PR #196, opened 2018-06-13
Add support for well-known type IDs

elprans:
Scalar types in `std` now have predetermined IDs.  This is useful for
binary driver implementations, as the codec mapping can be set up at
compile time.

We use the UUID variant 0 with sequential node id so that standard type
ids can be treated as simple integers starting from 1.

### DIFF
--- a/edb/api/__init__.py
+++ b/edb/api/__init__.py
@@ -1,0 +1,17 @@
+#
+# This source file is part of the EdgeDB open source project.
+#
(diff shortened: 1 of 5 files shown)


edb/lang/schema/std.py
```diff
@@ -27,6 +28,37 @@
 from . import delta as sd
 
 
+_TYPE_IDS = None
+
+
+def load_type_ids():
```
1st1: This exact code seems to be duplicated in `edb/lang/schema/objects.py`

PR #199, opened 2018-06-15
edgeql: Change syntax for empty tuple type.

vpetrovykh:
Empty tuple type is now indicated by `tuple<>`, whereas `tuple` without
angle brackets is now syntactically invalid.

### DIFF
--- a/edb/lang/edgeql/ast.py
+++ b/edb/lang/edgeql/ast.py
@@ -241,8 +241,8 @@ class _TypeName(Expr):
 class TypeName(_TypeName):
     name: str  # name is used for types in named tuples
     maintype: ObjectRef
-    subtypes: typing.Union[typing.List[_TypeName], None]
-    dimensions: typing.Union[typing.List[int], None]
+    subtypes: typing.Optional[typing.List[_TypeName]]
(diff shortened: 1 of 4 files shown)


edb/lang/edgeql/parser/grammar/expressions.py
```diff
@@ -1046,13 +1046,19 @@ def reduce_NodeName(self, *kids):
         maintype = kids[0].val
 
         # maintype cannot be 'array'
```
elprans: Fix the comment please. ("maintype cannot be a collection")

edb/lang/edgeql/ast.py
```diff
@@ -241,7 +241,7 @@ class _TypeName(Expr):
 class TypeName(_TypeName):
     name: str  # name is used for types in named tuples
     maintype: ObjectRef
-    subtypes: typing.List[_TypeName]
+    subtypes: typing.Union[typing.List[_TypeName], None]
```
1st1: `typing.Union[typing.List[_TypeName], None]` -> `typing.Optional[typing.List[_TypeName]]`

PR #205, opened 2018-06-23
graphql: Use a `filter` argument for more flexible filtering.

vpetrovykh:
Rather than taking arguments corresponding to the individual fields, use
a single `filter` argument to specify the filtering conditions. This is
more generic and easier to add more functionality to in the future.

Notice that this is based on top of the "gql-types" branch, so you may want to review that one first..

### DIFF
--- a/edb/lang/graphql/translator.py
+++ b/edb/lang/graphql/translator.py
@@ -42,6 +42,7 @@ class GraphQLTranslatorContext:
         self.vars = {}
         self.fields = []
         self.path = []
+        self.filter = None
         self.include_base = [False]
         self.gql_schema = gt.Schema(gqlcore)
         self.gqlcore_schema = gqlcore._gql_schema
@@ -263,10 +264,7 @@ class GraphQLTranslator(ast.NodeVisitor):
         target = self.get_field_type(
             prevt, node.name,
             args={
-                arg.name:
-                    (self._context.vars[arg.value.value]
-                     if isinstance(arg.value, gqlast.Variable) else
-                     arg.value.topython())
+                arg.name: self._get_field_arg_value(arg)
                 for arg in node.arguments
             },
             context=node.context)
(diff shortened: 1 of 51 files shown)


edb/lang/graphql/translator.py
```diff
@@ -487,26 +482,51 @@ def get_path_prefix(self, end_trim=None):
...
+
+        if op:
+            value = self.visit(node.value)
+
+            if name_parts == 'contains':
+                value.value = f'%{value.value}%'
+            elif name_parts == 'startswith':
+                value.value = f'{value.value}%'
+            elif name_parts == 'endswith':
+                value.value = f'%{value.value}'
```
1st1: These seem to be a special case of the "LIKE" operator. So I'd replace all of these with more generic "like" and "ilike" GraphQL filters.

edb/lang/graphql/types.py
```diff
@@ -89,8 +106,13 @@ def __init__(self, edb_schema):
...
             query=query,
-            types=[objt for name, objt in self._gql_objtypes.items()
-                   if name != 'Query'],
+            types=[
+                objt for name, objt in
+                sorted(itertools.chain(self._gql_objtypes.items(),
+                                       self._gql_inobjtypes.items()),
+                       key=lambda x: x[0])
+                if name != 'Query'
+            ],
```
1st1: This is a bit hard to read, I'd suggest to extract `sorted(...)` into a separate assignment statement `types = sorted(...)`.

1st1: From the performance point of view it makes more sense to first filter types and sort them later.

vpetrovykh: OK, I've refactored the code a little bit to make it less obscure.

edb/lang/graphql/translator.py
vpetrovykh: I've removed the "contains", "startswith" and "endswith" and replaced them with "like" and "ilike".

vpetrovykh:
I'm closing this one because I need to rebase it on "master" instead of the intermediate "gql-types".

PR #207, opened 2018-06-27
graphql: Use a `filter` argument for more flexible filtering.

vpetrovykh:
Rather than taking arguments corresponding to the individual fields, use
a single `filter` argument to specify the filtering conditions. This is
more generic and easier to add more functionality to in the future.

### DIFF
--- a/edb/lang/graphql/types.py
+++ b/edb/lang/graphql/types.py
@@ -23,7 +23,9 @@ from graphql import (
     GraphQLSchema,
     GraphQLObjectType,
     GraphQLInterfaceType,
+    GraphQLInputObjectType,
     GraphQLField,
+    GraphQLInputObjectField,
     GraphQLArgument,
     GraphQLList,
     GraphQLNonNull,
@@ -33,7 +35,9 @@ from graphql import (
     GraphQLBoolean,
     GraphQLID,
 )
+import itertools
 
+from edb.lang.common import ast
 from edb.lang.edgeql import ast as qlast
 from edb.lang.edgeql import codegen
 from edb.lang.edgeql.parser import parse_fragment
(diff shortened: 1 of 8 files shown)


edb/lang/graphql/types.py
```diff
@@ -153,36 +178,72 @@ def get_fields(self, typename):
...
+        comp = eq + ['gte', 'gt', 'lte', 'lt']
+        string = comp + ['like', 'ilike']
+
+        self._make_generic_input_type(GraphQLBoolean, *eq)
+        self._make_generic_input_type(GraphQLID, *eq)
+        self._make_generic_input_type(GraphQLInt, *comp)
+        self._make_generic_input_type(GraphQLFloat, *comp)
+        self._make_generic_input_type(GraphQLString, *string)
+
+    def _make_generic_input_type(self, base, *ops):
```
1st1: No need to use `*args` here as far as I can see.

edb/lang/schema/_graphql.eql
```diff
@@ -19,7 +19,7 @@
 
 CREATE FUNCTION graphql::short_name(std::str) -> std::str
     FROM SQL $$
-        SELECT regexp_replace($1, '.+?::(.+$)', '\1');
+        SELECT regexp_replace($1, '.+?::(.+$)', '\1') || 'Type';
```
1st1: Does this have to be an SQL function? Can you rewrite it to EdgeQL? (we'll disallow using SQL functions everywhere outside of `std.eql` soon).

vpetrovykh: OK, I'll need to expose `regexp_replace` in `std` first.

edb/lang/graphql/types.py
vpetrovykh: Oh, that's right, this is an unnecessary complication at this point.

PR #209, opened 2018-06-28
graphql: Add `orderBy` argument.

vpetrovykh:
All fields that return a list of objects can now be ordered by one or
more of the scalar fields on that object.


edb/lang/graphql/translator.py
```diff
@@ -452,12 +456,17 @@ def _validate_fragment_type(self, frag, spread):
...
-        if not f_arg:
-            return None
+    def _visit_arguments(self, arguments):
+        where = offset = limit = None
+        orderby = []
+
+        for arg in arguments:
+            if arg.name == 'filter':
+                where = self.visit(arg.value)
+            elif arg.name == 'orderBy':
```
1st1: Do we want `orderBy` or `sort`?  I've seen some implementations using the latter.  `sort` would be easier to type (shorter, no capital letters) and would read just fine too. @elprans thoughts?

edb/lang/graphql/types.py
```diff
@@ -238,11 +244,65 @@ def _make_generic_input_type(self, base, ops):
...
+                ),
+                nulls=GraphQLInputObjectField(
+                    self._gql_ordertypes['nullsOrderingEnum'],
+                    default_value='SMALLEST',
+                ),
+            )
+        )
+
+    @lru_cache(maxsize=None)
+    def get_order_fields(self, typename):
```
1st1: Are you sure it's a good idea to create an unbounded LRU cache?

vpetrovykh: Actually that's a copy-paste remnant of early incarnations where there was more re-use of these helper function results. It sort of made sense for `get_fields`, but doesn't make sense for `get_order_fields` or `get_filter_fields` (which are now called just once during init). I think I'll remove that altogether as optimizations are not yet the focus here.

edb/lang/graphql/translator.py
vpetrovykh: If we're considering `sort` maybe we should also consider `order`. The idea being that `order` is kinda referring to the keyword used in SQL and EdgeQL.

PR #210, opened 2018-06-28
graphql: Add slicing support.

vpetrovykh:
Add `after`, `first`, `before` and `last` arguments to anything that
returns an object list. The arguemnts have the same semantics as in
Relay Connections regardless of whether the field returns a Connection
or a plain List.

### DIFF
--- a/edb/lang/graphql/translator.py
+++ b/edb/lang/graphql/translator.py
@@ -29,7 +29,7 @@ from edb.lang.graphql import ast as gqlast, parser as gqlparser
 from edb.lang.schema import error as s_error
 
 from . import types as gt
-from .errors import GraphQLValidationError, GraphQLCoreError
+from . import errors as g_errors
 
 
 class GraphQLTranslatorContext:
@@ -44,8 +44,7 @@ class GraphQLTranslatorContext:
         self.path = []
         self.filter = None
         self.include_base = [False]
-        self.gql_schema = gt.Schema(gqlcore)
-        self.gqlcore_schema = gqlcore._gql_schema
+        self.gqlcore = gqlcore
         self.query = query
 
 
@@ -60,11 +59,11 @@ class GraphQLTranslator(ast.NodeVisitor):
         assert isinstance(name, str)
 
         try:
-            return self._context.gql_schema.get(name)
+            return self._context.gqlcore.get(name)
 
         except s_error.SchemaError:
             if context:
-                raise GraphQLValidationError(
+                raise g_errors.GraphQLValidationError(
                     f"{name!r} does not exist in the schema",
                     context=context)
             raise
@@ -79,7 +78,7 @@ class GraphQLTranslator(ast.NodeVisitor):
 
         if target is None:
             if context:
-                raise GraphQLValidationError(
+                raise g_errors.GraphQLValidationError(
                     f"field {name!r} is " +
                     f"invalid for {base.short_name}",
                     context=context)
@@ -94,7 +93,7 @@ class GraphQLTranslator(ast.NodeVisitor):
         }
 
         gqlresult = gql_proc(
-            self._context.gqlcore_schema,
+            self._context.gqlcore._gql_schema,
             self._context.query,
             variable_values={
                 name[1:]: val for name, val in self._context.variables.items()
@@ -104,7 +103,7 @@ class GraphQLTranslator(ast.NodeVisitor):
 
         if gqlresult.errors:
             for err in gqlresult.errors:
-                raise GraphQLCoreError(
+                raise g_errors.GraphQLCoreError(
                     err.message,
                     line=err.locations[0].line,
                     col=err.locations[0].column,
@@ -165,7 +164,7 @@ class GraphQLTranslator(ast.NodeVisitor):
             self.visit(node.variables)
 
         # base Query needs to be configured specially
-        base = self._context.gql_schema.get('Query')
+        base = self._context.gqlcore.get('Query')
 
         # special treatment of the selection_set, different from inner
         # recursion
@@ -202,7 +201,7 @@ class GraphQLTranslator(ast.NodeVisitor):
                     cond = cond.value
 
                 if not isinstance(cond, bool):
-                    raise GraphQLValidationError(
+                    raise g_errors.GraphQLValidationError(
                         f"'if' argument of {directive.name} " +
                         "directive must be a Boolean",
                         context=directive.context)
@@ -271,7 +270,7 @@ class GraphQLTranslator(ast.NodeVisitor):
         path.append(Step(name=node.name, type=target))
 
         if not top and fail:
-            raise GraphQLValidationError(
+            raise g_errors.GraphQLValidationError(
                 f"field {node.name!r} can only appear at the top-level Query",
                 context=node.context)
 
@@ -328,6 +327,7 @@ class GraphQLTranslator(ast.NodeVisitor):
             self._prepare_field(node)
 
         json_mode = False
+        is_shadowed = prevt.is_field_shadowed(node.name)
 
         # determine if there needs to be extra subqueries
         if not prevt.dummy and target.dummy:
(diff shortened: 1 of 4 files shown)


edb/lang/graphql/translator.py
```diff
@@ -459,12 +465,72 @@ def _validate_fragment_type(self, frag, spread):
...
+                    after = int(arg.value.value)
+                    assert after >= 0
+                    # The +1 is to make 'after' into an appropriate index.
+                    #
+                    # 0--a--1--b--2--c--3-- ... we call element at
+                    # index 0 (or "element 0" for short), the element
+                    # immediately after the mark 0. So after "element
+                    # 0" really means after "index 1".
+                    after += 1
+            except BaseException:
```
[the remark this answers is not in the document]
1st1: You shouldn't catch `BaseException` in Python unless you want to handle things like `SystemExit` and `KeyboardInterrupt`. So change it to `except Exception`.

`raise GraphQLValidationError from` is fine at least for now.

edb/lang/graphql/translator.py
```diff
@@ -459,12 +465,72 @@ def _validate_fragment_type(self, frag, spread):
...
+                limit = min(first, limit)
+        if last is not None:
+            if limit is not None:
+                if last < limit:
+                    offset = (offset or 0) + limit - last
+                    limit = last
+            else:
+                # there wasn't any limit, so we can define last in
+                # terms of offset alone
+                offset = parse_fragment(f'''
```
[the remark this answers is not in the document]
1st1: > I'm more convinced by the argument that it's easier to find missing features by expectedFailure marked tests than to remember places where the implementation is sub-optimal.

We have GitHub issues for tracking missing features.

edb/lang/graphql/translator.py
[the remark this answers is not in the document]
vpetrovykh: Ok, right. I'll amend that to `Exception`.

vpetrovykh:
I've removed `assert` and replaced their usage with exceptions.

edb/lang/graphql/translator.py
```diff
@@ -459,12 +464,68 @@ def _validate_fragment_type(self, frag, spread):
...
-                orderby = self.visit_order(arg.value)
+            try:
+                if arg.name == 'filter':
+                    where = self.visit(arg.value)
+                elif arg.name == 'order':
+                    orderby = self.visit_order(arg.value)
+                elif arg.name == 'first':
+                    first = arg.value.value
+                    if first < 0:
+                        raise ValueError(f"'{arg.name}' cannot be negative")
```
1st1: `raise ValueError(f"'{arg.name}' cannot be negative")` -> raise ValueError(f"{arg.name!r} cannot be negative")

edb/lang/graphql/translator.py
```diff
@@ -459,12 +464,68 @@ def _validate_fragment_type(self, frag, spread):
...
+        if last is not None:
+            if limit is not None:
+                if last < limit:
+                    offset = (offset or 0) + limit - last
+                    limit = last
+            else:
+                # FIXME: there wasn't any limit, so we can define last
+                # in terms of offset alone without negative OFFSET
+                # implementation
+                pass
```
1st1: Raise an exception here. Even `NotImplementedError` is better than silently ignoring the problem.

edb/lang/graphql/translator.py
```diff
@@ -459,12 +464,68 @@ def _validate_fragment_type(self, frag, spread):
...
+                    # The +1 is to make 'after' into an appropriate index.
+                    #
+                    # 0--a--1--b--2--c--3-- ... we call element at
+                    # index 0 (or "element 0" for short), the element
+                    # immediately after the mark 0. So after "element
+                    # 0" really means after "index 1".
+                    after += 1
+            except Exception:
+                raise GraphQLValidationError(
+                    f"invalid value for '{arg.name}': {arg.value.value!r}",
```
1st1: `{arg.name!r}`

edb/lang/graphql/types.py
```diff
@@ -656,7 +664,8 @@ def get_field_type(self, name, argsmap=None):
         fkey = get_fkey(name, argsmap)
         target = None
 
-        assert isinstance(name, str)
+        if not isinstance(name, str):
+            raise Exception(f"field name must be a string")
```
1st1: f"field name must be a str, got {type(name).__name__}"

edb/lang/graphql/types.py
```diff
@@ -468,8 +474,10 @@ def __init__(self, schema, **kwargs):
         if edb_base is None and self.edb_type:
             edb_base = schema.edb_schema.get(self.edb_type)
 
-        assert edb_base is not None
-        assert schema is not None
+        if edb_base is None:
+            raise Exception("cannot determine the EdgeDB base type")
```
1st1: Can you add a new exception class `GraphQLTranslationError` and use it in places like this?

edb/lang/graphql/translator.py
vpetrovykh: OK.

edb/lang/graphql/translator.py
vpetrovykh: Right, `NotImplementedError` fits well here.

edb/lang/graphql/translator.py
vpetrovykh: OK.

edb/lang/graphql/types.py
vpetrovykh: OK.

edb/lang/graphql/types.py
vpetrovykh: Good point, I'll add `GraphQLTranslationError` for this stuff.

vpetrovykh:
I've added the changes from the comments.

edb/lang/graphql/translator.py
```diff
@@ -459,12 +464,68 @@ def _validate_fragment_type(self, frag, spread):
...
+                    #
+                    # 0--a--1--b--2--c--3-- ... we call element at
+                    # index 0 (or "element 0" for short), the element
+                    # immediately after the mark 0. So after "element
+                    # 0" really means after "index 1".
+                    after += 1
+            except Exception:
+                raise GraphQLValidationError(
+                    f"invalid value for {arg.name!r}: {arg.value.value!r}",
+                    context=arg.context)
```
1st1: `raise GraphQLValidationError(...) from None`

edb/lang/graphql/translator.py
```diff
@@ -459,12 +464,68 @@ def _validate_fragment_type(self, frag, spread):
...
+        if last is not None:
+            if limit is not None:
+                if last < limit:
+                    offset = (offset or 0) + limit - last
+                    limit = last
+            else:
+                # FIXME: there wasn't any limit, so we can define last
+                # in terms of offset alone without negative OFFSET
+                # implementation
+                raise NotImplementedError
```
1st1: I suggest to raise the newly added `GraphQLTranslationError('negative offsets are not supported')`

edb/lang/graphql/types.py
```diff
@@ -468,8 +474,11 @@ def __init__(self, schema, **kwargs):
         if edb_base is None and self.edb_type:
             edb_base = schema.edb_schema.get(self.edb_type)
 
-        assert edb_base is not None
-        assert schema is not None
+        if edb_base is None:
+            raise GraphQLTranslationError(
+                "cannot determine the EdgeDB base type")
```
1st1: `f"cannot determine the EdgeDB base type for {self.edb_type.__name__}"`

^ do we have a unittest for this?

vpetrovykh: Good point, let me see if I can still reach this exception. It may be a relic of early versions.

edb/lang/graphql/translator.py
vpetrovykh: This isn't accurate. This exception would be raised on otherwise legal GraphQL (e.g. argument `last: 3`). Technically the argument is fine according to the spec and semantically valid. It is not implemented because ideally we'd rather implement it using EdgeQL `OFFSET -3`. That is an implementation details, though. Negative `OFFSET` is an EdgeQL feature and it currently is not implemented. The end user doesn't necessarily know anything about negative offsets.

edb/lang/graphql/types.py
1st1: Looks like `schema.get()` cannot return `None`.

edb/lang/graphql/types.py
```diff
@@ -468,8 +474,11 @@ def __init__(self, schema, **kwargs):
...
+        if schema is None:
+            raise GraphQLTranslationError("schema is missing")
```
1st1: You use `schema` a few lines above, this check is pointless.

edb/lang/graphql/translator.py
1st1: Sure, then raise 

```python
raise GraphQLTranslationError(
    f'limit={limit} and offset={offset}  translate to a negative OFFSET in EdgeQL '
    f'which is currently unsupported')
```

edb/lang/graphql/types.py
elprans: `schema.get()` without a `default=` raises a `SchemaError`.

edb/lang/graphql/translator.py
1st1: I mean this isn't something that we'll have for a long time, but if someone encounters this exception it's good that they will know roughly what's going on (and that the error isn't their fault).

edb/lang/graphql/types.py
```diff
@@ -656,7 +665,9 @@ def get_field_type(self, name, argsmap=None):
         fkey = get_fkey(name, argsmap)
         target = None
 
-        assert isinstance(name, str)
+        if not isinstance(name, str):
+            raise GraphQLTranslationError(
+                f"field name must be a str, got {type(name).__name__}")
```
1st1: Maybe this check should be before the `fkey = get_fkey(...)` line?

edb/lang/graphql/types.py
```diff
@@ -443,6 +405,40 @@ def get(self, name, **kwargs):
...
+        return tuple(result)
+
+
+class CompleteShadowingFields:
+    def __contains__(self, item):
+        return True
+
+
+ALL_FIELDS = CompleteShadowingFields()
+NO_FIELDS = set()
```
1st1: Should a `frozenset()` be used here? Can you add a comment clarifying these?

edb/lang/graphql/types.py
```diff
@@ -456,6 +418,20 @@ def __new__(mcls, name, bases, dct):
         return cls
 
 
+class CompleteShadowingFields:
```
1st1: This class name is weird and deserves a docstring.

vpetrovykh: I've thrown out that class and updated the code a bit to do without that extra complication.

Issue #202, opened 2018-06-20
Empty set results in a tuple with a `None` element.


vpetrovykh:
It is currently possible to create a tuple with `None` as one of its elements from an `{}`.

Tests involving this issue: 
- [ ] `test_edgeql_expr_tuple_14` created directly from literals
- [ ] `test_edgeql_expr_tuple_15` created from views using literals
- [ ] `test_edgeql_scope_tuple_12` created using an empty link property

PR #216, opened 2018-07-02
Strengthen nullability checks in the query compiler

elprans:
This round of fixes addresses specifically the handling of empty set
literals and empty set expressions within tuples.

Fixes: #202.

### DIFF
--- a/edb/server/pgsql/ast.py
+++ b/edb/server/pgsql/ast.py
@@ -126,7 +126,6 @@ class BaseRangeVar(Base):
     """Range variable, used in FROM clauses."""
 
     alias: Alias
-    nullable: bool
 
     @property
     def is_distinct(self):
@@ -154,8 +153,6 @@ RangeTypes = typing.Union[BaseRangeVar, _Ref]
 
 class BaseRelation(Base):
     name: str
-    # Whether the query output can contain NULLs.
-    nullable: bool = True
 
 
 class Relation(BaseRelation, EdgeQLPathInfo):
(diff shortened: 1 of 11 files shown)


edb/server/pgsql/compiler/pathctx.py
```diff
@@ -766,3 +811,22 @@ def get_path_output_or_null(
...
+        if isinstance(expr, pgast.Query):
+            tl_len = len(expr.target_list)
+            if tl_len != 1:
+                raise RuntimeError(
+                    f'subquery used as a value returns {tl_len} columns')
+
+            return is_nullable(expr.target_list[0].val, env=env)
+        else:
+            import edb.lang.common.markup
+            edb.lang.common.markup.dump(expr)
```
1st1: debug code?

edb/server/pgsql/ast.py
```diff
@@ -336,6 +337,10 @@ class DML(Base):
     # List of expressions returned
     returning_list: typing.List[ResTarget]
 
+    @property
+    def target_list(self):
+        return self.returning_list
```
1st1: Maybe it's better to define an `is_nullable()` method?

1st1:
LGTM, but please take a look at the review nit-comments before merging.

edb/server/pgsql/ast.py
elprans: This is not related to nullability. This is to avoid having to do `if isinstance(DML): returning_list.append(...) else target_list.append(...)`

1st1: It's weird that a base class makes `target_list` and `returning_list` the same, even though they have distinctly different names with a different meaning, no?

PR #218, opened 2018-07-04
Postpone expression cardinality analysis until scope tree is complete

elprans:
Currently, cardinality inference is happening at the same time as the
expression is compiled.  This is problematic, since there are cases
where the scope tree can be modified by further processing of the
expression, which could affect the inference result.  The most common
such case is expressions in tuples.

To avoid possible inconsistencies, postpone the cardinality inference
until the very end of statement compilation when the scope tree is
complete and final.

### DIFF
--- a/edb/lang/edgeql/compiler/pathctx.py
+++ b/edb/lang/edgeql/compiler/pathctx.py
@@ -23,7 +23,6 @@
 import typing
 
 from edb.lang.ir import ast as irast
-from edb.lang.ir import inference as irinference
 
 from edb.lang.schema import objects as s_obj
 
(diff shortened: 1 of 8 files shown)


edb/lang/edgeql/compiler/pathctx.py
```diff
@@ -91,14 +91,9 @@ def set_path_alias(
...
-    if cardinality != irast.Cardinality.ONE:
-        raise errors.EdgeQLError(
-            'possibly more than one element returned by an expression '
-            'where only singletons are allowed',
-            context=expr.context)
+        expr: irast.Base, *,
+        scope: typing.Optional[irast.ScopeTreeNode]=None,
+        ctx: context.ContextLevel) -> irast.Cardinality:
+    if scope is None:
+        scope = ctx.path_scope
```
1st1: I'd have "scope" and "schema" parameters (so no "ctx") for this function.

edb/lang/edgeql/compiler/stmtctx.py
```diff
@@ -76,6 +79,9 @@ def init_context(
 def fini_expression(
         ir: irast.Base, *,
         ctx: context.ContextLevel) -> irast.Command:
+    for cb in ctx.completion_work:
+        cb(ctx=ctx)
+
```
1st1: Add `ctx.completion_work.clear()`?

edb/lang/edgeql/compiler/pathctx.py
elprans: At this point it devolved into a pointless wrapper around the general `infer_cardinality`, so I'll just drop it altogether.

PR #220, opened 2018-07-11
edgeql: Disallow nested arrays.

vpetrovykh:
Currently nested arrays are disallowed in EdgeDB, so these tests should
all come up with errors.

### DIFF
--- a/edb/lang/edgeql/compiler/func.py
+++ b/edb/lang/edgeql/compiler/func.py
@@ -63,8 +63,10 @@ def compile_FunctionCall(
         fctx.in_func_call = True
         args, kwargs, arg_types = process_func_args(expr, funcname, ctx=fctx)
 
+        fatal_array_check = len(funcs) == 1
         for funcobj in funcs:
-            if check_function(funcobj, arg_types):
+            if check_function(expr, funcname, funcobj, arg_types,
+                              fatal_array_check=fatal_array_check):
                 break
         else:
             raise errors.EdgeQLError(
@@ -88,8 +90,11 @@ def compile_FunctionCall(
 
 
 def check_function(
+        expr: qlast.FunctionCall,
+        funcname: sn.Name,
         func: s_func.Function,
-        arg_types: typing.Iterable[s_obj.Object]) -> bool:
+        arg_types: typing.Iterable[s_obj.Object],
+        fatal_array_check: bool = False) -> bool:
     if not func.paramtypes:
         if not arg_types:
             # Match: `func` is a function without parameters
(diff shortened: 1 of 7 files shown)


elprans:
The fix for this would be to prohibit passing `array` to any function declared as `func(any) -> array<any>`. This can be done in the function handler of the IR compiler.

edb/lang/edgeql/compiler/func.py
```diff
@@ -63,8 +63,10 @@ def compile_FunctionCall(
         fctx.in_func_call = True
         args, kwargs, arg_types = process_func_args(expr, funcname, ctx=fctx)
 
+        fatal_array_check = len(funcs) == 1
```
1st1: I don't like this name. How about `nested_array_check`

edb/lang/edgeql/compiler/func.py
```diff
@@ -133,6 +139,24 @@ def check_function(
...
+            # 'array<any>', and the argument is also an 'array', then
+            # this is an invalid function invocation.
+            rt = func.returntype
+
+            if (pt.name == 'std::any' and
+                    isinstance(rt, s_types.Array) and
+                    rt.get_subtypes()[0].name == 'std::any' and
+                    isinstance(at, s_types.Array)):
+
+                if fatal_array_check:
```
1st1: Re-arrange IFs:

```python
if (pt.name == 'std::any' and
        fatal_array_check and
        ...)
```

tests/test_edgeql_functions.py
```diff
@@ -279,6 +281,17 @@ class TestEdgeQLFunctions(tb.QueryTestCase):
             [[{'name': 'Elvis'}, {'name': 'Yury'}]]
         ])
 
+    async def test_edgeql_functions_array_agg_13(self):
+        # FIXME: please update the error message once an error is
+        # actually produced
```
1st1: Outdated comment?

tests/test_edgeql_expressions.py
```diff
@@ -796,6 +796,44 @@ class TestExpressions(tb.QueryTestCase):
             [],
         ])
 
+    @unittest.expectedFailure
+    async def test_edgeql_expr_array_10(self):
+        # FIXME: please update the error message once an error is
+        # actually produced
+        with self.assertRaisesRegex(exc.EdgeQLError, 'nested array'):
```
elprans: This should be easy to fix in the array literal handler.

tests/test_edgeql_expressions.py
```diff
@@ -796,6 +796,44 @@ class TestExpressions(tb.QueryTestCase):
...
+                SELECT [[1, 2], [3, 4]];
+            ''')
+
+    @unittest.expectedFailure
+    async def test_edgeql_expr_array_11(self):
+        # FIXME: please update the error message once an error is
+        # actually produced
+        with self.assertRaisesRegex(exc.EdgeQLError, 'nested array'):
+            await self.con.execute(r'''
+                SELECT [array_agg({1, 2})];
```
elprans: Ditto

vpetrovykh:
OK, I've updated the literal handler to check for nested arrays.

PR #223, opened 2018-07-17
edgeql: Update the type specification syntax.

vpetrovykh:
Introduce type expressions - syntactical constructs denoting types and
type composition.

Operator `|` produces a union type that is a duck type having the
intersection of all of the links and properties of the component types.
Semantically any object that is an instance of at least one of the
component types is also an instance of the union type.

Operator `&` produces an intersection type that is a duck type having
the union of all of the links and properties of the component types.
Semantically only an object that is an instance of all of the component
types is also an instance of the intersection type.

Syntactically type expressions are general and can be used anywhere
where a type can be used. There may be some exceptions, notably only
base types can be used in `extending` specification.


edb/lang/edgeql/ast.py
```diff
@@ -67,6 +67,7 @@ class SetOperator(EdgeQLOperator):
 ILIKE = EdgeQLMatchOperator('ILIKE')
 NOT_ILIKE = EdgeQLMatchOperator('NOT ILIKE')
 
+# FIXME: are these even valid any more?
```
elprans: No, these are obsolete and unused.

edb/lang/edgeql/ast.py
```diff
@@ -235,45 +244,66 @@ class UnaryOp(Expr):
...
-class _TypeName(Expr):
-    pass
+class TypeOf(TypeExpr):
+    expr: Expr
 
 
-class TypeName(_TypeName):
-    name: str  # name is used for types in named tuples
+class TypeName(TypeExpr):
+    name: str  # XXX: name is used for types in named tuples
```
elprans: Why `XXX`?

vpetrovykh: Oh, that's right, I wanted to come back and fix it in the same fashion as a couple of other things in the parser. It should be a named tuple for parser use, rather than weird reuse of the TypeName.

edb/lang/edgeql/codegen.py
```diff
@@ -586,11 +600,22 @@ def visit_NoneTest(self, node):
...
+        )
+        if parenthesize:
+            self.write('(')
         if node.name is not None:
             self.write(ident_to_str(node.name), ': ')
         if isinstance(node.maintype, edgeql_ast.Path):
             self.visit(node.maintype)
         else:
+
+            if node.maintype is None:
```
elprans: leftover debug

edb/lang/edgeql/ast.py
vpetrovykh: Hm, I thought it was something else, nevermind. The comment is accurate and the `XXX` is not necessary.

edb/lang/edgeql/ast.py
vpetrovykh: OK, removing.

edb/lang/edgeql/compiler/expr.py
```diff
@@ -82,6 +80,21 @@ def compile_BinOp(
...
 
 
+@dispatch.compile.register(qlast.IsOp)
+def compile_IsOp(
+        expr: qlast.Base, *, ctx: context.ContextLevel) -> irast.Set:
+    try_folding = True
+
+    op_node = compile_type_check_op(expr, ctx=ctx)
+
+    if try_folding:
```
elprans: This check is always `True`, remove the `if`.

edb/lang/edgeql/compiler/schemactx.py
```diff
@@ -40,12 +40,16 @@
...
 
 def get_schema_object(
-        name: typing.Union[str, qlast.ObjectRef],
+        name: typing.Union[str, qlast.ObjectRef, qlast.TypeName],
         module: typing.Optional[str]=None, *,
         item_types: typing.Optional[typing.List[s_obj.ObjectMeta]],
         ctx: context.ContextLevel,
         srcctx: typing.Optional[parsing.ParserContext] = None) -> s_obj.Object:
 
+    if isinstance(name, qlast.TypeName):
```
elprans: It would be better to do this at call site, where there is more context as to what is going on.

edb/lang/edgeql/compiler/viewgen.py
```diff
@@ -173,11 +173,18 @@ def _normalize_view_ptr_expr(
     qlexpr = None
 
     if is_polymorphic:
-        source = qlast.TypeFilter(
-            expr=qlast.Path(steps=[qlast.Source()]),
-            type=qlast.TypeName(maintype=steps[0]))
+        ptype = steps[0]
+        # FIXME: what are the options here?
+        if isinstance(ptype, qlast.TypeName):
```
elprans: Wouldn't it be `TypeExpr` here?

edb/lang/edgeql/parser/grammar/expressions.py
```diff
@@ -314,11 +314,12 @@ def reduce_ShapeElementWithSubShape(self, *kids):
 
         shape = kids[1].val
 
-        if shape and isinstance(shape[0], qlast.ObjectRef):
+        # shape elements can have a path starting with a TypeName,
+        # this indicates a polymorphic shape and TypeName must be
+        # extracted from the path steps
+        if shape and isinstance(shape[0], qlast.TypeName):
```
elprans: Polymorphic shapes should support `FullTypeExpr`

tests/test_edgeql_syntax.py
```diff
@@ -2368,39 +2385,33 @@ def test_edgeql_syntax_ddl_attribute_05(self):
         CREATE ABSTRACT ATTRIBUTE std::foo tuple<Baz, tuple<int64, foo::Bar>>;
         """
 
-    @tb.must_fail(errors.EdgeQLSyntaxError,
-                  r"Unexpected 'array'", line=2, col=44)
+    # illegal semantically, but syntactically valid
```
elprans: Do we have tests for "semantic illegality" of this?

edb/lang/edgeql/compiler/expr.py
```diff
@@ -573,13 +586,9 @@ def try_fold_binop(
 
 
 def compile_type_check_op(
-        expr: qlast.BinOp, *, ctx: context.ContextLevel) -> irast.TypeCheckOp:
-    # <Expr> IS <Type>
+        expr: qlast.IsOp, *, ctx: context.ContextLevel) -> irast.TypeCheckOp:
+    # <Expr> IS <TypeExpr>
     left = dispatch.compile(expr.left, ctx=ctx)
-    with ctx.new() as subctx:
-        subctx.path_as_type = True
```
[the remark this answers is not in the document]
vpetrovykh: Dropped.

edb/lang/edgeql/compiler/expr.py
vpetrovykh: OK.

edb/lang/edgeql/parser/grammar/expressions.py
vpetrovykh: OK, changing to `TypeExpr`.

edb/lang/edgeql/compiler/viewgen.py
vpetrovykh: Yeah. I'm also not sure there's an `if..else` situation there. A polymorphic query *would* have a TypeExpr and that's that. So I'll use an `isinstance` test instead of a `len` test.

edb/lang/edgeql/compiler/schemactx.py
vpetrovykh: OK, I'm moving this bit of logic out to the call sites.

vpetrovykh:
I've just manually squashed all the fixes into the main commit. Once it clears the tests, I'll merge it to master.

Issue #217, opened 2018-07-03
Add support for the validated "required" qualifier in computable declarations


vpetrovykh:
A `REQUIRED` expression would be statically validated to be non-empty. This enables such things as required computables, which is useful for types and views. With this feature views can be reasoned about in a similar fashion to types with required links and properties.

e.g.
```
# if friends are not a required link, this would be valid
SELECT User {
    name,
    foo := REQUIRED (User.friends.name ?? 'n/a')
};

# assuming all links and properties are already required this is valid
SELECT Issue {
    title,
    foo := REQUIRED Issue.owner.email
};
```

An additional benefit is that if the schema changes and some required links or properties stop being required (say, `age` becomes non-required), the views and expressions that relied on this will fail compile-time validation and thus provide an early indication of something being broken.

Syntactically `REQUIRED` keyword would be used similarly to `DETACHED`.

elprans:
With the rework of cardinality specification in #303, the `REQUIRED` as an expression is not necessary, and should be expressed as a `required` qualifier on the computable declaration, in line with how regular links and properties are defined:

```
SELECT Issue {
    required foo := Issue.owner.email 
};
```

1st1:
@elprans was this implemented in the linked PR? Close if it was.

elprans:
@vpetrovykh, please look into writing some tests for this.

PR #224, opened 2018-07-20
edgeql: Add `REQUIRED expr` syntax.

vpetrovykh:
Issue #217.


edb/lang/edgeql/parser/grammar/precedence.py
```diff
@@ -109,6 +109,22 @@ class P_TYPEAND(Precedence, assoc='left', tokens=('AMPER',)):
     pass
 
 
+class P_EXISTS(Precedence, assoc='right', tokens=('EXISTS',)):
```
elprans: Please make these all equal (`rel_to_last='='`)

vpetrovykh: OK, I've updated the precedence.

PR #232, opened 2018-08-15
Untangle handling of scope in views

elprans:
This round fixes a number of long-standing warts in the handling of
complex scope scenarios in the presence of query iterator
(`FOR` statements), inline expression aliases (`SELECT foo := ...`)
and computables in general.

The main changes here are improving the consistency of scope namespace
handling, as well as explicit disambiguation of computable pointer
paths, so in `SELECT User { name := lower(User.name) }` the shape
element `User.name` and the `User.name` reference in the computable
expression have different path ids.

Inline expression aliases are no longer treated as equivalent to a
`WITH` clause view, because that breaks `FOR` statements
(the iterator alias follows the `WITH` clause logically, hence
the views there cannot refer to the iterator symbol).  The SQL compiler
has also been fixed to handle shaped computables containing a `FOR`
expression.

Finally, view types defined by nested shapes are no longer created
twice, which should improve compiler performance on queries with lots of
nested shapes.

### DIFF
--- a/edb/lang/edgeql/compiler/context.py
+++ b/edb/lang/edgeql/compiler/context.py
@@ -112,8 +112,16 @@ class ContextLevel(compiler.ContextLevel):
     aliased_views: typing.Dict[str, s_nodes.Node]
     """A dictionary of views aliased in a statement body."""
 
-    view_class_map: typing.Dict[s_nodes.Node, s_nodes.Node]  # noqa
-    """Object mapping (used by schema-level views)."""
+    schema_view_cache: typing.Dict[s_nodes.Node, s_nodes.Node]  # noqa
+    """Type cache used by schema-level views."""
+
+    expr_view_cache: typing.Dict[typing.Tuple[qlast.Base, str],
+                                 irast.Set]  # noqa
+    """Type cache used by expression-level views."""
+
+    shape_type_cache: typing.Dict[typing.Tuple[qlast.ShapeElement, ...],
+                                  s_nodes.Node]
+    """Type cache for shape expressions."""
 
     class_view_overrides: typing.Dict[s_name.SchemaName, s_nodes.Node]  # noqa
     """Object mapping used by implicit view override in SELECT."""
@@ -208,7 +216,9 @@ class ContextLevel(compiler.ContextLevel):
             self.view_nodes = {}
             self.view_sets = {}
             self.aliased_views = collections.ChainMap()
-            self.view_class_map = {}
+            self.schema_view_cache = {}
+            self.expr_view_cache = {}
+            self.shape_type_cache = {}
             self.class_view_overrides = {}
 
             self.clause = None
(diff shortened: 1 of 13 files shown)


edb/lang/edgeql/compiler/context.py
```diff
@@ -112,8 +112,16 @@ class ContextLevel(compiler.ContextLevel):
...
-    view_class_map: typing.Dict[s_nodes.Node, s_nodes.Node]  # noqa
-    """Object mapping (used by schema-level views)."""
+    schema_view_cache: typing.Dict[s_nodes.Node, s_nodes.Node]  # noqa
+    """Type cache used by schema-level views."""
+
+    expr_view_cache: typing.Dict[typing.Tuple[qlast.Base, str],
+                                 irast.Set]  # noqa
+    """Type cache used by expression-level views."""
+
+    shape_type_cache: typing.Dict[typing.Tuple[qlast.ShapeElement],
```
1st1: Is key a tuple of single `qlast.ShapeElement` element?

edb/lang/edgeql/compiler/setgen.py
```diff
@@ -435,7 +438,7 @@ def _is_computable_ptr(
         force_computable: bool=False,
         ctx: context.ContextLevel) -> bool:
     try:
-        qlexpr, _, _ = ctx.source_map[ptrcls]
+        qlexpr, *_ = ctx.source_map[ptrcls]
```
1st1: `ctx.source_map[ptrcls][0]`?

edb/lang/edgeql/compiler/setgen.py
```diff
@@ -630,10 +633,6 @@ def computable_ptr_set(
         ctx: context.ContextLevel) -> irast.Set:
     """Return ir.Set for a pointer defined as a computable."""
     ptrcls = rptr.ptrcls
-
-    # Must use an entirely separate context, as the computable
-    # expression is totally independent from the surrounding query.
-    subctx = stmtctx.init_context(schema=ctx.schema)
     self_ = rptr.source
```
1st1: Maybe we can use a bit less confusing name than `self_`?

edb/lang/edgeql/compiler/setgen.py
```diff
@@ -765,3 +719,73 @@ def _check_cardinality(ctx):
...
+        rptr: irast.Pointer,
+        source: irast.Set,
+        source_scls: s_nodes.Node,
+        inner_source_path_id: irast.PathId,
+        qlctx: context.ContextLevel,
+        ctx: context.ContextLevel) -> typing.ContextManager:
+    @contextlib.contextmanager
+    def newctx():
+        with ctx.new() as subctx:
+            subctx.class_view_overrides = {}
```
1st1: Would it make sense to make `_get_computable_ctx` a method on the Context object?

edb/lang/edgeql/compiler/context.py
elprans: Good catch. It's a tuple of many.

edb/lang/edgeql/compiler/setgen.py
elprans: Fixed.

edb/lang/edgeql/compiler/setgen.py
elprans: Renamed to `source_set`

edb/lang/edgeql/compiler/setgen.py
elprans: Not really.  I strive to keep `Context` as close to a dumb dataclass as possible with only very simple logic.

PR #240, opened 2018-10-05
Syntax

vpetrovykh:
Syntax changes regarding `ROLE` and `ON TARGET DELETE`.


edb/lang/edgeql/ast.py
```diff
@@ -127,6 +127,14 @@ class Cardinality(s_enum.StrEnum):
     DEFAULT = ''
 
 
+class Cascade(s_enum.StrEnum):
```
elprans: I think `LinkTargetDeleteAction` is a more appropriate name for this enumeration.

edb/lang/schema/parser/grammar/declarations.py
```diff
@@ -904,6 +907,7 @@ def _process_pointerspec(self, p: PointerSpec):
         attributes = []
         policies = []
         properties = []
+        on_delete = []
```
elprans: `on_delete` shouldn't be a list.

edb/lang/schema/parser/grammar/declarations.py
```diff
@@ -914,6 +918,8 @@ def _process_pointerspec(self, p: PointerSpec):
                 properties.append(spec)
             elif isinstance(spec, esast.Policy):
                 policies.append(spec)
+            elif isinstance(spec, esast.OnTargetDelete):
+                on_delete.append(spec)
```
elprans: Can we reuse the enum from `edgeql.ast`?

vpetrovykh:
I've amended the commits to reflect the changes @elprans suggested.

elprans:
LGTM

PR #244, opened 2018-10-12
Add some JSON functions to std

vpetrovykh:
Added functions:
- `json_typeof`
- `json_array_elements`
- `json_array_unpack`
- `json_object_unpack`
- `json_nav`

Also add support for checking index bounds and `[]` for JSON.

### DIFF
--- a/edb/lang/schema/_std.eql
+++ b/edb/lang/schema/_std.eql
@@ -204,6 +204,31 @@ CREATE FUNCTION std::current_datetime() -> std::datetime
 CREATE FUNCTION std::uuid_generate_v1mc() -> std::uuid
     FROM SQL 'SELECT edgedb.uuid_generate_v1mc()';
 
+# JSON functions
+CREATE FUNCTION std::json_typeof($json: std::json) -> std::str
+    FROM SQL FUNCTION 'jsonb_typeof';
+
+CREATE FUNCTION std::json_array_unpack($arr: std::json) -> SET OF std::json
+    FROM SQL FUNCTION 'jsonb_array_elements';
+
+CREATE FUNCTION std::json_object_unpack($obj: std::json) ->
+        SET OF tuple<std::str, std::json>
+    FROM SQL FUNCTION 'jsonb_each';
+
+CREATE FUNCTION std::json_get($json: std::json, VARIADIC $path: std::str) ->
+        OPTIONAL std::json
+    FROM SQL FUNCTION 'jsonb_extract_path';
+
+CREATE FUNCTION std::json_to_str($json: std::json) -> std::str
+    FROM SQL $$
+        SELECT $1::text
+    $$;
+
+CREATE FUNCTION std::str_to_json($str: std::str) -> std::json
+    FROM SQL $$
+        SELECT $1::jsonb
+    $$;
+
 CREATE ABSTRACT CONSTRAINT std::constraint
 {
     SET errmessage := 'invalid {__subject__}';
--- a/edb/server/pgsql/compiler/expr.py
+++ b/edb/server/pgsql/compiler/expr.py
@@ -19,6 +19,7 @@
 
 """Compilation handlers for non-statement expressions."""
 
+import json
 import typing
 
 from edb.lang.common import ast
@@ -159,11 +160,17 @@ def compile_TypeCast(
 @dispatch.compile.register(irast.IndexIndirection)
 def compile_IndexIndirection(
         expr: irast.Base, *, ctx: context.CompilerContextLevel) -> pgast.Base:
-    # Handle Expr[Index], where Expr may be std::str or array<T>.
-    # For strings we translate this into substr calls, whereas
-    # for arrays the native slice syntax is used.
+    # Handle Expr[Index], where Expr may be std::str, array<T> or
+    # std::json. For strings we translate this into substr calls,
+    # whereas for arrays the native index access is used. JSON is
+    # handled by using the `->` accessor. Additionally, in all of the
+    # above cases a boundary-check is performed on the index and an
+    # exception is potentially raised.
     is_string = False
     arg_type = _infer_type(expr.expr, ctx=ctx)
+    # line, column and filename are captured here to be used with the
+    # error message
+    exc_details = get_exc_details(expr.index)
 
     with ctx.new() as subctx:
         subctx.expr_exposed = False
@@ -172,6 +179,28 @@ def compile_IndexIndirection(
 
     if isinstance(arg_type, s_scalars.ScalarType):
         b = arg_type.get_topmost_concrete_base()
+
+        # JSON accessors already handle negative offset and count from
+        # 0, so we don't need any of the additional processing of the
+        # index expression that str and arrays require.
+        if b.name == 'std::json':
+            # At this point the index is either a text or some
+            # integer. If the index is an integer, cast it into int.
+            index_t = _infer_type(expr.index, ctx=ctx)
+            int_t = ctx.env.schema.get('std::anyint')
+            if index_t.issubclass(int_t):
+                index = pgast.TypeCast(
+                    arg=index,
+                    type_name=pgast.TypeName(
+                        name=('int',)
+                    )
+                )
+
+            return pgast.FuncCall(
+                name=('edgedb', '_json_index'),
+                args=[subj, index, exc_details]
+            )
+
         is_string = b.name == 'std::str'
 
     one = pgast.Constant(val=1)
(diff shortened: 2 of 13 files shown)


edb/lang/schema/_std.eql
```diff
@@ -185,6 +185,26 @@ CREATE FUNCTION std::current_datetime() -> std::datetime
 CREATE FUNCTION std::uuid_generate_v1mc() -> std::uuid
     FROM SQL 'SELECT edgedb.uuid_generate_v1mc()';
 
+# JSON functions
+CREATE FUNCTION std::json_typeof(std::json) -> std::str
+    FROM SQL FUNCTION 'jsonb_typeof';
+
+CREATE FUNCTION std::json_array_elements(std::json) -> array<std::json>
```
elprans: I think `json_array_as_array` is a better name for this.

edb/server/pgsql/metaschema.py
```diff
@@ -919,6 +978,109 @@ def __init__(self):
...
+            returns=('text',),
+            volatility='immutable',
+            text=self.text)
+
+
+class JSONIndexWithBoundsFunction(dbops.Function):
+    """Get a JSON element or raise an out-of-bounds exception."""
+    text = r'''
+        SELECT
+            CASE WHEN jsonb_typeof(val) = 'object' THEN (
```
1st1:
```
CASE jsonb_typeof(val)
WHEN 'object' THEN
...
WHEN 'text' THEN
...
```

edb/server/pgsql/metaschema.py
```diff
@@ -919,6 +978,109 @@ def __init__(self):
...
+                CASE WHEN pg_typeof(index)::text = 'text' THEN
```
1st1: I'd use `IF` here for clarity.

edb/server/pgsql/metaschema.py
```diff
@@ -122,20 +122,77 @@ class RaiseExceptionFunction(dbops.Function):
...
+            name=('edgedb', '_raise_exception_on_null'),
+            args=[('val', ('anyelement',)), ('exc', ('text',)),
+                  ('msg', ('text',)), ('det', ('text',))],
+            returns=('anyelement',),
+            volatility='immutable',
+            text=self.text)
+
+
+class RaiseExceptionOnEmptyStringFunction(dbops.Function):
+    """Return the passed string or raise an exception if it's n empty."""
```
elprans: typo: `it's n empty`

tests/test_edgeql_expressions.py
```diff
@@ -834,6 +835,24 @@ class TestExpressions(tb.QueryTestCase):
...
+                r'array index 10 is out of bounds'):
+            await self.con.execute("""
+                SELECT [1, 2, 3][10];
+            """)
+
+    async def test_edgeql_expr_array_16(self):
+        with self.assertRaisesRegex(
+                # FIXME: possibly a different error should be used here
+                exc.UnknownEdgeDBError,
+                r'array index -7 is out of bounds'):
```
elprans: Why does it say `index -7` when you're passing `-10`?

tests/test_edgeql_expressions.py
```diff
@@ -906,189 +925,28 @@ class TestExpressions(tb.QueryTestCase):
...
-            [True],
-            [True],
-            [True],
-        ])
+    async def test_edgeql_expr_string_04(self):
+        # There's an inherent problem with negative offsets that they
+        # resolve into some other offset that gets passed to substr.
+        # So potentially this can be prettified by introducing an
+        # extra layer of a function call or some other mechanism for
+        # storing the "original" offset value.
```
elprans: In this case it's better to give _no_ index at all than to give a misleading value.

edb/server/_testbase.py
```diff
@@ -384,7 +384,9 @@ class BaseQueryTestCase(DatabaseTestCase):
         res = await self.con.execute(query)
         # sort the query result by using the supplied key
         for r in res:
-            r.sort(key=key)
+            # don't bother sorting empty things
+            if r:
+                r.sort(key=key)
```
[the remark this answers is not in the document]
vpetrovykh: I ran into this issue when I was updating the tests for this feature. But in principle I can factor out this tiny change and make it into a separate commit.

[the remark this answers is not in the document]
1st1: yes, do that

edb/server/pgsql/compiler/expr.py
```diff
@@ -170,6 +175,26 @@ def compile_IndexIndirection(
...
     if isinstance(arg_type, s_scalars.ScalarType):
         b = arg_type.get_topmost_concrete_base()
+
+        # JSON accessors already handle negative offset and count from
+        # 0, so we don't need any of the additional processing of the
+        # index expression that str and arrays require.
+        if b.name == 'std::json':
+            anyint_t = ctx.env.schema.get('std::anyint')
+            # integer indexes must be cast into an int
+            if _infer_type(expr.index, ctx=ctx).issubclass(anyint_t):
```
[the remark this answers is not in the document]
vpetrovykh: OK

tests/test_edgeql_expressions.py
vpetrovykh: That's an artefact of the fact that the conversion from negative index to "real" index happens before this step. So the original index is lost by this point in code. At this point it's possibly a good idea to have a generic negative index adjusting function to be used in `array_get` and `_array_index` that would allow preserving the original index value along side with the new adjusted value. As long as this isn't a terrible anti-pattern for SQL, I'd implement that.

edb/server/pgsql/compiler/expr.py
```diff
@@ -198,20 +223,15 @@ def compile_IndexIndirection(
...
-        index = pgast.TypeCast(
-            arg=index,
-            type_name=pgast.TypeName(
-                name=('int',)
-            )
-        )
         result = pgast.FuncCall(
-            name=('substr',),
-            args=[subj, index, one]
+            name=('edgedb', '_string_index'),
```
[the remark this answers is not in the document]
vpetrovykh: I'm just following what appears to be a pre-existing convention. @elprans, is there some reason for it?

edb/server/_testbase.py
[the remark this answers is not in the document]
vpetrovykh: OK, refactored as https://github.com/edgedb/edgedb/pull/250

edb/server/pgsql/metaschema.py
vpetrovykh: `IF` only exists in the imperative PLPGSQL, which we're trying to avoid in favor of pure SQL.

edb/server/pgsql/metaschema.py
vpetrovykh: Good point.

edb/lang/schema/_std.eql
```diff
@@ -205,6 +205,26 @@ CREATE FUNCTION std::current_datetime() -> std::datetime
 CREATE FUNCTION std::uuid_generate_v1mc() -> std::uuid
     FROM SQL 'SELECT edgedb.uuid_generate_v1mc()';
 
+# JSON functions
+CREATE FUNCTION std::json_typeof($json: std::json) -> std::str
+    FROM SQL FUNCTION 'jsonb_typeof';
+
+CREATE FUNCTION std::json_array_as_array($arr: std::json) -> array<std::json>
```
elprans: Drop this in favor of upcoming support for json casting (`<array<json>>json_array_value`)

tests/test_edgeql_expressions.py
vpetrovykh: I've fixed it with a separate function that normalizes the indexes.

edb/server/pgsql/compiler/expr.py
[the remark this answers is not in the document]
vpetrovykh: I'll rename the helpers wholesale in a different PR.

tests/test_edgeql_expressions.py
vpetrovykh: There are a number of issues with indexes and slices in general that I'll be addressing in a separate PR. These issues are not JSON-specific.

edb/server/pgsql/metaschema.py
```diff
@@ -919,6 +978,221 @@ def __init__(self):
...
+            returns=('jsonb',),
+            volatility='immutable',
+            strict=True,
+            text=self.text)
+
+
+class JSONIndexByBigintFunction(dbops.Function):
+    """Get a JSON element by bigint index or raise an exception."""
+    text = r'''
+        SELECT
```
elprans: I'm thinking we shouldn't copy-paste the function body for all argument variants.  Please either make alternatives into small wrappers that call the main implementation with an up-casted argument, or cast at the call site.

vpetrovykh: In this case it saves me from checking whether JSON indexing is done by some form of integer vs. text and upcasting the integer to `bigint` in Python. But I could move that logic from implicit Postgres polymorphism into Python entirely.

PR #248, opened 2018-10-16
Implement new syntax for function parameters

1st1:
Paremeter kind and parameter type are now different things.
The following parameter kinds are supported right now:

 * regular positional parameters
 * VARIADIC
 * NAMED ONLY

The syntax for declaring functions & constraints not longer
allows to specify unnamed parameters -- names are required.

Old syntax:

    CREATE FUNCTION foo(int64)
    CREATE FUNCTION foo(VARIADIC SET OF str)
    CREATE FUNCTION foo($a: VARIADIC SET OF str)

New syntax:

    CREATE FUNCTION foo($p: int64)
    CREATE FUNCTION foo(VARIADIC $name: SET OF str)
    CREATE FUNCTION foo(VARIADIC $a: SET OF str)

Syntax for declaring named-only parameters (not yet supported by the
compiler):

    CREATE FUNCTION foo($pos1: int64, $pos2: str,
                        NAMED ONLY $n1: str, NAMED ONLY $n2: int64)

Named-only parameters will only be supported by functions (not
constraints).

### DIFF
--- a/edb/lang/edgeql/utils.py
+++ b/edb/lang/edgeql/utils.py
@@ -18,6 +18,7 @@
 
 
 import copy
+import itertools
 import typing
 
 from edb.lang.common import ast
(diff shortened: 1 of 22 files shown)


edb/lang/edgeql/utils.py
```diff
@@ -51,21 +52,27 @@ def inline_parameters(ql_expr: qlast.Base, args: typing.Dict[str, qlast.Base]):
...
-            container = []
-            result.append(qlast.Array(elements=container))
+        if varparam is not None and i == varparam:
+            result[n] = qlast.Array(elements=varargs)
+            varargs.append(e)
+        else:
+            if varparam is not None and i >= varparam:
+                varargs.append(e)
+            else:
+                result[n] = e
```
vpetrovykh: Maybe change `if ... else: if ... else` to `if ... elif ... else` to improve readability.

Issue #251, opened 2018-10-17
Improve support for JSON casts


elprans:
Currently, casting to and from `std::json` is supported only for scalar types. This proposes to extend JSON casting support to collection types as well as allowing to cast objects to `std::json`.

New rules:

1. `<json>Object` returns a `json` value containing the output representation of the object.  That is, the result is the same as `SELECT Object` in JSON mode, including the type shape.

2. Scalar casts now work only for _scalar_ JSON values, so casting a non-scalar JSON value to `std::str` is an error, and `<json>'{"foo":"bar"}'` returns a JSON *string* (`"{\"foo\":\"bar\"}"`), not a JSON object.

3. Non-scalar string-to-JSON and JSON-to-string casting is accomplished by the new `str_to_json(str) -> json` and `json_to_str(json) -> str` functions correspondingly.

4. `<json>array_value` returns a JSON array 

5. `<array<type>>json_value` returns an EdgeQL array of the given type.  `null` values in input produce an error (if the target type is not `json`, use `json_array_unpack` for fine-grained casting).

5. `<json>tuple_value` returns a JSON array and conversely `<tuple<...>>json_value` converts a JSON array into a tuple. `null` elements in input produce an error (if the target element type is not `json`).

6. `<json>named_tuple_value` returns a JSON object and conversely `<tuple<name: type...>>json_value` converts a JSON object/array into a named tuple. `null` values in input produce an error (if the target type is not `json`).

vpetrovykh:
Currently the casts from JSON to tuples are broken. The JSON docs will need to be updated once this is fixed.

haikyuu:
I am getting an error while trying to cast a string to an enum

```
error {
message: "cannot cast 'std::json' to 'default::Position'",
type: 'QueryError',
code: 67108864
 }
```
Query

```
INSERT Job {
            position := <Position>$position
          }
```
And i'm passing a string as a variable.

Is this related? or should strings be considered as str instead of `json`?

1st1:
@vpetrovykh Please check if str->enum casts work in EdgeQL+HTTP / GraphQL contexts correctly.

1st1:
@elprans @vpetrovykh Should this be closed now?

elprans:
I think we need to allow casting `json` to `anyenum` directly.  @vpetrovykh please look into this.

elprans:
Remaining work items in this issue:

[...]

PR #259, opened 2018-10-18
Add support for casting objects and tuples to std::json

elprans:
`<json>Object` now always returns a `std::json` value containing
the output representation of the object. That is, the result is
the same as the output of `SELECT Object` in JSON mode, including
the type shape.

Now, casting a non-named tuple to `std::json` returns a JSON array,
and casting a named tuple to `std::json` returns a JSON object.

Casting non-scalar JSON values to scalars is now an error, as well as
casting a non-matching JSON scalar value.

Issue: #251.

### DIFF
--- a/edb/lang/edgeql/compiler/expr.py
+++ b/edb/lang/edgeql/compiler/expr.py
@@ -44,6 +44,7 @@ from . import pathctx
 from . import setgen
 from . import schemactx
 from . import typegen
+from . import viewgen
 
 from . import func  # NOQA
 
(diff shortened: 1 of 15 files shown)


vpetrovykh:
> Casting non-scalar JSON values to scalars is not an error, as well as
> casting a non-matching JSON scalar value.

I assume you mean "...is now an error...".

elprans:
> I assume you mean "...is now an error...".
Yes.  Thanks.

PR #262, opened 2018-10-20
edgeql: Add more implicit casting tests.

vpetrovykh:
I've added [mostly] failing implicit cast tests.

I've also corrected the existing test in accordance with what the `_implicit_numeric_cast_map` looks like to me.


tests/test_edgeql_expressions.py
```diff
@@ -615,11 +616,94 @@ class TestExpressions(tb.QueryTestCase):
...
+        # IF should also force implicit casts of the two options
+        await self.assert_query_result(r"""
+            SELECT 3 / (2 IF TRUE ELSE 2.0);
+            SELECT 3 / (2 IF random() > -1 ELSE 2.0);
+
+            SELECT 3 / (2 IF FALSE ELSE 2.0);
+            SELECT 3 / (2 IF random() < -1 ELSE 2.0);
+        """, [
+            [1],
+            [1],
```
1st1: It's going to be 1.5 here and on the prev line.

1st1: We determine the type statically, not dynamically.

vpetrovykh: Yeah, let me amend that.

PR #264, opened 2018-10-22
Add support for `on target delete` clause for links

elprans:
The new `on target delete` clause of concrete links is a mechanism
to specify what happens when a target object of a link instance
is deleted.

Currently implemented modes are:

* `RESTRICT` -- produce an error indicating that deletion cannot be
  done due to an existing link. This is the default.
* `DEFERRED RESTRICT` -- same as the above, but the check is postponed
  until the end of the current transaction, if there is no transaction,
  equivalent to `RESTRICT`.
* `SET EMPTY` -- remove the link, can only be used with links that are
  not required.
* `DELETE SOURCE` -- remove the link and the source of the link
  (cascaded deletion).

### DIFF
--- a/edb/lang/schema/constraints.py
+++ b/edb/lang/schema/constraints.py
@@ -32,19 +32,24 @@ from . import name as sn
 from . import named
 from . import objects as so
 from . import referencing
+from . import utils
 
 
 class CumulativeBoolExpr(s_expr.ExpressionText):
     @classmethod
-    def merge_values(cls, ours, theirs, schema):
-        if ours and theirs and ours != theirs:
-            result = '({}) and ({})'.format(ours, theirs)
-        elif not ours and theirs:
-            result = theirs
-        else:
-            result = ours
+    def merge_values(cls, target, sources, field_name, *, schema):
+        def join(values):
+            # Make the list unique without losing the order.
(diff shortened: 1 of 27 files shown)


edb/lang/schema/constraints.py
```diff
@@ -19,6 +19,7 @@
 
 import itertools
 
+from edb.lang.common.ordered import OrderedSet
```
1st1: Module import?

tests/schemas/link_tgt_del.eschema
```diff
@@ -0,0 +1,44 @@
+#
+# This source file is part of the EdgeDB open source project.
+#
+# Copyright 2008-2016 MagicStack Inc. and the EdgeDB authors.
```
vpetrovykh: Why "2008-2016"? Looks like a bad copy-paste.

tests/test_link_target_delete.py
```diff
@@ -0,0 +1,277 @@
+#
+# This source file is part of the EdgeDB open source project.
+#
+# Copyright 2016-present MagicStack Inc. and the EdgeDB authors.
```
vpetrovykh: Again, looks like bad-copy-paste: "2016-present"

Issue #265, opened 2018-10-22
Change precedence rules for views and casts.


vpetrovykh:
Currently the following results in an error:
```
SELECT <json>Foo { bar };
```
Currently, the only valid cast of an object + shape is `json`. So it's basically the only use-case that we have for this combination of syntax.

Making casts lower priority may lead to confusion of what you're casting in complex expressions. So the better solution is to make view higher priority than cast. 

Moving the view operator to a higher precedence than cast would put it into the same category as `[]` and `foo()` (the precedence of `(` between the function name and its arguments). There's no ambiguity there since they are all kind of postfix operators and their actual sequence determines the sequence of application.

In addition the following will change:
```
SELECT EXISTS Foo { bar };
# this will *no longer be an error*, but will be equivalent to
SELECT EXISTS Foo;
```

```
SELECT -1{__type__: {name}};
# this will *become an error* now, the correct usage would be:
SELECT (-1){__type__: {name}};
```

A more subtle effect will be for `DETACHED`:
```
WITH Bar := (SELECT Bar FILTER .name = 'qwerty')
SELECT DETACHED Foo {
   bar,
   baz := Bar
};
```
The above will detach *both* `Foo` and `Bar`, effectively ignoring the `Bar` defined in the WITH block. Under the current precedence rules only `Foo` would be detached.

PR #267, opened 2018-10-23
Precedence

vpetrovykh:
Update precedence, add some tests that rely on precedence.

Issue #265

### DIFF
--- a/tests/test_session.py
+++ b/tests/test_session.py
@@ -117,3 +117,6 @@ class TestSession(tb.QueryTestCase):
 
             [['entity', 'user']]
         ])
+
+    # TODO: see if nested WITH block with different modules can be
+    # used to screw things up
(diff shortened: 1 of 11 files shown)


tests/test_session.py
```diff
@@ -117,3 +117,6 @@ class TestSession(tb.QueryTestCase):
 
             [['entity', 'user']]
         ])
+
+    # TODO: see if nested WITH block with different modules can be
```
elprans: Is this TODO still relevant?

vpetrovykh: Absolutely, I'll add the tests that I have shown to you in a separate small PR. Maybe I'll manage to come up with more obscure ways of nesting default modules, I'll need to look at the scopes tests for additional inspiration.

PR #270, opened 2018-10-24
Bytes

1st1:
(see individual commits)

### DIFF
--- a/edb/lang/edgeql/parser/grammar/lexer.py
+++ b/edb/lang/edgeql/parser/grammar/lexer.py
@@ -43,7 +43,7 @@ class EdgeQLLexer(lexer.Lexer):
     MERGE_TOKENS = {('NAMED', 'ONLY')}
 
     NL = 'NL'
-    MULTILINE_TOKENS = frozenset(('SCONST',))
+    MULTILINE_TOKENS = frozenset(('SCONST', 'BCONST'))
     RE_FLAGS = re.X | re.M | re.I
 
     # Basic keywords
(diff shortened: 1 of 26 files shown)


edb/lang/edgeql/parser/grammar/lexer.py
```diff
@@ -115,6 +115,26 @@ class EdgeQLLexer(lexer.Lexer):
...
+                )
+                (?P<BQ>
+                    (
+                        ' | "
+                    )
+                )
+                (?:
+                    (
+                        (\\\\ | \\['"] | \n | .)*?
+                        # we'll validate escapes codes in the parser
```
elprans: *escape*

edb/server/pgsql/metaschema.py
```diff
@@ -1145,6 +1145,31 @@ def __init__(self):
...
+class BytesIndexWithBoundsFunction(dbops.Function):
+    """Get a bytes character or raise an out-of-bounds exception."""
+    text = '''
+        SELECT edgedb._raise_exception_on_empty(
+            substr(
+                val,
+                edgedb._normalize_array_index(index, length(val)),
+                1),
+            'invalid_parameter_value',
+            'bytes index ' || index::text || ' is out of bounds',
```
elprans: I'd use the term "byte string" instead of "bytes"

tests/test_edgeql_expressions.py
```diff
@@ -1077,7 +1077,7 @@ class TestExpressions(tb.QueryTestCase):
         with self.assertRaisesRegex(
                 # FIXME: possibly a different error should be used here
                 exc.UnknownEdgeDBError,
-                r'string index 10 is out of bounds'):
+                r'str index 10 is out of bounds'):
```
elprans: I don't think "str" is better than "string" here.

PR #274, opened 2018-10-25
Make regular division return floats and add a floor division operator

elprans:
'/' now always performs a floating-point division.  The new '//'
operator peforms floor division explicitly.

The motivation behind this change is to allow greater accuracy in
complex expressions where the type of the numeric arguments is not
necessarily explicitly known, such as in polymorphic functions.


edb/lang/schema/scalars.py
```diff
@@ -228,13 +228,21 @@ def _is_assignment_castable_impl(source: str, target: str) -> bool:
...
+        ('std::int64', 'std::int64', 'std::float64'),
         ('std::float32', 'std::float32', 'std::float32'),
         ('std::float64', 'std::float64', 'std::float64'),
         ('std::decimal', 'std::decimal', 'std::decimal'),
         ('std::decimal', 'std::int64', 'std::decimal'),
     ],
 
+    ast.ops.FLOORDIV: [
+        ('std::int64', 'std::int64', 'std::int64'),
+        ('std::float32', 'std::float32', 'std::int64'),
```
1st1: We want the result of `<float32>1 // <float32>2 + <float32>1` to be a `float32`.

elprans:
@1st1 @vpetrovykh  I rebased this over the float literal change, so the changes are far less controversial this time.  Please review.

PR #277, opened 2018-10-25
Add a number of error checks before adding a function to the schema

1st1:
The following scenarios cause an error now:

* Different names for NAMED ONLY parameteres in overloaded functions

* Creating a SET OF EdgeQL function (not supported yet)

* Overloading a polymorphic function with a non-polymorphic function
  of a different return type

* Overloading a non-polymorphic function with a polymorphic function
  of a different return type

* Overloading a polymorphic function with a polymorphic function
  of a different return type


edb/lang/schema/functions.py
```diff
@@ -285,17 +323,52 @@ def _add_to_schema(self, schema):
...
+                    f'function: overloading another function with different '
+                    f'return type {func.return_typemod.to_edgeql()}'
+                    f'{func.return_type.name}',
+                    context=self.source_context)
+
+        if (language == qlast.Language.EdgeQL and
+                any(p.typemod is ft.TypeModifier.SET_OF for p in params)):
+            raise ql_errors.EdgeQLError(
+                f'cannot create {get_signature()} function: '
+                f'SET OF parameters in EdgeQL functions are '
```
elprans: *user-defined* EdgeQL functions

PR #279, opened 2018-10-26
Stop interpreting literals into Python values in constants

elprans:
Currently, the "value" field in various `Constant` AST nodes is a Python
value, which is obtained by interpreting the corresponding EdgeQL
literal.  When rendered back into code, the Python value is then
converted back into an EdgeQL or SQL value.  This approach is inconvenient
and is potentially lossy, so stop doing it and store the actual literal
value as a string, converting to Python only when necessary for
interpretation (such as constant folding).

### DIFF
--- a/edb/lang/edgeql/ast.py
+++ b/edb/lang/edgeql/ast.py
@@ -232,27 +232,72 @@ class FunctionCall(Expr):
     agg_set_modifier: typing.Optional[SetModifier]
 
 
-class Constant(Expr):
-    value: typing.Union[int, str, float, bool, bytes, decimal.Decimal]
+class BaseConstant(Expr):
+    value: str
+
+    @classmethod
+    def from_python(cls, val: object) -> 'BaseConstant':
+        if isinstance(val, str):
+            return StringConstant.from_python(val)
+        elif isinstance(val, bool):
+            return BooleanConstant(value='true' if val else 'false')
+        elif isinstance(val, int):
+            return IntegerConstant(value=str(val))
+        elif isinstance(val, decimal.Decimal):
+            if val.to_integral_value() == val:
+                return IntegerConstant(value=str(val))
+            else:
+                return FloatConstant(value=str(val))
+        elif isinstance(val, float):
+            return FloatConstant(value=str(val))
+        elif isinstance(val, bytes):
+            return BytesConstant.from_python(value=val)
+        else:
+            raise ValueError(f'unexpected constant type: {type(val)!r}')
 
 
-class StringConstant(Constant):
+class StringConstant(BaseConstant):
     quote: str
 
     @classmethod
-    def from_pystr(cls, s: str):
+    def from_python(cls, s: str):
         s = s.replace('\\', '\\\\')
         value = quote.quote_literal(s)
         return cls(value=value[1:-1], quote="'")
 
 
-class RawStringConstant(Constant):
+class RawStringConstant(BaseConstant):
     quote: str
 
     @classmethod
-    def from_pystr(cls, s: str):
+    def from_python(cls, s: str):
         value = quote.quote_literal(s)
         return cls(value=value[1:-1], quote="'")
+
+
+class BaseRealConstant(BaseConstant):
+    is_negative: bool = False
+
+
+class IntegerConstant(BaseRealConstant):
+    pass
+
+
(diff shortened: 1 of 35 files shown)


edb/lang/edgeql/ast.py
```diff
@@ -232,29 +232,74 @@ class FunctionCall(Expr):
...
+            return BooleanConstant(value='true' if val else 'false')
+        elif isinstance(val, int):
+            return IntegerConstant(value=str(val))
+        elif isinstance(val, decimal.Decimal):
+            if val.to_integral_value() == val:
+                return IntegerConstant(value=str(val))
+            else:
+                return FloatConstant(value=str(val))
+        elif isinstance(val, float):
+            return FloatConstant(value=str(val))
```
1st1: Should we use `val.hex()` for floats?

edb/lang/ir/ast.py
```diff
@@ -135,25 +135,43 @@ class EmptySet(Set):
...
     pass
 
 
-class RawStringConstant(Constant):
+class RawStringConstant(BaseConstant):
     pass
 
 
+class IntegerConstant(BaseConstant):
+    value: str
```
1st1: `value` is already specified in `BaseConstant`

edb/lang/ir/ast.py
```diff
@@ -135,25 +135,43 @@ class EmptySet(Set):
...
+
+
+class FloatConstant(BaseConstant):
+    value: str
+
+
+class BooleanConstant(BaseConstant):
+    value: bool
```
1st1: Sure you want `BooleanConstant` to be an exception and have a non-string `value`?

edb/lang/edgeql/ast.py
elprans: IIRC, `from_python(<float>)` is only used in the GraphQL translator. I'll remove support for float coercion altogether once the GraphQL AST gets the same treatment as here.

edb/lang/ir/staeval.py
```diff
@@ -0,0 +1,263 @@
...
+        unop: irast.UnaryOp,
+        schema: s_schema.Schema) -> irast.BaseConstant:
+
+    real_t = schema.get('std::anyreal')
+    int_t = schema.get('std::anyint')
+
+    result_type = unop.expr.scls
+
+    if (result_type.issubclass(real_t) and
+            unop.op in (ast.ops.UMINUS, ast.ops.UPLUS)):
```
1st1: Can't you simply do nothing if it's a unary plus?

edb/lang/ir/ast.py
1st1: Moreover, it seems that you always pass a string to it, not a bool.

edb/lang/ir/staeval.py
```diff
@@ -0,0 +1,263 @@
...
+        return float(ir.value)
+
+
+@const_to_python.register(irast.StringConstant)
+def str_const_to_python(
+        ir: irast.StringConstant,
+        schema: s_schema.Schema) -> str:
+
+    rv = qlquote.quote_literal(ir.value)
+    return pyast.literal_eval(f"''{rv}''")
```
1st1: I'd use something like this instead of literal eval:

```python
import ast
import re


def unescape(st):

    str_re = re.compile(r'''
        (?P<slash> \\\\) |
        (?P<dq> \\") |
        (?P<sq> \\') |
        (?:\\x(?P<x>[0-7][0-9a-fA-F])) |
        (?:\\u(?P<u>[0-9a-fA-F]{4})) |
        (?:\\U(?P<U>[0-9a-fA-F]{8})) |
        (?P<n>\\n) |
        (?P<t>\\t) |
        (?P<r>\\r)
    ''', re.X)

    subs = {
        'slash': '\\',
        'dq': '"',
        'sq': "'",
        't': '\t',
        'n': '\n',
        'r': '\r',
    }

    def cb(m):
        g = m.lastgroup
        try:
            return subs[g]
        except KeyError:
            pass

        return chr(int(m.group(g), 16))

    return str_re.sub(cb, st)


x = r'aaaa a \u0062 \U00001234 >\ < \\ \' \" \n \t ! \r asd \x62'
print(x)
print('UNESC', unescape(x))
print('LITEV', ast.literal_eval(f'"""{x}"""'))

```

PR #284, opened 2018-10-28
Make "any" a keyword; drop std::any; add std::anyscalar.

1st1:
* std::any is no longer a thing.  "any" is now a keyword and should be
  used without a "std::" prefix.

* The new "std::anyscalar" is the base type for all scalars, including
  abstract base types like "std::anyint".

* "std::len(array)" was fixed to return non-empty set on empty arrays.


edb/server/pgsql/delta.py
```diff
@@ -458,6 +458,7 @@ def apply(self, schema, context):
                 context=self.source_context)
 
         op = dbops.CreateFunction(dbf)
+        # print(op.code(None))
```
elprans: Remove this?

PR #286, opened 2018-10-29
edb.test: Make it possible to specify the output format explicitly

elprans:
The new `--output-format` option allows choosing from the following:

* 'auto' -- auto-detect the output format (current behavior, default)
* 'simple' -- simple terse output
* 'stacked' -- group output by the test suite
* 'verbose' -- echo each test result on a separate line.

### DIFF
--- a/edb/tools/test/__init__.py
+++ b/edb/tools/test/__init__.py
@@ -37,6 +37,10 @@ from . import styles
               help='increase verbosity')
 @click.option('-q', '--quiet', is_flag=True,
               help='decrease verbosity')
+@click.option('--output-format',
+              type=click.Choice(runner.OutputFormat.__members__),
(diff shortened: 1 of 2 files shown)


1st1:
We already have the `-v` flag that clashes (or just the same?) with the proposed `--output-format=verbose`.

elprans:
You will have extra verbosity around test collection and bootstrap, test progress will use the `MultilineRenderer`.

1st1:
But MultilineRenderer can break with extra verbosity, it relies on tests being run quiet.

1st1:
Just raise an error if verbose mode is on and MultilineRenderer is selected.

elprans:
Alright, I added a check.

PR #292, opened 2018-10-30
Tests

vpetrovykh:
Added more tests and fixed cardinality inference for many cases (function calls, index dereferencing, slices, etc.)

### DIFF
--- a/tests/test_edgeql_json.py
+++ b/tests/test_edgeql_json.py
@@ -65,7 +65,7 @@ class TestEdgeQLJSON(tb.QueryTestCase):
 
             [True],
             [False],
-            [None],
+            [None],  # JSON null
 
             [[2, 'a', 3.456]],
             [[2, 'a', 3.456, [['b', 1]]]],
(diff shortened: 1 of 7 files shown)


tests/test_edgeql_json.py
```diff
@@ -1024,3 +1024,139 @@ class TestEdgeQLJSON(tb.QueryTestCase):
...
+                {'2', '4', '4'}, {'0', 'b'}, {'bar', 'foo'}, {'1', '2'},
+                default := <json>'N/A'
+            )
+            ORDER BY _;
+        ''', [
+            # We expect 3 * 2 * 2 * 2 = 24 results per JSONTest.data.
+            # There are 3 non-empty data properties.
+            # Out of them all only 4 will have a valid path.
+            # So we expect 24 * 3 - 4 = 68 default 'N/A' results
+            [None, None] + ['N/A'] * 68 + [{'bingo': '42!'}, {'bingo': '42!'}],
```
elprans: Worth mentioning that these `None` values here are actually the interpretation of JSON `null` values, and not an (invalid) leakage of SQL NULL.

vpetrovykh: OK, I've updated comments in JSON tests and annotated occurrences of `None` that come from a legitimate `null`.

1st1:
(merged by hand)

PR #298, opened 2018-11-01
Implement OPERATOR DDL

elprans:
The new `{CREATE|UPDATE|DROP} OPERATOR` DDL commands formalize the
process of adding an operator to the schema.  They use the same
parameter and return declaration syntax as functions, although
named-only parameters are obviously not allowed.

The operator DDL is restricted to the `std` module and its only
currently implemented functionality is to alias PostgreSQL
operators to EdgeQL operators.

This approach allows systematic and exhaiustive definition of the
allowed operator space, as well as making it possible to generate proper
polymorphic SQL.

This PR also includes reorganization of the standard library sources to 
help with the coming addition of explicit operator definitions.


tests/test_link_target_delete.py
```diff
@@ -387,9 +387,10 @@ class TestLinkTargetDeleteDeclarative(stb.QueryTestCase):
 
             await self.assert_query_result(r'''
```
1st1: Typo in the commit message:

```
Qualify all queries in target delet tests with correct filters
                                   ^
```

Issue #70, opened 2017-07-21
Strings in EdgeQL and eschema.


vpetrovykh:
Currently EdgeQL strings are very simple and don't even have line continuation. The original rationale was that explicit string concatenation solves the problem without inventing new syntax. It also allows us to treat `$$`-strings as essentially raw strings, where everything is exactly as written, including \-escapes.

Line continuation is somewhat desirable for the cases when logically there should be no newlines, but the amount of text is large (e.g. descriptions).

Schema aggravates this problem by being white-space sensitive. Currently schema parser will automagically dedent raw strings *and* handle line continuation symbol before handing the strings off to the EdgeQL parser. The case for needing nice line continuation in the Schema is stronger.

```yml
concept Foo:
    description :=
        "This is all one" +
        "string without newlines."
```
v.s. something like
```yml
concept Foo:
    description :=
        "This is all one \
        string without newlines."
        # dedent will get rid of the whitespace before the word "string"
```

There's a `:>` operator in the schema that attempts to streamline strings, but because it lacks a terminal symbol it is difficult to highlight the string after it with regular expressions. This operator needs to be removed.

An option to consider is adding additional string types for `$$`-string. From lower ascii set only alphabetic characters, digits and "_" are allowed in the `$foo_1$` marker. This can be used to add a special separator for different string types. We seem to need strings that automatically dedent themselves, so examples of syntax for that could be:
- `$d:$ ...content... $$` 
- `$d:foo_1$ ...content... $foo_1$`
- `$:<$ ...content... $$` 
- `$foo_1:<$ ...content... $foo_1$`

TODO summary:
- [x] remove `:>`
- [x] remove very exceptional line continuation rule from eschema
- [x] add more string literal types to EdgeQL
- [x] add line continuation to some/all string literals in EdgeQL

PR #299, opened 2018-11-01
eschema: Remove `:>` syntax.

vpetrovykh:
The `:>` syntax is too obscure. Using `:=` with an explicit string is
better.


edb/lang/edgeql/parser/grammar/lexer.py
```diff
@@ -174,6 +174,14 @@ class EdgeQLLexer(lexer.Lexer):
                 (?P=Q)
              '''),
 
+        Rule(token='BADSCONST',
```
1st1: Please add a comment explaining that this will match any string-like thing that wasn't matched by string regexps above

vpetrovykh:
Related to #70

PR #308, opened 2018-11-08
edgeql: Add a `str` function.

vpetrovykh:
The `str` function is meant to be like a parametrized cast. When called
without the format parameter it is equivalent to casting via `<str>`.

### DIFF
--- a/tests/test_edgeql_functions.py
+++ b/tests/test_edgeql_functions.py
@@ -689,3 +689,176 @@ class TestEdgeQLFunctions(tb.QueryTestCase):
         ''', [
             [6.8],
         ])
+
+    @unittest.expectedFailure
+    async def test_edgeql_functions_str_01(self):
+        # at the very least the cast <str> should be equivalent to
+        # a call to str() without explicit format for simple scalars
+        await self.assert_query_result(r'''
+            WITH DT := current_datetime()
+            # FIXME: the cast has a "T" and the str doesn't for some reason
+            SELECT <str>DT = str(DT);
+
+            WITH D := current_date()
+            SELECT <str>D = str(D);
+
+            SELECT <str>123 = str(123);
+            SELECT <str>123.456 = str(123.456);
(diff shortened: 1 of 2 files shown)


tests/test_edgeql_functions.py
```diff
@@ -689,3 +689,176 @@ class TestEdgeQLFunctions(tb.QueryTestCase):
...
             [6.8],
         ])
+
+    @unittest.expectedFailure
+    async def test_edgeql_functions_str_01(self):
+        # at the very least the cast <str> should be equivalent to
+        # a call to str() without explicit format for simple scalars
+        await self.assert_query_result(r'''
+            WITH DT := current_datetime()
+            # FIXME: the cast has a "T" and the str doesn't for some reason
```
elprans: This is because we actually use a normalization hack in `<str>datetime`:

https://github.com/edgedb/edgedb/blob/6da6f58c933deaabb520053d2672101170eb6ee2/edb/server/pgsql/compiler/typecomp.py#L129-L145

tests/test_edgeql_functions.py
```diff
@@ -689,3 +689,176 @@ class TestEdgeQLFunctions(tb.QueryTestCase):
...
+            {},
+        ])
+
+    @unittest.expectedFailure
+    async def test_edgeql_functions_str_03(self):
+        await self.assert_query_result(r'''
+            WITH DT := <datetime>'2018-05-07 15:01:22.306916-05'
+            SELECT str(DT, 'HH:MI A.M.');
+        ''', [
+            # there's some implicit conversion happening, is that
```
elprans: Postgres converts `timestamp with timezone` into the local timezone of the server.  We run tests with the UTC timezone.  Start the server with `--timezone=UTC` (I think we should do this by default).

vpetrovykh:
I've made a more general PR #317 to replace this one.

PR #309, opened 2018-11-08
Update how time and date is handled.

vpetrovykh:
`std::time` is now `std::naivetime` and is no longer time zone aware.

The function `current_time` is now `local_time`.


edb/lib/std/30-datetimefuncs.eql
```diff
@@ -28,9 +28,9 @@ std::current_date() -> std::date
 
 
 CREATE FUNCTION
-std::current_time() -> std::time
+std::local_time() -> std::naivetime
```
1st1: I propose to change the order to `time_local()`; I'd like all time-related functions to be grouped together.

vpetrovykh: We have:
- `current_date`
- `current_datetime`
- `local_time`

Flipping the order seems a bit counter-intuitive.

elprans: I vote for `current_localtime()`

vpetrovykh:
I've pushed a revised version of the update with new time and date functions.

edb/lang/schema/scalars.py
```diff
@@ -232,7 +232,9 @@ def _is_assignment_castable_impl(source: str, target: str) -> bool:
         ('std::str', 'std::str', 'std::str'),
         ('std::bytes', 'std::bytes', 'std::bytes'),
         ('std::datetime', 'std::timedelta', 'std::datetime'),
-        ('std::time', 'std::timedelta', 'std::time'),
+        ('std::naivedatetime', 'std::timedelta', 'std::naivedatetime'),
```
1st1: `std::naivedatetime` -> `std::naive_datetime`? @elprans

edb/lib/std/30-datetimefuncs.eql
```diff
@@ -21,21 +21,62 @@
...
+std::datetime_now(mode: std::str='') -> std::datetime
 {
-    FROM SQL 'SELECT current_date';
+    FROM SQL $$
+    SELECT (
+        CASE WHEN "mode" = 'transaction' THEN
+            transaction_timestamp()
+        WHEN "mode" = 'statement' THEN
+            statement_timestamp()
+        ELSE
```
1st1: I don't like that this function can accept any garbage.  I also think that having more than 2 different modes of operation is way too much for a single function. Let's discuss splitting this into three functions.

elprans: `std::datetime_of_transaction`, `std::datetime_of_statement`, `std::datetime_now` seem like reasonable names to me.

vpetrovykh: I've updated the PR.

edb/lib/std/30-datetimefuncs.eql
```diff
@@ -21,21 +21,66 @@
 
 
 CREATE FUNCTION
-std::current_date() -> std::date
+std::datetime_now() -> std::datetime
```
elprans: Perhaps, `datetime_current` would be a better name, making some distance from SQL `now()` which is a transaction timestamp.

vpetrovykh: Sure, let's call it `datetime_current`. Avoiding confusion with SQL `now()` is important here.

PR #313, opened 2018-11-14
schema: Make Object.id a required field and always populate it

elprans:
Every schema item now has a unique id assigned at creation.

### DIFF
--- a/edb/lang/common/uuidgen.py
+++ b/edb/lang/common/uuidgen.py
@@ -17,17 +17,11 @@
 #
 
 
-import asyncpg
-import typing
+import os
+import uuid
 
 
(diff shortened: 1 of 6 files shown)


edb/lang/common/uuidgen.py
```diff
@@ -0,0 +1,27 @@
...
+#
+
+
+import random
+import uuid
+
+
+def uuid1mc() -> uuid.UUID:
+    """Generate a v1 UUID using a pseudo-random multicast node address."""
+    node = random.getrandbits(48) | (1 << 40)
```
1st1: Let's use `os.urandom(6)`

PR #317, opened 2018-11-14
edgeql: Add constructor functions for multiple data types.

vpetrovykh:
The `str` function is meant to be like a parametrized cast. When called
without the format parameter it is equivalent to casting via `<str>`.

`str` usage with `json` scalars is not equivalent to casting, it is
producing a string representation of JSON. The inverse function `json`
is therefore parsing a string, not casting either:
- `str_to_json` becomes `json`
- `json_to_str` becomes `str`

Add various date and time constructor functions.

Add `bool` function that converts values of other types to a boolean based
on how "truthy" the values are. E.g. `0`, `''` and `[]` are converted to
`False` and `23`, `'q'`, and `[1, 2]` are converted to `True`.

### DIFF
--- a/edb/lang/graphql/translator.py
+++ b/edb/lang/graphql/translator.py
@@ -118,7 +118,7 @@ class GraphQLTranslator(ast.NodeVisitor):
         for el in eql[0].result.elements:
             # swap in the json bits
             if (isinstance(el.compexpr, qlast.FunctionCall) and
-                    el.compexpr.func == 'str_to_json'):
+                    el.compexpr.func == 'to_json'):
                 name = el.expr.steps[0].ptr.name
                 el.compexpr.args[0].arg = qlast.StringConstant.from_python(
                     json.dumps(gqlresult.data[name], indent=4))
--- a/edb/lang/graphql/types.py
+++ b/edb/lang/graphql/types.py
@@ -572,7 +572,7 @@ class GQLBaseType(metaclass=GQLTypeMeta):
         '''
 
         if self.dummy:
-            return parse_fragment(f'''str_to_json("xxx")'''), None, None
+            return parse_fragment(f'''to_json("xxx")'''), None, None
 
         eql = parse_fragment(f'''
             SELECT {self.edb_base_name} {{
--- a/edb/lib/std/20-genericfuncs.eql
+++ b/edb/lib/std/20-genericfuncs.eql
@@ -44,7 +44,7 @@ CREATE FUNCTION
 std::len(array: array<anytype>) -> std::int64
 {
     FROM SQL $$
-    SELECT COALESCE(array_length("array", 1), 0)::bigint
+    SELECT cardinality("array")::bigint
     $$;
 };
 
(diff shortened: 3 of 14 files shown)


edb/lib/std/70-constructorfuncs.eql
```diff
@@ -0,0 +1,512 @@
...
+            to_timestamp(s, fmt)
+        END
+    )
+    $$;
+};
+
+
+CREATE FUNCTION
+std::datetime(year: std::int64, month: std::int64, day: std::int64,
+              hour: std::int64, min: std::int64, sec: std::float64)
```
elprans: We should consider making the time part optional, i.e default to `0`.

edb/lib/std/70-constructorfuncs.eql
```diff
@@ -0,0 +1,527 @@
...
+        'null'::jsonb, '""'::jsonb, '[]'::jsonb, '{}'::jsonb, 'false'::jsonb,
+        '0'::jsonb
+    )
+    $$;
+};
+
+
+# Any scalar that's not explicitly covered is "true" by virtue of existing.
+CREATE FUNCTION
+std::bool(val: anyscalar) -> std::bool
```
elprans: I don't think a catch-all converter is a good idea.  If there is no explicit conversion, it is undefined.

edb/lib/std/70-constructorfuncs.eql
```diff
@@ -0,0 +1,527 @@
...
+{
+    FROM SQL $$
+    SELECT TRUE
+    $$;
+};
+
+
+# Any Object is "true" by virtue of existing.
+CREATE FUNCTION
+std::bool(val: Object) -> std::bool
```
elprans: I'm not sure this is useful.

edb/lib/std/70-constructorfuncs.eql
elprans: > that's why we're introducing the bool constructor to get the truthy values of many more things.

Not really. The point of the constructors is to allow parametrization, not necessarily to extend the domain of inputs.

vpetrovykh: > Not really. The point of the constructors is to allow parametrization, not necessarily to extend the domain of inputs.

Contructors in general, sure. However, it appears that `bool` in particular is mainly serving the purpose of allowing a larger range of types to be treated as familiar truthy values similar to other languages. Unfortunately, there's no consistent agreement on truthiness among other languages (say Python, JS, and C) even though there are some common practices (such as treating numerical 0 as equivalent to "false" and non-zero as "true").

I think it's convenient to extend `bool` to every type because then you could ask "is a particular arbitrary set truthy". This can come up as a potentially useful `FILTER`.

```
SELECT Foo
FILTER Foo.bar != '' AND Foo.baz > 0;

# vs

SELECT Foo
FILTER bool(Foo.bar) AND bool(Foo.baz);

# or even

SELECT Foo
FILTER all({bool(Foo.bar), bool(Foo.baz)});
```
Imagine more than 2 links/props that are used in the same fashion in the above example and the general application of `bool` conveys the intent more clearly than explicit listing of various possibilities, I think.

vpetrovykh: After a discussion with @elprans and @1st1 we came to the conclusion that `bool` as truthiness is not as good a feature as it first appeared.

PR #327, opened 2018-11-21
Make Object.name a SchemaField

1st1:
(Please review individual commits)

### DIFF
--- a/edb/server/pgsql/delta.py
+++ b/edb/server/pgsql/delta.py
@@ -368,10 +368,11 @@ class FunctionCommand:
     def get_table(self, schema):
         return self._table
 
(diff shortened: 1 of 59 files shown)


edb/server/pgsql/delta.py
```diff
@@ -1653,7 +1664,8 @@ def apply(self, schema, context=None):
         self.table_name = common.objtype_name_to_table_name(
             self.new_name, catenate=False)
 
-        objtype.original_class.name = scls.name
+        schema = objtype.original_class.set_field_value(
```
elprans: Have you tried removing this outright?  Might be here for historical reasons.

PR #329, opened 2018-11-22
edgeql: Add more library functions.

vpetrovykh:
Add more generic functions to `std` module like min, max, round, etc.

Add some specialized functions to `math` module.

### DIFF
--- a/edb/lib/math.eql
+++ b/edb/lib/math.eql
@@ -1,0 +1,236 @@
+#
+# This source file is part of the EdgeDB open source project.
+#
+# Copyright 2018-present MagicStack Inc. and the EdgeDB authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+CREATE MODULE math;
+
+
+CREATE FUNCTION
+math::abs(x: std::anyreal) -> std::anyreal
+{
+    FROM SQL FUNCTION 'abs';
+};
+
+
+CREATE FUNCTION
+math::ceil(x: std::int64) -> std::float64
+{
+    FROM SQL FUNCTION 'ceil';
+};
+
+
+CREATE FUNCTION
+math::ceil(x: std::float64) -> std::float64
+{
+    FROM SQL FUNCTION 'ceil';
+};
+
+
+CREATE FUNCTION
+math::ceil(x: std::decimal) -> std::decimal
+{
+    FROM SQL FUNCTION 'ceil';
+};
+
+
+CREATE FUNCTION
+math::floor(x: std::int64) -> std::float64
+{
+    FROM SQL FUNCTION 'floor';
+};
+
+
+CREATE FUNCTION
+math::floor(x: std::float64) -> std::float64
+{
+    FROM SQL FUNCTION 'floor';
+};
+
+
+CREATE FUNCTION
+math::floor(x: std::decimal) -> std::decimal
+{
+    FROM SQL FUNCTION 'floor';
+};
+
+
+CREATE FUNCTION
+math::ln(x: std::int64) -> std::float64
+{
+    FROM SQL FUNCTION 'ln';
+};
+
+
+CREATE FUNCTION
+math::ln(x: std::float64) -> std::float64
+{
+    FROM SQL FUNCTION 'ln';
+};
+
+
+CREATE FUNCTION
+math::ln(x: std::decimal) -> std::decimal
+{
+    FROM SQL FUNCTION 'ln';
+};
+
+
+CREATE FUNCTION
+math::lg(x: std::int64) -> std::float64
+{
+    FROM SQL FUNCTION 'log';
+};
+
+
+CREATE FUNCTION
+math::lg(x: std::float64) -> std::float64
+{
+    FROM SQL FUNCTION 'log';
+};
+
+
+CREATE FUNCTION
+math::lg(x: std::decimal) -> std::decimal
+{
+    FROM SQL FUNCTION 'log';
+};
+
+
+CREATE FUNCTION
+math::log(x: std::decimal, NAMED ONLY base: std::decimal) -> std::decimal
+{
+    FROM SQL $$
+    SELECT log("base", "x")
+    $$;
+};
+
+
+# math::mean
+# -----------
+# The mean function returns an empty set if the input is empty set. On
+# all other inputs it returns the mean for that input set.
+CREATE FUNCTION
+math::mean(vals: SET OF std::decimal) -> std::decimal
+{
+    FROM SQL FUNCTION 'avg';
+};
+
+
+CREATE FUNCTION
+math::mean(vals: SET OF std::int64) -> std::decimal
+{
+    FROM SQL FUNCTION 'avg';
+};
+
+
+CREATE FUNCTION
+math::mean(vals: SET OF std::float64) -> std::float64
+{
+    FROM SQL FUNCTION 'avg';
+};
+
+
+# math::stddev
+# ------------
+CREATE FUNCTION
+math::stddev(vals: SET OF std::decimal) -> std::decimal
+{
+    FROM SQL FUNCTION 'stddev';
+};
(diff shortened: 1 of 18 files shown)


edb/lib/std/30-strfuncs.eql
```diff
@@ -19,11 +19,84 @@
 ## String functions
 
 
-# std::lower
-# ----------
+CREATE FUNCTION
+std::ascii(val: std::str) -> std::int64
```
elprans: Given that this returns an integer Unicode codepoint, this should be called `ord()`.

edb/lib/std/30-strfuncs.eql
```diff
@@ -19,11 +19,84 @@
...
+
+CREATE FUNCTION
+std::upper(s: std::str) -> std::str
+{
+    FROM SQL FUNCTION 'upper';
+};
+
+
+CREATE FUNCTION
+std::title(s: std::str) -> std::str
```
elprans: I think spelling this out as `titlecase()` would be clearer. Likewise for `lower()` and `upper()`.

1st1: I think they all should have "str_" prefix. I'm -0 on the "case" suffix.

edb/lib/std/30-strfuncs.eql
vpetrovykh: Done

edb/lib/std/30-strfuncs.eql
vpetrovykh: I've added the "str_" prefix.

edb/lib/std/30-strfuncs.eql
```diff
@@ -19,11 +19,84 @@
...
+{
+    FROM SQL FUNCTION 'initcap';
+};
+
+
+CREATE FUNCTION
+std::lpad(s: std::str, n: std::int64, fill: std::str=' ') -> std::str
```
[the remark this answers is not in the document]
vpetrovykh: Done.

edb/lib/math.eql
```diff
@@ -0,0 +1,212 @@
...
+math::variance(vals: SET OF std::anyfloat) -> OPTIONAL std::float64
+{
+    FROM SQL FUNCTION 'variance';
+};
+
+
+# math::var_pop
+# -------------
+CREATE FUNCTION
+math::var_pop(vals: SET OF std::decimal) -> OPTIONAL std::decimal
```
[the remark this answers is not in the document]
vpetrovykh: I've renamed the "variance" to "var" as that is the typical way it's written in statistics formulas.

vpetrovykh:
I've annotated the x-fails with more details comments about the observed failure. Please take a look at them (you can search for "BUG").
@elprans, the last test in the file isolates the bug I was showing to you in the office.

edb/lib/std/30-strfuncs.eql
```diff
@@ -91,3 +91,86 @@ std::`>` (l: std::str, r: std::str) -> std::bool
 CREATE INFIX OPERATOR
 std::`>=` (l: std::str, r: std::str) -> std::bool
     FROM SQL OPERATOR r'>=';
+
+
+CREATE FUNCTION
+std::ord(val: std::str) -> std::int64
```
1st1: I'd drop `ord()` and `chr()`.  I only used them a few times in my life and I fail to come up with a good example of how they can be used in EdgeQL.

edb/lib/std/30-strfuncs.eql
```diff
@@ -91,3 +91,86 @@ std::`>` (l: std::str, r: std::str) -> std::bool
...
+
+CREATE FUNCTION
+std::str_trim(s: std::str, tr: std::str=' ') -> std::str
+{
+    FROM SQL FUNCTION 'btrim';
+};
+
+
+CREATE FUNCTION
+std::str_reverse(s: std::str) -> std::str
```
1st1: Let's have a polymorphic `reverse()` function that works on bytes, strings, and arrays.

edb/lib/std/20-genericfuncs.eql
```diff
@@ -103,3 +103,108 @@ std::random() -> std::float64
...
+    SET initial_value := False;
+    FROM SQL FUNCTION 'bool_or';
+};
+
+
+# std::round
+# ----------
+
+CREATE FUNCTION
+std::round(val: std::decimal) -> std::decimal
```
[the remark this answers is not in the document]
1st1: @vpetrovykh this still needs to be addressed.

edb/lib/std/30-strfuncs.eql
vpetrovykh: There's no function for reversing bytea or arrays in Postgres. How much sense does the function for reversing a string make? It seems to be a good generalization for languages that deal with iterables and sequences generically, but since we don't have these broad concents in EdgeQL, does it really make sense to reverse strings? There's also the trouble with reversing arbitrary unicode.

edb/lib/std/30-strfuncs.eql
vpetrovykh: Dropped.

edb/lib/std/30-strfuncs.eql
1st1: Yeah, drop it.

1st1:
`std::round()` should be `FROM SQL FUNCTION 'round'`; please fix

tests/test_edgeql_functions.py
```diff
@@ -1239,3 +1243,1141 @@ class TestEdgeQLFunctions(tb.QueryTestCase):
...
+            {True},
+            {True},
+            {True},
+            {True},
+        ])
+
+    # BUG: we don't yet validate that the return cardinality is 1 here
+    #
+    # Population standard deviation is not defined for an empty set.
+    @unittest.expectedFailure
```
1st1: Please rebase onto the most recent `master` branch and start using `@edb.tools.test.not_implemented(reason)` and `@edb.tools.test.xfail(reason)` decorators.

vpetrovykh: I've rebased and updated the tests.

edb/lib/std/30-arrayfuncs.eql
```diff
@@ -36,13 +36,13 @@ std::array_unpack(array: array<anytype>) -> SET OF anytype
 
 
 CREATE FUNCTION
-std::array_contains(array: array<anytype>, what: anytype) -> std::bool
+std::array_contains(haystack: array<anytype>, needle: anytype) -> std::bool
```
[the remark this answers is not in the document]
elprans: `array` and `element` seem better.

edb/lib/std/20-genericfuncs.eql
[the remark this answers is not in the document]
vpetrovykh:
```
cannot create the `std::round(val: std::decimal, d: std::int64)` function: overloading "FROM SQL FUNCTION" functions is allowed only when all functions point to the same SQL function
```
Basically, there's a version of `round` where you get to specify the decimal digits and it has to be explicitly written in SQL because it involves a downcast to `int4`. This seems to clash with `FROM SQL FUNCTION 'round';`.

tests/test_edgeql_functions.py
```diff
@@ -1239,3 +1243,965 @@ class TestEdgeQLFunctions(tb.QueryTestCase):
...
+            SELECT math::ceil(<decimal>-2.5);
+        ''', [
+            {2},
+            {3},
+            {-2},
+            {3},
+            {-2},
+        ])
+
+    # BUG: invoking the polymorphic function ceil() with an int
```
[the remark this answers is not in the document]
vpetrovykh: Fixed.

edb/lib/std/30-strfuncs.eql
vpetrovykh: Dropped.

edb/lib/std/30-arrayfuncs.eql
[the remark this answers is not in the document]
vpetrovykh: OK, I'm not terribly attached to the haystack/needle. Array/element will do just fine.

vpetrovykh:
I've left a comment regarding `round` in the first comment thread that mentioned it. 

TL;DR the `FROM SQL FUNCTION` is incompatible with the third `round` definition.

edb/lib/std/30-arrayfuncs.eql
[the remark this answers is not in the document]
vpetrovykh: Hm. Do you object to the haystack/needle terminology in other contexts? I use that for `find`, but I can change that to something else in principle.

edb/lib/std/20-genericfuncs.eql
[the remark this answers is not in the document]
1st1: got it

PR #337, opened 2018-11-28
Minimize the number of SQL statements in std bootstrap and cache them

elprans:
Bundling the entire std bootstrap sequence into a single script block
allows for somewhat faster execution and makes it easier to cache the
script.  Additionally, the std schema pickle is now cached in the data
directory.

### DIFF
--- a/edb/lang/common/devmode.py
+++ b/edb/lang/common/devmode.py
@@ -1,0 +1,101 @@
+#
+# This source file is part of the EdgeDB open source project.
+#
+# Copyright 2011-present MagicStack Inc. and the EdgeDB authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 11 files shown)


edb/lang/common/devmode.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+def read_dev_mode_cache(cache_key, path):
+    full_path = get_dev_mode_cache_dir() / path
+
+    if full_path.exists():
+        with open(full_path, 'rb') as f:
+            src_hash = f.read(16)
+            if src_hash == cache_key:
+                return pickle.load(f)
```
1st1: Let's issue a warning if there's an exception during pickling (and ignore it).

edb/lang/common/devmode.py
```diff
@@ -0,0 +1,94 @@
...
+
+def is_in_dev_mode() -> bool:
+    devmode = os.environ.get('__EDGEDB_DEVMODE', '0')
+    return devmode.lower() not in ('0', '', 'false')
+
+
+def get_dev_mode_cache_dir() -> os.PathLike:
+    if is_in_dev_mode():
+        root = pathlib.Path(__file__).parent.parent.parent.parent
+        cache_dir = (root / 'build' / 'cache')
```
1st1: Let's ensure that `root / 'build'` exists (it must if you're in the dev mode).  This would help us when we move `edb.lang.common` to `edb.common`.

PR #338, opened 2018-11-29
Restrict usage of abstract scalars and anytype.

vpetrovykh:
Casting into an abstract scalar or anytype is now forbidden.

Technically anytype is a valid type that is the supertype for both
anyscalar and Object. It is legal to use it in any place where a type is
expected. However, in most places that would create an illegal
expression, such as an illegal cast to an abstract scalar (via either
<anytype> or [IS anytype]). It is also illegal to use anytype in
EXTENDING clause (it creates ambiguity w.r.t. how the list of base types
is linearized). It is legal when anytype appears in function
signature definitions or as the type expression in IS operator.

Anytype cannot be a link or property target for the same reason as
anyscalar: the current implementation in unable to support this kind of
polymorphism (unlike object polymorphism).

Anytype is also special in the fact that it doesn't have a corresponding
entry in the schema module (no Type object).

### DIFF
--- a/edb/lang/edgeql/compiler/inference/types.py
+++ b/edb/lang/edgeql/compiler/inference/types.py
@@ -298,7 +298,16 @@ def __infer_typeref(ir, env):
 
 @_infer_type.register(irast.TypeCast)
 def __infer_typecast(ir, env):
-    return infer_type(ir.type, env)
+    stype = infer_type(ir.type, env)
+
+    # is_polymorphic is synonymous to get_is_abstract for scalars
+    if stype.is_polymorphic(env.schema):
+        raise ql_errors.EdgeQLError(
+            f'cannot cast into an abstract scalar '
+            f'{stype.get_displayname(env.schema)}',
+            context=ir.context)
+
+    return stype
 
 
 @_infer_type.register(irast.Stmt)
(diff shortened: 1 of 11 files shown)


edb/lang/edgeql/compiler/inference/types.py
```diff
@@ -298,7 +298,16 @@ def __infer_typeref(ir, env):
...
 @_infer_type.register(irast.TypeCast)
 def __infer_typecast(ir, env):
-    return infer_type(ir.type, env)
+    stype = infer_type(ir.type, env)
+
+    # is_polymorphic is synonymous to get_is_abstract for scalars
+    if stype.is_polymorphic(env.schema):
+        raise ql_errors.EdgeQLError(
+            f'cannot cast into an abstract scalar '
+            f'{stype.get_name(env.schema)}',
```
1st1: get_displayname()

edb/lang/schema/schema.py
```diff
@@ -286,6 +287,10 @@ def _get(self, name, *, getter, default, module_aliases):
             if result is not None:
                 return result
 
+        # if nothing was found, check if it's the special anytype
+        if name == 'anytype':
+            return s_pseudo.Any.create()
```
1st1: `Schema.get()` shouldn't resolve `anytype`. It's not in the schema.

vpetrovykh: You're right, it's a remnant from an attempt to make `Expr IS anytype` work. I'll remove it for now.

edb/lang/edgeql/compiler/inference/types.py
vpetrovykh: Corrected.

Issue #330, opened 2018-11-23
Concatenation operator


vpetrovykh:
We may want a separate concatenation operator as opposed to overloading `+` for that purpose.

The rationale is that `+` is sort of intuitive for string concatenation, but less so when we consider arrays. Although *concatenation* operation is actually a common one for those.

The issue arises if we consider that we want mathematical structures (maybe inspired by numpy) like arrays, vectors and matrices. Mathematically `+` is well-defined for them and it's definitely not a concatenation. Concatenation is still a useful operator for those mathematical structures (building up a larger vector by concatenating smaller components).

So it appears that there's a general concatenation operation that is applicable to a range of types. The proposal is to add a `|` operator for it.

Postgres uses `||` for both string and array concatenation, but the issue with that is that this symbol is also commonly used in other languages as "boolean or" (e.g. C, Java, JS, Swift, Go). For many of these the following expression is an idiom:
```
some_var || 'default'
```
In Postgres this would be a string concatenation of `some_var` and the string literal `'default'`, but in these other languages this expression means something similar to "take the value of some_var or 'default' if the variable is empty". The latter would be `some_var ?? 'default'` in EdgeQL.

So using Postrges syntax for this operator is not a good option. There aren't any other well-established symbols that are overwhelmingly commonly used as concatenation operator, so we can use `|` as being least likely to be misleading.

1st1:
The solution to this problem is simple: + is concatenation for strings and isn't defined for arrays. Big -1 on a separate concatenation operator: having it for strings is weird, having it for arrays is generally useless (concatenating arrays is 100x less common than concatenating strings)

PR #345, opened 2018-12-03
Implement array, string and byte concatenation via the `++` operator

elprans:
The current `+` operator is a poor choice for concatenation, especially
in the array case, since the latter may be interpreted as an
element-wise vector addition.  Instead, use `++` consistently as the
concatenation operator for all data types that support it.

Closes: #330.

### DIFF
--- a/edb/server/pgsql/backend.py
+++ b/edb/server/pgsql/backend.py
@@ -330,10 +330,6 @@ class Backend:
         plan.generate(block)
         ql_text = block.to_string()
 
-        if debug.flags.delta_execute:
-            debug.header('Delta Script')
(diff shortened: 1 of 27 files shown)


edb/server/pgsql/backend.py
```diff
@@ -330,10 +330,6 @@ def process_delta(self, delta, schema):
         plan.generate(block)
         ql_text = block.to_string()
 
-        if debug.flags.delta_execute:
-            debug.header('Delta Script')
-            debug.dump_code(ql_text, lexer='sql')
-
```
1st1: Is this intentional?

PR #353, opened 2018-12-06
Add test runner markers for more elaborate xfail marking

elprans:
Currently, we use `@unittest.expectedFailure` regardless of the reason
of the test not passing.  It is useful to a) discriminate between bugs
and not-yet-implemented features; and b) be able to provide a textual
explanation for the test's failure.  This commit adds two decorators:
`edb.tools.test.xfail` and `edb.tools.test.not_implemented` to
implement this.

### DIFF
--- a/edb/tools/test/__init__.py
+++ b/edb/tools/test/__init__.py
@@ -27,9 +27,15 @@ import click
 
 from edb.tools import edbcommands
 
+from .decorators import not_implemented
+from .decorators import xfail
+
 from . import loader
 from . import runner
 from . import styles
+
+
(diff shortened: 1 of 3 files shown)


edb/tools/test/__init__.py
```diff
@@ -27,6 +27,9 @@
 
 from edb.tools import edbcommands
 
+from .decorators import not_implemented  # noqa
+from .decorators import xfail  # noqa
```
1st1: Alternatively you can define `__all__ = ('not_implemented', 'xfail')` in this file

PR #365, opened 2018-12-13
edgeql: Verify that operators and implicit casts work.

vpetrovykh:
Add exhaustive tests for all valid scalar types and all defined
operators except `??` and `IF...ELSE` (these 2 operators have a
different implementation from the others). The tests validate that
the operators are legal between specific scalar types and that the
return type is also inferred correctly.

### DIFF
--- a/edb/lang/edgeql/compiler/expr.py
+++ b/edb/lang/edgeql/compiler/expr.py
@@ -356,6 +356,16 @@ def compile_IfElse(
 
     if_expr_type = inference.infer_type(if_expr, ctx.env)
     else_expr_type = inference.infer_type(else_expr, ctx.env)
+    cond_expr_type = inference.infer_type(condition, ctx.env)
+
+    # make sure that the condition is actually boolean
+    bool_t = ctx.env.schema.get('std::bool')
+    if not cond_expr_type.issubclass(ctx.env.schema, bool_t):
+        raise errors.EdgeQLError(
+            'if/else condition must be of type {}, got: {}'.format(
+                bool_t.get_displayname(ctx.env.schema),
+                cond_expr_type.get_displayname(ctx.env.schema)),
+            context=expr.context)
 
     result = if_expr_type.find_common_implicitly_castable_type(
         else_expr_type, schema=ctx.env.schema)
@@ -363,8 +373,8 @@ def compile_IfElse(
     if result is None:
         raise errors.EdgeQLError(
             'if/else clauses must be of related types, got: {}/{}'.format(
-                if_expr_type.get_name(ctx.env.schema),
-                else_expr_type.get_name(ctx.env.schema)),
+                if_expr_type.get_displayname(ctx.env.schema),
+                else_expr_type.get_displayname(ctx.env.schema)),
             context=expr.context)
 
     ifelse = irast.IfElseExpr(
@@ -472,7 +482,7 @@ def compile_TypeFilter(
     if not isinstance(arg_type, s_objtypes.ObjectType):
         raise errors.EdgeQLError(
             f'invalid type filter operand: '
-            f'{arg_type.get_name(ctx.env.schema)} '
+            f'{arg_type.get_displayname(ctx.env.schema)} '
             f'is not an object type',
             context=expr.expr.context)
 
@@ -480,7 +490,7 @@ def compile_TypeFilter(
     if not isinstance(typ, s_objtypes.ObjectType):
         raise errors.EdgeQLError(
             f'invalid type filter operand: '
-            f'{typ.get_name(ctx.env.schema)} is not an object type',
+            f'{typ.get_displayname(ctx.env.schema)} is not an object type',
             context=expr.type.context)
 
     return setgen.class_indirection_set(arg, typ, optional=False, ctx=ctx)
--- a/edb/lib/std/25-numoperators.eql
+++ b/edb/lib/std/25-numoperators.eql
@@ -24,6 +24,16 @@
 # for each permutation of common integer and floating-point
 # operand types to avoid casting overhead, as these operations
 # are very common.
+#
+# Our implicit casts do not coincide with PostgreSQL. In particular we
+# do not implicitly cast floats into decimals. The philosophy behind
+# that is that using decimal arithmetic should be opt-in. On the other
+# hand, if decimals are used they should not be accidentally switched
+# to floating point arithmetic. One of the consequences of this is
+# that we need to explicitly define operators for every legal
+# combination of floats and decimals (this concerns comparison
+# operators) as unlike PostgreSQL we cannot rely on every kind of
+# number casting implicitly into decimal.
 
 # EQUALITY
 
@@ -158,12 +168,22 @@ std::`?=` (l: OPTIONAL std::float64, r: OPTIONAL std::float64) -> std::bool
 
 
 CREATE INFIX OPERATOR
-std::`=` (l: std::decimal, r: std::decimal) -> std::bool
+std::`=` (l: std::decimal, r: std::anyreal) -> std::bool
     FROM SQL OPERATOR r'=';
 
 
 CREATE INFIX OPERATOR
-std::`?=` (l: OPTIONAL std::decimal, r: OPTIONAL std::decimal) -> std::bool
+std::`?=` (l: OPTIONAL std::decimal, r: OPTIONAL std::anyreal) -> std::bool
+    FROM SQL EXPRESSION;
+
+
+CREATE INFIX OPERATOR
+std::`=` (l: std::anyreal, r: std::decimal) -> std::bool
+    FROM SQL OPERATOR r'=';
+
+
+CREATE INFIX OPERATOR
+std::`?=` (l: OPTIONAL std::anyreal, r: OPTIONAL std::decimal) -> std::bool
     FROM SQL EXPRESSION;
 
 
@@ -300,12 +320,22 @@ std::`?!=` (l: OPTIONAL std::float64, r: OPTIONAL std::float64) -> std::bool
 
 
 CREATE INFIX OPERATOR
-std::`!=` (l: std::decimal, r: std::decimal) -> std::bool
+std::`!=` (l: std::decimal, r: std::anyreal) -> std::bool
     FROM SQL OPERATOR r'<>';
 
 
 CREATE INFIX OPERATOR
-std::`?!=` (l: OPTIONAL std::decimal, r: OPTIONAL std::decimal) -> std::bool
+std::`?!=` (l: OPTIONAL std::decimal, r: OPTIONAL std::anyreal) -> std::bool
+    FROM SQL EXPRESSION;
+
+
+CREATE INFIX OPERATOR
+std::`!=` (l: std::anyreal, r: std::decimal) -> std::bool
+    FROM SQL OPERATOR r'<>';
+
+
+CREATE INFIX OPERATOR
+std::`?!=` (l: OPTIONAL std::anyreal, r: OPTIONAL std::decimal) -> std::bool
     FROM SQL EXPRESSION;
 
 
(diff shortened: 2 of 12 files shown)


edb/lib/std/25-numoperators.eql
```diff
@@ -934,19 +1038,29 @@ std::`%` (n: std::decimal, d: std::decimal) -> std::decimal
...
 
+# need an explicit operator for int64 in order to guarantee the result
+# is float64 and not decimal
 CREATE INFIX OPERATOR
-std::`^` (n: std::float64, p: std::float64) -> std::float64
-    FROM SQL OPERATOR '^';
+std::`^` (n: std::int64, p: std::int64) -> std::float64
+{
+    FROM SQL OPERATOR r'*';
+    FROM SQL 'SELECT (n ^ p)';
```
elprans: Shouldn't you be casing `p` to `float8` here to achieve the desired effect?

edb/lib/std/25-numoperators.eql
```diff
@@ -300,12 +318,22 @@ std::`?!=` (l: OPTIONAL std::float64, r: OPTIONAL std::float64) -> std::bool
...
 
 CREATE INFIX OPERATOR
-std::`!=` (l: std::decimal, r: std::decimal) -> std::bool
+std::`!=` (l: std::decimal, r: std::anyreal) -> std::bool
+    FROM SQL OPERATOR r'<>';
+
+
+CREATE INFIX OPERATOR
+std::`?!=` (l: OPTIONAL std::decimal, r: OPTIONAL std::anyreal) -> std::bool
+    FROM SQL EXPRESSION;
```
1st1: Please use this style consistently:

```
{
     FROM SQL $$
     SELECT round("val")
     $$;
 };
```

edb/lib/std/25-numoperators.eql
```diff
@@ -378,14 +406,33 @@ std::`>` (l: std::float64, r: std::float64) -> std::bool
...
-std::`>` (l: std::anyreal, r: std::decimal) -> std::bool
+std::`>` (l: std::decimal, r: std::decimal) -> std::bool
+    FROM SQL OPERATOR r'>';
+
+
+CREATE INFIX OPERATOR
+std::`>` (l: std::float64, r: std::decimal) -> std::bool
+    FROM SQL $$
+        SELECT "l" > "r";
+    $$;
```
1st1: Please use this formatting style:

```
{
     FROM SQL $$
     SELECT round("val")
     $$;
 };
```

edb/lib/std/25-numoperators.eql
```diff
@@ -934,19 +1038,29 @@ std::`%` (n: std::decimal, d: std::decimal) -> std::decimal
...
+    FROM SQL OPERATOR r'*';
+    FROM SQL 'SELECT (n ^ p)';
+};
 
 
 CREATE INFIX OPERATOR
-std::`^` (n: std::int64, p: std::int64) -> std::decimal
+std::`^` (n: std::float32, p: std::float32) -> std::float32
 {
     FROM SQL OPERATOR '*';
```
1st1: Need a comment here clarifying why `^` is created from `*`.

tests/test_edgeql_calls.py
```diff
@@ -1141,3 +1141,137 @@ class TestEdgeQLFuncCalls(tb.QueryTestCase):
...
+        The operator IS is normally static, so special care needs to
+        be taken when it's used inside polymorphic functions.
+    ''')
+    async def test_edgeql_calls_33(self):
+        await self.con.execute('''
+            CREATE FUNCTION test::call33(
+                a: anytype
+            ) -> bool
+                FROM EdgeQL $$
+                    SELECT a IS float32
```
1st1: We should just prohibit using `IS` in polymorphic functions on polymorphic types.  We have no ETA when (if ever) we can add support for this.  So instead of xfail, please make it raise a proper exception.

tests/test_edgeql_calls.py
```diff
@@ -1141,3 +1141,137 @@ class TestEdgeQLFuncCalls(tb.QueryTestCase):
...
+
+                DROP FUNCTION test::call33(
+                    a: anyreal
+                );
+            ''')
+
+    @test.xfail('''
+        All of the results are False, whereas some are expected to be
+        True because the operator '^' should be polymorphic.
+    ''')
```
1st1: ?

Testing the return type of `^` operator can be done without putting it into a polymorphic function.  The purpose of this test is to show that `float32 ^ float32` is `float32`, right?  AFAIK you can fix this by adding a specialized `^` for `float32` and cast the result there.

tests/test_edgeql_expressions.py
```diff
@@ -380,6 +381,574 @@ class TestExpressions(tb.QueryTestCase):
...
+        '<float32>1',
+        '<float64>1',
+        # decimal
+        '<decimal>1',
+    ]
+    unorderable = scalar_samples[:4]
+    orderable = scalar_samples[4:-6]
+    datetime = scalar_samples[5:-6]
+    numeric = scalar_samples[-6:]
+    nonnumeric = scalar_samples[:-6]
```
1st1: ¯\_(ツ)_/¯ Don't do this.  Refactor as follows:

```
unorderable = [
        '<bool>True',
         '<uuid>"d4288330-eea3-11e8-bc5f-7faf132b1d84"',
         '<bytes>b"Hello"',
         '<json>"Hello"',
]

orderable = [
		         '<str>"Hello"',
         '<datetime>"2018-05-07T20:01:22.306916+00:00"',
         '<naive_datetime>"2018-05-07T20:01:22.306916"',
         '<naive_date>"2018-05-07"',
         '<naive_time>"20:01:22.306916"',
         '<timedelta>"20:01:22.306916"',
]

...

scalar_samples = unorderable + orderable + ...
```

tests/test_edgeql_functions.py
```diff
@@ -2254,9 +2267,21 @@ class TestEdgeQLFunctions(tb.QueryTestCase):
...
+    @unittest.skip('''
+        The cast distance of int16 is 2 for mean(int64) and
+        mean(float64). Currently instead of an error one of the
+        implementaitons will be used at random, so the test failure is
+        unstable.
+
+        Once the cast distance ambiguity is fixed the next issue will
+        be making sure that only decimals produce decimal result,
+        everything else is float32 or float64.
+    ''')
```
1st1: Just fix this -- `compiler/func.py` lines 96..110.

edb/lib/std/25-numoperators.eql
vpetrovykh: It is my understanding that at the level of SQL operators we don't want to have unnecessary casts. As far as I can tell (albeit based on the signature of `pow` and my limited tests with `maxint64 ^ some_integer_power`) PostgreSQL actually returns `float8` for two integer inputs.

A possible test of the underlying implementation might be `10^309` which is greater than maxfloat64 and should result in an overflow error, while `10^308` is still a `float64`. If this test sounds reasonable, I can add it.

I can add the above comment about PostgreSQL to that specific definition, too.

edb/lib/std/25-numoperators.eql
vpetrovykh: Right, and I should probably copy `/` instead of the `*` anyways because power behaves more like division than multiplication w.r.t. types and other meta-information. This is simply because `2^-2` is `1/4`.

tests/test_edgeql_expressions.py
vpetrovykh: Yeah, this is a historical artifact of how that testing set came to be. I'll refactor for the sake of sanity and clarity.

edb/lib/std/25-numoperators.eql
vpetrovykh: I've updated indentation to be consistent.

edb/lib/std/25-numoperators.eql
vpetrovykh: We consistently omit `{...}` for one-liners like `FROM SQL EXPRESSION`.

tests/test_edgeql_functions.py
vpetrovykh: Currently, a "fix" for ambiguity will result in breaking many tests (such as `ln(<int16>10)` being unable to choose between `int64` and `float64` implementations). Fixing that, will backfire on how `float32` + `int16` works, namely it will result in a `float64` or it will require explicit versions for every numeric scalar of every std and math function. We also don't yet have validation that the SQL function body returns what is promised in EdgeQL. All of this sounds like a whole separate PR that's concentrating on *functions*.

The bottom line is that this fix is not in the scope of this PR, but defining sane *implicit casts* and *operators* is. So it seems that much like with `decimal` we need to adopt a specific doctrine w.r.t. `float32`. Just how special is it? How contagious? And why? Similarly, how special are `int16` and `int32`?

tests/test_edgeql_calls.py
vpetrovykh: Dropping the test because polymorphic functions will not be supported in user-space.

tests/test_edgeql_calls.py
vpetrovykh: I'll make a separate small PR forbidding `IS` with polymorphic types.

vpetrovykh:
I've added tests for `??` and `IF/ELSE`.

edb/lang/edgeql/compiler/expr.py
```diff
@@ -356,6 +356,16 @@ def compile_IfElse(
...
     else_expr_type = inference.infer_type(else_expr, ctx.env)
+    cond_expr_type = inference.infer_type(condition, ctx.env)
+
+    # make sure that the condition is actually boolean
+    bool_t = ctx.env.schema.get('std::bool')
+    if not cond_expr_type.issubclass(ctx.env.schema, bool_t):
+        raise errors.EdgeQLError(
+            'if/else condition must be of type {}, got: {}'.format(
+                bool_t.get_name(ctx.env.schema),
+                cond_expr_type.get_name(ctx.env.schema)),
```
1st1: `.get_name()` -> `.get_displayname()`

edb/lib/std/25-numoperators.eql
```diff
@@ -24,6 +24,14 @@
 # for each permutation of common integer and floating-point
 # operand types to avoid casting overhead, as these operations
 # are very common.
+#
+# Our implicit casts do not coincide with PostgreSQL. In particular we
+# do not implicitly cast floats into decimals. The philosophy behind
+# that is that using decimal arithmetic should be opt-in. One of the
```
1st1: "decimal arithmetic should be opt-in." and that once you're doing decimal arithmetic you won't accidentally switch to doing floating point arithmetic.

edb/lib/std/25-numoperators.eql
```diff
@@ -930,23 +1034,43 @@ CREATE INFIX OPERATOR
 std::`%` (n: std::decimal, d: std::decimal) -> std::decimal
 {
     FROM SQL OPERATOR r'%';
-    FROM SQL 'SELECT n - floor(n / d) * d';
+    FROM SQL 'SELECT n - floor("n" / "d") * d';
```
1st1: `d` -> `"d"`

edb/lib/std/25-numoperators.eql
```diff
@@ -930,23 +1034,43 @@ CREATE INFIX OPERATOR
...
 
 
+# need an explicit operator for int64 in order to guarantee the result
+# is float64 and not decimal
 CREATE INFIX OPERATOR
-std::`^` (n: std::float64, p: std::float64) -> std::float64
-    FROM SQL OPERATOR '^';
+std::`^` (n: std::int64, p: std::int64) -> std::float64
+{
+    # We cheat here a bit by copying the `/` operator, since there is
```
1st1: "by copying the `/` operator" -> "by copying most of SQL operator metadata from the `/` SQL operator"

tests/test_edgeql_functions.py
```diff
@@ -2038,9 +2039,21 @@ class TestEdgeQLFunctions(tb.QueryTestCase):
             {50000000000000000000000000},
         ])
 
+    @unittest.skip('''
```
1st1: Please add "edb.server._testbase.skip()" decorator (just alias it to unittest.skip)

edb/lang/edgeql/compiler/expr.py
vpetrovykh: Replaced in that file.

edb/lib/std/25-numoperators.eql
vpetrovykh: Added the clarification that decimals should not be accidentally switched back to floats.

edb/lib/std/25-numoperators.eql
vpetrovykh: Corrected.

edb/lib/std/25-numoperators.eql
vpetrovykh: Replaced this comment with a slightly longer more explicit version in this file.

tests/test_edgeql_functions.py
vpetrovykh: Added an alias to the "decorators.py".

PR #368, opened 2018-12-14
Rework schema item attributes

elprans:
Schema attributes are named values associated with schema items and are 
designed to hold arbitrary user-defined schema-level metadata.

There are several shortcomings with the current implementation that need
tweaking.

First issue: attributes can currently be of any primitive type, which makes
it really hard to represent the attribute collection as a coherent whole
without resorting to some string serialization, and that can go sideways
quickly if the attribute type is reasonably complex. Thus, we stop
pretending to support arbitrary types in attributes, and only support
`std::str`.  Complex structures can be stored as JSON and coerced to a
specific EdgeQL type explicitly, so this change has no practical usability
loss.  This also means that abstract attribute declaration does not require
a type specification anymore, although we may revisit this in the future
and allow `std::str` sub-types to be used, to, for example, allow
specifying constraints on attribute values. Abstract attribute declarations
continue to be required for disambiguation purposes.

Second issue: standard schema item aspects, like link cardinality, are 
represented as attributes in the introspection schema. There is no reason
to do it, and this creates confusing duality, since some aspects are
represented as properties and links of the appropriate
`schema::Object` subtype. We stop doing this, and only reserve
`schema::Object.attributes` to things that are true annotations and bear no
effect on the core schema behavior. Currently the only attributes that
satisfy this requirement are `"title"` and `"description"`.

In DDL, the `SET <attribute-name> := <attribute-value>` is changed to
`SET ATTRIBUTE <attribute-name> := <attribute-value>`.

In schema DSL we allow setting attribute values on foreign schema items 
using the regular declaration syntax with a qualified name:

    type othermod::Foo:
        attribute attr := 'value'

Third issue: automatic inheritance of attributes is sometimes inconvenient
or undesirable. A "description" attribute, for example, is unlikely to be
useful when inherited. Thus, we make all attributes non-inheritable by
default, with an option to opt-in to automatic inheritance with the new
`INHERITABLE` qualifier for `CREATE ATTRIBUTE`:

    CREATE INHERITABLE ATTRIBUTE foo;


edb/lang/schema/objects.py
```diff
@@ -249,6 +249,15 @@ def __get__(self, instance, owner):
             return self
 
 
+class RefDict(struct.Struct):
+
+    local_attr = struct.Field(str)
+    attr = struct.Field(str)
+    backref_attr = struct.Field(str, default='subject')
+    requires_explicit_inherit = struct.Field(bool, default=False)
+    ref_cls = struct.Field(type)
```
1st1: BTW, can we make all of these fields frozen?

edb/lib/schema.eql
```diff
@@ -22,62 +22,22 @@
 
 CREATE MODULE schema;
 
-CREATE MODULE stdattrs;  # standard schema attributes need to be in a separate
```
1st1: Amen 🙏

docs/edgeql/ddl/attributes.rst
```diff
@@ -65,13 +67,13 @@ Remove a :ref:`schema attribute <ref_datamodel_attributes>`.
...
-    DROP ATTRIBUTE <name> ;
+    DROP ABSTRACT ATTRIBUTE <name> ;
 
 Description
 -----------
 
-``DROP ATTRIBUTE`` removes an existing schema attribute from the database
-schema.
+``DROP ABSTRACT ATTRIBUTE`` removes an existing schema attribute from
+the database schema.
```
1st1: Should probably mention that it can drop `INHERITABLE` attributes too, without the need to use the qualifier.

PR #369, opened 2018-12-14
Make sure that stdmode and testmode flags are used consistently.

vpetrovykh:
Forbid usage of `FROM SQL` and polymorphic parameter and return types
for user-defined functions.

Add tests for other commands that should not be accessible at user-level
(creating new operators or casts).

### DIFF
--- a/edb/lang/schema/functions.py
+++ b/edb/lang/schema/functions.py
@@ -613,13 +613,36 @@ class CreateFunction(CreateCallableObject, FunctionCommand):
         return_typemod = self.scls.get_return_typemod(schema)
         from_function = self.scls.get_from_function(schema)
         has_polymorphic = params.has_polymorphic(schema)
+        polymorphic_return_type = return_type.is_polymorphic(schema)
         named_only = params.find_named_only(schema)
 
-        if return_type.is_polymorphic(schema) and not has_polymorphic:
+        # Certain syntax is only allowed in "EdgeDB developer" mode,
+        # i.e. when populating std library, etc.
+        if not context.stdmode and not context.testmode:
+            if has_polymorphic or polymorphic_return_type:
+                raise ql_errors.EdgeQLError(
+                    f'cannot create `{signature}` function: '
+                    f'generic types are not supported in '
+                    f'user-defined functions',
+                    context=self.source_context)
+            elif from_function:
+                raise ql_errors.EdgeQLError(
+                    f'cannot create `{signature}` function: '
+                    f'"FROM SQL FUNCTION" is not supported in '
+                    f'user-defined functions',
+                    context=self.source_context)
+            elif language != qlast.Language.EdgeQL:
+                raise ql_errors.EdgeQLError(
+                    f'cannot create `{signature}` function: '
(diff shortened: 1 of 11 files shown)


edb/lang/schema/functions.py
```diff
@@ -613,9 +613,32 @@ def _create_begin(self, schema, context):
...
         named_only = params.find_named_only(schema)
 
-        if return_type.is_polymorphic(schema) and not has_polymorphic:
+        # Certain syntax is only allowed in "EdgeDB developer" mode,
+        # i.e. when populating std library, etc.
+        if not context.stdmode and not context.testmode:
+            if has_polymorphic or polymorphic_return_type:
+                raise ql_errors.EdgeQLError(
+                    f'cannot create `{signature}` function: '
+                    f'polymorphic types are not supported in '
```
elprans: I'm thinking we went with slightly incorrect terminology with respect to `any` and friends.  IMO, `generics` or `generic types` is a more precise term.  Polymorphism is a property of functions and operators, so saying that something is a "polymorphic function" is OK.

elprans: \cc @1st1  ^

1st1: +1

elprans:
> Please add a test that you can't create any function in one of our standard modules (std, math, etc).

Not just added, but modified as well. `std` should be readonly post-bootstrap.

edb/lang/schema/functions.py
vpetrovykh: Wording updated.

vpetrovykh:
I've added more tests for the things that should not be accessible to the user.

edb/lang/schema/functions.py
```diff
@@ -613,13 +613,36 @@ def _create_begin(self, schema, context):
...
+            if has_polymorphic or polymorphic_return_type:
+                raise ql_errors.EdgeQLError(
+                    f'cannot create `{signature}` function: '
+                    f'generic types are not yet supported in '
+                    f'user-defined functions',
+                    context=self.source_context)
+            elif from_function:
+                raise ql_errors.EdgeQLError(
+                    f'cannot create `{signature}` function: '
+                    f'"FROM SQL FUNCTION" is not yet supported in '
```
1st1: "is not yet" -> "is not"

edb/lang/schema/functions.py
```diff
@@ -613,13 +613,36 @@ def _create_begin(self, schema, context):
...
+                    f'user-defined functions',
+                    context=self.source_context)
+            elif language != qlast.Language.EdgeQL:
+                raise ql_errors.EdgeQLError(
+                    f'cannot create `{signature}` function: '
+                    f'"FROM {language}" is not yet supported in '
```
1st1: "is not yet" -> "is not"

edb/lang/schema/functions.py
vpetrovykh: Done

edb/lang/schema/functions.py
vpetrovykh: Done

PR #370, opened 2018-12-16
Improve polymorphic resolution and implicit casting

elprans:
This PR has several commits improving the behavior of polymorphic 
function and operator resolution, as well as improvements for 
function and operator aliasing in std.


edb/lang/schema/objects.py
```diff
@@ -223,16 +223,17 @@ def __repr__(self):
 
 class SchemaField(Field):
 
-    __slots__ = ('default', 'hashable')
+    __slots__ = ('default', 'hashable', 'stdonly')
```
1st1: We already have `public` to allow fields to be set from DDL.  We could allow setting **any** field when in stdmode, and then we don't need your flag.  IMHO that's a better solution than adding yet another flag.

elprans: Sure.  This would require a separate commit to mark all the necessary fields as public, as we don't currently rely on it in DDL at all.

elprans: Another difficulty here is that some fields are set programmatically by the DDL pipeline as a result of some syntax.  For example `Function.language` and `Function.code` are technically public, since you can set them with a DDL command, but then also `SET language :=` should be illegal.  Since we do the check in `AlterObjectProperty`, it makes it hard to distinguish between the two.

elprans: OK, I renamed `public` to `allow_ddl_set` and switched everything to this whitelist in an additional commit.

Issue #376, opened 2018-12-21
Assignment casts


vpetrovykh:
Assignment casts are in a lot of ways a convenience and usability feature. For example, we're pretty sure that we want to allow casting a `str` into `uuid` or `datetime` on assignment (such as in `INSERT` or `UPDATE`). It just reduces the verbosity.

Things get a bit weirder with numerical scalars, there we also seem to want to be able to assign any numerical type to any other one (as long as the explicit cast exists). Let's say we failed to find an implicit cast that does the trick, so we start looking for an assignment cast (maybe from `float64` to `int64` like in `foo := 3^5`). Unlike the directed graph of implicit casts that had no cycles, the assignment cast graph necessarily has cycles (remember that technically any numeric is fair game as assignment to any other numeric).

Every hop on the implicit cast graph is technically "safe". It incurs a slight performance penalty, but the idea is that in principle the data does not degrade, nor does it cause exceptions. Assignment casts are not safe: they can degrade the data (`float64` -> `int64`) or even cause exceptions (`str` -> `int64`) on some inputs. So it seems that it is no longer a safe assumption that just because *A* is assignment-castable to *B* and *B* is assignment-castable to *C* that it's safe to conclude that *A* is assignment-castable to *C*. For example imagine that we decide one day that `uuid` is assignment-castable to `str`. This in no way implies that it is possible to assignment-cast `uuid` to `int64`. So assignment casts are *non-transitive*. This means that unless we have an explicit one-hop assignment cast declared between two types, it's not fair game to say there is one.

The above is complicated by the fact that we may be dealing with types that are derived from the basic scalars, so they will not have explicit assignment casts declared for the exact types. So we would need to look for possible valid casts specifically between the earliest non-abstract parent types.

1st1:
So what is this issue about? Do we have a bug to fix? What's the action?

vpetrovykh:
Are we in agreement that assignment casts are non-transitive? I can adjust the algorithms for them and that would fix currently failing assignment cast tests. 

Basically the algorithm for finding assignment cast distance would be:
For every possible assignment cast add the implicit cast distance from the source to cast source and the distance from the cast target to target + 1 (the +1 is for the single assignment cast jump itself). From this point on the distance can be used in our existing assignment cast algorithms. This would allow defining cyclical assignment casts without causing ambiguity since only one hop would be used anyway.

vpetrovykh:
[...]

PR #377, opened 2018-12-21
Fix assignment casts.

vpetrovykh:
Assignment casts between different numeric types had gaps so that it was
not possible to assign a `float32` to an `int32`, etc. In general the
assignment casts logic seemed to rely on these casts being transitive
just as implicit casts.

The new assignment casts are non-transitive and have to be explicitly
defined for all valid pairs. However, in broad terms, implicit casts can
be used to connect the end points of an explicit assignment cast.

### DIFF
--- a/edb/lang/schema/casts.py
+++ b/edb/lang/schema/casts.py
@@ -108,8 +108,33 @@ def find_common_castable_type(
 @functools.lru_cache()
 def is_assignment_castable(
         schema, source: s_types.Type, target: s_types.Type) -> bool:
-    dist = _is_reachable(schema, {'assignment': True}, source, target, 0)
-    return dist != _NOT_REACHABLE
+    # assignment castable has 3 components:
+    # 1) there must exists some one-hop assignment cast, s.t.
+    # 2) there's an implicit cast available from source to assn_cast.source
+    # 3) there's an implicit cast available from assn_cast.target to target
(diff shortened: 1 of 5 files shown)


vpetrovykh:
Basically, I have at least some of the simple "every numeric to every other numeric" assignment casts working in it. I'm not too clear yet on what SQL should be generated, though, so the current fix I have merely addresses the question of "is there a possible assignment cast?". So I'd like some feedback before I proceed with more changes.

(related to #376)

edb/lang/schema/casts.py
```diff
@@ -108,8 +108,15 @@ def find_common_castable_type(
...
         schema, source: s_types.Type, target: s_types.Type) -> bool:
-    dist = _is_reachable(schema, {'assignment': True}, source, target, 0)
-    return dist != _NOT_REACHABLE
+
+    # Assignment casts are valid only as one-hop casts.
+    casts = schema.get_casts_to_type(target, assignment=True)
+    if not casts:
+        return False
+
+    sources = {c.get_from_type(schema) for c in casts}
```
1st1:
```
for c in casts:
    if c.get_from_type(schema) == source:
        return True
return False
```

vpetrovykh: Fixed.

PR #379, opened 2019-01-04
Remove the dependency of IR on schema

elprans:
At present, we record schema objects into IR directly, which
means that in order to process the IR, the derived schema must be kept
and passed around.  This also makes any IR serialization impractical, as
the associated schema would need to be serialized as well.

This change moves the necessary information from the schema objects into
the IR directly removing the IR dependency on the schema.

### DIFF
--- a/edb/lang/common/ast/base.py
+++ b/edb/lang/common/ast/base.py
@@ -20,6 +20,7 @@
 import copy
 import collections.abc
 import functools
+import sys
 import typing
 
 import typing_inspect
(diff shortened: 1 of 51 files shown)


edb/lang/common/ast/base.py
```diff
@@ -53,8 +54,11 @@ def __new__(mcls, name, bases, dct):
         if '__annotations__' not in dct:
             return cls
 
+        globalns = dict(sys.modules[cls.__module__].__dict__)
```
1st1: Make a ChainMap here please

elprans: `typing` wants a real dict here, unfortunately

PR #382, opened 2019-01-07
Use schema object ids instead of names as much as possible

elprans:
Type names are now used exclusively for annotation, actual compilation
uses type ids instead.  Most objects are now reflected into the backend
using their ids.  Callables and other polymorphic items are notable
exceptions.


edb/lang/ir/ast.py
```diff
@@ -63,11 +63,10 @@ def __repr__(self):
 class TypeRef(Base):
     # The id of the referenced type
     id: uuid.UUID
+    # The module id of the referenced type
+    module_id: typing.Optional[uuid.UUID]
```
1st1: Should it be optional?

Issue #219, opened 2018-07-06
Types, type expressions, and type operators


1st1:
This issue documents our latest stance on how types and type-related operations should be defined in EdgeQL.

## 1. Union Types

For the **UNION** set operator we compute a Union Type for its lefthand and righthand sets.  We define *union types* only for Object Types; there is no concept of a *union type* for Scalar or Collection types.

A union type for scalar (or collection) types is a type that all of the union-ed types can **implicitly** cast to.

A union type for a single object type is that type. E.g. a union type for `Foo` object type is `Foo`.

A union type for two or more object types is an object type that describes an **intersection** of all properties and links of its types.  For example, for the following two types:

```
type Foo:
  property a -> int
  property b -> int

type Bar:
  property a -> int
```

a union type would be defined as:

```
type UnionOfFooAndBar:
  property a -> int
```

The following EdgeQL operators can produce union types:

* `UNION`
* `IF..ELSE`
* `??`—coalesce operator

## 2. Type Expressions

In EdgeQL we have a few places where types can appear in:

* type casts: `<type><expr>` as in `<int>"123"`;

* righthand side of the **IS** operator: `<expr> IS <type>` as in `(SELECT ...) IS User`;

* in **IS** operator in paths: `<path>[IS <type>]` as in `Card.<deck[IS User]`;

* DDL commands, e.g. `CREATE REQUIRED LINK <name> -> <type>`; etc.

This can be generalized by replacing the `<type>` grammar production with a `<type_expr>`—a type expression.

Type Expressions can appear only in the afore mentioned situations and are defined as follows:

* `Name` production, e.g.`NAME` or `NAME::NAME`.

* `Name < ... >` set of productions; allows to specify collection and scalar types like `array<int>`.

* `&` operator is used to create Intersection Types (detailed in a section below).

* `|` operator is used to create Union Types; it has lower precedence than `&`.

* `(` and `)` can be used for grouping.

A few valid examples of type expressions:

* `User | SystemUser`—a union type of type `User` and type `SystemUser`.

* `int16 | int32`—will evaluate to `int32` as that is what type `{<int16>1} UNION {<int32>2}` expression would evaluate to.

* `std::array<std::int64>`—an array of `int64`.

* `(User | SystemUser) & Named`—an intersection type of `Named` with a union type of `User` and `SystemUser`.

## 3. Intersection Types

Intersection types for scalar and collection types are not defined.

[...]

PR #389, opened 2019-01-09
Add rudimentary support for the INTROSPECT operator

elprans:
This adds support for the `INTROSPECT` operator, which is a way to
perform introspection of a given type expression.  Support for
the `TYPEOF` expression is added also, which is a way to perform
static compile-time type inference for an arbitrary expression.

Currently, `INTROSPECT` is limited to the types that are present in the
schema, i.e. introspection of collection types and views is not
supported yet.

I've also converted the uses of `__type__` on scalar expressions to
`INTROSPECT TYPEOF`, as `__type__` on non-object types is slated for
removal.

Issue #219.

### DIFF
--- a/docs/datamodel/colltypes.rst
+++ b/docs/datamodel/colltypes.rst
@@ -38,8 +38,11 @@ heterogeneous data.
 
     .. code-block:: edgeql-repl
 
-        db> SELECT ('foo', 42).__type__.name;
-        {"std::tuple<std::str, std::int64>"}
+        db> SELECT ('foo', 42);
+        {('foo', 42)}
+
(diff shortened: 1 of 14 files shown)


docs/datamodel/colltypes.rst
```diff
@@ -38,8 +38,11 @@ heterogeneous data.
 
     .. code-block:: edgeql-repl
 
-        db> SELECT ('foo', 42).__type__.name;
-        {"std::tuple<std::str, std::int64>"}
+        db> SELECT ('foo', 42);
+        {('foo', 42)}
+
+        db> SELECT (INTROSPECT TYPEOF ('foo', 42)).name;
```
1st1: Need to document `INTROSPECT` and `TYPEOF`

Issue #356, opened 2018-12-08
Maintain non-decimal/decimal arithmetic separation


vpetrovykh:
Decimals are great for precision, but they are slow. In many cases speed is more important for applications. Our philosophy is that we don't want to automatically upgrade non-decimal arithmetic to decimals because this is difficult to trace, but has a real impact on performance. On the other hand once you opt-in, decimals are contagious and will generally stay decimals.

Additionally, since the point of decimals is the perfect precision we disallow mixing decimals and floats (without explicit casts) since most float literals don't actually represent exact values, while decimal literals are exact. Mixing them causes ambiguity.

This has consequences to certain functions like `sum`, `mean`, `stddev`, `var`, etc. which in Postgres will return a decimal on integer input. We want to make their return type `int64` and `float64` for non-decimal inputs and `decimal` *only* if the input set is a `decimal` set. So we need to cast the output of these aggregates back into the non-decimal types.

We may add an optimization, such that when the compiler sees that the integer input set is being explicitly cast into `<decimal>` for these functions, both the input cast and the output cast will be omitted from SQL, thus resulting in the correct result and avoiding the casting performance penalty.

vpetrovykh:
Related to this there's a whole bunch of considerations w.r.t. operators and casts.

We start by realizing that it's very inconvenient if we can't compare decimals to floats (`a < b` or even the possibly far less useful `a = b`). So the impulse is to suggest that although `+` and other arithmetic is not defined for a mix of decimal and float, but `=` and `<` are defined for all numerical type combinations.

Comparison operators are related to each other via boolean operators, though. We expect that the following should be equivalent:
```
SELECT a = b;
SELECT NOT (a != b);
SELECT a >= b AND a <= b;
SELECT NOT (a > b OR a <b);
SELECT a IN b AND b IN a;
```
It makes sense that these expressions are all equivalent, because otherwise refactoring expressions may become a very unintuitive and error-prone process. For the moment let's ignore the possible types for which `=` and `!=` are defined, but `>` and `<` are not (we probably don't have unorderable types right now - meaning any type is valid in the `ORDER BY` clause, which is another useful property to have).

A natural extension of the above concept would seem to lead to the following also being equivalent:
```
SELECT a = b;
SELECT (a,) = (b,);
SELECT [a] = [b];
```
After all, we want to be able to compare tuples or arrays and sort results based on them. And it would be a very surprising thing if extracting that single element from the container would yield different results.

[...]

PR #390, opened 2019-01-10
Add "recursuve" field to operators.

vpetrovykh:
In principle comparison operators for arrays and tuples should have
their behavior derived from the comparison of the elements. We mark
these operators as "recursive" so that the compiler knows how to
validate them.

Only operators with the operands as either all arrays or all tuples can
be recursive. It's illegal to mix recursive and non-recursive versions
of an operator when overloading.

### DIFF
--- a/edb/edgeql/compiler/func.py
+++ b/edb/edgeql/compiler/func.py
@@ -28,6 +28,7 @@ from edb.ir import ast as irast
 from edb.ir import typeutils as irtyputils
 from edb.ir import utils as irutils
 
+from edb.schema import functions as s_func
 from edb.schema import name as sn
 from edb.schema import types as s_types
 
@@ -149,7 +150,8 @@ def compile_operator(
         ctx: context.ContextLevel) -> irast.OperatorCall:
 
     env = ctx.env
-    opers = env.schema.get_operators(op_name, module_aliases=ctx.modaliases)
+    schema = env.schema
+    opers = schema.get_operators(op_name, module_aliases=ctx.modaliases)
 
     if opers is None:
         raise errors.QueryError(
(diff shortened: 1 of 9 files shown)


vpetrovykh:
This is related to #356 in that the way arrays and tuples of decimals are compared to non-decimals is different from the way their elements would compare.

edb/edgeql/compiler/func.py
```diff
@@ -179,7 +181,64 @@ def compile_operator(
...
+
+                # Now that we have an operator, we need to validate that it
+                # can be applied to the tuple or array elements.
+                submatched = validate_recursive_operator(
+                    opers, args[0], args[1], ctx=ctx)
+
+                if len(submatched) != 1:
+                    # generate the error message corresponding to the number
+                    # of matches
+                    matched = submatched
```
1st1: I think the comment doesn't match the code, no?

edb/schema/operators.py
```diff
@@ -60,6 +60,9 @@ class Operator(s_func.CallableObject, s_abc.Operator):
     commutator = so.SchemaField(
         so.Object, default=None, compcoef=0.99)
 
+    recursive = so.SchemaField(
```
elprans: Make sure you read and populate the field in `intromech.read_operators`

edb/edgeql/compiler/func.py
vpetrovykh: I've reworded the comment to clarify what I mean.

edb/schema/operators.py
vpetrovykh: Done.

1st1: Not quite, you also need to update `pgsql/datasources/schema/operators.py`

vpetrovykh: OK, added `0.recursive` to the query there.

PR #403, opened 2019-01-17
docs: Update function declaration syntax.

vpetrovykh:
Remove the `$` from function declaration syntax and make `typequal`
production more explicit.

### DIFF
--- a/docs/datamodel/functions.rst
+++ b/docs/datamodel/functions.rst
@@ -19,11 +19,19 @@ A function may be defined in EdgeDB Schema using the ``function`` declaration:
 
     # where <argspec> is:
 
-    [ <argkind> ] $<argname>: [ <typequal> ] <argtype> [ = <default> ]
+    [ <argkind> ] <argname>: [ <typequal> ] <argtype> [ = <default> ]
+
+    # <argkind> is:
+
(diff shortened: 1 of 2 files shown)


docs/datamodel/functions.rst
```diff
@@ -19,11 +19,19 @@ A function may be defined in EdgeDB Schema using the ``function`` declaration:
...
-    [ <argkind> ] $<argname>: [ <typequal> ] <argtype> [ = <default> ]
+    [ <argkind> ] <argname>: [ <typequal> ] <argtype> [ = <default> ]
+
+    # <argkind> is:
+
+    [ { variadic | named only } ]
+
+    # <typequal> is:
+
+    [ { set of | optional } ]
```
1st1: Do support 'SINGLETON' there?  It's the default, and I'm wondering if our parser understands it.  If not -- it's ok, if yay -- need to document that.

vpetrovykh: No, we don't support an explicit `SINGLETON` there. In fact, `SINGLETON` is no longer a keyword at all. We have another keyword `SINGLE` to be used specifically with links/properties.

1st1: Alright then.

PR #417, opened 2019-01-24
Implement UNION, ??, and IF/ELSE as generic operators

elprans:
See individual commits.

### DIFF
--- a/edb/edgeql/compiler/func.py
+++ b/edb/edgeql/compiler/func.py
@@ -300,12 +300,17 @@ def compile_operator(
     matched_params = oper.get_params(env.schema)
     rtype = matched_call.return_type
 
-    if oper_name == 'std::UNION' and rtype.is_object_type():
-        # Special case for UNION operator, instead of common parent type,
-        # we return a union type.
-        left_type = setgen.get_set_type(args[0].expr, ctx=ctx).material_type(
+    if oper_name in {'std::UNION', 'std::IF'} and rtype.is_object_type():
+        # Special case for the UNION and IF operators, instead of common
+        # parent type, we return a union type.
+        if oper_name == 'std::UNION':
+            larg, rarg = (a.expr for a in args)
+        else:
+            larg, rarg = (a.expr for a in args[1:])
+
+        left_type = setgen.get_set_type(larg, ctx=ctx).material_type(
             ctx.env.schema)
-        right_type = setgen.get_set_type(args[1].expr, ctx=ctx).material_type(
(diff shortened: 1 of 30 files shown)


edb/edgeql/compiler/func.py
```diff
@@ -468,8 +485,15 @@ def finalize_args(bound_call: polyres.BoundCall, *,
...
         val_material_type = barg.valtype.material_type(ctx.env.schema)
         param_material_type = paramtype.material_type(ctx.env.schema)
-        if not val_material_type.issubclass(
-                ctx.env.schema, param_material_type):
+
+        compatible = (
+            val_material_type.issubclass(ctx.env.schema, param_material_type)
+            and (not param_material_type.is_tuple()
+                 or (param_material_type.get_element_names() ==
+                     val_material_type.get_element_names()))
```
1st1: `element_types`?

edb/edgeql/parser/grammar/ddl.py
```diff
@@ -1775,6 +1775,9 @@ def reduce_POSTFIX(self, *kids):
     def reduce_PREFIX(self, *kids):
         self.val = kids[0].val
 
+    def reduce_NARY(self, *kids):
```
1st1: ternary?

1st1: We can add N-ary when we need to, but if/else it ternary, right?

tests/test_edgeql_syntax.py
```diff
@@ -2368,10 +2368,10 @@ def test_edgeql_syntax_updatefor_01(self):
 
     def test_edgeql_syntax_coalesce_01(self):
         """
-        SELECT a ?? x;
-        SELECT a ?? x.a;
-        SELECT a ?? x.a[IS ABC];
-        SELECT (a ?? x.a[IS ABC]@aaa + 1);
+        SELECT (a ?? x);
```
1st1: Why these changes?

edb/edgeql/parser/grammar/ddl.py
elprans: Yes, but ternary and n-ary are the same from the standpoint of handling them, so it makes little sense in having a separate kind for every value of N.

1st1: Arent' binary and ternary and n-ary all the same from the standpoint of handling them? ;)

edb/edgeql/compiler/func.py
1st1: Can you add a comment?

PR #442, opened 2019-02-01
docs: Document on target delete triggers.

vpetrovykh:
Document all the different behavior options for when a link target is
deleted without explicitly updating the link.

### DIFF
--- a/docs/datamodel/links.rst
+++ b/docs/datamodel/links.rst
@@ -81,6 +81,10 @@ declaration in the context of a ``type`` declaration:
             [ readonly := {true | false} ]
             [ <attribute-declarations> ]
             [ <constraint-declarations> ]
+            [ on target delete { restrict |
+                                 allow |
+                                 delete source |
+                                 deferred restrict } ]
 
     # shorthand form for computable link declaration:
 
@@ -133,6 +137,29 @@ Parameters:
 
 :eschema:synopsis:`<constraint-declarations>`
     :ref:`Constraint <ref_datamodel_constraints>` declarations.
+
+:eschema:synopsis:`on target delete`
+    On target delete options cover the situation when the target
+    object of a link is deleted without explicitly updating the link.
+
+:eschema:synopsis:`restrict`
+    Prohibit deleting the link target as long as the source object exists.
+    This is the default behavior.
+
+:eschema:synopsis:`allow`
+    Allow dropping the connection between the source and target when
+    the target is deleted.
+
+:eschema:synopsis:`delete source`
+    Delete the source object if any link target is deleted. This means
+    that for ``multi`` links the source object will be deleted
+    if even one of the link targets is deleted (e.g. automatically
+    dissolving a team when all team members are critical and one has
+    been deleted).
+
+:eschema:synopsis:`deferred restrict`
+    Same as ``restrict``, but the check is performed at the end of
+    transaction instead of immediately.
 
 
 Concrete links can also be defined using the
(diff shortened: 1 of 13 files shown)


docs/datamodel/links.rst
```diff
@@ -134,6 +139,36 @@ Parameters:
 :eschema:synopsis:`<constraint-declarations>`
     :ref:`Constraint <ref_datamodel_constraints>` declarations.
 
+.. _ref_datamodel_links_on_delete:
+
+On target delete options cover the situation when the target object of
```
1st1: :synapsis:`on target delete`

docs/datamodel/links.rst
```diff
@@ -134,6 +139,36 @@ Parameters:
...
+a link is deleted without explicitly updating the link:
+
+:eschema:synopsis:`restrict`
+    Prohibit deleting the link target as long as the parent object exists.
+    This is the default behavior.
+
+:eschema:synopsis:`delete source`
+    Delete the parent object if any link target is deleted. This may
```
1st1: I'm not sure that child/parent treminology is the correct one to use here. @elprans?

docs/datamodel/links.rst
```diff
@@ -134,6 +139,36 @@ Parameters:
...
+    not be desirable behavior for ``multi`` links since the parent
+    object will be deleted if *just* one of the link targets is
+    deleted.
+
+:eschema:synopsis:`set empty`
+    Set the link to be empty when the target object is deleted. This
+    may not be desirable behavior for ``multi`` links since the link
+    will be set to an empty set if *just* one of the link targets is
```
1st1: not sure what "just one of the link targets deleted" means here

docs/datamodel/links.rst
elprans: Yeah, let’s stick to calling it *source*.

docs/datamodel/links.rst
elprans: Reading this I’m starting to think that `set empty` is a misnomer. What we actually *want* here is to simply delete the corresponding link and nothing else. So, yeah, the link set becomes *empty* if the last target was removed, but that is a side effect, not an explicit action itself. We should rename the action to `accept` or `delete link` or something along these lines.

Similarly, I think `delete source` on a multilink doesn’t strike me like a particularly useful thing. Perhaps we should ban this for now.

vpetrovykh: `delete source` for multilinks is perhaps more unusual, but it's perfectly reasonable for situations where the multilink belongs to some grouping where the integrity of the group is paramount: like a competition team (can't have incomplete teams). 

I'll drop the `set default` as it seems to be broken right now.

I'll rename `set empty` to `accept`.

docs/datamodel/links.rst
vpetrovykh: ok

docs/datamodel/links.rst
vpetrovykh: Maybe `allow` is better than `accept`? Our default mode is "restrict" so "accept/reject" vs. "allow/restrict".

vpetrovykh: I've updated the PR. It's no longer just documentation, but also has parser and test changes.

docs/datamodel/links.rst
```diff
@@ -134,6 +138,29 @@ Parameters:
...
+
+:eschema:synopsis:`delete source`
+    Delete the source object if any link target is deleted. This means
+    that for ``multi`` links the source object will be deleted
+    if even one of the link targets is deleted (e.g. automatically
+    dissolving a team when all team members are critical and one has
+    been deleted).
+
+:eschema:synopsis:`deferred restrict`
+    Similar to restrict, but no action is taken if the link is
```
elprans: Better wording: "Same as `restrict`, but the check is performed at the end of transaction instead of immediately".

vpetrovykh: Fixed

PR #444, opened 2019-02-02
Fix an issue with repl mishandling multiple `;`

vpetrovykh:
The `;` is supposed to be a separator, so multiple `;` in a row should
have no effect. REPL should not be attempting to execute a statement
that's just a `;`.

### DIFF
--- a/edb/repl/__init__.py
+++ b/edb/repl/__init__.py
@@ -71,7 +71,8 @@ def is_multiline():
         return False
 
     if text.endswith(';'):
-        return not lexutils.split_edgeql(text)
+        _, incomplete = lexutils.split_edgeql(text, script_mode=False)
+        return incomplete is not None
 
     return True
 
@@ -331,10 +332,10 @@ class Cli:
                         results.append(
                             self.connection._execute_graphql(command))
                     elif qm is context.QueryMode.Normal:
-                        for query in lexutils.split_edgeql(command):
+                        for query in lexutils.split_edgeql(command)[0]:
                             results.append(self.connection.fetch(query))
                     else:
-                        for query in lexutils.split_edgeql(command):
+                        for query in lexutils.split_edgeql(command)[0]:
                             results.append(self.connection.fetch_json(query))
 
                 except KeyboardInterrupt:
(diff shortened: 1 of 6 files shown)


vpetrovykh:
I'm not sure if it's intended, but the "alt + enter" way of terminating a statement in REPL no longer works if the statement is not terminated by a `;`. 

As far as I can tell the problem is that for a single statement unterminated by `;` the `out` is going to be an empty list, the `buffer` will not be empty and the `script_mode` will be `False`. So a `None` is returned, which repl cannot iterate over.

I'm a bit unclear why `split_edgeql` ever needs to return `None` though? Is this necessary for anything?

elprans:
That seems like a bug to me

edb/repl/__init__.py
```diff
@@ -71,7 +73,17 @@ def is_multiline():
         return False
 
     if text.endswith(';'):
-        return not lexutils.split_edgeql(text)
+        try:
+            return lexutils.split_edgeql(text) is None
```
elprans: I find this interface to be quite convoluted.  Why can't this logic be put _inside_ the `split_edgeql()` function and have it return both the list of statements _and_ whether it thinks the last statement has been terminated?  Perhaps as a `(complete, incomplete)` tuple.  That way the repl can execute the `complete` sequence, and populate the prompt with the text from `incomplete`:

```
db> SELECT 1; SELECT 2 <alt+enter>
{1}
db>SELECT 2█
```

edb/repl/__init__.py
```diff
@@ -71,7 +73,17 @@ def is_multiline():
...
+        except lexer.UnterminatedStringError:
+            # unterminated string is definitely an indication of a
+            # multiline input
+            return True
+        except base_lexer.LexError:
+            # it's OK to have lexing errors at this stage, since this
```
elprans: Could we have a mode where unknown tokens are simply ignored?

edb/repl/__init__.py
elprans: Also, this approach removes the need for the `script_mode` argument, since the decision whether to execute the `incomplete` part would fall onto the caller.

PR #445, opened 2019-02-03
Implement std::IN as a derivative of std::=

elprans:
The membership operators, `std::IN` and `std::NOT IN` must be applicable
to the same domain as `std::=`, since `[NOT] IN` is basically an
element-wise application of `=`.  Blindly copying all the `=`
definitions would be tedious now and in the future, so introduce a
mechanism to declare an operator as a "derivative" of another operator,
which would tell the compiler to use the origin operator when doing
call resolution.  For simplicity, derivative operators cannot be
polymorphic.

Also fixes an issue with tuple inputs to `IN`, where the right operand was
erroneously passed as a TupleVar instead of a value.

### DIFF
--- a/edb/edgeql/compiler/func.py
+++ b/edb/edgeql/compiler/func.py
@@ -236,6 +236,32 @@ def compile_operator(
 
         args.append((arg_type, arg_ir))
 
+    # Check if the operator is a derived operator, and if so,
+    # find the origins.
+    origin_op = opers[0].get_derivative_of(env.schema)
+    if origin_op is not None:
+        # If this is a derived operator, there should be
+        # exactly one form of it.  This is enforced at the DDL
+        # level, but check again to be sure.
+        if len(opers) > 1:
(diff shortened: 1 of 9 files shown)


elprans:
Removed.

edb/lib/std/25-setoperators.eql
```diff
@@ -21,34 +21,23 @@
...
-
-CREATE INFIX OPERATOR
-std::`NOT IN` (e: decimal, s: SET OF float64) -> std::bool
-    FROM SQL EXPRESSION;
-
-
-CREATE INFIX OPERATOR
-std::`NOT IN` (e: float64, s: SET OF decimal) -> std::bool
-    FROM SQL EXPRESSION;
+    SET derivative_of := 'std::=';
```
vpetrovykh: Is there a reason why `NOT IN` is derived from `=` as opposed to `!=`? It seems slightly odd.

PR #447, opened 2019-02-05
docs: Add details about `schema::Database`.

vpetrovykh:


docs/datamodel/databases.rst
```diff
@@ -3,3 +3,14 @@
 =========
 Databases
 =========
+
+An EdgeDB deployment can have multiple databases in it.
```
elprans: The correct term would be *cluster*

docs/datamodel/databases.rst
```diff
@@ -3,3 +3,14 @@
...
+:eql:stmt:`CREATE DATABASE` EdgeQL command adds a new database to the
```
elprans: *The*

docs/datamodel/databases.rst
```diff
@@ -3,3 +3,14 @@
...
+EdgeDB cluster. This also creates a special ``schema::Database``
```
elprans: This is oddly specific.  In fact, no special objects are created, since `schema::Database` is a view into a system catalog (which is also not a real table).  Simply say that it is possible to get a list of all databases with an example.

docs/datamodel/databases.rst
vpetrovykh: OK

docs/datamodel/databases.rst
vpetrovykh: OK

docs/datamodel/databases.rst
vpetrovykh: I've removed all the "object" talk and left just the sample query.

docs/datamodel/databases.rst
```diff
@@ -3,3 +3,14 @@
 =========
 Databases
 =========
+
+An EdgeDB cluster can have multiple databases in it. The
+:eql:stmt:`CREATE DATABASE` EdgeQL command adds a new database to the
+EdgeDB cluster.
+
+The following command will get a list of all databases currently
```
elprans: *The following command will get a list of all databases present in the cluster:*

vpetrovykh: ok

PR #448, opened 2019-02-07
Make type operation use consistent.

vpetrovykh:
The way link targets were specified in eschema was different from DDL
for multiple type links.

DDL and EdgeQL operator `IS` make use of type union operator `|` instead
of a `,` to specify multiple types. This is now the case for eschema,
too.

Union types can now be specified as link targets in eschema and in DDL.
They can also be used as the RHS of the `IS` operator.

Union types do not yet implement the "common" property/link logic.

Documentation has been updated.

### DIFF
--- a/docs/edgeql/operators/type.rst
+++ b/docs/edgeql/operators/type.rst
@@ -4,8 +4,40 @@
 Type Expressions
 ================
 
-This section describes EdgeQL expressions related to *types*.
+This section describes EdgeQL expressions related to *types*. Some of
+these expressions produce :ref:`introspection schema objects
+<ref_datamodel_modules_schema>`.
 
+Consider the following schema describing a few types used in an image
+creation and critiquing system. Some of the code examples in this
+section will refer to these types:
+
+.. code-block:: eschema
+
+    type Named:
+        required property name -> str
+
+    type User extending Named
+
+    type AdminUser extending User
+
+    abstract type Text:
+        required property body -> str
+
+    abstract type Authored:
+        required link author -> User:
+            constraint exclusive
+
+    type Comment extending Text, Authored:
+        property rating -> int64
+
+    type Image extending Authored, Named:
+        required property data -> bytes
+
+    type Gallery:
+        # something hand-picked by the resource administrators
+        required link curator -> AdminUser
+        required multi link content -> Authored
 
 IS
 ==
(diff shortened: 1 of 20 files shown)


docs/edgeql/operators/type.rst
```diff
@@ -35,8 +35,54 @@ IS
...
+.. eql:operator:: TYPEUNION: A | B
+
+    :optype A: object type
+    :optype B: object type
+    :resulttype: union type
+    :index: union type
+
+    Create a *union type* from types ``A`` and ``B``.
+
+    A *union type* is a way to describe the type of a set resulting from a
```
elprans: It's the other way around. The type of the result of a `UNION` expression is a union type, but we don't define union types through a `UNION`.  It is also not exclusive to the object types. There's a better definition in #219.

docs/edgeql/operators/type.rst
```diff
@@ -35,8 +35,54 @@ IS
...
+    :eql:op:`UNION` of two or more unrelated object types. This
+    expression can only be used on the right-hand-side of
+    :eql:op:`IS` or as the :ref:`concrete link
+    <ref_datamodel_links_concrete>` target type.
+
+    Consider the *union type* ``Text | Named``. Every ``User`` is of
+    that *union type* (because ``User`` extends ``Named``):
```
elprans: I think the emphasis is only necessary in the first occurrence of the term.

edb/edgeql/compiler/setgen.py
```diff
@@ -410,13 +410,13 @@ def resolve_ptr(
                 msg = (f'{near_endpoint.get_displayname(ctx.env.schema)} '
                        f'has no property {pointer_name!r}')
                 if target:
-                    msg += f'of type {target.get_name(ctx.env.schema)!r}'
+                    msg += f' of type {target.get_name(ctx.env.schema)!r}'
```
elprans: `get_displayname` should be used here.

edb/edgeql/compiler/setgen.py
```diff
@@ -410,13 +410,13 @@ def resolve_ptr(
...
 
             elif direction == s_pointers.PointerDirection.Outbound:
                 msg = (f'{near_endpoint.get_displayname(ctx.env.schema)} '
                        f'has no link or property {pointer_name!r}')
                 if target:
-                    msg += f'of type {target.get_name(ctx.env.schema)!r}'
+                    msg += f' of type {target.get_name(ctx.env.schema)!r}'
```
elprans: ditto.

edb/eschema/parser/grammar/declarations.py
```diff
@@ -132,6 +132,16 @@ def reduce_RowRawString(self, *kids):
         self.val = kids[0].parse_as_typelist()
 
 
+class TypeExpr(Nonterm):
```
elprans: Is the shadowing of the previous definition of the `TypeExpr` production intentional?

edb/schema/links.py
```diff
@@ -324,15 +333,20 @@ def _cmd_tree_from_ast(cls, schema, astnode, context):
...
+                else:
+                    target_typerefs = [target]
+
+                for typeref in target_typerefs:
+                    target_type = utils.resolve_typeref(typeref, schema=schema)
+
+                    if not isinstance(target_type, s_objtypes.ObjectType):
+                        raise errors.InvalidLinkTargetError(
+                            f'invalid link target, expected object type, got '
+                            f'{target_type.__class__.__name__}',
```
elprans: `get_displayname`

edb/eschema/parser/grammar/declarations.py
vpetrovykh: This is the only `TypeExpr` in eschema grammar.

elprans: Nevermind, this is a different grammar.

docs/edgeql/operators/type.rst
vpetrovykh: I'll change it to say that the type of the result of a `UNION` of two objects is an *example* of a union type.

Although #219 has indeed a more complete description that's also the problem. Currently, we don't have most of that implemented, so in the spirit of describing what actually is, I described the specific usages that are implemented and valid: link targets, `UNION` results and `IS`. These features were present even before the syntax change and generalization of type operations.

Type intersection is not implemented at all. Neither are type unions of scalar types.

docs/edgeql/operators/type.rst
vpetrovykh: I wanted to emphasize it consistently in this section.

edb/edgeql/compiler/setgen.py
vpetrovykh: ok

edb/edgeql/compiler/setgen.py
vpetrovykh: ok

edb/schema/links.py
vpetrovykh: ok

vpetrovykh:
I've added more type operators to the docs and a little tweak to `__type__`.

At this point the docs cover the `TYPEOF`, `INTROSPECT`, and `|` (type union) as they are implemented. The relationship between these type operators and link targets, `__type__` and `IS` is also covered.

docs/edgeql/operators/type.rst
elprans: Unions scalar types _are_ implemented, otherwise it wouldn't be possible to do a `UNION` over scalars.  A union of two scalar types is their common implicitly-castable-to type.

vpetrovykh:
Hmm... now that I look at it, we have a lot of examples referring to some schema without that schema being present anywhere in an obvious way. For example, look into "paths" or "statements".

I'm at this point rather confused what the standard is supposed to be. Defining a schema on every page seems a bit excessive and we need something for our examples, preferably something with real data, too. Basically we can't really pretend that each code snippet is perfectly valid stand-alone thing.

PR #452, opened 2019-02-12
docs: Document link aliasing via shapes.

vpetrovykh:
When a link is aliased in a shape, the original link properties are
transferred to the alias.

### DIFF
--- a/docs/edgeql/expressions/shapes.rst
+++ b/docs/edgeql/expressions/shapes.rst
@@ -154,3 +154,47 @@ Cardinality is normally statically inferred from the query, so
 overruling this inference may only be done to *relax* the cardinality,
 so it is not valid to specify the ``single`` qualifier for a computable
 expression that may return multiple items.
+
+
+Link Properties
++++++++++++++++
+
+A query could use a shape to create an alias to a real link. In this
+case, the link properties on that link are preserved on the aliased
+link as well. Consider the following schema:
+
+.. code-block:: eschema
+
+    abstract link friends:
+        property since -> datetime
+
+    type User:
+        required property name -> str
+        multi link friends -> User
+
+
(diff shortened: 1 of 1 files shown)


docs/edgeql/expressions/shapes.rst
```diff
@@ -154,3 +154,44 @@ Cardinality is normally statically inferred from the query, so
...
+        property since -> datetime
+
+    type User:
+        required property name -> str
+        multi link friends -> User
+
+Suppose that for a certain special subset of these users some
+interface requires that instead of the link ``friends`` the link
+should be called ``associates``, but the link property ``since`` must
+remain untouched. This can be done by using a shape with a computable
```
elprans: > link property ``since`` must remain untouched

I'm not sure what you mean here.  `SpecialUser.associates` is derived from `User.friends', so it inherits all link properties automatically.  Now, what you've written simply tells EdgeDB to include that property in the output.

I would build this description more specifically around the notion of aliasing links (that when you do `foo := Spam.bar`, `foo` inherits from `bar`).

docs/edgeql/expressions/shapes.rst
```diff
@@ -154,3 +154,44 @@ Cardinality is normally statically inferred from the query, so
...
+.. code-block:: eschema
+
+    abstract link friends:
+        property since -> datetime
+
+    type User:
+        required property name -> str
+        multi link friends -> User
+
+Suppose that for a certain special subset of these users some
```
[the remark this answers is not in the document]
vpetrovykh: I've reworded the whole section.

docs/edgeql/expressions/shapes.rst
vpetrovykh: I've reworded the whole section.

PR #453, opened 2019-02-14
docs: Minor tweak to attribute docs.

vpetrovykh:
Clarify that attributes can only be of type `str`.

Make the use of quotes consistent in text and examples for attribute
values.

### DIFF
--- a/docs/edgeql/ddl/attributes.rst
+++ b/docs/edgeql/ddl/attributes.rst
@@ -118,7 +118,7 @@ Create an object type ``User`` and set its ``title`` attribute to
 .. code-block:: edgeql
 
     CREATE TYPE User {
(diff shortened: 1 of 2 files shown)


docs/edgeql/ddl/attributes.rst
```diff
@@ -118,7 +118,7 @@ Create an object type ``User`` and set its ``title`` attribute to
 .. code-block:: edgeql
 
     CREATE TYPE User {
-        SET ATTRIBUTE title := 'User type';
+        SET ATTRIBUTE title := "User type";
```
1st1: Why this change?

vpetrovykh: To be consistent with the text above the example.

PR #454, opened 2019-02-14
docs: Fix the `code-block` spec for graphql schema.

vpetrovykh:
The language used to specify the graphql schema is different from
graphql itself. However, there are enough similarities that a graphql
lexer can be used for highlighting purposes.

### DIFF
--- a/tests/test_docs.py
+++ b/tests/test_docs.py
@@ -162,6 +162,10 @@ class TestDocSnippets(unittest.TestCase):
                 pass
             elif block.lang == 'graphql':
                 graphql_parser.parse(block.code)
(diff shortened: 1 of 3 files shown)


tests/test_docs.py
```diff
@@ -162,6 +162,10 @@ def run_block_test(self, block):
                 pass
             elif block.lang == 'graphql':
                 graphql_parser.parse(block.code)
+            elif block.lang == 'graphql-schema':
+                # The graphql-schema can be highlighter using graphql
```
1st1: can be highlighter -> can be highlighted

vpetrovykh: Fixed

PR #456, opened 2019-02-15
Add a 'gen-prism-grammars' command to 'edb'.

vpetrovykh:
This command generates the prism.js grammars for edgeql and eschema as
well as the synopsis versions of these. The grammars are based on the
keywords as defined in the respective parsers and the introspection of
the schema of an actual EdgeDB database (for purposes of establishing
operators, builtin types and functions).


1st1:
Actually I'd like this script to be in the edgedb-site repo.

1st1:
And it doesn't have to be in the `edb` group of tools.

vpetrovykh:
Cancelling pull request since the code will be in the edgedb-site repo.

PR #469, opened 2019-02-23
edgeql: Augment DELETE syntax.

vpetrovykh:
Make `DELETE ...` a sugar for `DELETE (SELECT ...)`i as this happens to
be the only sensible use of DELETE. This means that all the same
clauses as for SELECT are available for DELETE.

Update the documentation for DELETE.

### DIFF
--- a/docs/edgeql/statements/delete.rst
+++ b/docs/edgeql/statements/delete.rst
@@ -10,18 +10,33 @@ DELETE
 
 .. eql:synopsis::
 
-    [ WITH <with-spec> [ , ... ] ]
-    DELETE <expr> ;
+    [ WITH <with-item> [, ...] ]
+
+    DELETE <expr>
+
+    [ FILTER <filter-expr> ]
+
+    [ ORDER BY <order-expr> [direction] [THEN ...] ]
+
+    [ OFFSET <offset-expr> ]
+
+    [ LIMIT  <limit-expr> ] ;
 
 :eql:synopsis:`WITH`
     Alias declarations.
 
     The ``WITH`` clause allows specifying module aliases as well
-    as expression aliases that can be referenced by the ``UPDATE``
+    as expression aliases that can be referenced by the ``DELETE``
     statement.  See :ref:`ref_eql_with` for more information.
 
-:eql:synopsis:`DELETE <expr>`
-    Remove objects returned by *expr* from the database.
+:eql:synopsis:`DELETE ...`
+    The entire :eql:synopsis:`DELETE ...` statement is syntactic
+    sugar for ``DELETE (SELECT ...)``. Therefore, the base
+    :eql:synopsis:`<expr>` and the following :eql:synopsis:`FILTER`,
+    :eql:synopsis:`ORDER BY`, :eql:synopsis:`OFFSET`, and
+    :eql:synopsis:`LIMIT` clauses shape the set to
+    be deleted the same way an explicit :ref:`SELECT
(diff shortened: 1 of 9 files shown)


docs/edgeql/statements/delete.rst
```diff
@@ -21,7 +30,53 @@ DELETE
...
 :eql:synopsis:`DELETE <expr>`
-    Remove objects returned by *expr* from the database.
+    Form the base of the set to be deleted. This base set can be
+    further trimmed with other clauses.
+
+:eql:synopsis:`FILTER <filter-expr>`
+    The optional ``FILTER`` clause, where :eql:synopsis:`<filter-expr>`
+    is any expression that has a result of type :eql:type:`bool`. The
+    condition is evaluated for every element in the base set produced
+    by the ``DELETE`` clause.  The result of the evaluation of the
```
1st1: > produced
     by the ``DELETE`` clause

This is ambiguous. 

I guess you've just copy/pasted docs from the SELECT statement. We don't want that kind of duplication. I suggest to do it another way: copy/paste the synopsis and simply refer to the SELECT statement for the explanation of how different clauses work.  First paragraph should mention that `DELETE ...` is a sugar for `DELETE (SELECT ...)`; that should be enough.

edb/edgeql/ast.py
```diff
@@ -461,7 +461,13 @@ class UpdateQuery(SubjStatement):
 
 
 class DeleteQuery(SubjStatement):
+    # The extra fields are the same as for SELECT, because
+    # "DELETE ..." is a sugar for "DELETE (SELECT ...)"
```
1st1: Make a common parent class for SelectQuery and DeleteQuery

docs/edgeql/statements/delete.rst
vpetrovykh: I've reworded that section.

edb/edgeql/ast.py
vpetrovykh: Added a common parent.

docs/edgeql/statements/delete.rst
```diff
@@ -10,18 +10,33 @@ DELETE
...
     Alias declarations.
 
     The ``WITH`` clause allows specifying module aliases as well
-    as expression aliases that can be referenced by the ``UPDATE``
+    as expression aliases that can be referenced by the ``DELETE``
     statement.  See :ref:`ref_eql_with` for more information.
 
 :eql:synopsis:`DELETE <expr>`
-    Remove objects returned by *expr* from the database.
+    In principle, the entire ``DELETE ...`` statement is syntactic
```
1st1:
```suggestion
    The :eql:synopsis:`DELETE <expr>` statement is syntactic sugar for 
```

edb/edgeql/ast.py
```diff
@@ -433,14 +433,18 @@ class ReturningStatement(Statement):
     result_alias: str
 
 
-class SelectQuery(ReturningStatement):
+class AllFiltersStatement(Statement):
```
elprans: I'd prefer this to be called `SelectClauseStatement` or something like that.

docs/edgeql/statements/delete.rst
vpetrovykh: OK, I'll update the synopsis section to `DELETE ...` instead.

edb/edgeql/ast.py
vpetrovykh: OK

PR #470, opened 2019-02-25
GraphiQL, command statuses, repl fixes

1st1:


ansarizafar:
I would suggest using [Graphql Playground](https://github.com/prisma/graphql-playground) in place of GraphiQL. Graphql playground has better UI/features and It is also used by other more popular graphql platforms like Apollo and Prisma.

1st1:
@ansarizafar Thank you for the suggestion! It does look like GraphQL Playground is a bit more advanced than GraphiQL. However, when compared toe to toe, they aren't that much different.  What I like about GraphiQL is that its UI has less bells and whistles and thus should be more straightforward for new users, and it has an easy option to embed our logo.

GraphiQL or not, it's important to keep in mind that this is only a temporary solution. At some point we plan to have a web UI for EdgeDB that will have a GraphQL IDE of some sorts integrated in it.

PR #474, opened 2019-02-26
Implement EdgeQL over HTTP port

1st1:


edb/server/server.py
```diff
@@ -88,7 +91,15 @@ def _get_pgaddr(self):
                          portconf)
             return
 
-        port = http_graphql_port.HttpGraphQLPort(
+        if portconf.protocol == 'http+graphql':
```
elprans: The scheme should really be the other way around: `graphql+http`.

PR #478, opened 2019-02-27
Rename max* and min* constraints for consistency

1st1:
Rename the following constraints:

* minexclusive -> min_ex
* maxexclusive -> max_ex
* minlength -> min_len
* maxlength -> max_len

The initial motivation was to rename just the {min|max}exclusive
constraints, because their name mentally clashes with std::exclusive.

{min|max}length were the also renamed to play nice with {min|max}_ex
names, as well as with the std::len function.

### DIFF
--- a/docs/datamodel/constraints.rst
+++ b/docs/datamodel/constraints.rst
@@ -169,7 +169,7 @@ The standard library defines the following constraints:
         scalar type max_100 extending int64:
             constraint max(100)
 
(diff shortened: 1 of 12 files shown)


1st1:
@elprans Please feel free to hit the merge button.

1st1:
I'd also rename the std::length constraint to something that isn't so similar to `std::len` (and yet a completely different thing).  That can be done in a separate PR though.

PR #483, opened 2019-03-08
Automatically use prepared statements for queries over HTTP

1st1:


edb/server/compiler/compiler.py
```diff
@@ -136,6 +136,13 @@ def __init__(self, connect_args: dict, data_dir: str):
             self._data_dir = None
             self._std_schema = None
 
+    def _hash_sql(self, sql: bytes, **kwargs: bytes):
+        h = hashlib.sha1(sql)
+        for param, val in kwargs.items():
+            h.update(param.encode('latin1'))
```
elprans: Are `kwargs` the EdgeQL query argument dict? We use `\w+` for `IDENT`, which matches any Unicode word character, so we should probably use `utf8` here.

1st1: No, `kwargs` is an additional set of qualifiers for the compiled query, like its output mode.

PR #490, opened 2019-03-12
Add a 'gen-meta-grammars' command to 'edb'.

vpetrovykh:
This command generates the meta information useful for creation of
various grammars and highlighters. In particular, this information is
used by the pygments grammars used in REPL.

The command `make pygments` will regenerate an up-to-date grammar based
on the current state of EdgeDB.


Makefile
```diff
@@ -15,3 +15,8 @@ docs:
 
 postgres:
 	python setup.py build_postgres
+
+
+pygments:
+	edb gen-meta-grammars edgeql > edb/edgeql/pygments/meta.py
+	edb gen-meta-grammars > edb/eschema/pygments/meta.py
```
1st1: is eschema default? why?  There should be no default language for 'gen-meta-grammars'

edb/tools/gen_meta_grammars.py
```diff
@@ -0,0 +1,169 @@
...
+def main(names):
+    code = [
+        f'# AUTOGENERATED BY EdgeDB WITH\n'
+        f'#     $ edb gen-pygments-grammars {" ".join(names)}\n'
+        f'#\n'
+        f'# TIMESTAMP: {datetime.datetime.now():%Y-%m-%d %H:%M:%S}\n'
+        f'\n\n'
+    ]
+
+    con = edgedb.connect(user='edgedb', database='edgedb')
```
1st1: `con` should be created in `gen_meta_grammars()`; `main()` should be called in `try..finally` with `con.close()` in the `finally` block

Makefile
1st1: ah, I see, there's a default "all languages" -- no need for that default. One should always specify what they need (otherwise what happens if we add 3 new languages and the tool starts to generate shit that nobody expected it to generate)

vpetrovykh: Default is *everything*. Currently it's edgeql + eschema, but in the future this argument will be irrelevant since we'll just have one single grammar, most likely.

edb/tools/gen_meta_grammars.py
vpetrovykh: OK, I'll update it.

Makefile
1st1: just drop the default; always require at least one language.

vpetrovykh: OK, fair point about explicit specification of the languages..

Issue #498, opened 2019-03-13
Change "eschema" grammar


vpetrovykh:
The language used to defined *eschema* is currently very similar to DDL and very different (case-sensitive, whitesapce-sensitive). Which means that in principle it's a whole new language that needs to be documented (by us) and learned (by the users). This is a non-trivial burden. There's also some confusion as to which is primary source of truth - DDL or eschema?

The new approach is to make the Schema Definition Language (SDL) to be much more closely related to DDL. The rule of thumb is to make it so that if you take a `CREATE ...` statement from DDL and strip out all the `CREATE` and `SET` keywords you should end up with a valid SDL declaration. The goal of DDL is to specify how to change the DB from one state to the next, the goal of SDL is to describe the final state. This is why SDL just mirrors `CREATE` and `SET` DDL since we're not interested in the exact details of the steps.

This is related to #58

PR #500, opened 2019-03-14
eschema: Convert eschema grammar to SDL closer to DDL.

vpetrovykh:
The language used to defined eschema is currently very similar to DDL
and very different (case-sensitive, whitesapce-sensitive). Which means
that in principle it's a whole new language that needs to be documented
(by us) and learned (by the users). This is a non-trivial burden.
There's also some confusion as to which is primary source of truth - DDL
or eschema?

The new approach is to make the Schema Definition Language (SDL) to be
much more closely related to DDL. The rule of thumb is to make it so
that if you take a CREATE ... statement from DDL and strip out all
the CREATE and SET keywords you should end up with a valid SDL
declaration. The goal of DDL is to specify how to change the DB from
one state to the next, the goal of SDL is to describe the final state.
This is why SDL just mirrors CREATE and SET DDL since we're not
interested in the exact details of the steps.

The grammar has been stripped and moved to the EdgeQL grammar tree, but
the package is still called "eschema" and has a separate codegen and ast.

The command `edb gen-meta-grammars` now only has one valid argument
`edgeql`. However, we still keep two pygments grammars using it
(potentially useful if there's different quirks that need to be
accounted for).

Drop the parentheses from the `EXTENDING` list in both SDL and DDL.

Issue #498.


Makefile
```diff
@@ -20,5 +20,5 @@ postgres:
 pygments:
 	out=$$(edb gen-meta-grammars edgeql) && \
 		echo "$$out" > edb/edgeql/pygments/meta.py
-	out=$$(edb gen-meta-grammars edgeql eschema) && \
+	out=$$(edb gen-meta-grammars edgeql) && \
```
1st1: We don't need eschema/pygments anymore

edb/eschema/codegen.py
```diff
@@ -61,28 +63,29 @@ def _visit_specs(self, node):
...
             if getattr(node, 'properties', None):
-                self._visit_list(node.properties)
+                self.visit_list(node.properties, terminator=';')
             if getattr(node, 'attributes', None):
-                self._visit_list(node.attributes)
+                self.visit_list(node.attributes, terminator=';')
             if getattr(node, 'constraints', None):
-                self._visit_list(node.constraints)
+                self.visit_list(node.constraints, terminator=';')
             if getattr(node, 'policies', None):
```
elprans: There are no 'policies' anymore.

edb/edgeql/parser/grammar/sdl.py
```diff
@@ -0,0 +1,975 @@
...
+
+    def _block2(self, lbrace, sc1, cmdlist, sc2, rbrace):
+        self.val = cmdlist.val
+
+
+def commands_block(parent, *commands, opt=True):
+    if parent is None:
+        parent = ''
+
+    clsdict = collections.OrderedDict()
```
1st1: No need to use OrderedDict since Python 3.6 (except for LRU caches)

edb/edgeql/pygments/meta.py
```diff
@@ -1,7 +1,7 @@
 # AUTOGENERATED BY EdgeDB WITH
 #     $ edb gen-pygments-grammars edgeql
 #
-# TIMESTAMP: 2019-03-12 19:46:27
+# TIMESTAMP: 2019-03-14 13:13:13
```
1st1: No need for this change. In fact, I think this was a bad idea of mine to include the TIMESTAMP at all.  We can always use git to see when the file was last updated.  So please drop the TIMESTAMP comment generation from `edg gen-meta-grammars`

Makefile
vpetrovykh: Pygments dropped from code and make target.

edb/eschema/codegen.py
vpetrovykh: Dropped policies.

edb/edgeql/pygments/meta.py
vpetrovykh: Done

edb/edgeql/parser/grammar/sdl.py
vpetrovykh: Dropped.

PR #509, opened 2019-03-19
Update standard library builtin types.

vpetrovykh:
Add Named and UniquelyNamed as abstract types to std.

### DIFF
--- a/edb/edgeql/pygments/__init__.py
+++ b/edb/edgeql/pygments/__init__.py
@@ -25,8 +25,13 @@ from . import meta
 
 __all__ = ['EdgeQLLexer']
 
+# some unreserved keywords have special handling because they clash
+# with builtins
+EXCEPTION_KW = {'named'}
 
-unreserved_keywords = meta.EdgeQL.unreserved_keywords
+
+unreserved_keywords = sorted(
+    set(meta.EdgeQL.unreserved_keywords) - EXCEPTION_KW)
 reserved_keywords = meta.EdgeQL.reserved_keywords
 builtins = sorted(set(
     meta.EdgeQL.type_builtins + meta.EdgeQL.constraint_builtins +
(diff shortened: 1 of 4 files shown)


edb/edgeql/pygments/__init__.py
```diff
@@ -25,8 +25,13 @@
 
 __all__ = ['EdgeQLLexer']
 
+# some unreserved keywords have special handling because they clash
+# with builtins
+EXCEPTION_KW = {'named'}
```
1st1: Don't hardcode this; names like this should be available via `meta.py`

tests/test_edgeql_ddl.py
```diff
@@ -1968,6 +1968,29 @@ class TestEdgeQLDDL(tb.DDLTestCase):
...
             """)
 
+    @test.xfail('''
+        UniquelyNamed is an abstract type defined in std. Extending
+        from it causes the following error:
+
+        SchemaError: Property 'test::test|name@@test|Status' is
+        already present in the schema
+
+        This doesn't seem to happen when extending simply Named.
```
1st1: Yeah, add a test that Named works.

edb/edgeql/pygments/__init__.py
vpetrovykh: I'm not sure yet how to codify it properly. The basic idea is that unreserved keywords like this will generally have some special context, like another keyword before or after, or potentially some other symbol. Well, technically, *every* unreserved keyword has some special context, hence "unreserved", but I don't yet know what's the best way to capture it.
Maybe I could do some sort of tuple like `(before, kw, after)`, so for named it would be `(before=None, kw='named', after='only')`.

tests/test_edgeql_ddl.py
vpetrovykh: Added a "Named" test.

edb/edgeql/pygments/__init__.py
vpetrovykh: The bottom line is that there's a more general problem that needs to be solved - how to handle unreserved keywords. That can be addressed in a separate PR, but for the moment the hardcode at least indicates that something special is being done to what otherwise would be a keyword.

vpetrovykh:
OK, I'll drop this PR and instead create a PR with tests reproducing the bug that this PR uncovered, since it's not module-dependent.

PR #515, opened 2019-03-21
Merge edb.eschema into edb.edgeql.

vpetrovykh:


edb/edgeql/declarative.py
```diff
@@ -29,11 +29,10 @@
 from edb.edgeql import ast as qlast
 from edb.edgeql import codegen as qlcodegen
 from edb.edgeql import compiler as qlcompiler
+from edb.edgeql import parser as edge_parser
```
1st1: just `parser`

edb/edgeql/parser/__init__.py
```diff
@@ -17,12 +17,12 @@
 #
 
 
-from .parser import EdgeQLExpressionParser, EdgeQLBlockParser
+from . import parser as edge_parser
```
1st1: just `parser`

tests/test_docs.py
```diff
@@ -23,8 +23,7 @@
 
 from graphql.language import parser as graphql_parser
 
-from edb.edgeql import parser as edgeql_parser
-from edb.eschema import parser as schema_parser
+from edb.edgeql import parser as edge_parser
```
1st1: just `ql_parser`

edb/edgeql/declarative.py
vpetrovykh: ok

edb/edgeql/parser/__init__.py
vpetrovykh: ok

tests/test_docs.py
vpetrovykh: ok

PR #517, opened 2019-03-22
Implement initial support for user authentication

elprans:
This adds generic authentication framework via the new `cfg::Auth` 
configuration object.  The two initial authentication modes are `Trust` and
`SCRAM` (SCRAM-SHA-256).  The default is `SCRAM`, which means that a valid
role password is required to connect.   Since new instances obviously don't
have that configured, add a way to connect to a special
"admin" UNIX socket, which is protected by OS permissions, and which does
not enforce internal authentication.

To make things easier, the development server (which is now started with
`edb server`) has `Trust` auth configured by default.


edb/server/dbview/dbview.pyx
```diff
@@ -260,18 +261,43 @@ cdef class DatabaseConnectionView:
...
-    def get_system_overrides(self):
+    def get_sys_query(self, key: str) -> str:
+        return self._sys_queries[key]
+
+    async def reload_config(self):
+        conn = await self._server.new_pgcon(defines.EDGEDB_SUPERUSER_DB)
+
+        query = self.get_sys_query('config')
+        try:
+            result = await conn.simple_query(query.encode(), ignore_data=False)
```
1st1: `query.encode()` this should be cached

Neustradamus:
Good job!

Linked to:

PR #520, opened 2019-03-25
Add support for managing roles via a CLI

elprans:
This is the first command of a series of CLI commands that allow
basic manipulation of the EdgeDB instance and its schema using command
line syntax.  The basic notion is that `edgedb {create|alter|drop} ...`
follow a familiar DDL syntax, minus the unnecessary punctuation.

`edgedb {create,alter,drop} role` is chosen to be implemented as the first
such command to make it easier to set up a role from the command line
during initial EdgeDB setup.

### DIFF
--- a/edb/cli/mng.py
+++ b/edb/cli/mng.py
@@ -1,0 +1,182 @@
+#
+# This source file is part of the EdgeDB open source project.
+#
+# Copyright 2008-present MagicStack Inc. and the EdgeDB authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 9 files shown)


edb/cli/mng.py
```diff
@@ -0,0 +1,184 @@
...
+
+    return alters
+
+
+@create.command(name='role')
+@click.argument('role-name', type=str)
+@options(_role_options)
+@click.pass_context
+def create_role(ctx, role_name, **kwargs):
+    print('!!!', kwargs)
```
1st1: debug code

Issue #350, opened 2018-12-05
Cardinality validation for aggregates mapped onto SQL


vpetrovykh:
This is related to #349 

For standard library functions there's a need to map them onto existing SQL functions. Sometimes aggregate functions in SQL can return `NULL` for cases that in EdgeDB we would consider invalid. Consider `stddev` (sample standard deviation) that is not defined for sets of cardinality < 2. On the other hand for all other sets it is a single number. So in our definition of cardinality we expect the return value to be of cardinality exactly 1, neither more nor less (much like `count`). Unfortunately Postgres will return `NULL` for the input sets that are too small and we need a way of handling that with some sort of graceful error message instead.

The proposal is to have a special attribute like `empty_result_error` that provides a way to customize the error message. While a `NULL` test of the result is only added if the function return type is declared to be of cardinality >= 1 (not `OPTIONAL` or `SET OF`).

1st1:
This can be closed now, right? @vpetrovykh

vpetrovykh:
The examples of `SELECT math::stddev({})` or `SELECT math::stddev({1})` should not return any value as they have no valid value, much like `1/0`. Sample standard deviation is not defined for sets of fewer than 2 elements.

The desired behavior would be for these types of functions to raise an error instead of silently producing invalid output. The error message needs to reflect the fact that the function is not defined for certain input.

elprans:
Fixed in #521

PR #521, opened 2019-03-25
Raise an exception on invalid input to mean(), var(), and stddev()

elprans:
`math::mean()` is not defined on empty input set, `var()` and `stddev()`
are not defined on sets with cardinality less than 2.  Rather then
silently returning an empty set (which is a result of an underlying SQL
function returning NULL), raise an explicit exception.

This commit also adds a missing `InvalidValueError` mapping that fixes
the incorrect error being raised on parameter validation errors.  This
commit also wires in support for passing the query source context on
runtime errors, if available.


1st1:
Fixes https://github.com/edgedb/edgedb/issues/350

PR #523, opened 2019-03-27
Add support for collection types in schema

elprans:
Collection types are now supported much better in various schema contexts,
such as property types, and function parameter and return types.

Arrays of tuples are still not supported because casting record[] to a 
specific composite type array is not implemented in Postgres.

### DIFF
--- a/tests/test_edgeql_calls.py
+++ b/tests/test_edgeql_calls.py
@@ -1141,3 +1141,66 @@ class TestEdgeQLFuncCalls(tb.QueryTestCase):
                 [1, 2],
             ]
         )
+
+    async def test_edgeql_calls_33(self):
+        # Tuple argument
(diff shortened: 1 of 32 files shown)


tests/test_edgeql_calls.py
```diff
@@ -1141,3 +1141,65 @@ class TestEdgeQLFuncCalls(tb.QueryTestCase):
                 [1, 2],
             ]
         )
+
+    async def test_edgeql_calls_33(self):
+        # Tuple argument
+
+        await self.con.execute('''
+            CREATE FUNCTION test::call33(
+                a: tuple<int64, tuple<int64>>
```
1st1: Need to add tests for named tuples (as an argument and as a return value)

PR #531, opened 2019-03-28
Add initial support for enumerated types

elprans:
Enumerated types are a special class of a scalar type that are defined as a
list of explicit string values using the following syntax:

CREATE SCALAR TYPE <name> EXTENDING enum<'val1'[,...'valN']>

Over the wire, and in text serialization, the enum values are transferred 
as strings, but they sort according to the order of enum value definitions, 
and every enum type is distinct, i.e. there is no implicit cast between 
them.

Currently, there is no way to alter the composition of enumerated type 
after it was created.

### DIFF
--- a/docs/datamodel/scalars/enum.rst
+++ b/docs/datamodel/scalars/enum.rst
@@ -1,0 +1,27 @@
+.. _ref_datamodel_scalars_enum:
+
+Enumerated Types
+================
+
+.. eql:type:: std::enum
+
+    :index: enum
+
+    An enumerated type is a data type consisting of an order list of values.
+
+    An enumeration type can be declared in a schema declaration using
+    the following syntax:
+
+    .. eql:synopsis::
+
+        enum "<" <enum-values> ">"
+
+    Where :eql:synopsis:`<enum-values>` is a comma-separated list of
+    quoted string constants comprising the enum type.  Currently, the
+    only valid application of the enum declaration is to define an
+    enumerated scalar type:
+
+    .. code-block:: edgeql-repl
+
+        db> CREATE SCALAR TYPE my_enum_t EXTENDING enum<'One', 'Two'>;
+        CREATE TYPE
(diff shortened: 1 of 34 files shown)


docs/datamodel/scalars/enum.rst
```diff
@@ -0,0 +1,27 @@
...
+    :index: enum
+
+    An enumerated type is a data type consisting of an order list of values.
+
+    An enumeration type can be declared in a schema declaration using
+    the following syntax:
+
+    .. eql:synopsis::
+
+        enum "<" <enum-values> ">"
```
1st1: ```scalar type <name> extending enum "<" <enum-values> ">"```

docs/edgeql/ddl/scalars.rst
```diff
@@ -48,6 +48,10 @@ subtype with constraints.
 :eql:synopsis:`EXTENDING <supertype>`
     Optional clause specifying the *supertype* of the new type.
 
+    The :eql:synopsis:`<supertype>` is an
+    :eql:type:`enumerated type <std::enum>` declaration, in which
+    case an enumerated scalar type is defined.
```
1st1: Why are we saying that `<supertupe>` must be an enum declaration in a *generic* section about scalars?

1st1: The wording of this paragraph is weird. Rephrase to something like this: 

To define an enumerated scalar type specify :eql:type:`enumerated type <std::enum>` declaration in :eql:synopsis:`<supertype>` .

elprans: That sentence is missing a leading "if" and was intended like this:

"
    If the :eql:synopsis:`<supertype>` is an
    :eql:type:`enumerated type <std::enum>` declaration, an enumerated scalar type is defined.
"

docs/datamodel/scalars/enum.rst
elprans: This bit is specifically about the syntax for an anonymous enum type declaration.

PR #533, opened 2019-03-28
Stop acessing collection internals directly

elprans:
Lots of code is reaching directly into collection type attributes.  This
is bad and will break if the collection type is a schema object.  Fix
this, and make sure the accessor APIs are schema-aware.


edb/edgeql/compiler/cast.py
```diff
@@ -325,12 +327,12 @@ def _cast_tuple(
 
     # For tuple-to-tuple casts we generate a new tuple
     # to simplify things on sqlgen side.
-    new_names = list(new_stype.element_types)
+    new_subtypes_list = list(new_subtypes.items())
```
1st1: You can use `new_subtypes_list` on line 320 and avoid creating an extra object

PR #541, opened 2019-03-29
docs: Separate data model from syntax synopsis.

vpetrovykh:
Remove synopsis from Data Model section.

Create a new section in EdgeQL for type syntax and type expressions.
Move all the type-specific syntax into this section.

### DIFF
--- a/docs/datamodel/colltypes.rst
+++ b/docs/datamodel/colltypes.rst
@@ -7,6 +7,28 @@ Collection Types
 *Collection types* are special generic types used to group homogeneous or
 heterogeneous data.
 
+.. eql:type:: std::array
+
+    :index: array
+
+    Arrays represent a one-dimensional homogeneous ordered list.
+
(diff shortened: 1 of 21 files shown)


docs/datamodel/colltypes.rst
```diff
@@ -7,28 +7,40 @@ Collection Types
...
 
-    .. eql:synopsis::
+    .. code-block:: edgeql-repl
 
-        tuple "<" <element-type>, [<element-type>, ...] ">"
+        db> SELECT [1, 2];
+        {[1, 2]}
 
-    A named tuple:
+        db> SELECT (INTROSPECT TYPEOF [1, 2]).name;
```
1st1: doesn't work right now :( drop it

vpetrovykh: Removed the broken `INTROSPECT` from the docs.

PR #544, opened 2019-03-30
protocol and other improvements

1st1:


setup.py
```diff
@@ -63,7 +63,7 @@
     ],
 
     'docs': [
-        'Sphinx',
+        'Sphinx~=1.8.1',
```
elprans: Why not 2.0?

PR #545, opened 2019-03-30
Tighten the handling of array type dimension information and other fixes

elprans:


edb/schema/types.py
```diff
@@ -521,22 +528,9 @@ def from_subtypes(cls, schema, subtypes, typemods=None, *, name=None):
...
-        if isinstance(stype, cls):
-            # There is no array of arrays, only multi-dimensional arrays.
-            element_type = stype.get_element_type(schema)
-            if not dimensions:
-                dimensions.append(-1)
-            dimensions += stype.dimensions
-        else:
-            element_type = stype
-            dimensions = []
+        dimensions = [-1]
```
1st1: Please add a comment here, explaining that this means a 1 dimensional unbound array

PR #553, opened 2019-04-02
Implement `edgedb configure` CLI command

elprans:
The new `configure` CLI subcommand allows for reasonably complete control
over the system configuration and is largely equivalent to the
`CONFIGURE SYSTEM` statement.

The available subcommands mirror those of `CONFIGURE SYSTEM`:

    edgedb configure set <param> <value>
    edgedb configure reset <param>
    edgedb configure insert <cfgobj> --prop1=val1 --prop2=val2 ...
    edgedb configure reset <cfgobj> --prop1=val1 ...

### DIFF
--- a/edb/server/server.py
+++ b/edb/server/server.py
@@ -131,19 +131,32 @@ class Server:
         )
 
     async def _restart_mgmt_port(self, nethost, netport):
+        await self._mgmt_port.stop()
+
         try:
-            self._mgmt_port = self._new_port(
+            new_mgmt_port = self._new_port(
                 mng_port.ManagementPort,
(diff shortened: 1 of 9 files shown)


edb/server/server.py
```diff
@@ -131,19 +131,31 @@ def _new_port(self, portcls, **kwargs):
...
             await self._mgmt_port.start()
+            raise
+
+        try:
+            await new_mgmt_port.start()
         except Exception:
+            try:
+                await new_mgmt_port.stop()
+            except Exception:
+                pass
```
1st1: log this exception

PR #557, opened 2019-04-02
docs: Update the docs examples.

vpetrovykh:
Change the tutorial view example from DDL to SDL.

Use short path in UPDATE examples.


vpetrovykh:
We're keeping DDL for adding a view in the tutorial.

PR #573, opened 2019-04-03
Ignore IPv6 component of 'localhost'

elprans:
On many systems 'localhost' resolves to _both_ IPv4 and IPv6
addresses, even if the system is not capable or configured to
handle the IPv6 connections.  Due to the common nature of this
issue explicitly disable the AF_INET6 component of 'localhost'.

### DIFF
--- a/edb/server/mng_port/port.py
+++ b/edb/server/mng_port/port.py
@@ -96,9 +96,11 @@ class ManagementPort(baseport.Port):
     async def start(self):
         await super().start()
 
+        nethost = await self._fix_localhost(self._nethost, self._netport)
+
         tcp_srv = await self._loop.create_server(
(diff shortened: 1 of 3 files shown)


edb/server/mng_port/port.py
```diff
@@ -93,12 +95,56 @@ def new_edgecon_id(self):
...
+        infos = [a for a in localhost if a[0] == socket.AF_INET]
+
+        if not infos:
+            # "localhost" did not resolve to an IPv4 address,
+            # let create_server handle the situation.
+            return hosts
+
+        # Replace 'localhost' with explicitly resolved AF_INET addresses.
+        hosts.pop(idx)
+        for info in infos:
```
1st1: `in reversed(infos)`

PR #577, opened 2019-04-04
Fix repl script mode

elprans:
Unbreak the "script from stdin" mode: fix handling of password prompt,
as well as exception rendering and exit code.


tests/test_cli.py
```diff
@@ -72,3 +72,7 @@ class TestCLI(tb.ConnectedTestCase, tb.CLITestCaseMixin):
         result = self.run_cli('create', 'role', 'foo', '--allow-login',
                               '--password', input='foo-pass\n')
         self.assertIn('input is not a TTY', result.output)
+
+    async def test_cli_repl_script(self):
```
1st1: I'd also test that an edgeql error causes a non-zero return code.

PR #579, opened 2019-04-05
Switch to SERIALIZABLE as the default transaction isolation level

elprans:
Continuing the theme of safety and consistency by default, switch the
default transaction isolation level to `SERIALIZABLE`.  Remove the
`READ COMMITTED` level as it does not provide any tangible benefits over
the higher isolation levels.  `REPEATABLE READ` remains as an option for
scenarios that are pathological for the current implementation of
`SERIALIZABLE` in Postgres, as well as for use in hot-standbys.


edb/lib/sys.edgeql
```diff
@@ -155,3 +155,12 @@ sys::get_version_as_str() -> std::str
         ) IF v.stage != <sys::version_stage>'final' ELSE ''
     $$;
 };
+
+
+CREATE FUNCTION
+sys::get_transaction_isolation() -> std::str
```
1st1: do you want to make it an enum?

edb/server/compiler/errormech.py
```diff
@@ -52,6 +52,10 @@ class PGError(enum.Enum):
     InvalidDatetimeFormatError = '22007'
     DatetimeError = '22008'
 
+    TransactionSerializationFailure = '40001'
+    TransactionIntegrityConstraintViolation = '40002'
```
1st1: this one isn't mapped to anything

edb/lib/sys.edgeql
elprans: Not really.  It makes no difference output-wise, and it's not an argument

PR #589, opened 2019-04-05
docs: Cross-reference the formatting tables from functions.

vpetrovykh:
Regular expression functions and converter functions use special
formatting specs that are described separately. Now the functions
contain a link to the corresponding spec.

### DIFF
--- a/docs/edgeql/funcops/datetime.rst
+++ b/docs/edgeql/funcops/datetime.rst
@@ -450,8 +450,9 @@ Date and Time
 
     The :eql:type:`datetime` value can be parsed from the input
     :eql:type:`str` *s*. By default, the input is expected to conform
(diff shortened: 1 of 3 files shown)


docs/edgeql/funcops/datetime.rst
```diff
@@ -505,6 +503,8 @@ Date and Time
     value can be parsed from the input :eql:type:`str` *s* with an
     optional *fmt* argument or it can be given in terms of its
     component parts: *year*, *month*, *day*, *hour*, *min*, *sec*.
+    For more details on formatting see :ref:`here
```
1st1: Please make "For more details" a separate paragraph in all places

vpetrovykh: OK

PR #598, opened 2019-04-10
docs: Add more cross-references to datamodel.

vpetrovykh:


docs/datamodel/props.rst
```diff
@@ -34,3 +33,13 @@ source objects can link to the same target via
 out the number of property targets, a :ref:`collection type
 <ref_datamodel_collection_types>` target by itself is considered a
 *single* target.
+
+
+See Also
+--------
+
+Link
```
1st1: Is "link" supposed to be here? Or is it "property"?

docs/datamodel/scalars/bytes.rst
```diff
@@ -28,3 +28,12 @@ Bytes
...
         {false}
+
+
+See Also
+--------
+
+Scalar type
+:ref:`SDL <ref_eql_sdl_scalars>`,
+:ref:`DDL <ref_eql_ddl_scalars>`,
+and :ref:`introspection <ref_eql_introspection_scalar_types>`.
```
1st1: I'd also link relevant operators and functions from each data type page

docs/datamodel/props.rst
vpetrovykh: Fixed

docs/datamodel/scalars/bytes.rst
vpetrovykh: I've linked the sections from funcops.

PR #601, opened 2019-04-10
Add documentation for server installation and configuration

elprans:


docs/admin/setup.rst
```diff
@@ -0,0 +1,70 @@
...
+EdgeDB server.  The data managed by the instance is usually stored in
+a single directory, which is called the *data directory*.
+
+The EdgeDB server creates and populates a data directory automatically
+on the first run.  The location of the data directory can be specified
+using the ``-D`` option:
+
+.. code-block:: bash
+
+    $ edgedb-server -D /data/directory/path
```
vpetrovykh: Is there a difference between `edgedb-server` and `edb server`?

elprans: `edb` is the develoment-mode command, we don't install it in production builds.

PR #625, opened 2019-04-17
Separate static error handler from one that requires schema.

vpetrovykh:
Many errors don't require the schema to be properly hadled, these should
be handled first. If it is determined that an error needs some schema
data to be resolved, only then the schema should be fetched.

### DIFF
--- a/edb/server/compiler/errormech.py
+++ b/edb/server/compiler/errormech.py
@@ -20,6 +20,7 @@
 import enum
 import json
 import re
+import typing
 import uuid
 
 from edb import errors
@@ -31,7 +32,7 @@ from edb.pgsql import common
 from edb.pgsql import types
 
 
-class PGError(enum.Enum):
+class PGErrorCode(enum.Enum):
 
     IntegrityConstraintViolationError = '23000'
     RestrictViolationError = '23001'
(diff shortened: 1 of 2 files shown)


edb/server/compiler/errormech.py
```diff
@@ -252,3 +270,72 @@ def interpret_backend_error(schema, fields):
...
+
+    elif code in constraint_errors:
+        source = None
+
+        for errtype, ere in constraint_res.items():
+            m = ere.match(message)
+            if m:
+                error_type = errtype
+                break
+        # no need for else clause since it would have been handled by
```
elprans: You have a potential `NameError` for `error_type` here.

vpetrovykh: OK, I've added an initialization for `error_type` just in case. Incidentally I removed source from this code block since it's not used (but oddly wasn't flagged as unused by flake).

1st1:
Please don't merge this without my review.

edb/server/compiler/errormech.py
```diff
@@ -31,6 +32,10 @@
 from edb.pgsql import types
 
 
+class SchemaNeededForError:
+    '''A sentinel used to signal that a particular error requires a schema.'''
+
+
 class PGError(enum.Enum):
```
1st1: While you're here, could you please rename it to PGErrorCode?

edb/server/compiler/errormech.py
```diff
@@ -139,24 +155,29 @@ def interpret_backend_error(schema, fields):
     column_name = fields.get('c')
     constraint_name = fields.get('n')
 
-    if code == PGError.NotNullViolationError:
-        source_name = pointer_name = None
-
-        if schema_name and table_name:
-            tabname = (schema_name, table_name)
+    return ErrorDetails(
```
1st1: I don't like functions that have polymorphic return types.  I this case, it's either `ErrorDetails` or an actual exception instance.  I suggest to have one function to build ErrorDetails, and another that acts on it.

edb/server/compiler/errormech.py
```diff
@@ -139,24 +155,29 @@ def interpret_backend_error(schema, fields):
...
 
-        if pointer_name is not None:
-            pname = f'{source_name}.{pointer_name}'
+    message = err_details.message
+    detail = err_details.detail
+    detail_json = err_details.detail_json
+    code = err_details.code
+    schema_name = err_details.schema_name
+    table_name = err_details.table_name
+    constraint_name = err_details.constraint_name
```
1st1: Just use `err_details` in the below code.  This packing/unpacking looks suspicious.

edb/server/compiler/errormech.py
```diff
@@ -230,10 +248,10 @@ def interpret_backend_error(schema, fields):
             message, details=detail if detail else None)
 
     elif code == PGError.InvalidTextRepresentation:
```
1st1: Make a frozenset of codes that require schema. Have just one `if code in SCHEMA_CODES` clause.

edb/server/compiler/errormech.py
```diff
@@ -31,6 +32,10 @@
 from edb.pgsql import types
 
 
+class SchemaNeededForError:
```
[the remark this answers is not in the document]
vpetrovykh: OK

edb/server/compiler/errormech.py
vpetrovykh: OK

edb/server/compiler/errormech.py
vpetrovykh: OK

edb/server/compiler/errormech.py
vpetrovykh: OK, I've added a handler for some of these error details.

edb/server/compiler/errormech.py
vpetrovykh: I've put the codes that unambiguously require schema in `SCHEMA_CODES`. There are some error codes (notably, constraint_errors) that don't always require the schema and which I didn't put into the `SCHEMA_CODES` set.

PR #627, opened 2019-04-19
edgeql: Add a "session_only" attribute to functions.

### LINKED ISSUE #406: Add "session_only" attribute to functions
Some functions like `sys::advisory_lock()` can't be executed via EdgeQL REST protocol; such functions will have their `session_only` attribute set to `true`.  The compiler should be modified to support two additional modes: "session" and "no session".

vpetrovykh:
If a function is marked as "session_only" it can't be executed over REST
protocol since that protocol doesn't provide any guarantees regarding
the session used.

Fixes #406.

### DIFF
--- a/docs/edgeql/ddl/functions.rst
+++ b/docs/edgeql/ddl/functions.rst
@@ -45,8 +45,9 @@ CREATE FUNCTION
 
     # where <subcommand> is one of
 
-      FROM <language> <functionbody>
+      SET session_only := {true | false}
       SET ANNOTATION <annotation-name> := <value>
+      FROM <language> <functionbody>
 
 
 Description
@@ -131,6 +132,16 @@ Subcommands
 
 ``CREATE FUNCTION`` allows specifying the following subcommands in its
 block:
+
+:eql:synopsis:`SET session_only := {true | false}`
+    If ``true``, the function is only valid in contexts where there is
+    a well-defined session. In particular, this function cannot be
+    used over an HTTP port, within the body of another
+    non-session-only function, as part of a view definition, or as a
+    default value in definitions. This field is ``false`` by default.
+    Examples of session-only functions: :eql:func:`sys::sleep`,
+    :eql:func:`sys::advisory_lock`, :eql:func:`sys::advisory_unlock`,
+    :eql:func:`sys::advisory_unlock_all`.
 
 :eql:synopsis:`SET ANNOTATION <annotation-name> := <value>`
     Set the function's :eql:synopsis:`<annotation-name>` to
--- a/docs/edgeql/sdl/functions.rst
+++ b/docs/edgeql/sdl/functions.rst
@@ -35,8 +35,9 @@ commands <ref_eql_ddl_functions>`.
 
     function <name> ([ <argspec> ] [, ... ]) -> <returnspec>
     "{"
-        from <language> <functionbody> ;
+        session_only := {true | false} ;
         [ <annotation-declarations> ]
+        from <language> <functionbody> ;
     "}" ;
 
     # where <argspec> is:
--- a/edb/edgeql/compiler/__init__.py
+++ b/edb/edgeql/compiler/__init__.py
@@ -128,7 +128,8 @@ def compile_ast_to_ir(tree,
                       implicit_tid_in_shapes=False,
                       schema_view_mode=False,
                       disable_constant_folding=False,
-                      json_parameters=False):
+                      json_parameters=False,
+                      session_mode=False):
     """Compile given EdgeQL AST into EdgeDB IR."""
 
     if debug.flags.edgeql_compile:
@@ -144,7 +145,8 @@ def compile_ast_to_ir(tree,
         implicit_tid_in_shapes=implicit_tid_in_shapes,
         schema_view_mode=schema_view_mode,
         disable_constant_folding=disable_constant_folding,
-        json_parameters=json_parameters)
+        json_parameters=json_parameters,
+        session_mode=session_mode)
 
     if path_prefix_anchor is not None:
         path_prefix = anchors[path_prefix_anchor]
@@ -252,7 +254,10 @@ def compile_func_to_ir(func, schema, *,
         tree, schema, anchors=anchors, func=func,
         security_context=security_context, modaliases=modaliases,
         implicit_id_in_shapes=implicit_id_in_shapes,
-        implicit_tid_in_shapes=implicit_tid_in_shapes)
+        implicit_tid_in_shapes=implicit_tid_in_shapes,
+        # the body of a session_only function can contain calls to
+        # other session_only functions
+        session_mode=func.get_session_only(schema))
 
     return ir
 
--- a/edb/edgeql/compiler/func.py
+++ b/edb/edgeql/compiler/func.py
@@ -87,6 +87,14 @@ def compile_FunctionCall(
     else:
         matched_call = matched[0]
 
+    func = matched_call.func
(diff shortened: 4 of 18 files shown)


tests/test_edgeql_introspection.py
```diff
@@ -637,6 +637,29 @@ class TestIntrospection(tb.QueryTestCase):
...
+                    name,
+                    session_only
+                }
+                FILTER .name IN {'std::count', 'sys::advisory_lock'}
+                ORDER BY .name;
+            """,
+            [
+                {
+                    "name": "std::count",
+                    "session_only": {},
```
elprans: Why is this an empty set as opposed to the default `False`?

edb/pgsql/compiler/relgen.py
```diff
@@ -250,6 +250,10 @@ def _get_set_rvar(
         rvars = process_set_as_tuple_indirection(ir_set, stmt, ctx=ctx)
 
     elif isinstance(ir_set.expr, irast.FunctionCall):
+        if not ctx.session_mode and ir_set.expr.session_only:
+            raise RuntimeError(
+                f'{ir_set.expr.func_shortname} requires a session')
```
1st1:
```suggestion
                f'{ir_set.expr.func_shortname}() requires a session')
```

edb/lib/sys.edgeql
```diff
@@ -64,6 +64,7 @@ sys::sleep(duration: std::duration) -> std::bool
 CREATE FUNCTION
 sys::advisory_lock(key: std::int64) -> std::bool
 {
+    SET session_only := True;
```
1st1: Add it also for sys::sleep()

edb/pgsql/compiler/relgen.py
```diff
@@ -250,6 +250,10 @@ def _get_set_rvar(
         rvars = process_set_as_tuple_indirection(ir_set, stmt, ctx=ctx)
 
     elif isinstance(ir_set.expr, irast.FunctionCall):
+        if not ctx.session_mode and ir_set.expr.session_only:
+            raise RuntimeError(
```
elprans: This check should be done in the EdgeQL translation phase, where we have the source context.  Also, `QueryError` should be raised, since this is a user error.

tests/test_edgeql_introspection.py
vpetrovykh: Oh, that's probably a bug then. I think I've seen this before with introspection queries specifically. I'll see if I can find more examples and put them into a separate test.

edb/lib/sys.edgeql
vpetrovykh: Added.

edb/pgsql/compiler/relgen.py
vpetrovykh: OK, I've refactored the code to make this check at EdgeQL compile time.

docs/edgeql/ddl/functions.rst
```diff
@@ -45,6 +45,7 @@ CREATE FUNCTION
 
     # where <subcommand> is one of
 
+      SET session_only := {true | false}
```
elprans: Move this to below the `FROM` clause to group with `SET ANNOTATION`.

docs/edgeql/sdl/functions.rst
```diff
@@ -35,6 +35,7 @@ commands <ref_eql_ddl_functions>`.
 
     function <name> ([ <argspec> ] [, ... ]) -> <returnspec>
     "{"
+        session_only := {true | false} ;
```
elprans: ditto

edb/edgeql/compiler/__init__.py
```diff
@@ -144,7 +145,8 @@ def compile_ast_to_ir(tree,
         implicit_tid_in_shapes=implicit_tid_in_shapes,
         schema_view_mode=schema_view_mode,
         disable_constant_folding=disable_constant_folding,
-        json_parameters=json_parameters)
+        json_parameters=json_parameters,
+        session_mode=session_mode)
```
elprans: Session mode is a global switch, so it should be in the `Environment` portion of the context.

edb/edgeql/compiler/func.py
```diff
@@ -87,6 +87,13 @@ def compile_FunctionCall(
     else:
         matched_call = matched[0]
 
+    func = matched_call.func
+    func_name = func.get_shortname(env.schema)
+
+    if not ctx.session_mode and func.get_session_only(env.schema):
+        raise errors.QueryError(
+            f'{func_name}() requires a session', context=expr.context)
```
elprans: Perhaps a clearer "the {func_name}() function cannot be called in a non-session context"?

tests/test_session.py
```diff
@@ -180,3 +184,143 @@ class TestSession(tb.QueryTestCase):
             )
         finally:
             await tx.rollback()
+
+    async def test_session_only_function_01(self):
```
elprans: I don't think the positive tests are all that useful.  The session_only is applied to the whole query, so testing "various SELECT scenrarios" is redundant.

docs/edgeql/ddl/functions.rst
vpetrovykh: OK, I grouped the "SET" commands together.

docs/edgeql/sdl/functions.rst
vpetrovykh: Grouped them the same as in DDL.

edb/edgeql/compiler/func.py
vpetrovykh: Updated wording

edb/edgeql/compiler/__init__.py
vpetrovykh: OK, refactored to be an `Environment` attribute.

tests/test_session.py
vpetrovykh: Removed extra 2 tests.

docs/edgeql/ddl/functions.rst
```diff
@@ -45,8 +45,9 @@ CREATE FUNCTION
 
     # where <subcommand> is one of
 
-      FROM <language> <functionbody>
+      SET session_only := {true | false}
```
1st1:
```suggestion
      SET session_only := { true | false }
```

edb/schema/functions.py
```diff
@@ -633,6 +633,9 @@ class Function(CallableObject, s_abc.Function):
     initial_value = so.SchemaField(
         expr.Expression, default=None, compcoef=0.4, coerce=True)
 
+    session_only = so.SchemaField(
```
1st1: You also want to add a validation that all overloaded functions have the same `session_only` setting.  Similarly you will later do the same for `volatility`.

tests/test_edgeql_introspection.py
1st1: I don't see this as being resolved.  Why is `{}` still here?

docs/edgeql/ddl/functions.rst
vpetrovykh: The `{true | false}` is the current de facto standard in the docs.

tests/test_edgeql_introspection.py
vpetrovykh: I made a separate branch with some more schema introspection tests, but then forgot to update this specific one. I've changed it to `False` now and marked it with `xfail`.

edb/schema/functions.py
vpetrovykh: Yes, good catch. Check added.

PR #633, opened 2019-04-25
tests: Add some introspection tests.

vpetrovykh:
Add introspection tests for regular and schema types.

### DIFF
--- a/tests/test_edgeql_introspection.py
+++ b/tests/test_edgeql_introspection.py
@@ -939,6 +939,112 @@ class TestIntrospection(tb.QueryTestCase):
         self.assertIsNotNone(result[0].links[0].target.id)
         self.assertIsNotNone(result[0].properties[0].target.id)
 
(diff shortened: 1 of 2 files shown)


tests/test_edgeql_introspection.py
```diff
@@ -939,6 +939,123 @@ class TestIntrospection(tb.QueryTestCase):
...
+        await self.assert_query_result(
+            r'''
+                WITH MODULE schema
+                SELECT ObjectType {
+                    name,
+                    is_abstract
+                }
+                FILTER .name LIKE 'test::%'
+                ORDER BY .name;
+            ''',
```
1st1: I don't like tests like this because they break when we add another type to a schema. Select explicitly the names you're testing.

vpetrovykh: I've trimmed the test to a couple specific types.

PR #637, opened 2019-04-26
Implement schema object reference tracking in expressions

elprans:
Currently any references to schema objects used in schema expressions do
not participate in reference tracking.  The fix is conceptually
simple--record all schema refs in an `ObjectSet` in every `Expression`
instance.

In practice this requires _all_ schema expressions to be compilable to IR,
which is especially tricky if the expression in question is part of some
generic construct, such as an abstract constraint.  So, the changes here
also include the necessary tweaks to the EdgeQL compiler, mostly around
callable resolution.  Another issue is potential object reference cycles
caused by expressions, so DDL generated by schema diffs now has to separate
operations involving expression attributes.

### DIFF
--- a/edb/edgeql/compiler/__init__.py
+++ b/edb/edgeql/compiler/__init__.py
@@ -129,7 +129,8 @@ def compile_ast_to_ir(tree,
                       schema_view_mode=False,
                       disable_constant_folding=False,
                       json_parameters=False,
-                      session_mode=False):
+                      session_mode=False,
+                      allow_abstract_opers=False):
     """Compile given EdgeQL AST into EdgeDB IR."""
 
     if debug.flags.edgeql_compile:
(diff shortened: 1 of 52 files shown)


edb/edgeql/compiler/__init__.py
```diff
@@ -129,7 +129,8 @@ def compile_ast_to_ir(tree,
                       schema_view_mode=False,
                       disable_constant_folding=False,
                       json_parameters=False,
-                      session_mode=False):
+                      session_mode=False,
+                      allow_abstract_opers=False):
```
1st1: Can we please call it `allow_abstract_operators`?

edb/edgeql/compiler/context.py
```diff
@@ -158,6 +173,18 @@ def __init__(self, *, schema, path_scope,
         self.json_parameters = json_parameters
         self.session_mode = session_mode
         self.allow_abstract_opers = allow_abstract_opers
+        self.allow_generic_type_output = allow_generic_type_output
+        self.schema_refs = set()
+        self.func_params = func_params
+        self.parent_object_type = parent_object_type
+
+    def get_schema_object(self, name, *,
```
1st1: Maybe rename to `get_track_object()`

PR #641, opened 2019-05-01
Fix optimistic execute / cached SQL statements to work in tx with DDL

1st1:


edb/server/dbview/dbview.pyx
```diff
@@ -174,7 +174,9 @@ cdef class DatabaseConnectionView:
 
     cdef lookup_compiled_query(self, bytes eql, bint json_mode,
                                bint expect_one):
-        if self._tx_error or not self._query_cache_enabled:
+        if (self._tx_error or
+                not self._query_cache_enabled or
+                self._in_tx_with_ddl):
```
elprans: Shouldn't DDL statements bump dbver to invalidate the comp cache?

1st1: They should and they do. The fix makes everything work while *inside a transaction with DDL* (dover is bumped when a transaction is committed)

elprans: I see.  I'm a bit concerned about the impact on large scripts wrapped into a transaction (i.e. `cat script.edgeql | edgedb --single-transaction`), so we'll need to revisit this to see if we can come up with a more granular solution.

1st1: We can, but we'll need to first implement the pre-lexing etc., so i'll just merge this as is for now.

PR #642, opened 2019-05-06
Document the binary protocol

elprans:
This adds the documentation of the wire format of the binary protocol.

### DIFF
--- a/docs/internals/protocol/dataformats.rst
+++ b/docs/internals/protocol/dataformats.rst
@@ -1,0 +1,380 @@
+.. _ref_proto_dataformats:
+
+=================
+Data Wire Formats
+=================
+
+This section describes the data wire format of standard EdgeDB types.
+
+
+.. _ref_protocol_fmt_array:
+
+array<>
+=======
+
+The array values are represented as the following structure:
+
+.. code-block:: c
+
+    struct ArrayValue {
+        // Number of dimensions, currently must
+        // always be 1.
+        int32       ndims;
+        // Reserved.
+        int32       reserved;
+        // Dimension data.
+        Dimension   dimensions[ndims];
+        // Element data, the number of elements
+        // in this array is the sum of dimension sizes:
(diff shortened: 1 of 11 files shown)


docs/internals/protocol/dataformats.rst
```diff
@@ -0,0 +1,374 @@
...
+        byte        data[length];
+    };
+
+
+.. _ref_protocol_fmt_uuid:
+
+std::uuid
+=========
+
+The :eql:type:`std::uuid` values are represented as an array of 16 unsigned
```
1st1:
```suggestion
The :eql:type:`std::uuid` values are represented as a sequence of 16 unsigned
```

docs/internals/protocol/dataformats.rst
```diff
@@ -0,0 +1,374 @@
...
+
+
+.. _ref_protocol_fmt_datetime:
+
+std::datetime
+=============
+
+The :eql:type:`std::datetime` values are represented as a 64-bit integer,
+most sigificant byte first.  The value is the number of *microseconds*
+between the encoded datetime and January 1st 2000, 00:00 UTC.
```
1st1: Maybe it's worth mentioning that `946684800` is the *proper* number of seconds that should be used to convert between 2000/1970 timestamps.

docs/internals/protocol/messages.rst
```diff
@@ -0,0 +1,926 @@
...
+        byte            data[32];
+    };
+
+
+.. _ref_protocol_msg_server_parameter_status:
+
+ParameterStatus
+===============
+
+Sent by: server
```
1st1: Nit: some "Sent by: server" end with a period and some don't.

1st1:
Should we add a "close" message? Or we do that in a separate pr?

elprans:
> Should we add a "close" message? Or we do that in a separate pr?

In a separate PR that actually adds the message.

PR #654, opened 2019-05-11
docs: Replace function call

pedronc:
Function 'unnest' does not exist in the released version (alpha-1) , so it should be replaced by 'array_unpack '


1st1:
Thank you!

PR #659, opened 2019-05-17
Fix a couple of protocol issues found when documenting it

elprans:
The startup sequence has been shortened, the connection parameters are no
longer sent as a separate message, and are instead sent as part of the
initial `ClientHandshake` message.

The `ServerKeyData` message now contains a sequence of 32 bytes instead of
a 32-bit integer.

This is a breaking change, but no bump to the protocol version, since we're
in alpha.

### DIFF
--- a/edb/pgsql/dbops/types.py
+++ b/edb/pgsql/dbops/types.py
@@ -59,6 +59,28 @@ class TypeExists(base.Condition):
         ''')
 
 
+def type_oid(name):
+    if len(name) == 2:
+        typnamespace, typname = name
+    else:
+        typname = name[0]
+        typnamespace = 'pg_catalog'
+
+    qry = textwrap.dedent(f'''\
+        SELECT
+            typ.oid
+        FROM
+            pg_catalog.pg_type typ
+            INNER JOIN pg_catalog.pg_namespace nsp
+                ON nsp.oid = typ.typnamespace
+        WHERE
+            typ.typname = {ql(typname)}
+            AND nsp.nspname = {ql(typnamespace)}
+    ''')
+
+    return base.Query(qry)
+
+
 CompositeTypeExists = TypeExists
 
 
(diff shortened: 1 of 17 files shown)


elprans:
PR updated with the revert of Postgres OID elision in array and tuple data.

edb/pgsql/dbops/types.py
```diff
@@ -59,6 +59,28 @@ def code(self, block: base.PLBlock) -> str:
         ''')
 
 
+def type_oid(name):
```
1st1: Will this query be run for every type?  Maybe read them all in one query?

edb/server/mng_port/edgecon.pyx
```diff
@@ -1299,8 +1357,26 @@ cdef class EdgeConnection:
 
         out_buf.write_int16(<int16_t>argsnum)
 
-        in_len = frb_get_len(&in_buf)
-        out_buf.write_cstr(frb_read_all(&in_buf), in_len)
+        if array_tids:
+            # we have array parameters, ensure all of them
+            # have correct element OIDs as per Postgres' expectations.
+            array_tid_map = {a[0]: a[1] for a in array_tids}
```
1st1: Is there a way to avoid dict here?

edb/pgsql/dbops/types.py
elprans: This query is ran for every `CREATE TYPE` DDL, to *write* the newly added type's OID.  This isn't really performance-critical.  The *reading* of OIDs for the newly added types is done by a single query all at once.

edb/server/mng_port/edgecon.pyx
elprans: The only alternative seems to be doing an O(N) search on the passed tuple `len(array_tids)` times.  Iterating over the tuple isn't very convenient either.

elprans: Can I store dicts in query units?

PR #661, opened 2019-05-24
edgeql: Add 'volatility' attribute.

vpetrovykh:
The value of volatility can be one of "IMMUTABLE", "STABLE", and
"VOLATILE". The attribute translates into the PostgreSQL volatility
categories. The default is "VOLATILE".

Functions, operators and casts all have volatility.

This PR doesn't update docs or set-up any checks for the declared volatility. The reason is that it's not yet clear whether a user can declare arbitrary volatility of an EdgeQL function or whether that should be inferred from the EdgeQL function body. So this may or may not end up being a user-facing attribute.


1st1:
Why don't we make `IMMUTABLE` as a default for functions?  This is the usual case.  The behaviour should be as follows: the default is `IMMUTABLE`. If the function uses a non-immutable function, operator, or selects object types, we require the user to manually set the `volatility` attribute.  @elprans thoughts?

elprans:
@1st1 We may end up inferring the volatility as Victor suggested above.  Defining `IMMUTABLE` as a hard default might be dangerous.  My suggestion is to merge this to fix the behavior of `std` operators and functions, and revisit the issue of inference and volatility UX in a separate issue.

1st1:
>  revisit the issue of inference and volatility UX in a separate issue.

ok

edb/lib/schema.edgeql
```diff
@@ -35,6 +35,10 @@ CREATE SCALAR TYPE schema::operator_kind_t EXTENDING std::str {
     CREATE CONSTRAINT std::one_of ('INFIX', 'POSTFIX', 'PREFIX', 'TERNARY');
 };
 
+CREATE SCALAR TYPE schema::volatility_t EXTeNDING std::str {
```
1st1:
```suggestion
CREATE SCALAR TYPE schema::volatility_t EXTENDING std::str {
```

1st1: Why a constraint? Didn't we want to have an enum for this as there's actually an ordering that makes sense for volatility values?

edb/lib/std/25-booloperators.edgeql
```diff
@@ -34,70 +34,96 @@
...
+std::`?!=` (l: OPTIONAL std::bool, r: OPTIONAL std::bool) -> std::bool {
+    SET volatility := 'IMMUTABLE';
     FROM SQL EXPRESSION;
+};
 
 
 CREATE INFIX OPERATOR
-std::`>=` (l: std::bool, r: std::bool) -> std::bool
+std::`>=` (l: std::bool, r: std::bool) -> std::bool {
+    SET volatility := 'IMMUTABLE';
```
1st1: You set a default volatility for all operators to `IMMUTABLE`. Why this duplication?

edb/lib/std/30-jsonfuncs.edgeql
```diff
@@ -64,201 +68,279 @@ std::json_get(
...
     FROM SQL $$
         SELECT array_agg(j)
         FROM jsonb_array_elements(val) AS j
     $$;
+};
 
 
-CREATE CAST FROM std::bool TO std::json
+CREATE CAST FROM std::bool TO std::json {
+    SET volatility := 'STABLE';
```
1st1: Just curious, why isn't this immutable?

edb/lib/std/25-booloperators.edgeql
elprans: Somewhat confusingly, he set a default on the *introspection* type, the overall default is `VOLATILE`:

https://github.com/edgedb/edgedb/pull/661/files#diff-de65a2723408cb4f1ab36bd8fed914e1R402

edb/lib/std/30-jsonfuncs.edgeql
elprans: Because `to_jsonb` isn't immutable.

elprans: Although it's worth investigating exactly why and for which specific types, since `to_jsonb()` is generic.

edb/lib/std/25-booloperators.edgeql
1st1: * What's the point of setting it on the introspection type? 
* Why is it different? 
* Should we set the default for operators to IMMUTABLE?

edb/lib/std/30-jsonfuncs.edgeql
1st1: I suggest to add comments then to every function that isn't IMMUTABLE explaining why.

edb/lib/std/25-booloperators.edgeql
elprans: We should't be setting defaults on introspection types.  Operators aren't special (but they tend to be immutable).

edb/lib/std/30-jsonfuncs.edgeql
elprans: +1

edb/lib/schema.edgeql
vpetrovykh: At the moment enums proper don't work nicely with assignment casts, defaults, etc. Fixing that is a separate orthogonal issue, once enums are usable in the schema we can make a PR that updates the current usage of the constraint to an actual enum type (for volatility and for other schema enums possibly). 

I am looking into fixing the issues with enums right now.

edb/lib/std/25-booloperators.edgeql
vpetrovykh: I've removed the defaults for volatility except the one for the base class. I left it in mainly for purposes of showing that there IS a default value, which otherwise would remain buried in python.

edb/lib/std/30-jsonfuncs.edgeql
vpetrovykh: I've updated all the schema definitions to have volatility specified explicitly - this would make it easier to read and not have to guess what the default is.

I've added more comments about the chosen volatility value.

PR #668, opened 2019-06-07
tests: Fix expectedFailure test_constraints_ddl_04

schettino72:
Use correct operator for string concatenation, `++`.

Seems this test is failing just because it is using `+` instead `++`.  Or I am missing something?


1st1:
Thank you!

PR #669, opened 2019-06-09
docs: edgeql/lexical, fix escape sequences examples.

schettino72:
Output from REPL is not *raw strings*... I am assuming that is expected behaviour and docs are wrong.


1st1:
Thank you!

PR #671, opened 2019-06-19
Fix missing semi-colon in migrations example

florimondmanca:
Just learnt the hard way that property definitions should end with a semi-colon. 😄 They do in the rest of the migrations guide, at least, so I figured this was a typo.

https://edgedb.com/docs/edgeql/ddl/migrations#examples

### DIFF
--- a/docs/edgeql/ddl/migrations.rst
+++ b/docs/edgeql/ddl/migrations.rst
@@ -81,7 +81,7 @@ syntax:
 
     CREATE MIGRATION init TO {
         type User {
-            property username -> str
+            property username -> str;
         }
     };
 


1st1:
@vpetrovykh hm, I think this is a bug actually.  Shouldn't we allow to omit a trailing semicolon in a block?

vpetrovykh:
I think the original intent was to allow omitting trailing semicolons inside blocks. So I think it's a bug.

florimondmanca:
Okay, lemme fill in a proper bug report then!

florimondmanca:
Actually, maybe it's not a bug, it's a feature™. I just re-read your comment @1st1:

> Shouldn't we allow to omit a **trailing** semicolon in a block?

I wasn't actually running the specific migration from the guide, but rather this one:

```
CREATE MIGRATION `0001_initial` TO {
  type Band {
    required property name -> str
  }
  type Album {
    required property title -> str
    required property year -> int64
    required link band -> Band
  }
};
```

Which yields:

```
EdgeQLSyntaxError: Unexpected 'required'
### CREATE MIGRATION `0001_initial` TO {
###   type Band {
###     required property name -> str
###   }
###   type Album {
###     required property title -> str
###     required property year -> int64
###     ^
###     required link band -> Band
###   }
### };
```

The docs made me think that semicolons were always optional, but it seems that they're required for multiple-line blocks.

To be honest, optional trailing semicolons looks like a confusing feature to me. In particular, the second example from the guide uses a trailing semicolon. So shouldn't we always require a semicolon, for the sake of consistency?

In any case, I think adding the semicolon here may be less confusing, even though it *is* possible to omit it right now.

vpetrovykh:
Broadly speaking, the intent is to treat semicolons as _separators_ (and to allow trailing `;`) as opposed to _terminators_. This is similar to how `,` are treated in a Python dict literal. This is convenient for nested one-liners, but for multi-line specifications it's often better to terminate with a semicolon so that then the specifications order can be easily rearranged.

florimondmanca:
Ah, I see. That actually makes a lot more sense now, thanks.

Perhaps we should consider adding this subtlety to [Punctuation](https://edgedb.com/docs/edgeql/lexical#punctuation)? Currently it only says "EdgeQL uses ; as a statement separator", which I don't think encompasses the case of in-block declarations.

1st1:
> Perhaps we should consider adding this subtlety to Punctuation? Currently it only reads "EdgeQL uses ; as a statement separator", which I think doesn't encompass the case of in-block declarations yet.

Yeah, sure.

florimondmanca:
Okay. :+1: I’ll close this PR since this turned out to not be an actual issue.

Issue #179, opened 2018-05-11
equivalent to \d of PostgreSQL ?


matrixise:
Hi,

with PostgreSQL, we can see the description of a table, is it possible in the REPL?

elprans:
There's no such shortcut in the repl yet, but you can run introspection queries.

To list object types:

```SELECT schema::ObjectType { name };```

To list properties an links of an object type:

```
    WITH Object := (SELECT schema::ObjectType
                    FILTER .name = '<fully-qualified object type name>`)
    SELECT Object.pointers {name};
```

For example:

```
    WITH Object := (SELECT schema::ObjectType
                    FILTER .name = 'std::Object')
    SELECT Object.pointers {name};
```

vpetrovykh:
It makes sense for us to implement a similar set of commands to PostgreSQL `\d`. In principle, we now have one more mechanism in addition to introspection queries that can help us - `DESCRIBE`. Let's consider some of the features, first.

Many PostgreSQL commands allow to specify a [pattern](https://www.postgresql.org/docs/12/app-psql.html#APP-PSQL-PATTERNS) to be matched. The pattern rules are a mix of shell file name patterns, regexp, and identifier quoting. Which incidentally means that `*`, `.`, and `?` don't have their regexp meaning and require extra care if they are needed in regexp. For example, instead of the regexp `.+` one should use `?+`, etc. I would argue that this is one of the areas where EdgeDB should try to clean up the confusion:
- There's not a lot of value in forcing identifiers to lower-case (unless quoted) for EdgeDB simply because we have case-sensitive identifiers often of mixed case. However, we may want to make the matches case-insensitive by default.
- Rather than rely on identifier quoting mechanism to force exact case match I would suggest adding a flag to the command to toggle case-sensitivity. This will reduce the overall complexity of rules and feature interaction for patterns.
- There's no real incentive to use glob-like syntax, since we use `::` as a module/name separator instead of `.` for an equivalent concept in PostgreSQL. So we would end up with `*::*` where PostgreSQL has `*.*`, which is neither here nor there in terms of similarity to something familiar. So it would make sense then to only use regexp patterns to provide flexibility in matches. The upside of this is that the users won't have to learn 2 flavors of regexp - one for `\d` commands and another for actual queries.
[...]

PR #672, opened 2019-06-22
repl: add commands \dr list roles, \dm list modules, \dt list types, …

schettino72:
…\d describe type.

### DIFF
--- a/edb/repl/__init__.py
+++ b/edb/repl/__init__.py
@@ -366,7 +366,7 @@ class Cli:
 
         print(f'Type {result[0].name}:')
         for pt in result[0].pointers:
-           print(f'  {pt.type} {pt.name} -> {pt.target}')
+            print(f'  {pt.type} {pt.name} -> {pt.target}')
 
     @_command('psql', R'\psql',
               'open psql to the current postgres process',
@@ -441,7 +441,8 @@ class Cli:
             print('> ' + '\n> '.join(srv_tb.strip().split('\n')))
             print()
 
-    def fetch(self, query: str, params: dict=None, *, json: bool, retry: bool=True):
+    def fetch(self, query: str, params: dict=None, *,
+              json: bool, retry: bool=True):
         self.ensure_connection()
         self.context.last_exception = None
 
--- a/edb/repl/__init__.py
+++ b/edb/repl/__init__.py
@@ -23,6 +23,7 @@ import os
 import select
 import subprocess
 import sys
+from string import Template
 
 import edgedb
 import immutables
@@ -347,26 +348,42 @@ class Cli:
             self.command_list_types(args)
             return
         type_name = args
-        result, _ = self.fetch(
-            f'WITH type:= INTROSPECT {type_name}' + '''
+        tmpl_prop = '''
+                name,
+                required,
+                cardinality,
+                target : {name},
+                constraints: {
+                  name,
+                  is_abstract,
+                },
+        '''
+        result, _ = self.fetch(Template(
+            '''WITH type:= INTROSPECT ${type_name}
             SELECT type {
               name,
+              is_abstract,
+              is_final,
+              bases: {name},
               annotations: {name},
-              pointers:= (
-                SELECT type.pointers {
-                  name,
-                  type := type.pointers.__type__.name[len('schema::'):],
-                  target := type.pointers.target.name
+              properties: {
+                ${tmpl_prop}
+              },
+              links: {
+                ${tmpl_prop}
+                properties: {
+                  ${tmpl_prop}
                 }
-              )
+              },
             };
-            ''',
+            ''').substitute(type_name=type_name, tmpl_prop=tmpl_prop),
             json=False
         )
 
-        print(f'Type {result[0].name}:')
-        for pt in result[0].pointers:
-            print(f'  {pt.type} {pt.name} -> {pt.target}')
+        # render.render_binary(self.context, result, max_width=80)
+        # print('==============================\n')
+        render.render_sdl(self.context, result[0], max_width=80)
+
 
     @_command('psql', R'\psql',
               'open psql to the current postgres process',
--- a/edb/repl/__init__.py
+++ b/edb/repl/__init__.py
@@ -297,6 +297,77 @@ class Cli:
         for dbn in result:
             print(f'  {dbn}')
 
+    @_command('dr', R'\dr', 'list roles')
+    def command_list_roles(self, args):
+        result, _ = self.fetch(
+            '''
+                SELECT name := sys::Role.name
+                ORDER BY name ASC
+            ''',
+            json=False
+        )
+        print('List of roles:')
+        for role in result:
+            print(f'  {role}')
+
+    @_command('dm', R'\dm', 'list modules')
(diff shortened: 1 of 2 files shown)


schettino72:
refs #179 

Not sure you already have a design on how commands should look like.
Let me know what you guys think. 
Output should probably be in table format like psql...

1st1:
I like this.  @elprans your thoughts?

elprans:
I think this is useful.  Perhaps `\d <type>` should produce SDL output, i.e

```
> \d std::Object
type std::Object {
  link __type__ -> schema::Type;
  property id -> std::uuid;
}
```

1st1:
> I think this is useful. Perhaps \d <type> should produce SDL output, i.e

+1, although it's not that easy to implement, right?  I mean ideally we should list all indexes, all constraints, etc.

elprans:
> although it's not that easy to implement, right

Why not? All of that is introspectable and can be fetched in a single query. But just the links/properties would be good for the initial implementation.

schettino72:
My original idea was to have `\dx` commands to output in tabular text. And have another command to dump the schema (SDL or DDL).

So I agree there should be a way to generate SDL, just concerned about having some consistency in the command interface. But of course you guys decision.

Right now there is no way generate a "schema" dump right?
Does the SDL support Administration statements like "ROLE" ?

If `\d` displays SDL... what should `\dr` should display? how the properties `allow_login`, `is_superuser` should be displayed? (right now it only display role name)

elprans:
> If \d displays SDL... what should \dr should display? how the properties allow_login, is_superuser should be displayed? (right now it only display role name)

I don't think we should be overly concerned with the consistency of `\d` output, and instead find the most useful and readable format for each use case.  Tables make sense for simple schema items, but object types may contain lots of stuff and SDL-ish presentation makes sense to me from the familiarity and readability standpoint.  @1st1 @vpetrovykh what do you guys think?

>  Does the SDL support Administration statements like "ROLE" ?

It doesn't.

> Right now there is no way generate a "schema" dump right?

Not right now, but there will be `edgedb dump` in Alpha 2.

1st1:
> I don't think we should be overly concerned with the consistency of \d output, and instead find the most useful and readable format for each use case. [...] @1st1 @vpetrovykh what do you guys think?

+1.  There's nothing inherently special about tabular output.  We will use it where it makes sense, but in this case SDL makes more sense from the usability standpoint.

vpetrovykh:
Yeah, the \d commands are convenience shortcuts made for humans, so the output should be whatever is good for readability, they don't have to produce valid SDL.

schettino72:
Ok, got it. I will work on it when i have time.

schettino72:
Finally did some work on this. Still WIP and lots to do...
Just want to check if you guys thing this is an ok approach.
Basically I just created an "SDLRenderer".

I am using the `User` type from `cards` (in test folder sample schema).

```
abstract type Named {
    }
}

type User extending Named {
    multi link deck -> Card {
        property count -> int64;
    }

    property deck_cost := sum(.deck.cost);

    multi link friends -> User {
        property nickname -> str;
        # how the friend responded to requests for a favor
        #property favor -> array<bool>
    }

    multi link awards -> Award {
        constraint exclusive;
    }
}
```

So far I get the following output:

```
abstract type Named extending Object {
        constraint exclusive;
    }
}

type User extending Named {
    property deck_cost -> int64;
        constraint exclusive;
    }
    multi link awards -> Award {
        constraint exclusive;
    }
    multi link deck -> Card {
        property count -> int64;
    }
    multi link friends -> User {
        property nickname -> str;
    }
}
```

Notes:

- I could not find `delegated` info of `constraint` from introspection
- Also could not figure when a property is defined in a type or in a base type, actually I think it is nice to show properties from base types. But probably should have some kind of indication.
- is possible to get the order in which properties/links were defined?
- other differences I still need to work on

elprans:
> I could not find delegated info of constraint from introspection

This is a bug.  

> Also could not figure when a property is defined in a type or in a base type, actually I think it is nice to show properties from base types. But probably should have some kind of indication.

Yes.  I'm working on a patch to add this kind of tracking to the inherited objects.

> is possible to get the order in which properties/links were defined?

It's undefined.

> other differences I still need to work on 

The approach looks good to me. Thanks!

funkyfuture:
> Yeah, the \d commands are convenience shortcuts made for humans

not really with non-querty keyboard layouts b/c of the `\` with regards to typing. regarding readability, i'd say it's convenient for Postgres users, not all humans.

1st1:
@elprans do your last two PRs unblock this one?

elprans:
@1st1 Mostly, but we still need to expose the information on whether a particular field was inherited or defined locally in the introspection schema (i.e. the `field_inh_map`).

1st1:
@elprans can you work on that soon?

@schettino72 Eduardo, are you interested in finishing this PR?  We'd like to get it merged as soon as we can.

elprans:
The other thing to consider here is whether or not do we want make this a proper command, i.e. `DESCRIBE TYPE Foo` as opposed to a CLI feature.

1st1:
> The other thing to consider here is whether or not do we want make this a proper command, i.e. DESCRIBE TYPE Foo as opposed to a CLI feature.

I'm open to that.  That would make it much easier to implement third party tools/repls.  @victor do you want to spec this in a new issue?

schettino72:
Sorry guys. I had zero time for open-source lately. I am hoping I will have time to get back in a week or two.

MrNaif2018:
@1st1 @elprans I can finish this pr if needed. What should I do?

elprans:
@MrNaif2018 We need to weigh the pros and cons of the two approaches and decide which to pursue:

1) Implement the "list" and the "describe" the way @schettino72 is proposing in this PR: i.e. through introspection queries and client-side rendering.

2) Implement this functionality as server commands, i.e `DESCRIBE TYPE` that simply returns a pre-rendered string representation of the desired schema item.

The second option requires less code duplication and, IMO, better UX (although we could, with some effort, recognize `DESCRIBE` as a client-side command, like we do for slash commands).  On the other hand, the first option does not require adding new syntax, and is more flexible, as it is easier to customize the output rendering (color highlighting, width-formatting etc).

@1st1 @vpetrovykh @ambv what are your thoughts?

MrNaif2018:
My thoughts are for the second option, because the more client libraries the more code repetition. But can't \dr and other commands internally call describe type and have all the color rendering then?

MrNaif2018:
My thoughts are for the second option, because the more client libraries the more code repetition. But can't \dr and other commands internally call describe type and have all the color rendering then?

1st1:
> Implement this functionality as server commands, i.e DESCRIBE TYPE that simply returns a pre-rendered string representation of the desired schema item.

This, the second option.  Even though we'll need to add some new syntax, that will make it easier for people to write tools / ide plugins / repls / etc for EdgeDB.

1st1:
@vpetrovykh Let's make this a high priority.

vpetrovykh:
There's a number of changes that impact the implementation of this PR. I'm closing it in favour of the reworked version (PR #853).

PR #679, opened 2019-07-18
docs: Add a custom JS sphinx plugin.

vpetrovykh:
The custom plugin makes it possible to document the JS driver in a
manner similar to the rest of the EdgeDB docs.

### DIFF
--- a/edb/tools/docs/js.py
+++ b/edb/tools/docs/js.py
@@ -1,0 +1,148 @@
+#
+# This source file is part of the EdgeDB open source project.
+#
+# Copyright 2019-present MagicStack Inc. and the EdgeDB authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


edb/tools/docs/js.py
```diff
@@ -0,0 +1,137 @@
...
+from docutils import nodes as d_nodes
+from docutils.parsers.rst import directives
+
+from sphinx import addnodes as s_nodes
+from sphinx.domains import javascript as js
+from sphinx.locale import _
+from sphinx.util import docfields
+
+
+def merge_dicts(*dicts):
```
1st1: You can just `merged = {**d1, **d2, **d3}`, or

```python
merged = dict(
  **d1,
  **d2,
  key1=val1,
  key2=val2
)
```

so no need for this.

edb/tools/docs/js.py
```diff
@@ -0,0 +1,137 @@
...
+        res.update(d)
+
+    return res
+
+
+class JSFieldMixin:
+    def make_xref(self, rolename, domain, target,
+                  innernode=d_nodes.emphasis, contnode=None, env=None):
+
+        if domain == 'js' and rolename:
```
1st1: Can domain be something else here?

1st1:
I'm curious -- isn't there a JS domain for Sphinx already?

edb/tools/docs/js.py
vpetrovykh: Right. I've updated the code and dropped the helper `merge_dicts`.

edb/tools/docs/js.py
vpetrovykh: Hmm... probably not.

PR #686, opened 2019-07-31
Reimplement SDL through DDL

elprans:
Currently, SDL is implemented as a separate layer that translates SDL AST
into schema directly.  There are several problems with the current 
implementation and the approach in general:

1) it is hard to share code between SDL and DDL schema
definition paths; 2) definition validation and dependency management is
very hacky; 3) DDL code languishes, since SDL is the primary way to define
the schema.

The most obvious solution is to reimplement SDL in terms of DDL, which is
what I've done here.

This is a rather large patch, since there's unfortunately no way to get 
away with a fundamental change like this without breaking half of the test
suite.

Here's a rough outline of the changes here:

- a separate SDL AST is removed, since SDL is syntactically just DDL
with stripped `CREATE` and `SET` keywords;

- a new 'SDL to DDL' translation layer is added that breaks the SDL
  definitions into granular DDL operations and takes care of reference
  management;

- a new generic dependency resolver is added to produce correct
  migrations from schema diffs;

- the inherit-by-reference approach toward schema ref inheritance
  (pointers, constraints etc) is replaced with material inheritance,
  where refs are always derived in the subclass even in cases of
  pure inheritance and the refs in parent classes are now recorded
  as bases explicitly in the derived ref;

- the above allowed to replace the gnarly `derive()` codepath with
  a simple DDL op;

- introspection of unset schema object fields now correctly returns
  the default value;

- new infrastructure to track whether a field was inherited or set
  locally;

- fixes to the general schema reference tracking.


1st1:
Glanced through it, obviously too much for a thorough review. Let's merge this!

Issue #683, opened 2019-07-26
GraphQL mutations


vpetrovykh:
For now we'll address the 3 basic mutations: `INSERT`, `UPDATE`, and `DELETE`.

# Delete

The `DELETE` mutations is the simplest of them. It has the same signature as a query (essentially because any `SELECT` can be wrapped into a `DELETE`). It also returns the deleted objects, so that their details can be queried:

```graphql
mutation delete_Foo(
    # same as query parameters
    filter: ...,
    order: ...,
    first: ...,
    last: ...,
    before: ...,
    after: ...
) {
    id
}
```

# Insert

The `INSERT` mutation supports inserting one or more new objects. Links of those new objects can themselves be either new objects or existing ones.

```graphql
mutation insert_Foo(
    data: [{
        foo_prop1: 42,
        foo_prop2: "aardvark",
        some_link: [{
            data: {
                name: "new item"
            }
        }, {
            data: {
                # existing item by ID
                id: "...UUID..."
            }
        }, {
            # objects can be specified via 
            # the same interface as a query
            filter: ...,
            order: ...,
            first: ...,
            last: ...,
            before: ...,
            after: ...
        }]
    }]
) {
    id
    foo_prop1
    foo_prop2
    some_link {
        id
        name
    }
}
```

The `data` field and all the other selection fields are mutually
exclusive. Specifying both for the same object is an error.

# Update

The `UPDATE` mutation supports setting as well as some other common
operations. The basis for `UPDATE` is the same as for a regular query
since we first need to select the objects to update. Then we to
specify the data to be updated and the operations to be performed on
it.

```graphql
mutation update_Foo(
    # same as query parameters
    filter: ...,
    order: ...,
    first: ...,
    last: ...,
    before: ...,
    after: ...
    # actual update     
    data: {
        foo_prop1: {add: 1},
        foo_prop2: {set: "aardvark"},
        some_link: {append: 
            [{
                data: {
                    # existing item by ID
                    id: "...UUID..."
                }
            }, {
                # objects can be specified via 
                # the same interface as a query
                filter: ...,
                order: ...,
                first: ...,
                last: ...,
                before: ...,
                after: ...
            }]
        }
    }
) {
    id
    ...
}
```

The `data` field and all the other selection fields are mutually
exclusive. Specifying both for the same object is an error.

vpetrovykh:
Documentation updated in #697 
Closing the issue.

PR #688, opened 2019-08-03
Gql

vpetrovykh:
Implementation of specs outlined in #683.

The operations that require self-reference (like "append", etc.) currently can't be implemented due to a scoping bug.

### DIFF
--- a/edb/graphql/codegen.py
+++ b/edb/graphql/codegen.py
@@ -17,6 +17,7 @@
 #
 
 
+import json
 from edb.common.ast import codegen
 
 
@@ -130,7 +131,8 @@ class GraphQLSourceGenerator(codegen.SourceGenerator):
         self._visit_arguments(node)
 
     def visit_StringValue(self, node):
-        self.write(f'"{node.value}"')
+        # the GQL string works same as JSON string
+        self.write(json.dumps(node.value))
 
     def visit_IntValue(self, node):
         self.write(node.value)
--- a/edb/graphql/types.py
+++ b/edb/graphql/types.py
@@ -102,6 +102,7 @@ GQL_TO_OPS_MAP = {
 
 
 HIDDEN_MODULES = s_schema.STD_MODULES - {'std'}
+TOP_LEVEL_TYPES = {'Query', 'Mutation'}
 
 
 class GQLCoreSchema:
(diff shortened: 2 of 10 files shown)


edb/graphql/types.py
```diff
@@ -808,9 +826,24 @@ def get_field_type(self, name):
...
     edb_type = 'stdgraphql::Mutation'
 
-    def __init__(self, schema, **kwargs):
-        self.modules = schema.modules
-        super().__init__(schema, **kwargs)
+    def get_field_type(self, name):
+        fkey = (name, self.dummy)
+        target = None
+
+        print('AAA mut:', name, flush=True)
```
elprans: Drop this and amend the commit please.

edb/server/bootstrap.py
```diff
@@ -358,6 +358,7 @@ def _process_delta(delta, schema):
 async def _populate_data(std_schema, schema, conn):
     script = '''
         INSERT stdgraphql::Query;
+        INSERT stdgraphql::Mutation;
```
elprans: AFAIR, this was causing a crash in bootstrap due to a bug I fixed, but you put the fix into a subsequent commit, please rearrange correctly.  Alternatively, just squash the commits.

edb/graphql/codegen.py
```diff
@@ -130,7 +130,9 @@ def visit_Directive(self, node):
         self._visit_arguments(node)
 
     def visit_StringValue(self, node):
-        self.write(f'"{node.value}"')
+        # only need to escape '"'
+        esc = node.value.replace('"', r'\"')
```
1st1: What if a string has a trailing slash, e.g. `blah\`

edb/graphql/types.py
vpetrovykh: Done.

edb/server/bootstrap.py
vpetrovykh: OK, moved the fix to the earlier commit. The graphql tests pass for that commit now.

edb/graphql/codegen.py
vpetrovykh: Right. This should be using the JSON converter to get the correct string representation since the underlying value is a Python string. I've updated the handling of StringValue node in codegen and translator (translator uses `from_python`).

PR #700, opened 2019-08-18
Remove explicit index names

elprans:
Like constraints, indexes are really defined by the *expression*, and
the need to specify a synthetic name is more of a hinderance, due to
possible inheritance conflicts, so remove the ability to specify index
names.

Also make indexes actually properly inheritable in the schema.

### DIFF
--- a/docs/datamodel/indexes.rst
+++ b/docs/datamodel/indexes.rst
@@ -22,7 +22,7 @@ or more properties directly:
 
     type User {
         property name -> str;
-        index name_idx on (__subject__.name);
+        index on (__subject__.name);
     }
 
 With the above, ``User`` lookups by the ``name`` property will be faster,
@@ -41,7 +41,7 @@ of the host object type:
     type User {
         property firstname -> str;
         property lastname -> str;
-        index name_idx on (str_lower(
+        index on (str_lower(
             __subject__.firstname + ' ' + __subject__.lastname));
     }
 
@@ -51,7 +51,7 @@ Similarly indexes may refer to the link properties if the *subject* is a link:
 
     abstract link friends_base {
         property nickname -> str;
-        index nickname_idx on (__subject__@nickname);
+        index on (__subject__@nickname);
     }
 
 The index expression must not reference any variables other than the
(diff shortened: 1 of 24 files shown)


docs/datamodel/indexes.rst
```diff
@@ -51,7 +51,7 @@ Similarly indexes may refer to the link properties if the *subject* is a link:
 
     abstract link friends_base {
         property nickname -> str;
-        index nickname_idx on (__subject__@nickname);
+        index on (__subject__@nickname);
```
1st1: I keep forgetting — do we really need `(` / `)` around the `expr` part?

edb/pgsql/intromech.py
```diff
@@ -574,20 +574,24 @@ def interpret_indexes(self, table_name, indexes):
...
-                    'internal metadata inconsistency',
-                    details=f'Index {index_name} is defined in schema, but'
+            if index_data['is_local']:
+                try:
+                    pg_indexes.remove((subj_table_name, index_name))
+                except KeyError:
+                    raise errors.SchemaError(
+                        'internal metadata inconsistency',
+                        details=(
+                            f'Index {index_name} is defined in schema, but'
```
1st1:
```suggestion
                            f'Index {index_name} is defined in schema, but '
```

tests/test_indexes.py
```diff
@@ -79,12 +78,34 @@ class TestIndexes(tb.DDLTestCase):
             }]
         )
 
+        await self.con.execute(
+            """
+                ALTER TYPE test::Person
+                DROP INDEX ON ((.first_name, .last_name));
```
1st1: Do we have a test that tests how we fail if a user tries to drop an index with an expression we couldn't find?

elprans: Added a test.

docs/datamodel/indexes.rst
elprans: https://github.com/edgedb/edgedb/blob/f9a73382b0e08634cfec423ef608c630a835d521/edb/edgeql/parser/grammar/commondl.py#L115-L118

PR #705, opened 2019-09-02
Make nested views properly introspectable

elprans:
Currently, only the first level of a view is introspectable, which makes
nested views useless in contexts where proper introspection is
necessary, such as in GraphQL reflection.

Making schema views fully introspectable requires two changes: the names
of nested derived views must be unique and properly formed (i.e not use
the generic derived mangling), the target refs in of the derived
pointers in the view must be kept.

Also, `DROP VIEW` is currently broken when used on a view with nesting,
since it leaves behind the derived types created for nested views.  That
is also fixed by this commit.

### DIFF
--- a/edb/schema/delta.py
+++ b/edb/schema/delta.py
@@ -22,6 +22,7 @@ from __future__ import annotations
 import base64
 import collections
 import collections.abc
+import typing
 
 from edb import errors
 
@@ -413,7 +414,7 @@ class CommandContext:
         self.altered_targets = set()
 
     @property
-    def modaliases(self):
+    def modaliases(self) -> typing.Mapping[typing.Optional[str], str]:
         maps = [t.modaliases for t in reversed(self.stack)]
         maps.append(self._modaliases)
         return collections.ChainMap(*maps)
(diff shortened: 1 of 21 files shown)


vpetrovykh:
For GraphQL we'd need to prefix nested persistent views with something other than `__` since that is reserved for GraphQL special needs (https://graphql.github.io/graphql-spec/June2018/#sec-Reserved-Names). I'm tempted to go with a simple `View__`. This is something I can do once the main PR is committed, probably.

elprans:
The reason I used `__` is because it is reserved in EdgeQL also, so there's no risk of collision with regular types.

vpetrovykh:
My point is that how we reflect the names into GraphQL is entirely arbitrary and doesn't have to block or dictate the naming convention for EdgeQL. The `__` makes sense for EdgeDB.

edb/schema/delta.py
```diff
@@ -454,6 +454,14 @@ def enable_recursion(self):
     def canonical(self):
         return any(ctx.op.canonical for ctx in self.stack)
 
+    def in_deletion(self, offset=0):
```
1st1: Can you add docstrings for these two?

edb/schema/objects.py
```diff
@@ -119,12 +119,13 @@ class Field(struct.ProtoField):  # derived from ProtoField for validation
...
                  'compcoef', 'inheritable', 'simpledelta',
-                 'merge_fn', 'ephemeral', 'introspectable', 'allow_ddl_set')
+                 'merge_fn', 'ephemeral', 'introspectable',
+                 'allow_ddl_set', 'weak_ref')
 
     def __init__(self, type_, *, coerce=False,
                  compcoef=None, inheritable=True,
                  simpledelta=True, merge_fn=None, ephemeral=False,
-                 introspectable=True, **kwargs):
+                 introspectable=True, weak_ref=False, **kwargs):
```
1st1: Let's start documenting fields/attrs. What does `weak_ref=True` do?

Issue #684, opened 2019-07-27
SELECT Object IF EXISTS Object ELSE INSERT 


cosmophile:
Hi, I tried to use the IF...ELSE and EXISTS clauses in order to construct a query that either selects an item if it exists or else inserts it. So since I can do the following:
```
SELECT 1 IF EXISTS (SELECT Person FILTER .name = 'Ryan Gosling') 
....... ELSE (SELECT 0);
```
...I tried to do
```
SELECT Person FILTER .name = 'Ryan Gosling' 
....... IF EXISTS (SELECT Person FILTER .name = 'Ryan Gosling') 
....... ELSE (SELECT <int64>{});
```
However I get the error: **QueryError: operator 'std::IF' cannot be applied to operands of type 'std::bool', 'std::bool', 'std::int64'**

Am I making some syntax error or is this kind of selection not allowed? I assumed it would work because the documentation on IF states `anytype` as the left assignment, but I have not been able to figure out how to make it work with my schema types. 

Thank you very much for your help.

cosmophile:
Soon after opening this and trying various permutations, I realized that the left and right assignments need to be identical in type and be a singleton on both ends if I want to make an insert statement.

Closing this.

1st1:
@vpetrovykh should we enhance our error message here?

cosmophile:
Another related question / error message confusion: 

This is how I am inserting a Person if they don't exist:
```
SELECT ( SELECT Person FILTER .name = 'Zendaya' LIMIT 1 ) 
....... IF 'Zendaya' IN Person.name 
....... ELSE (INSERT Person { last_name := 'Zendaya'  }  );
```
I am also able to use `FOR x IN ...` to separately _either_ insert or select multiple persons. However, if I try to use the above statement to try to do both at once in my `UNION` clause (i.e, select x if it exists or insert it if it doesn't), like this:
```
FOR x IN { 
....... ( name := 'Zendaya' ), 
....... ( name := 'Madonna') 
....... } UNION ( 
....... SELECT ( SELECT Person FILTER .name = x.name LIMIT 1 ) 
....... IF x.name IN Person.name 
....... ELSE (INSERT Person { last_name := x.name  }  ) 
....... );
```
I get the following error message (even if both Persons exist): **InternalServerError: missing FROM-clause entry for table "__derived__|x@@w-1~2"**

[...]

PR #707, opened 2019-09-03
edgeql: Update signature and error reporting of IF..ELSE.

vpetrovykh:
Make the signature of the `std::IF` operator actually match the order of
the operands in EdgeQL (if_true, condition, if_false).

Provide a more specific error hint for the mismatched types of the IF
operator.

Clarifies the error message for #684.

### DIFF
--- a/edb/edgeql/compiler/func.py
+++ b/edb/edgeql/compiler/func.py
@@ -328,11 +328,18 @@ def compile_operator(
             )
 
         if not matched:
+            hint = ('Consider using an explicit type cast or a conversion '
+                    'function.')
+
+            if op_name == 'std::IF':
+                hint = (f"The IF and ELSE result clauses must be of "
+                        f"compatible types, while the condition clause must "
+                        f"be 'std::bool'. {hint}")
+
             raise errors.QueryError(
                 f'operator {str(op_name)!r} cannot be applied to '
                 f'operands of type {types}',
-                hint='Consider using an explicit type cast or a conversion '
-                     'function.',
+                hint=hint,
                 context=qlexpr.context)
         elif len(matched) > 1:
             if in_abstract_constraint:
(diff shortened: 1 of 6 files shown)


edb/edgeql/compiler/func.py
```diff
@@ -328,11 +328,18 @@ def compile_operator(
             )
 
         if not matched:
+            if op_name == 'std::IF':
+                hint = ("The IF and ELSE result clauses must be of compatible "
+                        "types, while the condition clause must be "
+                        "'std::bool'.")
+            else:
+                hint = ('Consider using an explicit type cast or a conversion '
```
elprans: I don't think these are mutually exclusive.  This is still a valid hint in the 'std::IF' case.

edb/lib/std/25-setoperators.edgeql
```diff
@@ -92,7 +92,7 @@ std::`??` (l: OPTIONAL anytype, r: SET OF anytype) -> SET OF anytype {
 
 
 CREATE TERNARY OPERATOR
-std::`IF` (condition: bool, if_true: SET OF anytype,
+std::`IF` (if_true: SET OF anytype, condition: bool,
```
elprans: Why do you think this is necessary or important?

edb/edgeql/compiler/func.py
vpetrovykh: Right, I'll merge them for IF.

edb/lib/std/25-setoperators.edgeql
vpetrovykh: Because the error message ends up listing the signature types in an odd order, which confused me when I tried `SELECT 1 IF True ELSE False` since the error message complained about operands being `std::bool`, `std::int64`, `std::bool`. This makes it look like there's something wrong with the error message.

elprans: OK, makes sense.

PR #711, opened 2019-09-05
Expose schema object attribute inheritance in introspection

elprans:
Every inheriting schema object introspection now has an
`inherited_attrs` property containing an array of names of attributes
that were inherited from a parent as opposed to defined locally.


1st1:
How about calling it 'inherited'? I don't like 'Attrs', but making it 'attributes' would be too long.

elprans:
`inherited` is confusing IMO since it may be attributed to the object itself and not it's attributes.  We may spell `attrs` as `fields` if that helps (and we also need to settle on that in general),

Issue #709, opened 2019-09-04
GraphQL port unable to handle covariant object types


elprans:
The following schema crashes the GraphQL schema adaptor:

```
type Bar;
type Bar2 extending Bar;

type Foo {
   link blah -> Bar
}

type Foo2 extending Foo {
   link blah -> Bar2;
}
```

```
InternalServerError: unhandled errors in a TaskGroup; 1 sub errors: (AssertionError)
 + AssertionError: Foo.bar expects type "Bar" but Foo2Type.bar provides type "Bar2".
```

PR #712, opened 2019-09-06
graphql: Fix an inheritance bug.

vpetrovykh:
Inheriting a link and changing its target type caused an error in
GraphQL.

GraphQL inheritance does not allow changing the field type, so when
reflecting pointers we must use the top-most base pointer version. It is
possible to access the additional fields (if any) of the subclassed
target type if the subtype is references by name:

```
query {
    Foo {
        bar {
            common
            ... on Bar2 {
                bar2_prop
            }
        }
    }
}
```

Closes: #709.


elprans:
> GraphQL inheritance does not allow changing the field type

Is this per the spec, or are we papering over graphql-core's behavior?

vpetrovykh:
> > GraphQL inheritance does not allow changing the field type
> 
> Is this per the spec, or are we papering over graphql-core's behavior?

The spec is a bit vague:
https://graphql.github.io/graphql-spec/June2018/#sec-Objects
https://graphql.github.io/graphql-spec/June2018/#sec-Interfaces

The spec doesn't address the specific case of inheritance like this, but it forbids to have "different" fields with the same name and it *requires* the object to have *all* the fields of its interfaces. This would seem like a contradiction in case of implementing 2 interfaces with a field of the same name and related type.

Strictly speaking, the object *is* allowed to implement a field of the same name as the interface and of a sub-type of the interface type, so under this reading it can be argued that this should be legal, whether this was intended and there are any actual GraphQL implementations that do this is another question. So maybe this should be allowed once I reimplement the GraphQL validation.

elprans:
This also doesn't seem to solve the nested view case.

```
CREATE VIEW FooView := (SELECT Foo { bar: { baz := '1' } })
```

```
            query {
                FooView {
                    bar {
                        baz
                    }
                }
            }
```

Having to explicitly specify the derived type of the nested view in the query seems barely usable.

vpetrovykh:
Views are a bit tricky in any case as they kinda generate their own types (maybe that's what the GraphQL "extended" types are for, but I'm still a bit unsure about it), since they basically introduce new fields. On the other hand we should be able to treat them as their own types fairly consistently since we happen to know at compile time whether something is a view or not.

vpetrovykh:
A more practical concern is whether other tools will work with our interpretation of the spec w.r.t. covariant object types for field targets. E.g. will GraphiQL interpret the schema correctly and have all the right auto-completion?

PR #715, opened 2019-09-13
tests: Assorted tests of schema migrations.

vpetrovykh:
Test that performing 2 or more SDL migration steps result in the same
final schema as simply applying the final SDL to a blank database.

### DIFF
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -29,6 +29,7 @@ from edb.schema import delta as s_delta
 from edb.schema import ddl as s_ddl
 from edb.schema import links as s_links
 from edb.schema import objtypes as s_objtypes
+from edb.tools import test
 
 
 class TestSchema(tb.BaseSchemaLoadTest):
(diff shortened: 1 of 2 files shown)


tests/test_schema.py
```diff
@@ -593,6 +594,35 @@ def _assert_migration_consistency(self, schema_text):
...
+        multi_migration = self.load_schema(migrations[0])
+        for i, state in enumerate(migrations[1:]):
+            multi_migration = self.run_ddl(multi_migration, f'''
+                CREATE MIGRATION m{i} TO {{
+                    {state}
+                }};
+                COMMIT MIGRATION m{i};
+            ''', 'test')
+
+        diff = s_ddl.delta_modules(single_migration, multi_migration, ['test'])
```
elprans: Is there a reason you're using `delta_modules` and not `delta_schema` here?

vpetrovykh: The `delta_schema` didn't work (was throwing exceptions about getting a Module instead of str), so I figured that I can get a bunch of tests to work this way and worry about why `delta_schema` didn't work later.
You can try changing the code to use `delta_schema` and see the error. I'm not certain whether there's something in the way I initialize the schemas that's wrong or whether it's a bug proper.

elprans: I don't see how what you're doing here differs from `_assert_migration_consistency` which uses `delta_schema` without an issue.

vpetrovykh: For some of the migration tests I have I'd also like to test what happens to actual data in the process (like whether the types, props and links get renamed or dropped, etc.). This would probably be done in a separate test class. The question is whether there's any value in doing some schema validation via introspection queries for these tests in addition to validating how they handle data? Or would that introspection be redundant in light of already having these schema tests?

vpetrovykh: > I don't see how what you're doing here differs from `_assert_migration_consistency` which uses `delta_schema` without an issue.

Yeah, I agree, at a glance I expected it to work, but it didn't. I can try poking at the error in more details to see if it's something relatively simple.

vpetrovykh: Well, the tests are already useful as is, as far as I can tell. So I'll commit this batch and build more tests on top of it so that you can get access to this and look at them too.

elprans: This test class is actually supposed to test the `GET MIGRATION` consistency.  If you want to do fine-grained migration testing, make a separate class please.

vpetrovykh: > This test class is actually supposed to test the `GET MIGRATION` consistency. If you want to do fine-grained migration testing, make a separate class please.

Well, the main reason why I stuck my tests in this class was that in addition to testing incremental SDL migrations I _also_ wanted to validate the final state w.r.t. `GET MIGRATION` consistency. (The point is to validate that the final state is probably sane *and* that the multiple SDL migrations do actually result in that same final state as a single jump.)

My question about doing more fine-grained migration testing w.r.t. data is mainly about whether the introspection queries would be redundant or not given that step-by-step migration of the same SDL would already be tested separately.

elprans: Schema dump is reliable enough, so you just need to compare the result of the step-by-step migration with the result of a direct load of the final step.

PR #717, opened 2019-09-17
Fix misc migrations issues identified in recently added tests

elprans:
Fixes:

* Skip inheritance checks in canonical delta plans;
* Fix handling of inherited links and properties that become
  required in the child type;
* Fix deparsing of constraint args;
* Fix handling of dependent views in SDL.

### DIFF
--- a/edb/schema/expr.py
+++ b/edb/schema/expr.py
@@ -170,6 +170,28 @@ class ExpressionList(typed.FrozenTypedList, type=Expression):
 
         return result
 
+    @classmethod
+    def compare_values(cls, ours, theirs, *,
+                       our_schema, their_schema, context, compcoef):
+        """See the comment in Object.compare_values""".
+        if not ours and not theirs:
+            basecoef = 1.0
+        elif (not ours or not theirs) or (len(ours) != len(theirs)):
+            basecoef = 0.2
+        else:
+            similarity = []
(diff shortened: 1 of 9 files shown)


edb/schema/expr.py
```diff
@@ -170,6 +170,27 @@ def merge_values(cls, target, sources, field_name, *, schema):
 
         return result
 
+    @classmethod
+    def compare_values(cls, ours, theirs, *,
+                       our_schema, their_schema, context, compcoef):
+        if not ours and not theirs:
```
vpetrovykh: This kind of method would benefit from a comment explaining roughly what it does. Also documenting the meaning of the arguments is helpful especially when they are similarly named: e.g. how is `ours` different from `our_schema`? If I understand correctly this is the kind of method that we would tweak if the migration doesn't recognize a rename or something, right? But in order to tweak it, some understanding of the original intent would be great.

elprans: Fair enough.

This is a common callback that is used when we do schema comparisons. `ours` and `theirs` are instances of `cls` and `our_schema` and `their_schema` are the corresponding schemas in which they're defined.  `compcoef` is whatever was specified for the field.  The method returns a coefficient of similarity of the values, from `0` to `1`.

PR #721, opened 2019-09-17
tests: Add data migration tests.

vpetrovykh:
An important feature of migrations is that data is supposed to be
preserved when possible (renaming, adding constraints, etc.). In
addition to tests verifying that multiple SDL migrations create the same
schema as a single SDL migration, data integrity tests make sure that
those migrations don't destroy data unexpectedly.

### DIFF
--- a/tests/test_edgeql_data_migration.py
+++ b/tests/test_edgeql_data_migration.py
@@ -1,0 +1,2526 @@
+#
+# This source file is part of the EdgeDB open source project.
+#
+# Copyright 2019-present MagicStack Inc. and the EdgeDB authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+import edgedb
+
+from edb.testbase import server as tb
+from edb.tools import test
+
+
+class TestEdgeQLDataMigration(tb.DDLTestCase):
+    """Test that migrations preserve data under certain circumstances.
+
+    Renaming, changing constraints, increasing cardinality should not
+    destroy data.
+
+    The test cases here use the same migrations as
+    `test_migrations_equivalence`, therefore the test numbers should
+    match for easy reference, even if it means skipping some.
+    """
+
+    _counter = 0
+
+    @property
+    def migration_name(self):
+        self._counter += 1
+        return f'm{self._counter}'
+
+    async def _migrate(self, migration):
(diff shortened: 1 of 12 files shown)


elprans:
Keep in mind that what you're really testing here is a heuristic that depends on the "difference coefficient" between the two schema objects.  It is possible to push the diff to the point where `ALTER` becomes a `DROP/CREATE`.  We might also be altering the heuristic in a way that may break tests like theses.

I think it would be better to actually test deterministic DDL commands in this case.

vpetrovykh:
Well, don't we want some "least surprise" behavior for SDL migrations, though? Like renames and refactoring should usually "just work". We want to bring SDL to a point where we reasonably trust migrating from one schema to another. We want to test that we generate reasonable DDL, not just that there may _exist_ reasonable DDL getting us from one schema to the next.

Then, in addition to these we can also test some specific DDL that would not be a default SDL migration behavior, like actually destroying and re-creating a property in what otherwise would have looked like a simple rename.

elprans:
Fair points.  Some testing of simple deltas which should obviously not cross the drop threshold is fine.

That said, the recommended migration workflow would include inspection of DDL commands and confirmation of the generated delta, so, again, let's not concentrate on testing the heuristic too much.

tests/test_edgeql_data_migration.py
```diff
@@ -0,0 +1,1345 @@
...
+                constraint max_len_value(10);
+                # add a constraint
+                constraint min_len_value(5);
+            }
+            type Base {
+                property foo -> constraint_length;
+            }
+        """
+        with self.assertRaisesRegex(
+                edgedb.InternalServerError,
```
elprans: Postgres errors should not leak, please don't write tests against them.  The error should be translated properly.

tests/test_edgeql_data_migration.py
```diff
@@ -0,0 +1,1345 @@
...
+            }],
+        )
+
+    @test.xfail('''
+        edgedb.errors.InternalServerError: column "foo" cannot be cast
+        automatically to type bigint
+
+        Ideally there should be some way of migrating regardless of
+        the potential data loss. As it stands, an extra step of column
+        dropping or renaming might be needed.
```
elprans: We can add a `USING` clause to the `SET TYPE` operation as the mechanism to implement type changes that are not implicitly castable.  But this test is a good example of a case that requires human intervention, since the data coercion expression might be non-trivial.  Replace this with a type migration that changes to an  autocoercible type.

tests/test_edgeql_data_migration.py
```diff
@@ -0,0 +1,2526 @@
...
+                'bar': {'id': data[0].bar.id}
+            }],
+        )
+
+    @test.xfail('''
+        edgedb.errors.InternalServerError: column Derived~2.bar does
+        not exist
+
+        A problem arises when a link was inherited, but then needs
+        to be "moved up" in the inheritance hierarchy. This is the
```
elprans: I usually visualize the inheritance hierarchy like an org chart, i.e the most generic type on top, so "moved up" here is confusing.

vpetrovykh: I changed the wording to refer to derived types instead.

Issue #730, opened 2019-09-24
Sequence doesn't work as a required property.


vpetrovykh:
Consider the following:
```
# self incrementing integer sequence
scalar type int_seq_t extending sequence;

type DefaultTest8 {
    required property number -> int_seq_t {
        # The number values are automatically
        # generated, and are not supposed to be
        # directly writable.
        readonly := true;
    }
}
```

Attempting to `INSERT DefaultTest8` fails, because the `number` is required, but is missing (it works fine if the property is not `required`). This is counter-intuitive.

vpetrovykh:
This example originally comes from the front-page schema (`type Issue`).

vpetrovykh:
This goes into the exact semantics of `required`. The way I understand it, it's basically a guarantee that the property will be non-empty for all objects. Whether or not that value needs to be explicitly supplied with an `INSERT` depends on whether this property has a default, and a self-incrementing scalar always has an implicit default "next value".

Incidentally, this makes me wonder, do we currently allow computables to be classified as `required`? We should be able to validate that the computable expression is guaranteed to be non-empty and providing such a guarantee may be useful. Let's say that `full_name` is a computable and gets used everywhere in forms & webpages, but it may be very nice to be able to enforce that it is _in fact_ never empty. This would mean that if one of the fields it relies upon stops being required, it would be a migration error.

vpetrovykh:
@elprans, @1st1 what are your thoughts?

elprans:
@vpetrovykh See #117 and #349

elprans:
Re the semantics of `required`, you got that right.  It guarantees that the property or link will always be there when selected.  How it got into the database is not important.

vpetrovykh:
I've added a couple more `INSERT` tests for the sequences. They have unexpected results (as far as I can tell), but additionally trying to replicate them in REPL gave me even stranger results.

1st1:
@vpetrovykh Please close issues when you commit.  You should always add a `Fixes: #NNN` to let github close things automatically.

PR #731, opened 2019-09-24
tests: Add a test of self-incrementing property.

vpetrovykh:
It should be legal to omit a required self-incrementing property when
inserting a new object.

Issue: #730

### DIFF
--- a/tests/test_edgeql_insert.py
+++ b/tests/test_edgeql_insert.py
@@ -1024,6 +1024,24 @@ class TestInsert(tb.QueryTestCase):
             ]
         )
 
+    async def test_edgeql_insert_default_05(self):
+        # Issue #730
+        await self.con.execute(r'''
+            # The 'number' property is supposed to be
+            # self-incrementing and read-only.
+            INSERT test::DefaultTest8;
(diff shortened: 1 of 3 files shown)


tests/test_edgeql_insert.py
```diff
@@ -1024,6 +1024,76 @@ class TestInsert(tb.QueryTestCase):
...
+        await self.assert_query_result(
+            r'''
+                WITH MODULE test
+                SELECT DefaultTest8.number;
+            ''',
+            {1, 2, 3}
+        )
+
+    @test.xfail('''
+        Presumably, forcing a specific sequence value in an INSERT
```
elprans: This is a wrong assumption to make.  Sequences are a standalone resource in the database and having a serial column basically means you have the implicit default of `nextval('my_sequence')`.  Inserting explicit values does not affect the underlying sequence at all.

vpetrovykh: OK, I've removed the 2 tests that `INSERT` a specified value instead of relying on the automatic sequence as they don't really test anything in particular.

PR #734, opened 2019-09-25
When replacing functions in migrations perform deletions first

elprans:
When a function is replaced in a migration with a function of the same
name, but different signature, make sure the old function is dropped
before the new function is created, otherwise the polymorphic checker
might not be happy.

### DIFF
--- a/edb/schema/ordering.py
+++ b/edb/schema/ordering.py
@@ -24,7 +24,9 @@ import collections
 from edb.common import topological
 
 from . import delta as sd
+from . import functions as s_func
 from . import inheriting
+from . import name as sn
 from . import objects as so
 from . import referencing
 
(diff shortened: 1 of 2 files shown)


edb/schema/ordering.py
```diff
@@ -294,6 +296,13 @@ def _trace_op(op, opstack, depgraph, renames, renames_r, strongrefs,
             # happen first.
             deps.add(('delete', op.classname))
 
+            if isinstance(obj, s_func.Function):
+                old_funcs = old_schema.get_functions(
+                    sn.shortname_from_fullname(op.classname),
+                    default=[])
+                for old_func in old_funcs:
+                    deps.add(('delete', old_func.get_name(old_schema)))
+
```
1st1: I'm not sure I follow.  What if you have two functions:

```
foo(a: float64) -> int64
foo(a: int64) -> int64
```

and you want to alter one of them, e.g.

```ALTER FUNCTION foo(a: int64) ...```

Wouldn't this code mean that `foo(a: float64)` variant gets deleted?

elprans: No.  This applies only to the commands *already* in the migration. We must sort `DROP FUNCTION foo` before `CREATE FUNCTION foo`, otherwise the `CREATE FUNCTION` will complain that an incompatible signature already exists.

1st1: ok.

Issue #739, opened 2019-09-29
DDL and Migrations tests


vpetrovykh:
Use SDL migrations to test some common DDL functionality and to validate "least surprising" behavior of simple SDL migrations themselves.

Tests should cover the following:
- [x] rename `type`
- [x] drop `type`
- [x] new `type` inheriting a `property`
- [x] `property` renaming
- [x] `property` type change
- [x] `property` made required
- [x] `property` dropping
- [x] `property` inheritance moved from derived to base type
- [x] `property` inheritance moved from base to derived type
- [x] new `type` inheriting a `link`
- [x] `link` renaming
- [x] `link` type change
- [x] `link` made required
- [x] `link` dropping
- [x] `link` inheritance moved from derived to base class
- [x] `link` inheritance moved from base to derived class
- [x] inherited `link` target change
- [x] `link` cardinality change
- [x] new `constraint` added to `property`
- [x] new `constraint` added to `scalar`
- [x] new `constraint` added to `link`
- [x] `constraint` dropping
- [x] change `function` call signature
- [x] change `function` return type
- [x] add new `function` with different call signature (make function polymorphic)
- [x] change `function` used in a computable
- [x] change `property` into a computable
- [x] change a computable into regular `property`
- [x] change `link` into a computable
- [x] change a computable into regular `link`
- [x] replace a `type` with an emulating `view`
- [x] new `link property`
- [x] `link property` renaming
- [x] `link property` type change
- [x] `link property` dropping
- [x] `link property` inheritance moved from derived to base type
- [x] `link property` inheritance moved from base to derived type
- [x] `view` property type change
- [x] `view` link type change
- [x] `view` property "renaming"
- [x] `view` link "renaming"
- [x] add/remove `annotation`
- [x] add/remove `inheritable annotation`
- [x] change between `inheritable annotation` and regular `annotation`
- [x] add/remove `index`
- [x] `collection` type as property
- [x] `collection` type as function signature and return

vpetrovykh:
Most of these are covered by #737 
I noticed a few areas missing as I was compiling the list, though. I'll make a small PR to add tests for them.

1st1:
Annotations?

1st1:
Indexes?

elprans:
Add tests for collection types in schema, i.e. collection function parameters, return types and property types.

PR #744, opened 2019-10-02
tests: Add more migration tests.

vpetrovykh:
New tests cover:
- changing of link target
- converting between regular and computable links
- introducing polymorphic functions
- changing links and properties in schema views
- adding/removing annotations
- annotation inheritance
- adding/removing indexes

Issue #739.


vpetrovykh:
I think there are a few of the bugs here that I could fix. It might make more sense to have separate PRs for the fixes.

PR #747, opened 2019-10-03
Fix an issue with incorrect deltas for functions.

### LINKED ISSUE #746: SDL with polymorphic functions is broken
Currently, if a polymorphic function is introduced via SDL the migration then produces only _one_ of the variants presented in SDL.

Example of problematic SDL:
```edgeql
function hello12(a: int64) -> str
    from edgeql $$
        SELECT 'hello' ++ <str>a
    $$;

function hello12(a: str) -> str
    from edgeql $$
        SELECT 'hello' ++ a
    $$;
```

vpetrovykh:
Functions are defined by their name and their call signatures. Both need
to be taken into account when producing a delta from SDL, otherwise
polymorphic functions end up with only one version.

Fixes #746.

### DIFF
--- a/edb/edgeql/declarative.py
+++ b/edb/edgeql/declarative.py
@@ -338,6 +338,18 @@ def trace_View(
 
 
 @trace_dependencies.register
+def trace_Function(node: qlast.CreateFunction, *, ctx: DepTraceContext):
+    # Functions are defined by their name + call signature, so we need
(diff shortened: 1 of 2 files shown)


1st1:
@vpetrovykh merge this as is and then fix https://github.com/edgedb/edgedb/issues/751

PR #749, opened 2019-10-03
Fix a view migration "renaming" a view property.

### LINKED ISSUE #748: Migration that renames a `VIEW` property are broken
Currently a migration that renames a `property` of a `VIEW` fails due to an `InternalServerError`.
The `test_edgeql_migration_37` from #744 is an example of this.

vpetrovykh:
View properties and links do not actually exist as columns in real
tables, so when migrations that rename them are reflected into SQL they
must not produce any changes to the tables.

Fixes #748.

### DIFF
--- a/edb/pgsql/delta.py
+++ b/edb/pgsql/delta.py
@@ -2097,50 +2097,74 @@ class PointerMetaCommand(MetaCommand, sd.ObjectCommand,
 
     def rename_pointer(self, pointer, schema, context, old_name, new_name):
         if context:
+            # before proceeding with renaming, make sure that this is
+            # not a computable
+            pointer = self.get_object(schema, context)
+            source = pointer.get_source(schema)
+            if source is not None:
+                is_computable = source.get_is_derived(schema)
+
+                # potentially this is a link property, so source may
+                # be a link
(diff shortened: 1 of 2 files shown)


edb/pgsql/delta.py
```diff
@@ -2074,50 +2074,74 @@ def get_columns(cls, pointer, schema, default=None):
...
 
-            if host and old_name != new_name:
-                if (new_name.endswith('std::source') and
-                        not host.scls.generic(schema)):
-                    pass
-                else:
-                    old_col_name = common.edgedb_name_to_pg_name(old_name.name)
-                    new_col_name = common.edgedb_name_to_pg_name(new_name.name)
+                if host and old_name != new_name:
+                    if (new_name.endswith('std::source') and
```
1st1: Please see if `==` works.  `endswith` isn't a reliable check.

vpetrovykh: Apparently it works, at least all the tests pass.

1st1:
Can you look for other places with endswith? if there are any please fix them in separate prs

vpetrovykh:
There does not appear to be any more suspicious `endswith` usage.

PR #750, opened 2019-10-03
tests: Add a couple of unusual views tests.

vpetrovykh:
These views are creating computable links with computable link
properties. Which they then migrate to rename or change property target
type.

Issue #739.


vpetrovykh:
The `view` here introduces a whole new computable `link` with its own `link property`. Although I simply used constant values for them, the more "realistic" scenario for something like this would be when there is an explicit intermediate object with properties that plays this role and we want to emulate _actual_ `link` and `link properties` instead via a `view`.

Issue #693, opened 2019-08-16
Set up Mypy


ambv:
Let's start with:
* common/ast
* edgeql/compiler

elprans:
The majority of typing errors in the compiler have been fixed in e7b95b8a233 and cac8d93fb213.   Next big areas to clean up are the `ir` and `schema` packages.

1st1:
I'm closing this one. We'll open new typing issues as we go.

PR #752, opened 2019-10-03
Track type coverage of the codebase with tests

ambv:
Tracking per function.  Also allows for tracking per line within functions but
that's not that important and way more noisy as it would require changing tests
on almost every change to every file, very annoying.  This is also why we're
not using Mypy's builtin per-line coverage reporting here.

Partially-typed function signatures are counted as untyped.  This is to
differentiate between old "documentation-style" annotations and actual
annotations covered by Mypy.

Those tests are set up to also fail when the actual percentage is *higher* than
the expected percentage.  This is so that we can raise the expectation in time
and keep the bar.

This is part of #693.

### DIFF
--- a/tests/test_type_coverage.py
+++ b/tests/test_type_coverage.py
@@ -1,0 +1,265 @@
+from __future__ import annotations
+
+import ast
+from dataclasses import dataclass
+import os
+from pathlib import Path
+import tokenize
+from typing import List, Union
+import unittest
+
+import edb
+
+EDB_DIR = Path(edb.__file__).parent
+
+
+@dataclass
+class TypeCoverage(ast.NodeVisitor):
+    typed_functions: int
+    untyped_functions: int
+    typed_lines: int
+    untyped_lines: int
+
+    def __init__(self) -> None:
+        self.typed_functions = 0
+        self.untyped_functions = 0
+        self.typed_lines = 0
+        self.untyped_lines = 0
+        self._classdefs: List[ast.ClassDef] = []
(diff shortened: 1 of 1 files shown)


1st1:
Maybe we should flip the check to `>=`?  Or you need to update it so that we can merge it :)

elprans:
Yeah, maybe allowing 10-15% jitter to the positive side.  No need to fix the numbers all the time, but there's still a stick to force the update now an then.

1st1:
I'd just use `>=`.  We can bump the number until we can enable mypy strict on all code and then we don't need this test, right?

ambv:
Sure, mypy strict would be perfect but that's a bit optimistic in the short-to-medium term :-)

Alright, I'll change to >=.

1st1:
@ambv Click the green button once the CI is ready.

BTW, please squash your commits into one when you do.  In general we want to keep commits separate only when each of them is logically complete.  @elprans will update the handbook right now.

ambv:
We can disable other forms of merging if squashing is something you always want to do.

elprans:
We don't want to *always* do a squash.  In fact, it is desirable to keep logically separate changes separate.  But small follow-up cleanups and fixes to the main commit should be squashed to reduce the noise in the commit log.

1st1:
> We can disable other forms of merging if squashing is something you always want to do.

If explaining our commit policies proves to be hard we can switch to "one commit per PR" rule and enforce it.  But for now I think we should give the core devs freedom to decide on their own.

Issue #755, opened 2019-10-03
Annotation migration fails intermittently


vpetrovykh:
The test `test_migrations_equivalence_annotation_01` fails intermittently.

I'll record one of the tracebacks here, just in case:
```
Traceback (most recent call last):
  File "/home/victor/dev/magicstack/edgedb/tests/test_schema.py", line 2084, in test_migrations_equivalence_annotation_01
    """])
  File "/home/victor/dev/magicstack/edgedb/tests/test_schema.py", line 603, in _assert_migration_equivalence
    self._assert_migration_consistency(migrations[-1])
  File "/home/victor/dev/magicstack/edgedb/tests/test_schema.py", line 588, in _assert_migration_consistency
    diff = s_ddl.delta_schemas(baseline_schema, test_schema)
  File "/home/victor/dev/magicstack/edgedb/edb/schema/ddl.py", line 87, in delta_schemas
    old, new, old_schema=schema2, new_schema=schema1))
  File "/home/victor/dev/magicstack/edgedb/edb/schema/objects.py", line 1143, in delta_sets
    their_schema=old_schema)
  File "/home/victor/dev/magicstack/edgedb/edb/schema/objects.py", line 778, in compare
    context=context)
  File "/home/victor/dev/magicstack/edgedb/edb/schema/objects.py", line 796, in compare_field_value
    compcoef=field.compcoef or 0.5)
  File "/home/victor/dev/magicstack/edgedb/edb/schema/objects.py", line 1466, in compare_values
    their_schema=their_schema, context=context))
  File "/home/victor/dev/magicstack/edgedb/edb/schema/objects.py", line 770, in compare
    their_value = other.get_field_value(their_schema, field_name)
  File "/home/victor/dev/magicstack/edgedb/edb/schema/objects.py", line 642, in get_field_value
    schema, field_name, allow_default=allow_default)
  File "/home/victor/dev/magicstack/edgedb/edb/schema/objects.py", line 635, in _get_schema_field_value
    f'{self!r} object has no value for field {field_name!r}')
edb.schema.objects.FieldValueNotFoundError: <AnnotationValue: at 0x7f27da85bf60> object has no value for field 'annotation'
```

PR #756, opened 2019-10-04
Prevent incorrect inheritance of non-inheritable annotations

elprans:
When propagating annotations down the inheritance chain, make sure
the annotation is actually inheritable, otherwise the schema will
end up with a partially-inherited annotation in a descendant object.
This really isn't specific to annotations, it's just that only the
annotations may be declared as non-inheritable currently.

Fixes: #755

### DIFF
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -381,7 +381,7 @@ _123456789_123456789_123456789 -> str
             })
         )
 
-    def test_schema_annotation_inheritance(self):
+    def test_schema_annotation_inheritance_01(self):
         schema = self.load_schema("""
             abstract annotation noninh;
             abstract inheritable annotation inh;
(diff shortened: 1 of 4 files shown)


tests/test_schema.py
```diff
@@ -406,6 +406,30 @@ def test_schema_annotation_inheritance(self):
         self.assertEqual(
             Object2.get_annotation(schema, 'test::inh'), 'inherit me')
 
+    def test_schema_annotation_inheritance_02(self):
```
vpetrovykh: Is there a reason why we need both of the tests in this PR? They seem to be kinda doing the same thing: checking the schema for annotations on the derived type. What is the thing that one of them tests that the other doesn't?

elprans: They are largely equivalent, the difference is in how we check the effects.

Issue #758, opened 2019-10-04
SDL and DDL for INDEX are not equivalent


vpetrovykh:
Depending on whether an `INDEX` is created via SDL or DDL it will end up with a different "name". This results in a "schema  difference" where no difference is expected.

See `test_migrations_equivalence_index_01`.

vpetrovykh:
This bug highlights an issue with using the expression to identify an `INDEX` or a `CONSTRAINT`. The expression always exists in some default module context and this context must be preserved at the moment of creation. This can be done by normalizing the expression and explicitly injecting `WITH MODULE blah` into it. However, it's useful to also preserve the original (shorter) expression for purposes of rendering and user-friendly identification. This can be stored in a special system `ANNOTATION`, which also makes this data introspectable.

PR #761, opened 2019-10-08
Index

vpetrovykh:
Add a new ANNOTATION `schema::system` for storing details used by
EdgeDB. Users should not change this annotation directly.

Record the original INDEX expression text in the `system` annotation for
that index. This way it is possible to use that to display a more
human-friendly expression during introspection as well. It is also used
to identify the INDEX itself in SDL and DDL.

Update the syntax so that index annotations are allowed. Also, disallow
indexes as top-level statements in DDL or SDL. They must be nested
inside a type or a link statement.
Also update the docs.

Fixes: #758

### DIFF
--- a/docs/datamodel/annotations.rst
+++ b/docs/datamodel/annotations.rst
@@ -19,6 +19,9 @@ The following are the annotations which can be set on any schema item:
 - ``title``
 - ``description``
 
+There's also a special ``schema::system`` annotation that should never
+be altered directly by the user, but can be useful in introspection.
+
 For example, consider the following declaration:
 
 .. code-block:: sdl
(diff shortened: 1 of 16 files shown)


elprans:
`schema::system` -- not in love with such a nondescript name, I'd use the actual meaning of the value being held: `schema::origexpr`.

vpetrovykh:
OK, I'll retract this PR, split it into separate pieces:
1) Adding annotations to indexes and tightening index syntax + docs
2) Creating a special string *field* `origexpr` on indexes for storing the original expression. So all the special annotation logic will instead be working with that field on index.

PR #764, opened 2019-10-11
edgeql: Store original index expression in a special field.

vpetrovykh:
Record the original INDEX expression text in the "origexpr" field for
that index. This way it is possible to use that to display a more
human-friendly expression during introspection as well. It is also used
to identify the INDEX itself in SDL and DDL.

Fixes: #758

### DIFF
--- a/edb/schema/delta.py
+++ b/edb/schema/delta.py
@@ -1371,11 +1371,6 @@ class AlterObjectProperty(Command):
             (value is None or
                 (isinstance(value, collections.abc.Container) and not value))
 
-        old_value_empty = \
-            (self.old_value is None or
-                (isinstance(self.old_value, collections.abc.Container) and
-                    not self.old_value))
-
         parent_ctx = context.current()
         parent_op = parent_ctx.op
         field = parent_op.get_schema_metaclass().get_field(self.property)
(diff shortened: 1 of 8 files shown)


edb/schema/delta.py
```diff
@@ -1399,6 +1399,8 @@ def _get_ast(self, schema, context):
             value = qlast.Tuple(elements=[
                 qlast.BaseConstant.from_python(el) for el in value
             ])
+        elif value is None:
```
elprans: The `if new_value_empty` check above should handle this.  Remove ` and old_value_empty`, it's wrong.

edb/schema/indexes.py
```diff
@@ -113,14 +118,25 @@ def _classname_from_ast(cls, schema, astnode, context):
     @classmethod
     def _classname_quals_from_ast(cls, schema, astnode, base_name,
                                   referrer_name, context):
-        subject = schema.get(referrer_name, None)
-        if subject is None:
-            return ()
+        expr_text = None
+
+        # check if we have annotation with original text
```
elprans: Outdated comment here.

vpetrovykh: Update the commend.

edb/schema/delta.py
vpetrovykh: OK, I've removed the `value` check in favor of `new_value_empty`. As a result of this `old_value_empty` is no longer used anywhere so I dropped it.

PR #765, opened 2019-10-15
Allow indexes to be annotated.

vpetrovykh:
Adding annotations to index is now possible.


vpetrovykh:
I've factored out the commit that just cleans up the unused AST nodes. It happens to increase the `test_type_coverage_edgeql` coverage percentage due to all the deletions in such a way that the next commit decreases the coverage. Although the total still increases for the entire PR.

PR #767, opened 2019-10-17
Validate that we can trace all the expression AST nodes.

vpetrovykh:
Edgeql AST nodes now have a `__abstract_node__` field which marks nodes
that should never be instantiated.

Refactor EdgeQL AST nodes hierarchy so that things that aren't valid expressions
no longer inherit from Expr. Introduce a base class `Command` from which
all EdgeQL commands inherit. `Statement` is a `Command` that is also valid
expression (i.e. EdgeQL query). Clean up a few unused AST nodes.

Add a test that checks that all eligible expression AST nodes are
traceable. Add tracing for several missing AST nodes.

Add return type annotations to tracer.py and adjust the coverage
percentage.

### DIFF
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -751,6 +751,46 @@ class TestGetMigration(tb.BaseSchemaLoadTest):
 
         self._assert_migration_consistency(schema)
 
+    def test_get_migration_04(self):
+        # validate that we can trace DETACHED
+        schema = r'''
+            type Foo {
+                property bar -> int64;
+            };
+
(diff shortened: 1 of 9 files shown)


tests/test_schema.py
```diff
@@ -751,6 +751,46 @@ def test_get_migration_03(self):
 
         self._assert_migration_consistency(schema)
 
+    def test_get_migration_04(self):
+        # validate that we can trace DETACED
```
elprans: typo

vpetrovykh: Fixed

PR #773, opened 2019-10-18
[WIP] Enable type checking of edb.schema

ambv:
Each commit is a separate file covered by MonkeyType. I haven't reformatted it to satisfy flake8 so that the diff is as easy to review as possible.

MonkeyType covering new functions actually increases the number of Mypy errors for a while. I will be fixing those new errors in subsequent commits.

[work in progress]

### DIFF
--- a/edb/schema/annos.py
+++ b/edb/schema/annos.py
@@ -31,6 +31,11 @@ from . import objects as so
 from . import utils
 
 
+if typing.TYPE_CHECKING:
+    from . import objtypes as s_objtypes
+    from . import schema as s_schema
+
+
 class Annotation(inheriting.InheritingObject):
     # Annotations cannot be renamed, so make sure the name
     # has low compcoef.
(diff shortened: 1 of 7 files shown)


edb/schema/annos.py
```diff
@@ -31,6 +31,11 @@
 from . import utils
 
 
+if typing.TYPE_CHECKING:
```
1st1: `from typing import *  # NoQA`?

ambv: I have a script to refactor all those at a later point. Same what I used for object to typing.Any yesterday. I can apply that in this pull request in the last commit. For now I'd like the clarity of which Tuple is which, and so on.

1st1: Sure!

1st1: Sure!

1st1:
Since this is a WIP I'm not sure if you expect a review or comments at this point, but so far so good, all types appear to be correct!

ambv:
Just wanted to show some code since I've been poking at all this for a while with no visible effect so far. I gave up on fully automated coverage but this is pretty good already. We'll upstream this to MonkeyType.

ambv:
Aborted, will do separate PRs per file.

PR #779, opened 2019-10-22
Fix some inconsistencies in ir.ast

elprans:
`ir.ast.Set.expr` is defined as `ir.ast.Base`, which is wrong, since it
must always be an instance of `ir.ast.Expr`.  Fix this, and the
relevant typing annotations in the EdgeQL frontend.

This also fixes a couple of remaining typing errors in
`edb.edgeql.compiler`, so enable mypy checks on it.

### DIFF
--- a/edb/edgeql/compiler/expr.py
+++ b/edb/edgeql/compiler/expr.py
@@ -121,7 +121,11 @@ def compile_DetachedExpr(
 
 
 @dispatch.compile.register(qlast.Set)
-def compile_Set(expr: qlast.Set, *, ctx: context.ContextLevel) -> irast.Base:
+def compile_Set(
+    expr: qlast.Set,
+    *,
+    ctx: context.ContextLevel
+) -> typing.Union[irast.Set, irast.Expr]:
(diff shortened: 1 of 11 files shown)


edb/edgeql/compiler/expr.py
```diff
@@ -121,7 +121,11 @@ def compile_DetachedExpr(
 
 
 @dispatch.compile.register(qlast.Set)
-def compile_Set(expr: qlast.Set, *, ctx: context.ContextLevel) -> irast.Base:
+def compile_Set(
+    expr: qlast.Set,
+    *,
+    ctx: context.ContextLevel
+) -> typing.Union[irast.Set, irast.Expr]:
```
vpetrovykh: After all the changes, I'm not sure that this ever returns anything other than a `irast.Set`. Does it?

elprans: The bigunion branch calls `dispatch.compile` with this return type.

vpetrovykh: I see.

PR #782, opened 2019-10-23
Add schema.ddl.ddl_text_from_schema(schema)

elprans:
The new `ddl_text_from_schema()` helper returns valid DDL text
for a specified schema, and is, essentially, an implementation of a
schema dump.  It is now used as another check for the migration
unit tests.

### DIFF
--- a/edb/schema/ddl.py
+++ b/edb/schema/ddl.py
@@ -41,6 +41,7 @@ from . import expr as s_expr
 from . import objtypes  # NOQA
 from . import constraints  # NOQA
 from . import functions  # NOQA
+from . import migrations
 from . import operators  # NOQA
 from . import indexes  # NOQA
 from . import links  # NOQA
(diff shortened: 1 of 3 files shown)


edb/server/compiler/compiler.py
```diff
@@ -419,7 +419,8 @@ def _compile_and_apply_migration_command(
                 return self._compile_and_apply_ddl_command(ctx, ddl_plan)
 
             elif isinstance(cmd, s_migrations.GetMigration):
-                ddl_text = s_ddl.ddl_text_from_migration(schema, migration)
+                delta = migration.get_delta(schema)
+                ddl_text = s_ddl.ddl_text_from_delta(schema, delta)
```
1st1: This actually reminds me about storing migrations in the DB and then we can drop the requirement to always do them in a transaction, right?

elprans: Possibly. We'll need to review the design once we move on to making the proper migration cli.

elprans:
The test failure is weird and intermittent.  Investigating.

elprans:
The failure turned out to be due to invalid assumptions about the state of the schema.  Intermittance due to devmode cache.

PR #783, opened 2019-10-23
Rewrite "import typing" into "from typing import *"

ambv:
There's a script based on Black that does this in edb/tools/rewrite_typing.py.

The script checks for shadowing of typing by names local to the file in which case it skips such a file. The script also skips files that say "import typing  # NoQA" so we can force keeping nominal imports for some files where it's needed (like the flake8 plugin).


edb/ir/scopetree.py
```diff
@@ -209,15 +209,15 @@ def descendant_namespaces(self) -> typing.Set[str]:
...
+    def unfenced_descendants(self) -> Iterator['ScopeTreeNode']:
         """An iterator of node's unfenced descendants including self."""
         yield self
         for child in tuple(self.children):
             if not child.fenced:
                 yield from child.unfenced_descendants
 
     @property
-    def strict_unfenced_descendants(self) -> typing.Iterator['ScopeTreeNode']:
+    def strict_unfenced_descendants(self) -> Iterator['ScopeTreeNode']:
```
1st1: We can probably also fix things like `Iterator['ScopeTreeNode']` to `Iterator[ScopeTreeNode]` if that's easy to handle in the script you have; and if not this can wait/be done manually.

1st1:
We'll also need to write a script to move `from typing import *` right under `from __future__ import annotations` just for aesthetics. That can be done with just iterating over lines in file without lib2to3; we can do it later (or I can do that myself)

ambv:
This fails a few code quality tests, will fix now.

1st1:
> This fails a few code quality tests, will fix now.

Yeah, I approved because in general it looks great. We'll need to fix the CI anyways :)

edb/tools/rewrite_typing.py
```diff
@@ -0,0 +1,165 @@
+"""
```
elprans: Missing copyright boilerplate

edb/tools/rewrite_typing.py
```diff
@@ -0,0 +1,165 @@
...
+
+from __future__ import annotations
+
+import os
+from pathlib import Path
+import sys
+from typing import *  # NoQA
+
+from black import lib2to3_parse, Visitor, Leaf, Node, LN, syms
```
elprans: Add black to dev dependencies?

ambv:
The star-import is also out of place now I guess but we can and should fix it with isort or something instead of potato sorting by hand.

edb/tools/rewrite_typing.py
ambv: We should lint it. Will add.

edb/ir/scopetree.py
ambv: OK, replaced all string annotations with proper names.

1st1: Awesome!

PR #786, opened 2019-10-23
Fix #785: Name 'qlast.Stmt' is not defined

### LINKED ISSUE #785: Name 'qlast.Stmt' is not defined
mypy 0.740 reports as of be705ce77d6a5f6d66a94691b4f7d0fb555b34b0 (possibly sooner):

edb/edgeql/compiler/stmt.py:488: error: Name 'qlast.Stmt' is not defined

ambv:


1st1:
Just fix it in #783

PR #788, opened 2019-10-23
Introduce flake8-bugbear

ambv:
We mostly care about two warnings:
- B007: not using a control loop variable inside its own loop
- B006: using mutable data structures as argument defaults

### DIFF
--- a/edb/common/markup/__init__.py
+++ b/edb/common/markup/__init__.py
@@ -33,7 +33,7 @@ class MarkupCapableMeta(type):
     def __new__(mcls, name, bases, dct):
         cls = super().__new__(mcls, name, bases, dct)
         if 'as_markup' in dct:
-            serializer.serializer.register(cls)(getattr(cls, 'as_markup'))
+            serializer.serializer.register(cls)(cls.as_markup)
         return cls
 
 
--- a/edb/common/markup/renderers/terminal.py
+++ b/edb/common/markup/renderers/terminal.py
@@ -187,7 +187,7 @@ class Buffer:
                     result.append(item[2].apply(item[1]))
                 offset += len(item[1])
             else:
-                assert False
+                raise AssertionError(f"Unexpected element: {el}")
 
         return ''.join(result)
 
--- a/edb/pgsql/compiler/relctx.py
+++ b/edb/pgsql/compiler/relctx.py
@@ -959,10 +959,13 @@ def range_for_pointer(
 
 def rvar_for_rel(
         rel: Union[pgast.BaseRelation, pgast.CommonTableExpr], *,
-        lateral: bool=False, colnames: List[str]=[],
+        lateral: bool=False, colnames: Optional[List[str]]=None,
         ctx: context.CompilerContextLevel) -> pgast.PathRangeVar:
 
     rvar: pgast.PathRangeVar
+
+    if colnames is None:
+        colnames = []
 
     if isinstance(rel, pgast.Query):
         alias = ctx.env.aliases.get(rel.name or 'q')
--- a/edb/schema/delta.py
+++ b/edb/schema/delta.py
@@ -226,7 +226,7 @@ class Command(struct.MixedStruct, metaclass=CommandMeta):
     def after(self, command):
         if isinstance(command, CommandGroup):
             for op in command:
-                self.before_ops.add(command)
+                self.before_ops.add(op)
         else:
             self.before_ops.add(command)
 
--- a/edb/server/cluster.py
+++ b/edb/server/cluster.py
@@ -155,7 +155,7 @@ class Cluster:
 
         return edgedb.connect(**connect_args)
 
-    def init(self, *, server_settings={}):
+    def init(self, *, server_settings=None):
         cluster_status = self.get_status()
 
         if not cluster_status.startswith('not-initialized'):
--- a/edb/testbase/server.py
+++ b/edb/testbase/server.py
@@ -186,7 +186,9 @@ class TestCase(unittest.TestCase, metaclass=TestCaseMeta):
 _default_cluster = None
 
 
-def _init_cluster(data_dir=None, *, cleanup_atexit=True, init_settings={}):
+def _init_cluster(data_dir=None, *, cleanup_atexit=True, init_settings=None):
+    if init_settings is None:
+        init_settings = {}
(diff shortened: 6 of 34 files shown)


edb/common/markup/__init__.py
```diff
@@ -33,7 +33,7 @@ class MarkupCapableMeta(type):
     def __new__(mcls, name, bases, dct):
         cls = super().__new__(mcls, name, bases, dct)
         if 'as_markup' in dct:
-            serializer.serializer.register(cls)(getattr(cls, 'as_markup'))
```
ambv: Explanation: getattr() with a string literal and no third argument is the same as a regular attribute get.

edb/common/markup/renderers/terminal.py
```diff
@@ -187,7 +187,7 @@ def check_folded_fit(pos, data, width):
                     result.append(item[2].apply(item[1]))
                 offset += len(item[1])
             else:
-                assert False
```
ambv: Explanation: this assert lacks context and would be turned off by -O.

pyproject.toml
```diff
@@ -0,0 +1,3 @@
+[tool.black]
+line-length = 79
+target-version = ['py37', 'py38']
```
elprans: Can we also disable string normalization?  Primarily because we want black to leave string prefixes alone. `r` and `R` mean different things for MagicPython.

setup.py
```diff
@@ -68,6 +68,7 @@
     'test': [
         'black~=19.3b0',
         'flake8~=3.7.5',
+        'flake8~=19.8.0',
```
elprans: You probably meant `flake8-bugbear` here

edb/pgsql/compiler/relctx.py
```diff
@@ -959,11 +959,14 @@ def range_for_pointer(
 
 def rvar_for_rel(
         rel: Union[pgast.BaseRelation, pgast.CommonTableExpr], *,
-        lateral: bool=False, colnames: List[str]=[],
+        lateral: bool=False, colnames: Optional[List[str]]=None,
```
ambv: I realize this goes contrary to some advice I recently gave. To reiterate that advice first: avoid Optionals if possible since None is viral. For instance, you should set arguments to related empty containers and be done with it.

However, this example connects two hairy edge cases of Python:
- setting mutable data structures as argument defaults; and
- the fact that a single string is a valid `Sequence[str]` (because strings are iterable).

We don't want to have a list as a default value because if a single call mutates it, the other calls will receive the mutated version.

On the other hand, we can't set a sensible immutable default instead because changing `List[str]` to `Sequence[str]` would undermine our ability to type check one of the most common errors: putting a single element in place of a list of elements.

So in the end we made the list optional :(

edb/schema/delta.py
```diff
@@ -226,7 +226,7 @@ def has_subcommands(self):
     def after(self, command):
         if isinstance(command, CommandGroup):
             for op in command:
-                self.before_ops.add(command)
+                self.before_ops.add(op)
```
ambv: nyan

edb/server/cluster.py
```diff
@@ -155,7 +155,7 @@ def connect(self, **kwargs):
 
         return edgedb.connect(**connect_args)
 
-    def init(self, *, server_settings={}):
+    def init(self, *, server_settings=None):
```
ambv: Since `server_settings` isn't used in the body of this method at all, it can very well be None with no supporting code to transform this into a dictionary.

edb/testbase/server.py
```diff
@@ -395,8 +397,8 @@ def _assert_list_shape(data, shape):
 
             shape_iter = _list_shape_iter(shape)
 
-            i = 0
-            for i, el in enumerate(data):
+            _data_count = 0
```
ambv: I renamed the `i` since it's used outside of the loop and has a purpose.

pyproject.toml
ambv: We can, sure. Care to elaborate how they're different? Python doesn't treat them differently.

setup.py
ambv: Truth.

pyproject.toml
elprans: MagicPython highlights `r"foo"` as a regexp, and `R"foo"` as an uninterpreted raw string, since the majority of uses of raw strings are regexps.

edb/pgsql/compiler/relctx.py
elprans: Is there a linter for this? I.e. "don't use Sequence[str] as annotation"?

tests/test_edgeql_expressions.py
```diff
@@ -893,7 +893,7 @@ class TestExpressions(tb.QueryTestCase):
 
         for left in uuids[:-1]:
             for right in uuids[1:]:
-                for op, not_op in [('>=', '<'), ('<=', '>')]:
+                for op, _not_op in [('>=', '<'), ('<=', '>')]:
```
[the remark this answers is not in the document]
elprans: Copy-paste or oversight.  @vpetrovykh did you intend to use the inverted operator for these tests like you did in other instances?

edb/pgsql/compiler/relctx.py
ambv: Well, it's a valid annotation. It's just much less useful than people think. We might write our own linter for this if you think it's important.

pyproject.toml
elprans: It it's possible to tell black to only normalize the quotes, and not prefixes, that's fine too.

ambv:
We mostly care about two warnings:
- B007: not using a control loop variable inside its own loop
- B006: using mutable data structures as argument defaults

tests/test_edgeql_expressions.py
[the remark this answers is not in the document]
vpetrovykh: That's a copy-pasting artifact + oversight. So the intent was to make sure that the ordering works out the same way as for str. We already validated symmetry of operators for str by this point, so it's sufficient to test that for all ordering operators bytes and uuids "similar" to strings (trivially castable from those strings) behave the same way. So there's no need for a separate grouping of `not_op`, but instead all 4 operators should just be iterated over by `op`.

```python
for op in ('>=', '<', '<=', '>'):
    ...
```

PR #792, opened 2019-10-24
Fix problems with test teardown when running `edb test`

elprans:
This is yet another *attempt* to convince our multiprocessed
unittest to actually do final test teardown properly.

Both `unittest` and `multiprocessing` are at fault here, with
multiple issues to address.

Problem ONE: same `tearDownClass` called multiple times.  This is
a result of misunderstanding of what `unittest.TestSuite` actually
is.  Our current teardown fix iterates over the registered
`TestSuite` instances and calls `_tearDownPreviousClass` on them.
This is wrong, because `TestSuite` is essentially just a bunch
of methods that mutate a *single result instance*, so they're really
basically static methods of `TestResult`, and so must be called
**once**, and not per `TestSuite` instance as is now.

Problem TWO: `multiprocessing.Pool` worker finalization does not
work.  When running with `-jN`, where `N` is greater than one,
we use the standard process pool to run tests in parallel.
Currently, the teardown portion is implemented via
`multiprocessing.util.Finalize`, which is a complicated proxy
for `atexit()`.  Using `atexit()` for finalization is a fraught
undertaking, since there's no guarantee if any object would be
alive at this point (hence the `Finalize` hacks with its
`weakref` use).

A sane approach is to actually provide a counterpart of
the `initialize` argument to the Pool constructor to ensure
correct worker cleanup and call it just after the worker
loop concludes.  The only way to do this currently is to
monkeypatch `multiprocessing.pool.worker()`.

But this is not all.  Since `multiprocessing` does not actually
care about correct worker funalization, it basically just
calls `Process.terminate()` on them as soon as `Pool.terminate()`
is called, without trying to be nice and `join()` them properly.
Hence another monkeypatch to convince it to do so.

### DIFF
--- a/edb/tools/test/runner.py
+++ b/edb/tools/test/runner.py
@@ -18,6 +18,7 @@
 
 
 from __future__ import annotations
+from typing import *  # NoQA
 
 import collections.abc
 import enum
@@ -36,7 +37,6 @@ import tempfile
 import threading
 import time
 import types
-from typing import *  # NoQA
 import unittest.case
 import unittest.result
 import unittest.runner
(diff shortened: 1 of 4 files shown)


edb/tools/test/runner.py
```diff
@@ -124,20 +126,18 @@ def _run(self, test, result):
...
-        suite = suite_cache.cache[case_class]
-    except KeyError:
-        suite = suite_cache.cache[case_class] = StreamingTestSuite()
-
-    if isinstance(case, collections.abc.Iterable):
-        # Got a test suite
-        for test in case:
-            suite.run(test, result)
+def _run_test(workload):
+    if workload is teardown_suite:
```
1st1: `is`? Maybe using `==` is safer in this case?

elprans: Ah, this is actually unused. Removed.

PR #797, opened 2019-10-30
tests: Add collections to migration tests.

vpetrovykh:
Test that at least the basic SDL migrations and some basic DDL for
working with collections work. The scope of the tests largely invovles
arrays and tuples in properties and schema views.

Fix assorted issues that prevented collections from being recorded in
the schema or being incorrectly altered.

Issue: #739


edb/edgeql/declarative.py
```diff
@@ -489,6 +485,33 @@ def _register_item(
...
+    # If we have any type ops, get a flat list of their operands.
+    targets = qlast.get_targets(expr)
+    for target in targets:
+        # We care about subtypes dependencies, because
+        # they can either be custom scalars or illegal
+        # ObjectTypes (then error message will depend on
+        # dependency tracing)
+        if target.subtypes:
+            for subtype in target.subtypes:
+                # Recurse!
```
1st1: Intense!

edb/pgsql/delta.py
```diff
@@ -2324,7 +2324,7 @@ def adjust_pointer_storage(self, pointer, schema, orig_schema, context):
 
                         alter_type = dbops.AlterTableAlterColumnType(
                             old_ptr_stor_info.column_name,
-                            common.qname(*new_type))
+                            common.quote_type(*new_type))
```
elprans: Remove the unpacking here. `quote_type` takes a tuple as its argument

vpetrovykh: Done

PR #800, opened 2019-10-30
Introduce edb.common.checked

ambv:
Those data structures are supposed to replace edb.common.typed.  These are the main selling points:
* they're fully type-annotated themselves;
* they are generic data structures enabling better annotating of other code; 
* they use *that same generic syntax* to perform runtime checking; and
* the parametrized types don't have to be pre-declared on module level for the type to be picklable.

Example usage:
```pycon
>>> from edb.common.checked import CheckedList
>>> cl = CheckedList[int]((1, 2, 3))
>>> cl.append(4)
>>> cl
edb.common.checked.CheckedList[int]([1, 2, 3, 4])
>>> cl.append("s")
...
ValueError: <class 'edb.common.checked.CheckedList[int]'> accepts only values of type <class 'int'>, got <class 'str'>
```

Other remarks:
* CheckedSet and FrozenCheckedSet implement all methods on the builtin data structures, unlike TypedSet and FrozenTypedSet.
* There is no separate OrderedCheckedDict since CheckedDict is ordered already.
* Just like in edb.common.typed there is no FrozenCheckedDict, we might add that later if needed. 
* edb.common.checked has a 50% test coverage whereas edb.common.typed had 36%.  In both cases the coverage is underreported because the test module is pre-imported before
coverage started.
* There is no need for separate parametrized classes like `StrList`, `IntSet` anymore. We can just use the generic types.
* Tests were copied from `test_typed` and extended to cover frozen collections as well.
* `edb.common.checked` collections have reprs that don't hide their true nature. For a compact representation use `str(collection)` instead of `repr(collection)`. I hope that's OK.
* Coercion isn't implemented in `edb.common.typed` but in `edb.common.struct` and `edb.schema.objects` (it looks like they share some design). I think it would be cleaner to move this functionality to the checked collections in form of an alternative constructor called simply `CheckedSet[int].coerce(iterable)`.
* Since they need to perform runtime type checks, `edb.common.checked` collections don't accept type variables, unions, and other special types. Parametrization needs to use concrete or abstract builtin types, user classes, or extension types.
* Type annotations on edb.common.checked classes are not covariant for simplicity. If we ever need that, it's relatively easy to change but they might be more verbose.
* Since this was not used anywhere, `edb.common.checked` doesn't implement "accept None" variants. This could be re-added by supporting `Optional[X]` as a type parameter but looks like we don't need it.
* The implementation doesn't depend on runtime type parametrization provided by `typing.Generic[T]` because this one is inadequate for pickling, coercion and type checks at `__init__` time. Instead, we're building picklable parametrized subclasses at runtime with `ParametricType.__class_getitem__`.
* The implementation has *some* code duplication due to type annotations being sometimes different between mutable and frozen variants of the same data structure. `ParametricType` itself cannot be generic because it implements `__class_getitem__` for our collections. I tried to minimize the duplication and think the result is fine. Mypy is very happy with those types.
* The implementation does not allow for customizing the underlying container type.  This ability of edb.common.typed was not used anywhere.

### DIFF
--- a/tests/common/test_checked.py
+++ b/tests/common/test_checked.py
@@ -1,0 +1,370 @@
+#
+# This source file is part of the EdgeDB open source project.
+#
+# Copyright 2011-present MagicStack Inc. and the EdgeDB authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from __future__ import annotations
+from typing import *  # NoQA
+
+import pickle
+import unittest
+
+from edb.common.checked import CheckedDict
+from edb.common.checked import CheckedList
+from edb.common.checked import CheckedSet
+from edb.common.checked import FrozenCheckedList
+from edb.common.checked import FrozenCheckedSet
+
+
+class CheckedDictTests(unittest.TestCase):
+    def test_common_checked_checkeddict_basics(self) -> None:
+        StrDict = CheckedDict[str, int]
+        assert StrDict({"1": 2})["1"] == 2
+        assert StrDict(foo=1, initdict=2)["initdict"] == 2
+
+        sd = StrDict(**{"1": 2})
+        assert sd["1"] == 2
+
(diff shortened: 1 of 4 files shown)


tests/common/test_checked.py
```diff
@@ -0,0 +1,370 @@
...
+        tl = StrSet()
+        tl.add("1")
+        tl.update(("2", "3"))
+        tl |= ["4"]
+        tl |= ("5",)
+        tl = tl | StrSet(["6"])
+        tl = {"0"} | tl
+        assert set(tl) == {"0", "1", "2", "3", "4", "5", "6"}
+
+        tl = "67896789" - tl  # sic, TypedSet used to coerce arguments, too.
```
ambv: Note: do we want to preserve this? Looks a bit error-prone to me but IDK.

PR #801, opened 2019-10-30
Read the module path in test_checked dynamically

ambv:
This allows us to use edb test `tests/common/test_checked.py`

### DIFF
--- a/tests/common/test_checked.py
+++ b/tests/common/test_checked.py
@@ -72,8 +72,9 @@ class CheckedDictTests(unittest.TestCase):
         FooDict = CheckedDict[str, Foo]
 
         td = FooDict(bar=Bar(), foo=Foo())
+        module_path = self.__module__
         expected = (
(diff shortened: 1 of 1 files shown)


tests/common/test_checked.py
```diff
@@ -160,8 +161,9 @@ class Bar(Foo):
 
         tl = FooList([Bar(), Foo()])
         cls_name = self.BaseList.__name__
+        module_path = self.__module__
         expected = (
-            f"edb.common.checked.{cls_name}[common.test_checked."
+            f"edb.common.checked.{cls_name}[{module_path}."
```
1st1: Btw, please use `self.assertEquals()` etc instead of the `assert` statement

1st1: We used to use `pytest` at some point, then switched to a custom test running framework, and after that settled on using unittest. There are probably a few more tests that use the `assert` statement, all it says is that code is oooold.

PR #803, opened 2019-10-30
Fix coverage integration in tests

elprans:
Current coverage integration is buggy in two ways:

First, in `-j1` mode, the test discovery is ran in the same process
as tests, so coverage might report module-level statements as uncovered,
since they've been executed before coverage started.  The fix is to
start coverage _before_ any test discovery.

Second, in `-jN` mode, any code that isn't part of the server process
isn't covered at all.  The fix here is to explicitly start coverage in
the test worker, and not just in the server process entrypoints.

I also added a check that the argument to `--cov` is not a file path,
since it's now possible to get very confused by
`edb test --cov test/file.py` by thinking you've enabled coverage for
everything and are testing `test/file.py`, whereas in fact you've
enabled coverage for `test/file.py` and are testing everything.

### DIFF
--- a/edb/tools/test/__init__.py
+++ b/edb/tools/test/__init__.py
@@ -19,11 +19,18 @@
 
 from __future__ import annotations
 
+import contextlib
+import functools
 import os
+import pathlib
 import sys
+import tempfile
 import unittest
 
 import click
+
+import edb
+from edb.common import devmode
 
 from edb.tools.edb import edbcommands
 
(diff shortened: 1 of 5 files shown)


edb/tools/test/__init__.py
```diff
@@ -109,6 +117,88 @@ def test(*, files, jobs, include, exclude, verbose, quiet, debug,
...
+        files=files,
+        jobs=jobs,
+        output_format=output_format,
+        warnings=warnings,
+        failfast=failfast,
+    )
+
+    if cov:
+        for pkg in cov:
+            if '\\' in pkg or '/' in pkg:
```
ambv: nit: os.sep would be enough

1st1:
```suggestion
            if '\\' in pkg or '/' in pkg or pkg.endswith('.py'):
```

elprans: I had thought about that, but then using `'/'` as separator is kinda also valid on Windows nowadays.

edb/tools/test/__init__.py
```diff
@@ -65,7 +72,8 @@
 @click.option('-x', '--failfast', is_flag=True,
               help='stop tests after a first failure/error')
 @click.option('--cov', type=str, multiple=True,
-              help='file path or package name to measure code coverage for')
+              help='package name to measure code coverage for,'
+                   'can be specified multiple times')
```
1st1:
```suggestion
                   'can be specified multiple times, e.g. "edb.common edb.server"')
```

PR #807, opened 2019-11-01
Remove edb.common.typed

ambv:
There's four separate commits that do their own respective things here:

* Remove TypedSet, replace with CheckedSet;
* Allow subclassing checked collections;
* Make frozen types in edb.common.checked hashable; and finally
* Remove edb.common.typed, replacing TypedLists with CheckedLists.

I'd keep the four separate commits on merge since they have their own respective logic and commit messages.


edb/common/checked.py
```diff
@@ -85,14 +82,29 @@ def __class_getitem__(
...
+            return generic.__class_getitem__(params)  # type: ignore
+
         return type(name, bases, type_dict)
 
     def __reduce__(self) -> Tuple[Any, ...]:
         assert self.types is not None
-        base: Type[ParametricType] = self.__class__.__bases__[0]
+        cls: Type[ParametricType] = self.__class__
         container = getattr(self, "_container", ())
+        if cls.__name__.endswith("]"):
```
1st1: We're using `typing_inspect` module by Ivan (https://github.com/ilevkivskyi/typing_inspect/blob/master/typing_inspect.py) -- can we perhaps write a cleaner check here using his APIs or typing APIs?

tests/common/test_checked.py
```diff
@@ -307,6 +307,22 @@ def __hash__(self):
         )
         assert repr(tl) in {expected_template.format(e) for e in expected}
 
+    def test_common_checkedset_pickling(self):
+        StrSet = self.BaseSet[str]
+        sd = StrSet({"123", "456"})
+
+        assert sd.type is str
```
1st1: If possible please update to use `self.assertIs` etc (just so that we don't forget)

edb/common/checked.py
```diff
@@ -171,6 +171,12 @@ class FrozenCheckedList(
     def __init__(self, iterable: Iterable[T] = ()) -> None:
         super().__init__()
         self._container = [self._check_type(element) for element in iterable]
+        self._hash = -1
+
+    def __hash__(self) -> int:
+        if self._hash == -1:
+            self._hash = hash(tuple(self._container))
```
1st1: I'd also add `type(self)` to the hashed tuple.

edb/common/struct.py
```diff
@@ -345,8 +344,10 @@ def _check_field_type(self, field, name, value):
             if field.coerce:
                 ftype = field.type[0]
 
-                if issubclass(ftype, (typed.AbstractTypedSequence,
-                                      typed.AbstractTypedSet)):
+                if issubclass(ftype, (checked.CheckedList,
```
elprans: Is it possible to have a common abc for this, e.g. `checked.Collection`?

1st1:
Great job 👍  I've left a few nit comments.

edb/common/checked.py
ambv: I mimicked what the equivalent was doing in edb.common.typed but sure, will update.

tests/common/test_checked.py
ambv: Will do.

edb/common/checked.py
ambv: `typing_inspect` will not work: the runtime type checking facility required for us to use a custom `ParametricType` base class which provides `__class_getitem__` that does exactly what we want: constructs a real subclass on the fly with the parametrized types applied.

Since it's a real subclass (unlike what `Generic[T]` does which is to give you an *instance* of `GenericAlias` which itself is not a real type), it has to have a unique name. So we give it a name that includes the generics.

Checking if we're dealing with such dynamic subclass by looking at the last character of the name is maybe not the most elegant thing in the entire module but it's the easiest solution for what we need.

For some context, the reason for the check under which you commented is to identify if we're pickling an instance of this dynamically generated type. In such case, we need to use `__restore__` for unpickling to recreate this dynamic type just-in-time.

If we're dealing with a type that's not the dynamically generated one, that means we are dealing with a subclass of it. If that subclass is seen from global state, we don't have to do anything special, the dynamic base type is already present and properly parametrized by the time our subclass is created.

edb/common/struct.py
ambv: Lists share a base class, Sets do, too. The reason I went with concrete classes is that the bases are not exported in `__all__` and as such I don't really consider them public API.

Why the preference?

elprans: Easy to forget to add another concrete class to the check when adding new collection types.

ambv: I agree, did that then.

edb/common/checked.py
1st1: Please copy/paste this comment into code :)

PR #809, opened 2019-11-01
tests: Update the "tutorial" test.

vpetrovykh:
Bring the tutorial test code up-to-date with the actual tutorial
that appears in the docs.


vpetrovykh:
This test case came up because in the original "tutorial" the `datetime` casts were using very much non-ISO strings.

PR #811, opened 2019-11-04
Add the new DESCRIBE command

elprans:
Currently, there is no way for an EdgeDB client to obtain an SDL or DDL 
representation of a given part of a schema or a part of a schema, except by
using introspection queries and implementing client-side code generation.
Since the facilities to produce correct schema dumps already exist in the
server, it makes a lot of sense to expose the functionality as commands:

    DESCRIBE { <schema-object> [, ...] | SCHEMA }
        [AS {DDL|SDL|TEXT}] [VERBOSE] ;

`DDL`, `SDL` and `TEXT` determine the output format, SDL being the default.

`<schema-object>` is a valid reference to a schema object: a (possibly 
qualified) name, or an object type followed by a name. If the object is 
global, then the type is required.

If the entire schema is to be described, the sole `SCHEMA` keyword is used.

Examples:

    # Return the description of the `User` object type as SDL
    DESCRIBE TYPE User AS SDL;

    # Return the description of the `orders` module as DDL.
    DESCRIBE MODULE orders AS DDL;

    # Return the description of the entire schema as DDL.
    DESCRIBE SCHEMA AS DDL;

    # Return the annotated description of the `User` type with verbosity
    DESCRIBE TYPE User AS TEXT VERBOSE;


edb/repl/__init__.py
```diff
@@ -53,7 +53,7 @@
 
 STATUSES_WITH_OUTPUT = frozenset({
     'SELECT', 'INSERT', 'DELETE', 'UPDATE',
-    'GET MIGRATION',
+    'GET MIGRATION', 'DESCRIBE', 'DESCRIBE'
```
1st1: One `'DESCRIBE'` more than needed

edb/schema/casts.py
```diff
@@ -143,7 +144,8 @@ def get_cast_fullname(
         name=sn.get_specialized_name(shortname, *quals))
 
 
-class Cast(s_anno.AnnotationSubject, s_func.VolatilitySubject, s_abc.Cast):
+class Cast(s_anno.AnnotationSubject, s_func.VolatilitySubject, s_abc.Cast,
+           qlkind=qltypes.SchemaObjectClass.CAST):
```
1st1: Why do you pass `qlkind` to metaclass constructor and not just keep it as a regular class attribute?

edb/schema/objects.py
```diff
@@ -442,9 +443,26 @@ def get_refdict_for_class(cls, refcls):
         else:
             raise KeyError(f'{cls} has no refdict for {refcls}')
 
+    @property
+    def is_schema_object(cls) -> bool:
+        return cls in ObjectMeta._schema_types
```
1st1: Maybe instead of `is_schema_object` we can just use `issubclass` to make mypy et al happier?

elprans: Will do in a separate refactor. This just follows what we currently have.

edb/schema/casts.py
elprans: I think this approach is slightly cleaner since it doesn't imply there's a mutable attribute on a class and we don't need to parse the class dict to validate it.

PR #815, opened 2019-11-05
Fix remaining edb.pgsql.compiler typing errors

elprans:
This makes mypy (strictly) happy with `edb.pgsql.compiler`.

### DIFF
--- a/edb/common/compiler.py
+++ b/edb/common/compiler.py
@@ -19,75 +19,113 @@
 
 from __future__ import annotations
 
+from typing import *  # NoQA
+
 import collections
 import re
 
 
+ContextLevel_T = TypeVar('ContextLevel_T', bound='ContextLevel')
+
+
 class ContextLevel:
-    def on_pop(self, prevlevel):
+    _stack: CompilerContext[ContextLevel]
+
+    def __init__(self, prevlevel: Optional[ContextLevel], mode: Any) -> None:
         pass
 
-    def new(self, mode=None):
-        return self._stack.new(mode, self)
+    def on_pop(self, prevlevel: Optional[ContextLevel]) -> None:
+        pass
+
+    def new(
+        self: ContextLevel_T,
+        mode: Any=None,
+    ) -> CompilerContextManager[ContextLevel_T]:
+        stack = cast(CompilerContext[ContextLevel_T], self._stack)
+        return stack.new(mode, self)
 
 
-class CompilerContextManager:
-    def __init__(self, context, mode, prevlevel):
+class CompilerContextManager(Generic[ContextLevel_T]):
+    def __init__(
+        self,
+        context: CompilerContext[ContextLevel_T],
+        mode: Any,
+        prevlevel: Optional[ContextLevel_T],
+    ) -> None:
         self.context = context
         self.mode = mode
         self.prevlevel = prevlevel
 
-    def __enter__(self):
-        self.context.push(self.mode, self.prevlevel)
-        return self.context.current
+    def __enter__(self) -> ContextLevel_T:
+        return self.context.push(self.mode, self.prevlevel)
 
-    def __exit__(self, exc_type, exc_value, traceback):
+    def __exit__(self, exc_type: Any, exc_value: Any, traceback: Any) -> None:
         self.context.pop()
 
 
-class CompilerContext:
-    def __init__(self):
-        self.stack = []
-        self.push(None)
+class CompilerContext(Generic[ContextLevel_T]):
+    stack: List[ContextLevel_T]
+    ContextLevelClass: Type[ContextLevel_T]
+    default_mode: Any
 
-    def push(self, mode, prevlevel=None):
-        if prevlevel is None:
+    def __init__(self) -> None:
+        self.stack = []
+        self._push(None, initial=True)
+
+    def push(
+        self,
+        mode: Any,
+        prevlevel: Optional[ContextLevel_T] = None,
+    ) -> ContextLevel_T:
+        return self._push(mode, prevlevel)
+
+    def _push(
+        self,
+        mode: Any,
+        prevlevel: Optional[ContextLevel_T] = None,
+        *,
+        initial: bool = False,
+    ) -> ContextLevel_T:
+        if prevlevel is None and not initial:
             prevlevel = self.current
         level = self.ContextLevelClass(prevlevel, mode)
-        level._stack = self
+        # XXX: typing fu
+        level._stack = cast(CompilerContext[ContextLevel], self)
         self.stack.append(level)
         return level
 
-    def pop(self):
+    def pop(self) -> None:
         level = self.stack.pop()
         level.on_pop(self.stack[-1] if self.stack else None)
 
-    def new(self, mode=None, prevlevel=None):
+    def new(
+        self,
+        mode: Any = None,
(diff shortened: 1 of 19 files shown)


edb/common/compiler.py
```diff
@@ -76,17 +99,17 @@ def _current(self):
...
-    def nextval(self, name='default'):
+    def nextval(self, name: str = 'default') -> int:
         self.counts[name] += 1
         return self.counts[name]
 
 
-class AliasGenerator(Counter):
-    def get(self, hint=None):
+class AliasGenerator(SimpleCounter):
+    def get(self, hint: Optional[str]=None) -> str:
```
ambv: If an empty string can be a sentinel, it's generally preferred. In this case could 'v' be the default?

edb/ir/scopetree.py
```diff
@@ -585,13 +585,15 @@ def find_unfenced(self, path_id: pathid.PathId) \
...
+        result: Optional[ScopeTreeNode] = None
+
         for node in self.descendants:
             if node.unique_id == unique_id:
-                return node
+                result = node
+                break
 
-        return None
+        return result
```
ambv: Did you have to make those changes to make Mypy happy?

edb/pgsql/common.py
```diff
@@ -108,7 +108,7 @@ def get_module_backend_name(module, prefix='edgedb_'):
     return edgedb_name_to_pg_name(f'{prefix}{module}', len(prefix))
 
 
-def edgedb_name_to_pg_name(name, prefix_length=0):
+def edgedb_name_to_pg_name(name, prefix_length=0) -> str:
```
ambv: `name` and `prefix_length` look like easy targets to type, too?

edb/pgsql/compiler/dml.py
```diff
@@ -440,7 +440,7 @@ def compile_insert_shape_element(
             # anyway.
             insvalctx.volatility_ref = context.NO_VOLATILITY
 
-        insvalctx.path_scope[ir_stmt.subject.path_id] = insert_stmt
+        # insvalctx.path_scope[ir_stmt.subject.path_id] = insert_stmt
```
ambv: Is this supposed to stay commented out?

edb/pgsql/compiler/dml.py
```diff
@@ -415,7 +415,7 @@ def process_insert_body(
                     shape_el, None, ctx=scopectx)
                 prop_elements.append(tuple_el)
 
-        valtuple = pgast.TupleVar(elements=prop_elements, named=True)
+        valtuple = pgast.TupleVarBase(elements=prop_elements, named=True)
```
ambv: Why the change? Esp. that typically I understand a class called "*Base" to be a building block, not something to be directly instantiated.

edb/pgsql/compiler/pathctx.py
```diff
@@ -124,6 +133,7 @@ def get_path_var(
             src_path_id = path_id
         else:
             src_path_id = path_id.src_path()
+            assert src_path_id is not None
```
ambv: This is subtle and I can't see. I looked in ir/pathid.py and in the src_path() method but I don't quite get it. Why do we know this will never be None?

edb/pgsql/compiler/pathctx.py
```diff
@@ -265,19 +280,6 @@ def get_path_var(
...
-        else:
-            outer_path_id = drilldown_path_id
-
-        path_id_map = {
-            outer_path_id: next(iter(source_rel.path_scope))
-        }
-
-        drilldown_path_id = map_path_id(
-            drilldown_path_id, path_id_map)
-
```
ambv: Dead code?

mypy.ini
```diff
@@ -16,6 +16,22 @@ warn_unused_configs = True
 follow_imports = True
 ignore_errors = False
 
+[mypy-edb.pgsql.compiler.*]
+follow_imports = True
+ignore_errors = False
+# Equivalent of --strict on the command line:
+disallow_subclassing_any = True
+disallow_any_generics = True
+; disallow_untyped_calls = True
```
ambv: Use # for comments and comment why you're disabling this. In checked.py the only reason is use of @overload which requires the default implementation to be untyped (this is where PEP 484 disagrees with Mypy I guess).

edb/ir/scopetree.py
elprans: Yes, otherwise it complained that I was returning `Any`

edb/pgsql/compiler/dml.py
elprans: `TupleVarBase` differs from `TupleVar` in that the former is just a "shape" of a tuple, whereas the latter actually holds the values.  In this case we need the one without the values.  Another approach is to make `TupleElement.val` to be `Optional`, but that opens a much bigger can of worms.

edb/ir/scopetree.py
ambv: That looks like a bug. It probably can't unwind the double generator right?

edb/pgsql/compiler/pathctx.py
elprans: It is correlated with `path_id.rptr()`, which we check above.  If `path_id.rptr()` is not `None`, then `src_path()` would not be.  Both are essentially `[:-1]` slices on the path id.

edb/pgsql/compiler/pathctx.py
elprans: Dead code.

mypy.ini
elprans: VSCode did the `;` thing, not me :-)  `disallow_untyped_calls` is unset because the compiler calls out to lots of untyped code in `edb.pgsql`, which is beyond the scope here.

edb/common/compiler.py
```diff
@@ -19,55 +19,78 @@
...
+
+
 class ContextLevel:
     def on_pop(self, prevlevel):
         pass
 
-    def new(self, mode=None):
+    def new(
+        self: ContextLevel_T,
+        mode=None,
```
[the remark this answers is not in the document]
elprans: Now is.

edb/common/compiler.py
```diff
@@ -19,55 +19,78 @@
...
         return self._stack.new(mode, self)
 
 
-class CompilerContextManager:
-    def __init__(self, context, mode, prevlevel):
+class CompilerContextManager(Generic[ContextLevel_T]):
+    def __init__(
+        self,
+        context: CompilerContext[ContextLevel_T],
+        mode: Any,
```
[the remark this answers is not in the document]
elprans: Yeah, depends on actual context implementation, usually some enum.

edb/common/compiler.py
elprans: Fixed.

edb/ir/scopetree.py
elprans: Yeah, this was probably some intermediate issue.  It's not reproducing now, so I reverted back to the original formulation.

edb/pgsql/common.py
elprans: Yes.

edb/pgsql/compiler/dml.py
elprans: It's supposed to be deleted, which I've now done.

PR #816, opened 2019-11-07
Enable type checking of edb.schema.objects

ambv:
Also enabled edb.schema.name which was a dependency.

### DIFF
--- a/edb/edgeql/compiler/cast.py
+++ b/edb/edgeql/compiler/cast.py
@@ -35,6 +35,7 @@ from edb.ir import utils as irutils
 from edb.schema import casts as s_casts
 from edb.schema import functions as s_func
 from edb.schema import modules as s_mod
+from edb.schema import objects as s_objects
 from edb.schema import types as s_types
 
 from edb.edgeql import ast as qlast
@@ -113,8 +114,10 @@ def compile_cast(
             ir_set, orig_stype, new_stype, srcctx=srcctx, ctx=ctx)
 
     else:
-        json_t = ctx.env.get_track_schema_object('std::json')
-
+        json_t = cast(
+            s_objects.InheritingObjectBase,
+            ctx.env.get_track_schema_object('std::json'),
+        )
         if (new_stype.issubclass(ctx.env.schema, json_t) and
                 ir_set.path_id.is_objtype_path()):
             # JSON casts of objects are special: we want the full shape
--- a/edb/schema/name.py
+++ b/edb/schema/name.py
@@ -25,13 +25,21 @@ import functools
 from edb import errors
 
 
+if TYPE_CHECKING:
+    NameT = TypeVar("NameT", bound="SchemaName")
+
+
 class SchemaName(str):
     __slots__ = ('module', 'name')
 
     module: str
     name: str
 
-    def __new__(cls, name, module=None):
+    def __new__(
+        cls: Type[NameT],
+        name: Union[SchemaName, str],
+        module: Optional[str] = None,
+    ) -> NameT:
         if not name:
             raise NameError('name must not be empty')
 
@@ -52,7 +60,10 @@ class SchemaName(str):
                 else:
                     _module = module
 
-        result = super().__new__(cls, _module + '::' + _name)
+        # Ignore below since Mypy doesn't believe you can pass `_fullname` to
+        # object.__new__.
+        _fullname = f"{_module}::{_name}"
+        result = super().__new__(cls, _fullname)  # type: ignore
         result.name = _name
         result.module = _module
 
@@ -62,14 +73,16 @@ class SchemaName(str):
         return (self.module, self.name)
 
     @staticmethod
-    def is_qualified(name):
+    def is_qualified(name: Union[SchemaName, str]) -> bool:
         return isinstance(name, SchemaName) or '::' in name
 
 
 class UnqualifiedName(SchemaName):
 
-    def __new__(cls, name):
-        result = str.__new__(cls, name)
+    def __new__(cls: Type[NameT], name: str) -> NameT:
+        # Ignore below since Mypy doesn't believe you can pass `name` to
+        # object.__new__.
+        result = str.__new__(cls, name)  # type: ignore
         result.name = name
         result.module = ''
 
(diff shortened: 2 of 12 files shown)


edb/schema/objects.py
```diff
@@ -71,24 +89,30 @@ def default_field_merge(target: Object, sources: List[Object],
...
         self.context = context
         self.pair = pair
 
-    def __enter__(self):
+    def __enter__(self) -> None:
         self.context.push(self.pair)
 
-    def __exit__(self, exc_type, exc_value, traceback):
+    def __exit__(
+        self, exc_type: type, exc_value: Exception, traceback: Any
```
elprans: Strictly speaking, `exc_type`, `exc_value` and `traceback` are optional in `__exit__`.  It's a bit annoying that Mypy forces you to explicitly type methods that it should know the signature of, them being part of the language spec.

edb/schema/objects.py
```diff
@@ -170,7 +205,7 @@ def __init__(self, type_, *, coerce=False,
         self.coerce = coerce
         self.allow_ddl_set = allow_ddl_set
 
-        self.compcoef = compcoef
+        self.compcoef = compcoef  # XXX what's up with it defaulting to None?
```
elprans: `None` means "don't compare"

edb/schema/objects.py
```diff
@@ -228,23 +263,24 @@ def required(self):
         return True
 
     @property
-    def is_schema_field(self):
+    def is_schema_field(self) -> bool:
         return False
 
-    def __get__(self, instance, owner):
+    def __get__(self, instance: T, owner: Type[T]) -> Optional[Field]:
+        # XXX what is happening in this method?
```
elprans: When a `<Field.name>` is accessed on a class, we return the `Field` instance, but when it's accessed on an instance, we return `None` to indicate "no value for this field", since, if there was a value, it would have been in the object's __dict__ and we wouldn't get here.

Now, `Field` is supposed to be non-optional (as opposed to `SchemaField`), but `EphemeralCollection` still relies on this.

edb/schema/objects.py
```diff
@@ -653,7 +712,13 @@ def _get_schema_field_value(self, schema, field_name, *,
         raise FieldValueNotFoundError(
             f'{self!r} object has no value for field {field_name!r}')
 
-    def get_field_value(self, schema, field_name, *, allow_default=True):
+    def get_field_value(
```
elprans: The mypy plugin might help with the return type here.

edb/schema/objects.py
ambv: I'll fix the optional bit.

1st1: `traceback: Any` -> `traceback: Optional[types.TracebackType]`
`exc_type: type` -> `exc_type: Optional[Type[Exception]]`

edb/schema/objects.py
ambv: Yes but just above you're declaring Field.compcoef to be strictly a float. I'll make it `Optional[float]` there.

ambv:
This is ready.

edb/edgeql/compiler/cast.py
```diff
@@ -113,8 +114,10 @@ def compile_cast(
             ir_set, orig_stype, new_stype, srcctx=srcctx, ctx=ctx)
 
     else:
-        json_t = ctx.env.get_track_schema_object('std::json')
-
+        json_t = cast(
+            s_objects.InheritingObjectBase,
```
1st1: The actual type is going to be `edb.schema.scalars.ScalarType`

edb/schema/name.py
```diff
@@ -62,14 +73,16 @@ def as_tuple(self):
...
         return isinstance(name, SchemaName) or '::' in name
 
 
 class UnqualifiedName(SchemaName):
 
-    def __new__(cls, name):
-        result = str.__new__(cls, name)
+    def __new__(cls: Type[NameT], name: str) -> NameT:
+        # Ignore below since Mypy doesn't believe you can pass `name` to
+        # object.__new__.
```
1st1: This is a bug of mypy, right?

edb/schema/objects.py
```diff
@@ -43,52 +47,113 @@
...
+class NoDefaultT(enum.Enum):
+    """Used as a sentinel indicating that a named argument wasn't passed.
+
+    Trick from https://github.com/python/mypy/issues/7642.
+    """
+    NoDefault = 0
+
+
+NoDefault: Final = NoDefaultT.NoDefault
+T = TypeVar("T")
```
1st1: Should lines 87..93 be moved up into the `if TYPE_CHECKING:` block?

edb/schema/objects.py
```diff
@@ -43,52 +47,113 @@
...
+        def __init__(self, from_iter: Optional[Iterable[CovT]] = None) -> None:
+            ...
+
+
+class NoDefaultT(enum.Enum):
+    """Used as a sentinel indicating that a named argument wasn't passed.
+
+    Trick from https://github.com/python/mypy/issues/7642.
+    """
+    NoDefault = 0
```
1st1: 🤦‍♂️

edb/schema/objects.py
```diff
@@ -228,23 +314,31 @@ def required(self):
...
-    def __get__(self, instance, owner):
+    @overload
+    def __get__(self, instance: None, owner: Type[Object]) -> Field[Type[T]]:
+        ...
+
+    @overload  # NoQA: F811
+    def __get__(self, instance: Object, owner: Type[Object]) -> T:
+        ...
+
+    def __get__(self, instance, owner):  # NoQA: F811
```
1st1: Some mad typing here...  💥

edb/schema/name.py
ambv: Yes and no. `object` itself does not accept kwargs since Python 2.0 or something:
```pycon
>>> object.__new__(object, name='hey')
  File "<stdin>", line 1, in <module>
TypeError: object() takes no arguments
```
But in this case `super()` will actually not be `object` yet because we're using multiple inheritance and Mypy doesn't see it.

edb/edgeql/compiler/cast.py
ambv: Yeah, we can change this but we only need `issubclass` below. Since it's a cast to get Mypy quiet, doesn't really matter much.

edb/schema/objects.py
ambv: Some of them are used in the runtime context, e.g. in casts, type aliases and in generics that are not inside annotations. But I see that not all, I'll move 'em.

PR #817, opened 2019-11-08
An assortment of fixes for DESCRIBE infidelities

elprans:
* Fix rendering of view and computables

  Use the correct `CREATE VIEW` syntax for views and prefer the shorthand 
  syntax for computable links and properties.

* Emit abstract constraint parameters correctly

  The parameter definitions of abstract constraints are currently not emitted
  altogether.  Fix this.

* Stop marking Constraint.errmessage as non-inherited if interpolated

  In abstract constraints `errmessage` is usually a string template, which
  may contain references to constraint arguments, in which case argument
  values are interpolated into the error message, and it is marked as
  non-inherited.  The last bit is actually not necessary, since the
  interpolation is reproducible.

* Render union types correctly in description of link targets

  This ensures that the original type union expression is rendered instead of
  an internal name for the generated union type.

* Omit derived types (such as unions) from DESCRIBE output

  Derived types are a result of some DDL construct, such as a union type 
  declaration, as should not appear as full types.

Props to @1st1 for noticing these.

### DIFF
--- a/edb/schema/constraints.py
+++ b/edb/schema/constraints.py
@@ -569,6 +569,31 @@ class CreateConstraint(ConstraintCommand,
 
         return cmd
 
+    def _apply_fields_ast(self, schema, context, node):
+        super()._apply_fields_ast(schema, context, node)
+
+        if isinstance(node, qlast.CreateConstraint):
+            params = []
+            for op in self.get_subcommands(type=s_func.ParameterCommand):
+                props = op.get_struct_properties(schema)
+                pname = s_func.Parameter.paramname_from_fullname(props['name'])
+                if pname == '__subject__':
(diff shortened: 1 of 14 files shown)


edb/schema/constraints.py
```diff
@@ -180,8 +180,9 @@ def get_concrete_constraint_attrs(
                         for name, val in args_map.items()}
 
             args_map['__subject__'] = '{__subject__}'
-            attrs['errmessage'] = attrs['errmessage'].format(**args_map)
-            inherited.pop('errmessage', None)
+            message_template = attrs['errmessage']
```
1st1: Typo in the commit message 'makring' -> 'marking'

PR #828, opened 2019-11-13
edgeql: Change `FROM` to `USING` keyword.

vpetrovykh:
In contexts where `FROM` appears to indicate the language or original
used change the syntax to use `USING` keyword. This primarily affects
function, cast and operator bodies.

In contexts where `FROM` indicates some initial state, the keyword
remains unchanged. This concernes migrations and cast signatures.


elprans:
We also need to change expr := for computables and views to USING

vpetrovykh:
Oh, that's right. I don't think we have an issue for this and it's big enough that it needs documentation of our thinking process. I'll make an issue.

PR #832, opened 2019-11-15
edgeql: Remove SELECT-like clauses from FOR.

### LINKED ISSUE #743: Remove SELECT-like clauses from the FOR statement
Currently, the `FOR` statement accepts the `FILTER`, `ORDER BY`, `OFFSET` and `LIMIT` clauses that apply to the *result* of the `UNION`. In essense, `FOR x IN <expr> UNION <expr> FILTER <flt>` is equivalent to `SELECT <implicit_alias> := (FOR x IN <expr> UNION <expr>) FILTER <flt>`

The problem is that the above transformation is arguably not intuitive, since it is possible to *also* assume that the `FILTER` clause etc. apply to the expression in the `UNION` clause.  The current
implementation is also incorrect, since it doesn't hide the iterator variable from the followup clauses, and doing that would also be an exception from all other statements.

For those reasons, remove the clause sugar support from the `FOR` statement until we figure out an intuitive way of reintroducing them.

vpetrovykh:
The `FOR` statement now only has the basic form of:

    FOR x IN {iterexpr} UNION expr;

All the `FILTER`, `ORDER BY`, `LIMIT` and `OFFSET` clauses have been
removed due to the potential ambiguity of their meaning (as evidenced by
even some tests with incorrect semantics).

If those clauses are necessary, then the correct procedure is to wrap
the whole `FOR ...` expression into a `SELECT`.

When rewriting the tests to move the SELECT-like clauses some potential
scoping issues were dicovered.

Fixes #743.

### DIFF
--- a/docs/edgeql/statements/for.rst
+++ b/docs/edgeql/statements/for.rst
@@ -17,15 +17,7 @@ FOR
 
     FOR <variable> IN "{" <iterator-set> [, ...]  "}"
 
-    UNION <output-expr>
-
-    [ FILTER <filter-expr> ]
-
-    [ ORDER BY <order-expr> [direction] [THEN ...] ]
-
-    [ OFFSET <offset-expr> ]
-
-    [ LIMIT  <limit-expr> ] ;
+    UNION <output-expr> ;
 
 :eql:synopsis:`FOR <variable> IN "{" <iterator-set> [, ...]  "}"`
     The ``FOR`` clause has this general form:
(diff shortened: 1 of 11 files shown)


vpetrovykh:
Please note that the new failing tests highlight bugs that have nothing to do with the syntax change. They just happened to be produced because of the rewrite of originally awkward syntax. I'll create a separate issue for them.

vpetrovykh:
I've updated the docs and removed the extra clauses from `FOR` description.

Issue #694, opened 2019-08-16
Optimize the IR compiler


ambv:
Currently the IR compiler is pure Python and it takes a long time for most EdgeQL queries, most likely in edgeql/compiler/func.py. Investigate and implement ways to make it faster.

ambv:
An aggregation of `edb/server/compiler/compiler.py:_compile_ql_query()` calls over all unit tests shows that:
- `compile_ast_to_ir` takes 80% of time
- from that, 42% is just `compile_BinOp`

This is the first thing I'll be looking at optimizing. Interestingly, there is a lot of time spent in those as well:
- `edb/schema/types.py:derive_subtype` takes 25% of the cumulative time
- `edb/common/ast/base.py:_check_type` takes 18% of the cumulative time and the longest "per call"

The cast distance calculation that was suspected as taking a lot of time, is not *that* expensive alone (although it still clocks at 10.5% of the cumulative time). I'll look into it while investigating `compile_BinOp`.

In terms of code structure, we're creating a ton of AST objects (over 2 mil in my sampling) which I'll look into decreasing. `edb/common/ast/base.py:__init__` also does a lot of work (cumulative time clocking at 18% which I suspect is due to type checking that I already covered above).

We're calling a lot of `__hash__` functions (50 mil calls during my sampling), out of which `UUID.__hash__` is dominant at 47 mil calls (taking 22.5 seconds in total). Maybe we can decrease that.

Raw data: https://gist.github.com/ambv/7b9e6b72d923ec9845c549e8fc47e6eb

1st1:
Thanks! A few questions:

* What was the query you profiled? Did you profile more than one?
* Ideally we need to profile like 10 queries of various complexity and features. Use `edb inittestdb` to generate test databases and then you can profile some queries from our tests.
* Can you share the test script you used to generate this?

For example, the high cost of `compile_BinOp` could also be explained if the benchmark query was `SELECT a + b`. Moreover, since operators in EdgeQL are compiled in the same way we compile function calls, the high cost of compiling them is explained by the same complex quadratic algorithm we use for our multiple dispatch.

> We're calling a lot of __hash__ functions (50 mil calls during my sampling), out of which UUID.__hash__ is dominant at 47 mil calls (taking 22.5 seconds in total). Maybe we can decrease that.

[...]

PR #835, opened 2019-11-15
Add @profile() decorator

ambv:
It lives under `edb.tools.profiling`. This decorator has a few unique properties:

- it allows for reusing the same profiler instance across calls, making it feasible to use also for functions called millions of times

- it uses temporary files in ways that enable using it across processes

- then provides an `aggregate()` function to gather multiple saved profile files into one result that can be observed

There is a simple command-line client for the `aggregate()` functionality invoked by `python -m edb.tools.profiling`.

Note: by default on macOS temporary files are created in a deterministic directory which is *not* /tmp/.  If you don't pass `dir=` to the `@profile()` decorator, also don't pass a directory to the aggregate function (or the command-line client) and the function will find the directory for you.

There are tests.

### DIFF
--- a/edb/tools/profiling.py
+++ b/edb/tools/profiling.py
@@ -37,7 +37,7 @@ PREFIX = "edgedb_"
 SUFFIX = ".pstats"
 
 
-T = TypeVar("T")
+T = TypeVar("T", bound=Callable[..., Any])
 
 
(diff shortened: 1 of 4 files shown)


ambv:
This is related to work on #694.

Issue #632, opened 2019-04-25
Improve backtick quoting


vpetrovykh:
Currently backticks themselves cannot be used inside a backtick-quoted identifier. The proposal is to use double backticks to mean a single backtick character in the identifier:

```edgeql
SELECT `a ``quoted`` identifier`;
```

Also, the error message for starting a quoted identifier with an `@` should be more user-friendly, like "identifiers cannot start with '@'"

Sija:
> The proposal is to use double backticks to mean a single backtick character in the identifier:

I'd suggest more popular backslash (`\`) character.

PR #839, opened 2019-11-18
edgeql: Update backtick quoting.

vpetrovykh:
Add the ability to use a backtick inside a backtick-quoted identifier by
using a double-backtick "``". Primarily the double-backtick is chosen
over backslash-quoting "\`" because the latter could be misleading. No
other backslash-escape sequence is interpreted inside backticks so
using, say "\n", in an identifier would be accepted, but not result in a
newline character.

Update the error message for identifiers starting with "@" to be more
specific.

Fixes: #632.

### DIFF
--- a/tests/test_edgeql_syntax.py
+++ b/tests/test_edgeql_syntax.py
@@ -759,6 +759,9 @@ aa';
         SELECT foo::`bar`;
         SELECT `foo`::bar;
         SELECT `foo`::`bar`;
+        SELECT `foo``bar`;
+        SELECT `foo`::`bar```;
+        SELECT `foo::bar`;
 
 % OK %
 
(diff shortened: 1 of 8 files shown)


tests/test_edgeql_syntax.py
```diff
@@ -759,6 +759,7 @@ def test_edgeql_syntax_name_01(self):
         SELECT foo::`bar`;
         SELECT `foo`::bar;
         SELECT `foo`::`bar`;
```
1st1: Please add `SELECT `foo::bar` test

vpetrovykh: I've added a syntax test with a (valid) trailing backtick in the identifier.

vpetrovykh:
I've added a functional test creating a type with spaces and a backtick in its name. Then performing `INSERT` and `SELECT` on it and verifying the `__type__.name` has _one_ backtick and not two.

vpetrovykh:
```
SELECT `foo::bar`;
```
I've added that to  _syntax_ test (which works fine), but it appears that functionally this is causing problems. I'll open a separate issue for that.

PR #841, opened 2019-11-18
edgeql: Ban "::" from being used inside quoted identifiers.

### LINKED ISSUE #840: Identifier quoting clashes with fully-qualified names
The point of identifier quoting is that it allows using non-alphanumeric characters or reserved keywords as identifiers. So consider the following:
```
CREATE TYPE `foo::Bar`;
```
It will be interpreted as a single identifier, which then gets prefixed with the `default` module name to create a fully-qualified name with module `default` and name `foo::Bar`. Unfortunately, during processing it gets normalized into a single string `"default::foo::Bar"`, which then further down the line gets (mis-)interpreted as module `default::foo` and name `Bar`.
```
edgedb> CREATE TYPE `foo::Bar`;
CREATE
edgedb> WITH MODULE schema
....... SELECT ObjectType {name};
UnknownModuleError: module 'default::foo' is not in this schema
```

So the question is what is the right fix here?
- We could ban `::` inside quoted identifiers, for the same reasons we ban `@` and `__` at the beginning of them already. It causes clashes and it's problematic to disambiguate in some circumstances (such as in `__type__.name` string).
- We could review what it means to be a fully-qualified name and whether it really *has* to be 2 identifier tokens separated by a `::` token or whether it's OK to just have a single quoted identifier with the `::` inside of it. This would mean that the type from the example would simply be equivalent to a plain `foo::Bar`, meaning module `foo` type-name `Bar`.

Personally, I'm leaning towards a ban. @1st1 and @elprans , what do you think?

vpetrovykh:
Using "::" inside a quoted identifier leads to ambiguity in some
representations so it is illegal.

Fixes #840.

### DIFF
--- a/docs/edgeql/lexical.rst
+++ b/docs/edgeql/lexical.rst
@@ -25,11 +25,12 @@ There are two ways of writing identifiers in EdgeQL: plain and quoted.
 The plain identifiers are similar to many other languages, they are
 alphanumeric with underscores and cannot start with a digit. The
 quoted identifiers start and end with a *backtick*
-```quoted.identifier``` and can contain any characters inside, but
-must not start with an ampersand (``@``). If there's a need to include
-a backtick character as part of the identifier name a double-backtick
-sequence (``````) should be used: ```quoted``identifier``` will result
-in the actual identifier being ``quoted`identifier``.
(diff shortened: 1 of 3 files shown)


vpetrovykh:
Docs updated.

PR #842, opened 2019-11-18
Add dependencies for building postgres (fixes #838)

### LINKED ISSUE #838: Build instructions are not full
Looking at https://edgedb.com/docs/internals/dev/:

1. (minor) We might want to add `git submodule update --init` as most projects do, as it took me few minutes to find out :)
2. Currently, postgres (or a module of?) is built as part of edgedb develop installation, and I needed `libreadline-dev zlib1g-dev uuid-dev` on ubuntu bionic to build it. I don't see it mentioned somewhere.

(no PR as I'm not sure how to describe latter in particular minumum versions... or maybe postgres modules will be obsolete soon enough)

tailhook:


docs/internals/dev.rst
```diff
@@ -20,7 +20,14 @@ Linux or macOS.  Windows is not currently supported.
...
 * Bison 1.875 or later;
 * Flex 2.5.31 or later;
-* Perl 5.8.3 or later.
+* Perl 5.8.3 or later;
+* Zlib (zlibg1-dev on Ubuntu);
+* Readline dev package;
+* Libuuid dev package.
+
+.. zlib, readline and libuuid are required to build postgres. Should be removed
+   when custom postgres build is no longer needed.
```
1st1: We'll probably continue having postgres as a submodule even when we have a hosted version etc. It just makes it easier for us pin a specific dev version of postgres; and for many people to build edgedb locally without bothering with installing a new system postgres.

Issue #640, opened 2019-05-01
Create an enum for the "sys::get_transaction_isolation()" function


1st1:
@elprans would you be OK with that?

PR #845, opened 2019-11-19
edgeql: Add an enum for transaction isolation modes.

vpetrovykh:
Add `sys::transaction_isolation_t` enum to denote possible transaction
isolation modes. The `sys::get_transaction_isolation` function now
returns that enum.

Fixes: #640.

### DIFF
--- a/docs/edgeql/funcops/sys.rst
+++ b/docs/edgeql/funcops/sys.rst
@@ -135,14 +135,29 @@ System
         {'1.0-alpha.1'}
 
 
-.. eql:function:: sys::get_transaction_isolation() -> str
+.. eql:function:: sys::get_transaction_isolation() -> \
+                        sys::transaction_isolation_t
 
     Return the isolation level of the current transaction.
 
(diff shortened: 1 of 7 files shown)


docs/edgeql/funcops/sys.rst
```diff
@@ -135,14 +135,29 @@ System
...
+
+.. eql:type:: sys::transaction_isolation_t
+
+    :index: enum transaction isolation
+
+    :eql:type:`Enum <enum>` indicating the possible transaction
+    isolation modes.
+
+    This enum takes the following values: ``READ UNCOMMITTED``, ``READ
+    COMMITTED``, ``REPEATABLE READ``, ``SERIALIZABLE``.
```
elprans: We only support `REPEATABLE READ` and `SERIALIZABLE`, please drop the irrelevant modes.

vpetrovykh: OK, I removed the extra modes from docs and actual enum.

PR #847, opened 2019-11-20
Disable runtime checks in production by default

ambv:
Runtime type checks in edb.common.ast and edb.common.checked are taking up to
15% of the compilation cost based on profiling done as part of #694.

This change disables runtime type checks  by default unless an
EDGEDB_DEBUG_TYPECHECK=1 flag is passed in the environment.  Note: this is also
true for tests with the exception of tests of the runtime typechecking
capability itself.

In the case of edb.common.checked, disabling type checks cannot unfortunately
be done by replacing the data structures with builtin untyped versions due to
picklability.  Instead, we're shadowing the type checking classmethods from the
abstract classes with identity functions on the concrete types.

Benchmarks to follow.

### DIFF
--- a/edb/common/debug.py
+++ b/edb/common/debug.py
@@ -108,6 +108,9 @@ class flags(metaclass=FlagsMeta):
     disable_qcache = Flag(
         doc="Disable server query cache. Parse/Execute will always recompile.")
 
+    typecheck = Flag(
+        doc="Perform runtime type checking.")
+
 
 @contextlib.contextmanager
 def timeit(title='block'):
--- a/edb/common/ast/base.py
+++ b/edb/common/ast/base.py
@@ -28,6 +28,7 @@ from typing import *  # NoQA
 
 import typing_inspect
 
+from edb.common import debug
 from edb.common import markup
 from edb.common import typeutils
 
@@ -462,7 +463,11 @@ def _check_mapping_type(type_, value, raise_error, instance_type):
         _check_type(vtype, v, raise_error)
 
 
-def _check_type(type_, value, raise_error):
+def _check_type_passthrough(type_, value, raise_error):
+    pass
+
+
+def _check_type_real(type_, value, raise_error):
     if type_ is None:
         return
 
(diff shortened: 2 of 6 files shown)


1st1:
@elprans Do you want typechecking to be enabled by a separate debug flag? I was thinking about enabling it always in tests & when EDGEDB_DEBUG_SERVER is set.

elprans:
I feel we should avoid lumping many behaviors into a single flag, so a separate flag is OK.  We should enable it in [devmode] and tests by default for now.

[devmode]: https://github.com/edgedb/edgedb/blob/450d5d47e0e4e5fb4ac3451eade2d9478c5ab768/edb/tools/edb.py#L43-L45

1st1:
> We should enable it in devmode and tests by default for now.

Yeah, that's what I'm talking about.

edb/common/debug.py
```diff
@@ -141,6 +142,26 @@ def print(*args):
     builtins.print(*args)
 
 
+def enable_typechecks():
+    if 'edb.common.checked' in sys.modules:
```
1st1: I'd just import these modules as the overhead is low. Otherwise seems very fragile -- what if something changes in how we load the system and enable_typechecks() is called before these imports.

ambv: The modules themselves enable or disable type checking on import based on what `debug.flags.typecheck` is saying. Because of that, they import `edb.common.debug`.

Therefore, we cannot indiscriminately import here if the module wasn't imported before. Otherwise during import of `edb.common.debug` we will end up importing `edb.common.checked` which in turn will end up importing `edb.common.debug`. An import cycle.

But to have `--devmode` enable runtime typechecks by default, we *do* need to update the status of runtime typechecking using `edb.common.debug.init_debug_flags()`. This is because by the time `edb.tools.edb.edbcommands()` is called, `edb.common.checked` and `edb.common.ast.base` were already imported and type checks were disabled.

1st1:
**some context**

I've done some syntax tweaks recently and runtime types checking was extremely useful, simply because our parsers aren't yet typed. Strictly typing them is going to be quite challenging and it's hard to put an ETA on that. Elvis also finds runtime type checking useful. So we better keep it.

OTOH we don't change syntax that frequently, so having runtime type checking working in prod or in regular local `edb test` runs doesn't make a lot of sense. Having them in CI and as an opt-in via a debug flag should be sufficient.

edb/common/debug.py
1st1: Got it.

ambv:
This is rebased.

PR #849, opened 2019-11-20
Implement support for describing functions

1st1:
The following commands are now supported:

* `DESCRIBE OBJECT <funcname>`;
* `DESCRIBE FUNCTION <funcname>`.

Also fix the way the DESCRIBE command outputs its result -- it
should be as if it selects a raw string literal.

### DIFF
--- a/edb/edgeql/compiler/stmt.py
+++ b/edb/edgeql/compiler/stmt.py
@@ -30,6 +30,7 @@ from edb.ir import ast as irast
 from edb.ir import typeutils as irtyputils
 
 from edb.schema import ddl as s_ddl
+from edb.schema import functions as s_func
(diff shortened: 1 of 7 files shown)


edb/edgeql/compiler/stmt.py
```diff
@@ -423,19 +424,43 @@ def compile_DescribeStmt(
...
                             objref.itemclass),
                     )
 
-                obj = schemactx.get_schema_object(
-                    objref,
-                    item_types=itemtypes,
-                    ctx=ictx,
-                )
-                items.append(obj.get_name(ictx.env.schema))
+                if objref.itemclass is None or \
```
elprans: Backslash and not `(`?

PR #851, opened 2019-11-21
Add basic `\d` command to REPL.

vpetrovykh:
Add `\d NAME` command to REPL. It is a shortcut for `DESCRIBE OBJECT`
where the `NAME` is plain text name of the concept that needs to be
described. No backtick name quoting is needed, but it can be used.

For example, the following are equivalent:

    \d O bje`ct
    \d `O bje``ct`

Largely the backticks are optional so that when keywords are used as
names no quoting is necessary. For example, describing a user-created
type `Commit` could be done like this:

    \d Commit
    \d default::Commit
    \d default::`Commit`

Issue: #179

### DIFF
--- a/edb/edgeql/parser/grammar/lexer.py
+++ b/edb/edgeql/parser/grammar/lexer.py
@@ -229,7 +229,7 @@ class EdgeQLLexer(lexer.Lexer):
              regexp=r'''
                     __[^\W\d]\w*__
                     |
-                    `__.*?__`
+                    `__([^`]|``)*__`(?!`)
                 '''),
 
         Rule(token='IDENT',
@@ -279,7 +279,10 @@ class EdgeQLLexer(lexer.Lexer):
             self.handle_error(txt)
 
         elif rule_token == 'QIDENT':
-            if txt[1] == '@':
+            if txt == '``':
+                self.handle_error(f'Identifiers cannot be empty',
+                                  exact_message=True)
+            elif txt[1] == '@':
                 self.handle_error(f'Identifiers cannot start with "@"',
                                   exact_message=True)
             elif '::' in txt:
--- a/edb/repl/__init__.py
+++ b/edb/repl/__init__.py
@@ -31,8 +31,10 @@ import immutables
 
 from prompt_toolkit import application as pt_app
 from prompt_toolkit import completion as pt_complete
+from prompt_toolkit import document as pt_document
 from prompt_toolkit import enums as pt_enums
 from prompt_toolkit import filters as pt_filters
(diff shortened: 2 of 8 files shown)


edb/edgeql/parser/grammar/lexer.py
```diff
@@ -238,7 +238,7 @@ class EdgeQLLexer(lexer.Lexer):
 
         Rule(token='QIDENT',
              next_state=STATE_KEEP,
-             regexp=r'`([^`]|``)*`'),
+             regexp=r'`([^`]|``)+`'),
```
1st1: I'm not sure I get these changes. Can you add a couple of syntax tests that wouldn't work with old regexes?

vpetrovykh: The change is so that empty string is no longer a valid identifier. It seems like a kind of identifier that is likely to cause errors and confusion. Such as introspected name would be `"modname::"`, which although technically correct, looks more like some kind of error.

I can change it back if we really want to keep empty strings as valid names, I'm just not sure it's a good idea.

I can add some tests for empty string names.

1st1: > The change is so that empty string is no longer a valid identifier. 

yeah, add a test for that!

vpetrovykh: I've updated the tests and adjusted the regexp back to the original so that the error message about empty identifiers could be more specific.

edb/repl/__init__.py
```diff
@@ -299,6 +301,35 @@ def command_list_dbs(self, args):
         for dbn in result:
             print(f'  {dbn}')
 
+    @_command('d', R'\d NAME', 'describe concept')
+    def command_describe_object(self, args):
+        name = utils.normalize_name(args)
```
```suggestion
        name = utils.normalize_name(args.strip())
```

edb/repl/utils.py
```diff
@@ -82,3 +84,39 @@ def split_edgeql(script: str, *, script_mode=True):
...
+def _strip_quote(part):
+    if (part != '```'
+            and len(part) > 2
+            and part.startswith('`') and part.endswith('`')):
+        return part[1:-1].replace('``', '`')
+    else:
+        return part
+
+
+def normalize_name(name):
```
1st1: Type-annotate new functions pls

edb/repl/__init__.py
vpetrovykh: Done

edb/repl/utils.py
vpetrovykh: Done.

PR #853, opened 2019-11-21
Add more \l REPL commands.

vpetrovykh:
Add `\lm [PATTERN]` command for listing modules, optionally filtered by
name using the specified regexp pattern.
Add `\lmI [PATTERN]` command for case-sensitive pattern matching.

Add `\lr [PATTERN]` command for listing roles, also with option to
filter by name using regexp.
Add `\lrI [PATTERN]` command for case-sensitive pattern matching.

The PATTERN is purely regexp, it should not be quoted in any way, but
instead will be treated as a raw string for a regexp search.

Issue: #179.

Co-authored-by: Eduardo Schettino <schettino72@gmail.com>

### DIFF
--- a/edb/repl/__init__.py
+++ b/edb/repl/__init__.py
@@ -45,6 +45,7 @@ from edb.errors import base as base_errors
 
 from edb.common import term
 from edb.edgeql import pygments as eql_pygments
+from edb.edgeql import quote as eql_quote
 
 from edb.server import buildmeta
 
@@ -60,7 +61,7 @@ STATUSES_WITH_OUTPUT = frozenset({
 
 
 @functools.lru_cache(100)
-def is_multiline_text(text):
+def is_multiline_text(text) -> bool:
     text = text.strip()
 
     if text in Cli.exit_commands:
@@ -80,7 +81,7 @@ def is_multiline_text(text):
 
 
 @pt_filters.Condition
-def is_multiline():
+def is_multiline() -> bool:
     doc = pt_app.get_app().layout.get_buffer_by_name(
         pt_enums.DEFAULT_BUFFER).document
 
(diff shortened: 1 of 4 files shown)


vpetrovykh:
Special thanks to @schettino72 on whose PR #672 this work is based.

edb/repl/utils.py
```diff
@@ -124,3 +124,17 @@ def normalize_name(name: str) -> str:
...
+
+
+def get_filter_based_on_pattern(pattern: str,
+                                field: str = '.name',
+                                flag: str = '') -> str:
+    if flag:
+        flag = f'(?{flag})'
+
+    if pattern:
+        return f'FILTER re_test($${flag}{pattern}$$, {field})'
```
1st1: `pattern` must be escaped. For example, your code will break if someone inputs `\l $$`. People can also do an EdgeQL injection this way. Make pattern a parameter of the query; pass flags as parameters.

vpetrovykh: I've updated the code to use named query parameters and passing the string as a value to those.

edb/repl/__init__.py
```diff
@@ -289,17 +290,85 @@ def command_connect(self, args: str) -> None:
...
+
+    @_command('lrI', R'\lrI [PATTERN]',
+              'list roles (case sensitive pattern match)')
+    def command_list_roles_i(self, args: str) -> None:
+        self._command_simplelist(
+            args.strip(),
+            query='SELECT name := sys::Role.name',
+            name_field='name',
+            item_name='roles',
+            case_sensitive=True,
```
1st1: To avoid code dup please follow what we did for the `command_describe_object` method

vpetrovykh: OK, I've used the same pattern for "I" flag versions of the specific `\l` commands.

PR #858, opened 2019-11-22
Draw a flame graph SVG during profile data aggregation

ambv:
Now when running `python -m edb.tools.profiling` to aggregate results from
profiling, it will generate three files instead of one:

- a single aggregate .prof file for future use (it's faster to load a single
file);

- a classic pstats text profile Top Calls listing; and

- an SVG flame graph.

The flame graph goes both up and down.  The upper one shows the call stack so
there's relative scaling effects in play.  If that makes it hard to see at
first glance which leaf functions run for the longest, the lower graph shows
the usage graph (reverse call stack).  That presents the relative timing of
a given function, plus percentage-wise which callers call it most often.

Example result:
Screenshot 2019-11-22 17 40 09

### DIFF
--- a/edb/tools/profiling.py
+++ b/edb/tools/profiling.py
@@ -186,7 +186,7 @@ class profile:
             # Mypy is wrong below, `stats` is there on all pstats.Stats objects
             render_svg(ps.stats, svg_path)  # type: ignore
         except ValueError as ve:
-            print("Cannot display flame graph: {ve}", file=sys.stderr)
+            print(f"Cannot display flame graph: {ve}", file=sys.stderr)
         print(
             f"Processed {success + failure} files, {failure} failed.",
             file=sys.stderr,
--- a/edb/tools/profiling.py
+++ b/edb/tools/profiling.py
@@ -23,6 +23,7 @@ import atexit
 import cProfile
 import dataclasses
 import functools
+import os
 import pathlib
 import pstats
 import sys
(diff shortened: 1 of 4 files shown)


elprans:
Can we use https://github.com/brendangregg/FlameGraph/ to make it zoomable?

1st1:
Awesome! You can expose this as a command, e.g. `edb profiling`.

> Can we use https://github.com/brendangregg/FlameGraph/ to make it zoomable?

Big +1

ambv:
I think I should be able to add zooming to the SVG files we're producing ourselves here. Then we avoid extra dependencies and keep the extra functionality (the reverse graph).

elprans:
@ambv Up to you.  `flamegraph.pl` is a single file, and I wouldn't mind vendoring it in. The dependency on Perl to run it is a valid concern, though.

1st1:
> Perl to run it is a valid concern

Perl is available on pretty much any dev machine :) Doesn't git still depends on it?

tailhook:
> @ambv Up to you. `flamegraph.pl` is a single file, and I wouldn't mind vendoring it in. The dependency on Perl to run it is a valid concern, though.

There's also rust version of the tool :) https://crates.io/crates/inferno
Unfortunately they don't publish binaries but `cargo install inferno` should work.

ambv:
An example (click through for a zoomable and searchable version):

ambv:
This is ready.

PR #861, opened 2019-11-23
Add \lT command listing scalar types.

vpetrovykh:
Add `\lT [PATTERN]` command for listing scalar types, optionally
filtered by name using the specified regexp pattern.
Add `\lTI [PATTERN]` command for case-sensitive pattern matching.

Issue #179

### DIFF
--- a/edb/repl/__init__.py
+++ b/edb/repl/__init__.py
@@ -48,8 +48,9 @@ from edb.edgeql import quote as eql_quote
 from edb.server import buildmeta
 
 from . import context
-from . import utils
 from . import render
+from . import table
+from . import utils
 
 
 STATUSES_WITH_OUTPUT = frozenset({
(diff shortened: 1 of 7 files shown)


mypy.ini
```diff
@@ -140,3 +140,11 @@ ignore_errors = False
 [mypy-edb.tools.profiling]
 follow_imports = True
 ignore_errors = False
+
+[mypy-edb.repl.table]
```
elprans: Let's use strict mode for new code

vpetrovykh: Updated.

edb/repl/__init__.py
```diff
@@ -362,6 +363,80 @@ def _command_simplelist(self, pattern: str, query: str,
...
 
+    @_command('lT', R'\lT [PATTERN]', 'list scalar types')
+    def command_list_scalar_types(self, args: str, *,
+                                  case_sensitive: bool = False) -> None:
+        self._list_scalar_types(args.strip(), case_sensitive=case_sensitive)
+
+    @_command('lTI', R'\lTI [PATTERN]',
+              'list scalar types (case sensitive pattern match)')
+    def command_list_scalar_types_i(self, args: str) -> None:
+        self.command_list_scalar_types(args, case_sensitive=True)
```
```suggestion
        self._list_scalar_types(args.strip(), case_sensitive=True)
```

edb/repl/__init__.py
```diff
@@ -362,6 +363,80 @@ def _command_simplelist(self, pattern: str, query: str,
                 print(f'No {item_name} found matching '
                       f'{eql_quote.quote_literal(pattern)}')
 
+    @_command('lT', R'\lT [PATTERN]', 'list scalar types')
+    def command_list_scalar_types(self, args: str, *,
+                                  case_sensitive: bool = False) -> None:
```
```suggestion
```

edb/repl/__init__.py
```diff
@@ -362,6 +363,80 @@ def _command_simplelist(self, pattern: str, query: str,
...
+        self._list_scalar_types(args.strip(), case_sensitive=case_sensitive)
```
```suggestion
        self._list_scalar_types(args.strip(), case_sensitive=False)
```

edb/repl/table.py
```diff
@@ -0,0 +1,170 @@
...
+
+class ColumnSpec(NamedTuple):
+    field: str
+    title: str
+    width: int
+    align: str
+
+
+def render_table(
+    title: str,
```
1st1: I'd make them all keyword-only

edb/repl/table.py
```diff
@@ -0,0 +1,170 @@
...
+    data: Iterable[edgedb.Object],
+    max_width: int = 0,
+    *,
+    file: typing.TextIO = sys.stdout,
+) -> None:
+    if max_width == 0:
+        return
+
+    # find the amount of space available for text
+    avail_width = max(0, max_width - (3 * len(columns) - 1))
```
1st1: Pls add a comment to explain this formula

edb/repl/table.py
```diff
@@ -0,0 +1,170 @@
...
+import typing  # needed for linting TextIO
+
+import edgedb
+
+
+class ColumnSpec(NamedTuple):
+    field: str
+    title: str
+    width: int
+    align: str
```
1st1: Use https://mypy.readthedocs.io/en/latest/literal_types.html#literal-types -- a literal type with `left`, `right`, `center` values

vpetrovykh: Done

edb/repl/table.py
vpetrovykh: Comment added.

edb/repl/table.py
vpetrovykh: Sure.

PR #862, opened 2019-11-23
Set rlimit for max open files on server start if necessary

1st1:


edb/server/main.py
```diff
@@ -374,10 +375,21 @@ def server_options(func):
...
-
     logsetup.setup_logging(kwargs['log_level'], kwargs['log_to'])
     exceptions.install_excepthook()
 
+    try:
+        fno_limits = resource.getrlimit(resource.RLIMIT_NOFILE)
+    except resource.error:
+        logger.warning('could not read RLIMIT_NOFILE')
+    else:
+        if fno_limits[0] < 2000:
```
elprans: Can we define this as a constant, ideally with a comment.

1st1: done

tailhook:
I think it never works on linux, even for root:
```
>>> resource.setrlimit(
...     resource.RLIMIT_NOFILE,
...     (2048, resource.RLIM_INFINITY))
  File "<stdin>", line 3, in <module>
ValueError: not allowed to raise maximum limit
```

1st1:
> I think it never works on linux, even for root:

Alright, I'll condition the code to only run on macOS.

elprans:
No, the error is because you've tried to raise the hard limit.  If you only raise the soft limit, it works:

```
Python 3.8.0 (default, Nov 20 2019, 09:01:43)
[GCC 9.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import resource
>>> resource.getrlimit(resource.RLIMIT_NOFILE)
(1024, 524288)
>>> soft, hard = resource.getrlimit(resource.RLIMIT_NOFILE)
>>> resource.setrlimit(resource.RLIMIT_NOFILE, (2048, hard))
>>> resource.getrlimit(resource.RLIMIT_NOFILE)
(2048, 524288)
```

PR #865, opened 2019-11-23
server: Only set rlimit nofile on macos

1st1:
Setting it on Linux seems to be prohibited anyways.

### DIFF
--- a/edb/server/defines.py
+++ b/edb/server/defines.py
@@ -30,7 +30,7 @@ EDGEDB_VISIBLE_METADATA_PREFIX = r'EdgeDB metadata follows, do not modify.\n'
 # By default, at least on macOS, the max number of open FDs
 # is 256, which is low and can cause 'edb test' to hang.
 # We try to bump the rlimit on server start if pemitted.
-EDGEDB_RLIMIT_NOFILE = 2048
+EDGEDB_MIN_RLIMIT_NOFILE = 2048
 
 
 _MAX_QUERIES_CACHE = 1000
--- a/edb/server/main.py
+++ b/edb/server/main.py
@@ -313,6 +313,22 @@ def run_server(args):
         cluster.stop()
 
 
+def bump_rlimit_nofile() -> None:
+    try:
+        fno_soft, fno_hard = resource.getrlimit(resource.RLIMIT_NOFILE)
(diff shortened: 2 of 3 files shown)


tailhook:
It's not prohibited per se. You can't set a hard limit to unlimited as *root* (there is a system limit, don't remember which one). And you can't bump hard limit as a *non-root*.

Other than that. I think it's not common to do that in Linux in apps. There are systemd and docker which do the task (docker bumps it to 64Ki by default).

So I'm okay with changes

elprans:
@tailhook If we do `setrlimit` at all, we should do it consistently on all platforms.

1st1:
> @tailhook If we do setrlimit at all, we should do it consistently on all platforms.

Couple of thoughts:

1. We only want to do this if the current limit is too low and we know it might cause problems.

2. I agree we should do it consistently as then people on Linux & macOS will have the same problems / same mode of operation -- easier for us to troubleshoot. I'll update the PR.

tailhook:
> @tailhook If we do setrlimit at all, we should do it consistently on all platforms.

Okay, let me elaborate:

1. It's hard limit that is a problem on OS X, right?
2. No way ot bump hard limit on Linux.

So no setting rlimit, right? We can print a warning. And set limits on test runner on OS X (I don't think there are many users running OS X on production). Is that consistent enough?

> We only want to do this if the current limit is too low and we know it might cause problems.

What is "too low"? Nginx and redis compares rlimit to max connections. This makes sense. But 8192?

elprans:
> It's hard limit that is a problem on OS X, right?

No, we're bumping into the *soft* limit, which is easy to raise.

> What is "too low"?

Less than 2048, or whatever we define as the minimum, which might be dynamic, based on the maximum number of server connections, like nginx.

1st1:
>> It's hard limit that is a problem on OS X, right?
>
> No, we're bumping into the soft limit, which is easy to raise.

Correct. Soft limits exist to be bumped :)

tailhook:
> No, we're bumping into the soft limit, which is easy to raise.

Okay, then you should do getrlimit and keep hard limit the same, instead of infinity. That works on linux too.

1st1:
@elprans see the updated

PR #870, opened 2019-11-25
Replace uuid.UUID with pgproto.UUID

ambv:
To this effect, `edb.common.uuidgen` grows a UUID type and three factory functions (`uuid3`, `uuid4`, and `uuid5`).  This is functionally equivalent but shaved 32.2 million function calls and 0.5% of run time in unit tests.

~~Note: there is still a lot of `uuid.UUID` objects in EdgeDB, I suspect that most are due to `uuidgen.uuid1mc()` which AFAICT cannot be replaced with `pgproto.UUID`. Case in point, the number of calls to `uuid.UUID.__hash__` fallen from 48 mil calls to 34 mil calls with this change but that's still far from zero.~~ Update: `uuid1mc()` calls now also wrap `pgproto.UUID`.


edb/server/compiler/compiler.py
```diff
@@ -252,6 +253,7 @@ def _process_delta(self, ctx: CompileContext, delta, schema):
 
         return schema, delta
 
+    @profile(prefix="_uuid_new", dir="/tmp/", reuse=False)
```
1st1: Probably don't want this to be committed?

ambv:
FTR, I'll be replacing `uuid1mc` instances as well.

1st1:
> FTR, I'll be replacing uuid1mc instances as well.

In this PR? I'll wait with reviewing until you push an update then.

edb/common/uuidgen.py
```diff
@@ -18,12 +18,42 @@
...
+    # Note: cannot use pgproto.UUID since it's UUID v1
     node = int.from_bytes(os.urandom(6), byteorder='little') | (1 << 40)
     return uuid.uuid1(node=node)
+
+
+# type-ignores below because the first argument to uuid.UUID is a string
+# called `hex` which is not something that pgproto.UUID supports.
+
+
+def uuid3(namespace: uuid.UUID, name: str) -> uuid.UUID:
```
1st1: I'm just curious, could you use `-> UUID` instead of `uuid.UUID` here?

edb/schema/_types.py
```diff
@@ -4,28 +4,30 @@
...
+    'std::int32': uuidgen.UUID('00000000-0000-0000-0000-000000000104'),
+    'std::int64': uuidgen.UUID('00000000-0000-0000-0000-000000000105'),
+    'std::float32': uuidgen.UUID('00000000-0000-0000-0000-000000000106'),
+    'std::float64': uuidgen.UUID('00000000-0000-0000-0000-000000000107'),
+    'std::decimal': uuidgen.UUID('00000000-0000-0000-0000-000000000108'),
+    'std::bool': uuidgen.UUID('00000000-0000-0000-0000-000000000109'),
+    'std::datetime': uuidgen.UUID('00000000-0000-0000-0000-00000000010a'),
+    'std::local_datetime': uuidgen.UUID(
+        '00000000-0000-0000-0000-00000000010b'
+    ),
```
1st1: You should actually update the `edb/tools/gen_types.py` command -- it's used to generate this file.

I'd suggest to update it to do this (to make sure we don't need to break lines to fit in 79):

```python
from edb.common import uuidgen

UUID = uuidgen.UUID

TYPE_IDS = {
    'anytype': UUID('00000000-0000-0000-0000-000000000001'),
    'anytype': UUID('00000000-0000-0000-0000-000000000001'),
   ...
```

edb/common/uuidgen.py
ambv: "UUID" is seen as a type alias to uuid.UUID anyway. I wanted to make this explicit that Mypy does not see our UUIDs at all.

edb/schema/_types.py
ambv: OK, I'll do that.

edb/common/uuidgen.py
1st1: Got it, thanks!

ambv:
```
asyncpg 0.19:       1147781485 function calls (1125288168 primitive calls) in 825.407 seconds
asyncpg 0.20:       1114944817 function calls (1096078482 primitive calls) in 814.988 seconds
asyncpg 0.20+uuid1: 1045854326 function calls (1028864057 primitive calls) in 791.847 seconds
```

Calls to `builtins.hash()` significantly down:
- asyncpg 0.19: 57.6 mil
- asyncpg 0.20: 43 mil
- asyncpg 0.20+uuid1: 9 mil 

Tests runtime consistently >13s down. `__hash__` is gone from the flame graph.

Current flame graph on `_compile_ql_query()`:
[](http://langa.pl/random/graphs/compile-ql-query-2019-11-25.svg)

1st1:
So `UUID.__hash__` calls is no longer a thing?

ambv:
Not a single trace within this profile.

PR #876, opened 2019-11-26
Add more \l commands.

vpetrovykh:
Add `\lc [PATTERN]` command for listing casts, optionally filtered by
affected type names using the specified regexp pattern.
Add `\lcI [PATTERN]` command for case-sensitive pattern matching.

Add `\lt [PATTERN]` command for listing object types, optionally filtered by
name using the specified regexp pattern.
Add `\ltI [PATTERN]` command for case-sensitive pattern matching.

Issue: #179

### DIFF
--- a/edb/repl/__init__.py
+++ b/edb/repl/__init__.py
@@ -58,6 +58,9 @@ STATUSES_WITH_OUTPUT = frozenset({
     'GET MIGRATION', 'DESCRIBE',
 })
 
+# Maximum width for rendering output and tables.
+MAX_WIDTH = 120
+
 
 @functools.lru_cache(100)
 def is_multiline_text(text) -> bool:
@@ -337,7 +340,7 @@ class Cli:
             flag = 'i'
 
         filter_clause, qkw = utils.get_filter_based_on_pattern(
-            pattern, name_field, flag)
+            pattern, [name_field], flag)
 
         try:
             result, _ = self.fetch(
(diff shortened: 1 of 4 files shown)


edb/repl/__init__.py
```diff
@@ -435,6 +435,151 @@ def _list_scalar_types(
...
+                            width=1, align='left'),
+                        table.ColumnSpec(
+                            field='kind', title='Type of Cast',
+                            width=1, align='left'),
+                        table.ColumnSpec(
+                            field='volatility', title='Volatility',
+                            width=1, align='left'),
+                    ],
+                    data=result,
+                    max_width=min(max_width, 120),
```
1st1: Please move 120 into a constant (MAX_TABLE_WIDTH or something)

edb/repl/utils.py
```diff
@@ -127,16 +127,21 @@ def normalize_name(name: str) -> str:
...
         flag = f'(?{flag})'
 
     qkw = {}
 
     if pattern:
-        qkw[f're_{field}'] = flag + pattern
-        return f'FILTER re_test(<str>$`re_{field}`, {field})', qkw
+        # the pattern is the same for all fields, so we only use one
+        # variable
+        qkw[f're_{fields[0]}'] = flag + pattern
```
1st1: Please replace `f're_{fields[0]}'` with `'re_filter'`. This business with using the first element of the `fields` mapping for a parameter name is extremely confusing.

edb/repl/utils.py
```diff
@@ -127,16 +127,21 @@ def normalize_name(name: str) -> str:
...
+        result = []
+        for field in fields:
+            result.append(f're_test(<str>$`re_{fields[0]}`, {field})')
+        return f'FILTER {" OR ".join(result)}', qkw
 
     else:
         # no FILTER clause necessary
```
1st1: `return '', {}`

edb/repl/utils.py
```diff
@@ -127,16 +127,21 @@ def normalize_name(name: str) -> str:
...
 
     qkw = {}
 
     if pattern:
-        qkw[f're_{field}'] = flag + pattern
-        return f'FILTER re_test(<str>$`re_{field}`, {field})', qkw
+        # the pattern is the same for all fields, so we only use one
+        # variable
+        qkw[f're_{fields[0]}'] = flag + pattern
+        result = []
```
1st1:
```suggestion
        filters = []
```

edb/repl/__init__.py
vpetrovykh: It's also used for other output formatting, so I'll just call it `MAX_WIDTH` and add a short comment.

edb/repl/utils.py
vpetrovykh: Yeah, at this point that's probably better. Originally I just thought that it made sense to correlate the name of the regexp parameter with the name of the prop it's applied to, but with this update it's no longer the case.

edb/repl/utils.py
vpetrovykh: OK

edb/repl/utils.py
vpetrovykh: Done

PR #881, opened 2019-11-26
doc: Fixes Map -> Headers, byte[16] -> uuid

tailhook:
`Map` is probably some legacy it isn't defined anywhere.

`byte[16]` where used as type descriptor id makes it more clear that it
refers to a type in type description array (or list of blocks). We use
uuid in type descriptor docs anyway.

### DIFF
--- a/docs/internals/protocol/overview.rst
+++ b/docs/internals/protocol/overview.rst
@@ -52,6 +52,10 @@ The following data types are used in the descriptions:
                int16  key;
                bytes  value;
            };
+    * - ``uuid``
+      - a simple array of 16 bytes with no length prefix, equivalent to
+        ``byte[16]``
+
 
 
 Message Format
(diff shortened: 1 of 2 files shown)


docs/internals/protocol/overview.rst
```diff
@@ -52,6 +52,10 @@ The following data types are used in the descriptions:
                int16  key;
                bytes  value;
            };
+    * - ``uuid``
+      - a simple array of 16 bytes with no length prefix, equivalent to
+        `byte[16]`
+
```
1st1: I think you need to shift this new block one or two spaces to the right. Align with line 41

tailhook: I've fixed indentation earlier. This is just single backticks against double backticks that driving me crazy when switching between rst and md :)

1st1: Yeah :)  We added a special test for that because we were so tired of hunting single backticks down.

Issue #884, opened 2019-11-27
Raise a clear error if the data dir is incompatible


elprans:
When we change the internals that affect the layout of things in Postgres, the old data dirs become broken in subtle ways, so there must be a clear check for this.  Basically, an equivalent of catalog version in Postgres.

PR #887, opened 2019-11-27
Add an explicit database instance compatibility check

elprans:
EdgeDB now checks whether a given database instance is compatible with
the currently running version of the server.  There are two checks:
major version (which is implied to be incompatible), and an additional
"format version" check which is intended to protect from incompatible
changes happening during development.

Fixes: #884.

### DIFF
--- a/edb/server/defines.py
+++ b/edb/server/defines.py
@@ -26,6 +26,9 @@ EDGEDB_SUPERUSER_DB = 'edgedb'
 EDGEDB_ENCODING = 'utf-8'
 EDGEDB_VISIBLE_METADATA_PREFIX = r'EdgeDB metadata follows, do not modify.\n'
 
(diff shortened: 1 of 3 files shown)


edb/server/defines.py
```diff
@@ -26,6 +26,9 @@
 EDGEDB_ENCODING = 'utf-8'
 EDGEDB_VISIBLE_METADATA_PREFIX = r'EdgeDB metadata follows, do not modify.\n'
 
+# Increment this whenever the database layout or stdlib changes.
+EDGEDB_CATALOG_VERSION = 20191127
```
1st1: I'd increase the resolution and add hours/minutes.

```suggestion
EDGEDB_CATALOG_VERSION = 2019_11_27_00_00
```

1st1: (in case a few incompatible changes are made within one day)

PR #890, opened 2019-11-28
Write system config override _before_ the connection is restarted

elprans:
The change to the `listen_addresses` (and `listen_port`) configuration
parameters triggers a management port restart, which drops the backend
connection.  Make sure we persist the configuration change _before_ that
happens.


edb/server/dbview/dbview.pyx
```diff
@@ -349,6 +349,9 @@ cdef class DatabaseIndex:
             )
             op_value = op.coerce_value(op_value, allow_missing=allow_missing)
 
+        self._sys_config = op.apply(config.get_settings(), self._sys_config)
```
1st1: Can you add a comment here?

PR #892, opened 2019-11-28
Add a hint to REPL when a trailing ";" causes an error.

vpetrovykh:
Special REPL commands don't require a trailing ";". Many commands take a
name or a pattern and interpret ";" as a valid part of the name/pattern.
This typically results in an emty result or an error. When this occurs
REPL should suggest omitting the trailing ";".

### DIFF
--- a/edb/repl/__init__.py
+++ b/edb/repl/__init__.py
@@ -454,6 +454,7 @@ class Cli:
             else:
                 print(f'No {item_name} found matching '
                       f'{eql_quote.quote_literal(pattern)}')
+                self._command_semicolon_hint(pattern)
 
     def _render_sdl(self, sdl: str) -> None:
         desc_doc = pt_document.Document(sdl)
@@ -524,11 +525,13 @@ class Cli:
         flags: AbstractSet[str],
         arg: Optional[str]
     ) -> None:
-        name = arg
         verbose = 'verbose' in flags
 
+        # normalize name
+        name = utils.normalize_name(arg or '')
+
         if not name:
-            print(f'The name {name!r} is not valid', flush=True)
+            print(f'The name {arg!r} is not valid', flush=True)
             return
 
         try:
@@ -539,6 +542,13 @@ class Cli:
                 ''',
                 json=False
             )
+        except edgedb.InvalidReferenceError as exc:
+            render.render_exception(self.context, exc)
+            # if the error itself doesn't have a hint, render our
+            # generic hint
+            if not exc._hint:
+                self._command_semicolon_hint(arg)
+
         except edgedb.EdgeDBError as exc:
             render.render_exception(self.context, exc)
         else:
@@ -640,6 +650,7 @@ class Cli:
                 if pattern:
                     print(f'No scalar types found matching '
                           f'{eql_quote.quote_literal(pattern)}.')
+                    self._command_semicolon_hint(arg)
                 elif 'system' not in flags:
                     print(R'No user-defined sclar types found. Try \lTS.')
 
@@ -702,6 +713,7 @@ class Cli:
                 if pattern:
                     print(f'No object types found matching '
                           f'{eql_quote.quote_literal(pattern)}.')
+                    self._command_semicolon_hint(arg)
                 elif 'system' not in flags:
                     print(R'No user-defined object types found. Try \ltS.')
 
@@ -770,8 +782,9 @@ class Cli:
                 if pattern:
                     print(f'No casts found matching '
                           f'{eql_quote.quote_literal(pattern)}.')
+                    self._command_semicolon_hint(arg)
                 else:
-                    print(f'No casts found.')
+                    print('No casts found.')
 
     def _command_psql(
         self,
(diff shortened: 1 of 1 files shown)


edb/repl/__init__.py
```diff
@@ -617,6 +622,7 @@ def _command_list_scalar_types(
             )
         except edgedb.EdgeDBError as exc:
             render.render_exception(self.context, exc)
+            self._command_semicolon_hint(arg)
```
1st1: Semicolons shouldn't be the cause of an error. They can only be if the input isn't escaped properly. What's the use case that triggers a hint in this except clause?

vpetrovykh: Oh, you're right in case of `\l` commands. The only way it can cause an error is in `\d` because the `InvalidReferenceError` is raised for names that can't be found.

I've removed this from lists' error handling.

edb/repl/__init__.py
```diff
@@ -541,6 +544,7 @@ def _command_describe_object(
...
```
1st1: Drop this

edb/repl/__init__.py
```diff
@@ -704,6 +709,7 @@ def _command_list_object_types(
                           f'{eql_quote.quote_literal(pattern)}.')
                 elif 'system' not in flags:
                     print(R'No user-defined object types found. Try \ltS.')
+                self._command_semicolon_hint(arg)
```
1st1: This should be in the `if pattern` block

edb/repl/__init__.py
```diff
@@ -642,6 +646,7 @@ def _command_list_scalar_types(
                           f'{eql_quote.quote_literal(pattern)}.')
                 elif 'system' not in flags:
                     print(R'No user-defined sclar types found. Try \lTS.')
+                self._command_semicolon_hint(arg)
```
1st1: This should be in the `if pattern` block

edb/repl/__init__.py
```diff
@@ -772,6 +778,7 @@ def _command_list_casts(
                           f'{eql_quote.quote_literal(pattern)}.')
                 else:
                     print(f'No casts found.')
+                self._command_semicolon_hint(arg)
```
1st1: ditto

edb/repl/__init__.py
```diff
@@ -850,6 +857,10 @@ def _command_errverbose(
             print('> ' + '\n> '.join(srv_tb.strip().split('\n')))
             print()
 
+    def _command_semicolon_hint(self, arg: Optional[str]) -> None:
+        if arg and arg[-1] == ';':
```
```suggestion
        if arg.strip() and arg.strip()[-1] == ';':
```

1st1:
FWIW I'd issue a warning if `;` is in the pattern regardless of whether it is trailing or not.

edb/repl/__init__.py
```diff
@@ -772,6 +778,7 @@ def _command_list_casts(
                           f'{eql_quote.quote_literal(pattern)}.')
                 else:
                     print(f'No casts found.')
```
```suggestion
                    print('No casts found.')
```

edb/repl/__init__.py
1st1: Ah, alright, just read your remark re `InvalidReferenceError` -- I think `InvalidReferenceError` needs to be handled with a dedicated `except InvalidReferenceError` clause and a nice error message.  And a hint if `;` was there.

edb/repl/__init__.py
vpetrovykh: OK

edb/repl/__init__.py
vpetrovykh: OK

edb/repl/__init__.py
vpetrovykh: The first `arg` mention ensures that `arg` is neither None nor `''`, both of which would cause index errors for `arg[-1]`. I've added the `strip`.

edb/repl/__init__.py
vpetrovykh: Done

edb/repl/__init__.py
vpetrovykh: OK

edb/repl/__init__.py
vpetrovykh: OK

edb/repl/__init__.py
```diff
@@ -850,6 +864,10 @@ def _command_errverbose(
             print('> ' + '\n> '.join(srv_tb.strip().split('\n')))
             print()
 
+    def _command_semicolon_hint(self, arg: Optional[str]) -> None:
+        if arg and arg.strip()[-1] == ';':
```
1st1: This can be an IndexError unless you do it this way (I suggested this exact code before):

```suggestion
        if arg.strip() and arg.strip()[-1] == ';':
```

edb/repl/__init__.py
1st1: This line is no longer needed since you handle it now in the `except edgedb.InvalidReferenceError as exc` above.

PR #897, opened 2019-11-29
Make it easier to invoke the full compiler directly in scripts

elprans:
Add `compiler.compile_edgeql_script()` as a slight modification of the
existing `compiler.compile_bootstrap_script()` to make it easier to
invoke the compiler directly.

### DIFF
--- a/edb/common/devmode.py
+++ b/edb/common/devmode.py
@@ -129,7 +129,7 @@ def read_dev_mode_cache(cache_key, path):
 
     if full_path.exists():
         with open(full_path, 'rb') as f:
-            src_hash = f.read(16)
+            src_hash = f.read(len(cache_key))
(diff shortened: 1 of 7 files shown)


1st1:
Pls add a test of compile_edgeql_script (load one of test schemas and compile a query). This is becoming a private API that we need to keep working without surprises.

edb/common/devmode.py
```diff
@@ -129,7 +129,7 @@ def read_dev_mode_cache(cache_key, path):
 
     if full_path.exists():
         with open(full_path, 'rb') as f:
-            src_hash = f.read(16)
+            src_hash = f.read(20)
```
1st1: Should this be a constant somewhere?

PR #899, opened 2019-11-29
Fix comparison of collections of distinct (but compatible) types

elprans:
The current recursive resolution of operators on collections incorrectly
assumes that a mere presence of a valid operator for the elements of the
collection is enough, and that would be OK, if operators on collections
were actually implemented in terms of the element-wise application of
the correct operator of the element types.  Unfortunately, current
Postgres does not support comparison of arrays or rows of distinct
types, even if a corresponding operator for the element types exists.
Fortunately, since the generic recursive collection operator is defined
as polymorphic on all parameters, the resulting call binding will force
a cast on the arguments to bring them to the same type.

### DIFF
--- a/tests/test_edgeql_expressions.py
+++ b/tests/test_edgeql_expressions.py
@@ -1751,35 +1751,16 @@ class TestExpressions(tb.QueryTestCase):
             # this operation should always be valid and True
             await self.assert_query_result(query, {True})
 
-    # FIXME: once the arrays and tuples work fully with the following
-    # tests, we can integrate tests of the comparison operators for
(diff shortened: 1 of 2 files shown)


tests/test_edgeql_expressions.py
```diff
@@ -1793,29 +1774,6 @@ class TestExpressions(tb.QueryTestCase):
         )
 
     async def test_edgeql_expr_valid_collection_04(self):
-        await self.assert_query_result(
```
1st1: Did you want to just drop all these tests? Are they invalid EdgeQL?

PR #910, opened 2019-11-29
Unwrap singledispatch calls in profiling

ambv:
This requires a replacement tracing singledispatch that keeps track of who called the generic single dispatch function and what implementation ended up being used.  Those recorded traces are saved to sidecar files when `edb.tools.profiling.profile()` is used.

On profile aggregation, the sidecar file data is used to remove `wrapper` functions from the data and replace them with correct links between callers and callees (concrete implementations of the singledispatch function).

Example graph without the singledispatch `wrapper` obscuring everything: http://langa.pl/random/graphs/compile_ast_to_ir-2019-11-29.svg (this is `compile_ast_to_ir` over tests; so that you can compare it with previous graphs).


edb/common/vendor/tracing_singledispatch.py
```diff
@@ -0,0 +1,154 @@
+#
+# This source file is part of the EdgeDB open source project.
+#
+# Copyright 2016-present MagicStack Inc. and the EdgeDB authors.
```
1st1: Replace this comment block with

```
Portions copyright 2019-present MagicStack Inc. and the EdgeDB authors.
Portions copyright 2001-2019 Python Software Foundation.
License: PSFL.
```
```

**Comment by 1st1** [comment]
How to use `patch_functools`? Does it have to be called manually? Can you add a help-docstring somewhere? Or even make ` edb/tools/profiling` a package with a `README.rst`?

**Comment by ambv** [comment]
> How to use `patch_functools`? Does it have to be called manually?
Yes.

> Can you add a help-docstring somewhere? Or even make ` edb/tools/profiling` a package with a `README.rst`?
Good idea actually.

File: edb/tools/profiling/README.md [hunk 2]
```diff
@@ -0,0 +1,190 @@
...
+
+This profiler is using `atexit` to ensure all data is written to disk
+before the program is done.  However, terminating a process with
+a signal may circumvent `atexit` from executing.  Similarly, an
+exception in an `atexit` handler called sooner may prevent our handler
+from running.  If data is visibly incomplete, use `save_every_n_calls=1`
+(which is the default).
+
+This profiler will not be able to trace calls that don't trigger
+`sys.settrace()` callbacks.
```
1st1: 😍

1st1:
Please squash into one commit and merge!

1st1:
@ambv Consider making this a `edb profiling` command; would be slightly more convenient than `python -m edb.tools.profiling`.

The `edb` command is only accessible to us, developers of EdgeDB. It's not installed for EdgeDB users. We can change/break/remove `edb` subcommands in any way we see fit. Backwards compat is not a concern here, just whatever UX is more convenient.

PR #912, opened 2019-11-29
Make EdgeDB work with Mypy 0.750

ambv:


edb/schema/pseudo.py
```diff
@@ -70,7 +70,7 @@ def __eq__(self, other: s_types.Type) -> bool:
                 self.name == other.name)
 
 
-AnyT = typing.TypeVar('AnyT', bound='Any')
+AnyT = typing.TypeVar('AnyT', 'Any', 'AnyTuple')
```
ambv: @1st1, if you changed this instead of doing the other changes to `pseudo.py`, that fixes the problem that Mypy was finding.

1st1: Lukasz, I've already merged https://github.com/edgedb/edgedb/commit/f30744d3ef4bd9f893ab56a5b5d1b705be215077. While it's a bit more impactful I also typed the `pseudo.py` while at it, so let's just close this one.

1st1: Good to know this wasn't a bug in mypy, btw. I thought it was.

PR #913, opened 2019-11-29
Add documentation for name resolution.

vpetrovykh:
Add an explanation of how name resolution in EdgeQL works to EdgeQL
overview.

### DIFF
--- a/docs/edgeql/overview.rst
+++ b/docs/edgeql/overview.rst
@@ -86,6 +86,29 @@ behavior.
 A *simple path* is a path which begins with a set reference.
 
 
+.. _ref_eql_fundamentals_name_resolution:
+
+Name Resolution
+===============
+
+In EdgeQL a name can either be *fully-qualified*, i.e. of the form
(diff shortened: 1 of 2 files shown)


docs/edgeql/overview.rst
```diff
@@ -86,6 +86,29 @@ behavior.
...
+manner:
+
+1) Look for a match to the short name in the current module (typically
+   ``default``, but it can be changed).
+2) Look for a match to the short name in the ``std`` module.
+
+Normally the current module is called ``default``, which is
+automatically created in any new database. It is possible to override
+the current module globally on the session level with a ``SET
+MODULE blah`` :ref:`command <ref_eql_statements_session_set_alias>`. It is
```
ambv: We should decide on a dictionary of placeholder names but my intuition is that they should **not** be `blah`, `foo`, `bar`, and so on.

vpetrovykh: I'll swap it to "my_module" in this example, I think.

1st1: @ambv good point.

PR #917, opened 2019-12-02
Implement std::bigint type

elprans:
The new `std::bigint` scalar type is an arbitrary precision integral
type.  It is implemented as a domain of PostgreSQL `numeric` type that
includes only integers.  The motivation for a distinct type like that is
that many platforms lack a true multi-precision decimal type, but
implement an arbitrary-precision integer type (JavaScript is a prominent
example).

The `<integer-literal>n` (e.g '10n` literals now produce values of
`std::bigint`.  To obtain a `decimal`, value `<float-literal>n`
(e.g `1.0n`) must now be used.

The `std::bigint` and `std::decimal` maintain the relationship that is
similar to the relationship between `std::int64` and `std::float64`.
All sized integer types are implicitly castable to `std::bigint`, and
`std::bigint` itself can implicitly cast to `std::decimal`.  Some
decimal operations, like `round()` and `//` have been redefined to
return `std::bigint` values.


edb/lib/math.edgeql
```diff
@@ -28,50 +28,50 @@ math::abs(x: std::anyreal) -> std::anyreal
...
+math::ceil(x: std::int64) -> std::int64
 {
     SET volatility := 'IMMUTABLE';
-    USING SQL FUNCTION 'ceil';
+    USING SQL FUNCTION 'edgedb.ceil';
 };
 
 
 CREATE FUNCTION
 math::ceil(x: std::float64) -> std::float64
```
tailhook: Why you didn't change this one to return bigint?

1st1: Why? `math::ceil(1e300)` should be `1e300`, not a bigint (which would also be slow).

1st1: 1e300 is an imprecise float, ceiling it shouldn't suddenly yield a precise result. IMO float operations and functions have nothing to do with decimals/bigints.

edb/lib/std/20-genericfuncs.edgeql
```diff
@@ -188,11 +198,11 @@ std::enumerate(
 # ----------
 
 CREATE FUNCTION
-std::round(val: std::int64) -> std::float64
+std::round(val: std::int64) -> std::int64
```
1st1: ouch

edb/lib/math.edgeql
```diff
@@ -28,50 +28,50 @@ math::abs(x: std::anyreal) -> std::anyreal
...
 {
     SET volatility := 'IMMUTABLE';
-    USING SQL FUNCTION 'ceil';
+    USING SQL FUNCTION 'edgedb.ceil';
 };
 
 
 CREATE FUNCTION
-math::ceil(x: std::decimal) -> std::decimal
+math::ceil(x: std::decimal) -> std::bigint
```
1st1: I'm not sure about this. Why doesn't it return decimal (like the float64 variant returns float64)?

edb/lib/std/20-genericfuncs.edgeql
```diff
@@ -55,6 +55,16 @@ std::len(array: array<anytype>) -> std::int64
 # std::sum
 # --------
 
+CREATE FUNCTION
+std::sum(s: SET OF std::bigint) -> std::bigint
```
1st1: You need to document all new signatures. Like in this case you should update `funcops/set.rst`.

edb/lib/math.edgeql
elprans: float variant returns float, because we have no other choice. float64 has a vastly larger range than int64.  decimal -> bigint does not have this problem, as bigint is able to represent all integer parts of decimal.  `ceil(float) -> int` in Python 3.

1st1: So you're breaking the symmetry in variants of the `ceil` function just because we can?

I mean in order to get an `anyint` for `ceil(0.1)` you need an explicit cast as its result is `float64`. For `ceil(0.1n)` you suddenly don't need a cast, it's result is `bigint`.  This asymmetry will propagate through arithmetic expressions, i.e. `(ceil(0.1) + 1) IS anyint` is `false`, but `(ceil(0.1n) + 1n) is anyint` is `true`. 

While this isn't a big deal I don't think this kind of asymmetry is good for variants **of the same function**. So I'm -1.

edb/lib/math.edgeql
tailhook: That's what python3 does:
```
>>> type(math.ceil(1e300))
<class 'int'>
```
Python2 returned a float. So I assume that there was some reasoning behind that.

docs/datamodel/scalars/numeric.rst
```diff
@@ -77,6 +77,36 @@ from :eql:type:`str` and :eql:type:`json`.
 ----------
 
 
+.. eql:type:: std::bigint
+
+    :index: numeric bigint
+
+    Arbitrary precision integer.
+
+    The EdgeDB philosophy is that using decimal type should be an
```
tailhook: This should be updated for bigint? And the next paragraph?

docs/datamodel/scalars/numeric.rst
```diff
@@ -93,15 +123,13 @@ from :eql:type:`str` and :eql:type:`json`.
 
     All of the following types can be explicitly cast into decimal:
     :eql:type:`str`, :eql:type:`json`, :eql:type:`int16`,
-    :eql:type:`int32`, :eql:type:`int64`, :eql:type:`float32`, and
-    :eql:type:`float64`.
+    :eql:type:`int32`, :eql:type:`int64`, :eql:type:`float32`,
+    :eql:type:`float64`, and :eql:type:`bigint`.
 
-    A decimal type has it's own literal:
+    A decimal literal is a float literal followed by 'n':
```
tailhook: I would say this need a clarification. That either decimal dot is required or `e` ?

docs/edgeql/funcops/numerics.rst
```diff
@@ -237,6 +237,30 @@ Numerics
...
 
 
+.. eql:function:: std::to_bigint(s: str, fmt: OPTIONAL str={}) -> decimal
+
+    :index: parse bigint
+
+    Create a :eql:type:`bigint` value.
+
+    Parse a :eql:type:`bigint` from the input *s* and optional format
+    specification *fmt*.
```
tailhook: A link to a format specification?

edb/lib/std/25-numoperators.edgeql
```diff
@@ -1381,11 +1472,11 @@ std::`//` (n: std::float64, d: std::float64) -> std::float64
 
 
 CREATE INFIX OPERATOR
-std::`//` (n: std::decimal, d: std::decimal) -> std::decimal
+std::`//` (n: std::decimal, d: std::decimal) -> std::bigint
```
vpetrovykh: For similar reasons as for `ceil`, `floor` and `round` this should return `decimal`.

edb/lib/std/25-numoperators.edgeql
```diff
@@ -1497,6 +1599,13 @@ std::`^` (n: std::float64, p: std::float64) -> std::float64 {
 };
 
 
+CREATE INFIX OPERATOR
+std::`^` (n: std::bigint, p: std::bigint) -> std::bigint {
```
vpetrovykh: Power cannot be guaranteed to return non-fractional result, because of negative powers, so the return type is `decimal`.

tests/test_edgeql_expressions.py
```diff
@@ -1346,13 +1361,59 @@ class TestExpressions(tb.QueryTestCase):
...
+                    await self.assert_query_result(
+                        f"""
+                            SELECT ({right} {op} {left}) IS {ldesc.typename};
+                        """,
+                        [True],
+                    )
+
+        for left, ldesc in get_test_items(anynumeric=True):
+            for right in get_test_values(anyint=True):
+                # floor division with anynumeric always results in bigint
```
vpetrovykh: Shouldn't be the case anymore. Type is preserved instead.

tests/test_edgeql_functions.py
```diff
@@ -3006,7 +3023,7 @@ class TestEdgeQLFunctions(tb.QueryTestCase):
         )
 
         await self.assert_query_result(
-            r'''SELECT round(<decimal>1.2) IS decimal;''',
+            r'''SELECT round(<decimal>1.2) IS bigint;''',
```
vpetrovykh: `IS decimal`

tests/test_edgeql_functions.py
```diff
@@ -3652,7 +3669,7 @@ class TestEdgeQLFunctions(tb.QueryTestCase):
         )
 
         await self.assert_query_result(
-            r'''SELECT math::ceil(<decimal>2.5) IS decimal;''',
+            r'''SELECT math::ceil(<decimal>2.5) IS bigint;''',
```
vpetrovykh: `IS decimal`

tests/test_edgeql_functions.py
```diff
@@ -3709,7 +3726,7 @@ class TestEdgeQLFunctions(tb.QueryTestCase):
         )
 
         await self.assert_query_result(
-            r'''SELECT math::floor(<decimal>2.5) IS decimal;''',
+            r'''SELECT math::floor(<decimal>2.5) IS bigint;''',
```
vpetrovykh: `IS decimal`

docs/edgeql/funcops/numerics.rst
elprans: It's below

docs/datamodel/scalars/numeric.rst
elprans: This is literally clarified by an example after the `:`.  I added a `e` example as well.

docs/datamodel/scalars/numeric.rst
elprans: Done

edb/lib/math.edgeql
elprans: OK, changed signatures to retain the type.

edb/lib/math.edgeql
elprans: > So I assume that there was some reasoning behind that.

Mostly historical.  It's likely that pre-Python 3 `math.ceil` predates the `long` type.

docs/datamodel/scalars/numeric.rst
```diff
@@ -93,18 +123,19 @@ from :eql:type:`str` and :eql:type:`json`.
...
 
     .. code-block:: edgeql-repl
 
-        db> SELECT 42n IS decimal;
-        {true}
         db> SELECT 1.23n IS decimal;
         {true}
 
+        db> SELECT 1e+100n IS decimal;
+        {true}
```
1st1: Do you want to add an example of a bigint here? Our syntax isn't common.

edb/api/errors.txt
```diff
@@ -140,5 +140,6 @@
 0x_FF_02_01_00   QueryArgumentError
 0x_FF_02_01_01   MissingArgumentError
 0x_FF_02_01_02   UnknownArgumentError
+0x_FF_02_01_03   InvalidArgumentValueError
```
1st1: You seem to be using this new error for both TypeError and ValueError.  Should we add InvalidArgumentTypeError and carefully separate them?

edb/edgeql/ast.py
```diff
@@ -274,10 +274,10 @@ def from_python(cls, val: typing.Any) -> BaseConstant:
         elif isinstance(val, int):
             return IntegerConstant(value=str(val))
         elif isinstance(val, decimal.Decimal):
-            if val.to_integral_value() == val:
-                return IntegerConstant(value=str(val.to_integral_value()))
+            if val.as_tuple().exponent >= 0:
```
1st1: I'm not sure this change is needed...

elprans:
Is this good to be merged?

PR #922, opened 2019-12-03
std::anynumeric tightening

elprans:
* Prohibit implicit comparisons between anynumeric and anyfloat

Comparing decimals and bigints with floats without an explicit cast is now
prohibited.  This is mostly for consistency with our general stance on the
decimal/float separation, but also because current operators don't work for
collection comparisons, which require a common implicit cast, and there is
none between floats and decimals.

* Prohibit NaN as a std::decimal value

NaN in PostgreSQL numeric is a weird thing.  First of all, none of the 
special IEEE 754 values are supported (i.e. there is no infinity).  Zero 
division always raises an exception, so `0/0` is not a thing either. 
Similarly, other "undefined" operations, such as `log(-1)` are also an 
error.

Given this apparent lack of usefulness, and a significant number of 
potential pitfalls (comparison weirdness, virality etc), it seems better to
just ban the use of NaNs in `std::decimal`.

### DIFF
--- a/edb/lib/std/25-numoperators.edgeql
+++ b/edb/lib/std/25-numoperators.edgeql
@@ -41,11 +41,6 @@
 # decimal casts as both directions can lose precision. Discussion
 # about precision loss of float to numeric casts can be found here:
 # https://www.postgresql.org/message-id/5A937D7E.60305%40anastigmatix.net
-# The practical consequence is that whenever the cast is needed, such
-# as for comparison operators, it is preferable to cast decimal into
-# float and not vice-versa. So we can go with Postgres behavior in
-# regards to the comparison operators (but not arithmetic) with
-# decimal as one of the operands.
 
 # EQUALITY
 
@@ -246,7 +241,7 @@ std::`=` (l: std::decimal, r: std::decimal) -> std::bool {
 
 
 CREATE INFIX OPERATOR
-std::`=` (l: std::decimal, r: std::anyreal) -> std::bool {
+std::`=` (l: std::decimal, r: std::anyint) -> std::bool {
     SET volatility := 'IMMUTABLE';
     USING SQL OPERATOR r'=';
 };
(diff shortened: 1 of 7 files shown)


edb/lib/std/25-numoperators.edgeql
```diff
@@ -1781,8 +1781,21 @@ CREATE CAST FROM std::float32 TO std::bigint {
 
 
 CREATE CAST FROM std::float32 TO std::decimal {
-    SET volatility := 'IMMUTABLE';
-    USING SQL CAST;
+    SET volatility := 'STABLE';
+    USING SQL $$
+        SELECT
+            (CASE WHEN val != 'NaN' AND val != 'Infinity' THEN
```
1st1: Does this handle negative infinity?

tests/test_edgeql_datatypes.py
```diff
@@ -241,6 +241,39 @@ class TestEdgeQLDT(tb.QueryTestCase):
...
+            )
+
+    async def test_edgeql_dt_decimal_02(self):
+        with self.assertRaisesRegex(
+            edgedb.InvalidValueError,
+            "invalid value for std::decimal",
+        ):
+            await self.con.execute(
+                r'''
+                    SELECT <decimal><float64>'Infinity'
```
1st1: Add a negative infinity test

elprans: Done

edb/lib/std/25-numoperators.edgeql
elprans: Fixed.

edb/lib/std/25-numoperators.edgeql
```diff
@@ -1781,8 +1776,24 @@ CREATE CAST FROM std::float32 TO std::bigint {
...
+        SELECT
+            (CASE WHEN val != 'NaN'
+                       AND val != 'Infinity'
+                       AND val != '-Infinity'
+            THEN
+                val::numeric
+            ELSE
+                edgedb._raise_specific_exception(
+                    'invalid_text_representation',
+                    'invalid value for numeric: ' || quote_literal(val),
```
1st1:
```suggestion
                    'invalid value for decimal: ' || quote_literal(val),
```
?

elprans: This gets replaced by `errormech` error normalization.

PR #928, opened 2019-12-04
Only trace singledispatch while profiling is enabled

ambv:
This solves invalid timings which caused misalignment in the flamegraphs.

Part of #694


ambv:
Before:
Screenshot 2019-12-04 at 02 17 12

After:
Screenshot 2019-12-04 18 02 20

PR #932, opened 2019-12-04
edgeql: Fix `.>` for tuples.

vpetrovykh:
The `.>` should be perfectly equivalent to `.` in tuples. In particular
when used with index dereferencing e.g. `SELECT tup.>0.>1`.

### DIFF
--- a/tests/test_edgeql_syntax.py
+++ b/tests/test_edgeql_syntax.py
@@ -1763,6 +1763,19 @@ aa';
         SELECT TUP.0.1n.2;
         """
 
+    def test_edgeql_syntax_path_28(self):
+        """
+        SELECT TUP.1.1;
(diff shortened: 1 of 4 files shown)


tests/test_edgeql_syntax.py
```diff
@@ -1763,6 +1763,19 @@ def test_edgeql_syntax_path_27(self):
         SELECT TUP.0.1n.2;
         """
 
+    def test_edgeql_syntax_path_28(self):
+        """
+        SELECT TUP.1.1;
+        SELECT TUP.>1.1;
+        SELECT TUP.>1.>1;
```
1st1: Please add a functional test.

vpetrovykh: OK

Issue #921, opened 2019-12-03
Semantics of new-line escaping in edgedb


tailhook:
Currently, all spaces are preserved, like in python:
```
edgedb>SELECT "hello \ 
.......       xxx";                    
{'hello       xxx'}
```
But I believe that's legacy and I very much like rust's semantics:
```
rust> println!("{:?}", "hello \
              world")
"hello world"
```
(i.e. stripping all whitespace at the start of the next line

This allows nicely indented long string phrases, and also allows breaking non-whitespace-separated texts:
```
WITH
    url := "http://example.com/\
            some/very/long/path"
```
(which is super-ugly with python semantics)

The downside is that you need to put whitespace before continuation:
```
second_line_indented := "def x():\n    \
                             return 'x'"
```
But with python's semantics, inside and outside indentation typically do not match anyway.

What do you think?

1st1:
+1 for rust semantics.

vpetrovykh:
I'm not convinced that the convenience of writing paragraphs of plain text is actually outweighed by the inconvenience of writing anything with indentation.
Compare the following:
```
WITH a := "\
some text or code
         that has
      indentation"
SELECT a;
```
```
WITH a := "some text or code\n         \
           that has\n      \
           indentation"
SELECT a;
```
I think that we need some mechanism flexible enough to either ignore leading whitespace or to preserve it, both are very useful and valid use-cases.

elprans:
I'm also not convinced that the Rust semantics is strictly better. 

> I think that we need some mechanism

We already have a mechanism: string concatenation (which is optimized away by the constant folder).

```
WITH
    url := "http://example.com/"
           ++ "some/very/long/path"
```

1st1:
@elprans 
> I'm also not convinced that the Rust semantics is strictly better.

@vpetrovykh 
>  Compare the following:

> ```
>  WITH a := "some text or code\n         \
>             that has\n      \
>             indentation"
>  SELECT a;
>  ```

Why don't you just use our multi-line raw string for this? **Using raw literals for code is more preferable anyways**:

```
WITH a := r"some text or code         
            that has      
            indentation"
SELECT a;
```

I personally think that for single-line strings the Rust way is more convenient -- that's what you actually want for formatting long paragraphs of **text**.  For code use raw multi-line raw literals.

But **for text**

```
a := 'Lorem ipsum dolor sit amet, consectetur \
      adipiscing elit, sed do eiusmod tempor'
```

is way more readable than

[...]

PR #935, opened 2019-12-05
edgeql: Update the semantics of line continuation in strings.

vpetrovykh:
Regular strings in EdgeQL will now treat line continuation (`\`
terminating a line) as discarding not only the newline, but also all the
whitespace leading up to the next non-whitespace character. Thus the
following are equivalent:
```
SELECT 'hello world';
SELECT 'hello \
        world';
```

Issue #921.

### DIFF
--- a/edb/edgeql/parser/grammar/expressions.py
+++ b/edb/edgeql/parser/grammar/expressions.py
@@ -20,7 +20,6 @@
 from __future__ import annotations
 
 import collections
-import re
 import typing
 
 from edb.common import parsing, context
@@ -939,8 +938,8 @@ class EscapedStringConstant(Nonterm):
         quote = match.group('Q')
         val = match.group('body')
 
-        # handle line continuations
-        val = re.sub(r'\\\n', '', val)
+        # collapse the whitespace after a line continuation
+        val = lexutils.collapse_newline_whitespace(val)
 
         self.val = qlast.StringConstant(value=val, quote=quote)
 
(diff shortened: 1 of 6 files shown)


edb/edgeql/parser/grammar/expressions.py
```diff
@@ -947,8 +946,8 @@ def reduce_SCONST(self, str_tok):
         quote = match.group('Q')
         val = match.group('body')
 
-        # handle line continuations
-        val = re.sub(r'\\\n', '', val)
+        # collapse the whitespace after a line continuation
+        val = lexutils.to_sql_text(val)
```
elprans: At this point we're very far from any SQL, so I'd choose a different name here, like "collapse_newline_whitespace", which is also more informative as to what's going on.

vpetrovykh: Yeah, I originally thought it was going to be closer to SQL. I renamed the function to `collapse_newline_whitespace`.

edb/edgeql/parser/grammar/lexutils.py
```diff
@@ -151,3 +151,10 @@ def cb(m):
         return chr(int(m.group(g), 16))
 
     return STRING_ESCAPE_RE.sub(cb, st)
+
+
+STRING_TO_SQL_RE = re.compile(r'\\\n\s*')
```
elprans: This should get renamed as well.

vpetrovykh: Done

PR #943, opened 2019-12-05
Add an error hint for incorrect string line continuation.

vpetrovykh:
Detect when a line continuation `\` in a regular string is followed by
space before a newline and provide a custom error message with a hint
for this situation.

### DIFF
--- a/tests/test_edgeql_expressions.py
+++ b/tests/test_edgeql_expressions.py
@@ -3053,11 +3053,39 @@ class TestExpressions(tb.QueryTestCase):
         )
 
         await self.assert_query_result(
+            r'''SELECT 'bb\
+            aa \
+
+
+            bb';
+            ''',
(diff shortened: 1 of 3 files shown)


tests/test_edgeql_expressions.py
```diff
@@ -3052,13 +3052,41 @@ class TestExpressions(tb.QueryTestCase):
...
+                _hint="newline has to immediately follow '\\'"):
+            await self.con.execute(
+                r"SELECT 'bb\   "
+                "\naa';"
+            )
+
+    async def test_edgeql_expr_string_11(self):
+        with self.assertRaisesRegex(
+                edgedb.QueryError,
+                r"invalid string literal: invalid escape sequence '\\ '"):
```
1st1: How about `_hint=r"for a backspace character type \\"`?

PR #947, opened 2019-12-06
Disable days and months units in std::duration

tailhook:
This also removes `duration_get` as it doesn't make sense with this
semantics (we'll return it back when `cal::relativedelta` is alive).

Other than that it looks simpler than I expected.

The controversial part is if we want to allow `<duration>'10 days'` (by normalizing it into seconds). Currently I have it disabled, but technically we can re-enable that in the future (for example `timedelta(days=10)` is perfectly valid and clear in python).

### DIFF
--- a/docs/edgeql/funcops/datetime.rst
+++ b/docs/edgeql/funcops/datetime.rst
@@ -564,25 +564,13 @@ Date and Time
         {<duration>'1:20:45'}
 
 
-.. eql:function:: std::duration_to_seconds(cur: duration) -> float64
+.. eql:function:: std::duration_to_seconds(cur: duration) -> decimal
 
     Return duration as total number of seconds in interval.
 
     .. code-block:: edgeql-repl
 
         db> SELECT duration_to_seconds(<duration>'1 hour');
-        {3600.0}
+        {3600.0d}
         db> SELECT duration_to_seconds(<duration>'10 second 100 millis');
-        {10.1}
-
-
-.. eql:function:: std::duration_to_micros(cur: duration) -> int64
-
-    Return duration as total number of microseconds.
-
-    .. code-block:: edgeql-repl
-
-        db> SELECT duration_to_micros(<duration>'1 hour');
-        {3600000000}
-        db> SELECT duration_to_micros(<duration>'10 second 100 millis');
-        {10100000}
+        {10.1d}
--- a/docs/edgeql/funcops/datetime.rst
+++ b/docs/edgeql/funcops/datetime.rst
@@ -562,3 +562,27 @@ Date and Time
         {<duration>'1:20:45'}
         db> SELECT to_duration(seconds := 4845);
         {<duration>'1:20:45'}
+
+
+.. eql:function:: std::duration_to_seconds(cur: duration) -> float64
+
+    Return duration as total number of seconds in interval.
+
+    .. code-block:: edgeql-repl
+
+        db> SELECT duration_to_seconds(<duration>'1 hour');
+        {3600.0}
+        db> SELECT duration_to_seconds(<duration>'10 second 100 millis');
+        {10.1}
+
+
+.. eql:function:: std::duration_to_micros(cur: duration) -> int64
+
+    Return duration as total number of microseconds.
+
+    .. code-block:: edgeql-repl
+
+        db> SELECT duration_to_micros(<duration>'1 hour');
+        {3600000000}
+        db> SELECT duration_to_micros(<duration>'10 second 100 millis');
+        {10100000}
--- a/docs/edgeql/funcops/datetime.rst
+++ b/docs/edgeql/funcops/datetime.rst
@@ -342,20 +342,22 @@ Date and Time
 
     Truncate the input duration to a particular precision.
 
-    The valid *unit* values are the same as for :eql:func:`datetime_trunc`.
+    The valid *unit* values are:
+    - ``'microseconds'``
+    - ``'milliseconds'``
+    - ``'seconds'``
+    - ``'minutes'``
+    - ``'hours'``
 
     .. code-block:: edgeql-repl
 
         db> SELECT duration_trunc(
-        ...     <duration>'3 days 15:01:22', 'day');
-        {'3 days'}
+        ...     <duration>'15:01:22', 'hours');
+        {'15:00:00'}
 
         db> SELECT duration_trunc(
-        ...     <duration>'15:01:22.306916', 'minute');
+        ...     <duration>'15:01:22.306916', 'minutes');
         {'15:01:00'}
-
-    The usual caveat that :eql:type:`duration` doesn't automatically
-    convert units applies to how truncation works.
 
 
 ----------
--- a/docs/edgeql/funcops/datetime.rst
+++ b/docs/edgeql/funcops/datetime.rst
@@ -47,9 +47,6 @@ Date and Time
     * - :eql:func:`date_get`
       - :eql:func-desc:`date_get`
 
-    * - :eql:func:`duration_get`
-      - :eql:func-desc:`duration_get`
-
     * - :eql:func:`datetime_trunc`
       - :eql:func-desc:`datetime_trunc`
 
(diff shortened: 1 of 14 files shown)


docs/edgeql/funcops/datetime.rst
```diff
@@ -299,82 +296,6 @@ Date and Time
 ----------
 
 
-.. eql:function:: std::duration_get(dt: duration, el: str) -> float64
```
elprans: We should maybe add a function or a cast to bigint returning the number of microseconds in the duration.

edb/pgsql/metaschema.py
```diff
@@ -1478,6 +1478,38 @@ def __init__(self) -> None:
             text=self.text)
 
 
+class DurationInFunction(dbops.Function):
+    """Cast text into duration, ensuring there is no month units"""
+    text = r'''
+        DECLARE
```
elprans: Please make this a pure SQL function.  PL/pgSQL is very slow.  See `StrToBigint` for an example of a similar cast.

edb/lib/cal.edgeql
```diff
@@ -292,7 +292,9 @@ std::`-` (l: cal::local_datetime, r: std::duration) -> cal::local_datetime {
 CREATE INFIX OPERATOR
 std::`-` (l: cal::local_datetime, r: cal::local_datetime) -> std::duration {
```
elprans: We don't seem to have tests for `local_datetime` and `datetime` subtraction.  Please add a couple.

docs/edgeql/funcops/datetime.rst
tailhook: We don't need bigint here, because internally there's only int64 microseconds.

edb/lib/cal.edgeql
tailhook: Just find out that we need to fix python bindings for the test.

tailhook:
@elprans please review new commits.

I think we can push this now, and add test later. I think we have a chicken-egg problem here: we need to update python bindings to use timedelta to make valid tests for `-` and we need this PR for edgedb-python to have comprehensive suite of duration/timedelta tests in edgedb-python.

docs/edgeql/funcops/datetime.rst
```diff
@@ -562,3 +562,27 @@ Date and Time
         {<duration>'1:20:45'}
         db> SELECT to_duration(seconds := 4845);
         {<duration>'1:20:45'}
+
+
+.. eql:function:: std::duration_to_seconds(cur: duration) -> float64
```
1st1: First, I think we should return a decimal here.

Second, do we need this as a function?  Why don't we just enable `<decimal>` and `<float64>` casts for `std::duration`?

docs/edgeql/funcops/datetime.rst
```diff
@@ -562,3 +562,27 @@ Date and Time
...
+
+    .. code-block:: edgeql-repl
+
+        db> SELECT duration_to_seconds(<duration>'1 hour');
+        {3600.0}
+        db> SELECT duration_to_seconds(<duration>'10 second 100 millis');
+        {10.1}
+
+
+.. eql:function:: std::duration_to_micros(cur: duration) -> int64
```
1st1: Not sure this is needed, especially if we have a decimal cast. I think we should try to keep the number of std functions to the functional minimum and avoid adding trivial functions.

docs/edgeql/funcops/datetime.rst
tailhook: Decimal cast is a bit unclear: what units should be? Python users would say seconds but javascript users would expect milliseconds.

edb/pgsql/metaschema.py
```diff
@@ -1478,6 +1478,41 @@ def __init__(self) -> None:
...
+class DurationInFunction(dbops.Function):
+    """Cast text into duration, ensuring there is no month units"""
+    text = r'''
+        SELECT
+            CASE WHEN
+                EXTRACT(MONTH FROM v.column1) != 0 OR
+                EXTRACT(YEAR FROM v.column1) != 0 OR
+                EXTRACT(DAY FROM v.column1) != 0
+            THEN
+                edgedb._raise_specific_exception(
```
1st1: I don't see any `assertRaises` tests to test this cast.

docs/edgeql/funcops/datetime.rst
tailhook: Hm, the idea was that float can have precision issues. Decimal might work. But @elprans specifically asked for microseconds for some reason.

edb/pgsql/metaschema.py
```diff
@@ -1478,6 +1478,41 @@ def __init__(self) -> None:
             text=self.text)
 
 
+class DurationInFunction(dbops.Function):
+    """Cast text into duration, ensuring there is no month units"""
```
vpetrovykh: ...no month or day units.

docs/edgeql/funcops/datetime.rst
elprans: I was thinking about an integer number, which can only be of microseconds, but a fractional number of seconds as a `decimal` works just as well.

edb/pgsql/metaschema.py
```diff
@@ -1478,6 +1478,41 @@ def __init__(self) -> None:
...
+            CASE WHEN
+                EXTRACT(MONTH FROM v.column1) != 0 OR
+                EXTRACT(YEAR FROM v.column1) != 0 OR
+                EXTRACT(DAY FROM v.column1) != 0
+            THEN
+                edgedb._raise_specific_exception(
+                    'invalid_datetime_format',
+                    'invalid input syntax for type duration: '
+                        || quote_literal(val),
+                    '{"hint":"You can''t use days or units ' ||
```
elprans: Please use neutral style here: "Units bigger than days cannot be used for std::duration.".  Also, [PostgreSQL error style guide](https://www.postgresql.org/docs/current/error-style-guide.html) is excellent.

tailhook:
Looks like all the comments are addressed (including the test for `-` operator, it turned out not that hard).

But I'm in doubt that `local_datetime - local_datetime` should return `duration` (as currently in this PR). I think `cal::relativedelta` would make much more sense (because we don't know how many hours passed between dates when we don't know timezone, but we know how many days have passed).

tailhook:
Removed `local_datetime - local_datetime` operator and marked test as `not_implemented`

edb/lib/cal.edgeql
tailhook: Done

edb/lib/std/30-datetimefuncs.edgeql
```diff
@@ -56,32 +56,33 @@ std::datetime_get(dt: std::datetime, el: std::str) -> std::float64
 
 
 CREATE FUNCTION
-std::duration_get(dt: std::duration, el: std::str) -> std::float64
+std::datetime_trunc(dt: std::datetime, unit: std::str) -> std::datetime
```
1st1: This looks like a new undocumented function. Why was it added?

tailhook:
So now we have `local_date - local_date` return relative delta, and `local_datetime - local_datetime` return relative delta (except the type in our to do list so far). The question is what `local_time - local_time` should return.

The current summary is:

Returning `cal::relativedelta`:
1. [upside] Consistent with all `cal::local_*` types
2. [upside] The type is in same module 
3. [downside] It isn't clear are there any circumstances where `time - time` returns non zero days

Returning `cal::duration`:
1. [upside] There are no days field in duration, so it's always clear what is returned 
2. [neutral] Duration can be cast into `cal::relativedelta` so things like `date - date + (time - time)` work fine anyway

Personally I find consistency of "return duration whenever its enough to represent value" is more important "than consistent with all `cal::` types". This is similar to what we have with float vs int: we use int while it's enough to represent what we need (i.e. length of an array) and convert it to a float (or decimal) when we need fractions or mix types (`123 / 2` or `123 + 1.5`).

tailhook:
On the other hand, python doesn't have any arithmetic on time type:
```
>>> time(10) + timedelta(hours=16)
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for +: 'datetime.time' and 'datetime.timedelta'
>>> time(10) - time(8)
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for -: 'datetime.time' and 'datetime.time'
```
So we may drop it for now as well. 

The upside of removing time arithmetic at all is that currently time arithmetic is silently wrapping, which is very controversial behavior:
```
edgedb> select <cal::local_time>'20:12:00' + <std::duration>'5 hours';         
{<local_time>'01:12:00'}
edgedb> select 9223372036854775807 + 1;
NumericOutOfRangeError: std::int64 out of range
```

vpetrovykh:
`local_time` is analogous to modular arithmetic (it actually is exactly modular arithmetic with the modulus being some representation of 24 hours). `LT + 0 = LT + 24h = LT + 1day`. So it operates under the built-in assumption that `24h = 1 day`, which is only true for `relativedelta`, but not true for `duration`. So `local_time` arithmetic implicitly operates on `relativedelta` and will have to implicitly convert `duration` into a `relativedelta` if it were allowed to operate on `duration`. Basically if we allow roll-over in `local_time` arithmetic, we're implicitly using `relativedelta`. If `23:50:00 + 2hours = 01:50:00`, we have necessarily assumed no daylight savings, we've assumed that one day is exactly like the next and is 24 hours.

`local_time` has nothing to do with actual time, it's all about clock behavior. `23:50:00 + 2hours` is *not* asking what time will it be in 2 hours, but it's asking what the clock will show if you advance it by 2 hours. It would be odd to say that it's illegal to advance the clock past midnight in one direction of another. So `relativedelta` is a datatype that deals with clock and calendar values. It emulates turning calendar pages and adjusting a wall-clock. `duration` is attempting to capture information about physical time between two events, disregarding the quirks of representation of that time. Whether daylight savings happened or not, `<duration>'2hours'` means exactly the same amount of time, but it may represent the `<relativedelta>'1 hour'`,  `<relativedelta>'2 hours'`, or `<relativedelta>'3 hours'` (the latter telling you what adjustment you'd be doing to your clock, not actual time passing).

edb/lib/std/30-datetimefuncs.edgeql
```diff
@@ -56,32 +56,33 @@ std::datetime_get(dt: std::datetime, el: std::str) -> std::float64
...
     USING SQL $$
-    SELECT date_part("el", "dt")
+    SELECT date_trunc("unit", "dt")
     $$;
 };
 
 
 CREATE FUNCTION
-std::datetime_trunc(dt: std::datetime, unit: std::str) -> std::datetime
+std::duration_trunc(dt: std::duration, unit: std::str) -> std::duration
```
vpetrovykh: If `duration` is absolute time expressed unambiguously then `duration_trunc` is nonsensical for any unit above 'hours'. In fact, it should not even acknowledge units bigger than 'hours`. It may be superfluous. Why do we need it?

This function is more useful for `relativedelda` in my opinion.

tailhook:
Fixed comments. The only doubtful thing is that postgres' units are:
    - ``'microseconds'``
    - ``'milliseconds'``
    - ``'second'``
    - ``'minute'``
    - ``'hour'``

I.e. mixed plurals and singulars. These work with `datetime_get` and `datetime_trunc`, but I've changed units all to be **plural**. Is it okay for them to be different? (singulars kinda make sense for `EXTRACT` but aren't very justified, and it looks like postgres just reuse them for `date_trunc`)

vpetrovykh:
Making all unit names plural is reasonable. It's self-consistent and is slightly easier to remember than a mix.

edb/lib/std/30-datetimefuncs.edgeql
```diff
@@ -56,32 +56,46 @@ std::datetime_get(dt: std::datetime, el: std::str) -> std::float64
...
 {
-    # date_trunc of timestamptz is STABLE in PostgreSQL
-    SET volatility := 'STABLE';
+    SET volatility := 'IMMUTABLE';
     USING SQL $$
-    SELECT date_trunc("unit", "dt")
+    SELECT CASE WHEN "unit" in ('microseconds', 'milliseconds')
+        THEN date_trunc("unit", "dt")
+        WHEN "unit" in ('seconds', 'minutes', 'hours')
+        THEN date_trunc(substring("unit" for length("unit")-1), "dt")
```
vpetrovykh: As far as I can tell Postgres accepts both singular and plural units, so there's no need for truncating the "s" from the end of the unit name.

tailhook: Yes, you're right. Unfortunately, I don't see it documented anywhere :(

vpetrovykh: Oh, the docs only list "microseconds" and "milliseconds" as plurals, the rest are singular. That's annoying. Actually, at this point if I had to choose I'd probably be more inclined to pick singular unit format. It makes no difference for `duration`, but `relativedelta` would benefit from using singular units because it's a little annoying to use "centuries" and guess the correct plural of "millennium" ("millennia" is correct Latin plural, but "millenniums" is also acceptable in English).

PR #953, opened 2019-12-09
WIP: Add initial rust module

tailhook:
Here is the first attempt. Currently it just imports an empty module. Basically, it's to help up ensure that the module builds well on all our workstations, CI, and packages well.

Questions: how to put the module into `PYTHONPATH` of the build?
* In current commit, you need `PYTHONPATH=build/rustlib` to make it work
* It looks like cython modules are just put into sources. So should I just copy the module somewhere to `edb/edgeql/_rust_lexer.cpython-bla-bla.so` ?

Notes to myself:
* Update submodule url to public URL before merging in

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "edb/server/pgproto"]
 	path = edb/server/pgproto
 	url = https://github.com/MagicStack/py-pgproto.git
+[submodule "edgedb-rust"]
+	path = edgedb-rust
+	url = https://github.com/edgedb/edgedb-rust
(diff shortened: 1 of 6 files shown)


elprans:
@tailhook Is it possible to use https://pypi.org/project/setuptools-rust/ to integrate this properly?

elprans:
Also: https://github.com/getsentry/milksnake

tailhook:
Add setuptools-rust. Milksnake is for cffi bindings (I think we'd be better with normal cpython bindings they should be faster and we don't have to support multiple pythons).

PR #954, opened 2019-12-10
Omit populating `.parent` for AST nodes

### LINKED ISSUE #938: Get rid of AST.parent
Screenshot_20191205_121201

`AST.parent` is a historical vestige that is used just in a couple of places, most notably, EdgeQL codegen and the AST transformer (which itself isn't used very much).  Look into getting rid of this, as it's up there in the function cost chart (`AST.__init__`)

ambv:
Fixes #938

### DIFF
--- a/edb/common/ast/base.py
+++ b/edb/common/ast/base.py
@@ -192,29 +192,8 @@ class AST(object, metaclass=MetaAST):
                 f'cannot instantiate abstract AST node '
                 f'{self.__class__.__name__!r}')
 
-        object.__setattr__(self, 'parent', None)
+        object.__setattr__(self, 'parent', kwargs.get('parent', None))
         self._init_fields(kwargs)
-
-        # XXX: use weakref here
-        for arg, value in kwargs.items():
-            if hasattr(self, arg):
-                if is_ast_node(value):
-                    object.__setattr__(value, 'parent', self)
-                elif isinstance(value, list):
-                    for v in value:
-                        if is_ast_node(v):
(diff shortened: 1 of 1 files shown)


edb/common/ast/base.py
```diff
@@ -192,30 +192,9 @@ def __init__(self, **kwargs):
                 f'cannot instantiate abstract AST node '
                 f'{self.__class__.__name__!r}')
 
-        object.__setattr__(self, 'parent', None)
+        object.__setattr__(self, 'parent', kwargs.get('parent', None))
```
1st1: I don't think this line is needed, or?..

edb/common/ast/base.py
```diff
@@ -192,30 +192,9 @@ def __init__(self, **kwargs):
...
         self._init_fields(kwargs)
```
1st1: Can you inline this function? It's not used anywhere else but `AST.__init__` is called a lot. Inlining will shave the calls number a bit and is completely safe in this case (even though this is a micro opt).

edb/common/ast/base.py
ambv: The attribute has to exist due to checks in multiple places and without the setter one caller would have to be changed.

edb/common/ast/base.py
ambv: I did but this is a miniscule difference since there is just one call per `__init__` call. It's less readable for dubious benefit.

edb/common/ast/base.py
1st1: What caller?

PR #959, opened 2019-12-12
Fix a bunch of issues with dependencies in SDL.

vpetrovykh:
SDL was very sensitive to definition order to the point where simple
linear dependencies caused failures. Generally, "using" something
(function, annotation, constraint, etc.) before declaring it caused
errors when tring to create an SDL migration. Dependencies on other
modules were similarly broken.

Schema migration tests now attempt to dump the schema into SDL (using
the same code-path as used by `DESCRIBE`). Then the schema produced by
the generated SDL is compared to the original to check correctness of
generated SDL syntax and semantics.

The current version fixes many dependency issues, however, it also
highlighted a couple of deficiencies in other areas, which should be
fixed independently:
1) SDL for index does not support fields, so there's no equivalent to a
valid `SET origexpr := ...` in SDL.
2) There are some cases where `exclusive` constraints get resolved as
`default::std::exclusive` instead of `std::exclusive`. This seems to
happen after the step that deals with SDL conversion to DDL and
dependency resolution. The issue is intermittent given the same input
SDL.

### DIFF
--- a/edb/edgeql/codegen.py
+++ b/edb/edgeql/codegen.py
@@ -810,7 +810,7 @@ class EdgeQLSourceGenerator(codegen.SourceGenerator):
             else:
                 self.write(ident_to_str(node.name.module), '::',
                            ident_to_str(node.name.name))
-        if node.create_if_not_exists:
+        if node.create_if_not_exists and not self.sdlmode:
             self.write(' IF NOT EXISTS')
         if after_name:
             after_name()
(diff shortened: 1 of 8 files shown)


tailhook:
I don't understand from the description: does it allow use before definition? Is it a goal?

vpetrovykh:
Yes, the goal is to make SDL order-independent or at least more so than it was. There are cases where entities may reference each other and a re-arranging is impossible within SDL:
```
# should fail in the current master
module default { 
    type Foo { 
        link bar := (
            SELECT Bar FILTER .name = 'foo'
        ); 
        property name -> str; 
    } 
     
    type Bar {  
        link foo := (
            SELECT Foo FILTER .name = 'bar'
        );  
        property name -> str;  
    }  
} 
```
But the main point is that in a declarative language the order shouldn't matter.

vpetrovykh:
Oooh, I see tests that didn't fail for me locally failed in CI. I'll look at them tomorrow.

elprans:
>     1. SDL for index does not support fields, so there's no equivalent to a
>        valid `SET origexpr := ...` in SDL.
> 
>     2. There are some cases where `exclusive` constraints get resolved as
>        `default::std::exclusive` instead of `std::exclusive`. This seems to
>        happen after the step that deals with SDL conversion to DDL and
>        dependency resolution. The issue is intermittent given the same input
>        SDL.

These don't seem terribly complicated.  Please fix as part of this PR rather than xfailing a whole lot of existing tests.

PR #962, opened 2019-12-13
edgeql: Update SDL syntax allowing setting fields on indexes.

vpetrovykh:
The `index` in SDL now allows to set `origexpr` field explicitly, just
as DDL.

### DIFF
--- a/edb/edgeql/parser/grammar/sdl.py
+++ b/edb/edgeql/parser/grammar/sdl.py
@@ -541,6 +541,7 @@ class AnnotationDeclarationShort(Nonterm):
 #
 sdl_commands_block(
     'CreateIndex',
+    SetField,
     SetAnnotation)
 
 
--- a/edb/schema/indexes.py
+++ b/edb/schema/indexes.py
@@ -162,16 +162,24 @@ class CreateIndex(IndexCommand, referencing.CreateReferencedInheritingObject):
         schema = super()._create_begin(schema, context)
         index = self.scls
 
+        expr = self.get_attribute_value('expr')
+        origexpr = index.get_field_value(schema, 'origexpr')
         # Check if the index doesn't have "origexpr" explicitly specified.
-        if index.get_field_value(schema, 'origexpr') is None:
+        if origexpr is None:
             # Add annotation with the original expression.
-            expr = self.get_attribute_value('expr')
             self.set_attribute_value('origexpr', expr.origtext)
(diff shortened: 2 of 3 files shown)


vpetrovykh:
This is one of the changes that I stumbled upon when dealing with SDL. It's nice and self-contained, so I factored it out.

I'm not sure if we consider `SET origexpr` part of the interface that's supposed to be exposed to the user, so I didn't update the documentation. Let me know if I need to do that, though.

elprans:
> I'm not sure if we consider SET origexpr part of the interface that's supposed to be exposed to the user

Um. Now I'm having second thoughts about this change.  Since we're discarding the usual session module aliases when processing the SDL declarations, `origexpr` should really be used to render the `on (<origexpr>)` part.

vpetrovykh:
Isn't that beside the point? The question ultimately is whether it's legal to set "origexpr" in SDL the same way as DDL. And if not, why not? Basically, what makes it valid to set "origexpr" in DDL, but *invalid in all cases* for SDL?

Incidentally, I should probably come up with a test case that breaks "origexpr" in some horrible way in DDL. Like mismatching the current module and the module of the declared type (and index).

vpetrovykh:
Also, `SET origexpr` was broken in DDL because it failed to override the "expr" value in introspection queries, so that got fixed here as well.

elprans:
> Isn't that beside the point? The question ultimately is whether it's legal to set "origexpr" in SDL the same way as DDL.

DDL uses the session module aliases, and SDL doesn't.  The whole point of expression rewrite that mangles the original expression in the first place is to *record the current module alias*. This is not a problem in SDL.

elprans:
Also, `origexpr` in DDL exists only for the benefit of introspection and reproducible dumps.  It does not make sense at all as a user-facing feature.

vpetrovykh:
Regarding user-facing feature, see my comment about `DESCRIBE AS DDL`. It currently produces the fully-disambiguated expression _and_ a `SET origexpr` to compensate for this. So unless we come up with some clever way of avoiding that, the "origexpr" has to be available to the user on some level in DDL at least.

vpetrovykh:
> DDL uses the session module aliases, and SDL doesn't. The whole point of expression rewrite that mangles the original expression in the first place is to _record the current module alias_.

Fair enough, it may be unnecessary feature in SDL, but give me a little time to see if I can come up with some odd use-case that needs it.

PR #966, opened 2019-12-14
Cache ptrref_from_ptrcls

elprans:
The original intent of this PR was to make `ptrref_from_ptrcls()` easier to cache by reducing its signature.  Doing that, however, made the function slower (since it now must do more work).  Then, while working on fixing #969, I realized that we actually need the `ptrref` -> `ptrcls` cache, so I then implemented the `ptrcls -> ptrref` cache and its inverse.  This brought the `get_movie` benchmark to the status quo:

compile_edgeql_script_webapp_get_movie: Mean +- std dev: 213 ms +- 5 ms

Looking at the profile details for this benchmark, `ptrref_from_ptrcls()` now uses sligthly fewer cycles: -12%, but the overall effect of the cumulative change is negligible.

### DIFF
--- a/edb/edgeql/compiler/context.py
+++ b/edb/edgeql/compiler/context.py
@@ -115,6 +115,34 @@ class PendingCardinality(NamedTuple):
     callbacks: List[PointerCardinalityCallback]
 
 
+class PointerRefCache(
+    Dict[
+        Tuple[s_pointers.PointerLike, s_pointers.PointerDirection],
+        irast.BasePointerRef,
+    ]
+):
+
+    _rcache: Dict[irast.BasePointerRef, s_pointers.PointerLike]
+
+    def __init__(self) -> None:
+        super().__init__()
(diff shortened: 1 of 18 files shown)


edb/edgeql/compiler/context.py
```diff
@@ -115,6 +115,34 @@ class PendingCardinality(NamedTuple):
...
+    ]
+):
+
+    _rcache: Dict[irast.BasePointerRef, s_pointers.PointerLike]
+
+    def __init__(self) -> None:
+        super().__init__()
+        self._rcache = {}
+
+    def __setitem__(
```
1st1: Since you don't have a `__getitem__` maybe it's better to rename this to `set_ptrcls_for_ref`?

elprans: This is a dict subclass, `__getitem__` is inherited.

PR #971, opened 2019-12-16
WIP: Add rust tokenizer

tailhook:
Probably the only important change is that I unescape tokens in the tokenizer itself, rather than in parser. And put unescaped value into AST.  This also means escaping back in codegen.

Tests fail probably because I've not fixed Postgres codegen, but the error is weird.

To do:
* [ ] finally unify RawString vs String, and EscapedString vs String nodes
* [ ] remove traces of the old lexer
* [ ] use (parsed) value from token for int/float/decimal types


tailhook:
Hm, github does't run tests on it :( Current error on tests is:
```
  File "/usr/local/lib/python3.8/dist-packages/click/core.py", line 535, in invoke                                                                  [560/9587]
    return callback(*args, **kwargs)                                                                                                                          
  File "/work/edb/tools/test/__init__.py", line 144, in test                                                                                                  
    result = run()                                                                                                                                            
  File "/work/edb/tools/test/__init__.py", line 248, in _run                                                                                                  
    result = test_runner.run(suite)                                                                                                                           
  File "/work/edb/tools/test/runner.py", line 690, in run                                                                                                     
    tb.setup_test_cases(                                                                                                                                      
  File "/work/edb/testbase/server.py", line 807, in setup_test_cases                                                                                          
    return asyncio.run(_run())                                                                                                                                
  File "/usr/lib/python3.8/asyncio/runners.py", line 43, in run                                                                                               
    return loop.run_until_complete(main)                                                                                                                      
  File "/usr/lib/python3.8/asyncio/base_events.py", line 608, in run_until_complete                                                                           
    return future.result()                                                                                                                                    
  File "/work/edb/testbase/server.py", line 804, in _run                                                                                                      
    g.create_task(controller(                                                                                                                                 
  File "/work/edb/common/taskgroup.py", line 169, in __aexit__                                                                                                
    raise me from None                                                                                                                                        
edb.common.taskgroup.TaskGroupError: unhandled errors in a TaskGroup; 2 sub errors: (InternalServerError)                                                     
 + InternalServerError: invalid syntax (<unknown>, line 1)                                                                                                    
 |   File "/work/edb/testbase/server.py", line 801, in controller                                                                                             
 |     await coro(*args)                                                                                                                                      
 |   File "/work/edb/testbase/server.py", line 830, in _setup_database                                                                                        
 |     await dbconn.execute(setup_script)                                                                                                                     
 |   File "/work/.eggs/edgedb-0.7.0a3-py3.8-linux-x86_64.egg/edgedb/asyncio_con.py", line 97, in execute                                                      
 |     await self._protocol.simple_query(query)                                                                                                               
 |   File "edgedb/protocol/protocol.pyx", line 452, in simple_query                                                                                           
                                                                                                                                                              
                                                                                                                                                              
 + InternalServerError: invalid syntax (<unknown>, line 1)                                                                                                    
 |   File "/work/edb/testbase/server.py", line 801, in controller                                                                                             
 |     await coro(*args)                                                                                                                                      
 |   File "/work/edb/testbase/server.py", line 830, in _setup_database                                                                                        
 |     await dbconn.execute(setup_script)                                                                                                                     
 |   File "/work/.eggs/edgedb-0.7.0a3-py3.8-linux-x86_64.egg/edgedb/asyncio_con.py", line 97, in execute                                                      
 |     await self._protocol.simple_query(query)
 |   File "edgedb/protocol/protocol.pyx", line 452, in simple_query
```

elprans:
> Hm, github does't run tests on it

I just pushed a fix to the workflow, please rebase.

tailhook:
Got working tests. Also added a tiny to-do list into the top comment, thats mostly a cleanup and a bit optimization. I think this PR is ready for a first review.

edb/edgeql/codegen.py
```diff
@@ -548,13 +563,26 @@ def visit_Parameter(self, node: qlast.Parameter) -> None:
...
+                        self.write(d, node.value, d)
+                    return
+        self.write(repr(node.value))
 
     def visit_RawStringConstant(self, node: qlast.RawStringConstant) -> None:
-        if node.quote.startswith('$'):
-            self.write(node.quote, node.value, node.quote)
-        else:
-            self.write('r', node.quote, node.value, node.quote)
+        if node.value.isprintable():
```
elprans: I'm not sure about this.  Current code retains the `r` prefix.  This seems to treat both kinds of strings equally. Why?

edb/edgeql/parser/grammar/expressions.py
```diff
@@ -907,80 +906,16 @@ def reduce_NFCONST(self, *kids):
...
-            raise EdgeQLSyntaxError(
-                f"invalid string literal: invalid line continuation",
-                hint="newline has to immediately follow '\\'",
-                context=str_tok.context)
-
-        quote = match.group('Q')
-        val = match.group('body')
-
-        # collapse the whitespace after a line continuation
-        val = lexutils.collapse_newline_whitespace(val)
```
elprans: The current lexer has a nice property: it's mostly non-destructive.  We can reconstruct the input stream by concatenating the tokens back. This property might be important in some contexts.  @vpetrovykh should know more.

edb/edgeql/parser/grammar/expressions.py
```diff
@@ -1258,7 +1193,8 @@ def reduce_empty(self, *kids):
 
 class Identifier(Nonterm):
     def reduce_IDENT(self, *kids):
-        self.val = kids[0].val
+        self.val = kids[0].clean_value
+        assert self.val != 'constraint', self.val
```
elprans: Why this `assert`?

edb/edgeql/codegen.py
tailhook: I don't see a reason why not? This is just the way the user writes strings. Semantically the strings are equivalent. It's like making a difference between single and double-quotes.

edb/edgeql/parser/grammar/expressions.py
tailhook: Technically there is `token.text()` which keeps everything and `token.value()`, which extracts semantically valid value. But I think we can get rid of `token.text()` in the future to avoid allocations. I don't see the reason to keep it.

edb/edgeql/parser/grammar/expressions.py
tailhook: I've debugged some issue. Sorry.

edb/edgeql/parser/grammar/expressions.py
tailhook: Or rather `.value` and `.clean_value` in pythonic tokens

tailhook: Well, that was about the token level. The AST is "abstract". You can't reconstruct exact text anymore (you're skipping comments and whitespace). So there is no reason to pick some single thing that we can reconstruct: quotes. *(there is even a concept called CST that keeps everything)*

elprans: Yes, I didn't mean reconstructing from AST, only that the lexer has this capability:

> We can reconstruct the input stream by concatenating the tokens back.

tailhook:
Rebased and added `SUPERUSER` non-reserved keyword.

tailhook:
Oh, I can't merge it because it's done against `add_rust` branch. Will open new one.

Issue #969, opened 2019-12-14
Fetching fields on untyped backwards links should not work


tailhook:
The query like:
```
SELECT User.<owner { title };
```
Should be an error, while fetching `id` should work.

PR #972, opened 2019-12-16
Make unions produced by reverse link navigation universal

elprans:
Currently, a reverse link path expression (e.g `User.<owner`) produces a 
regular union type of all possible types that have an `owner` link to
`User`.  While this may seem fine at a glance, this behavior actually leads
to fragile queries due to the "action at a distance" effect. The meaning of
the expression might change unintentionally via the addition of new types
(not all of which are directly controlled by the user).

For this reason, make it so that reverse link navigation produces a 
univesral union, i.e the union of all possible types, which is
`std::Object`.  Then, to access further links or properties (other than the
ones defined on `std::Object`), an explicit type filter operator
(`[IS ..]`) must be used, so instead of `User.<owner.number` one must to 
write `User.<owner[IS Issue].number`.

While working on this I also realized that there is a subtle issue related
to how link properties are treated in the context of link unions.  First of
all, this commit actually makes paths like
`User.<owner[IS Issue]@note` work.  Notably, the implementation here
*does not* attempt to generate link unions based on inheritance resolution
of link source, so in `Foo.<link[IS Bar]` the link `Bar.link` must actually
exist.  Whether to change this so that a union of `link` links of all
subtypes of `Bar` is generated is left for further discussion and work.

Fixes: #969

### DIFF
--- a/edb/schema/utils.py
+++ b/edb/schema/utils.py
@@ -242,8 +242,33 @@ def get_class_nearest_common_ancestor(schema, classes):
         return None
 
 
-def minimize_class_set_by_least_generic(schema, classes):
-    """Minimize the given Object set by filtering out all superclasses."""
+def minimize_class_set_by_most_generic(
+    schema: s_schema.Schema,
+    classes: Iterable[so.InheritingObjectBase]
+) -> List[so.InheritingObjectBase]:
+    """Minimize the given set of objects by filtering out all subclasses."""
+
+    classes = list(classes)
+    mros = [set(p.get_ancestors(schema).objects(schema)) for p in classes]
+    count = len(classes)
+    smap = itertools.starmap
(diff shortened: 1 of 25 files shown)


edb/schema/utils.py
```diff
@@ -242,8 +242,33 @@ def get_class_nearest_common_ancestor(schema, classes):
...
+    count = len(classes)
+    smap = itertools.starmap
+
+    # Return only those entries that do not have other entries in their mro
+    result = [
+        scls for i, scls in enumerate(classes)
+        if not any(smap(set.__contains__,
+                        ((mros[i], classes[j])
+                         for j in range(count) if j != i)))
+    ]
```
1st1: 🤦‍♂️

Issue #741, opened 2019-09-30
Place restrictions on the use of DML statements 


elprans:
The behavior of DML CTEs in postgres dictates that they: a) must be a top-level CTE; b) always execute to completion.  These restrictions must be respected by the semantics of EdgeQL DML.  The top-level CTE requirement can only be satisfied if the use of correlated vars inside a DML statement is prohibited, i.e:

```SELECT (Foo, (INSERT Bar { foo := Foo }));```

The execute-to-completion requirement can be satisfied by prohibiting the use of DML inside a conditional construct, like `??` or `IF`.

See #110 for an earlier discussion and #684 for an example user issue.

1st1:
What about cases like `SELECT (Foo, (INSERT Bar { foo := Foo })) FILTER false`?

elprans:
`FILTER false` is not considered a conditional, since the body of the statement is logically evaluated first.

PR #975, opened 2019-12-17
Place restrictions on the use of DML statements

elprans:
The behavior of DML CTEs in Postgres dictates that they: a) must be a
top-level CTE; b) always execute to completion. These restrictions must
be respected by the semantics of EdgeQL DML. The top-level CTE
requirement can only be satisfied if the use of correlated vars inside a
DML statement is prohibited, i.e:

    SELECT (Foo, (INSERT Bar { foo := Foo }));

The execute-to-completion requirement can be satisfied by prohibiting
the use of DML inside a conditional construct, like `??` or `IF`.

Closes: #741

### DIFF
--- a/edb/edgeql/compiler/func.py
+++ b/edb/edgeql/compiler/func.py
@@ -200,6 +200,14 @@ def compile_FunctionCall(
     return setgen.ensure_set(fcall, typehint=rtype, path_id=path_id, ctx=ctx)
 
 
+#: A dictionary of conditional callables and the indices
+#: of the arguments that are evaluated conditionally.
+CONDITIONAL_OPS = {
+    'std::IF': {0, 2},
+    'std::??': {1},
+}
+
+
 def compile_operator(
         qlexpr: qlast.Base, op_name: str, qlargs: typing.List[qlast.Base], *,
         ctx: context.ContextLevel) -> irast.Set:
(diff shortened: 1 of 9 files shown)


edb/edgeql/compiler/func.py
```diff
@@ -200,6 +200,12 @@ def compile_FunctionCall(
     return setgen.ensure_set(fcall, typehint=rtype, path_id=path_id, ctx=ctx)
 
 
+CONDITIONAL_OPS = {
+    'std::IF': {0, 2},
+    'std::??': {1},
+}
```
1st1: Can you add a comment explaining what's in sets? It's clear from the code below that it's about arguments, but only because I see the full diff.

edb/edgeql/compiler/stmt.py
```diff
@@ -234,6 +240,13 @@ def compile_GroupQuery(
 @dispatch.compile.register(qlast.InsertQuery)
 def compile_InsertQuery(
         expr: qlast.InsertQuery, *, ctx: context.ContextLevel) -> irast.Set:
+
+    if ctx.in_conditional is not None:
+        raise errors.QueryError(
+            'INSERT statements cannot be used inside conditionals',
```
1st1: conditionals > conditional operators?

edb/edgeql/compiler/stmt.py
```diff
@@ -292,6 +305,13 @@ def compile_InsertQuery(
 @dispatch.compile.register(qlast.UpdateQuery)
 def compile_UpdateQuery(
         expr: qlast.UpdateQuery, *, ctx: context.ContextLevel) -> irast.Set:
+
+    if ctx.in_conditional is not None:
+        raise errors.QueryError(
+            'UPDATE statements cannot be used inside conditionals',
```
1st1: ditto; "conditionals" feels like jargon

edb/ir/scopetree.py
```diff
@@ -362,8 +402,20 @@ def attach_subtree(self, node: ScopeTreeNode) -> None:
...
+            visible, visible_finfo = self.find_visible_ex(path_id)
             if visible is not None:
+                if visible_finfo is not None and visible_finfo.factoring_fence:
+                    # This node is already present in the surrounding
+                    # scope and cannot be factored out, such as
+                    # a reference to a correlated set inside a DML
+                    # statement.
+                    raise InvalidScopeConfiguration(
+                        f'cannot reference correlated set '
+                        f'{path_id.pformat()!r} here',
```
1st1: not sure I can parse this error message. What does "here" mean in this context?

elprans: The "here" is pointed to by the parse context arrow.

PR #976, opened 2019-12-17
doc: Add zero cardinality and zero type id

tailhook:


docs/internals/protocol/messages.rst
```diff
@@ -305,6 +305,11 @@ Format:
     };
 
     enum Cardinality {
+
+        // Zero cardinality is used in statements which don't return
+        // any result, such as CREATE DATABASE
+        ZERO = 0x6e,
```
elprans: I think `ZERO` is ambiguous here, because it still implies you get a result, it's just an empty set, but of a certain type.  Whereas, `CREATE DATABASE` does not return *at all*, not even an empty set.  So, something like `NO_RESULT` would be clearer.

docs/internals/protocol/typedesc.rst
```diff
@@ -23,6 +23,11 @@ to binary data that EdegDB can decode and work with.
 A *decoder* is used to decode data from EdgeDB native format to
 data types native to the driver.
 
+There is one special type with *type id* of zero:
+``00000000-0000-0000-0000-000000000000``. When you describe this type, result
+contains zero *blocks*. It's used when statement returns no meaningful results,
```
1st1:
```suggestion
contains zero *blocks*. It's used when a statement returns no meaningful results,
```

docs/internals/protocol/typedesc.rst
```diff
@@ -23,6 +23,11 @@ to binary data that EdegDB can decode and work with.
...
+for example for ``CREATE DATABASE example`` statement.
```
1st1:
```suggestion
e.g. the ``CREATE DATABASE example`` statement.
```

docs/internals/protocol/typedesc.rst
```diff
@@ -23,6 +23,11 @@ to binary data that EdegDB can decode and work with.
 A *decoder* is used to decode data from EdgeDB native format to
 data types native to the driver.
 
+There is one special type with *type id* of zero:
+``00000000-0000-0000-0000-000000000000``. When you describe this type, result
```
1st1:
```suggestion
``00000000-0000-0000-0000-000000000000``. The describe result of this type
```

docs/internals/protocol/typedesc.rst
tailhook: Well, actually is there any difference?

elprans: I think stylistically we should avoid using `e.g.`, `i.e.`, `etc`.  "for example, the `CREATE DATABASE foo` statement" would read better

1st1:
Thanks, Paul!

PR #978, opened 2019-12-17
Add a cookbook section and a an article about links

tailhook:
The idea of the cookbook is to have some tutorial-like explanations, but not in chronological order. I.e. guides and examples for specific topics.

Please, review carefully. I'm not sure I fully understand underlying concepts.

Also, take a closer look (and maybe add more text?) to the last query or two. Because I find the implicit join between Person's in high-level select and `Person` in the `Peer` view kinda subtle and non-obviuos.

[Rendered links cookbook](https://github.com/edgedb/edgedb/blob/cookbook_links/docs/cookbook/links.rst)

### DIFF
--- a/docs/cookbook/index.rst
+++ b/docs/cookbook/index.rst
@@ -1,0 +1,12 @@
+========
+Cookbook
+========
+
+This sections shows common idioms, approaches and recipes for using EdgeDB.
+
+.. toctree::
+    :maxdepth: 1
+    :hidden:
+
+    aggregates
+    links
--- a/docs/cookbook/index.rst
+++ b/docs/cookbook/index.rst
@@ -2,7 +2,7 @@
 Cookbook
 ========
 
-This sections shows common idioms, approaches and recipes for using EdgeDB.
+These sections show common idioms, approaches, and recipes for using EdgeDB.
 
 .. toctree::
     :maxdepth: 1
--- a/docs/cookbook/links.rst
+++ b/docs/cookbook/links.rst
@@ -4,7 +4,7 @@
 Links
 =====
 
-Link items define a specific relationship between two object types.
+Links define a specific relationship between two object types.
 Links have a direction, but can be traversed in both ways forward and
 :ref:`backward <ref_cookbook_links_bw>`.
 
--- a/docs/cookbook/links.rst
+++ b/docs/cookbook/links.rst
@@ -215,7 +215,7 @@ Now the next step is quite simple, just add a filter:
     ......... } FILTER .first_name = 'Ryan';
     {Object {
         first_name: 'Ryan',
-        collegues: {
+        colleagues: {
             Object { first_name: 'Ana' },
             Object { first_name: 'Harrison' },
         }
--- a/docs/cookbook/links.rst
+++ b/docs/cookbook/links.rst
@@ -12,7 +12,7 @@ Links have a direction, but can be traversed in both ways forward and
 Movie Example
 =============
 
-To define a relationship we use ``link`` keyword:
+To define a relationship we use the ``link`` keyword:
 
 .. code-block:: sdl
 
--- a/docs/cookbook/links.rst
+++ b/docs/cookbook/links.rst
@@ -161,7 +161,7 @@ Or more complex example:
         }
     }
 
-Now you may notice that the Ryan (Gosling) is mentioned as a colleague of
+Now you may notice that Ryan Gosling is mentioned as a colleague of
 himself. To fix it we need few more concepts.
 
 First note that the request above is an equivalent of:
--- a/docs/cookbook/links.rst
+++ b/docs/cookbook/links.rst
@@ -120,10 +120,11 @@ traversal ``.<`` operator:
         }
     }}
 
-You might also note that we've added ``[IS Movie]`` type filter. This is how
-backward link traversal works: EdgeDB fetches every object in the entire
-database having the field ``actors`` which is a ``Person``. So we narrow down
-the set of objects to ``Movie`` and select a title from it.
+You might also note that we've added ``[IS Movie]``, which we call
+:eql:op:`type intersection <ISINTERSECT>` operator. This is how backward link
+traversal works: EdgeDB fetches every object in the entire database having the
+field ``actors`` which is a ``Person``. So we narrow down the set of objects to
+``Movie`` and select a title from it.
 
 All other tools work on backward link:
 
--- a/docs/cookbook/links.rst
+++ b/docs/cookbook/links.rst
@@ -91,14 +91,13 @@ In the movie example above, we have only shown a forward link traversal:
 
 Here is a another example of using forward link. This time we only return
 last names of the artists as plain string (not an object). In this case, we
-need to alias the field with ``:=``. This example uses **forward link** ``.>``
-operator to make the code clearer:
+need to alias the field with ``:=``:
 
 .. code-block:: edgeql-repl
 
     tutorial> SELECT Movie {
     .........     title,
-    .........     starring := Movie.>actors.last_name,
+    .........     starring := Movie.actors.last_name,
     ......... };
     {Object {
         title: 'Blade Runner 2049',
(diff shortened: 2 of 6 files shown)


docs/cookbook/links.rst
```diff
@@ -0,0 +1,239 @@
...
+    .........     movies := Person.<cast[IS Movie].title,
+    ......... } FILTER .first_name = 'Ryan';
+    {Object {
+        first_name: 'Ryan',
+        movies: {
+            'Blade Runner 2049',
+        }
+    }}
+
+You might also note that we've added ``[IS Movie]`` cast. This is how backward
```
elprans: > we've added ``[IS Movie]`` cast

We call this the "type filter" operator.  "Cast" implies no filtering, and a possibility of error being raised, whereas type filters always succeed (but may return an empty set),

docs/cookbook/links.rst
```diff
@@ -0,0 +1,239 @@
...
+    {Object {
+        title: 'Blade Runner 2049',
+        starring: {
+            'Ford',
+            'Gosling',
+            'de Armas',
+        }
+    }}
+
+To find all movies that a person is starred in we use a **backward link**
```
elprans: >  we use a **backward link**

I'm more in favor of using "forward link traversal" and "reverse link traversal".  Adding "traversal" is important, because technically there is no such thing as a "backward link" -- there is only one link, you just traverse it in both directions.

docs/cookbook/links.rst
```diff
@@ -0,0 +1,239 @@
...
+Now you may notice that the Ryan (Gosling) is mentioned as a colleague of
+himself. To fix it we need few more concepts.
+
+First note that the request above is an equivalent of:
+
+.. code-block:: edgeql-repl
+
+    tutorial> SELECT Person {
+    .........     first_name,
+    .........     collegues := (
```
elprans: *colleagues*

docs/glossary.rst
```diff
@@ -15,6 +15,16 @@ Glossary
       syntax used to define the structuring of schemas. Common DDL
       statements include ``CREATE``, ``DROP``, and ``ALTER``.
 
+   link
+      Link items define a specific relationship between two object types. Link
+      instances relate one object to one or more different objects.
+
+      Comparing to SQL link is a way to define relationship, foreign key,
```
elprans: Again, drawing parallels to SQL here isn't necessary, we can concentrate this on a separate page.

docs/cookbook/links.rst
```diff
@@ -0,0 +1,239 @@
...
+            required link director -> Person;
+            multi link cast -> Person;
+        }
+        type Person {
+            required property first_name -> str;
+            required property last_name -> str;
+        }
+    }
+
+In terms of SQL database we defined two foreign keys on the table containing
```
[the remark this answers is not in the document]
tailhook: The point is that it should show up in search, if you search for foreign key.

docs/cookbook/links.rst
```diff
@@ -0,0 +1,239 @@
...
+
+Note: we wrapped a backward link access by ``SELECT`` subquery.
+
+Still we can't filter out by ``Person.id != Person.id`` because EdgeDB can't
+distinguish them. To make that work we should factor the inner query out:
+
+.. code-block:: edgeql-repl
+
+    tutorial> WITH
+    .........     Peer := (SELECT Person.<cast[IS Movie].cast)
```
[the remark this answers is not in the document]
tailhook: Yes makes sense. I thought I've checked for that on real data, but now I kinda realized what was the mistake in the data :)

docs/cookbook/links.rst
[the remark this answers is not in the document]
elprans: The EdgeDB for SQL users page will show up.

docs/cookbook/links.rst
```diff
@@ -0,0 +1,221 @@
+.. _ref_cookbook_links:
+
+=====
+Links
+=====
+
+Link items define a specific relationship between two object types.
```
1st1:
```suggestion
Links define a specific relationship between two object types.
```

docs/cookbook/index.rst
```diff
@@ -0,0 +1,12 @@
+========
+Cookbook
+========
+
+This sections shows common idioms, approaches and recipes for using EdgeDB.
```
1st1:
```suggestion
These sections show common idioms, approaches, and recipes for using EdgeDB.
```

docs/cookbook/links.rst
```diff
@@ -0,0 +1,221 @@
...
+Links have a direction, but can be traversed in both ways forward and
+:ref:`backward <ref_cookbook_links_bw>`.
+
+
+Movie Example
+=============
+
+To define a relationship we use ``link`` keyword:
```
1st1:
```suggestion
To define a relationship we use the ``link`` keyword:
```

docs/cookbook/links.rst
```diff
@@ -0,0 +1,221 @@
...
+            title: 'Blade Runner 2049',
+            actors: {
+                Object { first_name: 'Harrison' },
+                Object { first_name: 'Ryan' },
+                Object { first_name: 'Ana' },
+            }
+        }
+    }
+
+Here is a another example of using forward link. This time we only return
```
1st1:
```suggestion
Here is another example of using a forward link. This time we only return
```

docs/cookbook/links.rst
```diff
@@ -0,0 +1,221 @@
...
+            first_name: 'Ryan',
+            colleagues: {
+                Object { first_name: 'Ana' },
+                Object { first_name: 'Harrison' },
+                Object { first_name: 'Ryan' },
+            }
+        }
+    }
+
+Now you may notice that the Ryan (Gosling) is mentioned as a colleague of
```
1st1:
```suggestion
Now you may notice that Ryan Gosling is mentioned as a colleague of
```

docs/cookbook/links.rst
```diff
@@ -0,0 +1,221 @@
...
+    .........         SELECT Person.<actors[IS Movie].actors {
+    .........             first_name,
+    .........         }
+    .........     ),
+    ......... } FILTER .first_name = 'Ryan';
+
+Note: we wrapped a backward link access by ``SELECT`` subquery.
+
+Still we can't filter out by ``Person != Person`` because EdgeDB can't
+distinguish them. To make that work we should give the inner query an alias:
```
1st1: @elprans Is this the way you'd phrase this? Specifically "``Person != Person`` because EdgeDB can't
distinguish them" doesn't sound correct to me.

elprans: This paragraph is actually outdated.  The query has been reformulated.

tailhook: ".. distinguish the object on the left and the object on the right." ? 

This is a bit ugly wording I agree, but other than that it's not outdated.

elprans: But it is. You no longer filter `Person != Person`, you do `Peer != Person`, and that is not ambiguous.

tailhook: The idea of this introduction is to show "why" you need to create an alias. Because you could add a FILTER right away if only you have a way to distinguish a Person from the inner query and a Person from the outer query.

elprans: You don't have to create an alias.  You're still thinking of a previous formulation that was incorrect.

This is sufficient:

```
    tutorial> SELECT Person {
    .........     first_name,
    .........     colleagues := (
    .........         SELECT Person.<actors[IS Movie].actors {
    .........             first_name,
    .........         } FILTER Person.<actors[IS Movie].actors != Person
    .........     ),
    ......... } FILTER .first_name = 'Ryan';
```

Aliasing simply makes the query more readable (which is, btw, one of big reasons why aliases exist and why they work the way they work).

tailhook: Well, I'm not thinking of a previous formulation. I was just unaware that it's possible :) (there are cases in SQL where the alias is required, perhaps, thats why)

tailhook:
Okay, I've rewritten it again. Including "type filters" -> "intersections" (that's why I've rebased the whole branch, so look at last few commits).

tailhook:
Ping! I understand that I'm inefficient with this. But until we have docs, I don't have any more efficient way to learn the language :)

PR #979, opened 2019-12-17
Rename `color_enum_t` to `Color` in documentation

tailhook:


tailhook:
We discussed that in slack. And it looks like `property color -> Color` is much nicer than `property color -> color_enum_t`. So the consensus was something like: let's make enums equal to object types rather than scalars, as far as I understand. For many scalars types like `int64` there is no reason of naming a variable `int64`.

vpetrovykh:
Let's formalize the thoughts on this in #983

1st1:
This is now part of https://github.com/edgedb/edgedb/pull/999, closing this one.

PR #984, opened 2019-12-17
edgeql: Drop `.>` path operator.

### LINKED ISSUE #982: Drop `.>` as syntax for forward links.
The `.>` is perfectly equivalent to `.` in paths in all cases. This makes it redundant. The original reasons for keeping it boil down to symmetry with `.<` and nothing else. In practice, it doesn't appear all that useful or helpful, but it can create a little user confusion (e.g. "_is_ there a difference between this and `.`?")

In the final balance of things it appears that this syntax should be dropped.

vpetrovykh:
Drop the redundant `.>` path operator in favor of using only `.`.

Fixes #982.

### DIFF
--- a/docs/edgeql/expressions/paths.rst
+++ b/docs/edgeql/expressions/paths.rst
@@ -34,7 +34,7 @@ The individual path components are:
 :eql:synopsis:`<step-direction>`
     It can be one of the following:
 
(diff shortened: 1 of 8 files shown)


elprans:
Any updates to the docs?

vpetrovykh:
> Any updates to the docs?

Yeah, thanks. I keep forgetting the docs. There was one line.

PR #992, opened 2019-12-19
Dump/restore omnibus

1st1:
~This is still WIP:~

- [x] Fix `test_cli` and `test_server_config` (maybe few others)
- [x] Add docs for new commands
- [x] Add a couple of basic functional tests

### DIFF
--- a/edb/edgeql/compiler/func.py
+++ b/edb/edgeql/compiler/func.py
@@ -21,8 +21,7 @@
 
 
 from __future__ import annotations
-
-import typing
+from typing import *  # NoQA
 
 from edb import errors
 
@@ -35,6 +34,7 @@ from edb.schema import functions as s_func
 from edb.schema import modules as s_mod
 from edb.schema import name as sn
 from edb.schema import operators as s_oper
+from edb.schema import scalars as s_scalars
 from edb.schema import types as s_types
 
 from edb.edgeql import ast as qlast
@@ -137,7 +137,7 @@ def compile_FunctionCall(
         # form of the function is actually used.
         env.schema_refs.add(func)
 
-    func_initial_value: typing.Optional[irast.Set]
+    func_initial_value: Optional[irast.Set]
 
     if matched_func_initial_value is not None:
         iv_ql = qlast.TypeCast(
@@ -155,7 +155,7 @@ def compile_FunctionCall(
     path_id = pathctx.get_expression_path_id(rtype, ctx=ctx)
 
     if rtype.is_tuple():
-        rtype = typing.cast(s_types.Tuple, rtype)
+        rtype = cast(s_types.Tuple, rtype)
         tuple_path_ids = []
         nested_path_ids = []
         for n, st in rtype.iter_subtypes(ctx.env.schema):
@@ -209,7 +209,7 @@ CONDITIONAL_OPS = {
 
 
 def compile_operator(
-        qlexpr: qlast.Base, op_name: str, qlargs: typing.List[qlast.Base], *,
+        qlexpr: qlast.Base, op_name: str, qlargs: List[qlast.Base], *,
         ctx: context.ContextLevel) -> irast.Set:
 
     env = ctx.env
@@ -383,6 +383,19 @@ def compile_operator(
                 hint = (f"The IF and ELSE result clauses must be of "
                         f"compatible types, while the condition clause must "
                         f"be 'std::bool'. {hint}")
+            elif op_name == '+':
+                str_t = cast(s_scalars.ScalarType,
+                             env.schema.get('std::str'))
+                bytes_t = cast(s_scalars.ScalarType,
+                               env.schema.get('std::bytes'))
+                if (
+                    (ltype.issubclass(env.schema, str_t) and
+                        rtype.issubclass(env.schema, str_t)) or
+                    (ltype.issubclass(env.schema, bytes_t) and
+                        rtype.issubclass(env.schema, bytes_t)) or
+                    (ltype.is_array() and rtype.is_array())
+                ):
+                    hint = 'Consider using the "++" operator for concatenation'
 
             raise errors.QueryError(
                 f'operator {str(op_name)!r} cannot be applied to '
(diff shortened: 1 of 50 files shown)


1st1:
@elprans this is ready for ya.

1st1:
I've added docs and a basic fuzz test `test_dump_fuzz_01`.

1st1:
The CI will of course be red until we merge https://github.com/edgedb/edgedb-python/pull/58

edb/edgeql/compiler/func.py
```diff
@@ -383,6 +383,24 @@ def compile_operator(
...
+                    (ltype.issubclass(env.schema, str_t) and
+                        rtype.issubclass(env.schema, str_t)) or
+                    (ltype.issubclass(env.schema, bytes_t) and
+                        rtype.issubclass(env.schema, bytes_t)) or
+                    (ltype.is_array() and rtype.is_array())
+                ):
+                    raise errors.QueryError(
+                        f"operator '+' cannot be applied to "
+                        f"operands of type {types}; "
+                        f"use the '++' operator instead",
```
elprans: I feel like `"use the '++' operator instead"` logically belongs to the `hint`

edb/lib/std/99-postlude.edgeql
```diff
@@ -0,0 +1,22 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+ALTER MODULE std {
```
elprans: Why does this have to be in a "postlude"?

edb/pgsql/delta.py
```diff
@@ -3587,7 +3587,7 @@ def apply(self, schema, context):
                 dbops.Update(
                     table=table, record=updaterec, condition=condition))
 
-        self.attach_alter_table(context)
+        # self.attach_alter_table(context)
```
elprans: ?

edb/schema/modules.py
```diff
@@ -29,7 +29,9 @@
 
 class Module(so.UnqualifiedObject, s_anno.AnnotationSubject,
              qlkind=qltypes.SchemaObjectClass.MODULE):
-    pass
+
+    builtin = so.SchemaField(
```
elprans: Setting of this field should be protected by `stdmode`

edb/edgeql/compiler/func.py
1st1: It does, but what if the hint isn't rendered? A hint is something that might not actually be the right thing in many cases; it's our attempt to suggest a solution. In this case we know the only right solution -- use "++". So I'd keep it right in the error message.

edb/lib/std/99-postlude.edgeql
1st1: Because `builtin := true` in prelude requires `std::bool` to exist.

edb/edgeql/compiler/func.py
1st1: https://stratechery.com/2019/aws-mongodb-and-the-economic-realities-of-open-source/

1st1: I think hints are good when we have an idea of how a user can fix their stuff, but not absolutely sure. Besides, hints might not be rendered in all environments. In this case we know that the only way to fix the error is using "++", it's simple. So I'd keep this right in the error message.

edb/server/mng_port/edgecon.pyx
```diff
@@ -1447,3 +1471,293 @@ cdef class EdgeConnection:
...
+            }
+
+            disable_trigger_q = ''
+            enable_trigger_q = ''
+            for table in tables:
+                disable_trigger_q += (
+                    f'ALTER TABLE {table} DISABLE TRIGGER ALL;'
+                )
+                enable_trigger_q += (
+                    f'ALTER TABLE {table} ENABLE TRIGGER ALL;'
```
elprans: This might conflict with a hack in constraint trigger logic.  It shouldn't cause an error, but would make things slower, as there would be an unnecessary trigger invocation.  I will see if it's possible to avoid using `DISABLE TRIGGER` on my end.

edb/pgsql/delta.py
1st1: Yeah, this is actually a question to you.  What does `self.attach_alter_table(context)`?  The method isn't defined on this class. Commenting it out makes `ALTER MODULE` function seemingly fine.

edb/server/mng_port/edgecon.pyx
1st1: >  I will see if it's possible to avoid using DISABLE TRIGGER on my end.

I can just drop this for now. Disabling/enabling triggers didn't have any effect on performance of our movies db dump/restore cycle.  Is it important to disable triggers?

edb/edgeql/compiler/func.py
elprans: I disagree. The purpose of a hint is exactly for this: point users to a way to fix the error.  The idea is that hint and/or details are always rendered when the error is user-facing, e.g in repl.

edb/lib/std/99-postlude.edgeql
elprans: OK.

edb/pgsql/delta.py
elprans: Right. This was misplaced.  It is only for objects that are defined in terms of tables in Postgres.  Remove the line.

edb/server/mng_port/edgecon.pyx
elprans: Triggers might misfire depending on the order in which we restore stuff, so it's best to disable them for restore.

edb/schema/modules.py
1st1: `allow_ddl_set=False` for this field.

edb/edgeql/compiler/func.py
1st1: You seem to disagree stronger so I'll update the code to use hints. :)

Issue #983, opened 2019-12-17
Naming conventions for scalars and object types.


vpetrovykh:
Currently there's a convention in EdgeDB to name scalar types using `lower_case` with an added `_t` at the end. Object types are named using `CamelCase`.

The `_t` is meant to easily distinguish scalar type names and other lower-case names like functions and links. Abstract links are especially important here since they exist in the same namespace as the scalar types and can't use the same name. Although the problem of clashing names is a real one and needs to be solved, adding `_t` is somewhat awkward as a solution. In the general case of all scalar types it may be that "give better names" is the only generic reasonable advice.

However, there's a subset of scalar types that may have a different solution - enums. We could have a convention that the enum type names should be `ALLCAPS` to kind of remind what these types are. Using `ALLCAPS` for special constants is common practice in some programming languages and enums are conceptually similar to specialized constants.

Object type names should probably keep using `CamelCase`.

Whatever convention we agree on, should be reflected in our own built-in libraries. Incidentally, we still have some remains of "old-style enums" that emulated enum behavior by using a `one_of` constraint. The good news is that we only have things that are functionally enums that follow the `_t` scalar naming, other built-in scalars just have plain readable names.

tailhook:
What are examples of the languages which use ALL_CAPS enums?

Both Python and Rust use CamelCase for all custom types: objects, scalars and enums. And that is fine to me.Python has ALL_CAPS for the variants of the enum, which kinda makes sense for python as they are essentially constants. But not rust. Also we, don't discuss the variants, because they

vpetrovykh:
I mentioned that `ALLCAPS` are used for special constants, like `STD_LIB` and `STD_MODULES` in our own codebase. 

More importantly we need some way to come up with names for a common case of scalars - enums, which doesn't make them visually similar to object types. In Python, for example, there's no meaningful difference between a `str` and some user-created type because they are all subtypes of  `object` and behaviors can be implemented via magic methods. This is not the case for EdgeDB types, there are fundamental differences in behaviors between scalars and objects, which in the past we tried to unify, but encountered a number of problems (including gross inefficiency and really messy semantics).

tailhook:
I don't understand what is such an important difference? In python, there is also a lot of difference between immutable and mutable types, and between of subclasses of the built-in scalar and normal classes, and namedtuples and data classes. But we can live with that.

elprans:
[...]

PR #999, opened 2019-12-20
Rename enums and datetime types to follow the CamelCase convention

1st1:
In our recent @edgedb/engineering discussions we decided that we want to stop using `_t` suffixes for enums in the `schema::` module.  `_t` (short for "type") doesn't really tell the user much about the type, it is redundant by all means.  However, renaming `schema::cardinality_t` into `schema::cardinality` means that it's no longer possible to create an abstract property `abstract property cardinality` in the "schema" module.

There are two ways of resolving this potential naming issue:

1. Let our users be creative. After all EdgeQL does not enforce any naming convention or even language for identifiers.
2. Adopt the CamelCase convention for scalar types.

Regarding the (2) option: historically in EdgeDB the convention was that object types use CamelCase naming, and scalar types use snake_case. We can adopt CamelCase for either all scalar types in our builtin modules or for some.

This PR adopts CamelCase convention for all builtin enum types, as well as for `std::DateTime`, `std::Duration`, `cal::LocalDateTime`, `cal::LocalDate`, and `cal::LocalTime`.

Issue #983.
This PR obsoletes #979.


1st1:
Update: we decided that we will only use CamelCase for our enums and object types. Scalar types in the stdlib will continue using snake_case. For user-defined types we'll recommend using the CamelCase convention.

PR #1001, opened 2019-12-20
Cache type_to_typeref

ambv:
Before:
```
compile_edgeql_script: Mean +- std dev: 2.60 ms +- 0.07 ms
compile_edgeql_script_webapp_configure_system: Mean +- std dev: 44.0 ms +- 3.6 ms
compile_edgeql_script_webapp_get_movie: Mean +- std dev: 289 ms +- 55 ms
compile_edgeql_script_webapp_get_person: Mean +- std dev: 316 ms +- 73 ms
compile_edgeql_script_webapp_get_user: Mean +- std dev: 150 ms +- 4 ms
compile_edgeql_script_webapp_insert_movie: Mean +- std dev: 155 ms +- 2 ms
compile_edgeql_script_webapp_insert_person: Mean +- std dev: 59.6 ms +- 2.2 ms
compile_edgeql_script_webapp_insert_review: Mean +- std dev: 91.4 ms +- 1.7 ms
```

After:
```
compile_edgeql_script: Mean +- std dev: 2.44 ms +- 0.06 ms (6.2% faster)
compile_edgeql_script_webapp_configure_system: Mean +- std dev: 35.0 ms +- 0.9 ms (21% faster)
compile_edgeql_script_webapp_get_movie: Mean +- std dev: 256 ms +- 63 ms (11.5% faster)
compile_edgeql_script_webapp_get_person: Mean +- std dev: 325 ms +- 83 ms (2.8% SLOWER)
compile_edgeql_script_webapp_get_user: Mean +- std dev: 142 ms +- 3 ms (5.4% faster)
compile_edgeql_script_webapp_insert_movie: Mean +- std dev: 140 ms +- 3 ms (9.7% faster)
compile_edgeql_script_webapp_insert_person: Mean +- std dev: 54.6 ms +- 1.5 ms (8.4% faster)
compile_edgeql_script_webapp_insert_review: Mean +- std dev: 84.6 ms +- 2.0 ms (8.5% faster)
```

For big queries the jitter is higher due to GC activity as the cache needs to be cleaned out.

Other observations:
- only caching in the Environment didn't allow me to cache all places (see two marked spots) and got me thinking that it would be better to have a per-Schema "persistent" in-memory cache for this that would survive between compilation rounds; of course benchmarks would have to clean it up but for Real World use that would make the caches' hit rate much higher;
- I looked into why `get_person()` is actually slower now but have no answers yet;
- for benchmarks we want to disable GC when the function is being measured, given that those functions are not long-lived this should help with the jitter a lot;
- benchmarking with Wi-Fi off works better for me in terms of minimizing jitter;
- there are two failing tests due to the cache now but I'm not entirely sure if those are bugs, please look at them (test_edgeql_ir_scope_tree_19; test_edgeql_ir_scope_tree_24).

### DIFF
--- a/edb/edgeql/compiler/__init__.py
+++ b/edb/edgeql/compiler/__init__.py
@@ -596,7 +596,7 @@ def get_param_anchors_for_callable(
     if inlined_defaults:
         anchors['__defaults_mask__'] = irast.Parameter(
             name='__defaults_mask__',
-            typeref=irtyputils.type_to_typeref(
+            typeref=irtyputils.type_to_typeref(  # note: no cache
                 schema,
                 cast(s_scalars.ScalarType, schema.get('std::bytes')),
             ),
--- a/edb/edgeql/compiler/casts.py
+++ b/edb/edgeql/compiler/casts.py
@@ -71,7 +71,9 @@ def compile_cast(
 
     elif irutils.is_untyped_empty_array_expr(ir_expr):
         # Ditto for empty arrays.
-        new_typeref = irtyputils.type_to_typeref(ctx.env.schema, new_stype)
+        new_typeref = irtyputils.type_to_typeref(
+            ctx.env.schema, new_stype, cache=ctx.env.type_ref_cache
+        )
         return setgen.ensure_set(
             irast.Array(elements=[], typeref=new_typeref), ctx=ctx)
 
@@ -161,8 +163,12 @@ def _cast_to_ir(
         new_stype: s_types.Type, *,
         ctx: context.ContextLevel) -> irast.Set:
 
-    orig_typeref = irtyputils.type_to_typeref(ctx.env.schema, orig_stype)
-    new_typeref = irtyputils.type_to_typeref(ctx.env.schema, new_stype)
+    orig_typeref = irtyputils.type_to_typeref(
+        ctx.env.schema, orig_stype, cache=ctx.env.type_ref_cache
+    )
+    new_typeref = irtyputils.type_to_typeref(
+        ctx.env.schema, new_stype, cache=ctx.env.type_ref_cache,
+    )
     cast_name = cast.get_name(ctx.env.schema)
     cast_ir = irast.TypeCast(
         expr=ir_set,
@@ -185,8 +191,12 @@ def _inheritance_cast_to_ir(
         new_stype: s_types.Type, *,
         ctx: context.ContextLevel) -> irast.Set:
 
-    orig_typeref = irtyputils.type_to_typeref(ctx.env.schema, orig_stype)
-    new_typeref = irtyputils.type_to_typeref(ctx.env.schema, new_stype)
+    orig_typeref = irtyputils.type_to_typeref(
+        ctx.env.schema, orig_stype, cache=ctx.env.type_ref_cache,
+    )
+    new_typeref = irtyputils.type_to_typeref(
+        ctx.env.schema, new_stype, cache=ctx.env.type_ref_cache,
+    )
     cast_ir = irast.TypeCast(
         expr=ir_set,
         from_type=orig_typeref,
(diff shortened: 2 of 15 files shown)


ambv:
I did more measuring and sleuthing on this.

Stats from all unit tests:
- 799k calls
- 83.47% hit rate
- 0.000096% hit rate for types with custom typenames
- 0.000376% name hint mismatch (see below)
- 2.7% not cached at all (recursive calls for tuple subtypes)

### Conclusion 1
Due to the extremely low hit rate on types with custom typenames, I removed the custom typename from the cache key, making it smaller and thus faster for the common case.

### Conclusion 2
The two tests that are failing are failing because:
- while the type in question is properly recognized and cached;
- the schema changed during compilation (saved a bunch of derived types);
- and so the *name hint* is different.

Since the name hint is only used in annotations, and both name hints properly describe the type anyway, it is an acceptable compromise to allow for the name hints to not be derived types. I added an additional check for now to ensure that the name hints are correct every time. My intuition was that it's not worth it because the name hint mismatch rate is extremely low while the check cost would be paid on all cache hits. But it looks like it doesn't really make the cache perform much worse.

### More benchmarks
#### Before
```
compile_edgeql_script_simple: Mean +- std dev: 2.60 ms +- 0.07 ms
compile_edgeql_script_webapp_configure_system: Mean +- std dev: 44.0 ms +- 3.6 ms
compile_edgeql_script_webapp_get_movie: Mean +- std dev: 289 ms +- 55 ms
compile_edgeql_script_webapp_get_person: Mean +- std dev: 316 ms +- 73 ms
compile_edgeql_script_webapp_get_user: Mean +- std dev: 150 ms +- 4 ms
compile_edgeql_script_webapp_insert_movie: Mean +- std dev: 155 ms +- 2 ms
compile_edgeql_script_webapp_insert_person: Mean +- std dev: 59.6 ms +- 2.2 ms
compile_edgeql_script_webapp_insert_review: Mean +- std dev: 91.4 ms +- 1.7 ms
```

#### After (without name hint correction)
```
compile_edgeql_script_simple: Mean +- std dev: 2.43 ms +- 0.07 ms (6.54% faster)
compile_edgeql_script_webapp_configure_system: Mean +- std dev: 34.9 ms +- 1.0 ms (20.68% faster)
compile_edgeql_script_webapp_get_movie: Mean +- std dev: 257 ms +- 68 ms (11.07% faster)
compile_edgeql_script_webapp_get_person: Mean +- std dev: 315 ms +- 79 ms (0.32% faster)
compile_edgeql_script_webapp_get_user: Mean +- std dev: 142 ms +- 4 ms (5.33% faster)
compile_edgeql_script_webapp_insert_movie: Mean +- std dev: 143 ms +- 3 ms (7.74% faster)
compile_edgeql_script_webapp_insert_person: Mean +- std dev: 54.3 ms +- 1.1 ms (8.89% faster)
compile_edgeql_script_webapp_insert_review: Mean +- std dev: 83.8 ms +- 1.7 ms (8.32% faster)
```

#### After (with name hint correction)
```
compile_edgeql_script_simple: Mean +- std dev: 2.52 ms +- 0.11 ms (3.08% faster)
compile_edgeql_script_webapp_configure_system: Mean +- std dev: 34.9 ms +- 0.7 ms (20.68% faster)
compile_edgeql_script_webapp_get_movie: Mean +- std dev: 250 ms +- 58 ms (13.49% faster)
compile_edgeql_script_webapp_get_person: Mean +- std dev: 323 ms +- 84 ms (2.22% SLOWER)
compile_edgeql_script_webapp_get_user: Mean +- std dev: 143 ms +- 5 ms (4.67% faster)
compile_edgeql_script_webapp_insert_movie: Mean +- std dev: 143 ms +- 3 ms (7.74% faster)
compile_edgeql_script_webapp_insert_person: Mean +- std dev: 55.2 ms +- 2.1 ms (7.38% faster)
compile_edgeql_script_webapp_insert_review: Mean +- std dev: 85.3 ms +- 2.0 ms (6.67% faster)
```

ambv:
This is ready.

PR #1004, opened 2019-12-20
Put caching in all invocations of ptrref_from_ptrcls

ambv:


ambv:
We're leaving this out after conversation with @eprans.

PR #1015, opened 2019-12-24
Fix two ancestry propagation bugs found by @ambv in #1014

elprans:
The recomputation of ancestry on `issubclass()` calls is currently
masking two bugs in the correct propagation of the `ancestors` field:

  * `ReferencedObject.derive_ref()` results in `ancestors` not matching
    the actual ancestry if the object being derived already exists
    and the base does not match the actual base.

  * When comparing schemas, the `ancestors` field does not participate
    in the comparison function, which means that the result delta will
    omit objects that had only their ancestors changed.

### DIFF
--- a/edb/pgsql/common.py
+++ b/edb/pgsql/common.py
@@ -202,9 +202,22 @@ def get_objtype_backend_name(id, module_id, *, catenate=True, aspect=None):
     return convert_name(name, suffix=suffix, catenate=catenate)
 
 
-def get_pointer_backend_name(id, module_id, *, catenate=False):
+def get_pointer_backend_name(id, module_id, *, catenate=False, aspect=None):
+    if aspect is None:
+        aspect = 'table'
+
+    if aspect not in ('table', 'index'):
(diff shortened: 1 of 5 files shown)


edb/pgsql/common.py
```diff
@@ -202,9 +202,22 @@ def get_objtype_backend_name(id, module_id, *, catenate=True, aspect=None):
...
+        aspect = 'table'
+
+    if aspect not in ('table', 'index'):
+        raise ValueError(
+            f'unexpected aspect for pointer backend name: {aspect!r}')
+
     name = s_name.Name(module=str(module_id), name=str(id))
-    return convert_name(name, suffix='', catenate=catenate)
+
+    if aspect != 'table':
```
1st1:
```suggestion
    if aspect == 'index':
```

elprans: No, I like the current formulation better.  Only the `table` aspect does not require a suffix, but reversing the condition would basically require amending it every time a new aspect is added.

PR #1025, opened 2019-12-27
Fixes mock scram auth with new edgedb-python

tailhook:
The edgedb-python that requires this is:
https://github.com/edgedb/edgedb-python/pull/60

### DIFF
--- a/edb/server/bootstrap.py
+++ b/edb/server/bootstrap.py
@@ -506,7 +506,7 @@ async def _compile_sys_queries(schema, cluster):
 
 async def _populate_misc_instance_data(cluster):
 
-    mock_auth_nonce = scram.B64(scram.generate_nonce())
+    mock_auth_nonce = scram.generate_nonce()
     ver = buildmeta.get_version()
     json_instance_data = {
         'version': {


edb/server/bootstrap.py
```diff
@@ -506,7 +506,7 @@ def _process_delta(delta, schema):
 
 async def _populate_misc_instance_data(cluster):
 
-    mock_auth_nonce = scram.B64(scram.generate_nonce())
+    mock_auth_nonce = scram.generate_nonce()
```
elprans: `json.dumps()` doesn't like `bytes`, so the correct fix is to still encode this here, and decode in edgecon.pyx:490.

tailhook: It isn't encoded anymore. It returns strings in after https://github.com/edgedb/edgedb-python/pull/60 because in the spec scram nonce is a string of ASCII-printable characters rather than an array of bytes (In the current implementation it means that base64 in moved into the function)

elprans: Ah, so it fails because we need to bump edgedb-python?  Got it.

Neustradamus:
Good job!

Linked to:

PR #1026, opened 2019-12-29
Initial support for using a remote Postgres cluster as a backend

elprans:
This adds the new `--postgres-dsn` option to `edgedb-server`.  When 
specified instead of `-D`, EdgeDB will connect to the specified cluster 
address instead of using a local data directory.

Supporting this requires a bunch of changes around how roles and 
connections are treated.  Since most DBaaS restrict direct superuser 
access, and instead provide a regular role with specific privileges, the 
bootstrap process became somewhat more involved.

### DIFF
--- a/docs/admin/setup.rst
+++ b/docs/admin/setup.rst
@@ -68,3 +68,46 @@ method with the ``trust`` method:
 .. code-block:: bash
 
     $ edgedb --admin -h </data/dir> configure insert auth --method=trust
+
+
+Using remote PostgreSQL cluster as a backend
+============================================
+
+By default, EdgeDB creates and manages a local PostgreSQL database instance,
+however it is also possible to use a remote PostgreSQL instance, as long as it
+is version 12 or later.  Superuser access to the cluster is *required*.
+
+To setup EdgeDB using a remote PostgreSQL instance, instead of ``-D``,
+specify the ``-P`` (or ``--postgres-dsn``) option when starting the EdgeDB
+server:
+
+.. code-block:: bash
+
+    $ edgedb-server \
+        -P postgres://user:password@host:port/database?opt=val
+
+The format of the connection string generally follows that of `libpq`_,
+including support for specifying the connection parameters via
+`environment variables <postgres envvars>`_ and reading passwords from
+`the password file <postgres passfile>`_.  Unlike libpq, EdgeDB will treat
+unrecognized options as `PostgreSQL settings <postgres settings>`_ to be used
+for the connection.  Multiple hosts in the connection string are unsupported.
+
+.. note::
+
+    PostgreSQL DBaaS providers normally do not allow direct superuser access
+    to the database instance, which might prevent EdgeDB from working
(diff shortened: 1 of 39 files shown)


edb/pgsql/metaschema.py
```diff
@@ -2399,9 +2399,9 @@ def _generate_role_views(schema):
             a.rolname                                   AS name,
             a.rolsuper                                  AS is_superuser,
             a.rolcanlogin                               AS allow_login,
-            a.rolpassword                               AS password
+            (d.description)->>'ph'                      AS password
```
tailhook: Do you really think that compressing/obfuscating this field is justified?

edb/server/pgcon/pgcon.pyx
```diff
@@ -949,6 +981,22 @@ cdef class PGProto:
...
 
+    cdef make_auth_password_md5_message(self, bytes salt):
+        cdef WriteBuffer msg
+
+        msg = WriteBuffer.new_message(b'p')
+
+        user = self.pgaddr.get('user') or ''
+        password = self.pgaddr.get('password') or ''
+
+        # 'md5' + md5(md5(password + username) + salt))
```
tailhook: Is this encoding of a password is still considered safe? I would expect PBKDF with thousands of iterations here too. (perhaps, argon2_ffi?)

docs/admin/setup.rst
```diff
@@ -68,3 +68,31 @@ method with the ``trust`` method:
 .. code-block:: bash
 
     $ edgedb --admin -h </data/dir> configure insert auth --method=trust
+
+
+Using remote PostgreSQL cluster as a backend
```
tailhook: Please, document environment and pgpass file

edb/server/pgcon/pgcon.pyx
elprans: This is what RDS Postgres advertises.  Ideally, they'd need to configure it to use SCRAM, but they didn't.

edb/pgsql/metaschema.py
elprans: Not really.  Security through obscurity does not work.  We have to rely on a) That access to `pg_shdescription` is sufficiently restricted; and b) That SCRAM makes brute-force attacks infeasible.

tailhook: Yes. I mean, rename it to `password_hash` or something :)

edb/server/pgcon/pgcon.pyx
tailhook: Ah, sorry, I've misread this as our method for storing passwords.

docs/admin/setup.rst
elprans: Done.

edb/pgsql/metaschema.py
elprans: Ah, OK. Done.

elprans: Ah, OK. Done.

PR #1031, opened 2019-12-31
Use 32-bit length prefixed strings in descriptors

tailhook:
This requires new edgedb-python


1st1:
Will merge this tomorrow after all other PRs and the dump/restore PR I'm working on.

PR #1032, opened 2019-12-31
Introduce memory profiling

ambv:
Generates memory usage SVG graphs.

Example:
[Screenshot 2019-12-31 at 19 24 16](http://langa.pl/random/graphs/get_movie_memory-2019-12-31.svg)

Explanation:
- the top of the graph is aggregate memory pressure generated by calling a given function;
- if you hover over a block, you see how much memory as well as the actual line code;
- the actual memory allocations happen in the orangish leaf nodes on the bottom, so if you're looking to optimize, look for the widest bars **at the bottom** of the graph;
- the greenish blocks have more than one child;
- the gray blocks only have a single child and don't perform significant allocations so they are boring; I only left them here for traceback completion.

This has edbperf integration in a separate PR.

### DIFF
--- a/edb/tools/profiling/profiler.py
+++ b/edb/tools/profiling/profiler.py
@@ -24,11 +24,14 @@ See README.md in this package for more details.
 from __future__ import annotations
 from typing import *  # NoQA
 
+import ast
 import atexit
 import builtins
 import cProfile
 import dataclasses
 import functools
+import hashlib
+import linecache
 import os
 import pathlib
 import pickle
@@ -59,6 +62,7 @@ if TYPE_CHECKING:
     LineNo = int
     FunctionName = str
     FunctionID = Tuple[ModulePath, LineNo, FunctionName]
+    LineID = Tuple[ModulePath, LineNo]
     # cc, nc, tt, ct, callers
     PrimitiveCallCount = int  # without recursion
     CallCount = int  # with recursion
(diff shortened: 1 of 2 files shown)


ambv:
By the way, @elprans, look how in the example graph most allocations are `Schema._update_refs_to`:
Screenshot 2019-12-31 at 19 55 42

edb/tools/profiling/profiler.py
```diff
@@ -268,6 +272,39 @@ def accumulate_singledispatch_traces(self) -> Dict[FunctionID, CallCounts]:
...
 
 
+def profile_memory(func: Callable[[], Any]) -> MemoryFrame:
+    """Profile memory and return a tree of statistics.
+
+    Feed those to `render_memory_svg()` to write an SVG.
+    """
+    import tracemalloc
+
+    tracemalloc.start(1024)
```
elprans: What process did you use to arrive at this particular number of frames?

ambv: 1. I wanted a number that was the next perfect square after a safe number that covered all recursion in our benchmarks.
2. 1000 is the recursion limit.

It's an arbitrary number but I wanted the flame graph to be able to show all allocations.

elprans: :+1:

ambv:
`_update_refs_to` is one of the most commonly called functions in our codebase at 2.73% of calls:
Screenshot 2019-12-31 at 20 00 00

It would make sense to optimize it.

1st1:
So how can I use this new tool? Can you update the readme?

PR #1034, opened 2019-12-31
Use immutables.Map instead of a frozenset to store schema crossrefs

elprans:
The memory profiler added in #1032 has quickly pointed to
`Schema._update_refs_to()` as being the largest memory consumer during
query compilation.  This is due to the use of `frozenset` to store the
ids of referring objects.  Replacing it with `immutables.Map` results in
a significant reduction of memory use.

BEFORE:

    compile_edgeql_script_webapp_get_movie.memory:
        2900.13KiB / 6610 blocks

AFTER:

    compile_edgeql_script_webapp_get_movie.memory:
        855.99KiB / 7227 blocks (*70.48% less*)

### DIFF
--- a/edb/schema/schema.py
+++ b/edb/schema/schema.py
@@ -283,8 +283,6 @@ class Schema(s_abc.Schema):
         if not objfields:
             return self._refs_to
 
-        id_set = frozenset((scls.id,))
-
         with self._refs_to.mutate() as mm:
             for field in objfields:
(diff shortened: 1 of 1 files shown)


1st1:
Was there any speed improvement?

elprans:
Not conclusive on the benchmarks, but the test suite takes 10s less for me with this.

Issue #930, opened 2019-12-04
Add details to InternalServerError suggesting to file an issue


elprans:
`InternalServerError` is always a bug, even in cases where we simply don't translate the underlying Postgres error correctly.  To make this clear, the details (or hint) field should have a suggestion to open an issue on Github.  In the future we may extend this to automatically pre-populate the issue with relevant information as well.

PR #1035, opened 2020-01-02
Suggest to open a bug report on InternalServerError

elprans:
`InternalServerError` is always a bug, so put a hint to open an issue on
Github.

Fixes: #930.

Add Github issue templates

This adds issue templates for bugs and feature requests and a link to 
Spectrum for questions.

### DIFF
--- a/edb/server/mng_port/edgecon.pyx
+++ b/edb/server/mng_port/edgecon.pyx
@@ -1298,16 +1298,26 @@ cdef class EdgeConnection:
                 exc = RuntimeError(
                     'unhandled error while calling interpret_backend_error()')
 
-        fields = None
+        fields = {}
         fields_len = 0
         if (isinstance(exc, errors.EdgeDBError) and
                 type(exc) is not errors.EdgeDBError):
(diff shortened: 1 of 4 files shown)


edb/server/mng_port/edgecon.pyx
```diff
@@ -1298,16 +1298,26 @@ cdef class EdgeConnection:
...
             fields_len = <int16_t><uint16_t>(len(fields))
 
+        internal_error_code = errors.InternalServerError.get_code()
         if not exc_code:
-            exc_code = errors.InternalServerError.get_code()
+            exc_code = internal_error_code
+
+        if (exc_code == internal_error_code
+                and not fields.get(base_errors.FIELD_HINT)):
+            fields[base_errors.FIELD_HINT] = (
```
tailhook: Hm, modification of a dict aliased from `exc._attrs` at 1306. How likely this can be a problem?

elprans: Yeah, it would be better to make a copy.

Issue #1030, opened 2019-12-30
Use -H for hostname in CLI; use -h for help


1st1:
I think using `-h` to display help is a better UX than what we have currently. I propose to use capital `-H` for host names and a capital `-P` for ports (for symmetry with `-H`).

tailhook:
Fixed by #1039

PR #1039, opened 2020-01-06
Command-line fixes

### LINKED ISSUE #1029: Subcommands in edgedb-repl
# Backround

Currently, they look like:
```
edgedb create role <rolename>
edgedb create datbase <dbname>
```
And help looks like:
```
> edgedb --help
Usage: edgedb [OPTIONS] COMMAND [ARGS]...

Options:
  -h, --host TEXT
  -p, --port INTEGER
  -u, --user TEXT
  -d, --database TEXT
  --admin
  --password / --no-password
  --password-from-stdin
  -?, --help                  Show this message and exit.

Commands:
  alter
  configure
  create
  drop
> edgedb create --help
Usage: edgedb create [OPTIONS] COMMAND [ARGS]...

Options:
  -h, --host TEXT
  -p, --port INTEGER
  -u, --user TEXT
  -d, --database TEXT
  --admin
  --password / --no-password
  --password-from-stdin
  -?, --help                  Show this message and exit.

Commands:
  role

```

# Proposal

What I would like to have:
```
edgedb create-role <rolename>
edgedb create-datbase <dbname>
```
And help like this:
```
> edgedb --help
Usage: edgedb [OPTIONS] COMMAND [ARGS]...

Options:
  -h, --host TEXT
  -p, --port INTEGER
  -u, --user TEXT
  -d, --database TEXT
  --admin
  --password / --no-password
  --password-from-stdin
  -?, --help                  Show this message and exit.

Commands:
  alter-role
  configure-role
  create-role
  create-database
  drop-role
```

tailhook:
Fixes #1029, #1030

I've left `configure` as is, but if we want to make it a form similar to other (verb-noun) they should have been `insert-config`/`reset-config`/`set-config`, which I'm not sure are good. Anyway, current `configure` is good for discoverability.

### DIFF
--- a/docs/cli/edgedb.rst
+++ b/docs/cli/edgedb.rst
@@ -23,17 +23,17 @@ queries and seeing results interactively.
 Options
 =======
 
-:cli:synopsis:`-?, --help`
+:cli:synopsis:`-h, --help`
     Show help about the command and exit.
 
-:cli:synopsis:`-h <hostname>, --host=<hostname>`
+:cli:synopsis:`-H <hostname>, --host=<hostname>`
     Specifies the host name of the machine on which the server is running.
     If :cli:synopsis:<hostname> begins with a slash (``/``), it is used
     as the directory where the command looks for the server Unix-domain
     socket.  Defaults to the value of the ``EDGEDB_HOST`` environment
     variable.
 
-:cli:synopsis:`-p <port>, --port=<port>`
+:cli:synopsis:`-P <port>, --port=<port>`
     Specifies the TCP port or the local Unix-domain socket file extension
     on which the server is listening for connections.  Defaults to the value
     of the ``EDGEDB_PORT`` environment variable or, if not set, to ``5656``.
--- a/docs/admin/installation.rst
+++ b/docs/admin/installation.rst
@@ -131,4 +131,4 @@ to the server container:
 
     docker run --link=edgedb-server --rm -it \
         edgedb/edgedb:latest \
-        edgedb -u edgedb -h edgedb-server
+        edgedb -u edgedb -H edgedb-server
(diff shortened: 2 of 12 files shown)


tailhook:
Right, sorry. Also, i've just noticed:
```
  -P, --postgres-dsn TEXT         DSN of a remote Postgres cluster, if using
  -I, --bind-address TEXT         IP address to listen on
  -P, --port INTEGER              port to listen on
```
(not sure why click doesn't check for that!)
Which `-P` is right here? I would probably drop short alias for `--postgres-dsn` at all, I don't see a reason for it to be short (so we keep symmetry in client and server, an also both `-I` and `-P` are upper case).

tailhook:
Updated docs and also dropped `-P` alias for `--postgres-dsn`. Let me know if you think it's needed.

1st1:
> I would probably drop short alias for --postgres-dsn

+1. We should only use one letter aliases for commonly used parameters anyways.

1st1:
> (not sure why click doesn't check for that!)

That's quite scary actually.

docs/cli/edgedb.rst
```diff
@@ -23,17 +23,17 @@ queries and seeing results interactively.
 Options
 =======
 
-:cli:synopsis:`-?, --help`
+:cli:synopsis:`-h, --help`
```
elprans: I would keep `-?` as well, its a common convention on Windows.

tailhook: On Unixes it must be quoted because it's a shell pattern. Isn't it on new windows terminal is the same too? (It looks' like `?` is a [globbing pattern on powershell](https://www.leeholmes.com/blog/2015/01/23/the-wonderful-world-of-filtering-and-globbing/)). And rust implementation will need more work for having another alias for help command (I know that's unfortunate). So I had no motivation to do that. Also on windows it's likely `/?`, no?

Do you still think it worth it?

elprans: OK, let's drop it.

PR #1040, opened 2020-01-06
Fix datetime_get functions' units

### LINKED ISSUE #958: Validate names of datetime units to only support plurals
This includes:

* `datetime_get`
* `datetime_trunc`

`duration_get` is already works. Maybe other?

tailhook:
Fixes #958

Except what discussed in #958 I've also made the following fixes:
1. `epoch` -> `midnightseconds` for `cal::time_get` (epoch doesn't make sense there)
2. Remove `epoch` and other useless units for `cal::date_get`. We can introduce `epochdays` if needed but I don't think that seconds since epoch makes sense for the type.

### DIFF
--- a/docs/edgeql/funcops/datetime.rst
+++ b/docs/edgeql/funcops/datetime.rst
@@ -267,7 +267,7 @@ Date and Time
 
     Extract a specific element of input date by name.
 
-    The :eql:type:`local_date` scalar has the following elements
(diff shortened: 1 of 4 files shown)


1st1:
Don't forget to squash :)

tailhook:
> Don't forget to squash :)

It's easier to remember when it's default :)

Issue #846, opened 2019-11-19
Need to inject limits into arbitrary [nested] queries.


vpetrovykh:
We want to be able to inject limits into REPL queries in order to prevent a bad query which accidentally fetches billions of objects to cause problems and server-side crashes. This injection needs to happen at all levels of a query, such as in nested shapes and every other place where cardinality is not bounded.

PR #1045, opened 2020-01-06
Add an option to inject an implicit limit into read queries

elprans:
The client can now, via a special header to the Parse, request the server
to inject a constatnt limit into all read queries. This is useful for
contexts where the precise cardinality of output is less important, and
injecting limits ensures that large linked sets in shapes don't bring
everything down.

This requires edgedb/edgedb-python#67.

Fixes: #846.

### DIFF
--- a/edb/edgeql/compiler/func.py
+++ b/edb/edgeql/compiler/func.py
@@ -40,6 +40,7 @@ from edb.edgeql import ast as qlast
 from edb.edgeql import qltypes as ft
 from edb.edgeql import parser as qlparser
 
+from . import astutils
 from . import casts
 from . import context
 from . import dispatch
@@ -111,8 +112,6 @@ def compile_FunctionCall(
             f'{func_name}() cannot be called in a non-session context',
             context=expr.context)
 
-    final_args, params_typemods = finalize_args(matched_call, ctx=ctx)
-
     matched_func_params = func.get_params(env.schema)
     variadic_param = matched_func_params.find_variadic(env.schema)
     variadic_param_type = None
@@ -130,6 +129,12 @@ def compile_FunctionCall(
     )
 
     matched_func_initial_value = func.get_initial_value(env.schema)
+
+    final_args, params_typemods = finalize_args(
+        matched_call,
+        is_polymorphic=is_polymorphic,
+        ctx=ctx,
+    )
 
     if not in_abstract_constraint:
         # We cannot add strong references to functions from
@@ -418,9 +423,6 @@ def compile_operator(
                     hint=f'Possible variants: {detail}.',
                     context=qlexpr.context)
 
-    final_args, params_typemods = finalize_args(
-        matched_call, actual_typemods=actual_typemods, ctx=ctx)
-
     oper = matched_call.func
     assert isinstance(oper, s_oper.Operator)
     env.schema_refs.add(oper)
@@ -428,6 +430,19 @@ def compile_operator(
 
     matched_params = oper.get_params(env.schema)
     rtype = matched_call.return_type
+
+    is_polymorphic = (
+        any(p.get_type(env.schema).is_polymorphic(env.schema)
+            for p in matched_params.objects(env.schema)) and
+        rtype.is_polymorphic(env.schema)
+    )
+
+    final_args, params_typemods = finalize_args(
+        matched_call,
+        actual_typemods=actual_typemods,
+        is_polymorphic=is_polymorphic,
+        ctx=ctx,
+    )
 
     if oper_name in {'std::UNION', 'std::IF'} and rtype.is_object_type():
         # Special case for the UNION and IF operators, instead of common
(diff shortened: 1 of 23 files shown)


tailhook:
Shell crash on any query with `AttributeError: 'BlockingIOConnection' object has no attribute '_fetchall` for me.

tailhook:
Well, I have expected it to be "implicit default limit" rather than "implicit cap on a limit" like it is now:
```
tutorial> SELECT Movie { actors:= (SELECT .actors LIMIT 10) } LIMIT 10;        
{
    Object {
        actors: {
            Object { id: <uuid>'f829820a-3102-11ea-aae4-b7994d9e7100' },
            Object { id: <uuid>'f8298746-3102-11ea-aae4-932bee730fae' },
            ...
        }
    },
    Object {
        actors: {
            Object { id: <uuid>'fa614b3e-3102-11ea-aae4-b39899e09731' },
            Object { id: <uuid>'fa614ca6-3102-11ea-aae4-f3a757190c74' },
            ...
        }
    },
    ...
}
```
Is the former much harder to implement?

tailhook:
Also maybe I'm misreading something but but it looks like limit is applied at the wrong place:
```
        (SELECT/*<pg.SelectStmt at 0x7ff1bb40fa90>*/
                    "Movie~3".id AS "Movie_value~1",
                    "Movie~3".__type__ AS "__type___identity~1"
                FROM
                    "edgedb_e78289cc-2e3c-11ea-b579-d14117bfcd46"."c81b5e48-3102-11ea-828d-9d7cda4b45e5" AS "Movie~3"
                LIMIT 
            (SELECT/*<pg.SelectStmt at 0x7ff1bd111220>*/ (3)::int8 AS "expr~1_value~1" )
            ) AS "Movie~4"
        OFFSET (SELECT/*<pg.SelectStmt at 0x7ff1bb40f970>*/ (10)::int8 AS "expr~7_value~1"  )
        LIMIT (SELECT/*<pg.SelectStmt at 0x7ff1bccb0070>*/ (10)::int8 AS "expr~8_value~1"        )
    )
```
Note, i've applied `\limit 2`. But regardless of that, the `OFFSET` is applied after internal limit, rather than before. So there is no way to retrieve any records after that 100.

edb/edgeql/compiler/func.py
```diff
@@ -533,9 +533,18 @@ def compile_call_arg(arg_ql: qlast.Expr, *,
             dispatch.compile(arg_ql, ctx=fencectx),
             ctx=fencectx)
 
-        return setgen.scoped_set(
-            setgen.ensure_stmt(arg_ir, ctx=fencectx),
-            ctx=fencectx)
+        stmt = setgen.ensure_stmt(arg_ir, ctx=fencectx)
+        if (ctx.expr_exposed and ctx.implicit_limit
+                and isinstance(stmt, irast.SelectStmt) and not stmt.limit):
+            stmt.limit = setgen.ensure_set(
```
tailhook: It looks like `dispatch.compile` already calls `ensure_set` at least for constants.

elprans: Mypy doesn’t understand singledispatch

elprans:
> Well, I have expected it to be "implicit default limit" rather than "implicit cap on a limit" like it is now:

This would require a general way to infer that the argument sets are explicitly bounded.  Our current cardinality inference machinery does not allow that.  It can be done in principle, but is more involved.  Given the implicit limit mode is for interactive REPL only, I'd argue that this isn't very important at this time.

elprans:
> So there is no way to retrieve any records after that 100 (2 in my case)

Right, my "don't mess with existing LIMIT clause" code was off.  Thanks for catching this!

edb/server/mng_port/edgecon.pyx
```diff
@@ -773,14 +797,35 @@ cdef class EdgeConnection:
...
+        for k, v in headers.items():
+            buf.write_int16(<int16_t><uint16_t>k)
+            buf.write_len_prefixed_utf8(str(v))
+
     async def parse(self):
         cdef:
             bint json_mode
             bytes eql
+            dict headers
+            int implicit_limit = 0
```
1st1: Are you aware that this is C 'int' type (16 bit)? I think you're better off not declaring `implicit_limit` in a `cdef` (that will make it a Python object, and thus it will be of unlimited precision)

tailhook:
Since `...` is on its own line most of the time (i.e. no 100 results fit single line), we might want to print a nice comment on it:
```
  { some entry }
   ...  # more results hidden, \limit 100
}
```

edb/server/mng_port/edgecon.pyx
elprans: `int` is 32 bits on most platforms.  I can declare it as `uint64_t`.  Unlimited precision for limit makes little sense, as you are unlikely to get ten quintillion entries in your result :smiley:, and it's better to have defined limits for things in the protocol, even if they're internal/undocumented bits.

1st1:
> Since ... is on its own line most of the time (i.e. no 100 results fit single line), we might want to print a nice comment on it:

+1.

edb/server/mng_port/edgecon.pyx
1st1: > I can declare it as uint64_t

Yeas, that would work.

Issue #985, opened 2019-12-17
Testing data dump features


vpetrovykh:
We need to test how EdgeDB features interact with dump/restore mechanisms. Features that need to be covered:

- [x] annotations
- [x] single links/props
- [x] multi links/props
- [x] link props
- [x] all data types (scalars)
- [x] array props (a couple of scalars)
- [x] tuple and named tuple props (a few scalars)
- [x] same for link props
- [x] constraints (delegated, abstract, concrete)
- [x] indexes
- [x] data should be verifiable 
- [x] functions
- [x] custom scalar
- [x] computable props
- [x] computable links
- [x] computables (a couple complex ones)
- [x] readonly
- [x] required
- [x] computable with custom functions
- [x] inheritance order is preserved
- [x] custom enum (check as default, data and schema type)
- [x] aliases
- [x] separate module
- [x] cross module usage 
- [x] self-referential type
- [x] `on target delete` 
- [x] link target as union type

vpetrovykh:
Readonly appears to be broken, see #1008

vpetrovykh:
It appears that tuples don't work as link props.

vpetrovykh:
It appears that multi links with union types don't generate the right results. Sticking with single links in dump for now.

vpetrovykh:
The only thing I don't have a test for (because it didn't work for me in REPL) is `readme`. Other than that, feel free to close this.

vpetrovykh:
No more changes to test data and schema should be done at this point. Closing the ticket.

PR #1046, opened 2020-01-07
tests: Add `readonly` properties and links to dump tests.

vpetrovykh:
Issue: #985.

### DIFF
--- a/tests/schemas/dump01_default.esdl
+++ b/tests/schemas/dump01_default.esdl
@@ -386,3 +386,64 @@ type SourceA {
         on target delete deferred restrict;
     };
 };
+
+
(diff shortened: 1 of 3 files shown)


vpetrovykh:
I've run into an issue with how `readonly` interacts with inheritance while writing up this test. From a preliminary discussion with @elprans we decided to restrict the inheritance behavior of `readonly` which would make the original problematic case illegal. That particular solution is future-compatible and compatible with the current dump test schema. I'll open a separate issue and PR for the `readonly` problem.

1st1:
@vpetrovykh  From now on please don't add any tests into dump01.

Issue #1047, opened 2020-01-07
Overloading read-only flag


vpetrovykh:
Currently it's possible to override a `readonly` flag when extending a type, which seems fine until we consider the following issues:
```
type A {
   property foo -> str;
}

type B extending A {
   overloaded property foo -> str {
      readonly := True;
   }
}
```
Currently the above schema will not allow the command `UPDATE B SET {foo := 'new'}`, but will allow the `UPDATE A SET {foo := 'new'}`, which will happily update all of `B`. There are 3 options for resolving this:
1) Treat `readonly` like cardinality and completely forbid overloading it in any way.
2) Treat `readonly` like constraints and implement a run-time check instead of a compile-time check for this.
3) Treat `readonly` as a flag that can only be overloaded from `True` to `False` - this is easy to implement as a static check, but leads to possibly counter-intuitive behavior.

Another example to consider when deciding the appropriate strategy is highlighted by multiple inheritance and by union types:
```
type C {
   property foo -> str {
      readonly := False;
   }
}
type D {
   property foo -> str {
      readonly := True;
   }
}
type E extending C, D;
type F extending D, C;

type G {
   link bar0 -> C | D;
   link bar1 -> E;
}
```
Should the `readonly` flag in E and F be same or different (based on inheritance order)? What should `UPDATE G.bar0 SET {foo := 'new'}` vs.`UPDATE G.bar1 SET {foo := 'new'}` do?

Note that, In principle, `readonly` defines valid data operations, whereas `constraint` defines the data domain, so they don't have to behave similarly.

We can implement option 1) as the most conservative for the time being and then decide if we want to relax it in any way later.

vpetrovykh:
This incidentally makes me wonder how we handle interaction between `on target delete allow` and `readonly`. More specifically, should we forbid `on target delete allow` if `readonly` flag is set for the link?

elprans:
This seems to be done.

PR #1048, opened 2020-01-07
edgeql: Forbid redefinition of read-only flag.

vpetrovykh:
The read-only flag presents possible problems of interpretation if it's
allowed to be redefined in inheritance hierarchy. Specifically, consider
the following:

    type A {
        property foo -> str {
            readonly := True
        }
    }

    type B {
        property foo -> str {
            readonly := False;  # not omitted for clarity
        }
    }

    type C extending A, B;

If this is legal, then regardless of how we interpret the correct way to
inherit `readonly` in `C`, we will have unintuitive behavior of objects of
type `C`. It's not obvious to the end-user whether the property `foo` of
C should be more like A or like B.

The solution is to forbid mixing `readonly` flags much like it's
forbidden to mix or redefine cardinality via inheritance.

Issue: #1047

### DIFF
--- a/edb/schema/pointers.py
+++ b/edb/schema/pointers.py
@@ -53,11 +53,9 @@ MAX_NAME_LENGTH = 63
 
 
 def merge_cardinality(target: Pointer, sources: List[so.Object],
-                      field_name: str, *, schema) -> object:
+                      field_name: str, *, schema: s_schema.Schema) -> object:
     current = None
     current_from = None
-
-    target_source = target.get_source(schema)
 
     for source in [target] + list(sources):
         nextval = source.get_explicit_field_value(schema, field_name, None)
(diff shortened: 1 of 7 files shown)


edb/schema/pointers.py
```diff
@@ -92,6 +92,78 @@ def merge_cardinality(target: Pointer, sources: List[so.Object],
...
+        if nextval is not None:
+            if current is None:
+                current = nextval
+                current_from = source
+            elif current is not nextval:
+                current_from_source = current_from.get_source(schema)
+                source_source = source.get_source(schema)
+
+                tgt_repr = (
+                    f'{target_source.get_displayname(schema)}.'
```
elprans: Please use `get_verbosename(with_parent=True)` instead of hand-rolling reprs.

vpetrovykh: Cool, I'll update the `readonly` and `cardinality` error messages to use this instead.

PR #1049, opened 2020-01-07
Simplify limit clause for simple case

tailhook:
Originally limit clause is compiled as:
```
LIMIT
(SELECT/*<pg.SelectStmt at 0x7ff168032f70>*/
    (10)::int8 AS "expr~13_value~1"
    )
```
With this patch it's compiled as:
```
LIMIT 10
```

### DIFF
--- a/edb/pgsql/compiler/clauses.py
+++ b/edb/pgsql/compiler/clauses.py
@@ -152,6 +152,9 @@ def compile_limit_offset_clause(
     if ir_set is None:
         return None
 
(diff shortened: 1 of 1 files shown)


1st1:
> I think this sort of optimization should be done on a more general level, i.e. unwrap constant sets when we know the expression in scope is a singleton.

@elprans do you want to create a to-do issue for yourself and close this PR then?

elprans:
Opened #1050 for a general fix for this.

PR #1051, opened 2020-01-08
Filter irrelevant `readonly` values in DESCRIBE.

vpetrovykh:
Prevent `SET readonly := False` from being generated by `DESCRIBE ... AS
TEXT` if the value is, in fact, inherited. This reduces the noise as
most links and properties are not read-only by default.

### DIFF
--- a/edb/schema/pointers.py
+++ b/edb/schema/pointers.py
@@ -874,6 +874,21 @@ class PointerCommand(constraints.ConsistencySubjectCommand,
         else:
             return super().compile_expr_field(schema, context, field, value)
 
+    def _apply_field_ast(self, schema, context, node, op):
+        if context.descriptive_mode:
+            # When generating AST for DESCRIBE AS TEXT, we want to
+            # omit 'readonly' flag if it's inherited and it actually
+            # has the default value.
+            if op.property == 'readonly':
+                pointer_obj = self.get_object(schema, context)
+                field = type(pointer_obj).get_field('readonly')
+                dval = field.default
+
+                if op.source == 'inheritance' and op.new_value is dval:
+                    return
+
+        super()._apply_field_ast(schema, context, node, op)
+
 
 class SetPointerType(
         referencing.ReferencedInheritingObjectCommand,
--- a/edb/edgeql/codegen.py
+++ b/edb/edgeql/codegen.py
@@ -815,17 +815,7 @@ class EdgeQLSourceGenerator(codegen.SourceGenerator):
         if after_name:
             after_name()
 
-        commands = [
-            comm for comm in node.commands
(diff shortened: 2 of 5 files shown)


1st1:
Can you give before & after output?

1st1:
Also I don't see tests.

vpetrovykh:
I've added a describe test to illustrate that an explicit `readonly := False` is actually preserved, but none of the other inherited defaults for `readonly := False` appear in the other describe tests.

Prior to this change _all_ of the `readonly := False` were filtered out aggressively in codegen, which is not really optimal because ASTs tend to not know the difference between local and inherited fields.

edb/schema/pointers.py
```diff
@@ -874,6 +874,22 @@ def compile_expr_field(self, schema, context, field, value):
         else:
             return super().compile_expr_field(schema, context, field, value)
 
+    def _apply_field_ast(self, schema, context, node, op):
+        if context.descriptive_mode:
+            # When generating AST for DESCRIBE AS TEXT, we want to
+            # omit 'readonly' flag if it's inherited and it actually
+            # has the default value.
+            field_name = 'readonly'
+            if op.property == field_name:
```
1st1:
```suggestion
            if op.property == 'readonly':
```

edb/schema/pointers.py
```diff
@@ -874,6 +874,22 @@ def compile_expr_field(self, schema, context, field, value):
...
+                pointer_obj = self.get_object(schema, context)
+                field = type(pointer_obj).get_field(field_name)
```
1st1:
```suggestion
                field = type(pointer_obj).get_field('readonly')
```

edb/schema/pointers.py
```diff
@@ -874,6 +874,22 @@ def compile_expr_field(self, schema, context, field, value):
         else:
             return super().compile_expr_field(schema, context, field, value)
 
+    def _apply_field_ast(self, schema, context, node, op):
+        if context.descriptive_mode:
+            # When generating AST for DESCRIBE AS TEXT, we want to
+            # omit 'readonly' flag if it's inherited and it actually
+            # has the default value.
+            field_name = 'readonly'
```
1st1:
```suggestion
```

vpetrovykh: I've updated this code. Also rebased on master and resolved the coverage percentage conflict.

edb/schema/pointers.py
vpetrovykh: Updated.

edb/schema/pointers.py
vpetrovykh: Updated

PR #1056, opened 2020-01-08
edgeql: Make it invalid to `DROP REQUIRED` if it wasn't defined.

vpetrovykh:
`DROP REQUIRED` doesn't make sense when the required qualifier is not
set or is inherited. It can also lead to unintuitive results if it's
allowed since `DROP REQUIRED` cannot actually override an inherited `SET
REQUIRED`.

The solution is to only allow `DROP REQUIRED` in the contexts of a link
or property on which it was explicitly set.

### DIFF
--- a/edb/schema/delta.py
+++ b/edb/schema/delta.py
@@ -829,6 +829,7 @@ class ObjectCommand(Command, metaclass=ObjectCommandMeta):
         context: CommandContext,
         *,
         name: Optional[str] = None,
+        default: Union[so.Object, so.NoDefaultT, None] = so.NoDefault,
     ) -> so.Object:
         if name is None:
             name = self.classname
@@ -836,7 +837,7 @@ class ObjectCommand(Command, metaclass=ObjectCommandMeta):
             if rename is not None:
                 name = rename
         metaclass = self.get_schema_metaclass()
-        return schema.get(name, type=(metaclass,))
+        return schema.get(name, type=(metaclass,), default=default)
 
     def compute_inherited_fields(self, schema, context):
         result = {}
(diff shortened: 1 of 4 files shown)


vpetrovykh:
One of the weird side-effects of the original implementation was that not only was `DROP REQUIRED` allowed somewhere down the inheritance line, but it would also result in the inherited property to be overloaded as `required property` if the ancestor was required. Leading to the following:
```
db> CREATE TYPE Foo {CREATE REQUIRED PROPERTY name -> str};
CREATE
db> CREATE TYPE Bar EXTENDING Foo {ALTER PROPERTY name {DROP REQUIRED}};
CREATE
db> ALTER TYPE Foo {ALTER PROPERTY name {DROP REQUIRED}};
ALTER
db> INSERT Bar;
MissingRequiredError: missing value for required property default::Bar.name
db> \d Bar
type default::Bar extending default::Foo {
    required single link __type__ -> schema::Type {
        readonly := true;
    };
    required single property id -> std::uuid {
        readonly := true;
    };
    overloaded required single property name -> std::str;
};
```

edb/schema/delta.py
```diff
@@ -1405,6 +1406,36 @@ def _cmd_tree_from_ast(cls, schema, astnode, context):
                 schema,
                 context.modaliases,
             )
+        elif field.name == 'required' and new_value is False:
```
1st1: `new_value is False` is a bit weird -- can it be replaced with `not new_value`?

1st1: Add a comment if `None` is a legitimate value you're guarding against.

vpetrovykh: Hmmm, the value comes from the AST, so it should never be `None`, unless the AST is malformed. AST creation is entirely under our control, so a malformed AST would be a separate bug which we don't need to handle here. I'll change the expression to `not new_value`.

edb/schema/delta.py
```diff
@@ -1405,6 +1406,36 @@ def _cmd_tree_from_ast(cls, schema, astnode, context):
...
+        elif field.name == 'required' and not new_value:
+            # disallow dropping required that is not locally set
+            parent_obj = parent_op.get_object(schema, context, default=None)
+            errmsg = None
+
+            if parent_obj:
+                local_val = parent_obj.get_explicit_local_field_value(
+                    schema, 'required', None)
+
+                if local_val is not True:
```
1st1: `if not local_val`?

Please avoid using `is True` or `is False` checks in Python code.

vpetrovykh: OK, I'll change it to truthy test. I'll also change the variable name to `local_required` so that the condition reads as `if not local_required:` which conveys the intent a little better, I think.

PR #1058, opened 2020-01-08
Change SET ANNOTATION to CREATE/ALTER ANNOTATION

elprans:
Currently, the same syntax is used both to create and to alter
annotation values on schema objects: `SET ANNOTATION foo := 'value'`.
The current handling of the alter path is hacky and unreliable.
Additionally, there's no valid reason for the syntax (aside from
history) to differ from the usual `CREATE/ALTER` verbs used for every
other object.

### DIFF
--- a/edb/schema/ordering.py
+++ b/edb/schema/ordering.py
@@ -93,6 +93,7 @@ def linearize_delta(
 
     for _key, info in ordered:
         op = info['op']
+        # Elide empty ALTER statements from output.
         if (isinstance(op, sd.AlterObject)
                 and not len(op.get_subcommands())):
             continue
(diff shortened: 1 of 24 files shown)


elprans:
This is incomplete, please don't merge yet.

edb/schema/ordering.py
```diff
@@ -93,9 +93,9 @@ def linearize_delta(
 
     for _key, info in ordered:
         op = info['op']
-        if (isinstance(op, sd.AlterObject)
-                and not len(op.get_subcommands())):
-            continue
+        # if (isinstance(op, sd.AlterObject)
```
vpetrovykh: Remove the commented code?

elprans: Oops. Good catch.  This shouldn't be commented out.

vpetrovykh: Then we ideally need a test for whatever that code does :)

PR #1059, opened 2020-01-08
Set `ssl` if `sslmode` is not truthy

ambv:
Otherwise raises a "referenced before assignment" error.


1st1:
Can you also double check this code in edgedb-python and asyncpg?

Issue #924, opened 2019-12-03
Fix repl to render std::bigint correctly


1st1:
```
yury> select 1n;
{1}    # <- add 'n'
yury> select 1.0n;
{1.0n}
```

tailhook:
We kinda discussed this in the bigint PR and I would not waste much time on this as we're going to have Rust repl in alpha 3 (unless you have a very simple way to do that).

1st1:
I don't think fixing this will take a lot of time, but yeah, the priority is relatively low.

PR #1062, opened 2020-01-08
Fix repl to render std::bigint values properly

1st1:
Issue #924


1st1:
This depends on https://github.com/edgedb/edgedb-python/pull/69

PR #1067, opened 2020-01-09
graphql: Fix an issue with abstract link targets.

### LINKED ISSUE #990: Backward links in views may break GraphQL
The following is legal view and should not break GraphQL:
```
view X := User {
   # this will make a computable link
   # pointing to std::Object
   owned := User.<owner
}
```

vpetrovykh:
Abstract link targets confused the reflection mechanism which tried to
find matching input data for possible insert mutation. Abstract link
targets cannot be directly inserted using data, but only by referring to
existing data with filters.

Aliases should not be exposed for input at all.

Fixes #990.

### DIFF
--- a/edb/graphql/types.py
+++ b/edb/graphql/types.py
@@ -280,10 +280,17 @@ class GQLCoreSchema:
         }
 
     def _get_insert_args(self, typename):
+        # The data can only be a specific non-interface type, if no
+        # such type exists, skip it as we cannot accept unambiguous
+        # data input. It's still possible to just select some existing
+        # data.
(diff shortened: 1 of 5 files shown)


vpetrovykh:
It would effectively fail to parse the GrapQL mutation for those cases. I guess I could add a couple of tests to make sure they fail. As I really don't want to use GraphQL introspection to check (it's unwieldy).

1st1:
> I guess I could add a couple of tests to make sure they fail.

That would work! They should ideally fail in a nice way, i.e. not with some random internal error.

Issue #974, opened 2019-12-17
Update tutorial for Alpha 2


tailhook:
At least `module default {}` is missing. But this I've just stumbled upon. Didn't check other things.

Not sure what is a good time though (before or right after a release?)

tailhook:
Hm, I see it in the master but not on site. Do we publish site docs from a branch?

1st1:
@vpetrovykh could you please take a look at this when you have a minute?

vpetrovykh:
So the code in master should be OK (I've explicitly verified that I could run the tutorial commands in REPL for the current edgedb version). We do actually test the syntax of snippets used in the docs (including SDL in migraitons). Having said that, I need to update the description of SDL in the docs because the way it refers to modules is not accurate anymore. I'll probably overall need to review mentions of "SDL" and "module" in the actual descriptive text rather than in code snippets.

PR #1068, opened 2020-01-09
docs: Update the documentation for SDL.

vpetrovykh:
Update the SDL overview documentaiton to indicate that the SDL describes
the schema as a whole as opposed to a single module.

Add a section to describe SDL module blocks.

Make doc tests aware that SDL may or may not contain a module block and
thus require slightly different handling of the syntax test.

Fixes: #974.

### DIFF
--- a/docs/edgeql/ddl/migrations.rst
+++ b/docs/edgeql/ddl/migrations.rst
@@ -22,26 +22,14 @@ Create a new migration.
         [ ... ]
     "}" ;
 
-    [ WITH [ <module-alias> := ] MODULE <module-name> ]
-    CREATE MIGRATION <name> "{"
-        <ddl-command> ;
-        [ ... ]
-    "}" ;
-
 
 Description
 -----------
 
-``CREATE MIGRATION`` defines a new schema migration for a specific module.
-If *name* is qualified with a module name, then the migration is created
-for that module, owtherwise it is created for the current module, as
-determined by the session or the ``WITH`` block.
-
-There are two forms of ``CREATE MIGRATION`` as shown in the synopsis above.
-The first form uses a specific description of the target schema state and
-generates the necessary DDL commands automatically based on the current and
-the target state.  The second form uses explicit DDL command specifications
-for the migration.
+``CREATE MIGRATION`` defines a migration of the schema to a new state.
+The target schema state is described using :ref:`SDL <ref_eql_sdl>`
+and the migration generates the necessary :ref:`DDL <ref_eql_ddl>`
+commands behind the scenes based on the current and the target state.
 
 **Important:** ``CREATE MIGRATION`` and the follow-up
 :eql:stmt:`COMMIT MIGRATION` must be executed in a transaction block.
@@ -65,11 +53,6 @@ Parameters
     Module contents defined using the declarative :ref:`EdgeDB schema
     definition language<ref_eql_sdl>`.
 
-:eql:synopsis:`<ddl-command>`
-    A list of arbitrary DDL commands.  :ref:`Database <ref_admin_databases>`,
-    :ref:`module <ref_eql_ddl_modules>`, and migration commands cannot be
-    used here.
-
 
 Examples
 --------
@@ -86,21 +69,6 @@ syntax:
             }
         }
     };
-
-Create a new migration for the "payments" module using explicit DDL:
-
-.. code-block:: edgeql
-
-    START TRANSACTION;
-
-    CREATE MIGRATION alter_tx {
-        ALTER TYPE payments::Payment CREATE PROPERTY amount -> str;
-        ALTER TYPE payments::CreditCard CREATE PROPERTY cvv -> str;
-    };
-
-    COMMIT MIGRATION alter_tx;
-
-    COMMIT;
 
 
 COMMIT MIGRATION
@@ -138,7 +106,7 @@ Commit the "alter_tx" migration:
 
 .. code-block:: edgeql
 
-    COMMIT MIGRATION payments::alter_tx;
+    COMMIT MIGRATION init;
 
 
 DROP MIGRATION
--- a/docs/edgeql/sdl/index.rst
+++ b/docs/edgeql/sdl/index.rst
@@ -13,13 +13,18 @@ language and the imperative low-level :ref:`DDL <ref_eql_ddl>`.
 
 SDL is a declarative language optimized for human readability and
 expressing the state of the EdgeDB schema without getting into the
-details of how to arrive at that state.  Each *.esdl* file represents
-a complete schema state for a particular
-:ref:`module <ref_datamodel_modules>`.
+details of how to arrive at that state.  Each *SDL* block represents
+the complete schema state for a given :ref:`database
+<ref_datamodel_databases>`.
 
 Syntactically, an SDL declaration mirrors the ``CREATE`` DDL for the
 corresponding entity, but with all of the ``CREATE`` and ``SET``
-keywords omitted.
+keywords omitted.  The typical SDL structure is to use :ref:`module
+blocks <ref_eql_sdl_modules>` with the rest of the declarations being
+nested in their respective modules.
+
+Since SDL is declarative in nature, the specific order of
+declarations of module blocks or individual items does not matter.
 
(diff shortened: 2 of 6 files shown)


vpetrovykh:
So as I mentioned earlier the code samples (in tutorial and otherwise) in the master branch are up-to-date. But the descriptions of SDL, modules and migrations are in need of fixing.

vpetrovykh:
I've removed mention of DDL migrations since we don't actually test them currently, so in all likelihood they are broken in some ways.

docs/edgeql/sdl/index.rst
```diff
@@ -13,13 +13,19 @@ language and the imperative low-level :ref:`DDL <ref_eql_ddl>`.
...
+<ref_datamodel_databases>`.
 
 Syntactically, an SDL declaration mirrors the ``CREATE`` DDL for the
 corresponding entity, but with all of the ``CREATE`` and ``SET``
-keywords omitted.
+keywords omitted.  These individual declarations must either use
+:ref:`fully-qualified names <ref_eql_fundamentals_name_resolution>` or
+can be enclosed into a *module block* where only short names are
+allowed at top-level and any short name is assumed to belong either to
+the :ref:`module <ref_datamodel_modules>` named by the block or to
```
1st1: This is one long sentence. Please break it into 2-3 sentences and add an example.

docs/edgeql/sdl/index.rst
```diff
@@ -13,13 +13,19 @@ language and the imperative low-level :ref:`DDL <ref_eql_ddl>`.
...
+``std``.  Since SDL is declarative in nature, the specific order of
```
1st1: "Since" should be a new paragraph.

docs/edgeql/sdl/index.rst
```diff
@@ -13,13 +13,19 @@ language and the imperative low-level :ref:`DDL <ref_eql_ddl>`.
...
+details of how to arrive at that state.  Each *SDL* block represents
+the complete schema state for a given :ref:`database
+<ref_datamodel_databases>`.
 
 Syntactically, an SDL declaration mirrors the ``CREATE`` DDL for the
 corresponding entity, but with all of the ``CREATE`` and ``SET``
-keywords omitted.
+keywords omitted.  These individual declarations must either use
+:ref:`fully-qualified names <ref_eql_fundamentals_name_resolution>` or
+can be enclosed into a *module block* where only short names are
```
1st1: `*module block*` should ref `ref_eql_sdl_modules`

docs/edgeql/sdl/modules.rst
```diff
@@ -0,0 +1,111 @@
...
+-----------
+
+The module block declaration defines a new module much like
+:eql:stmt:`CREATE MODULE`.  Unlike its DDL counterpart the module
+block can have sub-declarations:
+
+:sdl:synopsis:`<schema-declarations>`
+    Define various schema items that belong to this module.
+
+Also unlike :eql:stmt:`CREATE MODULE` command, a module block with the
```
1st1: Drop "Also"

tests/test_docs.py
```diff
@@ -254,7 +254,11 @@ def run_block_test(self, block):
                 if lang == 'edgeql':
                     ql_parser.parse_block(snippet)
                 elif lang == 'sdl':
-                    ql_parser.parse_sdl(f'module default {{ {snippet} }}')
+                    # the snippet itself may or may not contain a module block
+                    if 'module' in snippet:
```
1st1: This is rather fragile check. Add a simple regexp to check for `module\s+\w+\s+{`

docs/edgeql/ddl/migrations.rst
```diff
@@ -22,26 +22,14 @@ Create a new migration.
...
-determined by the session or the ``WITH`` block.
-
-There are two forms of ``CREATE MIGRATION`` as shown in the synopsis above.
-The first form uses a specific description of the target schema state and
-generates the necessary DDL commands automatically based on the current and
-the target state.  The second form uses explicit DDL command specifications
-for the migration.
+``CREATE MIGRATION`` defines a migration of the schema to a new state.
+The target schema state is described using :ref:`SDL <ref_eql_sdl>`
+and the migration generates the necessary :ref:`DDL <ref_eql_ddl>`
```
tailhook: I know that it's not a new text, but I still don't get it: why we say DDL commands are generated? I see it as a schema is changed. I have never seen DDL when used CREATE MIGRATION.

docs/edgeql/sdl/modules.rst
```diff
@@ -0,0 +1,111 @@
...
+
+    module <ModuleName> "{"
+      [ <schema-declarations> ]
+      ...
+    "}"
+
+Implicit module declaration happens when another SDL declaration uses
+:ref:`fully-qualified name <ref_eql_fundamentals_name_resolution>` for
+the item that is being declared.  The *module* part of the
+*fully-qualified* name implies that a module by that name must also
```
tailhook: What does this mean? It means that before `type xxx::Yyy {}` I must have an empty block `module xxx {}` ?

docs/edgeql/ddl/migrations.rst
1st1: I'd fix this by replacing `and the migration generates` with `and the migration generates ... behind the scenes`

docs/edgeql/sdl/index.rst
vpetrovykh: OK

docs/edgeql/sdl/index.rst
vpetrovykh: OK, I'll try re-wording it and maybe moving it after the main canonical example. I'll add it in the form "this is an alternative way to do things" with its own example equivalent to the canonical.

docs/edgeql/sdl/index.rst
vpetrovykh: OK

docs/edgeql/sdl/modules.rst
vpetrovykh: OK

tests/test_docs.py
vpetrovykh: Fair enough.

docs/edgeql/ddl/migrations.rst
vpetrovykh: I've changed the wording to "behind the scenes".

We used to have a command `GET MIGRATION` that actually allowed to see the generated DDL, but lots of things since then have changed. The command is currently broken, but we might have some equivalent way to view the representation of the migration in the future.

docs/edgeql/sdl/modules.rst
vpetrovykh: I'll move it closer to the example using this style (which appears later on in the description).

vpetrovykh: I've also reworded it to say that the module will be automatically created based on the fully-qualified name.

docs/edgeql/ddl/migrations.rst
1st1: > The command is currently broken

Is it? I think we have a few tests that depend on it working.

tests/test_docs.py
```diff
@@ -254,7 +254,21 @@ def run_block_test(self, block):
                 if lang == 'edgeql':
                     ql_parser.parse_block(snippet)
                 elif lang == 'sdl':
-                    ql_parser.parse_sdl(f'module default {{ {snippet} }}')
+                    # the snippet itself may either containt a module
+                    # block or have a fully-qualified top-level name
+                    if re.match(
+                            r'''(?xm)
+                                (\bmodule\s+\w+\s*{) |
+                                (^.*
```
1st1: Hah, this is an overkill, IMO, but if it works it's fine :)

docs/edgeql/ddl/migrations.rst
vpetrovykh: Well it's sufficiently broken that when I tried it in REPL with a simple migration it failed (I think it ultimately failed to find some name in the schema, although I don't recall the error). But the main point is that we don't test it too much even if we test it at all, so it's not surprising that it would be broken in some cases.

PR #1070, opened 2020-01-10
Untangle the CREATE DDL path

elprans:
Currently, the order of processing of `CreateObject` subtrees is rather
awkward.  The generic implementation simply iterates over referenced
object subcommands, and various schema classes tack on their own hacks
to implement inheritance propagation etc.  On top of that, handling of
anonymous derived objects, like unions and collection types is hacky and
buggy as well.  To fix this, make the generic `CreateObject.apply` truly
generic, which makes it behave like `AlterObject.apply` already does.
This necessiated the addition of a concept of "prerequisites" for a
command, i.e. subcommands that must run before the object itself is
created.

### DIFF
--- a/edb/schema/inheriting.py
+++ b/edb/schema/inheriting.py
@@ -430,20 +430,23 @@ class CreateInheritingObject(InheritingObjectCommand, sd.CreateObject):
 
         return schema
 
-    def _create_refs(self, schema, context, scls, refdict):
+    def _create_innards(self, schema, context):
         if not context.canonical:
-            local_refs = list(self.get_subcommands(metaclass=refdict.ref_cls))
-            refdict_whitelist = context.inheritance_refdicts
-            if ((refdict_whitelist is None
-                    or refdict.attr in refdict_whitelist)
-                    and (context.inheritance_merge is None
-                         or context.inheritance_merge)):
-                schema = self.inherit_classref_dict(schema, context, refdict)
-            for op in local_refs:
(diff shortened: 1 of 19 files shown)


edb/schema/inheriting.py
```diff
@@ -430,20 +430,23 @@ def _create_begin(self, schema, context):
...
-        else:
-            return super()._create_refs(schema, context, scls, refdict)
+            cmd = sd.CommandGroup()
+            mcls = self.get_schema_metaclass()
+
+            for refdict in mcls.get_refdicts():
+                refdict_whitelist = context.inheritance_refdicts
+                if ((refdict_whitelist is None
+                        or refdict.attr in refdict_whitelist)
+                        and (context.inheritance_merge is None
```
1st1: Three-valued logic?

edb/schema/referencing.py
```diff
@@ -407,7 +397,7 @@ def _propagate_ref_creation(self, schema, context,
                 # with derive_subtype().
                 cmd.set_attribute_value('is_derived', True)
 
-        schema, _ = cmd.apply(schema, context)
+        # schema, _ = cmd.apply(schema, context)
```
1st1: ?

1st1:
Do we need to add a minimal DDL test?

edb/schema/inheriting.py
elprans: Yes. It's hairy, and necessary, unfortunately.

PR #1071, opened 2020-01-10
Add args to the server for automatic temp cluster bootstrap

1st1:
This adds the following args to the edgedb server:

* --temp-dir - to pick a random temp dir for the db cluster
* --auto-shutdown - to shutdown automatically when last connection
  disconnects
* --echo-runtime-info - to echo JSON with runtime parameters
  (runstate dir & the port the server listens on)

### DIFF
--- a/edb/server/buildmeta.py
+++ b/edb/server/buildmeta.py
@@ -18,6 +18,7 @@
 
 
 from __future__ import annotations
+from typing import *  # NoQA
 
 import enum
 import os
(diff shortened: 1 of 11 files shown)


1st1:
Ready for review but don't merge yet; I'm about to start experimenting with jest.

1st1:
Interestingly, the new test used to fail on Python 3.7 consistently and works flawlessly on 3.8. 🤷‍♂️

elprans:
Do you want to enable mypy properly on these modules?  You seem to have added a bunch of type annotations.

1st1:
> Do you want to enable mypy properly on these modules? You seem to have added a bunch of type annotations.

Not right now.

PR #1075, opened 2020-01-10
docs: Add a section on polymorphic queries.

vpetrovykh:
Provide a description of polymorphic nested queries and a usage example
in the "Overview" section. Ensure that this is searchable and
discoverable by a user looking for polymorphic or nested queries.

### DIFF
--- a/docs/edgeql/overview.rst
+++ b/docs/edgeql/overview.rst
@@ -229,3 +229,91 @@ A query is evaluated recursively using the following procedure:
        the results of the invocations are collected to form a single set.
 
 6. Collect the results of all iterations to obtain the final result set.
+
+
+.. _ref_eql_polymorphic_queries:
+
+Polymorphic Queries
+===================
+
+:index: poly polymorphism nested shapes
+
+A link target can be an abstract type, thus allowing objects of
+different extending types to be referenced.  This necessitates writing
+*polymorphic queries* that could fetch different data depending on the
+type of the actual objects.  Consider the following schema:
+
+.. code-block:: sdl
+
+    abstract type Named {
+        required property name -> str {
+            delegated constraint exclusive;
+        }
+    }
+
+    type User extending Named {
+        property avatar -> str;
+        multi link favorites -> Named;
+    }
+
+    type Game extending Named {
+        property price -> int64;
+    }
+
+    type Article extending Named {
+        property url -> str;
+    }
+
+Every ``User`` can have its ``favorites`` link point to either other
+``User``, ``Game``, or ``Article``.  To fetch data related to
+different types of objects in the ``favorites`` link the following
+syntax can be used:
+
+.. code-block:: edgeql
+
+    SELECT User {
+        name,
+        avatar,
+        favorites: {
+            # common to all Named
+            name,
+
+            # specific to Games
+            [IS Game].price,
+
+            # specific to Article
+            [IS Article].url,
+
+            # specific to User
+            [IS Game].avatar,
+
+            # a computable value tracking how many favorites
+            # does my favorite User have?
+            favorites_count := count(
+                # start the path at the root of the shape
+                User.favorites[IS User].favorites)
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


docs/edgeql/overview.rst
```diff
@@ -229,3 +229,74 @@ A query is evaluated recursively using the following procedure:
...
+
+
+.. _ref_eql_polymorphic_queries:
+
+Polymorphic Queries
+===================
+
+:index: nested shapes
+
+Sometimes it is useful to have an abstract type as a link target, thus
```
elprans: Should we also mention unions here?

vpetrovykh: We could _mention_ type unions and link to it without specifically blowing up this example. The type union section has its own polymorphic example already (the only one prior to this and not searchable).

1st1: I don't like words like "sometimes" in reference documentation.  I'd replace with "A link target can be an abstract type ... "

docs/edgeql/overview.rst
```diff
@@ -229,3 +229,74 @@ A query is evaluated recursively using the following procedure:
...
+
+    type Game extending Named {
+        property price -> int64;
+    }
+
+    type Article extending Named {
+        property url -> str;
+    }
+
+The above schema describes some platform where there are users, games
```
1st1: "some platform" reads weird. I'd drop this sentence. The next one `Each User can add any of these entities` can be edited to "Every User`can have its `favorites` link point to either other User, Game, or Article"

docs/edgeql/overview.rst
```diff
@@ -229,3 +229,74 @@ A query is evaluated recursively using the following procedure:
...
+and articles.  Each ``User`` can add any of these entities to
+``favorites``.  A natural question to ask then is how to fetch all this
```
1st1: "A natural question to ask then is how" -- "To fetch data related to different types of objects in the 'favorites' link the following syntax can be used"?

docs/edgeql/overview.rst
```diff
@@ -229,3 +229,74 @@ A query is evaluated recursively using the following procedure:
...
+            # does my favorite User have?
+            favorites_count := count(
+                # start the path at the root of the shape
+                User.favorites[IS User].favorites)
+        }
+    }
+
+The :eql:op:`[IS TypeName] <ISINTERSECT>` construct can be used in
+:ref:`paths <ref_eql_expr_paths>` to restrict the target to a specific
+type.  When it is used in :ref:`shapes <ref_eql_expr_shapes>` it
```
1st1: "When it is used" sentence is redundant here. You literally just showed that 20 words earlier.

docs/edgeql/overview.rst
vpetrovykh: OK

docs/edgeql/overview.rst
vpetrovykh: OK

docs/edgeql/overview.rst
vpetrovykh: OK

docs/edgeql/overview.rst
vpetrovykh: This is deliberate. It is the first time the `[IS type]` is used or mentioned. Given that this is the "overview" section there's no reason to believe that the reader is already familiar with it. So the paragraph is meant to reinforce the idea that there are, in fact, two different usages of this syntax (both of which appear in the example).

docs/edgeql/overview.rst
```diff
@@ -229,3 +229,91 @@ A query is evaluated recursively using the following procedure:
...
 
 6. Collect the results of all iterations to obtain the final result set.
+
+
+.. _ref_eql_polymorphic_queries:
+
+Polymorphic Queries
+===================
+
+:index: nested shapes
```
1st1: add "polymorphism"?

1st1: also "poly"

vpetrovykh: Done.

PR #1084, opened 2020-01-11
Remove the EMIT OIDS option from DESCRIBE

elprans:
This was originally implemented for dump, but that uses a different
approach now, so EMIT OIDS is no longer necessary.

Fixes: #1082


elprans:
> You forgot to delete 'emit' and 'oids' unreserved keywords.

Good catch. Removed.

PR #1093, opened 2020-01-13
Fixes for server command-line

tailhook:
* rename `--pidfile` -> `--pidfile-dir`
* more correct description of `--runstate-dir`

### DIFF
--- a/edb/server/main.py
+++ b/edb/server/main.py
@@ -416,11 +416,11 @@ def bump_rlimit_nofile() -> None:
                 logger.warning('could not set RLIMIT_NOFILE')
 
 
-def get_runstate_dir() -> Optional[str]:
+def _get_runstate_dir_default() -> str:
     try:
         return buildmeta.get_build_metadata_value("RUNSTATE_DIR")
     except buildmeta.MetadataError:
-        return None
+        return '<data-dir>'
 
 
 _server_options = [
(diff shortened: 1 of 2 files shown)


edb/server/main.py
```diff
@@ -468,7 +475,8 @@ def bump_rlimit_nofile() -> None:
     click.option(
         '--runstate-dir', type=PathPath(), default=None,
         help='directory where UNIX sockets will be created '
-             '("/run" on Linux by default)'),
+             '(data-dir or {!r} by default)'
```
elprans: This will print `or 'None'` for non-production builds, which isn't very nice.  I'd rewrite `get_runstate_dir()` as:

```python

def _get_runstate_dir_default():
    try:
        return buildmeta.get_build_metadata_value("RUNSTATE_DIR")
    except buildmeta.MetadataError:
        return '<data-dir>`
```

elprans: and then:

```python
help = f'directory where UNIX sockets and other temporary '
       f'runtime files will be placed ({_get_runstate_dir_default()} by default)'
```
```

tailhook: As far as I understand the `--runstate-dir` is required if no `RUNSTATE_DIR` present in buildmeta and not a `--devmode`.  So I'm not sure how far we want to go here with explanations.

elprans: > As far as I understand the --runstate-dir is required if no RUNSTATE_DIR present in buildmeta and not a --devmode.

Buildmeta and `devmode` are complementary, either is supposed to be present.

tailhook: Okay. Updated.

1st1:
@elprans I think you should cherry-pick this to the a2 branch.

PR #1096, opened 2020-01-13
Add type annotations for schema name and expr modules

RobertoPrevato:


edb/schema/expr.py
```diff
@@ -213,44 +258,48 @@ def compare_values(cls, ours, theirs, *,
...
+                module=modaliases[alias_name],
             )
         )
 
     return qltree
 
 
-def get_expr_referrers(schema, obj):
+def get_expr_referrers(schema: s_schema.Schema, 
+                       obj: so.Object) -> Dict[Any, Any]:
```
elprans: The correct annotation for the return type is `Dict[so.Object, str]`

RobertoPrevato:
@elprans I have slight doubts about this piece of code in `compiled` method:

```
singletons=singletons or (),
```
I used the "or ()" only because of the existing type annotation for the called function.

And also about `get_expr_referrers` annotated with a return type of `Dict[Any, Any]`. Can it be improved, how  `referrers` look like?

edb/schema/expr.py
RobertoPrevato: You answered while I was asking the question, so fast! I do it.

edb/schema/name.py
```diff
@@ -39,7 +39,7 @@ def __new__(
         cls: Type[NameT],
         name: Union[SchemaName, str],
         module: Optional[str] = None,
-    ) -> NameT:
+    ) -> Any:
```
elprans: This change doesn't seem to be correct.

edb/schema/name.py
```diff
@@ -79,7 +79,7 @@ def is_qualified(name: Union[SchemaName, str]) -> bool:
 
 class UnqualifiedName(SchemaName):
 
-    def __new__(cls: Type[NameT], name: str) -> NameT:
+    def __new__(cls: Type[NameT], name: str) -> Any:
```
elprans: You can omit the annotation on `cls` and change the return annotation to `UnqualifiedName`.

edb/schema/name.py
RobertoPrevato: It looked wrong to me, too. I did it because, if left "NameT", mypy complains this way:

image

`edb/schema/name.py:70:9: error: Returning Any from function declared to return "NameT"`

How should it be fixed? It looks like # type: ignore gives Any.

RobertoPrevato: Should I remove the type annotation from cls as well?

edb/schema/name.py
RobertoPrevato: This presents the same problem described in the comment above. (type: ignore -> Any; returning Any from function declared to return ...)

edb/schema/name.py
elprans: No, keep the `cls` annotation. We need the `TypeVar` here to make `UnqualifiedName.__new__` happy.  Use `cast` on the result: `cast(NameT, result)`

edb/schema/expr.py
```diff
@@ -99,20 +117,28 @@ def from_ast(
...
+        as_fragment: bool = False,
+        modaliases: Optional[Mapping[Optional[str], str]] = None,
+        parent_object_type: Optional[so.ObjectMeta] = None,
+        anchors: Optional[
+            Mapping[Union[str, qlast_.SpecialAnchorT], so.Object]
+        ] = None,
+        path_prefix_anchor: Optional[qlast_.SpecialAnchorT] = None,
+        allow_generic_type_output: bool = False,
+        func_params: Optional[s_func.ParameterLikeList] = None,
+        singletons: Optional[Sequence[s_types.Type]] = None,
```
elprans: Just use the `()` as the default here instead of `None` and drop `Optional`.

edb/schema/name.py
RobertoPrevato: I think so, since I cannot use `isinstance` in this case, with a generic.

PR #1104, opened 2020-01-14
Keep coverage results when running edb test --cov

RobertoPrevato:
* ~~instead of using a temporary folder, puts `coverage` results inside a folder "coverage_results" (folder added to .gitignore)~~
* ~~at every new run with --cov option, this folder is deleted if it exists, then created under `cwd`~~
* before exiting the context of the temporary folder, copies the `.coverage` file containing results into `cwd`
* this allows to create reports with `coverage` cli, for example `coverage html`, then `firefox htmlcov/index.html`

### DIFF
--- a/edb/tools/test/__init__.py
+++ b/edb/tools/test/__init__.py
@@ -154,6 +154,12 @@ def test(*, files, jobs, include, exclude, verbose, quiet, debug,
     sys.exit(result)
 
 
+def get_coverage_results_folder_path() -> str:
+    container_folder = pathlib.Path().absolute() / 'coverage_results'
+    os.makedirs(container_folder, exist_ok=True)
+    return str(container_folder)
+
+
 @contextlib.contextmanager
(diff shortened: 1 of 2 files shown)


edb/tools/test/__init__.py
```diff
@@ -154,6 +154,31 @@ def test(*, files, jobs, include, exclude, verbose, quiet, debug,
     sys.exit(result)
 
 
+def _prepare_coverage_result_folder(container_folder: pathlib.Path) -> None:
```
elprans: I think we can still use a temporary directory for intermediate files, but write the final `.coverage` file the the current directory.  That way you don't need this code.  Dealing with other generated files (such as HTML) would be left to the user.

RobertoPrevato: Thanks! I didn't know that .coverage output file is sufficient to generate reports (I understand now it aggregates all results). I'll do it :+1:

PR #1120, opened 2020-01-16
Expose --insecure-bootstrap-defaults

ambv:
This allows bootstrapping a database that lets anybody in from any address.

Also: for non-managed databases, don't attempt to set buffer, cache and memory
sizes.


ambv:
For the record: we decided not to expose this because it creates a *permanent* insecurity for the given database. Instead, we will be providing a TCP-based version of `--admin` that is available only per run.

PR #1131, opened 2020-01-17
graphql: Add bigint tests and fix some translation bugs.

vpetrovykh:
Currently we don't distinguish between limited and unlimited
precision/size numbers in GraphQL. Which means that by default it
makes sense to treat `Int` as `bigint` and `Float` as `decimal`
(especially for variables). This preserves the precision when it's
needed and can be downcast to the specific type the expression
requires.

Since `int64` is implicitly castable into `bigint` it's possible to
treat small `Int` literals as `int64` and large ones as `bigint`.

Stop mapping `bigint` onto `Float` and use `Int` instead.

This technically breaks the sprict spec for `Int` and `Float`.
Introducing custom types would be a separate PR.

Issue #1138


edb/edgeql/ast.py
```diff
@@ -265,9 +268,12 @@ def from_python(cls, val: typing.Any) -> BaseConstant:
         elif isinstance(val, bool):
             return BooleanConstant(value='true' if val else 'false')
         elif isinstance(val, int):
-            return IntegerConstant(value=str(val))
+            if val > MAX_INT64:
+                return BigintConstant(value=f'{val}n')
+            else:
+                return IntegerConstant(value=str(val))
```
1st1: What it it's smaller than `-MAX_INT64`? ;)  In general I don't like hacks like this and I'm not sure this is a good place to put them.

vpetrovykh: Oops! I fixed the `int64` range to include lower and upper bounds.
The reason why I put this here is that unlike `decimal.Decimal` Python doesn't have a clean way of distinguishing between `int64` and `bigint` and we have the check for `decimal.Decimal` right here.

vpetrovykh:
*sigh*
It appears that either I missed it last time or they ret-conned it, but the specs for `Int` and `Float` in GraphQL restrict them to be `int32` and `float64` respectively. It looks like we need to review how custom scalars work and perhaps define the same base scalars as we have in EdgeDB.

edb/edgeql/ast.py
1st1: @elprans what do you think of this `if`?

elprans: I think it's valid.  We decode bigints into Python int, so I don't see why the reverse would be incorrect.

1st1: I think this can lead to weird bugs at some point, like constant folder suddenly producing bigints for int32*int32 or something. I think it would be cleaner to handle this at the GraphQL level. Here we should just raise an error in case we detect an overflow. Idk, just something about this change doesn't look right to me. If you think it's fine then OK.

elprans: Well, then we need some `int` subclass in Python, otherwise bigint values become inexpressible.

1st1: IMO we should raise an error instead of returning a BigIntConstant. The BigInt conversion should be handled in the GraphQL code directly.

1st1: How so?  Just add `BigintConstant.from_python()` and be done with it. No need to a subclass.

vpetrovykh: So you're suggesting that at the call site we always know which AST to generate, whether it's `IntegerConstant` or `BigintConstant`? Well, then why do we not apply that logic to all of the constant nodes and why do we have a `BaseConstant.from_python` at all? My understanding of this method was that it's precisely for the purpose of handling a value that could be anything.

1st1: > So you're suggesting that at the call site we always know which AST to generate, 

In GraphQL code — yes.

> that logic to all of the constant nodes and why do we have a BaseConstant.from_python at all?

Do you see any such places where suddenly returning BigInt wouldn't be a bug?

> My understanding of this method was that it's precisely for the purpose of handling a value that could be anything.

I have a different understanding: it's there to generate a `std::int64` (the default int literal type) in most places. Silently injecting `std::bigint` there can potentially break things, instead of raising an error "i can't express your constant as int64, please fix or use BigIntConstant.from_python()"

vpetrovykh: I am confused by why do we have `from_python` in `BaseConstant` - an abstract AST node - at all? This version of `from_python` is currently handling mapping a lot of the different Python values to some specific "constant" AST. If it's there, it should do the right thing for all values it handles. It is happy to distinguish `float` and `decimal.Decimal`, so why shouldn't it distinguish `int` based on the range? It is _illegal_ to use an `IntegerConstant` with a value > 2**63, but the master implementation will gladly do just that.
If the intent is to always use the specific non-abstract constant AST nodes, then let's drop this method from the abstract base class.

As far as usage of this converter in GraphQL, all I'd end up doing is moving the same `if MIN_INT64 ...` code into the call-site in GraphQL where I'm processing an "int-looking" literal which may be an `int64` or may be `bigint` quite legitimately. Since `int64` is implicitly castable into `bigint` it's fine to always use `int64` for smaller literals. 
The EdgeQL queries:
```
SELECT Foo { bar } FILTER .bar > 2222222222222222222222222n;
SELECT Foo { bar } FILTER .bar > 2;
```
In GraphQL would be:
```
query q1 {
  Foo(
    filter: {bar: {gt: 2222222222222222222222222}}
  ) {bar}
}

query q2 {
  Foo(
    filter: {bar: {gt: 2}}
  ) {bar}
}
```

elprans: We use `from_python` mainly to convert `schema.Object.field` to `edgeql.AST.field`.  Since fields are typed anyway, we don't really need to have the dispatch in `BaseConstant`, we can just reimplement it in `AlterObjectProperty._get_ast` using the field type information.  GraphQL can do it's own thing.  We don't have constant folding other than string concatenation, so it's not an issue for that use case either.

Bottom line: the proper action is to kill `BaseConstant.from_python`.

1st1: In this particular PR, please just don't touch `from_python`. Fix this in GraphQL.

1st1: We can drop `from_python` in a separate PR.

vpetrovykh: OK, I've moved this logic to GraphQL code, leaving `BaseConstant.from_python` unaltered.

edb/graphql/types.py
```diff
@@ -56,6 +57,32 @@
...
 
 
+def coerce_int(value):
+    if isinstance(value, int):
+        num = value
+    else:
+        try:
+            num = int(value)
+        except ValueError:
+            num = int(float(value))
```
1st1: This line is weird. Can you explain why you use `int(float(...` here?

1st1:
@elprans Please review this PR too.

edb/graphql/types.py
vpetrovykh: That's copy-pasted from graphql core, I just changed the min/max value test. But it does look weird so we probably should drop that cast. (Perhaps the intent was to coerce `1.0` or `1e5` to an integer, but that's still an odd thing to do, since this is supposedly processing a string representation).

PR #1136, opened 2020-01-17
Only set buffer, cache and memory sizes in managed clusters

ambv:
Additionally, sprinkled some typing here and there.

### DIFF
--- a/edb/server/config/__init__.py
+++ b/edb/server/config/__init__.py
@@ -18,6 +18,7 @@
 
 
 from __future__ import annotations
+from typing import *  # NoQA
 
 from .ops import OpLevel, OpCode, Operation, lookup
 from .ops import spec_to_json, to_json, from_json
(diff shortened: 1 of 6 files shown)


edb/server/config/__init__.py
```diff
@@ -33,19 +34,19 @@
...
     'OpLevel', 'OpCode', 'Operation',
-    'ConfigType', 'Port',
+    'ConfigType',
     'load_spec_from_schema',
     'generate_config_query',
 )
 
 
-_settings = None
+_settings = Spec()
```
ambv: Note: I changed the default because there were many places where the server is using `get_settings()` indiscriminately without checking if the settings are None or not.

In a real scenario this won't happen so initially I put an assert in `get_settings()` to check they are in fact not None. It made a bunch of tests fail which could be worked around (clumsily but functionally). What broke the camel's back for me though was that the compiler depends on the settings to figure out if we are in test mode and if to do constant folding. Working around *that* would be too much trouble so I just changed the default here, all tests pass.

PR #1137, opened 2020-01-17
Add annotations to schema inheriting

RobertoPrevato:
@elprans, this is the draft PR for `inheriting.py`. It's a work in progress; it's down to 10 mypy errors, but there are unclear parts for me. Some parts are totally work in progress, for example the mind-bending annotations I drafted for `injeriting.py` `delta_bases` function.

1. the function InheritingObjectCommand._apply_rebase_ast seems to be never used in the tests
2. I need help with the delta_bases function, and the three places with TODOs :blush:

### DIFF
--- a/edb/edgeql/ast.py
+++ b/edb/edgeql/ast.py
@@ -623,7 +623,7 @@ class SetSpecialField(BaseSetField):
 class ObjectDDL(DDLCommand, CompositeDDL):
     __abstract_node__ = True
     name: ObjectRef
-    commands: typing.Sequence[DDLCommand]
+    commands: typing.List[DDLCommand]
 
 
 class CreateObject(ObjectDDL):
--- a/edb/schema/inheriting.py
+++ b/edb/schema/inheriting.py
@@ -39,10 +39,9 @@ if TYPE_CHECKING:
 
 
 class InheritingObjectCommand(sd.ObjectCommand):
-
-    def _create_begin(self,
-                      schema: s_schema.Schema,
-                      context: sd.CommandContext) -> s_schema.Schema:
+    def _create_begin(
+        self, schema: s_schema.Schema, context: sd.CommandContext
+    ) -> s_schema.Schema:
         schema = super()._create_begin(schema, context)
 
         if not context.canonical:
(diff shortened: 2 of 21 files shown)


edb/edgeql/ast.py
```diff
@@ -623,7 +623,7 @@ class SetSpecialField(BaseSetField):
 class ObjectDDL(DDLCommand, CompositeDDL):
     __abstract_node__ = True
     name: ObjectRef
-    commands: typing.Sequence[DDLCommand]
+    commands: typing.List[DDLCommand]
```
RobertoPrevato: I am not sure this change is right. But it looks reasonable, since we use the method "append" on commands property, so assuming it looks like a `List`?

elprans: Yes, `List` is correct.

edb/schema/inheriting.py
```diff
@@ -29,27 +30,47 @@
 from . import name as sn
 from . import objects as so
 from . import utils
+from edb.schema import schema as s_schema
+
+if TYPE_CHECKING:
+    from edb.schema import objtypes as s_objtypes
```
elprans: `inheriting.py` is a generic module, you shouldn't be using concrete implementations for the annotations.

edb/schema/inheriting.py
```diff
@@ -29,27 +30,47 @@
...
             schema = self._update_inherited_fields(schema, context)
 
         return schema
 
-    def _alter_begin(self, schema, context, scls):
-        schema = super()._alter_begin(schema, context, scls)
+    def _alter_begin(self,
+                     schema: s_schema.Schema,
+                     context: sd.CommandContext,
+                     scls: s_objtypes.ObjectType) -> s_schema.Schema:
```
elprans: The correct type is `so.InheritingObjectBase`

edb/schema/inheriting.py
```diff
@@ -58,11 +79,18 @@ def _update_inherited_fields(self, schema, context):
             schema, 'inherited_fields', inherited_fields)
         return schema
 
-    def inherit_fields(self, schema, context, scls, bases, *, fields=None):
+    def inherit_fields(
+            self,
+            schema: s_schema.Schema,
+            context: sd.CommandContext,
+            scls: s_objtypes.ObjectType,
```
elprans: Ditto

edb/schema/inheriting.py
RobertoPrevato: I understand: the concept of inheritance is generic. Just to be 100% sure, you mean that I shouldn't import `objtypes` module because it contains concrete classes, right? This wasn't clear to me while I started annotating this module: the main reason is that the `scls` variable (so.InheritingObjectBase) is used calling methods that are not defined in this base class, like `get_default_base_name` and `inheritable_fields`, which are defined on higher level classes.

In such situations, can we add abstract methods on the base class, to clarify the meaning of the class and its "interface" / "protocol", and satisfy mypy at the same time?

For example, adding to _InheritingObjectBase_:
```py
    def inheritable_fields(self) -> Iterable[Any]:
        raise NotImplementedError

    @classmethod
    def get_default_base_name(self) -> str:
        raise NotImplementedError
```

This is how I would do it (actually, in such cases I use @abstactmethod decorator from abc); do you agree with such approach or do you apply something else?

PR #1139, opened 2020-01-17
Make sys::get_version() return the actual version of the running server

elprans:
Currentlly, `sys::get_version()` returns the version of of a server that
was used to *bootstrap* an instance, since version metadata is written
into the instancedata storage.  This is obviously wrong, since one might
run a different, but compatible, version of EdgeDB on the same data
instance.  Fix this by switching `sys::get_version()` to use the session
state table instead.

### DIFF
--- a/edb/server/buildmeta.py
+++ b/edb/server/buildmeta.py
@@ -21,9 +21,11 @@ from __future__ import annotations
 from typing import *  # NoQA
 
 import enum
+import json
 import os
 import pathlib
-from typing import *  # NoQA
+
+import immutables as immu
 
 import edb
 from edb.common import devmode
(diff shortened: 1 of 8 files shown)


edb/server/buildmeta.py
```diff
@@ -171,3 +171,32 @@ def get_version() -> Version:
...
+
+_version_dict: Optional[Dict[str, Any]] = None
+
+
+def get_version_dict() -> Dict[str, Any]:
+    global _version_dict
+
+    if _version_dict is None:
+        ver = get_version()
+        _version_dict = {
```
1st1: Make it immutable? :)

elprans: `immutables.Map`?

1st1: yeah

PR #1142, opened 2020-01-18
Make expression aliases properly introspectable

elprans:
`schema::Type` now has an `expr` property that is non-empty if the
`Type` is a type variant, i.e. produced by an alias declaration.  The
new `is_from_alias` boolean computable is also added to simplify checks.

The REPL is taught to filter out type variants produced by aliases from
`\lt` and friends, and a new `\la` command is added to list aliases
specifically.

### DIFF
--- a/edb/lib/schema.edgeql
+++ b/edb/lib/schema.edgeql
@@ -47,6 +47,11 @@ CREATE ABSTRACT TYPE schema::Object {
 CREATE ABSTRACT TYPE schema::Type EXTENDING schema::Object;
 CREATE TYPE schema::PseudoType EXTENDING schema::Type;
 
+ALTER TYPE schema::Type {
+    CREATE PROPERTY expr -> std::str;
+    CREATE PROPERTY is_from_alias := EXISTS(.expr);
+};
+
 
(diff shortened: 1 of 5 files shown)


edb/lib/schema.edgeql
```diff
@@ -47,6 +47,11 @@ CREATE ABSTRACT TYPE schema::Object {
 CREATE ABSTRACT TYPE schema::Type EXTENDING schema::Object;
 CREATE TYPE schema::PseudoType EXTENDING schema::Type;
 
+ALTER TYPE schema::Type {
```
1st1: I don't quite like using schema::Type for this. Can we add schema::ExpressionAlias?

elprans: Well, those objects *are* types (technically abstract type variants), so this introspection reflection is actually true. Separating them would require a whole bunch of hacks, because you would need to replicate the whole `Type` hierarchy only in the `Alias` form, i.e. `ObjectAlias`, `TupleAlias` etc. I actually tried going in that direction and it went ugly real fast.

1st1: ok

PR #1144, opened 2020-01-18
Implement the "edb wipe" command

elprans:
This development tool allows removing all traces of EdgeDB from the 
specified PostgreSQL cluster.  Useful for cases when debugging bootstrap 
and setting up new clusters is expensive.

There are a couple of prerequisite commits that are tagging along this
PR as well.

### DIFF
--- a/edb/tools/wipe.py
+++ b/edb/tools/wipe.py
@@ -1,0 +1,200 @@
+#
+# This source file is part of the EdgeDB open source project.
+#
+# Copyright 2016-present MagicStack Inc. and the EdgeDB authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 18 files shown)


edb/tools/wipe.py
```diff
@@ -0,0 +1,192 @@
...
+    '-D',
+    '--data-dir',
+    type=AbsPath(),
+    help='database cluster directory')
+@click.option(
+    '-y',
+    'yes',
+    is_flag=True,
+    help='assume Yes response to all questions')
+def wipe(*, postgres_dsn, data_dir, yes):
```
1st1: Would be great to also add `--dry` to do a dry run.

elprans: Done.

PR #1146, opened 2020-01-19
Fix new issue url

Fogapod:
Old link uses empty template: https://github.com/edgedb/edgedb/issues/new?template=bug_report.md
New link: https://github.com/edgedb/edgedb/issues/new?template=bug_report.yml


elprans:
The URL in the exception is actually correct, it's the template files that were named incorrectly. Now fixed.  Thanks for noticing!

PR #1149, opened 2020-01-19
Add annotations for schema inheriting module

RobertoPrevato:
This is a brand new PR because I made a mistake with the previous branch and preferred to create a new one.

### DIFF
--- a/edb/schema/inheriting.py
+++ b/edb/schema/inheriting.py
@@ -18,9 +18,11 @@
 
 
 from __future__ import annotations
+from typing import *  # NoQA
 
 from edb.common import struct
 from edb.edgeql import ast as qlast
+from edb.schema import schema as s_schema
 
 from edb import errors
 
@@ -31,9 +33,16 @@ from . import objects as so
 from . import utils
 
 
+if TYPE_CHECKING:
+    from edb.schema import referencing as s_referencing
+    from edb.schema import types as s_types
+
+
 class InheritingObjectCommand(sd.ObjectCommand):
 
-    def _create_begin(self, schema, context):
+    def _create_begin(self,
+                      schema: s_schema.Schema,
+                      context: sd.CommandContext) -> s_schema.Schema:
         schema = super()._create_begin(schema, context)
 
         if not context.canonical:
@@ -41,15 +50,22 @@ class InheritingObjectCommand(sd.ObjectCommand):
 
         return schema
 
-    def _alter_begin(self, schema, context, scls):
-        schema = super()._alter_begin(schema, context, scls)
+    def _alter_begin(self,
+                     schema: s_schema.Schema,
+                     context: sd.CommandContext,
+                     scls: so.InheritingObjectBase) -> s_schema.Schema:
+        schema = super()._alter_begin(schema, context, scls)  # type: ignore
 
+        assert isinstance(schema, s_schema.Schema)
         if not context.canonical:
             schema = self._update_inherited_fields(schema, context)
 
         return schema
 
-    def _update_inherited_fields(self, schema, context):
+    def _update_inherited_fields(self,
+                                 schema: s_schema.Schema,
+                                 context: sd.CommandContext
+                                 ) -> s_schema.Schema:
         current_inh_fields = self.scls.get_inherited_fields(schema)
         new_inh_fields = self.compute_inherited_fields(schema, context)
         inherited_fields = current_inh_fields.update(new_inh_fields)
@@ -58,11 +74,19 @@ class InheritingObjectCommand(sd.ObjectCommand):
             schema, 'inherited_fields', inherited_fields)
         return schema
 
-    def inherit_fields(self, schema, context, scls, bases, *, fields=None):
+    def inherit_fields(self,
+                       schema: s_schema.Schema,
+                       context: sd.CommandContext,
+                       scls: so.InheritingObjectBase,
+                       bases: Tuple[so.Object, ...],
+                       *,
+                       fields: Optional[Iterable[str]] = None
+                       ) -> s_schema.Schema:
         mcls = self.get_schema_metaclass()
 
         if fields is not None:
-            field_names = set(scls.inheritable_fields()) & set(fields)
+            field_names: Iterable[str] = set(scls.inheritable_fields()) \
+                & set(fields)
         else:
             field_names = scls.inheritable_fields()
 
@@ -99,10 +123,29 @@ class InheritingObjectCommand(sd.ObjectCommand):
 
         return schema
 
-    def get_inherited_ref_layout(self, schema, context, refdict):
+    def get_inherited_ref_layout(
+        self,
+        schema: s_schema.Schema,
+        context: sd.CommandContext,
+        refdict: so.RefDict
+    ) -> Dict[
+        s_schema.Schema,
+        Tuple[
+            s_referencing.CreateReferencedObject,
+            qlast.ObjectDDL,
+            List[InheritingObject],
+        ],
+    ]:
         attr = refdict.attr
         bases = self.scls.get_bases(schema)
-        refs = {}
+        refs: Dict[
+            s_schema.Schema,
+            Tuple[
+                s_referencing.CreateReferencedObject,
+                qlast.ObjectDDL,
+                List[InheritingObject],
+            ],
+        ] = {}
 
         for base in bases.objects(schema):
             base_refs = base.get_field_value(schema, attr)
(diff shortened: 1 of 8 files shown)


edb/schema/objects.py
```diff
@@ -2031,6 +2031,13 @@ class InheritingObjectBase(Object):
         bool,
         default=False, compcoef=0.909)
 
+    def inheritable_fields(self) -> Iterable[Any]:
```
RobertoPrevato: I added two "abstract" methods, raising NotImplementedError, to explicit the expected protocol of this class, since later it's assumed that these two methods exist on the implementers. I am not sure you agree with this approach, but I think it makes sense.

edb/schema/inheriting.py
```diff
@@ -29,40 +30,70 @@
 from . import name as sn
 from . import objects as so
 from . import utils
+from edb.schema import schema as s_schema
```
elprans: `from . import schema as s_schema` for visual consistency with other imports.  Also, please insert imports in lexicographical order.

edb/schema/objects.py
elprans: This is fine, thanks!

edb/schema/inheriting.py
RobertoPrevato: You are right, Elvis. I indeed used Black and made the mistake to not separate the diff chunks. I will fix as you ask, and the next time I will pay attention to not repeat.

RobertoPrevato:
@elprans, the file should be correct now, can you please review?

edb/schema/inheriting.py
```diff
@@ -127,8 +170,20 @@ def get_inherited_ref_layout(self, schema, context, refdict):
...
+        refdict: so.RefDict,
+        present_refs: Dict[
+            s_schema.Schema,
+            Tuple[
+                s_referencing.CreateReferencedObject,
+                qlast.ObjectDDL,
+                List[InheritingObject],
+            ],
+        ],
+    ) -> Dict[Any, Any]:
```
elprans: `Dict[str, Type[sd.ObjectCommand]]`

edb/schema/inheriting.py
```diff
@@ -389,13 +495,22 @@ class AlterInherit(sd.Command):
     astnode = qlast.AlterAddInherit, qlast.AlterDropInherit
 
     @classmethod
-    def _cmd_tree_from_ast(cls, schema, astnode, context):
+    def _cmd_tree_from_ast(
+        cls,
+        schema: s_schema.Schema,
+        astnode: Any,
```
elprans: `qlast.ObjectDDL`

edb/schema/inheriting.py
```diff
@@ -446,10 +563,17 @@ def _create_innards(self, schema, context):
...
+        result = super()._create_innards(schema, context)
+        assert isinstance(result, s_schema.Schema)
+        return result
 
     @classmethod
-    def _cmd_tree_from_ast(cls, schema, astnode, context):
+    def _cmd_tree_from_ast(
+        cls,
+        schema: s_schema.Schema,
+        astnode: Any,
```
elprans: `qlast.ObjectDDL`

edb/schema/inheriting.py
```diff
@@ -461,7 +585,13 @@ def _cmd_tree_from_ast(cls, schema, astnode, context):
 
         return cmd
 
-    def _apply_field_ast(self, schema, context, node, op):
+    def _apply_field_ast(
+        self,
+        schema: s_schema.Schema,
+        context: sd.CommandContext,
+        node: Any,
```
elprans: `qlast.ObjectDDL`

edb/schema/inheriting.py
```diff
@@ -529,7 +665,12 @@ def inherit_classref_dict(self, schema, context, refdict):
 class AlterInheritingObject(InheritingObjectCommand, sd.AlterObject):
 
     @classmethod
-    def _cmd_tree_from_ast(cls, schema, astnode, context):
+    def _cmd_tree_from_ast(
+        cls,
+        schema: s_schema.Schema,
+        astnode: Any,
```
elprans: `qlast.ObjectDDL`

edb/schema/inheriting.py
```diff
@@ -742,21 +920,31 @@ def delta(cls, old, new, *, context=None, old_schema, new_schema):
...
+        value: Any,
+    ) -> None:
+        assert isinstance(scls, so.InheritingObjectBase)
         inherited_fields = scls.get_inherited_fields(schema)
         delta.add(sd.AlterObjectProperty(
             property=fname, old_value=None, new_value=value,
             source='inheritance' if inherited_fields.get(fname) else None))
 
-    def inheritable_fields(self):
+    def inheritable_fields(self) -> Iterable[Any]:
```
elprans: `Iterator[str]`

edb/schema/inheriting.py
```diff
@@ -770,17 +958,17 @@ def get_topmost_concrete_base(self, schema):
         raise errors.SchemaError(
             f'{self.get_verbosename(schema)} has no non-abstract ancestors')
 
-    def get_base_for_cast(self, schema):
-        if self.is_enum(schema):
+    def get_base_for_cast(self, schema: s_schema.Schema) -> so.Object:
+        if self.is_enum(schema):  # type: ignore
```
elprans: Consider moving this check to the type that actually implements `is_enum`.

edb/schema/inheriting.py
```diff
@@ -770,17 +958,17 @@ def get_topmost_concrete_base(self, schema):
...
             # all enums have to use std::anyenum as base type for casts
             return schema.get('std::anyenum')
         else:
             return self.get_topmost_concrete_base(schema)
 
     @classmethod
-    def get_root_classes(cls):
+    def get_root_classes(cls) -> Tuple[s_types.Type, ...]:
```
elprans: `Tuple[sn.Name]`

edb/schema/inheriting.py
RobertoPrevato: Yes, it looks better :+1:

edb/schema/inheriting.py
RobertoPrevato: @elprans, can we use a generic like in the solution I propose in my last commit? I tried this way because using exactly ObjectDDL was then causing problems in the annotations in `annos` module, and I understand that subclasses will always be called with their specific ast node (?).

By the way, here and in the other places where I left "Any", I did exactly because I needed your input (not because I was thinking it was right to leave Any).

elprans: > causing problems in the annotations in annos module, and I understand that subclasses will always be called with their specific ast node (?).

Yes, and the subclasses should be called with a subclass of `ObjectDDL`, so there should be no problem.  A generic is overkill here, since you're only annotating a single argument.  Can you show the actual error you're seeing?

RobertoPrevato: `Argument 2 of "_cmd_tree_from_ast" is incompatible with supertype "CreateInheritingObject"; supertype defines the argument type as "ObjectDDL"`

image

I know it could be resolved using qlast.ObjectDDL and assert in the function body that it's an instance of the expected type, but this way looks also not optimal to me.

elprans: @ambv Aren't overloaded methods in subtypes supposed to be covariant with respect to the arguments?

RobertoPrevato: Maybe there is a better way in mypy.

FWIW, in C# this would be a normal case for a generic, but I agree that in Python generics looks more like "heavy machinery".

```csharp
  public class ObjectDDL
  {

  }

  public class SpecificObjectDDL : ObjectDDL
  {
    public string Foo { get; set; }
  }

  public class Example<T> where T : ObjectDDL
  {
    public virtual void Something(T arg)
    {

    }
  }

  public class SpecificExample : Example<SpecificObjectDDL>
  {
    public override void Something(SpecificObjectDDL arg)
    {
      // throw exception if arg is null (not mandatory)
      Console.WriteLine(arg.Foo);
    }
  }
```

elprans: Yes, I always forget, this is the Liskov Substitution Principle.  The explicit use of a covariant generic is needed in the parent class.

RobertoPrevato: In conclusion, what pattern do you prefer in this case? Do you prefer subclasses to have the generic annotation (ObjectDDL) and have assert isinstance?

elprans: The solution in your last commit is correct, Roberto.

Issue #1151, opened 2020-01-20
The dangers of using `decimal` with JSON should be documented.


vpetrovykh:
I realized that any time we use `<json>` on a `decimal`, the default JSON decoder would simply read it as a `float` of some sort. This affects both GraphQL and EdgeQL over HTTP (and in general any EdgeQL query that relies on JSON conversion). So it appears like a subtle foot-gun, where you try a query in REPL and it works just fine, but then it silently loses precision in actual code because you used `edgeql+http` protocol.

I think that we need to make sure that the docs for `decimal` as well as for "EdgeQL over HTTP" and "GraphQL over HTTP" should have warnings about using decimals, much like the one in "EdgeDB JavaScript Driver" "Datatypes" section.

_Technically_, this is a JSON decoder issue rather than our issue, but we should still be very clear about the problems you might run into when working with `decimal` values.

PR #1153, opened 2020-01-20
tests: Adjust some test to check `decimal` exact value.

vpetrovykh:
A few tests that return `decimal` values now explicitly check the "json"
and "binary" results. The "binary" protocol is supposed to preserve type
information and actually produce a `decimal.Decimal` in Python. However,
fetching data as "json" can only produce a `float`.

Update the docs for the datatypes `bigint` and `decimal` with a note
about potential problems with JSON conversion.

Fixes: #1151.

### DIFF
--- a/edb/testbase/server.py
+++ b/edb/testbase/server.py
@@ -23,6 +23,7 @@ import asyncio
 import atexit
 import collections
 import contextlib
+import decimal
 import functools
 import inspect
 import json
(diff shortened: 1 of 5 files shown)


edb/testbase/server.py
```diff
@@ -483,6 +484,11 @@ def _assert_generic_shape(path, data, shape):
                     self.fail(
                         f'{message}: not isclose({data}, {shape}) '
                         f'{_format_path(path)}')
+            elif isinstance(shape, decimal.Decimal):
```
elprans: Why not simply add to the isinstance tuple below?

vpetrovykh: Done.

1st1:
@elprans this needs to be backported to a2

elprans:
> this needs to be backported to a2

Please use the new "needs-backport" label for this.

Issue #1145, opened 2020-01-18
Drop `from_python` from `BaseConstant`


vpetrovykh:
We use `from_python` mainly to convert `schema.Object.field` to `edgeql.AST.field`.  Since fields are typed anyway, we don't really need to have the dispatch in `BaseConstant`, we can just reimplement it in `AlterObjectProperty._get_ast` using the field type information.  GraphQL can do it's own thing.  We don't have constant folding other than string concatenation, so it's not an issue for that use case either.

Bottom line: the proper action is to kill `BaseConstant.from_python`.

_Originally posted by @elprans in https://github.com/edgedb/edgedb/pull/1131_

PR #1157, opened 2020-01-21
Refactor `BaseConstant.from_python`.

vpetrovykh:
Move the `from_python` dispatch into a utils function. Replace the usage
of the dispatch with calling `from_python` on specific classes where
possible.

Fixes: #1145.

### DIFF
--- a/edb/schema/utils.py
+++ b/edb/schema/utils.py
@@ -20,6 +20,7 @@
 from __future__ import annotations
 
 import collections
+import decimal
 import itertools
 from typing import *  # NoQA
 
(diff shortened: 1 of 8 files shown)


edb/edgeql/utils.py
```diff
@@ -119,3 +120,27 @@ def inline_anchors(
...
+MIN_INT64 = -2 ** 63
+
+
+def const_ast_from_python(val: Any) -> qlast.BaseConstant:
+    if isinstance(val, str):
+        return qlast.StringConstant.from_python(val)
+    elif isinstance(val, bool):
+        return qlast.BooleanConstant(value='true' if val else 'false')
+    elif isinstance(val, int):
+        if MIN_INT64 <= val <= MAX_INT64:
```
elprans: I thought we decided against this, no?

vpetrovykh: It's literally illegal then to create that `IntegerConstant` for cases outside the range. Why is it the only one that we're OK with messing up? The place where the utility function is used is generic, so technically there could be a large integer there.

vpetrovykh: I moved it to a different utils - in `schema`, closer to `delta.py`

edb/schema/utils.py
```diff
@@ -658,3 +659,27 @@ def _intersection_error(schema: s_schema.Schema,
...
+def const_ast_from_python(val: Any) -> qlast.BaseConstant:
+    if isinstance(val, str):
+        return qlast.StringConstant.from_python(val)
+    elif isinstance(val, bool):
+        return qlast.BooleanConstant(value='true' if val else 'false')
+    elif isinstance(val, int):
+        if MIN_INT64 <= val <= MAX_INT64:
+            return qlast.IntegerConstant(value=str(val))
+        else:
+            return qlast.BigintConstant(value=f'{val}n')
```
1st1: I think we should raise an error in this case.

vpetrovykh: OK, I've put a `ValueError` on that branch.

The GraphQL translator has its own ifs covering the int64/bigint case.

PR #1162, opened 2020-01-26
work in progress on annotating referencing.py

RobertoPrevato:
@elprans, sorry, I am stuck annotating `referencing` module. I am down to 25 problems from original 77, but there are a few parts that I don't understand.

1. What is the best way to address `"_create_innards" / "_delete_innards" undefined in superclass`, for example in `ReferencedObjectCommand`? I read how the class is used, and I understand it works by multi-inheritance, so for example in the case of "CreateConstraint" class, the super()._create_innards goes to the implementation in `CreateInheritingObject`. I don't know what is the best way to let mypy understand it, and make something that respects the original purpose of the code.
2. I have trouble understanding the `op` property that a `CommandContext` can have sometimes (set in `delta.Command._cmd_tree_from_ast` method). Is it an `Optional[Command]`?
3. The same for the `scls` property that a referrer context can have.

I would find it easier to understand these elements if there were more comments and possibly docstrings to describe some details (for example `op`, `scls`).

### DIFF
--- a/edb/schema/delta.py
+++ b/edb/schema/delta.py
@@ -487,6 +487,7 @@ class CommandContext:
         self.renamed_objs = set()
         self.altered_targets = set()
         self.schema_object_ids = schema_object_ids
+        self.op: Optional[Command] = None
 
     @property
     def modaliases(self) -> Mapping[Optional[str], str]:
@@ -768,7 +769,9 @@ class ObjectCommand(Command, metaclass=ObjectCommandMeta):
         if subnode is not None:
             node.commands.append(subnode)
 
-    def _get_ast_node(self, schema, context):
+    def _get_ast_node(self,
+                      schema: s_schema.Schema,
+                      context: CommandContext) -> qlast.ObjectDDL:
         return self.__class__.astnode
 
     def _get_ast(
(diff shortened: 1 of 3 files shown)


elprans:
@RobertoPrevato I'm working on annotating `delta.py`, and all of these will be sorted out.  You can put this on hold until I'm done.  Thanks!

1st1:
@elprans Are you still working on annotating delta.py?

elprans:
Yes. I'm almost done with that.

elprans:
I landed my typing PR, this should now be unblocked.

RobertoPrevato:
Hi @elprans, I still get the mypy complaints I described in my first comment, can you please help me? I am not sure if I miss something.

* I don't know what is the best way to handle _create_innards and _delete_innards since they are used as mixins and mypy complains about these methods being undefined in the superclass
* about `scls`, for example in `ReferencedObjectCommand._create_innards`, is "referrer_ctx" of type delta.CommandContext, and is its `scls` a schema.Type?

elprans:
>  don't know what is the best way to handle _create_innards and _delete_innards since they are used as mixins and mypy complains about these methods being undefined in the superclass

That doesn't seem to be correct. `ReferencedObjectCommand` inherits from `ObjectCommand` which defines `_create_innards`.  Perhaps you're not spelling the signature right? It must match that of the superclass.

> about scls, for example in ReferencedObjectCommand._create_innards, is "referrer_ctx" of type delta.CommandContext, and is its scls a schema.Type?

`referrer_ctx` is an `ObjectCommandContext[Object]`, and `scls` is just `Object`.

elprans:
> That doesn't seem to be correct.

Ah, sorry I was looking at the wrong class.  The correct fix would be to move the `_create_...` callbacks to the `CreateReferencedObject` class.

RobertoPrevato:
Closing this PR, since the work on annotations was started on a new branch after the work on delta.py.

PR #1169, opened 2020-01-30
Implement `JSON_ELEMENTS` IO format for responses

tailhook:
This is important for the non-interactive REPL mode, to allow fetching
large batches of the data without keeping everything in memory (and also
without replicating the whole JSONizing process on top of a binary
protocol).

I've tested this against patched Rust REPL. Need to implement similar feature in python bindings to have tests here.


docs/internals/protocol/messages.rst
```diff
@@ -303,8 +303,16 @@ Format:
...
         BINARY = 0x62,
-        JSON = 0x6a
+
+        // Returns a single row and single field that contains
+        // a resultset as a single JSON array
+        JSON = 0x6a,
+
+        // Returns a single JSON string per row, preferred over JSON format
+        // because might be used for larger responses
+        JSON_ROWS = 0x4a,
```
elprans: Perhaps `JSON_ELEMENTS` is a better name?  We don't refer to result entries as "rows" generally.  (Sorry, I know I suggested the "rows" name earlier).

edb/pgsql/compiler/output.py
```diff
@@ -525,5 +531,8 @@ def top_output_as_value(
 
         return stmt
 
+    elif env.output_format is context.OutputFormat.JSON_ROWS:
```
elprans: This branch seems redundant.

tailhook:
Added a test

tailhook:
@1st1, the test is here. I need your approval, as you have requested changes.

edb/server/compiler/compiler.py
```diff
@@ -95,11 +95,23 @@ class CompileContext:
...
 
 
 pg_ql = lambda o: pg_common.quote_literal(str(o))
 
 
+def _convert_format(inp: enums.IoFormat) -> pg_compiler.OutputFormat:
+    try:
+        return _IO_FORMAT_MAP[inp]
+    except KeyError:
+        raise RuntimeError(f"unimplemented io format {inp!r}")
```
1st1:
```suggestion
        raise RuntimeError(f"IO format {inp!r} is not supported")
```

PR #1178, opened 2020-02-03
Rust tokenizer

tailhook:
This issue supercedes #971 (latter was done diferent branch than master).


edb/common/parsing.py
```diff
@@ -97,9 +99,10 @@ def for_lex_token(mcls, token):
 
 
 class Token(parsing.Token, metaclass=TokenMeta):
-    def __init__(self, parser, val, context=None):
+    def __init__(self, parser, val, clean_value, context=None):
```
elprans: What is `clean_value` and how is it different cfrom `val`?

tests/test_edgeql_syntax.py
```diff
@@ -383,12 +390,12 @@ def test_edgeql_syntax_constants_34(self):
 
     def test_edgeql_syntax_constants_35(self):
         r"""
-        SELECT r"\n\w\d";
+        SELECT r'\n\w\d';
         """
 
     def test_edgeql_syntax_constants_36(self):
         r"""
-        SELECT $aa$\n\w\d$aa$;
```
elprans: This seems like a valid test, no?

tests/test_edgeql_syntax.py
```diff
@@ -383,12 +390,12 @@ def test_edgeql_syntax_constants_34(self):
 
     def test_edgeql_syntax_constants_35(self):
         r"""
-        SELECT r"\n\w\d";
```
elprans: Double quotes are valid for string literal, please keep the test as is.

tests/test_edgeql_syntax.py
```diff
@@ -398,16 +405,18 @@ def test_edgeql_syntax_constants_37(self):
 
     def test_edgeql_syntax_constants_38(self):
         r"""
-        SELECT "\n";
```
elprans: Ditto.

tests/test_edgeql_syntax.py
```diff
@@ -948,28 +957,28 @@ def test_edgeql_syntax_name_22(self):
         """
 
     @tb.must_fail(errors.EdgeQLSyntaxError,
-                  r'Identifiers cannot contain "::"', line=2, col=16)
+                  'backtick-quoted name can\'t contain `::`', line=2, col=16)
```
elprans: Please use "cannot" instead of "can't", we avoid the use of contractions in error messages.

tests/test_schema_syntax.py
```diff
@@ -358,6 +359,8 @@ def test_eschema_syntax_type_19(self):
         };
         """
 
+    @test.xfail('not sure how to test multi-line semantics, '
```
elprans: @vpetrovykh Any suggestions here?

edb/common/parsing.py
tailhook: `val` is token text and `clean_value` is the unescaped (for stings) and decoded value (for ints)

tests/test_edgeql_syntax.py
tailhook: Fixed

tests/test_edgeql_syntax.py
tailhook: Fixed

tests/test_edgeql_syntax.py
tailhook: Fixed

tests/test_edgeql_syntax.py
tailhook: Fixed

elprans:
@vpetrovykh Please review.

edb/common/parsing.py
vpetrovykh: As long as we don't decode float values, we should be fine.

vpetrovykh: As long as we don't decode float values, we should be fine.

tests/test_schema_syntax.py
vpetrovykh: Why is it `xfailed`? This should be testable even with strings that forget their original "structure" because they still have some actual value that makes it to the codegen and everything.

To give you some perspective of why the test is even there, the original SDL was indentation-sensitive (was using indentation instead of `{ ... }`) so multi-line strings could potentially screw up the indentation. The current SDL is no longer sensitive, so the test is far more trivial now and just ends up testing the use of multi-line strings in a realistic context.

edb/edgeql/parser/grammar/expressions.py
```diff
@@ -918,80 +917,16 @@ def reduce_NFCONST(self, *kids):
...
-
 class BaseStringConstant(Nonterm):
 
-    def reduce_EscapedStringConstant(self, *kids):
-        self.val = kids[0].val
-
-    def reduce_RawStringConstant(self, *kids):
-        self.val = kids[0].val
+    def reduce_SCONST(self, token):
+        self.val = qlast.RawStringConstant(value=token.clean_value)
```
vpetrovykh: If we're only using one AST node in the parser it should be `StringConstant` rather than `RawStringConstant`. In general, it appears that we should then remove the `RawStringConstant` altogether having only one type of AST node.

edb/edgeql/ast.py
```diff
@@ -266,22 +265,17 @@ def from_python(cls, val: typing.Any) -> BaseConstant:
...
 
 
 class RawStringConstant(BaseConstant):
-    quote: str
 
     @classmethod
     def from_python(cls, s: str):
-        value = quote.quote_literal(s)
-        return cls(value=value[1:-1], quote="'")
+        return cls(value=s)
```
vpetrovykh: This node should probably be dropped altogether as there's no more distinction between this and `StringConstant`.

edb/edgeql/codegen.py
```diff
@@ -545,13 +560,26 @@ def visit_Parameter(self, node: qlast.Parameter) -> None:
...
-            self.write('r', node.quote, node.value, node.quote)
+        if node.value.isprintable():
+            for d in ("'", '"', '$$'):
+                if d not in node.value:
+                    if '\\' in node.value and d != '$$':
+                        self.write('r', d, node.value, d)
+                    else:
+                        self.write(d, node.value, d)
+                    return
+        self.write(repr(node.value))
```
vpetrovykh: Drop this as.

edb/edgeql/codegen.py
```diff
@@ -545,13 +560,26 @@ def visit_Parameter(self, node: qlast.Parameter) -> None:
...
-        self.write(node.quote, node.value, node.quote)
+        if node.value.isprintable():
+            for d in ("'", '"', '$$'):
+                if d not in node.value:
+                    if '\\' in node.value and d != '$$':
+                        self.write('r', d, node.value, d)
+                    else:
+                        self.write(d, node.value, d)
+                    return
+        self.write(repr(node.value))
```
vpetrovykh: The idea is that we want codegen to be able to output human-friendly code (and at some point it will have different modes of operation: human-friendly, verbose, compact, etc.), which means that certain strings should be raw specifically because it makes it easier to read them. I like the rules for detecting quotes, but I would say that if a string contains a `$$` value we probably should assume that it's likely one raw (dollar) string inside another raw (dollar) string. So instead of potentially rendering it as a `repr` (thus escaping newlines and making it hard to read) we should render it as nested dollar-strings using `dollar_quote_literal` utility to make sure that our outer dollar-quotes aren't matched inside.

edb/edgeql/ast.py
vpetrovykh: This node should probably be dropped altogether as there's no more distinction between this and `StringConstant`.

edb/common/parsing.py
tailhook: Well, currently rust decodes float values too. But we in this change we don't use anything except strings (that's left for to do). So the question is not relevant for this PR, but still what's wrong with decoding floats?

tests/test_schema_syntax.py
tailhook: Fixed here: 9392b823 but I'm in doubt that the test makes sense (i.e. tests anything new). As far as I remember I've disabled it to discuss later.

tests/test_edgeql_expressions.py
```diff
@@ -3070,8 +3070,7 @@ class TestExpressions(tb.QueryTestCase):
     async def test_edgeql_expr_string_10(self):
         with self.assertRaisesRegex(
                 edgedb.QueryError,
-                r'invalid string literal: invalid line continuation',
-                _hint="newline has to immediately follow '\\'"):
+                r"invalid string literal: invalid escape sequence '\\ '",):
```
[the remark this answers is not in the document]
tailhook: Well, the hint isn't exactly well explaining for me. Backslash-space is as invalid escape sequence as any other. I would probably change it to something like `consider removing trailing whitespace`.

edb/edgeql/parser/grammar/expressions.py
tailhook: Done

edb/edgeql/codegen.py
tailhook: Done.

edb/edgeql/codegen.py
tailhook: Didn't know such function exists. But there is a problem with it:

Do we allow non-printable characters in the text? I.e. some ugly characters like backspace or escape? Or should dollar_quote_literal work only with printable characters too?

edb/edgeql/ast.py
tailhook: Done

tests/test_edgeql_expressions.py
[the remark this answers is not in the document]
tailhook: https://github.com/edgedb/edgedb/pull/1178/commits/5170881be123a412efcf5acd50763519496de7c0 -- is it clear enough for you?

edb/edgeql/codegen.py
tailhook: Added into `isprintable` branch but it doesn't seem to change much as `\n` is non-printable.

edb/common/parsing.py
vpetrovykh: Floats are often not representable exactly as a string, which means that the value provided by the user is going to be different from the factual value the float is decoded into. This presents two problems:
1) If you display that float value in an error message it may actually differ from the string the user supplied, which is inconvenient and potentially misleading.
2) If your decoder is different from the one Postgres uses then you'd reconstruct and submit the wrong value. Guaranteeing 100% that you're using the same decoder is probably impractical at this point.

Additionally, there's no reason to decode any numerical literals because we don't actually do any arithmetic on them and potentially this arithmetic may yield different results than Postgres.

We need to clean up the value string to remove the `_` from it, but that's purely a string operation.

edb/edgeql/codegen.py
vpetrovykh: Non-printable characters is a good point. So the first check should be for them, but there should also be a short white-list of non-printable characters that are OK (such as `\r`, `\n`, `\t` if it's for some reason considered non-printable). If the string contains any non-white-listed non-printable characters then it *must* be rendered as an escaped string (via the `repr`). The logic for the rest remains as outlined above.

tests/test_schema_syntax.py
vpetrovykh: It's OK even if it turns out to be redundant for this specific lexer and parser version. These syntax tests are fast and cheap enough. Also, this can come in handy when testing a parser re-write.

edb/common/parsing.py
tailhook: I think there are reversible algorithms for float to string conversion, but it's not for this PR anyway

edb/edgeql/codegen.py
tailhook: Tweaked a bit in b3459547 but let's not stuck in this loop indefinitely.

Fogapod:
It looks like rust compiler is required to compile edgedb after this commit.
Should it be included into this list now? https://github.com/edgedb/edgedb/blob/master/docs/internals/dev.rst#building-locally

PR #1179, opened 2020-02-03
Fix bug of bootstrapping server with EDGEDB_DEBUG_DELTA_EXECUTE

tailhook:
This was broken by a81203a6892303c8097b82437cbeec8916833c91

### DIFF
--- a/edb/server/compiler/compiler.py
+++ b/edb/server/compiler/compiler.py
@@ -520,7 +520,7 @@ class Compiler(BaseCompiler):
 
         if debug.flags.delta_execute:
             debug.header('Delta Script')
-            debug.dump_code('\n'.join(sql), lexer='sql')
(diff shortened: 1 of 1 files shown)


edb/server/compiler/compiler.py
```diff
@@ -520,7 +520,7 @@ def _compile_and_apply_ddl_command(self, ctx: CompileContext, cmd):
 
         if debug.flags.delta_execute:
             debug.header('Delta Script')
-            debug.dump_code('\n'.join(sql), lexer='sql')
+            debug.dump_code(b'\n'.join(sql), lexer='sql')
```
1st1: Don't you need to decode it? Or it works with bytes too?

tailhook: I've added a `b` (as Elivs suggested) and run bootstrap. The bug fixed. I'm not sure how all of that works actually.

PR #1192, opened 2020-02-05
Update the pygments grammar.

vpetrovykh:
Bring the pygments grammar up-to-date with the current keywords.

This affects the current REPL keyword highlighting.


1st1:
Maybe we should make this a separate package? @elprans

vpetrovykh:
I think that we use this in REPL and defining sphinx domains in docs, so since both of those are part of this repo it would be rather unintuitive to factor out the pygments package.

elprans:
There's value of making it a separate package: third-party projects.  It's useful to be able to highlight EdgeQL in Sphinx-generated docs.

vpetrovykh:
So I guess we need an issue for this:
> There's value of making it a separate package: third-party projects. It's useful to be able to highlight EdgeQL in Sphinx-generated docs.

PR #1196, opened 2020-02-06
Add `\li` commands to REPL.

### LINKED ISSUE #1195: Add a command for viewing indexes
The `\d+` command (and the corresponding `DESCRIBE`) should list the indexes for a type.

We also need a `\li` command to list all indexes, so that there's an overview of what indexes have been set up in the database already.

Related to #179

vpetrovykh:
`\li <pattern>` - list all indexes explicitly defined by the user [on
types matching the pattern].

The following options are avaialble:
* `I` - make the pattern match case-sensitive
* `S` - add indexes defined on system objects
* `+` - verbose mode shows implicitly defined indexes (coming from
        exclusive constraints)

Note that so far we only have implicit indexes on system objects, so `S`
without a `+` doesn't do anything.

Fixes #1195

### DIFF
--- a/edb/repl/__init__.py
+++ b/edb/repl/__init__.py
@@ -233,6 +233,20 @@ class Cli:
                     callback=self._command_list_casts,
                 ),
 
+                cmd.Command(
+                    trigger='li',
+                    desc='list indexes',
+                    group='Introspection',
+                    arg_name='PATTERN',
+                    arg_optional=True,
+                    flags={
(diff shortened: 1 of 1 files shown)


vpetrovykh:
The introspection query used there is actually an interesting showcase of EdgeQL's expressive power. Using union, type filtering, computables and re-shaping shapes to ultimately get the different objects into a uniform result table. Doing this in hand-written SQL would be a nightmare.

PR #1197, opened 2020-02-06
Use postgres/notify to invalidate query cache on DDL

1st1:
This makes it possible to use different EdgeDB servers connected
to the same Postgres cluster.  With this change, when a DDL command
is committed, a `NOTIFY` command is issued which all EdgeDB
connections listen to.

### DIFF
--- a/edb/server/pgcon/pgcon.pyx
+++ b/edb/server/pgcon/pgcon.pyx
@@ -22,6 +22,7 @@ import codecs
 import hashlib
 import json
 import os.path
+import weakref
 
 cimport cython
 cimport cpython
@@ -91,6 +92,7 @@ def _build_init_con_script() -> bytes:
             ('', {pg_ql(defines.DEFAULT_MODULE_ALIAS)}, 'A'),
             ('server_version', {pg_ql(buildmeta.get_version_json())}, 'R');
 
+        LISTEN __edgedb_ddl__;
     ''').encode('utf-8')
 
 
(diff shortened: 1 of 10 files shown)


1st1:
I'm unsure how to test this though. We should probably have a test for this in our cloud infra project.

elprans:
We can test this locally by starting a second server with —postgres-dsn of the first one.

edb/server/pgcon/pgcon.pyx
```diff
@@ -929,7 +941,18 @@ cdef class PGProto:
 
         elif mtype == b'A':
             # NotificationResponse
-            self.buffer.discard_message()
+            self.buffer.read_int32()  # discard pid
+            channel = self.buffer.read_null_str().decode()
+            payload = self.buffer.read_null_str().decode()
+            self.buffer.finish_message()
+
+            if channel == '__edgedb_ddl__':
```
ambv: We should log if an unsupported notification arrives. EdgeDB users cannot access Postgres directly so they won't be using those themselves, no? And even if they did, we should at least log if the channel starts with `__edgedb_`.

ambv:
I'll let Elvis approve this but it looks pretty clean. Postgres is a great base for this.

edb/server/pgcon/pgcon.pyx
1st1: > We should log if an unsupported notification arrives.

They shouldn't. We explicitly listen to the `__edgedb_ddl__` channel. But I'll add logging anyways just in case and for future debug.

> EdgeDB users cannot access Postgres directly so they won't be using those themselves, no?

No, we don't expose this yet.

1st1:
@elprans Please review again.

tests/test_server_proto.py
```diff
@@ -1888,6 +1892,138 @@ def on_log(con, msg):
...
+
+            INSERT Test { foo := 123 };
+        ''')
+
+        self.assertEqual(
+            await self.con.fetchone('SELECT Test.foo LIMIT 1'),
+            123
+        )
+
+        with tempfile.TemporaryDirectory() as tmp:
```
elprans: Might be worth factoring this out into `cluster.py` or `testbase` once we start adding other tests like this.

1st1: Yeah, as soon as we need a second test like this I'll add a reusable harness.

PR #1200, opened 2020-02-06
Fix a minor REPL bug with commented out lines.

vpetrovykh:
REPL should clean out the comments when breaking down the query text.
Otherwise it's possible to make REPL submit an empty query (type up a
comment and hit Alt+Enter) and cause a ProtocolError by not having an
actual query.


tailhook:
Sure. There is a test for that: https://github.com/edgedb/edgedb-rust/blob/master/edgeql-parser/tests/preparser.rs#L38-L43

Issue #1213, opened 2020-02-11
Add the "current_database()" to our builtins


vpetrovykh:
Postgres has a nice `current_database()` function that returns the name of the database you're connected to. This is a nice utility function to have. The proposal is to map it onto `sys::get_current_database()`.

The main reason I ran into this was that I was working on a fix to handle "cannot drop the currently open database" error and realized that the easiest way to write a test for this would be to actually ask the name of the test database first and then attempt to delete it, rather then set up a whole different test where I control the test database name.

1st1:
We also need current user. I see to options:

* add a `sys::get_session_info() -> tuple<user: str, database: str>` function. We can add more fields to the return tuple in future releases.

* have individual functions for every functionality bit -- `sys::get_current_database()`, `sys::get_current_user()`, etc.

The latter option seems like a simpler one to implement and manage, so I like it better.

elprans:
+1 for the separate functions

vpetrovykh:
Should it be `sys::get_current_user()` or `sys::get_current_role()`?

1st1:
Role, but you can't map that function to anything though (we can't use postgres for that). I'll add the current role myself.

elprans:
> but you can't map that function to anything though (we can't use postgres for that)

Why not?  We have a 1:1 mapping between EdgeDB roles and Postgres roles.

vpetrovykh:
> Role, but you can't map that function to anything though (we can't use postgres for that). I'll add the current role myself.

Yeah, I'm confused as to what the problem with having a function with SQL body `SELECT current_user;` is?

elprans:
> I'm confused as to what the problem

It wouldn't work currently, because we always connect to postgres as "edgedb" currently.  We need to fix this by switching the authorization context when EdgeDB picks a backend connection.  Is `current_user` blocking something?  If not, I'd postpone this.

elprans:
Closing this, since the subject is done, and `current_user` depends on a functional role system, which we don't have yet.

PR #1215, opened 2020-02-11
Add `get_current_database` function to `sys` module.

vpetrovykh:
Reflect Postgres' `current_database()` into
`sys::get_current_database()`.

Issue #1213.

### DIFF
--- a/edb/lib/sys.edgeql
+++ b/edb/lib/sys.edgeql
@@ -187,3 +187,12 @@ sys::get_transaction_isolation() -> sys::TransactionIsolation
     SET force_return_cast := true;
     USING SQL FUNCTION 'edgedb._get_transaction_isolation';
 };
+
+
+CREATE FUNCTION
+sys::get_current_database() -> str
+{
+    # The results won't change within a single statement.
+    SET volatility := 'STABLE';
+    USING SQL FUNCTION 'current_database';
+};
--- a/tests/test_database.py
+++ b/tests/test_database.py
@@ -23,7 +23,7 @@ from edb.testbase import server as tb
 
 
 class TestDatabase(tb.ConnectedTestCase):
-    async def test_database_create01(self):
+    async def test_database_create_01(self):
         await self.con.execute('CREATE DATABASE mytestdb;')
 
         try:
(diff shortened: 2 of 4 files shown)


edb/lib/sys.edgeql
```diff
@@ -187,3 +187,12 @@ sys::get_transaction_isolation() -> sys::TransactionIsolation
     SET force_return_cast := true;
     USING SQL FUNCTION 'edgedb._get_transaction_isolation';
 };
+
+
+CREATE FUNCTION
+sys::get_current_database() -> str
+{
+    # The results won't change within a single statement.
+    SET volatility := 'STABLE';
```
1st1: The results won't change ever. You might as well say it's immutable.

vpetrovykh: Is it possible to potentially rename a database? Same question for the role.

vpetrovykh: My understanding was that `IMMUTABLE` actually means universally the same, like arithmetic, for example. So same output across databases and across all time.

vpetrovykh:
I added the docs update to the commit.

edb/lib/sys.edgeql
elprans: The underlying function is marked as stable, and so must be the wrapper.

1st1: Right.

1st1:
Do we need the word `current` in the name?

vpetrovykh:
> Do we need the word `current` in the name?

I think so. There are potentially more databases that can be seen via, `SELECT sys::Database{name};`. So it seems that some qualifier is needed and "current" is fairly self-explanatory.

1st1:
OK, let's keep the verbose name. It's not going to be frequently used anyways (which is another + for the verbose version)

tests/test_database.py
```diff
@@ -42,3 +42,16 @@ class TestDatabase(tb.ConnectedTestCase):
             await conn.aclose()
         finally:
             await self.con.execute('DROP DATABASE mytestdb;')
+
+    async def test_database_current_database_01(self):
+        await self.con.execute('CREATE DATABASE currentdb;')
```
1st1: Wait, this is a bit too slow. Just add this test to test_database_create_01

vpetrovykh: Errr, sorry, I just merged. I can make a quick PR with tests consolidation.

Issue #1221, opened 2020-02-12
What are the minimum hardware requirements?


haikyuu:
- EdgeDB Version: Alpha 2
- OS Version: Ubuntu 18

Steps to Reproduce:

1. Following along with the tutorial (queries work just fine)
2. When i Expose GraphQL port, it took a while and i had to `ctrl+c` it.
`== aborting query and closing the connection ==`
After exposing another port, i get this error:

```
+ OSError: [Errno 12] Cannot allocate memory
 |   File "/usr/lib/x86_64-linux-gnu/edgedb-1-alpha2/lib/python3.8/site-packages/edb/server/procpool/pool.py", line 191, in _spawn_for_pool
 |     worker = await self._spawn_worker()
 |   File "/usr/lib/x86_64-linux-gnu/edgedb-1-alpha2/lib/python3.8/site-packages/edb/server/procpool/pool.py", line 187, in _spawn_worker
 |     await worker._spawn()
 |   File "/usr/lib/x86_64-linux-gnu/edgedb-1-alpha2/lib/python3.8/site-packages/edb/server/procpool/pool.py", line 85, in _spawn
 |     self._proc = await asyncio.create_subprocess_exec(
 |   File "/usr/lib/x86_64-linux-gnu/edgedb-1-alpha2/lib/python3.8/asyncio/subprocess.py", line 236, in create_subprocess_exec
 |     transport, protocol = await loop.subprocess_exec(
 |   File "uvloop/loop.pyx", line 2749, in subprocess_exec
 |   File "uvloop/loop.pyx", line 2707, in __subprocess_run
 |   File "uvloop/handles/process.pyx", line 596, in uvloop.loop.UVProcessTransport.new
 |   File "uvloop/handles/process.pyx", line 98, in uvloop.loop.UVProcess._init
```
repeated three times.

When i went to Digital Ocean droplet Graphs, i saw that CPU usage was 100%
![image](https://user-images.githubusercontent.com/8558836/74332971-40c0d080-4d97-11ea-81ef-4f67bfa938a9.png)

Note i was using the minimal Droplet 
![image](https://user-images.githubusercontent.com/8558836/74333068-7960aa00-4d97-11ea-9314-1a4c9cb984e3.png)

haikyuu:
Upgrading the droplet solves the issue:
![image](https://user-images.githubusercontent.com/8558836/74333454-466ae600-4d98-11ea-98b8-e1776f8c5866.png)

This one works seamlessly

[...]

PR #1222, opened 2020-02-12
Update Graphql section in tutorial - concurrency 1

haikyuu:
This will make the tutorial work on machines with less than `1GB` of RAM. Addresses #1221


1st1:
Thank you!

PR #1224, opened 2020-02-16
Annotations for schema referencing

RobertoPrevato:


edb/schema/referencing.py
```diff
@@ -38,12 +38,26 @@
...
 
 
+if TYPE_CHECKING:
+    from edb.schema import inheriting as s_inheriting
+
+
 class ReferencedObject(so.Object, derivable.DerivableObjectBase):
 
-    def get_referrer(self, schema):
+    def get_referrer(self, schema: s_schema.Schema) -> Optional[so.Object]:
```
RobertoPrevato: This is my biggest doubt of my current PR. 

In practice, the `ReferencedObject` class is inherited by classes that all define a `subject` field: therefore they all have a `get_subject` method independently from each other (s_annos.AnnotationValue, s_indexes.Index, s_constraints.Constraint).

In this kind of situation, I would define the `get_subject` as an `@abstractmethod` on the `ReferencedObject` class, to make it explicit that implementers declare such method - and make both humans reading the code and mypy happy.

Since the `get_subject` method is generated dynamically from the `subject = SchemaField`, I think abstractmethod doesn't fit here. I also noticed that you don't use abstractmethod, so I think you don't like that pattern.

I also considered specifying a class property:

```
get_subject: Callable[[ReferencedObject, s_schema.Schema], Optional[so.Object]]
```

But this looks confusing, too, and misleading for humans who read the code. Asserting that self is an instance of the classes having a subject, for the method that requires `get_subject`, violates the Single Responsibility Principle but looks like the clearest solution.

edb/schema/referencing.py
```diff
@@ -245,47 +279,66 @@ def _classname_from_ast(cls, schema, astnode, context):
...
+                      schema: s_schema.Schema,
+                      context: sd.CommandContext
+                      ) -> Type[qlast.DDLOperation]:
         subject_ctx = self.get_referrer_context(context)
         ref_astnode = getattr(self, 'referenced_astnode', None)
         if subject_ctx is not None and ref_astnode is not None:
-            return ref_astnode
+            # type ignore below, because asserting issubclass doesn't work
+            # in the current stable version of mypy
+            return ref_astnode  # type: ignore
```
RobertoPrevato: My understanding is that:

```
assert issubclass(ref_astnode, qlast.DDLOperation)
```

Should make mypy happy. Because if you assert that _A_ is a subclass of _B_, you assert that A is a type of B ("Type[B]").

Instead, mypy still considers `ref_astnode` to be `Any`.

---
Another solution would be to do:

```python
        ref_astnode: Type[qlast.DDLOperation] = getattr(self,
                                                        'referenced_astnode',
                                                        None)
        if subject_ctx is not None and ref_astnode is not None:
            return ref_astnode
```

But it doesn't give any concrete value; since `getattr` can return anything.

edb/schema/referencing.py
elprans: A regular method raising a `NotImplementedError` should suffice.  I agree that the implementation should be moved to the classes that actually declare `subject`.

edb/schema/referencing.py
```diff
@@ -643,13 +757,21 @@ def _validate(self, schema, context):
...
         rec = context.current().enable_recursion
         context.current().enable_recursion = False
-
         referrer_ctx = self.get_referrer_context(context)
-        referrer = referrer_ctx.scls
+
+        # type ignore below, because mypy doesn't trust referrer_ctx to be an
+        # instance of sd.ObjectCommandContext (considers it <nothing>)
+        assert isinstance(referrer_ctx, sd.ObjectCommandContext)
+        referrer = referrer_ctx.scls  # type: ignore
```
RobertoPrevato: In this case, and another identical one, I used "type: ignore" because, while asserting that referrer_ctx is an instance of sd.ObjectCommandContext (or sd.ObjectCommandContext[so.Object]) should be sufficient, mypy still complains that "<nothing> as no property `scls`".

sd.ObjectCommandContext defines a property scls as generic "Object_T" which is bound to so.Object.

To me, this looks like a limitation in mypy, it looks like it gets confused for the generic class.

edb/schema/referencing.py
```diff
@@ -108,8 +123,9 @@ def derive_ref(
         existing = refcoll.get(schema, refname, default=None)
 
         if existing is not None:
-            cmdcls = sd.ObjectCommandMeta.get_command_class_or_die(
-                sd.AlterObject, type(self))
+            cmdcls: Type[sd.Command] = \
```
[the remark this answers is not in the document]
elprans: Also, the correct annotation would likely be `ObjectCommand`

edb/schema/referencing.py
```diff
@@ -164,15 +182,16 @@ def derive_ref(
...
+        derived: ReferencedT = schema.get(derived_name)
 
         return schema, derived
 
 
 class ReferencedInheritingObject(inheriting.InheritingObject,
                                  ReferencedObject):
-
-    def get_implicit_bases(self, schema):
+    def get_implicit_bases(self,
```
elprans: The return type would be the same as `self`, so making this generic (ie. `self: ReferencedInheritingObjectT`) would improve the ergonomics of usage of this method.

edb/schema/referencing.py
```diff
@@ -181,20 +200,30 @@ def get_implicit_bases(self, schema):
 
 class ReferencedObjectCommandMeta(sd.ObjectCommandMeta):
     _transparent_adapter_subclass: ClassVar[bool] = True
-
-    def __new__(mcls, name, bases, clsdct, *,
-                referrer_context_class=None, **kwargs):
+    _referrer_context_class: Optional[
```
elprans: This is a class attribute, so a `ClassVar` is needed.

edb/schema/referencing.py
RobertoPrevato: Super :blush: in this case, I try to add a method raising `NotImplementedError` to the class, and I see if all tests pass. I am correcting another point right now.

edb/schema/referencing.py
```diff
@@ -546,8 +640,9 @@ def _build_alter_cmd_stack(self, schema, context, scls, *, referrer=None):
             else:
                 obj = None
 
-        cmd = delta
+        cmd: Union[sd.DeltaRoot, sd.AlterObject[Any]] = delta
```
elprans: Just `sd.Command` would suffice, since that's what you've annotated the return type to be.

edb/schema/referencing.py
```diff
@@ -546,8 +640,9 @@ def _build_alter_cmd_stack(self, schema, context, scls, *, referrer=None):
...
         for obj in reversed(object_stack):
+            assert obj is not None
```
elprans: Nitpick: `assert` after `get_referrer` instead would be cleaner.

edb/schema/referencing.py
elprans: Yes, that's fine.  The comment explaining why `type: ignore` is needed is sufficient.

edb/schema/referencing.py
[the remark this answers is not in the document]
RobertoPrevato: Without this annotation, mypy tells: _"Argument 1 to "get_command_class_or_die" of "ObjectCommandMeta" has incompatible type "Type[CreateObject[Any]]"; expected "Type[AlterObject[Any]]"_

when the cmdcls is not annotated. This warning from mypy is similar to the one inside `_propagate_ref_deletion` method. I noticed that these warnings didn't happen when I started working on annotating referencing, before the refactoring / annotation of delta.py.

edb/schema/referencing.py
RobertoPrevato: I just tried to add `ClassVar`, a new error is reported by mypy when this property is set inside the `__new__` method, close to its definition.
`Cannot assign to class variable "_referrer_context_class" via instance`.

However, since `__new__` there belongs to a metaclass, shouldn't mypy tolerate setting a class attribute here?

edb/schema/delta.py
```diff
@@ -642,6 +643,7 @@ def __init__(
         self.renamed_objs: Set[so.Object] = set()
         self.altered_targets: Set[so.Object] = set()
         self.schema_object_ids = schema_object_ids
+        self.op: Optional[Command] = None
```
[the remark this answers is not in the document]
RobertoPrevato: Indeed - this was debris from my previous attempts (before the chat we had on Thursday evening).

edb/schema/referencing.py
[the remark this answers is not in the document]
RobertoPrevato: Also, changing the type annotation like this:
```python
cmdcls: sd.ObjectCommand[so.Object]
```

Results in new errors:

image

edb/schema/referencing.py
```diff
@@ -124,6 +140,7 @@ def derive_ref(
             old_bases = existing.get_bases(schema)
 
             if new_bases != old_bases:
+                assert isinstance(new_bases, so.ObjectList)
```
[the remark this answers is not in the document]
RobertoPrevato: This doesn't seem to work. Even if I move the declaration of new_bases, even if I specify its type at line 114; I still need this assertion to make mypy happy.

Otherwise it says: _"object" has no attribute "objects"_

edb/schema/referencing.py
RobertoPrevato: It looks like mypy is not checking so deeply. If I remove this assertion and move it after `get_referrer`, after the various operations; mypy complains that obj in the new iteration could be None.

In the screenshot below, note that the assertion is commented (it is placed after the call to get_referrer)

image

edb/schema/referencing.py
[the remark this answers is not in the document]
elprans: OK, fine then.

edb/schema/referencing.py
elprans: Ah, sorry, I missed that this is a metaclass. No `ClassVar` then.

edb/schema/referencing.py
elprans: OK

1st1:
LGTM! Please squash before merging.

PR #1234, opened 2020-02-18
edb.schema cleanups (part 1)

elprans:
This is a series of commits with cleanups and refactorings
that are necessary for the schema reflection/introspection
rewrite.


edb/schema/delta.py
```diff
@@ -1197,43 +1177,49 @@ def compute_inherited_fields(
...
-            val = self._resolve_attr_value(
-                op.new_value, op.property, field, schema)
+            value = self._resolve_attr_value(
+                raw_value, attr_name, field, schema)
 
-            if isinstance(val, s_expr.Expression) and not val.is_compiled():
-                val = self.compile_expr_field(schema, context, field, val)
+            if (isinstance(value, s_expr.Expression)
+                    and not value.is_compiled()):
+                value = self.compile_expr_field(schema, context, field, value)
```
1st1: Is it safe to cache compiled expressions?

elprans: Yes, in a sense that both EdgeQL AST and the set of references in the expression are immutable.  IR AST is not safe to cache in general, but it's not being stored currently.

PR #1235, opened 2020-02-18
Fix tuple protocol in protocol version 0.8

### LINKED ISSUE #1220: Protocol inconsistency in handling tuples
# Issue

Tuple codec for output from edgedb is basically this (sans error handling):
```
        elem_count = <Py_ssize_t><uint32_t>hton.unpack_int32(frb_read(buf, 4))
        for i in range(elem_count):
            frb_read(buf, 4)  # reserved
            elem_len = hton.unpack_int32(frb_read(buf, 4))
            elem_codec = <BaseCodec>self.fields_codecs[i]
            elem = elem_codec.decode(frb_slice_from(&elem_buf, buf, elem_len))
```
And tuple codec for input to edgedb is:
```
        elem_data = WriteBuffer.new()
        for i in range(objlen):
            sub_codec = <BaseCodec>(self.fields_codecs[i])
            sub_codec.encode(elem_data, item)

        buf.write_int32(4 + elem_data.len())  # buffer length
        buf.write_int32(<int32_t><uint32_t>objlen)
        buf.write_buffer(elem_data)
```

I.e. it has **omitted reserved bytes**.

While descriptors look equivalent:
```
InputDescr [
    BaseScalar(
        BaseScalarTypeDescriptor {
            id: 00000000-0000-0000-0000-000000000101,
        },
    ),
    Tuple(
        TupleTypeDescriptor {
            id: 19b56b7d-56b4-0004-38f0-fef54e18d8e2,
            element_types: [
                TypePos(
                    0,
                ),
            ],
        },
    ),
]
Output Descr [
    BaseScalar(
        BaseScalarTypeDescriptor {
            id: 00000000-0000-0000-0000-000000000101,
        },
    ),
    Tuple(
        TupleTypeDescriptor {
            id: 19b56b7d-56b4-0004-38f0-fef54e18d8e2,
            element_types: [
                TypePos(
                    0,
                ),
            ],
        },
    ),
]
```

**Update:** same issue with `NamedTuple`

# Possible Fixes

1. Document current behavior
2. Drop old tuple type descriptor and bump major protocol version

Since this is a server-sent descriptor I don't see a way to deprecate this without major protocol upgrade. The protocol negotiation could be used to revert gracefully though.

tailhook:
Fixes #1220


edb/server/mng_port/edgecon.pyx
```diff
@@ -310,13 +315,24 @@ cdef class EdgeConnection:
 
         self.buffer.finish_message()
 
-        if major != PROTO_VER_MAJOR or minor != PROTO_VER_MINOR or nexts > 0:
+        self.protocol_version = major, minor
```
elprans: Also need to fix the version check in dump `restore()` in this file.

edb/server/main.py
```diff
@@ -491,6 +511,10 @@ def _get_runstate_dir_default() -> str:
         '--auto-shutdown', type=bool, default=False, is_flag=True,
         help='shutdown the server after the last management ' +
              'connection is closed'),
+    click.option(
+        '--max-protocol', type=str, callback=_protocol_version,
```
1st1: Perhaps I missed something, but why do we need this flag?

elprans: It was my suggestion initially.  We need this for the ability to test protocol downgrade in clients, but since we agreed that we'd be dropping support for older protocols in clients for now, this isn't strictly necessary now.

PR #1238, opened 2020-02-18
Generalize ParametricType and use it for ObjectCollections

elprans:
`ObjectCollection` subclasses are fairly similar to the checked
collections found in `common.checked`, and the reflection infrastructure
could benefit from a more specific collection type specification in
fields.

This also adds limited support for forward references in type arguments.

### DIFF
--- a/edb/common/parametric.py
+++ b/edb/common/parametric.py
@@ -1,0 +1,159 @@
+#
+# This source file is part of the EdgeDB open source project.
+#
+# Copyright 2011-present MagicStack Inc. and the EdgeDB authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from __future__ import annotations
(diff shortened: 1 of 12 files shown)


edb/common/parametric.py
```diff
@@ -0,0 +1,159 @@
...
+    @classmethod
+    def is_fully_resolved(cls) -> bool:
+        return not cls._forward_refs
+
+    @classmethod
+    def resolve_types(cls, globalns: Dict[str, Any]) -> None:
+        if cls.types is None:
+            raise TypeError(
+                f"{cls!r} is not parametrized"
+            )
```
ambv: super nit: pretty sure this fits in one line, no?

edb/schema/objects.py
```diff
@@ -1764,8 +1811,11 @@ def _reduce_to_ref(  # type: ignore
 class ObjectIndexBase(ObjectCollection, container=tuple):
     _key: KeyFunction
 
-    def __init_subclass__(cls, *, key: KeyFunction):
-        cls._key = key
+    def __init_subclass__(cls, *, key: Optional[KeyFunction] = None) -> None:
+        if key is not None:
+            cls._key = key
+        elif cls._key is None:
+            raise TypeError('missing required "key" class argument')
```
ambv: Not clear to me why this is necessary.

ambv:
Generic ObjectList is 🔥🔥🔥

edb/schema/objects.py
elprans: ObjectIndex must define a `key` function, but once it's defined, subclasses don't need to redefine it.

elprans: This includes parametrized subclasses, obviously.

ambv: Oh, of course, I didn't read it properly myself. Thanks!

Issue #1175, opened 2020-02-01
Tuple/ GraphQL bug


agritheory:
- EdgeDB Version: alpha2
- OS Version: Ubunut 18.04

```
InternalServerError: unhandled errors in a TaskGroup; 1 sub errors: (AttributeError)
 + AttributeError: 'Tuple' object has no attribute 'get_topmost_concrete_base'
 |   File "/usr/lib/x86_64-linux-gnu/edgedb-1-alpha2/lib/python3.8/site-packages/edb/server/baseport.py", line 66, in new_compiler
 |     await compiler_worker.call('connect', dbname, dbver)
 |   File "/usr/lib/x86_64-linux-gnu/edgedb-1-alpha2/lib/python3.8/site-packages/edb/server/procpool/pool.py", line 120, in call
 |     raise exc
```

Steps to Reproduce:
Schema is below; I'm trying to set up graphql on a database with a tuple scalar... which where I think this error is coming from.

Schema (Tuples)
```
SCALAR TYPE root_account_type_enum EXTENDING enum<'Asset', 'Liability', 'Equity', 'Income', 'Expense'>;
        SCALAR TYPE report_type_enum EXTENDING enum<'BalanceSheet', 'ProfitAndLoss'>;
```
Schema (Object)
```
      type Account {
        required property name -> str;
        required property account_number -> str;
        required link company -> Company;
        required link account_currency -> Currency;
        link parent_account -> Account;
        required property root_account_type -> root_account_type_enum;
        required property report_type -> report_type_enum;
        property has_subledger -> bool;
        link children_accounts := .<parent_account[IS Account];
        index on (.name) {
            annotation indextitle := 'Account Index';
        }
      }
```

ngryman:
I think I'm facing a similar issue when using `tuple` or `enum`. However, I have two distinct error messages.

I'm using the latest Docker image:
* EdgeDB Version: `1-alpha2`
* OS Version: `Debian GNU/Linux 9 (stretch)`

## Tuple

Given the following SDL:
```perl
type Task {
  property recurrence -> tuple<int16, int16, int16, int16, int16>;
}
```

I have the exact same error that is reported here when I try to enable GraphQL.

## Enum

When using an `enum` with GraphQL already enabled:
```perl
scalar type FieldType extending enum<'text', 'numeric', 'boolean'>;

type Field {
  required property type -> FieldType;
}
```

I have a different error in my GraphQL explorer:
```
{
  "errors": [
    {
[...]

PR #1253, opened 2020-02-21
graphql: Add an explicit error for reflecting tuples.

vpetrovykh:
Currently tuples are not supported in GraphQL. The GraphQL documentation
is also updated to reflect this fact.

Issue: #1175.

### DIFF
--- a/docs/clients/99_graphql/index.rst
+++ b/docs/clients/99_graphql/index.rst
@@ -36,6 +36,19 @@ Pointing your browser to ``http://127.0.0.1:8888/explore``
 will bring up a `GraphiQL`_ interface to EdgeDB. This interface can be
 used to try out queries and explore the GraphQL capabilities.
 
+
+Known Limitations
+=================
+
+- Due to the differences between EdgeQL and GraphQL syntax
+  :eql:type:`enum <std::enum>` types which have values that cannot be
+  represented as GraphQL identifiers (e.g. ``'N/A'`` or ``'NOT
+  APPLICABLE'``) cannot be properly reflected into GraphQL enums.
+
+- EdgeDB :eql:type:`tuples <std::tuple>` are not supported in GraphQL
+  reflection currently.
+
+
 .. __: http://graphql.org/docs/queries/
 
 .. _`GraphiQL`: https://github.com/graphql/graphiql
--- a/edb/graphql/types.py
+++ b/edb/graphql/types.py
@@ -299,6 +299,11 @@ class GQLCoreSchema:
             if name in self._gql_enums:
                 target = self._gql_enums.get(name)
 
+        elif edb_target.is_tuple():
+            edb_typename = edb_target.get_verbosename(self.edb_schema)
+            raise g_errors.GraphQLCoreError(
+                f"Could not convert {edb_typename} to a GraphQL type.")
+
         else:
             base_target = edb_target.get_topmost_concrete_base(self.edb_schema)
             bt_name = base_target.get_name(self.edb_schema)


docs/clients/99_graphql/index.rst
```diff
@@ -36,6 +36,18 @@ Pointing your browser to ``http://127.0.0.1:8888/explore``
 will bring up a `GraphiQL`_ interface to EdgeDB. This interface can be
 used to try out queries and explore the GraphQL capabilities.
 
+
+Known Limitations
+=================
+
+- Due to the differences between EdgeQL and GraphQL syntax some
+  :eql:type:`enum <std::enum>` types cannot be properly reflected.
```
1st1: Can you clarify this? What enum types? All of them?

docs/clients/99_graphql/index.rst
```diff
@@ -36,6 +36,18 @@ Pointing your browser to ``http://127.0.0.1:8888/explore``
...
+
+- EdgeDB :eql:type:`tuples <std::tuple>` are not supported in GraphQL
+  reflection currently (due to the fact that GraphQL Lists are
+  supposed to be homogeneous and tuples are heterogeneous).
```
1st1: Tuples are not lists, they are closer to objects, although lack ids. I suggest to drop the `currently (due .*)` part.

edb/graphql/types.py
```diff
@@ -299,6 +299,14 @@ def _convert_edb_type(self, edb_target):
             if name in self._gql_enums:
                 target = self._gql_enums.get(name)
 
+        elif edb_target.is_tuple():
+            edb_typename = edb_target.get_verbosename(self.edb_schema)
+            raise g_errors.GraphQLCoreError(
+                f"Could not convert {edb_typename} to"
+                f" a GraphQL type. "
```
1st1: Again, I'd just say that tuples are not currently supported without elaborating.

docs/clients/99_graphql/index.rst
vpetrovykh: Well, an unnamed tuple is just an ordered heterogeneous list. In fact, every tuple is defined by its element types and ordering and not by the optionally provided names. The point is that names are completely secondary in tuples unlike in object types. They are more like heterogeneous lists with a secondary optional indexing method by names.

Consider this: 
```
db> SELECT (1, 2) = (a := 1, b:= 2);
{true}
db> SELECT (b:= 2, a := 1) = (a := 1, b:= 2);
{false}
```

edb/graphql/types.py
vpetrovykh: Fair enough to remove the extra details from this message.

docs/clients/99_graphql/index.rst
vpetrovykh: It's covered in the linked enum description, but I can include it here as well.

docs/clients/99_graphql/index.rst
1st1: > Well, an unnamed tuple is just an ordered heterogeneous list.

In languages that have "heterogeneous lists". Tuples are also frequently referred to as records. I'm not sure why we're having this debate though; for the user-facing documentation you just want to say in simple words what's supported and what's not without diving into too much detail/cs terminology.

docs/clients/99_graphql/index.rst
1st1: You say "some enum types" -- what does "some" mean? Does the linked description explains what "some" mean in this sentence? Sorry, I just really can't parse this and hence potentially nitpicking.

docs/clients/99_graphql/index.rst
```diff
@@ -36,6 +36,20 @@ Pointing your browser to ``http://127.0.0.1:8888/explore``
...
 used to try out queries and explore the GraphQL capabilities.
 
+
+Known Limitations
+=================
+
+- Due to the differences between EdgeQL and GraphQL syntax
+  :eql:type:`enum <std::enum>` types which have values that use
+  symbols not found in identifiers (e.g. ``'N/A'`` or ``'NOT
+  APPLICABLE'``) cannot be properly reflected into GraphQL enums.
```
1st1: Can we use GraphQL strings for EgdeQL enums?

docs/clients/99_graphql/index.rst
```diff
@@ -36,6 +36,20 @@ Pointing your browser to ``http://127.0.0.1:8888/explore``
 will bring up a `GraphiQL`_ interface to EdgeDB. This interface can be
 used to try out queries and explore the GraphQL capabilities.
 
+
+Known Limitations
+=================
+
+- Due to the differences between EdgeQL and GraphQL syntax
+  :eql:type:`enum <std::enum>` types which have values that use
+  symbols not found in identifiers (e.g. ``'N/A'`` or ``'NOT
```
1st1: "types which have values that use symbols not found in identifiers" -> 'enum types that have values that cannot be represented as GraphQL identifiers (e.g. "N/A")'

docs/clients/99_graphql/index.rst
vpetrovykh: Maybe I'm overthinking it here. I'll drop the "due..." part.

docs/clients/99_graphql/index.rst
vpetrovykh: Fair enough. I'll update.

docs/clients/99_graphql/index.rst
vpetrovykh: Well, technically yes, we could simply always reflect enums as GraphQL strings. Without making use of GraphQL enum type at all. On the other hand if we only sometimes reflect enums as strings that would be a bit confusing. 

We can probably even have custom types for each "enum-reflected-as-string" in GraphQL so that we can at least provide the list of valid values in the description there. But I'm not sure if that's the right approach.

PR #1256, opened 2020-02-22
graphql: Change the auto-generated type name template.

vpetrovykh:
When object types are reflected into GraphQL, we create an interface
with the original type name and an underlying type implementing that
interface with a name that has "\_Type" appended to it. The idea is that
since the convention in EdgeDB is to have camel-case type names adding
an "\_" will result in fewer cases of type name clashes.

Fixes: #1175.

### DIFF
--- a/docs/clients/99_graphql/index.rst
+++ b/docs/clients/99_graphql/index.rst
@@ -48,6 +48,12 @@ Known Limitations
 - EdgeDB :eql:type:`tuples <std::tuple>` are not supported in GraphQL
   reflection currently.
 
+- Every non-abstract EdgeDB object type is simultaneously an interface
+  and an object in terms of GraphQL type system, which means that for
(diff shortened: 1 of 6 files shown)


1st1:
> adding an "_" will result in fewer cases of type name clashes.

So this wouldn't eliminate them. What happens when there's a name clash?

vpetrovykh:
> So this wouldn't eliminate them. What happens when there's a name clash?

The GraphQL initializatin fails with `Schema must contain unique named types but contains multiple types named "Foo_Type"`. Since we generate a whole bunch of intermediate type names (interface, type, filter object, order object, insert object, etc.) there are a myriad of ways to have a type name collision if you really try. The most likely accidental one was the the `Type` suffix, but I can review the rest of the names to make sure they all break camel-case in some way so that at least "best practice" names are always safe.

PR #1261, opened 2020-02-23
Add annotations for schema constraints

RobertoPrevato:
Hi @elprans,
Annotating constraints.py was simpler than inheriting and referencing, but I have some doubts.

* there are still 3 "Any"
* I temporarily added "type: ignore" in a few cases

I add comments below

### DIFF
--- a/edb/schema/constraints.py
+++ b/edb/schema/constraints.py
@@ -420,9 +420,6 @@ class ConstraintCommand(
     ) -> Tuple[str, ...]:
         if isinstance(astnode, qlast.CreateConstraint):
             return ()
-
-        # assert isinstance(astnode, qlast.ConstraintOp) # THIS BREAKS!
-
         exprs = []
         args = cls._constraint_args_from_ast(schema, astnode, context)
         for arg in args:
@@ -615,8 +612,8 @@ class CreateConstraint(ConstraintCommand,
         if referrer_ctx is None:
             return super()._create_begin(schema, context)
 
-        # the reason for the following type ignore is explained
-        # in a comment in referencing.py
+        # type ignore below because mypy doesn't believe
+        # referrer_ctx is sd.ObjectCommandContext[so.Object]
         subject = referrer_ctx.scls  # type: ignore
         if subject.is_scalar() and subject.is_enum(schema):
             raise errors.UnsupportedFeatureError(
--- a/edb/schema/constraints.py
+++ b/edb/schema/constraints.py
@@ -39,6 +39,20 @@ from . import referencing
 from . import utils
 
 
+if TYPE_CHECKING:
+    from edb.common import parsing as c_parsing
+    from edb.schema import scalars as s_scalars
+    from edb.schema import schema as s_schema
+
+
+T = TypeVar('T')
+
+
+def _assert_not_none(value: Optional[T]) -> T:
+    assert value is not None
+    return value
+
+
 class Constraint(referencing.ReferencedInheritingObject,
                  s_func.CallableObject, s_abc.Constraint,
                  qlkind=ft.SchemaObjectClass.CONSTRAINT):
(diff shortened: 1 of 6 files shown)


edb/schema/delta.py
```diff
@@ -933,7 +933,8 @@ class ObjectCommand(
     scls: so.Object_T
     _delta_action: ClassVar[str]
     _schema_metaclass: ClassVar[Optional[Type[so.Object_T]]]
-    astnode: ClassVar[Type[qlast.DDLOperation]]
+    astnode: ClassVar[Union[Type[qlast.DDLOperation],
+                            List[Type[qlast.DDLOperation]]]]
```
RobertoPrevato: Inside `constraints.py`, astnode class property is used to specify a list of `DDLOperation` types, rather than a single type. So, this change to delta.py looks correct as it describes the reality of the other module.

However, this change creates complications inside delta.py (TODO you can see below). Can you please recommend the best direction?

edb/schema/constraints.py
```diff
@@ -372,27 +468,43 @@ def _constraint_args_from_ast(cls, schema, astnode, context):
...
+        field: so.Field[Any],
+        value: s_expr.Expression,
+    ) -> s_expr.Expression:
         from edb.edgeql import compiler as qlcompiler
 
         if field.name in ('expr', 'subjectexpr'):
             if isinstance(self, CreateConstraint):
                 params = self._get_params(schema, context)
             else:
+                # TODO: this line is not hit in any test, is the assertion
```
RobertoPrevato: How can we cover this block with tests?

edb/schema/constraints.py
```diff
@@ -372,27 +468,43 @@ def _constraint_args_from_ast(cls, schema, astnode, context):
...
                 anchors['__subject__'] = referrer_ctx.op.scls
 
+            # TODO: it is true, the Parameter does not inherit Object,
+            # how to resolve the type: ignore below?
             return s_expr.Expression.compiled(
                 value,
                 schema=schema,
                 modaliases=context.modaliases,
-                anchors=anchors,
+                anchors=anchors,  # type: ignore
```
RobertoPrevato: Without `type: ignore`, mypy says:

```
Argument "anchors" to "compiled" of "Expression" has incompatible type "Dict[str, Parameter]"; expected "Optional[Mapping[Union[str, Type[SpecialAnchor]], Object]]"mypy(error)
```

It is right, because in the actual `Mapping[str, Parameter]`, Parameter does not inherit `Object`.
I am not sure what is the best way to proceed here.

edb/schema/constraints.py
```diff
@@ -679,15 +859,28 @@ class AlterConstraint(ConstraintCommand,
...
+        assert isinstance(cmd, AlterConstraint)
 
         if isinstance(astnode, (qlast.CreateConcreteConstraint,
                                 qlast.AlterConcreteConstraint)):
-            subject_ctx = context.get(ConsistencySubjectCommandContext)
+            # TODO: how to make the following in a more "type-compliant" way?
+            # ConsistencySubjectCommandContext is an empty mixin used to
+            # group some classes of ObjectCommand
+            subject_ctx = context \
+                .get(ConsistencySubjectCommandContext)  # type: ignore
```
RobertoPrevato: As in the "TODO" I wrote above. The empty mixin is used to group some types of ObjectCommand, I would like your guidance on how to refactor the code to make mypy happy and respect the original purpose of the code. 

As a side note, I understand well the "error" reported by mypy and its ground. At the same time, I think this use of empty mixin is even elegant and part of the beauty of Python (it looks a pity to fix!).

edb/schema/constraints.py
elprans: `Parameter` *does* inherit from `Object` (via `ObjectFragment`). I'm not sure why mypy complains here.

edb/schema/constraints.py
elprans: We don't allow `ALTER`-ing constraint expressions currently, so the `else` branch should probably be an assertion error instead.

edb/schema/constraints.py
elprans: There's no reason to keep `ConsistencySubjectCommandContext` a pure mixin, make it inherit from `ObjectCommandContext`.

edb/schema/constraints.py
RobertoPrevato: I double checked, and this is the `Parameter` class from `ir` module, which doesn't inherit Object (specified in the return type of `compiler.get_param_anchors_for_callable`). The `Parameter` class that is an `ObjectFragment` is defined in `functions` module under schema.

edb/schema/constraints.py
```diff
@@ -39,6 +39,20 @@
...
+    from edb.common import parsing as c_parsing
+    from edb.schema import scalars as s_scalars
+    from edb.schema import schema as s_schema
+
+
+T = TypeVar('T')
+
+
+def _assert_not_none(value: Optional[T]) -> T:
+    assert value is not None
```
1st1: Please replace this with an `if value is None: raise TypeError('...')`

edb/schema/constraints.py
1st1: I think in cases like this we shouldn't use assertions but an explicit `if not isinstance: raise`.

1st1: Drawing the line where to put an `assert` expression and where to use an explicit exception is kinda hard though. In this case we have an `if` on one thing but check another thing in the `else`; the code isn't easy to read as result.

edb/schema/constraints.py
RobertoPrevato: @elprans, when I modify the `ConsistencySubjectCommandContext` this way:

```py
class ConsistencySubjectCommandContext(
    sd.ObjectCommandContext[ConsistencySubject]
):
    # context mixin
    pass
```

mypy complains: "Value of type variable "CommandContextToken_T" of "get" of "CommandContext" cannot be "ConsistencySubjectCommandContext"mypy(error)"

I read the code several times, and I don't see why mypy comes to that conclusion.

1. the ConsistencySubject is a so.Object
2. ObjectCommandContext is CommandContextToken[ObjectCommand[so.Object_T]]
3. CommandContext.get expects a type of CommandContextToken_T

Can you please give me an hint on this?

edb/schema/constraints.py
RobertoPrevato: @1st1, I followed your advice. I generally prefer to raise `if not instance` too, especially because assertions can be disabled when running Python with -O flag. But I have mixed feelings for the assertions used to pass mypy validation: assertions in this case look reasonable because it's something about the programmer rather than the end user.

In this specific case, I altered the code to raise `if not isinstance(self, CreateConstraint)`, since the "else block" was not tested and Elvis said that scenario is not supported.

edb/schema/constraints.py
RobertoPrevato: To add something more:

1. the dictionary returned by `qlcompiler.get_param_anchors_for_callable` is `Dict[str, irast.Parameter]`
2. irast.Parameter does not inherit from schema `Object`, hence the error report from mypy

Since this comment in `edb.ir.ast`:
```
        anchors:
            Predefined symbol table.  Maps identifiers
            (or ``qlast.SpecialAnchor`` instances) to specified
            schema objects or IR fragments.
```
says that `anchors` keys can be of type `str` or `qlast.SpecialAnchor`, and values can be "schema objects _or_ IR fragments", the place to correct is the type annotation for `anchors`:

instead of 
```py
    anchors: Optional[
        Mapping[Union[str, qlast.SpecialAnchorT], s_obj.Object]
    ] = None,
```

this:
```py
    anchors: Optional[
        Mapping[Union[str, qlast.SpecialAnchorT],
                Union[s_obj.Object, irast.Base]]
    ] = None,
```
**Note:** `stmtctx.py.init_context` already annotates `anchors` this way.

However, mypy still complains that `Dict[str, irast.Parameter] is not compatible with `Mapping[Union[str, qlast.SpecialAnchorT], Union[s_obj.Object, irast.Base]]`. IMHO, this looks surprising because `Mapping[str, irast.Base]` should match.

This happens because of the way union on keys and values are handled.

For example, the following two annotations are not equivalent:
```py
def example(value: Mapping[Union[str, int], Union[str, int]]):
    ...


def example2(value: Union[Mapping[str, int],
                          Mapping[int, str],
                          Mapping[str, str],
                          Mapping[int, int]]):
    ...
```

RobertoPrevato: @elprans, my conclusion is that `type: ignore` here is the best thing, because of these two things: https://github.com/python/mypy/issues/8477

Even changing the annotation for anchors everywhere to be proper (I gave a try to this), then we have issues in places where we would need to use the key as union of two possible key types.

elprans: That's fine.

edb/schema/constraints.py
elprans: Smells like a mypy bug. Let's keep things as is.

Issue #1254, opened 2020-02-21
The `DESCRIBE` command should preserve original expressions where possible


vpetrovykh:
Currently `DESCRIBE MODULE ... AS SDL` provides a normalized expression for the default values as opposed to the original expression.

See [this spectrum post](https://spectrum.chat/edgedb/general/production-ready~8f220310-d1c7-4b92-bb48-b637e0e7a447?m=MTU4MjI4Mjk4MTI1Ng==)

PR #1263, opened 2020-02-25
Adjust the DESCRIBE AS SDL and AS TEXT logic.

vpetrovykh:
Change the logic when DESCRIBE command shows the original expression.
Basically if the normalized expression appears to only introduce a
default module that is the same as the module the object belongs to then
SDL and TEXT versions of describe can safely use the original short
expression. Otherwise, a normalized expression must be given to avoid
confusion.

Fixes: #1254

### DIFF
--- a/edb/schema/delta.py
+++ b/edb/schema/delta.py
@@ -790,11 +790,12 @@ class CommandContext:
         *,
         schema: Optional[s_schema.Schema] = None,
         modaliases: Optional[Mapping[Optional[str], str]] = None,
-        declarative: bool = False,
+        declarative: bool = False,  # SDL
+        descriptive_mode: bool = False,  # TEXT
+        verbose_mode: bool = False,  # VERBOSE
         stdmode: bool = False,
         testmode: bool = False,
         disable_dep_verification: bool = False,
-        descriptive_mode: bool = False,
         schema_object_ids: Optional[
             Mapping[Tuple[str, Optional[str]], uuid.UUID]
         ] = None,
@@ -804,11 +805,12 @@ class CommandContext:
         self._cache: Dict[Hashable, Any] = {}
         self._values: Dict[Hashable, Any] = {}
         self.declarative = declarative
+        self.descriptive_mode = descriptive_mode
+        self.verbose_mode = verbose_mode
         self.schema = schema
         self._modaliases = modaliases if modaliases is not None else {}
         self.stdmode = stdmode
         self.testmode = testmode
-        self.descriptive_mode = descriptive_mode
         self.disable_dep_verification = disable_dep_verification
         self.renames: Dict[str, str] = {}
         self.renamed_objs: Set[so.Object] = set()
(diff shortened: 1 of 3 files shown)


edb/schema/delta.py
```diff
@@ -2176,12 +2176,50 @@ def _get_expr_field_ast(
...
+                if (num_aliases == 0 or
+                    # No aliases at all or a single module alias are
+                    # both indicators that we might be able to use
+                    # orig_expr.
+                    (num_aliases == 1
+                        and isinstance(aliases[0], qlast.ModuleAliasDecl)
+                        # and that the module matches
+                        and aliases[0].module == parent_op.classname.module)):
+
+                    expr_ql = edgeql.parse_fragment(self.new_value.origtext)
```
1st1: I'm not sure we need this stuff. Too much guessing and I don't like that.

@elprans your thoughts?

vpetrovykh: In defence of this, the main point is that we can rely on the normalization code to create a fairly predictable query here and if the query doesn't fit this, it pretty much implies that it wasn't normalized so it doesn't need adjustments.

edb/schema/delta.py
```diff
@@ -2176,12 +2176,45 @@ def _get_expr_field_ast(
...
-            expr_ql = edgeql.parse_fragment(self.new_value.origtext)
-        else:
+        expr_ql = None
+        if context.descriptive_mode or context.declarative:
+            # Potentially we want to use the original expression
+            # instead of the normalized one, but this can only be done
+            # safely if the normalized form has sets the same module
+            # as the one that hosts the object we're in.
+            #
+            # We also make a special exception for indexes which
```
elprans: The comment here seems to be outdated.  Also, indexes *can* use functions, the scope of index expressions is exactly like that of constraints.

edb/schema/delta.py
elprans: This should also check if `origtext` != `text` to account for cases when `WITH MODULE` was actually specified by the user.  I would also factor this check out and place it near `imprint_expr_context()` to make sure they don't go out of sync.

elprans: > @elprans your thoughts?

I think that by adding `and origtext != text` to the condition the check is made unambiguous.  Implementing it as a helper near `imprint_source_context()`  rather than open-coding here should further solidify this.

1st1:
I think we should do a simpler thing:

* for SDL to always output the normalized expression.
* for TEXT to always output the original expression.
* for TEXT VERBOSE to always output the normalized expression.

vpetrovykh:
The specific version matters a lot for indexes, so regardless of SDL, DDL or TEXT we would want to output the original expr version for them so that it can actually be used to refer to the real index.

We can do as you suggested otherwise for the different modes.

1st1:
What's up with this PR?

vpetrovykh:
So is the consensus then:
* for SDL to always output the normalized expression.
* for TEXT to always output the original expression.
* for TEXT VERBOSE to always output the normalized expression.
* for indexes _always_ output the original, regardless of the mode since they are identified by the original text.

1st1:
> for TEXT to always output the original expression.

@elprans should we do that? Maybe we should always output the normalized one because it would reflect the correct state of the schema including renames etc.?

elprans:
> because it would reflect the correct state of the schema including renames etc.?

The "original" here does not necessarily mean the initial text, it just means that the expression is unmangled by the injected `WITH` block.  Renames will alter all expressions where the renamed object appears.

> for indexes always output the original, regardless of the mode since they are identified by the original text.

Constraints also.

1st1:
@vpetrovykh Please finalize this PR.

vpetrovykh:
I've rebased the PR and made constraints obey the same rule as indexes w.r.t. preserving and showing the original expression at all times because they are basically defined by it.

This causes `test_get_migration_multi_module_03` to fail due to a default module being injected into the finalized version of the expression and that default module is inferred to be different in SDL and DDL. This is an inherent ambiguity that 99% of the time doesn't matter as expressions for indexes or constraints typically involve pointers directly on the object and some constants and thus don't rely on a specific default module (`WITH MODULE foo` injected into the expression in the process of normalization). This seems more of an issue with how the expressions are normalized internally than an actual issue with grammar.

elprans:
@vpetrovykh I rebased on top of my "DDL identity" work that's in master, so index and constraint hacks are no longer necessary.  Not sure whether we still want to merge this given the pending normalization rework.

vpetrovykh:
This appears to be no longer necessary due to #1522

PR #1268, opened 2020-03-04
Corrects the annotation for anchors, improves constraints.py

RobertoPrevato:
I finally understood that `Union` defines an _intersection_ of types, not a _sum_ of possible types. I was misled by its name and its closeness with concrete types, all along. It somehow looks like an interface in Java or C#, with the difference that it is an implicit interface under the _type POV_, while being explicit under the _possible concrete types POV_.

Then it became clear that the way we were thinking of the [`anchors`](https://github.com/edgedb/edgedb/pull/1261#discussion_r386141078) was wrong.

This is **not** the proper way:
```python
    anchors: Optional[
        Mapping[Union[str, qlast.SpecialAnchorT],
                Union[s_obj.Object, irast.Base]]
    ] = None,
```

This is the proper way:
```python
AnchorsKeyType = typing.TypeVar("AnchorsKeyType", str, qlast.SpecialAnchorT)
AnchorsValueType = typing.TypeVar("AnchorsValueType", so.Object, Base)

    anchors: Optional[
        Mapping[AnchorsKeyType, AnchorsValueType]
    ] = None,
```

### DIFF
--- a/edb/edgeql/compiler/__init__.py
+++ b/edb/edgeql/compiler/__init__.py
@@ -169,11 +169,11 @@ def compile_ast_to_ir(
     modaliases: Optional[Mapping[Optional[str], str]] = None,
     anchors: Optional[
         Mapping[
-            Union[str, qlast.SpecialAnchorT],
-            Union[irast.Base, s_obj.Object],
+            irast.AnchorsKeyType,
+            irast.AnchorsValueType
         ]
     ] = None,
-    path_prefix_anchor: Optional[qlast.SpecialAnchorT] = None,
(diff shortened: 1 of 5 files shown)


edb/edgeql/compiler/__init__.py
```diff
@@ -504,8 +504,7 @@ def compile_func_to_ir(
 
     ir = compile_ast_to_ir(
         tree, schema,
-        anchors=param_anchors,  # type: ignore
```
RobertoPrevato: This `type: ignore` resolved itself, after my change. :sparkles:

Issue #1269, opened 2020-03-05
Bugs with line continuation in strings


vpetrovykh:
This happens in the current master branch.

Try this:
```
SELECT "some text \
will be completely \
        ignored";
```

There's also another bug that's harder to express in REPL, but I've made a test case for it. Basically, the line continuation is unhappy if it's followed by `\r` as opposed to `\n`. This should not be the case as either one of the two may appear as a newline depending which system this is coming from. I personally ran across this while playing around with queries submitted from the browser in tutorial.

As far as I can tell both of these are probably lexer issues.

PR #1270, opened 2020-03-05
tests: Add a couple of line continuation tests.

vpetrovykh:
Add a couple of tests for line continuation corner cases.

Issue #1269

### DIFF
--- a/tests/test_edgeql_expressions.py
+++ b/tests/test_edgeql_expressions.py
@@ -3057,7 +3057,7 @@ class TestExpressions(tb.QueryTestCase):
 
             bb';
             ''',
-            ['bbaa bb'],
+            ['bbaa \n\n            bb'],
         )
 
         await self.assert_query_result(
@@ -3085,7 +3085,6 @@ class TestExpressions(tb.QueryTestCase):
                 r"SELECT 'bb\   aa';"
             )
 
-    @test.xfail('The string literal is missing some content')
     async def test_edgeql_expr_string_12(self):
         # Issue #1269
         await self.assert_query_result(
@@ -3096,7 +3095,6 @@ aa \
             ['bbaa bb'],
         )
 
-    @test.xfail('alternative newline is not accepted after line continuation')
     async def test_edgeql_expr_string_13(self):
         # Issue #1269
         #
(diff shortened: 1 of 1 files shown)


tests/test_edgeql_expressions.py
```diff
@@ -3085,6 +3085,32 @@ class TestExpressions(tb.QueryTestCase):
...
+        # Issue #1269
+        #
+        # \r or \n should both be allowed after line continuation symbol
+        await self.assert_query_result(
+            "SELECT 'bb\\\n   aa';",
+            ['bbaa'],
+        )
+
+        await self.assert_query_result(
+            "SELECT 'bb\\\r   aa';",
```
tailhook: This should probably has `\r\n` or another test for `\r\n` because turning `'a\\\r\nb` into `a\nb` sounds bad.

tests/test_edgeql_expressions.py
```diff
@@ -3057,7 +3057,7 @@ class TestExpressions(tb.QueryTestCase):
 
             bb';
             ''',
-            ['bbaa bb'],
+            ['bbaa \n\n            bb'],
```
vpetrovykh: Actually, I strongly disagree with this change. The whole point of our `\` is that is consumes all white-space until next non-white-space, thus allowing to format a single logical string as you like. Consider this:
```
SELECT 'Always remember that: "\

        Long strings are very \
        useful because they let \
        the client make rendering \
        decisions.\

        " is all in one line (or, perhaps, \
        one paragraph)';
```

The point is that we wanted to find a way to make our multi-line strings somewhat easier to format as single logical line and only folding a single line after `\` is not that helpful. In practice I would never use that feature if I had several newlines, but as my example above shows I would be quite happy with the option of visually separating text blobs with newlines while keeping them all a contiguous string.

@1st1, @elprans what do you think about handling whitespace after `\`?

1st1: I also disagree with the change. Let's handle `\` the way rust handles them:

Screen Shot 2020-03-05 at 12 17 27 PM

1st1: That means that the behavior we implemented in a2 and current master branch is the correct one.

tailhook: Okay. Fixed that

PR #1272, opened 2020-03-06
Install rust repl as `edgedb` and pythonic as `edgedb-old`

tailhook:
(draft because I would like to see how tests fail on CI)


tailhook:
Okay, so tests keep working because they actually don't invoke a command-line tool, but rather use pythonic import of the internals. Do we want to redirect those tests? Duplicate them? Or keep everything as is?

elprans:
It appears that this installs the binary into `edb` i.e. "in-place", which is unnecessary.

elprans:
> pythonic import of the internals. Do we want to redirect those tests? Duplicate them? Or keep everything as is?

Consider the in-tree CLI tests to be deprecated.  We will switch to writing tests in the Rust repo.

tailhook:
> It appears that this installs the binary into edb i.e. "in-place", which is unnecessary.

This is how it works:
1. `venv/bin/edgedb` refers to a generated module `edb/_gen_edgedb.py`
2. `edb/_gen_edgedb.py` looks for a binary in it's own directory

Note: it's probably bad idea to run the binary from the build dir. And what is the alternative? Putting binary directly to the `venv/bin`? Well, I don't know the reasoning why it haven't been done this way in setuptools-rust. But it doesn't look like it's easy to fix without rewrtiting a bit chunk of setuptools-rust.

I should probably add the path to `.gitignore`, though.

Another question is what we want to have for "production" packages. I guess we want to use `cargo install` for that. That might mean, we need some flag in setup.py to disable installation of the repl?

elprans:
> Note: it's probably bad idea to run the binary from the build dir. And what is the alternative? Putting binary directly to the venv/bin

Yes.  This is the setuptools way.  The `_gen_edgedb.py` can go there as well.

elprans:
> That might mean, we need some flag in setup.py to disable installation of the repl?

No need. The packaging rules can simply omit the installation of the binary.

tailhook:
>> Note: it's probably bad idea to run the binary from the build dir. And what is the alternative? Putting binary directly to the venv/bin
>
> Yes. This is the setuptools way. The _gen_edgedb.py can go there as well.

I don't understand what do you propose here, rewriting another part of setuptools-rust? Without understanding the reasons why things put this way, right? Also we don't need `_get_edgedb.py` in this case, so I'm not sure I understand.

elprans:
> Without understanding the reasons why things put this way, right?

My understanding is that they simply didn't care.  But polluting the source tree with build products is *not* the correct way of building things in Python.

> I don't understand what do you propose here, rewriting another part of setuptools-rust?

I have a feeling that we can get away with some fiddling of `inplace` once more.  I'll take a look.

elprans:
Looks like there's no reasonable way to convince setuptools to install the non-Python "script" properly, so the current approach is the one we'll have to live with.  I added the build products to .gitignore.  Now testing if the current packaging still works.

elprans:
Now that `edgedb-cli` has been split into a separate repo, this PR is obsolete.

PR #1274, opened 2020-03-07
Add annotations for schema scalars.py

RobertoPrevato:


1st1:
Good job!

edb/schema/scalars.py
```diff
@@ -183,13 +232,18 @@ class CreateScalarType(ScalarTypeCommand, inheriting.CreateInheritingObject):
     astnode = qlast.CreateScalarType
 
     @classmethod
-    def _cmd_tree_from_ast(cls, schema, astnode, context):
+    def _cmd_tree_from_ast(
+        cls,
+        schema: s_schema.Schema,
+        astnode: qlast.DDLOperation,
+        context: sd.CommandContext,
+    ) -> Union[CreateScalarType, sd.CommandGroup]:
```
elprans: You don't need to be very specific as to the return type here, `sd.Command` is sufficient.

PR #1277, opened 2020-03-09
Import tokenizer from edgedb-rust

tailhook:
I'm not sure do we want to run `cargo test` from `edb test` or keep it separately run (I've added it as a separate step for CI).

Also, I'm going to drop `edgeql-pytest` crate as that tests python bindings which should already have been tested as part of the edgedb tests.

Another point here: I've just put sources here without importing the whole history. Let me know if you think importing history is a good idea.


tailhook:
> SHould we rename `edgeql-python` to something more meaningful? It made sense inside a rust repo, but being inside a predominantly Python repo it makes less sense.

We can. Not sure what the name makes more sense, though.

1st1:
> I'm not sure do we want to run cargo test from edb test or keep it separately run (I've added it as a separate step for CI).

Would it be possible to test the rust code entirely with Python?

tailhook:
> Would it be possible to test the rust code entirely with Python?

This doesn't make sense for "unit" tests, as there are a little bit more that used from python in lexer (preparser that is used by repl to delimit statements). And unit tests may test smaller internal things not exposed to python at all.

tailhook:
It looks like cargo is able to discover packages in subfolders too, as long as `Cargo.toml` of the workspace is in the root folder.

So we can put them somewhere in `edb/edgeql-parser` and `edb/edgeql-rust` ? Given that:
1. Python module called `edb._edgeql_rust` (and we can't put it deeper, i.e. more dots in the path, because of some limitations in cpython crate).
2. For `edgeql-parser` it's important to keep folder name same as crate name, so that the structure doesn't look weird (it's not a technical limitation), as we're probably going to publish it on crates.io (marking it as internal implementation) to publish REPL too.

tailhook:
Moved things to `edb/edgeql-parser` and `edb/edgeql-rust`

.github/workflows/tests.yml
```diff
@@ -67,6 +67,9 @@ jobs:
         # and don't want them to be reinstalled in an "isolated env".
         pip install --no-use-pep517 --no-deps -e .[test,docs]
 
+    - name: Cargo test
```
elprans: Can we invoke `cargo test` from the python testsuite like we do other external tools (mypy etc)?

tailhook: Sure, we can, as I've written in the very first message. I'm just not sure it makes much sense (i.e. it will look like a single test, etc.)

PR #1289, opened 2020-03-18
Upgrade to graphql-core 3.0.3

tailhook:
Things to note:

1. I strip `__` in front of some names, because new graphql-core doesn't allow them:
    ```
    Name '__SettingAliasAugmented__of_group_Type' must not begin with '__', which is reserved by GraphQL introspection.
    ```
    There should be a better way instead of stripping, but I failed to find one. Need some help. @vpetrovykh ? (FWIW, I don't want to hack graphql-core to allow such names if possible)
2. xfail tests: literal integers are now rejected when float is expected. Looks like it's a feature of the graphql but is not spec-compliant. I'll look into fixing it.
3. I have to double-check that all used class names are referred by their new names (in case some branch is not covered by tests).

But overall the PR is close to be ready for merging. The upside is that one expected failure is now fixed.

### DIFF
--- a/edb/graphql/_patch_core.py
+++ b/edb/graphql/_patch_core.py
@@ -22,7 +22,7 @@ from __future__ import annotations
 
 def patch_graphql_core():
     import graphql
-    import graphql.utils.type_comparators as type_comparators
+    import graphql.utilities.type_comparators as type_comparators
 
     old_is_type_sub_type_of = type_comparators.is_type_sub_type_of
 
--- a/edb/graphql/codegen.py
+++ b/edb/graphql/codegen.py
@@ -50,13 +50,13 @@ class GraphQLSourceGenerator(codegen.SourceGenerator):
             self.write(' on ')
             self.visit(node.type_condition)
 
-    def visit_Name(self, node):
+    def visit_NameNode(self, node):
         self.write(node.value)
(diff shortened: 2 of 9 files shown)


1st1:
@vpetrovykh please review. I'll also take a look later.

tailhook:
Another interesting case, introspection request like this:
```
query($val: String!) {
    __type(name: $val) {
        __typename
        name
        kind
    }
}
{"val": "Object"}
```
Doesn't work:
```
QueryError: Variable '$val' got invalid value <StringValueNode instance>; Expected type String. String cannot represent a non string value: <StringValueNode instance>
```
While previously it worked but returned `null` instead of actual result. Investigating...

tailhook:
> Another interesting case, introspection request like this:

It looks like we can make this work but, caching of introspection queries was wrong: it looks like response was cached with default values of variables and never refreshed. I'll fix support of variables, but I'm going to skip the caching issue in this PR and fix it in follow-up one (perhaps after graphql rewrite).

tailhook:
So I've fixed (1) by prefixing. (2) will be fixed as part of #1294 (because we already have error scanning there). And I've rechecked all the places as part of (3).

So this PR should be ready to land now.

PR #1291, opened 2020-03-19
Add annotations for links

RobertoPrevato:


edb/schema/links.py
```diff
@@ -100,20 +106,27 @@ class Link(sources.Source, pointers.Pointer, s_abc.Link,
...
 
-    def compare(self, other, *, our_schema, their_schema, context=None):
+    def compare(
+        self,
+        other: so.Object,
+        *,
+        our_schema: s_schema.Schema,
+        their_schema: s_schema.Schema,
+        context: Optional[so.ComparisonContext] = None,
+    ) -> Union[float, NotImplemented]:
```
elprans: In my pending branch I replaced returning `NotImplemented` with `raise NotImplementedError`, please do the same here.

edb/schema/links.py
```diff
@@ -236,15 +280,23 @@ def _cmd_tree_from_ast(cls, schema, astnode, context):
...
-    def _apply_field_ast(self, schema, context, node, op):
-        objtype = context.get(LinkSourceCommandContext)
+    def _apply_field_ast(
+        self,
+        schema: s_schema.Schema,
+        context: sd.CommandContext,
+        node: qlast.DDLOperation,
+        op: sd.AlterObjectProperty,
+    ) -> None:
+        objtype = context.get(LinkSourceCommandContext)  # type: ignore
```
elprans: Consider using `get_referrer_context()` instead.

edb/schema/links.py
```diff
@@ -277,16 +331,25 @@ def _apply_field_ast(self, schema, context, node, op):
...
         if refdict.attr != 'pointers':
             return cmd
 
-        parent_ctx = context.get(LinkSourceCommandContext)
+        # type ignore below, because LinkSourceCommandContext is used
+        # as mixin in ObjectTypeCommandContext; asserting that parent_ctx
+        # is an instance of ObjectTypeCommandContext confuses mypy that
+        # considers it as "<nothing>"; fixing requires changing how
+        # this context is obtained
+        parent_ctx = context.get(LinkSourceCommandContext)  # type: ignore
```
elprans: Consider using `get_referrer_context()` instead.

edb/schema/links.py
```diff
@@ -374,11 +437,24 @@ class SetTargetDeletePolicy(sd.Command):
     astnode = qlast.OnTargetDelete
 
     @classmethod
-    def _cmd_from_ast(cls, schema, astnode, context):
-        return sd.AlterObjectProperty(property='on_target_delete')
+    def _cmd_from_ast(
```
elprans: This class is dead code, remove it please.

RobertoPrevato: Do you mean the whole `SetTargetDeletePolicy`? If I remove this class, tests fail during arrangement phase.

elprans: Ah, nevermind, it's not dead, I misread the code.

edb/schema/links.py
```diff
@@ -40,14 +40,20 @@
...
 
-def merge_actions(target: so.Object, sources: List[so.Object],
-                  field_name: str, *, schema) -> object:
+def merge_actions(
+    target: so.InheritingObject,
+    sources: List[so.Object],
+    field_name: str,
+    *,
+    schema: s_schema.Schema,
+) -> object:
```
ambv: I know you didn't change this now but I noticed this. Returning `object` from a function means that you declare that the only safe methods to call and attributes to access are those that are present on the base `object` type, which is very little. Such a returned value is very annoying to use.

Did you mean `Any` here?

RobertoPrevato: Thanks, Łukasz. I just looked into this, and indeed `Any` is the right annotation here: the function `merge_actions` is used as `merge_fn`, whose protocol is defined in `schema.objects` as returning Any. Pushing to fix this.

PR #1293, opened 2020-03-20
Strict typing mode for edb.schema.{objects, types}

elprans:
This enables strict type checking for `edb.schema.objects` and `edb.schema.types`.  As usual,
this uncovered some class hierarchy wrongness and other infidelities, so there's associated code movement. This also improves the mypy plugin, most notably fixing support for the explicit typing annotations on fields.

### DIFF
--- a/edb/schema/objects.py
+++ b/edb/schema/objects.py
@@ -61,7 +61,7 @@ if TYPE_CHECKING:
         def __call__(
             self,  # not actually part of the signature
             target: InheritingObject,
-            sources: List[Object],
+            sources: Iterable[Object],
             field_name: str,
             *,
             schema: s_schema.Schema,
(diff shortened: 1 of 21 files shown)


edb/schema/objects.py
```diff
@@ -382,20 +387,18 @@ def required(self) -> bool:
...
+        self,
+        instance: Optional[Object],
+        owner: Type[Object],
+    ) -> Optional[T]:
         if instance is not None:
             raise FieldValueNotFoundError(self.name)
         else:
-            return self
+            raise AttributeError(
+                f"type object {owner.__name__!r} "
```
ambv: nit: I prefer leading whitespace in such situations since if you do this consistently, it's easier to spot a missing space.

Issue #187, opened 2018-06-01
GraphQL features and overview of existing technologies


vpetrovykh:
GraphQL is designed to be an abstraction for APIs, not specifically
for DBs. There has been some attempts at adapting GraphQL to query DBs
instead of the native query languages.

Prisma
======

Prisma is one of the latest GraphQL abstraction layers on top of a SQL
DB.

Data model in Prisma allow adding information about specific GraphQL
fields via directives. There's `unique`, `relation`, `default`, and
`rename`:

* `unique` denotes scalars with unique values
* `default` specifies the default value for a scalar
* `relation` is used to disambiguate how 2 types are related and also
  what the onDelete behavior is
* `rename` allows renaming types and fields

From the defined data model Prisma automatically produces a GraphQL
schema with all the various querying options. An example of such a
schema can be found here:
https://gist.github.com/gc-codesnippets/f302c104f2806f9e13f41d909e07d82d

Queries
-------

For a type Foo defined in the data model the following querying
options will be created:

* `foo` - get a single object using the `where` argument
* `foos` - get a number of objects using `where` as well as `orderBy`,
  `skip`, `after`, `before`, `first`, `last` to filter, order and
  slice the results.
* `foosConnection` - similar to `foos` (same arguments), but returning
  relay connections (https://facebook.github.io/relay/graphql/connections.htm)
* `node` - get a single object by its ID (this is generic for all
  types defined in the data model as all of them will be implementing
  the Node interface)

The `where` argument type contains all the complexity of the filtering
expressions. It is composable via `AND`, and `OR` and it has a whole
bunch of fields specialized in other operations for each scalar fields
of the base type. Here's an example of `where` fields generated for a
Type that has a `name` of type `String`:
```
    name: String
    name_not: String
    name_in: [String!]
    name_not_in: [String!]
    name_lt: String
    name_lte: String
    name_gt: String
    name_gte: String
    name_contains: String
    name_not_contains: String
    name_starts_with: String
    name_not_starts_with: String
    name_ends_with: String
    name_not_ends_with: String
```

The `where` argument also supports nesting based on relations:
```
query {
  posts(where: {
    author: {
      age_gt: 18
    }
  }) {
    id
    title
    author {
      name
      age
    }
  }
}
```

In addition to pagination/cursor support various aggregation features
are also exposed via Connection fields, e.g.:
```
# Count all posts with a title containing 'GraphQL'
query {
  postsConnection(where: {
    title_contains: "GraphQL"
[...]

PR #1294, opened 2020-03-20
Rewriting GraphQL requests for constant extraction

tailhook:
There are two kinds of failing tests:
1. Variables in introspection queries aren't part of the cache key
2. Transforming errors on variables to errors on constants

*(Note: this PR is on top of `graphql_core_3` branch, so approvals are unuseful here, I'll recreate PR when that branch is landed)*


ambv:
Man, that's a huge change.

tailhook:
> Man, that's a huge change.

Well, generally it's because it includes the whole graphql-parser and its tests. Perhaps, you can compare starting from that: https://github.com/edgedb/edgedb/compare/f135c5ca7d9e49af8825798691f56f6b912e8fa1...graphql_rewrite . Pretty small: 
image

tests/test_http_graphql_query.py
```diff
@@ -1190,8 +1190,12 @@ def test_graphql_functional_arguments_21(self):
     def test_graphql_functional_arguments_22(self):
         with self.assertRaisesRegex(
                 edgedb.QueryError,
-                r"invalid value for 'after'",
-                _line=5, _col=32):
+                # this error message is subpar, but this is what we get
+                # from postgres, because we transfer bigint values to postgres
+                # as strings
+                r'invalid input syntax for type bigint: "aaaaa"',
```
vpetrovykh: There's a historical reason why this is a string and the short version of the story is that this was inspires by Relay Connections (see here: https://github.com/edgedb/edgedb/issues/187). Depending on whether we still find that style useful here, maybe we can change the type to be an integer like "first" and "last".

vpetrovykh:
Other than my non-blocking comment, this looks fine.

tests/test_http_graphql_query.py
tailhook: Well, generally I don't think that pagination using `OFFSET` (which `after` and `before` are currently mapped to) is a good solution. Pagination should be coupled with sort order and use sorting key as `after` and `before`. But this is irrelevant to this PR. Please, create a new issue if you think this is worth changing now.

PR #1298, opened 2020-03-24
[DO NOT MERGE] Test for "InternalServerError" being returned from an open connection

ambv:
That happens on `CONFIGURE SYSTEM` commands which end up recreating the
listening sockets.

### DIFF
--- a/tests/test_database.py
+++ b/tests/test_database.py
@@ -45,3 +45,20 @@ class TestDatabase(tb.ConnectedTestCase):
             await conn.aclose()
         finally:
             await self.con.execute('DROP DATABASE mytestdb;')
+
+    async def test_database_connection_reset_after_configure(self):
+        conn = await self.connect(database="edgedb")
+        try:
+            await conn.execute(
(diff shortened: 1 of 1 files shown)


tests/test_database.py
```diff
@@ -45,3 +45,20 @@ class TestDatabase(tb.ConnectedTestCase):
...
         finally:
             await self.con.execute('DROP DATABASE mytestdb;')
+
+    async def test_database_connection_reset_after_configure(self):
+        conn = await self.connect(database="edgedb")
+        try:
+            await conn.execute(
+                "CONFIGURE SYSTEM SET listen_addresses := {'0.0.0.0', '::'}"
+            )
+            await conn.execute(
```
ambv: This `execute()` raises an InternalServerError.

ambv:
I basically need to know what course of action we should take here.

elprans:
I think the current behavior of silently killing existing connections is bad.  This needs be handled gracefully, i.e. keep existing connections, but accept new connections only on the configured address. This probably would require some rework in `port.py`.

PR #1299, opened 2020-03-25
Rewriting GraphQL queries for constant extraction

tailhook:
Rebased #1294

But it would be better to apply #1277 first (that sets rust workspace and directory structure)


1st1:
I'll also take a quick look at this right after the call

edb/graphql/translator.py
```diff
@@ -108,6 +123,23 @@ class TranspiledOperation(NamedTuple):
...
+
+    def __getitem__(self, key):
+        self.touched.add(key)
+        return super().__getitem__(key)
+
+    def values(self):
+        raise NotImplementedError()
+
+    def items(self):
+        raise NotImplementedError()
```
1st1: I'd add `keys()` / `__contains__()` too to make sure one doesn't accidentally lose tracking when debugging/writing more code.

edb/graphql/translator.py
```diff
@@ -204,18 +229,25 @@ def visit_DocumentNode(self, node):
...
                 result = graphql.execute(
                     self._context.gqlcore.graphql_schema,
                     self._context.document_ast,
                     operation_name=operation_name,
-                    variable_values=self._context.variables)
+                    variable_values=vars)
+                for var in vars.touched:
+                    self._context.vars[var] = (
+                        self._context.vars.get(var)
+                        ._replace(critical=True)
```
1st1: For better readability

```suggestion
                        self._context.vars.get(var)._replace(critical=True)
```

edb/graphql/translator.py
```diff
@@ -1510,24 +1547,100 @@ def value_node_from_pyvalue(val: Any):
...
+
+def convert_errors(
+    errs: List[gql_error.GraphQLError], *,
+    substitutions: Optional[Dict[str, Tuple[str, int, int]]],
+) -> List[gql_error.GraphQLErrors]:
+    result = []
+    for err in errs:
+        m = REWRITE_TYPE_ERROR.match(err.message)
+        if not m:
+            # we allow convefsion from Int to Float, and that is allowed by
```
1st1:
```suggestion
            # We allow conversion from Int to Float, and that is allowed by
```

edb/graphql/translator.py
```diff
@@ -1510,24 +1547,100 @@ def value_node_from_pyvalue(val: Any):
...
+            # graphql spec. It's unclear why graphql-core chokes on this
```
1st1: Can we make a PR to their repo?

edb/graphql/translator.py
```diff
@@ -1510,24 +1547,100 @@ def value_node_from_pyvalue(val: Any):
...
+            if INT_FLOAT_ERROR.match(err.message):
```
1st1: This is fragile. Do we have a functional test for this specific hack to track regressions?

graphql-rewrite/src/common.rs
```diff
@@ -0,0 +1,335 @@
...
+use num_traits::cast::ToPrimitive;
+
+use crate::tokenizer::{Kind as T, Token, TokenStream};
+use crate::helpers::{punct, ident, kind, name};
+use crate::position::Pos;
+
+/// Text abstracts over types that hold a string value.
+/// It is used to make the AST generic over the string type.
+pub trait Text<'a>: 'a {
+    type Value: 'a + From<&'a str> + AsRef<str> + std::borrow::Borrow<str> + PartialEq + Eq + PartialOrd + Ord + fmt::Debug + Clone;
```
1st1: You typically limit line length to 80, is there a reason why this line is longer? (just trying to understand the style)

graphql-rewrite/src/entry_point.rs
```diff
@@ -0,0 +1,274 @@
...
+            };
+        }
+        oper.ok_or_else(|| Error::NotFound("no operation found".into()))?
+    };
+    let (tokens, end_pos) = token_array(s)?;
+    let mut src_tokens = TokenVec::new(tokens);
+    let mut tokens = Vec::with_capacity(src_tokens.len());
+
+    let mut variables = Vec::new();
+    //let mut var_decl = Vec::new();
```
1st1: dead code

graphql-rewrite/src/entry_point.rs
```diff
@@ -0,0 +1,274 @@
...
+            PyTokenKind::Spread => false,
+            PyTokenKind::Colon => false,
+            PyTokenKind::Equals => false,
+            PyTokenKind::At => false,
+            PyTokenKind::BracketL => false,
+            PyTokenKind::BracketR => false,
+            PyTokenKind::BraceL => false,
+            PyTokenKind::BraceR => false,
+            PyTokenKind::Pipe => false,
+            PyTokenKind::Int => true,
```
1st1: Wouldn't it be better to just match only the ones that need `true` / `unreachable()` and let the rest fall through to `false`?

tests/test_http_graphql_query.py
```diff
@@ -2794,6 +2796,19 @@ def test_graphql_functional_variables_39(self):
...
+                r"Only scalar defaults are allowed\. "
+                r"Variable 'val' has non-scalar default value\."):
+            self.graphql_query(r"""
+                query($val: FilterFloat = {eq: 3.0}) {
+                    User(filter: {score: $val}) {
+                        id,
+                    }
+                }
+            """)
+
```
1st1: Should we add a couple of functional end to end tests here to test the new functioality?  @vpetrovykh

edb/graphql/translator.py
tailhook: I've put everything that involves values. Listing keys does seem harmless. So kinda, graphql code is expected to never do that. And if we debugging by listing somewhere inside the executor, it should be fine.

edb/graphql/translator.py
tailhook: Well, The first doesn't fit 79 chars as far as I remember. I don't find second more readable (the method name and argument name are much farther from each other).

edb/graphql/translator.py
tailhook: There was discussion, and they rejected it:

https://webcache.googleusercontent.com/search?q=cache:RKvRoXUpQEgJ:https://github.com/graphql-python/graphql-core/issues/171+&cd=1&hl=en&ct=clnk&gl=ua&client=firefox-b-d

Can open another issue in the new repo, but I'm not sure we have to wait for that to be patched and released to make progress on this PR.

edb/graphql/translator.py
tailhook: This is caught by `test_graphql_functional_variables_03` and `test_graphql_functional_variables_10`.

graphql-rewrite/src/common.rs
tailhook: Well, this is contributed code (to graphql-parser).

edb/graphql/translator.py
tailhook: Changed to temporary var.

graphql-rewrite/src/entry_point.rs
tailhook: Fallthrough is bad, because if you add another token, you must carefully consider if it needs true or false here. And compiler would not help you find places where there is a fall through branch.

tests/test_http_graphql_query.py
tailhook: We have tests of the rewrite in the rust. And current functional tests have both constants in the graphql and errors on that constants. So I feel that tests are fine as they are.

graphql-rewrite/src/entry_point.rs
tailhook: Thanks

tailhook:
Rebased and moved to `edb/graphql-rewrite`

edb/graphql/translator.py
1st1: > I don't find second more readable (the method name and argument name are much farther from each other).

FWIW it's the current code style we're using everywhere for code like this.

graphql-rewrite/src/entry_point.rs
1st1: Understood.

tailhook:
Extracted all traces of `graphql-parser` code to [its own repo](https://github.com/edgedb/graphql-parser).

PR #1300, opened 2020-03-25
Document Dump/Restore protocols and a file format

tailhook:
Also changed `DUMP_PROTO_VER` constant name to `DUMP_FORMAT_VER` as it looks like it is actually a file format version not the network protocol (the constant is internal, though).


tailhook:
Added also dump file format description

tailhook:
Added restore protocol description

docs/internals/dump_format.rst
```diff
@@ -0,0 +1,135 @@
+Dump File Format
+================
+
+This description uses the same :ref:`conventions <ref_protocol_conventions>`
```
1st1: We should preface this with a warning that the dump format isn't yet stable and might change before 1.0.

tailhook: Added a warning

docs/internals/dump_format.rst
```diff
@@ -0,0 +1,137 @@
...
+as the protocol description.
+
+.. warning:: Dump file format may change until EdgeDB 1.0
```
1st1, line 7: Better wording:

```suggestion
.. warning:: The dump file format is not final and is subject to change before EdgeDB 1.0.
```

docs/internals/dump_format.rst
```diff
@@ -0,0 +1,137 @@
...
+        int8            mtype;
+
+        // SHA1 hash sum of block data
+        byte            sha1sum[20];
+
+        // Length of message contents in bytes,
+        // including self.
+        int32           message_length;
+
+        // Data itself. Can be treated in opaque way by a client.
```
1st1, line 39:
```suggestion
        // Data itself. Should be treated in opaque way by a client.
```

docs/internals/dump_format.rst
```diff
@@ -0,0 +1,137 @@
...
+        byte            data[message_length];
+    }
+
+Usually internal structure of the block doesn't matter, all client
+need to do is:
```
1st1, line 44:
```suggestion
Upon receiving a protocol dump data message, the dump client should:
```

docs/internals/dump_format.rst
```diff
@@ -0,0 +1,137 @@
...
+
+* Replace packet type:
+    * ``@`` (0x40) → ``H`` (0x48)
+    * ``=`` (0x3d) → ``D`` (0x44)
+* Prepend SHA1 checksum to the block
+* Write/read the original data bytes
+  (excluding packet type byte but including the length)
+
```
1st1, line 52:
```suggestion
* Append the entire dump protocol message disregarding the
  first byte (the message type).

```

PR #1301, opened 2020-03-25
Reimplement schema object shell mechanism

elprans:
There are cases when schema objects cannot be referred to directly from a
schema delta: forward references and deltas resulting from schema 
comparison.  In these cases we replace schema object references with 
isntances of `ObjectRef` (or its numerous subclasses) that are a 
schema-independent way of describing a schema object in *some* schema. For
most objects it's just a name.

The current implementation of `ObjectRef` is problematic because it tries
to masquerade as a regular schema `Object` (mostly for historical typing
reasons), but since it's not a real object, the schema code is peppered
with checks in places where `ObjectRef` is not expected, and, since it
masquerades as an `Object` the typing doesn't help much (or, worse, gets in
the way).

Luckily, with the current state of things, the use cases for object shells
are much more limited, and the above mentioned masquerading is not
necessary.  This commit adds a new implementation of `ObjectShell`
(and `TypeShell`, etc.) that does the necessary minimum in a much cleaner
way.

(The PR contains multiple independent commits, please review separately.)

### DIFF
--- a/edb/schema/inheriting.py
+++ b/edb/schema/inheriting.py
@@ -373,8 +373,7 @@ class InheritingObjectCommand(sd.ObjectCommand[so.InheritingObjectT]):
 
         base_refs: List[so.InheritingObjectT] = []
         for b in getattr(astnode, 'bases', None) or []:
-            obj = utils.ast_to_typeref(b, modaliases=modaliases, schema=schema)
-            # assert isinstance(obj, so.InheritingObject)
+            obj = utils.ast_to_type(b, modaliases=modaliases, schema=schema)
             base_refs.append(cast(so.InheritingObjectT, obj))
 
         return cls._validate_base_refs(schema, base_refs, astnode, context)
--- a/edb/schema/inheriting.py
+++ b/edb/schema/inheriting.py
@@ -680,7 +680,7 @@ class CreateInheritingObject(
                 base_names.append(b.name)
         else:
             assert isinstance(bases, so.ObjectList)
-            base_names = list(bases.names(schema, allow_unresolved=True))
+            base_names = list(bases.names(schema))
 
         # Filter out implicit bases
         explicit_bases = [
--- a/edb/schema/inheriting.py
+++ b/edb/schema/inheriting.py
@@ -272,11 +272,7 @@ class InheritingObjectCommand(sd.ObjectCommand[so.InheritingObjectT]):
 
             obj = schema.get(cmd.classname, default=None)
             if obj is None:
-                cmd.set_attribute_value(
-                    refdict.backref_attr,
-                    so.ObjectRef(name=scls.get_name(schema)),
-                )
-
+                cmd.set_attribute_value(refdict.backref_attr, scls)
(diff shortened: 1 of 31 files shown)


edb/schema/scalars.py
```diff
@@ -269,20 +274,28 @@ def _cmd_tree_from_ast(
...
         bases = create_cmd.get_attribute_value('bases')
-        is_enum = False
-        if len(bases) == 1 and isinstance(bases._ids[0], AnonymousEnumTypeRef):
-            # type ignore below because this class elements is set
-            # directly on __dict__
-            elements = bases._ids[0].elements  # type: ignore
+        is_enum = (
+            len(bases) == 1
+            and bases.first(schema).get_name(schema) == 'std::anyenum'
+        )
```
1st1: What if there are more bases and one of them is an enum? The variable name `is_enum` is confusing; add a comment?

edb/schema/inheriting.py
```diff
@@ -473,38 +474,38 @@ def _apply_field_ast(
...
     old_bases: Iterable[str], new_bases: Iterable[str]
 ) -> Tuple[
-    Tuple[so.ObjectRef, ...],
+    Tuple[so.ObjectShell, ...],
     Tuple[
         Tuple[
-            List[so.ObjectRef],
-            Union[str, so.ObjectRef, Tuple[str, so.ObjectRef]],
+            List[so.ObjectShell],
+            Union[str, so.ObjectShell, Tuple[str, so.ObjectShell]],
```
1st1: Probably should use type aliases to make type signatures like this easier to digest

edb/schema/inheriting.py
```diff
@@ -636,49 +637,64 @@ def _apply_field_ast(
...
+        context: sd.CommandContext,
+        bases: Any,
+    ) -> List[str]:
+
+        mcls = self.get_schema_metaclass()
+        default_base = mcls.get_default_base_name()
+
+        if isinstance(bases, so.ObjectCollectionShell):
+            base_names = [
+                b.name for b in bases.items if b.name is not None
```
1st1: What does `if b.name is not None` check guard from?

edb/schema/links.py
```diff
@@ -317,10 +317,10 @@ def _apply_field_ast(
                         node.target = expr.qlast
                     else:
                         t = op.new_value
-                        assert isinstance(t, so.Object)
+                        assert isinstance(t, (so.Object, so.ObjectShell))
```
1st1: In places like this, don't you want to explicitly resolve `ObjectShell` into an `Object` and then pass it down?

elprans: Not really, because the delta you're decompiling to DDL might not have actually been applied, so the object isn't necessarily in the schema.

edb/schema/scalars.py
elprans: This is rewritten in a subsequent commit.  At this particular point the check is done in `_validate_base_refs()` earlier.

PR #1303, opened 2020-03-26
Switch to Rust CLI by default

elprans:
Devmode install now builds and uses `edgedb/edgedb-cli` by default.


haikyuu:
How can we use the dev build? Is it the nightly version available in apt (ubuntu)?

elprans:
@haikyuu The nightly build is available in the `.nightly` repo:

```
echo deb https://packages.edgedb.com/apt bionic.nightly main \
    | sudo tee /etc/apt/sources.list.d/edgedb.list
```

There is also the `:nightly` docker image.

PR #1310, opened 2020-03-27
Add annotations for schema functions

RobertoPrevato:
Adds annotations for schema functions, including a small refactoring of how `get_shortname` is used in `Parameter` class: its signature is made consistent with the one inherited from `QualifiedObject`, and replaces the places where it is used with a new function `get_parameter_name`.

### DIFF
--- a/edb/schema/functions.py
+++ b/edb/schema/functions.py
@@ -703,20 +703,6 @@ class CallableCommand(sd.QualifiedObjectCommand[CallableObject]):
             params.append(param)
         return FuncParameterList.create(schema, params)
 
-    def _alter_innards(
-        self,
-        schema: s_schema.Schema,
-        context: sd.CommandContext,
-    ) -> s_schema.Schema:
-        # type ignore below, because CallableCommand is used as mixin,
-        # in classes where super()._alter_innards is bound
-        schema = super()._alter_innards(schema, context)  # type: ignore
-
-        for op in self.get_subcommands(metaclass=Parameter):
-            schema = op.apply(schema, context=context)
-
-        return schema
-
     @classmethod
     def _get_param_desc_from_ast(
         cls,
@@ -754,6 +740,23 @@ class CallableCommand(sd.QualifiedObjectCommand[CallableObject]):
             params.append(param)
 
         return schema, params
+
+
+class AlterCallableObject(CallableCommand,
+                          sd.AlterObject[CallableObject]):
+    astnode = qlast.AlterCallableObject
+
+    def _alter_innards(
+        self,
+        schema: s_schema.Schema,
+        context: sd.CommandContext,
+    ) -> s_schema.Schema:
+        schema = super()._alter_innards(schema, context)
+
+        for op in self.get_subcommands(metaclass=Parameter):
+            schema = op.apply(schema, context=context)
+
+        return schema
 
 
 class CreateCallableObject(CallableCommand, sd.CreateObject[CallableObject]):
--- a/edb/schema/functions.py
+++ b/edb/schema/functions.py
@@ -46,7 +46,7 @@ if TYPE_CHECKING:
 
 def param_as_str(
     schema: s_schema.Schema,
-    param: Union[ParameterDesc, Parameter]
+    param: Union[ParameterDesc, Parameter],
 ) -> str:
     ret = []
     kind = param.get_kind(schema)
@@ -134,7 +134,7 @@ class ParameterDesc(ParameterLike):
         schema: s_schema.Schema,
         modaliases: Mapping[Optional[str], str],
         num: int,
-        astnode: qlast.FuncParam
+        astnode: qlast.FuncParam,
     ) -> ParameterDesc:
 
         paramd = None
@@ -199,7 +199,7 @@ class ParameterDesc(ParameterLike):
         cls,
         schema: s_schema.Schema,
         context: sd.CommandContext,
-        cmd: CreateParameter
+        cmd: CreateParameter,
     ) -> Tuple[s_schema.Schema, ParameterDesc]:
         props = cmd.get_attributes(schema, context)
         props['name'] = Parameter.paramname_from_fullname(props['name'])
@@ -221,7 +221,7 @@ class ParameterDesc(ParameterLike):
         schema: s_schema.Schema,
         func_fqname: sn.Name,
         *,
-        context: sd.CommandContext
+        context: sd.CommandContext,
     ) -> sd.Command:
         CreateParameter = sd.ObjectCommandMeta.get_command_class_or_die(
             sd.CreateObject, Parameter)
@@ -250,7 +250,7 @@ class ParameterDesc(ParameterLike):
         schema: s_schema.Schema,
         func_fqname: sn.Name,
         *,
-        context: sd.CommandContext
+        context: sd.CommandContext,
     ) -> sd.Command:
         DeleteParameter = sd.ObjectCommandMeta.get_command_class_or_die(
             sd.DeleteObject, Parameter)
@@ -309,7 +309,7 @@ class Parameter(so.ObjectFragment, ParameterLike):
         self,
         schema: s_schema.Schema,
         *,
-        with_parent: bool=False
+        with_parent: bool = False,
     ) -> str:
         vn = super().get_verbosename(schema)
         if with_parent:
@@ -406,7 +406,7 @@ class PgParams(NamedTuple):
     def from_params(
         cls,
         schema: s_schema.Schema,
-        params: Union[Sequence[ParameterLike], ParameterLikeList]
+        params: Union[Sequence[ParameterLike], ParameterLikeList],
     ) -> PgParams:
         pg_params = []
         named = []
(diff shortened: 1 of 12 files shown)


edb/schema/functions.py
```diff
@@ -237,7 +244,7 @@ def as_delete_delta(
         cmd = DeleteParameter(classname=param_name)
 
         if self.type.is_collection() and not self.type.is_polymorphic(schema):
-            param = schema.get(param_name)
+            param: s_types.Type = schema.get(param_name)
```
elprans:
```suggestion
            param = schema.get(param_name, type=Parameter)
```

Also, the annotation of `parent` `cleanup_schema_collection()` is incorrect: it should be `Object`, not `Type`.

edb/schema/functions.py
```diff
@@ -291,20 +298,28 @@ def get_verbosename(
         else:
             return vn
 
-    def get_shortname(self, schema: s_schema.Schema) -> str:
+    # type ignore below, because `get_shortname` in objects.QualifiedObject
+    # has return type of sn.Name; the following method instead *needs* to
+    # return a str
+    def get_shortname(self, schema: s_schema.Schema) -> str:  # type: ignore
```
elprans: This doesn't seem to be sound. Can you try removing this implementation and use the default one instead?

edb/schema/functions.py
```diff
@@ -313,27 +328,37 @@ def as_str(self, schema: s_schema.Schema) -> str:
...
 
 
-class ParameterCommand(referencing.StronglyReferencedObjectCommand,
-                       schema_metaclass=Parameter,
-                       context_class=ParameterCommandContext,
-                       referrer_context_class=CallableCommandContext):
+# type ignore below, because adding [Parameter] and consequently making
+# Parameter a referencing.ReferencedObject breaks the code
+class ParameterCommand(
+    referencing.StronglyReferencedObjectCommand,  # type: ignore
```
elprans: Can you still add the `[Parameter]` subscript and `type: ignore` that?

edb/schema/functions.py
```diff
@@ -667,17 +666,27 @@ def is_blocking_ref(
...
-    def _get_params(self, schema: s_schema.Schema, context):
+    def _get_params(
+        self,
+        schema: s_schema.Schema,
+        context: sd.CommandContext
+    ) -> so.ObjectList[Parameter]:
         params = []
         for cr_param in self.get_subcommands(type=ParameterCommand):
-            param = schema.get(cr_param.classname)
+            param: Parameter = schema.get(cr_param.classname)
```
elprans:
```suggestion
            param = schema.get(cr_param.classname, type=Parameter)
```

This spelling is better because it also performs runtime checking of the actual returned object.

edb/schema/functions.py
```diff
@@ -667,17 +666,27 @@ def is_blocking_ref(
...
             params.append(param)
         return FuncParameterList.create(schema, params)
 
-    def _alter_innards(self, schema: s_schema.Schema, context):
-        schema = super()._alter_innards(schema, context)
+    def _alter_innards(
```
elprans: Can you define an `AlterCallableObject` and move this there instead?

edb/schema/functions.py
```diff
@@ -1021,6 +1063,8 @@ def _create_begin(self, schema: s_schema.Schema, context):
                     f'invalid default value {p_default.text!r} of parameter '
                     f'{p.get_displayname(schema)!r}: {ex}',
                     context=self.source_context)
+            # 024
```
elprans: What does this comment refer to?

edb/schema/functions.py
```diff
@@ -1052,8 +1096,15 @@ def _create_begin(self, schema: s_schema.Schema, context):
...
     @classmethod
-    def _cmd_tree_from_ast(cls, schema: s_schema.Schema, astnode, context):
+    def _cmd_tree_from_ast(
+        cls,
+        schema: s_schema.Schema,
+        astnode: qlast.DDLOperation,
+        context: sd.CommandContext
+    ) -> sd.Command:
         cmd = super()._cmd_tree_from_ast(schema, astnode, context)
+        # 002
```
elprans: Likewise

edb/schema/functions.py
RobertoPrevato: Sorry, this is debris (placeholders I occasionally put while analyzing the code - usually I don't forget those).

edb/schema/functions.py
RobertoPrevato: Like the other - removing.

edb/schema/functions.py
RobertoPrevato: I looked into this for a while, and I would like your guidance to know the approach you prefer.
The `get_shortname` from the ancestor `QualifiedObject` is annotated and designed to return a `sn.SchemaName`: all callers are using exactly this type and wouldn't work with a regular `str`.
The `get_shortname` from the "interface" `ParameterLike` is annotated to return a `str`. The `Parameter` class is a descendant of both classes, and it implements the signature defined in `ParameterLike`, in fact a `sn.Name` is unwrapped into a `str` inside the method `paramname_from_fullname`.

There are currently 7 annotated functions `get_shortname`: 5 of them are annotated to return `str`, and 2 of them to return a more specialised `SchemaName` (QualifiedObject and EphemeralCollection). Both the functions in `QualifiedObject` and `Parameter` class are used by callers that rely on `SchemaName` and `str` respectively.

To make the functions consistent, we can for example refactor the code in `ParameterLike` to return `SchemaName`. In such case, how the following function should be modified? Would it make sense to return a SchemaName having `module` property equal to `fullname` input parameter?

```python
    @classmethod
    def paramname_from_fullname(cls, fullname: sn.Name) -> str:
        parts = str(fullname.name).split('@@', 1)
        if len(parts) == 2:
            return sn.unmangle_name(parts[0])
        elif '::' in fullname:
            return sn.Name(fullname).name
        else:
            return fullname
```

elprans: The callers that rely on `Parameter.get_shortname` returning `str` should call another function that wraps `get_shortname`, and `Parameter.get_shortname` should be removed (i.e. revert to the implementation in `QualifiedObject`):

```python
class Parameter:
    ...
    def get_parameter_name(self, schema) -> str:
        return self.get_shortname(schema).name
```

edb/schema/functions.py
elprans: Don't forget to actually *use* the newly defined class by extending from it in the appropriate `AlterFunction` and `AlterOperator`

edb/schema/functions.py
RobertoPrevato: Thanks Elvis for the instructions - I was in doubt earlier that you indeed meant this.
I worked in that direction, you can see in [this commit](https://github.com/edgedb/edgedb/commit/00af9ca4b1a32737b3870b941955aa4dc4b54a56) the files I had to change to be able to run the tests (to get the test database populated properly). Several tests are failing, and I get several of these:

```
    await self._protocol.simple_query(query)
  File "edgedb/protocol/protocol.pyx", line 497, in simple_query
edgedb.errors.InvalidReferenceError: improperly formed name: module is not specified: a
```

To find out when the old function is called, I added a function that logs the trace whenever the `Parameter.get_shortname` function is used; and it seems the only place where it is still used is from `/edgedb/edb/schema/objects.py", line 656, in get_displayname
    return str(self.get_shortname(schema))` - but of course I don't want to change this piece of code that is rightly abstracted from `Parameter` class to do something else. But maybe I only need to modify still the `get_verbosename` in Functions. Just writing this is giving me a bit more of insight. :smile:

elprans: Right. `Parameter` name generation is a bit backwards.  Here's how I propose to fix this:

```python
    def get_shortname(self, schema: s_schema.Schema) -> sn.Name:
        return sn.Name(
            module=self.get_name(schema).module,
            name=self.get_parameter_name(schema),
        )

    def get_displayname(self, schema: s_schema.Schema) -> str:
        return self.get_parameter_name(schema)
```

RobertoPrevato: It looks good now!

edb/schema/functions.py
```diff
@@ -747,7 +749,24 @@ def _get_param_desc_from_delta(
         return schema, params
 
 
-class CreateCallableObject(CallableCommand, sd.CreateObject):
+class AlterCallableObject(CallableCommand,
+                          sd.AlterObject[CallableObject]):
+    astnode = qlast.AlterCallableObject
```
elprans, line 754: This is an abstract class, there's no need to bind it to a particular AST node.

edb/schema/functions.py
```diff
@@ -1152,7 +1224,7 @@ class RenameFunction(sd.RenameObject, FunctionCommand):
     pass
 
 
-class AlterFunction(sd.AlterObject, FunctionCommand):
+class AlterFunction(sd.AlterObject[Function], FunctionCommand):
```
elprans, line 1227: `AlterCallableObject[Function]`.   Also, `AlterOperator` should inherit from `AlterCallableObject` as well.

PR #1315, opened 2020-03-31
Always derive collection types via schema

elprans:
Currently, collection types (`Tuple` and `Array`) are a weird special
case, whereby collection type instances are normally not tracked by a
schema, except when used by another schema object.  This duality
requires a complex web of classes with lots of hacks to circumvent
typing violations.  The complexity of this implementation far outweighs
any possible benefit of deriving ephemeral collection types outside of the
schema.

This commit reworks the whole thing so that there are now only one class
for each collection kind with the usual traits of the schema object.
The changes are mostly about changing a bunch of functions that may
derive new collection types to return the modified schema object.

### DIFF
--- a/edb/edgeql/compiler/__init__.py
+++ b/edb/edgeql/compiler/__init__.py
@@ -505,6 +505,7 @@ def compile_func_to_ir(
     )
 
     assert isinstance(ir, irast.Statement)
+    schema = ir.schema
 
     return_type = func.get_return_type(schema)
     if (not ir.stype.issubclass(schema, return_type)
--- a/edb/ir/typeutils.py
+++ b/edb/ir/typeutils.py
@@ -101,6 +101,18 @@ def is_abstract(typeref: irast.TypeRef) -> bool:
     return typeref.is_abstract
 
 
+def is_persistent_tuple(typeref: irast.TypeRef) -> bool:
+    if is_tuple(typeref):
+        if typeref.material_type is not None:
+            material = typeref.material_type
+        else:
+            material = typeref
+
+        return material.in_schema
+    else:
+        return False
+
+
 def type_to_typeref(
     schema: s_schema.Schema,
     t: s_types.Type,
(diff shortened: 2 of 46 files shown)


edb/edgeql/compiler/__init__.py
```diff
@@ -505,6 +505,7 @@ def compile_func_to_ir(
     )
 
     assert isinstance(ir, irast.Statement)
+    schema = ir.schema
```
1st1, line 508: Ha, I'm wondering why we haven't detected this before

edb/ir/typeutils.py
```diff
@@ -101,6 +101,18 @@ def is_abstract(typeref: irast.TypeRef) -> bool:
     return typeref.is_abstract
 
 
+def is_persistent_tuple(typeref: irast.TypeRef) -> bool:
```
1st1, line 104: Why not `is_material_tuple`? What's the difference between "material" and "persistent"?

edb/schema/types.py
```diff
@@ -902,8 +983,16 @@ def create_shell(
         typemods: Any = None,
         name: Optional[str] = None,
     ) -> ArrayTypeShell:
+        if not typemods:
```
1st1, line 986: Do we even use typemods for arrays? Isn't it easier to remove the parameter?

edb/edgeql/compiler/__init__.py
elprans: Because functions cannot produce schema objects other than themselves and collection types, which were previously handled differently.

edb/ir/typeutils.py
elprans: "material" means non-derived, i.e. not an ephemeral subtype derived for compilation purposes. "persistent" means that the type is used in schema objects (other than transiently in some expression), and will be loaded with schema introspection.

edb/schema/types.py
elprans: I'd thought about cleaning this up later. Simply removing this would break typing, so I decided not to bother for now.

edb/ir/typeutils.py
1st1: Is there a place you can document this and alike terminology?

elprans: Yes. I'll make a pass and add some docs.

PR #1317, opened 2020-04-02
Graphql rewrite stage 2

tailhook:
1. Implement Int/Float/Boolean variables extraction (excluding `first: 1` as a special case)
2. implement extraction of variable's default values
2. Extract variables from `@include(if:$var)` and `@skip(if:$var)` (previously they were detected too late, so were effectively disabling query cache)

### DIFF
--- a/edb/graphql-rewrite/src/entry_point.rs
+++ b/edb/graphql-rewrite/src/entry_point.rs
@@ -7,7 +7,8 @@ use edb_graphql_parser::query::Definition;
 use edb_graphql_parser::query::{Operation, InsertVars, InsertVarsKind};
 use edb_graphql_parser::query::{Document, parse_query, ParseError};
 use edb_graphql_parser::tokenizer::Kind::{StringValue, BlockString};
-use edb_graphql_parser::tokenizer::Kind::{IntValue, Punctuator, Name};
+use edb_graphql_parser::tokenizer::Kind::{IntValue, FloatValue};
+use edb_graphql_parser::tokenizer::Kind::{Punctuator, Name};
 use edb_graphql_parser::tokenizer::{TokenStream, Token};
 use edb_graphql_parser::common::{unquote_string, Type};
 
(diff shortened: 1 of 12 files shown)


edb/graphql/translator.py
```diff
@@ -373,10 +377,7 @@ def _should_include(self, directives):
 
                 if isinstance(cond, gql_ast.VariableNode):
                     varname = cond.name.value
-                    var = self._context.vars[varname]
-                    # mark the variable as critical
-                    self._context.vars[varname] = var._replace(critical=True)
-                    value = var.val
+                    value = self._context.vars[varname].val
```
vpetrovykh, line 380: "Critical" variables used to be part of the cache key (because they affect query shape). How is caching of different shapes due to include/skip directive variables handled now?

I tried to run queries in GraphiQL in this branch and it appears that there's no problem in correctly handling these variables, but I wonder what the mechanism is. The reason is that if this mechanism uses something other than this "critical" flag, we should probably drop the code handling it.

tailhook: Unfortunately, they weren't part of a cache key. There was (and still is) a merely if statement that always recompiled if there is a critical variables.

Currently variables from the include/exclude are present in the cache key:

tailhook: Old mechanism, which is still here:
https://github.com/edgedb/edgedb/blob/16f43cd44383373908f06eedca92923fbbb74d72/edb/server/http_graphql_port/protocol.pyx#L236
Is still used for introspection queries. I would like this to be handled in some other way, but we can do it in a separate PR.

tailhook:
> Tests are passing though, so my comment is probably non-blocking.

Well it looks like there are a whole lot of issues were hidden behind python fallback :(

The biggest issue is probably that floating point constants can be used as decimals. But if we make it as `Float!` variable it can be non-exact. Not sure what to do yet.

tailhook:
Another hidden issue is that `mutation {Object(field: {clear: true})..}` can't be substituted. Detecting `clear: ` looks weird, but looks like what we need to do :(

tailhook:
> Another hidden issue is that `mutation {Object(field: {clear: true})..}` can't be substituted. Detecting `clear: ` looks weird, but looks like what we need to do :(

Well, probably we can bring this thing as a critical variable and allow variables there... (that means fixing critical variables, also).

tailhook:
Fixed "clear" issue. Fixed issues with decimals. And fixed thing @vpetrovykh shown above.

tailhook:
@vpetrovykh do you want to do another round of review, or just merge it in?

edb/graphql-rewrite/src/entry_point.rs
```diff
@@ -145,12 +215,79 @@ pub fn rewrite(operation: Option<&str>, s: &str) -> Result<Entry, Error> {
...
 
     let mut variables = Vec::new();
+    let mut defaults = BTreeMap::new();
+    let mut key_vars = BTreeSet::new();
+    let mut value_positions = HashSet::new();
 
+    visit_directives(&mut key_vars, &mut value_positions, &oper);
+
+    for var in &oper.variable_definitions {
+        if var.name.starts_with("_edb_arg__") {
```
1st1, line 230: Please use `__{..}__` pattern as we do for other similar things -- identifiers like that are reserved for EdgeDB in EdgeQL and everywhere else.

tailhook: GraphQL also has `__xxx` pattern reserved for introspection purposes. So it chokes on that, see https://github.com/edgedb/edgedb/pull/1289

PR #1322, opened 2020-04-06
Assorted cleanups to schema (mostly around functions)

elprans:
See individual commits

### DIFF
--- a/edb/edgeql/compiler/stmt.py
+++ b/edb/edgeql/compiler/stmt.py
@@ -587,9 +587,8 @@ def init_stmt(
     ctx.stmt = irstmt
     if ctx.toplevel_stmt is None:
         parent_ctx.toplevel_stmt = ctx.toplevel_stmt = irstmt
-        ctx.path_scope = parent_ctx.path_scope
-    else:
-        ctx.path_scope = parent_ctx.path_scope.attach_fence()
+
+    ctx.path_scope = parent_ctx.path_scope.attach_fence()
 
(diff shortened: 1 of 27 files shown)


edb/edgeql/compiler/stmt.py
```diff
@@ -587,9 +587,8 @@ def init_stmt(
     ctx.stmt = irstmt
     if ctx.toplevel_stmt is None:
         parent_ctx.toplevel_stmt = ctx.toplevel_stmt = irstmt
-        ctx.path_scope = parent_ctx.path_scope
-    else:
-        ctx.path_scope = parent_ctx.path_scope.attach_fence()
+
+    ctx.path_scope = parent_ctx.path_scope.attach_fence()
```
1st1, line 591: > they would end up sharing the same scope node inappropriately.

Can we have a test for that?

edb/schema/functions.py
```diff
@@ -403,53 +430,6 @@ class DeleteParameter(ParameterCommand, sd.DeleteObject[Parameter]):
     pass
 
 
-class PgParams(NamedTuple):
```
1st1, line 406: Finally!

edb/edgeql/compiler/stmt.py
elprans: Basically all function tests are the test for that with this rework, hence the change.

tailhook:
Merged to publish a PR on top of this

PR #1323, opened 2020-04-07
Represent EdgeQL function bodies as Expressions

elprans:
Currently, function bodies are always parsed and stored as string
literals, which are only compiled when the `CreateFunction` delta
command is applied.  This has obvious downsides: validation happens too
late, and there's no dependency tracing for objects referenced in the
function body, not to mention the ugly hacks in the parser that deparse
the `USING (<expr>)` form back to a string.

Co-authored-by: Paul Colomiets (@tailhook)

### DIFF
--- a/tests/test_edgeql_calls.py
+++ b/tests/test_edgeql_calls.py
@@ -212,7 +212,7 @@ class TestEdgeQLFuncCalls(tb.QueryTestCase):
                 VARIADIC a: int64
             ) -> int64
                 USING EdgeQL $$
-                    SELECT <int64>sum(array_unpack(a));;
+                    SELECT <int64>sum(array_unpack(a))
                 $$;
         ''')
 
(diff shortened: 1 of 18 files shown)


tests/test_edgeql_calls.py
```diff
@@ -520,16 +520,16 @@ class TestEdgeQLFuncCalls(tb.QueryTestCase):
             CREATE FUNCTION test::inner(
                 a: anytype
             ) -> int64
-                USING EdgeQL $$
```
1st1, line 523: Wait, so this syntax is no longer supported?

tests/test_schema.py
```diff
@@ -3789,13 +3790,19 @@ def test_describe_01(self):
 
             r"""
             function stdgraphql::short_name(name: std::str) -> std::str {
+                orig_nativecode :=
+                    r"SELECT (((name)[5:] IF (name LIKE 'std::%') ELSE
```
1st1, line 3794: It's unfortunate we can no longer retain the original source code of functions.

elprans: We actually do store the original source, this is an artifact of parsing and codegening it.  Need more refactoring to avoid the roundtrip.

tests/test_edgeql_calls.py
elprans: It is, I just cleaned up some tests to use more concise syntax.

tailhook:
Merged to publish a branch on top of it.

PR #1324, opened 2020-04-07
Use token list in compile_eql rather than a string

tailhook:
This is a prerequisite for constant extraction PR. Not sure if this needs to be merged separately

### DIFF
--- a/edb/edgeql/parser/__init__.py
+++ b/edb/edgeql/parser/__init__.py
@@ -19,8 +19,13 @@
 
 from __future__ import annotations
 
+from typing import *
+
 from . import parser as ql_parser
 from .. import ast as qlast
+
+if TYPE_CHECKING:
+    from edb import _edgeql_rust
 
 
 def parse_fragment(expr):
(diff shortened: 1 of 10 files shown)


edb/server/compiler/compiler.py
```diff
@@ -166,7 +168,8 @@ def compile_edgeql_script(
     compiler._std_schema = std_schema
     compiler._bootstrap_mode = True
 
-    units = compiler._compile(ctx=ctx, eql=eql.encode())
+    tokens = _edgeql_rust.tokenize(eql)
+    units = compiler._compile(ctx=ctx, eql_tokens=tokens)
```
elprans, line 172: nit: just call them `tokens`, it's clear that they're EdgeQL from the context.

edb/edgeql/parser/__init__.py
```diff
@@ -19,9 +19,14 @@
 
 from __future__ import annotations
 
+from typing import TYPE_CHECKING, List
```
elprans, line 22: Why not just `from typing import *` as everywhere else?

edb/server/compiler/compiler.py
tailhook: Fixed

edb/edgeql/parser/__init__.py
tailhook: Not sure. Probably just general allergy to `*`. Fixed.

PR #1326, opened 2020-04-08
Break the schema/compiler import loop and organize compiler options

elprans:
This is a long overdue cleanup that makes `edb.edgeql.compiler` properly
importable from schema modules by making its entrypoint imports lazy.

Another cleanup organizes compilation options into a separate dataclass.
The number of options has grown rather large to keep passing them
around as regular kwargs anymore.

### DIFF
--- a/edb/edgeql/compiler/__init__.py
+++ b/edb/edgeql/compiler/__init__.py
@@ -128,55 +128,61 @@ dispatch.py
 from __future__ import annotations
 from typing import *
 
+# WARNING: this package is in a tight import loop with various modules
+# in edb.schema, so no direct imports from either this package or
+# edb.schema are allowed at the top-level.  If absolutely necessary,
+# use the lazy-loading mechanism.
+
+import functools
+
 from edb import errors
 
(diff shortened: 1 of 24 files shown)


edb/edgeql/compiler/__init__.py
```diff
@@ -128,55 +128,61 @@
...
+    irast = None
+    ireval = None
+    stmtctx_mod = None
+
 
+#: Compiler modules lazy-load guard.
+_LOADED = False
 
+
+def compiler_entrypoint(func: Callable[..., Any]) -> Callable[..., Any]:
```
1st1, line 170: Alternatively you can use a module-level `__getattr__`

1st1: I think that would be even cleaner.

elprans: I don't think `__getattr__` would help here, since we need to populate the local namespace in the module itself. It would only work if the internals were imported via a proxy module with `__getattr__` which doesn't seem cleaner.

PR #1327, opened 2020-04-08
Implement the notebook port

1st1:
This is a draft. I'm now experimenting with parsing the serialized binary proto in JS/Browser.


1st1:
@elprans This is ready for your review.

1st1:
> An aside: maybe we should name this a "notebook port" and not "tutorial port"?

Love it, done.

PR #1329, opened 2020-04-08
Implement volatility inference

elprans:
This adds the ability to infer expression volatility in the EdgeQL
compiler.  The implementation is very similar to cardinality inference.

### DIFF
--- a/edb/edgeql/compiler/inference/volatility.py
+++ b/edb/edgeql/compiler/inference/volatility.py
@@ -1,0 +1,316 @@
+#
+# This source file is part of the EdgeDB open source project.
+#
+# Copyright 2019-present MagicStack Inc. and the EdgeDB authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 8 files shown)


edb/edgeql/compiler/inference/volatility.py
```diff
@@ -0,0 +1,316 @@
+#
+# This source file is part of the EdgeDB open source project.
+#
+# Copyright 2008-present MagicStack Inc. and the EdgeDB authors.
```
vpetrovykh, line 4: This probably should be 2020 and not 2008. I wonder how much does this matter, though in terms of any real consequences. I assume it matters a little bit, otherwise why would we even bother, so it probably needs to be at least somewhat accurate.

elprans:
> I assume we're currently not inferring the volatility for EdgeQL functions and don't enforce that the declared volatility is not lower than the inferred, right?

Yes, although that step is trivial to add once this is merged.

PR #1330, opened 2020-04-09
Add annotations for schema pointers

RobertoPrevato:
Adds annotations for schema pointers, include refactoring of Pointer.derive_ref method to make it compatible with its ancestor class ReferencedObject.

### DIFF
--- a/edb/ir/typeutils.py
+++ b/edb/ir/typeutils.py
@@ -39,6 +39,7 @@ if TYPE_CHECKING:
     import uuid
 
     from edb.schema import name as s_name
+    from edb.schema import objects as s_objects
     from edb.schema import schema as s_schema
 
     PtrRefCacheKey = Tuple[s_pointers.PointerLike, s_pointers.PointerDirection]
@@ -398,6 +399,13 @@ def ptrref_from_ptrcls(
 
     ircls: Type[irast.BasePointerRef]
 
+    source_ref: Union[Optional[irast.TypeRef],
+                      Optional[s_objects.Object]]
+    target_ref: Union[Optional[irast.TypeRef],
+                      Optional[s_objects.Object]]
+    out_source: Union[Optional[irast.TypeRef],
+                      Optional[s_objects.Object]]
+
     if isinstance(ptrcls, irast.TupleIndirectionLink):
         ircls = irast.TupleIndirectionPointerRef
     elif isinstance(ptrcls, irast.TypeIntersectionLink):
@@ -414,8 +422,6 @@ def ptrref_from_ptrcls(
             s_mod.Module, name.module).id
     else:
         raise AssertionError(f'unexpected pointer class: {ptrcls}')
-
-    out_source: Optional[irast.TypeRef]
 
     target = ptrcls.get_far_endpoint(schema, direction)
     if isinstance(target, s_types.Type):
@@ -473,6 +479,7 @@ def ptrref_from_ptrcls(
         union_ptrs = set()
 
         for component in union_of.objects(schema):
+            assert isinstance(component, s_pointers.Pointer)
             material_comp = component.material_type(schema)
             union_ptrs.add(material_comp)
 
@@ -607,6 +614,7 @@ def cardinality_from_ptrcls(
         out_cardinality = None
         dir_cardinality = None
     else:
+        assert isinstance(out_card, qltypes.SchemaCardinality)
         out_cardinality = qltypes.Cardinality.from_schema_value(
             required, out_card)
         # Determine the cardinality of a given endpoint set.
--- a/edb/ir/typeutils.py
+++ b/edb/ir/typeutils.py
@@ -39,7 +39,6 @@ if TYPE_CHECKING:
     import uuid
 
     from edb.schema import name as s_name
-    from edb.schema import objects as s_objects
     from edb.schema import schema as s_schema
 
     PtrRefCacheKey = Tuple[s_pointers.PointerLike, s_pointers.PointerDirection]
@@ -399,12 +398,9 @@ def ptrref_from_ptrcls(
 
     ircls: Type[irast.BasePointerRef]
 
-    source_ref: Union[Optional[irast.TypeRef],
-                      Optional[s_objects.Object]]
-    target_ref: Union[Optional[irast.TypeRef],
-                      Optional[s_objects.Object]]
-    out_source: Union[Optional[irast.TypeRef],
-                      Optional[s_objects.Object]]
+    source_ref: Optional[irast.TypeRef]
+    target_ref: Optional[irast.TypeRef]
+    out_source: Optional[irast.TypeRef]
 
     if isinstance(ptrcls, irast.TupleIndirectionLink):
         ircls = irast.TupleIndirectionPointerRef
@@ -424,7 +420,8 @@ def ptrref_from_ptrcls(
         raise AssertionError(f'unexpected pointer class: {ptrcls}')
 
     target = ptrcls.get_far_endpoint(schema, direction)
-    if isinstance(target, s_types.Type):
+    if target is not None and not isinstance(target, irast.TypeRef):
+        assert isinstance(target, s_types.Type)
         target_ref = type_to_typeref(schema, target, cache=typeref_cache)
     else:
         target_ref = target
@@ -443,8 +440,11 @@ def ptrref_from_ptrcls(
         )
         source_ref = None
     else:
-        if isinstance(source, s_types.Type):
-            source_ref = type_to_typeref(schema, source, cache=typeref_cache)
+        if source is not None and not isinstance(source, irast.TypeRef):
+            assert isinstance(source, s_types.Type)
+            source_ref = type_to_typeref(schema,
+                                         source,
+                                         cache=typeref_cache)
         else:
             source_ref = source
         source_ptr = None
--- a/edb/schema/pointers.py
+++ b/edb/schema/pointers.py
@@ -380,7 +380,7 @@ class Pointer(referencing.ReferencedInheritingObject,
             ptr = schema.get(fqname, default=None)
             if ptr is None:
                 schema, ptr = self.derive_ref(
-                    schema, source, target,
+                    schema, source, target=target,
                     derived_name_base=derived_name_base, **kwargs)
         assert isinstance(ptr, Pointer)
         return schema, ptr
@@ -389,18 +389,17 @@ class Pointer(referencing.ReferencedInheritingObject,
         shortname = self.get_shortname(schema)
         return sn.Name(module='__', name=shortname.name)
 
-    def derive_ref(  # type: ignore
+    def derive_ref(
         self,
         schema: s_schema.Schema,
-        source: so.QualifiedObject,
-        target: Optional[s_types.Type] = None,
+        referrer: so.QualifiedObject,
         *qualifiers: str,
+        target: Optional[s_types.Type] = None,
         mark_derived: bool = False,
         attrs: Optional[Dict[str, Any]] = None,
         dctx: Optional[sd.CommandContext] = None,
         **kwargs: Any,
     ) -> Tuple[s_schema.Schema, Pointer]:
-
         if target is None:
             if attrs and 'target' in attrs:
                 target = attrs['target']
@@ -410,11 +409,11 @@ class Pointer(referencing.ReferencedInheritingObject,
         if attrs is None:
             attrs = {}
 
-        attrs['source'] = source
+        attrs['source'] = referrer
         attrs['target'] = target
 
         return super().derive_ref(
-            schema, source, mark_derived=mark_derived,
+            schema, referrer, mark_derived=mark_derived,
             dctx=dctx, attrs=attrs, **kwargs)
 
     def is_pure_computable(self, schema: s_schema.Schema) -> bool:
@@ -1045,14 +1044,6 @@ class PointerCommand(
         elif target_ref is not None:
             self._set_pointer_type(schema, astnode, context, target_ref)
 
-#    @classmethod
-#    def _extract_union_operands(cls, expr, operands):
-#        if expr.op == 'UNION':
-#            cls._extract_union_operands(expr.op_larg, operands)
-#            cls._extract_union_operands(expr.op_rarg, operands)
-#        else:
-#            operands.append(expr)
-#
     def compile_expr_field(
         self,
         schema: s_schema.Schema,
@@ -1068,8 +1059,9 @@ class PointerCommand(
             anchors: Dict[str, Any] = {}
 
             if field.name == 'expr':
+                # type ignore below, because the class is used as mixin
                 parent_ctx = context.get_ancestor(
-                    s_sources.SourceCommandContext,   # type: ignore
+                    s_sources.SourceCommandContext,  # type: ignore
                     self
                 )
                 assert parent_ctx is not None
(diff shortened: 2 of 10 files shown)


edb/schema/pointers.py
```diff
@@ -241,30 +255,50 @@ def get_verbosename(self, schema, *, with_parent: bool=False) -> str:
...
             return self
         else:
             return non_derived_parent
 
-    def get_near_endpoint(self, schema, direction):
+    def get_near_endpoint(
+        self,
+        schema: s_schema.Schema,
+        direction: PointerDirection,
+    ) -> Optional[so.Object]:
```
RobertoPrevato, line 269: I have doubts that I am annotating properly `get_near_endpoint` and `get_far_endpoint` (Optional[so.Object] instead of Optional[s_types.Type]), but since in the current code, `source` and `target` can be swapped depending on the pointer direction; and since they are already annotated as schema fields (`Object` and `Type), I am not sure whether SchemaFields should be modified, or these functions return type annotation made more high-level as I opted for.

edb/schema/pointers.py
```diff
@@ -335,27 +380,26 @@ def get_derived(self, schema, source, target, *,
...
 
     def derive_ref(
-        self: referencing.ReferencedT,
+        self,
         schema: s_schema.Schema,
-        source: s_sources.Source,
-        target: Optional[s_types.Type] = None,
+        referrer: so.QualifiedObject,
         *qualifiers: str,
+        target: Optional[s_types.Type] = None,
```
RobertoPrevato, line 397: To make the function signature compatible with the one defined in the ancestor `ReferencedObject`, I opted for moving the `target` argument among the keyword arguments. I think this is the most reasonable solution since we are using `**kwargs` here.

edb/schema/pointers.py
```diff
@@ -917,54 +1044,66 @@ def _process_create_or_alter_ast(self, schema, astnode, context):
         elif target_ref is not None:
             self._set_pointer_type(schema, astnode, context, target_ref)
 
-    @classmethod
-    def _extract_union_operands(cls, expr, operands):
```
RobertoPrevato, line 921: I removed this function because it's not used in the whole test suite, and even from the typing point of view it was giving an hard time. Is it correct to remove it?

edb/ir/typeutils.py
```diff
@@ -398,6 +399,13 @@ def ptrref_from_ptrcls(
 
     ircls: Type[irast.BasePointerRef]
 
+    source_ref: Union[Optional[irast.TypeRef],
```
elprans, line 402: I'm not sure this is correct. The `_ref` variables cannot be anything other than a `TypeRef`.  I suggest flipping the `if isinstance(target, s_type.Type)` check below to check for `TypeRef` and adding `assert isinstance(s_types.Type)`

```python
   if not isinstance(target, irast.TypeRef):
        assert isinstance(target, s_types.Type)
        target_ref = type_to_typeref(schema, target, cache=typeref_cache)
    else:
        target_ref = target
```

edb/schema/pointers.py
```diff
@@ -203,12 +210,12 @@ class Pointer(referencing.ReferencedInheritingObject,
         merge_fn=merge_cardinality)
 
     union_of = so.SchemaField(
-        so.ObjectSet,
+        so.ObjectSet[so.Object],
```
elprans, line 213:
```suggestion
        so.ObjectSet['Pointer'],
```

edb/schema/pointers.py
```diff
@@ -203,12 +210,12 @@ class Pointer(referencing.ReferencedInheritingObject,
...
         default=None,
         coerce=True)
 
     intersection_of = so.SchemaField(
-        so.ObjectSet,
+        so.ObjectSet[so.Object],
```
elprans, line 218:
```suggestion
        so.ObjectSet['Pointer'],
```

edb/schema/pointers.py
elprans: If the pointer is a link property, then the near endpoint would be a `Link` instance, which isn't a `Type`, so this is correct.

edb/schema/pointers.py
```diff
@@ -241,30 +255,50 @@ def get_verbosename(self, schema, *, with_parent: bool=False) -> str:
...
         if direction == PointerDirection.Outbound:
             return self.get_target(schema)
         else:
             return self.get_source(schema)
 
-    def set_target(self, schema, target):
+    def set_target(
+        self,
+        schema: s_schema.Schema,
+        target: s_objtypes.ObjectType,
```
elprans, line 288: The correct annotation for `target` is `s_types.Type`.

edb/schema/pointers.py
```diff
@@ -322,8 +359,16 @@ def merge_targets(cls, schema, ptr, t1, t2, *, allow_contravariant=False):
 
             return schema, current_target
 
-    def get_derived(self, schema, source, target, *,
-                    derived_name_base=None, **kwargs):
+    def get_derived(
+        self,
+        schema: s_schema.Schema,
+        source: so.Object,
```
elprans, line 365:
```suggestion
        source: sources.Source,
```

edb/schema/pointers.py
```diff
@@ -452,7 +499,11 @@ def allow_ref_propagation(
         constext: sd.CommandContext,
         refdict: so.RefDict,
     ) -> bool:
-        return not self.get_source(schema).is_view(schema)
+        from edb.schema import objtypes as s_objtypes
+
+        object_type = self.get_source(schema)
+        assert isinstance(object_type, s_objtypes.ObjectType)
```
elprans, line 505: `s_types.Type` would be enough, you don't need to go as narrow as `ObjectType` here

edb/schema/pointers.py
```diff
@@ -498,30 +552,44 @@ def get_is_derived(self, schema: s_schema.Schema) -> bool:
     def get_is_local(self, schema: s_schema.Schema) -> bool:
         return True
 
-    def get_union_of(self, schema):
+    def get_union_of(
+        self,
+        schema: s_schema.Schema,
+    ) -> Optional[so.ObjectSet[s_types.Type]]:
```
elprans, line 558:
```suggestion
    ) -> Optional[so.ObjectSet[PointerLike]]:
```

edb/schema/pointers.py
```diff
@@ -643,30 +715,35 @@ def _parse_computable(
         expr: qlast.Base,
         schema: s_schema.Schema,
         context: sd.CommandContext,
-    ) -> Tuple[s_schema.Schema, s_types.Type, Optional[Pointer]]:
+    ) -> Tuple[s_schema.Schema, s_types.Type, Union[None,
```
elprans, line 718: `Optional[PointerLike]`

edb/schema/pointers.py
```diff
@@ -847,13 +968,17 @@ def _process_create_or_alter_ast(self, schema, astnode, context):
             self.set_attribute_value('cardinality', astnode.cardinality)
 
         parent_ctx = self.get_referrer_context(context)
+        assert parent_ctx is not None
```
elprans, line 971:
```suggestion
        parent_ctx = self.get_referrer_context_or_die(context)
```

edb/schema/pointers.py
elprans: Yeah, seems dead

edb/schema/pointers.py
```diff
@@ -917,54 +1044,66 @@ def _process_create_or_alter_ast(self, schema, astnode, context):
...
                 options=qlcompiler.CompilerOptions(
                     modaliases=context.modaliases,
                     schema_object_context=self.get_schema_metaclass(),
                     anchors=anchors,
                     path_prefix_anchor=path_prefix_anchor,
-                    singletons=singletons,
+                    singletons=frozenset(singletons),
                 ),
             )
+            assert isinstance(expression, s_expr.Expression)
```
elprans, line 1087: Why is this `assert` necessary?

edb/schema/sources.py
```diff
@@ -74,7 +74,7 @@ def populate_pointer_set_for_source_union(
     components: Iterable[Source],
     union: Source,
     *,
-    modname: str = '__derived__',
+    modname: Optional[str] = '__derived__',
```
elprans, line 77: This change doesn't look correct. `modname` is not `Optional` in this function (which is why it's given a default).

tests/test_type_coverage.py
```diff
@@ -245,7 +245,7 @@ def test_cqa_type_coverage_repl(self) -> None:
         self.assertFunctionCoverage(EDB_DIR / "repl", 100.0)
 
     def test_cqa_type_coverage_schema(self) -> None:
-        self.assertFunctionCoverage(EDB_DIR / "schema", 89.88)
+        self.assertFunctionCoverage(EDB_DIR / "schema", 93.99)
```
1st1, line 248: 🍻

edb/schema/sources.py
RobertoPrevato: The ambiguity that caused doubt in me, is that the `get_or_create_union_pointer` function in `pointers` was already annotated with `modname: Optional[str] = None`. This argument is not altered in `get_or_create_union_pointer`, and it is passed directly to `s_sources.populate_pointer_set_for_source_union`.

Should `get_or_create_union_pointer` be modified to have a default value as well, or  should we assert that `modname` is not None before calling the function in `sources`?

edb/schema/pointers.py
RobertoPrevato: I did a mistake here: `value` was erroneously annotated as 'Any'. I corrected by saying that `value: s_expr.Expression`.

RobertoPrevato: I probably copied the annotations from the parent class in `delta.ObjectCommand`.

edb/schema/pointers.py
RobertoPrevato: I initially annotated this way. But then I had several issues in other methods, especially `get_or_create_union_pointer`, for example due to this variable swap:

```python
    if direction is PointerDirection.Inbound:
        source, target = target, source
```

I opted for keeping the source annotated as a more high-level object for this reason. Should I keep the `source` parameter in `get_or_create_union_pointer` as `so.Object`, and then assert that source is an instance of `sources.Source` before calling `genptr.get_derived` method?

edb/schema/sources.py
elprans: OK, the other way to fix this is to make it `Optional`, default to `None` and do a runtime default to `__derived__`. (`if modname is None: modname = '__derived__'`)

edb/schema/pointers.py
RobertoPrevato: Unfortunately this doesn't work because the Type argument of `ObjectSet` must be a subtype of "edb.schema.objects.Object". The `PseudoPointer` is not even an `objects.Object`.

edb/schema/pointers.py
elprans: The `source` parameter to  `get_or_create_union_pointer` should be `Source` as well.

RobertoPrevato: Ok, in this case I propose to add a `type: ignore` in the line where they are swapped. Because MyPy complains properly there, that:

```
Incompatible types in assignment (expression has type "Type", variable has type "Source")
Incompatible types in assignment (expression has type "Source", variable has type "Type")
```

edb/ir/typeutils.py
RobertoPrevato: If I apply this change, an internal server error happens while populating the test database:

```
edb test -j4
Collected 2974/2974 tests.
Using up to 4 processes to run tests.
Populating test databases... Task exception was never retrieved
future: <Task finished name='Task-142' coro=<<coroutine without __name__>()> exception=ConnectionAbortedError()>
  File "edgedb/protocol/protocol.pyx", line 798, in connect
  File "edgedb/protocol/asyncio_proto.pyx", line 60, in wait_for_message
ConnectionAbortedError


Shutting down test cluster... OK.
  File ".../venv/bin/edb", line 11, in <module>
    load_entry_point('edgedb-server', 'console_scripts', 'edb')()
  File ".../venv/lib/python3.8/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File ".../venv/lib/python3.8/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File ".../venv/lib/python3.8/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File ".../venv/lib/python3.8/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File ".../venv/lib/python3.8/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File ".../edgedb/edb/tools/test/__init__.py", line 153, in test
    result = run()
  File ".../edgedb/edb/tools/test/__init__.py", line 265, in _run
    result = test_runner.run(suite)
  File ".../edgedb/edb/tools/test/runner.py", line 690, in run
    tb.setup_test_cases(
  File ".../edgedb/edb/testbase/server.py", line 838, in setup_test_cases
    return asyncio.run(_run())
  File "/usr/local/lib/python3.8/asyncio/runners.py", line 43, in run
    return loop.run_until_complete(main)
  File "/usr/local/lib/python3.8/asyncio/base_events.py", line 612, in run_until_complete
    return future.result()
  File ".../edgedb/edb/testbase/server.py", line 835, in _run
    g.create_task(controller(
  File ".../edgedb/edb/common/taskgroup.py", line 169, in __aexit__
    raise me from None
edb.common.taskgroup.TaskGroupError: unhandled errors in a TaskGroup; 3 sub errors: (InternalServerError)
 + InternalServerError: 
 |   File ".../edgedb/edb/testbase/server.py", line 832, in controller
 |     await coro(*args)
 |   File ".../edgedb/edb/testbase/server.py", line 861, in _setup_database
 |     await dbconn.execute(setup_script)
 |   File ".../edgedb-python/edgedb/asyncio_con.py", line 165, in execute
 |     await self._protocol.simple_query(query)
 |   File "edgedb/protocol/protocol.pyx", line 497, in simple_query


 + InternalServerError: 
 |   File ".../edgedb/edb/testbase/server.py", line 832, in controller
 |     await coro(*args)
 |   File ".../edgedb/edb/testbase/server.py", line 861, in _setup_database
 |     await dbconn.execute(setup_script)
 |   File ".../edgedb-python/edgedb/asyncio_con.py", line 165, in execute
 |     await self._protocol.simple_query(query)
 |   File "edgedb/protocol/protocol.pyx", line 497, in simple_query


 + InternalServerError: 
 |   File ".../edgedb/edb/testbase/server.py", line 832, in controller
 |     await coro(*args)
 |   File ".../edgedb/edb/testbase/server.py", line 861, in _setup_database
 |     await dbconn.execute(setup_script)
 |   File ".../edgedb-python/edgedb/asyncio_con.py", line 165, in execute
 |     await self._protocol.simple_query(query)
 |   File "edgedb/protocol/protocol.pyx", line 497, in simple_query

```

By the way, is there a way to get more information from the test server, like a full traceback?

edb/schema/pointers.py
RobertoPrevato: Among the first comments, this is the only one for which I didn't make a change in the current branch.

elprans: It always returns `None`, so try annotating it as such.

Issue #1158, opened 2020-01-22
Prohibit or handle long database names


elprans:
I finally realized why `test_dump01_restore_compatibility` is failing on macOS.  The path to the dump file is longer than the PostgreSQL's limit of 63 characters, and since the test uses the full path name for the database name, it fails, because Postgres silently truncates long names (it issues a NOTICE response, but not an error).

We have graceful handling of other schema classes in those cases, but not databases.

1st1:
Should this be closed now?

elprans:
No, the linked PR is related, but does not fix the issue at hand.

PR #1331, opened 2020-04-09
Prohibit database names longer than 63 characters.

vpetrovykh:
Fixes: #1158

### DIFF
--- a/edb/schema/database.py
+++ b/edb/schema/database.py
@@ -19,13 +19,17 @@
 
 from __future__ import annotations
 
+from edb import errors
+
 from edb.edgeql import ast as qlast
 from edb.edgeql import qltypes
+from edb.schema import defines as s_def
 
 from . import abc as s_abc
 from . import annos as s_anno
 from . import delta as sd
 from . import objects as so
+from . import schema as s_schema
 
 
 class Database(so.GlobalObject, s_anno.AnnotationSubject, s_abc.Database,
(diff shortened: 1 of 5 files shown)


edb/schema/database.py
```diff
@@ -39,7 +42,24 @@ class DatabaseCommandContext(sd.ObjectCommandContext):
...
+    def _create_begin(
+        self,
+        schema: s_schema.Schema,
+        context: sd.CommandContext,
+    ) -> s_schema.Schema:
+        schema = super()._create_begin(schema, context)
+
+        # Validate that the database name is fewer than 64 characters
+        name = self.get_attribute_value('name')
+        if len(str(name)) > 63:
```
elprans, line 55: Please use a constant instead of a magic number.  We have `MAX_NAME_LENGTH` in `pointers.py`, factor it out somewhere.

1st1: `server/defines.py` would be an OK place since we have so many other constants there.

elprans: Yes, although I'd move it out or we'll be importing `server` everywhere.

vpetrovykh: I've factored out the 63 constant into `server/defines.py` from `database.py`, `pointers.py` and `common.py`.

tests/test_database.py
```diff
@@ -45,3 +45,9 @@ class TestDatabase(tb.ConnectedTestCase):
             await conn.aclose()
         finally:
             await self.con.execute('DROP DATABASE mytestdb;')
+
+    async def test_database_create_02(self):
+        with self.assertRaisesRegex(edgedb.SchemaDefinitionError,
+                                    r'Database names longer than 63 '
```
elprans, line 51: You missed a hardcoded spot

edb/schema/pointers.py
```diff
@@ -27,6 +27,7 @@
 from edb.edgeql import ast as qlast
 from edb.edgeql import compiler as qlcompiler
 from edb.edgeql import qltypes
+from edb.server import defines
```
elprans, line 30: Let's move the defines out of `server`.  I don't like the cross-pollination

tests/test_database.py
vpetrovykh: OK, I've updated the test.

edb/schema/pointers.py
vpetrovykh: OK, moved to a separate file in `schema`

PR #1332, opened 2020-04-10
Fix support of arguments unused by postgres

tailhook:
This PR has to be prerequisite for constant extraction.

### DIFF
--- a/edb/edgeql/compiler/stmtctx.py
+++ b/edb/edgeql/compiler/stmtctx.py
@@ -31,6 +31,7 @@ from edb import errors
 from edb.common import parsing
 
 from edb.ir import ast as irast
+from edb.ir import utils as irutils
 from edb.ir import typeutils
 
 from edb.schema import functions as s_func
@@ -204,9 +205,12 @@ def fini_expression(
             context=srcctx,
         )
 
+    used_params = {p.name for p in irutils.get_parameters(ir)}
+
     result = irast.Statement(
         expr=ir,
         params=ctx.env.query_parameters,
+        unused_params=frozenset(set(ctx.env.query_parameters) - used_params),
         views=ctx.view_nodes,
         source_map=ctx.source_map,
         scope_tree=ctx.path_scope,
--- a/edb/pgsql/compiler/__init__.py
+++ b/edb/pgsql/compiler/__init__.py
@@ -42,13 +42,15 @@ from .context import OutputFormat  # NOQA
 
 
 def compile_ir_to_sql_tree(
-        ir_expr: irast.Base, *,
-        output_format: Optional[OutputFormat]=None,
-        ignore_shapes: bool=False,
-        explicit_top_cast: Optional[irast.TypeRef]=None,
-        singleton_mode: bool=False,
-        use_named_params: bool=False,
-        expected_cardinality_one: bool=False) -> pgast.Base:
+    ir_expr: irast.Base, *,
+    output_format: Optional[OutputFormat]=None,
+    ignore_shapes: bool=False,
+    explicit_top_cast: Optional[irast.TypeRef]=None,
+    singleton_mode: bool=False,
+    use_named_params: bool=False,
+    expected_cardinality_one: bool=False,
+    argmap: Dict[str, int]=None,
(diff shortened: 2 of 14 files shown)


edb/edgeql/compiler/stmtctx.py
```diff
@@ -204,9 +205,12 @@ def fini_expression(
             context=srcctx,
         )
 
+    used_params = {p.name for p in irutils.get_parameters(ir)}
```
elprans, line 208: This is inefficient because it searches the entire AST tree with a callback on every node.  Let's use an accumulator in the context instead, adding every time an `ir.Parameter` is constructed.

edb/pgsql/compiler/__init__.py
```diff
@@ -89,28 +92,31 @@ def compile_ir_to_sql_tree(
...
-        expected_cardinality_one: bool=False,
-        pretty: bool=True) -> Tuple[str, Dict[str, int]]:
+    ir_expr: irast.Base, *,
+    output_format: Optional[OutputFormat]=None,
+    ignore_shapes: bool=False,
+    explicit_top_cast: Optional[irast.TypeRef]=None,
+    use_named_params: bool=False,
+    expected_cardinality_one: bool=False,
+    pretty: bool=True,
+    argmap: Dict[str, int]=None,
```
elprans, line 102: I'm afraid this makes the API harder to use.  The `argmap` doesn't *appear* as a required argument, whereas now it has to be carefully constructed by the caller or the compilation will crash.  It seems like building `argmap` is a concern separate from filtering out unused params, can we keep it inside `compile_ir_to_sql`?

elprans:
I wonder if "parking" unused parameters in a SQL CTE would be a cleaner approach: it wouldn't require the param bookkeeping and bind repacking.

tests/test_edgeql_expressions.py
```diff
@@ -834,6 +834,27 @@ class TestExpressions(tb.QueryTestCase):
...
+    async def test_edgeql_expr_variables_01(self):
+        await self.assert_query_result(
+            r'''SELECT (INTROSPECT TYPEOF <int64>$0).name;''',
+            {'std::int64'},
+            variables=(7,),
+        )
+
+    async def test_edgeql_expr_variables_02(self):
+        await self.assert_query_result(
+            r'''SELECT <str>$1 ++ (INTROSPECT TYPEOF <int64>$0).name;''',
```
1st1, line 846: + @elprans 

I'm not sure it's supposed to work like this. I think one should be able to have a query in `INTROSPECT TYPEOF` with a `$0` parameter and then use `$0` in their outer query, as in:

```
SELECT <str>$0 ++ (INTROSPECT TYPEOF <int64>$0).name;
```

That allows one to insert any query into `INTROSPECT TYPEOF` kind of construct without worrying about potential name/positional parameters collision.

edb/edgeql/compiler/stmtctx.py
tailhook: This is how `query_parameters` work. But transformation that makes it unused happens later in the transformation as far as I understand (i.e. unused ones are present in `query_parameters`).

tests/test_edgeql_expressions.py
1st1: Disregard this comment; Elvis convinced me that the current approach is fine.

+1 to inlining unused SQL parameters in a CTE.

edb/edgeql/compiler/stmtctx.py
elprans: In `typegen.py:82` make a copy of `query_parameters` in `subctx` and discard it afterward. That way you will ensure that only used parameters end up in `query_parameters`

tailhook: Could there be any other operators that discard values of arguments?

tailhook: Could there be an optimization pass that discard things like `0*$x`?

elprans: Not currently, no.

tailhook:
Closed in favor of #1337

PR #1336, opened 2020-04-13
Add annotations for schema lproperties

RobertoPrevato:


edb/schema/lproperties.py
```diff
@@ -311,21 +375,37 @@ class AlterProperty(
...
 
-    def _apply_field_ast(self, schema, context, node, op):
+    def _apply_field_ast(
+        self,
+        schema: s_schema.Schema,
+        context: sd.CommandContext,
+        node: qlast.DDLOperation,
+        op: sd.AlterObjectProperty,
+    ) -> None:
+        # TODO: this method is not covered by tests
```
RobertoPrevato, line 400: @elprans, this function is not covered by tests. Can you please give me an hint, on what kind of test should be arranged to cover this scenario? What is `op.new_value` in this context?

edb/schema/lproperties.py
```diff
@@ -46,14 +46,19 @@
...
+    def derive_ref(  # type: ignore
+        self,
         schema: s_schema.Schema,
         source: s_sources.Source,
         target: Optional[s_types.Type] = None,
-        attrs: Optional[Mapping[str, Any]] = None,
+        attrs: Optional[Dict[str, Any]] = None,
         **kwargs: Any,
-    ) -> Tuple[s_schema.Schema, referencing.ReferencedT]:
+    ) -> Tuple[s_schema.Schema, pointers.Pointer]:
```
elprans, line 60: You need to return `Property` here, not `Pointer`. That'll make mypy happy.

edb/schema/lproperties.py
elprans: Interesting.  This a deparsing of an `ALTER PROPERTY SET TYPE <foo>`.  To cover this, add a test to `TestGetMigration` that changes the type of a property. `new_value` in this case is `types.Type`.

RobertoPrevato: I did several attempts to configure a test, following the same patterns already defined in `TestGetMigration` methods, but it looks like I am not arranging the right test. I confirm that the method is not hit by current suite.

Can you show me how to apply a second migration to a schema already modified by a first migration in code?

elprans: Yes, we seem to have a fairly large coverage gap. I'm working on fixing the hardness in `TestGetMigration` to support testing multi-step migration deparse.

elprans: The enabled tests are a much deeper rabbit hole than I anticipated.  Let's not block this PR on that.

RobertoPrevato: Thanks! I was going to propose the same.

PR #1337, opened 2020-04-13
Fix handling of parameters unused by postgres

tailhook:
Parameters are added to the CTE to become know to postgres.

Supersedes #1332

### DIFF
--- a/edb/ir/ast.py
+++ b/edb/ir/ast.py
@@ -63,6 +63,7 @@ Set (
 
 from __future__ import annotations
 
+import dataclasses
 import typing
 import uuid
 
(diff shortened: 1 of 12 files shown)


edb/ir/ast.py
```diff
@@ -405,11 +406,25 @@ class Command(Base):
     __abstract_node__ = True
 
 
+@dataclasses.dataclass(frozen=True)
```
elprans, line 409: Why a dataclass instead of an AST node?

tailhook: 1. I don't think this should be visited by `dispatch.compile` (in particular because TypeRef here is a copy)
2. It contains mix of AST types: `ir.ast.TypeRef` and `schema.types.Type`
3. I would make it `Tuple` as in "source_map" but I prefer data classes instead of tuples.

Other than that, I didn't give it much thought.

PR #1341, opened 2020-04-14
Move the curried constraint parameter value to the params link in introspection

elprans:
Currently, we expose the applied constraint args in the `args` link of
the `schema::Constraint` type.  But it already has the `params` link it
inherits from `CallableObject`, so it makes sense to use that link
instead.

### DIFF
--- a/edb/lib/schema.edgeql
+++ b/edb/lib/schema.edgeql
@@ -167,8 +167,7 @@ CREATE ABSTRACT TYPE schema::VolatilitySubject {
 CREATE TYPE schema::Constraint
     EXTENDING schema::CallableObject, schema::InheritingObject
 {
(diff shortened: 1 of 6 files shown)


edb/lib/schema.edgeql
```diff
@@ -167,8 +167,7 @@ CREATE ABSTRACT TYPE schema::VolatilitySubject {
 CREATE TYPE schema::Constraint
     EXTENDING schema::CallableObject, schema::InheritingObject
 {
-    CREATE MULTI LINK args -> schema::Parameter {
-        CREATE CONSTRAINT std::exclusive;
+    ALTER LINK params {
         CREATE PROPERTY value -> std::str;
```
1st1, line 171: Maybe call it `argvalue`?

PR #1343, opened 2020-04-14
Separate used-defined and system object type hierarchies

elprans:
Currently, we use `std::Object` as the root object type for everything,
including the object types defined by the system.  This patch separates
the user-defined and system object hierarchies by adding the new
`std::BaseObject` type which is now the true root type.  `std::Object`
extends from `std::BaseObject` and becomes the default base for
user-defined object types.  All types declared in `schema::', `cfg::`
and `sys::` are made to extend from `BaseObject` directly.


tailhook:
This should have a bump of `EDGEDB_CATALOG_VERSION` too

PR #1344, opened 2020-04-15
EdgeQL constant extraction

tailhook:
This only extracts integers. Other types should be easier as most special cases are on integers.

### DIFF
--- a/edb/server/mng_port/edgecon.pyx
+++ b/edb/server/mng_port/edgecon.pyx
@@ -762,7 +762,8 @@ cdef class EdgeConnection:
 
         if self.debug:
             self.debug_print('Cache key', entry.key())
-            self.debug_print('Extra variables', entry.variables())
+            self.debug_print('Extra variables', entry.variables(),
+                             'after', entry.first_extra())
 
         query_unit = self.dbview.lookup_compiled_query(
             entry.key(), io_format, expect_one, implicit_limit)
--- a/edb/server/mng_port/edgecon.pyx
+++ b/edb/server/mng_port/edgecon.pyx
@@ -27,7 +27,7 @@ import traceback
 cimport cython
 cimport cpython
 
-from typing import List
+from typing import List, Optional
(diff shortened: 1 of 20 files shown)


edb/server/mng_port/edgecon.pyx
```diff
@@ -736,12 +754,19 @@ cdef class EdgeConnection:
         object io_format,
         bint expect_one,
         uint64_t implicit_limit,
-    ):
+    ) -> CompiledQuery:
         if self.debug:
             self.debug_print('PARSE', eql)
 
+        entry = rewrite(eql)
```
elprans, line 761: I think "preparse" (tokenize+extract args) would be a better term for this stage.  Also, `Entry` is nondescript. `PreparsedQuery` seems better.

edb/server/mng_port/edgecon.pyx
```diff
@@ -785,9 +811,14 @@ cdef class EdgeConnection:
 
         if not cached and query_unit.cacheable:
             self.dbview.cache_compiled_query(
-                eql, io_format, expect_one, implicit_limit, query_unit)
+                entry.key(), io_format, expect_one, implicit_limit, query_unit)
 
-        return query_unit
+        return CompiledQuery(
```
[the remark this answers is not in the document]
tailhook: `QueryUnit` comes from compiler. Things we add here come from server itself. It's ugly to modify structure with stubs got from compiler.

edb/server/mng_port/edgecon.pyx
tailhook: Well we have [preparser module](https://github.com/edgedb/edgedb/blob/master/edb/edgeql-parser/src/preparser.rs) which is basically statement delimiter.

elprans: How about "normalize" then?

tailhook: Renamed.

PR #1345, opened 2020-04-15
Schema annotations for migrations and objtypes

RobertoPrevato:


edb/schema/links.py
```diff
@@ -119,6 +119,17 @@ def has_user_defined_properties(self, schema: s_schema.Schema) -> bool:
...
                      if not p.is_special_pointer(schema)])
 
+    def get_source_type(
+        self,
+        schema: s_schema.Schema
+    ) -> s_types.Type:
+        from . import types as s_types
+
+        source = self.get_source(schema)
+        if not isinstance(source, s_types.Type):
```
elprans, line 129: Let's use an `assert` statement here like we do in other places.

edb/schema/objtypes.py
```diff
@@ -236,12 +257,12 @@ def get_or_create_union_type(
         module=module,
     )
 
-    objtype = schema.get_by_id(type_id, None)
+    objtype: Optional[so.Object] = schema.get_by_id(type_id, None)
```
elprans, line 260: `get_by_id` also supports the `type` argument, please use that.

edb/schema/objtypes.py
```diff
@@ -257,7 +257,7 @@ def get_or_create_union_type(
         module=module,
     )
 
-    objtype: Optional[so.Object] = schema.get_by_id(type_id, None)
+    objtype = schema.get_by_id(type_id, None, type=so.Object)
```
elprans, line 260: Why not `ObjectType`?

RobertoPrevato: Sorry, a mistake.

PR #1346, opened 2020-04-16
Make sure to_intXX and to_floatXX functions don't crash due to inlining

elprans:
In certain situations Postgres might fail to optimize away the `"fmt" is
NULL` branch of `CASE` when inlining the function, triggering the
constant folder to attempt to evaluate an invalid cast.  Fix this by
wrapping the casts into functions formulated in a way that prevents
inlining.

### DIFF
--- a/edb/pgsql/metaschema.py
+++ b/edb/pgsql/metaschema.py
@@ -185,6 +185,114 @@ class StrToDecimal(dbops.Function):
         )
 
 
+class StrToInt64NoInline(dbops.Function):
+    """String-to-int64 cast with noinline guard.
+
+    Adding a LIMIT clause to the function statement makes it
+    uninlinable due to the Postgres inlining heuristic looking
+    for simple SELECT expressions only (i.e. no clauses.)
+
+    This might need to change in the future if the heuristic
+    changes.
+    """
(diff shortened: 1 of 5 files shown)


edb/pgsql/metaschema.py
```diff
@@ -185,6 +185,106 @@ def __init__(self) -> None:
         )
 
 
+class StrToInt64NoInline(dbops.Function):
+    """String-to-int64 cast with noinline guard."""
+    text = r'''
+        SELECT
+            "val"::bigint
+        LIMIT
```
1st1, line 193: Do you know why LIMIT 1 is non-ininable? Maybe you can document that?

tailhook: It's probably not "limit 1". It probably nested function call that is not inlinable. But I'm not sure.

elprans: No, it's `LIMIT` specifically: https://github.com/postgres/postgres/blob/24d2d38b1eb86c0b410ad0f07f66566a83c6f05c/src/backend/optimizer/util/clauses.c#L4501-L4529

I added a comment to clarify.

PR #1348, opened 2020-04-16
edgeql-parser: Expose keywords lists from the crate

tailhook:
This is needed for proper highlighting in repl (see edgedb/edgedb-cli#27)


elprans:
We seem to have two lists of keywords now: this and the one in `edgeql.parser.grammar.keywords`.  The latter is used as the source for various generators (highlighting etc).  Can we do something about eliminating the duplication?  Generate one from another?

1st1:
> Generate one from another?

Ideally generate Rust from Python's `keywords.py`.

tailhook:
I don't like the idea of relying on codegen on every little thing. We can never add any new keyword without lots of precautions, deprecation process and so on. So I don't think adding a single line few times a year at most justifies another chunk of codegen. We are already not in a stage where we frequently add new keywords.

elprans:
> We can never add any new keyword without lots of precautions, deprecation process and so on

This only applies to reserved keywords which are the minority.

> I don't like the idea of relying on codegen on every little thing.

OK, can we make the rust keywords exported and accessible in Python then?

tailhook:
> OK, can we make the rust keywords exported and accessible in Python then?

Done.

PR #1351, opened 2020-04-19
Add annotations for schema sources and std

RobertoPrevato:


edb/schema/utils.py
```diff
@@ -909,8 +910,11 @@ def get_config_type_shape(
             elem_path.append(qlast.Ptr(ptr=qlast.ObjectRef(name=pn)))
 
             ptype = p.get_target(schema)
+            assert ptype is not None
 
             if ptype.is_object_type():
+                assert isinstance(ptype, s_objtypes.ObjectType)
```
elprans, line 916: Just replace the if condition with the isinstance check, since you're doing it anyway.

Issue #1340, opened 2020-04-13
Make query arguments required by default; add syntax to make them optional


1st1:
Currently query arguments are optional but it makes more sense to make them required by default. E.g. in

```
SELECT <OPTIONAL int64>$0 + <int64>$1;
```

the `$0` parameter would be optional and `$1` would be required.

See also https://github.com/edgedb/edgedb/pull/1338#discussion_r407655411

In the context of this change we should also consider allowing use of `SET OF` in the argument cast, e.g.

```
SELECT sim(<OPTIONAL SET OF int64>$0);
# or
SELECT sim(<SET OF int64>$0);
```

The main problem here, IMO, is that with this proposal `SET OF` and `OPTIONAL` appear to have slightly different semantics as opposed to when used in function signatures.

1st1:
@tailhook Do we need to update the docs? Could you also clarify what specifically was implemented in the merged PR? (we'll need that in changelog, cc @vpetrovykh)

tailhook:
Documentation is in #1373. Shortly: by default arguments made required instead of optional and added `<OPTIONAL type>$x`; `<REQUIRED type>$x` is also supported (but no-op).

elprans:
Done.

PR #1352, opened 2020-04-20
Required arguments

tailhook:
This is the first part of fixing #1340.

Unfortunately, I haven't found any changes this made in generated SQL. So I might be missing something.

### DIFF
--- a/edb/server/mng_port/edgecon.pyx
+++ b/edb/server/mng_port/edgecon.pyx
@@ -1051,7 +1051,6 @@ cdef class EdgeConnection:
             self.write(self.make_command_complete_msg(query_unit))
             return
 
-        bound_args_buf = self.recode_bind_args(bind_args, compiled)
 
         process_sync = False
         if self.buffer.take_message_type(b'S'):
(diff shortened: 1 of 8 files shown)


tailhook:
> Unfortunately, I haven't found any changes this made in generated SQL. So I might be missing something.

On the other hand, it fixed the issue with EdgeQL normalization, so should be fine.

elprans:
>  I haven't found any changes this made in generated SQL

We are likely not actually handling optional arguments correctly in all cases.

edb/server/mng_port/edgecon.pyx
```diff
@@ -1619,6 +1620,9 @@ cdef class EdgeConnection:
                 frb_read(&in_buf, 4)  # reserved
             in_len = hton.unpack_int32(frb_read(&in_buf, 4))
             out_buf.write_int32(in_len)
+            if in_len < 0:
+                name = query.query_unit.in_type_args[i]
```
1st1, line 1624: Please add a comment that `in_len == -1` means that the argument value is NULL.

PR #1353, opened 2020-04-20
Completes annotations for schema package

RobertoPrevato:


edb/schema/expraliases.py
```diff
@@ -33,17 +33,20 @@
...
+    sd.ObjectCommandContext[Any],
+    s_anno.AnnotationSubjectCommandContext
+):
     pass
 
 
 class AliasCommand(
-    sd.QualifiedObjectCommand[s_types.Type],
-    s_types.TypeCommand,
+    sd.QualifiedObjectCommand[so.QualifiedObject],
```
RobertoPrevato, line 48: This was annotated as `sd.QualifiedObjectCommand[s_types.Type]`, but mypy complains: `Type argument "edb.schema.types.Type" of "QualifiedObjectCommand" must be a subtype of "edb.schema.objects.QualifiedObject"`.
I understand so.QualifiedObject is the right type argument here.

elprans: Use `InheritingType`

edb/schema/expraliases.py
```diff
@@ -79,12 +82,15 @@ def command_for_ast_node(
         modaliases = cls._modaliases_from_ast(schema, astnode, context)
         ctx = AliasCommandContext(
             schema,
-            op=sd._dummy_object,
+            op=sd._dummy_object,  # type: ignore
```
elprans, line 85: `_dummy_command`

edb/schema/expraliases.py
```diff
@@ -79,12 +82,15 @@ def command_for_ast_node(
...
             scls=sd._dummy_object,
             modaliases=modaliases,
         )
 
+        mapping: Dict[Type[qlast.ObjectDDL], Any]
```
elprans, line 92: Why `Any` and not `Type[sd.Command]`?

edb/schema/ordering.py
```diff
@@ -68,7 +68,8 @@ def linearize_delta(
     ordered = list(topological.sort(depgraph, allow_unresolved=True,
                                     return_record=True))
 
-    parents: Dict[Tuple[Type[sd.ObjectCommand], str], sd.ObjectCommand]
+    parents: Dict[Tuple[Type[sd.ObjectCommand[Any]], str],
```
elprans, line 71: s/Any/Object/

mypy.ini
```diff
@@ -196,83 +196,7 @@ no_implicit_reexport = True
...
-disallow_untyped_defs = True
-disallow_incomplete_defs = True
-check_untyped_defs = True
-disallow_untyped_decorators = True
-no_implicit_optional = True
-warn_unused_ignores = True
-warn_return_any = True
-no_implicit_reexport = True
-
-[mypy-edb.schema.types.*]
```
elprans, line 275: Missed `schema._types`

tests/test_type_coverage.py
```diff
@@ -245,7 +247,7 @@ def test_cqa_type_coverage_repl(self) -> None:
         self.assertFunctionCoverage(EDB_DIR / "repl", 100.0)
 
     def test_cqa_type_coverage_schema(self) -> None:
-        self.assertFunctionCoverage(EDB_DIR / "schema", 98.69)
+        self.assertFunctionCoverage(EDB_DIR / "schema", 100.0)
```
elprans, line 250: 🎆

tests/test_type_coverage.py
```diff
@@ -54,7 +54,9 @@ def visit_FunctionDef(
             if arg.annotation:
                 self.assert_no_strings(arg.annotation)
                 typed_count += 1
-            elif index == 0 and self._classdefs and arg.arg in {"self", "cls"}:
+            elif index == 0 and self._classdefs and arg.arg in {"self",
+                                                                "cls",
+                                                                "mcls"}:
```
RobertoPrevato, line 59: Here I modified to include `mcls`, because some `@classmethod` are defined with first argument with such name. Since Flake8 doesn't complain about this, I thought logical to update this piece of code.

edb/schema/expraliases.py
RobertoPrevato: I tried `_dummy_command` earlier, then I hesitated since mypy wants an `ObjectCommand` and it's necessary to add a new `_dummy_object_command = ObjectCommand(classname="dummy")`. :+1: pushing it now

edb/schema/expraliases.py
RobertoPrevato: mypy complains in all assignments below: the dictionary key is contravariant.
image

I tried to use `TypeVar` to define a generic type, even using the `covariant=True` argument, but mypy is not happy anyway. In C#, this would be a case for a generic, because the key type argument for a Dictionary would be contravariant, too (e.g. no `Dictionary<IEnumerable, IEnumerable>` but rather `Dictionary<K, V> where V : IEnumerable`). I soon add a comment with the error when using a TypeVar.

RobertoPrevato: I deleted my previous comment (...). I got confused here, and missed the fact that I should have annotated each of the class attributes with `: Dict[Type[qlast.NamedDDL], Type[sd.Command]]`. :facepalm: 
Pushing a fix.

PR #1355, opened 2020-04-21
Implement OPTIONAL query parameters

tailhook:


edb/server/compiler/compiler.py
```diff
@@ -415,17 +421,17 @@ def _compile_ql_query(
                     named=named)
                 if array_params:
                     in_array_backend_tids = {p[0]: p[1] for p in array_params}
+                in_type_args = {name: name in optional_params
```
elprans, line 424: Could we also use a dataclass here so that we don't have to change it again once we implement the `SET OF` modifier?

edb/ir/ast.py
```diff
@@ -413,6 +413,9 @@ class Param:
     name: str
     """Parameter name"""
 
+    optional: bool
```
elprans, line 416: Let's use `required: bool` instead for consistency with pointers and the cardinality inference machinery.

edb/server/compiler/compiler.py
tailhook: Fixed

tailhook: Also moved array tids to the structure

edb/ir/ast.py
tailhook: Fixed. Also fixed `irast.Parameter(optional)` for consistency (but we can revert that commit if you wish).

elprans:
Need a documentation update for this also.

tailhook:
> Need a documentation update for this also.

Yes. Let me do it as a separate PR.

edb/edgeql/compiler/expr.py
```diff
@@ -401,6 +401,13 @@ def compile_TypeCast(
     target_typeref = typegen.ql_typeexpr_to_ir_typeref(expr.type, ctx=ctx)
     ir_expr: irast.Base
 
+    is_parameter = isinstance(expr.expr, qlast.Parameter)
+    if not is_parameter and expr.modifier:
+        raise errors.QueryError(
+            'cardinality modifiers REQUIRED and OPTIONAL only allowed '
```
1st1, line 407: `only allowed` -> `are only allowed`

edb/pgsql/ast.py
```diff
@@ -367,6 +368,15 @@ class NullRelation(ReturningQuery):
     where_clause: BaseExpr
 
 
+@dataclasses.dataclass(frozen=True)
+class Param:
```
1st1, line 372: Why isn't this a regular AST node?

edb/server/mng_port/edgecon.pyx
```diff
@@ -1606,38 +1606,45 @@ cdef class EdgeConnection:
...
 
         # number of elements in the tuple
-        argsnum = hton.unpack_int32(frb_read(&in_buf, 4))
+        recv_args = hton.unpack_int32(frb_read(&in_buf, 4))
+        decl_args = len(query.query_unit.in_type_args or ())
+
+        if recv_args != decl_args:
+            raise errors.QueryError(
+                f"invalid argument count required: "
+                f"{decl_args}, got: {recv_args}")
```
1st1, line 1615: `{decl_args}` -> `expected {decl_args}, got`

edb/pgsql/ast.py
elprans: Because it's metadata and not part of SQL syntax, i.e. we don't codegen or otherwise handle it as a standalone node.

1st1: What's the downside of still using an AST node for it?

elprans: Mainly for dispatch purposes, we have `singledispatch` coverage tests and things like that.  I think it's cleaner to have non-AST structures to represent things that aren't encoding a piece of syntax.

1st1: 👍

PR #1368, opened 2020-04-25
raise QueryError when compiling query with duplicate elements definition

### LINKED ISSUE #1321: Prohibit pointer name shadowing in shapes
Currently, EdgeDB would silently accept queries like:

```
SELECT Foo { bar, bar }
```

Confusingly, even though the output contains two properties, internally the latter definition "wins", shadowing the first one:

```
db> SELECT Foo { bar, bar := 'spam' }
{Object { name: 'spam', name: 'spam' }}
```

This behavior is even more confusing in the polymorphic shapes.

We *could* fix this by adding internal disambiguation, but then the ambiguity will simply shift to the output, where one would have two identically named fields.  IMO, it would be best to simply disallow such queries.

nsidnev:
closes #1321

### DIFF
--- a/edb/edgeql/compiler/viewgen.py
+++ b/edb/edgeql/compiler/viewgen.py
@@ -149,11 +149,21 @@ def _process_view(
 
     for shape_el in elements:
         with ctx.newscope(fenced=True) as scopectx:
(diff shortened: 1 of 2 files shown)


1st1:
Thank you, Nik!

PR #1373, opened 2020-04-29
Document option vars

tailhook:
This is a followup on #1355

### DIFF
--- a/docs/edgeql/expressions/parameters.rst
+++ b/docs/edgeql/expressions/parameters.rst
@@ -1,0 +1,170 @@
+.. _ref_eql_expr_params:
+
+Parameters
+==========
+
+:edb-alt-title: Query Parameters
+
+A parameter reference is used to indicate a value that is supplied externally
+to an EdgeQL expression.
+
+An example of query, that can be parametrized:
+
+.. code-block:: edgeql-repl
+
+    db> SELECT Person { birth_date } FILTER .first_name = 'John';
+    {
+        Object {birth_date: <cal::local_date>'1971-01-01'},
+    }
+
+The equivalent parametrized query:
+
+.. code-block:: edgeql-repl
+
+    db> SELECT Person { birth_date } FILTER .first_name = <str>$name;
+    Parameter <str>$name: John
+    {
+        Object {birth_date: <cal::local_date>'1971-01-01'},
+    }
+
+Note how the actual parameter value is asked by REPL in a separate prompt,
+after the whole query is accepted.
+
+Here's how parameters can be used with language bindings. e.g. with Python:
+
+.. code-block:: python
+
+    await connection.fetchone(
+        "SELECT Person { birth_date } FILTER .first_name = <str>$name",
+        name="John")
+
+And with JavaScript:
+
+.. code-block:: javascript
+
+    await connection.fetchOne(
+        "SELECT Person { birth_date } FILTER .first_name = <str>$name",
+        {name: "John"})
+
+
+Parameter Types
+---------------
+
+The following values can be turned into parameters:
+
+* Any :ref:`scalar type <ref_datamodel_scalar_types>`
+* :eql:type:`Array <array>` of any scalar type
+
+The parameter type must be specified in angular brackets in front of the
+variable name (the syntax is similar to type casts):
+
+.. code-block:: edgeql
+
+    INSERT Person {
+        first_name := <str>$first_name,
+        age := <int32>$age,
+        interests := <array<str>>$interests,
+    }
+
+See :ref:`ref_eql_types` for more info on specifying types.
+
+
+Optional Parameters
+-------------------
+
+By default, query parameters are ``REQUIRED`` which means that the query would
+fail if parameter value is an empty set. You can use ``OPTIONAL`` keyword for
+optional parameters.
+
+Example in Python:
+
+.. code-block:: python
+
+    for name, age in [("John", 33), ("Jack", None)]:
+        await conn.fetchone("""
+            INSERT Person {
+                first_name := <str>$name,
+                age := <OPTIONAL str>$age,
+            }
+        """, name=name, age=age)
+
+Example in JavaScript:
+
+.. code-block:: javascript
+
+    for(let [name, age] of [["John", 33], ["Jack", null]]) {
+        await conn.fetchOne("""
+            INSERT Person {
+                first_name := <str>$name,
+                age := <OPTIONAL str>$age,
+            }
+        """, {name, age})
+    }
+
+Note: denoting a *scalar EdgeQL empty set* is language-specific. We use
+``None`` in Python and ``null`` in JavaScript.
+
+The ``REQUIRED type_name`` is also a valid, although, redundant syntax:
+
+.. code-block:: edgeql
+
+    INSERT Person {
+        first_name := <REQUIRED str>$name,
+        age := <OPTIONAL str>$age,
+    }
+
+
+What Can be Parametrized?
+-------------------------
+
+Any data manipulation language (DML) statement can be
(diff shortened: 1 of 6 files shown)


docs/edgeql/expressions/parameters.rst
```diff
@@ -0,0 +1,173 @@
...
+    Parameter <str>$name: John
+    {
+        Object {birth_date: <cal::local_date>'1971-01-01'},
+    }
+
+Note how the actual parameter value is asked by REPL in a separate prompt,
+after the whole query is accepted.
+
+While parameters are useful in REPL, their power is mostly when using language
+bindings. Example of using parameters in Python:
```
1st1, line 34: Hm, "their power is mostly" is incomplete. I'd replace this paragraph with something like "Here's how parameters can be used with language bindings, e.g. with Python:"

docs/edgeql/expressions/parameters.rst
```diff
@@ -0,0 +1,173 @@
...
+
+Parameter Types
+---------------
+
+The following values can be turned into parameters:
+
+* Any :ref:`scalar type <ref_datamodel_scalar_types>`
+* :eql:type:`Array <array>` of any scalar type
+
+Parameter type is specified in the angular brackets in front of variable
```
1st1, line 59: "the variable name"

docs/edgeql/expressions/parameters.rst
```diff
@@ -0,0 +1,173 @@
...
+name, similarly to normal type cast:
+
```
1st1, line 61:
```suggestion
The parameter type must be specified in angular brackets in front of the variable
name (the syntax is similar to type casts):

```

docs/edgeql/expressions/parameters.rst
```diff
@@ -0,0 +1,173 @@
...
+        interests := <array<str>>$interests,
+    }
+
+See :ref:`ref_eql_types` for more info on specifying types.
+
+
+Optional Parameters
+-------------------
+
+By default parameter types are ``REQUIRED`` which means that query will fail
```
1st1, line 76:
```suggestion
By default, query parameters are ``REQUIRED`` which means that the query would fail
```

docs/edgeql/expressions/parameters.rst
```diff
@@ -0,0 +1,173 @@
...
+if parameter value is nil. You can use ``OPTIONAL`` keyword for optional
```
1st1, line 77:
```suggestion
if parameter value isn't provided. You can use the ``OPTIONAL`` keyword to denote optional
```

docs/edgeql/expressions/parameters.rst
```diff
@@ -0,0 +1,173 @@
...
+arguments.
```
1st1, line 78:
```suggestion
parameters.
```

docs/edgeql/expressions/parameters.rst
```diff
@@ -0,0 +1,173 @@
...
+                first_name := <str>$name,
+                age := <OPTIONAL str>$age,
+            }
+        """, name=name, age=age)
+
+Example in JavaScript:
+
+.. code-block:: javascript
+
+    for(let [name, age] of [("John", 33), ("Jack", null)]) {
```
1st1, line 96:
```suggestion
    for(let [name, age] of [["John", 33], ["Jack", null]]) {
```

docs/edgeql/expressions/parameters.rst
```diff
@@ -0,0 +1,173 @@
...
+        await conn.fetchOne("""
+            INSERT Person {
+                first_name := <str>$name,
+                age := <OPTIONAL str>$age,
+            }
+        """, {name, age})
+    }
+
+Note: there is a language-specific nil value, ``None`` in Python example and
```
1st1, line 105: We don't use the "nil" term anywhere in the docs and IMO we shouldn't introduce it here.

```suggestion
In most languages there is a way to indicate "absence of value", e.g. in Python we use ``None``, and we use
```

docs/edgeql/expressions/parameters.rst
```diff
@@ -0,0 +1,173 @@
...
+parametrized. Which means you can parametrize ``SELECT``, ``INSERT``
+and ``UPDATE`` statements of arbitrary nesting and complexity. Any constant
+value can be turned into a parameter in such queries.
+
+``CONFIGURE`` statements and schema definition language (SDL) can **not** be
+parametrized. And data definition language (DDL) has limitations on
+what can be parametrized, so it's not recommended to parametrize
+DDL statements.
+
+Some of the limitations can be lifted in the future versions.
```
1st1, line 131:
```suggestion
Some of the limitations might be lifted in the future versions.
```

docs/edgeql/expressions/parameters.rst
tailhook: "Isn't provided" is not true. We require `fetchone(..., param=None)` explicitly. And would fail before reaching postgres otherwise. So I couldn't come up with anything more clear than `nil`.

docs/edgeql/expressions/parameters.rst
tailhook: Yes. We don't have a term for that. And this is a problem. This one waits for fixing https://github.com/edgedb/edgedb/pull/1373#pullrequestreview-403197099

docs/edgeql/expressions/parameters.rst
tailhook: Changed.

docs/edgeql/expressions/parameters.rst
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+See :ref:`ref_eql_types` for more info on specifying types.
+
+
+Optional Parameters
+-------------------
+
+By default, query parameters are ``REQUIRED`` which means that the query would fail
+if parameter value is nil. You can use ``OPTIONAL`` keyword for optional
```
1st1, line 76: Then this:

```suggestion
By default, query parameters are ``REQUIRED`` which means that the query would fail
if parameter value is an empty set. You can use ``OPTIONAL`` keyword for optional
```

docs/edgeql/expressions/parameters.rst
```diff
@@ -0,0 +1,171 @@
...
+    for(let [name, age] of [["John", 33], ["Jack", null]]) {
+        await conn.fetchOne("""
+            INSERT Person {
+                first_name := <str>$name,
+                age := <OPTIONAL str>$age,
+            }
+        """, {name, age})
+    }
+
+Note: there is a language-specific nil value, ``None`` in Python example and
```
1st1, line 104:
```suggestion
Note: denoting an EdgeQL empty set is language-specific. We use ``None`` in Python and
```

docs/edgeql/expressions/parameters.rst
1st1: Alright, see the below suggestion.

docs/edgeql/expressions/parameters.rst
tailhook: I'm sure that "empty set" would be confusing to me. `set()` is an empty set in python. But, let's wait for user feedback.

docs/edgeql/expressions/parameters.rst
```diff
@@ -101,8 +101,8 @@ Example in JavaScript:
         """, {name, age})
     }
 
-Note: denoting an EdgeQL empty set is language-specific. We use ``None`` in Python and
-``null`` in JavaScript.
+Note: denoting an EdgeQL empty set is language-specific. We use ``None`` in
```
1st1, line 104: I agree it's a bit confusing. Last nit-edit suggestion:

```suggestion
Note: denoting a *scalar EdgeQL empty set* is language-specific. We use ``None`` in
```

PR #1376, opened 2020-04-30
Propagate dev_mode and psql path to CLI

tailhook:
Fixes edgedb/edgedb-cli#32

Requires edgedb/edgedb-cli#34 to be applied first.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -298,9 +298,9 @@ class develop(setuptools_develop.develop):
     def run(self, *args, **kwargs):
         _check_rust()
         build = self.get_finalized_command('build')
-        build_base = pathlib.Path('build').resolve()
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -298,10 +298,12 @@ class develop(setuptools_develop.develop):
     def run(self, *args, **kwargs):
         _check_rust()
         build = self.get_finalized_command('build')
+        build_base = pathlib.Path('build').resolve()
```
elprans, line 301: Use `build.build_base` here.

tailhook: Rather than fixing this. I've updated all `Path('build')` invocations to use `build_base` everywhere as a separate commit.

PR #1382, opened 2020-05-06
Fix erroneous limit injection into singleton properties in shapes

elprans:
Implicit limit is only applicable to multi properties and links.  What's
worse, the injection will trigger a false error when protected
properties like `id` are selected:

    db> SELECT Foo {id}
    error: cannot assign to id

Fix this by omitting the implicit limit injection on pointers that are
known to be singletons.

Per report from @RobertoPrevato

### DIFF
--- a/tests/test_server_proto.py
+++ b/tests/test_server_proto.py
@@ -2548,6 +2548,8 @@ class TestServerProtoDDL(tb.NonIsolatedDDLTestCase):
                 r"""
                     WITH MODULE test
                     SELECT FL_B {
+                        id,
+                        __type__,
                         a,
                     } ORDER BY .n
                 """,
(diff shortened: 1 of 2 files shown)


tests/test_server_proto.py
```diff
@@ -2548,6 +2548,7 @@ async def test_server_proto_fetch_limit_01(self):
                 r"""
                     WITH MODULE test
                     SELECT FL_B {
+                        id,
```
1st1, line 2551: Can you also add a separate test for `__tid__`?

elprans: > Can you also add a separate test for `__tid__`?

`__tid__` is magical and is not selectable explicitly.

elprans: I added a test for `__type__`, which *can* be selected and is also protected.

PR #1392, opened 2020-05-12
Add `--version` option

tailhook:
Perhaps, this can make writing tests for `edgedb server` easier. Regardless of that it looks like very expected thing on (almost) any executable.

### DIFF
--- a/edb/server/main.py
+++ b/edb/server/main.py
@@ -515,6 +515,9 @@ _server_options = [
         '--max-protocol', type=str, callback=_protocol_version,
         default='.'.join(map(str, mng_port.CURRENT_PROTOCOL)),
         help='maximum supported and advertized client protocol version'),
+    click.option(
+        '--version', is_flag=True,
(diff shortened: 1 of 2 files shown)


1st1:
@elprans I think we have some rather complicated versioning machinery, would this one liner do the correct thing?

edb/server/main.py
```diff
@@ -515,6 +515,7 @@ def _protocol_version(
         '--max-protocol', type=str, callback=_protocol_version,
         default='.'.join(map(str, mng_port.CURRENT_PROTOCOL)),
         help='maximum supported and advertized client protocol version'),
+    click.version_option(),
```
elprans, line 518: We need to use `buildmeta.get_version()` here for consistency with `sys::get_version()` and server logs.

tailhook: Fixed. Unfortunately `buildmeta.get_version()` doesn't work at import time, hence PR is a bit more complex.

PR #1395, opened 2020-05-13
Fix an issue with SDL ordering.

### LINKED ISSUE #1383: Cannot access inherited property in computed property
schema
```
type Class extending HasAvailability {
      property schedule := (
        SELECT (.<class[IS Course]).scheduledAt
      );
    }
    abstract type HasAvailability {
     multi link availableAt -> TimeSpot;
    }
    abstract type HasSchedule {
     multi link scheduledAt -> TimeSpot;
    }
    type Course extending HasSchedule{
    }
```

```
object type 'default::Course' has no link or property 'scheduledAt'
```

Note that it works when used in a query

vpetrovykh:
The computables were not tracing the dependency correctly, resulting in
occasional issues with SDL migrations.

When a computable contains a backwards link the dependency is now
using the correct link direction for tracking.

Also only pointers that have the same source as the computable are
"inherited" (to preserve link properties). All other paths get treated
as an arbitrary expression.

Fixes #1383

### DIFF
--- a/edb/edgeql/tracer.py
+++ b/edb/edgeql/tracer.py
@@ -65,7 +65,7 @@ class ObjectType(Type):
         super().__init__(name)
         self.pointers = {}
 
-    def is_pointer(self):
+    def is_pointer(self) -> bool:
         return False
 
     def getptr(self, schema, name):
@@ -86,7 +86,7 @@ class Pointer:
         self.target = target
         self.pointers = {}
 
-    def is_pointer(self):
+    def is_pointer(self) -> bool:
         return True
 
     def getptr(self, schema, name):
@@ -94,6 +94,9 @@ class Pointer:
 
     def get_target(self, schema):
         return self.target
+
+    def get_source(self, schema):
+        return self.source
 
     def get_name(self, schema):
         return self.name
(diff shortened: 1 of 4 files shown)


edb/edgeql/tracer.py
```diff
@@ -86,7 +86,7 @@ def __init__(self, name, *, source=None, target=None):
         self.target = target
         self.pointers = {}
 
-    def is_pointer(self):
+    def is_pointer(self) -> bool:
```
1st1, line 89: Speaking of typing, could you please strict-type the rest of this file and related files in a follow up PR?

edb/edgeql/tracer.py
```diff
@@ -406,9 +409,9 @@ def trace_Path(node: qlast.Path, *,
                     if ptr is None:
                         # Invalid pointer reference, bail.
                         return
-
-                    source_name = ptr.source.get_name(ctx.schema)
-                    ctx.refs.add(f'{source_name}@{step.ptr.name}')
+                    else:
```
1st1, line 412: Why this `else`? It does nothing at the cost of reduced readability.

edb/edgeql/tracer.py
vpetrovykh: OK, I'll try typing this file.

edb/edgeql/tracer.py
vpetrovykh: Yeah, this is definitely a remnant of some intermediate code I was trying out. Removed.

PR #1396, opened 2020-05-13
A train of fixes

elprans:
A bunch of assorted fixes that accumulated in my various branches.


edb/repl/render_json.py
```diff
@@ -185,6 +185,7 @@ def _bool(
 def _empty(
     o: None,
     repl_ctx: context.ReplContext,
+    render_ctx: RenderContext,
```
1st1, line 188: You're still using the Python repl?

edb/schema/functions.py
```diff
@@ -646,10 +644,12 @@ def param_is_inherited(
                 p for p in new_params
                 if not param_is_inherited(new_schema, new, p)
             ]
+            add = delta.add_prerequisite
```
1st1, line 647: nit: I'd call right here -- the code would be simpler (albeit a bit more repetitive).

edb/repl/render_json.py
elprans: Not usually.

PR #1399, opened 2020-05-14
Implement string-splitting function: to_array(str, delim)

elprans:
This implements the reverse of `to_str(array, delim)`:

    db> SELECT to_array('1, 2, 3', ', ');
    {['1', '2', '3']}

### DIFF
--- a/tests/test_edgeql_functions.py
+++ b/tests/test_edgeql_functions.py
@@ -2099,6 +2099,27 @@ class TestEdgeQLFunctions(tb.QueryTestCase):
             [''],
         )
 
+    async def test_edgeql_functions_to_array_01(self):
+        await self.assert_query_result(
+            r'''SELECT to_array('one, two, three', ', ');''',
+            [['one', 'two', 'three']],
+        )
+
+        await self.assert_query_result(
+            r'''SELECT to_array('', ', ');''',
+            [[]],
+        )
+
+        await self.assert_query_result(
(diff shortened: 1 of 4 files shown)


1st1:
I don't think using the `to_` conversion helps here. I'd go for `str_split()`.

elprans:
Yes, but we have `to_str(array, delim)` and not `array_join()`, so  `str_split` really will be a special case as far as naming conventions go.

1st1:
Alright. I've looked at our docs and yeah, we better follow the convention now. Green light 🟢

tests/test_edgeql_functions.py
```diff
@@ -2099,6 +2099,22 @@ async def test_edgeql_functions_to_str_08(self):
...
+        )
+
+        await self.assert_query_result(
+            r'''SELECT to_array('', ', ');''',
+            [[]],
+        )
+
+        await self.assert_query_result(
+            r'''SELECT to_array('foo', ', ');''',
+            [['foo']],
```
1st1, line 2115: What happens if the delimiter is an empty string? Please add a test (and document).

elprans: Done.

haikyuu:
A bit late on the party, but why not rename both? `array_join` and `str_split` are better IMHO.

elprans:
@haikyuu Aye, I opened #1401 to discuss this further.

PR #1400, opened 2020-05-15
Add typing annotations to tracer.py.

vpetrovykh:


elprans:
Please add the module to `mypy.ini` so that typechecking actually takes place

vpetrovykh:
> Please add the module to `mypy.ini` so that typechecking actually takes place

Oh, it felt like I was forgetting something. I'll update it.

edb/edgeql/tracer.py
```diff
@@ -368,28 +416,31 @@ def trace_Path(node: qlast.Path, *,
...
-                    src_name = src.get_name(ctx.schema)
-                    if isinstance(src, Pointer) and src.source is not None:
-                        src_src_name = src.source.get_name(ctx.schema)
-                        source_name = f'{src_src_name}@{src_name}'
-                    else:
-                        source_name = src_name
-                    ctx.refs.add(f'{source_name}@{step.ptr.name}')
+                    return None
+
+                elif isinstance(ptr, (s_links.Link, Pointer)):
```
1st1, line 427: @elprans do we have this weird union type here?

elprans: Yes.  When we trace SDL to extract the dependency graph we use makeshift standins for schema objects that are found in the SDL file being analyzed.  Existing schema objects would be regular schema objects, hence this union.

1st1: Alright. Is this PR ready?

vpetrovykh: I've added some cosmetic changes to imports. I'm trying to avoid name clashes between `typing` and the local classes, so I import specific items from `typing` rather than `from typing import *` or `import typing`. Selectively importing items from typing is more consistent with `from typing import *` used everywhere else in the project.

edb/edgeql/tracer.py
```diff
@@ -59,60 +67,85 @@ class Type(NamedObject):
...
-    subject: typing.Optional[sn.Name] = None,
-    path_prefix: typing.Optional[sn.Name] = None,
-    module: typing.Optional[str] = None,
-    objects: typing.Dict[str, object],
-    params: typing.Dict[str, str],
-) -> typing.FrozenSet[sn.Name]:
+    schema: s_schema.Schema,
+    source: Optional[sn.Name] = None,
+    subject: Optional[sn.Name] = None,
+    path_prefix: Optional[Union[str, sn.Name]] = None,
```
elprans, line 144: `Name` is a subclass of `str`, so the Union isn't needed strictly speaking.

vpetrovykh: Hmmm... I'm about 99% certain that mypy complained if I only used `sn.Name`, but let me try removing that to reduce noise, if possible.

vpetrovykh: I've replaced the Union with just `str`, leaving other places where it's pure `sn.Name` unchanged.

PR #1404, opened 2020-05-23
Document increment/decrement operations for UPDATE.

### LINKED ISSUE #165: We need `+=` and `-=` operators for UPDATE
When updating some link with `*` cardinality it is often useful to only add or remove targets rather than redefine the entire target set. To this end we can add `+=` and `-=` operators in place of `:=`. They would only appear in `SET` clause of `UPDATE` statements.

vpetrovykh:
Add documentation and examples for `+=` and `-=` assignemnts that can be
used in the `SET` block of an `UPDATE`.

Closes #165.

### DIFF
--- a/docs/edgeql/statements/update.rst
+++ b/docs/edgeql/statements/update.rst
@@ -43,12 +43,24 @@ selected by *update-selector-expr* and, optionally, filtered by
     information.
 
 :eql:synopsis:`SET <shape>`
-    A :ref:`shape <ref_eql_expr_shapes_update>` expression with
-    the new values for the links of the updated object.
+    A :ref:`shape <ref_eql_expr_shapes_update>` expression with the
+    new values for the links of the updated object. There are three
+    possible assignment operations permitted within the ``SET`` shape:
 
     .. eql:synopsis::
 
-        SET { <link> := <update-expr> [, ...] }
+        SET { <field> := <update-expr> [, ...] }
+
+        SET { <field> += <update-expr> [, ...] }
+
+        SET { <field> -= <update-expr> [, ...] }
+
+    The most basic assignment is the ``:=``, which just sets the
+    :eql:synopsis:`<field>` to the specified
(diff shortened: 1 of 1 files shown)


docs/edgeql/statements/update.rst
```diff
@@ -43,13 +43,25 @@ selected by *update-selector-expr* and, optionally, filtered by
...
-    the new values for the links of the updated object.
+    A :ref:`shape <ref_eql_expr_shapes_update>` expression with the
+    new values for the links of the updated object. There are three
+    possible assignment operations permitted within the ``SET`` shape:
 
     .. eql:synopsis::
 
         SET { <link> := <update-expr> [, ...] }
 
+        SET { <link> += <update-expr> [, ...] }
```
elprans, line 54: The use of link here is somewhat misleading, some may think only links are allowed in the `UPDATE` shape.  What about properties?

elprans:
Let's also add examples to the cheatsheet (and the tutorial).

docs/edgeql/statements/update.rst
vpetrovykh: I suppose I could call it "name" instead or "field". The latter being a familiar generic reference to what we internally dub "pointers". At the end of the day it's just a placeholder, so "field" is probably what most people would intuitively understand best.

vpetrovykh:
> Let's also add examples to the cheatsheet (and the tutorial).

Yeah, I'm working on the cheatsheet update, too.

vpetrovykh:
> You also need to add `+=` and `-=` operators to https://edgedb.com/docs/edgeql/funcops/set

They are not really operators, they represent assignments, exactly like `:=` is not an operator. They cannot appear anywhere where operators can, they don't appear in the operators list in the introspection, they don't have a signature even. The thing on their left is a symbolic label, not a set or type. Correct me if I'm wrong, but they are only allowed in the context of `UPDATE ... SET`, which is where they are described.

PR #1405, opened 2020-05-23
Update cheatsheet.

vpetrovykh:
Update the REPL and UPDATE sections.


docs/cheatsheet/update.rst
```diff
@@ -55,7 +55,7 @@ Updating a multi link by adding one more item:
         AND
         .directors.last_name = 'Villeneuve'
     SET {
-        actors := .actors UNION (
+        actors += (
```
1st1, line 58: Can we also add -= etc?

vpetrovykh: Example added.

Issue #1107, opened 2020-01-15
Investigate logic of `Flush` message


tailhook:
# Motivation

When we were using `Flush` instead of `Sync` in the (Rust) REPL the shell could hang on error (errors aren't flushed or something)

# Investigate

1. Take a look at `Flush`, it might not flush everything as it should
2. Errors should always flush (or at least by default), even if there is no `Flush` message, there is no sense in optimizing that failure path.

1st1:
> Take a look at Flush, it might not flush everything as it should

It does, but it might take some time for the buffer to be actually sent by uvloop/kernel.

> Errors should always flush (or at least by default), even if there is no Flush message, there is no sense in optimizing that failure path.

Yes, let's do that. The current behavior is kind of an oversight, not an optimization.

elprans:
> It does, but it might take some time for the buffer to be actually sent by uvloop/kernel.

Do we use `TCP_NODELAY`?

1st1:
it's on by default in 3.8

1st1:
i'd suggest to explicitly sync for now and in principle. Using flush is needed only for edge cases.

tailhook:
> It does, but it might take some time for the buffer to be actually sent by uvloop/kernel.

Well, I waited at least on the order of seconds. It's not ndelay.

> i'd suggest to explicitly sync for now and in principle. Using flush is needed only for edge cases.

It what I used to do. But @elprans suggested `Flush`.

Also because Sync breaks transaction, can request fail because schema changed between prepare and execute?

elprans:
> Well, I waited at least on the order of seconds. It's not ndelay.

This is because `flush_on_error = False`

> i'd suggest to explicitly sync for now and in principle. Using flush is needed only for edge cases.

We still need to fix it, though.  Also, `Parse/Execute` with a single `Sync` is conceptually simpler, since it's a single transaction unit.

1st1:
> Also because Sync breaks transaction, can request fail because schema changed between prepare and execute?

It doesn't break explicit transactions.

1st1:
ToDo:

- [ ] Remove hardcode about peaking 'S' messages
- [ ] Flush errors automatically
- [ ] Make sure our Flush triggers Postgres' Flush.
- [ ] Create testing rig for the binary proto
- [ ] Fix edgedb-python and edgedb-js to use Flush instead of Sync after Parse/Describe
[...]

PR #1406, opened 2020-05-25
Implement new protocol testing framework; start fixing flush

1st1:
Please review the commits individually.

@tailhook The second commit that fixes Flush is the one relevant for the rust protocol implementation. Note that it only addresses #1107 partially.


tailhook:
> @tailhook The second commit that fixes Flush is the one relevant for the rust protocol implementation. Note that it only addresses #1107 partially.

I'm still not sure if this is enough to use flush flawlessly.

1st1:
> I'm still not sure if this is enough to use flush flawlessly.

Paul, can you elaborate please? It would also help if you could remind me why we want to send Flush to Postgres -- I can't think of a scenario where this is actually useful.

tailhook:
> > I'm still not sure if this is enough to use flush flawlessly.
> 
> Paul, can you elaborate please?

Still don't recall and had no time to experiment with it. The concern is probably cherry-picking Sync messages here and there (and if we check for sync, why we don't check for flush in the same places?), but no known real use-case yet.

> It would also help if you could remind me why we want to send Flush to Postgres -- I can't think of a scenario where this is actually useful.

I didn't propose and didn't support this change in any way. You have mentioned this as you expected it be like that all the time. If we can send a request to postgres and receive data without sync, we don't need to forward flush to postgres (I don't thing I'll do that in repl, but the documentation looks like it should work).

1st1:
> The concern is probably cherry-picking Sync messages here and there (and if we check for sync, why we don't check for flush in the same places?)

Well, the implementation should work along these lines:

* A user types `SELECT 1; [ENTER]`
* REPL sends `{PREPARE} SELECT 1; {FLUSH}`
* If an error occurs we display it to the user and REPL sends `{SYNC}`
* no error - REPL sends `{EXECUTE}; {SYNC}`

> I didn't propose and didn't support this change in any way.  ... If we can send a request to postgres and receive data without sync, we don't need to forward flush to postgres

Fair enough. We tightly control all messaging with Postgres and issue `{FLUSH}` ourselves when necessary. So I think we're good.

PR #1408, opened 2020-05-25
Reimplement schema reflection

elprans:
This is a clean-slate schema reflection and introspection implementation
that uses generated EdgeQL instead of hardcoded SQL to store and introspect
schemas in the database.

The old approach has numerous problems:

* its implementation predated formal introspection schema, and
  so complex opencoded views are necessary as an emulation layer
  to allow EdgeQL queries on schema metadata, which is both hard
  to maintain and hard to make efficient (views cannot have
  indexes);

* schema readout in `intromech.py` is a hardcoded fragile mess,
  especially with regards to keeping correct dependency read order;

* schema readout is slow because it runs multiple queries.

In contrast, the new implementation is completely autogenerated from Python
class structure in `edb.schema`, uses native EdgeDB layout directly, so no
translation for the public introspection schema is necessary, and the
schema readout happens as a single query that populates the schema object
directly which is more efficient and reliable than going through the DDL
API.


edb/lib/cal.edgeql
```diff
@@ -16,9 +16,7 @@
 # limitations under the License.
 #
 
-CREATE MODULE cal {
-    SET builtin := true;
-};
+CREATE MODULE cal;
```
1st1, line 19: Why this change? Do `schema::Module` objects still have a `builtin` property? How to set it?

elprans: *All* standard schema objects now have the property and have it set automatically at bootstrap.

edb/schema/ddl.py
```diff
@@ -397,6 +404,54 @@ def apply_ddl(
...
+    )
+
+    return schema
+
+
+def apply_ddl_script_ex(
+    ddl_text: str,
+    *,
+    schema: s_schema.Schema,
+    modaliases: Optional[Mapping[Optional[str], str]] = None,
```
1st1, line 431: Probably need to create an alias for `Mapping[Optional[str], str]` (another time, for sure)

edb/schema/reflection/__init__.py
```diff
@@ -17,6 +17,14 @@
 #
 
 
-ALTER MODULE std {
-    SET builtin := true;
-};
+from .reader import parse_into  # NoQA
```
1st1, line 20: Git got interestingly confused here...

edb/schema/reflection/reader.py
```diff
@@ -0,0 +1,227 @@
...
+from edb.schema import schema as s_schema
+
+from . import structure as sr_struct
+
+
+def parse_into(
+    schema: s_schema.Schema,
+    data: Sequence[str],
+    schema_class_layout: Dict[Type[s_obj.Object], sr_struct.SchemaTypeLayout],
+) -> s_schema.Schema:
```
1st1, line 45: Can you add a docstring? My assumption is that we're parsing & applying the introspection JSON here?

edb/common/devmode.py
```diff
@@ -124,27 +124,34 @@ def get_dev_mode_cache_dir() -> os.PathLike:
...
+def write_dev_mode_cache(obj, cache_key, path, *, pickled=True):
     full_path = get_dev_mode_cache_dir() / path
 
     try:
         with tempfile.NamedTemporaryFile(
                 mode='wb', dir=full_path.parent, delete=False) as f:
             f.write(cache_key)
-            pickle.dump(obj, file=f, protocol=pickle.HIGHEST_PROTOCOL)
+            if pickled:
+                pickle.dump(obj, file=f, protocol=pickle.HIGHEST_PROTOCOL)
```
1st1, line 152: Why not call pickle.dump/load directly at write_dev_mode_cache/read_dev_mode_cache callsites? This argument/if check is flaky.

elprans: Because handling unpickling errors at the call sites would be a tedious repetition.  Would a wrapper function be better?

1st1: ah, right. Up to you, wrapper also works. This is nitpicking though.

PR #1410, opened 2020-05-26
docs: Fix enum fields rendering; fix uint/int references

1st1:


docs/internals/protocol/dataformats.rst
```diff
@@ -101,12 +101,23 @@ represented as:
...
 
-The :eql:type:`std::str` values are represented as a UTF-8 encoded byte string.
+The :eql:type:`std::str` values are represented as a UTF-8 encoded byte string,
+prefixed with an ``int32`` length field:
+
+.. code-block:: c
+
+    struct String {
+        // Number of bytes.
+        int32       length;
```
elprans, line 111: Why `int32` and not `uint32`?

elprans: Actually, I don't think this is correct.  Length prefix is part of the wrapping structure, not the encoded data itself.

1st1: https://github.com/edgedb/edgedb-js/blob/master/src/codecs/text.ts#L23-L31

1st1: https://github.com/MagicStack/py-pgproto/blob/master/codecs/text.pyx#L24-L48

PR #1412, opened 2020-05-26
Build CLI in debug mode when installing in devmode

elprans:


tailhook:
I'm not sure this is needed. This enables some extra assertions which is good. But also makes code ~10x slower. In my tests performance is still fine, though.

We currently build release builds with debugging symbols anyway, so tracebacks are there nevertheless. So not sure `--debug` is needed.

elprans:
My motivation for `--debug` is not the symbols but compilation speed.  Release builds are taking an exorbitant amount of time to build with LTO etc.

PR #1415, opened 2020-05-26
Rename `to_str` to `array_join` and `to_array` to `str_split`.

### LINKED ISSUE #1401: Consider renaming `to_str(array)` and `to_array(str)` to `str_split` and `array_join` respectively.
`to_str(array, delimiter)` and `to_array(str, delimiter)` are, at a glance, consistent with other conversion/constructor functions, but `to_array(str, delimiter)` is an odd and undisoverable way of naming a string-splitting function, and `str_split()` would be a better name.  On the other hand, `to_str(array, delimiter)` has a better claim at being similar to other "formatting" functions, but is also distinct in that it converts a collection type into a scalar type, so naming it `array_join` would not be too bad.

vpetrovykh:
Rename `to_str(array, delimiter)` and `to_array(str, delimiter)` to more
discoverable and familiar `array_join(array, delimiter)` and
`str_split(str, delimiter)` respectively.

Fixes #1401

Co-authored-by: Shankar Jha <shankarj67@gmail.com>

### DIFF
--- a/docs/edgeql/funcops/array.rst
+++ b/docs/edgeql/funcops/array.rst
@@ -32,11 +32,8 @@ Array
     * - :eql:func:`find`
       - Find the index of an element in the array.
 
-    * - :eql:func:`to_str`
+    * - :eql:func:`array_join`
       - Render an array to a string.
-
-    * - :eql:func:`to_array`
-      - Split a string into an array using a delimiter.
 
     * - :eql:func:`array_agg`
       - :eql:func-desc:`array_agg`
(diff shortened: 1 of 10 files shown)


vpetrovykh:
There's an issue with this PR at the moment: the `test_dump01_restore_compatibility` is failing. I'm guessing this is because the standard library functions used in the test schema changed. I'm not certain whether I should just generate a fresh alpha3 dump file or if we want some way of handling this with the older dump (or if such handling is even possible).

ambv:
After alphas we will have to handle old dumps for sure but while in alpha in my opinion we can for simplicity just ignore this and recreate the dump file.

1st1:
> There's an issue with this PR at the moment: the test_dump01_restore_compatibility is failing. I'm guessing this is because the standard library functions used in the test schema changed. I'm not certain whether I should just generate a fresh alpha3 dump file or if we want some way of handling this with the older dump (or if such handling is even possible).

What function does that dump depend on? to_str or to_array? How?

I suggest this:

* let's say it depends on `to_array`
* we keep to_array but mark it deprecated (create an issue to delete it before 1.0) and undocument it. We add a note to 1.0 release suggesting the update strategy for dumps created with a2.
* we add str_split and document it.

1st1:
> After alphas we will have to handle old dumps for sure but while in alpha in my opinion we can for simplicity just ignore this and recreate the dump file.

Yeah, in this very case I feel like we can preserve compatibility with a2 dumps, see my comment. It's not for free, but isn't super hard either.

1st1:
@vpetrovykh I've updated this PR branch myself with a few fixes.

PR #1416, opened 2020-05-26
Better numeric constants

tailhook:
This PR removes `0.` support and adds `1_024` support.

The underscore semantics mostly adheres to rust semantics: allow arbitrary number underscores between digits and at the end of the integer, fractional part or exponent, but not at the start of the integer or decimal part. Unlike in Rust starting exponent with underscore (`1e_10` is **not** allowed).

### DIFF
--- a/tests/test_edgeql_expressions.py
+++ b/tests/test_edgeql_expressions.py
@@ -380,6 +380,11 @@ class TestExpressions(tb.QueryTestCase):
                     r'''SELECT <int64>'3689348814741900000000000' ''',
                 )
 
+        with self.assertRaisesRegex(edgedb.EdgeQLSyntaxError,
+                                    'expected digit after dot'):
+            async with self.con.transaction():
+                await self.con.fetchone('SELECT 0. ')
+
     async def test_edgeql_expr_op_02(self):
         await self.assert_query_result(
             r'''SELECT 40 ^ 2;''',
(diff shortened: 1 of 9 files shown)


1st1:
@elprans please also take a look. The semantics of `_` makes sense to me.

tests/test_edgeql_expressions.py
```diff
@@ -380,6 +397,11 @@ async def test_edgeql_expr_literals_02(self):
                     r'''SELECT <int64>'3689348814741900000000000' ''',
                 )
 
+        with self.assertRaisesRegex(edgedb.EdgeQLSyntaxError,
+                                    'expected digit after dot'):
```
elprans, line 401: These sort of tests would probably be better placed in `test_edgeql_syntax`.

1st1: +1. Better keep all syntax tests consolidated in one place.

tailhook: I've move tests from edgeql_expressions. But left other ones in ddl. Or should I move DDL tests too?

elprans: If the purpose of those tests is to test uninterpreted handling of literals without necessarily hitting Postgres, then yes.

tailhook: Done.

PR #1421, opened 2020-05-27
tests: Add 'self.repl()' shortcut to launch repl right in the test

1st1:


edb/testbase/server.py
```diff
@@ -289,6 +291,40 @@ async def connect(cls, *,
             cluster=cluster, database=database, user=user, password=password)
         return await edgedb.async_connect(**conargs)
 
+    def psql(self):
+        """Open interactive psql right in the test.
```
elprans, line 295: Why `psql` and not `edgedb-cli`?

1st1: Good idea, I've implemented that. The only problem is that it uses our legacy REPL because the new CLI doesn't recognize EDGEDB_PASSWORD env var https://github.com/edgedb/edgedb-cli/issues/57

edb/testbase/server.py
```diff
@@ -289,6 +292,36 @@ async def connect(cls, *,
...
+            'python', '-m', 'edb.cli',
+            '--host', conargs['host'],
+            '--port', str(conargs['port']),
+            '--database', self.con.dbname,
+            '--user', conargs['user'],
+        ]
+
+        env = os.environ.copy()
+        if password := conargs.get('password'):
+            env['EDGEDB_PASSWORD'] = password
```
elprans, line 315: Both repls support `--password-from-stdin`, could you use that?

1st1: No, because you need to attach repl process' stdin to the main process' stdin.

PR #1422, opened 2020-05-27
Fix backend_id for std::bigint

1st1:
backend_ids (OIDs) for non-standard Postgres scalar types should
be repopulated after restoring database from a template snapshot
(as OIDs are not persisted across dumps).

### DIFF
--- a/edb/server/bootstrap.py
+++ b/edb/server/bootstrap.py
@@ -497,39 +497,66 @@ async def _init_stdlib(cluster, conn, testmode, global_ids):
     else:
         cluster.restore_database(edbdef.EDGEDB_TEMPLATE_DB, tpldbdump)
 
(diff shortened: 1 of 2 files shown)


edb/server/bootstrap.py
```diff
@@ -497,39 +497,66 @@ async def _init_stdlib(cluster, conn, testmode, global_ids):
...
         )
-
-        schema, sql = compile_bootstrap_script(
+        _, sql = compile_bootstrap_script(
             compiler,
             stdlib.reflschema,
-            tids_query,
+            '''
+            UPDATE schema::ScalarType
+            FILTER .builtin AND NOT .is_abstract
```
elprans, line 521: Filter on `.builtin` is not necessary

PR #1423, opened 2020-05-27
Test both Python & Rust CLI implementations for dump/restore

1st1:


1st1:
@elprans Since we have a bit of time before release please tweak the package infra to install edgedb-cli so that we can test it.

Issue #1320, opened 2020-04-05
Prohibit indexes on non-single pointers


elprans:
Currently an attempt to declare an index on a `multi` pointer just crashes with `InternalServerError: column "target" does not exist`.

1st1:
@vpetrovykh this is low priority one but would be nice to fix soon.

vpetrovykh:
So, does this mean we prohibit indexes with expressions of cardinality > 1?

PR #1424, opened 2020-05-27
Prohibit indexes on non-single expressions.

vpetrovykh:
Ensure that the index expression cardinality is never more than ONE.

Fixes: #1320

### DIFF
--- a/edb/schema/indexes.py
+++ b/edb/schema/indexes.py
@@ -291,7 +291,7 @@ class CreateIndex(
                 singletons = [subject]
                 path_prefix_anchor = qlast.Subject().name
 
-            return type(value).compiled(
+            expr = type(value).compiled(
                 value,
                 schema=schema,
                 options=qlcompiler.CompilerOptions(
@@ -302,6 +302,15 @@ class CreateIndex(
                     singletons=frozenset(singletons),
                 ),
             )
+
+            # Check that the inferred cardinality is no more than 1
+            if expr.irast.cardinality.is_multi():
+                raise errors.ResultCardinalityMismatchError(
+                    f'possibly more than one element returned by '
+                    f'the index expression where only singletons '
+                    f'are allowed')
+
+            return expr
         else:
             return super().compile_expr_field(schema, context, field, value)
 
--- a/tests/test_edgeql_ddl.py
+++ b/tests/test_edgeql_ddl.py
@@ -5113,3 +5113,41 @@ class TestEdgeQLDDL(tb.DDLTestCase):
             ''',
             [1],
         )
+
+    async def test_edgeql_ddl_index_01(self):
+        with self.assertRaisesRegex(
+            edgedb.ResultCardinalityMismatchError,
+            r"possibly more than one element returned by the index expression"
+        ):
+            await self.con.execute(r"""
+                CREATE TYPE Foo {
+                    CREATE MULTI PROPERTY a -> int64;
+                    CREATE INDEX ON (.a);
+                }
+            """)
+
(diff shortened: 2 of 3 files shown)


tests/test_edgeql_ddl.py
```diff
@@ -5113,3 +5113,41 @@ async def test_edgeql_ddl_required_08(self):
...
+
+    async def test_edgeql_ddl_index_01(self):
+        with self.assertRaisesRegex(
+            edgedb.ResultCardinalityMismatchError,
+            r"the index expression has cardinality MANY"
+        ):
+            await self.con.execute(r"""
+                CREATE TYPE Foo {
+                    CREATE MULTI PROPERTY a -> int64;
+                    CReATE INDEX ON (.a);
```
elprans, line 5125: Please fix capitalization here.

tests/test_edgeql_ddl.py
```diff
@@ -5113,3 +5113,41 @@ async def test_edgeql_ddl_required_08(self):
...
+    async def test_edgeql_ddl_index_02(self):
+        with self.assertRaisesRegex(
+            edgedb.ResultCardinalityMismatchError,
+            r"the index expression has cardinality MANY"
+        ):
+            await self.con.execute(r"""
+                CREATE TYPE Foo {
+                    CREATE PROPERTY a -> int64;
+                    CREATE PROPERTY b -> int64;
+                    CReATE INDEX ON ({.a, .b});
```
elprans, line 5138: .. and here.

tests/test_edgeql_ddl.py
```diff
@@ -5113,3 +5113,41 @@ async def test_edgeql_ddl_required_08(self):
...
+    async def test_edgeql_ddl_index_03(self):
+        with self.assertRaisesRegex(
+            edgedb.ResultCardinalityMismatchError,
+            r"the index expression has cardinality MANY"
+        ):
+            await self.con.execute(r"""
+                CREATE TYPE Foo {
+                    CREATE PROPERTY a -> int64;
+                    CREATE PROPERTY b -> int64;
+                    CReATE INDEX ON (array_unpack([.a, .b]));
```
elprans, line 5151: .. and here

edb/schema/indexes.py
```diff
@@ -302,6 +302,16 @@ def compile_expr_field(
                     singletons=frozenset(singletons),
                 ),
             )
+
+            # Check that the inferred cardinality is no more than 1
+            expr_card = expr.irast.cardinality
+            if expr_card.is_multi():
+                schema_card = expr_card.to_schema_value()[1]
+                raise errors.ResultCardinalityMismatchError(
+                    f'the index expression has cardinality {schema_card}, '
```
elprans, line 311: Please use `possibly more than one element returned by an expression where only singletons are allowed` like we do in all other places where we require singleton expressions.

tests/test_edgeql_ddl.py
vpetrovykh: OK

tests/test_edgeql_ddl.py
vpetrovykh: OK

tests/test_edgeql_ddl.py
vpetrovykh: OK

edb/schema/indexes.py
vpetrovykh: Updated the error message

1st1:
Please also update the changelog :)

PR #1425, opened 2020-05-27
Add a hint when extending scalar type by object

### LINKED ISSUE #1016: Bad error reporting on create type
```
CREATE TYPE MyStr EXTENDING str;                                     
ERROR: InternalServerError: 'NoneType' object has no attribute 'is_schema_field'
```
Version `1.0-alpha.2+dev.391.g2276144b`

I obviously needed `CREATE SCALAR TYPE` and would expect it to be grammar error (i.e. something like `TYPE is unexpected, expected one of SCALAR, DATABASE, ...`

tailhook:
Fixes #1016

### DIFF
--- a/edb/errors/compiler.py
+++ b/edb/errors/compiler.py
@@ -3,43 +3,50 @@ from edb import schema
 
 
 class BadTypeObject(errors.InvalidReferenceError):
-    def __init__(self, obj_id, obj_type, target_type, obj_name=None):
-        self.obj_id = obj_id
-        self.obj_type = obj_type.get_schema_class_displayname()
-        self.obj_name = obj_name
-        self.target_type = target_type.get_schema_class_displayname()
+    def __init__(self, obj_id, obj_type, target_type,
+                 obj_name=None, context=None):
+        self._obj_id = obj_id
+        self._obj_type = obj_type.get_schema_class_displayname()
+        self._obj_name = obj_name
+        self._target_type = target_type.get_schema_class_displayname()
+        self._context = context
         super().__init__(
             f'schema object {obj_id!r} exists, but is not '
             f'{target_type.get_schema_class_displayname()}')
 
-    def update(self, obj_name=None, statement=None):
-        self.obj_name = obj_name or self.obj_name
+    def update(self, obj_name=None, source_context=None, statement=None):
+        self._obj_name = obj_name or self._obj_name
+        if source_context:
+            self._context = source_context
+            self.set_source_context(source_context)
 
         if statement:
             # Time to finalize our error, with potentially new details
-            if self.obj_name:
+            if self._obj_name:
                 message = (
-                    f'schema object `{self.obj_name}` '
-                    f'(uuid: {self.obj_id}) exists, but is not '
-                    f'{self.target_type}'
+                    f'schema object `{self._obj_name}` '
+                    f'(uuid: {self._obj_id}) exists, but is not '
+                    f'{self._target_type}'
                 )
             else:
                 message = (
-                    f'schema object {self.obj_id!r} exists, but is not '
-                    f'{self.target_type}'
+                    f'schema object {self._obj_id!r} exists, but is not '
+                    f'{self._target_type}'
                 )
 
             # specific error hints go here
             if isinstance(statement, schema.objtypes.CreateObjectType):
-                if self.obj_type == 'scalar type':
+                if self._obj_type == 'scalar type':
                     return errors.InvalidReferenceError(
                         message,
-                        hint=f'name refers to the {self.obj_type}',
+                        hint=f'name refers to the {self._obj_type}',
                         details=f'consider using CREATE SCALAR TYPE',
+                        context=self._context,
                     )
 
             # if nothing else, just convert to regular error
-            return errors.InvalidReferenceError(message)
+            return errors.InvalidReferenceError(message,
+                                                context=self._context)
 
         return self
 
--- a/edb/errors/compiler.py
+++ b/edb/errors/compiler.py
@@ -1,0 +1,49 @@
+from edb import errors
+from edb import schema
+
+
+class BadTypeObject(errors.InvalidReferenceError):
+    def __init__(self, obj_id, obj_type, target_type, obj_name=None):
+        self.obj_id = obj_id
+        self.obj_type = obj_type.get_schema_class_displayname()
+        self.obj_name = obj_name
+        self.target_type = target_type.get_schema_class_displayname()
+        super().__init__(
+            f'schema object {obj_id!r} exists, but is not '
+            f'{target_type.get_schema_class_displayname()}')
+
+    def update(self, obj_name=None, statement=None):
+        self.obj_name = obj_name or self.obj_name
+
+        if statement:
+            # Time to finalize our error, with potentially new details
+            if self.obj_name:
+                message = (
+                    f'schema object `{self.obj_name}` '
+                    f'(uuid: {self.obj_id}) exists, but is not '
+                    f'{self.target_type}'
+                )
+            else:
+                message = (
+                    f'schema object {self.obj_id!r} exists, but is not '
+                    f'{self.target_type}'
+                )
+
+            # specific error hints go here
+            if isinstance(statement, schema.objtypes.CreateObjectType):
+                if self.obj_type == 'scalar type':
+                    return errors.InvalidReferenceError(
+                        message,
+                        hint=f'name refers to the {self.obj_type}',
+                        details=f'consider using CREATE SCALAR TYPE',
+                    )
+
+            # if nothing else, just convert to regular error
+            return errors.InvalidReferenceError(message)
+
+        return self
+
+    def __reduce__(self):
+        # This error is internal and should be converted into
+        # InvalidReferenceError on process boundary
+        return (errors.InvalidReferenceError, self.args)
--- a/edb/schema/inheriting.py
+++ b/edb/schema/inheriting.py
@@ -21,6 +21,7 @@ from __future__ import annotations
 from typing import *
 
 from edb import errors
+from edb.errors import compiler as c_errors
 
 from edb.common import struct
 from edb.edgeql import ast as qlast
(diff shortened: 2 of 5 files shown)


tailhook:
This is how it looks like now:
create_type_hint

edb/schema/inheriting.py
```diff
@@ -378,12 +379,29 @@ def _classbases_from_ast(
...
+                    metaclass=cls.get_schema_metaclass(),
+                )
+            except errors.InvalidReferenceError as initial_error:
+                from edb.schema import scalars as s_scalars
+                try:
+                    shell = utils.ast_to_object_shell(
+                        b,
+                        modaliases=modaliases,
+                        schema=schema,
+                        metaclass=s_scalars.ScalarType,
```
elprans, line 396: Use `ast_to_object`, omit `metaclass` argument and use `get_verbosename()` on the returned object.  That way you don't hardcode the check for scalars and are able to check for general metaclass mismatch.

tailhook: 1. I don't know what to do later. I can suggest `CREATE SCALAR TYPE` only for scalars. What `get_verbosename()` gives me is something like `scalar type 'xxx'`. Should I use `startswith`? (this doesn't work for enums) Should I use `get_schema_class_displayname()`?
2. Why use `ast_to_object` if that's a wrapper around shell with an extra step?

elprans: If you want just the `SCALAR TYPE` part, then `get_schema_class_displayname()` would do.

tailhook: Done

elprans: You still have the check hardcoded to scalars.  Remove `metaclass=` to make this work for any mismatched inheritance.

tailhook: Sorry, forgot this one. It doesn't change anything, though, because the hint is specifically for scalar types. I don't know what hint should be for other types. And basically I can't even recall what types can even be besides scalars and objects off the top of my head.

elprans: All inheritance goes through this codepath, including links, constraints etc.

tailhook: So I'm not sure what else I should do? Any test that I can do for non-types?

tailhook:
Current test failure I don't understand:
```
edgedb> CREATE TYPE test::ScalarType06 EXTENDING str;
ERROR: InvalidReferenceError: improperly formed name 'str': module is not specified
  Hint: type is scalar
  Detail: consider using CREATE SCALAR TYPE
edgedb> CREATE TYPE test::ScalarType06 EXTENDING std::str;
error: object type 'std::str' does not exist
  ┌─ query:1:42
  │
1 │ CREATE TYPE test::ScalarType06 EXTENDING std::str;
  │                                          ^^^^^^^^^ type is scalar
  │
  = consider using CREATE SCALAR TYPE

```
This looks like some recent change (after the initial PR) and it looks like an unrelated bug.

elprans:
> This looks like some recent change (after the initial PR) and it looks like an unrelated bug.

Fixed in #1546

1st1:
@elprans is this ready to be merged? @tailhook please rebase.

1st1 (approved):

tailhook:
I think it's not ready because of this:
>> All inheritance goes through this codepath, including links, constraints etc.
>
> So I'm not sure what else I should do? Any test that I can do for non-types?

1st1:
> I think it's not ready because of this:

Well, you could gate this with an `if` that would check if the ast node is CREATE TYPE. But ideally we should see how we handle cases like `CREATE TYPE blah EXTENDING some_constraint` and `CREATE ABSTRACT CONSTRAINT blah EXTENDING some_type` etc.

tailhook:
Fixed hint for `CREATE TYPE` -> `CREATE SCALAR TYPE` and also restore context (code span) for all inheritance errors (there is no code span for all such errors in the current master).

edb/errors/compiler.py
```diff
@@ -0,0 +1,56 @@
...
+        self._obj_id = obj_id
+        self._obj_type = obj_type.get_schema_class_displayname()
+        self._obj_name = obj_name
+        self._target_type = target_type.get_schema_class_displayname()
+        self._context = context
+        super().__init__(
+            f'schema object {obj_id!r} exists, but is not '
+            f'{target_type.get_schema_class_displayname()}')
+
+    def update(self, obj_name=None, source_context=None, statement=None):
```
1st1, line 17: I'd prefer not to create exceptions outside of `edb.errors`. We started with this and it took us quite some time to clean everything up. I'd suggest generalizing this and implementing directly in the `edb.errors.base.EdgeDBError` class.

@elprans thoughts?

elprans: Technically, it's not a new error type since it's reduced back to `InvalidReferenceError`, but I'm a bit worried about the error "update" pattern.  Perhaps passing the necessary context to `schema.get` is cleaner?

1st1: > Technically, it's not a new error type since it's reduced back to InvalidReferenceError

But that's my point. I don't want us to have 10s of intermediate error types reducing to `edb.errors` eventually. This complexity is unnecessary and is a sure road to spaghetti code.

tailhook: So here is the original traceback, marked with places where we add context:
```
      File "/work/edb/schema/types.py", line 1987, in _create_begin
        schema = super()._create_begin(schema, context)
      File "/work/edb/schema/inheriting.py", line 595, in _create_begin
        schema = super()._create_begin(schema, context) <<< HERE WE ADD STATEMENT
      File "/work/edb/schema/inheriting.py", line 44, in _create_begin
        schema = super()._create_begin(schema, context)
      File "/work/edb/schema/delta.py", line 2119, in _create_begin
        self.validate_create(schema, context)
      File "/work/edb/schema/types.py", line 2017, in validate_create
        bases = self.get_resolved_attribute_value(
      File "/work/edb/schema/delta.py", line 1804, in get_resolved_attribute_value
        value = self._resolve_attr_value(
      File "/work/edb/schema/delta.py", line 361, in _resolve_attr_value
        val = v.resolve(schema) <<< HERE WE ADD CONTEXT (code span)
      File "/work/edb/schema/types.py", line 473, in resolve
        return schema.get(
      File "/work/edb/schema/schema.py", line 339, in get
        return self.get_generic(
      File "/work/edb/schema/schema.py", line 1690, in get_generic
        return self._base_schema.get(
      File "/work/edb/schema/schema.py", line 339, in get
        return self.get_generic(
      File "/work/edb/schema/schema.py", line 1238, in get_generic
        obj = self._get(name, <<< HERE WE ADD REAL NAME OF THE OBJECT
      File "/work/edb/schema/schema.py", line 962, in _get
        result = getter(self, fqname)
      File "/work/edb/schema/schema.py", line 1232, in getter
        obj = schema.get_by_id(obj_id, type=type, default=None)
      File "/work/edb/schema/schema.py", line 1174, in get_by_id
        raise errors.InvalidReferenceError(
    edb.errors.InvalidReferenceError: schema object UUID('00000000-0000-0000-0000-000000000101') exists, but is not object type
```

Roughly:
1. Statement layer should be brought 13 layers below
2. Context (code span) should be brought 8 layers below
3. Real name should be brought 3 layers below

This roughtly translates as:
1. 13 functions should have a new argument
2. 8 function should have 2 new arguments
3. 3 functions should have 3 new arguments

Looking at it now I think it's probably possible to merge 2-3 into one point, but this doesn't improve situation to a manageable level.

While I understand that you need to wrap your head around that, but generally providing more context to the error is not only hugely used in Rust, it's also quite often used in python, like when you do:
```
except KeyError as e:
    raise FormValidationError(...)
```
... you essentially add more information to the error. This is also cheaper on success  part (comparing to passing all that context via args through call chain), and probably even on error case too.

We can bikeshed about how specifically we add that (i.e. method is called `update` rather than `context` like in Rust, because `context` term is used for source code position, etc.).

> I'd suggest generalizing this and implementing directly in the edb.errors.base.EdgeDBError class.

I'm not sure that `update()` method at the base error is very good, as many errors could have different context information needed. Please share your thoughts on how this should be generalized.

> But that's my point. I don't want us to have 10s of intermediate error types reducing to edb.errors eventually. This complexity is unnecessary and is a sure road to spaghetti code.

1. We have fields in this error object that can't be serialized efficiently and are useless outside of the compiler anyway
2. Without fields, I don't think it's viable to parse that information from the text message

Generally I don't see any problem in compiler-only exceptions. We already have exceptions, like `KeyError`, `ValueError`, etc. that should not leave compiler boundary and should be translated into a valid client error.

elprans: The passing of a bunch of contextual information trough the stack isn't new in this piece of code, which is why we pass `DeltaContext` everywhere. Here we just need to see how we can associate the extra info with that.  For example, the source span is already present in the `source_context` attribute of the current DDL command, which you can get from the context.  Likewise we can retain the reference to the original AST there also.

Thus, no new arguments would be needed, except at the `schema.py` level, where the call stack is fairly shallow and there is already precedent in other similar arguments.

tailhook: > The passing of a bunch of contextual information trough the stack isn't new in this piece of code, which is why we pass DeltaContext everywhere.

It stops propagating at one layer before resolve, where I add code span. Which is fixable, but...

> For example, the source span is already present in the source_context

This is not that source context. We need to pick specific name not the whole statement.

> Thus, no new arguments would be needed, except at the schema.py level, where the call stack is fairly shallow

I still see 8 layers above resolve. And CommandContext is not enough, we should make a more specific context object for that, that includes specific name.

> and there is already precedent in other similar arguments.

This is specifically frightening. Just a single simple hint needs another argument in the core and otherwise simple function `get_by_id`. More hints -- more places like that.

----

That said, it looks like we can aggregate all three `update`'s into a single error transformation at around `resolve` call, and we can use `CommandContext` instead of statement itself when figuring out hint. But I'm not convinced that we should pass the whole context forward.

msullivan:
I think I'd call https://github.com/edgedb/edgedb/pull/2896 good enough for this?

Issue #1339, opened 2020-04-13
ISE when not submitting enough query arguments


1st1:
```
>>> c.fetchone('select <int64>$0 + 1')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/yury/dev/venvs/edgedb/lib/python3.8/site-packages/edgedb/blocking_con.py", line 107, in fetchone
    return self._protocol.sync_execute_anonymous(
  File "edgedb/protocol/blocking_proto.pyx", line 91, in edgedb.protocol.blocking_proto.BlockingIOProtocol.sync_execute_anonymous
  File "edgedb/protocol/blocking_proto.pyx", line 77, in edgedb.protocol.blocking_proto.BlockingIOProtocol._iter_coroutine
  File "edgedb/protocol/protocol.pyx", line 542, in execute_anonymous
  File "edgedb/protocol/protocol.pyx", line 339, in _execute
edgedb.errors.InternalServerError: bind message supplies 0 parameters, but prepared statement "" requires 1
```

RobertoPrevato:
I cannot reproduce this issue anymore. Now a user friendly exception is thrown: `edgedb.errors.QueryError: invalid argument count, expected: 1, got: 0`.

```
/edgedb-python/edgedb/blocking_con.py", line 107, in fetchone
    return self._protocol.sync_execute_anonymous(
  File "edgedb/protocol/blocking_proto.pyx", line 91, in edgedb.protocol.blocking_proto.BlockingIOProtocol.sync_execute_anonymous
  File "edgedb/protocol/blocking_proto.pyx", line 77, in edgedb.protocol.blocking_proto.BlockingIOProtocol._iter_coroutine
  File "edgedb/protocol/protocol.pyx", line 541, in execute_anonymous
  File "edgedb/protocol/protocol.pyx", line 338, in _execute
edgedb.errors.QueryError: invalid argument count, expected: 1, got: 0
```

PR #1427, opened 2020-05-27
Adds tests for #1339

RobertoPrevato:


tests/test_server_proto.py
```diff
@@ -685,6 +685,27 @@ async def test_server_proto_args_07(self):
             await self.con.fetchone(
                 'select schema::Object {name} filter .id=$id', id='asd')
 
+    async def test_server_proto_args_08(self):
```
elprans, line 688: I'd add a similar test for named args also

RobertoPrevato: You had a very good point: indeed, it fails with another kind of error with named arguments.

```
File "edgedb/protocol/protocol.pyx", line 998, in edgedb.protocol.protocol.SansIOProtocol.encode_args
RuntimeError: expected named arguments, got positional arguments
```

1st1: That RuntimeError should be changed to `errors.InterfaceError`

1st1:
Roberto, could you please rebase the PR onto most recent master?

tests/test_server_proto.py
1st1: This PR is becoming stale, let's revive it @RobertoPrevato

elprans:
Blocked on edgedb/edgedb-python#208

1st1:
@fmoor Frederick, can you take a look at this PR and see if we should close it or if we should rebase it?

fmoor:
> see if we should close it or if we should rebase it?

I didn't find any equivalent tests in master.

fmoor:
@1st1 all the clients handle this check before sending the wrong number of arguments to the server. Maybe we don't need tests here?

PR #1428, opened 2020-05-27
Adds built-in function to convert unix timestamp to datetime

### LINKED ISSUE #1183: No `to_datetime` variant for Unix timestamp conversion.
`to_datetime(epochseconds: float64) -> datetime` is missing.

RobertoPrevato:
* Internally it uses the PostgreSQL function `to_timestamp`.
* Adds tests
* Adds description in documentation together with other datetime converters

Should close #1183

### DIFF
--- a/docs/edgeql/funcops/datetime.rst
+++ b/docs/edgeql/funcops/datetime.rst
@@ -386,6 +386,7 @@ Date and Time
                   std::to_datetime(year: int64, month: int64, day: int64, \
                     hour: int64, min: int64, sec: float64, timezone: str) \
                     -> datetime
+                  std::to_datetime(epochseconds: float64) -> datetime
 
     :index: parse datetime
 
@@ -416,7 +417,7 @@ Date and Time
         ...   <cal::local_datetime>'January 1, 2019 12:00AM', 'HKT');
         {<datetime>'2018-12-31T16:00:00+00:00'}
 
-    Yet another way to construct a the :eql:type:`datetime` value
+    Another way to construct a the :eql:type:`datetime` value
     is to specify it in terms of its component parts: *year*, *month*,
     *day*, *hour*, *min*, *sec*, and *timezone*
 
@@ -426,6 +427,13 @@ Date and Time
         ...     2018, 5, 7, 15, 1, 22.306916, 'UTC');
         {<datetime>'2018-05-07T15:01:22.306916+00:00'}
 
+    Finally, it is also possible to convert a Unix timestamp to a
+    :eql:type:`datetime`
+
+    .. code-block:: edgeql-repl
+
+        db> SELECT to_datetime(1590595184.584);
+        {<datetime>'2020-05-27T15:59:44.584000000Z'}
 
 ------------
 
--- a/docs/edgeql/funcops/datetime.rst
+++ b/docs/edgeql/funcops/datetime.rst
@@ -386,7 +386,9 @@ Date and Time
                   std::to_datetime(year: int64, month: int64, day: int64, \
                     hour: int64, min: int64, sec: float64, timezone: str) \
                     -> datetime
+                  std::to_datetime(epochseconds: decimal) -> datetime
                   std::to_datetime(epochseconds: float64) -> datetime
+                  std::to_datetime(epochseconds: int64) -> datetime
 
     :index: parse datetime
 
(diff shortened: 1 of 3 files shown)


docs/edgeql/funcops/datetime.rst
```diff
@@ -416,7 +417,7 @@ Date and Time
         ...   <cal::local_datetime>'January 1, 2019 12:00AM', 'HKT');
         {<datetime>'2018-12-31T16:00:00+00:00'}
 
-    Yet another way to construct a the :eql:type:`datetime` value
+    Another way to construct a the :eql:type:`datetime` value
```
RobertoPrevato, line 420: I removed "Yet" here, because it looked a bit cumbersome when adding a new option below.

edb/lib/std/70-converters.edgeql
```diff
@@ -317,6 +317,16 @@ std::to_datetime(year: std::int64, month: std::int64, day: std::int64,
 };
 
 
+CREATE FUNCTION
+std::to_datetime(epochseconds: std::float64) -> std::datetime
```
1st1, line 321: I think we also want to add a decimal variant of the function (Postgres seems to support that):

`std::to_datetime(epochseconds: std::decimal) -> std::datetime`

docs/edgeql/funcops/datetime.rst
1st1: +1

edb/lib/std/70-converters.edgeql
RobertoPrevato: Something interesting is happening.
I added a function to handle decimal:

```
CREATE FUNCTION
std::to_datetime(epochseconds: std::decimal) -> std::datetime
```

It works from the edgedb CLI:

```
SELECT <str>to_datetime(517795200.00n);
{'1986-05-30T00:00:00+00:00'}
```

But it fails when I run tests in Python:
```
edgedb.errors.QueryError: function to_datetime is not unique
```

It looks like it works in the CLI implemented with Rust, but fails with the Python code. I'll need to study how the Python code handles these overloads and if it's using Decimal class for decimal.

1st1: Hm, Rust or no Rust doesn't make any difference as both interact with the DB via our binary protocol. Can you also copy/paste your Python test code here?

1st1: And please push the latest code to this PR.

elprans: I think this is due to implicit cast from int to float and decimal.  Adding an explicit int64 variant should help.

RobertoPrevato: Pushed: tests won't pass, but if you try with the CLI, you should see the result I pasted above.

RobertoPrevato: The Python test code is:

```py
    async def test_edgeql_functions_unix_to_datetime_03(self):
        dt = await self.con.fetchone(
            'SELECT <str>to_datetime(517795200);'
        )
        self.assertEqual('1986-05-30T00:00:00+00:00', dt)

    async def test_edgeql_functions_unix_to_datetime_04(self):
        dt = await self.con.fetchone(
            'SELECT <str>to_datetime(517795200.00n);'
        )
        self.assertEqual('1986-05-30T00:00:00+00:00', dt)
```

RobertoPrevato: Checking with the explicit `int64` version.

1st1: > I think this is due to implicit cast from int to float and decimal. Adding an explicit int64 variant should help.

Yeah, but how do you explain repl vs unittest difference in behavior?

elprans: `SELECT <str>to_datetime(517795200);` should fail in repl also.  Roberto seems to have been testing on a bigint literal, which would not trigger a lookup ambiguity.

1st1: > SELECT <str>to_datetime(517795200); should fail in repl also. Roberto seems to have been testing on a bigint literal, which would not trigger a lookup ambiguity.

Good catch :)

1st1:
Thanks, Roberto!

PR #1431, opened 2020-05-28
Support bootstrap cache in production mode

elprans:
Bootstrap cache makes complete sense in production builds also, so
enable it in `--no-devmode`.

### DIFF
--- a/edb/server/bootstrap.py
+++ b/edb/server/bootstrap.py
@@ -22,6 +22,7 @@ from typing import *
 
 import json
 import logging
+import os
 import pathlib
 import pickle
 import re
@@ -338,7 +339,7 @@ async def _make_stdlib(testmode: bool, global_ids) -> StdlibBits:
     schema = s_schema.Schema()
     schema, _ = s_mod.Module.create_in_schema(schema, name='__derived__')
 
-    current_block = None
+    current_block = dbops.PLTopBlock()
 
     std_texts = []
     for modname in s_schema.STD_LIB + ('stdgraphql',):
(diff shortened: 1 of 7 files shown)


tests/test_type_coverage.py
```diff
@@ -199,7 +199,7 @@ def test_cqa_type_coverage_cli(self) -> None:
         self.assertFunctionCoverage(EDB_DIR / "cli", 38.71)
 
     def test_cqa_type_coverage_common(self) -> None:
-        self.assertFunctionCoverage(EDB_DIR / "common", 36.96)
+        self.assertFunctionCoverage(EDB_DIR / "common", 37.05)
```
1st1, line 202: This thing is dead now.

edb/server/bootstrap.py
```diff
@@ -463,37 +446,93 @@ async def _make_stdlib(testmode: bool, global_ids) -> StdlibBits:
...
+    schema = stdlib.stdschema
+    reflschema = stdlib.reflschema
+
+    topblock = dbops.PLTopBlock()
+    plans = []
+
+    context = sd.CommandContext()
+    context.stdmode = True
+
+    for ddl_cmd in edgeql.parse_block(ddl_text):
```
1st1, line 462: I assume this `for` is copy/pasted from `compiler.py`, no? If so I'd factor it out into a helper function in server/compiler.py

elprans: No, it's not copy/pasted from `compiler.py`, there are important differences from normal processing of DDL.

Issue #1218, opened 2020-02-12
The `\d` command is a bit broken


vpetrovykh:
Currently you can do this:
```
db> \d all
function std::all(vals: SET OF std::bool) ->  std::bool {
    volatility := 'IMMUTABLE';
    using sql
;};


db> CREATE FUNCTION all() -> bool USING EdgeQL 'SELECT True';
CREATE
db> \d all
function default::all() ->  std::bool using (SELECT True);


db> CREATE TYPE all;
CREATE
db> \d all
function default::all() ->  std::bool using (SELECT True);


db> \d default::all
function default::all() ->  std::bool using (SELECT True);
```

The function `default::all` successfully masks the type `default::all` (same happens if you had an abstract link `default::all`). I think that other cases of masking are impossible as we require names for types, links, props, and constraints to be universally unique, but only functions are exempt from that. We need some way to handle this, maybe some optional flags for the `\d` command that make it target specific schema types.

1st1:
Yeah, we should mirror `\l*` commands to `\d*`, e.g. `\df`, `\dt`, etc.

When there's a name collision for `\d` we should probably list all descriptions.

1st1:
@vpetrovykh Please verify if there's anything left to do here.

vpetrovykh:
This is definitely still an issue as of `EdgeDB 1.0-alpha.3+dev.214.gd0f51837`.

elprans:
I believe this has been fixed in #1432

PR #1432, opened 2020-05-28
Make DESCRIBE list multiple matches.

vpetrovykh:
Make DESCRIBE command list multiple matches for the specific name if
available. This can arise when there's a type and a function by the
same name or when a user-created type or function masks a std type or
function.

Issue #1218.

### DIFF
--- a/edb/edgeql/compiler/stmt.py
+++ b/edb/edgeql/compiler/stmt.py
@@ -23,7 +23,9 @@
 from __future__ import annotations
 
 from typing import *
+import textwrap
 
+from collections import defaultdict
 from edb import errors
 
 from edb.ir import ast as irast
@@ -492,7 +494,7 @@ def compile_DescribeStmt(
                     f'cannot describe full schema as {ql.language}')
         else:
             modules = []
-            items: List[str] = []
+            items: DefaultDict[str, List[str]] = defaultdict(list)
             referenced_classes: List[s_obj.ObjectMeta] = []
 
             objref = ql.object
@@ -502,7 +504,6 @@ def compile_DescribeStmt(
                 modules.append(objref.name)
             else:
                 itemtype: Optional[Type[s_obj.Object]] = None
-                found = False
 
                 name: str
                 if objref.module:
@@ -520,40 +521,80 @@ def compile_DescribeStmt(
                                 itemclass)
                         )
 
-                if (itemclass is None or
-                        itemclass is qltypes.SchemaObjectClass.FUNCTION):
+                last_exc = None
+                # Search in the current namespace AND in std. We do
+                # this to avoid masking a `std` object/function by one
+                # in a default module.
+                search_ns = [ictx.modaliases]
+                # Only check 'std' separately if the current
+                # modaliases don't already include it.
+                if ictx.modaliases.get(None, 'std') != 'std':
+                    search_ns.append({None: 'std'})
 
-                    try:
-                        funcs: Tuple[s_func.Function, ...] = (
-                            ictx.env.schema.get_functions(
-                                name,
-                                module_aliases=ictx.modaliases)
-                        )
-                    except errors.InvalidReferenceError:
-                        pass
-                    else:
-                        for func in funcs:
-                            items.append(func.get_name(ictx.env.schema))
-                        found = True
+                # Search in the current namespace AND in std.
+                for aliases in search_ns:
+                    # Use the specific modaliases instead of the
+                    # context ones.
+                    with ictx.subquery() as newctx:
+                        newctx.modaliases = aliases
+                        # Get the default module name
+                        modname = aliases[None]
+                        # Is the current item a function
+                        is_function = (itemclass is
+                                       qltypes.SchemaObjectClass.FUNCTION)
 
-                if not found:
-                    if itemclass is not qltypes.SchemaObjectClass.ALIAS:
-                        condition = None
-                        label = None
-                    else:
-                        condition = (
-                            lambda obj:
-                            obj.get_alias_is_persistent(ctx.env.schema)
-                        )
-                        label = 'alias'
-                    obj = schemactx.get_schema_object(
-                        objref,
-                        item_type=itemtype,
-                        condition=condition,
-                        label=label,
-                        ctx=ictx,
-                    )
-                    items.append(obj.get_name(ictx.env.schema))
+                        # We need to check functions if we're looking for them
+                        # specifically or if this is a broad search. They are
+                        # handled separately because they allow multiple
+                        # matches for the same name.
+                        if (itemclass is None or is_function):
+                            try:
+                                funcs: Tuple[s_func.Function, ...] = (
+                                    newctx.env.schema.get_functions(
+                                        name,
+                                        module_aliases=aliases)
+                                )
+                            except errors.InvalidReferenceError:
+                                pass
+                            else:
+                                for func in funcs:
+                                    items[f'function_{modname}'].append(
+                                        func.get_name(newctx.env.schema))
+
+                        # Also find an object matching the name as long as
+                        # it's not a function we're looking for specifically.
+                        if not is_function:
+                            try:
+                                if itemclass is not \
+                                        qltypes.SchemaObjectClass.ALIAS:
+                                    condition = None
+                                    label = None
+                                else:
+                                    condition = (
+                                        lambda obj:
+                                        obj.get_alias_is_persistent(
+                                            ctx.env.schema
+                                        )
+                                    )
+                                    label = 'alias'
+                                obj = schemactx.get_schema_object(
+                                    objref,
+                                    item_type=itemtype,
+                                    condition=condition,
+                                    label=label,
+                                    ctx=newctx,
+                                )
+                                items[f'other_{modname}'].append(
+                                    obj.get_name(newctx.env.schema))
+                            except errors.InvalidReferenceError as exc:
+                                # Record the exception to be possibly
+                                # raised if no matches are found
+                                last_exc = exc
+
+            # If we already have some results, suppress the exception,
+            # otherwise raise the recorded exception.
+            if not items and last_exc:
+                raise last_exc
 
             verbose = ql.options.get_flag('VERBOSE')
 
@@ -570,14 +611,48 @@ def compile_DescribeStmt(
                     f'cannot handle describe language {ql.language}'
                 )
 
+            # Based on the items found generate main text and a
+            # potential comment about masked items.
+            defmod = ictx.modaliases.get(None, 'std')
+            default_items = []
+            masked_items = set()
+            for objtype in ['function', 'other']:
+                defkey = f'{objtype}_{defmod}'
+                mskkey = f'{objtype}_std'
+
+                default_items += items.get(defkey, [])
+                if defkey in items and mskkey in items:
+                    # We have a match in default module and some masked.
+                    masked_items.update(items.get(mskkey, []))
+                else:
+                    default_items += items.get(mskkey, [])
+
+            # Throw out anything in the masked set that's already in
+            # the default.
+            masked_items.difference_update(default_items)
+
             text = method(
                 ctx.env.schema,
                 included_modules=modules,
-                included_items=items,
+                included_items=default_items,
                 included_ref_classes=referenced_classes,
                 include_module_ddl=False,
                 include_std_ddl=True,
             )
+            if masked_items:
+                text += ('\n\n'
+                         '# The following builtins are masked by the above:'
+                         '\n\n')
+                masked = method(
+                    ctx.env.schema,
+                    included_modules=modules,
+                    included_items=masked_items,
+                    included_ref_classes=referenced_classes,
+                    include_module_ddl=False,
+                    include_std_ddl=True,
+                )
+                masked = textwrap.indent(masked, '# ')
+                text += masked
 
         ct = typegen.type_to_typeref(
             ctx.env.get_track_schema_type('std::str'),
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -4823,3 +4823,61 @@ class TestDescribe(tb.BaseSchemaLoadTest):
             r"function test::foo() ->  std::str "
             r"using (SELECT r'\1');"
         )
+
+    def test_describe_poly_01(self):
+        self._assert_describe(
+            """
+            scalar type all extending str;
+
+            type Object {
+                property real -> bool;
+            }
+
+            function all() -> bool {
+                using (
+                    SELECT true
+                );
+            }
+            """,
+
+            'DESCRIBE OBJECT all AS TEXT',
+
+            """
+            function test::all() -> std::bool using (SELECT
+                true
+            );
+            scalar type test::all extending std::str;
+
+            # The following builtins are masked by the above:
+
+            # function std::all(vals: SET OF std::bool) ->  std::bool {
+            #     volatility := 'IMMUTABLE';
+            #     using sql
+            # ;};
+            """,
+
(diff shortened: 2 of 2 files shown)


vpetrovykh:
This only addresses part of the issue #1218. `DESCRIBE` and `\d` will now display multiple matches. There's still no specialized version of `\d`.

edb/edgeql/compiler/stmt.py
```diff
@@ -512,30 +511,50 @@ def compile_DescribeStmt(
...
-                        ctx=ictx,
-                    )
-
-                    items.append(obj.get_name(ictx.env.schema))
+                last_exc = None
+                # Search in the current namespace AND in std.
+                for aliases in [ictx.modaliases, {}]:
+                    # Use the specific modaliases instead of the
+                    # context ones.
+                    ictx.modaliases = cast(Dict[Optional[str], str], aliases)
```
1st1, line 519: So you've overridden `ictx.modaliases` to `{}` in this call, so the following operations can potentially be broken. Instead of this monkey patching I'd add a `modaliases` param to `get_schema_object` and pass `aliases` explicitly (and `ictx.env.schema.get_functions` already has a `module_aliases` arg for that exact purpose).

elprans: Why is this hack necessary at all?  Unqualified names are always searched in `std` regardless of `modaliases`.

vpetrovykh: > So you've overridden `ictx.modaliases` to `{}` in this call, so the following operations can potentially be broken. 

What operations would be broken? This is a `DESCRIBE`, so we have no operation or expression there, just a name that we need to match. The `modaliases` just specify which is the default module.

> Instead of this monkey patching I'd add a `modaliases` param to `get_schema_object` and pass `aliases` explicitly (and `ictx.env.schema.get_functions` already has a `module_aliases` arg for that exact purpose).

I'm patching the `ictx` because it is used in both the `get_functions` and `get_schema_object` and the latter takes the whole context from which it extracts `modaliases`.

Now that I look at it closer I think I can use `module_aliases` and `module` parameters of those two calls, respectively. This way I can avoid patching the context.

vpetrovykh: > Why is this hack necessary at all? Unqualified names are always searched in `std` regardless of `modaliases`.

The motivation for the hack is to avoid a user-created object/function in default module (from `ictx.modalises`) to mask a `std` built-in with the same name.

Look at the test: it has a user-defined function `all` and a user-defined scalar `all`. If I don't check `std` separately the user-defined function would mask the `std::all` which is not correct for `DESCRIBE all`. Same would happen if I defined a `default::Object`, it would mask the `std::Object` (I'll add that to the test as well). This happens because if the object is found in a default module, the search doesn't proceed to `std`.

elprans: Why is masking suddenly an issue in the case of `DESCRIBE`?  Yes, you masked a `std` object, and `DESCRIBE` should abide by the common name resolution rules and not invent its own thing. If you want do describe a masked schema object, qualify it

1st1: > What operations would be broken? This is a DESCRIBE, so we have no operation or expression there, just a name that we need to match. The modaliases just specify which is the default module.

IDK, the code can be refactored later. Violating separation principles by monkey-patching things isn't best practice, period.

> Why is masking suddenly an issue in the case of DESCRIBE? Yes, you masked a std object, and DESCRIBE should abide by the common name resolution rules and not invent its own thing. 

+1.

vpetrovykh: I type `\d all` and the standard library function is missing, because I happen to be in a module that defines it's own `all`. This is bad because the signatures of the `std::all` and the `default::all` could be completely incompatible and thus in an actual query never confused, acting like polymorphic versions.

The argument for harm of masking types is not so straightforward, but still, it's oddly misleading that just because I happen to be in a particular module (via `SET MODULE foo`) I should suddenly not see built-ins.

1st1: > This is bad because the signatures of the std::all and the default::all could be completely incompatible and thus in an actual query never confused, acting like polymorphic versions.

It's equally confusing to show the std version, as if you type a query with `all()` in the same repl window the std version wouldn't be called anyways. The `\d` command must show what you would actually execute.

That said, `\d` can potentially show other things that are shadowed in a comment, like 

```
db> \d all
function test::all(blah) {
  ...
}

# The test:all() function shadows std::all() function because
# the "test" module is set as default module for this session.
```

vpetrovykh: > If you want do describe a masked schema object, qualify it

How would I know it's masked? This is one way to discover this by using `\d`/`DESCRIBE`. Presumably I use `\d`/`DESCRIBE` precisely because I don't have a perfect recollection about the thing I'm asking about. It's more informative to show builtins even if they are masked, perhaps especially if they are masked to make this fact more obvious.

vpetrovykh: With functions it's weird to not show the polymorphic matches if they happen to be in `std`.

Consider this:
```
CREATE FUNCTION all(a: int64) -> bool USING (a = count(Object));
```
This just makes `all` a polymorphic function, no? I should expect both versions to work... which currently they don't:
```
db0> SELECT all(3);
{false}
db0> SELECT all(true);
error: could not find a function variant all
  ┌─ query:1:8
  │
1 │ SELECT all(true);
  │        ^^^^^^^^^^ error

db0> SELECT std::all(true);
{true}
```
This actually seems like a bug, not a feature. Perhaps I need to update the original ticket to include this issue.

1st1: > This just makes all a polymorphic function, no?

No, we don't allow cross-module polymorphism.

vpetrovykh: > No, we don't allow cross-module polymorphism.

Is there a reason as to why not? I mean, is it a hard limitation? If it's not a hard limitation, then it's a bit of a foot gun - I add a utility function to my default module, thinking I'll extend, say, `all` to my own [object] type or whatnot. Suddenly my other queries using the built-in `all` on booleans stop working. In shapes, computables, or just in my EdgeQL code. Even though there's technically no ambiguity w.r.t. the function signature.

It's very counter-intuitive that defining this, will break perfectly reasonable queries that use the built-in `len`:
```
module default {
    type Foo {
        multi property vals -> str;
        ...
    }
    function len(a: Foo) -> int64 using (SELECT sum(len(a.vals)));
}
```

1st1: > Is there a reason as to why not? 

Because it would be a bit crazy and hard to reason about.

> I add a utility function to my default module, thinking I'll extend, say, all to my own [object] type or whatnot.

Sure we can come up with examples rationalizing this. But it would still be a confusing feature in many contexts.

> It's very counter-intuitive that defining this, will break perfectly reasonable queries that use the built-in len:

It would be very counter-intuitive to me (no kidding) if it worked. If you define a `len()` function in your Python script it's not surprising it completely shadows the built-in. Or take C++ that supports function overloading -- it would be weird if defining a "foo()" function in one namespace means the compiler would now "try" "foo()" from other namespaces too.

IMO enabling this kind of cross-module polymorphism is unprecedented and would be confusing to many users including myself.

vpetrovykh: But this isn't any more "cross-module" than the usual "std is always accessible". Why defining a function with a definitely *distict* signature makes `std` unreachable?

`len` and `all` might be odd cases, but this makes it really hard defining a `to_str` for my own scalar or object type without screwing myself over big-time.

My reasoning is: I called `len('as')` so I'm looking for a `default::len(str)` function, but I find `default::len(Foo)`, which is obviously not a match, soooo I keep looking in `std`, when match not found. I mean, they are completely different functions with distinct signatures, why would they clash?

Python is a bad example, because being duck-typed it handles polymorphism differently, by asking you to hide it inside the function body using `isinstance` and such. On the other hand strictly-typed languages are usually quite happy to distinguish signatures based on parameter types (and order). Basically, it's like whatever namespace you're in, you also import all of `std` directly into it and *then* define more things.

1st1: > But this isn't any more "cross-module" than the usual "std is always accessible". Why defining a function with a definitely distict signature makes std unreachable?

Because our current reference resolution is based on names, and not other things like signatures. You're proposing to implement a new resolution algorithm to be able to magically overload the few functions we have in `std::`. I'm standing by what I said earlier -- I find the proposed behavior confusing and the motivation to implement it weak.

> len and all might be odd cases, but this makes it really hard defining a to_str for my own scalar or object type without screwing myself over big-time.

Let's say you have more modules than just "default" -- because why would we add a feature that only works when there's one module. In this case let's say your scalar is defined in module "a":

```
module a {
   type myscalar {...}
   function to_str(myscalar) ...
}
```

and now module "b":

```
module b {
   type MyType {
      property foo -> myscalar;
      property foo_name := to_str(.foo)
   }
}
```

Thing is `foo_name` would call `std::to_str`. You can fix it to call `a::to_str()` but that would go against your idea of magically overloading `std::to_str()`.

You can scale your example to queries. Essentially, the idea of overloading std module functions in your custom schema would only work reliably in trivial schemas with one module.  Or we need to add some other mechanisms to be able to achieve your goal.

Ultimately, there's no chance we would implement such an elaborate departure from the current semantics before 1.0.

1st1: IMO the only way of implementing what you want is to allow to directly overload `std::to_str(MyType)` or something like that. This can be a backwards compatible post 1.0 feature.

vpetrovykh: C++ is closer to our case, but it's not the same. There you can use `using namespace foo` or `using foo::bar`, but there's no implicit `using namespace std`. However, having said that if you import 2 namespaces with functions of identical name and distinct signatures, C++ is happy enough to resolve them.

http://cpp.sh/3sge2

Basically, to the extent that C++ allows function polymorphism it's happy enough to resolve them even if you import them by 2 consecutive `using namespace ...` directives. Well, at least "a C++" allows it, I'm quite a bit rusty in it at this point so I'm not sure how different versions handle it.

So to complete this analogy, I'd expect that EdgeQL works similar to having: 
```c++
using namespace std;
using namespace mymodule; // or whatever is the "current" module
...
```
Although, this analogy is not perfect in case of an exact name/signature match, which should result in `mymodule` masking `std`.

P.S. None of this changes how explicit module names get resolved, obviously those things can only be looked for in that specific module, as the name suggests. This is only for unqualified names.
P.P.S. Our own algorithm obviously has the capacity to find the functions in the "default" module (whatever that may be) or in `std` if none are found. I'm guessing that it's largely a matter of removing an `if` before looking into `std` and also validating that the stuff in `std` has a different signature... I assume we handle those match candidates based on signature already, so that doesn't sound like overly complex code change. Is there perhaps some additional issue in underlying SQL? Polymorphic resolution from inside function body, etc.?

1st1: > Basically, to the extent that C++ allows function polymorphism it's happy enough to resolve them even if you import them by 2 consecutive using namespace ... directives. 

But it does that when you **explicitly** "use" namespaces, not implicitly.  I.e. we don't have syntax like

```
WITH MODULE a, MODULE b
SELECT to_str(...)
```

Exhibit (A).

> P.S. None of this changes how explicit module names get resolved, obviously those things can only be looked for in that specific module, as the name suggests. This is only for unqualified names.

That was clear.

> I'm guessing that it's largely a matter of removing an if before looking into std and also validating that the stuff in std has a different signature...

Please stop twisting the topic. We are not discussing a removal of an `if`. We're discussing a semantic change to how functions are resolved. Without an explicit syntax like in (A) you're basically proposing to resolve an unqualified function name by looking it up in all modules in the DB until it's found. IMO this is completely unacceptable, regardless of how simple on the surface it is to implement.

Anyways, I don't think you're listening to my arguments so let me just put my -1 here. 

Please address the comments on this PR so that it works correctly with the current semantics of EdgeQL.

vpetrovykh: I'll rewrite this PR to use comments as a way to notify the user about types and functions being masked. After all, `DESCRIBE` and `\d` are supposed to provide insight into the current schema and masking built-ins is useful insight.

Regarding the behavior of unqualified name resolution, I'll open a separate issue where we can give it some proper consideration.

tests/test_schema.py
```diff
@@ -4817,3 +4817,61 @@ def test_describe_escape(self):
...
+            }
+            """,
+
+            'DESCRIBE OBJECT all AS TEXT',
+
+            """
+            function test::all() -> std::bool using (SELECT
+                true
+            );
+            scalar type test::all extending std::str;
```
1st1, line 4843: @elprans Do we intentionally support functions and types having the same names?

1st1: IIRC we did at some point of time to allow converter functions share the same name as types they convert to, but then we adopted the `to_<typename>` naming and planned to remove this "feature".

vpetrovykh: This may be worthwhile to discuss alongside the #1439 since it also touches on names uniqueness and how are functions different (or not) from other entities. Functions are somewhat special, but are they special enough is a good question.

elprans: We don't "intentionally" allow this, there's just no name clash, because all function names are mangled.

1st1: Well, I explained my -1 in #1439 and I think that name clashes should be prohibited.

vpetrovykh: My understanding was that at the moment we make `\d` actually describe the status quo, as opposed to change behavior. At the moment the functions don't name clash.

1st1: > At the moment the functions don't name clash.

Hm, looks like they do... you have `test::all` type and `test::all()` function, no?

But you're right, this PR should work withing the semantics implemented in the master branch. It looks OK to me, but I'll let @elprans to review it. The `for aliases in search_ns:` still looks questionable to me.

tests/test_schema.py
```diff
@@ -4817,3 +4817,61 @@ def test_describe_escape(self):
...
+
+            # The following builtins are masked by the above:
+
+            # function std::all(vals: SET OF std::bool) ->  std::bool {
+            #     volatility := 'IMMUTABLE';
+            #     using sql
+            # ;};
```
1st1, line 4850: I don't think this is needed. IMO `# The std::all builtin is masked by the above:` would suffice. Interested users can always describe the fully qualified name and that description would be properly syntax highlighted.

tests/test_schema.py
vpetrovykh: When I said "they don't name clash" I meant that function and object by same name doesn't cause an error at the moment. But yeah, I am specifically testing cases where multiple entities have the same short name in the test, including a function and a a scalar.

tests/test_schema.py
vpetrovykh: I was thinking about it. My reasoning is that masking isn't common and it's nicer if you get full info in a single command invocation rather than having to then look up separately. Basically this won't affect 99% schemas and will make the rare schemas that mask things a little easier to navigate and reason about.

1st1: I'm fine to keep it the way you implemented. It's a bit too noisy for my taste, but I hear your reasoning too.

1st1: @elprans thoughts?

1st1:
Perhaps we want to address https://github.com/edgedb/edgedb/issues/1465 first.

vpetrovykh:
I've rebased the PR. The status quo in name resolution remains, so the potentially masked "std" functions or types are shown in a comment.

PR #1433, opened 2020-05-28
Enable the ALTER FUNCTION command; strengthen the dump02 test.

1st1:


haikyuu:
Is it possible to alter the body of the function?

I always get this error while iterating on a function: `a function with the same signature is already defined` and i need to drop it first in order to change the body

edb/edgeql/parser/grammar/ddl.py
```diff
@@ -1638,6 +1641,33 @@ def reduce_DropFunction(self, *kids):
...
+    'AlterFunction',
+    SetFieldStmt,
+    CreateAnnotationValueStmt,
+    AlterAnnotationValueStmt,
+    DropAnnotationValueStmt,
+    opt=False
+)
+
+
+class AlterFunctionStmt(Nonterm):
```
elprans, line 1658: You also need to reflect changes to volatility in Postgres by adding code to `pgsql.delta.AlterFunction`.  The check would be `if self.get_attribute_value('volatility'): <code-as-in-CreateFunction>` (with `dbops.CreateFunction(func, or_replace=True)`).  Incidentally, you should be able to support body changes this way as well.

1st1: Good point. I'll see if I can add some testmode stuff to introspect Postgres volatility and test this thoroughly.

1st1:
> Is it possible to alter the body of the function?

I'll take a look, sure!

1st1:
>  Is it possible to alter the body of the function?

Implemented and will be part of alpha 3.

PR #1440, opened 2020-05-30
Add more unicode features to the dump test.

vpetrovykh:
Test non-trivial unicode names for:
- annotation
- scalar type
- type
- property
- link
- link property
- index (on a property with unicode name)
- function (including parameters and type)
- module
- constraint
- database

### DIFF
--- a/tests/schemas/dump02_default.esdl
+++ b/tests/schemas/dump02_default.esdl
@@ -18,12 +18,14 @@
 
 abstract annotation `🍿`;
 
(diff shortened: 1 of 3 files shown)


vpetrovykh:
I came across a couple of issues with constraints and link props, but they don't seem to be in the dump, but rather in actual DDL processing. I'll file separate bugs.

1st1:
Thanks Victor!

PR #1443, opened 2020-05-30
Use lxml 4.5.1

ambv:
The 4.4 series bundled an outdated libxml2 and the full build was failing for
me on Python 3.8 running on macOS Catalina due to the emoji in
internals/protocol/dataformats.rst.

### DIFF
--- a/edb/tools/docs/eql.py
+++ b/edb/tools/docs/eql.py
@@ -1045,8 +1045,9 @@ class StatementTransform(s_transforms.SphinxTransform):
 
     def apply(self):
         for section in self.document.traverse(d_nodes.section):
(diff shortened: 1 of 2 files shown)


ambv:
The error was fun as well, try to find what's going on based on this wonderful error reporting:
```
Exception occurred:
  File "src/lxml/etree.pyx", line 1871, in lxml.etree._ElementTree._assertHasRoot
AssertionError: ElementTree not initialized, missing root
```

1st1:
> The error was fun as well, try to find what's going on based on this wonderful error reporting:

Yeah, not fun indeed.

PR #1445, opened 2020-05-30
Add params validation logic to abstract constrains

1st1:


edb/schema/constraints.py
```diff
@@ -621,7 +621,108 @@ def _create_begin(
     ) -> s_schema.Schema:
         referrer_ctx = self.get_referrer_context(context)
         if referrer_ctx is None:
-            return super()._create_begin(schema, context)
+            schema = super()._create_begin(schema, context)
+
+            # Must be set after _create_begin
```
elprans, line 626: Put this behind `if not context.canonical` to avoid re-running checks on fully-validated deltas, and possibly factor into a helper method.

1st1: done

PR #1447, opened 2020-05-31
Require explicit OPTIONAL qualifiers for properties and links

1st1:
* Mass transform unqualified properties and links in tests

* Include OPTIONAL in generated DDL

* allow for compatible dump/restore

Co-authored-by: Yury Selivanov <yury@edgedb.com>


cthoyt:
FWIW I think making this explicit is good

1st1:
Closing this; read https://github.com/edgedb/rfcs/blob/master/text/1002-optionality-qualifier.rst for more details.

> FWIW I think making this explicit is good

Thank you for sharing this! In general I agree, explicit is usually better than implicit. In this case, however, after we did the change we saw that ESDL files became extremely verbose. It seems that in this specific case we can get away with having OPTIONAL implicit. Please also take a look at the linked RFC!

PR #1450, opened 2020-05-31
Fix `IndexCommand.get_object()`

ambv:
Without this change, altering a type to add an index to it results in:

```
TypeError: get_object() got an unexpected keyword argument 'default'
```

### DIFF
--- a/edb/schema/indexes.py
+++ b/edb/schema/indexes.py
@@ -185,14 +185,35 @@ class IndexCommand(
         quals = sn.quals_from_fullname(name)
         return tuple(quals[-2:])
 
-    # type ignore below, because parent class defines a compatible overload
-    def get_object(  # type: ignore
+    @overload
+    def get_object(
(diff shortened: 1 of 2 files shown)


ambv:
Sigh. Mypy failed here with:
```
edb/schema/indexes.py:188:5: error: Signature of "get_object" incompatible with supertype "QualifiedObjectCommand"
edb/schema/indexes.py:188:5: error: Signature of "get_object" incompatible with supertype "ObjectCommand"
edb/schema/indexes.py:198:53: error: Argument "default" to "get_object" of "QualifiedObjectCommand" has incompatible type "Union[so.Object_T, NoDefaultT, None]"; expected "Union[Index, NoDefaultT]"
Found 3 errors in 1 file (checked 307 source files)
```

but in fact `get_object()` **does have** the necessary overloads to satisfy the type here. Investigating.

Issue #1457, opened 2020-06-01
Creating type named `str` in default confuses edgedb


tailhook:
- EdgeDB Version: 1.0-a3

Steps to reproduce:
```
db> CREATE TYPE str;
OK: CREATE
db> SELECT "x";
ERROR: InternalServerError: insufficient data left in message
```
Compiled query assumes it's uuid for some reason:
```
                                                                                                                     
        (SELECT/*<pg.SelectStmt at 0x7fe5da5545e0>*/
                    ($1)::uuid AS "expr~1_identity~1"     
            ) AS "expr-1~2"                                                                
            INNER JOIN "edgedb_8e7140a2-a409-11ea-8bbb-dbade5f70ff0"."e779608c-a409-11ea-9ca9-e5640d314547" AS "str~1"
                ON ("expr-1~2"."expr~1_identity~1" = "str~1".id)
```
Explicit variables crash like this:
```
db> SELECT <str>$0;
ERROR: InternalServerError: cannot describe type default::str
  Hint: This is most likely a bug in EdgeDB. Please consider opening an issue ticket at https://github.com/edgedb/edgedb/issues/new?template=bug_report.md
  Server traceback:
      Traceback (most recent call last):
        File "/work/edb/server/procpool/worker.py", line 75, in worker
          res = await meth(*args)
        File "/work/edb/server/compiler/compiler.py", line 1480, in compile_eql_tokens
          return self._compile(ctx=ctx, tokens=eql_tokens)
        File "/work/edb/server/compiler/compiler.py", line 1141, in _compile
          comp: dbstate.BaseQuery = self._compile_dispatch_ql(ctx, stmt)
        File "/work/edb/server/compiler/compiler.py", line 1104, in _compile_dispatch_ql
          return self._compile_ql_query(ctx, ql)
        File "/work/edb/server/compiler/compiler.py", line 637, in _compile_ql_query
          in_type_data, in_type_id = sertypes.TypeSerializer.describe(
        File "/work/edb/server/compiler/sertypes.py", line 335, in describe
          type_id = builder._describe_type(
        File "/work/edb/server/compiler/sertypes.py", line 133, in _describe_type
          subtypes = [self._describe_type(st, view_shapes,
        File "/work/edb/server/compiler/sertypes.py", line 133, in <listcomp>
          subtypes = [self._describe_type(st, view_shapes,
        File "/work/edb/server/compiler/sertypes.py", line 328, in _describe_type
          raise errors.InternalServerError(
[...]

PR #1458, opened 2020-06-01
Emit fully qualified parameter types in constant extraction

1st1:
Addresses #1457.


1st1:
While the original issue is relatively easy to address we have another one: if a user aliases 'std' to some other module things break, e.g.:

```
WITH std AS MODULE math SELECT 'hello'
```

I've couple of ideas how we can fix this:

1. Prohibit re-aliasing 'std'
2. Add `__std__` alias for 'std' -- dunder names are reserved and this is kinda similar to Python's `__builtin__`.

Thoughts?

elprans:
> Add `__std__` alias for `std` 

This would be a more reliable approach for references that are generated internally.

PR #1463, opened 2020-06-02
Add a __std__ alias for 'std'; fix constant extraction

1st1:
Add a '__std__' alias for the 'std' module.

This fixes constant extraction in queries wit a
'WITH std AS MODULE ...' clause.

Issue #1457.

### DIFF
--- a/edb/edgeql-parser/src/keywords.rs
+++ b/edb/edgeql-parser/src/keywords.rs
@@ -122,6 +122,7 @@ pub const CURRENT_RESERVED_KEYWORDS: &[&str] = &[
     "__source__",
     "__subject__",
     "__type__",
+    "__std__",
     "alter",
     "and",
     "anytuple",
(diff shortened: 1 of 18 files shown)


1st1:
> There are a bunch of hardcoded references to std:: in edb.edgeql.compiler, we must update them also.

Most of them just resolve some type in std and that's unrelated to modaliases. The only cases I found where `std::` names are actually injected into the IR are:

* serialization of EdgeQL types to IR; if std is shadowed, we should compile `int64` to `__std__::int64`

* complex array casts, where we inject `std::array_agg` and other functions.

I've fixed both issues and added tests for them.

elprans:
> Most of them just resolve some type in std and that's unrelated to modaliases. 

Modaliases affect `schema.get()`, so those references might get affected also.

1st1:
> Modaliases affect schema.get(), so those references might get affected also.

You use `get_track_schema_type` and `get_track_schema_object` helpers in most places that don't pass `ctx.modaliases` to `schema.get`

1st1:
Basically I've updated all places where `ctx.env.modaliases` are used with `std::` names.

PR #1469, opened 2020-06-05
Allow explit OPTIONAL qualifiers in DDL and SDL

### LINKED ISSUE #1342: Require 'required' or 'optional' modifier for properties and links

ambv:
Also makes DESCRIBE output OPTIONAL by default in both DDL and SDL.

Fixes #1342

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,5 @@
-.PHONY: docs cython postgres postgres-ext pygments
+.PHONY: build docs cython postgres postgres-ext pygments
+.DEFAULT_GOAL := build
 
 SPHINXOPTS:="-W -n"
 
@@ -22,5 +23,5 @@ pygments:
 		echo "$$out" > edb/edgeql/pygments/meta.py
 
 
-update:
-	pip install -Ue .[test]
+build:
+	pip install -Ue .[docs,test]
--- a/Makefile
+++ b/Makefile
@@ -20,3 +20,7 @@ postgres:
 pygments:
 	out=$$(edb gen-meta-grammars edgeql) && \
 		echo "$$out" > edb/edgeql/pygments/meta.py
+
+
+update:
+	pip install -Ue .[test]
--- a/edb/edgeql/codegen.py
+++ b/edb/edgeql/codegen.py
@@ -1225,8 +1225,14 @@ class EdgeQLSourceGenerator(codegen.SourceGenerator):
         keywords = []
         if self.sdlmode and node.declared_overloaded:
             keywords.append('OVERLOADED')
-        if node.is_required:
-            keywords.append('REQUIRED')
+            if node.is_required:
+                keywords.append('REQUIRED')
+        else:
+            if node.is_required is True:
+                keywords.append('REQUIRED')
+            elif node.is_required is False:
+                keywords.append('OPTIONAL')
+            # else: `is_required` is None
         if node.cardinality:
             keywords.append(node.cardinality.as_ptr_qual().upper())
         keywords.append('PROPERTY')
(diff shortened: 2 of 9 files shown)


ambv:
Note: there's a test failing left here: `SELECT OPTIONAL`
Do we want this syntax to work?

Makefile
```diff
@@ -20,3 +20,7 @@ postgres:
 pygments:
 	out=$$(edb gen-meta-grammars edgeql) && \
 		echo "$$out" > edb/edgeql/pygments/meta.py
+
+
+update:
```
elprans, line 25: I'd call this "build" and make the default target.

Makefile
```diff
@@ -20,3 +20,7 @@ postgres:
...
+	pip install -Ue .[test]
```
ambv, line 26: This isn't related but I added this because I was tired of forgetting to add `[test]` when I update the build.

tests/schemas/issues.esdl
```diff
@@ -68,7 +68,7 @@ scalar type issue_num_t extending std::str;
 
 type Comment extending Text, Owned {
     required link issue -> Issue;
-    link parent -> Comment;
+    optional link parent -> Comment;
```
ambv, line 71: I sprinkled a few `optional` declarations in this file as a form of testing the SDL support.

edb/schema/reflection/structure.py
```diff
@@ -418,11 +418,12 @@ def generate_structure(
             ptr = schema_objtype.getptr(schema, fn)
 
             if fn in ownfields:
+                qual = "REQUIRED" if field.required else "OPTIONAL"
                 if ptr is None:
                     schema = _run_ddl(
                         f'''
                             ALTER TYPE {rschema_name} {{
-                                CREATE {'REQUIRED' if field.required else ''}
+                                CREATE {qual}
```
elprans, line 426: There are more `CREATE LINK|PROPERTY` places in this module, please update all of them.

tests/test_edgeql_syntax.py
```diff
@@ -755,6 +755,11 @@ def test_edgeql_syntax_required_01(self):
         SELECT REQUIRED (User.groups.description);
         """
 
+    def test_edgeql_syntax_optional_01(self):
+        """
+        SELECT OPTIONAL (User.groups.description);
```
elprans, line 760: This is actually dead syntax and needs to be removed.  We decided to express this via a cast expression.

ambv: I was wondering about it hence I left it unimplemented. OK, will remove the test.

edb/schema/reflection/structure.py
ambv: Shit, that happens when you get interrupted 😖

Makefile
1st1: +1.

Makefile
1st1: I'd also add [docs].

edb/edgeql/codegen.py
```diff
@@ -1225,8 +1225,14 @@ def visit_CreateConcreteProperty(
         keywords = []
         if self.sdlmode and node.declared_overloaded:
             keywords.append('OVERLOADED')
-        if node.is_required:
-            keywords.append('REQUIRED')
+            if node.is_required:
+                keywords.append('REQUIRED')
+        else:
+            if node.is_required is True:
```
1st1, line 1231: I always have an itch when I see clauses like this even that there's nothing wrong with them per se. I'd replace with
`if node.is_required is not None: if node.is_required: ... else: ... `

ambv: The current approach is less nested ```¯\_(ツ)_/¯```

Makefile
elprans: @ambv, please update the target here.

PR #1475, opened 2020-06-15
Remove reliance on Postgres table inheritance

elprans:
EdgeDB currently relies on PostgreSQL table inheritance to support the 
generic query semantics as well as implement certain aspects of schema 
information propagation.  While it made sense at the point of initial 
implementation, nowadays the query compiler and the schema management 
implementation in EdgeDB are significantly more mature, to the point where
Postgres' inheritance semantics actually get in the way.  Another point
against using table inheritance is that it is incompatible with native
partitioning, and we want to be able to expose that at some point.

The removal of reliance on Postgres table inheritance requies emulating its
behavior in queries and in DDL.  For read queries we emulate inheritance by
generating SQL views which contain a union over the target type table and
the tables of its descendants.  Union views are not updatable, so for DML
we have to emulate by generating multiple SQL DML statements for each
concrete table.  The DDL side of things is mostly untouched, and boils down
to the removal of inhibition of the inherited object reflection.


edb/schema/links.py
```diff
@@ -347,6 +347,9 @@ def inherit_classref_dict(
         context: sd.CommandContext,
         refdict: so.RefDict,
     ) -> sd.CommandGroup:
+        if self.scls.get_computable(schema) and refdict.attr != 'pointers':
```
1st1, line 346: Can you add a comment?

edb/common/context.py
```diff
@@ -164,7 +173,7 @@ def _find_line(self, point, offset=0, *, line_offsets):
         return linestart, lineend
 
     def get_line_snippet(
-            self, point, max_length=120, *, offset=0, line_offsets):
+            self, point, max_length=1200, *, offset=0, line_offsets):
```
1st1, line 176: Is this debug?

edb/pgsql/common.py
```diff
@@ -188,17 +188,19 @@ def get_scalar_backend_name(id, module_id, catenate=True, *, aspect=None):
 def get_objtype_backend_name(id, module_id, *, catenate=True, aspect=None):
     if aspect is None:
         aspect = 'table'
-    if aspect != 'table' and not re.match(
+    if aspect not in {'table', 'inhview'} and not re.match(
```
1st1, line 191: Do you have an enum for aspects? Would help with documentation of the code too.

edb/pgsql/deltadbops.py
```diff
@@ -478,7 +495,7 @@ def create_constraint(self, constraint):
 
         self.add_command(my_alter)
 
-        if not constraint.is_natively_inherited():
+        if constraint.requires_triggers():
```
1st1, line 498: Is the comment below outdated?

edb/pgsql/deltadbops.py
```diff
@@ -211,6 +226,15 @@ def is_natively_inherited(self):
         """Determine if this constraint can be inherited natively."""
         return self._type == 'check'
 
+    def requires_triggers(self):
+        return (
+            not self.is_natively_inherited()
```
1st1, line 231: I assume `is_natively_inherited` isn't needed anymore?

edb/server/bootstrap.py
```diff
@@ -654,8 +656,9 @@ async def _init_stdlib(cluster, conn, testmode, global_ids):
 async def _execute_ddl(conn, sql_text):
     try:
         if debug.flags.delta_execute:
-            debug.header('Delta Script')
-            debug.dump_code(sql_text, lexer='sql')
+            # debug.header('Delta Script')
+            # debug.dump_code(sql_text, lexer='sql')
+            print(sql_text)
```
1st1, line 661: Revert?

tests/test_server_proto.py
```diff
@@ -37,6 +37,7 @@
 class TestServerProto(tb.QueryTestCase):
 
     ISOLATED_METHODS = False
+    SERIALIZED = True
```
1st1, line 40: Was there some flaky test that needs this?

edb/pgsql/delta.py
```diff
@@ -1436,157 +1513,91 @@ def get_source_and_pointer_ctx(cls, schema, context):
...
+        bases = set(obj.get_bases(schema).objects(schema))
+        orig_bases = set(obj.get_bases(orig_schema).objects(orig_schema))
+
+        for new_base in bases - orig_bases:
+            self.schedule_inhviews_update(
+                schema, context, new_base, update_ancestors=True)
+
+        for old_base in orig_bases - bases:
+            self.schedule_inhviews_update(
+                schema, context, old_base, update_ancestors=True)
```
1st1, line 1573: Wouldn't this sometimes update same ancestors multiple times?

edb/pgsql/common.py
elprans: It wouldn't be a single enum.  Every schema object has its own set.

edb/common/context.py
elprans: Debug.

edb/pgsql/deltadbops.py
elprans: It's a bit misleadingly named now, but the check is still valid.  I'll adjust.

tests/test_server_proto.py
elprans: The whole test suite is flaky because it uses `ISOLATED_METHODS=False` and so is subject to races with itself, because not all test cases use transactions.

edb/server/bootstrap.py
elprans: Done

edb/pgsql/delta.py
elprans: No, everything is eventually get collected into a set, so there are no repetitions.  You can schedule the view update on the same target as many times as you'd like.

Issue #1472, opened 2020-06-09
Changing user password by ALTER USER doesn't work with Amazon RDS


ambv:
RDS doesn't expose an actual superuser to customers. Instead, a special user is provided to customers that sort of serves that role. However, when trying to change the password, we get:

```
edgedb> ALTER ROLE edgedb {
.......     SET password := 'new password';
....... };
Error: ERROR: InternalServerError: must be superuser to alter superusers
```

Technically all EdgeDB users are superusers so the expected behavior would be that we can set the password.

PR #1477, opened 2020-06-16
Fix handling of ROLE commands on RDS

elprans:
Although bootstrap on RDS accounts for the Postgres role configuration
specifics on RDS, `CREATE ROLE` and `ALTER ROLE` currently fail, because
the compiler and the DDL subsystem have no knowledge of those specifics.
Fix this.

Fixes: #1472

### DIFF
--- a/edb/pgsql/delta.py
+++ b/edb/pgsql/delta.py
@@ -3781,13 +3781,25 @@ class CreateRole(ObjectMetaCommand, adapts=s_roles.CreateRole):
         role = self.scls
         schema = ObjectMetaCommand.apply(self, schema, context)
 
+        membership = list(role.get_bases(schema).names(schema))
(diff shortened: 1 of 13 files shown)


edb/pgsql/delta.py
```diff
@@ -3807,18 +3819,37 @@ def apply(
...
                 builtin=role.get_builtin(schema),
-            ),
-        )
+            )
+        if self.has_attribute_value('is_superuser'):
+            superuser_flag = False
+            if context.backend_superuser_role:
+                # If the cluster is exposing an explicit superuser role,
+                # become a member of that instead of creating a superuser
+                # role directly,
```
1st1, line 3837:
```suggestion
                # role directly.
```

PR #1483, opened 2020-06-19
Aliases

vpetrovykh:
I'm quite puzzled here by what is `_handle_view_op` trying to do. It crashes because even if I remove the alias from the schema so that I can create the new alias, the `prev_ir = cls._compile_view_expr(prev_expr.qlast, classname, schema, context)` dies on the duplicate name bug because it literally tries to re-compile previous view on the old schema, where it exists by definition. This appears to be very wrong. We either need to extract the data necessary for the delta from the schema directly or have some hack that just allows for `_compile_view_expr` to overwrite existing schema objects when given a certain flag.

See `test_migrations_equivalence_42` to trigger this.

### DIFF
--- a/edb/edgeql/compiler/stmtctx.py
+++ b/edb/edgeql/compiler/stmtctx.py
@@ -153,6 +153,11 @@ def fini_expression(
             view_own_pointers = view.get_pointers(ctx.env.schema)
             for vptr in view_own_pointers.objects(ctx.env.schema):
                 _elide_derived_ancestors(vptr, ctx=ctx)
+                ctx.env.schema = vptr.set_field_value(
+                    ctx.env.schema,
+                    'is_from_alias',
+                    True,
+                )
 
                 tgt = vptr.get_target(ctx.env.schema)
                 assert tgt is not None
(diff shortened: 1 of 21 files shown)


edb/edgeql/compiler/stmtctx.py
```diff
@@ -153,6 +153,11 @@ def fini_expression(
             view_own_pointers = view.get_pointers(ctx.env.schema)
             for vptr in view_own_pointers.objects(ctx.env.schema):
                 _elide_derived_ancestors(vptr, ctx=ctx)
+                ctx.env.schema = vptr.set_field_value(
+                    ctx.env.schema,
+                    'is_from_alias',
```
elprans, line 158: This field shouldn't be necessary,  There is already an `is_derived` field that already serves the same purpose.

edb/schema/delta.py
```diff
@@ -239,23 +240,23 @@ def get_local_attribute_value(
         else:
             return None
 
-    def get_attribute_source_context(
+    def get_orig_attribute_value(
```
elprans, line 243: Why this spurious method reordering?

elprans: Ignore, this is from my commit.

elprans:
I merged #1494 which should fix most cases of alias alteration.

elprans:
This is superseded by #1494 and #1510.

PR #1484, opened 2020-06-19
Indexes

vpetrovykh:
This is generally how I plan to tackle the issue of re-compiling the expressions affected by an alter. `DROP/CREATE` would be performed where currently the check for affecting expressions is made.

See `test_migrations_equivalence_index_03` to trigger this.

### DIFF
--- a/edb/schema/delta.py
+++ b/edb/schema/delta.py
@@ -240,16 +240,6 @@ class Command(struct.MixedStruct, metaclass=CommandMeta):
         else:
             return None
 
-    def get_attribute_source_context(
-        self,
-        attr_name: str,
-    ) -> Optional[parsing.ParserContext]:
-        op = self.get_attribute_set_cmd(attr_name)
-        if op is not None:
-            return op.source_context
-        else:
-            return None
-
     def get_orig_attribute_value(
         self,
         attr_name: str,
@@ -257,6 +247,16 @@ class Command(struct.MixedStruct, metaclass=CommandMeta):
         op = self.get_attribute_set_cmd(attr_name)
         if op is not None:
             return op.old_value
+        else:
+            return None
+
+    def get_attribute_source_context(
+        self,
+        attr_name: str,
+    ) -> Optional[parsing.ParserContext]:
+        op = self.get_attribute_set_cmd(attr_name)
+        if op is not None:
+            return op.source_context
         else:
             return None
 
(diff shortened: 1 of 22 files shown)


edb/schema/objects.py
```diff
@@ -2553,15 +2583,57 @@ def delta_property(
...
+        )
+
+        # Check to see if this field's inherited status has changed.
+        # If so, this is defintely a change.
+        our_ifs = ours.get_inherited_fields(our_schema)
+        their_ifs = theirs.get_inherited_fields(their_schema)
+
+        if (fname in our_ifs) != (fname in their_ifs):
+            # The change in inherited status decreases the similarity.
+            similarity *= 0.5
```
elprans, line 2633: I would set a higher coef here.  Generally if you just want something to register as "altered" you simply need any value < 1, but you should be careful not to tip the overall coefficient product to below 0.6 where it would no longer be considered an "alter".

edb/schema/delta.py
```diff
@@ -990,36 +990,111 @@ def _build_alter_cmd_stack(
...
+                    # schedule it to be re-created.
+                    idx_subj = ref.get_subject(schema)
+
+                    dump(ref.get_expr(schema).origqlast, schema=schema, marker='delta.py:1020')
+                    dump(idx_subj, schema=schema, marker='delta.py:1021')
+                    dump(idx_subj.get_name(schema), schema=schema, marker='delta.py:1022')
+
+                    module, shortname = idx_subj.get_name(schema).split('::', 1)
+
+                    finalize_ast.append(
```
elprans, line 1026: See `ReferencedObject.delete` and `_propagate_field_alter` on how to use `_build_alter_cmd_stack` to build arbitrary alteration subtrees.  This should also need to be factored out in a way that allows moving index-specific operations to `indexes.py` etc.

edb/schema/delta.py
```diff
@@ -1000,18 +1001,52 @@ def _build_alter_cmd_stack(
...
+                    # entity is a function, alter it to drop/create the
+                    # body.
+                    delta_drop, cmd_drop = ref._get_command_stack(
+                        schema, context, AlterObject)
+                    delta_create, cmd_create = ref._get_command_stack(
+                        schema, context, AlterObject)
+
+                    # Copy own fields into the create command.
+                    value = ref.get_explicit_field_value(schema, fn, None)
+                    cmd_drop.set_attribute_value(fn, None)
```
elprans, line 1043: I don't think this will work in the general case. You cannot simply "drop" a body of a function object and same applies to the computables.  What you *can* do is replace the expression with something like `<exprtype>{}`.

vpetrovykh: I've updated the code to use a dummy body for functions. It uses a string cast, selecting an object, or an `{}` with a cast (for collections, since we can't directly cast a `str` into an arbitrary array or tuple).

edb/schema/indexes.py
```diff
@@ -359,6 +359,20 @@ class DeleteIndex(
...
+    def _apply_fields_ast(
+        self,
+        schema: s_schema.Schema,
+        context: sd.CommandContext,
+        node: qlast.DDLOperation,
+    ) -> None:
+        # When deleting index the original expression acts as a way to
+        # identify the index, so we must retrieve it and write it as "expr".
+        for fop in self.get_subcommands(type=sd.AlterObjectProperty):
+            if fop.property == 'expr':
```
elprans, line 371: Note: use `self.get_attribute_value()` (and its variants) instead of an open-coded loop in situations like this.

elprans:
Merged as part of #1510.

PR #1493, opened 2020-06-23
Implement `SET password_hash` for roles

### LINKED ISSUE #1492: Implement `CREATE ROLE SET password_hash`
This is needed for [`DESCRIBE ROLES`](https://github.com/edgedb/edgedb/issues/1491) to emit complete statements.

tailhook:
Fixes #1492

### DIFF
--- a/edb/schema/roles.py
+++ b/edb/schema/roles.py
@@ -50,6 +50,13 @@ class Role(so.GlobalObject, so.InheritingObject,
         allow_ddl_set=True,
         inheritable=False)
 
+    password_hash = so.SchemaField(
+        str,
+        default=None,
+        allow_ddl_set=True,
+        ephemeral=True,
+        inheritable=False)
+
 
 class RoleCommandContext(
         sd.ObjectCommandContext[Role],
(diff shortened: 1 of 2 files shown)


edb/schema/roles.py
```diff
@@ -49,6 +49,13 @@ class Role(so.GlobalObject, so.InheritingObject,
         allow_ddl_set=True,
         inheritable=False)
 
+    password_hash = so.SchemaField(
```
elprans, line 52: Why not just allow setting hash to the `password` field directly?

tailhook: Do you mean detecting prefix in the password? I think it's weird. While it's unlikely that your password would start with `SCRUM-SHA-256$`, but still matching on that text sounds like a bad design. Or this isn't what you propose?

elprans: OK, I wanted to see if you considered that option.  A separate DDL field is OK too.

edb/schema/roles.py
```diff
@@ -72,9 +79,24 @@ def _process_role_body(
...
             cmd.set_attribute_value('password', salted_password)
 
+        password_hash = cmd.get_attribute_value('password_hash')
+        if password_hash is not None:
+            if not password_hash.startswith('SCRAM-SHA-256$'):
+                raise errors.InvalidValueError(
+                    'only SCRAM-SHA-256 hashes are supported',
+                    context=astnode.context,
+                )
+            cmd.set_attribute_value('password', password_hash)
```
elprans, line 98: Please add a check that the hash is well-formed. (`scram.parse_verifier` should do the trick).

Neustradamus:
Good job!

Linked to:

PR #1494, opened 2020-06-23
Remove most direct references to schema objects at DDL>IR phase

elprans:
The DDL AST -> Delta IR translation phase is not supposed to rely on
objects being present in the schema, and there are a few places where
that still happens, most notably in the handling of Aliases.  The latter
is because `CREATE ALIAS` currently tries to pretend it's really a
`CREATE TYPE ... USING (expr)`, and to know wheter it's `CREATE TYPE` or
`CREATE SCALAR TYPE` it needs to compile the alias expression at AST
translation phase.  Fix this by making aliases into thin wrappers around
the nested type command.  A side-effect of this is that we now have an
explicit `Alias` object in the schema.

### DIFF
--- a/edb/schema/delta.py
+++ b/edb/schema/delta.py
@@ -22,6 +22,7 @@ from typing import *
 
 import collections
 import collections.abc
+import contextlib
 import itertools
 import uuid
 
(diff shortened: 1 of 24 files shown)


edb/schema/delta.py
```diff
@@ -827,6 +828,13 @@ def store_value(self, key: Hashable, value: Any) -> None:
     def get_value(self, key: Hashable) -> Any:
         return self._values.get(key)
 
+    @contextlib.contextmanager
+    def suspend_dep_verification(self) -> Iterator[CommandContext]:
+        dep_ver = self.disable_dep_verification
+        self.disable_dep_verification = True
+        yield self
```
1st1, line 835: try..finally, no?

elprans: Yes, of course. Good catch.

PR #1497, opened 2020-06-26
Describe config

### LINKED ISSUE #1490: Implement DESCRIBE CONFIG
Needed for `edgedb dump --all` edgedb/edgedb-cli#81.

tailhook:
Adds `DESCRIBE SYSTEM CONFIG`. I'm not sure how to test that. Given that some changes (listen_adresses) interfere with all other tests, some with few tests (`Auth` only blocks auth test, ports blocks more).

Also I'm not sure:
1. Do we need to emit `CONFIG SYSTEM RESET Auth`, as there is a single auth record on empty database, at least in dev mode
2. How to handle "set" parameters like `listen_addresses`. Currently, we can't assign an empty set. But `RESET` may mean a different thing for purposes of dump.

Fixes #1490

### DIFF
--- a/edb/pgsql/metaschema.py
+++ b/edb/pgsql/metaschema.py
@@ -1857,10 +1857,6 @@ class SysConfigFunction(dbops.Function):
         BEGIN
         RETURN QUERY EXECUTE $$
             WITH
-                data_dir AS
-                    (SELECT setting AS dir FROM pg_settings
-                     WHERE name = 'data_directory'),
-
                 config_spec AS
                     (SELECT
                         s.key AS name,
--- a/edb/pgsql/metaschema.py
+++ b/edb/pgsql/metaschema.py
@@ -31,6 +31,7 @@ from edb.common import exceptions
 from edb.common import uuidgen
 
 from edb.edgeql import qltypes
+from edb.edgeql import quote as qlquote
 
 from edb.schema import migrations  # NoQA
 from edb.schema import modules as s_mod
@@ -38,6 +39,8 @@ from edb.schema import name as sn
 from edb.schema import objtypes as s_objtypes
 
 from edb.server import defines
+from edb.server import compiler as edbcompiler
+from edb.server import bootstrap as edbbootstrap
 
 from . import common
 from . import dbops
@@ -471,7 +474,6 @@ class ExtractJSONScalarFunction(dbops.Function):
             args=[('val', ('jsonb',)), ('json_typename', ('text',)),
                   ('msg', ('text',), 'NULL'), ('det', ('text',), "''")],
             returns=('text',),
-            # Same volatility as jsonb_assert_type
             volatility='stable',
             text=self.text)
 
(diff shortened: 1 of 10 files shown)


elprans:
> 1. Do we need to emit `CONFIG SYSTEM RESET Auth`

We don't.  The `Trust` method is only added in dev mode for convenience.

>  2. How to handle "set" parameters like `listen_addresses`. Currently, we can't assign an empty set.

You cannot assign an empty set, because `listen_addresses` is declared as `required`.  What is the scenario in which you need to write an empty set there?

edb/pgsql/metaschema.py
```diff
@@ -1836,6 +1839,115 @@ def __init__(self) -> None:
...
+
+        cfg = schema.get('cfg::Config')
+        items = []
+        for pn, p in cfg.get_pointers(schema).items(schema):
+            if pn in ('id', '__type__'):
+                continue
+            ptype = p.get_target(schema)
+            mult = p.get_cardinality(schema) is qltypes.SchemaCardinality.MANY
+            if isinstance(ptype, s_objtypes.ObjectType):
+                if pn not in {'sessobj', 'sysobj'}:
```
elprans, line 1923: These are test config objects that only get populated in dev mode.  There's no need to filter them out in general.

tailhook:
> You cannot assign an empty set, because `listen_addresses` is declared as `required`. What is the scenario in which you need to write an empty set there?

This is weird because the default is an empty set. 

The scenario is obviously "DESCRIBE SYSTEM CONFIG" used for dump/restore. I don't think omitting property is good idea here. Especially because this is generic code that works for all "multi" properties. And some future multi-property can have a non-empty default but having valid behavior with an empty value set manually.

edb/pgsql/metaschema.py
elprans: I don't think we should really care about devmode vs non-devmode scenario.  The worst case that would happen is that you'll have a couple of extra definitions in `cfg`, no big deal.  On the other hand, we will certainly be adding testing stuff to `cfg`, and it's easy to overlook this filter here, making it incomplete and moot.

tailhook: I still don't understand what to do:
1. These are objects and currently each object have a special treatment. I do know that to do with `auth` and `ports` but don't how to serialize `sessobj` and `sysobj`. Also If I understand correctly `sessobj` has something to the current session, so does not need to be in dump regardless of the dev/non-dev mode.
2. Worst case is dump that has to be edited manually, not "extra records that are silently ignored"

elprans: OK, let's keep the check as is to progress.  Ideally, we'd need to generalize the handling of nested config objects to avoid having to add code here when we add stuff to `cfg::` (i.e not hardcode `Port` and `Auth`, but iterate over `cfg::Config` links instead).  You actually don't need to declare subclasses `dbops.Function` to declare functions, you can construct a `dbops.Function` directly in an introspection loop in `generate_more_support_functions`.

tailhook:
Looks like this should be good enough at this point. We can do more generic handling of ports and auth later on.

PR #1498, opened 2020-06-26
Fix constants test that was skipped before

tailhook:


tests/test_edgeql_syntax.py
```diff
@@ -3923,3 +3876,54 @@ def test_edgeql_syntax_describe_04(self):
         """
         DESCRIBE TYPE foo::Bar AS DDL VERBOSE;
         """
+
+
+class TestEdgeQLUnderscoreConstants(server_base.QueryTestCase):
```
elprans, line 3881: Arguably, this isn't a pure syntax test, so I'd move into some other functional suite to avoid adding another suite (query suites have setup overhead).

tailhook: The point was to test constants. There is a comment about that: it's impossible to tests constants without DDL (because of constant extraction). It was your request to move test here :)

elprans: > There is a comment about that: it's impossible to tests constants without DDL

If you only need to test the lexer behavior, then just use the docstring-style with a simple query.  These tests call the parser directly and bypass constant extraction.

tailhook: So there was a test in expressions. You asked to move it into the syntax: https://github.com/edgedb/edgedb/pull/1416#pullrequestreview-418523029

Now I see that yes, I need to have a functional test, because syntax test doesn't ensure that constant is parsed as the right token and that the numeric value is right.

So if I move simple `SELECT 1_024` to the `test_edgeql_expressions` where should I move functional tests that involve DDL?

elprans: My earlier review and request was specifically about pure syntax tests:

> If the purpose of those tests is to test uninterpreted handling of literals without necessarily hitting Postgres, then yes.

Also, I think we need some sort of internal config toggle similar to `__internal_testmode` to disable constant extraction instead of relying on the DDL hack, because the latter will stop working if we start extracting from DDL also.

1st1: +1

tailhook: Another thing that I've just noticed is that DDL tests are run sequentially because of possible deadlocks. So this test in current mode can't be placed anywhere except `test_edgeql_ddl.py` as far as I understand.

elprans: That's a good point. I'm ok with `test_edgeql_ddl` as the place for this.

elprans:
It looks like @1st1 beat you to it in 7de95af461.  Closing.

PR #1500, opened 2020-06-26
Add `make clean`

ambv:
This allows us to clean outdated builds and dependencies from the repo directory, as well as retrigger Cython by removing compiled .c and .h files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,3 +25,7 @@ pygments:
 
 build:
 	pip install -Ue .[docs,test]
+
+
+clean:
+	git clean -Xf -e "!/*.code-workspace" -e "!/*.vscode"
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,7 @@ docs/_build
 
 !/.coveragerc
 /.vscode
+*.code-workspace
 /.pytest_cache
 /.mypy_cache
 /.vagga


Makefile
```diff
@@ -25,3 +26,13 @@ pygments:
...
 build:
 	pip install -Ue .[docs,test]
+
+
+clean:
+	rm -fr $(ROOT).eggs/
+	rm -fr $(ROOT)build/
+	rm -fr $(ROOT)dist/
+	rm -fr $(ROOT)docs/_build/
+	find $(ROOT)edb/server -path pgproto -prune -name "*.c" -delete
```
1st1, line 36: Maybe we use `git clean` that would take care of relevant c files automatically?

ambv: `git clean -X` sounds like a good plan, let me try this. I didn't know about it 🤦🏻‍♂️

Makefile
```diff
@@ -25,3 +25,7 @@ pygments:
 
 build:
 	pip install -Ue .[docs,test]
+
+
+clean:
+	git clean -Xn -e "!/*.code-workspace" -e "!/*.vscode"
```
1st1, line 31: Hm, can you add these to `.gitignore`?

ambv: Sure, will do.

Just note that this command says: "don't remove those". I have them in my global gitignore in my dotfiles and `git clean -X` also looks at your ~/.gitignore file. I don't *want* it to zap my workspace file, all of my configuration is there.

1st1: Most of us use vscode so having these in .gitignore makes sense.

ambv:
It's green and ready for merging.

Makefile
```diff
@@ -25,3 +25,7 @@ pygments:
 
 build:
 	pip install -Ue .[docs,test]
+
+
+clean:
+	git clean -Xf -e "!/*.code-workspace" -e "!/*.vscode"
```
1st1, line 31: Do you still need `-e "!/*.code-workspace" -e "!/*.vscode"` even though they're in gitignore?

The reason I'm asking is because I can see myself removing these after a few months thinking they are redundant. It wouldn't be great if I accidentally break your setup that way.

ambv: Rephrasing my comment above: those specify *exceptions*. I specifically **don't** want `git clean` to nuke them because cleaning the build does not equate deleting my editor configuration.

1st1: Apologies, I didn't quite understand how `git clean -X` works.

PR #1505, opened 2020-06-30
Add the syntax of altering a computable to the docs.

vpetrovykh:
The DDL docs for links and properties now include the syntax for
altering a computable expression and examples of using that syntax.

### DIFF
--- a/docs/edgeql/ddl/links.rst
+++ b/docs/edgeql/ddl/links.rst
@@ -153,14 +153,17 @@ Define a new link ``interests`` on the ``User`` object type:
         CREATE MULTI LINK friends -> User
     };
 
-Define a new link ``friends_in_same_town`` as a computable on the
-``User`` object type:
+Define a new link ``special_group`` as a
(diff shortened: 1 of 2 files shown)


docs/edgeql/ddl/links.rst
```diff
@@ -153,14 +153,16 @@ Define a new link ``interests`` on the ``User`` object type:
...
 
 .. code-block:: edgeql
 
     ALTER TYPE User {
-        CREATE LINK friends_in_same_town := (
-            SELECT __source__.friends FILTER .town = __source__.town)
+        CREATE LINK special_group := (
+            SELECT __source__.friends
+            FILTER .town = __source__.town
+        )
```
elprans, line 165: Add a link to the "Computables" section, which explains what `__source__` is etc.

vpetrovykh: I added links to computables from the links and properties examples.

PR #1507, opened 2020-07-01
Describe roles

tailhook:
Based on the #1497


edb/edgeql/ast.py
```diff
@@ -66,6 +66,12 @@ class CardinalityModifier(s_enum.StrEnum):
     Required = 'REQUIRED'
 
 
+class DescribeGlobal(s_enum.StrEnum):
+    Schema = 'SCHEMA'
+    SystemConfig = 'SYSTEM CONFIG'
+    Roles = 'ROLES'
```
1st1, line 72: Maybe `DESCRIBE ROLE;` would be OK to describe all roles? That way we don't need to add another unreserved keyword (which are still perceived as regular keywords by users).

edb/edgeql/codegen.py
```diff
@@ -1659,10 +1659,10 @@ def visit_ReleaseSavepoint(self, node: qlast.ReleaseSavepoint) -> None:
 
     def visit_DescribeStmt(self, node: qlast.DescribeStmt) -> None:
         self.write(f'DESCRIBE ')
-        if node.object:
-            self.visit(node.object)
+        if isinstance(node.object, qlast.DescribeGlobal):
+            self.write(str(node.object))
```
1st1, line 1663: Please add `DescribeGlobal.to_edgeql()` method like in enums in `edb/edgeql/qltypes.py`. Makes things a little clearer in places like this.

tests/test_roles.py
```diff
@@ -0,0 +1,27 @@
...
+        roles = next(iter(await self.con.fetchall("""DESCRIBE ROLES""")))
+        base1 = roles.index('CREATE SUPERUSER ROLE `base1`;')
+        base2 = roles.index('CREATE SUPERUSER ROLE `base2`;')
+        child1 = roles.index('CREATE SUPERUSER ROLE `child1`')
+        child2 = roles.index('CREATE SUPERUSER ROLE `child2`')
+        child3 = roles.index('CREATE SUPERUSER ROLE `child3`')
+        assert child1 > base1, roles
+        assert child2 > base2, roles
+        assert child3 > child2 and child3 > base1, roles
+        assert "SET password_hash := 'SCRAM-SHA-256$4096:" in roles, roles
```
1st1, line 27: Need a similar test for `DESCRIBE CONFIG`

tests/test_roles.py
```diff
@@ -0,0 +1,27 @@
...
+                SET password := 'test'
+            };
+        """)
+        roles = next(iter(await self.con.fetchall("""DESCRIBE ROLES""")))
+        base1 = roles.index('CREATE SUPERUSER ROLE `base1`;')
+        base2 = roles.index('CREATE SUPERUSER ROLE `base2`;')
+        child1 = roles.index('CREATE SUPERUSER ROLE `child1`')
+        child2 = roles.index('CREATE SUPERUSER ROLE `child2`')
+        child3 = roles.index('CREATE SUPERUSER ROLE `child3`')
+        assert child1 > base1, roles
```
1st1, line 24:
```suggestion
        self.assertGreater(child1, base1, roles)
```

edb/pgsql/metaschema.py
```diff
@@ -2591,6 +2672,7 @@ async def generate_support_functions(conn, schema):
         dbops.CreateFunction(IsinstanceFunction()),
         dbops.CreateFunction(GetSchemaObjectNameFunction()),
         # dbops.CreateFunction(QuoteLiteralFunction()),
```
elprans, line 2674: Remove this?

tailhook: Yes. That was in describe config PR, haven't rebased yet at that point.

tests/test_roles.py
tailhook: As I've written in #1497, I don't know how to make a valid test for this. Should I create a new instance just for DESCRIBE CONFIG test? We can probably make an instance for `dump --all` and then reuse it for `DESCRIBE CONFIG`, because too many instances doesn't sound a great idea.

1st1: Can you just run `DESCRIBE CONFIG` and check that the command was successful at all? That's a very surface test, but still a test.

tailhook: Ah, yes. At the point of writing this, I've assumed that syntax tests do run queries. But now as I know they aren't I've added some smoke test.

Issue #1461, opened 2020-06-02
RFC1000 implementation


elprans:
Will use this to track the implementation of [RFC 1000](https://github.com/edgedb/rfcs/blob/master/text/1000-migrations.rst).

haikyuu:
Great news. 🎉

PR #1510, opened 2020-07-02
Initial implementation of the new migration system (RFC 1000)

elprans:
This puts in place the initial bits of work on implementing RFC 1000: new
syntax for the `MIGRATION` commands as well as the necessary plumbing
changes to record and track migration objects.

Issue: #1461


edb/schema/expr.py
```diff
@@ -234,6 +243,12 @@ def qlast(self) -> qlast_.Base:
             self._qlast = qlparser.parse_fragment(self.text)
         return self._qlast
 
+    @property
+    def origqlast(self) -> qlast_.Base:
+        if self._origqlast is None:
+            self._origqlast = qlparser.parse_fragment(self.origtext)
```
1st1, line 249: Is it safe to cache it like this?

edb/schema/pointers.py
```diff
@@ -193,6 +193,14 @@ class Pointer(referencing.ReferencedInheritingObject,
     computable = so.SchemaField(
         bool,
         default=None,
+        compcoef=0.99,
+    )
+
+    # True, ff this pointer is defined in an Alias.
```
1st1, line 199:
```suggestion
    # True, if this pointer is defined in an Alias.
```

edb/schema/referencing.py
```diff
@@ -687,6 +686,41 @@ def _validate(
...
+        context: sd.CommandContext,
+        bases: Any,
+    ) -> Sequence[str]:
+
+        mcls = self.get_schema_metaclass()
+        default_base = mcls.get_default_base_name()
+
+        if isinstance(bases, so.ObjectCollectionShell):
+            base_names = [
+                b.name for b in bases.items if b.name is not None
```
1st1, line 686: It's hard to trace the API while reviewing, so I'll just ask: why `.name` and not `.get_name(schema)`?

edb/testbase/lang.py
```diff
@@ -323,8 +323,8 @@ def run_ddl(cls, schema, ddl, default_module=defines.DEFAULT_MODULE_ALIAS):
                 ddl_plan = s_ddl.compile_migration(
                     ddl_plan, target_schema, current_schema)
 
-            elif isinstance(stmt, qlast.Delta):
-                # APPLY MIGRATION
+            elif isinstance(stmt, qlast.CommitMigration):
+                # COMMIT MIGRATION
```
1st1: Please tell me one last thing about a detail for the connection pool. It's something I wanted to discuss and then forgot to ask ->

edb/schema/functions.py
```diff
@@ -970,6 +970,24 @@ def get_verbosename(
         sn = self.get_shortname(schema)
         return f"function '{sn}{params.as_str(schema)}'"
 
+    def get_dummy_body(self, schema: s_schema.Schema) -> expr.Expression:
+        rt = self.get_return_type(schema)
+
+        if rt.is_scalar():
+            # scalars and enums can be cast from a string
+            text = f'SELECT <{rt.get_displayname(schema)}>""'
```
1st1, line 1023: I'm not sure I understand how casting from empty string is correct here.

1st1: Ah, this isn't supposed to be a valid expression in runtime. I suggest adding a docstring to this method explaining what's it for.

edb/schema/functions.py
```diff
@@ -970,6 +970,24 @@ def get_verbosename(
...
+        elif rt.is_object_type():
+            # just grab an object of the appropriate type
+            text = f'SELECT {rt.get_displayname(schema)} LIMIT 1'
+        else:
+            # Can't easily create a valid cast, so just cast empty set
+            # into the given type. Technically this potentially breaks
+            # cardinality requirement, but since this is a dummy
+            # expression it doesn't matter at the moment.
+            text = f'SELECT <{rt.get_displayname(schema)}>{{}}'
```
1st1, line 1032: Maybe this is the only branch that's actually needed?

edb/schema/expr.py
elprans: This is actually dead code.

edb/schema/referencing.py
elprans: It's not an object it's an `ObjectShell`, but it may be worth imitating the object API on it.

edb/testbase/lang.py
elprans: Well, all the current tests that test migrations have been converted, there are plenty of them.

edb/schema/referencing.py
1st1: +1 to imitate it, for sure.

edb/schema/functions.py
elprans: Not really. We cannot return an empty set from a function that is not returning an `OPTIONAL`, and most scalar functions are like that.  But the `else` branch is not entirely correct either, it should be `SELECT {rt.get_displayname(schema)} LIMIT 1`

edb/schema/objects.py
```diff
@@ -1329,24 +1252,174 @@ def as_shell(self, schema: s_schema.Schema) -> ObjectShell:
...
+            fn for fn, f in type(self).get_fields().items() if f.ddl_identity
+        ]
+
+        ddl_identity: Optional[Dict[str, Any]]
+        if ddl_id_fields:
+            ddl_identity = {}
+            for fn in ddl_id_fields:
+                v = self.get_field_value(schema, fn)
+                if v is not None:
+                    ddl_identity[fn] = v
```
1st1, line 1269: Does None mean default (= insignificant for identity) for all fields?

edb/schema/objects.py
```diff
@@ -1329,24 +1252,174 @@ def as_shell(self, schema: s_schema.Schema) -> ObjectShell:
...
+        new_schema: Optional[s_schema.Schema],
     ) -> None:
         ff = cls.get_fields(sorted=True).items()
         fields = {fn: f for fn, f in ff
                   if f.simpledelta and not f.ephemeral and f.introspectable}
 
         if old and new:
-            if old_schema is None:
-                raise ValueError("`old` provided but `old_schema is None")
+            assert \
```
1st1, line 1417: nit: IMO a regular `if ... raise AssertionError` is better in cases where so many `\` are used.

edb/schema/objects.py
```diff
@@ -1514,7 +1598,7 @@ def delta_sets(
             # and `old_schema`.
             comp = x.compare(
                 y,
-                our_schema=new_schema,
+                our_schema=new_schema,  # type: ignore
```
1st1, line 1601: add `assert new_schema is not None`?

docs/edgeql/ddl/migrations.rst
```diff
@@ -7,60 +7,103 @@ Migrations
...
 :eql-statement:
 
-Create a new migration.
+Start a migration block.
 
 .. eql:synopsis::
 
-    [ WITH [ <module-alias> := ] MODULE <module-name> ]
-    CREATE MIGRATION <name> TO "{"
+    START MIGRATION <name> TO "{"
```
1st1, line 19:
```suggestion
    START MIGRATION TO "{"
```

docs/edgeql/ddl/migrations.rst
```diff
@@ -7,60 +7,103 @@ Migrations
...
         <sdl-declaration> ;
         [ ... ]
     "}" ;
 
+Parameters
+----------
+
+:eql:synopsis:`<sdl-declaration>`
+    Module contents defined using the declarative :ref:`EdgeDB schema
```
1st1, line 28: It's not necessarily contents of one module. Can we say "Schema defined using ..."?

docs/edgeql/ddl/migrations.rst
```diff
@@ -7,60 +7,103 @@ Migrations
...
 
 Description
 -----------
 
-``CREATE MIGRATION`` defines a migration of the schema to a new state.
-The target schema state is described using :ref:`SDL <ref_eql_sdl>`
-and the migration generates the necessary :ref:`DDL <ref_eql_ddl>`
-commands behind the scenes based on the current and the target state.
+``START MIGRATION`` defines a migration of the schema to a new state. The
+target schema state is described using :ref:`SDL <ref_eql_sdl>` amd describes
```
1st1, line 35:
```suggestion
target schema state is described using :ref:`SDL <ref_eql_sdl>` and describes
```

docs/edgeql/ddl/migrations.rst
```diff
@@ -71,84 +114,194 @@ syntax:
...
+
+.. eql:synopsis::
+
+    POPULATE MIGRATION ;
+
+Description
+-----------
+
+``POPULATE MIGRATION`` is used within a migration block started by
+:eql:stmt:``START MIGRATION`` to automatically fill the migration with
```
1st1, line 173:
```suggestion
:eql:stmt:`START MIGRATION` to automatically fill the migration with
```

docs/edgeql/ddl/migrations.rst
```diff
@@ -71,84 +114,194 @@ syntax:
...
-
+    ABORT MIGRATION ;
 
 Description
 -----------
 
-``COMMIT MIGRATION`` runs the DDL commands defined by the given migration.
-Once the migration is committed, it cannot be dropped.
+``ABORT MIGRATION`` is used to abort a migration block started by
+:eql:stmt:``START MIGRATION``.  Issuing ``ABORT MIGRATION`` outside of a
```
1st1, line 139:
```suggestion
:eql:stmt:`START MIGRATION`.  Issuing ``ABORT MIGRATION`` outside of a
```

docs/edgeql/ddl/migrations.rst
```diff
@@ -71,84 +114,194 @@ syntax:
...
+
+Description
+-----------
+
+``COMMIT MIGRATION`` runs the commands defined by the current migration and
+records the migration as the most recent migration in the database.
 
-:eql:synopsis:`<name>`
-    The name of the migration to discard.
+**Important:** ``COMMIT MIGRATION`` must be executed in a migration block.
```
1st1, line 297:
```suggestion
**Important:** ``COMMIT MIGRATION`` must be executed in a migration block
initiated by ``START MIGRATION``.
```

docs/edgeql/ddl/migrations.rst
```diff
@@ -71,84 +114,194 @@ syntax:
...
 
 
 Example
 -------
 
-Remove the "init" migration:
+Execute and create the current migration:
```
1st1, line 303:
```suggestion
Create and execute the current migration:
```

docs/tutorial/queries.rst
```diff
@@ -245,11 +243,11 @@ the ``first_name`` and ``last_name`` properties. This time we will use
     .........         }
     .........     }
     ......... };
-    CREATE MIGRATION
-    tutorial> COMMIT MIGRATION movies;
+    START MIGRATION
+    tutorial> POPULATE MIGRATION;
+    POPULATE MIGRATION
+    tutorial> COMMIT MIGRATION;
```
1st1, line 249: Can COMMIT MIGRATION automatically call POPULATE MIGRATION if it wasn't called before? That would simplify the initial manual schema creation in the tutorial(s) (although while complicating the spec a bit)

edb/edgeql/ast.py
```diff
@@ -662,28 +662,45 @@ def name(self) -> ObjectRef:
...
 
 
-class CreateMigration(CreateObject, Delta):
-    parents: typing.List[ObjectRef]
-    target: typing.Any
+class CreateMigration(CreateObject, Migration):
 
+    parent: typing.Optional[ObjectRef] = None
+    message: typing.Optional[str] = None
+    # XXX: compat
```
1st1, line 673: Can you elaborate here

edb/server/compiler/compiler.py
```diff
@@ -770,70 +764,220 @@ def _compile_and_apply_ddl_command(self, ctx: CompileContext, cmd):
...
+                modaliases=None,
+                single_unit=True,
+            )
+
+        elif isinstance(ql, qlast.AbortMigration):
+            mstate = current_tx.get_migration_state()
+            if mstate is None:
+                raise errors.QueryError(
+                    'unexpected ABORT MIGRATION:'
+                    ' not currently in a migration block',
```
1st1, line 964: nit: I see you moving the space to the beginning of literal continuation line, whereas I (and I thought we all) used to keep it at the end. Not sure why we're switching the coding style.

edb/schema/schema.py
```diff
@@ -1028,3 +1033,43 @@ def _get_operators(
         Tuple[s_oper.Operator, ...],
         tuple(schema._id_to_type[oid] for oid in objids),
     )
+
+
+@functools.lru_cache()
```
1st1, line 1038: I'd limit the size of this cache. Would 10 work?

edb/schema/functions.py
elprans: Actually the `else` branch is about collections.

edb/schema/objects.py
elprans: `None` means there's no information for this identity field at the moment.  I need to see if this can be turned into a hard error.

edb/schema/objects.py
elprans: I'm still holding hope that we might be able to disable assertion checks at some point given the amount of them we've been adding to make mypy happy.

edb/schema/objects.py
elprans: No, we can't assert it there, it's correlated with the collection being empty or not.  That whole function is a bit of a mess, we'll probably need to find a way to refactor it to avoid all this.

docs/tutorial/queries.rst
elprans: We should be using the CLI in the tutorials.  All `MIGRATION` commands are now considered low-level plumbing and we shouldn't encourage newcomers to use them or make them deceptively easy.

edb/server/compiler/compiler.py
elprans: This was Łukasz's suggestion. When used consistently as a style, it makes it less likely that you forget the space at the end of the line or have it stripped by the editor accidentally (happened to me a few times).

edb/schema/schema.py
elprans: Why so small?  Schema objects are lightweight (`PyObject header` + `uuid` basically), so the default limit shouldn't eat away at the memory too much.

edb/server/compiler/compiler.py
1st1: Well, now we have 5% of strings in new style and 95% in old. 🤷‍♂️ I wouldn't do this.

edb/schema/schema.py
1st1: Schema objects are small but they hold references to potentially large trees of other objects. So my question still stands: would a small number like 10 work? How often is this function called?

PR #1512, opened 2020-07-03
Forbid reference to link properties not from links

### LINKED ISSUE #1471: InternalServerError: 'NoneType' object has no attribute 'ptrref'
- EdgeDB Version: 1.0-alpha.3+dev.247.g04b74247

Steps to Reproduce:
Error occurs when a computed property has a link property path on a subquery, eg.

```
WITH MODULE schema
SELECT ObjectType {
  name,

  # error occurs with this line:
  annotationValues := (SELECT .annotations)@value

  # these are ok:
  # annotationValues := .annotations@value
  # annotationNames := (SELECT .annotations).name
  # annotationNames := .annotations.name
  
} FILTER EXISTS .annotations
```

nsidnev:
closes #1471

### DIFF
--- a/edb/edgeql/compiler/setgen.py
+++ b/edb/edgeql/compiler/setgen.py
@@ -264,6 +264,15 @@ def compile_path(expr: qlast.Path, *, ctx: context.ContextLevel) -> irast.Set:
             if ptr_expr.type == 'property':
                 # Link property reference; the source is the
                 # link immediately preceding this step in the path.
+                if path_tip.rptr is None:
+                    source = get_set_type(path_tip, ctx=ctx)
+                    s_dn = source.get_displayname(ctx.env.schema)
+                    raise errors.InvalidReferenceError(
+                        f"link property {ptr_name!r} "
+                        f"should be accessed from link, not from {s_dn!r}",
+                        context=ptr_expr.ptr.context,
+                    )
+
                 if isinstance(path_tip.rptr.ptrref,
                               irast.TypeIntersectionPointerRef):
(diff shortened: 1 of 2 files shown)


edb/edgeql/compiler/setgen.py
```diff
@@ -264,6 +264,15 @@ def compile_path(expr: qlast.Path, *, ctx: context.ContextLevel) -> irast.Set:
             if ptr_expr.type == 'property':
                 # Link property reference; the source is the
                 # link immediately preceding this step in the path.
+                if path_tip.rptr is None:
+                    source = get_set_type(path_tip, ctx=ctx)
+                    s_dn = source.get_displayname(ctx.env.schema)
+                    raise errors.InvalidReferenceError(
```
elprans, line 270: I'd use "unexpected reference to link property {ptr_name!r} outside of a path expression", and I think it should be an `EdgeQLSyntaxError` (we normally use `InvalidReferenceError` to indicate a syntactically-correct reference to a non-existent object).

nsidnev: I've changed error and message.

> we normally use `InvalidReferenceError` to indicate a syntactically-correct reference to a non-existent object

Thanks for the explanation. I raised an `InvalidReferenceError` following the message in the issue, but the syntax error here really looks more appropriate, given what is written in the issue:
> The only valid form of link property reference is .link@prop.

nsidnev:
There is a strange error in CI, and I don’t think that it is related to this PR. Maybe restarting a task can fix it?

PR #1514, opened 2020-07-07
Test a3 dumps; drop pure-Python dump/restore tooling

1st1:


ambv:
Is this part of some issue? I missed when this was decided.

1st1:
> I missed when this was decided.

When what was decided?

ambv:
To remove the Python dump/restore.

1st1:
> To remove the Python dump/restore.

Ah, right. Is this a controversial decision? We have the tooling implemented in Rust and by default that is what you get via the `edgedb` command. Invoking the pure-Python dump/restore via `edgedb-old` (I think) isn't practical nor there are use cases for that. The Python dump/restore implementation isn't even maintained anymore, and testing it adds like 15-20 seconds to the overall `edb test` time.

So at this point, I think we should drop it. And we also should drop pure Python repl, but I'll wait until @elprans no longer needs it.

ambv:
It's not controversial, I mostly thought this was part of the Python CLI removal which I missed. I sometimes still use it myself, too, because I can pdb into it.

1st1:
> Python CLI removal which I missed

Well, ultimately, maintaining two CLIs would be hard for us. So we'll have to have just one.

> I sometimes still use it myself, too, because I can pdb into it.

Maybe we should make Rust CLI more debuggable? What are the actual use cases?

ambv:
> What are the actual use cases?

For me it's a poor man's wireshark ;-) Helped me understand what is going on with connections and the protocol.

elprans:
> For me it's a poor man's wireshark ;-) Helped me understand what is going on with connections and the protocol.

Python CLI uses edgedb-python, there's nothing there that can't be reproduced in a Python REPL with `connection.fetch`

PR #1517, opened 2020-07-08
Skip defaults in `DISPLAY SYSTEM CONFIG`

tailhook:


tailhook:
Also implemented skipping empty sets, like `listen_address` (treating empty set as default).

PR #1518, opened 2020-07-08
Remove reliance on GHA target matrix in nightly builds

ambv:
Use Jinja for the same purpose. This allows for different Linux distributions to independently publish their artifacts when their own build and test was green.

It also speeds up execution by not building all Docker images on every matrix iteration.

Fixes edgedb/edgedb-pkg#4

### DIFF
--- a/.github/workflows.src/nightly.tpl.yml
+++ b/.github/workflows.src/nightly.tpl.yml
@@ -10,61 +10,39 @@ on:
       - nightly
 
 jobs:
-<% if targets.linux %>
-  build-linux:
+<% for tgt in targets.linux %>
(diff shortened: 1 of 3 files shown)


ambv:
It's probably easier to look at the generated `.github/workflows/nightly.yml` file in isolation and not as a diff.

elprans:
> Fixes edgedb/edgedb-pkg#4

Not completely.  We still want to look into the possibility of caching the images to avoid building them on every run.

ambv:
Okay, but let's make atomic changes. This one is a big change in the workflows, let's make it separately from any other change. I also want to split Build from Test but didn't want to mangle this in one big diff.

PR #1519, opened 2020-07-08
Allow empty set as a valid value for multi configuration properties

elprans:
`CONFIGURE SYSTEM SET multiprop := {}` is currently prohibited for no
good reason, so make amends and fix this.

It's unclear if we should allow empty sets in `single` config properties,
so I didn't touch those.

### DIFF
--- a/edb/ir/ast.py
+++ b/edb/ir/ast.py
@@ -402,10 +402,6 @@ class Set(Base):
         return f'<ir.Set \'{self.path_id}\' at 0x{id(self):x}>'
 
 
-class EmptySet(Set):
-    pass
-
-
 class Command(Base):
     __abstract_node__ = True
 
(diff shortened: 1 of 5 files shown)


elprans:
Hm. I can't reproduce the failure on my box :-|

elprans:
> Should I revert 1c0473d now?

Yes, it shouldn't be necessary once this PR is merged.

elprans:
I'm not sure why it's failing.  The CLI is simply crashing with `Error: future has timed out`, and I can't see why this test is special and why this change might have affected it (other dump tests are fine).  @tailhook any ideas how to debug?

elprans:
Yea, it's an unrelated failure, seems like an intermittent protocol or binding bug.  Filed as https://github.com/edgedb/edgedb-rust/issues/25 for now.

PR #1521, opened 2020-07-09
Allow annotations to be renamed.

### LINKED ISSUE #762: Annotations cannot be renamed
Currently we don't allow renaming abstract annotations. There's no specific reason to keep it forbidden.

vpetrovykh:
Add the syntax for ALTER ABSTRACT ANNOTATION.
Update the documentation.

Fixes #762.

### DIFF
--- a/docs/edgeql/ddl/annotations.rst
+++ b/docs/edgeql/ddl/annotations.rst
@@ -61,6 +61,72 @@ Declare an annotation ``extrainfo``.
     CREATE ABSTRACT ANNOTATION extrainfo;
 
 
+ALTER ABSTRACT ANNOTATION
+=========================
+
+:eql-statement:
+
+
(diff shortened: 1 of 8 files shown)


docs/edgeql/ddl/annotations.rst
```diff
@@ -61,6 +61,96 @@ Declare an annotation ``extrainfo``.
...
+
+.. eql:synopsis::
+
+    ALTER ABSTRACT ANNOTATION <name>
+    [ "{" ] <subcommand>; [...] [ "}" ];
+
+    # where <subcommand> is one of
+
+      RENAME TO <newname>
+      EXTENDING ...
```
elprans, line 80: Abstract annotations don't inherit, so remove this.

vpetrovykh: OK, I'll remove that from docs and grammar.

PR #1522, opened 2020-07-09
Expression normalization

vpetrovykh:
I still have tests to reconcile, but this is the general approach to normalization I have.

### DIFF
--- a/edb/edgeql/codegen.py
+++ b/edb/edgeql/codegen.py
@@ -756,6 +756,21 @@ class EdgeQLSourceGenerator(codegen.SourceGenerator):
             self._write_keywords(' EXTENDING ')
             self.visit_list(node.bases, newlines=False)
 
+    def _ddl_clean_up_commands(
+        self,
+        commands: Sequence[qlast.DDLOperation],
+    ) -> Sequence[qlast.DDLOperation]:
+        # Always omit orig_expr fields from output since we are
+        # using the original expression in TEXT output
+        # already.
+        return [
+            c for c in commands
+            if (
+                not isinstance(c, qlast.SetField)
+                or not c.name.startswith('orig_')
+            )
+        ]
+
     def _ddl_visit_body(
         self,
         commands: Sequence[qlast.DDLOperation],
@@ -772,18 +787,7 @@ class EdgeQLSourceGenerator(codegen.SourceGenerator):
                 )
             ]
 
-        if self.descmode:
-            # Omit orig_expr fields from output since we are
-            # using the original expression in TEXT output
-            # already.
-            commands = [
-                c for c in commands
-                if (
-                    not isinstance(c, qlast.SetField)
-                    or not c.name.startswith('orig_')
-                )
-            ]
-
+        commands = self._ddl_clean_up_commands(commands)
         if len(commands) == 1 and allow_short and not (
             isinstance(commands[0], qlast.ObjectDDL)
         ):
@@ -1523,7 +1527,8 @@ class EdgeQLSourceGenerator(codegen.SourceGenerator):
             if node.commands:
                 self.write(' {')
                 self._block_ws(1)
-                self.visit_list(node.commands, terminator=';')
+                commands = self._ddl_clean_up_commands(node.commands)
+                self.visit_list(commands, terminator=';')
                 self.new_lines = 1
             else:
                 self.write(' ')
--- a/edb/edgeql/compiler/normalization.py
+++ b/edb/edgeql/compiler/normalization.py
@@ -1,0 +1,480 @@
+#
+# This source file is part of the EdgeDB open source project.
+#
+# Copyright 2020-present MagicStack Inc. and the EdgeDB authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either nodeess or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+"""EdgeQL expression normalization functions."""
+
+
+from __future__ import annotations
+from typing import *
+
+import functools
+
+from edb.common.ast import base
+from edb.edgeql import ast as qlast
+from edb.schema import schema as s_schema
+from edb.schema import name as sn
+
+
+@functools.singledispatch
+def normalize(
+    node: Any,
+    *,
+    schema: s_schema.Schema,
+    modaliases: Mapping[Optional[str], str],
+    localnames: AbstractSet[str] = frozenset(),
+) -> None:
+    raise AssertionError(f'normalize: cannot handle {node!r}')
+
+
+def _normalize_recursively(
+    node: qlast.Base,
+    field: str,
+    value: Any,
+    *,
+    schema: s_schema.Schema,
+    modaliases: Mapping[Optional[str], str],
+    localnames: AbstractSet[str] = frozenset(),
+) -> None:
+    # We only want to handle fields that need to be traversed
+    # recursively: Base AST and lists. Other fields are essentially
+    # expected to be processed by the more specific handlers.
+    if isinstance(value, qlast.Base):
+        normalize(
+            value,
+            schema=schema,
+            modaliases=modaliases,
+            localnames=localnames,
+        )
+    elif isinstance(value, list):
+        if value and isinstance(value[0], qlast.Base):
+            for el in value:
+                normalize(
+                    el,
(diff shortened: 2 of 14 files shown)


edb/edgeql/compiler/normalization.py
```diff
@@ -0,0 +1,175 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either nodeess or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+"""EdgeQL nodeession normalization functions."""
```
elprans, line 20: heavy typo here

1st1:
@vpetrovykh make sure you add a test case for this: https://github.com/edgedb/edgedb/discussions/1588#discussioncomment-42247

vpetrovykh:
The actual functionality is there, but I still need to fix mypy complaints of the form:
- `Incompatible default for argument "ignorenames" (default has type "FrozenSet[<nothing>]", argument has type "Set[str]")`
- `Unsupported target for indexed assignment ("Mapping[Optional[str], str]")`
And write up a proper commit message.

Still, the code can be reviewed already.

edb/edgeql/codegen.py
```diff
@@ -756,6 +756,21 @@ def _ddl_visit_bases(self, node: qlast.BasesMixin) -> None:
...
+        commands: Sequence[qlast.DDLOperation],
+    ) -> Sequence[qlast.DDLOperation]:
+        # Always omit orig_expr fields from output since we are
+        # using the original expression in TEXT output
+        # already.
+        return [
+            c for c in commands
+            if (
+                not isinstance(c, qlast.SetField)
+                or not c.name.startswith('orig_')
```
elprans, line 770: With the new normalization we no longer need the `orig_` hacks at all.  Would it make sense to clean that up as part of this PR or would you like to do a followup?

edb/edgeql/compiler/normalization.py
```diff
@@ -0,0 +1,539 @@
...
+                [
+                    normalize(
+                        el,
+                        schema=schema,
+                        modaliases=modaliases,
+                        ignorenames=ignorenames,
+                    ) for el in value
+                ]
+            )
+
```
elprans, line 80:
```
else:
    raise AssertionError
```

edb/edgeql/compiler/normalization.py
```diff
@@ -0,0 +1,539 @@
...
+@normalize.register
+def normalize_DDL(
+    node: qlast.DDL,
+    *,
+    schema: s_schema.Schema,
+    modaliases: Mapping[Optional[str], str],
+    ignorenames: Set[str] = frozenset(),
+) -> qlast.DDL:
+    raise ValueError(f'normalize: cannot handle {node!r}')
```
elprans, line 113: Let's use `AssertionError` for unexpected code paths.

edb/edgeql/compiler/normalization.py
```diff
@@ -0,0 +1,539 @@
...
+
+@normalize.register
+def normalize_Base(
+    node: qlast.Base,
+    *,
+    schema: s_schema.Schema,
+    modaliases: Mapping[Optional[str], str],
+    ignorenames: Set[str] = frozenset(),
+) -> qlast.Base:
+    node = copy.copy(node)
```
elprans, line 90: I think we should just normalize destructively and not do `copy()`, which is shallow and may give a false sense that normalization returns a proper copy of the source tree.

edb/edgeql/compiler/normalization.py
```diff
@@ -0,0 +1,539 @@
...
+        if funcs:
+            # As long as we found some functions, they will be from
+            # the same module (the first valid resolved module for the
+            # function name will mask "std").
+            _, fname = funcs[0].get_name(schema).as_tuple()
+            _, module, name = sn.split_name(sn.shortname_from_fullname(fname))
+            node.func = (module, name)
+        elif None in modaliases:
+            # Even if the name was not resolved in the schema it may
+            # be the name of the object being defined, as such the
```
elprans, line 470: What do you mean by "the name of the object being defined"?  A body of a recursive function?

edb/schema/constraints.py
```diff
@@ -517,6 +512,43 @@ def get_ddl_identity_fields(
...
             return id_fields
 
+    @classmethod
+    def _ignorenames_from_ast(
+        cls,
+        schema: s_schema.Schema,
+        astnode: qlast.DDLOperation,
+        context: sd.CommandContext,
+    ) -> Set[str]:
+        ignorenames = super(ConstraintCommand, cls)._ignorenames_from_ast(
```
elprans, line 522:
```suggestion
        ignorenames = super()._ignorenames_from_ast(
```

edb/schema/constraints.py
```diff
@@ -517,6 +512,43 @@ def get_ddl_identity_fields(
...
+            # but we can get them from the schema.
+            objref = astnode.name
+
+            # Merge the context modaliases and the command modaliases
+            # to correctly infer the module where the constraint shoud
+            # be.
+            modaliases = dict(context.modaliases)
+            modaliases.update(
+                cls._modaliases_from_ast(schema, astnode, context))
+            module = modaliases.get(objref.module, objref.module)
```
elprans, line 540: This isn't necessary. `Schema.get()` takes `modaliases` as argument.

edb/schema/delta.py
```diff
@@ -608,6 +608,21 @@ def _modaliases_from_ast(
 
         return modaliases
 
+    @classmethod
+    def _ignorenames_from_ast(
```
elprans, line 612: This isn't an internal method, you're calling it from code that is outside of the class, so trim the leading underscore.

edb/schema/delta.py
```diff
@@ -721,10 +736,12 @@ def __init__(
         op: Command_T,
         *,
         modaliases: Optional[Mapping[Optional[str], str]] = None,
+        ignorenames: Set[str] = frozenset(),
```
elprans, line 739: `ignorenames` is too generic of a name to be used for a very specific purpose in the general context.  How about `localnames`?  Please also add the type annotation above and a comment on it with the purpose of the attribute.

edb/schema/pointers.py
```diff
@@ -1130,6 +1132,8 @@ def _process_create_or_alter_ast(
 
         This may be called in the context of either Create or Alter.
         """
+        from edb.edgeql.compiler import normalization as qlnorm
```
elprans, line 1135: Is there an import loop here?  You should be able to import the compiler modules normally.

edb/edgeql/codegen.py
vpetrovykh: I would probably make that a separate clean-up PR. Basically we want to land the name normalization and then we can trim this and make further improvements.

edb/edgeql/compiler/normalization.py
vpetrovykh: More like hand-written DDL for recursive types:
```
CREATE TYPE Foo {
    CREATE MULTI LINK children -> Foo;
}
```
Basically when this is not coming from schema, but is created by hand, the command can easily be appearing like that instead of being separated into `CREATE`/`ALTER`.

edb/edgeql/compiler/normalization.py
vpetrovykh: Fixed.

edb/edgeql/compiler/normalization.py
vpetrovykh: Actually this was deliberately omitted, in this generic field normalization we recursively process other AST nodes and lists of them, but we ignore other types that are assumed to be leaves in this recursion (and handled by the AST-node-specific handler).

I've renamed the function and added a comment to address this.

edb/edgeql/compiler/normalization.py
vpetrovykh: Sure, I've changed `ValueError` to `AssertionError`.

edb/schema/constraints.py
vpetrovykh: Done.

edb/schema/constraints.py
vpetrovykh: That's right, I've passed the modaliases directly to `get`.

edb/schema/delta.py
vpetrovykh: Leading underscore removed.

edb/schema/delta.py
vpetrovykh: Sure, `localnames` sounds good.

edb/schema/pointers.py
vpetrovykh: Hmm... it appears that there is no import loop, apparently that's a relic of some earlier refactoring.

Moved the import to the top.

edb/edgeql/compiler/normalization.py
vpetrovykh: Destructive normalization it is then.

Issue #1402, opened 2020-05-19
Number literals handle overflow and underflow inconsistently


vpetrovykh:
Edgedb version:  EdgeDB 1.0-alpha.3+dev.198.g9e8c0613
REPL version: edgedb-cli 0.1.0

I was a little surprised by the way finite-bit numbers are handled. The overflow and underflow doesn't happen consistently.

Consider the following for overflow:
```
victor> SELECT 11111111111111111111111111111111111111111111;
ERROR: EdgeQLSyntaxError: can't parse integer: number too large to fit in target type
```
So far so good, the `int64` overflowed as expected.

Let's see what the float does:
```
victor> SELECT 1e999;
{inf}
victor> SELECT 111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111.0;
{inf}
```
That's peculiar, I was expecting an error, much like with the `int64`. So let's try slightly different literals:
```
victor> SELECT <float64>'inf';
{inf}
victor> SELECT <float64>'1e999';
ERROR: NumericOutOfRangeError: "1e999" is out of range for type std::float64
victor> SELECT <float64>1e999n;
ERROR: NumericOutOfRangeError: "1000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000" is out of range for type std::float64
```

Similarly, there's an underflow issue:
```
victor> SELECT 1e-999;
{0}
victor> SELECT <float64>1e-999n;
[...]

PR #1523, opened 2020-07-09
Implement oveflow handling for floating point

tailhook:
#1402

### DIFF
--- a/edb/edgeql-rust/src/normalize.rs
+++ b/edb/edgeql-rust/src/normalize.rs
@@ -145,11 +145,17 @@ pub fn normalize<'x>(text: &'x str)
                 push_var(&mut rewritten_tokens, "__std__::float64",
                     next_var(variables.len()),
                     tok.start, tok.end);
+                let value = tok.value.replace("_", "").parse()
+                    .map_err(|e| Error::Tokenizer(
+                        format!("can't parse std::float64: {}", e),
+                        tok.start))?;
+                if value == f64::INFINITY || value == -f64::INFINITY {
+                    return Err(Error::Tokenizer(
+                        format!("number is out of range for std::float64"),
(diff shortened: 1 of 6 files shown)


edb/edgeql-rust/src/normalize.rs
```diff
@@ -145,11 +145,17 @@ pub fn normalize<'x>(text: &'x str)
                 push_var(&mut rewritten_tokens, "__std__::float64",
                     next_var(variables.len()),
                     tok.start, tok.end);
+                let value = tok.value.replace("_", "").parse()
+                    .map_err(|e| Error::Tokenizer(
+                        format!("can't parse std::float64: {}", e),
+                        tok.start))?;
+                if value == f64::INFINITY || value == -f64::INFINITY {
+                    return Err(Error::Tokenizer(
+                        format!("number is out of range for std::float64"),
```
1st1, line 154: What error would be raised for `SELECT <std::float32>1e10000`?  Would it be "number is out of range for std::float64" or "number is out of range for std::float32"? The latter would be less confusing to users.

tailhook: We don't have float32 syntax. it's basically just `<float32>` cast value with float64 syntax. On float32 postgres does the job:
```
edgedb> SELECT <float32>1e99;
ERROR: NumericOutOfRangeError: value out of range: overflow
edgedb> SELECT 1e99;
{1000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000}
```

On the other hand, we can use bare `float` to make it less confusing.

1st1: > ERROR: NumericOutOfRangeError: value out of range: overflow

This is a nice enough error, all is good

PR #1525, opened 2020-07-09
Split Build and Test phases into separate jobs

### LINKED ISSUE #1454: Split the package build-and-test workflow job into two jobs: build and test.
Currently, package build and built package tests are done as steps of the same job.  Unfortunately, it seems like Github Actions have an (undocumented) job run time limit on macOS of 60 minutes (the documentation claims that the limit is 6 hours).  It might be a fluke or a temporary bug.  Regardless, separating build and test into separate jobs seems like a cleaner arrangement anyway.

ambv:
Fixes #1454

### DIFF
--- a/.github/workflows.src/nightly.tpl.yml
+++ b/.github/workflows.src/nightly.tpl.yml
@@ -24,13 +24,6 @@ jobs:
         PKG_PLATFORM_VERSION: "<< tgt.platform_version >>"
         EXTRA_OPTIMIZATIONS: "true"
 
-    - name: Test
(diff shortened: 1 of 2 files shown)


elprans:
You didn't seem to have updated the actual workflow files.

ambv:
jon-stewart-really

elprans:
What about the release flow? Or do you want to test this on the nightly first?

ambv:
> Or do you want to test this on the nightly first?

Yes. It feels wrong to make changes to the release flow that I won't be testing now. If I got hit by a bus and the nightlies turned out to be borked, the old release flow will still work.

As soon as the nightly is green I will copy the changes to the main release workflow.

PR #1531, opened 2020-07-13
Replace Spectrum badge with GitHub discussions

Fogapod:


1st1:
Thanks! You can also create similar PRs for edgedb-python and edgedb-js repos if you have time.

PR #1538, opened 2020-07-14
Update the syntax docs for `REQUIRED`/`OPTIONAL`.

vpetrovykh:
Update the DDL and SDL docs for links and properties.

### DIFF
--- a/docs/edgeql/ddl/links.rst
+++ b/docs/edgeql/ddl/links.rst
@@ -21,7 +21,8 @@ CREATE LINK
     [ WITH <with-item> [, ...] ]
     {CREATE|ALTER} TYPE <TypeName> "{"
       [ ... ]
-      CREATE [ REQUIRED ] [{SINGLE | MULTI}] LINK <name>
+      CREATE [{REQUIRED | OPTIONAL}] [{SINGLE | MULTI}]
+        LINK <name>
         [ EXTENDING <base> [, ...] ] -> <type>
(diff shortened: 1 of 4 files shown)


docs/edgeql/ddl/links.rst
```diff
@@ -80,6 +82,11 @@ Parameters
     the *required* attribute, i.e it is not possible to make a
     required link non-required by extending it.
 
+:eql:synopsis:`OPTIONAL`
+    If specified, the link is considered *optional* for the parent
```
1st1, line 86: OPTIONAL is default, so it doesn't matter if it's specified or not. Please reflect that it's the default.

docs/edgeql/ddl/props.rst
```diff
@@ -78,6 +79,11 @@ Parameters
     inherit the *required* attribute, i.e it is not possible to make a
     required property non-required by extending it.
 
+:eql:synopsis:`OPTIONAL`
+    If specified, the property is considered *optional* for the parent
+    object type, i.e. it is possible for the property to resolve to an
+    empty value.
```
1st1, line 85: ditto

docs/edgeql/ddl/links.rst
vpetrovykh: Updated to clarify it's the default.

docs/edgeql/ddl/props.rst
vpetrovykh: OK

PR #1541, opened 2020-07-14
Fix restore all

tailhook:
This does two things:
1. It doesn't create a database+user named after unix system user. I don't think think makes sense now.
2. Assumes that `edgedb` user is created (and has no password or extending) in `describe roles`, so it alters it

The (1) fixes issue of database+user being created when we're trying to `restore --all` into that database (so we don't conflict on `CREATE ROLE`, note: we can't know which role will be on target system when generating dump, so can't use alter or drop role statements to fix that).

(2) fixes just `edgedb` since it's always created. I still don't like assymetry of `alter role` for `edgedb` vs `create role` for others in `describe roles` (it's fine in dump, but it kinda weird in statement which is meant to "describe").

### DIFF
--- a/edb/pgsql/metaschema.py
+++ b/edb/pgsql/metaschema.py
@@ -1980,6 +1980,7 @@ class DescribeRolesAsDDLFunction(dbops.Function):
         roles = inhviewname(schema, role_obj)
         member_of = role_obj.get_pointers(schema).get(schema, 'member_of')
         members = inhviewname(schema, member_of)
+        qi_superuser = qlquote.quote_ident(defines.EDGEDB_SUPERUSER)
         text = f"""
             WITH RECURSIVE
             dependencies AS (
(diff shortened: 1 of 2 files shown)


elprans:
`edgedb-server` has the `--default-database-user` and `--default-database`, which could be both set to `edgedb` to inhibit the "current user+db" behavior at bootstrap.  I don't think we should remove this altogether, because the drivers default to the current user for auth and this would make manually-created user instances a bit more awkward to use.

tailhook:
> `edgedb-server` has the `--default-database-user` and `--default-database`, which could be both set to `edgedb` to inhibit the "current user+db" behavior at bootstrap. I don't think we should remove this altogether, because the drivers default to the current user for auth and this would make manually-created user instances a bit more awkward to use.

So this means that all service files must contain `--default-database-user` and `--default-database`, right?

I'm still not sure this is the best way to handle that. Because we have to explain that to everyone trying to do `restore --all` to a manually created instance.

What about other part of the PR. I.e. `ALTER ROLE edgedb` thing?

elprans:
> So this means that all service files must contain `--default-database-user` and `--default-database`, right?

It's only relevant for instance bootstrap.  I actually think that service files shouldn't attempt to bootstrap the datadir implicitly (i.e need to add the `--no-bootstrap` flag or something).

> I'm still not sure this is the best way to handle that. Because we have to explain that to everyone trying to do `restore --all` to a manually created instance.

It's a tricky thing, but perhaps the `CREATE OR REPLACE` syntax would help.

> What about other part of the PR. I.e. `ALTER ROLE` edgedb thing?

That one is fine.

PR #1543, opened 2020-07-14
Add DDL syntax to change whether a referenced schema object is "owned"

elprans:
Links, properties, constraints, indexes etc are either inherited from
the parent type(s), or are defined directly in a given type, in which
case the referenced object is "owned" by the type.  Currently, there is
no way to consistently express the change from "inherited" to "owned" or
vice-versa in DDL, so add the new `SET OWNED` and `DROP OWNED` clauses
to the appropriate `ALTER` commands.  This also renames the `is_local`
object field to `is_owned` everywhere for consistency.


edb/edgeql-parser/src/tokenizer.rs
```diff
@@ -809,6 +809,7 @@ pub fn is_keyword(s: &str) -> bool {
         | "__subject__"
         | "__type__"
         | "__std__"
+        | "abort"
```
1st1, line 812: This seem to be unrelated to the PR

edb/edgeql-parser/src/tokenizer.rs
```diff
@@ -844,6 +845,7 @@ pub fn is_keyword(s: &str) -> bool {
         | "optional"
         | "or"
         | "order"
+        | "populate"
```
1st1, line 848: ditto

edb/edgeql-parser/src/tokenizer.rs
elprans: I synced the current list of keywords (forgot to in the original migration pr).

PR #1553, opened 2020-07-17
Update createdb.rst

cosmophile:
SDL commit missing final COMMIT command.

### DIFF
--- a/docs/tutorial/createdb.rst
+++ b/docs/tutorial/createdb.rst
@@ -77,6 +77,8 @@ Migrations have to be done inside a :ref:`transaction
     POPULATE MIGRATION
     tutorial> COMMIT MIGRATION;
     COMMIT MIGRATION
+    tutorial> COMMIT;
+    COMMIT
 
 The name of a migration doesn't matter much beyond providing a way to
 specify the particular migration which must be committed. Once the


elprans:
Hi @cosmophile!  We have changed how migrations work, and in a4+ they no longer require an explicit transaction block, because `START MIGRATION` and `COMMIT MIGRATION` form a transaction block implicitly now.  The sentence preceding the example you've modified should actually be removed, but the example itself is correct.

cosmophile:
Okay I see! Then I will open an edgedb-js issue, because the following did not work for me to commit a schema to a new database (I'm basically testing a js-only workflow for my db): 

```
    await conn.execute('START TRANSACTION')
    await conn.execute(`START MIGRATION TO {${schema}}`);
    await conn.execute('POPULATE MIGRATION')
    await conn.execute('COMMIT MIGRATION')
```

Whereas the following worked:
```
    await conn.execute('START TRANSACTION')
    await conn.execute(`START MIGRATION TO {${schema}}`);
    await conn.execute('POPULATE MIGRATION')
    await conn.execute('COMMIT MIGRATION')
    await conn.execute('COMMIT')
```

elprans:
@cosmophile It didn't work because you started a transaction and didn't commit it.  Remove `START TRANSACTION` and it'll work.

cosmophile:
Doh.

PR #1554, opened 2020-07-17
Upgrade rust to 1.45

tailhook:
Needed for: https://github.com/edgedb/edgedb-cli/pull/102

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -61,6 +61,13 @@ jobs:
         sudo apt-get update
         sudo apt-get install -y uuid-dev libreadline-dev bison flex
 
+    - name: Install rust toolchain
+      uses: actions-rs/toolchain@v1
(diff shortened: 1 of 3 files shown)


tailhook:
Do we need to update it somewhere in the build infrastructure?

elprans:
> Do we need to update it somewhere in the build infrastructure?

Yes, [here](https://github.com/edgedb/edgedb-pkg/blob/master/integration/linux/build/update.sh#L65)

elprans:
Also, we probably want to use `https://github.com/actions-rs/toolchain` explicitly and not rely on the toolchain installed in the actions image.

tailhook:
> Also, we probably want to use `https://github.com/actions-rs/toolchain` explicitly and not rely on the toolchain installed in the actions image.

Oh. Didn't know we rely on that :(

PR #1555, opened 2020-07-17
Implement casts between enums and json

msullivan:
Part of #251.

### DIFF
--- a/edb/edgeql/compiler/casts.py
+++ b/edb/edgeql/compiler/casts.py
@@ -128,7 +128,7 @@ def compile_cast(
                 subctx.implicit_tid_in_shapes = False
                 viewgen.compile_view_shapes(ir_set, ctx=subctx)
         elif (orig_stype.issubclass(ctx.env.schema, json_t)
-              and new_stype.is_enum(ctx.env.schema) and False):
+              and new_stype.is_enum(ctx.env.schema)):
             # Casts from json to enums need some special handling
             # here, where we have access to the enum type. Just turn
             # it into json->str and str->enum.
--- a/edb/edgeql/compiler/casts.py
+++ b/edb/edgeql/compiler/casts.py
@@ -127,6 +127,15 @@ def compile_cast(
                 subctx.implicit_id_in_shapes = False
                 subctx.implicit_tid_in_shapes = False
                 viewgen.compile_view_shapes(ir_set, ctx=subctx)
+        elif (orig_stype.issubclass(ctx.env.schema, json_t)
+              and new_stype.is_enum(ctx.env.schema)):
+            # Casts from json to enums need some special handling
+            # here, where we have access to the enum type. Just turn
+            # it into json->str and str->enum.
+            str_typ = ctx.env.get_track_schema_object('std::str')
+            assert isinstance(str_typ, s_types.Type)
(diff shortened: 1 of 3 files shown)


edb/edgeql/compiler/casts.py
```diff
@@ -127,6 +127,15 @@ def compile_cast(
                 subctx.implicit_id_in_shapes = False
                 subctx.implicit_tid_in_shapes = False
                 viewgen.compile_view_shapes(ir_set, ctx=subctx)
+        elif (orig_stype.issubclass(ctx.env.schema, json_t)
+              and new_stype.is_enum(ctx.env.schema)):
+            # Casts from json to enums need some special handling
+            # here, where we have access to the enum type. Just turn
+            # it into json->str and str->enum.
+            str_typ = ctx.env.get_track_schema_object('std::str')
```
elprans, line 135: Use `get_track_schema_type` and then you won't need the `assert`

edb/lib/std/25-enumoperators.edgeql
```diff
@@ -105,3 +105,9 @@ CREATE CAST FROM std::anyenum TO std::str {
     SET volatility := 'IMMUTABLE';
     USING SQL CAST;
 };
+
+# The reverse cast from json to enum is handled in compile_cast
```
elprans, line 109: I think we still want an explicit declaration of the reverse cast. Use `USING SQL EXPRESSION` to mark it as "compiler-handled".

tests/test_edgeql_enums.py
```diff
@@ -188,3 +188,27 @@ async def test_edgeql_enums_assignment_04(self):
...
+    async def test_edgeql_enums_json_cast_01(self):
+        self.assertEqual(
+            await self.con.query(
+                "SELECT <json><test::color_enum_t>'RED'"
+            ),
+            ['"RED"'])
+
+        await self.assert_query_result(
+            "SELECT <test::color_enum_t><json>'RED'",
+            ['RED'])
```
1st1, line 201: `self.assert_query_result` checks both JSON and binary compile modes automatically. SO I'd also add a test without an explicit `<json>` cast, testing that JSON mode handles enums just fine too:

```
            "SELECT <test::color_enum_t>'RED'",
            ['RED'])
```

msullivan: I can add that test, but it looks like it already passes even without my changes

edb/edgeql/compiler/casts.py
```diff
@@ -128,12 +128,11 @@ def compile_cast(
                 subctx.implicit_tid_in_shapes = False
                 viewgen.compile_view_shapes(ir_set, ctx=subctx)
         elif (orig_stype.issubclass(ctx.env.schema, json_t)
-              and new_stype.is_enum(ctx.env.schema)):
+              and new_stype.is_enum(ctx.env.schema) and False):
```
elprans, line 131:
```suggestion
              and new_stype.is_enum(ctx.env.schema)):
```

msullivan: oops! debug code slipped in the last update

1st1: Was it caught by tests?

elprans: Yes, tests failed.

PR #1563, opened 2020-07-20
Implement casts from json to array types

msullivan:
Part of #251.

### DIFF
--- a/edb/edgeql/compiler/casts.py
+++ b/edb/edgeql/compiler/casts.py
@@ -141,8 +141,8 @@ def compile_cast(
                   ctx.env.schema, json_t)):
             # Turn casts from json->array<T> into json->array<json>
             # and array<json>->array<T>.
-            _, json_array_typ = s_types.Array.create(
-                ctx.env.schema, element_type=json_t)
+            ctx.env.schema, json_array_typ = s_types.Array.from_subtypes(
+                ctx.env.schema, [json_t])
             json_array_ir = compile_cast(
                 ir_expr, json_array_typ, srcctx=srcctx, ctx=ctx)
             return compile_cast(
--- a/edb/edgeql/compiler/casts.py
+++ b/edb/edgeql/compiler/casts.py
@@ -56,7 +56,8 @@ def compile_cast(
         new_stype: s_types.Type, *,
         srcctx: Optional[parsing.ParserContext],
         ctx: context.ContextLevel,
-        modifier: Optional[qlast.CardinalityModifier]=None) -> irast.Set:
+        cardinality_mod: Optional[qlast.CardinalityModifier]=None
+) -> irast.Set:
 
     if isinstance(ir_expr, irast.EmptySet):
         # For the common case of casting an empty set, we simply
@@ -131,7 +132,8 @@ def compile_cast(
             # it into json->str and str->enum.
             str_typ = ctx.env.get_track_schema_type('std::str')
             str_ir = compile_cast(ir_expr, str_typ, srcctx=srcctx, ctx=ctx)
-            return compile_cast(str_ir, new_stype, modifier=modifier,
+            return compile_cast(str_ir, new_stype,
+                                cardinality_mod=cardinality_mod,
                                 srcctx=srcctx, ctx=ctx)
         elif (orig_stype.issubclass(ctx.env.schema, json_t)
               and isinstance(new_stype, s_types.Array)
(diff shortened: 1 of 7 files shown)


tests/test_edgeql_casts.py
```diff
@@ -2162,6 +2162,32 @@ async def test_edgeql_casts_json_10(self):
...
 
+    async def test_edgeql_casts_json_11(self):
+        await self.assert_query_result(
+            r"SELECT <array<int64>><json>[1, 1, 2, 3, 5]",
+            [[1, 1, 2, 3, 5]]
+        )
+
+        with self.assertRaisesRegex(
+                # FIXME: This should be a different error
+                edgedb.InternalServerError,
```
1st1, line 2173: The simplest way would be to map these via `errormech.py` to `errors.InvalidValueError`

edb/edgeql/compiler/casts.py
```diff
@@ -135,6 +135,18 @@ def compile_cast(
             str_typ = ctx.env.get_track_schema_type('std::str')
             str_ir = compile_cast(ir_expr, str_typ, srcctx=srcctx, ctx=ctx)
             return compile_cast(str_ir, new_stype, srcctx=srcctx, ctx=ctx)
+        elif (orig_stype.issubclass(ctx.env.schema, json_t)
+              and isinstance(new_stype, s_types.Array)
+              and not new_stype.get_subtypes(ctx.env.schema)[0].issubclass(
+                  ctx.env.schema, json_t)):
+            # Turn casts from json->array<T> into json->array<json>
+            # and array<json>->array<T>.
+            _, json_array_typ = s_types.Array.create(
```
1st1, line 144:
```suggestion
            ctx.env.schema, json_array_typ = s_types.Array.create(
```

elprans: Also, use `from_subtypes`, which wraps `create` with extra sanity checks.

tests/test_edgeql_casts.py
msullivan: OK. I'll do a follow-up PR that improves a bunch of JSON related errors.

elprans: Yes, we need to map `wrong_object_type` (42809) to `InvalidValueError`.  That can be done in a separate PR, though.

msullivan:
Oh, this does the wrong thing when there are null elements in the array. (Which seems to be inherited from existing behavior casting from array<json> to array<something else>). I'll look into this

edb/pgsql/compiler/expr.py
```diff
@@ -232,14 +232,31 @@ def compile_TypeCast(
...
-        return pgast.FuncCall(
+        res = pgast.FuncCall(
             name=func_name,
             args=[pg_expr],
         )
 
     else:
         raise RuntimeError('cast not supported')
 
+    if expr.modifier and expr.modifier == expr.modifier.Required:
```
elprans, line 243:
```suggestion
    if expr.modifier is expr.modifier.Required:
```

edb/edgeql/compiler/casts.py
```diff
@@ -55,7 +55,8 @@ def compile_cast(
         ir_expr: Union[irast.Set, irast.Expr],
         new_stype: s_types.Type, *,
         srcctx: Optional[parsing.ParserContext],
-        ctx: context.ContextLevel) -> irast.Set:
+        ctx: context.ContextLevel,
+        modifier: Optional[qlast.CardinalityModifier]=None) -> irast.Set:
```
elprans, line 59: Lets use the opportunity to rename `modifier` to something more descriptive, like `cardinality_mod`.

tests/test_edgeql_casts.py
```diff
@@ -2162,6 +2162,50 @@ async def test_edgeql_casts_json_10(self):
...
+                r'expected json number, null; got json string'):
+            await self.con.query_one(
+                r"SELECT <array<int64>><json>['asdf']")
+
+        with self.assertRaisesRegex(
+                # FIXME: This should be a different error
+                edgedb.InternalServerError,
+                r'expected json number, null; got json string'):
```
elprans, line 2181: The error message is somewhat misleading. I wonder if we can simply avoid mentioning `null` in `jsonb_assert_type`.

elprans: This is material for a followup PR, though.

Issue #1350, opened 2020-04-18
Add __tname__ object magic field & session config to enable it


1st1:
For REPL & the tutorial it's a great UX to display type names when we render results. Introspecting schema is a suboptimal solution as it's slow and not always practical to do.

tailhook:
I also think that we may want to configure all implicit fields rather than just `__tname__`. I.e. `__tid__` is not used in repl now, unless you enable `introspect-types` which should rely on the new `__tname__` instead. And `id` is also not needed anywhere except repl (for empty shapes).

I'm not sure this should be a session config, rather than per-request config, like we do `implicit_limit` though (I'd like to disable it for statically typed queries like `list-databases` in CLI).

fmoor:
Also, I don't think clients for statically typed languages like go will have much use for __tid__, __tname__ etc. since query results  will be marshaled into user defined types.

elprans:
@1st1 is this done?

1st1:
I think so

PR #1567, opened 2020-07-22
Add a new header to the Prepare message to implicitly inline type names

1st1:
The Prepare protocol message now has a new header to make the compiler
implicitly inline a "__tname__ := .__type__.name" computable.  This is
especially useful for REPL-like applications.

### DIFF
--- a/edb/edgeql/compiler/viewgen.py
+++ b/edb/edgeql/compiler/viewgen.py
@@ -55,6 +55,8 @@ if TYPE_CHECKING:
     from edb.schema import lproperties as s_props
     from edb.schema import sources as s_sources
 
+    ShapePtr = Tuple[irast.Set, s_pointers.Pointer, qlast.ShapeOp]
+
 
 def process_view(
     *,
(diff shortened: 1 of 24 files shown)


1st1:
Issue: https://github.com/edgedb/edgedb/issues/1350

edb/edgeql/compiler/viewgen.py
```diff
@@ -1005,13 +1007,84 @@ def has_implicit_tid(
     )
 
 
+def has_implicit_tname(
+        stype: s_types.Type, *,
```
elprans, line 1011: Let's use consistent formatting for newly added code.

elprans (approved):

1st1:
@elprans @tailhook I'd like to merge this asap to land the edgecon.pyx refactorings and unblock my ports rework. The REPL works with this PR as is, although it crashes when `\set introspect-types on` is set (https://github.com/edgedb/edgedb-cli/pull/108 addresses that).

elprans (approved):

PR #1570, opened 2020-07-22
Improve RENAME DDL reconstruction from schema diff

elprans:
This makes it so that parent object renames are processed before child
object alteration is considered, which correctly skips delta hunks that
resulted in rename propagation, as well as reassembles the delta tree in
a way that results in only a single DDL command.

### DIFF
--- a/edb/schema/constraints.py
+++ b/edb/schema/constraints.py
@@ -58,9 +58,36 @@ def _assert_not_none(value: Optional[T]) -> T:
     return value
 
 
+def merge_constraint_params(
+    constraint: Constraint,
+    supers: List[Constraint],
+    field_name: str,
+    *,
+    ignore_local: bool,
+    schema: s_schema.Schema,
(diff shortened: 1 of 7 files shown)


edb/schema/constraints.py
```diff
@@ -319,12 +319,12 @@ def _validate_subcommands(
...
-            if cname in {'subject', 'subjectexpr'}:
-                raise errors.InvalidConstraintDefinitionError(
-                    f'{cname} is not a valid constraint annotation',
-                    context=command.context)
+            if isinstance(command, qlast.BaseSetField):
+                cname = command.name
+                if cname in {'subject', 'subjectexpr'}:
+                    raise errors.InvalidConstraintDefinitionError(
+                        f'{cname} is not a valid constraint annotation',
+                        context=command.context)
```
1st1, line 327: I don't see a test for this. Can we add one?

PR #1572, opened 2020-07-22
Implement casts from json to tuple types

msullivan:
The strategy I used was to do it at the IR level by making calls to
`json_get`.

Finishes #251, I think.

### DIFF
--- a/edb/edgeql/compiler/casts.py
+++ b/edb/edgeql/compiler/casts.py
@@ -104,13 +104,15 @@ def compile_cast(
         # and is always a wider domain, so we simply reassign
         # the stype.
         return _inheritance_cast_to_ir(
-            ir_set, orig_stype, new_stype, cardinality_mod, ctx=ctx)
+            ir_set, orig_stype, new_stype,
+            cardinality_mod=cardinality_mod, ctx=ctx)
 
     elif new_stype.issubclass(ctx.env.schema, orig_stype):
         # The new type is a subtype, so may potentially have
         # a more restrictive domain, generate a cast call.
         return _inheritance_cast_to_ir(
-            ir_set, orig_stype, new_stype, cardinality_mod, ctx=ctx)
+            ir_set, orig_stype, new_stype,
+            cardinality_mod=cardinality_mod, ctx=ctx)
 
     elif orig_stype.is_array():
         return _cast_array(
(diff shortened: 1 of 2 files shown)


tests/test_edgeql_casts.py
```diff
@@ -2204,6 +2204,112 @@ async def test_edgeql_casts_json_11(self):
...
+                r"""SELECT <tuple<a: int64, b: int64>> to_json('"test"')"""
+            )
+
+        with self.assertRaisesRegex(
+                edgedb.InvalidValueError,
+                r'invalid null value in cast'):
+            await self.con.query(
+                r"""SELECT <tuple<json, json>> to_json('[3000]')"""
+            )
+
```
1st1, line 2312: Would a `<tuple<tuple<int64, str>, tuple<str, int64>> some_json` work (nested tuples)? What about `<tuple<array<tuple ... >>>` types?

edb/edgeql/compiler/casts.py
```diff
@@ -207,7 +212,9 @@ def _cast_to_ir(
 def _inheritance_cast_to_ir(
         ir_set: irast.Set,
         orig_stype: s_types.Type,
-        new_stype: s_types.Type, *,
+        new_stype: s_types.Type,
+        cardinality_mod: Optional[qlast.CardinalityModifier],
```
1st1, line 216: Let's make `cardinality_mod` a keyword-only (like it is in `compile_cast()`)

edb/edgeql/compiler/casts.py
```diff
@@ -356,6 +364,46 @@ def _find_cast(
...
+
+    new_subtypes = dict(new_stype.iter_subtypes(ctx.env.schema))
+
+    with ctx.new() as subctx:
+        subctx.anchors = subctx.anchors.copy()
+        source_alias = subctx.aliases.get('a')
+        subctx.anchors[source_alias] = ir_set
+
+        elements = []
+        for new_el_name, new_st in new_subtypes.items():
```
1st1, line 382: Why not `for ... in new_stype.iter_subtypes(ctx.env.schema)`?  That would avoid allocating a dictionary just for iteration.

msullivan: Oh, yeah, obviously. (I had copy-pasted and adapted from some other tuple casting code that made a dict so it could ask the length, it looks like)

tests/test_edgeql_casts.py
msullivan: That ought to work. I'll add some tests.

tests/test_edgeql_casts.py
```diff
@@ -2204,6 +2204,135 @@ async def test_edgeql_casts_json_11(self):
...
+                r"""
+                    SELECT <tuple<a: int64, b: int64>>
+                    to_json('{"a": 1, "b": "2"}')
+                """
+            )
+
+        # This isn't really the best error message for this.
+        with self.assertRaisesRegex(
+                edgedb.InvalidValueError,
+                r'invalid null value in cast'):
```
1st1, line 2300: Hm, right. @elprans how can we have a better one here? Or maybe not bother?

PR #1574, opened 2020-07-23
Add explicit declarations for casts from json to collections

elprans:
This is to explicitly indicate support for the newly added `std::json`
casts.

### DIFF
--- a/edb/lib/std/30-jsonfuncs.edgeql
+++ b/edb/lib/std/30-jsonfuncs.edgeql
@@ -153,12 +153,24 @@ CREATE CAST FROM anytuple TO std::json {
 };
 
 
+CREATE CAST FROM std::json TO anytuple {
+    SET volatility := 'STABLE';
+    USING SQL EXPRESSION;
+};
+
+
 CREATE CAST FROM std::json TO array<json> {
     SET volatility := 'IMMUTABLE';
     USING SQL $$
         SELECT array_agg(j)
         FROM jsonb_array_elements(val) AS j
     $$;
+};
+
+
+CREATE CAST FROM std::json TO array<anytype> {
+    SET volatility := 'STABLE';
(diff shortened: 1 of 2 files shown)


edb/server/defines.py
```diff
@@ -27,7 +27,7 @@
 EDGEDB_VISIBLE_METADATA_PREFIX = r'EdgeDB metadata follows, do not modify.\n'
 
 # Increment this whenever the database layout or stdlib changes.
-EDGEDB_CATALOG_VERSION = 2020_07_14_14_00
+EDGEDB_CATALOG_VERSION = 2020_07_23_00_00
```
msullivan, line 30: Aha, I've missed doing this!

edb/lib/std/30-jsonfuncs.edgeql
```diff
@@ -153,6 +153,12 @@ CREATE CAST FROM anytuple TO std::json {
 };
 
 
+CREATE CAST FROM std::json TO anytuple {
+    SET volatility := 'STABLE';
+    USING SQL EXPRESSION;
+};
```
1st1, line 159: @elprans Why were the tests passing without this?

elprans: Because these are special casts (`FROM SQL EXPRESSION`), which are handled by the compiler directly and not via the generic dispatch.

1st1: I don't think this answers my question.

1st1: Is the answer that these formally defined casts are there just for the introspection?

elprans: Yes, for introspection and completeness sake.

PR #1575, opened 2020-07-23
Disallow using std::exclusive on scalar types

### LINKED ISSUE #1182: InternalServerError error creating scalar type
- EdgeDB Version: `1.0-alpha.2+dev.505.g5b68ec24`

Steps to Reproduce:

```
START TRANSACTION;
CREATE MIGRATION init TO {
    module default {
        scalar type unique_name extending str {
            constraint exclusive;
        }
    }
};
COMMIT MIGRATION init;
```
Error: `InternalServerError: column "78096142-4760-11ea-ad7d-c3dbbf5ced49" does not exist`

I found another InternalServerError while playing with this:
```
START TRANSACTION;
CREATE MIGRATION init TO {
    module default {
        scalar type unique_name extending str {
            constraint exclusive on (str_lower(__subject__));
        }
    }
};
COMMIT MIGRATION init;
```
Error: `InternalServerError: argument of CHECK must be type boolean, not type text`

I'm not sure if the latter is a bug, I'm using it wrong, but `InternalServerError` is scary.

Is it possible to declare `exclusive` constraint in scalar so that it checks lowercase value?
My plan to use it is:
```
type User {
    required property nickname -> unique_name;
    # ...
    index on (str_lower(__subject__.nickname));
}
```
`unique_name` has other constraints, but I removed them to simplify my report.

msullivan:
Fixes #1182.

### DIFF
--- a/edb/lib/std/50-constraints.edgeql
+++ b/edb/lib/std/50-constraints.edgeql
@@ -46,6 +46,7 @@ std::expression EXTENDING std::constraint
 CREATE ABSTRACT CONSTRAINT
 std::exclusive EXTENDING std::constraint
 {
+    SET is_aggregate := true;
     SET errmessage := '{__subject__} violates exclusivity constraint';
     USING (std::_is_exclusive(__subject__));
 };
--- a/edb/schema/constraints.py
+++ b/edb/schema/constraints.py
@@ -28,6 +28,7 @@ from edb.edgeql import compiler as qlcompiler
 from edb.edgeql import qltypes as ft
 from edb.edgeql import parser as qlparser
 from edb.edgeql import utils as qlutils
+from edb.schema import scalars as s_scalars
 
 from . import abc as s_abc
 from . import annos as s_anno
@@ -44,7 +45,6 @@ from . import utils
 
 if TYPE_CHECKING:
     from edb.common import parsing as c_parsing
-    from edb.schema import scalars as s_scalars
     from edb.schema import schema as s_schema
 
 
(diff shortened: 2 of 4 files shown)


edb/lib/std/50-constraints.edgeql
```diff
@@ -46,6 +46,7 @@ std::expression EXTENDING std::constraint
 CREATE ABSTRACT CONSTRAINT
 std::exclusive EXTENDING std::constraint
 {
+    SET is_aggregate := true;
```
1st1, line 49: We don't formally use "aggregate" in EdgeQL, we have `SET OF` for that. Maybe `is_setof` would work better here?

1st1: @elprans

elprans: Well, `SET OF` is the cardinality modifier on the argument/return type, but we _do_ call these "aggregate parameters" and "aggregate functions": https://edgedb.com/docs/edgeql/overview#aggregates.

edb/schema/constraints.py
```diff
@@ -733,6 +736,12 @@ def _populate_concrete_constraint_attrs(
                 f'subjectexpr is already defined for {name!r}'
             )
 
+        if (isinstance(subject_obj, s_scalars.ScalarType)
+                and constr_base.get_is_aggregate(schema)):
+            raise errors.InvalidConstraintDefinitionError(
+                f'Constraint {name} may not be used on scalar types'
```
elprans, line 742:
```suggestion
                f'{constr_base.get_verbosename(schema)} may not be used on scalar types'
```

edb/lib/std/50-constraints.edgeql
1st1: Alright.

edb/schema/constraints.py
elprans: Hm. Sorry, `abstract constraint 'std::exclusive'` is a tad misleading and irrelevant in this context, so let's do
`f'constraint {constr_base.get_displayname(schema)} may not ...` (to elide the word "abstract")

Issue #1164, opened 2020-01-28
Add constraint syntax to object types


elprans:
Currently we allow declaring constraints only on properties and links (syntactically), but not on object types.  This is largely an oversight, since we already have all the necessary mechanisms.

dlax:
Does this mean something like this?
```
type Person {
    required property first_name -> str;
    required property last_name -> str;
    constraint exclusive on ((__subject__.first_name, __subject__.last_name));
}
```

elprans:
> Does this mean something like this?

Yes, that's correct.

CodesInChaos:
The examples and tests all use `constraint exclusive`. Does this feature work with other constraints as well?

elprans:
> Does this feature work with other constraints as well?

Absolutely.  This works:

```
create type User {
    create property first_name -> str;
    create property last_name -> str;

    create constraint max_value(20)
        on (len(array_join([__subject__.first_name, __subject__.last_name], " ")))
    {
        set errmessage := 'combined length of first and last name must not exceed 20 characters'
    }
};
```

> The examples and tests all use `constraint exclusive`

Yes, we need more tests and examples: #1609.

PR #1581, opened 2020-07-24
Support adding constraints to object types in DDL

msullivan:
SDL support to come in a follow-up.

Work on #1164.

### DIFF
--- a/docs/edgeql/ddl/objects.rst
+++ b/docs/edgeql/ddl/objects.rst
@@ -221,7 +221,7 @@ The following subcommands are allowed in the ``ALTER TYPE`` block:
     :eql:stmt:`DROP PROPERTY` for details.
 
 :eql:synopsis:`ALTER CONSTRAINT <constraint-name> ...`
-    Alter the definition of a constraint for this link.  See
+    Alter the definition of a constraint for this object type.  See
     :eql:stmt:`ALTER CONSTRAINT` for details.
 
 :eql:synopsis:`DROP CONSTRAINT <constraint-name>;`
--- a/docs/edgeql/ddl/objects.rst
+++ b/docs/edgeql/ddl/objects.rst
@@ -90,7 +90,7 @@ The following subcommands are allowed in the ``CREATE TYPE`` block:
     :eql:stmt:`CREATE PROPERTY` for details.
 
 :eql:synopsis:`CREATE CONSTRAINT <constraint-name> ...`
-    Define a concrete constraint for this link.  See
+    Define a concrete constraint for this object type.  See
     :eql:stmt:`CREATE CONSTRAINT` for details.
 
 :eql:synopsis:`CREATE INDEX ON <index-expr>`
--- a/docs/edgeql/ddl/objects.rst
+++ b/docs/edgeql/ddl/objects.rst
@@ -28,6 +28,7 @@ CREATE TYPE
       CREATE ANNOTATION <annotation-name> := <value>
       CREATE LINK <link-name> ...
       CREATE PROPERTY <property-name> ...
+      CREATE CONSTRAINT <constraint-name> ...
       CREATE INDEX ON <index-expr>
 
 Description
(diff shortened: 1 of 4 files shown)


docs/edgeql/ddl/objects.rst
```diff
@@ -88,6 +89,10 @@ The following subcommands are allowed in the ``CREATE TYPE`` block:
     Define a new property for this object type.  See
     :eql:stmt:`CREATE PROPERTY` for details.
 
+:eql:synopsis:`CREATE CONSTRAINT <constraint-name> ...`
+    Define a concrete constraint for this link.  See
```
elprans, line 93:
```suggestion
    Define a concrete constraint for this object type.  See
```

docs/edgeql/ddl/objects.rst
```diff
@@ -212,6 +220,14 @@ The following subcommands are allowed in the ``ALTER TYPE`` block:
     Remove a property item from this object type.  See
     :eql:stmt:`DROP PROPERTY` for details.
 
+:eql:synopsis:`ALTER CONSTRAINT <constraint-name> ...`
+    Alter the definition of a constraint for this link.  See
```
elprans, line 224:
```suggestion
    Alter the definition of a constraint for this object type.  See
```

docs/edgeql/ddl/objects.rst
```diff
@@ -212,6 +220,14 @@ The following subcommands are allowed in the ``ALTER TYPE`` block:
...
+    :eql:stmt:`ALTER CONSTRAINT` for details.
+
+:eql:synopsis:`DROP CONSTRAINT <constraint-name>;`
+    Remove a constraint from this link.  See
```
elprans, line 228:
```suggestion
    Remove a constraint from this object type.  See
```

tests/test_constraints.py
```diff
@@ -933,6 +933,61 @@ async def test_constraints_ddl_06(self):
                 };
             """)
 
+    async def test_constraints_ddl_07(self):
+        await self.con.execute("""
+            SET MODULE test;
+            CREATE TYPE Test {
+                CREATE PROPERTY first_name -> str;
+                CREATE PROPERTY last_name -> str;
+                CREATE CONSTRAINT exclusive on (__subject__.first_name);
```
elprans, line 942: Make sure that only `SINGLE` pointers are allowed, and test constraints on links too.

elprans: Per chat, we'll allow `MULTI` if it's only one ref in expression, so that it's completely equivalent to a constraint defined directly on a pointer.

edb/schema/constraints.py
```diff
@@ -825,6 +827,27 @@ def _populate_concrete_constraint_attrs(
...
+                ),
+            )
+            assert isinstance(final_subjectexpr.irast, ir_ast.Statement)
+
+            if final_subjectexpr.irast.cardinality.is_multi():
+                refs = ir_utils.get_longest_paths(final_expr.irast)
+                if len(refs) > 1:
+                    raise errors.InvalidConstraintDefinitionError(
+                        "Constraint with multi cardinality may not "
+                        "reference multiple fields",
```
elprans, line 847: We use "pointers" instead of "fields" (and that's how the internal abstraction is called).  That said, we also go out of our way to _not_ use that term in the documentation and instead use "link or property". @edgedb/engineering, interested to hear opinions on whether we want a single term to mean "link or property", and what should it be.

1st1: "Link or property" is fine. Single terms tend to be confusing if they're not obvious. Ideally we should use either "link" or "property" in error messages when we know which one is more relevant.

tests/test_constraints.py
```diff
@@ -933,6 +933,108 @@ async def test_constraints_ddl_06(self):
...
+            await self.con.execute("""
+                CREATE TYPE test::Label {
+                    CREATE PROPERTY text -> str;
+                };
+                CREATE TYPE test::ObjCnstr3 {
+                    CREATE LINK label -> test::Label;
+                    CREATE CONSTRAINT exclusive on (__subject__.label);
+                };
+                INSERT test::ObjCnstr3 {
+                    label := (SELECT (INSERT test::Label {
```
elprans, line 1024: No need to wrap `INSERT` with `SELECT` here.

PR #1592, opened 2020-07-25
Fix handling of literal empty sets of object types

elprans:
Literal empty set of object type (i.e. `<Foo>{}`) is currently
erroneously returning an entire set of `Foo` instead of an empty set.

Fixes: #1589

### DIFF
--- a/tests/test_edgeql_select.py
+++ b/tests/test_edgeql_select.py
@@ -3797,6 +3797,42 @@ class TestEdgeQLSelect(tb.QueryTestCase):
                 } ORDER BY .number;
                 """)
 
+    async def test_edgeql_select_empty_object_01(self):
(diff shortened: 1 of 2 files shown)


tests/test_edgeql_select.py
```diff
@@ -3797,6 +3797,33 @@ async def test_edgeql_select_empty_05(self):
...
                 """)
 
+    async def test_edgeql_select_empty_object_01(self):
+        await self.assert_query_result(
+            r'''
+            WITH MODULE test
+            SELECT <Issue>{}
+            ''',
+            [],
+        )
```
1st1, line 3807: While it would be equivalent, can you also add a test for `SELECT count(<Issue>{}) = 0`?

PR #1599, opened 2020-07-27
Support adding constraints to object types in SDL

### LINKED ISSUE #1164: Add constraint syntax to object types

msullivan:
Fixes #1164.

Depends on https://github.com/edgedb/edgedb/pull/1581.

### DIFF
--- a/tests/schemas/constraints.esdl
+++ b/tests/schemas/constraints.esdl
@@ -112,7 +112,9 @@ type Object {
 type ObjCnstr {
     required property first_name -> str;
     required property last_name -> str;
+    link label -> Label;
     constraint exclusive on (__subject__.first_name);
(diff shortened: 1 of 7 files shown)


tests/schemas/constraints.esdl
```diff
@@ -109,6 +109,12 @@ type Object {
     property c_my_enum -> constraint_my_enum;
 }
 
+type ObjCnstr {
+    required property first_name -> str;
+    required property last_name -> str;
+    constraint exclusive on (__subject__.first_name);
```
elprans, line 115: Let's add a test for constraints on links too.

PR #1602, opened 2020-07-28
Disallow empty set literals in FOR ... UNION statements

### LINKED ISSUE #1585: InternalServerError: local variable 'derived' referenced before assignment

This code causes error:
```edgeql
START TRANSACTION;
START MIGRATION TO {
    module default {
        function create_user() -> str
        {
            using EdgeQL $$
                FOR x IN {}
                UNION ()
            $$;
        };
    };
};
```

This seems related to `FOR x IN {}`.
Adding code to UNION block does not change anything, replacing `{}` with value removes error.

msullivan:
They don't do anything useful and some internals can't handle them.
Fixes #1585.

### DIFF
--- a/edb/edgeql/parser/grammar/expressions.py
+++ b/edb/edgeql/parser/grammar/expressions.py
@@ -112,7 +112,7 @@ class ByExprList(ListNonterm, element=ByExpr, separator=tokens.T_COMMA):
 
 class SimpleFor(Nonterm):
     def reduce_For(self, *kids):
-        r"%reduce FOR Identifier IN Set \
+        r"%reduce FOR Identifier IN NonEmptySet \
                   UNION OptionallyAliasedExpr"
         self.val = qlast.ForQuery(
             iterator_alias=kids[1].val,
@@ -956,6 +956,11 @@ class NamedTupleElementList(ListNonterm, element=NamedTupleElement,
     pass
 
 
+class NonEmptySet(Nonterm):
+    def reduce_LBRACE_TrailingCommaExprList_RBRACE(self, *kids):
+        self.val = qlast.Set(elements=kids[1].val)
+
+
 class Set(Nonterm):
     def reduce_LBRACE_OptExprList_RBRACE(self, *kids):
         self.val = qlast.Set(elements=kids[1].val)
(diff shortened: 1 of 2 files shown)


tests/test_edgeql_syntax.py
```diff
@@ -2609,6 +2609,14 @@ def test_edgeql_syntax_updatefor_01(self):
         UNION (UPDATE Foo FILTER (Foo.id = x.0) SET {bar := x.1});
         """
 
+    @tb.must_fail(errors.EdgeQLSyntaxError,
+                  r"Unexpected '}'",
+                  line=2, col=27)
+    def test_edgeql_syntax_empty_for_01(self):
+        """
+        SELECT (FOR x in {} UNION ());
+        """
```
1st1, line 2618: This bans a simple case, what about

```
WITH set := <schema::Object>{}
SELECT (FOR x in set UNION ());
```

?

Wouldn't it be better to fix the underlying problem instead of banning this at the grammar level?

cc @elprans

edb/edgeql/parser/grammar/expressions.py
```diff
@@ -956,6 +956,11 @@ class NamedTupleElementList(ListNonterm, element=NamedTupleElement,
     pass
 
 
+class NonEmptySet(Nonterm):
```
1st1, line 959: -1 on doing it this way. 

An `Unexpected {}` error would be confusing to users.  Instead this should be handled directly in `Set.reduce_LBRACE_OptExprList_RBRACE` with an `if not elements: raise SyntaxError("nice error message, with a suggestion")`

tests/test_edgeql_syntax.py
msullivan: @elplans suggested the syntactic approach because your snippet isn't valid syntax: the set in a FOR needs to be a set-literal.

That said, `WITH s := {} SELECT (FOR x in {s} UNION ());` triggers the same crash, so I'm going to abandon the syntactic approach.

elprans: It was my suggestion to ban this at the syntax level because `FOR` requires a set literal syntactically (otherwise we have ambiguous grammar).  The underlying problem is that the variable you iterate over in `FOR` must have a concrete type and an empty set literal `{}` is of type `anytype`.  Since you can't put a cast _before_ the empty set literal in `FOR` it makes sense to just make it a syntax error.

That said, we still need to add a check that the `FOR` variable is not of a generic type (your example shows how one can still trigger the bug).

1st1: @msullivan 
> the set in a FOR needs to be a set-literal.

Right!

@elprans 
> It was my suggestion to ban this at the syntax level because FOR requires a set literal syntactically (otherwise we have ambiguous grammar). 

Alright. I guess we can resurrect this PR with the approach I suggested in https://github.com/edgedb/edgedb/pull/1602#pullrequestreview-456944615

But the `WITH s := {} SELECT (FOR x in {s} UNION ());` case should also be fixed, ideally, so maybe we try to fix both things in one place.

PR #1604, opened 2020-07-28
Don't crash on empty set as FOR iterator.

### LINKED ISSUE #1585: InternalServerError: local variable 'derived' referenced before assignment

msullivan:
This catches the issue and reports an error in the compiler.

Fixes #1585.

I figure there's no compelling reason to also catch the special
syntactic case in the parser, given that we need to handle it in a
more general way in the compiler anyway?

### DIFF
--- a/edb/edgeql/compiler/schemactx.py
+++ b/edb/edgeql/compiler/schemactx.py
@@ -214,7 +214,7 @@ def derive_view(
                         ptrcls=ptr, ctx=ctx)
 
     else:
-        raise RuntimeError("unsupported type in derive_view")
+        raise TypeError("unsupported type in derive_view")
 
     ctx.view_nodes[derived.get_name(ctx.env.schema)] = derived
 
--- a/edb/edgeql/compiler/schemactx.py
+++ b/edb/edgeql/compiler/schemactx.py
@@ -213,6 +213,9 @@ def derive_view(
                     stmtctx.pend_pointer_cardinality_inference(
                         ptrcls=ptr, ctx=ctx)
 
+    else:
+        raise RuntimeError("unsupported type in derive_view")
+
     ctx.view_nodes[derived.get_name(ctx.env.schema)] = derived
 
     if preserve_shape and stype in ctx.env.view_shapes:
--- a/edb/edgeql/compiler/stmt.py
+++ b/edb/edgeql/compiler/stmt.py
@@ -132,8 +132,8 @@ def compile_ForQuery(
             iterator_stmt = setgen.new_set_from_set(
                 iterator_view, preserve_scope_ns=True, ctx=scopectx)
 
-            ptr_target = inference.infer_type(iterator_stmt, ctx.env)
-            anytype = ptr_target.find_any(ctx.env.schema)
+            iterator_type = inference.infer_type(iterator_stmt, ctx.env)
+            anytype = iterator_type.find_any(ctx.env.schema)
             if anytype is not None:
                 raise errors.QueryError(
                     'FOR statement has iterator of indeterminate type',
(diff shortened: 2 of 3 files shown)


edb/edgeql/compiler/schemactx.py
```diff
@@ -213,6 +213,9 @@ def derive_view(
                     stmtctx.pend_pointer_cardinality_inference(
                         ptrcls=ptr, ctx=ctx)
 
+    else:
+        raise RuntimeError("unsupported type in derive_view")
```
1st1, line 217: `TypeError`?

It would be propagated as an `InternalServerError` anyways, but `TypeError` looks like a better choice here.

edb/edgeql/compiler/stmt.py
```diff
@@ -131,6 +132,14 @@ def compile_ForQuery(
             iterator_stmt = setgen.new_set_from_set(
                 iterator_view, preserve_scope_ns=True, ctx=scopectx)
 
+            ptr_target = inference.infer_type(iterator_stmt, ctx.env)
+            anytype = ptr_target.find_any(ctx.env.schema)
+            if anytype is not None:
```
1st1, line 137: Would `ptr_target.is_polymorphic()` work here?

Speaking of `edb/schema/types.py`, `Type.contains_any()` implementation is off w.r.t. its return type.

edb/edgeql/compiler/schemactx.py
elprans: I think we should just standardize on something like an `AssertionError` to stop wondering about which exception to use in case of an unexpected state on the insides.  I picked `AssertionError` because we also have lots of `assert` around.

1st1: For cases like this I'd suggest something like `UnreachableError`.

1st1: Or even `Never(BaseException)`

edb/edgeql/compiler/stmt.py
elprans: > Would `ptr_target.is_polymorphic()` work here?

It would, but `find_any` gives you the benefit of providing the source context.

> `Type.contains_any()` implementation is off w.r.t. its return type.

How is it wrong? Seems OK to me.

1st1: > It would, but find_any gives you the benefit of providing the source context.

Got it.

> How is it wrong? Seems OK to me.

My bad, I thought it calls `find_any` instead of `is_any` for some reason.

edb/edgeql/compiler/schemactx.py
msullivan: `AssertionError` seems best to me if we want to standarize, since a lot of such errors are going to be assertion rerors

edb/edgeql/compiler/stmt.py
msullivan: Though we could just use the type itself without any trouble in the cases that seem likely here

edb/edgeql/compiler/stmt.py
```diff
@@ -132,8 +132,8 @@ def compile_ForQuery(
             iterator_stmt = setgen.new_set_from_set(
                 iterator_view, preserve_scope_ns=True, ctx=scopectx)
 
-            ptr_target = inference.infer_type(iterator_stmt, ctx.env)
-            anytype = ptr_target.find_any(ctx.env.schema)
+            iterator_type = inference.infer_type(iterator_stmt, ctx.env)
```
elprans, line 135: `ir.ast.Set` already has its type inferred, so you can just do `setgen.get_set_type(foo, ctx=ctx)`.  `infer_type` is smart enough to avoid re-inferring the type and is roughly the same, but `get_set_type` is cleaner.

PR #1606, opened 2020-07-29
Handle nested aliased types in GraphQL.

### LINKED ISSUE #722: GraphQL + Views and nested views
Consider the nested view case.

```
CREATE VIEW FooView := (SELECT Foo { bar: { baz := '1' } })
```

```
            query {
                FooView {
                    bar {
                        baz
                    }
                }
            }
```

Having to explicitly specify the derived type of the nested view in the query seems barely usable, especially since nested view names are difficult to look up and are necessarily verbose themselves.

The preferred solution here would be to implement GraphQL backend that can handle types with interfaces which have fields of the same name and related types. Currently, this scenario seems to be at least one of the possible interpretations of the GraphQL spec.

The spec is a bit vague:
https://graphql.github.io/graphql-spec/June2018/#sec-Objects
https://graphql.github.io/graphql-spec/June2018/#sec-Interfaces

The spec doesn't address the specific case of inheritance like this, but it forbids to have "different" fields with the same name and it requires the object to have all the fields of its interfaces. This would seem like a contradiction in case of implementing 2 interfaces with a field of the same name and related type.

Strictly speaking, the object is allowed to implement a field of the same name as the interface and of a sub-type of the interface type, so under this reading it can be argued that this should be legal, whether this was intended and there are any actual GraphQL implementations that do this is another question. So maybe this should be allowed once I reimplement the GraphQL validation.

This is similar to #709.

vpetrovykh:
Aliased types have layout similar to their original tpyes, but with
potentially additional links or properties. This means that the GraphQL
reflected fields must have additional arguments on them. This, in turn,
means that the aliased type cannot be simply derived from the original
type's interface because the arguments would then have to match as well.
So aliased types are reflected without attempting to preserve inheritance,
but instead the focus is on making all the fields accessible.

Fixes #722.

### DIFF
--- a/edb/graphql/types.py
+++ b/edb/graphql/types.py
@@ -57,6 +57,93 @@ from edb.schema import objects as s_objects
 from . import errors as g_errors
 
 
+'''
+This module is responsible for mapping EdgeDB types onto the GraphQL
+types. However, this is an imperfect mapping because not all the types
+or relationships between them can be expressed.
+
+# Aliased Types
+
+Aliased types present a particular problem. Basically, they break
+inheritance and GraphQL fragments become useless for them.
+
+Consider a link friends in this alias:
+```
+    type User {... multi link friends -> User}
+    type SpecialUser extending User {property special-> str}
+    alias UserAlias := User {friends: {some_new_prop := 'foo'}}
+```
+
+In GraphQL we will have type UserType implementing interfaces User,
+Object and SpecialUserType implementing SpecialUser, User, Object. The
+trouble starts with our implicit aliased type for the aliased friends
+link that targets __UserAlias__friends. This type gets reflected into
+a GraphQL _edb__UserAlias__friends that implements... what? We have 2
+options:
+
+    1) Implement interfaces mirroring the EdgeDB types: User, Object.
+
+    2) Implement it's own interface (or just omit interfaces here,
+       since if the interface is unique it's not adding anything).
+
+Case 2) preserves all the fields defined in the alias to be accessible
+and filterable, etc., but loses inheritance information.
+
+Case 1) leads to the following additional choices:
+
+    a) The friends target is still interface User, then the field
+       some_new_prop will not appear in the nesting, but will require
+       a specialized fragment:
+
+        query {
+            UserAlias {
+                friends {
+                    ... on _edb__UserAlias_friends {
+                        some_new_prop
+                    }
+                }
+            }
+        }
+
+    b) We can make the target of friends of UserAlias to be the actual
+       type _edb__UserAlias__friends (similar to case 2), but then we
+       cannot use the typed fragment `... on SpecialUser` construct
+       inside it because the SpecialUser is a sibling of our aliased
+       type and will cause a GraphQL validation error.
+
+    c) The field target can be a union type, but it will still only
+       have fields that are common to all union members and will
+       require awkward inlined typed fragments to work with just like
+       the first bullet point.
+
+In the end I think that rather than preserving the inheritance and
+then essentially forcing the use of `... on _edb__UserAlias_friends`
+just to access the very field for which the alias was created in the
+first place it's better to bite the bullet accept that in GraphQL
+aliased types reflection removes all inheritance info, but at least
+provide all the fields as per normal. The reasoning being that the
+fields and data are probably much more important for practical
+purposes than inheritance purity. To allow the SpecialUser
+polymorphism I'd rather suggest to the user to bake it into the alias
+like so:
+```
+    alias UserAlias := User {
+        friends: {
+            some_new_prop := 'foo',
+            [IS SpecialUser].special,
+        }
+    }
+```
+
+Also, note that for the same reasons as outlined above that make
+aliased types into a sibling branch in GraphQL, we can't give an
+accurate __typename for them, like we can in EdgeQL (__type__.name)
+because the "correct" types would violate the declared GraphQL type
+hierarchy. So aliased types are necessarily opaque in GraphQL in all
+these ways. Unlike in EdgeQL.
+'''
+
+
 def coerce_int64(value):
     if isinstance(value, int):
         num = value
@@ -243,6 +330,11 @@ class GQLCoreSchema:
         module, shortname = name.split('::', 1)
         if module in {'default', 'std'}:
             if shortname.startswith('__'):
+                # Use '_edb' prefix to mark derived and otherwise
+                # internal types. We opt out of '__edb' because we
+                # still rely on the first occurrence of '__' in
+                # GraphQL names to separate the module from the rest
+                # of the name in some code.
                 return '_edb' + shortname
             else:
                 return shortname
@@ -291,6 +383,11 @@ class GQLCoreSchema:
                 return el_type
             else:
                 target = GraphQLList(GraphQLNonNull(el_type))
+
+        elif edb_target.is_view(self.edb_schema):
+            target = self._gql_objtypes.get(
+                edb_target.get_name(self.edb_schema)
+            )
 
         elif edb_target.is_object_type():
             target = self._gql_interfaces.get(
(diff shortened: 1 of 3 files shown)


edb/edgeql/compiler/inference/cardinality.py
```diff
@@ -355,7 +355,7 @@ def __infer_set(
             else:
                 rptrref_card = rptrref.dir_cardinality
 
-            if rptrref_card.is_single():
+            if rptrref_card is not None and rptrref_card.is_single():
```
elprans, line 358: This is suspicious.  In what scenario is it `None`?

edb/graphql/types.py
```diff
@@ -395,7 +400,7 @@ def get_fields(self, typename):
         if typename == 'Query':
             for name, gqltype in sorted(self._gql_interfaces.items(),
                                         key=lambda x: x[1].name):
-                if name in TOP_LEVEL_TYPES:
+                if name in TOP_LEVEL_TYPES or gqltype.name.startswith('_edb'):
```
1st1, line 403: What's `_edb`?

vpetrovykh: Basically it's an indicator that the underlying EdgeDB entity is some kind of internal thing (has a name starting with `__`, which is illegal for user-land). This is true of nested aliased types that are auto-generated, for example. It makes sense for them to exist as types in the GraphQL schema, but they should not be exposed at top-level queriable things.

See [the code](https://github.com/edgedb/edgedb/blob/15980391ce5535e938c45f836548efed1307c360/edb/graphql/types.py#L242) where this name translation happens.

I can add a comment here to clarify, though.

1st1: > has a name starting with __, which is illegal for user-land

Yeah, that's why I asked, because I see only one `_` underscore, not two.

edb/edgeql/compiler/inference/cardinality.py
vpetrovykh: Try the following in `graphqlfunctional` test database:
```
SELECT default::SettingAliasAugmented {
    of_group: {
        name_upper
    }
} FILTER .of_group.name_upper > "U";
```
(This is also triggered by the `test_graphql_functional_alias_05` from this PR)

I was using the reasoning that in an "if/else" situation everything that's not explicitly in the "if" is intended to fall under else, so I didn't think too much before adding the `is not None` part. You're right that it may indicate a different bug instead.

edb/graphql/types.py
vpetrovykh: I think that at some point we were debating whether to use `__` or `_` prefix in GraphQL-reflected names for internal stuff and decided that `__` was probably better left to the actual GraphQL-specific things (`__typename`, etc.), but our own translated internal entities would be prefixed with `_`, which later got to be `_edb` to potentially allow marking other flavours of internal stuff with a different `_something` prefix.

vpetrovykh:
I looked into the fragments and inheritance. So we've seen this problem before. Basically, the aliased types break inheritance and GraphQL fragments become useless for them.

Consider a link `friends` in this alias:
```
type User {... multi link friends -> User}
type SpecialUser extending User {property special-> str}
alias UserAlias := User {friends: {some_new_prop := 'foo'}}
```
In GraphQL we will have type `UserType` implementing interfaces `User`, `Object` and `SpecialUserType` implementing `SpecialUser`, `User`, `Object`.
The trouble starts with our implicit aliased type for the aliased friends link that targets `__UserAlias__friends`. This type gets reflected into a GraphQL `_edb__UserAlias__friends` that implements... what? We have 2 options:
1) Implement interfaces mirroring the EdgeDB types: `User`, `Object`.
2) Implement it's own interface (or just omit interfaces here, since if the interface is unique it's not adding anything).

Case 2. was implemented in the PR I have, it preserves all the fields defined in the alias to be accessible and filterable, etc.

Case 1. leads to the following additional choices:
* The `friends` target is still interface `User`, then the field `some_new_prop` will not appear in the nesting, but will require a specialized fragment `query {UserAlias { friends { ... on _edb__UserAlias_friends { some_new_prop }}}}`
* We can make the target of `friends` of `UserAlias` to be the actual type `_edb__UserAlias__friends` (similar to case 2), but then we cannot use the typed fragment `... on SpecialUser` construct inside it because the `SpecialUser` is a sibling of our aliased type and will cause a GraphQL validation error.
* The field target can be a union type, but it will still only have fields that are common to all union members and will require awkward inlined typed fragments to work with just like the first bullet point.

In the end I think that rather than preserving the inheritance and then essentially forcing the use of ... on `_edb__UserAlias_friends` just to access the very field for which the alias was created in the first place it's better to bite the bullet accept that in GraphQL aliased types reflection removes all inheritance info, but at least provide all the fields as per normal. The reasoning being that the fields and data are probably much more important for practical purposes than inheritance purity. To allow the `SpecialUser` polymorphism I'd rather suggest to the user to bake it into the alias like so:
```
alias UserAlias := User {
    friends: {
        some_new_prop := 'foo',
        [IS SpecialUser].special,
    }
}
```

Also, note that for the same reasons as outlined above that make aliased types into a sibling branch in GraphQL, we can't give an accurate `__typename` for them, like we can in EdgeDB (`__type__.name`) because the "correct" types would violate the declared GraphQL type hierarchy. So aliased types are necessarily opaque in GraphQL in all these ways. Unlike in EdgeDB.

edb/graphql/types.py
1st1: Are the `_edb` things visible in introspection? I'd suggest to use `__edb_` prefix -- it's unlikely to clash with anything in GraphQL and is distinctly a private system name.

1st1:
> Implement it's own interface (or just omit interfaces here, since if the interface is unique it's not adding anything).

I think this is an OK approach.

> it's better to bite the bullet accept that in GraphQL aliased types reflection removes all inheritance info

Yeah, it seems that alternative approaches are way more complicated and the gain is little at best.

edb/graphql/types.py
```diff
@@ -57,6 +57,93 @@
...
+        }
+    }
+```
+
+Also, note that for the same reasons as outlined above that make
+aliased types into a sibling branch in GraphQL, we can't give an
+accurate __typename for them, like we can in EdgeDB (__type__.name)
+because the "correct" types would violate the declared GraphQL type
+hierarchy. So aliased types are necessarily opaque in GraphQL in all
+these ways. Unlike in EdgeDB.
```
elprans, line 143: s/EdgeDB/EdgeQL/

edb/edgeql/compiler/inference/cardinality.py
vpetrovykh: This was resolved by a separate issue.

edb/graphql/types.py
vpetrovykh: I've added a comment regarding the choice of prefix and its meaning.

edb/graphql/types.py
vpetrovykh: Done

edb/graphql/types.py
```diff
@@ -243,6 +330,11 @@ def get_gql_name(self, name):
         module, shortname = name.split('::', 1)
         if module in {'default', 'std'}:
             if shortname.startswith('__'):
+                # Use '_edb' prefix to mark derived and otherwise
+                # internal types. We opt out of '__edb' because we
+                # still rely on the first occurrence of '__' in
```
elprans, line 335: If it's a leading `__`, why would that be ambiguous with a module separator that would be on the interior of the name?

vpetrovykh: Because the place where it matters the first occurrence of `__` is used as a separator for the module name. It's possible to work around it there, but it seems to be trading one awkward name for another.

vpetrovykh: Actually, there's a simpler reason - leading `__` is reserved by GraphQL and the validator throws an exception when it sees that.

Issue #1614, opened 2020-07-31
Cardinality calculation causes error


vpetrovykh:
This issue appears in the ff869ab54d63968081d54e99b4dfb8f2b62f9ce8

Steps to Reproduce:

In `graphqlfunctional` database, try:
```
SELECT default::SettingAliasAugmented {
    of_group: {
        name_upper
    }
} FILTER .of_group.name_upper > "U";
```

This causes the following error:
```
ERROR: InternalServerError: 'NoneType' object has no attribute 'is_single'
  Hint: This is most likely a bug in EdgeDB. Please consider opening an issue ticket at https://github.com/edgedb/edgedb/issues/new?template=bug_report.md
  Server traceback:
      Traceback (most recent call last):
        File "/home/victor/dev/magicstack/edgedb/edb/server/procpool/worker.py", line 75, in worker
          res = await meth(*args)
        File "/home/victor/dev/magicstack/edgedb/edb/server/compiler/compiler.py", line 1683, in compile_eql_tokens
          return self._compile(ctx=ctx, tokens=eql_tokens)
        File "/home/victor/dev/magicstack/edgedb/edb/server/compiler/compiler.py", line 1319, in _compile
          comp: dbstate.BaseQuery = self._compile_dispatch_ql(ctx, stmt)
        File "/home/victor/dev/magicstack/edgedb/edb/server/compiler/compiler.py", line 1282, in _compile_dispatch_ql
          return self._compile_ql_query(ctx, ql)
        File "/home/victor/dev/magicstack/edgedb/edb/server/compiler/compiler.py", line 589, in _compile_ql_query
          ir = qlcompiler.compile_ast_to_ir(
        File "/home/victor/dev/magicstack/edgedb/edb/edgeql/compiler/__init__.py", line 175, in wrapper
          return func(*args, **kwargs)
        File "/home/victor/dev/magicstack/edgedb/edb/edgeql/compiler/__init__.py", line 224, in compile_ast_to_ir
          ir_expr = stmtctx_mod.fini_expression(ir_set, ctx=ctx)
        File "/home/victor/dev/magicstack/edgedb/edb/edgeql/compiler/stmtctx.py", line 116, in fini_expression
          cb(ctx=ctx)
        File "/home/victor/dev/magicstack/edgedb/edb/edgeql/compiler/stmtctx.py", line 512, in _infer_pointer_cardinality
          inferred_card = infer_expr_cardinality(irexpr=irexpr, ctx=ctx)
[...]

PR #1615, opened 2020-07-31
Do not confuse default link base as "derived base"

elprans:
When handling computable pointers, make sure that we're not peeling
the derived pointer to the top `std::link` or `std::pointer`.  In
fact, `PointerRef.base_ptr` should never contain the top pointer.

Fixes: #1614.

### DIFF
--- a/edb/ir/typeutils.py
+++ b/edb/ir/typeutils.py
@@ -541,13 +541,17 @@ def ptrref_from_ptrcls(
     base_ptr: Optional[irast.BasePointerRef]
     if is_derived:
         base_ptrcls = ptrcls.get_bases(schema).first(schema)
-        base_ptr = ptrref_from_ptrcls(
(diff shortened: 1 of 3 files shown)


edb/ir/typeutils.py
```diff
@@ -541,13 +541,17 @@ def ptrref_from_ptrcls(
...
     if is_derived:
         base_ptrcls = ptrcls.get_bases(schema).first(schema)
-        base_ptr = ptrref_from_ptrcls(
-            ptrcls=base_ptrcls,
-            direction=direction,
-            schema=schema,
-            cache=cache,
-            typeref_cache=typeref_cache,
-        )
+        top = type(base_ptrcls).get_default_base_name()
```
1st1, line 544:
```suggestion
        top_name = type(base_ptrcls).get_default_base_name()
```

PR #1621, opened 2020-08-01
Optimize handling of singleton right operand in ??

elprans:
When the cardinality of the right operand in ?? is known to be exactly
`ONE`, we can use a scalar SQL `COALESCE` without any precautions.  If
it's `AT_MOST_ONE`, use a fact that `(SELECT (<empty_rel>))` is `NULL`
to avoid using `force_optional`, because the latter largely defeats the
optimization.

This also fixes cardinality inference of certain type intersection
paths.

### DIFF
--- a/edb/edgeql/compiler/inference/cardinality.py
+++ b/edb/edgeql/compiler/inference/cardinality.py
@@ -272,11 +272,11 @@ def __infer_type_introspection(
     return ONE
 
 
-def _is_visible(
+def _find_visible(
     ir: irast.Set,
     scope_tree: irast.ScopeTreeNode,
     env: context.Environment,
-) -> bool:
+) -> Optional[irast.ScopeTreeNode]:
     parent_fence = scope_tree.parent_fence
     if parent_fence is not None:
         if scope_tree.namespaces:
@@ -284,9 +284,9 @@ def _is_visible(
         else:
             path_id = ir.path_id
 
-        return parent_fence.is_visible(path_id)
+        return parent_fence.find_visible(path_id)
     else:
-        return False
+        return None
 
 
 @_infer_cardinality.register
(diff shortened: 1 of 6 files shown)


edb/edgeql/compiler/inference/cardinality.py
```diff
@@ -276,17 +276,17 @@ def _is_visible(
     ir: irast.Set,
     scope_tree: irast.ScopeTreeNode,
     env: context.Environment,
-) -> bool:
+) -> Optional[irast.ScopeTreeNode]:
```
1st1, line 279: Rename to `def _find_visible`?

edb/edgeql/compiler/inference/cardinality.py
```diff
@@ -297,8 +297,10 @@ def __infer_set(
     singletons: Collection[irast.PathId],
     env: context.Environment,
 ) -> qltypes.Cardinality:
-    if _is_visible(ir, scope_tree, env) or ir.path_id in singletons:
+    if ir.path_id in singletons:
         return ONE
+    if node := _is_visible(ir, scope_tree, env):
```
1st1, line 302:
```suggestion
    if node := _is_visible(ir, scope_tree, env) is not None:
```

edb/ir/ast.py
```diff
@@ -563,7 +563,7 @@ class CallArg(ImmutableBase):
     __ast_mutable_fields__ = frozenset(('cardinality',))
 
     expr: Set
-    cardinality: qltypes.Cardinality = qltypes.Cardinality.ONE
+    cardinality: qltypes.Cardinality = qltypes.Cardinality.UNKNOWN
```
1st1, line 566: sigh

1st1:
> This also fixes cardinality inference of certain type intersection
paths.

Can we add a test for that?

Issue #1618, opened 2020-08-01
Changing the text of an annotation via SDL returns InternalServerError: 'NoneType' object has no attribute 'get_name'


CodesInChaos:
- EdgeDB Version: Alpha 4
- OS Version: Ubuntu 20

Create the following schema via SDL:

```
START MIGRATION TO {
	module default {
		abstract annotation description;
		
		abstract link Ordered {
			annotation description := "1";
		}
	}
};
populate migration;
commit migration;
```

then execute the same migration again, with a different value for the description (e.g. `"2"`).

It fails with the following error:

```
ERROR: InternalServerError: 'NoneType' object has no attribute 'get_name'
  Hint: This is most likely a bug in EdgeDB. Please consider opening an issue ticket at https://github.com/edgedb/edgedb/issues/new?template=bug_report.md
  Server traceback:
      Traceback (most recent call last):
        File "/usr/lib/x86_64-linux-gnu/edgedb-server-1-alpha4/lib/python3.8/site-packages/edb/server/procpool/worker.py", line 75, in worker
          res = await meth(*args)
        File "/usr/lib/x86_64-linux-gnu/edgedb-server-1-alpha4/lib/python3.8/site-packages/edb/server/compiler/compiler.py", line 1704, in compile_eql_tokens_in_tx
          return self._compile(ctx=ctx, tokens=eql_tokens)
        File "/usr/lib/x86_64-linux-gnu/edgedb-server-1-alpha4/lib/python3.8/site-packages/edb/server/compiler/compiler.py", line 1319, in _compile
          comp: dbstate.BaseQuery = self._compile_dispatch_ql(ctx, stmt)
        File "/usr/lib/x86_64-linux-gnu/edgedb-server-1-alpha4/lib/python3.8/site-packages/edb/server/compiler/compiler.py", line 1242, in _compile_dispatch_ql
          return self._compile_ql_migration(ctx, ql)
        File "/usr/lib/x86_64-linux-gnu/edgedb-server-1-alpha4/lib/python3.8/site-packages/edb/server/compiler/compiler.py", line 937, in _compile_ql_migration
          ddl_query = self._compile_and_apply_ddl_stmt(
        File "/usr/lib/x86_64-linux-gnu/edgedb-server-1-alpha4/lib/python3.8/site-packages/edb/server/compiler/compiler.py", line 748, in _compile_and_apply_ddl_stmt
          block, new_types = self._process_delta(ctx, delta)
[...]

PR #1624, opened 2020-08-01
Use ddl_identity for annotation refs in annotation value commands

elprans:
Currently, reflection of annotation value commands relies on the
`annotation` reference to be present in the command, which is not always
the case, such as when only the annotation value is altered.  Use the
new DDL identity mechanism to always guarantee that the reference is
there in a canonical representation of a command.  While at it, do some
minor refactoring to make DDL identity handling sligtly more formal and
general.

Fixes: #1618.

### DIFF
--- a/edb/schema/annos.py
+++ b/edb/schema/annos.py
@@ -67,7 +67,7 @@ class AnnotationValue(
         so.Object, compcoef=1.0, default=None, inheritable=False)
 
     annotation = so.SchemaField(
-        Annotation, compcoef=0.429)
+        Annotation, compcoef=0.429, ddl_identity=True)
 
     value = so.SchemaField(
         str, compcoef=0.909)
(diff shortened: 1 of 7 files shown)


edb/schema/annos.py
```diff
@@ -67,7 +67,7 @@ class AnnotationValue(
         so.Object, compcoef=1.0, default=None, inheritable=False)
 
     annotation = so.SchemaField(
-        Annotation, compcoef=0.429)
+        Annotation, compcoef=0.429, ddl_identity=True)
```
1st1, line 70: Why is it called "DDL identity"? The name isn't obvious.

edb/schema/delta.py
```diff
@@ -1660,6 +1667,12 @@ def has_ddl_identity(self, aspect: str) -> bool:
             and self.ddl_identity.get(aspect) is not None
         )
 
+    def set_ddl_identity(self, aspect: str, value: Any) -> None:
+        if self.ddl_identity is None:
+            self.ddl_identity = {}
+
+        self.ddl_identity[aspect] = value
```
1st1, line 1674: Is it OK if we're overriding some aspect that's already there?  Or should there be separate API for that?

elprans: Overwriting existing values is harmless here.

edb/schema/annos.py
elprans: The `Field` attribute has this comment:

```
    #: Whether the field is used to identify the object
    #: in DDL operations and schema reflection when object
    #: name is insufficient.
    ddl_identity: bool
```

PR #1628, opened 2020-08-03
Implement basic select-or-insert via coalesce

### LINKED ISSUE #684: SELECT Object IF EXISTS Object ELSE INSERT

msullivan:
Implement a select-or-insert idiom using
`(SELECT ...) ?? (INSERT ...)`.

We only allow it under restricted circumstances (when the select and
insert both use the same value for an exclusive property) and it is
compiled using ON CONFLICT.

This doesn't work when the values are constants yet, because the
normalizer treats each constant separately. I'll change that in a
follow-up.

Mostly fixes #684, modulo follow-up.


edb/edgeql/compiler/stmt.py
```diff
@@ -254,16 +258,105 @@ def compile_GroupQuery(
...
+            ptr.is_id_pointer(schema) or
+            any(c.issubclass(schema, exclusive_constr)
+                for c in ptr.get_constraints(schema).objects(schema))
+        )
+        if not is_unique or not ptr.is_property(schema):
+            error("FILTER is not on an exclusive property")
+
+    shape_props = {}
+    for shape_set, _ in rhs.subject.shape:
+        # what's the base_ptr deal here???
```
elprans, line 331: `Foo { some_ptr }` produces a virtual subtype of `Foo` (let's call it `Foo1`), where `Foo1.some_ptr` is a derivative of `Foo.some_ptr`.  `base_ptr` in `rptr` points to the original `Foo.some_ptr` (which would be the actual property in schema).

edb/edgeql/compiler/stmt.py
```diff
@@ -254,16 +258,105 @@ def compile_GroupQuery(
...
 
 
-@dispatch.compile.register(qlast.InsertQuery)
-def compile_InsertQuery(
-        expr: qlast.InsertQuery, *, ctx: context.ContextLevel) -> irast.Set:
+def simple_stmt_eq(lhs: irast.Base, rhs: irast.Base) -> bool:
+    if (
+        isinstance(lhs, irast.BaseConstant)
+        and isinstance(rhs, irast.BaseConstant)
+        and lhs.typeref == rhs.typeref  # XXX?
```
elprans, line 265: How about using `evaluate_to_python_val` here to compute the value of the constant for comparison?

edb/edgeql/compiler/stmt.py
```diff
@@ -254,16 +258,105 @@ def compile_GroupQuery(
...
+            any(c.issubclass(schema, exclusive_constr)
+                for c in ptr.get_constraints(schema).objects(schema))
+        )
+        if not is_unique or not ptr.is_property(schema):
+            error("FILTER is not on an exclusive property")
+
+    shape_props = {}
+    for shape_set, _ in rhs.subject.shape:
+        # what's the base_ptr deal here???
+        base_ptr = shape_set.rptr.ptrref.base_ptr
```
elprans, line 332: You can use `ptrcls_from_ptrref` to obtain the pointer schema object from the descriptor and then compare it directly with `filtered_ptrs` (well, `get_nearest_non_derived_parent` thereof).

edb/ir/ast.py
```diff
@@ -715,7 +715,7 @@ class MutatingStmt(Stmt):
 
 
 class InsertStmt(MutatingStmt):
-    pass
+    on_conflict: typing.Optional[typing.List[PointerRef]] = None
```
elprans, line 718: SQL `ON CONFLICT` accepts constraint names as conflict targets, let's use a constraint object directly.  I suggest adding a `ConstraintRef` to IR:

```python
class ConstraintRef(ImmutableBase):
    # The id of the constraint
    id: uuid.UUID
    # The module id of the constraint
    module_id: uuid.UUID
```

You can obtain the name of the SQL constraint with `common.get_constraint_backend_name()`

edb/edgeql/compiler/stmt.py
msullivan: Oh, yeah. I looked at doing that for the whole comparison and didn't because it wouldn't work for parameters but it should work here.

Does the typeref `==` comparison actually work, by the way?

elprans: It does so by accident because we maintain a `type -> typeref` cache.  If you want to compare them reliably, you have to compare `typeref.id`, but it'd be much better to do `ir_typeref_to_type` and compare the type object instead.

edb/ir/ast.py
msullivan: That turns out to not get the right constraint name, but I think I figured it out.

tests/test_edgeql_insert.py
```diff
@@ -1715,3 +1715,121 @@ async def test_edgeql_insert_correlated_bad_02(self):
                     Subordinate,
                 );
             ''')
+
+
+class TestInsertCoalesce(tb.QueryTestCase):
```
1st1, line 1720: I wouldn't add another class as that would slow down test start time (every test case gets its own DB which takes some time).

edb/edgeql/compiler/expr.py
```diff
@@ -71,10 +72,54 @@ def compile_Path(
...
+        lir = setgen.ensure_set(dispatch.compile(left, ctx=lctx), ctx=lctx)
+        lir = setgen.scoped_set(setgen.ensure_stmt(lir, ctx=lctx), ctx=lctx)
+    with ctx.newscope() as rctx:
+        rir = setgen.ensure_set(
+            stmt.compile_InsertQuery(right, ctx=rctx, conditioned_on=lir),
+            ctx=rctx
+        )
+        rir = setgen.scoped_set(setgen.ensure_stmt(rir, ctx=rctx), ctx=rctx)
+
+    # Now we flip-flop the coalesce so that the SELECT gets executed
```
elprans, line 93: The comment seems outdated since we use the `UNION` approach now.

PR #1632, opened 2020-08-04
Make query normalization use same variable for same constants

msullivan:
This is needed to make select-or-insert work when using constants,
since otherwise this hides the equality of constants from the
compiler.

Work on #684.

### DIFF
--- a/edb/edgeql-rust/src/normalize.rs
+++ b/edb/edgeql-rust/src/normalize.rs
@@ -1,4 +1,5 @@
 use std::collections::{HashMap, BTreeSet};
+use std::hash::{Hash, Hasher};
 
 use edgeql_parser::tokenizer::{TokenStream, Kind};
 use edgeql_parser::position::Pos;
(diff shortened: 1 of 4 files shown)


edb/edgeql-rust/src/normalize.rs
```diff
@@ -7,11 +7,11 @@ use bigdecimal::BigDecimal;
 use crate::tokenizer::{CowToken, decode_string};
 
 
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Eq, Hash, Clone)]
 pub enum Value {
     Str(String),
     Int(i64),
-    Float(f64),
+    Float(u64), // ... we store the raw bits here so that this can be Eq
```
tailhook, line 14: Let's make it `[u8; 8]` if we want to store bits. But I'm not sure if this is optimal.

msullivan:
Closing this per redesign in #1639 making this not necessary for now

PR #1654, opened 2020-08-11
Tiny fix of DESCRIBE CURRENT MIGRATION

tailhook:
Current master fails on empty `proposed` list.

### DIFF
--- a/edb/server/compiler/compiler.py
+++ b/edb/server/compiler/compiler.py
@@ -882,10 +882,10 @@ class Compiler(BaseCompiler):
                     'confirmed': confirmed,
                     'proposed': [{
                         'statements': [{
-                            'text': proposed[0],
+                            'text': item,
                         }],
                         'confidence': 1.0,
-                    }],
+                    } for item in proposed],
                 }).encode('unicode_escape').decode('utf-8')
 
                 desc_ql = edgeql.parse(


elprans:
The current coding is intentional: it sends only the first statement of each proposed scenario instead of the whole script.  Clarification: current implementation does not yet support variants properly.

tailhook:
> The current coding is intentional: it sends only the first statement of each proposed scenario instead of the whole script. Clarification: current implementation does not yet support variants properly.

Hm. Okay. But this should be fixed anyway:
```
Error: ERROR: InternalServerError: tuple index out of range
  Server traceback:
        File "/work/edb/server/procpool/worker.py", line 75, in worker
          res = await meth(*args)
        File "/work/edb/server/compiler/compiler.py", line 1712, in compile_eql_tokens_in_tx
          return self._compile(ctx=ctx, tokens=eql_tokens)
        File "/work/edb/server/compiler/compiler.py", line 1327, in _compile
          comp: dbstate.BaseQuery = self._compile_dispatch_ql(ctx, stmt)
        File "/work/edb/server/compiler/compiler.py", line 1250, in _compile_dispatch_ql
          return self._compile_ql_migration(ctx, ql)
        File "/work/edb/server/compiler/compiler.py", line 885, in _compile_ql_migration
          'text': proposed[0],
      IndexError: tuple index out of range
```

elprans:
> But this should be fixed anyway:

Ah, yes, of course.

elprans:
The `IndexError` here was fixed some time ago.

PR #1656, opened 2020-08-11
Fix issues with nested inserts

msullivan:
When an enclosing select-or-insert (which will soon become INSERT
UNLESS CONFLICT) conflicts, don't do nested inserts. Also, when there
are nested FORs, the inner inserts need to happen for each nested FOR.

Work on #684.

### DIFF
--- a/edb/pgsql/compiler/dml.py
+++ b/edb/pgsql/compiler/dml.py
@@ -1241,21 +1241,6 @@ def process_link_values(
                 row_query, ir_stmt.subject.path_id, env=input_rel_ctx.env)
             dispatch.visit(ir_expr, ctx=input_rel_ctx)
 
-            if is_insert:
-                target_rvar = pathctx.get_path_rvar(
-                    input_rel,
-                    ir_expr.path_id,
-                    aspect='identity',
-                    env=input_rel_ctx.env,
-                )
-                pathctx.put_path_rvar(
-                    input_rel,
-                    ir_stmt.subject.path_id,
-                    rvar=target_rvar,
-                    aspect='identity',
-                    env=input_rel_ctx.env,
-                )
-
             if (
                 isinstance(ir_expr.expr, irast.Stmt)
                 and ir_expr.expr.iterator_stmt is not None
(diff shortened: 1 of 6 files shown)


edb/pgsql/compiler/dml.py
```diff
@@ -437,6 +471,27 @@ def process_insert_body(
                             path_id=ictx.rel.path_id, ctx=ctx)
         iterator_id = pathctx.get_path_identity_var(
             select, iterator_set.path_id, env=ctx.env)
+
+        # debug.header("IT IS THAT ITERATOR SHIT")
```
1st1, line 475: Let's not commit any debug code, even when it's commented out

msullivan: Oh, crap.

edb/pgsql/compiler/dml.py
```diff
@@ -1172,9 +1225,11 @@ def process_link_values(
         if iterator_cte is not None:
             iterator_rvar = relctx.rvar_for_rel(
                 iterator_cte, lateral=True, ctx=subrelctx)
-            relctx.include_rvar(row_query, iterator_rvar,
-                                path_id=iterator_cte.query.path_id,
-                                ctx=subrelctx)
+            # XXX???
+            if iterator_cte.query.path_id:
```
elprans, line 1229: Perhaps pass the `(iterator_cte, iterator_set)` tuple to avoid having to fetch the `path_id`?

edb/pgsql/compiler/dml.py
```diff
@@ -1222,40 +1227,24 @@ def process_link_values(
...
+            relctx.include_rvar(row_query, iterator_rvar,
+                                path_id=iterator_set.path_id,
+                                ctx=subrelctx)
 
         with subrelctx.newscope() as sctx, sctx.subrel() as input_rel_ctx:
             input_rel = input_rel_ctx.rel
-            if iterator_cte is not None:
-                input_rel_ctx.path_scope[iterator_cte.query.path_id] = \
+            if iterator is not None:
+                input_rel_ctx.path_scope[iterator[1].query.path_id] = \
```
elprans, line 1241:
```suggestion
                input_rel_ctx.path_scope[iterator[0].path_id] = \
```

PR #1658, opened 2020-08-12
Fix schema diff regression

elprans:
Optimizations in commit 589ed705b8b0 introduced a regression that
resulted in corrupt diffs being generated intermittently which resulted
in `test_constraints_exclusive_migration` failing.


elprans:
> Is it possible to design a deterministically failing test for this?

Yes, once we start writing proper tests for the new migration flow.

Issue #1639, opened 2020-08-06
Specialize 'select-or-insert' and 'update-or-insert'


elprans:
Recent work on adding the select-or-insert and update-or-insert idioms to EdgeQL (#1628, #684) spawned a discussion on whether using a general-purpose conditional operator `??` is a good idea after all.

The approach committed in #1628 has lots of technical limitations, and, although most of those can be fixed with more work, the main issue with it remains: it gives a misleading appearance of generic support for conditional DML which is impossible as long as it is in PostgreSQL.

For this reason I propose reverting the select-or-insert and update-or-insert to a specialized form of the `INSERT` statement:

```
INSERT Type { <shape> } 
UNLESS CONFLICTS [ <conflict-target> [ ELSE <alternative-stmt> ] ]
```

where `alternative-stmt` may be any other query, including `UPDATE` or `DELETE`.

This is semantically much closer to PostgreSQL's `INSERT ... ON CONFLICT ...` and allows explicit specification of the conflict target (i.e. the constraint expression).

elprans:
An important question to answer in this formulation is whether the `Type` reference is bound in the `ELSE` clause.  If it is, then the alternative statements would not need a `FILTER`, so 

```
INSERT User { ... } UNLESS CONFLICTS .name 
ELSE SELECT User
``` 

and

```
INSERT User { ... } UNLESS CONFLICTS .name
ELSE UPDATE User SET { ... }
```

FWIW, PostgreSQL binds the name of the table in `ON CONFLICT DO UPDATE`.

Another option would be to add a special keyword to refer to the conflicting object (e.g `__conflicting__`), but I prefer that less.

vpetrovykh:
If I understand this proposal correctly I can write:
```
INSERT Foo {bar := 1} UNLESS CONFLICTS
```
Which will return an `{}` in case of conflict instead of raising an exception. So it should be legal to write:
```
SELECT (INSERT Foo {bar := 1} UNLESS CONFLICTS) ?? (SELECT Foo FILTER .bar = 1)
```
or, if I want a different fallback:
```
SELECT (INSERT Foo {bar := 1} UNLESS CONFLICTS) ?? (SELECT Foo FILTER .bar = 42)
```

How is using an `ELSE` clause really helping here? Especially if the `Foo` reference is bound as suggested, then the last expression would have to be:
```
WITH Foo2 := Foo  # going to need this for selecting a default 
                  # that's something other than the conflicted value
[...]

PR #1659, opened 2020-08-12
Initial implementation on ON CONFLICT

msullivan:
Add support for `INSERT ... ON CONFLICT`. This is an initial take,
without ELSE, so it replaces what has already been implemented as the
coalescing based select-or-insert.

This also adds support to the lexer/parser for the `ELSE` branch, but
doesn't implement them yet.

First steps of #1639.


edb/edgeql/compiler/stmt.py
```diff
@@ -228,86 +227,58 @@ def simple_stmt_eq(lhs: irast.Base, rhs: irast.Base,
...
 ) -> irast.ConstraintRef:
-    def error(s: str) -> NoReturn:
+    ctx.partial_path_prefix = stmt.subject
+
+    # We compile the name here so we can analyze it, but we don't do
+    # anything else with it.
+    cspec_res = setgen.ensure_set(dispatch.compile(
+        constraint_spec, ctx=ctx), ctx=ctx)
+
+    if not cspec_res.rptr:
```
elprans, line 242: This isn't enough, you should also check that: 1) `rptr.source` is the same as `InsertStmt` subject and that the cardinality of the property is `single`.

edb/edgeql/compiler/stmt.py
```diff
@@ -362,6 +333,22 @@ def compile_InsertQuery(
...
+
+            if constraint_spec:
+                with ictx.new() as constraint_ctx:
+                    stmt.on_conflict = compile_insert_on_conflict_constraint(
+                        stmt, constraint_spec, ctx=constraint_ctx)
+            else:
+                stmt.on_conflict = True
+
+            if else_branch:
+                raise errors.QueryError(
```
1st1, line 347: nit: an `UnsupportedFeatureError` error would be more appropriate. See the full list here: https://github.com/edgedb/edgedb/blob/master/edb/api/errors.txt#L26

edb/edgeql/compiler/stmt.py
1st1: And then consider raising `InvalidTypeError`.

tests/test_edgeql_insert.py
```diff
@@ -1795,28 +1794,18 @@ async def test_edgeql_insert_dependent_03(self):
...
-            variables=("test",)
-        )
-
-        await self.assert_query_result(
-            query,
-            [{"name": "test"}],
-            variables=("test",)
-        )
+        await self.con.execute(query)
+        await self.con.execute(query)
```
1st1, line 1805: Not sure what the test here is. Do we need to execute two times and not check the result?

msullivan: We're checking that the inner inserts don't happen again when the outer insert conflicts. I'll add a comment.

1st1: Then you need to check the results. The first one is supposed to return an object, the second is supposed to be an empty set.

msullivan: I can add checks, sure. The reason I skipped them is because I had other tests for the core behavior, and this was just trying to test the nested inserts

1st1: Got it. Since it isn't hard to add specific checks I usually do that anyways, you never know what kind of regressions we might have.

edb/edgeql/compiler/stmt.py
msullivan: It doesn't seem like we raise InvalidTypeError much at all elsewhere? Am I missing something?

1st1: I don't think we're too diligent with narrowing down error types, we need to improve with that. Most of the time we raise `QueryError` simply because it's a habit (and often a copy/paste). That said, our errors are hierarchical, and InvalidTypeError is a subclass of QueryError, and it seems that in this very case it's an appropriate one.

1st1: BTW, `api/errors.txt` is not set in stone, so if you see that we're misusing QueryError in some common cases feel free to propose adding new types there.

edb/edgeql/parser/grammar/expressions.py
```diff
@@ -582,6 +585,35 @@ def reduce_SimpleShapePointer_REMASSIGN_Expr(self, *kids):
...
+class OnConflictElse(Nonterm):
+    def reduce_ELSE_Expr(self, *kids):
+        self.val = kids[1].val
+
+    def reduce_empty(self, *kids):
+        self.val = None
+
+
+class OnConflictCause(Nonterm):
+    def reduce_UNLESS_CONFLICT_OnConflictSpecifier_OnConflictElse(self, *kids):
```
elprans, line 605: This allows `ELSE` without an explicit conflict specifier, I think we should not allow that (at least initially).

tests/test_edgeql_insert.py
```diff
@@ -1718,6 +1718,66 @@ async def test_edgeql_insert_correlated_bad_02(self):
...
+
+        query2 = r'''
+        SELECT
+         ((INSERT test::Person {name := <str>$0} UNLESS CONFLICT)
+          ?? (SELECT test::Person FILTER .name = <str>$0));
+        '''
+
+        res = await self.con.query(query2, "test2")
+        res2 = await self.con.query(query2, "test2")
+        assert res == res2
```
1st1, line 1746: self.assertEqual

1st1: we don't use the assert statement in tests

msullivan:
I think this should be good to merge?

PR #1668, opened 2020-08-15
Fix INSERT nested in an UPDATE in a FOR

msullivan:


edb/pgsql/compiler/dml.py
```diff
@@ -1300,7 +1298,8 @@ def process_link_values(
         )
     )
 
-    if is_insert:
+    if len(ctx.toplevel_stmt.ctes) > old_cte_count:
```
elprans, line 1301: This seems a bit flimsy.  Right now we don't use CTEs for anything other than DML, but that may change without notice.  I wonder if it is possible to do `pathctx.put_path_bond(parent_insert_path_id)` on the dependent DML CTE instead where we handle the `pseudo_iterator`?

elprans: On a second thought, putting a bond in the dependent DML CTE will not be enough if said DML is nested, so the external tracking and injection approach here is better.  That said, let's use an explicit DML statement tracker instead of `toplevel_stmt.ctes`.

msullivan:
The other way doesn't work. Fixing it by reusing the iterator machinery in the same way is hampered by the fact that UPDATE with an iterator does not work. I will open an issue.

PR #1672, opened 2020-08-17
graphql: Add an `exists` filter operation.

### LINKED ISSUE #1655: GraphQL does not support `EXISTS`
There seems to be no way to test for empty set in GraphQL filters.

vpetrovykh:
The `exists` filter operation appears at every nested level regardless
of whether the level represents a link or a property. The operation
takes either a `true` or `false`, which translates to `EXISTS` and `NOT
EXISTS` EdgeQL respectively.

Fixes #1655

### DIFF
--- a/tests/test_http_graphql_query.py
+++ b/tests/test_http_graphql_query.py
@@ -3227,6 +3227,210 @@ class TestGraphQLFunctional(tb.GraphQLTestCase):
             ]
         })
 
+    def test_graphql_functional_exists_01(self):
+        self.assert_graphql_query_result(r"""
+            query {
+                User(
+                    filter: {profile: {exists: true}},
+                    order: {name: {dir: ASC}}
+                ) {
+                    name
+                    profile {
+                        name
+                    }
+                }
+            }
+        """, {
+            "User": [
+                {
+                    "name": "Alice",
+                    "profile": {
+                        "name": "Alice profile",
+                    },
+                },
+                {
+                    "name": "Bob",
+                    "profile": {
+                        "name": "Bob profile",
+                    },
+                },
+            ]
+        })
+
+    def test_graphql_functional_exists_02(self):
+        self.assert_graphql_query_result(r"""
+            query {
+                User(
+                    filter: {profile: {exists: false}},
+                    order: {name: {dir: ASC}}
+                ) {
+                    name
+                    profile {
+                        name
+                    }
+                }
+            }
+        """, {
+            "User": [
+                {
+                    "name": "Jane",
+                    "profile": None,
+                },
+                {
+                    "name": "John",
+                    "profile": None,
+                },
+            ]
+        })
+
+    def test_graphql_functional_exists_03(self):
+        self.assert_graphql_query_result(r"""
+            query {
+                User(
+                    filter: {groups: {settings: {exists: false}}},
+                    order: {name: {dir: ASC}}
+                ) {
+                    name
+                    groups {
+                        name
+                        settings {
+                            name
+                        }
+                    }
+                }
+            }
+        """, {
+            "User": [
+                {
+                    "name": "Alice",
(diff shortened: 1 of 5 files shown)


tests/test_http_graphql_query.py
```diff
@@ -3227,6 +3227,207 @@ def test_graphql_functional_order_04(self):
             ]
         })
 
+    def test_graphql_functional_exists_01(self):
+        self.assert_graphql_query_result(r"""
+            query {
+                User(
+                    filter: {profile: {exists: true}}
```
1st1, line 3234: Can you add a test where there's a link `exists`?

vpetrovykh:
I actually ran into a weird bug here it has to do with compiling `{exists: true}` into `EXISTS expr` and `{exists: false}` into `NOT EXISTS expr`. When the literal `true` or `false` in GraphQL is replaced with a variable (which apparently happens by default at the moment) the resulting EdgeQL query + variables has an unused boolean variable. This causes a problem when the variable corresponding to `exists` changes value, but the underlying EdgeQL doesn't get recompiled, leaving the "original" `EXISTS` or `NOT EXISTS` in place and so getting the result intermittently wrong (depending on the cached GraphQL history).

I fixed that by compiling into `EXISTS (path_expr) = <bool>$var` this way using a variable for `exists` works out correctly because the EdgeQL query handles the variable change without special case recompiling.

Regarding docs, though, why are GraphQL sections apparently hidden in the [documentation site](https://edgedb.com/docs/clients/99_graphql/index)? It's possible to search and find "GraphQL Mutations" and other sections via search, but not directly like other clients' docs. Was it hidden on purpose? I forget if we basically wanted to move it away from the main docs or not.

tests/test_http_graphql_query.py
vpetrovykh: There are link/property names that will fail to produce a GraphQL reflection, such as `and`, `or`, `not`. The name `exists` falls into the same category. We currently don't have specific tests for those either. There are probably weirder name clashes possible that cannot be resolved.

Should I add a bunch of test schemas for names that are legal in EdgeDB, but fail in our GraphQL reflection?

1st1: Hm. Maybe *our* graphql "clauses" should be prefixed with `__`?  How does Hasura deal with this?

1st1:
> Regarding docs, though, why are GraphQL sections apparently hidden in the documentation site? It's possible to search and find "GraphQL Mutations" and other sections via search, but not directly like other clients' docs. Was it hidden on purpose? I forget if we basically wanted to move it away from the main docs or not.

Seems to be a regression. Try to debug it.

tests/test_http_graphql_query.py
elprans: Hasura only allows operators on fields, and since scalars have no properties there's no conflict.  Our issue is that we allow operating on objects and so operator names can clash with property names.

> Maybe our graphql "clauses" should be prefixed with `__`

This will inconvenience 99.9% of users for the benefit of 0.1%.  I think it's fine if we reflect properties that conflict with GraphQL "operator" keywords with a `_` suffix, e.g. `filter_` and couple that with good error messages and documentation.

1st1: Maybe there's also way to use `@annotations` to disambiguate? (I don't quite recall where they can appear though)

vpetrovykh:
> > Regarding docs, though, why are GraphQL sections apparently hidden in the documentation site? It's possible to search and find "GraphQL Mutations" and other sections via search, but not directly like other clients' docs. Was it hidden on purpose? I forget if we basically wanted to move it away from the main docs or not.
> 
> Seems to be a regression. Try to debug it.

I'll merge this change (which contains the actual documentation updates) and look into why the site is not rendering that section separately (since it might be a site issue and not the docs issue as such).

PR #1673, opened 2020-08-17
Fix quoting of identifiers in SQL.

### LINKED ISSUE #1667: DDL results in SQL syntax error
- EdgeDB Version: EdgeDB 1.0-alpha.4+dev.130.g9e96c03f
- OS Version: LInux

Steps to Reproduce:

Try this in REPL:
```
CREATE TYPE Foo {CREATE LINK left -> Foo};
```

I get an error:
```
  Server traceback:
      edb.errors.InternalServerError: syntax error at or near "left"
```

Note that this also happens for `rgiht`. It appears that the identifier `left` is not quoted and clashes with a SQL keyword.

vpetrovykh:
EdgeQL identifiers that are the same as SQL keywords should be quoted.
Unreserved SQL keywords are exempt from this rule.

Fixes #1667.

### DIFF
--- a/edb/pgsql/common.py
+++ b/edb/pgsql/common.py
@@ -81,7 +81,10 @@ def needs_quoting(string):
                string.replace('_', 'a').isalnum())
     return (
         not isalnum or
-        string.lower() in pg_keywords.by_type[pg_keywords.RESERVED_KEYWORD] or
+        string.lower() in pg_keywords.by_type[
+            pg_keywords.RESERVED_KEYWORD] or
+        string.lower() in pg_keywords.by_type[
+            pg_keywords.TYPE_FUNC_NAME_KEYWORD] or
         string.lower() != string
     )
 
--- a/tests/test_edgeql_ddl.py
+++ b/tests/test_edgeql_ddl.py
@@ -879,6 +879,21 @@ class TestEdgeQLDDL(tb.DDLTestCase):
                 ALTER TYPE Child ALTER LINK target DROP OWNED;
             """)
 
+    async def test_edgeql_ddl_27(self):
+        # Test that identifiers that are SQL keywords get quoted.
+        # Issue 1667
+        await self.con.execute("""
+            CREATE TYPE test::Foo {
+                CREATE PROPERTY left -> str;
+                CREATE PROPERTY smallint -> str;
+                CREATE PROPERTY natural -> str;
+                CREATE PROPERTY null -> str;
+                CREATE PROPERTY `like` -> str;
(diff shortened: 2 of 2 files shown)


edb/pgsql/common.py
```diff
@@ -81,7 +81,10 @@ def needs_quoting(string):
                string.replace('_', 'a').isalnum())
     return (
         not isalnum or
-        string.lower() in pg_keywords.by_type[pg_keywords.RESERVED_KEYWORD] or
+        string.lower() in pg_keywords.by_type[
+            pg_keywords.RESERVED_KEYWORD] or
+        string.lower() in pg_keywords.by_type[
+            pg_keywords.TYPE_FUNC_NAME_KEYWORD] or
```
elprans, line 87: There's also `COL_NAME_KEYWORD`.

vpetrovykh: Oh, I tried that and I got this during initialization:
```
2. asyncpg.exceptions.UndefinedObjectError: type smallint does not exist


    query, line 127, in query
        | 
        > CREATE  FUNCTION edgedb._derive_uuid(id uuid, variant "smallint")
          ^
        | RETURNS  uuid
...
```

vpetrovykh: Incidentally, creating a type or a link with name `smallint` doesn't cause any problems (maybe I should add that as another test case).

elprans: Sigh, this is crazy.  They basically hardcoded a bunch of alternative spellings for types as keywords and even though you can do `"int2"`, you cannot do `"smallint"`.  Makes little sense to me, but fine, not quoting `COL_NAME_KEYWORD` is actually fine, since the only place where they require quoting in SQL are unqualified function names, and all our function calls are fully-qualifed.

tests/test_edgeql_ddl.py
```diff
@@ -879,6 +879,13 @@ async def test_edgeql_ddl_26(self):
                 ALTER TYPE Child ALTER LINK target DROP OWNED;
             """)
 
+    async def test_edgeql_ddl_27(self):
+        # Test that identifiers that are SQL keywords get quoted.
+        # Issue 1667
+        await self.con.execute("""
+            CREATE TYPE test::Foo {CREATE LINK left -> test::Foo};
```
1st1, line 886: Can you add few more links, like a edgeql-quoted `CREATE` and `LINK`, for example.

vpetrovykh: OK, added a few more names.

PR #1684, opened 2020-08-20
A round of fixes to the migration system

elprans:
This fixes a bunch of infidelities in the Delta -> DDL pipeline.  The fixes
are tested by the existing test suite which is ran with the
(introduced in this commit) `migrations_via_ddl` debug flag, which verifies
that the migration DDL text produces a valid schema.


docs/edgeql/ddl/migrations.rst
```diff
@@ -230,28 +230,33 @@ in the current migration block in the specified output format:
     .. code-block::
 
         {
-          // List of confirmed migration
-          // statements.
+          // Name of the parent migration
```
1st1, line 233: (should be `.. code-block:: javascript`)

edb/common/debug.py
```diff
@@ -123,6 +123,9 @@ class flags(metaclass=FlagsMeta):
     log_metrics = Flag(
         doc="Log verbose statistics on connections and compiler behavior.")
 
+    migrations_via_ddl = Flag(
+        doc="Always use generated DDL when running migrations.")
```
1st1, line 127: use -> apply or execute?

docs/edgeql/ddl/migrations.rst
elprans: I tried that, it errors, because it's not really valid JS.

edb/common/debug.py
elprans: No, "use" is more appropriate, because currently we generate the DDL, but use the source delta tree instead.  Regardless, this is a temporary flag until we fix all current testsuite failures and enable this behavior by default.

edb/schema/delta.py
```diff
@@ -145,12 +143,21 @@ def delta_objects(
             ),
         )
 
-    for p in altered:
-        delta.add(p)
+    delta.update(alters)
+
+    deleted_order: Iterable[so.Object]
+    deleted = old - {y for _, (s, y) in comparison_map.items() if s > 0.6}
```
1st1, line 149: Maybe make 0.6 (that's used in a few places) a constant with a comment?

PR #1687, opened 2020-08-21
Fix broken errors in some correlated set cases

msullivan:
InvalidScopeConfiguration carries references to ScopeTreeNodes in it,
which causes grief because those fail to pickle due to the weakrefs in
them. (Though pickling them doesn't seem like the right thing to do
even if it wasn't for the weakref.)

Handle this in the same way that it is handled for the current cases
of this message that work: by catching it and reraising. (See
`register_set_in_scope`).

Other solutions to this might be better: we could just drop the nodes
from InvalidScopeConfiguration, since it doesn't seem like we ever
make use of them in any way.

Catching and rethrowing has the advantage that we can provide a better
source context, though, since one does not seem to be available in the
places InvalidScopeConfiguration is thrown. (Though that could also be
changed.)

### DIFF
--- a/edb/edgeql/compiler/stmt.py
+++ b/edb/edgeql/compiler/stmt.py
@@ -779,12 +779,9 @@ def compile_Shape(
             f'shapes cannot be applied to '
             f'{expr_stype.get_verbosename(ctx.env.schema)}'
         )
-    try:
-        view_type = viewgen.process_view(
-            stype=expr_stype, path_id=expr.path_id,
-            elements=shape.elements, ctx=ctx)
-    except irast.InvalidScopeConfiguration as e:
-        raise errors.QueryError(e.args[0], context=shape.context) from e
+    view_type = viewgen.process_view(
+        stype=expr_stype, path_id=expr.path_id,
+        elements=shape.elements, parser_context=shape.context, ctx=ctx)
 
     return setgen.ensure_set(expr, type_override=view_type, ctx=ctx)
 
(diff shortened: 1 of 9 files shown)


edb/edgeql/compiler/stmt.py
```diff
@@ -779,9 +779,12 @@ def compile_Shape(
...
             f'{expr_stype.get_verbosename(ctx.env.schema)}'
         )
-    view_type = viewgen.process_view(
-        stype=expr_stype, path_id=expr.path_id,
-        elements=shape.elements, ctx=ctx)
+    try:
+        view_type = viewgen.process_view(
+            stype=expr_stype, path_id=expr.path_id,
+            elements=shape.elements, ctx=ctx)
+    except irast.InvalidScopeConfiguration as e:
```
elprans, line 786: `InvalidScopeConfigration` gets thrown by `ScopeNode.attach_path()`, so perhaps it'll be better to invent a wrapper around it that takes source context as argument and reraises?

msullivan: Should we just drop InvalidScopeConfiguration (or at least drop the node arguments and add source context), and pass in source contexts on all the call sites? There don't seem to be that many.

elprans: The original motivation behind a bespoke exception was to keep `scopetree` entirely self-contained and unit-testable, and the intended purpose of the node arguments was to handle it and make the error message a bit more helpful by finding the source context for the other appearance of the symbol.  Alas, those aspirations didn't really materialize yet.  Let's drop the node arguments.

1st1:
> Other solutions to this might be better: we could just drop the nodes
from InvalidScopeConfiguration, since it doesn't seem like we ever
make use of them in any way.

@elprans Can we do that?

msullivan:
@elprans ping

edb/errors/__init__.py
```diff
@@ -152,6 +152,10 @@ class InvalidPropertyTargetError(InvalidTargetError):
     _code = 0x_04_02_01_02
 
 
+class InvalidScopeConfigurationError(InvalidTypeError):
```
elprans, line 155: Probably not worth an explicit public error class, let's use `InvalidReferenceError` instead.

1st1: Also we don't manually add types to this file. That's done via modifying `edb/api/errors.txt` and running `edb gen-errors`, as the comment on top of the file suggests.

> Probably not worth an explicit public error class, let's use InvalidReferenceError instead.

Maybe call it `ScopeError` and add it?

elprans: I see no point in being hyper-granular in query error classes.  `InvalidReferenceError` is currently raised for cases where you refer to something that doesn't exist, or exists and is of a wrong type. In this case, the object you're referencing exists, but you can't refer to it in that scope.

PR #1696, opened 2020-08-25
Fix derived name mangling in schema

elprans:
The current mangling schema isn't properly reversible, and we actually rely
on it being so.  Fix this by making sure that
`unmangle(mangle(X)) == X` always holds regardless of the depth of 
mangling.

This creates an incompatibility with earlier dumps, because we carry an 
internal `name -> id` map in them.  Fortunately, it's easy to fix by 
retaining the old mangling code and using it to lookup in schema id maps 
produced by older servers.

*NOTE*: the dump tests are failing because the current version derivation is wrong,
and it seemed wrong to work that around in code.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -19,6 +19,11 @@ jobs:
     steps:
     - uses: actions/checkout@v2
       with:
+        fetch-depth: 0
+        submodules: false
+
+    - uses: actions/checkout@v2
+      with:
         fetch-depth: 50
         submodules: true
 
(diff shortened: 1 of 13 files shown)


1st1:
> NOTE: the dump tests are failing because the current version derivation is wrong,

Do tests pass locally?

elprans:
> Do tests pass locally?

They do if I munge the local version to say `a5.dev...`

1st1:
> They do if I munge the local version to say a5.dev...

Not sure I follow, but we need to make sure that the CI and running tests locally works after merging this, obviously.

Issue #1681, opened 2020-08-19
migrations: Rejecting proposed statement doesn't work


tailhook:
Here is the log:
```
[2020-08-19T12:49:59Z DEBUG edgedb::migrations::create] Executing `DESCRIBE CURRENT MIGRATION AS JSON`
Following DDL statements will be applied:
    CREATE TYPE default::Type3 {
        CREATE OPTIONAL SINGLE PROPERTY field3 -> std::str;
    };
Apply the DDL statements? [y,n,l,c,b,s,q,?]
no
[2020-08-19T12:50:00Z DEBUG edgedb::migrations::create] Executing `ALTER CURRENT MIGRATION REJECT PROPOSED`
Error: ERROR: InternalServerError: local variable 'query' referenced before assignment
  Hint: This is most likely a bug in EdgeDB. Please consider opening an issue ticket at https://github.com/edgedb/edgedb/issues/new?template=bug_report.md
  Server traceback:
      Traceback (most recent call last):
        File "/work/edb/server/procpool/worker.py", line 75, in worker
          res = await meth(*args)
        File "/work/edb/server/compiler/compiler.py", line 1733, in compile_eql_tokens_in_tx
          return self._compile(ctx=ctx, tokens=eql_tokens)
        File "/work/edb/server/compiler/compiler.py", line 1348, in _compile
          comp: dbstate.BaseQuery = self._compile_dispatch_ql(ctx, stmt)
        File "/work/edb/server/compiler/compiler.py", line 1271, in _compile_dispatch_ql
          return self._compile_ql_migration(ctx, ql)
        File "/work/edb/server/compiler/compiler.py", line 1010, in _compile_ql_migration
          return query
      UnboundLocalError: local variable 'query' referenced before assignment
```

PR #1697, opened 2020-08-25
Initial implementation of ALTER CURRENT MIGRATION REJECT PROPOSED

elprans:
This adds the capability of making "guided" schema diffs, where certain 
predicted operations have been prohibited.

Fixes: #1681

### DIFF
--- a/edb/schema/delta.py
+++ b/edb/schema/delta.py
@@ -40,7 +40,7 @@ from edb.common import verutils
 from edb.edgeql import ast as qlast
 from edb.edgeql import compiler as qlcompiler
 from edb.edgeql import qltypes
-
+from edb.edgeql import quote as qlquote
 
 from . import expr as s_expr
 from . import name as sn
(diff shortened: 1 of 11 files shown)


1st1:
There are two unrelated commits in this PR. Please rebase

elprans:
Rebased.

edb/schema/delta.py
```diff
@@ -94,7 +96,13 @@ def delta_objects(
 
         full_matrix.append((x, y, similarity))
 
-    full_matrix.sort(key=lambda v: v[2], reverse=True)
+    full_matrix.sort(
+        key=lambda v: (
+            1.0 - v[2],
```
1st1, line 110: Just curious, why is this better than reversed=True?

elprans: Because sorting names in reverse is kinda weird, esp. for tests.

tests/test_edgeql_ddl.py
```diff
@@ -6132,3 +6136,98 @@ async def test_edgeql_ddl_describe_migration_json_03(self):
...
+            'confirmed': ['ALTER TYPE test::Test RENAME TO test::Test3'],
+            'complete': False,
+            'proposed': {
+                'statements': [{
+                    'text': 'CREATE TYPE test::Test2;',
+                }],
+                'confidence': 1.0,
+                'operation_id': 'CREATE TYPE test::Test2',
+                'prompt': (
+                    "did you create object type 'test::Test2'?"
```
1st1, line 6222: I'm not sure I following the test. There was the 'test::Test' type. We then migrated to 'test::Test2' and 'test::Test3' types, and confirmed that we renamed 'Test' to 'Test3'.  Why do we ask about 'Test2' now?

elprans: Because that's how `DESCRIBE CURRENT MIGRATION` works.  It's the decision of the frontend whether or not to show the actual prompt, or just accept the `CREATE` (this is driven by `confidence` and `safe`).

1st1: What happens if I reject this suggestion?

elprans: The suggested list will be empty and the "complete" key will also be `false`, which would indicate to the frontend that the server has exhausted all its suggestions and the user must complete the migration manually (or rollback to an earlier point).

PR #1698, opened 2020-08-25
Replace default SCM version with a custom scheme

ambv:
This changes behavior from the default in setuptools-scm which is to use `git describe`.  The problem with `git describe` is that on a given branch it describes the HEAD commit using the latest tag *on that branch* or *in a merged branch*.  When, say, the `releases/v1.0a4` branch is cut off and later only a few commits are cherry-picked from master, and one of those commits gets tagged as `v1.0a4`, this tag never appears on `master` again.  And thus, `git describe` never properly recognizes this as the latest version tag.

This change instead looks at *any* tag in the repository starting with `v`.  If the current HEAD commit is tagged with one of those tags, that's used for the version exactly.  Otherwise, the latest tag is always used and incremented using the smallest possible increment.

Finally, a `dev` sub-version is used which counts all commits on the current branch.  This addresses the issue of it being sortable, while keeping the implementation simple.

By the way, local version suffixes like `+g1478d96c.d20200826` are not changing and are still provided by setuptools-scm's default implementation.

One known limitation of this current implementation is that the latest tagged version is always used for non-exact dev builds.  We are unlikely to build new releases of EdgeDB from old branch commits, and on a release branch if we need another release, we will move the tag, making a release an exact match again.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,6 +22,7 @@ import os
 import os.path
 import pathlib
 import platform
+import re
 import shutil
 import subprocess
 import textwrap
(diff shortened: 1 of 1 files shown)


ambv:
This commit's version is for example: `edgedb-server==1.0a5.dev5020+g67903476`.

PR #1701, opened 2020-08-26
Add examples of CLI usage.

### LINKED ISSUE #1021: Authentication is under documented
First, it took me ~30 minutes to figure out:
```
edgedb --admin configure insert auth --method=scram --priority=1
```
And I'm not sure is priority right.

Second, how does `create role` works? I.e. I can create a role and set a password. I still don't know to give that role access to a given database.

vpetrovykh:
Add examples for configuring a database and then connecting to it using
CLI (similar to the admin commands) to the cheat sheets.

Fixes #1021

### DIFF
--- a/docs/cheatsheet/cli.rst
+++ b/docs/cheatsheet/cli.rst
@@ -1,0 +1,91 @@
+.. _ref_cheatsheet_cli:
+
+CLI
+===
+
+Create a database instance ``my_instance``:
+
+.. code-block:: bash
+
+    $ edgedb server init my_instance
+
+Create a database:
+
+.. code-block:: bash
+
+    $ edgedb -I my_instance create-database my_new_project
+    OK: CREATE
+
+Create a new role (other than default ``edgedb``):
+
+.. code-block:: bash
+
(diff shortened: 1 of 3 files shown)


docs/cheatsheet/cli.rst
```diff
@@ -0,0 +1,85 @@
+.. _ref_cheatsheet_cli:
+
+CLI
+===
+
+Create a database:
+
+.. code-block:: bash
+
+    $ edgedb --admin create-database my_new_project
```
1st1, line 10: I think this is the old way of doing this, no? @elprans

elprans: Yes. `--admin` is deprecated, all examples should use `-I<instance-name>` instead.  Also, we should be recommending creating instances instead of databases for new projects.

vpetrovykh: I've installed edgedb on my Ubuntu laptop from scratch, re-tried the commands and updated the cheat sheet accordingly.

1st1:
@elprans please also take a look

PR #1710, opened 2020-08-27
Add missing qlkind annotation to Index schema object implementation

elprans:
Without this, `DESCRIBE CURRENT MIGRATION` on a migration with indexes
will crash.

### DIFF
--- a/edb/schema/indexes.py
+++ b/edb/schema/indexes.py
@@ -25,6 +25,7 @@ from edb import errors
 from edb.common import ast
 from edb.edgeql import ast as qlast
 from edb.edgeql import compiler as qlcompiler
+from edb.edgeql import qltypes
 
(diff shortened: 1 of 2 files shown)


edb/schema/indexes.py
```diff
@@ -41,7 +42,11 @@
     from . import types as s_types
 
 
-class Index(referencing.ReferencedInheritingObject, s_anno.AnnotationSubject):
+class Index(
+    referencing.ReferencedInheritingObject,
+    s_anno.AnnotationSubject,
+    qlkind=qltypes.SchemaObjectClass.INDEX,
```
1st1, line 48: Can you add some check to the metaclass to require `qlkind` for "final" classes? Perhaps we can add something like `__abstract__ = True` for non-final ones.

PR #1711, opened 2020-08-28
Some minor debugging improvements

msullivan:
Better output for RelRangeVars in debug dumps and fix dump to not
break as badly when its output is taken over by remote-pdb.

### DIFF
--- a/edb/common/markup/renderers/terminal.py
+++ b/edb/common/markup/renderers/terminal.py
@@ -641,7 +641,14 @@ def render(markup, *, ensure_newline=True, file=None, renderer=Renderer):
     if file is None:
         file = sys.stdout
 
-    fileno = file.fileno()
+    try:
+        fileno = file.fileno()
+    except OSError:
+        # This is a hack to try to get nice colorized dump output over
+        # a remote-pdb connection. If the output is redirected to
+        # something without fileno, use what ought to be stdout's fileno
+        # to decide on color, etc.
+        fileno = 1
     max_width = term.size(fileno)[1]
 
     style_table = None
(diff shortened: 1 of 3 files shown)


edb/pgsql/ast.py
```diff
@@ -205,6 +205,12 @@ def query(self) -> BaseRelation:
         else:
             return self.relation
 
+    def __repr__(self) -> str:
+        return (
+            f'<pg.{self.__class__.__name__} '
+            f'name={self.relation.name!r} at 0x{id(self):x}>'
```
1st1, line 211:
```suggestion
            f'name={self.relation.name!r} at {id(self):#x}>'
```

:)

edb/common/markup/renderers/terminal.py
```diff
@@ -641,7 +641,14 @@ def render(markup, *, ensure_newline=True, file=None, renderer=Renderer):
...
 
-    fileno = file.fileno()
+    try:
+        fileno = file.fileno()
+    except OSError:
+        # This is a hack to try to get nice colorized dump output over
+        # a remote-pdb connection. If the output is redirected to
+        # something without fileno, use what ought to be stdout's fileno
+        # to decide on color, etc.
+        fileno = 1
```
1st1, line 651: I'm not so sure about this one, but let's see if it works.

msullivan: Yeah that's about where I am on it.

1st1: Can you use `sys.stdout.fileno()` instead of hardcoding 1?

msullivan: No, because the whole thing being hacked around here is that sys.stdout has been rebound to a TextIO by the remote debugger or something. Arguably the "right" thing to do here is to drop this hack and just output an unstyled dump, but I really do want the color in this case...

PR #1712, opened 2020-08-28
Add an env variable to force using a cache even if the keys mismatch

msullivan:
This is useful for speeding up local testing in some cases and for
making it possible to run unit tests when there is an issue that
causes bootstrapping to fail (since failing unit tests are a lot
easier to narrow down).

I can rename the environment variable if somebody has a name they
prefer.

### DIFF
--- a/edb/common/debug.py
+++ b/edb/common/debug.py
@@ -75,8 +75,8 @@ class flags(metaclass=FlagsMeta):
     bootstrap = Flag(
         doc="Debug server catalog bootstrap.")
 
-    cache_yolo = Flag(
-        doc="Disable consistency check.")
+    bootstrap_cache_yolo = Flag(
+        doc="Disable bootstrap cache consistency check.")
 
     edgeql_parser = Flag(
         doc="Debug EdgeQL parser (rebuild grammar verbosly).")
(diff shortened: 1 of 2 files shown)


edb/server/buildmeta.py
```diff
@@ -140,7 +140,7 @@ def read_data_cache(
     if full_path.exists():
         with open(full_path, 'rb') as f:
             src_hash = f.read(len(cache_key))
-            if src_hash == cache_key:
+            if src_hash == cache_key or os.getenv('EDGEDB_CACHE_YOLO') == '1':
```
elprans, line 143: Please add it to `common.debug.flags` for discoverability.

edb/common/debug.py
```diff
@@ -75,6 +75,9 @@ class flags(metaclass=FlagsMeta):
     bootstrap = Flag(
         doc="Debug server catalog bootstrap.")
 
+    cache_yolo = Flag(
```
elprans, line 78: perhaps `bootstrap_cache_yolo` or do you mean for this to apply to all cache everywhere?

msullivan: Good question. Is `edb.server.buildmeta.read_data_cache` used for other caches? (I don't really know what all caches we have.)

elprans: `read_data_cache` is only used for stdlib cache in devmode (server and also serverless schema tests). I think it's fine to name it `bootstrap_cache_yolo`

Issue #1715, opened 2020-08-28
migration defining a function crashes with "InternalServerError: list index out of range"


CodesInChaos:
- EdgeDB Version: alpha5
- OS Version: ubuntu 20

Run the following migration:
```
START MIGRATION TO {
module default {
	function canonicalize_name(s:str) -> str
		using (s);
}
};
POPULATE MIGRATION;
COMMIT MIGRATION;
```

it crashes with:

> ERROR: InternalServerError: list index out of range
>   Hint: This is most likely a bug in EdgeDB. Please consider opening an issue ticket at https://github.com/edgedb/edgedb/issues/new?template=bug_report.md
>   Server traceback:
>       Traceback (most recent call last):
>         File "/usr/lib/x86_64-linux-gnu/edgedb-server-1-alpha5/lib/python3.8/site-packages/edb/pgsql/compiler/__init__.py", line 83, in compile_ir_to_sql_tree
>           qtree = dispatch.compile(ir_expr, ctx=ctx)
>         File "/usr/lib/x86_64-linux-gnu/edgedb-server-1-alpha5/lib/python3.8/functools.py", line 875, in wrapper
>           return dispatch(args[0].__class__)(*args, **kw)
>         File "/usr/lib/x86_64-linux-gnu/edgedb-server-1-alpha5/lib/python3.8/site-packages/edb/pgsql/compiler/expr.py", line 67, in compile_Set
>           return output.top_output_as_value(ctx.rel, ir_set, env=ctx.env)
>         File "/usr/lib/x86_64-linux-gnu/edgedb-server-1-alpha5/lib/python3.8/site-packages/edb/pgsql/compiler/output.py", line 684, in top_output_as_value
>           arg=stmt.target_list[0].val,
>       IndexError: list index out of range
>       
>       The above exception was the direct cause of the following exception:
>       
>       Traceback (most recent call last):
>         File "/usr/lib/x86_64-linux-gnu/edgedb-server-1-alpha5/lib/python3.8/site-packages/edb/server/procpool/worker.py", line 75, in worker
>           res = await meth(*args)
>         File "/usr/lib/x86_64-linux-gnu/edgedb-server-1-alpha5/lib/python3.8/site-packages/edb/server/compiler/compiler.py", line 1816, in compile_eql_tokens_in_tx
>           return self._compile(ctx=ctx, tokens=eql_tokens)
[...]

PR #1723, opened 2020-08-28
Fix compilation of functions with non-statement bodies

elprans:
The current top-level output handling assumes that the expression is
always a statement, but that wouldn't always be the case when compiling
the body of a function like the following:

    CREATE FUNCTION identity(arg: str) -> str USING (arg);

Fixes: #1715

### DIFF
--- a/edb/pgsql/compiler/pathctx.py
+++ b/edb/pgsql/compiler/pathctx.py
@@ -791,9 +791,8 @@ def get_path_output(
         ptr_info: Optional[pg_types.PointerStorageInfo]=None,
         env: context.Environment) -> pgast.OutputVar:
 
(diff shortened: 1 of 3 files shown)


edb/pgsql/compiler/pathctx.py
```diff
@@ -959,6 +959,10 @@ def get_path_serialized_output(
     # must be kept outside of get_path_output() generic.
     aspect = 'serialized'
 
+    view_path_id_map = getattr(rel, 'view_path_id_map', None)
```
1st1, line 962: Untyped?

elprans: Hm. It's a copypasta from a similar method. Might actually not be needed anymore.

Issue #1688, opened 2020-08-21
Restrict the usage of DML in DDL and within shapes.


vpetrovykh:
There are some limitations to using DML (`INSERT`, `DELETE`, `UPDATE`)
statements as part of other expressions:

* They cannot appear as branches of conditional expressions (#684, #1639).
* They shouldn't appear inside function bodies (although this
  limitation can be relaxed in the future #1661).
* They cannot be correlated with other query components, basically
  they must always be independent of the rest of the query in such a
  way that refactoring them into a `WITH` block doesn't change the
  semantics.

The last point results in some additional restrictions that need to be enforced:

* DML cannot appear as a computable in a shape
* DML cannot appear as a computable in a type or aliased type
  declaration at all (even in a `WITH` block), because that
  potentially results in issues with correlation of the DML and the
  rest of the object whenever that type appears in a query.

It is, however, legal to use DML in a `default` expression, for example:

```
type Foo {
    required property val -> str;
}

type Bar {
    required link foo -> Foo {
        default := (INSERT Foo)
    }
}
```

Which means that if `foo` is omitted at creation time, a new `Foo`
object will be created automatically for that link.

elprans:
@vpetrovykh Is this done?

vpetrovykh:
This appears to be done. Special case of possibly allowing DML in functions is a separate issue #1661

PR #1726, opened 2020-08-29
Forbid usage of DML in some special cases.

vpetrovykh:
DML is forbidden in shapes, aliased types, and computable definitions.

Issue #1688

### DIFF
--- a/edb/edgeql/compiler/options.py
+++ b/edb/edgeql/compiler/options.py
@@ -66,6 +66,14 @@ class GlobalCompilerOptions:
     #: definitions.
     func_params: Optional[s_func.ParameterLikeList] = None
 
+    #: The name that can be used in a "DML is disallowed in ..."
+    #: error. When this is not None, any DML should cause an error.
+    in_ddl_context_name: Optional[str] = None
+
+    #: Sometimes (like in queries compiled form GraphQL) it may be OK
+    #: to contain DML in the top-level shape computables.
+    allow_top_level_shape_dml: bool = False
+
 
 @dataclass
 class CompilerOptions(GlobalCompilerOptions):
--- a/edb/schema/pointers.py
+++ b/edb/schema/pointers.py
@@ -870,6 +870,7 @@ class PointerCommandOrFragment(
 
         source = schema.get(source_name, type=s_objtypes.ObjectType)
 
+        ptr_name = self.get_displayname()
         expression = s_expr.Expression.compiled(
             s_expr.Expression.from_ast(expr, schema, context.modaliases),
             schema=schema,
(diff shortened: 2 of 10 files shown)


edb/edgeql/compiler/options.py
```diff
@@ -66,6 +66,14 @@ class GlobalCompilerOptions:
     #: definitions.
     func_params: Optional[s_func.ParameterLikeList] = None
 
+    #: The name that can be used in a "DML is disallowed in ..."
+    #: error. When this is not None, any DML should cause an error.
+    in_ddl: Optional[str] = None
```
1st1, line 71: I'd rename to something like

```suggestion
    in_ddl_context_name: Optional[str] = None
```

edb/edgeql/compiler/options.py
```diff
@@ -66,6 +66,14 @@ class GlobalCompilerOptions:
...
+
+    #: The query comes from compiling GraphQL, so it may contain DML
+    #: in the top-level shape computables.
+    from_graphql: bool = False
```
1st1, line 75: I wouldn't expose any of graphql requirements or semantics to the compiler, so I'd give it a more generic name

```suggestion
    allow_top_level_shape_dml: bool = False
```

edb/edgeql/compiler/options.py
vpetrovykh: Renamed

edb/edgeql/compiler/options.py
vpetrovykh: Renamed

edb/edgeql/compiler/viewgen.py
```diff
@@ -613,6 +613,8 @@ def _normalize_view_ptr_expr(
                 is_update=is_update,
             )
 
+            # Mutation shapes don't obey the same rules as select shapes.
+            shape_expr_ctx.in_shape = not is_mutation
```
elprans, line 617: This seems superfluous.  Use `defining_view` below instead: `ctx.defining_view is not None and ctx.defining_view.get_expr_type() is s_types.ExprType.Select`

edb/schema/pointers.py
```diff
@@ -896,6 +896,8 @@ def _parse_computable(
 
         source = schema.get(source_name, type=s_objtypes.ObjectType)
 
+        ptr_name = sn.shortname_from_fullname(
```
elprans, line 899: Please don't open-code object names, use standard helpers: `get_displayname_static()`, `get_verbosename_static()` etc. For this case you can add a shortcut similar to `get_verbosename()` on `ObjectCommand` to get `displayname`.

Issue #1730, opened 2020-08-29
Can't create tuple property containing custom scalar


CodesInChaos:
- EdgeDB Version: nightly (1.0a6.dev5038+g7cd6dcc4)
- OS Version: ubuntu 20

```
START MIGRATION TO {
module default {
	scalar type MyScalar extending str;

	type User {
		required property tup -> tuple<x:MyScalar>;
	}
}
};
POPULATE MIGRATION;
COMMIT MIGRATION;
```

fails with 
> ERROR: InvalidReferenceError: type 'default::MyScalar' does not exist

elprans:
Yes, the dependency resolver gets it wrong in this case.

PR #1731, opened 2020-08-30
Fix handling of collections of newly created types

elprans:
When a collection type is defined in terms of one or more *new* types
that were defined in the same migration, the DDL compilation would crash
because collection types currently assume that their consituent types
exist at the time of DDL AST processing, which isn't always true.

Fix this by not attempting to resolve component types too early.  Also,
while at it, clean up the implementation of implicit collection type
creation and cleanup by using conditional commands instead of the
current `ensure_collection_type` hack.

Fixes: #1730

### DIFF
--- a/edb/schema/casts.py
+++ b/edb/schema/casts.py
@@ -153,7 +153,7 @@ def get_cast_fullname(
     from_type: s_types.TypeShell,
     to_type: s_types.TypeShell,
 ) -> sn.Name:
-    quals = [from_type.name, to_type.name]
+    quals = [from_type.get_name(schema), to_type.get_name(schema)]
     shortname = sn.Name(f'{module}::cast')
     return sn.Name(
         module=shortname.module,
@@ -245,6 +245,18 @@ class CastCommand(sd.QualifiedObjectCommand[Cast],
 
         return get_cast_fullname(schema, 'std', from_type, to_type)
 
+    def canonicalize_attributes(
+        self,
+        schema: s_schema.Schema,
+        context: sd.CommandContext,
+    ) -> s_schema.Schema:
+        schema = super().canonicalize_attributes(schema, context)
+        schema = s_types.materialize_type_in_attribute(
+            schema, context, self, 'from_type')
+        schema = s_types.materialize_type_in_attribute(
+            schema, context, self, 'to_type')
+        return schema
+
 
 class CreateCast(CastCommand, sd.CreateObject[Cast]):
     astnode = qlast.CreateCast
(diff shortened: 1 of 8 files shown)


CodesInChaos:
Isn't this actually fixing #1730?

elprans:
> Isn't this actually fixing #1730?

Good catch.  Yes, I mislabeled this.

edb/schema/casts.py
```diff
@@ -153,7 +153,7 @@ def get_cast_fullname(
     from_type: s_types.TypeShell,
     to_type: s_types.TypeShell,
 ) -> sn.Name:
-    quals = [from_type.name, to_type.name]
```
1st1, line 156: Hm, why does `.name` even work? Maybe disable it in favor of `.get_name()`?

edb/schema/casts.py
```diff
@@ -245,6 +245,18 @@ def _classname_from_ast(
 
         return get_cast_fullname(schema, 'std', from_type, to_type)
 
+    def canonicalize_attributes(
+        self,
+        schema: s_schema.Schema,
+        context: sd.CommandContext,
+    ) -> s_schema.Schema:
```
1st1, line 252: Add a docstring? This looks like a new method added to a few places

edb/schema/delta.py
```diff
@@ -2477,6 +2477,10 @@ def apply(
 class DeleteObject(ObjectCommand[so.Object_T], Generic[so.Object_T]):
     _delta_action = 'delete'
 
+    #: If True, apply the command only if the object has no referrers
+    #: in the schema.
+    if_unused = struct.Field(bool, default=False)
```
1st1, line 2482: I'd rename to `only_if_unused` or even more verbose `apply_only_if_unused`. This isn't the most common thing so verbosity isn't going to slow down coding much.

edb/schema/casts.py
1st1: I assume this is for collection types, right?

edb/schema/delta.py
elprans: This naming is consistent with other `if_` conditions on commands: `if_not_exists`, `if_exists`, etc.  Should we rename those also?

edb/schema/casts.py
elprans: It's not a new method, just an overload of an existing one.  I'll make sure there is a comment on the main definition, but I don't think there is a point in copy-pasting the docstring into every overload.

edb/schema/casts.py
elprans: > Hm, why does `.name` even work? Maybe disable it in favor of `.get_name()`?

Historical.  It makes sense to remove direct attribute access, but I'd rather do in in a followup PR.

edb/schema/delta.py
1st1: I'd rename to `apply_if_unused`, `apply_if_exists`, etc. /2c

elprans: OK, I'll follow up separately.

Issue #1699, opened 2020-08-26
Fix volatile keys with SELECT UNLESS CONFLICT ELSE


msullivan:
Follow up to #1639.

Currently we duplicate the constraint expression in both the INSERT and the SELECT generated in the ELSE branch. If the expression is volatile, it might evaluate to different values in each branch, which is bad.

We need to hoist the computation out, at least when it is nontrivial.

1st1:
@msullivan Has this been fixed?

msullivan:
We disallow volatile keys there now, so we can't miscompile anything, but we don't *support* it, which we should.

PR #1735, opened 2020-08-31
Overhaul how FOR interacts with iterators

### LINKED ISSUE #1674: FOR ... UNION (UPDATE ...) broken in some cases
`FOR noob in {"Alice", "Bob"} UNION (UPDATE User FILTER .name = tag SET {name:=noob++"!"});`
produces
`Error: missing FROM-clause entry for table "str~4"`

Probably fixing this is a matter of applying the same techniques used for INSERT.

### LINKED ISSUE #1686: Directly nested FOR ... FOR ... INSERT broken
`FOR a in {"a", "b"} UNION (FOR b in {"c", "d"} UNION (INSERT Note {name:=b}));` only does two inserts (though will return each object twice).

`FOR a in {"a", "b"} UNION (FOR b in {a++"c", a++"d"} UNION (INSERT Note {name:=b}));` generates invalid SQL.

The issue is that only the inner iterator gets lifted into a CTE.

### LINKED ISSUE #1724: Loosen restrictions/fix compilation on correlation of FOR variables with DML
Currently we disallow queries like
```
FOR a in {"a", "b"} UNION (
    FOR b in {"a", "b"} UNION (
            INSERT Note {name := a++b}));
```
and
```
FOR a in {"foo", "bar"} UNION (
    SELECT (INSERT Note {name:=a}) {name});
```
with "cannot reference correlated set" errors.

And then there are some related examples that I had thought we banned but actually we just (I think) miscompile:
```
FOR a in {"foo", "bar"} UNION (
    (a, (INSERT Note {name:=a})));
```
This produces a cross product with 4 results when I think that is should only produce 2.

msullivan:
We now always compile FOR iterators when they are seen at SELECT
statements, rather than looking back up the tree to find them when we
encounter them at DML statements. At the compilation site, we check
whether there is DML in the body and hoist the iterator to a CTE if
so.

This fixes some issues involving nested iterators and mixed SELECT/DML
inside iterators and allows us to loosen the restrictions on
correlation of iterator variables.

It should make fixing #1699 (volatile keys with UNLESS CONFLICT ELSE)
practical. Hopefully it also clears the way for additional refactors.

Closes #1724. Closes #1674. Closes #1686.

### DIFF
--- a/edb/ir/utils.py
+++ b/edb/ir/utils.py
@@ -290,3 +290,12 @@ def get_iterator_sets(stmt: irast.Stmt) -> Sequence[irast.Set]:
         iterators.extend(stmt.hoisted_iterators)
 
     return iterators
+
+
+def contains_dml(stmt: irast.Stmt) -> bool:
+    """Check whether a statement contains any DML in a subtree."""
+    # If this ends up being a perf problem, we can use a visitor
+    # directly and cache.
+    res = ast.find_children(stmt, lambda x: isinstance(x, irast.MutatingStmt),
+                            terminate_early=True)
+    return bool(res)
--- a/edb/pgsql/compiler/dml.py
+++ b/edb/pgsql/compiler/dml.py
@@ -148,7 +148,7 @@ def init_dml_stmt(
 
     pathctx.put_path_bond(ctx.rel, ir_stmt.subject.path_id)
     if ctx.enclosing_cte_iterator:
-        pathctx.put_path_bond(ctx.rel, ctx.enclosing_cte_iterator.set.path_id)
+        pathctx.put_path_bond(ctx.rel, ctx.enclosing_cte_iterator.path_id)
 
     return DMLParts(
         dml_ctes=dml_map,
@@ -309,13 +309,12 @@ def wrap_dml_cte(
     return dml_rvar
 
 
-def update_iterator_cte(
-        ir: irast.Set,
-        cte: pgast.CommonTableExpr,
-        ctx: context.CompilerContextLevel) -> pgast.IteratorCTE:
-    last = ctx.enclosing_cte_iterator
-    old = last.all_ids if last else frozenset()
-    return pgast.IteratorCTE(ir, cte, old | frozenset([ir.path_id]))
+def merge_iterator_scope(iter: Optional[pgast.IteratorCTE],
+                         select: pgast.SelectStmt,
+                         *, ctx: context.CompilerContextLevel) -> None:
+    while iter:
+        ctx.path_scope[iter.path_id] = select
+        iter = iter.parent
 
 
 def fini_dml_stmt(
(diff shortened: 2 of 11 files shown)


edb/ir/utils.py
```diff
@@ -290,3 +290,12 @@ def get_iterator_sets(stmt: irast.Stmt) -> Sequence[irast.Set]:
         iterators.extend(stmt.hoisted_iterators)
 
     return iterators
+
+
+def contains_dml(stmt: irast.Stmt) -> bool:
+    """Check whether a statement contains any DML in a subtree."""
+    # If this ends up being a perf problem, we can use a visitor
+    # directly and cache.
+    res = ast.find_children(stmt, lambda x: isinstance(x, irast.MutatingStmt),
```
1st1, line 299: How often will this thing re-traverse the tree? Do we need to add a cache here?

edb/pgsql/compiler/dml.py
```diff
@@ -571,9 +555,9 @@ def process_insert_body(
 
         subctx.expr_exposed = False
 
-        if iterator_cte is not None:
+        if iterator is not None:
             subctx.path_scope = ctx.path_scope.new_child()
-            subctx.path_scope[iterator_cte.query.path_id] = select
+            subctx.path_scope[iterator.cte.query.path_id] = select
```
elprans:
```suggestion
            subctx.path_scope[iterator.set.path_id] = select
```

edb/pgsql/compiler/dml.py
```diff
@@ -698,12 +682,12 @@ def compile_insert_else_body(
...
-                iterator_set, iterator_cte = iterator
                 iterator_rvar = relctx.rvar_for_rel(
-                    iterator_cte, lateral=True, ctx=ictx)
+                    iterator.cte, lateral=True, ctx=ictx)
+                # XXX: what is up with the path ids here.
                 relctx.include_rvar(ictx.rel, iterator_rvar,
-                                    path_id=iterator_set.path_id, ctx=ictx)
-                ictx.path_scope[iterator_cte.query.path_id] = ictx.rel
+                                    path_id=iterator.set.path_id, ctx=ictx)
+                ictx.path_scope[iterator.cte.query.path_id] = ictx.rel
```
elprans:
```suggestion
                ictx.path_scope[iterator.set.path_id] = ictx.rel
```

edb/ir/utils.py
msullivan: Once per FOR, so it is quadratic in the nesting depth of FOR statements, which didn't seem that likely to be bad to me.

edb/pgsql/compiler/stmt.py
```diff
@@ -49,8 +50,27 @@ def compile_SelectStmt(
 
         query = ctx.stmt
 
-        if not isinstance(stmt.result.expr, irast.MutatingStmt):
-            iterators = irutils.get_iterator_sets(stmt)
+        iterators = irutils.get_iterator_sets(stmt)
+        if iterators and irutils.contains_dml(stmt):
+            # If we have iterators and we contain nested DML
```
elprans, line 55: We need to prove that the iterators are hoist-able, i.e. that they're not correlated with anything except among themselves.

msullivan:
I just realized that there are lingering issues with queries like:
```
FOR noob in {"Phil Emarg", "Madeline Hatch"} UNION (
    INSERT Person { name := noob, notes := (INSERT Note { name := noob } ) });
```
where the inner INSERT will need to directly join on both the FOR and the outer INSERT, since the outer INSERT only can output the iterator identity and not the values too.

This shouldn't be too hard to fix.

edb/pgsql/compiler/dml.py
```diff
@@ -459,80 +529,11 @@ def process_insert_body(
...
-            aspect='identity', ctx=ctx)
-
-    else:
-        iterator_cte = None
-        iterator_id = None
+    iterator_id: Optional[pgast.BaseExpr] = None
+    iterator = ctx.enclosing_cte_iterator
+    if iterator is not None:
+        # XXX: THAT PATH BOND????
+        pathctx.put_path_bond(insert_stmt, iterator.path_id)
```
elprans, line 536: You do this below, is it needed earlier?

edb/pgsql/compiler/dml.py
```diff
@@ -301,6 +309,43 @@ def wrap_dml_cte(
...
 
+def merge_iterator_scope(iterator: Optional[pgast.IteratorCTE],
+                         select: pgast.SelectStmt,
+                         *, ctx: context.CompilerContextLevel) -> None:
+    while iterator:
+        ctx.path_scope[iterator.path_id] = select
+        iterator = iterator.parent
+
+
+def merge_iterator(iterator: Optional[pgast.IteratorCTE],
```
elprans, line 320: Please use formatting for function signatures consistent with other in the file.

edb/pgsql/compiler/dml.py
msullivan: No

Issue #1661, opened 2020-08-13
Prohibit DML queries in function bodies 


vpetrovykh:
The proposal is to prohibit `INSERT`, `UPDATE`, and `DELETE` statements in function bodies.

Currently these statements cannot appear in conditional expressions. More specifically, they are executed unconditionally, so it is semantically misleading to place them in a conditional branch that may or may not be executed - so this is forbidden. There's a specialized form of these statements that allows them to be conditionally executed #1639.

However, functions don't have the same unconditional execution policy and if these statements were to appear in functions, this could lead to unpredictable results, especially since function calls may be optimized by the query planner based on volatility and where they appear. In order to prevent unspecified behavior we currently choose to ban data manipulation inside function bodies.

Fogapod:
Does this include `DELETE` statement?

elprans:
> Does this include `DELETE` statement?

Yes.

To clarify, the decision to prohibit DML in functions plugs an undefined behavior hole until we figure out how to allow this safely.

vpetrovykh:
I've updated the top post to include `DELETE` there as well.

Fogapod:
I hope you understand that this would be a big breaking change. Is there is absolutely no other way around at the moment?
The ability to use DDL in functions was there for months and several alphas (if not from 1st alpha?).
Some people already integrated it (including myself):
- https://spectrum.chat/edgedb/general/personal-accomplishment-edgedb-authentication-from-svelte-js~ff16a60c-91b2-4e85-aa13-1aa3d22d9500

elprans:
> I hope you understand that this would be a big breaking change.

Absolutely!  Hence the "under discussion" label here.

> The ability to use DML in functions was there for months

You might have noticed that functions with DML have serious issues: 1) it's impossible to see the effect of a data-modifying function in a calling query;  2) depending on where the call site in the query is, the function might not even get called, or it might, the behavior is undefined.  We'd like to avoid inheriting PostgreSQL mistakes and it's not too late yet.

> Is there is absolutely no other way around at the moment?

We can add a configuration knob that enables the current behavior, e.g. `CONFIGURE SYSTEM/SESSION SET allow_dml_in_functions := true` (it'll be disabled by default).

Furhter, we will look into finding a way to allow DML-containing functions to be used in a safe and predictable manner, but before we do that it's imperative that we don't promote their use in new code.

1st1:
[...]

PR #1741, opened 2020-09-02
Add tracking of DML inside function bodies.

vpetrovykh:
Track the presence of DML statemnts and function calls that appear
inside function bodies.

Issue #1661

### DIFF
--- a/edb/edgeql/compiler/stmt.py
+++ b/edb/edgeql/compiler/stmt.py
@@ -333,6 +333,9 @@ def compile_InsertQuery(
             context=expr.context,
         )
 
+    # Record this node in the list of potential DML expressions.
+    ctx.env.dml_exprs.append(expr)
+
     with ctx.subquery() as ictx:
         stmt = irast.InsertStmt()
         init_stmt(stmt, expr, ctx=ictx, parent_ctx=ctx)
@@ -416,6 +419,9 @@ def compile_UpdateQuery(
             context=expr.context,
         )
 
+    # Record this node in the list of potential DML expressions.
+    ctx.env.dml_exprs.append(expr)
+
     with ctx.subquery() as ictx:
         stmt = irast.UpdateStmt()
         init_stmt(stmt, expr, ctx=ictx, parent_ctx=ctx)
@@ -483,6 +489,9 @@ def compile_DeleteQuery(
             'DELETE statements cannot be used inside conditional expressions',
             context=expr.context,
         )
+
+    # Record this node in the list of potential DML expressions.
+    ctx.env.dml_exprs.append(expr)
 
     with ctx.subquery() as ictx:
         stmt = irast.DeleteStmt()
--- a/edb/schema/functions.py
+++ b/edb/schema/functions.py
@@ -985,6 +985,9 @@ class Function(CallableObject, VolatilitySubject, s_abc.Function,
     session_only = so.SchemaField(
         bool, default=False, compcoef=0.4, coerce=True, allow_ddl_set=True)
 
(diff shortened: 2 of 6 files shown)


vpetrovykh:
So do we want to have some environment variable or some other way of controlling whether the DML in functions is an error or a warning?

elprans:
> So do we want to have some environment variable or some other way of controlling whether the DML in functions is an error or a warning?

A per-database configuration variable.  We obviously need to implement per-database configuration support first.

vpetrovykh:
OK, so if you have no other objections, we can merge this and then update the handling of DML in functions once we have per-database configs.

edb/schema/functions.py
```diff
@@ -1122,22 +1125,30 @@ def compile_function(
...
-                session_mode=session_only,
-            ),
+            options=options,
         )
 
         ir = compiled.irast
         assert isinstance(ir, irast.Statement)
         schema = ir.schema
 
+        if options.dml_exprs:
```
elprans, line 1147: Please use `irast.Statement` to communicate the characteristics of IR.

edb/edgeql/compiler/stmt.py
```diff
@@ -484,6 +490,9 @@ def compile_DeleteQuery(
             context=expr.context,
         )
 
+    # Record this node in the list of potential DML expressions.
+    ctx.env.options.dml_exprs.append(expr)
```
elprans, line 494: It seems to me that a future-proof version of this should be in the form of an inferer-visitor, like the one we have for volatility.  That way we can infer if a chain of function calls is, in fact, a mutation.

edb/schema/functions.py
vpetrovykh: OK, I'll move that from the context options to `irast.Statement`.

edb/edgeql/compiler/stmt.py
vpetrovykh: The benefit of the current approach is that it allows us to mark a specific piece of code as the culprit in the error or warning. Visitors that infer this for the whole subtree will just broadly state whether the body is or isn't DML-containing, which, in case of this being due to another DML-containing function call is not as useful.

elprans: Nothing prevents the mutation detection visitor to return a list of exprs instead of a boolean answer.

edb/edgeql/compiler/options.py
```diff
@@ -74,6 +75,10 @@ class GlobalCompilerOptions:
     #: to contain DML in the top-level shape computables.
     allow_top_level_shape_dml: bool = False
 
+    #: A list of DML expressions (statements and DML-containing
+    #: functions) that appear in a function body.
+    dml_exprs: List[qlast.Base] = dc_field(default_factory=list)
```
elprans, line 80: Please move this to `Environment`.  `CompilerOptions` are options passed _to_ the compiler, they shouldn't be used to pass information back to the caller or mutated.

vpetrovykh: Done

edb/edgeql/compiler/stmt.py
vpetrovykh: In this case I don't really see the benefit of having a separate visitor when this can be done during compile pass. Additionally, if/when we allow DML-containing functions to exist, then we'll need to correctly resolve the (potentially aliased in `WITH` block) name, etc. to correctly infer whether a given expression is DML-containing. This would be just duplication of the same code form compiling function calls.

elprans (approved):

PR #1746, opened 2020-09-04
Fix issue with coalescing inside UPDATE

msullivan:
The check for optionality in get_set_rvar was not being done in the
scope of the path, so we could miss optionality in some cases, such as
two xfailing tree tests.

### DIFF
--- a/edb/pgsql/compiler/relgen.py
+++ b/edb/pgsql/compiler/relgen.py
@@ -161,9 +161,9 @@ def get_set_rvar(
         is_empty_set = isinstance(ir_set, irast.EmptySet)
 
         path_scope = relctx.get_scope(ir_set, ctx=subctx)
-        scope = path_scope or subctx.scope_tree
+        new_scope = path_scope or subctx.scope_tree
         is_optional = (
-            scope.is_optional(path_id) or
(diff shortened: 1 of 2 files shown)


edb/pgsql/compiler/relgen.py
```diff
@@ -172,7 +174,6 @@ def get_set_rvar(
                 ir_set=ir_set, stmt=stmt, ctx=subctx)
             subctx.pending_query = subctx.rel = stmt
 
-        path_scope = relctx.get_scope(ir_set, ctx=subctx)
         if path_scope:
```
elprans, line 177: A slightly better fix would be to move this `if` block to before `is_optional`.  `update_scope` reassigns `ctx.scope_tree` to a correct value.

elprans: Argh, disregard. This breaks the optional wrapper.

elprans (approved):

edb/pgsql/compiler/relgen.py
msullivan: Yeah, that was the first thing I tried, but moving it ahead of the prepare optional excitingly does not work.

PR #1747, opened 2020-09-04
Fix the test case retry loop

msullivan:
It was failing due to not starting a new transaction.  I tested by
putting some diagnostics in and running it until I observed that the
retry happened sucessfully.

### DIFF
--- a/edb/testbase/server.py
+++ b/edb/testbase/server.py
@@ -114,9 +114,11 @@ class TestCaseMeta(type(unittest.TestCase)):
                     if try_no == 3:
                         raise
                     else:
-                        self.loop.run_until_complete(self.con.execute(
-                            'ROLLBACK;'
-                        ))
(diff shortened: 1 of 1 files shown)


edb/testbase/server.py
```diff
@@ -114,9 +114,11 @@ def wrapper(self, *args, __meth__=meth, **kwargs):
                     if try_no == 3:
                         raise
                     else:
-                        self.loop.run_until_complete(self.con.execute(
-                            'ROLLBACK;'
-                        ))
+                        if getattr(self, 'ISOLATED_METHODS'):
```
elprans, line 117: Perhaps condition the entire loop in this `getattr`?  It probably makes no sense to retry tests that we haven't implicitly established a transaction for.

msullivan:
I think this is ready to merge?

elprans (approved):

PR #1750, opened 2020-09-04
Allow the usage of EXISTS in constraints.

vpetrovykh:
EXISTS operator is now compiled inside constraint expressions.

Issue #1727


vpetrovykh:
It looks like this PR uncovered a bug with `required` - see `test_edgeql_ddl_required_10` and `test_edgeql_ddl_required_11`.

elprans:
> It looks like this PR uncovered a bug with `required`

Please open an issue.

elprans (approved):

PR #1753, opened 2020-09-05
Rename string handling funtions to use "start" and "end".

### LINKED ISSUE #1752: Rename some string handling functions and deprecate the old versions.
From discussion in Slack a point was raised:
> Just noticed something now of an engineering nature though:
> https://edgedb.com/docs/edgeql/funcops/string#function::std::str_lpad
> This reminded me of similar methods in Rust that were eventually deprecated in favour of calling them "start" and "end" instead of "right" and "left" due to languages that write right to left (and even up to down):
> https://doc.rust-lang.org/std/primitive.str.html#method.trim_left
> The new methods did the exact same thing but were just given different names, while the old method names were left alone to accommodate legacy code.

This means that we want to rename string handling functions to use "start" and "end" such as `std::str_trim_start`, etc. The old functions should remain for some time, but they need to be marked as deprecated. To this end we also introduce a `deprecated` annotation which can mark deprecated items as well as provide a useful note like what should be used instead.

The documentation needs to include deprecated functions, but they should also be clearly labeled as such.

vpetrovykh:
Rename string handling functions to use "start" and "end" such as
`std::str_trim_start`, etc. The old functions are marked as
deprecated. To this end `deprecated` annotation is introduced.

Fixes #1752


docs/edgeql/funcops/string.rst
```diff
@@ -241,10 +241,38 @@ String
...
+        db> SELECT str_pad_start('much too long', 10);
+        {'much too l'}
+        db> SELECT str_pad_start('short', 10, '.:');
+        {'.:.:.short'}
+
+
+----------
+
+
 .. eql:function:: std::str_lpad(string: str, n: int64, fill: str = ' ') -> str
```
1st1, line 267: I'd add a separate page collecting all deprecated functions.

1st1 (approved):
LGTM, but let's have a separate dedicated page for all deprecated functions.

docs/edgeql/funcops/string.rst
vpetrovykh: OK, I've made a separate page for that.

PR #1755, opened 2020-09-08
Improve database initialization

tailhook:
1. Added `--bootstrap-script` and `--bootstrap-command` arguments to run some
   commands just after bootstrap
2. Renamed `--bootstrap` to `--bootstrap-only`
3. Automatic user and database (either --default-database or unix user name)
   are not created at the server restart any more (only at bootstrap)

A little bit related to edgedb/edgedb-docker#3 but not exactly that

### DIFF
--- a/edb/server/main.py
+++ b/edb/server/main.py
@@ -483,7 +483,7 @@ _server_options = [
         '--bootstrap-command', metavar="QUERIES",
         help='run the commands when initializing the database. '
              'Queries are executed by default user within default '
-             'database. May be used with or without `--bootstrap`.'),
+             'database. May be used with or without `--bootstrap-only`.'),
     click.option(
         '--bootstrap-script', type=PathPath(), metavar="PATH",
         help='run the script when initializing the database. '
--- a/edb/server/main.py
+++ b/edb/server/main.py
@@ -35,6 +35,7 @@ import socket
 import sys
 import tempfile
 import typing
+import warnings
 
 import uvloop
 
@@ -137,9 +138,13 @@ def _init_cluster(cluster, args: ServerConfig) -> bool:
         'default_database_user': args.default_database_user,
         'testmode': args.testmode,
         'insecure': args.insecure,
+        'bootstrap_script': args.bootstrap_script,
+        'bootstrap_command': args.bootstrap_command,
     }
 
-    need_restart = asyncio.run(bootstrap.bootstrap(cluster, bootstrap_args))
+    need_restart = asyncio.run(
+        bootstrap.ensure_bootstrapped(cluster, bootstrap_args)
+    )
 
     global _server_initialized
     _server_initialized = True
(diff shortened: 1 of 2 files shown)


edb/server/main.py
```diff
@@ -469,6 +479,16 @@ def _protocol_version(
     click.option(
         '--default-database-user', type=str, default=getpass.getuser(),
         help='the name of the default database owner'),
+    click.option(
+        '--init-command', metavar="QUERIES",
```
elprans, line 483: I still like `--bootstrap-command` better, because it signifies that this would _only_ be executed during bootstrap.  To clarify things, we can rename `--bootstrap` to `--bootstrap-only`.

edb/server/main.py
```diff
@@ -286,9 +294,9 @@ def run_server(args: ServerConfig):
                 abort('Could not start database cluster in %s',
                       args.data_dir)
 
-            need_cluster_restart = _init_cluster(cluster, args)
-
-            if need_cluster_restart and pg_cluster_started_by_us:
+            need_cluster_restart = _init_cluster(cluster, args,
+                                                 pg_cluster_init_by_us)
```
elprans, line 298: I'm not sure this is correct.  `pg_cluster_init_by_us` indicates whether or not we ran `initdb`.  It would always be `False` for RDS instances and so `--init-command` wouldn't get executed.  Instead, let's move everything, including the creation of the default database under the `if new_template_db_id` and stop doing "partial bootstrap".

tailhook: Hm, yes, now I see that `_ensure_edgedb_template_database(..)` is actually our flag. But I'm still not sure why we need those alter database: https://github.com/edgedb/edgedb/blob/ea3a6ec9196a5c0f6bcd34f80a31bf562eff369a/edb/server/bootstrap.py#L207-L230
It looks like we want to skip any database mutations if DB is bootstrapped (i.e. when template database exists). Including the cited snippet and creating superuser at: https://github.com/edgedb/edgedb/blob/ea3a6ec9196a5c0f6bcd34f80a31bf562eff369a/edb/server/bootstrap.py#L1004-L1011 
Or am I misunderstanding something?

elprans: Yes, skip all mutations post-bootstrap (i.e. move them into the bootstrap process).

edb/server/main.py
tailhook: Sounds good.

edb/server/main.py
tailhook: Fixed.

edb/server/main.py
```diff
@@ -461,14 +468,27 @@ def _protocol_version(
...
         '--default-database', type=str, default=getpass.getuser(),
         help='the name of the default database to create'),
     click.option(
         '--default-database-user', type=str, default=getpass.getuser(),
         help='the name of the default database owner'),
+    click.option(
+        '--bootstrap-command', metavar="QUERIES",
+        help='run the commands when initializing the database. '
+             'Queries are executed by default user within default '
+             'database. May be used with or without `--bootstrap`.'),
```
elprans, line 486:
```suggestion
             'database. May be used with or without `--bootstrap-only`.'),
```

edb/server/main.py
```diff
@@ -461,14 +468,27 @@ def _protocol_version(
...
+    click.option(
+        '--bootstrap-script', type=PathPath(), metavar="PATH",
+        help='run the script when initializing the database. '
+             'Script run by default user within default database. '
+             'May be used with or without `--bootstrap`.'),
```
elprans, line 491:
```suggestion
             'May be used with or without `--bootstrap-only`.'),
```

elprans (approved):
LGTM (+minor nit)

PR #1758, opened 2020-09-10
tests: Add more tests centering on migrations.

vpetrovykh:
Add a test showing an issue with tracing function bodies.
Add tests for renaming abstract annotations.
Add tests for `DESCRIBE CURRENT MIGRATION` workflow.

A number of these tests are not currently passing.


vpetrovykh:
This is just to highlight different problems with migrations. Separate issues and PRs would probably be necessary to address the `xfail` scenarios.

elprans (approved):

elprans (approved):

PR #1762, opened 2020-09-11
Fix broken SET OF/OPTIONAL interactions

### LINKED ISSUE #1640: Bug in interaction of OPTIONAL and SET OF
- EdgeDB Version: 1.0-alpha.4+dev.92.g20b0bd7f
- OS Version: Linux

Steps to Reproduce:

```
victor> CREATE TYPE Foo;
OK: CREATE
victor> SELECT (<str>Foo.id) ?? (<str>count(Foo));
{}
```
The expected result is `{'0'}`, since we expect the `count` of an empty set be called.

msullivan:
The interaction between SET OF and OPTIONAL needs more careful
handling than is done currently.
Basically, the idea is:
1. Merging an optional path with a non-optional one should result in a
   non-optional path, *unless* the non-optional path was fenced.
2. When an path is factored out past an optional parent, it should
   become optional. This rule allows us to avoid marking *all* paths
   appearing in an optional argument as optional, which would result
   in worse codegen.

Applying these rules is complicated by the fact that when we compile
arguments, we don't know whether it is a SET OF or an OPTIONAL. We
solve this by tracking how many non-optional paths have been merged
into a previously optional node and doing a post-hoc fixup of all the
paths encountered in SET OF and OPTIONAL arguments.

Fixes #1640.

### DIFF
--- a/edb/edgeql/compiler/func.py
+++ b/edb/edgeql/compiler/func.py
@@ -639,15 +639,16 @@ def process_path_log(arg_ctx: Optional[context.ContextLevel],
         # while compiling an argument and find and adjust the
         # optionality of any factored out nodes after the fact.
         for path_id in arg_ctx.path_log:
-            assert arg_scope is not None
-            # If the node is still here, nothing to do
-            desc = arg_scope.find_descendant(path_id)
-            if desc:
-                continue
-            visible = arg_scope.find_visible(path_id)
-            if visible:
-                if visible.optional_count:
-                    visible.optional_count -= 1
+            for prefix in path_id.iter_prefixes(include_ptr=True):
+                assert arg_scope is not None
+                # If the node is still here, nothing to do
+                desc = arg_scope.find_descendant(prefix)
+                if desc:
+                    continue
+                visible = arg_scope.find_visible(prefix)
+                if visible:
+                    if visible.optional_count:
+                        visible.optional_count -= 1
         arg_ctx.path_log = []
 
 
--- a/edb/edgeql/compiler/func.py
+++ b/edb/edgeql/compiler/func.py
@@ -94,7 +94,7 @@ def compile_FunctionCall(
         ctx.env.options.schema_object_context is s_constr.Constraint
     )
 
-    args, kwargs = compile_call_args(expr, funcname, ctx=ctx)
+    args, kwargs, arg_ctxs = compile_call_args(expr, funcname, ctx=ctx)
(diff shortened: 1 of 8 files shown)


edb/ir/scopetree.py
```diff
@@ -790,6 +828,14 @@ def find_descendant_and_ns(
 
         return None, frozenset(), None
 
+    def is_optional_upto(self, ancestor: Optional[ScopeTreeNode]) -> bool:
+        nobe: Optional[ScopeTreeNode] = self
```
elprans, line 832:
```suggestion
        node: Optional[ScopeTreeNode] = self
```

edb/edgeql/compiler/func.py
```diff
@@ -713,4 +756,9 @@ def finalize_args(
 
         args.append(call_arg)
 
+        # If we have any logged paths left over and our enclosing
+        # context is logging paths, propagate them up.
+        if arg_ctx and arg_ctx.path_log and ctx.path_log is not None:
```
elprans, line 761: Should it be an error to have leftover paths in the log and enclosing context NOT logging paths?

edb/edgeql/compiler/func.py
```diff
@@ -610,13 +621,40 @@ def compile_call_args(
...
+            for prefix in path_id.iter_prefixes(include_ptr=True):
+                assert arg_scope is not None
+                # If the node is still here, nothing to do
+                desc = arg_scope.find_descendant(prefix)
+                if desc:
+                    continue
+                visible = arg_scope.find_visible(prefix)
+                if visible:
+                    if visible.optional_count:
+                        visible.optional_count -= 1
```
elprans, line 651: Is it possible for `visible.optional_count` to be > 1 here?

elprans (approved):
Looks good, I like the approach!  I've left a couple of questions in the review comments.

edb/edgeql/compiler/func.py
msullivan: No, that happens all the time. For example in `SELECT Foo.num + 1`, we'll log `Foo.num`, we won't process it, but the enclosing context doesn't care.

edb/edgeql/compiler/func.py
msullivan: Yeah. If we have something like `SELECT (Foo.num ?= 0, Foo.num ?= Foo.num)`, the optional_count will go up to 2 and then come back down to 0. If we have ``SELECT (Foo.num ?= 0, Foo.num = Foo.num)`, it will go up to 2 and stay there.

PR #1763, opened 2020-09-12
Fix some guided migration scenarios.

vpetrovykh:
Allow `REJECT PROPOSED` when the "proposed" is empty.

Add `CREATE`/`DROP` commands for modules to the `DESCRIBE CURRENT
MIGRATION AS JSON`.

### DIFF
--- a/edb/server/compiler/compiler.py
+++ b/edb/server/compiler/compiler.py
@@ -959,42 +959,47 @@ class Compiler(BaseCompiler):
                 generate_prompts=True,
                 guidance=mstate.guidance,
             )
-            top_command = next(iter(diff.get_subcommands()))
 
-            if (orig_cmdclass := top_command.get_annotation('orig_cmdclass')):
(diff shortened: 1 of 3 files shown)


edb/server/compiler/compiler.py
```diff
@@ -959,42 +959,50 @@ def _compile_ql_migration(self, ctx: CompileContext, ql: qlast.Migration):
...
                 guidance=mstate.guidance,
             )
-            top_command = next(iter(diff.get_subcommands()))
 
-            if (orig_cmdclass := top_command.get_annotation('orig_cmdclass')):
-                top_cmdclass = orig_cmdclass
-            else:
-                top_cmdclass = type(top_command)
+            from edb.common.markup import dump
+            dump(diff, marker='compiler.py:967')
```
elprans, line 964: Forgotten dump

elprans (approved):

PR #1773, opened 2020-09-15
Disallow DELETE and UPDATE in many cases without FILTER

### LINKED ISSUE #1742: Require a FILTER on DELETE and UPDATE?
There was some discussion earlier about requiring FILTER on DELETE and UPDATE statements, to help prevent mishaps. `FILTER true` could be used when operating on everything is actually desired.

(Relevant: https://www.youtube.com/watch?v=i_cVJgIz_Cs)

msullivan:
This patch errs on the side of disallowing, since it is pretty easy to
workaround.

We allow DML when the source is filtered, comes from a query that is
not a set reference, when it is a FOR iterator, and when it is the set
being inserted in the ELSE clause of INSERT ... ELSE.
(FOR is allowed because the meaning seems to clearly imply applying to
everything ...and because it made the implementation easy.)

Most of the test cases for things we allow are existing ones.

I can follow up with a documentation update if there is agreement on
this one.

Closes #1742.

### DIFF
--- a/edb/edgeql/compiler/stmt.py
+++ b/edb/edgeql/compiler/stmt.py
@@ -420,47 +420,6 @@ def compile_InsertQuery(
     return result
 
 
-def allow_update_or_delete(
-        subject: irast.Set,
-        *,
-        ctx: context.ContextLevel) -> bool:
-    # Allow deleting (...).foo if we would allow deleting (...)
-    if subject.rptr and allow_update_or_delete(subject.rptr.source, ctx=ctx):
-        return True
-
-    # This is a bit of a hack, but allow deleting paths that are in
-    # the factoring_allowlist. This lets us delete FOR iterators and
-    # the INSERT set in INSERT ... ELSE.
-    if (ctx.path_scope.parent
-            and subject.path_id in ctx.path_scope.parent.factoring_allowlist):
-        return True
-
-    sexpr = subject.expr
-
-    if not sexpr:
-        return False
-
-    # Allow non SELECT and filtered things
-    if not isinstance(sexpr, irast.SelectStmt) or sexpr.where:
-        return True
-
-    return allow_update_or_delete(sexpr.result, ctx=ctx)
-
-
-def enforce_constrained_dml(
-        expr: qlast.Expr,
-        subject: irast.Set,
-        where: Optional[qlast.Expr],
-        *,
-        ctx: context.ContextLevel) -> None:
-    if not where and not allow_update_or_delete(subject, ctx=ctx):
-        typ = "DELETE" if isinstance(expr, qlast.DeleteQuery) else "UPDATE"
-        raise errors.QueryError(
-            f'{typ} statement requires a FILTER clause',
-            context=expr.context
-        )
-
-
 @dispatch.compile.register(qlast.UpdateQuery)
 def compile_UpdateQuery(
         expr: qlast.UpdateQuery, *, ctx: context.ContextLevel) -> irast.Set:
@@ -469,6 +428,12 @@ def compile_UpdateQuery(
         raise errors.QueryError(
             'UPDATE statements cannot be used inside conditional expressions',
             context=expr.context,
+        )
+
+    if not expr.where:
+        raise errors.QueryError(
+            f'UPDATE statement requires a FILTER clause',
+            context=expr.context
         )
 
     # Record this node in the list of potential DML expressions.
@@ -480,8 +445,6 @@ def compile_UpdateQuery(
 
         subject = dispatch.compile(expr.subject, ctx=ictx)
         assert isinstance(subject, irast.Set)
-
-        enforce_constrained_dml(expr, subject, where=expr.where, ctx=ictx)
 
         subj_type = inference.infer_type(subject, ictx.env)
         if not isinstance(subj_type, s_objtypes.ObjectType):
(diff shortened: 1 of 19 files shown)


1st1:
@elprans Please review

elprans:
I'm finding the implementation of the rule somewhat fuzzy.  For example this is allowed:

```
WITH foo := (SELECT User UNION User2) DELETE foo;
```

But this isn't:

```
WITH foo := (SELECT User) DELETE foo;
```

IMO, we should only error on `DELETE Foo` if `Foo` is a direct reference to a universal set of `Foo`, since the intent of this change is to catch cases of unintentional `DELETE`/`UPDATE` in a quick REPL scenario rather than in an elaborate query scenario.

msullivan:
That particular case doesn't feel that fuzzy; `foo` is basically just a direct alias to `User`. Disallowing DML of direct aliases to types seems potentially helpful and not potentially hurtful, so my inclination is to keep it, though I don't feel strongly.

elprans:
> Disallowing DML of direct aliases to types seems potentially helpful and not potentially hurtful

Well, semantically `foo := (SELECT User UNION User2)` is also a direct alias, albeit to a union type.  Why is it different?

msullivan:
> Well, semantically `foo := (SELECT User UNION User2)` is also a direct alias, albeit to a union type. Why is it different?

I'd say for exactly that reason, which is that its a union type and not a base type. But I also wouldn't really expect any users to think about it that way instead of as the documented "merge two sets" operation. Moreover, union types I don't think are a concept we talk about in the documentation at all

1st1:
I'm going to go ahead and suggest the most straightforward and stupid-simple option: always disallow DELETEs and UPDATEs without FILTER.

Cons: will be annoying to the few users who *really* know what they're doing and writing relatively complex queries with WITH-aliases and subqueries.

Pros: easy to implement; we can start relaxing the rule after 1.0.

vpetrovykh:
I am inclined to agree with @1st1. In an ideal world I prefer a more purist approach and would want the rule to be semantically consistent and rigorous. However, since this is meant to curb an accidental catastrophic [REPL] error, I'm fine supporting a rule that is brain-dead simple and worry about how to refine it later.

As far as invalidating previous (dumped) expressions, well, we *can* inject the `FILTER True` clause to all `UPDATE` and `DELETE` expressions that lack a `FILTER` when restoring the dump. Basically, the user has already decided that the expression was OK, so we should just roll with it.

msullivan:
That seems fine with me. It's a little unfortunate for `UNLESS CONFLICT` but otherwise mostly fine. I went ahead and implemented it, but can make other changes instead if we want.

1st1:
> It's a little unfortunate for UNLESS CONFLICT but otherwise mostly fine. 

Can you please clarify?

msullivan:
> > It's a little unfortunate for UNLESS CONFLICT but otherwise mostly fine.
> 
> Can you please clarify?

The set already gets bound to the existing object, but then we still need to add a FILTER, so we end up with
```
INSERT Person {name := "Emmanuel Villip"} UNLESS CONFLICT
ON .name ELSE (UPDATE Person FILTER true SET { tag := "redo" })
```

1st1:
> The set already gets bound to the existing object, but then we still need to add a FILTER, so we end up with

That's indeed unfortunate. Can we make one exception from the rule, and allow filter-less UPDATE when they're directly in the rhs of `UNLESS CONFLICT`?

elprans:
I have pondered on this issue for a while, and I don't think the syntactic solution is the best approach.

Even if we implement the "is-universal-set" inference, there would still remain a big issue with requiring an explicit `FILTER false`: it breaks composability.  It is no longer safe to write a query composer that does `DELETE $foo`, where `$foo` is determined dynamically, or even `DELETE foo()`, where `foo()` is a function that returns some set.  And so, query builders and other pieces of code that construct queries dynamically would be forced to always inject `FILTER false`, which defeats its purpose to a large extent.

Considering that the original motivation here is to prevent accidental unbound deletion or update, I propose a different solution: report the "unfiltered DML" as a special lint condition (or a warning) in the query descriptor that is returned by the `Prepare` message.  The REPL can then use the warning to prompt the user to confirm their action, and language bindings can emit the warning into a log, or, optionally, turn it into an error.  The `FILTER false` would still be a way to avoid the issuance of the warning.  The advantage of such a lint/warning system is that it is generic and can be extended to cover other dubious query situations (e.g. use of volatiles, bad correlation etc).

We've had some discussion about this with the team on Monday and there seems to be agreement that the lint/warning solution is better.  It would also avoid the need for the introduction of backwards-incompatible syntax changes, and so we don't have to rush this in right now.

msullivan:
OK, that seems reasonable. What's the path forward for building that--how much engineering/design effort is it likely to take?

I'm not sure if it exactly makes sense to use for my volatile stuff--or at least, we'd have to be willing to say that if you have volatile warnings then we don't  really give any semantics for what happens with them.

1st1:
> OK, that seems reasonable. What's the path forward for building that--how much engineering/design effort is it likely to take?

We'll need to update the https://github.com/edgedb/edgedb/discussions/1849 document to include a spec for warnings. And settle on how we are going to encode warnings/errors in binary.

We'll then need to decide how exactly we attach warnings to the Prepare response message: every warning would use its own header, or we combine all warnings in one.

As for the implementation, we can approach it in several stages. We don't need the `throw` syntax day one, we can start with implementing just the exception hierarchy and warnings.

Overall I'd say we can do the entire thing in about a week.

elprans:
> I'm not sure if it exactly makes sense to use for my volatile stuff

I don't think so either, it was just an example of a possible case where the semantics is consistent with the model, but the apparent intent is wrong.

1st1:
Closing this as we decided to not pursue this idea. Instead we'll implement a warnings system to confirm if the user actually wants to execute a FILTER-less DELETE or UPDATE. Here's an issue for that https://github.com/edgedb/edgedb/issues/3137

PR #1775, opened 2020-09-15
Fix INSERT-OR-SELECT when combined with shape

### LINKED ISSUE #1743: Insert-or-select fails when it's combined with a shape.
- EdgeDB Version: EdgeDB 1.0-alpha.6+dev.5050.gb8aa1c99
- OS Version: Linux

Steps to Reproduce:
```
CREATE TYPE default::Foo {
    CREATE OPTIONAL SINGLE PROPERTY name -> std::str {
        CREATE CONSTRAINT std::exclusive;
    };
};
CREATE TYPE default::Bar {
    CREATE REQUIRED SINGLE LINK foo -> default::Foo;
};
INSERT Foo {name := 'foo0'};
```
Then the following happens:
```
victor> SELECT (INSERT Bar {foo := (INSERT Foo {name := 'foo0'} UNLESS CONFLICT ON .name ELSE (SELECT Foo))}) {id};
{Object {id: 948f4c8e-ed81-11ea-955a-5b32015497c5}}
victor> SELECT (INSERT Bar {foo := (INSERT Foo {name := 'foo0'} UNLESS CONFLICT ON .name ELSE (SELECT Foo))}) {id, foo};
ERROR: InternalServerError: more than one row returned by a subquery used as an expression
  Server traceback:
      edb.errors.InternalServerError: more than one row returned by a subquery used as an expression
```
So it's not that the cardinality check fails, it's that the SQL causes an effect similar to a cardinality problem if foo is included in the final shape to be returned.

msullivan:
Fixes #1743.

### DIFF
--- a/edb/pgsql/compiler/dml.py
+++ b/edb/pgsql/compiler/dml.py
@@ -377,7 +377,7 @@ def fini_dml_stmt(
         assert len(parts.dml_ctes) == 1
         cte = next(iter(parts.dml_ctes.values()))[0]
         relctx.add_type_rel_overlay(
-            ir_stmt.subject.typeref, 'unIon', cte,
+            ir_stmt.subject.typeref, 'union', cte,
             dml_stmts=dml_stack, path_id=ir_stmt.subject.path_id, ctx=ctx)
     elif isinstance(ir_stmt, irast.DeleteStmt):
         relctx.add_type_rel_overlay(
--- a/edb/pgsql/compiler/dml.py
+++ b/edb/pgsql/compiler/dml.py
@@ -372,8 +372,12 @@ def fini_dml_stmt(
     # referencing this class yields the expected results.
     dml_stack = get_dml_stmt_stack(ir_stmt, ctx=ctx)
     if isinstance(ir_stmt, irast.InsertStmt):
+        # The union CTE might have a SELECT from an ELSE clause, which
+        # we don't actually want to include.
+        assert len(parts.dml_ctes) == 1
+        cte = next(iter(parts.dml_ctes.values()))[0]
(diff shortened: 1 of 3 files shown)


edb/pgsql/compiler/dml.py
```diff
@@ -372,8 +372,12 @@ def fini_dml_stmt(
     # referencing this class yields the expected results.
     dml_stack = get_dml_stmt_stack(ir_stmt, ctx=ctx)
     if isinstance(ir_stmt, irast.InsertStmt):
+        # The union CTE might have a SELECT from an ELSE clause, which
+        # we don't actually want to include.
+        assert len(parts.dml_ctes) == 1
+        cte = next(iter(parts.dml_ctes.values()))[0]
         relctx.add_type_rel_overlay(
-            ir_stmt.subject.typeref, 'union', union_cte,
+            ir_stmt.subject.typeref, 'unIon', cte,
```
elprans, line 380:
```suggestion
            ir_stmt.subject.typeref, 'union', cte,
```

elprans: Not sure about the assertion here.  How would this work on updates to parent types?

elprans: You most likely need to reassemble the `UNION` here without the `ELSE` part.

msullivan: Oops on the `unIon`! I added that to make sure this is where a particular union in the generated code was coming from :P

msullivan: This branch is only for INSERTs. The assert should always be true.

elprans (approved):

PR #1776, opened 2020-09-15
Make sure FOR iterator statement correlates correctly

msullivan:
This fixes an issue introduced in #1757, where declaring a view for
the iterator was done in a temporary scope so we lost the tree
modifications.

### DIFF
--- a/edb/edgeql/compiler/stmt.py
+++ b/edb/edgeql/compiler/stmt.py
@@ -113,52 +113,51 @@ def compile_ForQuery(
         stmt = irast.SelectStmt()
         init_stmt(stmt, qlstmt, ctx=sctx, parent_ctx=ctx)
 
-        with sctx.newscope(fenced=True) as scopectx:
-            iterator_ctx = None
-            if (ctx.expr_exposed and ctx.iterator_ctx is not None
-                    and ctx.iterator_ctx is not sctx):
-                iterator_ctx = ctx.iterator_ctx
+        # Compile the iterator
+        iterator_ctx = None
+        if (ctx.expr_exposed and ctx.iterator_ctx is not None
+                and ctx.iterator_ctx is not sctx):
+            iterator_ctx = ctx.iterator_ctx
 
-            if iterator_ctx is not None:
-                iterator_scope_parent = iterator_ctx.path_scope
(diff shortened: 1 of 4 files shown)


msullivan:
OK this isn't right yet, since `WITH X := {1, 2} SELECT ((FOR x in {X} UNION (SELECT x)), X);` still computes a cross product, although that bug predates the original one.

edb/edgeql/compiler/stmt.py
```diff
@@ -132,6 +132,7 @@ def compile_ForQuery(
 
             iterator_view = stmtctx.declare_view(
                 iterator, qlstmt.iterator_alias,
+                temporary=False,
```
elprans: `scopectx` should probably be `iterator_ctx` here (assuming `else: iterator_ctx = scopectx` is added above)

msullivan: OK, I'll try that. Basically every single thing about this PR has changed since that version, though :P

edb/edgeql/compiler/stmt.py
```diff
@@ -132,7 +132,9 @@ def compile_ForQuery(
 
             iterator_view = stmtctx.declare_view(
                 iterator, qlstmt.iterator_alias,
-                path_id_namespace=path_id_ns, ctx=scopectx)
+                new_namespace=False,
+                path_id_namespace=path_id_ns,
+                ctx=iterator_ctx or scopectx)
```
elprans, line 137: In fact, the actual bug seems to be that `scopectx` is superfluous and only gets in the way as `declare_view` puts `path_namespace` on it only for it to get thrown away.  `declare_view` iself provides enough isolation, there's no need to surround it with fences.  Try simply unnesting the whole `with ... as scopectx` block and replace `scopectx` with `sctx`.  This also obviates the need for `bodyctx` nesting below.  The `new_namespace` hack isn't needed too.

msullivan: Great catch.

I did that, and threw some slight refactoring/rearrangement in because dropping such a big chunk of code into the same nesting level without any visual breaks was giving me the willies :P

elprans (approved):

PR #1779, opened 2020-09-16
Disallow correlation with iterator set if body contains DML

### LINKED ISSUE #1739: Properly reject correlation with FOR iterator containing DML
We should reject queries like `SELECT (Person, (FOR x in {Person} UNION ((INSERT Note {name := x.name}))));`

Currently we generate invalid SQL.

msullivan:
Currently we generate invalid SQL, since we would need to lift the
correlated set into CTEs, which we don't want to deal with.

Fixes #1739.

(Stacked on top of https://github.com/edgedb/edgedb/pull/1776, can't merge yet)

### DIFF
--- a/tests/test_edgeql_insert.py
+++ b/tests/test_edgeql_insert.py
@@ -1188,6 +1188,42 @@ class TestInsert(tb.QueryTestCase):
             ],
         )
 
+    async def test_edgeql_insert_for_bad_01(self):
+        with self.assertRaisesRegex(
+            edgedb.errors.QueryError,
+            "cannot reference correlated set",
+        ):
+            await self.con.execute("""
+                WITH MODULE test
+                SELECT (Person,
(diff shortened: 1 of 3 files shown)


tests/test_edgeql_insert.py
```diff
@@ -1188,6 +1188,42 @@ async def test_edgeql_insert_for_17(self):
...
+            "cannot reference correlated set",
+        ):
+            await self.con.execute("""
+                WITH MODULE test
+                SELECT (Person,
+                        (FOR x in {Person} UNION (
+                             SELECT (INSERT Note {name := x.name}))));
+            """)
+
+    async def test_edgeql_insert_for_bad_03(self):
```
elprans, line 1215: Let's add a test with a deeper `INSERT` placement to ensure the recursive nature of DML search (thinking of possible future refactors).

elprans (approved):

PR #1785, opened 2020-09-17
Don't trigger recursion errors on medium-size set literals

### LINKED ISSUE #475: RecursionError when compiling unions over many queries
Query to repro:

```
WITH MODULE test
FOR X IN {
     (1,), (1,), (1,), (1,), (1,), (1,), (1,), (1,), (1,), (1,), 
     (1,), (1,), (1,), (1,), (1,), (1,), (1,), (1,), (1,), (1,), 
     (1,), (1,), (1,), (1,), (1,), (1,), (1,), (1,), (1,), (1,), 
     (1,), (1,), (1,), (1,), (1,), (1,), (1,), (1,), (1,), (1,), 
}
UNION 1;
```

msullivan:
Currently set literals get compiled into a chain of UNIONs, which
causes a big blowup in the nesting depth of the AST. As a result, we
trip a recursion error on set literals containing 60 one-element tuples,
which is a solidly medium-size input.

Fix this by compiling set literals into a tree of UNIONs instead of a
chain.

Fixes #475 to my satisfaction though others may disagree


elprans (approved):
LGTM as a quick interim fix.  Ideally, we need an n-ary op here.

PR #1791, opened 2020-09-18
Restore find_path_output as an optimization

msullivan:
This will fix an issue where fetch something twice for expr and
serialized, or similar, including sometimes doing two function calls!

A fix I made in #1762 made find_path_output no longer necessary for
correctness, so I removed it, but it will still important as an
optimization.

Restore the optimization bit of that code, and add a comment so I
don't forget and delete it again ;P.

### DIFF
--- a/edb/pgsql/compiler/pathctx.py
+++ b/edb/pgsql/compiler/pathctx.py
@@ -772,6 +772,19 @@ def _get_rel_path_output(
     return result
 
 
+def find_path_output(
+        rel: pgast.BaseRelation, path_id: irast.PathId, ref: pgast.BaseExpr, *,
+        env: context.Environment) -> Optional[pgast.OutputVar]:
+    if isinstance(ref, pgast.TupleVarBase):
+        return None
+
+    for key, other_ref in rel.path_namespace.items():
+        if _same_expr(other_ref, ref) and key in rel.path_outputs:
+            return rel.path_outputs.get(key)
+    else:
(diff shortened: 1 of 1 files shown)


msullivan:
Oh, oops. I ran the test suite locally and somehow then forgot it had failed between yesterday and this morning and it is broken. I'm going to close this for now because I think the problem is serious.

msullivan:
Updated with a fix that I'm not really happy with.

elprans:
> Updated with a fix that I'm not really happy with.

Yeah, this was always fairly flimsy.  Perhaps we should investigate if pushing expression computation into a `LATERAL` relation instead of putting it directly into the target list hurts performance.  If not, that'll be a more reliable fix.  We'd still end up with fatter tuples, though.

edb/pgsql/compiler/pathctx.py
```diff
@@ -171,9 +171,13 @@ def get_path_var(
     var: Optional[pgast.BaseExpr]
 
     if astutils.is_set_op_query(rel):
+        # We disable the find_path_output optimizaiton when doing
+        # UNIONs to avoid situations where they have different numbers
+        # of columns.
         cb = functools.partial(
             get_path_output_or_null,
             env=env,
+            no_find_output=True,
```
elprans, line 180: Let's name this `disable_output_fusion` or something like that.

elprans (approved):

PR #1806, opened 2020-09-22
Fix another FOR correlation bug I introduced earlier

msullivan:


edb/edgeql/compiler/stmt.py
```diff
@@ -128,8 +128,18 @@ def compile_ForQuery(
...
+        # correlation between the iterator and the enclosing
+        # query, since the correlation imposes compilation issues
+        # we aren't willing to tackle.
+        if contains_dml:
+            ictx.path_scope.factoring_allowlist.update(
+                ctx.iterator_path_ids)
         iterator_view = stmtctx.declare_view(
             iterator, qlstmt.iterator_alias,
+            factoring_fence=contains_dml,
+            temporary=False,
```
elprans, line 142: The underlying cause seems to be a bug in `attach_subtree`.  Removing `temporary` here helps in this particular case, but the bug is still there.  Here's an example that's broken and doesn't use `FOR`:

```
WITH MODULE test
SELECT count(
    (Card.name, (SELECT (SELECT Card)))
)
```

msullivan: Per slack discussion, the real broken example is
```
        WITH MODULE test
        SELECT (
            Card.name,
            (WITH X := (SELECT Card) SELECT X.name),
        )
```

1st1:
Hm, is this PR stale?

msullivan:
I'm closing this because I'm shortly submitting a PR that takes this approach (don't use `temporary` in `declare_view`) and applies it to all uses of declare_view, which fixes Elvis's issue as well.

Issue #1794, opened 2020-09-18
Mark optional types in the AST as Optional


msullivan:
A *lot* of fields in especially our qlast but also everywhere else are not marked Optional despite being optional. 

This causes two big problems:
1. We don't get typechecking related to the optionality
2. In a bunch of cases mypy will decide that branches are unreachable and not typecheck them at all.

Ideally, mypy would catch that we were putting None into these non-Optional fields, but we have evaded it.

msullivan:
I started fixing some of this and it might be less bad than I feared but I also discovered that schema.get basically always is inferred as returning None which seems pretty bad

elprans:
> schema.get basically always is inferred as returning None

Huh? I don't see that being the case.  `reveal_type` shows the correct type for a couple of call sites I checked.

PR #1807, opened 2020-09-22
Add a bunch more types to qlast

msullivan:
This causes way less grief than I feared it would.

Work on #1794. (Still want to look over irast too.)


elprans (approved):
I wonder if we should make an attrs-like mypy plugin to auto-generate constructor typings for AST nodes.

msullivan:
Yeah. Looking into that is on my TODO list.

PR #1809, opened 2020-09-23
Use workflow_dispatch instead of repository_dispatch

ambv:
This will allow us to test actions on branches more easily and doesn't require remembering the `curl` incantation to trigger a workflow.

### DIFF
--- a/.github/workflows.src/release.tpl.yml
+++ b/.github/workflows.src/release.tpl.yml
@@ -1,8 +1,8 @@
 name: Build Test and Publish a Release
 
 on:
-  repository_dispatch:
-    types: ["release"]
+  workflow_dispatch:
+    inputs: {}
 
 jobs:
 <% for tgt in targets.linux %>
@@ -12,15 +12,15 @@ jobs:
     steps:
     - name: Determine package version
       shell: bash
-      env:
-        BRANCH: "${{ github.event.client_payload.branch }}"
-      run: echo ::set-output name=version::${BRANCH#releases/}
+      run: |
+        echo ::set-output name=branch::${GITHUB_REF#refs/heads/}
+        echo ::set-output name=version::${GITHUB_REF#refs/heads/releases/}
       id: whichver
 
     - name: Build
       uses: edgedb/edgedb-pkg/integration/linux/build/<< tgt.name >>@master
       env:
-        SRC_REF: "${{ github.event.client_payload.branch }}"
+        SRC_REF: "${{ steps.whichver.outputs.branch }}"
         PKG_VERSION: "${{ steps.whichver.outputs.version }}"
         PKG_REVISION: "<current-date>"
         PKG_PLATFORM: "<< tgt.platform >>"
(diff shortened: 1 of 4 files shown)


.github/workflows.src/release.tpl.yml
```diff
@@ -12,15 +12,13 @@ jobs:
...
-        BRANCH: "${{ github.event.client_payload.branch }}"
-      run: echo ::set-output name=version::${BRANCH#releases/}
+      run: echo ::set-output name=version::${GITHUB_REF#releases/}
       id: whichver
 
     - name: Build
       uses: edgedb/edgedb-pkg/integration/linux/build/<< tgt.name >>@master
       env:
-        SRC_REF: "${{ github.event.client_payload.branch }}"
+        SRC_REF: $GITHUB_REF
```
ambv, line 21: Not sure if this will work? If not, we need to update edgedb-pkg to use GITHUB_REF instead.

.github/workflows.src/release.tpl.yml
```diff
@@ -12,15 +12,13 @@ jobs:
...
-      env:
-        BRANCH: "${{ github.event.client_payload.branch }}"
-      run: echo ::set-output name=version::${BRANCH#releases/}
+      run: echo ::set-output name=version::${GITHUB_REF#releases/}
       id: whichver
 
     - name: Build
       uses: edgedb/edgedb-pkg/integration/linux/build/<< tgt.name >>@master
       env:
-        SRC_REF: "${{ github.event.client_payload.branch }}"
```
elprans, line 23: I think `GITHUB_REF` has this format: `refs/heads/<branch-name>`

ambv: Ah, alright: ${GITHUB_REF#refs/heads/}

.github/workflows.src/release.tpl.yml
elprans: The correct syntax is "${{ env.GITHUB_REF }}", I think.

elprans: Also, you can normalize the branch name from `GITHUB_REF` in the "Determine package version" step, just add another output

.github/workflows.src/release.tpl.yml
```diff
@@ -12,15 +12,14 @@ jobs:
     steps:
     - name: Determine package version
       shell: bash
-      env:
-        BRANCH: "${{ github.event.client_payload.branch }}"
-      run: echo ::set-output name=version::${BRANCH#releases/}
+      run: echo ::set-output name=branch::${GITHUB_REF#refs/heads/}
```
elprans, line 15: I think you meant to do this:

```
run: |
    echo ::set-output name=branch::${GITHUB_REF#refs/heads/}
    echo ::set-output name=version::${GITHUB_REF#refs/heads/releases/}
```

ambv: Ha, I noticed while I was reading 😂

Screenshot 2020-09-23 at 19 00 42
Screenshot 2020-09-23 at 19 00 52

elprans (approved):

PR #1810, opened 2020-09-24
Use pointer object ID instead of name for table column names

elprans:
Currently, we use the name of a link or property as the name of a 
corresponding column in the PostgreSQL table.  When we switched to the use
of object IDs for table names, column names were left as is, because there
was still reliance on table inheritance.  Now that table inheritance was
replaced with direct use of `UNION`, the requirement of matching column
names no longer exists and we can switch to rename-resistant column mapping
instead.

### DIFF
--- a/edb/edgeql/compiler/context.py
+++ b/edb/edgeql/compiler/context.py
@@ -484,6 +484,9 @@ class ContextLevel(compiler.ContextLevel):
     defining_view: Optional[s_types.Type]
     """Whether a view is currently being defined (as opposed to be compiled)"""
 
+    compiling_update_shape: bool
+    """Whether an UPDATE shape is currently being compiled."""
+
     in_conditional: Optional[parsing.ParserContext]
     """Whether currently in a conditional branch."""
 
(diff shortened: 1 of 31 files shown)


tests/test_edgeql_ir_scopetree.py
```diff
@@ -37,8 +37,13 @@ class TestEdgeQLIRScopeTree(tb.BaseEdgeQLCompilerTest):
                           'cards.esdl')
 
     UUID_RE = re.compile(
```
1st1, line 39: Why is it called UUID_RE if it really matches something else?

1st1 (approved):

tests/test_edgeql_ir_scopetree.py
elprans: It used to match uuid literals exactly, but it now matches something else. I'll rename.

edb/edgeql/compiler/context.py
```diff
@@ -484,6 +484,9 @@ class ContextLevel(compiler.ContextLevel):
     defining_view: Optional[s_types.Type]
     """Whether a view is currently being defined (as opposed to be compiled)"""
 
+    compiling_update_shape: bool
```
msullivan, line 487: It looks like this is dead? Nothing seems to use it

elprans: Oops. Yes, it's part of an aborted attempt to condition the resolution of descendants in ptrrefs.

Issue #1812, opened 2020-09-24
InternalServerError: <Link 48b8eb51-fe77-11ea-bb24-0de48e5974e3 at 0x0x7ff9d977dc10>


isidentical:
```
asttests> SELECT array_agg((SELECT ast::Dict {keys} FILTER count(.keys) = 2 LIMIT 1).keys);
ERROR: InternalServerError: <Link 48b8eb51-fe77-11ea-bb24-0de48e5974e3 at 0x0x7ff9d977dc10>
  Hint: This is most likely a bug in EdgeDB. Please consider opening an issue ticket at https://github.com/edgedb/edgedb/issues/new?template=bug_report.md
  Server traceback:
      Traceback (most recent call last):
        File "/usr/lib/x86_64-linux-gnu/edgedb-server-1-alpha5/lib/python3.8/site-packages/edb/server/procpool/worker.py", line 75, in worker
          res = await meth(*args)
        File "/usr/lib/x86_64-linux-gnu/edgedb-server-1-alpha5/lib/python3.8/site-packages/edb/server/compiler/compiler.py", line 1795, in compile_eql_tokens
          return self._compile(ctx=ctx, tokens=eql_tokens)
        File "/usr/lib/x86_64-linux-gnu/edgedb-server-1-alpha5/lib/python3.8/site-packages/edb/server/compiler/compiler.py", line 1429, in _compile
          comp: dbstate.BaseQuery = self._compile_dispatch_ql(ctx, stmt)
        File "/usr/lib/x86_64-linux-gnu/edgedb-server-1-alpha5/lib/python3.8/site-packages/edb/server/compiler/compiler.py", line 1392, in _compile_dispatch_ql
          return self._compile_ql_query(ctx, ql)
        File "/usr/lib/x86_64-linux-gnu/edgedb-server-1-alpha5/lib/python3.8/site-packages/edb/server/compiler/compiler.py", line 628, in _compile_ql_query
          out_type_data, out_type_id = sertypes.TypeSerializer.describe(
        File "/usr/lib/x86_64-linux-gnu/edgedb-server-1-alpha5/lib/python3.8/site-packages/edb/server/compiler/sertypes.py", line 335, in describe
          type_id = builder._describe_type(
        File "/usr/lib/x86_64-linux-gnu/edgedb-server-1-alpha5/lib/python3.8/site-packages/edb/server/compiler/sertypes.py", line 172, in _describe_type
          subtypes = [self._describe_type(st, view_shapes,
[...]

PR #1814, opened 2020-09-24
Fix result type serialization crash

elprans:
If a pointer is defined in a sufficiently deeply nested query,
`view_shapes` wouldn't get populated for it, which is really a separate
issue, but the crux here is that `TypeSerializer` should not rely on
object being present in the `view_shapes` map at all.

Fixes: #1812.

### DIFF
--- a/edb/server/compiler/sertypes.py
+++ b/edb/server/compiler/sertypes.py
@@ -233,9 +233,9 @@ class TypeSerializer:
                 links.append(not ptr.is_property(self.schema))
 
             t_rptr = t.get_rptr(self.schema)
-            if t_rptr is not None:
+            if t_rptr is not None and (rptr_ptrs := view_shapes.get(t_rptr)):
                 # There are link properties in the mix
-                for ptr in view_shapes[t_rptr]:
+                for ptr in rptr_ptrs:
                     if ptr.singular(self.schema):
                         subtype_id = self._describe_type(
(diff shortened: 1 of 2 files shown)


edb/server/compiler/sertypes.py
```diff
@@ -233,9 +233,9 @@ def _describe_type(self, t, view_shapes, view_shapes_metadata,
                 links.append(not ptr.is_property(self.schema))
 
             t_rptr = t.get_rptr(self.schema)
-            if t_rptr is not None:
+            if t_rptr is not None and (rptr_ptrs := view_shapes.get(t_rptr)):
```
1st1, line 236: I'm not sure this is a correct fix... So there can be a type that has a shape but for some reason the type descriptor for it would be empty? If you believe that the fix is correct then add a detailed comment explaining why (more detailed than what's in the commit message).

elprans: It's not a type, it's a link, which would only have a shape if you've got link properties in your shape.  The reason why this code mostly works is because we use `defaultdict` in the compiler for the `view_shapes` map.  Other references to `view_shapes` in `sertypes` use `.get()` appropriately.

1st1: > It's not a type, it's a link

Ah, I see now

1st1 (approved):

Issue #1770, opened 2020-09-13
edgedb.errors.InternalServerError: could not serialize result in worker subprocess


isidentical:
This happens ~oftenly while I'm inserting tons of nodes. I don't know whether it is relevant or not, but I have a multiprocessing setup (ProcessPoolExecutor) which I insert using 1 transaction per process, and the transactions include the creation up to 10k nodes. (setup is the same with my other issues)

```
Traceback (most recent call last):
  File "/home/isidentical/reiz.io/reiz/db/insert.py", line 121, in insert
    if conversion := convert(child, connection):
  File "/home/isidentical/reiz.io/reiz/db/insert.py", line 105, in convert
    ", ".join(convert(value, connection) for value in child), "{}"
  File "/home/isidentical/reiz.io/reiz/db/insert.py", line 105, in <genexpr>
    ", ".join(convert(value, connection) for value in child), "{}"
  File "/home/isidentical/reiz.io/reiz/db/insert.py", line 99, in convert
    {"id": "<uuid>" + repr(str(insert(child, connection).id))},
  File "/home/isidentical/reiz.io/reiz/db/insert.py", line 124, in insert
    return connection.query_one(query)
  File "/home/isidentical/.venv38/lib/python3.8/site-packages/edgedb/blocking_con.py", line 108, in query_one
    return self._protocol.sync_execute_anonymous(
  File "edgedb/protocol/blocking_proto.pyx", line 91, in edgedb.protocol.blocking_proto.BlockingIOProtocol.sync_execute_anonymous
  File "edgedb/protocol/blocking_proto.pyx", line 77, in edgedb.protocol.blocking_proto.BlockingIOProtocol._iter_coroutine
  File "edgedb/protocol/protocol.pyx", line 526, in execute_anonymous
  File "edgedb/protocol/protocol.pyx", line 220, in _parse
edgedb.errors.InternalServerError: could not serialize result in worker subprocess
```

isidentical:
From the same context (if needed can share some queries, but would be tricky to run them since they are all auto-generated, long, and multi-step things). Just sharing if it would make sense since I am unable to create a short reproducer, can close the issue if it doesn't make any sense.
```
Traceback (most recent call last):
  File "/home/isidentical/reiz.io/reiz/db/insert.py", line 158, in inject_project
    with connection.transaction():
  File "/home/isidentical/.venv38/lib/python3.8/site-packages/edgedb/blocking_con.py", line 108, in query_one
[...]

Issue #1334, opened 2020-04-12
Database breaking randomly


gsouf:
- EdgeDB Version: 1.0-alpha.2+g16f75ece5.d20200122 
- OS Version: Running in docker - Host unbuntu 19.10

Steps to Reproduce:

Cannot say, it produced randomly with a database that was working well. The db broke, the js client returned this error:

```
code: 16777216
message: "cannot issue new command"
type: "InternalServerError"
```

Stoping a restarting the docker instance solved the issue.

Please let me know how I can provide more details if it occurs again

gsouf:
I believe I found what happens. 

First scalar types breaks graphql. Second, removing the scalar type breaks https interfaces.

Not related to the JS client. As I said in my initial message.

 Here are the steps:

1. create a scalar type:

```
CREATE scalar type FooOrBar extending enum<"foo", "bar">;
```

2. Try to load something via graphql. Graphql is broken after adding this scalar type.

3. Remove the scalar type.

4. Try to load something via graphql or edgeql over http.

elprans:
This seems to be a problem specific to HTTP ports: #1314

gsouf:
@elprans it happened again today after I ran some mutation on my database. Some of them contained syntax errors.
It's it's related to concurent connexion then it might be because some errors don't have a proper handling and cause connection to remain active (or believed to be active). 

I can run more tests if required.

ambv:
Reproduced this as well. Agree with @elvis that this looks like a duplicate of #1314.

1st1:
@ambv Are you looking into this one or should I look at this?

ambv:
I didn't do anything here besides reproducing the problem. Feel free to assign to me.

1st1:
> I didn't do anything here besides reproducing the problem. Feel free to assign to me.

It's OK, please take a look tomorrow.

Issue #1314, opened 2020-03-30
Cannot issue new command


haikyuu:
- EdgeDB Version: `EdgeDB 1.0-alpha.3+dev.102.g176dac70`
- OS Version: Ubuntu

Steps to Reproduce:

1. Expose an HTTP Port 
2. Send 2-3 requests at the same time

```
{
  message: 'cannot issue new command',
  type: 'InternalServerError',
   code: 16777216
 }
```

elprans:
Does this happen on a particular query for you or regardless of a query?

1st1:
> Send 2-3 requests at the same time

What's your `concurrency` setting?

haikyuu:
Concurrency is set to 4

haikyuu:
But 4 doesn't mean 4 requests at the same time max. Does it?

haikyuu:
> Does this happen on a particular query for you or regardless of a query?

I did a `reset Port` and created a new Port and i was able to send queries just fine.

It seems some errors mess up with Ports.

haikyuu:
When that happens, all queries return that same error. Even if i send one query.

1st1:
> But 4 doesn't mean 4 requests at the same time max. Does it?

It means that only 4 previously unseen queries can be compiled in parallel. But in any case, the server shouldn't error with what you see, so this is most definitely a bug. I'll investigate.

haikyuu:
@1st1 I think i've found a reproduction scenario.

- Create HTTP Port.
- Make requests
- Alter schema
- `cannot issue new command`

ambv:
I can reproduce this. After altering schema, the subsequent (compiled) query works once more, then fails with ISE `"unexpected message type '3'"` and subsequent requests fail with ISE `"cannot issue new command"`.

PR #1817, opened 2020-09-24
Handle CloseComplete Postgres response in _parse_execute_to_buf

elprans:
The `_parse_execute_to_buf`, which is used for JSON-based queries is
currently forgetting to handle the `CloseComplete` (3) response message
from Postgres despite engaging in prepared statement cleanup like its
cousin `_parse_execute` (which handles `CloseComplete` correctly).  This
leads to crashes when the prepared statement cache buffer overflows, and
is most prominent on HTTP ports, however it's also possible to trigger
this via a normal connection, because prepared statements also get
invalidated on schema version changes, as @vpetrovykh uncovered recently
in #1780.

Fixes: #1334
Fixes: #1314
Fixes: #1770

### DIFF
--- a/edb/server/pgcon/pgcon.pyx
+++ b/edb/server/pgcon/pgcon.pyx
@@ -368,11 +368,12 @@ cdef class PGProto:
                     if store_stmt:
                         self.prep_stmts[stmt_name] = dbver
 
-                elif mtype in {b'C', b'n', b'2', b'I'}:
+                elif mtype in {b'C', b'n', b'2', b'I', b'3'}:
                     # CommandComplete
                     # NoData
                     # BindComplete
                     # EmptyQueryResponse
+                    # CloseComplete
                     self.buffer.discard_message()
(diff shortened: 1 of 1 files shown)


1st1 (approved):
We do need to ignore this message, so LGTM. But I really don't understand this part:

>  This leads to crashes when the prepared statement cache buffer overflows

does this undiscarded message lead to some error that prevents the cleanup?

elprans:
> does this undiscarded message lead to some error that prevents the cleanup?

This undiscarded message leads to a protocol crash in `fallthrough`.  It is _triggered_ by an overflow in prepared statement cache, because that's when `Close` is issued in `_parse_execute` (in `before_prepare()`)

1st1:
>  It is triggered by an overflow in prepared statement cache, because that's when Close is issued in _parse_execute (in before_prepare())

Ah, OK, I get it now.

PR #1829, opened 2020-09-26
Add std::throw() and 'never' pseudo-type.

1st1:
@elprans This is draft, but I'd appreciate you reviewing it now so that I know if I'm on the right path.

The PR adds a new `never` pseudo-type and a `std::throw()` function.

### TODO

- [x] Add `never`
- [x] Add `std::throw`
- [ ] Add basic `std::throw` tests
- [ ] Add tests that `never` is castable to object types and collections
- [ ] Add tests that `never` can't be a valid parameter type, a valid property target type, etc
- [ ] Add `never` to `api/types.txt` -- type of `SELECT std::throw()` should be `never`
- [ ] Add tests that `std::throw()` can appear virtually in every kind of expression

### Motivation

**Runtime assertions:**

```
WITH
    MODULE schema
SELECT Type {
    id,
    name,
    kind := 'object' IF Type IS ObjectType ELSE
            'scalar' IF Type IS ScalarType ELSE
            'array' IF Type IS Array ELSE
            'tuple' IF Type IS Tuple ELSE
            std::throw('unknown type kind'),
    [IS ScalarType].enum_values,
    [IS ObjectType].union_of,
    # ...
}
```

The above would fail if we add new kind of type in future EdgeDB versions. That allows, for example, to use discriminated union types in Python or TypeScript, making `enum_values` to be only available when `kind = 'scalar'`. We could, of course, return `'unknown'` instead of using `std::throw` and handle that in the client code, but that's way less elegant.

**Combating empty sets:**

```
SELECT (
    (SELECT schema::Type {id, name} FILTER .name = <str>$name).id ?? 
    throw('could not resolve type: ' ++ <str>$name)
)
```

The query would return an ID of a type by name and throw an explicit user friendly exception if no type is found.

### DIFF
--- a/edb/edgeql-parser/src/keywords.rs
+++ b/edb/edgeql-parser/src/keywords.rs
@@ -163,6 +163,7 @@ pub const CURRENT_RESERVED_KEYWORDS: &[&str] = &[
     "like",
     "limit",
     "module",
+    "never",
     "not",
     "offset",
     "optional",
--- a/edb/api/errors.txt
+++ b/edb/api/errors.txt
@@ -110,6 +110,8 @@
 0x_05_03_00_01   TransactionSerializationError
 0x_05_03_00_02   TransactionDeadlockError
 
+0x_05_04_00_00   RuntimeUserError
+
 
 ####
 
--- a/edb/edgeql-parser/src/tokenizer.rs
+++ b/edb/edgeql-parser/src/tokenizer.rs
@@ -885,6 +885,7 @@ pub fn is_keyword(s: &str) -> bool {
         | "lock"
         | "match"
         | "move"
+        | "never"
         | "notify"
         | "prepare"
         | "partition"
--- a/edb/edgeql/ast.py
+++ b/edb/edgeql/ast.py
@@ -228,6 +228,10 @@ class AnyTuple(PseudoObjectRef):
     pass
 
 
+class Never(PseudoObjectRef):
+    pass
+
+
 class Anchor(Expr):
     __abstract_node__ = True
     name: str
--- a/edb/edgeql/codegen.py
+++ b/edb/edgeql/codegen.py
@@ -672,6 +672,9 @@ class EdgeQLSourceGenerator(codegen.SourceGenerator):
     def visit_AnyTuple(self, node: qlast.AnyTuple) -> None:
         self.write('anytuple')
 
+    def visit_Never(self, node: qlast.AnyType) -> None:
+        self.write('never')
+
     def visit_TypeCast(self, node: qlast.TypeCast) -> None:
         self.write('<')
         self.visit(node.type)
--- a/edb/edgeql/compiler/func.py
+++ b/edb/edgeql/compiler/func.py
@@ -761,21 +761,38 @@ def finalize_args(
             assert isinstance(paramtype, s_types.Array)
             paramtype = list(paramtype.get_subtypes(ctx.env.schema))[0]
 
-        # Check if we need to cast the argument value before passing
-        # it to the callable.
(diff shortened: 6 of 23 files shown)


edb/edgeql-parser/src/keywords.rs
```diff
@@ -163,6 +163,7 @@ pub const CURRENT_RESERVED_KEYWORDS: &[&str] = &[
     "like",
     "limit",
     "module",
+    "never",
```
1st1, line 166: Perhaps we should add it as an abstract final `std::never` type, but that would require some hacks.

1st1: OTOH `never` isn't exactly a popular property/link/function/parameter name.

tailhook:
The motivation section probably lacks the following: `never` is a return type of the `std::throw`.

I like this thing, but I'm worrying that we can have some same issues that Rust has with never (i.e. `!` in Rust) being a full-featured type (it isn't yet in stable Rust, but parts of its functionality are there). I'm not sure I can summarize these issues, though. And of course, it doesn't mean they apply to EdgeDB but I think this shouldn't be treated lightly.

Why we need to do it now? Is there anything that can't be added later in backwards-compatible manner?

Also there is an issue with exceptions in postgres that make valid requests fail due to some inlining and eager evaluation. @elprans, we probably should have some link on this because this pops up quite often, and I fail to judge if this could be a problem in each case (i.e. my intuition now is that we must avoid exceptions as much as possible).

elprans:
> Also there is an issue with exceptions in postgres that make valid requests fail due to some inlining and eager evaluation. 

This isn't specific to the proposed `throw()`.  Any fallible operation has the same issue.  The real fix, and a prerequisite for `throw()` would be to start tracking expression fallibility properly and generate appropriate code in conditionals that guarantees conditional execution.

1st1:
> i.e. my intuition now is that we must avoid exceptions as much as possible

I don't think I agree with that. A query can fail with a variety of reasons (zero division is the simplest example) and exceptions are essential in Postgres' and EdgeDB's execution model. 

That said, the main purpose of `std::throw()` is to actually never throw, unless there's something that really breaks the contract between the query and the code that executes it.

> Why we need to do it now? Is there anything that can't be added later in backwards-compatible manner?

Sure, but the same logic applies to every new feature. I wanted this to happen for a long period of time, and while working on it I discovered that we might want to fix a few things, like make `anytype` not a reserved keyword and better define EdgeQL spec, w.r.t. how coalesce treats fallible operands. Overall I think this feature is essential to combat empty sets quirks (the old and surprising SQL's eager NULL behavior), so I'd like this to be a 1.0 feature (if that's realistic and manageable, of course). The priority isn't high though.

tailhook:
> > i.e. my intuition now is that we must avoid exceptions as much as possible
> 
> I don't think I agree with that. A query can fail with a variety of reasons (zero division is the simplest example) and exceptions are essential in Postgres' and EdgeDB's execution model.
> 
> That said, the main purpose of `std::throw()` is to actually never throw, unless there's something that really breaks the contract between the query and the code that executes it.
> 

I agree with all of that. My concern is only that postgres doesn't work with exceptions well, but @elprans's answer explains what we have to do to make them reliable. The problem is that it sounds like quite heavyweight.

> > Why we need to do it now? Is there anything that can't be added later in backwards-compatible manner?
> 
> Sure, but the same logic applies to every new feature. 

Yes, that's exactly my point: consistently apply the rule to every new feature at this moment.

1st1:
### We'll likely implement the https://github.com/edgedb/edgedb/discussions/1849 proposal instead of this PR. This can be closed for now.

Issue #1784, opened 2020-09-17
Inconsistency with volatile calls


msullivan:
1) `FOR x IN {1, 2, 3} UNION (x, random());` returns identical values
2) `FOR x IN {1, 2, 3} UNION (SELECT (x, random()));` returns different values
3) `SELECT (FOR x IN {1, 2, 3} UNION (SELECT (x, random())));` returns identical values (!)

I /think/ that all 3 ought to return different values but 2 and 3 definitely should do the same thing.

1st1:
Yeah, this is a known issue, see also https://github.com/edgedb/edgedb/issues/492.

It would be great if we were always consistent when volatile funcs are used without `CORRELATED TO`. Perhaps we should remove hacks that @elprans added to make edge cases work, implement the new syntax, and educate users.

msullivan:
Yeah, I hit this while thinking about #492. 

My intuition about it was that it clearly makes sense according to our semantics that `SELECT (Object, random())` should always return the same value (and, in fact, it would be pretty wrong if it didn't!), the obvious seeming interpretation of our semantics is that `FOR x IN {1, 2, 3} UNION (x, random())` should return different values, and so it should be possible to accomplish whatever we need to do without new syntax. So I figured I'd work through some examples of what workarounds were needed, which would help illuminate to me what is needed from sugar to make it easier.

(Thinking about it a little more, the most obvious reading of our semantics I think also implies that ORDER BY random() ought to work, too, though.)

(There's a plausible semantics where it is *always* allowed to get factored out, also, but my understanding is there are cases where we need that to not be the case. And it's less nice, IMO.)

elprans:
>  Perhaps we should remove hacks

Those aren't hacks.  Our query execution model should be well-defined for things to mostly work intuitively without explicit `CORRELATED` hints.  Specifically, when iterating over an expression that serves as a shape source, we must evaluate all expressions in the shape `N` times, where `N` is the cardinality of the source set (including, possibly 0).  Same applies to `FOR` expressions.  So, in this case, I propose to extend implicit `CORRELATED TO` injection to volatiles inside the body of `FOR`.  Should be a relatively simple fix.

elprans:
> I think also implies that ORDER BY random() ought to work, too, though.

[...]

PR #1836, opened 2020-09-29
Correlate volatile functions with enclosing FORs and in clauses

msullivan:
This means that things like `ORDER BY random()` and
`FILTER random() < 0.5` will work.

Part of #1784.

### DIFF
--- a/edb/pgsql/compiler/clauses.py
+++ b/edb/pgsql/compiler/clauses.py
@@ -120,6 +120,9 @@ def setup_iterator_volatility(
     path_id = iterator.path_id
     ref: Optional[pgast.BaseExpr] = None
 
+    # We use a callback scheme here to avoid inserting volatility ref
+    # columns unless there is actually a volatile operation that
+    # requires it.
     def get_ref() -> pgast.BaseExpr:
         nonlocal ref
         if ref is None:
--- a/edb/pgsql/compiler/clauses.py
+++ b/edb/pgsql/compiler/clauses.py
@@ -77,6 +77,58 @@ def fini_stmt(
                 )
 
 
+def get_volatility_ref(
+        path_id: irast.PathId, *,
+        ctx: context.CompilerContextLevel) -> pgast.BaseExpr:
+    """Produce an appropriate volatility_ref from a path_id."""
+
+    ref: Optional[pgast.BaseExpr] = relctx.maybe_get_path_var(
+        ctx.rel, path_id, aspect='identity', ctx=ctx)
+    if not ref:
+        rvar = relctx.maybe_get_path_rvar(
+            ctx.rel, path_id, aspect='value', ctx=ctx)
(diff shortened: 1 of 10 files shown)


edb/pgsql/compiler/relgen.py
```diff
@@ -1954,18 +1956,19 @@ def _compile_func_epilogue(
...
-        # Apply the volatility reference.
-        # See the comment in process_set_as_subquery().
-        func_rel.where_clause = astutils.extend_binop(
-            func_rel.where_clause,
-            pgast.NullTest(
-                arg=ctx.volatility_ref,
-                negated=True,
+    if expr.volatility is qltypes.Volatility.VOLATILE:
+        for ref in ctx.volatility_ref:
+            if isinstance(ref, context.NoVolatilitySentinel):
```
elprans, line 1961: I didn't like the `NO_VOLATILITY` hack very much before, but in this formulation I like it even less.  Perhaps we could track the `INSERT` subject `path_id` instead and condition computable correlation in `process_set_as_subquery` on that instead.

edb/pgsql/compiler/clauses.py
```diff
@@ -77,6 +77,58 @@ def fini_stmt(
...
+        ctx: context.CompilerContextLevel) -> None:
+    if iterator is None:
+        return
+
+    old = () if is_cte else ctx.volatility_ref
+
+    path_id = iterator.path_id
+    ref: Optional[pgast.BaseExpr] = None
+
+    def get_ref() -> pgast.BaseExpr:
```
elprans, line 123: Can you explain why a callback approach is needed, please?

msullivan: I touched on it in the comment I added in `context.py` but could add a comment here too: it's so we don't add the volatility ref column to the relation unless it actually gets used.

edb/pgsql/compiler/clauses.py
```diff
@@ -96,8 +150,12 @@ def compile_iterator_expr(
         # for path identity of the iterator expression.  This is
         # necessary to maintain correct correlation for the state
         # of iteration in DML statements.
-        relctx.ensure_bond_for_expr(
-            iterator_expr.expr.result, iterator_query, type='uuid', ctx=subctx)
+        # The already_existed check is to avoid adding in bogus volatility refs
+        # when we reprocess an iterator that was hoisted.
+        if not already_existed:
```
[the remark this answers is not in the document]
msullivan: I don't see how at a quick glance, at least

edb/pgsql/compiler/relgen.py
msullivan: Yeah, I think that should work.

msullivan:
I think this should be good now?

elprans (approved):
Yep, looks good.

Issue #1833, opened 2020-09-27
Enum improvements


1st1:
### Introduction

While working on https://github.com/edgedb/edgedb/pull/1829, I discovered that we need to make our error types and codes introspectable just like we allow to introspect all aspects of EdgeDB schemas and config.

Our errors are currently defined in https://github.com/edgedb/edgedb/blob/master/edb/api/errors.txt and that file is used to auto-generate code on the server with `edb gen-errors`. When we add a new exception kind we need to re-run `edb gen-errors` for EdgeDB itself, but also to recompile tools like `edgedb` REPL so that it knows about the newly added error codes.

Lastly, it would be nice if `std::throw()` (added in #1829) could be told what exact error code it should use. Asking users to specify error codes by their int value wouldn't be a good UX; it would be much better if they could say `std::throw('message', error := <sys::Errors>'InvalidValueError')`.

We can of course define an enum in EdgeDB, but current enums are string enums. To store the actual error codes we'd need to provide a special function like `sys::get_error_code(error: sys::Errors)` or have a `sys::ErrorCodes { name: sys::Errors, code: int64 }` object type. Both solutions are suboptimal.

This issue proposes adding two new things:

1. Integer enums
2. New syntax for enum literals

### Integer Enums

I propose to allow defining integer enums with the following syntax:

```
CREATE SCALAR TYPE sys::Errors EXTENDING enum<
  InternalServerError := 0x_01_00_00_00, 
  UnsupportedFeatureError := 0x_02_00_00_00,
  # ...
>;
```

We already use a similar syntax to type named tuples. When an int enum is defined we can automatically create internal functions that can cast integers to our enum, or enum names to integers; those functions can be fast and use a JSON map internally to make lookups efficient.

### New syntax for enum literals

The current syntax for enums `SELECT <schema::Cardinality>'ONE'` feels verbose and hacky. I propose to treat enums like special singleton objects and allow using paths syntax to reference them, as in:

```
SELECT schema::Cardinality.ONE;
```

This new syntax would make working with enums in EdgeQL just as easy and natural as in high-level programming languages. It would also play nice with various query builders for EdgeQL.

For string enums, the value of the _enum path_ would be a string subtype, as in:

```
db> SELECT <str>schema::Cardinality.ONE;
{'ONE'}
```

and for string enums the value would be int-castable:

```
db> SELECT <int64>sys::Errors.InternalServerError;
{16777216}   # = 0x_01_00_00_00
```

-----

[...]

PR #1843, opened 2020-09-29
Implement ident-based syntax for enums

1st1:
Issue https://github.com/edgedb/edgedb/issues/1833


edb/pgsql/metaschema.py
```diff
@@ -2230,7 +2230,12 @@ def __init__(self) -> None:
 class SysGetTransactionIsolation(dbops.Function):
     "Get transaction isolation value as text compatible with EdgeDB's enum."
     text = r'''
-        SELECT upper(setting)
+        SELECT
+            CASE setting
+                WHEN 'repeatable read' THEN 'RepeatableRead'
+                WHEN 'serializable' THEN 'Serializable'
+                ELSE 'Unknown'  -- should never happen
```
elprans, line 2237: Raise an exception here instead.

1st1: done

elprans (approved):

Issue #1772, opened 2020-09-14
Migration features


vpetrovykh:
The general idea is that anything with a name should support renaming. We also want creation and deletion work for all items (this would make migrations reliable at least in the simple cases).

Create/drop:
- [x] abstract annotation
- [x] annotation
- [x] abstract constraint
- [x] property constraint
- [x] link constraint 
- [x] type constraint 
- [x] scalar type
- [x] enum type
- [x] object type
- [x] abstract link
- [x] link
- [x] abstract property 
- [x] property 
- [x] link property 
- [x] alias
- [x] index 
- [x] function

Rename:
- [x] abstract annotation
- [x] annotation
- [x] abstract constraint
- [x] property constraint
- [x] link constraint 
- [x] type constraint 
- [x] scalar type
- [x] enum type
- [x] object type
- [x] abstract link
- [x] link
- [x] abstract property 
- [x] property 
- [x] link property 
- [x] alias
- [x] function

Additionally we want to alter some things without a name change when they are already in use:
- [x] alter alias expression 
- [x] alter function body
- [x] alter computable link
- [x] alter computable property 
- [x] add a type to extending 
- [x] remove a type from extending 
- [x] change required status
- [x] change from single to multi property 
- [x] change from single to multi link

1st1:
cc @elprans @msullivan

msullivan:
:firework: :tada:

PR #1844, opened 2020-09-30
Convert data migration tests to also test DESCRIBE MIGRATION.

vpetrovykh:
Most of the data migration tests now also apply the migration
in a step-by-step fashion based on the DESCRIBE MIGRATION output.

Issue #1772


elprans (commented):
Most of the new xfails ("no edgeql class") should be trivially fixable, let's fix them as part of this PR to avoid masking a large chunk of tests and letting them rot.

elprans (approved):

PR #1846, opened 2020-09-30
Add a command to generate dumps for tests

### LINKED ISSUE #1736: Generate test dumps for a4 and a5 releases
We need to generate dumps for a4 and a5: https://github.com/edgedb/edgedb/tree/master/tests/dumps/dump01

Would be great if we could automate that with some subcommand under the `edb` tool. Also need a checklist for releases.

ambv:
Includes all test cases where the database name starts with `dump`.

Fixes #1736


1st1 (commented):
* The `TestDump01_CLI` class should be renamed to `TestDump01`
* I'd add a `.gitignore` file to ignore `*_dev*.dump` files so that they can't be accidentally committed
* IMO we're not yet ready to start committing dump02 files. So I'd add a special class-level attribute to dump test cases like `STABLE_DUMP_CHECK = True`. So far it will only be set on the `Test01` case.

ambv:
All historical dumps are now enabled and work.

1st1 (approved):

Issue #1830, opened 2020-09-27
Opaque unions produced by naked backlinks aren't handled correctly in the schema


elprans:
Per report in #1824, the following schema produces `SchemaError: cannot get 'name' value: item '<uuid>' is not present in the schema`:

```
abstract type Action {
   required link user -> User;
}

type Post extending Action;

type User {
  posts := .<user;
}
```

PR #1847, opened 2020-09-30
Fix handling of computables with bare backlinks

elprans:
The bare backlink computable (e.g. `.<link`) should resolve into
`std::BaseObject`, while still being handled as an appropriate opaque union
internally.

Fixes: #1830

### DIFF
--- a/edb/schema/functions.py
+++ b/edb/schema/functions.py
@@ -965,19 +965,19 @@ class Function(CallableObject, VolatilitySubject, s_abc.Function,
         qlast.Language, default=None, compcoef=0.4, coerce=True)
 
     from_function = so.SchemaField(
-        str, default=None, compcoef=0.4, introspectable=False)
+        str, default=None, compcoef=0.4)
 
     from_expr = so.SchemaField(
-        bool, default=False, compcoef=0.4, introspectable=False)
+        bool, default=False, compcoef=0.4)
 
     force_return_cast = so.SchemaField(
-        bool, default=False, compcoef=0.9, introspectable=False)
+        bool, default=False, compcoef=0.9)
 
     sql_func_has_out_params = so.SchemaField(
-        bool, default=False, compcoef=0.9, introspectable=False)
+        bool, default=False, compcoef=0.9)
 
     error_on_null_result = so.SchemaField(
-        str, default=None, compcoef=0.9, introspectable=False)
+        str, default=None, compcoef=0.9)
 
     initial_value = so.SchemaField(
         expr.Expression, default=None, compcoef=0.4, coerce=True)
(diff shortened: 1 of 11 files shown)


edb/schema/functions.py
```diff
@@ -965,19 +965,19 @@ class Function(CallableObject, VolatilitySubject, s_abc.Function,
...
-        bool, default=False, compcoef=0.9, introspectable=False)
+        bool, default=False, compcoef=0.9)
 
     sql_func_has_out_params = so.SchemaField(
-        bool, default=False, compcoef=0.9, introspectable=False)
+        bool, default=False, compcoef=0.9)
 
     error_on_null_result = so.SchemaField(
-        str, default=None, compcoef=0.9, introspectable=False)
+        str, default=None, compcoef=0.9)
```
1st1, line 980: Does this mean that all these fields are exposed to the `schema::`?

1st1: Or we now only expose what's specified in `schema.edgeql` and the rest is automatically hidden?

1st1: NVM, I should have started with reviewing individual commits.

edb/schema/types.py
```diff
@@ -1918,6 +1940,16 @@ def get_ast(
     ) -> Optional[qlast.DDLOperation]:
         if self.get_attribute_value('expr'):
             return None
+        elif (
+            (union_of := self.get_attribute_value('union_of')) is not None
```
1st1, line 1944: nice use of `:=`. I'm still not used to it.

1st1 (approved):

edb/schema/pointers.py
```diff
@@ -865,7 +865,7 @@ def _parse_computable(
         expr: qlast.Base,
         schema: s_schema.Schema,
         context: sd.CommandContext,
-    ) -> Tuple[s_schema.Schema, s_types.Type, Optional[PointerLike]]:
+    ) -> Tuple[s_schema.Schema, s_types.TypeShell, Optional[PointerLike]]:
```
msullivan, line 868: (What's a type shell?)

msullivan (approved):
This seems plausible to me but also I don't really understand any of the relevant schema components.

edb/schema/pointers.py
elprans: When we build a schema delta IR tree, we cannot refer to actual schema objects directly, because they might not exist until we actually run the IR.  So, we use `ObjectShell` and its derivatives.  It's somewhat analogous to `TypeRef` in EdgeQL IR, except we don't usually need to fully describe the underlying object in a `Shell`, only the means to properly look it up in the schema once available.

PR #1851, opened 2020-10-02
Eliminate deferred cardinality checking

msullivan:
Make the cardinality checker do a full top-down cardinality inference
of all parts of the tree, performing cardinality checks and filling in
cardinality fields in the AST.

This allows us to ditch all of the at_stmt_fini machinery and
once_pointer_cardinality_is_inferred machinery.

It fixes some incorrect inference of function/operator argument
cardinalities that had been causing me trouble when trying to do
checks for volatility cardinality restrictions; these were cauesd by
not being able to find the proper scope in the callback.

### DIFF
--- a/edb/edgeql/compiler/inference/cardinality.py
+++ b/edb/edgeql/compiler/inference/cardinality.py
@@ -322,15 +322,6 @@ def _find_visible(
         else:
             path_id = ir.path_id
 
-        # This is a nasty hack, I think. If a path is an unfenced
-        # descendant, don't find a visible node above us. This fixes
-        # issues like in test_edgeql_scope_filter_01 where we mean to
-        # refer to our ought-to-be-namespaced child but instead find
-        # something else.
-        if ((nobe := scope_tree.find_descendant(path_id))
-                and nobe.fence == scope_tree.fence):
-            return None
-
         return parent_fence.find_visible(path_id)
     else:
         return None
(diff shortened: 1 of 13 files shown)


edb/edgeql/compiler/inference/cardinality.py
```diff
@@ -284,32 +322,248 @@ def _find_visible(
         else:
             path_id = ir.path_id
 
+        # This is a nasty hack, I think. If a path is an unfenced
+        # descendant, don't find a visible node above us. This fixes
+        # issues like in test_edgeql_scope_filter_01 where we mean to
+        # refer to our ought-to-be-namespaced child but instead find
+        # something else.
```
msullivan, line 329: Or maybe we should just xfail test_edgeql_scope_filter_01?

elprans: Hm. Yea, the `strip_weak_namespaces` business is all wrong.  What happens if you run cardinality inference _before_ `strip_weak_namespaces`?

msullivan: Yeah, that works!

I had actually tried that earlier, and it didn't work, but when I went back so I could explain what didn't work about it, and it turns out I had moved cardinality inference to happen in between the two parts of stripping weak namespaces, which unsurprisingly works less well than doing it either before or after.

elprans (approved):
:+1:

PR #1853, opened 2020-10-02
Prohibit taking implicit cross products with volatile operations

### LINKED ISSUE #1784: Inconsistency with volatile calls

msullivan:
I wanted error spans on these messages, which meant threading around
a few more contexts in the compiler.

I needed to make the bodies of FOR scoped in order to have the
iterator variable recognized as having single cardinality inside the
body when the body is just an expression and not an explicit subquery.

Fixes #1784.

### DIFF
--- a/edb/edgeql/compiler/inference/cardinality.py
+++ b/edb/edgeql/compiler/inference/cardinality.py
@@ -43,6 +43,8 @@ from edb.ir import utils as irutils
 from edb.ir import typeutils
 from edb.edgeql import ast as qlast
 
+from . import volatility
+
 from .. import context
 
 if TYPE_CHECKING:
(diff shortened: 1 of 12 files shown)


edb/edgeql/compiler/stmt.py
```diff
@@ -181,14 +181,19 @@ def compile_ForQuery(
...
-            view_scls=ctx.view_scls,
-            view_rptr=ctx.view_rptr,
-            result_alias=qlstmt.result_alias,
-            view_name=ctx.toplevel_result_view_name,
-            forward_rptr=True,
-            ctx=sctx)
+        with sctx.newscope(fenced=True) as bctx:
+            stmt.result = setgen.scoped_set(
+                compile_result_clause(
+                    qlstmt.result,
```
elprans, line 187: `astutils.ensure_qlstmt(qlstmt.result)` here should be more efficient as it wouldn't introduce extraneous fences for expressions that are already statements.

edb/edgeql/compiler/inference/cardinality.py
```diff
@@ -186,20 +188,49 @@ def _coalesce_cardinality(
...
+    # check this by assuming that a voltile operation is AT_MOST_ONE
+    # and making sure that the resulting cartesian cardinality isn't
+    # multi.
+    for i, vol in enumerate(vols):
+        if vol == VOLATILE:
+            cards2 = list(cards)
+            cards2[i] = AT_MOST_ONE
+            if cartesian_cardinality(cards2).is_multi():
+                raise errors.QueryError(
+                    "can not take cross product of volatile operation",
```
elprans, line 212: This also seems to trip up even if I put the volatile in `WITH`:

```
WITH rand := random() SELECT ({1, 2}, rand)
```

edb/edgeql/compiler/stmt.py
msullivan: More efficient in what sense? Compiler efficiency, or will the generated code be improved?

`ensure_qlstmt` breaks a few tests---I haven't dug into it yet though.

elprans: I was mainly concerned with scope tree complexity, but the benefit here is negligible.  Let's keep your initial version then.  Technically, `ensure_qlstmt` should be equivalent to setting up a fence, but there's probably some weird interaction with iterator state.

elprans (approved):

PR #1861, opened 2020-10-06
Move a bunch of the arguments in cardinality inference into a context

msullivan:
This cleans things up a bit and makes it easier to extend in the
future.

### DIFF
--- a/edb/edgeql/compiler/inference/cardinality.py
+++ b/edb/edgeql/compiler/inference/cardinality.py
@@ -55,6 +55,14 @@ MANY = qltypes.Cardinality.MANY
 AT_LEAST_ONE = qltypes.Cardinality.AT_LEAST_ONE
 
 
+class CardCtx(NamedTuple):
+    env: context.Environment
+    inferred_cardinality: Dict[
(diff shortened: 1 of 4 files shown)


edb/edgeql/compiler/inference/cardinality.py
```diff
@@ -55,6 +55,14 @@
 AT_LEAST_ONE = qltypes.Cardinality.AT_LEAST_ONE
 
 
+class CardCtx(NamedTuple):
+    env: context.Environment
+    inferred_cardinality: Dict[
+        Tuple[irast.Base, irast.ScopeTreeNode],
+        qltypes.Cardinality]
+    singletons: Collection[irast.PathId]
```
elprans, line 63: Am I missing something or is `singletons` not used anymore?

msullivan: Yeah, it doesn't seem to be used. One of the goals of this is so that changes like removing singletons could be less noisy. I could drop it in this PR or the next.

elprans (approved):

PR #1863, opened 2020-10-07
Add alpha 6 changelog.

vpetrovykh:


docs/changelog/1_0_a6.rst
```diff
@@ -0,0 +1,63 @@
...
+======
+
+* Introduce new enum definition syntax (:eql:gh:`#1843`).
+
+  Instead of using strings, the enums are now defined using identifiers:
+
+  .. code-block:: edgeql
+
+    CREATE SCALAR TYPE schema::Cardinality
+        EXTENDING enum<ONE, MANY>;
```
tailhook, line 19: Example should show CamelCase option, since this is our new convention.

docs/changelog/1_0_a6.rst
```diff
@@ -0,0 +1,63 @@
...
+
+* Rename built-in schema enums to use CamelCase names (:eql:gh:`#1843`).
+* Rename string handling functions to use ``start`` and ``end`` in
+  their name (:eql:gh:`#1752`).
```
tailhook, line 23: Should we list functions? There are only 4 of them as far as I know.

docs/changelog/1_0_a6.rst
```diff
@@ -0,0 +1,63 @@
...
+* Fix how :ref:`FOR <ref_eql_statements_for>` statement correlates
+  values (:eql:gh:`#1776`).
+* Fix handling of collections of newly created types in SDL (:eql:gh:`#1730`).
+* Fix handling of function definitions in SDL (:eql:gh:`#1649`).
+* Fix interactions of ``SET OF`` and ``OPTIONAL`` arguments (:eql:gh:`#1640`).
+* Implementation of more of the features of the new migration syntax
+  (`RFC 1000 <migrations_>`_).
+
+CLI
+===
```
tailhook, line 43: This title while is valid is still a bit confusing.

This is for the rust-cli/repl:
* Require instance name for most `edgedb server` commands.
* Add version check to show a warning when tools should be updated
* Add support for installing server using Docker (`edgdb server install --method=docker`)
* Add `edgedb self-upgrade` command

docs/changelog/1_0_a6.rst
Dhghomon: How about calling it upper camel case? Probably not super important but I remember at least two people on Twitch who were annoyed at how Rust's structs, enums etc. were written in what it called camel case which they insisted was pascal case. Eventually the Rust docs ended up calling it upper camel case and that's what the error message gives now:

>warning: type `someStruct` should have an upper camel case name

Doing that now might avoid some pedantry later on.

tailhook: Yes, upper camel case sounds better than pascal case (nobody remember Pascal these days :) )
*(But this probably should be in a different comment)*

docs/changelog/1_0_a6.rst
1st1, line 19:
```suggestion
        EXTENDING enum<One, Many>;
```

docs/changelog/1_0_a6.rst
1st1: Agree with @tailhook; please least all functions (old, and their new equivs) in a table.

1st1 (approved):
Please fix the nits and let's merge.

docs/changelog/1_0_a6.rst
vpetrovykh: I changed the example.

vpetrovykh: I changed the example.

docs/changelog/1_0_a6.rst
vpetrovykh: I've added the list of affected functions.

docs/changelog/1_0_a6.rst
vpetrovykh: I added the changes from `edgedb-cli` (sorry I missed that repo in my first attempt).

I am not fond of CLI title, but I'm also not sure what's a good simple name for this section.

Issue #1625, opened 2020-08-02
SDL migration including function definition fails on commit with "InternalServerError: relation ... does not exist"


CodesInChaos:
- EdgeDB Version: alpha 4
- OS Version: Ubuntu 20

On an empty database run the following migration:

```
START MIGRATION TO {
module default {
	type User {
		optional property name -> str;
	}

	function get_user(name:str) -> User
	USING EdgeQL $$
		select User {
			name,
		}
		filter {
			.name = name
		}
		limit 1
	$$;
}

};
POPULATE MIGRATION;
COMMIT MIGRATION;
```

The commit fails with the following error:

> InternalServerError: relation "edgedb_22db0c59-d1e6-11ea-bd31-efd5a96c52d1.a7cb9161-d4ab-11ea-8e37-2d90af0bf1de_t" does not exist

Fogapod:
Same issue:
```edgeql
START TRANSACTION;
START MIGRATION TO {
    module default {
        type User;

        function create_user() -> SET OF User
        {
            using EdgeQL $$
            SELECT User
            $$;
        }
    };
};

POPULATE MIGRATION;
```

Temp fix: `CREATE TYPE User;` before migration.

Fogapod:
There's more to this:
```edgeql
START TRANSACTION;
CREATE TYPE User;
START MIGRATION TO {
    module default {
        type User {
            required property nickname -> str;
        }

        function create_user() -> SET OF User
        {
            using EdgeQL $$
            SELECT User
            FILTER .nickname = 'test'
            $$;
        }
    };
};

POPULATE MIGRATION;
COMMIT MIGRATION;
# ERROR: InternalServerError: column User~2.nickname does not exist
```

Fogapod:
The same error also happens when type referenced by function is updated in the same migration.
I encountered this when I renamed one of the fields in type and also created function referencing this type.

If I remove function definition from migration, run it, add function back and run again, it works fine.
Creating function before migration using DDL also fixes this.

elprans:
Yeah, this is an issue with internal SQL command ordering, I'll look into it soon.

msullivan:
The issue here is that the function references the type's inheritance view and not the table, and inheritance views are always all updated at the end of the ddl script. So even if the table is created/updated before the function, the view still isn't.

msullivan:
This is fixed

PR #1866, opened 2020-10-07
Add an xfailed migration test

msullivan:


elprans (approved):
I think this was also previously reported in #1625.

PR #1867, opened 2020-10-08
Add support for per-database configuration

elprans:
EdgeDB currently supports SESSION and SYSTEM configuration scopes.  Add
a new DATABASE scope that lives in between SESSION and SYSTEM.


edb/pgsql/metaschema.py
```diff
@@ -94,6 +111,118 @@ def __init__(self) -> None:
...
+
+class AlterCurrentDatabaseSetArray(dbops.Function):
+    """Alter a configuration parameter of the current database."""
+    text = '''
+    BEGIN
+        EXECUTE 'ALTER DATABASE ' || quote_ident(current_database())
+        || ' SET ' || quote_ident(parameter) || ' = '
+        || coalesce(
+            (SELECT
+                array_to_string(array_agg(q.v::text), ',')
```
tailhook, line 202: What's the upside of using different encodings for different types, especially unsafe ones, like this, instead of using JSON for all values?

edb/edgeql/codegen.py
```diff
@@ -1628,25 +1628,25 @@ def visit_FuncParam(self, node: qlast.FuncParam) -> None:
             self.visit(node.default)
 
     def visit_ConfigSet(self, node: qlast.ConfigSet) -> None:
-        self.write('CONFIGURE')
-        self.write(' SYSTEM' if node.system else ' SESSION')
+        self.write('CONFIGURE ')
+        self.write(node.scope)
```
tailhook, line 1632: Looks like it emits `DATABASE` rather than `CURRENT DATABASE`?

tailhook (commented):
Overall looks good.
* No docs
* No dump support yet, right?

edb/edgeql/codegen.py
elprans: Good catch.

edb/pgsql/metaschema.py
elprans: 1. PostgreSQL settings are always either strings or lists of strings.
2. There's no way to put `ALTER SYSTEM` into a function, so we cannot use JSON encoding for that operation anyway.

edb/edgeql/codegen.py
1st1: I see this comment marked as "resolved" but I don't see how it was fixed.

1st1: In general I'd suggest adding an explicit `to_edgeql()` method to the enum, like other enums have.

edb/pgsql/metaschema.py
```diff
@@ -63,6 +63,23 @@ def __init__(self, conn):
         self.db = conn
 
 
+class DBConfigTable(dbops.Table):
+    def __init__(self) -> None:
+        super().__init__(name=('edgedb', '_db_config'))
```
1st1, line 68: I'm just curious: do we need to use a leading `_` for SQL-level tables and helpers?

edb/pgsql/metaschema.py
```diff
@@ -94,6 +111,118 @@ def __init__(self) -> None:
         ),
 
 
+class AlterCurrentDatabaseSetString(dbops.Function):
+    """Alter a configuration parameter of the current database."""
```
1st1, line 115: I'd clarify with

```suggestion
    """Alter a PostgreSQL configuration parameter of the current database."""
```

for this and the following functions.

edb/server/dbview/dbview.pxd
```diff
@@ -17,6 +17,13 @@
 #
 
 
+cpdef enum SideEffects:
+
+    SchemaChanges = 1
+    DatabaseConfigChanges = 2
+    SystemConfigChanges = 4
```
1st1, line 24:
```suggestion
    SchemaChanges = 1 << 0
    DatabaseConfigChanges = 1 << 1
    SystemConfigChanges = 1 << 2
```

to make it obvious that this is a bitmask enum

edb/server/pgcon/pgcon.pyx
```diff
@@ -1089,12 +1096,29 @@ cdef class PGProto:
...
             self.buffer.finish_message()
 
-            if channel == '__edgedb_ddl__':
-                dbver = bytes.fromhex(payload)
-                if self.edgecon_ref is not None:
-                    edgecon = self.edgecon_ref()
-                    if edgecon is not None:
-                        edgecon.on_remote_ddl(dbver)
+            if channel == '__edgedb_sysevent__':
+                event_data = json.loads(payload)
```
1st1, line 1100: Maybe event_data should be an array of events?

edb/edgeql/codegen.py
1st1: nvm, I see this was addressed in a follow up commit

1st1 (approved):
I've left a few nit comments. Good job!

elprans:
Added four commits with fixes: configuration scopes are no longer mixed up inappropriately and `DESCRIBE CURRENT DATABASE CONFIG` is implemented.  I'll follow up with dump integration and docs in a separate PR.

1st1:
> Added four commits with fixes: configuration scopes are no longer mixed up inappropriately and DESCRIBE CURRENT DATABASE CONFIG is implemented. I'll follow up with dump integration and docs in a separate PR.

And also please prohibit DML in functions now that we have all the blocks for that.

1st1 (approved):

PR #1868, opened 2020-10-08
Better titles in changelog

tailhook:


elprans (approved):
@ambv Please cherry-pick this into a6

docs/changelog/1_0_a6.rst
```diff
@@ -45,16 +45,22 @@ EdgeQL
 * Implementation of more of the features of the new migration syntax
   (`RFC 1000 <migrations_>`_).
 
-CLI
-===
+
+Command-Line Tools
```
1st1, line 49:
```suggestion
Command Line Tools
```

elprans: Au contraire, "command-line" seems to be the appropriate spelling for an adjective form: https://en.wikipedia.org/wiki/Command-line_interface

1st1: Before I suggested I did a quick google search and saw that Apple uses "Command Line Tools" for their stuff. Even that linked wiki article uses the dashless version frequently, if not more frequently than the one with the dash. 🤷‍♂️

1st1: I'm actually curious what @Dhghomon thinks about this: google shows a 50/50 split in search results and lots of inconsistent uses.

1st1: I went ahead and merged as is. If Dave suggests a different spelling we can fix in a follow up commit.

Dhghomon: tl;dr: Yes, let's keep the hyphen.

Aesthetically my brain says Command Line Tools looks better and hyphens are being used less and less correctly as time goes on (because compound modifiers are so vague) but the hyphen always does make it clearer that it's an adjective and Microsoft even has it in its style guide with the hyphen:

https://docs.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/c/command-line

https://communicatingtechnically.com/2016/05/12/commandline-tool-command-line-tool-command-line-tool/

I like the second link there by someone who went through the same agonizing process of trying to decide on one.

Not that I'd go and hyphenate another company's docs unasked if they went with Command Line Tools and preferred it that way, mind you.

PR #1871, opened 2020-10-08
`buildmeta.get_version()` does not return a string

ambv:


edb/tools/gen_test_dumps.py
```diff
@@ -107,7 +110,7 @@ def die(msg):
     help="number of parallel processes to use",
 )
 def gen_test_dumps(*, jobs, tests_dir):
-    version = buildmeta.get_version()
+    version = str(buildmeta.get_version())
```
1st1, line 113: I know it's rush hour, but maybe you should unpack the tuple it returns instead of doing str replace.

elprans (approved):

edb/tools/gen_test_dumps.py
ambv: This won't change much since I will still want to translate "alpha" into "a", and so on. I agree that your suggestion is more elegant. We'll maybe save one line that way.

1st1: It's not about saving one line, it's about the code being easier to follow (and perhaps more resilient to changes in the `Version` however unlikely they are). But since this is a peripheral tool and not core, I don't care that much.

PR #1877, opened 2020-10-10
Fix an issue with repeating the same proposed DDL.

vpetrovykh:
Some tests went into infinite loops when trying to apply DDL generated
by DESCRIBE MIGRATION. This was due to DDL being proposed over and over
(as if the previous DDL didn't apply correctly). It appears that the
diff was handling missing and default values incorrectly instead.

Issue #1772

### DIFF
--- a/edb/schema/objects.py
+++ b/edb/schema/objects.py
@@ -1141,6 +1141,18 @@ class Object(s_abc.Object, s_abc.ObjectContainer, metaclass=ObjectMeta):
         explicit: bool = False,
     ) -> float:
         fname = field.name
+
+        # If a field is not inheritable (and thus cannot be affected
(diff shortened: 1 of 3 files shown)


edb/schema/objects.py
```diff
@@ -1123,6 +1123,21 @@ def compare_field_value(
             assert isinstance(result, (float, int))
             return result
 
+        # If a field is not inheritable (and thus cannot be affected
```
elprans, line 1126: Please move this check into `compare_obj_field_value` (the `if explicit` branch), for clarity.

vpetrovykh: I moved the logic to `compare_obj_field_value`.

elprans (approved):

Issue #1841, opened 2020-09-29
Implement consistent fixup of expressions when referenced schema objects are renamed


elprans:
Modules, Types, links and properties are now represented in the backend in a name-independent way, which is to say that renames of those schema objects are "free" from the standpoint of compiled expressions.  That said, the original names may still be stored in the original text of EdgeQL expressions in the schema.  

We need to implement a consistent mechanism that would "fixup" renamed references in schema expressions.  A straightforward way to implement this is to add a reference token map to `expr.Expression` and perform string replacements.

PR #1878, opened 2020-10-13
Fixup expressions when schema objects are renamed

msullivan:
We do this by adding infrastructure to track where in the AST
references to schema objects appear, and then using that to rewrite
expressions when we need to do a rename.

There are still some gaps in this, but I wanted to get something up
for review as a start:
 * Renaming types used in function arguments makes it impossible to
   delete the functions (since the argument types are encoded in
   the underlying name and I haven't implemented renaming there yet).
 * Migrations that rename a property used in an object constraint
   don't always work, since the DDL-based recompilation can interfere
   with how the migration wants to do it.
 * Lots of cases don't have tests written yet.

This is much of the way on #1841.

### DIFF
--- a/edb/edgeql/compiler/context.py
+++ b/edb/edgeql/compiler/context.py
@@ -195,6 +195,11 @@ class Environment:
     schema_refs: Set[s_obj.Object]
     """A set of all schema objects referenced by an expression."""
 
+    schema_ref_exprs: Optional[Dict[s_obj.Object, Set[qlast.Base]]]
+    """Map from all schema objects referenced to the ast referants.
+
+    This is used for rewriting expressions in the schema after a rename. """
+
     created_schema_objects: Set[s_obj.Object]
     """A set of all schema objects derived by this compilation."""
 
@@ -231,6 +236,7 @@ class Environment:
         self.view_shapes_metadata = collections.defaultdict(
             irast.ViewShapeMetadata)
         self.schema_refs = set()
+        self.schema_ref_exprs = {} if options.track_schema_ref_exprs else None
         self.created_schema_objects = set()
         self.ptr_ref_cache = PointerRefCache()
         self.type_ref_cache = {}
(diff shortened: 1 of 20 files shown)


edb/pgsql/delta.py
```diff
@@ -2867,6 +2870,9 @@ def apply(
         prop = self.scls
         schema = PropertyMetaCommand.apply(self, schema, context)
 
+        if self.metadata_only:
```
elprans, line 2875: Perhaps we can generalize this by skipping `metadata_only` instances in `Adapter.adapt()` called by `compiler.py:_process_delta`?

edb/schema/delta.py
```diff
@@ -1234,77 +1234,77 @@ def _propagate_if_expr_refs(
...
+                        dummy = None
+                    elif isinstance(ref, s_func.Function):
+                        dummy = ref.get_dummy_body(schema)
+
+                    for fn in fns:
+                        # Do the switcheraroos
+                        value = ref.get_explicit_field_value(schema, fn, None)
+                        if isinstance(value, s_expr.Expression):
+                            # Strip the "compiled" out of the expression
+                            value = s_expr.Expression(
```
elprans, line 1297: I wonder if this should be a method on `Expression` (a `compiled` counterpart).

edb/edgeql/compiler/context.py
```diff
@@ -292,13 +307,15 @@ def get_track_schema_object(  # NoQA: F811
                     scls_type=s_aliases.Alias,
                     field_name='type',
                 )
-                self.schema_refs.update(alias_objs)
+                for obj in alias_objs:
+                    self.add_schema_ref(obj, expr)  # is this right?
```
elprans, line 311: > is this right?

Yes

edb/schema/delta.py
```diff
@@ -1234,77 +1234,77 @@ def _propagate_if_expr_refs(
...
         if expr_refs:
             ref_desc = []
-            for ref, fn in expr_refs.items():
+            for ref, fns in expr_refs.items():
+                really_apply = False
+
                 from . import functions as s_func
                 from . import indexes as s_indexes
                 from . import pointers as s_pointers
+                from . import constraints as s_cnstr
```
elprans, line 1254: Downstream imports aren't ideal, we should clean this up by doing some form of dispatch at some point.

elprans (approved):
LGTM!

edb/pgsql/delta.py
msullivan: I'm not sure how to make that work; we do need to process those instances, just stop after applying the schema changes and handling children.

elprans: The idea is that you don't "adapt" `metadata_only` commands into equivalent `pgsql.Command` instances. It's fine to leave this to a followup cleanup.

1st1:
wow, good job!

PR #1882, opened 2020-10-15
Fix migrations that rename properties used in constraints

msullivan:
Work on #1841.


edb/schema/ordering.py
```diff
@@ -62,6 +62,9 @@ def linearize_delta(
     # tree is broken up into linear branches, which are then sorted
     # and reassembled back into a tree.
 
+    # A map of commands to root->command paths through the tree.
+    # Nodes are duplicated so the interior nodes of the path are
+    # distinct. (Right?)
```
elprans, line 67: Right.

elprans (approved):

PR #1884, opened 2020-10-15
Add support for chained schemas

elprans:
The new `ChainedSchema` implementation added here takes two regular
schemas and chains them together a-la `ChainMap`.  This is necessary to
separate the read-only `std` portion of schema from user-defined schema
in order to save on serialization cost of transmitting the schema over
RPC.

### DIFF
--- a/edb/edgeql/compiler/setgen.py
+++ b/edb/edgeql/compiler/setgen.py
@@ -783,7 +783,7 @@ def type_intersection_set(
     arg_type = get_set_type(source_set, ctx=ctx)
 
     result = schemactx.apply_intersection(arg_type, stype, ctx=ctx)
-    if result.stype is arg_type:
+    if result.stype == arg_type:
         return source_set
 
     poly_set = new_set(stype=result.stype, ctx=ctx)
--- a/edb/schema/objects.py
+++ b/edb/schema/objects.py
@@ -499,10 +499,6 @@ class ObjectMeta(type):
         refdicts = collections.OrderedDict()
         mydicts = {}
 
-        if '__slots__' in clsdict:
-            raise TypeError(
-                f'cannot create {name} class: __slots__ are not supported')
-
         if name in mcls._all_types:
             raise TypeError(
                 f'duplicate name for schema class: {name}, already defined'
@@ -522,6 +518,7 @@ class ObjectMeta(type):
 
             v.name = k
             myfields[k] = v
+            del clsdict[k]
 
             if v.is_schema_field:
                 getter_name = f'get_{v.name}'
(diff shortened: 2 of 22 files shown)


edb/edgeql/compiler/setgen.py
```diff
@@ -783,7 +783,7 @@ def type_intersection_set(
     arg_type = get_set_type(source_set, ctx=ctx)
 
     result = schemactx.apply_intersection(arg_type, stype, ctx=ctx)
-    if result.stype is arg_type:
+    if result.stype == arg_type:
```
1st1, line 786: Ouch. This is a tricky thing to debug. Is it possible to tell mypy to prohibit using `is` on certain types?

edb/schema/objects.py
```diff
@@ -817,8 +820,8 @@ def get_verbosename(
         dname = self.get_displayname(schema)
         return f"{clsname} '{dname}'"
 
-    def __init__(self, *, _private_init: bool) -> None:
-        pass
+    def __init__(self, *, _private_init: uuid.UUID) -> None:
```
1st1, line 823: I'd call it "private_id"

```suggestion
    def __init__(self, *, _private_id: uuid.UUID) -> None:
```

edb/schema/schema.py
```diff
@@ -1004,19 +1284,401 @@ def __init__(
...
+        self,
+        obj_id: uuid.UUID,
+        default: Union[so.Object_T, so.NoDefaultT, None] = so.NoDefault,
+        *,
+        type: Optional[Type[so.Object_T]] = None,
+    ) -> Optional[so.Object_T]:
+        try:
+            return self._top_schema.get_by_id(
+                obj_id, type=type)
+        except errors.InvalidReferenceError:
```
1st1, line 1535: Maybe we optimize this one too by checking if the `_top_schema` has this id? Exceptions are expensive.

edb/schema/schema.py
```diff
@@ -1004,19 +1284,401 @@ def __init__(
...
+
+    def get_global(  # NoQA: F811
+        self,
+        objtype: Type[so.Object_T],
+        name: str,
+        default: Union[so.Object_T, so.NoDefaultT, None] = so.NoDefault,
+    ) -> Optional[so.Object_T]:
+        try:
+            return self._top_schema.get_global(objtype, name)
+        except errors.InvalidReferenceError:
```
1st1, line 1566: Ditto, optimize this one too? Checking `_globalname_to_id` on `top_schema` should do it

1st1 (approved):

edb/edgeql/compiler/setgen.py
elprans: mypy has to grow a few extra hooks and then we'll be able to do this with a plugin.

1st1: Summoning @msullivan

PR #1885, opened 2020-10-16
Upgrade mypy to 0.790

elprans:
mypy-0.790 has a critical dmypy infinite loop fix and a bunch of other
small improvements.


msullivan (approved):
Wooo finally

PR #1886, opened 2020-10-16
Finish supporting renames of objects used in expressions

### LINKED ISSUE #1841: Implement consistent fixup of expressions when referenced schema objects are renamed

msullivan:
- Generalize the scheme for testing these things
- Rename indexes when the expression changes
- Support aliases
- Fix some issues with computables

I think we've caught everything now.

Fixes #1841.

### DIFF
--- a/edb/pgsql/delta.py
+++ b/edb/pgsql/delta.py
@@ -1723,6 +1723,7 @@ class RenameIndex(IndexCommand, RenameObject, adapts=s_indexes.RenameIndex):
 
         return schema
 
+
 class AlterIndexOwned(
     IndexCommand,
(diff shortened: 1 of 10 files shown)


elprans (approved):
Nice work!

PR #1893, opened 2020-10-17
Change how inheriting cardinality is determined.

vpetrovykh:
Cardinality cannot be changed once it is set, this affects the
logic of checking whether it is inherited or not, since it may
be inherited and yet also explicitly specified.

Issue #1772

### DIFF
--- a/edb/schema/inheriting.py
+++ b/edb/schema/inheriting.py
@@ -156,10 +156,13 @@ class InheritingObjectCommand(sd.ObjectCommand[so.InheritingObjectT]):
             else:
                 ours = None
 
-            inherited = result is not None and ours is None
+            inherited = self.is_field_inherited(
+                schema, field_name, bases, result, ours)
+
             inherited_fields_update[field_name] = inherited
 
-            if (result is not None or ours is not None) and result != ours:
+            if (result is not None or ours is not None) and (
+                    result != ours or inherited):
                 schema = self.inherit_field(
                     schema,
                     context,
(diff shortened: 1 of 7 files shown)


edb/schema/inheriting.py
```diff
@@ -156,10 +156,24 @@ def inherit_fields(
             else:
                 ours = None
 
-            inherited = result is not None and ours is None
+            inherited = False
+            if field_name == 'cardinality':
```
elprans, line 160: I'm not sure I like this approach of hardcoding logic for 'cardinality' specifically.  Perhaps there should be a `Field` attribute instead: `strict_inheritance` or something like that.

vpetrovykh: Well, `cardinality` is already special. We have a cardinality-specific merge function (separate from, say, other special merges for  `readonly` and `required`). I'm not sure we gain anything here by adding a flag, but a generic flag will obscure what is actually being handled, I think.

vpetrovykh: The reason why I bring up the cardinality-specific merge function is that it exists because cardinality has slightly different rules when it comes to inheritance, where you basically can't omit it like other things.

elprans: `cardinality` is a field present only on `Pointer` objects.  It is weird to open-code for it in a generic piece of code in `inheriting.py`.  So, please either overload this in `pointers.py` or make a field attribute.

vpetrovykh: Ah, OK, point taken about `inheriting.py` vs `pointers.py`. I'll overload `pointers.py` then.

vpetrovykh:
I've added another commit related to handling cardinality to this PR.

edb/schema/inheriting.py
vpetrovykh: Inheritance has been refactored and overloaded in `pointers.py` to accommodate cardinality.

elprans (approved):

PR #1900, opened 2020-10-20
Stricter typing for schema.get

msullivan:
Get rid of the basically unconstrainted output type variable in the
type of schema.get(). Explicitly return so.Object when no `type`
argument is passed.

My motivation for this is that I had run into a few cases where the
output type was being inferred as `None` and so code wasn't getting
typechecked. And when I spot-checked some other call sites, the first
two I picked had the same problem. When I fixed the typing, though,
this didn't catch /that/ many places that had been missing
type-checking. I had to fix a handful of errors, and running with
--warn-unreachable now reports 144 errors instead of 150.

So maybe this isn't worth landing? It does give better type checking,
just not as impactfully as I had expected.

### DIFF
--- a/edb/edgeql/compiler/context.py
+++ b/edb/edgeql/compiler/context.py
@@ -289,9 +289,10 @@ class Environment:
         label: Optional[str] = None,
         condition: Optional[Callable[[s_obj.Object], bool]] = None,
     ) -> Optional[s_obj.Object]:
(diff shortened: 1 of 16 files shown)


elprans:
Looks like a rebase is needed

elprans (approved):
The tightening of `schema.get()` calls themselves is a worthy improvement.

PR #1903, opened 2020-10-21
Prohibit functions from sharing shortname with types.

### LINKED ISSUE #1465: Prohibit functions from sharing shortname with types
The following scenario should be prohibited:

```
type foo { ... }
create function foo { ... }
```

See also https://github.com/edgedb/edgedb/issues/1439 for additional context.

vpetrovykh:
Functions can no longer share their name with other schema objects. The
reasoning is that we may want to implement traits as some kind of
specialized functions and name clashes with types would cause problems.

Fixes #1465

### DIFF
--- a/edb/schema/delta.py
+++ b/edb/schema/delta.py
@@ -2118,6 +2118,16 @@ class CreateObject(ObjectCommand[so.Object_T], Generic[so.Object_T]):
 
         props = self.get_resolved_attributes(schema, context)
         metaclass = self.get_schema_metaclass()
+
+        # Check if functions by this name exist
+        fn = props.get('name')
+        if fn is not None:
+            funcs = schema.get_functions(fn, tuple())
+            if funcs:
+                raise errors.SchemaError(
+                    f'{funcs[0].get_verbosename(schema)} is already present '
+                    f'in the schema {schema!r}')
+
         schema, self.scls = metaclass.create_in_schema(schema, **props)
 
         if not props.get('id'):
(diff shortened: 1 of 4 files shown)


edb/schema/delta.py
```diff
@@ -2132,6 +2132,16 @@ def _create_begin(
...
         props = self.get_resolved_attributes(schema, context)
         metaclass = self.get_schema_metaclass()
+
+        # Check if functions by this name exist
+        fn = props.get('name')
+        if fn is not None:
+            funcs = schema.get_functions(fn, tuple())
+            if funcs:
+                raise errors.SchemaError(
+                    f'{type(funcs[0]).__name__} {fn!r} is already present '
```
elprans, line 2142: Use `get_verbosename` here.

edb/schema/functions.py
```diff
@@ -1242,14 +1242,22 @@ def _create_begin(
...
                 f'is already defined',
                 context=self.source_context)
 
+        # Check if other schema objects with the same name (ignoring
+        # signature, of course) exist.
+        if other := schema.get(
+                other_name := sn.SchemaName(shortname, fullname.module),
+                None):
+            raise errors.SchemaError(
+                f'{type(other).__name__} {other_name!r} is already present '
```
elprans, line 1258: Ditto.

edb/schema/delta.py
vpetrovykh: Done

edb/schema/functions.py
vpetrovykh: Done

elprans (approved):

PR #1909, opened 2020-10-23
Fix renaming and deleting abstract constraints

msullivan:
The parameter renaming logic is shared with functions (for now only
used when renaming due to a type change).

Work on #1772.

### DIFF
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -4270,6 +4270,29 @@ class TestGetMigration(tb.BaseSchemaLoadTest):
             }
         """])
 
+    def test_schema_migrations_equivalence_rename_abs_constraint_01(self):
+        # This drops and recreates the constraint; we'd prefer if that
+        # didn't happen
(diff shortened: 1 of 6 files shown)


tests/test_schema.py
```diff
@@ -4162,6 +4162,29 @@ def test_schema_migrations_equivalence_rename_refs_06(self):
             alias Alias3 := Remark { command := .remark ++ "!" };
         """])
 
+    def test_schema_migrations_equivalence_rename_abs_constraint_01(self):
```
elprans, line 4165: Please add an (xfailing?) test using the `assert_describe_migration` framework (see `test_edgeql_data_migraiton`).

msullivan: Looks like we already have the xfailed test we want `test_edgeql_migration_describe_constraint_01`, so I just updated its description

msullivan:
The parameter renaming logic is shared with functions (for now only
used when renaming due to a type change).

Work on #1772.

PR #1911, opened 2020-10-23
Improve test parallelizability

elprans:
We have a bunch of test suites currently that aren't parallelized. Reason number one is absense of proper error recovery handling, which is easily fixable by proper application of `assertRaisesRegexTx`. Reason number two is DDL concurrency and possible deadlocking, which is fixed by making copies of test databases for each runner worker.  Altogether, this meaningfully improves test run times on machines with high core count, but is also useful to speed up selective test running if the selected test suite happens to be non-parallelizable currently.


edb/testbase/server.py
```diff
@@ -643,8 +644,17 @@ class DatabaseTestCase(ClusterTestCase, ConnectedTestCaseMixin):
...
-    ISOLATED_METHODS = True
+    # or affect non-transactional state, in which case
+    # TRANSACTION_ISOLATION must be set to False
+    TRANSACTION_ISOLATION = True
+
+    # By default, tests from the same testsuite may be ran
+    # in parallel in several test worker processes.  However,
+    # certain cases might exhibit pathological locking behavior,
+    # or are parallel-unsafe altogether, in which case
+    # PARALLELISM_GRANULARITY must be set to 'database' or 'system'.
```
1st1, line 657: This comment doesn't actually explain what "database" or "system" do.

tests/test_cli.py
```diff
@@ -1,72 +0,0 @@
...
-# limitations under the License.
-#
-
-
-import edgedb
-
-from edb.testbase import server as tb
-
-
-class TestCLI(tb.ConnectedTestCase, tb.OldCLITestCaseMixin):
```
1st1, line 25: Is this removal intentional? The commit message doesn't mention it.

edb/edgeql/parser/grammar/ddl.py
```diff
@@ -424,12 +424,22 @@ def reduce_AlterFinal(self, *kids):
 # CREATE DATABASE
 #
 class CreateDatabaseStmt(Nonterm):
-    def reduce_CREATE_DATABASE_AnyNodeName_OptCreateRoleCommandsBlock(
-        self,
-        *kids,
-    ):
+    def reduce_CREATE_DATABASE_regular(self, *kids):
+        """%reduce CREATE DATABASE AnyNodeName OptCreateDatabaseCommandsBlock
```
1st1, line 428: And this commit actually adds a new feature to EdgeQL that needs to be documented and perhaps deserves a commit message.

elprans: It's a hidden feature, only for `testmode`.  There is no intention to expose it for now.

tests/test_cli.py
elprans: The old CLI is slated for removal and I couldn't be bothered to spend time to fix something that's going to get removed very soon anyway.

edb/testbase/server.py
elprans: Added an explanation.

edb/edgeql/parser/grammar/ddl.py
1st1: Ah, I see you raise an error in `_cmd_tree_from_ast`. All's good.

elprans:
I hid the parallel-mode behind a debug flag, so this should be safe to merge.

1st1 (approved):

PR #1913, opened 2020-10-26
Stop storing instances of schema objects in the schema container

elprans:
Currently, the `Schema` object stores schema object dicts but _also_ the
Python object instances of the corresponding schema objects and schema
object collections.  This is detrimental for schema serialization as it
adds overhead when unpickling or introspecting the schema.  Fix this by
constructing the necessary objects on the fly instead.  The extra
runtime overhead is mitigated by a reasonably sized LRU cache.

This also includes some schema field getter optimizations that balance
the runtime performance hit from the above so the effect of this patch
on compilation performance is neutral.

### DIFF
--- a/edb/schema/abc.py
+++ b/edb/schema/abc.py
@@ -18,25 +18,33 @@
 
 
 from __future__ import annotations
-
 import typing
-
-
-if typing.TYPE_CHECKING:
-    ObjectContainer_T = typing.TypeVar(
-        'ObjectContainer_T', bound='ObjectContainer'
-    )
 
 
 class Schema:
     pass
 
 
+class Reducible:
+    """An interface implemented by all non-builtin objects stored in schema."""
+
+    def schema_reduce(self) -> typing.Any:
+        """Return a primitive representation of the object.
+
+        The return value must consist of primitive Python objects.
+        """
+        raise NotImplementedError
+
+    @classmethod
+    def schema_restore(
+        cls,
+        data: typing.Any,
+    ) -> Reducible:
+        """Restore object from data returned by *schema_reduce*."""
+        raise NotImplementedError
+
+
 class Object:
-    pass
-
-
-class ObjectContainer:
     pass
 
 
--- a/edb/schema/objects.py
+++ b/edb/schema/objects.py
@@ -24,6 +24,7 @@ import builtins
 import collections
 import collections.abc
 import enum
+import functools
 import uuid
 
 from edb import errors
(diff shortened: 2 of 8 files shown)


edb/schema/objects.py
```diff
@@ -786,6 +852,27 @@ class Object(s_abc.Object, s_abc.ObjectContainer, metaclass=ObjectMeta):
 
     _fields: Dict[str, SchemaField[Any]]
 
+    def schema_reduce(self) -> Tuple[str, uuid.UUID]:
+        return type(self).__name__, self.id
+
+    @classmethod
+    @functools.lru_cache(maxsize=10240)
+    # mypy hates lru_cache
+    def schema_restore(  # type: ignore
```
1st1, line 861: I'm curious, will mypy ignore this function here but still keep validating that it's being called correctly from other places?

edb/schema/abc.py
```diff
@@ -18,25 +18,27 @@
...
-
 class Schema:
     pass
 
 
-class Object:
-    pass
+class Reducible:
+
+    def schema_reduce(self) -> typing.Any:
```
1st1, line 30: Please add detailed docstrings to both.

edb/schema/objects.py
```diff
@@ -505,30 +518,79 @@ def __new__(
...
                     # The getter was defined explicitly, move on.
                     continue
-                clsdict[getter_name] = (
-                    lambda self, schema, *, _fn=v.name:
-                        self._get_schema_field_value(schema, _fn)
-                )
+
+                ftype = field.type
+                if issubclass(ftype, s_abc.Reducible):
+                    def reducible_getter(
```
1st1, line 544: I'd add a comment that you care about perf so these are closures and not toplevel funcs with functools.partial wrapping them.

1st1 (approved):

edb/schema/objects.py
elprans: Mypy complains about the signature not matching that of the parent, because `@lru_cache` adds some properties to the returned callable (cache information).  This has no effect on the actual function signature.

msullivan: Though using @lru_cache does probably lose the attribute types as far as mypy is concerned

edb/schema/schema.py
```diff
@@ -818,21 +906,49 @@ def add(
 
         return self._replace(**updates)  # type: ignore
 
+    def add(
+        self,
+        id: uuid.UUID,
+        sclass: Type[so.Object],
+        data: Mapping[str, Any],
+    ) -> FlatSchema:
+        reducible_fields = sclass.get_reducible_fields()
```
msullivan, line 915: What's going on here?

msullivan (approved):
This seems pretty reasonable.

Are there any brief comments that could be added somewhere to explain the new scheme a little?

edb/schema/schema.py
elprans: Basically we don't want non-primitive Python object instances in the schema anymore.  And so, if you have a field that has a type `schema.objects.Object` or `objects.ObjectCollection`, etc, we call `schema_reduce` on it to get a primitive tuple.  `schema_restore` reverses the process and is called when you call the corresponding getter.  Specifically, objects are reduced to a `(classname, id)` tuple, and object collections are reduced to a `(classname, typeargs, tuple<id>, tuple<attr, val>)`.

PR #1926, opened 2020-10-29
Add a `DROP EXPRESSION` sub-command to DDL grammar.

vpetrovykh:
We need a command to express the change from a computable to
non-computable links and properties.

Issue #1772


msullivan (approved):
LGTM.

Is this just the grammar support, and it doesn't work in the backend yet? (Otherwise it's probably worth adding a functionality test)

vpetrovykh:
Yeah, this is just grammar support. Backend support would be the next step.

PR #1931, opened 2020-10-29
Use tuples instead of maps to hold object data in schema

elprans:
Currently, we use instances of `immutables.Map` to hold schema object
"dicts" in the schema container.  This is somewhat wasteful, because we
store map keys repeatedly for every object, but the type layout is known
and static.  This can be fixed by storing object data as tuples instead,
which poses a different tradeoff: most fields in objects are `None`, but
that is still more efficient than duplicating sparse map keys.  More
importantly, however, unpickling `immutables.Map` is significantly
slower than unpickling raw tuples.  For example on the `dump01` test
schema I get the following figures:

Tuples:

- pickle size: 353,560 bytes
- unpickle time: 7,643,447 nanoseconds

immutables.Map:

- pickle size: 384,861 bytes
- unpickle time: 97,480,621 nanoseconds

### DIFF
--- a/edb/schema/objects.py
+++ b/edb/schema/objects.py
@@ -23,6 +23,7 @@ from typing import *
 import builtins
 import collections
 import collections.abc
+import copy
 import enum
 import functools
 import uuid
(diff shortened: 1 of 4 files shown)


1st1 (approved):
The code looks OK, but it's super confusing that the PR is against the still not landed lazy-schema branch. Let's please just merge this all.

PR #1932, opened 2020-10-29
Use a table instead of functions for system data

elprans:
It turns out that calling an immutable constant-returning function is a
lot slower than selecting from a table if the size of the returned
datum is considerable, an order of magnitude slower, in fact.  So,
switch system data storage to use a regular table, this shaves off a few
milliseconds from the cold compiler startup time.

### DIFF
--- a/edb/pgsql/metaschema.py
+++ b/edb/pgsql/metaschema.py
@@ -2159,7 +2159,11 @@ class SysConfigFunction(dbops.Function):
             (s.value->>'typemod') AS typemod,
             (s.value->>'backend_setting') AS backend_setting
         FROM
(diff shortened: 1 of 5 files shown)


1st1:
> this shaves off about ~50ms from the cold compiler startup time.

wow

1st1 (approved):

elprans:
>> this shaves off about ~50ms from the cold compiler startup time.

> wow

Turns out I wasn't measuring this correctly.  The difference is still large: 500microseconds vs 3000microseconds for the introspection query, so the savings are ~2.5ms, not 50.

PR #1933, opened 2020-10-30
Implement `DROP EXPRESSION`

vpetrovykh:
Implement `DROP EXPRESSION` command to make property or link no longer
computable.

Issue #1772

### DIFF
--- a/edb/schema/delta.py
+++ b/edb/schema/delta.py
@@ -2854,6 +2854,7 @@ class AlterSpecialObjectProperty(Command):
         field = parent_cls.get_field(propname)
 
         new_value: Any = astnode.value
+        old_value: Any = None
 
         if field.type is s_expr.Expression:
             if parent_cls.has_field(f'orig_{field.name}'):
@@ -2861,17 +2862,20 @@ class AlterSpecialObjectProperty(Command):
                     schema, parent_op.qlast, field.name)
             else:
                 orig_text = None
-            new_value = s_expr.Expression.from_ast(
-                astnode.value,
-                schema,
-                context.modaliases,
-                context.localnames,
-                orig_text=orig_text,
-            )
+
+            if astnode.value is not None:
+                new_value = s_expr.Expression.from_ast(
+                    astnode.value,
+                    schema,
+                    context.modaliases,
+                    context.localnames,
+                    orig_text=orig_text,
+                )
 
         return AlterObjectProperty(
             property=astnode.name,
             new_value=new_value,
+            old_value=old_value,
             source_context=astnode.context,
         )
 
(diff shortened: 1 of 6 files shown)


edb/schema/delta.py
```diff
@@ -2850,24 +2850,39 @@ def _cmd_tree_from_ast(
...
+                )
+            else:
+
+                # SetSpecialField 'expr' to None is `DROP EXPRESSION`,
+                # so we're going to need the old value in order to see
+                # that a change happened.
+                from . import pointers as s_pointers
+
+                pointer = cast(
+                    s_pointers.Pointer, schema.get(parent_op.classname))
```
elprans, line 2878: You cannot rely on `schema` at the `cmd_tree_from_ast` stage.  This has to go into the delta application phase, which, in this case would be `canonicalize_attributes` in `pointers.py`, and you'll likely also need to set up some sentinel in `_process_create_or_alter_ast` there too.

vpetrovykh: I'm not sure I follow. The pointer in question *has* to already exist in the schema because we are in the `DROP EXPRESSION` branch. So there must be a pointer already defined and we can use it to get the `old_value`. What am I missing?

elprans: > What am I missing?

The DDL script is not guaranteed to be executed statement-by-statement. Consider the following script:

```
ALTER TYPE Foo CREATE LINK foo := (...);
ALTER TYPE Foo ALTER LINK foo DROP EXPRESSION;
```

Also, whenever you find yourself importing a specific module into a general one, it's likely that you're not doing it in the correct spot.

edb/schema/pointers.py
```diff
@@ -1001,6 +1006,67 @@ def _deparse_name(
...
+            cmd._process_create_or_alter_ast(schema, astnode, context)
+        else:
+            expr_cmd = qlast.get_ddl_field_command(astnode, 'expr')
+            if expr_cmd is not None:
+                expr = expr_cmd.value
+                if expr is None:
+                    # `DROP EXPRESSION` detected
+                    aop = sd.AlterObjectProperty(
+                        property='expr',
+                        new_value=s_expr.Expression(text='DROP EXPRESSION'),
```
elprans, line 1031: What's the reason for the `DROP EXPRESSION` hack? Seems like setting it to `None` and then using `get_attribute_set_cmd()` in `canonicalize_attributes` would be cleaner.

vpetrovykh: I guess that works (I just realized that's almost what `get_local_attribute_value` is doing behind the scenes).

elprans (approved):

PR #1934, opened 2020-10-30
Refactor EdgeQL tokenization to always include the source string

elprans:
When the switch to the Rusty EdgeQL lexer was made, we lost the ability
to track the original query source text, which gets in the way of
server-side error reporting.  This refactor introduces `edgeql.Source`
as the canonical way of representing an EdgeQL query source, including
the normalized version, and switches everything to use it.


edb/edgeql/tokenizer.py
```diff
@@ -0,0 +1,123 @@
...
+        return self._tokens
+
+    def first_extra(self) -> Optional[int]:
+        return None
+
+    def extra_count(self) -> int:
+        return 0
+
+    def extra_blob(self) -> bytes:
+        return b''
```
1st1, line 55: Please add docstrings for the class and all methods.

edb/edgeql/tokenizer.py
```diff
@@ -0,0 +1,123 @@
...
+
+    @classmethod
+    def from_string(cls, text: str) -> Source:
+        return cls(text=text, tokens=tokenize(text))
+
+
+class NormalizedSource(Source):
+
```
1st1, line 63: Please add a docstring.

edb/edgeql/tokenizer.py
```diff
@@ -0,0 +1,123 @@
...
+        return _tokenize(eql)
+    except TokenizerError as e:
+        message, position = e.args
+        hint = _derive_hint(eql, message, position)
+        raise errors.EdgeQLSyntaxError(
+            message, position=position, hint=hint) from e
+
+
+def normalize(eql: str) -> Entry:
+    try:
```
1st1, line 106: Please add a docstring.

1st1 (approved):

msullivan (approved):

PR #1937, opened 2020-10-30
Use an inferred volatility for functions

### LINKED ISSUE #1888: Infer callable volaitlity instead of defaulting to VOLATILE
We are currently defaulting to `VOLATILE` for functions instead of inferring volatility from the function body, which is a problem since we're now placing more restrictions on where volatiles can appear in an expression.

msullivan:
We do this by adding a new inferred_volatility field to
VolatilitySubject and falling back to that (via a new pseudo-field
"effective volatility") when volatility isn't specified.

I did it that way instead of just populating volatility with the
inferred value because that caused us to include 'SET volatility :=
...' in our DESCRIBE output when volatility was inferred.

We then expose `inferred_volatility` and `effective_volatility` via
introspection. The change in the behavior/meaning of the volatility
field here is something of a compatibility break.

Not sure if there is a better way to structure this all that would
preserve the meaning of the `volatility` field for introspection?

Fixes #1888.

### DIFF
--- a/edb/schema/functions.py
+++ b/edb/schema/functions.py
@@ -26,6 +26,8 @@ from typing import *
 
 from edb import errors
 
+from edb.common import verutils
+
 from edb.edgeql import ast as qlast
 from edb.edgeql import compiler as qlcompiler
 from edb.edgeql import qltypes as ft
@@ -975,7 +977,7 @@ class AlterCallableObject(
     sd.AlterObject[CallableObjectT],
 ):
 
-    def get_ast(
+    def _get_ast(
         self,
         schema: s_schema.Schema,
         context: sd.CommandContext,
@@ -1274,6 +1276,34 @@ class FunctionCommand(
             )
         return val
 
+    def canonicalize_attributes(
+        self,
+        schema: s_schema.Schema,
+        context: sd.CommandContext,
+    ) -> s_schema.Schema:
+        schema = super().canonicalize_attributes(schema, context)
+        # When volatility is altered, we need to force a
+        # reconsideration of nativecode if it exists in order to check
+        # it against the new volatility or compute the volatility on a
+        # RESET.  This is kind of unfortunate.
+        if (
+            isinstance(self, sd.AlterObject)
+            and self.has_attribute_value('volatility')
+            and not self.has_attribute_value('nativecode')
+            and (nativecode := self.scls.get_nativecode(schema)) is not None
+        ):
+            self.set_attribute_value(
+                'nativecode',
+                expr.Expression.not_compiled(nativecode)
+            )
+
+        # Resolving 'nativecode' has side effects on has_dml and
+        # volatility, so force it to happen as part of
+        # canonicalization of attributes.
+        super().get_resolved_attribute_value(
+            'nativecode', schema=schema, context=context)
+        return schema
+
     def compile_function(
         self,
         schema: s_schema.Schema,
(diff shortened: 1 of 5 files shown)


1st1:
> I did it that way instead of just populating volatility with the
inferred value because that caused us to include 'SET volatility :=
...' in our DESCRIBE output when volatility was inferred.

And why is that bad?

> We then expose inferred_volatility and effective_volatility via
introspection. The change in the behavior/meaning of the volatility
field here is something of a compatibility break.

IMO this is confusing. What volatility will the compiler use if the specified volatility is different from the inferred one?

msullivan:
> > I did it that way instead of just populating volatility with the
> > inferred value because that caused us to include 'SET volatility :=
> > ...' in our DESCRIBE output when volatility was inferred.
> 
> And why is that bad?
> 

Noisy, ugly, causes a lot of issues with migrations tests.

> > We then expose inferred_volatility and effective_volatility via
> > introspection. The change in the behavior/meaning of the volatility
> > field here is something of a compatibility break.
> 
> IMO this is confusing. What volatility will the compiler use if the specified volatility is different from the inferred one?

We use the specified volatility, but specifying a volatility less strict than the inferred volatility is an error. This means you can declare a function to be stricter than it actually is, but not less strict.

1st1:
>> And why is that bad?
>
> Noisy, ugly

Well, "ugly" isn't an objective metric, and as for noisy, I'm not sure I agree either. I, personally, don't mind seeing volatility when I inspect functions in repl.

> causes a lot of issues with migrations tests.

Maybe we can tweak tests somehow?

> We use the specified volatility

Right.

> but specifying a volatility less strict than the inferred volatility is an error.

This is great.

That said, I still think `inferred_{name}` properties a bit confusing to the user. Can we make it non-introspectable, so that it's just an implementation detail? That way we wouldn't need to document it or care about it much.

elprans:
In this case, I think, `inferred_volatility` is the _default_ value for the `volatility` field.  In most other cases the default field value is hardcoded, but here we'd compute it instead.  In practice this means that the definition will be as follows:

```
CREATE ABSTRACT TYPE schema::VolatilitySubject EXTENDING schema::Object {
    CREATE PROPERTY inferred_volatility -> schema::Volatility;    
    CREATE PROPERTY volatility -> schema::Volatility {
        SET default := .inferred_volatility
    };
};
```

Like with other fields, we don't show its value in `DESCRIBE` (unless verbose), if the default value was used.

edb/lib/schema.edgeql
```diff
@@ -167,11 +167,11 @@ CREATE ABSTRACT TYPE schema::CallableObject
...
 
 CREATE ABSTRACT TYPE schema::VolatilitySubject EXTENDING schema::Object {
-    CREATE PROPERTY volatility -> schema::Volatility {
-        # NOTE: this default indicates the default value in the python
-        # implementation, but is not itself a source of truth
-        SET default := 'Volatile';
-    };
+    CREATE PROPERTY volatility -> schema::Volatility;
+    CREATE PROPERTY inferred_volatility -> schema::Volatility;
+    CREATE REQUIRED PROPERTY effective_volatility := (
```
1st1, line 172: I don't think this is needed given that you already have the `get_effective_volatility()` helper function.

edb/schema/functions.py
```diff
@@ -618,10 +620,30 @@ def get_ast(self, schema: s_schema.Schema) -> List[qlast.FuncParam]:
...
+    # when the volatility was inferred.
     volatility = so.SchemaField(
-        ft.Volatility, default=ft.Volatility.Volatile,
+        ft.Volatility, default=None,
         compcoef=0.4, coerce=True, allow_ddl_set=True)
 
+    inferred_volatility = so.SchemaField(
+        ft.Volatility, default=None)
+
+    def get_effective_volatility(
```
elprans, line 639: This isn't properly documented, but you are actually allowed to overload the normal getter for a field value, and so this should be a `def get_volatility()` instead.  See `Constraint.get_subject()` for an example on how to get to the raw field value in this case.  Oh, and the fallback to `Volatile` is likely not needed anymore: the volatility should always either be inferred, or specified explicitly (in the case of SQL-backed functions in std).

msullivan:
> In this case, I think, `inferred_volatility` is the _default_ value for the `volatility` field. In most other cases the default field value is hardcoded, but here we'd compute it instead. In practice this means that the definition will be as follows:
> 
> ```
> CREATE ABSTRACT TYPE schema::VolatilitySubject EXTENDING schema::Object {
>     CREATE PROPERTY inferred_volatility -> schema::Volatility;    
>     CREATE PROPERTY volatility -> schema::Volatility {
>         SET default := .inferred_volatility
>     };
> };
> ```
> 
> Like with other fields, we don't show its value in `DESCRIBE` (unless verbose), if the default value was used.

We don't appear to allow partial paths like that in defaults.

edb/lib/schema.edgeql
msullivan: We don't populate these tables with that sort of helper function value, though

1st1: That was my point -- let's not add properties to `schema::` unless we need to. And it seems like `effective_volatility` can be computed in Python whenever we need it, so there's no strong need to have it in `schema::`.

msullivan:
I think we ought to export the "real" volatility to `schema::` somehow,
though, and I don't think we've yet figured out a good way to do it as the
`volatility` property.

edb/lib/schema.edgeql
msullivan: It's not really required for anything to work, but the idea for adding it was so that the correct value is available via reflection easily

msullivan:
I've updated this with a version that makes the stored schema field for specified volatility named `specified_volatility` while keeping the name as `volatility` as far as DDL is concerned. This lets us expose the real volatility as `volatility` in the reflection schema (though the other parts are also exposed so that we can compute it).

I also spent some time working on an approach that would always store the real volatility in `volatility` and that would use another flag to indicate whether it was computed or not. This was most of the way towards working (some migration issues still needed to be fixed) but was a lot more hairy and filled with special casing.

msullivan:
(There are still some problems with this and ALTER changing volatility, but I'm working on that)

msullivan:
OK, this is in good shape now @elprans

elprans (approved):
Looks much better!

PR #1949, opened 2020-11-04
Make migration constraint renames actually generate renames

msullivan:
Previously we would drop the constraint and create a new one,
which is bad because it requires checking the constraint for
the whole table.

Fixing this required:
 * Making parameter comparisions compare based on just the
   real param name, not the fully encoding one that includes
   the constraint/function name.
 * Having concrete constraint comparisons take into account
   in-flight renames of base constraints when comparing names.

(Should we try to move to a world where all these sorts of derived
objects don't have global derived names? They cause a lot of
grief. But probably the change would also.)

Work on #1772

### DIFF
--- a/edb/schema/ddl.py
+++ b/edb/schema/ddl.py
@@ -230,16 +230,6 @@ def delta_schemas(
 
                 result.add(create)
 
-    objects = sd.DeltaRoot(canonical=True)
-
-    # FIXME: We (mostly accidentally) rely on the ordering of
-    # schemaclasses, which is totally implicit from our import
-    # structures but also deterministic.
-    # This is because in some cases involving renames, we need to
-    # process the rename before we process consumers of it.
-    # In particular, I think ObjectType needs to be processed pretty late.
-    #
-    # This should be explicit and not implicit.
     schemaclasses = [
         schemacls
         for schemacls in so.ObjectMeta.get_schema_metaclasses()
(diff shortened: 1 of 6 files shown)


edb/schema/ddl.py
```diff
@@ -230,6 +230,14 @@ def _filter(schema: s_schema.Schema, obj: so.Object) -> bool:
...
     objects = sd.DeltaRoot(canonical=True)
 
+    # FIXME: We (mostly accidentally) rely on the ordering of
+    # schemaclasses, which is totally implicit from our import
+    # structures but also deterministic.
+    # This is because in some cases involving renames, we need to
+    # process the rename before we process consumers of it.
+    # In particular, I think ObjectType needs to be processed pretty late.
+    #
+    # This should be explicit and not implicit.
```
elprans, line 240: We used to have an explicit ordering, but it was removed in favor of topological sorting of the diff in `ordering.py`, so it would be best to fix that instead of resurrecting the explicit order.

msullivan: The trickiness here isn't (at least for the case here) that we don't sort them in the right order, but that the rename means we won't generate any diff at all for other parts. Probably that can be merged into the ordering process; it will require detecting now-unnecessary name changes and dropping them.

In general though, the diff process itself is sensitive to ordering, since whether a rename of some other object has been processed first will affect the comparison; if something else changed too, that ordering might decide whether we consider it an ALTER or not, right?

elprans: Yes, that's right.  We already do inheritance topo sort in the diff routine, but will likely need to include other kinds of dependencies.  My point isn't that we shouldn't sort at all, but that we should sort by dependencies explicitly rather than by some predefined order.

msullivan: I took a different approach, which is to re-compute the diff until it stabilizes. This should let us be agnostic to ordering and improves the DDL we generate for some existing cases.

elprans (approved):

PR #1956, opened 2020-11-05
Enable DDL for altering enums.

vpetrovykh:
Enum types can now be altered, but the alter can only add new labels,
not change the existing ones in any way. This ensures that the existing
data is always valid.

### DIFF
--- a/edb/schema/scalars.py
+++ b/edb/schema/scalars.py
@@ -324,6 +324,11 @@ class CreateScalarType(
 
                 shell = bases[0]
                 assert isinstance(shell, AnonymousEnumTypeShell)
+                if len(set(shell.elements)) != len(shell.elements):
+                    raise errors.SchemaDefinitionError(
+                        f'enums cannot contain duplicate values',
+                        context=astnode.bases[0].context,
+                    )
                 create_cmd.set_attribute_value('enum_values', shell.elements)
                 create_cmd.set_attribute_value('is_final', True)
                 create_cmd.set_attribute_value('bases', [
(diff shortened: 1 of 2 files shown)


edb/schema/scalars.py
```diff
@@ -280,6 +280,11 @@ def _cmd_tree_from_ast(
 
                 shell = bases[0]
                 assert isinstance(shell, AnonymousEnumTypeShell)
+                if len(set(shell.elements)) != len(shell.elements):
+                    raise errors.SchemaDefinitionError(
+                        f'enum labels are not unique',
```
1st1, line 285: Better error message: there are duplicate enum values

edb/schema/scalars.py
```diff
@@ -354,10 +359,50 @@ def apply(
...
+            all_bases = []
+
+            for bases, pos in self.added_bases:
+                for base in bases:
+                    if isinstance(base, AnonymousEnumTypeShell):
+                        is_enum_base = True
+                    elif isinstance(base, s_types.TypeShell):
+                        is_enum_base = base.resolve(schema).is_enum(schema)
+                    else:
+                        is_enum_base = base.is_enum(schema)
```
1st1, line 379: This code seems wrong. If you have two new bases "Enum, NonEnum" the `is_enum_base` would be `False` after the loop.

edb/schema/scalars.py
vpetrovykh: Updated.

edb/schema/scalars.py
vpetrovykh: The `is_enum_base` is only used inside the loop to weed out the cases where there's something other than enums in the bases list. Beyond the loop it is assumed that the bases are enums-only, since otherwise the loop would have produced an error.

1st1 (approved):

edb/schema/scalars.py
1st1: my bad, you're right

edb/schema/scalars.py
```diff
@@ -354,10 +359,53 @@ def apply(
...
+                        raise errors.SchemaError(
+                            f'cannot add another type as supertype, '
+                            f'enumeration must be the only supertype specified'
+                        )
+
+                # Since all bases are enums at this point, error
+                # messages only mention about enums.
+                if pos:
+                    raise errors.SchemaError(
+                        f'cannot add another enum as supertype '
```
msullivan, line 392: I think this needs some punctuation between the two parts.

PR #1958, opened 2020-11-05
Change the lower bound of type annotations tag to 0x80

1st1:
This way it's easy to test if a tag is an annotation or not
with a simple bitwise check: 0x80 & tag == 0x80.

### DIFF
--- a/edb/server/compiler/sertypes.py
+++ b/edb/server/compiler/sertypes.py
@@ -463,7 +463,7 @@ class TypeSerializer:
             return ArrayDesc(
                 tid=tid, dim_len=dim_len, subtype=codecs_list[pos])
 
-        elif (t >= 0x7f and t <= 0xff):
+        elif (t >= 0x80 and t <= 0xff):
             # Ignore all type annotations.
             desc.read_len32_prefixed_bytes()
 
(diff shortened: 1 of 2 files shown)


edb/server/compiler/sertypes.py
```diff
@@ -480,7 +481,9 @@ def parse(cls, typedesc: bytes) -> TypeDesc:
             desc = cls._parse(wrapped, codecs_list)
             if desc is not None:
                 codecs_list.append(desc)
-        return desc
+        if not codecs_list:
+            raise errors.InternalServerError('could not parse type descriptor')
+        return codecs_list[-1]
```
tailhook, line 486: Not sure how this is related to the issue...

tailhook (approved):

tailhook:
Also, please fix protocol docs!

edb/server/compiler/sertypes.py
1st1: It's in a separate commit

1st1:
> Also, please fix protocol docs!

I think I did, there's a change in the `docs/internals/protocol/typedesc.rst` file. Anything else?

tailhook:
> > Also, please fix protocol docs!
> 
> I think I did, there's a change in the `docs/internals/protocol/typedesc.rst` file. Anything else?

No as long as I know. Was it originally there, and I've just missed that? Ah, nevermind.

PR #1959, opened 2020-11-06
Reparse query on syntax error

### LINKED ISSUE #1496: Bad error reporting after constant extraction
```
edgedb> SELECT ''' 1 ''';
error: Unexpected '('
  ┌─ query:1:10
  │
1 │ SELECT ''' 1 ''';
  │          ^^^^^^ error
```
There is no paren in original text so error is misleading.

tailhook:
Here are couple of things to watch out:
1. First commit starts sending original query text to compiler instead of cache key (which is a compressed and perhaps normalized query)
2. I'm not sure if it's save to reuse `CommandContext` as it is done in the PR

Fixes #1496

### DIFF
--- a/edb/edgeql/tokenizer.py
+++ b/edb/edgeql/tokenizer.py
@@ -20,6 +20,7 @@ from __future__ import annotations
 from typing import *
 
 import re
+import hashlib
 
 from edb._edgeql_rust import tokenize as _tokenize, TokenizerError, Token
 from edb._edgeql_rust import normalize as _normalize, Entry
@@ -33,11 +34,15 @@ TRAILING_WS_IN_CONTINUATION = re.compile(r'\\ \s+\n')
 class Source:
 
     def __init__(self, text: str, tokens: List[Token]) -> None:
+        self._cache_key = hashlib.blake2b(text.encode('utf-8')).digest()
         self._text = text
         self._tokens = tokens
 
     def text(self) -> str:
         return self._text
+
+    def cache_key(self) -> bytes:
+        return self._cache_key
 
     def variables(self) -> Dict[str, Any]:
         return {}
@@ -63,6 +68,7 @@ class NormalizedSource(Source):
 
     def __init__(self, normalized: Entry, text: str) -> None:
         self._text = text
+        self._cache_key = normalized.key()
         self._tokens = normalized.tokens()
         self._variables = normalized.variables()
         self._first_extra = normalized.first_extra()
@@ -71,6 +77,9 @@ class NormalizedSource(Source):
 
     def text(self) -> str:
         return self._text
+
+    def cache_key(self) -> bytes:
+        return self._cache_key
 
     def variables(self) -> Dict[str, Any]:
         return self._variables
@@ -90,12 +99,6 @@ class NormalizedSource(Source):
     @classmethod
     def from_string(cls, text: str) -> NormalizedSource:
         return cls(normalize(text), text)
-
-    @classmethod
-    def from_string_with_cache_key(cls, text: str) -> (NormalizedSource, str):
-        """Returns a tuple of Source and canonical cache key for the query"""
-        normalized = normalize(text)
-        return cls(normalized, text), normalized.key()
 
 
 def tokenize(eql: str) -> List[Token]:
--- a/edb/edgeql/tokenizer.py
+++ b/edb/edgeql/tokenizer.py
@@ -61,8 +61,8 @@ class Source:
 
 class NormalizedSource(Source):
 
-    def __init__(self, normalized: Entry) -> None:
-        self._text = normalized.key()
+    def __init__(self, normalized: Entry, text: str) -> None:
+        self._text = text
         self._tokens = normalized.tokens()
         self._variables = normalized.variables()
         self._first_extra = normalized.first_extra()
@@ -89,7 +89,13 @@ class NormalizedSource(Source):
 
     @classmethod
     def from_string(cls, text: str) -> NormalizedSource:
-        return cls(normalize(text))
+        return cls(normalize(text), text)
+
+    @classmethod
+    def from_string_with_cache_key(cls, text: str) -> (NormalizedSource, str):
+        """Returns a tuple of Source and canonical cache key for the query"""
+        normalized = normalize(text)
+        return cls(normalized, text), normalized.key()
 
 
 def tokenize(eql: str) -> List[Token]:
--- a/edb/server/compiler/compiler.py
+++ b/edb/server/compiler/compiler.py
@@ -1415,6 +1415,32 @@ class Compiler(BaseCompiler):
         ctx: CompileContext,
         source: edgeql.Source,
     ) -> List[dbstate.QueryUnit]:
+        try:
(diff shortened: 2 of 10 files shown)


edb/edgeql/tokenizer.py
```diff
@@ -89,7 +89,12 @@ def extra_blob(self) -> bytes:
 
     @classmethod
     def from_string(cls, text: str) -> NormalizedSource:
-        return cls(normalize(text))
+        return cls(normalize(text), text)
```
elprans, line 92: Perhaps make `cache_key` a member of `Source` instead and then you won't have to use a tuple?

edb/server/compiler/compiler.py
```diff
@@ -1415,6 +1415,30 @@ def _compile(
...
+            if isinstance(source, edgeql.NormalizedSource):
+                # try non-normalized source
+                try:
+                    original = edgeql.Source.from_string(source.text())
+                    ctx = dataclasses.replace(ctx, original)
+                    return self._try_compile(ctx=ctx, source=original)
+                except errors.EdgeQLSyntaxError as e:
+                    raise e
+                else:
+                    raise AssertionError(
```
1st1, line 1431: I'd raise `InternalServerError` and add more context: the normalized query text; the type and text of the original error; the type and text of the new error. I know this is unlikely to happen, but if it does it would help us to debug this.

1st1 (approved):
lgtm, but please address my comment before merging.

edb/edgeql/tokenizer.py
1st1: +1. I actually didn't like this method too.

tailhook: It though about this. But we already transfer query to the complier two times (once as a source and other time as a token list). Adding a cache_key here would add a third time. I remember we discussed few megabyte-sized queries, so that amount of tranfer may be significant. But cache_key is useless in the compiler.

edb/server/compiler/compiler.py
tailhook: Both of the errors are in `__source__`:
```
        File "/work/edb/server/compiler/compiler.py", line 1419, in _compile
          return self._try_compile(ctx=ctx, source=source)
        File "/work/edb/server/compiler/compiler.py", line 1453, in _try_compile
          statements = edgeql.parse_block(source)
        File "/work/edb/edgeql/parser/__init__.py", line 69, in parse_block
          return parser.parse(source)
        File "/work/edb/common/parsing.py", line 415, in parse
          raise self.get_exception(
      edb.errors.EdgeQLSyntaxError: Unexpected '('

      During handling of the above exception, another exception occurred:

        File "/work/edb/server/compiler/compiler.py", line 1427, in _compile
          raise RuntimeError("synthetic error");
      RuntimeError: synthetic error

      During handling of the above exception, another exception occurred:

        File "/work/edb/server/procpool/worker.py", line 75, in worker
          res = await meth(*args)
        File "/work/edb/server/compiler/compiler.py", line 1869, in compile
          return self._compile(ctx=ctx, source=source)
        File "/work/edb/server/compiler/compiler.py", line 1432, in _compile
          raise AssertionError(
      AssertionError: Normalized and non-normalized query errors differ
```

tailhook:
With more testing I've found following things:
1. The attached test doesn't reproduce original issue, neither I could reproduce it in test_edgeql_expressions test. Is there something that disables constant extraction in tests?
2. Most likely "Normalized and non-normalized query errors differ" error would be because these queries differ in number of arguments of the query (i.e. many queries have at least one constant). I'm not sure we need to do anything about that though, looks like syntatic errors happen before checking exhaustiveness of the arguments so for known cases it works fine.

edb/edgeql/tokenizer.py
1st1: > I remember we discussed few megabyte-sized queries

Usually that was about SQL queries we generate. I'm not sure how long it would take to compile a megabyte of EdgeQL, but that time will likely dwarf the socket io time.

That said, I agree, transferring cache_key to compilers does seem wasteful.

1st1:
> The attached test doesn't reproduce original issue, neither I could reproduce it in test_edgeql_expressions test. Is there something that disables constant extraction in tests?

We have a `EDGEDB_DEBUG_EDGEQL_DISABLE_NORMALIZATION` flag but it's not set by default. I suggest adding a `print()` and double checking that normalization does happen during your specific test.

> Most likely "Normalized and non-normalized query errors differ" error would be because these queries differ in number of arguments of the query (i.e. many queries have at least one constant). I'm not sure we need to do anything about that though, looks like syntatic errors happen before checking exhaustiveness of the arguments so for known cases it works fine.

Can you check if we raise `EdgeQLSyntaxError` in the IR and SQL compilers?

edb/edgeql/tokenizer.py
elprans: Maybe make `cache_key` a hash of source then?

tailhook:
> > Most likely "Normalized and non-normalized query errors differ" error would be because these queries differ in number of arguments of the query (i.e. many queries have at least one constant). I'm not sure we need to do anything about that though, looks like syntatic errors happen before checking exhaustiveness of the arguments so for known cases it works fine.
> 
> Can you check if we raise `EdgeQLSyntaxError` in the IR and SQL compilers?

Quick grepping yields no results:
```
$ rg EdgeQLSyntaxError edb/ir
$ rg EdgeQLSyntaxError edb/pgsql
```
So we're probably safe?

(still investigating tests, they don't go through normalizer code path, but yet I don't know why)

tailhook:
> > The attached test doesn't reproduce original issue, neither I could reproduce it in test_edgeql_expressions test. Is there something that disables constant extraction in tests?
> 
> We have a `EDGEDB_DEBUG_EDGEQL_DISABLE_NORMALIZATION` flag but it's not set by default. I suggest adding a `print()` and double checking that normalization does happen during your specific test.

Well, I forgot that we don't optimize `execute` yet (because it doesn't have arguments). So using `query` reproduces the original issue.

edb/edgeql/tokenizer.py
```diff
@@ -89,7 +89,12 @@ def extra_blob(self) -> bytes:
...
+
+    @classmethod
+    def from_string_with_cache_key(cls, text: str) -> (NormalizedSource, str):
```
1st1, line 95: Please add a docstring describing what "cache key" is used for and what it is.

1st1 (approved):
Looks good, but please add a docstring to from_string_with_cache_key

edb/edgeql/tokenizer.py
tailhook: Almost lost this comment. Yes making it hash makes sense. I don't remember why it wasn't done this way from the start. Perhaps for easier debugging (i.e. `cache_key` contains text with variables extracted but in textual form, not as a list of tokens). I'll try to export something else for debugging and use a hash.

tailhook: Done

tailhook:
Made cache_key a hash instead of full text. And moved it back into the Source class.

elprans (approved):

PR #1961, opened 2020-11-07
Fix some non-delegated constraint issues

### LINKED ISSUE #1938: Unique constraint on abstract type doesn't work?
Here is the error:
```
edgedb> CREATE ABSTRACT TYPE Base { create property test1 -> str; create property test2 -> str; create constraint exclusive on ((.test1, .test2));};
OK: CREATE
edgedb> CREATE TYPE Test2 EXTENDING Base;
error: could not resolve partial path
  ┌─ query:1:28
  │
1 │ CREATE TYPE Test2 EXTENDING Base;
  │                            ^^^^^^^ error

edgedb> CREATE TYPE Test2 EXTENDING default::Base;
error: could not resolve partial path
  ┌─ query:1:28
  │
1 │ CREATE TYPE Test2 EXTENDING default::Base;
  │                            ^^^^^^^ error
```
Also note that error message is a way off, so I guess it refers to a constraint definition rather than the DDL here.

There are three possible action points:
1. Fix constraint to work
2. Or disable some kinds of constraints in abstract types (at least temporarily)
3. And figure out what's wrong with the error position and how we can fix that (I think we need to ship actual source code of the expression via error message to correctly point to the error).

**Update:** when using `create-migration` error pops up only on `migrate` the migration is created without errors.

msullivan:
* Don't fail when extending a type with a non-delegated object
   constraint that uses an implicit path prefix. Fixes #1938.
 * Fix non-delegated exclusive constraints not working properly in many
   cases.
 * As an (apparently correct) side effect of this, INSERTing something
   that would violate an exclusive constraint at the same time as
   DELETE/UPDATEing the conflicting object now fails.
 * Include correct TABLE and SCHEMA entries when raising unique_violation.
   (Instead of individuals characters from the name string :P)

### DIFF
--- a/edb/pgsql/common.py
+++ b/edb/pgsql/common.py
@@ -189,6 +189,15 @@ def get_scalar_backend_name(id, module_id, catenate=True, *, aspect=None):
     return convert_name(name, aspect, catenate)
 
 
+def get_aspect_suffix(aspect):
+    if aspect == 'table':
+        return ''
+    elif aspect == 'inhview':
+        return 't'
+    else:
+        return aspect
+
+
 def get_objtype_backend_name(id, module_id, *, catenate=True, aspect=None):
     if aspect is None:
         aspect = 'table'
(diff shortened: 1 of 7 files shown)


elprans (approved):
:+1:

PR #1964, opened 2020-11-10
Disallow scalar types with more than one concrete base

### LINKED ISSUE #1790: Prohibit inheriting scalar types from more than one non-abstract scalar type
Currently:

```
select> CREATE SCALAR TYPE test::myint EXTENDING std::int64, std::str;
OK: CREATE
```

We have to prohibit this by enforcing that there must be exactly one base non-abstract scalar type in the ancestry of a new non-abstract scalar type.

msullivan:
Fixes #1790.

### DIFF
--- a/edb/schema/scalars.py
+++ b/edb/schema/scalars.py
@@ -222,45 +222,51 @@ class ScalarTypeCommand(
     schema_metaclass=ScalarType,
     context_class=ScalarTypeCommandContext,
 ):
-    def apply(
+    def validate_scalar_ancestors(
+        self,
+        ancestors: Sequence[so.SubclassableObject],
+        schema: s_schema.Schema,
+        context: sd.CommandContext,
+    ) -> None:
+        concrete_ancestors = {
+            ancestor for ancestor in ancestors
+            if not ancestor.get_is_abstract(schema)
+        }
+        # Filter out anything that has a subclass relation with
+        # every other concrete ancestor. This lets us allow chains
+        # of concrete scalar types while prohibiting diamonds (for
+        # example if X <: A, B <: int64 where A, B are concrete).
+        # (If we wanted to allow diamonds, we could instead filter out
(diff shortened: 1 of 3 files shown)


edb/schema/scalars.py
```diff
@@ -222,7 +222,45 @@ class ScalarTypeCommand(
     schema_metaclass=ScalarType,
     context_class=ScalarTypeCommandContext,
 ):
-    pass
+    def apply(
```
elprans, line 225: It's better to overload `validate_create()` and `validate_alter()` instead of `apply()` (you almost never want to overload `apply()`).

edb/schema/scalars.py
```diff
@@ -222,7 +222,45 @@ class ScalarTypeCommand(
...
+                c1 for c1 in concrete_ancestors
+                if not all(c1 == c2 or c1.issubclass(schema, c2)
+                           or c2.issubclass(schema, c1)
+                           for c2 in concrete_ancestors)
+            }
+
+            if len(concrete_ancestors) > 1:
+                raise errors.SchemaError(
+                    f'scalar type may not have more than '
+                    f'one concrete base type',
```
elprans, line 260:
```suggestion
                    f'one concrete base type', context=self.source_context
```

edb/schema/scalars.py
msullivan: This isn't why I did it that way originally, but that approach doesn't catch the case where an existing abstract base has a new concrete base added (since the AlterScalarType that updates ancestors is generated after `validate_alter` is called, I think).

I think that we could have validate_alter explicitly search all descendants when bases are added and explicitly do this check on them. Is that an improvement over overloading `apply()`?

elprans: Yea, I think that would be a cleaner approach, I think.

elprans (approved):

PR #1965, opened 2020-11-11
Turn schema names into tuples

elprans:
Currently, `schema.Name` is a weird subclass of `str` which adds the
`module` and the `name` attributes to it to access the name components. 
There are problems with this approach: it allows mixing parsed and unparsed
names freely, since all names are strings, and also wastes space since the
name's components are stored alongside the full, doublecolon-separated,
name representation.  This is evident on the size of the `dump01` schema
dump, which is ~25% smaller as a result of this change.

[There is a bunch of preparatory commits, please review individually]


edb/edgeql/compiler/context.py
```diff
@@ -67,7 +67,7 @@ def __init__(
         source: s_sources.Source,
         *,
         ptrcls: Optional[s_pointers.Pointer],
-        ptrcls_name: Optional[s_name.Name] = None,
+        ptrcls_name: Optional[s_name.QualifiedName] = None,
```
1st1, line 70: nit: I'd call it `QualName`

edb/edgeql/compiler/clauses.py
```diff
@@ -48,7 +49,7 @@ def compile_where_clause(
     with ctx.newscope(fenced=True) as subctx:
         subctx.path_scope.unnest_fence = True
         ir_expr = dispatch.compile(where, ctx=subctx)
-        bool_t = ctx.env.get_track_schema_type('std::bool')
+        bool_t = ctx.env.get_track_schema_type(sn.QualifiedName('std', 'bool'))
```
1st1, line 52: Would be cool to have `sn.qname('std::bool')` with an LRU cache inside. That would save some keystrokes and maybe even make things a bit faster.

edb/schema/name.py
```diff
@@ -20,102 +20,244 @@
...
-        nqname = name.name
-    else:
-        module = None
-        nqname = name
+NameT = TypeVar("NameT", bound="Name")
+QualifiedNameT = TypeVar("QualifiedNameT", bound="QualifiedName")
+UnqualifiedNameT = TypeVar("UnqualifiedNameT", bound="UnqualifiedName")
+
+
+if TYPE_CHECKING:
```
1st1, line 35: Can you add a comment explaining why you have this if here? I don't get it, for example.

1st1 (approved):
Impressive pr!

edb/common/topological.py
```diff
@@ -29,48 +30,85 @@ class UnresolvedReferenceError(Exception):
...
 
-def sort(graph, *, return_record=False, allow_unresolved=False):
-    adj = defaultdict(OrderedSet)
-    loop_control = defaultdict(OrderedSet)
+K = TypeVar('K')
+V = TypeVar('V', contravariant=True)
+T = TypeVar('T')
+
+
+class DepGraphEntry(Generic[K, V, T]):
```
msullivan, line 43: I always like to see dicts-as-objects going away

edb/edgeql/compiler/clauses.py
msullivan: :+1:

edb/edgeql/compiler/context.py
elprans: Done.

edb/schema/name.py
elprans: Added a comment.

edb/edgeql/compiler/clauses.py
elprans: I dunno.

`sn.QualName('std', 'bool')`
`sn.qname('std::bool')` 

Doesn't seem like it's worth an extra wrapper.  Even `lru_cache` has a cost.

msullivan (approved):
Nice to see this particular confusion resolved!

edb/edgeql/compiler/clauses.py
1st1: imo the latter is easier to type; python's lru_cache is in C and so caching by a str is extremely fast.

tailhook (approved):

edb/graphql/types.py
```diff
@@ -980,16 +1156,20 @@ def get_order_fields(self, typename):
...
                         self._gql_ordertypes['Ordering']
                     )
-            elif t.is_object_type():
+            elif isinstance(t, s_objtypes.ObjectType):
                 # It's a link so we need the link's type order input
                 t_name = t.get_name(self.edb_schema)
-                fields[name] = self._gql_ordertypes[t_name]
-
-            # We ignore pointers that aren't scalars or objects.
+                # XXX: why is below not a GraphQLField?
```
vpetrovykh, line 1162: You are right, this should be wrapped into a GraphQLInputObjectField, just like above.

edb/edgeql/compiler/clauses.py
tailhook: It's also memory cost. Compilers aren't lightweight already. So I'm with @elprans here.

edb/graphql/types.py
```diff
@@ -1022,27 +1203,36 @@ def _define_types(self):
...
                     fields=partial(self.get_fields, t_name),
-                    resolve_type=lambda obj, info: obj,
+                    resolve_type=_type_resolver,
                     description=self._get_description(t),
                 )
-            self._gql_interfaces[t_name] = gqltype
+            # XXX: graphql-core typing annotations don't
+            #      seem to indicate that it is fine to mix
+            #      GraphQLObjectType with GraphQLInterfaceType
+            self._gql_interfaces[t_name] = gqltype  # type: ignore
```
vpetrovykh, line 1221: Conceptually interfaces are type-like, they represent potentially more than one type, but otherwise are used in most of the same places. They would have slightly different properties (such as that types can implement interfaces, but obviously not other types, etc.). I wonder if potentially asserting that what we use as `interfaces` is actually a `GraphQLInterfaceType` specifically would solve the typing complaint. Perhaps there are other places as well where we would need a similar assert. I can do that after you merge this PR, though.

edb/edgeql/compiler/clauses.py
elprans: Aside from performance considerations, I specifically didn't want to keep it easy to treat names as unparsed strings.  A handy casting shortcut creates too much incentive to keep names as strings.  Also, there aren't _that_ many places where the name of the object is a hardcoded string.

PR #1966, opened 2020-11-12
Pick off some (very) low handing fruit in compiler perf area

elprans:
A few simple micro-optimisations and caches result in ~20% perf boost.
The baseline is the names-as-tuples branch, which regressed a bit compared
to master (which spurred me to look into clawing some of that back).

    compile_edgeql_script_simple: Mean +- std dev: 1.48 ms += 0.02 ms (6.18% faster)
    compile_edgeql_script_webapp_configure_system: Mean +- std dev: 19.4 ms += 0.2 ms (18.40% faster)
    compile_edgeql_script_webapp_get_movie: Mean +- std dev: 117 ms += 2 ms (19.55% faster)
    compile_edgeql_script_webapp_get_person: Mean +- std dev: 135 ms += 3 ms (22.30% faster)
    compile_edgeql_script_webapp_get_user: Mean +- std dev: 72.4 ms += 0.9 ms (23.68% faster)
    compile_edgeql_script_webapp_insert_movie: Mean +- std dev: 71.3 ms += 1.1 ms (16.42% faster)
    compile_edgeql_script_webapp_insert_person: Mean +- std dev: 20.3 ms += 0.2 ms (16.76% faster)
    compile_edgeql_script_webapp_insert_review: Mean +- std dev: 36.0 ms += 0.6 ms (14.46% faster)


edb/common/struct.py
```diff
@@ -347,14 +359,6 @@ def _init_fields(
 
             setattr(self, field_name, value)
 
-    def __setattr__(self, name: str, value: Any) -> None:
```
1st1, line 350: I'm wondering if we can speed up `_init_fields` by not calling `_getdefault` and instead implementing `__getattr__` to call it dynamically when needed.

edb/schema/name.py
```diff
@@ -179,15 +179,20 @@ def mangle_name(name: str) -> str:
...
 
 
+mangle_re_1 = re.compile(r'(?<![|])\|(?![|])')
+mangle_re_2 = re.compile(r'(?<![&])&(?![&])')
+
+
 def unmangle_name(name: str) -> str:
-    name = re.sub(r'(?<![|])\|(?![|])', '::', name)
-    name = re.sub(r'(?<![&])&(?![&])', '@', name)
+    name = mangle_re_1.sub('::', name)
```
1st1, line 187: Does this func need an lru_cache?

edb/common/compiler.py
```diff
@@ -95,7 +94,7 @@ def _push(
             prevlevel = self.current
             level = self.ContextLevelClass(prevlevel, mode)
         # XXX: typing fu
```
1st1, line 96: looks like the XXX is outdated

edb/common/ordered.py
```diff
@@ -28,35 +28,29 @@
...
+            self.map = {v: None for v in iterable}
+        else:
+            self.map = {}
 
-    def add(self, item: K, *, last: Optional[bool] = None) -> None:
-        self.map[item] = item
-        if last is not None:
-            self.map.move_to_end(item, last=last)
+    def add(self, item: K) -> None:
+        self.map[item] = None
```
1st1, line 41: You can push an item to be first easily:

```
>>> a = {'a': 1, 'b': 2, 'c': 3}
>>> b = {'c': None}
>>> b.update(a)
>>> b
{'c': 3, 'a': 1, 'b': 2}
```

You can also push it last, if you just reconstruct the dict. Both are O(n) technically, although the code sample above should be fast enough as it's all in C.

1st1 (approved):

edb/schema/name.py
elprans: No, it's called by another function that is already cached.

edb/common/ordered.py
elprans: Yes, but we don't need this anywhere.  The only place where this is used is DDL subcommands, and there I just used a list.

edb/common/struct.py
elprans: Overloading `__getattr__` has a steep perf penalty even if it's very simple.  It also kills typing checks.

1st1: > Overloading __getattr__ has a steep perf penalty even if it's very simple. 


Why? It would only be called for slots that weren't initialized via the constructor. Essentially lazy defaults. You can even cache defaults (write them in) when `__getattr__` is invoked on them.

> It also kills typing checks.

Why? Literally nothing changes.

elprans: > Why? Literally nothing changes.

Mypy stops typechecking attribute accesses when there is a `__getattr__` on a class.

PR #1971, opened 2020-11-14
Support renaming functions

msullivan:
We do this by adding a func_id field that is used to determine the
backend name and that is shared between all functions sharing a name.

### DIFF
--- a/edb/schema/functions.py
+++ b/edb/schema/functions.py
@@ -1709,7 +1709,8 @@ class RenameFunction(RenameCallableObject[Function], FunctionCommand):
         if target:
             raise errors.SchemaError(
                 f"can not rename function to '{new_name!s}' because "
-                f"a function with the same name already exists, and renaming into an overload is not supported",
+                f"a function with the same name already exists, and "
+                f"renaming into an overload is not supported",
                 context=self.source_context)
 
 
--- a/edb/schema/functions.py
+++ b/edb/schema/functions.py
@@ -21,6 +21,7 @@ from __future__ import annotations
 
 import abc
 import types
+import uuid
 from typing import *
 
 from edb import errors
@@ -29,6 +30,7 @@ from edb.edgeql import ast as qlast
 from edb.edgeql import compiler as qlcompiler
 from edb.edgeql import qltypes as ft
 from edb.edgeql import parser as qlparser
+from edb.common import uuidgen
 
 from . import abc as s_abc
 from . import annos as s_anno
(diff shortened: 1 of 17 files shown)


edb/schema/functions.py
```diff
@@ -1645,7 +1665,52 @@ def _apply_field_ast(
...
+        if len(existing) > 1:
+            raise errors.SchemaError(
+                'renaming an overloaded function is not allowed',
+                context=self.source_context)
+
+        target = schema.get_functions(new_name, ())
+        if target:
+            raise errors.SchemaError(
+                f"can not rename function to '{new_name!s}' because "
+                f"it already exists",
```
elprans, line 1712:
```suggestion
                f"a function with the same name already exists, and renaming into an overload is not supported",
```

elprans (approved):

1st1:
I don't like the `func_id` name. I'd suggest `backend_name` instead.

tests/test_edgeql_ddl.py
```diff
@@ -2642,6 +2642,141 @@ async def test_edgeql_ddl_function_29(self):
...
+
+            CREATE FUNCTION test::foo(s: int64) -> str {
+                USING (SELECT <str>s)
+            };
+        """)
+
+        with self.assertRaisesRegex(
+                edgedb.SchemaError,
+                r"renaming an overloaded function is not allowed"):
+            await self.con.execute("""
```
1st1, line 2705: So how would I do rename 2 or more functions sharing the same name? Recreate all functions from scratch?

msullivan: Yeah, unfortunately.

1st1: Can we allow a special form of syntax like `ALTER FUNCTION foo::bar RENAME TO foo::baz` -- note that there are no parenthesis there. @elprans

elprans: Actually, I think we should allow parentheses-free form to refer to a _single_, non-overloaded function, like Postgres does.  The vast majority of user-defined functions are not overloaded, and having to retype full parameter definition is an unnecessary hassle (which is why Postgres implemented this).

1st1: I'm not against that. But if we have more than one function defined for the name, would it make sense to allow renaming all of them with this syntax (I think it would)?

elprans: Would `DROP FUNCTION foo::bar` drop all overloads too?

1st1: That's a good question.  Probably that would be too dangerous. Maybe we should have `ALTER ALL FUNCTION foo::bar RENAME TO` and `DROP ALL FUNCTION foo::bar`

PR #1975, opened 2020-11-16
Make sure an alter can be generated for low-similarity pairs

elprans:
For low similarity scores (< 0.6) we generate a `DROP` followed by a
`CREATE`, treating the pair as disinct objects.  However, the intent might
still be to actually alter (and possibly rename) the existing type and
hence we must handle the rejection of `DROP`/`CREATE` appropriately.

### DIFF
--- a/edb/schema/constraints.py
+++ b/edb/schema/constraints.py
@@ -268,11 +268,22 @@ class Constraint(referencing.ReferencedInheritingObject,
         self_schema: s_schema.Schema,
         other_schema: s_schema.Schema,
         context: so.ComparisonContext,
-    ) -> sd.Command:
+    ) -> sd.ObjectCommand[Constraint]:
         return super().as_alter_delta(
             other,
             self_schema=self_schema,
             other_schema=other_schema,
+            context=context,
+        )
+
+    def as_delete_delta(
+        self,
+        *,
+        schema: s_schema.Schema,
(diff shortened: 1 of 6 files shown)


edb/schema/delta.py
```diff
@@ -115,6 +133,7 @@ def delta_objects(
 
     seen_x = set()
     seen_y = set()
+    x_counter: Dict[so.Object_T, int] = collections.defaultdict(int)
```
msullivan, line 136: This seems dead?

edb/schema/constraints.py
```diff
@@ -268,14 +268,25 @@ def as_alter_delta(
...
-    ) -> sd.Command:
+    ) -> sd.ObjectCommand[Constraint]:
         return super().as_alter_delta(
             other,
             self_schema=self_schema,
             other_schema=other_schema,
             context=context,
         )
 
+    def as_delete_delta(
```
msullivan, line 279: Is this needed, if it just makes a super call?

msullivan (approved):
Seems good

edb/schema/constraints.py
elprans: mypy wants this for some reason

edb/schema/delta.py
elprans: Ah, yeah. Part of followup PR

PR #1978, opened 2020-11-17
Remove xfail on a test fixed in #1976

msullivan:


elprans (approved):
One down, score to go :-)

PR #1979, opened 2020-11-17
Put all edgeql modules in the same pgsql schema

msullivan:
This just does the bare minimum to make it work, and I plan to go on a
rampage deleting now-dead code in a follow up.

I could include that rampage in this PR if that is preferred.

### DIFF
--- a/edb/pgsql/common.py
+++ b/edb/pgsql/common.py
@@ -38,6 +38,7 @@ from edb.schema import operators as s_opers
 from edb.schema import pointers as s_pointers
 from edb.schema import scalars as s_scalars
 from edb.schema import types as s_types
+from edb.schema import schema as s_schema
 
 from . import keywords as pg_keywords
 
@@ -91,6 +92,7 @@ def needs_quoting(string):
 
 
 def qname(*parts):
+    assert len(parts) <= 3, parts
     return '.'.join([quote_ident(q) for q in parts])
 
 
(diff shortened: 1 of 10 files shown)


edb/pgsql/common.py
```diff
@@ -122,7 +122,9 @@ def quote_type(type_):
 
 
 def get_module_backend_name(module, prefix='edgedb_'):
-    return edgedb_name_to_pg_name(f'{prefix}{module}', len(prefix))
+    # We now put everything in the same schema, but haven't yet cleaned up
+    # all the call sites.
+    return "edgedb"
```
1st1, line 127: I'm wondering if there's any advantage in keeping the `std::` namespace in a separate schema. @elprans ?

elprans, line 127: We should pick a different schema name.  `edgedb` is reserved for internal structures, functions etc.  Additionally, the standard library should also use a separate schema.  The reason here is simple: dumps, logical replication, upgrades etc are all much simpler if we cleanly separate system structures from user-defined structures.

I propose:

`edgedbstd`: standard library and internal helpers (`metaschema`)
`edgedbinstdata`: unique instance data (already present)
`edgedbpub`: user data

1st1: +1. Except I'd change `edgedbinstdata` to edgedbinst` (or `edgedbpub` to `edgedbpubdata`). I like the former.

msullivan: OK, I did a split between `edgedbstd` for the standard library (everything in STD_MODULES) and `edgedbpub` for everything else.

I haven't moved the metaschema helpers from `edgedb` to `edgedbstd`; that seems orthogonal to this change.

elprans (approved):
Don't forget to bump CATALOG_VERSION

edb/pgsql/common.py
```diff
@@ -121,8 +123,11 @@ def quote_type(type_):
     return first + last
 
 
-def get_module_backend_name(module, prefix='edgedb_'):
-    return edgedb_name_to_pg_name(f'{prefix}{module}', len(prefix))
+def get_module_backend_name(module):
+    # standard modules go into "edgedbstd", user ones into "edgedbpub"
+    assert isinstance(module, str) and "-" not in module, (
```
1st1, line 128: Do we have uuid cast to strings floating around?

elprans: Actually, a dash in the module name isn't prohibited, I can do 

```
CREATE MODULE `foo-bar`
```

so this will trip up.  I suggest dropping the check.

1st1: @msullivan Sully, could you please add one test that uses a quoted module name with a space in it?

PR #1980, opened 2020-11-18
mypy: Turn off the global ignore_errors=True

msullivan:
I turned it back on for a few modules, but fixed/type: ignored errors in
most of them.

I also rampaged through mypy.ini and merged a lot of the sections.


edb/common/ast/base.py
```diff
@@ -234,8 +234,8 @@ def _check_type_real(type_, value, raise_error):
 
 
 class AST(object, metaclass=MetaAST):
-    __fields = []
-    __ast_frozen_fields__ = frozenset()
+    __fields = []  # type: List[str]
+    __ast_frozen_fields__ = frozenset()  # type: AbstractSet[str]
```
elprans, line 238: Why type comments and not type annotations as usual?

elprans (approved):
A smaller diff than I expected :-)

edb/common/ast/base.py
msullivan: Oh, lol: because the struct mechanism interprets type annotations, which means that annotating these breaks everything!

(I had meant to write a comment explaining this and got distracted)

1st1: Haha, yes, a comment is definitely needed here :)

PR #1988, opened 2020-11-20
Fix deletion of collection types when they are orphaned

msullivan:
This fixes a bunch of test_schema tests with DDL migrations turned on.

### DIFF
--- a/edb/schema/delta.py
+++ b/edb/schema/delta.py
@@ -2715,6 +2715,11 @@ class DeleteObject(ObjectCommand[so.Object_T], Generic[so.Object_T]):
     #: in the schema.
     if_unused = struct.Field(bool, default=False)
 
+    #: Potential references to this object that we know are being
+    #: deleted, which we use to resolve if_unused.
+    expiring_refs = struct.Field(AbstractSet[so.Object],
+                                 default=frozenset())
+
     def _delete_begin(
         self,
         schema: s_schema.Schema,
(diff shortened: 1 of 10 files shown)


edb/schema/delta.py
```diff
@@ -2715,6 +2715,11 @@ class DeleteObject(ObjectCommand[so.Object_T], Generic[so.Object_T]):
     #: in the schema.
     if_unused = struct.Field(bool, default=False)
 
+    #: Potential references to this object that we know are being
+    #: deleted, which we use to resolve if_unused.
+    expiring_refs = struct.Field(AbstractSet[so.Object],
+                                 default=frozenset())
```
1st1, line 2721: Should this be an `ephemeral` field?

elprans: It's not a `schema.Field` on an object, just a delta command field.

edb/schema/types.py
```diff
@@ -861,6 +865,7 @@ def get_schema_class_displayname(cls) -> str:
     def as_colltype_delete_delta(
         self,
         schema: s_schema.Schema,
+        expiring_refs: AbstractSet[so.Object],
```
elprans, line 868: Let's make this a kwarg

msullivan:
Any more feedback, or is this ready to merge?

elprans (approved):
LGTM

PR #1996, opened 2020-11-25
Fix a cluster of bugs related to rebasing and constraints

msullivan:
* SET OWNED needs to proceed recursively (or else things like
   constraints and link properties will be lost when a parent is
   dropped)
 * When parents are dropped, we need to go through the object and
   find parent objects for refs and rebase those
 * We need to drop a view while rebasing
 * AlterConstraint's pgsql delta can't use the original_schema since
   the constraint might not have existed in the original schema

### DIFF
--- a/tests/test_edgeql_ddl.py
+++ b/tests/test_edgeql_ddl.py
@@ -1084,6 +1084,7 @@ class TestEdgeQLDDL(tb.DDLTestCase):
             CREATE TYPE Child EXTENDING Parent {
                 ALTER PROPERTY name {
                     SET OWNED;
+                    ALTER CONSTRAINT exclusive SET OWNED;
                 };
             };
         """)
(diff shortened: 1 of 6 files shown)


tests/test_edgeql_ddl.py
```diff
@@ -1043,6 +1043,182 @@ async def test_edgeql_ddl_28(self):
...
+            SET MODULE test;
+
+            CREATE TYPE Parent {
+                CREATE PROPERTY name -> str {
+                    CREATE CONSTRAINT exclusive;
+                };
+            };
+            CREATE TYPE Child EXTENDING Parent {
+                ALTER PROPERTY name {
+                    SET OWNED;
```
elprans, line 1086: Note: you don't need to `SET OWNED` if there are other operations in your `ALTER`, it's set implicitly since you're making local alters.

elprans (approved):
:+1:

tests/test_edgeql_ddl.py
msullivan: Roger. In this case I think it makes things more clear.

Issue #117, opened 2017-12-12
Computable links validation


vpetrovykh:
In EdgeDB a computable link is indistinguishable in behavior from a regular link. Which means that just like a regular link a computable link must point to *distinct* targets. This means that if we cannot prove that a computable link target only contains distinct objects, we must raise a compile time error.

E.g. the following is an error:
```
SELECT Issue {
    foo := Issue.owner UNION User
};
```
To resolve the error, an explicit `DISTINCT` can be added:
```
SELECT Issue {
    foo := DISTINCT (Issue.owner UNION User)
};
```

In particular, Views with some defined computable links are indistinguishable from Concepts with those same links for the purposes of querying. For example, a path involving a View is guaranteed to produce distinct objects.

1st1:
@elprans are we ready to close this?

elprans:
This is still outstanding.

1st1:
@elprans please lay out the plan to fix this for @vpetrovykh or others.

elprans:
The plan would be to introduce the new inferer for distinctness and enforce it like we do cardinality.

PR #1997, opened 2020-11-25
Multiplicity inference and enforcement

vpetrovykh:
Implement some basic multiplicity inference.
    
The inference is currently hidden behind a `validate_multiplicity`
compiler option due to current limitations of using `DISTINCT` to "fix"
multiplicity inference issues. These issues arise in EdgeQL generated
for schema objects and more broadly for specifying multi links with link
properties using `FOR`.
    
Currently, only the multiplicity tests are actually running with the flag enabled.
    
Issue #117

### DIFF
--- a/edb/edgeql/compiler/inference/multiplicity.py
+++ b/edb/edgeql/compiler/inference/multiplicity.py
@@ -1,0 +1,690 @@
+#
+# This source file is part of the EdgeDB open source project.
+#
+# Copyright 2020-present MagicStack Inc. and the EdgeDB authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+"""EdgeQL multiplicity inference.
+
+A top-down multiplicity inferer that traverses the full AST populating
+multiplicity fields and performing multiplicity checks.
+"""
+
+
+from __future__ import annotations
+from typing import *
+
+import functools
+
+from edb import errors
+
+from edb.edgeql import qltypes
+
+from edb.schema import pointers as s_pointers
+
+from edb.ir import ast as irast
+from edb.ir import typeutils as irtyputils
+
+from . import cardinality
+from . import context as inference_context
+
+
+ZERO = qltypes.Multiplicity.ZERO
+ONE = qltypes.Multiplicity.ONE
+MANY = qltypes.Multiplicity.MANY
+
+
+def _max_multiplicity(
+    args: Iterable[qltypes.Multiplicity]
+) -> qltypes.Multiplicity:
+    # Coincidentally, the lexical order of multiplicity is opposite of
+    # order of multiplicity values.
+    arg_list = list(args)
+    if not arg_list:
+        return ZERO
+    else:
+        return min(arg_list)
+
+
+def _common_multiplicity(
+    args: Iterable[irast.Base],
+    *,
+    scope_tree: irast.ScopeTreeNode,
+    ctx: inference_context.InfCtx,
+) -> qltypes.Multiplicity:
+    return _max_multiplicity(
+        infer_multiplicity(a, scope_tree=scope_tree, ctx=ctx) for a in args)
+
+
+@functools.singledispatch
+def _infer_multiplicity(
+    ir: irast.Expr,
+    *,
+    scope_tree: irast.ScopeTreeNode,
+    ctx: inference_context.InfCtx,
+) -> qltypes.Multiplicity:
+    # return MANY
+    raise ValueError(f'infer_multiplicity: cannot handle {ir!r}')
+
+
+@_infer_multiplicity.register
+def __infer_none(
+    ir: None,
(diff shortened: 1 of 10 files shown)


edb/edgeql/compiler/inference/multiplicity.py
```diff
@@ -0,0 +1,701 @@
...
+
+
+@_infer_multiplicity.register
+def __infer_type_introspection(
+    ir: irast.TypeIntrospection,
+    *,
+    scope_tree: irast.ScopeTreeNode,
+    ctx: inference_context.InfCtx,
+) -> qltypes.Multiplicity:
+    # The result is always ONE, but we still want to actually
```
elprans, line 122: This comment seems out of place.

edb/edgeql/compiler/inference/multiplicity.py
```diff
@@ -0,0 +1,701 @@
...
+) -> None:
+    for shape_set, _ in ir.shape:
+        new_scope = cardinality._get_set_scope(shape_set, scope_tree)
+        if shape_set.expr and shape_set.rptr:
+            expr_mult = infer_multiplicity(
+                shape_set.expr, scope_tree=new_scope, ctx=ctx)
+
+            ptrref = shape_set.rptr.ptrref
+            if (expr_mult is MANY and
+                    not ptrref.out_target.is_scalar and
```
elprans, line 142: Use helpers from `irtypeutils.py` instead of accessing `TypeRef` members directly.

edb/edgeql/compiler/inference/multiplicity.py
```diff
@@ -0,0 +1,701 @@
...
+    *,
+    is_mutation: bool=False,
+    scope_tree: irast.ScopeTreeNode,
+    ctx: inference_context.InfCtx,
+) -> qltypes.Multiplicity:
+    rptr = ir.rptr
+    new_scope = cardinality._get_set_scope(ir, scope_tree)
+
+    if rptr is not None:
+        # Validate the source regardless
```
elprans, line 186: Regardless of what?

edb/edgeql/compiler/inference/multiplicity.py
```diff
@@ -0,0 +1,701 @@
...
+    if rptr is not None:
+        rptrref = rptr.ptrref
+
+        # Validate the source
+        infer_multiplicity(rptr.source, scope_tree=new_scope, ctx=ctx)
+
+        if isinstance(rptr.ptrref, irast.TupleIndirectionPointerRef):
+            # All bets are off for tuple elements.
+            return MANY
+        elif ir.typeref.is_scalar or ir.typeref.collection:
```
elprans, line 201: Likewise, use `irtyputils` helpers

edb/edgeql/compiler/inference/multiplicity.py
```diff
@@ -0,0 +1,701 @@
...
+                schema = ctx.env.schema
+                exclusive_constr = schema.get(
+                    'std::exclusive', type=s_constr.Constraint)
+                # We should only have some kind of path terminating in a
+                # property here.
+                assert isinstance(rptrref, irast.PointerRef)
+                _ptr = schema.get_by_id(rptrref.id)
+                assert isinstance(_ptr, s_pointers.Pointer)
+                ptr = _ptr.get_nearest_non_derived_parent(ctx.env.schema)
+                is_unique = (
```
elprans, line 217: There is actually a `Pointer.is_exclusive()` helper for this.

edb/edgeql/compiler/inference/multiplicity.py
```diff
@@ -0,0 +1,701 @@
...
+    ir: irast.SelectStmt,
+    *,
+    scope_tree: irast.ScopeTreeNode,
+    ctx: inference_context.InfCtx,
+) -> qltypes.Multiplicity:
+
+    assert ir.iterator_stmt is not None
+    itexpr = ir.iterator_stmt.expr
+
+    if not isinstance(ir.result, irast.Set):
```
elprans, line 432: `SelectStmt.result` is always a `Set`, this check is redundant.

elprans (approved):
Looks good.  Some minor nits overall.

edb/edgeql/compiler/inference/multiplicity.py
vpetrovykh: This is actually one of the places where you could sneak multiplicity-breaking expression without triggering an error:
```
WITH X := Foo {bar := {Bar, Bar}}
SELECT INTROSPECT TYPEOF X.bar;
```
By the time we get to validating multiplicity, the original expression in the `TYPEOF` is lost. Technically, this is in the same category of nitpicking of whether we validate "unused" expressions, but if we want full consistency then this place should be checked as well.

I've added a longer comment to clarify.

edb/edgeql/compiler/inference/multiplicity.py
vpetrovykh: Updated to use `is_object` instead.

edb/edgeql/compiler/inference/multiplicity.py
vpetrovykh: Yeah, that's old comment (and the line is duplicated, too). Fixed.

edb/edgeql/compiler/inference/multiplicity.py
vpetrovykh: Updated to use `is_object` instead.

edb/edgeql/compiler/inference/multiplicity.py
vpetrovykh: Updated.

edb/edgeql/compiler/inference/multiplicity.py
vpetrovykh: Removed.

PR #1999, opened 2020-11-26
Fix migrations modifying collection aliases

msullivan:
Some Array fields in ArrayExprAlias weren't being serialized because
the introspection query was looking at its parent
CollectionExprAlias. Similar for tuples. Skip that.

The obvious tests for this didn't catch it because if the alias was
created during the same transaction, the reflection didn't come into
play.

Also, hide the CollectionExprAliases from reflection.

### DIFF
--- a/edb/schema/types.py
+++ b/edb/schema/types.py
@@ -1279,6 +1279,8 @@ class ArrayExprAlias(
     Array,
     qlkind=qltypes.SchemaObjectClass.ALIAS,
 ):
+    # N.B: Don't add any SchemaFields to this class, they won't be
+    # reflected properly (since this inherits from the concrete Array).
     pass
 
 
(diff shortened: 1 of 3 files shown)


edb/schema/types.py
```diff
@@ -2105,7 +2109,14 @@ class CreateCollectionExprAlias(
...
     sd.CreateObject[CollectionExprAliasT],
 ):
-    pass
+    def canonicalize_attributes(
+        self,
+        schema: s_schema.Schema,
+        context: sd.CommandContext,
+    ) -> s_schema.Schema:
+        schema = super().canonicalize_attributes(schema, context)
+        self.set_attribute_value('internal', True)
```
elprans, line 2118: It probably makes more sense to set 'internal' in `_handle_alias_op` as the type attribute, and I think this should also be done for non-collection alias types.

msullivan:
Oh, wait, we can't hide the type because that would break the `type` link in `schema::Alias`. Just reverting that bit and fixing the bug.

elprans (approved):

PR #2005, opened 2020-11-30
Initial implementation of capabilities

tailhook:
This is a part of RFC1004

To Do:
* [x] figure out what to do with session mode
* [x] remove `advisory_lock*` functions (I've kept until we figure out session mode)
* [x] add `SERVER_HEADER_CAPABILITIES` header
* [x] figure out how to do unit tests
* [x] `SESSION_CONFIG` and `PERSISTENT_CONFIG` capabilities are not implemented

Publishing draft so that @1st1 could take a look.

Regarding `session_mode`:
1. I don't think that this flag enabled or not should depend on kind of "port". It should depend on some command-line flag or server/database setting.
2. The `session_mode/session_only` name is misleading for `sleep` function.
3. It's not just compiler internal stuff, it's also a `session_only` flag in function that is a public API.


tailhook:
Unfortunately, neither current python nor current JavaScript currently support extra headers:
```
  File "/work/.eggs/edgedb-0.12.0a1-py3.8-linux-x86_64.egg/edgedb/blocking_con.py", line 101, in query
    return self._protocol.sync_execute_anonymous(
  File "edgedb/protocol/blocking_proto.pyx", line 91, in edgedb.protocol.blocking_proto.BlockingIOProtocol.sync_execute_anonymous
  File "edgedb/protocol/blocking_proto.pyx", line 77, in edgedb.protocol.blocking_proto.BlockingIOProtocol._iter_coroutine
  File "edgedb/protocol/protocol.pyx", line 560, in execute_anonymous
  File "edgedb/protocol/protocol.pyx", line 232, in _parse
  File "edgedb/protocol/protocol.pyx", line 162, in edgedb.protocol.protocol.SansIOProtocol.reject_headers
edgedb.errors.BinaryProtocolError: unexpected headers
```
So we have to:
1. Gate new header by the protocol version
2. Update python bindings before applying this PR :(

edb/server/mng_port/edgecon.pyx
```diff
@@ -824,6 +867,10 @@ cdef class EdgeConnection:
             self.debug_print('SIMPLE QUERY', eql)
 
         if self.dbview.in_tx_error():
+            if not (allow_capabilities & enums.Capability.TRANSACTION):
```
elprans, line 870: Which scenario triggers this?  Is this check really needed?

elprans:
Looks good to me so far!

edb/server/mng_port/edgecon.pyx
tailhook: Perhaps something like this:
```
with conn.try_transaction() as tx:
    try:
        do_something()
    except:
        tx.query("ROLLBACK")  # << here we disable transaction statements, per RFC1004
```

edb/server/mng_port/edgecon.pyx
```diff
@@ -824,6 +868,10 @@ cdef class EdgeConnection:
...
+                raise errors.DisabledCapabilityError(
+                    f"Cannot recover transaction, disallowed capability"
```
1st1, line 873: I suggest to move this check into `_recover_script_error`, where you know what was compiled, and where a correct error would be raised if the user didn't actually sent ROLLBACK or ROLLBACK TO SAVEPOINT. Then rephrase the error to

"Cannot execute ROLLBACK command; the transactions capability is disabled"

edb/server/mng_port/edgecon.pyx
```diff
@@ -812,8 +841,23 @@ cdef class EdgeConnection:
...
+        if headers:
+            for k, v in headers.items():
+                if k == QUERY_OPT_ALLOW_CAPABILITIES:
+                    if len(v) != 8:
+                        raise errors.BinaryProtocolError(
+                            f'capabilities header must be exactly 8 bytes'
+                        )
+                    allow_capabilities = hton.unpack_uint64(
+                        cpython.PyBytes_AS_STRING(v))
+                    allow_capabilities |= ALWAYS_ALLOWED
```
1st1, line 856: This code is repeated in a couple of places. Please factor it out into a method, the file is long enough already.

edb/server/notebook_port/protocol.pyx
```diff
@@ -31,10 +31,32 @@ from edb.common import markup
...
 
 
+ALLOWED_CAPABILITIES = (
+    enums.Capability.QUERY |
+    enums.Capability.MODIFICATIONS |
+    enums.Capability.DDL
+)
+
+
+def capability_error(current: enums.Capability) -> Exception:
```
1st1, line 46: Make it a method on the Capability enum

1st1 (commented):
So far so good.

1st1:
> remove advisory_lock* functions (I've kept until we figure out session mode)

If the functions are used in the test suite let's make the test-only.

> figure out what to do with session mode

I haven't seen any significant complication in the code. What are your latest thoughts on this?

edb/api/errors.txt
```diff
@@ -39,6 +39,10 @@
 
 0x_03_03_00_00   ResultCardinalityMismatchError
 
+0x_03_04_00_00   UnsupportedCapabilityError
```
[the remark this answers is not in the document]
tailhook: Added. But I'm not very sure if this distinction is useful at all.

edb/server/http_edgeql_port/protocol.pyx
```diff
@@ -31,10 +31,31 @@ from edb.common import markup
...
 
 
+ALLOWED_CAPABILITIES = (
+    enums.Capability.QUERY |
+    enums.Capability.MODIFICATIONS
+)
+
+
+def capability_error(current: enums.Capability) -> Exception:
+    for item in enums.Capability:
```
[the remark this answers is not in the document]
tailhook: This isn't a problem but yeah, fixed the enum.

edb/server/compiler/enums.py
```diff
@@ -46,14 +46,32 @@ class ResultCardinality(strenum.StrEnum):
...
+    MODIFICATIONS     = 0b00001         # noqa
+    SESSION_CONFIG    = 0b00010         # noqa
+    TRANSACTION       = 0b00100         # noqa
+    DDL               = 0b01000         # noqa
+    PERSISTENT_CONFIG = 0b10000         # noqa
+
+    QUERY             = 0b00001 << 32   # noqa
+    SESSION_MODE      = 0b00010 << 32   # noqa
+
+    PUBLIC = (
```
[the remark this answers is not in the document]
tailhook: Added a comment. I think names are descriptive enough. And moved definition out the class.

edb/server/compiler/enums.py
```diff
@@ -46,14 +46,32 @@ class ResultCardinality(strenum.StrEnum):
...
-    QUERY = enum.auto()
-
-    ALL = DDL | TRANSACTION | SESSION | QUERY
+class Capability(enum.IntFlag):
+
+    MODIFICATIONS     = 0b00001         # noqa
+    SESSION_CONFIG    = 0b00010         # noqa
+    TRANSACTION       = 0b00100         # noqa
+    DDL               = 0b01000         # noqa
+    PERSISTENT_CONFIG = 0b10000         # noqa
```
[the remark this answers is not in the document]
tailhook: Fixed

edb/server/mng_port/edgecon.pyx
```diff
@@ -100,6 +99,33 @@ cdef object log_metrics = logging.getLogger('edb.server.metrics')
...
+DEF ALL_CAPABILITIES = 0xFFFFFFFFFFFFFFFF
+
+cdef uint64_t ALWAYS_ALLOWED = (
+    enums.Capability.QUERY |
+    enums.Capability.SESSION_MODE
+)
+cdef uint64_t PUBLIC_CAPABILITIES = enums.Capability.PUBLIC
+
+
+def capability_error(
```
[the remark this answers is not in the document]
tailhook: I'd rather make it a method on error object, but we can't add methods on error object :( Moved to enum.

edb/server/mng_port/edgecon.pyx
tailhook: Thanks. Fixed!

edb/server/mng_port/edgecon.pyx
tailhook: Done. Thanks!

tailhook:
> > remove advisory_lock* functions (I've kept until we figure out session mode)
> 
> If the functions are used in the test suite let's make the test-only.
> 

They are in the test suite of python bindings. Does python package use testmode server? I think so. So yes we can do.

> > figure out what to do with session mode
> 
> I haven't seen any significant complication in the code. What are your latest thoughts on this?

There isn't any much complication in the code per se (except they are broken now). But my thoughts are the same:
1. We can't make reliable `advisory_lock` after connnection pooling, keeping them in testmode and using insde the transaction is fine
2. Sleep also should be only available in test mode (and `session_only` flag on it makes no sense) 
3. So there is no justification either for `session_only` flag or for `SESSION_MODE` (internal) capability. While latter is an internal thing, `session_only` is a public flag for functions.

tailhook:
Minimal fix for session_mode functions is something along the lines of: https://github.com/edgedb/edgedb/pull/2005/commits/cecffa53a17e02a959bc2caabdc1ce0e89ec5b8b
But I still think that `session_only` flag must be removed or at least disallowed for user functions.

1st1:
>  But I still think that session_only flag must be removed or at least disallowed for user functions.

Let's make sleep/lock functions testmode-only. You can drop the session_only flag.

edb/server/mng_port/edgecon.pyx
```diff
@@ -1120,7 +1177,14 @@ cdef class EdgeConnection:
...
 
         buf = WriteBuffer.new_message(b'1')  # ParseComplete
-        buf.write_int16(0)  # no headers
+
+        buf.write_int16(1)
+        buf.write_int16(SERVER_HEADER_CAPABILITIES)
+        buf.write_int32(sizeof(uint64_t))
+        buf.write_int64(<int64_t>(
+            PUBLIC_CAPABILITIES & compiled_query.query_unit.capabilities
+        ))
```
1st1, line 1186: @elprans I'm gonna merge this now, but we need to have a quick discussion about common header serialization format. Did you want to use type descriptors for that?

elprans: >  Did you want to use type descriptors for that?

I think that was one of the options, but I don't recall if there was a more detailed plan of action.  For example, would we transmit the type descriptor for each header, or would that be hardcoded?

In any case, I think the current string-based form with arbitrary rules and ad-hoc validation is quite bad.

1st1 (commented):
Let's remove the session_only flag and merge this.

1st1:
Also don't forget to cleanup the docs for sleep/locks since they're becoming testmode only.

1st1:
And let's prefix sleep/lock functions with `_` to make it clear they aren't public.

tailhook:
Added underscores and updated docs (I've removed testmode functions, is it what I needed to do with them?). We could also probably add a namespace for testmode functions instead of throwing them in arbitrary modules with underscores, probably.

tailhook:
So what we have to do to get this merged:
1. Merge python: https://github.com/edgedb/edgedb-python/pull/132 (test will fail)
2. Release a python library (tests will fail, perhaps `skip wheel tests` label should help)
3. Add the new version into `setup.py` in this PR and merge

tailhook:
Last commit removes `QUERY` capability and so the whole public/private capability split. The query capability has never been checked and was always allowed anyway. And was kept so long just because it already was there (and there was `SESSION_MODE` private capability just commit before). We can revert last commit if you wish.

1st1:
I merged the PR; that said I have a question re

> Release a python library (tests will fail, perhaps skip wheel tests label should help)

Why would tests fail? Why can't you skip them if they're run against older edgedb?

tailhook:
> Why would tests fail? Why can't you skip them if they're run against older edgedb?

Sounds plausible. Do we have such a mechanism already?

elprans (approved):
LGTM. Good work!

PR #2008, opened 2020-12-01
Fix some DDL/migration bugs relating to aliases

msullivan:
* Don't generate a 'DROP SCALAR TYPE' when an alias is ALTERed away
   from a scalar.
 * Fix several different things that could go wrong when ALTERing aliases,
   especially with regard to generating DeleteObjects for the old underlying
   type.

All of the test_schema_migrations_equivalence_collections_ tests now pass

Work on #1987.

### DIFF
--- a/tests/test_edgeql_data_migration.py
+++ b/tests/test_edgeql_data_migration.py
@@ -6865,12 +6865,6 @@ class TestEdgeQLDataMigration(tb.DDLTestCase):
             [{'a': 'hello', 'b': 42}],
         )
 
-    @test.xfail('''
-        The "complete" flag is not set even though the DDL from
-        "proposed" list is used.
-
-        This happens on the second migration.
-    ''')
     async def test_edgeql_migration_eq_collections_06(self):
         await self.migrate(r"""
             type Base {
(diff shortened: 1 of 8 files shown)


tests/test_edgeql_data_migration.py
```diff
@@ -6903,10 +6897,11 @@ async def test_edgeql_migration_eq_collections_06(self):
         )
 
     @test.xfail('''
-        The "complete" flag is not set even though the DDL from
-        "proposed" list is used.
+        ISE: column <blah> cannot be cast automatically to type
```
elprans, line 6900: This type of tests needs to be converted to use the `DESCRIBE CURRENT MIGRATION` mechanism, since altering a type requires user input.

elprans: Clarification: no need to do the conversion in this PR.

elprans (approved):

tests/test_edgeql_data_migration.py
msullivan: What does "requires" mean exactly here? We do all sorts of wild stuff in this file without describe current migration, and using POPULATE MIGRATION; COMMIT MIGRATION; works for most cases too?

This version ought to produce a better error than an ISE, too, right?

elprans: > This version ought to produce a better error than an ISE, too, right?

Yes. `POPULATE MIGRATION` should abort if user input is required to properly generate a migration.  What you're seeing currently is a Postgres equivalent of such an error.  I'm working on this bit  right now.

msullivan: Even `migrate` here kind of simulates user input, though, right? It repeatedly selects the proposed DDL.

elprans: > kind of simulates user input, though, right?

It simulates agreement with proposed DDL, but here there is no way to even generate any propositions, since we don't know how to convert from one data to another if there's no defined cast.  Similarly, when a `required` property is added, we must know the value to use to back-populate existing entries.

Issue #2010, opened 2020-12-01
Compiler bug not allowing functions to return sets of tuples


Dhghomon:
- EdgeDB Version: Alpha 6
- OS Version: Windows 10 (Ubuntu Windows 10 subsystem)

I discovered this when experimenting with creating my own functions that return sets.

Steps to Reproduce:

Create any function of this sort:

```
create function find_string(input: str) -> SET OF tuple<str, int64>
  using(
    SELECT(Person.name, find(Person.name, input))
);
```

Here's the output: `ERROR: InternalServerError: a column definition list is only allowed for functions returning "record".`

@elprans mentioned the following on the error:

```
It's a compiler bug

It compiles the body query as if it was an anonymous tuple, but it isn't in this context and so Postgres complains.  SQL arcana.
```

I think it was during Chapter 12 when the schema looked [like this](https://github.com/edgedb/easy-edgedb/blob/master/chapter12/code.md).

PR #2011, opened 2020-12-01
Fix functions returning a set of tuples

elprans:
A tuple type returned from a function would be declared as a named
composite type in Postgres and it is an error to specify a column
definition list for such type, even if it matches the definition.
There are checks for this elsewhere, but set-returning function didn't
handle this properly.

Fixes: #2010

### DIFF
--- a/edb/pgsql/compiler/relgen.py
+++ b/edb/pgsql/compiler/relgen.py
@@ -1929,10 +1929,13 @@ def _process_set_func(
         colnames = list(subtypes)
     else:
         colnames = [ctx.env.aliases.get('v')]
-        coldeflist = []
 
-    if expr.sql_func_has_out_params:
-        # SQL functions declared with OUT params reject column definitions.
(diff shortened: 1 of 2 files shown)


edb/pgsql/compiler/relgen.py
```diff
@@ -1929,10 +1929,13 @@ def _process_set_func(
         colnames = list(subtypes)
     else:
         colnames = [ctx.env.aliases.get('v')]
-        coldeflist = []
 
-    if expr.sql_func_has_out_params:
-        # SQL functions declared with OUT params reject column definitions.
+    if (
+        # SQL functions declared with OUT params or returning
+        # name composite types reject column definitions.
```
msullivan, line 1935:
```suggestion
        # named composite types reject column definitions.
```

msullivan (approved):

1st1 (approved):

Issue #2006, opened 2020-11-30
Run edgedb on digitalocean's cloud storage


tailhook:
Currently it fails because can't run on non-superuser account. It's unclear whether `doadmin` default user can do all the things we want from it. Is there a simple way to test that? Can we that on bootstrap when using `--postgres-dsn` instead of failing on non-superuser?

elprans:
EdgeDB requires certain privileges that are, by default, only granted to superusers.  Most importantly, we need to be able to define operators and operator classes.  On RDS, although the full superuser is not accessible, `rds_superuser` includes all privileges we require.  If DO provides a similar role, we can use that, if not, then it would be fairly hard to make EdgeDB run there.

Here's one relevant thread: https://www.digitalocean.com/community/questions/managed-postgres-clusters-pg_restore-errors-for-superuser-privileges

elprans:
OK, so the good news is that I managed to get bootstrap run successfully on DO.  The bad news is that DO requires TLS and we don't support that yet for backend connections.

haikyuu:
Azure added support for postgres 12 https://docs.microsoft.com/en-us/azure/postgresql/flexible-server/concepts-supported-versions

PR #2013, opened 2020-12-02
Refactor remote Postgres support to be more generic and robust

elprans:
The current support for managed Postgres instances via `--postgres-dsn`
is limited to RDS and is rather clunky.  In order to support more
managed hosts and with less specific code, implement a notion of
"backend parameters", which include things like the cluster superuser
role and a capability bitmask that informs EdgeDB of the availability
(or lack of) certain features.

With this, EdgeDB can be bootstrapped on a DO instance and I plan to
test more providers soon.

Issue: #2006

### DIFF
--- a/edb/server/pgcluster.py
+++ b/edb/server/pgcluster.py
@@ -20,6 +20,7 @@ from __future__ import annotations
 from typing import *
 
 import asyncio
+import enum
 import locale
 import logging
 import os
@@ -34,6 +35,8 @@ import time
 
 import asyncpg
 from asyncpg import serverversion
+
+from edb.common import uuidgen
 
 from edb.server import buildmeta
 from edb.server import defines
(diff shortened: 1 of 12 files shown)


edb/server/pgcluster.py
```diff
@@ -71,14 +74,62 @@ class ClusterError(Exception):
...
+    NONE = 0
+    #: Whether CREATE ROLE .. SUPERUSER is allowed
+    SUPERUSER_ACCESS = 1 << 0
+    #: Whether reading PostgreSQL configuration files
+    #: via pg_file_settings is allowed
+    CONFIGFILE_ACCESS = 1 << 1
+    #: Whether the PostgreSQL server supports the C.UTF-8 locale
+    C_UTF8_LOCALE = 1 << 2
+
+    ALL = SUPERUSER_ACCESS | CONFIGFILE_ACCESS | C_UTF8_LOCALE
```
1st1, line 88: I suggest to not include computed flags into the enum. Including it makes it a proper member, making it show up when you iterate over individual flags etc. It'd just rename to `ALL_BACKEND_CAPABILITIES`

edb/server/pgcluster.py
```diff
@@ -767,24 +826,95 @@ def get_remote_pg_cluster(dsn: str) -> RemoteCluster:
...
+            configfile_access = True
+        else:
+            configfile_access = True
+
+        if configfile_access:
+            caps |= BackendCapabilities.CONFIGFILE_ACCESS
+
+        tx = conn.transaction()
+        await tx.start()
+        rname = ''.join(random.choice(string.ascii_letters) for _ in range(10))
```
1st1, line 872: I'd use a UUID here. It's highly unlikely but a collision is possible in which case you'd get an exception you don't currently catch.

1st1 (approved):

tailhook (approved):
Looks great! Can non-having no `c.utf-8` locale be a problem for something? (i.e. sort order for text fields will be different?)

elprans:
> Can having no `c.utf-8` locale be a problem for something?

Actually, I don't think I fully intended to switch collation to `c.utf-8` here.  Re-reading some earlier commits, I think we should keep it as `c` everywhere for now.

Issue #2001, opened 2020-11-26
Cardinality inference for computable properties


tailhook:
- EdgeDB Version: 1-alpha6

Could not create a required computable property:
```
required property deleted := EXISTS(.<element[IS DeletionRecord]);
```
The error is:
```
possibly an empty set returned by an expression for the computable 'deleted' declared as 'required'
```
Since `EXISTS` never returns empty set I expect this to work.

PR #2015, opened 2020-12-03
Fix cardinality inference on operators

elprans:
The operator cardinality inference currently doesn't take into account
the optionality of the return type, which results in incorrect lower
cardinality bound for most operators.  Function inference handles this
correctly.

Fixes: #2001

### DIFF
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -5561,7 +5561,7 @@ class TestDescribe(tb.BaseSchemaLoadTest):
                 CREATE OPTIONAL MULTI LINK intersection_of
                     -> schema::ObjectType;
                 CREATE OPTIONAL MULTI LINK union_of -> schema::ObjectType;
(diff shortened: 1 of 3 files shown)


1st1 (approved):

tests/test_schema.py
```diff
@@ -5561,7 +5561,7 @@ def test_schema_describe_builtins_01(self):
                 CREATE OPTIONAL MULTI LINK intersection_of
                     -> schema::ObjectType;
                 CREATE OPTIONAL MULTI LINK union_of -> schema::ObjectType;
-                CREATE OPTIONAL SINGLE PROPERTY is_compound_type := (
+                CREATE REQUIRED SINGLE PROPERTY is_compound_type := (
```
1st1, line 5564: Is this related to the change?

msullivan (approved):

tailhook (approved):

PR #2016, opened 2020-12-03
Fix a number of cases where renames are combined with other operations

msullivan:
This requires tracking renames as we go when doing _cmd_tree_from_ast
and write_meta, so that we can correctly identify the enclosing object
name even when it has changed.

Also, make doing renames work even if it is not the first thing in the
ALTER. This requires abandoning apply fragments in alter_begin (which
weirdly required a fixup to scalar base validation, but caused no
other trouble).

(I had thought that fixing that was necessarily tangled up with the
rest; I think I might have been wrong, but it doesn't seem worth
untangling and some of the main changes are necessary for pending PRs
also.)

Progress on #1987. Fixes:
 * test_schema_migrations_equivalence_27
 * test_schema_migrations_equivalence_rename_refs_05
 * test_schema_migrations_equivalence_rename_refs_06


elprans (approved):
:+1:

Issue #2004, opened 2020-11-30
Combine INLINE_TYPEIDS and INLINE_TYPENAMES headers


tailhook:
This could be a single header `IMPLICIT_FIELDS` which is a bitmask (i don't know why they are called `INLINE`, but names aren't too important, as we use numbers in the protocol).

There was also discussion of whether `IMPLICIT_ID` should be added to the bitmask, but I don't think we have a decision on that yet.

Or if we don't want bitmasks, we should make better validation of the headers, because anything like `TRUE` is `false` isn't very great idea. Any non-`true/false` value should be an error, instead of current:
```
                elif k == QUERY_OPT_INLINE_TYPEIDS:
                    inline_typeids = v.lower() == b'true'
                elif k == QUERY_OPT_INLINE_TYPENAMES:
                    inline_typenames = v.lower() == b'true'
```

tailhook:
This also should be gated on protocol version 0.9. Because currently it works on python bindings which have protocol 0.8 and doesn't force upgrade. This can be problematic in some scenarios.

elprans:
> This could be a single header `IMPLICIT_FIELDS` which is a bitmask

I agree that "inline" is a misnomer here.  I'm not sure about combining into a bitmask, though.  We already have one non-flag query option: `implicit_limit`.  IMO, it makes more sense to keep options as separate headers.

> Or if we don't want bitmasks, we should make better validation of the headers.

@1st1 did we decide to make headers binary and use the normal I/O codecs for them?

1st1:
> @1st1 did we decide to make headers binary and use the normal I/O codecs for them?

I remember that you and I discussed that and both agreed that it's better than ad-hoc custom ascii values for every other header. So +1 from me.

PR #2019, opened 2020-12-03
Assume inline_typeids for protocol <= 0.8

tailhook:
Was broken in d7b08f7b281144a20f9ca0c559a457037fa38b6e
In particular, JavaScript tests are broken because of this.


tailhook:
We also need to gate new headers, but I left this to #2004 to have less conflicts with #2005

1st1 (approved):

PR #2020, opened 2020-12-03
Fix injection of implicit limit in some cases

elprans:
Currently, implicit limit is injected into `SET OF` arguments of all
generic functions (with the original intent to cover `array_agg`-like
functions).  Unfortunately, this breaks `std::max` and the like, and
while we could add an extra condition to check for the array return
type, it seems like the best thing to do is to hardcode for `array_agg`
for now.  A better implementation might be to slice all output arrays,
but that seems like more effort than its worth.

Additionally, stop implicit injection if inside a `json` cast.  JSON
strings, unlike structured output, have no indication of truncation, and
so the result would appear wrong.  A future improvement might add JSON
string truncation and some magic tokens to indicate truncation.

`std::max` bug reported by @ambv.

### DIFF
--- a/edb/edgeql/compiler/stmt.py
+++ b/edb/edgeql/compiler/stmt.py
@@ -75,12 +75,13 @@ def compile_SelectQuery(
             sctx.partial_path_prefix = ctx.partial_path_prefix
             stmt.implicit_wrapper = True
 
-        if expr.limit is not None:
-            sctx.inhibit_implicit_limit = True
-        elif ((ctx.expr_exposed or sctx.stmt is ctx.toplevel_stmt)
-                and ctx.implicit_limit and not sctx.inhibit_implicit_limit):
+        if (
+            (ctx.expr_exposed or sctx.stmt is ctx.toplevel_stmt)
+            and ctx.implicit_limit
+            and expr.limit is None
+            and not ctx.inhibit_implicit_limit
+        ):
             expr.limit = qlast.IntegerConstant(value=str(ctx.implicit_limit))
-            sctx.inhibit_implicit_limit = True
 
         stmt.result = compile_result_clause(
             expr.result,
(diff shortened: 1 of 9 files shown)


edb/edgeql/compiler/stmt.py
```diff
@@ -1022,8 +1023,28 @@ def compile_result_clause(
...
+            rexpr = astutils.ensure_ql_select(result_expr)
+            if (
+                sctx.implicit_limit
+                and rexpr.limit is None
+                and not sctx.inhibit_implicit_limit
+            ):
+                # Inline alias is special: it's both "exposed",
+                # but also subject for further processing, so
+                # make sure we don't mangle it with an implicit
+                # limit.
```
msullivan, line 1035: Hm, so does this not get limited at all, then?

edb/edgeql/compiler/stmt.py
```diff
@@ -75,12 +75,13 @@ def compile_SelectQuery(
...
-        if expr.limit is not None:
-            sctx.inhibit_implicit_limit = True
-        elif ((ctx.expr_exposed or sctx.stmt is ctx.toplevel_stmt)
-                and ctx.implicit_limit and not sctx.inhibit_implicit_limit):
+        if (
+            (ctx.expr_exposed or sctx.stmt is ctx.toplevel_stmt)
+            and ctx.implicit_limit
+            and expr.limit is None
+            and not ctx.inhibit_implicit_limit
+        ):
```
msullivan, line 83: (What's the change happening here?)

msullivan (approved):
Seems reasonable. A couple questions mostly for my own edification.

edb/edgeql/compiler/stmt.py
elprans: It does, by the outer query.

```
SELECT 
   _ := (foo)  # <- implicit nested SELECT here, don't limit
LIMIT
   <implicit>
```

edb/edgeql/compiler/stmt.py
elprans: The previous coding attempted to solve this case:

```
SELECT
   (SELECT ... LIMIT <implicit>) # nested
LIMIT
   100 # more than implicit
```

But this is stupid and unnecessary, because the CLI would trim excessive entries anyway, and so the new approach doesn't care if your explicit `LIMIT` is greater than implicit, it'll clamp to the least of the two.

PR #2021, opened 2020-12-03
Avoid needlessly altering cardinality in some cases

msullivan:
The logic for determining whether to output a property alter had (I
believe) a precedence bug, which resulted in outputting `SET SINGLE`
and `SET MULTI` in some migrations where they didn't change.
Clean that up.

### DIFF
--- a/tests/test_edgeql_data_migration.py
+++ b/tests/test_edgeql_data_migration.py
@@ -20,6 +20,7 @@
 import json
 import os.path
 import uuid
+import textwrap
 
 import edgedb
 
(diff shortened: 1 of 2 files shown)


tests/test_edgeql_data_migration.py
```diff
@@ -2860,6 +2867,43 @@ async def test_edgeql_migration_eq_14(self):
...
+            }
+
+            type Derived extending Base {
+                overloaded required property foo -> str;
+            }
+        """, populate=True)
+
+        await self.assert_describe_migration({
+            'confirmed': [
+                textwrap.dedent("""\
```
elprans, line 2893: I wonder if we should maybe teach `assert_describe_migration` to dedent stuff automatically.

elprans (approved):

tests/test_edgeql_data_migration.py
msullivan: Yeah, that's a good idea. I went and did it.

Issue #12, opened 2017-07-08
Need a tool for testing the examples from the docs efficiently.


vpetrovykh:
Typos and outdated syntax can creep into documentation examples. We need a tool to test them together with the rest of the test suite. In particular that means that we will need to make the examples use module names in a self-consistent and sane manner.

Issue #11, opened 2017-07-07
Add function-like signatures to constraints


1st1:
While working with @vpetrovykh on #8, we realized that `$param` in constraints isn't strictly defined, which causes problems when you try to use function calls in your constraints' exprs (in fact, everything seems to be working by accident right now).

So when we changed

```
CREATE CONSTRAINT std::enum INHERITING std::constraint {
    SET errmessage := '{subject} must be one of: {param}.';
    SET expr := (subject IN $param);
};
```

to

```
CREATE CONSTRAINT std::enum INHERITING std::constraint {
    SET errmessage := '{subject} must be one of: {param}.';
    SET expr := (array_contains($param, subject));
};
```

the compiler could no longer compile the `expr` expression, as the type of `$param` is unknown.

We propose to add function-like signatures to constraints, as in:

```
CREATE CONSTRAINT std::enum($param: array<any>) INHERITING std::constraint {
    SET errmessage := '{subject} must be one of: {param}.';
    SET expr := (array_contains($param, subject));
};
```

which would at least make inference possible for parameters of constraints.

There's still an open issue of how to determine the type of the `subject` link though.

Thoughts?

PR #2023, opened 2020-12-04
Reduce verbosity/duplication in ordering._trace_op

msullivan:
There's a 4-7 line idiom in _trace_op that for looking up a key and
creating a new entry if necessary that is repeated ten times. Refactor
it into a nested function.

I decided to refactor it because otherwise I was about to add
repetitions #11 and #12 :P.


elprans (approved):
Oh, thank you! This has been an eye sore

PR #2026, opened 2020-12-04
Fix dropping an overloaded pointer in both a parent and child class

msullivan:
This ran into a fiddly issue where the DROP OWNED was nested inside of
redundant AlterPropertys as an artifact of how opbranches are
constructed.

Progress on #1987: Fixes:
 * test_schema_migrations_equivalence_03
 * test_schema_migrations_equivalence_04

### DIFF
--- a/edb/schema/ordering.py
+++ b/edb/schema/ordering.py
@@ -776,18 +776,8 @@ def _extract_op(stack: Sequence[sd.Command]) -> List[sd.Command]:
         parent_op = alter_delta
         new_stack.append(parent_op)
 
-    op = stack[-1]
-    # We don't want duplicate nested Alters on the same object (which
-    # can happen when an operation is transformed from a different
-    # type), so prune it away as necessary.
-    if (
-        isinstance(op, sd.AlterObject) and
-        isinstance(parent_op, sd.AlterObject) and
-        op.classname == parent_op.classname
-    ):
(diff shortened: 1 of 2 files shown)


edb/schema/ordering.py
```diff
@@ -776,8 +776,18 @@ def _extract_op(stack: Sequence[sd.Command]) -> List[sd.Command]:
         parent_op = alter_delta
         new_stack.append(parent_op)
 
-    stack[-2].discard(stack[-1])
-    parent_op.add(stack[-1])
-    new_stack.append(stack[-1])
+    op = stack[-1]
+    # We don't want duplicate nested Alters on the same object (which
```
elprans, line 780: Hm, how hard would it be to prevent this artifact in the first place?  ISTM we might be throwing away data by pruning it like this.

msullivan: So the situation where it comes up here was caused by `150b143d09dab20c31876a7636d31b091ba7ee5c`, which moved the AlterProperty/AlterOwned for a property underneath its delete, which results in a doublly nested AlterProperty once `_extract_op` messes around with things.

I think we can just directly attach the AlterOwned to the Delete, though, without an intermediate Alter, which will neatly solve the problem.

elprans (approved):
Sweet

PR #2029, opened 2020-12-04
Fix dropping inherited properties with constraints

msullivan:
The main bug is that inherited constraints would end up using the
subjectexpr of the abstract constraint in computing their name quals,
if the concrete parent had no subjectexpr. This resulted in it having
an incorrect name, causing it to be missed in places.

Additionally, the DDL text we generated contained subcommands nested
in a DropProperty. Prevent that, since it isn't valid in the concrete
syntax.

I think that changing how constraint names are computed means I had to
bump the version?

Progress on #1987. Fixes test_schema_migrations_equivalence_31.

### DIFF
--- a/edb/schema/delta.py
+++ b/edb/schema/delta.py
@@ -1708,22 +1708,29 @@ class ObjectCommand(
 
                     setattr(node, ast_attr, attr_val)
 
-            # No subcommands for delete are permitted
-            if not isinstance(self, DeleteObject):
-                # Keep subcommands from refdicts and alter fragments (like
-                # rename, rebase) in order when producing DDL asts
-                refdicts = tuple(x.ref_cls for x in mcls.get_refdicts())
-                for op in self.get_subcommands():
-                    if (
-                        isinstance(op, AlterObjectFragment)
-                        or (isinstance(op, ObjectCommand) and
-                            issubclass(op.get_schema_metaclass(), refdicts))
-                    ):
-                        self._append_subcmd_ast(schema, node, op, context)
+            # Keep subcommands from refdicts and alter fragments (like
+            # rename, rebase) in order when producing DDL asts
+            refdicts = tuple(x.ref_cls for x in mcls.get_refdicts())
+            for op in self.get_subcommands():
+                if (
+                    isinstance(op, AlterObjectFragment)
+                    or (isinstance(op, ObjectCommand) and
+                        issubclass(op.get_schema_metaclass(), refdicts))
+                ):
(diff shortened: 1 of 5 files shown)


edb/schema/delta.py
```diff
@@ -1708,16 +1708,18 @@ def _apply_fields_ast(
...
-            refdicts = tuple(x.ref_cls for x in mcls.get_refdicts())
-            for op in self.get_subcommands():
-                if (
-                    isinstance(op, AlterObjectFragment)
-                    or (isinstance(op, ObjectCommand)
-                        and issubclass(op.get_schema_metaclass(), refdicts))
-                ):
-                    self._append_subcmd_ast(schema, node, op, context)
+            # No subcommands for delete are permitted
+            if not isinstance(self, DeleteObject):
```
elprans, line 1712: Hm, I wonder if we're not papering (future?) bugs with this.  If the delta tree contains DDL-renderable commands under `Delete` then either of these should be true: a) the commands shouldn't have been attached to `Delete` in the first place; or b) the subcommands should be incidental to `Delete` and so non-renderable themselves.

msullivan: The subcommands here will be `Delete`s of referenced objects, which are definitely renderable themselves. Currently we generate DDL for those subcommands basically all the time. For example, if we drop an object with a property, we'll get
```
DROP TYPE default::Test {
    DROP PROPERTY foo;
};
```

Looking into it a bit more, it seems that many sorts of objects have an undocumented feature to allow some subcommands for `DROP` in the DDL, but Property is missing it. Lots of the objects that do have it are missing parts, too (for example, DROP ANNOTATION isn't allowed as a DROP subcommand anywhere)

So I can think of two ways forward here:
 1. This patch's approach, where we skip generating subcommands. (Possibly with some extra asserts added to prevent papering over issues. Maybe we could generate the subcommands, verify they are all Drops, and then clear them?). Probably then also remove the undocumented DROP subcommand feature from DDL.
 2. Make the DROP subcommand feature complete. Support it for every sort of object and every sort of subobject.

I prefer option 1, because I think that DROPing all the subcomponents explicitly in the generated migration DDL is pointlessly noisy.

elprans: 1 sounds good to me.  Nested deletes are basically the result of `_canonicalize`, so we just need to verify that there are only subcommands that would have been generated by it, which at a first glance seem to be any `Delete` subcommands.

msullivan: Alright, I updated it to do some more checking

elprans (approved):

PR #2030, opened 2020-12-05
Make DDL migrations the default

### LINKED ISSUE #1987: Move to always using MIGRATIONS_VIA_DDL

msullivan:
Remove the `migrations_via_ddl` flag and add a `legacy_migrations`
flag with an inverted meaning.

We can tear out the old code in a follow-up.

Three tests needed to be xfailed:
 * Two tests that changed the body and return type of a function that
   was used in an expression. These tests passed as
   assert_migration_equivalence tests but the migrations would fail
   with ISE from pg if you *actually* tried to execute them, so this
   is not a real regression.
 * test_schema_migrations_equivalence_23, which converts a type into
   an alias. We are probably going to need to add syntax to DDL to do
   this sort of modification.
   This migration basically did actually work when I tried it,
   though it didn't delete the underlying pgsql table and I suspect
   more weirdness also still lurked.

Close #1987.

### DIFF
--- a/edb/common/debug.py
+++ b/edb/common/debug.py
@@ -126,9 +126,6 @@ class flags(metaclass=FlagsMeta):
     log_metrics = Flag(
         doc="Log verbose statistics on connections and compiler behavior.")
 
-    legacy_migrations = Flag(
-        doc="Use the legacy non-DDL mode when running migrations.")
-
     disable_docs_edgeql_validation = Flag(
         doc="Disable validation of edgeql in docs (for site build)")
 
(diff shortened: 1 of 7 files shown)


elprans (approved):
Is there value in keeping `legacy_migrations`?

msullivan:
> Is there value in keeping `legacy_migrations`?

I dunno, probably not. I figured I'd do a quick flag-flipping switch to get the right tests running by default before I start tearing code out. (Though it looks like the tearing code out might not go very deep anyway, so maybe it wasn't worth splitting out.)

msullivan:
I changed my mind and decided to actually just delete it all.

elprans (approved):

elprans:
> I changed my mind and decided to actually just delete it all.

:+1:.  Don't forget to amend the commit message :-)

PR #2033, opened 2020-12-07
Add installation instructions for CLI

tailhook:
Only `curl | sh` added so far. I think eventually we need to add repository instructions too, but I don't know the good format of that. As we don't have them for server too.


elprans (approved):
LGTM.

PR #2038, opened 2020-12-09
Fix DISTINCT clobbering link props inside it.

vpetrovykh:
Currently this solves the problem for some cases, but not others.

### DIFF
--- a/edb/edgeql/compiler/stmt.py
+++ b/edb/edgeql/compiler/stmt.py
@@ -75,6 +75,16 @@ def compile_SelectQuery(
             sctx.partial_path_prefix = ctx.partial_path_prefix
             stmt.implicit_wrapper = True
 
+            # We need to preserve view_rptr if this SELECT is just
+            # an implicit wrapping of a single DISTINCT, because otherwise
+            # using a DISTINCT to satisfy link multiplicity requirement
+            # will kill the link properties.
+            #
+            # This includes problems with initializing the schema itself.
+            if (isinstance(expr.result, qlast.UnaryOp) and
+                    expr.result.op == 'DISTINCT'):
+                sctx.view_rptr = ctx.view_rptr
+
         if expr.limit is not None:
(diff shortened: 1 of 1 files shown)


vpetrovykh:
Here's the setup that breaks:
```
CREATE TYPE Child {CREATE REQUIRED PROPERTY val -> str};
CREATE TYPE Parent {CREATE MULTI LINK children -> Child {CREATE PROPERTY lp -> int64}};
INSERT Child {val := 'q'};
INSERT Child {val := 'w'};

# This breaks:
INSERT Parent {
    # if you remove DISTINCT from here, it works, but DISTINCT 
    # messes up cardinality inference for v.0 of all things.
    children := DISTINCT (
        FOR v IN {enumerate({'q', 'w'})}
        UNION (
            SELECT Child {
                @lp := v.0
            }
            FILTER .val = v.1
        )
    ),
};
```

edb/edgeql/compiler/stmt.py
```diff
@@ -75,6 +75,16 @@ def compile_SelectQuery(
...
             stmt.implicit_wrapper = True
 
+            # We need to preserve view_rptr if this SELECT is just
+            # an implicit wrapping of a single DISTINCT, because otherwise
+            # using a DISTINCT to satisfy link multiplicity requirement
+            # will kill the link properties.
+            #
+            # This includes problems with initializing the schema itself.
+            if (isinstance(expr.result, qlast.UnaryOp) and
+                    expr.result.op == 'DISTINCT'):
```
elprans, line 85: This is a general problem of nesting computable link properties. Hardcoding for `DISTINCT` specifically is not going to help much.  I have some considerations on how to refactor link property derivation in general, the idea is to attach them to the derived target type instead of trying to pass through the remote pointer trough context.

elprans (approved):

msullivan (approved):

PR #2039, opened 2020-12-10
Add alpha 7 changelog.

vpetrovykh:


docs/changelog/1_0_a7.rst
```diff
@@ -0,0 +1,86 @@
...
+
+This changelog summarizes new features and breaking changes in
+EdgeDB 1.0 alpha 7 "Lalande".
+
+
+EdgeQL
+======
+
+* Implement consistent update of expressions when referenced schema
+  objects are renamed (:eql:gh:`#1841`).
```
1st1, line 13: Overall this looks like you simply copy/pasted git log here. We need more editing, grouping of related commit messages, and some examples.

docs/changelog/1_0_a7.rst
```diff
@@ -0,0 +1,86 @@
...
+* Initial implementation of capabilities from `RFC 1004 <robust_>`_
+  (:eql:gh:`#2005`).
+* Move request methods into Executor interface (`#76
+  <https://github.com/edgedb/edgedb-js/pull/76>`_) as part of the `RFC
+  1004 <robust_>`_ changes.
+* Fix `edgedb-python <https://github.com/edgedb/edgedb-python>`_
+  driver forward compatibility with the new edgedb protocol versions.
+* Update the `edgedb-python <https://github.com/edgedb/edgedb-python>`_ driver
+  to 0.12.0a1.
+* Allow configuring __tid__ injected property via private APIs in
```
1st1, line 73: this shouldn't be mentioned

docs/changelog/1_0_a7.rst
```diff
@@ -0,0 +1,86 @@
...
+We now have an improved spec for client API (`RFC 1004 <robust_>`_).
+Rolling out the support for the full spec will be done in the next
+release, but some implementation work has already started.
+
+* Initial implementation of capabilities from `RFC 1004 <robust_>`_
+  (:eql:gh:`#2005`).
+* Move request methods into Executor interface (`#76
+  <https://github.com/edgedb/edgedb-js/pull/76>`_) as part of the `RFC
+  1004 <robust_>`_ changes.
+* Fix `edgedb-python <https://github.com/edgedb/edgedb-python>`_
```
1st1, line 69: remove this

docs/changelog/1_0_a7.rst
```diff
@@ -0,0 +1,86 @@
...
+  driver forward compatibility with the new edgedb protocol versions.
+* Update the `edgedb-python <https://github.com/edgedb/edgedb-python>`_ driver
```
1st1, line 71: remove this

docs/changelog/1_0_a7.rst
```diff
@@ -0,0 +1,86 @@
...
+  to 0.12.0a1.
+* Allow configuring __tid__ injected property via private APIs in
+  python driver (`#105 <https://github.com/edgedb/edgedb-python/pull/105>`_).
+* Allow passing ``implicit_typenames`` header in the python driver (`#105
```
1st1, line 75: remove this

docs/changelog/1_0_a7.rst
```diff
@@ -0,0 +1,86 @@
...
+
+
+Bindings
+========
+
+We now have an improved spec for client API (`RFC 1004 <robust_>`_).
+Rolling out the support for the full spec will be done in the next
+release, but some implementation work has already started.
+
+* Initial implementation of capabilities from `RFC 1004 <robust_>`_
```
1st1, line 64: remove this

docs/changelog/1_0_a7.rst
```diff
@@ -0,0 +1,86 @@
...
+  (:eql:gh:`#1996`).
+* Fix functions returning a set of tuples (:eql:gh:`#2010`).
+* Fix cardinality inference on operators (:eql:gh:`#2001`).
+* Implementation of more of the features of the `RFC 1000
+  <migrations_>`_ migration syntax, such as improving the proposed
+  migration DDL, adding migration hash, fixing issues with renames,
+  fixing issues with changing bases, improving expression dependency
+  tracing, improving handling changes to aliases, adding DDL commands
+  to better control fields like ``errmessage`` or ``default`` and
+  other changes (:eql:gh:`#1987`).
```
1st1, line 43: Most of items here are individual commit message. Most are related to migrations. I suggest grouping all migration related changes in its own section and elaborating on them.

docs/changelog/1_0_a7.rst
```diff
@@ -0,0 +1,86 @@
...
+  1004 <robust_>`_ changes.
+* Fix `edgedb-python <https://github.com/edgedb/edgedb-python>`_
+  driver forward compatibility with the new edgedb protocol versions.
+* Update the `edgedb-python <https://github.com/edgedb/edgedb-python>`_ driver
+  to 0.12.0a1.
+* Allow configuring __tid__ injected property via private APIs in
+  python driver (`#105 <https://github.com/edgedb/edgedb-python/pull/105>`_).
+* Allow passing ``implicit_typenames`` header in the python driver (`#105
+  <https://github.com/edgedb/edgedb-python/pull/105>`_).
+* Fix the python driver query cache to account for expected
```
1st1, line 77: remove this

1st1 (commented):
This needs more work.

docs/changelog/1_0_a7.rst
vpetrovykh: OK

docs/changelog/1_0_a7.rst
vpetrovykh: OK

docs/changelog/1_0_a7.rst
vpetrovykh: OK

docs/changelog/1_0_a7.rst
vpetrovykh: We normally reported driver version changes, why not now?

docs/changelog/1_0_a7.rst
vpetrovykh: I thought that forward compatibility was a noteworthy enough feature of any tool that interacts with other tools.

docs/changelog/1_0_a7.rst
vpetrovykh: I was thinking that much like we had an ongoing effort to implement RFC1000 mentioned in past changelogs (alpha 4, 5, 6), a similar mention is worth for starting implementaiton of RFC1004.

docs/changelog/1_0_a7.rst
vpetrovykh: OK, I'll make a separate section for migrations because it is the motivation behind a lot of the individual changes. Pure EdgeQL changes will shrink to things like `DATABASE` config and `bigint` vs. `decimal` clarification.

I'll add some kind of introductory paragraph to the migration section.

vpetrovykh:
I've added a few examples for some of the features and moved all migration stuff into its own section. Now migration also has an example + list of features that got improvements, so unlike in previous releases this one will look like the migrations are a lot more fleshed out.

I'm not sure about why should some of the items you marked for removal should be removed (I've left comments why I included them), but feel free to drop them of course.

I've also made a few tiny tweaks to the actual docs.

vpetrovykh:
Also, I ran into a tokenizer bug while I was at this (thanks to the doc snippets tests) - https://github.com/edgedb/edgedb/pull/2040

For the moment I just avoid the offending code, but that's weird how this remained unnoticed for so long. I think the constant extractor is typically masking this.

docs/changelog/1_0_a7.rst
```diff
@@ -0,0 +1,218 @@
...
+* Moving a type between modules (:eql:gh:`#1890`).
+* Changing base types and changing where constraints are defined
+  (:eql:gh:`#1996`).
+
+
+Command-Line Tools
+==================
+
+* Deprecate ``--default-database`` and ``--default-database-user``
+  (:eql:gh:`#1879`).
```
tailhook, line 189: This is not `Command-line Tools` change, it's server command-line.

docs/changelog/1_0_a7.rst
```diff
@@ -0,0 +1,218 @@
...
+* Add a new header to the ``Prepare`` message to implicitly inline
```
tailhook, line 190: Don't think we need this

tailhook (commented):
On the actual command-line tools:
1. Add ``edgedb server logs``
2. Add ``--connect-timeout``
3. Add ``--dsn``
4. Bugfix: `\lt` doesn't show cryptic internal types (unions and intersections)
5. Bugfix: added shutdown timeout of second (previously could hang indefinitely)
6. Default user is now `edgedb` and default database `edgedb`. Default database is not named as a user any more.
7. When installing server DEBIAN_FRONTEND is now `noninteractive` by default and is overridable
8. Bugfix: repl now supports full range of datetime values
9. Bugfix: ``server install`` now friendlier on linuxes without systemd allowing foreground run
10. Repl by default shows typenames, and effectively `\set introspect-types` setting is removed
11. Non-interactive error messages are prefixed by `edgedb error`, to quickly spot which tool has errored in scripts
12. Add `migration-log` command
13. Add `server info` command *(should merge today)*

docs/changelog/1_0_a7.rst
1st1: This is just a minor bug fix. We typically don't mention these in changelogs, and neither do other projects.

docs/changelog/1_0_a7.rst
1st1: Right, let's keep this, but change the version to 0.12.0.  We should never mention alpha versions in our chnagelogs.

docs/changelog/1_0_a7.rst
1st1: This has no effect on user APIs -- it's an internal change.

docs/changelog/1_0_a7.rst
vpetrovykh: Removed

docs/changelog/1_0_a7.rst
vpetrovykh: OK

docs/changelog/1_0_a7.rst
vpetrovykh: Removed

docs/changelog/1_0_a7.rst
vpetrovykh: Removed

docs/changelog/1_0_a7.rst
vpetrovykh: Removed

1st1 (approved):
Much better!

1st1:
@elprans want to take a look before we merge?

docs/changelog/1_0_a7.rst
1st1: @vpetrovykh Have you added what Paul suggested yet?

vpetrovykh:
One sec! I'm adding the CLI update from @tailhook

docs/changelog/1_0_a7.rst
vpetrovykh: Yes

docs/changelog/1_0_a7.rst
```diff
@@ -0,0 +1,242 @@
...
+
+We use :eql:stmt:`DESCRIBE CURRENT MIGRATION AS JSON <DESCRIBE CURRENT
+MIGRATION>` to see what EdgeDB is proposing. The JSON format makes it
+easier to potentially integrate this with other tools. For this
+example it's worth turning on ``json`` output mode for edgedb REPL:
+
+.. code-block:: edgeql-repl
+
+    db> \set output-mode json
+    db[tx]> DESCRIBE CURRENT MIGRATION AS JSON;
```
elprans, line 115: Why are we exhibiting raw `DESCRIBE` instead of the `create-migration` CLI?

vpetrovykh: Largely because I'm not as familiar with create-migration CLI (since I pretty much use `DESCRIBE` in tests), but also because it is smaller in terms of explanations needed for the example, like you don't need to touch on the schema directory and file organization in it to do an example. The full-blown `create-migration` CLI example might be good fodder for the blog? Either for the release or even a standalone one?

I guess this is a work-in-progress vs. finished tool consideration. For work-in-progress showing lower level guts can be more interesting because it shows capability (like JSON that can be integrated with a tool), whereas when the tool is finished you probably want to show off how nice and smooth it is without bothering with the internals. So although it's my personal experience bias that lead me to do `DESCRIBE` in REPL example, but I think it's a good insight into our work at this stage.

elprans: OK, fair enough.

> but I think it's a good insight into our work at this stage.

Please make this clear in the description.  `DESCRIBE CURRENT MIGRATION AS JSON` is not intended to be used directly.

vpetrovykh: I've added another paragraph mentioning the tool and that the example is showing low-level commands.

docs/changelog/1_0_a7.rst
```diff
@@ -0,0 +1,248 @@
...
+Server Command-Line
+===================
+
+* Make ``server install`` friendlier on linuxes without systemd
+  allowing foreground run (`#171
+  <https://github.com/edgedb/edgedb-cli/pull/171>`_).
+* When installing server ``DEBIAN_FRONTEND`` is now ``noninteractive`` by
+  default and is overridable (`#188
+  <https://github.com/edgedb/edgedb-cli/pull/188>`_).
+* Add ``edgedb server logs`` (`#172
```
tailhook, line 222: nit: either use `edgedb server xxx` everywhere or `server xxx` everywhere (I prefer former), currently it's inconsinstent.

docs/changelog/1_0_a7.rst
```diff
@@ -0,0 +1,248 @@
...
+* ``\lt`` in REPL doesn't show implicit internal types (unions and
+  intersections) (`#169 <https://github.com/edgedb/edgedb-cli/issues/169>`_).
+* Remove ``\set introspect-types`` in REPL and show typenames by
+  default.
+
+
+Server Command-Line
+===================
+
+* Make ``server install`` friendlier on linuxes without systemd
```
tailhook, line 216: I'm still not sure this fits `Server Command-Line`. I would say only `--default-database` and `--default-database-user` do.

docs/changelog/1_0_a7.rst
```diff
@@ -0,0 +1,248 @@
...
+  allowing foreground run (`#171
+  <https://github.com/edgedb/edgedb-cli/pull/171>`_).
+* When installing server ``DEBIAN_FRONTEND`` is now ``noninteractive`` by
+  default and is overridable (`#188
+  <https://github.com/edgedb/edgedb-cli/pull/188>`_).
+* Add ``edgedb server logs`` (`#172
+  <https://github.com/edgedb/edgedb-cli/pull/172>`_).
+* Add ``server info`` command.
+* Add shutdown timeout of 1 second.
```
tailhook, line 225: This is super-unclear. Better:
* When exiting from the REPL with Ctrl+D client now waits max of 1 second for server response

But generally, if you skipped all bugfixes anyway, this might go unnoticed too.

vpetrovykh: Perhaps I could just drop it from the highlights.

docs/changelog/1_0_a7.rst
vpetrovykh: I'm kinda going with the idea that tools/commands that interact with the server with the purpose of monitoring/installing/managing some aspect of the server are "server" tools. Whereas the tools focusing on data (so schema or querying) are "regular" CLI.

docs/changelog/1_0_a7.rst
vpetrovykh: Updated

elprans (approved):

PR #2040, opened 2020-12-10
Tokenizer bug for bigint.

vpetrovykh:
Rust tokenizer seems to fail on new bigint literals: `123e100n`.

### DIFF
--- a/tests/test_edgeql_syntax.py
+++ b/tests/test_edgeql_syntax.py
@@ -25,7 +25,6 @@ from edb import errors
 from edb.testbase import lang as tb
 from edb.edgeql import generate_source as edgeql_to_source
 from edb.edgeql.parser import parser as edgeql_parser
-from edb.tools import test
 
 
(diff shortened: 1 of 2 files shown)


tailhook:
Added a fix.

tests/test_edgeql_syntax.py
```diff
@@ -25,6 +25,7 @@
 from edb.testbase import lang as tb
 from edb.edgeql import generate_source as edgeql_to_source
 from edb.edgeql.parser import parser as edgeql_parser
+from edb.tools import test
```
1st1, line 28: I don't think this import is needed.

1st1 (approved):
@tailhook lgtm, please drop the weird import in the test file.

tests/test_edgeql_syntax.py
1st1, line 28:
```suggestion
```

PR #2046, opened 2020-12-11
Fix accessing links on objects that come from functions and the like

### LINKED ISSUE #1887: edb.errors.InternalServerError: there is no range var for (__derived__::expr~15) source in <pg.SelectStmt at 0x7f9393526e80>
```
SELECT
    ast::Expr
FILTER
    count (.value[IS ast::Call].args) = 2
    AND (
        WITH
            items := array_agg(
                (
                    FOR _singleton
                    IN {.value[IS ast::Call]}
                    UNION (SELECT _singleton.args ORDER BY @index)
                )
            )
        SELECT
            items[0] IS ast::Name
    )
LIMIT 10;
```

while playing with the edgeql, I noticed this throws an `edb.errors.InternalServerError`, so reporting it. Here is the [schema](https://github.com/reizio/reiz.io/blob/master/static/Python-reiz.edgeql)

msullivan:
semi_join would fail in some cases when the object came from somewhere
that didn't provide a source (like a function call or an array
indexing).

Also, drop the initial_value on array_agg, which isn't needed and
causes trouble when it can't be type inferred (such as in the bug
report for the main issue).

Closes #1887.

### DIFF
--- a/edb/lib/std/30-arrayfuncs.edgeql
+++ b/edb/lib/std/30-arrayfuncs.edgeql
@@ -26,6 +26,7 @@ std::array_agg(s: SET OF anytype) -> array<anytype>
     CREATE ANNOTATION std::description :=
         'Return the array made from all of the input set elements.';
     SET volatility := 'IMMUTABLE';
+    SET initial_value := [];
     USING SQL FUNCTION 'array_agg';
 };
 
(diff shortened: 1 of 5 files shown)


elprans:
> Also, drop the initial_value on array_agg, which isn't needed and causes trouble when it can't be type inferred (such as in the bug report for the main issue).

This doesn't seem right and causes a bunch of tests to fail.  The "cannot determine type of empty array" error actually comes from Postgres, not EdgeDB, because I suspect we're failing to generate the correct SQL cast for the initial value in this case.

msullivan:
Yeah, indeed. New version I think gets it right.

(My syncing setup had failed apparently, so I ran the test suite on the wrong code before putting this PR up :P)

elprans (approved):

PR #2048, opened 2020-12-12
Untangle the handling of cardinality in the schema

elprans:
There is currently a fair amount of brittle special-casing around the 
handling of `cardinality` and `required` in pointers, specifically around
the handling of inherited and computed status, and how those and other
fields are emitted in `DESCRIBE`.

Now, there's nothing fundamentally special in `cardinality` or
`required` that prevents them to be handled in a generic fashion, except 
that we need to keep track of whether the values were inferred or specified
explicitly.  For this we introduce the `computed_fields` array that is
substantially similar to the existing `inherited_fields` attribute, but
tracks the "computed" status instead.

This change has a visible effect: `DESCRIBE AS DDL|SDL` now does _not_ 
output the cardinality attributes if they weren't specified explicitly in
the original schema, meaning that `link foo -> Bar` would be described as
`link foo -> Bar`, and an explicit `single link foo -> Bar` would be
described as such.  This will result in a need to make a simple migration
for databases restored from previous versions of EdgeDB for applications
that use SDL schemas (in order to correctly reflect the
"inherited" and "computed" status of fields that was discarded previously).

This patch also moves some hardcoded `DESCRIBE` bits into generic `Field` 
attributes.

### DIFF
--- a/edb/edgeql/compiler/stmt.py
+++ b/edb/edgeql/compiler/stmt.py
@@ -262,7 +262,8 @@ def compile_insert_unless_conflict(
         )
 
     ptr = ptr.get_nearest_non_derived_parent(schema)
-    if ptr.get_cardinality(schema) != qltypes.SchemaCardinality.One:
+    ptr_card = ptr.get_cardinality(schema)
+    if not ptr_card.is_single():
         raise errors.QueryError(
             'UNLESS CONFLICT property must be a SINGLE property',
             context=constraint_spec.context,
@@ -1128,6 +1129,11 @@ def compile_query_subject(
         and view_rptr.ptrcls_name is not None
         and expr_rptr is not None
         and expr_rptr.direction is s_pointers.PointerDirection.Outbound
+        and expr_rptr.source.rptr is None
+        and (
+            view_rptr.source.get_bases(ctx.env.schema).first(ctx.env.schema).id
+            == expr_rptr.source.typeref.id
+        )
         and (
             view_rptr.ptrcls_is_linkprop
             == (expr_rptr.ptrref.source_ptr is not None)
--- a/edb/edgeql/qltypes.py
+++ b/edb/edgeql/qltypes.py
@@ -81,12 +81,29 @@ class SchemaCardinality(s_enum.StrEnum):
     '''This enum is used to store cardinality in the schema.'''
     One = 'One'
     Many = 'Many'
+    Unknown = 'Unknown'
+
+    def is_multi(self) -> bool:
+        if self is SchemaCardinality.One:
+            return False
+        elif self is SchemaCardinality.Many:
+            return True
+        else:
+            raise ValueError('cardinality is unknown')
+
+    def is_single(self) -> bool:
+        return not self.is_multi()
+
+    def is_known(self) -> bool:
+        return self is not SchemaCardinality.Unknown
 
     def as_ptr_qual(self) -> str:
         if self is SchemaCardinality.One:
             return 'single'
-        else:
+        elif self is SchemaCardinality.Many:
             return 'multi'
+        else:
+            raise ValueError('cardinality is unknown')
 
     def to_edgeql(self) -> str:
         return self.as_ptr_qual().upper()
--- a/edb/pgsql/metaschema.py
+++ b/edb/pgsql/metaschema.py
@@ -2761,6 +2761,8 @@ def _generate_database_views(schema):
                 AS {qi(ptr_col_name(schema, Database, 'internal'))},
             datname AS {qi(ptr_col_name(schema, Database, 'name'))},
             datname AS {qi(ptr_col_name(schema, Database, 'name__internal'))},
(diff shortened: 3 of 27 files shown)


edb/schema/inheriting.py
```diff
@@ -514,26 +514,18 @@ def _classbases_from_ast(
...
-        op: sd.AlterObjectProperty,
-    ) -> None:
+        field: str,
+        astnode: Type[qlast.DDLOperation],
+    ) -> Optional[str]:
         if (
-            op.property in {'is_abstract', 'is_final'}
-            and not isinstance(self, sd.DeleteObject)
+            field in {'is_abstract', 'is_final'}
+            and issubclass(astnode, qlast.CreateObject)
```
msullivan, line 501: Could some of this kind of logic be made data-driven with field properties?

elprans: Yes. The plan is to follow this up with some sort of structure that can be specified on the field. Ideally, the info should be bi-directional, i.e. used in `_cmd_tree_from_ast`.  I decided not to do this in this PR, since it's already fairly big and `get_ast_attr_for_field` was already there.

edb/schema/inheriting.py
```diff
@@ -156,56 +156,33 @@ def inherit_fields(
...
 
-            if (result is not None or ours is not None) and (
-                    result != ours or inherited):
-                schema = self.inherit_field(
-                    schema,
-                    context,
-                    field=field,
-                    value=result,
-                    inherited=inherited,
+            if (
```
msullivan, line 162: A comment walking through the cases here could help?

msullivan (approved):
This is a really great cleanup

edb/schema/inheriting.py
msullivan: :+1:

msullivan:
And this depends on https://github.com/edgedb/edgedb/pull/2032?

edb/edgeql/compiler/stmt.py
```diff
@@ -1107,6 +1108,11 @@ def compile_query_subject(
         and view_rptr.ptrcls_name is not None
         and expr_rptr is not None
         and expr_rptr.direction is s_pointers.PointerDirection.Outbound
+        and expr_rptr.source.rptr is None
+        and (
+            view_rptr.source.get_bases(ctx.env.schema).first(ctx.env.schema).id
+            == expr_rptr.source.typeref.id
+        )
```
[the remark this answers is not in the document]
elprans: This checks that in `Foo { computable := Foo.bar }` the start of the path on right hand side of `:=` matches the subject of the shape, in which case we treat it as a property alias.  This is in contrast to `Foo { computable := Bar.bar }`, which is _not_ an alias.  This check was missing previously to unpleasant effect.

I'll add a comment.

edb/pgsql/metaschema.py
```diff
@@ -3678,9 +3684,8 @@ def _generate_config_type_view(
             continue
 
         pp_type = pp.get_target(schema)
-        pp_multi = (
-            pp.get_cardinality(schema) is qltypes.SchemaCardinality.Many
-        )
+        pp_card = pp.get_cardinality(schema)
+        pp_multi = pp_card.is_multi()
```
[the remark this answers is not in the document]
elprans: This is an artifact of an intermediate refactoring step, where `get_cardinality()` could return `None`.  Can be shortened now.

elprans:
> And this depends on #2032?

Yes.

msullivan:
I think we should add `computed_fields` to `schema::Object`, since `inherited_fields` is exposed and the information is visible in DESCRIBE

edb/edgeql/qltypes.py
```diff
@@ -81,12 +81,29 @@ class SchemaCardinality(s_enum.StrEnum):
...
+        elif self is SchemaCardinality.Many:
+            return True
+        else:
+            raise ValueError('cardinality is unknown')
+
+    def is_single(self) -> bool:
+        return not self.is_multi()
+
+    def is_known(self) -> bool:
+        return self is not SchemaCardinality.Unknown
```
1st1, line 98: nice internal API

edb/schema/links.py
```diff
@@ -311,6 +311,24 @@ def _cmd_tree_from_ast(
...
+    def get_ast_attr_for_field(
+        self,
+        field: str,
+        astnode: Type[qlast.DDLOperation],
+    ) -> Optional[str]:
+        if (
+            field == 'required'
+            and issubclass(astnode, qlast.CreateConcreteLink)
+        ):
+            return 'is_required'
```
1st1, line 323: Maybe we finally can drop `is_`?

1st1 (approved):

edb/schema/links.py
elprans: Yes, I have that on my TODO list, but keep procrastinating on it.  Will tackle that soon.

PR #2058, opened 2020-12-15
Remove duplication of metainformation in delta command class decls

elprans:
Delta command implementations are currently associated with their
subject schema classes via the `schema_metaclass` class argument, which
is used by `ObjectCommandMeta` to populate the registry.  The class
argument is now fully redundant, because `ObjectCommand` is a `Generic`
and we can extract the schema class from its arguments instead.

This change has a benefit of increased validation of declarations,
making sure there aren't inconsistent annotations, and, more
importantly, makes it easier to use `__init_subclass__` for registries,
which, in turn, removes the need for a few metaclasses.

### DIFF
--- a/edb/common/adapter.py
+++ b/edb/common/adapter.py
@@ -29,10 +29,10 @@ class AdapterError(Exception):
     pass
 
 
+_adapters: Dict[Any, Dict[type, Adapter]] = {}
+
+
 class Adapter(type):
-    adapters: ClassVar[Dict[Any, Dict[type, Adapter]]] = {}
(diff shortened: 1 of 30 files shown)


msullivan (approved):
Looks great. Getting rid of metaclasses is nice.

Are any of the others easy to put on the chopping block?

elprans:
> Are any of the others easy to put on the chopping block?

Ideally, every metaclass that does pure registration-type stuff should be converted to an equivalent `__init_subclass__`, but I don't think we have any left in `edb.schema`.

PR #2060, opened 2020-12-16
Formalize and simplify handling of special schema object fields

elprans:
Schema object fields are not created equal.  Some, like `cardinality` or `is_owned`, require specialized treatment.  The current DDL AST binding system works at the AST node class level, and so to support existing custom field handlers we have a set of duplicate AST nodes and parser productions for each concrete schema object (e.g. for `SET OWNED` we have `AlterLinkOwned`, `AlterPropertyOwned`, `AlterIndexOwned`, etc).

Migration improvements in the pipeline require a lot more specialization like this, and it's clear that the current system has too much pointless boilerplate.  To alleviate this, add a mechanism to bind `AlterObjectFragment` implementations to specific fields.  Convert `SET/DROP OWNED` to it, and implement a custom handler for `SET SINGLE/MULTI` as a demonstration of the concept.

### DIFF
--- a/edb/schema/constraints.py
+++ b/edb/schema/constraints.py
@@ -1215,9 +1215,10 @@ class RenameConstraint(
 
 class AlterConstraintOwned(
     referencing.AlterOwned[Constraint],
+    field='is_owned',
     referrer_context_class=ConsistencySubjectCommandContext,
 ):
-    astnode = qlast.AlterConstraintOwned
+    pass
 
 
 class AlterConstraint(
--- a/edb/schema/constraints.py
+++ b/edb/schema/constraints.py
@@ -435,12 +435,15 @@ class ConstraintCommand(
 
         return args
 
-    def get_verbosename(self) -> str:
+    def get_verbosename(self, parent: Optional[str] = None) -> str:
         mcls = self.get_schema_metaclass()
         vname = mcls.get_verbosename_static(self.classname)
         if self.get_attribute_value('is_abstract'):
             vname = f'abstract {vname}'
-        return vname
+        if parent is not None:
+            return f'{vname} of {parent}'
+        else:
+            return vname
 
     def compile_expr_field(
         self,
--- a/edb/schema/delta.py
+++ b/edb/schema/delta.py
@@ -23,6 +23,7 @@ from typing import *
 import collections
 import collections.abc
 import contextlib
+import functools
 import itertools
 import uuid
 
@@ -353,6 +354,8 @@ class Command(struct.MixedStruct, metaclass=CommandMeta):
 
     #: AlterObjectProperty lookup table for get|set_attribute_value
     _attrs: Dict[str, AlterObjectProperty]
+    #: AlterSpecialObjectField lookup table
+    _special_attrs: Dict[str, AlterSpecialObjectField[so.Object]]
 
     def __init__(self, **kwargs: Any) -> None:
         super().__init__(**kwargs)
@@ -360,6 +363,7 @@ class Command(struct.MixedStruct, metaclass=CommandMeta):
         self.before_ops = []
         self.qlast: qlast.DDLOperation
         self._attrs = {}
+        self._special_attrs = {}
 
     def copy(self: Command_T) -> Command_T:
         result = super().copy()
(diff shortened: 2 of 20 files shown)


edb/schema/pointers.py
```diff
@@ -1545,6 +1560,80 @@ def _cmd_tree_from_ast(
...
+        if orig_card == new_card:
+            # The actual value hasn't changed, nothing to do here.
+            return schema
+
+        # Cardinality alters of pointers used in expressions is prohibited.
+        # Eventually we may be able to relax this by revalidating the
+        # expressions and making sure their meaning and cardinality does not
+        # change.
+        vn = scls.get_verbosename(schema, with_parent=True)
+        schema = self._prohibit_if_expr_refs(
```
msullivan, line 1592: Hm, why does it need to be prohibit instead of propagate? (The comment in in the propagate call for SetPointerType is wrong, and we don't actually prohibit type changes; I think that we do actually propagate them correctly now and generate errors and updates where necessary?)

edb/schema/pointers.py
```diff
@@ -1545,6 +1560,80 @@ def _cmd_tree_from_ast(
...
 
+class AlterPointerUpperCardinality(
+    referencing.ReferencedInheritingObjectCommand[Pointer_T],
+    inheriting.AlterInheritingObjectFragment[Pointer_T],
+    sd.AlterSpecialObjectField[Pointer_T],
+    PointerCommandOrFragment[Pointer_T],
+):
+    """Handler for the "cardinality" field changes."""
+
+    def _alter_begin(
```
msullivan, line 1571: There's obviously a *lot* of similarity here with SetPointerType. I think a bunch of the similarity couldn't be removed without unpleasant shoehorning/excess abstraction, but I think probably the error message computation and `_set_type`/`_alter_upper_cardinality` could be consolidated?

edb/schema/pointers.py
```diff
@@ -1545,6 +1560,80 @@ def _cmd_tree_from_ast(
...
+            # are not altered in the same op, and this is an error.
+            if non_altered_bases:
+                bases_str = ', '.join(
+                    b.get_verbosename(schema, with_parent=True)
+                    for b in non_altered_bases
+                )
+
+                vn = scls.get_verbosename(schema, with_parent=True)
+
+                raise errors.SchemaDefinitionError(
```
msullivan, line 1614: This is new behavior, right? Add a test?

edb/schema/pointers.py
```diff
@@ -1300,11 +1300,26 @@ def _process_create_or_alter_ast(
             )
 
         if astnode.cardinality is not None:
-            self.set_attribute_value(
-                'cardinality',
-                astnode.cardinality,
-                source_context=astnode.context,
-            )
+            if isinstance(self, sd.CreateObject):
```
msullivan, line 1303: This create/alter split really is unfortunate, yeah

edb/schema/delta.py
```diff
@@ -1470,6 +1471,53 @@ def _propagate_if_expr_refs(
 
         return schema
 
+    def _prohibit_if_expr_refs(
```
msullivan, line 1474: I added a comment elsewhere indicating that I'm not 100% sure why we need to prohibit instead of propagate, but assuming we do, I think that `_propagate_if_expr_refs` could be made to prohibit by having a flag that disables the main `if isinstance(` branch?

If you prefer to have it separated out, the `fixer` argument is dead and can be dropped.

msullivan (approved):
I left a bunch of comments/questions about specific issues involving AlterPropertyUpperCardinality.

The general mechanism here all looks solid.

edb/schema/pointers.py
elprans: I'm not 100% sure what the correct approach to propagation should be.  Should cardinality changes automatically affect cardinality of computable expressions that might use this pointer?  IMO, we should only do autopropagation if referrers are unaffected, and require an explicit opt-in otherwise.

edb/schema/pointers.py
elprans: Yeah, some deduplication seems prudent.

edb/schema/pointers.py
msullivan: Actually I think I might be wrong here, since we won't actually do all the updates necessary in actual schema stuff for things like functions; some of that can be fixed just by setting `metadata_only=False`, though

edb/schema/pointers.py
1st1: +1

edb/schema/pointers.py
msullivan: My inclination would be to automatically propagate the changes, though an opt-in would probably work fine too? Do you have syntax ideas for that?

1st1: > And then what should ALTER TYPE Foo ALTER PROPERTY bar SET DEFAULT := 'new default' do? Should it propagate to subtypes even though they have an explicit override set? Probably not.

I'm not sure I follow this example. The prevalent mode of operation for edgedb users will be modifying their schema via SDL. In which case the user would just change your schema to 

```
type Foo {
    property bar -> str { default := 'new default' }
}

type Sub extending Foo {
    overloaded property bar -> str { default := 'spam' }
}
```

And we can compile this to a targeted DDL that would just update `Foo.bar`. No need for auto-propagation or for CASCADE/RESTRICT modifiers.  Basically our DDL should be robust enough to support SDL and migrations, and I don't think we need to pack more features in it than what's absolutely required.

elprans:
I pushed a bunch of improvements which also affect existing checks somewhat.  Please review the trailing commit.

edb/schema/pointers.py
elprans: > I'm not sure I follow this example. The prevalent mode of operation for edgedb users will be modifying their schema via SDL.

The `CASCADE` thing is hypothetical, I didn't mean that it is necessary for this PR.

edb/schema/constraints.py
```diff
@@ -435,12 +435,15 @@ def _constraint_args_from_ast(
 
         return args
 
-    def get_verbosename(self) -> str:
+    def get_verbosename(self, parent: Optional[str] = None) -> str:
```
1st1, line 438: I'd make it a keyword-only arg

1st1 (approved):
lgtm

PR #2061, opened 2020-12-16
New client-side errors

tailhook:
Needed for edgedb/edgedb-js#82 as part of implementing RFC1004

### DIFF
--- a/edb/api/errors.txt
+++ b/edb/api/errors.txt
@@ -138,9 +138,9 @@
 
 0x_FF_00_00_00   ClientError
 0x_FF_01_00_00   ClientConnectionError
-0x_FF_01_01_00   ConnectionFailedError
-0x_FF_01_01_01   ConnectionFailedTemporarilyError
-0x_FF_01_02_00   ConnectionTimeoutError
+0x_FF_01_01_00   ClientConnectionFailedError
+0x_FF_01_01_01   ClientConnectionFailedTemporarilyError  #SHOULD_RECONNECT
+0x_FF_01_02_00   ClientConnectionTimeoutError  #SHOULD_RECONNECT
 
 0x_FF_02_00_00   InterfaceError
 
--- a/edb/api/errors.txt
+++ b/edb/api/errors.txt
@@ -138,6 +138,9 @@
 
 0x_FF_00_00_00   ClientError
 0x_FF_01_00_00   ClientConnectionError
+0x_FF_01_01_00   ConnectionFailedError
+0x_FF_01_01_01   ConnectionFailedTemporarilyError
+0x_FF_01_02_00   ConnectionTimeoutError
 
 0x_FF_02_00_00   InterfaceError
 
--- a/edb/tools/gen_errors.py
+++ b/edb/tools/gen_errors.py
@@ -145,8 +145,9 @@ class ErrorsTree:
 
         self._all_codes = set()
         self._all_names = set()
+        self._all_tags = dict()
 
-    def add(self, b1, b2, b3, b4, name):
+    def add(self, b1, b2, b3, b4, name, tags):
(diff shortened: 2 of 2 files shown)


edb/api/errors.txt
```diff
@@ -138,6 +138,10 @@
 
 0x_FF_00_00_00   ClientError
 0x_FF_01_00_00   ClientConnectionError
+0x_FF_01_01_00   ConnectionFailedError
+0x_FF_01_01_01   ConnectionFailedTemporarilyError
+0x_FF_01_02_00   ConnectionTimeoutError
+0x_FF_01_03_00   AuthenticationFailedError
```
elprans, line 144: How is this `AuthenticationFailedError` different from `AuthenticationError` above?  Also, I think we need to prefix all these with `Client`

tailhook: My mistake. I was looking for it as a subclass of ClientError. Fixed.

1st1: > Also, I think we need to prefix all these with Client

+1.

edb/api/errors.txt
```diff
@@ -138,6 +138,9 @@
 
 0x_FF_00_00_00   ClientError
 0x_FF_01_00_00   ClientConnectionError
+0x_FF_01_01_00   ConnectionFailedError
+0x_FF_01_01_01   ConnectionFailedTemporarilyError
+0x_FF_01_02_00   ConnectionTimeoutError
```
1st1, line 143:
```suggestion
0x_FF_01_01_00   ClientConnectionFailedError
0x_FF_01_01_01   ClientConnectionFailedTemporarilyError
0x_FF_01_02_00   ClientConnectionTimeoutError
```

tailhook: What other connection could fail? I don't think we expose any server-side connection errors with similar granularity. Keeping length of class names (including errors) sane is useful. I would rather `ClientConnectionError -> ConnectionError`.

elprans: The `Client` prefix is customary for errors generated by the _client_.

> What other connection could fail? I don't think we expose any server-side connection errors with similar granularity. 

Not yet, but it's easy to imagine connection errors stemming from server-side API integrations and foreign data wrappers.

elprans (approved):

edb/api/errors.txt
```diff
@@ -138,6 +138,9 @@
 
 0x_FF_00_00_00   ClientError
 0x_FF_01_00_00   ClientConnectionError
+0x_FF_01_01_00   ClientConnectionFailedError
+0x_FF_01_01_01   ClientConnectionFailedTemporarilyError  #SHOULD_RECONNECT
```
1st1, line 142: Paul, last bit of bikeshedding here, I promise! Should we call this TRY_RECONNECT? Also, should we adopt enum-like syntax and use CamelCase for tags?

cc @elprans

tailhook: "try reconnect" sounds like a method name. I doesn't sound like an actionable tag for error. But we can review tag names again when we have `retry`. I'm not sure what the full list of tags we need.
 
Both in python and javascript error tags are constants, not types. In python, even if that would be an enum tag names would be snake case. If in Rust they will be types we will turn snake case into a camel case. Also I like snake case here since it's clear how to turn it into camel case without ambiguities (as there is no temptation to write `AsyncIOSomething` which is `async_i_o_something` or `async_io_something` or `asyncio_something`, so other way around it might be ambiguous)

elprans: +1 for constant-style names.  Don't have a strong opinion on `TRY` vs `SHOULD`, but the latter sounds better to me.

1st1 (approved):

Issue #2057, opened 2020-12-15
Dump crashes


tailhook:
- EdgeDB Version: 1-alpha.6, 1-alpha.7, ec7a0e609

Applied migration:
```
CREATE MIGRATION m1u4jg4u3dkb737rpbwe4jeb6q735ptn65qgb2zd6p3ep2zl4nickq
    ONTO initial
{
  CREATE TYPE default::A;
  CREATE TYPE default::B {
      CREATE REQUIRED SINGLE LINK link_a -> default::A;
  };
  ALTER TYPE default::A {
      CREATE OPTIONAL MULTI LINK link_b := (.<link_a[IS default::B]);
  };
};
```
Got this:
```
[...]

PR #2063, opened 2020-12-17
Make sure computable pointers don't appear in dump

elprans:
The dump/restore implementation currently erroneously considers
computable properties and links as real and includes them in the dump.
This slipped through testing because of another bug, which causes
columns to be generated for `single` properties and links. `multi`
computables, however, are broken and it is impossible to generate a dump
of a schema containing them.

Fix this by properly excluding non-dumpable pointers.  Unfortunately,
this bug causes previously-generated dumps to contain extraneous (empty)
columns, so this patch includes a procedure to rewrite the binary `COPY`
stream to eject those columns.

A portion of this patch will be backported to a7 and perhaps earlier to
make sure those versions can produce working dumps on schemas with
`multi` computables.

Issue: #2057


edb/server/compiler/compiler.py
```diff
@@ -2139,6 +2139,11 @@ async def describe_database_restore(
         )
         ctx.state.start_tx()
 
+        dump_with_extraneous_computables = (
+            dump_server_ver is not None
+            and dump_server_ver < (1, 0, verutils.VersionStage.ALPHA, 8)
```
tailhook, line 2144: Is it still fine to use on patched version of edgedb?

edb/server/compiler/compiler.py
```diff
@@ -2139,6 +2139,11 @@ async def describe_database_restore(
         )
         ctx.state.start_tx()
 
+        dump_with_extraneous_computables = (
+            dump_server_ver is not None
```
tailhook, line 2143: Why version could be `None`? I would say it's better to assert on it, this condition sounds like we have "always latest version" dumps, which is probably nonsensical.

tailhook (approved):
Other than two questions, looks good to me.

edb/server/pgcon/pgcon.pyx
```diff
@@ -947,10 +956,90 @@ cdef class PGProto:
...
+
+        real_ncols = ncols + len(elided_cols)
+        frb_init(&rbuf, data, data_len)
+
+        elide = <int8_t*>cpythonx.PyMem_Calloc(
+            <size_t>real_ncols, sizeof(int8_t))
+
+        try:
+            for col in elided_cols:
+                elide[col] = 1
```
1st1, line 989: This calloc and the initialization of `elide` can me be moved outside of this function to `async def _restore`

1st1 (approved):

edb/server/compiler/compiler.py
1st1: I'm also curious why it can be `None` -- perhaps add a comment at least?

elprans: The earliest versions did not have the version header in dumps. I'll add a comment

edb/server/compiler/compiler.py
elprans: Yes. The backpatched version only takes care of the `multi` computable case, leaving the old behavior of dumping `single` computable columns in place.

edb/server/compiler/compiler.py
tailhook: Then `None` should be treated as a smaller version not larger. I.e. `dump_with_extraneous_computables` should be `True` for `None` not `False`.

tailhook: Ah, I see you fixed this.

PR #2066, opened 2020-12-17
Remove unused ptr_item_id column from pointer tables

elprans:
The `ptr_item_id` column is an unused vestige of the past, when a single
table could contain links or properties from multiple types.  The only
reason it stayed around is for dump compatibility, but now that we have
a mechanism to elide columns from dumps, so `ptr_item_id` goes away.


1st1 (approved):
So it looks like we needed to implement dump columns eliding anyways :)

PR #2067, opened 2020-12-18
Support computed link properties

### LINKED ISSUE #2003: Can't set computable property for abstract links
- EdgeDB Version: Alpha 6
- OS Version: Official docker container

It seems that computables that refer to other link properties do not work in SDL:
```
START MIGRATION TO {
    module default {
        abstract link my_link {
            property prop := (
                1
            );
        }
    }
};
```
The equivalent for types works just fine:
```
START MIGRATION TO {
    module default {
        type MyType {
            property prop := (
                1
            );
        }
    }
};
```

All testing done on a fresh (DROP -> CREATE) database.

msullivan:
The biggest chunk of the diff in lines is just moving a
compile_expr_field function around and adding a little to it to
support computed props in abstract pointers.

Fixes #2003.

### DIFF
--- a/edb/edgeql/compiler/stmtctx.py
+++ b/edb/edgeql/compiler/stmtctx.py
@@ -76,7 +76,9 @@ def init_context(
         # cardinality inference, so we set up the scope tree in
         # the necessary fashion.
         for singleton in options.singletons:
-            path_id = pathctx.get_path_id(singleton, ctx=ctx)
+            path_id = (pathctx.get_path_id(singleton, ctx=ctx)
+                       if isinstance(singleton, s_types.Type)
+                       else pathctx.get_pointer_path_id(singleton, ctx=ctx))
             ctx.env.path_scope.attach_path(path_id, context=None)
 
         ctx.path_scope = ctx.env.path_scope.attach_fence()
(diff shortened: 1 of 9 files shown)


edb/schema/pointers.py
```diff
@@ -1066,6 +1058,74 @@ def _parse_computable(
...
+                        type=s_objtypes.ObjectType
+                    )
+                    schema, view = base_obj.derive_subtype(
+                        schema,
+                        name=sn.QualName("__derived__",
+                                         "FakeAbstractLinkBase"),
+                        mark_derived=True, transient=True)
+                    schema, source = source.derive_ref(
+                        schema, view, target=view,
+                        mark_derived=True, transient=True)
```
msullivan, line 1102: Note that there is a big chunk of logic added here, it wasn't a pure move

edb/edgeql/compiler/stmtctx.py
```diff
@@ -76,7 +76,9 @@ def init_context(
         # cardinality inference, so we set up the scope tree in
```
elprans, line 76: This comment needs a touchup.

elprans (approved):
LGTM. A few minor nits.

edb/schema/pointers.py
```diff
@@ -1066,6 +1058,74 @@ def _parse_computable(
...
+
+                anchors[qlast.Source().name] = source
+                assert isinstance(source, (s_types.Type, Pointer))
```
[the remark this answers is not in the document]
msullivan: Doesn't quite work, unfortunately; would require changing the type of `singletons` again and that causes downstream effects worse than this

PR #2068, opened 2020-12-19
Fix a potential memory leak in taskgroup

achimnol:
refs edgedb/edgedb-python#140

* This change does not affect the existing behavior of short-lived
  task groups, but will improve memory usage of long-lived task groups
  when there are many tasks spawned repeatedly for a long time.

* In aiotools, the test case uses a virtual clock that makes
  asyncio.sleep to deterministically complete instantly, but here
  it seems not to have such thing.  I've reduced the number of
  repetitions and time scale, but not sure it's stable or not.

### DIFF
--- a/edb/common/taskgroup.py
+++ b/edb/common/taskgroup.py
@@ -25,6 +25,7 @@ import itertools
 import sys
 import textwrap
 import traceback
+import weakref
 
 
 class TaskGroup:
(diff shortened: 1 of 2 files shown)


1st1 (approved):
lgtm. Thanks for the contribution!

PR #2075, opened 2020-12-22
Fix "cannot describe type" ISE in cast+array+empty set interactions

### LINKED ISSUE #1586: InternalServerError: cannot describe type default::User

Steps to Reproduce:

Run: ` SELECT <array<User>>{};`

Schema:
```esdl
type User {
            required property nickname -> str;
            required property email -> str;
            required property email_verified -> bool {
                default := false;
            };
            required property password -> bytes;
            required property created_at -> datetime {
            };
        };
```

msullivan:
Fixes #1586.

### DIFF
--- a/edb/server/compiler/sertypes.py
+++ b/edb/server/compiler/sertypes.py
@@ -207,16 +207,6 @@ class TypeSerializer:
             raise errors.SchemaError(f'unsupported collection type {t!r}')
 
         elif isinstance(t, s_objtypes.ObjectType):
-            if t not in view_shapes:
-                # If we have an object type that isn't attached to a view,
-                # it must be the case that it doesn't actually appear in
-                # the output, so just return *something* that is well
-                # formed.
-                return self._describe_type(
-                    self.schema.get('std::uuid'), view_shapes,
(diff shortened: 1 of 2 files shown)


edb/server/compiler/sertypes.py
```diff
@@ -205,7 +206,17 @@ def _describe_type(self, t, view_shapes, view_shapes_metadata,
...
 
-        elif view_shapes.get(t):
+        elif isinstance(t, s_objtypes.ObjectType):
+            if t not in view_shapes:
+                # If we have an object type that isn't attached to a view,
+                # it must be the case that it doesn't actually appear in
+                # the output, so just return *something* that is well
+                # formed.
+                return self._describe_type(
+                    self.schema.get('std::uuid'), view_shapes,
```
elprans, line 216: I feel like a better fix would be to pretend that a `view_shape` exists, but is empty, or, even better, consisting entirely of implicit pointers.  Otherwise there is a danger of the frontend getting confused by an object not being represented properly.

elprans (approved):

PR #2076, opened 2020-12-22
Add `eql:section-intro-page` directives + add placeholder index pages

jaclarke:


1st1 (approved):
Looks great. Let's merge as is and add more text when we have the new docs in front of our eyes.

PR #2083, opened 2020-12-23
Make the error message for function variants more descriptive.

### LINKED ISSUE #1994: Error message when function variant cannot be found
This is another quick error message change on my wishlist based on my experiences. Take the following:

```
CREATE FUNCTION give_string() -> str USING ('Hi guyz');
CREATE FUNCTION give_string(input: str) -> str USING (input);
CREATE FUNCTION give_string(input: int64) -> str USING (<str>input);
SELECT give_string(8.6);
```

That gives the following error message:

```
error: could not find a function variant give_string
```

I remember missing the word 'variant' in the message and thinking I got the wrong function a few times.

A change to something like this seems easiest:

```
error: function give_string exists but not with this input. Hint: type **DESCRIBE FUNCTION give_string** to see existing variants.
```

vpetrovykh:
Update the error message to indicate that it's the function input and
not the missing function that's the issue. Also add a hint about what
can be done to view function variants.

Fixes #1994

### DIFF
--- a/edb/edgeql/compiler/func.py
+++ b/edb/edgeql/compiler/func.py
@@ -97,15 +97,49 @@ def compile_FunctionCall(
     args, kwargs, arg_ctxs = compile_call_args(expr, funcname, ctx=ctx)
     matched = polyres.find_callable(funcs, args=args, kwargs=kwargs, ctx=ctx)
     if not matched:
+        alts = [f.get_signature_as_str(env.schema) for f in funcs]
+        sig: List[str] = []
+        # This is used to generate unique arg names.
+        argnum = 0
+        for argtype, _ in args:
+            # Skip any name colliding with kwargs.
+            while f'arg{argnum}' in kwargs:
+                argnum += 1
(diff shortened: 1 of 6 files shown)


edb/edgeql/compiler/func.py
```diff
@@ -98,7 +98,10 @@ def compile_FunctionCall(
     matched = polyres.find_callable(funcs, args=args, kwargs=kwargs, ctx=ctx)
     if not matched:
         raise errors.QueryError(
-            f'could not find a function variant {funcname}',
+            f'function {funcname} exists, but not with this input',
```
elprans, line 101: This is only marginally better.  We should show a full function signature that we failed to find, like Postgres does:

`function len(integer) does not exist`

This gives more context, i.e. the actual parameter types we tried to match.  We can use the `hint` field to show existing overloads, if any.

1st1:
@vpetrovykh ping

elprans (approved):

PR #2084, opened 2020-12-24
Serialize DDL operations with an advisory lock

1st1:


edb/server/pgcon/pgcon.pyx
```diff
@@ -1122,6 +1124,28 @@ cdef class PGConnection:
         finally:
             self.after_command()
 
+    async def acquire_ddl_lock(self):
+        await self.parse_execute_json(
+            f'''
+            SELECT pg_advisory_lock({DDL_LOCK_ID});
```
elprans, line 1130: I seem to recall that we've decided to only _try_ taking the lock here (i.e. ` pg_try_advisory_lock`) and raise `ConcurrentDDLError`, instead of waiting for the lock to become available.  The `try_` approach pushes timeouts and retry logic to the client, in line with the other "retry" loops we have.

1st1:
This is superseded by https://github.com/edgedb/edgedb/pull/2183

PR #2087, opened 2020-12-25
Get rid of the remnants of the old expression normalization support

elprans:
The new expression normalization mechanism that was added in #1522
(0d110e3a75e5) left the support infrastructure for the old system in
place, namely: all the `orig_expr` bits.  This finishes the job by
ripping those bits off.


edb/edgeql/compiler/normalization.py
```diff
@@ -42,6 +45,46 @@ def normalize(
...
 
 
+def renormalize_compat(
+    norm_qltree: qlast.Base,
+    orig_text: str,
+    *,
+    schema: s_schema.Schema,
+    localnames: AbstractSet[str] = frozenset(),
+) -> qlast.Base:
+    """Renormalize an expression normalized with imprint_expr_context()."""
```
1st1, line 55: Would help if you could clarify in this docstring what renormalization would actually do to it.

1st1: Looks like this is needed for compatibility with pre a7 dumps?

1st1: And it essentially re-creates the old WITH-style normalization?

1st1 (approved):

edb/edgeql/compiler/normalization.py
elprans: It takes the old `WITH-style` normalization and renormalizes it into the new explicit module name injection.

1st1: So... can you add that to the docstring?

Issue #2089, opened 2020-12-28
EdgeQL migrations text is canonicalized in dump and can't be restored.


tailhook:
- EdgeDB Version: 1.0-alpha.8+dev.5302.g451f69c1a

# Minimal Reproduction

Migration:
```
CREATE MIGRATION m1d3g5tgdhecqihyl22x2az2ejlo5mcppyczgrbgfptw6pt5x6rvnq
    ONTO initial
{
  create type default::Type1;
};
```
(intentionally edited to lowercase `create type`)

Applied migration, dumped, trying to restore:
```
edgedb error: restoring database "edgedb": Error initiating restore protocol: ERROR: SchemaDefinitionError: specified migration name does not match the name derived from the migration contents: 'm1d3g5tgdhecqihyl22x2az2ejlo5mcppyczgrbgfptw6pt5x6rvnq', expected 'm1tg6epg6qho3vxdppjvg3s2uqihxkzxn45nwxsmuqq6e532bn27vq'
```

In dump, migration appears as:
```
CREATE APPLIED MIGRATION m1d3g5tgdhecqihyl22x2az2ejlo5mcppyczgrbgfptw6pt5x6rvnq ONTO initial {
    CREATE TYPE default::Type1;
};
```

# The Actual Issue

There are two things:
1. Currently, manually edited migrations can't survive dump/restore
2. If we change some representation of some AST node, we have a problem. This is how it happened for the first time for me:

Here is the schema:
```
module default {
  scalar type Color extending enum<Red, Blue>;
  type Type1 {
    property color -> Color {
        default := <Color>'Red';
    };
  }
};
```
EdgeDB 1.0-alpha.6+gb1863a9f0.d20201225 generates the following migration:
```
CREATE MIGRATION m1xd2ulp7ago2ziweohefvqrfqui4qyx2mrpknceo3kkbyythyhbka
    ONTO initial
{
  CREATE TYPE default::Type1;
  CREATE FINAL SCALAR TYPE default::Color EXTENDING enum<Red, Blue>;
  ALTER TYPE default::Type1 {
      CREATE OPTIONAL SINGLE PROPERTY color -> default::Color {
          SET default := <default::Color>'Red';
      };
  };
};
```
But after I load this migration with `1.0-alpha.8+dev.5302.g451f69c1a` and dump it, i get the following in the dump:
```
[...]

PR #2090, opened 2020-12-28
Preserve original text representation of migrations in dumps

elprans:
Migration name is derived from the original token stream of the
migration body.  Hence, when a migration is dumped, the original tokens
must be used.

Fixes: #2089

### DIFF
--- a/edb/edgeql/ast.py
+++ b/edb/edgeql/ast.py
@@ -691,6 +691,7 @@ class MigrationBody(DDL):
 class CreateMigration(CreateObject, Migration):
 
     body: MigrationBody
+    script: typing.Optional[str] = None
     parent: typing.Optional[ObjectRef] = None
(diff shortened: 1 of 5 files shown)


edb/edgeql/ast.py
```diff
@@ -691,6 +691,7 @@ class MigrationBody(DDL):
 class CreateMigration(CreateObject, Migration):
 
     body: MigrationBody
+    script: typing.Optional[str] = None
```
1st1, line 694: This is basically non-normalized `body` of migration? Maybe call it `raw_body`?

1st1 (approved):

edb/edgeql/ast.py
elprans: It is not _necessarily_ non-normalized.  It could be a codegen-ed body too.

PR #2103, opened 2021-01-06
Preparatory set of patches for `ALTER PROPERTY .. SET TYPE`

elprans:
See individual commits.

### DIFF
--- a/edb/common/topological.py
+++ b/edb/common/topological.py
@@ -42,8 +42,15 @@ T = TypeVar('T')
 
 class DepGraphEntry(Generic[K, V, T]):
 
+    #: The graph node
     item: V
+    #: An optional set of dependencies for the graph node as lookup keys.
     deps: MutableSet[K]
+    #: An optional set of *weak* dependencies for the graph node as
+    #: lookup keys.  The difference from regular deps is that weak deps
+    #: that cause cycles are ignored.  Essentially, weak deps dictate
+    #: a _preference_ in order rather than a requirement.
+    weak_deps: MutableSet[K]
     merge: Optional[MutableSet[K]]
     loop_control: MutableSet[K]
     extra: Optional[T]
@@ -55,6 +62,7 @@ class DepGraphEntry(Generic[K, V, T]):
         merge: Optional[MutableSet[K]] = None,
         loop_control: Optional[MutableSet[K]] = None,
         extra: Optional[T] = None,
+        weak_deps: Optional[MutableSet[K]] = None,
     ) -> None:
         self.item = item
         if deps is None:
@@ -65,6 +73,9 @@ class DepGraphEntry(Generic[K, V, T]):
             loop_control = set()
         self.loop_control = loop_control
         self.extra = extra
+        if weak_deps is None:
+            weak_deps = set()
+        self.weak_deps = weak_deps
 
 
 def sort_ex(
@@ -74,9 +85,19 @@ def sort_ex(
 ) -> Iterator[Tuple[K, DepGraphEntry[K, V, T]]]:
 
     adj: Dict[K, OrderedSet[K]] = defaultdict(OrderedSet)
+    weak_adj: Dict[K, OrderedSet[K]] = defaultdict(OrderedSet)
     loop_control: Dict[K, OrderedSet[K]] = defaultdict(OrderedSet)
 
     for item_name, item in graph.items():
+        if item.weak_deps:
+            for dep in item.weak_deps:
+                if dep in graph:
+                    weak_adj[item_name].add(dep)
+                elif not allow_unresolved:
+                    raise UnresolvedReferenceError(
+                        'reference to an undefined item {} in {}'.format(
+                            dep, item_name))
+
         if item.merge is not None:
             for merge in item.merge:
                 if merge in graph:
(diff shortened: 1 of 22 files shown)


edb/schema/objects.py
```diff
@@ -1460,24 +1460,27 @@ def init_delta_command(
...
         schema: s_schema.Schema,
+        context: sd.CommandContext,
         cmdtype: Type[sd.ObjectCommand_T],
         *,
         classname: Optional[sn.Name] = None,
         referrer: Optional[Object] = None,
         **kwargs: Any,
-    ) -> Tuple[sd.DeltaRoot, sd.ObjectCommand_T]:
-        root_cmd, parent_cmd = self.init_parent_delta_branch(
+    ) -> Tuple[sd.CommandGroup, sd.ObjectCommand_T, sd.ContextStack]:
```
msullivan, line 1480: Might be worth adding a comment to init_delta_branch about this

edb/schema/delta.py
```diff
@@ -2150,6 +2170,30 @@ def get_resolved_attribute_value(
 
         return value
 
+    def resolve_attribute_value(
```
msullivan, line 2173: Can this share code with `get_resolved_attribute_value`?

edb/schema/delta.py
```diff
@@ -3540,11 +3598,27 @@ def regular_cmd_from_ast(
...
+                and not astnode.value.elements
+            ):
                 # empty set
                 new_value = None
 
+            elif isinstance(astnode.value, qlast.TypeExpr):
+                if not isinstance(parent_op, QualifiedObjectCommand):
+                    raise AssertionError(
+                        'cannot determine module for derived compound type'
+                        ' parent operation is not a QualifiedObjectCommand'
```
msullivan, line 3612: I'm not really sure how to parse this sentence; maybe some punctuation is intended after the first line?
(Not that it really matters, as an assertion)

edb/schema/inheriting.py
```diff
@@ -170,13 +171,20 @@ def inherit_fields(
...
+                if isinstance(sav, sd.AlterObjectProperty):
+                    schema = self.scls.set_field_value(
+                        schema, field_name, result)
+                else:
+                    deferred_complex_ops.append(sav)
 
         schema = self._update_inherited_fields(
             schema, context, inherited_fields_update)
 
+        for op in deferred_complex_ops:
```
msullivan, line 185: Why's this deferral needed? Add a comment?

edb/schema/pointers.py
```diff
@@ -1477,16 +1462,28 @@ def _alter_begin(
...
+        context.current().enable_recursion = orig_rec
         scls = self.scls
 
+        vn = scls.get_verbosename(schema, with_parent=True)
+
         orig_target = scls.get_target(orig_schema)
         new_target = scls.get_target(schema)
 
+        if new_target is None:
+            # This will happen if `REST TYPE` was called
```
msullivan, line 1477: REST -> RESET

edb/schema/ordering.py
```diff
@@ -482,6 +487,25 @@ def record_field_deps(
...
 
+    def write_dep_matrix(
+        dependent: str,
+        dependent_tags: Tuple[str, ...],
+        dependency: str,
+        dependency_tags: Tuple[str, ...],
+        *,
+        as_weak: bool = False,
+    ) -> None:
+        if as_weak:
```
msullivan, line 498: I'd push the `if as_week` into the loops?

(But: great helper function.)

edb/common/topological.py
```diff
@@ -44,6 +44,7 @@ class DepGraphEntry(Generic[K, V, T]):
 
     item: V
     deps: MutableSet[K]
+    weak_deps: MutableSet[K]
```
msullivan, line 47: Could you add some comments on what the semantics of weak_deps are and how they are implemented?

msullivan (approved):
Looks good. Some nits and doc requests.

edb/schema/delta.py
elprans: Yeah, I missed a colon there

edb/server/compiler/compiler.py
```diff
@@ -874,9 +874,12 @@ def _compile_ql_migration(self, ctx: CompileContext, ql: qlast.Migration):
             mstate = mstate._replace(current_ddl=all_ddl)
             if debug.flags.delta_plan:
                 debug.header('Populate Migration DDL AST')
```
1st1, line 876: This line is redundant

edb/schema/pointers.py
```diff
@@ -482,13 +470,13 @@ def merge_targets(
...
                 vn = ptr.get_verbosename(schema)
+                t1_vn = t1.get_verbosename(schema)
                 t2_vn = t2.get_verbosename(schema)
                 raise errors.SchemaError(
                     f'cannot redefine {vnp} as {t2_vn}',
-                    details=(
-                        f'{vn} targets {t2_vn} that is not related '
-                        f'to a type found in this link in the parent type: '
-                        f'{t1.get_displayname(schema)!r}.'))
+                    details=f'{vn} is defined as {t1_vn} in a parent type, '
```
1st1, line 477: "in the parent type"? Also, I assume that `t1_vn` will include the parent type name, right?

1st1 (approved):

edb/schema/pointers.py
elprans: No it won't

elprans: And also, we are talking about _a_ parent type, because there may be several.

1st1: Can we include the actual name of the parent to the error message? Why can't we be specific here?

elprans: That would require some refactoring as we aren't passing the necessary context through the call stack. I can follow up with a cleanup later.

PR #2112, opened 2021-01-08
Fix an issue with `DESCRIBE MODULE std`.

### LINKED ISSUE #1858: DESCRIBE MODULE std is broken
Per report in #1857:

```
elvis> DESCRIBE MODULE std AS SDL;
ERROR: InternalServerError: unhashable type: 'CheckedList[str]'
```

vpetrovykh:
Add codegen for casts. Update codegen for operators and constraints.
Ensure that the generated ASTs for operators, constraints and casts are
complete.

Fixes #1858

### DIFF
--- a/edb/schema/objects.py
+++ b/edb/schema/objects.py
@@ -1191,6 +1191,10 @@ class Object(s_abc.Object, ObjectContainer, metaclass=ObjectMeta):
             val = self.get_explicit_field_value(schema, fn, default=None)
             if val is None:
                 continue
+            elif isinstance(val, collections.abc.MutableSequence):
+                # Turn the list into something hashable so it can be
+                # put in a set.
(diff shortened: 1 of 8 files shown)


edb/schema/objects.py
```diff
@@ -1185,6 +1185,10 @@ def hash_criteria(
             val = self.get_explicit_field_value(schema, fn, default=None)
             if val is None:
                 continue
+            elif isinstance(val, checked.CheckedList):
```
elprans, line 1188:
```suggestion
            elif isinstance(val, collections.abc.MutableSequence):
```

elprans (approved):

edb/schema/objects.py
vpetrovykh: Updated.

PR #2115, opened 2021-01-12
Add support for type coercion expression in `ALTER ... SET TYPE`

elprans:
`ALTER {PROPERTY|LINK} .. SET TYPE` now requires an explicit conversion 
expression specified in the `USING` clause, if the new type is not 
assignment-castable from the old type.

This doesn't include the documentation updates, but is otherwise ready
for review.

### DIFF
--- a/edb/edgeql/ast.py
+++ b/edb/edgeql/ast.py
@@ -631,6 +631,7 @@ class SetPointerType(SetField):
     name: str = 'target'
     special_syntax: bool = True
     value: TypeExpr
+    cast_expr: typing.Optional[Expr]
 
 
 class NamedDDL(DDLCommand):
--- a/edb/pgsql/delta.py
+++ b/edb/pgsql/delta.py
@@ -59,7 +59,9 @@ from edb.schema import types as s_types
 from edb.common import markup
 from edb.common import ordered
 from edb.common import topological
+from edb.common import uuidgen
 
+from edb.ir import pathid as irpathid
 from edb.ir import typeutils as irtyputils
 from edb.ir import utils as irutils
 
@@ -141,10 +143,21 @@ class MetaCommand(sd.Command, metaclass=CommandMeta):
 
     @classmethod
     def as_markup(cls, self, *, ctx):
-        node = super().as_markup(self, ctx=ctx)
+        node = markup.elements.lang.TreeNode(name=str(self))
 
-        for op in self.pgops:
-            node.add_child(node=markup.serialize(op, ctx=ctx))
+        for dd in self.pgops:
+            if isinstance(dd, AlterObjectProperty):
+                diff = markup.elements.doc.ValueDiff(
+                    before=repr(dd.old_value), after=repr(dd.new_value))
+
+                if dd.new_inherited:
+                    diff.comment = 'inherited'
+                elif dd.new_computed:
+                    diff.comment = 'computed'
+
+                node.add_child(label=dd.property, node=diff)
+            else:
+                node.add_child(node=markup.serialize(dd, ctx=ctx))
 
         return node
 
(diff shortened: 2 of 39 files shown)


edb/schema/scalars.py
```diff
@@ -139,6 +142,21 @@ def implicitly_castable_to(
         assert isinstance(right, s_types.Type)
         return s_casts.is_implicitly_castable(schema, left, right)
 
+    def castable_to(
+        self,
+        other: s_types.Type,
+        schema: s_schema.Schema,
+    ) -> bool:
```
1st1, line 149: I'd add a docstring, which can be the commit message, verbatim.

edb/edgeql/ast.py
```diff
@@ -631,6 +631,7 @@ class SetPointerType(SetField):
     name: str = 'target'
     special_syntax: bool = True
     value: TypeExpr
+    expr: typing.Optional[Expr]
```
msullivan, line 634: cast_expr, maybe?

edb/pgsql/delta.py
```diff
@@ -2252,36 +2280,337 @@ def adjust_pointer_storage(self, pointer, schema, orig_schema, context):
...
+                # object.
+                and (not is_required or not expr_is_nullable)
+            )
+
+            need_temp_col = (
+                (is_multi and expr_is_nullable)
+                or (changing_col_type and not using_expr_is_trivial)
+            )
+
+            if not using_expr_is_trivial:
```
msullivan, line 2402: Add a comment about what's being set up here?

edb/pgsql/delta.py
```diff
@@ -2252,36 +2280,337 @@ def adjust_pointer_storage(self, pointer, schema, orig_schema, context):
...
+                    INTO _dummy_text;
+                '''))
+            else:
+                clean_nulls = dbops.Query(textwrap.dedent(f'''\
+                    DELETE FROM {tab} WHERE {col} IS NULL
+                '''))
+
+            self.pgops.add(clean_nulls)
+
+        elif aux_ptr_table is not None:
```
msullivan, line 2563: What's happening here?

edb/pgsql/delta.py
```diff
@@ -2252,36 +2280,337 @@ def adjust_pointer_storage(self, pointer, schema, orig_schema, context):
...
-                        ))
+                if (
+                    old_target != new_target
+                    and not type_change_ok
+                    and not pointer.is_endpoint_pointer(orig_schema)
+                ):
+                    self._alter_pointer_type(
+                        pointer, schema, orig_schema, context)
+
+    def _alter_pointer_type(self, pointer, schema, orig_schema, context):
```
msullivan, line 2291: This function is a doozy; it is probably worth considering whether it can be split up in a way that makes it easier to read/think about. (Maybe it can't be, though)

edb/schema/pointers.py
```diff
@@ -1484,10 +1532,72 @@ def _alter_begin(
...
 
-        schema = self._propagate_if_expr_refs(
-            schema, context, action=f'alter the type of {vn}')
-
         if not context.canonical:
+            assert orig_target is not None
+            assert new_target is not None
+            pop = self.get_parent_op(context)
+
+            if (
```
msullivan, line 1540: What is supposed to happen if a USING expression is specified when the value can be casted? It looks like it is silently ignored, which seems bad

msullivan (approved):
This seems pretty reasonable to me, though we should make sure something sensible happens when USING is specified on a value that *can* be casted.

edb/schema/pointers.py
elprans: Yikes. This is unintentional. Cast expr should always be evaluated.

edb/pgsql/delta.py
elprans: I tried, but the result looked even harder to follow for me and was much longer.

edb/pgsql/delta.py
elprans: The relevant comment is under the `if using_eql_expr is not None`, I'll add a pointer to it

edb/schema/expr.py
```diff
@@ -255,6 +258,42 @@ def schema_refs_from_data(
     ) -> FrozenSet[uuid.UUID]:
         return so.ObjectCollection.schema_refs_from_data(data[1])
 
+    @property
+    def stype(self) -> s_types.Type:
+        from edb.ir import ast as irast_
+
+        if not self.is_compiled():
```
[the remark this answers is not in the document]
elprans: Good idea.

PR #2118, opened 2021-01-16
Improve migration prompts

elprans:
Add collection of improvements intended to improve overall fidelity of
migration prompts in `DESCRIBE MIGRATION`.

Teach the delta tree reordering to take delta confidence into account, and
only gather subcommands if their `confidence` is `1.0`.  Further, use the
innermost command in a simple-chain command tree to determine user prompts,
for example in `ALTER TYPE Foo ALTER PROPERTY bar` the prompt will address
property `bar` and not just type `Foo`.

Finally, the optimize the client/server `DESCRIBE MIGRATION` loop for the
common case of client's acceptance of proposed DDL.  This removes the need
to rederive the proposed plan on every step and results in better prompts.

### DIFF
--- a/edb/schema/ddl.py
+++ b/edb/schema/ddl.py
@@ -18,9 +18,9 @@
 
 
 from __future__ import annotations
+from typing import *
 
 from collections import defaultdict
-from typing import *  # noqa
 
 from edb import edgeql
 from edb.edgeql import ast as qlast
@@ -233,6 +233,12 @@ def delta_schemas(
                 )
                 assert isinstance(create, sd.CreateObject)
                 create.if_not_exists = True
+                # We currently fully assume that modules are created
+                # or deleted and never renamed.  This is fine, because module
+                # objects are never actually referenced directly, only by
+                # the virtue of being the leading part of a fully-qualified
+                # name.
+                create.set_annotation('confidence', 1.0)
 
                 result.add(create)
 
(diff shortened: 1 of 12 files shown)


edb/schema/delta.py
```diff
@@ -201,18 +202,22 @@ def can_delete(name: sn.Name) -> bool:
                 or not can_delete(y_name)
                 or x_name in renames_x
             ):
+                if (
+                    (x_alter_variants[x] > 1 or can_create(x_name))
+                    and parent_confidence != 1.0
+                ):
+                    confidence = s
+                else:
+                    confidence = 1.0
```
msullivan, line 211: Why does parent confidence of 1.0 imply child confidence of 1.0? Because any child change that doesn't affect the parent confidence should be considered to be perfectly confident?

msullivan (approved):

edb/schema/pointers.py
```diff
@@ -228,21 +229,128 @@ def merge_target(
...
+        t1_cls = 'property' if isinstance(t1, s_abc.ScalarType) else 'link'
+        t2_cls = 'property' if isinstance(t2, s_abc.ScalarType) else 'link'
+
+        t1_source_vn = t1_source.get_verbosename(schema, with_parent=True)
+        if t2_source is None:
+            raise errors.SchemaError(
+                f'cannot redefine {vnp} as {t2_vn}',
+                details=(
+                    f'{vn} is defined as a {t1_cls} to {t1_vn} in'
+                    f' parent {t1_source_vn}'
```
1st1, line 297:
```suggestion
                    f' parent type {t1_source_vn}'
```

edb/schema/pointers.py
```diff
@@ -228,21 +229,128 @@ def merge_target(
...
+                ),
+            )
+        else:
+            t2_source_vn = t2_source.get_verbosename(schema, with_parent=True)
+            raise errors.SchemaError(
+                f'inherited {vnp} has a type conflict',
+                details=(
+                    f'{vn} is defined as a {t1_cls} to {t1_vn} in'
+                    f' parent {t1_source_vn} and as {t2_cls} in'
+                    f' parent {t2_source_vn}'
```
1st1, line 307:
```suggestion
                    f' parent type {t1_source_vn} and as {t2_cls} in'
                    f' parent type {t2_source_vn}'
```

1st1 (approved):

edb/schema/pointers.py
elprans: The result of `get_verbosename()` already includes the word "type".

edb/schema/pointers.py
elprans: The result of `get_verbosename()` already includes the word "type".

edb/schema/ddl.py
```diff
@@ -234,6 +234,9 @@ def _filter(schema: s_schema.Schema, obj: so.Object) -> bool:
                 assert isinstance(create, sd.CreateObject)
                 create.if_not_exists = True
 
+                if generate_prompts:
```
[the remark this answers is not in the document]
elprans: Added comments.

edb/schema/delta.py
elprans: Yeah, this is intended to handle the innards of fully-confident `CREATE`.  It's less clear for `ALTER`, because we may be confident about the `ALTER` (via guidance ban on `DROP`/`CREATE`), but that does not necessarily mean full confidence of the child alters.  Will likely need to do more tinkering here, I left a TODO here so we don't forget.

Issue #272, opened 2018-10-25
Consider implementing '\(expr)' string interpolation like in Swift


1st1:
Example:

```
SELECT User {
   friends_msg := '\(User.name) has \(count(User.friends)) friends'
}
```

Prior art:
* [Python f-strings](https://www.python.org/dev/peps/pep-0498/)
* [Swift interpolated strings](https://developer.apple.com/documentation/swift/string)

msullivan:
We still reject strings containing `\(` so we could do the original syntax or the f-string version.

I'm partial to f-strings but only because I'm very used to them.

1st1:
> We still reject strings containing \( so we could do the original syntax or the f-string version.

We should do `\(` because it's a drop in syntax without requiring a new string literal type. This is what Guido actually wanted to do with f-strings (just make all strings f-strings) but couldn't because of the obvious reasons.

Gobot1234:
Would this be subject to limitations that pre python 3.12 are subject to caused by its parser?

PR #2119, opened 2021-01-17
Implement required user input feedback in DESCRIBE MIGRATION

elprans:
Teach `DESCRIBE CURRENT MIGRATION AS JSON` to include a list of client 
prompts for expressions required to complete the migration.  The proposed
statement text will now include placeholders in the form of
`\(placeholder)` where the client response should be inserted.

This patch implements client input requests for `SET TYPE` operations
(`AlterPointerType`).

### DIFF
--- a/edb/schema/delta.py
+++ b/edb/schema/delta.py
@@ -411,6 +411,32 @@ class Command(struct.MixedStruct, metaclass=CommandMeta):
     def is_data_safe(self) -> bool:
         return False
 
+    def get_required_user_input(self) -> Dict[str, str]:
+        return {}
+
+    def record_diff_annotations(
+        self,
+        schema: s_schema.Schema,
+        orig_schema: Optional[s_schema.Schema],
+        context: so.ComparisonContext,
+    ) -> None:
+        """Record extra information on a delta obtained by diffing schemas.
+
+        This provides an apportunity for a delta command to annotate itself
+        in schema diff schenarios (i.e. migrations).
+
+        Args:
+            schema:
+                Final schema of a migration.
+
+            orig_schema:
+                Original schema of a migration.
+
+            context:
+                Schema comparison context.
+        """
+        pass
+
     def resolve_obj_collection(
         self,
         value: Any,
@@ -1462,6 +1488,16 @@ class ObjectCommand(Command, Generic[so.Object_T]):
                 subcmd.is_data_safe()
                 for subcmd in self.get_subcommands()
             )
+
+    def get_required_user_input(self) -> Dict[str, str]:
+        result: Dict[str, str] = self.get_annotation('required_input')
+        if result is None:
+            result = {}
+        for cmd in self.get_subcommands():
+            subresult = cmd.get_required_user_input()
+            if subresult:
+                result.update(subresult)
+        return result
 
     def get_friendly_description(
         self,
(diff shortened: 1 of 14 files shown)


edb/schema/delta.py
```diff
@@ -408,6 +408,17 @@ def adapt(cls: Type[Command_T], obj: Command) -> Command_T:
     def is_data_safe(self) -> bool:
         return False
 
+    def get_required_user_input(self) -> Dict[str, str]:
+        return {}
+
+    def record_diff_annotations(
```
msullivan, line 414: docstring?

tests/test_edgeql_data_migration.py
```diff
@@ -101,8 +114,38 @@ async def fast_forward_describe_migration(self, *, limit=None):
                     await self.con.execute('COMMIT MIGRATION;')
                     break
 
+                interpolations = {}
+
+                user_input_reqs = mig['proposed']['required_user_input']
+                if user_input_reqs:
+                    for var in user_input_reqs:
```
msullivan, line 121: Are these guaranteed to be in the right order? (I guess they might be, if `json` always respects the orderings?)

tests/test_edgeql_data_migration.py
```diff
@@ -3808,15 +3872,31 @@ async def test_edgeql_migration_eq_33(self):
...
+            r"""
+                type Child;
+                type Child2;
+
+                type Base {
+                    # change link type
+                    link foo -> Child2;
+                }
+            """,
+            user_input=[
```
msullivan, line 3896: Could you add a test that requires two user inputs?

msullivan (approved):
Looks good! Exciting to have this coming together.

Just a couple doc/test requests

1st1 (approved):

tailhook:
So what's the grammar here, is `\(something)` is a single token? Currently we have `\` disallowed, so it's fine, just double checking. (I must be doing replacement token-by-token, because otherwise I risk replacing something in quoted string. So this must be a part of tokenizer).

tailhook:
> So what's the grammar here, is `\(something)` is a single token? Currently we have `\` disallowed, so it's fine, just double checking.

I should ask have been asked if it's regex `\\\([a-zA-Z0-9_]+\)`? name be a valid identifier? or do we expect arbitrary strings in parenthesis?

elprans:
For the DDL template placeholders `\\\([a-zA-Z0-9_]+\)` is the correct pattern.  These placeholders are always system-generated.  That said, the planned string interpolation #272 uses `\()` in string literals and is supposed to work like a Python f-string.

tests/test_edgeql_data_migration.py
```diff
@@ -7930,6 +7979,36 @@ async def test_edgeql_migration_prompt_id_01(self):
...
+            type Bar { property foo -> int64 };
+        ''')
+
+        await self.assert_describe_migration({
+            'proposed': {
+                'statements': [{
+                    'text': '''
+                        ALTER TYPE test::Bar {
+                            ALTER PROPERTY foo {
+                                SET TYPE std::int64 USING (\\(cast_expr));
```
1st1, line 7997: I'd use an `r''' ... '''` string to have single slashes here. Would make it easier to read.

edb/schema/delta.py
elprans: Added

tests/test_edgeql_data_migration.py
elprans: Yes, `required_user_input` is an array, and we explicitly guarantee stable command order in diffs.

tests/test_edgeql_data_migration.py
elprans: Good point. Added that to `test_edgeql_migration_eq_collections_07`

PR #2122, opened 2021-01-18
Looks like a typo

Dhghomon:


1st1:
> Looks like a typo

No, the docs are correct. Sum of int32 can overflow, so the result is int64 on purpose. We don't have a variant for int16 because it will be upcasted to int32. Not sure why we have a variant for int32 though, perhaps to avoid it being upcasted to float64 -- @elprans ?

elprans:
Yeah, `int32` variants exist to disambiguate casting.

PR #2130, opened 2021-01-19
Implement syntax to specify a fill expression in SET REQUIRED

elprans:
When an existing link or property is altered to become `required`,
existing objects must be updated for the new constraint to be valid.
This implements a `USING` clause for `SET REQUIRED` which acts similarly
to the analogous clause of the `SET TYPE` operation.

### DIFF
--- a/edb/edgeql/codegen.py
+++ b/edb/edgeql/codegen.py
@@ -1551,6 +1551,22 @@ class EdgeQLSourceGenerator(codegen.SourceGenerator):
                 self.visit(node.cast_expr)
                 self.write(')')
 
+    def visit_SetPointerOptionality(
+        self,
+        node: qlast.SetPointerOptionality,
+    ) -> None:
+        if node.value is None:
+            self.write('RESET OPTIONALITY')
+        else:
+            if self._eval_bool_expr(node.value):
+                self.write('SET REQUIRED')
+            else:
+                self.write('SET OPTIONAL')
+            if node.fill_expr is not None:
+                self.write(' USING (')
+                self.visit(node.fill_expr)
+                self.write(')')
+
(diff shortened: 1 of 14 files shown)


edb/edgeql/codegen.py
```diff
@@ -1551,6 +1551,22 @@ def visit_SetPointerType(self, node: qlast.SetPointerType) -> None:
                 self.visit(node.cast_expr)
                 self.write(')')
 
+    def visit_SetPointerOptionality(
+        self,
+        node: qlast.SetPointerOptionality,
+    ) -> None:
+        if node.value is None:
+            self.write('RESET OPTIONALITY')
```
1st1, line 1559: Can't say I'm a big fan of this, but I see why this can be a better UX that distinct RESET REQUIRED and RESET OPTIONAL.

1st1 (approved):

edb/pgsql/delta.py
```diff
@@ -2258,12 +2258,142 @@ def adjust_pointer_storage(self, pointer, schema, orig_schema, context):
...
+                self.pgops.add(dbops.Query(update_qry))
+            else:
+                source = ptr.get_source(orig_schema)
+                src_tab = q(*common.get_backend_name(
+                    orig_schema,
+                    source,
+                    catenate=False,
+                ))
+
+                update_qry = textwrap.dedent(f'''\
```
msullivan, line 2336: For this, do we create a new link table and then delete the old one?

msullivan (approved):

edb/pgsql/delta.py
elprans: No, this simply inserts the result of evaluation of the `USING` clause into a link table for every object that has an empty link set.  I'll clarify in a comment.

Issue #1778, opened 2020-09-16
Disallow overloading SINGLETON/SET OF/OPTIONAL based on type?


msullivan:
Currently type-based function overloading is allowed to select argument type modifiers (SINGLETON/SET OF/OPTIONAL). There is substantial complexity in edb/edgeql/compiler/func.py and the scope tree to handle this (a bunch added recently, in #1762).

I think that argument type modifiers (SET OF, in particular, which induces subqueries) affect the semantics of the query enough that it should be easy and straightforward to determine them.

I propose that we require that all function overloads with matching signatures (number of positional args + names of name only arguments) have the same type modifiers. We can then decide which type modifiers to use based on just the function signature, before we necessarily know the types, and then figure out which exact instance later. (I wouldn't mind banning overloading SET OF here, either, since SET OF is a big deal, but I don't have any technical arguments.)

This will let us clean up function compilation to be able to know the argument kind in advance. (We could also accomplish this by having a typechecker that directly typechecks EdgeQL in advance, which I think is probably a good idea either way but is also probably a bunch of work.) It also means that an evaluation model can get away without needing to do a static typecheck. (I think. There might be other problems.)

---

There are currently 4 offending functions in the stdlib:
`std::to_str` (the deprecated array overload) and `cal::to_local_{datetime,date,time}`.
They can all be fixed without much trouble. (The second argument to the array overload for `to_str` can be made optional, though it is deprecated and being removed anyway, and the cal functions can have the fmt argument made named only, or made non-OPTIONAL and given a default of "" or something).

A lot of the OPTIONAL arguments aren't *really* OPTIONAL, I think, also; the optionality is just there to be a default value for a parameter and probably nobody ever plans to pass a potentially empty set in.

1st1:
> I propose that we require that all function overloads with matching signatures (number of positional args + names of name only arguments) have the same type modifiers.

+1, and I'm surprised that this isn't the case already. We should add a check for that in `schema/functions.py` to enforce.

> the cal functions can have the fmt argument made named only, or made non-OPTIONAL and given a default of "" or something).

The motivation for `fmt` to be an `OPTIONAL` stems from the desire to make `""` format an empty format, e.g. `to_str(1, '')` would produce an empty string. I think that's how Postgres' formatting functions are defined. We can have some other special default placeholder other than an empty string though.

tailhook:
[...]

PR #2132, opened 2021-01-19
Determine func/oper argument type modifiers in advance of compiling args

msullivan:
This allows us to compile arguments with the correct fences in the
first place, instead of trying to fix them up after the fact (which
does not work very well). The idea is that this is a prereq to pending
scope tree/namespace fixes I am working on.

This required implementing #1778, which disallows overloads with
different type modifiers that we can't resolve based on argument count
and names. Right now, to get something working and to unblock myself
on the chain of scope work this is at the head of, I do that check at
the *call* sites. This should be changed to be done at the definition
site in a follow-up.

To resolve the places in the stdlib where we break this new rule, I
made `{}` allowed as a default placeholder argument for non-OPTIONAL
arguments and used that to make the fmt argument for a number of cal
functions non-OPTIONAL. There are other choices here that might be better,
but this one only breaks compatability a little so I went with it for now.
This is easy to change (before 1.0).

This change allowed me to rip out a bunch nasty mechanisms for dealing
with OPTIONAL that I introduced several months ago (optional_count,
path_log).

### DIFF
--- a/edb/edgeql/compiler/context.py
+++ b/edb/edgeql/compiler/context.py
@@ -498,9 +498,6 @@ class ContextLevel(compiler.ContextLevel):
     """A set of schema objects for which the shadowing rewrite should be
        disabled."""
 
-    path_log: Optional[List[irast.PathId]]
-    """An optional list of path ids added to the scope tree in this context."""
-
     def __init__(
         self,
         prevlevel: Optional[ContextLevel],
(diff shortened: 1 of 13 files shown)


elprans (approved):
LGTM to me as an interim step, but

> There are other choices here that might be better, but this one only breaks compatability a little so I went with it for now. This is easy to change (before 1.0).

Can you elaborate a bit on future remedies you have in mind for this?

msullivan:
I'm going to close this, since we don't need it now. Maybe we'll want it later.

Issue #1381, opened 2020-05-06
Incorrect result for a value computed in WITH block


vpetrovykh:
Run the following in the existing scope tests dataset (from `test_edgeql_scope.py`):
```
WITH 
    MODULE test,
    avg := math::mean({len(Card.name)})
SELECT Card {name, a := avg};
```
Instead of being the same value across all results `a` is different for each as if the function call was inlined.

vpetrovykh:
I can even simplify it further to:
```
WITH 
    MODULE test,
    avg := count({Card.name})
SELECT Card {name, a := avg};
```
Or even
```
WITH 
    MODULE test,
    avg := Card.name
SELECT Card {name, a := avg};
```

1st1:
@elprans can we fix this for alpha 3?

1st1:
(if not move it to Alpha 4)

elprans:
The fix isn't trivial, so Alpha 4.

msullivan:
Also: `WITH X := {1, 2}, Y := X+1 SELECT (X, Y)`.
Should produce a cross-product output with 4 results, actually produces 2.

PR #2137, opened 2021-01-21
Make scope tree unique_ids actually be unique

msullivan:
We do this by getting rid of the use of "temporary" scopes in
declare_view (an approach that I previewed in #1806).

The purpose of using the temporary scopes was I think to allow the use
of the extra_scope mechanism to place the scope tree for a WITH
binding *underneath* the scope tree of a use of it. Unfortunately this
could cause duplication of unique_ids when it got grafted into two
independent subqueries, as well as some of other issues when the
fusing didn't behave right.

Fortunately, all that extra_scope machinery wasn't really needed. It
seems to work basically just fine to stick it in the scope tree where
it appears.

This fixes a handful of bugs, but also introduces some failures:
 * test_edgeql_tree_update_05 (flipping two nodes) generates invalid
   SQL, but I think fixing weak namespace stripping should fix this.
 * test_edgeql_volatility_with_01 (now split into a new test)
   fails because we now fail to detect volatile WITH bindings
   in situations like `X := random(), Y := X`. Fixing this should
   be straightforward.
 * test_edgeql_select_tvariant_04 fails because we now fail to
   deduplicate computed link sets in some cases. I haven't dug
   into this one yet.

This is work towards fixing #1381.

I'm putting this up now to get eyes on it, but if we want to wait on
landing it for me to fix some of the remaining problems, that would
make sense.

This wound up being a very long and painful journey to a pretty small
patch.

### DIFF
--- a/edb/edgeql/compiler/setgen.py
+++ b/edb/edgeql/compiler/setgen.py
@@ -229,7 +229,6 @@ def compile_path(expr: qlast.Path, *, ctx: context.ContextLevel) -> irast.Set:
                 'could not resolve partial path ',
                 context=expr.context)
 
-    extra_scopes = {}
     computables = []
     path_sets = []
 
@@ -283,8 +282,6 @@ def compile_path(expr: qlast.Path, *, ctx: context.ContextLevel) -> irast.Set:
                         is_binding=True,
                         ctx=ctx,
                     )
-
-                    extra_scopes[path_tip] = view_scope_info
                 else:
                     path_tip = class_set(stype, ctx=ctx)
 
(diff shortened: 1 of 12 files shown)


edb/edgeql/compiler/setgen.py
```diff
@@ -411,10 +411,28 @@ def compile_path(expr: qlast.Path, *, ctx: context.ContextLevel) -> irast.Set:
...
                     dispatch.compile(step, ctx=subctx), ctx=subctx)
 
+                # If the head of the path is a direct object
+                # reference, wrap it in an expression set to give it a
+                # new path id. This prevents the object path from being
+                # spuriously visible to computable paths defined in a shape
+                # at the root of a path. (See test_edgeql_select_tvariant_04
+                # for an example).
+                if (
+                    path_tip.path_id.is_objtype_path()
```
elprans, line 429: You probably also want to add `and not path_id.is_view_path()` to actually test that `path_id` is a direct reference, as `is_objtype_path` only tests if `path_id` represents an object type.

elprans (approved):

PR #2138, opened 2021-01-22
Properly handle non-DDL commands in migration blocks

elprans:
RFC 1000 prescribes that a non-DDL command, like `INSERT`, inside a
`START MIGRATION` block should not be executed right away and should be 
recorded as part of the migration script instead.  This patch basically 
implements that by adding the new `Query` delta command, which simply 
compiles a given EdgeQL statement.

Also, while at it, add proper checks around which statements can and cannot
appear in migration blocks.

### DIFF
--- a/edb/common/markup/renderers/terminal.py
+++ b/edb/common/markup/renderers/terminal.py
@@ -82,6 +82,12 @@ class Buffer:
         yield
         self.data.append((FOLDABLE_LINES_END, ))
 
+    @contextlib.contextmanager
+    def non_foldable_lines(self):
+        self.data.append((FOLDABLE_LINES_END, ))
+        yield
+        self.data.append((FOLDABLE_LINES_START, ))
+
     def mark_line_break(self):
         self.data.append((LINE_BREAK, ))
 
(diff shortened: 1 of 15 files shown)


edb/common/markup/renderers/terminal.py
```diff
@@ -429,6 +435,18 @@ def _render_lang_String(self, element):
...
+        with self.buffer.non_foldable_lines():
+            for line in element.str.splitlines():
+                self.buffer.new_line()
+                self.buffer.write(
+                    line,
+                    style=self.styles.literal
+                )
+            self.buffer.data.append((DEDENT_NO_NL, ))
+            self.buffer.new_line()
+            self.buffer.data.append((INDENT_NO_NL, ))
```
1st1, line 448: It's been long time we've improved anything in this file...

edb/server/mng_port/edgecon.pyx
```diff
@@ -1380,13 +1380,14 @@ cdef class EdgeConnection:
...
             else:
-                await conn.parse_execute(
-                    query_unit,         # =query
-                    self,               # =edgecon
-                    bound_args_buf,     # =bind_data
-                    use_prep_stmt,      # =use_prep_stmt
-                    state,              # =state
-                )
+                if query_unit.sql:
+                    await conn.parse_execute(
```
1st1, line 1384: You probably also want to update the `simple_query()` method.

edb/server/compiler/status.py
```diff
@@ -32,17 +32,17 @@ def get_status(ql: qlast.Base) -> bytes:
 
 @get_status.register(qlast.CreateObject)
 def _ddl_create(ql):
-    return b'CREATE'
+    return f'CREATE {ql.object_class}'.encode()
```
1st1, line 35: What if `object_class` enum value contains an underscore in it?

1st1 (approved):

edb/server/compiler/status.py
elprans: It's not supposed to, we use spaces there.

edb/server/mng_port/edgecon.pyx
elprans: OK, added `if query_unit.sql` guards everywhere where `edgecon` attempts to run it.

edb/common/markup/renderers/terminal.py
elprans: It's a rather nasty hack, but I couldn't find any prettier way to do this:

```
Delta(
   Query(
       multiline query text
       goes here
   )
)
```

PR #2148, opened 2021-01-24
Add readonly property to schema

kfrp:
ref [comment](https://github.com/edgedb/edgedb/discussions/2146#discussioncomment-306690)


elprans:
Thanks for the contribution, @kfrp!  Let's also add a test for this.  Amending `test_edgeql_introspection_meta_default_03` seems like a good target (select `readonly` alongside `required`, the `__type__` link should be `readonly: True`).

elprans (approved):
Thanks!

PR #2151, opened 2021-01-26
Replace weak namespace stripping with smarter rewriting

### LINKED ISSUE #1381: Incorrect result for a value computed in WITH block

msullivan:
Instead of "stripping" weak namespaces from all paths, instead rewrite
weak namespaces to be "absolute" by removing namespaces that don't
appear at the binding site for the path.

Fixes #1381.

### DIFF
--- a/edb/edgeql/compiler/stmtctx.py
+++ b/edb/edgeql/compiler/stmtctx.py
@@ -41,6 +41,8 @@ from edb.edgeql import ast as qlast
 from edb.edgeql import qltypes
 from edb.edgeql import parser as qlparser
 
+from edb.common.ast import visitor as ast_visitor
+
 from . import astutils
 from . import context
 from . import dispatch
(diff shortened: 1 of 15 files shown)


edb/edgeql/compiler/stmtctx.py
```diff
@@ -254,10 +250,106 @@ def fini_expression(
...
+            # in temporary scopes, so we need to just skip those.
+            if scope := tree.find_by_unique_id(path_scope_id):
+                _try_namespace_fix(scope, ir_set)
+
+
 def _elide_derived_ancestors(
     obj: Union[s_types.InheritingType, s_pointers.Pointer], *,
     ctx: context.ContextLevel
 ) -> None:
+
```
elprans, line 352: Looks like a spurious newline.

elprans (approved):
Nice! :tada:

msullivan:
Oh, oops, it looks like I never ran test_edgeql_ir_scopetree after I got rid of some hacks I used to use the old namespace stripping while doing test setup, so it need a lot of updating

msullivan:
I ended up writing https://gist.github.com/msullivan/0d3e3d52b9cca471ddcc29b31cc61def to automate the scope tree test fixing.

PR #2159, opened 2021-01-28
Fix a collection of issues with DML inheritance and overlays

msullivan:
* Make WITH bound DML operations appear in the returned values
   (by tracking the stack of current DML statements explicitly instead
    of through parent_stmt)
 * Make subtypes of a link type be returned by nested DML operations.
   (By tracking ancestors in the TypeRefs and updating their type
   overlays also).
 * Make the results of nested UPDATEs appear in returned values.
   (By constructing an overlay that filters out the updated objects
   and then unions them back in.)

There's also a refactor of the overlay datastructure to use nested
dicts, which is groundwork for making WITH overlays keep working in a
soon forthcoming PR to make WITH bound DML be executed the correct
number of times.

### DIFF
--- a/edb/edgeql/compiler/typegen.py
+++ b/edb/edgeql/compiler/typegen.py
@@ -224,7 +224,8 @@ def type_to_typeref(
     )
     include_ancestors = (
         expr_type is s_types.ExprType.Insert
-        or include_descendants
+        or expr_type is s_types.ExprType.Update
+        or expr_type is s_types.ExprType.Delete
     )
     return irtyputils.type_to_typeref(
         schema,
--- a/edb/edgeql/compiler/typegen.py
+++ b/edb/edgeql/compiler/typegen.py
@@ -222,9 +222,14 @@ def type_to_typeref(
         expr_type is s_types.ExprType.Update
         or expr_type is s_types.ExprType.Delete
     )
+    include_ancestors = (
+        expr_type is s_types.ExprType.Insert
+        or include_descendants
+    )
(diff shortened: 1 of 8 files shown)


edb/pgsql/compiler/relctx.py
```diff
@@ -1019,29 +1020,63 @@ def range_for_typeref(
...
-        )
+        qry = set_ops[0][1]
 
         for op, rarg in set_ops[1:]:
-            qry.op, qry.rarg = op, rarg
-            qry = pgast.SelectStmt(
-                all=True,
-                larg=qry
-            )
+            if op == 'filter':
```
elprans, line 1067: Hmm. I'm curious why the `except` op didn't work here?

edb/edgeql/compiler/typegen.py
```diff
@@ -222,9 +222,14 @@ def type_to_typeref(
         expr_type is s_types.ExprType.Update
         or expr_type is s_types.ExprType.Delete
     )
+    include_ancestors = (
+        expr_type is s_types.ExprType.Insert
+        or include_descendants
```
elprans, line 227: It's probably cleaner to restate DML types here in case `include_descendants` deviates from the current set of conditions.

elprans (approved):

edb/pgsql/compiler/relctx.py
msullivan: `except` compares the whole row, and we need to compare just the keys. (The whole row won't match, since it's an UPDATE)

Issue #2157, opened 2021-01-27
Improve documentation on the behavior of operators, especially logical with respect to empty sets and coalescing


elprans:
Per discussion, we need to clearly outline how scalar operators work, most importantly, logical operators in situations where either side is an empty set, and how to express common scenarios using `IF`, `??`, coalescing comparison, `all/any`, shapes and `FOR`.  Ideally, we should have a separate reference page, and integrate this knowledge into the book.

PR #2160, opened 2021-01-28
Add a cheatsheet focused on boolean expressions and gotchas.

vpetrovykh:
Issue #2157

### DIFF
--- a/docs/cheatsheet/boolean.rst
+++ b/docs/cheatsheet/boolean.rst
@@ -1,0 +1,233 @@
+.. _ref_cheatsheet_boolean:
+
+Boolean Expressions
+===================
+
+Boolean expressions can be tricky sometimes, so here are a handful of
+tips and gotchas.
+
+There's a fundamental difference in how ``{}`` is treated by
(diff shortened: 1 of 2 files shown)


docs/cheatsheet/boolean.rst
```diff
@@ -0,0 +1,173 @@
...
+.. code-block:: edgeql-repl
+
+    db> SELECT all(<bool>{});
+    {true}
+    db> SELECT any(<bool>{});
+    {false}
+
+There's no direct analogue to the boolean operator "short-circuiting"
+that's implemented in many other languages because in EdgeQL all parts
+of the expression get evaluated. However, there are expressions that
```
elprans, line 49:
```suggestion
that's implemented in many other languages because in EdgeQL the 
order of evaluation of subexpressions is generally not defined. 
However, there are expressions that
```

vpetrovykh: Fixed

vpetrovykh:
I've also added a bit more on the gotchas of `FOR` in these similar situations.

elprans (approved):

PR #2161, opened 2021-01-28
Refactor handling of context/error span handling

tailhook:
1. Track integer byte offsets thorough code
2. Only convert when error must be printed or sent to the client
3. Use Rust for position conversion

Notes:
1. It looks like `rebase_context` is unused anymore, right?
2. No test has failed when `as_markup` was broken. Where is it used?
3. We still use strings for queries internally, using bytes instead will be in a separate PR
4. We send byte offsets to the client (there are probably two users of these offsets currently: Rust client, which is fine, and Javascript tutorial, which is unreleased yet).
5. `column` is now calculated using [unicode-width](https://crates.io/crates/unicode-width) previously was in terms of unicode codepoints. But it is unused in Rust for rendering anyway (only may show up in errors that don't show snippet), so should not break anything.
6. utf16column and char_offset are currently unused. The former is intended for language server protocol (LSP) support and the latter we might consider sending to the client so that slicing string in python is easier.

Also I think that https://github.com/edgedb/ui/commit/4b97505abeffaf909346d5ffa4f7dc407edc0573 does it wrong: it uses byte offsets sent from server as a utf-16 offsets that JS expects. @jaclarke, should we also have `utf16offset` sent from server? Or should you overlay just target line(s) and use `utf16column` for slicing?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -214,6 +214,7 @@ dependencies = [
  "snafu",
  "thiserror",
  "twoway",
+ "unicode-width",
  "wasm-bindgen",
 ]
 
@@ -576,6 +577,12 @@ name = "unchecked-index"
 version = "0.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "eeba86d422ce181a719445e51872fa30f1f7413b62becb52e95ec91aa262d85c"
+
+[[package]]
+name = "unicode-width"
+version = "0.1.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9337591893a19b88d8d87f2cec1e73fad5cdfd10e5a6f349f498ad6ea2ffb1e3"
 
 [[package]]
 name = "unicode-xid"
--- a/docs/internals/protocol/messages.rst
+++ b/docs/internals/protocol/messages.rst
@@ -126,6 +126,43 @@ Format:
 See the :ref:`list of error codes <ref_protocol_error_codes>` for all possible
 error codes.
 
+Known headers:
+
+* 0x0001 ``HINT``: ``str`` -- error hint.
+* 0x0002 ``DETAILS``: ``str`` -- error details.
+* 0x0101 ``SERVER_TRACEBACK``: ``str`` -- error traceback from server
+  (is only sent in dev mode).
+* 0xFFF1 ``POSITION_START`` -- byte offset of the start of the error span.
+* 0xFFF2 ``POSITION_END`` -- byte offset of the end of the error span.
+* 0xFFF3 ``LINE_START`` -- one-based line number of the start of the
+  error span.
+* 0xFFF4 ``COLUMN_START`` -- one-based column number of the start of the
+  error span.
+* 0xFFF5 ``UTF16_COLUMN_START`` -- zero-based column number in UTF-16
+  encoding of the start of the error span.
+* 0xFFF6 ``LINE_END`` -- one-based line number of the start of the
+  error span.
+* 0xFFF7 ``COLUMN_END`` -- one-based column number of the start of the
+  error span.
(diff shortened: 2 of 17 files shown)


elprans:
> No test has failed when as_markup was broken. Where is it used?

It is used by `debug.dump()` and we dump errors when `EDGEDB_DEBUG_SERVER=1` is active, which it is by default when running the server in devmode: `edb server`.

With this patch it's broken for me (on `SELECT Foo`, where `Foo` does not exist):

```
1. edb.errors.InvalidReferenceError: object type or alias 'default::Foo' does not exist
Exception during serialization to markup: <builtins.TypeError: cannot use a bytes pattern on a string-like object>

1. edb.errors.InvalidReferenceError: object type or alias 'default::Foo' does not exist
Exception during serialization to markup: <builtins.AttributeError: 'ParserContext' object has no attribute '_points'>
```

You can reproduce this by running the server with `EDGEDB_DEBUG_SERVER=1` (or `edb server`) and you can get to the original error here:

https://github.com/edgedb/edgedb/blob/b1eec7ed6a649a16127f66447b9df8026854e9b4/edb/common/markup/serializer/base.py#L130-L133

jaclarke:
> Also I think that edgedb/ui@4b97505 does it wrong: it uses byte offsets sent from server as a utf-16 offsets that JS expects. @jaclarke, should we also have utf16offset sent from server?

You're right, it currrently doesn't work right if the input has multibyte characters in it. If the notebook protocol is only intended to be used for interactive browser based apps like the tutorial, it probably makes sense to just send only utf16 offsets for it.

> Or should you overlay just target line(s) and use utf16column for slicing?

I'm not sure what you mean, wouldn't it still need the offsets to know how long the error span is? And isn't the column position in bytes currently?

tailhook:
> > Also I think that [edgedb/ui@4b97505](https://github.com/edgedb/ui/commit/4b97505) does it wrong: it uses byte offsets sent from server as a utf-16 offsets that JS expects. @jaclarke, should we also have utf16offset sent from server?
> 
> You're right, it currrently doesn't work right if the input has multibyte characters in it. If the notebook protocol is only intended to be used for interactive browser based apps like the tutorial, it probably makes sense to just send only utf16 offsets for it.
> 

I think we should keep protocol the same everywhere, as we don't know how it will be used. Maybe somebody wants to make a Rust/Wasm or Go/Wasm frontend, or even non-electron desktop app for the notebook.

> > Or should you overlay just target line(s) and use utf16column for slicing?
> 
> I'm not sure what you mean, wouldn't it still need the offsets to know how long the error span is? And isn't the column position in bytes currently?

`utf8column` is what is described by the language server protocol LSP (they the field `character` though). I think it is used similar to this:
```
let lines = text.split(/\r\n?|\n/);
let prefix_text = lines[span.start.line].substr(0, span.start.utf8column)
let span_text = lines[span.start.line].substr(span.start.utf8column, span.end.utf8column)
let suffix_text = lines[span.end.line].substr(span.end.utf8column)
```
(example only works for single line for brevity)

I.e. if you process text line-by-line this is enough. So the question is, given that we will need `utf16column` for LSP, do we need another field that is utf16 offset from the start of the query?

tailhook:
> With this patch it's broken for me (on SELECT Foo, where Foo does not exist):

Fixed. Thanks!

jaclarke:
Oh, the start and end are both like `{line: number, column: number}`, I was thinking of what is currently sent, which is a start and end offset, and the start line and start column number:
image

In that case `utf16column` should be enough, I don't think we need a `utf16offset`.

tailhook:
https://github.com/edgedb/edgedb/pull/2161/commits/6b14d6bc2159316592d9e56c729445c4e3e3c1ef exposes all of those positions in the binary protocol.

elprans (approved):

1st1:
@tailhook Should this be merged now?

tailhook:
Done! Sorry, for some reason I thought this is already merged.

Issue #1669, opened 2020-08-15
Nested insert of single links improperly correlated


msullivan:
#1656 fixed issues with nested inserts when the outer insert had `ON CONFLICT` or was inside a `FOR`, but the fix only works for multi links.

For single links, which don't use a link table and so have a different code path, the nested insert is executed *before* the enclosing insert.

For the `FOR` case, where currently we will do one nested insert when we should do one per FOR iteration, this doesn't pose much of a problem and just needs some programming to be done.

The `ON CONFLICT` case, where currently we will execute the nested insert even if the enclosing one fails on conflict, things are trickier: we need to put something into the link column. @elprans suggests that we randomly generate a uuid for the initial enclosing insert, then UPDATE it once the nested one is done.

msullivan:
It looks like the FOR case got fixed (probably by the big overhaul of iterator/DML interaction I did a while back), but ON CONFLICT needs to be fixed

msullivan:
Thinking about it some more, UNLESS CONFLICT preventing the execution of nested DML feels kind of fiddly and non-compositional. But also it seems really useful, so probably we need to do it anyway.

I think then, though, that things that are WITH bound outside of the INSERT need to actually be independent (and they are currently not).

For example,
```
                WITH MODULE test,
                     N := (INSERT Note {name := "tag!" }),
                FOR name in {"Phil", "Madz"} UNION (
                    INSERT Person {
                        name :=  name,
                        note := N,
                    }
                );
```
should only do one insert.
And
```
        WITH MODULE test,
             N := (INSERT Note {name := "tag!" }),
        SELECT (
            INSERT Person {
                name :=  "Test",
                notes := N,
            } UNLESS CONFLICT
        ) {name};
```
should *always* do 1.

elprans:
> (and they are currently not).

Another example where volatile materialization would save the day

PR #2162, opened 2021-01-28
Make sure that WITH bound DML statements execute the correct number of times

msullivan:
This requires compiling them when we see with binding, so the
compilation happens in the correct context. The CTE will be picked up
dml_stmts when the variable is encountered.

To make FOR work, we need to also track it in dml_stmts and reuse
iterator CTEs when necessary.

Type and pointer overlays then also need to be updated whenever a DML
containing variable is used inside other DML.


msullivan:
This fixes something I discussed in a comment to #1669 but not the core issue itself.

elprans (approved):

PR #2163, opened 2021-01-29
Improve type safety in ObjectIndex implementations

elprans:
Currently, `ObjectIndex` keys are strings, but are supposed to be
`Name` to preserve the name structure.  This happened to touch a bit more
than just names, as I had to fix some of the AST annotations to make code
areas reachable for mypy.


edb/edgeql/compiler/setgen.py
```diff
@@ -586,6 +591,7 @@ def resolve_ptr(
     track_ref: Optional[Union[qlast.Base, Literal[False]]],
     ctx: context.ContextLevel,
 ) -> s_pointers.Pointer:
+    assert isinstance(pointer_name, str)
```
msullivan, line 594: Why this? Catching some untyped bad callsites?

edb/ir/ast.py
```diff
@@ -420,8 +420,8 @@ class Set(Base):
     path_id: PathId
     path_scope_id: typing.Optional[int]
     typeref: TypeRef
-    expr: Expr
-    rptr: Pointer
+    expr: typing.Optional[Expr]
```
msullivan, line 423: Wooo

edb/pgsql/metaschema.py
```diff
@@ -2916,13 +2925,18 @@ def _generate_database_views(schema):
...
-    Role = schema.get('sys::Role')
-    member_of = Role.getptr(schema, 'member_of')
-    bases = Role.getptr(schema, 'bases')
-    ancestors = Role.getptr(schema, 'ancestors')
-    annos = Role.getptr(schema, 'annotations')
-    int_annos = Role.getptr(schema, 'annotations__internal')
+def _generate_role_views(schema: s_schema.Schema) -> List[dbops.View]:
+    Role = schema.get('sys::Role', type=s_objtypes.ObjectType)
+    member_of = Role.getptr(schema, s_name.UnqualName('member_of'))
+    assert isinstance(member_of, s_links.Link)
```
msullivan, line 2931: Would it be worth adding a `type` argument to `getptr`?

msullivan (approved):

edb/edgeql/compiler/setgen.py
elprans: Yeah.  Not necessary anymore.

edb/pgsql/metaschema.py
elprans: Yeah, that'd be better.

PR #2169, opened 2021-01-30
Implement syntax to specify a conversion expression in SET SINGLE

elprans:
When an existing link or property is converted from `multi` to
`single`, existing objects must be updated to properly reduce the
existing multi sets to a singleton value.  This implements a `USING`
clause for `SET SINGLE` (and `RESET CARDINALITY`), which acts
similarly to the analogous clause of the `SET TYPE` and `SET REQUIRED`
operations.


edb/pgsql/delta.py
```diff
@@ -2304,6 +2304,209 @@ def adjust_pointer_storage(self, pointer, schema, orig_schema, context):
...
+        self,
+        schema: s_schema.Schema,
+        orig_schema: s_schema.Schema,
+        context: sd.CommandContext,
+    ) -> None:
+        ptr = self.scls
+        ptr_stor_info = types.get_pointer_storage_info(ptr, schema=schema)
+        old_ptr_stor_info = types.get_pointer_storage_info(
+            ptr, schema=orig_schema)
+        ptr_table = ptr_stor_info.table_type == 'link'
```
1st1, line 2317:
```suggestion
        ptr_has_table = ptr_stor_info.table_type == 'link'
```

?

1st1 (approved):

PR #2170, opened 2021-02-01
A collection of fixes to schema-level aliases

elprans:


edb/edgeql/compiler/schemactx.py
```diff
@@ -203,16 +203,27 @@ def derive_view(
...
-            inheritance_refdicts={'pointers'},
-            mark_derived=True,
-            transient=True,
-            preserve_path_id=preserve_path_id,
-            attrs=attrs,
-        )
+        existing = ctx.env.schema.get(
+            derived_name, default=None, type=type(stype))
+        if existing is not None:
+            if ctx.recompiling_schema_alias:
```
msullivan, line 209: Is there a comment worth adding here?

msullivan (approved):

PR #2172, opened 2021-02-01
Add Beta 1 changelog.

vpetrovykh:


docs/changelog/1_0_b1.rst
```diff
@@ -0,0 +1,332 @@
+==========
+1.0 Beta 1
+==========
+
+This changelog summarizes new features and breaking changes in
+EdgeDB 1.0 beta 1 "Sirius".
```
1st1, line 6: Maybe we should call this Sirius A. Then the second beta or the RC can be called Sirius B. Thoughts?

docs/changelog/1_0_b1.rst
```diff
@@ -0,0 +1,332 @@
...
+
+
+Migrations
+==========
+
+One of the major improvements in Beta release is the migrations
+support (`RFC 1000 <migrations_>`_). It is now possible to manage
```
1st1, line 13: This reads strange, since migrations have been there for a while. I'd rephrase to something along the "This Beta release features the fully implemented RFC 1000 for the first time"

1st1 (commented):
We'll also need to add details about the new ports, but that can be done after this is merged (and when the ports finally land)

docs/changelog/1_0_b1.rst
vpetrovykh: That would be kinda neat, but only if we pretty much nail down the guarantee that there will be exactly 2 betas, no more no less. With more betas the naming wouldn't really be as clean then (and with only 1 beta it would be odd, too).

docs/changelog/1_0_b1.rst
vpetrovykh: Fair point, I'll update the wording.

vpetrovykh:
Should the edgedb-js version tick to the next one? I mean, it's been at `0.11.0` since alpha 6 and we do mention some improvements in this changelog.

vpetrovykh:
I've also updated the migration example to include the help the interactive tool will display for all the options.

ambv (approved):

PR #2175, opened 2021-02-01
Don't execute nested DML when UNLESS CONFLICT ON fails

msullivan:
We already avoid executing nested DML for pointers that are stored in
link fields (since those are populated *after* the insert is done),
but we need to make sure it works for pointers stored inline as well.

To make this work, we reuse the machinery for selecting conflicting
rows for the ELSE clause, but perform an anti-join on it to generate
an iterator that we use for the INSERT, containing only the rows that
won't conflict.

Progress on #1669, but we still need to support when ON is not
specified explicitly.

### DIFF
--- a/edb/edgeql/compiler/stmt.py
+++ b/edb/edgeql/compiler/stmt.py
@@ -314,23 +314,23 @@ def compile_insert_unless_conflict(
 
     ctx.env.schema = schema
 
-    # Compile an else branch
-    else_info = None
-    if else_branch:
-        # Produce a query that finds the conflicting objects
-        nobe = qlast.SelectQuery(
-            result=insert_subject,
-            where=qlast.BinOp(
-                op='=',
-                left=constraint_spec,
-                right=anchor
-            ),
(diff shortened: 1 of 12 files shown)


edb/edgeql/compiler/stmt.py
```diff
@@ -314,23 +314,23 @@ def compile_insert_unless_conflict(
 
     ctx.env.schema = schema
 
+    # Produce a query that finds the conflicting objects
+    nobe = qlast.SelectQuery(
```
elprans, line 318: `node`?

tests/schemas/insert.esdl
```diff
@@ -51,6 +51,9 @@ type Person {
         default := "Nemo";
     };
     optional single property tag -> std::str;
+    required single property tag2 -> std::str {
+	default := "<n/a>";
```
elprans, line 55: Indentation seems weird

elprans (approved):

PR #2178, opened 2021-02-02
Fix lower cardinality checks on multi pointers

elprans:
The `required` constraint is currently not being enforced properly on
`multi` links and properties in cases where the value expression in
`INSERT` or `UPDATE` returns an empty set.  Fix this by treating the value
expression as `OPTIONAL` and check for `NULL` explicitly.

### DIFF
--- a/edb/pgsql/compiler/dml.py
+++ b/edb/pgsql/compiler/dml.py
@@ -69,6 +69,10 @@ class DMLParts(NamedTuple):
 
     range_cte: Optional[pgast.CommonTableExpr]
 
+    #: A list of CTEs that implement constraint validation at the
+    #: query level.
+    check_ctes: List[pgast.CommonTableExpr]
+
 
 def init_dml_stmt(
     ir_stmt: irast.MutatingStmt,
@@ -156,6 +160,7 @@ def init_dml_stmt(
         dml_ctes=dml_map,
         range_cte=range_cte,
         else_cte=else_cte,
+        check_ctes=[],
     )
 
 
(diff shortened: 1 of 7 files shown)


edb/pgsql/compiler/dml.py
```diff
@@ -372,6 +375,24 @@ def fini_dml_stmt(
 
     relctx.include_rvar(ctx.rel, union_rvar, ir_stmt.subject.path_id, ctx=ctx)
 
+    for check_cte in parts.check_ctes:
```
msullivan, line 378: Add a comment?

edb/pgsql/compiler/dml.py
```diff
@@ -1068,7 +1101,31 @@ def process_link_update(
         delqry = None
 
     if shape_op is qlast.ShapeOp.SUBTRACT:
-        return data_cte
+        prev_expr = ir_set.expr
+        ir_set.expr = None
+
+        # Turn the IR of the expression on the right side of :=
```
msullivan, line 1107: I think this setup needs some explanation because I had to spend a bunch of time puzzling on it and I'm not 100% sure I'm right:
1. We clear the expr on the ir_set so that it just represents a pointer to that element
2. We add an overlay to filter out the things we're deleting
3. We "recompile" the ir_set to compute the rows that will be remaining
4. We fix up the ir_set

Mucking around with the ir_set strikes me as kind of scary, so it definitely should be called out clearly, at least.

msullivan: (Should we only be doing this when we are actually enforcing the constraint?)

edb/pgsql/compiler/dml.py
msullivan: I don't think I understand why we need to collect the counts in the enclosing query

elprans: We need to force the evaluation to trigger a call to `raise`, since nothing else depends on the check CTE.  Will add a comment.

edb/pgsql/compiler/dml.py
elprans: Ugh, it was a quick hack, and it totally slipped my mind to fix this properly.  Your analysis is correct, and I'll add that as a comment.

msullivan (approved):
Should be good to go once comments are added

PR #2181, opened 2021-02-03
Don't execute nested DML when UNLESS CONFLICT fails

msullivan:
We accomplish this by synthesizing a SELECT that matches all
conflicting objects based on inspection of the exclusive constraints
in the object schema.

While I'm at it, I also made it work when there are subjectexprs,
which was an outstanding bug even with ON.

This still doesn't support object constraints, which will be a little
more hairy but still should be doable; I'll tackle that in a
follow-up.

Most of the way on #1669.

### DIFF
--- a/edb/edgeql/compiler/stmt.py
+++ b/edb/edgeql/compiler/stmt.py
@@ -25,9 +25,11 @@ from typing import *
 
 from collections import defaultdict
 import textwrap
+import copy
 
 from edb import errors
 from edb.common import context as pctx
+from edb.common.ast import visitor as ast_visitor
 
 from edb.ir import ast as irast
 from edb.ir import typeutils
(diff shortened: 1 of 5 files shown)


edb/edgeql/compiler/stmt.py
```diff
@@ -230,7 +232,146 @@ def compile_GroupQuery(
         context=expr.context)
 
 
+TBase = TypeVar('TBase', bound=qlast.Base)
```
elprans, line 235: Perhaps call this `Base_T` and move to `edgeql.ast`?  I think this is the approach we settled on with respect to typevars.

edb/edgeql/compiler/stmt.py
```diff
@@ -230,7 +232,146 @@ def compile_GroupQuery(
...
+
+
+def subject_substitute(ast: TBase, new_subject: qlast.Expr) -> TBase:
+    ast = copy.deepcopy(ast)
```
elprans, line 239: Can this be done by setting up a `__subject__` anchor before compiling the node instead?

elprans (approved):

edb/edgeql/compiler/stmt.py
msullivan: Yes and no:
 * Yes, but we have to use the subjectexpr twice, on each side of the equation, so we'd have to do it by compiling it twice, once with each side as an anchor, and then putting the results of those in as anchors. I think it would work just fine but it's not really any simpler, I guess. (Well, not any less code. Maybe simpler.)
 * But handling object constraints is going to require a souped-up version of the substitution approach, since we'll need to be able to pull things out of the shape based on path references

msullivan: (So my plan is to stick with the current approach)

PR #2183, opened 2021-02-03
Move the compiler state to the IO process

1st1:


edb/server/mng_port/edgecon.pyx
```diff
@@ -1663,7 +1681,8 @@ cdef class EdgeConnection:
             # in this situation we just silently exit.
             self.abort()
 
-        except ConnectionAbortedError:
+        except (ConnectionAbortedError, pgerror.BackendQueryCancelledError):
```
elprans, line 1684: Postgres might cancel on its own, or it might be external cancellation, should we always `abort()`?  Maybe only if `self._cancelled`?

1st1: > Postgres might cancel on its own

How and why?

elprans (approved):

edb/server/mng_port/edgecon.pyx
elprans: Due to a timeout `statement_timeout` or `lock_timeout`, and also on a hot standby node when the async recovery process conflicts with the query.

1st1: Alright, I've fixed it to only abort when we're cancelling the protocol.

edb/pgsql/delta.py
```diff
@@ -281,6 +283,172 @@ class AlterObjectProperty(MetaCommand, adapts=sd.AlterObjectProperty):
...
+                FROM
+                    pg_catalog.pg_shdescription
+                WHERE
+                    objoid = (
+                        SELECT oid
+                        FROM pg_database
+                        WHERE datname = {ql(edbdef.EDGEDB_TEMPLATE_DB)}
+                    )
+                    AND classoid = 'pg_database'::regclass::oid
+                FOR UPDATE
```
fantix, line 403: Looks like AWS RDS doesn't allow locking the `pg_shdescription` catalog (but it does allow `COMMENT` on them):

```sql
postgres=> SELECT * FROM pg_shdescription FOR UPDATE;
ERROR:  permission denied for table pg_shdescription
```

I couldn't find an AWS reference for this limit. This is blocking EdgeQL queries like `ALTER ROLE ... SET password := ...`, and causing some relevant tests to fail. Would it be a problem if we don't lock it pls?

elprans: Sigh.  We use the lock to prevent concurrent modification of the global schema.  We cannot simply remove the lock, need to find another way.

edb/server/dbview/dbview.pyx
```diff
@@ -291,15 +389,31 @@ cdef class DatabaseConnectionView:
             self._invalidate_local_cache()
 
         if not self._in_tx:
-            if query_unit.has_ddl:
-                self._db._signal_ddl(None)
+            if new_types:
+                self._db._update_backend_ids(new_types)
+            if query_unit.user_schema is not None:
+                self._in_tx_dbver = next_dbver()
```
fantix, line 395: I think `_in_tx_dbver` has been wrong here.

PR #2185, opened 2021-02-04
Improve "Unexpected token" type of error message.

vpetrovykh:
There are some error patterns where "unexpected token" is better
expressed as a different error message.

### DIFF
--- a/edb/edgeql/parser/parser.py
+++ b/edb/edgeql/parser/parser.py
@@ -21,7 +21,14 @@ from __future__ import annotations
 from edb import errors
 from edb.common import debug, parsing
 
-from .grammar import rust_lexer
+from .grammar import rust_lexer, tokens
+from .grammar import expressions as gr_exprs
+from .grammar import commondl as gr_commondl
+
+
+def a(word):
+    article = 'an' if word[0] in 'aeiou' else 'a'
+    return f'{article} {word}'
 
 
 class EdgeQLParserBase(parsing.Parser):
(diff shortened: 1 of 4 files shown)


edb/edgeql/parser/parser.py
```diff
@@ -36,9 +37,64 @@ def get_exception(self, native_err, context, token=None):
...
+                    # an attempt to distinguish some specific parsing
+                    # error in shapes.
+
+                    # Then check if LBRACE, LPAREN, or LBRACKET is in
+                    # the parser stack. Any one of those indicates
+                    # that the parser was in the middle of processing
+                    # a list of some sort (shape, tuple, or array) and
+                    # encountering an unexpected IDENT probably
+                    # indicates a missing COMMA instead.
+                    for i, (el, _) in enumerate(reversed(self.parser._stack)):
```
1st1, line 64: Why do we need a `for` here? Looks like we break in almost every branch. Also in the first iteration of this loop, `el` would be the same as `ltok`, right? So maybe there's no need in looping here?

edb/edgeql/parser/parser.py
```diff
@@ -36,9 +37,64 @@ def get_exception(self, native_err, context, token=None):
...
+                                    ltok.val in keywords.edgeql_keywords
+                                ):
+                                    # This is some [unreserved]
+                                    # keyword preceding the unexpected
+                                    # IDENT in a shape, which probably
+                                    # is not a case of a missing COMMA
+                                    # anymore.
+                                    break
+                                else:
+                                    msg = (f"Missing ',' before the shape "
```
1st1, line 78: What about `SELECT {1 2}`? It's not a shape.

1st1 (commented):
Love the new error messages!

elprans:
Hm, this approach feels flimsy and prone to get outdated with respect to the rest of the grammar.  A misplaced specific "missing ',' before <foo>" is much more confusing than a generic "unexpected <IDENT>".

1st1:
> Hm, this approach feels flimsy and prone to get outdated with respect to the rest of the grammar.

Yeah, I also don't like introspecting the token stream. It would be better to try implementing this directly in the grammar. Did you try that Victor? If you did, what specific conflicts were there (so that I can think of how we could avoid them)

edb/edgeql/parser/parser.py
vpetrovykh: You're right. I was just focusing on the unexpected `IDENT` errors, but I can reframe it so that we check for this, too, by swapping the order of ifs and looking at the parsing stack first to determine the context and then narrowing it down by considering the unexpected token itself. Since this is in the error path, the tiny overhead of looping through the stack first shouldn't matter.

vpetrovykh:
> > Hm, this approach feels flimsy and prone to get outdated with respect to the rest of the grammar.
> 
> Yeah, I also don't like introspecting the token stream. It would be better to try implementing this directly in the grammar. Did you try that Victor? If you did, what specific conflicts were there (so that I can think of how we could avoid them)

Yes, I tried a few places for false rules that catch `IDENT IDENT`, but that was producing ambiguity issues, which, we already know is a problem that conflicts with unreserved keywords (e.g. `required foo := 1`).

This is as brittle as grammar itself, which we tend not to refactor, but rather just add things to usually. Refactorings have been relatively few historically. And syntax tests checking for the errors would fails and indicate a problem in future refactorings. Essentially, I can make it more reliable by checking for specific non-terminals a bit more on the specialized error paths. This way it's far more likely that any change to the grammar that disrupts/renames non-terminals will simply result in swapping the specific error for generic.

elprans:
> This is as brittle as grammar itself, which we tend not to refactor, but rather just add things to usually. Refactorings have been relatively few historically.

Well, the problem is exactly that: someone adds productions that your `if` scaffold will start caching as a false positive.  In my opinion, unless we can improve parsing formally with productions, we shouldn't attempt to build ad-hoc "parsers" on the bare token stream.

vpetrovykh:
In order to make the ifs less susceptible to false positives I should pin determination of specific error paths to non-terminals. Because unlike tokens they carry more semantics and are less likely to result in coincidences.

Adding false rules to the grammar just to catch errors is generally a bad idea because false rules are overwhelmingly likely to produce ambiguities (because they represent variations of the valid grammar) and once added can be a pain to maintain.

Broadly speaking I don't see a particularly big problem in having an error processor that uses some knowledge of the underlying grammar to narrow down generic errors.

1st1:
Victor, what's the latest?

vpetrovykh:
I've updated the error message algorithm to rely on non-terminals in addition to tokens. There are more cases handles for "Missing comma" situations (including differentiating between function calls and tuples). The case when a statement expression is not wrapped in parentheses properly inside a function call is also handled.

tests/test_edgeql_syntax.py
```diff
@@ -1455,7 +1455,7 @@ def test_edgeql_syntax_shape_43(self):
         """
 
     @tb.must_fail(errors.EdgeQLSyntaxError,
-                  r"Unexpected 'foo'", line=3, col=27)
+                  r"Missing ',' before the shape item 'foo'", line=3, col=27)
```
elprans, line 1458: I don't think the definite article should be used in any of these messages, because, especially in the case of tuples and arrays, there may be multiple items `'foo'`.

tests/test_schema_syntax.py
```diff
@@ -1521,6 +1530,37 @@ def test_eschema_syntax_alias_03(self):
...
+            );
+        };
+        """
+
+    @tb.must_fail(errors.EdgeQLSyntaxError,
+                  r"Missing ',' before the tuple item '2'", line=3, col=32)
+    def test_eschema_syntax_alias_05(self):
+        """
+        module test {
+            alias FooBaz := (1 2);
```
elprans, line 1551: Are you sure it's a tuple and not just a fat-fingered parenthesized expr?

1st1: Good point. We should rewrite error messages to use "if", as in "hey, if this meant to be a tuple you'd want to have a comma between 1 and 2", or something along these lines.

tests/test_edgeql_syntax.py
vpetrovykh: I've reworded the error messages.

tests/test_schema_syntax.py
vpetrovykh: I've reworded the error messages to be in the form of "It appears that ..." since that is still accurate. Now the message states what the parser seems to infer (based on Nonterminals that's what the parser was trying to match). So whatever the human intent might have been, but the actual text appears to be <blah>. This phrasing should be helpful either way: if the intent matches, then the hint is relevant, if the intent is surprising, then the error indicates what the parser sees and that can be helpful in figuring out what went wrong in that chunk of code, too.

edb/edgeql/parser/parser.py
```diff
@@ -36,12 +43,144 @@ def get_exception(self, native_err, context, token=None):
...
+                elif (
+                    rule == 'shape' and
+                    token_kind == 'IDENT' and
+                    isinstance(ltok, parsing.Nonterm)
+                ):
+                    # Make sure that the previous element in the stack
+                    # is some kind of Nonterminal, because if it's
+                    # not, this is probably not an issue of a missing
+                    # COMMA.
+                    msg = (f"It appears that a ',' is missing in {a(rule)} "
```
elprans, line 125: We start error messages with a lowercase letter, because they are often prefixed with "Error: ".  Also, why does this have to be an error message and not a hint?

elprans (approved):

PR #2188, opened 2021-02-05
Allow omitting `-> type` in SDL block decls of computed pointers

msullivan:
This matches how DDL works and the existing output of
`describe ... as sdl`.

I removed some logic where the USING got dropped from the block of
commands because it messed up parse/codegen tests and doesn't seem to
cause any trouble on the schema side.


elprans (approved):
:+1:

1st1 (approved):

PR #2189, opened 2021-02-05
Make UNLESS CONFLICT's handling of nested DML work with object/computed cnstrs

### LINKED ISSUE #1669: Nested insert of single links improperly correlated

msullivan:
We do this by collecting all the pointers that are actually needed and
then substuting them into the definition of computed pointers/object
constraints.

Also fix a silly bug where we would generate bogus SQL DDL if a
constraint key contained a space.

Fixes #1669.

I think we could make it so that ELSE works for UNLESS CONFLICT
without needing ON, now. I think it's probably worth it to do that.

### DIFF
--- a/edb/pgsql/deltadbops.py
+++ b/edb/pgsql/deltadbops.py
@@ -206,10 +206,12 @@ class SchemaConstraintTableConstraint(ConstraintCommon, dbops.TableConstraint):
                           SCHEMA = '{schemaname}',
                           CONSTRAINT = '{constr}',
                           MESSAGE = '{errmsg}',
-                          DETAIL = 'Key ({plain_expr}) already exists.';
+                          DETAIL = 'Key ({esc_plain_expr}) already exists.';
                 END IF;
             '''.format(
                 plain_expr=origin_exprdata['plain'],
+                # XXX: do we want plain_expr here at all??
+                esc_plain_expr=origin_exprdata['plain'].replace("'", "''"),
                 new_expr=exprdata['new'],
                 table=common.qname(
                     schemaname,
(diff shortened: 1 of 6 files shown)


edb/pgsql/deltadbops.py
```diff
@@ -206,10 +206,12 @@ def get_trigger_proc_text(self):
                           SCHEMA = '{schemaname}',
                           CONSTRAINT = '{constr}',
                           MESSAGE = '{errmsg}',
-                          DETAIL = 'Key ({plain_expr}) already exists.';
+                          DETAIL = 'Key ({esc_plain_expr}) already exists.';
                 END IF;
             '''.format(
                 plain_expr=origin_exprdata['plain'],
+                # XXX: do we want plain_expr here at all??
+                esc_plain_expr=origin_exprdata['plain'].replace("'", "''"),
```
elprans, line 214: Use `pgsql.common.quote_literal` instead and concatenate `DETAIL` instead of interpolating directly.

tests/schemas/insert.esdl
```diff
@@ -64,6 +64,26 @@ type Person {
...
+type Person2a extending Person2 {
+    required single property last -> std::str;
+    constraint exclusive on ((__subject__.first, __subject__.last));
+}
+
+type Person2b extending Person2 {
+    optional single property last -> std::str;
+    property namespace := .first ++ " "; # har, har
+    property name {
+	using (.namespace ++ .last);
```
elprans, line 82: Broken indentation here

elprans (approved):

PR #2197, opened 2021-02-06
Fix type and cardinality inference for UNLESS CONFLICT

msullivan:
Also fix doing an nested INSERT in the ELSE clause, which I ran into
when testing cardinality inference.

### DIFF
--- a/edb/edgeql/compiler/inference/cardinality.py
+++ b/edb/edgeql/compiler/inference/cardinality.py
@@ -1035,15 +1035,24 @@ def __infer_insert_stmt(
         ir.result, is_mutation=True, scope_tree=new_scope, ctx=ctx
     )
 
-    if ir.on_conflict:
-        for part in [ir.on_conflict.select_ir, ir.on_conflict.else_ir]:
-            if part:
-                infer_cardinality(part, scope_tree=scope_tree, ctx=ctx)
-
     assert not ir.iterator_stmt, "InsertStmt shouldn't ever have an iterator"
 
     # INSERT without a FOR is always a singleton.
-    return ONE
+    if not ir.on_conflict:
+        return ONE
+    # ... except if UNLESS CONFLICT is used
+    else:
+        if ir.on_conflict.select_ir:
+            infer_cardinality(
+                ir.on_conflict.select_ir, scope_tree=scope_tree, ctx=ctx)
+
+        card = AT_MOST_ONE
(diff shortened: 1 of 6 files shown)


edb/edgeql/compiler/inference/cardinality.py
```diff
@@ -1035,15 +1035,24 @@ def __infer_insert_stmt(
...
     assert not ir.iterator_stmt, "InsertStmt shouldn't ever have an iterator"
 
     # INSERT without a FOR is always a singleton.
-    return ONE
+    if not ir.on_conflict:
+        return ONE
+    # ... except if UNLESS CONFLICT is used
+    else:
+        if ir.on_conflict.select_ir:
+            infer_cardinality(
```
1st1, line 1046: Should it be `card = infer_cardinality(...`?

msullivan: Aha! I had thought that we didn't need to worry about the cardinality of the select_ir because it would always be AT_MOST_ONE, but now that we support select_irs that look at multiple different exclusive constraints, it can be MANY!

`card = infer_cardinality(...` isn't quite what we need to do, since it shouldn't have any effect at all if there is no `ELSE`, but we do need to take it into account

Good catch!

tests/test_edgeql_insert.py
```diff
@@ -2264,43 +2264,113 @@ async def test_edgeql_insert_unless_conflict_01(self):
...
+                edgedb.QueryError,
+                "possibly more than one element returned by an expression "
+                "for a computable link 'foo' declared as 'single'"):
+            await self.con.query(r'''
+                WITH MODULE test,
+                    X := (INSERT Person {
+                        name := "Test",
+                        case_name := "Foo",
+                    }
+                UNLESS CONFLICT ELSE Person),
```
elprans, line 2369: I see why `UNLESS CONFLICT ELSE SELECT` returns `MANY`, but I'm not 100% sure it's the best semantics.  Intuitively, any `INSERT` is expected to be `AT_MOST_ONE` and I see potential for confusion from errors raised in queries like this.

Perhaps we should rethink allowing `UNLESS CONFLICT ELSE` without an explicit conflict spec?  Postgres bans bare `ON CONFLICT DO UPDATE` for this reason.

msullivan: I dropped it for now

elprans (approved):

PR #2198, opened 2021-02-06
Catch bad SDL definitions.

### LINKED ISSUE #2179: Bogus names and references in migrations are reported as ISEs
```
start migration to {
module test {
type Foo extending Bar;
}
};
```
is reported as `edb.common.topological.UnresolvedReferenceError: reference to an undefined item test::Bar in test::Foo`

and 
```
start migration to {
module test {
type Foo extending Foo;
}
};
```
is reported as a RecursionError

vpetrovykh:
Catch definitions referring to non-existing names or recursive
definitions early and provide an appropriate error message for them.

Fixes #2179

### DIFF
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -150,6 +150,29 @@ _123456789_123456789_123456789 -> str
         """
 
     @tb.must_fail(errors.InvalidReferenceError,
+                  "property 'test::bar' does not exist",
+                  position=59)
+    def test_schema_bad_prop_04(self):
+        """
+            abstract property foo extending bar;
+        """
+
+    @tb.must_fail(errors.InvalidReferenceError,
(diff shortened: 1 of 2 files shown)


tests/test_schema.py
```diff
@@ -149,6 +149,29 @@ def test_schema_bad_prop_03(self):
             };
         """
 
+    @tb.must_fail(errors.InvalidReferenceError,
+                  "pointer 'test::bar' does not exist",
```
elprans, line 153: can we use the real class name (i.e. 'link' or 'property') instead of 'pointer' here?

vpetrovykh: OK, I've updated the message to refer tot he specific type.

elprans (approved):

PR #2208, opened 2021-02-10
Stop labelling CLI examples as "bash" code blocks.

vpetrovykh:
The bash code blocks end up with an odd higlighting scheme, which is
unhelpful.

### DIFF
--- a/docs/changelog/1_0_b1.rst
+++ b/docs/changelog/1_0_b1.rst
@@ -38,7 +38,7 @@ We create a directory ``app_schema`` and write the above into a
 ``app_schema/schema.esdl`` file inside it. Then we can initialize our
 project database by running ``edgedb create-migration``:
 
-.. code-block:: bash
+.. code-block::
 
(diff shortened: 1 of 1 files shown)


1st1:
I'm not so sure about this. `bash` language gets special treatment in our UI code -- every line is separate and has a proper "COPY" functionality. @jaclarke 

I'd just remove the code highlighter for "bash" and keep our docs ReST files as is.

jaclarke:
What's the problem here? None of the bash code blocks here have anything highlighted anyway: https://edgedb-staging.vercel.app/docs/changelog/1_0_b1

Though there are a few places with odd highlighting like 'install' in the last block on this page: https://edgedb-staging.vercel.app/docs/tutorial/install

vpetrovykh:
This change was done in response to https://github.com/edgedb/edgedb-site/pull/66

Maybe highlighting changed between old and new site rendering?

vpetrovykh:
It appears that this change is not needed.

PR #2210, opened 2021-02-11
Hack: silence warnings from calling transaction()

msullivan:
edgedb-python has been updated to mark transaction() as deprecated,
which makes our test suite cry bloody murder. We should probably
actually fix this, but for now I just supressed the warnings.


msullivan:
(This didn't get anywhere near all of them, whoops)

PR #2211, opened 2021-02-11
Remove schema objects created as reflection proxies on deletes

msullivan:
Leaving this around is incorrect, and combined with other factors (we
don't seem to fire link deletion triggers in the schema tables?)  and
the use of deterministic IDs for collection types to produce truly
bizarre results when a collection type is deleted and then recreated.

### DIFF
--- a/edb/schema/reflection/writer.py
+++ b/edb/schema/reflection/writer.py
@@ -956,7 +956,7 @@ def write_meta_delete_object(
 
             blocks.append((parent_update_query, parent_variables))
 
-        # We need to delete any links create via reflection_proxy
+        # We need to delete any links created via reflection_proxy
         layout = classlayout[mcls]
         proxy_links = [
             link for link, layout_entry in layout.items()
--- a/edb/schema/reflection/writer.py
+++ b/edb/schema/reflection/writer.py
@@ -956,9 +956,27 @@ def write_meta_delete_object(
 
             blocks.append((parent_update_query, parent_variables))
 
+        # We need to delete any links create via reflection_proxy
+        layout = classlayout[mcls]
+        proxy_links = [
+            link for link, layout_entry in layout.items()
+            if layout_entry.reflection_proxy
+        ]
+
+        operations = []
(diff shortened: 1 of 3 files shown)


edb/schema/reflection/writer.py
```diff
@@ -956,9 +956,27 @@ def write_meta_delete_object(
...
+        # We need to delete any links create via reflection_proxy
+        layout = classlayout[mcls]
+        proxy_links = [
+            link for link, layout_entry in layout.items()
+            if layout_entry.reflection_proxy
+        ]
+
+        operations = []
+        if proxy_links:
+            # Link deletion triggers apparently don't work for the
```
elprans, line 968: It's not that triggers don't fire: we don't _have_ a forward cascade logic at all (i.e. `ON SOURCE DELETE DELETE TARGET`).

edb/schema/reflection/writer.py
```diff
@@ -956,9 +956,27 @@ def write_meta_delete_object(
 
             blocks.append((parent_update_query, parent_variables))
 
+        # We need to delete any links create via reflection_proxy
```
elprans, line 959:
```suggestion
        # We need to delete any links created via reflection_proxy
```

elprans (approved):

edb/schema/reflection/writer.py
msullivan: I think you missed my meaning here: we *do* have triggers for deleting entries from the link table when the source is deleted (but not for deleting the target object, like you said), and that's not working here, which results in leaving stray dangling entries in the link table).

elprans: Oh, that is unexpected.  There is no exception for system tables, they're treated exactly like user schema, so we might be papering over a bug here.

msullivan: Yeah, you are right, I was sloppy about this.

I dug into it and there are actually a couple of bugs involving inheritance and trigger tables.

PR #2126, opened 2021-01-18
Add intro + link descriptions

Dhghomon:
Feel free to take an axe to anything here

PR #2212, opened 2021-02-11
Revert cheatsheets title + make new title an alt title

jaclarke:
New cheatsheets title in #2126 is too long for the docs nav:
image

This reverts the title and makes the new title an alt title instead:
image


1st1 (approved):

1st1:
good catch

PR #2215, opened 2021-02-12
Multiplex binary protocol and HTTP via the same network interface

1st1:
The only missing part right now is a proper HTTP upgrade implementation. I'll deal with it in a follow up PR.


edb/server/protocol/protocol.pyx
```diff
@@ -175,5 +209,33 @@ cdef class HttpProtocol:
...
+        path = urllib.parse.unquote(request.url.path.decode('ascii'))
+        path = path.strip('/')
+        path_parts = path.split('/')
+
+        if len(path_parts) >= 3:
+            root, dbname, extname, *args = path_parts
+            db = self.server.get_db(dbname=dbname)
+
+            if root == 'db':
+                if extname == 'graphql':
```
elprans, line 222: I assume we will use some extension registry here instead of hardcoding, right?

elprans (approved):

edb/server/protocol/protocol.pyx
1st1: Yea, once we have it.

PR #2217, opened 2021-02-12
Run test workflow in shards

### LINKED ISSUE #2201: Improve test workflow performance
The current runtime of the test Github workflow has grown quite a bit, because we execute all tests in a single job (with `-j2`).  This can be improved by splitting the test suite into chunks that can then be executed in concurrent jobs.  Instance bootstrap is also quite slow and cannot be parallelized, so it'll have to be executed (and cached) in a separate job.

fantix:
This PR adds 3 new options to `edb test`:

* `edb test --shard 2/5` separates all collected tests into 5 shards, and run the 2nd shard only. The split is supposed to be deterministic.
* `edb test --time-file` generates a CSV file of time spent on each test. If the file exists, it'll be taken as an input for `--shard` to generate a more even split, and the CSV file will be updated after the run with new average running times.
* `edb test --list` outputs a list of all selected tests and exits.

The pipeline will fetch from a Gist for `--time-file`, and update that Gist if the tests pass in the master branch.

- [x] Replace the Gist URL with a new one by @edgedb-ci 
- [x] Add @edgedb-ci 's Github Personal Access Token with `gist` scope as Secret `GIST_TOKEN`
- [x] Update the branch protection rules to check for the `verify` job (or update the `test.yml` to rename the final job back to `test`?)

Fixes #2201

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -13,7 +13,7 @@ on:
     - cron: "0 */3 * * *"
 
 jobs:
-  test:
+  build:
     runs-on: ubuntu-18.04
 
     steps:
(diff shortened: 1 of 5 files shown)


1st1:
I wonder why it says "Some checks haven’t completed yet" here in the PR status, and yet everything is green in the Action tab? Also, maybe we run 32 jobs? :)

fantix:
Oh the 3rd TODO is for your first question:

>  Update the branch protection rules to check for the verify job (or update the test.yml to rename the final job back to test?)

(let me rename the `verify` job to `test`)

And sure, we can try 32 jobs ;) It'll be roughly 10 minutes I guess.

elprans:
@1st1 

> I wonder why it says "Some checks haven’t completed yet"

Because the workflow job which is listed in the branch protection rules got replaced.

@fantix 

> * Replace the Gist URL with a new one by @edgedb-ci

https://api.github.com/gists/8b722a65397f7c4c0df72f5394efa04c

> * Add @edgedb-ci 's Github Personal Access Token with `gist` scope as Secret `GIST_TOKEN`

Done.  Added as `CI_BOT_GIST_TOKEN` org secret.

edb/tools/test/__init__.py
```diff
@@ -80,8 +83,13 @@
               help='package name to measure code coverage for, '
                    'can be specified multiple times '
                    '(e.g --cov edb.common --cov edb.server)')
-def test(*, files, jobs, include, exclude, verbose, quiet, debug,
-         output_format, warnings, failfast, shuffle, cov, repeat):
+@click.option('--time-file', type=click.File('a+'), metavar='FILEPATH',
```
elprans, line 86: I'd call this `--log-running-times`

elprans: Or, rather `--running-times-log`

fantix: Oh actually it also reads this file for calculating distribution - should we create a separate option for reading?

1st1: Why a+?

edb/testbase/server.py
```diff
@@ -1505,3 +1510,65 @@ def start_edgedb_server(
...
+    if total_shards <= 1:
+        return cases
+
+    current_shard -= 1
+    new_test_est = 0.1
+    tests_with_est = []
+    new_tests = []
+    total_tests = 0
+    selected_tests = 0
+    for tests in cases.values():
```
elprans, line 1525: I wonder if we can get better performance by avoiding the splits of test suites (classes) as much as possible.  The setup for a test suite might be expensive.  We can time and record `_setup_database` calls to guide this decision,

edb/tools/test/__init__.py
fantix: It's both for reading and writing, and I'd like the file to be created if not exist. `r+` doesn't create the file, and `w+` overwrites the file on open.

edb/testbase/server.py
fantix: Yes, I thought about that and had some early patches to group test suites by the 60+ different database setups, or leave the non-database tests in a separate group. Then I became lazy and went for an easier way to get some results first. Now that we're here, I'm totally open to more attempts.

fantix:
> It'll be roughly 10 minutes I guess.
I was too optimistic - 11m13s. Shall we try harder - group by the database setup, cache virtualenv, etc?

.github/workflows/tests.yml
```diff
@@ -67,22 +87,316 @@ jobs:
...
+      with:
+        profile: minimal
+        toolchain: stable
+        default: true
+
+    - name: Install
+      run: |
+        # --no-use-pep517 because we have explicitly installed all deps
+        # and don't want them to be reinstalled in an "isolated env".
+        pip install --no-use-pep517 --no-deps -e .[test,docs]
```
elprans, line 244: I wonder if we could cache the entire virtualenv too to avoid doing this step in shards?

elprans (approved):
I think we can merge this as-is and continue working on possible improvements in a subsequent PR.  The improvement here is already substantial.

fantix:
Sounds good to me! I was trying the virtualenv cache this morning, will send another PR if that works. I'll merge this one now, thanks!

elprans:
@fantix, it seems like each shard is doing expensive instance bootstrap.  I think the cache check in the `build` step is incorrect, since it skips "Bootstrap EdgeDB Server" when it shouldn't.  `gen_build_cache_key` only stamps Postgres version and the grammars.  We need to add another key to track whether full instance bootstrap is needed.  This is controlled by src_hash = buildmeta.hash_dirs(CACHE_SRC_DIRS, extra_files=[__file__])` in `bootstrap.py`.

PR #2220, opened 2021-02-12
Update the documentation for links.

### LINKED ISSUE #967: Documentation on links is unclear
I'm looking at https://edgedb.com/docs/datamodel/links:

1. There is no single example of how links are defined
2. I think linking from SDL to DDL for explanation is ugly. If I understand correctly, most users should never use DDL directly.
3. Docs don't mention backwards link. I.e. if have `Ticket { link owner -> User; }` should I also define `User { multi link tickets -> Ticket }` or not?
4. What indexes are provided on the link by default and what indexes you need to define yourself? (I.e. if no ones provided then user needs indexes in 99.9% cases). 
5. There should be the discussion of forward vs backward links: I.e. whether `User` needs a multi link to `Session`, or whether `Session` needs a single link to `User` (probably we can come up with more controversial example).

**Update:** added (5)

vpetrovykh:
Cover in more detials the one-to-one, one-to-many, and many-to-many
scenarios as well as the use of backward link traversal.

Fixes #967

### DIFF
--- a/docs/datamodel/links.rst
+++ b/docs/datamodel/links.rst
@@ -4,7 +4,7 @@
 Links
 =====
 
-:index: link
+:index: link one-to-one one-to-many many-to-one many-to-many
 
 Link items define a specific relationship between two :ref:`object
 types <ref_datamodel_object_types>`.  Link instances relate one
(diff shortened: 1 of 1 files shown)


docs/datamodel/links.rst
```diff
@@ -14,16 +14,265 @@ There are two kinds of link item declarations: *abstract links*, and
...
 not tied to any particular object type. Typically this is done to set
 some :ref:`annotations <ref_datamodel_annotations>`, define
-:ref:`link properties <ref_datamodel_props>`, or setup :ref:`constraints
-<ref_datamodel_constraints>`.  Concrete links are defined on specific object
-types.
+:ref:`link properties <ref_datamodel_props>`, or setup
+:ref:`constraints <ref_datamodel_constraints>`.  Concrete links
+are defined on specific object types.
 
 Links are directional and have a *source* and one or more *targets*.
```
elprans, line 21: Links don't have multiple targets.  Rather, a single *source* may have multiple *links* of a kind.  It is easy to see the difference when you take link properties into account.

docs/datamodel/links.rst
```diff
@@ -14,16 +14,265 @@ There are two kinds of link item declarations: *abstract links*, and
...
+possible to specify common relationships between things:
+*one-to-one*, *one-to-many*, and *many-to-many*.
+
+
+One-to-One
+----------
+
+A *one-to-one* relationship represents a situation where everything
+is broken up into pairs and all the pairs are unique. For example,
+``Employee`` and ``ReservedParking``:
```
elprans, line 36:
```suggestion
A *one-to-one* relationship represents a situation where one object from 
a source set is linked to only one object in the target set, and vice versa.
For example, ``Employee`` and ``ReservedParking``:
```

docs/datamodel/links.rst
```diff
@@ -14,16 +14,265 @@ There are two kinds of link item declarations: *abstract links*, and
...
+    WITH Spot := (
+        SELECT ReservedParking FILTER .number = 42
+    )
+    SELECT Spot.<parking[IS Employee] {
+        name
+    };
+
+Backward link traversal requires to specify the original link's source
+type, but other than that it works the same way as forward traversal.
+
```
elprans, line 86: What about `Many-to-One`?

docs/datamodel/links.rst
```diff
@@ -14,16 +14,265 @@ There are two kinds of link item declarations: *abstract links*, and
...
+instead represents situations like ownership or hierarchies. For
+example, ``Person`` and ``Shoes``:
+
+.. code-block:: sdl
+
+    type Person {
+        required property name -> str {
+            constraint exclusive;
+        }
+        multi link cars -> Shoes {
```
elprans, line 101: cars?

docs/datamodel/links.rst
```diff
@@ -14,16 +14,265 @@ There are two kinds of link item declarations: *abstract links*, and
...
+    }
+    type Shoes {
+        required property description -> str {
+            # Just making sure that each description
+            # is unique like a name.
+            constraint exclusive;
+        }
+    }
+
+A ``Person`` can have potentially have more than one pair of
```
elprans, line 113: I think `Shoes` is an inconvenient example in the context of the English language, because you use the term "pair of ``Shoes``" when talking about a single object, so this might be somewhat confusing to non-native speakers.

docs/datamodel/links.rst
vpetrovykh: OK, I reworded it to talk about sources having links of certain kinds.

docs/datamodel/links.rst
vpetrovykh: Updated

docs/datamodel/links.rst
vpetrovykh: Many-to-one is mentioned in a backward link traversal description of one-to-many.

docs/datamodel/links.rst
vpetrovykh: An artifact from various iterations. Good catch :)

docs/datamodel/links.rst
vpetrovykh: Yeah. I'm looking for something that would be exclusively owned in a one-to-many way. So that's why I went for a personal item.

Maybe a ``Shirt`` or a ``Phone`` then? With ``Phone`` the example of getting multiple new ones may be a bit weird.

docs/datamodel/links.rst
elprans: That isn't enough.  `Many-to-One` is actually the most used option, so it deserves a top level mention.  Perhaps even before `One-to-One`.

vpetrovykh:
I've updated the order of presentation to be: many-to-one, one-to-one, many-to-many.
I've also updated the schema to use `Shirt` for the many-to-one example.

docs/datamodel/links.rst
```diff
@@ -14,16 +14,236 @@ There are two kinds of link item declarations: *abstract links*, and
...
+        }
+    }
+    type Shoes {
+        required property description -> str {
+            constraint exclusive;
+        }
+    }
+
+It's possible to include both links ``owner`` and ``shirts`` to a
+schema, making one of them a computable expression in terms of the
```
elprans, line 109: Perhaps add a link to some section that discusses computables, or, better, add an example?  In fact, putting backward traversal into a computable link is a common pattern.

elprans (approved):
LGTM, but please expand a bit on computables.

docs/datamodel/links.rst
vpetrovykh: Fair point. I might as well use this opportunity to show a version of the schema with the computable link. I think we also have a section where we talk about computables, so I can link to that as well.

vpetrovykh:
I updated the wording for computables.

docs/datamodel/links.rst
```diff
@@ -14,16 +14,259 @@ There are two kinds of link item declarations: *abstract links*, and
...
+
+    type Person {
+        required property name -> str {
+            constraint exclusive;
+        }
+        multi link shirts -> Shirt {
+            constraint exclusive;
+        }
+    }
+    type Shoes {
```
tailhook, line 102: s/Shoes/Shirt/

docs/datamodel/links.rst
```diff
@@ -14,16 +14,259 @@ There are two kinds of link item declarations: *abstract links*, and
...
+
+.. code-block:: edgeql
+
+    WITH M := (
+        SELECT Movie FILTER .title = "Matrix"
+    )
+    SELECT M.<likes[IS Person] {
+        name
+    };
+
```
tailhook, line 266: Somewhere here is a good place to introduce link properties?

docs/datamodel/links.rst
```diff
@@ -14,16 +14,259 @@ There are two kinds of link item declarations: *abstract links*, and
...
+following schema:
+
+.. code-block:: sdl
+
+    type Person {
+        required property name -> str {
+            constraint exclusive;
+        }
+        multi link shirts -> Shirt {
+            constraint exclusive;
```
tailhook, line 99: I would put some stress on `exclusive` here. This is what makes `many-to-one` instead of `many-to-many` in this case. Otherwise it's easy to miss when making your own schema.

docs/datamodel/links.rst
vpetrovykh: Replaced

docs/datamodel/links.rst
vpetrovykh: I've added a comment to the schema.

docs/datamodel/links.rst
vpetrovykh: Link properties are a messy topic all of their own. We have a few places scattered across the docs touching on them, but perhaps they should be reorganized.

PR #2221, opened 2021-02-12
Disallow implicit and assignment casting between mismatched named tuples

msullivan:
Also, include field names in the displayname of a named tuple, or else
the error message will just be total garbage.

### DIFF
--- a/tests/test_edgeql_datatypes.py
+++ b/tests/test_edgeql_datatypes.py
@@ -425,7 +425,7 @@ class TestEdgeQLDT(tb.QueryTestCase):
             [decimal.Decimal('1e100')],
         )
 
-    async def test_edgeql_named_tuple_mismatch_01(self):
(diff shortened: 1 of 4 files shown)


tests/test_edgeql_datatypes.py
```diff
@@ -424,3 +424,15 @@ async def test_edgeql_dt_decimal_05(self):
             [10.0**100],
             [decimal.Decimal('1e100')],
         )
+
+    async def test_edgeql_named_tuple_mismatch_01(self):
```
elprans, line 428: I'd also add a test for implicit casting `e.g. (a, b) UNION (b, a)`

elprans (approved):

msullivan:
While writing that test, I accidentally duplicated field names in a named tuple and discovered we don't catch that (and can ISE as a result), so I went ahead and fixed that also

PR #2226, opened 2021-02-13
Record DDL statements as migrations

elprans:
All DDL is now implicitly wrapped into a `CREATE MIGRATION` block.  This
allows safe interoperation between SDL-driven `edgedb migration`
workflows and direct modifications of the schema with DDL.


1st1 (approved):
lgtm

PR #2228, opened 2021-02-15
Extensions

elprans:
Add initial (provisional) support for `CREATE EXTENSION PACKAGE` and `CREATE EXTENSION` and add formal declarations for GraphQL et al.


edb/lib/ext/graphql.edgeql
```diff
@@ -17,6 +17,8 @@
 #
 
 
+CREATE EXTENSION PACKAGE graphql VERSION '1.0';
+
 CREATE MODULE stdgraphql;
```
1st1, line 22: Should this line be nested in `CREATE EXTENSION PACKAGE`?

edb/schema/schema.py
```diff
@@ -52,16 +52,25 @@
...
     sn.UnqualName('schema'),
     sn.UnqualName('math'),
     sn.UnqualName('sys'),
     sn.UnqualName('cfg'),
     sn.UnqualName('cal'),
+    sn.UnqualName('stdgraphql'),
 )
-STD_MODULES = frozenset(STD_LIB + (sn.UnqualName('stdgraphql'),))
 
+STD_SOURCES = (
```
1st1, line 65: Can you please add a comment what this is used for?

edb/server/dbview/dbview.pyx
```diff
@@ -555,7 +555,7 @@ cdef class DatabaseIndex:
         self._comp_sys_config = config.get_compilation_config(sys_config)
 
     def get_db(self, dbname):
-        return self._dbs[dbname]
+        return self._dbs.get(dbname)
```
1st1, line 558: I used `[]` on purpose here, to let it fail (it never should fail, unless there's a bug).

1st1 (approved):

edb/server/dbview/dbview.pyx
elprans: `POST /db/nonexistent/edgeql` crashes with a `KeyError` currently, and I don't think that catching `KeyError` is nicer than checking for `None`.  We can add another method: `maybe_get_db` or something.

edb/lib/ext/graphql.edgeql
elprans: I haven't actually implemented nested DDL for extensions, as we need to add a new tracking mechanism to attribute created objects to extensions for it to make sense, and that, in turn, requires a bit of refactoring.  Can be done post beta.

edb/server/dbview/dbview.pyx
1st1: Well, you might as well then fix connecting to a non-existent DB via binary.

1st1: Returning None here might make it a segfault now. Please check all get_db() callers.

1st1: Also, I wouldn't return None. I'd raise a proper `edb.errors` exception here, or at least a LookupError.

elprans: Done.

Issue #192, opened 2018-06-07
Discontinue `map` container type


vpetrovykh:
`map` container will be removed from EdgeDB.

We can have a `map` if and when we will implement it natively as a PostgreSQL type extension.

PR #2234, opened 2021-02-16
Add support for enabling extensions via SDL

elprans:
The new `using extension <extname> [version 'ver'];` syntax is the 
equivalent for `CREATE EXTENSION <extname>;` DDL.  We are stepping away 
from the usual custom of simply stripping the `CREATE` verb, because the 
resulting `extension <foo>;` is a bit _too_ terse and might at a glance be
confused with a declaration that puts the entire module into an extension.


edb/edgeql-parser/src/keywords.rs
```diff
@@ -45,7 +43,6 @@ pub const UNRESERVED_KEYWORDS: &[&str] = &[
     "named",
     "object",
     "of",
-    "oids",
```
1st1, line 48: Thanks God.

tests/test_edgeql_syntax.py
```diff
@@ -3057,6 +3057,31 @@ def test_edgeql_syntax_ddl_create_migration_06(self):
...
+
+    def test_edgeql_syntax_ddl_create_migration_08(self):
+        """
+        START MIGRATION TO {
+            using extension graphql;
+        };
+        """
+
+    @tb.must_fail(errors.EdgeQLSyntaxError,
+                  "'using extension' cannot be used inside a module block")
```
1st1, line 3075: I usually add col/line checks for this kind of tests.

1st1 (approved):

tests/test_edgeql_data_migration.py
```diff
@@ -7714,6 +7714,88 @@ async def test_edgeql_migration_non_ddl_statements(self):
...
+                }
+                FILTER .name = 'graphql'
+            """,
+            [{
+                'name': 'graphql',
+            }]
+        )
+
+        await self.con.execute('''
+            START MIGRATION TO {
```
dmgolembiowski, line 7751: Is this migration committing changes to the default module?

elprans: No.  Extensions are database-wide, and so are declared at the top level.

dmgolembiowski:
>> `map` container will be removed from EdgeDB.
>> 
>> We can have a `map` if and when we will implement it natively as a PostgreSQL type extension.
> Comment originally by @vpetrovykh https://github.com/edgedb/edgedb/issues/192#issue-330081977

Alright, please hold my beer.

PR #2241, opened 2021-02-17
Rip out binary.pyx/Timer; fix basic metrics

1st1:


elprans (approved):
Right.  We'd need to replace this with some more formal tracing framework, ideally some non-blocking event sink that acts as a sidecar to the server.

1st1:
> Right. We'd need to replace this with some more formal tracing framework, ideally some non-blocking event sink that acts as a sidecar to the server.

Yeah. Need to spec it and implement it (and actually use it!)

PR #2242, opened 2021-02-17
Fix interactions between inheritance and link deletion policies

### LINKED ISSUE #2231: `on target delete delete source` does not delete object extending abstract type
- EdgeDB Version: `1.0-alpha.7+g85a3931a7.d20210113` from [Docker Hub](https://hub.docker.com/layers/edgedb/edgedb/latest/images/sha256-ec9ec924f0d91769db32c6530f70406d637734f0dc579e4ca6ca73dd8c7e3f8e?context=explore)
- OS Version: Official Docker image running on `NixOS 21.03pre-git (x86_64)`

Steps to Reproduce:

1. `INSERT B { value := (INSERT C) };`
2. `DELETE C;`

Now `B` should be deleted through the `A.value` link target deletion policy, but it was instead set empty. Check with `SELECT B { value };`.

Schema:

```edgeql
abstract type A {
	link value -> C {
	};
}

type B extending A {}

type C {}
```

### LINKED ISSUE #2232: `constraint expression` from parent type is not checked on link target deletion
- EdgeDB Version: `1.0-alpha.7+g85a3931a7.d20210113` from [Docker Hub](https://hub.docker.com/layers/edgedb/edgedb/latest/images/sha256-ec9ec924f0d91769db32c6530f70406d637734f0dc579e4ca6ca73dd8c7e3f8e?context=explore)
- OS Version: Official Docker image running on `NixOS 21.03pre-git (x86_64)`

Steps to Reproduce:

1. `INSERT B { value := (INSERT C) };`
2. `DELETE C;`

Step 2 should throw an error because the constraint is no longer satisfied.

Schema:

```edgeql
type A {
	link value -> C {
		on target delete allow;
	};

	constraint expression on (
		EXISTS .value
	);
}

type B extending A {}

type C {}
```

msullivan:
Rework the creation of link deletion policies to ensure that the
proper triggers always get created and deleted in the presence of
inheritance, rebases, and changes to cardinality and link policies.

Actually creating all the proper triggers caused some grief updating
the reflection schema, so most links pointing to schema::Type now have
`ON TARGET DELETE DEFERRED RESTRICT`. (In addition to the magically
implicit `ON TARGET DELETE ALLOW` inserted for all refdicts.)

Fixes #2231. Fixes #2232.

### DIFF
--- a/edb/lib/schema.edgeql
+++ b/edb/lib/schema.edgeql
@@ -146,7 +146,7 @@ CREATE ABSTRACT TYPE schema::AnnotationSubject EXTENDING schema::Object {
     CREATE MULTI LINK annotations EXTENDING schema::reference
     -> schema::Annotation {
         CREATE PROPERTY value -> std::str;
-	ON TARGET DELETE DEFERRED RESTRICT;
+	ON TARGET DELETE ALLOW;
     };
 };
 
@@ -176,7 +176,7 @@ CREATE ABSTRACT TYPE schema::CallableObject
     EXTENDING schema::AnnotationSubject
 {
     CREATE MULTI LINK params -> schema::Parameter {
-        ON TARGET DELETE DEFERRED RESTRICT;
+        ON TARGET DELETE ALLOW;
     };
 
     CREATE LINK return_type -> schema::Type {
@@ -213,7 +213,7 @@ CREATE ABSTRACT TYPE schema::ConsistencySubject EXTENDING schema::Object {
     CREATE MULTI LINK constraints EXTENDING schema::reference
     -> schema::Constraint {
         CREATE CONSTRAINT std::exclusive;
-	ON TARGET DELETE DEFERRED RESTRICT;
+	ON TARGET DELETE ALLOW;
     };
 };
 
@@ -231,7 +231,7 @@ CREATE TYPE schema::Index EXTENDING schema::AnnotationSubject {
 CREATE ABSTRACT TYPE schema::Source EXTENDING schema::Object {
     CREATE MULTI LINK indexes -> schema::Index {
         CREATE CONSTRAINT std::exclusive;
-	ON TARGET DELETE DEFERRED RESTRICT;
+	ON TARGET DELETE ALLOW;
     };
 };
 
@@ -252,7 +252,7 @@ CREATE ABSTRACT TYPE schema::Pointer
 ALTER TYPE schema::Source {
     CREATE MULTI LINK pointers EXTENDING schema::reference -> schema::Pointer {
         CREATE CONSTRAINT std::exclusive;
-	ON TARGET DELETE DEFERRED RESTRICT;
+	ON TARGET DELETE ALLOW;
     };
 };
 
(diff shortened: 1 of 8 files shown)


edb/lib/schema.edgeql
```diff
@@ -287,7 +297,9 @@ CREATE TYPE schema::Property EXTENDING schema::Pointer;
 
 ALTER TYPE schema::Pointer {
     CREATE LINK source -> schema::Source;
-    CREATE LINK target -> schema::Type;
+    CREATE LINK target -> schema::Type {
+        ON TARGET DELETE DEFERRED RESTRICT;
```
elprans, line 301: Good call.  I wonder if we can actually turn some `ALLOW` in `structure.py` into this?

tests/test_edgeql_ddl.py
```diff
@@ -10270,6 +10270,244 @@ async def test_edgeql_ddl_captured_as_migration_01(self):
             }]
         )
 
+    async def test_edgeql_ddl_link_policy_01(self):
```
elprans, line 10273: Let's add some tests with constraints like case in #2232 to test that we're clearing up inline links.  I also have a feeling that `required multi` links are probably also broken with respect to this.

msullivan: I added a test for #2232, and then some broken ones for MULTI.

What is the story with the multi issue? We can't have our constraints operating over multiple tables, so we do the checking on the query side, which means our deletion trigger won't set off the error? Should we just disallow that feature combination?

tests/test_edgeql_ddl.py
```diff
@@ -10508,6 +10508,115 @@ async def test_edgeql_ddl_link_policy_09(self):
...
+        """)
+
+    @test.xfail('''
+        We can't properly compile the constraint here in the basic case,
+        let alone combined with DELETE ALLOW.
+
+        We probably should be rejecting the constraint.
+    ''')
+    async def test_edgeql_ddl_link_policy_11(self):
+        # Should we even be allowing (EXISTS .tgt) as a constraint?
```
elprans, line 10552: I see no reason why it couldn't be equivalent to `required`, but I'm also OK with punting this and prohibiting `EXISTS` along with other references to `multi` on object constraints.

tests/test_edgeql_ddl.py
msullivan: Or maybe we can make the ON DELETE ALLOW specifically understand how to handle required

tests/test_edgeql_ddl.py
```diff
@@ -10508,6 +10508,115 @@ async def test_edgeql_ddl_link_policy_09(self):
...
+            """)
+
+        await self.con.execute("""
+            DELETE Bar;
+            DELETE Tgt;
+        """)
+
+    @test.xfail('''
+        We don't handle this correctly because we do REQUIRED MULTI
+        enforcement on the query end.
```
elprans, line 10586: We do handle this in `_alter_pointer_cardinality` and `_alter_pointer_optionality`, though, so there's precedent.  No reason why the same technique couldn't work from a trigger.

elprans (approved):

Issue #2230, opened 2021-02-15
Update documentation for the migration workflows


elprans:
The Getting Started flow, tutorials, the cheatsheet, as well as the reference need to be updated to include the CLI-based migration flows.

1st1:
Yeah, the deadline for this is "yesterday". :)

Couple of notes re Getting Started:

* No DDL in it.
* Showcase CLI to download & install EdgeDB.
* Use CLI to apply migrations (not repl). Explain where to put our ESDL files, etc.
* Enabling GraphQL should be done via adding `using extension graphql` to the ESDL file (coming soon, @elprans is working on that)

dmgolembiowski:
Aaah this is so great! Thank you!

elprans:
@vpetrovykh, is this done?

PR #2244, opened 2021-02-17
Update the "getting started" docs to use migrations.

vpetrovykh:
Issue #2230.

### DIFF
--- a/docs/tutorial/queries.rst
+++ b/docs/tutorial/queries.rst
@@ -4,41 +4,48 @@
 =========
 
 After the database and schema are set up, we can add some actual data.
-For example, let's add "Blade Runner 2049" to the database. It's possible to
-add movie, director and actor data all at once:
+For this tutorial we will use the command-line REPL tool to do that,
+so let's start it up:
+
+.. code-block:: bash
+
+    $ edgedb -I tutorial
+
+Now, let's add "Blade Runner 2049" to the database. It's
+possible to add movie, director and actor data all at once:
 
 .. code-block:: edgeql-repl
 
-    tutorial> INSERT Movie {
-    .........     title := 'Blade Runner 2049',
-    .........     year := 2017,
-    .........     director := (
-    .........         INSERT Person {
-    .........             first_name := 'Denis',
-    .........             last_name := 'Villeneuve',
-    .........         }
-    .........     ),
-    .........     actors := {
-    .........         (INSERT Person {
-    .........             first_name := 'Harrison',
-    .........             last_name := 'Ford',
-    .........         }),
-    .........         (INSERT Person {
-    .........             first_name := 'Ryan',
-    .........             last_name := 'Gosling',
-    .........         }),
-    .........         (INSERT Person {
-    .........             first_name := 'Ana',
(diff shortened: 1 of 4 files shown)


docs/tutorial/queries.rst
```diff
@@ -152,127 +164,163 @@ there in the ``actors`` by using a :ref:`computable
...
+    'default::Person' optional? [y,n,l,c,b,s,q,?]
+    y
+    Created movie_schema/migrations/00002.edgeql, id:
+    m15dogciav42crguwkslmjoucq5ieg25hcgk7ewxfmcmzf7othksda
+
+Third and final step in this sequence is to apply the migration:
+
+.. code-block:: bash
+
+    $ edgedb migrate --schema-dir movie_schema
```
1st1, line 229: `--schema-dir` -- I don't think we need this in the tutorial. We should use the default location.

vpetrovykh: You're right that we can just use the default for this tutorial.

I wonder if it's still worth mentioning that the directory can be specified via this option, just to reassure a newcomer that they still can pick some sensible location for their schema in the project.

1st1: > I wonder if it's still worth mentioning that the directory can be specified via this option, just to reassure a newcomer that they still can pick some sensible location for their schema in the project.

I don't think we should. We can't cover all side questions that readers might have -- the tutorial would become hard to read. So let's keep it as simple as possible and to the point.

Dhghomon:
Just looked over the whole thing and only found one small typo - just delete **the** here:

questions to make sure that **the** EdgeDB correctly inferred the changes:
questions to make sure that EdgeDB correctly inferred the changes:

vpetrovykh:
I've cleaned up the getting started section and now I've also updated the GraphQL sections so that they use the `extension` syntax.

docs/tutorial/install.rst
```diff
@@ -31,7 +32,7 @@ Then, let's install and configure the EdgeDB server:
 .. code-block:: bash
 
     $ edgedb server install
-    $ edgedb server init
+    $ edgedb server init default
```
elprans, line 35: Perhaps use something like `myapp` here to highlight that the choice of the instance name is arbitrary and up to the user.  `default` might be interpreted like something you _have_ to use, or even as some weird argument to `init` ("init with defaults or something").

docs/tutorial/queries.rst
```diff
@@ -4,77 +4,82 @@
...
 
 After the database and schema are set up, we can add some actual data.
-For example, let's add "Blade Runner 2049" to the database. It's possible to
-add movie, director and actor data all at once:
+For this tutorial we will use the command-line REPL tool to do that,
+so let's start it up:
+
+.. code-block:: bash
+
+    $ edgedb
```
elprans, line 12: This should be `edgedb -I myapp` (or whatever instance name we choose for the tutorial).

vpetrovykh:
I rebased the getting started section on top of the #2263. The `-I tutorial` is now used consistently.

elprans (approved):

PR #2245, opened 2021-02-17
Set IntervalStyle to 'sql_standard'

### LINKED ISSUE #2229: configure Postgres connections to have proper `IntervalStyle`
Current behavior:

```
edgedb> SELECT to_str(<duration>'-24674.045854s', 'HH24hMImSS.USs');
{'-06h-51m-14.-45854s'}
edgedb> SELECT <duration>'-6h51m14.045854s';
{-18525.954146s}
edgedb> SELECT <duration>'-6h -51m -14.045854s';
{-24674.045854s}
edgedb> SELECT <duration>'-6h-51m-14.045854s';
ERROR: InvalidValueError: invalid input syntax for type std::duration: "-6h-51m-14.045854s"
```

Postgres:
> According to the SQL standard all fields of an interval value must have the same sign, so a leading negative sign applies to all fields; for example the negative sign in the interval literal '-1 2:03:04' applies to both the days and hour/minute/second parts. PostgreSQL allows the fields to have different signs, and traditionally treats each field in the textual representation as independently signed, so that the hour/minute/second part is considered positive in this example. If IntervalStyle is set to sql_standard then a leading sign is considered to apply to all fields (but only if no additional signs appear). Otherwise the traditional PostgreSQL interpretation is used. To avoid ambiguity, it's recommended to attach an explicit sign to each field if any field is negative.

**Action:** We want to have IntervalStyle set to `sql_standard`.

ambv:
Fixes #2229

### DIFF
--- a/edb/server/pgcon/pgcon.pyx
+++ b/edb/server/pgcon/pgcon.pyx
@@ -1201,6 +1201,9 @@ cdef class PGConnection:
         buf.write_bytestring(b'default_transaction_isolation')
         buf.write_bytestring(b'repeatable read')
 
+        buf.write_bytestring(b'intervalstyle')
+        buf.write_bytestring(b'sql_standard')
+
         buf.write_bytestring(b'user')
         buf.write_bytestring(self.pgaddr['user'].encode('utf-8'))
 
--- a/edb/server/pgcluster.py
+++ b/edb/server/pgcluster.py
@@ -152,6 +152,9 @@ class BaseCluster:
                 f'SET ROLE {self._default_session_auth};'
             )
 
+        await conn.execute(
+            "SET IntervalStyle = 'sql_standard';"
+        )
         return conn
 
     def get_runtime_params(self) -> BackendRuntimeParams:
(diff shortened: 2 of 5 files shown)


edb/server/pgcon/pgcon.pyx
```diff
@@ -135,6 +135,11 @@ async def connect(connargs, dbname):
             ignore_data=True,
         )
 
+    await pgcon.simple_query(
```
elprans, line 138: This should go into `connect()` in this file to avoid the extra query roundtrip.

ambv:
`AssertionError: data shape differs: '0:00:20' != '00:00:20' PATH: [0]["str_duration"]{0}`

1st1:
> AssertionError: data shape differs: '0:00:20' != '00:00:20' PATH: [0]["str_duration"]{0} while testing restores.

Fundamentally the data is still there and appears to be correct, so we can just fix that assertion in test_dump01.py. @elprans do you agree?

edb/server/pgcon/pgcon.pyx
1st1: Yeah, move it to pgcon.pyx:1192

elprans:
Yes, just fix the test.

elprans (approved):

1st1:
I've fixed a few other failing tests. Waiting for the CI to clear.

PR #2247, opened 2021-02-17
Improve CI

fantix:
Background of Github Action caching:

* Cache is immutable. That means, if `cache-hit` is `true` (`key` exact match), there will be no cache uploaded in the end.
* The `restore-keys` are used for partial match. Partially matching cache will be restored at the beginning, but `cache-hit` will be `false`. A new cache with `key` will be created in the end if the workflow succeeded.
* Cache created in a Git branch can only be accessed by itself and its descendants. Different pull requests (siblings) won't share cache.

Changes in this PR:

* Split down the build job, use separate cache by adding new `build_xxx` commands in setup.py
* Use more precise hash for each build step, test shards no longer need to download the build cache
* Optimize tests shard distribution for database setup
* Record database setup running time

Steps to build EdgeDB server (ordered):

| Code Name | Brief | Source | Target | Cache | Build Cache | Build command ____________________________________ |
|---|---|---|---|---|---|---|
| `venv` | Python dependencies | `setup.py` | `$VIRTUAL_ENV` | Same as target | Same as target | `pip download ...` `pip install ...` |
| `cli` | edgedb-cli in Rust | [edgedb/edgedb-cli](https://github.com/edgedb/edgedb-cli) @ master | `edb/cli/edgedb` | `build/cli` | `build/temp.arch/rust/cli` | `setup.py build_cli` |
| `rust` | Extensions in Rust | `edb/**/*.rs` `edb/**/Cargo.toml` | `edb/*.so` | `build/rust_extensions` `build/lib.arch/edb/*.so` | `build/temp.arch/rust/extensions` | `setup.py build_rust` |
| `ext` | Extensions in Cython | `edb/**/*.py[x/i]` `edb/**/*.px[d/i]` | `edb/**/*.so` | `build/extensions` `build/lib.arch/edb/**/*.so` | `build/temp.arch/edb` | `BUILD_EXT_MODE=py-only setup.py build_ext` |
| `parsers` | Parser table | `edb/edgeql/parser/grammer/**/*.py` `edgeql-parser/src/keywords.rs` | `edb/edgeql/parser/grammar/*.pickle` | `build/lib` | None | `setup.py build_parsers --inplace` |
| `postgres` | PostgreSQL | `postgres` | `build/postgres/install` | Same as target | None, fresh build | `setup.py build_postgres` |
| - | Editable installation | `setup.py` | `edgedb_server.egg-info` | `$VIRTUALENV/edgedb_server.egg-info` | None | `BUILD_EXT_MODE=skip pip install -e .` |
| `bootstrap` | EdgeDB server bootstrap (stdlib cache) | `CACHE_SRC_DIRS`: `edb/schema/**/*.py` `edb/lib/**/*.edgeql` `edb/edgeql/parser/**/*.py` `edb/edgeql/compiler/**/*.py` Additional: `edb/pgsql/metaschema/**/*.py` `edb/server/bootstrap.py` | `build/cache` | Same as target | None | `edb server --bootstrap-only` |

TODOs:

- [ ] Update the repo branch protection rule to use `test-conclusion` as the required job

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -32,40 +32,38 @@ jobs:
       with:
         python-version: 3.9
 
-    - name: Handle Pip dependency cache
-      uses: actions/cache@v2
-      id: depcache
+    # Build virtualenv
+
+    - name: Handle virtualenv
+      uses: syphar/restore-virtualenv@v1.1
+      id: venv-cache
       with:
-        path: deps
-        key: edbdeps-pip-${{ hashFiles('setup.py') }}
+        requirement_files: setup.py
 
-    - name: Download dependencies
-      if: steps.depcache.outputs.cache-hit != 'true'
+    - name: Install Python dependencies
+      if: steps.venv-cache.outputs.cache-hit != 'true'
       run: |
-        pip download --dest=deps .[test,docs]
+        pip download --dest=$VIRTUAL_ENV/deps .[test,docs]
+        pip install -U --no-index --find-links=$VIRTUAL_ENV/deps $VIRTUAL_ENV/deps/*
 
-    - name: Install Python deps
-      run: |
-        pip install -U --no-index --find-links=deps deps/*
+    # Prepare environment variables and shared artifacts
 
-    - name: Compute build cache key
-      run: |
-        mkdir -p .tmp
-        python setup.py --quiet gen_build_cache_key >.tmp/build_cache_key.txt
-
-    - name: Handle build cache
-      uses: actions/cache@v2
-      id: buildcache
-      with:
-        path: build
-        key: edbbuild-v2-${{ hashFiles('.tmp/build_cache_key.txt') }}
-        restore-keys: |
-          edbbuild-v2-
-
-    - name: Download the running times log file
+    - name: Compute cache keys and download the running times log
       env:
         GIST_TOKEN: ${{ secrets.CI_BOT_GIST_TOKEN }}
       run: |
+        mkdir -p .tmp
+        python setup.py -q ci_helper --type cli > .tmp/edgedbcli_git_rev.txt
+        python setup.py -q ci_helper --type rust >.tmp/rust_cache_key.txt
+        python setup.py -q ci_helper --type ext >.tmp/ext_cache_key.txt
+        python setup.py -q ci_helper --type parsers >.tmp/parsers_cache_key.txt
+        python setup.py -q ci_helper --type postgres >.tmp/postgres_git_rev.txt
+        python setup.py -q ci_helper --type bootstrap >.tmp/bootstrap_cache_key.txt
+        echo EDGEDBCLI_GIT_REV=$(cat .tmp/edgedbcli_git_rev.txt) >> $GITHUB_ENV
+        echo POSTGRES_GIT_REV=$(cat .tmp/postgres_git_rev.txt) >> $GITHUB_ENV
+        echo BUILD_LIB=$(python setup.py -q ci_helper --type build_lib) >> $GITHUB_ENV
+        echo BUILD_TEMP=$(python setup.py -q ci_helper --type build_temp) >> $GITHUB_ENV
+
         curl \
           -H "Accept: application/vnd.github.v3+json" \
           -u edgedb-ci:$GIST_TOKEN \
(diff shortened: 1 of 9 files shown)


edb/testbase/server.py
```diff
@@ -1524,63 +1529,140 @@ def start_edgedb_server(
...
+        else:
+            for test in tests:
+                total_tests += 1
+                est = stats.get(str(test), (new_test_est, 0))[0]
+                total_est += est
+                heapq.heappush(tests_with_est, (-est, total_tests, test))
+
+    target_est = total_est / total_shards  # target running time of one shard
+    shards_est = [(0, shard, set()) for shard in range(total_shards)]
+    cases = collections.OrderedDict()  # output
```
1st1, line 1582: nit: unless you're using some specific OrderedDict apis, you can use the plain python dict here (which are ordered too).

fantix: Oh, interesting! I was only to keep a consistent returning value as:

```python
def get_test_cases(tests):
    result = collections.OrderedDict()
```

1st1:
Is it possible to change the final action name from "test" to something more descriptive? I assume that's the action that verifies that all tests were run?

fantix:
> that's the action that verifies that all tests were run?

Yes, exactly - we should update the Github repo branch protection rule correspondingly too. I'll rename it to `verify`, any other suggestions please?

elprans:
"Test Conclusion" would be more descriptive.

.github/workflows/tests.yml
```diff
@@ -73,39 +73,45 @@ jobs:
...
-      if: steps.buildcache.outputs.cache-hit != 'true'
+      if: steps.buildcache.outputs.cache-hit != 'true' || steps.venvcache.outputs.cache-hit != 'true'
       run: |
         # --no-use-pep517 because we have explicitly installed all deps
         # and don't want them to be reinstalled in an "isolated env".
         pip install --no-use-pep517 --no-deps -e .[test,docs]
 
     - name: Bootstrap EdgeDB Server
-      if: steps.buildcache.outputs.cache-hit != 'true'
+      if: steps.buildcache.outputs.cache-hit != 'true' || steps.venvcache.outputs.cache-hit != 'true'
```
elprans, line 98: If I'm reading this correctly, the build step and the bootstrap step share the same cache?  Will bootstrap invalidation affect the build cache too?  Many PRs change the core source and so bootstrap cache would be invalidated much more frequently than the build cache which depends only on deps and Rust bits.

fantix: > the build step and the bootstrap step share the same cache?

Yes.

> Will bootstrap invalidation affect the build cache too?

Yes, because they are placed in the same `build/` folder, though the impact is partial - Github Action would still be able to pull in the previous cache as a partial match (`cache-hit = false`) of `restore-keys`. Looks like `pip install -e .` will still re-build the extensions, but the time is shorter than a complete rebuild.

Are the pickle files the only outcome of the bootstrap, please?  I wasn't sure if the `build/` directory was also updated. If we could separate the bootstrap cache, that would be the best. Actually, if we could confirm the following, I could make the caching more fine-grained and efficient, we don't even need the `build/` folder in each shard (?):

* `setup.py` -> virtualenv
* `postgres` + `pgproto` + Rust/Cython Files -> `*.so`
* `CACHE_SRC_DIRS` -> `*.pickle`
* `edb/edgeql/parser/grammer/**/*.py` -> ???
* edgedb-cli -> `edb/cli/edgedb`

I'm sorry if this is a dumb question due to my ignorance, but I'd like to ask here first before I spend too much time reading the wrong part of the code.

fantix: Never mind - I think I figured this out, will do some more testing here.

1st1: Can we have a separate cache for every item here:

> virtualenv + postgres + pgproto + Rust/Cython Files -> *.so

?

E.g. "postgres" is very unlikely to change, so it should always be cached separately (building it takes a while).

elprans: > Are the pickle files the only outcome of the bootstrap to be cached

The outcome of bootstrap is the contents of `build/cache/`, which includes a schema pickle and an SQL dump, so `CACHE_SRC_DIRS` -> `build/cache`

> `edb/edgeql/parser/grammer/**/*.py` -> 

these produce the parser table pickles (which are stored next to them). See `_compile_parsers` in `setup.py`.  All parsers share the grammar and so should be cached together (just take the hash of `edb.edgeql.parser.grammar` and `edgeql-parser/src/keywords.rs`.

edb/server/buildmeta.py
```diff
@@ -122,7 +123,10 @@ def hash_dir(dirname, ext, paths):
         hash_dir(dirname, ext, paths)
 
     if extra_files:
-        paths.extend(extra_files)
+        for extra_file in extra_files:
+            if hasattr(extra_file, 'resolve'):
```
elprans, line 127: This isn't going to save you from mypy, I'm afraid.  Use `isinstance(extra_file, pathlib.Path)`

fantix: Thanks!

1st1:
@fantix looks like the ci is red.

fantix:
It was green 🥀 looks like Github is giving 500 for concurrent artifact updates 🤔

1st1:
> It was green 🥀 looks like Github is giving 500 for concurrent artifact updates 🤔

Putting a for-loop around every cloud operation is considered a "good" practice. So that's what we likely need to do, somehow.

elprans:
> looks like Github is giving 500 for concurrent artifact update

You might want to split the cache key file and the time stats file into separate artifacts so that artifact uploads don't contain overlapping files.

fantix:
OK, I think all issues are addressed now. 32 shards = `5m 23s`, 16 shards = `6m 18s`, but looks like 32 shards easily get queued up when multiple PRs are running CI.

elprans (approved):
Lets stay with 16 for now.

PR #2249, opened 2021-02-18
Fix parsing datetimes with > 4 year digits

### LINKED ISSUE #2248: datetime casting is asymmetric
It seems surprising to me that some datetimes can be cast to str but not cast back to datetime. This happens if the year has more than 4 digits.

```
edgedb> SELECT to_datetime(999999999999);
{<datetime>'+33658-09-27T01:46:39.000064Z'}
edgedb> SELECT <datetime><str>to_datetime(999999999999);
ERROR: InvalidValueError: invalid input syntax for type std::datetime: '33658-09-27T01:46:39.000064+00:00'
  Hint: Please use ISO8601 format. Alternatively "to_datetime" function provides custom formatting options.
<!-- Please search existing issues to avoid creating duplicates. -->
```

- EdgeDB Version: package:1.0a8.dev5375+g8205d736c-2021020400~focal.nightly
- OS Version: Linux 5.8.0-43-generic #49~20.04.1-Ubuntu SMP Fri Feb 5 09:57:56 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

tailhook:
Fixes #2248

### DIFF
--- a/edb/pgsql/metaschema.py
+++ b/edb/pgsql/metaschema.py
@@ -1583,7 +1583,7 @@ class DatetimeInFunction(dbops.Function):
         SELECT
             CASE WHEN val !~ (
                     '^\s*(' ||
-                        '(\d{4}-\d{2}-\d{2}|\d{8})' ||
+                        '((?:\+\d+)?\d{4}-\d{2}-\d{2}|\d{8})' ||
                         '[ tT]' ||
                         '(\d{2}(:\d{2}(:\d{2}(\.\d+)?)?)?|\d{2,6}(\.\d+)?)' ||
                         '([zZ]|[-+](\d{2,4}|\d{2}:\d{2}))' ||
@@ -1604,7 +1604,7 @@ class DatetimeInFunction(dbops.Function):
                     )
                 )
             ELSE
-                val::timestamptz
+                regexp_replace(val, '^\s*\+', '')::timestamptz
             END;
     '''
 
@@ -1665,7 +1665,7 @@ class LocalDatetimeInFunction(dbops.Function):
             CASE WHEN
                 val !~ (
                     '^\s*(' ||
-                        '(\d{4}-\d{2}-\d{2}|\d{8})' ||
+                        '((?:\+\d+)?\d{4}-\d{2}-\d{2}|\d{8})' ||
                         '[ tT]' ||
                         '(\d{2}(:\d{2}(:\d{2}(\.\d+)?)?)?|\d{2,6}(\.\d+)?)' ||
                     ')\s*$'
@@ -1685,7 +1685,7 @@ class LocalDatetimeInFunction(dbops.Function):
                     )
                 )
             ELSE
-                val::timestamp
+                regexp_replace(val, '^\s*\+', '')::timestamp
             END;
     '''
 
--- a/tests/test_edgeql_casts.py
+++ b/tests/test_edgeql_casts.py
@@ -635,6 +635,18 @@ class TestEdgeQLCasts(tb.QueryTestCase):
             [True, True, True, True, True, True, True, True],
         )
 
+        await self.assert_query_result(
+            # validating that these are all in fact the same datetime
+            r'''
+                WITH x := {
+                    '+33658-09-27T01:46:39.000064Z',
+                }
+                SELECT <datetime>x =
+                    <datetime>'+33658-09-27T01:46:39.000064Z';
+            ''',
+            [True],
+        )
+
         async with self.assertRaisesRegexTx(
                 edgedb.InvalidValueError,
                 r'invalid input syntax'):
@@ -691,6 +703,18 @@ class TestEdgeQLCasts(tb.QueryTestCase):
                     <cal::local_datetime>'2018-05-07T20:01:22.306916';
             ''',
             [True, True, True],
+        )
+
+        await self.assert_query_result(
+            # validating that these are all in fact the same datetime
+            r'''
+                WITH x := {
+                    '+33658-09-27T01:46:39.000064',
+                }
+                SELECT <cal::local_datetime>x =
+                    <cal::local_datetime>'+33658-09-27T01:46:39.000064';
+            ''',
+            [True],
         )
 
         async with self.assertRaisesRegexTx(


1st1:
@vpetrovykh please review

vpetrovykh:
Is it illegal to do this?
`SELECT <datetime>''+3658-09-27T01:46:39.000064Z"`
Basically, my question is whether we should accept a leading `+` even for 4-digit year if we accept it for more digits. The rationale is that I'm not sure that `+33658-09-27T01:46:39.000064Z` is ISO8601, so if we allow deviation, we might as well allow that leading `+` in general.

Other than that the PR looks good.

edb/pgsql/metaschema.py
```diff
@@ -1665,7 +1665,7 @@ class LocalDatetimeInFunction(dbops.Function):
             CASE WHEN
                 val !~ (
                     '^\s*(' ||
-                        '(\d{4}-\d{2}-\d{2}|\d{8})' ||
+                        '((?:\+\d+)?\d{4}-\d{2}-\d{2}|\d{8})' ||
```
elprans, line 1668: We ought to support negative years too.

edb/pgsql/metaschema.py
```diff
@@ -1685,7 +1685,7 @@ class LocalDatetimeInFunction(dbops.Function):
                     )
                 )
             ELSE
-                val::timestamp
+                regexp_replace(val, '^\s*\+', '')::timestamp
```
elprans, line 1688: Use `to_timestamp(val, 'YYYY-MM-DDTHH:MI:SS.USZ')` instead.  That one is able to parse the sign and extended year digits.

edb/pgsql/metaschema.py
vpetrovykh: Yeah +1 on the negative year.
```
edgedb> SELECT to_datetime(-99999999999);
{<datetime>'-1199-02-15T14:13:21Z'}
edgedb> SELECT <datetime>"-1199-02-15T14:13:21Z";
ERROR: InvalidValueError: invalid input syntax for type std::datetime: '-1199-02-15T14:13:21Z'
  Hint: Please use ISO8601 format. Alternatively "to_datetime" function provides custom formatting options.
```

elprans:
> I'm not sure that `+33658-09-27T01:46:39.000064Z` is ISO8601

It is: https://en.wikipedia.org/wiki/ISO_8601#Years

edb/pgsql/metaschema.py
vpetrovykh: There's a problem with that one:
```
# SELECT to_timestamp('4368-04-02T04:26:41Z', 'YYYY-MM-DDTHH:MI:SS.USZ');
ERROR:  invalid value "Z" for "US"
DETAIL:  Value must be an integer.
```

Basically the ISO8601 standard makes the fractional seconds optional, while the specific parsing format makes them obligatory. That's why I didn't use it in the first place.

vpetrovykh:
> > I'm not sure that `+33658-09-27T01:46:39.000064Z` is ISO8601
> 
> It is: https://en.wikipedia.org/wiki/ISO_8601#Years

OK, as far as I can see they also refer to +0000 and -0001 as valid there. So a +/- prefix should be allowed.

edb/pgsql/metaschema.py
elprans: Since we use the regexp anyway, we can detect the presence of optional parts and adjust the format string accordingly.

1st1:
> OK, as far as I can see they also refer to +0000 and -0001 as valid there. So a +/- prefix should be allowed.

Feel free to push to Paul's branch with a follow up commit to fix things and add more tests. Let's merge this sooner rather than later.

edb/pgsql/metaschema.py
vpetrovykh: Sigh, we can probably look at the seconds and if they have a `.` use one format and of they don't have fractional value, use another. This would come at the cost of one more "if" (I guess `CASE WHEN`).

vpetrovykh:
> Feel free to push to Paul's branch with a follow up commit to fix things and add more tests. Let's merge this sooner rather than later.

OK, let me try to apply all the suggestions so far as another commit on top of this.

vpetrovykh:
And we need to fix the `<str>` cast from `datetime` as well, to avoid this:
```
edgedb> SELECT <str><datetime>'+111001-01-01T00:00:00Z';
{'111001-01-01T00:00:00+00:00'}
edgedb> SELECT <str><datetime>'-0002-01-01T00:00:00Z';
{'0002-01-01T00:00:00+00:00 BC'}
```
Notice the lack of "+" and the non-ISO8601 "BC".

vpetrovykh:
Investigating what happens to years outside 4-digit ISO8601 standard has uncovered the following:

### Postgres doesn't render or parse them correctly

The ISO8601 allows using years outside the 0000-9999 range, but mandates the use of + or - sign for those occasions. Our typical way of getting an ISO8601 timestamp was to use `to_json`, but we also considered a `text` cast and formatting with `to_char`:
```
edgedb=# SELECT make_timestamptz(12013, 7, 15, 8, 15, 23);  
    make_timestamptz     
-------------------------
 12013-07-15 08:15:23-04
(1 row)

edgedb=# SELECT to_json(make_timestamptz(12013, 7, 15, 8, 15, 23));
           to_json            
------------------------------
 "12013-07-15T08:15:23-04:00"
(1 row)

edgedb=# SELECT make_timestamptz(12013, 7, 15, 8, 15, 23)::text;
    make_timestamptz     
-------------------------
 12013-07-15 08:15:23-04
(1 row)

edgedb=# SELECT to_char(make_timestamptz(12013, 7, 15, 8, 15, 23), 'YYYY');
 to_char 
---------
 12013
(1 row)
```
Dates with negative years are harder to construct (`make_timestamptz` requires year to be strictly positive), but they can be the result of casting, parsing, or arithmetic. The end result is that the date is rendered with `BC` suffix instead of using ISO8601 approach of negative year:
```
edgedb=# SELECT to_json(make_timestamptz(0, 7, 15, 8, 15, 23));
ERROR:  date field value out of range: 0-07-15
edgedb=# SELECT to_json('1000-01-01T00:00:00 BC'::timestamptz);
              to_json              
-----------------------------------
 "1000-01-01T00:00:00-05:17:32 BC"
(1 row)

edgedb=# SELECT '1000-01-01T00:00:00 BC'::timestamptz::text;
              text               
---------------------------------
 1000-01-01 00:00:00-05:17:32 BC
(1 row)

edgedb=# SELECT to_char('1000-01-01T00:00:00 BC'::timestamptz, 'YYYY');
 to_char 
---------
 1000
(1 row)
```
Not only is the `-` not rendered, but the year 1000 BC corresponds to year -0999 ISO8601 (because 0001 = 1 CE, 0000 = 1 BCE, -0001 = 2 BCE, etc.). So it is numerically wrong if we wanted to construct our own ISO-compliant rendering format. We can try to `extract`:
```
edgedb=# SELECT extract(year from '1000-01-01T00:00:00 BC'::timestamptz);
 date_part 
-----------
     -1000
(1 row)
```
Well, we get the sign, but still would need to adjust the value.

Parsing is a problem for negative years:
```
edgedb=# SELECT to_timestamp('-1000-09-27T01:46:39Z', 'YYYY-MM-DDTHH:MI:SSZ');
          to_timestamp           
---------------------------------
 1000-09-27 00:46:39-05:17:32 BC
(1 row)
```
OK, so again we have the same thing as for rendering. The ISO year -1000 should map onto 1001 BCE and not 1000 BCE, but there's an additional issue stemming from that:
```
edgedb=# SELECT to_timestamp('0000-09-27T01:46:39Z', 'YYYY-MM-DDTHH:MI:SSZ');
          to_timestamp           
---------------------------------
 0001-09-27 00:46:39-05:17:32 BC
(1 row)

edgedb=# SELECT to_timestamp('-0001-09-27T01:46:39Z', 'YYYY-MM-DDTHH:MI:SSZ');
          to_timestamp           
---------------------------------
 0001-09-27 00:46:39-05:17:32 BC
(1 row)
```
All of this makes reliably parsing ISO8601-compliant dates with year outside the range of 0000-9999 hard for our own casting functions and near impossible for a user-defined format (i.e. no format string would correctly parse or render ISO8601-compliant BCE dates, so the user would have to rely on messier custom functions). What's worse is that currently cusotm formatting will silently handle the input incorrectly (either in rendering or in parsing).

### Other languages don't necessarily support these dates

Python simply only supports datetime values with a positive year. 

JavaScript will allow to construct a `Date` object with a negative year, but the support for parsing the negative ISO years is questionable. E.g. `moment` being otherwise a pretty good library will treat it as follows:
```
> moment('-1000-01-01').toISOString()
"1000-01-01T05:17:32.000Z"
> moment('-1000-01-01T00:00:00').toISOString()
null
```
Although we provide our own datatypes, this leaves the user who simply passes around JSON in a difficult situation.

In general, date/time support is complicated and anything but uniform across languages. Although almost every language will handle the core ISO8601 with 4-digit year correctly, this is not a certainty outside of that range.

### The usage of dates and timestamps

Any application that needs dates outside the range given by year 0000-9999 would probably need to define its own tools for doing so. Postgres has a lower bound on timestamp of 4713 BCE, so even if we supported correct ISO parsing and rendering of past dates, they would be of very limited utility. 

Realistically, most practical applications would be perfectly fine using the dates within a couple of centuries from the current moment and those are well-supported.

### Conclusion

It seems that the best way to proceed is to restrict the `datetime`, `local_datetime` and `local_date` to the range of positive years [1, 9999] and leave in place all of our checks and ISO conversion tricks we currently use. If we find some better solutions for these issues in the future, we can then either lift restrictions or introduce a new date/time datatype (equivalent to `numeric`).

PR #2251, opened 2021-02-19
Simplify starting EdgeDB on a random available network port

1st1:
While there, cleanup & fix minor bugs in pgcluster.py.

### DIFF
--- a/edb/server/cluster.py
+++ b/edb/server/cluster.py
@@ -20,14 +20,13 @@
 from __future__ import annotations
 
 import asyncio
-import errno
+import json
 import os
-import random
-import socket
 import subprocess
 import sys
 import tempfile
 import time
+import typing
 
 import edgedb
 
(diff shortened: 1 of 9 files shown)


edb/server/cluster.py
```diff
@@ -163,28 +137,53 @@ def init(self, *, server_settings=None):
...
+        stdout = subprocess.DEVNULL
+        if port == 0:
+            stdout = subprocess.PIPE
+
         self._daemon_process = subprocess.Popen(
             self._edgedb_cmd + extra_args,
-            stdout=sys.stdout, stderr=sys.stderr,
-            env=env)
+            stdout=stdout,
+            stderr=subprocess.DEVNULL,
```
elprans, line 169: Why is stderr going to /dev/null?

1st1: I'm not sure it's right to pipe it into sys.stderr. It was like that before and didn't help me debug while working on this PR one bit. I can revert to sys.stderr if you want.

edb/server/pgcluster.py
```diff
@@ -419,7 +404,11 @@ def start(self, wait=60, *, server_settings=None, port, **opts):
                     self._data_dir))
 
         extra_args = ['--{}={}'.format(k, v) for k, v in opts.items()]
-        extra_args.append('--port={}'.format(port))
+
+        # We connect to our PostgreSQL server via a Unix socket,
+        # with TCP disabled. Which means that we don't really
+        # care about the port number being unique or even available.
+        extra_args.append('--port=65535')
```
elprans, line 411: The Unix socket-only reasoning should allow us to run with the default port as well.  I don't see a reason why this is necessary.

edb/server/cluster.py
```diff
@@ -163,29 +137,54 @@ def init(self, *, server_settings=None):
...
+            cmd_port = 'auto'
+        else:
+            cmd_port = str(port)
 
         extra_args = ['--{}={}'.format(k.replace('_', '-'), v)
                       for k, v in settings.items()]
-        extra_args.append('--port={}'.format(self._effective_port))
+        extra_args.append(f'--port={cmd_port}')
+        if port == 0:
+            extra_args.append('--echo-runtime-info')
```
elprans, line 153: It would be really nice if the server supported passing a fileno here: `--echo-runtime-info=:4:`.  This way we would be able to leave the standard streams undisturbed.

elprans (approved):

PR #2252, opened 2021-02-19
Clamp datetime, local_datetime and local_date to the 1-9999 year range

elprans:
The default timestamp range of (4713 BC - 294276 AD) has problems:
Postgres isn't ISO compliant with years out of the 0-9999 range and
language compatibility is questionable.

See https://github.com/edgedb/edgedb/pull/2249#issuecomment-782353029
for an expanded rationale.


1st1 (approved):

vpetrovykh:
Looks good.

Issue #2246, opened 2021-02-17
CI triggered an intermittent bug


vpetrovykh:
CI triggered this intermittent bug:
```
ERROR: tearDownClass (test_server_proto.TestServerProtoDDL)
--------------------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/runner/work/edgedb/edgedb/edb/testbase/server.py", line 1011, in tearDownClass
    cls.loop.run_until_complete(drop_db())
  File "/opt/hostedtoolcache/Python/3.9.1/x64/lib/python3.9/asyncio/base_events.py", line 642, in run_until_complete
    return future.result()
  File "/home/runner/work/edgedb/edgedb/edb/testbase/server.py", line 1007, in drop_db
    await cls.admin_conn.execute(
  File "/opt/hostedtoolcache/Python/3.9.1/x64/lib/python3.9/site-packages/edgedb/asyncio_con.py", line 378, in execute
    await self._impl._protocol.simple_query(query)
  File "edgedb/protocol/protocol.pyx", line 567, in simple_query
edgedb.errors.ExecutionError: database "serverprotoddl_5003" is being accessed by other users
```

PR #2254, opened 2021-02-20
Handle Terminate messages appropriately

elprans:
The `Terminate` protocol message (`b'X'`) is the "negotiated" connection 
termination, when the client asks the server to disconnect and waits for 
the connection to drop.  This ensures that both the client and the server
are in agreement with respect to whether a connection between them still
exists, which is important for cases when a command such as
`DROP DATABASE` immediately follows the client disconnect.

Depends on edgedb/edgedb-python#166

Fixes: #2246

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -243,7 +243,7 @@ jobs:
 
     - name: Install edgedb-server and populate egg-info
       env:
-        CACHE_HIT: ${{ steps.postgres-cache.outputs.cache-hit }}
+        CACHE_HIT: ${{ steps.venv-cache.outputs.cache-hit }}
         BUILD_EXT_MODE: skip
       run: |
         if [[ "$CACHE_HIT" == "true" ]]; then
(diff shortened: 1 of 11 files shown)


edb/server/protocol/binary.pyx
```diff
@@ -298,7 +298,10 @@ cdef class EdgeConnection:
             self._transport.abort()
             self._transport = None
 
-    async def close(self):
+    def close(self):
+        if self.dbview is not None:
```
1st1, line 302: Hm, we need to put this in `connection_lost`.

elprans: Done. I've added the necessary bits to `abort()` and made sure it always gets called.  The `pinned_pgcon` abort business has been moved there as well.

.github/workflows/tests.yml
```diff
@@ -243,7 +243,7 @@ jobs:
 
     - name: Install edgedb-server and populate egg-info
       env:
-        CACHE_HIT: ${{ steps.postgres-cache.outputs.cache-hit }}
+        CACHE_HIT: ${{ steps.venv-cache.outputs.cache-hit }}
```
1st1, line 246: Let's move into a separate commit

1st1 (approved):

PR #2257, opened 2021-02-20
Discard Postgres connections on cancellation

elprans:
When we issue `pg_cancel_backend`, we must discard the Postgres
connection, because we cannot be completely sure about its state.
Postgres cancellation is based on Unix signals and is addressed to the whole
worker process and not a concrete operation.  The result is that we might be
racing with the currently running query and if that completes before the
cancellation signal reaches the backend, we'll be setting a trap for the
_next_ query that is unlucky enough to pick up this Postgres backend
from the connection pool.

### DIFF
--- a/edb/server/server.py
+++ b/edb/server/server.py
@@ -671,6 +671,12 @@ class Server:
         finally:
             self._release_sys_pgcon()
 
+    async def _cancel_and_discard_pgcon(self, pgcon, dbname) -> None:
+        try:
+            await self._cancel_pgcon_operation(pgcon)
+        finally:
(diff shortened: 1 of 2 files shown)


edb/server/server.py
```diff
@@ -671,6 +671,12 @@ async def _cancel_pgcon_operation(self, pgcon) -> bool:
         finally:
             self._release_sys_pgcon()
 
+    async def _cancel_and_discard_pgcon(self, pgcon, dbname) -> None:
+        try:
```
1st1, line 675: Please add a comment clarifying why we want to cancel before terminating.

1st1 (approved):

Issue #2265, opened 2021-02-23
Deleting abstract type doesn't work


tailhook:
- EdgeDB Version: 1.0b1-2021022220

Steps to Reproduce:

1. `DELETE AbstractType` returns empty set and doesn't delete anything
2. `DELETE SomeObject.some_link` where `some_link` refers to abstract type doesn't work too
3. `DELETE ConcreteType` works fine as well as `DELETE SomeObject.some_link[IS ConcreteType]`

I'm not sure about concrete types, whether they delete subclasses too.

PR #2270, opened 2021-02-25
Make sure DELETE propagates to child type properly

elprans:
`DELETE Foo` should remove all objects of type `Foo`, including all
objects of subtypes of `Foo`.  This currently doesn't work because a
"deletion" derived type does not get generated for `DELETE` subjects,
which, in turn, leads to a typeref that doesn't include type descendant
information.

Fixes: #2265

### DIFF
--- a/tests/test_edgeql_delete.py
+++ b/tests/test_edgeql_delete.py
@@ -26,13 +26,16 @@ class TestDelete(tb.QueryTestCase):
     SETUP = """
         START MIGRATION TO {
             module test {
-                type DeleteTest {
+                type LinkingType {
+                    multi link objs -> AbstractDeleteTest;
+                };
+
+                abstract type AbstractDeleteTest {
                     property name -> str;
                 };
 
-                type DeleteTest2 {
-                    property name -> str;
-                };
+                type DeleteTest extending AbstractDeleteTest;
+                type DeleteTest2 extending AbstractDeleteTest;
             };
         };
         POPULATE MIGRATION;
@@ -479,7 +482,7 @@ class TestDelete(tb.QueryTestCase):
                     (DELETE DeleteTest2);
             ''')
 
-    async def test_edgeql_update_in_conditional_bad_02(self):
+    async def test_edgeql_delete_in_conditional_bad_02(self):
         with self.assertRaisesRegex(
                 edgedb.QueryError,
                 'DELETE statements cannot be used'):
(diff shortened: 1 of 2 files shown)


tests/test_edgeql_delete.py
```diff
@@ -494,3 +498,50 @@ async def test_edgeql_update_in_conditional_bad_02(self):
...
+            INSERT DeleteTest { name := 'child of abstract 1' };
+            INSERT DeleteTest2 { name := 'child of abstract 2' };
+        """)
+
+        await self.assert_query_result(
+            r"""
+                WITH
+                    MODULE test,
+                    D := (
+                        DELETE
```
1st1, line 515: Do we have tests for UPDATEs on abstract types?

tests/test_edgeql_delete.py
```diff
@@ -494,3 +498,50 @@ async def test_edgeql_update_in_conditional_bad_02(self):
...
+                SELECT D { name } ORDER BY .name;
+            """,
+            [{
+                'name': 'child of abstract 1'
+            }, {
+                'name': 'child of abstract 2'
+            }],
+        )
+
+    @test.xfail('DELETE trips up link target deletion constraint improperly')
```
msullivan, line 529: Do you want me to take a look at the follow-up, here?

msullivan (approved):
This was quite a bit less painful than I feared; I guess we had all the infrastructure, we just weren't tickling it right?

tests/test_edgeql_delete.py
elprans: I've thought about it, and I'm not 100% sure it's a bug, because you can rewrite things as 

```
WITH ToDelete := (SELECT LinkingObject.link)
DELETE ToDelete
```

Perhaps we could special-case straight `DELETE LinkingObject.link`.

tests/test_edgeql_delete.py
elprans: We do.  The abstract bit is irrelevant, this was broken for non-abstract inheritance too.

elprans:
> I guess we had all the infrastructure, we just weren't tickling it right

Yes, the explanation is in the commit message: `typeref` wasn't generated with descendants and so the dml machinery wasn't generating the correct union of dml parts.

tests/test_edgeql_delete.py
msullivan: Actually, wait, shouldn't both of these fail? The link isn't set to `ALLOW`.

PR #2271, opened 2021-02-25
docs: Update the changelog & the clients index page

1st1:
* Mention new versions of edgedb-js, edgedb-python, and edgedb-go
  and how RFC1004 methods are called in them;

* Add a link to edgedb-go to the clients page.


vpetrovykh (approved):
LGTM

PR #2127, opened 2021-01-18
Make intro a bit friendlier

Dhghomon:
I really want to add the quote from We Can Do Better Than SQL as IMO it explains EdgeDB best in the fewest number of words - would have been great to see front and centre in September when first learning about EdgeDB. I added it as text but ideally it would be cool to replace the horn image with an image containing the quote (maybe a link to the blog post too when clicked):

https://www.edgedb.com/blog/we-can-do-better-than-sql/

PR #2273, opened 2021-02-25
Update the docs intro

1st1:


1st1:
Supersedes https://github.com/edgedb/edgedb/pull/2127

docs/intro.rst
```diff
@@ -3,15 +3,27 @@
 Introduction
 ============
 
+.. pull-quote::
+
+    The relational model is still the most effective method
+    of representing data. And SQL as a declarative,
```
elprans, line 9: A reference to SQL out of context feels a bit weird here, let's drop the quote for now and find better phrasing.

1st1: OK

PR #2274, opened 2021-02-26
Fix a bunch of invalid serializations involving shapes and collections

### LINKED ISSUE #1818: array_unpack error in some JSON output queries
`FOR X in {array_agg(0)} UNION (SELECT array_unpack(X));` fails in json mode with
`      edb.errors.InternalServerError: function unnest(json) does not exist`

msullivan:
A common theme in these issues is an object being contained inside a
collection such that the shape can't be produced near the top-level
but instead must live inside the collection.

 * When injecting default props into a non-view object set (which
   feels sketchy, but we do it sometimes), make sure to only do it
   *once* if the set is referenced multiple times.
 * Make sure things work when an object is output with a shape and
   without a specified one in the same query, and type name injection
   is on.
 * Make sure things work when an object apears in a tuple and is
   is also referenced with a shape through tuple projection.
 * Fix array_agg and similar operating on shapes; something can
   be used as a value and serialized, so it isn't safe to
   put a serialized value as the value aspect. We need to
   compile *both*. Fixes #1818.

An earlier iteration of this PR broke test_edgeql_calls_35, because of
limitations in how we generate tuple casts. I added an xfailed test
for an already existing related failure to that bug.

### DIFF
--- a/edb/pgsql/compiler/__init__.py
+++ b/edb/pgsql/compiler/__init__.py
@@ -90,6 +90,7 @@ def compile_ir_to_sql_tree(
 
         _ = context.CompilerContext(initial=ctx)
         ctx.singleton_mode = singleton_mode
+        ctx.expr_exposed = True
         qtree = dispatch.compile(ir_expr, ctx=ctx)
 
     except Exception as e:  # pragma: no cover
--- a/edb/pgsql/compiler/relgen.py
+++ b/edb/pgsql/compiler/relgen.py
@@ -2463,7 +2463,8 @@ def process_set_as_agg_expr(
             ctx=xctx)
 
     # Though if the result type is a scalar, the value should be good
-    # enough, so don't generate bad code.
+    # enough, so don't generate a bunch of unnessecary code to produce
+    # a serialized value when we can use value.
     if (
         output.in_serialization_ctx(ctx=ctx)
         and not irtyputils.is_scalar(ir_set.typeref)
--- a/edb/pgsql/compiler/relgen.py
+++ b/edb/pgsql/compiler/relgen.py
@@ -179,6 +179,13 @@ def get_set_rvar(
                 subctx.path_scope[path_id] = scope_stmt
             relctx.update_scope(ir_set, stmt, ctx=subctx)
 
+        # If the set is a binding, we need to treat it as exposed
+        # because we don't know at this point whether it will be
+        # reused in an exposed context.
+        # (We could avoid this with some more analysis, probably.)
+        if ir_set.is_binding:
+            subctx.expr_exposed = None
+
         rvars = _get_set_rvar(ir_set, ctx=subctx)
 
         if optional_wrapping:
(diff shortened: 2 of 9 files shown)


edb/pgsql/compiler/relgen.py
```diff
@@ -2386,32 +2406,77 @@ def process_set_as_agg_expr(
...
-        stmt, ir_set.path_id, set_expr, env=ctx.env)
+    cctx = ctx.subrel() if wrapper else ctx.new()
+    with cctx as xctx:
+        xctx.expr_exposed = False
+        process_set_as_agg_expr_inner(
+            ir_set, xctx.rel, aspect='value', wrapper=wrapper,
+            ctx=xctx)
+
+    # Though if the result type is a scalar, the value should be good
+    # enough, so don't generate bad code.
```
elprans, line 2466: Bad code?  I guess `compile_output` will force scalar array serialization anyway (I'd mention that in the comment), but I wonder why this formulation needs to exclude scalars specifically.

tests/test_edgeql_calls.py
```diff
@@ -1243,6 +1243,36 @@ async def test_edgeql_calls_35(self):
             ]
         )
 
+    @test.xfail('''
+       Fails with 'ISE: cannot cast type record to <stuff>', I think
+       because the subcomponents need to be cast also.
```
elprans, line 1248: Yeah, this is an annoying Postgres quirk: it refuses to cast composite types even if they're binary-compatible!  We are handling this elsewhere, e.g:

https://github.com/edgedb/edgedb/blob/6891b88ad419709dad5fea7b4e9c7281ec2f9019/edb/edgeql/compiler/viewgen.py#L707-L727

But we forget to do the same for function returns here:

https://github.com/edgedb/edgedb/blob/6891b88ad419709dad5fea7b4e9c7281ec2f9019/edb/schema/functions.py#L1428-L1436

(`implicitly_castable` does not necessarily mean that it is implicitly castable in Postgres, which is what `is_type_compatible` checks)

elprans (approved):

edb/pgsql/compiler/relgen.py
msullivan: By bad code I just mean "suboptimal code"; if it's just scalars, then we're fine not providing a "serialized" output, which means we can skip compiling it.

edb/pgsql/compiler/__init__.py
```diff
@@ -90,6 +90,7 @@ def compile_ir_to_sql_tree(
 
         _ = context.CompilerContext(initial=ctx)
         ctx.singleton_mode = singleton_mode
+        ctx.expr_exposed = True
```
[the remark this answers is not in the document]
msullivan: Unless I'm missing something, it looks like those are all in the edgeql compiler; there's one in the pgsql side that is `and`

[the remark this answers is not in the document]
elprans: Oh, that's right.

tests/test_edgeql_calls.py
msullivan: Going to merge and then will try to address in a follow-up

PR #2282, opened 2021-02-27
Fix metadata-related RDS issues

elprans:


fantix:
Thanks for the fast fix! I'll test on this PR soon.

fantix:
I'm not sure but maybe we should retry the lock polling by default? I can easily reproduce a locking race by running these tests in parallel (on RDS):

```shell
$ edb test -v -j2 -k test_edgeql_ddl_role
=======================================================================================================
ERROR: test_edgeql_ddl_role_01 (test_edgeql_ddl.TestEdgeQLDDL)
-------------------------------------------------------------------------------------------------------
  File "edb/testbase/server.py", line 127, in wrapper
    self.loop.run_until_complete(
  File "asyncio/base_events.py", line 642, in run_until_complete
    return future.result()
  File "tests/test_edgeql_ddl.py", line 6261, in test_edgeql_ddl_role_01
    await self.con.execute(r"""
  File "edgedb/asyncio_con.py", line 359, in execute
    await self._impl._protocol.simple_query(
  File "edgedb/protocol/protocol.pyx", line 578, in simple_query
edgedb.errors.TransactionSerializationError: Cannot serialize global DDL: efdea6f2-803e-11eb-b872-8962d99913de
```

If I replace the `-j2` with `-j1`, it passes perfectly.

I think this means any concurrent `CREATE ROLE` may hit a lock race on RDS.

fantix (approved):

PR #2283, opened 2021-02-27
Insert explicit casts when returning non-compatible tuples from a function

msullivan:
Also, update the type compatibility check to be recursive, in case
nested tuples don't match.

### DIFF
--- a/edb/schema/types.py
+++ b/edb/schema/types.py
@@ -2376,3 +2376,36 @@ def materialize_type_in_attribute(
         )
 
     return schema
+
+
+def is_type_compatible(
+    type_a: Type,
+    type_b: Type,
+    *,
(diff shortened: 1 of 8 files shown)


edb/schema/types.py
```diff
@@ -2376,3 +2376,36 @@ def materialize_type_in_attribute(
...
     return schema
+
+
+def is_type_compatible(
+    type_a: Type,
+    type_b: Type,
+    *,
+    schema: s_schema.Schema,
+) -> bool:
+
```
elprans, line 2387: Let's add a docstring here, explaining that this is needed to make EdgeQL implicit casts explicit in SQL, which has differing semantics.

elprans (approved):

Issue #2285, opened 2021-02-28
1-beta1 is slow compared to 1-alpha7


unmade:
### The Problem

After upgrading to `1-beta1` my tests run very slow - almost 3 minutes. With `1-alpha7` test suit runs in 30 seconds.

Do you know what could be the problem or how can I trace what causes this slowdown?

### My Setup:

- Test suit is running on macOS BigSur (rMBP 15 early 2013) and also in GitHub Actions
- The [project](https://github.com/unmade/shelf-back) is in Python with latest at the moment (0.13.0) `edgedb-python`
- EdgeDB run with [docker-compose](https://github.com/unmade/shelf-back/blob/master/docker-compose.yml) and with [services in GitHub Actions](https://github.com/unmade/shelf-back/blob/master/.github/workflows/test.yml#L18)
- Test suit includes:
    - [creating test database](https://github.com/unmade/shelf-back/blob/master/tests/conftest.py#L75-L90) - (runs once)
    - [creating connection pool](https://github.com/unmade/shelf-back/blob/master/tests/conftest.py#L93-L101) (runs once)
    - [migration](https://github.com/unmade/shelf-back/blob/master/tests/conftest.py#L104-L110) (runs once)
    - [tables clean up](https://github.com/unmade/shelf-back/blob/master/tests/conftest.py#L119-L123) (runs after each test)

elprans:
Thanks for a great report, @unmade!  The fix will be rolled into a b1 fixup revision tonight.

PR #2286, opened 2021-02-28
Make sure query source cache doesn't invalidate on COMMIT

elprans:
The server schema state machinery determines if query source caches are
still valid by the presence of user or global schema in the returned
query units.  The `COMMIT` statement handler, however, is currently
_always_ including the schemas in the returned query unit, which results
in improper query source invalidation on every `COMMIT`.

Fixes: #2285.

### DIFF
--- a/edb/server/compiler/dbstate.py
+++ b/edb/server/compiler/dbstate.py
@@ -427,7 +427,7 @@ class Transaction:
         return self._current.user_schema
 
     def get_user_schema_if_updated(self) -> Optional[s_schema.FlatSchema]:
-        if self._current.user_schema == self._state0.user_schema:
+        if self._current.user_schema is self._state0.user_schema:
             return None
         else:
             return self._current.user_schema
--- a/edb/server/compiler/dbstate.py
+++ b/edb/server/compiler/dbstate.py
@@ -426,8 +426,20 @@ class Transaction:
     def get_user_schema(self) -> s_schema.FlatSchema:
         return self._current.user_schema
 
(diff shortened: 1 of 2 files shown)


edb/server/compiler/dbstate.py
```diff
@@ -426,9 +426,21 @@ def get_schema(self, std_schema: s_schema.FlatSchema) -> s_schema.Schema:
     def get_user_schema(self) -> s_schema.FlatSchema:
         return self._current.user_schema
 
+    def get_user_schema_if_updated(self) -> Optional[s_schema.FlatSchema]:
+        if self._current.user_schema == self._state0.user_schema:
```
1st1, line 430: I'd use `is` -- that's what `==` does here after all.

edb/server/compiler/dbstate.py
```diff
@@ -426,9 +426,21 @@ def get_schema(self, std_schema: s_schema.FlatSchema) -> s_schema.Schema:
...
+            return None
+        else:
+            return self._current.user_schema
+
     def get_global_schema(self) -> s_schema.FlatSchema:
         return self._current.global_schema
 
+    def get_global_schema_if_updated(self) -> Optional[s_schema.FlatSchema]:
+        if self._current.global_schema == self._state0.global_schema:
```
1st1, line 439:
```suggestion
        if self._current.global_schema is self._state0.global_schema:
```

edb/server/compiler/dbstate.py
1st1, line 430:
```suggestion
        if self._current.user_schema is self._state0.user_schema:
```

1st1 (approved):

PR #2302, opened 2021-03-03
Make semicolons completely optional in SDL

elprans:
Currently, we require a semicolon to terminate non-block declarations in
SDL, i.e. any declaration that is not followed by a `{`.  This annoyed
me for quite some time, so I started looking if we actually _need_ to
have that terminator, and, as it turns out, we don't really.  Once I
removed the short statement terminator there was only one conflict: the
`ON` keyword being 1) unreserved; and 2) constraint's optional
`on (subject)` clashing with `on target delete`, e.g.:

    link foo {
        constraint foo
        on target delete
     }

vs

     link foo {
         constraint foo on (blah)
     }

This has been easy to fix by 1) making `ON` reserved; and 2) lexing
`on target` as a single token (like we do with `set type` and `named only`).

This also allowed removing a bunch of weird hacks in the SDL grammar, and
there's potential for further cleanup.


1st1 (approved):

tests/test_schema_syntax.py
```diff
@@ -111,7 +111,7 @@ def test_eschema_syntax_semicolon_02(self):
     def test_eschema_syntax_semicolon_03(self):
         """
         abstract type test::OwnedObject {
-            required link owner -> User;
+            required link owner -> User
```
1st1, line 114: Can you add a test with a default expression? Also test_eschema_syntax tests all have semicolons in them, can you randomly drop a few from a few tests?

tailhook (changes requested):
Please do not merge this. Details are in the slack, so that we don't split discussion into many pieces.

PR #2308, opened 2021-03-03
Fix use of contexts for constraint errors

### LINKED ISSUE #2305: Invalid constraint definition error without indication of where the error is
This error doesn't give the user much help locating the error in their file. It would be nice to have a line number and column.

```bash
$ edgedb -I workout create-migration
ERROR: InvalidConstraintDefinitionError: constraints cannot contain paths with more than one hop
```

schema.esdl
```
module default {
	type Exercise {
			constraint exclusive;
		};
	}
	type Set_ {
		link workout := .<sets[IS Workout];
		required link exercise -> Exercise;
		required property index -> int16 {
			constraint min_value(0);
		};
		required property weight -> float32;
		required property reps -> int16;
		required property completed -> bool {
			default := false;
		};

		constraint exclusive on ( (.index, .workout) );
	}
	type Workout {
		multi link sets -> Set_ {
			constraint exclusive;
		}
		required property started_at -> datetime {
			constraint exclusive;
		}
	}
};
```

- EdgeDB Version: package:1.0b2.dev5478+g25318fe0e-2021022200~focal.nightly
- EdgeDB CLI Version: edgedb-cli 1.0.0-beta.1
- OS Version: Linux 5.8.0-44-generic edgedb/edgedb-cli#50~20.04.1-Ubuntu SMP Wed Feb 10 21:07:30 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

msullivan:
Errors from constraints are missing contexts in most cases
(... whenever the expr text isn't __subject__). Always just use the
context we've got.

Fixes #2305.

### DIFF
--- a/edb/schema/constraints.py
+++ b/edb/schema/constraints.py
@@ -885,11 +885,6 @@ class CreateConstraint(
 
         attrs['args'] = args
 
-        if expr.text == '__subject__':
-            expr_context = sourcectx
-        else:
-            expr_context = None
-
         assert subject is not None
         path_prefix_anchor = (
             qlast.Subject().name if isinstance(subject, s_types.Type)
@@ -915,7 +910,7 @@ class CreateConstraint(
                 f'{name} constraint expression expected '
                 f'to return a bool value, got '
                 f'{expr_type.get_verbosename(schema)}',
-                context=expr_context
+                context=sourcectx
             )
 
         if subjectexpr is not None:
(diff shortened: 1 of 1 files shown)


elprans (approved):
Makes sense, although the error in #2305 is confusing in the presence of computables, since the `subject` itself isn't "too many hops", but the underlying computable expression is.  Can we do something about that?

msullivan:
I'm not sure if there is a simple approach; the contexts on the refs point to strings, and the compiled sets have thrown away the information that they came from computables, as far as I can tell.

PR #2313, opened 2021-03-04
Fix a cluster of constraint and rename issues

### LINKED ISSUE #2307: "error: constraint 'std::max_len_value' does not exist" during create-migration
EdgeDB Version: EdgeDB 1.0-beta.1+g2eee1afe4.d20210301

Changing schema from:
```
abstract type Event {
  required property createdAt -> datetime {
  }

  required link user -> User;
}

type Post extending Event {
  required property content -> str {
    constraint min_len_value(1);
    constraint max_len_value(280);
  }
}
```
to: 
```
abstract type Event {
  required property createdAt -> datetime {
  }

  required link user -> User;
}

abstract type HasContent {
  required property content -> str {
    constraint min_len_value(1);
    constraint max_len_value(280);
  }
}

type Post extending Event, HasContent {
}

type Reply extending Event, HasContent {
  required link post -> Post;
}
```

Run `edgedb create-migration`:
```
did you create object type 'default::HasContent'? [y,n,l,c,b,s,q,?]
y
did you alter object type 'default::Post'? [y,n,l,c,b,s,q,?]
y
did you create object type 'default::Reply'? [y,n,l,c,b,s,q,?]
y
did you drop constraint 'std::max_len_value' of property 'content'? [y,n,l,c,b,s,q,?]
n
did you alter property 'content' of object type 'default::Post'? [y,n,l,c,b,s,q,?]
y
error: constraint 'std::max_len_value' does not exist
  ┌─ query:3:9
  │  
3 │ ╭         ALTER CONSTRAINT std::max_len_value(280) ON (std::len(<std::str>__subject__)) {
4 │ │             DROP OWNED;
5 │ │         };
  │ ╰─────────^ error

Rolling back last operation...
did you alter property 'content' of object type 'default::Post'? [y,n,l,c,b,s,q,?]
n
did you alter property 'content' of object type 'default::Post'? [y,n,l,c,b,s,q,?]
```

### LINKED ISSUE #2311: Bogus rename prompts for constraints when a parent type is renamed
Migrating from

```
    abstract type Named {
    };

    type Org;

    abstract type OrgBound {
        required link org -> Org;
    };

    abstract type OrgUniquelyNamed
        extending Named, OrgBound
    {
        constraint exclusive on ((.name, .org))
    }
```

to

```
    abstract type Named {
    };

    type Org;
    abstract type Resource;

    abstract type OrgBound {
        required link org -> Org;
    };

    abstract type OrgUniquelyNamedResource
        extending Named, Resource, OrgBound
    {
        delegated constraint exclusive on ((.name, .org))
    }
```

I get:

```
did you rename constraint 'std::exclusive' of object type 'default::OrgUniquelyNamed' to 'std::exclusive'? [y,n,l,c,b,s,q,?]
l
Following DDL statements will be applied:
    ALTER TYPE default::OrgUniquelyNamed {
        ALTER CONSTRAINT std::exclusive ON ((.name, .org)) {
            SET DELEGATED;
        };
    };
```

which is triggered because the generated mangled name of the `std::exclusive` has changed, but this isn't a visible name and so a prompt should not be generated.

I suspect a similar issue exists for indexes.

msullivan:
Some things fixed:
 * Track renames while generating ASTs so we don't emit references
   to renamed objects
 * Make sure to include subjectexpr and args when generating
   ALTER CONSTRAINT asts
 * Correctly use the nearest *abstract* constraint when generating
   constraint names in all cases
 * Consistently handle which constraint properties are marked
   as inherited. (Though there is still something of a mess there.)
 * Handle renames of abstract constraints in situations with
   inheritance properly

Fixes #2307. Fixes #2311. Fixes some more not filed.

### DIFF
--- a/edb/schema/constraints.py
+++ b/edb/schema/constraints.py
@@ -139,7 +139,7 @@ class Constraint(
         obj = schema.get(name, type=Constraint)
 
         if not obj.generic(schema):
-            base = obj.get_bases(schema).objects(schema)[0]
+            base = obj.get_nearest_generic_parent(schema)
             base_name = context.get_obj_name(schema, base)
 
             quals = list(sn.quals_from_fullname(name))
@@ -424,6 +424,43 @@ class ConstraintCommand(
 
         return args
 
+    @classmethod
+    def as_inherited_ref_ast(
+        cls,
+        schema: s_schema.Schema,
+        context: sd.CommandContext,
+        name: sn.Name,
+        parent: so.Object,
+    ) -> qlast.ObjectDDL:
+        assert isinstance(parent, Constraint)
+        astnode_cls = cls.referenced_astnode  # type: ignore
+        nref = cls.get_inherited_ref_name(schema, context, parent, name)
+        args = []
+
+        parent_args = parent.get_args(schema)
+        if parent_args:
+            parent_args = parent.get_args(schema)
+            assert parent_args is not None
+            for arg_expr in parent_args:
+                arg = edgeql.parse_fragment(arg_expr.text)
+                args.append(arg)
+
+        subj_expr = parent.get_subjectexpr(schema)
+        if (
+            subj_expr is None
+            # Don't include subjectexpr if it was inherited from an
+            # abstract constraint.
+            or parent.get_nearest_generic_parent(
+                schema).get_subjectexpr(schema) is not None
+        ):
+            subj_expr_ql = None
+        else:
+            subj_expr_ql = edgeql.parse_fragment(subj_expr.text)
+
+        astnode = astnode_cls(name=nref, args=args, subjectexpr=subj_expr_ql)
+
+        return cast(qlast.ObjectDDL, astnode)
+
     def compile_expr_field(
         self,
         schema: s_schema.Schema,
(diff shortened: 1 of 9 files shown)


edb/schema/constraints.py
```diff
@@ -534,6 +571,7 @@ def get_ref_implicit_base_delta(
             if not b.get_abstract(schema)
             and b.generic(schema) and b.get_name(schema) != default_base
         ]
+        # assert not explicit_bases
```
elprans, line 574: Uncomment or delete?

edb/schema/constraints.py
```diff
@@ -809,6 +847,13 @@ def _populate_concrete_constraint_attrs(
         from edb.ir import ast as ir_ast
         from edb.ir import utils as ir_utils
 
+        bases = self.get_resolved_attribute_value(
+            'bases', schema=schema, context=context,
+        )
+        direct_base = bases.objects(schema)[0]
+        if not direct_base.generic(schema):
```
elprans, line 854: A comment here on why we're short-circuiting?

edb/schema/constraints.py
msullivan: Let's try uncommenting it, and if that works I'll delete the apparently dead code above it.

(The explicit_bases condition requires `get_abstract` and `generic` to disagree and I don't understand when that would happen?)

msullivan: :heavy_check_mark:

msullivan:
Updated to fix some more bugs and to get rid of the custom `_canonicalize` and the `_maybe_fix_name` atrocity

elprans (approved):
Nice to see this untangled!

PR #2329, opened 2021-03-06
Fix tuples in multi props and IN expressions

### LINKED ISSUE #2255: edb.errors.InternalServerError: column "target" is of type edgedbpub."71637d60-9aed-6158-3c48-7316b93c0197_t" but expression is of type bigint
- EdgeDB Version: 1.0a8.dev5340+g76e0688b-2021012101~focal.nightly
- OS Version: Ubuntu

```
START MIGRATION TO {
    module default {
        type Movie {
            multi property tags -> tuple<int64, str>;
        }
    }
}
```
```
INSERT Movie {title := 'hey', tags := {(10, 'foo'), (20, 'bar')}};
```
```
ERROR: InternalServerError: column "target" is of type edgedbpub."71637d60-9aed-6158-3c48-7316b93c0197_t" but expression is of type bigint
  Server traceback:
      edb.errors.InternalServerError: column "target" is of type edgedbpub."71637d60-9aed-6158-3c48-7316b93c0197_t" but expression is of type bigint
```

msullivan:
When inserting tuples into multi properties, we need to output
them as values first.

Expressions like `(1, 2) IN Foo.multi_prop` were broken because if a
tuple literal appears on the LHS of an `= ANY (` operation, pgsql
wants the subquery to contain one column per tuple field, which it won't
if we're getting the tuple straight from a table.

Instead of worrying about making the two sides agree, I just sidestepped
the issue by changing the compilation approach to compile
`x IN y` as approximately `EXISTS (SELECT ... WHERE x_output = y_output)`.

(Do we think this will cause any performance regressions? If it will
be a problem, I could instead worry about making the two sides agree.)

Also while looking at this I noticed that the optimized special-case
for `IN array_unpack(` was broken if the argument had multi
cardinality, so I fixed that also.

Fixes #2255.

### DIFF
--- a/edb/pgsql/compiler/relgen.py
+++ b/edb/pgsql/compiler/relgen.py
@@ -1239,7 +1239,7 @@ def process_set_as_membership_expr(
         left_arg, right_arg = (a.expr for a in expr.args)
 
         newctx.expr_exposed = False
-        left_expr = dispatch.compile(left_arg, ctx=newctx)
+        left_out = dispatch.compile(left_arg, ctx=newctx)
 
         right_arg = irutils.unwrap_set(right_arg)
         # If the right operand of [NOT] IN is an array_unpack call,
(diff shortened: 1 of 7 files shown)


edb/pgsql/compiler/relgen.py
```diff
@@ -1253,37 +1253,39 @@ def process_set_as_membership_expr(
         else:
             is_array_unpack = False
 
-        with newctx.subrel() as _, _.newscope() as subctx:
-            dispatch.compile(right_arg, ctx=subctx)
-            pathctx.get_path_value_output(
-                subctx.rel, right_arg.path_id,
-                env=subctx.env)
+        # We compile `x [NOT] IN y` into essentially
+        # `[NOT] EXISTS (SELECT ... WHERE x = y)`.
```
elprans, line 1257: Hm, a quick test shows that Postgres generates the same {semi|anti}-join plan for both formulations.  This is not a guarantee that this'll happen in every query, but seems OK to me at the moment.

elprans (approved):

Issue #2333, opened 2021-03-08
Problems comparing empty tuples/arrays


Fogapod:
- EdgeDB Version: `1-beta1`

I've been experimenting with `DISTINCT` and discovered a few things:
1) When different types are passed to distinct, it leaks implementation details (UNION) in error:
```
edgedb> select distinct {1, "1"};
ERROR: QueryError: operator 'UNION' cannot be applied to operands of type 'std::int64' and 'std::str'
  Hint: Consider using an explicit type cast or a conversion function.
```
2) It has some problems comparing empty tuples:
```
edgedb> select distinct {(), (), ()};
{(), (), ()}
```
Note: this doesn't work with empty arrays, compiler makes me annotate 2nd argument.

Trying to figure out what's going on, I tried comparing empty tuples manually. This is what I've got:
```
edgedb> select <tuple<str>>{} = ("",);
{}
edgedb> select <array<str>>{} = ["",];
{}
```

Not sure if these are issues that should be fixed or I misunderstood something.

elprans:
> When different types are passed to distinct, it leaks implementation details (UNION) in error:

This is not about `DISTINCT`, you cannot `UNION` two incompatible types, and the set literal `{ ... }` is a syntax sugar over the `UNION` operator.

>     2. It has some problems comparing empty tuples:
> 
> 
> ```
> edgedb> select distinct {(), (), ()};
> {(), (), ()}
> ```

This is a genuine bug.


> Trying to figure out what's going on, I tried comparing empty tuples manually. This is what I've got:
> 
> ```
> edgedb> select <tuple<str>>{} = ("",);
> {}
> edgedb> select <array<str>>{} = ["",];
> {}
> ```
> 
> Not sure if these are issues that should be fixed or I misunderstood something.

You are comparing an **empty set** of tuples/arrays vs a set containing **one** empty tuple or array.  They are not equivalent.  It's like comparing `None` to `[]` in Python.

Fogapod:
> This is not about `DISTINCT`, you cannot `UNION` two incompatible types, and the set literal `{ ... }` is a syntax sugar over the `UNION` operator.

I'm saying that `DISTINCT` operator produces error mentioning `UNION`, this is not what I expected it to do.

> You are comparing an **empty set** of tuples/arrays vs a set containing **one** empty tuple or array. They are not equivalent. It's like comparing `None` to `[]` in Python.

You can compare `None` and `[]` in Python and get `False`, not empty array or set.
[...]

PR #2335, opened 2021-03-08
Fix SELECT DISTINCT on empty and nested tuples

msullivan:
* For nested tuples, we need to pull out all the columns recursively,
   instead of just the top level ones.
 * For empty tuples, instead of trying to use DISTINCT, just set LIMIT 1.
   (Since the only question is whether the result is an empty set or a set
   containing the empty tuple.)

The addresses the actual bug in #2333, but there is some discussion of
potential error message improvements there also.

### DIFF
--- a/tests/test_edgeql_expressions.py
+++ b/tests/test_edgeql_expressions.py
@@ -4177,6 +4177,11 @@ aa \
 
     async def test_edgeql_expr_setop_12(self):
         await self.assert_query_result(
+            r'''SELECT DISTINCT {(), ()};''',
+            [[]],
+        )
+
+        await self.assert_query_result(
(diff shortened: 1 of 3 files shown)


tests/test_edgeql_expressions.py
```diff
@@ -4175,6 +4175,22 @@ async def test_edgeql_expr_setop_11(self):
             len(everything), len(distinct),
             'DISTINCT len(ObjectType.name) failed to filter out dupplicates')
 
+    async def test_edgeql_expr_setop_12(self):
+        await self.assert_query_result(
+            r'''SELECT DISTINCT (SELECT ({1,2,3}, ()) FILTER .0 > 1).1;''',
```
elprans, line 4180: Let's also add the explicit set literal case `SELECT DISTINCT {(), ()}` just in case set literals get some special handling.

elprans (approved):

PR #2336, opened 2021-03-08
Fix crash on empty sets inside set literals

### LINKED ISSUE #2154: ISE on empty set within set literal
- EdgeDB Version: `1.0-alpha.8+dev.5351.g93a3c34c4.d20210126`

Repro:

```
SELECT {False, <bool>{}};
```

msullivan:
EmptySet is not a BaseConstant, so it shouldn't appear in a
ConstantSet. Filter out empty sets when evaluating unions in the
constant evaluator.

Fixes #2154.

### DIFF
--- a/edb/ir/staeval.py
+++ b/edb/ir/staeval.py
@@ -166,20 +166,31 @@ def evaluate_OperatorCall(
 
 def _evaluate_union(
         opcall: irast.OperatorCall,
-        schema: s_schema.Schema) -> irast.ConstantSet:
+        schema: s_schema.Schema) -> irast.ConstExpr:
 
-    elements: List[irast.ConstExpr] = []
+    elements: List[irast.BaseConstant] = []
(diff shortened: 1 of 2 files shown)


edb/ir/staeval.py
```diff
@@ -166,20 +166,31 @@ def evaluate_OperatorCall(
...
-        typeref=next(iter(elements)).typeref,
-    )
+    if elements:
+        return irast.ConstantSet(
+            elements=tuple(elements),
+            typeref=next(iter(elements)).typeref,
+        )
+    else:
+        # We can only get an empty result if one of the arguments
+        # was empty, so empty_set must have been set.
```
elprans, line 192: The wording is a bit confusing: "We get an `EmptySet` if the UNION consisted exclusively of empty set literals" reads better IMO.

elprans (approved):

Issue #2296, opened 2021-03-01
Lots of failures migrating to empty schemas


msullivan:
After manually having run into issues several times trying to drop everything using a migration, I modified `_assert_migration_equivalence` to always migrate to an empty schema at the end (`migrations = migrations + ['']`), and it shakes out 9 test failures.

msullivan:
The failing tests are:
test_schema_migrations_equivalence_35
test_schema_migrations_equivalence_annotation_05
test_schema_migrations_equivalence_annotation_06
test_schema_migrations_equivalence_collections_26
test_schema_migrations_equivalence_compound_01
test_schema_migrations_equivalence_constraint_03
test_schema_migrations_equivalence_linkprops_08
test_schema_migrations_equivalence_linkprops_10
test_schema_migrations_equivalence_collections_15

elprans:
> I modified `_assert_migration_equivalence` to always migrate to an empty schema at the end

That's a good idea.

PR #2340, opened 2021-03-09
Implement DROP OWNED for annotation values

msullivan:
Work on #2296.

### DIFF
--- a/edb/schema/annos.py
+++ b/edb/schema/annos.py
@@ -70,15 +70,6 @@ class AnnotationValue(
     value = so.SchemaField(
         str, compcoef=0.909)
 
-    # Overloaded to mark it as not inheritable, since I found it
-    # basically impossible to get the inherited behavior to work right
-    # otherwise.
-    final = so.SchemaField(
-        bool,
-        inheritable=False,
-        compcoef=0.909,
-    )
-
     def __str__(self) -> str:
         return '<{}: at 0x{:x}>'.format(self.__class__.__name__, id(self))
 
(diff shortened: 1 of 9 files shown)


edb/schema/annos.py
```diff
@@ -70,6 +70,15 @@ class AnnotationValue(
     value = so.SchemaField(
         str, compcoef=0.909)
 
+    # Overloaded to mark it as not inheritable, since I found it
```
elprans, line 73: Hm, why is it `inheritable` in the base class in the first place?

edb/schema/reflection/writer.py
```diff
@@ -178,6 +178,17 @@ def _build_object_mutation_shape(
     if lprop_fields is None:
         lprop_fields = {}
 
+    # XXX: This is a hack around the fact that _update_lprops works by
```
elprans, line 181: Ugh

elprans (approved):

edb/schema/annos.py
msullivan: Having looked a bit I can't tell why it exists in the first place (apart from on annotation values).

There is syntax for declaring scalar types as FINAL, but as far as I could tell that doesn't really do anything? You can still subtype a final scalar. And there is syntax for doing SET FINAL as part of an alter on a wide variety of different objects, but mostly nothing good seems to happen if you do that. You can put it on a pointer, and then it will be left out when a subtype is created, and then there will be a crash when trying to execute the DDL? I'm guessing this is a vestigal feature?

The logic for how non-inheritable annotations are left out of children is based on it and lives in `inheriting`, which is I guess one reason why it it needs to exist on `SubclassableObject`. (At least absent some refactoring)

PR #2344, opened 2021-03-09
Fix some collection type deletion bugs

msullivan:
* Delete nested collection types on alias deletion
 * Don't delete nested types twice if they appear twice
 * Check context.is_deleting when checking about is_unused

Work on #2296.

There are some available cleanups that I'm going to put in a
follow-up.

### DIFF
--- a/edb/schema/types.py
+++ b/edb/schema/types.py
@@ -924,7 +924,6 @@ class Collection(Type, s_abc.Collection):
         self,
         schema: s_schema.Schema,
         *,
-        if_exists: bool = False,
         expiring_refs: AbstractSet[so.Object],
         view_name: Optional[s_name.QualName] = None,
     ) -> sd.Command:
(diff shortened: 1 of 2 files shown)


edb/schema/types.py
```diff
@@ -1868,7 +1868,7 @@ def as_colltype_delete_delta(
                 expiring_refs=expiring_refs,
             )
 
-        for el in self.get_subtypes(schema):
+        for el in set(self.get_subtypes(schema)):
```
elprans, line 1871: What's this for?

elprans (approved):

edb/schema/types.py
msullivan: Don't delete nested types twice if they appear twice

I'll add a comment

msullivan: Actually that wasn't really sufficient; I switched to using if_exists

PR #2347, opened 2021-03-10
Refactoring of some collection and function deletion stuff

msullivan:
Have CollectionType and CallableType rely on _canonicalize to populate
deletions, and get rid of as_colltype_delete_delta in favor of an
Optional-returning as_type_delete_if_dead defined on Type.

This lets us drop most of our special cases on Collection and on
compound types.


elprans (approved):
Nice cleanup.

PR #2350, opened 2021-03-10
Fix an issue with using ad-hoc aliases in SDL.

### LINKED ISSUE #2320: Problems using a symbol defined in a WITH block in SDL.
- EdgeDB Version: 1.0.0-beta.1
- OS Version: Linux

Steps to Reproduce:

The following schema fails to create a migration:
```
module default {
    type Foo {
        property val -> str;
        property comp := count((
            WITH x := .val
            SELECT y := Bar FILTER x = y.val
        ))
    }

    type Bar {
        property val -> str;
    }
};
```
The error message is:
```
ERROR: InvalidReferenceError: schema item 'std::x' does not exist
```

It appears that validation during tracing is at fault.

vpetrovykh:
Ad-hoc aliases inside WITH blocks or around SELECT results no longer
cause a tracer problem when appearing in expressions in SDL.

Fixes #1425
Fixes #2320

### DIFF
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -2181,6 +2181,112 @@ class TestGetMigration(tb.BaseSchemaLoadTest):
 
         self._assert_migration_consistency(schema)
 
+    def test_schema_get_migration_44(self):
+        schema = r'''
+        type Foo {
+            property val -> str;
+            property comp := count((
+                # Use an alias in WITH block in a computable
+                WITH x := .val
+                # Use an alias in SELECT in a computable
+                SELECT y := Bar FILTER x = y.val
+            ))
+        }
+
+        type Bar {
+            property val -> str;
+        }
+        '''
+
+        self._assert_migration_consistency(schema)
+
+    def test_schema_get_migration_45(self):
+        # Need to make sure that usage of ad-hoc aliases doesn't mask
+        # the real error.
+        with self.assertRaisesRegex(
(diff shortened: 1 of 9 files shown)


tests/test_schema.py
```diff
@@ -2174,6 +2174,112 @@ def test_schema_get_migration_43(self):
...
+
+        type Bar {
+            property val -> str;
+        }
+        '''
+
+        self._assert_migration_consistency(schema)
+
+    def test_schema_get_migration_45(self):
+        # The error should be about usage of mutations in computable,
```
elprans, line 2197: This comment seems to be outdated.

tests/test_schema.py
```diff
@@ -2174,6 +2174,112 @@ def test_schema_get_migration_43(self):
...
+        # not about aliases.
+        with self.assertRaisesRegex(
+            errors.SchemaDefinitionError,
+            "invalid mutation in computable property 'comp'"
```
elprans, line 2201: Observation: the phrasing of this error implies that there is a "valid mutation" in a computable property, when, in fact, all mutations are banned.

elprans (approved):

tests/test_schema.py
vpetrovykh: Reworded.

tests/test_schema.py
vpetrovykh: Reworded.

Issue #2353, opened 2021-03-11
Error creating migration: <QualName __::...> is not a computable pointer


Fogapod:
- EdgeDB Version: `1.0-beta.2+dev.5519.ga407daf8b`

I only changed type name:
```diff
 module default {
-    type UserGuildSettings {
+    type AccentSettings {
         required property user_id -> snowflake;
         required property guild_id -> snowflake;
         property accents -> array<tuple<name: str, severity: int16>> {
             default := <array<tuple<name: str, severity: int16>>>[]
         };
 
         # https://github.com/edgedb/edgedb/issues/1939#issuecomment-789197413
         # constraint exclusive on ((.user_id, .guild_id));
         property exclusive_hack {
             using ((.user_id, .guild_id));
             constraint exclusive;
         };
     }
 
     scalar type snowflake extending int64;
 };
```

Traceback:
```python
$ edgedb -I potato_bot create-migration 
edgedb error: ERROR: InternalServerError: <QualName __::exclusive_hack> is not a computable pointer
  Hint: This is most likely a bug in EdgeDB. Please consider opening an issue ticket at https://github.com/edgedb/edgedb/issues/new?template=bug_report.md
  Server traceback:
      Traceback (most recent call last):
        File "/usr/lib/x86_64-linux-gnu/edgedb-server-1-beta2-dev5519/lib/python3.9/site-packages/edb/edgeql/compiler/setgen.py", line 1029, in computable_ptr_set
          comp_info = ctx.source_map[ptrcls]
      KeyError: <Property 3bbf08be-8194-11eb-ac52-33c7be8b6238 at 0x0x7f75459dc880>
      
      During handling of the above exception, another exception occurred:
      
      Traceback (most recent call last):
        File "/usr/lib/x86_64-linux-gnu/edgedb-server-1-beta2-dev5519/lib/python3.9/site-packages/edb/server/compiler_pool/worker.py", line 312, in worker
          res = meth(*args)
        File "/usr/lib/x86_64-linux-gnu/edgedb-server-1-beta2-dev5519/lib/python3.9/site-packages/edb/server/compiler_pool/worker.py", line 198, in compile_in_tx
          units, cstate = COMPILER.compile_in_tx(cstate, *args, **kwargs)
        File "/usr/lib/x86_64-linux-gnu/edgedb-server-1-beta2-dev5519/lib/python3.9/site-packages/edb/server/compiler/compiler.py", line 2004, in compile_in_tx
[...]

PR #2361, opened 2021-03-13
Complete the fix for interdependent object renames

elprans:
The fix in #2359 corrected the tail end of the rename pipeline, but we
need to apply the cross-ref sort to the entry pass that replaces
expression fields with dummies as well.

Issue: #2353.


msullivan (approved):
Oops, was the test flaking?

elprans:
> Oops, was the test flaking?

Yeah: https://github.com/edgedb/edgedb/runs/2102717791?check_suite_focus=true

PR #2362, opened 2021-03-15
Add the ability to run test on remote Postgres

fantix:
Refs a28a7762, added the `--postgres-dsn` option to `edb test` to configure the test suite to run with a remote Postgres cluster

### DIFF
--- a/edb/tools/test/__init__.py
+++ b/edb/tools/test/__init__.py
@@ -88,9 +88,11 @@ __all__ = ('not_implemented', 'xfail', 'skip')
               help='Maintain a running time log file at FILEPATH')
 @click.option('--list', 'list_tests', is_flag=True,
               help='list all the tests and exit')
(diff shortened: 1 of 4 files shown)


edb/tools/test/__init__.py
```diff
@@ -88,9 +88,11 @@
               help='Maintain a running time log file at FILEPATH')
 @click.option('--list', 'list_tests', is_flag=True,
               help='list all the tests and exit')
+@click.option('--postgres-dsn', type=str,
+              help='Test ')
```
elprans, line 92:
```suggestion
              help='Use the specified Postgres cluster instead of starting a temporary local one.')
```

elprans (approved):

edb/tools/test/__init__.py
fantix: Oops, shouldn't have multi-tasked. Thanks for the fix!

Issue #2349, opened 2021-03-10
Error upgrading from 1-beta1 to nightly: InternalServerError: wrong element type


Fogapod:
- EdgeDB Version: `1-beta1` -> `1.0-beta.1+g2eee1afe4.d20210301`

I tried upgrading my docker installation to nightly and got an error:
```python
edgedb error: failed to restore "potato_bot": restoring database "edgedb": ERROR: InternalServerError: wrong element type
  Hint: This is most likely a bug in EdgeDB. Please consider opening an issue ticket at https://github.com/edgedb/edgedb/issues
/new?template=bug_report.md
  Server traceback:
      edb.errors.InternalServerError: wrong element type
```

Now I can't start my database:
```bash
$ edgedb server status
┌────────────┬───────┬─────────┬─────────────┐
│    Name    │ Port  │ Version │   Status    │
├────────────┼───────┼─────────┼─────────────┤
│ potato_bot │ 10701 │ 1-beta1 │ not running │
└────────────┴───────┴─────────┴─────────────┘
$ edgedb server start potato_bot
edgedb_potato_bot
$ edgedb server logs potato_bot
INFO 1 2021-03-07 20:44:22,264 edb.server: EdgeDB server (1.0-beta.1+g2eee1afe4.d20210301) starting.
INFO 1 2021-03-07 20:44:26,389 edb.server: Serving on 0.0.0.0:10701
INFO 1 2021-03-07 20:44:26,389 edb.server: Serving on /var/lib/edgedb/data/run/.s.EDGEDB.10701
INFO 1 2021-03-07 20:44:26,389 edb.server: Serving admin on /var/lib/edgedb/data/run/.s.EDGEDB.admin.10701
INFO 1 2021-03-10 11:05:51,735 edb.server: Shutting down.
INFO 1 2021-03-10 11:10:49,731 edb.server: EdgeDB server (1.0-beta.1+g2eee1afe4.d20210301) starting.
---- Exception occurred: database instance incompatible with this version of EdgeDB ----

1. edb.errors.ConfigurationError: database instance incompatible with this version of EdgeDB

### a lot of tracebacks ###

---- Details ----
Details: The database instance was initialized with EdgeDB format version 202103050000, but this version of the server expects
format version 202102190000
Hint: You need to recreate the instance and upgrade using dump/restore.
```

`edgedb.upgrade.potato_bot.dump/` was created, I assume this could be useful.

[...]

PR #2364, opened 2021-03-15
Fix handling of collection types of non-builtin scalars in dumps

elprans:
PostgreSQL encodes internal OID of element types in array and tuple
binary data.  Since only the builtin pg_catalog types have stable OIDs,
the data of arrays and tuples of all other types must be patched with
correct OID values.  We already do this in a limited form for input data
of array arguments, but the dump angle was completely overlooked.

Fixes: #2349.

### DIFF
--- a/edb/pgsql/metaschema.py
+++ b/edb/pgsql/metaschema.py
@@ -2834,6 +2834,72 @@ class GetBaseScalarTypeMap(dbops.Function):
         )
 
 
+class GetPgTypeForEdgeDBTypeFunction(dbops.Function):
+    """Return Postgres OID representing a given EdgeDB type."""
+
+    text = f'''
+        SELECT
+            coalesce(
+                (
+                    SELECT
+                        tn::regtype::oid
+                    FROM
+                        edgedb._get_base_scalar_type_map()
+                            AS m(tid uuid, tn text)
+                    WHERE
+                        m.tid = "typeid"
+                ),
+                (
+                    SELECT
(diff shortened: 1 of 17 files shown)


edb/pgsql/metaschema.py
```diff
@@ -2834,6 +2834,72 @@ def __init__(self) -> None:
...
+                            WHERE
+                                tid = elemid
+                        )
+                ),
+                edgedb.raise(
+                    NULL::bigint,
+                    'invalid_parameter_value',
+                    msg => (
+                        'cannot determine OID of EdgeDB type '
+                        || typeid::text
```
1st1, line 2884: Can we also pass the typename to this function to make this error more clear?

1st1 (approved):

edb/server/defines.py
```diff
@@ -34,7 +34,7 @@
 EDGEDB_SPECIAL_DBS = {EDGEDB_TEMPLATE_DB, EDGEDB_SYSTEM_DB}
 
 # Increment this whenever the database layout or stdlib changes.
-EDGEDB_CATALOG_VERSION = 2021_03_05_00_00
+EDGEDB_CATALOG_VERSION = 2021_03_15_00_00
```
tailhook, line 37: This probably means change can't be backported, right? :(

elprans: Yes, there's no way around this I'm afraid.

PR #2365, opened 2021-03-15
Fix failing ROLE DDL test on RDS

fantix:
`CREATE SUPERUSER ROLE` will create a non-super role on EdgeDB with backend Postgres that doesn't have super user like AWS RDS.

### DIFF
--- a/tests/test_edgeql_ddl.py
+++ b/tests/test_edgeql_ddl.py
@@ -6354,12 +6354,14 @@ type test::Foo {
         )
 
     async def test_edgeql_ddl_role_02(self):
+        from edb.server.pgcluster import BackendCapabilities
         await self.con.execute(r"""
             CREATE SUPERUSER ROLE foo2 {
                 SET password := 'secret';
             };
         """)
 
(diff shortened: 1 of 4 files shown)


tests/test_edgeql_ddl.py
```diff
@@ -6369,7 +6371,7 @@ async def test_edgeql_ddl_role_02(self):
             """,
             [{
                 'name': 'foo2',
-                'superuser': True,
+                'superuser': caps & BackendCapabilities.SUPERUSER_ACCESS,
```
elprans, line 6374: No, `superuser` does not actually reflect `rolsuper` directly from Postgres, but rather membership in the EDGEDB_SUPERGROUP role.  However, it appears that we're forgetting to make new roles members of it in `CreateRole`, which is why you get this failure.  IOW, fix `CreateRole.apply` in `pgsql/delta.py` to inherit from `EDGEDB_SUPERGROUP` (unconditionally) instead of from `instance_params.base_superuser` directly.

PR #2367, opened 2021-03-15
Filter out reverse links by type intersection earlier, to avoid false deps

### LINKED ISSUE #2321: Backward link expressions in computables result in overly broad cross-reference record, blocking deletions
Here's a test:

```
    def test_schema_migrations_equivalence_49(self):
        self._assert_migration_equivalence([r"""
            type Foo {
                link bars := .<foo[IS Bar];
                link spam := .<foo[IS Spam];
            };

            type Bar {
                link foo -> Foo;
            };

            type Spam {
                link foo -> Foo;
            };
        """, r"""
            type Foo {
                link spam := .<foo[IS Spam];
            };

            type Spam {
                link foo -> Foo;
            };
        """])

```

It crashes with:

```
edb.errors.SchemaError: cannot drop link 'foo' of object type 'default::Bar' because other objects in the schema depend on it
Detail: link 'spam' of object type 'default::Foo' depends on foo
```

And this happens because the schema reference tracking code, upon seeing `.<foo` in the definition of the `Foo.spam` computable, records **all** `foo` links pointing to `Foo` in the schema, despite an explicit type intersection narrowing that follows.  The offending bits are in `setgen.resolve_ptr()`.

msullivan:
Fixes #2321.

### DIFF
--- a/edb/edgeql/compiler/setgen.py
+++ b/edb/edgeql/compiler/setgen.py
@@ -629,28 +629,29 @@ def resolve_ptr(
                 ctx.env.add_schema_ref(ref, track_ref)
 
     else:
-        all_ptrs = near_endpoint.getrptrs(ctx.env.schema, pointer_name,
-                                          sources=far_endpoints)
-        # If this reverse pointer access is followed by intersections,
-        # we filter out any pointers that couldn't be picked up
-        # by the intersections. We don't do this to generate correct code,
-        # but instead just to avoid creating spurious dependencies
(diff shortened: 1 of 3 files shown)


elprans:
This appears to be interacting badly with common prefix factoring.  With this patch
`SELECT (Foo.<foo[IS Bar], Foo.<foo[IS Spam]);` starts returning a non-empty tuple.

elprans (approved):

Issue #2291, opened 2021-03-01
Failing tests on RDS


fantix:
- EdgeDB Version: 157e58b78315339ff620e04d7afcecb05f0505df
- OS Version: macOS 11.2.1 Darwin Kernel Version 20.3.0 xnu-7195.81.3~1/RELEASE_X86_64
- AWS RDS Version: 12.4

This is an issue to track all the failing tests when running on AWS RDS.

- [x] `ALTER ROLE ... SET password := ...` fails. Fixed in #2282 
- [x] test_edgeql_ddl_role_02 (test_edgeql_ddl.TestEdgeQLDDL)

   ```
   Traceback (most recent call last):
     File "edb/testbase/server.py", line 127, in wrapper
       self.loop.run_until_complete(
     File "asyncio/base_events.py", line 642, in run_until_complete
       return future.result()
     File "tests/test_edgeql_ddl.py", line 6284, in test_edgeql_ddl_role_02
       await self.assert_query_result(
     File "edb/testbase/server.py", line 534, in assert_query_result
       self._assert_data_shape(res, exp_result_json, message=msg)
     File "edb/testbase/server.py", line 742, in _assert_data_shape
       return _assert_generic_shape((), data, shape)
     File "edb/testbase/server.py", line 710, in _assert_generic_shape
       return _assert_list_shape(path, data, shape)
     File "edb/testbase/server.py", line 660, in _assert_list_shape
       _assert_generic_shape(
     File "edb/testbase/server.py", line 714, in _assert_generic_shape
       return _assert_dict_shape(path, data, shape)
     File "edb/testbase/server.py", line 619, in _assert_dict_shape
       _assert_generic_shape(path + (f'["{sk}"]',), data[sk], sv)
     File "edb/testbase/server.py", line 730, in _assert_generic_shape
       self.fail(
   AssertionError: data shape differs: False != True PATH: [0]["superuser"]
   ```

- [x] test_server_proto_ddlprop_01 (test_server_proto.TestServerProtoDdlPropagation)

   ```
   Traceback (most recent call last):
     File "edb/testbase/server.py", line 127, in wrapper
       self.loop.run_until_complete(
     File "asyncio/base_events.py", line 642, in run_until_complete
       return future.result()
     File "tests/test_server_proto.py", line 1895, in test_server_proto_ddlprop_01
       async with tb.start_edgedb_server(adjacent_to=self.con) as sd:
     File "edb/testbase/server.py", line 1495, in __aenter__
[...]

PR #2369, opened 2021-03-16
Auto-detect and set default --max-backend-connections

fantix:
* For local Postgres, default is computed based on total memory
* Local Postgres is now started with `pg_settings.max_connections` being the same as `--max-backend-connections`, meaning all Postgres connections may be used by EdgeDB.
* For remote Postgres, default is computed based on remote `pg_settings.max_connections` minus `pg_settings.superuser_reserved_connections` and other possible reserved
  connections like the one on RDS.
* If specified, the value of this option will be validated against above default/max value for remote Postgres.

Refs [#2291](https://github.com/edgedb/edgedb/issues/2291#issuecomment-788109778)

- [x] Tests to be added

### DIFF
--- a/edb/lib/_testmode.edgeql
+++ b/edb/lib/_testmode.edgeql
@@ -97,6 +97,12 @@ ALTER TYPE cfg::AbstractConfig {
         CREATE ANNOTATION cfg::internal := 'true';
         SET default := '';
     };
+
+    CREATE PROPERTY __pg_max_connections -> std::str {
+        CREATE ANNOTATION cfg::internal := 'true';
+        CREATE ANNOTATION cfg::backend_setting := '"max_connections"';
+        SET default := '';
+    };
 };
 
 
--- a/edb/server/main.py
+++ b/edb/server/main.py
@@ -36,6 +36,7 @@ import tempfile
 import typing
 import warnings
 
+import psutil
 import uvloop
 
 import click
@@ -55,6 +56,8 @@ from . import pgconnparams
 from . import pgcluster
 from . import protocol
 
+
+BYTES_OF_MEM_PER_CONN = 100 * 1024 * 1024 / 8  # 100Mb
 
 logger = logging.getLogger('edb.server')
 _server_initialized = False
@@ -271,10 +274,8 @@ def run_server(args: ServerConfig):
     cluster: Union[pgcluster.Cluster, pgcluster.RemoteCluster]
     if args.data_dir:
         if not args.max_backend_connections:
-            max_conns = _default_max_backend_connections()
-            new_args = args._asdict()
-            new_args['max_backend_connections'] = max_conns
-            args = ServerConfig(**new_args)
+            max_conns = _compute_default_max_backend_connections()
+            args = args._replace(max_backend_connections=max_conns)
             logger.info(f'Using {max_conns} max backend connections based on '
                         f'total memory.')
 
@@ -295,9 +296,7 @@ def run_server(args: ServerConfig):
             instance_params.max_connections -
             instance_params.reserved_connections)
         if not args.max_backend_connections:
-            new_args = args._asdict()
-            new_args['max_backend_connections'] = max_conns
-            args = ServerConfig(**new_args)
+            args = args._replace(max_backend_connections=max_conns)
             logger.info(f'Detected {max_conns} backend connections available.')
         elif args.max_backend_connections > max_conns:
             abort(f'--max-backend-connections is too large for this backend; '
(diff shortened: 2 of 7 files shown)


edb/server/main.py
```diff
@@ -487,13 +510,30 @@ def _protocol_version(
...
 def _validate_max_backend_connections(ctx, param, value):
-    if value < defines.BACKEND_CONNECTIONS_MIN:
+    if value is not None and value < defines.BACKEND_CONNECTIONS_MIN:
         raise click.BadParameter(
             f'the minimum number of backend connections '
             f'is {defines.BACKEND_CONNECTIONS_MIN}')
     return value
 
 
+def _default_max_backend_connections():
```
1st1, line 520:
```suggestion
def _compute_default_max_backend_connections():
```

edb/server/main.py
```diff
@@ -487,13 +510,30 @@ def _protocol_version(
...
+    try:
+        import psutil
```
1st1, line 522: We don't have `psutil` as a dependency in our `setup.py`, and since we ship our own Python in our packages it's a 0 chance that this import will ever succeed on a user installation.

I think we can add psutil as a dependency, this isn't the first time we need it. @elprans ok?

fantix: Added now 🖖

edb/server/main.py
```diff
@@ -270,7 +273,16 @@ def run_server(args: ServerConfig):
 
     cluster: Union[pgcluster.Cluster, pgcluster.RemoteCluster]
     if args.data_dir:
-        cluster = pgcluster.get_local_pg_cluster(args.data_dir)
+        if not args.max_backend_connections:
+            max_conns = _compute_default_max_backend_connections()
+            new_args = args._asdict()
+            new_args['max_backend_connections'] = max_conns
+            args = ServerConfig(**new_args)
```
fantix, line 280: use `NamedTuple._replace` here

edb/lib/_testmode.edgeql
```diff
@@ -97,6 +97,12 @@ ALTER TYPE cfg::AbstractConfig {
         CREATE ANNOTATION cfg::internal := 'true';
         SET default := '';
     };
+
+    CREATE PROPERTY __pg_max_connections -> std::str {
```
1st1, line 101: Hm, is this supposed to be accessed only by tests?

edb/server/main.py
```diff
@@ -56,6 +57,8 @@
 from . import protocol
 
 
+BYTES_OF_MEM_PER_CONN = 100 * 1024 * 1024 / 8  # 100Mb
```
elprans, line 60: Why are you dividing by 8 here?  

Also, 100MB per backend seems a bit low, but, on the other hand, actual worker memory utilization depends greatly on the complexity and number of queries in prepared statements, so we will need to build tools to query or estimate actual memory utilization and have the pool retire connections that have crossed a certain memory utilization threshold.

edb/server/main.py
```diff
@@ -487,13 +509,18 @@ def _protocol_version(
...
+    if value is not None and value < defines.BACKEND_CONNECTIONS_MIN:
         raise click.BadParameter(
             f'the minimum number of backend connections '
             f'is {defines.BACKEND_CONNECTIONS_MIN}')
     return value
 
 
+def _compute_default_max_backend_connections():
+    total_mem = psutil.virtual_memory().total
+    return int(total_mem / BYTES_OF_MEM_PER_CONN)
```
1st1, line 521:
```suggestion
    return max(int(total_mem / BYTES_OF_MEM_PER_CONN), 1)
```

or maybe the lower bound should be 2, idk.

tests/test_server_ops.py
```diff
@@ -130,3 +125,82 @@ async def test_server_ops_bootstrap_script(self):
...
+                max_allowed_connections=None,
+                postgres_dsn=f'postgres:///?user=postgres&port={port}&'
+                             f'host={host}',
+            ) as sd:
+                con = await edgedb.async_connect(
+                    user='edgedb', host=sd.host, port=sd.port)
+                try:
+                    max_connections = await con.query_one(
+                        'SELECT cfg::SystemConfig.__pg_max_connections '
+                        'LIMIT 1')
```
1st1, line 187: Hm, does it work without `LIMIT 1`? If it doesn't it's a bug

elprans: It's not a bug.  We don't have a notion of a singleton object, so the compiler is unable to prove that there is only one `cfg::SystemConfig`.

elprans: Also, this reminded me that we need to implement cardinality assertions instead of using `LIMIT 1`

edb/lib/_testmode.edgeql
fantix: Yeah I mean that's my intention - managed Postgres should define/access their own `max_connections` from the Postgres side, or if the Postgres is managed by EdgeDB, we could already configure this value through `max_backend_connections` (for now `pg_settings.max_connections` equals to EdgeDB `max_backend_connections`).

edb/server/main.py
fantix: Oh the `8` is for `Mb` bytes to bits, but if 100MBytes is already too low ... I'll definitely remove the `/ 8` then, and perhaps we could start with 100MBytes per backend assumption? That's approximilately 10 backend connections per GB of memory, or 82 for 8GB.

edb/server/main.py
fantix: Good catch! As we do have a "pool", perhaps 2 makes more sense.

edb/server/main.py
elprans: Oh.  Memory isn't usually measured in bits, so this got me confused.  Yes, we want at least 100MiB, but, like I said above, ideally this heuristic should be dynamic.

setup.py
```diff
@@ -49,6 +49,7 @@
     'immutables>=0.15',
     'parsing~=1.6.1',
     'prompt_toolkit==3.0.3',
+    'psutil>=5.8.0',
```
elprans, line 52: We prefer to pin with `~=` instead.

elprans (approved):

PR #2370, opened 2021-03-16
Add backend pool logging

fantix:
In order to lower the impact, backend connection events are logged only once at most every second per block.

### DIFF
--- a/edb/server/connpool/pool.py
+++ b/edb/server/connpool/pool.py
@@ -239,20 +239,25 @@ class Block(typing.Generic[C]):
     def log_connection(self, event: str, timestamp: float = 0) -> None:
         if not timestamp:
             timestamp = time.monotonic()
+
+        # Add to the backlog if we're in batching, regardless of the time
         if self._is_log_batching:
             self._log_events[event] = self._log_events.setdefault(event, 0) + 1
+
+        # Time check only if we're not in batching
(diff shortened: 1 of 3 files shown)


edb/server/connpool/pool.py
```diff
@@ -224,6 +236,37 @@ def _wakeup_next_waiter(self) -> None:
                 waiter.set_result(None)
                 break
 
+    def log_connection(self, event: str, timestamp: float = 0) -> None:
+        if not timestamp:
+            timestamp = time.monotonic()
+        if self._is_log_batching:
+            self._log_events[event] = self._log_events.setdefault(event, 0) + 1
+        elif timestamp - self._last_log_timestamp > MIN_LOG_TIME_THRESHOLD:
```
1st1, line 244: There can be a race here, as asyncio scheduled calls can be executed slightly past their time

```suggestion
        elif (
            timestamp - self._last_log_timestamp > MIN_LOG_TIME_THRESHOLD
            and not self._is_log_batching
        ):
```

1st1 (approved):

edb/server/connpool/pool.py
fantix: Right, thanks for the tip! The `elif` should've taken care of that. Let me add a comment

1st1: Of course, you're right! 🤦‍♂️

PR #2371, opened 2021-03-16
Add TCP keepalive to backend connections

fantix:
The current backend pooling algorithm seems to be not reusing existing connections before the `max_backend_connections` is reached. Before that is fixed, we need TCP keepalive to make test suite work in environment like GitHub Action + AWS RDS where TCP connections are likely proxied and dropped silently without `RST` after being idle for a while.

### DIFF
--- a/edb/server/pgcon/pgcon.pyx
+++ b/edb/server/pgcon/pgcon.pyx
@@ -23,6 +23,7 @@ import codecs
 import hashlib
 import json
 import os.path
+import socket
 
 cimport cython
 cimport cpython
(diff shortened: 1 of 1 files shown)


1st1 (approved):

edb/server/pgcon/pgcon.pyx
```diff
@@ -124,8 +128,17 @@ async def connect(connargs, dbname):
             lambda: PGConnection(dbname, loop, connargs), addr)
 
     else:
-        _, pgcon = await loop.create_connection(
+        trans, pgcon = await loop.create_connection(
             lambda: PGConnection(dbname, loop, connargs), host=host, port=port)
+        sock = trans.get_extra_info('socket')
+        sock.setsockopt(socket.SOL_SOCKET, socket.SO_KEEPALIVE, 1)
```
1st1, line 134: Can you also add a comment explaining why we bother with TCP keepalive here?

fantix: Sure thing!

PR #2376, opened 2021-03-17
Track and print scopetree children in order

msullivan:
This produces what I feel is more useful debug output, especially when
dumping the tree repeatedly as it evolves (since it is more stable).

I fixed test_edgeql_ir_scopetree automatically using this script:
https://gist.github.com/msullivan/0d3e3d52b9cca471ddcc29b31cc61def

### DIFF
--- a/edb/ir/scopetree.py
+++ b/edb/ir/scopetree.py
@@ -72,7 +72,7 @@ class ScopeTreeNode:
     nodes merged into it, but could become optional again.
     """
 
-    children: MutableSet[ScopeTreeNode]
+    children: List[ScopeTreeNode]
     """A set of child nodes."""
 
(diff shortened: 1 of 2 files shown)


elprans:
Can we measure a perf impact of this?  `OrderedSet` is a thin wrapper over dict, but it's a wrapper nonetheless.

msullivan:
It looks like it slowed things down by a percent or two (though I'm not totally sure I believe that). I tried replacing it with a list, which sped things up by a percent or two. 

Switching to a list could be bad for remove in big cases, but in practice should usually be faster. (And we already have way worse quadratic places than this new one.)

elprans (approved):

PR #2384, opened 2021-03-18
Add a generic min/max aggregate fallback

### LINKED ISSUE #1920: min/max trigger ISE when used on unsupported types
`std::min`/`std::max` nominally take `anytype` but the backing SQL function doesn't support `uuids` or tuples, among other things;

`SELECT min(Object.id);`, `SELECT min(Object)` and `SELECT min({(1,2), (3,4)});` all produce ISEs.

We should at least properly disallow the unsupported ones but since they have working less than operators it would be best if we could support it actually. (But I'm not sure if that is plausible.)

(I think there might be other similar things, but I couldn't find more at a quick glance.)

vpetrovykh:
The aggregates min and max cannot handle all of our scalar types due to
how they're implemented in Postgres. So we will use specialized min/max
aggregates for the types Postgres explicitly supports and rely on a generic
catch-all implmentation for all the rest.

We introduce a `fallback` flag for function declarations. The purpose of
the flag is to designate a particular implementation as the fallback for
a polymorphic function. There's only one fallback that can be
designated. This fallback function is exempt from the restriction of
polymorphic functions using the same underlying SQL function (so that
Postgres can actually resolve the polymorphic calls). It means that this
fallback implementation will in fact be used in all cases where we
statically cannot resolve the call to a more specific version of the
polymorphic function (such as in bodies of other polymorphic functions).

Currently the `fallback` flag is not intended to be set by the users of
EdgeDB, only by the system initialization scripts that define the
standard library.

Fixes #1920

Co-authored-by: Elvis Pranskevichus <elvis@edgedb.com>


docs/edgeql/funcops/set.rst
```diff
@@ -394,12 +394,19 @@ Set
...
+.. eql:function:: std::min(values: SET OF anyscalar) -> \
+                  OPTIONAL anyscalar
+                  std::min(values: SET OF array<anyscalar>) -> \
+                  OPTIONAL array<anyscalar>
 
     :index: aggregate
 
     Return the smallest value of the input set.
 
+    Currently, the ``min`` function does not accept :eql:type:`bool`,
```
elprans, line 406: IMO, it's better to list what the function _does_ accept, because we are operating from a specific list of definitions in Postgres.

vpetrovykh: In principle I can just put the entire set of polymorphic versions there, but the full list would be bigger than `to_str` [here](https://www.edgedb.com/docs/edgeql/funcops/string#function::std::to_str) (because of the flavors of the arrays).

The reason why I worded it this way is to convey the intent that we actually want to properly extend it to all orderables (at least all orderable scalars and types derived from them).

elprans (changes requested):
As discussed, we will try implementing a generic fallback via a `SELECT arg ORDER BY arg LIMIT 1` specialization as an `SQL EXPRESSION` function.

vpetrovykh:
I've reworked the approach. Now we have a fallback and the fix is backwards-compatible.

tests/test_edgeql_expressions.py
```diff
@@ -2170,6 +2170,26 @@ async def test_edgeql_expr_valid_collection_16(self):
...
+                query = f"""SELECT {fn}([{val}]) = [{val}];"""
+                await self.assert_query_result(query, {True})
+
+    async def test_edgeql_expr_valid_minmax_03(self):
+        for val in get_test_values():
+            for fn in ['min', 'max']:
+                # same as the previous test, but for tuples
+                query = f"""SELECT {fn}(({val},)) = ({val},);"""
+                await self.assert_query_result(query, {True})
+
```
msullivan, line 2192: What happens if we take `min` or `max` of an object type, especially one with nontrivial shape components?

I think it's worth adding a test for that, but if it doesn't work it's probably OK to disallow it for now. (Or for ever? It isn't particularly semantically meaningful.)

msullivan (approved):
Looks good to me, pending my question about object types.

tests/test_edgeql_expressions.py
elprans: Semantically the "value" of an object is its ID.  Shape components aren't considered to be part of that value (unless cast to `json`).  Even though there's no intrinsic external meaning to sorting objects by value, I don't think we should ban this case.  `ORDER BY` guarantees certain stable order of elements in set, which by itself might be useful in some cases and which is why we consider all types to be orderable.

edb/schema/functions.py
```diff
@@ -717,6 +717,14 @@ class CallableObject(
     abstract = so.SchemaField(
         bool, default=False, inheritable=False, compcoef=0.909)
 
+    fallback = so.SchemaField(
```
elprans, line 720: Please add a comment describing this field.  Also, don't forget to bump CATALOG_VERSION since this adds a new property to the reflection schema.

edb/lib/std/20-genericfuncs.edgeql
```diff
@@ -162,6 +162,57 @@ std::random() -> std::float64
 
 CREATE FUNCTION
 std::min(vals: SET OF anytype) -> OPTIONAL anytype
+{
```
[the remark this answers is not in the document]
vpetrovykh: Comment added.

edb/pgsql/compiler/relgen.py
```diff
@@ -1908,6 +1917,47 @@ def process_set_as_enumerate(
     return new_stmt_set_rvar(ir_set, stmt, aspects=aspects, ctx=ctx)
 
 
+def process_set_as_std_min_max(
+    ir_set: irast.Set,
+    stmt: pgast.SelectStmt,
+    *,
+    ctx: context.CompilerContextLevel,
+) -> SetRVars:
```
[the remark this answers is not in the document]
vpetrovykh: Comment added.

tests/test_edgeql_expressions.py
vpetrovykh: I've added tests for min/max on objects. They're ordered by their `id`, so min/max should respect that as well.

edb/schema/functions.py
vpetrovykh: Comment added and catalog version bumped.

edb/schema/functions.py
```diff
@@ -1159,6 +1159,23 @@ class Function(
...
+    # function. The fallback implementation is exempted from the
+    # limitation that all polymorphic functions have to map to the
+    # same function in Postgres. There can only be at most one
+    # fallback implementation for any given polymorphic function.
+    #
+    # The flag is intended for internal use for standard library
+    # functions.
+    fallback = so.SchemaField(
+        bool,
+        allow_ddl_set=True,
```
elprans, line 1173: Drop `allow_ddl_set` here as this isn't supposed to be settable from user DDL.

elprans (approved):

Issue #2380, opened 2021-03-18
create-migration fails when a property was moved to an abstract base type


ambv:
- EdgeDB Version: 1.0b1
- OS Version: macOS 10.15.7

Schema before:
```esdl
module commonplace {
    type User {
        required property name -> util::Slug {
            constraint exclusive;
        };
        index on (__subject__.name);
    };
};
```

Schema after:
```esdl
module commonplace {
    type User extending
        Named;

    type Content extending
        Named;

    abstract type Named {
        required property name -> util::Slug {
            constraint exclusive;
        };
        index on (__subject__.name);
    };
};
```

create-migration failing:
![2021-03-18 15 29 59](https://user-images.githubusercontent.com/55281/111643823-8ef89c80-87ff-11eb-80ee-7987f7544ff6.gif)

ambv:
Trying to drop the index first results in a different error, about the exclusivity constraint now:
```
did you drop constraint 'std::exclusive' of property 'name'? [y,n,l,c,b,s,q,?]
y
error: cannot drop inherited constraint 'std::exclusive' of property 'name' of object type 'commonplace::User'
  ┌─ query:3:9
  │
3 │         DROP CONSTRAINT std::exclusive;
  │         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ error
  │
  = constraint 'std::exclusive' of property 'name' of object type 'commonplace::User' is inherited from:
    - property 'name' of object type 'commonplace::Named'
```

Answering "no" to the question results in:
```
ERROR: InternalServerError: uncompiled expression in the field 'expr' of constraint <QualName commonplace::std|exclusive@commonplace|__||name&commonplace||User@da39a3ee5e6b4b0d3255bfef95601890afd80709>
```

elprans:
> create-migration failing:

Confirmed.

> Trying to drop the index first 

Hm, the error seems to suggest `User` is already inherited from `Named`.  You cannot drop constraints inherited from parent type.

> Answering "no" to the question results in:

This was fixed in b2.nightly.

msullivan:
@ambv what tool do you use to make the animations?

ambv:
> @ambv what tool do you use to make the animations?

Gifox 2 on macOS

PR #2386, opened 2021-03-19
Add a test for __typenames__ flag.

vpetrovykh:
Running a query with `__typenames__` flag causes an error for a built-in
type.

### DIFF
--- a/tests/test_edgeql_select.py
+++ b/tests/test_edgeql_select.py
@@ -6304,3 +6304,15 @@ class TestEdgeQLSelect(tb.QueryTestCase):
         for row in res:
             self.assertEqual(row[0].__tname__, "test::User")
             self.assertEqual(row[1].__tname__, "test::Issue")
+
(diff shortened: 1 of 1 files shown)


msullivan (approved):

tests/test_edgeql_select.py
```diff
@@ -6304,3 +6304,14 @@ async def test_edgeql_select_array_common_type_01(self):
         for row in res:
             self.assertEqual(row[0].__tname__, "test::User")
             self.assertEqual(row[1].__tname__, "test::Issue")
+
+    @test.xfail("""
+        SchemaDefinitionError: cannot alter object type 'std::Object':
+        module std is read-only
+    """)
```
msullivan, line 6311: Mention #2380?

vpetrovykh: Do you mean #2387 (which I added a reference to)?

I don't see how #2380 is related

PR #2390, opened 2021-03-19
Make schema::TupleElement extend std::BaseObject

msullivan:
Otherwise `SELECT Object` returns all the schema::TupleElements,
which is pretty silly.


elprans (approved):
Perhaps it's worth adding a check that types in `schema::` don't inherit from `Object`.

PR #2394, opened 2021-03-22
Fix a staeval cast bug that produces bizarre results

msullivan:
const_to_python casting would directly apply the cast to the "python
value", which is wrong because tuples and None have special
meaning. Since operator eval uses a code path that uses
const_to_python on casts, this would result in concatenation of a cast
of a constant set to produce bizarre results.

### DIFF
--- a/tests/test_edgeql_expressions.py
+++ b/tests/test_edgeql_expressions.py
@@ -4704,3 +4704,30 @@ aa \
             await self.con.query("""
                 SELECT Object ?? '';
             """)
+
+    async def test_edgeql_expr_static_eval_casts_01(self):
+        # The static evaluator produced some really silly results for
+        # these at one point.
+
+        await self.assert_query_result(
+            r'''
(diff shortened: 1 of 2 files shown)


tests/test_edgeql_expressions.py
```diff
@@ -4704,3 +4704,30 @@ async def test_edgeql_expr_invalid_object_scalar_op_01(self):
...
                 SELECT Object ?? '';
             """)
+
+    async def test_edgeql_expr_static_eval_casts_01(self):
+        # The static evaluator produced some really silly results for
+        # these at one point.
+
+        await self.assert_query_result(
+            r'''
+                WITH x := {1, 2}, SELECT ("wtf" ++ <str>x);
```
elprans, line 4714:
```suggestion
                WITH x := {1, 2}, SELECT ("ftw" ++ <str>x);
```

elprans (approved):

PR #2397, opened 2021-03-24
Run test suite daily on AWS, DO

fantix:
Also made the GitHub workflow YAMLs for the test suite Jinja2 templates because of duplicate code.

This depends a bunch of other PRs to run correctly, so it shall be merged as the last one. Demo available [here](https://github.com/edgedb/edgedb/actions/workflows/tests-managed-pg.yml)


.github/workflows.src/tests-managed-pg.tpl.yml
```diff
@@ -0,0 +1,218 @@
...
+      with:
+        aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+        aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+        aws-region: us-east-2
+<%- endmacro -%>
+
+name: Tests on Managed PostgreSQL
+
+on:
+  schedule:
```
elprans, line 15: Let's add a `workflow_dispatch` trigger too for the ability to do ad hoc runs.

elprans (approved):
Great work, @fantix!

PR #2398, opened 2021-03-24
Support SSL on backend connections

fantix:
This PR is mostly copied from asyncpg SSL support, aiming to be compatible with libpq sslmode behavior. Likely, the default sslmode is "prefer" - use SSL connection (without certificate verification) if the server supports SSL, or fallback to non-SSL otherwise.

### DIFF
--- a/edb/server/pgcon/pgcon.pyx
+++ b/edb/server/pgcon/pgcon.pyx
@@ -60,7 +60,7 @@ from edb.server import buildmeta
 from edb.server import compiler
 from edb.server import defines
 from edb.server.cache cimport stmt_cache
-from edb.server.pgconnparams import SSLMode
+from edb.server import pgconnparams
 from edb.server.protocol cimport binary as edgecon
 
 from edb.common import debug
@@ -169,7 +169,7 @@ async def _create_ssl_connection(protocol_factory, host, port, *,
     return new_tr, pg_proto
 
 
-class _Retry(Exception):
+class _RetryConnectSignal(Exception):
     pass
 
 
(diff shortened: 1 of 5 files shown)


elprans (approved):

elprans:
We'll probably also need to add support for sslcert, sslkey, sslrootcert, and sslcrl DSN args too.

edb/server/pgcon/pgcon.pyx
```diff
@@ -58,13 +60,13 @@ from edb.server import buildmeta
 from edb.server import compiler
 from edb.server import defines
 from edb.server.cache cimport stmt_cache
+from edb.server.pgconnparams import SSLMode
```
1st1, line 63: Please import the module, not the concrete type

edb/server/pgcon/pgcon.pyx
```diff
@@ -116,39 +118,141 @@ def _build_init_con_script() -> bytes:
...
+
+    pg_proto = protocol_factory()
+    pg_proto.is_ssl = do_ssl_upgrade
+    pg_proto.connection_made(new_tr)
+    new_tr.set_protocol(pg_proto)
+
+    return new_tr, pg_proto
+
+
+class _Retry(Exception):
```
1st1, line 172: I'd rename to `_RetryWithSSLSignal` and add `_RetryWithoutSSLSignal`

1st1 (commented):
Good stuff, but please address the comments before merging

edb/server/pgcon/pgcon.pyx
elprans: Hm, having two exceptions would complicate things.  In this implementation both the [no-ssl, ssl] and [ssl, no-ssl] are served by the same piece of code which doesn't actually care if a particular set of connection args has ssl or not.

1st1: Then `_RetryConnectSignal`

PR #2404, opened 2021-03-24
Fix a cluster of FOR and tuple binding bugs

msullivan:
The prototypical issues here that started this issue off are
test_edgeql_scope_binding_04 (a FOR test) and
test_edgeql_scope_binding_05 (a WITH/tuple test).

In test _04, a WITH bound FOR loop spuriously correlated with itself
when selected twice in subqueries, and in test _05, its translation to
WITH/tuples crashed in pgsql.

The cause of both involved path_id_mask:

 * For _05, the issue is that pull_path_namespace would pull a mapped
   path name even if its unmapped version was in path_id_mask. Since
   tuples do mapping, this would allow one name for a tuple element to
   escape incorrectly.

 * For _04, the issue is that all iterator statements were explicitly
   exempted from being placed into path_id_mask, which pretty
   straightforwardly causes them to spuriously correlate.
   Unfortunately this is required in a number of cases where
   FOR appears inside shape computables, so I tweaked the criteria
   to be closer to being correct.
   One key thing is that we don't want *all* FOR loops inside of
   shape computables to be exempted from path_id_mask
   (see test_edgeql_scope_binding_07)

Also to help write one of my tests, I made tuples equivalent to lists
for test case shapes, which lets us put them in sets.

### DIFF
--- a/edb/pgsql/compiler/dml.py
+++ b/edb/pgsql/compiler/dml.py
@@ -809,6 +809,7 @@ def compile_insert_else_body(
 
         dispatch.compile(else_select, ctx=ictx)
         pathctx.put_path_id_map(ictx.rel, subject_id, else_select.path_id)
+        ictx.rel.path_id_mask.discard(else_select.path_id)
 
         else_select_cte = pgast.CommonTableExpr(
             query=ictx.rel,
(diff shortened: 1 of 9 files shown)


edb/pgsql/compiler/dml.py
```diff
@@ -809,6 +809,7 @@ def compile_insert_else_body(
 
         dispatch.compile(else_select, ctx=ictx)
         pathctx.put_path_id_map(ictx.rel, subject_id, else_select.path_id)
+        ictx.rel.path_id_mask.discard(else_select.path_id)
```
elprans, line 812: Add a comment on why we're exposing `else_select` here?

elprans (approved):

PR #2407, opened 2021-03-26
Fix creation/alteration of scalar constraints used in containers

### LINKED ISSUE #1370: InternalServerError: cannot alter type
- EdgeDB Version: EdgeDB 1.0-alpha.2+g16f75ece5.d20200122
- OS Version: Windos 10 Educational, Version: 1909, Build: 18363.778

Steps to Reproduce:
```
START TRANSACTION;
CREATE MIGRATION m2 TO {
    module default {
        type User {
            required property scopes -> array<scope>;
        }
        scalar type scope extending str {
            constraint one_of (
                '1', '2'
            );
        }
    }
};
COMMIT MIGRATION m2;
COMMIT;
```

Error log:
```
InternalServerError: cannot alter type ""edgedb_e1d3ab02-7c59-11ea-9cc7-f7e49d2fb386"."87c9d94a-87f0-11ea-b4ba-db2bb9dbfd2b_domain"" because column "6716f2a1-87ef-11ea-bb56-fd1198a88276.scopes" uses it
```

May be related to this issue #1369.

msullivan:
Migrations that created a scalar type and a property that used it
inside a container were broken for a simple reason, which is that the
constraint creation would get priority sorted after the column
creation. Tweak things around to avoid that.

However, adding a constraint to an existing scalar type that was used
in a property inside of a container type
(like `property x -> array<positive_int>`) was broken in a more
fundamental way, because pgsql disallows doing that.

We can work around this in a laborious way:
 1. Find all of the properties whose type is a container type
    that contains this scalar. (Possibly transitively.)
 2. Change the type of all offending properties to an equivalent type
    that does not reference this scalar. This may require creating
    new types.
 3. Add the constraint.
 4. Restore the type of all offending properties. If existing data
    violates the new constraint, we will fail here. Delete any
    temporarily created types.

This is a lot more code than the importance of the particular case
really ought to justify, but we live in a fallen world.

The SET TYPEs work by generating automatic conversion code using
casts. All of this infrastructure existed already but getting it
to work for the tests I wrote required fixing a couple cast bugs
and a bug in the compilation of aggregate functions used directly
as a query top-level.

Fixes #1370.

### DIFF
--- a/edb/pgsql/compiler/relgen.py
+++ b/edb/pgsql/compiler/relgen.py
@@ -2523,12 +2523,11 @@ def process_set_as_agg_expr(
     expr = ir_set.expr
     assert isinstance(expr, irast.FunctionCall)
 
-    # If the func has an initial val, we need to create a wrapper to put
-    # the coalesces we insert in
+    # If the func has an initial val, we need to do the interesting
+    # work in subrels and provide a wrapper to put the coalesces in
     wrapper = None
     if expr.func_initial_value is not None:
-        with ctx.subrel() as subctx:
-            wrapper = subctx.rel
+        wrapper = stmt
 
     # When in a serialization context, we need to compute both a value
     # version and a serialized version of the aggregate function call,
@@ -2554,9 +2553,6 @@ def process_set_as_agg_expr(
             process_set_as_agg_expr_inner(
                 ir_set, xctx.rel, aspect='serialized', wrapper=wrapper,
                 ctx=xctx)
-
-    if wrapper:
-        ctx.rel = stmt = wrapper
 
     return new_stmt_set_rvar(ir_set, stmt, ctx=ctx)
 
--- a/edb/pgsql/compiler/relgen.py
+++ b/edb/pgsql/compiler/relgen.py
@@ -2556,7 +2556,7 @@ def process_set_as_agg_expr(
                 ctx=xctx)
 
     if wrapper:
-        stmt = wrapper
+        ctx.rel = stmt = wrapper
 
     return new_stmt_set_rvar(ir_set, stmt, ctx=ctx)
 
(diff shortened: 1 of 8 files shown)


edb/pgsql/compiler/relgen.py
```diff
@@ -2556,7 +2556,7 @@ def process_set_as_agg_expr(
                 ctx=xctx)
 
     if wrapper:
-        stmt = wrapper
+        ctx.rel = stmt = wrapper
```
elprans, line 2559: This seems fishy.  `wrapper` may come from a `ctx.subrel()` so we're doing some weird inversion here.  Why is this needed?

edb/pgsql/delta.py
```diff
@@ -1718,8 +1863,18 @@ def apply(
         orig_schema = schema
         schema = s_scalars.AlterScalarType.apply(self, schema, context)
         new_scalar = self.scls
+
+        used_props = []
+        if list(self.get_subcommands(type=s_constr.CreateConstraint)):
+            used_props = self._get_composite_refs(schema, context)
+            schema = self._undo_everything(schema, context, used_props)
```
elprans, line 1870: Maybe add `if used_props:` so this reads cleaner.

edb/pgsql/compiler/relgen.py
msullivan: When a aggregate function with an initial_val was used at the top-level of a query (without a SelectStmt wrapping it, which happens for fragments like the USING clauses of SET TYPE), the actual results were in one of the rvars in the returned SetRVars but they weren't actually making it into the result rel.

This is because in those cases, we were totally ignoring the input stmt in favor of creating a new statement, which was dubious. This fix was even more dodgy, and I've pushed a cleanup.

(Though it also feels off to me that `_process_toplevel_query` doesn't do anything with the rvars in the common case. But trying to `_include_rvars` there causes us to try to generate an infinite amount of SQL)

elprans (approved):
:+1:

PR #2410, opened 2021-03-29
Consider object exclusive constraints in cardinality inference

### LINKED ISSUE #2097: Possibly inconsistent cardinality determination
With the following schema:

```
type Person {
    constraint exclusive on ((.name,));
}
```

If I run:

```python
query_one_json("""SELECT Person FILTER .name = 'test'""")
```

I get:

```
edgedb.errors.InterfaceError: query cannot be executed with query_one_json() as it returns a multiset
```

But if I do:

```python
query_one_json("""SELECT Person FILTER .id = <uuid>'...'""")
```

I get a singular result, as expected.

I would have thought that since `constraint exclusive on ((.name,))` creates an index, the cardinality would have been the same as filtering on `id`. Please let me know if I am misunderstanding or doing something wrong.

Edit: upon further inspection, this does not seem to be a problem with the Python driver itself (so sorry for posting here), since I also encountered similar behavior when using `UNLESS CONFLICT ON`:

```
INSERT Person { name := 'test' }
UNLESS CONFLICT ON .name
ELSE (...)
```

Fails as well. That lead me to try:

```
type Person {
    required property name -> str { constraint exclusive };
}
```

Which, does resolve both problems, but isn't equivalent for the case of:

```
type Person {
    required property first_name -> str;
    required property last_name -> str;
    constraint exclusive on ((.first_name, .last_name,));
}
```

msullivan:
This makes object exclusive constraints more consistent with pointer
ones and allows easily filtering on compound exclusive constraints.

We do this by looking at tuple values in addition to directly pointer
references.

Fixes #2097. Lays the groundwork for tuple ON CONFLICT.

### DIFF
--- a/edb/edgeql/compiler/inference/cardinality.py
+++ b/edb/edgeql/compiler/inference/cardinality.py
@@ -1041,7 +1041,7 @@ def _analyse_filter_clause(
         obj_exclusives = get_object_exclusive_constraints(
             result_stype, ptr_set, ctx.env)
         for _, exc_ptrs in obj_exclusives:
-            if all(exc_ptr in ptr_set for exc_ptr in exc_ptrs):
+            if not (set(exc_ptrs) - ptr_set):
                 return AT_MOST_ONE
 
     return result_card
--- a/edb/edgeql/compiler/inference/cardinality.py
+++ b/edb/edgeql/compiler/inference/cardinality.py
@@ -36,8 +36,11 @@ from edb.common import parsing
 from edb.edgeql import qltypes
 
 from edb.schema import name as sn
+from edb.schema import types as s_types
 from edb.schema import objtypes as s_objtypes
 from edb.schema import pointers as s_pointers
+from edb.schema import constraints as s_constraints
+from edb.schema import expr as s_expr
 
 from edb.ir import ast as irast
 from edb.ir import utils as irutils
@@ -48,6 +51,7 @@ from . import volatility
 from . import context as inference_context
 
 from .. import context
+from .. import options
 
 
 AT_MOST_ONE = qltypes.Cardinality.AT_MOST_ONE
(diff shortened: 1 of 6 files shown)


edb/edgeql/compiler/inference/cardinality.py
```diff
@@ -932,6 +936,82 @@ def extract_filters(
     return None
 
 
+def _get_object_constraint_pointers(
```
msullivan, line 939: Hm, I did a bunch of work to only look at tuple constraints so that everything is pretty regular and works without any trouble. But now that I think about it, maybe that's all a waste of time and an unnecessary limitation: any constraint that is a stable function of the pointers it references should work. So maybe we should drop all of this logic and `irutils.extract_constraint_style_ptrs` and *just* base the decision on `subjectexpr.refs` (which we already look at as an optimization before looking more deeply at the subjectexpr body).

The advantage here would be that then this would work with constraints on stuff like `.first ++ .last` would work. To the extent that that is really an advantage, I guess.

Thoughts?

elprans: This would work only if the constraint expression result is not _more_ unique than a tuple of its inputs, which would be hard to prove in the general case.  To support inference on arbitrary expressions, we must verify that you've used an equivalent expression in your `FILTER` or `UNLESS CONFLICT`.

edb/edgeql/compiler/inference/cardinality.py
```diff
@@ -945,12 +1025,25 @@ def _analyse_filter_clause(
...
                 # Bingo, got an equality filter on a link with a
                 # unique constraint
                 return AT_MOST_ONE
 
+        # Then look at all the object exclusive constraints
+        result_stype = ctx.env.set_types[result_set]
+        obj_exclusives = get_object_exclusive_constraints(
+            result_stype, ptr_set, ctx.env)
+        for _, exc_ptrs in obj_exclusives:
+            if all(exc_ptr in ptr_set for exc_ptr in exc_ptrs):
```
elprans, line 1044: Perhaps `if not (exc_ptrs - ptr_set)`?

edb/edgeql/compiler/inference/cardinality.py
msullivan: Hm, why is being "more unique" a problem? The max cardinality is still ONE?

elprans:
```
function foo(first_name: str, last_name: str) -> str {
    USING (
        SELECT 
            first_name 
            ++ last_name 
            ++ <str>count((
                SELECT User 
                FILTER (.first_name, .last_name) = (first_name, last_name)
            ))
    )
}

type Foo { 
    property first_name -> str; 
    property last_name -> str; 
    constraint exclusive on (foo(.first_name, .last_name));
}
```

That said, as I was writing the above, I remembered that we shouldn't allow non-`Immutable` functions in constraint expressions, so this issue is probably moot.

msullivan: Oh, yeah; I said stable above but should have said immutable.

With that, do you think it's worth making this simplification/change?

elprans: Let's do it in a separate PR.

msullivan: Sure. I think that PR will end up being "delete half of this PR"

elprans (approved):

PR #2418, opened 2021-03-31
Prohibit mixing computable and regular links/props.

### LINKED ISSUE #2099: Disallow mixing computable and non-computable properties in inheritance
The EdgeQL compiler is currently not equipped to properly handle inheritance trees with different computable expressions, and until it does, prohibit overloading non-computable links and properties with with computables, and vice-versa (and also overloading a computable with another computable).

vpetrovykh:
Overloading computables presents some complication, so for the moment we
prohibit overloading computables or overloading a regular link/prop with
a computable.

Fixes #2099

### DIFF
--- a/edb/schema/pointers.py
+++ b/edb/schema/pointers.py
@@ -20,11 +20,12 @@ from __future__ import annotations
 from typing import *
 
 import collections.abc
+import enum
 import json
 
 from edb import errors
 
-from edb.common import enum
+from edb.common import enum as s_enum
 from edb.common import struct
 
 from edb.edgeql import ast as qlast
@@ -51,9 +52,15 @@ if TYPE_CHECKING:
     from . import sources as s_sources
 
 
-class PointerDirection(enum.StrEnum):
+class PointerDirection(s_enum.StrEnum):
     Outbound = '>'
     Inbound = '<'
+
+
+class LineageStatus(enum.Enum):
+    VALID = 0
+    MULTIPLE_COMPUTABLES = 1
+    MIXED = 2
 
 
 def merge_cardinality(
@@ -1381,6 +1388,15 @@ class PointerCommand(
             self._validate_pointer_def(schema, context)
         return schema
 
+    def _create_finalize(
+        self,
+        schema: s_schema.Schema,
+        context: sd.CommandContext,
+    ) -> s_schema.Schema:
+        schema = super()._create_finalize(schema, context)  # type: ignore
+        self._validate_computables(schema, context)
+        return schema
+
     def _alter_begin(
         self,
         schema: s_schema.Schema,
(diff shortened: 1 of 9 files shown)


edb/schema/pointers.py
```diff
@@ -1335,6 +1351,130 @@ def _alter_begin(
...
+        if scls.get_from_alias(schema):
+            return
+
+        is_computable = scls.is_pure_computable(schema)
+        is_owned = scls.get_owned(schema)
+        # Get the non-generic, explicitly declared ancestors as the
+        # limitations on computables apply to explicitly declared
+        # pointers, not just a long chain of inherited ones.
+        #
+        # Because this is potentially nested inside a command to
```
msullivan, line 1379: What is the case where this can come up? If the ancestor has been deleted and so is this object, maybe there is a more straightforward way to just skip all this logic? If the ancestor has been deleted and this object is being rebased, then it might be a bug elsewhere that the rebase hasn't happened first, though I'm not totally sure.

elprans: This is definitely masking a bug.

msullivan: Aha, it's because when a property is deleted, it first deletes its constraints (which will generate an alter of the child prop to delete *its* constraint) and *then* its children. (And that pattern shows up in cases where it doesn't need to because `ordering` generates wild and crazy diffs when inheritance is involved.) Maybe a cleaner expedient fix would be to skip validating computables unless something potentially relevant to computables changed (which might be a slight perf advantage too), but it's your call.

I found another bug while investigating this:
From 
```
abstract type Text { property x -> str { constraint exclusive } }         
abstract type Owned { property x -> str { constraint exclusive } }        
type Comment extending Text, Owned;
```
running `ALTER TYPE Text DROP PROPERTY x` gets us a fun `AttributeError: 'AlterProperty' object has no attribute 'scls'` crash. I'm going to follow up on that, which might end up cleaning up this whole snag.

edb/schema/pointers.py
```diff
@@ -1335,6 +1351,130 @@ def _alter_begin(
...
+            # We have a computable with some non-empty lineage. Which
+            # could be valid only if this is some aliasing followed by
+            # regular pointers only.
+            prev_shortname = head.get_shortname(schema)
+            prev_is_comp = True
+            for b in rest:
+                cur_is_comp = b.is_pure_computable(schema)
+                cur_shortname = b.get_shortname(schema)
+                if prev_is_comp:
+                    # Computables cannot overload, but they can alias
```
msullivan, line 1457: Could we check `from_alias` directly instead of guessing based on name equality?

msullivan (approved):
Looks good

edb/schema/pointers.py
vpetrovykh: `from _alias` would mark the pointer that was derived, but the thing that we want to ensure is that the original pointer actually had a different name, so it was aliased and not overridden. I don't think we have anything labelling the original pointer in that way.

vpetrovykh:
Is this still relevant or has it been superceded?

PR #2425, opened 2021-04-01
Populate scls in init_delta_command

msullivan:
Currently we are populating it in init_delta_branch but not
init_delta_command. This caused a bug when rebasing constraints
because of a drop in one parent where an scls was expected
but not present.

### DIFF
--- a/edb/schema/constraints.py
+++ b/edb/schema/constraints.py
@@ -458,12 +458,11 @@ class ConstraintCommand(
 
         base: Optional[so.Object] = None
         if isinstance(self, AlterConstraint):
-            scls = self.get_object(schema, context)
-            base = scls.get_subject(schema)
+            base = self.scls.get_subject(schema)
         else:
             referrer_ctx = self.get_referrer_context(context)
             if referrer_ctx:
-                base = referrer_ctx.op.get_object(schema, context)
+                base = referrer_ctx.op.scls
 
         if base is not None:
             # Concrete constraint
--- a/edb/schema/constraints.py
+++ b/edb/schema/constraints.py
@@ -458,11 +458,12 @@ class ConstraintCommand(
 
         base: Optional[so.Object] = None
         if isinstance(self, AlterConstraint):
-            base = self.scls.get_subject(schema)
+            scls = self.get_object(schema, context)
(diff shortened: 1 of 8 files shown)


elprans:
Hm. `self.scls` is populated when you `apply()` a command.  Is there code that is attempting to reference `.scls` on an unapplied command?  It probably should call `cmd.get_object(schema, context)` instead

msullivan:
Yeah, there is a bunch of code that references `.scls` on unapplied commands, generally because it is code that can be called both from an actually command application or because something got created with `init_delta_branch` and some method got called on it. 

I'll see how painful it is to change that (and then decide whether I want to make the argument for landing this diff as is)

elprans:
The trouble with relying on `.scls` on a non-applied command is that it wouldn't be set on a command produced by processing DDL AST.  In any case, an accessor method that checks `.scls` and falls back to `get_object` would be best here,  `.scls` shouldn't be referenced directly by outside code.

msullivan:
OK, I pushed a commit that backs out the addition of an scls assignment in `init_delta_command` while continuing to remove the previously existing assignments to scls (in `init_delta_branch` and `_propagate_if_expr_refs`).

I think it's the wrong approach and we should stick with the original diff. There are quite a few functions that get called both via apply and directly after being created with `init_delta_branch` or
`init_delta_command`. I think it is not very helpful and more than a little hurtful to disallow that code from using `scls`, especially with no typechecker support for catching the error. (This would be especially annoying when making a change that causes a code path to be called on an `init_delta_branch`d object that previously hadn't been.)

It's true that commands arising during processing the DDL ast wouldn't have `scls` on them, but I think the split between "DDL AST processing" and "canonicalizing/applying the tree" is a clearer one with much more segregated code paths than "has had apply() called on it" or not.

(I also kind of think that all the ast functions ought to not take a schema at all. My impression has been that most of the time they actually use the schema in a meaningful way, it is a bug.)

elprans:
> I think it's the wrong approach and we should stick with the original diff

OK, let's go with a minimal fix for now and think about possible cleanups later.

elprans (approved):

PR #2431, opened 2021-04-05
Add TransactionConflictError

fmoor:
TransactionConflictError is needed to make checking for retryable transaction errors simple.

See also the discussion following https://github.com/edgedb/edgedb-go/pull/123#pullrequestreview-625664587

### DIFF
--- a/edb/api/errors.txt
+++ b/edb/api/errors.txt
@@ -111,8 +111,9 @@
 0x_05_02_00_03   MissingRequiredError
 
 0x_05_03_00_00   TransactionError
-0x_05_03_00_01   TransactionSerializationError  #SHOULD_RETRY
-0x_05_03_00_02   TransactionDeadlockError  #SHOULD_RETRY
+0x_05_03_01_00   TransactionConflictError  #SHOULD_RETRY
+0x_05_03_01_01   TransactionSerializationError
+0x_05_03_01_02   TransactionDeadlockError
 
 
 ####


edb/api/errors.txt
```diff
@@ -111,8 +111,9 @@
 0x_05_02_00_03   MissingRequiredError
 
 0x_05_03_00_00   TransactionError
-0x_05_03_00_01   TransactionSerializationError  #SHOULD_RETRY
-0x_05_03_00_02   TransactionDeadlockError  #SHOULD_RETRY
+0x_05_03_01_00   TransactionConflictError
```
1st1, line 114: I think you can specify `#SHOULD_RETRY` for the new `TransactionConflictError` (and there shouldn't be need to specify it for the leaf error types, right @tailhook?)

fmoor: updated

1st1: @tailhook feel free to merge in the morning your time

tailhook (approved):

tailhook:
Merged. Thanks!

edb/api/errors.txt
fmoor: @1st1 were you expecting `edb gen-errors-json --client` to tag the deadlock and serialization errors with `SHOULD_RETRY`? It doesn't look like it does.

1st1: Yeah, I did, that was my question to Paul.

PR #2432, opened 2021-04-06
Generated classes for #2431

tailhook:
All of us forgot generating classes


tailhook:
Looks like errors here are due to breaking exceptions in python bindings. This is cyclic dependency since python bindings tests will also fail due to this incompatibility.

1st1 (approved):

PR #2436, opened 2021-04-06
add SHOULE_RETRY tag to transaction errors

fmoor:
These tags were erroneously left out of https://github.com/edgedb/edgedb/pull/2431.


fmoor:
going to make `edb gen-errors-json` infer tags from parent types instead of adding them explicitly

PR #2437, opened 2021-04-06
make errors inherit tags from their parents

fmoor:
The errors generated from edb/api/errors.txt now inherit tags from their parents so that subclasses don't need to explicitly define all their tags.


1st1 (approved):
I appreciate the simplification of the spaghetti code I had here before ;-)

PR #2438, opened 2021-04-07
Reserve 15 backend connections in testmode

fantix:
Some test cases will start a second EdgeDB server (default max_backend_connections=10), so we should reserve some backend connections for that. This is ideally calculated upon the edb test -j option, but that also depends on the total available memory. We are hard-coding 15 reserved connections here for simplicity.

### DIFF
--- a/edb/server/main.py
+++ b/edb/server/main.py
@@ -58,6 +58,7 @@ from . import protocol
 
 
 BYTES_OF_MEM_PER_CONN = 100 * 1024 * 1024  # 100MiB
+NUM_RESERVED_CONNS_IN_TESTMODE = 15
 
 logger = logging.getLogger('edb.server')
 _server_initialized = False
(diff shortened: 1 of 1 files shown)


edb/server/main.py
```diff
@@ -273,14 +274,24 @@ def run_server(args: ServerConfig):
 
     cluster: Union[pgcluster.Cluster, pgcluster.RemoteCluster]
     if args.data_dir:
-        if not args.max_backend_connections:
+        pg_max_connections = args.max_backend_connections
+        if not pg_max_connections:
             max_conns = _compute_default_max_backend_connections()
+            pg_max_connections = max_conns
```
elprans, line 280: Please add a comment here on why we're doing this (the text of the commit message is perfect).

elprans (approved):

PR #2442, opened 2021-04-07
Add pool simulation automation

fantix:
This PR adds QoS score to the simulation tests. Running for 10x time makes the score relatively stable and sensitive to major changes to the pool. Therefore, this simulation is made a workflow, configured to run on every commit or PR that touches the pool or the test. The generated reports are published here: https://edgedb.github.io/edgedb-pool-simulation/. The workflow will report failure if the total score is under 80 out of 100. For now it's stablized at about 93.

~In the following steps, I'll do the following:~

- [x] 1. Adjust the simulation pressure based on detected host performance. (done in #2480)
- [x] 2. Run the simulation for longer time in this workflow (like simply time x 10) for a more reliable result.
- [x] 3. Yield a QoS score (relative to host performance) for each simulation, and warn about a sudden score drop.
- [x] 4. (final goal) Add more simulations, and optimize the pool for non-starving situations like connection garbage collection, reusability, rebalancing, etc. (done in #2480)

### DIFF
--- a/tests/test_server_pool.py
+++ b/tests/test_server_pool.py
@@ -45,6 +45,7 @@ import os
 import random
 import statistics
 import string
+import sys
 import textwrap
 import time
 import typing
@@ -56,6 +57,14 @@ from edb.server import connpool
 
 # TIME_SCALE is used to run the simulation for longer time, the default is 1x.
 TIME_SCALE = int(os.environ.get("TIME_SCALE", '1'))
+
+# Running this script individually for the simulation test will exit with
+# code 0 if the final score is above MIN_SCORE, non-zero otherwise.
+MIN_SCORE = int(os.environ.get('MIN_SCORE', 80))
+
+# As the simulation test report is kinda large (~5MB with TIME_SCALE=10),
+# the CI automation will delete old reports beyond CI_MAX_REPORTS.
+CI_MAX_REPORTS = int(os.environ.get('CI_MAX_REPORTS', 50))
 
 C = typing.TypeVar('C')
 
@@ -502,7 +511,7 @@ class SimulatedCase(unittest.TestCase, metaclass=SimulatedCaseMeta):
 
         return js_data
 
-    def simulate_all_and_collect_stats(self):
+    def simulate_all_and_collect_stats(self) -> int:
         specs = {}
         for methname in dir(self):
             if not methname.startswith('test_'):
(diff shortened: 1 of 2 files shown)


1st1:
I'd change the sorting to make sure that the latest simulation is on top of the table.

But I have a bigger question: how will we use this? Like if there's a regression, how will we discover it? It's not like someone would browse these results every day.

fantix:
> I'd change the sorting to make sure that the latest simulation is on top of the table.

Good catch! Done.

> But I have a bigger question: how will we use this? Like if there's a regression, how will we discover it? It's not like someone would browse these results every day.

Just added QoS score logic, the CI should fail if the score is below 80; but I think it's okay not to block PR merges for now.

tests/test_server_pool.py
```diff
@@ -217,6 +250,111 @@ def calc_total_percentiles(
...
+            )
+        )
+
+    def calc_score(
+        self, value: float, *,
+        v100: float = 0,
+        v90: float,
+        v60: float,
+        v0: float = 2,
+    ) -> float:
```
1st1, line 281: Can you please add a comment somewhere explaining how the score is calculated?

tests/test_server_pool.py
```diff
@@ -340,33 +494,87 @@ def simulate_all_and_collect_stats(self):
...
-            f.write(html)
+        print('Final QoS score:', score)
+        return score
 
 
 class TestServerConnpoolSimulation(SimulatedCase):
 
     def test_server_connpool_1(self):
+        def calc_score(sim: Simulation):
+            return self.calc_group_score_with_weight(
```
1st1, line 561: Yeah, just glancing at it I've no idea how to use it. Docstrings would be hugely appreciated!

tests/test_server_pool.py
fantix: Oh yes, absolutely!

tests/test_server_pool.py
fantix: Will do!

fantix:
> Maybe we should also require that the score should be bigger than the score of the naive algo. But I see at least one sim where they're kinda bound to be the same.

That would be Sim 6 which I haven't found a proper scoring rule. As I continue to work on the single-db scenario, it might be soon necessary to have one. So yeah, checking that we beat the naive pool in QoS makes sense to me.

1st1 (approved):

Issue #2430, opened 2021-04-02
query header to toggle implicit object id


fmoor:
A query header that toggles implicit object ids should be added. This would make bindings in statically typed languages more flexible since they would not have to require that users provide an id field to decode the value into.

1st1:
Feel free to work on this Frederick!

PR #2443, opened 2021-04-08
add header to disable implicit ids

fmoor:
fixes https://github.com/edgedb/edgedb/issues/2430


edb/server/compiler/compiler.py
```diff
@@ -104,6 +104,7 @@ class CompileContext:
     implicit_limit: int = 0
     inline_typeids: bool = False
     inline_typenames: bool = False
+    inline_shapeids: bool = True
```
elprans, line 107:
```suggestion
    inline_objectids: bool = True
```

edb/server/compiler/compiler.py
```diff
@@ -549,10 +550,13 @@ def _compile_ql_query(
 
         single_stmt_mode = ctx.stmt_mode is enums.CompileStatementMode.SINGLE
 
-        can_have_implicit_fields = (
-            native_out_format and
-            single_stmt_mode
-        )
+        if not ctx.inline_shapeids:
+            can_have_implicit_fields = False
```
elprans, line 554: Why are object ids special here and not like typeid?

fmoor: I'm not sure exactly what you have in mind. I didn't want to break backward compatibility. Should this behavior be controlled entirely by the header instead of the combination of the header, protocol, and statement mode?

docs/internals/protocol/messages.rst
```diff
@@ -280,6 +280,10 @@ Known headers:
 * 0xFF04 ``ALLOW_CAPABILITIES``: ``uint64`` -- optional bitmask of
   capabilities allowed for this query.  See RFC1004_ for more information.
 
+* 0xFF05 ``IMPLICIT_SHAPEIDS`` -- "true" by default.
```
1st1, line 283: * I think it's better to call this `OMIT_OBJECTIDS`.

edb/server/compiler/compiler.py
1st1: Well, there can be a case of the user does not want to inline IDs but wants typenames. Your change disables typenames and typeids if object IDs are omitted.

edb/server/compiler/compiler.py
fmoor: fixed

edb/server/compiler/compiler.py
fmoor: oh got it!

fmoor: changed

docs/internals/protocol/messages.rst
fmoor: To invert the default value? Maybe `EXPLICIT_OBJECTIDS` is better? "omit" sounds like they will forcefully be omitted always.

1st1: > Maybe EXPLICIT_OBJECTIDS is better?

Yeah, much better, good idea.

edb/server/protocol/binary.pyx
```diff
@@ -1187,6 +1195,9 @@ cdef class EdgeConnection:
                 elif k == QUERY_HEADER_ALLOW_CAPABILITIES:
                     self.version_check("ALLOW_CAPABILITIES header", (0, 9))
                     allow_capabilities = parse_capabilities_header(v)
+                elif k == QUERY_HEADER_EXPLICIT_OBJECTIDS:
+                    self.version_check("EXPLICIT_OBJECTIDS header", (0, 9))
```
1st1, line 1199: Perhaps we should bump the proto to 0.10

1st1 (commented):
LGTM in general, I think we should bump the proto version. @elprans

elprans (approved):

edb/server/protocol/binary.pyx
fmoor: changed

Issue #2069, opened 2020-12-20
No way to supply --runstate-dir on edgedb init


isidentical:
When trying to setup an EdgeDB instance on a limited environment; EdgeDB throws this warnings,
```
$ ~/.edgedb/bin/edgedb server init --nightly default
...
2020-12-20T11:44:34.0903426Z CRITICAL 6748 2020-12-20 11:44:34,089 edb.server: cannot create the runstate directory: '/run/user/1001' does not exist; please use --runstate-dir to specify the correct location
```
but I couldn't find how to pass this as an option;
```
 $ ~/.edgedb/bin/edgedb server init --runstate-dir x --nightly test
error: Found argument '--runstate-dir' which wasn't expected, or isn't valid in this context

If you tried to supply `--runstate-dir` as a PATTERN use `-- --runstate-dir`

USAGE:
    edgedb server init [FLAGS] [OPTIONS] <name>

For more information try --help
```
or
```
 $ ~/.edgedb/bin/edgedb --runstate-dir server init x --nightly test 
error: Found argument '--runstate-dir' which wasn't expected, or isn't valid in this context

If you tried to supply `--runstate-dir` as a PATTERN use `-- --runstate-dir`

USAGE:
    edgedb [FLAGS] [OPTIONS] [SUBCOMMAND]

For more information try --help
```

NatureGeorge:
same issue in WSL2 (win10) for `1.0a7-2020122418~focal`

```bash
(base) zzf@LAPTOP-MGPQT3HG:~$ edgedb server init myinstance
┌──────────────────────┬───────────────────────────────────────────────┐
│ Instance Name        │ myinstance                                    │
│ Mode                 │ User Service                                  │
│ Data Directory       │ /home/zzf/.local/share/edgedb/data/myinstance │
│ Credentials Path     │ /home/zzf/.edgedb/credentials/myinstance.json │
│ Database Server Port │ 10701                                         │
│ Default User         │ edgedb                                        │
│ Default Database     │ edgedb                                        │
│ EdgeDB Version       │ 1.0a7-2020122418~focal                        │
└──────────────────────┴───────────────────────────────────────────────┘
WARNING 133 2020-12-25 22:02:52,021 py.warnings: /usr/lib/x86_64-linux-gnu/edgedb-server-1-alpha7/lib/python3.8/site-packages/edb/server/main.py:598: DeprecationWarning: Option `--bootstrap` is deprecated, use `--bootstrap-only`
[...]

PR #2444, opened 2021-04-09
Add `runstate_dir` option to `tb.start_edgedb_server()`

fantix:
When devmode is off, starting EdgeDB server with --postgres-dsn requires a valid runstate_dir (default `[/var]/run/edgedb` needs permission, and EdgeDB won't remove the directory). That caused the failure of the test `test_server_ops_detect_postgres_pool_size`, and breaking our nightly builds.

Refs #2069

### DIFF
--- a/tests/test_server_ops.py
+++ b/tests/test_server_ops.py
@@ -168,7 +168,7 @@ class TestServerOps(tb.TestCase):
         port = edgedb_cluster.find_available_port()
         actual = random.randint(50, 100)
 
-        async def test(host):
+        async def test(pgdata_path):
             bootstrap_command = (
                 r'CONFIGURE SYSTEM INSERT Auth '
                 r'{ priority := 0, method := (INSERT Trust) }'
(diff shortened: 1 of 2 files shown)


tests/test_server_ops.py
```diff
@@ -178,6 +179,7 @@ async def test(host):
                 max_allowed_connections=None,
                 postgres_dsn=f'postgres:///?user=postgres&port={port}&'
                              f'host={host}',
+                runstate_dir=None if devmode.is_in_dev_mode() else host,
```
elprans, line 182: Why are we passing `host` as `runstate_dir` here?

fantix: Oh that's a bad naming - it's actually the path to a temp dir. Let me rename it.

fantix: Oh that's a bad naming - it's actually the path to a temp dir. Let me rename it.

elprans (approved):

PR #2451, opened 2021-04-12
Add a way to check server health via an HTTP request

elprans:
This adds a new `/server/status/ready` HTTP endpoint, which responds
with "OK" if things are OK.  This is intended to be used as a simple
check target for service control plane.

### DIFF
--- a/edb/server/protocol/system_api.py
+++ b/edb/server/protocol/system_api.py
@@ -1,0 +1,153 @@
+#
+# This source file is part of the EdgeDB open source project.
+#
+# Copyright 2019-present MagicStack Inc. and the EdgeDB authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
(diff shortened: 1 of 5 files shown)


edb/server/protocol/system_api.py
```diff
@@ -0,0 +1,153 @@
...
+        response.close_connection = True
+        return
+
+
+async def handle_status_request(
+    request,
+    response,
+    server,
+):
+    result = await execute(server, "SELECT 'OK'", {})
```
1st1, line 85: An unhandled exception would propagate and cause the server to return a 400, text/plain response. Is that acceptable? Maybe it should always return application/json?

1st1 (approved):

edb/server/protocol/system_api.py
elprans: No, errors are handled in the caller, see above.

1st1: Ah, I forgot my own spaghetti. You're right.

PR #2452, opened 2021-04-13
Fix C.UTF-8 detection on remote Postgres

fantix:
On Debian/Ubuntu, this locale is written `C.UTF-8` with the dash, while on other platforms like CentOS or hosted Postgres like DigitalOcean, it's `C.utf8` without the dash. This caused our nightly builds to fail when testing with a locally-started remote Postgres on Debian/Ubuntu.

Refs: https://github.com/edgedb/edgedb-pkg/commit/e89832b09ef7cc747e8becc79301efa6cc665666


elprans (approved):
Ha! Good catch!

PR #2453, opened 2021-04-13
Push shapes further down in the IR tree when possible

msullivan:
This avoids bugs with queries like
`SELECT ((SELECT User {deck}), User.deck)`,
which was being miscompiled as if it was
`SELECT ((SELECT User) {deck}, User.deck)`

### DIFF
--- a/edb/pgsql/compiler/dml.py
+++ b/edb/pgsql/compiler/dml.py
@@ -1491,7 +1491,7 @@ def process_link_values(
             input_rel_ctx.shapes_needed_by_dml.add(shape_expr)
 
             if ptr_is_required and enforce_cardinality:
-                input_rel_ctx.force_optional.add(shape_expr.path_id)
+                input_rel_ctx.force_optional.add(ir_expr.path_id)
 
             dispatch.visit(ir_expr, ctx=input_rel_ctx)
 
--- a/edb/pgsql/compiler/dml.py
+++ b/edb/pgsql/compiler/dml.py
@@ -481,60 +481,45 @@ def get_dml_range(
         return range_cte
 
 
-def compile_iterator_ctes(
-    iterators: Iterable[irast.Set],
+def compile_iterator_cte(
+    iterator_set: irast.Set,
     *,
     ctx: context.CompilerContextLevel
 ) -> Optional[pgast.IteratorCTE]:
 
     last_iterator = ctx.enclosing_cte_iterator
 
-    seen = set()
-    p = last_iterator
-    while p:
-        seen.add(p.path_id)
-        p = p.parent
-
-    for iterator_set in iterators:
-        # Because of how the IR compiler hoists iterators, we may see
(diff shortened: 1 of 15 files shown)


edb/pgsql/compiler/dml.py
```diff
@@ -637,6 +637,9 @@ def process_insert_body(
         for shape_el, shape_op in ir_stmt.subject.shape:
             assert shape_op is qlast.ShapeOp.ASSIGN
 
+            if shape_el.path_id.is_linkprop_path():
```
elprans, line 640: What is this new check a result of?

msullivan: Oh, right:
it lets us drop the logic in `_compile_view_shapes_in_set` that checked for linkprop paths and adjusted the path_tip; because previously linkprops got filtered out during an INSERT by inspecting the source. I think we can drop that test now, actually?

msullivan:
OK, I think I can also fix this to push the shapes down for DML, also, which I think will let us get rid of hoisted_iterators

msullivan:
Also it seems that this turns `test_edgeql_scope_tuple_14`, which previously we miscompiled, into an internal server error. I suppose that's an improvement?

edb/pgsql/compiler/dml.py
```diff
@@ -1488,6 +1473,23 @@ def process_link_values(
                     row_query, ir_stmt.subject.path_id,
                     env=input_rel_ctx.env),)
 
+            # Look for a shape that might provide link property values
```
elprans, line 1476: Let's not forget to remove this hack once we implement linkprop stickiness on derived types.

edb/pgsql/compiler/expr.py
```diff
@@ -599,6 +599,17 @@ def _compile_set(
             shape=[expr for expr, _ in ir_set.shape],
             ctx=ctx,
         )
+    elif ir_set.shape and ir_set in ctx.shapes_needed_by_dml:
+        # hoo, boy.
```
elprans, line 603: Please elaborate :-)

elprans (approved):
LGTM.  Very nice to be rid of the hoisting hack!

edb/pgsql/compiler/expr.py
msullivan: oops I thought I had gotten rid of this one

edb/pgsql/compiler/dml.py
msullivan: I'm not sure that linkprop stickiness will help us with that?

I think we could get rid of this part of the hack by adding something like `nested_shape_source: Optional[Set]`, to Set, that `_compile_view_shapes_in_set` populates? So that the rummaging around only happens in one place instead of two?

(I think maybe we could also get rid of the rummaging around with a viewgen refactor, but I'm not sure)

PR #2458, opened 2021-04-15
Directly track nested "shape sources" in the IR

msullivan:
This lets us dispense with the grungy while loop in
`process_link_values` and (for better or worse) makes the information
more easily available.


msullivan:
I'm a little ambivalent on whether to actually merge this. I'm leaning yes, but I don't love adding such a limited-use field to Set.

msullivan:
I think that I do actually want to merge this, @elprans

elprans (approved):

PR #2463, opened 2021-04-17
server: Implement the --emit-server-status=<DEST> server option

elprans:
The new `--emit-server-status=<DEST>` option replaces
`--echo-runtime-info`.  Unlike the latter, `--emit-server-status`
accepts an explicit file path or a file descriptor to emit server
information to.  This lets server `stdout` be undisturbed, and adds
a way for future additions such as reporting server status via a
socket, which might be useful for a service supervisor.

This also does a bunch of cleanups in the `server_ops` tests.

### DIFF
--- a/edb/server/server.py
+++ b/edb/server/server.py
@@ -105,6 +105,7 @@ class Server:
         netport,
         auto_shutdown: bool=False,
         echo_runtime_info: bool = False,
+        status_sink: Optional[Callable[[str], None]] = None,
         startup_script: Optional[StartupScript] = None,
     ):
 
@@ -145,6 +146,7 @@ class Server:
         self._auto_shutdown = auto_shutdown
 
         self._echo_runtime_info = echo_runtime_info
+        self._status_sink = status_sink
 
         self._startup_script = startup_script
 
@@ -846,6 +848,14 @@ class Server:
                 "runstate_dir": str(self._runstate_dir),
             }
             print(f'\nEDGEDB_SERVER_DATA:{json.dumps(ri)}\n', flush=True)
+
+        if self._status_sink is not None:
+            status = {
+                "port": self._listen_port,
+                "socket_dir": str(self._runstate_dir),
+                "main_pid": os.getpid(),
+            }
+            self._status_sink(f'READY={json.dumps(status)}')
 
     async def stop(self):
         try:
--- a/edb/server/cluster.py
+++ b/edb/server/cluster.py
@@ -22,6 +22,7 @@ from __future__ import annotations
 import asyncio
 import json
 import os
+import socket
 import subprocess
 import sys
 import tempfile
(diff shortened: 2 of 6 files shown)


edb/server/server.py
```diff
@@ -847,6 +849,14 @@ async def start(self):
             }
             print(f'\nEDGEDB_SERVER_DATA:{json.dumps(ri)}\n', flush=True)
 
+        if self._status_sink is not None:
+            status = {
+                "port": self._listen_port,
+                "socket_dir": str(self._runstate_dir),
+                "main_pid": os.getpid(),
+            }
+            self._status_sink(f'READY={json.dumps(status)}')
```
1st1, line 858: Let's use JSON all the way:

```suggestion
            self._status_sink(json.dumps({"status": "ready", **status}))
```

elprans: The idea here is that `READY=` is the message type and JSON is payload.  It's also easier to match.

1st1: The only reason we had prefix with `--echo` is because stdout is full of garbage and we needed a marker. Here though, we can just readlines and parse JSON. So I'm -1 on the `READY=` prefix.

elprans: Markers are still needed, because `READY` wouldn't necessarily be the only kind of status message in there.  I'm not sure why line-oriented `<msg-kind>=<payload>` protocol is controversial here.  Pure JSON would require full parsing, which wouldn't be possible in shell without an external tool like `jq`.

(FWIW, I modeled this after systemd's `sd_notify`.)

1st1: >  I'm not sure why line-oriented <msg-kind>=<payload> protocol is controversial here.

Because you're forcing me to parse this primitive protocol. Instead of just reading the line and parsing JSON, I have one extra step of splitting by the first `=`. There's no reason for this.

How about we replace `"status":"ready"` with `"kind":`"ready-message"`?

elprans: > Because you're forcing me to parse this primitive protocol.

Is this really a problem? `<IDENT>=` is basically message framing, which allows matching of or skipping over messages without parsing the payload.  Hiding the message type in the payload is not a better design in this regard, IMO.

> There's no reason for this.

There is at least one, and I'm repeating myself here: parsing JSON is hard in shell without `jq`, which is a shame if all I needed was implementing `--wait-until-ready` reliably in shell and didn't care about actual information in the payload.

elprans: > There's no reason for this.

Another reason is that if you want to use the nice deserializaion into `struct`-s API in Go or Rust, you need to know the message shape beforehand.

1st1: Fair enough, lgtm

1st1 (approved):

PR #2465, opened 2021-04-18
Add support for EdgeDB multitenancy on a single Postgres cluster

elprans:
It is currently not possible to host multiple instances of EdgeDB on the
same Postgres cluster.  This complicates upgrades, since it is not always
convenient to spin up a parallel Postgres cluster just to upgrade EdgeDB,
especially if the upgrade needs to happen automatically.

Solve this by introducing a concept of EdgeDB *tenant*.  Each tenant is 
given a unique identifier, which is then used to qualify all globally 
visible Postgres objects (databases and roles).  It is possible to specify
the tenant id explicitly via the new `--postgres-tenant-id` server
command-line parameter.  The default tenant ID is derived from EdgeDB
catalog version, so it is now technically possible to start multiple 
incompatible EdgeDB versions on the same cluster.  Currently, the server 
would still refuse to do so, but upcoming work will introduce a mode where
this is allowed, and that'll make upgrades of EdgeDB running on a remote
cluster a much simpler story.

### DIFF
--- a/edb/schema/defines.py
+++ b/edb/schema/defines.py
@@ -20,14 +20,12 @@
 from __future__ import annotations
 
 # Maximum length of Postgres tenant ID.
-MAX_TENANT_ID_LENGTH = 1 + 10
-#                      ^    ^
-#          tenant scheme    tenant ID
+MAX_TENANT_ID_LENGTH = 10
 
 # Maximum length of names that are reflected 1:1 to Postgres:
-MAX_NAME_LENGTH = 63 - MAX_TENANT_ID_LENGTH - 1
-#                 ^                           ^
-#         max Postgres name len        tenant_id separator
+MAX_NAME_LENGTH = 63 - MAX_TENANT_ID_LENGTH - 1 - 1
+#                 ^                           ^   ^
+#    max Postgres name len     tenant_id scheme   tenant_id separator
 
 # Maximum number of arguments supported by SQL functions.
 MAX_FUNC_ARG_COUNT = 100
--- a/edb/schema/defines.py
+++ b/edb/schema/defines.py
@@ -19,9 +19,15 @@
 
 from __future__ import annotations
 
-# Maximum length of column name in Postgres. This limits the database
-# name length and affects column name mangling.
-MAX_NAME_LENGTH = 63
+# Maximum length of Postgres tenant ID.
+MAX_TENANT_ID_LENGTH = 1 + 10
+#                      ^    ^
+#          tenant scheme    tenant ID
+
+# Maximum length of names that are reflected 1:1 to Postgres:
+MAX_NAME_LENGTH = 63 - MAX_TENANT_ID_LENGTH - 1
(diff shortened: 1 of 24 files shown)


tests/test_server_ops.py
```diff
@@ -228,3 +228,43 @@ async def test(pgdata_path):
...
+                pgconnparams.ConnectionParameters(
+                    user='postgres',
+                    database='template1',
+                ),
+            )
+            self.assertTrue(cluster.ensure_initialized())
+            cluster.start()
+            try:
+                self.loop.run_until_complete(
+                    asyncio.wait([
```
fantix, line 264: `asyncio.wait()` doesn't propagate the errors, please use e.g. `taskgroup` instead.

edb/server/bootstrap.py
```diff
@@ -441,13 +460,13 @@ async def _make_stdlib(testmode: bool, global_ids) -> StdlibBits:
     schema = await _execute_edgeql_ddl(schema, stdglobals)
 
     _, global_schema_version = s_std.make_global_schema_version(schema)
```
fantix, line 462: I think `make_global_schema_version()` here also needs at least the `BackendRuntimeParams` from `ctx` to apply `CreateGlobalSchemaVersion` with the right tenant.

elprans: No, `make_global_schema_version` returns an abstract DDL command.  `_process_delta` is what compiles it into SQL and that gets the tenant id properly.

tests/test_server_ops.py
elprans: Good point.  Fixed.

fantix (approved):

edb/schema/defines.py
```diff
@@ -19,9 +19,15 @@
 
 from __future__ import annotations
 
-# Maximum length of column name in Postgres. This limits the database
-# name length and affects column name mangling.
-MAX_NAME_LENGTH = 63
+# Maximum length of Postgres tenant ID.
+MAX_TENANT_ID_LENGTH = 1 + 10
```
1st1, line 23: Do you maybe want to reserve yet another 11 characters for future use, reducing the max length to 40? We won't be able to do this ever again.

edb/server/main.py
```diff
@@ -599,6 +615,11 @@ def _validate_compiler_pool_size(ctx, param, value):
     click.option(
         '--postgres-dsn', type=str,
         help='DSN of a remote Postgres cluster, if using one'),
+    click.option(
+        '--postgres-tenant-id', type=str,
+        help='Specifies the tenant ID of this server when hosting'
+             ' multiple EdgeDB instances on one Postgres cluster',
```
1st1, line 621: Can you add some info on what's valid input for this option?

1st1 (approved):

edb/schema/defines.py
elprans: I don't think we need to reserve more.

Issue #2343, opened 2021-03-09
AssertionError: Normalized and non-normalized query errors differ


Fogapod:
- EdgeDB Version: `1-beta1`

Steps to Reproduce:

1. `select <tuple<"">>1;`
<details>
  <summary>Traceback</summary>

  ```python
  ERROR: InternalServerError: Normalized and non-normalized query errors differ
    Hint: This is most likely a bug in EdgeDB. Please consider opening an issue ticket at https://github.com/edgedb/edgedb/issues/new?template=bug_report.md
    Server traceback:
        Traceback (most recent call last):
          File "/usr/lib/x86_64-linux-gnu/edgedb-server-1-beta1/lib/python3.9/site-packages/edb/common/parsing.py", line 403, in parse
            self.parser.token(token)
          File "/usr/lib/x86_64-linux-gnu/edgedb-server-1-beta1/lib/python3.9/site-packages/parsing/__init__.py", line 209, in token
            self._act(token, tokenSpec)
          File "/usr/lib/x86_64-linux-gnu/edgedb-server-1-beta1/lib/python3.9/site-packages/parsing/__init__.py", line 233, in _act
            raise UnexpectedToken("Unexpected token: %r" % sym)
        parsing.errors.UnexpectedToken: Unexpected token: <Token LANGBRACKET "<">
  
        The above exception was the direct cause of the following exception:
  
        Traceback (most recent call last):
          File "/usr/lib/x86_64-linux-gnu/edgedb-server-1-beta1/lib/python3.9/site-packages/edb/server/compiler/compiler.py", line 1548, in _compile
            return self._try_compile(ctx=ctx, source=source)
          File "/usr/lib/x86_64-linux-gnu/edgedb-server-1-beta1/lib/python3.9/site-packages/edb/server/compiler/compiler.py", line 1580, in _try_compile
            statements = edgeql.parse_block(source)
          File "/usr/lib/x86_64-linux-gnu/edgedb-server-1-beta1/lib/python3.9/site-packages/edb/edgeql/parser/__init__.py", line 69, in parse_block
            return parser.parse(source)
          File "/usr/lib/x86_64-linux-gnu/edgedb-server-1-beta1/lib/python3.9/site-packages/edb/common/parsing.py", line 416, in parse
            raise self.get_exception(
        edb.errors.EdgeQLSyntaxError: Unexpected '<'
  
        During handling of the above exception, another exception occurred:
  
        Traceback (most recent call last):
[...]

PR #2473, opened 2021-04-20
remove query normalization error consistency check

fmoor:
fixes https://github.com/edgedb/edgedb/issues/2343

The error consistency check is dropped because errors from normalized and non-normalized queries are not guaranteed to be the same.

### DIFF
--- a/edb/server/compiler/compiler.py
+++ b/edb/server/compiler/compiler.py
@@ -1565,11 +1565,7 @@ class Compiler:
                 except errors.EdgeQLSyntaxError as denormalized_err:
                     raise denormalized_err
                 except Exception:
-                    raise AssertionError(
-                        "Normalized and non-normalized query errors differ")
-                else:
-                    raise AssertionError(
-                        "Normalized query is broken while original is valid")
+                    raise original_err
             else:
                 raise original_err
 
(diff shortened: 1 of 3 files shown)


edb/server/compiler/compiler.py
```diff
@@ -1565,11 +1565,7 @@ def _compile(
                 except errors.EdgeQLSyntaxError as denormalized_err:
                     raise denormalized_err
                 except Exception:
-                    raise AssertionError(
-                        "Normalized and non-normalized query errors differ")
-                else:
-                    raise AssertionError(
```
elprans, line 1571: This check is still useful to catch normalization errors.

elprans (commented):
Before we mark #2343 as closed, we should also fix the original issue, where `select <tuple<"">>1;` raises an ISE instead of `SyntaxError`

fmoor:
> we should also fix the original issue, where `select <tuple<"">>1;` raises an ISE instead of `SyntaxError`

@elprans is https://github.com/edgedb/edgedb/pull/2473/commits/b4356c547c4873da21a1e8d479250958cf639f8e what you had in mind for the fix?

elprans (approved):

PR #2475, opened 2021-04-20
Include catalog version in package version metadata

elprans:
This includes `EDGEDB_CATALOG_VERSION` as build metadata in the version
string in the form `cv{version}`.  Also, the build metadata is now used
in release builds also.

### DIFF
--- a/.github/workflows.src/nightly.tpl.yml
+++ b/.github/workflows.src/nightly.tpl.yml
@@ -10,14 +10,40 @@ on:
       - nightly
 
 jobs:
+  prep:
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.whichver.outputs.version }}
+      branch: ${{ steps.whichver.outputs.branch }}
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Determine package version
+      shell: bash
+      run: |
+        branch=${GITHUB_REF#refs/heads/}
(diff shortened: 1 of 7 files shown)


tailhook:
Changeset is okay. But I'm not sure if we want this.

Our current upgrade logic:
1. Major version is always dump/restore
2. New nightly version is also always dump/restore

I'm not sure adding an extra number to the equation is useful. What if it differs for the minor version? It may create false impression that we can bump schema number in the minor version, which will be very disruptive for production users.

So if we need to use something for namespacing postgres databases by version, if don't see why we cant use major (for stable) and git commit (for nightly), or combination of major+commit for latter. Or hash of thereof.

elprans:
> Our current upgrade logic:
> 
> 1. Major version is always dump/restore
> 2. New nightly version is also always dump/restore

I have changed the slotting behavior of nightlies to use the catalog version instead of the commit number, because we need to fix build retention in such a way as to not keep every nightly build forever, but also to make sure that we retain builds with unique catalog versions to make sure you can always install a compatible server and dump your databases, even if it was nightly.

> I'm not sure adding an extra number to the equation is useful. What if it differs for the minor version? It may create false impression that we can bump schema number in the minor version, which will be very disruptive for production users.

Adding catalog version to build metadata is the only reasonable way to accomplish the above.  I don't think anyone would care about a string of numbers in the version metadata.

PR #2480, opened 2021-04-21
Pool optimization

fantix:
This PR has 5 commits:

## Single block: don't create connection if more than 1 available

Before this PR, the pool will keep adding connections on any `acquire()` when it's managing connections to only one database. That means we'll soon reach the max capacity even if we have only one frontend connection, as far as we keep issuing EdgeQL queries. This wasn't a big issue, just a waste of connections, especially after we add GC in this PR.

This commit checks if the current `acquire()` and the next `acquire()` could return immediately. If yes, don't create a new connection.

## Don't rebalance if `max_capacity` is not reached.

If we still have space for more connections, don't actively rebalance the pool just yet. Because the blocks could just create more connections and don't have to worry about other blocks just yet. Rebalance will kick in when the max
capacity is hit; or we'll depend on the garbage collection (see below) to shrink the over-quota blocks.

## Implement pool "Garbage Collection" (discard unused conns)

This commit makes the per-block connection queue *a stack*, with each connection paired with a monotonic timestamp indicating when it was pushed into the stack. GC is implemented using the same `try_steal` logic to pull connections from the bottom of the stack until a preset timestamp limit is met. GC is now running every 2 minutes after the first idle connection is freed into the pool, and pauses if no idle connection is freed in the last 2 GC cycles. So theoretically an idle connection will be discarded between 2-4 minutes after it's returned to the pool and not otherwise stolen or rebalanced.

## pool: Add comments

Explaining the pool logic. This will be mapped to a blog post later.

This commit also added a simplified `_steal_conn()` to the `NaivePool` so that it doesn't hang if a new block is requested while there's no connection returned in old blocks.

## Prune unused connections when creating from template database

Creating a database from a template database requires the template database have no active connections. With the above changes, now it's possible to have pooled but unused connections to the "template" database while running `CREATE DATABASE xxx FROM yyy`. We simply do the same as we did before dropping a database - prune the unused pooled connections, or raise an error if we know the database has connections.

### DIFF
--- a/edb/server/connpool/pool.py
+++ b/edb/server/connpool/pool.py
@@ -99,6 +99,24 @@ class ConnectionState:
 
 
 class Block(typing.Generic[C]):
+    # A Block holds a number of connections to the same backend database.
+    # A Pool consists of one or more blocks; blocks are the basic unit of
+    # connection pool algorithm, while the pool itself also takes care of
+    # balancing resources between Blocks (because all the blocks share the same
+    # PostgreSQL `max_connections` limit), based on realtime statistics the
+    # block collected and populated.
+    #
+    # Instead of the regular round-robin queue, EdgeDB adopted an LIFO stack
+    # (conn_stack) for the connections - the most recently used connections are
+    # always yielded first. This allows us to run "garbage collection" or
+    # "connection stealing" to recycle the unused connections from the bottom
+    # of the stack (the least recently used ones), so that other blocks could
+    # reuse the spared resource.
+    #
+    # Block is coroutine-safe. Multiple tasks acquiring connections will be put
+    # in a waiters' queue (conn_waiters), if the demand cannot be fulfilled
+    # immediately without blocking/awaiting. When connections are ready in the
+    # stack, the next task in the queue will be woken up to continue.
 
     loop: asyncio.AbstractEventLoop
     dbname: str
@@ -149,18 +167,26 @@ class Block(typing.Generic[C]):
         self._log_events = {}
 
     def count_conns(self) -> int:
+        # The total number of connections in this block, including:
+        #  - Future connections that are still pending in connecting
+        #  - Idle connections in the stack
+        #  - Acquired connections (not in the stack)
         return len(self.conns) + self.pending_conns
 
     def count_waiters(self) -> int:
+        # The number of tasks that are blocked in acquire()
         return self.conn_waiters_num
 
     def count_queued_conns(self) -> int:
+        # Number of connections in the stack/pool
         return len(self.conn_stack)
 
     def count_pending_conns(self) -> int:
+        # Number of future connections that are still pending in connecting
         return self.pending_conns
 
     def count_conns_over_quota(self) -> int:
+        # How many connections over the quota
         return max(self.count_conns() - self.quota, 0)
 
     def count_approx_available_conns(self) -> int:
@@ -183,10 +209,17 @@ class Block(typing.Generic[C]):
     def try_steal(
         self, only_older_than: typing.Optional[float] = None
     ) -> typing.Optional[C]:
+        # Try to take one unused connection from the block without blocking.
+        # If only_older_than is provided, only the connection that was put in
+        # the stack before the given timestamp is returned. None will be
+        # returned if we cannot find such connection in this block.
+
         if not self.conn_stack:
             return None
 
         if only_older_than is not None:
+            # We only need to check the bottom of the stack - higher items in
+            # the stack only have larger timestamps
             oldest_conn = self.conn_stack[0]
             if self.conns[oldest_conn].in_stack_since > only_older_than:
                 return None
@@ -201,6 +234,12 @@ class Block(typing.Generic[C]):
         self.conn_waiters_num += 1
         try:
             attempts = 0
+
+            # Skip the waiters' queue if we can grab a connection from the
+            # stack immediately - this is not completely fair, but it's
+            # extremely hard to always take the shortcut and starve the queue
+            # without blocking the main loop, so we are fine here. (This is
+            # also how asyncio.Queue is implemented.)
             while not self.conn_stack:
                 waiter = self.loop.create_future()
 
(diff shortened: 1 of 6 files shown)


edb/server/connpool/pool.py
```diff
@@ -108,7 +109,7 @@ class Block(typing.Generic[C]):
     conn_acquired_num: int
     conn_waiters_num: int
     conn_waiters: typing.Deque[asyncio.Future[None]]
-    conn_queue: typing.Deque[C]
+    conn_stack: typing.Deque[typing.Tuple[float, C]]
```
1st1, line 112: Hm, maybe we put the timestamp into `ConnectionState`?

edb/server/connpool/pool.py
```diff
@@ -1095,13 +1094,42 @@ def _maybe_tick(self) -> None:
...
+                self._log_to_snapshot(
+                    dbname=block.dbname, event='conn-stolen')
+                self._schedule_transfer(block, conn, for_block)
+                self._blocks.move_to_end(block.dbname, last=True)
+                return
+        # If all the blocks are busy, simply wait in the queue to get one.
+        for block in self._blocks.values():
+            if block is for_block:
+                continue
+            if block.count_conns():
```
1st1, line 1113: Hm, why should we block until we acquire a connection from a random block? I'm not sure this is right.

1st1: Ah, this is for the NaivePool. Never mind!

edb/server/connpool/pool.py
```diff
@@ -301,6 +301,8 @@ class BasePool(typing.Generic[C]):
     _blocks: collections.OrderedDict[str, Block[C]]
 
     _is_starving: bool
+    # Indicates if any block is starving for connections, this usually means
+    # the number of active blocks is greater than the pool max capacity.
```
1st1, line 305: Since this is the only comment added in this commit, I find the " pool: Add comments " message a bit misleading :)

1st1 (approved):
lgtm overall, but let's discuss if we want to put connection timestamp in the stack on in the ConnectionState struct

edb/server/connpool/pool.py
fantix: Oh right - I do intended to add more here, but don't want to delay the release.

edb/server/connpool/pool.py
fantix: Yeah I think that'll work - it's just a little weird to have:

```python
    def try_steal(
        self, only_older_than: float = float("inf")
    ) -> typing.Optional[C]:
        if (
            self.conn_stack
            and self.conns[self.conn_stack[0]].in_stack_since < only_older_than
        ):
            conn = self.conn_stack.popleft()
            assert conn in self.conns  # <--- we don't need this any more
            return conn
        return None
```

But I guess it runs at about the same speed, so I'm basically fine with either.

fantix: This is easier to read:

```python
    def try_steal(
        self, only_older_than: typing.Optional[float] = None
    ) -> typing.Optional[C]:

        if not self.conn_stack:
            return None
        
        conn = self.conn_stack.popleft()
        assert conn in self.conns
        
        if only_older_than is not None:
            if self.conns[conn].in_stack_since > only_older_than:
                self.conn_stack.appendleft(conn)
                return None
            
        return conn
```

1st1: This is a bit more optimal, imo:

```python
    def try_steal(
        self, only_older_than: typing.Optional[float] = None
    ) -> typing.Optional[C]:

        if not self.conn_stack:
            return None
        
        if only_older_than is not None:
            top_conn = self.conn_stack[0]
            if self.conns[top_conn].in_stack_since > only_older_than:
                return None            
            
        return self.conn_stack.popleft()
```

edb/server/connpool/pool.py
```diff
@@ -556,6 +606,42 @@ async def _discard_conn(self, block: Block[C], conn: C) -> None:
...
+    #   Mode A: managing connections to only one database
+    #   Mode B: multiple databases, below max capacity
+    #   Mode C: reached max capacity, some tasks are waiting for connections
+    #   Mode D: some blocks are starving with zero connection
+    #
+    # Mode A is close to a regular connection pool - new connections are only
+    # created when there are not enough spare ones in the pool, and used
+    # connections are released back to the pool, cached for next acquisition
+    # (unless being idle for too long and GC will recycle them). In the same
+    # code base, there are usually an early-exit for Mode A.
```
1st1, line 625: > In the same code base, there are usually an early-exit for Mode A.

Not sure I understand this sentence. Clarify?

edb/server/connpool/pool.py
```diff
@@ -556,6 +606,42 @@ async def _discard_conn(self, block: Block[C], conn: C) -> None:
...
+    #
+    # Mode B is simply an extension of Mode A for multiple databases. Each
+    # block in Mode B acts just like Mode A, with minimal difference like less
+    # aggressive connection creation. Different blocks could freely create new
+    # connections on demand, racing with each other naturally by the traffic.
```
1st1, line 630: by the traffic -> by the demand for Postgres connections

edb/server/connpool/pool.py
```diff
@@ -556,6 +606,42 @@ async def _discard_conn(self, block: Block[C], conn: C) -> None:
...
+    # to "transfer" a connection from another block. This is achieved by
+    # rebalancing the pool based on calculated per-block quotas recalibrated
+    # in periodic "ticks" (see _tick()).
+    #
+    # In extreme cases, the number of blocks may go beyond the max capacity.
+    # This is Mode D when even each block takes only at most one connection,
+    # there are still some starved blocks that have no connections at all.
+    # Mode D reuses the framework of Mode C but runs separate logic in a
+    # different if-else branch. In short, the pool reallocates the limited
+    # total number of connections to different blocks in turn like round-robin.
```
1st1, line 644: "in turn like round-robin." -> "in a round-robin fashion"

1st1 (approved):
Love the new comments, aside from a couple of minor nits. Really good job, Fantix!

Issue #2459, opened 2021-04-16
Inconsistent casting of `bytes`


vpetrovykh:
Weird thing about `<json>` casts is that `bytes` can be cast when part of a shape, but not directly.

```
db> SELECT <json>Foo{z := b'abc'};
{"{\"z\": \"\\\\x616263\"}"}
db> SELECT <json>b'abc';
error: cannot cast 'std::bytes' to 'std::json'
  ┌─ query:1:14
  │
1 │ SELECT <json>b'abc';
  │              ^^^^^^ error
```

vpetrovykh:
The problem is that there's no real standard for representing binary data in JSON. So we can solve this by picking one approach and declare it as standard for how EdgeDB will treat binary data w.r.t. JSON. We can use base 64 encoding for this representation and thus allow consistent conversion/casting between `bytes` and `json` in any context where these casts may appear. This also means that invalid base 64 encoding would result in an error when casting from `json` to `bytes` is attempted.

The alternative of disallowing any casting of `bytes` in all cases has the downside that it prevents users from even potentially accessing binary data via JSON interface in the rare cases when they might want to. So rather than forbidding this, documenting the format and letting the users handle the parsing of JSON is a more user-friendly solution.

PR #2482, opened 2021-04-21
implement casting between bytes and json

fmoor:
fixes https://github.com/edgedb/edgedb/issues/2459

Casting between bytes and json is enabled by encoding bytes to base64 encoded json strings.

### DIFF
--- a/edb/ir/typeutils.py
+++ b/edb/ir/typeutils.py
@@ -107,12 +107,12 @@ def is_abstract(typeref: irast.TypeRef) -> bool:
 
 def is_json(typeref: irast.TypeRef) -> bool:
     """Return True if *typeref* describes the json type."""
-    return typeref.id == s_obj.get_known_type_id('std::json')
+    return typeref.real_base_type.id == s_obj.get_known_type_id('std::json')
 
 
 def is_bytes(typeref: irast.TypeRef) -> bool:
     """Return True if *typeref* describes the bytes type."""
-    return typeref.id == s_obj.get_known_type_id('std::bytes')
+    return typeref.real_base_type.id == s_obj.get_known_type_id('std::bytes')
 
 
 def is_persistent_tuple(typeref: irast.TypeRef) -> bool:
--- a/edb/ir/typeutils.py
+++ b/edb/ir/typeutils.py
@@ -32,6 +32,7 @@ from edb.schema import pseudo as s_pseudo
 from edb.schema import scalars as s_scalars
 from edb.schema import types as s_types
 from edb.schema import objtypes as s_objtypes
+from edb.schema import objects as s_obj
 from edb.schema import utils as s_utils
 
 from . import ast as irast
@@ -102,6 +103,16 @@ def is_generic(typeref: irast.TypeRef) -> bool:
 def is_abstract(typeref: irast.TypeRef) -> bool:
     """Return True if *typeref* describes an abstract type."""
     return typeref.is_abstract
+
+
+def is_json(typeref: irast.TypeRef) -> bool:
+    """Return True if *typeref* describes the json type."""
+    return typeref.id == s_obj.get_known_type_id('std::json')
+
+
+def is_bytes(typeref: irast.TypeRef) -> bool:
+    """Return True if *typeref* describes the bytes type."""
+    return typeref.id == s_obj.get_known_type_id('std::bytes')
 
 
 def is_persistent_tuple(typeref: irast.TypeRef) -> bool:
--- a/edb/pgsql/compiler/output.py
+++ b/edb/pgsql/compiler/output.py
@@ -28,8 +28,10 @@ from edb.ir import ast as irast
 from edb.ir import typeutils as irtyputils
 
 from edb.schema import defines as s_defs
+from edb.schema import casts as s_casts
 
 from edb.pgsql import ast as pgast
+from edb.pgsql import common
 from edb.pgsql import types as pgtypes
 
 from . import astutils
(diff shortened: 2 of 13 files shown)


elprans (commented):
Looks good. Please add a test with shape casts to JSON as well.

tests/test_edgeql_json.py
```diff
@@ -1408,12 +1408,11 @@ async def test_edgeql_json_slice_03(self):
...
-                edgedb.QueryError, r'cannot cast.*bytes.*to.*json.*'):
-
-            await self.con.execute(r"""
-                SELECT <json>b'foo';
-            """)
+        await self.assert_query_result(
+            r""" SELECT <json>b'foo'; """,
+            ['Zm9v'],
+            ['"Zm9v"'],
+        )
```
1st1, line 1415: Please also add tests when a json type is in a shape, e.g.

```
SELECT <json>User {
  name,
  some_bytes := b'aaa'
}
```

and the same query without `<json>` but when compiled in JSON output mode.

fmoor: Casting shapes with `<json>` and `query_json()` don't work with the current changes. I've been tinkering with the compiler for a while trying to find the relative parts. If anyone has any pointers let me know.

elprans: It's an interesting issue, actually.  See `compile_TypeCast` in `edgeql/compiler/expr.py`.  Specifically the trailing `else:` branch sets up a context to compile the casted expression.  What we are missing here is another context flag that would tell the shape processor in `viewgen.py` that it is inside a JSON cast, and that every shape element value should be cast to `<json>`.  We do this type of recursion for collections, but not shapes, evidently.  This isn't an issue for most types, since PostgreSQL's cast is the same as ours, but `bytes` exposed the bug here.

fmoor: changed

edb/edgeql/compiler/expr.py
```diff
@@ -523,11 +523,20 @@ def compile_TypeCast(
                 # JSON wants type shapes and acts as an output sink.
                 subctx.expr_exposed = True
                 subctx.inhibit_implicit_limit = True
+
+            if (
+                isinstance(expr.type, qlast.TypeName) and
+                isinstance(expr.type.maintype, qlast.ObjectRef) and
+                expr.type.maintype.name == 'json'
```
fmoor, line 530: Is there a better way to check if the target type is `std::json`?

elprans: The `if` check above (`if target_stype.contains_json(subctx.env.schema):`) is doing exactly that.

fmoor:
This query `SELECT <json>schema::Type{name}` seems to trigger an edge case: 

[edb.errors.SchemaError: cannot redefine property 'builtin' of object type 'schema::Object' as scalar type 'std::json'](https://github.com/edgedb/edgedb/pull/2482/checks?check_run_id=2460148479#step:16:650)

I haven't started debugging this yet, but it seems like there are two inconsistencies with this error
1. the property `builtin` is not in the result shape
2. even if it was `builtin` of type `std::bool` should be castable to json right?

elprans:
It's not about castability, you cannot change the type of an existing link or property in a shape.  This is problem with this approach.  I think the solution would be to not to redefine the original properties, but inject magic computables instead, e.g `__json_name := <json>.name` and then teach the SQL compiler to look for these when serializing shapes.

fmoor:
> I think the solution would be to not to redefine the original properties, but inject magic computables instead, e.g `__json_name := <json>.name` and then teach the SQL compiler to look for these when serializing shapes.

I have no idea where to start with this :confused:

edb/edgeql/compiler/expr.py
fmoor: changed

fmoor:
> It's not about castability, you cannot change the type of an existing link or property in a shape.

Why does it only complain about the `builtin` property? It works fine with everything else.

elprans:
> Why does it only complain about the `builtin` property?

Because we do implicit rewrites on introspection schema objects to support field defaults, and evidently that rewrite interacts badly with this.  See `get_schema_reflection_default`

1st1:
@fmoor @elprans Hm, what's the status of this PR?

fmoor:
I think it is postponed. Maybe it should just be closed for now? @elprans

elprans:
I marked it as draft. We'll come back to this later.

1st1:
Should this one be closed in favor of https://github.com/edgedb/edgedb/pull/2920?

msullivan:
I took a different approach than the proposed one.

In addition to making bytes render as base64 when cast to json, we need to have them render that way in json output mode too. Once that is fixed, the whole object casting problem goes away, since casting objects to json simply relies on first serializing the object in a json output mode.

edb/pgsql/compiler/output.py
```diff
@@ -530,8 +543,18 @@ def serialize_expr_to_json(
...
-    elif path_id.is_collection_path() and not expr.ser_safe:
-        val = coll_as_json_object(expr, styperef=path_id.target, env=env)
+    elif irtyputils.is_collection(styperef) and not expr.ser_safe:
+        val = coll_as_json_object(expr, styperef=styperef, env=env)
+
+    elif (
+        irtyputils.is_bytes(styperef)
+        and not expr.ser_safe
+    ):
+        cast_name = s_casts.get_cast_fullname_from_names(
```
elprans, line 553: I'd add a TODO: note here to remind us to find a way to generalize this, since `bytes` wouldn't be the last scalar with a custom JSON cast.

elprans (approved):

edb/ir/typeutils.py
```diff
@@ -104,6 +105,16 @@ def is_abstract(typeref: irast.TypeRef) -> bool:
     return typeref.is_abstract
 
 
+def is_json(typeref: irast.TypeRef) -> bool:
+    """Return True if *typeref* describes the json type."""
+    return typeref.id == s_obj.get_known_type_id('std::json')
```
1st1, line 110: Should we use `.base_type.id` here?

msullivan: Yeah, good catch!

Issue #2332, opened 2021-03-08
Feature request: multiple string prefixes at once


Fogapod:
Currently you must use either `r` or `b`, not both:
```
edgedb> br"";
error: prefix "br" is not allowed for strings, allowed: `b`, `r`
  ┌─ query:1:1
  │
1 │ br"";
  │ ^ error
```
Prefixes can be combined in python, would be nice if EdgeDB could support this too.

1st1:
Yeah, it's an oversight we don't have them.

PR #2483, opened 2021-04-21
enable multiple string prefixes

fmoor:
fixes https://github.com/edgedb/edgedb/issues/2332

### DIFF
--- a/edb/edgeql-parser/tests/tokenizer.rs
+++ b/edb/edgeql-parser/tests/tokenizer.rs
@@ -557,6 +557,10 @@ fn strings() {
     assert_eq!(tok_typ(r#" b""  "#), [BinStr]);
     assert_eq!(tok_str(r#" b''  "#), [r#"b''"#]);
     assert_eq!(tok_typ(r#" b''  "#), [BinStr]);
+    assert_eq!(tok_str(r#" br""  "#), [r#"br"""#]);
+    assert_eq!(tok_typ(r#" br""  "#), [BinStr]);
+    assert_eq!(tok_str(r#" br''  "#), [r#"br''"#]);
+    assert_eq!(tok_typ(r#" br''  "#), [BinStr]);
     assert_eq!(tok_err(r#" ``  "#),
         "Unexpected `backtick quotes cannot be empty`");
 
@@ -572,6 +576,10 @@ fn strings() {
     assert_eq!(tok_typ(r#" b"hello"  "#), [BinStr]);
     assert_eq!(tok_str(r#" b'hello'  "#), [r#"b'hello'"#]);
     assert_eq!(tok_typ(r#" b'hello'  "#), [BinStr]);
+    assert_eq!(tok_str(r#" rb"hello"  "#), [r#"rb"hello""#]);
+    assert_eq!(tok_typ(r#" rb"hello"  "#), [BinStr]);
+    assert_eq!(tok_str(r#" rb'hello'  "#), [r#"rb'hello'"#]);
+    assert_eq!(tok_typ(r#" rb'hello'  "#), [BinStr]);
     assert_eq!(tok_str(r#" `hello`  "#), [r#"`hello`"#]);
     assert_eq!(tok_typ(r#" `hello`  "#), [BacktickName]);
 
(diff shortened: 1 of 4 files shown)


1st1:
Please add a functional query test too. Like `SELECT br'aaa' ++ rb'ccc'`

1st1:
Actually, I'm not sure we should support all combinations of literal prefixes. Maybe we just start with `rb`, short for `raw bytes literal`?

fmoor:
> Actually, I'm not sure we should support all combinations of literal prefixes. Maybe we just start with `rb`, short for `raw bytes literal`?

I don't think of `rb` as "raw bytes". I think of it as `raw AND bytes` in which case prohibiting `bytes AND raw` doesn't make sense.

1st1:
> I don't think of rb as "raw bytes".

I also don't think of rb that way. I specifically said "raw bytes literal" ;)

In any case, I just double checked what Python does, and Python 3.9 allows both `rb` and `br` prefixes (to my surprise, because I think it wasn't like that before), so I withdraw my suggestion.

edb/edgeql-parser/tests/tokenizer.rs
```diff
@@ -634,6 +634,8 @@ fn strings() {
     assert_eq!(tok_typ(" b\"hel\nlo\" "), [BinStr]);
     assert_eq!(tok_str(" b'hel\nlo' "), ["b'hel\nlo'"]);
     assert_eq!(tok_typ(" b'hel\nlo' "), [BinStr]);
+    assert_eq!(tok_str(" rb'hel\nlo' "), ["rb'hel\nlo'"]);
+    assert_eq!(tok_str(" br'hel\nlo' "), ["br'hel\nlo'"]);
```
tailhook, line 638: More tests needed:
1. Add `tok_typ` assertion, that makes sure that this is a binary string.
2. Add test that ensures that something that is not allowed in binary string like `\xFF` is allowed in raw string
3. A tests that disallows escaping quotes `rb'test\'`
4. Different quotes

Maybe look at other binary and raw string cases. This is quite sensitive piece of code. That we want all combinations of things tested against, in case we further optimize some tokenizer code paths in future. (Tiny exception I'm okay with: don't have to copy all the tests with `rb` and `br`, that seems to be trivially proven to work).

tailhook: Ah, yes, I misread `c if c as u32 > 0x7f` as if it doesn't allow hex escapes > 0x7f, while it doesn't allow actual character codes like that.

tailhook:
And yes, Yury right, that we also need a functional test, preferably that compares bytes to raw bytes:
```
SELECT br"test\raw" == b"test\\raw"
```
To make it work, decodiing `BinStr` has to be adjusted: https://github.com/edgedb/edgedb/blob/34bac5578414c61029138f2a6953373d7e3be1be/edb/edgeql-rust/src/tokenizer.rs#L554
(perhaps by bringing prefix parsing into the `unquote_bytes` function instead, like we do in `unquote_string`)

fmoor:
I've added a functional query test.

edb/edgeql-rust/src/tokenizer.rs
```diff
@@ -654,7 +654,15 @@ impl<'a> From<SpannedToken<'a>> for CowToken<'a> {
     }
 }
 
-fn unquote_bytes<'a>(s: &'a str) -> Result<Vec<u8>, String> {
+fn unquote_bytes<'a>(value: &'a str) -> Result<Vec<u8>, String> {
+    if value.starts_with("rb") || value.starts_with("br") {
```
tailhook, line 658: Let's do something like this:
```
let idx = value.find(|c| c == '\'' || c == '"').ok_or_else(...)?;
let prefix = value[..idx]
match prefix {
    "rb" | "br" => 
    "b" => 
    _ => Err(format!("unimplemented prefix {:?}", prefix)),
}
```
I think this would make potential new syntaxes safer.

tailhook (commented):
Please also add few smoke tests on the new `unquote_bytes` function.

edb/edgeql-rust/src/tokenizer.rs
fmoor: changed

tailhook (approved):

Issue #1418, opened 2020-05-26
Remove old CLI Python implementation


1st1:

elprans:
@ambv, there are a few references to Python CLI (`-m edb.cli`), which need to be replaced with appropriate invocations of `edgedb-cli` in the testing/packaging infra.

PR #2484, opened 2021-04-23
Remove Python-based CLI

elprans:
The old Python-based has been fully replaced by the Rust CLI and is no
longer needed in any context.

Closes: #1418


docs/internals/dev.rst
```diff
@@ -58,22 +57,15 @@ Python "venv" with all dependencies and commands installed into it.
...
-      $ cd edgedb-python
-      $ pip install -v -e .
-
 #. Build edgedb (the build will take a while):
 
    .. code-block:: bash
 
       $ cd ../edgedb
-      $ pip install -v -e ".[test,docs]"
+      $ pip install -v -e ".[test]"
```
1st1, line 65: Why removing `docs`?

1st1 (approved):

docs/internals/dev.rst
elprans: Because `test` implies `docs` in `setup.py`.

PR #2491, opened 2021-04-23
Skip simulation tests

fantix:
Now that we have the separate simulation test CI, these tests are not
needed in regular test runs.


tests/test_server_pool.py
```diff
@@ -374,6 +375,7 @@ def on_disconnect(self):
 
 class SimulatedCaseMeta(type):
     def __new__(mcls, name, bases, dct):
+        debug_pool = os.environ.get('EDGEDB_TEST_DEBUG_POOL')
```
elprans, line 378: Why not just make this check in `setUp`?

fantix: That makes a lot more sense! Fixed now.

elprans (approved):

PR #2492, opened 2021-04-24
Emit data catalog incompatibility to server status sink

elprans:
When `--emit-server-status` is given and the data directory is
determined to be incompatible with the current server, emit the
following message to the status sink:

  INCOMPATIBLE={"data_catalog_version":..., "expected_catalog_version":...}


fantix (approved):
LGTM

PR #2497, opened 2021-04-26
Switch build date and git commit in build version metadata

elprans:
Make build date come first to make sure newer builds are considered
newer versions.


fantix (approved):
This makes sense to me. 👍 

I'm not sure but is there any need to compare new type of versions starting with `+d` with old type `+g` (because for now`"+d" < "+g"`)?

elprans:
> I'm not sure but is there any need to compare new type of versions starting with

This is only relevant if all preceding parts of the version are equal, including `.devNNN`

PR #2498, opened 2021-04-26
Add an option to ignore other tenant in single-tenant mode

elprans:
To allow in-place upgrades, add the new `--ignore-other-postgres-tenant`
server flag, which will ask the server to ignore the existing tenant
with a different catalog version in single-tenant mode.

### DIFF
--- a/edb/server/args.py
+++ b/edb/server/args.py
@@ -61,8 +61,8 @@ class ServerConfig(NamedTuple):
     insecure: bool
     data_dir: pathlib.Path
     postgres_dsn: str
-    postgres_tenant_id: Optional[str]
-    ignore_other_postgres_tenant: bool
+    tenant_id: Optional[str]
+    ignore_other_tenants: bool
     log_level: str
     log_to: str
     bootstrap_only: bool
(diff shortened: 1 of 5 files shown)


1st1:
Maybe the server should always ignore other tenants? I'm worried that we're adding too much complexity for this prematurely.

elprans:
> Maybe the server should always ignore other tenants?

No, that'd result in much confusion, as in "where is my data??" when you accidentally run a newer release on an existing cluster.

edb/server/args.py
```diff
@@ -213,6 +214,13 @@ def _writer(status: str) -> None:
              ' Must be an alphanumeric ASCII string, maximum'
              f' {schema_defines.MAX_TENANT_ID_LENGTH} characters long.',
     ),
+    click.option(
+        '--ignore-other-postgres-tenant',
```
1st1, line 218:
```suggestion
        '--ignore-other-tenants',
```

elprans: Should I rename `--postgres-tenant-id` to `--tenant-id` as well?

1st1: Yes, i see no point in having 'postgres' in those arg names

elprans: Done.

1st1 (approved):

Issue #1385, opened 2020-05-08
Add support for existence conditionals to all top-level DDL commands.


elprans:
The `CREATE IF NOT EXISTS` and `DROP IF EXISTS` variants are useful.

We already have an ad-hoc `CREATE MODULE <...> IF NOT EXISTS`, and there's internal suppert for `ALTER IF EXISTS`.

1st1:
Yeah, we should support it for all schema objects.

1st1:
@elprans Is this 2.0 material?

PR #2500, opened 2021-04-27
Add `IF NOT EXISTS` clause to ROLE creations DDL.

vpetrovykh:
`CREATE ROLE` command benefits from the same `IF NOT EXISTS` clause
as `CREATE MODULE`.

Issue #1385

### DIFF
--- a/edb/edgeql/codegen.py
+++ b/edb/edgeql/codegen.py
@@ -877,10 +877,10 @@ class EdgeQLSourceGenerator(codegen.SourceGenerator):
             else:
                 self.write(ident_to_str(node.name.module), '::',
                            ident_to_str(node.name.name))
-        if node.create_if_not_exists and not self.sdlmode:
-            self.write(' IF NOT EXISTS')
         if after_name:
             after_name()
+        if node.create_if_not_exists and not self.sdlmode:
+            self.write(' IF NOT EXISTS')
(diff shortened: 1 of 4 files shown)


edb/edgeql/codegen.py
```diff
@@ -971,7 +971,11 @@ def visit_AlterDropInherit(self, node: qlast.AlterDropInherit) -> None:
             self.visit_list(node.bases)
 
     def visit_CreateDatabase(self, node: qlast.CreateDatabase) -> None:
-        self._visit_CreateObject(node, 'DATABASE')
+        def after_name() -> None:
```
elprans, line 974: This isn't relevant to this PR, I think.

elprans (approved):
Please fix the nit, otherwise LGTM

edb/edgeql/codegen.py
vpetrovykh: Right, but it's in general trying to account for `CREATE DATABASE foo FROM bar` syntax. This exists in the grammar and seems to be use din some testing (https://github.com/edgedb/edgedb/blob/29cec52cf3609389503ed21ef2911657f5f994a6/edb/testbase/server.py#L946-L947).

I can wrap that into a separate tiny codegen/syntax PR.

PR #2501, opened 2021-04-27
Properly escape underscores in a LIKE filter in _get_cached_reflection()

elprans:
Underscore has a special meaning of "any character" in LIKE patterns, so
the `WHERE` clause in `_get_cached_reflection()` is currently picking up
functions that aren't actual reflection helpers resulting in a server
crash in `introspect_db`.


fantix (approved):
👍

PR #2502, opened 2021-04-27
Server shutdown optimization

fantix:
In case of:

1. The EdgeDB process group received SIGINT (regular CTRL + C in terminal)
2. ~Compiler main process exited~
3. ~Compiler subprocess exited~ <- we'll restart the compiler worker process
4. ~Postgres exited~ <- we'll put the server in an error state in a different PR. For now, just mute the error.

This PR will make sure of the same exclusive execution path to stop the server.

* ~SIGINT/SIGTERM can be sent twice - first to initialize the shutdown,
  second (or more) will lead to sys.exit(1)~
* ~In the first 1000ms after the shutdown is initialized, SIGINT/SIGTERM
  will be ignored to protect the shutdown process from e.g. double kill~ <- doesn't make sense any more

Also fixed a wrong indication of server shutting down for systemd.

### DIFF
--- a/edb/server/main.py
+++ b/edb/server/main.py
@@ -215,6 +215,7 @@ def _run_server(
             )
 
         loop.add_signal_handler(signal.SIGTERM, terminate_server, ss, loop)
+        loop.add_signal_handler(signal.SIGINT, terminate_server, ss, loop)
 
         # Notify systemd that we've started up.
         _sd_notify('READY=1')
@@ -222,11 +223,9 @@ def _run_server(
         try:
             loop.run_forever()
         finally:
-            try:
-                logger.info('Shutting down.')
-                loop.run_until_complete(ss.stop())
-            finally:
-                _sd_notify('STOPPING=1')
+            _sd_notify('STOPPING=1')
+            logger.info('Shutting down.')
+            loop.run_until_complete(ss.stop())
 
 
 def run_server(args: srvargs.ServerConfig, *, do_setproctitle: bool=False):
--- a/edb/server/pgcon/pgcon.pyx
+++ b/edb/server/pgcon/pgcon.pyx
@@ -79,6 +79,10 @@ DEF COPY_SIGNATURE = b"PGCOPY\n\377\r\n\0"
 
 
 cdef object CARD_NO_RESULT = compiler.ResultCardinality.NO_RESULT
+cdef object POSTGRES_SHUTDOWN_ERR_CODES = {
+    '57P01': 'admin_shutdown',
+    '57P02': 'crash_shutdown',
+}
 
 
 cdef bytes INIT_CON_SCRIPT = None
@@ -1635,6 +1639,14 @@ cdef class PGConnection:
         while self.buffer.take_message():
             if not self.parse_notification():
                 mtype = self.buffer.get_message_type()
+                if mtype == b'E':  # ErrorResponse
+                    er_cls, fields = self.parse_error_message()
+                    msg = POSTGRES_SHUTDOWN_ERR_CODES.get(fields['C'])
+                    if msg:
+                        msg = fields.get('M', msg)
+                        # TODO: handle PostgreSQL shutdown here
+                        # For now, we just silently skip this message
+                        continue
                 raise RuntimeError(
                     f'unexpected message type {chr(mtype)!r} in IDLE state')
 


tailhook (commented):
Apart from tiny comments. I think the core problem is in compiler pool:
https://github.com/edgedb/edgedb/blob/9bfea9fbcaef580bac5222837cd412b1c6bcc400/edb/server/compiler_pool/worker.py#L386

We fork processes here, but neither wait them to finish nor even reap stopped processes I think. Looks like we have to turn "parent compiler" into a full-blown process manager here.

Note: in addition to compiler processes dying on shutdown it's also possible that they are dying killed by OOM and in future we will be killing them on timeout. So we probably want to make parent process a pristine copy of a compiler to fork new processes from, not as equally enaged worker.

fantix:
> Apart from tiny comments. I think the core problem is in compiler pool:
> https://github.com/edgedb/edgedb/blob/9bfea9fbcaef580bac5222837cd412b1c6bcc400/edb/server/compiler_pool/worker.py#L386
> 
> We fork processes here, but neither wait them to finish nor even reap stopped processes I think. Looks like we have to turn "parent compiler" into a full-blown process manager here.
> 
> Note: in addition to compiler processes dying on shutdown it's also possible that they are dying killed by OOM and in future we will be killing them on timeout. So we probably want to make parent process a pristine copy of a compiler to fork new processes from, not as equally enaged worker.

@tailhook Thanks for the review! Yes, I kinda agree with you that we probably need a process manager. My question here is - maybe for @1st1 too - why we're calling `fork()` 4 times in the subprocess assuming we have 4 cores, instead of calling `asyncio.create_subprocess_exec(sys.xecutable, '-m', WORKER_MOD, ...)` 4 times, so that the `compiler_pool.pool.Pool` will become the process manager in the main EdgeDB process? Is it to leverage shared memory for low-end machines?

I'm okay with either `fork() + exec*() + 4 x fork()` or `4 x (fork() + exec*())`, but the proposed design here is as follows, please kindly review (@1st1 too):
* Implement a proper process manager that e.g. takes care of SIGCHLD with `waitpid()` etc.
* Be able to start / stop processes at runtime
* Restart workers that natually died from - I don't know - core dumps?
* Handle OOM properly (assume we don't have a memory leakage - we keep running with the remaining workers until the last one dies from OOM, then we tell the user we're out to compile new queries; or the main process OOMs and we skip the trouble and bail out?)
* Handle timeout properly (kill and restart the slow process that didn't respond in a certain amount of time, with `SIGTERM` followed by `SIGKILL` after a while?)

tailhook:
> @fantix
> 
> > Implement a proper process manager that e.g. takes care of SIGCHLD with waitpid() etc.
> 
> I'm not sure why. The current mechanism is using a UNIX socket - if it breaks for any reason the child process is deemed to be dead. We can follow up with a kill(childpid) if we want. Same for the server - if the server suddenly dies, the child compiler process detects the broken pipe and instantly exits.
> 
> I don't want us to implement sigchld and waitpid() because they are they own can of worms and needed when you want a generic mechanism. In our case, we don't need generic, we have our own solution.
> 

You're missing the point. The problem isn't exactly SIGCHLD vs tracking sockets. The problem is that whenever compiler dies, somebody need to restart it. And that restart has to be fork from some other already loaded compiler (otherwise this optimization will just go away with time).

> > Compiler processes are stateless and mostly synchronous (or should be soon). So we can just to with KILL immediately (when we detect that compilation is timed out). Otherwise we just introduce bugs (by interrupting things that aren't supposed to be interrupted, like "finally" statements). SIGTERM that doesn't have a handler would work too, though (just so we can log the difference between KILL by OOM, and own process killer).
> 
> No. Starting them is very slow, can be a matter of seconds.
> 
> My plan is to make them use synchronous IO (they don't need to be async anymore) and just use a mechanism similar to KeyboardInterrupt. When they use blocking IO there's no risk of wrecking anything.

This is very dangerous thing. There are at least three issues with this:
1. To make sure that state is in sync we have to close the connection to server (which by assumption above means worker is dead for the supervisor) and reconnect. But this is minor and can be fixed.
2. Any `finally` and `__exit__` clauses could be interrupted (hi [PEP-419](https://www.python.org/dev/peps/pep-0419/)). This may include failing to close the socket from (1), context vars (unlikely as implemented in C but any abstractions on top can fail), and a lot of other things. This can also lead to memory leaks.
3. Non-python code is not interrupted in this way, so moving more code into Rust will make it less responsive.

All of this can be avoided by super-careful coding. Avoiding important code in finally, and any third-party finally/context managers. Making "dirty" flag on connection. Then we have to mask out signal so that it doesn't interrupt the main loop itself, or any our cache cleaning routines. I don't think it worth it

Timeouting compilation should not be a common case. And the time to restart of the worker should be same order of magnitude that the compilation timeout itself. (i.e. it's unlikely that we will set compilation timeout of 100ms, or something). So trying to recover compiler will not save for DoS by malicious connection pushing huge statements to the compiler: the time that compilers tries to process these queries is the issue not the time to restart compiler on a timeout.

So this is very similar to our issue with, whether to close postgres connection or try to recover. I think we should just kill the worker for now and work towards optimizing it (including the startup time) in Rust.

tailhook:
> Thanks for all the comments! Here's my plan:

Plan is good. Just to make sure that everybody is on the same page: It looks like issue of `systemctl stop` (and so nightly tests on CLI) will not fix until (3) (or maybe (2), but I'm not sure), will be done. Although we can engage timeout and kill signal. Which should be fine for dev instances.

elprans:
To clarify, we use `fork()` to share Python code objects and, most importantly, the parser tables that take ages to unpickle.  I agree with @tailhook that killing a misbehaving compiler is a most reliable course of action, because spawning a new process is cheap if we continue to use the fork server approach.

1st1:
> This is very dangerous thing. There are at least three issues with this:

The dangerous thing is to continue using asyncio where it shouldn't be used.

> To make sure that state is in sync we have to close the connection to server (which by assumption above means worker is dead for the supervisor) and reconnect. But this is minor and can be fixed.

I'm not sure what this is about. Please clarify.

> Any finally and __exit__ clauses could be interrupted (hi PEP-419). This may include failing to close the socket from (1), context vars (unlikely as implemented in C but any abstractions on top can fail), and a lot of other things. This can also lead to memory leaks.

This is unrelated to our current architecture. At least I don't see how. Please clarify.

> Non-python code is not interrupted in this way, so moving more code into Rust will make it less responsive.

Rust code in the compiler process is going to be synchronous, most likely. In any case I'm against supporting hypothetical scenarios right now. Years might pass until we have enough Rust in our compiler infra to justify async there.

1st1:
>  I agree with @tailhook that killing a misbehaving compiler is a most reliable course of action

I think no one was disputing that.

tailhook:
> My plan is to make them use synchronous IO (they don't need to be async anymore) and just use a mechanism similar to KeyboardInterrupt. When they use blocking IO there's no risk of wrecking anything.

@1st1 I'm referring to this. It looks like you're suggesting to use signals (mechanism similar to KeyboardInterrupt) to enforce timeouts on compiler. Which that whole post was about.

I'm all for getting rid of async from compiler.

fantix:
> It looks like issue of `systemctl stop` (and so nightly tests on CLI) will not fix until (3) (or maybe (2), but I'm not sure), will be done. Although we can engage timeout and kill signal. Which should be fine for dev instances.

Right, this PR is now not fixing much, only leaving 2 stubs (postgres exit message and SIGINT for future fast shutdown) for the next steps. Looking at the CLI nightly test systemd logs, I feel like there are likely more than one issue to fix for shutdown to handle all cases properly, so please bare with me and keep the nightly as it is now.

PR #2504, opened 2021-04-27
Add `schema::Object.name_parts` property

elprans:
Currently there is no convenient way to get structured object name in
the introspection schema, because `schema::Object.name` is a string.
Fix this by adding `Object.name_parts` which is an array of strings
representing the parts of the name: module and unqualified name.

### DIFF
--- a/edb/lib/schema.edgeql
+++ b/edb/lib/schema.edgeql
@@ -44,6 +44,7 @@ CREATE SCALAR TYPE schema::TypeModifier
 # Base type for all schema entities.
 CREATE ABSTRACT TYPE schema::Object EXTENDING std::BaseObject {
     CREATE REQUIRED PROPERTY name -> std::str;
+    CREATE REQUIRED PROPERTY name_parts -> array<std::str>;
     CREATE REQUIRED PROPERTY internal -> std::bool {
         SET default := false;
     };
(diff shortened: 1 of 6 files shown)


1st1:
What's the motivation for adding this? Why not just split name by '::' in edgeql? How will you use this exactly?

elprans:
> What's the motivation for adding this?

We went to considerable lengths to stop mixing names with strings internally, this is a continuation of that principle.  I actually think that `.name` should be an array, but that's a bit too radical at this point.

The sequences PR is using this to dump sequence names properly with appropriate quoting.  I also considered adding a `quoted_name` computable, but our `quote_ident` function isn't up to snuff.  Considering that introspection would primarily be used for various builders, quoting safety is important.  

Another common use case is to filter by module name.

> Why not just split name by '::' in edgeql

Well, it's fairly inconvenient, especially if all you want is to get a properly quoted identifier.  And, although we ban `::` even in quoted identifiers, blindly splitting by `::` somehow still feels unsafe, especially if we consider possible future support for nested modules.

1st1:
Since the PR lacks the docs, I'm not sure how will this work when we have nested namespaces. Will `foo::bar::Spam` be represented as `['foo', 'bar', 'Spam']` or as `['foo::bar', 'Spam']`?

Basically my worry is that everybody starts hardcoding `name_parts[0]` and `name_parts[1]`, which would break badly when we introduce nested namespaces.

Also, why does this have to be an array? Why not a namedtuple?

`name_parts -> tuple<mod: array<str>, name: str>`

elprans:
> Also, why does this have to be an array? Why not a namedtuple?

Good point.  Alright, I'll make it internal until we figure the proper API out.

Issue #2441, opened 2021-04-07
Restoring a sequence type from a dump resets the counter.


vpetrovykh:
- EdgeDB Version: EdgeDB 1.0-beta.2+dev.5562.g64ae7622e
- OS Version: Linux

Steps to Reproduce:

1. Create a database with the schema
    ```
    scalar type MySeq extending sequence;
    type Foo {
        required property num -> MySeq;
    };
    ```
2. Insert a few objects: `FOR x IN {1, 2, 3, 4} UNION (INSERT Foo);`
3. Dump the database.
4. Drop/create the database (or in any other way prepare a blank database to restore to).
5. Restore the database.
6. Run `INSERT Foo`.
7. Observe that the `MySeq` sequence was reset:
    ```
    db> select Foo.num;
    {1, 2, 3, 4, 1}
    ```

PR #2508, opened 2021-04-28
Add functions to manipulate sequence state and preserve sequence state in dumps

elprans:
### Implement functions to explicitly advance or reset a sequence value

The new `std::sequence_next(schema::ScalarType)` function
returns the next value for the specified sequence type.

The new `std::sequence_reset(schema::ScalarType, std::int64)`
function sets the _current_ value of the specified sequence, the
next call to `sequence_next` will return the next value in sequence.

The new `std::sequence_reset(schema::ScalarType)`
function resets the specified sequence to its initial state, so
the next call to `sequence_next` will return the start value of the
sequence.

### Make sure sequence state gets included in dumps

The current value of a sequence type is a piece of data that must be
preserved in the database dump.

Fixes: #2441

### DIFF
--- a/docs/edgeql/funcops/index.rst
+++ b/docs/edgeql/funcops/index.rst
@@ -65,6 +65,7 @@ singletons. Examples of aggregate functions include built-ins such as
     json
     bytes
     math
+    sequence
     sys
     uuid
     deprecated
--- a/docs/edgeql/funcops/sequence.rst
+++ b/docs/edgeql/funcops/sequence.rst
@@ -1,0 +1,103 @@
+.. _ref_eql_funcops_sequence:
+
+========
+Sequence
+========
+
+:edb-alt-title: Sequence Manipulation Functions
+
+
+.. list-table::
+    :class: funcoptable
+
+    * - :eql:func:`sequence_next`
+      - :eql:func-desc:`sequence_next`
+
+    * - :eql:func:`sequence_reset`
+      - :eql:func-desc:`sequence_reset`
+
+
+---------
+
+
+.. eql:function:: std::sequence_next(seq: schema::ScalarType) -> int64
+
+    Advance the sequence to its next value and return that value.
+
+    Sequence advancement is done atomically, each concurrent session and
+    transaction will receive a distinct sequence value.
+
+    .. code-block:: edgeql-repl
+
+       db> SELECT sequence_next(INTROSPECT MySequence);
+       {11}
+
+
+---------
+
+
+.. eql:function:: std::sequence_reset(seq: schema::ScalarType) -> int64
+                  std::sequence_reset( \
+                    seq: schema::ScalarType, val: int64) -> int64
+
+    Reset the sequence to its initial state or the specified value.
+
+    The single-parameter form resets the sequence to its initial state, where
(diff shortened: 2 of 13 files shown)


1st1:
> The new std::sequence_set_cur_value(schema::ScalarType, std::int64)
function sets the current value of the specified sequence.

Given that we have `datetime_current()` I wouldn't use the `cur` here and just spell it out in full. The name is already long. That said, I would omit "value" -- it's not needed here -- so it would be:

`std::sequence_get_next()`
`std::sequence_set_next()`
`std::sequence_set_current()`

But then I don't like the `get_next()` function because it's not clear if it's advancing the internal sequence counter or not. So I'd call it `std::sequence_advance()` or just `std::sequence_next()`

Lastly, do we really need `std::sequence_set_current()` *and* `std::sequence_set_next()`? Can we have just one function?

elprans:
> Lastly, do we really need `std::sequence_set_current()` _and_ `std::sequence_set_next()`? Can we have just one function?

Then it would have to be `set_current()`, because that is needed in dumps.  Postgres has both variants, albeit as an argument to `nextval()`, and `set_next()` semantics is the default behavior.

> call it `std::sequence_advance()` just `std::sequence_next()`

I considered these.  `advance()` doesn't make it clear that it returns the value and isn't very discoverable, and `next()` is a bit too close to `set_next()`, but I can live with that, I think.

1st1:
So I'd go with two for now: `std::sequence_next()` and `std::sequence_set_current()`.

elprans:
Actually, the use case for `set_next()` is to be able to reset the sequence back to the initial value.  If we only have `set_current()`, there will be no way to generate a value of `1`.

1st1:
I'd be fine with `std::sequence_next()`, `std::sequence_set_current()`, and `std::sequence_set_next()`.

1st1 (commented):
Can you please add some docs too? They don't have to be elaborate at this point.

elprans:
Renamed.

1st1 (approved):
LGTM, but please add some docs!

elprans:
Docs added.

msullivan (approved):
Seems good, though the current/next distinction is kind of a shame

tailhook:
> Actually, the use case for set_next() is to be able to reset the sequence back to the initial value. If we only have set_current(), there will be no way to generate a value of 1.

Can we have `sequence_set_current(type, val)` and `sequence_reset(type)` instead ?

elprans:
> Can we have `sequence_set_current(type, val)` and `sequence_reset(type)` instead ?

We could, but I'm concerned we could miss some other important use case for setting the next value of the sequence exactly.

Another option is to do what Postgres does and use a single function with a kwarg:

```
sequence_set(type, val, NAMED ONLY used)
```

Where `used` would determine if `val` should be considered the current or next value.  I think multiple functions is clearer, though.

1st1:
We've decided to have one function with two variants:

* std::sequence_reset(type) -- reset to initial value
* std::sequence_reset(type, value) -- reset the current value to the supplied one

elprans:
Changed to `sequence_reset`

PR #2509, opened 2021-04-28
Fix migration prompting while adding required pointers via parent types

### LINKED ISSUE #2486: Ask for default when adding new required property
- EdgeDB Version: 1-beta.1
- EdgeDB CLI Version: beta.1
- OS Version: macOS Big Sur

Steps to Reproduce:

I made a minimal repro of how I got my schema into an inconsistent state.

1. Create an initial schema

```
module default {
    type BlogPost {
        property title -> str;
    }
};
```

2. Add some data;

```
INSERT BlogPost;
```

3. Update
```
module default {
    type BlogPost {
        property title -> str;
    }
};
```

4. Run `create-migration` (works fine). If you make an existing non-required property required, the migration tool prompts you for a default, but not if you define a new required property.

5. Inconsistent state:

Running migrate:

```
$ edgedb -I inconsistent migrate
edgedb error: ERROR: MissingRequiredError: missing value for required property 'name' of object type 'default::BlogPost'
```

Runing create-migration (e.g. after setting a default for .content):

```
$ edgedb -I inconsistent create-migration
edgedb error: Database must be updated to the last migration on the filesystem for `create-migration`. Run:
  edgedb migrate
```

msullivan:
The key idea is that we force a `CREATE PROPERTY ... { SET REQUIRED  ... }`
to be generated when appropriate even if we would normally elide it.

Forcing this to happen exposed a funny corner case in which the
diffing mechanism sometimes generates DELETE/CREATE pairs for the
__type__ link when a type is renamed. This might be worth dealing with
more robustly, but I worked around this for now by having the compiler
regenerate the proposed plan after applying DDL that does a rename.
This turns out to have been a good idea in general, since it fixed
issues where we would generate prompts that referred to old names
instead of new ones.

Fixes #2486.

### DIFF
--- a/edb/schema/delta.py
+++ b/edb/schema/delta.py
@@ -2163,6 +2163,7 @@ class ObjectCommand(Command, Generic[so.Object_T]):
                         # and that have their value actually changed.
                         not fop.new_inherited
                         or context.descriptive_mode
+                        or self.ast_ignore_ownership()
                     )
                     and (
                         fop.old_value != new_value
@@ -2604,6 +2605,10 @@ class ObjectCommand(Command, Generic[so.Object_T]):
         if self.annotations is None:
             self.annotations = {}
         self.annotations[name] = value
+
+    def ast_ignore_ownership(self) -> bool:
+        """Whether to force something into the AST even though it is owned"""
+        return False
 
 
 class ObjectCommandContext(CommandContextToken[ObjectCommand[so.Object_T]]):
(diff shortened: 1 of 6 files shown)


edb/schema/delta.py
```diff
@@ -2605,6 +2606,10 @@ def set_annotation(self, name: str, value: Any) -> None:
             self.annotations = {}
         self.annotations[name] = value
 
+    def ast_ignore_ownership(self) -> bool:
+        """Whether to force something into the AST even though it is owned"""
```
elprans, line 2610: Did you mean "not owned" here?

elprans (approved):

PR #2510, opened 2021-04-28
Make DROP MODULE actually drop module contents

### LINKED ISSUE #1989: DROP MODULE doesn't delete the module contents
```
x3> create module test; create type test::foo;
OK: CREATE
OK: CREATE
x3> drop module test;
OK: DROP
x3> describe schema;
{
  'CREATE MODULE default IF NOT EXISTS;
CREATE TYPE test::foo;',
}
x3> create module test; create type test::foo;
OK: CREATE
ERROR: SchemaError: ObjectType <QualName test::foo> is already present in the schema <FlatSchema gen:1 at 0x7f738eabdcd0>
```

The objects are inaccessible while the module is gone, but will appear in `describe schema` and in introspection tables and if the module is recreated, all the objects in it are already there.

msullivan:
This is done in the most heavy handed way possible: by emulating
schema migrations. Deleting lots of things at once in our sytem is
kind of hard, and we've put a lot of effort into making it work with
schema migrations.

Fixes #1989.

### DIFF
--- a/edb/schema/modules.py
+++ b/edb/schema/modules.py
@@ -88,7 +88,7 @@ class DeleteModule(ModuleCommand, sd.DeleteObject[Module]):
         # Modules aren't actually stored with any direct linkage
         # to the objects in them, so explicitly search for objects
         # in the module (excluding the module itself).
-        has_objects = bool(any(schema.get_objects(
+        has_objects = bool(list(schema.get_objects(
             included_modules=[self.classname],
             excluded_items=[self.classname],
         )))
--- a/edb/schema/modules.py
+++ b/edb/schema/modules.py
@@ -29,7 +29,6 @@ from edb.edgeql import qltypes
 from . import annos as s_anno
 from . import delta as sd
 from . import schema as s_schema
-from . import objects as so
 
 
 class Module(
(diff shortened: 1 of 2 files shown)


elprans:
I don't think recursive deletion should be the default behavior, it's too dangerous.  Instead, can we prohibit deletion of a module containing objects and, perhaps, add a syntactic switch to enable deletion of contained objects?  I.e. `rmdir` vs `rm -r` or `DROP SCHEMA` vs `DROP SCHEMA CASCADE` in Postgres.

elprans (approved):

edb/schema/modules.py
```diff
@@ -73,3 +75,26 @@ class AlterModule(ModuleCommand, sd.AlterObject[Module]):
...
+        context: sd.CommandContext,
+    ) -> None:
+        super()._validate_legal_command(schema, context)
+
+        # For now, we disallow deleting non-empty modules.
+
+        # Modules aren't actually stored with any direct linkage
+        # to the objects in them, so explicitly search for objects
+        # in the module (excluding the module itself).
+        has_objects = bool(any(schema.get_objects(
```
elprans, line 91: I don't think you need `any` here

msullivan: No, you do: it returns an iterator, and `bool` on an iterator is `True`. There's no need to be clever here, though; I'll just make it `list`

elprans: Ah, right. Let's leave any in then. Constructing a list seems wasteful, since there could be lots of objects in a module

PR #2516, opened 2021-04-29
Support switching to binary protocol with HTTP upgrade

fantix:
The server now follows HTTP/1.1 upgrade protocol to promote a single
frontend connection from HTTP-based protocol to EdgeDB binary protocol.
The protocol is named `edgedb-binary` and it has to be exactly the same
in the Upgrade header, e.g. `Upgrade: edgedb-binary`.

Previous mechanism to detect if a new connection is HTTP or binary by
looking at the first byte still remains, so old clients are still
supported. But new clients are suggested to use HTTP upgrade on
connect in order to switch to the binary protocol (or remain in HTTP).

### DIFF
--- a/edb/server/main.py
+++ b/edb/server/main.py
@@ -196,6 +196,7 @@ def _run_server(
         echo_runtime_info=args.echo_runtime_info,
         status_sink=args.status_sink,
         startup_script=args.startup_script,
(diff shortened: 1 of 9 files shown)


fantix:
Closed for #2610.

PR #2518, opened 2021-04-29
Add the _datetime_range_buckets function

elprans:
`_datetime_range_buckets` generates a set of datetime pairs representing
buckets within a given datetime range.  The function isn't public,
because a proper public implementation will need range types and
`cal::relativedelta`, neither of which are implemented.


fmoor (approved):

1st1 (commented):
Nice. We need this in the stdlib indeed.

Issue #2522, opened 2021-05-04
ISE when running `describe schema as sdl` if an extension is enabled


jaclarke:
- EdgeDB Version: 1.0-beta.2+dev5631.d2021050409.gec3bd1e8f.cv202104290000

@colinhacks reported in slack that `describe schema as sdl` fails on the tutorial schema. It seems that it's specifically having the notebook extension (or any extension) enabled that causes the error.

Steps to Reproduce:
```
edgedb> describe schema as sdl;
{'module default{};'}
edgedb> create extension graphql;
OK: CREATE EXTENSION
edgedb> describe schema as sdl;
ERROR: InternalServerError: ''
  Hint: This is most likely a bug in EdgeDB. Please consider opening an issue ticket at https://github.com/edgedb/edgedb/issues/new?template=bug_report.md
  Server traceback:
      Traceback (most recent call last):
        File "/home/james/dev/edgedb/edb/server/compiler_pool/worker.py", line 312, in worker
          res = meth(*args)
        File "/home/james/dev/edgedb/edb/server/compiler_pool/worker.py", line 173, in compile
          units, cstate = COMPILER.compile(
        File "/home/james/dev/edgedb/edb/server/compiler/compiler.py", line 1986, in compile
          units = self._compile(ctx=ctx, source=source)
        File "/home/james/dev/edgedb/edb/server/compiler/compiler.py", line 1565, in _compile
          return self._try_compile(ctx=ctx, source=source)
        File "/home/james/dev/edgedb/edb/server/compiler/compiler.py", line 1618, in _try_compile
          comp, capabilities = self._compile_dispatch_ql(ctx, stmt)
        File "/home/james/dev/edgedb/edb/server/compiler/compiler.py", line 1539, in _compile_dispatch_ql
          query = self._compile_ql_query(ctx, ql)
        File "/home/james/dev/edgedb/edb/server/compiler/compiler.py", line 563, in _compile_ql_query
          ir = qlcompiler.compile_ast_to_ir(
        File "/home/james/dev/edgedb/edb/edgeql/compiler/__init__.py", line 177, in wrapper
          return func(*args, **kwargs)
        File "/home/james/dev/edgedb/edb/edgeql/compiler/__init__.py", line 225, in compile_ast_to_ir
          ir_set = dispatch_mod.compile(tree, ctx=ctx)
        File "/usr/lib/python3.9/functools.py", line 877, in wrapper
          return dispatch(args[0].__class__)(*args, **kw)
[...]

PR #2523, opened 2021-05-04
Fix DESCRIBE SCHEMA AS SDL in presence of unqualified objects

elprans:
Modules aren't the only unqualified objects in the schema, there are
also extensions.

Fixes: #2522


vpetrovykh (approved):
LGTM

PR #2524, opened 2021-05-05
Fix two related issues with UPDATEs

msullivan:
* Execute nested DML once for each UPDATEd object in all cases.
   (Previously it worked only when the relevant links were multi)
 * Execute nested DML on *all* subtypes, not just the first processed

The core of the solution to both issues is to create an IteratorCTE
based on the range CTE.

### DIFF
--- a/edb/pgsql/compiler/dml.py
+++ b/edb/pgsql/compiler/dml.py
@@ -144,7 +144,7 @@ def init_dml_stmt(
     ):
         dml_cte = pgast.CommonTableExpr(
             query=pgast.SelectStmt(),
-            name=ctx.env.aliases.get(hint='m')
+            name=ctx.env.aliases.get(hint='melse')
         )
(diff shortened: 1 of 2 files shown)


edb/pgsql/compiler/dml.py
```diff
@@ -144,7 +144,7 @@ def init_dml_stmt(
     ):
         dml_cte = pgast.CommonTableExpr(
             query=pgast.SelectStmt(),
-            name=ctx.env.aliases.get(hint='m')
+            name=ctx.env.aliases.get(hint='melse')
```
elprans, line 147: Is this hunk intentional?

elprans (approved):

edb/pgsql/compiler/dml.py
msullivan: Yeah, I was trying to make sure I understood where the `m`s were coming from

PR #2526, opened 2021-05-05
Prevent illegal UPDATEs to covariantly overloaded link types

### LINKED ISSUE #2324: Prevent illegal UPDATEs to covariantly overloaded link types
If we have a schema like
```
type Tgt;
type SubTgt extending Tgt {
    required property foo -> str;
};

type Foo {
    property name -> str;
    required link x -> Tgt;
};
type Bar extending Foo {
    overloaded required link x -> SubTgt;
};
```

we can do `UPDATE Foo FILTER .name = "whatever" SET { x := (INSERT Tgt) }`
and we will wind up in an illegal state.

This is because we allow covariant overloading of pointer types but that is unsound since they are mutable.

Options:
1. Statically disallow UPDATEs that SET pointers that have their types overridden in subclasses.
2. Disallow these illegal UPDATEs with run-time checks.

My gut reaction is prefer static checks, and I think most cases wouldn't be inconvenienced. But @elprans argued for dynamic checks, suggesting the case of an API endpoint where "you only get an id of an object to update and an id of a new link target", and I find this reasonably compelling.

There will be a runtime cost, but it shouldn't be that bad, and we only need to pay the cost in situations that would be prohibited if we went the static route.

msullivan:
Insert dynamic type checks in situations where it can occur.

One slightly fiddly part is making sure that we can extract the
`__type__` from the places we need it (since doing this check can
require `__type__` in places where it otherwise wouldn't appear anywhere
in the AST). This is solved with a bit of a hack: UpdateStmt is
augmented with a PointerRef to `BaseObject.__type__`, from which we can
extract whatever we need.

Also fixes creation of covariantly overloaded multi links.

Fixes #2324.

### DIFF
--- a/edb/pgsql/compiler/dml.py
+++ b/edb/pgsql/compiler/dml.py
@@ -917,7 +917,7 @@ def compile_insert_shape_element(
 
 def process_update_body(
     *,
-    ir_stmt: irast.MutatingStmt,
+    ir_stmt: irast.UpdateStmt,
     update_cte: pgast.CommonTableExpr,
     dml_parts: DMLParts,
     typeref: irast.TypeRef,
@@ -1000,6 +1000,19 @@ def process_update_body(
                         else:
                             val = dispatch.compile(updvalue, ctx=scopectx)
 
+                        assert isinstance(updvalue, irast.Stmt)
+                        val = check_update_type(
+                            val,
+                            val,
+                            is_subquery=True,
+                            ir_stmt=ir_stmt,
+                            ir_set=updvalue.result,
+                            subject_typeref=typeref,
+                            shape_ptrref=ptrref,
+                            actual_ptrref=actual_ptrref,
+                            ctx=scopectx,
+                        )
+
                         val = pgast.TypeCast(
                             arg=val,
                             type_name=pgast.TypeName(name=ptr_info.column_type)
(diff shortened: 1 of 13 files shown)


tests/schemas/updates.edgeql
```diff
@@ -25,6 +25,14 @@ INSERT test::Status {
     name := 'Closed'
 };
 
+INSERT test::MajorLifeEvent {
+    name := 'Broke a Type System'
```
elprans, line 29: :rofl:

elprans (approved):

edb/pgsql/compiler/dml.py
```diff
@@ -1068,6 +1081,104 @@ def process_update_body(
...
+    base_ptrref = irtyputils.find_actual_ptrref(
+        ir_stmt.subject.typeref, shape_ptrref)
+    # We skip the check if either the base type matches exactly
+    # or the shape type matches exactly. FIXME: *Really* we want to do
+    # a subtype check, here, though, since this could do a needless
+    # check if we have multiple levels of overloading, but we don't
+    # have the infrastructure here.
+    if (
+        not irtyputils.is_object(ir_set.typeref)
+        or base_ptrref.out_target == actual_ptrref.out_target
```
elprans, line 1113: Perhaps check `.id` equality instead?  Should be more reliable.

PR #2527, opened 2021-05-06
Update the CLI documentation.

vpetrovykh:
I've updated much of the CLI documentaiton, but there's still the `server` group that I need to cover. I'll do that as a separate PR.


1st1:
Victor, please update the `cli/index.rst`. I'd drop the table because we'll have it all displayed in the TOC anyways. Need to add installation instructions (and copy the bullet list from https://www.edgedb.com/install).

You can also add this paragraph:

> EdgeDB includes the all-encompassing edgedb command-line tool. It provides an idiomatic way to do just about everything: install EdgeDB, spin up a local instance, open a REPL, execute queries, manage auth roles, introspect a database schema, create migrations, and more. Install it with one shell command.

vpetrovykh:
Update the PR to move install/uninstall instructions to the overview blurb. Updated the wording of the overview as well.

PR #2528, opened 2021-05-06
Update the documentaiton of server CLI commands.

vpetrovykh:


vpetrovykh:
There are a handful of things I left out:
* `--help` or `help` at most of the levels, because it's not terribly informative and it exists at the `edgedb` level of the docs.
* `edgedb revert` largely because I don't really understand what "Revert a major instance upgrade" actually entails and what scenario it covers. Minimalist information here might be more misleading than helpful, though.
* `--system` and `--start-conf` options in `edgedb server init`, because I have no idea what they are and the help is not clarifying that.

fantix:
> `--start-conf`

I happened to run into this one - I could help adding something later today.

docs/cli/edgedb_server_init.rst
```diff
@@ -0,0 +1,63 @@
...
+    ``edgedb``.
+
+:cli:synopsis:`--default-user=<default-user>`
+    Specifies the default user name (created during initialization,
+    and saved in credentials file). Defaults to: ``edgedb``.
+
+:cli:synopsis:`--port=<port>`
+    Specifies which port should the instance be configured on. By
+    default a random port will be used and recorded in the credentials
+    file.
```
fantix, line 63:
```suggestion
    file.

:cli:synopsis:`--start-conf=<start-conf>`
    Configures how the new instance should start: ``auto`` (default)
    for automatic start with the system / user session, ``manual`` to
    turn that off so that the instance can be manually started on
    demand.
```

vpetrovykh: Thanks, I've reworded this slightly and added it to the docs.

docs/cli/edgedb_server_destroy.rst
```diff
@@ -0,0 +1,33 @@
+.. _ref_cli_edgedb_server_destroy:
+
+
+=====================
+edgedb server destroy
+=====================
+
+Stop an EdgeDB instance.
```
tailhook, line 8: It doesn't just stop. It removes all the data.

docs/cli/edgedb_server_destroy.rst
```diff
@@ -0,0 +1,33 @@
...
+.. cli:synopsis::
+
+     edgedb server destroy [OPTIONS] <name>
+
+
+Description
+===========
+
+``edgedb server destroy`` is a terminal command for destroying an
+EdgeDB instance. The instance must be :ref:`stopped
```
tailhook, line 19: Instance doesn't have to be stopped first.

tailhook (approved):

tailhook:
> edgedb revert largely because I don't really understand what "Revert a major instance upgrade" actually entails and what scenario it covers. Minimalist information here might be more misleading than helpful, though.

When you do upgrade, which involves dump/restore, we keep old database directory around. Revert removes the new database directory and moves old one in place, then modifies service file to use previous version of the EdgeDB.

> --system

Doesn't work. We didn't decide what to do with it. Probably have to remove or hide it from the help output too.

> and --start-conf options in edgedb server init, because I have no idea what they are and the help is not clarifying that.

`--start-conf=manual` prevents from starting EdgeDB on init and later when system is reboot (or user is logged in). By default (`--start-conf=auto`) database is expected to be running at all times (managed by systemd/launchctl), which is not what you want if you have multiple instances on a laptop.

vpetrovykh:
Thanks! I've updated the docs.

docs/cli/edgedb_server_destroy.rst
vpetrovykh: Fixed

docs/cli/edgedb_server_destroy.rst
vpetrovykh: Fixed

PR #2530, opened 2021-05-06
Apply volatility refs for collection computables

msullivan:


elprans (approved):

msullivan:
A mistake that I make a *lot* is that I fix the last lint error, and because it was something super simple (like an extra line), I don't bother to run lint again to check. But I forget that lint isn't run on tests unless it passes on edb, so I miss a test lint error.

PR #2543, opened 2021-05-12
Make USING expressions propagate to subtypes

msullivan:
USING expressions for SET REQUIRED/SET TYPE/similar don't get applied
to subtypes properly.

Fix this by cloning `AlterSpecialObjectField` objects along with
their USING expressions, and modifying
`ObjectCommand.set_attribute_value` to use an existing
`AlterSpecialObjectField` even if it doesn't have an actual value
in it yet.


msullivan:
(reported by @colinhacks)

elprans (approved):

PR #2545, opened 2021-05-13
Inject volatility_ref on object types also

msullivan:
This ensures that nested volatile computables get executed the right
number of times. Since the calls to get_path_var are only done when
there is actually a volatile, it shouldn't degrade code quality in
other cases.

### DIFF
--- a/tests/schemas/volatility_setup.edgeql
+++ b/tests/schemas/volatility_setup.edgeql
@@ -59,7 +59,15 @@ CREATE FUNCTION test::err_volatile() -> float64 {
     $$;
 };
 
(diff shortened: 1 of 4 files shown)


tests/schemas/volatility_setup.edgeql
```diff
@@ -59,7 +59,15 @@ CREATE FUNCTION test::err_volatile() -> float64 {
     $$;
 };
 
+CREATE FUNCTION test::rand_int(top: int64) -> int64 {
```
elprans, line 62: nit: this seems to be irrelevant to this PR

elprans (approved):

PR #2548, opened 2021-05-18
Compiler pool: restart dead workers

fantix:
* Promoted the first worker into a Queen (talking to the "Hive" in the same protocol for requests to spawn workers)
* Also removed async code from workers
* Fixed zombie worker issue by daemonizing
* The current design avoids using waitpid() on worker processes

(Before this PR, we were expecting the worker processes to exit on a broken UNIX socket. But in cases like OOM the worker process would die in a zombie because we didn't (want to?) `waitpid()` on it. This PR simply used two forks to avoid the zombie issue, and keep `waitpid()` away.)

The current design has 2 issues:
* It doesn't survive edgedb/edgedb-cli#349 because even though the (user) systemd process took over the daemonized worker processes, they are still considered a part of the service and lingered longer than the main process, thus triggering the CentOS bug [0017461](https://bugs.centos.org/view.php?id=17461#c38441).
* The double-fork breaks the EdgeDB server process tree hierarchy - it's kinda weird.

I'm now leaning towards a different design: the Queen just wait on all its subprocesses, and immediately fork a new one if any worker exited with non-zero code, until all workers exit with zero. And we don't need the Queen-Hive communication any more, as well as the double-forking. Theoretically, edgedb/edgedb-cli#349 won't be an issue.

### DIFF
--- a/edb/server/compiler_pool/amsg.py
+++ b/edb/server/compiler_pool/amsg.py
@@ -20,32 +20,32 @@
 from __future__ import annotations
 
 import asyncio
+import functools
+import logging
 import os
+import socket
 import struct
+import typing
 
 
+logger = logging.getLogger("edb.server")
 _uint64_unpacker = struct.Struct('!Q').unpack
 _uint64_packer = struct.Struct('!Q').pack
 
 
-class PoolClosedError(Exception):
-    pass
+def _wakeup_waiter(waiter, fut):
+    if not waiter.done():
+        waiter.set_result(fut.result())
 
 
-class BaseFramedProtocol(asyncio.Protocol):
+class MessageStream:
+    """Data stream that yields messages."""
 
-    def __init__(self, *, loop, con_waiter=None):
-        self._loop = loop
+    def __init__(self):
         self._buffer = b''
-        self._transport = None
-        self._con_waiter = con_waiter
         self._curmsg_len = -1
-        self._closed = False
 
-    def process_message(self, msg):
-        raise NotImplementedError
-
-    def data_received(self, data):
+    def feed_data(self, data):
         # TODO: rewrite to avoid buffer copies.
         self._buffer += data
         while self._buffer:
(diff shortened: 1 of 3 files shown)


1st1:
> Promoted the first worker into a Queen (talking to the "Hive" in the same protocol for requests to spawn workers)

-1 on this idea.

Let's keep it simple:

* On small machines, when the number of workers is <= 4, we just spawn new worker processes when needed.

* On beefier machines, when the number of workers is > 4, we have one more additional process -- we can call it "template worker". The template worker can do only one thing -- fork new workers. Nothing else. If it dies - we shut down the server, but it shouldn't ever die because it'll pretty much do nothing most of the time.

1st1:
> I'm now leaning towards a different design: the Queen just wait on all its subprocesses, and immediately fork a new one if any worker exited with non-zero code, until all workers exit with zero.

Although this is very similar to my "template worker" proposal, essentially all compiler processes will be children of the template process, right?

fantix:
> > I'm now leaning towards a different design: the Queen just wait on all its subprocesses, and immediately fork a new one if any worker exited with non-zero code, until all workers exit with zero.
> 
> Although this is very similar to my "template worker" proposal, essentially all compiler processes will be children of the template process, right?

Yes, and essentially the same as the current PR (sorry about the wording "Promoted the first worker into a Queen" - the Queen is just the template process, no promotion process). The difference is only how subprocesses are managed.

1st1:
> Yes, and essentially the same as the current PR (sorry about the wording "Promoted the first worker into a Queen" - the Queen is just the template process, no promotion process). The difference is only how subprocesses are managed.

Yeah, "promoting" sounds extremely confusing, as it implies we'd have some sort of consensus algorithm etc. That's way too much complexity, let's keep this as stupid as possible.

Is this ready for a review or you're still working on it?

fantix:
> Is this ready for a review or you're still working on it?

Right, I'd like to try to remove the double-forking and Queen-Hive communication, but add `waitpid()` in the template process.

1st1:
> Queen-Hive communication

Can we also drop the queen/hive terminology? I find it confusing in our very simple setup (and workers don't communicate with the queen that much, they should communicate directly to the IO process)

PR #2551, opened 2021-05-19
Compiler pool worker restart and mgmt

fantix:
### Background

1. The compiler worker processes may exit due to 1) out of memory 2) future timeout kills 3) other unexpected issues.
2. We want the EdgeDB server to keep running if a worker exits like that, and start a new worker to maintain the pool size.
3. Starting a compiler worker is time-consuming. Therefore in the existing design, we only loaded the necessary parser spec and everything in the first worker, and used the low-level `os.fork()` to create other workers to 1) save time reloading those preset, and 2) leverage shared memory.
4. We didn't get involved into the unreliable `SIGCHLD` solution (thus no `waitpid()`) before, so we reused the same UNIX socket for IPC between the server and the workers to also maintain the worker's lifetime: the worker will organically exit if the connection is broken because of e.g. the server is killed, or on the other hand the server will know if the worker dies and disconnects.
5. We'd like to keep the design similarly simple in this PR too.
6. However, the exisiting design will leave zombie processes behind in the scenario described in (1) and (2) because we never waited on the subprocesses. (except for the first worker - the parent of all reamining workers. asyncio will automatically wait on it in `SIGCHLD` handler, but all the remaining workers will be re-parented to a supervisor daemon - PID=1 or systemd user daemon.)

### Solution A

This is implemented in #2548 and abandoned due to complexity. The idea was to keep avoiding `SIGCHLD` and `waitpid()`, and make use of the IPC socket to restart the dead workers. Therefore, we altered the role of the first worker into a "template process" (queen in #2548), that will accept requests from the IPC socket and spawn workers. In order to avoid zombie processes without `waitpid()`, we had to introdue double-forking to daemonize the workers. However, that gets into the same issue of edgedb/edgedb-cli#349.

### Solution B

Implemented in this PR. Similar to solution A, we have a preloaded idle "template process", which keeps waiting on its child processes to exit. If a child exits with an error or gets killed (return code > 0 or killing signal is not SIGINT), the template process will simply fork a new child, until all child processes exits normally. In this case, we won't have zombies (memory is happy), or daemonized processes (CentOS is happy), and the design is simple (no additional IPCs).

On the server side, we keep attaching newly-connected workers to the pool, and exit if the template process dies.

In addition, this PR also made workers asyncio-free.

Closes #2548
Fixes edgedb/edgedb-cli#349

### DIFF
--- a/edb/server/compiler_pool/pool.py
+++ b/edb/server/compiler_pool/pool.py
@@ -36,7 +36,6 @@ import immutables
 from edb.server import pgcluster
 
 from edb.common import debug
-from edb.common import taskgroup
 
 from . import amsg
 from . import queue
@@ -100,7 +99,7 @@ class Worker:
     async def _attach(self, init_args_pickled: bytes):
         self._manager._stats_spawned += 1
 
-        self._con = await self._server.get_by_pid(self._pid)
+        self._con = self._server.get_by_pid(self._pid)
 
         await self.call(
             '__init_worker__',
@@ -141,12 +140,12 @@ class Worker:
             exc.__formatted_error__ = data[0]
             raise exc
 
-    async def close(self):
+    def close(self):
         if self._closed:
             return
         self._closed = True
         self._manager._stats_killed += 1
-        self._manager._workers.discard(self)
+        self._manager._workers.pop(self._pid, None)
         self._manager._report_worker(self, action="kill")
         try:
             os.kill(self._pid, signal.SIGTERM)
@@ -154,10 +153,10 @@ class Worker:
             pass
 
 
-class Pool:
+class Pool(amsg.ServerProtocol):
 
     _workers_queue: queue.WorkerQueue[Worker]
-    _workers: Set[Worker]
+    _workers: Dict[int, Worker]
 
     def __init__(
         self,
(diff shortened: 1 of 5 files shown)


1st1 (commented):
LGTM.

1st1:
The PR looks good to me. Any open questions?

Would be super cool to add some tests but I have no idea how. If you also can't figure out how we can add some tests in reasonable time - we can merge without tests.

fantix:
> The PR looks good to me. Any open questions?

Thanks for checking! I don't have questions so far. The TODO here is to shutdown the server when the template process exited, I'll include that fix in the server shutdown PR after the signal handler PR.

> Would be super cool to add some tests but I have no idea how. If you also can't figure out how we can add some tests in reasonable time - we can merge without tests.

I'm exactly working on adding tests - I'll give it a small time window, so it's either tomorrow or some time later.

edb/server/compiler_pool/pool.py
```diff
@@ -209,20 +209,21 @@ async def _attach_worker(self, pid: int, init_args, init_args_pickled):
...
+        ):
+            logger.info("All %s compiler Workers are ready.", self._pool_size)
+            self._ready_evt.set()
 
-        await self._server.start()
-        self._running = True
+        return worker
 
+    @functools.lru_cache(maxsize=None)
+    def _get_init_args(self):
```
fantix, line 226: Meanwhile, I did find a question regarding the worker's init args. As now we may start workers at any time throughout the server's lifetime, is it okay to keep using the cached init args like `DatabaseState` cached at the time of the server start? Looks like `__sync__` in `worker.py` is used to update the worker state, but I'm not so sure.

1st1: It's OK to keep using the cached args -- the IO process will send the updated bits with the first request to the worker. We can also update the args if necessary (and cache the new version) but I'm not sure it's worth the work.

PR #2554, opened 2021-05-21
Fix a collection of bugs exposed while varying ordering

msullivan:


edb/pgsql/compiler/expr.py
```diff
@@ -614,8 +614,8 @@ def _compile_shape(
     for element in result.elements:
         # The ref might have already been added by the nested shape
         # processing, so add it conditionally.
-        pathctx.put_path_serialized_var_if_not_exists(
-            ctx.rel, element.path_id, element.val, env=ctx.env)
+        pathctx.put_path_serialized_var(
```
elprans, line 617: What's the story here?  Also, the comment is wrong now.

elprans (approved):

edb/pgsql/compiler/expr.py
msullivan: new comment:
```
        # We want to force, because the path id might already exist
        # serialized with a different shape, and we need ours to be
        # visible. (Anything needing the old one needs to have pulled
        # it already: see the "unfortunate hack" in
        # process_set_as_tuple.)
```

PR #2557, opened 2021-05-24
DRAFT: signalctl module

fantix:


tests/common/test_signalctl.py
```diff
@@ -0,0 +1,388 @@
...
+import sys
+import tempfile
+import textwrap
+
+from edb.common import signalctl
+from edb.testbase import server as tb
+
+
+@contextlib.asynccontextmanager
+async def spawn(method):
```
1st1, line 36: I like magic, but this might be a bit too much. Is there any advantages over the approach we use in uvloop? https://github.com/MagicStack/uvloop/blob/master/tests/test_signals.py#L142

1st1 (commented):
Overall this looks good, I think we should try integrating this in edb/server. Can be as part of this PR, as the API might actually change once we start using it

fantix:
Thanks for reviewing! I’ll add to edbserver tmr morning.

tests/common/test_signalctl.py
fantix: The advantage is very limited, like being covered by mypy and flake.

edb/server/main.py
```diff
@@ -363,9 +358,17 @@ def run_server(args: srvargs.ServerConfig, *, do_setproctitle: bool=False):
                         ),
                     )
 
-                _run_server(
-                    cluster, args, runstate_dir, internal_runstate_dir,
-                    do_setproctitle=do_setproctitle,
+                loop = asyncio.new_event_loop()
```
1st1, line 361: nit: Can we try using asyncio.run here?

edb/common/signalctl.py
```diff
@@ -0,0 +1,174 @@
...
+                self._register_waiter(signal, waiter)
+            try:
+                try:
+                    signal = await waiter
+                except asyncio.CancelledError as e:
+                    if not fut.done():
+                        fut.cancel(msg=e)
+                else:
+                    if not fut.done():
+                        fut.cancel(msg=SignalError(signal))
```
1st1, line 121: Add an `assert signal is not None` before this line

edb/common/signalctl.py
```diff
@@ -0,0 +1,174 @@
...
+            for signal in cancel_on:
+                self._register_waiter(signal, waiter)
+            try:
+                try:
+                    signal = await waiter
+                except asyncio.CancelledError as e:
+                    if not fut.done():
+                        fut.cancel(msg=e)
+                else:
+                    if not fut.done():
```
1st1, line 120: Add a comment:
```
# `fut` is still running, which means that `waiter` was woken up
# by a signal.
```

edb/server/pgcon/pgcon.pyx
```diff
@@ -1635,6 +1639,14 @@ cdef class PGConnection:
         while self.buffer.take_message():
             if not self.parse_notification():
                 mtype = self.buffer.get_message_type()
+                if mtype == b'E':  # ErrorResponse
+                    er_cls, fields = self.parse_error_message()
+                    msg = POSTGRES_SHUTDOWN_ERR_CODES.get(fields['C'])
+                    if msg:
+                        msg = fields.get('M', msg)
```
1st1, line 1646: hm, this variable is unused, no? Also, what's the plan on handling the shutdown?

edb/server/main.py
fantix: Sure, let me try that.

edb/server/pgcon/pgcon.pyx
fantix: Oh sorry, this hunk was moved to #2574 - I'll update this PR and remove this change.

1st1 (approved):
LGTM.

Next step would be to implement `async for signal in ...` and react to signals while the server is running.

PR #2561, opened 2021-05-25
Increase the confidence of simple create pointer migrations

### LINKED ISSUE #2547: Non-interactive `create-migration` doesn't work in a very simple case
- EdgeDB Version: 1-beta1, 1-beta2, 1.0-beta.3+dev5645.d2021051711.ge3eaeab87.cv202104290000

Steps to Reproduce:

1. Create a schema: `module default { type Item; }` and apply to the database
2. Modify the schema file to:
```
module default {
    type Item {
        property x -> str;
    };
};
```
3. Run `edgedb create-migration --non-interactive`:
```
Server is about to apply the following migration:
    ALTER TYPE default::Item {
        CREATE PROPERTY x -> std::str;
    };
But confidence is 0.9618666666666666 (minimum is 0.99999)
```

I expect this change to run smoothly in non-interactive mode.

msullivan:
The slight churn in the produced deltas caused us to produce worse
migration prompts in some cases (a "did you alter <parent>?") instead
of a useful prompt about what was really done, so I hacked the code
for DESCRIBE MIGRATION AS JSON to generate better prompts by
rebuilding the delta from the AST. Sorry. This improves our prompts in
some existing cases too.

Fixes #2547.

### DIFF
--- a/edb/server/compiler/compiler.py
+++ b/edb/server/compiler/compiler.py
@@ -1026,12 +1026,27 @@ class Compiler:
                         mstate.target_schema,
                         guided_diff,
                     )
-
                     proposed_steps = []
 
                     if proposed_ddl:
-                        for ddl_text, top_op in proposed_ddl:
-                            prompt_id, prompt_text = top_op.get_user_prompt()
+                        for ddl_text, ast, top_op in proposed_ddl:
+                            # get_ast has a lot of logic for figuring
+                            # out when an op is implicit in a parent
+                            # op. get_user_prompt does not have any of
+                            # that sort of logic, which makes it
+                            # susceptible to producing overly broad
+                            # messages. To avoid duplicating that sort
+                            # of logic, we recreate the delta from the
+                            # AST, and extract a user prompt from
+                            # *that*.
+                            # This is stupid, and it is slow.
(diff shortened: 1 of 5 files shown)


edb/server/compiler/compiler.py
```diff
@@ -1026,12 +1026,27 @@ def _compile_ql_migration(
...
+                            # out when an op is implicit in a parent
+                            # op. get_user_prompt does not have any of
+                            # that sort of logic, which makes it
+                            # susceptible to producing overly broad
+                            # messages. To avoid duplicating that sort
+                            # of logic, we recreate the delta from the
+                            # AST, and extract a user prompt from
+                            # *that*.
+                            # This is stupid, and it is slow.
+                            top_op2 = s_ddl.cmd_from_ddl(
```
elprans, line 1043: Hmm, this is quite unfortunate.  We should be able to extract the elision logic from `get_ast` to a helper method and try to reuse that in `get_user_prompt` and friends.  That would be quite involved, though, so I'm fine with the hack for now.

msullivan: Yeah, that's about where I am.

msullivan:
(Approve it?)

elprans (approved):

PR #2566, opened 2021-05-27
Implement removal/reordering of enum elements

### LINKED ISSUE #2564: Support removing/reordering elements in enums
I was talking to a friend about edgedb and the literal first question that they had was "Can you remove elements from an enum? Because you can't in postgres."

We inherit this limitation from postgres ourselves, but it shouldn't be too hard to do it anyway:
1. Modify the type of every single referencing property to replace the enum with `str`
2. Delete and recreate the enum type
3. Undo step 1

This is unpleasant and nasty but most of the infrastructure needed to do it was built so that we could support adding constraints to scalar types that are used inside of collection types; supporting modifying enums would be a *much* less marginal application of that chunk of code :P. See #2407.

msullivan:
Mostly this is able to leverage the machinery built in #2407 to
handle adding constraints to scalar types that are contained in
composite types.

Fixes #2564.

### DIFF
--- a/edb/pgsql/delta.py
+++ b/edb/pgsql/delta.py
@@ -1669,22 +1669,31 @@ class RebaseScalarType(ScalarTypeMetaCommand,
 
 class AlterScalarType(ScalarTypeMetaCommand, adapts=s_scalars.AlterScalarType):
 
-    def _get_composite_refs(
+    def _get_problematic_refs(
         self,
         schema: s_schema.Schema,
         context: sd.CommandContext,
-    ) -> List[Tuple[s_props.Property, s_types.TypeShell]]:
+        *,
+        composite_only: bool,
+    ) -> Optional[Tuple[
+        Set[s_funcs.Function],
+        List[Tuple[s_props.Property, s_types.TypeShell]],
+    ]]:
         """Find problematic references to this scalar type that need handled.
 
-        Postgres has an irritating limitation that a constraint may
-        not be added to a domain type if that domain type appears in a
-        *composite* type that is used in a column somewhere.
+        This is used to work around two irritating limitations of Postgres:
+          1. That elements of enum types may not be removed or reordered
+          2. That a constraint may not be added to a domain type if that
+             domain type appears in a *composite* type that is used in a
+             column somewhere.
 
-        We don't want to have that limitation, and we need to do a decent
-        amount of work to work around this.
+        We don't want to have these limitations, and we need to do a decent
+        amount of work to work around them.
 
-        1. Find all of the properties whose type is a container type
-           that contains this scalar. (Possibly transitively.)
+        1. Find all of the affected properties. For case 2, this is any
+           property whose type is a container type that contains this
+           scalar. (Possibly transitively.) For case 1, the container type
+           restriction is dropped.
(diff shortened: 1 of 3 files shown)


edb/pgsql/delta.py
```diff
@@ -1771,18 +1798,33 @@ def _undo_everything(
 
         acmd = CommandMeta.adapt(cmd)
         self.pgops.add(acmd)
+
+        for func in funcs:
+            # Force function deletions at the SQL level without ever
```
elprans, line 1803: This is problematic.  Currently, PostgreSQL is fairly ambivalent to dangling references to functions, but that is changing.  Version 14 added new (SQL standard) syntax for function definitions that actually pays attention to function bodies and tracks refs like we do:

```
postgres=# CREATE FUNCTION foo() RETURNS int RETURN (SELECT 1);
CREATE FUNCTION
postgres=# CREATE FUNCTION bar() RETURNS int RETURN (foo());
CREATE FUNCTION
postgres=# SELECT bar();
 bar
─────
   1
(1 row)

postgres=# DROP FUNCTION foo();
ERROR:  cannot drop function foo() because other objects depend on it
DETAIL:  function bar() depends on function foo()
HINT:  Use DROP ... CASCADE to drop the dependent objects too.
```

So we have to either _also_ drop and recreate all referring SQL objects, find some other way to accomplish this, or learn to live with the limitations of enums.

After all, if your enums are unstable like that, you can always fall back to a `str` property with a `one_of` constraint.  Sure, it'll use more space, but you'd be completely free from the issues of Postgres enums.

edb/pgsql/delta.py
```diff
@@ -1815,26 +1865,39 @@ def apply(
...
+
+        schema = ScalarTypeMetaCommand.apply(self, schema, context)
+
         if new_enum_values:
             type_name = common.get_backend_name(
                 schema, new_scalar, catenate=False)
 
-            if old_enum_values != new_enum_values:
+            if needs_recreate:
+                cond = dbops.EnumExists(type_name)
```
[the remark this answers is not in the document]
msullivan: Probably isn't.

edb/pgsql/delta.py
msullivan: We already topo sort the functions, so that part is fine. I'm updating the PR to also drop/recreate constraints. Is there anything else to handle?

elprans: Defaults, indexes, pretty much everything that could reference a function from an expression.

msullivan: OK, I did defaults and indexes (for some reason I thought we did defaults manually in the query compiler, and didn't realize indexes supported that sort of full expression). Looking through the schema, I think that's everything?

There's also a little bit of mechanism around for default values for scalar types, but it seems to be vestigal, with no way to use it? (Oh, and it looks like the pgsql side of things for scalar type defaults will only put constant values into the pgsql schema?)

elprans (approved):
I'm a bit scared at the amount of churn such enum alteration might cause, especially since label addition is essentially free, but we should be able to surface the "migration cost" warnings in the migrations to inform users of how heavy the migration would be, so I think it's worth keeping.

PR #2568, opened 2021-05-28
Compiler pool: Fix the worker start failure check

fantix:
The template process will now abort if the accumulated number of spawns exceeds
2 x default_num_workers, but reset this number to zero if no spawning in 1 second.


1st1 (approved):
LGTM, I hope this fixes the test :)

PR #2569, opened 2021-05-29
Fix deduplication of computable link paths

msullivan:
There are two components to the fix:
 * The scope tree didn't properly do longest-prefix detection
   for computable paths, because of how the head of a computable
   gets wrapped in a branch to protect it from the compiled body
   of the computable.

   The fix to this part is a bit involved:
     * Have attach_subtree detect the BRANCH-covered case and
       be able to merge BRANCH-covered nodes.
     * Track which nodes are merged into which other nodes,
       so that the list of fences that compile_path sets as
       full fences temporarily is correct.
     * Generate the fence_points in attach_path directly, so that
       computable paths *always* use them, not just when actually
       compiling the computable. (Which keeps the merging from
       breaking when the path is registered in another place.)
     * When merging a visible node, use fuse_subtree to fuse its
       children recursively instead of just picking them up from
       the tree traversal. This allows our detection to work when
       there are multiple computables in a path.

 * Fixing the scope tree makes a lot of simple cases work (because
   relying on the semi-joins that get done when the underlying
   paths are referenced work), but handling tricky cases like
   a negative filter requires reworking computable semi-joins
   in the pgsql compiler.

   We need to compile the source and evaluate the body of the
   computable with the source joined in and visible, and *then*
   do a semi-join on the final result.

At one point, things were broken when there was an implicit limit but
not when there wasn't, so I tweaked assert_query_result to make it
easy to insert an implicit limit.

### DIFF
--- a/edb/edgeql/compiler/pathctx.py
+++ b/edb/edgeql/compiler/pathctx.py
@@ -91,7 +91,6 @@ def register_set_in_scope(
         ir_set: irast.Set, *,
         path_scope: Optional[irast.ScopeTreeNode]=None,
         optional: bool=False,
-        fence_points: FrozenSet[irast.PathId]=frozenset(),
         ctx: context.ContextLevel) -> List[irast.ScopeTreeNode]:
     if path_scope is None:
         path_scope = ctx.path_scope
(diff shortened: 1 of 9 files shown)


msullivan:
(OK, I think this is solid now)

elprans (approved):

PR #2574, opened 2021-06-03
Handle Postgres unavailability

fantix:
* Mark the server as PG_UNAVAILABLE if we received SHUTDOWN notification from Postgres through the system db
* Return an error if PG_UNAVAILABLE, when it's trying to acquire a pgcon
* Keep retrying to reconnect to the system db until success
* The existing frontend connections won't be disconnected while the Postgres is unreachable, they will resume to normal after we have Postgres back.


edb/api/errors.txt
```diff
@@ -21,6 +21,9 @@
 
 0x_01_00_00_00   InternalServerError
 
+0x_01_00_00_01   PostgresUnavailableError
```
elprans, line 24: I don't think this should be classified as an `InternalServerError`.  Postgres crashes may happen because of external circumstances like hardware failure, memory pressure etc.  We should add a new category, e.g. `AvailabilityError` (it could also be made retryable, I think).

edb/api/errors.txt
```diff
@@ -128,6 +128,13 @@
 0x_07_01_00_00   AuthenticationError
 
 
+####
+
+0x_08_00_00_00   AvailabilityError
+
+0x_08_00_00_01   PostgresUnavailableError  #SHOULD_RETRY
```
1st1, line 135: Let's rename to `BackendUnavailableError`. Even though we're bound to Postgres, we usually don't leak that in error messages/types.

edb/server/server.py
```diff
@@ -351,11 +356,22 @@ def get_compilation_system_config(self):
         return self._dbindex.get_compilation_system_config()
 
     async def acquire_pgcon(self, dbname):
-        return await self._pg_pool.acquire(dbname)
+        if self._pg_unavailable_msg is not None:
+            raise errors.PostgresUnavailableError(
+                'Postgres is not available: ' + self._pg_unavailable_msg
+            )
+
+        while True:
```
1st1, line 364: Should we have a limit here?

edb/server/server.py
```diff
@@ -708,11 +724,15 @@ async def _acquire_sys_pgcon(self):
...
-            await self.__sys_pgcon.set_server(self)
+        if self.__sys_pgcon is None or not self.__sys_pgcon.is_healthy():
+            conn, self.__sys_pgcon = self.__sys_pgcon, None
+            if conn is not None:
+                self._sys_pgcon_ready_evt.clear()
+                conn.abort()
+            # We depend on the reconnect on connection_lost() of __sys_pgcon
+            await self._sys_pgcon_ready_evt.wait()
+            if self.__sys_pgcon is None:
+                self._sys_pgcon_waiter.release()
```
1st1, line 735: Hm, we usually don't check if `_acquire_sys_pgcon` returns an actual connection (not None) before using what it returned. Perhaps we should raise an error if we fail to obtain a connection (and audit all callers of the function?)

edb/server/server.py
```diff
@@ -788,22 +808,71 @@ def _on_remote_system_config_change(self):
...
+                    conn = await self._pg_connect(defines.EDGEDB_SYSTEM_DB)
+                    break
+                except ConnectionError:
+                    # Keep retrying as far as:
+                    #   1. The EdgeDB server is still serving,
+                    #   2. We still cannot connect to the Postgres cluster, or
+                    pass
+                except pgcon_errors.BackendError as e:
+                    #   3. The Postgres cluster is still starting up
+                    if e.fields['C'] != '57P03':
```
1st1, line 844: Hm, looks like we're now checking these codes in multiple places. How about we add methods to the BackedErorr type, like `is_something()`?

1st1 (commented):
looks great, I think it needs some minor cleanup

edb/server/server.py
fantix: Sure! `_pg_pool.max_capacity + 1` sounds like a reasonable limit

edb/server/server.py
fantix: Discussed offline, conclusion included in the next CRF

1st1 (approved):

PR #2579, opened 2021-06-06
Implement a toy evaluator model

msullivan:
It can handle a pretty big chunk of the non-DML side of the language.
It operates against a hard-coded in-memory database that contains
the `cards` test database and some other stuff.

The *biggest* limitations right now is that it doesn't have any
understanding of a schema (so no inheritance or schema-defined
computables) and that it doesn't understand cardinality at all, so
all shape elements are returned as lists.

I've found this pretty useful in trying to double-check whether real
compiler behaviors seem correct or not.

### DIFF
--- a/edb/tools/toy_eval_model.py
+++ b/edb/tools/toy_eval_model.py
@@ -49,14 +49,15 @@ import functools
 T = TypeVar('T')
 
 
+# This is just documentation I guess.
 SCHEMA = '''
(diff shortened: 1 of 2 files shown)


1st1:
Since this is a "toy model" maybe we merge it to the master branch?

elprans (approved):
Sure, let's

PR #2580, opened 2021-06-07
Support Stolon+Consul (acting as a Stolon proxy)

fantix:
This is a prototype supporting Stolon+Consul high availability. Launch the EdgeDB server with:

```
--postgres-dsn=postgresql://user:pass@<ignored host/port>/dbname
--ha-cluster=stolon+consul:///<stolon cluster name>
```

EdgeDB will subscribe to local Consul http://127.0.0.1:8500/v1/kv/stolon/cluster/{cluster_name}/clusterdata for updates from Stolon, and switch over to the new master Postgres when there is one.

TODO:

- [x] better understand Stolon protocol, e.g. detect partition and disconnect
- [x] remove hard-code
- [x] add tests

PR is built on #2557 and #2574.


1st1:
Please merge the signals PR and rebase this to simplify the review

edb/server/args.py
```diff
@@ -313,6 +316,14 @@ def _writer(status: str) -> None:
         '--auto-shutdown', type=bool, default=False, is_flag=True,
         help='shutdown the server after the last ' +
              'connection is closed'),
+    click.option(
+        '--ha-mode', default='off',
```
elprans, line 320: I wonder if we should use the dsn-like URI format for this instead of separate options, e.g:

`--ha-cluser=stolon+consul://<consul-agent-addr>:8500/<cluster-name>`

This ways we'd also be able to configure Consul connection parameters if needed (e.g. auth, TLS etc).

1st1: +1

.github/workflows.src/tests.inc.yml
```diff
@@ -208,6 +216,31 @@
           cp build/postgres/stamp build/postgres/install/
         fi
 
+    # Build Stolon
+
+    - name: Set up Go
```
elprans, line 221: Hmm. How much does this slow down the test run?  I'm thinking this should be a separate workflow like the connection pool tests.

fantix: For the latest (and the only) successful run, it's 8m15s comparing to recent master 7m55s-ish. The test itself is about 1m20s.

1st1:
Should this one be closed in favor of #2920?

fantix:
Sure, it's included in #2920 anyways.

Issue #2565, opened 2021-05-26
add edgedb-server --auto-shutdown-after=Ns


fmoor:
Adding an `--auto-shutdown-after` flag that takes a duration as it's value could make some use cases simpler. Specifically for testing. It may be more convenient for some testing scenarios to create several short lived connections that don't survive the entire duration of the suite. In this situation `--auto-shutdown` can't be used because the server may stop prematurely. 

see also https://github.com/edgedb/edgedb-python/pull/184#discussion_r640133011

PR #2581, opened 2021-06-07
add --auto-shutdown-after option to edgedb-server

fmoor:
fixes https://github.com/edgedb/edgedb/issues/2565

The new option shuts down the server when all clients have disconnected and no new clients have connected in N seconds.

### DIFF
--- a/edb/server/args.py
+++ b/edb/server/args.py
@@ -84,7 +84,7 @@ class ServerConfig(NamedTuple):
     echo_runtime_info: bool
     emit_server_status: str
     temp_dir: bool
-    auto_shutdown: bool
+    auto_shutdown_after: float
 
     startup_script: Optional[StartupScript]
     status_sink: Optional[Callable[[str], None]]
@@ -310,9 +310,13 @@ _server_options = [
         help='create a temporary database cluster directory '
              'that will be automatically purged on server shutdown'),
     click.option(
-        '--auto-shutdown', type=bool, default=False, is_flag=True,
+        '--auto-shutdown', type=bool, default=False, is_flag=True, hidden=True,
         help='shutdown the server after the last ' +
              'connection is closed'),
+    click.option(
+        '--auto-shutdown-after', type=float, default=-1.0,
+        help='shutdown the server after the last connection has been closed '
+             'for N seconds. N < 0 is treated as infinite.'),
     click.option(
         '--version', is_flag=True,
         help='Show the version and exit.')
@@ -371,6 +375,17 @@ def parse_args(**kwargs: Any):
                 " Please create the database explicitly.",
                 DeprecationWarning,
             )
+
+    if kwargs['auto_shutdown']:
+        warnings.warn(
+            "The `--auto-shutdown` option is deprecated, use "
+            "`--auto-shutdown-after` instead.",
+            DeprecationWarning,
+        )
+        if kwargs['auto_shutdown_after'] < 0:
+            kwargs['auto_shutdown_after'] = 0
+
+    del kwargs['auto_shutdown']
 
     if kwargs['temp_dir']:
         if kwargs['data_dir']:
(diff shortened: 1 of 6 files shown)


edb/server/args.py
```diff
@@ -313,6 +314,10 @@ def _writer(status: str) -> None:
         '--auto-shutdown', type=bool, default=False, is_flag=True,
         help='shutdown the server after the last ' +
              'connection is closed'),
+    click.option(
+        '--auto-shutdown-after', type=float, default=0.0,
+        help='shutdown the server after N seconds of inactivity. '
```
elprans, line 319:
```suggestion
        help='shutdown the server after N seconds of inactivity '
             'after the last connection is closed`
```

`--auto-shutdown` essentially becomes `--auto-shutdown-after=0`.  Perhaps we should also deprecate it.  @1st1?

fmoor: changed. I deprecated and hid the `--auto-shutdown` option.

edb/server/server.py
```diff
@@ -199,15 +205,30 @@ def on_binary_client_connected(self) -> str:
 
     def on_binary_client_authed(self):
         self._binary_proto_num_connections += 1
+
+        if self._auto_shutdown_handler:
```
elprans, line 209: I'd put this into `on_binary_client_connected` instead, otherwise a delay in auth might cause premature termination.

elprans: And I think the `_binary_proto_num_connections` should be moved there also.

edb/server/args.py
```diff
@@ -372,6 +377,13 @@ def parse_args(**kwargs: Any):
                 DeprecationWarning,
             )
 
+    if kwargs['auto_shutdown']:
```
elprans, line 380: Let's turn this into `auto_shutdown_after = 0` and remove the handling of `auto_shutdown`.

edb/server/args.py
```diff
@@ -372,6 +377,13 @@ def parse_args(**kwargs: Any):
...
+        warnings.warn(
+            "The `--auto-shutdown` option is depricated, use "
```
elprans, line 382:
```suggestion
            "The `--auto-shutdown` option is deprecated, use "
```

edb/server/server.py
fmoor: changed

edb/server/server.py
```diff
@@ -195,19 +201,34 @@ def get_pg_dbname(self, dbname: str) -> str:
...
+
         if not self._binary_proto_num_connections and self._auto_shutdown:
             self._accepting_connections = False
             raise SystemExit
 
+        if not self._binary_proto_num_connections and self._auto_shutdown_after >= 0:
+
+            def shutdown():
+                self._accepting_connections = False
+                raise SystemExit
```
elprans, line 227: @1st1, @fantix Is there a better way other than raising `SystemExit`?

1st1: A better way would be to have a special "server_done" future object and to do `await asyncio.wait(server.server_forever(), server_done)` in main.py.

1st1: Actually, since we're talking about auto-shutdown for CI and there are no open connections to EdgeDB, I think this is OK. Just add an XXX, once we have facilities to exit in a better fashion we'll fix this.

edb/server/server.py
fmoor: moving `_binary_proto_num_connections` caused some test to hang so I moved it back

edb/server/server.py
fmoor: added XXX comment

edb/server/args.py
fmoor: changed

edb/server/args.py
fmoor: changed

1st1 (approved):

PR #2582, opened 2021-06-08
add IF NOT EXISTS to DESCRIBE ROLES output

fmoor:
This allows restoring from a dump when the dumped roles already exist in the target instance.


1st1:
@elprans should this be merged?

elprans:
I think we need an option to `DESCRIBE ROLES` to enable conditionals instead of always emitting them.  Exact syntax TBD

1st1:
Syntax proposal: `DESCRIBE ... AS CONDITIONAL DDL;`

fmoor:
Closing because this is very stale.

PR #2584, opened 2021-06-08
Get rid of the temporary scope mechanism

msullivan:
The temporary scope mechanism is a bit of kludgy machinery in the
middle of an important system (process_view), and it would be nice to
have it be gone.

The main technical motivation for this change is that it enables doing
proper cardinality checking on computable pointers that are declared
in shapes not exposed to output.

There wasn't much to fix when removing it: some path_ids need to be
added to the path_log, and some hacky code dealing with UNLESS
CONFLICT ELSE clauses needed to be tweaked.

Main downside to this change is that it makes the scope tree a little
noisier.


elprans (approved):
Nice. This turned out to be much simpler than I'd expected.  Probably because of all the earlier cleanups.  Thanks!

PR #2590, opened 2021-06-10
Implement anonymous shapes

### LINKED ISSUE #2533: Add support for anonymous shapes
Currently, there is no standard and convenient way of selecting multiple expressions of varying cardinality together.  Tuples work well for singletons only, and to use a shape one has to select some object to attach the computables to (e.g the GraphQL extension uses the `stdgraphql::Query` and `stdgraphql::Mutation` singletons for this purpose).

## Proposal

Add a special object type (let's call it `std::VirtualObject`) and allow shapes to be anonymous, i.e not attached to an expression on the left.

For example:

```
SELECT {
   foo := <some-expr>,
   bar := <other-expr>
}
```

The result of such query is dynamically generated `std::VirtualObject` instance.  

Anonymous shapes can be used everywhere where a regular shape can, including in the `WITH` clause:

```
WITH Foo := {comp1 := ..., comp2 := ...}
SELECT Foo { comp1, comp2 }
```

This provides a generic way of associating together sets of values that aren't otherwise associated in any way in the schema.

msullivan:
Add a std::VirtualObject singleton object, similar to
stdgraphql::Query, and make `SELECT { foo := expr, ... }` implicity
operate on a detached VirtualObject.

For annoying LR(1) reasons, we don't support link properties in
anonymous shapes, but that doesn't seem like a big problem.

Elvis suggested removing stdgraphql::Query/Mutation, but I'm holding
off on that in this version because those have an extra __typename
field that would require some handling and I didn't want to muck
around with the graphql compiler in this version. (I tried making
them aliases but trying to create aliases in std library modules
triggers errors about them being read-only.)

Fixes #2533.

### DIFF
--- a/edb/lib/std/60-baseobject.edgeql
+++ b/edb/lib/std/60-baseobject.edgeql
@@ -44,6 +44,11 @@ CREATE ABSTRACT TYPE std::Object EXTENDING std::BaseObject {
         'Root object type for user-defined types';
 };
 
+CREATE TYPE std::VirtualObject EXTENDING std::BaseObject {
+    CREATE ANNOTATION std::description :=
+        'Object type for anonymous shapes';
+};
+
 # 'USING SQL EXPRESSION' creates an EdgeDB Operator for purposes of
 # introspection and validation by the EdgeQL compiler. However, no
 # object is created in Postgres and the EdgeQL->SQL compiler is expected
--- a/edb/server/bootstrap.py
+++ b/edb/server/bootstrap.py
@@ -928,6 +928,7 @@ async def _populate_data(schema, compiler, conn):
     script = '''
         INSERT stdgraphql::Query;
         INSERT stdgraphql::Mutation;
+        INSERT std::VirtualObject;
     '''
 
     schema, sql = compile_bootstrap_script(compiler, schema, script)
--- a/edb/edgeql/ast.py
+++ b/edb/edgeql/ast.py
@@ -517,7 +517,7 @@ class ShapeElement(OffsetLimitMixin, OrderByMixin, FilterMixin, Expr):
 
 
 class Shape(Expr):
-    expr: Expr
+    expr: typing.Optional[Expr]
     elements: typing.List[ShapeElement]
 
 
--- a/edb/edgeql/codegen.py
+++ b/edb/edgeql/codegen.py
@@ -495,10 +495,8 @@ class EdgeQLSourceGenerator(codegen.SourceGenerator):
 
     def visit_Shape(self, node: qlast.Shape) -> None:
         if node.expr is not None:
-            # shape.expr may be None in Function RETURNING declaration,
-            # when the shape is used to described a returned struct.
             self.visit(node.expr)
-        self.write(' ')
+            self.write(' ')
         self._visit_shape(node.elements)
 
     def _visit_shape(self, shape: Sequence[qlast.ShapeElement]) -> None:
--- a/edb/edgeql/compiler/astutils.py
+++ b/edb/edgeql/compiler/astutils.py
@@ -73,7 +73,8 @@ def is_ql_empty_set(expr: qlast.Expr) -> bool:
 
 def is_ql_path(qlexpr: qlast.Expr) -> bool:
     if isinstance(qlexpr, qlast.Shape):
-        qlexpr = qlexpr.expr
+        if qlexpr.expr:
+            qlexpr = qlexpr.expr
 
     if not isinstance(qlexpr, qlast.Path):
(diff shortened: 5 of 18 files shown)


1st1:
@vpetrovykh 

> Elvis suggested removing stdgraphql::Query/Mutation

Victor, once this PR is in, please implement the changes in the GraphQL backend.

msullivan:
> Victor, once this PR is in, please implement the changes in the GraphQL backend.

If we think that using aliases is the right thing, I can fix the stuff that causes aliases to not work in the standard library, but it might just be best to have the graphql compiler use `std::VirtualObject` directly?

edb/server/bootstrap.py
```diff
@@ -928,6 +928,7 @@ async def _populate_data(schema, compiler, conn):
     script = '''
         INSERT stdgraphql::Query;
         INSERT stdgraphql::Mutation;
+        INSERT std::VirtualObject;
```
elprans, line 931: It would be much better if we make each instance of `VirtualObject` distinct, i.e. have a different id.  Automatic injection of `id := uuid_generate_v1mc()` should do the trick.

msullivan: Hm, what's the rationale for caring about that? I don't expect anybody to really be looking at the id of the anonymous queries much, and so I don't see the advantage of adding more magic here, I guess, instead of having it be straightforward syntactic sugar?
Two other points:
 * We don't allow assigning to `id` as a computable now, but it should be too hard to override that for this case if we need to
 * Injecting a call to `uuid_generate_v1mc()` might cause it to need materialization

elprans: > Hm, what's the rationale for caring about that?

Equality.

```
WITH
   Foo := {foo := 1}
   Bar := {bar := '2'}
SELECT Foo = Bar;
```

This should return `false`.

msullivan: Fair, though it seems a little marginal to me still.

It looks like just injecting the `id` wouldn't be sufficient, though; some pgsql-side work would be needed to actually believe that `id` is the object value

1st1: > This should return false.

+1

elprans: > some pgsql-side work would be needed to actually believe that `id` is the object value

Wouldn't materialization solve this naturally?

msullivan: > Wouldn't materialization solve this naturally?

That's a good point: I tried it out on my materialization branch, and it can make the computed `id` get viewed as the object value.

It then tried to join that subquery with the actual backing table (in order to ensure a source rvar), which resulted in an empty output. So if we want to generate fake ids dynamically, we'd need to ensure that we *never* try to join it against the real table for `std::VirtualObject`.

That's probably doable but I think it would be fragile and hacky.

elprans: > That's probably doable but I think it would be fragile and hacky.

It doesn't have to be.  We can use the same machinery we use to do ACL-style filtering on type objects (grep `get_access_policy_filters`).

elprans: Hm, on a second thought, that wouldn't really help here.

> That's probably doable but I think it would be fragile and hacky.

So, if we force materialization on all properties of the anon shape, there would technically be no need to join the actual "source" rvar there, right?  We just need to make sure we're not joining it frivolously, which we shouldn't be doing anyway for performance reasons.

msullivan:
After a bunch of offline discussion, we agreed on sticking with this approach for now, but banning `DISTINCT` on virtual objects.

While I was at it, I also banned different variants of equality, but I can drop those bans if people prefer.

1st1 (approved):

elprans (approved):
Impl lgtm, please add documentation.

edb/lib/std/60-baseobject.edgeql
```diff
@@ -44,6 +44,11 @@ CREATE ABSTRACT TYPE std::Object EXTENDING std::BaseObject {
         'Root object type for user-defined types';
 };
 
+CREATE TYPE std::VirtualObject EXTENDING std::BaseObject {
```
elprans, line 47: This needs to be `final`, I think

msullivan: Agreed that that would be good, unfortunately final doesn't work currently: https://github.com/edgedb/edgedb/issues/2607

msullivan:
> Impl lgtm, please add documentation.

I'm going to merge as is and do docs in a follow up so that I can get the parser changes committed and not need to rebuild grammars as I switch branches

PR #2593, opened 2021-06-11
Docs

### LINKED ISSUE #2401: Improve documentation and tutorials on non-trivial mutations
EdgeQL allows non-trivial mutations: nested `INSERT` and `UPDATE` and insert-or-update-or-select via `UNLESS CONFLICT`.  The tutorial/cheatsheet documentation level is currently very skimpy with respect to the examples those.

vpetrovykh:
Update the cheatsheets with more `INSERT` examples.
    
Add more `UNLESS CONFLICT` and nested `INSERT` examples.
    
Fixes #2401

### DIFF
--- a/docs/cheatsheet/insert.rst
+++ b/docs/cheatsheet/insert.rst
@@ -97,29 +97,104 @@ Insert several nested objects at once:
 ----------
 
 
-"Upserts" as well as and other combinations of :eql:stmt:`INSERT` and
-some alternative operation are possible:
+Sometimes it's necessary to check whether some object exists and
+create it if it doesn't. If this type of object has an exclusive
+property, the ``UNLESS CONFLICT`` clause can make the ``INSERT``
+command indempotent. So running such a command would guarantee that a
+copy of the object exists without the need for more complex logic:
 
 .. code-block:: edgeql
 
-    WITH MODULE people
+    # Try to create a new User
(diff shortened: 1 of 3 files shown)


docs/cheatsheet/insert.rst
```diff
@@ -97,29 +97,104 @@ Insert several nested objects at once:
...
+create it if it doesn't. If this type of object has an exclusive
+property, the ``UNLESS CONFLICT`` clause can make the ``INSERT``
+command indempotent. So running such a command would guarantee that a
+copy of the object exists without the need for more complex logic:
+
+.. code-block:: edgeql
+
+    # Try to create a new User
+    INSERT User {
+        name := "Łukasz Langa",
```
elprans, line 110: I'm sure Łukasz wouldn't mind, but I think it'll be better to use less identifying names in examples.

vpetrovykh: I can use something more generic, but I think we already have `Łukasz` in a couple of places (probably because examples used were matching stuff used in blog posts).

1st1: Let's use something generic.

vpetrovykh:
I've changed the name to something generic in this PR.

elprans (approved):

PR #2594, opened 2021-06-11
Add an auto-generated cast table to the docs.

vpetrovykh:
The table shows all the possible casts for the built-in scalars.


1st1:
Victor, can you add a screenshot of how this will look like in the docs?

vpetrovykh:
I'll need to probably apply some styling to this to make it look sane, because the table is large. Vertical text for the top row should help keep things a bit nicer.

1st1:
> I'll need to probably apply some styling to this to make it look sane, because the table is large. Vertical text for the top row should help keep things a bit nicer.

That will be complicated. I suggest to generate a rst file with simple text/bullet list, that you can include (there should be a directive for that and if it doesn't work we can make it work)

vpetrovykh:
I have uploaded a screenshot with a table and vertical text for top row. The corresponding changes to the edgedb-ui are very small (https://github.com/edgedb/ui/pull/316).

For an overall view of casts a table is better than many lists I think. Although maybe we can add a cast list at the end of each individual scalar page.

Screenshot_20210616_120419

1st1:
How does it look on mobile?

1st1:
Also It's not clear what 'impl' means.

1st1:
> "impl" - in addition to explicit and assignment cast, will be cast
  automatically when used in a context requiring the target type

Is a bit hard to parse, please add an example with `e.g. in ...`

I'd drop the legend under the table and just have a bullet list explaining what's what right above it.

vpetrovykh:
> How does it look on mobile?

This is what iPhone 6/7/8 looks like in chrome:
Screenshot_20210616_125932
iPhone X:
Screenshot_20210616_130100
iPad:
Screenshot_20210616_130241

vpetrovykh:
I've added more explicit descriptions of types of casts and provided examples for them.

vpetrovykh:
The cast table has been trimmed to play nice with smaller screens (see https://github.com/edgedb/ui/pull/316).

1st1 (approved):

1st1:
Thanks Victor!

PR #2595, opened 2021-06-11
Ban "$" as the first character in identifiers.

vpetrovykh:
Much like the leading "@" is banned in backtick-quoted identifiers, so
is the leading "$".

### DIFF
--- a/edb/edgeql-parser/src/tokenizer.rs
+++ b/edb/edgeql-parser/src/tokenizer.rs
@@ -296,6 +296,11 @@ impl<'a> TokenStream<'a> {
                                 "backtick-quoted name cannot \
                                     start with char `@`"));
                         }
+                        if val.starts_with("`$") {
+                            return Err(Error::unexpected_static_message(
+                                "backtick-quoted name cannot \
+                                    start with char `$`"));
+                        }
                         if val.contains("::") {
                             return Err(Error::unexpected_static_message(
                                 "backtick-quoted name cannot \
--- a/tests/test_edgeql_syntax.py
+++ b/tests/test_edgeql_syntax.py
@@ -1048,6 +1048,17 @@ aa';
         SELECT ``::Bar;
         """
 
+    @tb.must_fail(errors.EdgeQLSyntaxError, line=2, col=16)
+    def test_edgeql_syntax_name_27(self):
+        """
+        SELECT `$event`;
+        """
+
+    def test_edgeql_syntax_name_28(self):
+        """
+        SELECT `ok$event`;
+        """
+
     def test_edgeql_syntax_shape_01(self):
         """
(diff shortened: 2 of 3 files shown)


edb/edgeql-parser/src/tokenizer.rs
```diff
@@ -296,6 +296,11 @@ impl<'a> TokenStream<'a> {
                                 "backtick-quoted name cannot \
                                     start with char `@`"));
                         }
+                        if val.starts_with("`$") {
+                            return Err(Error::unexpected_static_message(
+                                "backtick-quoted name cannot \
```
elprans, line 301: I think "backtick-quoted" is redundant in the message.  These characters are banned in all identifiers, quoted or not.

vpetrovykh: True, but only in a very technical sense.
```
SELECT <str>$ident;   # is legal
SELECT <str>`$ident`; # is not legal
```
And to most people the subtlety of `$ident` being a "dollar, followed by an identifier" vs being a special identifier is obscure, especially since parameters work very much like identifiers.

Whereas the error message gets to the core of it, that you cannot write certain things inside backticks.

tests/test_edgeql_syntax.py
```diff
@@ -1048,6 +1048,17 @@ def test_edgeql_syntax_name_26(self):
...
 
+    @tb.must_fail(errors.EdgeQLSyntaxError, line=2, col=16)
+    def test_edgeql_syntax_name_27(self):
+        """
+        SELECT `$event`;
+        """
+
+    def test_edgeql_syntax_name_28(self):
+        """
+        SELECT `ok$event`;
```
1st1, line 1059: I'd ban `$` everywhere, not just when it's the first character. We can relax this ban later, if there's a case for it.

vpetrovykh: As opposed to... literally everything else? Why specifically `$`? There are lots of odd symbols that are allowed inside backticks:
```WITH `[@];-)... "'#//*/` := 42 SELECT `[@];-)... "'#//*/`;```

The simplest usage of backticks is to just allow reserved keywords to potentially be used as identifiers (typically in shapes intended for output), such as `order` or `filter`. However, we also considered that the identifiers might reflect some filenames or even entire paths (for modules, here's a bit of an extreme example: ```WITH MODULE `~/.proj/schema/types` SELECT ...```).

My point is that if you're considering banning `$` everywhere, maybe a bigger review of what we allow in backtick-quoted identifiers is needed.

elprans: Yeah, I also think banning `$` altogether (as opposed to the beginning of ident) isn't warranted.

1st1 (approved):

PR #2596, opened 2021-06-11
Discard disconnected compiler workers from the pool queue

fantix:
This was a bug in the compiler pool that disconnected workers were never removed from the queue, that concurrent compilation may run into the disconnected worker and trigger an error.

### DIFF
--- a/edb/server/compiler_pool/pool.py
+++ b/edb/server/compiler_pool/pool.py
@@ -452,6 +452,19 @@ class Pool(amsg.ServerProtocol):
 
         return preargs, callback
 
+    async def _acquire_worker(self):
+        while (
+            worker := await self._workers_queue.acquire()
+        ).get_pid() not in self._workers:
+            # The worker was disconnected; skip to the next one.
+            pass
+        return worker
+
+    def _release_worker(self, worker):
+        # Skip disconnected workers
+        if worker.get_pid() in self._workers:
+            self._workers_queue.release(worker)
+
     async def compile(
         self,
         dbname,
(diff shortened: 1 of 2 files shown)


edb/server/compiler_pool/pool.py
```diff
@@ -452,6 +452,19 @@ def sync_worker_state_cb(
 
         return preargs, callback
 
+    async def _acquire_worker(self):
+        while (
+            worker := await self._workers_queue.acquire()
+        ).get_pid() not in self._workers:
+            # The worker was disconnected; skip to the next one.
+            pass
+        return worker
```
elprans, line 461: Can all workers in the queue be disconnected?  If so, then this potentially still leaks a disconnected worker (the last one).

edb/server/compiler_pool/pool.py
```diff
@@ -452,6 +452,19 @@ def sync_worker_state_cb(
 
         return preargs, callback
 
+    async def _acquire_worker(self):
+        while (
+            worker := await self._workers_queue.acquire()
```
1st1, line 457: Can this loop complete with no connected worker found?

edb/server/compiler_pool/pool.py
fantix: > Can all workers in the queue be disconnected?

Yes!

> this potentially still leaks a disconnected worker (the last one).

But the last one in the queue won't escape the `.get_pid() not in self._workers` check? `_workers_queue.acquire()` will block after that, until a worker whose pid is in `self._workers` is found 🤔

edb/server/compiler_pool/pool.py
fantix: I don't think so, it'll block until a connected one is found I think.

edb/server/compiler_pool/pool.py
elprans: Oh, right. `acquire()` blocks.

1st1 (approved):

PR #2600, opened 2021-06-16
Do a big overhaul of scope tree prefix factoring

msullivan:
The meat of this rework is in the "unfenced path" case of attach_path,
which had issues when factoring out of an unfenced BRANCH.
In particular, consider the case with a tree like:
```
FENCE {
    FENCE {
        User
    },
    BRANCH,
}
```
and we want to attach User to the BRANCH.
(This would be generated by `((SELECT User), ((User),).0)`,
see test_edgeql_scope_branch_01).
This should result in User being factored to the top FENCE, but
currently this does not happen, because the factoring procedure is
to look at all descendant nodes and then look for any unfenced
descendants of an ancestor; because the existing User is fenced,
it is not found. However, if the tuple was processed in the other
order, we get the right results.

In order to factor, at most one of the paths can be fenced. This means
that the correct approach is to search unfenced paths as long as we
have not seen a fence between the attachment node and the ancestor we
are searching for. Thus, `find_unfenced` becomes `find_factoring_point`
and is able to subsume the old call to `find_descendant_and_ns`.
We also need to search for the *highest* such point.

As part of this, fix some issues in computing which namespaces to strip
when factoring, and make sure that the namespace stripping is applied
to all relevant paths.

I also pulled the error checking for factoring into its own function,
which I think helps make the happy case flow clearer.

Also fix an issue that I introduced in #2569 while fixing computable
paths. In #2569 I identified a BRANCH-protected computable path just
by looking to see if it was covered by a single BRANCH. This is wrong,
since there are other ways to introduce something covered by a BRANCH!
Add a computable_branch flag to directly represent this case.

### DIFF
--- a/edb/ir/scopetree.py
+++ b/edb/ir/scopetree.py
@@ -51,6 +51,9 @@ class ScopeTreeNode:
     fenced: bool
     """Whether the subtree represents a SET OF argument."""
 
+    computable_branch: bool
+    """Whether this is a branch created to hide heads in a computable path"""
+
     protect_parent: bool
     """Whether the subtree represents a scope that must not affect parents."""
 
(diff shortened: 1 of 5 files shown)


elprans (approved):
I like the cleanups a lot!

PR #2604, opened 2021-06-19
Fix an SDL ordering bug.

### LINKED ISSUE #1824: Backlinks don't work in computables when the link is defined in an abstract class
- EdgeDB Version: Alpha 5
- OS Version: Mac OS 10.15.6

Steps to Reproduce:

Create a schema where a computable link uses a backlink for a link defined in an abstract class

For example the following schema:
```
abstract type Action{
    required property txid -> str{
        constraint exclusive;
    }
    required property timestamp -> int32;
    
        required link user -> User;

    
}

type Post extending Action{
  ...
}

type User {
  ...
  posts := .<user[IS Post];
}
```
produces the error: `InternalServerError: reference to an undefined item default::Post@user in default::User@posts`

If I don't include the IS intersect, i.e.:
```
type User {
  ...
  posts := .<user;
}
```
I get the following error: `SchemaError: cannot get 'name' value: item '3b64a89d-50f4-e418-acaa-c19c327e41d9' is not present in the schema <Schema gen:4106 at 0x7f857a0be820>`

vpetrovykh:
The bare backlink used in a computable in SDL is now getting processed
after the potential forward versions of that link.

Fixes #1824

### DIFF
--- a/edb/edgeql/tracer.py
+++ b/edb/edgeql/tracer.py
@@ -76,7 +76,13 @@ class Annotation(NamedObject):
 
 
 class Type(NamedObject):
-    pass
+    def is_scalar(self) -> bool:
+        return False
+
+
+class ScalarType(Type):
+    def is_scalar(self) -> bool:
+        return True
 
 
 TypeLike = Union[Type, s_types.Type]
@@ -138,6 +144,9 @@ SourceLike_T = TypeVar("SourceLike_T", bound="SourceLike")
 class ObjectType(Type, Source):
 
     def is_pointer(self) -> bool:
+        return False
+
+    def is_scalar(self) -> bool:
         return False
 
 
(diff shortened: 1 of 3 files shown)


edb/edgeql/tracer.py
```diff
@@ -578,8 +585,21 @@ def trace_Path(
                         # process it on the next step.
                         pass
                     else:
-                        # otherwise we cannot say anything about the target,
-                        # so bail.
+                        # No type intersection, so the only type that
+                        # it can be is "Object", which is trivial.
+                        # However, we need to make it dependent on
+                        # every link of the same name now.
+                        for fqname, ptr in ctx.objects.items():
```
elprans, line 592: You need to check that `ptr` is actually a `Pointer` here.

vpetrovykh: Added the `Pointer` check.

elprans (approved):

PR #2609, opened 2021-06-22
Fix a cluster of bugs having to do with OPTIONAL or tuples or both

### LINKED ISSUE #2602: Optional arguments seem broken for tuples
~`SELECT (User.name, User.avatar.name) ?? ('hm', 'n/a');` returns only a row for Alice instead of one for every user~

`SELECT (User.name, User.avatar.name) ?? (User.name, 'n/a')` gives an ISE, subquery must return only one column

`SELECT (SELECT ('no', 'no') FILTER false) ?? ('a', 'b');` returns no rows
though `SELECT <tuple<str, str>>{} ?? ('a', 'b')` does

... and `WITH X := (User.name, User.avatar.name), SELECT X ?? ('hm', 'n/a');` generates a silly ISE when trying to compute an implicit cast for a TupleExprAlias, but when I fix that, it produces
```
{
  ('Alice', 'Dragon'),
  ('Bob', 'Dragon'),
  ('Carol', 'Dragon'),
  ('Dave', 'Dragon'),
}
```
which is pretty frighteningly wrong!

msullivan:
* Always wrap function arguments in a SELECT, which fixes an
   issue with optional function arguments that are already SELECTs.
 * Don't overwrite the path_scope_id of an optional argument
   during finalize_args if it already has one
 * Check the scope tree for optionality in both the new and old
   original scope (needed to properly handle WITH bound optional args)
 * Populate optional wrappers with *serialized* tuple vars as
   well as value ones, to avoid column mismatches
 * Don't force non-nullability of tuple elements in _get_path_output,
   since that prevents things from working in an OPTIONAL way.
   Instead, enforce non-nullability of tuple elements as part of
   tuple creation.
 * Make Tuple.find_common_implicitly_castable_type always create
   *Tuples*, even if it an aliased Tuple. Currently it will try to
   *make a new TupleExprAlias which will not work.
 * Disallow passing empty tuples (or tuples of empty tuples) as
   optional arguments, since we can't represent a missing empty
   tuple. We could fix this probably by adding some dummy data for
   empty tuples and letting that be NULL, but it seems pretty
   marginal at this point.

Fixes #2602.

### DIFF
--- a/edb/edgeql/compiler/func.py
+++ b/edb/edgeql/compiler/func.py
@@ -41,7 +41,6 @@ from edb.edgeql import ast as qlast
 from edb.edgeql import qltypes as ft
 from edb.edgeql import parser as qlparser
 
-from . import astutils
 from . import casts
 from . import context
 from . import dispatch
@@ -314,6 +313,8 @@ def compile_operator(
             arg_ir = setgen.scoped_set(
                 setgen.ensure_stmt(arg_ir, ctx=fencectx),
                 ctx=fencectx)
+
+            arg_ir = setgen.ensure_set(arg_ir, srcctx=qlarg.context, ctx=ctx)
 
             arg_ctxs[arg_ir] = fencectx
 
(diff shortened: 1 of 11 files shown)


elprans (approved):

elprans (approved):
LGTM, but it seems the new assertions are failing on some tests.

msullivan:
The failing test is a column count mismatch when compiling
```SELECT <tuple<int64, int64>>{} ?? (1, 2)```

This previously worked basically by accident (it went through the COALESCE optimization); `SELECT <tuple<int64, int64>>{} ?? {(1, 2), (3,4)}` doesn't work.

This change also breaks:
```
SELECT Foo.asdf ?? (1, 2);
```

Other related queries that fail for the same reason (both on master and with this change):
```
SELECT <tuple<int64, int64>>{} UNION (1, 2);
SELECT Foo.asdf UNION (1, 2);
```

elprans:
> The failing test is a column count mismatch when compiling

OK, let's xfail it then (and other related cases).

msullivan:
I figured out how to solve the problem, which ended up resulting in deleting about half the code from the original version of the PR, and picking up like four more bug fixes

elprans (approved):

PR #2610, opened 2021-06-22
[DRAFT] RFC 1008 PoC: TLS and ALPN

fantix:
Draft for PoC purposes, please don't waste time reviewing yet.

Refs edgedb/rfcs#37

### DIFF
--- a/edb/server/args.py
+++ b/edb/server/args.py
@@ -89,6 +89,10 @@ class ServerConfig(NamedTuple):
     startup_script: Optional[StartupScript]
     status_sink: Optional[Callable[[str], None]]
 
+    tls_certfile: Optional[pathlib.Path]
+    tls_keyfile: Optional[pathlib.Path]
+    allow_no_tls: bool
+
 
 class PathPath(click.Path):
     name = 'path'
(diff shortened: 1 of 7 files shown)


edb/server/args.py
```diff
@@ -317,6 +321,23 @@ def _writer(status: str) -> None:
...
+             'signed certificate will be generated and used instead (for '
+             'development only!).'),
+    click.option(
+        '--tls-keyfile', type=PathPath(),
+        help='Specify a path to a file containing the private key. If not '
+             'present, the private key will be taken from --tls-certfile as '
+             'well. If the private key is protected by a password, specify '
+             'with an environment variable EDGEDB_TLS_PRIVATE_KEY_PASSWORD.'),
+    click.option(  # TODO: make the default False
+        '--allow-no-tls', type=bool, default=True, is_flag=True, hidden=True,
```
1st1, line 338: This one is only for `edb server` and not for `edgedb-server`? I think these options are shared between the two

edb/server/protocol/protocol.pyx
```diff
@@ -75,25 +79,50 @@ cdef class HttpProtocol:
...
+                outgoing = ssl.MemoryBIO()
+                incoming = ssl.MemoryBIO()
+                incoming.write(data)
+                sslobj = self.server._sslctx.wrap_bio(
+                    incoming, outgoing, server_side=True
+                )
+                sslobj.do_handshake()
+            except ssl.SSLWantReadError:
+                pass
+            except Exception:
```
1st1, line 101: Probably we should use less genetic exception type here

edb/server/protocol/protocol.pyx
```diff
@@ -75,25 +79,50 @@ cdef class HttpProtocol:
...
                 # This is, most likely, our binary protocol,
                 # as its first message kind is `V`.
                 #
                 # Switch protocols now (for compatibility).
-                binproto = binary.EdgeConnection(
-                    self.server, self.external_auth)
-                self.transport.set_protocol(binproto)
-                binproto.connection_made(self.transport)
-                binproto.data_received(data)
+                self._switch_to_binary_protocol(data)
```
1st1, line 121: Only when `if self.server._tls_compat`?

1st1 (commented):
Nice!

edb/server/protocol/protocol.pyx
fantix: My bad ... this is actually needed by `_return_binary_error()` - I'll make it more readable.

edb/server/args.py
fantix: Yes you're right, `edgedb-server` has these options too. At least `edgedb server` the CLI won't have it.

PR #2614, opened 2021-06-22
Try out TypeGuard for ScopeTreeNodeWithPathId

msullivan:


msullivan:
First take of this failed because the typing_extensions that the edgedb python package depends on is fixed to 3.7.4; I sent a PR there to bump that (and then do we need to do a version bump?), but we can also just if TYPE_CHECKING the import here, so.

edb/ir/scopetree.py
```diff
@@ -21,6 +21,8 @@
 
 from __future__ import annotations
 from typing import *
+if TYPE_CHECKING:
+    from typing_extensions import TypeGuard
```
elprans, line 25: Don't forget to add a dependency to `setup.py`

msullivan: The stubs are mypy internal, and we aren't actually importing it yet at runtime

msullivan:
Any reason not to merge this, since it doesn't actually need the runtime dep?

elprans (approved):

PR #2619, opened 2021-06-24
Ban backlinks on computable links.

vpetrovykh:
Following a computable link backwards is technically possible, but
generally suboptimal. Until such a time when we might have reasonable
strategies for implementing this, the backlinks on computables are
banned.

### DIFF
--- a/edb/edgeql/compiler/setgen.py
+++ b/edb/edgeql/compiler/setgen.py
@@ -664,6 +664,15 @@ def resolve_ptr(
                     ctx.env.add_schema_ref(
                         p.get_nearest_non_derived_parent(ctx.env.schema),
                         track_ref)
+            for ptr in ptrs:
+                if ptr.is_pure_computable(ctx.env.schema):
+                    vname = ptr.get_verbosename(ctx.env.schema,
+                                                with_parent=True)
+                    raise errors.InvalidReferenceError(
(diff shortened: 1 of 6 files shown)


edb/edgeql/compiler/setgen.py
```diff
@@ -653,6 +653,15 @@ def resolve_ptr(
                     ctx.env.add_schema_ref(
                         p.get_nearest_non_derived_parent(ctx.env.schema),
                         track_ref)
+            for ptr in ptrs:
+                if ptr.is_pure_computable(ctx.env.schema):
+                    vname = ptr.get_verbosename(ctx.env.schema,
+                                                with_parent=True)
+                    raise errors.InvalidReferenceError(
+                        f'cannot follow backlink {pointer_name!r} because '
+                        f'the {vname} is computable',
```
elprans, line 662:
```suggestion
                        f'{vname} is computable',
```

`{get_verbosename()} is computable` would be rendered as `link "foo" is computable`, so I don't think an article is needed here, as `link "foo"` is a _title_ in this context, I think.

elprans (approved):

edb/edgeql/compiler/setgen.py
vpetrovykh: This is one of those cases where I would defer to a native speaker. @msullivan?

Issue #2617, opened 2021-06-24
InternalServerError: operator is not unique: timestamp with time zone `-` duration_t


jaclarke:
- EdgeDB Version: 1.0-beta.3+dev5686.d2021062412.g99e1f5eec.cv202106090000

Steps to Reproduce:

```edgeql
select datetime_current() - <duration>'10s';
```

elprans:
This seems to be a regression from the addition of `relative_duration`.  Somehow our operator permutation tests aren't catching this.

PR #2621, opened 2021-06-25
When aliasing SQL functions in stdlib, make sure the return types match

elprans:
The return value of all EdgeQL functions must always satisfy the declared
return type.  However, currently, when declaring a `USING SQL FUNCTION`
function we don't properly enforce the rule, which allows for broken
declarations, e.g. `std::datetime_current` returning a plain
`timestamptz` instead of `edgedb.timestamptz_t`, which would then
confuse downstream function and operator dispatch.  Fix this by adding a
validation check for `USING SQL FUNCTION` functions, and fix all
violations in the stdlib.

Fixes: #2617

### DIFF
--- a/edb/pgsql/delta.py
+++ b/edb/pgsql/delta.py
@@ -890,6 +890,75 @@ class FunctionCommand:
 
         return self.make_function(func, sql_text, schema)
 
+    def sql_rval_consistency_check(
+        self,
+        cobj: s_funcs.CallableObject,
+        expr: str,
+        schema: s_schema.Schema,
+    ) -> dbops.Command:
+        fname = cobj.get_verbosename(schema)
+        rtype = types.pg_type_from_object(
+            schema,
+            cobj.get_return_type(schema),
+            persistent_tuples=True,
+        )
+        rtype_desc = '.'.join(rtype)
+
+        # Determine the actual returned type of the SQL function.
+        # We can't easily do this by looking in system catalogs because
+        # of polymorphic dispatch, but, fortunately, there's pg_typeof().
+        # We only need to be sure to actually NOT call the target function,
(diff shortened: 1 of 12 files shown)


edb/pgsql/delta.py
```diff
@@ -890,34 +890,113 @@ def compile_edgeql_function(self, func: s_funcs.Function, schema, context):
...
+    ) -> dbops.Command:
+        fname = cobj.get_verbosename(schema)
+        rtype = types.pg_type_from_object(
+            schema,
+            cobj.get_return_type(schema),
+            persistent_tuples=True,
+        )
+        rtype_desc = '.'.join(rtype)
+
+        f_test = textwrap.dedent(f'''\
```
msullivan, line 907: Add at least a quick comment here explaining what this is accomplishing.

If you can stomach it, it might also be worth explaining *how*/*why* this works since I certainly don't understand it why this version works and other variants don't.

msullivan (approved):
Looks good. Actually validating this should save us a lot of grief!

edb/pgsql/delta.py
```diff
@@ -890,34 +890,113 @@ def compile_edgeql_function(self, func: s_funcs.Function, schema, context):
...
+        check = dbops.Query(text=f'''
+            PERFORM
+                edgedb.raise_on_not_null(
+                    NULLIF(
+                        pg_typeof(NULL::{qt(rtype)}),
+                        {f_test}
+                    ),
+                    'invalid_function_definition',
+                    msg => format(
+                        '%s is declared to return SQL "%s", but '
```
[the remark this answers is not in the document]
elprans: Fixed.

edb/pgsql/delta.py
elprans: Done.

PR #2625, opened 2021-06-28
Implement path-like syntax for accessing enum types' members

1st1:
The current syntax requires casting of string values into enums, e.g.
`<Color>'Red'`. There are a few shortcomings of this syntax:

* The casting is cumbersome and makes enums feel like non-first-class
  citizens of the spec.

* In EdgeQL query builders for TypeScript (and Python) enum values
  will use the standard attribute access syntax, which would make
  casting in EdgeQL look confusing.

* Our REPL, currently, renders enum values as idents, without qualifying
  the actual type name, e.g. `Red`.  With the new syntax it would be
  possible to switch the REPL (and the interactive tutorial / EdgeDB
  Studio) to render fully-qualifies enum member names, as in
  'default::Color.Red'.

This commit implements path-like syntax for accessing enum values.
Instead of `SELECT <Color>'Red';` it is now possible to run
`SELECT Color.Red;`.

### DIFF
--- a/docs/datamodel/scalars/enum.rst
+++ b/docs/datamodel/scalars/enum.rst
@@ -19,7 +19,14 @@ Enum
 
         scalar type Color extending enum<Red, Green, Blue>;
 
-    :eql:op:`Casting <CAST>` is required to obtain an
+    Enum values can then be accessed directly:
+
+    .. code-block:: edgeql-repl
+
+        db> SELECT Color.Red IS Color;
+        {true}
+
+    :eql:op:`Casting <CAST>` can be used to obtain an
     enum value in an expression:
 
     .. code-block:: edgeql-repl
@@ -27,6 +34,8 @@ Enum
         db> SELECT 'Red' IS Color;
         {false}
         db> SELECT <Color>'Red' IS Color;
+        {true}
+        db> SELECT <Color>'Red' = Color.Red;
         {true}
 
     .. note::
--- a/edb/edgeql/compiler/setgen.py
+++ b/edb/edgeql/compiler/setgen.py
@@ -28,6 +28,7 @@ import contextlib
 
 from edb import errors
 
+from edb.common import levenshtein
 from edb.common import parsing
 
 from edb.ir import ast as irast
@@ -38,6 +39,7 @@ from edb.schema import name as s_name
 from edb.schema import objtypes as s_objtypes
 from edb.schema import pointers as s_pointers
 from edb.schema import pseudo as s_pseudo
+from edb.schema import scalars as s_scalars
 from edb.schema import sources as s_sources
 from edb.schema import types as s_types
 from edb.schema import utils as s_utils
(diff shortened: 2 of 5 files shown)


edb/edgeql/compiler/setgen.py
```diff
@@ -780,6 +790,98 @@ def _is_computable_ptr(
...
+    if step2.direction is not None:
+        step2_direction = s_pointers.PointerDirection(step2.direction)
+    if step2_direction is not s_pointers.PointerDirection.Outbound:
+        raise errors.QueryError(
+            f"enum types do not support backward link navigation",
+            context=step2.context,
+        )
+    if step2.type == 'property':
+        raise errors.QueryError(
+            f"enum types do not support link properties",
```
elprans, line 831:
```suggestion
            f"unexpected reference to link property '{step2.name}' outside of a path expression",
```

edb/edgeql/compiler/setgen.py
```diff
@@ -780,6 +790,98 @@ def _is_computable_ptr(
...
+            context=step2.context,
+        )
+
+    if nsteps > 2:
+        raise errors.QueryError(
+            f"enum types only support two-step paths: "
+            f"`enum name`.`member name`",
```
elprans, line 838:
```suggestion
            f"invalid property reference on a primitive type expression"
```

edb/edgeql/compiler/setgen.py
```diff
@@ -780,6 +790,98 @@ def _is_computable_ptr(
...
+
+    assert isinstance(source, s_scalars.ScalarType)
+    enum_values = source.get_enum_values(ctx.env.schema)
+    assert enum_values is not None
+
+    if ptr_name not in enum_values:
+        raise errors.InvalidReferenceError(
+            f"'{source.get_displayname(ctx.env.schema)!r}' enum "
+            f"has no member called {ptr_name!r}; choose one from "
+            f"{set(enum_values)!r}",
```
elprans, line 865: Not sure if listing all enum members is totally safe, because there can be lots of them.  Perhaps just do the levenstein thing?

docs/datamodel/scalars/enum.rst
```diff
@@ -19,14 +19,21 @@ Enum
...
+
+    :eql:op:`Casting <CAST>` can be used to obtain an
     enum value in an expression:
 
     .. code-block:: edgeql-repl
 
         db> SELECT 'Red' IS Color;
         {false}
-        db> SELECT <Color>'Red' IS Color;
+        db> SELECT <Color>'Red' IS Color.Red;
```
elprans, line 36: This doesn't seem correct: `Color.Red` isn't a type.  Did you mean `SELECT Color.Red IS Color`?

1st1: Good catch, fixed

elprans (approved):

PR #2626, opened 2021-06-28
Extract some of get_path_var into helper functions

msullivan:
There are two commits here; the first one is definitely good, while
the second is a bit more marginal, though I think probably still a net
improvement?

### DIFF
--- a/edb/pgsql/compiler/pathctx.py
+++ b/edb/pgsql/compiler/pathctx.py
@@ -207,11 +207,11 @@ def get_path_var(
     assert src_path_id is not None
 
     # Find which rvar will have path_id as an output
-    src_aspect, rel_rvar, accidentally_found_var = _find_rel_rvar(
+    src_aspect, rel_rvar, found_path_var = _find_rel_rvar(
         rel, path_id, src_path_id, aspect=aspect, env=env)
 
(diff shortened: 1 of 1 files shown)


edb/pgsql/compiler/pathctx.py
```diff
@@ -204,52 +204,14 @@ def get_path_var(
...
-                rel, src_path_id, aspect='source', env=env)
-
-    if rel_rvar is None and alt_aspect is not None:
-        # There is no source range var for the requested aspect,
-        # check if there is a cached var with less specificity.
-        var = rel.path_namespace.get((path_id, alt_aspect))
-        if var is not None:
-            put_path_var(rel, path_id, var, aspect=aspect, env=env)
-            return var
+    if accidentally_found_var:
```
elprans, line 213: Why "accidentally"?

msullivan: I guess it's not really "accidental", but is instead "in passing", or something. The issue is that (annoyingly) the check for presence in an existing tuple needs to be done after the initial attempts to find an rvar but before some of the later ones.

I'll just drop "accidentally"

elprans (approved):

PR #2627, opened 2021-06-28
Specify the mypy plugin by path instead of module

msullivan:
This will make it so that `mypy edb` works, instead of needing to do
`python3 -m mypy edb`.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,6 +1,6 @@
 [mypy]
 python_version = 3.8
-plugins = edb.tools.mypy.plugin
+plugins = edb/tools/mypy/plugin.py
 follow_imports = normal
 ignore_missing_imports = True
 warn_redundant_casts = True


mypy.ini
```diff
@@ -1,6 +1,6 @@
 [mypy]
 python_version = 3.8
-plugins = edb.tools.mypy.plugin
+plugins = edb/tools/mypy/plugin.py
```
elprans, line 3: Would this work if I run mypy from a different working directory?

msullivan: Somewhat surprisingly, yes

msullivan:
(Oh, it looks like the current version will work when edgedb is installed in the current venv. I still want to make this change, though, so it doesn't need to be installed, and so we run the same version of the plugin either way we run mypy)

elprans (approved):

PR #2635, opened 2021-06-30
consolidate protocol cardinality enums

fmoor:
Instead of having an enum for top level query cardinality and a second for shape field cardinality merge them into one enum.

### DIFF
--- a/edb/server/compiler/sertypes.py
+++ b/edb/server/compiler/sertypes.py
@@ -22,7 +22,6 @@
 from __future__ import annotations
 
 import dataclasses
-import enum
 import io
 import struct
 import typing
@@ -37,6 +36,8 @@ from edb.schema import objects as s_obj
 from edb.schema import objtypes as s_objtypes
 from edb.schema import scalars as s_scalars
 from edb.schema import types as s_types
+
+from . import enums
 
 
 _int32_packer = struct.Struct('!l').pack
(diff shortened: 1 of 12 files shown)


edb/server/compiler/sertypes.py
```diff
@@ -38,6 +37,8 @@
 from edb.schema import scalars as s_scalars
 from edb.schema import types as s_types
 
+from edb.testbase import protocol
```
elprans, line 40: Importing this from `testbase` here is a bit weird.  I think we should move the definition here and import it into `testbase/protocol/messages` instead.

fmoor: changed

elprans: Let's also get rid of `ResultCardinality` and change `render_cardinality` to use the consolidated enum

fmoor: changed

elprans: Hm, I don't see changes to `render_cardinality` in `binary.pyx` in the diff still.

edb/server/protocol/binary.pyx
```diff
@@ -1198,14 +1198,14 @@ cdef class EdgeConnection:
 
     cdef char render_cardinality(self, query_unit) except -1:
         if query_unit.cardinality is CARD_NO_RESULT:
-            return <char>(b'n')
-        elif query_unit.cardinality is CARD_ONE:
-            return <char>(b'o')
+            return <char>(CARD_NO_RESULT.value)
```
elprans, line 1201: I don't think there's a reason to explicitly filter for enum values here.  Just return `<char>query_unit.cardinality.value` directly.

fmoor: changed

elprans (approved):
LGTM! Last minor nit: please capitalize the commit message title.

1st1 (approved):

CodesInChaos:
Is there a reason the values of this options are so weird, instead of being the combination of three bits (zero, one, more-than-one) or just consecutive small integers?

fmoor:
> Is there a reason the values of this options are so weird

These values correspond to ASCII characters. I agree that it is not what I would expect for enum values. However, it is what was already being used in some parts of the protocol and changing those parts would have caused even more breakage in clients.

elprans:
Use of ASCII for various enum values in the protocol is a PostgreSQL practice that we inherited.  It helps somewhat when looking at protocol buffer dumps.

CodesInChaos:
I didn't realize they were ASCII, that makes sense.

PR #2636, opened 2021-07-01
Validate explicit type specification in computables

elprans:
Currently, the following would be accepted without error:

```
ALTER TYPE Foo CREATE PROPERTY bar -> str { USING (1) };
```

The explicit type specification would be silently ignored in favor of
the type inferred from the expression.  This is obviously wrong, so make
sure we raise a `SchemaDefinitionError` in cases like this.

### DIFF
--- a/edb/edgeql/ast.py
+++ b/edb/edgeql/ast.py
@@ -1295,12 +1295,19 @@ def get_ddl_field_value(
     return cmd.value if cmd is not None else None
 
 
+def get_ddl_subcommand(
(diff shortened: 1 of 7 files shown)


msullivan (approved):
Looks good. Should avoid some silly issues

1st1 (approved):

elprans:
I changed my mind regarding banning the syntax outright.  I re-rolled the PR to perform validation instead.

msullivan (approved):

PR #2637, opened 2021-07-01
Materialize volatile (and some other) sets in binding constructs

### LINKED ISSUE #2421: Support materializing sets of values in queries
Currently we need to recompute sets essentially each time they appear in the scope tree and whenever a computable property is referenced. This causes big problems with volatile computables and WITH bindings.

We should be able to fix this by "materializing" values instead of recomputing them. This will involve packing up sets into arrays, and unpacking when necessary.

This may be able to help with other issues too, such as tricky cases with FOR and computable properties.

msullivan:
Materializing a set of values consists of:
 * Serializing each of the values in native mode, but with computed
   shape fields included.
 * Then, if the value is possibly multi, aggregating it into an array.

We consider materializing a set of values in two of our core binding
constructs:
 * WITH statements
 * ad-hoc computed pointers.
 * We don't materialize the argument to FOR, because I thought that
   FOR's iteration makes it unnecessary. I found some bugs involving
   FOR, though, so maybe that is wrong.

We materialize a set in the following cases:
 * If it is volatile, or uses a bound variable that is volatile
 * If it appears in a computable and refers to a bound variable.
   This is somewhat overly broad, but is intended to allow us to
   capture FOR variables inside a shape, and similar patterns.

The materialized sets are computed at the enclosing statement:
at the WITH statement, or at the shape subject expression.

We detect when a materialized set is referenced by its type/ptrcls.

To support this, queries on the pgast side gain some new fields for
tracking materialized (or packed) data. packed_path_rvar_map tracks
what rvars contain materialized data while packed_path_outputs
contains the actual output vars.

The most involved part of this is unpack_rvar, which extracts a set
from a packed rvar. It inspects the type of the set, generates SQL to
unpack it, and populate the current statement with the appropriate
rvars for accessing it and its fields. Materialized links (to
materialized shapes) and multi properties of shapes are placed in
packed rvars of their own.

Generating path_ids for tuple paths and shape paths on the fly, in
pgsql, is at least a little hair-raising. (Some dubious machinery
needed to be added to typeutils for the latter.)

In order to ensure that shapes (and tuples that contain shapes) can be
correctly output or correctly *rematerialized*, we need to reconstruct
a serialized version when we unpack. Doing this for tuples means we
need to do it for any shapes they contain, which means that unpack
needs to be driven by a recursive traversal of the type.
This also drives a reconstruction of the *value* of tuples, so that
a tuple containing objects can safely compared for equality and the
like, without inadvertantly comparing materialized shape data.

Known issues:
 * We are a little over-eager to materialize. There are many cases
   where we only use a value once where we will materialize.
   (But also cases where we only use values once where we need to.)
 * We materialize AT_MOST_ONE sets as if they were MANY, serializing
   them into an array. This was an implementation expedient; it
   shouldn't be too hard to use NULL instead.

 * Materializing arrays that contain shapes is pretty busted. The
   tricky part is that unpacking rvars with shapes works by populating
   the current rel with rvars for each of the shape fields. This is
   a problem with arrays, because we can't get at those until after
   some elimination form has been done on the array, and there are
   several of those we'll need to deal with.

 * As I was doing final cleanups on this PR, I found a big cluster of
   bugs involving FOR/tricky uses of WITH (
   test_edgeql_volatility_for_like_hard_*,
   test_edgeql_volatility_for_hard_*).
   I am fairly concerned about these, but didn't want to delay
   getting a PR up for review any longer.

Testing this was a little fiddly, since assert_query_result is pretty
useless when the output is volatile. It was important to test in both
json and native mode (since I had regularly bugs that broke only one
of them!), so I ended up building some of my own infrastructure for that.

This fixes some existing tests (including test_edgeql_scope_tuple_14),
though not as many as I hoped. (I was hoping to get some of the FOR
tests, which I'll hopefully take more of a look at soon.)

Closes #2421.

Co-authored-by: Elvis Pranskevichus <elvis@edgedb.com>

### DIFF
--- a/edb/edgeql/compiler/stmtctx.py
+++ b/edb/edgeql/compiler/stmtctx.py
@@ -50,6 +50,7 @@ from . import inference
 from . import options as coptions
 from . import pathctx
 from . import setgen
+from . import viewgen
 from . import schemactx
 
 
@@ -115,6 +116,27 @@ def fini_expression(
         and pathctx.get_set_scope(ir, ctx=ctx) is None
     ):
         ir = setgen.scoped_set(ir, ctx=ctx)
+
+    # Make sure that all materialized sets have their views compiled
+    v = ast_visitor.NodeVisitor()
+    v.node_visit(ir)
+    children = [x for x in v.memo if isinstance(x, irast.Stmt)]
+    for stmt in children:
+        for key in list(stmt.materialized_sets):
+            mat_set = stmt.materialized_sets[key]
+            ir_set = mat_set.materialized
+            if len(mat_set.uses) <= 1:
+                del stmt.materialized_sets[key]
+            else:
+                with ctx.new() as subctx:
+                    subctx.implicit_tid_in_shapes = False
+                    subctx.implicit_tname_in_shapes = False
+                    assert ir_set.path_scope_id is not None
+                    new_scope = ctx.env.scope_tree_nodes.get(
+                        ir_set.path_scope_id)
+                    assert new_scope
+                    subctx.path_scope = new_scope
+                    viewgen.compile_view_shapes(ir_set, ctx=subctx)
 
     # The inference context object will be shared between
     # cardinality and multiplicity inferrers.
@@ -227,7 +249,7 @@ def fini_expression(
         multiplicity=multiplicity,
         stype=expr_type,
         view_shapes={
-            src: [ptr for ptr, _ in ptrs]
+            src: [ptr for ptr, op in ptrs if op != qlast.ShapeOp.MATERIALIZE]
             for src, ptrs in ctx.env.view_shapes.items()
         },
         view_shapes_metadata=ctx.env.view_shapes_metadata,
@@ -295,9 +317,11 @@ def _try_namespace_fix(
         replacement = scope.find_visible(prefix)
         if (
             replacement and replacement.path_id
-            and replacement.path_id != prefix
+            and replacement.path_id.namespace != obj.path_id.namespace
         ):
-            new = obj.path_id.replace_prefix(prefix, replacement.path_id)
+            new = obj.path_id.strip_namespace(
+                obj.path_id.namespace - replacement.path_id.namespace)
+
             obj.path_id = new
             break
 
--- a/edb/edgeql/compiler/stmtctx.py
+++ b/edb/edgeql/compiler/stmtctx.py
@@ -118,9 +118,8 @@ def fini_expression(
         ir = setgen.scoped_set(ir, ctx=ctx)
 
     # Make sure that all materialized sets have their views compiled
-    v = ast_visitor.NodeVisitor()
-    v.node_visit(ir)
-    children = [x for x in v.memo if isinstance(x, irast.Stmt)]
+    children = ast_visitor.find_children(
+        ir, lambda n: isinstance(n, irast.Stmt))
     for stmt in children:
         for key in list(stmt.materialized_sets):
             mat_set = stmt.materialized_sets[key]
--- a/edb/ir/ast.py
+++ b/edb/ir/ast.py
@@ -739,7 +739,7 @@ class TypeCast(ImmutableExpr):
 
 class MaterializedSet(Base):
     # Hide uses to reduce spew; we produce our own simpler uses
-    __ast_hidden__ = {'uses'}
+    __ast_hidden__ = {'use_sets'}
     materialized: Set
     # We really only want the *paths* of all the places it is used,
     # but we need to store the sets to take advantage of weak
--- a/edb/ir/ast.py
+++ b/edb/ir/ast.py
@@ -69,7 +69,7 @@ import dataclasses
 import typing
 import uuid
 
-from edb.common import ast, compiler, parsing
+from edb.common import ast, compiler, parsing, markup
 
 from edb.schema import modules as s_mod
 from edb.schema import name as sn
@@ -182,7 +182,10 @@ class BasePointerRef(ImmutableBase):
     __ast_hidden__ = {'children'}
 
     # cardinality fields need to be mutable for lazy cardinality inference.
-    __ast_mutable_fields__ = frozenset(('dir_cardinality', 'out_cardinality'))
+    # and children because we update pointers with newly derived children
+    __ast_mutable_fields__ = frozenset(
+        ('dir_cardinality', 'out_cardinality', 'children')
+    )
 
     # The defaults set here are mostly to try to reduce debug spew output.
     name: sn.QualName
@@ -475,6 +478,7 @@ class ComputableInfo(typing.NamedTuple):
     path_id: PathId
     path_id_ns: typing.Optional[WeakNamespace]
     shape_op: qlast.ShapeOp
+    should_materialize: bool
 
 
 class Statement(Command):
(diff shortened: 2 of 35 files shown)


edb/edgeql/compiler/stmtctx.py
```diff
@@ -116,6 +117,27 @@ def fini_expression(
     ):
         ir = setgen.scoped_set(ir, ctx=ctx)
 
+    # Make sure that all materialized sets have their views compiled
+    v = ast_visitor.NodeVisitor()
+    v.node_visit(ir)
+    children = [x for x in v.memo if isinstance(x, irast.Stmt)]
```
elprans, line 123: Why not `ast.find_children`?

edb/ir/ast.py
```diff
@@ -733,6 +737,30 @@ class TypeCast(ImmutableExpr):
     sql_expr: bool
 
 
+class MaterializedSet(Base):
+    # Hide uses to reduce spew; we produce our own simpler uses
+    __ast_hidden__ = {'uses'}
```
elprans, line 742: Did you mean `use_sets` here?

edb/pgsql/ast.py
```diff
@@ -394,17 +400,29 @@ class Query(ReturningQuery):
...
+        PathRangeVar,
+    ]
 
     argnames: typing.Dict[str, Param]
 
     ctes: typing.List[CommonTableExpr]
 
+    def get_rvar_map(self, flavor: str) -> typing.Dict[
+            typing.Tuple[irast.PathId, str], PathRangeVar]:
+        if flavor == 'packed':
```
elprans, line 421: Please use an explicit list of arms for `flavor` here with an `AssertionError` in else in case we add more of them in the future.

edb/pgsql/compiler/relctx.py
```diff
@@ -69,12 +74,18 @@ def pull_path_namespace(
 
     for source_q in source_qs:
         s_paths: Set[Tuple[irast.PathId, str]] = set()
-        if hasattr(source_q, 'path_outputs'):
-            s_paths.update(source_q.path_outputs)
-        if hasattr(source_q, 'path_namespace'):
-            s_paths.update(source_q.path_namespace)
-        if isinstance(source_q, pgast.Query):
-            s_paths.update(source_q.path_rvar_map)
+        if flavor == 'normal':
```
elprans, line 77: Please use an explicit list of arms for `flavor` here with an `AssertionError` in else in case we add more of them in the future.

edb/pgsql/compiler/relctx.py
```diff
@@ -85,21 +96,31 @@ def pull_path_namespace(
             # Skip pulling paths that match the path_id_mask before or after
             # doing path id mapping. We need to look at before as well
             # to prevent paths leaking out under a different name.
-            if (
+            if flavor == 'normal' and (
```
elprans, line 99: I'd invert the check to `flavor != 'packed'` and add a comment why `path_id_mask` doesn't apply for it.

edb/pgsql/compiler/relctx.py
```diff
@@ -225,16 +259,17 @@ def include_specific_rvar(
...
             pathctx.put_path_rvar_if_not_exists(
-                stmt, path_id, rvar, aspect=aspect, env=ctx.env)
+                stmt, path_id, rvar, flavor=flavor, aspect=aspect, env=ctx.env)
 
         scopes = [ctx.scope_tree]
         parent_scope = ctx.scope_tree.parent
         if parent_scope is not None:
             scopes.append(parent_scope)
 
+    if flavor == 'normal':
```
elprans, line 272: Ditto

edb/pgsql/compiler/relctx.py
```diff
@@ -736,6 +780,295 @@ def maybe_get_scope_stmt(
...
+                path_id, alias, multi=False, packed=False,
+                ref=ref if simple else None,
+            ))
+
+        if not simple:
+            if not multi:
+                # Sigh, have to wrap in an array so we can unpack.
+                ref = pgast.ArrayExpr(elements=[ref])
+
+            qry.from_clause[0:0] = [
```
elprans, line 972: `qry.from_clause.insert(0, ...)`?

edb/pgsql/compiler/relgen.py
```diff
@@ -1176,6 +1220,25 @@ def process_set_as_subquery(
...
                     stmt, subrvar, ir_source.path_id, ctx=ctx)
 
+        # If we are looking at a materialized computable, running
+        # get_set_rvar on the source above may have made it show
+        # up. So try to lookup the rvar again, and if we find it,
+        # skip compiling the computable.
+        if ir_source and (new_rvar := _lookup_set_rvar(ir_set, ctx=ctx)):
+            if semi_join:
+                # We need to use DISTINCT, instead of doing an actual
+                # semi-join, unfortunately
```
elprans, line 1230: why?

edb/pgsql/types.py
```diff
@@ -153,6 +153,13 @@ def pg_type_from_scalar(
     return column_type
 
 
+def pg_type_array(tp: Tuple[str, ...]) -> Tuple[str, ...]:
```
elprans, line 156: Use this new helper in `pg_type_from_object` below?

elprans (approved):
Monumental effort, Sully!  The overall approach and implementation look good to me.  A few cosmetic issues here and there.

Have you measured how this affects compilation perf?

edb/edgeql/compiler/stmtctx.py
msullivan: I was going to say it was because find_children doesn't properly handle multiple levels of nested data, and before I created MaterializedSet, I just stored that tuple in a dict.
But *neither* looks at dict *values*, so that isn't it either.

I think maybe I needed it for some debugging I was trying to do?
In any case, it can be `find_children`.

edb/pgsql/compiler/relgen.py
msullivan: Because semi-joins work by taking a fresh, unique copy of the set (from a new base rel) and filtering it down via semi-join against the subquery. But we need to exfiltrate actual data out of the subquery, not just whether something is present.

edb/pgsql/ast.py
msullivan: Fun fact: for a while, I had a `semi-packed` flavor that behaved like `packed` in some places and `normal` in others. It was pretty bad.

edb/pgsql/compiler/relgen.py
msullivan: Hm, I think maybe we could do something like:
```
SELECT ...
    FROM <new base relation>
    CROSS JOIN LATERAL (
        SELECT ... FROM <the subquery with the rvar unpacking>
        WHERE <join condition>
        LIMIT 1
    )
```
Do you think that would be better?

edb/pgsql/compiler/relctx.py
msullivan: Heh: the reason it doesn't apply is because for basically the whole development I had a separate `include_packed_rvar` (I think possibly straight out of your original branch) with a comment above it saying `XXX: can this be merged with include_rvar??`.
All the tests pass without it, but it made me nervous.

msullivan: Oh, hah, actually it looks like the situation is that this check is needed in *at least one of* `_pull_path_namespace` and `include_specific_rvar`.

msullivan:
Oh, I forgot to measure the performance before I merged it. I'll do that today.

1st1:
This is huge, Sully. :confetti_ball:

PR #2638, opened 2021-07-02
Adjust the precedence of DETACHED to match EXISTS/DISTINCT

msullivan:
Currently the situation is that `SELECT DETACHED User { x := User.name }`
parses as `SELECT (DETACHED User) { x := User.name }`,
but the reference to `User` in the shape is still expected to refer
to the same `User` being selected over. (See
`test_edgeql_scope_detached_{07,08}`)

I claim that this makes no sense: `User` is DETACHED, and nothing
outside that detached subexpression ought to be able to correlate with
it.
If we want references inside the body of the shape to correlate, then,
we need to raise the precdence of `DISTINCT` so that the above example
parses as `SELECT DETACHED (User { x := User.name })`.

This also helps solve some other issues with the current implementation:
 * Currently the example query I gave doesn't quite work: it infers
   `x` as multi, and returns the results as sets
 * `SELECT (User.name, { x := User.name })`, which is sugar for
   `SELECT (User.name, (DETACHED VirtualObject) { x := User.name })`,
   ISEs at SQL execution, because the cardinality of `x` is inferred as
   single but `x` is compiled with `User` detached.

These issues are caused by a mismatch between `process_view`, which does
not adopt the namespace of its source, and `_get_computable_ctx`, which
does. In this change, we wrap a DISTINCT expression in another set, to
prevent its freshly generated namespace from leaking out.

I can think of two other ways forward:
 * Accept my claim that a reference to `User` inside
   `SELECT (DETACHED User) { x := User.name }` shouldn't correlate
   with the `User` outside, but don't change the precedence. This breaks
   two tests, but no big deal.
   If somebody really cares about this, they can write
   `SELECT (DETACHED (User { x := User.name }))`, but probably they'd
   actually just use `.name`.

   Implementing this is trivial: keep the change to `expr.py`, revert all
   the parser changes.

   To be honest, I like this approach about as much as actually
   merging this PR. For whatever reason it didn't occur to me that it
   was a reasonable proposal until after I had done the parser changes.
   Oops.

 * Just fix the mismatch between `process_view` and `_get_computable_ctx`,
   preserving both the current parse tree and the (I think) intended semantics.
   I don't like this, because it results in things in the shape being
   essentially detached without actually being detached.

   We could also maybe try to find a way where references to the type
   being selected refers to the detached type (which is in some namespace),
   while other references stay undetached? This might be the actually intended
   semantics, but it's not obvious to me immediately how to implement this,
   and I think it doesn't make a lot of sense.

### DIFF
--- a/docs/edgeql/lexical.rst
+++ b/docs/edgeql/lexical.rst
@@ -254,7 +254,7 @@ content with *dollar-quotes* in an unambiguous manner:
 More specifically, a delimiter:
 
 * Must start with an ASCII letter or underscore
-* Has following characters that can be digits 0-9, underscores or 
+* Has following characters that can be digits 0-9, underscores or
   ASCII letters
 
 .. _ref_eql_lexical_bytes:
@@ -438,14 +438,14 @@ EdgeQL operators listed in order of precedence from lowest to highest:
     * - :eql:op:`*<MULT>`, :eql:op:`/<DIV>`,
         :eql:op:`//<FLOORDIV>`, :eql:op:`%<MOD>`
     * - :eql:op:`??<COALESCE>`
-    * - :eql:op:`DISTINCT`, unary :eql:op:`-<UMINUS>`
+    * - :eql:op:`DISTINCT`, :eql:op:`EXISTS`,
+	    :ref:`DETACHED <ref_eql_with_detached>`, unary :eql:op:`-<UMINUS>`
     * - :eql:op:`^<POW>`
     * - :eql:op:`type cast <CAST>`
     * - :eql:op:`array[] <ARRAYIDX>`,
         :eql:op:`str[] <STRIDX>`,
         :eql:op:`json[] <JSONIDX>`,
         :eql:op:`bytes[] <BYTESIDX>`
-    * - :ref:`DETACHED <ref_eql_with_detached>`
 
 .. |neq| replace:: !=
 .. _neq: ./funcops/generic#operator::NEQ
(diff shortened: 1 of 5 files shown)


docs/edgeql/lexical.rst
```diff
@@ -438,14 +438,14 @@ EdgeQL operators listed in order of precedence from lowest to highest:
     * - :eql:op:`*<MULT>`, :eql:op:`/<DIV>`,
         :eql:op:`//<FLOORDIV>`, :eql:op:`%<MOD>`
     * - :eql:op:`??<COALESCE>`
-    * - :eql:op:`DISTINCT`, unary :eql:op:`-<UMINUS>`
+    * - :eql:op:`DISTINCT`, :eql:op:`EXISTS`,
+	    :ref:`DETACHED <ref_eql_with_detached>`, unary :eql:op:`-<UMINUS>`
```
elprans, line 442: Curios, why isn't `DETACHED` an `eql:op:` like everything else?  @vpetrovykh?

elprans (approved):

docs/edgeql/lexical.rst
msullivan: Huh, it's documented as part of `WITH`?

vpetrovykh: Yeah, comparing it to a top-level `WITH` block seemed like the simplest way to describe this operator, so it's ended up in that part of the documentation rather than with the rest of the operators. Basically, this operator messes with scopes and symbols, rather than actually just doing something to the set of values it is applied to.

PR #2648, opened 2021-07-05
Fix restore of dump version 0.11

tailhook:
This fixes CLI command error (https://github.com/edgedb/edgedb-cli/pull/369/checks?check_run_id=2933758980#step:5:941).

To reproduce: just make a dump and restore the database having at least one object type via CLI from the branch `up_proto`.

Not sure how best to test this? Should I just copy my ugly test dump to the source code? Or should we rely on normal dump/restore cycle test after CLI are merged.


elprans (approved):
OK to rely on new CLI in tests

1st1:
Why wasn't this picked up by our test suite?

PR #2650, opened 2021-07-05
Fix docs build

jaclarke:
Moving the `Cardinality` enum into `edb/server/compiler/enums.py` in https://github.com/edgedb/edgedb/pull/2635 broke the docs build since this enum is rendered on the `docs/internals` pages, and sphinx was trying to import `edb.server.compiler` and it's dependencies.

Importing `Cardinality` directly from `edb.server.compiler.enums` doesn't work (I think because of `__init__.py`), so I've moved the enum into `edb/server/enums.py` to make the docs build work. @elprans Is there a better place to put this enum?


elprans (commented):
Perhaps we should use `edb.protocol` and move this, and relevant parts from `sertypes` and `testbase.protocol` there.  @1st1?

1st1:
> Perhaps we should use edb.protocol and move this, and relevant parts from sertypes and testbase.protocol there. @1st1?

Yeah, sure, makes sense.

elprans (approved):

Issue #2643, opened 2021-07-02
Make server command-line equivalent environment variables have the uniform `EDGEDB_SERVER_` prefix


elprans:
Currently the environment variables corresponding to server command-line flags are prefixed with `EDGEDB_`, which makes them overlap with the environment variables used for client connections, which might lead to surprises.  Fix this by prefixing the server variables uniformly with `EDGEDB_SERVER_` and deprecate the old set.  The docker image is affected in a similar way.

fmoor:
@elprans we aren't updating the prefix for the development debug variables too are we?

elprans:
No, debug vars may stay the same as they already have a prefix.

tailhook:
I'm not sure I nderstand the problem. It was always nice that `EDGEDB_PORT` works both for server and client. So if you change it both server runs on a new port and client does that too.

elprans:
> 'm not sure I nderstand the problem. It was always nice that `EDGEDB_PORT` works both for server and client.

It's cool if you intended that, but is very bad if you didn't.  IMO, explicit separation is more reliable here.

PR #2655, opened 2021-07-06
Remove environment variables from server CLI

fmoor:
related to https://github.com/edgedb/edgedb/issues/2643

This change removes the sparse, unused and undocumented  environment variables from the server CLI.

### DIFF
--- a/edb/server/args.py
+++ b/edb/server/args.py
@@ -200,7 +200,7 @@ def _status_sink_fd(fileno: int) -> Callable[[str], None]:
 
 _server_options = [
     click.option(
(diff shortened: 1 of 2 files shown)


fmoor:
@elprans did I get this right? I could only find 3 variables used by the server and they don't seem to be documented anywhere so I'm not sure what deprecating them amounts to except for not removing them. :confused:

elprans:
> I could only find 3 variables used by the server and they don't seem to be documented anywhere

Most of the env vars currently are exposed by the [docker image](https://github.com/edgedb/edgedb-docker).  Given the rather sporadic use of env vars in the server itself, I think we can get away with dropping them altogether.

elprans (approved):

PR #2656, opened 2021-07-07
Update the CLI docs.

vpetrovykh:
Make sure that the CLI docs are reflecting the current state of the
`edgedb` commands. Group the docs in a similar fashion to how they are
grouped in the CLI help.


docs/admin/configuration.rst
```diff
@@ -22,17 +22,17 @@ parameters using EdgeQL.  For example:
     CONFIGURE SYSTEM SET listen_addresses := {'127.0.0.1', '::1'};
 
 
-edgedb configure
-================
+edgedb config
```
tailhook, line 25: Oh, sorry, this is going to be reverted: https://github.com/edgedb/edgedb-cli/issues/376

docs/changelog/1_0_b1.rst
```diff
@@ -103,7 +103,7 @@ schema to be:
     };
 
 And we apply the changes by using :ref:`edgedb
-create-migration <ref_cli_edgedb_create_migration>` and :ref:`edgedb
+create-migration <ref_cli_edgedb_migration_create>` and :ref:`edgedb
```
tailhook, line 106: You've only changed reference but not the actual command

docs/changelog/1_0_b1.rst
```diff
@@ -126,7 +126,7 @@ migrate <ref_cli_edgedb_migrate>` commands again:
 At this point we may want to actually create a default channel "Main"
 and make the ``channel`` link required. So we alter the schema to make
 the link required and run :ref:`edgedb create-migration
```
tailhook, line 128: Same here.

docs/changelog/1_0_b1.rst
```diff
@@ -191,7 +191,7 @@ apply the changes:
 The above example shows some of the interactions with the EdgeDB
 migration management tools. We will keep improving the inference
 engine that guides the prompts of :ref:`edgedb create-migration
-<ref_cli_edgedb_create_migration>`. However, if the suggestion engine
+<ref_cli_edgedb_migration_create>`. However, if the suggestion engine
```
tailhook, line 194: And here

docs/cli/edgedb_self/edgedb_self_install.rst
```diff
@@ -0,0 +1,49 @@
...
+
+.. cli:synopsis::
+
+    edgedb self install [OPTIONS]
+
+
+Description
+===========
+
+``edgedb self install`` is a terminal command used to install the CLI
```
tailhook, line 18: Maybe mention something along the lines of "if you downloaded the binary manually to a temporary location, this is how you install it into standard location"

tailhook (approved):
Overall looks good. I've not checked files that have been split into multiple, if there are any changes.

Also we might want to mention somewhere that connection options could also be specified at the end of the command too.

docs/changelog/1_0_b1.rst
vpetrovykh: This is deliberate, because this is a changelog reference to the command as it used to be called when it was introduced, but linked to the actual non-deprecated version of the command.

docs/cli/edgedb_self/edgedb_self_install.rst
vpetrovykh: Updated.

docs/changelog/1_0_b1.rst
tailhook: Oops. Sorry, you're right!

vpetrovykh:
I've also added an `edgedb authenticate` section.

docs/admin/configuration.rst
vpetrovykh: I've reverted that bit (and other similar mentions) back to `configure`.

elprans:
@vpetrovykh Merge this?

elprans (approved):

PR #2658, opened 2021-07-07
Don't ask for conversion expressions when changing type of computable

msullivan:


elprans (approved):
There are more issues with computables at pgsql level (we apparently make columns for them and then forget to alter), but this is a good fix on its own.

PR #2659, opened 2021-07-07
Support changing function bodies in migrations

msullivan:
This has a few parts:
 * Turn up the nativecode coefficient
 * Fix generation of AlterFunction asts
 * Fix codegen of AlterFunction

Work on #1772.


elprans (approved):
Also needs a test to check if the return type is validated on the new body

PR #2660, opened 2021-07-08
A string of cleanups as prep for computables-as-aliases

elprans:
See individual commits

### DIFF
--- a/edb/schema/delta.py
+++ b/edb/schema/delta.py
@@ -1112,7 +1112,7 @@ class CommandContext:
         self.backend_runtime_params = backend_runtime_params
         self.affected_finalization: Dict[
             Command,
-            List[Tuple[Command, ObjectCommand[so.Object], List[str]]],
+            List[Tuple[Command, AlterObject[so.Object], List[str]]],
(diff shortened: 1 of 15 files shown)


edb/schema/delta.py
```diff
@@ -2523,6 +2376,15 @@ def compile_expr_field(
             f'{cdn} {self.classname!r}'
         )
 
+    def get_dummy_expr_field_value(
+        self,
+        schema: s_schema.Schema,
+        context: CommandContext,
+        field: so.Field[Any],
+        value: Any,
+    ) -> Optional[s_expr.Expression]:
```
msullivan, line 2392: Maybe add a comment?

edb/schema/referencing.py
```diff
@@ -435,12 +440,6 @@ def get_top_referrer_op(
         return op
 
 
-class StronglyReferencedObjectCommand(
```
msullivan, line 438: Kind of happy to see this go

msullivan (approved):
Great cleanups!

PR #2664, opened 2021-07-09
RFC 1008: TLS and ALPN

fantix:
* Load certs from `data_dir` if no `--tls-cert-file` given
* Generate dev certs if none found in dev mode
* ~Run tests using `_localtest_PORT` temporary instances~

Closes #2610

### DIFF
--- a/edb/server/args.py
+++ b/edb/server/args.py
@@ -91,7 +91,7 @@ class ServerConfig(NamedTuple):
 
     tls_cert_file: Optional[pathlib.Path]
     tls_key_file: Optional[pathlib.Path]
-    optional_tls: bool
+    allow_cleartext_connections: bool
 
 
 class PathPath(click.Path):
@@ -334,9 +334,8 @@ _server_options = [
              'well. If the private key is protected by a password, specify '
              'with an environment variable EDGEDB_TLS_PRIVATE_KEY_PASSWORD.'),
     click.option(
-        '--optional-tls', type=bool, is_flag=True, hidden=True,
-        help='Run the server in TLS-compatible mode, supporting clients that '
-             'use both TLS and cleartext transports.'),
+        '--allow-cleartext-connections', type=bool, is_flag=True, hidden=True,
+        help='Also allow client connections in cleartext in addition to TLS.'),
     click.option(
         '--version', is_flag=True,
         help='Show the version and exit.')
--- a/edb/server/args.py
+++ b/edb/server/args.py
@@ -320,19 +320,21 @@ _server_options = [
         '--auto-shutdown-after', type=float, default=-1.0,
         help='shutdown the server after the last connection has been closed '
              'for N seconds. N < 0 is treated as infinite.'),
-    click.option(
-        '--tls-cert-file', type=PathPath(),
-        help='Specify a path to a single file in PEM format containing the TLS'
-             'certificate as well as any number of CA certificates needed to '
-             'establish the certificate’s authenticity. If not present, a '
-             'self-signed certificate will be generated and used instead (for '
-             'development only!).'),
+    lambda has_devmode, func: click.option(
+        '--tls-cert-file', type=PathPath(), required=not has_devmode,
+        help='Specify a path to a single file in PEM format containing the '
+             'TLS certificate as well as any number of CA certificates needed '
+             'to establish the certificate’s authenticity.' + (
+                 ' If not present, a self-signed certificate will be '
+                 'generated and used instead.' if has_devmode else '')
+    )(func),
     click.option(
         '--tls-key-file', type=PathPath(),
         help='Specify a path to a file containing the private key. If not '
              'present, the private key will be taken from --tls-cert-file as '
              'well. If the private key is protected by a password, specify '
-             'with an environment variable EDGEDB_TLS_PRIVATE_KEY_PASSWORD.'),
+             'with an environment variable '
+             'EDGEDB_SERVER_TLS_PRIVATE_KEY_PASSWORD.'),
     click.option(
         '--allow-cleartext-connections', type=bool, is_flag=True, hidden=True,
         help='Also allow client connections in cleartext in addition to TLS.'),
@@ -342,10 +344,15 @@ _server_options = [
 ]
 
 
-def server_options(func):
-    for option in reversed(_server_options):
-        func = option(func)
-    return func
+def server_options(has_devmode=False):
+    def decorator(func):
+        for option in reversed(_server_options):
+            try:
+                func = option(func)
+            except TypeError:
+                func = option(has_devmode, func)
+        return func
+    return decorator
 
 
 def parse_args(**kwargs: Any):
--- a/edb/server/args.py
+++ b/edb/server/args.py
@@ -89,6 +89,10 @@ class ServerConfig(NamedTuple):
     startup_script: Optional[StartupScript]
     status_sink: Optional[Callable[[str], None]]
 
+    tls_cert_file: Optional[pathlib.Path]
+    tls_key_file: Optional[pathlib.Path]
+    optional_tls: bool
+
 
 class PathPath(click.Path):
     name = 'path'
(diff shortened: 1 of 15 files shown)


edb/server/args.py
```diff
@@ -316,6 +320,23 @@ def _writer(status: str) -> None:
...
+             'self-signed certificate will be generated and used instead (for '
+             'development only!).'),
+    click.option(
+        '--tls-key-file', type=PathPath(),
+        help='Specify a path to a file containing the private key. If not '
+             'present, the private key will be taken from --tls-cert-file as '
+             'well. If the private key is protected by a password, specify '
+             'with an environment variable EDGEDB_TLS_PRIVATE_KEY_PASSWORD.'),
+    click.option(
+        '--optional-tls', type=bool, is_flag=True, hidden=True,
```
elprans, line 337: I'd name this `--allow-cleartext-connections` for clarity.

edb/server/cluster.py
```diff
@@ -245,6 +248,7 @@ async def test(timeout):
                     conn = await edgedb.async_connect(
                         host=str(self._runstate_dir),
                         port=self._effective_port,
+                        admin=True,
```
elprans, line 251: We are deprecating and removing the `admin` argument in bindings.  Why is this necessary?

edb/server/main.py
```diff
@@ -217,6 +223,50 @@ async def _run_server(
...
+                    "_generate_dev_cert",
+                    "--cert-file",
+                    str(cert_file),
+                    "--key-file",
+                    str(key_file),
+                ]
+            )
+            return str(cert_file), str(key_file)
+
+    abort("Cannot start EdgeDB server without TLS certificate.")
```
elprans, line 267:
```suggestion
    abort("Cannot start EdgeDB server without a TLS certificate, use the `--tls-cert-file` command-line argument to specify it.")
```

edb/server/server.py
```diff
@@ -811,36 +819,33 @@ async def run_startup_script_and_exit(self):
     async def _start_servers(self, host, port):
         nethost = await _fix_localhost(host)
 
+        sslctx = ssl.SSLContext()
+        sslctx.load_cert_chain(
```
elprans, line 823: We should handle cert issues here and wrap them with a nice error.  Specifically, if `tls_key_file` is missing and not present in `cert_file`.

edb/server/server.py
```diff
@@ -1010,3 +1016,7 @@ async def _fix_localhost(host):
         hosts.insert(idx, addr)
 
     return hosts
+
+
+def _tls_private_key_password():
+    return os.environ.get('EDGEDB_TLS_PRIVATE_KEY_PASSWORD', '')
```
elprans, line 1022: Let's use `EDGEDB_SERVER` as the var prefix.

edb/server/cluster.py
fantix: It's not - it's just easier to make the tests pass for now, given that we still have a few `_admin_query()` calls. I can fix those to use the CLI instead of the Python binding.

edb/server/server.py
```diff
@@ -852,13 +856,58 @@ async def _start_servers(self, host, port):
...
+                        "the private key file using the `--tls-key-file` "
+                        "command-line argument?"
+                    ) from e
+                else:
+                    raise StartupError(
+                        "Cannot load TLS certificates - please double check "
+                        "if the specified certificate files are valid."
+                    )
+
+            raise StartupError("") from e
```
elprans, line 906:
```suggestion
            raise StartupError from e
```

edb/server/protocol/protocol.pyx
```diff
@@ -189,13 +231,89 @@ cdef class HttpProtocol:
...
+        else:
+            # It's either HTTP as the negotiated protocol, or the negotiation
+            # failed and we have no idea what ALPN the client has set. Here we
+            # just start talking in HTTP, and let the client bindings decide if
+            # this is an error based on the ALPN result.
+            self._init_http_parser()
+
+    async def _return_binary_error(self, proto):
+        await proto.write_error(errors.BinaryProtocolError(
+            'TLS Required',
```
elprans, line 283: Can we add a test for this condition?  `edgedb-python` auto-downgrades to cleartext on cert errors, we can utilize that.

elprans (approved):
LGTM overall.  @1st1, please take a look too.

edb/server/args.py
```diff
@@ -316,6 +320,22 @@ def _writer(status: str) -> None:
         '--auto-shutdown-after', type=float, default=-1.0,
         help='shutdown the server after the last connection has been closed '
              'for N seconds. N < 0 is treated as infinite.'),
+    click.option(
+        '--tls-cert-file', type=PathPath(),
+        help='Specify a path to a single file in PEM format containing the TLS'
```
1st1, line 325:
```suggestion
        help='Specify a path to a single file in PEM format containing the TLS '
```

edb/server/args.py
```diff
@@ -316,6 +320,22 @@ def _writer(status: str) -> None:
...
+             'certificate as well as any number of CA certificates needed to '
+             'establish the certificate’s authenticity. If not present, a '
```
1st1, line 327: I'd do this differently.

`edb server` should have `--tls-cert-file` optional.
`edgedb-server` should have it required.

With adjusted help section for each tool.

edb/server/args.py
```diff
@@ -316,6 +320,22 @@ def _writer(status: str) -> None:
...
+             'self-signed certificate will be generated and used instead (for '
+             'development only!).'),
+    click.option(
+        '--tls-key-file', type=PathPath(),
+        help='Specify a path to a file containing the private key. If not '
+             'present, the private key will be taken from --tls-cert-file as '
+             'well. If the private key is protected by a password, specify '
+             'with an environment variable EDGEDB_TLS_PRIVATE_KEY_PASSWORD.'),
```
1st1, line 335:
```suggestion
             'with an environment variable EDGEDB_SERVER_TLS_PRIVATE_KEY_PASSWORD.'),
```

edb/server/cluster.py
```diff
@@ -216,69 +218,59 @@ async def _read_server_status(stream: asyncio.StreamReader):
...
+            raise ClusterError(
+                f'could not connect to edgedb-server '
+                f'within {timeout} seconds') from None
+
+    def _admin_query(self, query, wait_until_available="0s"):
+        return subprocess.call(
+            [
+                sys.executable,
+                "-m",
+                "edb.cli",
```
1st1, line 256: I think we should call the `edgedb` command here explicitly; `edb.cli` is just a wrapper.

edb/server/main.py
```diff
@@ -217,6 +223,51 @@ async def _run_server(
...
 
+def _init_tls_certs(
+    args: srvargs.ServerConfig,
+    cert_dir: Optional[tempfile.TemporaryDirectory],
+):
+    cert_path = None
+    if args.tls_cert_file:
+        cert_file = pathlib.Path(args.tls_cert_file).resolve()
+        if args.tls_key_file is None:
+            return str(cert_file), None
```
1st1, line 234: I'm just curious -- why not just return Path objects?

edb/server/protocol/protocol.pyx
```diff
@@ -189,13 +231,89 @@ cdef class HttpProtocol:
...
+            binproto.data_received(data)
+        return binproto
+
+    def _init_http_parser(self):
+        self.parser = httptools.HttpRequestParser(self)
+        self.current_request = HttpRequest()
+
+    async def _forward_first_data(self, data):
+        # As we stole the "first data", we need to manually send it back to
+        # the SSLProtocol. The hack here is highly-coupled with uvloop impl.
```
1st1, line 248: Perhaps we should add an explicit method for this in uvloop, like `inject_data` or something. But that can be done separately if you think it's a good idea.

1st1 (approved):
Great work! I've left some comments.

edb/server/main.py
fantix: Oh yes, `Path` objects would do better! It's been a few versions of changes and I cannot remember how it ended up in strings. Let me push a fix.

edb/server/cluster.py
fantix: But the `edgedb` command is not always in the `PATH` - finding the path to the binary is the same code as `edb.cli`(`.__main__`) I think?

elprans: This method is only used in tests to configure test clusters, using `edb.cli` here is appropriate.

edb/server/protocol/protocol.pyx
fantix: > `edgedb-python` auto-downgrades to cleartext on cert errors

Oh it's actually the opposite - it auto-downgrades to cleartext if the `SSLError` is not a cert error. That means the server is likely an old one without TLS support, and the client binding is allowed to use cleartext.

I just added 2 tests using `edb.protocol.protocol`, that should cover the case.

edb/server/server.py
fantix: Oops my bad - missed one error message here.

edb/server/args.py
fantix: Had to add a hidden `edb testserver` sub-command for the tests to leverage the cert auto-generation feature, at least the tests passed now.

elprans: What is the reason for avoiding generating dev certs in `edb server`?

fantix: Oh we also generate dev certs in `edb server`, it's just that neither `edb server` nor `edgedb-server` (`python -m edgedb.server.main`) is suitable for running the tests. `edb server` has `insecure=True` and `edgedb-server` doesn't generate the certs.

edb/server/cluster.py
fantix: Ohhh this is biting the nightly build test where we don't have the `edgedb` CLI binary. I'll change it to use `edb.protocol.protocol` to connect to the UNIX socket directly and run admin queries there.

elprans: We actually have it in `PATH` there, so it might be easier to simply do use unprefixed "edgedb" here?

fantix: Ah yes, sure thing!

PR #2666, opened 2021-07-09
Make ObjectShell a Generic type

elprans:
`ObjectShell` impersonates a schema object of a specified type, so it
makes sense to expose the underlying object type to the type checker.


msullivan (approved):
I like this

PR #2668, opened 2021-07-09
Delete the test_edgeql_ir_scope_tree tests

msullivan:
They are pretty pointless, since I just automatically update them
whenever I break them.
I've kept around some that test for error conditions.


elprans (approved):
Yep, they clearly outlived their purpose.

PR #2669, opened 2021-07-09
Fix an issue with cycle detection.

vpetrovykh:
Migrations introducing cycles into the schema should not cause
InternalServerError.
    
The SDL cycle detection appears to be robust and the error message has
been adjusted.
    
The DDL cycle detection has a few corner cases where the tests are
failing.

### DIFF
--- a/edb/edgeql/declarative.py
+++ b/edb/edgeql/declarative.py
@@ -189,10 +189,42 @@ class TraceContextBase:
         if extra_name:
             fq_name = s_name.QualName(
                 module=fq_name.module,
-                name=f'{fq_name.name}:{extra_name}',
+                name=f'{fq_name.name}@@{extra_name}',
             )
 
         return name, fq_name
+
+
+def get_verbosename_from_fqname(
+    fq_name: s_name.QualName,
+    ctx: DepTraceContext,
+) -> str:
+    traceobj = ctx.objects[fq_name]
+    assert traceobj is not None
+
+    name = str(fq_name)
+    clsname = traceobj.__class__.__name__.lower()
+    ofobj = ''
+
+    if isinstance(traceobj, qltracer.Alias):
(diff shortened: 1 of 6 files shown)


vpetrovykh:
So there was a problem with an `IndexError` being produced when trying to generate the `CycleError` message. I fixed that.

The `topological.sort()` is used in a number of places and at least some of them are user-facing that need better error messages, so I'm wondering if it's better:
0) to mess with some custom `isinstance` when the `CycleError` is generated 
1) to optionally pass in a custom error generating function for the cases when the errors need to be prettier
2) to store `item` in addition to `path` in the Cycle error and let the call-site unpack those and generate better error message if needed

vpetrovykh:
I should probably add some DDL tests for triggering this as well.

elprans (approved):

elprans:
>  I'm wondering if it's better

Handling a `CycleError` seems like the most Pythonic way.  You can update it to contain all the information you need to produce a good user-facing error.

1st1:
> Handling a CycleError seems like the most Pythonic way. You can update it to contain all the information you need to produce a good user-facing error.

+1

edb/edgeql/declarative.py
```diff
@@ -189,12 +189,44 @@ def get_fq_name(
...
             fq_name = s_name.QualName(
                 module=fq_name.module,
-                name=f'{fq_name.name}:{extra_name}',
+                name=f'{fq_name.name}@@{extra_name}',
             )
 
         return name, fq_name
 
 
+def get_verbosename_from_fqname(
```
elprans, line 198: Why not use `get_verbosename_static()`?

vpetrovykh: I'm not sure it would be compatible with the way tracer names things.

PR #2677, opened 2021-07-14
Avoid marking `--tls-cert-file` as required on Click level

elprans:
This breaks executable invocations that have nothing to do with TLS,
like `edgedb-server --version`.  The check in `parse_args()` is
sufficient.


1st1 (approved):

fantix (approved):
Also reverted the `edb testserver` subcommand.

PR #2680, opened 2021-07-14
Refactor: optionally generate certs in the server

fantix:
Use cryptography to generate self-signed certs if the switch `--generate-self-signed-cert` is set (default for dev mode).

Also fixed to use the latest edgedb-python (`--tls-ca-cert`).

### DIFF
--- a/edb/server/args.py
+++ b/edb/server/args.py
@@ -320,15 +320,12 @@ _server_options = [
         '--auto-shutdown-after', type=float, default=-1.0,
         help='shutdown the server after the last connection has been closed '
              'for N seconds. N < 0 is treated as infinite.'),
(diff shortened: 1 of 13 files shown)


edb/server/args.py
```diff
@@ -39,6 +39,8 @@
 
 
 BYTES_OF_MEM_PER_CONN = 100 * 1024 * 1024  # 100MiB
+TLS_CERT_FILE_NAME = "certificate.crt"
```
elprans, line 42: I think we should name these `edbtlscert.pem` and `edbprivkey.pem` to indicate that these are edgedb's files, and `*.pem` is a more appropriate extension given the content format.

elprans (approved):

PR #2689, opened 2021-07-15
Allow generate cert with --bootstrap-only

fantix:
Now `edb server --bootstrap-only --generate-self-signed-cert` would work together.

@elprans You menthoned there were some other comments re #2680, I could include the fix in this PR.


elprans:
> You menthoned there were some other comments

Yeah, the only material thing is that I think we should put generated certs into `internal_runstate_dir` instead of a separate temporary directory.

elprans (approved):

elprans (approved):

PR #2691, opened 2021-07-16
Include cert path in the runtime info

fantix:
This is used in CLI test suite.

### DIFF
--- a/edb/server/server.py
+++ b/edb/server/server.py
@@ -954,6 +954,7 @@ class Server:
             ri = {
                 "port": self._listen_port,
                 "runstate_dir": str(self._runstate_dir),
+                "tls_cert_file": self._tls_cert_file,
             }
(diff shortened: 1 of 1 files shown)


edb/server/server.py
```diff
@@ -954,6 +954,7 @@ async def start(self):
             ri = {
                 "port": self._listen_port,
                 "runstate_dir": str(self._runstate_dir),
+                "tls_cert_file": self._tls_cert_file,
```
elprans, line 957: This is OK, but we should fix the callers to use the non-deprecated option below.

elprans (approved):

edb/server/server.py
fantix: Oh cool! Was still wondering why we have both echo and emit. Will fix the CLI test laters :+1:

PR #2692, opened 2021-07-16
Use anonymous shapes in GraphQL translation.

vpetrovykh:


edb/lib/ext/graphql.edgeql
```diff
@@ -21,19 +21,6 @@ CREATE EXTENSION PACKAGE graphql VERSION '1.0';
 
 CREATE MODULE stdgraphql;
```
elprans, line 22: Can we get rid of this module altogether?

vpetrovykh: We have a helper function in there that converts the names of types into their GraphQL names. It's something that can be inlined I suppose. Just a bit messy.

elprans: Having a top-level stdlib module just for a helper function is messy too.  As a bonus, you'd be able to actually handle hypothetical cases where `default::` isn't your default module more easily that way.

vpetrovykh: OK, I'll go inline it. maybe it won't be too messy.

vpetrovykh: Gaaah! The short answer is that inlining it breaks `test_graphql_functional_alias_04`. I'm going to need to examine this a bit more thoroughly, because currently there's at least 2 different ways it gets broken: can't prove cardinality ONE for the typename (in the alias computed link) and also getting the wrong result for the type name. They are probably related, but it's a bit messy, so I'm trying to minimize the offending expression.

1st1: > can't prove cardinality ONE for the typename

You can use the new "assert_single()" for that

vpetrovykh:
This is currently blocked by the following bugs:
- https://github.com/edgedb/edgedb/issues/2707
- https://github.com/edgedb/edgedb/issues/2708

Whereas https://github.com/edgedb/edgedb/issues/2707 can be worked around by using an `assert_single` hack, but the https://github.com/edgedb/edgedb/issues/2708 doesn't seem to have a simple workaround.

elprans (approved):

Issue #2402, opened 2021-03-24
Add support for runtime cardinality assertion


elprans:
There are many occasions in queries and schema where the compiler must prove that the expression result has a certain cardinality.  Not all cases are provable statically, either in principle or due to inference shortcomings, so we customarily use the `LIMIT 1` workaround on such expressions, which is bad, because it simply hides the fact that an expression might return multiple values, effectively negating the soundness check.

Instead of `LIMIT 1` there should be an explicit runtime assertion instruction, which would raise an exception at runtime if the desired cardinality is violated.

For example:

```
INSERT Obj { foo := ASSERT SINGLE (SELECT Foo FILTER <complex-filter-eliminating-all-but-one>) }
```

msullivan:
This could be a function, right? `assert_single(...)`? (Albeit maybe one that requires special handling in pgsql)

elprans:
Syntactically, sure it can be function-like.  That said, I generally lean toward preferring operator syntax for special stuff like this, which is why, for example, `DETACHED` is an operator and not a function.

PR #2695, opened 2021-07-16
Implement std::assert_single()

elprans:
The new `std::assert_single(input: SET OF anytype) -> anytype` function
checks the cardinality of the input set, and if it is greater than 1,
raises a `CardinalityViolationError`.  This runtime cardinality
assertion is intended to be used in place of the current `LIMIT 1` hack
to safely place a singleton cardinality invariant on an expression where
static inference cannot help.

Fixes: #2402


tests/test_edgeql_expressions.py
```diff
@@ -4778,3 +4778,59 @@ async def test_edgeql_normalization_missmatch_01(self):
...
+            }
+        """)
+
+        await self.assert_query_result("""
+            SELECT assert_single((
+                SELECT User { name } FILTER .name ILIKE "He Who%"
+            ))
+        """, [{
+            "name": "He Who Remains",
+        }])
```
msullivan, line 4795: Could you add a test with a computed property, and a test that does some computation on a shape result (instead of directly outputting it)?

edb/pgsql/compiler/relgen.py
```diff
@@ -1902,6 +1907,80 @@ def process_set_as_expr(
...
+def process_set_as_singleton_assertion(
+    ir_set: irast.Set,
+    stmt: pgast.SelectStmt,
+    *,
+    ctx: context.CompilerContextLevel,
+) -> SetRVars:
+    expr = ir_set.expr
+    assert isinstance(expr, irast.FunctionCall)
+
+    with ctx.subrel() as newctx:
```
msullivan, line 1919: Could you add a comment that shows the SQL we're generating here?

msullivan (approved):

tests/test_edgeql_expressions.py
1st1: > and a test that does some computation on a shape result (instead of directly outputting it)?

+1

1st1 (approved):

PR #2679, opened 2021-07-14
Add `identifier` and `description` annotations to operators

jaclarke:

PR #2702, opened 2021-07-19
Fix unpacking of materialized persistent tuples

msullivan:
Should resolve the issue blocking #2679.

### DIFF
--- a/tests/test_edgeql_volatility.py
+++ b/tests/test_edgeql_volatility.py
@@ -427,6 +427,15 @@ class TestEdgeQLVolatility(tb.QueryTestCase):
             [False, False, True, True],
         )
 
+    async def test_edgeql_volatility_with_05(self):
+        await self.con.execute(r'''
+            CREATE TYPE Foo { CREATE PROPERTY asdf -> tuple<float64> };
+        ''')
+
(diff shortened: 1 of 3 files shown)


msullivan:
(It does not resolve the issue)

elprans (approved):

tests/test_edgeql_volatility.py
```diff
@@ -427,6 +427,19 @@ async def test_edgeql_volatility_with_04(self):
...
+            CREATE TYPE Foo { CREATE PROPERTY asdf -> tuple<float64> };
+        ''')
+
+        await self.con.query(r'''
+            WITH X := (random(),) SELECT X.0;
+        ''')
+
+        await self.con.query(r'''
+            WITH X := {(random(),),(random(),)} SELECT X.0;
+        ''')
```
1st1, line 441: Since this test doesn't check the output, I assume it tests that the compiler can compile this query?

1st1 (approved):

tests/test_edgeql_volatility.py
msullivan: Yeah, and the result is a random number so there's not a whole lot of testing we can do on it

PR #2709, opened 2021-07-20
Fix some issues with tuple/union interactions

msullivan:
* Don't pull_namespace in UNION and coalesce, since that can
   improperly expose NULL values to correlated sets.
 * Filter types with matching material_types in range_for_typeref
   to avoid incorrect duplicates.

### DIFF
--- a/tests/test_edgeql_select.py
+++ b/tests/test_edgeql_select.py
@@ -2393,7 +2393,6 @@ class TestEdgeQLSelect(tb.QueryTestCase):
             ],
         )
 
-    @test.xfail('Too many results')
     async def test_edgeql_select_setops_13a(self):
         await self.assert_query_result(
             r"""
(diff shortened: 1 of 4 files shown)


tests/test_edgeql_select.py
```diff
@@ -2530,6 +2528,28 @@ async def test_edgeql_select_setops_19(self):
...
+            )
+        ''')
+        self.assertEqual(len(res), 4)
+        for row in res:
+            self.assertNotEqual(row[1].id, None)
+
+    async def test_edgeql_select_setops_21(self):
+        res = await self.con.query(r'''
+            SELECT (
+                'oh no' ?? (SELECT Issue.time_spent_log.body FILTER false),
```
vpetrovykh, line 2545: Why is the order of `??` operands like that? Normally I'd expect to put the "potentially empty set" as the first operand and the "backup" as the second.

msullivan: Yeah, indeed that would make more sense as a useful query, but this is the order that was broken :P

(In a more practical broken query, I guess both sides would be potentially empty)

elprans (approved):

PR #2715, opened 2021-07-22
Fix source remapping in computables to work when there are WITH bindings

msullivan:
I think this should unblock #2692.

### DIFF
--- a/edb/edgeql/compiler/setgen.py
+++ b/edb/edgeql/compiler/setgen.py
@@ -469,14 +469,22 @@ def compile_path(expr: qlast.Path, *, ctx: context.ContextLevel) -> irast.Set:
 
                 scope_set = scoped_set(scope_set, ctx=subctx)
 
-        mapped = ctx.view_map.get(path_tip.path_id)
-        if mapped is not None:
-            path_tip = new_set(
-                path_id=mapped.path_id,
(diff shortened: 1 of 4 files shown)


edb/edgeql/compiler/setgen.py
```diff
@@ -1401,7 +1409,10 @@ def newctx() -> Iterator[context.ContextLevel]:
             remapped_source = new_set_from_set(
                 rptr.source, rptr=rptr.source.rptr,
                 preserve_scope_ns=True, ctx=ctx)
-            subctx.view_map[inner_path_id] = remapped_source
+            key = inner_path_id.strip_namespace(inner_path_id.namespace)
+            assert key not in subctx.view_map
+            subctx.view_map[key] = (inner_path_id, remapped_source)
```
elprans, line 1414: Hmm. I wonder if using bare `path_id` here is safe.  Is it possible to construct a case where this would get overwritten incorrectly?  Sibling `WITH` bindings maybe?

msullivan: It looks like `view_map` is always *empty*, (created from the qlctx view_map), so that shouldn't be an issue

elprans (approved):

PR #2716, opened 2021-07-23
Use the new location for dev datadir in `edb inittestdb`

elprans:


fantix (approved):
Thanks! This is good as it matches the [CLI check](https://github.com/edgedb/edgedb-cli/pull/380/files#diff-1b33b1af1a8e35c4f324cac560774fc81865de0181472f33fd544f199e8cf931R55) so that this dev instance won't show up in `edgedb instance status`.

PR #2722, opened 2021-07-24
Adds missing `__slots__` for `edb.schema.objects.Field`

sobolevn:
Refs https://github.com/python/mypy/pull/10864


elprans (approved):
Thanks!

PR #2725, opened 2021-07-26
Fix notebook protocol version handling + improve error message when query contains params

### LINKED ISSUE #2618: Notebook queries failing with error: `'>=' not supported between instances of 'NoneType' and 'tuple'`
- EdgeDB Version: 1.0-beta.3+dev5686.d2021062412.g99e1f5eec.cv202106090000

Steps to Reproduce:

1. Setup tutorial database from edgedb/ui
2. Run a query with the notebook protocol, where the result is a set of objects:
```bash
$ curl --request POST \
>   --url http://localhost:5656/db/tutorial/notebook \
>   --header 'Content-Type: application/json' \
>   --data '{"queries":["SELECT User {name}"]}'
{"kind": "results", "results":[{"kind": "error", "error": ["Error", "'>=' not supported between instances of 'NoneType' and 'tuple'", {}]

curl --request POST \
>   --url http://localhost:5656/db/tutorial/notebook \
>   --header 'Content-Type: application/json' \
>   --data '{"queries":["SELECT User"]}'
{"kind": "results", "results":[{"kind": "error", "error": ["Error", "'>=' not supported between instances of 'NoneType' and 'tuple'", {}]

curl --request POST \
>   --url http://localhost:5656/db/tutorial/notebook \
>   --header 'Content-Type: application/json' \
>   --data '{"queries":["SELECT User.name"]}'
{"kind": "results", "results":[{"kind": "data", "data": ["AAAAAAAAAAAAAAAAAAABAQ==", "AgAAAAAAAAAAAAAAAAAAAQE=", "RAAAAA8AAQAAAAVBbGljZUQAAAAQAAEAAAAGQmlsbGllRAAAABEAAQAAAAdDYW1lcm9uRAAAAA4AAQAAAAREYW5h", "U0VMRUNU"]}]}

$ curl --request POST \
>   --url http://localhost:5656/db/tutorial/notebook \
>   --header 'Content-Type: application/json' \
>   --data '{"queries":["SELECT {1, 2, 3}"]}'
{"kind": "results", "results":[{"kind": "data", "data": ["AAAAAAAAAAAAAAAAAAABBQ==", "AgAAAAAAAAAAAAAAAAAAAQU=", "RAAAABIAAQAAAAgAAAAAAAAAAUQAAAASAAEAAAAIAAAAAAAAAAJEAAAAEgABAAAACAAAAAAAAAAD", "U0VMRUNU"]}]}
```

### LINKED ISSUE #2630: Update notebook protocol to return protocol version
Protocol version 0.11 introduces changes in the type descriptor structure, so the protocol version is now required to decode the codecs, which the notebook protocol currently doesn't return.

jaclarke:
- Pass protocol version to compiler from `compile_notebook` (Fixes #2618)
- Return protocol version in notebook response http header (Fixes #2630)
- Improve error message when query contains parameters (Fixes https://github.com/edgedb/edgedb.com/issues/205)

### DIFF
--- a/edb/server/protocol/protocol.pyx
+++ b/edb/server/protocol/protocol.pyx
@@ -53,6 +53,7 @@ cdef class HttpResponse:
     def __cinit__(self):
         self.status = HTTPStatus.OK
         self.content_type = b'text/plain'
+        self.headers = {}
         self.body = b''
         self.close_connection = False
 
@@ -187,6 +188,7 @@ cdef class HttpProtocol:
             b'1.0',
             b'400 Bad Request',
             b'text/plain',
+            {},
             f'{type(ex).__name__}: {ex}'.encode(),
             True)
 
@@ -203,7 +205,8 @@ cdef class HttpProtocol:
             self.transport.resume_reading()
 
     cdef _write(self, bytes req_version, bytes resp_status,
-                bytes content_type, bytes body, bint close_connection):
+                bytes content_type, dict headers, bytes body,
+                bint close_connection):
         if self.transport is None:
             return
         data = [
@@ -212,8 +215,14 @@ cdef class HttpProtocol:
             b'Content-Length: ', f'{len(body)}'.encode(), b'\r\n',
         ]
 
+        for key, value in headers.items():
+            data.append(f'{key}: {value}\r\n'.encode())
+
         if debug.flags.http_inject_cors:
             data.append(b'Access-Control-Allow-Origin: *\r\n')
+            if headers:
+                data.append(b'Access-Control-Expose-Headers: ' + \
+                    ', '.join(headers.keys()).encode() + b'\r\n')
 
         if close_connection:
             data.append(b'Connection: close\r\n')
@@ -228,6 +237,7 @@ cdef class HttpProtocol:
             request.version,
             f'{response.status.value} {response.status.phrase}'.encode(),
             response.content_type,
+            response.headers,
             response.body,
             response.close_connection)
 
--- a/edb/server/protocol/protocol.pyx
+++ b/edb/server/protocol/protocol.pyx
@@ -53,7 +53,7 @@ cdef class HttpResponse:
     def __cinit__(self):
         self.status = HTTPStatus.OK
         self.content_type = b'text/plain'
-        self.headers = {}
+        self.custom_headers = {}
         self.body = b''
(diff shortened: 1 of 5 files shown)


tests/test_http_notebook.py
```diff
@@ -58,6 +58,7 @@ def test_http_notebook_01(self):
             results,
             {
                 'kind': 'results',
+                'protocol_version': [0, 11],
```
1st1, line 61: Can you try this?

```suggestion
                'protocol_version': [0, ...],
```

would hate to update this test every time we bump the proto version

jaclarke: What should the ellipsis do here? It just shows an error that ellipsis isn't equal to 11.

elprans: Could we use HTTP headers instead of embedding in response body?

1st1: Yeah, using a header makes sense. So `EdgeDB-Protocol-Version: 0.11`? (Apparently `X-` prefix shouldn't be used anymore https://stackoverflow.com/questions/3561381/custom-http-headers-naming-conventions)

1st1: Also, @jaclarke, any issues with using a header?

@elprans can a custom HTTP header be stripped by some proxy? We need a protocol version to decode data, so this isn't some auxiliary stuff.

elprans: Stripping random response headers is super bad mojo, so I wouldn't worry about that.

jaclarke: > Also, @jaclarke, any issues with using a header?

Not that I can see, other than we need to send an extra CORS header to be able to access the custom header in the browser, but that's not really a problem. (https://developer.mozilla.org/en-US/docs/Glossary/CORS-safelisted_response_header)

edb/server/protocol/protocol.pyx
```diff
@@ -53,6 +53,7 @@ cdef class HttpResponse:
     def __cinit__(self):
         self.status = HTTPStatus.OK
         self.content_type = b'text/plain'
+        self.headers = {}
```
1st1, line 56:
```suggestion
        self.custom_headers = {}
```

1st1 (approved):
LGTM, please squash before merging.

PR #2728, opened 2021-07-26
Prohibit navigational nesting in UPDATE

elprans:
Currently, we allow the following `UPDATE` syntax:

```
UPDATE Foo
FILTER .foo_name = 'First'
SET {
   bar: {
       spam := true
   } FILTER .bar_name = 'Blah'
}
```

The intuitive interpretation is that it should update an object of name
`Blah` specifically linked to a `Foo` object of name `First`.  However,
currently this is what the statement gets transformed into:

```
UPDATE Foo
FILTER .foo_name = 'First'
SET {
   bar := (SELECT .bar {
       spam := true
   } FILTER .bar_name = 'Blah')
}
```

It's the standard `SELECT` transform, but it's quite wrong for `UPDATE`,
since instead of updating a single linked object, it actually _replaces_
the entire linked set `bar`.

We intend to support the nested `UPDATE` case properly soon, but for now
we should just ban the syntax.

While at it, fix a bug whereby UPDATEs that reassign a link to an
overlapping set don't get the link properties erased properly.

### DIFF
--- a/tests/test_dump01.py
+++ b/tests/test_dump01.py
@@ -1821,6 +1821,8 @@ class DumpTestCaseMixin:
             ],
         )
 
+        """XXX: uncomment the below once direct link property updates are
+                implemented
         # validate read-only
         with self.assertRaisesRegex(
                 edgedb.QueryError,
@@ -1872,6 +1874,7 @@ class DumpTestCaseMixin:
                         rol1: {@rolp11 := 1},
                     };
                     ''')
+        """
 
 
 class TestDump01(tb.StableDumpTestCase, DumpTestCaseMixin):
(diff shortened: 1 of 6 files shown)


tests/test_dump01.py
```diff
@@ -1821,6 +1821,8 @@ async def _ensure_schema_data_integrity(self):
             ],
         )
 
+        """XXX: uncomment the below once direct link property updates are
+                implemented
```
1st1, line 1825: Do we have an issue for that?

1st1 (approved):

msullivan (approved):

PR #2730, opened 2021-07-27
Changelog for beta 3 release.

vpetrovykh:
I am possibly missing some of the bindings changes as my understanding is that the major one was the RFC 1008, but perhaps I should include other stuff. I'll give it another pass.

Also the ordering and the examples might need tweaking.

However the bulk of changes should be there.


docs/changelog/1_0_b3.rst
```diff
@@ -0,0 +1,203 @@
...
+      - Get various metadata about the project
+    * - ``project upgrade``
+      - Upgrade EdgeDB instance used for the current project
+    * - ``self upgrade``
+      - Upgrade this edgedb binary
+    * - ``self install``
+      - Install server
+
+* Implemented `RFC 1008: TLS and ALPN <rfc1008_>`_.
+* Add an option to generate self-signed certificates.
```
elprans, line 188: self-signed certs are part of TLS and ALPN work, no need to break out a separate item.  Also, I'd add a blurb from the motivation section of the RFC here to explain why we are adding TLS: it is to properly support protocol selection, but also proper transport security by default.

vpetrovykh: I've reworked the TLS section.

docs/cli/edgedb_server/edgedb_server_reset_password.rst
```diff
@@ -1,47 +0,0 @@
-.. _ref_cli_edgedb_server_reset_password:
```
elprans, line 1: This should be kept (although renamed to `instance reset-auth`)

vpetrovykh:
There should be no drastic mismatches between the docs and the changelog now. However, there are still a few places where the pre-project way of setting up instances is used. I'll make a separate PR to update those as they don't really impact the changelog.

docs/cli/edgedb_server/edgedb_server_reset_password.rst
vpetrovykh: Yeah, I just wanted to confirm the new name. I'll add it back in a moment.

1st1:
Victor, can you split the work into two commits: one that fixes the docs, and one that adds the changelog?

vpetrovykh:
> Victor, can you split the work into two commits: one that fixes the docs, and one that adds the changelog?

Sure, but the changelog will depend on the updated docs (because of all the `:ref:` and other cross-links). Should I just have them as two commits for this one PR?

docs/changelog/1_0_b3.rst
```diff
@@ -0,0 +1,231 @@
...
+  .. code-block:: edgeql
+
+    SELECT Color.Red;
+
+  is equivalent to
+
+  .. code-block:: edgeql
+
+    SELECT <Color>'Red';
+* Allow removal and reordering of :eql:type:`enum` elements
```
1st1, line 61: Add an empty newline before this line

1st1 (approved):

docs/changelog/1_0_b3.rst
vpetrovykh: Added.

vpetrovykh:
I've split this into 3 commits under the same PR.

docs/cli/edgedb_server/edgedb_server_reset_password.rst
vpetrovykh: I've added it back as `instance reset-auth`

PR #2732, opened 2021-07-27
Fix some materialization related issues

msullivan:
The main thing here is to rework the selection of which materialized
set is compiled to ensure that we always compile one with an
appropriate rptr source, if it is a computable.

Additionally we need to store materialized computables on a non-SELECT
shape with the shape statement, not with whatever edgeql statement is
surrounding it, which requires shifting how things are tracked.

These reworkings fixed some volatility cross product errors we were
missing.

Modify the test_volatility tests to also insert typenames, and fix
some issues that shook out.

As part of dealing with the changes to non-SELECT shape compilation, I
needed to make the visibility checks in cardinality inference and
subquery compilation be done from the parent *branch* instead of the
parent *fence*. I am not super confident about this change.

### DIFF
--- a/edb/edgeql/compiler/stmt.py
+++ b/edb/edgeql/compiler/stmt.py
@@ -1049,9 +1049,9 @@ def compile_Shape(
         shape: qlast.Shape, *, ctx: context.ContextLevel) -> irast.Set:
     shape_expr = shape.expr or qlutils.ANONYMOUS_SHAPE_EXPR
     with ctx.new() as subctx:
-        subctx.qlstmt = shape
+        subctx.qlstmt = astutils.ensure_qlstmt(shape)
         subctx.stmt = stmt = irast.SelectStmt()
-        ctx.env.compiled_stmts[shape] = stmt
+        ctx.env.compiled_stmts[subctx.qlstmt] = stmt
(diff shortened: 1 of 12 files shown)


edb/edgeql/compiler/stmt.py
```diff
@@ -1049,21 +1049,38 @@ def compile_Shape(
...
-    if not isinstance(expr_stype, s_objtypes.ObjectType):
-        raise errors.QueryError(
-            f'shapes cannot be applied to '
-            f'{expr_stype.get_verbosename(ctx.env.schema)}',
-            context=shape.context,
-        )
-    view_type = viewgen.process_view(
-        stype=expr_stype, path_id=expr.path_id,
-        elements=shape.elements, parser_context=shape.context, ctx=ctx)
+        subctx.qlstmt = shape
```
elprans, line 1052: Use `astutils.ensure_qlstmt` here (and use the result as a key below for `compiled_stmts`) and you should avoid the need for a `Union` in the dictionaries, since it seems the value is only used as a key for lookups.

msullivan: Good call

elprans (approved):

Issue #2696, opened 2021-07-17
Trying to update listen_addresses triggers ISE: address already in use


Fogapod:
- EdgeDB Version: 1.0-beta.2+ga7130d5c7.cv202104290000 
- OS Version: Ubuntu 20.04

Steps to Reproduce:

1. create server instance
2. edgedb -I SERVER_NAME configure set listen_addresses 127.0.0.1 ::1 0.0.0.0

```python

edgedb error: ERROR: InternalServerError: [Errno 98] error while attempting to bind on address ('0.0.0.0', 10700): address already in use
  Hint: This is most likely a bug in EdgeDB. Please consider opening an issue ticket at https://github.com/edgedb/edgedb/issues/new?template=bug_report.md
  Server traceback:
      Traceback (most recent call last):
        File "edb/server/protocol/binary.pyx", line 1694, in edb.server.protocol.binary.EdgeConnection.main
        File "edb/server/protocol/binary.pyx", line 950, in simple_query
        File "edb/server/protocol/binary.pyx", line 1045, in _simple_query
        File "edb/server/protocol/binary.pyx", line 1002, in edb.server.protocol.binary.EdgeConnection._simple_query
        File "edb/server/protocol/binary.pyx", line 1405, in _execute_system_config
        File "edb/server/dbview/dbview.pyx", line 515, in apply_config_ops
        File "edb/server/dbview/dbview.pyx", line 649, in apply_system_config_op
        File "/usr/lib/x86_64-linux-gnu/edgedb-server-1-beta2/lib/python3.9/site-packages/edb/server/server.py", line 629, in _on_system_config_set
          await self._restart_servers_new_addr(value, self._listen_port)
        File "/usr/lib/x86_64-linux-gnu/edgedb-server-1-beta2/lib/python3.9/site-packages/edb/server/server.py", line 573, in _restart_servers_new_addr
          self._servers, _ = await self._start_servers(nethost, netport)
        File "/usr/lib/x86_64-linux-gnu/edgedb-server-1-beta2/lib/python3.9/site-packages/edb/server/server.py", line 786, in _start_servers
          tcp_srv = await self._loop.create_server(
        File "uvloop/loop.pyx", line 1768, in create_server
      OSError: [Errno 98] error while attempting to bind on address ('0.0.0.0', 10700): address already in use
```

Despite the error, value is changed in config:
```
edgedb> SELECT cfg::DatabaseConfig.listen_addresses;
[...]

PR #2733, opened 2021-07-27
Fix certgen issue with multiple hosts

fantix:
If `listen_addresses` was set to multiple hosts, the certgen would fail. This PR fixes this issue.

Also allowed to specify `--bind-address` multiple times.

### DIFF
--- a/edb/server/main.py
+++ b/edb/server/main.py
@@ -233,7 +233,7 @@ async def _run_server(
 
 
 def _generate_cert(
-    cert_dir: pathlib.Path, listen_host: str
+    cert_dir: pathlib.Path, listen_host: Union[str, Iterable]
 ) -> Tuple[pathlib.Path, Optional[pathlib.Path]]:
     logger.info("Generating self-signed TLS certificate.")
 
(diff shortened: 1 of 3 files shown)


edb/server/main.py
```diff
@@ -233,7 +233,7 @@ async def _run_server(
 
 
 def _generate_cert(
-    cert_dir: pathlib.Path, listen_host: str
+    cert_dir: pathlib.Path, listen_host: Union[str, Iterable]
```
1st1, line 236: Maybe fix the caller code to always pass a `list[str]`?

fantix: Sure thing! I would also like to update `edgedb-server` to take multiple `-I` values, e.g. `edgedb-server -I 127.0.0.1 -I 192.168.2.22`, because one could anyway set the system config `listen_addresses` to multiple hosts.

1st1:
I don't see how this PR deals with port=auto when multiple host names are specified. As discussed in https://github.com/edgedb/edgedb/issues/2696 -- let's do this:

1. If port=auto was specified, the server should start without creating an admin socket. There's no point in it anyways, since admin socket is only used by the CLI, and port=auto is only used for CI/test harnesses.
2. If `--emit-server-status` was specified, the server should output a list of tuples for the hosts/ports it listens on.

fantix:
> I don't see how this PR deals with port=auto when multiple host names are specified

Oh this PR was only intended to address the cert generation issue. I have another PR for #2696

fantix:
Merged in #2737.

PR #2734, opened 2021-07-27
Prune tmpns out of the scope tree after compilation

msullivan:
This is mostly just to improve the debugging experience


elprans (approved):
The ghost of temp namespaces is back! :-)

PR #2735, opened 2021-07-27
Reduce the size of pools on low memory systems

elprans:
When available system RAM is tight (<1GiB), reduce the sizes of the
compiler pool and the PostgreSQL connection pool to a minimum (2 and 4).

### DIFF
--- a/edb/server/args.py
+++ b/edb/server/args.py
@@ -38,7 +38,8 @@ from . import buildmeta
 from . import defines
 
 
-BYTES_OF_MEM_PER_CONN = 100 * 1024 * 1024  # 100MiB
+MIB = 1024 * 1024
+RAM_MIB_PER_CONN = 100
 TLS_CERT_FILE_NAME = "edbtlscert.pem"
 TLS_KEY_FILE_NAME = "edbprivkey.pem"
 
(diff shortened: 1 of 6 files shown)


fantix (approved):

edb/server/args.py
```diff
@@ -38,7 +38,8 @@
 from . import defines
 
 
-BYTES_OF_MEM_PER_CONN = 100 * 1024 * 1024  # 100MiB
+MIB = 1024 * 1024
+RAM_MB_PER_CONN = 100
```
1st1, line 42:
```suggestion
RAM_MIB_PER_CONN = 100
```

Or just rename `MIB` to `MB`

edb/server/pgcluster.py
```diff
@@ -470,6 +470,7 @@ def start(self, wait=60, *, server_settings=None, **opts):
             # we're using superuser only now (so all connections available),
             # and we don't support reserving connections for now
             'max_connections': str(self._instance_params.max_connections),
+            'max_locks_per_transaction': 256,
```
1st1, line 473: Can you add a comment explaining why we configure this and why it's set to 256?

1st1 (approved):
lgtm overall, there are a couple nit comments though

PR #2739, opened 2021-07-28
Remove `--admin` usage from the docs.

vpetrovykh:
Rewrite the passwordless connection creation example without the use of
deprecated `--admin`. Add that example to the `CONFIGRE` documentaion as
well.

### DIFF
--- a/docs/admin/setup.rst
+++ b/docs/admin/setup.rst
@@ -68,5 +68,5 @@ method with the ``Trust`` method:
 
 .. code-block:: bash
 
-    $ edgedb -I <instance-name> --admin configure insert Auth \
-             --method=Trust --priority=0
+    $ edgedb -I <instance-name> -c "CONFIGURE INSTANCE INSERT \
+        Auth {priority := 0, method := (INSERT Trust)}"
--- a/docs/edgeql/admin/configure.rst
+++ b/docs/edgeql/admin/configure.rst
@@ -80,6 +80,16 @@ Set the same parameter, but for the current database:
 
     CONFIGURE CURRENT DATABASE SET query_work_mem := '4MB';
 
+Add a Trust authentication method for "my_user":
+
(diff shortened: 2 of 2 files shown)


docs/admin/setup.rst
```diff
@@ -57,16 +57,3 @@ In JavaScript:
...
------------------------------------
-
-If you are doing testing and don't want to bother with passwords or other
-authentication, it is possible to override the default password authentication
-method with the ``Trust`` method:
-
-.. code-block:: bash
-
-    $ edgedb -I <instance-name> --admin configure insert Auth \
-             --method=Trust --priority=0
```
1st1, line 72: I think we should keep this, just replace the bash line with EdgeQL?

elprans: Yeah, drop `--admin` and replace `configure` with `query "CONFIGURE INSTANCE ..."`

1st1: Do we even need to replace `configure` with EdgeQL? I mean if you know the password this all should go just fine even without `--admin`

elprans: Yeah, the motivation to use EdgeQL in the example is to avoid the CLI `configure` arcana, but I'm fine either way.

vpetrovykh: Actually, now that I read the whole "Administration" section I wonder what is it trying to target and whether we should be mentioning `edgedb project` anywhere here instead of the manual instance setup.

1st1: > what is it trying to target

People who need to manage live or otherwise complicated/manual deployments of EdgeDB.

> we should be mentioning edgedb project

I guess we should add a note to the top of the page that most of this stuff isn't needed during regular EdgeDB development. `edgedb project` covers 99% of those needs.

1st1 (approved):

PR #2741, opened 2021-07-28
Allow compiler pool to be of size 1

elprans:
Currently, the process pool code is asserting the pool size to be
greater than 1 for no apparent reason.  Take the opportunity to
editorialize the nearby log messages a bit.


fantix (approved):
Good for me

1st1 (approved):

PR #2744, opened 2021-07-29
More properly fix issue with introspection defaults when prop used twice

msullivan:
This was worked around in #2667, but the main underlying issue was
that the pull_namespace on the type_cte meant that one Pointer
accessing required would cause it to eagerly get pulled up into other
queries.

Fix that, and revert the workaround from #2667.


elprans (approved):
:+1:

PR #2748, opened 2021-07-29
Turn on AST `__init__` synthesis for edgeql AST

msullivan:
I needed to fix a defer related bug in the plugin, and I felt that it
was a little too harsh to ban *all* container construction defaults,
so I added a `container_factory` sentinel object for when we still
want the behavior. (Probably we want eventually want to drop all the
magic use of type annotations to produce the defaults and specify them
explicitly?)

### DIFF
--- a/edb/common/ast/base.py
+++ b/edb/common/ast/base.py
@@ -50,6 +50,9 @@ class Field:
         self.meta = field_meta
 
 
+container_factory: Any = object()
+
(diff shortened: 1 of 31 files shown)


elprans (approved):
> and I felt that it was a little too harsh to ban all container construction defaults

Why?  Seems like we're almost there.

PR #2752, opened 2021-07-30
Add --allow-unsecure-http-clients and --instance-name params...

1st1:
... to the EdgeDB server.

### DIFF
--- a/edb/server/args.py
+++ b/edb/server/args.py
@@ -99,6 +99,8 @@ class ServerConfig(NamedTuple):
     allow_insecure_binary_clients: bool
     allow_insecure_http_clients: bool
 
+    instance_name: Optional[str]
+
 
 class PathPath(click.Path):
     name = 'path'
@@ -379,6 +381,11 @@ _server_options = [
         envvar="EDGEDB_SERVER_ALLOW_INSECURE_HTTP_CLIENTS",
         type=bool, is_flag=True, hidden=True,
         help='Allow non-TLS client HTTP connections.'),
+    click.option(
+        '--instance-name',
+        envvar="EDGEDB_SERVER_INSTANCE_NAME",
+        type=str, default=None, hidden=True,
+        help='Server instance name.'),
     click.option(
         '--version', is_flag=True,
         help='Show the version and exit.')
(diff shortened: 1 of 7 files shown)


edb/server/args.py
```diff
@@ -95,7 +95,9 @@ class ServerConfig(NamedTuple):
     tls_cert_file: Optional[pathlib.Path]
     tls_key_file: Optional[pathlib.Path]
     generate_self_signed_cert: bool
-    allow_cleartext_connections: bool
+
+    allow_unsecure_binary_clients: bool
```
elprans, line 99: insecure?

edb/server/protocol/protocol.pyx
```diff
@@ -133,8 +136,9 @@ cdef class HttpProtocol:
                 return
             else:
                 # HTTP.
+                print('!!!', self.allow_unsecure_http_clients)
```
elprans, line 139: Forgotten debug print

1st1: fixed

edb/server/args.py
1st1: yeah, fixed

elprans (approved):

tests/test_server_ops.py
```diff
@@ -403,6 +404,12 @@ async def test_server_ops_no_cleartext(self):
                 self.assertIn('location', resp_headers)
                 self.assertTrue(
                     resp_headers['location'].startswith('https://'))
+
+                self.assertIn('strict-transport-security', resp_headers)
+                # By default we enforce HTTPS via HSTS on all routes.
+                self.assertEqual(
+                    resp_headers['strict-transport-security'],
+                    'max-age=31536000; includesubdomains; preload')
```
elprans, line 412: I don't think we should force `includeSubdomains` and `preload` here. We have no business in enforcing HSTS opinions on the entire domain on which EdgeDB is running.

1st1: good catch, makes sense

1st1: Fixed

elprans (approved):

PR #2753, opened 2021-07-30
Make `dispatch.compile` always return irast.Set

msullivan:
In practice it *did* already do this, except for ConfigSet and
ConfigReset.  Those required a little fiddling.

This lets us drop a lot of calls to `ensure_set`.

### DIFF
--- a/edb/edgeql/compiler/expr.py
+++ b/edb/edgeql/compiler/expr.py
@@ -22,7 +22,7 @@
 
 from __future__ import annotations
 
-import typing
+from typing import *
 
 from edb import errors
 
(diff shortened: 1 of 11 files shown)


1st1 (commented):
Nice!

edb/edgeql/compiler/expr.py
```diff
@@ -576,7 +572,7 @@ def compile_Introspect(
 @dispatch.compile.register(qlast.Indirection)
 def compile_Indirection(
         expr: qlast.Indirection, *, ctx: context.ContextLevel) -> irast.Set:
-    node = dispatch.compile(expr.arg, ctx=ctx)
+    node: Union[irast.Set, irast.Expr] = dispatch.compile(expr.arg, ctx=ctx)
```
elprans, line 575: Is this annotation still needed?

edb/edgeql/compiler/setgen.py
```diff
@@ -1102,6 +1099,12 @@ def scoped_set(
     return ir_set
 
 
+def lol_set(
```
elprans, line 1102: What is this for?

msullivan: Oh oops, that was something I used while doing the migration

elprans (approved):

edb/edgeql/compiler/expr.py
msullivan: Yeah, it's reassigned to later

PR #2754, opened 2021-07-30
Ensure a bare collection type is created when creating collection aliases

elprans:
Currently, when an expression alias of a collection type is defined we
don't explicitly create an anonymous collection type corresponding to
the type returned by the alias expression.  This would become
problematic once computed pointers will become defined in terms of an
anonymous expression alias, because queries and backend table
definitions expect naked collection types.

For example, `alias Foo := (1, 2)` will now generate the following
pseudo-delta tree:

    CreateAlias("Foo@alias")
      CreateTupleExprAlias("Foo")
         CreateTuple("tuple<int64, int64>")

### DIFF
--- a/edb/schema/expraliases.py
+++ b/edb/schema/expraliases.py
@@ -326,18 +326,20 @@ class DeleteAlias(
 ):
     astnode = qlast.DropAlias
 
-    def _delete_begin(
+    def _canonicalize(
         self,
         schema: s_schema.Schema,
         context: sd.CommandContext,
-    ) -> s_schema.Schema:
-        if not context.canonical:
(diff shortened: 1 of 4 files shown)


msullivan (approved):
Looks good.

It might be worth adding a test that verifies that the expected type is making it into the schema, but it also might be too marginal to bother since it will be pretty implied by the follow-up stuff.

elprans:
> It might be worth adding a test that verifies that the expected type is making it into the schema

The test I modified in this PR does exactly that (albeit by counting the tuple types, not checking the names directly).

msullivan:
Yeah, I had meant more explicitly making sure it is the "right" type

PR #2756, opened 2021-07-30
Provide a default protocol_version in new_compiler_context

msullivan:
sertypes now consults protocol_version in certain cases, and so
having a None protocol_version can cause sertypes to crash.
(This came up in edgedb-perf.)

Solve this by adding a protocol_version argument to
new_compiler_context and passing it in from callsites.

My original plan was to make protocol_version default to
CURRENT_PROTOCOL, but `edb.server.compiler` is imported from within
`edb.server.protocol`, and so doing a top-level import caused some
cycle-induced errors, and I did this instead.

It would be possible to stick with the original plan by either doing
the import inside of new_compiler_context or moving CURRENT_PROTOCOL.

I could switch to one of those if someone prefers.


msullivan:
Oh, actually, I think avoiding the cycle is super easy, so I am going to do that?

fmoor (approved):
LGTM

elprans:
> having a None protocol_version can cause sertypes to crash

Hm. edgedb-perf works for me.  How do I reproduce this?

msullivan:
> > having a None protocol_version can cause sertypes to crash
> 
> Hm. edgedb-perf works for me. How do I reproduce this?

Try `compile_edgeql_script_webapp_insert_review`?

elprans (changes requested):
I'm not happy with how protocol version is proliferating here.  Let's set `CURRENT_PROTOCOL` as the default in `CompileContext`.  It should be possible to move the constants to `protocol/__init__.py` and `import` them in pyx files instead of relying on `pxi`.

msullivan:
Alright, I moved them into `defines.py`, which dodges all the circularity issues and is a reasonable place anyway

elprans (approved):

PR #2762, opened 2021-08-03
Do some optimizations of AST.__init__

msullivan:
The main thing here is not iterating over the full fields list, but
instead iterating over the list of attributes with factory defaults
and the kwargs dict separately.

This gets us benchmark speedups in the 2-5% range, mostly.
(We already picked up 1% from the previous AST changes.)

### DIFF
--- a/edb/common/ast/base.py
+++ b/edb/common/ast/base.py
@@ -236,8 +236,8 @@ class AST:
         d = self.__dict__
 
         for field_name, factory in self._field_factories:
-            value = factory()
-            d[field_name] = value
+            if field_name not in kwargs:
+                d[field_name] = factory()
 
         d.update(kwargs)
 
--- a/edb/common/ast/base.py
+++ b/edb/common/ast/base.py
@@ -232,14 +232,12 @@ class AST:
                 f'cannot instantiate abstract AST node '
                 f'{self.__class__.__name__!r}')
 
-        # Put things directly into __dict__, as an optimization
-        d = self.__dict__
-
+        # Make kwargs directly into our __dict__
         for field_name, factory in self._field_factories:
             if field_name not in kwargs:
-                d[field_name] = factory()
+                kwargs[field_name] = factory()
 
-        d.update(kwargs)
+        self.__dict__ = kwargs
 
         should_check_types = __debug__ and _check_type is _check_type_real
         if should_check_types:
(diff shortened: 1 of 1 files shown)


edb/common/ast/base.py
```diff
@@ -219,24 +232,19 @@ def __init__(self, **kwargs):
...
-                value = field.factory()
-            else:
-                value = field.default
+        # Put things directly into __dict__, as an optimization
+        d = self.__dict__
 
-            if should_check_types:
-                self.check_field_type(field, value)
+        for field_name, factory in self._field_factories:
+            value = factory()
```
1st1, line 239: I'm not sure if unconditionally calling `factory()` is better than checking if `field_name not in kwargs` first. I mean it can be slightly faster until someone adds a more expensive factory.

1st1: Can you add an if check and see if it costs too much? If it does we might as well compile this file with Cython.

msullivan: It's a tiny bit worse but it's pretty marginal

msullivan: Oh, but if we do that, then we can actually just make the updates directly to `kwargs`, then stick `kwargs` into `__dict__`, which gets us a bit more of a speedup

1st1: Even better

edb/common/ast/base.py
```diff
@@ -219,24 +232,17 @@ def __init__(self, **kwargs):
...
-            else:
-                value = field.default
+        # Make kwargs directly into our __dict__
+        for field_name, factory in self._field_factories:
+            if field_name not in kwargs:
+                kwargs[field_name] = factory()
 
-            if should_check_types:
-                self.check_field_type(field, value)
+        self.__dict__ = kwargs
```
1st1, line 240: Can you move this line after the `if should_check_types` line?

1st1 (approved):

edb/common/ast/base.py
```diff
@@ -219,24 +232,17 @@ def __init__(self, **kwargs):
...
+        for field_name, factory in self._field_factories:
+            if field_name not in kwargs:
+                kwargs[field_name] = factory()
 
-            if should_check_types:
-                self.check_field_type(field, value)
+        should_check_types = __debug__ and _check_type is _check_type_real
+        if should_check_types:
+            for k, v in kwargs.items():
+                self.check_field_type(self._fields[k], v)
```
elprans, line 243: OT: now that this all is properly typed, should we get rid of runtime type checking code altogether?  It's behind a debug flag anyway.

elprans (approved):

PR #2763, opened 2021-08-03
Speed up server startup by caching _get_pickled_init_args

msullivan:
On my machine this speeds up startup on my normal collection of test
instances from ~9.5s to ~5.5s.

### DIFF
--- a/edb/server/compiler_pool/pool.py
+++ b/edb/server/compiler_pool/pool.py
@@ -250,6 +250,7 @@ class Pool(amsg.ServerProtocol):
         )
         return init_args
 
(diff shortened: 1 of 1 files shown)


1st1 (approved):
Good to go, good catch

1st1:
An even better fix: fold `_get_pickled_init_args()` into `_get_init_args()`. It's easier to reason about that way.

1st1:
TIL: https://docs.python.org/3/library/functools.html?highlight=functools#functools.cache

elprans (approved):

PR #2769, opened 2021-08-05
Update the make scripts that use the dev server.

vpetrovykh:
Use a subprocess to send queries to the dev server rather than rely on
connecting to a specific port or instance.


edb/tools/gen_cast_table.py
```diff
@@ -161,14 +147,28 @@ def gen_cast_table():
...
-    con = None
     try:
-        con = edgedb.connect(user=edgedb_defines.EDGEDB_SUPERUSER,
-                             database=edgedb_defines.EDGEDB_SUPERUSER_DB,
-                             port=5656)
-        main(con)
+        res = subprocess.run([
+            'edb',
+            'cli',
+            '--json',
```
elprans, line 154: This would shortly become `edb cli query --json`, so let's save ourselves an extra PR, change it here and wait until that lands.

vpetrovykh: Sure, I'll wait and move the `--json` flag.

vpetrovykh: I've amended the code to use `edb cli query -Fjson` as per latest CLI implementation.

elprans (approved):

PR #2771, opened 2021-08-05
Change the path to _localdev

fantix:
The _localdev directory is now side-by-side to the data_dir containing
CLI-managed local instances, instead of inside the data_dir, so that the
CLI could be lifted from checking special cases for _localdev.

PR #2779, opened 2021-08-06
Sync inittestdb data-dir with that of devmode server (again)

elprans:
The location has been changed in #2771, but `inittestdb` was overlooked
again, so factor out the directory logic into a helper in `devmode` to
make sure this doesn't happen the next time we change the path :-)


fantix (approved):
Oh yes, forgot that again, thanks!

PR #2789, opened 2021-08-09
Make find_children search into nested containers

msullivan:
This is important because otherwise it misses shapes. This somehow
hasn't caused any visible bugs yet but caused trouble in a pending PR
I have.

While at it, substantially clean up the function and drop support for the unused child_traverse.


1st1 (commented):
Can we add a test for this?

elprans (approved):

PR #2790, opened 2021-08-10
Change how JSON scalars are reflected into GraphQL.

vpetrovykh:
`std::json` is now reflected into a custom JSON scalar type in Graphql.
This is an escape hatch so the output is rendered as seamless JSON (not
as a string encoding JSON). The input values can only be supplied as a
variable at the moment (because of issues of parsing and validating
GraphQL types otherwise).

### DIFF
--- a/edb/graphql/translator.py
+++ b/edb/graphql/translator.py
@@ -1210,12 +1210,9 @@ class GraphQLTranslator:
             val = self.visit(node)
 
             if target.is_json:
-                # JSON must use a converter function and not a cast
-
-                return qlast.FunctionCall(
-                    func='to_json',
-                    args=[val]
-                )
+                # JSON can only come as a variable and will already be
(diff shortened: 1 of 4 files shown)


vpetrovykh:
In theory we could even hack things so that we accept JSON literals in the GraphQL query directly, but that would need some rework around variable extraction (for GraphQL) as the boundaries for variables are more fuzzy and the types of variables are impossible to guess without schema knowledge.

elprans (approved):
LGTM

vpetrovykh:
I've added a formal test for passing simple (boolean, number, string, null) as well as composite JSON values as variables.

1st1 (approved):

1st1:
Don't forget to update the changelog, this item deserves a big mention with an explanation.

PR #2795, opened 2021-08-10
Fix deprecated usages of SSLContext

elprans:
Fix `SSLContext`-related deprecation warnings by setting it up with
explicit protocol/purpose flags.


fantix (approved):
Thanks! Looks neat.

PR #2796, opened 2021-08-10
Materialize references to *all* visible sets from inside pointers

msullivan:
There are still a couple of related-seeming issues, one of which is
pretty scary (test_edgeql_scope_source_rebind_01), so we're not done
yet, but this fixes a bunch.

### DIFF
--- a/edb/edgeql/compiler/viewgen.py
+++ b/edb/edgeql/compiler/viewgen.py
@@ -582,7 +582,7 @@ def _normalize_view_ptr_expr(
                 is_insert=is_insert, is_update=is_update, ctx=ctx)
             materialized = setgen.should_materialize(
                 irexpr, ptrcls=ptrcls,
-                binding_pessimism=True, skipped_bindings={path_id},
+                materialize_visible=True, skipped_bindings={path_id},
                 ctx=ctx)
             ptr_target = inference.infer_type(irexpr, ctx.env)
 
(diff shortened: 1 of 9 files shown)


edb/edgeql/compiler/viewgen.py
```diff
@@ -1313,7 +1313,7 @@ def _compile_view_shapes_in_set(
             element = setgen.extend_path(
                 path_tip,
                 ptr,
-                unnest_fence=True,
+                unnest_fence=False,
```
elprans, line 1316: Hm.  Curious why is this needed and why doesn't this trip `test_edgeql_ir_scope_tree_bad_03`?

msullivan: Having an `unnest_fence` there was causing spurious failures in cases like `test_edgeql_scope_ref_outer_03` and the new `test_edgeql_scope_ref_outer_07`, where we define an alias that refers to an outer scope and then use the alias in a semi-joiny way.

It doesn't trip `test_edgeql_ir_scope_tree_bad_03` because we use an unnest fence when normalizing ptr exprs

msullivan: Actually I should just drop the `unnest_fence` argument altogether if I'm doing this, too

elprans: Oh, that's a nice cleanup!

elprans (approved):

PR #2803, opened 2021-08-11
Fix bug with computables and rebinding

msullivan:
The issue was that a computable reference on a rebound and filtered
set was referring to the original set. The idea of the fix is that in
`_get_computable_ctx`, we need to always use the
`inner_source_path_id` when setting up the remapping.

The particular handling of `subns` I'm not totally sure about to be
honest. This is the thing that most matched the original and passed
the tests.


elprans (approved):
Seems OK to me.

PR #2806, opened 2021-08-12
Get rid of `new_event_loop` mess in cluster/pgcluster

elprans:
Currently, `edgedb-server` and `postgres` harnesses in `server.cluster`
and `server.pgcluster` expose sync APIs for things like `init()` and
`start()` and then rather unceremoniously create new asyncio event loops
to actually be able to do I/O with the started server reasonably.  This
is obviously a problem if an event loop is already present in the
process, as evidenced by the contortions in some of the `test_server_ops_`
tests.  More importantly, though, sync API in `Cluster` prevents the use
of asyncio subprocesses, which I'd like to use to channel Postgres logs
properly.

In this patch I went over the call stacks that currently do the
`new_event_loop` stuff, made them `async` and threw away a bunch
of hacks in the process.  Also, I had to properly type-annotate the
cluster code, because I can't make myself do refactoring without the
help of a typechecker anymore.

### DIFF
--- a/edb/server/cluster.py
+++ b/edb/server/cluster.py
@@ -18,17 +18,19 @@
 
 
 from __future__ import annotations
+from typing import *
 
 import asyncio
 import json
 import os
+import pathlib
 import socket
 import subprocess
 import sys
 import tempfile
 import time
-import typing
 
+import asyncpg
 import edgedb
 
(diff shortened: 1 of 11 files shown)


1st1 (approved):
lgtm

PR #2807, opened 2021-08-12
Update the dirver's versions mentioned in changelog.

vpetrovykh:


docs/changelog/1_0_b3.rst
```diff
@@ -255,12 +255,12 @@ instance information, so all the bindings need to be updated to their
 latest versions in order to properly work with this release.
 
 * Implement `RFC 1008 <rfc1008_>`_ features for `edgedb-python
-  <https://github.com/edgedb/edgedb-python>`_ and release v0.16.0
+  <https://github.com/edgedb/edgedb-python>`_ and release v0.17.0
```
1st1, line 258:
```suggestion
  <https://github.com/edgedb/edgedb-python>`_ and release v0.17.x
```

It's already 0.17.2 and will continue to be a moving target

vpetrovykh: I'll update the python version to v0.17.x

According to what I just pulled I can see that edgedb-js v0.15.0 and that's already in the changelog.

vpetrovykh: And I've just noticed the edgedb-js release, so I updated that to v0.15.x

1st1 (approved):

Issue #2819, opened 2021-08-16
Use asyncio subprocesses for Postgres cluster management 


elprans:
Currently we use plain `subprocess` when running various PostgreSQL binaries, which makes it difficult to work with subprocess I/O streams, including logging.  Switch to `asyncio` subprocess API to fix this.

PR #2824, opened 2021-08-17
Use asyncio subprocess API in PostgreSQL cluster management

elprans:
This switches all uses of `subprocess` in `pgcluster` to
`asyncio.create_subprocess_exec` and sets up consistent capturing and
logging of subprocess standard streams.

Some slight massaging of log formatting is also here: timestamp is now
formatted as proper ISO date, and logger names are a darker shade of
grey.

Fixes: #2819

image

### DIFF
--- a/edb/server/pgcluster.py
+++ b/edb/server/pgcluster.py
@@ -28,15 +28,15 @@ import os
 import os.path
 import pathlib
 import re
+import shlex
 import shutil
-import subprocess
-import sys
-import tempfile
+import textwrap
 import time
 
 import asyncpg
 from asyncpg import serverversion
 
+from edb.common import supervisor
 from edb.common import uuidgen
 
 from edb.server import buildmeta
@@ -46,8 +46,12 @@ from edb.pgsql import common as pgcommon
 from . import pgconnparams
 
 
-logger = logging.getLogger(__name__)
-
+logger = logging.getLogger('edb.pgcluster')
+pg_dump_logger = logging.getLogger('pg_dump')
+pg_ctl_logger = logging.getLogger('pg_ctl')
+pg_config_logger = logging.getLogger('pg_config')
+initdb_logger = logging.getLogger('initdb')
+postgres_logger = logging.getLogger('postgres')
 
 get_database_backend_name = pgcommon.get_database_backend_name
 get_role_backend_name = pgcommon.get_role_backend_name
@@ -138,13 +142,13 @@ class BaseCluster:
         ] = None
         self._default_session_auth: Optional[str] = None
         self._pg_config_path = pg_config_path
+        self._pg_config_data: Dict[str, str] = {}
         self._pg_bin_dir: Optional[str] = None
         if instance_params is None:
             self._instance_params = (
                 get_default_runtime_params().instance_params)
         else:
             self._instance_params = instance_params
-        self._init_env()
 
     def get_db_name(self, db_name: str) -> str:
         return get_database_backend_name(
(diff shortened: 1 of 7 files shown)


edb/server/pgcluster.py
```diff
@@ -1104,3 +985,220 @@ async def _get_reserved_connections(
...
+        logger=logger,
+        level=level,
+        log_stdout=log_stdout,
+        log_stderr=log_stderr,
+        capture_stdout=capture_stdout,
+        capture_stderr=capture_stderr,
+        **kwargs,
+    )
+
+    exit_code, stdout_lines, stderr_lines = await asyncio.wait_for(
```
1st1, line 1039: Maybe use taskgroup?

edb/server/pgcluster.py
```diff
@@ -561,110 +556,54 @@ async def start(
...
+        self._daemon_process, *loggers = await _start_logged_subprocess(
+            [self._postgres, '-D', str(self._data_dir), *extra_args],
+            capture_stdout=False,
+            capture_stderr=False,
+            logger=postgres_logger,
+            log_processor=postgres_log_processor,
         )
 
+        for logger_coro in loggers:
+            asyncio.create_task(logger_coro)
```
1st1, line 568: Use `edb.common.supervisor`?

1st1 (approved):
Awesome.

edb/server/pgcluster.py
elprans: Why is a taskgroup better here?  `wait_for` shouldn't leak tasks, no?

1st1: > Why is a taskgroup better here?

More predictable failing than gather if something goes wrong?

1st1: Up to you, the current code is ok too

tailhook (commented):
Sorry for the late feedback, but two things bother me here:
1. You use plain `decode()` everywhere. But crashing in case postgres spits some binary text or incomplete utf-8 isn't great idea. (do we even ensure that locale is utf-8-based?). So at least `.decode('utf-8', 'replace')` should be used.
2. Timestamps could easily be offsetted by seconds when you use this method of capturing logs (I mean async reading; although threads would do only marginal difference). At least we have to make postgres print timestamps, parse them and print original timestamps.

elprans:
> 1. You use plain `decode()` everywhere.

`utf-8` is the default in Python 3, and we set Postgres `LC_MESSAGES` to a `utf-8` locale explicitly (`client_encoding` also).  That said, it doesn't hurt to be defensive, so I'll add `errors='replace'` there.

> 2. Timestamps could easily be offsetted by seconds ... At least we have to make postgres print timestamps, parse them and print original timestamps.

Unfortunately, we will have to live with this.  If we use Postgres' timestamps, then the log might become non-monotonic, and some logging systems (e.g. Loki) rely on timestamps being monotonic.  As an option, we can retain the original timestamp in the message, but what's the actual problem?

tailhook:
> > 1. Timestamps could easily be offsetted by seconds ... At least we have to make postgres print timestamps, parse them and print original timestamps.
> 
> Unfortunately, we will have to live with this. If we use Postgres' timestamps, then the log might become non-monotonic, and some logging systems (e.g. Loki) rely on timestamps being monotonic.

I think Loki should be fixed soon, if not already: https://grafana.com/docs/loki/latest/design-documents/2021-01-ordering-constraint-removal/
Also they say "monotonically increasing timestamps, partitioned by stream", so maybe these could probably be different streams? (i.e. there is a process that can split them up). 

> As an option, we can retain the original timestamp in the message, but what's the actual problem?

Any time you need to correlate logs with any external logs it can be a problem. I.e. debugging transaction conflicts or deadlocks. Or just slow behavior.

I'm also concerned that sometimes EdgeDB wouldn't be able to keep up with postgres logs or would crash (i.e. on OOM) with dropping important part of log in memory. So hopefully nobody will run Postgres by EdgeDB in a large-scale production. But at least we can fix timestamps.

PR #2829, opened 2021-08-18
Reformat the data model documentation.

vpetrovykh:
Change the "Data Model" section to "Concepts" and update the layout.
Also describe the concepts starting from "instance" and so on.


vpetrovykh:
I left the "Abstract Types" subsection mostly as is, because `anytype` is hard to allocate anywhere specific and leaving it as the only dangling bit seems odd.

docs/datamodel/abstract.rst
```diff
@@ -4,6 +4,13 @@
 Abstract Types
 ==============
 
+These types are used in introspection as well as in definitions of
```
1st1, line 7: Quoting https://github.com/edgedb/edgedb/discussions/2823:

> Abstract Types -> drop it. Move information about abstract scalar types to the Scalar Types section, anytuple to Collection Types, etc.

docs/datamodel/functions.rst
```diff
@@ -29,6 +29,34 @@ cardinality:
     {2}
 
 
+User-defined Functions
+----------------------
+
+It is also possible to define custom functions. For example, consider
+a function that adds an exclamation mark ``'!'`` at the end of the
+string or produces ``'!!!'`` if the input is omitted or empty:
```
1st1, line 37: The `!!!` part is unnecessary, let's keep it stupid-simple. Just concatenating `!` is sufficient.

docs/datamodel/functions.rst
```diff
@@ -29,6 +29,34 @@ cardinality:
...
+.. code-block:: sdl
+
+    function exclamation(word: optional str = {}) -> str
+        using ((word ++ '!') ?? '!!!');
+
+In order to make sure that the function is called when the argument is
+an empty set ``{}`` we make the argument :ref:`optional
+<ref_eql_fundamentals_optional>`. We also provide a default value of
+``{}`` if the argument is omitted entirely. Here are some results this
+function produces:
```
1st1, line 48: Yeah, the "optional" stuff is cool, but the whole purpose of this section is to say just one thing: "you can define your own functions". That's it. This section isn't about anything else, and especially it's not about `optional` arguments.

docs/datamodel/index.rst
```diff
@@ -1,50 +1,116 @@
...
+is created with an empty database called "edgedb". If you wish to add
+more, you can use the :eql:stmt:`CREATE DATABASE` EdgeQL command.
+Conversely, the :eql:stmt:`DROP DATABASE` command removes a database.
+
+The following command will get a list of all databases present in the
+instance:
+
+.. code-block:: edgeql
+
+    SELECT sys::Database.name;
```
1st1, line 58: Perhaps mention `\l` in the interactive shell?

docs/datamodel/index.rst
```diff
@@ -1,50 +1,116 @@
...
+
+Modules
+=======
+
+Every database has a schema that fully describes it. Schema consists
+of logical units called  **modules**, which act as namespaces.
+Modules contain user-defined types, functions, etc. A module has a
+name that is unique inside a database. The same schema object name can
+be used in different modules without conflict.  For example, both
+``module1`` and ``module2`` can contain a ``User`` object type.
```
1st1, line 70: Good stuff!

docs/datamodel/index.rst
```diff
@@ -1,50 +1,116 @@
...
+
+* ``std``: standard types, functions and other declarations
+* ``schema``: types describing the :ref:`introspection <ref_eql_introspection>`
+  schema
+* ``sys``: system-wide entities, such as user roles and
+  :ref:`databases <ref_datamodel_databases>`
+* ``cfg``: configuration and settings
+* ``stdgraphql``: GraphQL-related types
+* ``math``: algebraic and statistical functions
+* ``default``: the default module for user-defined types, functions, etc.
```
1st1, line 85: Add a TODO here to link to the future `Standard Library` section.

docs/datamodel/index.rst
```diff
@@ -1,50 +1,116 @@
...
+for faster querying, and **computed** properties or links to define
+useful expressions (e.g. ``property email := .user_name ++ '@' ++
+.provider_name``).
+
+**Expression Aliases** let you use existing types under new names to
+build on them without touching the original -- both in your schema or
+on the fly inside a query.
+
+You can also create your own **functions**, strongly typed along with
+everything else in EdgeDB.
```
1st1, line 116: Love this new section, great job. Now we just need a cool diagram, but we'll add it later.

1st1 (commented):
This is really great. I've left a few comments.

1st1:
> I left the "Abstract Types" subsection mostly as is, because anytype is hard to allocate anywhere specific and leaving it as the only dangling bit seems odd.

`anytype` should be listed in `Concepts / Functions`, perhaps add `Generic` functions for that with an example. It would be way cooler to introduce it that way, because generic functions are super cool.

docs/datamodel/functions.rst
vpetrovykh: Fixed.

docs/datamodel/functions.rst
vpetrovykh: FIxed.

docs/datamodel/index.rst
vpetrovykh: Added.

docs/datamodel/index.rst
vpetrovykh: Added

1st1 (approved):
OK to go and let's continue refactoring the docs

Issue #2827, opened 2021-08-18
Cardinality violation is not detected


vpetrovykh:
- EdgeDB Version: 1-beta3 (docker)
- OS Version: linux

Steps to Reproduce:

```
test> create type Foo;
OK: CREATE TYPE
test> create type Bar {create required single link foo -> Foo};
OK: CREATE TYPE
test> insert Foo;
{default::Foo {id: 33ba1e22-ffb7-11eb-a3f3-0f3f7e1586b7}}
test> insert Bar {foo := Foo};
{default::Bar {id: 367723c6-ffb7-11eb-a3f3-1fd1c76544b5}}
test> insert Foo;
{default::Foo {id: 382e27a0-ffb7-11eb-a3f3-079f3251eb68}}
test> insert Bar {foo := Foo};
ERROR: InternalServerError: duplicate key value violates unique constraint "2dbcc6bd-ffb7-11eb-9cd5-bd146c902983_pkey"
  Hint: This is most likely a bug in EdgeDB. Please consider opening an issue ticket at https://github.com/edgedb/edgedb/issues/new?template=bug_report.md
  Server traceback:
      edb.errors.InternalServerError: duplicate key value violates unique constraint "2dbcc6bd-ffb7-11eb-9cd5-bd146c902983_pkey"
```

PR #2830, opened 2021-08-18
Enforce pointer cardinality statically in mutations

elprans:
Make sure we actually raise a cardinality assertion error when an
expression to the right of `:=` in mutations violates the declared
pointer cardinality.

Fixes: #2827

### DIFF
--- a/tests/test_edgeql_insert.py
+++ b/tests/test_edgeql_insert.py
@@ -3643,3 +3643,15 @@ class TestInsert(tb.QueryTestCase):
                     (INSERT Person { name := 'foo' })
                 )
             ''')
+
+    async def test_edgeql_insert_cardinality_assertion(self):
+        async with self.assertRaisesRegexTx(
+                edgedb.QueryError,
+                "possibly more than one element returned by an expression "
+                "for a link 'sub' declared as 'single'"):
+            await self.con.query(r'''
+                INSERT InsertTest {
+                    l2 := 10,
+                    sub := Subordinate,
+                }
+            ''')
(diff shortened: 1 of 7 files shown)


1st1 (approved):

tests/test_schema.py
```diff
@@ -7209,7 +7209,7 @@ def test_schema_describe_10(self):
                         UPDATE Foo
                         FILTER .val = 1
                         SET {
-                            val += 1
+                            val := .val + 1
```
msullivan, line 7212: Why does this need to change?

msullivan (approved):
Main change looks good but I'm kind of weirded out by the `+=` thing? (Though that seems probably a separate issue)

msullivan:
Oh, lol, indeed.

tests/test_edgeql_insert.py
```diff
@@ -3643,3 +3643,15 @@ async def test_edgeql_insert_and_delete_01(self):
                     (INSERT Person { name := 'foo' })
                 )
             ''')
+
+    async def test_edgeql_insert_cardinality_assertion(self):
+        async with self.assertRaisesRegexTx(
```
1st1, line 3648: Can you add a test for UPDATE, simple & straightforward like this one for INSERT?

elprans: I already did, see the diff

vpetrovykh (approved):
Looks good and I see that you also fixed the wording for some errors from "computed link" to just "link" when dealing with `INSERT` or `UPDATE`.

tests/test_edgeql_insert.py
1st1: Got it, there was a test already, GH just collapsed the view of it.

PR #2831, opened 2021-08-18
Fix test instance destruction regression

elprans:
This is a regression from #2806 which made test EdgeDB instances linger
after test conclusion.

### DIFF
--- a/edb/tools/test/runner.py
+++ b/edb/tools/test/runner.py
@@ -797,6 +797,7 @@ class ParallelTextTestRunner:
                     )
 
                 async def _setup():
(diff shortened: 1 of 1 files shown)


edb/tools/test/runner.py
```diff
@@ -797,6 +797,7 @@ def run(self, test, selected_shard, total_shards, running_times_log_file):
                     )
 
                 async def _setup():
+                    nonlocal cluster
```
1st1, line 800: :facepalm:

1st1 (approved):

PR #2836, opened 2021-08-18
Report the values of `_Py_HashSecret` and `random.seed()` in test failures

elprans:
This should hopefully help with the diagnostics of extreme Heisenbugs.
Unfortunately, it is not possible to directly specify the exact value of
`_Py_HashSecret`, so a minor CPython patch would be necessary to make
this data useful.

### DIFF
--- a/edb/tools/test/cpython_state.py
+++ b/edb/tools/test/cpython_state.py
@@ -1,0 +1,34 @@
+#
+# This source file is part of the EdgeDB open source project.
+#
+# Copyright 2021-present MagicStack Inc. and the EdgeDB authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+from __future__ import annotations
+from typing import *
+
+import ctypes
+
+
+class _Py_HashSecret_t(ctypes.Union):
+    _fields_ = [
+        ('uc', ctypes.c_ubyte * 24),
+    ]
+
+
+def get_py_hash_secret() -> bytes:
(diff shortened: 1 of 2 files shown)


edb/tools/test/cpython_hacks.py
```diff
@@ -0,0 +1,34 @@
...
+
+class _Py_HashSecret_t(ctypes.Union):
+    _fields_ = [
+        ('uc', ctypes.c_ubyte * 24),
+    ]
+
+
+def get_py_hash_secret() -> bytes:
+    hashsecret = _Py_HashSecret_t.in_dll(ctypes.pythonapi, '_Py_HashSecret')
+    return bytes(hashsecret.uc)
```
1st1, line 34: Do we need to care about 32 bit builds? (I'm not if _Py_HashSecret_t is defined differently for those)

edb/tools/test/cpython_hacks.py
```diff
@@ -0,0 +1,34 @@
+#
```
1st1, line 1: Can we please give a descriptive name to this module -- `secret_hash.py` or something.

1st1 (approved):

edb/tools/test/cpython_hacks.py
msullivan: maybe in a `cpython_hacks` directory <_<

elprans: I think cpython_hacks is descriptive here as we are reaching into the interpreter internals specific to cpython and it's likely this wouldn't be the last such hack

1st1: I disagree. The hack part is an implementation detail, I couldn't care less how a module works.

edb/tools/test/cpython_hacks.py
elprans: It is always 24 bytes long: https://github.com/python/cpython/blob/b2f68b190035540872072ac1d2349e7745e85596/Include/pyhash.h#L56-L57

elprans:
Here's a CPython patch to go along with this: https://github.com/elprans/cpython/commit/0cece3de55b26d247ebaaf7c811f93c2f8d0a268

edb/tools/test/cpython_hacks.py
elprans: Renamed to `cpython_state`. I'd like to avoid naming the whole module after the function.

1st1: Sure, that's a good name!

PR #2837, opened 2021-08-18
Change query type argument type from tuple to object

fmoor:
To support checking query argument cardinality during the [prepare phase](https://www.edgedb.com/docs/internals/protocol/overview#granular-flow) the argument type is changed from tuple to object because the former does not carry cardinality information at the protocol level and the later does.

The protocol version is bumped to 0.12


edb/server/compiler/compiler.py
```diff
@@ -537,6 +539,65 @@ def _compile_ql_script(
...
+            arg_ast, 
+            schema=schema,
+            options=arg_compile_options,
+        )
+
+        in_type_data, in_type_id = sertypes.TypeSerializer.describe(
+                schema=ir_expr.schema,
+                typ=ir_expr.stype,
+                view_shapes={},
+                view_shapes_metadata={},
```
elprans, line 593:
```suggestion
                view_shapes=ir_expr.view_shapes,
                view_shapes_metadata=ir_expr.view_shapes_metadata,
```

edb/server/compiler/compiler.py
```diff
@@ -537,6 +539,65 @@ def _compile_ql_script(
...
+            # todo fmoor: this doesn't handle qualified type names
+            type_name = schema_type.get_name(schema).get_local_name().name
+
+            elem = qlast.ShapeElement(
+                expr=qlast.Path(steps=[qlast.Ptr(
+                    ptr=qlast.ObjectRef(name=f'__arg_{name}'),
+                    direction=s_pointers.PointerDirection.Outbound,
+                )]),
+                compexpr=qlast.TypeCast(
+                    expr=qlast.Parameter(name=name, optional=not required),
```
fmoor, line 560: The next puzzle here is that it doesn't matter if `optional` is set to `True` or `False`. In both cases the field cardinality in the resulting descriptor data is `0x41 (ONE)`. It should be `0x6f (AT_MOST_ONE)` when `optional=True`.

Any ideas @elprans?

elprans: This seems like a bug, let me see.

elprans: You need to specify optionality on the `TypeCast` node, not `Parameter`.  I think `Parameter.optional` is a vestige that needs removal.

edb/server/compiler/compiler.py
```diff
@@ -537,6 +539,73 @@ def _compile_ql_script(
...
+
+        arg_compile_options = qlcompiler.CompilerOptions(
+            modaliases=current_tx.get_modaliases(),
+            implicit_tid_in_shapes=False,
+            implicit_id_in_shapes=False,
+            implicit_tname_in_shapes=False,
+            apply_query_rewrites=False,
+        )
+
+        ir_expr = qlcompiler.compile_ast_to_ir(
```
fmoor, line 592: @elprans it seems like compiling an empty shape does not respect `implicit_id_in_shapes=False` and inserts an `id` in the result anyways. Should this behavior change or instead I could namespace arguments and then have `TypeSerializer.describe()` filter based on an argument prefix?

elprans: This seems to be driven by [this](https://github.com/edgedb/edgedb/blob/929cf4134988cc4e578c79f82d86f24e3bda47af/edb/edgeql/compiler/viewgen.py#L1219).  @msullivan will know more, but it seems to be a safer bet to specialize `describe()` and filter `id` out.

edb/server/compiler/compiler.py
```diff
@@ -537,6 +539,70 @@ def _compile_ql_script(
...
+    def _compile_query_args(
+            self, 
+            ctx: CompileContext,
+            subtypes: List[Tuple[str, Any, bool]], 
+    ) -> Tuple[bytes, uuid.UUID]:
+        # TODO(fmoor) perhaps check for zero parameters and 
+        # return a static value (b'...', uuid.UUID(b'...'))
+
+        current_tx = ctx.state.current_tx()
+        schema = current_tx.get_schema(self._std_schema)
```
fmoor, line 551: Does this look like the right schema object to be using? I am getting a few errors like this from `s_utils.typeref_to_ast()`:

```
edgedb.errors.SchemaError: cannot get item data: item 'f3000700-db1e-1114-fca2-12a8e050d706' is not present in the chema <FlatSchema gen:6104 at 0x7f5dad066490>
```

elprans: You should be using the schema object returned with the results of the compilation of the main query (`ir.schema`)

edb/server/compiler/compiler.py
```diff
@@ -537,6 +537,71 @@ def _compile_ql_script(
...
+    def _compile_query_args(
+            self,
+            ctx: CompileContext,
+            schema: s_schema.Schema,
+            subtypes: List[Tuple[str, Any, bool]],
+    ) -> Tuple[bytes, uuid.UUID]:
+        # TODO(fmoor) perhaps check for zero parameters and
+        # return a static value (b'...', uuid.UUID(b'...'))
+        if not subtypes:
+            return b'\x01(\xda\xba\tJ\x8b\x86h\xf8!T\xb8]\xf6\xe6\x0e\x00\x00', uuid.UUID('28daba09-4a8b-8668-f821-54b85df6e60e')
```
fmoor, line 549: @elprans is something like this valid? I have no idea if the descriptor id is stable or not. Skipping compilation of zero arguments side steps a class of errors where `qlcompiler.compile_ast_to_ir()` raises an exception about some type already existing that only happens when the AST being compiled is ~ `SELECT {};`.

elprans: What is the error you're seeing?  If manually constructed `SELECT {}` (as in shape, not empty set), doesn't work then `SELECT std::FreeObject` should surely work.  I'd very much like to avoid hardcoding serialized descriptors like this.

edb/server/compiler/compiler.py
```diff
@@ -684,9 +760,13 @@ def _compile_ql_query(
                 ir.schema, params_type = s_types.Tuple.create(
                     ir.schema, element_types={}, named=False)
 
-            in_type_data, in_type_id = sertypes.TypeSerializer.describe(
-                ir.schema, params_type, {}, {},
-                protocol_version=ctx.protocol_version)
+            if ctx.protocol_version >= (0, 12):
```
fmoor, line 763: I _think_ this the last bug. Which branch should be taken when the protocol is `None` (HTTP)? This currently fails with `TypeError: '>=' not supported between instances of 'NoneType' and 'tuple'`

elprans: The user-facing HTTP interface is not changing, so the HTTP extension should always use the latest version (and set it explicitly).

edb/server/compiler/compiler.py
```diff
@@ -537,6 +537,76 @@ def _compile_ql_script(
...
+            self,
+            ctx: CompileContext,
+            schema: s_schema.Schema,
+            subtypes: List[Tuple[str, Any, bool]],
+    ) -> Tuple[bytes, uuid.UUID]:
+        current_tx = ctx.state.current_tx()
+
+        if not subtypes:
+            arg_ast = qlast.SelectQuery(
+                orderby=[],
```
elprans, line 550: You shouldn't be needing to set `orderby` to `[]` explicitly.

edb/server/compiler/compiler.py
```diff
@@ -537,6 +537,76 @@ def _compile_ql_script(
...
+                        expr=qlast.Parameter(name=name),
+                        type=s_utils.typeref_to_ast(schema, schema_type),
+                    ),
+                    operation=qlast.ShapeOperation(op=qlast.ShapeOp.ASSIGN),
+                )
+
+                elements.append(elem)
+
+            arg_ast = qlast.SelectQuery(
+                orderby=[],
```
elprans, line 581: Ditto

edb/server/compiler/compiler.py
```diff
@@ -537,6 +537,76 @@ def _compile_ql_script(
...
+                    card = qlast.CardinalityModifier.Optional
+
+                elem = qlast.ShapeElement(
+                    expr=qlast.Path(steps=[qlast.Ptr(
+                        ptr=qlast.ObjectRef(name=f'__arg_{name}'),
+                        direction=s_pointers.PointerDirection.Outbound,
+                    )]),
+                    compexpr=qlast.TypeCast(
+                        cardinality_mod=card,
+                        optional=not required,
```
elprans, line 571: `TypeCast` doesn't have an `optional` attribute.  I'm surprised `mypy` didn't complain here, since our mypy plugin is injecting strict `__init__` signatures.  @msullivan, ideas?

elprans (approved):
LGTM.  A few nits.  Did you test this with the updated binding?

1st1 (commented):
I think this is overkill. We should instead hack TypeSerializer` to serialize object-like shape without the need to construct edgeql and compile it to IR. While that will be a bit more hacky, we really don't mind doing that just once. I'm OK with this PR being merged now (and making edgedb-python use the new protocol), but we must redo this in a follow up PR.

1st1:
> I think this is overkill. 

@fmoor Frederick, to be clear, I think this PR has to be merged asap and we should proceed with fixing the bindings. What I suggest in my comment is an optimization, we can do it later.

edb/server/compiler/compiler.py
fmoor: changed

edb/server/compiler/compiler.py
fmoor: changed

edb/server/compiler/compiler.py
fmoor: changed

fmoor:
> Did you test this with the updated binding?

Yes I did. Both the server and binding testsuites pass on my machine.

PR #2839, opened 2021-08-19
Fix sorting of apparent server package version

elprans:
Include the date as the first component of build metadata in release
versions to ensure proper sortability.  This is a regression from #2475.

In addition, switch to using the top commit date instead of current
date, as the latter is encoded in package revision already.

Fixes: edgedb/edgedb-cli#477

### DIFF
--- a/.github/workflows.src/release.tpl.yml
+++ b/.github/workflows.src/release.tpl.yml
@@ -21,10 +21,14 @@ jobs:
           echo ::error::"${branch} is not a release branch."
           exit 1
         fi
+        commitish=$(git rev-parse --verify --quiet "${GITHUB_REF}" | cut -c1-9)
+        date=$(env TZ=UTC git show -s \
+               --format=%cd --date=format-local:%Y%m%d%H "$commitish")
         ver=${branch#releases/}
-        ver+=+g$(git rev-parse --verify --quiet "${GITHUB_REF}" | cut -c1-9)
+        ver+=+d${date}
+        ver+=.g${commitish}
         ver+=.cv$(sed -n -r 's/EDGEDB_CATALOG_VERSION = ([0-9_]+)/\1/p' \
-                   edb/server/defines.py | sed 's/_//g')
+                  edb/server/defines.py | sed 's/_//g')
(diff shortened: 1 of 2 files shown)


1st1:
Does this have any backwards compatibility problems? Like older (current) versions of the CLI being broken, or maybe something else?

elprans:
> Does this have any backwards compatibility problems? Like older (current) versions of the CLI being broken, or maybe something else?

As far as I understand, this should _fix_ the compatibility problem introduced in #2475

laosb:
Just hitting this. Any release schedule for the fix?

tailhook:
@laosb, sorry, for delaying this, but it also requires fix for CLI if it's fixed this way.

@elprans, I'm still confused by the change. Why #2475 changed stable version numbers? It was supposed to fix nightlies so was supposed to change nightlies only, no? Neither CLI nor system package manager can't reasonably act on change of EDGEDB_CATALOG_VERSION within the single stable version anyway.

elprans:
> Why #2475 changed stable version numbers

We need the catver metadata for auto-upgrades.  Is the addition of build metadata to stable versions still breaking even if properly sortable?

tailhook:
> > Why #2475 changed stable version numbers
> 
> We need the catver metadata for auto-upgrades.

What kind of auto-upgrades? For any kind of niche use case we can actually fetch json and look for catalog version.

> Is the addition of build metadata to stable versions still breaking even if properly sortable?

It's hard to say. Version comparison should work. Although, you effectively made `revision` field useless. But there maybe quirks here and there as this wasn't how it was intended to work and this is quite hard to catch :shrug:  We should also probably do some fine-tuning of `edgedb server list-versions` as longer version numbers look ugly there.

elprans:
>  For any kind of niche use case we can actually fetch json and look for catalog version.

That's the thing.  JSON indexes are populated from the build metadata in the version.

> Although, you effectively made `revision` field useless

Notice that in this diff I'd switched the date field in version to be that of the top commit, and not _current_ date.  So `+d` is the source code version, and the revision is the package build date.

tailhook (approved):

PR #2842, opened 2021-08-20
Drop direction from PointerRef

msullivan:
This makes PointerRef more directly a reflection of schema pointers,
and lets me delete the recently added `outbound_ptr` that I added to
allow reversing of a inbound PointerRef.

Direction is already tracked in Pointer and in PathId, which allows us
to reconstruct it wherever we need to.


elprans (approved):
Good cleanup.

PR #2844, opened 2021-08-20
Fix cardinality inference in "optionality-preserving" functions

elprans:
A certain number of standard aggregate functions are guaranteed to
preserve the lower cardinality bound of their argument.  These functions
are `std::assert_single`, `std::enumerate`, `std::min`, and `std::max`.
Unfortunately, there is currently no way to signal that lower
cardinality is preserved, and so we introduce the new std-only
`preserves_cardinality` field, which is also exposed in the public
introspection schema for the benefit of query builders and such.

While doing this, I realised that we also mishandle functions with
non-OPTIONAL arguments that return `OPTIONAL` or `SET OF`, so I fixed
inference of those too.

### DIFF
--- a/edb/edgeql/compiler/inference/cardinality.py
+++ b/edb/edgeql/compiler/inference/cardinality.py
@@ -672,32 +672,73 @@ def __infer_func_call(
     scope_tree: irast.ScopeTreeNode,
     ctx: inference_context.InfCtx,
 ) -> qltypes.Cardinality:
-    SetOfType = qltypes.TypeModifier.SetOfType
+    return_card = (
+        MANY if ir.typemod is qltypes.TypeModifier.SetOfType else
+        AT_MOST_ONE if ir.typemod is qltypes.TypeModifier.OptionalType else
+        ONE
+    )
 
-    args = []
-    # process positional args
-    for arg, typemod in zip(ir.args, ir.params_typemods):
-        arg.cardinality = infer_cardinality(
-            arg.expr,
-            scope_tree=scope_tree, ctx=ctx)
-        if typemod is not SetOfType:
-            args.append(arg.expr)
+    ret_lower_bound, ret_upper_bound = _card_to_bounds(return_card)
 
-    # the cardinality of the function call depends on the cardinality
-    # of non-SET_OF arguments AND the cardinality of the function
-    # return value
-    if ir.typemod is SetOfType:
-        return MANY
+    if ir.preserves_optionality:
+        # This is a generic aggregate function which preserves the
+        # optionality of its generic argument.  For simplicity we
+        # are deliberately not checking the parameters here as that
+        # would have been done at the time of declaration.
+        arg_cards = []
+
+        for arg in ir.args:
+            arg.cardinality = infer_cardinality(
+                arg.expr, scope_tree=scope_tree, ctx=ctx)
+            arg_cards.append(arg.cardinality)
+
+        arg_card = zip(*(_card_to_bounds(card) for card in arg_cards))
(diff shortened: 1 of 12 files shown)


edb/lib/std/20-genericfuncs.edgeql
```diff
@@ -321,6 +333,7 @@ std::max(vals: SET OF duration) -> OPTIONAL duration
         'Return the greatest value of the input set.';
     SET volatility := 'Immutable';
     SET force_return_cast := true;
+    SET preserves_optionality := true;
```
1st1, line 336: Please add a check to `edb/schema/function.py` to check if all variants of function specify `preserves_optionality` if one sets it.

elprans: Good call.  I actually planned to add a whole bunch of validation, but somehow forgot to do it :man_facepalming:

msullivan: We would be able to support them not doing it, though, right? Disallowing it would be because we don't think we'd ever want to on purpose?

edb/edgeql/compiler/inference/cardinality.py
```diff
@@ -672,32 +672,65 @@ def __infer_func_call(
...
     else:
-        if args:
-            return _common_cardinality(
-                args, scope_tree=scope_tree, ctx=ctx,
-            )
-        else:
-            if ir.typemod is qltypes.TypeModifier.OptionalType:
-                return AT_MOST_ONE
+        # For regular non-OPTIONAL functions, the general rule of
+        # Cartesian cardinality of argument applies, although we still
```
msullivan, line 701: s/argument/arguments/

edb/edgeql/compiler/inference/cardinality.py
```diff
@@ -672,32 +672,65 @@ def __infer_func_call(
...
             else:
-                return ONE
+                all_singletons = False
+
+        if not singleton_args:
+            # Either no arguments at all, or all arguments are non-singletons,
+            # so the declared return cardinality is as specific as we can get.
+            return return_card
+        else:
+            _check_op_volatility(singleton_args, singleton_arg_cards, ctx=ctx)
```
msullivan, line 728: I think we need to do `_check_op_volatility` on at least the OPTIONAL arguments. (Though I'd probably do it for SET OF ones too, unless we find an actual use-case and are convinced it works?)

msullivan (approved):
LGTM, module the volatility stuff

edb/lib/std/20-genericfuncs.edgeql
elprans: Semantically, all overloads should behave the same, I can't think of a use case where a specific overload would need to be exempted.

edb/edgeql/compiler/inference/cardinality.py
elprans: I'm not sure `_check_op_volatility` would work correctly for `SET OF` args.  Consider this hypothetical function:

```
func string_agg(items: SET OF str, delimiter: str) -> str { ... } 
```

and then

```SELECT string_agg({"foo", "bar"}, <str>random())```

This is a perfectly legal call, but since `items` is `MANY`, `_check_op_volatility` would trip, despite no Cartesian product actually being made.

edb/lib/std/20-genericfuncs.edgeql
elprans: Fixed.

1st1 (approved):

PR #2849, opened 2021-08-23
raise AuthenticationError when Auth for role is not configured

### LINKED ISSUE #2848: ISE: `unimplemented auth method: NoneType` when inserting new user on clean instance
- EdgeDB Version: 1-beta3-451baf4
- OS Version: ubuntu 18.04

Steps to Reproduce:

1. initialize new project with new `EdgeDB` instance in docker with `edgedb project init`
2. login into EdgeDB shell using `edgedb` command and create new user with password:
```
CREATE SUPERUSER ROLE edgedb_scram {
    SET password := 'edgedb_scram_password'
};

CONFIGURE INSTANCE INSERT Auth {
    user := 'edgedb_scram',
    method := (INSERT SCRAM),
    priority := 2
};
```
3. exit current shell session and try running `edgedb` command again to login into shell
```
$ edgedb
edgedb error: Error authenticating: ERROR: InternalServerError: unimplemented auth method: NoneType
  Server traceback:
        File "edb/server/protocol/binary.pyx", line 1689, in edb.server.protocol.binary.EdgeConnection.main
        File "edb/server/protocol/binary.pyx", line 445, in auth
      edb.errors.InternalServerError: unimplemented auth method: NoneType
```

As I understand it, this is because the default user (`edgedb`) created by the EdgeDB bootstrap process is not being added to `cfg::Auth`. This can be verified by selecting the existing `cfg::Auth` objects before inserting a new one:
```
SELECT cfg::Auth {
    user,
    method,
    priority
};
```
The output will be an empty set (`{}`).

But in any case, user should not receive such error from the EdgeDB server in case of authentication failure. I would rather expect something like `AuthenticationError: no authentication method configured for the "edgedb" role` or something similar.

It's possible I missed something about roles configuration, but that part of EdgeDB is poorly documented right now, IMO.

nsidnev:
Closes #2848

### DIFF
--- a/edb/server/server.py
+++ b/edb/server/server.py
@@ -1191,6 +1191,8 @@ class Server:
                 if match:
                     return auth.method
 
+        raise errors.AuthenticationError(f'no authentication method configured for {user!r} role')
+
     def get_sys_query(self, key):
         return self._sys_queries[key]
 
--- a/edb/server/server.py
+++ b/edb/server/server.py
@@ -1191,7 +1191,9 @@ class Server:
                 if match:
                     return auth.method
 
-        raise errors.AuthenticationError(f'no authentication method configured for {user!r} role')
+        raise errors.AuthenticationError(
+            f"no authentication method configured for {user!r} role"
+        )
 
     def get_sys_query(self, key):
         return self._sys_queries[key]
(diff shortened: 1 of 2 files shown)


1st1 (approved):
LGTM, thanks so much @nsidnev!

Issue #2622, opened 2021-06-25
InternalServerError: duplicate key value violates unique constraint "9802c422-d5d1-11eb-9b10-459c32f1a94a_source_target_key"


isidentical:
- EdgeDB Version: Beta 1
- OS Version: Ubuntu 20.04

Steps to Reproduce:

1. Use this schema
```
START MIGRATION TO {
    module ast {
        type Compare {
            multi property ops -> cmpop;
        }
        scalar type cmpop extending enum<'Eq', 'NotEq', 'Lt', 'LtE', 'Gt', 'GtE', 'Is', 'IsNot', 'In', 'NotIn'> {}
    }
};

POPULATE MIGRATION;
COMMIT MIGRATION;
```

2. Execute the following command
```
bug_report> INSERT ast::Compare {ops := {<ast::cmpop>'Eq', <ast::cmpop>'Eq'}};
ERROR: InternalServerError: duplicate key value violates unique constraint "9802c422-d5d1-11eb-9b10-459c32f1a94a_source_target_key"
  Hint: This is most likely a bug in EdgeDB. Please consider opening an issue ticket at https://github.com/edgedb/edgedb/issues/new?template=bug_report.md
  Server traceback:
      edb.errors.InternalServerError: duplicate key value violates unique constraint "9802c422-d5d1-11eb-9b10-459c32f1a94a_source_target_key"
```

This is kinda weird because, you can use multiple *different* operators as well as you can individually construct a set of `Eq` objects. But the problem arises when you have multiple of them on a link.

elprans:
This is because currently we prohibit multiplicity > 1 in properties (i.e. the values must be unique), and the error isn't handled correctly here.  However, there's a plan to remove property multiplicity constraints: https://github.com/edgedb/edgedb/discussions/2054

isidentical:
Thanks for the quick response @elprans! Indeed #2054 seems really cool. I'll probably switch the current model to use `property ops -> array<cmpop>` instead of `multi property ops -> cmpop` to support multiple same values.

elprans:
Yeah, using an array instead of a `multi` property is a good workaround.  Furthermore, if you treat the whole array as a single value, i.e. you don't often need to filter on individual elements or modify the array frequently, an array is better than a multi property, because it's more compact and local to the host object.

PR #2853, opened 2021-08-24
Make properties non-distinct by default

elprans:
All values in EdgeQL are considered to be multisets, which are a
generalization of the set concept that allows duplicate elements.  For
brevity we refer to multisets as just "sets", and use "proper set" or
"distinct set" when referring to proper mathematical sets.

In EdgeDB the data is modeled as a directed graph, where objects are
vertices and links are edges.  Query path expressions (`Foo.bar.baz`)
are graph traversal operators and, by definition, always return a
*distinct set*.  This means that there can be no link of the same name
between a pair of objects, and this also means that a set of objects
pointed to by a `multi` link is always distinct.  Computed links must
follow this rule too so as to behave exactly like materialized links
externally.  To illustrate, consider a common case of finding a set of
"friends of friends" of a particular user.  In EdgeQL this is simply
`SELECT User.friends.friends`.  Proper set semantics here allows this
operation to use a more efficient semi-join in place of a regular join
(especially if the latter is later coupled with a use of `DISTINCT`).  A
majority of queries on object paths are like that and so this behavior
is desirable.

Properties are different, because their value is scalar, and scalars
lack true identity, which makes any proper set mechanics on scalar sets
expensive due to explicit elimination of duplicates.  Furthermore,
non-distinct scalar sets are actually *desirable* in many queries,
especially where analytics and tuples are involved.  Finally, we should
enforce consistent multiplicity rules on computed properties, and static
inference of multiplicity on scalars is weak and will effectively force
users to pollute queries with pointless and expensive `DISTINCT`.

Here we change EdgeDB behavior to allow multiplicity greater than one in
`multi` properties.

See edgedb/edgedb#2054 for the discussion.

Fixes: #2622.

### DIFF
--- a/edb/pgsql/compiler/dml.py
+++ b/edb/pgsql/compiler/dml.py
@@ -1292,40 +1292,216 @@ def process_link_update(
         if shape_op is qlast.ShapeOp.SUBTRACT:
             data_rvar = relctx.rvar_for_rel(data_select, ctx=ctx)
 
-            # Drop requested link records.
-            delqry = pgast.DeleteStmt(
-                relation=target_rvar,
-                where_clause=astutils.new_binop(
-                    lexpr=astutils.new_binop(
-                        lexpr=source_ref,
-                        op='=',
-                        rexpr=pgast.ColumnRef(
-                            name=[target_alias, 'source'],
+            if target_is_scalar:
+                # MULTI properties are not distinct, and since `-=` must
+                # be a proper inverse of `+=` we cannot simply DELETE
+                # all property values matching the `-=` expression, and
+                # instead have to resort to careful deletion of no more
+                # than the number of tuples returned by the expression.
+                # Here, we rely on the "ctid" system column to refer to
+                # specific tuples.
+                #
+                # DELETE
+                #   FROM <link-tab>
+                # WHERE
+                #   ctid IN (
+                #     SELECT
+                #       shortlist.ctid
+                #     FROM
+                #       (SELECT
+                #         source,
+                #         target,
+                #         count(target) AS cnt
+                #        FROM
+                #         <data-expr>
+                #        GROUP BY source, target
+                #       ) AS counts,
+                #       LATERAL (
+                #         SELECT
+                #           candidates.ctid
+                #         FROM
+                #           (SELECT
+                #             ctid,
+                #             row_number() OVER (
+                #               PARTITION BY data
+                #               ORDER BY data
+                #             ) AS rn
+                #           FROM
+                #             <link-tab>
+                #           WHERE
+                #             source = counts.source
+                #             AND target = counts.target
+                #           ) AS candidates
+                #         WHERE
+                #           candidates.rn <= counts.cnt
+                #       ) AS shortlist
+                #   );
+
+                val_src_ref = pgast.ColumnRef(
+                    name=[data_rvar.alias.aliasname, 'source'],
+                )
+                val_tgt_ref = pgast.ColumnRef(
+                    name=[data_rvar.alias.aliasname, 'target'],
+                )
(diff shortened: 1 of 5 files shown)


tests/test_edgeql_update.py
```diff
@@ -2349,6 +2353,68 @@ async def test_edgeql_update_subtract_02(self):
...
+                    .name
+            """,
+            [
+                {
+                    'str_tags': {'1', '2', '3', '4'},
+                },
+                {
+                    'str_tags': {'4'},
+                },
+            ],
```
1st1, line 2415: Can you test that `-= {}` works too?

1st1 (approved):

tests/test_edgeql_update.py
elprans: Done.

edb/pgsql/compiler/dml.py
```diff
@@ -1289,40 +1289,214 @@ def process_link_update(
...
-                        op='=',
-                        rexpr=pgast.ColumnRef(
-                            name=[target_alias, 'source'],
+            if target_is_scalar:
+                # MULTI properties are not distinct, and since `-=` must
+                # be a proper inverse of `+=` we cannot simply DELETE
+                # all property values matching the `-=` expression, and
+                # instead have to resort to careful deletion of no more
+                # than the number of tuples returned by the expression.
+                # Here, we rely on the "ctid" system column here to refer
```
1st1, line 1298:
```suggestion
                # Here, we rely on the "ctid" system column to refer
```

tests/test_edgeql_update.py
```diff
@@ -2349,6 +2353,126 @@ async def test_edgeql_update_subtract_02(self):
...
+                    'str_tags': ['1', '2', '2', '3', '4', '5'],
+                },
+            ],
+        )
+
+        await self.con.execute("""
+            UPDATE UpdateTest
+            FILTER .name LIKE 'update-test-subtract-non-distinct-%'
+            SET {
+                str_tags -= {'1', '2', '2', '3', '5'}
```
1st1, line 2393: Side question, is it legal to use both `-=` and `+=` in the same `SET` for the same property (it should be, I guess)? Can you add tests for that too?

1st1:
Do we need to update the docs?

tests/test_edgeql_update.py
elprans: > Side question, is it legal to use both `-=` and `+=` in the same `SET` for the same property

It isn't:

```
error: duplicate definition of property 'p' of object type 'default::Foo'
  ┌─ query:1:37
  │
1 │ UPDATE Foo SET { p -= {1, 2, 2, 3}, p += {1} };
```

elprans:
> Do we need to update the docs?

Yes, but given all the pending reorg, I dare not to touch that folder for now.

1st1:
I'm gonna be testing and merging https://github.com/edgedb/edgedb/pull/2846 now. Other than that there will be no reorg until @vpetrovykh returns.

tests/test_edgeql_update.py
1st1: Obviously out of scope of this PR, but I'm curious what you think about it? Should we allow if post 1.0?

elprans: No.  It would be a complicated thing to support and it potentially opens the door for unintended copy-paste bugs.

vpetrovykh:
It occurs to me that having non-distinct multi props makes introducing "properties on properties" much harder (largely because identifying which specific property value needs a "prop-prop" becomes harder there and would be prone to errors). So if in the future we have some fancy structures on links (like "links on links") we probably wouldn't find it easy to introduce or manage the same things for properties.

Issue #2876, opened 2021-08-27
Concurrent complex `UNLESS CONFLICT` queries can spuriously raise ConstraintViolationError


msullivan:
If we, in two concurrent transactions, do:
```
FOR name IN {'name'} UNION (
    SELECT (
        INSERT Counter {
            name := name,
            value := 1,
        } UNLESS CONFLICT ON .name
        ELSE (
            UPDATE Counter
            SET { value := .value + 1 }
        )
    ).value
)
```
we spuriously generate a `ConstraintViolationError`, instead of a `TransactionSerializationError` (which can be automatically retried).

The issue is with `UNLESS CONFLICT`s that use "conflict CTEs" to determine whether to suppress the INSERT, rather than relying on `ON CONFLICT` to do it. Both queries start with a snapshot of an empty table, and so both try to do the INSERT; without there being an `ON CONFLICT`, postgres raises a constraint violation instead of failing to serialize the transaction.

Unfortunately, it is necessary in cases like this to not specify `ON CONFLICT`, because we depend on the lack of `ON CONFLICT` to cause an error if a single query tries to insert two conflicting objects (for example, if the query above because `FOR name IN {'foo', 'foo'}`). See https://github.com/edgedb/edgedb/issues/2193 for a detailed discussion of this issue.

Some possible approaches to this issue include:
 1. Implement the "FILTER" strategy discussed in #2193 
 2. Stick with the "ERROR" approach, but implement it by explicitly detecting and raising an error on constraint violations. This will probably be able to build on some of the machinery being built for https://github.com/edgedb/edgedb/issues/2845
 3. Retry on (some?) `ConstraintViolationError`s

msullivan:
Alright I think we're good here.

PR #2877, opened 2021-08-27
Always use serializable isolation

### LINKED ISSUE #2875: Conflicting data can be inserted into related types in concurrent transactions
If we have
```
CREATE TYPE Foo {
    CREATE PROPERTY name -> str {
        CREATE CONSTRAINT exclusive;
    };
};
CREATE TYPE Bar EXTENDING Foo;
```
and we do, in concurrent transactions `INSERT Foo { name := "foo" }` and `INSERT Bar { name := "foo" }`, both will succeed.

This is real bad.

It appears that the triggers we use for enforcing these cross-table exclusive constraints aren't getting it done here. I haven't dug into the literature yet to see what promises exactly postgres makes and whether it is possible to coax it into behaving. If not, it might be shadow table time.

msullivan:
If we stick with this, we'll probably want to yank all of our
isolation mode toggles.

Test suite time doesn't seem affected.

Will fix #2875, #2876.

### DIFF
--- a/edb/server/compiler/compiler.py
+++ b/edb/server/compiler/compiler.py
@@ -1301,8 +1301,8 @@ class Compiler:
             ctx.state.start_tx()
 
             sql = 'START TRANSACTION'
-            if ql.isolation is not None:
-                sql += f' ISOLATION LEVEL {ql.isolation.value}'
+            # N.B: Currently we ignore the isolation level. We should
+            # remove it if we keep it this way.
             if ql.access is not None:
                 sql += f' {ql.access.value}'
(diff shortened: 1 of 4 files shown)


1st1:
Can you bootstrap DB for our webapp-bench with this patch and without it? What would be the time difference?

elprans:
> Can you bootstrap DB for our webapp-bench

I did, and there is no difference at all.  Adding an exclusive constraint on the common parent type didn't make a difference either.

elprans (approved):

PR #2879, opened 2021-08-27
Always allow FINAL for CREATE SCALAR TYPE

msullivan:
It's written into migrations that were generated before.


1st1 (approved):
We have to merge this now, but the question what to do with this next is still open.

msullivan:
Actually, I think this isn't quite right, and we should only allow it for enums. I'm gonna start over

Issue #2845, opened 2021-08-21
Exclusive constraint doesn't trigger in repeated DML to another type


vpetrovykh:
- EdgeDB Version: EdgeDB 1.0-beta.4+dev5910.d2021082020.gfa96e3ffc.cv202108020000
- OS Version: linux

Steps to Reproduce:

```
test> START MIGRATION TO {
.....     module default {
.....         type Foo {
.....             required property name -> str {
.....                 constraint exclusive;
.....             }
.....             link foo -> Foo;
.....         }
..... 
.....         type Bar extending Foo;
.....     }
..... };
..... POPULATE MIGRATION;
..... COMMIT MIGRATION;
OK: START MIGRATION
OK: POPULATE MIGRATION
OK: COMMIT MIGRATION
test> WITH name := 'Alice'
..... INSERT Foo {
.....     name := name,
.....     foo := (
.....         INSERT Bar {
.....             name := name,
.....         }
.....     )
..... };
{default::Foo {id: 868ff1f8-022d-11ec-8da0-773572bc9714}}
test> select Foo {name, foo: {name}};
{
  default::Foo {name: 'Alice', foo: default::Bar {name: 'Alice'}},
  default::Bar {name: 'Alice', foo: {}},
}
```

The `exclusive` constraint on `Foo` is not deferred, so it should apply to `Bar`, therefore inserting `Alice` as both `Foo` and `Bar` should be impossible.

PR #2881, opened 2021-08-30
Correctly raise exclusive constraints when doing INSERT to related types

msullivan:
Since we can't trust triggers to work in this case, we need to do it
ourselves.

This is progress on #2845, but we still need to handle it properly for
UPDATE also.

### DIFF
--- a/edb/edgeql/compiler/viewgen.py
+++ b/edb/edgeql/compiler/viewgen.py
@@ -77,7 +77,6 @@ def process_view(
         return view_scls
 
     with ctx.newscope(fenced=True) as scopectx:
-        scopectx.in_temp_scope = True
         scopectx.path_scope.is_temporary = True
         view_path_id_ns = None
         new_path_id = path_id
--- a/edb/edgeql/compiler/viewgen.py
+++ b/edb/edgeql/compiler/viewgen.py
@@ -77,6 +77,7 @@ def process_view(
         return view_scls
 
     with ctx.newscope(fenced=True) as scopectx:
+        scopectx.in_temp_scope = True
         scopectx.path_scope.is_temporary = True
         view_path_id_ns = None
         new_path_id = path_id
--- a/edb/pgsql/compiler/dml.py
+++ b/edb/pgsql/compiler/dml.py
@@ -926,13 +926,16 @@ def compile_insert_else_body_failure_check(
     else_fail = on_conflict.else_fail
     if not else_fail:
         return
+
     # Copy the type rels from the possibly conflicting earlier DML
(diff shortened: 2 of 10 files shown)


msullivan:
This is mostly there but I realized that I didn't actually test the *nested* case, which isn't quite right

edb/pgsql/compiler/dml.py
```diff
@@ -891,6 +917,52 @@ def compile_insert_else_body(
...
+                val=pgast.StringConstant(
+                    val=(
+                        f'duplicate key value violates unique '
+                        f'constraint "{cid};schemaconstr"'
+                    )
+                ),
+            ),
+            pgast.NamedFuncArg(
+                name='constraint',
+                val=pgast.StringConstant(val=f"{cid};schemaconstr")
```
elprans, line 957: Need to pull this into `pgsql.common` to avoid hardcoding PostgreSQL object naming specifics in multiple places.

edb/edgeql/compiler/viewgen.py
```diff
@@ -77,6 +77,7 @@ def process_view(
         return view_scls
 
     with ctx.newscope(fenced=True) as scopectx:
+        scopectx.in_temp_scope = True
```
elprans, line 80: Do we need a second `in_temp_scope` flag?  Is it always correlated with `path_scope.is_temporary`?  Perhaps test for that instead (add a `ScopeTree` helper if traversal is needed)?

msullivan: I mostly did it this way to avoid chasing up the tree but that's gonna be super marginal so I think you are right

edb/schema/utils.py
```diff
@@ -626,6 +626,22 @@ def get_class_nearest_common_ancestor(
...
+
+    return nearests
+
+
+def get_class_nearest_common_ancestor(
+    schema: s_schema.Schema,
+    classes: Iterable[so.InheritingObjectT]
+) -> Optional[so.InheritingObjectT]:
+    common_list = get_class_nearest_common_ancestors(schema, classes)
+    # FIXME: This is arbitrary and probably not meaningful
```
elprans, line 644: Perhaps we could add a check to see if combining `classes` results in a meaningful MRO?  For example,

```
type A
type B
type C extending B, A
type D extending A, B

get_class_nearest_common_ancestor([C, D]) raises
```

msullivan: My plan was to go through and audit the call sites (of which there are only a single digit number) and try to work out what the best approach for each case is, but until I do that preserve the current behavior

elprans (approved):

PR #2885, opened 2021-08-31
Upgrade parsing

elprans:
I've spent a couple of evenings on seeing if I can reduce the time
needed to regenerate parser tables, and it turned out that I could:
93% of that run time is now gone.  This is all in parsing 2.0 which I
just released.  I went a little further here and tweaked our `preload()`
function to build parsers in parallel and taught `edb test` to preload
parsers.  With these changes the edit/test cycle for our parsers is a
whole lot better.


edb/server/compiler_pool/worker.py
```diff
@@ -374,7 +374,7 @@ def main():
     # worker cannot start correctly.
     max_worker_spawns = numproc * 2
 
-    ql_parser.preload()
+    ql_parser.preload(allow_rebuild=False)
```
1st1, line 377: Good.

1st1 (approved):

PR #2888, opened 2021-09-02
Fix the flakey test_schema_recursive_05

msullivan:
This has two components:
 1. Make sort_by_cross_refs prefer reporting non-trivial cycles
    over trivial ones, since anything that has a cycle with a
    computed pointers will also have a trivial cycle, since
    computed pointers get inlined.
 2. Sort the endpoints in the error message so that the output
    there is more deterministic too.

It still won't be totally deterministic which cycle gets reported, if
there are multiple ones, though.


elprans (approved):
:+1:

PR #2889, opened 2021-09-02
Fix handling of function overloads on object type parameters

elprans:
Currently, an attempt to declare a function overload with a
difference in an object type parameter crashes with an ISE:

    Error: InternalServerError: function "..." already exists with same argument types

This is because objects are passed to functions by identity and Postgres
is unable to distinguish between the implementations, since all would
have the same `uuid`-typed parameter.

A straightforward solution for this would involve adding a dummy
argument of the composite type corresponding to the object type
relation.  This will only work for compile-time dispatch, i.e.
`SELECT foo(Obj)` would always call the `foo(Obj)` variant, even if more
specific `foo()` implementations exist for subtypes of `Obj`.  This
behavior is very unintuitive and so we must do runtime function dispatch.
Unfortunately, there is no way to do such dispatch in Postgres without
resorting to `EXECUTE` in PL/pgSQL, which would be very slow, so the
solution here is to put all overloaded implementations into a single
function body and switch based on the `__type__` value passed as an
additional shadow argument for the object parameter.

This approach currently places a number of restrictions on overloads of
object type functions: 1) there must be no difference in any but one
parameter type, i.e. this is single dispatch, and 2) the names of
all parameters must match, because all implementations share a single
function definition.  The second restriction is an implementation
caveat, as we can easily normalize the names of positional arguments,
but I decided to keep that out of scope of this patch.

### DIFF
--- a/edb/pgsql/delta.py
+++ b/edb/pgsql/delta.py
@@ -832,7 +832,11 @@ class FunctionCommand:
             if compile_defaults and param_default is not None:
                 default = self.compile_default(func, param_default, schema)
 
-            args.append((param.get_parameter_name(schema), pg_at, default))
+            pn = param.get_parameter_name(schema)
+            args.append((pn, pg_at, default))
+
+            if param_type.is_object_type():
+                args.append((f'__{pn}__type', ('uuid',), None))
 
         return args
 
@@ -852,10 +856,27 @@ class FunctionCommand:
     def compile_sql_function(self, func: s_funcs.Function, schema):
         return self.make_function(func, func.get_code(schema), schema)
 
+    def _compile_edgeql_function(
+        self,
+        schema: s_schema.Schema,
+        context: sd.CommandContext,
+        func: s_funcs.Function,
+        body: s_expr.Expression,
+    ) -> s_expr.Expression:
+        return s_funcs.compile_function(
+            schema,
+            context,
+            body=body,
+            params=func.get_params(schema),
+            language=ql_ast.Language.EdgeQL,
+            return_type=func.get_return_type(schema),
+            return_typemod=func.get_return_typemod(schema),
+        )
+
     def fix_return_type(
             self, func: s_funcs.Function, nativecode, schema, context):
 
-        return_type = self._get_attribute_value(schema, context, 'return_type')
+        return_type = func.get_return_type(schema)
         ir = nativecode.irast
 
         if not (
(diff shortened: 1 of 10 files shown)


edb/schema/functions.py
```diff
@@ -1445,7 +1449,66 @@ def compile_function(
                     context=body.qlast.context,
                 )
 
-        return_type = self._get_attribute_value(schema, context, 'return_type')
+        return expr
+
+    @classmethod
+    def compile_function(
```
msullivan, line 1455: I spent a long while being confused about the new `compile_function`/`compile_this_function` before I realized that `compile_function` was a `@classmethod`. I guess a comment could be added but to be honest it's pretty clear once I realized that.

edb/pgsql/delta.py
```diff
@@ -888,7 +906,158 @@ def compile_edgeql_function(self, func: s_funcs.Function, schema, context):
...
+                        UNION ALL
+                        SELECT
+                            target AS ancestor,
+                            index
+                        FROM
+                            edgedb."_SchemaObjectType__ancestors"
+                            WHERE source = {qi(type_param_name)}
+                        ) a
+                    WHERE ancestor IN ({impl_ids})
+                    ORDER BY index
```
msullivan, line 992: Would it be worth it to try to put an index on (source, index)? Probably not, because the sets probably aren't very big?

msullivan (approved):
This is a great piece of work, and it'll be really nice to have this feature working

edb/pgsql/delta.py
elprans: There is a unique `(source, target)` index on all links, and that is also necessary to enforce proper distinctness.

edb/schema/functions.py
elprans: I think it'll be even clearer if we turn it into a function, it doesn't need to be a classmethod

edb/pgsql/metaschema.py
```diff
@@ -3966,6 +3966,32 @@ async def generate_support_views(
...
+    )
+
+    idx_bn = types.get_pointer_storage_info(
+        ObjectType_ancestors_index,
+        schema=schema,
+    ).column_name
+
+    alias_view = dbops.View(
+        name=('edgedb', f'_SchemaObjectType__ancestors'),
+        query=(f'SELECT *, {qi(idx_bn)} AS index FROM {q(*anc_bn)}')
```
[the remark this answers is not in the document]
elprans: I generalized the whole thing to use the object type view machinery.

PR #2890, opened 2021-09-02
Fix volatility inference of functions taking object arguments

elprans:
A reference to an object `Parameter` is at most `Stable`, because it
triggers a table scan.

### DIFF
--- a/edb/edgeql/compiler/inference/volatility.py
+++ b/edb/edgeql/compiler/inference/volatility.py
@@ -145,7 +145,13 @@ def __infer_set(
     env: context.Environment,
 ) -> InferredVolatility:
     if ir.rptr is not None:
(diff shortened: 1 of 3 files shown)


edb/edgeql/compiler/inference/volatility.py
```diff
@@ -209,7 +210,10 @@ def __infer_param(
     ir: irast.Parameter,
     env: context.Environment,
 ) -> InferredVolatility:
-    return IMMUTABLE
+    if irtyputils.contains_object(ir.typeref):
```
msullivan, line 213: Put the comment about the table scan here too?

msullivan (approved):

elprans:
I think I found a more precise formulation.

msullivan (approved):

Issue #2834, opened 2021-08-18
Migration fails when an object-handling function is introduced.


vpetrovykh:
- EdgeDB Version: 1-beta3 (docker)
- OS Version: linux

Steps to Reproduce:

1. Setup a project.
2. Use the following schema:
   ```
    module default {
        type Person {
            property name -> str;
        }
        function get_name(p: Person) -> str using (p.name);
    }
   ```
3. Run `edgedb migration create` and then `edgedb migrate`
   ```
    $ edgedb migration create
    did you create object type 'default::Person'? [y,n,l,c,b,s,q,?]
    > y
    did you create function 'default::get_name'? [y,n,l,c,b,s,q,?]
    > y
    Created ./dbschema/migrations/00001.edgeql, id: m1jgp4q5mu7e44wqupbm2znfhnuef7picezzz77acazp5wpyqkbjaa
    $ edgedb migrate
    edgedb error: ERROR: InternalServerError: relation "edgedbpub.2ee91729-0052-11ec-bc00-95d4256e2ffc_t" does not exist
      Hint: This is most likely a bug in EdgeDB. Please consider opening an issue ticket at https://github.com/edgedb/edgedb/issues/new?template=bug_report.md
      Server traceback:
          edb.errors.InternalServerError: relation "edgedbpub.2ee91729-0052-11ec-bc00-95d4256e2ffc_t" does not exist   
   ```

This also fails in REPL:
```
EdgeDB 1.0-beta.3+g8ab4a7be0.cv202108020000 (repl 1.0.0-beta.3)
Type \help for help, \quit to quit.
edgedb> START MIGRATION TO {
.......   module default {
.......     type Person {
.......       property name -> str;
.......     }
.......     function get_name(p: Person) -> str using (p.name);
.......   }
....... };
....... 
....... POPULATE MIGRATION;
....... COMMIT MIGRATION;
OK: START MIGRATION
OK: POPULATE MIGRATION
Error: ERROR: InternalServerError: relation "edgedbpub.4e6d4a01-0052-11ec-b910-3958c99cb3e5_t" does not exist
  Hint: This is most likely a bug in EdgeDB. Please consider opening an issue ticket at https://github.com/edgedb/edgedb/issues/new?template=bug_report.md
  Server traceback:
      edb.errors.InternalServerError: relation "edgedbpub.4e6d4a01-0052-11ec-b910-3958c99cb3e5_t" does not exist
```

[...]

PR #2892, opened 2021-09-02
Fix ISE when an object function is created alongside its object

elprans:
The culprit is the tardiness of type inheritance view creation.  Make it
eager for `CreateObjectType`.

Fixes: #2834.


msullivan (approved):
Seems reasonable

PR #2894, opened 2021-09-03
proto 0.12: Accept zero-length arguments data

tailhook:


tailhook:
@fmoor this has to be updated in Go bindings too (current change is backwards compatible, but we're going to fix that before releasing rc1 with protocol 0.12)

elprans (approved):

PR #2896, opened 2021-09-03
Better error messages on schema type mismatches

### LINKED ISSUE #1016: Bad error reporting on create type

msullivan:
Make `Schema.get` produce its own error on a type mismatch instead of
using `get_by_id`, since it has access to the name and so can produce
a better message. Also pass in the sourcectx when resolving names as
part of shell construction, so that the errors will have source
contexts.

In `declarative`, replace some type asserts with type checks and error messages, 
and add some infrastructure for producing the error messages.

Fixes #1016.

### DIFF
--- a/edb/common/english.py
+++ b/edb/common/english.py
@@ -1,0 +1,26 @@
+#
+# This source file is part of the EdgeDB open source project.
+#
+# Copyright 2009-present MagicStack Inc. and the EdgeDB authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+from __future__ import annotations
+from typing import *
+
+
+def add_a(word):
+    article = 'an' if word[0] in 'aeiou' else 'a'
+    return f'{article} {word}'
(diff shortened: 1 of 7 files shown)


edb/edgeql/declarative.py
```diff
@@ -1200,6 +1202,39 @@ def _resolve_type_expr(
...
+) -> Optional[qltracer.NamedObject]:
+
+    obj = ctx.objects.get(refname)
+
+    if obj is not None and not isinstance(obj, tracer_type):
+        obj_type = TRACER_TO_REAL_TYPE_MAP[type(obj)]
+        real_type = TRACER_TO_REAL_TYPE_MAP[tracer_type]
+        raise errors.InvalidReferenceError(
+            f'{str(refname)!r} exists, but is a '
+            f'{obj_type.get_schema_class_displayname()!r}, '
```
elprans, line 1230: Hm, perhaps we could lose the quotes here.  `'test::X' exists, but is a scalar type, not a object type` reads better to me (minus a more obvious article error).

msullivan: Sure

msullivan: I fixed the articles too

edb/common/english.py
```diff
@@ -0,0 +1,26 @@
...
+# limitations under the License.
+#
+
+
+from __future__ import annotations
+from typing import *
+
+
+def add_a(word):
+    article = 'an' if word[0] in 'aeiou' else 'a'
```
elprans, line 25: Should be close enough for our use cases.

elprans (approved):

PR #2900, opened 2021-09-04
Fix handling of the `optional` specifier in shapes

elprans:
Currently, the `optional` specifier in computed pointers in query
shapes is thoroughly ignored, because it is declared as `bool` and not
`Optional[bool]` in the AST making `optional` and absence of optionality
specifier impossible to distinguish.  Fix this.

### DIFF
--- a/edb/edgeql/compiler/inference/cardinality.py
+++ b/edb/edgeql/compiler/inference/cardinality.py
@@ -358,26 +358,26 @@ def _infer_pointer_cardinality(
     ptrcls: s_pointers.Pointer,
     ptrref: Optional[irast.BasePointerRef],
     irexpr: irast.Base,
-    specified_required: bool = False,
+    specified_required: Optional[bool] = None,
     specified_card: Optional[qltypes.SchemaCardinality] = None,
     is_mut_assignment: bool = False,
     shape_op: qlast.ShapeOp = qlast.ShapeOp.ASSIGN,
     source_ctx: Optional[parsing.ParserContext] = None,
-
     scope_tree: irast.ScopeTreeNode,
     ctx: inference_context.InfCtx,
 ) -> None:
 
     env = ctx.env
 
-    # Convert the SchemaCardinality into Cardinality used for inference.
-    if not specified_required and specified_card is None:
-        ir_specified_card = None
+    if specified_required is None:
+        spec_lower_bound = None
     else:
-        ir_specified_card = qltypes.Cardinality.from_schema_value(
-            specified_required,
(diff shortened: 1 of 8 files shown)


edb/edgeql/compiler/inference/cardinality.py
```diff
@@ -403,45 +403,47 @@ def _infer_pointer_cardinality(
         else:
             inferred_card = expr_card
 
-    if ir_specified_card is None:
+    if spec_upper_bound is None and spec_lower_bound is None:
```
msullivan, line 406: Do you think it would read cleaner if we dropped the special case, which should fall out of the rest of the logic?
(Answer could easily be no)

edb/edgeql/compiler/viewgen.py
```diff
@@ -922,11 +922,21 @@ def _normalize_view_ptr_expr(
...
-            specified_required = base_required
 
-            if (shape_el.cardinality is not None
-                    and base_ptrcls is not None
-                    and shape_el.cardinality != base_cardinality):
+            # Inferred optionality overrides that of the base pointer
+            # if pointer is not `required`.
+            if shape_el.required is not None:
+                specified_required = shape_el.required
+            elif base_required is True:
```
msullivan, line 930: Why only if it's true?

edb/edgeql/compiler/viewgen.py
```diff
@@ -922,11 +922,21 @@ def _normalize_view_ptr_expr(
             specified_required = shape_el.required
```
msullivan, line 922: The logic flow here still seems a little weird; why is all the required stuff conditioned on the base cardinality?

msullivan (approved):
Seems reasonable; had a couple questions

edb/edgeql/compiler/viewgen.py
elprans: Because we still want `foo := 1` to infer as `ONE` even if `foo` is defined as `optional` in the base type.  I'll clarify the comment above.

edb/edgeql/compiler/viewgen.py
elprans: Because if we don't know the upper bound, then we don't know the lower either.

edb/edgeql/compiler/inference/cardinality.py
elprans: It's a rather common case, so I'd leave it as an optimization (in fairness it's likely marginal, but it still elides quite a bit of code).

PR #2901, opened 2021-09-04
Implement std::assert_exists()

elprans:
The new `std::assert_exists(input: SET OF anytype) -> SET OF anytype`
function checks the cardinality of the input set, and if it is less
than 1, raises a `CardinalityViolationError`.  This assertion
complements `std::assert_single()` in that it makes it possible to
satisfy lower cardinality bound requirements in computed pointers:

    SELECT Obj {
        required link foo := assert_exists((SELECT Foo))
    }

### DIFF
--- a/docs/stdlib/set.rst
+++ b/docs/stdlib/set.rst
@@ -35,6 +35,9 @@ Set
     * - :eql:func:`assert_single`
       - :eql:func-desc:`assert_single`
 
+    * - :eql:func:`assert_exists`
+      - :eql:func-desc:`assert_exists`
+
     * - :eql:func:`count`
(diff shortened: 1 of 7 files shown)


msullivan (approved):
Looks good!

docs/stdlib/set.rst
```diff
@@ -304,6 +307,32 @@ Set
 ----------
 
 
+.. eql:function:: std::assert_exists(s: SET OF anytype) -> SET OF anytype
+
+    :index: cardinality existence
```
1st1, line 312:
```suggestion
    :index: cardinality existence empty
```

1st1 (approved):

PR #2915, opened 2021-09-08
Get rid of FOR scope hacks

elprans:
The "iterator_target" metadata together with `iterator_ctx` in context
are remnants of the now removed iterator hoisting.  They're actually
causing incorrect scope such as in the newly added
`test_edgeql_for_in_computable_11`.  This removal umasked a failure in
`test_edgeql_for_in_computable_02`, which is due to issues in how
materialization reasons are tracked across recompilation of shapes.  The
bug is unrelated as is demonstrated by
`test_edgeql_for_in_computable_02b`, which currently fails with or
without the changes in this patch.

### DIFF
--- a/edb/edgeql/compiler/context.py
+++ b/edb/edgeql/compiler/context.py
@@ -86,11 +86,6 @@ class ViewRPtr:
 
 
 @dataclasses.dataclass
-class StatementMetadata:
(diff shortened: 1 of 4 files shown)


msullivan (approved):
Great! Happy to have this gone

PR #2919, opened 2021-09-08
Update renamed command

kwikiel:
Rationale: 
warning: The 'create-migration' subcommand was renamed.
Use 'edgedb migration create' instead.


elprans (approved):
Thanks!

Issue #2158, opened 2021-01-27
Add support for remote cluster multi-tenancy


elprans:
Currently it is assumed that there is always a 1:1 correspondence between an EdgeDB instance and a remote Postgres cluster.  This isn't necessary and makes in-place EdgeDB upgrades hard to implement.  A solution would be to partition the Postgres cluster between EdgeDB "tenants" by mangling database names using a tenant id, which would be passed as a command-line argument to `edgedb-server` alongside `--postgres-dsn`.  This approach can also be used to host an arbitrary number of EdgeDB instances in a single Postgres cluster.

PR #2920, opened 2021-09-08
Adaptive Postgres HA Support

fantix:
Depends on #2580, implementation of #2887 

This PR supports a "adaptive" Postgres HA cluster which doesn't actively notify EdgeDB of failovers.

- [ ] Add tests.


edb/server/pgcon/pgcon.pyx
```diff
@@ -1590,6 +1605,18 @@ cdef class PGConnection:
                     self.connected = True
                     break
 
+                elif mtype == b'S':
+                    # ParameterStatus
+                    name, value = self.parse_parameter_status_message()
+                    if name == 'in_hot_standby':
+                        # Postgres 14 and above has this; lower Postgres is
```
elprans, line 1612: `ParameterStatus` will also be sent once the backend changes its state, so we need to handle it in the query loop also.

edb/server/pgcon/pgcon.pyx
```diff
@@ -90,6 +90,14 @@ cdef bytes INIT_CON_SCRIPT = None
 
 def _build_init_con_script() -> bytes:
     return (f'''
+        SELECT CASE WHEN pg_is_in_recovery() THEN
```
elprans, line 93: Perhaps make this conditional on PostgreSQL<=13 to avoid paying the price of the query on modern versions.

setup.py
```diff
@@ -55,7 +55,7 @@
     'parsing~=2.0',
     'psutil~=5.8',
     'setproctitle~=1.2',
-    'setuptools_scm~=6.0',
+    'setuptools_scm~=6.0,!=6.3.0',
```
elprans, line 58: Curious, what's wrong with 6.3.0?

fantix: I think it's [missing a dependency](https://github.com/pypa/setuptools_scm/issues/625)

edb/server/ha/base.py
```diff
@@ -0,0 +1,54 @@
...
+        raise NotImplementedError
+
+    def get_master_addr(self) -> Optional[Tuple[str, int]]:
+        raise NotImplementedError
+
+
+def get_backend(parsed_dsn: urllib.parse.ParseResult) -> Optional[HABackend]:
+    backend, _, sub_scheme = parsed_dsn.scheme.partition("+")
+    if backend == "stolon":
+        from . import stolon
```
1st1, line 50: Any reason for a dynamic import?

edb/server/ha/passive.py
```diff
@@ -0,0 +1,121 @@
...
+            self._unhealthy_timer_handle = (
+                asyncio.get_running_loop().call_later(
+                    UNHEALTHY_MIN_TIME, self._maybe_failover
+                )
+            )
+            self._pgcon_count = max(
+                self._cluster_protocol.get_active_pgcon_num(), 0
+            ) + 1
+            logger.warning(
+                "Passive HA cluster is unhealthy. "
```
1st1, line 73: Is "passive HA" some established terminology? Can we drop "passive" if it's not?

edb/server/ha/passive.py
```diff
@@ -0,0 +1,121 @@
...
+
+class State(enum.Enum):
+    HEALTHY = 1
+    UNHEALTHY = 2
+    FAILOVER = 3
+
+
+class PassiveHASupport:
+    _state: State
+    _unhealthy_timer_handle: Optional[asyncio.TimerHandle]
```
1st1, line 42: I like how you encapsulated the state machine in this class, but can you also add a docstring summarizing it? (You can copy the relevant bits from the GH discussion)

edb/server/ha/stolon.py
```diff
@@ -0,0 +1,243 @@
...
+
+    async def _start_watching(self):
+        raise NotImplementedError
+
+    def on_cluster_data(self, data):
+        cluster_status = data.get("cluster", {}).get("status", {})
+        master_db = cluster_status.get("master")
+        cluster_phase = cluster_status.get("phase")
+        if cluster_phase != "normal":
+            logger.debug("Stolon cluster phase: %s", cluster_phase)
```
1st1, line 91: nit:
```suggestion
            logger.debug("Stolon cluster phase: %r", cluster_phase)
```

an so in other places; that will add quotes to strings making it easier to read logs

edb/server/ha/stolon.py
```diff
@@ -0,0 +1,243 @@
...
+
+    def data_received(self, data: bytes):
+        self._parser.feed_data(data)
+
+    def connection_lost(self, exc):
+        self._consul_backend.connection_lost()
+
+    def on_status(self, status: bytes):
+        if self._parser.get_status_code() != 200:
+            self._transport.close()
```
1st1, line 163: Should we add debug logging here?

edb/server/pgcon/pgcon.pyx
```diff
@@ -289,8 +305,26 @@ async def connect(connargs, dbname, tenant_id):
             ignore_data=True,
         )
 
-    if INIT_CON_SCRIPT is None:
-        INIT_CON_SCRIPT = _build_init_con_script()
+    if 'in_hot_standby' in pgcon.parameter_status:
+        # in_hot_standby is present in Postgres 14 and above
```
1st1, line 309: Is it *always* present?

```suggestion
        # in_hot_standby is always present in Postgres 14 and above
```

edb/server/pgcon/pgcon.pyx
```diff
@@ -435,9 +474,12 @@ cdef class PGConnection:
             self.msg_waiter = None
 
     def set_server(self, server):
-        assert defines.EDGEDB_SYSTEM_DB in self.dbname
         self.server = server
 
+    def mark_as_system_db(self):
+        assert defines.EDGEDB_SYSTEM_DB in self.dbname
```
1st1, line 480: Why does this check use `in` and not `==`?

edb/server/pgcon/pgcon.pyx
```diff
@@ -1783,6 +1835,15 @@ cdef class PGConnection:
 
         self.buffer.finish_message()
 
+    cdef parse_parameter_status_message(self):
+        cdef:
+            str name
+            str value
+        assert self.buffer.get_message_type() == b'S'
+        name = self.buffer.read_null_str().decode()
+        value = self.buffer.read_null_str().decode()
```
1st1, line 1844: I'd add debug logging here.

edb/server/server.py
```diff
@@ -109,13 +112,17 @@ def __init__(
...
         startup_script: Optional[srvargs.StartupScript] = None,
+        backend_passive_ha: bool = False,
     ):
 
         self._loop = asyncio.get_running_loop()
 
         # Used to tag PG notifications to later disambiguate them.
         self._server_id = str(uuid.uuid4())
 
+        # Increase-only counter to reject outdated connects
```
1st1, line 123:
```suggestion
        # Increase-only counter to reject outdated connections
```

1st1 (approved):
Looks good to me. I only have a few nit comments.

edb/server/ha/base.py
fantix: This was to break a circular import `ha.base` <-> `ha.base.stolon`

edb/server/ha/passive.py
fantix: Per discussion, we'll rename it to "adaptive".

edb/server/pgcon/pgcon.pyx
fantix: Yes according to my test, I tried to read the Postgres GUC code but it may take too much time.

edb/server/pgcon/pgcon.pyx
fantix: It's related to multi-tenancy: https://github.com/edgedb/edgedb/issues/2158

Issue #2926, opened 2021-09-09
Overloaded function cannot be deleted


vpetrovykh:
- EdgeDB Version: EdgeDB 1.0-beta.4+dev5960.d2021090721.gdac3f0c40.cv202109030000
- OS Version: linux

It used to work in `1-beta3`, so it must be a recent regression.

Steps to Reproduce:

```
db> create type Person {
...     create property name -> str;
...     create property strength -> int16;
... };
OK: CREATE TYPE
db> create function fight(one: Person, two: Person) -> str 
... using (one.name ++ two.name);
OK: CREATE FUNCTION
db> create function fight(names: str, one: int16, two: str) -> str
... using (names ++ <str>one ++ two);
OK: CREATE FUNCTION
db> drop function fight(one: Person, two: Person);
edgedb error: InternalServerError: unexpected overload on multiple parameters in object-taking function
  Hint: This is most likely a bug in EdgeDB. Please consider opening an issue ticket at https://github.com/edgedb/edgedb/issues/new?template=bug_report.md
  Server traceback:
      Traceback (most recent call last):
        File "/home/victor/dev/magicstack/edgedb/edb/server/compiler_pool/worker.py", line 308, in worker
          res = meth(*args)
        File "/home/victor/dev/magicstack/edgedb/edb/server/compiler_pool/worker.py", line 176, in compile
          units, cstate = COMPILER.compile(
        File "/home/victor/dev/magicstack/edgedb/edb/server/compiler/compiler.py", line 2059, in compile
          units = self._compile(ctx=ctx, source=source)
        File "/home/victor/dev/magicstack/edgedb/edb/server/compiler/compiler.py", line 1621, in _compile
          return self._try_compile(ctx=ctx, source=source)
        File "/home/victor/dev/magicstack/edgedb/edb/server/compiler/compiler.py", line 1674, in _try_compile
          comp, capabilities = self._compile_dispatch_ql(ctx, stmt)
        File "/home/victor/dev/magicstack/edgedb/edb/server/compiler/compiler.py", line 1568, in _compile_dispatch_ql
          self._compile_and_apply_ddl_stmt(ctx, ql),
        File "/home/victor/dev/magicstack/edgedb/edb/server/compiler/compiler.py", line 769, in _compile_and_apply_ddl_stmt
          return self._compile_and_apply_ddl_stmt(ctx, cm)
[...]

PR #2927, opened 2021-09-09
Fix object function overload validation and handling

elprans:
This patch narrows the function overload check to only the functions
that are actually overloaded with a different object type in the same
argument, so

    func(a: Foo, b: int32)
    func(a: Bar, b: int64)

is illegal, but

    func(a: Foo, b: int32)
    func(a: Bar, b: int32)
    func(a: str, b: int32, c: int64)

is perfectly OK.  This also consolidates the validation together with
the search for the overloaded parameter position so there is no
potential for internal discrepancies.

Fixes: #2926.

### DIFF
--- a/edb/schema/functions.py
+++ b/edb/schema/functions.py
@@ -26,6 +26,7 @@ from typing import *
 
 from edb import errors
 
+from edb.common import parsing
 from edb.common import struct
 from edb.common import verutils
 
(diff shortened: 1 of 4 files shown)


1st1:
Do we currently have any functions that have overloads with different sets of parameters?

msullivan (approved):

elprans:
> Do we currently have any functions that have overloads with different sets of parameters?

In the stdlib?  I don't think we have _any_ object-receiving functions there at all, let alone overloaded ones.  This is purely for used-defined stuff.

1st1:
> In the stdlib? I don't think we have any object-receiving functions there at all, let alone overloaded ones.

I was asking about _any_ functions. I don't like the fact that we allow overloaded variants to have extra parameters.

elprans:
> I was asking about _any_ functions. I don't like the fact that we allow overloaded variants to have extra parameters.

Yes, various constructors, such as `to_datetime()` use differing numbers of arguments.  Why is this a problem?

1st1:
> Yes, various constructors, such as to_datetime() use differing numbers of arguments.

Ah, that's right. Never mind then.

edb/schema/functions.py
```diff
@@ -1251,6 +1252,119 @@ def get_dummy_body(self, schema: s_schema.Schema) -> s_expr.Expression:
...
+                        f'function with differences in the names of '
+                        f'parameters is not supported',
+                        context=srcctx,
+                        details=(
+                            f"Other function is defined as `{other_sig}`"
+                        )
+                    )
+
+                diff_param = this_diff_param
+                overloads.append(f)
```
1st1, line 1361: You can only check against the first overload and then break the top loop.

elprans: I can't: I need all matching overloads to construct the common function body.

PR #2930, opened 2021-09-10
Make computed pointers consistent.

### LINKED ISSUE #2558: Make behavior of explicit computable links consistent with shapes
- EdgeDB Version: beta2

Steps to Reproduce:

Currently the following schema treats `User.tweets` as detached in queries:

```ts
module default {
  type User {
    property name -> str;
    multi link tweets := User.<author[IS Tweet];
    # treated differently than:
    # multi link tweets := .<author[IS Tweet];
  }
  type Tweet {
    link author -> User;
  }
}
```

Moreover, switching between the explicit and implicit "User" is treated as a schema change by `edgedb create-migration`.

These should be treated as equivalent for the sake of consistency with how the dot shortcut works inside query shapes: https://edgedb.slack.com/archives/C09QSK72B/p1621902198005900

vpetrovykh:
Computed links and properties defined in the schema should behave the
same way as the equivalent computed links and properties defined ad-hoc
directly in the query.

Fixes #2558

### DIFF
--- a/edb/edgeql/compiler/setgen.py
+++ b/edb/edgeql/compiler/setgen.py
@@ -1282,8 +1282,13 @@ def computable_ptr_set(
     newctx: Callable[[], ContextManager[context.ContextLevel]]
 
     if qlctx is None:
-        # Schema-level computable, completely detached context
-        newctx = ctx.detached
+        # Schema-level computed link or property, the context should
+        # still have a source.
+        newctx = _get_schema_computed_ctx(
+            rptr=rptr,
+            source=source_set,
+            ctx=ctx)
+
     else:
         newctx = _get_computable_ctx(
             rptr=rptr,
@@ -1339,6 +1344,38 @@ def computable_ptr_set(
     maybe_materialize(ptrcls, comp_ir_set, ctx=ctx)
 
     return comp_ir_set
+
+
+def _get_schema_computed_ctx(
+    *,
+    rptr: irast.Pointer,
+    source: irast.Set,
+    ctx: context.ContextLevel
+) -> Callable[[], ContextManager[context.ContextLevel]]:
+
+    @contextlib.contextmanager
+    def newctx() -> Iterator[context.ContextLevel]:
+        with ctx.detached() as subctx:
+            source_stype = get_set_type(source, ctx=ctx)
+
+            source_scope = pathctx.get_set_scope(rptr.source, ctx=ctx)
+            if source_scope and source_scope.namespaces:
+                subctx.path_id_namespace |= source_scope.namespaces
+
+            inner_path_id = pathctx.get_path_id(source_stype, ctx=subctx)
+
+            remapped_source = new_set_from_set(
+                rptr.source,
+                rptr=rptr.source.rptr,
+                preserve_scope_ns=True,
+                ctx=ctx,
+            )
+            key = inner_path_id.strip_namespace(inner_path_id.namespace)
+            subctx.view_map[key] = ((inner_path_id, remapped_source),)
+
+            yield subctx
+
+    return newctx
 
 
 def _get_computable_ctx(
--- a/edb/edgeql/compiler/context.py
+++ b/edb/edgeql/compiler/context.py
@@ -658,6 +658,7 @@ class ContextLevel(compiler.ContextLevel):
                 self.anchors = prevlevel.anchors.copy()
                 self.modaliases = prevlevel.modaliases.copy()
                 self.aliased_views = collections.ChainMap()
+                self.view_map = collections.ChainMap()
                 self.class_view_overrides = {}
                 self.expr_exposed = prevlevel.expr_exposed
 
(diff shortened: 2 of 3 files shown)


edb/edgeql/compiler/setgen.py
```diff
@@ -1346,6 +1351,51 @@ def computable_ptr_set(
...
+                source_stype, ctx=subctx).merge_namespace(subns)
+
+            subctx.pending_stmt_full_path_id_namespace = frozenset(subns)
+
+            remapped_source = new_set_from_set(
+                rptr.source, rptr=rptr.source.rptr,
+                preserve_scope_ns=True, ctx=ctx)
+            key = inner_path_id.strip_namespace(inner_path_id.namespace)
+            old = subctx.view_map.get(key, ())
+            subctx.view_map[key] = ((inner_path_id, remapped_source),) + old
```
elprans, line 1392: You still want this to be `detached` and you only need to populate the `view_map` in the detached context, so something like this:

```suggestion
        with ctx.detached() as subctx:
            source_stype = get_set_type(source, ctx=ctx)
            subctx.view_map = ctx.view_map.new_child()

            source_scope = pathctx.get_set_scope(rptr.source, ctx=ctx)
            if source_scope and source_scope.namespaces:
                subctx.path_id_namespace |= source_scope.namespaces

            inner_path_id = pathctx.get_path_id(source_stype, ctx=subctx)

            remapped_source = new_set_from_set(
                rptr.source,
                rptr=rptr.source.rptr,
                preserve_scope_ns=True,
                ctx=ctx,
            )
            key = inner_path_id.strip_namespace(inner_path_id.namespace)
            old = subctx.view_map.get(key, ())
            subctx.view_map[key] = ((inner_path_id, remapped_source),) + old
```

vpetrovykh: Updated.

elprans: Actually, even this version is too liberal, you don't want to preserve previous values of `view_map` in the detached context either, it should be exactly one entry.  Apply this patch:

```patch
diff --git a/edb/edgeql/compiler/context.py b/edb/edgeql/compiler/context.py
index 956806aa7..7252f82c6 100644
--- a/edb/edgeql/compiler/context.py
+++ b/edb/edgeql/compiler/context.py
@@ -658,6 +658,7 @@ class ContextLevel(compiler.ContextLevel):
                 self.anchors = prevlevel.anchors.copy()
                 self.modaliases = prevlevel.modaliases.copy()
                 self.aliased_views = collections.ChainMap()
+                self.view_map = collections.ChainMap()
                 self.class_view_overrides = {}
                 self.expr_exposed = prevlevel.expr_exposed

diff --git a/edb/edgeql/compiler/setgen.py b/edb/edgeql/compiler/setgen.py
index 530df46b9..66ee0bb2c 100644
--- a/edb/edgeql/compiler/setgen.py
+++ b/edb/edgeql/compiler/setgen.py
@@ -1357,8 +1357,6 @@ def _get_schema_computed_ctx(
     def newctx() -> Iterator[context.ContextLevel]:
         with ctx.detached() as subctx:
             source_stype = get_set_type(source, ctx=ctx)
-            subctx.view_map = ctx.view_map.new_child()
-
             source_scope = pathctx.get_set_scope(rptr.source, ctx=ctx)
             if source_scope and source_scope.namespaces:
                 subctx.path_id_namespace |= source_scope.namespaces
@@ -1372,8 +1370,7 @@ def _get_schema_computed_ctx(
                 ctx=ctx,
             )
             key = inner_path_id.strip_namespace(inner_path_id.namespace)
-            old = subctx.view_map.get(key, ())
-            subctx.view_map[key] = ((inner_path_id, remapped_source),) + old
+            subctx.view_map[key] = ((inner_path_id, remapped_source),)

             yield subctx

```

May also be worth writing a query that has a computed link referring to the type of an enclosing shape, e.g.:

```
SELECT Foo {
   bar: {
      bar_comp
   }
}
```

where `bar_comp` has references to `Foo`.  These *must not* be correlated with `Foo` in the query.

vpetrovykh: > May also be worth writing a query that has a computed link referring to the type of an enclosing shape

Oh, cool, I'll add a test for that.

vpetrovykh: > Actually, even this version is too liberal, you don't want to preserve previous values of `view_map` in the detached context either, it should be exactly one entry.

So... this "previous value" of `view_map` would be coming from where exactly? A `WITH` block of the `CREATE TYPE` command? I'm trying to understand a way to _have_ some previous values there that would not be valid in the nested scope.

elprans: Nested computables, e.g. if you turn the above test to:

```
SELECT Spam {
   foo := (SELECT Foo {
     bar := (SELECT Bar {
        bar_comp
     })
   })
}
```

`bar_comp` should only have a mapping for `Bar -> Spam.foo.bar`, but not for `Foo -> Spam.foo`

elprans (approved):

PR #2931, opened 2021-09-10
Get rid of setuptools-scm

elprans:
We aren't actually benefiting from `setuptools-scm` anymore because we
call `git` and format our version manually anyway, and avoiding the
`setuptools-scm` dependency removes some of it's transivite deps from
the runtime list.

This also moves `edb.server.buildmeta` to `edb.buildmeta`, because it
represents the entire distribution not just the server part and reduces
the possibility of import issues at `setup.py` time.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,7 +24,8 @@ __pycache__/
 docs/_build
 /AUTHORS
 /ChangeLog
-tests/dumps/**/*_dev*.dump
+/tests/dumps/**/*_dev*.dump
+/edb/_buildmeta.py
 
 /.coverage*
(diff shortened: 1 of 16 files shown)


fantix (approved):
Looks good to me!

nit: can we move `get_version()`, `get_version_dict()` and `get_version_json()` to their original place before `get_version_from_scm()` so that Git history looks nicer?

elprans:
> nit: can we move `get_version()`, `get_version_dict()` and `get_version_json()` to their original place before `get_version_from_scm()` so that Git history looks nicer?

Done.

PR #2939, opened 2021-09-15
Fix "compiler worker process closed" issue

fantix:
This should fix the "the connection to the compiler worker process is unexpectedly closed" issue.

### DIFF
--- a/edb/server/compiler_pool/pool.py
+++ b/edb/server/compiler_pool/pool.py
@@ -452,9 +452,9 @@ class Pool(amsg.ServerProtocol):
 
         return preargs, callback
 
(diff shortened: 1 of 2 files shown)


edb/server/compiler_pool/pool.py
```diff
@@ -452,9 +452,9 @@ def sync_worker_state_cb(
 
         return preargs, callback
 
-    async def _acquire_worker(self):
+    async def _acquire_worker(self, **kwargs):
```
elprans, line 455: Let's spell the arguments explicitly (I think it's just `condition` here).

elprans (approved):

PR #2941, opened 2021-09-16
RC1 Changelog.

vpetrovykh:
This is the first iteration. I assume we'll add more details potentially about the way we process environment and default settings, etc. This would affect CLI and bindings.

### DIFF
--- a/docs/changelog/1_0_rc1.rst
+++ b/docs/changelog/1_0_rc1.rst
@@ -1,0 +1,162 @@
+========
+1.0 RC 1
+========
+
+This changelog summarizes new features and breaking changes in
+EdgeDB 1.0 RC 1 "Ran".
+
+
+Migrations
+==========
+
+We continue fixing bugs that affect schema definitions and migrations:
+
+* Always escape newlines when string literals appear in automatically
+  generated code, such as in migrations (:eql:gh:`#2704`).
+* Fix behavior of ``SET MULTI`` to make sure that it does not imply
+  ``SET REQUIRED`` (:eql:gh:`#2778`).
+* Make properties non-distinct by default (:eql:gh:`#2853`).
+
+  All values in EdgeQL are considered to be :ref:`multisets
+  <ref_eql_fundamentals_set>`, which are a generalization of the set
+  concept that allows duplicate elements. For brevity we refer to
+  multisets as just "sets", and use "proper set" or "distinct set"
+  when referring to proper mathematical sets.
+
+  In EdgeDB the data is modeled as a directed graph, where objects are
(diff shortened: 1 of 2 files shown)


docs/changelog/1_0_rc1.rst
```diff
@@ -0,0 +1,139 @@
...
+========
+
+This changelog summarizes new features and breaking changes in
+EdgeDB 1.0 RC 1 "Ran".
+
+
+EdgeQL
+======
+
+* Drop ``FINAL`` from the syntax (:eql:gh:`#2607`).
```
1st1, line 12: Have we actually dropped the syntax? @msullivan

vpetrovykh: From what I understand we no longer support it for most cases:
```
db> create final scalar type foo extending str;
error: UnsupportedFeatureError: FINAL is not supported
  ┌─ query:1:1
  │
1 │ create final scalar type foo extending str;
  │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ error
```
But I guess we have a way of allowing it in dumps and migrations: #2886

So as a language feature it's not longer part of any actual construct and is deprecated, but not outright an error for some legacy cases. Which I think makes it fair to say that we've "dropped" it as we don't actually want anyone to use it.

vpetrovykh:
I've added the "Bindings" section with a mention of new protocol version. The release versions for Python and JS bindings are just guesses based on the last version that I see for those bindings. The Go driver version appears to be a recent release, so I used the version we already released (around Aug 26).

vpetrovykh:
I've also reorganized a bit by splitting off schema-related stuff from the EdgeQL section into the Migrations section.

docs/changelog/1_0_rc1.rst
```diff
@@ -0,0 +1,162 @@
...
+  sets expensive due to explicit elimination of duplicates.
+  Furthermore, non-distinct scalar sets are actually *desirable* in
+  many queries, especially where analytics and tuples are involved.
+  Finally, we should enforce consistent multiplicity rules on computed
+  properties, and static inference of multiplicity on scalars is weak
+  and will effectively force users to pollute queries with pointless
+  and expensive :eql:op:`DISTINCT`.
+
+  Here we change EdgeDB behavior to allow multiplicity greater than
+  one in ``multi`` properties.
```
1st1, line 49:
```suggestion
  Here we change EdgeDB behavior to allow multiplicity *greater than
  one* in ``multi`` properties.
```

1st1 (approved):

PR #2946, opened 2021-09-20
Bugfixes and tests for HA and backend connection

fantix:
* bugfix: no backend SSL through UNIX socket
* bugfix: Stolon backend failover not working in particular situation
* improve: close pgcon properly when retrying with non-SSL connection
* test: added tests for adaptive backend HA
* test: ported asyncpg connection test for backend pgcon


edb/server/ha/adaptive.py
```diff
@@ -21,11 +21,14 @@
...
 import enum
 import logging
+import os
 
 from . import base
 
 
-UNHEALTHY_MIN_TIME = 30
+UNHEALTHY_MIN_TIME = int(os.getenv(
+    'EDGEDB_BACKEND_ADAPTIVE_HA_UNHEALTHY_MIN_TIME', 30
```
1st1, line 30: Is this a public API?

edb/testbase/server.py
```diff
@@ -1806,3 +1814,27 @@ def get_cases_by_shard(cases, selected_shard, total_shards, verbosity, stats):
               f' / {int(total_est / 60)}m {int(total_est % 60)}s, '
               f'{len(setups)}/{setup_count} databases to setup.')
     return cases
+
+
+def find_available_port(port_range=(49152, 65535), max_tries=1000):
```
1st1, line 1819: I think we already have this function defined somewhere. Also, perhaps, instead of a loop, it should just create a tmp socket pair (or create a socket and bind to 0) and reuse its port.

1st1 (commented):
overall lgtm

edb/testbase/server.py
fantix: > I think we already have this function defined somewhere.

Right, it was removed in [#2251](https://github.com/edgedb/edgedb/pull/2251/files#diff-04df2656d0211c7f0c32b3ae2705297d32ea26bcdb7c4f9f1282b3a9016eb25dL43).

> create a tmp socket pair (or create a socket and bind to 0) and reuse its port.

I'm not sure what you meant by socket pair (they are UNIX sockets right?), but I'm changing the method into "bind to 0".

1st1: > I'm not sure what you meant by socket pair (they are UNIX sockets right?), 

You can pass AF_INET family to `socketpair()`.

> but I'm changing the method into "bind to 0".

Whatever is simpler.

edb/server/ha/adaptive.py
fantix: I see - the value itself is not test-only, so let me create a new PR to document about the HA support. What about `docs/internals/backend_ha.rst`?

1st1: > I see - the value itself is not test-only

Then use the `EDGEDB_SERVER` prefix? I'm confused... How is that the variable is for the tests but its value isn't?

>  What about docs/internals/backend_ha.rst?

Sure

setup.py
```diff
@@ -93,7 +93,6 @@
     'types-MarkupSafe~=1.1',
     'types-pkg-resources~=0.1.3',
     'types-typed-ast~=1.4.2',
-    'types-requests~=2.25.6',
```
elprans, line 96: `requests-xml` has `requests` as a transitive dependency, so we have to keep this, I think.

fantix: Right, actually I added it when I was using `requests` in the test - but keeping it is also fine, in case someone needs requests in the future.

elprans (approved):

PR #2949, opened 2021-09-21
pgsql.delta: Remove the uses of ad-hoc priority and rework inh views

elprans:
There are also preparatory and post- cleanup commits here, please review individually.

### DIFF
--- a/edb/pgsql/delta.py
+++ b/edb/pgsql/delta.py
@@ -161,9 +161,7 @@ class MetaCommand(sd.Command, metaclass=CommandMeta):
         return schema
 
     def generate(self, block: dbops.PLBlock) -> None:
-        for op in sorted(
-                self.pgops, key=lambda i: getattr(i, 'priority', 0),
-                reverse=True):
+        for op in self.pgops:
             op.generate(block)
 
     @classmethod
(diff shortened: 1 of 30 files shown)


edb/pgsql/delta.py
```diff
@@ -2044,21 +2073,13 @@ def _redo_everything(
...
         cmd = sd.DeltaRoot()
 
-        for prop, new_typ in props:
-            delta_alter, cmd_alter, alter_context = prop.init_delta_branch(
+        for prop, _ in props:
+            delta_alter, cmd_alter, _ = prop.init_delta_branch(
                 schema, context, cmdtype=sd.AlterObject)
-            cmd_alter.set_attribute_value(
-                'target', prop.get_target(orig_schema))
             cmd_alter.set_attribute_value(
```
msullivan, line 2077: Why does the default update need to get split off?

edb/pgsql/delta.py
```diff
@@ -3425,13 +3502,7 @@ def _alter_pointer_type(self, pointer, schema, orig_schema, context):
...
 
         if changing_col_type:
-            self.schedule_inhviews_update(
-                schema,
-                context,
-                source,
-                update_descendants=True,
-                update_ancestors=True,
-            )
+            self.create_inhview(schema, context, source)
```
msullivan, line 3502: Is the idea that we do the batching for things like property creation that would be very expensive to do eagerly, but just do things eagerly for less hot things? (Like here?)

msullivan (approved):
This is great! Really nice cleanups. I wonder if after these cleanups it might be getting close to possible to turn on type-checking here...

edb/pgsql/delta.py
elprans: Because the default expression might reference objects that need to be re-created first.

edb/pgsql/delta.py
elprans: Yeah.  Batching is an optimization and needs to be used sparingly and carefully, lest the original problems that we're attempting to solve here recur.

msullivan: :+1:

1st1 (approved):
Impressive cleanup

PR #2950, opened 2021-09-21
Add `suggested_pool_concurrency` server-side setting

1st1:


edb/server/server.py
```diff
@@ -150,6 +150,10 @@ def __init__(
         self._max_backend_connections = max_backend_connections
         self._compiler_pool = None
         self._compiler_pool_size = compiler_pool_size
+        self._suggested_client_pool_size = max(
+            min(compiler_pool_size, defines.MAX_SUGGESTED_CLIENT_POOL_SIZE),
```
elprans, line 154: Why is client pool size determined by the compiler pool size?  I think it should be a function of the backend connection pool size, since most queries don't go into the compiler.

1st1: Yeah, it's a copy/paste typo, good catch. Fixed.

elprans (approved):

fantix:
👍

tailhook:
@1st1 what is the point of minimum of 10? I don't think there is too much vaiue in having more than backend connections. And `BACKEND_CONNECTIONS_MIN` is 4.

PR #2956, opened 2021-09-22
Use ubuntu-latest and Python 3.10 in CI

fantix:
See also: https://github.com/pypa/pip/issues/5865

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -63,7 +63,7 @@ DOCS_DEPS = [
     'docutils~=0.17.0',
     'lxml~=4.6.3',
     'Pygments~=2.10.0',
-    'Sphinx~=4.1.2',
+    'Sphinx~=4.2.0',
     'sphinxcontrib-asyncio~=0.3.0',
 ]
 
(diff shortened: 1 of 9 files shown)


fantix:
Sphinx 4.2.0 is needed for Python 3.10

setup.py
```diff
@@ -100,6 +100,7 @@
     CYTHON_DEPENDENCY,
     'packaging>=21.0',
     'setuptools-rust~=0.12.1',
+    'wheel~=0.37.0',  # needed by PyYAML and immutables, refs pypa/pip#5865
```
elprans, line 103: Ugh. This could be a problem in edgedb-pkg too.  Could you elaborate on how it fails, please?

fantix: `wheel` and `Cython` are in the [build `requires` of `PyYAML`](https://github.com/yaml/pyyaml/blob/master/pyproject.toml), and `pip download` won't fetch them at all. Because `PyYAML` hasn't provided a wheel release for Python 3.10, installing from source tarball will fail in the CI here:

https://github.com/edgedb/edgedb/blob/cb61a47469c282e2219875747dd3257d528adfc5/.github/workflows/tests.yml#L46-L47

Same for `immutables` but only `wheel` is needed.

elprans: Oh, so I'd remove the version pin from `wheel` then.

elprans (approved):

PR #2958, opened 2021-09-22
Add mutation tests.

vpetrovykh:
Nothing too fancy, just a few more mutation tests that deal with link properties. No failures.


msullivan (approved):
Good tests, happy we mass them all :)

PR #2959, opened 2021-09-23
Enable compile-time multiplicity enforcement for computed links

elprans:
A computed link should be indistinguishable from a regular link in queries.
This means that just like regular links, a computed link
_must_ resolve to a distinct set.  After this, it is a compile-time error
to use any expression that could not be proved to have multiplicity of at
most 1.  Queries where static inference is impossible should use
`assert_distinct` or `DISTINCT` depending on whether duplicates are
expected.

Fixes: #117

### DIFF
--- a/edb/edgeql/compiler/inference/cardinality.py
+++ b/edb/edgeql/compiler/inference/cardinality.py
@@ -660,6 +660,16 @@ def __infer_func_call(
         _, arg_upper = arg_card
         return _bounds_to_card(CardinalityBound.ONE, max(arg_upper))
 
+    elif ir.func_shortname == sn.QualName('std', 'assert_distinct'):
+        arg_cards = []
+
+        for arg in ir.args:
+            arg.cardinality = infer_cardinality(
+                arg.expr, scope_tree=scope_tree, ctx=ctx)
+            arg_cards.append(arg.cardinality)
+
+        return max_cardinality(arg_cards)
+
     elif ir.preserves_optionality:
         # This is a generic aggregate function which preserves the
         # optionality of its generic argument.  For simplicity we
--- a/edb/edgeql/compiler/inference/cardinality.py
+++ b/edb/edgeql/compiler/inference/cardinality.py
@@ -47,8 +47,9 @@ from edb.ir import utils as irutils
 from edb.ir import typeutils
 from edb.edgeql import ast as qlast
 
-from . import volatility
 from . import context as inference_context
+from . import utils as inf_utils
+from . import volatility
 
 from .. import context
 
@@ -112,24 +113,6 @@ def _bounds_to_card(
         lower.as_required(),
         upper.as_schema_cardinality(),
     )
-
-
-def _get_set_scope(
-        ir_set: irast.Set,
-        scope_tree: irast.ScopeTreeNode,
-        ctx: inference_context.InfCtx) -> irast.ScopeTreeNode:
-
-    if ir_set.path_scope_id:
-        new_scope = ctx.env.scope_tree_nodes.get(ir_set.path_scope_id)
-        if new_scope is None:
-            raise errors.InternalServerError(
-                f'dangling scope pointer to node with uid'
-                f':{ir_set.path_scope_id} in {ir_set!r}'
-            )
-    else:
-        new_scope = scope_tree
-
-    return new_scope
 
 
 def cartesian_cardinality(
@@ -337,22 +320,6 @@ def __infer_type_introspection(
     return ONE
 
 
-def _find_visible(
-    ir: irast.Set,
-    scope_tree: irast.ScopeTreeNode,
-) -> Optional[irast.ScopeTreeNode]:
-    parent_branch = scope_tree.parent_branch
-    if parent_branch is not None:
-        if scope_tree.namespaces:
-            path_id = ir.path_id.strip_namespace(scope_tree.namespaces)
-        else:
-            path_id = ir.path_id
-
-        return parent_branch.find_visible(path_id)
-    else:
-        return None
-
-
 def _infer_pointer_cardinality(
     *,
     ptrcls: s_pointers.Pointer,
@@ -501,7 +468,7 @@ def _infer_shape(
     ctx: inference_context.InfCtx,
 ) -> None:
     for shape_set, shape_op in ir.shape:
-        new_scope = _get_set_scope(shape_set, scope_tree, ctx=ctx)
+        new_scope = inf_utils.get_set_scope(shape_set, scope_tree, ctx=ctx)
         if shape_set.expr and shape_set.rptr:
             ptrref = shape_set.rptr.ptrref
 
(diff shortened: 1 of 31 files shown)


tests/test_edgeql_update.py
```diff
@@ -2937,10 +2939,10 @@ async def test_edgeql_update_add_dupes_03(self):
                     UPDATE UpdateTestSubSubType
                     FILTER .name = 'add-dupes'
                     SET {
-                        tags += {
+                        tags += assert_distinct({
                             (UPDATE Tag FILTER .name = 'fun' SET {flag := 1}),
                             (UPDATE Tag FILTER .name = 'wow' SET {flag := 2}),
-                        }
+                        })
```
vpetrovykh, line 2945: Do we actually *need* an `assert_distinct` here? I thought that `+=` and `-=` already will make linked sets distinct in the end. I mean if I already have `Tag` "wow" linked as `tags`, doing `tags += (select Tag filter .name = 'wow')` should be legal, so who cares if the set after `+=` or `-=` is not distinct if the result must be reconciled anyways?

vpetrovykh (approved):
LGTM. I do have a question about one of the tests, but it probably doesn't impact the PR overall.

edb/edgeql/qltypes.py
```diff
@@ -172,10 +174,35 @@ def _missing_(cls, name):
...
+        return self is MultiplicityValue.ONE
+
+    def is_many(self) -> bool:
+        return self is MultiplicityValue.MANY
+
+    def is_zero(self) -> bool:
+        return self is MultiplicityValue.ZERO
+
+
+@dataclasses.dataclass(frozen=True, eq=False)
```
msullivan, line 193: Probably worth adding a quick comment here

edb/edgeql/compiler/inference/cardinality.py
```diff
@@ -660,6 +660,16 @@ def __infer_func_call(
         _, arg_upper = arg_card
         return _bounds_to_card(CardinalityBound.ONE, max(arg_upper))
 
+    elif ir.func_shortname == sn.QualName('std', 'assert_distinct'):
```
msullivan, line 663: Why does this need special handling?

edb/pgsql/compiler/relgen.py
```diff
@@ -2003,6 +2008,143 @@ def process_set_as_existence_assertion(
...
+    #        (CASE WHEN
+    #            <target_set>
+    #            IS DISTINCT FROM
+    #            lag(<target_set>) OVER (ORDER BY <target_set>)
+    #        THEN <target_set>
+    #        ELSE edgedb.raise(ConstraintViolationError))
+    #    FROM
+    #        (SELECT <target_set>, row_number() OVER () AS i) AS q
+    #    ORDER BY
+    #        q.i
```
msullivan, line 2049: ordering by the check_expr too, right?

msullivan (approved):
This looks pretty solid. I think that fixing the xfails should be pretty simple

edb/pgsql/compiler/relgen.py
```diff
@@ -2003,6 +2012,151 @@ def process_set_as_existence_assertion(
...
+            ),
+        ])
+
+        pathctx.put_path_id_map(newctx.rel, ir_set.path_id, ir_arg_set.path_id)
+
+    aspects = ('value', 'source')
+
+    func_rvar = relctx.new_rel_rvar(ir_set, newctx.rel, ctx=ctx)
+    relctx.include_rvar(stmt, func_rvar, ir_set.path_id,
+                        pull_namespace=False, aspects=aspects, ctx=ctx)
```
msullivan, line 2155: Dropping the `pull_namespace=False` fixes the 4 ISE xfails

msullivan:
Oh, and you need to bump `EDGEDB_CATALOG_VERSION`

msullivan:
Aha, and the problem with linkprops is that `compile_call_arg` always wraps call arguments in a `SelectStmt` for fencing and implicit limit insertion purposes. Compiling operator arguments works a little different, so it dodges that.

tests/test_edgeql_update.py
elprans: This is a good point.  I think we can disable multiplicity checks for `-=` and `+=`

edb/edgeql/compiler/inference/cardinality.py
elprans: `std::assert_distinct` preserves the cardinality of its argument, it's completely transparent.

edb/pgsql/compiler/relgen.py
elprans: Right.

edb/edgeql/qltypes.py
elprans: Actually, I think I'll make the structured multiplicity private to inference.  I originally made it this way to minimize the `multiplicity.py` diff, but it had grown very large anyway.

tests/test_edgeql_update.py
1st1: +1

edb/pgsql/compiler/relgen.py
elprans: Oh, good.

edb/edgeql/compiler/inference/utils.py
```diff
@@ -0,0 +1,63 @@
+#
+# This source file is part of the EdgeDB open source project.
+#
+# Copyright 2008-present MagicStack Inc. and the EdgeDB authors.
```
1st1, line 4: Definitely not 2008 ;)

edb/pgsql/compiler/relgen.py
```diff
@@ -2003,6 +2008,143 @@ def process_set_as_existence_assertion(
...
+    # Generate a distinct set assertion as the following SQL:
+    #
+    #    SELECT
+    #        <target_set>,
+    #        (CASE WHEN
+    #            <target_set>
+    #            IS DISTINCT FROM
+    #            lag(<target_set>) OVER (ORDER BY <target_set>)
+    #        THEN <target_set>
+    #        ELSE edgedb.raise(ConstraintViolationError)) AS check_expr
```
1st1, line 2045: Should we add a new class of errors -- RuntimeAssertionError?

edb/edgeql/compiler/inference/multiplicity.py
```diff
@@ -189,21 +193,33 @@ def _infer_shape(
...
+                and irtyputils.is_object(ptrref.out_target)
+            ):
+                ctx.env.schema, ptrcls = irtyputils.ptrcls_from_ptrref(
+                    ptrref, schema=ctx.env.schema)
+                assert isinstance(ptrcls, s_pointers.Pointer)
+                desc = ptrcls.get_verbosename(ctx.env.schema)
+                if not is_mutation:
+                    desc = f"computed {desc}"
                 raise errors.QueryError(
                     f'possibly not a strict set returned by an '
```
1st1, line 216: strict -> distinct?

edb/pgsql/compiler/relgen.py
elprans: Why would that be better than a `ConstraintViolationError` or a `CardinalityViolationError`?

edb/edgeql/compiler/inference/multiplicity.py
elprans: Sure.

edb/edgeql/compiler/inference/utils.py
elprans: Fixed.

PR #2962, opened 2021-09-23
Use an explicit extra dep group to install build deps in CI

elprans:
Changes in #2956 broke package builds, mostly because of missing
parentheses in `install_requires`, but I also think putting build deps
into an extra group would be a cleaner workaround.


fantix (approved):
Oh yes sure, this makes a lot more sense. Thanks!

PR #2964, opened 2021-09-23
Make forward_rptr work for assert_distinct

msullivan:
Update some of the related for tests, also.

### DIFF
--- a/tests/test_edgeql_for.py
+++ b/tests/test_edgeql_for.py
@@ -251,10 +251,6 @@ class TestEdgeQLFor(tb.QueryTestCase):
             }
         )
 
-    @test.xfail("""
-        assert_distinct causes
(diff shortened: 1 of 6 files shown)


tests/test_edgeql_for.py
```diff
@@ -346,7 +342,7 @@ async def test_edgeql_for_in_computable_02b(self):
         await self.assert_query_result(
             r'''
                 SELECT User {
-                    select_deck := (
+                    select_deck := ((
```
elprans, line 345: This hunk seems to be unrelated.

msullivan: Yeah, though it's related in the sense that it's a follow-up to your multiplicity PR involving FOR

elprans (approved):

PR #2966, opened 2021-09-23
Fix missing pg_config_path in nightly

fantix:


elprans (approved):

elprans (commented):
> To use openssl from Homebrew on macOS, run:

Perhaps we should automate this in `setup.py` instead of forcing the developer to jump through hoops?

fantix:
Yeah sure, let me try detecting homebrew and openssl installation, or overrides.

elprans:
I would also try `pkg-config --libs --cflags openssl` to support those who use MacPorts.

fantix:
Detection on macOS will be added in a separate PR.

PR #2968, opened 2021-09-23
Always use bundled Postgres client binaries

elprans:
The internal pgcluster code is currently set up to allow using arbitrary
Postgres installations for client binaries (`pg_config`, `psql`, etc),
which isn't particularly useful because we always ship Postgres binaries
as part of EdgeDB even when a remote cluster is used.


fantix (approved):
Very clean fix! 👍

PR #2969, opened 2021-09-24
Update the docs.

vpetrovykh:
Move things from datamodel to standard library. Update some of the
descriptive text.


docs/datamodel/functions.rst
```diff
@@ -18,17 +18,39 @@ of :eql:type:`int64`:
     db> SELECT len({'hello', 'world'});
     {5, 5}
 
+This behavior is known as an *element-wise* operation. Many built-in
+and user-defined functions operate on elements. In case of multiple
+arguments, a cross-product of all the input sets is computed to
+determine all the input combinations that the function needs to be
+applied to. After that, the function is applied to each element of the
+cross-product in turn to generate the output set. This implies that if
```
1st1, line 26: "in turn" -> "in order"? this sentence is a bit hard to read.

docs/stdlib/datetime.rst
```diff
@@ -5,7 +5,263 @@
...
+
+    See functions :eql:func:`cal::to_relative_duration`, and :eql:func:`to_str`
+    and date/time :eql:op:`operators <DTMINUS>` for more ways of working with
+    :eql:type:`cal::relative_duration`.
+
+
+.. _ref_std_datetime_ops:
+
+=======================
+Functions and Operators
```
1st1, line 263: Hm, I'm not sure it's a good idea to push the func/ops table so far down. How about we move the table to the very top and make the first few rows to link to the actual types? The types description would be right below the table, which essentially is a TOC of the page.

docs/stdlib/json.rst
```diff
@@ -4,7 +4,120 @@
...
+  homogeneous, and must not contain *null*
+
+A *regular* :eql:type:`tuple` is converted into a JSON *array* when cast
+into :eql:type:`json`, whereas a *named* :eql:type:`tuple` is converted
+into a JSON *object*. These casts are not reversible, i.e. it is not
+possible to cast a JSON value directly into a :eql:type:`tuple`.
+
+
+=======================
+Functions and Operators
```
1st1, line 119: Ditto, the toc table should stay on top of the page.

docs/stdlib/numerics.rst
```diff
@@ -4,7 +4,224 @@
...
+        }
+
+    The sequence is bound to the scalar type, not to the property, so
+    if multiple properties use the same ``sequence`` type they will
+    share the same counter. For each distinct counter, a separate
+    scalar type that is extending ``sequence`` should be used.
+
+
+====================================================
+Arithmetic Operators and Numeric Converter Functions
```
1st1, line 223: ditto

1st1 (changes requested):
Overall this looks awesome, but I think we need to move all type descriptions under the TOC table.

vpetrovykh:
I've moved the func/ops tables up to be a sort of TOC for all the pages in the standard library.

I've also added `BaseObject` and `Object` to stdlib docs.

docs/stdlib/numerics.rst
vpetrovykh: ok

docs/stdlib/json.rst
vpetrovykh: ok

docs/stdlib/datetime.rst
vpetrovykh: ok

docs/datamodel/functions.rst
vpetrovykh: This is hard to word right. The "in turn" here is referring to "the function is applied to each element". The order doesn't matter and it's just a way of saying "apply element-wise".

If I change it to "in order" it will look like it refers to "in order to generate the output set".

I think both of these are about the same level of awkward wording, so I've tried splitting the sentence instead.

1st1 (approved):

PR #2970, opened 2021-09-24
Fix test_backend_connect suite on non-devmode servers

elprans:
The `backend_connect` test suite is currently written to assume a dev
server, which has defaults for all arguments, whereas a prod server
requires explicit data dir/DSN and TLS configuration.


fantix:
Ah that's why it failed. Thanks for the fix!

PR #2974, opened 2021-09-27
Fix a number of issues with various "pass-through" functions/operators

msullivan:
Unary operations like assert_single, mostly need to have pull_namespace
turned on, so that packed rvars get pulled up properly.

Binary SET OF things like UNION and if/else, though, need
pull_namespace=False, to prevent a filtered path in one expression
side from clobbering the same path in another. This packed rvar issue
mostly doesn't apply here, because we currently don't support unioning
computed stuff.

This all feels pretty fragile, though.

### DIFF
--- a/edb/pgsql/compiler/pathctx.py
+++ b/edb/pgsql/compiler/pathctx.py
@@ -780,6 +780,12 @@ def get_packed_path_var(
             val=pgast.NullConstant())
 
         assert query.larg and query.rarg
+        assert (
+            isinstance(query.rarg, pgast.SelectStmt)
+            and isinstance(query.rarg.from_clause[0], pgast.RelRangeVar)
+            and isinstance(
+                query.rarg.from_clause[0].relation, pgast.NullRelation)
+        )
         query.rarg.target_list.append(restarget)
 
         nullref = pgast.ColumnRef(name=[alias], nullable=True)
--- a/edb/pgsql/compiler/pathctx.py
+++ b/edb/pgsql/compiler/pathctx.py
@@ -768,6 +768,25 @@ def get_packed_path_var(
 
     query = rvar.query
     assert isinstance(query, pgast.Query)
+
+    top_query = query
+    # Handle the optional case in a hacky way, since it is the only
+    # one that should come up currently. Eventually this should be
+    # handled properly.
+    if astutils.is_set_op_query(query):
+        alias = env.aliases.get('null')
+        restarget = pgast.ResTarget(
+            name=alias,
+            val=pgast.NullConstant())
+
+        assert query.larg and query.rarg
+        query.rarg.target_list.append(restarget)
+
+        nullref = pgast.ColumnRef(name=[alias], nullable=True)
+        _put_path_output_var(query.rarg, path_id, aspect, nullref, env=env)
+
+        query = query.larg
+
     rel_rvar = get_path_rvar(
         query, path_id, flavor='packed', aspect=aspect, env=env)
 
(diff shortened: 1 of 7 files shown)


edb/pgsql/compiler/pathctx.py
```diff
@@ -768,6 +768,25 @@ def get_packed_path_var(
 
     query = rvar.query
     assert isinstance(query, pgast.Query)
+
+    top_query = query
+    # Handle the optional case in a hacky way, since it is the only
+    # one that should come up currently. Eventually this should be
+    # handled properly.
+    if astutils.is_set_op_query(query):
```
elprans, line 776: Is this for the `emptyrel` side of `optional_wrapping`, if so, it probably makes sense to to check for a `NullRelation` explicitly to be safer here.

edb/pgsql/compiler/relctx.py
```diff
@@ -183,7 +182,7 @@ def include_rvar(
         overwrite_path_rvar: bool=False,
         pull_namespace: bool=True,
         flavor: str='normal',
-        aspects: Optional[Tuple[str, ...]]=None,
+        aspects: Optional[Collection[str]]=None,
```
elprans, line 185: Alas, `Collection[str]` includes `str` itself, and I'm always uneasy when there's a signature like this.  Perhaps `Tuple[str, ...] | AbstractSet[str] | None`?

edb/pgsql/compiler/relgen.py
```diff
@@ -2401,6 +2444,17 @@ def _process_set_func_with_ordinality(
             pathctx.put_path_value_var(
                 ctx.rel, element.path_id, element.val, env=ctx.env)
 
+    if (output.in_serialization_ctx(ctx) and ir_set.shape
+            and not ctx.env.ignore_object_shapes):
+        ensure_source_rvar(ir_set, ctx.rel, ctx=ctx)
+        exprcomp._compile_shape(ir_set, ir_set.shape, ctx=ctx)
+
+    var = pathctx.maybe_get_path_var(
+        ctx.rel, ir_set.path_id, aspect='serialized', env=ctx.env)
```
elprans, line 2453: What's going on here?

tests/test_edgeql_select.py
```diff
@@ -6274,3 +6286,105 @@ async def test_edgeql_select_reverse_overload_03(self):
                     SELECT User.<whatever
                 ''',
             )
+
+    async def test_edgeql_function_source_01a(self):
+        # TODO: I think we might want to eliminate this sort of shape
```
elprans, line 6291: Yes, except, I think it would be the array literal that should erase types.

elprans (approved):
LGTM!  A few nits here and there.

tests/test_edgeql_select.py
msullivan: That would be a bit more of a breaking change, since right now shapes will appear through array literals. That might be fine, though?

edb/pgsql/compiler/relgen.py
msullivan: Adding a comment:
```
    # If there is a shape specified on the argument to enumerate, we need
    # to compile it here manually, since we are skipping the normal
    # code path for it.
```

tests/test_edgeql_select.py
msullivan: And array_agg also would need the treatment, I think

msullivan: Oh, right now we do some pretty silly stuff: `SELECT [Card {name}, Card {element}]` in json mode will produce output like `[{"name": "Imp"}, {"element": "Fire"}],`

In native mode is just breaks

Issue #2975, opened 2021-09-28
ISE when `??` in used singleton schema expressions (constraints, indexes)


elprans:
Using `??` in a constraint expression currently results in an ISE.  See https://github.com/edgedb/edgedb/discussions/2907#discussioncomment-1393308

PR #2976, opened 2021-09-28
Fix handling of `??` in schema singleton expressions

elprans:
`std::??` was overlooked in the singleton schema expression code path.

Fixes: #2975


msullivan (approved):
LGTM

PR #2983, opened 2021-09-29
backend pool: fix hanging concurrent acquire due to discard

fantix:
When a frontend connection is broken, the server may cancel the pending
backend query and discard the pgcon. When at the same time the backend
pool is being acquired, we should create a new pgcon after discarding.

This fixes the unstable test_proto_connection_lost_cancel_query, and
also refines the warning of releasing an unhealthy pgcon.


elprans (approved):
Good fix!

PR #2987, opened 2021-09-30
Send capabilities header in CommandDataDescription message

elprans:
The `CommandDataDescription` message should be equivalent to
`PrepareComplete` message and should send the re-parsed query
capabilities to the client.

### DIFF
--- a/edb/server/protocol/binary.pyx
+++ b/edb/server/protocol/binary.pyx
@@ -1383,7 +1383,13 @@ cdef class EdgeConnection:
             WriteBuffer msg
 
         msg = WriteBuffer.new_message(b'T')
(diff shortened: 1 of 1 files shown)


1st1 (commented):
Is this safe to land? If we're not releasing updated edgedb-js/edgedb-python they wouldn't parse this well.

elprans:
> they wouldn't parse this well.

The bindings are supposed to ignore unknown headers.  At least edgedb-python does.

fmoor (approved):
LGTM 

Maybe `CAPABILITIES` should be added to CommandDataDescription's list of Known Headers in the docs?

elprans:
I'll follow up with a doc update in a separate PR.

PR #2997, opened 2021-10-01
Add a nightly CI to cover (old) Postgres versions

fantix:
As suggested by @elprans , we could possibly use the [composite run steps](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action) in the future to replace the jinja-generated duplicates.


elprans (approved):
Excellent!

PR #3048, opened 2021-10-19
Also run on Postgres 14 but with xfail

fantix:


1st1:
> @1st1 Looks like the following test is failing constantly under pg14:

cc @elprans 

Yeah, this is a bug in pg14. Elvis, should we xfail this specifically under pg14 and merge this?

elprans:
Yes, let's xfail the specific test.  I think an env var with Postgres version would be the simplest way to do it: `EDGEDB_TEST_POSTGRES_VERSION`

1st1 (approved):

Issue #2314, opened 2021-03-04
`create-migration` prompts loop if answering 'n' when altering a constraint


jaclarke:
EdgeDB Version: EdgeDB 1.0-beta.1+g2eee1afe4.d20210301

When altering a constraint then running `create-migration`, the prompts loop if you answer 'no' to the drop/create prompts.

Migrating schema from:
```
module default {
  type User {
    required property username -> str {
      constraint exclusive;
      constraint regexp(r'\w+');
    }
  }
};
```
to:
```
module default {
  type User {
    required property username -> str {
      constraint exclusive;
      constraint regexp(r'[a-z0-9_]+');
    }
  }
};
```
and running `edgedb create-migration`:
```
did you create constraint 'std::regexp' of property 'username'? [y,n,l,c,b,s,q,?]
n
did you create constraint 'std::regexp' of property 'username'? [y,n,l,c,b,s,q,?]
n
did you create constraint 'std::regexp' of property 'username'? [y,n,l,c,b,s,q,?]
y
did you drop constraint 'std::regexp' of property 'username'? [y,n,l,c,b,s,q,?]
n
did you drop constraint 'std::regexp' of property 'username'? [y,n,l,c,b,s,q,?]
n
did you drop constraint 'std::regexp' of property 'username'? [y,n,l,c,b,s,q,?]
y
Created ./dbschema/migrations/00002.edgeql, id: m1fadkben4m45k2zh4juemhccnr25awzrdg4s3clpgalsl7lg5gi7q
```

msullivan:
From a quick glance, it looks like this happens for basically anything other than top-level object changes, because the logic for extracting a prompt and the logic for figuring out what was rejecting don't match up.

PR #3080, opened 2021-10-23
Fix many migration proposal rejection issues

msullivan:
One major issue was that the logic for determining what the proposed
change that was being rejected did not match the logic for computing
the proposed change. This is fixed by pulling the info about the
rejected change directly from last_proposed, which also has the
advantage of eliminating another schema diff.

The downside of this is that reject no longer rejects anything unless
DESCRIBE CURRENT MIGRATION AS JSON has been run in the current
transaction. This works fine with the CLI, though. If this is a
problem, we can reuse the logic from DESCRIBE AS JSON to reproduce it
when needed.

can_alter and friends also needed to be changed to get types directly
from the relevant objects, since they were getting confused by sclass
being Pointer (which corresponds to the type of the pointers refdict.)

I had to adjust the confidence generation to prevent emitting a 1.0
confidence for some of the (very not 1.0 confidence!) suggestions
in my tests.

I then fiddled around a bit to try to make some of the messages in
those cases more granular.

### DIFF
--- a/edb/schema/delta.py
+++ b/edb/schema/delta.py
@@ -1764,7 +1764,7 @@ class ObjectCommand(Command, Generic[so.Object_T]):
         self,
         *,
         parent_op: Optional[Command] = None,
-    ) -> Tuple[str, str]:
+    ) -> Tuple[CommandKey, str]:
         """Return a human-friendly prompt describing this operation."""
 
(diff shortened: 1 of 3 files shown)


msullivan:
This will also fix #2314

PR #3103, opened 2021-10-27
Add timeout configs

1st1:


tailhook (commented):
Few nits while skimming over PR

edb/ir/statypes.py
```diff
@@ -0,0 +1,285 @@
...
+                    (?P<microseconds>\d{0,3})
+                    (?P<submicro>\d*)
+                )?
+            )?
+        )?
+        \s*
+        $
+    ''', re.X)
+
+    _pg_parser = re.compile(r'''
```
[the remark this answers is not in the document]
1st1: > And also needs (?: instead of using capturing for groups that don't need to be captures.

It does not, I'm only using named groups.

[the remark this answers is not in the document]
1st1: > This regex looks too complex and too error-prone.

That's why I have tests... I mean we have a lot of regexps in our code base, this isn't the most complex one.

edb/pgsql/codegen.py
```diff
@@ -740,7 +740,7 @@ def visit_CoalesceExpr(self, node):
         self.write(')')
 
     def visit_AlterSystem(self, node):
-        self.write('ALTER INSTANCE ')
+        self.write('ALTER SYSTEM ')
```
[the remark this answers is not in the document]
1st1: This is SQL code generator.

1st1:
@elprans @fantix this is ready for review.

@fantix please pay special attention to https://github.com/edgedb/edgedb/pull/3103/commits/9be2501e0842571724427b9283aecc58e2f30a5f (and also https://github.com/edgedb/edgedb/pull/3103/commits/a26cc505258b6cd1f73c3310d39cad5df4b92cbf to see how "system_config" is encoded in `binary.pyx`).

@fantix with this merged we need to:

* update our bindings to 0.13 protocol. The only difference: the protocol shouldn't assume that the 'S' Settings message contains UTF8 data -- it can contain arbitrary data. Please coordinate with @jaclarke and @fmoor 
* update our REPL to read the sent `system_config` setting, parse it and extract the value of `session_idle_timeout` (if the server speaks 0.13 protocol). Then use that value to periodically send 'Sync' message to the server to avoid REPL being disconnected for idling.
* update our REPL to set `session_idle_transaction_timeout` setting to 5 minutes (the default is 10 seconds and is meant for machines, not humans). Perhaps we should make this configurable in REPL.

fmoor:
> The only difference: the protocol shouldn't assume that the 'S' Settings message contains UTF8 data -- it can contain arbitrary data.

Should the documentation for each known setting include its data type?

1st1:
> Should the documentation for each known setting include its data type?

Yes, it should. I've just pushed a commit documenting that.

1st1:
@colinhacks I'm not sure where to document the new `cfg::memory` type. Perhaps we should document the entire `cfg` module in the Standard Library section.

fmoor:
Should there be an entry for `cfg::memory` in `docs/internals/protocol/dataformats.rst`?

edb/server/server.py
```diff
@@ -535,6 +539,34 @@ async def reload_sys_config(self):
...
+        setting = self._config_settings[setting_name]
+        if setting.report:
+            self.create_task(
+                self.load_reported_config(), interruptable=True)
+
+    def get_report_config_data(self) -> bytes:
+        return self._report_config_data
+
+    async def load_reported_config(self):
+        syscon = await self._acquire_sys_pgcon()
```
elprans, line 552: This seems to only reload the config.  I can't seem to find where it sends the updated config to the clients.

edb/pgsql/metaschema.py
```diff
@@ -236,6 +258,122 @@ def __init__(self) -> None:
...
+                ELSE
+                    edgedb.raise(
+                        NULL::edgedb.memory_t,
+                        msg => (
+                            'unable to parse memory size "' || "val" || '"'
+                        )
+                    )
+            END
+        FROM LATERAL (
+            SELECT regexp_match(
```
elprans, line 309: IMO, the error in the `CASE` block is clearer, so I'd change the unit parsing here to `\w+`

edb/server/server.py
1st1: Yes, we'll need to fix this in RC3.

edb/pgsql/metaschema.py
1st1: Fixed in https://github.com/edgedb/edgedb/pull/3103/commits/97fc60a15629030bda11822898e962b41afaec99

edb/server/server.py
```diff
@@ -527,6 +539,34 @@ async def reload_sys_config(self):
...
+            data = await syscon.parse_execute_extract_single_data_frame(
+                self.get_sys_query('report_configs'),
+                b'__report_configs',
+                dbver=0, use_prep_stmt=True, args=(),
+            )
+            self._report_config_data = (
+                struct.pack('!L', len(self._report_config_typedesc)) +
+                self._report_config_typedesc +
+                data
+            )
```
fantix, line 563: Even though it's very unlikely, but should we handle the case when `data` is `None`? Or perhaps the underlying query should raise an error instead of returning `None`?

fantix (commented):
* Looks like `docs/edgeql/admin/configure.rst` is not updated yet.
* Looks cool to me!

edb/server/pgcon/errors.py
```diff
@@ -41,6 +41,11 @@
 ERROR_CHECK_VIOLATION = '23514'
 ERROR_EXCLUSION_VIOLATION = '23P01'
 
+# Class 25 - Invalid Transaction State
+ERROR_IDLE_IN_TRANSACTION_TIMEOUT = '25P03'
+
+###
```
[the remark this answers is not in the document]
1st1: >  trying to add something here

I think I added 25P03, or was it there?

edb/server/server.py
1st1: > but should we handle the case when data is None?

`self._report_config_typedesc` would either be set to a non-None value or not set at all (prompting an AttributeError). I think we're fine here.

1st1:
@fantix https://github.com/edgedb/edgedb/pull/3134

edb/server/server.py
fantix: Oh sorry I mean `data` returned from `syscon.parse_execute_extract_single_data_frame(...)`

PR #3112, opened 2021-10-28
discussion: handle new tasks on server shutdown

fantix:
For new tasks created after the server is requessted to shutdown and before the loop actually stops, we have 3 options to handle them:

1. Don't even create the `coroutine` object to avoid the `RuntimeWarning` (this PR). The downside is the inconsistent `server.create_task()` API that takes a coroutine factory rather than a coroutine object.
2. Don't call `create_task()` when `server._accept_new_tasks == False`. This adds a lot of duplicate checks before calling `server.create_task()`.
3. Create the tasks anyways, and carefully check every task and modify them to honor the shutdown event. We don't even need the "interruptable" tasks in this case. The risk is, the server may freeze at exit if in the future someone adds new tasks that cannot clean themselves at server exit.
  * We could use `SignalController().wait_for(coro, cancel_on=SIGxxx)` in most tasks, but this also means we'll have a lot of signal-triggered cancellation.
  * At the same time, certain server-exit sanity checks in the tasks are unavoidable. Even if we choose other options, these checks are also necessary for "un-interruptable" tasks to stop, because those tasks may have already started before the server is requested to stop.

### DIFF
--- a/edb/server/protocol/binary.pyx
+++ b/edb/server/protocol/binary.pyx
@@ -1153,7 +1153,7 @@ cdef class EdgeConnection:
     def signal_side_effects(self, side_effects):
         if side_effects & dbview.SideEffects.SchemaChanges:
             self.server.create_task(
-                self.server._signal_sysevent(
+                lambda: self.server._signal_sysevent(
                     'schema-changes',
                     dbname=self.get_dbview().dbname,
                 ),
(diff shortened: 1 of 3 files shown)


edb/server/protocol/binary.pyx
```diff
@@ -1153,30 +1153,30 @@ cdef class EdgeConnection:
     def signal_side_effects(self, side_effects):
         if side_effects & dbview.SideEffects.SchemaChanges:
             self.server.create_task(
```
1st1, line 1155: To make it less ambiguous perhaps we should call this `schedule_task`. But please allow me to think more about this. This can be fixed in rc3 too.

PR #3121, opened 2021-11-03
Make sure `False AND {}` is `{}`

vpetrovykh:
The `std::AND` operator is not `OPTIONAL`, so we cannot safely fall back
to SQL `AND`.  Use the same technique as we do for `OR`.  Technically,
we could formulate this to only use the altered implementation if any of
the operands is optional, but that is left as a future optimization.

### DIFF
--- a/tests/schemas/issues.esdl
+++ b/tests/schemas/issues.esdl
@@ -106,6 +106,15 @@ type Issue extending Named, Owned, Text {
     property tags -> array<str>;
 }
 
+# This is used to test correct behavior of boolean operators: NOT,
+# AND, OR. It targets especially interactions of properties and {}.
+#
+# Issue can be used to test similar interaction for links.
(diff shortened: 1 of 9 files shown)


vpetrovykh:
I've added more tests for single and multi properties interacting with boolean operators.

elprans (approved):

tests/schemas/issues.esdl
```diff
@@ -106,6 +106,15 @@ type Issue extending Named, Owned, Text {
     property tags -> array<str>;
 }
 
+# This is used to test correct behavior of boolean operators: NOT,
+# AND, OR. It targets especially interactions of properties and {}.
+#
+# Issue can be used to test similar interaction for links.
+type BooleanTest extending Named {
```
elprans, line 113: Probably not the best place for this.  Why not `test_expressions`?

PR #3122, opened 2021-11-03
Force delete/recreate when referenced object deleted/recreated

### LINKED ISSUE #3118: Dependent objects (like indexes) not handled when property is deleted/recreated
As an example, see test `test_schema_migrations_equivalence_54`:
We wish to migrate from
```
type User {
      index on (__subject__.name);
};
```
to
```
scalar type Slug extending str;
abstract type Named {
    required property name -> Slug;
    index on (__subject__.name);
};
type User extending Named;
```

For better or worse (... likely worse), the initially inferred migration path is to delete the property `name` and let it get recreated when we inherit it. We produce this migration:
```
ALTER TYPE default::User {
    DROP PROPERTY name;
    EXTENDING default::Named LAST;
};
ALTER TYPE default::User {
    ALTER INDEX ON (__subject__.name) DROP OWNED;
};
```
The problem is that the index becomes invalid when we drop name; if we are dropping `name`, we need to also drop the index on it, and let it get recreated.

msullivan:
If an object that a referring object depends on is deleted and then
recreated, we currently don't have anything we can do except also do
that for the referring object as well.

Fixes #3118.

### DIFF
--- a/edb/schema/ddl.py
+++ b/edb/schema/ddl.py
@@ -271,13 +271,12 @@ def delta_schemas(
     ]
 
     assert not context.renames
-    # We retry performing the diff until we stop finding new
-    # renames. This allows us to be agnostic to the order that we
-    # process schemaclasses.
-    old_count = -1
-    while old_count != len(context.renames):
-        old_count = len(context.renames)
-        context.deletions.clear()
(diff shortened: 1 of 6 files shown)


elprans (approved):
Happy to see many xfails fixed!

PR #3123, opened 2021-11-03
Make "and" and "or" in GraphQL conform to Postgres truthtable.

vpetrovykh:
In EdgeQL all operators handle {} the same way. If the operands are
required, then the result is {} if any of the inputs are {}. This
applies to boolean operators as well, so that `not`, `and`, `or` will
all produce {} as the result if any of the operands are {}. This case
can then be explicitly handled by adding `??` to the operands or the the
result.

In Postgres `False AND NULL` is `False` and `True OR NULL` is `True`.
In the general case this inconsistency can cause hard to debug subtle
issues.

In GraphQL we don't have arbitrary expressions, but rather a very
restricted subset of them appearing as "filter". In that particular
subset of usage it makes sense to use "short-circuiting" definition of
"and" and "or" operators because that best conforms to the expectations
in this use-case.

### DIFF
--- a/edb/graphql/translator.py
+++ b/edb/graphql/translator.py
@@ -1344,20 +1344,20 @@ class GraphQLTranslator:
         return qlast.Array(elements=self.visit(node.values))
 
     def visit_ObjectValueNode(self, node):
-        # this represents some expression to be used in filter
-        result = []
-        for field in node.fields:
-            result.append(self.visit(field))
-
-        return self._join_expressions(result)
+        # This represents some expression to be used in filter. In
+        # case of multiple expressions they are implicitly combined
+        # using AND.
+        return self._visit_list_generalized_bool_op(node.fields, 'AND')
 
     def visit_ObjectFieldNode(self, node):
         fname = node.name.value
 
         # handle boolean ops
         if fname == 'and':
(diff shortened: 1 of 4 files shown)


edb/graphql/translator.py
```diff
@@ -1612,15 +1612,42 @@ def _visit_list_of_inputs(self, inputlist, op):
...
             return None
-        elif len(exprs) == 1:
-            return exprs[0]
+        elif len(nodes) == 1:
+            return self.visit(nodes[0])
+
+        # The short-circuiting value is True for OR and False for AND.
+        short = op == 'OR'
+
+        return qlast.IfElse(
```
elprans, line 1638: I don't think using `IfElse` would be friendly to the query plan, it's very likely Postgres wouldn't be able to use bitmap index scans if a `CASE WHEN` is involved.  I'm afraid we don't have a choice but to introduce the short-circuiting versions of `AND` and `OR`, possibly as functions.

vpetrovykh: I've added `sys::__pg_and` and `sys::__pg_or` functions marked as "internal".

msullivan (approved):
Remember to bump EDGEDB_CATALOG_VERSION, but this seems good to me

PR #3126, opened 2021-11-03
Document Prometheus metrics

colinhacks:


docs/internals/observability.rst
```diff
@@ -0,0 +1,64 @@
...
+
+All EdgeDB instances expose a Prometheus-compatible ``/metrics`` endpoint on
+port ``9090``. The following metrics are made available.
+
+
+Processes
+^^^^^^^^^
+
+``compiler_process_spawns_total``
+  **Counter.** Total number of compiler processes spawned.
```
1st1, line 14: Hm, I think we should auto-generate this, we have all this data right in metrics.py. @vpetrovykh please take a look.

colinhacks: Admirable goal but this took like 5 minutes to put together and it's easy to update. Might be a premature optimization (until we have way more metrics).

1st1: > Generating might be a premature optimization (until we have way more metrics).

We will and this documentation will go out of sync in no time. We need to do this properly. I'm asking @vpetrovykh for help here because he has just done something similar (even more complex actually) for our cast table -- this is easier.

1st1: I'm OK with committing this now just for the purpose of publishing the blog post. But we'll need a follow up to make it proper right after rc2.

1st1 (approved):

docs/internals/observability.rst
1st1: @colinhacks if you merge this now, please create an issue for generating this automatically and assign to @vpetrovykh

PR #3140, opened 2021-11-05
Do not emit `ALTER SYSTEM` when it's not supported by the backend

elprans:
Currently, we unconditionally compile `CONFIGURE INSTANCE` to `ALTER 
SYSTEM`, even if the latter is not supported by the backend, which is 
common for cloud Postgres.  When unsupported, emit a config op JSON record
and let `dbview.apply_system_config_op()` deal with it.

### DIFF
--- a/edb/api/errors.txt
+++ b/edb/api/errors.txt
@@ -145,6 +145,13 @@
 
 ####
 
+0x_09_00_00_00   BackendError
+
+0x_09_00_01_00   UnsupportedBackendFeatureError
+
+
+####
+
 0x_F0_00_00_00   LogMessage
 
 0x_F0_01_00_00   WarningMessage
--- a/edb/pgsql/compiler/config.py
+++ b/edb/pgsql/compiler/config.py
@@ -27,6 +27,7 @@ from edb.schema import objects as s_obj
 from edb.edgeql import qltypes
 
 from edb.pgsql import ast as pgast
+from edb.pgsql import params as pgparams
 
 from . import astutils
 from . import clauses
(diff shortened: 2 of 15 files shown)


edb/pgsql/compiler/config.py
```diff
@@ -47,7 +48,14 @@ def compile_ConfigSet(
...
 
-    if op.scope is qltypes.ConfigScope.INSTANCE and op.backend_setting:
+    if (
+        op.scope is qltypes.ConfigScope.INSTANCE
+        and op.backend_setting
+        and (
+            ctx.env.backend_runtime_params.instance_params.capabilities
+            & pgparams.BackendCapabilities.CONFIGFILE_ACCESS
+        )
+    ):
```
1st1, line 58: So this means that ALTER INSTANCE is a no-op on systems without CONFIGFILE_ACCESS?

elprans: Currently, effectively yes, but the idea is that we'll add special handling in dbview to apply these changes via an appropriate side channel.  This PR basically makes it so that `CONFIGURE INSTANCE` doesn't crash.

1st1 (approved):

edb/api/errors.txt
```diff
@@ -24,6 +24,7 @@
 ####
 
 0x_02_00_00_00   UnsupportedFeatureError
+0x_02_00_01_00   UnsupportedBackendFeatureError
```
1st1, line 27: I'd consider making it a top-level separate type.

1st1 (approved):

edb/pgsql/compiler/config.py
```diff
@@ -48,6 +49,14 @@ def compile_ConfigSet(
     result: pgast.BaseExpr
 
     if op.scope is qltypes.ConfigScope.INSTANCE and op.backend_setting:
+        if not (
+            ctx.env.backend_runtime_params.instance_params.capabilities
+            & pgparams.BackendCapabilities.CONFIGFILE_ACCESS
+        ):
+            raise errors.UnsupportedBackendFeatureError(
+                "configuring backend parameters via CONFIGURE INSTANCE"
+                " is not supported by the current backend"
```
1st1, line 58: Do we have something like "backend name" or "backend type" in our system? E.g. it would be cool to explicitly say "this isn't supported on RDS" or "... on Stolon"

elprans: We don't.  We do feature-based inference of backend capabilities, but, strictly speaking, there isn't a way to get a good "title" for the host kind.

edb/api/errors.txt
elprans: Changed.  I added a new base `BackendError` and inherited `UnsupportedBackendFeatureError` from it.

Issue #1252, opened 2020-02-21
Mechanism for reflecting tuples into GraphQL


vpetrovykh:
Currently we don't support reflecting tuples into GraphQL because GraphQL does not support heterogeneous lists and because reflecting named tuples as an "Object" with fields presents a problem of partial fetching.

I have an idea of how tuples can be generically reflected, but this would make them read-only w.r.t. GraphQL. So it's not the best solution, but it would at least address those usages where read-only approach is acceptable.

The idea is to reflect all tuples as a GraphQL Object. For unnamed tuples the fields would be given some generic numbered name like `field0`, `field1`, etc. This gets around the issue of not supporting heterogeneous lists. However, the objects in GraphQL don't have to be fetched completely (unlike the underlying tuples, which are a single unit), so we need a mechanism for addressing that. The solution I propose is to expose the tuple property as `<prop_name>_tuple` or some similar mangled name. This means that the `<prop_name>_tuple` can be an ad-hoc computable of any type we like in the underlying EdgeQL query, so we can reshape the actual tuple in any way.

Consider this example:
```
type Foo {
    # some really fancy tuple with non-trivial structure
    property data -> 
        tuple<int64, str, array<float64>, tuple<a: int64, b: float64>>
}
```
The GraphQL query would look something like this:
```
query {
  Foo(
    filter: {data_tuple: {field0: {eq: 42}}}
  ) {
    data_tuple {
      field0
      field2
      field3 {
        b
      }
    }
  }
}
```
And the resulting EdgeQL would be something like:
```
SELECT Foo {
    data_tuple := (
        field0 := .data.0,
        field2 := .data.2,
        field3 := (a := .data.3.a)
    )
} FILTER .data.0 = 42;
```

We wouldn't support any operations on the fields inside the tuple (such as filtering), so that we could still maintain the integrity of the tuple data. So each field is optional to fetch, but cannot be altered in any other way when fetching.

@1st1, @elprans what are your thoughts on this?

vpetrovykh:
Actually, I realized we can reflect a tuple field insert by making the input object for it have all its fields and non-nullable, so that they must be specified. So then inserting any tuple structure becomes an all-or-nothing proposition in GraphQL.

elprans:
This seems reasonable to me.  Not a high-priority for now, though.

1st1:
[...]

PR #3145, opened 2021-11-06
Reflect more native types to GraphQL.

vpetrovykh:
Reflect EdgeDB tuples as JSON to GraphQL.

Reflect `array<json>` as List[JSON!] to GraphQL. There's no reason for
this restriction anymore as JSON van now be reflected.

Reflect `array<tuple<...>>` as List[JSON!]` to GraphQL, but make it
read-only because it involves a tuple.

Issue #1252.

### DIFF
--- a/edb/graphql/types.py
+++ b/edb/graphql/types.py
@@ -437,14 +437,8 @@ class GQLCoreSchema:
     ) -> Optional[GraphQLOutputType]:
         target: Optional[GraphQLOutputType] = None
 
-        # only arrays can be validly wrapped, other containers don't
-        # produce a valid graphql type
         if isinstance(edb_target, s_types.Array):
             subtype = edb_target.get_subtypes(self.edb_schema)[0]
-            if str(subtype.get_name(self.edb_schema)) == 'std::json':
-                # cannot expose arrays of JSON
-                return None
-
             el_type = self._convert_edb_type(subtype)
             if el_type is None:
                 # we can't expose an array of unexposable type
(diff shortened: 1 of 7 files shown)


edb/graphql/types.py
```diff
@@ -437,14 +437,8 @@ def _convert_edb_type(
     ) -> Optional[GraphQLOutputType]:
         target: Optional[GraphQLOutputType] = None
 
-        # only arrays can be validly wrapped, other containers don't
-        # produce a valid graphql type
         if isinstance(edb_target, s_types.Array):
             subtype = edb_target.get_subtypes(self.edb_schema)[0]
-            if str(subtype.get_name(self.edb_schema)) == 'std::json':
```
1st1, line 444: Use `issubtype` here

edb/graphql/types.py
```diff
@@ -806,6 +799,14 @@ def get_insert_fields(
                 intype = self._wrap_input_type(ptr, intype)
                 fields[name] = GraphQLInputField(intype)
 
+            elif (
+                isinstance(edb_target, s_types.Array) and
```
elprans, line 803: Use `edb_target.contains_array_of_tuples`

edb/graphql/types.py
```diff
@@ -897,6 +898,14 @@ def get_update_fields(
 
                 fields[name] = GraphQLInputField(intype)
 
+            elif (
+                isinstance(edb_target, s_types.Array) and
+                edb_target.get_subtypes(
+                    self.edb_schema)[0].is_tuple(self.edb_schema)
```
elprans, line 904: Ditto

elprans (approved):

edb/graphql/types.py
vpetrovykh: These are deleted lines, so in the new version we don't perform that check. Subtype is used for the recursion.

edb/graphql/types.py
vpetrovykh: Done

edb/graphql/types.py
vpetrovykh: Done

Issue #3115, opened 2021-11-02
Unexpected `function is defined recursively` error


colinhacks:
I'm having trouble adding the following user-defined function to my schema:

```
function len_strings(words: str) -> int64
    using (len(words));
```

I'm getting this error:

```
$ edgedb migration create                 
error: function 'default::len_strings(words: str)' is defined recursively
    ┌─ ./dbschema/default.esdl:265:3
    │  
265 │ ╭   function len_strings(words: str) -> int64
266 │ │     using (len(words));
    │ ╰───────────────────────^ error

edgedb error: cannot proceed until .esdl files are fixed
```

This error only appears if my function name begins with `len`. I assume this is a broader problem in the loop detection code.

PR #3164, opened 2021-11-10
Fix false positive cycle detections in functions

elprans:
The function body tracer is checking for references to functions and
then uses the name prefix to look for all overloaded function
definitions to include in the dependency list.  Unfortunately, the
prefix checking code simply uses `startswith`, which could grab
completely unrelated functions if they happen to have a name with
a given prefix.

Fixes: #3115

### DIFF
--- a/edb/edgeql/tracer.py
+++ b/edb/edgeql/tracer.py
@@ -310,20 +310,20 @@ class TracerContext:
         self, ref: qlast.ObjectRef
     ) -> Set[sn.QualName]:
         refs = set()
-        prefixes = []
+        prefixes = set()
 
         if ref.module:
             # replace the module alias with the real name
             module = self.modaliases.get(ref.module, ref.module)
-            prefixes.append(f'{module}::{ref.name}')
+            prefixes.add(f'{module}::{ref.name}')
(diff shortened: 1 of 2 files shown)


edb/edgeql/tracer.py
```diff
@@ -310,20 +310,20 @@ def get_ref_name_starstwith(
...
-            prefixes.append(f'{self.module}::{ref.name}')
-            prefixes.append(f'std::{ref.name}')
+            prefixes.add(f'{self.module}::{ref.name}')
+            prefixes.add(f'std::{ref.name}')
 
         for objname in self.objects.keys():
-            for prefix in prefixes:
-                if str(objname).startswith(prefix):
-                    refs.add(objname)
+            short_name = str(objname).split('@@', 1)[0]
```
1st1, line 324: Don't we have a helper for this?

1st1 (approved):

edb/edgeql/tracer.py
elprans: No, this is homegrown mangling in SDL tracer, not the schema mangling.

PR #3165, opened 2021-11-10
Fix more issues with injecting typenames into std:: objects

msullivan:
We need to be able to insert views for injection inside of WITH/FOR
bindings, so we generalize expr_exposed to have an intermediate
exposure level corresponding to bindings.

To make problems like this less obscure, we make them *worse* by
adding an assert that prevents injecting into non-view shapes in
general.

### DIFF
--- a/edb/edgeql/compiler/viewgen.py
+++ b/edb/edgeql/compiler/viewgen.py
@@ -446,7 +446,7 @@ def _compile_qlexpr(
         shape_expr_ctx.partial_path_prefix = source_set
 
         if is_mutation and ptrcls is not None:
-            shape_expr_ctx.expr_exposed = True
+            shape_expr_ctx.expr_exposed = context.Exposure.EXPOSED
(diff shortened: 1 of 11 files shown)


edb/edgeql/compiler/viewgen.py
```diff
@@ -1202,11 +1202,15 @@ def _inline_type_computable(
     stype: s_objtypes.ObjectType,
     compname: str,
     propname: str,
+    is_mutation: bool,
     *,
     shape_ptrs: List[ShapePtr],
     ctx: context.ContextLevel,
 ) -> None:
     assert isinstance(stype, s_objtypes.ObjectType)
+    # Injecting into non-view objects /usually/ works, but it fails if the
```
elprans, line 1211: The "usually works" part is wrong too, because user-defined schema types shouldn't be mutated either (if you add or change anything, it should be a new type id), so this is a good change.

elprans (approved):

PR #3167, opened 2021-11-10
Produce better error on bad arguments to set constructors

1st1:
Currently a query like "SELECT {1, 2, '3'}" produces a hard to
interpret error:

   operator 'UNION' cannot be applied to operands
   of type 'std::int64' and 'std::str'

Fix the error message to a more comprehensible:

   set constructor has arguments of incompatible types
   'std::int64' and 'std::str'

While there also fix the source context of the generated error
to point to the location of the error.

Fixes: https://github.com/edgedb/edgedb/issues/2333

### DIFF
--- a/edb/edgeql/ast.py
+++ b/edb/edgeql/ast.py
@@ -281,6 +281,10 @@ class BinOp(Expr):
     right: Expr
 
 
+class SetConstructorOp(BinOp):
+    op: str = 'UNION'
+
+
 class WindowSpec(Clause, OrderByMixin):
     orderby: typing.List[SortExpr]
     partition: typing.List[Expr]
--- a/edb/edgeql/compiler/expr.py
+++ b/edb/edgeql/compiler/expr.py
@@ -26,6 +26,7 @@ from typing import *
 
 from edb import errors
 
+from edb.common import context as ctx_utils
 from edb.edgeql import qltypes as ft
 
 from edb.ir import ast as irast
@@ -70,6 +71,7 @@ def compile_Path(
     return stmt.maybe_add_view(setgen.compile_path(expr, ctx=ctx), ctx=ctx)
 
 
+@dispatch.compile.register(qlast.SetConstructorOp)
 @dispatch.compile.register(qlast.BinOp)
 def compile_BinOp(
         expr: qlast.BinOp, *, ctx: context.ContextLevel) -> irast.Set:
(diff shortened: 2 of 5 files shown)


1st1:
Before:

Screenshot from 2021-11-10 09-59-29

After:

Screenshot from 2021-11-10 09-59-55

----

I do realize this isn't the most elegant fix. But it's not the first time we re-interpret errors based on the message -- we do that all the time for backend errors. I've added tests to catch regressions, so we should be OK with this.

edb/edgeql/compiler/expr.py
```diff
@@ -70,12 +71,18 @@ def compile_Path(
...
 def compile_BinOp(
         expr: qlast.BinOp, *, ctx: context.ContextLevel) -> irast.Set:
 
     op_node = func.compile_operator(
-        expr, op_name=expr.op, qlargs=[expr.left, expr.right], ctx=ctx)
+        expr,
+        op_name=expr.op,
+        qlargs=[expr.left, expr.right],
+        ctx=ctx,
+        setunion=isinstance(expr, qlast.SetUnionOp)
```
elprans, line 84: Why not put this `isinstance` inside `compile_operator` instead of adding an argument?

edb/edgeql/compiler/expr.py
```diff
@@ -70,12 +71,27 @@ def compile_Path(
...
 
-    op_node = func.compile_operator(
-        expr, op_name=expr.op, qlargs=[expr.left, expr.right], ctx=ctx)
+    try:
+        op_node = func.compile_operator(
+            expr, op_name=expr.op, qlargs=[expr.left, expr.right], ctx=ctx)
+    except errors.QueryError as e:
+        if not isinstance(expr, qlast.SetUnionOp):
+            raise
+        prefix = "operator 'UNION' cannot be applied to operands of type "
```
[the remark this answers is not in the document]
1st1: The latest approach doesn't have this problem, but adds a couple of tests to make sure

edb/edgeql/compiler/expr.py
1st1: Done :man_facepalming:

edb/edgeql/ast.py
```diff
@@ -281,6 +281,10 @@ class BinOp(Expr):
     right: Expr
 
 
+class SetUnionOp(BinOp):
```
elprans, line 284: I'd name this `SetConstructorOp` for clarity (and add a comment that it's needed for `{...}`)

1st1: Good idea, updated.

elprans (approved):

PR #3168, opened 2021-11-10
Set session_idle_timeout and session_idle_transaction_timeout

1st1:
Now that our REPL reads the session_idle_timeout setting and sends
periodic SYNC messages to avoid disconnects, we're ready to flip
the timeouts back to what was planned: 60 and 10 seconds
correspondingly.


1st1:
@colinhacks We need to update the blog post now to mention that the new timeouts are actually enabled.

elprans (approved):

fantix (approved):

PR #3169, opened 2021-11-10
Some cleanup to recent type computable inlining changes

msullivan:
PR #3165 added an assert that the stype was a view, but exempted
mutations from that check. Exempting mutations is spurious and
unnecessary (though it seemed necessary at one point during
development).

### DIFF
--- a/edb/edgeql/compiler/viewgen.py
+++ b/edb/edgeql/compiler/viewgen.py
@@ -1202,15 +1202,15 @@ def _inline_type_computable(
     stype: s_objtypes.ObjectType,
     compname: str,
     propname: str,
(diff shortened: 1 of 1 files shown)


elprans (approved):

1st1:
Can you figure out some sort of regression test for this?

msullivan:
> Can you figure out some sort of regression test for this?

I'm not sure I follow. There's no bug being fixed here, the change is just making an assert stronger.

1st1:
> I'm not sure I follow. There's no bug being fixed here, the change is just making an assert stronger.

Never mind then, I looked at the changes very quickly and notices that there are no tests.

PR #3172, opened 2021-11-11
Drop materialization special case in compile_query_subject view insertion

msullivan:
It is hacky, and not really needed anymore. Removing it fixed a test.

Removing it also caused test_edgeql_insert_enumerate_01 to fail when
doing type injection since it exposed an issue that allowed id and
`__tname__` to be injected twice in some queries that inserted twice
to the same object. I decided that the real root cause of that issue
was that schemactx copied view_shapes to child views without updating
the pointers to point to the child, so I changed that.

There wound up being a lot of churn on that side, as I had to commit
some violence to get mypy to not lose track of things because of type
intersections.

### DIFF
--- a/edb/edgeql/compiler/schemactx.py
+++ b/edb/edgeql/compiler/schemactx.py
@@ -28,13 +28,13 @@ from edb import errors
 
 from edb.common import parsing
 
-from edb.schema import abc as s_abc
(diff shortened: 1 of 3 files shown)


elprans (approved):
:+1:

PR #3188, opened 2021-11-16
Stop filtering out IPv6 interfaces for "localhost" if IPv6 is supported

1st1:


fantix (approved):
Looks good to me!

Should we also fix `_fix_wildcard_host()` to include `::1`?

1st1:
@elprans @fantix I ended up fixing many things. Please review again.

edb/server/server.py
```diff
@@ -1703,17 +1704,41 @@ async def _resolve_localhost() -> List[str]:
...
+
+    wc_hosts = []
+
+    if ipv4_wc in ip_hosts:
+        wc_hosts.append('0.0.0.0')
+
+    if ipv6_wc in ip_hosts:
+        wc_hosts.append('::')
+
+    if wc_hosts:
```
elprans, line 1733: I think you have to manage the IPv4 and IPv6 sets separately here.  Otherwise you'll discard IPv4 hosts if I specified (`127.0.0.1, ::`).

1st1: Right, fixed

elprans (approved):

PR #3191, opened 2021-11-16
Fix generated SET TYPE in migration when a create is split into an alter

### LINKED ISSUE #3183: Changing type of overloaded link leads to weird errors during migration creation
- EdgeDB Version: 1-rc2
- OS Version: xubuntu 20.04.3

Steps to Reproduce:

Running `edgedb migration create --non-interactive --allow-unsafe` using the schema below returns the following error:

> edgedb error: SchemaError: link 'entity' of object type 'default::DogVersion' cannot be cast automatically from object type 'default::Versioned' to object type 'default::Dog'
  Hint: You might need to specify a conversion expression in a USING clause
  Server traceback: [...]

The interactive version is even worse, it gets stuck in an endless loop attempting to roll back a step.

Schema:
```
module default {
	abstract type Version {
		required link entity -> Versioned {
			readonly := true;
		}
	}

	abstract type Barks {
	}

	abstract type Versioned {
	}

	type DogVersion extending Version, Barks {
		overloaded required link entity -> Dog;
	}

	type Dog extending Versioned {
		multi link versions := (SELECT Dog.<entity[IS DogVersion]);
	}

	type CatVersion extending Version {
		overloaded required link entity -> Cat;
	}

	type Cat extending Versioned {
	}
}
```

This example is very fragile, even apparently unrelated changes like removing `Barks` from `Dog` or renaming types can lead to it successfully creating a migration.

The migration it produces when I remove `Barks` from `Dog`:

```
CREATE MIGRATION m1qqz546xpyqxoha4ipcw3upqbnzzudphdkihdojuu6bbui4r53gwq
    ONTO initial
{
  CREATE ABSTRACT TYPE default::Barks;
  CREATE ABSTRACT TYPE default::Versioned;
  CREATE TYPE default::Cat EXTENDING default::Versioned;
  CREATE ABSTRACT TYPE default::Version {
      CREATE REQUIRED LINK entity -> default::Versioned {
          SET readonly := true;
      };
  };
  CREATE TYPE default::CatVersion EXTENDING default::Version {
      ALTER LINK entity {
          SET OWNED;
          SET REQUIRED;
          SET TYPE default::Cat;
      };
  };
  CREATE TYPE default::Dog EXTENDING default::Versioned;
  CREATE TYPE default::DogVersion EXTENDING default::Version {
      ALTER LINK entity {
          SET OWNED;
          SET REQUIRED;
          SET TYPE default::Dog;
      };
  };
  ALTER TYPE default::Dog {
      CREATE MULTI LINK versions := (SELECT
          default::Dog.<entity[IS default::DogVersion]
      );
  };
};
```

I also had a variant where interactive mode asked for a conversion expression for such a link.

msullivan:
When an ALTER LINK that changes type gets split off from a CREATE TYPE
into an ALTER TYPE, we may need to include a USING statement. The body
of the USING statement doesn't actually matter---it's only needed because
the DDL side can't figure out that the cast expr isn't needed.

Fixes #3183.

### DIFF
--- a/edb/schema/links.py
+++ b/edb/schema/links.py
@@ -362,12 +362,38 @@ class CreateLink(
                         assert isinstance(t, (so.Object, so.ObjectShell))
                         node.target = utils.typeref_to_ast(schema, t)
             else:
+                old_type = pointers.merge_target(
+                    self.scls,
+                    list(self.scls.get_bases(schema).objects(schema)),
+                    'target',
+                    ignore_local=True,
+                    schema=schema,
+                )
                 assert isinstance(op.new_value, (so.Object, so.ObjectShell))
+                new_type = (
+                    op.new_value.resolve(schema)
+                    if isinstance(op.new_value, so.ObjectShell)
+                    else op.new_value)
+
+                new_type_ast = utils.typeref_to_ast(schema, op.new_value)
+                cast_expr = None
+                # If the type isn't assignment castable, generate a
+                # USING with a nonsense cast. It shouldn't matter,
+                # since there should be no data to cast, but the DDL side
+                # of things doesn't know that since the command is split up.
+                if old_type and not old_type.assignment_castable_to(
+                        new_type, schema):
+                    cast_expr = qlast.TypeCast(
+                        type=new_type_ast,
+                        expr=qlast.Set(elements=[]),
+                    )
                 node.commands.append(
                     qlast.SetPointerType(
-                        value=utils.typeref_to_ast(schema, op.new_value),
+                        value=new_type_ast,
+                        cast_expr=cast_expr,
                     )
                 )
+
         elif op.property == 'on_target_delete':
             node.commands.append(qlast.OnTargetDelete(cascade=op.new_value))
         else:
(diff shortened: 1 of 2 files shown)


edb/schema/links.py
```diff
@@ -362,12 +362,38 @@ def _apply_field_ast(
...
+
+                new_type_ast = utils.typeref_to_ast(schema, op.new_value)
+                cast_expr = None
+                # If the type isn't assignment castable, generate a
+                # USING with a nonsense cast. It shouldn't matter,
+                # since there should be no data to cast, but the DDL side
+                # of things doesn't know that since the command is split up.
+                if old_type and not old_type.assignment_castable_to(
+                        new_type, schema):
+                    cast_expr = qlast.TypeCast(
```
elprans, line 386: Hmm. I'm slightly surprised this works, since type casts from a known object type to another object type are ought to be illegal:

```
edgedb> CREATE TYPE Foo;
OK: CREATE TYPE
edgedb> CREATE TYPE Bar;
OK: CREATE TYPE
edgedb> SELECT <Bar>Foo;
error: QueryError: cannot cast object type 'default::Foo' to 'default::Bar', use `...[IS default::Bar]` instead
  ┌─ query:1:13
  │
1 │ SELECT <Bar>Foo;
  │             ^^^ error
```

elprans: Empty set. OK

elprans (approved):

PR #3197, opened 2021-11-18
Add support for Azure to `--backend-dsn`

elprans:
Azure requires a custom-superuser quirk similar to RDS and GCP.


fmoor (approved):

edb/server/pgcluster.py
```diff
@@ -834,6 +834,7 @@ async def _get_cluster_type(
         managed_clouds = {
             'rds_superuser': RemoteCluster,    # Amazon RDS
             'cloudsqlsuperuser': RemoteCluster,    # GCP Cloud SQL
+            'azure_pg_admin': RemoteCluster,    # Azure Postgres
```
1st1, line 837: Can you handle a possible KeyError (raise a nice error) when you're indexing this dict a few lines below?

1st1 (approved):

Issue #3198, opened 2021-11-18
ALTER ... DROP EXTENDING succeeds even if the class isn't actually a parent


msullivan:
or in fact, even if it doesn't exist.
`ALTER TYPE Foo DROP EXTENDING NonExists` works fine.

PR #3199, opened 2021-11-18
Don't have {} act as a wildcard in assert_query_result

### LINKED ISSUE #3196: Don't have `{}` act as a wildcard in `assert_query_result`
For assert_query_result, the empty dict `{}` currently acts as a "wild card", that will match anything.

This is because it checks each element in the dict, and so if the dict is empty, it doesn't look at the returned data at all.

I think we should make sure that `{}` only matches dictionaries, and should add in a separate wildcard facility.
We should maybe also have it check that the result dictionary doesn't contain any *extra* keys, except for `id` and `__tname__`. (We don't really have a good way to distinguish between named tuples and objects in this interface, so I don't think we can do the *best* thing and *insist* that those are inserted when appropriate).

msullivan:
For assert_query_result, the empty dict `{}` acts as a "wild card",
that will match anything (because it checks each element in the dict,
and so if the dict is empty, it doesn't look at the returned data at all).

In addition to tests intentionally using it as a wildcard, there are
lots of tests that wanted it to mean "null or empty set".

For the former, I put in the actual value or put in the type,
which acts as a wildcard but in a slight more typed way.

The latter I fixed.

Fixes #3196.

### DIFF
--- a/tests/test_edgeql_ddl.py
+++ b/tests/test_edgeql_ddl.py
@@ -591,7 +591,7 @@ class TestEdgeQLDDL(tb.DDLTestCase):
                 } ORDER BY TestSelfLink2.foo2;
             """,
             [
-                {'bar2': {}, 'foo2': 'Alice'},
+                {'bar2': [], 'foo2': 'Alice'},
                 {'bar2': {'Alice'}, 'foo2': 'Bob'},
                 {'bar2': {'Alice', 'Bob'}, 'foo2': 'Carol'}
             ],
@@ -3519,7 +3519,7 @@ class TestEdgeQLDDL(tb.DDLTestCase):
             r"""
                 SELECT int_func_1();
             """,
-            [{}],
+            [1],
         )
 
     async def test_edgeql_ddl_function_07(self):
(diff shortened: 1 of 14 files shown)


tests/test_edgeql_ddl.py
```diff
@@ -6445,9 +6445,14 @@ async def test_edgeql_ddl_extending_04(self):
...
                         .annotations@value;
             """,
-            {}
+            []
         )
 
+    @test.xfail('''
+        Default value ought to get reset back to non-existent, since it
+        was inherited? (Or actually maybe not, since the prop is owned
+        by then?)
```
msullivan, line 6454: This test is the one questionable one. It wanted to test that when a parent type is dropped, a owned inherited property that was inheriting the default should lose the default.

It wasn't quite testing this, because it was dropping one parent a second time (#3198).

When that is fixed, though, dropping the remaining parent doesn't cause the default to clear.

Maybe that is right, though? The child property is owned. (Though the default is inherited.)

elprans (approved):

tests/test_edgeql_ddl.py
elprans: The answer to such DDL questions is: how would you express an equivalent SDL migration.  Losing a parent in SDL usually means you lose everything you inherited from it, so IMO it's natural that DDL follows the behavior.

Issue #3185, opened 2021-11-15
Investigate splitting and moving FILTER clauses


msullivan:
As a simple example of when we want this,
```
SELECT User
FILTER User.name = 'Alice' AND User.deck.name = 'Imp'
```
generates `WHERE EXISTS ...` as the filter because the filter clause has cardinality many. We could probably get more efficient queries by breaking it up.

It's also worth investigating whether we get into trouble with expensive shape construction inside filtered selects? I think elvis said @jaclarke ran into an introspection query that caused trouble?

PR #3200, opened 2021-11-18
Filter NULL arguments out when calling non-strict functions/operators

### LINKED ISSUE #3195: Functions don't always have correct strictness behavior with empty sets
Consider, in our `issues` test repo:
```
create function foo(x: str, y: int64) -> str USING (x);

SELECT BooleanTest { name, val, x := foo(.name, .val) };
```

Here, we'll incorrectly output `.name`, even when `.val` is empty.

This only happens downstream of something actually being represented as `NULL` in SQL, like when it comes from a table; `SELECT foo("test", (SELECT 1 FILTER false))` behaves correctly

msullivan:
We have two related problems with a related solution:
 - Our implementation of the AND operator, which carefully
   ensures that we return NULL if either input is NULL,
   is a huge performance problem that prevents the use of indexes
 - User defined functions with implementations that aren't strict in
   their arguments can end up producing values when they shouldn't.

To fix this, we add NOT NULL filters when calling functions/operators
with potentially NULL values. To avoid doing this *too* much (and to
make it work more often in singleton mode), we add a notion of
function/operators having a strict implementation, which allows
the check to be skipped. Currently all built-in functions and operators
except for AND and OR are marked as being strict.

Fixes some #3185 related issues.
Fixes #3195.

### DIFF
--- a/edb/edgeql/compiler/func.py
+++ b/edb/edgeql/compiler/func.py
@@ -245,11 +245,6 @@ def compile_FunctionCall(
     else:
         tuple_path_ids = []
 
-    impl_is_strict = (
-        func.get_impl_is_strict(env.schema)
-        and func.get_language(env.schema) != qlast.Language.EdgeQL
-    )
-
     fcall = irast.FunctionCall(
         args=final_args,
         func_shortname=func_name,
@@ -274,7 +269,7 @@ def compile_FunctionCall(
         variadic_param_type=variadic_param_type,
         func_initial_value=func_initial_value,
         tuple_path_ids=tuple_path_ids,
-        impl_is_strict=impl_is_strict,
+        impl_is_strict=func.get_impl_is_strict(env.schema),
     )
 
     ir_set = setgen.ensure_set(fcall, typehint=rtype, path_id=path_id, ctx=ctx)
--- a/edb/edgeql/compiler/func.py
+++ b/edb/edgeql/compiler/func.py
@@ -245,6 +245,11 @@ def compile_FunctionCall(
     else:
         tuple_path_ids = []
 
+    impl_is_strict = (
+        func.get_impl_is_strict(env.schema)
+        and func.get_language(env.schema) != qlast.Language.EdgeQL
+    )
+
     fcall = irast.FunctionCall(
         args=final_args,
         func_shortname=func_name,
@@ -269,6 +274,7 @@ def compile_FunctionCall(
         variadic_param_type=variadic_param_type,
         func_initial_value=func_initial_value,
         tuple_path_ids=tuple_path_ids,
+        impl_is_strict=impl_is_strict,
     )
 
     ir_set = setgen.ensure_set(fcall, typehint=rtype, path_id=path_id, ctx=ctx)
@@ -592,6 +598,7 @@ def compile_operator(
         typeref=typegen.type_to_typeref(rtype, env=env),
         typemod=oper.get_return_typemod(env.schema),
         tuple_path_ids=[],
+        impl_is_strict=oper.get_impl_is_strict(env.schema),
     )
 
     _check_free_shape_op(node, ctx=ctx)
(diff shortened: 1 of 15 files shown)


edb/pgsql/compiler/expr.py
```diff
@@ -315,6 +315,24 @@ def compile_SliceIndirection(
...
+) -> List[pgast.BaseExpr]:
+    args = [dispatch.compile(a.expr, ctx=ctx) for a in expr.args]
+    for ref, ir_arg, typemod in zip(args, expr.args, expr.params_typemods):
+        if (
+            not expr.impl_is_strict
+            and ir_arg.cardinality.can_be_zero()
+            and ref.nullable
+            and typemod == ql_ft.TypeModifier.SingletonType
+        ):
+            raise RuntimeError(
```
elprans, line 330: Is this a temporary assertion?  If not, then this should be `errors.UnsupportedFeatureError`

edb/schema/functions.py
```diff
@@ -740,6 +740,9 @@ class CallableObject(
     abstract = so.SchemaField(
         bool, default=False, inheritable=False, compcoef=0.909)
 
+    impl_is_strict = so.SchemaField(
+        bool, default=True, compcoef=0.4)
```
elprans, line 744: We have to enforce this by 1) Creating functions with `STRICT` if this is `True`; or 2) For `FROM SQL FUNCTION` functions, validate that the underlying function is `STRICT`.

edb/pgsql/compiler/expr.py
msullivan: I made it a RuntimeError because that's what all the similar errors are, but they should probably all be UnsupportedFeatureError

edb/schema/functions.py
msullivan: 1. Currently we just ignore this on user created functions, though I'll make it go and error instead if it is explicitly specified.

2. Do you know of a way to automatically do that validation?

elprans: See how we use `sql_rval_consistency_check`.  This check would be a lot simpler, you'd only need to query `pg_proc.proisstrict` and `edgedb.raise` if it doesn not match `impl_is_strict`.

elprans: > if it is explicitly specified

It should be impossible to specify in user DDL unless the Field is declared with `allow_ddl_set`, so we're good there.

edb/edgeql/compiler/func.py
```diff
@@ -245,6 +245,11 @@ def compile_FunctionCall(
     else:
         tuple_path_ids = []
 
+    impl_is_strict = (
+        func.get_impl_is_strict(env.schema)
+        and func.get_language(env.schema) != qlast.Language.EdgeQL
```
1st1, line 250: Can you add a comment explaining the `and != edgeql` part?

edb/pgsql/compiler/expr.py
1st1: Can this surface in a user query? If so, a RuntimeErorr would become an ISE. Maybe we cleanup non-ISE RuntimeError in this file in a separate commit as part of this PR?

tests/test_edgeql_ddl.py
```diff
@@ -8898,7 +8896,10 @@ async def test_edgeql_ddl_enum_05(self):
                  CREATE CONSTRAINT expression ON (
                      <str>.num != asdf2()
                  );
-                 CREATE INDEX ON (asdf(.color));
+                 # FIXME: An index on asdf(.color) isn't supported currently
+                 # because passing nullable arguments requires filtering.
+                 # CREATE INDEX ON (asdf(.color));
+                 CREATE INDEX ON (.color);
```
1st1, line 8902: Unless you want to fix this as part of this PR, I'd create an x-fail test so that we don't forget about this.

edb/edgeql/compiler/func.py
msullivan: Reworked this bit away

tests/test_edgeql_ddl.py
msullivan: I just fixed it instead

msullivan:
I implemented setting strict on functions we define via edgeql or sql, if all the arguments are required singleton arguments. I also added a verification check for all SQL functions used to back functions or operators; this actually found a bug, with `++` on arrays.

I didn't add a verification check for "using SQL operator", because the operators are pretty heavily overloaded and so the check would have needed to check against the argument types also to be useful. This is certainly doable, but I checked the tables manually and the only operator backed by a non-strict function is `||` for arrays, but we actually call `array_cat` on arrays directly anyways, so it didn't seem like it was worth the hassle.

1st1 (commented):
lgtm

elprans (approved):
Looks good!

PR #3203, opened 2021-11-19
Implement index and slicing of json strings.

vpetrovykh:
Even though Postgres doesn't have support for index or slicing
operations on a json string, we can implement it similarly to how we
implemented `++` for json strings.

### DIFF
--- a/edb/pgsql/metaschema.py
+++ b/edb/pgsql/metaschema.py
@@ -1097,9 +1097,9 @@ class AssertJSONTypeFunction(dbops.Function):
                     msg => coalesce(
                         msg,
                         (
-                            'expected json '
+                            'expected JSON '
                             || array_to_string(typenames, ' or ')
-                            || '; got json '
+                            || '; got JSON '
                             || coalesce(jsonb_typeof(val), 'UNKNOWN')
                         )
                     ),
@@ -1742,7 +1742,7 @@ class StringIndexWithBoundsFunction(dbops.Function):
                 1
             ),
             'invalid_parameter_value',
-            'string index ' || "index"::text || ' is out of bounds',
+            "typename" || ' index ' || "index"::text || ' is out of bounds',
             "detail"
         )
     '''
@@ -1754,6 +1754,7 @@ class StringIndexWithBoundsFunction(dbops.Function):
                 ('val', ('text',)),
                 ('index', ('bigint',)),
                 ('detail', ('text',)),
+                ('typename', ('text',), "'string'"),
             ],
             returns=('text',),
             # Same volatility as raise_on_empty
@@ -1936,7 +1937,7 @@ class JSONIndexByTextFunction(dbops.Function):
                     val -> index,
                     'invalid_parameter_value',
                     msg => (
-                        'json index ' || quote_literal(index)
+                        'JSON index ' || quote_literal(index)
                         || ' is out of bounds'
                     ),
                     detail => detail
@@ -1947,7 +1948,7 @@ class JSONIndexByTextFunction(dbops.Function):
                     NULL::jsonb,
                     'wrong_object_type',
                     msg => (
-                        'cannot index json ' || jsonb_typeof(val)
+                        'cannot index JSON ' || jsonb_typeof(val)
                         || ' by ' || pg_typeof(index)::text
                     ),
                     detail => detail
@@ -1958,10 +1959,13 @@ class JSONIndexByTextFunction(dbops.Function):
                     NULL::jsonb,
                     'wrong_object_type',
                     msg => (
-                        'cannot index json '
+                        'cannot index JSON '
                         || coalesce(jsonb_typeof(val), 'UNKNOWN')
                     ),
-                    detail => detail
+                    detail => (
+                        '{"hint":"Retrieving an element by a string index '
+                        || 'is only available for JSON objects."}'
+                    )
                 )
             END
     '''
@@ -1992,7 +1996,7 @@ class JSONIndexByIntFunction(dbops.Function):
                     NULL::jsonb,
                     'wrong_object_type',
                     msg => (
-                        'cannot index json ' || jsonb_typeof(val)
+                        'cannot index JSON ' || jsonb_typeof(val)
                         || ' by ' || pg_typeof(index)::text
                     ),
                     detail => detail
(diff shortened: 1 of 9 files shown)


edb/pgsql/metaschema.py
```diff
@@ -2038,14 +2041,23 @@ def __init__(self) -> None:
...
+                        FROM jsonb_array_elements(val)
+                    ),
+                    start, stop
+                ))
+            )
+            WHEN 'string' THEN (
+                to_jsonb(_slice(val#>>'{}', start, stop))
+            )
+            ELSE
+                edgedb.jsonb_assert_type(val, ARRAY['array'])
```
msullivan, line 2059: Probably make this an `edgedb.raise`, like for regular indexing?

msullivan (approved):
Looks good! One requested tweak to error behavior.

edb/pgsql/metaschema.py
vpetrovykh: The indexing errors will be raised by the `_slice`, this error is for when the underlying `json` type is not sliceable.

I could inline `edgedb.raise` from `edgedb.jsonb_assert_type`:
```
                edgedb.raise(
                    NULL::jsonb,
                    'wrong_object_type',
                    msg => coalesce(
                        msg,
                        (
                            'expected json '
                            || array_to_string(typenames, ' or ')
                            || '; got json '
                            || coalesce(jsonb_typeof(val), 'UNKNOWN')
                        )
                    ),
                    detail => detail
                )
```

I'm not sure that inlining this would be an improvement.

msullivan: Hm, my actual complaint here is that it will say 'expected json array' when that isn't quite right, but the right fix is just passing in 'string' in the call to jsonb_assert_type.

(It might be worth switching to jsonb_assert_type for JSONIndexByIntFunction and JSONIndexByTextFunction, too, then, but no strong feeling there)

tests/test_edgeql_json.py
```diff
@@ -1378,6 +1381,43 @@ async def test_edgeql_json_slice_03(self):
...
+            r'''SELECT to_json('"hello"')[-100:100];''',
+            ['hello'],
+            ['"hello"'],
+        )
+
+        await self.assert_query_result(
+            r'''SELECT to_json('"hello"')[100:-100];''',
+            [''],
+            ['""'],
+        )
```
1st1, line 1419: I don't see error test cases, e.g.: 

`SELECT to_json('"hello"')[20]`

`SELECT to_json('["hello"]')[0][20]`

`SELECT to_json('["hello"]')[0][100]`

and also that empty slices like below returning empty strings

`SELECT to_json('"hello"')[:-100]`

`SELECT to_json('["hello"]')[0][100:]`

edb/pgsql/metaschema.py
1st1: +1 to Sully here: in the "ELSE" case just raise an error clearly explaining that the `[]` operator works on JSON arrays and strings, but not on `jsonb_typeof(val)`.

1st1 (changes requested):
Nice work but I think it needs a bit more tweaking before landing

1st1:
Also we need to update the docs.

edb/pgsql/metaschema.py
vpetrovykh: OK, I've adjusted the error message to be more consistent with the other slice and index errors for json. I've also added a hint with the information regarding the correct usage.

Speaking of hints, I seem to recall we used to display them in REPL, but I cannot see them for this or for some other errors, like `select <datetime>'1234-11-11';` informing about the expected ISO format. The hints do appear to reach `errormech.py`, though.

edb/pgsql/metaschema.py
```diff
@@ -1961,7 +1962,10 @@ class JSONIndexByTextFunction(dbops.Function):
                         'cannot index json '
                         || coalesce(jsonb_typeof(val), 'UNKNOWN')
                     ),
-                    detail => detail
+                    detail => (
+                        '{"hint":"Retrieving an element by an string index '
```
1st1, line 1966:
```suggestion
                        '{"hint":"Retrieving an element by a string index '
```

edb/pgsql/metaschema.py
```diff
@@ -1961,7 +1962,10 @@ class JSONIndexByTextFunction(dbops.Function):
...
+                        || 'is only available for json arrays and objects."}'
```
1st1, line 1967: Why do we allow that for arrays?

Also please use all-caps spelling of JSON

edb/pgsql/metaschema.py
```diff
@@ -2014,7 +2026,10 @@ class JSONIndexByIntFunction(dbops.Function):
                         'cannot index json '
                         || coalesce(jsonb_typeof(val), 'UNKNOWN')
                     ),
-                    detail => detail
+                    detail => (
+                        '{"hint":"Retrieving an element by an integer index '
+                        || 'is only available for json arrays and strings."}'
```
1st1, line 2031:
```suggestion
                        || 'is only available for JSON arrays and strings."}'
```

edb/pgsql/metaschema.py
```diff
@@ -2038,14 +2053,34 @@ def __init__(self) -> None:
...
+            ELSE
+                edgedb.raise(
+                    NULL::jsonb,
+                    'wrong_object_type',
+                    msg => (
+                        'cannot slice json '
+                        || coalesce(jsonb_typeof(val), 'UNKNOWN')
+                    ),
+                    detail => (
+                        '{"hint":"Slicing is only available for json arrays'
```
1st1, line 2079:
```suggestion
                        '{"hint":"Slicing is only available for JSON arrays'
```

tests/test_edgeql_json.py
```diff
@@ -447,6 +439,53 @@ async def test_edgeql_json_accessor_22(self):
...
+            r'''select to_json('"hello"')[0] = <json>'h';''',
+            [True],
+        )
+
+        await self.assert_query_result(
+            r'''select to_json('"hello"')[-2] = <json>'l';''',
+            [True],
+        )
+
+    async def test_edgeql_json_accessor_24(self):
```
1st1, line 453: I'd just combine test_edgeql_json_accessor_{24..27} into one test.

edb/pgsql/metaschema.py
vpetrovykh: Fixed.

tests/test_edgeql_json.py
vpetrovykh: We do this quite often. Historically I think it just helped getting a sense of which features worked and which still needed fixing/refining.

I can merge it in this case, since the error handling is introduced in one PR here.

edb/pgsql/metaschema.py
vpetrovykh: You are right, we don't allow this for arrays, we just have a more specialized error for them. Wording fixed.

I actually used lower-case `json` on purpose here because that is the name of the EdgeDB JSON scalar. But if you think that we should make it `JSON`, would that also apply to the main error message, such as `json index 10 is out of bounds`? There are a few of those, too.

vpetrovykh:
I think hints aren't always rendered by REPL, see https://github.com/edgedb/edgedb-cli/issues/559

edb/pgsql/metaschema.py
1st1: > I actually used lower-case json on purpose here because that is the name of the EdgeDB JSON scalar.

Then just spell it `std::json`?

tests/test_edgeql_json.py
1st1: Let's merge into one -- the test suite will run a bit faster.

1st1 (approved):

edb/pgsql/metaschema.py
vpetrovykh: After reviewing the error messages themselves I came to the conclusion that using `JSON` is more appropriate because a lot of the messages refer to `JSON string` or `JSON object` or `JSON array`, which are various valid concepts, but are not distinct types. Also the ambiguous meaning of `std::json array` would be unhelpful (is it `array<json>` or `json` that happens to be an array internally?)

1st1:
@elprans Elvis please cherry-pick this into 1.x

PR #3204, opened 2021-11-19
Allow multiple `--emit-server-status` arguments

elprans:
It is possible to be wanting to specify multiple server status sinks.
One such use case is the server docker image entrypoint, which uses
`--emit-server-status` to monitor and report auto-generated TLS
certificates, but this functionality is disabled if
`--emit-server-status` was already specified.

### DIFF
--- a/edb/server/args.py
+++ b/edb/server/args.py
@@ -103,7 +103,7 @@ class ServerConfig(NamedTuple):
     auto_shutdown_after: float
 
     startup_script: Optional[StartupScript]
-    status_sink: Optional[Callable[[str], None]]
+    status_sinks: List[Callable[[str], None]]
 
     tls_cert_file: Optional[pathlib.Path]
     tls_key_file: Optional[pathlib.Path]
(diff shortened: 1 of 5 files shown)


edb/server/args.py
```diff
@@ -710,32 +711,37 @@ def parse_args(**kwargs: Any):
...
+                status_sink = _status_sink_file(
+                    status_sink_addr[len('file://'):])
+            elif status_sink_addr.startswith('fd://'):
+                try:
+                    fileno = int(status_sink_addr[len('fd://'):])
+                except ValueError:
+                    abort(
+                        f'invalid file descriptor number in '
+                        f'--emit-server-status: '
+                        f'{status_sink_addr[len("fd://")]!r}'
```
1st1, line 728:
```suggestion
                        f'{status_sink_addr[len("fd://"):]!r}'
```

edb/server/args.py
```diff
@@ -710,32 +711,37 @@ def parse_args(**kwargs: Any):
...
+                    )
+
+                status_sink = _status_sink_fd(fileno)
+            elif m := re.match(r'(^\w+)://', status_sink_addr):
```
1st1, line 732:
```suggestion
            elif m := re.match(r'^(\w+)://', status_sink_addr):
```

1st1 (approved):

PR #3206, opened 2021-11-19
Support path prefix in link constraints

msullivan:
Clean up some other code in schema dealing with path_prefix_anchor
while we're at it.


elprans (approved):

1st1:
@elprans Should we merge this and include in 1.x?

PR #3209, opened 2021-11-23
Docs revisions

colinhacks:
Related to: https://github.com/edgedb/edgedb.com/pull/400

- Remove references to `tls-verify-hostname`
- Clarify Connection docs
- Fix typos and errors in docs
- Address Victor's feedback


docs/clients/connection.rst
```diff
@@ -6,26 +6,37 @@ Connection
...
+- Explicitly pass a DSN or :ref:`instance name
   <ref_reference_connection_instance_name>`
-  to the client creation function:
+  into the client creation function:
   ``edgedb.createClient`` in JS, ``edgedb.create_client()`` in Python, and
   ``edgedb.CreateClient`` in Go.
 
+  .. code-block:: typescript
+
+    const client = await edgedb.createClient({
```
1st1, line 34: No need for `await` here.

docs/reference/edgeql/cardinality.rst
```diff
@@ -0,0 +1,168 @@
...
+cardinality of each input and apply the following transformations, based on
+the type qualifier (or lack thereof) for each:
+
+.. code-block::
+
+  element-wise:  N -> N
+  optional:      N -> N == 0 ? 1 : N
+  aggregate:     N -> 1
+
+The cardinality of the resulting set is the product of the resulting numbers.
```
vpetrovykh, line 167: This is an oversimplification. What this chart and rule will actually give you is the _number of function invocations_. The actual final cardinality of the result set is going to depend on the cardinality of each of those invocations. It so happens that a great many functions in the standard library return exactly 1 result per invocation (`count`, `len`, `in`, pretty much all of the arithmetic operators, etc.). However, some have *optional* results, like `json_get(<json>[0,1,2], '10')` will actually return `{}`. Others return *multiple* (declared as `set of`) results, like `array_unpack([0,1,2])` which will return `{0, 1, 2}`. So the cardinality of the result ultimately depends of what kind of function you're calling and what can it possibly return for the particular inputs.

You can also define custom functions that return some set for each given input:
```
edgedb> create function foo(x: str) -> set of str 
....... using (select r := array_unpack(str_split(x, '')) filter r in {'a', 'e', 'i', 'o', 'u'});
OK: CREATE FUNCTION
edgedb> select foo({'hello', 'world'});
{'e', 'o', 'o'}
edgedb> select foo({'qqq', 'www'});
{}
```
Currently we don't allow custom functions with aggregate arguments, but they certainly don't have to produce only a single result either.

vpetrovykh (approved):
I left a few comments that should be addressed.

I like moving the pedantic details into the cardinality section, but if that's the reference it should certainly end up with an accurate picture in the end. Currently it still oversimplifies things.

docs/edgeql/select.rst
```diff
@@ -22,19 +22,47 @@ seen simple queries that select primitive values.
...
+  db> with x := {1, 2, 3, 4, 5}
+  ... select x
+  ... order by x desc;
+  {5, 4, 3, 2, 1}
+  db> with x := {1, 2, 3, 4, 5}
+  ... select x
+  ... offset 1 limit 3;
+  {2, 3, 4}
+
+This queries can also be rewritten to use inline aliases, like so:
```
vpetrovykh, line 47:
```suggestion
These queries can also be rewritten to use inline aliases, like so:
```

docs/reference/edgeql/cardinality.rst
```diff
@@ -0,0 +1,174 @@
...
+^^^^^^^^^^^^^^^
+
+When defining functions, all inputs are element-wise by default. The
+``set of`` :ref:`type qualifier  <ref_sdl_function_typequal>` is used to
+designate an input as *aggregate*. Currently this modifier is not supported
+for user-defined functions, but it is used by certain standard library
+functions.
+
+Similarly the ``optional`` qualifier marks the input as optional; an operation
+will be executed is an optional input is empty or omitted, whereas passing an
```
vpetrovykh, line 152:
```suggestion
will be executed is an optional input is empty, whereas passing an
```
Being `optional` has nothing to do with omitting the input altogether. You can only omit an argument if either there's a polymorphic version of the same function without the argument or if the argument has a _default_ value defined. In many cases it's useful to define both optional and default like so `x: optional str = {}`, precisely so that it can be omitted, but that's not always the case. E.g. obviously the first argument to `??` cannot be omitted.

docs/reference/edgeql/cardinality.rst
```diff
@@ -0,0 +1,174 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^
+
+To compute the number of times a function/operator will be invoked, take the
+cardinality of each input and apply the following transformations, based on
+the type qualifier (or lack thereof) for each:
+
+.. code-block::
+
+  element-wise:  N -> N
+  optional:      N -> N == 0 ? 1 : N
```
vpetrovykh, line 170:
```suggestion
  optional:      N -> max(1, N)
```
I feel like a more generic function such as "max" is a better fit than the ternary "... ? ... : ..." operator that's only present in some programming languages.

1st1 (approved):
Massive!

PR #3213, opened 2021-11-23
Replace `--generate-self-signed-cert` with `--tls-cert-mode`

elprans:
The `--generate-self-signed-cert` option and the corresponding
`EDGEDB_SERVER_GENERATE_SELF_SIGNED_CERT` are deprecated are replaced
with the `--tls-cert-mode` (`EDGEDB_SERVER_TLS_MODE`) enum with the
following values:

* `require_file` -- requires TLS certificate and key to be specified
  and to be valid files;
* `generate_self_signed` -- permits the server to auto-generate
  a self-signed certificate, possibly storing the result in
  `--tls-cert-file` and `--tls-key-file` if those are set, but
  point to non-existent files.

The default is determined by `--security` as before.

### DIFF
--- a/docs/changelog/1_0_rc2.rst
+++ b/docs/changelog/1_0_rc2.rst
@@ -301,8 +301,13 @@ Server configuration
     - ``never``
 
 ``EDGEDB_SERVER_TLS_CERT_MODE``
-    - ``file == default``
-    - ``self_signed`` — generate self-signed on bootstrap and when expired
+    - ``require_file`` — requires a valid TLS certificate and key to be
+      specified
+    - ``generate_self_signed`` generate self-signed certificate and private
+      key on bootstrap if certificate or key are not specified or missing
+    - ``default`` (equals to ``require_file`` if ``EDGEDB_SERVER_SECURITY``
+      is set to ``strict``, equals to ``generate_self_signed`` if
+      ``EDGEDB_SERVER_SECURITY`` is set to ``insecure_dev_mode``)
 
 ``EDGEDB_DOCKER_SHOW_GENERATED_CERT``
     - ``always == default``
--- a/edb/server/args.py
+++ b/edb/server/args.py
@@ -25,7 +25,6 @@ import os
 import pathlib
 import re
 import warnings
-import sys
 import tempfile
 
 import click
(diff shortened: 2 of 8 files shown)


edb/server/args.py
```diff
@@ -395,27 +405,48 @@ def _writer(status: str) -> None:
...
+        type=click.Choice(
+            ['default'] + list(ServerTlsCertMode.__members__.values()),
+            case_sensitive=True,
+        ),
+        default='default',
+        help='Specifies what to do when the TLS certificate and key are '
+             'either not specified or are missing.  When set to '
+             '"require_file", the TLS certificate and key must be specified '
+             'in the --tls-cert-file and --tls-key-file options and both must '
+             'exist.  When set to "self_Signed" a new self-signed certificate '
```
fantix, line 436:
```suggestion
             'exist.  When set to "generate_self_signed" a new self-signed certificate '
```

fantix (approved):
Submitting my mobile version of LGTM here - I will try to run it locally tmr morning.

docs/changelog/1_0_rc2.rst
```diff
@@ -301,8 +301,9 @@ Server configuration
     - ``never``
 
 ``EDGEDB_SERVER_TLS_CERT_MODE``
-    - ``file == default``
-    - ``self_signed`` — generate self-signed on bootstrap and when expired
+    - ``require_file`` (``== default`` if EDGEDB_SERVER_SECURITY is ``strict``)
+    - ``generate_self_signed`` generate self-signed certificate and private
+      key on bootstrap if certificate or key are not specified or missing.
```
1st1, line 306: You probably want this in rc3 changelog

elprans: We had this in rc2 changelog :laughing:

tailhook: We probably want to delete this from rc2 and add to rc3?

docs/changelog/1_0_rc2.rst
```diff
@@ -301,8 +301,9 @@ Server configuration
     - ``never``
 
 ``EDGEDB_SERVER_TLS_CERT_MODE``
-    - ``file == default``
-    - ``self_signed`` — generate self-signed on bootstrap and when expired
+    - ``require_file`` (``== default`` if EDGEDB_SERVER_SECURITY is ``strict``)
```
tailhook, line 304: What is default if security is not `strict`?

tailhook (approved):

edb/server/args.py
elprans: Fixed.

docs/changelog/1_0_rc2.rst
elprans: Clarified.

docs/changelog/1_0_rc2.rst
elprans: Yes.

PR #3215, opened 2021-11-25
Require both TLS certificate and key to be explicitly specified

elprans:
Whenever either `--tls-cert-file` or `--tls-key-file` are specified, the other must be also.

### DIFF
--- a/edb/server/args.py
+++ b/edb/server/args.py
@@ -692,27 +692,28 @@ def parse_args(**kwargs: Any):
     if kwargs['tls_key_file'] and not kwargs['tls_cert_file']:
         abort('When --tls-key-file is set, --tls-cert-file must also be set.')
 
+    if kwargs['tls_cert_file'] and not kwargs['tls_key_file']:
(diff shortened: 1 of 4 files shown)


1st1 (approved):

tailhook (commented):
Appending keyfile to the cert file if key file is not specified looks very dangerous. I would always assume that the cert file is public and feel free to copy it somewhere for everyone's access, just to realize there is also a private key.

It's probably fine for reading. And it's probably better to assume that use wants combined thing only when `key_file == cert_file`. Or at least when `cert_file` is omitted, not `key_file`.

elprans:
> Appending keyfile to the cert file if key file is not specified looks very dangerous

Well, this is fairly common practice.  To a point where Python's TSL API [explicitly allows passing a combined `.pem`](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.load_cert_chain).  That said, I agree that here the convenience is likely not worth the additional risk and we should make both arguments required if either is specified.

fantix (approved):

1st1 (approved):

PR #3216, opened 2021-11-25
Prevent mutation of free objects

elprans:
Free objects are immutable, so any attempt at DML on them is an error.


1st1 (approved):
Do we have tests that you can't inherit from FreeObjects?

1st1:
Also need to make sure that it's not possible to have links pointing to FreeObjects (or computed links that resolve into a FreeObject).

elprans:
I'll add the other checks in a separate PR.

PR #3217, opened 2021-11-26
Prohibit using FreeObject as a base or linking to it

elprans:
Neither makes sense semantically.

### DIFF
--- a/edb/schema/indexes.py
+++ b/edb/schema/indexes.py
@@ -23,6 +23,7 @@ from typing import *
 from edb import edgeql
 from edb import errors
 from edb.common import ast
+from edb.common import parsing
 from edb.common import verutils
 from edb.edgeql import ast as qlast
 from edb.edgeql import compiler as qlcompiler
@@ -201,6 +202,7 @@ class IndexCommand(
         *,
         name: Optional[sn.Name] = None,
         default: Union[Index, so.NoDefaultT] = so.NoDefault,
+        sourcectx: Optional[parsing.ParserContext] = None,
     ) -> Index:
         ...
 
@@ -212,6 +214,7 @@ class IndexCommand(
         *,
         name: Optional[sn.Name] = None,
         default: None = None,
+        sourcectx: Optional[parsing.ParserContext] = None,
     ) -> Optional[Index]:
         ...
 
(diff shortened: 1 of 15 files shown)


edb/schema/indexes.py
```diff
@@ -222,6 +225,7 @@ def get_object(  # NoQA: F811
         *,
         name: Optional[sn.Name] = None,
         default: Union[Index, so.NoDefaultT, None] = so.NoDefault,
+        sourcectx: Optional[parsing.ParserContext] = None,
```
1st1, line 228: This new param isn't used in the body of the function

edb/schema/objtypes.py
```diff
@@ -100,10 +100,15 @@ def is_object_type(self) -> bool:
...
 
     def is_free_object_type(self, schema: s_schema.Schema) -> bool:
-        return self.issubclass(
-            schema,
-            schema.get('std::FreeObject', type=ObjectType),
-        )
+        if self.get_name(schema) == sn.QualName('std', 'FreeObject'):
+            return True
+
+        FreeObj = schema.get('std::FreeObject', type=ObjectType, default=None)
```
1st1, line 106: nit: I'd keep the full name ;)

```suggestion
        FreeObject = schema.get('std::FreeObject', type=ObjectType, default=None)
```

tests/test_schema.py
```diff
@@ -275,26 +275,44 @@ def test_schema_overloaded_link_04(self):
...
 
+    @tb.must_fail(
+        errors.InvalidLinkTargetError,
+        "object type 'std::FreeObject' is not a valid link target",
+        position=69,
+    )
+    def test_schema_bad_link_03(self):
+        """
+            type Object {
+                link foo -> FreeObject
```
1st1, line 312: Can you add one more test testing `link foo := (SELECT {a := 1})`?

1st1 (approved):
I've left a couple nit comments. I'm not sure if what's in this PR is enough to prohibit computed links with expression returning a FreeObject; if it is, it needs a test. If not, a follow up PR?

elprans:
> f it is, it needs a test. If not, a follow up PR?

A follow up, because the fix is not straightforward.

Issue #3218, opened 2021-11-27
Warning during bootstrap


tailhook:
- EdgeDB Version: 1.0-dev.6229+004227f

Warns during bootstrap:
```
Successfully installed 1.0-dev.6229+004227f
Initializing EdgeDB instance...
WARNING 268 2021-11-27T16:25:05.691 py.warnings: /home/user1/.local/share/edgedb/portable/1.0-dev.6229/lib/python3.10/site-packages/edb/server/server.py:1288: RuntimeWarning: coroutine '_signal_sysevent' was never awaited
  await binary.EdgeConnection.run_script(
```

fantix:
should be fixed together with #3112

tailhook:
Then #3112 should be prioritized I think. Seeing this warning each time you create an instance on 1.0 would be unfortunate. Or maybe we can come up with a simpler fix for just this warning?

PR #3220, opened 2021-11-29
Mute coroutine warning at shutdown by immediate cancellation

fantix:
After revisiting #3112, I found a more reliable option 4 to address the "coroutine ... was never awaited" warning - cancel the late tasks immediately after creating them.

Closes #3112

### DIFF
--- a/edb/server/protocol/binary.pyx
+++ b/edb/server/protocol/binary.pyx
@@ -1194,6 +1194,8 @@ cdef class EdgeConnection:
         return query_unit
 
     def signal_side_effects(self, side_effects):
+        if not self.server._accept_new_tasks:
+            return
         if side_effects & dbview.SideEffects.SchemaChanges:
             self.server.create_task(
                 self.server._signal_sysevent(
@@ -2231,7 +2233,10 @@ cdef class EdgeConnection:
         return out_buf
 
     def connection_made(self, transport):
-        if not self.server._accepting_connections:
+        if (
+            not self.server._accepting_connections
+            or not self.server._accept_new_tasks
+        ):
             transport.abort()
             return
 
@@ -2307,6 +2312,7 @@ cdef class EdgeConnection:
                     # reaches the backend, we'll be setting a trap for the
                     # _next_ query that is unlucky enough to pick up this
                     # Postgres backend from the connection pool.
+                    # TODO(fantix): hold server shutdown to complete this task
                     self.server.create_task(
                         self.server._cancel_and_discard_pgcon(
                             self._pinned_pgcon,
--- a/edb/server/protocol/binary.pyx
+++ b/edb/server/protocol/binary.pyx
@@ -1194,7 +1194,7 @@ cdef class EdgeConnection:
         return query_unit
 
     def signal_side_effects(self, side_effects):
-        if not self.server._accept_new_tasks:
+        if self.server.is_shutting_down():
             return
         if side_effects & dbview.SideEffects.SchemaChanges:
             self.server.create_task(
@@ -2234,8 +2234,8 @@ cdef class EdgeConnection:
 
     def connection_made(self, transport):
         if (
-            not self.server._accepting_connections
-            or not self.server._accept_new_tasks
+            not self.server.is_accepting_connections()
+            or self.server.is_shutting_down()
         ):
             transport.abort()
             return
--- a/edb/server/protocol/binary.pyx
+++ b/edb/server/protocol/binary.pyx
@@ -2233,10 +2233,7 @@ cdef class EdgeConnection:
         return out_buf
 
     def connection_made(self, transport):
-        if (
-            not self.server.is_accepting_connections()
-            or self.server.is_shutting_down()
-        ):
+        if not self.server.is_serving():
             transport.abort()
             return
 
--- a/edb/server/server.py
+++ b/edb/server/server.py
@@ -149,7 +149,7 @@ class Server(ha_base.ClusterProtocol):
 
         self._serving = False
         self._initing = False
-        self._accept_new_tasks = False
+        self._shutting_down = False
 
         self._cluster = cluster
         self._pg_addr = self._get_pgaddr()
@@ -208,7 +208,6 @@ class Server(ha_base.ClusterProtocol):
 
         self._binary_proto_id_counter = 0
         self._binary_conns = collections.OrderedDict()
-        self._accepting_connections = False
 
         self._servers = {}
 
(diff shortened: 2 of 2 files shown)


tailhook:
Can we prevent the task from starting instead?

tailhook:
For the case #3218, I would prefer doing `if not self.server.shutting_down()` in `signal_side_effects`. Because in this case we explicitly know that these side effects are irrelevant after shutdown. But I'm not sure about other (future) tasks.

fantix:
Yeah, that's exactly option 2 in #3112.

tailhook:
> Yeah, that's exactly option 2 in #3112. We may need both option 2 & 4 for both explicitness and correctness.

Do we have too many of such tasks? I think I have only seen that one in #3218 so far. Also there is no option for so I'm not sure what you mean.

fantix:
> Do we have too many of such tasks?

There're a few, but not too many. I've pushed [a fix](https://github.com/edgedb/edgedb/pull/3220/commits/55d192c904938f671be5dafa4298322d8ea0809e) to add a check to most of them.

> > Yeah, that's exactly option 2 in #3112. We may need both option 2 & 4 for both explicitness and correctness.
>
>  Also there is no option for so I'm not sure what you mean.

Oh I mean this in #3112 description:

> 2. Don't call `create_task()` when `server._accept_new_tasks == False`. This adds a lot of duplicate checks before calling `server.create_task()`.

edb/server/protocol/binary.pyx
```diff
@@ -1193,6 +1193,8 @@ cdef class EdgeConnection:
         return query_unit
 
     def signal_side_effects(self, side_effects):
+        if not self.server._accept_new_tasks:
```
tailhook, line 1196: This should have some "public" method or property. It's fine to use underscored attribute in `server.py` but not here.

edb/server/protocol/binary.pyx
```diff
@@ -2230,7 +2232,10 @@ cdef class EdgeConnection:
         return out_buf
 
     def connection_made(self, transport):
-        if not self.server._accepting_connections:
+        if (
+            not self.server._accepting_connections
+            or not self.server._accept_new_tasks
```
tailhook, line 2237: Same thing here. I'm not sure whether `_accepting_connections` and `_accept_new_tasks` should be exposed as a single public property/method or different ones, though.

tailhook (commented):
I would prefer `if not self.is_shutting_down()` or something similar. `_accept_new_tasks` doesn't say enough. Like does it temporary not accepting tasks? Should we retry in a timeout? Or maybe it doesn't accept because it is starting up?

edb/server/protocol/binary.pyx
fantix: Looks like we don't need `_accepting_connections` any more - I've created `Server.is_accepting_connections()` based on `Server._serving`.

fantix:
> I would prefer `if not self.is_shutting_down()` or something similar. `_accept_new_tasks` doesn't say enough. Like does it temporary not accepting tasks? Should we retry in a timeout? Or maybe it doesn't accept because it is starting up?

Well `_accept_new_tasks` is not 100% `!is_shutting_down()`, particularly in e.g. `Server.run_startup_script_and_exit()` before `Server.start()` is called. I've renamed `_accept_new_tasks` to `_shutting_down`, and use `Server._task_group is None` in `Server.create_task()` to check if we're allowed to create tasks. This should cover all cases.

edb/server/server.py
```diff
@@ -1564,17 +1580,29 @@ def create_task(self, coro, *, interruptable):
...
+                    f"    {path}, line {lineno}, in {method}"
+                    for path, lineno, method in origin
+                )
+            else:
+                detail += (
+                    "; set sys.set_coroutine_origin_tracking_depth() to debug"
+                )
+            logger.debug("Task is not started and ignored: %s", detail)
+            task = self.__loop.create_task(coro)
+            task.cancel()
```
tailhook, line 1599: If we are to keep this code, please move it to a function. Because when reading this code 60% of `craete_task` being cancellation is a distraction.

tailhook (commented):
> > I would prefer `if not self.is_shutting_down()` or something similar. `_accept_new_tasks` doesn't say enough. Like does it temporary not accepting tasks? Should we retry in a timeout? Or maybe it doesn't accept because it is starting up?
> 
> Well `_accept_new_tasks` is not 100% `!is_shutting_down()`, particularly in e.g. `Server.run_startup_script_and_exit()` before `Server.start()` is called. I've renamed `_accept_new_tasks` to `_shutting_down`, and use `Server._task_group is None` in `Server.create_task()` to check if we're allowed to create tasks. This should cover all cases.

Sorry, but my concern still holds here. I didn't mean literally `is_shutting_down` by the way. So `run_startup_script_and_exit` is bootstrap process, right? I.e. it's not expected for the server to be running for a long time, so updating schema doesn't make sense, right?

So it might be:
* `is_server_running()` or `is_active()`
* `server.is_serving()` or `server.is_accepting_connections()`
* `is_ready_to_shutdown()` (I think bootstrap fits this well)

I mean we can make a flag that checks both bases. Although, my concern with `is_accepting_connections` and similar is that someone might assume it should be false early or during a huge load. Maybe you can come up with even better name.

fantix:
At the moment, internally the `Server` has 4 states:
* `_initing=True`: while `Server.init()` is running
* `_serving=True`: since the TCP socket started to accept connections, before `Server.stop()` is called
* `_shutting_down=True`: after `Server.stop()` is called
* In the rest of the time, the server stays in a "default" state.

Bootstrap is like: `default` -> `initing` -> `default` (run_startup_script_and_exit)
Server from start to stop: `default` -> `initing` -> `default` (startup_script) -> `serving` -> `shutting_down`

* The server will abort new client connections immediately if not `serving`
* The server will not trigger `__edgedb_sysevent__` in `shutting_down` and `default` because the `sys_pgcon` is likely closing. This should be fixed I think to always trigger the event, but not handling the notification.

tailhook:
@fantix, from your description I'm still confused why we need `_discard_coro`? If it's just in case "we forgot something", let's make warnings visible at least in dev mode (although, since nobody were concerned about these warnings previously, until we saw them on installation, it maybe not enough).

edb/server/protocol/binary.pyx
```diff
@@ -1194,6 +1194,8 @@ cdef class EdgeConnection:
         return query_unit
 
     def signal_side_effects(self, side_effects):
+        if self.server.is_shutting_down():
```
elprans, line 1197: This would be `false` in bootstrap mode.  I think we need a single condition, e.g. `is_active()` to cover both the "is_shutting_down()" and "not started at all" case.

elprans:
I'm also not convinced of the utility of `_discard_coro`.  We are trading Python's standard "coroutine not awaited" for a custom and lower visibility "task ignored".  An explicit check in `signal_side_effects` should be enough for now.

1st1:
> I'm also not convinced of the utility of _discard_coro.

FWIW that function is one reason why I don't want to merge this PR.

PR #3223, opened 2021-11-29
Produce proper errors for FILTER clauses on required subshapes

msullivan:


tests/test_edgeql_select.py
```diff
@@ -6969,3 +6969,14 @@ async def test_edgeql_select_concat_null_01(self):
                 {'name': 'triangle', 'val': 10, 'x': [10, 0]},
             ],
         )
+
+    async def test_edgeql_select_subshape_filter_01(self):
+        async with self.assertRaisesRegexTx(
+            edgedb.QueryError,
+            "possibly an empty set returned",
```
elprans, line 6976: Ideally we should produce a good hint on what to do to fix the error with the use of `assert_exists`.

elprans (approved):

PR #3226, opened 2021-11-29
Avoid generating EXISTS where clauses for type filters

msullivan:
Since we don't run card inference on type rewrites, the where clause
cardinality always gets compiled as if it is multi, which it isn't.

Run card inference on type rewrites, and tweak compilation of FILTER
to be both more natural reading and to crash on this sort of bug
instead.

### DIFF
--- a/edb/edgeql/compiler/stmtctx.py
+++ b/edb/edgeql/compiler/stmtctx.py
@@ -136,11 +136,20 @@ def fini_expression(
         scope_tree=ctx.path_scope,
         ctx=inf_ctx,
     )
+    cardinality = inference.infer_cardinality(
+        ir,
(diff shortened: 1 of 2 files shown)


edb/edgeql/compiler/stmtctx.py
```diff
@@ -136,12 +136,21 @@ def fini_expression(
         scope_tree=ctx.path_scope,
         ctx=inf_ctx,
     )
+    cardinality = inference.infer_cardinality(
```
elprans, line 139: Hm. Why a second call to `infer_cardinality`?

msullivan: That's weird.

elprans (approved):

PR #3235, opened 2021-12-02
Fix errors in CLI config file docs

msullivan:


elprans (approved):

docs/cli/index.rst
```diff
@@ -118,13 +118,14 @@ The ``cli.toml`` has the following structure. All fields are optional:
...
     implicit-properties = false   # Print implicit properties of objects
-    implicit-limit = 100          # Set implicit LIMIT
+    limit = 100                   # Set implicit LIMIT
                                   # Defaults to 100, specify 0 to disable
     input-mode = "emacs"          # Set input mode. One of: vi, emacs
     output-format = "default"     # Set output format.
                                   # One of: default, json, json-pretty,
                                   # json-lines
-    print-stats = false           # Print statistics on each query
+    print-stats = "off"           # Print statistics on each query.
```
1st1, line 127: @tailhook @fantix 

This seems like a bug -- why `"off"` here and `false` for other bool props?

colinhacks: This is a ternary option "off"/"query"/"detailed" 

https://github.com/edgedb/edgedb-cli/blob/3aa3b0cb59e6455aaf0a856d9615ee12a81782d5/src/repl.rs#L38

1st1: Perhaps we should standardize other bool prefs to "on" / "off". Or the users will be confused.

1st1: Perhaps we should standardize other bool prefs to "on" / "off". Or the users will be confused.

@msullivan The setting help should list all valid setting values.

msullivan: Yeah that's the next line.

tailhook: @1st1 
> Perhaps we should standardize other bool prefs to "on" / "off". Or the users will be confused. 

We do allow `on`/`off` in the shell. But this is toml. Boolean values should be boolean values as specified by toml (so you don't have to use quotes and generally it's friendly for people knowing toml).

PR #3236, opened 2021-12-02
Add GCP deployment guide

fmoor:


docs/guides/deployment/gcp.rst
```diff
@@ -0,0 +1,191 @@
...
+
+Persist TLS Certificate
+=======================
+
+.. code-block:: bash
+
+   kubectl create secret generic cloudsql-tls-credentials \
+       --from-literal=tlskey="$(
+           kubectl exec deploy/edgedb -c=edgedb -- \
+               cat /etc/ssl/edgedb/edbprivkey.pem
```
elprans, line 153: Let's replace this with a call to `edgedb-show-secrets.sh --format=raw EDGEDB_SERVER_TLS_KEY` and `edgedb-show-secrets.sh --format=raw EDGEDB_SERVER_TLS_CERT`

elprans (approved):

docs/guides/deployment/gcp.rst
```diff
@@ -0,0 +1,203 @@
...
+       --from-literal=instance=${INSTANCE_CONNECTION_NAME}=tcp:5432
+
+Deploy EdgeDB
+=============
+
+.. code-block:: bash
+
+   $ URL="https://raw.githubusercontent.com"
+   $ URL="${URL}/edgedb/edgedb-deploy/dev/gcp/deployment.yaml"
+   $ wget $URL
```
fmoor, line 137: Is there a better way to write this url so that it doesn't violate the 80 character line length limit? This looks strange in a guide:
image

If I use a backslash the second line is treated as output rather than part of the command and the copy button doesn't copy it.
image

cc @elprans

elprans: This is a site bug, backslash should be interpreted correctly.

Issue #3234, opened 2021-12-02
Defaulting to `EMPTY LAST` in `ORDER BY .. DESC` prevents the use of backward index scan resulting in poor perf


elprans:
See #3232 for context.  We should avoid emitting `EMPTY LAST` implicitly (both in GraphQL and EdgeQL translation).

msullivan:
Additionally, placing the ORDER BY value in a subquery contributes to the problem

PR #3238, opened 2021-12-03
Swap the default behavior of sorting by emptys sets

msullivan:
That is, make `EMPTY LAST` the default for `ASC` and `EMPTY FIRST` the
default for `DESC`. This seems like worse behavior, but it is pretty
marginal, and it is what SQL can do quickly.

It turns out we had zero tests for this behavior.

Partial fix to #3234.

### DIFF
--- a/docs/reference/edgeql/select.rst
+++ b/docs/reference/edgeql/select.rst
@@ -66,8 +66,8 @@ SELECT
     an empty set when evaluating an *expression* are sorted *after*
     all other values; if ``EMPTY FIRST`` is specified, then they
     are sorted *before* all other values.  If neither is specified,
-    ``EMPTY FIRST`` is assumed when ``ASC`` is specified or implied,
-    and ``EMPTY LAST`` when ``DESC`` is specified.
+    ``EMPTY LAST`` is assumed when ``ASC`` is specified or implied,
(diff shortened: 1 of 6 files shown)


elprans (approved):

vpetrovykh:
> This seems like worse behavior, but it is pretty
> marginal, and it is what SQL can do quickly.

I suppose that this is a marginal enough case (because it can be trivially overridden) that the fact that `{}` is fairly reasonable interpretation of "smallest" for the purpose of ordering can be overlooked, but is it *really* that big of a performance boost?

msullivan:
2ms vs 60ms, on a test doing 
`SELECT Block { hash, height } ORDER BY .height ASC EMPTY LAST LIMIT 50`
vs
`SELECT Block { hash, height } ORDER BY .height ASC EMPTY FIRST LIMIT 50`, respectively, on a table with a million rows.

(And, also with a pending PR of mine that tries to avoid putting the body of ORDER BY in a subquery.)

vpetrovykh:
Wow! OK, faster default it is then.

msullivan:
Oh, wait! We should not do this!

Postgres lets you specify where to put NULLS in an index! We should do that instead.

PR #3239, opened 2021-12-03
Avoid putting subqueries in SQL ORDER BY clauses

msullivan:
pg does a bad job understanding how to use indexes if the body of an
ORDER BY is a subquery. We need, in general, for it to be a subquery,
but most common cases do not, so try to collapse it in a simple expression.

Fixes related issue to #3234.

### DIFF
--- a/edb/pgsql/compiler/clauses.py
+++ b/edb/pgsql/compiler/clauses.py
@@ -244,6 +244,69 @@ def compile_filter_clause(
     return where_clause
 
 
+def _get_target_from_range(
+    target: pgast.BaseExpr, rvar: pgast.BaseRangeVar
+) -> Optional[pgast.BaseExpr]:
+    """Try to read a target out of a very simple rvar.
+
+    The goal here is to allow collapsing trivial pass-through subqueries.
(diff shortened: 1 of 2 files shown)


edb/pgsql/compiler/clauses.py
```diff
@@ -256,8 +319,12 @@ def compile_orderby_clause(
 
             # In OPDER BY we compile ir.Set as a subquery:
```
elprans, line 320: Typo fix opportunity:

```suggestion
            # In ORDER BY we compile ir.Set as a subquery:
```

edb/pgsql/compiler/clauses.py
```diff
@@ -244,6 +244,69 @@ def compile_filter_clause(
...
+        or len(query.target_list) != 1
+        or len(query.from_clause) != 1
+    ):
+        return query
+
+    val = _get_target_from_range(
+        query.target_list[0].val, query.from_clause[0])
+    if val:
+        return val
+    return query
```
elprans, line 307:
```suggestion
    else:
        return query
```

elprans (approved):

PR #3243, opened 2021-12-03
Substantially broaden what can appear as the argument to FOR

msullivan:
Introduce a notion of AtomicExpr and allow it as the argument to FOR.

### DIFF
--- a/edb/edgeql/parser/parser.py
+++ b/edb/edgeql/parser/parser.py
@@ -117,8 +117,8 @@ class EdgeQLParserBase(parsing.Parser):
                     else:
                         msg = f'Unexpected {token.text()!r}'
                 elif rule == 'for iterator':
-                    msg = ("Complex expressions in a FOR query must be "
-                           "parenthesized")
+                    msg = ("Complex expressions in a FOR iterator clause must "
+                           "be parenthesized")
 
                     if i > 0:
                         context = pctx.merge_context([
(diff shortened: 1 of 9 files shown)


edb/edgeql/parser/parser.py
```diff
@@ -116,8 +117,14 @@ def get_exception(self, native_err, context, token=None):
                     else:
                         msg = f'Unexpected {token.text()!r}'
                 elif rule == 'for iterator':
-                    msg = (f"missing '{{' before {token.text()!r} "
-                           f"in a FOR query")
+                    msg = ("Complex expressions in a FOR query must be "
```
elprans, line 120:
```suggestion
                    msg = ("Complex expressions in a FOR iterator clause must be "
```

elprans: Or, maybe "missing parentheses around complex expression in a FOR iterator clause"

msullivan: If we do the latter suggestion, should we also change "Statement used as an expression must be enclosed in a set of parentheses" to "Missing parentheses around statement used as expression"?

elprans: Sure

elprans (approved):

edb/edgeql/parser/parser.py
msullivan: Hm, I think I prefer the first approach, actually. Fine to stick with that?

PR #3249, opened 2021-12-06
Support single-role and single-database mode

fantix:
This PR is a WIP to support backends that doesn't allow `CREATE ROLE` or `CREATE DATABASE` like Heroku Postgres.

If `CREATE ROLE` is not allowed:

* The same backend user will be used for everything.
* Bootstrap only one EdgeDB user `edgedb`, mapped to the same backend user.
* `--tenant-id` is disallowed, EdgeQL `CREATE ROLE`, `DROP ROLE` will raise an error.
* Because we cannot `COMMENT ON` the role, the `pg_authid` metadata is stored in `edgedbinstdata.instdata['single_role_metadata']`
* `sys::Role` is shadowed by a completely different set of views.
* The template_db and system_db are still created as usual.

If `CREATE DATABASE` is not allowed:
* Assumes `CREATE ROLE` is not allowed - or else you could just create a new role with `CREATE DATABASE` permission
* No template database will be created
* Data which is usually found in a system database is stored in the same database
* Metadata of the "template database" and "system database" is stored as `edgedbinstdata.instdata['__edgedbtpl__metadata']` and `edgedbinstdata.instdata['__edgedbsys__metadata']`

TODO:

- [x] Drop `test-heroku.yml`, and add it to `test-pg-versions.yml`
- [ ] Add Terraform scripts to run tests on real Heroku Postgres
- [x] `test_server_proto_configure_compilation` is failing (but not local?)
- [x] I think the concurrent schema change is broken in this branch, like `edb test -j2 tests/test_edgeql_expressions.py` will fail.

### DIFF
--- a/edb/pgsql/delta.py
+++ b/edb/pgsql/delta.py
@@ -5364,7 +5364,8 @@ class DatabaseMixin:
                     SELECT
                         edgedb.raise(
                             NULL::uuid,
-                            msg => 'operation is not supported by the backend'
+                            msg => 'operation is not supported by the backend',
+                            exc => 'feature_not_supported'
                         )
                     INTO _dummy_text
                     '''
@@ -5431,7 +5432,8 @@ class RoleMixin:
                     SELECT
                         edgedb.raise(
                             NULL::uuid,
-                            msg => 'operation is not supported by the backend'
+                            msg => 'operation is not supported by the backend',
+                            exc => 'feature_not_supported'
                         )
                     INTO _dummy_text
                     '''
--- a/edb/pgsql/delta.py
+++ b/edb/pgsql/delta.py
@@ -371,14 +371,10 @@ class AlterGlobalSchemaVersion(
         else:
             backend_params = params.get_default_runtime_params()
 
-        instance_params = backend_params.instance_params
-        capabilities = instance_params.capabilities
-        tenant_id = instance_params.tenant_id
-
         tpl_db_name = common.get_database_backend_name(
-            edbdef.EDGEDB_TEMPLATE_DB, tenant_id=tenant_id)
+            edbdef.EDGEDB_TEMPLATE_DB, tenant_id=backend_params.tenant_id)
 
-        if capabilities & params.BackendCapabilities.SUPERUSER_ACCESS:
+        if backend_params.has_superuser_access:
             # Only superusers are generally allowed to make an UPDATE
             # lock on shared catalogs.
             lock = dbops.Query(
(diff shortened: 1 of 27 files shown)


edb/server/bootstrap.py
```diff
@@ -451,14 +485,6 @@ async def _make_stdlib(
     std_plans.append(schema_version)
     plan.generate(current_block)
 
-    stdglobals = '\n'.join([
```
elprans, line 466: I don't think this is correct.  If you remove this, the global schema would not have the correct record for the default superuser at bootstrap time (as well as the devmode schema cache).

edb/pgsql/delta.py
```diff
@@ -5315,12 +5353,32 @@ class DeleteModule(ModuleMetaCommand, adapts=s_mod.DeleteModule):
...
+class DatabaseMixin:
+    def ensure_has_create_database(self, backend_params):
+        if not backend_params.has_create_database:
+            self.pgops.add(
+                dbops.Query(
+                    f'''
+                    SELECT
+                        edgedb.raise(
+                            NULL::uuid,
+                            msg => 'operation is not supported by the backend'
```
elprans, line 5365:
```suggestion
                            exc => 'feature_not_supported'
```

edb/pgsql/delta.py
```diff
@@ -5359,12 +5420,32 @@ def apply(
...
+class RoleMixin:
+    def ensure_has_create_role(self, backend_params):
+        if not backend_params.has_create_role:
+            self.pgops.add(
+                dbops.Query(
+                    f'''
+                    SELECT
+                        edgedb.raise(
+                            NULL::uuid,
+                            msg => 'operation is not supported by the backend'
```
elprans, line 5432:
```suggestion
                            msg => 'operation is not supported by the backend',
                            exc => 'feature_not_supported'
```

edb/server/bootstrap.py
```diff
@@ -250,13 +243,77 @@ async def _is_pristine_cluster(ctx: BootstrapContext) -> bool:
         return False
 
 
+async def _is_pristine_cluster(ctx: BootstrapContext) -> bool:
```
elprans, line 246: There's a lot of branching here, can we simplify this method to look for single-db metadata specifically, and, if not found, fallback to regular tenant detection?

elprans (approved):
This looks good to me overall!  Let's address the remaining items and merge.

edb/server/bootstrap.py
fantix: Yeah, with the `--with-backend-capability`/`--without-backend-capability` options to be added, I think we should raise an error here in `ensure_bootstrapped()` if:

1. The backend capability indicates e.g. the lack of `CREATE DATABASE` permission, and
2. The backend is already bootstraped differently, e.g. not in single-db mode

And the `_is_pristine_cluster()` only needs to check for single-db metadata like you said. I'll update the PR.

fantix: UPDATE: Oh no, the regular detection couldn't find out if the backend is already bootstrapped in a single-role mode, because it's only checking the roles of `[<tenant_hash>_]edgedb_supergroup`. In this case, continuting to bootstrap as a pristine cluster will fail.

fantix: Pushed a fix to 1) raise errors and 2) run regular tenant detection first, then check single modes in an easier way.

fantix:
Sample run on simulated single modes: https://github.com/edgedb/edgedb/runs/4600878649

PR #3259, opened 2021-12-08
Fix referring to linkprops on WITH bound links

### LINKED ISSUE #3233: InternalServerError when using link property in sub query
- EdgeDB Version: 1.0-dev.6184+1833814

Steps to Reproduce:

```
edgedb> WITH
.......   __scope_0_Movie := (DETACHED default::Movie)
....... SELECT (__scope_0_Movie) {
.......   id,
.......   characters := (
.......     WITH
.......       __scope_1_Person := (__scope_0_Movie.characters)
.......     SELECT (__scope_1_Person) {
.......       name,
.......       @character_name
.......     }
.......   )
....... };
edgedb error: InternalServerError: None
```

Schema: https://github.com/edgedb/edgedb-js/blob/master/qb/dbschema/default.esdl

msullivan:
Fixes #3233.
Except maybe this is all bogus and should be rejected?

The interactions with type intersections are broken, but the relevant
cases of type intersection/linkprops seem broken anyway.

### DIFF
--- a/edb/pgsql/compiler/relgen.py
+++ b/edb/pgsql/compiler/relgen.py
@@ -714,6 +714,11 @@ def process_set_as_link_property_ref(
         else:
             link_prefix = ir_source
 
+        # If this is a WITH bound link, dive through a bit to get to it.
+        while link_prefix.rptr is None:
+            assert isinstance(link_prefix.expr, irast.SelectStmt)
+            link_prefix = link_prefix.expr.result
+
         source_scope_stmt = relctx.get_scope_stmt(
             ir_source.path_id, ctx=newctx)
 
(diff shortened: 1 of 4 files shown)


edb/pgsql/compiler/relgen.py
```diff
@@ -714,6 +714,11 @@ def process_set_as_link_property_ref(
         else:
             link_prefix = ir_source
 
+        # If this is a WITH bound link, dive through a bit to get to it.
```
elprans, line 717: We should disallow this.  There is no semantic "persistence" of the fact that you've traversed the link in a `WITH` expression.  Past that point it's just a set.

PR #3261, opened 2021-12-09
Fix bound doubly nested shapes refering to props in a long way

### LINKED ISSUE #3245: InternalServerError when selecting computed property created from link property
- EdgeDB Version: 1.0-dev.6184+1833814

Steps to Reproduce:
```
edgedb> WITH
.......   __scope_0_Movie := (DETACHED default::Movie)
....... SELECT (__scope_0_Movie) {
.......   id,
.......   characters := (
.......     WITH
.......       __scope_1_Person := (
.......         SELECT __scope_0_Movie.characters {
.......           __linkprop_character_name := __scope_0_Movie.characters@character_name
.......         }
.......       )
.......     SELECT (__scope_1_Person) {
.......       name,
.......       single @character_name := (__scope_1_Person.__linkprop_character_name)
.......     }
.......   )
....... };
edgedb error: InternalServerError: more than one row returned by a subquery used as an expression
  Server traceback:
      edb.errors.InternalServerError: more than one row returned by a subquery used as an expression
```

Schema: https://github.com/edgedb/edgedb-js/blob/a410b1630214db8dc2c19b71b94d26e6b348a917/qb/dbschema/default.esdl (with data inserted in test setup: https://github.com/edgedb/edgedb-js/blob/a410b1630214db8dc2c19b71b94d26e6b348a917/qb/test/setupTeardown.ts#L37)

msullivan:
This includes about half of #3259, which is being abandoned, in order
to get it to work with linkprops.

There are still some issues when the link is computed, but hopefully
this unsticks some of the query builder problems.

Fixes #3245.

### DIFF
--- a/edb/pgsql/compiler/relgen.py
+++ b/edb/pgsql/compiler/relgen.py
@@ -907,9 +907,10 @@ def process_set_as_path(
     if is_linkprop:
         backtrack_src = ir_source
         ctx.disable_semi_join.add(backtrack_src.path_id)
-        assert ir_source.rptr is not None
+
+        assert backtrack_src.rptr
         while backtrack_src.path_id.is_type_intersection_path():
-            backtrack_src = ir_source.rptr.source
+            backtrack_src = backtrack_src.rptr.source
             ctx.disable_semi_join.add(backtrack_src.path_id)
 
     semi_join = (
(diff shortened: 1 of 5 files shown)


edb/pgsql/compiler/relgen.py
```diff
@@ -1221,7 +1222,21 @@ def process_set_as_subquery(
             stmt.where_clause = astutils.extend_binop(
                 stmt.where_clause, cond_expr)
 
-    return _new_subquery_stmt_set_rvar(ir_set, stmt, ctx=ctx)
+    rvars = _new_subquery_stmt_set_rvar(ir_set, stmt, ctx=ctx)
+    # If the inner set also exposes a pointer path souce, we need to
```
elprans, line 1226:
```suggestion
    # If the inner set also exposes a pointer path source, we need to
```

elprans (approved):

PR #3277, opened 2021-12-14
Add digitalocean deployment guide

fmoor:


docs/guides/deployment/digitalocean.rst
```diff
@@ -0,0 +1,105 @@
...
+
+.. code-block:: bash
+
+   ssh root@$IP "
+       docker run \
+           --detach \
+           --restart always \
+           --name edgedb \
+           --publish 5656:5656 \
+           --volume /var/lib/edgedb/data:/var/lib/edgedb/data \
```
elprans, line 74: The data volume is not necessary when `SERVER_BACKEND_DSN` is used.

docs/guides/deployment/digitalocean.rst
```diff
@@ -0,0 +1,105 @@
...
+           --env EDGEDB_SERVER_PASSWORD=$PASSWORD \
+           --env EDGEDB_SERVER_TLS_CERT_MODE=generate_self_signed \
+           --env EDGEDB_SSERVER_BACKEND_DSN=$URI \
```
elprans, line 77: Typo

```suggestion
           --env EDGEDB_SERVER_BACKEND_DSN=$URI \
```

docs/guides/deployment/digitalocean.rst
```diff
@@ -0,0 +1,105 @@
...
+* DigitalOcean account
+* ``doctl`` CLI (`install <doclt-install_>`_)
+
+.. _doctl-install: https://docs.digitalocean.com/reference/doctl/how-to/install/
+
+
+Create a managed PostgreSQL instance
+====================================
+
+Note the URI that is printed in the output from this command. It will be used
```
elprans, line 23: Is there a way to obtain the URI with a CLI command?  Ideally, the guide should be completely copy-pastable to reduce space for user error.

docs/guides/deployment/digitalocean.rst
fmoor: What about persisting the generated TLS certificate?

docs/guides/deployment/digitalocean.rst
fmoor: Most `doctl` commands provide a way to get single values likes, but the `doctl database create` command does not have that functionality. The nicest way would to use the `--output json` option with `jq` but that adds another dependency. Other wise we could do some ugly text parsing with grep & cut, sed, awk etc. Not sure which is better.

There also does not seem to be a programmatic way to get the URI for a database is all you  know is the name. You need the id to use `doctl database get`.

docs/guides/deployment/digitalocean.rst
elprans: Ideally, it needs to be extracted and stored as a secret like we do for Fly.

fmoor: I don't think that digitalocean has a builtin secrets management solution. I couldn't find one. We would have to choose a secrets management solution like Vault or something and send the reader off to complete some other tutorial before continuing with ours.

fmoor: Or we could change the tutorial to use Kubernetes which has secrets management.

docs/guides/deployment/digitalocean.rst
elprans: `jq` isn't an unreasonable thing to rely on, it's widely available.

docs/guides/deployment/digitalocean.rst
elprans: Hm, I actually think we should change this to use our pre-built image instead.

docs/guides/deployment/digitalocean.rst
```diff
@@ -0,0 +1,180 @@
...
+Configure the backend postgres DSN. To simplify the initial deployment, let's
+instruct EdgeDB to run in insecure mode (with password authentication off and
+an autogenerated TLS certificate). We will secure the instance once things are
+up and running.
+
+.. code-block:: bash
+
+   $ ssh root@$IP "
+   cat << EOF > /etc/edgedb/env
+   EDGEDB_SERVER_BACKEND_DSN=${DSN}
```
fmoor, line 126: It looks like the docs generation requires a `\` at the end of a line for it to treat the next line as part of the same command.
image
This snippet can't be copied using the copy button when rendered to HTML. It only copies the first line :(

fmoor: cc @elprans

elprans: Rewrite it as

```
printf "..." | ssh root@$IP -T "cat > /etc/edgedb/env"
```

fmoor: I still can't get the string to fit on one line so it has the same problem.
image

I could do it with two commands but this seems strange in a guide.
image

elprans: Try

```
printf "EDGEDB_SERVER_BACKEND_DSN=${DSN}\n\
EDGEDB_SERVER_SECURITY=insecure_dev_mode\n" | ...
```

fmoor: There has to be a space before the `\` for it to count as a line continuation by the doc generation system but this one does work. The placement of `\n`s is a little strange but its not too bad.
image

elprans (approved):

PR #3287, opened 2021-12-16
Update GROUP syntax to match RFC 1009, implment it in the model

msullivan:
The RFC is at https://github.com/edgedb/rfcs/blob/master/text/1009-group.rst.

Update the parser, AST, codegen, and tracer to implement the new
syntax.

Implement GROUP in the toy model and write a few simple tests for it.


tests/test_edgeql_syntax.py
```diff
@@ -2680,13 +2661,61 @@ def test_edgeql_syntax_group_05(self):
...
+    def test_edgeql_syntax_group_08(self):
+        """
+        GROUP
+            User
+        USING
+            letter := (.name)[0]
+        BY {letter, age, ROLLUP(rank, status)};
+        """
+
+    def test_edgeql_syntax_group_09(self):
```
elprans, line 2712: Let's add some tests with nested grouping sets

elprans (approved):

PR #3288, opened 2021-12-16
Fix CI Cython/Rust extension cache mixing issue

fantix:
Refs #2978

### DIFF
--- a/.github/workflows.src/tests.inc.yml
+++ b/.github/workflows.src/tests.inc.yml
@@ -154,12 +154,12 @@
         if [[ "$CACHE_HIT" != "true" ]]; then
           rm -rf ${BUILD_LIB}
           mkdir -p build/rust_extensions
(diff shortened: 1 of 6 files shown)


.github/workflows.src/tests.inc.yml
```diff
@@ -180,8 +181,9 @@
           rm -rf ${BUILD_LIB}
           mkdir -p ./build/extensions
           rsync -aP ./build/extensions/ ${BUILD_LIB}/
-          python setup.py -v build_ext
+          BUILD_EXT_MODE=py-only python setup.py -v build_ext
           rsync -aP ${BUILD_LIB}/ ./build/extensions/
```
elprans, line 185: Let's use `-v` instead of `-P`, because the latter actually implies `--partial` (and we don't want that), and also, `--progress` output gets garbled in the logs.

elprans (approved):

PR #3292, opened 2021-12-17
[toy-model] Don't deduplicate grouping sets

msullivan:


edb/tools/toy_eval_model.py
```diff
@@ -679,12 +679,12 @@ def eval_Group(node: qlast.GroupQuery, ctx: EvalContext) -> Result:
         Tuple[Tuple[ByElement, ...], Tuple[Data, ...]],
         Tuple[Dict[ByElement, Data], List[Data]]
     ] = {}
-    # Rebuild the set tuple from all_keys to both deduplicate
+    # Rebuild the set tuples from all_keys to both deduplicate
```
elprans, line 682: Dedup comment now outdated?

elprans (approved):

msullivan:
... this doesn't actually fix the issue, it just breaks it differently

edb/tools/toy_eval_model.py
msullivan: It still deduplicates internally to the key, which is actually what the comment *originally* referred to

PR #3299, opened 2021-12-20
Exit with status 11 on longer-than-supported runstate dir path

elprans:
This needs to be a specific exit code for the CLI to be able to identify
the condition better and present a better error message.

While at it, change the default "invalid usage" exit code to `2`, which
is what seems to be commonly used to indicate usage errors.

### DIFF
--- a/edb/server/args.py
+++ b/edb/server/args.py
@@ -47,14 +47,14 @@ TLS_KEY_FILE_NAME = "edbprivkey.pem"
 logger = logging.getLogger('edb.server')
 
 
-class FatalConfigurationError(Exception):
+class InvalidUsageError(Exception):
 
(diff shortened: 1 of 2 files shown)


fantix (approved):

tailhook:
> 
> While at it, change the default "invalid usage" exit code to 64 (`EX_USAGE`).

Not sure about 64 as default. Rust's clap (and so edgedb) has reverted that choice (and made `2`, although it was `1` for years before):
https://github.com/clap-rs/clap/pull/1653#issuecomment-581104426

elprans:
> has reverted that choice

OK, I'll make it `2`, it doesn't really matter to me as long as it's _different_ from a general crash returning `1`.

PR #3308, opened 2021-12-22
Add backend capability option

fantix:
* Also overwrite detected capabilities with the ones from bootstrap.
* Also wrapped `feature_not_supported` pgerrors into `UnsupportedBackendFeatureError`

```
  --backend-capabilities [[~]SUPERUSER_ACCESS] [[~]CONFIGFILE_ACCESS] [[~]C_UTF8_LOCALE] [[~]CREATE_ROLE] [[~]CREATE_DATABASE]
                                  A space-separated set of backend
                                  capabilities, which are required to be
                                  present, or absent if prefixed with ~.
                                  EdgeDB will only start if the actual backend
                                  capabilities match the specified set.
                                  However if the backend was never
                                  bootstrapped, the capabilities prefixed with
                                  ~ will be *disabled permanently* in EdgeDB
                                  as if the backend never had them.
```

### DIFF
--- a/edb/server/args.py
+++ b/edb/server/args.py
@@ -34,6 +34,7 @@ from edb import buildmeta
 from edb.common import devmode
 from edb.common import enum
 from edb.schema import defines as schema_defines
+from edb.pgsql import params as pgsql_params
 
 from . import defines
 
@@ -132,6 +133,11 @@ class ServerConfig(NamedTuple):
 
     instance_name: Optional[str]
 
+    backend_capability_bound: Tuple[
+        List[pgsql_params.BackendCapabilities],
+        List[pgsql_params.BackendCapabilities],
+    ]
+
 
 class PathPath(click.Path):
     name = 'path'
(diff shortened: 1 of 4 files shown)


edb/server/args.py
```diff
@@ -160,6 +166,44 @@ def convert(self, value, param, ctx):
             self.fail(f"{value!r} is not a valid integer", param, ctx)
 
 
+class CapabilityBoundingSet(click.ParamType):
+    name = 'capability'
+
+    def __init__(self):
+        self.choices = list(
```
fantix, line 173: Ah, this is better to be a dict.

edb/server/args.py
```diff
@@ -509,6 +554,17 @@ def _writer(status: str) -> None:
         envvar="EDGEDB_SERVER_INSTANCE_NAME",
         type=str, default=None, hidden=True,
         help='Server instance name.'),
+    click.option(
+        '--backend-capability-bound',
+        envvar="EDGEDB_SERVER_BACKEND_CAPABILITY_BOUND",
```
elprans, line 559: I don't think this should be called a "BOUND", because, unlike systemd's `CapabilityBoundingSet`, we start from "all detected capabilities are enabled" and not from "all capabilities are disabled".  Let's simply call it "--backend-capabilities"

elprans (approved):

PR #3310, opened 2021-12-23
Add Heroku Postgres test

fantix:
Run tests on real Heroku Postgres weekly. Sample run: https://github.com/edgedb/edgedb/actions/runs/1616127940

Some tests are failing because Heroku Postgres has `COLLATE="en_US.UTF-8"` while EdgeDB assumes `COLLATE="C"`. These tests are marked as "xfail" for now until EdgeDB supports single-db mode on backend that doesn't have `COLLATE="C"`. Before that, string ordering would follow the local collation inconsistently on backends like Heroku Postgres.

### DIFF
--- a/.github/heroku-postgres/main.tf
+++ b/.github/heroku-postgres/main.tf
@@ -1,0 +1,29 @@
+terraform {
+  required_providers {
+    heroku = {
+      source = "heroku/heroku"
+      version = "~> 4.0"
+    }
+  }
+}
+
+//resource "heroku_app" "default" {
+//  name = "edgedb-heroku-ci"
+//  region = "us"
+//
+//  organization {
+//    name = "edgedb"
+//  }
+//}
+
+resource "heroku_addon" "database" {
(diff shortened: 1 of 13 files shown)


.github/heroku-postgres/main.tf
```diff
@@ -0,0 +1,29 @@
+terraform {
+  required_providers {
+    heroku = {
+      source = "heroku/heroku"
+      version = "~> 4.0"
+    }
+  }
+}
+
+//resource "heroku_app" "default" {
```
elprans, line 10: Remove commented-out code?

elprans (approved):

.github/heroku-postgres/main.tf
fantix: Oh actually this was commented because the Heroku account running tests doesn't have permission to delete an App. If an idle Heroku App is generating unnecessary bills, we could promote the account and uncomment to drop the App.

PR #3315, opened 2022-01-05
Pull assert_data_shape out of testbase.server and use it for model tests

msullivan:


elprans (approved):

1st1:
I'm not sure I like the name 'bag' here, imo `tb.unordered` would read better.

PR #3316, opened 2022-01-05
Server.create_task() will fail if called unexpectedly

### LINKED ISSUE #3218: Warning during bootstrap

fantix:
Hence, all `Server.create_task()` calls are guarded with an additional
`if Server._accept_new_tasks` check.

This is simply Option 2 in #3112.

Closes #3112, closes #3220, fixes #3218

### DIFF
--- a/edb/server/server.py
+++ b/edb/server/server.py
@@ -557,7 +557,7 @@ class Server(ha_base.ClusterProtocol):
 
     def schedule_reported_config_if_needed(self, setting_name):
         setting = self._config_settings[setting_name]
-        if setting.report:
+        if setting.report and self._accept_new_tasks:
             self.create_task(
                 self.load_reported_config(), interruptable=True)
 
@@ -1106,6 +1106,9 @@ class Server(ha_base.ClusterProtocol):
             raise
 
     def _on_remote_ddl(self, dbname):
+        if not self._accept_new_tasks:
+            return
+
         # Triggered by a postgres notification event 'schema-changes'
         # on the __edgedb_sysevent__ channel
         async def task():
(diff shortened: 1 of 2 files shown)


edb/server/server.py
```diff
@@ -557,7 +557,7 @@ async def reload_sys_config(self):
 
     def schedule_reported_config_if_needed(self, setting_name):
         setting = self._config_settings[setting_name]
-        if setting.report:
+        if setting.report and self._accept_new_tasks:
             self.create_task(
```
1st1, line 561: Why can't `self.create_task` return early when `not self._accept_new_tasks`?

Also, `create_task(foo)` could wrap the foo coroutine into this:

```
async wrapper():
    if self._accept_new_tasks:
        await foo
```

With all that, we probably don't need to litter the code with many checks for `self._accept_new_tasks`, no?

fantix: Unfortunately neither of the suggested methods work - if a coroutine is created but not consumed in a Task, it'll raise a warning.

1st1: Oh, I see. :/

1st1 (approved):

PR #3320, opened 2022-01-07
Make it possible to prohibit bare DDL

elprans:
RFC 1000 prescribes a configuration parameter that can be used to disable
all DDL outside of migrations.  This commit adds
`allow_bare_ddl` to do just that.  `edgedb migrate` can automatically set
it on a database to prevent accidental migration state skew.


edb/lib/_testmode.edgeql
```diff
@@ -108,6 +111,9 @@ ALTER TYPE cfg::AbstractConfig {
         SET default := <std::duration>'0 seconds';
     };
 
+    CREATE PROPERTY enumprop -> cfg::TestEnum {
+        SET default := <cfg::TestEnum>'One';
```
1st1, line 115: this should also work, right?

```suggestion
        SET default := cfg::TestEnum.One;
```

edb/pgsql/metaschema.py
```diff
@@ -4551,36 +4554,29 @@ def _generate_schema_ver_views(schema: s_schema.Schema) -> List[dbops.View]:
...
-        def _cast(val: Any) -> str:
-            return f'{q(*func_name)}({val})'
+    cast_sql, _ = compiler.compile_ir_to_sql(
+        cast_ir,
+        use_named_params=True,
+        singleton_mode=True,
+    )
 
-    return _cast
+    return lambda val: cast_sql.replace('__replaceme__', val)
```
1st1, line 4579: ouch, is this really necessary?

tests/test_server_config.py
```diff
@@ -972,11 +972,15 @@ async def test_server_proto_configure_describe_system_config(self):
             conf4 = "CONFIGURE INSTANCE SET memprop := <cfg::memory>'100MiB';"
             await self.con.execute(conf4)
 
+            conf5 = "CONFIGURE INSTANCE SET enumprop := <cfg::TestEnum>'Two';"
```
1st1, line 975: Can you also add a test that `cfg::TestEnum.Two` works?

1st1 (approved):

edb/pgsql/metaschema.py
elprans: This is the simplest and reliable way.  SQL AST is immutable, so it isn't possible to do this easily with an AST visitor.

edb/lib/_testmode.edgeql
elprans: Right, this works now too.

tests/test_server_config.py
1st1: Unresolving this thread -- can you add this test in this commit?

elprans: No, because we cannot make `DESCRIBE INSTANCE CONFIG` return a non-casting version.  There is a dot-notation test elsewhere.

PR #3327, opened 2022-01-08
Fix performance of queries containing references to multi-links from DML

elprans:
When compiling link references that have a DML statement as a source,
we normally produce an "overlay" UNION containing the existing link
relation as well as the effects of the mutation.  Unfortunately,
Postgres seems to cope badly with JOINs containing UNIONs which are
composed of a mix of RelRange and RangeSubselect, producing horrible
HashJoin plans scanning the entire UNION.  Luckily, there's a
straightforward workaround: push the JOIN condition into UNION operands
and simply CROSS JOIN the UNION laterally.  While this is theoretically
beneficial for all kinds of UNIONs, we restrict this optimization to overlay
UNIONs only to limit the possibility of breakage as not all  UNIONs are
guaranteed to have correct path namespace and translation maps set up.

### DIFF
--- a/edb/pgsql/compiler/relctx.py
+++ b/edb/pgsql/compiler/relctx.py
@@ -412,6 +412,7 @@ def new_primitive_rvar(
     ir_set: irast.Set,
     *,
     path_id: irast.PathId,
+    lateral: bool,
     ctx: context.CompilerContextLevel,
 ) -> pgast.PathRangeVar:
     if not ir_set.path_id.is_objtype_path():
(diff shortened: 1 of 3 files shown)


edb/pgsql/compiler/relctx.py
```diff
@@ -1157,11 +1199,54 @@ def rel_join(
...
+    query: pgast.SelectStmt,
+    right_rvar: pgast.RangeSubselect,
+    *,
+    ctx: context.CompilerContextLevel,
+) -> None:
+    def _inject_filter(component: pgast.Query) -> None:
+        condition = None
+
+        for path_id in right_rvar.query.path_scope:
+            lref = maybe_get_path_var(
```
msullivan, line 1212: Do you think it's worth trying to unify `_inject_filter` and the main logic in `_plain_join`?

msullivan (approved):

edb/pgsql/compiler/relctx.py
elprans: They are subtly different, I don't think unifying them would help readability.

PR #3329, opened 2022-01-10
Add `eql:index-boost` directive

jaclarke:
The long lost sibling of this PR: https://github.com/edgedb/edgedb.com/pull/196

### DIFF
--- a/edb/tools/docs/eql.py
+++ b/edb/tools/docs/eql.py
@@ -609,6 +609,25 @@ class EQLReactElement(d_rst.Directive):
         return [node]
 
 
+class EQLIndexBoost(d_rst.Directive):
+
+    has_content = False
+    optional_arguments = 0
+    required_arguments = 1
+
+    def run(self):
+        try:
+            boost = float(self.arguments[0])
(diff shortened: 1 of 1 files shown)


1st1:
What specific pages will we boost with this? Maybe we should tweak our ranking algo instead?

jaclarke:
Originally it was proposed to boost the datamodel pages and de-emphasize the ddl, and changelog pages: https://edgedb.slack.com/archives/CNECHA9EW/p1615416583244700?thread_ts=1615391108.240000&cid=CNECHA9EW.

1st1:
Will this directive be applied to an entire page?

colinhacks:
> Will this directive be applied to an entire page?
As I understand it the boost is applied to all indexed items on a page. Also if it's defined on an index page, the boost is applied to the entire subtree, which is cool. It seems deeper `boost-index` values override shallower ones: https://github.com/edgedb/edgedb.com/blob/d9ddc707c0ca0560c86b99c2ef678ca003351eaa/dataBuild/docs.ts#L555

PR #3331, opened 2022-01-10
Always include the definition context namespace in computable contexts

### LINKED ISSUE #3323: Another bad for-loop/computable/with interaction
```
with props := (                                                           
  for h in User union (                                                   
    select h {namelen := len(h.name)}                                     
  )                                                                       
)                                                                         
select props {                                                            
  name,                                                                   
  namelen # <- adding this property causes cartesian product              
};
```

produces a cartesian product. Defining the computable with just `.name` instead of `h.name` works.

I thought it was related to https://github.com/edgedb/edgedb/pull/2947 but who knows.

reported by @jaclarke

msullivan:
We need to include the *original* source namespace in our ctx
namespace when compiling computables. The current mechanism of trying
to look up in view_sets or failing that using the source namespace
from the computable use, but this was failing to find it in some cases
with FOR.

Fix this by instead directly pulling in the namespace from qlctx. The
inclusion of qlctx's namespace nicely allows us to ditch so later
logic as well.

Additionally we need to merge the namespace into *both* sides in
get_view_map_remapping, to handle cases like referencing a `FOR`
variable where the current ns doesn't get merged in.

Fixes #3323.

### DIFF
--- a/tests/test_edgeql_scope.py
+++ b/tests/test_edgeql_scope.py
@@ -3521,7 +3521,7 @@ class TestEdgeQLScope(tb.QueryTestCase):
             )
             select props {
               name,
-              namelen # <- adding this property causes cartesian product
+              namelen
             };
             ''',
             tb.bag([
--- a/tests/test_edgeql_scope.py
+++ b/tests/test_edgeql_scope.py
@@ -3063,7 +3063,7 @@ class TestEdgeQLScope(tb.QueryTestCase):
             ]
         )
 
-    async def test_edgeql_scope_ref_outer_02(self):
+    async def test_edgeql_scope_ref_outer_02a(self):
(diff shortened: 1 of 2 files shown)


tests/test_edgeql_scope.py
```diff
@@ -3491,3 +3510,45 @@ async def test_edgeql_select_outer_rebind_04(self):
...
+        await self.assert_query_result(
+            r'''
+            with props := (
+              for h in User union (
+                select h {namelen := len(h.name)}
+              )
+            )
+            select props {
+              name,
+              namelen # <- adding this property causes cartesian product
```
elprans, line 3524: Drop this comment.

elprans (approved):

PR #3332, opened 2022-01-10
Mark `test_edgeql_expr_valid_collection_15` as xfail on Postgres 14 only

elprans:
Seems like this is a regression in 14.

### DIFF
--- a/tests/test_edgeql_expressions.py
+++ b/tests/test_edgeql_expressions.py
@@ -2170,8 +2170,10 @@ class TestExpressions(tb.QueryTestCase):
         )
 
     @test.xfail(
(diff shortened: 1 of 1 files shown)


msullivan (approved):

fantix:
It's also [failing](https://github.com/edgedb/edgedb/runs/4770366476?check_suite_focus=true#step:19:5089) on Postgres 12?

elprans:
> It's also [failing](https://github.com/edgedb/edgedb/runs/4770366476?check_suite_focus=true#step:19:5089) on Postgres 12?

Yikes.  That's a different (and more unpleasant) failure mode.

PR #3348, opened 2022-01-19
Convert the docs to use lower-case keywords consistently.

vpetrovykh:
The client library examples are left as is. The queries appear inside
strings and so upper-case keywords may be desirable in that instance.

https://github.com/edgedb/easy-edgedb/pull/49 should be merged so that building the entire site works, unless you toggle easy-edgedb off from the build.


vpetrovykh:
I've added a few more commits that address some other docs issues.

colinhacks (approved):

PR #3355, opened 2022-01-20
Update test client to use the latest edgedb-python

fantix:
Depends on edgedb/edgedb-python#286


msullivan (commented):
What is the story for why `testbase/connection.py` is growing all of this machinery? Is it because the test suite depended on old details of the internals of edgedb-python and now that those have changed we need to reimplement them on the testbase side?

fantix:
Yes, also because I tried not to inherit from the “Connection” class from edgedb-python and re-implement the interfaces in testbase.

elprans (approved):

PR #3362, opened 2022-01-24
Disable unrendered unicode bidirectional override characters

tailhook:
See CVE-2021-42574

Rust has a good explanation of what's the problem: https://blog.rust-lang.org/2021/11/01/cve-2021-42574.html


1st1:
lgtm; @elprans I think we should make this part of rc5

elprans (approved):

Issue #3347, opened 2022-01-19
InternalServerError: object type <QualName default::default|XXX@view~1> already exists


metmirr:
- EdgeDB Version: 1.0-rc.4+1850eff
- OS Version: MacOS 12.1

Steps to Reproduce:

1. `edgedb project init` ✅ 
2. `edgedb instance create sample` ✅ 
3. `edgedb -I sample migration create` ✅ 
4. `edgedb -I sample migrate` ✅ 
5. Go to graphiql playground

Error shown in the graphiql:
```json
{
  "errors": [
    {
      "message": "TypeError: Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but 'Filterdefault|ProcessedEventsLog@view~1' does not.\n\nNames must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but 'default|ProcessedEventsLog@view~1' does not.\n\nNames must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but 'default|ProcessedEventsLog@view~1' does not.\n\nNames must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but 'Orderdefault|ProcessedEventsLog@view~1' does not.\n\nNames must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but 'delete_default|ProcessedEventsLog@view~1' does not."
    }
  ]
}
``` 

Note: There are no errors during steps.

Schema:
```
using extension graphql;

module default {
  type ProcessedEventsLog {
    required property event_id -> bigint;
    required property indexer_head -> int64;
    required property chain_head -> int64;
    required property last_scanned_block -> int64;
    required property updated_at -> datetime;
  }

  type Transfer {
    required property from -> str;
    required property to -> str;
    required property amount -> float64;
    required property happened_in_block -> int64;
  }

  alias ProcessorStatus := (
    select ProcessedEventsLog {
      indexer_head,
      last_scanned_block,
      in_sync:= .last_scanned_block >= .indexer_head
    }
    order by .last_scanned_block DESC
    limit 1
  )
}
```

PR #3367, opened 2022-01-25
Reflect more complex aliases.

vpetrovykh:
Aliases using objects, filters and other clauses should be reflected
into GraphQL.

Issue #3347

### DIFF
--- a/edb/graphql/types.py
+++ b/edb/graphql/types.py
@@ -1258,13 +1258,6 @@ class GQLCoreSchema:
                 # The aliased types actually only reflect as an object
                 # type, but the rest of the processing is identical to
                 # interfaces.
-                if '~' in gql_name:
-                    # These are internal aliases, but they define the
-                    # actual shapes we need to reflect.
-                    gql_name = '_edb' + gql_name.replace('|', '_') \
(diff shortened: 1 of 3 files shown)


vpetrovykh:
This should fix a number of issues with alias reflection into GraphQL, but more elaborate aliases may still have problems.

edb/graphql/types.py
```diff
@@ -1248,6 +1248,13 @@ def _define_types(self) -> None:
                 # The aliased types actually only reflect as an object
                 # type, but the rest of the processing is identical to
                 # interfaces.
+                if '~' in gql_name:
+                    # These are internal aliases, but they define the
+                    # actual shapes we need to reflect.
+                    gql_name = '_edb' + gql_name.replace('|', '_') \
```
elprans, line 1254: Ugh, this is unfortunate.  I have a branch where nested alias naming is fixed properly.  If we want to fix this here now, ideally we'd need to synchronize the mangling schema.

msullivan:
The issue was fixed in #3751 by fixing the names actually assigned to these nested aliases

elprans (approved):

PR #3368, opened 2022-01-25
Deeply nested aliases have problems in GraphQL

vpetrovykh:
wip


vpetrovykh:
This is intended to be on top of #3367

PR #3370, opened 2022-01-25
Update quickstart docs to use correct query

joeriddles:
Update `docs > guides > quickstart.rst > 5. Run some queries`.

Fix the filter to use `ILIKE`, like the paragraph above the code sample explains.


colinhacks:
Good catch! I decided to switch from using `ilike` to simply filtering by the `Movie.id` property, and I did a terrible job of switching. Closing this in favor of https://github.com/edgedb/edgedb/pull/3372. Thanks!

PR #3373, opened 2022-01-25
Optimize optional compilation of simple paths

msullivan:
Currently we put an optional wrapper around optional arguments in all
cases, include ones such as
`SELECT Issue { te :=  .time_estimate ?? -1 }` in which it would be sound
to simply rely on time_estimate being stored as NULL in the database
and omit the optional wrapper.

We can omit the optional wrapper when the argument is a path in which
the LHS is visible and the pointer is a single non-computed property.

### DIFF
--- a/edb/pgsql/compiler/relgen.py
+++ b/edb/pgsql/compiler/relgen.py
@@ -504,7 +504,7 @@ def can_omit_optional_wrapper(
     is visible and bar is a single non-computed property, which we know
     will be stored as NULL in the database.
     """
-    if (
+    return (
         ir_set.expr is None
         and not ir_set.path_id.is_objtype_path()
         and ir_set.rptr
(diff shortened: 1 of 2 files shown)


edb/pgsql/compiler/relgen.py
```diff
@@ -492,6 +492,32 @@ def set_as_subquery(
...
+    """Determine whether it is safe to omit the optional wrapper.
+
+    Doing so is safe when the expression is guarenteed to result in
+    a NULL and not an empty set.
+
+    Currently the only such case implement is a path `foo.bar` where foo
+    is visible and bar is a single non-computed property, which we know
+    will be stored as NULL in the database.
+    """
+    if (
```
elprans, line 507: Why not just `return` the condition?

elprans (approved):

edb/pgsql/compiler/relgen.py
msullivan: I had in my head that there would be a chain of conditions... but then there was just one :P

msullivan:
@elprans I also added another case for trivial selects, which I think is also pretty low-risk. (It's needed for the case that prompted this, which is group by using clauses)

PR #3374, opened 2022-01-25
Fix card inference for OFFSET/LIMIT clauses

msullivan:
OFFSET and LIMIT can reduce cardinality


elprans (approved):
Please cherry-pick to 1.x

Issue #211, opened 2018-06-29
Add support for negative `OFFSET`


vpetrovykh:
It would be nice to support negative `OFFSET` values. The interpretation would be similar to Python, where a negative index indicates counting from the end.

In terms of implementation negative `OFFSET` is equivalent to an order reversal, `LIMIT` and another order reversal:

```
SELECT Foo
ORDER BY Foo.bar ASC NULLS LAST
OFFSET -3;

# equivalent to
SELECT (
    SELECT Foo
    ORDER BY Foo.bar DESC NULLS FIRST
    LIMIT 3
  )
ORDER BY Foo.bar ASC NULLS LAST;
```

vpetrovykh:
This feature would also be used in GraphQL to implement `last` argument for pagination.

msullivan:
How would we handle this in the general case, where the argument to `OFFSET` is dynamic? That is probably actually the typical case, since the constant will be replaced with a query parameter.

I guess the question is less "how", since I think it is definitely doable (maybe: order by row_number()*sign(offset), then offset by abs(offset), then order by row_number), and more "is it worth doing that"?

elprans:
`OFFSET`/`LIMIT` without `ORDER BY` is almost always meaningless anyway, so one possibility is that we require `ORDER BY` for negatives?

msullivan:
I don't think that really addresses my concern, which is that the argument to OFFSET is dynamic and we don't know whether it is negative or not, and so need to generate code that handles both cases.

elprans:
> we don't know whether it is negative or not, and so need to generate code that handles both cases.

Ah, that's a good point.  Perhaps negative slicing should be explicit syntactically?  E.g `LIMIT NEGATIVE -10` or something?

PR #3375, opened 2022-01-25
Handle negative OFFSET and LIMIT

msullivan:
Nothing important here, but might as well not produce an ISE.


edb/api/errors.txt
```diff
@@ -112,6 +112,8 @@
 0x_05_01_00_00   InvalidValueError
 0x_05_01_00_01   DivisionByZeroError
 0x_05_01_00_02   NumericOutOfRangeError
+0x_05_01_00_03   InvalidLimitError
```
msullivan, line 115: Is it worth adding new error types for this? We could just use InvalidValueError directly? (Does adding error types require updating the bindings?)

elprans: Probably not, because we are mulling possibly adding support for negative `LIMIT`/`OFFSET` (see #211).

msullivan: Alright, so just make it an `InvalidValueError`?

elprans: Yeah, that seems reasonable.

elprans (approved):

msullivan:
Going to cherry-pick this since it is wildly low risk

PR #3387, opened 2022-01-30
Weed out unnecessary reserved keywords

elprans:
Currently we have a number of keywords that are reserved for no technical
reason.  They are all part of utility commands and will never legally
appear in places where an identifier is expected.  (Except
`empty`, which is part of `ORDER BY` syntax, but doesn't need to be
reserved either).

More specifically, this commit downgrades: `abort`, `declare`, `empty`,
`populate`, `release`, `reset`, `start`.

Some of those are interesting type/property names so there is value in
unreserving these.


1st1 (approved):
backport to 1.0?

PR #3388, opened 2022-01-30
Unreserve the `ORDER` keyword

elprans:
`ORDER` is only used in combination with `BY` in the `ORDER BY` clause,
so teach the tokenizer to fuse the two keywords into a single token.
This allows unreserving `ORDER`, which is a massive win because "Order"
is a very useful name for a type.


1st1 (approved):
cherry-pick for 1.0?

msullivan (approved):

PR #3391, opened 2022-01-31
Add an assertion that materialized references are not duplicated

msullivan:
This turns some cases that are currently miscompiled into ones
that trigger an ISE, which is an improvement.


elprans (approved):

msullivan:
@elprans I'd like this in 1.0. Thoughts?

elprans:
An ISE is always better than a miscompile, so if there's low risk of false assertion failures, go ahead and cherry-pick this.

PR #3397, opened 2022-02-03
Reduce the number of calls to delta_schema in the migration happy path

msullivan:
Get rid of all calls to schemas_are_equal except to check if a migration
is complete once we already think it should be. Also, distinguish between
the "we have nothing proposed" case and the "we have consumed all proposed
advice" case, which saves one schema diff at the end.

This brings the time to migrate to constraints.esdl down on my machine
from 78s to 12s.

### DIFF
--- a/edb/server/compiler/compiler.py
+++ b/edb/server/compiler/compiler.py
@@ -810,7 +810,6 @@ class Compiler:
             )
 
             context = self._new_delta_context(ctx)
-            orig_schema = schema
             schema = delta.apply(schema, context=context)
 
             if mstate.last_proposed:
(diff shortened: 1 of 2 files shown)


edb/server/compiler/compiler.py
```diff
@@ -1150,15 +1149,18 @@ def _compile_ql_migration(
...
+                        proposed_desc = None
 
                 desc = json.dumps({
                     'parent': (
                         str(mstate.parent_migration.get_name(schema))
                         if mstate.parent_migration is not None
                         else 'initial'
                     ),
-                    'complete': s_ddl.schemas_are_equal(
+                    'complete': not proposed_desc and s_ddl.schemas_are_equal(
```
elprans, line 1163:
```suggestion
                    'complete': proposed_desc is None and s_ddl.schemas_are_equal(
```

elprans (approved):

Issue #3385, opened 2022-01-29
UDF invalid reference error


elprans:
### Discussed in https://github.com/edgedb/edgedb/discussions/3383

<div type='discussions-op-text'>

<sup>Originally posted by **rdlaitila** January 29, 2022</sup>
Wondering if I'm hitting a bug or unsupported EdgeQL language support for UDFs. For the given UDF:

```
module default {
	function test(foo: str, bar: str) -> str {
		using (
			with
				f := foo,
				b := bar,
				r := f ++ b,
			select r
		)
	}
}
```

When attempting to create the migrations via `edgedb migrations create` I get the following error:

```
edgedb error: InvalidReferenceError: schema item 'std::f' does not exist
```

Using edgedb-cli: 1.0.0-rc.6+4b9d831
Using edgedb-instance: 1.0-rc.4+1850eff

UDF documentation appears sparse for the early releases, wondering if this is just user error. Thanks!</div>

PR #3398, opened 2022-02-03
Fix tracing of WITH blocks in SDL

elprans:
Currently, the `WITH` block tracer forgets to apply the effects of prior
entries to subsequent entries, so

    with
      f := foo,
      b := bar,
      r := f ++ b,
    select r

fails with `InvalidReferenceError: schema item 'std::f' does not exist`
when processing the `r` alias.

Fixes: #3385


vpetrovykh (approved):
LGTM

PR #3401, opened 2022-02-03
Reject setting a non-empty type as abstract

### LINKED ISSUE #3399: Setting a type as abstract preserves existing data
```
edgedb> create type Foo;
OK: CREATE TYPE
edgedb> insert Foo;
{default::Foo {id: 6d4cab68-8528-11ec-b417-ab253b541761}}
edgedb> alter type Foo { set abstract; };
OK: ALTER TYPE
edgedb> select Foo;
{default::Foo {id: 6d4cab68-8528-11ec-b417-ab253b541761}}
```

Since it isn't possible to insert into an abstract type, I would expect that the abstract type shouldn't be able to contain data either?

It seems that on the pgsql schema side we don't have any special handling of abstract types at all, and they get tables and everything.

msullivan:
Fixes #3399.

### DIFF
--- a/edb/pgsql/delta.py
+++ b/edb/pgsql/delta.py
@@ -2952,6 +2952,7 @@ class AlterObjectType(ObjectTypeMetaCommand,
         schema: s_schema.Schema,
         context: sd.CommandContext,
     ) -> s_schema.Schema:
+        orig_schema = schema
         schema = super().apply(schema, context=context)
         objtype = self.scls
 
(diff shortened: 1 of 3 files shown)


edb/pgsql/delta.py
```diff
@@ -2969,6 +2972,39 @@ def apply(
...
+        ))
+
+        vn = self.scls.get_verbosename(schema)
+        check_qry = textwrap.dedent(f'''\
+            SELECT
+                edgedb.raise(
+                    NULL::text,
+                    'cardinality_violation',
+                    msg => {common.quote_literal(
+                            f"may not make non-empty {vn} abstract")},
```
elprans, line 2999:
```suggestion
                            f"cannot make non-empty {vn} abstract")},
```

elprans (approved):

PR #3402, opened 2022-02-04
Produce a real error message when an annotation value is not a string

msullivan:


elprans (approved):

edb/schema/annos.py
```diff
@@ -290,8 +292,10 @@ def _cmd_tree_from_ast(
             astnode.value, schema=schema)
 
         if not isinstance(value, str):
-            raise ValueError(
-                f'unexpected value type in annotation: {value!r}')
+            raise errors.InvalidValueError(
+                'annotation values must be strings',
```
1st1, line 296:
```suggestion
                'annotation values must be strings, got ${typename}',
```

edb/schema/annos.py
```diff
@@ -366,8 +370,10 @@ def _cmd_tree_from_ast(
                 astnode.value, schema=schema)
 
             if not isinstance(value, str):
-                raise ValueError(
-                    f'unexpected value type in AnnotationValue: {value!r}')
+                raise errors.InvalidValueError(
+                    'annotation values must be strings',
```
1st1, line 374: ditto

edb/schema/annos.py
```diff
@@ -282,18 +282,23 @@ def _cmd_tree_from_ast(
...
         assert isinstance(cmd, CreateAnnotationValue)
         annoname = sn.shortname_from_fullname(cmd.classname)
 
-        value = qlcompiler.evaluate_ast_to_python_val(
+        value, ir = qlcompiler.evaluate_ast_to_python_val_and_ir(
             astnode.value, schema=schema)
 
-        if not isinstance(value, str):
+        if (
+            not isinstance(ir.stype, s_scalar.ScalarType)
```
elprans, line 296: Use `ir.stype.is_scalar()` and then you don't need an import?

msullivan: Actually we don't even need to check that it's a scalar, do we?

elprans: We don't.

PR #3408, opened 2022-02-07
add `SHOULD_RETRY` tag to `TransactionConflictError` inheritors

nsidnev:
This is a small change, but I think it might be useful. I needed it for my driver so that I didn't have to invent a tag inheritance mechanism when generating errors.

`ClientConnectionFailedTemporarilyError` and its inheritors have tag duplication for this case:
https://github.com/edgedb/edgedb/blob/master/edb/api/errors.txt#L165-L167

Similar to the `TransactionConflictError` exception and its inheritors in the python driver:
https://github.com/edgedb/edgedb-python/blob/master/edgedb/errors/__init__.py#L375-L387

### DIFF
--- a/edb/api/errors.txt
+++ b/edb/api/errors.txt
@@ -120,8 +120,8 @@
 
 0x_05_03_00_00   TransactionError
 0x_05_03_01_00   TransactionConflictError  #SHOULD_RETRY
-0x_05_03_01_01   TransactionSerializationError
-0x_05_03_01_02   TransactionDeadlockError
+0x_05_03_01_01   TransactionSerializationError  #SHOULD_RETRY
+0x_05_03_01_02   TransactionDeadlockError  #SHOULD_RETRY
 
 
 ####
--- a/edb/tools/gen_errors.py
+++ b/edb/tools/gen_errors.py
@@ -181,6 +181,9 @@ class ErrorsTree:
 
             self.add(desc)
 
+    def get_errors(self):
+        return self._tree
+
     def get_parent(self, code):
         b1, b2, b3, b4 = code
 
--- a/tests/test_api_errors.py
+++ b/tests/test_api_errors.py
@@ -17,9 +17,12 @@
 #
 
 
+import pathlib
 import unittest
 
+import edb
 from edb import errors
+from edb.tools import gen_errors
 
 
 class TestErrorsClasses(unittest.TestCase):
@@ -59,3 +62,20 @@ class TestErrorsClasses(unittest.TestCase):
         # intended for client libraries.
 
         self.assertFalse(hasattr(errors, 'ClientError'))
+
+
+class TestErrorsTags(unittest.TestCase):
+    errors_path = pathlib.Path(edb.__path__[0]) / 'api' / 'errors.txt'
+
+    def test_api_errors_tags_01(self):
+        tree = gen_errors.ErrorsTree()
+        tree.load(self.errors_path)
+
+        for (code, desc) in tree.get_errors().items():
+            parent = tree.get_parent(code)
+            if parent is None:
+                continue
+
+            self.assertTrue(
+                all(tag in desc.tags for tag in parent.tags),
+                f"{desc.name} error doesn't inherit all {parent.name} tags")


elprans (approved):

1st1:
@fantix @jaclarke @fmoor @tailhook take a look at this PR. When we merge it we'll need to update the clients to pick up this change.

tailhook:
> This is a small change, but I think it might be useful. I needed it for my driver so that I didn't have to invent a tag inheritance mechanism when generating errors.
> 

Technically this is not a change. And if we want to have explicit inherited tags we have to make a check/test for that.

> `ClientConnectionFailedTemporarilyError` and its inheritors have tag duplication for this case: https://github.com/edgedb/edgedb/blob/master/edb/api/errors.txt#L165-L167
> 

These are not inherited, so it's not the same. We note inheritance by matching prefix. Superclass is one that has `00` in trailing bytes. Note `https://twitter.com/nickschifrin/status/1492203844155150339` has `01_01` at the end while others have `02_00`, and `03_00`.

> Similar to the `TransactionConflictError` exception and its inheritors in the python driver: https://github.com/edgedb/edgedb-python/blob/master/edgedb/errors/__init__.py#L375-L387

Yes, in python driver tags are inherited by code generator. I think we should be doing that everywhere, unless we have some bug.

----

So either we enforce this at the source level, and in this case this PR needs a test for that.

Or we keep it as is and propagate tags in code generation (we can recheck all the bindings if you've seen some of them miss this functionality)

tailhook (changes requested):
> This is a small change, but I think it might be useful. I needed it for my driver so that I didn't have to invent a tag inheritance mechanism when generating errors.
> 

Technically this is not a change. And if we want to have explicit inherited tags we have to make a check/test for that.

> `ClientConnectionFailedTemporarilyError` and its inheritors have tag duplication for this case: https://github.com/edgedb/edgedb/blob/master/edb/api/errors.txt#L165-L167
> 

These are not inherited, so it's not the same. We note inheritance by matching prefix. Superclass is one that has `00` in trailing bytes. Note `https://twitter.com/nickschifrin/status/1492203844155150339` has `01_01` at the end while others have `02_00`, and `03_00`.

> Similar to the `TransactionConflictError` exception and its inheritors in the python driver: https://github.com/edgedb/edgedb-python/blob/master/edgedb/errors/__init__.py#L375-L387

Yes, in python driver tags are inherited by code generator. I think we should be doing that everywhere, unless we have some bug.

----

So either we enforce this at the source level, and in this case this PR needs a test for that.

Or we keep it as is and propagate tags in code generation (we can recheck all the bindings if you've seen some of them miss this functionality)

nsidnev:
> We note inheritance by matching prefix.

Oh cool, I missed that correlation, thanks!

> So either we enforce this at the source level, and in this case this PR needs a test for that.

No problem, I can add it if this PR is decided to merge. But I'm not sure where to add tests in this repository. Wouldn't it be better to instead add appropriate tests to the python driver (and others) directly affected by this change?

> (we can recheck all the bindings if you've seen some of them miss this functionality)

Sorry, I am not very familiar with the implementation of other drivers.

fmoor:
Making the attributes explicit might make implementing errors in arbitrary languages a little simpler. Although, we probably should put a note in `errors.txt` about the error code hierarchy and how we expect error checking in clients to work. Namely that if a client checks for a more general error and EdgeDB returns a more specific error that Inherits from the general error that the client is checking for the check in the client should match. Or another way, checking EdgeDB errors should not be a simple error code comparison. 

> I'm not sure where to add tests in this repository

I think this is the place https://github.com/edgedb/edgedb/blob/master/tests/test_api_errors.py

tailhook (approved):
The code looks good to me so I'm removing my "request changes" request.

But I would say that I'm still on this side:
> Or we keep it as is and propagate tags in code generation

This is mostly intuitive (i.e. how languages with inheritance usually work anyway), so I didn't come up with a good argument for the other side. So I leave it to @elprans to decide.

1st1:
I also think that the tools should propagate the flags to child types.

Although when we introduce error types tools will become much simpler

tailhook:
> Ok, I don't really mind leaving it as it is. I'll also implement tags inheritance in elixir driver when compiling the error module.
> 
> But can we add an explanation of the inheritance mechanism to the binary driver documentation? Perhaps other driver developers may have the same misunderstanding about that.

Sure, better docs are always encouraged. If you see a place in the docs where this information fits, please make a PR. If you're need help with that, let's continue discussing it here.

PR #3462, opened 2022-02-12
Fix typo in abstract type name in docs

amaster507:
The extended abstracted type was used as `HasEmail` but was delcared as `Email`


elprans (approved):
Good catch!

1st1 (approved):

elprans:
Thanks!

PR #3482, opened 2022-02-14
Bump bundled Postgres to 14.2

elprans:
The Memoize and record hashing regressions were fixed in 14.2, so we can
upgrade now.


fantix (approved):
:+1:

Issue #3454, opened 2022-02-11
Error on initial deployment to Google Cloud


bondo:
- EdgeDB Version: 1.0+08d3e21
- OS Version: Google cloud node version 1.21.6-gke.1500, local OS Ubuntu 20.04.3

Steps to Reproduce:

1. Follow the guide https://www.edgedb.com/docs/guides/deployment/gcp
2. Replace the region with europe-north1
3. In the "Deploy EdgeDB" section the pods never get ready

The logs says:
```
$ kubectl logs deployment/edgedb --container edgedb
Bootstrapping EdgeDB instance on remote Postgres cluster...
INFO 35 2022-02-11T17:32:14.977 edb.server: starting EdgeDB server 1.0+08d3e21, built on 2022-02-10T00:04Z from revision e16e60d1e (2022-02-09T01:08Z) for x86_64-unknown-linux-gnu
---- Exception occurred: invalid literal for int() with base 10: 'suik52Q' ----

1. ValueError: invalid literal for int() with base 10: 'suik52Q'

---- Traceback ----

    /usr/bin/edgedb-server, line 8, in <module>
        > sys.exit(main())
    /usr/lib/x86_64-linux-gnu/edgedb-server-1/lib/python3.10/site-packages/click/core.py, line 829, in __call__
        > return self.main(*args, **kwargs)
    /usr/lib/x86_64-linux-gnu/edgedb-server-1/lib/python3.10/site-packages/click/core.py, line 782, in main
        > rv = self.invoke(ctx)
    /usr/lib/x86_64-linux-gnu/edgedb-server-1/lib/python3.10/site-packages/click/core.py, line 1066, in invoke
        > return ctx.invoke(self.callback, **ctx.params)
    /usr/lib/x86_64-linux-gnu/edgedb-server-1/lib/python3.10/site-packages/click/core.py, line 610, in invoke
        > return callback(*args, **kwargs)
    /usr/lib/x86_64-linux-gnu/edgedb-server-1/lib/python3.10/site-packages/edb/server/main.py, line 615, in main
        > server_main(**kwargs)
    /usr/lib/x86_64-linux-gnu/edgedb-server-1/lib/python3.10/site-packages/edb/server/main.py, line 604, in server_main
        > asyncio.run(run_server(server_args))
    /usr/lib/x86_64-linux-gnu/edgedb-server-1/lib/python3.10/asyncio/runners.py, line 44, in run
        > return loop.run_until_complete(main)
[...]

PR #3483, opened 2022-02-14
Fix IPv6 address parsing in postgres URI

fmoor:
Plain IPv6 addresses specified in square brackets in the connection URI are now parsed correctly.

fixes https://github.com/edgedb/edgedb/issues/3454


elprans (approved):
Please cherry-pick into stable/1.x

PR #3484, opened 2022-02-14
Fix IPv6 address parsing in postgres URI

fmoor:
Plain IPv6 addresses specified in square brackets in the connection URI are now parsed correctly.

fixes https://github.com/edgedb/edgedb/issues/3454


fmoor:
RE: https://github.com/edgedb/edgedb/pull/3483#pullrequestreview-882147064

PR #3501, opened 2022-02-16
Make sure the compiler pool preserves Python "isolated mode"

elprans:
If we are running in isolated mode, as production builds will, make sure
the compiler pool passes it on.


fantix (approved):
👍

PR #3506, opened 2022-02-16
fix candidate for #2567: We generate a pointless self join

### LINKED ISSUE #2567: We generate a pointless self-join on simple link references
Part of the generated code for `SELECT User.deck.name` is

```
edgedbpub."2ba72ba6-bd92-11eb-86bd-39cdc1a3919c_t" AS "User~2"
INNER JOIN edgedbpub."2ba72ba6-bd92-11eb-86bd-39cdc1a3919c_t" AS "User~3"
    ON ("User~2".id = "User~3".id)
```

SonOfLilit:
Fixes #2567 and hopefully doesn't break anything else.

@elprans says "Indeed, we need to check if path_rvar is in scope_stmt before re-including it."

@msullivan this is in your court I believe.

### DIFF
--- a/tests/test_edgeql_sql_codegen.py
+++ b/tests/test_edgeql_sql_codegen.py
@@ -20,7 +20,6 @@
 import os.path
 
 from edb.testbase import lang as tb
-from edb.tools import test
 
 from edb.common.ast import visitor as ast_visitor
 
@@ -69,16 +68,18 @@ class TestEdgeQLSQLCodegen(tb.BaseEdgeQLCompilerTest):
 
         # Make sure that User is only selected from *once* in the query
         user_obj = self.schema.get('default::User')
-        user_id_str = str(user_obj.id)
         self.assertEqual(
-            sql.count(user_id_str), 1, "User table referenced more than once: " + sql
+            sql.count(str(user_obj.id)),
+            1,
+            "User table referenced more than once: " + sql
         )
 
         # Make sure that Profile is only selected from *once* in the query
         profile_obj = self.schema.get('default::Profile')
-        profile_id_str = str(profile_obj.id)
         self.assertEqual(
-            sql.count(profile_id_str), 1, "Profile table referenced more than once: " + sql
+            sql.count(str(profile_obj.id)),
+            1,
+            "Profile table referenced more than once: " + sql
         )
 
     def test_codegen_elide_optional_wrapper(self):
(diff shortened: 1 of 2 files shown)


elprans:
Not sure what's going on with the CLA bot here.  The webhook succeeded, but it's somehow still listed as failed.

tests/test_edgeql_sql_codegen.py
```diff
@@ -37,7 +36,7 @@ class TestEdgeQLSQLCodegen(tb.BaseEdgeQLCompilerTest):
     """
 
     SCHEMA = os.path.join(os.path.dirname(__file__), 'schemas',
-                          'cards.esdl')
+                          'graphql.esdl')
```
msullivan, line 39: Could you switch this back to `cards`, restore the original `no_self_join` test, and then add your modified test back in as a new test, except using the `issues` schema instead of `graphql` (using `Issue.priority`, maybe?)

I like adding a test that uses a `single` link, but would like to keep the `multi` link test also.

And `issues` and `cards` are the most widely used test schemas internally, so I'd rather just use those when possible, since they are more immediately familiar.

See `test_codegen_elide_optional_wrapper` below for an example of accessing the issues schema; the names to pass to `schema.get` will be `issues::Issue` (instead of `default::`).

msullivan (commented):
This looks great. I've got some testing nits, but otherwise looks good!

tests/test_edgeql_sql_codegen.py
msullivan: Oh, it could also use `cards` and `User.avatar`

msullivan: Oh, or we could modify this test suite to just *always* use `issues` and drop the use of `cards` (replacing `User.deck.name` with `User.todo.name`, maybe).

That's probably the right thing to do, but you don't need to do it as part of this PR.

(I usually reach for `cards` as my go to database when testing things, because I like the theming and because I think the test *data* for it is better, but `issues` is probably a better test schema in general because it exercises the feature set better; `cards` doesn't have any optional properties, which is why I pulled in `issues` also, but probably I should have just switched to `issues` when I did that)

SonOfLilit:
Cards, and to my memory Issues, don't have single links from one table to
another, which is the simplest case at least in terms of ease of debugging.
However, I can add tests based on those, sure.

msullivan:
`User.avatar` is a single link in `cards`. It would also be totally fine to use `issues`, which has a lot of single links.

msullivan (approved):
Looks good to me! Thanks for the fix.

elprans (approved):
Thanks!

PR #3508, opened 2022-02-17
Fix --ignore-other-tenants regression

elprans:
The `--ignore-other-tenants` was broken in #3249, which, in turn, breaks
in-place upgrades.  Alas, I couldn't come up with an easy way to test
this, because servers are started in subprocesses, which makes it hard
to mock things.


fantix:
Fixed in #3511, thanks!

PR #3515, opened 2022-02-19
Make sure wait_for_sync() does NOT discard error messages

elprans:
Postgres may generate errors between `CommandComplete` and
`ReadyForQuery`, such as when a single-statement transaction
fails to commit.  Make sure we don't ignore such errors or else we'll
erroneously report that the query was a success when it wasn't!

### DIFF
--- a/edb/server/pgcon/pgcon.pyx
+++ b/edb/server/pgcon/pgcon.pyx
@@ -570,19 +570,31 @@ cdef class PGConnection:
             await self.after_command()
 
     async def wait_for_sync(self):
-        while True:
-            if not self.buffer.take_message():
-                await self.wait_for_message()
-            mtype = self.buffer.get_message_type()
-            if mtype == b'Z':
-                self.parse_sync_message()
-                return
-            else:
-                if not self.parse_notification():
-                    if PG_DEBUG or self.debug:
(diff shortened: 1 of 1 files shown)


msullivan (approved):
Good catch

edb/server/pgcon/pgcon.pyx
```diff
@@ -570,19 +570,26 @@ cdef class PGConnection:
...
     async def wait_for_sync(self):
+        error = None
         while True:
             if not self.buffer.take_message():
                 await self.wait_for_message()
             mtype = self.buffer.get_message_type()
             if mtype == b'Z':
                 self.parse_sync_message()
-                return
+                break
```
1st1, line 580: OK, this is subtle. Please put the entire `while True` in a `try..finally`, with `finally` checking on `error is not None`. 

This will prevent us from adding another return (and not break) somewhere inside this block in the future. Also, please add a comment to the finally block explaining the current conundrum.

1st1 (changes requested):

edb/server/pgcon/pgcon.pyx
1st1: `try..finally` would also save us from an unrelated Python-level exception masking the more important and potentially trickier Postgres error.

1st1 (approved):

PR #3520, opened 2022-02-20
Fix typo in parameters docs (examples)

x1B:
I may be missing something, but shouldn't the parameter name in the EdgeQL query match the property name in the JS substitutions arg?


elprans (approved):
Good catch! Thanks!

PR #3527, opened 2022-02-21
Raise a proper error on TypeIntrospection in simple expressions

### LINKED ISSUE #3494: Sequence default fails in migration
### Discussed in https://github.com/edgedb/edgedb/discussions/3493

<div type='discussions-op-text'>

<sup>Originally posted by **tchak** February 15, 2022</sup>
I have the following schema:

```
module default {
  scalar type ProjectNumber extending sequence;

  type Project {
    required property number -> ProjectNumber {
      constraint exclusive;
    }
  }
}
```

I expect `number` to be autofilled. But when I use this schema from TypeScript driver type signature refuses to let me create a Project without specifying a `number` property. I tried the following schema but it failed on migration:

```
module default {
  scalar type ProjectNumber extending sequence;

  type Project {
    required property number -> ProjectNumber {
      constraint exclusive;
      default := sequence_next(introspect ProjectNumber);
    }
  }
}
```

Is this is a bug or intended behaviour ?</div>

msullivan:
Fixes #3494.

At least kind of. We could also probably support this. The main problem
is just that sequence_next wants the type as an extra argument.

### DIFF
--- a/edb/pgsql/compiler/expr.py
+++ b/edb/pgsql/compiler/expr.py
@@ -315,6 +315,14 @@ def compile_SliceIndirection(
     return result
 
 
+@dispatch.compile.register(irast.TypeIntrospection)
+def compile_TypeIntrospection(
+        expr: irast.TypeIntrospection, *,
+        ctx: context.CompilerContextLevel) -> pgast.BaseExpr:
+    raise errors.UnsupportedFeatureError(
+        'type introspection not supported in simple expressions')
+
+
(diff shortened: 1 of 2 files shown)


elprans (approved):
Would be nice to actually support this case properly (e.g. `TypeIntrospection` can compile into JSON or a tuple in simple expression mode), but a proper error is OK for now.

msullivan:
Cherry-picked into stable/1.x

PR #3528, opened 2022-02-21
Proper errors when a scalar type tries to extend nothing

msullivan:


edb/schema/scalars.py
```diff
@@ -370,6 +370,11 @@ def _cmd_tree_from_ast(
             create_cmd = cmd
 
         if isinstance(astnode, qlast.CreateScalarType):
+            if not astnode.bases:
+                raise errors.SchemaError(
+                    f'scalar type must have a supertype',
+                    context=astnode.context,
```
elprans, line 376: Let's add a hint with an example, e.g. `hint="For example: scalar type {name} extending str"`

elprans (approved):

msullivan:
Cherry-picked to stable/1.x

PR #3529, opened 2022-02-21
Disable rebasing scalar types

msullivan:
Currently doing this can allow pg and edgedb to disagree about a column's type!

### DIFF
--- a/edb/schema/scalars.py
+++ b/edb/schema/scalars.py
@@ -573,18 +573,16 @@ class RebaseScalarType(
             self.validate_scalar_bases(schema, context)
 
         else:
-            if not self.scls.get_abstract(schema):
-                old_concrete = self.scls.get_topmost_concrete_base(schema)
+            old_concrete = self.scls.maybe_get_topmost_concrete_base(schema)
 
             schema = super().apply(schema, context)
 
             self.validate_scalar_bases(schema, context)
 
-            if not self.scls.get_abstract(schema):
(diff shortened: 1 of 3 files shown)


elprans (changes requested):
Mm. I don't think it's a good idea to disable rebasing outright.  Adding or inserting (or removing) a new (abstract) base is fine.  What we _do_ want to prohibit is changing the topmost non-abstract parent.

edb/schema/scalars.py
```diff
@@ -570,10 +570,22 @@ def apply(
...
         else:
+            if not self.scls.get_abstract(schema):
+                old_concrete = self.scls.get_topmost_concrete_base(schema)
+
             schema = super().apply(schema, context)
 
-        self.validate_scalar_bases(schema, context)
+            self.validate_scalar_bases(schema, context)
+
+            if not self.scls.get_abstract(schema):
```
elprans, line 583: Why are abstract scalars special?  Wouldn't this loophole let me do

```
abstract scalar type Foo extending str;
scalar type Bar extending Foo;
```

and then `alter scalar type Foo ... extending int64` with bad consequences for `Bar`?  I think we at least ought to check if the abstract scalar has no subtypes.

msullivan: Yeah, you're right. The reason was because abstract scalars might not have a concrete base at all, but that can be handled differently

elprans (approved):

msullivan:
Cherry-picked into stable/1.x

PR #3536, opened 2022-02-22
Don't consider a linkprop on a multi link to be multi in a constraint on it

msullivan:
Makes using EXISTS/?!= work on linkprop constraints.
Helps mitigate #1727.

### DIFF
--- a/tests/test_edgeql_ddl.py
+++ b/tests/test_edgeql_ddl.py
@@ -8114,7 +8114,7 @@ type default::Foo {
                 """)
 
     @test.xfail('''
-        EXISTS constraint violation not raised for MULTI property.
+        Reports an schema error. Maybe that is exactly what we want?
     ''')
     async def test_edgeql_ddl_constraint_04(self):
         # Test for #1727. Usage of EXISTS in constraints.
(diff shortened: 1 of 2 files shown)


tests/test_edgeql_ddl.py
```diff
@@ -8114,7 +8114,7 @@ async def test_edgeql_ddl_constraint_03(self):
                 """)
 
     @test.xfail('''
-        EXISTS constraint violation not raised for MULTI property.
+        Reports an schema error. Maybe that is exactly what we want?
```
elprans, line 8117: Ideally, we should support this, but raising SchemaError is a lot better than accepting it and then _not_ raising the constraint violation.

elprans (approved):

tests/test_edgeql_ddl.py
msullivan: I guess we could enforce complex constraints manually at the DML side...

msullivan:
cherry picked for 1.x

tests/test_edgeql_ddl.py
elprans: Same way we enforce `required`.

PR #3545, opened 2022-02-23
preparser: fix backslash at the end of raw string

tailhook:
Fixes edgedb/edgedb-cli#676

Also adds examples of binary raw strings to docs


1st1:
@elprans Should we cherry-pick this to 1.x?

PR #3546, opened 2022-02-23
Fix collection of migration issues with inherited props

### LINKED ISSUE #3542: Unable to create migration altering a property from required to optional
- EdgeDB Version: EdgeDB CLI 1.1.0+8bc2c20
- OS Version: MacOS Catalina 10.15.7

Steps to Reproduce:

1. Create a new EdgeDB project (version 1.0) with `default` module as below:
```
module default {
    abstract type Person {
    }

    type PC extending Person {
    }
}
```
2. Create and apply migration
```
user@local> edgedb migration create
did you create object type 'default::Person'? [y,n,l,c,b,s,q,?]
> y
did you create object type 'default::PC'? [y,n,l,c,b,s,q,?]
> y
Created ./dbschema/migrations/00001.edgeql, id: m1clwlsejlohyzkgrc234l4etgml2mnq6ttpfvvpaykmrfujj5lzaq

user@local> edgedb migrate
Applied m1clwlsejlohyzkgrc234l4etgml2mnq6ttpfvvpaykmrfujj5lzaq (00001.edgeql)
Note: adding first migration disables DDL. More info: https://edgedb.com/p/bare_ddl

```
3. Remove `required` from `default::Person::name` property and create new migration.
```
user@local> edgedb migration create
did you make property 'name' of object type 'default::Person' optional? [y,n,l,c,b,s,q,?]
> y
edgedb error: EdgeDB could not resolve migration with the provided answers. Please retry with different answers.

```

As you see from the terminal output above, the migration to make property `default::Person::name` optional could not be created. The error message is quite confusing `EdgeDB could not resolve migration with the provided answers. Please retry with different answers`.

How can I fix it?

msullivan:
* Propagate RESETs to non-owned children
 * Don't merge modifications to owned children when ordering migrations.
   The modification to the parent won't apply to the child, so we have
   to explicitly do the child.
 * Prune required_user_input whose placeholders have been elided from the
   actual DDL generated.

Fixes #3542.

### DIFF
--- a/edb/schema/ordering.py
+++ b/edb/schema/ordering.py
@@ -72,6 +72,8 @@ def linearize_delta(
         commands.
     """
 
+    # delta.dump()
+
     # We take the scatter-sort-gather approach here, where the original
     # tree is broken up into linear branches, which are then sorted
     # and reassembled back into a tree.
@@ -843,7 +845,12 @@ def _trace_op(
                     for ref in refs:
                         write_ref_deps(ref, referrer, this_name_str)
 
-                if isinstance(obj, referencing.ReferencedInheritingObject):
+                if (
+                    isinstance(obj, referencing.ReferencedInheritingObject)
+                    # Changes to owned objects can't necessarily be merged
+                    # in with parents, so we make sure not to.
+                    and not obj.get_owned(new_schema)  # XXX ???
+                ):
                     implicit_ancestors = [
                         b.get_name(new_schema)
                         for b in obj.get_implicit_ancestors(new_schema)
--- a/edb/schema/ordering.py
+++ b/edb/schema/ordering.py
@@ -72,8 +72,6 @@ def linearize_delta(
         commands.
     """
 
-    # delta.dump()
-
     # We take the scatter-sort-gather approach here, where the original
     # tree is broken up into linear branches, which are then sorted
     # and reassembled back into a tree.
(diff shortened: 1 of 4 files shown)


edb/schema/ordering.py
```diff
@@ -72,6 +72,8 @@ def linearize_delta(
         commands.
     """
 
+    # delta.dump()
```
elprans, line 75:
```suggestion
```

edb/schema/ordering.py
```diff
@@ -843,7 +845,12 @@ def write_ref_deps(
                     for ref in refs:
                         write_ref_deps(ref, referrer, this_name_str)
 
-                if isinstance(obj, referencing.ReferencedInheritingObject):
+                if (
+                    isinstance(obj, referencing.ReferencedInheritingObject)
+                    # Changes to owned objects can't necessarily be merged
+                    # in with parents, so we make sure not to.
+                    and not obj.get_owned(new_schema)  # XXX ???
```
elprans, line 852: Is `XXX` still relevant?  If so, add a comment why?

tests/test_edgeql_data_migration.py
```diff
@@ -8785,6 +8785,108 @@ async def test_edgeql_migration_annotation_05(self):
...
+        child is still optional.
+
+        For this to work, we need to process the *child* first,
+        but in order for the reverse case above to work we need to
+        process the *parent* first.
+
+        I don't know if there is any way to fix this sort of thing without
+        exposing this kind of semantic understanding to ordering.
+
+        Maybe suppressing some kinds of intermediate-state errors during
```
elprans, line 8866: We can do that, but only if the intermediately-broken-state exists _within_ a tree corresponding to a single DDL command.  We already do this for deletions.  Please open a ticket so we don't forget to look into this later.

elprans: One possible solution would be to add DDL syntax, e.g. `SET REQUIRED RECURSIVE` or something.

elprans (approved):

msullivan:
Cherry-picked to 1.x

PR #3548, opened 2022-02-24
Restructure the changelog

1st1:


1st1:
@vpetrovykh please push the 1.1 changes directly to this branch.

-------

Here's how this looks:


Screen Shot 2022-02-23 at 5 08 06 PM

elprans:
This unfortunately breaks links we inserted into release tags, but I guess it isn't a big deal?

1st1:
We can just keep the files where they were, so the urls will keep working.

vpetrovykh:
I was adding a few of the PRs submitted by the community in the changelog.
Also I think I may have included some changes that actually made it into 1.0 here as I wasn't sure that omitting them altogether makes sense.

So this one is a bit longer than I expected.

elprans (approved):

docs/changelog/1_x.rst
```diff
@@ -0,0 +1,75 @@
...
+  link (:eql:gh:`#3536`).
+* Disable changing the concrete base of a :ref:`scalar type
+  <ref_datamodel_scalar_types>` (:eql:gh:`#3529`).
+* Avoid generating pointless self joins (:eql:gh:`#2567`).
+* Fix IPv6 address parsing (:eql:gh:`#3454`).
+* If a type has object instances, it cannot be made ``abstract``
+  (:eql:gh:`#3399`).
+* Fix an issue that sometimes caused :eql:kw:`with` block variables to
+  be unusable (:eql:gh:`#3385`).
+* The ``start`` keyword is now reserved (:eql:gh:`#3396`).
```
elprans, line 50: This bit is unnecessary.  We re-reserved `start` in 1.0

PR #3550, opened 2022-02-24
Add on-demand compiler pool scaling

fantix:
Use --compiler-pool-scaling-mode=on_demand to switch to the new mode,
which will a spawn new compiler worker process every 3 seconds if the
compiling requests keep queueing up, up to num of CPUs; after 60 seconds
without compiling requests, the compiler pool will scale down to
--compiler-pool-size with a default of 1 under on-demand mode.

### DIFF
--- a/edb/server/args.py
+++ b/edb/server/args.py
@@ -94,6 +94,41 @@ class BackendCapabilitySets(NamedTuple):
     must_be_absent: List[pgsql_params.BackendCapabilities]
 
 
+class CompilerPoolScalingMode(enum.StrEnum):
+    Fixed = "fixed"
+    OnDemand = "on_demand"
+
+    def __init__(self, name):
+        self.pool_class = None
+
+    def set_class(self, cls):
+        self.pool_class = cls
+        return cls
+
+    def compute_default_pool_size(self):
+        if self.value == self.Fixed:
+            return compute_default_compiler_pool_size()
+        else:
+            return 1
+
+
+class OptionWithDynamicHelp(click.Option):
+    def __init__(self, *args, help_func, **kwargs):
+        self._help = None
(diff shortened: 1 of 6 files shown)


edb/server/args.py
```diff
@@ -94,6 +94,41 @@ class BackendCapabilitySets(NamedTuple):
...
 
 
+class CompilerPoolScalingMode(enum.StrEnum):
+    Fixed = "fixed"
+    OnDemand = "on_demand"
+
+    def __init__(self, name):
+        self.pool_class = None
+
+    def set_class(self, cls):
```
1st1, line 104: I think `assign_implementation` would be less confusing. Also add a comment that this is a decorator

edb/server/compiler_pool/pool.py
```diff
@@ -736,6 +703,201 @@ async def describe_database_restore(
...
+    async def _start(self):
+        await self._create_template_proc(retry=False)
+
+    async def _create_template_proc(self, retry=True):
+        self._template_proc_scheduled = False
+        if not self._running:
+            return
+        self._template_proc_version += 1
+        try:
+            self._template_transport = await self._create_worker_process(
```
1st1, line 750: It's a bit hard to follow the changes -- are we still using fork for the "fixed" pool?

1st1: This is in part because of using inheritance. Can you add comments here and there explaining the key points: where we spawn, where we fork, and how we do that?

fantix: Yeah sure, np

1st1:
Should we merge this one?

elprans (approved):
:+1:

PR #3555, opened 2022-02-25
Remote compiler

fantix:
Depends on #3550.

## Overall architecture

cloud demo drawio

### EdgeDB Server:

```
edgedb-server --compiler-pool-mode remote --compiler-pool-addr /path/to/socket --compiler-pool-size 4
```

When the mode is set to `remote`, the pool size means how many concurrent compile requests can the EdgeDB server issue at most, default is `2`.

### Compiler Server:

```
edgedb-server compiler /path/to/socket --pool-size 4
```

The compiler server will also create another socket file under the same dir for internal communication. Pool size means the number of worker processes the compiler server will maintain, default to the number of CPUs.

## Schema

Information needed to compile:

* "**std schema**" (system schema): Immutable constructs, identical among all EdgeDB instances (clients):
  * `backend_runtime_params` - for now just assume it's identical among clients;
  * Shared among the same catalog_version, so identical among clients:
    * `std_schema`
    * `refl_schema`
    * `schema_class_layout`
* "**schema n**" (client schema): Different across clients, and changes on the fly:
  * `global_schema`
  * `system_config`
  * DatabaseStates
    * `dbname` (key)
    * `user_schema`
    * `reflection_cache`
    * `database_config`

cloud demo-第 2 页 drawio

* All EdgeDB instances (clients) share the same system schema;
* Compiler server always keep in sync with all connected clients;
* Compiler server keeps only one copy of the std schema to bootstrap new worker subprocesses;
* Compiler server only keeps pickled copies of the client schemas;
* Each compiler worker has an LRU cache for client schemas;
* If a client disconnects from the compiler server, its cache in the worker is invalidated;
* The compiler server manages the cache in all workers by sending additional RPC parameters as instructions which are followed by the workers.

### Client-based Requests

This includes `compile()`, `compile_notebook()` and `compile_graphql()`.

* If some workers have the schema for the client, choose the most-recently-used worker among them;
* Or else, choose a worker with the least client schemas.

### Transaction-based Requests

This is only `compile_in_tx()`, but `compile()` may also affect the compiler state.

Each worker keeps only one most-recent compiler state, the compiler server keeps the SHA1 of the pickled states in sync with the workers. Between the EdgeDB instance and the compiler worker, EdgeDB firstly calls `compile_in_tx()` with the SHA1 of the previous pickled state, and retry with the full pickled state if the compiler server replies the SHA1 missed the cache. Like a regular EdgeDB/compiler, the worker is put to the end of the queue for higher probability of cache hit.

Benchmarks: 4 CPUs, 4 worker process, 4 test processes running the server test suite, semaphore=4

* Cache hit: 97%
* Transaction pickled state average size: 165KB (23KB ~ 1.26MB)

UPDATE: the compiler server should generate a sequential number to reflect the different unique pickled states, instead of calculating the SHA1 of hundreds of KBs of data.

### Other Requests

Other requests like `try_compile_rollback()` or anything on the COMPILER is simply using the workers at the beginning of the queue.

## TODOs

- [x] Fix mypy & pyflakes
- [x] Extract duplicate code
- [x] Manually test most branches
- [x] Add automated tests (https://github.com/edgedb/edgedb/runs/5487539100?check_suite_focus=true)
- [ ] Maybe use `dbview.dbver`?
- [ ] CLI interface - do we want DSN?

### DIFF
--- a/edb/schema/schema.py
+++ b/edb/schema/schema.py
@@ -493,6 +493,23 @@ class FlatSchema(Schema):
         self._refs_to = immu.Map()
         self._generation = 0
 
+    def eq(self, other: FlatSchema) -> bool:
+        if self._id_to_data != other._id_to_data:
+            return False
+        if self._id_to_type != other._id_to_type:
+            return False
+        if self._shortname_to_id != other._shortname_to_id:
+            return False
+        if self._name_to_id != other._name_to_id:
+            return False
+        if self._globalname_to_id != other._globalname_to_id:
+            return False
+        if self._refs_to != other._refs_to:
+            return False
+        if self._generation != other._generation:
+            return False
+        return True
+
     def _replace(
         self,
         *,
--- a/edb/schema/schema.py
+++ b/edb/schema/schema.py
@@ -493,23 +493,6 @@ class FlatSchema(Schema):
         self._refs_to = immu.Map()
         self._generation = 0
 
-    def eq(self, other: FlatSchema) -> bool:
-        if self._id_to_data != other._id_to_data:
-            return False
-        if self._id_to_type != other._id_to_type:
-            return False
-        if self._shortname_to_id != other._shortname_to_id:
-            return False
-        if self._name_to_id != other._name_to_id:
-            return False
-        if self._globalname_to_id != other._globalname_to_id:
-            return False
-        if self._refs_to != other._refs_to:
-            return False
-        if self._generation != other._generation:
-            return False
-        return True
-
     def _replace(
         self,
         *,
--- a/edb/server/args.py
+++ b/edb/server/args.py
@@ -630,11 +630,16 @@ _compiler_options = [
         type=int,
         callback=_validate_compiler_pool_size,
         default=compute_default_compiler_pool_size(),
+        help=f"Number of compiler worker processes. Defaults to "
+             f"{compute_default_compiler_pool_size()}.",
     ),
     click.option(
-        "--cache-size",
+        "--client-schema-cache-size",
         type=int,
         default=100,
+        help="Number of client schemas each worker could cache at most. The "
+             "compiler server is not affected by this setting, it keeps a "
+             "pickled copy of the client schema of all active clients."
     ),
     click.argument("socket_path"),
 ]
@@ -800,6 +805,9 @@ def parse_args(**kwargs: Any):
     )
     if kwargs['compiler_pool_size'] is None:
         if kwargs['compiler_pool_mode'] == CompilerPoolMode.Remote:
+            # this reflects to a local semaphore to control concurrency,
+            # 2 means this is a small EdgeDB instance that could only issue
+            # at max 2 concurrent compile requests at a time.
             kwargs['compiler_pool_size'] = 2
         else:
             kwargs['compiler_pool_size'] = compute_default_compiler_pool_size()
--- a/edb/server/args.py
+++ b/edb/server/args.py
@@ -279,6 +279,19 @@ def _validate_compiler_pool_size(ctx, param, value):
     return value
 
 
+def _validate_host_port(ctx, param, value):
+    if value is None:
+        return None
+    address = value.split(":", 1)
+    if len(address) == 1:
+        return address[0], defines.EDGEDB_REMOTE_COMPILER_PORT
+    else:
+        try:
+            return address[0], int(address[1])
+        except ValueError:
+            raise click.BadParameter(f'port must be int: {address[1]}')
+
+
 def compute_default_compiler_pool_size() -> int:
     total_mem = psutil.virtual_memory().total
     total_mem_mb = total_mem // MIB
@@ -466,7 +479,10 @@ _server_options = [
     ),
     click.option(
         '--compiler-pool-addr',
-        type=str,
+        callback=_validate_host_port,
+        help=f'Specify the host[:port] of the compiler pool to connect to, '
+             f'only used if --compiler-pool-mode=remote. Default host is '
+             f'localhost, port is {defines.EDGEDB_REMOTE_COMPILER_PORT}',
     ),
     click.option(
         '--echo-runtime-info', type=bool, default=False, is_flag=True,
@@ -642,8 +658,20 @@ _compiler_options = [
              "pickled copy of the client schema of all active clients."
     ),
     click.option(
-        "--socket-path",
-        required=True,
+        '-I', '--listen-addresses', type=str, multiple=True,
+        default=('localhost',),
+        help='IP addresses to listen on, specify multiple times for more than '
+             'one address to listen on. Default: localhost',
+    ),
+    click.option(
+        '-P', '--listen-port', type=PortType(),
+        help=f'Port to listen on. '
+             f'Default: {defines.EDGEDB_REMOTE_COMPILER_PORT}',
+    ),
+    click.option(
+        '--runstate-dir', type=PathPath(), default=None,
+        help="Directory to store UNIX domain socket file for IPC, a temporary "
+             "directory will be used if not specified.",
     ),
 ]
 
@@ -816,7 +844,12 @@ def parse_args(**kwargs: Any):
             kwargs['compiler_pool_size'] = compute_default_compiler_pool_size()
     if kwargs['compiler_pool_mode'] == CompilerPoolMode.Remote:
         if kwargs['compiler_pool_addr'] is None:
-            abort('--compiler-pool-mode=remote requires --compiler-pool-addr')
+            kwargs['compiler_pool_addr'] = (
+                "localhost", defines.EDGEDB_REMOTE_COMPILER_PORT
+            )
+    elif kwargs['compiler_pool_addr'] is not None:
+        abort('--compiler-pool-addr is only meaningful '
+              'under --compiler-pool-mode=remote')
 
     if kwargs['temp_dir']:
         if kwargs['data_dir']:
--- a/edb/server/args.py
+++ b/edb/server/args.py
@@ -97,6 +97,7 @@ class BackendCapabilitySets(NamedTuple):
 class CompilerPoolMode(enum.StrEnum):
     Fixed = "fixed"
     OnDemand = "on_demand"
+    Remote = "remote"
 
     def __init__(self, name):
         self.pool_class = None
@@ -133,6 +134,7 @@ class ServerConfig(NamedTuple):
     max_backend_connections: Optional[int]
     compiler_pool_size: int
     compiler_pool_mode: CompilerPoolMode
+    compiler_pool_addr: str
     echo_runtime_info: bool
     emit_server_status: str
     temp_dir: bool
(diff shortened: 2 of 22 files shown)


1st1:
Screen Shot 2022-02-25 at 10 25 09 AM

by @fantix

edb/server/main.py
```diff
@@ -609,19 +609,31 @@ def server_main(**kwargs):
...
+    if ctx.invoked_subcommand is None:
+        server_main(**kwargs)
+
+
+@main.command()
+@srvargs.compiler_options
+def compiler(**kwargs):
+    from edb.server.compiler_pool import server as compiler_server
+
+    asyncio.run(compiler_server.server_main(**kwargs))
```
1st1, line 636: Does this have to be defined in this file? Maybe `compiler_main.py`?

edb/server/compiler_pool/remote_worker.py
```diff
@@ -0,0 +1,481 @@
...
+
+
+INITED: bool = False
+clients: immutables.Map[int, ClientSchema] = immutables.Map()
+BACKEND_RUNTIME_PARAMS: pgparams.BackendRuntimeParams = (
+    pgparams.get_default_runtime_params()
+)
+COMPILER: compiler.Compiler
+LAST_STATE: Optional[compiler.dbstate.CompilerConnectionState] = None
+STD_SCHEMA: s_schema.FlatSchema
```
1st1, line 59: FWIW if you want to avoid code duplication we can move this state into a class (as well as `__sync__` and `__init_worker__` methods etc). At this stage you should know if the code is 99% similar or if it's different enough to warrant copy/paste.

edb/server/args.py
```diff
@@ -109,8 +110,12 @@ def assign_implementation(self, cls):
     def compute_default_pool_size(self):
         if self.value == self.Fixed:
             return compute_default_compiler_pool_size()
-        else:
+        elif self.value == self.OnDemand:
             return 1
+        elif self.value == self.Remote:
+            return 2
```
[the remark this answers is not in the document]
fantix: (1) Sure thing!

(2) 

> When the mode is set to remote, the pool size means how many concurrent compile requests can the EdgeDB server issue at most, default is `2`.

Well this doesn't really have a concrete reason, 2 is just something on top of my head.

edb/server/compiler_pool/server.py
```diff
@@ -0,0 +1,547 @@
...
+            other_state = other.dbs.get(dbname)
+            if other_state is None:
+                dbs = dbs.set(dbname, state)
+            elif state is not other_state:
+                dbs = dbs.set(dbname, state.diff(other_state))
+        return ClientSchema(
+            dbs,
+            None
+            if self.global_schema is other.global_schema
+            else self.global_schema,
```
1st1, line 95: nit: I personally find this hard to read, I'd just move this into a variable.

edb/server/compiler_pool/server.py
```diff
@@ -0,0 +1,547 @@
...
+            self._release_worker(worker, put_in_front=False)
+
+    async def _request(self, method_name, msg):
+        worker = await self._acquire_worker()
+        try:
+            return await worker.call(method_name, msg=msg)
+        finally:
+            self._release_worker(worker)
+
+    async def call(self, protocol, req_id, msg):
```
1st1, line 432: I'm confused... This isn't a "call an RPC", this is "handle a call", right? If so it should be named something like "handle_client_call", no?

1st1 (commented):
Mostly nit picking

edb/server/args.py
```diff
@@ -640,6 +650,28 @@ def server_options(func):
...
 
+_compiler_options = [
+    click.option(
+        "--pool-size",
+        type=int,
+        callback=_validate_compiler_pool_size,
+        default=compute_default_compiler_pool_size(),
+    ),
+    click.option(
+        "--cache-size",
```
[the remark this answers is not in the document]
fantix: It's the cached client schemas in workers - I'll rename and add help.

edb/server/args.py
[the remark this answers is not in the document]
1st1: I still don't understand why is this `2` or for whatever it worth what any other number would even mean in this context.

edb/server/main.py
fantix: I usually use a `entry_point` in `setup.py` to inject the dependencies of the CLI modules (example [here](https://python-gino.org/docs/en/1.0/tutorials/fastapi.html#create-models-and-api) using `importlib` and Poetry plugins), in order to avoid a late import (which is necessary to avoid a circular import) like this:

`main.py`:

```python
@click.group(...)
def main(...):
    ...

from . import compiler_main  # noqa: late import to avoid circular import
```

`compiler_main.py`:

```python
from . import main

@main.command()
def compiler(...):
    from edb.server.compiler_pool import server
    ...server.server_main()...
```

For the EdgeDB server, I don't think we need a dependency injection because we're not a extensible library. So I was leaning towards just putting the `compiler` subcommand in `edb.server.main`, because a late import cannot hide too much implementation details (like 1 line vs 5 lines, and we still see the "compiler" keyword in the main server).

Even with that said, I'm still fine with adding the `from . import compiler_main` here, just reply 🚀 ;)

edb/server/compiler_pool/remote_worker.py
fantix: Ah yes - I'll likely re-organize the code with module-level functions, because the `compile_*` methods are largely different. Let me try now.

edb/server/compiler_pool/server.py
```diff
@@ -0,0 +1,547 @@
...
+    _tx_state_id_seq = (_tx_state_id_seq + 1) % (2 ** 63 - 1)
+    return _tx_state_id_seq
+
+
+class PickledState(typing.NamedTuple):
+    user_schema: typing.Optional[bytes]
+    reflection_cache: typing.Optional[bytes]
+    database_config: typing.Optional[bytes]
+
+    def diff(self, other: PickledState):
```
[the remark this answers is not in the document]
fantix: Will do!

edb/server/compiler_pool/server.py
fantix: Right!

edb/server/compiler_pool/server.py
```diff
@@ -0,0 +1,547 @@
...
+        if self._inited.is_set():
+            logger.debug("New client %d connected.", client_id)
+            if self._backend_runtime_params != backend_runtime_params:
+                raise RuntimeError(
+                    "Incompatible client: backend_runtime_params"
+                )
+            # if self._std_schema != std_schema:
+            #     raise RuntimeError("Incompatible client: std_schema")
+            # if self._refl_schema != refl_schema:
+            #     raise RuntimeError("Incompatible client: refl_schema")
```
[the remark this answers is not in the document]
fantix: Oh forgot about this one! I agree with you, but it was mismatching for some reason ... let me see now.

edb/server/compiler_pool/server.py
fantix: Yes!

edb/server/main.py
1st1: I was asking why not add another script "edgedb_compiler_pool" to https://github.com/edgedb/edgedb/blob/f4c1ced364a94e8206297fb970ef494c9b1a2de3/setup.py#L820-L824 and keep everything related in a separate file from main.py?

edb/server/args.py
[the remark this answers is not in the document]
fantix: Before this PR, only 1 compile request can be submitted at the same time per UNIX connection. We will have only 1 UNIX connection between each EdgeDB server and the compiler server, so after this PR, multiple compile requests can "multiplex" on the same UNIX connection. In this case, `--compiler-pool-size` is used to set the max concurrency of the "multiplex", so that the compiler server won't be overloaded by a single EdgeDB server.

edb/server/main.py
fantix: Ohhhh I see what you meant. I think @elprans had some quick thoughts to use a subcommand under `edgedb-server` (to minimize the number of executables?).

edb/server/compiler_pool/server.py
```diff
@@ -0,0 +1,545 @@
...
+            schema_class_layout,
+            global_schema,
+            system_config,
+        ) = pickle.loads(init_args_pickled)
+
+        if self._inited.is_set():
+            logger.debug("New client %d connected.", client_id)
+            if self._backend_runtime_params != backend_runtime_params:
+                raise state_mod.IncompatibleClient("backend_runtime_params")
+            if not self._std_schema.eq(std_schema):
```
elprans, line 238: This check seems very expensive.  Why can't we use catalog_version or some other indicator of compatibility instead?

fantix: It was a leftover issue from PoC. Using `catalog_version` makes sense, I'm pushing a fix.

edb/server/main.py
elprans: Yeah. Exposing entrypoints in prod builds is a pain, so it would be best take the "single entrypoint" approach to operational modes.

edb/server/args.py
```diff
@@ -640,6 +654,33 @@ def server_options(func):
...
+    ),
+    click.option(
+        "--client-schema-cache-size",
+        type=int,
+        default=100,
+        help="Number of client schemas each worker could cache at most. The "
+             "compiler server is not affected by this setting, it keeps a "
+             "pickled copy of the client schema of all active clients."
+    ),
+    click.argument("socket_path"),
```
elprans, line 674: Let's make this an option as well.  I know it's not optional, but then neither is `-D`

edb/schema/schema.py
```diff
@@ -493,6 +493,23 @@ def __init__(self) -> None:
         self._refs_to = immu.Map()
         self._generation = 0
 
+    def eq(self, other: FlatSchema) -> bool:
```
elprans, line 496: Seems like a dead function

edb/server/args.py
```diff
@@ -463,6 +464,10 @@ def _writer(status: str) -> None:
              'automatically scale up (to --compiler-pool-size workers ) and '
              'down to the demand. Default to "fixed".',
     ),
+    click.option(
+        '--compiler-pool-addr',
```
elprans, line 468: I'm not sure a Unix socket for the remote pool is worth it operationally.  It's a lot easier to operate with TCP in prod.

elprans: to clarify: I'm OK if we support Unix too, it's just that it shouldn't be the only mode

elprans: On a second thought, let's not make our lives harder and support TCP only.

edb/server/args.py
```diff
@@ -619,6 +624,36 @@ def server_options(func):
...
+    click.option(
+        "--client-schema-cache-size",
+        type=int,
+        default=100,
+        help="Number of client schemas each worker could cache at most. The "
+             "compiler server is not affected by this setting, it keeps a "
+             "pickled copy of the client schema of all active clients."
+    ),
+    click.option(
+        "--socket-path",
```
elprans, line 645: This should be `--listen-addresses` / `--listen-port` per the comment re socket above.

elprans (approved):

PR #3556, opened 2022-02-25
Add an "in" operator to GraphQL.

### LINKED ISSUE #3443: GraphQL: Filter for multiple IDs
It would be very hand to be able to filter for multiple IDs from the GraphQL API.

Example:
```
query {
  Person(filter: {id: {in: ["foo", "bar"]} }) {
    name
  }
}
```

vpetrovykh:
Much like "eq" and "neq", the "in" operator is available for all
scalars.

Fixes #3443

### DIFF
--- a/edb/graphql/translator.py
+++ b/edb/graphql/translator.py
@@ -1383,6 +1383,14 @@ class GraphQLTranslator:
                 # The binary operator that we need here is "="
                 op = '='
 
+            elif op == 'IN':
+                # Instead of wrapping the values in an array, wrap
+                # them in a set
+                value = qlast.FunctionCall(
+                    func='array_unpack',
+                    args=[value],
+                )
+
             elif self._context.right_cast is not None:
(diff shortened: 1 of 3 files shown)


edb/graphql/translator.py
```diff
@@ -1383,6 +1383,11 @@ def visit_ObjectFieldNode(self, node):
                 # The binary operator that we need here is "="
                 op = '='
 
+            elif op == 'IN':
+                # Instead of wrapping the values in an array, wrap
+                # them in a set
+                value = qlast.Set(elements=value.elements)
```
elprans, line 1389: Currently, `in array_unpack($array)` works better than `in <set>`.

vpetrovykh: Fixed.

elprans (approved):

msullivan:
Cherry-picked to 1.2

PR #3557, opened 2022-02-25
Add some tests for default values in GraphQL.

vpetrovykh:


tests/test_http_graphql_mutation.py
```diff
@@ -1420,6 +1420,66 @@ def test_graphql_mutation_insert_multiple_01(self):
             "Profile": [],
         })
 
+    def test_graphql_mutation_insert_default_01(self):
+        # Test insert object with a required property with a default.
+        data = {
+            'name': 'New InactiveUser01',
+            'age': 99,
+            'active': False,
+            'score': 0,
```
1st1, line 1429: maybe try setting values different from the default? What's the purpose of this PR -- I don't see any code fixes, just new tests...

vpetrovykh: We already have tests that set other values, this one is making sure that:
1) GraphQL allows omitting required fields with a default
2) The values get correctly set and sent back

elprans (approved):

msullivan:
Cherry-picked to 1.2

PR #3571, opened 2022-03-02
Determine func/oper argument type modifiers in advance of compiling args

msullivan:
This is based on a previous attempt, #2132.

This allows us to compile arguments with the correct fences in the
first place, instead of trying to fix them up after the fact.

The key insight here is that if a function argument could be either
OPTIONAL or SINGLETON, it is safe to treat it as optional. The
generated code will be a bit worse, since it will include an optional
wrapper.

This change allowed me to rip out a bunch nasty mechanisms for dealing
with OPTIONAL that I introduced during the first few months of my
employment and that I have been trying to kill ever since
(optional_count, path_log).

This is about code-size-neutral but the added code is boring plumbing
(to find typemods before fully typechecking) and the removed code is
fiddly hacks.

This also proves that it should be possible to precompute a scope tree
without needing to do move typechecking before scope tree inference.

### DIFF
--- a/edb/buildmeta.py
+++ b/edb/buildmeta.py
@@ -44,7 +44,7 @@ from edb.common import verutils
 
 
 # Increment this whenever the database layout or stdlib changes.
-EDGEDB_CATALOG_VERSION = 2022_01_07_00_00
+EDGEDB_CATALOG_VERSION = 2022_03_01_00_00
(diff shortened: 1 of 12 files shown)


msullivan:
Sure, done

elprans (approved):

PR #3576, opened 2022-03-03
Fix issues involving scalar set identity/bond of a union

### LINKED ISSUE #3525: Some sort of scoping error with a non-object binding
This query works properly:
```
with
  options := {'aneuploid', 'balkanize', 'defenestrate', 'sanguine'},
  pairs := (select Pair filter .word1 = 'hatch' and .similarity = <float64>$sim),
select options filter (
    with opt_pair := (
        select Pair filter .word1 = options and .word2 in pairs.word2),
    select count(opt_pair) = count(distinct opt_pair.similarity)
) limit 1;
```
but this should-be-equivalent query runs for a long time and then returns no data
```
with
  options := {'aneuploid', 'balkanize', 'defenestrate', 'sanguine'},
  word2 := (select Pair filter .word1 = 'hatch' and .similarity = <float64>$sim).word2,
select options filter (
    with opt_pair := (
        select Pair filter .word1 = options and .word2 in word2),
    select count(opt_pair) = count(distinct opt_pair.similarity)
) limit 1;
```

wrapping the reference to `word2` in the filter clause with a select fixes it.

Schema is 
```
module default {
    type Pair {
        required property word1 -> str;
        required property word2 -> str;
        required property similarity -> float64;
        constraint exclusive on ((.word1, .word2));
        index on ((.word1, .word2));
        index on ((.word1, .similarity));
    }
}
```

msullivan:
1. When _get_path_var_in_setop fails, clear the NULL entries
   so that a future call doesn't succeed based on the presence
   of those.
2. Skip ensure_bond_for_expr on scalar view sets. I don't think
   it is needed anymore, and causes performance issues on the
   #3525 test case. (Fixing this also would fix the bug in the
   absence of fix 1).
3. Don't let values relations produce identity aspects. This
   fixes a bug that currently only shows up without normalization
   on so the test for it doesn't *really* actually test it...

Fixes #3525

### DIFF
--- a/edb/pgsql/compiler/pathctx.py
+++ b/edb/pgsql/compiler/pathctx.py
@@ -411,7 +411,26 @@ def _get_path_var_in_setop(
         else:
             all_null = False
 
-    if all_null:
+    # Fail if no subquery had the path or, for scalar identity paths,
+    # if any did not have it.
+    #
+    # We need to do this for scalar identity because scalar identity
+    # is only used for volatility refs, and it is OK if looking it up
+    # fails, because we create a backup volatility ref---but it is
+    # *not* OK for it to succeed and produce NULL in some cases.
+    if all_null or (
+        aspect == 'identity' and optional and not path_id.is_objtype_path()
+    ):
+        # If *none* of the subqueries had it, we have to remove them all
+        # before erroring, lest a future call see them and decide
+        # they really exist.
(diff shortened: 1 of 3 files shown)


elprans (approved):

msullivan:
Cherry-picked for 1.2

PR #3585, opened 2022-03-07
Fix card inference bug and support chained cardinality restriction

### LINKED ISSUE #3566: Cardinality Inference on cascading
- EdgeDB Version: 1.0
- OS Version: macOS Monterey 12.0.1

Steps to Reproduce:

1.I got a problem about cardinality Inference
image

- The red condition causes cardinality exception;
- The blue condition works

2.  related schema:
```  
create abstract type smartlist { 
    create required property name -> str;
    create property desc -> str; 
  };
  
  create type sml_cf_usage extending smartlist {
    alter property name {
      create constraint exclusive; 
    }; 
  }; 

  create type lease_asset { 
    create property sys_asset_id -> uuid; 
    create required property asset_no -> str { 
      create constraint exclusive; 
    };
    create property area -> decimal { 
      create constraint min_value(0); 
    } ;
    create property _create_time -> cal::local_datetime; 
    create property _modify_time -> cal::local_datetime; 
     
    create link entity -> dim_entity; 
    create link _creator -> user; 
    create link _modifier -> user; 
  };

  create type lease_cf_usage {
    create property lease_sdate -> cal::local_datetime; 
    create property lease_edate -> cal::local_datetime; 
    create property total_amount -> decimal; 
    
    create link cf_usage -> sml_cf_usage; 
    create required link asset -> lease_asset { 
    } ;
    
    create constraint exclusive on ((.cf_usage, .asset));
  } ;

  create type lease_cf_payment { 
    create property cf_id -> uuid; 
    create property cf_date -> cal::local_datetime; 
    create property eff_sdate -> cal::local_datetime; 
    create property eff_edate -> cal::local_datetime; 
    create property cf_amount -> decimal; 
    create property cf_amount_calc -> decimal; 
    
    create link cf_type -> sml_cf_type; 
    create link cf_status -> sml_cf_status; 
    create required link lease_cf_usage -> lease_cf_usage { 
    } ;
  } ;

```

Cardinality Inference on cascading has some problem.

msullivan:
There is a soundness bug in card inference where we only look at the
rightmost element in the path. So if a filter is on `.foo.name`, we
check if `name` has an exclusive constraint *on the subject object*,
which is wrong.

To avoid this breaking code that relied on this in an unsound way
but could be sound, this is paired with implementation of chained
cardinality restriction on filters.

That is, make `select Foo filter .bar.baz = 'key'` infer cardinality
`AtMostOne` if `bar` and `baz` both have exclusive constraints.

Previously, it would have been necessary to write
`select Foo filter .bar = (select Bar filter .baz = 'key')` if you wanted
cardinality One.

Closes #3566.

### DIFF
--- a/edb/edgeql/compiler/inference/cardinality.py
+++ b/edb/edgeql/compiler/inference/cardinality.py
@@ -839,6 +839,7 @@ def _is_ptr_or_self_ref(
                 env.set_types[ir_set] == srccls
                 or (
                     (rptr := ir_set.rptr) is not None
+                    and env.set_types[rptr.source] == srccls
                     and srccls.maybe_get_ptr(
                         env.schema,
                         rptr.ptrref.shortname.get_local_name()
--- a/edb/edgeql/compiler/inference/cardinality.py
+++ b/edb/edgeql/compiler/inference/cardinality.py
@@ -875,7 +875,14 @@ def extract_filters(
             if op_card.is_multi():
                 pass
 
-            elif _is_ptr_or_self_ref(left, result_set, env):
+            elif (
+                (left_matches := _is_ptr_or_self_ref(left, result_set, env))
+                or _is_ptr_or_self_ref(right, result_set, env)
+            ):
+                # If the match was on the right, flip the args
+                if not left_matches:
+                    left, right = right, left
+
                 if infer_cardinality(
                     right, scope_tree=scope_tree, ctx=ctx,
                 ).is_single():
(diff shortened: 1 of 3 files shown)


msullivan:
Oh, though it now occurs to me that improving cardinality inference (to view more things as single) is also a compatibility hazard, since it changes the output shape...

elprans (approved):

elprans:
It is, but not fixing this is also bad.

msullivan:
Cherry-picked for 1.2

PR #3586, opened 2022-03-07
Add `collapsible` directive to docs

jaclarke:


docs/guides/deployment/aws_aurora_ecs.rst
```diff
@@ -87,6 +87,8 @@ your terminal:
    /ec2/v2/home#NIC:search=ec2-security-group
 
 
+.. collapsible::
```
1st1, line 90: @jaclarke will we change this to `:edb-collapsed`?

1st1:
@jaclarke James, I think we agreed to not make this a directive, but instead make it a role attached to the title -- `:edb-collapsed:`.

jaclarke:
Ok, I misunderstood, I thought the comment was about changing the name. Though what's the difference with using a role here instead? Currently the directive is effectively just an alias for `.. class:: _collapsible`, which just adds a class to the section, the same as a role would right?

1st1:
> Though what's the difference with using a role here instead?

I thought that using a directive would require an extra nesting... but it looks like it doesn't.

Can you add a unittest?

1st1:
> I thought that using a directive would require an extra nesting... but it looks like it doesn't.

Can you put a single paragraph in this directive?

colinhacks:
I'd like to land this since it would be useful in some of the guides I'm writing. But I don't think the current version is quite what I need.

It seems like everything below ``.. edb-collapsed::`` gets added to the collapsible block? It would be nice to be able to add collapsible blocks in the middle of a post. I got this to work with `code-block` like so:

```
.. code-block::
   :class: _collapsible

  RANDOM PARAGRSPH!
  RANDOM PARAGRSPH!
  RANDOM PARAGRSPH!
  RANDOM PARAGRSPH!
  RANDOM PARAGRSPH!
```

But I couldn't figure out how to render a generic collapsible block:

```
outer para 

.. collapsible::

  random paragraph

  .. code-block::
  
    inner code block

  another paragraph

outer para
```

Plus, I'm not a big fan of the design, with the fade-out and the "Show more" button that only shows up if the element is sufficiently tall. 

Screen Shot 2022-05-13 at 12 51 21 PM

I'd prefer to have a straightforward `details/summary` implementation with the triangular arrow, like they have on the Fly docs. 

Screen Shot 2022-05-13 at 12 47 45 PM

Though perhaps this should be a separate directive?

1st1:
@jaclarke can you please rebase this? I think we need to get it merged :)

PR #3587, opened 2022-03-07
Fix recursive computed ordering issue.

### LINKED ISSUE #3499: Filter on computed link on computed link
### Discussed in https://github.com/edgedb/edgedb/discussions/3498

<div type='discussions-op-text'>

<sup>Originally posted by **tchak** February 15, 2022</sup>
I have the following schema:

```
module default {
  type Version {
    multi link fields -> Field {
      property position -> int32 {
        default := 0;
      }
    }
    multi link ordered_fields := (select .fields order by @position);
    multi link sections := (select .ordered_fields filter .type = 'section');
  }

  type Field {
    required property type -> str;
    multi link versions := .<fields[is Version];
  }
}
```

I am getting the following error on migrate:

```
error: object type 'default::Field' has no link or property 'type'
    ┌─ ./dbschema/default.esdl:174:59
    │
174 │     multi link sections := (select .ordered_fields filter .type = 'section');
    │                                                           ^^^^^ error
```

I don't know if I am doing something wrong or if this is a bug. I have the same issue with a filter on an already filtered computed. I am also pretty sure I have something similar working on single computed links.</div>

vpetrovykh:
Recursively infer the targets of computed links and properties when
ordering SDL.

Fix #3499

### DIFF
--- a/edb/edgeql/declarative.py
+++ b/edb/edgeql/declarative.py
@@ -40,6 +40,7 @@ from edb import errors
 from edb.common import parsing
 from edb.common import topological
 from edb.common import english
+from edb.common.ordered import OrderedSet
 
 from edb.edgeql import ast as qlast
 from edb.edgeql import codegen as qlcodegen
@@ -402,6 +403,12 @@ def sdl_to_ddl(
         for decl_ast in declarations:
             trace_dependencies(decl_ast, ctx=tracectx)
 
+    # Before sorting normalize all ordering, to make sure that errors
+    # are consistent.
+    for ddlentry in ddlgraph.values():
+        ddlentry.deps = OrderedSet(sorted(ddlentry.deps))
+        ddlentry.weak_deps = OrderedSet(sorted(ddlentry.weak_deps))
+
     try:
         ordered = topological.sort(ddlgraph, allow_unresolved=False)
     except topological.CycleError as e:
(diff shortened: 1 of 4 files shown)


edb/edgeql/declarative.py
```diff
@@ -1078,6 +1089,14 @@ def _get_pointer_deps(
     if pointer in ctx.objects:
         result.add(pointer)
 
+    # HACK: Add all pointers that have this pointer (link, actually)
```
msullivan, line 1092: I'm a bit confused here; why doesn't the property dependency get picked up normally by tracing?

msullivan (approved):

edb/edgeql/declarative.py
vpetrovykh: To pick up a link property we need the tracer to be aware when the context of the expression is a path that denotes a link, currently the tracer only concerns itself with the link target though.

Ideally we'd want the tracer to be better, but as I said in my overall comment this might be a good reason to change the entire approach rather than try to teach the existing tracer about the link properties. The existing tracer is trying to get things approximately right with less deep analysis.

msullivan: Oh, roger, got it.

msullivan:
Cherry-picked to 1.2

PR #3596, opened 2022-03-10
Use path_id instead of type at the root of filter inference

### LINKED ISSUE #3590: Incorrect cardinality inference on update
A cardinality inference bug in `update` queries. Using the [query builder schema](https://github.com/edgedb/edgedb-js/blob/master/qb/dbschema/default.esdl) the following query has inferred cardinality `Many`:

```
WITH
  __scope_0_Hero := DETACHED default::Hero
UPDATE __scope_0_Hero
FILTER (__scope_0_Hero.name = "Spider-Man")
SET {
  name := ("The Amazing " ++ __scope_0_Hero.name)
}
```

Though `Hero.name` has an exclusive constraint. Regression possibly introduced in https://github.com/edgedb/edgedb/pull/3585.

msullivan:
This fixes some false positives and false negatives.
Fixes #3590.

### DIFF
--- a/edb/edgeql/compiler/inference/cardinality.py
+++ b/edb/edgeql/compiler/inference/cardinality.py
@@ -835,7 +835,7 @@ def _is_ptr_or_self_ref(
         return (
             isinstance(srccls, s_objtypes.ObjectType)
             and (
-                env.set_types[ir_set] == srccls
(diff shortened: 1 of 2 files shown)


elprans (approved):

msullivan:
Cherry-picked to 1.2

PR #3603, opened 2022-03-11
Don't enforce exclusivity of non-exclusive obj constraints in UPDATE

msullivan:
In the case that was reported, also, this bug only triggered
because of two suboptimal codegen issues:
  * We generate the conflict checks even when there
    is a single update of a type without children
  * We generate conflict checks for a single UPDATE on a
    type without subtypes.

Those are worth following up on. (The first one is marked with a TODO
already, but I think the second one is just an oversight.)


elprans (approved):

msullivan:
Cherry-picked to 1.2

PR #3605, opened 2022-03-11
Don't generate self-update conflict code if the type has no children

msullivan:


elprans (approved):

msullivan:
I'm planning to cherry-pick this. @elprans, do you agree?

elprans:
Go for it.

msullivan:
Cherry-picked to 1.2, though without the test

PR #3607, opened 2022-03-11
Only emit UPDATE conflict checks when a relevant pointer is modified

msullivan:
That is, only when the a pointer mentioned in an exclusive constraint
is referenced.


elprans (approved):

msullivan:
Cherry picked to 1.2

PR #3608, opened 2022-03-12
Disable exclusive triggers when there aren't any other types in play

msullivan:


tests/test_edgeql_update.py
```diff
@@ -3419,6 +3419,16 @@ async def test_edgeql_update_cardinality_assertion(self):
                 }
             ''')
 
+    @test.xfail('''
```
msullivan, line 3422: We need to decide if this behavior change is unacceptable, or find some way around it.

elprans: Well, we haven't really defined the behavior either way, so I think it's fine if the behavior changes like so.

elprans (approved):

tests/test_edgeql_update.py
msullivan: The catch is that the behavior then depends on whether or not there is a subtype, which is kind of hinky. Not necessarily disqualifying, though, since it's such an edge case. (And you could always opt in to the trigger-based behavior by creating a dummy subtype, if you somehow cared.)

elprans: How hard would it be to repurpose the `unless conflict` machinery to handle constraints also and get rid of triggers altogether?

msullivan: Currently, when used to raise errors, the `conflict` machinery makes sure to exclude all the data from the base relations from the queries it is doing. Before I did this, I think it was causing a 25% slowdown in the test suite, or some such.

That said, that perf trouble might be fixable, though it's a question mark. It would probably take a couple days to get working?

Would the idea be to try to always implement the "no trigger" style behavior?

elprans: Getting rid of triggers would be ideal, yeah.

msullivan: That's not what I mean with the question about the desired behavior, though. There are two possible semantics here:
 * Whether to raise an error is based on comparing to the state at the *start* of the statement. That is, deleting something and then adding something with the same key *still errors*. That's the semantics you get when using triggers.
 * The reverse, where if you delete something and then add something with the same key, it *works*. That's the semantics you get when triggers are off but there is a constraint on the table (it seems).

We would need to make a choice for which to implement

elprans: Well, technically there isn't any causality between the statements, because the order of execution is undefined (unless you explicitly correlate the statements), so Option 1 seems to make more sense.

msullivan:
Cherry-picked to 1.2

tests/test_edgeql_update.py
msullivan: We talked out of band and decided that for now, the behavior here is unspecified, and we can revisit it later. It's pretty marginal.

PR #3613, opened 2022-03-15
Enforce newly created exclusive constraints across tables

msullivan:
And on rebases also.

In order to make it work for constraints on abstract links, abstract link views need
to be fixed to include `source`, which is a bit hacky.

That is a change to the pgsql schema, but we can avoid any trouble from that by
forcing a view update whenever we are generating code that might reference
the `source` of an abstract link view. I tested that bit manually.

### DIFF
--- a/edb/pgsql/delta.py
+++ b/edb/pgsql/delta.py
@@ -219,6 +219,31 @@ class MetaCommand(sd.Command, metaclass=CommandMeta):
             raise AssertionError(f"there is no {ctxcls} in context stack")
         ctx.op.inhview_updates.add(source)
 
+    def schedule_post_inhview_update_command(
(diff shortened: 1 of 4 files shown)


edb/pgsql/delta.py
```diff
@@ -1768,6 +1812,20 @@ def apply(
             constraint, schema, context, self.source_context)
         self.pgops.add(op)
 
+        # If the constraint is being added to existing data,
+        # we need to enforce it on the existing data. (This only
+        # matters for
```
elprans, line 1817: Looks like truncated comment

elprans (approved):

msullivan:
Cherry picked to 1.2

PR #3616, opened 2022-03-16
Properly track modaliases inside transactions

### LINKED ISSUE #3604: `set module` is not fully rolled back with compiled query
- EdgeDB Version: 2.0-dev.6420+local
- OS Version: macOS

Steps to Reproduce:

```
edgedb> start transaction;
OK: START TRANSACTION
edgedb[tx]> set module t2;
OK: SET ALIAS
edgedb[tx]> select AAA {};
{}
edgedb[tx]> rollback;
OK: ROLLBACK TRANSACTION
edgedb> select AAA {};
{}
```

The final `select` is expected to result like this:

```
edgedb> select AAA {a:=1};
error: InvalidReferenceError: object type or alias 'default::AAA' does not exist
  ┌─ query:1:8
  │
1 │ select AAA {a:=1};
  │        ^^^ error
```

Schema:

```
module t2 {
    type AAA {}
}
```

msullivan:
This was causing us to use invalid cache keys.

Also, read from and write to the same cache when in txes with SETs.

Fixes #3604.

### DIFF
--- a/edb/server/compiler/compiler.py
+++ b/edb/server/compiler/compiler.py
@@ -1877,8 +1877,7 @@ class Compiler:
                 if comp.requires_restart:
                     unit.config_requires_restart = True
 
-                if ctx.state.current_tx().is_implicit():
-                    unit.modaliases = ctx.state.current_tx().get_modaliases()
+                unit.modaliases = ctx.state.current_tx().get_modaliases()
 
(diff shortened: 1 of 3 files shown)


msullivan:
Cherry-picked to 1.2

PR #3618, opened 2022-03-16
Initial draft of wasm server integration

tailhook:
This is as simple as it can get:
1. Has single version of extension without any configuration
2. Relies on wasm server listening on `$runstate_dir/.s.wasm_ext`
3. Uses unix socket with one connection per request (similarly to SCGI)
4. Uses Pythonic Pickle to serialize request/response

Technically it works, but doesn't make too much sense to integrate right now. Anyone having access to [edgedb/edgedb-wasm](https://github.com/edgedb/edgedb-wasm) repo (still private) can play with it though.

### DIFF
--- a/edb/wasm/wasm_ext.py
+++ b/edb/wasm/wasm_ext.py
@@ -1,0 +1,80 @@
+import logging
+import asyncio
+import http
+import json
+import pickle
+
+from functools import partial
+
+from edb import errors
+from edb.server.protocol import protocol
+from edb.server.dbview import dbview
+#from edb.server import server
+
+log = logging.getLogger(__name__)
+
+
+def handle_error(
+    request: protocol.HttpRequest,
+    response: protocol.HttpResponse,
+    error: Exception,
+):
+    er_type = type(error)
+    if not issubclass(er_type, errors.EdgeDBError):
+        er_type = errors.InternalServerError
+
+    # TODO(tailhook) figure out whether we want to expose error message here
+    # for arbitrary errors
+    response.body = json.dumps({
+        'kind': 'error',
+        'error': {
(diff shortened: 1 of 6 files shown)


edb/wasm/wasm_ext.py
```diff
@@ -0,0 +1,116 @@
...
+            'type': er_type.__name__,
+        }
+    }).encode()
+    response.content_type = b'application/json'
+    response.status = http.HTTPStatus.INTERNAL_SERVER_ERROR
+    response.close_connection = True
+
+
+async def rpc_request(  # type: ignore  # match statement
+    server,  # : server.Server,  # circular import
```
elprans, line 47: Use `if typing.TYPE_CHECKING:` to guard imports for type annotations.

elprans:
Can we add some basic tests here?

edb/wasm/wasm_ext.py
tailhook:
```python
if typing.TYPE_CHECKING:
    from edb.server import server
```
Yields this error:
```
    edb/server/protocol/protocol.pyx, line 36, in init edb.server.protocol.protocol
        > from edb.wasm import wasm_ext
    /work/edb/wasm/wasm_ext.py, line 52, in <module>
        > server: server.Server,

NameError: name 'server' is not defined
```

This doesn't help. Is this because of import from cython?

tailhook: Ah, right. Forgot about that :(

tailhook:
This is generally ready, but I'm not sure if we want to fix GC issue somewhere else.

tailhook:
Okay, removed that hack.

PR #3621, opened 2022-03-16
Skip state sync if the state in pgcon is identical to dbview

fantix:
Refs #3581


elprans:
Seems like the wrong bug is referenced

fantix:
> Seems like the wrong bug is referenced

Ah, fixed!

edb/server/protocol/binary.pyx
```diff
@@ -1607,10 +1622,14 @@ cdef class EdgeConnection:
             and not query_unit.tx_commit
             and not query_unit.tx_rollback
         ):
-            state = _dbview.serialize_state()
+            orig_state = state = _dbview.serialize_state()
         new_types = None
         conn = await self.get_pgcon()
         try:
+            if conn.last_state is state:
```
elprans, line 1629: Would it make more sense to compare by value instead?

edb/server/protocol/binary.pyx
```diff
@@ -1668,6 +1691,14 @@ cdef class EdgeConnection:
             side_effects = _dbview.on_success(query_unit, new_types)
             if side_effects:
                 self.signal_side_effects(side_effects)
+            if not _dbview.in_tx():
+                state = _dbview.serialize_state()
+                if state is not orig_state:
+                    # In 3 cases the state is changed:
+                    #   1. The non-tx query changed the state
+                    #   2. The state is ynced with dbview (orig_state is None)
```
elprans, line 1699:
```suggestion
                    #   2. The state is synced with dbview (orig_state is None)
```

elprans (approved):
LGTM

PR #3631, opened 2022-03-18
Generate fresh ids for free shapes when possible

msullivan:
Free shapes generated from cross products are not guarenteed to have
unique ids, however. So while
  `for x in User union { asdf := x.name ++ "!", x := .id}`
generates unique ids,
  `select (User, { asdf := User.name ++ "!", x := .id})`
does not.

I'm not sure if there is a good way to avoid this. We ban cross products
of volatile computations for this reason, but doing that to free objects
seems aggressive?

### DIFF
--- a/edb/edgeql/compiler/inference/cardinality.py
+++ b/edb/edgeql/compiler/inference/cardinality.py
@@ -207,9 +207,7 @@ def _common_cardinality(
 
 
 def _is_singleton_type(typeref: irast.TypeRef) -> bool:
-    if typeref.material_type:
-        typeref = typeref.material_type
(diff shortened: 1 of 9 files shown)


edb/edgeql/compiler/inference/cardinality.py
```diff
@@ -207,9 +207,7 @@ def _common_cardinality(
 
 
 def _is_singleton_type(typeref: irast.TypeRef) -> bool:
```
elprans, line 209: Remove the helper altogether?

elprans (approved):

PR #3632, opened 2022-03-18
Fix a multiplicity computation bug for tuples

msullivan:
Currently we track the multiplicity of each element of a tuple
and preserve it, but this doesn't work in general: Consider
`(User, Card)`, where both elements (and the result) have multiplicty
ONE but `(User, Card).1` ought to have multiplicity MANY.

Fix this by taking cardinality and cross products into account
and only preserving element multiplicity for elements that can't
get duplicated.

It might be possible to do better by taking into account correlation
more, though: this will report that (Card.name, Card.cost).name
is MANY.


elprans (approved):

msullivan:
Cherry-picked for 1.3

PR #3633, opened 2022-03-18
Remove the physical FreeObject from the database

msullivan:
In #3631, the compiler was made to bypass the actual table. Now, skip
populating it at all.

### DIFF
--- a/edb/lib/std/60-baseobject.edgeql
+++ b/edb/lib/std/60-baseobject.edgeql
@@ -44,7 +44,7 @@ CREATE ABSTRACT TYPE std::Object EXTENDING std::BaseObject {
         'Root object type for user-defined types';
 };
 
-CREATE ABSTRACT TYPE std::FreeObject EXTENDING std::BaseObject {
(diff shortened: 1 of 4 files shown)


edb/lib/std/60-baseobject.edgeql
```diff
@@ -44,7 +44,7 @@ CREATE ABSTRACT TYPE std::Object EXTENDING std::BaseObject {
         'Root object type for user-defined types';
 };
 
-CREATE TYPE std::FreeObject EXTENDING std::BaseObject {
+CREATE ABSTRACT TYPE std::FreeObject EXTENDING std::BaseObject {
```
elprans, line 47: Hmm, technically `abstract` means that there can be no instance of this type _exactly_, so this qualifier would be a lie.  I think we should instead prohibit `insert FreeObject` explicitly and leave the type non-abstract.

msullivan: Yeah, fair

elprans (approved):

Issue #3583, opened 2022-03-07
Improve backend pool performance


fantix:
## Background

Within a frontend connection (edgecon), the user could set module aliases and session-level configs. These are called the (edgecon) "state" in code and available throughout the lifecycle of the frontend connection. The state takes effect in queries by 1) setting its values into a temporary table `_edgecon_state` in the backend connection (pgcon) used by the frontend connection, and 2) calling `pg_catalog.set_config()` in Postgres.

As a frontend connection may use different backend connections for different queries (when not in a transaction), the frontend connection needs to do a "state sync" before running the query on the backend connection. Currently, this is always done when running a query outside of a transaction, and it's done in a reset-and-recreate way which is taking about 200 microseconds. 

Queries from extensions (like the EdgeQL over HTTP extension) are not currently doing state sync, because they are usually ad-hoc. The module alias is probably not an issue because the compiling step will fail if the query is trying to use an alias only in the state, but the session-level config - if didn't reset - will affect the queries, like the `query_execution_timeout` config.

## Action Items

* Skip state sync if the state in pgcon is identical to dbview
* Change state sync to applying the diff instead of resetting everything
* Pin pgcon to edgecon "softly" outside of a tx if state in dbview is not the default one, so it's more likely to reuse the pgcon with the same state
* Reset session config for queries from extensions

1st1:
> Change state sync to applying the diff instead of resetting everything

I think this is an unnecessary complication. RESET should be fine and most connections won't change the state.

msullivan:
Are we still planning to do any follow ups here?

PR #3640, opened 2022-03-21
Reuse the same pgcon in edgecon if possible

fantix:
Refs #3583


msullivan:
Should this go into a potential 1.3

elprans:
We decided to shelve this for now.

PR #3646, opened 2022-03-21
minor documentation fixes

Marnixvdb:
Three small fixes. Two to make code examples more consistent and one to correct a confusing mistake on the page about modelling relations.

### DIFF
--- a/docs/datamodel/constraints.rst
+++ b/docs/datamodel/constraints.rst
@@ -41,10 +41,10 @@ the length of a string.
   type User {
     required property username -> str {
       # as custom constraint
-      constraint expression on (len(__subject__) < 25);
+      constraint expression on (len(__subject__) <= 25);
 
       # with built-in
-      constraint max_len_value(24);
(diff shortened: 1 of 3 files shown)


docs/datamodel/constraints.rst
```diff
@@ -44,7 +44,7 @@ the length of a string.
       constraint expression on (len(__subject__) < 25);
```
colinhacks, line 44: Can you just change this to `<= 25` so we don't have different numbers?

colinhacks (changes requested):

colinhacks:
Looks great! Left a tiny nitpick, then let's merge

docs/datamodel/constraints.rst
Marnixvdb: Of course! I changed it to <= 25 and updated the max_len_value condition to 25 as well, otherwise they wouldn't match.

colinhacks (approved):

PR #3652, opened 2022-03-22
Fix a bug involving computable shadowing

msullivan:
Because aliased_views is mutable, we have a bug where a
computable property that references the type it is on
can have that name point to a different alias for it when
it gets recompiled.

Since the aliased_views mutability is depended on currently,
fix this by taking a snapshot of aliased_views when doing
process_view.

Also delete some unneeded code that I thought ought to have been
handling this case :P.


elprans (approved):

msullivan:
Cherry-picked for 1.3

PR #3656, opened 2022-03-23
Add an `FOR GROUP` construct and desugaring code from GROUP

msullivan:
The implementation strategy for our nice shape-based `GROUP` is
essentially to implement the original `GROUP ... INTO` proposal
and desugar the real `GROUP` into it.

For development and testing purposes, I want it to be possible to
write tests and dev instance queries using the internal `GROUP`,
so this adds syntax for it under the name `FOR GROUP`.

This allows us to reuse with slight modification most of the tests
that are lying around for the old GROUP, so this includes major
updates to that test suite.

This also implement the desugaring (including some optimizations) from
real GROUP into `FOR GROUP`, and updates the toy model to use the
desugaring.

### DIFF
--- a/edb/edgeql/compiler/options.py
+++ b/edb/edgeql/compiler/options.py
@@ -78,6 +78,12 @@ class GlobalCompilerOptions:
     #: to contain DML in the top-level shape computables.
     allow_top_level_shape_dml: bool = False
 
+    #: Is this a dev instance of the compiler
+    dev_instance: bool = False
+
+    #: Is the compiler running in testmode
+    testmode: bool = False
+
 
 @dataclass
 class CompilerOptions(GlobalCompilerOptions):
--- a/edb/edgeql/compiler/options.py
+++ b/edb/edgeql/compiler/options.py
@@ -79,7 +79,7 @@ class GlobalCompilerOptions:
     allow_top_level_shape_dml: bool = False
 
     #: Is this a dev instance of the compiler
-    dev_instance: bool = False
+    devmode: bool = False
 
     #: Is the compiler running in testmode
     testmode: bool = False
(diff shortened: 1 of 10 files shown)


edb/edgeql/compiler/stmt.py
```diff
@@ -235,6 +235,26 @@ def compile_ForQuery(
...
 
+@dispatch.compile.register(qlast.InternalGroupQuery)
+def compile_InternalGroupQuery(
+    expr: qlast.InternalGroupQuery, *, ctx: context.ContextLevel
+) -> irast.Set:
+    # We want to disallow use of FOR GROUP except for when running
+    # a dev build. Check for the presence of a test schema element
+    # to detect whether we're on a dev build.
+    if not expr.from_desugaring and not ctx.env.schema.get(
+        'cfg::TestSessionConfig', None
```
elprans, line 246: I'd be more rigorous and pass the `__internal_testmode` config value in `CompilerOptions` from `_compile_ql_query` instead.

msullivan: That's a pain for local development, though, because `__internal_testmode` can only be set per-session

elprans: Fine, move the test schema hack up there then.  I'd still like `testmode` to show up as a `CompilerOptions` flag.

msullivan: Yeah good point

edb/edgeql/compiler/options.py
```diff
@@ -78,6 +78,12 @@ class GlobalCompilerOptions:
     #: to contain DML in the top-level shape computables.
     allow_top_level_shape_dml: bool = False
 
+    #: Is this a dev instance of the compiler
+    dev_instance: bool = False
```
elprans, line 82:
```suggestion
    devmode: bool = False
```

elprans (approved):

PR #3662, opened 2022-03-23
Delete some code marked "remove after postgres 11"

msullivan:
The code forced custom scalar array_agg elements to their base
type. Dropping this code broke an array_get test, which revealed a
more general bug issue where forcing polymorphic scalars to their base
type caused trouble with array_get defaults.

This also fixed a bug where arrays of custom scalars could be casted
between even though the actual scalar type could not.

Additionally, removing a now-unused import perturbed the import order
in such a way as to break things, so I sunk an import down into a function

### DIFF
--- a/edb/schema/constraints.py
+++ b/edb/schema/constraints.py
@@ -29,7 +29,6 @@ from edb.edgeql import compiler as qlcompiler
 from edb.edgeql import qltypes as ft
 from edb.edgeql import parser as qlparser
 from edb.edgeql import utils as qlutils
-from edb.schema import scalars as s_scalars
 
 from . import abc as s_abc
 from . import annos as s_anno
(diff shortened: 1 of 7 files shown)


edb/schema/constraints.py
```diff
@@ -671,6 +670,7 @@ def _populate_concrete_constraint_attrs(
         from edb.ir import utils as ir_utils
         from . import pointers as s_pointers
         from . import links as s_links
+        from . import scalars as s_scalars
```
elprans, line 673: This new import seems unused?

elprans: Nevermind.

elprans (approved):

msullivan:
Cherry-picked for 1.3

PR #3665, opened 2022-03-23
Don't emit the contents of array_agg and the like *twice*

msullivan:
Currently we will duplicate an array_agg/array literal in most cases
other than appearing at the very toplevel, once for value and once for
serialized. Avoid this by pretending the serialized output is the value
if we know it is only being used for "not NULL" style things; use eta
expansion to ensure this.

### DIFF
--- a/edb/edgeql/compiler/eta_expand.py
+++ b/edb/edgeql/compiler/eta_expand.py
@@ -118,7 +118,7 @@ def needs_eta_expansion_expr(
     """
     if isinstance(ir, irast.SelectStmt):
         return needs_eta_expansion(
-            ir.result, is_processed=bool(ir.where or ir.orderby), ctx=ctx)
+            ir.result, has_clauses=bool(ir.where or ir.orderby), ctx=ctx)
 
     if isinstance(stype, s_types.Array):
         if isinstance(ir, irast.Array):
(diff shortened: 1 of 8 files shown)


edb/ir/typeutils.py
```diff
@@ -133,6 +133,30 @@ def is_persistent_tuple(typeref: irast.TypeRef) -> bool:
...
 
 
+def contains_predicate(
+    typeref: irast.TypeRef,
+    pred: Callable[[irast.TypeRef], bool],
+) -> bool:
+    if pred(typeref):
+        return True
+
+    if typeref.intersection:
```
elprans, line 143: Would be prettier with `elif`/`else` :-)

edb/edgeql/compiler/eta_expand.py
```diff
@@ -148,6 +149,7 @@ def needs_eta_expansion_expr(
 def needs_eta_expansion(
     ir: irast.Set,
     *,
+    is_processed: bool = False,
```
elprans, line 152: What does `is_processed` mean exactly?

elprans (approved):

edb/edgeql/compiler/eta_expand.py
msullivan: Renamed to `has_clauses`

PR #3667, opened 2022-03-24
Implement GROUP

msullivan:
This is mostly an implementation of `FOR GROUP`, since we already have
code for compiling real `GROUP` to `FOR GROUP`.

The general case materializes the group. As an optimization for common
aggregation cases, we find calls to aggregate functions that reference
the group and compile them separately and included them in the
grouping output.


elprans (approved):
Monumental!  LGTM!

PR #3672, opened 2022-03-24
Rebalance deeply nested `UNION` operators

### LINKED ISSUE #3669: Bulk Inserting Error: maximum recursion depth exceeded
I'm getting the error `edgedb.errors.InternalServerError: maximum recursion depth exceeded while calling a Python object` when I'm bulk-inserting ~50 small objects in one query.

The schema is:
```
  type Person {
    required property first_name -> str;
    property last_name -> str;
    multi property tags -> str;
    property ordered_tags -> array<str>;
    required property inserted_at -> datetime {
        readonly := true;
    }
    property created_at -> datetime;
    multi link people_i_follow -> Person {
        on target delete allow;
    }
    required property slug -> str {
        constraint exclusive;
    }
    multi link my_followers := .<people_i_follow[is Person];
    link best_friend -> Person {
        on target delete allow;
    }
    property full_name :=
      .first_name ++ ' ' ++ .last_name
      if exists .last_name
      else .first_name;
  }
```

And the code creates this query:
```
with
model_1 := (insert Person {first_name := <str>$first_name_1, slug := <str>$slug_1}),
model_2 := (insert Person {first_name := <str>$first_name_2, slug := <str>$slug_2}),
...
model_50 := (insert Person {first_name := <str>$first_name_50, slug := <str>$slug_50}),
models := (model_1 union model_1 ... union model_50)
select models {id}
```

My current workarounds are 1) only bulk inserting a small number of objects at a time ~3, and 2) asynchronously inserting objects individually.

msullivan:
We could do other associative operators if we wanted, like ++. If we
wanted to do +, we'd have to do it after type inference so that we
don't do it on floating point values, where it doesn't hold.

Fixes #3669.

### DIFF
--- a/edb/edgeql/compiler/expr.py
+++ b/edb/edgeql/compiler/expr.py
@@ -71,10 +71,57 @@ def compile_Path(
     return stmt.maybe_add_view(setgen.compile_path(expr, ctx=ctx), ctx=ctx)
 
 
+def _balance(
+    elements: Sequence[qlast.Expr],
+    ctor: Callable[
+        [qlast.Expr, qlast.Expr, Optional[ctx_utils.ParserContext]],
+        qlast.Expr
+    ],
+    context: Optional[ctx_utils.ParserContext],
+) -> qlast.Expr:
+    mid = len(elements) // 2
+    ls, rs = elements[:mid], elements[mid:]
+    ls_context = rs_context = None
+    if len(ls) > 1 and ls[0].context and ls[-1].context:
+        ls_context = ctx_utils.merge_context([
+            ls[0].context, ls[-1].context])
+    if len(rs) > 1 and rs[0].context and rs[-1].context:
+        rs_context = ctx_utils.merge_context([
+            rs[0].context, rs[-1].context])
+
+    return ctor(
+        (
+            _balance(ls, ctor, ls_context)
+            if len(ls) > 1 else ls[0]
+        ),
+        (
+            _balance(rs, ctor, rs_context)
+            if len(rs) > 1 else rs[0]
+        ),
+        context,
+    )
+
+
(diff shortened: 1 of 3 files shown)


edb/edgeql/compiler/expr.py
```diff
@@ -666,3 +692,17 @@ def flatten_set(expr: qlast.Set) -> List[qlast.Expr]:
             elements.append(el)
 
     return elements
+
+
+def collect_binop(expr: qlast.Expr, op: str) -> List[qlast.Expr]:
+    elements = []
+    def _go(expr: qlast.Expr) -> None:
```
elprans, line 699: Perhaps make this into an iterative traversal to further avoid recursion issues?

tests/test_edgeql_select.py
```diff
@@ -6288,6 +6288,34 @@ async def test_edgeql_select_big_set_literal(self):
 
         assert len(res) == 100
 
+    async def test_edgeql_select_big_unions(self):
+        res = await self.con.query("""
+            SELECT (
+                 (1,) union (1,) union (1,) union (1,) union (1,) union
```
1st1, line 6294: Can you add a similar test with `{1, 2, 3, 4, 5, 6, ..., 100}`? That used to crash with a recursive error.

elprans: We already have a test for that.

edb/edgeql/compiler/expr.py
msullivan: I think I had it in my head that there was some pass before this point that would do a full tree traversal, and so the important bit was just dodging the ~10x stack frame blowup in pgsql. But that's wrong, and this was the first thing hit when I made it much bigger

elprans (approved):

PR #3676, opened 2022-03-25
Fix a collection of json casting bugs

### LINKED ISSUE #3671: Casting arrays or tuples from JSON `null` is broken
```
foo> select <array<str>>to_json('null');
edgedb error: InvalidValueError: cannot extract elements from a scalar
```

Expected result: cast should return an appropriately typed empty set.

msullivan:
* Produce empty sets when a toplevel null is cast to an array or tuple
 * Force a NULL check on casts to array<json>, where previously we
   could spuriously produce an empty array instead of an empty set.

Fixing the latter without causing further trouble involved fixing the
exposedness handling of indirections on tuple literals.

Fixes #3671.


elprans (approved):

msullivan:
Cherry picked for 1.x

PR #3682, opened 2022-03-28
Binary protocol channeling over HTTP

1st1:
This is WIP. Pushing it to unblock @jaclarke.

-------

ToDo:

- [ ] Discuss the protocol change with the team; implement it in edgedb-python and edgedb-go.
- [ ] Limit capabilities of compiled queries (prohibit transactions, ddl, session, etc)
- [ ] Think about deprecating the 'execute' message
- [ ] Add proper HTTP path validation.
- [ ] Add proper version negotiation and checks
- [ ] Error out on empty messages; on extra data left in buffer, etc.
- [ ] Implement basic JWT
- [ ] Implement error reporting
- [ ] Add low-level proto tests


edb/server/protocol/protocol.pyx
```diff
@@ -452,6 +463,32 @@ cdef class HttpProtocol:
...
+
+                    out = await binary.eval_buffer(
+                        self.server,
+                        database,
+                        self.current_request.body,
+                    )
+
+                    response.body = out
+                    response.status = http.HTTPStatus.OK
+                    response.content_type = b'application/x.edgedb'
```
elprans, line 480: I'd use something like `application/vnd.edgedb.v019.raw`

edb/server/protocol/studio_ext.pyx
```diff
@@ -0,0 +1,107 @@
...
+import mimetypes
+
+import immutables
+
+from edb import errors
+
+from edb.common import debug
+from edb.common import markup
+
+STATIC_FILES_DIR = 'build/studio'
```
elprans, line 33: Use `get_shared_data_dir_path() / 'studio'` instead to make this work with production builds.

setup.py
```diff
@@ -428,6 +428,39 @@ def _compile_cli(build_base, build_temp):
...
+    # run build
+    env = dict(os.environ)
+    # With CI=true (set in GH CI) `yarn build` fails if there are any
+    # warnings. We don't need this check in our build so we're disabling
+    # this behavior.
+    env['CI'] = ''
+    # env['BUILD_PATH'] = '../../build/studio'
+    subprocess.run(
+        ['yarn', 'build'], check=True, cwd='edgedb-studio/web', env=env)
+    subprocess.run(['rm', '-r', 'build/studio'])
```
elprans, line 460:
```suggestion
    subprocess.run(['rm', '-r', buildmeta.get_shared_data_dir_path() / 'studio'])
```

edb/server/protocol/protocol.pyx
```diff
@@ -452,6 +463,32 @@ cdef class HttpProtocol:
                     self.server,
                 )
                 return
+            if (path_parts[0] == 'admin' and
```
elprans, line 466: I'd prefer this to be `'ui'` instead to match `edgedb ui`, i.e. `https://edgedb.host:port/ui`

edb/server/protocol/protocol.pyx
1st1: `vnd` is for officially registered "vendor" protocols. We have to use `x.` before that happens.

elprans: OK, but I'd still include the version and encoding

elprans (approved):

1st1:
@elprans all fixed, please take a look at the two new commits

edb/server/protocol/protocol.pyx
1st1: I did include that..

1st1:
> Fixup

Oups, sorry. Thanks for fixing.

PR #3685, opened 2022-03-28
Don't reverse the order of large set literals/union chains

msullivan:
The issue here is that the explicit iterative traversal accidentally
went right-to-left instead of left-to-right and that because
SetConstructorOp has an op field that is set to `UNION`, it got
rebalanced a *second* time.

### DIFF
--- a/edb/edgeql/compiler/expr.py
+++ b/edb/edgeql/compiler/expr.py
@@ -200,7 +200,7 @@ def compile_Set(
             bigunion = _balance(
                 elements,
                 lambda l, r, c: qlast.SetConstructorOp(
-                    left=l, right=r, context=c),
+                    left=l, right=r, rebalanced=True, context=c),
                 expr.context
             )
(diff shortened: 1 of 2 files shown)


elprans (approved):

edb/edgeql/compiler/expr.py
```diff
@@ -697,11 +697,11 @@ def flatten_set(expr: qlast.Set) -> List[qlast.Expr]:
 def collect_binop(expr: qlast.BinOp) -> List[qlast.Expr]:
     elements = []
 
-    stack = [expr.left, expr.right]
+    stack = [expr.right, expr.left]
     while stack:
         el = stack.pop()
         if isinstance(el, qlast.BinOp) and el.op == expr.op:
-            stack.extend([el.left, el.right])
+            stack.extend([el.right, el.left])
```
1st1, line 704: What I thought :)

1st1 (approved):

PR #3688, opened 2022-03-28
Fix json -> array<json> cast in stdlib

msullivan:
This bug was fixed in #3676, but via a compiler hack that was easily
cherry-pickable to 1.x. Now we can just fix it in the library.

### DIFF
--- a/edb/lib/std/30-jsonfuncs.edgeql
+++ b/edb/lib/std/30-jsonfuncs.edgeql
@@ -241,12 +241,8 @@ CREATE CAST FROM std::json TO anytuple {
 CREATE CAST FROM std::json TO array<json> {
     SET volatility := 'Immutable';
     USING SQL $$
-    SELECT CASE jsonb_typeof(val)
-    WHEN 'null' THEN NULL
-    ELSE
-        (SELECT array_agg(j)
-         FROM jsonb_array_elements(val) AS j)
-    END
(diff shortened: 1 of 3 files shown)


edb/lib/std/30-jsonfuncs.edgeql
```diff
@@ -241,8 +241,12 @@ CREATE CAST FROM std::json TO anytuple {
 CREATE CAST FROM std::json TO array<json> {
     SET volatility := 'Immutable';
     USING SQL $$
-        SELECT array_agg(j)
-        FROM jsonb_array_elements(val) AS j
+    SELECT CASE jsonb_typeof(val)
```
elprans, line 244: [`NULLIF`](https://www.postgresql.org/docs/current/functions-conditional.html#FUNCTIONS-NULLIF) would probably be faster here.

msullivan: I'm missing the obvious way to to use NULLIF here?

elprans: Try

```
SELECT array_agg(j) 
FROM jsonb_array_elements(nullif(val, 'null'::jsonb)) as j;
```

msullivan: Oh, aha

elprans (approved):

Issue #3679, opened 2022-03-27
datetime_get() cannot be used in index on an expression


fspoettel:
Not sure if this is an issue or working as expected - reported because the CLI advised it.

- EdgeDB Version: 1.2+5aecabc
- OS Version: Linux 5.13.0-35-generic # 40~20.04.1-Ubuntu SMP Mon Mar 7 09:18:32 UTC 2022 x86_64 GNU/Linux

Steps to Reproduce:

1. Create an index on an expression using `datetime_get()`:

```
  type Scrobble {
    required property scrobbled_at -> datetime;
    index on (datetime_get(.scrobbled_at, 'year'));
  }
```

2. Run `edgedb migration create`. This creates:
```
CREATE MIGRATION m1xpvg7vrb3cor7l3uilniureplrjmwloeqbm4jhhd6koyb2ljw5sq
    ONTO m1vs3yy7z3ipjnmg724bdztubd6hvvn2m3hnzuq5hl4vajb2qf23gq
{
  ALTER TYPE default::Scrobble {
      CREATE INDEX ON (std::datetime_get(.scrobbled_at, 'year'));
  };
};
```

3. Run `edgedb migrate`. This returns:
```
edgedb error: InternalServerError: functions in index expression must be marked IMMUTABLE
  Hint: This is most likely a bug in EdgeDB. Please consider opening an issue ticket at https://github.com/edgedb/edgedb/issues/new?template=bug_report.md
  Server traceback:
      edb.errors.InternalServerError: functions in index expression must be marked IMMUTABLE
```

1st1:
@vpetrovykh @msullivan I'm not sure why `datetime_get` isn't immutable; seems like a bug.

elprans:
We should also check volatility properly at EdgeDB level, this shouldn't be a Postgres ISE.

msullivan:
`datetime_get` seems to be legitimately `STABLE` and not `IMMUTABLE`: https://github.com/edgedb/edgedb/blob/v1.2/edb/lib/std/30-datetimefuncs.edgeql#L61-L62

We use timestamptz to represent datetime, and operations on timestamptz seem to be STABLE

This is kind of unfortunate.

elprans:
Well, actually it's only `STABLE` because of timezone configurability in Postgres.  EdgeDB hardcodes in-database timezone to UTC and does not allow changing it, so I _think_ we can get away with a lie about volatility of our wrapper.

msullivan:
[...]

PR #3690, opened 2022-03-28
Check that index expressions are immutable

msullivan:
This requires also having volatility inference consider properties of
objects marked as singletons to be immutable; indexes can't call
STABLE functions but can of course access the table.

Fixes the ISE bug in #3679.


elprans (approved):

msullivan:
Cherry-picked to 1.3

PR #3702, opened 2022-03-31
Parsing/AST for globals

msullivan:


edb/edgeql-rust/src/normalize.rs
```diff
@@ -211,7 +213,11 @@ pub fn normalize<'x>(text: &'x str)
             }
             Kind::Keyword
             if (matches!(&(&tok.value[..].to_uppercase())[..],
-                "CONFIGURE"|"CREATE"|"ALTER"|"DROP"|"START"))
+                         "CONFIGURE"|"CREATE"|"ALTER"|"DROP"|"START")
+                || (last_was_set &&
+                    matches!(&(&tok.value[..].to_uppercase())[..],
```
elprans, line 218: Hmm, any reason why we can't support parameters in `set global`?

msullivan: I don't think there's any fundamental reason, other than that we currently don't support parameters for `SessionStateQuery`. We probably could?

I didn't bother because my understanding is that `SET GLOBAL` is really just for the CLI, and that real usages will be setting globals directly via language bindings.

I can look at supporting parameters, but I don't think it's a blocker for merging stuff

elprans: It's not a blocker, but our long term goal is to add support for parameters to all statements, including DDL, so it would be nice to have support in newly added statements if we can afford that.  Though I think we'd need some preparatory work first.

elprans (approved):

PR #3704, opened 2022-03-31
Fix local temp cluster init to be compatible with the latest edgedb cli

1st1:

PR #3705, opened 2022-03-31
Pin the version of edgedb-cli that we use

msullivan:
This will ensure that edgedb and the cli stay in sync, require a test
suite run for cli updates, and make it possible to bisect beyond
cli/server interface breakages.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -546,7 +546,7 @@ class ci_helper(setuptools.Command):
             print(binascii.hexlify(ext_hash).decode())
 
         elif self.type == 'cli':
-            print(_get_edgedbcli_rev())
+            print(EDGEDBCLI_COMMIT)
 
         elif self.type == 'build_temp':
             print(pathlib.Path(build.build_temp).resolve())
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -114,6 +114,7 @@
 RUST_VERSION = '1.53.0'  # Also update docs/internal/dev.rst
 
 EDGEDBCLI_REPO = 'https://github.com/edgedb/edgedb-cli'
+EDGEDBCLI_COMMIT = '87cb7e08452f37b97e95475df43006fa6131c0bf'
```
1st1, line 117: Should we use tags instead of commits?

msullivan:
> Should be merged after #3704

This currently pins to the previous cli version, so as written it is designed to be merged *before*, and for #3704 to be made to update the pin.

The test failures were because I missed updating the code that *CI* uses to figure out the cli version, which is fixed now.

(That said, if #3704 is made to go green before this, we should merge that first and I'll update the pin on this one)

setup.py
msullivan: Done

elprans (approved):

1st1 (approved):

PR #3718, opened 2022-04-05
Add `make studio` target for convenience

elprans:


1st1 (approved):

1st1:
Wait, call it `make ui`

1st1:
We should also rename `build_studio` to `build_ui`

elprans:
I'll open a separate PR with renames.

PR #3719, opened 2022-04-05
Revert "Pin the version of edgedb-cli that we use (#3705)"

elprans:
The consensus is that we should strive to keep CLI compatibility with
old servers, including the internal development/test features (possibly
in deprecated status).

This reverts commit 902a02cc3699ca180f090cb1e75f28af0bff3111.


msullivan (changes requested):
I'd like to keep the mechanism for pinning, so that we can use it to temporarily fix the `edgedb` build when necessary. I'll send up an alternate proposal PR

PR #3700, opened 2022-03-31
Fix a collection of nested shape path reference issues

msullivan:
* When computing potentially visible sets from a computable, don't
   skip the source if the reference to the source occurs outside the
   scope of the computable. This fixes cases where the source is used
   in a WITH binding outside the shape, and allows those to be
   properly materialized. Fixes #3678.
 * When remapping paths to computed sources in compile_path,
   don't copy the remapped set's expr/rptr. They shouldn't be
   needed and this could result in a TypeIntersection rptr
   on a set with an expr, which broke card inference when
   the source path had a type intersection.
 * Suppress recompiling a computable when the path has been
   remapped. The computable must already be visible, and it
   was breaking things on the pgsql side.

This cleans up most of the leftover problems from #3261, I think.

PR #3724, opened 2022-04-05
Mild rework of new_set_from_set

msullivan:
* Distinguish between None and "keep the old value", for
   fields where None is valid. One of the issues in #3700
   was partially caused by a passed in None causing a value
   to be kept (though it that case the best fix was to *always*
   keep the old value).
 * Make `preserve_scope_ns` the default behavior by replacing it
   with a new argument, `merge_current_ns`, that has inverted
   behavior and is also False by default.
   `preserve_scope_ns=True` was passed by every call site but two,
   and only one call site used its default value.

### DIFF
--- a/edb/edgeql/compiler/setgen.py
+++ b/edb/edgeql/compiler/setgen.py
@@ -25,6 +25,7 @@ from __future__ import annotations
 from typing import *
 
 import contextlib
+import enum
 
 from edb import errors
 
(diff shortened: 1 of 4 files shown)


edb/edgeql/compiler/setgen.py
```diff
@@ -133,32 +133,36 @@ def get_set_type(
 
 def new_set_from_set(
         ir_set: irast.Set, *,
-        preserve_scope_ns: bool=False,
+        merge_current_ns: bool=False,
+        path_scope_id: Optional[int | Literal['']]='',
```
elprans, line 137: Not sure I like `''` as a sentinel.  Maybe use an explicit singleton (a-la `NoDefault`)?

msullivan: Sure

elprans (approved):

PR #3735, opened 2022-04-07
Make schema::Type.from_alias include generated target types

msullivan:
Aliases with nested shapes need to create schema types for
those views, and it gives the names like `__UserAlias__deck`.

These are types that have `is_view` as True inside the compiler
but don't have an `expr` field, so this fact isn't exposed
to introspection at all.

An alternative would be to add a new field (`is_view`, perhaps)
instead of changing `from_alias`, but I think that having
`from_alias` be true on the target types generated from alias
definitions makes sense.

Needed to fix edgedb/edgedb-studio#13.

### DIFF
--- a/edb/lib/schema.edgeql
+++ b/edb/lib/schema.edgeql
@@ -76,7 +76,8 @@ CREATE TYPE schema::PseudoType EXTENDING schema::Type;
 
 ALTER TYPE schema::Type {
     CREATE PROPERTY expr -> std::str;
-    CREATE PROPERTY from_alias := EXISTS(.expr);
+    CREATE PROPERTY expr_type -> int64;
+    CREATE PROPERTY from_alias := EXISTS(.expr_type);
     # Backwards compatibility.
     CREATE PROPERTY is_from_alias := .from_alias;
 };
--- a/edb/lib/schema.edgeql
+++ b/edb/lib/schema.edgeql
@@ -76,8 +76,7 @@ CREATE TYPE schema::PseudoType EXTENDING schema::Type;
 
 ALTER TYPE schema::Type {
     CREATE PROPERTY expr -> std::str;
-    CREATE PROPERTY expr_type -> int64;
-    CREATE PROPERTY from_alias := EXISTS(.expr_type);
+    CREATE PROPERTY from_alias -> bool;
     # Backwards compatibility.
     CREATE PROPERTY is_from_alias := .from_alias;
 };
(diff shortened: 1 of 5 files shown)


edb/lib/schema.edgeql
```diff
@@ -76,7 +76,8 @@ CREATE TYPE schema::PseudoType EXTENDING schema::Type;
 
 ALTER TYPE schema::Type {
     CREATE PROPERTY expr -> std::str;
-    CREATE PROPERTY from_alias := EXISTS(.expr);
+    CREATE PROPERTY expr_type -> int64;
```
elprans, line 79: Do we really need to expose `expr_type`?

msullivan: That is where `is_view` is computed from in the server code, so it's the main source of truth for this. 

I think it would also work to expose `alias_is_persistent`?

Or we could add a new `Field`

1st1: What's `expr_type`? Why is it an int?

msullivan: It's an enum that indicates what kind of statement (select, insert, update, etc) a type is based on. The actual value is not really important here, but its presence is currently the main source of truth for whether a type is a view or not.

I'm going to rename `alias_is_persistent` to `is_persistent_alias` and expose that instead

msullivan: Or maybe just add a `from_alias` field?

1st1: Can we call it `is_alias`?

elprans: We don't use `is_` prefixes in schema anymore.

msullivan:
Alright I just made it a real field

1st1:
Can you add a test?

msullivan:
> Can you add a test?

Sure

edb/schema/types.py
```diff
@@ -104,6 +104,13 @@ class Type(
         ExprType,
         default=None, compcoef=0.909)
 
+    # True for views.  This should always match the value of
+    # `bool(expr_type)`, but can be exported in the introspection
+    # schema without revealing weird internals.
+    from_alias = so.SchemaField(
+        bool,
+        default=False, compcoef=0.909)
```
elprans, line 112: Since this is a derivative of `expr_type` should we make `compcoef=1`?

elprans (approved):

edb/schema/types.py
msullivan: Yeah, sure, that makes more sense.

(tbh I'm not sure that the compcoef of `expr_type` makes sense anyway?)

PR #3725, opened 2022-04-05
Add node and yarn to dev dependencies

fmoor:

PR #3738, opened 2022-04-07
Require python 3.10

msullivan:


docs/reference/dev.rst
```diff
@@ -18,7 +18,7 @@ Linux or macOS.  Windows is not currently supported.
 * C compiler (GCC or clang);
 * Rust compiler and Cargo 1.53 or later;
 * autotools;
-* Python 3.9 dev package;
+* Python 3.10 dev package;
```
1st1, line 21: Oh, while here, can you add one more commit mentioning the requirement for yarn/node?

1st1 (approved):

docs/reference/dev.rst
msullivan: That was already in the pending #3725, which I just went and merged

PR #3739, opened 2022-04-07
Make CONFIGURE INSTANCE work in multi-statement execute blocks

msullivan:
We have machinery for splitting out queries that need to be executed
on their own instead of grouped together (`single_unit`), so extend it
to work for CONFIGURE commands and mark CONFIGURE INSTANCE as
single_unit.

This isn't particularly important, but it is the same infrastructure
that will make SET GLOBAL work in multi-statement execute blocks,
and the error message was wrong in any case.

### DIFF
--- a/edb/server/compiler/compiler.py
+++ b/edb/server/compiler/compiler.py
@@ -1567,6 +1567,7 @@ class Compiler:
 
         sql = (sql_text.encode(),)
 
+        single_unit = False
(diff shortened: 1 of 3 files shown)


elprans (commented):
I'm not sure this is a sound change. Semantically we expect `con.execute('block')` to be atomic, i.e. an implicit transaction.

PR #3743, opened 2022-04-11
Implement DDL commands for globals

msullivan:
This is all the infrastructure needed for non-computed globals that
can't be used inside functions yet.

### DIFF
--- a/edb/schema/globals.py
+++ b/edb/schema/globals.py
@@ -119,7 +119,7 @@ class GlobalCommand(
 
         if isinstance(target_ref, s_pointers.ComputableRef):
             raise errors.SchemaDefinitionError(
-                "computable globals are not yet implemented",
+                "computed globals are not yet implemented",
                 context=self.source_context,
             )
         elif (self.get_local_attribute_value('expr')) is not None:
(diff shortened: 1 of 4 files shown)


edb/schema/globals.py
```diff
@@ -80,3 +90,289 @@ class Global(
...
+    ) -> s_schema.Schema:
+        from . import pointers as s_pointers
+
+        schema = super().canonicalize_attributes(schema, context)
+        target_ref = self.get_local_attribute_value('target')
+        inf_target_ref: Optional[s_types.TypeShell[s_types.Type]]
+
+        if isinstance(target_ref, s_pointers.ComputableRef):
+            raise errors.SchemaDefinitionError(
+                "computable globals are not yet implemented",
```
elprans, line 122:
```suggestion
                "computed globals are not yet implemented",
```

edb/schema/globals.py
```diff
@@ -80,3 +90,289 @@ class Global(
...
+                context=self.source_context,
+            )
+        elif (self.get_local_attribute_value('expr')) is not None:
+            raise errors.SchemaDefinitionError(
+                "computable globals are not yet implemented",
```
elprans, line 127:
```suggestion
                "computed globals are not yet implemented",
```

elprans (approved):

PR #3744, opened 2022-04-11
Implement storage and retrieval of global variables

msullivan:
Split into a few commits.

### DIFF
--- a/edb/lib/std/30-jsonfuncs.edgeql
+++ b/edb/lib/std/30-jsonfuncs.edgeql
@@ -74,7 +74,6 @@ std::json_get(
     $$;
 };
 
-
 CREATE INFIX OPERATOR
 std::`=` (l: std::json, r: std::json) -> std::bool {
     CREATE ANNOTATION std::identifier := 'eq';
--- a/edb/pgsql/compiler/clauses.py
+++ b/edb/pgsql/compiler/clauses.py
@@ -382,6 +382,7 @@ def fini_toplevel(
 
 def populate_argmap(
     params: List[irast.Param],
+    globals: List[irast.Global],
     *,
     ctx: context.CompilerContextLevel,
 ) -> None:
@@ -400,3 +401,13 @@ def populate_argmap(
             index=index,
             required=param.required,
         )
+    for param in globals:
+        ctx.argmap[param.name] = pgast.Param(
+            index=len(ctx.argmap) + 1,
+            required=param.required,
+        )
+        if param.has_present_arg:
+            ctx.argmap[param.name + "present__"] = pgast.Param(
+                index=len(ctx.argmap) + 1,
+                required=True,
+            )
--- a/edb/server/compiler/compiler.py
+++ b/edb/server/compiler/compiler.py
@@ -696,6 +696,13 @@ class Compiler:
                         array_type_id=array_tid,
                     )
 
+            globals = None
+            if ir.globals:
+                globals = [
+                    (str(glob.global_name), glob.has_present_arg)
+                    for glob in ir.globals
+                ]
+
             if ctx.protocol_version >= (0, 12):
                 in_type_data, in_type_id = \
                     sertypes.TypeSerializer.describe_params(
@@ -733,6 +740,7 @@ class Compiler:
                 sql=(sql_bytes,),
                 sql_hash=sql_hash,
                 cardinality=result_cardinality,
+                globals=globals,
                 in_type_id=in_type_id.bytes,
                 in_type_data=in_type_data,
                 in_type_args=in_type_args,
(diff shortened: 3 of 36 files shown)


1st1:
Is this ready for review?

msullivan:
> Is this ready for review?

Yes

edb/lib/std/30-jsonfuncs.edgeql
```diff
@@ -74,6 +74,16 @@ std::json_get(
     $$;
 };
 
+CREATE FUNCTION
+std::json_opt(json: OPTIONAL std::json) -> std::json
```
elprans, line 78: Hm, not sure this needs to be a public function.  Why not just use `?? <json>'null'`?

msullivan: It seemed like a useful enough function, and I was getting squeamish about the amount of code we were generating, but it's probably fine to just do `??`

elprans: scalar `??` shouldn't be too bad?

elprans (approved):
LGTM, but I'd like @1st1 to take a quick look at state/proto changes as well.

edb/lib/std/30-jsonfuncs.edgeql
1st1: Yeah, I'm also against adding sugar functions. Let's get rid of it.

edb/pgsql/compiler/clauses.py
```diff
@@ -400,3 +401,13 @@ def populate_argmap(
             index=index,
             required=param.required,
         )
+    for param in globals:
+        ctx.argmap[param.name] = pgast.Param(
+            index=len(ctx.argmap) + 1,
+            required=param.required,
+        )
+        if param.has_present_arg:
+            ctx.argmap[param.name + "_present"] = pgast.Param(
```
1st1, line 410: better use `f'__{param_name}_present__'` -- we ban `__` names specifically to be able to avoid collisions in cases like this.

edb/server/compiler/compiler.py
```diff
@@ -1878,6 +1895,14 @@ def _try_compile(
                             'transaction block')
 
                     unit.system_config = True
+                elif comp.config_scope is qltypes.ConfigScope.GLOBAL:
+                    unit.system_config = True
+                    # FIXME: We want to be able to
```
1st1, line 1900: ?

edb/server/compiler/compiler.py
```diff
@@ -1878,6 +1895,14 @@ def _try_compile(
...
+                    if statements_len > 1:
+                        raise errors.QueryError(
```
1st1, line 1902:
```suggestion
                        raise errors.UnsupportedFeatureError(
```

edb/server/config/ops.py
```diff
@@ -68,15 +69,25 @@ class Operation(NamedTuple):
...
+                return None
+            else:
+                raise errors.ConfigurationError(
+                    f'unknown setting {self.setting_name!r}') from None
 
-    def coerce_value(self, setting: spec.Setting, *,
+    def coerce_value(self, setting: Optional[spec.Setting], *,
                      allow_missing: bool = False):
+        # A missing setting indicates that this is a global, not a setting
+        if not setting:
```
1st1, line 85: This smells. Do we have to reuse this machinery for globals?

edb/server/protocol/binary.pyx
```diff
@@ -2270,12 +2280,17 @@ cdef class EdgeConnection:
...
             assert recv_args == query.first_extra, \
                 f"argument count mismatch {recv_args} != {query.first_extra}"
-            out_buf.write_int16(<int16_t>(recv_args + query.extra_count))
-        else:
-            out_buf.write_int16(<int16_t>recv_args)
+            num_args += query.extra_count
+        if query.query_unit.globals:
+            num_args += len(query.query_unit.globals)
+            for _, has_present_arg in query.query_unit.globals:
+                num_args += has_present_arg
```
1st1, line 2291:
```suggestion
                if has_present_arg:
                    num_args += 1
```

1st1 (commented):
I've left a few comments

edb/lib/std/30-jsonfuncs.edgeql
1st1: In general, adding a new stdlib function is a big deal. Let's not add them in some PRs hidden amongst bigger changes. We should have a separate PR/review process for new APIs.

edb/server/config/ops.py
msullivan: I think overall reusing the Operation machinery here is good, but I rearranged the logic here to be driven by the `scope`, and never attempting to look up a `Setting` for a global, which I think smells a lot less

PR #3750, opened 2022-04-13
Fix some issues with defaults

msullivan:
My initial goal here was to delete `ptr_default_to_col_default` and
not set column defaults at all, since we inject the defaults ourselves
anyway. I then realized that the default does serve the actual purpose
of populating data when a new column is created--and that there was a
bug, since multi pointers and things with complex defaults would not
get this treatment, so I used the _alter_pointer_optionality path
to populate that data instead of defaults.

But it also turns out that link properties do actually require having
the defualt column set, since we don't inject the values there. I
didn't want to try to make that change now, so instead I just added an
error for the cases where we can't set a column default on a link
property.

This means I didn't actually accomplish the initial goal, but did fix
a bunch of bugs, so still a win.

### DIFF
--- a/edb/pgsql/delta.py
+++ b/edb/pgsql/delta.py
@@ -4636,7 +4636,8 @@ class PropertyMetaCommand(CompositeMetaCommand, PointerMetaCommand):
                         raise errors.UnsupportedFeatureError(
                             f'default value for '
                             f'{prop.get_verbosename(schema, with_parent=True)}'
-                            f' is too complicated for a link property default',
+                            f' is too complicated; link property defaults '
+                            f'must not depend on database contents',
                             context=self.source_context)
 
                     cols = self.get_columns(
--- a/edb/pgsql/delta.py
+++ b/edb/pgsql/delta.py
@@ -3251,12 +3251,8 @@ class PointerMetaCommand(MetaCommand):
         default_value = None
 
         if default is not None:
-            if isinstance(default, s_expr.Expression):
-                default_value = schemamech.ptr_default_to_col_default(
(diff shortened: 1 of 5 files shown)


edb/pgsql/delta.py
```diff
@@ -4610,6 +4628,17 @@ def _create_property(
...
 
+                    if (
+                        isinstance(src.scls, s_links.Link)
+                        and not default_value
+                        and prop.get_default(schema)
+                    ):
+                        raise errors.UnsupportedFeatureError(
+                            f'default value for '
+                            f'{prop.get_verbosename(schema, with_parent=True)}'
+                            f' is too complicated for a link property default',
```
elprans, line 4639: Can we elaborate what "too complicated" actually means here?

edb/pgsql/delta.py
```diff
@@ -4610,6 +4628,18 @@ def _create_property(
...
 
+                    if (
+                        isinstance(src.scls, s_links.Link)
+                        and not default_value
+                        and prop.get_default(schema)
+                    ):
+                        raise errors.UnsupportedFeatureError(
+                            f'default value for '
+                            f'{prop.get_verbosename(schema, with_parent=True)}'
+                            f' is too complicated; link property defaults '
```
elprans, line 4639: perhaps `default value for .. must be a constant expression`?

elprans (approved):

edb/pgsql/delta.py
msullivan: Well, it can be stuff like `random()`, though

PR #3751, opened 2022-04-14
Fix a collection of issues with aliases

msullivan:
* Fix migrating an alias between different types. Previously,
   when doing this, constraints would be added to the view object
   even though they are normally left off; the fix was to make the
   object commands in define_alias canonical.
 * Make the rptr weak_ref field be stored with ON TARGET DELETE ALLOW,
   which fixes some issues when migrating away from a complex alias.
 * Support comparision between ObjectType and ScalarType so that
   you can migrate between them.
 * Fix a number of issues with aliases that have a view inside of a
   tuple. The main piece here is generating proper names for all views
   that appear in aliases, not just ones that appear with an rptr.

### DIFF
--- a/edb/edgeql/compiler/context.py
+++ b/edb/edgeql/compiler/context.py
@@ -232,6 +232,9 @@ class Environment:
     compiled_stmts: Dict[qlast.Statement, irast.Stmt]
     """A mapping of from input edgeql to compiled IR"""
 
+    actually_toplevel_result_view_name: Optional[s_name.QualName]
+    """The name of a view being defined as an alias."""
+
     def __init__(
         self,
         *,
@@ -271,6 +274,7 @@ class Environment:
         self.scope_tree_nodes = weakref.WeakValueDictionary()
         self.materialized_sets = {}
         self.compiled_stmts = {}
+        self.actually_toplevel_result_view_name = None
 
     def add_schema_ref(
             self, sobj: s_obj.Object, expr: Optional[qlast.Base]) -> None:
--- a/edb/edgeql/compiler/context.py
+++ b/edb/edgeql/compiler/context.py
@@ -232,7 +232,7 @@ class Environment:
     compiled_stmts: Dict[qlast.Statement, irast.Stmt]
     """A mapping of from input edgeql to compiled IR"""
 
-    actually_toplevel_result_view_name: Optional[s_name.QualName]
+    alias_result_view_name: Optional[s_name.QualName]
     """The name of a view being defined as an alias."""
 
     def __init__(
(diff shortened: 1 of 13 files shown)


edb/edgeql/compiler/context.py
```diff
@@ -232,6 +232,9 @@ class Environment:
     compiled_stmts: Dict[qlast.Statement, irast.Stmt]
     """A mapping of from input edgeql to compiled IR"""
 
+    actually_toplevel_result_view_name: Optional[s_name.QualName]
```
elprans, line 235: Drop `actually`.

edb/edgeql/compiler/context.py
```diff
@@ -271,6 +274,7 @@ def __init__(
         self.scope_tree_nodes = weakref.WeakValueDictionary()
         self.materialized_sets = {}
         self.compiled_stmts = {}
+        self.actually_toplevel_result_view_name = None
```
elprans, line 277: Pass it as argument instead of setting directly below?

edb/schema/ddl.py
```diff
@@ -427,6 +427,9 @@ def schemas_are_equal(
         schema_b_filters=schema_b_filters,
         linearize_delta=False,
     )
+    # if bool(diff.get_subcommands()):
```
elprans, line 430: leftovers

edb/schema/expraliases.py
```diff
@@ -516,6 +516,24 @@ def define_alias(
...
+    # If the delta includes a rebase, canonicalizing that will add
+    # back in those things---so mark everything as canonical to
+    # avoid that.
+    #
+    # Except that the collection operations we generated above came
+    # from shells, and are missing things that they need to be
+    # complete.
+    for op in derived_delta.get_subcommands(type=sd.ObjectCommand):
+        if not issubclass(op.get_schema_metaclass(), s_types.Collection):
+            op.canonical = True
```
elprans, line 535: Hmm. Perhaps `as_create_delta` should set `canonical` instead?

edb/edgeql/compiler/context.py
msullivan: I'm going with `alias_result_view_name`. I don't want it to just have the same name as `toplevel_result_view_name` in context, which is different

edb/schema/expraliases.py
msullivan: Aha: `as_create_delta` (and friends) do set `canonical`, but `ordering` blows it up some times. So I think we can just fix that

elprans (approved):

PR #3752, opened 2022-04-14
Fix `[NOT] IN array_unpack(foo)` when foo is empty

msullivan:
The array version of ANY/ALL produces NULL if there is a NULL
argument, so we need to coalesce in that case.


elprans (approved):

msullivan:
Cherry-picked to 1.x

PR #3755, opened 2022-04-15
Fix START TRANSACTION

1st1:


elprans (approved):

edb/server/pgcon/pgcon.pyx
```diff
@@ -1068,6 +1076,9 @@ cdef class PGConnection:
         try:
             if state is not None:
                 await self._parse_apply_state_resp(state)
+                if state_sync:
+                    await self.wait_for_sync()
+                    self.waiting_for_sync = True
```
fantix, line 1081: Maybe add a comment/GOTCHA here, specifying the reason why the order of these two lines are not wrong? I was confused in the first place without reading the context.

fantix (approved):
lgtm!

edb/server/pgcon/pgcon.pyx
1st1: Sure, done

1st1:
@msullivan @elprans please cherry pick to 1.x

PR #3769, opened 2022-04-19
Implement computed globals

msullivan:
Heavily reuse code paths from expraliases.


edb/schema/delta.py
```diff
@@ -2158,6 +2158,8 @@ def _deparse_name(
 
         if isinstance(name, sn.QualName):
             nname = sn.shortname_from_fullname(name)
+            if not isinstance(nname, sn.QualName):
```
elprans, line 2161: oops

elprans (approved):

PR #3773, opened 2022-04-20
Fix REPL run shortcut on macOS

fantix:
Refs edgedb/edgedb-cli#716

### DIFF
--- a/docs/guides/cheatsheet/repl.rst
+++ b/docs/guides/cheatsheet/repl.rst
@@ -22,8 +22,8 @@ on Windows/Linux, or Esc+Return on macOS.
 .. tip::
 
     The Alt-Enter / ESC+Return trick is also very useful when your cursor is in
-    the middle of a long query and you want to run it directly without moving the
-    cursor to the end of the query.
+    the middle of a long query and you want to run it directly without moving 
+    the cursor to the end of the query.
     
     macOS terminals allow you to configure the left Option key to act as ESC+,
     which makes Alt-Enter also work on macOS:
--- a/docs/guides/cheatsheet/repl.rst
+++ b/docs/guides/cheatsheet/repl.rst
@@ -11,8 +11,8 @@ a semicolon and press "ENTER".
     db> select 5;
     {5}
 
-Alternatively, you can run the query without a semicolon with esc+return on macOS
-or Alt-Enter on Windows/Linux.
+Alternatively, you can run the query without a semicolon by hitting Alt-Enter 
+on Windows/Linux, or Esc+Return on macOS.
 
 .. code-block:: edgeql-repl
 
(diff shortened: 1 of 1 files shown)


docs/guides/cheatsheet/repl.rst
```diff
@@ -11,14 +11,20 @@ a semicolon and press "ENTER".
     db> select 5;
     {5}
 
-Alternatively, you can run the query without a semicolon with ⌥-Enter on macOS
+Alternatively, you can run the query without a semicolon with esc+return on macOS
 or Alt-Enter on Windows/Linux.
```
1st1, line 15:
```suggestion
Alternatively, you can run the query without a semicolon by hitting an `ESC`
key immediately followed by `RETURN` on macOS, or Alt-Enter on Windows/Linux.
```

docs/guides/cheatsheet/repl.rst
```diff
@@ -11,14 +11,20 @@ a semicolon and press "ENTER".
...
 .. code-block:: edgeql-repl
 
     db> select 5
     {5}
 
+.. tip::
+
+    This esc+return or Alt-Enter trick is also very useful when your cursor is in
+    the middle of a long query and you want to run it directly without moving the
+    cursor to the end of the query.
```
1st1, line 26:
```suggestion
    The Alt-Enter / ESC+Return trick is also very useful when your cursor is in
    the middle of a long query and you want to run it directly without moving the
    cursor to the end of the query.
    
    Terminals like iTerm allow to configure left Option key to act as ESC+, which
    makes Alt-Enter also work on macOS.
```

docs/guides/cheatsheet/repl.rst
colinhacks: This works even if you hold down the `esc` key

docs/guides/cheatsheet/repl.rst
colinhacks: Let's mention the "Use Option as Meta option" here for Terminal.app users

colinhacks: > Terminals like iTerm allow to configure

Edit: "Terminals like iTerm allow *you* to configure *the* left Option"

1st1: Good catch and good suggestion on Terminal.app. I've updated with some detailed instructions for both iTerm and Terminal.app

docs/guides/cheatsheet/repl.rst
```diff
@@ -11,14 +11,26 @@ a semicolon and press "ENTER".
...
 .. code-block:: edgeql-repl
 
     db> select 5
     {5}
 
+.. tip::
+
+    The Alt-Enter / ESC+Return trick is also very useful when your cursor is in
+    the middle of a long query and you want to run it directly without moving the
+    cursor to the end of the query.
```
1st1, line 26:
```suggestion
    The Alt-Enter / ESC+Return trick is also very useful when your cursor is in
    the middle of a long query and you want to run it directly without moving 
    the cursor to the end of the query.
```

PR #3777, opened 2022-04-21
Sort refs in get_inherited_ref_layout

### LINKED ISSUE #3775: Broken order dependence when deriving types
If you do
```
create type A {
    create property foo := "!";
    create property bar -> str;
};
alter type A alter property foo using (.bar);
```

And then
```
create type B extending A;
```
we get a totally bogus `object type 'default::B' has no link or property 'bar'` error, complete with broken source span.

The cause here is that `inherit_fields` recompiles all expression fields as soon as it sees them, and so if the order that pointers are reprocessed does not match their internal dependency order, then things will break.

It seems that in practice they get processed in the order we get them back from the introspection tables, which in practice seems to be the order that they were created. So that's why most normal cases work, and this weird case that adds a dependency after the fact breaks. But as far as I can tell even the working cases work by accident, since we don't do anything to ensure the ordering

msullivan:
This makes deriving from a type with computed pointers work on
purpose, instead of by accident. (Previously it would work if it
happened to process them in the right order, which it usually would,
since it seems the database produced them in creation order
typically.)

Fixes #3775.

### DIFF
--- a/edb/schema/inheriting.py
+++ b/edb/schema/inheriting.py
@@ -203,12 +203,26 @@ class InheritingObjectCommand(sd.ObjectCommand[so.InheritingObjectT]):
 
         ancestors = set(self.scls.get_ancestors(schema).objects(schema))
         for base in bases.objects(schema) + (self.scls,):
-            base_refs: so.ObjectIndexBase[
+            base_refs: Dict[
                 sn.Name,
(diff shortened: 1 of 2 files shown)


elprans (approved):

msullivan:
Cherry picked

PR #3778, opened 2022-04-21
Add uuid_generate_v4() function to the stdlib.

### LINKED ISSUE #3774: [Feature] Expose the generate_uuid_v4() function from the uuid-ossp Postgres extension
OWASP suggests session keys contain **at least 128 bits** of random information. The ideal data type that is fixed-size and also optimised for fast lookup by Postgres is the UUID type, which also happens to be exactly 128 bits.

The EdgeDB standard library only exposes `uuid_generate_v1mc()` from the `uuid-ossp` Postgres extension. While this is sufficient for primary keys, UUID v1 should **never be used for any sort of authorization purpose**, the randomised MAC address accounts for only 48 bits. Given a previous key, it becomes _much_ easier to predict future keys ([UUID version differences](https://versprite.com/blog/universally-unique-identifiers/)).

See the below example for UUIDs generating using `uuid_generate_v1mc()` in quick succession:

```text
{91b24894-c0cb-11ec-b79f-cf8b1d655c91}
{91e0c44e-c0cb-11ec-b79f-9bd023a1c068}
{920794de-c0cb-11ec-b79f-af1198448bf5}
{92365404-c0cb-11ec-b79f-5b4335fd24ab}
{92695fd4-c0cb-11ec-b79f-676e47860109}
{9296a430-c0cb-11ec-b79f-175ed269c4b4}
```

Postgres exposes a `uuid_generate_v4()` function in the exactly the same extension as `uuid_generate_v1mc()` (`uuid-ossp`). UUID v4 possesses 122 bits of purely random information, which is acceptable for most applications that aren't particularly sensitive.

The stored v4 UUID could also be combined with the primary key UUID of the object to give 170 bits of random information, whilst being optimised for lookup and storage efficiency.

I've escalated this from a Discord reply:

> Yeah, exposing built-in Postgres functions is trivial.  Please open an issue

Would such a feature be implementable? The current solution is to generate the UUID v4 on the server before insertion, Node.js supports this natively via the `crypto.randomUUID()` function, but not a lot of people know this and rely instead on more third-party packages.

1st1:
Closes #3774.

### DIFF
--- a/edb/lib/std/30-uuidfuncs.edgeql
+++ b/edb/lib/std/30-uuidfuncs.edgeql
@@ -24,9 +24,15 @@ CREATE FUNCTION
 std::uuid_generate_v1mc() -> std::uuid {
     CREATE ANNOTATION std::description := 'Return a version 1 UUID.';
     SET volatility := 'Volatile';
-    USING SQL $$
-    SELECT edgedbext.uuid_generate_v1mc()
-    $$;
+    USING SQL FUNCTION 'edgedbext.uuid_generate_v1mc';
+};
+
+
+CREATE FUNCTION
+std::uuid_generate_v4() -> std::uuid {
+    CREATE ANNOTATION std::description := 'Return a version 4 UUID.';
+    SET volatility := 'Volatile';
+    USING SQL FUNCTION 'edgedbext.uuid_generate_v4';
 };
 
 
--- a/docs/stdlib/uuid.rst
+++ b/docs/stdlib/uuid.rst
@@ -45,4 +45,19 @@ UUID
     .. code-block:: edgeql-repl
 
         db> select uuid_generate_v1mc();
-        {'1893e2b6-57ce-11e8-8005-13d4be166783'}
+        {1893e2b6-57ce-11e8-8005-13d4be166783}
+
+
+---------
+
+
+.. eql:function:: std::uuid_generate_v4() -> uuid
+
+    Return a version 4 UUID.
+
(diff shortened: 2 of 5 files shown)


edb/lib/std/30-uuidfuncs.edgeql
```diff
@@ -30,6 +30,16 @@ std::uuid_generate_v1mc() -> std::uuid {
 };
 
 
+CREATE FUNCTION
+std::uuid_generate_v4() -> std::uuid {
+    CREATE ANNOTATION std::description := 'Return a version 4 UUID.';
+    SET volatility := 'Volatile';
+    USING SQL $$
+    SELECT edgedbext.uuid_generate_v4()
+    $$;
```
elprans, line 39:
```suggestion
    USING SQL FUNCTION 'edgedbext.uuid_generate_v4';
```

1st1: OK, I'll update the v1mc function to use this too.

1st1: I think your suggestion broke volatility tests. cc @msullivan

msullivan: It's a flake, I reran the tests.

I think the bug for those failures might actually be in postgres, but damned if I know how to prove that. I've only ever been able to repro it in CI. I was able to confirm that we are generating the correct code for those cases

PR #3784, opened 2022-04-22
Add functions for bitwise operations to std.

vpetrovykh:
Add `bit_and`, `bit_or`, `bit_xor`, `bit_not`, `bit_lshift`,
`bit_rshift` that work on all of EdgeDB integers.


edb/lib/std/26-bitwisefuncs.edgeql
```diff
@@ -0,0 +1,241 @@
...
+};
+
+
+CREATE FUNCTION
+std::bit_lshift(val: std::int32, n: std::int64) -> std::int32
+{
+    CREATE ANNOTATION std::description :=
+        'Bitwise shift left operator for 32-bit integers.';
+    SET volatility := 'Immutable';
+    USING SQL $$
```
1st1, line 226: As discussed on the call, this needs some extra handling for big vals of `n`. Also negative `n` shouldn't be allowed neither for rshift nor for lshit

1st1 (commented):
@vpetrovykh looks good, please fix lshift/rshift and write docs!

vpetrovykh:
See #3799 for more details.

edb/lib/std/26-bitwisefuncs.edgeql
```diff
@@ -0,0 +1,341 @@
...
+
+
+# In Postgres bitwise shift operators accept a 32-bit integer as the number of
+# bit positions that need to be shifted. However, in EdgeDB the default
+# integer literal is int64, so we should accept that and cast it down inside
+# the function body.
+#
+# In Postgres the number of bits shifted gets truncated using a positive mod
+# 32 (or mod 64 for int8). We do not want such truncation in EdgeDB. Shifting by 20 bits 2 times
+# should bethe same as shifitng by 40 bits once.
```
msullivan, line 175: `shifting`

vpetrovykh: Fixed typo.

edb/lib/std/30-arrayfuncs.edgeql
```diff
@@ -41,6 +41,41 @@ std::array_unpack(array: array<anytype>) -> SET OF anytype
 };
 
 
+CREATE FUNCTION
+std::array_fill(val: anytype, n: std::int64) -> array<anytype>
```
elprans, line 45: I wonder if we should add `repeat(val: anytype, n) -> set of anytype` instead.  Then this could be expressed as `array_agg(repeat('foo', 10))`

vpetrovykh: I'm tempted to say that it's worth having a `repeat` function all by itself in addition to `array_fill`, largely because calling an `array_agg` on it sounds like it would not be efficient at all, since aggregates in SQL are a whole category of their own that has to walk the data in a linear fashion.

elprans (approved):

PR #3795, opened 2022-04-25
Expose Function.backend_name to introspection

msullivan:
This is pretty marginal, but I want it to support some bad scripts I
use to rewrite queries between different instances of a database.

In general, while we don't *support* directly writing SQL queries
against our database, I think that the introspection schema ought
to still contain the information necessary for it to be possible.

### DIFF
--- a/edb/buildmeta.py
+++ b/edb/buildmeta.py
@@ -44,7 +44,7 @@ from edb.common import verutils
 
 
 # Increment this whenever the database layout or stdlib changes.
-EDGEDB_CATALOG_VERSION = 2022_04_21_00_01
+EDGEDB_CATALOG_VERSION = 2022_04_25_00_00
 EDGEDB_MAJOR_VERSION = 2
 
 
--- a/edb/lib/schema.edgeql
+++ b/edb/lib/schema.edgeql
@@ -432,6 +432,7 @@ CREATE TYPE schema::Function
     };
 
     CREATE MULTI LINK used_globals EXTENDING schema::ordered -> schema::Global;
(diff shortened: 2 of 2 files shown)


1st1:
I don't like exposing too many implementation details. 

Can you elaborate on "but I want it to support some bad scripts I use to rewrite queries between different instances of a database"? How that would be useful to anyone but you?

elprans (changes requested):
Not sure I like exposing internal implementation details in the introspection schema.  The ad-hoc rewrite scripts don't seem like a sufficient justification to me.

msullivan:
The more general point here is that, like I said, I think that it ought to be *possible* to write SQL scripts against an edgedb database using the information in `schema`. Users ask about this pretty frequently, and the answer we always give is "it's not supported or recommended, but it *is* possible, and here's where you can get the information".

I think we should expose the information needed to call an edgeql function

1st1:
We'll eventually provide an open read-only SQL interface to access the database that would de-mangle type names and (possibly) functions.

But what you propose is different -- you propose to expose the mangled implementation-detail name to users and I'm -1 on that for a few reasons:

1. We generally don't give access to the underlying backend now, and we certainly will not be giving that ability in our cloud offering.
2. I want us to have an ability to change backend names whenever we see fit. Be it after an upgrade or after some (possibly not directly related) migration.
3. EdgeQL functions already have extra arguments (globals, info about keyword argument, etc). Users might not be able to even call our functions from SQL because it's actually cumbersome to compose these extra args.

PR #3797, opened 2022-04-26
Drop broken special case for nested insert FOR

### LINKED ISSUE #3789: Guaranteed ISE on nested insert on a `required multi link`
Schema:

```
module default {
    type B {
        required multi link c -> default::C;
    };
    type C {
        required property x -> std::int64;
    };
};
```

Query:

```
insert B {
  c := (
    for i in {(x := 90), (x := 240)}
    union (
      insert C {
        x := i.x,
      }
    )
  ),
};
```

Result:

```
edgedb error: InternalServerError: duplicate key value violates unique constraint "96db17f1-c393-11ec-81bf-b3f038f3cc4b_source_target_key"
  Server traceback:
      edb.errors.InternalServerError: duplicate key value violates unique constraint "96db17f1-c393-11ec-81bf-b3f038f3cc4b_source_target_key"
```

Context in #3786

msullivan:
There is some code that detects when a nested insert contains a DML
inside an iterator, and explicitly reincludes it. I think that this
was a workaround for when the general DML/iterator mechanisms were
much more broken than they are now.

It causes an extra join, and in the case of multi required things, the
join is broken. Just drop that code.

Fixes #3789.

### DIFF
--- a/edb/pgsql/compiler/dml.py
+++ b/edb/pgsql/compiler/dml.py
@@ -1959,35 +1959,6 @@ def process_link_values(
 
             dispatch.visit(ir_expr, ctx=input_rel_ctx)
 
-            if (
-                isinstance(ir_expr.expr, irast.Stmt)
-                and ir_expr.expr.iterator_stmt is not None
-            ):
-                # The link value is computaed by a FOR expression,
-                # check if the statement is a DML statement, and if so,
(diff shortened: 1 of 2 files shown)


elprans (approved):

msullivan:
cherry-picked

PR #3801, opened 2022-04-26
Avoid generating optional wrappers for expressions that can't be empty

msullivan:


elprans (approved):

msullivan:
Cherry-picked for 1.4

PR #3814, opened 2022-05-02
Protocol v1.0 - part 1

fantix:
Pre-RFC: https://github.com/edgedb/edgedb/discussions/3772

Depends on https://github.com/edgedb/edgedb-python/pull/306


1st1 (commented):
Dropping pre 0.13 looks OK.

As for copying the protocol - this is fine. I going to implement a different approach:

```
binary.pyx: EdgeConnection
binary_v0.pyx: EdgeConnectionBackwardsCompatible extends EdgeConnection
```

`binary_v0.pyx` would have v0 methods prefixed with `def legacy_`.

Then `protocol.pyx` could just always instantiate `EdgeConnectionBackwardsCompatible`

---

Now, I think your approach also works so let's go with it.

fantix:
> Now, I think your approach also works so let's go with it.

Thanks for the quick response! That's exactly what I wanted to know.

However with my current approach, there're some complication of like switching the protocol object attached in the transport, and some in-and-out in the server. I'll try the inheritance approach very quickly and go on from there.

1st1 (approved):

PR #3820, opened 2022-05-03
Fix IN array_unpack for bigints

### LINKED ISSUE #3817: ISE in `foo IN array_unpack(<double-array-cast>)`
```
edgedb> select 1n in array_unpack(<array<bigint>><array<str>>to_json('["1"]'));
edgedb error: InternalServerError: syntax error at or near "ANY"
  Server traceback:
      edb.errors.InternalServerError: syntax error at or near "ANY"
```

msullivan:
The way we insert casts for the comparison operator breaks on
array_unpacks.

Also, we failed to take into account the scopes that we skipped over,
which caused trouble in the double cast cases.

Fixes #3817.

### DIFF
--- a/edb/pgsql/compiler/relgen.py
+++ b/edb/pgsql/compiler/relgen.py
@@ -1369,16 +1369,19 @@ def process_set_as_membership_expr(
         newctx.expr_exposed = False
         left_out = dispatch.compile(left_arg, ctx=newctx)
 
-        right_arg = irutils.unwrap_set(right_arg)
+        orig_right_arg = right_arg
+        unwrapped_right_arg = irutils.unwrap_set(right_arg)
         # If the right operand of [NOT] IN is an array_unpack call,
         # then use the ANY/ALL array comparison operator directly,
         # since that has a higher chance of using the indexes.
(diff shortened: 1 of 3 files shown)


edb/pgsql/compiler/relgen.py
```diff
@@ -1389,6 +1392,11 @@ def process_set_as_membership_expr(
         left_is_row_expr = astutils.is_row_expr(left_out)
 
         with newctx.subrel() as _, _.newscope() as subctx:
+            if is_array_unpack:
+                relctx.update_scope(orig_right_arg, subctx.rel, ctx=subctx)
```
elprans, line 1396: Why do we need to `update_scope` on both `orig_` and `unwrapped_`?

elprans (approved):

edb/pgsql/compiler/relgen.py
msullivan: Because if `unwrapped_` has a `path_scope_id`, then that takes precedence, but otherwise we want to use `orig_`'s

msullivan:
Cherry-picked for 1.4

PR #3824, opened 2022-05-04
Implement lint-on/lint-off

colinhacks:


tests/test_docs.py
```diff
@@ -121,8 +121,15 @@ def extract_code_blocks(self, source: str, filename: str):
...
 
         lines = source.split('\n')
+        lint_on = True
         for lineno, line in enumerate(lines, 1):
-            if len(line) > self.MAX_LINE_LEN:
+
+            if line.startswith('.. lint-off'):
+                lint_on = False
+            elif line.startswith('.. lint-on'):
+                lint_on = True
```
1st1, line 130: Can you also add basic pairing validation? Basically lint-off should always be followed by lint-on before EOF. That would prevent disabling lint accidentally after merge conflicts.

1st1 (approved):
Please address the comment before merging.

tests/test_docs.py
colinhacks: Done

PR #3834, opened 2022-05-05
Fix cardinality inference for casts from json

### LINKED ISSUE #3832: Casts from json can produce empty values but card inference doesn't understand
For example:
```
edgedb> select { required z := <str>to_json('null') };
[{"z": null}]
```

(Prompted by https://github.com/edgedb/edgedb/issues/3793)

msullivan:
Fixes #3832


elprans (approved):

msullivan:
Skipping cherry-picking this into 1.4, because 1.x lacks the `can_omit_optional_wrapper` optimization, and I don't think I want to add it in, and unfortunately (for annoying reasons) that optimization works around a bug that can only show up in our internal schema manipulation code

PR #3837, opened 2022-05-06
Execute script

fantix:
Depends on #3814, https://github.com/edgedb/edgedb-python/pull/310

- [ ] Add tests
- [ ] Return result of the last command in a script
- [ ] Always use `stmt_mode = "all"` in `EdgeConnection._compile()` and check in `server.compiler`
- [x] Fix tests
- [x] Support constant extraction on multiple-statement queries
- [ ] Drop remaining SimpleQuery

Sample scenarios of EdgeDB `Execute` generating Postgres messages using pseudo-code:

## Single-command EdgeQL

EdgeDB Protocol:

```
Execute(
    insert Hero { name := 'Daisy Johnson' }
)
```

Generated Postgres messages:

```
Parse("INSERT INTO heros (name) VALUES ($1)")
Bind('Daisy Johnson')
Execute()
Sync()
```

On protocol-level, transaction control commands like `START TRANSACTION`, `COMMIT`, `ROLLBACK [...]` are allowed.

## Multi-command EdgeQL

EdgeDB Protocol:

```
Execute(
    insert Hero { name := 'Daisy Johnson' };
    insert Hero { name := 'Phillip J. Coulson' };
)
```

Generated Postgres messages:

```
Parse("INSERT INTO heros (name) VALUES ($1)")
Bind('Daisy Johnson')
Execute()
Parse("INSERT INTO heros (name) VALUES ($1)")  # Maybe omitted if prepared statement rules apply
Bind('Phillip J. Coulson')
Execute()
Sync()
```

There is only one `Sync` per EdgeDB `Execute`, so it is either an atomic implicit transaction in Postgres, or a part of the outer existing transaction.

On protocol-level, transaction control commands are NOT allowed.

`Flush` messages may be added if `SET GLOBAL` is in the script, but that won't affect the atomicity. Only the values generated from the last command will be returned.

## Migration Commands

`START MIGRATION ...`, `COMMIT MIGRATION` and `ABORT MIGRATION` are allowed in scripts, but only if all `START MIGRATION ... ` have a matching `COMMIT MIGRATION` or `ABORT MIGRATION`. Implementation-wise, migration commands in scripts will never open or close a database transaction or savepoint, they are always a part of the outer implicit or explicit transaction.

In addition, there will be a new capability `MIGRATION_TX`, gating the ability of the migration commands to actually open or close a database transaction. The client bindings should always turn this capability off, so that the following practice is forbidden:

```python
client.execute("START MIGRATION TO ...")
...
client.execute("COMMIT MIGRATION")
```

Instead, the client could either run migration in a script:

```python
client.execute("""
    START MIGRATION TO ...;
    ...
    COMMIT MIGRATION;
""")
```

Or use an explicit transaction to wrap the separate migration commands:

```python
for tx in client.transaction():
    with tx:
        tx.execute("START MIGRATION TO ...")
        ...
        tx.execute("COMMIT MIGRATION")
        # Or just use a script again:
        tx.execute("""
            START MIGRATION TO ...;
            ...
            COMMIT MIGRATION;
        """)
```

## Client Bindings

Both `query()` and `execute()` will use the same `Parse`/`Execute` EdgeDB protocol messages, both accepting single-command and multi-command EdgeQL. The difference is, `execute()` will generate `Parse`/`Execute` messages with `output_format=NULL` (previously known as `io_format`), while `query()` will use `output_format=BINARY` (or `JSON` for the `query_*_json()` family).

# Rejected Ideas

## Transactional Script

EdgeDB Protocol:

```
Execute(
    start transaction;
    insert Hero { name := 'Daisy Johnson' };
    commit;
    start transaction;
    insert Hero { name := 'Phillip J. Coulson' };
    rollback;
)
```

Generated Postgres messages:

```
Parse("START TRANSACTION")
Bind()
Execute()
Parse("INSERT INTO heros (name) VALUES ($1)")
Bind('Daisy Johnson')
Execute()
Parse("COMMIT")
Bind()
Execute()
Sync()

# If we're in error state: (or always ROLLBACK after COMMIT to save round-trip time)
Parse("ROLLBACK")
Bind()
Execute()
Sync()

Parse("START TRANSACTION")
Bind()
Execute()
Parse("INSERT INTO heros (name) VALUES ($1)")  # Maybe omitted if prepared statement rules apply
Bind('Phillip J. Coulson')
Execute()
Sync()  # make sure the following ROLLBACK works, clearing `ignore_till_sync` pg state
Parse("ROLLBACK")
Bind()
Execute()
Sync()
```

Returning values is NOT allowed.

## Illegal Transactional Script

Statements outside transactions are not allowed:

```
Execute(
    start transaction;
    insert Hero { name := 'Daisy Johnson' };
    commit;
    insert Hero { name := 'Phillip J. Coulson' };
)
```

Mismatching transaction boundaries are not allowed:

```
Execute(
    insert Hero { name := 'Daisy Johnson' };
    commit;
    start transaction;
    insert Hero { name := 'Phillip J. Coulson' };
    rollback;
)
```


edb/server/compiler/dbstate.py
```diff
@@ -277,6 +277,79 @@ def has_ddl(self) -> bool:
...
+    # Cardinality of the result set.  Set to NO_RESULT if the
+    # unit represents multiple queries compiled as one script.
+    cardinality: enums.Cardinality = enums.Cardinality.NO_RESULT
+
+    out_type_data: bytes = sertypes.NULL_TYPE_DESC
+    out_type_id: bytes = sertypes.NULL_TYPE_ID.bytes
+    in_type_data: bytes = sertypes.NULL_TYPE_DESC
+    in_type_id: bytes = sertypes.NULL_TYPE_ID.bytes
+    in_type_args: Optional[List[Param]] = None
+    globals: Optional[List[str]] = None
```
msullivan, line 306: globals will be fully per QueryUnit, not at all per QueryUnitGroup

Though I guess you can skip worrying about it and I'll fix it when I do argument reworking

fantix: I see, thanks! I was adding this because global is used in `parse()` of `binary.pyx`, but I’ll leave as it is now.

fantix:
`START MIGRATION TO` and `COMMIT MIGRATION` are actually transaction starting and committing I think. I have to use separate `execute()` in tests for them. Maybe this will change in the future (?)

fmoor:
> START MIGRATION TO and COMMIT MIGRATION are actually transaction starting and committing I think.

Using the old protocol, I was able to run this query with the allow capabilities header transaction capabilities bit off and it didn't complain. Was the old behavior a bug?

docs/reference/protocol/messages.rst
```diff
@@ -257,6 +238,9 @@ Use:
 * ``JSON_ELEMENTS`` to return a single JSON string per top-level set element.
   This can be used to iterate over a large result set efficiently.
 
+* ``NONE`` to prevent the server from returning data, even if the EdgeQL
+  command does.
```
elprans, line 242:
```suggestion
  statement does.
```

edb/server/compiler/compiler.py
```diff
@@ -1899,13 +1884,13 @@ def _try_compile(
...
                 elif comp.config_scope is qltypes.ConfigScope.GLOBAL:
-                    unit.system_config = True
+                    unit.set_global = True
                     # FIXME: We want to be able to do this in a block
                     # with other commands.
-                    if statements_len > 1:
-                        raise errors.UnsupportedFeatureError(
-                            'SET GLOBAL cannot be executed in a block '
-                            'with other commands')
+                    # if statements_len > 1:
```
elprans, line 1890: This code can be deleted now?

edb/server/pgcon/pgcon.pyx
```diff
@@ -1096,10 +1213,9 @@ cdef class PGConnection:
                     if mtype == b'D':
                         # DataRow
                         if not has_result:
-                            raise errors.InternalServerError(
-                                f'query that was inferred to have '
-                                f'no data returned received a DATA package; '
-                                f'query: {query.sql}')
+                            # It's likely output_format=NONE, so just discard
```
elprans, line 1216: Can we pass output_format in the query unit to avoid guessing?

edb/server/protocol/binary.pyx
```diff
@@ -1611,6 +1562,14 @@ cdef class EdgeConnection:
                         # set conn.last_state correctly later
                         orig_state = None
 
+                if query_unit.tx_savepoint_rollback:
+                    # await self.recover_current_tx_info(conn)
```
elprans, line 1566:
```suggestion
```

tests/test_edgeql_data_migration.py
```diff
@@ -10869,18 +10869,18 @@ async def test_edgeql_migration_eq_collections_25(self):
                     alias Foo := [20];
                 }
             };
-            POPULATE MIGRATION;
-            COMMIT MIGRATION;
         """)
+        await self.con.execute("POPULATE MIGRATION")
```
elprans, line 10873: Why are these pulled out?  We ought to support `start/populate/commit` as a single script.  While these are implicit transaction boundaries, they are _not_ the same as explicit `start transaction/commit`.

fantix: @fmoor I'll amend a fix to allow migration commands in scripts

edb/server/pgcon/pgcon.pyx
fantix: Yup that's better - and makes script logic easier.

edb/server/protocol/binary.pyx
```diff
@@ -1077,18 +1077,27 @@ cdef class EdgeConnection:
...
 
                     if config_ops:
                         await _dbview.apply_config_ops(
                             conn,
                             config_ops)
 
-                    _dbview.on_success(query_unit, new_types)
+                    side_effects = _dbview.on_success(query_unit, new_types)
+                    if side_effects:
+                        # COMMIT MIGRATION breaks the atomicity as an exception
```
elprans, line 1099: Not sure this is correct.  Side effect signaling should only be done once the relevant transaction has been successfully committed.  `COMMIT MIGRATION` could be part of a transaction that can roll back, or, worse, have long-running statements which would mean that the recipients of side effect signals would not actually see the changes they've been signaled about.

fantix: If `COMMIT MIGRATION` is part of a transaction, `dbview.on_success('COMMIT MIGRATION')` won't return any side effects, because it'll be a release savepoint command.

Oppositely, when there is no outer transaction, a script may have more than one `COMMIT MIGRATION` (with some other statements between migrations), that means we'll have 2 batches of side_effects. For example:

```
start migration to { module default { type A {} } };
populate migration;
commit migration;  -- side effects

create type C {};  -- start implicit transaction
start migration to { module default { type B {} } };  -- make implicit transaction "explicit"
populate migration;
commit migration;  -- side effects again
```

This is really breaking atomicity of scripts in different ways, so perhaps we should disallow migration commands outside of explicit transactions, regardless of script or not? Like, we should disallow doing this in Python:

```
>>> import edgedb
>>> c=edgedb.create_client('aa')
>>> c.execute('start migration to { module default { type A {} } }; populate migration;')
>>> c.execute('commit migration')
```

elprans: Yes, we should disallow migration control outside of a transaction (implicit or explicit)

docs/reference/protocol/index.rst
```diff
@@ -302,21 +269,17 @@ messages:
...
-* a :ref:`ref_protocol_msg_sync` message is received.
+Each of the messages could contain one or more EdgeQL commands in the same text
+string. If more than one EdgeQL command is found in a single message, the
+server will treat the commands as an EdgeQL script. EdgeQL scripts are always
+atomic, they will be executed in an implicit transaction block if no explicit
+transactions are started beforehand. Therefore, EdgeQL scripts have limitations
+on the EdgeQL commands they could contain:
+
+* Transaction control commands are not allowed, like ``start transaction``,
+  ``commit``, ``rollback to savepoint`` or ``commit migration``.
```
elprans, line 280:
```suggestion
  ``commit``, ``declare savepoint``, or ``rollback to savepoint``.
```

docs/reference/protocol/index.rst
```diff
@@ -302,21 +269,17 @@ messages:
...
+* Non-transactional commands are not allowed, like ``create database``.
+* ``configure instance`` command is not allowed.
```
elprans, line 282:
```suggestion
* Non-transactional commands, like ``create database`` or ``configure instance`` are not allowed.
```

edb/server/compiler/compiler.py
```diff
@@ -636,136 +644,151 @@ def _compile_ql_query(
...
+
+        if ctx.protocol_version >= (0, 12):
+            in_type_data, in_type_id = \
+                sertypes.TypeSerializer.describe_params(
+                    schema=ir.schema,
+                    params=params,
+                    protocol_version=ctx.protocol_version,
+                )
+        else:
+            # Legacy protocol support
```
elprans, line 687: Drop this?  We shouldn't be supporting 0.11 and below anymore.

elprans (approved):
No obvious issues jumped out at me, so let's merge this and do post-merge review and testing.

edb/server/compiler/dbstate.py
```diff
@@ -226,6 +229,12 @@ class QueryUnit:
     tx_savepoint_rollback: bool = False
     tx_savepoint_declare: bool = False
 
+    # True if this unit is one of the *explicit* transaction control commands
+    # including `START TRANSACTION`, `COMMIT`, `ROLLBACK` and the SAVEPOINT
+    # commands, excluding migration control commands `START MIGRATION ...`,
+    # `COMMIT MIGRATION` and `ABORT MIGRATION`, even though they do control tx
+    tx_control: bool = False
```
1st1: Basically this is `tx_id is not None or tx_commit or tx_rollback or tx_savepoint_rollback or tx_savepoint_declare`? Maybe just make it a property with expression? Or, better, fold all the `tx_*` props into one Enum field.

edb/server/compiler/dbstate.py
```diff
@@ -298,6 +292,16 @@ class QueryUnit:
...
 
+    @property
+    def tx_control(self) -> bool:
+        return (
+            bool(self.tx_id)
+            or self.tx_rollback
+            or self.tx_commit
+            or self.tx_savepoint_declare
+            or self.tx_savepoint_rollback
+        )
```
1st1, line 303: I'm confused... The previous commit killed it... Still, I think that refactoring this to use a simple enum is better than having like 5 or 6 separate properties.

edb/server/compiler/compiler.py
fantix: I think this is for restoring pre-0.12 dumps - I'll complete the comment here.

edb/server/compiler/dbstate.py
fantix: This was a result of my confusion among code review fixes. I didn't force-push to clear this, so it was easier to review the next fix.

edb/server/compiler/dbstate.py
fantix: Sounds reasonable! I'll start a separate PR for this.

PR #3842, opened 2022-05-09
Support global variables on HTTP/graphql protocols

### LINKED ISSUE #3829: Add a mechanism for specifying global variables to HTTP/graphql endpoints

msullivan:
Adds a `globals` argument to the queries that works just like the
`variables` argument.

We pass it in JSON-encoded as one query argument, and fully reuse the
mechanisms for functions that use global variables.

Fixes #3829.

### DIFF
--- a/edb/server/protocol/edgeql_ext.pyx
+++ b/edb/server/protocol/edgeql_ext.pyx
@@ -55,6 +55,7 @@ async def handle_request(
         return
 
     variables = None
+    globals = None
     query = None
 
     try:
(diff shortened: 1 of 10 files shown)


edb/server/protocol/edgeql_ext.pyx
```diff
@@ -96,6 +106,10 @@ async def handle_request(
         if variables is not None and not isinstance(variables, dict):
             raise TypeError('"variables" must be a JSON object')
 
+        # XXX: wait, should we...
```
elprans, line 109: What's this comment in relation to?

msullivan: Honestly I don't remember. I deleted it.

1st1 (approved):
Create an issue to update the docs

PR #3846, opened 2022-05-10
Docs fixes

colinhacks:


vpetrovykh (approved):
Looks good! I think that the "How to read the docs" section has better flow now because it explains what the sections are rather than starting off with "go through..."

PR #3852, opened 2022-05-12
Implement ON SOURCE DELETE DELETE TARGET

msullivan:


edb/pgsql/delta.py
```diff
@@ -5106,7 +5120,19 @@ def get_trigger_name(self, schema, target,
 
         aspect += '-t'
 
-        return common.get_backend_name(
+        # Postgres applies triggers in alphabetical order, and
+        # get_backend_name produces essentially cryptographically
+        # random trigger names.
+        #
+        # All we want for now is for source triggers to applie first,
```
elprans, line 5127:
```suggestion
        # All we want for now is for source triggers to apply first,
```

elprans (approved):

PR #3855, opened 2022-05-12
Always correctly handle variadic arguments when producing AST from migrations

### LINKED ISSUE #3807: Migrations that alter/delete variadic functions are broken
They format the variadic argument as including the array type, which is wrong

msullivan:
In our schema, we store variadic types with an array argument, and we need
be consistent about stripping it away when producing ASTs.

Fixes #3807.


1st1 (approved):

msullivan:
Cherry picked for 1.4

PR #3856, opened 2022-05-12
Fix wrong type name in str->bigint cast error message

### LINKED ISSUE #3737: Typo in error message
- EdgeDB Version: 2.0-dev.6500+local

Error message when casting to 'bigint' mentions 'decimal' type:
```edgeql
select <bigint>'12345678901234567890n'
# InvalidValueError: invalid input syntax for type std::decimal: "12345678901234567890n"
```

msullivan:
This turned out to be much more interesting than a simple typo: since
our bigint type is really just a numeric (big *decimal*) with a
constraint on it, if a cast to bigint failed, the postgres error
message would mention numeric, which we would map to std::decimal.

Fix this by catching the error on the cast to numeric and reraising
our own error in that case. Some sources on the internet suggested
this might be slow, so I measured the different versions I tried:
having a separate cast_to_decimal_or_null function that did the error
handling was a slowdown, but inlining that and making the whole thing
plpgsql was actually a 2x speedup.

Fixes #3737.

### DIFF
--- a/edb/pgsql/metaschema.py
+++ b/edb/pgsql/metaschema.py
@@ -522,7 +522,7 @@ class StrToBigint(dbops.Function):
                     NULL::numeric,
                     'invalid_text_representation',
                     msg => (
-                        'invalid syntax for type edgedb.bigint_t: '
+                        'invalid input syntax for type edgedb.bigint_t: '
                         || quote_literal(val)
                     )
                 );
@@ -553,7 +553,7 @@ class StrToDecimal(dbops.Function):
                     NULL::numeric,
                     'invalid_text_representation',
                     msg => (
-                        'invalid syntax for type numeric: '
+                        'invalid input syntax for type numeric: '
                         || quote_literal(val)
                     )
                 )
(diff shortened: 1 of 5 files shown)


tests/test_edgeql_casts.py
```diff
@@ -1352,6 +1352,13 @@ async def test_edgeql_casts_numeric_06(self):
             [1],
         )
 
+    async def test_edgeql_casts_numeric_07(self):
+        async with self.assertRaisesRegexTx(
+                edgedb.InvalidValueError,
+                r'invalid input syntax for std::bigint'):
+            await self.con.query_single(
+                'SELECT <bigint>"100000n"')
```
1st1, line 1360: Add a test for `<decimal>'12313.132n'`?

edb/pgsql/metaschema.py
```diff
@@ -541,7 +553,7 @@ class StrToDecimal(dbops.Function):
                     NULL::numeric,
                     'invalid_text_representation',
                     msg => (
-                        'invalid syntax for numeric: '
+                        'invalid syntax for type numeric: '
```
1st1, line 556: Does it make sense to have different implementations for bigint/decimal casts?

msullivan: They have different behavior

msullivan:
Alright, I also made all of the cast failure messages consistent

edb/pgsql/metaschema.py
1st1: I see that, but the mechanisms are different -- one checks for NaN and other for exceptions. Is that because of how we implement the bigint domain (i.e. it raises?)

1st1 (approved):

PR #3857, opened 2022-05-13
Expose Tuple.named to introspection.

msullivan:
Context: 

> It appears there's not really a good way to tell if a tuple is a named tuple or not from introspection, but there are separate type descriptors for them, and they get decoded into different types (array vs. object/map) in clients. Currently the query builder assumes that if a introspected tuple type has keys that look like indexes that it's an unnamed tuple, so there's an edge case if the user creates a tuple type like tuple<`0`: str, `1`: str>, then query builder will infer the result type of that to be an array but the actual result will be an object. We can fix this in a slightly hacky way by parsing the name of the tuple, but would it be possible to change edgedb so the introspected keys are null for unnamed tuples?

### DIFF
--- a/edb/lib/schema.edgeql
+++ b/edb/lib/schema.edgeql
@@ -129,6 +129,7 @@ CREATE TYPE schema::TupleElement EXTENDING std::BaseObject {
 
 
 CREATE TYPE schema::Tuple EXTENDING schema::CollectionType {
+    CREATE REQUIRED PROPERTY named -> bool;
     CREATE MULTI LINK element_types EXTENDING schema::ordered
     -> schema::TupleElement {
         CREATE CONSTRAINT std::exclusive;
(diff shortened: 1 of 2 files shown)


edb/lib/schema.edgeql
```diff
@@ -129,6 +129,7 @@ CREATE TYPE schema::TupleElement EXTENDING std::BaseObject {
 
 
 CREATE TYPE schema::Tuple EXTENDING schema::CollectionType {
+    CREATE REQUIRED PROPERTY named -> bool;
```
1st1, line 132: If we have a test for introspecting tuples please enhance it with this prop check

msullivan: We don't specifically have a test that introspects Tuples

(In general there are a lot of specific things in `schema` that don't have tests directly for them. I think that all the mechanisms get tested, but not each individual field. Stuff like exposing a bool field is really just a config change on pretty well tested code paths.)

1st1: fair enough

1st1 (approved):

PR #3865, opened 2022-05-17
Fix ISE when selecting generated alias

### LINKED ISSUE #3863: ISE when selecting generated alias type
- EdgeDB Version: 2.0-dev.6578+local

Steps to Reproduce:

`select`ing any type that EdgeDB generates from an alias (the alias type itself is fine) causes an ISE with no message:
```
select default::__AwardAlias2__winner;

InternalServerError:
```

Schema: https://github.com/edgedb/edgedb/blob/master/tests/schemas/cards.esdl

msullivan:
Fixes #3863.

### DIFF
--- a/edb/edgeql/compiler/setgen.py
+++ b/edb/edgeql/compiler/setgen.py
@@ -298,6 +298,13 @@ def compile_path(expr: qlast.Path, *, ctx: context.ContextLevel) -> irast.Set:
 
                 if (stype.get_expr_type(ctx.env.schema) is not None and
                         stype.get_name(ctx.env.schema) not in ctx.view_nodes):
+                    if not stype.get_expr(ctx.env.schema):
+                        raise errors.InvalidReferenceError(
+                            f"cannot refer to alias link helper type "
+                            f"'{stype.get_name(ctx.env.schema)}'",
+                            context=step.context,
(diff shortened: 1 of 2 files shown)


edb/edgeql/compiler/setgen.py
```diff
@@ -298,6 +298,13 @@ def compile_path(expr: qlast.Path, *, ctx: context.ContextLevel) -> irast.Set:
 
                 if (stype.get_expr_type(ctx.env.schema) is not None and
                         stype.get_name(ctx.env.schema) not in ctx.view_nodes):
+                    if not stype.get_expr(ctx.env.schema):
+                        raise errors.InvalidReferenceError(
+                            f"cannot refer to alias link helper type "
```
elprans, line 303: Should we pretend it doesn't exist instead?

msullivan: I considered that, but it's visible in schema reflection, and if anybody is trying to select it they probably saw it in schema reflection, so this is a more useful error?

I don't feel strongly though

elprans: OK, fine by me.

elprans (approved):

PR #3875, opened 2022-05-18
Support client side routing in UI

jaclarke:
Handle unknown paths by responding with `index.html` file instead of 404 error, and let client side handle actual 404 errors in the UI.

### DIFF
--- a/edb/graphql/extension.pyx
+++ b/edb/graphql/extension.pyx
@@ -125,6 +125,13 @@ async def handle_request(
                         raise TypeError(
                             '"globals" must be a JSON object')
 
+        elif request.method == b'OPTIONS':
+            response.status = http.HTTPStatus.NO_CONTENT
+            response.custom_headers['Access-Control-Allow-Methods'] = \
+                'POST, GET, OPTIONS'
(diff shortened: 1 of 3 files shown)


edb/graphql/extension.pyx
```diff
@@ -125,6 +125,13 @@ async def handle_request(
                         raise TypeError(
                             '"globals" must be a JSON object')
 
+        elif request.method == b'OPTIONS':
+            response.status = http.HTTPStatus.NO_CONTENT
+            response.custom_headers['Access-Control-Allow-Methods'] = \
+                'POST, GET, OPTIONS'
+            response.custom_headers['Access-Control-Max-Age'] = '86400'
+            return
+
```
1st1, line 134: If we're hiding GraphQL from UI (see https://github.com/edgedb/edgedb-studio/pull/17#issuecomment-1158491258) this change is no longer needed?

1st1 (commented):
@jaclarke The commit message isn't describing the changes at all and I'm failing to understand why they are needed. Can you write a better commit message? Also the GraphQL part isn't going to be needed anymore, I think.

fantix (approved):

elprans (approved):

PR #3876, opened 2022-05-18
Graphql with strawberry

rednafi:
The images aren't working because they are stored in `edgedb.com` and the site is not deployed to Vercel yet. I don't have access to that.


vpetrovykh (commented):
How are the shapes in EdgeQL queries affected by the GraphQL API?

colinhacks (approved):

PR #3879, opened 2022-05-18
Add an environment variable to configure the cython opt level

msullivan:
make cython is a *lot* faster on -O0 (14s vs 68s), so setting -O0
when doing iterative developement will make life much easier.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -124,7 +124,11 @@ EXTRA_DEPS = {
     'build': BUILD_DEPS,
 }
 
-EXT_CFLAGS = ['-O2']
+EXT_CFLAGS: List[str] = []
+if flag := os.getenv('EDGEDB_OPT_CFLAG'):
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -124,7 +124,11 @@
     'build': BUILD_DEPS,
 }
 
-EXT_CFLAGS = ['-O2']
+EXT_CFLAGS: List[str] = []
+if flag := os.getenv('EDGEDB_OPT_CFLAG'):
```
elprans, line 128: We already have `EDGEDB_DEBUG` env var down below that affects Cython compilation, perhaps we can reuse that?

msullivan: Setting that actually slows down compilation, though, from 14s to 20s relative to just passing -O0

elprans (approved):

PR #3880, opened 2022-05-19
Don't discard globals on a COMMIT

msullivan:


tests/test_server_proto.py
```diff
@@ -517,6 +517,12 @@ async def test_server_proto_set_global_01(self):
 
         await self.con.query('reset global glob')
 
+    async def test_server_proto_set_global_02(self):
+        await self.con.execute('START TRANSACTION')
+        await self.con.execute('set global glob := 1')
```
elprans, line 522: This makes me wonder if we want to support transaction-scoped globals.

elprans (approved):

tests/test_server_proto.py
msullivan: I think that in practice with the new protocol, this won't matter at all?

msullivan: Since setting globals will be done on a language binding level and not via `SET GLOBAL`

elprans: Sure, we can circle back to this when and if this becomes needed at the language level.

PR #3881, opened 2022-05-19
Range types WIP

vpetrovykh:
This is a draft mostly to get some feedback on how we want to tackle certain technical aspects of range types.

### DIFF
--- a/edb/edgeql/compiler/casts.py
+++ b/edb/edgeql/compiler/casts.py
@@ -111,6 +111,13 @@ def compile_cast(
         return _cast_array(
             ir_set, orig_stype, new_stype, srcctx=srcctx, ctx=ctx)
 
+    elif (
+        orig_stype.is_range()
+        and not s_types.is_type_compatible(
+            orig_stype, new_stype, schema=ctx.env.schema)
+    ):
+        raise NotImplementedError
+
     elif orig_stype.issubclass(ctx.env.schema, new_stype):
         # The new type is a supertype of the old type,
         # and is always a wider domain, so we simply reassign
--- a/edb/edgeql/compiler/inference/types.py
+++ b/edb/edgeql/compiler/inference/types.py
@@ -178,6 +178,9 @@ def __infer_type_introspection(
     elif irtyputils.is_tuple(ir.typeref):
         return cast(s_objtypes.ObjectType,
                     env.schema.get('schema::Tuple'))
+    elif irtyputils.is_range(ir.typeref):
+        return cast(s_objtypes.ObjectType,
+                    env.schema.get('schema::Range'))
     else:
         raise errors.QueryError(
             'unexpected type in INTROSPECT', context=ir.context)
--- a/edb/edgeql/compiler/normalization.py
+++ b/edb/edgeql/compiler/normalization.py
@@ -517,9 +517,10 @@ def compile_TypeName(
     if isinstance(node.maintype, qlast.ObjectRef):
         # This is a specific path root, resolve it.
         if (not node.maintype.module and
-                # maintype names 'array' and 'tuple' specifically
+                # maintype names 'array', 'tuple', and 'range' specifically
                 # should also be ignored
-                node.maintype.name not in {'array', 'tuple', *localnames}):
+                node.maintype.name not in {'array', 'tuple', 'range',
+                                           *localnames}):
             maintype = schema.get(
                 node.maintype.name,
                 default=None,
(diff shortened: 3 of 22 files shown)


vpetrovykh:
> PostgreSQL comes with the following built-in range types:
> 
> * `int4range` — Range of integer, int4multirange — corresponding Multirange
> * `int8range` — Range of bigint, int8multirange — corresponding Multirange
> * `numrange` — Range of numeric, nummultirange — corresponding Multirange
> * `tsrange` — Range of timestamp without time zone, tsmultirange — corresponding Multirange
> * `tstzrange` — Range of timestamp with time zone, tstzmultirange — corresponding Multirange
> * `daterange` — Range of date, datemultirange — corresponding Multirange

They have their own OID and we probably want to map corresponding `range<blah>` types in EdgeDB to those. Unlike arrays, there does not seem to be a nice reference to them in `pg_type` like a `typarray` column for arrays. So we might want to pre-generate ranges for our built-in scalars ourselves and have a few more built-in range types (with stable OIDs) than Postgres.

vpetrovykh:
I haven't touched any aspect of this that relates to protocol when sending range values over the wire (I assume we send them as Postgres would in binary format with minimal meddling, though). Our clients will need to learn about these types, too. That's another aspect that this PR does not touch at the moment. In practical sense, it's useful for testing to have clients that can receive these values, but not crucial as values can be compared inside EdgeDB instead.

vpetrovykh:
Integer ranges are peculiar in Postgres in that they always try to get into the "[a,b)" format, no matter how you actually specify the boundaries. Since integers are discrete, this trick is generally always possible. E.g.
```
V2f145d6d13_edgedb=# select int8range(1, 4, '[]');
 int8range 
-----------
 [1,5)
(1 row)

V2f145d6d13_edgedb=# select int8range(1, 4, '(]');
 int8range 
-----------
 [2,5)
(1 row)
```
This does not affect unbounded ranges as infinity is not included in the bound:
```
V2f145d6d13_edgedb=# select int8range(NULL, 4);
 int8range 
-----------
 (,4)
(1 row)

V2f145d6d13_edgedb=# select lower_inc(int8range(NULL, 4));
 lower_inc 
-----------
 f
(1 row)
```
This feature may catch someone by surprise, but is not fundamentally wrong as the produced ranges are mathematically equivalent. 

This is not happening for a contiguous type:
```
V2f145d6d13_edgedb=# select numrange(1.1, 4.4);
 numrange  
-----------
 [1.1,4.4)
(1 row)

V2f145d6d13_edgedb=# select numrange(1.1, 4.4, '()');
 numrange  
-----------
 (1.1,4.4)
(1 row)
```

vpetrovykh:
What's our formalism for declaring a custom range type? At the moment we can just start with ranges for built-in scalars, but do we want to have more customization available in principle? Is there going to be a `create ... type` that does this job?

Currently all built-in range types are declared _implicitly_, much like arrays, but we might want explicit declarations even for `std` to have proper control over `canonical` or `subtype_diff` functions (`subtype_diff`, for example has to return a `float8` in all cases, so just automarically grabbing `-` operator won't quite work and it makes for efficient indexing, so omitting it is not a great idea).

So would these be a new type, like `create range type Foo extending range<float64>` or would they be a flavour of `create scalar type` like enums currently are?

vpetrovykh:
Casting.

1) We obviously want a textual representation of our intervals and the ability to cast to and from `str`. We also probably want that to be consistent with `json`. To disambiguate between an array `[1.2, 3.4]` and `range(1.2, 3.4, inchigh := true)` we can make the string representation simply be identical to a constructor call. That's verbose, so maybe this could be a good use for `[a..b)` format instead.

2) Casting between range types is evidently not supported in Postgres:
   ```
   V2f145d6d13_edgedb=# select int4range(1, 4) :: int8range;
   ERROR:  cannot cast type int4range to int8range
   LINE 1: select int4range(1, 4) :: int8range;
   ```
   This is a niche thing, so we can probably omit casting between ranges of compatible types (like within integer ranges or within float ranges). In principle, though, we could define casts for these if this kind of convenience is needed.

elprans:
> This is a draft mostly to get some feedback on how we want to tackle certain technical aspects of range types.

There is a reverse relationship via `pg_type.rngsubtype`.

> What's our formalism for declaring a custom range type?

There is none, range types are supposed to be automatically defined for all point scalars.

> We obviously want a textual representation of our intervals and the ability to cast to and from `str`.

Um, we don't have text I/O for other container types, so I think the need to support direct string cast for range types isn't as obvious.  I would punt this.

> Casting between range types is evidently not supported in Postgres

This is not a problem.  There are other casts that aren't supported directly which we implement via eta-expansion (e.g. tuples etc).

PR #3883, opened 2022-05-19
Fix some scoping issues for singleton set literals

msullivan:


elprans (approved):
Curious in what way is `newscope(fenced=True)` is different from wrapping into a `select` here?

msullivan:
> Curious in what way is `newscope(fenced=True)` is different from wrapping into a `select` here?

Didn't have a proper answer to this so wasn't merging, but I dug it up:

If we just use newscope like that, the path_id of the set will just be the path_id of whatever the internal bit is. Which means that if we have `{z}.foo`, `{z}` will have the path_id `z` and then the whole expression `z->foo`, which makes `z` visible, incorrectly.

msullivan:
(And I did a quick audit for other similar issues and didn't find any)

PR #3884, opened 2022-05-19
Cache sets in card inference independent of their call sites

### LINKED ISSUE #3859: InternalServerError: maximum recursion depth exceeded in comparison
- EdgeDB Version: 2.0-dev.6578+local

Steps to Reproduce:

```edgeql
SELECT default::Comment {
  issue := assert_exists((SELECT .issue {
    references := (SELECT .references LIMIT 1),
    priority := (SELECT .priority LIMIT 1),
    related_to := (SELECT .related_to LIMIT 1),
    status := (SELECT .status LIMIT 1),
    owner := (SELECT .owner LIMIT 1),
    time_spent_log := (SELECT .time_spent_log LIMIT 1),
    watchers := (SELECT .watchers LIMIT 1),
  } LIMIT 10)),
}

# InternalServerError: maximum recursion depth exceeded in comparison
```

Schema: https://github.com/edgedb/edgedb/blob/master/tests/schemas/issues.esdl

msullivan:
Sets have a semi-intrinsic cardinality (though for aliases it can
still depend on positions in the tree), and we want to cache that
independently from the cardinality of a *reference to a set*.

Importantly, we don't want to recheck everything whenever `singletons`
changes.

Fixes #3859.

### DIFF
--- a/edb/edgeql/compiler/inference/cardinality.py
+++ b/edb/edgeql/compiler/inference/cardinality.py
@@ -497,17 +497,34 @@ def _infer_set(
     scope_tree: irast.ScopeTreeNode,
     ctx: inference_context.InfCtx,
 ) -> qltypes.Cardinality:
-    result = _infer_set_inner(
-        ir, is_mutation=is_mutation,
-        scope_tree=scope_tree, ctx=ctx)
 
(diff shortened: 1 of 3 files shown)


elprans (approved):
Good fix!

PR #3887, opened 2022-05-20
Fix a cross-type constraint violation bug

msullivan:
An incorrect optimization would incorrectly decide that an inherited
constraint didn't matter. So in a hierarchy {X, Y} < B < A,
with a constraint on a field in A,
if we try to insert to both X and Y, we will find a constraint
on the nearest common ancestor B and would then decide that the
constraint didn't matter because it was declared not-delegated on
an ancestor, A.

Honestly I can't remember what I was thinking with that logic, which
seems clearly wrong.

### DIFF
--- a/edb/edgeql/compiler/conflicts.py
+++ b/edb/edgeql/compiler/conflicts.py
@@ -238,11 +238,6 @@ def _constr_matters(
     return (
         not constr.generic(schema)
         and not constr.get_delegated(schema)
-        and (
-            constr.get_owned(schema)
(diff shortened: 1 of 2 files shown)


elprans (approved):

msullivan:
Aha, I figured out what the optimization was for: it was sound in the original context that `_constr_matters` was used, which was a case where all ancestor constraints were processed also. Add a flag.

elprans (approved):

PR #3893, opened 2022-05-24
Update nextjs.rst

MangoPieface:
It resolves the example which currently causes the page to break. You can either update to the suggested change above or change the export to:

    export default HomePage


MangoPieface:
Adding a couple more finds into this PR.

PR #3894, opened 2022-05-24
Make INSERT UNLESS CONFLICT/access policy behavior consistent

msullivan:
When doing an INSERT UNLESS CONFLICT that conflicts with objects that
are not visible, consistently raise a constraint error.

Currently we raise a constraint error in cases where we can't use
postgres ON CONFLICT, such as if the body contains an insert to a
single link, but succeed with an empty set in ELSE otherwise.

I went with failing in both cases instead of succeding in both cases
for a couple reasons:
 1. The implementation of always failing is much more straightforward.
    Always succeeding is I think doable, by doing a conflict select
    that ignores access policies. The ELSE branch, then, would need
    to make sure to reapply the access policies.
 2. I don't like having the ELSE branch trigger but having no
    objects appear in it.

### DIFF
--- a/edb/pgsql/compiler/dml.py
+++ b/edb/pgsql/compiler/dml.py
@@ -816,6 +816,13 @@ def insert_needs_conflict_cte(
     if on_conflict.always_check or ir_stmt.conflict_checks:
         return True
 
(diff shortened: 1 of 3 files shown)


elprans (approved):
We need to document this behavior properly.

PR #3895, opened 2022-05-24
Support externally specified object ids

### LINKED ISSUE #3510: External UUIDs
Let's say I have data that already has a UUID from another source. Could I use that UUID as the id, or will edgeDB still create another UUID?

> Currently, it's prohibited.  Primarily because EdgeDB does not actually verify id uniqueness, so we can only trust ids we generate ourself. - @elprans

makes sense. I think this would be a good thing to have eventually though after a mechanism is added to verify uniqueness
I have data right now that has 2 unique ids from its previous and current data source. If I transition to edgeDB I will have to maintain these keys and also have a 3rd. Data syncs become more problematic because I have to either write back edgeDB id to my previous data source to know what exists already and what doesn't or use a different identifying field and do a insert of update statement. I see this becoming more problematic the more times one has to migrate their data between services that all generate their own unique identifiers.

> Verifying uniqueness is quite easy: `select BaseObject filter .id = <uuid>$id`. The problem is that it might be quite slow if you have a large database.  I hear your concern, though, so we'll think of something.
> Please open an issue - @elprans

msullivan:
Most of the work here is in inserting checks to prevent duplicate
`id`s, since we don't have a trigger on `id` for performance
reasons. We use the conflict mechanism for this. The dodgy bit is that
we have to do it with access policies disabled, to make sure they don't
hide something.

Fixes #3510.

### DIFF
--- a/edb/edgeql/compiler/conflicts.py
+++ b/edb/edgeql/compiler/conflicts.py
@@ -187,8 +187,18 @@ def _compile_conflict_select(
                 left=lhs, right=rhs,
             ))
 
-    insert_subject = qlast.Path(steps=[
-        s_utils.name_to_ast_ref(subject_typ.get_name(ctx.env.schema))])
+    # If the type we are looking at is BaseObject, then this must a
+    # conflict check we are synthesizing for an explicit .id. We need
+    # to ignore access policies in that case, since there is no
+    # trigger to back us up.
+    ignore_rewrites = (
+        str(subject_typ.get_name(ctx.env.schema)) == 'std::BaseObject')
+    if ignore_rewrites:
+        assert not obj_constrs
+        assert len(constrs) == 1 and len(constrs['id'][1]) == 1
+    insert_subject = ctx.create_anchor(setgen.class_set(
+        subject_typ, ignore_rewrites=ignore_rewrites, ctx=ctx
+    ))
 
     for constr in obj_constrs:
         subjectexpr = constr.get_subjectexpr(ctx.env.schema)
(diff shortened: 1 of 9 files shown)


edb/schema/pointers.py
```diff
@@ -664,8 +664,14 @@ def is_property(self, schema: s_schema.Schema) -> bool:
...
 
-    def is_protected_pointer(self, schema: s_schema.Schema) -> bool:
-        return self.get_shortname(schema).name in {'id', '__type__'}
+    def is_protected_pointer(
+        self, exprtype: s_types.ExprType, schema: s_schema.Schema
+    ) -> bool:
+        sn = self.get_shortname(schema).name
+        return (
+            sn == '__type__'
+            or (sn == 'id' and not exprtype.is_mutation())
```
elprans, line 673: Hm, I'd explicitly check for `Insert` here, because mutating `id` in `UPDATE` should remain prohibited.

edb/edgeql/compiler/conflicts.py
```diff
@@ -187,8 +187,18 @@ def _compile_conflict_select(
                 left=lhs, right=rhs,
             ))
 
-    insert_subject = qlast.Path(steps=[
-        s_utils.name_to_ast_ref(subject_typ.get_name(ctx.env.schema))])
+    # If the type we are looking at is BaseObject, then this must a
+    # conflict check we are synthesizing for an explicit .id. We need
+    # to ignore access policies in that case, since there is no
+    # trigger to back us up.
```
elprans, line 193: I'd also comment that this is safe to do, because we prevent you from inserting into an abstract type.

edb/schema/pointers.py
1st1: Do we have a test that it's prohibited? I think it'd make sense to add a few tests ensuring we have sane custom ID semantics.

msullivan: I made it `is_mutation` so that we get `cannot update property 'id': it is declared as read-only` message on update, which I thought was a better message. I can add a test for that, or I can change it

elprans: Oh, I forgot that there's another check down the line that protects it. It's fine then.

Issue #3877, opened 2022-05-18
Server crashes on startup if `--max-backend-connections` value is much lower than the number of databases


jaclarke:
- EdgeDB Version: 2.0-dev.6597+local
- OS Version: WSL 2 Ubuntu 18.04

Steps to Reproduce:

1. Have an instance with a large number of databases. Bug initally occurred after running `edb inittestdb`, but it can also be reproduced by creating a number of empty databases, eg:
```
create database db01;
create database db02;
...
create database db24;
create database db25;
```
2. Start server with `--max-backend-connections` set to a low value:
```
$ edb server --max-backend-connections=4
INFO 25489 2022-05-18T20:39:42.997 edb.server: starting EdgeDB server 2.0-dev.6597+local, built on 2022-05-18T19:39Z from revision fdba24ad4 (2022-05-18T01:05Z) for x86_64-unknown-linux-gnu
INFO 25489 2022-05-18T20:39:42.997 edb.server: development mode active
INFO 25507 2022-05-18T20:39:43.961 postgres: starting PostgreSQL 14.2 on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0, 64-bit
INFO 25507 2022-05-18T20:39:43.964 postgres: listening on Unix socket "/home/james/.local/share/edgedb/_localdev/.s.PGSQL.5432"
INFO 25507 2022-05-18T20:39:44.042 postgres: database system was shut down at 2022-05-18 20:39:40 BST
CRITICAL 25507 2022-05-18T20:39:44.043 postgres: the database system is starting up
INFO 25507 2022-05-18T20:39:44.241 postgres: database system is ready to accept connections
INFO 25489 2022-05-18T20:39:46.338 edb.server: Connection established to backend database: edgedb
INFO 25489 2022-05-18T20:39:46.338 edb.server: Connection established to backend database: db02
INFO 25489 2022-05-18T20:39:46.339 edb.server: Connection established to backend database: db01
===================================== Exception occurred: sorry, too many clients already ======================================
1. edb.server.pgcon.errors.BackendError: sorry, too many clients already

---------------------------------------------------------- Traceback -----------------------------------------------------------
    /home/james/dev/edgedb/edb/server/connpool/pool.py, line 514, in _connect
        > conn = await self._connect_cb(block.dbname)
    /home/james/dev/edgedb/edb/server/server.py, line 344, in _pg_connect
[...]

PR #3901, opened 2022-05-25
Fix backend connection pool over-acquire

elprans:
There is an off-by-one error in `acquire_pgcon` that allows server to
attempt to acquire a connection over the allowed
`max-backend-connections` limit.

Fixes: #3877


1st1 (approved):

fantix (approved):
Thanks for the fix!

PR #3902, opened 2022-05-25
Perform database introspection lazily

elprans:
Currently, server startup time is proportional to the number of
databases in the instance, so when this number is large, startup will be
unnecessarily slow.  Avoid this by postponing the database introspection
until the first connection to it.

### DIFF
--- a/edb/server/dbview/dbview.pyx
+++ b/edb/server/dbview/dbview.pyx
@@ -16,6 +16,7 @@
 # limitations under the License.
 #
 
+import asyncio
 import base64
 import json
 import os.path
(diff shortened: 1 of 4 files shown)


edb/server/dbview/dbview.pyx
```diff
@@ -76,17 +77,22 @@ cdef class Database:
         self._index = index
         self._views = weakref.WeakSet()
 
+        self._introspection_lock = asyncio.Lock()
```
1st1, line 80: Can we create it lazily?

1st1 (approved):

edb/server/dbview/dbview.pyx
elprans: Why? It's really cheap.

PR #3907, opened 2022-05-26
Refuse to start if we detect Postgres running in the target data dir

elprans:
A running Postgres cluster in the target data directory is normally an
indication that there was an unclean EdgeDB shutdown that left orphan
subprocesses behind.


fantix:
The diff looks good to me, but it doesn't match the PR title?

fantix (approved):

PR #3909, opened 2022-05-26
Fix floordiv and modulus operators.

vpetrovykh:
The floordiv (`//`) and modulus (`%`) operators had some corner cases
where they would produce the wrong result or an error due to the
implementation calculations.

Add some fuzztests for these operators.

### DIFF
--- a/edb/lib/std/25-numoperators.edgeql
+++ b/edb/lib/std/25-numoperators.edgeql
@@ -2077,11 +2077,12 @@ std::`/` (l: std::decimal, r: std::decimal) -> std::decimal {
 
 # FLOORDIV
 
-# PostgreSQL uses truncation division, so the -12 % 5 is -2, because
-# -12 // 5 is -2, but EdgeQL uses floor division, so -12 // 5 is -3,
-# and so -12 % 5 must be 3.  The correct divmod behavior is implemented
-# in the C extension functions, with the exception of std::decimal,
-# which is implemented using simple formulae instead.
+# PostgreSQL uses truncation division, so the -12 % 5 is -2, because -12 // 5
+# is -2, but EdgeQL uses floor division, so -12 // 5 is -3, and so -12 % 5
+# must be 3. The correct divmod behavior is implemented via the floor
+# function working specifically with numeric type. The numeric value needs to
+# be forced into using arbitrary precision by getting multiplied by
+# 1.0::numeric.
 
 CREATE INFIX OPERATOR
 std::`//` (n: std::int16, d: std::int16) -> std::int16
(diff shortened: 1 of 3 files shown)


vpetrovykh:
Do I need to merge this into some other branch so that it goes into 2.0 rather than the next 1.x?

vpetrovykh:
The modulus operator had some edge cases of overflow errors:
```
edgedb> select 2000000 %  9223372036854770000;
edgedb error: NumericOutOfRangeError: std::int64 out of range
```
None of the inputs are out of range, nor is the result out of range, but the intermediate calculation is: `((n % d) + d) % d`. Specifically the `(n % d) + d` part can be out of range for large values of `d`. So for now I just upcast the integers and then cast the results back.

The idea of using modulus to perform `//` runs into the problem of overflow errors in the modulus operator. So I reverted to the backup idea of using `floor(1.0::numeric * "n"::numeric / "d"::numeric)`.

elprans:
> Do I need to merge this into some other branch so that it goes into 2.0 rather than the next 1.x?

master is 2.0

tests/test_edgeql_expressions.py
```diff
@@ -879,6 +880,168 @@ async def test_edgeql_expr_mod_01(self):
...
+    async def test_edgeql_expr_mod_04(self):
+        # Fuzztest of int64 //
+        MAX = 9223372036854775807
+        vals = []
+        for i in range(500):
+            a = random.randrange(-MAX, MAX)
+            b = random.randrange(-MAX, MAX)
+            vals.append([i, a, b, a // b])
+
+        await self.assert_query_result(
```
msullivan, line 970: `assert_query_result` takes a `variables` arg; maybe it would be better to use that instead of building a big query?

msullivan (approved):
This looks good to me, pending your experiments with non-upcast variants.

tests/test_edgeql_expressions.py
vpetrovykh: Yeah, I've used variables and it also sped things up drastically.

edb/lib/std/25-numoperators.edgeql
```diff
@@ -2094,7 +2095,8 @@ std::`//` (n: std::int16, d: std::int16) -> std::int16
     # the common attributes of the SQL division operator while
     # overriding the main operator function.
     USING SQL OPERATOR r'/';
-    USING SQL 'SELECT floor("n"::numeric / "d"::numeric)::int2';
+    USING SQL
+        'SELECT floor(1.0::numeric * "n"::numeric / "d"::numeric)::int2';
```
elprans, line 2099: Now that we have fast and working `%`, should we use it here instead of `floor`?

vpetrovykh: We would need a way of handling the case when `d` is close to minimum integer value and subtracting the remainder would push it beyond the valid range. Similar issue as for `%`, but a bit harder to actually tackle, because we're supposed to unconditionally subtract the remainder.

PR #3924, opened 2022-05-30
Update tutorial to use consistent naming

boucher:
There are a few instances where the names being used don't quite match up. Since one of these is a graphic which I can't recreate, I've updated all the names to match what's shown in the graphic. Perhaps you'd prefer a different convention and would rather just update the graphic.

I also added a missing react import which isn't super important but did leave me scratching my head for a second.

### DIFF
--- a/docs/guides/tutorials/nextjs.rst
+++ b/docs/guides/tutorials/nextjs.rst
@@ -321,14 +321,20 @@ we inserted earlier.
 
 To fetch these from the homepage, we'll use ``useState``, ``useEffect``, and
 the built-in ``fetch`` API. At the top of the ``HomePage`` component in
-``pages/index.tsx``, replace the static data with .
+``pages/index.tsx``, replace the static data and add the missing imports.
 
 .. code-block:: tsx-diff
 
      // pages/index.tsx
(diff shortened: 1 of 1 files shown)


docs/guides/tutorials/nextjs.rst
```diff
@@ -460,12 +466,12 @@ Back in ``pages/index.tsx``, lets update our code to use the inferred
...
-  +    const [posts, setPosts] = useState<GetPosts | null>(null);
+  +    const [posts, setPosts] = useState<Posts | null>(null);
        // ...
 
      }
 
-Now, when we update our ``getPosts`` query, the type of our dynamically loaded
+Now, when we update our ``selectPosts`` query, the type of our dynamically loaded
 ``posts`` variable will update automatically—no need to keep our type
 definitions in sync with our API logic!
```
1st1, line 476:
```suggestion
Now, when we update our ``selectPosts`` query, the type of our dynamically
loaded ``posts`` variable will update automatically—no need to keep
our type definitions in sync with our API logic!
```

colinhacks (approved):

PR #3930, opened 2022-06-01
Expose 'body' and 'language' fields on schema::Function

msullivan:
'nativecode' becomes 'body', and gets exposed, while 'language' is
exposed and modified so that SQL is exported as 'Builtin'.

### DIFF
--- a/edb/lib/schema.edgeql
+++ b/edb/lib/schema.edgeql
@@ -49,6 +49,9 @@ CREATE SCALAR TYPE schema::AccessPolicyAction
 CREATE SCALAR TYPE schema::AccessKind
     EXTENDING enum<`Select`, UpdateRead, UpdateWrite, `Delete`, `Insert`>;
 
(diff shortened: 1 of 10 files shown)


edb/lib/schema.edgeql
```diff
@@ -49,6 +49,9 @@ CREATE SCALAR TYPE schema::AccessPolicyAction
 CREATE SCALAR TYPE schema::AccessKind
     EXTENDING enum<`Select`, UpdateRead, UpdateWrite, `Delete`, `Insert`>;
 
+CREATE SCALAR TYPE schema::FuncLanguage
+    EXTENDING enum<EdgeQL, Builtin>;
```
elprans, line 53: I think we should probably _not_ use an enum here, at least not initially, because it might turn out to be too inflexible for when (and if) we allow functions to be written in a language of choice and compiled into common binary format.

msullivan:
Closing in favor of #3944

PR #3948, opened 2022-06-08
Add some useful arithemtic for local date/time types.

vpetrovykh:
Add subtraction operator for pairs of `local_datetime`, `local_date`,
`local_time`.
Add normalization functions for `relative_duration`.
Add element extraction function `duration_get` for both `duration` and
`relative_duration`.


edb/lib/cal.edgeql
```diff
@@ -1177,6 +1415,141 @@ std::datetime_get(dt: cal::local_datetime, el: std::str) -> std::float64
 };
 
 
+CREATE FUNCTION
+std::duration_get(dt: cal::date_duration, el: std::str) -> std::float64
+{
+    CREATE ANNOTATION std::description :=
+        'Extract a specific element of input duration by name.';
+    SET volatility := 'Stable';
```
elprans, line 1423: Why `Stable`?

edb/lib/cal.edgeql
```diff
@@ -266,14 +318,51 @@ cal::date_get(dt: cal::local_date, el: std::str) -> std::float64
...
+    CREATE ANNOTATION std::description :=
+        'Convert 24-hour chunks into days.';
+    SET volatility := 'Immutable';
+    SET force_return_cast := true;
+    USING SQL FUNCTION 'justify_hours';
+};
+
+
+CREATE FUNCTION
+cal::normalize_relative_days(dur: cal::relative_duration)
```
1st1, line 342: Should this be defined for `date_duration` too? Also, why are we adding these functions? Are they essential?

edb/lib/cal.edgeql
```diff
@@ -601,27 +700,71 @@ std::`+` (l: std::duration, r: cal::local_date) -> cal::local_date
...
+    SET commutator := 'std::+';
+    SET force_return_cast := true;
+    USING SQL $$
+        SELECT ("l" + "r")::edgedb.timestamp_t
+    $$;
+};
+
+
+CREATE INFIX OPERATOR
+std::`-` (l: cal::local_date, r: cal::relative_duration) -> cal::local_datetime
```
1st1, line 753: Hm, should we also support multiplication operators for durations? (e.g. date_duration * 2)

edb/lib/cal.edgeql
```diff
@@ -1094,6 +1300,12 @@ CREATE CAST FROM cal::relative_duration TO std::json {
 };
 
 
+CREATE CAST FROM cal::date_duration TO std::json {
+    SET volatility := 'Stable';
```
1st1, line 1304: Can't it be immutable? ditto for other json<>date_duration; they're independent of current db config

edb/lib/cal.edgeql
```diff
@@ -266,14 +318,51 @@ cal::date_get(dt: cal::local_date, el: std::str) -> std::float64
...
+                    'millennium, month, quarter, seconds, week, year."}'
                 )
             )
         END
     $$;
 };
 
 
+CREATE FUNCTION
+cal::normalize_relative_hours(dur: cal::relative_duration)
```
1st1, line 330: I don't like the name -- it goes against our naming convention where the typename is the first part of the name. Should be `duration_normalize_hours`.

edb/lib/cal.edgeql
```diff
@@ -929,19 +1095,44 @@ CREATE INFIX OPERATOR
...
+CREATE INFIX OPERATOR
+std::`-` (l: cal::date_duration, r: cal::date_duration) -> cal::date_duration {
+    CREATE ANNOTATION std::identifier := 'minus';
+    CREATE ANNOTATION std::description :=
+        'Time interval subtraction.';
+    SET volatility := 'Immutable';
+    USING SQL $$
+    SELECT ("l" - "r")::edgedb.date_duration_t;
+    $$;
+};
```
1st1, line 1128: I don't see <, >, <=, >= defined for date_duration... are they defined automatically?

edb/lib/cal.edgeql
vpetrovykh: You're correct, only the version dealing with `timestamptz` is marked as "stable", the rest are "immutable". Fixed.

edb/lib/cal.edgeql
vpetrovykh: Maybe. I'd still be inclined to not consider it "essential" and I'm unsure about the correct way of applying `*`. What does "1 month 1 day 1 hour" * 40 produce? Do the days get normalized? Hours? Maybe `duration` just multiplies the elements as is, whereas `relative_duration` normalizes the hours and convers them into days, but leaves elements greater than days as is.

edb/lib/cal.edgeql
vpetrovykh: The implicit cast from `date_duration` to `relative_duration` makes comparison operators compatible with both types (or even a mix, which is desirable). Since they return `bool` and operate on `interval` under the hood, having specialized versions doesn't really provide any special benefit.

edb/lib/cal.edgeql
elprans: `timestamptz` version should also be `immutable`.  It is stable in Postgres because it has configurable timezones, EdgeDB doesn't.

edb/lib/cal.edgeql
elprans: Postgres' `stable` marking is due to configurability of timezone and interval output format.  EdgeDB has stable timezone (`UTC`), and stable output format, so we should be able to mark these as immutable.

edb/lib/cal.edgeql
vpetrovykh: This is one of those examples where `edgedb.raise` is stable and it's part of the implementation here.

The `edgedb.raise` itself has its own reason why it's [stable](https://github.com/edgedb/edgedb/blob/5d7a2dd3fd72dcb183ec080648d652bf99582c4d/edb/pgsql/metaschema.py#L1023-L1029). Basically we don't want them to be optimized by Postgres in a way that causes it to be evaluated (thus raising the exception) because it uses a constant (NULL) input regardless of the actual query where it's used.

edb/lib/cal.edgeql
```diff
@@ -266,14 +318,51 @@ cal::date_get(dt: cal::local_date, el: std::str) -> std::float64
...
+    CREATE ANNOTATION std::description :=
+        'Convert 30-day chunks into months.';
+    SET volatility := 'Immutable';
+    SET force_return_cast := true;
+    USING SQL FUNCTION 'justify_days';
+};
+
+
+CREATE FUNCTION
+cal::duration_normalize(dur: cal::relative_duration)
```
1st1, line 354: Can you normalize date_duration?

edb/lib/cal.edgeql
1st1: OK, let's not implement `*`, `"1 month 1 day 1 hour" * 40` does seem like an example to figure out in the future.

edb/lib/cal.edgeql
vpetrovykh: Added an overload for normalizing `date_duration` returning `date_duration` as the result.

vpetrovykh: Yes, and I also removed the convenience `duraiton_normalize` function that combines 2 other normalizations.

elprans (approved):

PR #3949, opened 2022-06-08
Implement partial constraints and indexes

### LINKED ISSUE #1610: Add the ability to define partial constraints and indexes
Currently, constraints and indexes are unconditional, i.e. it is impossible to exclude a certain subset of elements from being indexed or constrained.  Having the ability of omitting elements with low selectivity is beneficial for index size, since such query quals wouldn't even use the index at all.  

Adding a similar predicate to constraints is also useful.  For example, currently if one wanted to enforce uniqueness of user logins, the following could be used:

```
type User {
    required property username -> str;
    constraint exclusive on (.username);
}
```

However, if we consider that some user accounts can be marked as deleted (without actually deleting them), the constraint would prevent new users from claiming a previously used name.  A predicate on the constraint would solve this:

```
type User {
    required property username -> str;
    property deleted -> bool;
    constraint exclusive on (.username) except (.deleted);
};
```

I think an exclusion predicate is better than an inclusion predicate, i.e. `constraint ... where (not .deleted)`, but that is up for debate.

For consistency, non-aggregate constraints should also support predicates, in which case `constraint expression on (len(.username) > 5) except (.deleted)` is equivalent to `constraint expression on (len(.username) > 5 or .deleted)`

msullivan:
Closes #1610.

### DIFF
--- a/edb/pgsql/delta.py
+++ b/edb/pgsql/delta.py
@@ -2986,20 +2986,22 @@ class CreateIndex(IndexCommand, adapts=s_indexes.CreateIndex):
         singletons = [subject]
         path_prefix_anchor = ql_ast.Subject().name
 
+        options = qlcompiler.CompilerOptions(
+            modaliases=context.modaliases,
+            schema_object_context=cls.get_schema_metaclass(),
+            anchors={ql_ast.Subject().name: subject},
+            path_prefix_anchor=path_prefix_anchor,
+            singletons=singletons,
+            apply_query_rewrites=not context.stdmode,
+        )
+
         index_expr = index.get_expr(schema)
         ir = index_expr.irast
         if ir is None:
             index_expr = type(index_expr).compiled(
                 index_expr,
                 schema=schema,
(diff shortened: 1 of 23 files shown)


edb/pgsql/delta.py
```diff
@@ -3017,12 +3019,28 @@ def create_index(cls, index, schema, context):
...
+            except_expr = type(except_expr).compiled(
+                except_expr,
+                schema=schema,
+                options=options,
+            )
+        if except_expr:
+            except_tree = compiler.compile_ir_to_sql_tree(
+                except_expr.irast.expr, singleton_mode=True)
+            except_src = codegen.SQLSourceGenerator.to_source(except_tree)
+            except_src = f'({except_src}) is not true'
```
elprans, line 3033: We need to be clear in the documentation that if the predicate returns an empty set, we _allow_ the object to be indexed/constrained.

elprans (approved):

PR #3956, opened 2022-06-09
State over proto

fantix:
Depends on edgedb/edgedb-python#315

### DIFF
--- a/edb/server/compiler/sertypes.py
+++ b/edb/server/compiler/sertypes.py
@@ -39,6 +39,7 @@ from edb.schema import objtypes as s_objtypes
 from edb.schema import scalars as s_scalars
 from edb.schema import schema as s_schema
 from edb.schema import types as s_types
+from edb.ir import statypes
 
 from . import enums
 
@@ -70,6 +71,22 @@ CTYPE_INPUT_SHAPE = b'\x08'
 CTYPE_ANNO_TYPENAME = b'\xff'
 
 EMPTY_BYTEARRAY = bytearray()
+
+
+def _encode_str(data: str) -> bytes:
+    return data.encode('utf-8')
+
+
+def _decode_str(data: bytes) -> str:
+    return data.decode('utf-8')
+
+
+def _encode_bool(data: bool) -> bytes:
+    return b'\x01' if data else b'\x00'
+
+
+def _decode_bool(data: bytes) -> bool:
+    return bool(data[0])
 
 
 def cardinality_from_ptr(ptr, schema) -> enums.Cardinality:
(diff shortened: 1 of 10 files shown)


edb/server/dbview/dbview.pyx
```diff
@@ -380,9 +393,79 @@ cdef class DatabaseConnectionView:
...
+        state = {
+            'module': module,
+            'aliases': list(modaliases.items()),
+            'config': {k: v.value for k, v in session_config.items()},
+            'globals': {k: v.value for k, v in globals_.items()},
+        }
+        return serializer.encode(state)
+
+    cdef decode_state(self, type_id, data, protocol_version):
+        if type_id == b'\0' * 16:
```
msullivan, line 429: Make `b'\0' * 16` a global, probably?

edb/server/compiler/sertypes.py
```diff
@@ -602,11 +661,144 @@ def parse(cls, typedesc: bytes, protocol_version: tuple) -> TypeDesc:
...
+        schema, aliases_array = s_types.Array.from_subtypes(
+            schema, [alias_tuple])
+        state_shape['aliases'] = aliases_array
+
+        # config := cfg::Config { session_cfg1, session_cfg2, ... }
+        schema, config_type = simple_derive_type(
+            schema, 'cfg::Config', 'state_config'
+        )
+        input_shapes[config_type] = config_shape = {}
+        from edb.server.config import get_settings
```
1st1, line 734: Why is this import here? Circular dep?

edb/protocol/messages.py
```diff
@@ -789,6 +797,15 @@ class Execute(ClientMessage):
     expected_cardinality = EnumOf(UInt8, Cardinality,
                                   'Expected result cardinality.')
     command_text = String('Command text.')
+
+    state_typedesc_id = UUID('Updated session state descriptor ID.')
+    state_data = ArrayOf(
```
elprans, line 802: Should this be `FixedArrayOf(1, Dataelement)` then?  Also, what's the rationale behind making this an array?

elprans (approved):

edb/protocol/messages.py
fantix: Oh there wasn't a specific reason - in `Data` message it's also an array of `DataElement`.

elprans: I think there is a bit of a confusion proliferating here.  The `Data` message should not contain an array of `DataElement` either.  The `1` there is _not_ a number of array elements, it's a number of _columns_ in the underlying Postgres tuple.  EdgeDB always generates output tuples with a single element.  I think it will be much less confusing if we simply declare a reserved field in `Data`, and remove `DataElement`.

elprans (approved):

PR #3975, opened 2022-06-20
Upgrade a bunch of casts and functions from "Stable" to "Immutable".

vpetrovykh:
A lot of the casts and functions are labelled as "Stable" because of
either raising an exception or handling timestamptz. However, it seems
to be overzealous most of the time.

### DIFF
--- a/edb/lib/std/30-datetimefuncs.edgeql
+++ b/edb/lib/std/30-datetimefuncs.edgeql
@@ -58,8 +58,7 @@ std::datetime_get(dt: std::datetime, el: std::str) -> std::float64
 {
     CREATE ANNOTATION std::description :=
         'Extract a specific element of input datetime by name.';
-    # date_part of timestamptz is STABLE in PostgreSQL
-    SET volatility := 'Stable';
+    SET volatility := 'Immutable';
     USING SQL $$
     SELECT CASE WHEN "el" IN (
             'century', 'day', 'decade', 'dow', 'doy', 'hour',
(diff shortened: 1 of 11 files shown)


edb/lib/std/30-datetimefuncs.edgeql
```diff
@@ -468,18 +467,15 @@ std::`-` (v: std::duration) -> std::duration {
 
 ## String casts
 
-# Casts from text to any sort of date/time types are all STABLE in
-# PostgreSQL, but it may be the case that our restricted versions that
-# prohibit usage of timezone for non-timezone-aware types are actually
-# IMMUTABLE (as the corresponding casts from those types to text).
 CREATE CAST FROM std::str TO std::datetime {
+    # Stable because the input string can be in any time-zone.
```
elprans, line 471: I would also clarify what this means: a dependence on a mutable external timezone database (e.g. tzdata).

elprans (approved):

edb/lib/std/30-datetimefuncs.edgeql
```diff
@@ -468,18 +472,18 @@ std::`-` (v: std::duration) -> std::duration {
 
 ## String casts
 
-# Casts from text to any sort of date/time types are all STABLE in
-# PostgreSQL, but it may be the case that our restricted versions that
-# prohibit usage of timezone for non-timezone-aware types are actually
-# IMMUTABLE (as the corresponding casts from those types to text).
 CREATE CAST FROM std::str TO std::datetime {
+    # Stable because the input string can be in any time-zone. Time zones are
```
1st1, line 476:
```suggestion
    # Stable because the input string can contain an explicit time-zone. Time zones are
```

edb/lib/std/30-jsonfuncs.edgeql
```diff
@@ -255,112 +255,116 @@ CREATE CAST FROM std::json TO array<anytype> {
...
-    SET volatility := 'Stable';
+    SET volatility := 'Immutable';
     USING SQL $$
     SELECT edgedb.jsonb_extract_scalar(val, 'string');
     $$;
 };
 
 
 CREATE CAST FROM std::json TO std::datetime {
+    # Stable because the input string can be in any time-zone. Time zones are
```
1st1, line 364: ditto

1st1 (approved):
please take a look at my nit comment and then we can merge

PR #3979, opened 2022-06-21
Add support for authentication, formalize edgedb+http protocol

elprans:
This formalizes the binary protocol over HTTP, makes it available at the
`/db/<dbname>` path, and adds support for authentication via JWT
(encrypted and signed).

To test, use edgedb/edgedb-js#372


edb/server/args.py
```diff
@@ -85,8 +85,42 @@ class ServerTlsCertMode(enum.StrEnum):
...
+
+
+class ServerConnTransport(enum.StrEnum):
+
+    HTTP = "HTTP"
+    TCP = "TCP"
+
+
+class ServerAuthMethods:
+
```
1st1, line 101: This is needed just to make typing possible?

edb/server/args.py
```diff
@@ -594,16 +688,28 @@ def _writer(status: str) -> None:
             '`--tls-cert-mode=generate_self_signed`.  The default is `strict`.'
         ),
     ),
+    click.option(
+        '--jws-public-key-file',
+        type=PathPath(),
+        envvar="EDGEDB_SERVER_JWS_PUBLIC_KEY_FILE",
+        help='Specifies a path to a file containing a public key in PEM '
+             'format used to verify JWT signatures.'),
```
1st1, line 696: Can you make them hidden for now?

edb/server/protocol/binary.pyx
```diff
@@ -636,6 +668,81 @@ cdef class EdgeConnection:
...
+                jwt=encoded_token,
+            )
+            token = jwt.JWT(
+                key=skey,
+                algs=["RS256", "ES256"],
+                jwt=decrypted_token.claims,
+            )
+        except jwt.JWException as e:
+            raise errors.AuthenticationError(
+                f'authentication failed: {e.args[0]}'
```
1st1, line 713:
```suggestion
                f'JWT authentication failed: {e.args[0]}'
```

edb/server/protocol/binary.pyx
```diff
@@ -636,6 +668,81 @@ cdef class EdgeConnection:
...
+            ) from None
+        except Exception as e:
+            raise errors.AuthenticationError(
+                f'authentication failed: cannot decode JWT'
```
1st1, line 717: for easier future debug 

```suggestion
                f'authentication failed: cannot decode JWT ({e.args[0]})'
```

edb/server/protocol/protocol.pyx
```diff
@@ -416,91 +431,144 @@ cdef class HttpProtocol:
...
-            if extname == 'edgeql':
-                extname = 'edgeql_http'
+            if db is None:
+                return self._not_found(request, response)
+
+            extname = path_parts[2] if path_parts_len > 2 else None
+
+            if (
+                # Binary proto tunnelled through HTTP
+                extname is None
```
1st1, line 450:
```suggestion
                (extname is None and request.method == b'POST')
```

?

1st1 (approved):

edb/server/protocol/binary.pyx
elprans: These messages will already have "JWT" somewhere in them.

edb/server/protocol/binary.pyx
elprans: A better approach would probably be to log the original error, but still be nebulous in the error message returned to the client to avoid accidentally exposing something that shouldn't be exposed.

edb/server/args.py
elprans: Yes, and also to simplify rendering back to the command-line format.

devunt:
Hi! Does `edgedb+http` protocol supports transaction?

raddevon:
> Hi! Does `edgedb+http` protocol supports transaction?

No, the HTTP protocol is stateless, so each query runs on a brand new connection.

elprans:
That said, you can still send semicolon-separated multi-statement scripts if you need to.

PR #3980, opened 2022-06-21
Allow xfail to distinguish between errors and failures

msullivan:
Errors typically represent ISEs and other things that prevent
execution of the query, while failures typically represent incorrect
results. The latter is quite a bit worse, so modify our xfail to still
fail on those by default.

Happy to change the name if you have a better suggestion.

### DIFF
--- a/edb/tools/test/decorators.py
+++ b/edb/tools/test/decorators.py
@@ -25,7 +25,7 @@ import unittest
 skip = unittest.skip
 
 
-def xfail(reason, *, unless=False, allow_failure=False):
+def _xfail(reason, *, unless=False, allow_failure):
     def decorator(test_item):
         if unless:
             return test_item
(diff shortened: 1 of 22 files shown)


edb/tools/test/decorators.py
```diff
@@ -25,12 +25,13 @@
 skip = unittest.skip
 
 
-def xfail(reason, *, unless=False):
+def xfail(reason, *, unless=False, allow_failure=False):
```
elprans, line 28: Perhaps introduce another decorator for better visibility, e.g. `xerror`?

msullivan: And turn the old xfail into xerror?

elprans: Yeah.

elprans (approved):

PR #3983, opened 2022-06-22
Add range types.

vpetrovykh:
Add range types, basic operators and some helper functions. The
implemented range types are:
- range<int32>
- range<int64>
- range<float32>
- range<float64>
- range<decimal>
- range<datetime>
- range<local_datetime>
- range<local_date>

Partial implementation of RFC 1012.


edb/lib/cal.edgeql
```diff
@@ -1920,3 +1922,160 @@ std::max(vals: SET OF array<cal::date_duration>) -> OPTIONAL array<cal::date_dur
...
 };
+
+
+## Range functions
+
+CREATE FUNCTION
+std::range(
+    lower: optional cal::local_datetime = {},
+    upper: optional cal::local_datetime = {},
+    named only inclower: bool = true,
```
1st1, line 1933:
```suggestion
    named only inc_lower: bool = true,
```

Our "policy" is to include underscores between words in names.

edb/lib/cal.edgeql
```diff
@@ -1920,3 +1922,160 @@ std::max(vals: SET OF array<cal::date_duration>) -> OPTIONAL array<cal::date_dur
...
+                )::timestamp,
+                step::interval
+            )::edgedb.timestamp_t
+    $$;
+};
+
+
+CREATE FUNCTION
+std::range_unpack(
+    val: range<cal::local_date>
```
1st1, line 2011: this should have a second argument -- date_duration

edb/lib/cal.edgeql
```diff
@@ -1920,3 +1922,160 @@ std::max(vals: SET OF array<cal::date_duration>) -> OPTIONAL array<cal::date_dur
...
+                )::timestamp,
+                'P1D'::interval
+            )::edgedb.date_t
+    $$;
+};
+
+
+CREATE FUNCTION
+std::range_unpack(
+    val: range<cal::local_date>,
```
1st1, line 2042: We should also have functions to unpack ranges for ints/floats/etc?

1st1 (commented):
Mostly looks good to me

edb/lib/cal.edgeql
```diff
@@ -1920,3 +1922,160 @@ std::max(vals: SET OF array<cal::date_duration>) -> OPTIONAL array<cal::date_dur
...
+                    )
+                )::timestamp,
+                step::interval
+            )::edgedb.date_t
+    $$;
+};
+
+# Need to cast edgedb.date_t to date in order for the @> operator to work.
+CREATE FUNCTION std::contains(
+    haystack: range<cal::local_date>,
```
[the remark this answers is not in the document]
vpetrovykh: The file layout has to do with sequencing DDL in a way so that we can declare types before their usage in functions, etc. The current layout is what we also do for other `std` functions that apply to `cal` types - we defined the extra ones in the same file as `cal` things.

edb/lib/cal.edgeql
vpetrovykh: We do that in `31-rangefuncs.edgeql`.

vpetrovykh:
I've removed hardcoded UUIDs for the exposed range types, instead I use a similar trick to our built-in scalars to link UUIDs to Postgres types.

edb/schema/types.py
```diff
@@ -2658,6 +3113,12 @@ def labels_compatible(t_a: Type, t_b: Type) -> bool:
             return (
                 not isinstance(t_as, Tuple) and labels_compatible(t_as, t_bs)
             )
+        # elif isinstance(t_a, Range) and isinstance(t_b, Range):
```
elprans, line 3116: Dead code?

edb/schema/types.py
```diff
@@ -2544,6 +2944,53 @@ def get_dummy_expr_field_value(
...
+):
+
+    def get_dummy_expr_field_value(
+        self,
+        schema: s_schema.Schema,
+        context: sd.CommandContext,
+        field: so.Field[Any],
+        value: Any,
+    ) -> Optional[s_expr.Expression]:
+        # XXX: what's this dummy value?
```
elprans, line 2987: This is needed for some switcheroo shenanigans when we do alter propagation in expressions.  Using `range()` should be fine.

edb/schema/types.py
```diff
@@ -2108,6 +2133,370 @@ def get_underlying_schema_class(cls) -> typing.Type[Collection]:
...
+class Range(
+    Collection,
+    s_abc.Range,
+    qlkind=qltypes.SchemaObjectClass.RANGE_TYPE,
+    schema_name='range',
+):
+
+    element_type = so.SchemaField(
+        Type,
+        # We want a low compcoef so that array types are *never* altered.
```
elprans, line 2149:
```suggestion
        # We want a low compcoef so that range types are *never* altered.
```

edb/schema/types.py
```diff
@@ -2108,6 +2133,370 @@ def get_underlying_schema_class(cls) -> typing.Type[Collection]:
...
+        *,
+        name: Optional[s_name.QualName] = None,
+        **kwargs: Any,
+    ) -> typing.Tuple[s_schema.Schema, Range_T]:
+        if len(subtypes) != 1:
+            raise errors.SchemaError(
+                f'unexpected number of subtypes, expecting 1: {subtypes!r}')
+        stype = subtypes[0]
+
+        # XXX: need to make sure only anypoint is allowed here
```
elprans, line 2344: What's stopping you from implementing this check?

vpetrovykh: Implemented

edb/schema/types.py
vpetrovykh: Fixed typo

vpetrovykh:
I've added casts for range types following similar reasoning to their underlying type casts. Hopefully this is the last bit of fixing for ranges.

At the moment we don't have an agreed way of casting them to `str` or `json`. This is something we might want to implement soon because without it the range types are not as useful as props.

PR #3985, opened 2022-06-22
Bring back Parse message

elprans:
Upon further consideration it was decided that an explicit "parse only"
mode deserves a dedicated message rather than a magical type descriptor
id or a boolean flag in `Execute`.

Another aspect of this change is that a mismatching _input_ type
descriptor in `Execute` now causes a `ParameterTypeMismatchError`
instead of treating the condition as a non-error.

### DIFF
--- a/edb/server/protocol/binary.pyx
+++ b/edb/server/protocol/binary.pyx
@@ -1683,6 +1683,11 @@ cdef class EdgeConnection:
 
         buf = self.make_command_data_description_msg(compiled)
 
+        # Cache compilation result in anticipation that the client
+        # will follow up with an Execute immediately.
+        #
+        # N.B.: we cannot rely on query cache because not all units
+        # are cacheable.
         self._last_anon_compiled = compiled
         self._last_anon_compiled_hash = hash(query_req)
 
@@ -1692,6 +1697,7 @@ cdef class EdgeConnection:
     async def execute(self):
         cdef:
             QueryRequestInfo query_req
+            dbview.DatabaseConnectionView dbview
             bytes in_tid
             bytes out_tid
             bytes args
@@ -1704,6 +1710,8 @@ cdef class EdgeConnection:
         args = self.buffer.read_len_prefixed_bytes()
         self.buffer.finish_message()
 
+        dbview = self.get_dbview()
+
         if (
             self._last_anon_compiled is not None and
             hash(query_req) == self._last_anon_compiled_hash and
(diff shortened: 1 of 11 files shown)


edb/server/protocol/binary.pyx
```diff
@@ -1716,27 +1743,24 @@ cdef class EdgeConnection:
...
-            self.debug_print('EXECUTE', query)
+            self.debug_print('EXECUTE', query_req.text())
 
         # Clear the _last_anon_compiled so that the next Execute - if
-        # identical - will always lookup in the cache and honor the `cacheable`
-        # flag to compile the query again. Put it another way, this is the
-        # legacy Execute of the "anonymous" prepared statement.
+        # identical - will always lookup in the cache and honor the
+        # `cacheable` flag to compile the query again. Put it another way,
+        # this is the legacy Execute of the "anonymous" prepared statement.
```
fantix, line 1763:
```suggestion
        # `cacheable` flag to compile the query again.
```

edb/server/protocol/binary.pyx
```diff
@@ -1692,9 +1719,9 @@ cdef class EdgeConnection:
             )
             if query_unit_group is None:
                 if self.debug:
-                    self.debug_print('EXECUTE /REPARSE', query)
+                    self.debug_print('EXECUTE /CACHE MISS', query_req.text())
 
-                compiled = await self._parse(query, query_req)
+                compiled = await self._parse(query_req)
                 self._last_anon_compiled = compiled
                 self._last_anon_compiled_hash = hash(query_req)
```
fantix, line 1726:
```suggestion
```

This is no longer needed as we don't allow parse-only in `Execute`.

edb/testbase/server.py
```diff
@@ -402,11 +403,13 @@ async def init_cluster(
     if init_settings is None:
         init_settings = {}
 
+    log_level = 's' if not debug.flags.server else 'd'
```
fantix, line 406: yay!

fantix (approved):

1st1 (approved):

edb/server/protocol/binary.pyx
```diff
@@ -1881,6 +1894,21 @@ cdef class EdgeConnection:
                         # ERROR message immediately.
                         await self.restore()
 
+                    elif mtype == b'D':
+                        raise errors.BinaryProtocolError(
+                            "Describe message (D) is not supported in "
+                            "protocols greater 0.13")
```
colinhacks, line 1900:
```suggestion
                            "protocols greater than 0.13")
```

PR #3991, opened 2022-06-27
Support local instance with JWT auth

fantix:
Depends on #3875

- [x] Rename to drop public/private terms in JWE/JWS
- [x] Add option to generate JWE/JWS key pairs
- [x] Generate JWT tokens for local UI (done in the CLI edgedb/edgedb-cli#757)

## Cloud

edgedb-jwt-cloud drawio

## Local

edgedb-jwt-local drawio

### DIFF
--- a/edb/server/args.py
+++ b/edb/server/args.py
@@ -43,6 +43,8 @@ MIB = 1024 * 1024
 RAM_MIB_PER_CONN = 100
 TLS_CERT_FILE_NAME = "edbtlscert.pem"
 TLS_KEY_FILE_NAME = "edbprivkey.pem"
+JWE_KEY_FILE_NAME = "edbjwekeys.pem"
+JWS_KEY_FILE_NAME = "edbjwskeys.pem"
 
 
 logger = logging.getLogger('edb.server')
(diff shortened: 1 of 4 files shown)


edb/server/protocol/ui_ext.pyx
```diff
@@ -56,32 +57,49 @@ async def handle_request(
...
+                token = jwt.JWT(
+                    header={"alg": "ECDH-ES", "enc": "A256GCM"},
+                    claims=signed_token.serialize(),
+                )
+                token.make_encrypted_token(ekey)
+                urlpath = request.url.path.decode('ascii')
+                response.status = http.HTTPStatus.TEMPORARY_REDIRECT
+                response.custom_headers['Location'] = (
+                    f'{urlpath}?authToken={token.serialize()}'
+                )
```
fantix, line 96: Instead of doing the redirect here, can we send the token back in the `/ui/instance-info` request above?

elprans: This redirect largely makes the token ineffective, as then nothing would stop a malicious script from fishing out the token and proceeding with making requests to the local instance.  Instead, the CLI must insert the token into the URL it generates in `edgedb ui`.

fantix: Ah ok

PR #4000, opened 2022-06-29
Document different headers

tailhook:
* Binary headers now called `KeyValue`
* JSON headers are now called `Annotations`


tailhook:
Tests are fixed @elprans

elprans (approved):

PR #4001, opened 2022-06-29
Allow multi-statement requests to return data

elprans:
Currently, if a request contains multiple statements it is considered to
be a "script", i.e. it will not return any data messages back to the
caller.  This is a rather arbitrary limitation at this point, and lifting
it not only simplifies the internal logic, but also enables useful
client patterns, such as DML followed by a `select` without the need to
resort to complex `with` ladders.  Thus, make it so that `output_format`
always controls the output of the trailing statement regardless of how
many preceding statements were in the request (preceding statements
still have their output suppressed).

EdgeQL+HTTP is not supported yet, will be tackled in a followup PR.

### DIFF
--- a/edb/server/compiler/compiler.py
+++ b/edb/server/compiler/compiler.py
@@ -101,8 +101,8 @@ class CompileContext:
     state: dbstate.CompilerConnectionState
     output_format: enums.OutputFormat
     expected_cardinality_one: bool
-    stmt_mode: enums.CompileStatementMode
     protocol_version: Tuple[int, int]
+    skip_first: bool = False
     json_parameters: bool = False
     schema_reflection_mode: bool = False
     implicit_limit: int = 0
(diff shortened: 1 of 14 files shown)


tests/test_server_config.py
```diff
@@ -1255,6 +1255,8 @@ async def test_server_config_idle_connection_02(self):
 
             await asyncio.sleep(1)
 
+            await conn.recv_match(protocol.ParameterStatus)
```
fantix, line 1258: We could drop this after rebasing.

edb/server/compiler/compiler.py
```diff
@@ -1726,55 +1718,55 @@ def _try_compile(
...
-        # That means that the returned QueryUnit has to have the in/out codec
-        # information, correctly inferred "singleton_result" field etc.
-        single_stmt_mode = ctx.stmt_mode is enums.CompileStatementMode.SINGLE
         default_cardinality = enums.Cardinality.NO_RESULT
-
         statements = edgeql.parse_block(source)
         statements_len = len(statements)
 
-        if ctx.stmt_mode is enums.CompileStatementMode.SKIP_FIRST:
+        if ctx.skip_first:
```
fantix, line 1725: We can drop `skip_first` after dropping `binary_v0.pyx` right?

1st1 (approved):

PR #4004, opened 2022-06-30
Use generic execute machinery in `edgeql_http` extension

elprans:
Aside from requiring fewer hacks, this enables support for multi-statement requests.


edb/server/pgcluster.py
```diff
@@ -1240,6 +1241,15 @@ async def _log_subprocess_output(
     return []
 
 
+async def _safe_readline(stream: asyncio.StreamReader) -> bytes:
+    try:
+        line = await stream.readline()
+    except ValueError:
```
1st1, line 1247: Does stream actually recover from this error?

elprans: Yes

1st1 (approved):

msullivan (approved):
I like the refactorings

edb/server/protocol/binary.pyx
```diff
@@ -1341,135 +1071,37 @@ cdef class EdgeConnection:
...
-                    config_ops = [
-                        config.Operation.from_json(r[0][1:])
-                        for r in data]
-            finally:
-                await conn.wait_for_sync()
-        finally:
-            await conn.after_command()
-        return config_ops
+        query_unit = compiled.query_unit_group[0]
+        if query_unit.system_config:
```
fantix, line 1099: This was `if query_unit.config_requires_restart:`, is it a typo here to use `system_config`?

fantix: Fixed in https://github.com/edgedb/edgedb/pull/5250/commits/050316468741e32340df7e2c7c297885d1fa07fd

PR #4008, opened 2022-06-30
Add JSON casts to range types.

vpetrovykh:
Add casts to and from JSON to all our range types. The JSON generated is
structured as an object with the following fields: "lower", "inc_lower",
"upper", "inc_upper".

### DIFF
--- a/edb/pgsql/compiler/output.py
+++ b/edb/pgsql/compiler/output.py
@@ -557,6 +557,12 @@ def serialize_expr_to_json(
             env=env,
         )
 
+    elif irtyputils.is_range(styperef) and not expr.ser_safe:
+        val = pgast.FuncCall(
+            # Use the actual generic helper for converting anyrange to jsonb
+            name=('edgedb', 'range_to_jsonb'),
+            args=[expr], null_safe=True, ser_safe=True)
+
     elif irtyputils.is_collection(styperef) and not expr.ser_safe:
         val = coll_as_json_object(expr, styperef=styperef, env=env)
 
--- a/tests/test_edgeql_expressions.py
+++ b/tests/test_edgeql_expressions.py
@@ -5941,6 +5941,501 @@ aa \
                     )
                 ''')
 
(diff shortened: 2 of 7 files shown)


tests/test_edgeql_expressions.py
```diff
@@ -5941,6 +5941,238 @@ async def test_edgeql_expr_range_31(self):
                     )
                 ''')
 
+    async def test_edgeql_expr_range_32(self):
```
1st1, line 5944: I don't see any error cases. What happens if I pass various kinds of garbage to `select <range<cal::local_date>>to_json($1)`? Think missing some attributes, or an outright wrong kind of JSON (array, number 42, object literal with all wrong keys, etc)

1st1 (changes requested):
Needs tests for error conditions.

tests/test_edgeql_expressions.py
vpetrovykh: I've added a bunch more tests for various wrong JSON inputs.

vpetrovykh:
I've fixed the issue mentioned by @msullivan with json mode (where ranges nested in shapes would not cast to JSON properly) and added a test with a nested shape.

edb/pgsql/compiler/output.py
```diff
@@ -557,6 +557,14 @@ def serialize_expr_to_json(
             env=env,
         )
 
+    elif irtyputils.is_range(styperef) and not expr.ser_safe:
+        cast_name = s_casts.get_cast_fullname_from_names(
```
elprans, line 561: What's the point of getting the cast_name then?

vpetrovykh: Oops, no point. Fixed.

elprans (approved):
LGTM basically, but let's add serialization support to remove the need for `json_only` hack and have proper output tests

1st1 (approved):

PR #4014, opened 2022-06-30
Disallow shapes on computed links

### LINKED ISSUE #3369: InvalidReferenceError when creating an alias
- EdgeDB Version: rc4 (current master)
- OS Version: linux

Steps to Reproduce:

```
edgedb> create type Foo;
OK: CREATE TYPE
edgedb> create type Bar;
OK: CREATE TYPE
edgedb> alter type Foo create link bar := Bar{val := 42};
error: InvalidReferenceError: object type 'default::Bar' does not exist
  ┌─ query:1:35
  │
1 │ alter type Foo create link bar := Bar{val := 42};
  │                                   ^^^^^^^^^^^^^^ error
```

Notice that this is achieved in DDL, so it doesn't look like a typical case of SDL ordering that prevents visibility.

msullivan:
We currently produce nonsensical error messages saying a type does not
exist that clearly does.

Fixes #3369.

### DIFF
--- a/edb/schema/pointers.py
+++ b/edb/schema/pointers.py
@@ -1217,7 +1217,7 @@ class PointerCommandOrFragment(
             and target.is_view(expression.irast.schema)
         ):
             raise errors.UnsupportedFeatureError(
-                f'including a shape on schema-defined computed pointers '
+                f'including a shape on schema-defined computed links '
                 f'is not yet supported',
                 context=self.source_context,
             )
--- a/edb/schema/pointers.py
+++ b/edb/schema/pointers.py
@@ -1211,6 +1211,17 @@ class PointerCommandOrFragment(
                     context=srcctx,
                 )
 
+        if (
+            not isinstance(source, Pointer)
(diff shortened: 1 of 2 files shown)


edb/schema/pointers.py
```diff
@@ -1211,6 +1211,17 @@ def _parse_computable(
                     context=srcctx,
                 )
 
+        if (
+            not isinstance(source, Pointer)
+            and not source.is_view(schema)  # type: ignore
+            and target.is_view(expression.irast.schema)
+        ):
+            raise errors.UnsupportedFeatureError(
+                f'including a shape on schema-defined computed pointers '
```
elprans, line 1220: We tend to avoid using "pointer" in user-facing messages and documentation, so let's spell "properties" or "links" depending on what is being defined.

elprans (approved):

edb/schema/pointers.py
msullivan: oh, and in fact it is always a link

PR #4019, opened 2022-07-02
Allow customizing assertion messages in `assert_exists` and friends

elprans:
The `assert_exists`, `assert_distinct` and `assert_single` now accept an
optional `message` named-only argument that allows changing the default
error text to something more specific.


msullivan (approved):
Looks good, and useful

1st1 (approved):
Nice!

PR #4020, opened 2022-07-03
Replace open-coded range implementations with compiler-assisted generics

elprans:
Ranges are container types, albeit less generic than arrays, so it's
possible to use the same approach to avoid open-coding a bunch of
definitions in std for every concrete range type.

Also, implement a function to create empty ranges in a type-agnostic way.


edb/edgeql/compiler/casts.py
```diff
@@ -540,6 +553,169 @@ def _cast_tuple(
...
+                    ),
+                    type=ql_el_type,
+                ),
+            ],
+            kwargs={
+                "inc_lower": qlast.FunctionCall(
+                    func=('__std__', 'range_is_inclusive_lower'),
+                    args=[source_path],
+                ),
+                "inc_upper": qlast.FunctionCall(
```
msullivan, line 615: We need to handle empty, also?

msullivan (approved):
Seems approximately reasonable, modulo the need to handle empty either in this PR or a follow-up. (The existing code doesn't either.)

On a more general note, it is a shame to have to choose between duplicating a bunch of functions written in SQL and writing general code but as a compiler special case. We'll need to start moving beyond that at some point, especially since eventually our users will demand to write polymorphic functions.

edb/edgeql/compiler/casts.py
vpetrovykh: > We need to handle empty, also?

Presumably we'll have the `std::range_is_empty` function as well as the `empty` argument to the range constructor that can be used here.

elprans:
Pushed an update adding `empty := true` to `range()` and validation.

edb/pgsql/compiler/relgen.py
```diff
@@ -349,29 +351,38 @@ def _get_set_rvar(
...
-            elif str(expr.func_shortname) == 'std::max' and expr.func_sql_expr:
+            elif fname == 'std::max' and expr.func_sql_expr:
                 # Generic std::max
                 rvars = process_set_as_std_min_max(ir_set, stmt, ctx=ctx)
 
+            elif fname == 'std::range' and expr.func_sql_expr:
+                # Generic std::range() constructor
+                rvars = process_set_as_std_range(ir_set, stmt, ctx=ctx)
+
+            elif fname == 'std::range_of_empty' and expr.func_sql_expr:
```
vpetrovykh, line 381: Is this `std::range_of_empty` an artefact of the previous implementation that's no longer relevant? I don't see that function in `std`.

vpetrovykh (approved):
I think it's OK overall, but `range_of_empty` seems to be outdated in this last iteration.

vpetrovykh:
I agree with @msullivan that range casts need to also copy `empty` field, because the bounds of an empty range are `null` and so without the `empty := true` flag it would become an unbounded range of everything.

edb/pgsql/compiler/relgen.py
elprans: Yeah, forgot to clean that up.

elprans:
Yeah, good catch. I added the casting bit and a test for it.

msullivan (approved):

edb/edgeql/compiler/casts.py
```diff
@@ -540,6 +553,166 @@ def _cast_tuple(
...
+                "empty": qlast.TypeCast(
+                    expr=qlast.FunctionCall(
+                        func=('__std__', 'json_get'),
+                        args=[
+                            source_path,
+                            qlast.StringConstant(value='empty'),
+                        ],
+                    ),
+                    type=ql_bool_t
+                ),
```
vpetrovykh, line 709: If I read this correctly this should cause the `empty` argument to be an `{}` if the input JSON doesn't have an explicit `empty` field. Which, in turn, should cause the entire `range` call to result in an `{}` as well, because the named argument `empty` is not optional, it merely has a default and _can be omitted_ entirely.

elprans: These specializations are somewhat different from the normal compilation semantics.  The range constructor is implemented as a non-strict SQL function, so it _will_ get called on `NULL` input to even a non-optional argument (if called directly from a specialization like here), and the implementation uses `empty` in a way that tolerates nulls.

msullivan: Would it be better to pass False in to `json_get` explicitly and avoid any funny business?

msullivan: @vpetrovykh was right that this is sketchy. The non-strictness of the range constructor is not at all limited to being called directly from this kind of specialization:
```
edgedb> select range(<int64>{}, empty:=<optional bool>$0);
Parameter <bool>$0 (Ctrl+D for empty set `{}`): 
[{"inc_lower": false, "inc_upper": false, "lower": null, "upper": null}]
edgedb> select range(<int64>{}, inc_upper:=<optional bool>$0);
Parameter <bool>$0 (Ctrl+D for empty set `{}`): 
edgedb error: InvalidValueError: range constructor flags argument must not be null
```
as one example.
Both examples ought to return an empty set, since they pass an empty set to a non-optional argument, but one returns a value and the other raises an error.

PR #4032, opened 2022-07-07
Fix protov1 for dump/restore

fantix:


edb/server/protocol/binary.pyx
```diff
@@ -1952,7 +1952,10 @@ cdef class EdgeConnection(frontend.FrontendConnection):
 
         msg_buf = WriteBuffer.new_message(b'C')
         msg_buf.write_int16(0)  # no headers
+        msg_buf.write_int64(0)  # capabilities
```
elprans, line 1955: Do we need to worry about proto 0.x compatibility here?

fantix: Ah yes, should be copied to binary_v0

elprans (approved):

PR #4040, opened 2022-07-07
Start work on the changelog.

msullivan:
I think most of the things are on it, but we probably want to say
more.


docs/changelog/2_x.rst
```diff
@@ -0,0 +1,177 @@
...
+  output of ``GROUP`` are free objects that contain the aggregated
+  elements as a computed link:
+
+  .. code-block:: edgeql-repl
+
+	  db> group Card { name } by .element;
+	  {
+		{
+		  key: {element: 'Earth'},
+		  grouping: {'element'},
```
haikyuu, line 38: elements is the plural of element. So that's a bit confusing.
Does it pluralize the `element`? or is it just the api?

msullivan: I should change it to use an example where the key isn't named `element`

msullivan: changed it to the movies db

PR #4041, opened 2022-07-07
Expose instance name as `sys::get_instance_name()`

elprans:
The new `sys::get_instance_name()` function returns the instance name
configured for the server, or `_localdev` for development instances, or
`_unknown` otherwise.

While at it, replace the use of temporary table for server metadata with
custom Postgres GUC variables.

### DIFF
--- a/edb/buildmeta.py
+++ b/edb/buildmeta.py
@@ -44,7 +44,7 @@ from edb.common import verutils
 
 
 # Increment this whenever the database layout or stdlib changes.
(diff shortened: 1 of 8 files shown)


msullivan (approved):

1st1 (approved):
nice!!

1st1:
>  _unknown otherwise.

Maybe it should be an empty string instead?

elprans:
> Maybe it should be an empty string instead?

Not sure that would be more useful.  We can always use something like `_edgedb`, maybe together with the local hostname.

PR #4042, opened 2022-07-07
Remove the `/ui/instance-info` endpoint

elprans:
It is insecure and the necessary data can be queried by connecting to
`__edgedbsys__` with correct authentication and running introspection
queries.

Depends on #4039 and #4041


1st1 (approved):

elprans:
> We can drop the `/admin_binary_http` endpoint as well I think: https://github.com/edgedb/edgedb/blob/kill-instance-info-endpoint/edb/server/protocol/protocol.pyx#L452-L456

OK, dropped.

PR #4046, opened 2022-07-08
Add documentation for ranges.

vpetrovykh:
Update the standard library as well as the datamodel section.


docs/stdlib/range.rst
```diff
@@ -0,0 +1,650 @@
...
+  fields are optional, but the *inclusivity* fields ``inc_lower`` and
+  ``inc_upper`` are *mandatory*. This is to address the fact that whether the
+  range boundaries are included by default can vary based on system or context
+  and being explicit avoids subtle errors. The only exception to this are
+  empty ranges that can have just the ``"empty": true`` field.
+
+Comparing ranges
+^^^^^^^^^^^^^^^^
+
+Ranges can also be compared to each other ...
```
elprans, line 177: Seems like a placeholder section?

vpetrovykh: Oh, right. I ended up doing proper operators blocks instead.

elprans (approved):

PR #4049, opened 2022-07-09
Fix range errors.

vpetrovykh:
It should be an error to unpack an unbounded range.
Correctly translate the error when lower bound is greater than the upper
bound.


edb/server/compiler/errormech.py
```diff
@@ -374,6 +374,16 @@ def _static_interpret_cardinality_violation(_code, err_details):
...
 
 
+@static_interpret_by_code.register(
+    pgerrors.ERROR_DATA_EXCEPTION)
+def _static_interpret_data_exception(_code, err_details):
+    # Problem with range bounds
+    if err_details.message.startswith('range lower bound must be'):
+        return errors.InvalidValueError(err_details.message)
+
+    return errors.InternalServerError(err_details.message)
```
elprans, line 384: It's probably safe to simply translate all data exceptions to `InvalidValue`.

vpetrovykh: Fair point

elprans (approved):

PR #4059, opened 2022-07-11
Add dump tests for new 2.0 types/features

msullivan:
Set up the test so that it will do compatability dump tests as soon as
we add 2.0 dump snapshots.

Also, fix a sertypes bug it found.

### DIFF
--- a/tests/schemas/dump04_default.esdl
+++ b/tests/schemas/dump04_default.esdl
@@ -23,3 +23,31 @@ type Test1 {
         default := <array<tuple<name: str, severity: int16>>>[]
     };
 };
+
+type Test2 {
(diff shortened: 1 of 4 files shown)


tests/schemas/dump04_default.esdl
```diff
@@ -23,3 +23,31 @@ type Test1 {
         default := <array<tuple<name: str, severity: int16>>>[]
     };
 };
+
+type Test2 {
```
1st1, line 27: Can you also add some globals + some schema acl policies and test that they all get restored and can be introspected/verified after restore?

elprans (approved):

PR #4063, opened 2022-07-12
Remove asyncpg dependency

elprans:


fantix (approved):

edb/server/pgcluster.py
```diff
@@ -1005,9 +1023,15 @@ async def _get_reserved_connections(
...
     conn = await rcluster.connect()
     try:
-        user, dbname = await conn.fetchrow(
-            "SELECT current_user, current_database()"
-        )
+        data = json.loads(await conn.sql_fetch_val(
+            b"""
+            SELECT json_build_object(
+                'user', current_user,
+                'db', current_database()
```
fantix, line 1030:
```suggestion
                'dbname', current_database()
```

.github/workflows.src/tests-pg-versions.tpl.yml
```diff
@@ -63,21 +63,30 @@ jobs:
       if: ${{ matrix.single-mode }}
       shell: python
       run: |
-        import asyncio, asyncpg
+        import asyncio
+        import subprocess
+
+        from edb.server.pgcluster import get_pg_bin_dir
+
         async def main():
```
msullivan, line 71: Drop the async, since it's doing a blocking subprocess call and not doing any async stuff

elprans: It's doing `await get_pg_bin_dir()`, so need the async

1st1 (approved):

PR #4066, opened 2022-07-13
Fix stdlib dump compatibility with PostgreSQL 13

elprans:
PostgreSQL 14 emits generated `multirange_type_name` in RANGE type
definitions, which is incompatible with PostgreSQL 13.  Elide them
explicitly to avoid the compatibility issue.


msullivan (approved):
Sure, why not

Issue #1080, opened 2020-01-11
Nested INSERT bug


vpetrovykh:
```
db> CREATE TYPE Foo {CREATE LINK bar -> Object};
CREATE
db> INSERT Foo {bar := (INSERT Foo)};
QueryError: invalid reference to default::Foo: self-referencing INSERTs are not allowed
Hint: Use DETACHED if you meant to refer to an uncorrelated default::Foo set
### INSERT Foo {bar := (INSERT Foo)};
###
```
This looks like a bug to me. I was under the impression that the thing right after `INSERT` is supposed to be a type name, not an arbitrary expression, as such there cannot be a problem of self-reference. If memory serves me, we couldn't actually guarantee that self-references in an `INSERT` would work (say for use in a subquery selecting some link based on other yet-to-be-inserted data that appears syntactically in the `INSERT` statement).

As far as I recall this is illegal:
```
INSERT Bar {
    a := 'data',
    b := .a
};
```
Am I misunderstanding `INSERT`?

Just to clarify, the intent is to simple create 2 `Foo` objects where one is linked to the other. This is *not* intended or expected to be some sort of self-link.

elprans:
We prohibited this sort of self-reference for consistency with `UPDATE Foo`, where `Foo` *is* an expression.

elprans:
See #539 for context of this.

1st1:
TBH, I still don't understand why `INSERT Foo {bar := (INSERT Foo)};` doesn't work.   This looks like something obvious that doesn't work for some non-obvious reasons -- people will struggle with this.

> UPDATE Foo, where Foo is an expression.

Can "Foo" be an expression?

elprans:
> Can "Foo" be an expression?

I don't see why not.

```
WITH Foo := (SELECT User FILTER ...)
UPDATE Foo SET { ... };
```

`DELETE` is similar.

1st1:
DELETE requires parens, no? I.e. `DELETE (SELECT ...)`.

`INSERT Foo {bar := (INSERT Foo)};` seems more useful to me than `WITH Foo := (SELECT User FILTER ...) UPDATE Foo SET { ... };`.

vpetrovykh:
OK, so I understand why we might want to forbid this:
```
INSERT Foo {bar := (SELECT Foo LIMIT 1)};
```
This is genuinely ambiguous in terms of what the person might be trying to express:
1) The intent was to create a self-looping single object `Foo`.
[...]

PR #4070, opened 2022-07-13
Make sure that INSERT path ids in ELSE clauses differ from enclosing

msullivan:
That is, if we have
```
  INSERT Person {name := "Madz"}
  UNLESS CONFLICT ON (.name)
  ELSE (INSERT Person {name := "Maddy"})
```

make sure that the inner `Person` has a distinct pathid (by
namespacing it) from the outer `Person`. They don't correlate despite
the outer one being visible to the inner.

This doesn't fix any current bugs, but that's mostly a matter of
luck. It fixes an issue in an experimental PR I'm working on.

### DIFF
--- a/edb/edgeql/compiler/conflicts.py
+++ b/edb/edgeql/compiler/conflicts.py
@@ -42,6 +42,7 @@ from . import astutils
 from . import context
 from . import dispatch
 from . import inference
+from . import pathctx
 from . import setgen
 from . import typegen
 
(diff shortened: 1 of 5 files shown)


elprans:
Hmm, this kind-of brings back the [nested `INSERT` discussion](https://github.com/edgedb/edgedb/issues/1080#issuecomment-905856270)

msullivan:
Hm, I wouldn't hate always doing this, either

elprans:
> Hm, I wouldn't hate always doing this, either

Per discussion, we should prohibit nested `INSERT` in `ELSE` like we do it in nested shapes (because `ELSE` is a nested scope).

PR #4073, opened 2022-07-14
Add ACL and globals docs.

vpetrovykh:
Add documentaion for globals and for ACL, which indirectly depends
on globals.


docs/reference/sdl/acl.rst
```diff
@@ -0,0 +1,155 @@
...
+:eql:synopsis:`update read`
+    Apply the policy to all objects selected for an update. If an object does
+    not allow access under this kind of policy, it is not visible cannot be
+    updated.
+
+    Note that any object that cannot be selected, cannot be modified either.
+
+:eql:synopsis:`update write`
+    Apply the policy to all objects at the end of an update. If an updated
+    object violate this policy, an error is produced instead.
```
colinhacks, line 127:
```suggestion
    object violates this policy, an error is produced instead.
```

docs/reference/sdl/acl.rst
```diff
@@ -0,0 +1,155 @@
...
+Any sub-type :ref:`extending <ref_datamodel_inheritance>` a base type also
+inherits all the access policies of the base type.
+
+The access policy declaration options are as follows:
+
+:eql:synopsis:`<name>`
+    The name of the access policy.
+
+:eql:synopsis:`when (<condition>)`
+    Specifies which objects should this policy apply to. The
```
colinhacks, line 82:
```suggestion
    Specifies which objects this policy applies to. The
```

docs/reference/ddl/acl.rst
```diff
@@ -0,0 +1,142 @@
...
+    alter type <TypeName> "{"
+      [ ... ]
+      drop access policy <name> ;
+    "}"
+
+Description
+-----------
+
+The combination :eql:synopsis:`alter type ... drop access policy`
+removeds the specified access policy from a given object type.
```
colinhacks, line 135:
```suggestion
removes the specified access policy from a given object type.
```

docs/reference/sdl/globals.rst
```diff
@@ -0,0 +1,145 @@
...
+            select User filter .id = global current_user_id
+        );
+
+
+The valid SDL sub-declarations are listed below:
+
+:eql:synopsis:`default := <expression>`
+    Specifies the default value for the global variable as an EdgeQL
+    expression. The default value is used by the session if the value was not
+    explicitly specified or by the :ref:`reset
```
colinhacks, line 133:
```suggestion
    explicitly specified by the client or was reset with the :ref:`reset
```

docs/reference/sdl/index.rst
```diff
@@ -82,3 +82,5 @@ to the previous migration:
     functions
```
colinhacks, line 82: For consistency with the non-reference docs, can we use this ordering?

- annotations
- globals
- acl
- functions
- extensions

docs/reference/ddl/acl.rst
```diff
@@ -0,0 +1,142 @@
+.. _ref_eql_ddl_acl:
```
colinhacks, line 1: How do you feel about switching to use `ols` instead of `acl` in all the refs and filenames? I think that's a little more consistent with the messaging, since we seem to have converged on referring to this as "object-level security".

colinhacks (approved):

docs/reference/ddl/acl.rst
1st1: I think we should keep using ACL as that's an established term that people will be searching for. OLS might be more accurate, but no one knows it and the benefit of switching to using it is tiny.

vpetrovykh: Other than in the filenames and git commit messages I don't think I use the term ACL. The docs talk about "access policies" implementing "object-level security".

docs/reference/sdl/acl.rst
```diff
@@ -0,0 +1,155 @@
...
+    Note that any object that cannot be selected, cannot be modified either.
+
+:eql:synopsis:`using <expr>`
+    Specifies what the policy is with respect to a given eligible (based on
+    :eql:synopsis:`when` clause) object. The :eql:synopsis:`<expr>` has to be
+    a :eql:type:`bool` expression. The specific meaning of this value also
+    depends on whether this policy flavor is :eql:synopsis:`allow` or
+    :eql:synopsis:`deny`.
+
+    When omitted, it is assumed that this policy applies to all eligible
```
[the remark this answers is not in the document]
vpetrovykh: Ooops! I confirm that I also get this error.

PR #4078, opened 2022-07-15
Unbreak simple query in protocol v0

elprans:


1st1 (approved):

fantix (approved):
I'll write some test for v0 support.

PR #4087, opened 2022-07-15
Remove redundant phrase in range docs

fmoor:


fmoor:
Just realized I misunderstood this.

fmoor:
@vpetrovykh maybe this could be improved somehow? To exclude something and omit something are very similar ideas. Perhaps unbounded or infinite or something is more clear? :thinking:

PR #4098, opened 2022-07-18
add support for conditional top-level DDL statements

### LINKED ISSUE #1385: Add support for existence conditionals to all top-level DDL commands.

nsidnev:
This adds support for conditional statements for most top-level DDL statements.

I skipped a few statements that look like internals or ones that IMO don't need this:
1. `CREATE/DROP DATABASE` - PostgreSQL doesn't allow you to create databases conditionally + for such "external" objects EdgeDB has its custom logic. But I can try to add this if you want to.
2. `CREATE/DROP EXTENSION PACKAGE` - it seems that these are statements for internal use. But if needed, I can add support for that there too.
3. `CREATE PSEUDO TYPE` - it doesn't seem that the user should be able to do that anyway.
4. `CREATE/ALTER/DROP OPERATOR` - also looks like internal statements.
5. `CREATE/ALTER/DROP CAST` - the same as above.
6. `CREATE/ALTER/DROP MIGRATION` - it seems to me that the user should not be able to manipulate migrations in conditional way.

Also, if you want, I can break this PR up into several separate PRs, though most of additions here are tests and grammar changes.

Closes #1385

### DIFF
--- a/edb/schema/delta.py
+++ b/edb/schema/delta.py
@@ -3443,8 +3443,7 @@ class AlterObject(AlterObjectOrFragment[so.Object_T], Generic[so.Object_T]):
         schema: s_schema.Schema,
         context: CommandContext,
     ) -> s_schema.Schema:
-
-        if not context.canonical and self.if_exists:
+        if self.if_exists:
             scls = self.get_object(schema, context, default=None)
             if scls is None:
                 context.current().op.discard(self)
--- a/edb/schema/delta.py
+++ b/edb/schema/delta.py
@@ -3405,7 +3405,8 @@ class AlterObject(AlterObjectOrFragment[so.Object_T], Generic[so.Object_T]):
         cmd = super()._cmd_tree_from_ast(schema, astnode, context)
         assert isinstance(cmd, AlterObject)
 
-        # idk how, but somehow there could be an instance of CreateConcreteLink here
+        # idk how, but somehow there could be an instance
+        # of CreateConcreteLink here
         if isinstance(astnode, qlast.AlterObject):
             cmd.if_exists = astnode.alter_if_exists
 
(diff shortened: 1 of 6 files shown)


nsidnev:
@msullivan friendly ping :)

edb/schema/delta.py
```diff
@@ -3405,6 +3405,11 @@ def _cmd_tree_from_ast(
         cmd = super()._cmd_tree_from_ast(schema, astnode, context)
         assert isinstance(cmd, AlterObject)
 
+        # idk how, but somehow there could be an instance
```
msullivan, line 3408: overloaded functions declared in SDL have `CreateObject` asts (with flags indicating that they can be alters too)

msullivan (approved):
Looks great. Clean up one comment and this is solid.

elprans (changes requested):
The implementation looks solid, but before we merge this I'd like us to reflect a little bit on whether we want `CREATE` to be conditional, because it creates ambiguity as to what state the final object will be: the existing structure or the new structure.

elprans:
> The implementation looks solid, but before we merge this I'd like us to reflect a little bit on whether we want `CREATE` to be conditional, because it creates ambiguity as to what state the final object will be: the existing structure or the new structure.

To further clarify, I think `CREATE .. IF NOT EXISTS` might be better expressed as `DROP .. IF EXISTS; CREATE ...`

1st1:
> To further clarify, I think CREATE .. IF NOT EXISTS might be better expressed as DROP .. IF EXISTS; CREATE ...

Oh, that's a very good idea.

nsidnev:
> To further clarify, I think CREATE .. IF NOT EXISTS might be better expressed as DROP .. IF EXISTS; CREATE ...

But we already have support for this for modules and roles. And although I don't mind the `drop -> create` approach, I honestly don't see anything confusing here. If I use conditional creation, I understand that if the entity already exists, it won't be overwritten by this statement, and nothing will be done instead. 

And there is also a question how to be in the case when there is already some data for the type from conditional creation, but the user doesn't know if the type itself is actually in the schema. And just want to be sure that before any further change user will have the necessary type in the scheme.

PR #4101, opened 2022-07-18
Make ViewRPtr a dataclass

msullivan:


edb/edgeql/compiler/context.py
```diff
@@ -69,29 +69,18 @@ class ContextSwitchMode(enum.Enum):
     DETACHED = enum.auto()
 
 
+@dataclasses.dataclass
```
elprans, line 72:
```suggestion
@dataclasses.dataclass(kw_only=True)
```

elprans (approved):

edb/edgeql/compiler/context.py
msullivan: It has a `_: dataclasses.KW_ONLY` in it, to preserve the existing behavior of source being positional

But there's no real reason why that arg should be positional so I changed it to make everything kw_only

PR #4114, opened 2022-07-21
Fix properties of type array<uuid> in migrations

### LINKED ISSUE #4107: Properties or type `array<uuid>` are broken
```
edgedb> start migration to {
....... module default {
....... type Foo {
....... property myArray -> array<uuid>;
....... }; }; };
OK: START MIGRATION
edgedb error: SchemaError: Array ('d2a2fc36-29e2-20f2-9861-9f2633371c73') is already present in the schema <FlatSchema gen:1 at 0x7fd2da28fee0>
```

msullivan:
The issue is that the type array<uuid> was in an inconsistent
state. The shadow reflection schema uses array<uuid>, so it is always
in the reflected schema, but since it is created by the reflection
schema code and does not appear in the stdlib, it didn't appear in the
stdschema.

Fix this by adding any collection types that appear in the reflschema
into the stdschema.

Also to avoid related inconsistencies, skip loading objects that
appear in the stdschema when processing the json from the reflection
query. The query filters out qualified objects from the std schema but
not "global" ones, so collections and some other things were being
duplicated between _base_schema and _top_schema.

Fixes #4107.

### DIFF
--- a/edb/schema/reflection/reader.py
+++ b/edb/schema/reflection/reader.py
@@ -100,6 +100,9 @@ def parse_into(
         name = s_name.name_from_string(entry['name__internal'])
         layout = schema_class_layout[mcls]
 
+        if base_schema.has_object(objid) and str(name) != '__schema_version__':
+            continue
+
         if isinstance(obj, s_obj.QualifiedObject):
             name_to_id[name] = objid
         else:
--- a/edb/schema/reflection/reader.py
+++ b/edb/schema/reflection/reader.py
@@ -36,6 +36,7 @@ from edb.schema import name as s_name
 from edb.schema import objects as s_obj
 from edb.schema import operators as s_oper
 from edb.schema import schema as s_schema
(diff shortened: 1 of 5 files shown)


edb/schema/reflection/reader.py
```diff
@@ -100,6 +100,9 @@ def parse_into(
         name = s_name.name_from_string(entry['name__internal'])
         layout = schema_class_layout[mcls]
 
+        if base_schema.has_object(objid) and str(name) != '__schema_version__':
```
elprans, line 103: Why `__schema_version` specifically?  Is there a more general rule?

msullivan: I made it handle the Version types more generally. I don't know if there's something more general than that

msullivan: But it is because `__schema_version__` is special and get updated despite also being in the base schema

elprans (approved):

PR #4118, opened 2022-07-21
add `std::json_set` function

### LINKED ISSUE #3826: Update parts of JSON properties with a json_set function
Let's say initially `myMovie.json = {"a": 10, "b": 20}`

Instead of `update Movie set {myJson := to_json('{"a": 11, "b": 20}')};`, make it possible to do something like: `update Movie set {myJson := json_set(.myJson, 'a', .myJson['a'] + 1)}`.

It could just be syntactic sugar for cloning the whole JSON object and updating a deep field.

The syntax could be similar to `json_get` where the first argument specifies the JSON property, then the middle arguments are a path to a location within the JSON object to update, and the last argument is the value to set.

What if the path is invalid, e.g. `json_set(.myJson, 'a', 'x', 'y', 11)`? Maybe create singleton objects along the way to make it valid like `{"a": {"x": {"y": 11}}}`.

If there's an array along the path, could use numbers to specify indexes, and error if the path part isn't a number.

The command can potentially have a merge flag, so if `myMovie.myJson` initially equals `{"a": {"x": 5, "y": 6}}`, then `json_set(.myJson, 'a', to_json('{"y": 7, "z": 8}'))` could return `{"a": {"x": 5, "y": 7, "z": 8}}`.

nsidnev:
Closes #3826


nsidnev:
It ended with an SSL error in `test_server_config_db_config`, which should not be affected by this change. Maybe it is flaky one and can be fixed by retrying the job?

elprans:
> It ended with an SSL error in `test_server_config_db_config`, which should not be affected by this change. Maybe it is flaky one and can be fixed by retrying the job?

That's #4119

msullivan (commented):
This looks great. Implementation seems solid.

My one concern is more about the specification:
I think `create_if_missing` should probably be true by default, like it is in postgres.

Adding an element to a top level existing object seems like a pretty bread-and-butter case that should be pretty straightforward.

@elprans, thoughts?

elprans:
> I think `create_if_missing` should probably be true by default, like it is in postgres.

Yes

docs/stdlib/json.rst
```diff
@@ -353,6 +356,91 @@ possible to cast a JSON value directly into a :eql:type:`tuple`.
...
+.. eql:function:: std::json_set( \
+                    target: json, \
+                    path: array<str>, \
+                    new_value: optional json, \
+                    named only create_if_missing: bool = true, \
+                    named only empty_treatment: JsonSetEmptyTreatment = \
+                      JsonSetEmptyTreatment.ReturnEmpty) \
+                  -> optional json
+
+    Returns an updated JSON target with a new value.
```
1st1, line 368: Need to add that this is a new v2 function.

1st1 (changes requested):
Hm, the asymmetry between `json_get` and `json_set` worries me: in `json_get` the path arg is variadic, but here it's an array. I get why it's designed this way, but I don't like this.

1st1:
I think we should consider an alternative signature, where the new value is either the second positional argument, followed by a variadic path, or when new value is a named only arg.

elprans:
>  followed by a variadic path

Variadic arguments are actually a pain to use currently because we don't have syntax to unpack an array into a variadic.

1st1:
> Variadic arguments are actually a pain to use currently because we don't have syntax to unpack an array into a variadic.

Agree, but we can add syntax to unpack in the future. That said, I think adding a new function that has a closely related counterpart already in the stdlib designed in a different way is a bad idea.

elprans:
The only reasonable alternative is making the new value `named only`, because making it precede the variadic path could be utterly confusing to a reader when both arguments are variables.

1st1:
> The only reasonable alternative is making the new value named only

I'm OK with that. @nsidnev do you want to change the signature?

nsidnev:
> do you want to change the signature?

Yeah, named only argument looks reasonable. I will update this PR soon.

nsidnev:
Failed with `TransactionSerializationError` in `test_server_proto_ddlprop_01`, looks like a flaky one, which can be fixed by restarting the job

docs/stdlib/json.rst
```diff
@@ -353,6 +356,105 @@ possible to cast a JSON value directly into a :eql:type:`tuple`.
...
+        ...   'a',
+        ...   value := <json>{},
+        ...   empty_treatment := JsonSetEmptyTreatment.ReturnTarget,
+        ... );
+        {'{"a": 10, "b": 20}'}
+        db> select json_set(
+        ...   to_json('{"a": 10, "b": 20}'),
+        ...   'a',
+        ...   value := <json>{},
+        ...   empty_treatment := JsonSetEmptyTreatment.RaiseException,
```
1st1, line 437: @elprans How about we rename `JsonSetEmptyTreatment` to `JsonEmpty`? The former is offensively long to type.

```suggestion
        ...   empty_treatment := JsonEmpty.RaiseException,
```

1st1: On that note, `JsonEmpty.Error` is better than `JsonEmpty.RaiseException`

docs/stdlib/json.rst
```diff
@@ -353,6 +356,105 @@ possible to cast a JSON value directly into a :eql:type:`tuple`.
...
+        ... );
+        InvalidValueError: invalid empty JSON value
+        db> select json_set(
+        ...   to_json('{"a": 10, "b": 20}'),
+        ...   'a',
+        ...   value := <json>{},
+        ...   empty_treatment := JsonSetEmptyTreatment.UseJsonNull,
```
1st1, line 444:
```suggestion
        ...   empty_treatment := JsonEmpty.Null,
```

docs/stdlib/json.rst
msullivan: I'm +1 on these changes, I think

docs/stdlib/json.rst
msullivan: I think we should keep the `Use`, since every other value is a verb phrase. ("Error" unquestionably can be a verb in Software Engineer English, even if Merriam-Webster thinks it is only a noun.)

We could drop the `Json`, and make it `UseNull`. The SQL names the initial design borrowed from obviously needed to distinguish between SQL and json nulls, which we do not. (Though it could still be clearer if we did?)

nsidnev: > I think we should keep the Use, since every other value is a verb phrase

+1 to that. Or alternatively, I could remove the verbs from the other parts, but I like `ReturnEmpty` and `ReturnTarget`, so I'd prefer to keep them if possible.

msullivan (approved):
I am happy with this!

1st1 (approved):

edb/buildmeta.py
```diff
@@ -44,7 +44,7 @@
 
 
 # Increment this whenever the database layout or stdlib changes.
-EDGEDB_CATALOG_VERSION = 2022_07_22_00_00
+EDGEDB_CATALOG_VERSION = 2022_07_25_00_00
```
1st1, line 47:
```suggestion
EDGEDB_CATALOG_VERSION = 2022_07_26_00_00
```

1st1:
Thanks for the hard work, Nik!

PR #4123, opened 2022-07-22
Fix diamond pattern delegated constraints

### LINKED ISSUE #4105: Migration fails with sub-nested hierarchies
- EdgeDB Version: 2.0-rc.3+f035802, 3.0-dev.6761+local

Maybe related: #3289.

Steps to Reproduce:

```
edgedb> start migration to {}; populate migration; commit migration;
OK: START MIGRATION
OK: POPULATE MIGRATION
Error: InternalServerError: column old.bcab4e74-0793-11ed-81d4-a5e4b75261ac does not exist
```

Schema:
```
abstract type R {
    }
}

type S extending R {
    required property s -> str;
}


type T extending R {
    required property t -> str;
}


type V extending S, T;
```

msullivan:
If we have an inheritence tree like V <: S, T <: R, with R
declaring a delegated constraint, we need to check both S and T
in the constraint trigger for V.

This requires a decent amount of rearranging the guts of trigger
generation.

Also fix dropping a property with such a constraint, which was the
originally filed issue here.

Fixes #4105.

### DIFF
--- a/edb/buildmeta.py
+++ b/edb/buildmeta.py
@@ -44,7 +44,7 @@ from edb.common import verutils
 
 
 # Increment this whenever the database layout or stdlib changes.
-EDGEDB_CATALOG_VERSION = 2022_07_22_00_00
+EDGEDB_CATALOG_VERSION = 2022_07_25_00_00
 EDGEDB_MAJOR_VERSION = 3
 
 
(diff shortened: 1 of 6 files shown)


elprans (approved):
LGTM. This probably also requires a catver bump.

PR #4126, opened 2022-07-22
Gate explicit id setting behind a configuration flag

msullivan:
This deals with some potential security issues in some use cases
if it is always enabled.

### DIFF
--- a/edb/edgeql/compiler/viewgen.py
+++ b/edb/edgeql/compiler/viewgen.py
@@ -1106,7 +1106,10 @@ def _normalize_view_ptr_expr(
         else:
             msg = f'cannot assign to {ptrcls_sn.name}'
         if id_access and not ctx.env.options.allow_user_specified_id:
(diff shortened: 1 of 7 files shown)


edb/edgeql/compiler/viewgen.py
```diff
@@ -1078,26 +1078,39 @@ def _normalize_view_ptr_expr(
...
     ):
-        ptrcls_sn = ptrcls.get_shortname(ctx.env.schema)
         if is_polymorphic:
             msg = (f'cannot access {ptrcls_sn.name} on a polymorphic '
                    f'shape element')
         else:
             msg = f'cannot assign to {ptrcls_sn.name}'
-        raise errors.QueryError(msg, context=shape_el.context)
+        if id_access and not ctx.env.options.allow_user_specified_id:
+            hint = 'config setting allow_user_specified_id must be enabled'
```
elprans, line 1109:
```suggestion
            hint = 'consider enabling the "allow_user_specified_id" configuration parameter to allow setting custom object ids'
```

elprans (approved):

PR #4127, opened 2022-07-22
Eagerly introspect the extensions for each database

msullivan:
This is necessary because otherwise we won't accept http/graphql
connections until a connection is made from the native protocol.

### DIFF
--- a/edb/server/server.py
+++ b/edb/server/server.py
@@ -668,6 +668,23 @@ class Server(ha_base.ClusterProtocol):
             schema_class_layout=self._schema_class_layout,
         )
 
+    async def _acquire_intro_pgcon(self, dbname):
+        try:
+            conn = await self.acquire_pgcon(dbname)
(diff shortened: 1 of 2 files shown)


edb/server/server.py
```diff
@@ -754,23 +760,56 @@ async def introspect_db_config(self, conn):
...
+
         try:
-            dbs_query = self.get_sys_query('listdbs')
-            json_data = await syscon.sql_fetch_val(dbs_query)
-            dbnames = json.loads(json_data)
-        finally:
-            self._release_sys_pgcon()
+            extension_names_json = await conn.sql_fetch_val(
+                b'''
+                    SELECT json_agg(name) FROM edgedb."_SchemaExtension";
```
elprans, line 778: I'd really prefer this to be a properly compiled EdgeQL "sys_query" like the other introspection queries here.

elprans (approved):

PR #4129, opened 2022-07-22
Remove (mostly) the WHEN clause from access policies

msullivan:
I decided to go halfway with the compatibility break with old DDL: we
accept a WHEN in CREATE, and merge it with the expr, but we *don't*
accept modifications of WHEN in ALTER, which would require
persistently keeping track of whether a WHEN was present and handling
it in a bunch of places.

This should cover most existing DDL, I suspect.

Maybe it's only worth trying to deal with the compatability at all if
we do it fully, though.

### DIFF
--- a/edb/schema/policies.py
+++ b/edb/schema/policies.py
@@ -51,14 +51,6 @@ class AccessPolicy(
     data_safe=True,
 ):
 
-    condition = so.SchemaField(
-        s_expr.Expression,
-        default=None,
-        coerce=True,
-        compcoef=0.909,
-        special_ddl_syntax=True,
-    )
-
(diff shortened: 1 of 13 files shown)


edb/schema/policies.py
```diff
@@ -51,13 +51,12 @@ class AccessPolicy(
...
 ):
 
-    condition = so.SchemaField(
-        s_expr.Expression,
-        default=None,
-        coerce=True,
-        compcoef=0.909,
-        special_ddl_syntax=True,
-    )
+    # has_condition = so.SchemaField(
```
elprans, line 54: remove?

PR #4134, opened 2022-07-25
fix link anchor in docs

theMosaad:
old ref: https://www.edgedb.com/docs/edgeql/paths#ref-eql-paths (goes to start of the page)
new ref: https://www.edgedb.com/docs/edgeql/paths#ref-eql-paths-backlinks (goes to the appropriate section)


colinhacks (approved):
Thanks, good catch!

PR #4137, opened 2022-07-25
Apply listen_addresses correctly with wildcard

### LINKED ISSUE #3971: InternalServerError changing listen_addresses on bare metal install
I followed the instructions in the [bare metal deployment guide](https://www.edgedb.com/docs/guides/deployment/bare_metal) and discovered that I could not change the `listen_addresses` without restarting the server.

- EdgeDB Version: EdgeDB 1.4+9d299d7 (repl 1.1.2+f832277)
- OS Version: ubuntu 22.04

Steps to Reproduce:
```bash
$ sudo edgedb --tls-security=insecure --admin -P 5656
EdgeDB 1.4+9d299d7 (repl 1.1.2+f832277)
edgedb> alter role edgedb set password := 'super1secret';
OK: ALTER ROLE

edgedb> configure instance set listen_addresses := {'0.0.0.0'};
Error: InternalServerError: could not create any listen sockets

edgedb> \q
$ sudo systemctl restart edgedb-server-1.service
$ sudo edgedb --tls-security=insecure -P 5656 --password
Password for 'edgedb':
EdgeDB 1.4+9d299d7 (repl 1.1.2+f832277)
edgedb> configure instance set listen_addresses := {'0.0.0.0'};
OK: CONFIGURE INSTANCE
edgedb> \q
```

fantix:
* Shutdown conflicting servers first if old/new addresses has wildcard
* Don't fail the startup if only admin server can be started (moved to #4138)
* Fix ISE with IPv6 addresses
* Raise a proper error if the newly-configured listen_* doesn't work
  * For now it just says "Server updated its config but cannot serve on requested address/port, please see server log for more information.", but we can show the proper socket errors in the future.

Fixes #3971, see also: #2696

### DIFF
--- a/edb/server/server.py
+++ b/edb/server/server.py
@@ -896,8 +896,11 @@ class Server(ha_base.ClusterProtocol):
     async def _restart_servers_new_addr(self, nethosts, netport):
         if not netport:
             raise RuntimeError('cannot restart without network port specified')
-        nethosts = await _resolve_interfaces(nethosts)
+        nethosts, has_ipv4_wc, has_ipv6_wc = await _resolve_interfaces(
+            nethosts
+        )
         servers_to_stop = []
+        servers_to_stop_early = []
         servers = {}
         if self._listen_port == netport:
             hosts_to_start = [
@@ -906,7 +909,25 @@ class Server(ha_base.ClusterProtocol):
             for host, srv in self._servers.items():
                 if host == ADMIN_PLACEHOLDER or host in nethosts:
                     servers[host] = srv
+                elif host in ['::', '0.0.0.0']:
+                    servers_to_stop_early.append(srv)
                 else:
+                    if has_ipv4_wc:
+                        try:
+                            ipaddress.IPv4Address(host)
+                        except ValueError:
+                            pass
+                        else:
+                            servers_to_stop_early.append(srv)
+                            continue
+                    if has_ipv6_wc:
+                        try:
+                            ipaddress.IPv6Address(host)
+                        except ValueError:
+                            pass
+                        else:
+                            servers_to_stop_early.append(srv)
+                            continue
                     servers_to_stop.append(srv)
             admin = False
         else:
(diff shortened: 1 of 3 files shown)


edb/server/server.py
```diff
@@ -1466,7 +1499,7 @@ async def _start_servers(
                 if fut.result() is not None
             })
 
-        if not servers:
+        if not servers and not (admin and port):
             raise StartupError("could not create any listen sockets")
```
1st1, line 1503: Please add a short comment here.

edb/server/server.py
```diff
@@ -906,33 +909,63 @@ async def _restart_servers_new_addr(self, nethosts, netport):
             for host, srv in self._servers.items():
                 if host == ADMIN_PLACEHOLDER or host in nethosts:
                     servers[host] = srv
+                elif host in ['::', '0.0.0.0']:
+                    servers_to_stop_early.append(srv)
```
1st1, line 913: Should `hosts_to_start` be synced with `servers_to_stop_early`?

edb/server/server.py
1st1: And can we have a separate PR specifically for this change?

edb/server/server.py
```diff
@@ -906,33 +909,63 @@ async def _restart_servers_new_addr(self, nethosts, netport):
...
+
         if hosts_to_start:
-            new_servers, *_ = await self._start_servers(
-                hosts_to_start,
-                netport,
-                admin=admin,
-            )
-            servers.update(new_servers)
+            try:
+                new_servers, *_ = await self._start_servers(
```
elprans, line 943: Ideally we shouldn't fail all servers because one of them failed to start.  This can be fixed in a follow up PR, though.

edb/server/server.py
fantix: The first `if` is already checking `host in nethosts`, which is the same condition populating `hosts_to_start` - not sure if that answers the question.

edb/server/server.py
fantix: I think you already did that in #3210?

edb/server/server.py
fantix: Created #4138.

edb/server/server.py
elprans: Ah, right :-)

edb/server/server.py
1st1: Ah, I reread the code, it's all fine.

1st1 (approved):

PR #4138, opened 2022-07-25
Still start when only admin server is functional

fantix:
This prevents bricking the server with invalid listen_addresses or
listen_port followed by a restart.

### DIFF
--- a/edb/server/server.py
+++ b/edb/server/server.py
@@ -1466,7 +1466,9 @@ class Server(ha_base.ClusterProtocol):
                 if fut.result() is not None
             })
 
-        if not servers:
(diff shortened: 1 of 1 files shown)


edb/server/server.py
```diff
@@ -1466,7 +1466,9 @@ async def _start_servers(
                 if fut.result() is not None
             })
 
-        if not servers:
+        # Fail if none of the servers can be started, except when the admin
+        # server on a UNIX domain socket will be started.
+        if not servers and not (admin and port):
```
elprans, line 1471:
```suggestion
        if not servers and (not admin or port == 0):
```

elprans (approved):

PR #4144, opened 2022-07-26
Add simple Rust page

colinhacks:
Interim solution, we can go a proper conversion step when we have the resources. In the meantime I think it's important that we have a link in the sidebar. 

Should probably hold off on merging until `v0.3` lands, cc @tailhook


tailhook:
No need to hold off, since it refers to `latest` anyway. I'm going to release new version today.

1st1 (approved):

PR #4148, opened 2022-07-26
Update guides to use 2.x

fmoor:
Docker image references that were pinned are changed to use latest.

### DIFF
--- a/docs/guides/deployment/digitalocean.rst
+++ b/docs/guides/deployment/digitalocean.rst
@@ -221,5 +221,5 @@ your droplet and run the following.
 
 .. code-block:: bash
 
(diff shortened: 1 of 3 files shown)


docs/guides/deployment/digitalocean.rst
```diff
@@ -221,5 +221,5 @@ your droplet and run the following.
 
 .. code-block:: bash
 
-   $ apt-get update && apt-get install --only-upgrade edgedb-1
+   $ apt-get update && apt-get install --only-upgrade edgedb-server-2
```
fmoor, line 224: `edgedb-1` is not installed in the current digital ocean image only `edgedb-server-` and `edgedb-cli`

elprans (approved):

Issue #4142, opened 2022-07-25
Explicit references to subject type in schemas break in presence of inheritance


msullivan:
This happens with basically every place it can happen, but one example:
```
edgedb> create type A { create property z -> str; create property zed := A.z; };
OK: CREATE TYPE
edgedb> create type B extending A;
OK: CREATE TYPE
edgedb> insert A { z := "1" }; insert A { z := "2" }; insert B { z := "3" };
[{"id": "bb03847c-0a0b-11ed-b000-bf55cc7ccfd7"}]
[{"id": "bb03847d-0a0b-11ed-b000-87ad3b679fe5"}]
[{"id": "bb03847e-0a0b-11ed-b000-8b28fd0e4fe7"}]
edgedb> select B { multi oops := .zed };
[{"oops": ["1", "2", "3"]}]
edgedb> select B { zed };
edgedb error: InternalServerError: more than one row returned by a subquery used as an expression
```

Here, we define a computed property on `A` that directly references `A` -- this is obviously interpreted as referring to the source A.

But when we select it on `B`, it is intepreted as referring to `A` still!

msullivan:
#4826 fixed this for computeds, and https://github.com/edgedb/edgedb/pull/4146 fixed it for constraints. (I think; there may be some fiddliness there.) I think we probably could just fix this issue everywhere it happens.

We found this issue *right* before our 2.0 release, and I remember the consensus was that this wasn't plausible to fix and we should eventually transition towards having these schema expressions be in a "detached" context, where the type name of the enclosing type would refer to the whole type, not the current subject/source.

The plan then was that we would need to release some sort of tool to do the schema migration and rewrite migration history to support it.

To match the planned new semantics, access policies were defined with a `DETACHED`-style definition.

This leaves us in an annoying position:
1. Most schema expressions are *not* detached, but many of them are still broken. (But should be fixable)
2. But access policies *are* detached

Our options then are something like:
1. Move forward with the plan to make everything detached. This will involve building a schema fixing tool
2. Fix the rest of the bugs, call it a day. Leave access policies detached
3. Fix the rest of the bugs, and plan to eventually move access policies to *not* be detached

PR #4149, opened 2022-07-27
Fix a number of related access policy semantics issues

msullivan:
1. Fix semantics of which policies apply *inside* of access policies

While evaluating a rule, the subject type *of the rule* and
*all of its descendants* may be accessed without restriction.
I think this is the only coherent thing.

Implementing this requires being substantially more intentional in the
compiler about when we want to use a type rewrite: previously we would
essentially always use a type rewrite, except when in the process of
compiling the type rewrite CTE itself, and that decision was made on
the pgsql side. There existed an `ignore_rewrites` mechanism, but it
was so-far only used to compile a conflict CTE for BaseObject when
doing an INSERT with explicit id.

We expand the use of `ignore_rewrites` to tag all references to a type
and its descendants while compiling an access policy defined on the
type. Since it needs to be based at the *source* of the type, we need
to compile each rule individually in order to compile them with the
proper exclusions.

2. DML policies also are evaluated without restriction on those types

Currently SELECT policies will be applied during the evaluation of DML
polices.  Change this for consistency, to avoid bizarre scenarios such
as being unable to update a SELECT-visible object of a type that only
has a single `allow all` rule, because it evaluates as true when
interpreted as a `select` rule but false when interpreted as an
`update read` rule.

3. Evaluate access policy rules in a detached context

Non-detached schema exprs cause a lot of trouble when they interact
with inheritance (see #4142). We discovered that issue too late to
deal with it properly for 2.0, since it will require a compatibility
break and a recovery strategy, but let's get it right for new features.

### DIFF
--- a/edb/edgeql/compiler/context.py
+++ b/edb/edgeql/compiler/context.py
@@ -406,6 +406,9 @@ class ContextLevel(compiler.ContextLevel):
     that had rewrites in its components.
     """
 
+    suppress_rewrites: FrozenSet[s_types.Type]
+    """Types to suppress using rewrites on"""
+
     aliased_views: ChainMap[s_name.Name, s_types.Type]
     """A dictionary of views aliased in a statement body."""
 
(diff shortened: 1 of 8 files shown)


elprans (approved):
Please remember to document the amended behavior.

elprans (approved):

PR #4150, opened 2022-07-27
Prohibit access policy expressions with multi cardinality

msullivan:
I think it's semantically unclear what a multi cardinality value ought
to mean in a `deny` rule. There are two ways to think about what deny
rules should mean:
 1. Any object for which a deny rule returns `true` should be removed
    from the candidate set of objects
 2. We evaluate the following condition as a filter on each object
   `(allow_1 OR ... OR allow_n) AND NOT deny_1 ... AND NOT deny_m`

These definitions coincide if each rule has cardinality ONE, but they
differ if a rule can have cardinality AT_LEAST_ONE: if a rule returns
`{true, false}`, then definition 1 would exclude the object while
definition 2 would include it.

Currently the implementation follows definition 2, because it operates
based on boolean arithmetic; I think that if we had to choose that
definition 1 would be less surprising behavior, but that the best
option is to disallow multi cardinality, which lets our two ways of
thinking about the rules coincide.

We can always add it back in later if we want.

### DIFF
--- a/edb/schema/policies.py
+++ b/edb/schema/policies.py
@@ -164,6 +164,14 @@ class AccessPolicyCommand(
                     context=srcctx
                 )
 
+            if expression.irast.cardinality.is_multi():
+                srcctx = self.get_attribute_source_context(field)
+                raise errors.SchemaDefinitionError(
+                    f'possibly more than one element returned by {vname} '
+                    f'expression for the {pol_name} ',
+                    context=srcctx
+                )
+
             target = schema.get(sn.QualName('std', 'bool'), type=s_types.Type)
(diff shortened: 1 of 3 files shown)


msullivan:
We could also only disallow them in `deny` rules, if anybody is really excited about having them in `allow` rules (where there aren't any semantic problems), but my inclination is to not bother with the distinction

elprans (approved):
The intended interpretation of a "deny" rule is "exclude all objects matching filter", and not a straight logical negation of the `using` expression and I would argue that the current implementation is wrong and there is no ambiguity, but since there's no immediate consensus on this and we're so close to the release, r+.

PR #4162, opened 2022-07-28
Inconsistency in quickstart tutorial

ks93:
`year` vs. `release_year`


colinhacks:
Thanks, good catch! We're taking the long way round for this one though: https://github.com/edgedb/edgedb/pull/4163 It's more consistent with some other sample schemas we use.

PR #4182, opened 2022-08-01
Fix global defaults with nontrivial computation

### LINKED ISSUE #4173: Can't create required global with computation in default
This works

```
required global somestr -> str {
  default := '00000000-0000-0000-0000-000000000000';
};
```

But for some reason this doesn't:

```
required global someuuid -> uuid {
  default := <uuid>'00000000-0000-0000-0000-000000000000';
};
```

```
$ edgedb migration create
error: required globals must have a default
   ┌─ dbschema/default.esdl:16:3
   │  
16 │ ╭   required global someuuid -> uuid {
17 │ │     default := <uuid>'00000000-0000-0000-0000-000000000000';
18 │ │   };
   │ ╰───^ error
```

msullivan:
The SET DEFAULT in a global must never be split off into a separate
command by `declarative`. Aliases had a similar special case; extend
it to cover globals.

Fixes #4173.

### DIFF
--- a/edb/edgeql/declarative.py
+++ b/edb/edgeql/declarative.py
@@ -1024,7 +1024,8 @@ def _register_item(
             elif (isinstance(cmd, qlast.SetField)
                   and not cmd.special_syntax
                   and not isinstance(cmd.value, qlast.BaseConstant)
-                  and not isinstance(op, qlast.CreateAlias)):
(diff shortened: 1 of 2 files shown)


elprans (approved):

msullivan:
Cherry-picked for 2.x

PR #4183, opened 2022-08-01
Fix migration that removes policy using clause

### LINKED ISSUE #4175: `Unexpected EXPRESSION` during `edgedb migration create`
Starting schema:

```
type Test2 {
  property name -> str;
  access policy asdf allow all using true;
}
```

Create/apply migration

```
$ edgedb migration create
did you create object type 'default::Test2'? [y,n,l,c,b,s,q,?]
> y
Created dbschema/migrations/00042.edgeql, id: m1yrbvsi62ewhzkvl44oxo6ogbd7xx42rqdqxdq7kdkn7sf3ppjeza
$ edgedb migrate         
Applied m1yrbvsi62ewhzkvl44oxo6ogbd7xx42rqdqxdq7kdkn7sf3ppjeza (00042.edgeql)
```

Remove the `using` clause:

```diff
  type Test2 {
    property name -> str;
-   access policy asdf allow all using true;
+   access policy asdf allow all;
  }
```

Create migration

```
$ edgedb migration create
did you alter access policy 'asdf' of object type 'default::Test2'? [y,n,l,c,b,s,q,?]
> y
error: EdgeQLSyntaxError: Unexpected 'EXPRESSION'
  ┌─ query:2:36
  │
2 │     ALTER ACCESS POLICY asdf RESET EXPRESSION;
  │                                    ^^^^^^^^^^ error
Rolling back last operation...
```

You can work around this by answering `n`, in which case the migration will drop the old policy and add a new one with the same name.

msullivan:
I had made the (100% undocumented and untested) syntax for resetting a
policy USING clause be `RESET USING`. This was inconsistent with how
to reset every other single field that is set with `USING`, which is
`RESET EXPRESSION`, as well as the generic code in codegen that
generates it.  Switch it to `RESET EXPRESSION` and document it.

There was a schema test for removing a USING clause from a policy, but
that flow only ever operates on ASTs, and so the mismatch between the
codegen and the parser was undetecgted.

Fixes #4175.

### DIFF
--- a/docs/reference/ddl/access_policies.rst
+++ b/docs/reference/ddl/access_policies.rst
@@ -143,6 +143,7 @@ Alter access policy
         [ reset when ; ]
         { allow | deny } <action> [, <action> ... ; ]
         [ using (<expr>) ; ]
+        [ reset expression ; ]
         [ create annotation <annotation-name> := <value> ; ]
         [ alter annotation <annotation-name> := <value> ; ]
         [ drop annotation <annotation-name>; ]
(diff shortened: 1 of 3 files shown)


elprans (approved):

msullivan:
Cherry-picked for 2.x

PR #4187, opened 2022-08-02
Don't create constraints on derived views when adding a pointer to a type

### LINKED ISSUE #4186: Adding links to types that have a global view puts schema into wedged state
- EdgeDB Version: 
EdgeDB CLI 2.0.1+93a7fab
EdgeDB 2.0+710b779
- OS Version: Linux LTS

Can't create migration on provided schema for:
- removing .email index on member's object (member.esdl);
- removing .email exclusive constraint on member's object;

Steps to Reproduce:

1. Create first migration and apply;
2. remove .email index or exclusive constraint from member's object SDL;
3. create migration;

Schema:
[schema](https://github.com/lu-zen/edgedb-migration-creation-bug/)

msullivan:
If we have a derived schema view of a type (a global or an alias) and
try to add a pointer with a constraint on it to the base type, we will
also add the constraint to the derived, not real, view type. This is
a bad problem, because we do *not* include constraints and the like
when directly creating these views.

Make sure that we avoid doing this.

Also fix another bug I ran into when trying to test if this works in
the rebase case.

Fixes #4186.

### DIFF
--- a/edb/schema/globals.py
+++ b/edb/schema/globals.py
@@ -131,6 +131,10 @@ class GlobalCommand(
 
         expression = self.get_attribute_value('expr')
         assert isinstance(expression, s_expr.Expression)
+        # If it's not compiled, don't worry about it. This should just
+        # be a dummy expression.
+        if not expression.irast:
+            return schema
         assert isinstance(expression.irast, irast.Statement)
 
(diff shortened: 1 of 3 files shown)


elprans (approved):

msullivan:
Cherry-picked to 2.x

PR #4189, opened 2022-08-02
Allow multiple FDs per socket in activation

fantix:
Needed by edgedb/edgedb-cli#798

### DIFF
--- a/edb/server/main.py
+++ b/edb/server/main.py
@@ -191,7 +191,7 @@ async def _run_server(
             compiler_pool_addr=args.compiler_pool_addr,
             nethosts=args.bind_addresses,
             netport=args.port,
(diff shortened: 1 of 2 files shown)


elprans (approved):

fantix:
Do you mean that we should *merge* the activation socket ports with the configured ports, instead of overwriting? Like all `host:port` in activation sockets should work, while all `listen_addresses`:`listen_port` should work too?

elprans:
> Do you mean that we should _merge_ the activation socket ports with the configured ports, instead of overwriting? Like all `host:port` in activation sockets should work, while all `listen_addresses`:`listen_port` should work too?

It's fairly marginal, because in practice `--port` is the same as activated port, but do it if you feel like it.

PR #4190, opened 2022-08-02
Document the allow_user_specified_id settings

fmoor:


fmoor:
Is there more that should be said about this setting?

docs/stdlib/cfg.rst
```diff
@@ -85,6 +85,9 @@ Query planning
 Client connections
 ------------------
 
+:eql:synopsis:`allow_user_specified_id -> bool`
+  Makes it possible to set the `.id` property when inserting new objects.
```
elprans, line 89: We should add a warning about the security implications of this and that custom object ids should come from a trusted source.

docs/stdlib/cfg.rst
```diff
@@ -85,6 +85,19 @@ Query planning
...
+
+  Enabling this feature introduces some security vulnerabilities:
+
+  1. An unprivileged user can discover ids that already exist in the database
+     by trying to insert new values and noting when there is a constraint
+     violation on `.id` even if the user doesn't have access to the relevant
```
fmoor, line 95: I'm not sure I got all the details right. I pretty much just copied this: https://github.com/edgedb/edgedb/issues/3510#issuecomment-1189419879

docs/stdlib/cfg.rst
fmoor: added

1st1 (approved):

PR #4196, opened 2022-08-02
Avoid doing semi-joins after a sequence of single links

### LINKED ISSUE #4160: Slow query when filtering by second link
- EdgeDB Version: 1.4+ffe1237 
- OS Version: your CloudFormation template

Simplified schema:
```
Person { multi link personPosts := .<person[is Profile].profilePosts; }
Profile {
  link person -> Person {
  }
  multi link profilePosts := .<owner[is Post];
}
IGProfile {}
TTProfile {}
Post { required link owner -> Profile; }
IGPost { overloaded required link owner -> IGProfile; }
TTPost { overloaded required link owner -> TTProfile; }
```

These queries are both fast:
```
select Post filter .owner.id = <uuid>"";
select Person { personPosts } filter .id = <uuid>"";
```

But this one take 5+ seconds:
```
select Post filter .owner.person.id = <uuid>"";
```

Number of objects:
Person: 100
Profile: 150k
Post: 431k

I'm happy to share more details if it would be helpful.

msullivan:
If we have a query like
```
select Post filter .owner.person.id = <uuid>$uid
```
where `owner` is `single`, there is no need to do a semi-join on
`.person`. This can have major performance impacts.

Fixes #4160.

### DIFF
--- a/edb/pgsql/compiler/relgen.py
+++ b/edb/pgsql/compiler/relgen.py
@@ -1317,7 +1317,12 @@ def process_set_as_subquery(
 
             if is_objtype_path and not source_is_visible:
                 # Non-scalar computable semi-join.
-                semi_join = _source_path_needs_semi_join(ir_source, ctx=ctx)
+
+                # TODO: The basic path case has a more sophisticated
+                # understanding of when to do semi-joins. Using that
+                # naively here doesn't work, but perhaps it could be
+                # adapted?
+                semi_join = True
 
                 # We need to compile the source and include it in,
                 # since we need to do the semi-join deduplication here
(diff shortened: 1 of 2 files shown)


edb/pgsql/compiler/relgen.py
```diff
@@ -964,6 +964,32 @@ def process_set_as_path_type_intersection(
         ir_set, stmt, aspects=['value', 'source'], ctx=ctx)
 
 
+def _source_path_is_visible(
```
elprans, line 967: I would use something like `_needs_semi_join`

elprans (approved):

PR #4197, opened 2022-08-03
Add SCRAM authentication over HTTP

### LINKED ISSUE #4172: Add support for SCRAM auth to HTTP-tunneled binary protocol

fantix:
Fixes #4172

### DIFF
--- a/edb/server/protocol/auth/scram.py
+++ b/edb/server/protocol/auth/scram.py
@@ -32,7 +32,7 @@ from edb.common import markup
 
 SESSION_TIMEOUT = 30
 SESSION_HIGH_WATER_MARK = SESSION_TIMEOUT * 10
-sessions = collections.OrderedDict()
+sessions: collections.OrderedDict[str, tuple] = collections.OrderedDict()
 
 
 def handle_request(request, response, server):
--- a/edb/server/protocol/auth/scram.py
+++ b/edb/server/protocol/auth/scram.py
@@ -1,0 +1,316 @@
+#
+# This source file is part of the EdgeDB open source project.
+#
+# Copyright 2022-present MagicStack Inc. and the EdgeDB authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import base64
+import collections
+import hashlib
+import http
+import os
+import time
+from jwcrypto import jwt
+
+from edgedb import scram
+
+from edb.common import debug
+from edb.common import markup
+
+
+SESSION_TIMEOUT = 30
+SESSION_HIGH_WATER_MARK = SESSION_TIMEOUT * 10
+sessions = collections.OrderedDict()
+
+
+def handle_request(request, response, server):
+    if not request.authorization:
+        response.status = http.HTTPStatus.UNAUTHORIZED
+        response.custom_headers["WWW-Authenticate"] = "SCRAM-SHA-256"
+        return
+
+    scheme, _, auth_str = request.authorization.decode("ascii").partition(" ")
+    if scheme != "SCRAM-SHA-256":
+        response.body = (
+            b"Client selected an invalid SASL authentication mechanism"
+        )
+        response.status = http.HTTPStatus.UNAUTHORIZED
+        response.custom_headers["WWW-Authenticate"] = "SCRAM-SHA-256"
+        return
+
+    data = None
+    sid = None
+    try:
+        for kv_str in auth_str.split():
+            key, _, value = kv_str.rstrip(",").partition("=")
+            if key == "data":
+                data = base64.b64decode(value.strip('"')).strip()
+            elif key == "sid":
+                sid = value.strip('"')
+        if data is None:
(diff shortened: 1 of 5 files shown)


edb/server/protocol/auth/scram.py
```diff
@@ -0,0 +1,316 @@
...
+        if debug.flags.server:
+            markup.dump(ex)
+        response.body = str(ex).encode("ascii")
+        response.status = http.HTTPStatus.BAD_REQUEST
+        response.close_connection = True
+        return
+
+    if (
+        not server.get_jwe_key().has_public
+        or not server.get_jws_key().has_private
```
jaclarke, line 74: Should `has_public` and `has_private` be the other way around here? In `server.py` the `jwe_key` has the private key, and `jws_key` has the public key: https://github.com/edgedb/edgedb/blob/master/edb/server/server.py#L1611-L1628

fantix: Oh this was intentional - in `edb/server/server.py` it uses the keys for decryption and signature verification, `scram.py` instead uses them for signing and encryption. That is to say, the remote instance must be a "development" instance, which has both private and public keys in both JOSE keys. Production instances should not usually use the `--jose-key-mode=generate` and have their own ways to generate/utilize the JOSE keys, like the EdgeDB Cloud instances, and in this case, HTTP SCRAM login is disabled (I think that's okay?).

tests/test_http_auth.py
```diff
@@ -0,0 +1,170 @@
...
+    @classmethod
+    def get_api_path(cls) -> str:
+        return "/"
+
+    def test_http_auth_scram(self):
+        args = self.get_connect_args()
+
+        with self.http_con() as con:
+            _, headers, status = self.http_con_request(
+                con, {}, path="auth/scram"
```
elprans, line 40: Is there a point in duplicating the auth scheme in path when we already indicate it in headers?

fantix: So just `GET /auth`? Or like @jaclarke suggested, maybe something like `GET /auth_token`? I'm okay with either. Please advise.

elprans: Let's go with `GET /auth/token`

elprans (approved):

edb/server/protocol/auth/scram.py
```diff
@@ -0,0 +1,310 @@
...
+    rolerec = server.get_roles().get(user)
+    if rolerec is None or rolerec["password"] is None:
+        raise ValueError(f"invalid user: {user!r}")
+    ekey = server.get_jwe_key()
+    skey = server.get_jws_key()
+
+    namespace = "edgedb.server"
+    token = jwt.JWT(
+        header={"alg": "ES256" if skey["kty"] == "EC" else "RS256"},
+        claims={f"{namespace}.roles": {user: rolerec["password"]}},
```
elprans, line 299: Let's also include the standard `iat` (issued at) claim to control for token expiration.  The latter would probably be configurable somehow, but let's worry about that later.

jaclarke:
Not sure if this is a problem, but you can tell if a username exists without a password.

Invalid username:
```
sending first, Authorization: SCRAM-SHA-256 data=biwsbj1pbnZhbGlkLHI9dzloYTBjVVh5ZFp6VERFcTZBRUhKV3Nz 

got server first, status 401, WWW-Authenticate: SCRAM-SHA-256 sid=D18ke6ziWmrOT1Y3_1JfGg, data=cj13OWhhMGNVWHlkWnpUREVxNkFFSEpXc3NIVFN4UEZhbmJnOXNVRjJLS0JyZ0hwSUUscz12SWhya1lsR3F3aXZNdWs2bDMrV2pRPT0saT00MDk2

sending final, Authorization: SCRAM-SHA-256 sid=D18ke6ziWmrOT1Y3_1JfGg, data=Yz1iaXdzLHI9dzloYTBjVVh5ZFp6VERFcTZBRUhKV3NzSFRTeFBGYW5iZzlzVUYyS0tCcmdIcElFLHA9djNLOFdMeU0venI1SDZDcHBWNW9WTVdOQWRieWoyOGc0Y2ZmRmI0blRtRT0=

got server final, status 401, Authentication-Info: null
```

Valid username (with password), Invalid password provided:
```
sending first, Authorization: SCRAM-SHA-256 data=biwsbj10ZXN0LHI9NCtqcUUzZ042OXRieVI0OWp4Z1NkcDAx 

got server first, status 401, WWW-Authenticate: SCRAM-SHA-256 sid=gU7nK6JPkz-BZRdlkE5dbw, data=cj00K2pxRTNnTjY5dGJ5UjQ5anhnU2RwMDFnWUtJb1ZSYTF1aWRwTUZHcEJLQ2gyTnoscz0zM2F0QlFlQ1VEL2RkNDJTWXhhU0Z3PT0saT00MDk2 

sending final, Authorization: SCRAM-SHA-256 sid=gU7nK6JPkz-BZRdlkE5dbw, data=Yz1iaXdzLHI9NCtqcUUzZ042OXRieVI0OWp4Z1NkcDAxZ1lLSW9WUmExdWlkcE1GR3BCS0NoMk56LHA9Yi9QcTFqeHdyWldCZUl6SVN3ZWZZNXRBN2srV3ZYVVQySlNmdWs5YnRHWT0=

got server final, status 401, Authentication-Info: null
```

Valid username (without password):
```
sending first, Authorization: SCRAM-SHA-256 data=biwsbj10ZXN0MixyPVpDR3ZweGxxWTV6NDh0OWRNUkFoODVqcw== 

got server first, status 401, WWW-Authenticate: SCRAM-SHA-256
```

fantix:
Hmm right, the no-password case should also use the mock verifier. I'll push a fix

Neustradamus:
@fantix: Good job!

Linked to:
- https://github.com/edgedb/edgedb/pull/4172
- https://github.com/edgedb/edgedb/pull/4197

PR #4204, opened 2022-08-03
Rename test_dump04.py to test_dump_v2.py

msullivan:
Since we're likely going to want to introduce a new dump test for each
major version we release, I want to name them based on major release
version numbers, moving forward.

I'll cherry-pick this commit onto our 2.0 branch to build dumps.


elprans (approved):

msullivan:
Ha, if I cherry pick it to build the dumps, then it thinks the version is 2.1, so instead I am running `gen-test-dumps` as uncommitted changes with `EDGEDB_BUILD_IS_RELEASE=1` set

PR #4207, opened 2022-08-03
Make range() properly strict in its non-optional arguments

msullivan:
range() is currently not strict in its non-optional arguments:
 * NULL empty is treated as false
 * NULL inc_lower and inc_upper produce errors

Fix this by adding NULL checks when necessary.
Update casting from JSON, which relied upon this behavior.


elprans (approved):

msullivan:
Cherry-picked for 2.x

PR #4214, opened 2022-08-05
Always arm auto-shutdown timer when it's greater than zero

elprans:
Currently, the `--auto-shutdown-after` is always armed after last client
disconnection, which means that the server will continue to run
indefinitely if no client connections have happened.  Change the
behavior of `--auto-shutdown-after` to always arm the timer if the
timeout is greater than zero.  Zero timeout keeps the original behavior
of shutting down after last client disconnection.


fantix (approved):
That looks reasonable to me

PR #4217, opened 2022-08-05
Fix json -> array<json> cast of '[]'

msullivan:
We need to coalesce the result of array_agg so that we get [] instead
of {}, but for that to work we need to check for null input
separately.


vpetrovykh (approved):
LGTM

elprans (approved):

Issue #4230, opened 2022-08-08
Types with only computed fields break GraphQL schema


dardanbujupaj:
- EdgeDB Version: 2.1+9c4bc9f 
- OS Version: MacOS Monterey 12.4

Steps to Reproduce:

1. Create a database containing a type that contains only a computed value, e.g.:
```
using extension graphql;

module default {
  type User {
    property username := 1;
  }
}
```
2. Visit the graphiql explorer -> querying ist not possible and an error message containing something like this is shown: `TypeError: Input Object type UpdateCommentable must define one or more fields`.
3. To fix it, either remove the computed value or add a actual property/link

Schema:
Here is a schema for an actual use case (simplified) where this can happen.
```
using extension graphql;

module default {
  abstract type Commentable {
    multi link comments := .<subject[is Comment]
  }
  
  type User {
    required property username -> str;
  }

  type Board extending Commentable {
    required property name -> str;
    multi link notes := .<board[is Note];
    link owner -> User;
  }

  type Note extending Commentable {
    required property description -> str;
    link board -> Board;
  }

  type Comment extending Commentable {
    required link subject -> Commentable;
    required property content -> str;
  }
}
```

dardanbujupaj:
Seems to be resolved in recent versions.

PR #4231, opened 2022-08-08
Check for computables when generating update types

dardanbujupaj:
This PR adds a check for computed fields before generating update types.

This should fix https://github.com/edgedb/edgedb/issues/4230 as empty update types can't be generated anymore. 
Unfortunately I'm not able to test it properly as I have issues getting the local development server working...

### DIFF
--- a/edb/graphql/types.py
+++ b/edb/graphql/types.py
@@ -1298,7 +1298,7 @@ class GQLCoreSchema:
                 # only objects that have at least one non-readonly
                 # link/property are eligible
                 pointers = t.get_pointers(self.edb_schema)
-                if any(not (p.get_readonly(self.edb_schema) or 
+                if any(not (p.get_readonly(self.edb_schema) or
                        p.is_pure_computable(self.edb_schema))
                        for _, p in pointers.items(self.edb_schema)):
                     gqlupdatetype = GraphQLInputObjectType(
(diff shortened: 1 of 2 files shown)


dardanbujupaj:
I just realized, that a similar check should probably be introduced for the Insert types. But I'm a bit confused, as there's not even a check regarding the readonly properties.

vpetrovykh:
We definitely want a test in the test suite that actually reproduces the #4230 bug so that we can be sure that this fixes it as well as that future changed don't break it again.

Whether we need a similar check for insert types depends on whether this bug is reproducible in Insert types.

So @dardanbujupaj, if you feel up to it, you can come back after you figure out what went wrong with your local development server and write a test for this.

dardanbujupaj:
@vpetrovykh Finally got the test environment to work.

I updated the PR, but instead of adding explicit tests, I updated `tests/schemas/graphql_schema.esdl` to contain a type according to the description in https://github.com/edgedb/edgedb/issues/4230.

This leads to all `graphql_schema`-tests failing and the changes to `edb/graphql/types.py` resolve this as expected.

You can check the failing tests at commit https://github.com/edgedb/edgedb/pull/4231/commits/5db40e0fa6a1e0a0d0ab996f73676ef7bb6c9d5a with  `edb test -k test_graphql_schema`.

dardanbujupaj:
This seems to be fixed in recent releases 👍🏻

Issue #4245, opened 2022-08-09
Migrations fail for exclusive constraints on links with access policies


CosmicPangolin:
- EdgeDB Version:  nightly │ 3.0-dev.6840+0581c41, but also v2 kube container on GCP
- OS Version: MacOS Monterey 12.4

Steps to Reproduce:

1. Try to migrate the following schema.  migration create works, but migrate throws an invalid constraint definition error:
![image](https://user-images.githubusercontent.com/40650466/183781544-ab8c90be-4b39-47d3-8bb8-7e5a2bef42c2.png)

2. Delete access policy on entity, try again.  Success!

If you remove the access policy on Entity it's fine.  If you incrementally add access policies after migrating the core models it’s fine, but fails on additional migrations of those properties (such as moving the link from an abstract to extending type).

Schema: 
```using extension edgeql_http;

module default {

  global role -> str;
  global firebaseUserId -> str;

  abstract type Entity {
    access policy privateAndLockedByDefault allow all using (false);

    property insertTime -> datetime {
      default := datetime_current();
      readonly := true;
    }

    property updateTime -> datetime {
      default := datetime_current();
    }
  }

  type User extending Entity {
    link customer := .<user[is Customer];

    required property displayName -> str;
    required property email -> str { constraint exclusive };
    required property firebaseUserId -> str { constraint exclusive };
    
    property handle -> str { constraint exclusive };
    property mobile -> str;
    property birthday -> datetime;
  }

  type Customer {
    required link user -> User {
      constraint exclusive;
    }
  }
}

PR #4248, opened 2022-08-10
Disable access policy rewrite when compiling constraints

elprans:
Access policies are applied _before_ the constraint is checked anyway
and performing rewrites confuses constraint IR analysis.  This was
partially fixed in #3892, but not all the way.

Fixes: #4245


msullivan (approved):
This looks good. The error, `test_edgeql_migration_eq_collections_25`, is somewhat concerning but seems to clearly be an existing flake?

msullivan:
Something bizarre is happening. The failure seems to reproduce in CI; I haven't reproed it locally, and I can't understand a change in `pgsql` could cause a stack overflow during earlier stages of migrations.

elprans:
Looks like a bug to me, unrelated (but somehow uncovered by) the chage

PR #4261, opened 2022-08-15
Adapt setup.py/pyproject.toml to setuptools 54

elprans:
The "editable mode" has been changed in >=setuptools-54 to no longer
call `setup.py develop` on which we've been relying so far.  And
although the `SETUPTOOLS_ENABLE_FEATURE=legacy-editable` escape hatch
exists, it probably wouldn't for long, so it's a good idea to switch
everything to be compatible with the brave new world of PEP 517/660.

### DIFF
--- a/.github/workflows.src/tests.inc.yml
+++ b/.github/workflows.src/tests.inc.yml
@@ -33,10 +33,14 @@
 
     - name: Install Python dependencies
       if: steps.venv-cache.outputs.cache-hit != 'true'
+      shell: bash
       run: |
-        pip install wheel
-        pip download --use-feature=in-tree-build --dest=$VIRTUAL_ENV/deps .[test,docs,build]
-        pip install -U --no-index --find-links=$VIRTUAL_ENV/deps $VIRTUAL_ENV/deps/*
+        set -e
+        python -m pip install -U pip setuptools wheel build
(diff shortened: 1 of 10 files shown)


fantix:
I think the `python setup.py build_ext` command cannot find the `setuptools_rust` build dependency now.

elprans:
> I think the `python setup.py build_ext` command cannot find the `setuptools_rust` build dependency now.

You are not supposed to run `setup.py` commands without installing the build dependencies first.

fantix:
> You are not supposed to run setup.py commands without installing the build dependencies first.

I see. In that case, suggested change:

```diff
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,9 @@ casts:

 build:
        find edb -name '*.pyx' | xargs touch
-       pip install -Ue .[docs,test]
+       python -m pip install -U pip setuptools build
+       python -c "import build, os; os.system('python -m pip install ' + ' '.join(f'\'{x}\'' for x in build.ProjectBuilder('.').build_system_requires))"
+       python -m pip install --no-build-isolation -Ue .[docs,test]```

elprans:
> I see. In that case, suggested change:

Done.

fantix (approved):

Issue #4253, opened 2022-08-11
Add support for musl-based Linux systems


notramo:
- EdgeDB Version: latest
- OS Version: Void Linux (musl)

Installed using `curl https://sh.edgedb.com|sh`
It seems like the official package is dynamically linked against glibc. Could you statically link it so it works both on musl and glibc systems?
```
Found `edgedb.toml` in /home/notramo/Downloads/testing_db
Initializing project...
Specify the name of EdgeDB instance to use with this project [default: testing_db]:
> testing_db
Checking EdgeDB versions...
┌────────────────────────┬───────────────────────────────────────────┐
│ Project directory      │ /home/notramo/Downloads/teszt             │
│ Project config         │ /home/notramo/Downloads/teszt/edgedb.toml │
│ Schema dir (non-empty) │ /home/notramo/Downloads/teszt/dbschema    │
│ Installation method    │ portable package                          │
│ Version                │ 2.1+c73716f                               │
│ Instance name          │ testing_db                                │
└────────────────────────┴───────────────────────────────────────────┘
Version 2.1+c73716f is already installed
Initializing EdgeDB instance...
edgedb error: cannot bootstrap EdgeDB instance: bootstrap failed to start (command-line: "/home/notramo/.local/share/edgedb/portable/2.1/bin/edgedb-server" "--bootstrap-only" "--data-dir" "/home/notramo/.local/share/edgedb/data/.~testing_db.tmp" "--runstate-dir" "/run/user/1000/edgedb-testing_db" "--tls-cert-mode=generate_self_signed" "--jose-key-mode=generate" "--bootstrap-command" "\n            ALTER ROLE edgedb {\n                SET password_hash := \"SCRAM-SHA-256$4096:Bj9ILlReFmGJHKdhvtHEJQ==$NF1NAYemoy+1XDczB4HevsIUWmuS5+h0x0jxOGeBdSY=:1Nitolt/YdHIwhoKO3Y4dYg82xjMc7nXdYejsJcWEAg=\";\n            };\n            "): No such file or directory (os error 2)
```

elprans:
> Could you statically link it so it works both on musl and glibc systems?

That would be hard to do, because the server package is not a simple executable, it's an omnibus package of Python, Postgres, EdgeDB and a bunch of their dependencies.

A more realistic fix would be to have a musl-linked build, which should be doable.

notramo:
A musl-linked build would also be a good option, the CLI client already supports it.

PR #4273, opened 2022-08-17
Build musl versions of server

elprans:
Fixes: #4253

### DIFF
--- a/.github/workflows.src/nightly.targets.yml
+++ b/.github/workflows.src/nightly.targets.yml
@@ -95,6 +95,18 @@ targets:
           platform_version: aarch64
           family: generic
           runs_on: [self-hosted, linux, arm64]
+        - name: linuxmusl-x86_64
+          platform: linux
+          platform_version: x86_64
+          platform_libc: musl
+          family: generic
+          runs_on: [self-hosted, linux, x64]
+        - name: linuxmusl-aarch64
+          platform: linux
+          platform_version: aarch64
+          platform_libc: musl
+          family: generic
+          runs_on: [self-hosted, linux, arm64]
 
     macos:
         - name: macos-x86_64
--- a/.github/workflows.src/nightly.tpl.yml
+++ b/.github/workflows.src/nightly.tpl.yml
@@ -25,7 +25,7 @@ jobs:
       id: whichver
 
 <% for tgt in targets.linux %>
-<% set plat_id = tgt.platform + ("-{}".format(tgt.platform_version) if tgt.platform_version else "") %>
+<% set plat_id = tgt.platform + ("{}".format(tgt.platform_libc) if tgt.platform_libc else "") + ("-{}".format(tgt.platform_version) if tgt.platform_version else "") %>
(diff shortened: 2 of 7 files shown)


fantix (approved):

tailhook:
I'm not sure what this means. Is this a statically-linked musl build? Or does it require musl libc to be present on the system? Does it work on Alpine?

Will we have any issues with the musl build because of smaller default C stack? Will it fit standard python's recursive depth limit? Does postgres work well on such a small stack, or does it have a mechanism to adjust the stack size?

elprans:
> I'm not sure what this means. Is this a statically-linked musl build? 

Dynamically linked.

> Does it work on Alpine?

Yes, I also tested Void Linux and a `FROM scratch` container with a custom-built musl.

> Will it fit standard python's recursive depth limit? Does postgres work well on such a small stack, or does it have a mechanism to adjust the stack size?

Well, alpine is a fairly popular base image for Python, I've not seen anybody complain about it.  Postgres is also packaged by Alpine without any specific hacks, AFAIK stack size isn't hardcoded and queried via `RLIMIT_STACK` or a similar technique.  FWIW, the test suite passes just fine.

tailhook:
> AFAIK stack size isn't hardcoded and queried via `RLIMIT_STACK` or a similar technique.

Here is the quote from musl documentation:

> The default stack size for new threads on glibc is determined based on the resource limit governing the main thread’s stack (RLIMIT_STACK). It generally ends up being 2-10 MB.
>
> musl provides a default thread stack size of 128k (80k prior to 1.1.21).

Note: for other threads than main, you can change the stack size on thread creation. But as far as I know there is no such method for main thread with musl libc.

On the other hand:
> Since 1.1.21, musl supports increasing the default thread stack size via the PT_GNU_STACK program header, which can be set at link time via -Wl,-z,stack-size=N.

So this is a compile-time flag.

> Well, alpine is a fairly popular base image for Python, I've not seen anybody complain about it

It looks like Python build in Alpine is specially tweaked to account that:
* https://github.com/python/cpython/issues/76488#issuecomment-1093770247
* https://git.alpinelinux.org/aports/tree/main/python3/APKBUILD?h=3.16-stable#n119

So it looks like it isn't autodetected. Since you aren't using Alpine's package, can you make sure that this flag is set?

For postgres, I can't find anything relevant in Alpine package, so maybe it does some magic internally, or maybe nobody uses postgres Alpine package for anything serious. Hard to say.

elprans:
> So it looks like it isn't autodetected. Since you aren't using Alpine's package, can you make sure that this flag is set?

Sure.  FWIW, Python 3.11 no longer uses the C stack for Python calls, so this wouldn't be an issue once we switch.

notramo:
> For postgres, I can't find anything relevant in Alpine package, so maybe it does some magic internally, or maybe nobody uses postgres Alpine package for anything serious. Hard to say.

The Postgres official Docker images provide two versions: Debian Bullseye, and Alpine.
So I guess Postgres works well on Alpine if the upstream devs ship it.

PR #4275, opened 2022-08-18
Add Get Started section

colinhacks:
To run this locally check out the `getting-started` branches of both `edgedb/edgedb` and `edgedb/edgedb.com`.

--------

This is a change that's been percolating in my head for a while. This is just a rough draft to see if this is a change we want to pursue. You can run this using the `master` branch of `edgedb.com`, but you'll need to manually type http://localhost:3000/docs/intro/index into the URL bar to get to the docs home page.

Basic idea is to provide a simple "Get started" section up front that new users can go through more or less linearly. 

Screen Shot 2022-08-17 at 11 20 29 PM

- "Using the CLI" has installation instructions and a simple CLI cheatsheet. 
- "Defining a schema" is the SDL primer
- "Creating migrations" is a modified version of https://www.edgedb.com/docs/guides/migrations/index
- "Writing queries" is the EdgeQL primer
- "Running queries" is a new page breaking down the major concepts that are common to all client libraries: creating clients, connection resolution, query vs querySingle, etc. We can use the new code tabs widget to provide basic code samples (how to initialize a client, how to execute a Hello World query) for each client. 

The Guides section is moved down underneath EdgeQL. It contains stuff that isn't immediately relevant to new users (Deployment, Introspection, Migration patterns) — ideally these wouldn't be in your face when you open the docs for the first time.

Screen Shot 2022-08-17 at 11 30 45 PM

Tutorials is still under Guides but I could be convinced to put it under Get Started.


i0bs:
Awesome start to the PR! I love this idea, having a Get Started section gives first-time EdgeDB users a more simplified area for trying to learn things. I believe a lot of the current documentation is well-written and requires no revision to keep.

There are some things I'd like to suggest, however, to make user navigation better. Here's my basic outline:

## Make the Guides section unique.
The first thing I'd like to suggest is renaming "Tutorials" to "Using frameworks." This is because all of the examples we see underneath it are really just specific use cases for a framework like Flask or Next.js. Nothing wrong with that, I just think that the common consensus is that tutorials teach you how to use something specifically relating to the core service when, in reality, these are more like "extensions of applicable use."

Another suggestion I have is moving "Developing EdgeDB" under "Reference." Theory-wise, most people wanting to build EdgeDB locally are also trying to work near DSN and server configurations. 

## Add a Developers section.
I think there's a good chunk of users who want to develop something with/for EdgeDB as their own solution, despite already existing ways to do things.

An alternative suggestion I have is to introduce a "Developers" section that brings forth more detail towards standards for using EdgeDB/client libraries, any "no-no's" when using EdgeDB, and a general "how-to" for wanting to integrate services e.g. CLI or admin UI.

## Clarifying types.
This is a huge gripe point I've found to exist between many users of EdgeDB. There is a large amount of confusion between "Schema," "Standand Library" and "EdgeQL." Understandably, these all make sense to be kept separate because they're not the same. Schema represents schema definition for correlative typing, Standard Library has all builtin defined types for EdgeDB, and EdgeQL is.. well, a query language to help with execution.

The issue is that, right now the documentation makes it *really* painful to jump between all 3 of these when looking for something. The current documentation has no pointers, nor a search bar or consolidated access point. It wouldn't make sense for all of these things to be merged under 1 section, and neither would it be beneficial to the user navigating.

With this, I'd like to suggest something a little more interim but plausible: creating an "EdgeDB" section in the documentation. It would respect this hierarchy:

```
EdgeDB
|_ Overview
|_ EdgeQL
   |_ Primer
   |_ ...
|_ Schema
|  |_ Primer
|  |_ ...
|_ Standard Library
   |_ Generic
   |_ ...
```

The "Overview" for this section should go under more detail about the difference between all 3. What I've found to partially be an issue for many is duplicated section parts, such as "Primer" being in both EdgeQL and Schema. This causes a lot of user confusion by first-glance when trying to navigate.

The other half of the problem I've found is a lack of user education between what they are. I believe EdgeQL is already explained pretty well, and so is the Standard Library documentation. What seems to cause the most confusion is the Schema and how it exactly "differs" from the Standard Library.

See between "Types" in Standard Library and "Object Types" in Schema. Instead of the STL talking about the type itself, it shows information about operators that can be called upon a type. This makes the user have to navigate elsewhere now to Schema to learn about specific types. My suggestion is to either explain better in the Get Started overview or in the overview of this proposed section.

## Consistent pointers between STL and Schema.
Simply put, the issue I've talked about above creates a lot of confusion for the user navigating. There are some areas, such as "Enum" where the docs nicely point you to STL for more information.. and then a lot of areas where they aren't.

I think the suggestion above could be supplemented with us adding hyperlinks pointing to other areas, in case the user actually wanted details about that "other" thing. This may not be necessary, however, if the proposed structure brings back a lot more clarity.

1st1:
> An alternative suggestion I have is to introduce a "Developers" section that brings forth more detail towards standards for using EdgeDB/client libraries, any "no-no's" when using EdgeDB, and a general "how-to" for wanting to integrate services e.g. CLI or admin UI.

The name "Developers" is so generic that it tells me nothing, really. I'd just have all the content you suggest under the guides, but really, it should be also mentioned in the reference too (e.g. if there is a no-no for using indexes, it should be in the reference too)

> Clarify types.

I'm -1 on moving the standard library out of the top level. I also like the current layout of Schema / EdgeQL top-categories as they give EdgeDB a good into. Just grouping them under one category solves nothing, actually.


> This is a huge gripe point I've found to exist between many users of EdgeDB. There is a large amount of confusion between "Schema," "Standand Library" and "EdgeQL." 

There's some, I agree. But I also heard a lot of voices of people who love our docs and say it's logical and great. We shouldn't knee jerk and move things just because somebody said something isn't obvious. That said, I do agree with many people pointing that we lack proper cross-references in many cases.

> Consistent references between STL and Schema.

STL=Standard Library? We shorten it to "std". Other than that +1.

1st1:
@colinhacks To avoid this getting rotten we should merge soon and then continue iterating!

i0bs:
> I'm -1 on moving the standard library out of the top level. I also like the current layout of Schema / EdgeQL top-categories as they give EdgeDB a good into. Just grouping them under one category solves nothing, actually.

I agree that blindly nesting all of these under one section would practically solve nothing. Arguably, this idea could be dropped as a whole as long as there's more clarity between what all of these achieve.

That said, I want to think of more reasons for why they should be separated. I personally see them as the bread and butter of EdgeDB. Because of that, I wish for them to be documented to better explain their association between one another.

I agree with the rest of your responses, in hindsight, a lot of them give a better reason for why some things should be kept how they are, or have a lesser change just to make what currently exists more fitting.

colinhacks:
> Renaming "Tutorials" to "Using frameworks"

I'm good with this

> Add a Developers section.

I'm having a hard time visualizing what will go in here. Most tooling on top of EdgeDB can be built directly on top of our client libraries/introspection/etc. You can't really integrate with the CLI or UI right now in a meaningful way. 

> these all make sense to be kept separate because they're not the same

Exactly. The examples people are mostly referencing in Discord (backlinks, link properties, etc) are in the small set of concepts that kinda break the hierarchy. But the hierarchy is still broadly applicable and useful. 
  
> Creating an "EdgeDB" section in the documentation

Having a section called "EdgeDB" is redundent imo - everything in the docs is EdgeDB-related. 

> The "Overview" for this section should go under more detail about the difference between all 3

How is this different from https://www.edgedb.com/docs?

> The current documentation has no hyperlinks pointing from one section to another, nor a search bar

There is a search bar :) 

We're consistent with the cross-linking in the "See also" section at the bottom of every page. More in-page links would be  useful, or perhaps a way to add "marginal notes" linking to other relevant pages 🤔 Also we don't really have links from STL back to the relevant EdgeQL/Schema pages, which could use improvement.

docs/guides/deployment/digitalocean.rst
```diff
@@ -60,27 +60,45 @@ Use these variables to change the password for the default role ``edgedb``.
...
-         --password-from-stdin \
-         --trust-tls-cert \
-         --host $IP \
-         --non-interactive \
-         digitalocean
-   Authenticating to edgedb://edgedb@your-droplet-ip:5656/edgedb
-   Trusting unknown server certificate
+   $ echo edgedb://edgedb:$PASSWORD@$IP > dsn.txt
+
+Copy the value from ``dsn.txt``. Tun the following command to open a REPL
```
i0bs, line 74:
```suggestion
Copy the value from ``dsn.txt``. Run the following command to open a REPL
```

docs/guides/deployment/digitalocean.rst
```diff
@@ -60,27 +60,45 @@ Use these variables to change the password for the default role ``edgedb``.
...
+to the new instance.
```
i0bs, line 75:
```suggestion
instance to the newly established DSN.
```

docs/guides/deployment/digitalocean.rst
```diff
@@ -60,27 +60,45 @@ Use these variables to change the password for the default role ``edgedb``.
...
+   Authenticating to edgedb://edgedb@1.2.3.4:5656/edgedb
+   Trusting unknown server certificate:
+   SHA1:1880da9527be464e2cad3bdb20dfc430a6af5727
    Successfully linked to remote instance. To connect run:
-     edgedb -I digitalocean
+     edgedb -I my_instance
 
-You can now use the EdgeDB instance deployed on DigitalOcean as
-``digitalocean``, for example:
+You can now use the ``-I`` flag to execute CLI commands against your remote
```
i0bs, line 96:
```suggestion
You can now use the ``-I`` CLI flag to execute instance-based commands against your remote
```

docs/guides/deployment/digitalocean.rst
```diff
@@ -181,35 +199,47 @@ Set the security policy to strict.
...
 .. _ref_guide_deployment_digitalocean_link:
 
-Create a local link to the new EdgeDB instance
-==============================================
+Get your instance's DSN
+=======================
 
-To access the EdgeDB instance you've just provisioned on DigitalOcean from your
-local machine run the following command.
+Let's construct your instance's DSN (also known as a "connection string").
```
i0bs, line 207: Is this duplicated? Appears to be the exact same documentation as above for constructing a DSN.

i0bs: Relevant to lines 215-223 in the file change.

docs/intro/cli.rst
```diff
@@ -0,0 +1,128 @@
...
+**Windows Powershell**
+
+.. code-block:: powershell
+
+    PS> iwr https://ps1.edgedb.com -useb | iex
+
+Follow the prompts on screen to complete the installation. The script will
+download the ``edgedb`` command built for your OS and add a path to it to your
+shell environment. To test the installation, run ``edgedb --version`` from the
+command line.
```
i0bs, line 32:
```suggestion
command line. The result should look something like this:
```

docs/intro/cli.rst
```diff
@@ -0,0 +1,128 @@
...
+
+.. code-block:: bash
+
+    $ edgedb --version
+    EdgeDB CLI 1.x+abcdefg
+
+If you encounter a ``command not found`` error, you may need to open a new
+terminal window before the ``edgedb`` command is available.
+
```
i0bs, line 41:
```suggestion

If issues are still present, you may need to check your PATH environment variables to ensure
``edgedb`` is properly pointing to the source directory.

```

This is Windows OS specific, I've ran into this before and people may not realise that the PATH env var is not properly fitted for the terminal. Just a proposal, might not be needed at all to explain.

docs/intro/clients.rst
```diff
@@ -0,0 +1,391 @@
...
+Copy and paste the following simple script. This script initializes a
+``Client`` instance. Clients manage an internal pool of connections to your
+database and provide a set of methods for executing queries.
+
+.. note::
+
+  Note that we aren't passing connection information (say, a connection
+  URL) when creating a client. The client libraries can detect that
+  they are inside a project directory and connect to the project-linked
+  instance automatically. (More on this later.)
```
i0bs, line 136: What does `(More on this later.)` mean? Is it referenced elsewhere in the documentation? If so, we should hyperlink "project-linked instance automatically." If not, will we get this information in the future through something like a blog post? Unsure what this is meant to entail.

docs/intro/clients.rst
```diff
@@ -0,0 +1,391 @@
...
+
+    $ go run .
+
+  .. code-tab:: bash#.NET
+
+    $ dotnet run
+
+You should see a random number get printed to the console. This number was
+generated inside your EdgeDB instance using the built-in ``random()``
+function.
```
i0bs, line 252:
```suggestion
function from Python.
```

These are various languages, developers outside of the Python language will be confused.

docs/intro/migrations.rst
```diff
@@ -0,0 +1,240 @@
...
+
+That's it! You've created and applied your first EdgeDB migration. Your
+instance is now using the latest schema.
+
+
+
+Data migrations
+---------------
+
+Depending on the schema changes, you may be prompted to provide an EdgeQL
```
i0bs, line 131:
```suggestion
Depending on how the schema changes, you may be prompted to provide an EdgeQL
```

i0bs (changes requested):
First pass. Aiming for grammatical consistency, and small tidbits of clarification where it may be appropriate.

docs/intro/clients.rst
colinhacks: This is referring to the `random` function that's built into EdgeQL: https://www.edgedb.com/docs/stdlib/numbers#function::std::random 

I added a link to avoid confusion 👍

docs/intro/clients.rst
colinhacks: Added a ref to the "Connection" section later in the document (which is what I was referring to here). Good call.

docs/guides/deployment/digitalocean.rst
colinhacks: Hm, you're not connecting to the DSN, you're connecting to the instance the DSN refers to. I clarified the wording a bit.

docs/guides/deployment/digitalocean.rst
colinhacks: Good catch

docs/guides/deployment/aws_aurora_ecs.rst
```diff
@@ -26,41 +26,68 @@ Quick Install with CloudFormation
 We maintain a CloudFormation `template <cf-template_>`_ for easy automated
 deployment of EdgeDB in your AWS account.  The template deploys EdgeDB
 to a new ECS service and connects it to a newly provisioned Aurora PostgreSQL
-cluster.  The EdgeDB instance created by the template is exposed to the
-Internet and is protected by TLS and a password you provide.
+cluster. The created instance has a public IP address, has TLS configured, and
```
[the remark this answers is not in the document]
colinhacks: Technically is correct - all three clauses should start with the same part of speech. But it's still weird so I reworked it

docs/datamodel/links.rst
```diff
@@ -90,12 +86,149 @@ In the ``GroupChat`` example, the ``GroupChat.members`` link is now
...
+link`` and exclusivity constraint constitute a *one-to-one* relationship.
+
+.. _ref_guide_many_to_many:
+
+Many-to-many
+^^^^^^^^^^^^
+
+A *many-to-many* relation is the least constrained kind of relationship. There
+is no exclusivity or cardinality constraints in either direction. As an example
+consider a simple app where ``Users`` can "like" their favorite ``Movies``.
```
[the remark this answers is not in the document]
colinhacks: Eh, don't love "multiple of User" so I restructure the sentence to avoid a plural

i0bs (approved):
Glad to see these changes, I think navigation will be way better with this, and of course more clarification on a lot of areas that needed touching

docs/intro/instances.rst
```diff
@@ -0,0 +1,91 @@
...
+We can execute a query against our new instance with ``edgedb query``. Specify
+which instance to connect to by passing an instance name into the ``-I`` flag.
+
+.. code-block:: bash
+
+  $ edgedb query "select 3.14" -I my_instance
+  3.14
+
+Upgrading
+^^^^^^^^^
```
1st1, line 33: I feel like this is a bit too much for "getting started". Especially given that the preferred way is EdgeDB Projects -- maybe the quickstart should only reference `edgedb project` and nothing else. `edgedb instance` is sort of an advanced command that most users don't have to touch.

docs/intro/migrations.rst
```diff
@@ -0,0 +1,240 @@
...
+
+.. code-block::
+
+  .
+  ├── dbschema
+  │   ├── default.esdl
+  │   └── migrations
+  │       └── 00001.edgeql
+  │       └── ...
+  └── edgedb.toml
```
1st1, line 27: Maybe:

```suggestion
  ├── dbschema
  │   ├── default.esdl         // schema definition
  │   └── migrations           // created and managed by EdgeDB
  │       └── 00001.edgeql
  │       └── ...
  └── edgedb.toml
```

docs/intro/migrations.rst
```diff
@@ -0,0 +1,240 @@
...
+
+The schema itself is declaratively defined using EdgeDB's
+schema definition language.
```
1st1, line 30: Since this is a primer I'd say explicitly

```suggestion
schema definition language inside `*.esdl` files.
```

docs/intro/clients.rst
```diff
@@ -0,0 +1,393 @@
+.. _ref_intro_clients:
+
+================
+Client Libraries
+================
```
1st1, line 5: I just love this entire page!

1st1 (commented):
A few comments. I want to do one more pass (a bit more detailed one) tomorrow.

docs/intro/instances.rst
i0bs: +1 with @1st1, in hindsight this would be better.

We could touch up on `edgedb instance` though somewhere else where we may want to provide more advanced methods of handling things through CLI - but for a getting started page, this should be kept as simple as possible.

docs/intro/projects.rst
```diff
@@ -0,0 +1,188 @@
...
+Projects
+========
+
+It can be inconvenient to pass the ``-I`` flag every time you wish to run a
+CLI command.
+
+.. code-block:: bash
+
+  $ edgedb migration create -I my_instance
+
```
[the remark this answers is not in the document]
i0bs: To mitigate the chances of duplicating parts of the docs, I suggest we make a [reference to `client.rst` as seen here.](https://github.com/edgedb/edgedb/pull/4275/files/3bc237a65f17f923d4a469e63a85b0f9d10438d0#diff-b88aa49ad9290d2f942558c54fcc3b8a059306bb0da35f36d15b6d99c5b95b32R139-R149)

docs/edgeql/types.rst
```diff
@@ -150,11 +146,10 @@ The ``[is foo]`` "type intersection" syntax should not be confused with the
   {true}
   db> select {3.14, 2.718} is not int64;
   {true, true}
-  db> select Animal is Dog;
+  db> select Media is Movir;
```
i0bs, line 149: This is a typo in the latest changes

1st1 (approved):

PR #4282, opened 2022-08-20
Fix `range_unpack` bug.

vpetrovykh:
The `range_unpack` function was incorrectly excluding values close to
boundary, especially when the boundary was not itself inclusive.


msullivan (approved):
I think this all makes sense.

One thing that confused me is the lower bound behavior for discrete types like `int64`. It seems that the cause of this is that `int8range` and similar always canonicalize to a `[)` range. Sigh.

Issue #4277, opened 2022-08-19
Reload TLS cerificate files when they change


fmoor:
In many deployment scenarios TLS certificates are renewed on a regular basis. Currently edgedb needs to be restarted when certificate files are renewed. This potentially causes downtime and increased latency as caches are repopulated.

The server should watch the `--tls-cert-file` and `--tls-key-file` and reread them when they change so that the server doesn't need to be restarted.

PR #4290, opened 2022-08-23
Reload TLS certificate files when they change

fmoor:
fixes https://github.com/edgedb/edgedb/issues/4277


edb/server/main.py
```diff
@@ -239,6 +261,15 @@ async def _run_server(
...
             args.tls_cert_file, args.tls_key_file, tls_cert_newly_generated)
 
+        def reload_tls():
+            ss.reload_tls(args.tls_cert_file, args.tls_key_file)
+
+        handler = FileModifiedHandler(reload_tls)
+        observer = watchdog.observers.Observer()
+        observer.schedule(handler, args.tls_cert_file)
+        observer.schedule(handler, args.tls_key_file)
+        observer.start()
```
1st1, line 271: You have to shut it down when we're stopping the server https://python-watchdog.readthedocs.io/en/stable/quickstart.html#a-simple-example

edb/server/main.py
```diff
@@ -69,6 +71,26 @@
...
+        event: Union[
+            watchdog.events.DirModifiedEvent,
+            watchdog.events.FileModifiedEvent,
+        ]
+    ) -> None:
+        if isinstance(event, watchdog.events.DirModifiedEvent):
+            return
+
+        try:
+            self._reload_certs()
```
1st1, line 89: This is incorrect - you shouldn't call methods on Server from an arbitrary thread. You should schdule them with `Server._loop.call_soon_threadsafe`. Also need to shutdown the server if any error happens during reading the files or modifying the tls context.

fmoor: > Also need to shutdown the server if any error happens during reading the files or modifying the tls context.

Is using the original cert not an option if the new one is invalid?

elprans:
Is there an asyncio-compatible watching library we can use instead of spawning threads?

edb/server/main.py
1st1: I have a feeling we should abort in that case, but it's not a strong one. If we decide we want to abort, we should at least retry reading after some time and only abort on a second time. @elprans thoughts?

1st1:
> Is there an asyncio-compatible watching library we can use instead of spawning threads?

No popular and trusted one afaik. Maybe we spawn a thread and just check mtime in a loop every 5 seconds?

elprans:
> > Is there an asyncio-compatible watching library we can use instead of spawning threads?
> 
> No popular and trusted one afaik. Maybe we spawn a thread and just check mtime in a loop every 5 seconds?

Threads mean GIL.  If it can't be in the event loop, then a monitoring subprocess sending SIGHUP to the main process seems better to me.

1st1:
> Threads mean GIL. If it can't be in the event loop, then a monitoring subprocess sending SIGHUP to the main process seems better to me.

That's too complicated.

@fantix how about we add a hidden API to uvloop to expose libuv watchers?

fmoor:
There is https://github.com/samuelcolvin/watchfiles which is mostly an interface to https://github.com/notify-rs/notify but its not 1.0 yet.

1st1:
> There is https://github.com/samuelcolvin/watchfiles which is mostly an interface to https://github.com/notify-rs/notify but its not 1.0 yet.

It also uses threads for its async API.

fantix:
> @fantix how about we add a hidden API to uvloop to expose libuv watchers? That should be easy.

I'm looking into this now - we can start with tweaking MagicStack/uvloop#474, the basic API should be pretty straightforward.

~Even though we don't need the recursive watching in this PR, it'll be needed by the edgedb-python codegen `--watch` mode. However, [libuv doesn't support](https://docs.libuv.org/en/v1.x/fs_event.html#c.uv_fs_event_start) recursive watching out of the box on Linux, and libuv is likely NOT going to [add support for that](https://github.com/libuv/libuv/issues/1778). So IMO we should at least make the hidden uvloop API friendly-enough for implementing recursive watching, given aside of whether we want to support Linux recursive watching directly in uvloop. I'll try something on top of @jensbjorgensen's implementation, and start a new thread in the uvloop repo for further discussion.~

Oh, edgedb-python doesn't depend on uvloop, and `--watch` mode will probably be just an extra feature that doesn't necessarily require uvloop for watching changes recursively. So it'll be easy then, literally quoting Yury's "we add a hidden API to uvloop to (simply) expose libuv watchers".

1st1:
@fantix can you pick this one up given that we have the necessary new API in uvloop?

edb/server/main.py
fantix: I added a fast exponential retry & abort.

edb/server/server.py
```diff
@@ -1765,9 +1764,59 @@ def _tls_private_key_password():
...
+                            reload_tls,
+                            _file_modified,
+                            _event,
+                            retry + 1,
+                        )
+                    )
+            except Exception:
+                logger.critical(
+                    "Unexpected error occured during reload TLS certificates; "
+                    "shutting down.",
```
elprans, line 1807:
```suggestion
                    "error while reloading TLS certificate and/or key, "
                    "shutting down.",
```

edb/server/server.py
```diff
@@ -1765,9 +1764,59 @@ def _tls_private_key_password():
...
+            try:
+                self.reload_tls(tls_cert_file, tls_key_file)
+            except (StartupError, FileNotFoundError) as e:
+                if retry > defines._TLS_CERT_RELOAD_MAX_RETRIES:
+                    logger.critical(str(e))
+                    self._accepting_connections = False
+                    self._stop_evt.set()
+                else:
+                    delay = defines._TLS_CERT_RELOAD_EXP_INTERVAL * 2 ** retry
+                    logger.warning("%s; Retry in %.1f seconds.", e, delay)
```
elprans, line 1794:
```suggestion
                    logger.warning("%s; retrying in %.1f seconds.", e, delay)
```

edb/server/server.py
```diff
@@ -1765,9 +1764,59 @@ def _tls_private_key_password():
...
+                logger.critical(
+                    "Unexpected error occured during reload TLS certificates; "
+                    "shutting down.",
+                    exc_info=True,
+                )
+                self._accepting_connections = False
+                self._stop_evt.set()
+
+        self._tls_certs_watch_handles.append(
+            self.__loop._monitor_fs(str(tls_cert_file), reload_tls)
```
elprans, line 1814: Need to check if `tls_cert_file` and `tls_key_file` are the same file?

elprans:
Closing in favor of the merged #4297.

Issue #4278, opened 2022-08-19
Mechanism to triger config reload without restarting


fmoor:
There should be a way to reload configuration without restarting on `SIGHUP`.

elprans:
No, the server should reload config from external files (e.g. certs) when it receives `SIGHUP`

PR #4297, opened 2022-08-25
Enable config reloading on SIGHUP

fmoor:
Fixes https://github.com/edgedb/edgedb/issues/4278


edb/server/main.py
```diff
@@ -246,6 +246,15 @@ async def _run_server(
...
         )
 
+        def load_configuration(signum, frame):
+            logger.info("reloading configuration")
+            ss.get_loop().call_soon_threadsafe(
+                ss.load_tls, args.tls_cert_file, args.tls_key_file)
+            ss.get_loop().call_soon_threadsafe(
+                ss.load_jwcrypto, args.jws_key_file, args.jwe_key_file)
+
+        signal.signal(signal.SIGHUP, load_configuration)
```
fmoor, line 256: I think this overrides https://github.com/edgedb/edgedb/blob/master/edb/server/daemon/daemon.py#L208. :thinking:

1st1: Yeah, we should call the original handler too

1st1: (to enable log rotation)

fmoor: changed

elprans (approved):

PR #4325, opened 2022-09-07
Annotate all type names when INJECT_OUTPUT_TYPE_NAMES is set

fantix:
This includes type desc annotation for type names of objects and basic scalar types.

### DIFF
--- a/edb/server/compiler/sertypes.py
+++ b/edb/server/compiler/sertypes.py
@@ -397,7 +397,7 @@ class TypeSerializer:
                 buf.append(_uint16_packer(self.uuid_to_pos[el_type]))
 
             if self.inline_typenames:
-                self._add_annotation(mt, type_id.bytes)
+                self._add_typename_annotation(mt, type_id.bytes)
             self._register_type_id(type_id)
             return type_id
 
@@ -436,7 +436,7 @@ class TypeSerializer:
                 buf.append(_uint16_packer(self.uuid_to_pos[bt_id]))
 
             if self.inline_typenames:
-                self._add_annotation(mt)
+                self._add_typename_annotation(mt)
             self._register_type_id(type_id)
             return type_id
 
@@ -502,7 +502,7 @@ class TypeSerializer:
         else:
             return self._describe_type(t, {}, {}, protocol_version)
 
-    def _add_annotation(self, t: s_types.Type, type_id=None):
+    def _add_typename_annotation(self, t: s_types.Type, type_id=None):
         self.anno_buffer.append(CTYPE_ANNO_TYPENAME)
 
         self.anno_buffer.append(type_id or t.id.bytes)
--- a/edb/server/compiler/sertypes.py
+++ b/edb/server/compiler/sertypes.py
@@ -396,6 +396,8 @@ class TypeSerializer:
                 buf.append(el_name_bytes)
                 buf.append(_uint16_packer(self.uuid_to_pos[el_type]))
 
+            if self.inline_typenames:
+                self._add_annotation(mt, type_id.bytes)
             self._register_type_id(type_id)
             return type_id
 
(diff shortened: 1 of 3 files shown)


fantix:
The CLI survived this change, but the notebook tests are broken - I'll double check if edgedb.com could run properly.

1st1:
> The CLI survived this change, but the notebook tests are broken - I'll double check if edgedb.com could run properly.

That's edgedb-js issue, should be easily fixable. cc @jaclarke

edb/server/compiler/sertypes.py
```diff
@@ -396,6 +396,8 @@ def _describe_type(self, t, view_shapes, view_shapes_metadata,
                 buf.append(el_name_bytes)
                 buf.append(_uint16_packer(self.uuid_to_pos[el_type]))
 
+            if self.inline_typenames:
+                self._add_annotation(mt, type_id.bytes)
```
1st1, line 400: Is type_id a name or an id here?

fantix: It's a UUID of the derived object type, we need this to link the annotation back to the original object type descriptor.

1st1: In what cases do you need this?

I don't like to abuse the field that is called "name" to store IDs in it when it's convenient for us.

1st1: I mean we should always just inline plain object names, not IDs. If we need IDs then we need to reconsider how we implement this change (e.g. add a new annotation type).

fantix: Oh I think you misunderstood. As a reminder, the data structure of type descriptors is like the following:

```
uint8 type = 1     # ObjectShapeDescriptor
uuid id = 111...   # Type ID of the object
...                # Data of the 111 ObjectShapeDescriptor

uint8 type = 3     # ScalarTypeDescriptor
uuid id = 222...   # Type ID of the user-defined scalar type
uint16 base_idx =  # Index of the typedesc for the base type

uint8 type = 0xff  # TypeAnnotationDescriptor
uuid id = 111...   # References 111 ObjectShapeDescriptor
string name = ...  # Full type name of 111

uint8 type = 0xff  # TypeAnnotationDescriptor
uuid id = 222...   # References 222 ScalarTypeDescriptor
string name = ...  # Full type name of 222
```

For `ScalarTypeDescriptor`, the `material_type()` is itself, so we could just pass in the `mt` to `_add_annotation()`; but for `ObjectShapeDescriptor`, the current type is derived from the`mt = material_type()`, thus their IDs are different, we still need to pass in the `type_id` of the current type so that the annotation could reference the correct object typedesc.

1st1: Oh my, I'm sorry Fantix. Can we rename `_add_annotation` to `_add_typename_annotation`?

1st1 (approved):
@elprans please also take a look.

elprans:
Is this backwards-compatible?

edb/server/compiler/sertypes.py
fantix: No problem!

fantix:
> Is this backwards-compatible?

@elprans I believe so.

Clients using the binary protocol should ignore unexpected typename annotations, like the JS client looks safe:

https://github.com/edgedb/edgedb-js/blob/93ebc775701ed5274e78b8c24f5675401a477096/src/codecs/registry.ts#L257-L263

1st1:
Closing this one -- after a quick chat with @elprans we decided that this is a suboptimal solution. The selected type can actually be a complex thing (union/intersection/a combination of them) and then the type name would be unusable. Also the return type of `SELECT User {name}` isn't actually `User`, it's separate a derived type that isn't even a subtype of `User`, strictly speaking. A better strategy for `.edgeql` codegens is to just call the top-pevel dataclass/struct type `Data` and then derive the nested object type names from link names.

PR #4338, opened 2022-09-10
Better messages for required links hidden by policies

msullivan:
Use the error message field of assert_exists to put a good message
in. Additionally, track pending computeds being compiled and use those
to produce more detailed messages.
I only list the closest enclosing computed, though we could go further.


1st1:
cc @jaclarke James, please try this locally.

elprans (approved):

PR #4347, opened 2022-09-12
Fix cardinality in CommandDataDescription

fantix:
Previously the cardinality in CommandDataDescription can only be one of AT_MOST_ONE, MANY or NO_RESULT, now it could also be ONE or AT_LEAST_ONE.

### DIFF
--- a/edb/protocol/enums.py
+++ b/edb/protocol/enums.py
@@ -22,6 +22,8 @@ from typing import *
 
 import enum
 
+from edb.edgeql import qltypes as ir
+
 
 class Cardinality(enum.Enum):
     # Cardinality isn't applicable for the query:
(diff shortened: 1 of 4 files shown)


fmoor (approved):

edb/server/compiler/compiler.py
```diff
@@ -626,9 +626,15 @@ def _compile_ql_query(
         )
 
         if ir.cardinality.is_single():
-            result_cardinality = enums.Cardinality.AT_MOST_ONE
+            if ir.cardinality.can_be_zero():
```
elprans, line 629: I would factor this into a classmethod on the enum similarly to how we convert between cardinality representations elsewhere:

```
    @classmethod
    def from_ir_value(
        cls,
        card: ir.Cardinality
    ) -> Cardinality:
        ...
```

then, use the new helper from other such open-coded conversion places (e.g. `cardinality_from_ptr`)

msullivan (approved):
Seems right. Probably follow Elvis's suggestion, though.

edb/protocol/enums.py
```diff
@@ -41,3 +43,18 @@ class Cardinality(enum.Enum):
 
     # Cardinality is >= 1
     AT_LEAST_ONE = 0x4d
+
+    @classmethod
+    def from_ir_value(cls, card: ir.Cardinality) -> Cardinality:
+        if card == ir.Cardinality.AT_MOST_ONE:
```
elprans, line 49: Use `is` to test for enum (https://docs.python.org/3/library/enum.html#comparisons)

```suggestion
        if card is ir.Cardinality.AT_MOST_ONE:
```

elprans (approved):

PR #4353, opened 2022-09-14
Add a _use_sys_pgcon context manager

msullivan:
Makes the code nicer, and I'm planning to add some more uses.


elprans (approved):

1st1 (commented):
Just keep in mind that we have just one system connection. So this high level api should not create an illusion that you can have as many as you want, or you end up with a dead lock. Ideally you should make it non-reentrble

PR #4355, opened 2022-09-14
Implement a patch application mechanism

### LINKED ISSUE #4308: Allow modifying stdlib/metaschema during minor version upgrades
Currently we can't update stdlib functions during minor version upgrades, since they are stored in the database.

We have a plan for fixing this:
1. Add a patch number to the database alongside the catalog version. Store a list of patch scripts (allowing both SQL and EdgeQL) to run on minor version upgrade. When connecting to a database, check the database's patch number and apply any relevant patches.
2. Build some slight tooling for generating patches. Probably a tool that takes a list of function names or similar and generates the scripts. Nothing too smart.
3. Add CI tests that create a database using an earlier point release and upgrades it.

msullivan:
Include, in the source, a list of patches to apply, which can be either SQL or EdgeQL.

Then, in each database, track how many patches have been applied.

On server start, check if the system database has any unapplied patches. If so,
compile all the patches and apply them to every database that is missing them,
including the template database. Update the cached stdschema in the system database,
since that's the only one that matters. (Updating the stdschema is pretty marginal for
fixes to the bodies of functions, but would be important if we wanted to change a volatility,
for example.)

Also fix ALTER of a SQL function, which we'll need for most of these in practice.

Closes #4308.


edb/server/server.py
```diff
@@ -802,6 +808,91 @@ async def _introspect_dbs(self):
...
+                if sys:
+                    sql += syssql
+                logger.info("applying patch %d to database '%s'", num, dbname)
+                sql = tuple(x.encode('utf-8') for x in sql)
+                await conn.sql_fetch(sql)
+
+    async def _maybe_patch_db(self, dbname, patches):
+        logger.info("applying patches to database '%s'", dbname)
+
+        conn = await self.acquire_pgcon(dbname)
```
elprans, line 853: `_use_pgcon`?

edb/server/bootstrap.py
```diff
@@ -466,6 +474,79 @@ def compile_single_query(
...
+        INSERT INTO edgedbinstdata.instdata (key, json)
+        VALUES('num_patches', {val})
+        ON CONFLICT (key)
+        DO UPDATE SET json = {val};
+    """
+
+    # Pure SQL patches are simple
+    if kind == 'sql':
+        return (patch, update), (), schema
+
```
elprans, line 494: `assert` here that `kind` is `edgeql`?

edb/server/server.py
msullivan: these aren't the sys conn; we only use acquire_pgcon like this once, so I haven't added an abstraction

edb/testbase/server.py
```diff
@@ -1370,6 +1370,9 @@ async def _setup_database(dbname, setup_script, conn_args, stats):
         await admin_conn.execute(
             f'CREATE DATABASE {qlquote.quote_ident(dbname)};'
         )
+    except edgedb.DuplicateDatabaseDefinitionError:
```
elprans, line 1373: Why this bit?

msullivan: I want to be able to run the tests on a *database* that was created on and old version

elprans (approved):

edb/testbase/server.py
```diff
@@ -444,7 +444,8 @@ async def init_cluster(
         )
         destroy = False
 
-    if await cluster.get_status() == 'not-initialized':
+    pg_cluster = await cluster._get_pg_cluster()
+    if await pg_cluster.get_status() == 'not-initialized':
```
1st1, line 448: This code is a bit weird -- why didn't `cluster.get_status() == 'not-initialized'` work as expected?

msullivan: Because that `get_status` will try to do a bunch of other things that break in the case where we are trying to reuse an already created database

1st1: Hm, maybe we should refactor that code. Fine for now, but this is brittle.

PR #4366, opened 2022-09-15
Run the primary CI test suite on pushes to stable branches

msullivan:
This needs to be cherry picked to stable/2.x branch to make it take
effect.
Should we do the various other suites too?


elprans (approved):
Other suites (e.g HA) are chained to the main CI, so no need.

PR #4374, opened 2022-09-16
Fix signalctl for Python 3.11

fantix:
Python 3.11 deprecated the `msg` param in `Future.cancel()`, we can no longer depend on the `msg` to mark the reason of cancellations. That means, it is impossible to match recorded signals to CancelledErrors and replace them with SignalErrors.

The solution here is to raise an ExceptionGroup containing all signal errors, and its `__cause__` will be set to the cancellation error chain displaying where the signal hit the user code. User error will be merged into the same ExceptionGroup with signal errors, so be prepared to use the `except*` grammar to handle exceptions.

One exception is CancelledError itself. If `SignalController.wait_for()` is cancelled from outside (and no signals captured after that), a naked CancelledError will still be raised. This made it possible to chain `SignalController.wait_for()` to `asyncio.wait_for()` for timeouts. Similarly, a group of signal errors can still be found in the `__cause__`.

---

UPDATE: the use of ExceptionGroup is dropped because mypy isn't ready yet, now we're chain the signal errors with `__context__`.

### DIFF
--- a/edb/common/signalctl.py
+++ b/edb/common/signalctl.py
@@ -113,20 +113,26 @@ class SignalController:
         if cancel_on is None:
             cancel_on = self._signals
 
+        cancelled_by = []
+        outer_cancelled_at_last = False
+
         # The design here: we'll wait on a separate Future "waiter" for clean
-        # cancellation. The waiter might be waken up by 3 different events:
+        # cancellation. The waiter might be woken up by 3 different events:
         #   1. The given "fut" is done
         #   2. A signal is captured
         #   3. The "waiter" is cancelled by outer code.
-        # For 2 and 3, we'll interrupt the given "fut" by a cancellation
-        # carrying the reason as its message. Because our cancellation might be
+        # For 2, we'll cancel the given "fut" and record the signal in
+        # cancelled_by to raise as an ExceptionGroup in the next step; for 3,
+        # we cancel the given "fut" and propagate the CancelledError later.
+        #
+        # The complexity of this design is: because our cancellation might be
         # intercepted in the "fut" code - e.g. a finally block or except block
         # that traps (and hopefully re-raises) the CancelledError or
         # BaseException, we need a loop here to ensure all the nested blocks
         # are exhaustively executed until the "fut" is done, meanwhile the
         # signals may keep hitting the "fut" code blocks, and "wait_for" is
-        # ready to handle them properly, and chain the SignalError objects
-        # with the __context__ or __cause__ attribute as they happen.
+        # ready to handle them properly, and return all the SignalError objects
+        # in an ExceptionGroup preserving the order as they happen.
         while not fut.done():
             waiter = self._loop.create_future()
             cb = functools.partial(_release_waiter, waiter)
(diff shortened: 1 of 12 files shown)


fantix:
Oh mypy! https://github.com/python/mypy/issues/12840

elprans:
Hm, is there a way to retain Python 3.10 compatibility for now without too much hassle?  If not, then I think we should sit on this until Python 3.11 is out and all dependencies are in order.

fantix:
> is there a way to retain Python 3.10 compatibility for now without too much hassle?

Yeah I think so, instead of using ExceptionGroup, we could just chain the signal errors with `__cause__`, which should be just fine.

edb/common/signalctl.py
```diff
@@ -162,46 +173,42 @@ async def wait_for(self, fut, *, cancel_on=None):
...
+            #    attached as __context__ if event 2 happened.
+            if cancelled_by is not None:
+                e.__context__ = cancelled_by
+                rv = e
             else:
-                # Preserve whatever else is in the chain.
-                current.__context__ = ctx
-            current = current.__context__
-            ctx = ctx.__context__
+                raise
```
1st1, line 211: Should it be bare `raise` or `raise e`? I suspect the former, in which case we can drop line 209

edb/common/signalctl.py
```diff
@@ -162,46 +173,42 @@ async def wait_for(self, fut, *, cancel_on=None):
...
+            # 2. "fut" is cancelled - this usually means we caught a signal,
+            #    but it could also be other reasons, see below.
+            if cancelled_by is not None:
+                # Event 2 happened at least once
+                if outer_cancelled_at_last:
+                    # If event 3 is the last event, the outer code is probably
+                    # expecting a CancelledError, e.g. asyncio.wait_for().
+                    # Therefore, we just raise it with signal errors attached.
+                    ex.__context__ = cancelled_by
+                    rv = ex
```
1st1, line 192: Maybe just do a bare `raise` here?

edb/common/signalctl.py
```diff
@@ -137,16 +143,21 @@ async def wait_for(self, fut, *, cancel_on=None):
             try:
                 try:
                     signal = await waiter
-                except asyncio.CancelledError as e:
+                except asyncio.CancelledError:
                     # Event 3: cancelled by outer code.
                     if not fut.done():
-                        fut.cancel(msg=e)
+                        fut.cancel()
+                        outer_cancelled_at_last = True
```
1st1, line 150: The name is a bit cryptic. Maybe just call it `waiter_was_cancelled`?

fantix: `waiter_was_cancelled` means "event 3 happened at least once", while we need "event 3 is the last event", because without ExceptionGroup, we can only raise one error, and it should be the last one in user-code. For examplye, if an outer cancel - say `asyncio.wait_for()` expired - was handled in a "finally" block in user-code, but we captured a sginal in the "finally" block, we should raise the SignalError instead of the CancelledError, in which case this flag should be `False` even though "waiter was cancelled" once.

On the same page I'm still open to suggestions for naming.

edb/common/signalctl.py
fantix: It should be bare `raise` just because the traceback is one line shorter than `raise e` when `cancelled_by` is None.

If `cancelled_by` is not None, it is almost equivalent to:

```python
        except Exception as e:
            if cancelled_by is not None:
                e.__context__ = cancelled_by
            raise e  # looks like reraising itself won't reset __context__
```

1st1: yeah, so drop line 209

1st1 (approved):

PR #4381, opened 2022-09-19
Generate empty block when necessary.

Androidown:
Currently an empty ALTER TYPE ddl like
```
ALTER TYPE Foo {};
```
will be re-generated by codegen module as
```
ALTER TYPE FOO;
```
, which is invalid syntax.

This is bad because in production it will create a broken migration record with `script` field value as `ALTER TYPE FOO;`, and dump - restore that DB will fail with a syntax error (although I don't know if `include_migration` might be `false`  in a dump/restore process).

### DIFF
--- a/docs/intro/clients.rst
+++ b/docs/intro/clients.rst
@@ -37,6 +37,7 @@ libraries* for the following languages.
 - `Python <https://github.com/edgedb/edgedb-python>`_
 - `Rust <https://github.com/edgedb/edgedb-rust>`_
 - `.NET <https://github.com/edgedb/edgedb-net>`_
+- `Java <https://github.com/edgedb/edgedb-java>`_
 
 Unofficial (community-maintained) libraries are available for the following
 languages.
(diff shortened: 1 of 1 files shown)


msullivan:
The idea here looks reasonable. Could you look at the test failure?

1st1:
@Androidown where is this empty ALTER TYPE coming from? While the fix is correct, I think ultimately EdgeDB should reject this command as it's a no-op.

Androidown:
> @Androidown where is this empty ALTER TYPE coming from? While the fix is correct, I think ultimately EdgeDB should reject this command as it's a no-op.
> 

I agree that such ddl is a no-op and no migration record should be created. But currently you can reproduce this problem with the following DDL
```
CREATE TYPE X;
ALTER TYPE X {};
```
Both are valid edgeql syntax and will run as expect. But if you take a look at table schema::Migration, there will be a record of which the value of field script is `ALTER TYPE X;`.

PR #4394, opened 2022-09-20
Fix computed global scoping behavior

### LINKED ISSUE #4388: links off of global variables don't get semijoined
```
global_semi> create type X;
OK: CREATE TYPE
global_semi> create type Y { create link x -> X };
insert OK: CREATE TYPE
global_semi> insert X;
[{"id": "2008da90-3877-11ed-9bb0-a773f2cf7623"}]
global_semi> insert Y { x := assert_single(X) };
[{"id": "24a14e52-3877-11ed-9bb0-fb0d250ace11"}]
global_semi> insert Y { x := assert_single(X) };
[{"id": "24a14e53-3877-11ed-9bb0-6fdc29f2a593"}]
global_semi> select Y.x;
[{"id": "2008da90-3877-11ed-9bb0-a773f2cf7623"}]
global_semi> create global Yg := Y;
OK: CREATE GLOBAL
global_semi> select (global Yg).x;
[
  {"id": "2008da90-3877-11ed-9bb0-a773f2cf7623"},
  {"id": "2008da90-3877-11ed-9bb0-a773f2cf7623"}
]
```

msullivan:
Computed globals should not participate in path factoring, since
`(global Foo).x` looks much more like `(select Foo).x` or
`(detached Foo).x` than `Foo.x`.

Additionally, fix links off of computed globals not being deduplicated.

Fixes #4388.

### DIFF
--- a/edb/edgeql/compiler/expr.py
+++ b/edb/edgeql/compiler/expr.py
@@ -471,7 +471,10 @@ def compile_GlobalExpr(
     if glob.is_computable(ctx.env.schema):
         obj_ref = s_utils.name_to_ast_ref(
             glob.get_target(ctx.env.schema).get_name(ctx.env.schema))
-        return dispatch.compile(qlast.Path(steps=[obj_ref]), ctx=ctx)
+        # Wrap the reference in a subquery so that it does not get
+        # factored out or go directly into the scope tree.
+        qry = qlast.SelectQuery(result=qlast.Path(steps=[obj_ref]))
+        return dispatch.compile(qry, ctx=ctx)
 
     objctx = ctx.env.options.schema_object_context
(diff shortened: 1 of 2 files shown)


elprans (approved):
Let's document this also?

msullivan:
Do we have reasonable documentation or explanation of our longest path factoring anywhere? I only know of https://www.edgedb.com/docs/reference/edgeql/eval, which specifies it but kind of assumes the reader has some idea what is going on already.

elprans:
I think we should have a reasonable simple explanation of factoring and scopes somewhere high enough. @colinhacks do you have ideas on how we can better present scoping?   Maybe even with a picture?

PR #4397, opened 2022-09-20
Clean up some policy stuff

msullivan:
Always compile policies inside nested scopes, to avoid dumping stuff
at the top level.
Properly infer policy cardinality for INSERT. (Right now it is always
ONE, but that might change, and the backend was treating the missing
cardinality as MANY)

### DIFF
--- a/edb/edgeql/compiler/inference/cardinality.py
+++ b/edb/edgeql/compiler/inference/cardinality.py
@@ -1094,6 +1094,20 @@ def _infer_matset_cardinality(
         )
 
 
+def _infer_policy_cardinality(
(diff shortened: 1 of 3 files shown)


elprans (approved):
Is this fixing a bug or needs a test?

msullivan:
> Is this fixing a bug or needs a test?

Not really. I wouldn't be *surprised* if the scoping stuff fixes a bug, but I wasn't able to think of it. The card inference fix improves the generated code quality by avoiding using the MANY path for filter clauses spuriously. I added some asserts to check that just now, in liu of adding a codegen test.

Issue #4335, opened 2022-09-08
Slicing empty array produces empty set


msullivan:
```
edgedb> select (<array<str>>[])[0:];
{}
```
Correct behavior would be to produce `{[]}`.

Strings and bytes work correct. json strings work but json arrays are broken.

PR #4403, opened 2022-09-22
Fix slicing with an empty set

### LINKED ISSUE #4327: Slicing behaves incorrectly on NULL empty sets
Another instance of a classic edgedb bug.

Slicing is a regular (non-optional) operation, and so empty set inputs should result in an empty set output, like such:
```
edgedb> select [1,2,3][1:<int64>{}];
[]
```

But empty sets that get represented as NULL produce output as if the argument was omitted:
```
edgedb> select [1,2,3][1:<optional int64>$0];
Parameter <int64>$0 (Ctrl+D for empty set `{}`): 
[[2, 3]]
```

The root of the issue is that all slicing (missing arguments or not) is done by calling a `_slice` function, and we represent missing bounds as NULL.

aljazerzen:
fixes #4327

### DIFF
--- a/edb/pgsql/compiler/expr.py
+++ b/edb/pgsql/compiler/expr.py
@@ -286,9 +286,11 @@ def compile_SliceIndirection(
     with ctx.new() as subctx:
 
         # Postgres arrays use 32bit indexes, so EdgeDB inherits this limitation
-        def compile_and_cast_int(expr):
-            expr = dispatch.compile(expr, ctx=subctx)
-            return pgast.TypeCast(arg=expr, type_name=pgast.TypeName(name=("integer",)))
+        def compile_and_cast_int(expr: irast.Base) -> pgast.BaseExpr:
+            pg_expr = dispatch.compile(expr, ctx=subctx)
+            return pgast.TypeCast(
+                arg=pg_expr, type_name=pgast.TypeName(name=("integer",))
+            )
 
         subctx.expr_exposed = False
         subj = dispatch.compile(expr.expr, ctx=subctx)
@@ -299,14 +301,15 @@ def compile_SliceIndirection(
             start = compile_and_cast_int(expr.start)
 
         if expr.stop is None:
-            # Max index in EdgeQL is 2^32-2 i.e. 2147483646
-            # because during conversion to Postgres index, we 
-            # add 1 to the stop and then subtract one.
-            stop: pgast.BaseExpr = pgast.LiteralExpr(expr="2147483646")
+            # Max index in EdgeQL is 2^32-2 because during conversion
+            # to Postgres index, we add 1 to the stop and then subtract one.
+            stop: pgast.BaseExpr = pgast.LiteralExpr(expr=str(2**31-2))
         else:
             stop = compile_and_cast_int(expr.stop)
 
-    result = pgast.FuncCall(name=("edgedb", "_slice"), args=[subj, start, stop])
+    result = pgast.FuncCall(
+        name=("edgedb", "_slice"), args=[subj, start, stop]
+    )
     return result
 
 
--- a/edb/pgsql/compiler/expr.py
+++ b/edb/pgsql/compiler/expr.py
@@ -276,29 +276,37 @@ def compile_IndexIndirection(
 
 @dispatch.compile.register(irast.SliceIndirection)
 def compile_SliceIndirection(
-        expr: irast.SliceIndirection, *,
-        ctx: context.CompilerContextLevel) -> pgast.BaseExpr:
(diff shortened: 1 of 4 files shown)


aljazerzen:
Also fixes #4335

edb/pgsql/compiler/expr.py
```diff
@@ -276,29 +276,37 @@ def compile_IndexIndirection(
...
+) -> pgast.BaseExpr:
+
     # Handle Expr[Index], where Expr may be std::str, array<T> or
     # std::json. For strings we translate this into substr calls.
     # Arrays use the native slice syntax. JSON is handled by a
     # combination of unnesting aggregation and array slicing.
     with ctx.new() as subctx:
+
+        # Postgres arrays use 32bit indexes, so EdgeDB inherits this limitation
+        def compile_and_cast_int(expr):
```
msullivan, line 289: You'll need to put a type annotation on this for mypy to pass.
(mypy is pretty fast, and useful, so you'll probably want to run that locally even during development)

edb/pgsql/compiler/expr.py
```diff
@@ -276,29 +276,37 @@ def compile_IndexIndirection(
...
         else:
-            start = dispatch.compile(expr.start, ctx=subctx)
+            start = compile_and_cast_int(expr.start)
+
         if expr.stop is None:
-            stop: pgast.BaseExpr = pgast.NullConstant()
+            # Max index in EdgeQL is 2^32-2 i.e. 2147483646
+            # because during conversion to Postgres index, we 
+            # add 1 to the stop and then subtract one.
+            stop: pgast.BaseExpr = pgast.LiteralExpr(expr="2147483646")
```
msullivan, line 305: I'd compute the number instead of just writing it in

edb/pgsql/metaschema.py
```diff
@@ -2094,39 +2081,22 @@ class StringSliceImplFunction(dbops.Function):
...
+                edgedb._normalize_array_index(
+                    start, edgedb._length(val))
+            )
     '''
 
     def __init__(self) -> None:
         super().__init__(
             name=('edgedb', '_str_slice'),
             args=[
-                ('val', ('anyelement',)),
```
msullivan, line 2128: I would leave these arguments one per line. It all fits on one, but especially with this sort of nested data I think the line breaks help.

Feel free to add trailing commas, which should calm down your code formatter?

edb/pgsql/metaschema.py
```diff
@@ -2296,18 +2264,19 @@ class JSONSliceFunction(dbops.Function):
...
         SELECT
-            CASE jsonb_typeof(val)
-            WHEN 'array' THEN (
-                to_jsonb(_slice(
+            CASE
+            WHEN val IS NULL THEN NULL
+            WHEN jsonb_typeof(val) = 'array' THEN (
+                to_jsonb(edgedb._slice(
                     (
-                        SELECT array_agg(value)
```
msullivan, line 2303: Oops.

tests/test_edgeql_select.py
```diff
@@ -5171,6 +5171,98 @@ async def test_edgeql_select_slice_03(self):
...
+
+        await self.assert_query_result(
+            r"""
+            select [1,2,3][1:<int64>{}];
+            """,
+            [],
+        )
+
+        # try to trick the compiler and to pass NULL into edgedb._slice
+        await self.assert_query_result(
```
msullivan, line 5198: Can you test with the bottom bound as a NULL input too? And with both, I guess.

msullivan (commented):
This looks good! I left a few comments, but it should be good to merge once those are addressed!

msullivan (approved):
The remaining test failure is legitimate, since json values come back as strings in our native python bindings and as json in our json mode. You can fix this by making a direct call to query (like you do in other places), but `assert_query_result` allows specifying separate expected values for json and native modes.

Feel free to merge once the tests pass

edb/pgsql/compiler/expr.py
```diff
@@ -276,29 +276,40 @@ def compile_IndexIndirection(
...
+    expr: irast.SliceIndirection, *, ctx: context.CompilerContextLevel
+) -> pgast.BaseExpr:
+
     # Handle Expr[Index], where Expr may be std::str, array<T> or
     # std::json. For strings we translate this into substr calls.
     # Arrays use the native slice syntax. JSON is handled by a
     # combination of unnesting aggregation and array slicing.
     with ctx.new() as subctx:
+
+        # Postgres arrays use 32bit indexes, so EdgeDB inherits this limitation
```
elprans, line 288: Ideally we should document this, because our `std::[]` is confusingly defined as taking `int64`

PR #4415, opened 2022-09-23
Propagate database creation and deletion events to adjacent servers

elprans:
Make sure `CREATE DATABASE` and `DROP DATABASE` gets noticed properly by
adjacent servers.

This doesn't fully solve the `DROP DATABASE` case, as adjacent servers
might be maintaining inactive connections in their pools, this
preventing the database from getting dropped.  This is solvable, but is
somewhat messy, so I left this for a future patch.

### DIFF
--- a/tests/test_server_proto.py
+++ b/tests/test_server_proto.py
@@ -2178,6 +2178,40 @@ class TestServerProtoDdlPropagation(tb.QueryTestCase):
                             DROP ROLE ddlprop01;
                         ''')
 
+    @unittest.skipUnless(devmode.is_in_dev_mode(),
+                         'the test requires devmode')
+    async def test_server_adjacent_database_propagation(self):
+        if not self.has_create_database:
+            self.skipTest('create database is not supported by the backend')
+
+        conargs = self.get_connect_args()
+
+        server_args = {}
+        if self.backend_dsn:
+            server_args['backend_dsn'] = self.backend_dsn
+        else:
+            server_args['adjacent_to'] = self.con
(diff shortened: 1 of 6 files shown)


tests/test_server_proto.py
```diff
@@ -2178,6 +2178,40 @@ async def test_server_proto_ddlprop_01(self):
...
+        server_args = {}
+        if self.backend_dsn:
+            server_args['backend_dsn'] = self.backend_dsn
+        else:
+            server_args['adjacent_to'] = self.con
+        async with tb.start_edgedb_server(**server_args) as sd:
+
+            await self.con.execute('''
+                CREATE DATABASE test_db_prop;
+            ''')
```
1st1, line 2198: Can you also add a test for the "a database changes" case?

1st1 (approved):

tests/test_server_proto.py
elprans: Not sure what you mean by "a database changes" here

1st1: From code:

```
        # Triggered by a postgres notification event 'database-changes'
        # on the __edgedb_sysevent__ channel
```

elprans: Well, that's what this test does

1st1: OK, I got confused with the naming and by the presence of `_on_remote_database_deletion` -- what's it for?

elprans: Dead code.

PR #4417, opened 2022-09-23
Fix slicing arrays of tuples

aljazerzen:
- change `irast.SliceIndirection#expr` from `irast.Base` to `irast.Set`
- inline `edgedb._slice` call for arrays of tuples

We could actually inline for all arrays, now that the code is written. There would be more code generated, but could reduce number of functions in stdlib by 1.


msullivan (approved):
Looks great!

PR #4418, opened 2022-09-23
Improve invalid reference error

### LINKED ISSUE #915: Can we improve `type or view does not exist` message?
Here is the query:
```
SELECT User { id } FILTER email = <str>$email
```
And the error is:
```
edgedb.errors.InvalidReferenceError: object type or view 'email' does not exist
```

My points are:
1. It would be quite common error (as it closely matches SQL `WHERE email =`)
2. It's hard to understand what `object type or view` means in this context.

The best fix would be to look through all the names and make a suggestion:
```
edgedb.errors.InvalidReferenceError: object type or view 'email' does not exist. Try `User.email`?
```

If it's hard to implement we can use some heuristics (like 99% that lowercase name is a field). Or at least add an explanation `You can't reference field directly use 'Object.field'`.

aljazerzen:
Closes #915

image

The heuristic with all lower case is probably not 100%. I also hope this doesn't have many false positives.

### DIFF
--- a/edb/edgeql/compiler/schemactx.py
+++ b/edb/edgeql/compiler/schemactx.py
@@ -90,7 +90,6 @@ def get_schema_object(
         )
 
     except errors.QueryError as e:
-        print(item_type)
         s_utils.enrich_schema_lookup_error(
             e,
             name,
--- a/edb/edgeql/compiler/schemactx.py
+++ b/edb/edgeql/compiler/schemactx.py
@@ -97,6 +97,7 @@ def get_schema_object(
             item_type=item_type,
             condition=condition,
             context=srcctx,
+            compiler_ctx=ctx,
         )
         raise
 
--- a/edb/edgeql/compiler/schemactx.py
+++ b/edb/edgeql/compiler/schemactx.py
@@ -27,6 +27,7 @@ from typing import *
 from edb import errors
 
 from edb.common import parsing
+from edb.ir import typeutils
 
 from edb.schema import links as s_links
 from edb.schema import name as sn
@@ -89,15 +90,16 @@ def get_schema_object(
         )
 
     except errors.QueryError as e:
+        print(item_type)
         s_utils.enrich_schema_lookup_error(
             e,
             name,
             modaliases=ctx.modaliases,
             schema=ctx.env.schema,
             item_type=item_type,
+            pointer_parent=_get_partial_path_prefix_type(ctx),
             condition=condition,
             context=srcctx,
-            compiler_ctx=ctx,
         )
         raise
 
@@ -112,6 +114,19 @@ def get_schema_object(
             context=srcctx)
     else:
         return stype
+
+
+def _get_partial_path_prefix_type(
+    ctx: context.ContextLevel,
+) -> Optional[s_types.Type]:
+    if ctx is None:
+        return None
+    ppp = ctx.partial_path_prefix
+    if ppp is None or ppp.typeref is None:
+        return None
+
+    _, type = typeutils.ir_typeref_to_type(ctx.env.schema, ppp.typeref)
+    return type
 
 
 def _get_type_variant(
--- a/edb/schema/schema.py
+++ b/edb/schema/schema.py
@@ -1353,6 +1353,17 @@ class FlatSchema(Schema):
     ) -> NoReturn:
         refname = str(name)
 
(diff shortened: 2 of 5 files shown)


edb/schema/schema.py
```diff
@@ -1353,6 +1353,17 @@ def _raise_bad_reference(
...
+        # some hacky heuristics
+        may_be_field = (
+            label == 'object type or alias' and
+            refname.islower() and 
+            '::' not in refname
+        )
+
+        hint: Optional[str] = None
+        if may_be_field:
+            hint = f'did you mean entity field `.{refname}`?'
```
elprans, line 1365:
```suggestion
            hint = f'did you mean object property or link `.{refname}`?'
```

elprans (commented):
We should be able to make this more precise by inspecting the pointers of current path prefix (in context), possibly with Levenshtein distance

edb/schema/utils.py
```diff
@@ -818,30 +817,93 @@ def find_item_suggestions(
...
+    if compiler_ctx is None:
+        return iter([])
+    ppp = compiler_ctx.partial_path_prefix
+    if ppp is None or ppp.typeref is None:
+        return iter([])
+
+    schema, object = typeutils.ir_typeref_to_type(schema, ppp.typeref)
+
+    # TODO: is it okay to use all referrers?
+    suggestions = schema.get_referrers(object)
```
aljazerzen, line 859: @msullivan Is this okay?

aljazerzen:
I've redone this to use existing `enrich_schema_lookup_error` and extend it to also suggest `.link` or `.property`:

image

@msullivan Where should I put tests for this? Does it even need them?

msullivan:
(We chatted some directly about next steps on this)

aljazerzen:
@msullivan  I've changed lookup of pointers and added two tests. Feel free to merge when (or if) the tests pass.

edb/schema/utils.py
```diff
@@ -818,30 +817,81 @@ def find_item_suggestions(
...
+def find_fields_suggestions(
+    schema: s_schema.Schema,
+    item_type: Optional[so.ObjectMeta],
+    compiler_ctx: Optional[ContextLevel],
+) -> Iterable[Tuple[so.Object, str]]:
+    from . import types as s_types
+    from . import objtypes as s_objtypes
+    from edb.ir import typeutils
+
+    if item_type is not s_types.QualifiedType:
```
msullivan, line 845: I don't think this `item_type` stuff should be needed?

edb/schema/utils.py
```diff
@@ -37,6 +37,7 @@
     from . import schema as s_schema
     from . import types as s_types
     from edb.common import parsing
+    from edb.edgeql.compiler.context import ContextLevel
```
msullivan, line 40: I should have thought of this when we talked earlier, but I'd like to try to avoid having schema utils operate on pretty internal edgeql->IR compiler stuff like the context. I think it would be best to do the `context -> ObjectType` resolution in a helper in the edgeql compiler somewhere (`edb.edgeql.compiler.schemactx`, next to the enrich callsite, probably), and just pass in an `ObjectType`

msullivan: Oh, actually, looking a little closer, `enrich_...` already has an argument `collection`, which takes a collection of objects to consider instead of looking it up, and it's used for suggesting pointers when the dot *has* been included. 
Look into whether unifying those cases makes things nicer? (It might not)
(Which could probably be done by passing in a collection in both places, or just passing in the source in both places?)

aljazerzen: I've refactored the code so instead of `collection`, `pointer_parent: so.Object` is passed as an argument. We still need separate function for finding pointers because they should not be type checked, because missing dot is causing wrong type to be expected.

edb/schema/utils.py
aljazerzen: Yup, removed that.

edb/edgeql/compiler/schemactx.py
```diff
@@ -89,12 +90,14 @@ def get_schema_object(
         )
 
     except errors.QueryError as e:
+        print(item_type)
```
msullivan, line 93: Stray print

msullivan (approved):
Looks good to me.
Clean up the stray prints and merge!

PR #4422, opened 2022-09-23
Propagate database quarantine before dropping it

elprans:
Even though there may not be active connections to a database being
dropped, adjacent servers may still be keeping warm Postgres connections in
the pool, so use signaling to trigger remote pruning.

### DIFF
--- a/edb/common/retryloop.py
+++ b/edb/common/retryloop.py
@@ -1,0 +1,141 @@
+#
+# This source file is part of the EdgeDB open source project.
+#
+# Copyright 2022-present MagicStack Inc. and the EdgeDB authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+from __future__ import annotations
+from typing import (
+    Callable,
+    Optional,
+    Tuple,
+    Type,
+)
+
+import asyncio
+import random
+import time
+import types
+
+
+def const_backoff(delay: float) -> Callable[[int], float]:
+    return lambda _: delay
+
+
+def exp_backoff(
+    *,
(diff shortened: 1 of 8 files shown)


edb/server/server.py
```diff
@@ -1262,6 +1305,18 @@ async def _signal_sysevent(self, event, **kwargs):
...
 
+    def _on_remote_database_quarantine(self, dbname):
+        if not self._accept_new_tasks:
+            return
+
+        # Block new connections to the database.
+        self._block_new_connections.add(dbname)
+
+        async def task():
+            await self._pg_pool.prune_inactive_connections(dbname)
```
msullivan, line 1316: Can you just pass self._pg_pool.prune_inactive_connections(dbname) to `create_task`?

tests/test_server_proto.py
```diff
@@ -2212,6 +2212,10 @@ async def test_server_adjacent_database_propagation(self):
                     await con2.query("select 1")
                     await con2.aclose()
 
+            await self.con.execute('''
+                DROP DATABASE test_db_prop;
+            ''')
```
msullivan, line 2217: Should we test *recreating* a database with the same name, after a drop?

msullivan: (And connecting to the new one, I guess)

msullivan (approved):
Looks good, though a few questions

edb/common/retryloop.py
```diff
@@ -0,0 +1,141 @@
...
+
+class RetryIteration:
+
+    def __init__(self, loop: RetryLoop) -> None:
+        self._loop = loop
+
+    async def __aenter__(self) -> RetryIteration:
+        return self
+
+    async def __aexit__(
```
1st1, line 95: Around here I started to realize that this might be my code... :)

tests/test_server_proto.py
elprans: Done.

edb/server/server.py
```diff
@@ -1089,21 +1093,60 @@ async def _on_before_create_db_from_template(
...
         else:
-            # If, however, there are no open EdgeDB connections, prune
-            # all non-active postgres connection to the `dbname` DB.
+            # Block new connections to the database.
+            self._block_new_connections.add(dbname)
+
+            # Prune our inactive connections.
             await self._pg_pool.prune_inactive_connections(dbname)
 
+            # Signal adjacent servers to prune their connections to this
```
1st1, line 1109: Can you add a fast path here awaiting on `_pg_ensure_database_not_connected`?

1st1 (approved):

PR #4425, opened 2022-09-24
Make sure connections used by patching don't linger

elprans:
Current approach is rather heavy-handed and not very reliable

### DIFF
--- a/edb/server/connpool/pool.py
+++ b/edb/server/connpool/pool.py
@@ -1194,6 +1194,12 @@ class Pool(BasePool[C]):
         while (conn := block.try_steal()) is not None:
             conns.append(conn)
 
+        # Forcibly drop the block, so that the cleared out block can't
+        # have a connection transferred to it because it appears
+        # starving.
+        assert not block.count_waiters()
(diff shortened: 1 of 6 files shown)


edb/server/connpool/pool.py
```diff
@@ -1194,6 +1194,12 @@ async def prune_inactive_connections(self, dbname: str) -> None:
         while (conn := block.try_steal()) is not None:
             conns.append(conn)
 
+        # Forcibly drop the block, so that the cleared out block can't
+        # have a connection transferred to it because it appears
+        # starving.
+        assert not block.count_waiters()
+        self._blocks.pop(block.dbname)
```
elprans, line 1201: I'm afraid this is not correct, since popping the block does not actually cancel any in-flight connection transfers or new connections.  Also, I think we need to add an explicit method, like `evict_db` or something, because `prune_inactive_connections` does not necessarily imply that _all_ connections are dropped, just the inactive ones.

I've got a better solution coming up.

elprans:
Closing in favor of #4426

PR #4426, opened 2022-09-24
[try 2] Make sure connections used by patching don't linger

msullivan:
Current approach is rather heavy-handed and not very reliable

This is a retry of https://github.com/edgedb/edgedb/pull/4425 with a different branch name


msullivan:
Actually doing it like this is silly

PR #4429, opened 2022-09-24
Use direct connections instead of the connection pool when patching

elprans:
Using the regular connection pool unnecessarily pollinates the pool with
connections to databases, most importantly to the template database,
which should not maintain any connections.  Instead of fighting the
pool, connect to Postgres directly while patching.


msullivan (approved):
Yeah, this makes more sense

PR #4432, opened 2022-09-26
Only run ANALYZE on the conn in bootstrap when it exists

msullivan:


edb/server/bootstrap.py
```diff
@@ -1659,9 +1659,10 @@ async def _bootstrap(ctx: BootstrapContext) -> None:
         schema = s_schema.FlatSchema()
         schema = await _init_defaults(schema, compiler, tpl_ctx.conn)
 
-        await conn.sql_execute(
-            b"ANALYZE",
-        )
+        if backend_params.has_create_database:
+            await conn.sql_execute(
```
elprans, line 1663: Use `tpl_ctx.conn`, we still want `ANALYZE` on single-db backends.

elprans (approved):

PR #4433, opened 2022-09-26
Skip redundant prompts in digitalocean guide

fmoor:
Omitting the `--non-interactive` argument results in the CLI prompting for all connection options even if the DSN specifies all necessary options.


fmoor:
see also https://github.com/edgedb/edgedb-cli/issues/844

elprans (approved):

PR #4448, opened 2022-09-29
Handle bigint in index and slice

### LINKED ISSUE #4416: Decide on types of array indexing
As noted in #4403, Postgres supports only 32 bit integers and does an implicit cast from `bigint` into `integer` before applying `[]`.

Our `std::[]`, on the other hand, uses `bigint`.

Do we inherit this limitation, or try to work around it?

My opinion is that the limit is reasonable, with `bytea`, this allows up to 2GB. EdgeDB is (probably) not meant for such storage. Similarly, if arrays have more than 2.147.483.648 elements, the user is doing something wrong.

If we decide to inherit, we should:
- [x] correct `std::[]` for arrays, text, bytes and 
- [x] correct `std::[l:u]` (slicing),
- [ ] add some kind of warning in the docs,
- [x] correctly report index out of range errors.

aljazerzen:
closes #4416

### DIFF
--- a/edb/pgsql/metaschema.py
+++ b/edb/pgsql/metaschema.py
@@ -1904,31 +1904,79 @@ class ParseTriggerConditionFunction(dbops.Function):
 
 class NormalizeArrayIndexFunction(dbops.Function):
     """Convert an EdgeQL index to SQL index."""
+
     text = '''
-        SELECT (
+        SELECT
             CASE WHEN index < 0 THEN
-                length + index + 1
+                length::bigint + index + 1
             ELSE
                 index + 1
             END
-        )::int
     '''
 
     def __init__(self) -> None:
         super().__init__(
             name=('edgedb', '_normalize_array_index'),
             args=[('index', ('bigint',)), ('length', ('int',))],
-            returns=('int',),
+            returns=('bigint',),
             volatility='immutable',
+            strict=True,
+            text=self.text,
+        )
+
+
+class IntOrNullFunction(dbops.Function):
+    """
+    Convert bigint to int. If it does not fit, return NULL.
+    """
+
+    text = """
+        SELECT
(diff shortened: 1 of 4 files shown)


aljazerzen:
I've tested the performance impact of added code in stdlib with:

```
for i in _gen_series(1, 100000000)
union ( select [1, 2, 3][0:i]);
```

and 

```
for i in _gen_series(1, 100000000)
union (select [1, 2, 3][i]);
```

The impact is negligent.

When this is merged, I'll also add patch commit to `stable/2.x`.

edb/pgsql/metaschema.py
```diff
@@ -2099,22 +2168,26 @@ class StringSliceImplFunction(dbops.Function):
...
-                edgedb._normalize_array_index(
-                    start, edgedb._length(val))
+                pg_start,
+                pg_end - pg_start
             )
+        FROM (SELECT
+            GREATEST(0, edgedb._int_bound(edgedb._normalize_array_index(
+                start, edgedb._length(val)
+            ))) as pg_start,
+            GREATEST(0, edgedb._int_bound(edgedb._normalize_array_index(
```
msullivan, line 2178: Why is `GREATEST(` needed here?

msullivan (approved):
This looks good to me. This whole thing is kind of a pain, but oh well. 

Adding a doc warning is probably worth doing. Doesn't need to be this PR.

I would not bother backporting this to `2.x`.

edb/pgsql/metaschema.py
aljazerzen: Even though both bounds are int, when you subtract them to get length, it may overflow. 

This can happen, for example if you use `[1, 2, 3][-2147483648:12]`. Start will get converted into `-2147483648 + 3 + 1 = -2147483644`. Length would be `12 - (-2147483644) = overflow`.

This was my afternoon today.

PR #4454, opened 2022-09-29
Fix creating an enum and a variadic function that references at the same time

### LINKED ISSUE #4213: Custom type in function variadic argument is not found in migration
- EdgeDB Version:  2.0+88c1706, 2.1+52c90a7

This fails:

```
edgedb> start migration to {
.......     module default {
.......         scalar type E extending enum<a, b, c>;
.......         function f(variadic e: E) -> bool using (true);
.......     }
....... };
OK: START MIGRATION
OK: POPULATE MIGRATION
edgedb[tx]> commit migration;
edgedb error: InvalidReferenceError: type 'default::E' does not exist
```

But doing it separately succeeds:

```
edgedb> start migration to {
.......     module default {
.......         scalar type E extending enum<a, b, c>;
.......
.......     }
....... };
OK: START MIGRATION
OK: POPULATE MIGRATION
edgedb[tx]> commit migration;
OK: COMMIT MIGRATION
edgedb> start migration to {
.......     module default {
.......         scalar type E extending enum<a, b, c>;
.......         function f(variadic e: E) -> bool using (true);
.......     }
....... };
OK: START MIGRATION
OK: POPULATE MIGRATION
edgedb[tx]> commit migration;
OK: COMMIT MIGRATION
edgedb> \ds
module default {
    function f(variadic e: default::E) ->  std::bool using (true);
    scalar type E extending enum<a, b, c>;
};
```

msullivan:
The issue here is that function name mangling currently uses the *id*
of collection types. Collection type ids are deterministic and depend
on the ids of their components, but we need to generate the name
before the enum is created.

Just use the name.

This can't be cherry-picked to 2.0 because it changes the names of
lots of objects.

Fixes #4213.

### DIFF
--- a/edb/buildmeta.py
+++ b/edb/buildmeta.py
@@ -44,7 +44,7 @@ from edb.common import verutils
 
 
 # Increment this whenever the database layout or stdlib changes.
-EDGEDB_CATALOG_VERSION = 2022_09_29_00_00
+EDGEDB_CATALOG_VERSION = 2022_09_29_00_01
 EDGEDB_MAJOR_VERSION = 3
 
 
--- a/edb/schema/functions.py
+++ b/edb/schema/functions.py
@@ -858,12 +858,8 @@ class CallableObject(
         canonical_order = canonical_param_sort(schema, params)
         for param in canonical_order:
             pt = param.get_type_shell(schema)
-            if isinstance(pt, s_types.CollectionTypeShell):
-                quals.append(pt.get_schema_class_displayname())
-                pt_id = str(pt.get_id(schema))
-            else:
-                pt_id = str(pt.name)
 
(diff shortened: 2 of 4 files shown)


elprans:
Hm, is this safe in the face of a possible rename of collection type components?

msullivan:
Hm, I think it *should* be as safe as it is in the face of renames of non-collection types. (Which I'm pretty sure it is.) I'll double check.

(I also now have a pending patch that will just nuke the entire `get_id` mechanism for collection types and use generated ids and look things up by name like we do for everything else.)

msullivan:
I added a test that validates that it is safe in the face of that.

elprans (approved):

Issue #2383, opened 2021-03-18
ISE when doing UPDATE on result of union


msullivan:
With a schema like
```
type Bar { property x -> str; }
type Baz { property x -> str; }
```
If we do `WITH W := (Bar UNION Baz) UPDATE W SET { x := "1" }`, we get an ISE from pgsql about a column not existing.
If we make x a multi property, instead we get an "unexpected union link" out of `range_for_ptrref`.

I don't think it would be terrible if we rejected this (though it also may not be hard to support).

elprans:
Should be easy to support `Bar UNION Baz` is essentially equivalent to `Bar extending Baz`, we just need to generate DMLParts from `union_components` too.

msullivan:
Also, as a note, we'll need to make sure that we correctly find any constraints we need to generate conflict checks for

msullivan:
Huh, this seems to have been partially fixed at some point. It is broken when we do typename injection, but works without it.

This is actually a *problem*, though, because since it got fixed accidentally, the "correctly generate conflict checks" thing that I mentioned above didn't happen.

PR #4463, opened 2022-09-30
Get rid of deterministic collection id generation

msullivan:
It is a weird mismatch with the rest of the system, and introduces
scope for weird collision attacks if untrusted users to a system can
control explicitly inserted ids.


edb/api/types.txt
```diff
@@ -47,4 +47,13 @@
 00000000-0000-0000-0000-000000000111 cal::relative_duration
 00000000-0000-0000-0000-000000000112 cal::date_duration
 
+00000000-0000-0000-0000-000000000120 range<std|int32>
```
elprans, line 50: Hmm, what's the reason for static allocation for ranges?  We don't do this for arrays or tuples.

msullivan: Yeah, but in terms of how they map onto postgres, ranges are more like scalar types, and so they share scalar type's mechanism for mapping uuids to postgres type names

1st1: This smells a bit. Can you have ranges of extension types, like PostGIS scalars? Also why does this line has this weird `std|int32` syntax?

msullivan: I agree it smells a bit, but I think it's largely making explicit an existing implicit smell?

I think if you have ranges of extension types, it would need to have its own postgres type name?

Because those are the internal type names for range types, since we escape `::` into `|` inside the type argument part of names.

elprans: I still think there's probably no real need for static allocation here and we are simply missing a bit of indirection somewhere.  Also, IIRC static id allocation was introduced as an optimization so that clients can have pre-built codec tables, i.e. technically stable type ids are not a requirement even for built-in types.

elprans: Pushed https://github.com/edgedb/edgedb/pull/4463/commits/2012b8f0e6bfc09160c9f7e70cb2f24e588a6d69 here removing static allocation from ranges and treating them like arrays instead.

msullivan:
Oh, yeah, that is much better. Thanks!

edb/server/bootstrap.py
```diff
@@ -1050,7 +1050,7 @@ async def _init_stdlib(
             compiler,
             stdlib.reflschema,
             '''
-            UPDATE schema::Array
```
msullivan, line 1053: Huh, apparently https://github.com/edgedb/edgedb/issues/2383 got fixed somehow.

elprans (approved):

PR #4469, opened 2022-10-03
Cast uuid to object

### LINKED ISSUE #4435: Support casting from `uuid` to object types
We want to support doing things like `<User><uuid>$0` (and probably just `<User>$0`, with the same meaning?), and have it mean something like
`assert_exists((select detached User filter .id = <uuid>$0))`
(probably with a better error message passed to assert_exists).

This will make it a lot easier to convert ids into objects.

Offline we've discussed a bunch of alternative syntaxes and decided that just supporting casting was the least bad. If anyone wants I guess we can argue more here now.

aljazerzen:
Closes #4435.

I suggest that the two commits are reviewed separately - first one is just refactoring.

### DIFF
--- a/edb/edgeql/compiler/casts.py
+++ b/edb/edgeql/compiler/casts.py
@@ -946,10 +946,12 @@ def _find_object_by_id(
             new_stype.get_name(ctx.env.schema)
         )
 
+        from edb.edgeql.ast import BinOp, StringConstant, Path
+
         select_id = qlast.SelectQuery(
-            result=qlast.DetachedExpr(expr=qlast.Path(steps=[object_name])),
-            where=qlast.BinOp(
-                left=qlast.Path(
+            result=qlast.DetachedExpr(expr=Path(steps=[object_name])),
+            where=BinOp(
+                left=Path(
                     steps=[
                         qlast.Ptr(
                             ptr=qlast.ObjectRef(name='id'), direction='>'
@@ -961,17 +963,29 @@ def _find_object_by_id(
                 right=uuid_anchor,
             ),
         )
+
+        error_message = BinOp(
+            left=StringConstant(
+                value=(
+                    repr(new_stype.get_displayname(ctx.env.schema))
+                    + ' with id \''
+                )
+            ),
+            op='++',
+            right=BinOp(
+                left=qlast.TypeCast(
+                    expr=uuid_anchor,
+                    type=qlast.TypeName(maintype=qlast.ObjectRef(name='str')),
+                ),
+                op='++',
+                right=StringConstant(value='\' does not exist'),
+            ),
+        )
+
         ql = qlast.FunctionCall(
             func='assert_exists',
             args=[select_id],
-            kwargs={
-                'message': qlast.StringConstant(
-                    value=(
-                        repr(new_stype.get_displayname(ctx.env.schema))
-                        + ' with this id does not exist'
-                    )
-                )
-            },
+            kwargs={'message': error_message},
         )
 
         return dispatch.compile(ql, ctx=subctx)
--- a/edb/edgeql/compiler/casts.py
+++ b/edb/edgeql/compiler/casts.py
@@ -21,6 +21,7 @@
 
 
 from __future__ import annotations
+import copy
 
 from typing import *
 
@@ -98,7 +99,7 @@ def compile_cast(
         orig_stype.issubclass(ctx.env.schema, uuid_t)
         and new_stype.is_object_type()
     ):
-        return _find_object_by_id(ir_expr, new_stype, ctx=ctx)
+        return _find_object_by_id(ir_expr, new_stype, cardinality_mod, ctx=ctx)
 
     json_t = ctx.env.get_track_schema_type(
         sn.QualName('std', 'json'))
@@ -932,6 +933,7 @@ def _cast_array_literal(
 def _find_object_by_id(
     ir_expr: Union[irast.Set, irast.Expr],
     new_stype: s_types.Type,
+    cardinality_mod: Optional[qlast.CardinalityModifier],
     *,
     ctx: context.ContextLevel,
 ) -> irast.Set:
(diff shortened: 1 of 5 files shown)


edb/edgeql/compiler/casts.py
```diff
@@ -69,7 +69,7 @@ def compile_cast(
             ctx=ctx,
             srcctx=ir_expr.context)
 
-    elif irutils.is_untyped_empty_array_expr(ir_expr):
+    if irutils.is_untyped_empty_array_expr(ir_expr):
```
msullivan, line 73: Why change all the elif to if?

edb/edgeql/compiler/casts.py
```diff
@@ -919,3 +927,40 @@ def _cast_array_literal(
...
+                            ptr=qlast.ObjectRef(name='id'), direction='>'
+                        )
+                    ],
+                    partial=True,
+                ),
+                op='=',
+                right=uuid_anchor,
+            ),
+        )
+        ql = qlast.FunctionCall(func='assert_exists', args=[select_id])
```
msullivan, line 964: We should put a good error message in; assert_exists takes an optional msg argument

edb/edgeql/compiler/casts.py
aljazerzen: I find it more readable, because in this case I process this as guard statement - just a special case that does an early return.

I know that having `elif` results in same control flow, but I find them unnecessary.

msullivan: (Which definitely doesn't always make it worth doing this sort of thing, but in this kind of long function with a lot of cases, I think I like it)

edb/edgeql/compiler/casts.py
```diff
@@ -961,6 +961,17 @@ def _find_object_by_id(
                 right=uuid_anchor,
             ),
         )
-        ql = qlast.FunctionCall(func='assert_exists', args=[select_id])
+        ql = qlast.FunctionCall(
+            func='assert_exists',
+            args=[select_id],
+            kwargs={
+                'message': qlast.StringConstant(
```
msullivan, line 968: Can we make the error message here actually include the id?

tests/test_edgeql_casts.py
```diff
@@ -2631,3 +2631,24 @@ async def test_edgeql_casts_all_null(self):
...
                 json_only=json_only,
             )
+
+    async def test_edgeql_casts_uuid_to_object(self):
+        persons = await self.con.query('select Person { id }')
+
+        res = await self.con.query('select <Person><uuid>$0', persons[0].id)
+        self.assertEqual(1, len(res))
+
+        async with self.assertRaisesRegexTx(
```
msullivan, line 2641: Add a test where we apply it to a set of ids where *one* of the ids is missing and the rest are there, to make sure that that raises an error.

tests/schemas/casts.esdl
```diff
@@ -76,3 +76,7 @@ type ScalarTest {
     property p_decimal -> decimal;
     property p_json -> json;
 }
+
+type Person {
+    property first_name -> str;
```
msullivan, line 81: Just call it `name`?

msullivan (commented):
Looking solid

edb/edgeql/compiler/casts.py
msullivan: Yeah, that's a good question. Let's try benchmarking it.

tests/test_edgeql_casts.py
aljazerzen: Ha. It turns out that this does not raise an error. Did I do something wrong with contexts?

edb/edgeql/compiler/casts.py
msullivan: Otherwise I would go with something like
```
could not find <typename> with specified id
```
I think that including the id would be pretty nice, though. So if it is slow, it might be worth looking into way to make it not too bad?

tests/test_edgeql_casts.py
msullivan: Ok, I think we do need to make sure that the set appears in the scope tree, like this:
https://github.com/edgedb/edgedb/blob/master/edb/edgeql/compiler/casts.py#L480-L483

When you test this out, make sure to take a look at the scope tree with and without doing this, to try to see what's going on?

msullivan: (Interestingly, I bet that if you had `with` bound the set you were casting, and then cast that *name*, it would have worked. This is because compiling that *name* would have put the path in the scope tree)

edb/edgeql/compiler/casts.py
aljazerzen: I've tried using this:
```
> time edb cli query "select (for i in _gen_series(1, 10000000000) union (select <Person><uuid>'b90b4f92-4331-11ed-adee-17dbc27f8b02' filter .first_name = 'non-existent'));"
edb cli query   0.52s user 0.06s system 99% cpu 0.589 total
```
and it from 0.49s - 0.55s, no matter if I have ids in error message or not.

So I'm keeping it in.

tests/test_edgeql_casts.py
aljazerzen: Hmmm, after I added UUID into the error message this now passes. Is this the consequence of the strange scope tree rules?

aljazerzen:
This is mostly done, except for:
- `<User>$0` (not a cast, but a type param on query parameter)
- `<optional User>my_uuid`

edb/edgeql/compiler/casts.py
```diff
@@ -888,3 +927,65 @@ def _cast_array_literal(
...
+
+        uuid_anchor = subctx.create_anchor(
+            setgen.ensure_set(ir_expr, ctx=ctx), name='a'
+        )
+
+        object_name = s_utils.name_to_ast_ref(
+            new_stype.get_name(ctx.env.schema)
+        )
+
+        from edb.edgeql.ast import BinOp, StringConstant, Path
```
msullivan, line 949: edgedb style is to just use `qlast.BinOp`, etc

tests/test_edgeql_casts.py
msullivan: Ha, yeah.
In the original take, the reference to the anchor appeared only in a subquery. (And like, triply so: inside of SET OF argument, inside a select statement, inside a filter clause thereof.)
But once the error message was added, it was referred to in the scope *outside* of the SET OF argument to `assert_exists`, so that is where it gets bound...

msullivan (approved):
I think the core of this is ready to merge.

msullivan:
Oh, one more test idea: test `<User><optional uuid>$0` and `<User><uuid>{}`. Those ought to work. (I think they might not, since the way that the error message forces the scope handling is somewhat fragile--since msg is an OPTIONAL argument, it will get hoisted, but "optionally". Fix is like we discussed before, making sure the path is in the scope tree. We can talk about this offline too.)

msullivan:
Your plan for making user a direct argument seems reasonable.

PR #4472, opened 2022-10-04
Add `SHOULD_RETRY` tag to `IdleSessionTimeoutError`

tailhook:
While this error might be received a lot before `Parse`/`Execute`, the race condition with these is still possible.


elprans (approved):

1st1:
Should we backport to 2.x?

tailhook:
> Should we backport to 2.x?

Shouldn't be harmful. But also impact is quite limited (very rare race condition).

PR #4474, opened 2022-10-04
json_object_pack

### LINKED ISSUE #3747: Provide a way to construct arbitrary json objects
The only way to produce json objects is by casting from objects and named tuples and converting a string from `to_json`. 

I think we want something like
`function json_object_pack(els: array<tuple<str, json>>) -> json` (though it could also be `SET OF` instead?)

Annoyingly, postgres's json construction functions aren't great from an "arbitrary input" perspective, but it should still be doable?

aljazerzen:
Closes #3747

### DIFF
--- a/edb/pgsql/compiler/relgen.py
+++ b/edb/pgsql/compiler/relgen.py
@@ -3489,7 +3489,7 @@ def process_set_as_json_object_pack(
         ctx: context.CompilerContextLevel) -> SetRVars:
     expr = ir_set.expr
     assert isinstance(expr, irast.FunctionCall)
-    
+
     with ctx.subrel() as subctx:
         ir_arg = expr.args[0].expr
 
@@ -3505,8 +3505,7 @@ def process_set_as_json_object_pack(
 
         # construct the function call
         set_expr = pgast.FuncCall(
-            name = ("json_object_pack",),
-            args = [keys, values]
+            name=("jsonb_object_agg",), args=[keys, values]
         )
 
         # declare that the 'aspect=value' of ir_set (original set)
--- a/edb/pgsql/compiler/relgen.py
+++ b/edb/pgsql/compiler/relgen.py
@@ -3487,38 +3487,39 @@ def process_set_as_exists_expr(
 def process_set_as_json_object_pack(
         ir_set: irast.Set, stmt: pgast.SelectStmt, *,
         ctx: context.CompilerContextLevel) -> SetRVars:
-    assert isinstance(ir_set.expr, irast.FunctionCall)
-    func_call = ir_set.expr
-    pairs = func_call.args[0]
+    expr = ir_set.expr
+    assert isinstance(expr, irast.FunctionCall)
+    
+    with ctx.subrel() as subctx:
+        ir_arg = expr.args[0].expr
 
-    from edb.edgeql.compiler import setgen
+        # compile IR to pgast that may be a ref
+        arg_ref = dispatch.compile(ir_arg, ctx=subctx)
 
-    stype = setgen.get_set_type(pairs.expr, ctx=ctx)
+        # ensure that pgast ref is actually a value
+        arg_val = output.output_as_value(arg_ref, env=subctx.env)
 
-    keys = setgen.tuple_indirection_set(
-        pairs.expr,
-        source=stype,
-        ptr_name='0',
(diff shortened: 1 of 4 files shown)


edb/pgsql/compiler/relgen.py
```diff
@@ -3483,6 +3483,45 @@ def process_set_as_exists_expr(
...
+        # ensure that pgast ref is actually a value
+        arg_val = output.output_as_value(arg_ref, env=subctx.env)
+
+        # get first and the second fields of the tuple
+        keys = astutils.tuple_getattr(arg_val, ir_arg.typeref, '0')
+        values = astutils.tuple_getattr(arg_val, ir_arg.typeref, '1')
+
+        # construct the function call
+        set_expr = pgast.FuncCall(
+            name=("jsonb_object_agg",), args=[keys, values]
```
msullivan, line 3508: `jsonb_object_agg` will return `NULL` when there are no input rows, but we need to return an empty json object. We can solve that with a COALESCE.

edb/pgsql/compiler/relgen.py
```diff
@@ -3483,6 +3483,45 @@ def process_set_as_exists_expr(
...
 
 
+@_special_case('std::json_object_pack')
+def process_set_as_json_object_pack(
+        ir_set: irast.Set, stmt: pgast.SelectStmt, *,
+        ctx: context.CompilerContextLevel) -> SetRVars:
+    expr = ir_set.expr
+    assert isinstance(expr, irast.FunctionCall)
+
+    with ctx.subrel() as subctx:
```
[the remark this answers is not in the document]
msullivan: I think that there are a bunch of places where we create subrels without really needing to. We should try to clean some of that up.

There are also definitely a lot of places where we create subrels for purely technical internal reasons, because something or another in the compiler (often the path map) is connected to subrels. (https://github.com/edgedb/edgedb/blob/e1f0f06be29a35600dba0cb906d6a53d051ad8ff/edb/pgsql/compiler/relgen.py#L1877-L1887 for an example)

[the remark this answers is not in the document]
aljazerzen: > `stmt` is always just `ctx.rel`

I would suggest that we don't also pass `stmt` to all the functions then. It would cleanup the signatures quite a lot and only make us use two chars more :D

edb/pgsql/compiler/relgen.py
aljazerzen: Oh, I need to get attentive to that!

edb/pgsql/compiler/relgen.py
```diff
@@ -3483,6 +3483,45 @@ def process_set_as_exists_expr(
...
+        ir_arg = expr.args[0].expr
+
+        # compile IR to pgast that may be a ref
+        arg_ref = dispatch.compile(ir_arg, ctx=subctx)
+
+        # ensure that pgast ref is actually a value
+        arg_val = output.output_as_value(arg_ref, env=subctx.env)
```
msullivan, line 3500: A couple things:
1. Apparently `compile` always calls `output_as_value` on its result, so we shouldn't need to call `output_as_value`. (There are a bunch of existing call sites that I think do anyway.)
2. Let's try handling the special case (though actually pretty common; I checked, and an `array_unpack` of an array of tuples will do this) case of having a `TupleVar`. To do this, you're going to want to call `visit` instead of `compile` (so it skips computing the value), grab the path var out yourself with `get_path_value_var`, and then check on the result to see if it is a `TupleVar`

tests/test_edgeql_json.py
```diff
@@ -832,6 +832,19 @@ async def test_edgeql_json_object_unpack_05(self):
             [1, 2]
         )
 
+    async def test_edgeql_json_object_pack(self):
```
msullivan, line 835: We'll want some more tests. Try one with empty set input, with input from an `array_unpack`, and from an object property. (Object property is good because it will give you something that is definitely packed up as values, not as a TupleVar)

msullivan (commented):
This is great. I've suggestion one bug fix and two improvements

edb/pgsql/compiler/relgen.py
msullivan: This is one of those "classic edgedb bugs" that has shown up several times

PR #4479, opened 2022-10-05
Tweaks to speed up array indexing

msullivan:
While I was testing out my tuple argument stuff, it was *really* slow,
and the main culprit seemed to be array indexing.

One test case with maybe about a million elements total took long
enough to run that I gave up, so I went to investigate.

Just eliminating the "strict" on most of the array functions and their
helpers was enough to get the test runtime down to a managable 7s, but
eliminating the call to _int_or_null brought it down to 1.2s. I did
some experimentation, and merging the _int_or_null check into
_normalize_array_index also was able to bring it down to about
1.2s. (I'm not totally sure why it mattered so much.)

This meant I needed to adjust usage for the slice cases, so I pushed
the _int_bound into the call to _normalize_array_index.

I'm putting this up now in case you have any thoughts on the best way
to do this, but I think that I've probably messed the overflow edge
case handling and need to think about it a bit harder. Maybe we want a
separate `_normalize_array_slice_index` that does integrated slice
index normalizing and clamping?


aljazerzen:
Woah, that's quite a difference in time. I hadn't noticed when benching simple cases.

I agree on having separate `_normalize_array_index` and `_normalize_array_slice_index` and chugging all the logic in there. I wanted to do this already, but didn't want to make too much changes.

There are also some proxy functions for string and bytea functions that could be dropped if we inline a litte bit of code.

I'll also check the overflow cases.

edb/pgsql/metaschema.py
```diff
@@ -1920,7 +1922,6 @@ def __init__(self) -> None:
             args=[('index', ('bigint',)), ('length', ('int',))],
             returns=('bigint',),
```
aljazerzen, line 1923: this can now be int

edb/pgsql/metaschema.py
```diff
@@ -2043,8 +2040,8 @@ class StringIndexWithBoundsFunction(dbops.Function):
             "detail"
         )
         FROM (
-            SELECT edgedb._int_or_null(
-                edgedb._normalize_array_index("index", char_length("val"))
+            SELECT (
+                edgedb._normalize_array_index("index", char_length("val"))::int
```
aljazerzen, line 2044: with these changes, _normalize_array_index can return `int`, so cast here is not needed

edb/pgsql/metaschema.py
```diff
@@ -2081,8 +2077,8 @@ class BytesIndexWithBoundsFunction(dbops.Function):
             "detail"
         )
         FROM (
-            SELECT edgedb._int_or_null(
-                edgedb._normalize_array_index("index", length("val"))
+            SELECT (
+                edgedb._normalize_array_index("index", length("val"))::int
```
aljazerzen, line 2081: same

edb/pgsql/metaschema.py
```diff
@@ -2172,12 +2167,12 @@ class StringSliceImplFunction(dbops.Function):
...
-            ))) as pg_start,
-            GREATEST(0, edgedb._int_bound(edgedb._normalize_array_index(
-                stop, edgedb._length(val)
-            ))) as pg_end
+            GREATEST(0, edgedb._normalize_array_index(
+                edgedb._int_bound(start), edgedb._length(val)
+            ))::int as pg_start,
+            GREATEST(0, edgedb._normalize_array_index(
+                edgedb._int_bound(stop), edgedb._length(val)
+            ))::int as pg_end
```
aljazerzen, line 2175: `GREATEST` is not needed anymore. 

Now the output of `_normalize_array_index` is on interval `[0, 2147483647]`. Before the output of `_int_bound` was `[-2147483648, 2147483647]`, because `_normalize` could produce very negative numbers. See my comment here: https://github.com/edgedb/edgedb/pull/4448#discussion_r983950307

aljazerzen (approved):
The change to _normalize_array_index simplifies a lot.

msullivan:
> Woah, that's quite a difference in time. I hadn't noticed when benching simple cases.

I think that it wasn't exposed until strict got removed from all the relevant functions.

PR #4480, opened 2022-10-05
Parsing SQL with libpg_query and translating to our pg.ast

aljazerzen:
Just a proof of concept, does not have to reside here and have `libpg_query` as a git submodule.


edb/pypg_query/setup.py
```diff
@@ -0,0 +1,18 @@
+from setuptools import setup, Extension
+from Cython.Build import cythonize
+
+import os.path
+
+libpg_query = os.path.join('.', 'libpg_query')
+
+extensions = cythonize([
```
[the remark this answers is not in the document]
aljazerzen: Is the overall setup ok? To have libpg_query as a git submodule and build it from setup.py? And code for generating our AST from the json also part of this python package, named pypg_query?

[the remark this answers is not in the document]
elprans: Yeah, though I'd name the module `pg_sql_parser` or something like that, because it's not actually doing any queries, it's pure parsing.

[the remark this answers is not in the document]
elprans: Maybe even move it under `edb/pgsql/parser`?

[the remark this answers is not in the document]
aljazerzen: `edb/pgsql/parser` would make the most sense, yes.

.gitmodules
```diff
@@ -5,3 +5,7 @@
 [submodule "edb/server/pgproto"]
 	path = edb/server/pgproto
 	url = https://github.com/MagicStack/py-pgproto.git
+[submodule "edb/pypg_query/libpg_query"]
+	path = edb/pypg_query/libpg_query
+	url = https://github.com/lfittl/libpg_query.git
```
msullivan, line 10: `https://github.com/pganalyze/libpg_query` seems canonical?

edb/pgsql/parser/.gitignore
```diff
@@ -0,0 +1,5 @@
+*.so
+psqlparse/parser.c
+build/
+__pycache__/
+*.c
```
msullivan, line 5: Make sure files have the trialing newlines; a lot of tools get irritated about it

aljazerzen:
This is done from my side, but it requires some dev-ops. 

Can someone looks at gihub actions workflows I added? "Build libpg_query" step needs a cache, because it takes 42 sec.

setup.py
```diff
@@ -377,9 +405,10 @@ class build(setuptools_build.build):
         setuptools_build.build.sub_commands
         + [
             ("build_metadata", lambda self: True),
+            ("build_libpg_query", lambda self: True),
             ("build_parsers", lambda self: True),
             ("build_postgres", lambda self: True),
-            ("build_cli", lambda self: True),
+            # ("build_cli", lambda self: True),
```
elprans, line 411: Don't forget to uncomment this.

edb/pgsql/parser/ast_builder.py
```diff
@@ -0,0 +1,499 @@
+from typing import *
```
elprans, line 1: Please add the standard header to all newly-added files.

edb/pgsql/parser/.gitignore
```diff
@@ -0,0 +1,5 @@
+*.so
+psqlparse/parser.c
```
elprans, line 2: This `.gitignore` seems to be unnecessary, because all ignored files are covered by the top-level `.gitignore` (and `./build` is no longer produced).

setup.py
```diff
@@ -559,6 +588,22 @@ def run(self, *args, **kwargs):
...
+    user_options = []
+
+    def initialize_options(self):
+        pass
+
+    def finalize_options(self):
+        pass
+
+    def run(self):
+        if os.environ.get("EDGEDB_BUILD_PACKAGE"):
```
elprans, line 604: Remove this block.  `EDGEDB_BUILD_PACKAGE` indicates a production package (Postgres and the CLI are built separately in prod workflow because they require special handling, whereas libpq_query likely doesn't).

edb/pgsql/parser/ast_builder.py
```diff
@@ -0,0 +1,499 @@
...
+    materialized = None
+    if n["ctematerialized"] == "CTEMaterializeAlways":
+        materialized = True
+    elif n["ctematerialized"] == "CTEMaterializeNever":
+        materialized = False
+
+    return pgast.CommonTableExpr(
+        name=n["ctename"],
+        query=_build_query(n["ctequery"]),
+        # recursive=False,
```
elprans, line 198: Is this a TODO?

edb/pgsql/parser/ast_builder.py
```diff
@@ -0,0 +1,499 @@
...
+        op=op,
+        all=n["all"] if "all" in n else False,
+        larg=_maybe(n, "larg", _build_select_stmt),
+        rarg=_maybe(n, "rarg", _build_select_stmt),
+        ctes=_maybe(n, "withClause", _build_ctes),
+    )
+
+
+def _build_insert_stmt(n: Node) -> pgast.InsertStmt:
+    # Insert is currently unsupported,
```
elprans, line 130: Obsolete comment?

elprans (approved):
This looks good to me (modulo minor nits).

> Can someone looks at gihub actions workflows I added? "Build libpg_query" step needs a cache, because it takes 42 sec.

Should be a straightforward adaptation of the Postgres caching bits.  We can add that after merging this.

tests/test_pgsql_parse.py
```diff
@@ -0,0 +1,616 @@
...
+
+
+def parse_and_gen(sql: str) -> str:
+    ast = parser.parse(sql)
+    sql_stmts = [codegen.generate_source(stmt, pretty=False) for stmt in ast]
+    sql = "; ".join(sql_stmts)
+    return sql.replace("  ", " ").replace("( (", "((").replace(") )", "))")
+
+
+class TestEdgeQLSelect(tb.TestCase):
```
msullivan, line 34: I forgot this, but we have a framework for writing tests where input and output can be specified as strings. An example is here: https://github.com/edgedb/edgedb/blob/master/tests/test_edgeql_ir_card_inference.py#L31

Your call if you want to try to use it.

I think here it might actually make things worse, since you properly handle whitespace in your tests. I think that for the parse/codegen tests that we have for the edgeql parser, we actually do the comparison after stripping away all whitespace!

edb/pgsql/parser/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+from .parser import pg_parse
+from .ast_builder import build_queries
+
+from typing import *
+
+from edb.pgsql import ast as pgast
+
+
+def parse(sql_query: str) -> List[pgast.Query]:
+    import json
```
msullivan, line 11: Put the json import at the top level

msullivan (approved):
This looks good.

Like we discussed earlier, we should also add a ParserContext to the AST and populate it based on source information from the json. That can be a follow-up PR, though.

edb/pypg_query/demo.txt
```diff
@@ -0,0 +1,100 @@
+$ python sql_to_pgast.py 'SELECT first_name, Person.* FROM Persons p JOIN Posts ON p.id = posts.posted_by WHERE LENGTH(posts.title) > 10'
```
1st1: I suggest renaming to README.txt

1st1: And adding other relevant info explaining what this subpackage does.

edb/pgsql/parser/exceptions.py
```diff
@@ -0,0 +1,21 @@
+from typing import Optional
```
1st1, line 1: Please add our standard copyright comment to all new .py files.

tests/test_pgsql_parse.py
```diff
@@ -0,0 +1,616 @@
...
+        )
+        self.assertEqual(
+            parse_and_gen("UPDATE x SET a = 1, b = 2, c = 3"),
+            "UPDATE x SET a = 1, b = 2, c = 3",
+        )
+        self.assertEqual(
+            parse_and_gen("UPDATE x SET z = now()"), "UPDATE x SET z = now()"
+        )
+
+    @unittest.skip("unsupported SQL constructs")
```
1st1, line 492: What's the intent here? To eventually fix some of them? You should use xfail for such.

As for parsing unsupported commands like `PREPARE a123 AS SELECT a` -- you shouldn't fail on the first one, you should test failing on *every one* of them.

1st1 (commented):
Impressive work, please fix the nits I commented on (specifically the tests).

edb/pypg_query/demo.txt
aljazerzen: Oh, I commited this file just to show sully how far this is.

tests/test_pgsql_parse.py
aljazerzen: Intent is to save this collection of tests if we decide to support these features in the future.

Will change to xfail.

PR #4499, opened 2022-10-11
Properly propagate creation and deletion of extensions

elprans:
`CREATE EXTENSION` (and `DROP` too) are currently happily ignored by the
adjacent servers, but also, in some cases even by the local server.


fantix (approved):
LGTM!

PR #4500, opened 2022-10-11
Make it possible to specify and toggle server readiness state

elprans:
To distinguish between a "live" server and a "ready" server, add a new
`/server/state/alive` API to check for server liveness specifically.

The `/server/state/ready` check now depends not just on system liveness,
but also on the internal readiness state flag, the initial value of which
can be specified with `--initial-readiness-state={ready|not_ready}` or
its equivalent environment variable `EDGEDB_SERVER_INITIAL_READINESS_STATE`.

The readiness state can be set to `ready` via a `SIGUSR1` signal to the
main server process and to `not_ready` via a `SIGUSR2` signal.


edb/server/server.py
```diff
@@ -751,6 +758,16 @@ async def introspect_db(self, dbname):
...
+            extension_names_json = await conn.sql_fetch_val(
+                b'''
+                    SELECT json_agg(name) FROM edgedb."_SchemaExtension";
+                ''',
+            )
+            if extension_names_json:
+                extensions = set(json.loads(extension_names_json))
+            else:
+                extensions = set()
+
```
fantix, line 770: looks like u forgot to drop this debug code

fantix (approved):
OK for me on code level, curious about the answers to Paul's questions too.

nsidnev (approved):

elprans:
> 1. Can it be controlled by signals only?
> 2. Is it purely informational, or does it influence queries somehow?
> 3. Can the status be fetched by the EdgeDB binary protocol, rather than HTTP?
> I'm also missing context of what specific utility for this.

As @nsidnev pointed out above this is mostly intended for service orchestrators that need the ability to fence an EdgeDB node in a multi-node instance.  Signals are the simplest and most widely supported [way of controlling containerized tasks](https://www.nomadproject.io/api-docs/allocations#signal-allocation). 

> I'm thinking that this can be combined with our `disable_queries_with_message := "cannot apply migrations automatically"` concept we need for `migrate --dev-mode` (sorry no public discussion for this to link to). I don't mean it should be the same toggle, but I think it can be a coherent set of variables that are exposed both via HTTP and binary protocol.

The idea behind a "readiness" state is that the node becomes non-routeable in a service mesh.  Blocking queries in an already-established connection is too late and will only confuse the client.

elprans:
I've changed my mind re the original design in this PR.  Specifying the initial readiness state in the command line directly is problematic, because if that is part of an orchestrator task definition, then the
server might be put into `not_ready` state inadvertently when a task is restarted for any reason.  Thus, replace `--initial-readiness-state` with a monitored `--readiness-state-file`.  Changing files in containers should be as easy as sending a signal.

nsidnev:
I'm not sure about the new design... I liked the old one more.

> Specifying the initial readiness state in the command line directly is problematic, because if that is part of an orchestrator task definition, then the server might be put into not_ready state inadvertently when a task is restarted for any reason.

Although I admit that this part is true, but in addition to the flag there was a convenient environment variable, which, instead of a flag, can be set dynamically during the start of the task.

The difference is not too big, but with the previous one, we had to do something like that when defining the task for the orchestrator:
1. Сheck if we need to start the server with a readiness failure (this would most likely require inspecting the dynamic configuration storage when the task runs, rather than the parameters passed to the task, since they are tied to the task scheduled for the execution), and if so, write the value in the env
2. As soon as everything necessary is done, we just send a signal and the server becomes ready (in the meantime we update the storage, so the 1st part will be skipped when the orchestrator restarts the task)

After that, even on restart the server will be ready, and we won't have to worry about the state in the environment variables

With the current design we still need to do all of the above + some additional work:
1. We still need to check which state we need to write to the container and write this file + add an env var/flag for the server when we define the task
2. Once we have done our stuff, we need to update the file in the container directly + send a signal + update the storage so that when the task is restarted, the file with the desired state is created (or the file is not created and the env var/flag is not set for the server)

Perhaps I'm missing something, why the new design should be easier to use, I'd be happy to hear an explanation for that.

elprans:
Hm, I thought we would always write a state file to storage that survives restarts and state flips would only require writing to that file.  No signaling is necessary since the server is watching the file itself.

nsidnev:
> No signaling is necessary since the server is watching the file itself.

Ah, I got that. Then it's no longer an issue.

nsidnev (approved):

tailhook:
> The idea behind a "readiness" state is that the node becomes non-routeable in a service mesh. Blocking queries in an already-established connection is too late and will only confuse the client.

Sure. But do I understand correctly, that instance is fully functional (you can make queries and DDLs to it), but you can't connect to it through service mesh? So we are relying on service mesh to be flawless for the correctness of our approach?

As an example, can some cache on the service mesh side + IP reuse on hosting side ruin our database setup?

fantix (approved):

elprans:
> As an example, can some cache on the service mesh side + IP reuse on hosting side ruin our database setup?

I'm not sure I understand.  The fencing is still done by the server itself, there is no functional difference from this approach vs using some configuration value except that it's a lot easier for the multi-layered orchestration system to write files than it is to execute EdgeQL.

PR #4503, opened 2022-10-11
Make object type descriptor ids be derived from type name

jaclarke:
Using the object type name instead of the object type id makes type descriptor id's stable across different database instances.

### DIFF
--- a/edb/server/compiler/sertypes.py
+++ b/edb/server/compiler/sertypes.py
@@ -278,7 +278,7 @@ class TypeSerializer:
         elif isinstance(t, s_objtypes.ObjectType):
             # This is a view
             self.schema, mt = t.material_type(self.schema)
-            base_type_name = mt.get_displayname(self.schema)
+            base_type_name = str(mt.get_name(self.schema))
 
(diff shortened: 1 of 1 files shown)


edb/server/compiler/sertypes.py
```diff
@@ -278,7 +278,7 @@ def _describe_type(self, t, view_shapes, view_shapes_metadata,
         elif isinstance(t, s_objtypes.ObjectType):
             # This is a view
             self.schema, mt = t.material_type(self.schema)
-            base_type_id = mt.id
+            base_type_name = mt.get_displayname(self.schema)
```
elprans, line 281: Use `get_name` instead.  `get_displayname` is intended for user-visible messages and does not identify a (possibly derived) type uniquely.

msullivan: (`get_name` gives back a `Name`, but you can just call `str` on it to get a string.)

elprans (approved):

1st1 (approved):

PR #4511, opened 2022-10-12
Check for invalid arrays at the protocol level

### LINKED ISSUE #4405: Ensure that array arguments are 1-indexed and 1 dimentional
Postgres supports arbitrary array indexes:
```
SELECT '[5:10]={1,2,3,4,5,6}'::int[] -- array that starts indexing at 5
```
... but they default to being 1-indexed. Our compiler and stdlib assume that all arrays are 1-indexed, which is almost true: we never construct non-1-index array, but one may be passed as query argument.

This is probably the indexing offset: https://github.com/edgedb/edgedb-python/blob/master/edgedb/protocol/codecs/array.pyx#L82

> It should actually be pretty easy to add a check for this in the server's protocol handling code, because we already have code for doing special handling of array parameters. Postgres's encoding of arrays requires including the internal type ID of the element; we didn't want to make our clients need to understand that, so we have to handle it in our server: https://github.com/edgedb/edgedb/blob/a8633eb9d19e8696b0f04fce86a49435ac7124d0/edb/server/protocol/args_ser.pyx#L169-L177

> So that gives a pretty natural place to also check the dimensions and flags.

msullivan:
Currently we allow array arguments containing NULL and with unexpected
dims and bounds. Check and prevent this.

Fixes #4405.

Performance impact seemed negligable.

### DIFF
--- a/edb/server/protocol/args_ser.pyx
+++ b/edb/server/protocol/args_ser.pyx
@@ -103,6 +103,7 @@ cdef WriteBuffer recode_bind_args(
         int32_t decl_args
         ssize_t in_len
         ssize_t i
+        int32_t array_tid
         const char *data
         bint live = positions is None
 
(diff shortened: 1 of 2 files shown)


edb/server/protocol/args_ser.pyx
```diff
@@ -209,6 +205,55 @@ cdef WriteBuffer recode_bind_args(
...
+        ssize_t idx
+        ssize_t num
+        ssize_t tag
+        FRBuffer sub_buf
+
+    frb_slice_from(&sub_buf, in_buf, in_len)
+
+    val = hton.unpack_int32(frb_read(&sub_buf, 4)) # ndims
+    if val != 1 and val != 0:
+        raise errors.QueryError("unsupported array dimensions")
```
elprans, line 229: `InputDataError`?

msullivan: Sure. Made all the errors in args_ser.pyx be InputDataError

fantix (approved):
LGTM!

elprans:
I wonder if we could do this checking in SQL? (e.g use `_nullif_array_nulls`)

msullivan:
> I wonder if we could do this checking in SQL? (e.g use `_nullif_array_nulls`)

I considered that approach, but it didn't really seem nicer to me since we were already doing special processing for arrays and because we already are doing the NULL check in args_ser for the tuple argument case.

We'd possibly want to reuse the `param_cte` mechanism to avoid duplicating the calls to it, but that would be fine.

Mostly it just seems to make sense to me to keep our input protocol validation done in `args_ser`, though

elprans (approved):

PR #4514, opened 2022-10-12
Fix handling of link properties named 'id'

### LINKED ISSUE #4237: Create Index in abstract Link crashes Migration
- EdgeDB Version: 2.1
- OS Version: macOS 12.5 (M1 Pro, Also Checked in Docker Image with `amd64` architecture)

Steps to Reproduce:

1. Create Previous Schema (Works well!)
2. Add Index in Abstract Link
3. Create Migration and Apply, It Crashes.

```
edgedb error: InternalServerError: relation "edgedbpub.f6cc2eae-17e5-11ed-8e92-83df59ae4c33" does not exist
  Server traceback:
      edb.errors.InternalServerError: relation "edgedbpub.f6cc2eae-17e5-11ed-8e92-83df59ae4c33" does not exist
```

p.s. I'm new with EdgeDB, so It can be schema error. If my schema is wrong, Just Tell me please. :) I think this error message can be contain more information. 

Schema:

```
module default {
  # Base Model for Define Essential Values
  abstract type BaseModel {
    overloaded required property id -> uuid {
      constraint exclusive;
      readonly := true;
      default := uuid_generate_v1mc();
    };

    required property created_at -> datetime;
    required property updated_at -> datetime;
    property deleted_at -> datetime;
    required property deleted -> bool {
      default := false;
    };

    index on (.created_at) {
      annotation title := "CreatedAt Index";
    };
  }

  # Define User Model
  type User extending BaseModel {
    # Email is unique.
    required property email -> str {
      constraint exclusive;
    };

    required property password -> str;

      constraint min_len_value(3);
    };

    index on (.email) {
      annotation title := "Email Index";
    };
  }

  abstract link IdIndex {
    property id -> uuid;
    # ----------------------------------
    # here is the problem!!
    index on (__subject__@id);
    # ----------------------------------
  }

  # Define Memo and Comment Model
  type Memo extending BaseModel {
    required link created_by extending IdIndex -> User;
    multi link accessable_users extending IdIndex -> User;

    required property content -> str;
    required property tags -> array<str> {
      default := <array<str>>[];
    };
  }

  type Comment extending BaseModel {
    required link created_by extending IdIndex -> User;
    required link memo extending IdIndex -> Memo;
    required property content -> str;
  }
}
```

msullivan:
Link props named 'id' were being misidentified as special pointers,
and so didn't trigger the creation of link tables properly.

Fixes #4237.

### DIFF
--- a/edb/edgeql/compiler/viewgen.py
+++ b/edb/edgeql/compiler/viewgen.py
@@ -1097,9 +1097,8 @@ def _normalize_view_ptr_expr(
         )
 
     # Prohibit invalid operations on id and __type__
-    ptrcls_sn = ptrcls.get_shortname(ctx.env.schema)
     id_access = (
-        ptrcls_sn.name == 'id'
+        ptrcls.is_id_pointer(ctx.env.schema)
         and (
             not ctx.env.options.allow_user_specified_id
             or not exprtype.is_mutation()
@@ -1110,6 +1109,7 @@ def _normalize_view_ptr_expr(
         and (id_access or ptrcls.is_protected_pointer(ctx.env.schema))
         and not from_default
     ):
+        ptrcls_sn = ptrcls.get_shortname(ctx.env.schema)
         if is_polymorphic:
             msg = (f'cannot access {ptrcls_sn.name} on a polymorphic '
                    f'shape element')
--- a/edb/ir/typeutils.py
+++ b/edb/ir/typeutils.py
@@ -784,7 +784,7 @@ def is_id_ptrref(ptrref: irast.BasePointerRef) -> bool:
     """Return True if *ptrref* describes the id property."""
     return (
         str(ptrref.std_parent_name) == 'std::id'
-    )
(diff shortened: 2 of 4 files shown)


edb/edgeql/compiler/viewgen.py
```diff
@@ -1097,9 +1097,8 @@ def _normalize_view_ptr_expr(
         )
 
     # Prohibit invalid operations on id and __type__
-    ptrcls_sn = ptrcls.get_shortname(ctx.env.schema)
     id_access = (
-        ptrcls_sn.name == 'id'
```
elprans, line 1102: I have a nagging suspicion that this might not be the only such place.

elprans (approved):

edb/edgeql/compiler/viewgen.py
msullivan: Yeah.

Maybe we should just ban it?

PR #4521, opened 2022-10-13
Disable restore-keys for cython extension builds

msullivan:
Cython incremental builds seem to be unreliable?


msullivan:
Thoughts? I'm pretty sure this is what was causing me a lot of cache grief on the 2.x branch earlier. (And I've already applied this change on the 2.x branch)

fantix (approved):
Yeah I think this makes sense, I've also met unreliability if not cleaned.

PR #4529, opened 2022-10-14
Report failing access policy

### LINKED ISSUE #4095: Add custom error message on object level security constraints
Currently, when a access policy constraint is violated the error looks like this:
`edgedb error: EdgeDBError: access policy violation on insert of default::User`

It does not carry any info about what constraint has been violated (it can have more than 1 insert constraint for eg).

Would be great to be able to set a custom error message like we do on type/field level constraints.

aljazerzen:
Closes #4095

- refactor clauses.py
- report names of failed access policies

I have tests to do, but would ask for a quick peek if the major changes are ok.

### DIFF
--- a/edb/edgeql/compiler/policies.py
+++ b/edb/edgeql/compiler/policies.py
@@ -348,7 +348,12 @@ def compile_dml_write_policies(
             name = str(pol.get_shortname(schema))
 
             policies.append(
-                irast.WritePolicy(action=action, name=name, expr=ir_set)
+                irast.WritePolicy(
+                    expr=ir_set,
+                    action=action,
+                    name=name,
+                    error_msg=pol.get_errmessage(schema),
+                )
             )
 
         return irast.WritePolicies(policies=policies)
--- a/edb/edgeql/compiler/policies.py
+++ b/edb/edgeql/compiler/policies.py
@@ -39,8 +39,6 @@ from . import context
 from . import dispatch
 from . import setgen
 
-T = TypeVar("T")
-
 
 def get_access_policies(
     stype: s_objtypes.ObjectType, *, ctx: context.ContextLevel,
--- a/edb/edgeql/compiler/policies.py
+++ b/edb/edgeql/compiler/policies.py
@@ -39,6 +39,8 @@ from . import context
 from . import dispatch
 from . import setgen
 
+T = TypeVar("T")
+
 
 def get_access_policies(
     stype: s_objtypes.ObjectType, *, ctx: context.ContextLevel,
@@ -125,6 +127,8 @@ def get_rewrite_filter(
 ) -> Optional[qlast.Expr]:
     schema = ctx.env.schema
     pols = get_access_policies(stype, ctx=ctx)
+    if not pols:
+        return None
 
     ctx.anchors = ctx.anchors.copy()
 
@@ -317,27 +321,21 @@ def compile_dml_write_policies(
     mode: qltypes.AccessKind, *,
     ctx: context.ContextLevel,
 ) -> Optional[irast.WritePolicies]:
-    """Compile a policy filter for a DML statement at a particular type"""
+    """Compile policy filters and wrap them into irast.WritePolicies"""
     if not ctx.env.type_rewrites.get((stype, False)):
-        return None
-
-    pols = get_access_policies(stype, ctx=ctx)
-    if not pols:
         return None
 
     with ctx.detached() as _, ctx.newscope(fenced=True) as subctx:
         # TODO: can we make sure to always avoid generating needless
         # select filters
-        skip_subtypes = (stype, False) not in ctx.env.type_rewrites
-        result = setgen.class_set(
-            stype, path_id=result.path_id, skip_subtypes=skip_subtypes, ctx=ctx
-        )
-
-        subctx.anchors[qlast.Subject().name] = result
-        subctx.partial_path_prefix = result
+        _prepare_dml_policy_context(stype, result, ctx=subctx)
 
         schema = subctx.env.schema
         subctx.anchors = subctx.anchors.copy()
+
+        pols = get_access_policies(stype, ctx=ctx)
+        if not pols:
+            return None
 
         policies = []
         for pol in pols:
@@ -367,20 +365,10 @@ def compile_dml_read_policies(
     if not ctx.env.type_rewrites.get((stype, False)):
         return None
 
-    pols = get_access_policies(stype, ctx=ctx)
-    if not pols:
-        return None
-
     with ctx.detached() as _, ctx.newscope(fenced=True) as subctx:
         # TODO: can we make sure to always avoid generating needless
         # select filters
-        skip_subtypes = (stype, False) not in ctx.env.type_rewrites
-        result = setgen.class_set(
-            stype, path_id=result.path_id, skip_subtypes=skip_subtypes,
-            ctx=ctx)
-
-        subctx.anchors[qlast.Subject().name] = result
-        subctx.partial_path_prefix = result
+        _prepare_dml_policy_context(stype, result, ctx=subctx)
 
         condition = get_rewrite_filter(stype, mode=mode, ctx=subctx)
         if not condition:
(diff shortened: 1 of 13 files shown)


aljazerzen:
For this schema:
```
module default {
  global cur_user -> str;

  type Movie {
    property release_year -> int64;
    multi link actors -> Person;
    link director -> Person;

    access policy insert_only_by_aljaz
      allow insert using (
          (global cur_user = 'aljaz') ?? false
      );
    access policy prevent_cats
      deny insert using (.title = 'cats');

	access policy cur_watchers
      allow select using (
          (global cur_user = 'admin') ?? false
      );
  }  
}
```

This is how it displays the errors:
```
edgedb> insert Movie { title := 'space man' };
edgedb error: AccessPolicyError: access policy violation on insert of default::Movie (none of these allow policies match: __::insert_only_by_aljaz)

edgedb> set global cur_user := 'aljaz';
OK: SET GLOBAL

edgedb> insert Movie { title := 'space man' };
{default::Movie {id: a78638c4-4bd6-11ed-85fe-9374378a0a92}}


edgedb> insert Movie { title := 'cats' };
edgedb error: AccessPolicyError: access policy violation on insert of default::Movie (denied by policy __::prevent_cats)


edgedb> insert Movie { title := '1942' };
{default::Movie {id: ae02dbc6-4bd6-11ed-85fe-b79b097b96ae}}


edgedb> select Movie { title };
{}

edgedb> set global cur_user := 'admin';
OK: SET GLOBAL

edgedb> select Movie { title };
{default::Movie {title: 'space man'}, default::Movie {title: '1942'}}
```

edb/edgeql/compiler/policies.py
```diff
@@ -314,12 +311,58 @@ def try_type_rewrite(
...
 
-def compile_dml_policy(
+def compile_dml_write_policies(
     stype: s_objtypes.ObjectType,
     result: irast.Set,
     mode: qltypes.AccessKind, *,
     ctx: context.ContextLevel,
-) -> Optional[irast.PolicyExpr]:
+) -> Optional[irast.WritePolicies]:
+    """Compile a policy filter for a DML statement at a particular type"""
```
msullivan, line 320: update doc comment

edb/edgeql/compiler/policies.py
```diff
@@ -314,12 +311,58 @@ def try_type_rewrite(
     type_rewrites[rw_key] = filtered_set
 
 
-def compile_dml_policy(
+def compile_dml_write_policies(
```
msullivan, line 314: Think about whether there is a reasonably nice way to reduce the duplication between the read and the write side. There might not be (sometimes the cure to a little bit of duplication is worse than the disease), but think about it

edb/pgsql/compiler/dml.py
```diff
@@ -839,33 +838,81 @@ def compile_policy_check(
         dml_rvar = relctx.rvar_for_rel(dml_cte, ctx=ctx)
         relctx.include_rvar(ictx.rel, dml_rvar, path_id=subject_id, ctx=ictx)
 
-        cond_ref = clauses.compile_filter_clause(
-            policy_expr.expr, policy_expr.cardinality, ctx=ictx)
+        def raise_if(
```
msullivan, line 841: This looks solid and like it should work usefully.

I'd like to go a bit further in what we produce, though, if we can, and produce an error message that lists *all* of the failing deny rules. It will probably end up being a bit gnarlier than your approach, which is fairly elegant :P

aljazerzen:
The codegen is getting complex, I don't know how to simplify it.

This is generated SQL
```SQL
...
  "policy~1" AS MATERIALIZED (
    (
        SELECT
            edgedb.raise_on_null(
                nullif("expr-29~2"."expr~29_value~1", false),
                'insufficient_privilege',
                msg => 'access policy violation on insert of default::Movie (none of these allow policies match: __::insert_only_by_aljaz)',
                "table" => 'cd33f102-4ee2-11ed-93f1-01441a342449'
            ) AS error_allow,
            edgedb.raise_on_not_null(((
                SELECT string_agg(error_msg, ', ')
                FROM ((
                    SELECT 'denied by policy __::prevent_cats'
                    WHERE "expr-31~2"."expr~31_value~1"
                ) UNION (
                    SELECT 'denied by policy __::prevent_start_on_c'
                    WHERE "expr-35~2"."expr~35_value~1"
                )) AS t(error_msg))),
                'insufficient_privilege',
                msg => (
                    'access policy violation on insert of default::Movie (' || (((
                        SELECT string_agg(error_msg, ', ')
                        FROM ((
                            SELECT 'denied by policy __::prevent_cats'
                            WHERE "expr-31~2"."expr~31_value~1"
                        ) UNION (
                            SELECT 'denied by policy __::prevent_start_on_c'
                            WHERE "expr-35~2"."expr~35_value~1"
                        )) AS t(error_msg))
                    ) || ')')
                ),
                "table" => 'cd33f102-4ee2-11ed-93f1-01441a342449'
            ) AS error_deny
        FROM
...
```

edb/pgsql/compiler/astutils.py
```diff
@@ -198,6 +198,81 @@ def extend_binop(
...
+        result = stmt
+
+    for s in stmt_list:
+        if s is not None and s is not result:
+            result = pgast.SelectStmt(larg=result, op=op, rarg=s)
+
+    return result
+
+
+def conditional_string_agg(
```
msullivan, line 236: This is obscure enough that I'd probably just stick it in `dml.py` next to the call site

edb/pgsql/compiler/dml.py
```diff
@@ -839,33 +838,99 @@ def compile_policy_check(
...
+        else:
+            allow_expr = pgast.BooleanConstant(val='false')
+
+        allow_msgs = (pol.error_msg for pol, _ in allow if pol.error_msg)
+        if allow_msgs:
+            hint = 'none of these allow policies match: '
+            hint += ', '.join(allow_msgs)
+            allow_msg = f'{msg} ({hint})'
+        else:
+            allow_msg = f'{msg} (no allow policies)'
```
msullivan, line 894: I think I'd prefer to drop most of the extra helpful messages here, and just say `access policy violation` and then, if applicable, following it, the relevant specified error messages.

My reasoning here is that the particular error might be sensitive information, and we don't want to give out any more than what the user asked to give. I'm willing to be convinced that this is not necessary, though.

I think it would be best to be able to produce error messages from both allow and deny rules at the same time, too?

The good news is I think all of these nitpicks should simplify the code, not make it more complex :P

edb/pgsql/compiler/dml.py
```diff
@@ -839,33 +838,99 @@ def compile_policy_check(
...
+                        val=f'denied by policy {pol.error_msg}'
+                    ),
+                    cond,
+                )
+                for pol, cond in deny
+                if pol.error_msg
+            ]
+            deny_message = astutils.conditional_string_agg(deny_messages)
+            if deny_message:
+                deny_message = astutils.extend_concat(
```
msullivan, line 921: Once the errmessage stuff is there, we'll need to handle the case where it errors but nothing with an error message failed

msullivan (commented):
This is very solid looking, and it does a great job of handling pretty nasty requirements nicely.

I've got a hopefully final round of formatting nitpicks, though I'm also open to arguments about them :P

edb/pgsql/compiler/astutils.py
```diff
@@ -198,6 +198,81 @@ def extend_binop(
...
+            where_clause=cond,
+        )
+        for value, cond in pairs
+    ]
+    union = extend_select_op(None, *selects)
+
+    if not union:
+        return None
+
+    return pgast.SubLink(
```
msullivan, line 252: Is the SubLink needed? Can't you just return the Select?

edb/pgsql/compiler/dml.py
aljazerzen: I agree, given that this is security it should not be verbose when denying access.

msullivan (approved):
Looks great!

msullivan:
As reference for later documentation (in #4943), I believe the algorithm is:
 * Collect all "relevant" policy errmessage: this means every deny rule with an errmessage that was true, and every allow rule with an errmessage that was false
 * (Note that this means that either every allow rule with an errmessage will be included or none of them will)
 * If at least one message was collected, stick them all comma separated in the error message
 * Otherwise return a fully generic error message that does *not* indicate which rules failed

The rationale for this system is to avoid leaking any information that the developer did not opt-in to leak, in case the error messages are directly exposed to an end-user. *Why* a security policy failed is often sensitive information, so it needs to only be opt-in.

PR #4532, opened 2022-10-14
Fix casting empty sets to standard library enum types

### LINKED ISSUE #4530: ISE when casting result of `json_set`
```
edgedb> select <tuple<str>>json_set(to_json('["b"]'), "0", value := <json>"a");
edgedb error: InternalServerError: type "edgedbpub.7f1aef75-4be4-11ed-a553-a70cebd6e089_domain" does not exist
  Server traceback:
      edb.errors.InternalServerError: type "edgedbpub.7f1aef75-4be4-11ed-a553-a70cebd6e089_domain" does not exist
```

msullivan:
When making up path_ids for empty sets, we need to keep track of
whether the actual type came from a standard library module, so that
we can propertly figure out whether it is in edgedbstd or edgedbpub.

The issue where this first came up was when casting the result of
json_set: an optional wrapper is being inserted for the
`empty_treatment` argument in this cast case. (That seems like its own
performance bug, too.)

Fixes #4530.

### DIFF
--- a/edb/edgeql/compiler/pathctx.py
+++ b/edb/edgeql/compiler/pathctx.py
@@ -31,6 +31,7 @@ from edb.ir import typeutils as irtyputils
 
 from edb.schema import name as s_name
 from edb.schema import pointers as s_pointers
+from edb.schema import schema as s_schema
 from edb.schema import types as s_types
 
 from . import context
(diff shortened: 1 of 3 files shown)


elprans (approved):

msullivan:
I think I actually do not like this solution but let's ship it in 2.4 and then clean it up

PR #4535, opened 2022-10-16
Produce the correct value for '\b' escapes

### LINKED ISSUE #4534: Inserting \b byte gets changed in the database to 0x10
- EdgeDB Version: 2.3
- OS Version: Windows 10

This only seems to happen with the backspace, if you insert something like `b'\n'` it keeps it as the same value.
In my case this is breaking any value I am storing in bytes that has the hex value 0x08

Steps to Reproduce:

1. Have an object with a bytes property
```
type Test {
    required property data -> bytes;
} 
```
2. Insert an object with the bytes set as `b'\b'` (this is the ascii backspace which also corresponds to 0x08)
```
insert Test { data := b'\b' };
```
3. Select that inserted object and see that the bytes got changed to `b'\x10'`
```
edgedb> select Test { data };
{default::Test {data: b'\x10'}}
```

msullivan:
Fixes #4534

### DIFF
--- a/edb/edgeql-parser/src/helpers.rs
+++ b/edb/edgeql-parser/src/helpers.rs
@@ -82,7 +82,7 @@ fn _unquote_string<'a>(s: &'a str) -> Result<String, String> {
                                    .to_string())?;
                 match c {
                     c@'"' | c@'\\' | c@'/' | c@'\'' => res.push(c),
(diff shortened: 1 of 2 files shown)


1st1 (approved):

1st1:
Ouch. Can we have this in 2.x please? @elprans

PR #4538, opened 2022-10-17
Add ExprAlias types to public schema

### LINKED ISSUE #4493: ArrayExprAlias, TupleExprAlias, RangeExprAlias are missing from the public schema
This would be pretty unimportant, except that it means that `__type__` on those elements is missing (filter out by access policy). This is bad, since `__type__` is `required` -- we actually have a special-case that opts `__type__` out from access policy induced `assert_exists` insertion, and so it returns invalid data instead of crashing, here.

I think it should just be a matter of adding the types to `schema.edgeql`. The relevant schema type definitions are in `schema/types.py`

```
edgedb> CREATE ALIAS tuple_alias := ('foo', 10);
OK: CREATE ALIAS
edgedb> select schema::Tuple { name, __type__ } filter .name like 'default::%';
{Object {name: 'default::tuple_alias', __type__: {}}}
```

aljazerzen:
Closes #4493.

### DIFF
--- a/edb/lib/schema.edgeql
+++ b/edb/lib/schema.edgeql
@@ -123,6 +123,9 @@ CREATE TYPE schema::Array EXTENDING schema::CollectionType {
 };
 
 
+CREATE TYPE schema::ArrayExprAlias EXTENDING schema::Array;
+
+
 CREATE TYPE schema::TupleElement EXTENDING std::BaseObject {
     CREATE REQUIRED LINK type -> schema::Type {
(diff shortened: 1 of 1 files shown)


elprans (commented):
Why is this needed?  `__type__` of alias results should point to a plain `Type`.  `Alias` objects are technically internal.

msullivan:
The problems are spelled out in the linked issue

msullivan (approved):

PR #4550, opened 2022-10-19
Add/fix fields in `CREATE MIGRATION`

tailhook:
1. `message` field is now reflected in `schema::Migration`
2. Added `generated_by` field to the migration
3. DDL-generated migrations are now marked with `generated_by := MigrationGeneratedBy.DDLStatement`.

### DIFF
--- a/edb/edgeql/parser/grammar/ddl.py
+++ b/edb/edgeql/parser/grammar/ddl.py
@@ -332,15 +332,21 @@ class NestedQLBlock(ProductionTpl):
     def _process_body(self, body):
         fields = []
         stmts = []
+        uniq_check = set()
         for stmt in body:
             if isinstance(stmt, qlast.SetField):
-                if stmt.name in self.allowed_fields:
-                    fields.append(stmt)
-                else:
+                if stmt.name not in self.allowed_fields:
                     raise errors.InvalidSyntaxError(
                         f'unexpected field: {stmt.name!r}',
                         context=stmt.context,
                     )
+                if stmt.name in uniq_check:
+                    raise errors.InvalidSyntaxError(
+                        f'duplicate `SET {stmt.name} := ...`',
+                        context=stmt.context,
+                    )
+                uniq_check.add(stmt.name)
+                fields.append(stmt)
             else:
                 stmts.append(stmt)
 
@@ -2818,7 +2824,7 @@ class CreateMigrationBodyBlock(NestedQLBlock):
 
     @property
     def allowed_fields(self) -> typing.FrozenSet[str]:
-        return frozenset({'message'})
+        return frozenset({'message', 'generated_by'})
 
     @property
     def result(self) -> typing.Any:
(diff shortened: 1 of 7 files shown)


tailhook:
Also I've noted that you can't do `SET message := ` in `START MIGRATION`:
```
edgedb> start migration to { module default { type X; }};
OK: START MIGRATION
OK: POPULATE MIGRATION
edgedb[tx]> set message := 'from started';
error: EdgeQLSyntaxError: Unexpected 'message'
  ┌─ query:1:5
  │
1 │ set message := 'from started';
  │     ^^^^^^^ error

edgedb[tx:failed]> rollback;
```

This is a bug, right? But I'm not sure if it's fixable, as it conflicts with `set {module|global|alias}` syntax a bit.

msullivan:
> This is a bug, right? But I'm not sure if it's fixable, as it conflicts with `set {module|global|alias}` syntax a bit.

It's probably fixable, in this specific case of `set message`; probably not for arbitrary things?

edb/edgeql/parser/grammar/ddl.py
```diff
@@ -2865,22 +2871,44 @@ def reduce_CreateMigration(self, *kids):
...
+                case 'message':
+                    message = f.value
+                case 'generated_by':
+                    generated_by = f.value
+
         self.val = qlast.CreateMigration(
             name=kids[2].val.name,
             parent=kids[2].val.parent,
             body=kids[3].val.body,
+            message=message,
```
msullivan, line 2887: I think that if we populate the `commands` field here with `kids[3].val.fields`, the generic mechanisms for handling `SET field := ...` should pick up the commands and work without needing to have a `message` or `generated_by` field in `qlast.CreateMigration`, without needing to pick them out in the parser, and thus also without needing to handle them in `_cmd_from_ast`

msullivan (commented):
This is all fairly reasonable but I *think* that there is a simpler way that leverages our generic mechanisms for handling `SET`. Try to use that and follow up with me if it doesn't work?

tailhook:
> This is all fairly reasonable but I _think_ that there is a simpler way that leverages our generic mechanisms for handling `SET`. Try to use that and follow up with me if it doesn't work?

It works. Thanks! Feels a bit too magical for someone not familiar enough with the codebase, though.

tailhook:
Oh, by the way, I've just noticed that I've added uniqueness check into the quite generic check. Please double check that it's fine (looks like it's only used in two places, but I'm not sure). Maybe some `SET field` statements should be overridable?

msullivan:
> Oh, by the way, I've just noticed that I've added uniqueness check into the quite generic check. Please double check that it's fine (looks like it's only used in two places, but I'm not sure). Maybe some `SET field` statements should be overridable?

Yeah, that seems right.

msullivan (approved):
Oh I thought I approved this yesterday!

Issue #4350, opened 2022-09-13
Error when running a sequence of queries and returning multiple variables


dotlouis:
- EdgeDB Version: 2.1+52c90a7
- OS Version: macOS 12.5.1 (M1)

Hi all,

Given a schema that looks like this (simplified for readability):

```GraphQL
module default {
  type Submission {
    required link account -> auth::Account
    required link contact -> default::Contact
  }
  type Contact {
    required link account -> auth::Account
    required link sharedContact -> shared::Contact
    link submission := .<contact[is Submission];
  }
}

module shared {
  type Contact {
    property name -> str;
    multi link accountContacts := .<sharedContact[is default::Contact];
    multi link employments := .<contact[is shared::Employment];
  }

  type Employment {
      required link contact -> shared::Contact;
      required link company -> shared::Company
  }

  type Company {
      property name -> str;
      multi link employments := .<company[is shared::Employment];
  }
}
```

I have this query that uses WITH block to create nested objects once, and reuse them in subsequent queries:

```GraphQL
WITH
  account := (global currentAccount),
  sharedContact := (
    INSERT shared::Contact {
      name := "Colin"
    } 
    UNLESS CONFLICT ON shared::Contact.linkedinUrl
    ELSE (shared::Contact)
  ),
  contact := (
    INSERT default::Contact {
      account := account,
      sharedContact := sharedContact
    }
  ),
  submission := (
    INSERT default::Submission {
      account := account,
      contact := contact,
    }
  ),
  sharedCompany := (
    INSERT shared::Company {
      name := "EdgeDB"
    } 
    UNLESS CONFLICT ON shared::Company.linkedinUrl
    ELSE (shared::Company)
  ),
  sharedEmployment := (
    INSERT shared::Employment {
      contact := sharedContact,
      company := sharedCompany
    } 
    UNLESS CONFLICT ON shared::Employment.email
    ELSE (shared::Employment)
  )
select (submission, sharedEmployment)
```

However, this throws a rather obscure:

`InternalServerError: None`

I initially wanted to return only the submission (from the select)
```
select submission
```

But it throws a `QueryError: unused alias definition: 'sharedEmployment'`

So I've tried returning both to work around it with a tuple:
```
select (submission, sharedEmployment)
```
or with a free object:
```
select {
  submission := submission,
  sharedEmployment := sharedEmployement
}
```
Both of these methods throws: `InternalServerError: None`

There is a similar discussion around how to sequence queries and return only parts of them in the response:
[...]

PR #4556, opened 2022-10-19
Fix UNLESS CONFLICT ON for a not-inserted property

msullivan:
We were generating an anytype in our select and then getting confused
on the backend.

Found trying to reproduce #4350.

### DIFF
--- a/tests/test_edgeql_insert.py
+++ b/tests/test_edgeql_insert.py
@@ -3275,6 +3275,16 @@ class TestInsert(tb.QueryTestCase):
             q, [], variables={'n': "2"},
         )
 
+    async def test_edgeql_insert_unless_conflict_26(self):
+        # Test unless conflict on a property not actually mentioned
+        await self.con.execute('''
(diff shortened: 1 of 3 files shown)


tests/test_edgeql_insert.py
```diff
@@ -3275,6 +3275,16 @@ async def test_edgeql_insert_unless_conflict_25(self):
             q, [], variables={'n': "2"},
         )
 
+    async def test_edgeql_insert_unless_conflict_26(self):
+        # Test unless conflict on a property not actually mentioned
```
elprans, line 3279: Would be great to add a variant of this test where we _actually_ get a conflict, e.g when a conflicting value is from a default.

elprans (approved):

tests/test_edgeql_insert.py
msullivan: We have tests for the cases with defaults. This is for when there's *not* a default

PR #4557, opened 2022-10-20
Allow overloading `__type__` with a computed in shapes

### LINKED ISSUE #4552: Allow overloading `__type__` with a computed in shapes
Context:

> https://github.com/edgedb/edgedb-js/issues/326 — do we have to disallow computeds called __type__ in shapes even when they are returning a Type? This works for all other links but not for __type__ and that breaks our querybuilder which is cumbersome to get fixed.
I think part of the reasoning here is purely historical - at one time we used to use __type__ in our REPL. Also we ban __names__ in general but we can relax things here because this name __type__ is already defined.

And

CleanShot 2022-10-19 at 13 26 13@2x

---

We want this to be in the next 2.x release.

cc @jaclarke @msullivan

aljazerzen:
Closes #4552.

Note that this may break cli/clients:

```
edgedb> SELECT Movie {
.......   title,
.......   __type__ := (
.......     SELECT Movie.__type__ {
.......       id,
.......       name
.......     }
.......   )
....... };
{default::Movie {title: 'avengers', __type__: schema::ObjectType {id: 9a4fa88c-4fca-11ed-a72b-778447665de9, name: 'default::Movie'}}}
edgedb> SELECT Movie {
.......   title,
.......   __type__ := (
.......     SELECT Person.__type__ {
.......       id,
.......       name
.......     }
.......   )
....... };
edgedb error: QueryError: possibly an empty set returned by an expression for a computed property '__tname__' declared as 'required'
edgedb>
```


msullivan (approved):
Could you add a test, also?

I think doing dumb stuff with `__type__` will only cause problems when typename injection is on, so it shouldn't cause much client trouble.

1st1:
> I think doing dumb stuff with __type__ will only cause problems when typename injection is on, so it shouldn't cause much client trouble.

Why? Type names are injected via `__tname__`. Ideally whatever "dumb stuff" you do with `__type__` shouldn't disrupt `__tname__`.

1st1:
IOW can we ensure that `__tname__` just always works no matter what's done to `__type__`?

msullivan:
So it looks like the `__tname__` error there is just the one that comes up first, and even without type injection that example will produce errors because the cardinality isn't right: `__type__` is `single required`.
I tried this variant:
```
scope> SELECT User {
......   name,
......   __type__ := (introspect Card) { id, name },
...... };
```
which doesn't produce any errors, but does result in putting the wrong typename in the output:
```
  default::Card {
    name: 'Alice',
    __type__: schema::ObjectType {
      id: e98e931f-4f15-11ed-8887-45ec6913f825,
      name: 'default::Card',
    },
  },
```

Let me see if I can fix the typename thing. Typename injection is pretty fiddly, though, so we'll see.

1st1:
(And if we don't have a test that it's prohibited to overload `__id__`, `__tname__`, and `__tid__` we need to add it. We only want to allow `__type__` because we don't rely on it in REPL or elsewhere.

msullivan:
> (And if we don't have a test that it's prohibited to overload `__id__`, `__tname__`, and `__tid__` we need to add it. We only want to allow `__type__` because we don't rely on it in REPL or elsewhere.

Yeah we ban all the dunder names except `__type__` I think. (And `__type__` probably should have just be named `type`!)

1st1:
> And __type__ probably should have just be named type

We didn't want to shadow 'type' property/link name. So we went with a special dunder name for that, which I think is OK-ish approach

PR #4558, opened 2022-10-20
Fix inheritance code block example

pokedotdev:
The code block was missing the definition of the abstract type and the extension
[https://www.edgedb.com/docs/edgeql/sets#inheritance](https://www.edgedb.com/docs/edgeql/sets#inheritance)

image


1st1 (approved):

1st1:
Thank you!

PR #4565, opened 2022-10-21
Hack around the test flakes in dump tests

msullivan:
The problem is that `aclose` can return before `DROP DATABASE` will
succeed.  Work around this by sticking `DROP DATABASE` in a retry
loop.

I tested it by replacing `con2.aclose()` with
```
async def asdf():
    await asyncio.sleep(1)
    await con2.aclose()

asyncio.create_task(asdf())
```
This consistently failed before introducing the retry loop, which
fixed it.

### DIFF
--- a/edb/testbase/server.py
+++ b/edb/testbase/server.py
@@ -1283,7 +1283,15 @@ class DumpCompatTestCaseMeta(TestCaseMeta):
             finally:
                 self.__class__.con = oldcon
                 await con2.aclose()
-                await self.con.execute(f'DROP DATABASE {qdbname}')
+
+                # The connection might not *actually* be closed on the db
+                # side yet. This is probably a bug, but hack around it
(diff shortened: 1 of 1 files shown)


elprans (changes requested):
No, it's a bug with the `DROP DATABASE` implementation.  `aclose()` actually waits until its connection falls off.  I'll look into fixing this properly shortly and would rather not paper it over here to make sure that the fix works.

msullivan:
If you think a fix for the underlying bug is forthcoming pretty soon, we can skip this. But this bug does cause of a lot of pain (frequent flakes, release builds failing) and it would be nice to not have to deal with it when we release 2.6, for example.

If we do merge this workaround, we should certainly revert it when we actually fix the bug

elprans (approved):
OK, let's merge and revert when fixed properly.

PR #4568, opened 2022-10-22
Fix breakage due to --use-db-cache

msullivan:


elprans (approved):

elprans:
Remove the other `cache_file` though.

PR #4573, opened 2022-10-25
Make patches able to make modifications to schema layout

msullivan:
To do this, after applying any script component, we need to rerun
generate_structure, run any SQL it generates, and update all the
cached values it produces.

This also requires upgrading the regular edgeql script application
path to update reflschema and stdschema in sync and making
generate_structure able to run on an existing reflschema.

### DIFF
--- a/edb/server/bootstrap.py
+++ b/edb/server/bootstrap.py
@@ -608,10 +608,9 @@ def prepare_patch(
         # We need to delete all the support views and recreate them at the end
         support_view_commands = metaschema.get_support_views(
             reflschema, backend_params)
-        for cv in support_view_commands:
+        for cv in reversed(list(support_view_commands)):
             dv = dbops.DropView(
                 cv.view.name,
-                cascade=True,
                 conditions=[dbops.ViewExists(cv.view.name)],
             )
             dv.generate(preblock)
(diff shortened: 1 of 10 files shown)


edb/server/bootstrap.py
```diff
@@ -500,61 +536,137 @@ def prepare_patch(
...
+        ))
+
+        # This part is wildly hinky
+        # We need to delete all the support views and recreate them at the end
+        support_view_commands = metaschema.get_support_views(
+            reflschema, backend_params)
+        for cv in support_view_commands:
+            dv = dbops.DropView(
+                cv.view.name,
+                cascade=True,
```
elprans, line 614: Any way we can avoid `cascade` here?  It's pretty dangerous and can reach into parts that we didn't anticipate, hide bugs etc.

msullivan: For some reason, when I was thinking about this yesterday, figuring out the order to drop the views in seemed like it would be hard.

It obviously was not.

elprans (approved):

msullivan:
(Obviously none of this got tested on this branch on master, so testing mostly happened here https://github.com/edgedb/edgedb/pull/4582)

Issue #4564, opened 2022-10-21
"too many range table entries" on nested filtering in the presence of Access Policies


jackfischer:
- EdgeDB Version: 3.0-dev.7000+58a0109 
- OS Version: macOS 12.6

Steps to Reproduce:

1. Perform a highly nested select/filter with access policies enabled (example below). We are seeing this on other queries as well. No inserted data required.
`edgedb error: InvalidValueError: too many range table entries`

To restore normal edgedb behavior, disable access policies via `configure session set apply_access_policies := false`

QB

```
e.select(e.StrategyNode, node => ({
    tiptapId: true,
    metrics: metric => ({
      predictions: e.select(metric.predictingQuestion.responses, response => ({
        content: true,
        createdDatetime: true,
        limit: 1,
        order_by: {
          expression: response.createdDatetime,
          direction: e.DESC,
          empty: EMPTY_LAST,
        },
        target: e.select(metric.metricTargets, target => ({
          id: true,
          targetValue: true,
          filter: e.op(target.strategyNode, "=", e.select(e.StrategyNode, node => ({
            filter: e.op(node.tiptapId, "=", node.tiptapId),
          }))),
        })),
        tiptapId: node.tiptapId,
      })),
    }),
  }));
```

Generated EdgeQL

```
WITH
  __scope_0_defaultStrategyNode := DETACHED default::StrategyNode
SELECT __scope_0_defaultStrategyNode {
  tiptapId,
  metrics := (
    WITH
      __scope_1_defaultMetric := __scope_0_defaultStrategyNode.metrics
    SELECT __scope_1_defaultMetric {
      multi predictions := (
        WITH
          __scope_2_defaultSignal := __scope_1_defaultMetric.predictingQuestion.responses
        SELECT __scope_2_defaultSignal {
          content,
          createdDatetime,
          multi target := (
            WITH
              __scope_3_defaultMetricTarget := __scope_1_defaultMetric.metricTargets
            SELECT __scope_3_defaultMetricTarget {
              id,
              targetValue
            }
            FILTER (__scope_3_defaultMetricTarget.strategyNode = (WITH
              __scope_4_defaultStrategyNode := DETACHED default::StrategyNode
            SELECT __scope_4_defaultStrategyNode {
              id
            }
            FILTER (__scope_4_defaultStrategyNode.tiptapId = __scope_0_defaultStrategyNode.tiptapId)))
          ),
          single tiptapId := __scope_0_defaultStrategyNode.tiptapId
[...]

PR #4574, opened 2022-10-25
Add a 'future' system and a 'nonrecursive_policies' future

msullivan:
A 'feature' is a configuration flag that lives inside the database
itself, and can control major questions of language behavior that can
impact compilation (for example, nonrecursive_policies affects the
compilation of access policies in a way that can affect function
bodies). Features are explicitly built in to the language, and are set
up so that they can trigger other schema changes when they are
toggled.

The syntax is basically the same as for extensions. (The core
difference between features and extensions is that extensions ought to
be self-contained and not change the meaning of language features!)

The intent is to use features somewhat like python's
"from __future__ import ...".

Then, implement a 'nonrecursive_policies' feature.  This feature flag
disables the evaluation of policies within policies themselves. This
is probably what the default behavior should have been, and it is
important to provide this behavior ASAP, but it would be unreasonable
to unconditionally change the behavior in a point release.
See #4564 for one place where the current behavior has made a use-case
impossible.

I'd be thrilled if someone had a better name than nonrecursive_policies.

We will be able to cherry-pick this to 2.x using #4573.


1st1:
Please wait for my review

elprans (commented):
Implementation looks good to me, though we probably still want to bikeshed on the name a bit, e.g. it was suggested that we may want to use `future` or `future feature` to stress that this mechanism is intended for smooth deprecation of schema features (or bugs) and not another sort of extension-like or config-like mechanism.

1st1:
The PR looks good but I'd like to bikeshed about naming. I don't like `using feature xxx` because "feature" implies some kind of permanent knob. But we do want to drop these "features" eventually. Hence I'm -1 on having the "using feature" naming.

I like `using future nonrecursive_policies` or `using future simple_policies` more. The syntax suggests that it flips some behavior to what it will be in the future.

zackelan:
shim? [well-established term](https://en.wikipedia.org/wiki/Shim_(computing)) that I think has the connotations we want

and I like "simple" or "basic" policies over nonrecursive. maybe "extended" instead of recursive?

elprans (approved):

PR #4585, opened 2022-10-27
Implement MIGRATION REWRITE system

### LINKED ISSUE #4583: Implement START MIGRATION REWRITE
As described in https://github.com/edgedb/edgedb/discussions/4562

msullivan:
This allows the migration history (recorded in schema::Migration)
to be rewritten, as long as the final result matches the target.

Closes #4583.

I am going to add tests, but I've tested it manually and I think it
works well enough for @tailhook to start building on top of it.


edb/server/compiler/compiler.py
```diff
@@ -1410,38 +1472,211 @@ def _compile_ql_migration(
...
+                savepoint_name = current_tx.start_migration()
+                query = dbstate.MigrationControlQuery(
+                    sql=(b'SELECT LIMIT 0',),
+                    action=dbstate.MigrationAction.START,
+                    tx_action=None,
+                    cacheable=False,
+                    modaliases=None,
+                )
+
+            # Start from an empty schema except for `module default`
```
elprans, line 1580: Hm, why special treatment for `module default`?

edb/server/compiler/compiler.py
```diff
@@ -1410,38 +1472,211 @@ def _compile_ql_migration(
...
+            self._assert_not_in_migration_block(ctx, ql)
+            mrstate = self._assert_in_migration_rewrite_block(ctx, ql)
+
+            diff = s_ddl.delta_schemas(schema, mrstate.target_schema)
+            if list(diff.get_subcommands()):
+                if debug.flags.delta_plan:
+                    debug.header("COMMIT MIGRATION REWRITE mismatch")
+                    diff.dump()
+                raise errors.QueryError(
+                    'cannot commit migration rewrite: schemas do not match',
```
elprans, line 1616:
```suggestion
                    'cannot commit migration rewrite: schema resulting from rewrite does not match committed schema',
```

edb/server/compiler/compiler.py
```diff
@@ -927,6 +959,31 @@ def _compile_and_apply_ddl_stmt(
                 single_unit=False,
             )
 
+        # If we are in a migration rewrite, we also don't actually
+        # apply the DDL, just record it. (The DDL also needs to be a
+        # CreateMigration.)
+        if mrstate := current_tx.get_migration_rewrite_state():
+            if not isinstance(stmt, qlast.CreateMigration):
```
elprans, line 966: Not necessary in this iteration, but in principle we can allow any DDL statements and automatically collect them into a migration.

elprans (approved):
LGTM.  Please add tests.

edb/server/compiler/compiler.py
msullivan: What typically happens in practice is that a lone DDL statement will get automatically put into its own migration

edb/server/compiler/compiler.py
msullivan: A newly created database starts off containing a `module default`, but an empty `FlatSchema()` doesn't.

tailhook:
Tiny issue so far:
```
edgedb> start migration rewrite;
OK: START MIGRATION REWRITE
edgedb[tx]> fail;
error: EdgeQLSyntaxError: Unexpected 'fail'
  ┌─ query:1:1
  │
1 │ fail;
  │ ^^^^ error

edgedb[tx:failed]> abort migration rewrite;
edgedb error: TransactionError: current transaction is aborted, commands ignored until end of transaction block
```
(works with `start migration/abort migration`)

PR #4587, opened 2022-10-27
Allow JWT token auth in binary protocol

fantix:
The token should be included in the ClientHandshake message, under a new connection parameter named "token".

### DIFF
--- a/edb/server/protocol/binary.pyx
+++ b/edb/server/protocol/binary.pyx
@@ -494,7 +494,7 @@ cdef class EdgeConnection(frontend.FrontendConnection):
             # token in the HTTP header has higher priority than
             # the ClientHandshake message, under the scenario of
             # binary protocol over HTTP
-            token = self._extract_token_in_auth_data()
+            token = self._extract_token_from_auth_data()
             if token is None:
                 token = params.get('token')
             self._auth_jwt(user, token)
@@ -618,7 +618,7 @@ cdef class EdgeConnection(frontend.FrontendConnection):
         if user not in roles:
             raise errors.AuthenticationError('authentication failed')
 
-    def _extract_token_in_auth_data(self):
+    def _extract_token_from_auth_data(self):
         if not self._auth_data:
             raise errors.AuthenticationError(
                 'authentication failed: no authorization data provided')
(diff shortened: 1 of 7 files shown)


edb/server/protocol/binary.pyx
```diff
@@ -606,20 +608,23 @@ cdef class EdgeConnection(frontend.FrontendConnection):
...
         role = self.server.get_roles().get(user)
         if role is None:
             raise errors.AuthenticationError('authentication failed')
 
-        if not self._auth_data:
-            raise errors.AuthenticationError(
-                'authentication failed: no authorization data provided')
+        decrypt = False
+        if not encoded_token:
+            decrypt = True  # token in binary protocol over HTTP is encrypted
```
elprans, line 618: Let's just drop encryption of JWT and normalize this (i.e. check for `token` in `params` and then in `self._auth_data`)

elprans (commented):
Let's drop encryption of JWTs and password passing in tokens.  Correct token signature + role claims is sufficient for authentication.

edb/server/protocol/binary.pyx
```diff
@@ -469,6 +469,11 @@ cdef class EdgeConnection(frontend.FrontendConnection):
                 f'message: "database"'
             )
 
+        # token in the HTTP header has higher priority than the ClientHandshake
+        # message, under the binary-protocol-over-HTTP scenario
+        if self._auth_data is None:
+            self._auth_data = params.get('token')
+
```
elprans, line 476: Change to:

```suggestion
        token = self._auth_data
        if token is None:
            token = params.get('token')
```

and move under `authmethod_name == 'JWT':` to avoid clobbering `_auth_data` for cases when auth is not JWT

edb/server/protocol/binary.pyx
```diff
@@ -491,7 +491,13 @@ cdef class EdgeConnection(frontend.FrontendConnection):
         if authmethod_name == 'SCRAM':
             await self._auth_scram(user)
         elif authmethod_name == 'JWT':
-            self._auth_jwt(user)
+            # token in the HTTP header has higher priority than
+            # the ClientHandshake message, under the scenario of
+            # binary protocol over HTTP
+            token = self._extract_token_in_auth_data()
```
elprans, line 497:
```suggestion
            token = self._extract_token_from_auth_data()
```

elprans (approved):

PR #4604, opened 2022-11-01
Document `future`.

vpetrovykh:
Document what the `future` specification is and how it's intended to be used. Add the `nonrecursive_access_policies` future behavior to the docs.

Explain the current access policy behavior and the behavior we expect in EdgeDB 3.0.

### DIFF
--- a/docs/datamodel/access_policies.rst
+++ b/docs/datamodel/access_policies.rst
@@ -268,7 +268,7 @@ policies. This means that they can affect each other in various ways. Because
 of this great care needs to be taken when creating access policies based on
 objects other than the ones they are defined on. For example:
 
-.. code-block:: sdl-diff
+.. code-block:: sdl
 
     global current_user_id -> uuid;
     global current_user := (
@@ -277,11 +277,11 @@ objects other than the ones they are defined on. For example:
 
     type User {
       required property email -> str { constraint exclusive; };
-      required property is_admin -> bool;
+      required property is_admin -> bool { default := false };
 
       access policy admin_only
         allow all
-        using (global current_user.is_admin);
+        using (global current_user.is_admin ?? false);
     }
 
     type BlogPost {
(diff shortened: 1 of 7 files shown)


docs/datamodel/access_policies.rst
```diff
@@ -263,6 +263,59 @@ algorithm for resolving these policies.
...
+        using (global current_user ?= .author.id);
+    }
+
+In the above schema only the admin will see a non-empty ``author`` link,
+because only the admin can see any user objects at all. This means that
+instead of making ``BlogPost`` visible to its author, all non-admin authors
+won't be able to see their own posts. The above issue can be remedied by
+making the current user able to see their own ``User`` record.
+
+.. note::
```
1st1, line 302: Maybe warning?

vpetrovykh: Fair enough. I changed it to "warning".

msullivan (approved):
I'm happy with this, and also made a few tweaks.

(I had originally just meant to fix the error and do some very light copy editing, but wound up tweaking the content slightly)

docs/datamodel/access_policies.rst
```diff
@@ -263,6 +263,65 @@ algorithm for resolving these policies.
...
+
+  Starting with the upcoming EdgeDB 3.0, access policy restrictions will
+  **not**
+  apply to any access policy expression. This means that when reasoning about
+  access policies it is no longer necesary to take other policies into
+  account. Instead, all data is visible for the purpose of *defining* an access
+  policy.
+
+  This change is being made to simplify reasoning about access
+  policies and to allow certain patterns to be express
```
raddevon, line 316: express -> expressed

docs/datamodel/future.rst
```diff
@@ -0,0 +1,49 @@
...
+Future Behavior
+===============
+
+Any time that we add new functionality to EdgeDB we strive to do it in the
+least disruptive way possible. Deprecation warnings, documentation and guides
+can help make these transiotions smoother, but sometimes the changes are just
+too big, especially if they affect already existing functionality. It is often
+inconvenient dealing with these changes at the same time as upgrading to a new
+major version of EdgeDB. To help with this transition we introduce
+:ref:`future <ref_eql_sdl_future>` specification.
```
raddevon, line 13: I would add an article here: "**the** future specification"

docs/reference/ddl/future.rst
```diff
@@ -0,0 +1,81 @@
...
+
+
+Description
+-----------
+
+The command ``drop future`` disables a currently active future behavior for
+the current database. However, this is only possible for versions of EdgeDB
+when the behavior in question is not officially introduced. Once a particular
+behavior is introduced as the standard behavior in an EdgeDB release, it
+cannot be disabled. Running this command will simply denote that no special
```
raddevon, line 62: > Running this command will simply denote that no special action is needed to enable it in this case.

I have a little trouble following this. Are you saying here that running `drop future` with a behavior that is now standard in your current release will have no effect but will tell you that the feature does not need to be enabled? If so, I'd propose:

> Running ``drop future`` in this scenario will have no impact on the database but will inform you that \<whatever it informs>

Also, I read your sentence to say that, when I try to *dis*able the feature, I'm going to get back a message telling me that I don't need to do anything to *en*able it. Is that accurate?

PR #4605, opened 2022-11-01
Fix collection type using DDL and set global in notebook protocol

msullivan:
The collection type in DDL fixes some issues we've had with updating
the tutorial to use the latest edgedb, where a function taking
variadic args failed.

The approach to fixing it was to use the execute module instead of
trying to handle things directly.

Since this got us most of the way to having globals work in the
tutorial, I also got that working. The main snag there was the
capabilities system, which treats SET GLOBAL as a session
configuration command. This is true, but probably too strict: add a
new capabilities flag for SET GLOBAL, but make it implied by the
session config command, so as to not break protocol compatability.

### DIFF
--- a/edb/server/compiler/compiler.py
+++ b/edb/server/compiler/compiler.py
@@ -2015,7 +2015,7 @@ class Compiler:
             if ql.scope is qltypes.ConfigScope.SESSION:
                 capability = enums.Capability.SESSION_CONFIG
             elif ql.scope is qltypes.ConfigScope.GLOBAL:
-                capability = enums.Capability.SESSION_CONFIG
+                capability = enums.Capability.SET_GLOBAL
             else:
                 capability = enums.Capability.PERSISTENT_CONFIG
             return (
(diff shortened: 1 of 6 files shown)


edb/server/protocol/notebook_ext.pyx
```diff
@@ -33,8 +33,10 @@ from edb.common import markup
 from edb.server import compiler
 from edb.server import defines as edbdef
 from edb.server.compiler import OutputFormat
+from edb.server.compiler import dbstate
 from edb.server.compiler import enums
-from edb.server.protocol cimport args_ser
+# from edb.server.protocol cimport args_ser
```
elprans, line 38:
```suggestion
```

edb/server/compiler/compiler.py
```diff
@@ -2015,7 +2015,7 @@ def _compile_dispatch_ql(
             if ql.scope is qltypes.ConfigScope.SESSION:
                 capability = enums.Capability.SESSION_CONFIG
             elif ql.scope is qltypes.ConfigScope.GLOBAL:
-                capability = enums.Capability.SESSION_CONFIG
+                capability = enums.Capability.SET_GLOBAL
```
elprans, line 2018: Oops

elprans (approved):

PR #4609, opened 2022-11-01
minor doc updates and fixes, schema chapter

zackelan:


Sikarii (approved):

docs/datamodel/access_policies.rst
```diff
@@ -341,9 +340,6 @@ the author.
 
 "Disappearing" posts that become invisible after 24 hours.
 
-Blog posts are publicly visible except to users that have been ``blocked`` by
-the author.
-
```
raddevon, line 346: What's the rationale for removing this?

docs/datamodel/computeds.rst
```diff
@@ -68,7 +68,8 @@ next time you try to :ref:`create a migration <ref_intro_migrations>`.
 
   type Person {
     property first_name -> str;
-    required property name_upper := str_upper(.name); # invalid
+    # invalid, because first_name is not a required property
```
raddevon, line 71: Could we add something to indicate what we're talking about? Maybe the pointing down emoji (👇 )? Not sure if that would be appropriate in this context, but I think it would make it less ambiguous.

raddevon (approved):
I suggested a few very minor tweaks, but this looks great!

docs/datamodel/access_policies.rst
zackelan: Copy-paste error, see line 317 above

PR #4616, opened 2022-11-02
Only allow uuid generation functions as default values for id

msullivan:
Update some of the uuid documentation

### DIFF
--- a/edb/schema/pointers.py
+++ b/edb/schema/pointers.py
@@ -1649,7 +1649,7 @@ class PointerCommand(
         from edb.ir import utils as irutils
 
         # If we add more, we probably want a better mechanism
-        ID_WHITELIST = (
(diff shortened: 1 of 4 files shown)


edb/schema/pointers.py
```diff
@@ -1633,6 +1638,42 @@ def validate_object(
...
+        schema: s_schema.Schema,
+        context: sd.CommandContext,
+        expr: irast.Base,
+    ) -> None:
+        """If default is being set on id, check it against a whitelist"""
+        from edb.ir import ast as irast
+        from edb.ir import utils as irutils
+
+        # If we add more, we probably want a better mechanism
+        ID_WHITELIST = (
```
elprans, line 1652:
```suggestion
        ID_ALLOWLIST = (
```

elprans (approved):

PR #4622, opened 2022-11-03
Stop relying on hacks in ordering compiler enums.

vpetrovykh:
Multiplicity, cardinality and volatility enums have an intrinsic order that is relevant for the compiler. The enum definition order is now used to determine how the `<` comparison works for various enum values.


msullivan (approved):
Great cleanup

PR #4626, opened 2022-11-03
fix: indexing of an empty set

aljazerzen:
With #4479, stdlib's array indexing has been made non strict, which makes this produce an error:
```
select re_match(r"foo", 'other name')[0];
```
instead of an empty set.

This PR fixes that and adds tests for all kinds of indexing and slicing or empty sets.

### DIFF
--- a/edb/buildmeta.py
+++ b/edb/buildmeta.py
@@ -44,7 +44,7 @@ from edb.common import verutils
 
 
 # Increment this whenever the database layout or stdlib changes.
-EDGEDB_CATALOG_VERSION = 2022_11_01_00_00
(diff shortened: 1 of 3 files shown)


msullivan (approved):
Bump the catalog version too?

Could you try benchmarking it also, with a query that does a lot of indexing into an array? If it's a perf problem we'll need to rework

aljazerzen:
I ran the following benchmark and this commit has no effect on it.

```
time edb cli query "select (for i in _gen_series(1, 100000000) union (select [1, 2, 3][2])) limit 3;"
```

Issue #4567, opened 2022-10-21
DROP DATABASE can fail even after last connection is closed


msullivan:
https://github.com/edgedb/edgedb/pull/4565 hacks around this to prevent some test flakes, but it is worth fixing the bug.

We should revert https://github.com/edgedb/edgedb/pull/4565 when we do so.

msullivan:
`_ensure_database_not_connected` actually does have a retry loop that waits for all the connections to be dropped. I strongly suspect that the issue is that the connection pool is reestablishing connections

PR #4629, opened 2022-11-03
Put all of the DROP DATABASEs in the test suite behind a retry loop

msullivan:
This works around #4567.


elprans (approved):

msullivan:
@elprans is the description of the problem in the comment still accurate?

PR #4631, opened 2022-11-04
Add `EXCEPT` and `INTERSECT` set operators.

### LINKED ISSUE #4392: Add set intersection and difference operators
The hardest part of this is probably figuring out syntax/naming, since we probably don't want to add any reserved keywords.

We can use either symbol operators `&` and `\` or just function names.

```
select set_intersection({1,2,3}, {2,3,4}} == {2,3}
select set_difference({1,2,3}, {2,3,4}} == {1}
```
the implementation can be defined straightforwardly in terms of select:
```
set_intersection(X, Y) = select x := X filter x in (select Y)
set_difference(X, Y) = select x := X filter x not in (select Y)
```

vpetrovykh:
The `EXCEPT` and `INTERSECT` operators serve to complete the basic multiset functionality. They operate on multisets taking element multiplicty into account and so they cannot be easily substituted by a `FILTER` clause.

Fixes #4392

### DIFF
--- a/edb/pgsql/compiler/relgen.py
+++ b/edb/pgsql/compiler/relgen.py
@@ -1485,6 +1485,8 @@ def process_set_as_membership_expr(
 
 
 @_special_case('std::UNION')
+@_special_case('std::EXCEPT')
+@_special_case('std::INTERSECT')
 def process_set_as_setop(
     ir_set: irast.Set, *, ctx: context.CompilerContextLevel
 ) -> SetRVars:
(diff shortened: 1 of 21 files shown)


edb/pgsql/compiler/relgen.py
```diff
@@ -1535,6 +1535,108 @@ def process_set_as_setop(
     return new_stmt_set_rvar(ir_set, ctx.rel, ctx=ctx)
 
 
+@_special_case('std::EXCEPT')
```
msullivan, line 1538: I think all three of these are the same except for the value of `subqry.op`, right? Can we have them all be handled by `process_set_as_setop` and just inspect the call to decide which `op` to use?

vpetrovykh: Probably. I copy-pasted them in the beginning because it was easier to separate debugging process without messing up `UNION`.

vpetrovykh: OK, I've folded it back into one function for all 3 operators.

msullivan (approved):
Seems reasonable. Could you add a comment somewhere laying out how the different keyword distinctions behave?

Excited to have these operators! (Especially except)

msullivan:
Are we good to merge this?

PR #4640, opened 2022-11-07
Add example of TS/JS transaction w/ querybuilder

raddevon:
A user created a repo comparing a task model implementation in Knex and EdgeDB. User was under the impression it was not possible to use the querybuilder with transactions. This is part of an effort to better communicate that it is possible.

See https://discord.com/channels/841451783728529451/1038780233458909324/1039094288354652180

Companion to https://github.com/edgedb/edgedb-js/pull/482

### DIFF
--- a/docs/edgeql/transactions.rst
+++ b/docs/edgeql/transactions.rst
@@ -42,10 +42,23 @@ under the hood.
 TypeScript/JS
 ^^^^^^^^^^^^^
 
+Using an EdgeQL query string:
+
 .. code-block:: typescript
 
   client.transaction(async tx => {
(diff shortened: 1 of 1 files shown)


docs/edgeql/transactions.rst
```diff
@@ -42,12 +42,25 @@ under the hood.
...
+
+.. code-block:: typescript
+
+  const query = e.insert(e.Fish, {
+    name: 'Wanda'
+  });
+  client.transaction(async tx => {
+    await query.run(tx);
+  });
+
```
1st1, line 63: I think you should also take a look at edgedb-js docs specifically and see if we can improve them too

PR #4650, opened 2022-11-09
SQL resolver

### LINKED ISSUE #4540: SQL name resolution
For the upcoming SQL support, we essentially need to implement a basic SQL name resolver. We will not support all of the SQL, but we need following features:

(using the Movies schema)
```sql
-- map Movie to our ObjectType
-- map Movie.title to the correct property
SELECT Movie.title FROM Movie;

-- figure out that property title comes from ObjectType Movie
SELECT title FROM Movie;

-- case insentive by default, lowercase when strictly-typed
SELECT tItLe, `relase_year` FROM `movie`;

-- table aliases
SELECT m.title FROM Movie AS m;

-- resolve <link>_id
SELECT title FROM Movie JOIN Genre ON Movie.genre_id = Genre.id

-- resolve <parent_type>_id
SELECT c.title, m.release_year FROM Content c JOIN Movie m ON m.content_id = c.id

-- infer subquery types
SELECT title, t.name FROM (SELECT *, title as name FROM Movie) t JOIN Genre ON ...
```

## Inheritance

We have to decide what to do when selecting from a parent object. Should rows of child tables be included?

@1st1 says no, because SQL is low level and we should expose only this low-level data.

@msullivan says that we should at least provide some syntax to allow access to the inheriting views we already have so the SQL interface is not such a pain in the ass to use.

I think that we should avoid having two different syntaxes (or any kind of table name suffixes), because this adds much more surface for bugs and confusion. 

While it would be convenient to expose inheriting views, the exposed database would not be normalized so I agree with Yury to stick only to base tables.

## Names of multi-multi link tables

```
type Movie {
    multi link actors -> Person {
      property character_name -> str;
    };
}
type Person {
    link filmography := .<actors[is Content];
}
```

These tables do not have a name in EdgeQL, but need one in SQL. I suggest `movie_actors_person_filmography`:
- object name where link is defined,
- link name on that object
- object name where link is referenced,
- link name on that object.

aljazerzen:
Closes #4540 

Also a few fixes to `pgsql.ast` and `pgsql.parser`


tests/schemas/movies.esdl
```diff
@@ -1,7 +1,7 @@
 #
 # This source file is part of the EdgeDB open source project.
 #
-# Copyright 2016-present MagicStack Inc. and the EdgeDB authors.
```
msullivan, line 4: Go home `git diff`, you are drunk.

(I think it was such a small file that the copyright notice made it think it was the same.)

edb/testbase/server.py
```diff
@@ -1257,6 +1257,17 @@ async def squery(self, query):
         ''')
         return await self.scon.fetch(rewritten)
 
+    async def squery_values(self, query):
+        res = await self.squery(query)
+        return [list(r.values()) for r in res]
+
+    def assertShape(self, res, rows, cols, column_names=None):
```
msullivan, line 1264: Add a docstring for this

I would also say call it `assert_shape`, except that `assertShape` matches other assert methods. Your call.

a.py
```diff
@@ -0,0 +1,32 @@
+from edb.common import devmode
```
msullivan, line 1: Drop this

edb/pgsql/resolver/context.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+class ResolverContextLevel(compiler.ContextLevel):
+    schema: s_schema.Schema
+    names: NameGenerator
+
+    """Visible names in scope"""
+    scope: Scope
+
+    """Current table"""
```
msullivan, line 117: I think this description is wrong?

msullivan (commented):
I didn't wind up having time to review this today. I would like some more documentation---particularly, high-level documentation of the approach. I suspect that the heart of the approach is the data structures and how they are maintained, but I'm not sure

tests/schemas/movies.esdl
aljazerzen: Well, it is doing what it is supposed to. The problem is that the majority of the file is copyright.

edb/pgsql/resolver/resolver.py
```diff
@@ -0,0 +1,645 @@
...
+
+
+@dispatch._resolve.register
+def resolve_SelectStmt(
+    stmt: pgast.SelectStmt, *, ctx: Context
+) -> pgast.SelectStmt:
+
+    # UNION
+    if stmt.larg:
+        with ctx.isolated() as subctx:
```
msullivan, line 47: Should we be doing some column inference in the sides of a union?

edb/pgsql/resolver/resolver.py
```diff
@@ -0,0 +1,645 @@
...
+    alias: pgast.Alias,
+    *,
+    ctx: Context,
+) -> pgast.JoinExpr:
+
+    with ctx.empty() as subctx:
+        larg = dispatch.resolve(range_var.larg, ctx=subctx)
+        l_rel = subctx.scope.rel
+
+    with ctx.empty() as subctx:
```
msullivan, line 228: I think that with `LATERAL` joins the join arguments can reference earlier relations, so in the case we don't want an empty relation

edb/pgsql/resolver/resolver.py
```diff
@@ -0,0 +1,645 @@
...
+    cte = next((t for t in ctx.scope.ctes if t.name == relation.name), None)
+    if cte:
+        ctx.scope.rel.name = cte.name
+        ctx.scope.rel.columns = cte.columns.copy()
+        return pgast.Relation(name=cte.name, schemaname=None)
+
+    # lookup the object in schema
+    obj: Optional[s_objtypes.ObjectType] = None
+    if (relation.schemaname or 'public') == 'public':
+        object_type_name = relation.name[0].upper() + relation.name[1:]
```
msullivan, line 282: What's going on here?

edb/pgsql/resolver/resolver.py
```diff
@@ -0,0 +1,645 @@
...
+
+        obj = ctx.schema.get(  # type: ignore
+            object_type_name,
+            None,
+            module_aliases={None: 'default'},
+            type=s_objtypes.ObjectType,
+        )
+
+        if not obj:
```
msullivan, line 291: We also need a way to reference link tables

msullivan (commented):
Looking pretty good

msullivan:
Oh, one other thing I was wondering is if it would be nicer to directly return the `rel` Table instead of having it get implicitly returned via the context? There aren't that many places that need to handle that, I think, so it might be more explicit without needing to be that much more verbose. I'm not sure about this, though

edb/pgsql/resolver/context.py
```diff
@@ -0,0 +1,174 @@
...
+    rel: Table
+    "Current relation where ResTargets are added to"
+
+    join_relations: List[Table]
+    "Current relation where Join relations are added to"
+
+    tables: List[Table]
+    """Tables visible in this scope"""
+
+    ctes: List[CTE]
```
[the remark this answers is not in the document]
aljazerzen: Yeah, they could but then I would we duplicating stored table name in `Table.name` and as key in `tables`. And tables would have to be stored under two keys, because table can be referred with its table name and also with an alias.

Having it as a list is not really slower in most cases, because number of tables will very rarely exceed 3.

edb/pgsql/resolver/resolver.py
```diff
@@ -0,0 +1,645 @@
...
+
+
+@dispatch.resolve_range_var.register
+def resolve_RangeFunction(
+    range_var: pgast.RangeFunction,
+    alias: pgast.Alias,
+    *,
+    ctx: Context,
+) -> pgast.RangeFunction:
+    raise errors.UnsupportedFeatureError('range functions are not supported')
```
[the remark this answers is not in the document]
aljazerzen: Result of range function may be a table with named columns, depending on the function called. So we would need to hardcode resulting Table for each range function. It's just that.

edb/pgsql/resolver/resolver.py
```diff
@@ -0,0 +1,645 @@
...
+
+        result = context.CTE()
+        result.name = cte.name
+        result.columns = []
+
+        cols_query = subctx.scope.rel.columns
+        cols_names = cte.aliascolnames
+
+        if cols_names and len(cols_query) != len(cols_names):
+            raise errors.QueryError(
```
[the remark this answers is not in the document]
aljazerzen: I've included the context, but it doesn't work that well, because it reports location within the SQL query, not the whole EdgeQL query:

```
error: QueryError: unknown object `MessaGe`
  ┌─ query:1:15
  │
1 │ describe alter 'SELECT * FROM "messaGe"';
  │               ^ error
```

But this will work ok, when SQL will be parsed directly.

edb/pgsql/resolver/resolver.py
aljazerzen: Ok, I've implemented case insensitive lookup we talked about, but there is still a possible problem.

Our ObjectTypes are now exposed as tables with their name converted to lower case, because we want to refer to tables with ident without quotes. But if you have two different ObjectTypes with same name except for casing (`default::User` and `default::user`) they will be exposed as two tables with the same name.

edb/pgsql/resolver/relation.py
```diff
@@ -0,0 +1,254 @@
...
+            continue
+
+        if p.get_cardinality(ctx.schema).is_single():
+            # single links only for tables with at least one property
+            # besides source and target
+            l_pointers = p.get_pointers(ctx.schema).objects(ctx.schema)
+            if len(l_pointers) <= 2:
+                continue
+        return p
+    return None
```
aljazerzen, line 227: @msullivan I'm not sure if this is the best way to do a lookup for a link of an object.

edb/pgsql/resolver/relation.py
```diff
@@ -0,0 +1,222 @@
...
+    if stmt.larg or stmt.rarg:
+        assert stmt.larg and stmt.rarg
+
+        with ctx.isolated() as subctx:
+            larg, ltable = dispatch.resolve_relation(stmt.larg, ctx=subctx)
+
+        with ctx.isolated() as subctx:
+            rarg, rtable = dispatch.resolve_relation(stmt.rarg, ctx=subctx)
+
+        # valudate equal columns from both sided
```
msullivan, line 58: s/valudate/validate/

msullivan: and s/sided/sides/

msullivan:
I like the table returning setup

edb/pgsql/resolver/relation.py
msullivan: There's a `getptr` function that is probably what you want, instead of looping through

edb/pgsql/resolver/expr.py
```diff
@@ -0,0 +1,350 @@
...
+
+        if isinstance(col_name, pgast.Star):
+            return [(t, c) for t in ctx.scope.tables for c in t.columns]
+        else:
+            for table in ctx.scope.tables:
+                matched_columns.extend(_lookup_in_table(col_name, table))
+
+    elif len(name) == 2:
+        # look for the column in the specific table
+        tab_name = name[0]
```
msullivan, line 114: `tab_name, col_name = name`

edb/pgsql/resolver/range_var.py
```diff
@@ -0,0 +1,326 @@
...
+        name=cte.name,
+        aliascolnames=None,
+        query=query,
+        recursive=cte.recursive,
+        materialized=cte.materialized,
+    )
+    return node, result
+
+
+RANGE_FUNCTIONS_COLS = {
```
msullivan, line 218: Might be worth putting this in its own file?

msullivan (approved):
This is great work!

PR #4654, opened 2022-11-10
SQL introspection

aljazerzen:
Proof of concept for introspection in information_schema tables:

- a new schema `edgedbsql`
- 2 new views: `edgedbsql.tables` and `edgedbsql.columns`
- resolver redirects queries to `information_schema.X` to `edgedbsql.X`

Many columns in the two views are just `NULL`, I hope they are not needed. Looking at the code of analytical software that we want to work with our imitated database, they are not even needed. We do need some views from `pg_catalog`, namely `pg_class` and `pg_namespace`.

For example Rehash:
```sql
-- PostgreSQL.get_tables
SELECT s.nspname as table_schema,
       c.relname as table_name,
       a.attname as column_name,
       null as data_type
FROM pg_class c
JOIN pg_namespace s
ON c.relnamespace = s.oid
AND s.nspname NOT IN ('pg_catalog', 'information_schema')
JOIN pg_attribute a
ON a.attrelid = c.oid
AND a.attnum > 0
AND NOT a.attisdropped
WHERE c.relkind IN ('m', 'f', 'p')
UNION
SELECT table_schema,
       table_name,
       column_name,
       data_type
FROM information_schema.columns
WHERE table_schema NOT IN ('pg_catalog', 'information_schema')
```

@elprans Do you think that we need all of 61 tables in information_schema or are 2 enough? 

There is also middle ground where any references to unimplemented information_schema tables are compiled to `SELECT NULL, NULL, NULL, ... LIMIT 0`, so we don't bloat the metaschema.

(sorry for formatting the dbops commands)


edb/pgsql/metaschema.py
```diff
@@ -4131,121 +4131,128 @@ async def bootstrap(
...
-        dbops.CreateRange(Float32Range()),
-        dbops.CreateRange(Float64Range()),
-        dbops.CreateRange(DatetimeRange()),
-        dbops.CreateRange(LocalDatetimeRange()),
-        dbops.CreateFunction(RangeToJsonFunction()),
-        dbops.CreateFunction(RangeValidateFunction()),
-        dbops.CreateFunction(RangeUnpackLowerValidateFunction()),
-        dbops.CreateFunction(RangeUnpackUpperValidateFunction()),
-    ])
+    commands.add_commands(
```
msullivan, line 4134: I don't think the expanded indentation here makes things better and it churns up the diff

tests/test_sql_query.py
```diff
@@ -373,7 +373,7 @@ async def test_sql_query_28(self):
         self.assert_shape(res, 2, 2, ['name', 'title'])
 
     async def test_sql_query_29(self):
-        # link tables
+        # link tables<g
```
msullivan, line 376: spurious

edb/pgsql/resolver/relation.py
```diff
@@ -265,3 +280,70 @@ def _construct_column(p: s_pointers.Pointer, ctx: Context) -> context.Column:
         col.reference_as = dbname
 
     return col
+
+
+def _get_information_schema(table: str) -> List[str]:
+    # TODO: this should probably go somewhere else
```
msullivan, line 286: Indeed

msullivan (commented):
This looks like a basically solid approach, though we might want it out in a new file or something, since it could grow to be pretty big if we support more.

I don't have a super clear idea exactly what all needs to be supported; @elprans?

edb/pgsql/metaschema.py
aljazerzen: No, it doesn't make it better, but this is what black does.

I'm using "darker", that formats only code chunks that I've changed and that's why this was formatted.

I had reverted it a few times, but I'd be very happy if we could just merge the change, so I don't have to keep reverting it all the time I work on metaschema.

msullivan: Sigh. Maybe do something like
```
new = [
   ...
]
commands.add_commands(new)
```

aljazerzen:
@msullivan this is also ready for re-review

edb/pgsql/metaschema.py
```diff
@@ -5005,6 +5009,149 @@ def _generate_schema_alias_view(
...
+    sql_bool = 'information_schema.yes_or_no'
+    sql_card = 'information_schema.cardinal_number'
+    tables_and_columns = [
+        dbops.View(
+            name=('edgedbsql', 'tables'),
+            query=(
+                f'''
+        SELECT
+            'postgres'::{sql_ident} AS table_catalog,
+            'public'::{sql_ident} AS table_schema,
```
msullivan, line 5024: Is this right? We use the schemas, right?

edb/pgsql/metaschema.py
```diff
@@ -5005,6 +5009,149 @@ def _generate_schema_alias_view(
...
+            NULL::{sql_ident} AS self_referencing_column_name,
+            NULL::{sql_str} AS reference_generation,
+            NULL::{sql_ident} AS user_defined_type_catalog,
+            NULL::{sql_ident} AS user_defined_type_schema,
+            NULL::{sql_ident} AS user_defined_type_name,
+            'NO'::{sql_bool} AS is_insertable_into,
+            'NO'::{sql_bool} AS is_typed,
+            NULL::{sql_str} AS commit_action
+        FROM edgedb."_SchemaObjectType"
+        WHERE name ilike 'default::%'
```
msullivan, line 5036: Oh, we need to support non-default modules. We *could* filter out `std`/`schema`/etc but my inclination would be to leave them in?

edb/pgsql/metaschema.py
```diff
@@ -5005,6 +5009,149 @@ def _generate_schema_alias_view(
...
+            'BASE TABLE'::{sql_str} AS table_type,
+            NULL::{sql_ident} AS self_referencing_column_name,
+            NULL::{sql_str} AS reference_generation,
+            NULL::{sql_ident} AS user_defined_type_catalog,
+            NULL::{sql_ident} AS user_defined_type_schema,
+            NULL::{sql_ident} AS user_defined_type_name,
+            'NO'::{sql_bool} AS is_insertable_into,
+            'NO'::{sql_bool} AS is_typed,
+            NULL::{sql_str} AS commit_action
+        FROM edgedb."_SchemaObjectType"
```
msullivan, line 5035: `tables` should include link tables also

msullivan: and `columns` will need to be populated for link tables too

edb/pgsql/resolver/information_schema.py
```diff
@@ -0,0 +1,812 @@
...
+        ("check_option", "character_data"),
+        ("is_updatable", "yes_or_no"),
+        ("is_insertable_into", "yes_or_no"),
+        ("is_trigger_updatable", "yes_or_no"),
+        ("is_trigger_deletable", "yes_or_no"),
+        ("is_trigger_insertable_into", "yes_or_no"),
+    ],
+}
+
+# retrived with:
```
msullivan, line 796: This doesn't need to go into this PR, but since this is all generated data, I think it might be worth having it be explicitly generated by a tool (along with `range_functions.py`)

msullivan (commented):
Approach looks solid. We need to expose link tables also, though.

edb/pgsql/metaschema.py
aljazerzen: Yes indeed - I forgot we had modules. Can we map one module directly into schema name?

There will be a conflict if someone has a module named `information_schema`, but otherwise it should be ok.

msullivan: Yeah should be fine

aljazerzen:
Only multi property tables left to do.

edb/pgsql/metaschema.py
```diff
@@ -5011,6 +5015,260 @@ def _generate_schema_alias_view(
...
+            SELECT * FROM obj_ty
+        ) UNION ALL (
+            WITH qualified_links AS (
+                -- multi links and links with at least one property
+                -- (besides source and target)
+                SELECT link.id
+                FROM edgedb."_SchemaLink" link
+                JOIN edgedb."_SchemaProperty" AS prop ON link.id = prop.source
+                WHERE prop.computable IS NOT TRUE AND prop.internal IS NOT TRUE
+                GROUP BY link.id, link.cardinality
```
msullivan, line 5049: Logic that seems so simple when stuck in a python function can get a lot gnarlier inlined into a SQL query... :)

msullivan (approved):
Fantastic work here!

PR #4464, opened 2022-09-30
Reword and add uniformity to Standard Library documentation

i0bs:
This Pull Request seeks to improve the Standard Library documentation by rewording content written under types, operators and functions to be more uniform with the rest. I am marking this as a draft due to the abundant size of these docs and the amount of typos/grammatical inconsistencies needed to account for.

PR #4655, opened 2022-11-10
Clarify the `duration_get` function

raddevon:
The behavior of `duration_get` is inconsistent. For duration elements month or larger, the units are aggregated to return the final value. For duration elements smaller than month, only the literal value expressed in the duration for that unit is returned. Since this behavior could be unexpected, we need additional documentation around it.

### DIFF
--- a/docs/stdlib/datetime.rst
+++ b/docs/stdlib/datetime.rst
@@ -967,7 +967,7 @@ EdgeDB stores and outputs timezone-aware values in UTC.
     .. code-block:: edgeql-repl
 
         db> select duration_get(
-        ...   <cal::relative_duration>''20 hours 3600 seconds', 'hour');
+        ...   <cal::relative_duration>'20 hours 3600 seconds', 'hour');
         {21}
 
     Seconds and smaller units always return remaining time in that unit after
@@ -976,10 +976,10 @@ EdgeDB stores and outputs timezone-aware values in UTC.
     .. code-block:: edgeql-repl
 
         db> select duration_get(
-        ...   <cal::relative_duration>''20 hours 3600 seconds', 'seconds');
+        ...   <cal::relative_duration>'20 hours 3600 seconds', 'seconds');
         {0}
         db> select duration_get(
-        ...   <cal::relative_duration>''20 hours 3630 seconds', 'seconds');
+        ...   <cal::relative_duration>'20 hours 3630 seconds', 'seconds');
         {30}
 
     Normalization and truncation may help you deal with this. If your use case
--- a/docs/stdlib/datetime.rst
+++ b/docs/stdlib/datetime.rst
@@ -982,6 +982,15 @@ EdgeDB stores and outputs timezone-aware values in UTC.
         ...   <cal::relative_duration>''20 hours 3630 seconds', 'seconds');
         {30}
 
+    Normalization and truncation may help you deal with this. If your use case
+    allows for making assumptions about the duration of a month or a day, you
+    can make those conversions for yourself using the
+    :eql:func:`cal::duration_normalize_hours` or
+    :eql:func:`cal::duration_normalize_days` functions. If you got back a
+    duration as a result of a datetime calculation and don't need the level of
+    granularity you have, you can truncate the value with
+    :eql:func:`duration_truncate`.
+
 ----------
 
 
--- a/docs/stdlib/datetime.rst
+++ b/docs/stdlib/datetime.rst
@@ -865,14 +865,14 @@ EdgeDB stores and outputs timezone-aware values in UTC.
       and :eql:type:`cal::date_duration` for durations longer than a day,
       because a month is assumed to be 30 days exactly.
 
-    The :eql:type:`duration` scalar only has the smallest units available
-    for extraction.
+    The :eql:type:`duration` scalar has only ``'hour'`` and smaller units
+    available for extraction.
 
     The :eql:type:`cal::relative_duration` scalar has all of the units
     available for extraction.
 
-    The :eql:type:`cal::date_duration` scalar only has the largest and middle
-    size units available for extraction.
+    The :eql:type:`cal::date_duration` scalar only has ``'date'`` and larger
+    units available for extraction.
 
     .. code-block:: edgeql-repl
 
--- a/docs/stdlib/datetime.rst
+++ b/docs/stdlib/datetime.rst
@@ -860,10 +860,10 @@ EdgeDB stores and outputs timezone-aware values in UTC.
 
     Additionally, it's possible to convert a given duration into seconds:
 
-    - ``'totalseconds'`` - the number of seconds represented by the
-      duration. It will be approximate for :eql:type:`cal::relative_duration`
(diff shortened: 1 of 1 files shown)


docs/stdlib/datetime.rst
```diff
@@ -905,6 +903,48 @@ EdgeDB stores and outputs timezone-aware values in UTC.
         ...   <duration>'30 hours', 'totalseconds');
         {108000}
 
+    **Gotchas**
```
1st1, line 906: Can you try using `.. rubric:: Gotchas`; would it look good visually? Semantically it would still be a header, not a bold text, which is better in this case.

raddevon:
This isn't quite ready yet. Just noticed a mistake on my part. I'll get it fixed and push again.

docs/stdlib/datetime.rst
raddevon: I'll give that a shot. Thank you!

raddevon: That seemed to work well, @1st1. Thanks for the suggestion!

raddevon:
OK, we're good to go now. Review when ready!

docs/stdlib/datetime.rst
```diff
@@ -905,6 +903,77 @@ EdgeDB stores and outputs timezone-aware values in UTC.
...
+        ...   <cal::relative_duration>'1 year 400 days', 'year');
+        {1}
+
+    When you request a unit in the smallest size class, it will be pooled with
+    other durations in the same size class.
+
+    .. code-block:: edgeql-repl
+
+        db> select duration_get(
+        ...   <cal::relative_duration>''20 hours 3600 seconds', 'hour');
```
i0bs, line 963: This apparently has a test failing over ``seconds`` being used. This looks completely fine though, is there something wrong with the test invalidating?

docs/stdlib/datetime.rst
```diff
@@ -827,54 +827,52 @@ EdgeDB stores and outputs timezone-aware values in UTC.
                   std::duration_get(dt: cal::date_duration, \
                                     el: str) -> float64
 
-    Extract a specific element of input duration by name.
+    Return a specific element of input duration given a unit name.
```
i0bs, line 830: This might lead to a merge conflict if accepted before #4464, is L830s current writing in that OR okay, or should I bring this over to that one?

docs/stdlib/datetime.rst
i0bs, line 963:
```suggestion
        ...   <cal::relative_duration>'20 hours 3600 seconds', 'hour');
```

docs/stdlib/datetime.rst
```diff
@@ -905,6 +903,77 @@ EdgeDB stores and outputs timezone-aware values in UTC.
...
+        ...   <cal::relative_duration>''20 hours 3600 seconds', 'hour');
+        {21}
+
+    Seconds and smaller units always return remaining time in that unit after
+    accounting for the next larger unit.
+
+    .. code-block:: edgeql-repl
+
+        db> select duration_get(
+        ...   <cal::relative_duration>''20 hours 3600 seconds', 'seconds');
```
i0bs, line 972:
```suggestion
        ...   <cal::relative_duration>'20 hours 3600 seconds', 'seconds');
```

docs/stdlib/datetime.rst
```diff
@@ -905,6 +903,77 @@ EdgeDB stores and outputs timezone-aware values in UTC.
...
+        {0}
+        db> select duration_get(
+        ...   <cal::relative_duration>''20 hours 3630 seconds', 'seconds');
```
i0bs, line 975:
```suggestion
        ...   <cal::relative_duration>'20 hours 3630 seconds', 'seconds');
```

i0bs (changes requested):

docs/stdlib/datetime.rst
raddevon: Thanks for calling this out. There are elements I like in each of them. What about a hybrid of the two?

>Returns the element of a duration given a unit name.

I'll go ahead and update on my side.

docs/stdlib/datetime.rst
raddevon: I'm guessing that was a result of my errant quotes that you found? (Thanks for catching that, btw.) It looks like all the tests are passing now, but let me know if you still see an issue.

raddevon:
@vpetrovykh I believe these latest changes address your feedback from our chat on Friday. I've given more context around the reason for the size classes and introduced the normalization and truncation functions as ways to deal with this.

i0bs (approved):

docs/stdlib/datetime.rst
```diff
@@ -827,54 +827,52 @@ EdgeDB stores and outputs timezone-aware values in UTC.
...
+      Only for units ``'month'`` or larger or for units ``'hour'`` or smaller
+      will you receive a total across multiple units expressed in the original
+      duration. See *Gotchas* below for details.
 
     Additionally, it's possible to convert a given duration into seconds:
 
     - ``'totalseconds'`` - the number of seconds represented by the
       duration. It will be approximate for :eql:type:`cal::relative_duration`
-      and :eql:type:`cal::date_duration` with usints larger than a day,
+      and :eql:type:`cal::date_duration` for durations longer than a day,
```
vpetrovykh, line 865: It's approximate if the duration included units of months or larger, not just "longer than a day". As long as the units are just days, the `totalseconds` conversion will simply assume 24hr days and do an exact conversion. Approximations happen for larger units: months+.

docs/stdlib/datetime.rst
```diff
@@ -905,6 +903,93 @@ EdgeDB stores and outputs timezone-aware values in UTC.
...
+    will return only the number of days expressed as ``N days`` in your
+    duration. It will not add another day to the returned count for every 24
+    hours (defined as ``24 hours``) in the duration, nor will it consider the
+    months' constituent day counts in the returned value.
+
+    Specifying ``'decade'`` for ``el`` will total up all decades represented in
+    units ``'month'`` and larger, but it will not add a decade's worth of days
+    to the returned value as an additional decade. On the other end of the unit
+    spectrum, passing ``'hour'`` for ``el`` will not add a day's worth of hours
+    for each day in your duration.
```
vpetrovykh, line 933: The part about hours not rolling over into days seems to be a repetition of the point made in the previous paragraph.

vpetrovykh (commented):
Mostly looks good. I have a couple of comments. The one about "longer than a day" requires more precise phrasing. The second one is more of a judgement call on whether repetition is good in explanations.

docs/stdlib/datetime.rst
raddevon: You're right on this. I removed that sentence and joined the one about decades with the previous paragraph.

raddevon:
@vpetrovykh Thank you for looking this over. I think I've resolved both of the issues you found. Whenever you get a chance, take a look and approve if those changes do in fact resolve the issues adequately.

PR #4667, opened 2022-11-14
Fix putting a statement as the body of an access policy

msullivan:
Currently it isn't (by oversight) allowed. The particularly bad part,
though, is that while you can fix the parse by wrapping the statement
in parens, the DDL codegen will lose the parens again.

The easiest way to work around this is probably wrapping it in braces,
so it is a set literal.


elprans:
> the DDL codegen will lose the parens again

Let's fix that?

msullivan:
> > the DDL codegen will lose the parens again
> 
> Let's fix that?

The extra parens shouldn't be required. They aren't required in any other place a parenthesized expression appears in the schema.

elprans (approved):

PR #4671, opened 2022-11-15
Fix `max_ex_value` constraint.

vpetrovykh:
The constraint was not exclusive. This is now fixed.


vpetrovykh:
While answering @raddevon's question about `max_ex_value` constraint I realized that the constraint in `std` is actually not exclusive (`min_ex_value` was fine though).

elprans (approved):

vpetrovykh:
Ooops, forgot to bump catalog version after changing `std` library.

PR #4675, opened 2022-11-15
Add wrappers for many more bytes functions

msullivan:
This exposes:
 * convert_from
 * convert_to
 * encode
 * decode
 * sha{224,256,384,512}

I kept all of their names, but prefixed the non-SHA ones with
`bytes_`.  I'm not confident in either of those decisions and so would
be happy to have discussion on that.

We'll need to add some tests and documentation, but I'm putting this
PR up in advance of that because I want to use a hash function to
generate deterministic "random" numbers for testing purposes.

### DIFF
--- a/edb/lib/std/30-bytesfuncs.edgeql
+++ b/edb/lib/std/30-bytesfuncs.edgeql
@@ -32,6 +32,86 @@ std::bytes_get_bit(bytes: std::bytes, num: int64) -> std::int64
 };
 
 
+CREATE FUNCTION
+std::bytes_convert_from(bytes: std::bytes, src_encoding: std::str) -> std::str
+{
+    CREATE ANNOTATION std::description :=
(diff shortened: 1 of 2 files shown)


edb/lib/std/30-bytesfuncs.edgeql
```diff
@@ -32,6 +32,86 @@ std::bytes_get_bit(bytes: std::bytes, num: int64) -> std::int64
...
+{
+    CREATE ANNOTATION std::description :=
+        'Decode a textual representation of a binary string';
+    SET volatility := 'Immutable';
+    USING SQL FUNCTION 'decode';
+};
+
+
+CREATE FUNCTION
+std::sha224(bytes: std::bytes) -> std::bytes
```
elprans, line 76: I would consider adding `sha` into a `hash` module or something similar.

aljazerzen (approved):

PR #4676, opened 2022-11-16
Add dotnet

ShalokShalom:
This adds a link to dotnet

### DIFF
--- a/docs/intro/quickstart.rst
+++ b/docs/intro/quickstart.rst
@@ -402,8 +402,9 @@ likely be using one of EdgeDB's *client libraries* to execute queries. EdgeDB
 provides official libraries for
 `JavaScript/TypeScript <https://github.com/edgedb/edgedb-js>`__,
 `Go <https://github.com/edgedb/edgedb-go>`__,
-`Python <https://github.com/edgedb/edgedb-python>`__, and
-`Rust <https://github.com/edgedb/edgedb-rust>`_. Check out the :ref:`Clients
+`Python <https://github.com/edgedb/edgedb-python>`__, 
+`Rust <https://github.com/edgedb/edgedb-rust>`__, and
+`C# and F# <https://github.com/edgedb/edgedb-net>`_. Check out the :ref:`Clients
 <ref_intro_clients>` guide to get started with the language of your choice.
 
 .. _ref_quickstart_onwards:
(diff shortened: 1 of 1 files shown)


i0bs (approved):

docs/intro/quickstart.rst
```diff
@@ -395,8 +395,9 @@ likely be using one of EdgeDB's *client libraries* to execute queries. EdgeDB
 provides official libraries for
 `JavaScript/TypeScript <https://github.com/edgedb/edgedb-js>`__,
 `Go <https://github.com/edgedb/edgedb-go>`__,
-`Python <https://github.com/edgedb/edgedb-python>`__, and
-`Rust <https://github.com/edgedb/edgedb-rust>`_. Check out the :ref:`Clients
+`Python <https://github.com/edgedb/edgedb-python>`__, 
+`Rust <https://github.com/edgedb/edgedb-rust>`__, and
+`C# and F# <https://github.com/edgedb/edgedb-net>`_. Check out the :ref:`Clients
```
quinchs, line 400:
```suggestion
`.NET <https://github.com/edgedb/edgedb-net>`_. Check out the :ref:`Clients
```

quinchs (approved):

docs/intro/quickstart.rst
ShalokShalom: You provide Clojure support? Visual Basic and a lot of others?

ShalokShalom: As someone who is an avid F# fan, I can only say that I hate it, when a project advertises dotnet support, as if that's a language or as when they support the whole platform 😅

They let you look through the documentation, just to reveal your language is actually not supported.

i0bs: The client offers at least VB.NET, ASP.NET, C# and F# support. In most cases where you'll work with either VB or ASP frameworks this would be valid

ShalokShalom: Yeah, true. Why wouldn't it work with Clojure?

I was more speaking about the support via documentation.

If a feature is not documented, it could as well not exist.

quinchs: The binding work with any language that is able to import .NET class libraries, which include C#, VB.NET, F#, etc. We may not have documentation for other .NET languages but that does not mean that it's not supported.

msullivan:
Hm... I'm not sure why the tests are messed up.

PR #4689, opened 2022-11-18
TEST: Test most of the changes I want for 2.7

msullivan:
(This includes provisional versions of https://github.com/edgedb/edgedb/pull/4670 and https://github.com/edgedb/edgedb/pull/4688)


msullivan:
Very confused why the CLA bot is hassling me here but it isn't blocking here anyway so

PR #4691, opened 2022-11-18
Add the syntax for abstract indexes.

vpetrovykh:
Add the syntax for complex abstract index definitions as well as update the concrete index syntax to account for more index variety.

### DIFF
--- a/edb/edgeql/ast.py
+++ b/edb/edgeql/ast.py
@@ -1175,24 +1175,59 @@ class DropConcreteConstraint(ConcreteConstraintOp, DropObject):
     pass
 
 
+class IndexType(DDL):
+    name: ObjectRef
+    args: typing.List[Expr] = ast.field(factory=list)
+    kwargs: typing.Dict[str, Expr] = ast.field(factory=dict)
+
+
 class IndexCommand(ObjectDDL):
 
     __abstract_node__ = True
-    expr: Expr
-    except_expr: typing.Optional[Expr] = None
     object_class: qltypes.SchemaObjectClass = (
         qltypes.SchemaObjectClass.INDEX)
 
 
-class CreateIndex(IndexCommand, CreateObject):
+class IndexCode(Clause):
+    language: Language
+    code: str
+
+
+class CreateIndex(
+    CreateExtendingObject,
+    CallableObjectCommand,
+    IndexCommand,
+):
+    kwargs: typing.Dict[str, Expr] = ast.field(factory=dict)
+    index_types: typing.List[IndexType]
+    code: typing.Optional[IndexCode] = None
+
+
+class AlterIndex(AlterObject, IndexCommand):
     pass
 
 
-class AlterIndex(IndexCommand, AlterObject):
+class DropIndex(DropObject, IndexCommand):
     pass
 
 
-class DropIndex(IndexCommand, DropObject):
+class ConcreteIndexCommand(IndexCommand):
+
+    __abstract_node__ = True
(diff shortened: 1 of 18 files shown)


vpetrovykh:
I've also updated the related [RFC 1017](https://github.com/edgedb/rfcs/pull/66).

edb/edgeql/ast.py
```diff
@@ -1170,24 +1170,54 @@ class DropConcreteConstraint(ConcreteConstraintOp, DropObject):
     pass
 
 
+class IndexType(DDL):
+    name: typing.Union[tuple, str]
```
msullivan, line 1174: It's a `tuple[str]`, right?

edb/edgeql/codegen.py
```diff
@@ -1808,19 +1817,82 @@ def _after_index(self, node: qlast.IndexCommand) -> None:
...
+            for i, (name, arg) in enumerate(node.kwargs.items()):
+                if i > 0:
+                    self.write(', ')
+                self.write(f'{edgeql_quote.quote_ident(name)} := ')
+                self.visit(arg)
+            self.write(')')
+
     def visit_CreateIndex(self, node: qlast.CreateIndex) -> None:
+        def after_name() -> None:
+            if node.params:
```
msullivan, line 1841: What are params, here? How do they interact with kwargs?

edb/edgeql/parser/grammar/commondl.py
```diff
@@ -543,3 +561,111 @@ def reduce_ExtensionVersion(self, *kids):
...
+
+    def reduce_IndexExtArgList(self, *kids):
+        self.val = kids[0].val
+
+    def reduce_empty(self):
+        self.val = []
+
+
+class ProcessIndexArgsMixin:
+    def _process_arguments(self, arguments):
```
msullivan, line 622: This could just be a function, I think, instead of a mixin?

msullivan: Oh but there's an existing one for functions?

edb/edgeql/codegen.py
msullivan: Oh, I understand now. Only one can be present, though, right? Maybe have an assert to that effect?

stealth-sparrow:
Is it possible to add support for `array_ops` indexes?

```
CREATE ABSTRACT INDEX std::gin_array ON array<anytype> {
    USING SQL $$gin ((__col__))$$;
};
```

stealth-toucan:
Is it possible to add support for `array_ops` indexes?

```
CREATE ABSTRACT INDEX std::gin_array ON array<anytype> {
    USING SQL $$gin ((__col__))$$;
};
```

edb/edgeql/parser/grammar/commondl.py
vpetrovykh: At the moment I'm just following the same pattern as for functions.

edb/edgeql/codegen.py
vpetrovykh: No args or params in this iteration of the PR anymore.

vpetrovykh:
This implementation hard-codes the valid indexable types as well as the specific SQL that needs to be emitted to define the built-in Postgres indexes. In the future both of these may be defined more formally via DDL, though.

msullivan (approved):
This looks good to me. Can you add some DDL tests?

msullivan (approved):

PR #4694, opened 2022-11-18
Bootstrap command file

fmoor:
This is a follow up to https://github.com/edgedb/edgedb/pull/4693


fmoor:
Maybe the docker image should also be updated?

elprans (approved):

PR #4706, opened 2022-11-23
refactor sql resolver with dataclasses

aljazerzen:


edb/pgsql/resolver/context.py
```diff
@@ -20,12 +20,14 @@
...
 from copy import deepcopy
 from typing import *
+from dataclasses import dataclass, field
 import enum
 
 from edb.common import compiler
 from edb.schema import schema as s_schema
 
 
+@dataclass
```
elprans, line 30: Consider

```suggestion
@dataclass(kw_only=True)
```

PR #4713, opened 2022-11-24
Refactor compiler

aljazerzen:
server/compiler/compiler.py had 3000+ lines of code, with `compile_ql_migration` being 700+ lines long.

The main thing I did here was extracting `compile_ql_migration` and `compile_and_apply_ddl_stmt` into `ddl.py`. To do that, I had to convert object-oriented Compiler into functional code that uses CompilerContext for all state needed for compilation.

I suggest reviewing commits separately, otherwise the diffs are just too big.


msullivan (approved):
This seems good! Breaking up the migration stuff in particular is very nice.

Reviewing this kind of refactoring diff is hard, so I appreciate the detailed comment in the first commit. (Really I reviewed the comment, then read the commits to make sure that it looks reasonable and approximately does what was claimed.) Make sure that makes it into the final message.

PR #4714, opened 2022-11-24
Reset schema

### LINKED ISSUE #4351: Add support for `RESET SCHEMA TO initial`
Per [RFC 1000](https://github.com/edgedb/rfcs/blob/master/text/1000-migrations.rst).  `RESET SCHEMA TO <arbitrary-migration>` requires support for migration reversals, but `RESET SCHEMA TO initial` does not and is a requirement for `edgedb database wipe` or a similar action of factory-resetting a database.

aljazerzen:
Closes #4351

### DIFF
--- a/edb/server/compiler/ddl.py
+++ b/edb/server/compiler/ddl.py
@@ -1098,8 +1098,6 @@ def _reset_schema(
     current_tx = ctx.state.current_tx()
     schema = current_tx.get_schema(ctx.compiler_state.std_schema)
 
-    current_tx.start_migration()
-
     empty_schema = s_schema.ChainedSchema(
         ctx.compiler_state.std_schema,
         s_schema.FlatSchema(),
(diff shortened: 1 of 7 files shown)


edb/server/compiler/ddl.py
```diff
@@ -1076,3 +1079,68 @@ def _abort_migration_rewrite(
...
+        raise errors.QueryError(
+            f'Unknown schema version "{ql.target.name}". '
+            'Currently, only revision supported is "initial"',
+            context=ql.target.context,
+        )
+
+    current_tx = ctx.state.current_tx()
+    schema = current_tx.get_schema(ctx.compiler_state.std_schema)
+
+    current_tx.start_migration()
```
msullivan, line 1101: Hm, does this get cleaned up somewhere?

msullivan (approved):
Seems good, but could you double check that the `start_migration` call doesn't leave any state hanging?

PR #4727, opened 2022-11-29
Add an expand_inhviews debug flag that forces views to get expanded

msullivan:
This helps lay the groundwork for EXPLAIN: when referring to views (or
non-materialized CTEs), postgres's EXPLAIN output does not refer to
the alias names in our compiled query. We need it to use those alias
names, though, in order to correspond query plan nodes with parts of
our query.

### DIFF
--- a/edb/pgsql/compiler/clauses.py
+++ b/edb/pgsql/compiler/clauses.py
@@ -52,7 +52,14 @@ def get_volatility_ref(
     if not ref:
         rvar = relctx.maybe_get_path_rvar(
             stmt, path_id, aspect='value', ctx=ctx)
-        if rvar and isinstance(rvar.query, pgast.ReturningQuery):
+        if (
+            rvar
+            and isinstance(rvar.query, pgast.ReturningQuery)
+            # Expanded inhviews might be unions, which can't naively have
+            # a row_number stuck on; they should be safe to just grab
+            # the path_id value from, though
+            and rvar.tag != 'expanded-inhview'
+        ):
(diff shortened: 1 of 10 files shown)


edb/pgsql/compiler/clauses.py
```diff
@@ -52,7 +52,14 @@ def get_volatility_ref(
     if not ref:
         rvar = relctx.maybe_get_path_rvar(
             stmt, path_id, aspect='value', ctx=ctx)
-        if rvar and isinstance(rvar.query, pgast.ReturningQuery):
+        if (
+            rvar
+            and isinstance(rvar.query, pgast.ReturningQuery)
+            # XXX: Is this legit? Otherwise ran into trouble with
```
elprans, line 58: I'm not 100% sure here, feels dangerous to me.  A cleaner solution would be to inject a tuple of `(row_number(), relation_number)` into every member of the union op, where `relation_number` is just an index of the relation or something.

msullivan: I wound up handling this by specifically exempting expanded inhviews, which we know will be safe to just use the value for, and I didn't want to have the volatility ref be substantially different in expanded mode.

I /think/ that other unions shouldn't be able to sneak in here (they'll typically be wrapped in another select), but if they do we'll fail instead of doing the wrong thing. If we encounter that we can do your suggestion, but I don't want to implement it without a test case, and I don't want to generate that code for these expanded inhviews

elprans (approved):

PR #4728, opened 2022-11-29
Represent the end column when dumping a context

msullivan:


edb/common/context.py
```diff
@@ -129,9 +131,11 @@ def as_markup(cls, self, *, ctx):
                 lines.append(ctx_line)
                 line_numbers.append(start.line + i)
 
+        endcol = end.column if start.line == end.line else None
```
aljazerzen, line 134: Are you sure that end will never be null?

aljazerzen (approved):

edb/common/context.py
aljazerzen: On second though, no it cannot be None.
It's asserted by constructor and it also gets enforced by edgedb_rust: https://github.com/edgedb/edgedb/pull/4728/files#diff-1b935a02e66d82bbf000081591f8a0143012392b0ede79f0753071f1bb45e12fR52-R63

PR #4729, opened 2022-11-29
Postgres protocol extension

fantix:
# Outline

- [x] Extract base class for frontend connections
- [x] Add `compile_sql()` API
- [x] Support simple query
- [x] Handle error and return ErrorResponse
- [x] Add SQL compile cache
- [x] Support extended query
- [x] Support SCRAM auth
- [x] Support SSL
- [x] Handle `CancelRequest`
- [x] Drop `describe alter` hack
- [x] Add tests for simple query
- [x] Add tests for extended query
- [x] Support suspended portals that are immediately closed
- [ ] Support transactions
- [ ] Support SET
- [ ] Support md5 auth

# Difference to native PostgreSQL protocol

* Only read-only queries allowed
* Only SCRAM and Trust authentication methods are currently supported
* Transactions are not supported (yet?)
* Simple query may return slightly different errors as it's implemented in extended query for reusing backend prepared statements
* Extended queries without an explicit transaction:
  * Try to consume all portals/cursors in a single pipelined packet and end it with a `Sync` message to gain best compatibility (which is what most libraries do).
  * Sending packets separately (like `Parse`-`Bind` in the first `send()`, `Execute` in the second `send()`), network partition or sending non-extended-query messages will be treated as a separate pipeline with slightly lower performance.
  * In the end of the pipeline, the implicit transaction is always closed (with pending messages flushed) even if the client didn't or forgot to send a `Sync`, though the client won't receive a `ReadyForQuery` message. In the case of an error, the server still expects to see a `Sync` message to recover from the error. In short, it still looks like a single implicit transaction to the frontend client, but it may actually be done in several separate implicit transaction in the backend that matches the pipeline partitions. So clients should not rely on the implicit transaction if network partition is unavoidable; use explicit transactions (when it's ready) when strict data consistency is needed.
  * The portal will be destroyed in the end of a pipeline (even without a `Sync`) if it is already executed.
  * `Execute` with a limit is not supported.


aljazerzen:
Nice, got it working with psql! And my parser+resolver also work.

There was a problem when tried to connect via JDBC connection. It threw an error here: https://github.com/edgedb/edgedb/blob/5bdc35fa01d61b7bd69ca92f33aae378d367bed2/edb/server/protocol/frontend.pyx#L248

You may not have noticed, but I did a refactor of server/compiler/compiler.py which you'll have to merge. Just a heads up on the incoming conflicts.

elprans:
> I think we'll have to pin the backend pgcon to support client-cached prepared statements

We don't have to if we cache the prepared statement text in the server.  I.e. we can lie about the preparedness of the statement.

fantix:
> > I think we'll have to pin the backend pgcon to support client-cached prepared statements
> 
> We don't have to if we cache the prepared statement text in the server. I.e. we can lie about the preparedness of the statement.

Sure. What about "portals"? If the idea is to avoid pinning pgcons for "pg-ext" frontend connections as much as possible, I think we can do something like this for named portals/server-side cursors:

1. We assume the extended queries are short-lived, because we will temporarily hold a backend pgcon during executing them.
2. We expect portals - named or unnamed - to be consumed in consecutive messages as described in the extended query workflow. Any other unexpected messages in the middle will be considered as an implicit close action on the portals.
3. One exception would be fresh-new portals that haven't started to execute. We could similarly lie about the plannedness of the portals (bind again when requested to execute).
4. (3) has no performance overhead when this assumption is true: pipelined messages are rarely arriving separately on a TCP connection. Like if the pipelined `Parse + Bind + Execute` are received separately, the server would do "Parse" 3 times and "Bind" twice.

Or else if this is too complicated, we could always pin a backend pgcon for executed portals (or simply return an "unimplemented" error), while fresh-new portals could do the same lying hack.

elprans:
It's not clear how much of a usage named portals and cursors outside of transactions actually have in the wild.  So my take would be to raise `feature_not_supported` on an attempt to specify the row count in `Execute`.  This is the only scenario which we can't emulate (i.e suspended portals).

So, for clarity:

- if in a transaction: connection already pinned, so we can support portals and cursors without a problem.
- if not in a transaction: only fully-executed portals are supported, i.e. the one where you don't specify the row count in `Execute`.  `Parse` and `Bind` can be partially emulated, i.e. we maintain a prepared statement/portal mapping in our connection object and simply reissue `Parse`/`Bind` on subsequent `Execute`.

edb/server/protocol/pg_ext.pyx
```diff
@@ -0,0 +1,763 @@
...
+                    )
+
+                sql_source = await self.compile(query_str)
+                if len(sql_source) > 1:
+                    raise pgerror.new(
+                        pgerror.ERROR_SYNTAX_ERROR,
+                        "cannot insert multiple commands into a prepared "
+                        "statement",
+                    )
+                sql_text = sql_source[0].encode("utf-8")
```
aljazerzen, line 525: Result of parsing + resolving may also be SQL 0 statements - `sql_source[0]` will throw an error.

elprans (approved):

Neustradamus:
@fantix: Good job!

Linked to:

Issue #4744, opened 2022-11-30
Edgedb Buildpack can no longer create Postgresql Extensions in Heroku


Kiefna:

kage1414:
Has there been any progress on this issue? I'm currently waiting to deploy a new version of an app that uses EdgeDB

fantix:
Fixed in #5413

PR #4748, opened 2022-12-01
Note in Heroku guide that EdgeDB deployment is currently broken

raddevon:
Temporarily hiding this while the buildpack is broken due to a recent Heroku change. See https://github.com/edgedb/edgedb/issues/4744

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -82,7 +82,7 @@ language = None
 
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
-exclude_patterns = []
+exclude_patterns = ['guides/deployment/heroku.rst']
 
 # The reST default role (used for this markup: `text`) to use for all
 # documents.
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -82,7 +82,7 @@ language = None
 
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
-exclude_patterns = ['guides/deployment/heroku.rst']
+exclude_patterns = []
 
 # The reST default role (used for this markup: `text`) to use for all
 # documents.
--- a/docs/guides/deployment/heroku.rst
+++ b/docs/guides/deployment/heroku.rst
@@ -6,6 +6,13 @@ Heroku
 
 :edb-alt-title: Deploying EdgeDB to Heroku
 
+
+.. warning::
+
+    Deployment to Heroku is currently not working due to a change in Heroku's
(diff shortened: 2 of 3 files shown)


i0bs (approved):
Admonition may also work, looks good

i0bs:
> I was hoping there might be a way to just toggle this TOC entry off as a comment, but best I can tell, that's not possible. Comments require multiple lines and break any indented blocks they're in. 🫤

Best way I've found is to delete the TOC tree item which isn't ideal, between this and admonition they're the best takes

fmoor (approved):

docs/conf.py
```diff
@@ -82,7 +82,7 @@
 
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
-exclude_patterns = []
+exclude_patterns = ['guides/deployment/heroku.rst']
```
1st1, line 85: The correct way is to mark the rest file as `:orphan:`

CleanShot 2022-12-01 at 10 51 54@2x

1st1 (changes requested):
Use `:orphan:` instead of messing with the config

docs/conf.py
raddevon: Thanks, @1st1. I'll keep this in mind for the future. Wasn't sure the proper way to do this.

I think we might pull back on this, keeping the guide visible and adding a note at the top of the guide, that way we communicate we do intend to continue to support deployment to Heroku. I'll get this revised.

raddevon:
Take a look at this alternative approach when you have a chance, @1st1. This better communicates our intent to continue supporting Heroku as a deployment target. It doesn't require orphaning anything.

fmoor:
Maybe include a link to the github issue for this problem? It's nice to be able to get notifications for things like this.

raddevon:
@fmoor Great idea! I added a link to the issue.

msullivan:
Heroku is pretty broken in several ways, iirc, though

raddevon:
> Heroku is pretty broken in several ways, iirc, though

@msullivan Meaning solving the issue I've linked here will not fully fix Heroku deployment? Should I just revert to  the warning without the link to the Github issue?

msullivan:
Hm, I might be wrong about that. I think it could just be that a bunch of stuff about the tests are broken

raddevon:
I think this is ready to go if you don't mind approving, @1st1.

elprans (approved):

1st1 (approved):

PR #4751, opened 2022-12-01
Warn users not to run CLI as root

raddevon:


docs/cli/index.rst
```diff
@@ -25,6 +25,10 @@ on-screen instructions:
 
     $ curl --proto '=https' --tlsv1.2 -sSf https://sh.edgedb.com | sh
 
+.. warning::
+
+    You should not run the CLI as root.
```
fmoor, line 30: This is generally a good idea, but it might work. If this is addressing what I think it is the major pain point is that people try to run a server as root which fails with an error and really should not be done.

@elprans may have more nuanced guidance for this.

raddevon: Here's the Slack discussion that inspired this for context: https://discord.com/channels/841451783728529451/1043930111528935555/1044655465977622619

elprans: The fundamental issue here is that Postgres will refuse to run as root and so it is impossible to create instances as root.  Other operations are safe, so this should really be handled by the CLI in `create instance`.

fmoor: This also comes up when people try to run `edgedb-server-2` from the apt package as root. They usually don't understand what system-d is and that they don't actually need to start the server themselves.

fmoor: So i guess this bit should go in the bare metal deployment guide?

raddevon: @elprans @fmoor Thanks for your help on this. Check my understanding here: we should move this warning to the bare metal deployment guide, and the CLI itself needs to handle the scenario that the user tries to run `edgedb create instance` as root so there's no need to document that. Does that sound right?

Is the wording of the warning OK?

fmoor: > Does that sound right?

as I understand it, yes

> Is the wording of the warning OK?

The warning should say that `edgedb-server` can not be run as root. This is different than the CLI.

docs/guides/deployment/bare_metal.rst
```diff
@@ -77,6 +77,10 @@ default. You can start the server by enabling the unit.
 This will start the server on port 5656, and the data directory will be
 ``/var/lib/edgedb/1/data``.
 
+.. warning::
+
+    `edgedb-server` cannot be run as root.
+
```
raddevon, line 83: Not sure about the placement of this, but I think I at least have the wording right, @fmoor.

docs/guides/deployment/bare_metal.rst
```diff
@@ -77,6 +77,10 @@ default. You can start the server by enabling the unit.
 This will start the server on port 5656, and the data directory will be
 ``/var/lib/edgedb/1/data``.
 
+.. warning::
+
+    ``edgedb-server`` cannot be run as root.
+
```
raddevon, line 83: @fmoor Had to push again. Brain is permanently stuck on Markdown mode. 😅 

Placement look good here?

fmoor: looks fine to me.

fmoor (approved):

PR #4757, opened 2022-12-02
Fix: error message for function calls on derived types

### LINKED ISSUE #4749: Don't use derived type names in "function does not exist errors"
```
edgedb> with x := 'a', select sum(x);
error: QueryError: function "sum(arg0: __derived__::x)" does not exist
```
This is a bad error, we should say
```
edgedb> with x := 'a', select sum(x);
error: QueryError: function "sum(arg0: std::str)" does not exist
```

Operators get this right so fixing it is probably a simple matter:
```
edgedb> with x := 'a', select x ++ 1;
error: InvalidTypeError: operator '++' cannot be applied to operands of type 'std::str' and 'std::int64'
```

aljazerzen:
Closes #4749

### DIFF
--- a/edb/edgeql/compiler/func.py
+++ b/edb/edgeql/compiler/func.py
@@ -113,11 +113,13 @@ def compile_FunctionCall(
             # Skip any name colliding with kwargs.
             while f'arg{argnum}' in kwargs:
                 argnum += 1
+            ty = schemactx.get_material_type(argtype, ctx=ctx)
             sig.append(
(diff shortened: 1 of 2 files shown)


elprans:
Please add a test

elprans (approved):

msullivan (approved):
Good fix and nice cleanups.

Just to confirm, the operator stuff is separate from the fix, and so if I wanted to omit that in the cherry-pick I could?

PR #4759, opened 2022-12-02
Update mypy

aljazerzen:
Update mypy & refactor many typing annotations. This should probably be merged after 2.8


edb/pgsql/parser/ast_builder.py
```diff
@@ -46,8 +46,8 @@ def _maybe(
     return None
 
 
-def _ident(t: T) -> U:
```
msullivan, line 49: I'd drop the `T` and just make it `object` or `Any`. Not much point in having a type param only used once.

msullivan (approved):
Looks good!

We probably won't cherry-pick this to 2.8 unless it becomes a pain not to

msullivan:
Thanks for doing this!

Issue #4592, opened 2022-10-28
Can't cast to enum in index/constraint


msullivan:
```
scalar type OrderStatus extending enum<open, submitted, processing, complete, canceled>;

type Order {
  required property status -> OrderStatus;
  constraint exclusive on ((OrderStatus.open in .status));
}
```
produces `InternalServerError: functions in index expression must be marked IMMUTABLE`

It looks the problem is that the actual postgres-side cast is not marked IMMUTABLE.

EdgeQL creates casts to enums as immutable, even if they are legitimate casts that could fail. Even if we didn't, we would certain want this sort of *literal reference that cannot possibly fail* to be IMMUTABLE.

Not sure if there is a good way to fix this, though? Introduce wrapper casting functions for enums that we explicitly mark as IMMUTABLE?

msullivan:
It looks like array/string indexing has the same problem, since `_index` gets inferred as STABLE. That should be easy to fix though.

PR #4760, opened 2022-12-02
Immutable index

aljazerzen:
Fixes the comment from #4592

### DIFF
--- a/edb/pgsql/metaschema.py
+++ b/edb/pgsql/metaschema.py
@@ -1998,8 +1998,9 @@ class ArrayIndexWithBoundsFunction(dbops.Function):
             args=[('val', ('anyarray',)), ('index', ('bigint',)),
                   ('detail', ('text',))],
             returns=('anyelement',),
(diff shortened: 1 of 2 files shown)


aljazerzen:
When approved, I'll add the patch for backporting

aljazerzen:
Wait, should this also apply to `std::_slice`?

elprans:
> Wait, should this also apply to `std::_slice`?

Yes

msullivan:
I tested `_slice` at some point and it seemed fine? Add tests, I guess, but it might not need a change

msullivan (approved):

msullivan:
Or wait, should we just make raise_on_null immutable?

aljazerzen:
> Or wait, should we just make raise_on_null immutable?

I think we don't:

https://github.com/edgedb/edgedb/blob/immutable-index/edb/pgsql/metaschema.py#L1208-L1215

PR #4761, opened 2022-12-02
Add documentation contributing guidelines

raddevon:
I've created some basic guidelines for contributing to our documentation. Would love to get feedback on what is here. I'm sure I've missed some things.

I've been collecting some style considerations as well with the idea of creating a style guide, and I think this is the logical place to expose those.

Since this file is intended to be contributing guidelines generically (not specific to documentation), I think it would make sense to write up and include a basic set of guidelines for contributing code here as well. I don't feel like I'm prepared to do that on my own, but I'm happy to collaborate on it.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -1,0 +1,57 @@
+How to contribute to EdgeDB
+===========================
+
+Thank you for contributing to EdgeDB! We love our open source community and
+want to foster a healthy contributor ecosystem.
+
+To make sure the project can continue to improve quickly, we have a few
+guidelines designed to make it easier for your contributions to make it into
+the project.
+
+Contributing documentation
+--------------------------
+
+- **Avoid changes that don't fix an obvious mistake or add clarity.** This is
+  subjective, but try to look at your changes with a critical eye. Do they fix
+  errors in the original like misspellings or typos? Do they make existing
+  prose more clear or accessible while maintaining accuracy? If you answered
+  "yes" to either of those questions, this might be a great addition to our
+  docs! If not, consider starting a discussion instead to see if your changes
+  might be the exception to this guideline before submitting.
+- **Keep commits and pull requests small.** We get it. It's more convenient to
+  throw all your changes into a single pull request or even into a single
+  commit. The problem is that, if some of the changes are good and others don't
+  quite work, having everything in one bucket makes it harder to filter out the
+  great changes from those that need more work.
+- **Make spelling and grammar fixes in a separate pull request from any content
+  changes.** These changes are quick to check and important to anyone reading
+  the docs. We want to make sure they hit the live documentation as quickly as
+  possible without being bogged down by other changes that require more
+  intensive review.
+
+Documentation style
+~~~~~~~~~~~~~~~~~~~
+
+- **Lines should be no longer than 79 characters.**
+- **Remove trailing whitespace or whitespace on empty lines.**
+- **Surround references to parameter named with asterisks.** You may be tempted
+  to surround parameter names with double backticks (````param````). We avoid
+  that in favor of ``*param*``, in order to distinguish between parameter
+  references and inline code (which *should* be surrounded by double
+  backticks).
+- **EdgeDB is singular.** Choose "EdgeDB is" over "EdgeDB are" and "EdgeDB
+  does" over "EdgeDB do."
+- **Use American English spellings.** Choose "color" over "colour" and
+  "organize" over "organise."
+- **Use the Oxford comma.** When delineating a series, place a comma between
+  each item in the series, even the one with the conjunction. Use "eggs, bacon,
+  and juice" rather than "eggs, bacon and juice."
+- **Write in the simplest prose that is still accurate and expresses everything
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.rst
```diff
@@ -0,0 +1,57 @@
...
+- **Keep commits and pull requests small.** We get it. It's more convenient to
+  throw all your changes into a single pull request or even into a single
+  commit. The problem is that, if some of the changes are good and others don't
+  quite work, having everything in one bucket makes it harder to filter out the
+  great changes from those that need more work.
+- **Make spelling and grammar fixes in a separate pull request from any content
+  changes.** These changes are quick to check and important to anyone reading
+  the docs. We want to make sure they hit the live documentation as quickly as
+  possible without being bogged down by other changes that require more
+  intensive review.
```
i0bs, line 30: We should give instructions for how to build documentation from source

```suggestion
  intensive review.

Please see EdgeDB's guide for `building documentation <https://www.edgedb.com/docs/guides/contributing#writing-documentation>`_
from source.
```

i0bs (changes requested):
A few questions regarding the proposal:

- There's a `@Contributor` role for EdgeDB's Discord server that only one user has. Would this role continue to be handed out for those who make precedential contributions to the project? Should we have a section clarifying the role and how contributions correlate to receiving it?
- Guidelines are mentioned as different sections for documentation, but there are no requirements. Should users be creating Issues first and then create a Pull Request? Do they need to have any additional tools for code-related contributions, e.g. compiler changes needing to compare benchmark scores?

Some suggestions for adding sections:

- Having a section dedicated to explaining commit style and size may help. Some common questions that could be asked:
  - Do you use [conventional commits?](https://www.conventionalcommits.org/en/v1.0.0/#specification)
  - What's the maximum preferred amount of commits allowed in a Pull Request?
  - What are exceptions (if any) to a Pull Request allowing it to be large, e.g. making wide scale changes to compiler logic?
- General guidelines/requirements for EdgeDB, including all other client libraries, could be kept separate from dedicated sections for documentation, code and etc.

aljazerzen (approved):

zackelan (approved):

raddevon:
> * There's a `@Contributor` role for EdgeDB's Discord server that only one user has. Would this role continue to be handed out for those who make precedential contributions to the project? Should we have a section clarifying the role and how contributions correlate to receiving it?

I'm not seeing this role currently in the Discord.
 
> * Guidelines are mentioned as different sections for documentation, but there are no requirements. Should users be creating Issues first and then create a Pull Request? Do they need to have any additional tools for code-related contributions, e.g. compiler changes needing to compare benchmark scores?

I don't want to be super heavy handed here if we can help it. Maybe adding this under the first two paragraphs addresses your concern? I'm open to other suggestions.

```rest
These are guidelines rather than hard rules. If you want to submit a pull
request that strays from these, it might be a good idea to start a discussion
about it first. Otherwise, it's likely your pull request will not be merged.
```

> Having a section dedicated to explaining commit style and size may help. Some common questions that could be asked:
>       
>  * Do you use [conventional commits?](https://www.conventionalcommits.org/en/v1.0.0/#specification)

This seems to me like a heavy burden to put on contributors. Let's try something looser to start, and we can tighten it up if it's becoming too much of a burden. If the commit subject tells me what was done and the body tells me why (where that's not self-evident), I'm good with that.

>  * What's the maximum preferred amount of commits allowed in a Pull Request?

I'm not sure if there's a way to set a maximum that makes sense. Seems like anything would be arbitrary. I've talked to the team though, and it sounds like we want every PR to be addressing an issue in Github. That might provide a mechanism for us to suggest a scope for a PR that would avoid the problem of very large PRs.

>  * What are exceptions (if any) to a Pull Request allowing it to be large, e.g. making wide scale changes to compiler logic?

This could also be addressed in the issue discussion.

>  * General guidelines/requirements for EdgeDB, including all other client libraries, could be kept separate from dedicated sections for documentation, code and etc.

My thinking is that we could provide a set of guidelines that are general enough that they could apply equally to any repo. I'd like to iron those out here and then roll them out across the repos. If we start finding that individual projects have their own needs when it comes to contributing guidelines, we can start to diverge them at that point. To start, I'd like to focus on getting something fairly generic in place that balances making it easy for maintainers to review PRs against any additional burden placed on contributors. My hope is that we can do that without getting so specific as to necessitate dramatically different guidelines from one project to the next.

i0bs (approved):

i0bs:
```
These are guidelines rather than hard rules. If you want to submit a pull
request that strays from these, it might be a good idea to start a discussion
about it first. Otherwise, it's likely your pull request will not be merged.
```

This works, yeah. Main thought for suggestion is so Issues can be used for tracking something while keeping review discussion separate in the Pull Request

elprans (approved):

raddevon:
Guidelines for contributing code are a little thin, but that's awesome if they're sufficient as they are. I want the guidelines to be as minimal as they can be.

I'll leave this open for a while in the event anyone has further input. Otherwise, I intend to merge on Monday.

PR #4773, opened 2022-12-05
Fix: reset schema

### LINKED ISSUE #4766: `RESET SCHEMA TO INITIAL` leaves database in broken state
Somehow the schema version id (`__schema_version__`/`edgedb."_SchemaSchemaVersion"`) stored in the database is getting fouled up after running `RESET SCHEMA TO INITIAL`, such that subsequent DDL commands produce
```
edgedb.errors.TransactionSerializationError: Cannot serialize DDL: 0f51bd2b-72aa-11ed-9fc6-61c4bb93e1e1
```

It seems that `RESET SCHEMA TO INITIAL` leaves the database in a broken state where no DDL can be run successfully. This doesn't cause any problems in the normal testing mode because `test_edgeql_migration_reset_schema` happens to be the last test in that test suite that actually tries to modify the database.
(I ran into this trying to debug the NOCREATEDB test mode, which needs to reuse the same database for everything and where this ends up as a huge problem)

This can be observed by adding a `await self.con.query('create module default')` to the end of `test_edgeql_migration_reset_schema`.

(Related to that, reset schema should probably include an empty `module default` in the schema)

aljazerzen:
Fixes #4714
Closes #4766
Removes need for #4767

### DIFF
--- a/edb/server/compiler/ddl.py
+++ b/edb/server/compiler/ddl.py
@@ -1103,10 +1103,22 @@ def _reset_schema(
         s_schema.FlatSchema(),
         current_tx.get_global_schema(),
     )
+    empty_schema = s_ddl.apply_sdl(  # type: ignore
+        qlast.Schema(
+            declarations=[
+                qlast.ModuleDeclaration(
+                    name=qlast.ObjectRef(name='default'),
+                    declarations=[],
+                )
+            ]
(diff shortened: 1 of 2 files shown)


msullivan (approved):
Thanks for the quick fix!

PR #4774, opened 2022-12-05
Improve typing in pg dbops

aljazerzen:
I added typing annotations while debugging and trying to understand the code.

### DIFF
--- a/edb/pgsql/dbops/functions.py
+++ b/edb/pgsql/dbops/functions.py
@@ -20,6 +20,7 @@
 from __future__ import annotations
 
 import textwrap
+from typing import *
 
 from ..common import qname as qn
 from ..common import quote_ident as qi
@@ -77,7 +78,13 @@ class FunctionExists(base.Condition):
 
 
 class FunctionOperation:
-    def format_args(self, args, has_variadic, *, include_defaults=True):
+    def format_args(
+        self,
+        args: Sequence[str | Tuple[str, str, str]],
+        has_variadic: bool,
+        *,
+        include_defaults=True,
+    ):
         if not args:
             return ''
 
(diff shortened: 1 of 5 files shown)


elprans:
Did you run these past mypy?

aljazerzen:
Yes, it doesn't report errors. And it should start typechecking functions as soon as they get type annotations, right?

elprans:
Our default mypy config is pretty weak and might ignore errors.  My advice would be to add the newly-typed modules to one of stricter `tool.mypy.overrides` in `pyproject.toml` for more thorough checking.

edb/pgsql/dbops/functions.py
```diff
@@ -168,9 +191,15 @@ def code(self, block: base.PLBlock) -> str:
...
 class DropFunction(ddl.DDLOperation, FunctionOperation):
     def __init__(
-            self, name, args, *,
-            if_exists=False,
-            has_variadic=False, conditions=None, neg_conditions=None):
+        self,
+        name: Tuple[str, ...],
+        args: Sequence[str | Tuple[str, str, str]],
+        *,
+        if_exists=False,
```
msullivan, line 198: `: bool`?

edb/pgsql/dbops/tables.py
```diff
@@ -327,7 +328,11 @@ def code(self, block: base.PLBlock) -> str:
 
 
 class AlterTableBaseMixin:
-    def __init__(self, name, contained=False, **kwargs):
+
+    name: Tuple[str, ...]
+    contained: bool
+
+    def __init__(self, name: Tuple[str, ...], contained=False, **kwargs):
```
msullivan, line 335: `: bool`?

msullivan (approved):
Sure. Agree with Elvis that it is probably worth trying making some stuff stricter in pyproject.toml`

The usefulness of this will be pretty limited until (unless) we enable errors in `pgsql/delta.py`... And that's gonna be tough for a few reasons (one I think being that a bunch of dodgy python magic happens)

PR #4775, opened 2022-12-05
Immutable casts

### LINKED ISSUE #4592: Can't cast to enum in index/constraint

aljazerzen:
Closes #4592

### DIFF
--- a/edb/edgeql/compiler/casts.py
+++ b/edb/edgeql/compiler/casts.py
@@ -216,6 +216,9 @@ def compile_cast(
                 ctx=ctx,
             )
 
+    # Constraints and indexes require an immutable expression, but pg cast is
+    # only stable. In this specific case, we use cast wrapper function that
+    # is declared to be immutable.
     if new_stype.is_enum(ctx.env.schema):
         objctx = ctx.env.options.schema_object_context
         if objctx in (s_constr.Constraint, s_indexes.Index):
--- a/edb/edgeql/compiler/casts.py
+++ b/edb/edgeql/compiler/casts.py
@@ -96,19 +96,6 @@ def compile_cast(
             f'`...[IS {new_stype.get_displayname(ctx.env.schema)}]` instead',
             context=srcctx)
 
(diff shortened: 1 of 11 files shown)


edb/edgeql/compiler/casts.py
```diff
@@ -225,6 +216,13 @@ def compile_cast(
                 ctx=ctx,
             )
 
+    if new_stype.is_enum(ctx.env.schema):
```
msullivan, line 219: Add a comment why

edb/edgeql/compiler/options.py
```diff
@@ -116,13 +117,15 @@ class CompilerOptions(GlobalCompilerOptions):
 
     #: A set of schema types and links that should be treated
     #: as singletons in the context of this compilation.
-    singletons: Collection[Union[s_types.Type, s_pointers.Pointer]] = (
-        frozenset())
+    singletons: Collection[
+        s_types.Type | s_pointers.Pointer | s_constraints.ConsistencySubject
```
msullivan, line 121: I think that all ConsistencySubjects are already `Type` or `Pointer`. I'd add an assert at the callsite instead of changing the type here (and then would try to not make the downstream changes to type_remaps and view_sets?)

msullivan (approved):
Looks good. 

(Sorry, lost track of this one)

PR #4776, opened 2022-12-05
Use @> for std::contains with array parameters

stealth-toucan:
At the moment, EdgeDB uses `array_position` in `std::contains` when the haystack is an array. This change modifies `std::contains` to use the `@>` operator; this improves performance in the presence of indexes that support the `array_ops` operator class.

### DIFF
--- a/edb/buildmeta.py
+++ b/edb/buildmeta.py
@@ -44,7 +44,7 @@ from edb.common import verutils
 
 
 # Increment this whenever the database layout or stdlib changes.
-EDGEDB_CATALOG_VERSION = 2023_03_30_00_00
+EDGEDB_CATALOG_VERSION = 2023_03_30_00_01
 EDGEDB_MAJOR_VERSION = 3
 
 
--- a/edb/edgeql/compiler/options.py
+++ b/edb/edgeql/compiler/options.py
@@ -33,6 +33,8 @@ if TYPE_CHECKING:
     from edb.schema import pointers as s_pointers
     from edb.ir import pathid
 
+    SourceOrPathId = Union[s_types.Type, s_pointers.Pointer, pathid.PathId]
+
 
 @dataclass
 class GlobalCompilerOptions:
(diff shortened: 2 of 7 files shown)


msullivan:
This looks reasonable.

Did you test if it seems to help with indexes?

elprans:
Yeah `@>` is part of [`array_ops`](https://www.postgresql.org/docs/current/gin-builtin-opclasses.html) operator class which is indexable by GIN.  So, this wouldn't immediately help with the default b-tree index type we currently use, but will help once @vpetrovykh's work on exposing additional index types lands.

msullivan:
Another question here is: will postgres use a GIN index in response to a `@>` inside a function call, or will we need to special case it in our compile to take advantage of this. I guess postgres can probably inline the function? Would you be able to test *that*? Otherwise we'll need to take a different approach here

stealth-toucan:
When `needle` is an array, postgres will inline the function and take advantage of the index. When `needle` is an array element, postgres will fall back to a sequential scan; this has to do with postgres' criteria for inlining strict functions.

Unfortunately, I don't see a way to finagle inlining using SQL alone, but inlining can be achieved by making the function non-strict (`SET impl_is_strict := false;`). Technically we can write a strict function for this, but postgres will refuse to recognize it as such for the purposes of inlining because of decisions taken many years ago.

All of the above have been experimentally verified by manipulating the postgres backend of a live EdgeDB instance.

How would you like to proceed?

msullivan:
Sorry for not following up sooner. Thank you very much for your analysis and PR, it made this very easy on my side.

What I am going to do is:
 * Put up another PR that fixes the one downside that `SET impl_is_strict := false;` has (that it means the function can't be used in constraints/indexes)
 * Add `SET impl_is_strict := false;` to the element-wise `std::contains` and merge this PR

msullivan:
I dropped the `std::contains(haystack: array<anytype>, needle: array<anytype>)` version because it is inconsistent with the existing behavior of `contains`. It could probably be introduced with a name like `contains_all`, but I haven't done that because it seems a little marginal (and because I don't want to spend time writing tests for a new function right now :P).

@stealth-toucan if having that version of the function is important to you, could you please file another issue (or even open a new PR, with tests)?

msullivan:
Oh, annoyingly making `contains` not strict actually does still cause some more problems, because scalar constraints will be apparently still be evaluated on NULL values. I think there we actually want to always filter NULLs out, since it would be pretty nonsensical to be able to write a scalar constraint that coalesces or uses ?= or whatnot.
Need to check what happens with pointer constraints too

msullivan:
Alright, I ended up actually needing to do some stuff to get it to work

elprans (approved):

PR #4782, opened 2022-12-06
Cleanups and refactors of Expression

msullivan:
Two big typing changes:
* Make irast officially return irast.Statement. Basically every call
  site asserts this.
* Try to track whether an Expression has an irast in the type system.

These clean up a decent amount of crud and reduce the amount of crud I will
need to *add* to make pgsql/delta typecheck cleanly.

Also make a bunch of the classmethods into regular methods. Many of
the call sites were doing `type(expr).compiled(expr, ...)`, which is
just silly. I split that out if people don't like it, though.


elprans (approved):
Nice

PR #4800, opened 2022-12-08
Add type introspection to Parse and Execute

quinchs:
## Summary
This PR adds the ability to introspect type information when sending `Parse` and `Execute`.

### Overview
A new compilation flag `INTROSPECT_TYPE_INFORMATION = 1 << 3` has been added, when this flag is set, a type annotation descriptor is returned with the result descriptors:
```c
struct TypeIntrospectionDescriptor {
    // Indicates that this is an
    // Type introspection descriptor.
    uint8        type = 0xf0

    // ID of the descriptor the
    // annotation is for.
    uuid         id;

    // Annotation text.
    string       annotation;
};
```
with the annotation text being a binary encoded struct with the following format:
```c
struct TypeInformation {
  // The name of the type.
  string            name;

  // The id of the type in EdgeDB.
  uuid              id;

  // Ancestors of the type.
  uint16            ancestors_length;
  TypeInformation[] ancestors;
}
```

**Remarks**
The type information is encoded as a string to maintain backwards compatibility with the [TypeAnnotationDescriptor](https://www.edgedb.com/docs/reference/protocol/typedesc#type-annotation-descriptor), although, since this data is gatekept by a new flag, maybe we could change the format of the `TypeIntrospectionDescriptor` to not be encoded as a string?

### DIFF
--- a/edb/protocol/messages.py
+++ b/edb/protocol/messages.py
@@ -521,9 +521,10 @@ class Capability(enum.IntFlag):
 
 class CompilationFlag(enum.IntFlag):
 
-    INJECT_OUTPUT_TYPE_IDS   = 1 << 0    # noqa
-    INJECT_OUTPUT_TYPE_NAMES = 1 << 1    # noqa
-    INJECT_OUTPUT_OBJECT_IDS = 1 << 2    # noqa
+    INJECT_OUTPUT_TYPE_IDS       = 1 << 0    # noqa
+    INJECT_OUTPUT_TYPE_NAMES     = 1 << 1    # noqa
+    INJECT_OUTPUT_OBJECT_IDS     = 1 << 2    # noqa
+    INTROSPECT_TYPE_INFORMATION  = 1 << 3    # noqa
 
 
 class ErrorSeverity(enum.Enum):
--- a/edb/protocol/messages.py
+++ b/edb/protocol/messages.py
@@ -521,10 +521,12 @@ class Capability(enum.IntFlag):
 
 class CompilationFlag(enum.IntFlag):
 
-    INJECT_OUTPUT_TYPE_IDS       = 1 << 0    # noqa
-    INJECT_OUTPUT_TYPE_NAMES     = 1 << 1    # noqa
-    INJECT_OUTPUT_OBJECT_IDS     = 1 << 2    # noqa
-    INTROSPECT_TYPE_INFORMATION  = 1 << 3    # noqa
+    INJECT_OUTPUT_TYPE_IDS      = 1 << 0    # noqa
+    INJECT_OUTPUT_TYPE_NAMES    = 1 << 1    # noqa
+    INJECT_OUTPUT_OBJECT_IDS    = 1 << 2    # noqa
+    INTROSPECT_TYPE_INFORMATION = 1 << 3    # noqa
+    DETAILED_TYPE_INFORMATION   = 1 << 4    # noqa
+    DETAILED_SCALAR_INFORMATION = 1 << 5    # noqa
 
 
 class ErrorSeverity(enum.Enum):
--- a/edb/server/compiler/compiler.py
+++ b/edb/server/compiler/compiler.py
@@ -109,6 +109,8 @@ class CompileContext:
     inline_typenames: bool = False
     inline_objectids: bool = True
     introspect_type_information: bool = False
+    detailed_type_information: bool = False
+    detailed_scalar_information: bool = False
     schema_object_ids: Optional[Mapping[s_name.Name, uuid.UUID]] = None
     source: Optional[edgeql.Source] = None
     backend_runtime_params: pg_params.BackendRuntimeParams = (
@@ -468,6 +470,8 @@ class Compiler:
                     inline_typeids=False,
                     inline_typenames=True,
                     introspect_type_information=False,
+                    introspect_detailed_type_information=False,
+                    introspect_detailed_scalar_information=False,
                     json_parameters=False,
                     source=source,
                     protocol_version=protocol_version,
@@ -509,6 +513,8 @@ class Compiler:
         inline_objectids: bool = True,
         json_parameters: bool = False,
         introspect_type_information: bool = False,
+        detailed_type_information: bool = False,
+        detailed_scalar_information: bool = False,
     ) -> Tuple[dbstate.QueryUnitGroup,
                Optional[dbstate.CompilerConnectionState]]:
 
@@ -547,6 +553,8 @@ class Compiler:
             inline_typenames=inline_typenames,
             inline_objectids=inline_objectids,
             introspect_type_information=introspect_type_information,
+            detailed_type_information=detailed_type_information,
+            detailed_scalar_information=detailed_scalar_information,
             skip_first=skip_first,
             json_parameters=json_parameters,
             source=source,
@@ -580,6 +588,8 @@ class Compiler:
         inline_objectids: bool = True,
         json_parameters: bool = False,
         introspect_type_information: bool = False,
+        detailed_type_information: bool = False,
+        detailed_scalar_information: bool = False,
         expect_rollback: bool = False,
     ) -> Tuple[dbstate.QueryUnitGroup, dbstate.CompilerConnectionState]:
         if (
@@ -605,6 +615,8 @@ class Compiler:
             inline_typenames=inline_typenames,
             inline_objectids=inline_objectids,
             introspect_type_information=introspect_type_information,
+            detailed_type_information=detailed_type_information,
+            detailed_scalar_information=detailed_scalar_information,
             skip_first=skip_first,
             source=source,
             protocol_version=protocol_version,
@@ -1226,6 +1238,8 @@ def _compile_ql_query(
             ir.view_shapes, ir.view_shapes_metadata,
             inline_typenames=ctx.inline_typenames,
             introspect_type_information=ctx.introspect_type_information,
+            detailed_type_information=ctx.detailed_type_information,
+            detailed_scalar_information=ctx.detailed_scalar_information,
             protocol_version=ctx.protocol_version)
     else:
         out_type_data, out_type_id = \
--- a/edb/server/compiler/compiler.py
+++ b/edb/server/compiler/compiler.py
@@ -108,6 +108,7 @@ class CompileContext:
     inline_typeids: bool = False
     inline_typenames: bool = False
     inline_objectids: bool = True
+    introspect_type_information: bool = False
     schema_object_ids: Optional[Mapping[s_name.Name, uuid.UUID]] = None
     source: Optional[edgeql.Source] = None
     backend_runtime_params: pg_params.BackendRuntimeParams = (
@@ -466,6 +467,7 @@ class Compiler:
                     implicit_limit=implicit_limit,
                     inline_typeids=False,
                     inline_typenames=True,
+                    introspect_type_information=False,
                     json_parameters=False,
                     source=source,
                     protocol_version=protocol_version,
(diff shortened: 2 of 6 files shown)


tailhook:
1. Do we have to bump protocol version? I think this is the only way for the client to find out whether this flag can be used. @elprans ?
2. I think this kind of encoding of parents can be very inefficient with multiple inheritance in a diamond pattern:
    ```
      A
     / \
    B   C
     \ /
      D
      E
      F
    ```

    Will be encoded as:

    ```
    {A {B {D {E {F}}}, C {D {E {F}}}}}
    ```

    Because both B and C have D, E, F as ancestors. And this can quickly become a combinatorical explosion with a little bit more inheritance.

3. Minor, but: I'm not sure why you have picked `0xf0` as the annotation marker? It's neither beginning nor the end of the range.

> Although, since this data is gatekept by a new flag, maybe we could change the format of the TypeIntrospectionDescriptor to not be encoded as a string?

Yes I think so.

quinchs:
> I think this kind of encoding of parents can be very inefficient with multiple inheritance in a diamond pattern
Maybe we can encode ancestor information like so:
```c
struct TypeIntrospectionDescriptor {
    // Indicates that this is an
    // Type introspection descriptor.
    uint8        type = 0x0a

    // ID of the descriptor the
    // annotation is for.
    uuid              id;

    // Name of the type.
    string            name;

    // ID of the type.
    uuid              type_id;

    // Indexes within the 'ancestors' collection 
    // that are the direct parents of this type.
    uint16            ancestors_index_length;
    uint16[]          ancestors_indexes;

    // All ancestors of the current type.
    uint16            ancestors_length;
    TypeInformation[] ancestors;
    
};
```
with the annotation text being a binary encoded struct with the following format:
```c
struct TypeInformation {
  // The name of the type.
  string            name;

  // The id of the type in EdgeDB.
  uuid              id;

  // Ancestors of the type.
  uint16            ancestors_index_length;
  uint16[]          ancestors_indexes;
}
```

With this approach, `TypeInformation` only contains indexes pointing to the descriptors `ancestors` array, removing the duplicate definitions.

quinchs:
Alright here are some of the changes I've made to the structure:

```c
struct TypeIntrospectionDescriptor {
    // Indicates that this is an
    // Type introspection descriptor.
    uint8        type = 0xfe

    // ID of the descriptor the
    // annotation is for.
    uuid              id;

    // Name of the type.
    string            name;

    // ID of the type.
    uuid              type_id;

    // Indexes within the 'ancestors' collection 
    // that are the direct parents of this type.
    uint16            ancestors_index_length;
    uint16[]          ancestors_indexes;

    // All ancestors of the current type.
    uint16            ancestors_length;
    TypeInformation[] ancestors;
    
};

struct TypeInformation {
  // The name of the type.
  string            name;

  // The id of the type in EdgeDB.
  uuid              id;

  // Ancestors of the type.
  uint16            ancestors_index_length;
  uint16[]          ancestors_indexes;
};
```

Should the target codec index be encoded in the `TypeIntrospectionDescriptor ` as well as descriptor id? Personally my binding can use the ID to find the correct target codec, do other bindings need the index to find the target codec?

**My implementation in edgedb-net**
https://github.com/edgedb/edgedb-net/compare/feat/protocol-type-introspection

tailhook:
> Alright here are some of the changes I've made to the structure:

That would work. But a bit confusing that you call different things by the `id` name in `TypeInspectionDescriptor` and in `TypeInformation`. I also think that `TypeInformation` can be embedded as follows:
```c
struct TypeIntrospectionDescriptor {
    // Indicates that this is an
    // Type introspection descriptor.
    uint8        type = 0xfe

    // ID of the descriptor the
    // annotation is for.
    uuid              descriptor_id;

    TypeInformation info;

    // All ancestors of the current type.
    uint16            ancestors_length;
    TypeInformation[] ancestors;
    
};

struct TypeInformation {
  // The name of the type.
  string            name;

  // The id of the type in EdgeDB.
  uuid              type_id;

  // Indexes within the 'ancestors' collection 
  // that are the direct parents of this type.
  uint16            ancestors_index_length;
  uint16[]          ancestors_indexes;
};
```

Which is basically the same binary structure as in yours definition, just defined a little bit differently.

> 
> Should the target codec index be encoded in the `TypeIntrospectionDescriptor ` as well as descriptor id? Personally my binding can use the ID to find the correct target codec, do other bindings need the index to find the target codec?
> 

I think descriptor id is enough. I think indexes were used to conserve size of the structure. So if we need smaller structure here, we probably may use indexes **instead** of descriptor ids? But probably we don't need it.

tailhook:
If we have a separate array of `ancestors` anyways. Maybe it makes sense to have single ancestors information for the whole query?

I mean if we have `User.friends` which is also a `User`. And those are different descriptors (because of different fields). Under current spec it will duplicate the whole hierarchy of the `User` type.

Generally, I'm not sure how deep our hierarchies will be and how many shared types they will have, so it's hard to determine if this optimization is needed. At least it will not create a combinatorical explosion so less important than initial optimization.

quinchs:
> If we have a separate array of ancestors anyways. Maybe it makes sense to have single ancestors information for the whole query?

Should we return just one `TypeIntrospectionDescriptor` with all types found within the query? Maybe with the structure of:

```c
struct TypeIntrospectionDescriptor {
  // Indicates that this is an
  // Type introspection descriptor.
  uint8        type = 0xfe

  // All types that are used in codecs.
  uint16 codec_typeinfo_length;
  CodecTypeInformation[] codec_typeinfo;

  // All ancestor types found within the resulting types.
  uint16            ancestors_length;
  TypeInformation[] ancestors;
};

struct CodecTypeInformation {
  // The ID of the codec this type information is for.
  uuid              codec_id;

  // The name of the type.
  string            name;

  // The id of the type in EdgeDB.
  uuid              type_id;

  // Indexes within the 'ancestors' collection 
  // that are the direct parents of this type.
  uint16            ancestors_index_length;
  uint16[]          ancestors_indexes;
}

struct TypeInformation {
  // The name of the type.
  string            name;

  // The id of the type in EdgeDB.
  uuid              type_id;

  // Indexes within the 'ancestors' collection 
  // that are the direct parents of this type.
  uint16            ancestors_index_length;
  uint16[]          ancestors_indexes;
};
```

tailhook:
> > If we have a separate array of ancestors anyways. Maybe it makes sense to have single ancestors information for the whole query?
> 
> Should we return just one `TypeIntrospectionDescriptor` with all types found within the query? Maybe with the structure of:
> 

I was thinking of something like this:
```c
struct TypeIntrospectionDescriptor {
    // Indicates that this is an
    // Type introspection descriptor.
    uint8        type = 0xfe

    // ID of the descriptor the
    // annotation is for.
    uuid              descriptor_id;
    
  // The name of the type.
  string            name;

  // The id of the type in EdgeDB.
  uuid              type_id;

  // Indexes within the 'ancestors' collection 
  // that are the direct parents of this type.
  uint16            ancestors_index_length;
  uint16[]          ancestors_indexes;  // <<< these are indexes in `AncestorsDescriptor`
}

struct AncestorsDescriptor {
    uint8        type = 0xfd;  // <<<< Different kind of descriptor

    // All ancestors of the current type.
    uint16            ancestors_length;
    TypeInformation[] ancestors;
    
};
```
Not sure if this is the best idea, though.

quinchs:
Personally I don't like the idea of adding a second descriptor, It adds an extra bit of code complexity to bindings to store this descriptor in memory while building the codecs. I think combining the two together like [this](https://github.com/edgedb/edgedb/pull/4800#issuecomment-1346933138) makes it simpler to both write in the server and implement for bindings.

quinchs:
This format proves to be very easy to implement for bindings (at least in my case):

```c
struct TypeIntrospectionDescriptor {
  // Indicates that this is an
  // Type introspection descriptor.
  uint8                  type = 0xfe

  // All ancestor types found within the resulting types.
  uint16                 ancestors_length;
  TypeInformation[]      ancestors;

  // All types that are used in codecs.
  uint16                 codec_typeinfo_length;
  CodecTypeInformation[] codec_typeinfo;
};

struct CodecTypeInformation {
  // The ID of the codec this type information is for.
  uuid              codec_id;

  // The name of the type.
  string            name;

  // The ID of the type in EdgeDB.
  uuid              type_id;

  // Indexes within the 'ancestors' collection 
  // that are the direct parents of this type.
  uint16            ancestors_index_length;
  uint16[]          ancestors_indexes;
}

struct TypeInformation {
  // The name of the type.
  string            name;

  // The id of the type in EdgeDB.
  uuid              type_id;

  // Indexes within the 'ancestors' collection 
  // that are the direct parents of this type.
  uint16            ancestors_index_length;
  uint16[]          ancestors_indexes;
};
```

I've put the ancestor types first so the array can be fully read before being indexed by the codec information, allowing indexing when deserializing the codec type information.

Here's the updated implementation in .NET https://github.com/edgedb/edgedb-net/blob/feat/protocol-type-introspection/src/EdgeDB.Net.Driver/Binary/Descriptors/TypeIntrospectionDescriptor.cs

elprans:
> The ID of the codec this type information is for.

What is "codec" here?

quinchs:
> What is "codec" here?

Sorry, its the ID of the type descriptor that the annotation is for

quinchs:
I've implemented detailed scalar introspection, as well as a way to lower overhead with type introspection:

**New Flags**
+ `DETAILED_TYPE_INFORMATION = 1 << 4`
+ `DETAILED_SCALAR_INFORMATION = 1 << 5`

### Scalar changes
As before, specifying the `INJECT_OUTPUT_TYPE_NAMES` flag, you will receive an annotation with the type name of scalars:
```c
struct ScalarAnnotationDescriptor {
    // Indicates that this is a
    // Type Annotation descriptor.
    uint8        type = 0xff;

    // ID of the descriptor the
    // annotation is for.
    uuid         id;

    // The name of the scalar type.
    string       name;
};
```
By adding the new `DETAILED_SCALAR_INFORMATION` flag, the server will return ancestors information as a `ScalarAnnotationDescriptor` where each ancestors name and id is included, and a new `DetailedScalarAnnotationDescriptor`:
```c
struct DetailedScalarAnnotationDescriptor {
    // Indicates that this is a Detailed 
    // Type Annotation descriptor.
    uint8        type = 0xfe;

    // ID of the descriptor the
    // annotation is for.
    uuid         id;

    // The name of the scalar type.
    string       name;

    // The ancestors of the scalar type
    uint32      ancestors_length;
    uuid[]       ancestors;
}
```
where each id in the `ancestors` array points to a `ScalarAnnotationDescriptor` returned earlier. This does mean that bindings have to maintain state of all annotations returned to preform the linking of uuids but this shouldn't be an issue.

### Type introspection changes
When only specifying the `INTROSPECT_TYPE_INFORMATION` flag, the server returns only the type name and id - no ancestral information:
```c
struct TypeAnnotationDescriptor {
    // Indicates that this is a 
    // Type Introspection descriptor.
    uint8        type = 0xfc;

    // ID of the descriptor the 
    // annotation is for.
    uuid         id;

    // The name of the type.
    string       name;
}
```
By adding the new `DETAILED_TYPE_INFORMATION` flag, a `DetailedTypeAnnotationDescriptor` is returned for each descriptor that describes an object:
```c
struct DetailedTypeAnnotationDescriptor {
    // Indicates that this is a
    // Type introspection descriptor.
    uint8        type = 0xfd;

    // ID of the descriptor the
    // the annotation is for.
    uuid         id;

    // The name of the type.
    string       name;

    // The id of the type.
    uuid         type_id;

    // The direct ancestors ids of this type
    uint32      bases_length;
    uuid[]       bases;
};

struct AncestorTypeAnnotationDescriptor {
    // Indicates that this is a
    // Ancestor Type Annotation descriptor.
    uint8        type = 0xfb;

    // ID of the type.
    uuid         id;

    // The name of the type.
    string       name;

    // The direct ancestors ids of this type.
    uint32     bases_length;
    uuid[]      bases;
}

```
The ancestors of the object are sent as `AncestorTypeAnnotationDescriptor` described above, mapping `ancestors` collection to the `id` property of the ancestors' descriptor.

---

This format is again up to change, I've tried to meet all requirements without making a breaking change to the protocol, maybe in 3.0 we can remove the scalar flag as including scalar ancestor information *should* be default behavior.

elprans:
Superseded by #5334

PR #4807, opened 2022-12-12
Fix: computed from id

### LINKED ISSUE #4781: Can't assign `.id` as body of computed
- OS Version: ubuntu 22.04

Steps to Reproduce:
```bash
$ edgedb
EdgeDB 2.7+81de4dd (repl 2.3.0-dev.895+cba06af)
edgedb> SELECT schema::Type {
.......   XYZ := .id,
....... };
error: QueryError: cannot assign to XYZ
  ┌─ query:2:3
  │
2 │   XYZ := .id,
  │   ^^^^^^^^^^ consider enabling the "allow_user_specified_id" configuration parameter to allow setting custom object ids
```

aljazerzen:
Fixes #4781

The problem was that `XYZ` was a derived property which is extending `Object#id`. Because of that, `is_id_pointer` evaluates to True.

This commit basically moves the two checks for id modification and updates of readonly to an earlier place, where the properties have not yet been derived and `ptrcls` is simply None.

### DIFF
--- a/edb/edgeql/compiler/viewgen.py
+++ b/edb/edgeql/compiler/viewgen.py
@@ -880,6 +880,43 @@ def _normalize_view_ptr_expr(
                     f'(expecting {" or ".join(expected)})'
                 )
 
+    # Prohibit update of readonly
+    if exprtype.is_update() and ptrcls and ptrcls.get_readonly(ctx.env.schema):
+        raise errors.QueryError(
+            f'cannot update {ptrcls.get_verbosename(ctx.env.schema)}: '
+            f'it is declared as read-only',
+            context=compexpr and compexpr.context,
+        )
+
+    # Prohibit invalid operations on id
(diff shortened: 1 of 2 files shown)


elprans (commented):
Please add a test.

msullivan (approved):
This seems reasonable, but might it be better to make `is_id_pointer` only be true when the pointer is also named `id`, or some such?

aljazerzen:
Is id always named id? Can we do some kind of view/computed/alias that would make assigning to id possible via some other ident?

msullivan:
> Is id always named id? Can we do some kind of view/computed/alias that would make assigning to id possible via some other ident?

Yeah it should be. Don't think so.

msullivan:
I'm merging this now because I want to put it out in a release, but could you still please follow up on the suggested change?

PR #4815, opened 2022-12-13
Parse SQL transaction and set statments

aljazerzen:
@fantix Here are the statements I promised. You can merge them into your branch or we can do this in two separate PRs.


msullivan (approved):

edb/pgsql/ast.py
```diff
@@ -1005,3 +1005,87 @@ class IteratorCTE(ImmutableBase):
     path_id: irast.PathId
     cte: CommonTableExpr
     parent: typing.Optional[IteratorCTE]
+
+
+class Statement(Base):
```
fantix, line 1010: https://github.com/edgedb/edgedb/blob/9a396dcd76d1b8bbf8c78c2b42d7891b0aaa27bf/edb/pgsql/parser/__init__.py#L30

This would return `List[pgast.Base]` instead? Or perhaps `List[Union[pgast.Query, pgast.Statement]]`?

aljazerzen: Yep, you are right - type annotations inside ast_builder are more like suggestions than hard requirements, I guess.

aljazerzen:
This now also includes pg_catalog introspection and fixes/extensions for a few different pg AST nodes.

edb/pgsql/parser/ast_builder.py
```diff
@@ -214,6 +264,144 @@ def _build_delete_stmt(n: Node, c: Context) -> pgast.DeleteStmt:
...
+        isolation_mode=isolation_mode or pgast.IsolationMode.READ_COMMITTED,
+        access_mode=access_mode or pgast.AccessMode.READ_WRITE,
+        deferrable=deferrable or False,
+    )
+
+
+def _build_prepare(n: Node, c: Context) -> pgast.PrepareStmt:
+    return pgast.PrepareStmt(
+        name=n["name"],
+        argtypes=_maybe_list(n, c, "params", _build_base_expr),
```
fantix, line 394:
```suggestion
        argtypes=_maybe_list(n, c, "argtypes", _build_base_expr),
```

But looks like `_build_base_expr()` doesn't support `TypeName`.

edb/pgsql/codegen.py
```diff
@@ -894,5 +894,112 @@ def visit_Set(self, node: pgast.Set) -> None:
...
+            self.write(" READ ONLY")
+        if node.deferrable:
+            self.write(" DEFERRABLE")
+
+    def visit_PrepareStmt(self, node: pgast.PrepareStmt) -> None:
+        self.write(f"PREPARE {node.name} AS ")
+        self.visit(node.query)
+
+    def visit_ExecuteStmt(self, node: pgast.ExecuteStmt) -> None:
+        self.write(f"EXECUTE {node.name}")
```
fantix, line 975: args and params are not written? Tho actually we don't need this 2 codegen - `PREPARE` and `EXECUTE` will be translated to direct `Parse` and `Bind+Execute` messages I think.

edb/pgsql/codegen.py
```diff
@@ -894,5 +894,112 @@ def visit_Set(self, node: pgast.Set) -> None:
...
+        self.write("SET ")
+        if node.scope == pgast.OptionsScope.TRANSACTION:
+            self.write("LOCAL ")
+        self.write(node.name)
+        self.write(" TO ")
+        self.visit_list(node.args)
+
+    def visit_SetTransactionStmt(self, node: pgast.SetTransactionStmt) -> None:
+        self.write("SET ")
+        if node.scope != pgast.OptionsScope.TRANSACTION:
```
fantix, line 907:
```suggestion
        if node.scope == pgast.OptionsScope.TRANSACTION:
```

edb/pgsql/codegen.py
aljazerzen: I didn't do it, but I guess I should. I wont' take long

PR #4817, opened 2022-12-13
Link tutorial from appropriate doc pages

raddevon:
This was inspired by a Discord member who suggested we link to the [upsert tutorial](https://www.edgedb.com/tutorial/data-mutations/upsert) from the appropriate page and section in the documentation. I took the opportunity to liberally add links from documentation pages to related tutorials.

### DIFF
--- a/docs/datamodel/constraints.rst
+++ b/docs/datamodel/constraints.rst
@@ -221,4 +221,4 @@ using arbitrary EdgeQL expressions. The example below uses the built-in
   * - :ref:`Introspection > Constraints <ref_eql_introspection_constraints>`
   * - :ref:`Standard Library > Constraints <ref_std_constraints>`
   * - `Tutorial > Advanced EdgeQL > Constraints
-      <https://www.edgedb.com/tutorial/advanced-edgeql/constraints>`_
+      </tutorial/advanced-edgeql/constraints>`_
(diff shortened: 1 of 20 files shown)


docs/datamodel/constraints.rst
```diff
@@ -220,3 +220,5 @@ using arbitrary EdgeQL expressions. The example below uses the built-in
   * - :ref:`DDL > Constraints <ref_eql_ddl_constraints>`
   * - :ref:`Introspection > Constraints <ref_eql_introspection_constraints>`
   * - :ref:`Standard Library > Constraints <ref_std_constraints>`
+  * - `Tutorial > Advanced EdgeQL > Constraints
+      <https://www.edgedb.com/tutorial/advanced-edgeql/constraints>`_
```
elprans, line 224: @1st1 @jaclarke do we have some intersphinx capability to be able to use refs here to avoid hardcoding the URLs?

raddevon: @elprans I almost asked about this, but it didn't look like the tutorial is built in Sphinx so I assumed it isn't possible.

raddevon: @elprans These links are relative now, and I've added https://github.com/edgedb/website/issues/5 to address reference linking into the tutorial. For some reason, I had it in my head that we couldn't do relative links in Sphix when I was putting this together, so thanks for calling that out.

elprans (approved):

PR #4819, opened 2022-12-14
Get rid of the "unused alias definition" error

msullivan:
It makes incrementally debugging queries more annoying, makes writing
big DML queries where you don't care if you return the data harder,
and doesn't accomplish anything useful semantically.

We should just ditch it.

### DIFF
--- a/edb/edgeql/compiler/context.py
+++ b/edb/edgeql/compiler/context.py
@@ -186,16 +186,6 @@ class Environment:
 
     view_shapes_metadata: Dict[s_types.Type, irast.ViewShapeMetadata]
 
-    must_use_views: Dict[
-        s_types.Type,
-        Optional[Tuple[s_name.Name, Optional[parsing.ParserContext]]],
-    ]
-    """A set of views that *must* be used in an expression.
-
-    Once a view is used, we set it to None rather than removing it so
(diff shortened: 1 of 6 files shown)


msullivan:
(Once upon a time, not referencing an alias that did DML would have resulted in the DML not executing, but I fixed that a long time ago as a side effect of making with-bound DML execute the correct number of times if referenced from a `FOR` query in #2162)

elprans:
Hm, I think this would introduce undefined behavior for expressions with side effects. Do they get executed or not?

msullivan:
DML always gets executed when it appears in a WITH binding, exactly once. Where (and whether) the bound variable is referenced in the body shouldn't impact the behavior, or else things like
```
with X := (insert Foo)
for _ in {1, 2, 3} union (X)
```
would insert three times.

tailhook:
> (Once upon a time, not referencing an alias that did DML would have resulted in the DML not executing, but I fixed that a long time ago as a side effect of making with-bound DML execute the correct number of times if referenced from a `FOR` query in #2162)

Yeah, I remember this was introduced exactly to avoid this issue in users' code. This could be a language-server warning, but I agree that this is annoying as an error.

aljazerzen (approved):
These commits that only remove code are nice :D

PR #4821, opened 2022-12-14
Enforce id name

aljazerzen:
Follow up on #4807 

Should not have any effect, but will prevent future bugs.


msullivan (approved):
Looks good.

I also dropped the `std::target` check, which was made dead and didn't make any sense anyway.
(I just made the change myself because I think I'll cherry-pick it)

PR #4823, opened 2022-12-14
Add documentation for interactive shell commands

raddevon:
A couple of questions I had on this as I was working through it:

1. Is "Shell Commands" the right name for this section? We refer to them in the help at one point as "backslash commands" so maybe that is better? Maybe "REPL commands"?
2. Is `:cli:synopsis:` the right way to render the commands? It leads to some questionable highlighting, but I notice the highlighting of the command options earlier on this page are also a little wonky.

Open to any other feedback as well.

### DIFF
--- a/docs/cli/edgedb.rst
+++ b/docs/cli/edgedb.rst
@@ -143,34 +143,34 @@ many of the commands:
 - ``-s``- Show system objects
 - ``-c``- Case-sensitive pattern matching
 
-:cli:synopsis:`\\d [-v] NAME`
+:cli:synopsis:`\\d object [-v] NAME, \\describe object [-v] NAME`
   Describe schema object specified by *NAME*.
 
-:cli:synopsis:`\\ds, \\describe schema`
+:cli:synopsis:`\\ds, \\d schema, \\describe schema`
   Describe the entire schema.
 
 :cli:synopsis:`\\l, \\list databases`
   List databases.
 
-:cli:synopsis:`\\ls, \\list scalars [-sc] [PATTERN]`
+:cli:synopsis:`\\ls [-sc] [PATTERN], \\list scalars [-sc] [PATTERN]`
   List scalar types.
 
-:cli:synopsis:`\\lt, \\list types [-sc] [PATTERN]`
+:cli:synopsis:`\\lt [-sc] [PATTERN], \\list types [-sc] [PATTERN]`
   List object types.
 
-:cli:synopsis:`\\lr, \\list roles [-c] [PATTERN]`
+:cli:synopsis:`\\lr [-c] [PATTERN], \\list roles [-c] [PATTERN]`
   List roles.
 
-:cli:synopsis:`\\lm, \\list modules [-c] [PATTERN]`
+:cli:synopsis:`\\lm [-c] [PATTERN], \\list modules [-c] [PATTERN]`
   List modules.
 
-:cli:synopsis:`\\la, \\list aliases [-vsc] [PATTERN]`
+:cli:synopsis:`\\la [-vsc] [PATTERN], \\list aliases [-vsc] [PATTERN]`
   List expression aliases.
 
-:cli:synopsis:`\\lc, \\list casts [-c] [PATTERN]`
-  List casts.
+:cli:synopsis:`\\lc [-c] [PATTERN], \\list casts [-c] [PATTERN]`
+  List available conversions between types.
 
-:cli:synopsis:`\\li, \\list indexes [-vsc] [PATTERN]`
+:cli:synopsis:`\\li [-vsc] [PATTERN], \\list indexes [-vsc] [PATTERN]`
   List indexes.
 
 .. rubric:: Operations
--- a/docs/cli/edgedb.rst
+++ b/docs/cli/edgedb.rst
@@ -173,6 +173,11 @@ many of the commands:
 :cli:synopsis:`\\li [-vsc] [PATTERN], \\list indexes [-vsc] [PATTERN]`
   List indexes.
 
+.. rubric:: Database
+
+:cli:synopsis:`\\database create [NAME]`
+  Create a new database.
+
 .. rubric:: Data Operations
 
 :cli:synopsis:`\\dump FILENAME`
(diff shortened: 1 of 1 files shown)


docs/cli/edgedb.rst
```diff
@@ -129,3 +129,85 @@ Options
...
+:cli:synopsis:`\\e, \\edit [N]`
+  Spawn ``$EDITOR`` to edit the most recent history entry or history entry *N*.
+  The output of this will then be used as input into the shell.
+
+
+.. rubric:: Settings
+
+:cli:synopsis:`\\set [OPTION [VALUE]]`
+  With *VALUE*, the option named by *OPTION* will be set to the provided value.
+  If *VALUE* is omitted, the command will show the current value of *OPTION*.
```
zackelan, line 199: I think this might be more readable if these two were swapped? "if value is omitted..." followed by "if value is specified..."

docs/cli/edgedb.rst
```diff
@@ -129,3 +129,85 @@ Options
...
+  Restore the database dump at *FILENAME* into the currently connected
+  database.
+
+.. rubric:: Editing
+
+:cli:synopsis:`\\s, \\history`
+  Show a history of commands executed in the shell.
+
+:cli:synopsis:`\\e, \\edit [N]`
+  Spawn ``$EDITOR`` to edit the most recent history entry or history entry *N*.
```
zackelan, line 191: maybe a reference that `N` comes from the `\history` command above?

zackelan (commented):
if we want to match [Postgres](https://www.postgresql.org/docs/current/app-psql.html) terminology:

> Meta-commands are often called slash or backslash commands.

raddevon:
Switched this to draft because I recalled I'm waiting on some information from @tailhook before I finalize this.

`\help` says `\lm` has a `-I` option, but I don't think it actually does. It looks like `\lm DEFAULT` will match my `default` module even without `-I`. When I try to add `-I` to see what it does, the CLI complains with `Error parsing backslash command: error: Found argument '-I' which wasn't expected, or isn't valid in this context`.

I'm thinking maybe this is just an issue with the output of `\help`. Looks like `\lm [-I]` should be `\lm [PATTERN]` and `\lr [-I]` should be `\lr [PATTERN]`. Does this look right to you, @tailhook?

Hints for those commands also show a `-c` option, but that isn't documented in `\help`. The CLI will take it, but I can't tell what it does. What does that option do?

raddevon:
> if we want to match [Postgres](https://www.postgresql.org/docs/current/app-psql.html) terminology:
> 
> > Meta-commands are often called slash or backslash commands.

@zackelan I like that. I've changed it to "Backslash Commands" for now. Thank you!

raddevon:
> Switched this to draft because I recalled I'm waiting on some information from @tailhook before I finalize this.
> 
> `\help` says `\lm` has a `-I` option, but I don't think it actually does. It looks like `\lm DEFAULT` will match my `default` module even without `-I`. When I try to add `-I` to see what it does, the CLI complains with `Error parsing backslash command: error: Found argument '-I' which wasn't expected, or isn't valid in this context`.
> 
> I'm thinking maybe this is just an issue with the output of `\help`. Looks like `\lm [-I]` should be `\lm [PATTERN]` and `\lr [-I]` should be `\lr [PATTERN]`. Does this look right to you, @tailhook?
> 
> Hints for those commands also show a `-c` option, but that isn't documented in `\help`. The CLI will take it, but I can't tell what it does. What does that option do?

Got a resolution on this. `-c` replaced `-I`. Not clear if the direction was also flipped (from sensitive being default to insensitive) or if the capital letter was indicative that the option did the opposite of what it said (i.e., turning *off* case insensitivity rather than turning it *on*). Whatever the case, I've updated the documentation to reflect the current state of things.

Also submitted a PR for the CLI to update the help there: https://github.com/edgedb/edgedb-cli/pull/909

This is now ready for review.

docs/cli/edgedb.rst
```diff
@@ -129,3 +129,89 @@ Options
...
+.. rubric:: Introspection
+
+The introspection commands share a few common options that are available to
+many of the commands:
+
+- ``-v``- Verbose
+- ``-s``- Show system objects
+- ``-c``- Case-sensitive pattern matching
+
+:cli:synopsis:`\\d [-v] NAME`
```
tailhook, line 146: `\d` is currently a shortcut for `\describe` And we have `\describe object NAME` and `\describe schema`

docs/cli/edgedb.rst
```diff
@@ -129,3 +129,89 @@ Options
...
+  Describe schema object specified by *NAME*.
+
+:cli:synopsis:`\\ds, \\describe-schema`
```
tailhook, line 149:
```suggestion
:cli:synopsis:`\\ds, \\describe schema`
```

tailhook (commented):
We also support:
1. `\database {create,drop}`
2. `\migrate`, `\migration *`

Those are not described in the short help because we want that to be shorter, and these are unlikely to be commonly used from the repl

raddevon:
Thanks for the heads-up on these commands, @tailhook.

> We also support:
> 
>     1. `\database {create,drop}`

This one isn't working for me. I get `Error parsing backslash command: error: Found argument 'drop' which wasn't expected, or isn't valid in this context`. The hint suggests that only the `create` subcommand is supported for `\database`. Should I file a bug on that?

I'll get those migration commands added and your other suggestions implemented.

tailhook:
> Thanks for the heads-up on these commands, @tailhook.
> 
> > We also support:
> > ```
> > 1. `\database {create,drop}`
> > ```
> 
> This one isn't working for me. I get `Error parsing backslash command: error: Found argument 'drop' which wasn't expected, or isn't valid in this context`. The hint suggests that only the `create` subcommand is supported for `\database`. Should I file a bug on that?
> 
> I'll get those migration commands added and your other suggestions implemented.

Ah, sorry, we don't have `drop` yet. For some reason. Probably can do it together with `wipe` as part of edgedb/edgedb-cli#839

raddevon:
I think this is ready for another look, @tailhook. I've added the migration commands and `\database create`. I figure we can circle back once `\database drop` is implemented. I reworked some of the aliases based on your advice on Slack and wrote a better one-liner for `\list casts`. I believe that covers everything.

See what you think of it when you have time. Thanks!

tailhook (approved):

PR #4830, opened 2022-12-15
Allow JWT token auth in binary protocol

fantix:
Cherry-pick #4587, #4703, #4708 and #4829, and add support for both old and new CLI

Tested:

- [x] CLI 2.2.6+7eabbf9 `edgedb ui`
- [x] CLI 2.3.0 dev `edgedb ui`
- [x] UI password login (uses server-generated token)

### DIFF
--- a/edb/server/protocol/binary.pyx
+++ b/edb/server/protocol/binary.pyx
@@ -491,7 +491,7 @@ cdef class EdgeConnection(frontend.FrontendConnection):
             if self._auth_data:
                 token = self._extract_token_from_auth_data(self._auth_data)
             else:
-                token = params.get('token')
+                token = params.get('secret_key')
             self._auth_jwt(user, token)
         elif authmethod_name == 'Trust':
             self._auth_trust(user)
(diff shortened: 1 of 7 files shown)


edb/server/protocol/binary.pyx
```diff
@@ -667,22 +686,18 @@ cdef class EdgeConnection(frontend.FrontendConnection):
             ) from None
 
         if not claims.get(f"{namespace}.any_role"):
+            assert not is_legacy  # legacy CLI will only generate any_role JWTs
             token_roles = claims.get(f"{namespace}.roles")
-            if not isinstance(token_roles, dict):
+            if not isinstance(token_roles, list):
```
elprans, line 691: Legacy tokens would still have a dict here, no?

fantix: Yes, that's why I put an assertion above - I think we could also do:

```suggestion
            if not isinstance(token_roles, dict if is_legacy else list):
```

elprans: Well, yes, otherwise it will crash on a legacy token.

msullivan (approved):

AnsonHwang86:
@raddevon Can add document to this? Without document, I don't know how to use JWT in edgedb. thanks.

PR #4841, opened 2022-12-21
Instruct readers to install generator

raddevon:
`npx` running a remote package will continue to run the cached version, presumably until the cached version is deleted or until a version is specified. We want the user to manage the generator version alongside their other dependencies instead.

See Slack discussion: https://edgedb.slack.com/archives/CNECHA9EW/p1671488065449729

### DIFF
--- a/docs/guides/tutorials/nextjs.rst
+++ b/docs/guides/tutorials/nextjs.rst
@@ -365,7 +365,13 @@ typed EdgeQL queries easy and painless. The result type of our queries will be
 automatically inferred, so we won't need to manually type something like
 ``type Post = { id: string; ... }``.
 
-Generate the query builder with the following command.
+First, install the generator to your project.
+
+.. code-block:: bash
+
+  $ yarn add @edgedb/generate
+
+Then generate the query builder with the following command.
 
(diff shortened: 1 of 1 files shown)


raddevon:
This was the only place I found that needed to be changed. The Heroku guide currently has the reader run the generator without installing, but I'll make that change in the todo project it has the user pull down. No change necessary here. (That todo PR is forthcoming as soon as I get write permissions on the repo.)

The TS/JS client docs already recommend installing, so no change needed there either.

docs/guides/tutorials/nextjs.rst
```diff
@@ -365,7 +365,13 @@ typed EdgeQL queries easy and painless. The result type of our queries will be
 automatically inferred, so we won't need to manually type something like
 ``type Post = { id: string; ... }``.
 
-Generate the query builder with the following command.
+First, install the generator to your project.
+
+.. code-block:: bash
+
+  $ yarn add @edgedb/generate
```
jaclarke, line 372:
```suggestion
  $ yarn add -D @edgedb/generate
```

jaclarke (approved):

Issue #2467, opened 2021-04-19
Disallow volatile operations in schema computables


msullivan:
For example
```
type Foo {
    property rand := random();
}
```
or more excitingly `link best_friends := (SELECT .friends FILTER random() > .5)`
ought to both be disallowed.

RobbieMcKinstry:
Could you say a little more about why this is disallowed? I have a rather inoculous use case.

I'm trying to create a "Magic Link" for user logins. The link should expire after 24 hours, and I figured I could use a computed property to set that, since `datetime_current()` is allowable elsewhere in the schema definition (`default := datetime_current()`.

```
module default {
    abstract type MagicLink extending Timestamped {        
        required link user -> User;
        required property verification_code -> str;

        required property created_at -> datetime {
            readonly := true;
            default := datetime_current();
        }

        property expired := datetime_current() - .created_at < <duration>'24 hours';
    }
```

However, I get the following error:

```
$ edgedb migration create
[2022-12-05T19:59:06Z WARN  edgedb::version_check] Newer version of edgedb tool exists 2.2.6+7eabbf9 (current 2.0.1)
error: volatile functions are not permitted in schema-defined computed expressions
   ┌─ dbschema/default.esdl:21:29
   │
21 │         property expired := datetime_current() - .created_at < <duration>'24 hours';
   │                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ error

edgedb error: cannot proceed until .esdl files are fixed
```

franckdsf:
Similar use case as @RobbieMcKinstry , I don't get why this is prohibited.

fmoor:
Volatile means that each time a function is called during a transaction it could produce a different result even though it's inputs are the same. `datetime_current()` returns the current time which is different each time the function is called. This makes the function volatile. `datetime_of_transaction()` on the other hand always returns the same time during a given transaction so it is not volatile.

This is a valid schema as of EdgeDB 2.8:
```
module default {
  type MyThing {
    required property created_at -> datetime {
      default := datetime_of_transaction();
    };
    required property expired := (datetime_of_transaction() - .created_at) > <duration>'24h';
  };
}
```

RobbieMcKinstry:
That's sufficient for my use case -- I don't need the time down to the milisecond, just a close estimate of when it was created. I'll use `datetime_of_transaction()` instead. Thanks for the explanation! I appreciate it :)

PR #4842, opened 2022-12-21
Note volatile ops prohibited in schema computables

raddevon:
This is a common area of user confusion, particularly when trying to created timestamp properties. Hoping to better expose https://github.com/edgedb/edgedb/issues/2467 to users.

### DIFF
--- a/docs/datamodel/computeds.rst
+++ b/docs/datamodel/computeds.rst
@@ -33,6 +33,14 @@ field is referenced in a query.
   For more information, see the :ref:`EdgeQL > Select > Computeds
   <ref_eql_select_computeds>`.
 
+.. warning::
+
+  Volatile functions are not allowed in computed properties defined in schema.
+  This means that, for example, your schema-defined computed property cannot
+  call :eql:func:`datetime_current`, but it *can* call
+  :eql:func:`datetime_of_transaction` or :eql:func:`datetime_of_statement`.
+  This does *not* apply to computed properties outside of schema.
+
 .. _ref_dot_notation:
 
 Leading dot notation
@@ -121,6 +129,28 @@ The ``User.blog_posts`` expression above uses the *backlink operator* ``.<`` in
 conjunction with a *type filter* ``[is BlogPost]`` to fetch all the
 ``BlogPosts`` associated with a given ``User``. For details on this syntax, see
 the EdgeQL docs for :ref:`Backlinks <ref_eql_paths_backlinks>`.
+
+Created Timestamp
+^^^^^^^^^^^^^^^^^
+
+Using a computed property, you can timestamp when an object was created in your
+database.
+
+.. code-block:: sdl
+
+  type BlogPost {
+    property title -> str;
+    link author -> User;
+    required property created_at -> datetime {
+      readonly := true;
+      default := datetime_of_statement();
+    }
+  }
+
+When a ``BlogPost`` is created, :eql:func:`datetime_of_statement` will be
+called to supply it with a timestamp as the ``created_at`` property. You might
+also consider :eql:func:`datetime_of_transaction` if that's better suited to
+your use case.
 
 
 .. list-table::
--- a/docs/stdlib/datetime.rst
+++ b/docs/stdlib/datetime.rst
@@ -668,8 +668,8 @@ EdgeDB stores and outputs timezone-aware values in UTC.
 
     This function is non-volatile since it returns the current time when the
     transaction is started, not when the function is called. As a result, it
-    can be used in :ref:`computed properties <ref_datamodel_computed>`, even
-    when they are defined in schema.
+    can be used in :ref:`computed properties <ref_datamodel_computed>` defined
+    in schema.
 
 ----------
 
@@ -682,8 +682,8 @@ EdgeDB stores and outputs timezone-aware values in UTC.
 
     This function is non-volatile since it returns the current time when the
     statement is started, not when the function is called. As a result, it
-    can be used in :ref:`computed properties <ref_datamodel_computed>`, even
-    when they are defined in schema.
+    can be used in :ref:`computed properties <ref_datamodel_computed>` defined
+    in schema.
 
 ----------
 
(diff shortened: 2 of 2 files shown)


docs/stdlib/datetime.rst
```diff
@@ -652,6 +652,10 @@ EdgeDB stores and outputs timezone-aware values in UTC.
         db> select datetime_current();
         {<datetime>'2018-05-14T20:07:11.755827Z'}
 
+    This function is volatile since it always returns the current time when it
+    is called. As a result, it cannot be used in :ref:`computed properties
```
raddevon, line 656: @jaclarke I notice that my links to `ref_datamodel_computed` go to the correct page, but they scroll to the bottom of the page when I'm testing locally. Is that something I'm doing wrong, or is it a bug? The target is at the very top of that "Computed properties" page.

docs/datamodel/computeds.rst
```diff
@@ -33,6 +33,15 @@ field is referenced in a query.
   For more information, see the :ref:`EdgeQL > Select > Computeds
   <ref_eql_select_computeds>`.
 
+.. warning::
+
+  Volatile functions are not allowed in computed properties defined in schema.
+  This means that, for example, your schema-defined computed property cannot
+  call :eql:func:`datetime_current`, but it *can* call
+  :eql:func:`datetime_of_transaction` or :eql:func:`datetime_of_statement`.
+  This does *not* apply to computed properties in :ref:`free objects
```
fmoor, line 42: Is this exception specific to free objects or queries outside of the schema generally?

fmoor (commented):
Do we have an explicit definition of what volatile means somewhere in the docs? I'm not sure it is necessary for this change but it could be helpful.

raddevon:
> Do we have an explicit definition of what volatile means somewhere in the docs? I'm not sure it is necessary for this change but it could be helpful.

@fmoor We have this [in the SDL function declaration reference](https://www.edgedb.com/docs/reference/sdl/functions#description):

```rest
    Function volatility determines how aggressively the compiler can
    optimize its invocations.

    If not explicitly specified the function volatility is set to
    ``Volatile`` by default.

    * A ``Volatile`` function can modify the database and can return
      different results on successive calls with the same arguments.

    * A ``Stable`` function cannot modify the database and is
      guaranteed to return the same results given the same
      arguments *within a single statement*.

    * An ``Immutable`` function cannot modify the database and is
      guaranteed to return the same results given the same arguments
      *forever*.
```

Users aren't likely to find it when they're using functions but only when they're declaring them. This is the part that seems relevant for users who might want to consume a function in a computed property:

> A ``Volatile`` function can modify the database and can return different results on successive calls with the same arguments.

I could copy that to datamodel/computeds. I wonder if it's even entirely accurate though. Should this be "... can modify the database *or* can return different results..." rather than "and?" `datetime_current()` doesn't modify the database but is volatile because it can return different results given the same arguments, unless I'm missing something.

docs/datamodel/computeds.rst
raddevon: Are there other contexts where computed properties can be used besides in schema or in a free object? Maybe I need to change it to say, "This does *not* apply to uses of computed properties outside schema."

fmoor: computeds can be used in any shape, for example
```
select User { id, name, last_login := (select .logins order by .timestamp desc limit 1) }
```
The type here is `User` but the shape includes a computed link.

fmoor:
> @fmoor We have this [in the SDL function declaration reference](https://www.edgedb.com/docs/reference/sdl/functions#description)

That is good enough for me. As long as there is something somewhere that people can find/link to. As far as the accuracy of the definition, I have no idea.

docs/datamodel/computeds.rst
fmoor: > This does not apply to uses of computed properties outside the schema.

This seems correct to me.

raddevon: > computeds can be used in any shape, for example
> 
> ```
> select User { id, name, last_login := (select .logins order by .timestamp desc limit 1) }
> ```
> 
> The type here is `User` but the shape includes a computed link.

That makes sense. Thanks for clarifying! I've updated the language to catch all usages outside of schema.

docs/stdlib/datetime.rst
```diff
@@ -662,6 +666,10 @@ EdgeDB stores and outputs timezone-aware values in UTC.
 
     Return the date and time of the start of the current transaction.
 
+    This function is non-volatile since it returns the current time when the
+    transaction is started, not when the function is called. As a result, it
+    can be used in :ref:`computed properties <ref_datamodel_computed>`, even
+    when they are defined in schema.
```
fmoor, line 672: A bit of a nit. Maybe drop `, even when they are` since it implies this has something to do with computeds outside of schema?

docs/stdlib/datetime.rst
```diff
@@ -672,6 +680,10 @@ EdgeDB stores and outputs timezone-aware values in UTC.
 
     Return the date and time of the start of the current statement.
 
+    This function is non-volatile since it returns the current time when the
+    statement is started, not when the function is called. As a result, it
+    can be used in :ref:`computed properties <ref_datamodel_computed>`, even
+    when they are defined in schema.
```
fmoor, line 686: same here

fmoor (approved):
Looks good to me. Maybe a review from someone with a deeper understanding of the language is good though. I mostly don't know what I'm talking about :upside_down_face:.

raddevon:
Would love to get your eyes on this one too when you have time, @vpetrovykh, to confirm my understanding and accuracy before I merge.

docs/stdlib/datetime.rst
raddevon: My fear for leaving out any mention of definition in schema is that it might imply there are cases when a volatile function cannot be used in a computed property that is *not* inside schema.

> As a result, it can be used in computed properties.

Maybe if I reword like this?

> As a result, it can be used in schema-defined computed properties.

fmoor: I was thinking it would read 
> As a result, it can be used in computed properties defined in schema.

msullivan (approved):
Yeah this all seems correct.

msullivan:
Let's merge it

PR #4843, opened 2022-12-21
fix: alter abstract pointer set default

### LINKED ISSUE #4837: Improve error message on abstract property with default value
A user [encountered this on Discord](https://discord.com/channels/841451783728529451/849377751370432573/1054704821560684564) and assumed abstract properties were bugged. Adding a default value to an abstract property appears to be prohibited, but the error message does not indicate the problem:

```
edgedb error: InternalServerError: no referrer context for <class 'edb.schema.properties.AlterProperty'>
```

Can a more specific error be returned in this case to reflect that abstract properties do not support default values (or maybe that they only support `readonly` and annotations)?

aljazerzen:
Closes #4837

The problem was in that when creating a property with an enum default, CreateProperty command is split into two: CreateProperty & AlterProperty set default. Because the check for default on abstract pointers was implemented on CreateProperty, it was not throwing when default was added via AlterProperty.

This commit moves the check to PointerCommand - unifies Property & Link and Create & Alter.

### DIFF
--- a/edb/schema/pointers.py
+++ b/edb/schema/pointers.py
@@ -1404,14 +1404,16 @@ class PointerCommandOrFragment(
         value: s_expr.Expression,
         track_schema_ref_exprs: bool=False,
     ) -> s_expr.CompiledExpression:
+
         if field.name in {'default', 'expr'}:
-            parent_ctx = self.get_referrer_context_or_die(context)
-            source = parent_ctx.op.get_object(schema, context)
-            parent_vname = source.get_verbosename(schema)
-            ptr_name = self.get_verbosename(parent=parent_vname)
-            in_ddl_context_name = None
             if field.name == 'expr':
+                parent_ctx = self.get_referrer_context_or_die(context)
+                source = parent_ctx.op.get_object(schema, context)
+                parent_vname = source.get_verbosename(schema)
+                ptr_name = self.get_verbosename(parent=parent_vname)
                 in_ddl_context_name = f'computed {ptr_name}'
+            else:
+                in_ddl_context_name = None
 
             return self._compile_expr(
                 schema,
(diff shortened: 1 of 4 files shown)


edb/schema/pointers.py
```diff
@@ -1404,14 +1404,16 @@ def compile_expr_field(
...
-            ptr_name = self.get_verbosename(parent=parent_vname)
-            in_ddl_context_name = None
             if field.name == 'expr':
+                parent_ctx = self.get_referrer_context_or_die(context)
+                source = parent_ctx.op.get_object(schema, context)
+                parent_vname = source.get_verbosename(schema)
+                ptr_name = self.get_verbosename(parent=parent_vname)
                 in_ddl_context_name = f'computed {ptr_name}'
+            else:
+                in_ddl_context_name = None
```
aljazerzen, line 1416: This is just a refactor

edb/schema/pointers.py
```diff
@@ -1717,10 +1719,19 @@ def _cmd_tree_from_ast(
...
     ) -> sd.Command:
         cmd = super()._cmd_tree_from_ast(schema, astnode, context)
+        assert isinstance(cmd, PointerCommand)
+
         referrer_ctx = cls.get_referrer_context(context)
         if referrer_ctx is not None:
             if getattr(astnode, 'declared_overloaded', False):
                 cmd.set_attribute_value('declared_overloaded', True)
+        else:
+            # This is an abstract propery/link
```
msullivan, line 1729: s/propery/property/

msullivan, line 1729:
```suggestion
            # This is an abstract property/link
```

msullivan: hey I finally figured out how to make a suggestion!

msullivan (approved):
Looks good!

edb/schema/pointers.py
aljazerzen: My spell checker only shows first 100 mistakes in a file :confused:

PR #4844, opened 2022-12-21
Fix SQL expr & type array bounds parsing

aljazerzen:


edb/common/ast/codegen.py
```diff
@@ -52,6 +52,15 @@ def node_visit(self, node: base.AST) -> None:
...
 
+    def visit_indented(self, node: base.AST, indent: bool=False) -> None:
+        if indent:
+            self.new_lines = 1
+            self.char_indentation += 1
+        res = self.visit(node)
+        if indent:
+            self.char_indentation -= 1
+        return res
+
```
aljazerzen, line 63: I've added this helper that could probably simplify pg codegen a lot. If accepted here, I may do another refactor on it.

msullivan: Yeah that looks good. Should `indent` default to `True` here, though?

edb/common/ast/codegen.py
```diff
@@ -52,6 +52,22 @@ def node_visit(self, node: base.AST) -> None:
         visitor = getattr(self, method, self.generic_visit)
         return visitor(node)
 
+    def visit_indented(
+        self, node: base.AST, indent: bool = False, nest: bool = False
```
msullivan, line 56: Should indent be true by default?

edb/pgsql/ast.py
```diff
@@ -884,22 +878,21 @@ def copyfrom(self, other):
...
 class SubLinkType(enum.IntEnum):
-    EXISTS = enum.auto()
-    NOT_EXISTS = enum.auto()
-    ALL = enum.auto()
-    ANY = enum.auto()
-    EXPR = enum.auto()
+    pass
+    pass
+    pass
+    pass
```
msullivan, line 884: Maybe just one `pass`.

msullivan: Wait, no, maybe it can just go away :)

msullivan (approved):
Great cleanup

edb/common/ast/codegen.py
aljazerzen: Ah, yes. I wanted to just override `visit`, but typing complained extensively.

edb/pgsql/ast.py
aljazerzen: Maybe I should start re-reading my commits before pushing :D

PR #4845, opened 2022-12-21
Make nonrecursive_access_policies the only thing

msullivan:
We'll also need to update the documentation to reflect this, but
updating it now would change the live site too early, right?


1st1:
> but updating it now would change the live site too early, right?

Yes, let's not do it just now. You can create a separate issue so that we don't forget.

elprans (approved):

Issue #4814, opened 2022-12-13
Support for common SQL data analysis tools


aljazerzen:
Ref #4729

Tested tools:

- [x] dbt - [works](https://github.com/edgedb/edgedb/issues/4814#issuecomment-1403769986), but only for ephemeral models
- [x] Metabase - [tested](https://github.com/edgedb/edgedb/issues/4814#issuecomment-1403769986)
- [x] Cluvio - [tested](https://github.com/edgedb/edgedb/issues/4814#issuecomment-1403769986)
- [x] ~~Redash (couldn't deploy)~~
- [x] Tablau (works)
- [x] Fivetran ([works](https://github.com/edgedb/edgedb/issues/4814#issuecomment-1503152710))
- [x] Airbyte (works)
- [x] Hevo ([works](https://github.com/edgedb/edgedb/issues/4814#issuecomment-1409253738))
- [x] Stitch ([works](https://github.com/edgedb/edgedb/issues/4814#issuecomment-1503152710))
- [x] DataGrip (works)
- [x] pg_dump (is throwing "oid not found" errors, we have to hide more things)

----

Original post:

## Metabase

Result: "Suspended cursor is not supported"
Error at: https://github.com/edgedb/edgedb/blob/60380fd7ae75b0230234315ad897405061ac08f3/edb/server/protocol/pg_ext.pyx#L626-L628

I had to add support for parsing SET statement and resolving it to an empty relation. They are using:
```
SET extra_float_digits TO 3
```
I think we are probably save for ignore this specific SET statement.

Reproduce with local instance of edgedb and:
```
$ docker run -d -p 3000:3000 --name metabase --net=host metabase/metabase
```

## Cluvio

Error: ERROR: suspended cursor is not supported

To reproduce: I used their cloud to connect to my VPS with edgedb development env and a public IP.

## Dbt-cloud

Result:
```
Runtime Error

  dbt was unable to connect to the specified database.

  The database returned the following error:

  

    >Database Error

    unsupported SQL construct: unknown enum: {'TransactionStmt': {'kind': 'TRANS_STMT_BEGIN'}}

  

  Check your database credentials and try again. For more information, visit:

  https://docs.getdbt.com/docs/configure-your-profile
```
To reproduce: I used their cloud to connect to my VPS with edgedb development env and a public IP.

This means that we need transactions.

## Redash

I gave up trying to deploy an instance and connect to plain Postgres.

## Tablau

Result:
```
Can’t connect to PostgreSQL
[...]

PR #4849, opened 2022-12-22
SQL search path

aljazerzen:
Implements [search_path](https://www.postgresql.org/docs/15/ddl-schemas.html) as discussed in #4814


msullivan (approved):
Is there a way to test it?

aljazerzen:
I could add tests that call directly pg parser & resolver, but it'll be way easier when `SET search_path` is implemented.

So for now, I've tested locally and added xskip-ed tests that can be enabled when supported.

Issue #4854, opened 2022-12-27
Liveness & Readiness endpoints


XAMPPRocky:
In kubernetes, you typically expose a set of probes to allow the system to know when it's live, when it's ready to serve traffic, and when it has started up. As far as I can tell Edgedb currently doesn't have any documented way to check for it's liveness and readiness. It would be nice if there was documented way to check for Edgedb's liveness and readiness.

https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

elprans:
Documentation is, indeed, lacking for these.

The (HTTP) endpoints are: `/server/status/ready` and `/server/status/alive`

PR #4856, opened 2022-12-28
Document aliveness/readiness endpoints

raddevon:
to address #4854

A couple of questions to help me get this ready:

1. What do the endpoints return if the states are _not_ OK? (Is there any way to create or simulate these scenarios so that I can find out?)
2. Where else should this be documented? I believe this needs to be documented where I have it in the HTTP client documentation since it's tied to that, but I feel it's unlikely someone looking for this functionality will know to look there. I don't see an obvious place within the current docs' structure that this should go. I was thinking of adding a cheatsheet (at Guides > Cheatsheets) and then maybe adding a new top-level document under "Reference". Do those locations make sense? Anywhere else I'm missing?

### DIFF
--- a/docs/guides/deployment/digitalocean.rst
+++ b/docs/guides/deployment/digitalocean.rst
@@ -221,3 +221,10 @@ instance.
 
    The command groups ``edgedb instance`` and ``edgedb project`` are not
    intended to manage production instances.
+
+Health Checks
+=============
+
+Using the EdgeDB HTTP client, you can perform health checks to monitor the
+status of your EdgeDB instance. Learn how to use them with our :ref:`health
+checks guide <ref_guide_deployment_health_checks>`.
--- a/docs/guides/deployment/digitalocean.rst
+++ b/docs/guides/deployment/digitalocean.rst
@@ -225,6 +225,6 @@ instance.
 Health Checks
 =============
 
-Using the EdgeDB HTTP client, you can perform health checks to monitor the
-status of your EdgeDB instance. Learn how to use them with our :ref:`health
-checks guide <ref_guide_deployment_health_checks>`.
+Using an HTTP client, you can perform health checks to monitor the status of
+your EdgeDB instance. Learn how to use them with our :ref:`health checks guide
+<ref_guide_deployment_health_checks>`.
--- a/docs/guides/deployment/health_checks.rst
+++ b/docs/guides/deployment/health_checks.rst
@@ -6,38 +6,8 @@ Health Checks
 
 You may want to monitor the status of your EdgeDB instance. Is it up? Is it
 ready to take queries? This guide will show you to perform health checks using
-our HTTP client and the ``alive`` and ``ready`` endpoints.
+HTTP and the ``alive`` and ``ready`` endpoints.
 
-Enable the HTTP Client
-======================
-
-Add this line to your schema, outside the ``module`` block:
-
-.. code-block:: sdl
-
-    using extension edgeql_http;
-
-Then use the CLI to create a migration:
-
-.. code-block:: bash
-
-    $ edgedb migration create
-    did you create extension 'edgeql_http'? [y,n,l,c,b,s,q,?]
-    > y
-    Created dbschema/migrations/00003.edgeql, id: <your-migration-id>
-
-Your migration's filename and ID may be different.
-
-Apply the new migration:
-
-.. code-block:: bash
-
-    $ edgedb migrate
-    Applied <your-migration-id> (00003.edgeql)
-
-EdgeDB will now expose HTTP endpoints you can use to interact with the
-database.
-    
 
 Check Instance Aliveness
 ========================
--- a/docs/guides/deployment/health_checks.rst
+++ b/docs/guides/deployment/health_checks.rst
@@ -1,0 +1,73 @@
+.. _ref_guide_deployment_health_checks:
+
+=============
+Health Checks
+=============
+
+You may want to monitor the status of your EdgeDB instance. Is it up? Is it
+ready to take queries? This guide will show you to perform health checks using
+our HTTP client and the ``alive`` and ``ready`` endpoints.
+
+Enable the HTTP Client
+======================
+
+Add this line to your schema, outside the ``module`` block:
+
+.. code-block:: sdl
+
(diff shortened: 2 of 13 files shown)


XAMPPRocky:
FWIW just my opinion as a user on 2, I'd love to see a Kubernetes sub chapter in the deployment chapter that listed these endpoints.

elprans:
> 1. What do the endpoints return if the states are _not_ OK? (Is there any way to create or simulate these scenarios so that I can find out?)

Server will return a 50x response if not ready (or a network connection error).

> 2. Where else should this be documented?

I would add a common entry under Guides / Deployment and link to it from each guide.

raddevon:
> FWIW just my opinion as a user on 2, I'd love to see a Kubernetes sub chapter in the deployment chapter that listed these endpoints.

Thank you for the feedback, @XAMPPRocky! I think it might make sense for us to consider writing a deployment guide for Kubernetes, but I'm not sure about adding a link for Kubernetes that would just include this aliveness/readiness info. If and when we do write a guide on Kubernetes, sounds like it would be useful to also document this there or at least link to our existing documentation on it.

Do you feel like you would have found it at @elprans' suggested location: as a new link under deployment guides? Here's a mock-up of what that might be:
CleanShot 2022-12-29 at 07 44 41@2x


Would love to hear any thoughts, too, on the best title for this, if it does end up there. "Instance Status?" "Aliveness/Readiness?" Maybe something slightly more verbose like "Checking Instance Status?"

raddevon:
Re: naming, I ended up going with "Health Checks" at @elprans suggestion.

Marking this ready for review. I will prep a separate PR to add an "HTTP API" section to "Reference."

docs/guides/deployment/health_checks.rst
```diff
@@ -0,0 +1,73 @@
...
+
+=============
+Health Checks
+=============
+
+You may want to monitor the status of your EdgeDB instance. Is it up? Is it
+ready to take queries? This guide will show you to perform health checks using
+our HTTP client and the ``alive`` and ``ready`` endpoints.
+
+Enable the HTTP Client
```
elprans, line 11: `edgeql_http` is _not_ required, the liveness/readiness checks are always available.

docs/guides/deployment/digitalocean.rst
```diff
@@ -216,3 +216,10 @@ Set the security policy to strict.
 That's it! Refer to the :ref:`Construct the DSN
 <ref_guide_deployment_digitalocean_link>` section above to connect to your
 instance.
+
+Health Checks
+=============
+
+Using the EdgeDB HTTP client, you can perform health checks to monitor the
```
elprans, line 223:
```suggestion
Using an HTTP client, you can perform health checks to monitor the
```

raddevon:
Just noticed I missed a reference to needing to activate HTTP for a health check, but I have that removed now. I think this should be good to go.

raddevon:
Added you as reviewer here, @zackelan, since you took a look at my other similar PR. This is aliveness/readiness documentation but as part of the HTTP docs under "Client Libraries" and within the "Deployment" documentation.

zackelan:
> What do the endpoints return if the states are not OK? (Is there any way to create or simulate these scenarios so that I can find out?)

if you want to learn to fish:

```
cobalt ~/src/edgedb > rg -l alive
tests/test_http_graphql_query.py
tests/test_server_ops.py
edb/testbase/server.py
edb/server/pgcluster.py
edb/server/pgcon/pgcon.pyx
edb/server/protocol/system_api.py
edb/server/protocol/protocol.pxd
edb/server/server.py
edb/server/protocol/protocol.pyx
edb/server/protocol/binary.pyx
edb/common/markup/serializer/base.py
```

from that list, `system_api.py` has the actual endpoint implementations. it looks like the 'alive' endpoint does a `SELECT 'OK'` query against the `__edgedbsys__` Postgres database.

so you should never do this with a real instance, it voids the warranty, but...

```
cobalt ~ > docker run --rm -d -e EDGEDB_SERVER_SECURITY=insecure_dev_mode \
    -v /tmp/healthcheck-test:/run/edgedb --name healthcheck-test -p 5656:5656 \
    edgedb/edgedb:2.9
cobalt ~ > xh http://localhost:5656/server/status/alive
HTTP/1.1 200 OK
Content-Length: 4
Content-Type: application/json

"OK"
cobalt ~ > sudo psql -h /tmp/healthcheck-test -U postgres
psql (14.6, server 14.5)
Type "help" for help.

postgres=# \l
                                           List of databases
           Name            |       Owner        | Encoding | Collate |  Ctype  |   Access privileges
---------------------------+--------------------+----------+---------+---------+-----------------------
 V2f147ded60___edgedbsys__ | V2f147ded60_edgedb | UTF8     | C       | C.UTF-8 |
 V2f147ded60___edgedbtpl__ | V2f147ded60_edgedb | UTF8     | C       | C.UTF-8 |
 V2f147ded60_edgedb        | V2f147ded60_edgedb | UTF8     | C       | C.UTF-8 |
 postgres                  | postgres           | UTF8     | C       | C.UTF-8 |
 template0                 | postgres           | UTF8     | C       | C.UTF-8 | =c/postgres          +
                           |                    |          |         |         | postgres=CTc/postgres
 template1                 | postgres           | UTF8     | C       | C.UTF-8 | =c/postgres          +
                           |                    |          |         |         | postgres=CTc/postgres
(6 rows)

postgres=# UPDATE pg_database SET datallowconn = false WHERE datname = 'V2f147ded60___edgedbsys__';
UPDATE 1
postgres=# SELECT pg_terminate_backend(pg_stat_activity.pid) FROM pg_stat_activity
postgres-# WHERE pg_stat_activity.datname = 'V2f147ded60___edgedbsys__'
postgres-# AND pid <> pg_backend_pid();
 pg_terminate_backend
----------------------
 t
 t
(2 rows)
cobalt ~ > xh http://localhost:5656/server/status/alive
HTTP/1.1 503 Service Unavailable
Connection: close
Content-Length: 158
Content-Type: text/plain

{
    "error": {
        "message": "Postgres is not available: terminating connection due to administrator command",
        "type": "BackendUnavailableError",
        "code": 134217729
    }
}
```

zackelan (approved):

Issue #4717, opened 2022-11-25
"if orphan" documentation lacking and somewhat misleading


brassel:
- EdgeDB Version:
"2.7+a86e274"
- EdgeDB CLI Version:
EdgeDB CLI 2.2.6+7eabbf9
- OS Version:
macOS 12.6 (21G115)

Steps to Reproduce:

Run the following edgedb script in an empty database.

~~~
create module default if not exists;
create type default::Sub {
    create property name -> std::str;
};
create type default::Parent {
    create required link sub -> default::Sub {
        on source delete delete target if orphan;
    };
};
create type default::Parent2 {
    create link sub -> default::Sub {
        on source delete delete target if orphan;
    };
};
insert Sub {
  name := "sub 1"
};
insert Parent {
  sub := (select Sub filter .name = "sub 1" limit 1)
};
insert Parent2 {
  sub := (select Sub filter .name = "sub 1" limit 1)
};
delete Parent;
~~~

The result is 
~~~~
edgedb error: ConstraintViolationError: deletion of default::Sub (4e1116d8-6c9c-11ed-baf8-b3b891af78e4) is prohibited by link target policy
  Detail: Object is still referenced in link sub of default::Parent2 (4e1720d2-6c9c-11ed-baf8-b3f158647624).
~~~~

Schema:

msullivan:
I believe this is working as intended, but our documentation fails to document the subtleties here:

`IF ORPHAN` means that we try to delete the target if the target object is orphaned *with respect to that link*. So because the `Sub` object is not pointed to by any other `Parent` object through the `sub` link, when `Parent` is deleted we try to delete the `Sub`, which then triggers `Parent2`'s constraint.

PR #4858, opened 2022-12-28
Clarify `if orphan`

raddevon:
to address #4717

Please check the accuracy of this when you get a chance, @msullivan. I'm particularly concerned if I properly understand the implications I've described in the second paragraph of my note.

If my description _is_ accurate, I wonder if we can and should consider another way to handle beyond just documenting it as-is. I feel like the result I've described would be fairly surprising.

### DIFF
--- a/docs/datamodel/links.rst
+++ b/docs/datamodel/links.rst
@@ -368,8 +368,9 @@ delete`` clause.
 Under this policy, deleting a ``MessageThread`` will *unconditionally* delete
 its ``messages`` as well.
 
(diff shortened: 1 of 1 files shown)


msullivan (approved):
Yeah, that is all correct

msullivan:
We definitely *could* implement the other behavior.

I believe that Elvis was concerned about the performance implications, though it might not be that bad: we have to do the exact same checks when we actually *do* delete an object as we do to check if it is an orphan...

PR #4862, opened 2022-12-29
Fix zlibg1-dev name

qwolphin:
Looks like there is a mistake in the library name

```shell
wolphin@meow:~$ docker run -it ubuntu:22.10 bash
root@bbeedfbb7589:/# apt update
<...>

root@bbeedfbb7589:/# apt install zlibg1-dev
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
E: Unable to locate package zlibg1-dev

root@bbeedfbb7589:/# apt install zlib1g-dev
<...>
Need to get 14.5 MB of archives.
After this operation, 50.0 MB of additional disk space will be used.
Do you want to continue? [Y/n] 
```

### DIFF
--- a/docs/guides/contributing.rst
+++ b/docs/guides/contributing.rst
@@ -36,7 +36,7 @@ On Ubuntu 22.10, these can be installed by running:
 .. code-block:: bash
 
    $ apt install make gcc rust-all autotools-dev python3.11-dev \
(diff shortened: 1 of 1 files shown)


elprans (approved):
Thanks!

PR #4867, opened 2023-01-04
pg-ext: track and restore session settings

fantix:
- [x] Necessary changes to the SQL parser/codegen
- [x] Drop emulated portals / `Bind`
- [x] Encapsulate query behaviors with `SQLQueryUnit`
- [x] Support transactions
- [x] Track settings in simple query
- [x] Track settings in extended query
- [x] Handle `search_path` in simple query
- [x] Handle `search_path` in extended query
- [ ] Add tests
- [ ] `PREPARE` and `EXECUTE`?
- [ ] Two-phase transactions?
- [ ] md5 authentication

### Abstract

As we don't pin backend connections to frontend connections, pg-ext session settings must be stored in the frontend connections, and restored to the backend connection before using it.

### Solution

The way to achieve settings tracking is to track every successful message executed in the backend connection. As PostgreSQL doesn't return enough information about what is just executed, we need to attach additional information from the compiler to all the messages we send to the backend, that is the `SQLQueryUnit` struct. It contains necessary information to replay what just happened to Postgres in memory, particularly in the `ConnectionView` object, a.k.a. the "dbview". Before using a newly-borrowed backend connection, we just run a bulk `set_config()` prepared query with the settings from the `dbview` to restore the state. This is pretty much the same as how EdgeDB frontend connections handle their version of dbviews.

In PostgreSQL, session settings are transactional. That means, our replay logic in the dbview must be aware of the transaction state as well. This includes a full emulation of PostgreSQL transaction handling, including implicit/explicit transactions, savepoints and error states. At this point, we don't have to track local settings, because they are gone with the end of transactions, while a backend connection is always pinned during transactions - even implicit ones.

An exception is the "frontend-only" settings. For now, there is only the `search_path` config. They are totally emulated in the EdgeDB server, not affecting the backend connection at all. Unfortunately, this is transparent to the users who might just mix frontend-only settings with any other messages. Therefore, we'll have to defer its handling to when the backend logic hits a frontend-only query unit. For transaction handling, we're also tracking the frontend-only local settings separately, as we're not depending on the backend to track them here.

The transaction tracking is not only happening in the I/O server / dbview. In #4729, we did all compilation before batch execution to reduce round-trip time for performance. However, the compilation actually needs the frontend-only settings (`search_path`), while the queries may still modify them too. The solution here is to emulate the transaction state in the compilation step again in advance, so that compiler could assume all commands will succeed and use the right `search_path`. For simple query, this is done in `dbstate.SQLTransactionState` which is derived from the dbview and running the emulation in the compiler. For extended query, we will simply use a copy of the same dbview, because the compilation is always done on a query-by-query basis.


edb/pgsql/parser/ast_builder.py
```diff
@@ -807,8 +813,11 @@ def _build_a_expr(n: Node, c: Context) -> pgast.BaseExpr:
 
 def _build_func_call(n: Node, c: Context) -> pgast.FuncCall:
     n = _unwrap(n, "FuncCall")
+    name = tuple(_list(n, c, "funcname", _build_str))  # type: tuple[str, ...]
+    if name == ('set_config',) or name == ('pg_catalog', 'set_config'):
+        raise PSqlUnsupportedError(n)
     return pgast.FuncCall(
-        name=tuple(_list(n, c, "funcname", _build_str)),
+        name=name,
```
aljazerzen, line 820: This is not really the best place to throw this error.

Consider moving this to here: https://github.com/edgedb/edgedb/blob/master/edb/pgsql/resolver/expr.py#L263-L265

aljazerzen (approved):

aljazerzen:
Nice write-up.

edb/pgsql/ast.py
```diff
@@ -1052,30 +1061,24 @@ class RollbackToStmt(TransactionStmt):
...
-    REPEATABLE_READ = enum.auto()
-    READ_COMMITTED = enum.auto()
-    READ_UNCOMMITTED = enum.auto()
+class RollbackPreparedStmt(TwoPhaseTransactionStmt):
+    pass
 
 
-class AccessMode(enum.IntEnum):
-    READ_WRITE = enum.auto()
-    READ_ONLY = enum.auto()
```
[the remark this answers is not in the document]
fantix: Oh right, sorry about this! Looks like we just need the values to be strings - just like the other settings. Also as you might have already noticed, PostgreSQL is weird about transaction settings, `SET SESSION CHARASTICS ...` is actually setting those `default_transaction_*` settings, and `SET transaction_*` is actually only effective in (transaction-)local scope even without the `SET LOCAL` keyword 🤯 .

edb/pgsql/ast.py
```diff
@@ -1000,7 +1000,16 @@ class Statement(Base):
 
 class VariableSetStmt(Statement):
     name: str
+    args: ArgsList
+    scope: OptionsScope
+
+
+class ArgsList(Base):
     args: typing.List[BaseExpr]
```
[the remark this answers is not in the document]
fantix: This is used at:
 https://github.com/edgedb/edgedb/blob/2249ee521fb86a34758a8f5a03408b3d7cc8e403/edb/server/compiler/compiler.py#L546

If it's just a Python `list`, the generated result wouldn't be correct as expected.

edb/pgsql/ast.py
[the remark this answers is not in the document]
aljazerzen: Don't be sorry, I'm just trying to be funny.

> is actually only effective

I'm not even surprised about such inconsistencies anymore...

tests/test_pgext.py
```diff
@@ -842,8 +842,9 @@ async def test_sql_proto_extended_query_08(self):
             Execute("portal8"),
             Sync(),
         )
-        await self.conn.read(ParseComplete)
-        await self.conn.read(BindComplete)
+        # On Postgres we will also receive:
```
elprans, line 845: What does this comment mean exactly?

fantix: This is to mark a behavioral difference between vanilla Postgres and our implementation

elprans: Could it cause problems for clients?

fantix: Theoretically yes, like clients sending `Parse` wouldn't expect a `42P03: cursor already exists` error; but it's an error after all, so I guess most clients would just raise the error regardlessly?

elprans: OK, I guess we'll see, but ideally we shouldn't deviate too much from vanilla Postgres.

elprans (approved):

fantix:
I'll add more tests in a new PR

PR #4868, opened 2023-01-04
Add reference doc for the HTTP API

raddevon:
- Not sure about the order of the documents under "Reference." It seems somewhat haphazard but maybe there is a rationale here I'm not understanding. I had considered alphabetizing the items with multiple documents (EdgeQL, SDL, DDL, Binary Protocol, Administration) at the top of the list and alphabetizing all the single-document items below those. I don't know if it makes sense, but some kind of order is probably better than none at all. Like I said, though, maybe what we have currently is an order that I just don't understand.
- I created this as a single document. It didn't seem like enough content to justify breaking it up. I incorporated the existing "Observability" document into the new "HTTP API" document and removed the old document. Thought about splitting the HTTP API into multiple documents just to maintain a separate "Observability" doc, but I don't think it actually improves discoverability since those trees are collapsed by default when a user opens "Reference." They still need to know "Observability" is in "HTTP API" to see it, so I decided to just make it a single document. Maybe there's an argument for keeping the top-level "Observability" doc though and just replicating the content or even linking across. A single document with all the HTTP reference seemed slightly better, but I'm not married to it. The "Observability" ref is intact, so any links to it will still work as before.


docs/reference/http.rst
```diff
@@ -0,0 +1,218 @@
...
+
+Using HTTP, you may check the health of your EdgeDB instance, check metrics on
+your instance, and make queries.
+
+Your instance's URL takes the form of ``http://<hostname>:<port>/``. For
+queries, you will append ``db/<database-name>/edgeql``.
+
+.. note::
+
+    Here's how to determine your development EdgeDB instance's HTTP server URL:
```
1st1, line 14: @fantix @tailhook @elprans we should probably have a command for stuff like this. https://github.com/edgedb/edgedb-cli/issues/914

docs/reference/http.rst
```diff
@@ -0,0 +1,218 @@
...
+
+You may make queries via either the POST or GET HTTP method. Query requests can
+take the following fields:
+
+- ``query`` - contains the EdgeQL query string
+- ``variables``- contains a JSON object where the keys are the parameter names
+  from the query and the values are the arguments to be used in this execution
+  of the query.
+
+When using the GET method, supply ``query`` and ``variables`` as query
```
zackelan, line 171: maybe add an example of a GET request here? if for no other reason than to show that encoding all of this in the querystring is fiddly and for any non-trivial query they probably want to use POST with a request body

docs/reference/http.rst
```diff
@@ -0,0 +1,218 @@
...
+Note that the ``error`` field will only be present if an error
+actually occurred. The ``error`` will further contain the ``message``
+field with the error message string, the ``type`` field with the name
+of the type of error and the ``code`` field with an integer
+:ref:`error code <ref_protocol_error_codes>`.
+
+.. note::
+
+    Caution is advised when reading ``decimal`` or ``bigint`` values
+    using HTTP protocol because the results are provides in JSON
```
zackelan, line 210:
```suggestion
    using the HTTP protocol because the results are provided in JSON
```

docs/reference/http.rst
```diff
@@ -0,0 +1,218 @@
...
+
+Your instance's URL takes the form of ``http://<hostname>:<port>/``. For
+queries, you will append ``db/<database-name>/edgeql``.
+
+.. note::
+
+    Here's how to determine your development EdgeDB instance's HTTP server URL:
+
+    - The ``hostname`` will be ``localhost``
+    - Find the ``port`` by running ``edgedb instance list``. This will print a
```
zackelan, line 17: the hostname is in this output, so could pull both from there, and it would work for remote-linked instances as well.

docs/reference/http.rst
```diff
@@ -0,0 +1,218 @@
...
+
+.. code-block::
+
+    http://<hostname>:<port>/server/status/alive
+
+If your instance is alive, it will respond with a ``200`` status code and
+``"OK"`` as the payload. Otherwise, it will respond with a ``50x`` or a network
+error.
+
+Awakeness
```
zackelan, line 46:
```suggestion
Readiness
```

docs/reference/http.rst
raddevon: Do you mean that `edgedb instance list` gives you the hostname?

zackelan: yes - the "port" column is a bit misleadingly named, it's both hostname & port. here I used `edgedb instance link` on my laptop to link to an instance I have running on my desktop:

```
gold ~/src/edgedb-cli > ./target/debug/edgedb instance list
┌────────┬─────────┬──────────────┬─────────────┬────────┐
│ Kind   │ Name    │ Port         │ Version     │ Status │
├────────┼─────────┼──────────────┼─────────────┼────────┤
│ remote │ desktop │ cobalt:15656 │ 2.9+5ac35ff │ up     │
└────────┴─────────┴──────────────┴─────────────┴────────┘
gold ~/src/edgedb-cli > xh cobalt:15656/server/status/ready
HTTP/1.1 200 OK
Content-Length: 4
Content-Type: application/json

"OK"
```

raddevon: Ah, OK. So, for remote instances, you'll also get the hostname in the "Port" column.

Should I include this in a new paragraph in this note addressing production instances, or is there a case where someone would have a remote development instance? If this primarily addresses production, I could add something like this:

> To determine the URL of a production instance, you can get both the hostname and port of the instance from the "Port" column of the `edgedb instance list` table (formatted as `<hostname>:<port>`) if you have already linked to the instance. The same guidance on development database names applies here.

zackelan: AFAIK remote instances can be either production or development - I can see a remote dev instance being useful in that laptop/desktop case, and a remote production instance being useful for debugging.

raddevon: I'd love to get your thoughts, @zackelan, on this rework. Changed "development" in the original to "local" and added a paragraph for determining remote instance URL to incorporate your suggestion.

Thanks for the great feedback, by the way!

zackelan: LGTM!

zackelan (approved):

Issue #4796, opened 2022-12-08
Index on a mutable computed property is recognized as a valid SDL


crasite:
- EdgeDB Version: 2.7+a86e274
- EdgeDB CLI Version: 2.2.6+7eabbf9
- OS Version: macOS 13.0.1 (22A400) 

Explanation:  

- Schema which have an index on a mutable expression shouldn't be possible.  
- The following schema will create a valid migration file, but it won't be able to migrate.

Steps to Reproduce:

1. Create the following schema with `edgedb migration create`
2. This create a valid migration file which shouldn't be possible, because `link_count` is a mutable expression
3. Call `edgedb migrate`
4. An error `edgedb error: UnsupportedFeatureError: unexpectedly long path in simple expr` is shown

Schema:
```esdl
module default {
  Type Sample{
    multi link simple_link -> Sample;
    property link_count := count(.<simple_link[IS Sample]);
    index on (.link_count);
  }
}
```

aljazerzen:
The problem here is not the computed, but rather using aggregation and accessing links from indexes.

#4869 adds a check for this when migration is created.

This can be cherry-picked to be released in 2.8.

PR #4869, opened 2023-01-04
Fix: forbid aggregation in index expressions

aljazerzen:
Add a check on indexes that prevents aggregation, similar to what's already on the constraint expressions.

### DIFF
--- a/edb/schema/indexes.py
+++ b/edb/schema/indexes.py
@@ -292,6 +292,9 @@ class IndexCommand(
         value: s_expr.Expression,
         track_schema_ref_exprs: bool=False,
     ) -> s_expr.CompiledExpression:
+        from edb.ir import utils as irutils
+        from edb.ir import ast as irast
+
         singletons: List[s_types.Type]
         if field.name in {'expr', 'except_expr'}:
             # type ignore below, for the class is used as mixin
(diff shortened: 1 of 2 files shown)


aljazerzen:
This is the relevant chunk from constraints: 
https://github.com/edgedb/edgedb/blob/ba352d9944765951241bde86102a374f47cd90da/edb/schema/constraints.py#L863-L913

@msullivan what does this prevent exactly? https://github.com/edgedb/edgedb/blob/ba352d9944765951241bde86102a374f47cd90da/edb/schema/constraints.py#L885-L896

Because it did not catch this:
```
TYPE ObjIndex2 {
    LINK foo -> ObjIndex1;
    INDEX ON (.foo.name);
};
```
(but it did get caught by immutability check, which is also strange to me).

aljazerzen:
Ref #4796

msullivan (approved):

msullivan:
> This is the relevant chunk from constraints:
> 
> https://github.com/edgedb/edgedb/blob/ba352d9944765951241bde86102a374f47cd90da/edb/schema/constraints.py#L863-L913
> 
> @msullivan what does this prevent exactly?
> 
> https://github.com/edgedb/edgedb/blob/ba352d9944765951241bde86102a374f47cd90da/edb/schema/constraints.py#L885-L896
> 
> Because it did not catch this:
> 
> ```
> TYPE ObjIndex2 {
>     LINK foo -> ObjIndex1;
>     INDEX ON (.foo.name);
> };
> ```
> 
> (but it did get caught by immutability check, which is also strange to me).

It should prevent multi-hop paths like that. You've put your code after the immutability check, so it makes sense immutability triggered first. (Unless you originally had it before that check, in which case I don't know why it didn't work.)

The reason the immutability check catches it, by the way, is that accessing the link makes it STABLE, since it needs to access the database. The access to the object itself isn't, because it's marked as a "singleton" in the options.

PR #4873, opened 2023-01-05
DML in fill_expr & conv_expr

aljazerzen:
Because DML CTEs must be top-level, I had to pull them up to be on the same level as INSERT/UPDATE.

The problem is that INSERT & UPDATE statements are hardcoded SQL, so I had to generate source for CTEs separately and concat with INSERT/UPDATE statement SQL.

I did try to convert INSERT & UPDATE to be instead defined with pgast, but it was way too verbose (see the intermediate commit).

Note that DML is still not supported for `cast_expr`.

### DIFF
--- a/edb/pgsql/delta.py
+++ b/edb/pgsql/delta.py
@@ -299,7 +299,7 @@ class Query(MetaCommand, adapts=sd.Query):
         schema = super().apply(schema, context)
 
         assert self.expr.irast
-        sql_tree = compiler.compile_ir_to_sql_tree(
+        sql_tree, _ = compiler.compile_ir_to_sql_tree(
             self.expr.irast,
             output_format=compiler.OutputFormat.NATIVE_INTERNAL,
             explicit_top_cast=irtyputils.type_to_typeref(
@@ -1016,7 +1016,7 @@ class FunctionCommand(MetaCommand):
             if not irutils.is_const(ir.expr):
                 raise ValueError('expression not constant')
 
-            sql_tree = compiler.compile_ir_to_sql_tree(
+            sql_tree, _ = compiler.compile_ir_to_sql_tree(
                 ir.expr, singleton_mode=True)
             return codegen.SQLSourceGenerator.to_source(sql_tree)
 
@@ -3289,7 +3289,7 @@ class CreateIndex(IndexCommand, adapts=s_indexes.CreateIndex):
         table_name = common.get_backend_name(
             schema, subject, catenate=False)
 
-        sql_tree = compiler.compile_ir_to_sql_tree(
+        sql_tree, _ = compiler.compile_ir_to_sql_tree(
             ir.expr, singleton_mode=True)
 
         if isinstance(sql_tree, pg_ast.ImplicitRowExpr):
@@ -3307,7 +3307,7 @@ class CreateIndex(IndexCommand, adapts=s_indexes.CreateIndex):
                 options=options,
             )
         if except_expr:
-            except_tree = compiler.compile_ir_to_sql_tree(
+            except_tree, _ = compiler.compile_ir_to_sql_tree(
                 except_expr.irast.expr, singleton_mode=True)
             except_src = codegen.SQLSourceGenerator.to_source(except_tree)
             except_src = f'({except_src}) is not true'
@@ -3711,21 +3711,29 @@ class PointerMetaCommand(
         if isinstance(source_op, sd.CreateObject):
             return
 
-        with_sql = ''
+        tab = q(*ptr_stor_info.table_name)
+        target_col = ptr_stor_info.column_name
+        source = ptr.get_source(orig_schema)
+        src_tab = q(*common.get_backend_name(
+            orig_schema,
+            source,
+            catenate=False,
+        ))
+
+        # initial extern relvar (see docs of _compile_conversion_expr)
+        source_rel = textwrap.dedent(f'''\
+            SELECT * FROM {src_tab}
+        ''')
+        source_rel_alias = f'source_{uuidgen.uuid1mc()}'
+
         if self.conv_expr is not None:
-            (
-                _,
-                conv_expr_sql,
-                orig_rel_alias,
-                _,
-                with_sql,
-            ) = self._compile_conversion_expr(
-                pointer=ptr,
-                conv_expr=self.conv_expr,
+            (conv_expr_ctes, _) = self._compile_conversion_expr(
+                ptr,
+                self.conv_expr,
+                source_rel_alias,
                 schema=schema,
                 orig_schema=orig_schema,
                 context=context,
-                orig_rel_is_always_source=True,
                 target_as_singleton=False,
                 check_non_null=is_required and not is_multi
             )
@@ -3735,16 +3743,14 @@ class PointerMetaCommand(
                     'explicit conversion expression was expected'
                     ' for multi->single transition'
                 )
-            
-            # single -> multi
-            orig_rel_alias = f'alias_{uuidgen.uuid1mc()}'
-            conv_expr_sql = (
-                f'SELECT '
-                f'{qi(orig_rel_alias)}.{qi(old_ptr_stor_info.column_name)}'
-            )
 
-        tab = q(*ptr_stor_info.table_name)
-        target_col = ptr_stor_info.column_name
+            # single -> multi
+            conv_expr_ctes = textwrap.dedent(f'''\
+                _conv_rel(val, id) AS (
+                    SELECT {qi(old_ptr_stor_info.column_name)}, id
+                    FROM {qi(source_rel_alias)}
+                )
+            ''')
 
         if not is_multi:
             # Moving from pointer table to source table.
@@ -3763,9 +3769,12 @@ class PointerMetaCommand(
             self.pgops.add(alter_table)
 
             update_qry = textwrap.dedent(f'''\
-                {with_sql}
-                UPDATE {tab} AS {qi(orig_rel_alias)}
-                SET {qi(target_col)} = ({conv_expr_sql})
+                WITH
+                "{source_rel_alias}" AS ({source_rel}),
+                {conv_expr_ctes}
+                UPDATE {tab} AS _update
+                SET {qi(target_col)} = _conv_rel.val
+                FROM _conv_rel WHERE _update.id = _conv_rel.id
             ''')
             self.pgops.add(dbops.Query(update_qry))
 
@@ -3792,20 +3801,11 @@ class PointerMetaCommand(
             ))
 
             update_qry = textwrap.dedent(f'''\
-                {with_sql}
+                WITH
+                "{source_rel_alias}" AS ({source_rel}),
+                {conv_expr_ctes}
                 INSERT INTO {tab} (source, target)
-                (
-                    SELECT
-                        {qi(orig_rel_alias)}.id,
-                        q.val
-                    FROM
-                        {src_tab} AS {qi(orig_rel_alias)},
-                        LATERAL (
-                            {conv_expr_sql}
-                        ) AS q(val)
-                    WHERE
-                        q.val IS NOT NULL
-                )
+                (SELECT id, val FROM _conv_rel WHERE _conv_rel.val IS NOT NULL)
             ''')
 
             if not is_scalar:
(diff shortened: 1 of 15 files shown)


edb/pgsql/parser/ast_builder.py
```diff
@@ -857,7 +857,7 @@ def _build_insert_target(n: Node, c: Context) -> pgast.InsertTarget:
     )
 
 
-def _build_update_target(n: Node, c: Context) -> pgast.UpdateTarget:
+def _build_update_target(n: Node, c: Context) -> pgast.UpdateTarget:  # no QA
```
elprans, line 860: What is `no QA` for?

tests/test_edgeql_ddl.py
```diff
@@ -2839,6 +2839,37 @@ async def test_edgeql_ddl_ptr_set_type_validation(self):
                 ALTER TYPE Foo ALTER LINK l SET TYPE Spam USING (SELECT Spam)
             """)
 
+    async def test_edgeql_ddl_ptr_using_dml(self):
```
elprans, line 2842: Also need a test for mutations in conversion expressions.

edb/pgsql/delta.py
```diff
@@ -4199,9 +4227,9 @@ def _compile_conversion_expr(
         target_as_singleton: bool = True,
     ) -> Tuple[
         s_expr.Expression,  # Possibly-amended EdgeQL conversion expression
-        str,                # SQL text
-        str,                # original relation alias
-        bool,               # whether SQL expression is trivial
+        pg_ast.SelectStmt,  # SQL text
```
elprans, line 4230:
```suggestion
        pg_ast.SelectStmt,  # SQL AST
```

msullivan (commented):
I think that as implemented, the CTEs will always execute *once*, rather than *once per object being operated on*.

I think we need to compute the rows being operated on in a CTE, and get pgsql to make the CTEs dependent on the input CTE.

We have a mechanism for this: the context tracks an `IteratorCTE`, which during DML compilation represents the "enclosing" iteration CTE (which can correspond to a `for` loop or to another DML statement)

tests/test_edgeql_ddl.py
```diff
@@ -2839,6 +2839,37 @@ async def test_edgeql_ddl_ptr_set_type_validation(self):
...
+        await self.con.execute(r"""
+            CREATE TYPE Hello;
+            CREATE TYPE World {
+                CREATE LINK hell -> Hello;
+                CREATE MULTI LINK heaven -> Hello;
+            };
+            INSERT World {
```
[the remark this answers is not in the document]
msullivan: Also with no `World` objects and with a `World` object where `hell` is already populated. (The thing shouldn't run.)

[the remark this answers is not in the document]
msullivan: Which actually now makes me wonder about what the behavior of `multi -> single` should be if there is already at most one target. Currently it looks like we run the command anyway, which isn't consistent with how it works for setting required

tests/test_edgeql_ddl.py
aljazerzen: You mean casts?

Conversions should be covered by altering the link from multi to single.

edb/pgsql/parser/ast_builder.py
aljazerzen: Nothing anymore...

tests/test_edgeql_ddl.py
```diff
@@ -2839,6 +2839,37 @@ async def test_edgeql_ddl_ptr_set_type_validation(self):
...
+                    (INSERT Hello),
+                    (INSERT Hello),
+                }
+            };
+        """)
+        assert len(await self.con.query('SELECT Hello')) == 3
+
+        await self.con.execute(r"""
+            ALTER TYPE World {
+                ALTER LINK hell SET REQUIRED USING (INSERT Hello);
```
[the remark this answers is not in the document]
aljazerzen: I don't think that's possible. `__subject__` is not allowed and partial_path_prefix within the insert is set to Hello.

(A critic of EdgeQL: these two facts are really not obvious - I had to test them. This is a common pattern in SQL when every clause has different rules on how it's contents are resolved.)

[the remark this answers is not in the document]
msullivan: Sigh, yeah. It's `__source__` here. (Or just `World`).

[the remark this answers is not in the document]
aljazerzen:
```
> cannot reference correlated set 'World' here
```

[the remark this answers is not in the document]
msullivan: sigh, doing something like
```
with z := World.name, insert ...
```
does work, though.

I think the correlated set error is not correct here

tests/test_edgeql_ddl.py
```diff
@@ -2839,6 +2839,98 @@ async def test_edgeql_ddl_ptr_set_type_validation(self):
...
+            r"""
+            CREATE TYPE Hello;
+            CREATE TYPE World {
+                CREATE LINK hell -> Hello;
+                CREATE MULTI LINK heaven -> Hello;
+            };
+            """
+        )
+
+        await self.con.execute(
```
msullivan, line 2900: Check that no inserts happened

aljazerzen:
And yeah, type casting for multi is still busted.

edb/pgsql/parser/ast_builder.py
```diff
@@ -1031,12 +1031,21 @@ def _build_on_conflict(n: Node, c: Context) -> pgast.OnConflictClause:
...
         action=_build_str(n["action"], c),
         infer=_maybe(n, c, "infer", _build_infer_clause),
-        target_list=_build_targets(n, c, "targetList"),
+        target_list=_build_on_conflict_targets(n, c, "targetList"),
         where=_maybe(n, c, "where", _build_base_expr),
         context=_build_context(n, c),
     )
 
 
+def _build_on_conflict_targets(
```
msullivan, line 1040: Are these changes meant to be in this PR? What's the story?

edb/pgsql/delta.py
```diff
@@ -4358,46 +4351,97 @@ def _compile_conversion_expr(
...
+            # volatile expressions are executed once for each object.
+
+            # IR ast wrapping
+            typ = orig_schema.get(f'schema::ObjectType', type=s_types.Type)
+            outer_path = irast.PathId.from_type(
+                orig_schema, typ, typename=sn.QualName("std", "obj")
+            )
+
+            scope_iter = irast.ScopeTreeNode(
+                unique_id=10001,
```
msullivan, line 4373: Let's use negative numbers. (And bind them to names so we don't mention each number more than once.)

edb/pgsql/delta.py
```diff
@@ -4250,6 +4220,32 @@ def _compile_conversion_expr(
         new_target = not_none(pointer.get_target(schema))
 
         if conv_expr.irast is None:
+            if check_non_null and False:
```
msullivan, line 4223: Dead code?

tests/test_edgeql_ddl.py
```diff
@@ -2839,6 +2839,118 @@ async def test_edgeql_ddl_ptr_set_type_validation(self):
...
+            };
+            """
+        )
+        self.assertEqual(len(await self.con.query("SELECT Hello")), 6)
+
+        await self.con.execute(
+            r"""
+            ALTER TYPE World {
+                ALTER LINK hell SET REQUIRED USING (INSERT Hello);
+                ALTER LINK heaven SET SINGLE USING (INSERT Hello);
```
msullivan, line 2872: We definitely want a test that inserts in a way that refers to existing data.
I mentioned a workaround for the correlated set thing, but I think also we need to fix the correlated set error in this case, since I think it is an important use case.

msullivan (commented):
This is looking good!

aljazerzen:
@msullivan the failing test is:

```
============================================================================================================================================================================================================================================================================================
ERROR: test_edgeql_ddl_ptr_set_type_using_01 (test_edgeql_ddl.TestEdgeQLDDL.test_edgeql_ddl_ptr_set_type_using_01)
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Py_HashSecret: 71e54da6077ee492b7ab2342b94356ad37bbfe05756044a4
random.seed(): 55a8dab2709fd31f
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Server Traceback:
edb.errors.InternalServerError: duplicate key value violates unique constraint "8a7ee8ae-a660-11ed-b038-8fcadad7b31e_source_target_key"

Test Traceback:
  File "/home/aljaz/EdgeDB/edgedb/edb/testbase/server.py", line 150, in wrapper
    self.loop.run_until_complete(
  File "/usr/lib/python3.11/asyncio/base_events.py", line 653, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/home/aljaz/EdgeDB/edgedb/tests/test_edgeql_ddl.py", line 2573, in test_edgeql_ddl_ptr_set_type_using_01
    await self.con.execute("""
  File "/home/aljaz/EdgeDB/venv/lib/python3.11/site-packages/edgedb/abstract.py", line 297, in execute
    await self._execute(ExecuteContext(
  File "/home/aljaz/EdgeDB/edgedb/edb/testbase/connection.py", line 375, in _execute
    await self._protocol.execute(
  File "edgedb/protocol/protocol.pyx", line 557, in execute
  File "edgedb/protocol/protocol.pyx", line 462, in _execute
edgedb.errors.InternalServerError: duplicate key value violates unique constraint "8a7ee8ae-a660-11ed-b038-8fcadad7b31e_source_target_key"
```

Script with minimal repro:
```
CREATE TYPE script7::Bar {
    CREATE PROPERTY name -> str;
};
CREATE TYPE script7::SubBar EXTENDING script7::Bar;
CREATE TYPE script7::Foo {
    CREATE MULTI LINK m_l -> script7::Bar {
        CREATE PROPERTY lp -> str;
    };
};

INSERT script7::Bar {name := 'bar1'};
INSERT script7::SubBar {name := 'bar2'};

WITH
    bars := (SELECT script7::Bar),
INSERT script7::Foo {
    m_l := (
        FOR bar IN {enumerate(bars)}
        UNION (SELECT bar.1 { @lp := <str>(bar.0 + 1) })
    ),
};

WITH
    bars := (SELECT script7::Bar),
INSERT script7::Foo {
    m_l := (
        FOR bar IN {enumerate(bars)}
        UNION (SELECT bar.1 { @lp := <str>(bar.0 + 3) })
    )
};

ALTER TYPE script7::Foo ALTER LINK m_l {
    SET TYPE script7::SubBar USING (.m_l[IS script7::SubBar])
}
```

Problem is that this WHERE clause is not getting generated:
```diff sql
 UPDATE edgedbpub."6e886670-a63c-11ed-ac04-5f2ed5b130d2" AS "_source"
 SET the_column_name = (
     FROM LATERAL (
         FROM edgedbpub."6e1a196f-a63c-11ed-a297-25b0f88a8190_t" AS "Bar~1"
         CROSS JOIN LATERAL (
             FROM edgedbpub."6e4f916a-a63c-11ed-847e-4fcfb36d9113_t" AS "SubBar~1"
             SELECT "SubBar~1".id AS "indirection_value~1"
             WHERE ("Bar~1".id = "SubBar~1".id)
         ) AS "SubBar_indirection~2"
         SELECT "SubBar_indirection~2"."indirection_value~1" AS "indirection_value~2"
-        WHERE ("_source".target = "Bar~1".id)
     ) AS "expr-1~2"
     SELECT "expr-1~2"."indirection_value~2" AS "expr~1_serialized~1"
 );
```

edb/pgsql/delta.py
```diff
@@ -4420,35 +4364,177 @@ def _compile_conversion_expr(
...
+            unique_id=10001,
+        )
+        scope_body = irast.ScopeTreeNode(
+            unique_id=10002,
+            fenced=True
+        )
+        scope_body.children.extend(ir.scope_tree.children)
+
+        scope_root = irast.ScopeTreeNode(
+            unique_id=10000,
```
msullivan, line 4410: I'd use negative numbers so they don't collide. (And probably assign them all to variables instead of repeating the numbers)

edb/pgsql/delta.py
```diff
@@ -4420,35 +4364,177 @@ def _compile_conversion_expr(
...
+            path_id=outer_path,
+        )
+        scope_root.children.append(scope_iter)
```
msullivan, line 4413: Don't modify `children` directly, use `attach_child`. Modifying it directly won't set the parent pointer in the children. (Note that `attach_child` will remove it from a current parent, so if you are moving nodes from one tree to another you'll need to copy the list)
(Which I think doesn't matter here, but.)

edb/schema/pointers.py
```diff
@@ -2343,6 +2345,13 @@ def _alter_begin(
                         context=self.source_context,
                     )
 
+                if irutils.contains_dml(self.cast_expr.ir_statement):
```
msullivan, line 2348: I just realized that we can probably implement this using `ctid`. (And even without that we could do it for links, I think).
This is lower priority than fill, though, so we can push this off to 3.1 at this point, probably. File an issue to track it though.

tests/test_edgeql_ddl.py
```diff
@@ -2839,6 +2839,144 @@ async def test_edgeql_ddl_ptr_set_type_validation(self):
...
+            };
+            CREATE TYPE World {
+                CREATE PROPERTY foo -> str;
+                CREATE LINK hell -> Hello;
+            };
+            INSERT World { foo := 'hello' };
+            """
+        )
+
+        await self.con.execute(
```
msullivan, line 2971: Check that `Hello` inserted properly

edb/pgsql/delta.py
```diff
@@ -4086,28 +4061,32 @@ def _alter_pointer_type(self, pointer, schema, orig_schema, context):
...
 
+        tab = q(*old_ptr_stor_info.table_name)
+        target_col = old_ptr_stor_info.column_name
+        aux_ptr_table = None
+        aux_ptr_col = None
+
+        source_rel_alias = f'source_{uuidgen.uuid1mc()}'
+
         # There are two major possibilities about the USING claus:
         # 1) trivial case, where the USING clause refers only to the
```
msullivan, line 4072: We've eliminated the trivial case (which never really fired), so update the comment?

msullivan: Maybe explaining that we could consider actually detecting/optimizing it in the future?

edb/pgsql/delta.py
```diff
@@ -4275,28 +4227,42 @@ def _alter_pointer_type(self, pointer, schema, orig_schema, context):
...
-        str,                # SQL text
-        str,                # original relation alias
-        bool,               # whether SQL expression is trivial
+        str,  # SQL
+        bool,  # is_nullable
     ]:
+        """
+        Compile USING expression of an ALTER statement.
+
+        producing_ctes contract:
```
msullivan, line 4247: `produce_ctes`

edb/pgsql/delta.py
```diff
@@ -4420,35 +4364,181 @@ def _compile_conversion_expr(
...
+        typ = orig_schema.get(f'schema::ObjectType', type=s_types.Type)
+        outer_path = irast.PathId.from_type(
+            orig_schema, typ, typename=sn.QualName("std", "obj")
+        )
+
+        root_uid = -1
+        iter_uid = -2
+        body_uid = -3
+        # scope tree wrapping is roughy equivalent to:
+        # "(std::obj) uid:-1": {
```
msullivan, line 4399: I think we don't actually need to tag the outer node with `(std::obj)`

msullivan (approved):
Great work on this!

A few nits about docs and tests, and we'll want to address casts at some point, but I think we're good

PR #4878, opened 2023-01-07
Add syntax highlighting

ShalokShalom:
This looks prettier :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ Forget foreign keys; tabular data modeling is a relic of an older age, and it
 with modern languages. Instead, EdgeDB thinks about schema the same way you do:
 as **object types** containing **properties** connected by **links**.
 
-```
+```fs
 type Person {
   required property name -> str;
 }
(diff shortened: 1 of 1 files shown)


ShalokShalom:
You can preview it here:

raddevon:
This seems to work reasonably well for the SDL block (the first block), but the EdgeQL highlights (the other two blocks) are a bit odd. I'm also concerned about the implications even on the first block if we end up changing the example, since the highlighter is intended for a different syntax.

What do you think, @elprans? @msullivan? @1st1?

ShalokShalom:
Would you prefer that?

grafik

haikyuu:
Why not add support for the edgeQL and esdl in github?
[Here is how](https://github.com/github/linguist/blob/master/CONTRIBUTING.md).

It should be straightforward since edgedb already has the [textmate language](https://github.com/edgedb/edgedb-editor-plugin/blob/master/grammars/edgeql.tmLanguage)

ShalokShalom:
Ah, I didnt suggest this, since I thought Edge uses its own language. 
And Github adds only stuff with 300+ external repos

haikyuu:
Opened an issue for github syntax highlighting to separate from this PR https://github.com/edgedb/edgedb/issues/4893

1st1:
@raddevon 

> What do you think, @elprans? @msullivan? @1st1?

IMO this PR is great.

1st1 (approved):

PR #4881, opened 2023-01-10
docs: a stub for client libraries section

tailhook:
Also includes initial description for the rounding during encoding/decoding datetimes

### DIFF
--- a/docs/reference/bindings/datetime.rst
+++ b/docs/reference/bindings/datetime.rst
@@ -1,0 +1,92 @@
+.. _ref_bindings_datetime:
+
+==================
+Date/Time Handling
+==================
+
+EdgeDB has 6 types related to date and time handling:
+
+* :eql:type:`datetime` (:ref:`binary format <ref_protocol_fmt_datetime>`)
+* :eql:type:`duration` (:ref:`binary format <ref_protocol_fmt_duration>`)
+* :eql:type:`cal::local_datetime`
+  (:ref:`binary format <ref_protocol_fmt_local_datetime>`)
+* :eql:type:`cal::local_date`
+  (:ref:`binary format <ref_protocol_fmt_local_date>`)
+* :eql:type:`cal::relative_duration`
+  (:ref:`binary format <ref_protocol_fmt_relative_duration>`)
+* :eql:type:`cal::date_duration`
+  (:ref:`binary format <ref_protocol_fmt_date_duration>`)
+
+Usually we try to map those types to the respective language-native types,
+with the following caveats:
+
+* The type in standard library
+* It has enough range (EdgeDB has timestamps from year 1 to 9999)
+* And it has good enough precision (at least microseconds)
+
+If any of the above is not true we tend to provide a custom type in the client
+library itself that is convertible to whatever type language or any popular
+external dependencies have. Exceptions: JavaScript has millisecond precision on
+their ``Date`` type (which is actually a timestamp) and we decided that it's
+better to use that type by default even though it doesn't have enough
+precision.
+
+
+Precision
+=========
+
+:eql:type:`datetime`, :eql:type:`duration`, :eql:type:`cal::local_datetime` and
+:eql:type:`cal::relative_duration` all have precision of **1 millisecond**.
+
+This means that if language-native type have a bigger precision such as
+nanosecond, client library has to round that timestamp when encoding it for
+EdgeDB.
+
+We use **rouding to the nearest even** for that operation. Here are some
+examples of timetamps with high precision, and how they are stored in the
+database::
+
+    2022-02-24T05:43:03.123456789Z → 2022-02-24T05:43:03.123457Z
+
+    2022-02-24T05:43:03.000002345Z → 2022-02-24T05:43:03.000002Z
+    2022-02-24T05:43:03.000002500Z → 2022-02-24T05:43:03.000002Z
+    2022-02-24T05:43:03.000002501Z → 2022-02-24T05:43:03.000003Z
+    2022-02-24T05:43:03.000002499Z → 2022-02-24T05:43:03.000002Z
+
+    2022-02-24T05:43:03.000001234Z → 2022-02-24T05:43:03.000001Z
+    2022-02-24T05:43:03.000001500Z → 2022-02-24T05:43:03.000002Z
+    2022-02-24T05:43:03.000001501Z → 2022-02-24T05:43:03.000002Z
+    2022-02-24T05:43:03.000001499Z → 2022-02-24T05:43:03.000001Z
+
+.. note::
+
+   A quick refresher on rounding types: If we perform multiple operations of
+   summing while rounding half-up or rounding half-down, the error margin of
+   the resulting value tends to increase. If we round half-to-even instead,
+   the expected value of summing tends to be more accurate.
+
+Note as described in :ref:`datetime protocol documentation
+<ref_protocol_fmt_datetime>` the value is encoded as a *signed* microseconds
+delta since a fixed time. Some care must be taken when rounding negative
+microsecond values. See `tests for Rust implementation`_ for a good set of
+test cases.
+
+Rounding to the nearest even applies to all operations that client libraries
+perform, in particular:
+
+1. Encoding timestamps *and* time deltas (see the :ref:`list of types
+   <ref_bindings_datetime>`) to the binary format if precision of the native
+   type is higher than microseconds.
+2. Decoding timestamps *and* time deltas from the binary format is precision
+   of native type is lower than microseconds (applies for JavaScript for
+   example)
+3. Converting from EdgeDB specific type (if there is one) to native type and
+   back (depending on the difference in precision)
+4. Parsing a string to an EdgeDB specific type (this operation is optional to
+   implement, but if it is implemented, it must obey the rules)
+
+.. lint-off
+
+.. _tests for Rust implementation: https://github.com/edgedb/edgedb-rust/tree/master/edgedb-protocol/tests/datetime_chrono.rs
+
+.. lint-on
--- a/docs/reference/bindings/index.rst
+++ b/docs/reference/bindings/index.rst
@@ -1,0 +1,40 @@
+.. _ref_bindings_overview:
+
+================
+Client Libraries
+================
+
+EdgeDB client libraries are a bit higher level than usual database bindings.
+In particular, they contain:
+
+* Structured data retrieval
+* Connection pooling
+* Retrying of failed transactions and queries
+
+Additionally, client libraries might provide:
+
+* Code generation for type-safe database access
+* Query builder
+
+This is a **work-in-progress** reference for writing client libraries for
+EdgeDB.
+
+External Links:
+
+* :ref:`Official Client Libraries <ref_clients_index>`
+* :ref:`Binary protocol <ref_protocol_overview>`
+* `RFC 1004`_ - Robust Client API
+
+Contents:
+
+.. toctree::
+    :maxdepth: 3
+
+    datetime
+
+
+.. lint-off
+
+.. _RFC 1004: https://github.com/edgedb/rfcs/blob/master/text/1004-transactions-api.rst
+
+.. lint-on
--- a/docs/reference/protocol/dataformats.rst
+++ b/docs/reference/protocol/dataformats.rst
@@ -419,6 +419,9 @@ encoded as:
 
     0x00 0x02 0x2b 0x35 0x9b 0xc4 0x10 0x00
 
+See :ref:`client libraries <ref_bindings_datetime>` section for more info
+about how to handle different precision when encoding data.
+
 
 .. _ref_protocol_fmt_local_datetime:
 
@@ -435,6 +438,9 @@ encoded as:
 .. code-block:: c
 
     0x00 0x02 0x2b 0x35 0x9b 0xc4 0x10 0x00
+
+See :ref:`client libraries <ref_bindings_datetime>` section for more info
+about how to handle different precision when encoding data.
 
 
 .. _ref_protocol_fmt_local_date:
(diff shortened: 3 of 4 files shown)


docs/reference/bindings/datetime.rst
```diff
@@ -0,0 +1,85 @@
...
+    2022-02-24T05:43:03.000001234Z → 2022-02-24T05:43:03.000001Z
+    2022-02-24T05:43:03.000001500Z → 2022-02-24T05:43:03.000002Z
+    2022-02-24T05:43:03.000001501Z → 2022-02-24T05:43:03.000002Z
+    2022-02-24T05:43:03.000001499Z → 2022-02-24T05:43:03.000001Z
+
+Note as described in :ref:`datetime protocol documentation
+<ref_protocol_fmt_datetime>` the value is encoded as a *signed* microseconds
+delta since a fixed time. Some care must be taken when rounding negative
+microsecond values. See `tests for Rust implementation`_ for a good set of
+test cases.
```
raddevon, line 65:
```suggestion
.. note::

    As described in our :ref:`datetime protocol documentation
    <ref_protocol_fmt_datetime>`, the value is encoded as a *signed*
    microseconds delta since a fixed time. Some care must be taken when
    rounding negative microsecond values. See `tests for Rust implementation`_
    for a good set of test cases.
```

docs/reference/bindings/datetime.rst
```diff
@@ -0,0 +1,85 @@
...
+
+Rounding to the nearest even applies to all operations that client libraries
+perform, in particular:
```
raddevon, line 68:
```suggestion
EdgeDB client libraries round to the nearest even for all operations they
perform that require rounding, in particular:
```

docs/reference/bindings/datetime.rst
```diff
@@ -0,0 +1,85 @@
...
+1. Encoding timestamps *and* time deltas (see the :ref:`list of types
+   <ref_bindings_datetime>`) to the binary format if precision of the native
+   type is higher than microseconds.
+2. Decoding timestamps *and* time deltas from the binary format is precision
+   of native type is lower than microseconds (applies for JavaScript for
+   example)
+3. Converting from EdgeDB specific type (if there is one) to native type and
+   back (depending on the difference in precision)
+4. Parsing a string to an EdgeDB specific type (this operation is optional to
+   implement, but if it is implemented, it must obey the rules)
```
raddevon, line 79:
```suggestion
* Encoding timestamps *and* time deltas (see the :ref:`list of datetime types
  <ref_bindings_datetime>`) to the binary format if precision of the native
  type is greater than microseconds.
* Decoding timestamps *and* time deltas from the binary format if precision
  of the native type is less than microseconds (applies to JavaScript for
  example)
* Converting from EdgeDB-specific type (if there is one) to a native type and
  back when those types differ in their precision
* Parsing a string to an EdgeDB-specific type (Implementation is not required,
  but if it *is* implemented, it must use this rounding strategy.)
```

docs/reference/bindings/index.rst
```diff
@@ -0,0 +1,40 @@
+.. _ref_bindings_overview:
+
+================
+Client Libraries
```
raddevon, line 4:
```suggestion
Client libraries
```

docs/reference/bindings/index.rst
```diff
@@ -0,0 +1,40 @@
...
+================
+
+EdgeDB client libraries are a bit higher level than usual database bindings.
+In particular, they contain:
```
raddevon, line 8:
```suggestion
EdgeDB client libraries are higher level than typical database bindings. In
particular, they contain:
```

docs/reference/bindings/index.rst
```diff
@@ -0,0 +1,40 @@
...
+
+* Structured data retrieval
+* Connection pooling
+* Retrying of failed transactions and queries
+
+Additionally, client libraries might provide:
+
+* Code generation for type-safe database access
+* Query builder
```
raddevon, line 17:
```suggestion
* A query builder
```

docs/reference/protocol/dataformats.rst
```diff
@@ -419,6 +419,9 @@ encoded as:
 
     0x00 0x02 0x2b 0x35 0x9b 0xc4 0x10 0x00
 
+See :ref:`client libraries <ref_bindings_datetime>` section for more info
```
raddevon, line 422:
```suggestion
See the :ref:`client libraries <ref_bindings_datetime>` section for more info
```

docs/reference/protocol/dataformats.rst
```diff
@@ -436,6 +439,9 @@ encoded as:
...
```
raddevon, line 442:
```suggestion
See the :ref:`client libraries <ref_bindings_datetime>` section for more info
```

docs/reference/protocol/dataformats.rst
```diff
@@ -470,6 +476,9 @@ encoded as:
 
     0x00 0x00 0x00 0x0a 0x32 0xae 0xf6 0x00
 
+See :ref:`client libraries <ref_bindings_datetime>` section for more info
```
raddevon, line 479:
```suggestion
See the :ref:`client libraries <ref_bindings_datetime>` section for more info
```

docs/reference/protocol/dataformats.rst
```diff
@@ -505,6 +514,9 @@ encoded as:
     // months
     0x00 0x00 0x00 0x00
 
+See :ref:`client libraries <ref_bindings_datetime>` section for more info
```
raddevon, line 517:
```suggestion
See the :ref:`client libraries <ref_bindings_datetime>` section for more info
```

docs/reference/protocol/dataformats.rst
```diff
@@ -536,6 +548,9 @@ For example, the ``cal::relative_duration`` value
     // months
     0x00 0x00 0x00 0x1f
 
+See :ref:`client libraries <ref_bindings_datetime>` section for more info
```
raddevon, line 551:
```suggestion
See the :ref:`client libraries <ref_bindings_datetime>` section for more info
```

raddevon (commented):
This is great work, Paul! If I've not captured your meaning in any of my suggestions, I'm happy to work with you further on them. There's one in particular I need some help understanding before I can provide a suggestion. I commented on it to that effect.

docs/reference/bindings/datetime.rst
```diff
@@ -0,0 +1,85 @@
+.. _ref_bindings_datetime:
+
+==================
+Date/Time Handling
```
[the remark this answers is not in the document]
raddevon: Having second thoughts on this, actually. I think maybe it makes sense to keep your title case and for me to circle back on those that *don't* use title case to bring them in line. Having given it some more consideration and seeing that it does seem to be the prevailing style in other parts of the docs, I think it makes sense.

docs/reference/bindings/datetime.rst
```diff
@@ -0,0 +1,85 @@
...
+* The type in standard library
+* It has enough range (EdgeDB has timestamps from year 1 to 9999)
+* And it has good enough precision (at least microseconds)
+
+If any of the above is not true we tend to provide a custom type in the client
+library itself that is convertible to whatever type language or any popular
+external dependencies have. Exceptions: JavaScript has millisecond precision on
+their ``Date`` type (which is actually a timestamp) and we decided that it's
+better to use that type by default even though it doesn't have enough
+precision.
```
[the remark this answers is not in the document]
tailhook: For example, in Rust we have `edgedb...::Duration` type that can be converted into `std::time::Duration` (stdlib) and `chrono::Duration` (external library) and back via `value.try_into()`. Different languages have different ways to convert types. But I don't know what "compatible" type means.

docs/reference/bindings/datetime.rst
```diff
@@ -0,0 +1,84 @@
...
+* The type is in the language's standard library
+* The type has sufficient range (EdgeDB timestamps range from year 1 to 9999)
+* The type has sufficient precision (at least microseconds)
+
+If any of the above is not true we tend to provide a custom type in the client
+library itself that is convertible to whatever type language or any popular
+external dependencies have. Exceptions: JavaScript has millisecond precision on
+their ``Date`` type (which is actually a timestamp) and we decided that it's
+better to use that type by default even though it doesn't have enough
+precision.
```
raddevon, line 32: Thanks for the clarification, @tailhook. See if this works:

```suggestion
If any of the above criteria are *not* met, we usually provide a custom type
in the client library that can be converted to a language-native type or a 
type provided by a third-party library. Exception: The JavaScript ``Date``
type (which is actually a timestamp) has only millisecond precision, but we
determined it's better to use that type by default even though it doesn't have
sufficient precision.
```

Is there anything we can say here about why we made this call on the JavaScript ``Date`` conversion? I might not go into anything super complicated here, but if we could briefly explain why we made this call, I think it could be useful.

tailhook: Well, more thinking about that, I'm not sure that JavaScript is special. Maybe it's reverse: Rust is special (i.e. Rust has edgedb-specific types, primarily because REPL should represent types exactly). For the datetime type Rust and JavaScript are the only exceptions as far as I remember. I'm not sure about duration/relativedelta.

raddevon: @tailhook Sounds like it might be worth calling out both JavaScript _and_ Rust.

tailhook: I don't think adding more text will help. I think we can say something like "If these criteria don't met, please ping us in chat, or use your best judgement."

fmoor:
Would it make sense to add a note on why we chose `round to even` or a link to something that explains the advantages? This sounds very bizarre if you have never seen it before.

tailhook:
> Would it make sense to add a note on why we chose `round to even` or a link to something that explains the advantages? This sounds very bizarre if you have never seen it before.

Well, yes it sounds bizarre first time. But it makes sense in statistics. Also Postgres does that. Here are few links by quick googling:

1. https://stackoverflow.com/questions/45223778/is-bankers-rounding-really-more-numerically-stable
2. https://softwareengineering.stackexchange.com/questions/256265/why-do-some-languages-round-to-the-nearest-even-integer

I'm not sure these are good to put into the docs. Maybe someone knows the "canonical" link for that.

raddevon:
I went back and forth between thinking this might be out of the documentation scope and liking it as a note, so the user can see it's an aside but we give context inline rather than forcing the user to seek it outside our docs. I think I'm leaning slightly toward the latter, I think.

What would you think about something like this as a `.. note::`? (Also, have I accurately reflected our reasoning for doing this? This was my assumption when I noticed it.)

> You may not be familiar with rounding to even. Most rounding goes in the same direction every time: either up from the midpoint or down. This can be a problem because it means you are rounding more often in the direction you choose. Ideally, for rounded numbers to, in aggregate, be more representative of the values they were rounded from, you would want to round up half the time and down the other half.
>
> By rounding to the nearest even number, you are no longer always rounding in a single direction. If the even is down, you round down. If the even is up, you round up. Rounding to even allows you to balance out your rounding by considering a second factor in determining which direction you will round.

fmoor:
I personally would prefer a link to something external. It seems like that would give a more thorough explanation and example algorithm. It is also less text in our docs. The problem is, as @tailhook pointed out, that there is not a great canonical link to point to. I found [this one](https://blogs.sas.com/content/iml/2019/11/11/round-to-even.html) clear and to the point, but who knows how long it will be there.

fmoor:
Also, for round-to-even negative numbers are treated differently than positive numbers, but we don't explicitly mention what dates if any we consider negative. It seems we treat all dates as positive numbers. This information is given implicitly with the examples but it might be nice if this was explicitly called out. There are lots of dates that could be considered the threshold between positive and negative. All of them except maybe the big bang are rather arbitrary so it is hard to guess.

tailhook:
@fmoor
> Also, for round-to-even negative numbers are treated differently than positive numbers,

I think they also round to nearest even. And evens are the same. What is the difference?

> but we don't explicitly mention what dates if any we consider negative.

Well, intuitively dates are all positive. At least starting from year 1 which we support. (Unix) epoch timestamps can be negative, but this is not how we think of dates even if they are implemented like that.

But even with negative timestamps around any given epoch, rounding doesn't happen to be any different on whether it's positive or negative. Maybe some specific calculation doesn't work for negative numbers, but it doesn't change the docs unless we provide some example formulas.

@raddevon 
I would like to include good summary in the note, but I don't think what you propose summarizes it well. And I fail to come up with a good summary myself.

fmoor:
> I think they also round to nearest even. And evens are the same. What is the difference?
evens round towards zero and odds away from zero

> Well, intuitively dates are all positive.
This is subjective. Others may not have the same intuition. I was thinking of dates as a unix epoch and initially took the sign into account. This seems like an easy detail to include.

raddevon:
@tailhook I'm happy to rework my summary given some direction. This is how I understand it, but maybe the problem is that I _don't_ understand it, or at least not well enough. 😅

tailhook:
> @tailhook I'm happy to rework my summary given some direction. This is how I understand it, but maybe the problem is that I _don't_ understand it, or at least not well enough. sweat_smile

Well, I don't know. What you have written is intuitively kinda right. But is scientifically wrong :) Which is probably okay to some degree.

Thinking more about it would provide just some direction of the research rather than explaining. Like this:

> When you perform multiple operations of summing up and rounding values, the the error margin of the resulting value tends to increase. When using rounding half-to-even the expected value of such calculation should be the most accurate (while in the "half-up" method, for example, expected value tends to become bigger the more calculations are done)

raddevon:
That sounds good, @tailhook. I think that's what I meant, but I didn't do the best job getting it across. I'll rework my explanation around your suggestion. Thank you!

raddevon:
> If we perform multiple operations of summing while rounding half-up or rounding half-down, the error margin of the resulting value tends to increase. If we round half-to-even instead, the expected value of summing tends to be more accurate.

See if this seems accurate and still captures everything we want to capture, @tailhook. I tried to reframe it a bit more as an explanation in hopes of anchoring it more within the context, while hopefully retaining accuracy.

tailhook:
Oops, sorry. It looks like I've somehow pushed "merge" button by a mistake :( Feel free to make a PR with your suggestion.

raddevon:
It's my fault. I should have used "request changes" instead of "approve." 😅  I'll put in a PR. Thanks, @tailhook!

PR #4887, opened 2023-01-13
Replace for_each_query_in_set with an iterator based version

msullivan:
It is more convenient in most use cases

### DIFF
--- a/edb/pgsql/compiler/astutils.py
+++ b/edb/pgsql/compiler/astutils.py
@@ -151,18 +151,18 @@ def get_leftmost_query(query: pgast.Query) -> pgast.Query:
     return result
 
 
-def for_each_query_in_set(
-    qry: pgast.Query,
-    cb: Callable[[pgast.Query], Any],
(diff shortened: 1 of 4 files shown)


edb/pgsql/compiler/astutils.py
```diff
@@ -151,18 +151,18 @@ def get_leftmost_query(query: pgast.Query) -> pgast.Query:
...
-) -> List[Any]:
-    if is_set_op_query(qry):
-        assert qry.larg and qry.rarg
-        result = for_each_query_in_set(qry.larg, cb)
-        result.extend(for_each_query_in_set(qry.rarg, cb))
-    else:
-        result = [cb(qry)]
-
-    return result
+def each_query_in_set(qry: pgast.Query) -> Iterator[pgast.Query]:
```
msullivan, line 154: "Print out the nodes of a binary tree" was the first interview question I got when I interviewed for my first programming internship in 2005 :P

elprans (approved):

PR #4889, opened 2023-01-13
FTS index and functions

vpetrovykh:


edb/pgsql/delta.py
```diff
@@ -3263,19 +3265,54 @@ def create_index(cls, index, schema, context):
...
+                as_fragment=True,
+            )
+            kw_ir = kw_expr.irast
+            kw_sql_tree = compiler.compile_ir_to_sql_tree(
+                kw_ir.expr, singleton_mode=True)
+            sql = codegen.SQLSourceGenerator.to_source(kw_sql_tree)
+            # HACK: the compiled SQL is expected to have some unnecessary
+            # casts, strip casts to text as they mess with the requirement
+            # that index expressions are IMMUTABLE.
+            if sql.endswith('::text'):
```
msullivan, line 3300: Hmmmmm this is very sketchy. What casts were running afoul of this?

edb/schema/indexes.py
```diff
@@ -125,6 +129,23 @@ class Index(
         inheritable=False,
     )
 
+    params = so.SchemaField(
```
msullivan, line 132: Add a comment around here making clear when `params` appears and when `kwargs` does?

edb/schema/indexes.py
```diff
@@ -185,6 +206,57 @@ def get_displayname_static(cls, name: sn.Name) -> str:
         shortname = cls.get_shortname_static(name)
         return shortname.name
 
+    def get_all_kwargs(
+        self,
+        schema: s_schema.Schema,
+    ) -> s_expr.ExpressionDict:
+        kwargs = s_expr.ExpressionDict()
+        all_kw = self.__class__.get_field('kwargs').merge_fn(
```
msullivan, line 214: Does just `self.get_field` not work? (Or does that get resolved to something else?)
I'd probably go with `type(self)` over `self.__class__`

edb/schema/indexes.py
```diff
@@ -367,16 +458,56 @@ def _cmd_from_ast(
...
+    ) -> Optional[qlast.DDLOperation]:
+        astnode = super()._get_ast(schema, context, parent_node=parent_node)
+
+        kwargs = self.get_resolved_attribute_value(
+            'kwargs',
+            schema=schema,
+            context=context,
+        )
+        if kwargs:
+            astnode.kwargs = dict()
```
msullivan, line 476: Might be cleaner to do a dict comprehension

edb/schema/indexes.py
```diff
@@ -125,6 +129,23 @@ class Index(
...
+        coerce=True,
+        compcoef=0.4,
+        default=so.DEFAULT_CONSTRUCTOR,
+        inheritable=True,
+    )
+
+    kwargs = so.SchemaField(
+        s_expr.ExpressionDict,
+        coerce=True,
+        compcoef=0.4,
```
msullivan, line 143: Does `ALTER` work for any of this yet? If not maybe compcoef should be 0?

edb/pgsql/metaschema.py
```diff
@@ -4134,6 +4134,130 @@ def __init__(self) -> None:
         )
 
 
+class FTSParseQueryFunction(dbops.Function):
```
msullivan, line 4137: Is this going to be fast enough? I guess there's no way around having to do this kind of computation, if we're doing a string input query.

msullivan: I'd bet that looping instead of doing recursion would help, though?

edb/pgsql/delta.py
```diff
@@ -3263,19 +3265,57 @@ def create_index(cls, index, schema, context):
...
+                kw_expr = expr.ensure_compiled(
+                    schema=schema,
+                    options=kw_options,
+                    as_fragment=True,
+                )
+                kw_ir = kw_expr.irast
+                kw_sql_tree = compiler.compile_ir_to_sql_tree(
+                    kw_ir.expr, singleton_mode=True)
+                sql = codegen.SQLSourceGenerator.to_source(kw_sql_tree)
+                # HACK: the compiled SQL is expected to have some unnecessary
```
msullivan, line 3300: Hmmmmm this is very sketchy. What casts were running afoul of this?

edb/pgsql/delta.py
```diff
@@ -3263,19 +3265,57 @@ def create_index(cls, index, schema, context):
...
+            # constants)
+            # These are expected to be constants, so we don't have anchors,
+            # path prefixes, etc.
+            kw_options = qlcompiler.CompilerOptions(
+                modaliases=context.modaliases,
+                schema_object_context=cls.get_schema_metaclass(),
+                apply_query_rewrites=False,
+            )
+            for name, expr in kwargs.items():
+                # XXX: origin messes up compilation, but by this point we
```
msullivan, line 3288: How does origin mess up the compilation?

tests/test_indexes.py
```diff
@@ -231,3 +231,34 @@ async def test_index_05(self):
                 };
                 """
             )
+
```
msullivan, line 234: We also need some tests for creating new abstract indexes via extending, right?
(Though that's definitely lower priority than FTS, so if it turns out to be pretty busted and tricky to get working we might be able to punt on that for now)

edb/edgeql/parser/grammar/commondl.py
```diff
@@ -639,6 +657,20 @@ def _process_arguments(self, arguments):
 
         return kwargs
 
+    def _process_params_or_kwargs(self, bases, arguments):
+        params = []
+        kwargs = dict()
+
+        # If the definition is extending anotehr abstract index, then we
```
1st1, line 664:
```suggestion
        # If the definition is extending another abstract index, then we
```

edb/pgsql/delta.py
1st1: +1.

edb/schema/reflection/writer.py
```diff
@@ -383,6 +383,46 @@ def _build_object_mutation_shape(
 
             assignments.append(f'{ns}__internal := {shadow_target_expr}')
 
+        elif ftype is sr_struct.FieldType.EXPR_DICT:
+            target_expr = f'''
```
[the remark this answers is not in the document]
vpetrovykh: Updated

edb/schema/indexes.py
vpetrovykh: Set to 0

msullivan (approved):
OK, my understanding is that there are still a number of FTS changes planned, but the index side of things is mostly in place (except for ALTER?), and can be merged?
And then we'll follow up some more with FTS-specific PRs?

vpetrovykh:
Yeah, that's the plan.

PR #4891, opened 2023-01-13
Import asyncpg lazily as it's not always present

fantix:
This fixes the nightly CI

### DIFF
--- a/tests/test_sql_query.py
+++ b/tests/test_sql_query.py
@@ -21,6 +21,11 @@ import os.path
 from edb.testbase import server as tb
 from edb.tools import test
 
+try:
+    import asyncpg
+except ImportError:
+    pass
+
 
 class TestSQL(tb.SQLQueryTestCase):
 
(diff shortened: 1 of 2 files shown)


msullivan (approved):

tests/test_sql_query.py
```diff
@@ -17,7 +17,6 @@
 #
 
 import os.path
-import asyncpg
```
msullivan, line 20: You could also maybe do
```
try:
    import asyncpg
except ImportError:
    pass
```
and add a comment saying that the tests will be skipped during setup if asyncpg isn't present.

Not sure if that's better.

tests/test_sql_query.py
```diff
@@ -335,6 +334,8 @@ async def test_sql_query_24(self):
         self.assertEqual(res, [['Forrest Gump']])
 
     async def test_sql_query_25(self):
+        import asyncpg
```
elprans, line 337: Just guard the import at the top level with `try: except ImportError`.  These tests are not reachable if asyncpg is not available because of the guard in setup.

fantix: OK will do! Looks like flake8 is not complaining that `asyncpg` may be an undefined name.

PR #4896, opened 2023-01-17
Fix `cal::local_time` parsing from `str`.

### LINKED ISSUE #4884: Converting `cal::local_time` from string can result into `24:00:00`
- EdgeDB Version: 3.0-dev.7166+local

Steps to Reproduce:
```
edgedb> SELECT cal::to_local_time('23:59:59.999999500');
edgedb error: ProtocolEncodingError: ProtocolEncodingError: datetime is out of range
```

vpetrovykh:
Change casts from `str` and `json` to `cal::local_time` to disallow creating "24:00:00" value.
Change `cal::to_local_time` to disallow creating "24:00:00" value.

Fixes #4884

### DIFF
--- a/edb/pgsql/metaschema.py
+++ b/edb/pgsql/metaschema.py
@@ -2633,29 +2633,43 @@ class LocalTimeInFunction(dbops.Function):
     """Cast text into time using ISO8601 spec."""
     text = r'''
         SELECT
-            CASE WHEN val !~ (
-                    '^\s*(' ||
-                        '(\d{2}(:\d{2}(:\d{2}(\.\d+)?)?)?|\d{2,6}(\.\d+)?)' ||
-                    ')\s*$'
-                )
+            CASE WHEN date_part('hour', x.t) = 24
             THEN
                 edgedb.raise(
                     NULL::time,
                     'invalid_datetime_format',
                     msg => (
-                        'invalid input syntax for type time: '
+                        'cal::local_time field value out of range: '
                         || quote_literal(val)
-                    ),
-                    detail => (
-                        '{"hint":"Please use ISO8601 format. Examples: '
-                        || '18:43:27 or 18:43 Alternatively '
-                        || '\"to_local_time\" function provides custom '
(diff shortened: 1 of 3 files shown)


edb/pgsql/metaschema.py
```diff
@@ -2633,29 +2633,45 @@ class LocalTimeInFunction(dbops.Function):
...
-                val::time
-            END;
+                x.t
+            END
+        FROM (
+            SELECT
+                CASE WHEN val !~ (
+                        '^\s*(' ||
+                            '(\d{2}(:\d{2}(:\d{2}(\.\d+)?)?)?|\d{2,6}(\.\d+)?)' ||
+                        ')\s*$'
```
tailhook, line 2654: Is there any other time format that postgres supports?

tailhook (approved):

edb/pgsql/metaschema.py
vpetrovykh: Postgres is a bit sloppy about digits in between `:`. So it will accept omitted leading 0, like `'2:3:4'::time` is the same as `'02:03:04'::time`, whereas the ISO standard explicitly indicates that "hh:mm:ss" should have leading 0.

Another thing is that we don't accept reduced precision time that omits seconds altogether, while Postgres can accept "hh:mm" format. I forget what the problems were with reduced precision format at the moment.

msullivan (approved):

msullivan:
(Though at the risk of causing more trouble: would it be useful to /allow/ local_time past 24:00:00? It is useful for representing things like subway time tables when the arrival time is after midnight.)

tailhook:
> (Though at the risk of causing more trouble: would it be useful to /allow/ local_time past 24:00:00? It is useful for representing things like subway time tables when the arrival time is after midnight.)

Postgres doesn't support that (i.e. 24:00:00 max). Stack overflow advises to use interval for that in postgres.

msullivan:
@vpetrovykh good to merge this?

PR #4898, opened 2023-01-17
Static evaluation for SQL

aljazerzen:
Ref #4814 

This commit evaluates this AST:
```
FuncCall:
  name: ('pg_catalog', 'current_schemas')
  args:
  - TypeCast:
      arg:
        StringConstant:
          val: 't'
      typename: 
        name: ('pg_catalog', 'bool')
```
... to:
```
ArrayExpr:
  elements:
  - StringConstant:
      val: 'pg_catalog'
  - StringConstant:
      val: 'public'
```
... where `public` is search_path set by `SET search_path TO x`.

Tests don't yet work, @fantix suspects that it is bug in the pg protocol part. I've also enabled a related test, which also fails, see comments below.


tests/test_sql_query.py
```diff
@@ -599,42 +599,30 @@ async def test_sql_query_introspection_02(self):
...
+            create module my_module;
+            create type my_module::Foo;
+            create type default::Bar;
+        ''')
+        await self.con.query('SELECT my_module::Foo')
+        await self.scon.fetch('SELECT id FROM "my_module"."Foo";') # fails
+        await self.scon.fetch('SELECT id FROM "public"."Person";')
+        await self.scon.fetch('SELECT id FROM "default"."Person";')
+        await self.scon.fetch('SELECT id FROM "public"."Bar";') # fails
+        await self.scon.fetch('SELECT id FROM "default"."Bar";') # fails
```
aljazerzen, line 613: @fantix can you also look at this?

It behaves as it would not be using the new schema.

fantix: This is probably because `TRANSACTION_ISOLATION` on the test case class is by default `True` and the new types are only valid within `self.con` but not `self.scon`.

tests/test_sql_query.py
```diff
@@ -644,9 +632,28 @@ async def test_sql_schemas(self):
...
+        res = await self.squery_values('select current_schema;')
+        self.assertEqual(res, [['public']])
+
+        await self.scon.execute("set search_path to blah")
+        res = await self.squery_values("select current_schema")
+        self.assertEqual(res, [['blah']])
+
+        await self.squery_values('set search_path to blah;')
+        res = await self.squery_values('select current_schema;')
+        # self.assertEqual(res, [['blah']])
```
fantix, line 649: Ah this is failing because of the compile cache ... the previous `select current_schema;` is cached as the statically compiled `select 'public'...` thing and will keep working that way. I'm thinking maybe we could inject the values in the I/O server instead of the compiler, or we'll have to disable the compile cache when it is statically evaluated (a `cachable` flag somewhere?).

fantix: Update after the meeting:

1. It's not the compile cache (the cache key is a tuple of `hash(sql)` and the `fe_settings` which is just `search_path` for now), but rather a reused prepared statement from asyncpg.
2. I'll try to add a hack to re-compile on bind if `fe_settings` changed.

aljazerzen: 3. If:
   - this becomes too hacky or
   - we more functions that depend on the connection state or
   - we want to support `current_schemas( expr_that_is_not_const )`
   ... this static eval approach is no good and we have to implement such functions in SQL and evaluate them in Postgres. The hard part is "how to pass connection state from our server to postgres". It could be done in two ways (but we don't have a consensus which is less painfull):
   - store the state in a temp table which has to be populated every time a backend connection is received from the pool,
   - figure out which connection state variables are used during resolving and pass them to Postgres as query parameters.

aljazerzen:
@fantix is this waiting on me?

fantix:
> @fantix is this waiting on me?

Oh I could also try to fix the test, if that's the last thing remaining here.

aljazerzen:
Yes, please. I'm not exactly sure what to do - something with how we open connections in the testing framework?

fantix:
> Yes, please. I'm not exactly sure what to do - something with how we open connections in the testing framework?

No worries! By default the tests are ran in transactions, but obviously not for both the EdgeDB and SQL connections. By turning off `TRANSACTION_ISOLATION`, the schema added by `self.con.execute()` is then visible to `self.scon.fetch()` within the same test method.

aljazerzen:
Great, thanks.

@msullivan this is ready for review.

edb/pgsql/resolver/expr.py
```diff
@@ -259,15 +260,36 @@ def resolve_FuncCall(
...
+                and isinstance(value, pgast.StringConstant)
+                and isinstance(is_local, pgast.BooleanConstant)
+            ):
+                if (
+                    name.val == "search_path"
+                    and value.val == ""
+                    and not is_local.val
+                ):
+                    return value
+        raise errors.QueryError("unsupported", context=expr.context)
```
msullivan, line 283: Might want a better message?

edb/pgsql/resolver/static.py
```diff
@@ -0,0 +1,176 @@
...
+
+@eval.register
+def eval_SQLValueFunction(
+    expr: pgast.SQLValueFunction,
+    *,
+    ctx: Context,
+) -> pgast.BaseExpr:
+    from edb.pgsql.ast import SQLValueFunctionOP as op
+
+    pass_trough = [
```
msullivan, line 144: spelling: pass_through

tests/test_sql_query.py
```diff
@@ -34,6 +34,7 @@ class TestSQL(tb.SQLQueryTestCase):
     SETUP = os.path.join(
         os.path.dirname(__file__), 'schemas', 'movies_setup.edgeql'
     )
+    TRANSACTION_ISOLATION = False
```
msullivan, line 37: Add a comment saying why isolation is off.

msullivan: Actually it might be cleaner to add the extra objects being created to the schema for the whole test suite.

Otherwise you'll also want to modify that test to clean up the things created also

tests/test_sql_query.py
```diff
@@ -599,54 +600,57 @@ async def test_sql_query_introspection_02(self):
...
-        await self.migrate(
-            '''
-            mod my_module {
-                type Foo;
-            }
-            '''
-        )
-        await self.squery('SELECT id FROM "my_module"."Foo";')
-        await self.squery('SELECT id FROM "public"."Person";')
+    async def test_sql_query_schemas(self):
```
msullivan, line 603: Add a test for calling `set_config` explicitly too

msullivan (approved):
This seems reasonable, though we still might want to consider whether we can do without it by just rewriting calls to be to different functions we implement

PR #4900, opened 2023-01-18
Add debug flag for listening for DAP connections

aljazerzen:
This support my debugging workflow, please approve.


msullivan (approved):
Consider making the port configurable?

aljazerzen:
It's a random big number that has a low chance of clashing with anything, and it's convenient to stay the same, so you can configure your IDE.

I don't have a use for configuring the port and don't see a reason why anyone would, thus concluding it's not worth the effort.

PR #4903, opened 2023-01-19
Adjust how we do cardinality arithmetic

msullivan:
First, abstract away the repeated unzipping code and make it handle
the empty case more nicely.

Then, base it on a notion of addition and multiplication of bounds,
like @zchenb is doing in his draft formalism.

Union cardinality gets a lot cleaner, and cartesian a little cleaner.


edb/edgeql/compiler/inference/cardinality.py
```diff
@@ -121,28 +127,36 @@ def _bounds_to_card(
     )
 
 
+def _card_unzip(
+    args: Iterable[qltypes.Cardinality],
+) -> tuple[tuple[CardinalityBound, ...], tuple[CardinalityBound, ...]]:
+    card = list(zip(*(_card_to_bounds(a) for a in args)))
```
zchenb, line 133: This line is very difficult to understand, what does it do?

zchenb (commented):
All others look good.

edb/edgeql/compiler/inference/cardinality.py
msullivan: This is (believe it or not) a commonish idiom in python for "unzipping" a list of tuples (or of lists, or whatever).
`zip` takes an arbitrary number of "iterable" arguments, and so passing all of the tuples in a list-of-tuples to `zip` as arguments means that we zip all of the tuples together, which has the effect of unzipping the original list-of-tuples.

Downsides are that it is weird and confusing and that it doesn't really behave sensibly on an empty list. (Which it *can't* without knowing the arity of the tuples statically)

zchenb (approved):
Looks good!

PR #4909, opened 2023-01-20
Add introspection via pg_catalog

aljazerzen:
Uses underlying pg_catalog and joins-in our schema to export nice table and columns names.

Ref #4814

### DIFF
--- a/edb/pgsql/metaschema.py
+++ b/edb/pgsql/metaschema.py
@@ -5017,7 +5017,9 @@ def _generate_schema_alias_view(
 
 def _generate_sql_information_schema() -> List[dbops.Command]:
 
-    # a helper table used in many other views
+    # A helper view that contains all data tables we expose over SQL, excluding
+    # introspection tables.
+    # It contains table & schema names and associated module id.
     virtual_tables = dbops.View(
         name=('edgedbsql', 'virtual_tables'),
         query='''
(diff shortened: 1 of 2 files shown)


edb/pgsql/metaschema.py
```diff
@@ -5049,169 +5046,98 @@ def _generate_sql_information_schema() -> List[dbops.View]:
...
+    # A few tables in here were causing problems, so let's hide them as an
+    # implementation detail.
+    # To be more specific:
+    # - following tables were missing from information_schema:
+    #   Link.properties, ObjectType.links, ObjectType.properties
+    # - even though introspection worked, I wasn't able to select from some
+    #   tables in cfg and sys
+
+    # For making up oids of schemas that represent modules
+    uuid_to_oid = dbops.Function(
```
msullivan, line 5079: Hm, what do we think the odds of collision are? How big is oid?

msullivan (approved):
Seems like a big improvement!

edb/pgsql/metaschema.py
elprans: 32-bit signed integers, so technically this can overflow.

edb/pgsql/metaschema.py
```diff
@@ -5267,7 +5207,88 @@ def _generate_sql_information_schema() -> List[dbops.View]:
...
+            typowner,
+            typlen,
+            typbyval,
+            typtype,
+            typcategory,
+            typispreferred,
+            typisdefined,
+            typdelim,
+            typrelid,
+            typsubscript,
```
fantix, line 5224: Looks like `typsubscript` is only added in [PostgreSQL 14](https://www.postgresql.org/docs/14/catalog-pg-type.html), thus this breaks [using EdgeDB on PostgreSQL 13](https://github.com/edgedb/edgedb/actions/runs/3992636910/jobs/6857300141), which will still have a (final) release on [Nov 2025](https://www.postgresql.org/support/versioning/). @elprans should we fix the use of `typsubscript` or drop the support of PostgreSQL 13?

aljazerzen: We can just remove `typsubscript` from here. If this is a change in recent pg version, no tool will depend on it.

PR #4910, opened 2023-01-20
Parse SQL nullif, record indirection op & create table

aljazerzen:
Ref #4814

### DIFF
--- a/edb/pgsql/parser/ast_builder.py
+++ b/edb/pgsql/parser/ast_builder.py
@@ -184,6 +184,7 @@ def _build_stmt(node: Node, c: Context) -> pgast.Query | pgast.Statement:
             "TransactionStmt": _build_transaction_stmt,
             "PrepareStmt": _build_prepare,
             "ExecuteStmt": _build_execute,
+            "CreateStmt": _build_create,
         },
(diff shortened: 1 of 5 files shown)


edb/pgsql/parser/ast_builder.py
```diff
@@ -800,6 +869,14 @@ def _build_a_expr(n: Node, c: Context) -> pgast.BaseExpr:
             expr=_build_base_expr(n["rexpr"], c),
             context=_build_context(n, c),
         )
+    elif n['kind'] == 'AEXPR_NULLIF':
```
msullivan, line 872: Why is postgres like this

msullivan (approved):

edb/pgsql/parser/ast_builder.py
aljazerzen: I'm sure there are quite a lot non-handled cases still waiting to be discovered, but I just don't have the motivation to go and do them. There also just too, too much of them:

https://github.com/pganalyze/libpg_query/blob/15-latest/protobuf/pg_query.proto

I'll do them when they are needed for something.

PR #4911, opened 2023-01-20
SQL system columns

aljazerzen:
Support for this: https://www.postgresql.org/docs/14/ddl-system-columns.html

Needed for pg_dump & JDBC.


msullivan (approved):
Looks reasonable.

It won't be a problem that it only works for `ONLY`?

We could stick these things in the inhviews, if we had to.

aljazerzen:
Yes, these columns won't be available for inhview tables (without ONLY). But it's needed for pg_catalog anyways - at lease for now I don't think it's needed for all tables.

PR #4914, opened 2023-01-21
Implement nested modules

msullivan:
This implements the ABS proposal, discussed in #4883, but without use of
the `module` keyword.

It also doesn't reserve module names, which I'll do in a follow up.

Also postponed to a follow up: possibly moving a number of the existing
modules into the std:: namespace.

Also postponed to a follow up (that I will probably try to pawn off
onto @raddevon): documentation.

### DIFF
--- a/edb/edgeql/tracer.py
+++ b/edb/edgeql/tracer.py
@@ -306,10 +306,15 @@ def get_name(
     elif not module:
         module = current_module
     elif modaliases:
-        fq_module = modaliases.get(module)
+        if module:
+            first, sep, rest = module.partition('::')
+        else:
+            first, sep, rest = module, '', ''
+
+        fq_module = modaliases.get(first)
         if fq_module is not None:
             no_std = True
-            module = fq_module
+            module = fq_module + sep + rest
 
     qname = sn.QualName(module=module, name=ref.name)
     if type is None:
--- a/edb/edgeql/tracer.py
+++ b/edb/edgeql/tracer.py
@@ -21,7 +21,8 @@ from __future__ import annotations
 
 # Import specific things to avoid name clashes
 from typing import (Dict, FrozenSet, Generator, List, Mapping, Optional,
-                    Union, Set, Tuple, Iterable, Generic, TypeVar, Sequence)
+                    Union, Set, Tuple, Iterable, Generic, TypeVar, Sequence,
+                    AbstractSet)
 
 import functools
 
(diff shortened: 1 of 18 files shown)


tests/test_edgeql_ddl.py
```diff
@@ -5065,6 +5065,51 @@ async def test_edgeql_ddl_module_02(self):
...
+
+        await self.assert_query_result(
+            r'''
+            select foo::bar::Baz
+            ''',
+            []
+        )
+        await self.assert_query_result(
+            r'''
+            with module foo::bar
```
elprans, line 5107: Can we have some more tests for setting aliases at various depths (both via `with` and `set module`) and also some shadowing tests.

msullivan: Added more tests for setting aliases, fixed a couple bugs.

There's only really one shadowing behavior (shadowing a module under `std`) and it had a test in the first pass

edb/edgeql/tracer.py
```diff
@@ -281,11 +283,59 @@ def trace_refs(
         modaliases={},
         params=params,
         visited=set(),
+        local_modules=local_modules,
     )
     trace(qltree, ctx=ctx)
     return frozenset(ctx.refs)
 
 
+def get_name(
```
aljazerzen, line 292: Rename to `resolve_relative_name` or `lookup_relative_name`?

msullivan: I went with just `resolve_name`

msullivan:
I think we are good to go here; could I get a final review from @elprans or @aljazerzen?

elprans (approved):

tests/test_schema.py
```diff
@@ -746,6 +746,28 @@ def test_schema_duplicate_def_03(self):
...
+        errors.SchemaDefinitionError,
+        "module 'ext' is a reserved module name"
+    )
+    def test_schema_module_reserved_02(self):
+        """
+            module foo {
+                module ext {}
+            }
+        """
+
```
raddevon, line 770: We have a list of reserved module names with explanations of why they are reserved (see bottom of https://www.edgedb.com/docs/datamodel/index#terminology). These two are not currently on that list, so I'd like to add them. What are the reasons for `super` and `ext` being reserved? What do we use those for?

I'm guessing `super` lets us reference something in the parent, and I believe I recall mention of `ext` being for Postgres extensions? I may be confusing that with something else.

aljazerzen: `super` is reserved for future use, for reference to the parent module:

```
module hello {
   type Hello;

   module world {
       type World {
           link hello -> super::Hello;
       }
   }
}
```
(I don't think this is implemented yet, but it's a reasonable design)

`ext` is for extensions in general. And extensions in EdgeDB are not the same thing as Postgres extensions (as I initially believed).

PR #4915, opened 2023-01-21
Fix EdgeQL doc's number floordiv syntax

ShadiestGoat:
There is a small typo making arithmetic of the Numbers a bit harder to understand. The floordiv is written as a singular `/` rather than the correct `//`.


elprans (approved):
Thanks!

PR #4922, opened 2023-01-24
Fix Insert (Alternative Method)

zchenb:
Changed inserts to take a single object reference.

### DIFF
--- a/edb/edgeql/ast.py
+++ b/edb/edgeql/ast.py
@@ -573,8 +573,8 @@ class InternalGroupQuery(
     from_desugaring: bool = False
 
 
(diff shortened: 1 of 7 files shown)


zchenb:
I prefer the method of #4920, but the modified grammar will cause test failures. This is a workaround.

msullivan (approved):
This looks good!

What do you prefer about the other version? (Using `AtomicExpr` doesn't seem like the right thing)

zchenb:
This version may be better if AtomicExpr is not the right thing. Shall I go ahead and merge?

msullivan:
Go for it

PR #4924, opened 2023-01-25
[parser] Adjust associativity to make PtrQuals nonterminal more useful

msullivan:
By making MULTI/REQUIRED/OPTIONAL/SINGLE right associative, we can get
rid of an unfortunate bit in our parser where we need to inline the
body of PtrQuals.

The downside here is that this breaks some stuff: currently
`{required multi := 1}` works, with `multi` being the name of
the property, but this change breaks that.

### DIFF
--- a/edb/common/parsing.py
+++ b/edb/common/parsing.py
@@ -235,7 +235,7 @@ class PrecedenceMeta(type):
             doc = '%%%s' % assoc
 
             last = mcls.last.get((mcls, prec_group))
-            if last:
+            if last and rel_to_last:
                 doc += ' %s%s' % (rel_to_last, last.__name__)
 
             result.__doc__ = doc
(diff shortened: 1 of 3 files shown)


aljazerzen:
I think it is reasonable to say "`{required multi := 1}` was a bug are we are fixing it".

msullivan:
Even with this change, though, `{multi := 1}` works.

Like, it's weird, but `multi` is *not* a reserved keyword so it's hard to say that it is a *bug* that it can be used like this.
IDK.

(Neither are `optional` or `required` but `single` strangely is marked as "future reserved". I think maybe we should unreserve it unless we can think of some reason why it might get used differently than `multi`.)

aljazerzen (approved):

PR #4925, opened 2023-01-25
[parser] Implement an inliner for PtrQuals

msullivan:
There are currently places where the definition of PtrQuals gets
inlined into the definition of another nonterminal, because parsing it
in a nicely factored way requires 2 lookahead tokens and we are LR(1).

Implement an inliner for `PtrQuals` that will expands out a reference
to `InlinedPtrQuals` into all of the appropriate rules. This can be
used for the definition of `PtrQuals` itself and anywhere else.

This is a *competing* PR to #4924. It is more complex, but doesn't
break anything.

### DIFF
--- a/edb/edgeql/parser/grammar/expressions.py
+++ b/edb/edgeql/parser/grammar/expressions.py
@@ -20,6 +20,8 @@
 from __future__ import annotations
 
 import collections
+import itertools
+import types
 import typing
 
 from edb.common import parsing, context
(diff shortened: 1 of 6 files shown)


edb/edgeql/parser/grammar/expressions.py
```diff
@@ -549,135 +551,86 @@ class PtrQualsSpec(typing.NamedTuple):
...
+        ('MULTI', qltypes.SchemaCardinality.Many),
+        ('<e>', None),
+    ]
 
-    def reduce_MULTI(self, *kids):
-        self.val = PtrQualsSpec(cardinality=qltypes.SchemaCardinality.Many)
+    for name, func in list(cls.__dict__.items()):
+        if not (
+            name.startswith('reduce_') and isinstance(func, types.FunctionType)
+        ):
```
aljazerzen, line 572: I really don't think this is a good idea.

It is more concise, but it is also magic that takes a few minutes to understand.

aljazerzen (approved):
If you really want to merge this, please more documentation to `inline_ptr_quals`.

msullivan:
Part of the motivation here is that if we want to stay consistent in allowing things like `required multi` to mean a pointer named `multi` when we drop the requirement to always write `property`/`link` in SDL, that will require four more uses of inlined PtrQuals

edb/edgeql/parser/grammar/expressions.py
```diff
@@ -549,135 +551,83 @@ class PtrQualsSpec(typing.NamedTuple):
     cardinality: typing.Optional[qltypes.SchemaCardinality] = None
 
 
-class PtrQuals(Nonterm):
-    def reduce_OPTIONAL(self, *kids):
-        self.val = PtrQualsSpec(required=False)
+# Class decorator that inlines any references to InlinedPtrQuals in
+# a production.
```
elprans, line 555: Please add a comment _why_ this is necessary.

elprans (approved):

PR #4931, opened 2023-01-26
Use statement cache in EdgeQL script execution

fantix:
If prepared statements are not used in executing EdgeQL scripts (single API call with a query string that contains multiple EdgeQL queries separated by `;`), complex queries - even when executed frequently with proper EdgeQL compile cache - may constantly take dozens of milliseconds to run. This defect reduces the power of EdgeDB server with executing EdgeQL scripts, so this PR fixes it so that statement cache is leveraged when possible.

Also skip restoring identical state in the backend connection for EdgeQL script execution.

With this fix, the execution time is reduced to `6~7ms` in a sample test when fully warmed up, where it took `50~80ms` without this fix.


edb/server/pgcon/pgcon.pyx
```diff
@@ -824,14 +824,17 @@ cdef class PGConnection:
 
     cdef send_query_unit_group(
         self, object query_unit_group, object bind_datas, bytes state,
-        ssize_t start, ssize_t end,
+        ssize_t start, ssize_t end, int dbver, object parse_array
```
msullivan, line 827: Maybe add a comment that `parse_array` is really an *output*?

msullivan (approved):
:+1:

msullivan:
If you're confident in this, we can include it in 2.10?

fantix:
Yeah I think it's good, but let me quickly test on 2.x branch

fantix:
Test result says otherwise, let me see what's going wrong.

PR #4945, opened 2023-01-27
SQL current user & database

aljazerzen:
Support for `SELECT current_user, current_catalog`.

### DIFF
--- a/edb/server/compiler/compiler.py
+++ b/edb/server/compiler/compiler.py
@@ -632,7 +632,11 @@ class Compiler:
                     pass
                 else:
                     args['search_path'] = parse_search_path(search_path)
-                options = pg_resolver.Options(**args)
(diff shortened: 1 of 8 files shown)


edb/server/compiler/compiler.py
```diff
@@ -632,7 +632,11 @@ def parse_search_path(search_path_str: str) -> list[str]:
                     pass
                 else:
                     args['search_path'] = parse_search_path(search_path)
-                options = pg_resolver.Options(**args)
+                options = pg_resolver.Options(
+                    current_user='edgedb',
+                    current_database='edgedb',
+                    **args
+                )
```
aljazerzen, line 639: @fantix Can you hook in actual data here? Is this easy enough?

msullivan (approved):

PR #4946, opened 2023-01-27
Resolver GROUP BY alised column

aljazerzen:


tests/test_sql_query.py
```diff
@@ -504,6 +504,33 @@ async def test_sql_query_33(self):
         # this numbers change, so let's just check that there are 6 of them
         self.assertEqual(len(res[0]), 6)
 
+    async def test_sql_query_34(self):
+        # GROUP and ORDER BY aliased column
+
+        res = await self.squery_values(
+            """
+            SELECT substr(title, 2, 4) AS itl, count(*) FROM "Movie"
```
msullivan, line 512: :laughing:

tests/test_sql_query.py
```diff
@@ -504,6 +504,33 @@ async def test_sql_query_33(self):
...
+            GROUP BY itl
+            ORDER BY itl
+            """
+        )
+        self.assertEqual(res, [["avin", 1], ["orre", 1]])
+
+    async def test_sql_query_35(self):
```
msullivan, line 519: Do these tests actually test an ambiguous case?

msullivan (approved):
So much SQL

tests/test_sql_query.py
aljazerzen: No, it's not ambiguous, this all valid SQL. It's just that SQL has very particular rules on what names/aliases are allowed in which clauses. And that some tools rely on this behavior.

PR #4947, opened 2023-01-27
Fix SQL typecast to "char"

aljazerzen:
The problem is that this query:
```
SELECT 'a'::"char"
```
... get compiled to ...
```
SELECT 'a'::char
```

... which apparently is not the same.

The former is an ident that should resolve to pg_catalog.char, and the latter is a keyword (that is equivalent to pg_catalog.bpchar).

### DIFF
--- a/edb/pgsql/keywords.py
+++ b/edb/pgsql/keywords.py
@@ -68,7 +68,7 @@ pg_keywords = {
     "cast": ("CAST", RESERVED_KEYWORD),
     "catalog": ("CATALOG_P", UNRESERVED_KEYWORD),
     "chain": ("CHAIN", UNRESERVED_KEYWORD),
-    "char": ("CHAR_P", TYPE_FUNC_NAME_KEYWORD),
+    "char": ("CHAR_P", COL_NAME_KEYWORD),
     "character": ("CHARACTER", COL_NAME_KEYWORD),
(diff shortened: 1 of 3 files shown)


edb/pgsql/keywords.py
```diff
@@ -68,7 +68,7 @@
     "cast": ("CAST", RESERVED_KEYWORD),
     "catalog": ("CATALOG_P", UNRESERVED_KEYWORD),
     "chain": ("CHAIN", UNRESERVED_KEYWORD),
-    "char": ("CHAR_P", COL_NAME_KEYWORD),
+    "char": ("CHAR_P", TYPE_FUNC_NAME_KEYWORD),
```
elprans, line 71: I'm not quite sure this is the best fix. Maybe we should special-case `"char"` in codegen instead?

aljazerzen: Maybe here?

https://github.com/edgedb/edgedb/blob/master/edb/pgsql/common.py#L105-L131

elprans: No, I think none of that actually solves the confusion.  IMO, the most reliable way would be to translate unquoted `char` to `bpchar` at the parser level and always quote `"char"` in codegen.  In other words, use Postgres names.

elprans (approved):

PR #4957, opened 2023-01-31
pg-ext: bug fixes

fantix:
* Don't capture fatal errors in pg-ext protocol
* Support SQL_ASCII client encoding and error out on invalid encodings
* Handle client_encoding correctly
* Remember the last_state of pg-ext
* Forward ParameterStatus in simple query

### DIFF
--- a/edb/server/pgcon/pgcon.pyx
+++ b/edb/server/pgcon/pgcon.pyx
@@ -103,6 +103,7 @@ cdef dict POSTGRES_SHUTDOWN_ERR_CODES = {
 }
 
 cdef bytes INIT_CON_SCRIPT = None
+cdef object EMPTY_SQL_STATE = json.dumps({}).encode('utf-8')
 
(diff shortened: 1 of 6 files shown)


edb/server/pgcon/pgcon.pyx
```diff
@@ -103,6 +103,7 @@ cdef dict POSTGRES_SHUTDOWN_ERR_CODES = {
 }
 
 cdef bytes INIT_CON_SCRIPT = None
+cdef object EMPTY_SQL_STATE = json.dumps({}).encode('utf-8')
```
1st1, line 106:
```suggestion
cdef object EMPTY_SQL_STATE = "{}".encode('utf-8')
```

? I admire the readability of this but it does make the startup slower, however minuscule that would be

1st1 (approved):
Only did a cursory look at this PR but everything looked fine.

Issue #4441, opened 2022-09-28
Document path factoring behavior


msullivan:
In EdgeQL, common paths are "factored" to the closest surrounding subquery.

This is what allows
```
select User.first_name ++ ' ' ++ User.last_name
```
to return one row per user, as well as causing
```
select (User.name, count(User))
```
to always report `1` as the count, as well as letting
```
select (User.name, count(User.friends))
```
return each user's number of friends.

I don't believe that this is currently documented anywhere but https://www.edgedb.com/docs/reference/edgeql/eval, which does not explain it at all, it just shows how to define it.

vpetrovykh:
The PR #4959 (mentioned above) adds a section on path resolution to the docs.

PR #4959, opened 2023-01-31
Document path resolution

raddevon:
Addresses #4441

### DIFF
--- a/docs/edgeql/path_resolution.rst
+++ b/docs/edgeql/path_resolution.rst
@@ -34,7 +34,7 @@ using cartesian multiplication.
     {'Mina Murray', 'Jonathan Harker', 'Lucy Westenra', 'John Seward'}
 
 We assume this is what you want, but if your goal is to get the cartesian
-product, you can accomplish it one of two ways. You could use
+product, you can accomplish it one of three ways. You could use
 :eql:op:`detached`.
 
 .. code-block:: edgeql-repl
@@ -59,7 +59,7 @@ product, you can accomplish it one of two ways. You could use
       'John Seward',
     }
 
-Or you could use :ref:`with <ref_eql_with>` to attach a different symbol to
+You could use :ref:`with <ref_eql_with>` to attach a different symbol to
 your set of ``User`` objects.
 
 .. code-block:: edgeql-repl
@@ -85,6 +85,9 @@ your set of ``User`` objects.
       'John Seward',
     }
 
+Or you could leverage the effect scopes have on path resolution. More on that
+:ref:`in the Scopes section <ref_eql_path_resolution_scopes>`.
+
 The reason ``with`` works here even though the alias ``U`` refers to the exact
 same set is that we only assume you want the path factored in this way when you
 use the same *symbol* to refer to a set. This means operations with
@@ -96,6 +99,8 @@ That may leave you still wondering why ``U`` and ``User`` did not get a common
 path factored. ``U`` is just an alias of ``select User`` and ``User`` is the
 same symbol that we use in our name query. That's true, but EdgeDB doesn't
 factor in this case because of the queries' scopes.
+
+.. _ref_eql_path_resolution_scopes:
 
 Scopes
 ------
--- a/docs/edgeql/path_resolution.rst
+++ b/docs/edgeql/path_resolution.rst
@@ -105,7 +105,7 @@ factor in this case because of the queries' scopes.
 Scopes
 ------
 
-Scopes change the was path resolution works. Two sibling select queries — that
+Scopes change the way path resolution works. Two sibling select queries — that
 is, queries at the same level — do not have their paths factored even when they
 use a common symbol.
 
--- a/docs/edgeql/path_resolution.rst
+++ b/docs/edgeql/path_resolution.rst
@@ -1,0 +1,151 @@
+.. _ref_eql_path_resolution:
+
+===============
+Path resolution
+===============
+
+Element-wise operations with multiple arguments in EdgeDB are generally applied
+to the :ref:`cartesian product <ref_reference_cardinality_cartesian>` of all
+the input sets.
+
+.. code-block:: edgeql-repl
+
+    db> select {'aaa', 'bbb'} ++ {'ccc', 'ddd'};
+    {'aaaccc', 'aaaddd', 'bbbccc', 'bbbddd'}
+
+In some cases, this works out fine, but in others it doesn't make sense. Take
+this example:
+
+.. code-block:: edgeql
+
+    select User.first_name ++ ' ' ++ User.last_name;
+
+Should the result of this query be every ``first_name`` in your ``User``
+objects concatenated with every ``last_name``? That's probably not what you
+want.
+
+This is why, in cases where multiple element-wise arguments share a common path
+(``User.`` in this example), EdgeDB factors out the common path rather than
+using cartesian multiplication.
+
+.. code-block:: edgeql-repl
+
+    db> select User.first_name ++ ' ' ++ User.last_name;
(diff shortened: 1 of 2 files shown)


docs/edgeql/path_resolution.rst
```diff
@@ -0,0 +1,151 @@
...
+      'John Westenra',
+      'John Seward',
+    }
+
+The reason ``with`` works here even though the alias ``U`` refers to the exact
+same set is that we only assume you want the path factored in this way when you
+use the same *symbol* to refer to a set. This means operations with
+``User.first_name`` and ``User.last_name`` do get the common path factored
+while ``U.first_name`` and ``User.last_name`` do not and are resolved with
+cartesian multiplication.
```
zchenb, line 93: I am not sure but it seems that there are better explanation, that the clauses of with and the body of with live in different scopes (same reason as for the next section). Maybe Sully can confirm?

zchenb (commented):
Looks fine to me as an introductory explanation for path resolution except the commented line. The text did not mention the interaction with subqueries and semi-subqueries which are more advanced topics.

docs/edgeql/path_resolution.rst
zchenb: The technical reason is that `U` in `U.first_name` is really a binder that has not been instantiated. Putting it on a par with `User.last_name` seems strange.

raddevon: Any ideas for a better way to explain this @vpetrovykh or @msullivan? This was my takeaway from my conversation with Victor. I may have misunderstood.

docs/edgeql/path_resolution.rst
```diff
@@ -0,0 +1,151 @@
...
+using cartesian multiplication.
+
+.. code-block:: edgeql-repl
+
+    db> select User.first_name ++ ' ' ++ User.last_name;
+    {'Mina Murray', 'Jonathan Harker', 'Lucy Westenra', 'John Seward'}
+
+We assume this is what you want, but if your goal is to get the cartesian
+product, you can accomplish it one of two ways. You could use
+:eql:op:`detached`.
```
msullivan, line 38: There's a third way, too, which is to put both parts in a subquery like in the "Scopes" section

docs/edgeql/path_resolution.rst
msullivan: Both of these reasons are correct and are part of what is going on:
 * `U.first_name` and `User.last_name` don't have anything factored out because they don't have a common prefix
 * The `User` in the `with` binding and the `User` in the `select` body don't get factored because they are in separate subqueries

docs/edgeql/path_resolution.rst
```diff
@@ -0,0 +1,151 @@
...
+      ('Lucy', 'Harker'),
+      ('Lucy', 'Westenra'),
+      ('Lucy', 'Seward'),
+      ('John', 'Murray'),
+      ('John', 'Harker'),
+      ('John', 'Westenra'),
+      ('John', 'Seward'),
+    }
+
+Common symbols in nested scopes *are* factored when they use the same symbol.
```
msullivan, line 124: I think another good example here is `select (User.name, count(User.friends))`, which returns each user's name and the number of friends they have.

docs/edgeql/path_resolution.rst
raddevon: Makes sense. I will update it to reflect that. That will give me a natural segue into scopes too. Thanks @zchenb and @msullivan!

raddevon:
I believe I have addressed all of the feedback in this latest push. @msullivan and/or @zchenb, would you care to have another look when you have time?

docs/edgeql/path_resolution.rst
```diff
@@ -0,0 +1,177 @@
...
+path factored. ``U`` is just an alias of ``select User`` and ``User`` is the
+same symbol that we use in our name query. That's true, but EdgeDB doesn't
+factor in this case because of the queries' scopes.
+
+.. _ref_eql_path_resolution_scopes:
+
+Scopes
+------
+
+Scopes change the was path resolution works. Two sibling select queries — that
```
zchenb, line 108: `Scopes change the was path resolution works.` Not sure how to read this sentence. Could you clarify?

raddevon: That's a typo. Thanks for catching it! Just pushed a fix. Should be `way` instead of `was`.

zchenb (approved):

docs/edgeql/path_resolution.rst
```diff
@@ -0,0 +1,177 @@
...
+      'John Westenra',
+      'John Seward',
+    }
+
+You could use :ref:`with <ref_eql_with>` to attach a different symbol to
+your set of ``User`` objects.
+
+.. code-block:: edgeql-repl
+
+    edgedb> with U := (select User)
```
vpetrovykh, line 67:
```suggestion
    edgedb> with U := User
```
You don't need the `select` for aliasing simple expressions. You usually need `select` if you want to use some of the other clauses (`filter` or `order`, etc.).

docs/edgeql/path_resolution.rst
```diff
@@ -0,0 +1,177 @@
...
+    {('Fran', 3), ('Bam', 2), ('Emma', 3), ('Geoff', 1), ('Tyra', 1)}
+
+In this example, ``count``, like all aggregate function, creates a nested
+scope, but this doesn't prevent the paths from being factored as you can see
+from the results. If the paths were *not* factored, the friend count would be
+the same for all the result tuples and it would reflect the total number of
+``Person`` objects that are in *all* ``friends`` links rather than the number
+of ``Person`` objects that are in the named ``Person`` object's ``friends``
+link.
+
```
vpetrovykh, line 166: It's possibly worthwhile contrasting this with an example where both parts of that tuple are wrapped into some aggregate function:
```
    edgedb> select (array_agg(distinct Person.name), count(Person.friends));
    {(['Fran', 'Bam', 'Emma', 'Geoff'], 3)}
```
Here `Person` is not factored out because the scopes of aggregate function arguments are _sibling_ scopes, much like the two `select` sub-queries from earlier. So the interpretation is "get an array of all unique names" and "get the total number of people who are somebody's friend".

docs/edgeql/path_resolution.rst
```diff
@@ -0,0 +1,177 @@
...
+Clauses & Nesting
+^^^^^^^^^^^^^^^^^
+
+Most clauses are nested and are subjected to the same rules described above:
+common symbols are factored and assumed to refer to the same object as the
+outer query. This is because clauses like :ref:`filter
+<ref_eql_select_filter>` and :ref:`order by <ref_eql_select_order>` need to
+be applied to each row.
```
vpetrovykh, line 174:
```suggestion
be applied to each value in the result.
```
We generally avoid using "row" or "column" in our explanations as we don't actually have them.

docs/edgeql/path_resolution.rst
```diff
@@ -0,0 +1,177 @@
...
+
+The :ref:`limit <ref_eql_select_pagination>` clause is not nested in the scope
+because it needs to be applied globally to your query.
```
vpetrovykh, line 177:
```suggestion
because it needs to be applied globally to the entire result set of your query.
```

vpetrovykh (changes requested):

docs/edgeql/path_resolution.rst
raddevon: Good suggestion. I added this (nearly verbatim).

vpetrovykh (approved):
LGTM

msullivan (approved):

Issue #4787, opened 2022-12-07
Undocumented config options


jaclarke:
The config options `allow_bare_ddl` and `apply_access_policies` appear to not be documented anywhere. They probably should be documented in `stdlib/cfg` (and maybe `reference/configuration`?).

PR #4963, opened 2023-01-31
Document `allow_bare_ddl`, `apply_access_policies`

raddevon:
I have a technical issue here. The `` :ref:`configuration parameter <ref_std_cfg>` `` link anchors to nearly the end of the config page rather than the top where that anchor is. I've asked about this issue on Slack. I will fix it once I get that figured out.

The content should be ready for review though.

Addresses #4787

### DIFF
--- a/docs/datamodel/access_policies.rst
+++ b/docs/datamodel/access_policies.rst
@@ -322,6 +322,19 @@ making the current user able to see their own ``User`` record.
   following to the schema: ``using future nonrecursive_access_policies;``
 
 
+Disabling policies
+^^^^^^^^^^^^^^^^^^
+
+You may disable all access policies by setting the ``apply_access_policies``
+:ref:`configuration parameter <ref_std_cfg>` to ``false``.
+
+You may also toggle access policies using the "Disable Access Policies"
+checkbox in the "Config" dropdown in the EdgeDB UI (accessible by running
+the CLI command ``edgedb ui`` from inside your project). This is the most
+convenient way to temporarily disable access policies since it applies only to
+your UI session.
+
+
 Examples
 ^^^^^^^^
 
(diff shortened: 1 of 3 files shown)


docs/datamodel/access_policies.rst
```diff
@@ -322,6 +322,19 @@ making the current user able to see their own ``User`` record.
...
+^^^^^^^^^^^^^^^^^^
+
+You may disable all access policies by setting the ``apply_access_policies``
+:ref:`configuration parameter <ref_std_cfg>` to ``false``.
+
+You may also toggle access policies using the "Disable Access Policies"
+checkbox in the "Config" dropdown in the EdgeDB UI (accessible by running
+the CLI command ``edgedb ui`` from inside your project). This is the most
+convenient way to temporarily disable access policies since it applies only to
+your UI session.
```
1st1, line 335: Do we also have a flag for this in our CLI REPL?

docs/stdlib/cfg.rst
```diff
@@ -82,6 +82,34 @@ Query planning
...
+  When you create an instance, this is set to ``cfg::AllowBareDDL.AlwaysAllow``
+  until you run a migration. At that point it is set to
+  ``cfg::AllowBareDDL.NeverAllow`` because it's generally a bad idea to mix
+  migrations with bare DDL.
+
+.. _ref_std_cfg_apply_access_policies:
+
+:eql:synopsis:`apply_access_policies -> bool`
+  Determines whether access policies should be applied when running queries.
+  Setting this to ``false`` effctively puts you into super-user mode, ignoring
```
1st1, line 101:
```suggestion
  Setting this to ``false`` effectively puts you into super-user mode, ignoring
```

1st1 (approved):
Looks good aside a couple nits highlighted in the comments

docs/datamodel/access_policies.rst
tailhook: We don't have a setting for that. But you can issue a regular query to tune configuration option. I think this is "dangeorous don't use unless you know what you're doing" thing, so maybe query is okay?

Issue #4936, opened 2023-01-26
Implement triggers RFC


msullivan:
https://github.com/edgedb/rfcs/blob/master/text/1020-triggers.rst

`AFTER COMMIT` we will leave off at first.

PR #4964, opened 2023-02-01
Initial work on triggers

msullivan:
The implements parsing and schema storage of triggers (but doesn't
actually implement the *behavior*).

Progress on #4936.


aljazerzen (approved):
I've reviewed most of this because I'm plagiarizing it for rewrites, but I didn't look into the tracer-context-passing changes.

PR #4965, opened 2023-02-01
Parse SQL GREATEST, LEAST, sublink array, IS DISTINCT FROM

aljazerzen:


aljazerzen:
That's one of the motivations for PRQL :D

PR #4974, opened 2023-02-02
Implement manual error override configuration

### LINKED ISSUE #4625: Manual error status for EdgeDB server
# Motivation

`edgedb watch` command is expected to be running in the background and update schema in the database. Sometimes schema updates cannot be applied because of syntax of schema files is wrong or because database changes are conflicting. In those cases we have to notify users somehow that schema changes could not be applied.

It's hard to make notifications from the background process, so the simplest solution is to error out all clients. So all subsequent testing that user does fails with the same error until schema files are fixed.

# Overview

We use configuration mechanism to propagate error to all the clients that use the same database.

Example on conflicting type:
```
CONFIGURE DATABASE SET database_failure := "edgedb watch: SchemaError: property 'value' of object type 'default::Bar' cannot be cast automatically from scalar type 'std::int32' to scalar type 'std::str. Please fix syntax error before proceeding."
```
Example syntax error:
```
CONFIGURE DATABASE SET database_error := "edgedb watch: Error reading schema files. Please fix syntax error before proceeding."
CONFIGURE DATABASE SET database_error_info := '{"file": "dbschema/default.esdl", "line": 10, "column": 16, "message": "Syntax error: unexpected `)`"}'
```
Example when database upgrade needed:
```
CONFIGURE DATABASE SET database_warning := "edgedb watch: current EdgeDB version is 2.3+deadbee while minimum of 2.7 required by edgedb.toml"
```

# Specification

New error type must be introduced. `DevModeError` (better name?), probably a subclass of `BackendError` (or `ConfigurationError`?)

Config parameters:

* `database_error` -- this error is thrown in every client when any non-configuration query is performed (i.e. any CONFIGURE is allowed as well as selecting from `cfg::` and `sys::get_version*`, maybe all stdlib is allowed?)
* `database_error_info` -- JSON with extra information (format to be determined). This can be used in UI-based tools to provide nicer error. Has to be fetched manually via `SELECT cfg::Config`.
* `database_warning` -- sent as a warning to all the clients (when we start supporting warning mechanism)

This mechanism is error by default, but generally advisory. To opt-out of error you can do:
```
CONFIGURE SESSION SET database_error := {};
```
This mechanism will be used by CLI for two things:
1. For `edgedb watch` / `edgedb migrate` to fix the schema after resetting changes
2. `edgedb` (REPL) will print the failure and clean `database_error` for the session, so you can inspect things while error is there.

msullivan:
Implement a configuration field that causes the server to return an
error for all non-configuration commands.
See #4625 for details and motivation. Closes #4625.

The details here are a little different than proposed there.
There is one config property, `force_database_error`, that
accepts a json encoded string with the following format:
`false` means no error, and an object means to raise an error:
```
{
  "type": <name of exception class>,
  "message": <exception message>,
  "hint": <exception hint>,
  "details": <exception details>,
  "context": {
    "line": <line number>,
    "col": <column number>,
    "start": <start position>,
    "end": <end position>,
    "filename": <file name for error>
  }
}
```
All fields except for `type` are optional.

### DIFF
--- a/edb/server/compiler/compiler.py
+++ b/edb/server/compiler/compiler.py
@@ -1807,6 +1807,9 @@ def _try_compile(
     for i, stmt in enumerate(statements):
         is_trailing_stmt = i == statements_len - 1
         stmt_ctx = ctx if is_trailing_stmt else non_trailing_ctx
+
+        _check_force_database_error(stmt_ctx, stmt)
+
         comp, capabilities = _compile_dispatch_ql(
             stmt_ctx,
             stmt,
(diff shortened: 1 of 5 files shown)


edb/server/compiler/compiler.py
```diff
@@ -2328,6 +2331,42 @@ def _get_data_mending_desc(
...
 
+def _check_force_database_error(
+    ctx: CompileContext,
+    ql: qlast.Base,
+) -> None:
+    if isinstance(ql, qlast.ConfigOp):
+        return
+
+    try:
+        err = json.loads(_get_config_val(ctx, 'force_database_error'))
```
aljazerzen, line 2342: Will this deserialize JSON on every compile?

It's not that show, just unnecessary.

aljazerzen: And `json.loads` throws an error if it gets None

tailhook (approved):

PR #4978, opened 2023-02-03
Optimize trivial WITH-bound `group` uses

### LINKED ISSUE #4243: Better optimization for with-bound groups
Currently we have optimizations that will sink shapes and for loop bodies into the inside of our internal group statement.
These optimizations operate on the edgeql AST, not on IR.

A big problem here is that as a result, these optimizations don't apply when the `group` is bound with `with`, which seems a fairly natural thing to want to do. Also, unfortunately, the non-optimized path is buggier than the optimized path (since it leans much more heavily on materialization). See https://github.com/edgedb/edgedb/issues/4240, https://github.com/edgedb/edgedb/issues/4241.

-----
I haven't thought about this deeply yet, but the options I have thought of so far are:
1. Do some basic analysis to see how many times a with-bound group is used, and inline it when it is only one. This would require some care since the scope might be different at the in-lining point. We can probably work around that in either a principled way or a hacky hyper-local way.
2. Change the optimizations to operate at the IR level. Probably actually this still requires doing analysis to count the number of uses

msullivan:
If the last alias in a WITH block is a GROUP query that is immediately
used in a shape or a for loop, inline it so that our other
optimizations apply.

This probably covers 95% of WITH-bound group use cases, despite being
a "hyper-local hack", as I said in #4243.

Fixes #4243. (Mostly.)

### DIFF
--- a/edb/edgeql/desugar_group.py
+++ b/edb/edgeql/desugar_group.py
@@ -218,8 +218,9 @@ def try_group_rewrite(
     # immediately used, so that we can apply the other optimizations.
     match node:
         case qlast.SelectQuery(
-            aliases=[*_, qlast.AliasedExpr(
-                alias=alias, expr=qlast.GroupQuery() as grp)
+            aliases=[
+                *_,
+                qlast.AliasedExpr(alias=alias, expr=qlast.GroupQuery() as grp)
             ] as qaliases,
             result=qlast.Shape(
                 expr=qlast.Path(steps=[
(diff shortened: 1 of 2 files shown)


edb/edgeql/desugar_group.py
```diff
@@ -199,7 +214,38 @@ def try_group_rewrite(
...
     """
 
-    # TODO: would Python's new pattern matching fit well here???
+    # Inline trivial uses of aliases bound to a group and then
+    # immediately used, so that we can apply the other optimizations.
+    match node:
+        case qlast.SelectQuery(
+            aliases=[*_, qlast.AliasedExpr(
+                alias=alias, expr=qlast.GroupQuery() as grp)
+            ] as qaliases,
```
elprans, line 223: Perhaps better formatted as

```suggestion
            aliases=[
                *_, 
                qlast.AliasedExpr(
                  alias=alias, 
                  expr=qlast.GroupQuery() as grp
                )
            ] as qaliases,
```

?

elprans (approved):

PR #4979, opened 2023-02-03
Implement a scheme for using "view patterns" with pattern matching

msullivan:
There are a lot of details in the docstring, but the gist of it is
that this will allow us to implement pattern matching "helper
functions", which can make it a lot more pleasant to match on
our often complex AST constructs.


msullivan:
... I promise to remove at least 3 other metaclasses from the codebase to make up for this

aljazerzen (approved):
Sigh.

PR #4981, opened 2023-02-03
rewrites schema

aljazerzen:
- Parse mutation rewrites
- fix
- Add rewrites to schema

### DIFF
--- a/edb/schema/rewrites.py
+++ b/edb/schema/rewrites.py
@@ -1,0 +1,301 @@
+#
+# This source file is part of the EdgeDB open source project.
+#
+# Copyright 2008-present MagicStack Inc. and the EdgeDB authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+from __future__ import annotations
+from typing import *
+
+from edb import errors
+
+from edb.edgeql import ast as qlast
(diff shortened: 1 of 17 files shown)


edb/schema/rewrites.py
```diff
@@ -0,0 +1,301 @@
...
+
+        assert isinstance(astnode, qlast.RewriteCommand)
+
+        for kind in astnode.kinds:
+            cmd = super()._cmd_tree_from_ast(schema, astnode, context)
+            assert isinstance(cmd, RewriteCommand)
+
+            cmd.set_attribute_value('kind', kind)
+            group.add(cmd)
+        return group
```
aljazerzen, line 215: @msullivan This is the part I need you to checkout.

edb/schema/rewrites.py
```diff
@@ -0,0 +1,301 @@
...
+            # in_ddl_context_name = pol_name
+
+            kind = self._get_kind(schema)
+
+            anchors = {}
+            if qltypes.RewriteKind.Update == kind:
+                anchors['__old__'] = pathid.PathId.from_type(
+                    schema,
+                    source,
+                    typename=sn.QualName(module='__derived__', name='__old__'),
```
msullivan, line 141: Oh this reminds me I think `__old__ was left out of the anchors in the tracer

edb/schema/rewrites.py
```diff
@@ -0,0 +1,301 @@
...
+            return s_expr.Expression(text='false')
+        else:
+            raise NotImplementedError(f'unhandled field {field.name!r}')
+
+    def validate_object(
+        self,
+        schema: s_schema.Schema,
+        context: sd.CommandContext,
+    ) -> None:
+        # XXX: verify we don't have the same bug as access policies
```
msullivan, line 192: Oh yeah we'll need to think about linkprop interactions. Probably these won't work on linkprops unless we change a bunch about how we compile them

edb/schema/rewrites.py
msullivan: Yeah, this seems like a good way to handle it, but I'd add a comment explaining it both here and in CreateRewrite

msullivan (approved):
Looks good!

Issue #4942, opened 2023-01-26
SDL syntax simplifications from RFC 1022


msullivan:
https://github.com/edgedb/rfcs/blob/master/text/1022-freetypes.rst

There is still some discussion needed about how exactly things will be normalized.

PR #4983, opened 2023-02-03
Implement the SDL syntax changes from RFC 1022

msullivan:
The two main things are:
 * Allow : instead of -> for pointers
 * Allow omitting "link"/"property" in SDL for non-computed pointers
   when the target type is explicitly specified.

We do the link/property determination in a pretty simple-minded way
when resolving SDL to DDL, so DDL still has no notion of having an
unspecified pointer kind.

"Mostly" finishes #4942. There's more to be done if we
want to normalize to the new colon syntax.

I updated some of the tests schemas. People should check them and make
sure they are happy with them.


tests/schemas/issues.esdl
```diff
@@ -61,68 +61,68 @@ type LogEntry extending Owned, Text {
...
+    priority: Priority;
 
-    optional multi link watchers -> User;
+    optional multi watchers: User;
 
-    optional property time_estimate -> int64;
+    optional time_estimate: int64;
 
-    multi link time_spent_log -> LogEntry;
+    multi time_spent_log: LogEntry;
```
aljazerzen, line 91: This looks so much better. `link` & `property` add quite a lot of noise.

aljazerzen (approved):

1st1 (approved):

PR #4991, opened 2023-02-07
Fix SQL introspection for nested modules

aljazerzen:


msullivan (approved):

edb/pgsql/metaschema.py
```diff
@@ -5039,18 +5039,23 @@ def _generate_sql_information_schema() -> List[dbops.Command]:
...
             SELECT
                 id,
-                CASE SPLIT_PART(name, '::', 1)
-                    WHEN 'default' THEN 'public'
-                    ELSE SPLIT_PART(name, '::', 1)
-                END AS schema_name,
-                SPLIT_PART(name, '::', 1) AS module_name,
-                SPLIT_PART(name, '::', 2) AS table_name
+                REGEXP_REPLACE(name, '::[^:]*$', '') AS module_name,
+                SPLIT_PART(name, '::', -1) AS table_name
```
fantix, line 5046: Unfortunately `SPLIT_PART(...., -1)` is not yet supported in [Postgres 13](https://www.postgresql.org/docs/13/functions-string.html#id-1.5.8.10.7.2.2.31.1.1.1), but looks like this is the last one [failing on 13](https://github.com/edgedb/edgedb/actions/runs/4258973631/jobs/7410766473).

PR #4998, opened 2023-02-09
Fix custom function calls on the HTTP interface

### LINKED ISSUE #4989: custom function wasn't supported via HTTP
- EdgeDB Version: V2.9
- EdgeDB CLI Version:
- OS Version: WSL in widows 10

Steps to Reproduce:

1. run the query `select SellingOrder {computedPurchaseDateTime }` via repl will return the correct result.
2. run the query `select SellingOrder {computedPurchaseDateTime }` via http by insomnia will result null.

image

image

Schema:
```
type SellingOrder {
		property purchaseDateTime -> datetime;
		property computedPurchaseDateTime := get_fdow(.purchaseDateTime);
	}

	function get_fdow(dt: datetime) -> str # get the first day of specified week
        	using (
            	with pur_date := cal::to_local_datetime(dt,"America/Los_Angeles"),
            	weeks := datetime_get(pur_date, "isodow"),
            	first_date_of_week := pur_date - <duration>(<str>((weeks-1)*24*60*60*1000) ++ " milliseconds"),
            	week := str_split(<str>first_date_of_week, "T")[0],
            	select week
        );
```

msullivan:
The problem was that, if no global variables were passed,
we were passing in NULL for the global variable JSON object,
which causes strict functions to spuriously return NULL.

Fixes #4989

### DIFF
--- a/edb/server/protocol/edgeql_ext.pyx
+++ b/edb/server/protocol/edgeql_ext.pyx
@@ -125,7 +125,7 @@ async def handle_request(
             db,
             query,
             variables=variables or {},
-            globals_=globals_ or {},
+            globals_=globals_,
         )
     except Exception as ex:
         if debug.flags.server:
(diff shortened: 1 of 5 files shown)


fantix (approved):
LGTM

PR #5000, opened 2023-02-09
Fix quote_string function to quote backslash

tailhook:
This function is mostly used in the CLI for quoting configuration values. We haven't seen too many complex configuration values so far, so this was unnoticed.


msullivan (approved):
good catch

PR #5006, opened 2023-02-09
Implement rewrites

aljazerzen:
- [x] add to schema without inheritance
- [x] add to schema with inheritance and overriding
- [x] ql compile `__specfied__` 
- [x] ql compile `__subject__ `
- [x] ql compile `__old__` 
- [x] ql compile with partial path prefix (almost works)
- [x] pg compile Insert (works by just changing shape on ql side)
- [x] find all rewrites for child ObjectTypes and store them in `irast.UpdateStmt.rewrites`
- [x] pg compile Update
- [x] access policy interaction
- [x] volatile fields in updates (... also broken with access policies!)
- [x] chaining + recursion check
- [x] references to multi fields in `update` (probably needs some of the same stuff triggers use)

### DIFF
--- a/edb/edgeql/compiler/viewgen.py
+++ b/edb/edgeql/compiler/viewgen.py
@@ -35,6 +35,7 @@ from edb.common.typeutils import downcast, not_none
 from edb.ir import ast as irast
 from edb.ir import typeutils
 from edb.ir import utils as irutils
+import edb.ir.typeutils as irtypeutils
 
 from edb.schema import links as s_links
 from edb.schema import name as sn
@@ -844,8 +845,6 @@ def _compile_rewrites(
     s_ctx: ShapeContext,
     ctx: context.ContextLevel,
 ) -> irast.Rewrites:
-    import edb.ir.typeutils as irtypeutils
-
     # init
     anchors = prepare_rewrite_anchors(
         specified_ptrs, kind, stype, ctx
@@ -1014,10 +1013,9 @@ def _compile_rewrites_for_stype(
 
         subject_set, specified_set, old_set = anchors
 
-        # ???
         rewrite_view = view_scls
         if rewrite_source != view_scls.get_nearest_non_derived_parent(schema):
-            # XXX: CREAM
+            # FIXME: Caching?
             rewrite_view = downcast(
                 s_objtypes.ObjectType,
                 schemactx.derive_view(
@@ -1069,12 +1067,7 @@ def _compile_rewrites_for_stype(
             shape_ql = qlast.ShapeElement(
                 expr=qlast.Path(
                     steps=[
-                        qlast.Ptr(
-                            ptr=qlast.ObjectRef(
-                                name=ptrcls_sn.name,
-                                module=ptrcls_sn.module,
-                            ),
-                        ),
+                        qlast.Ptr(ptr=s_utils.name_to_ast_ref(ptrcls_sn)),
                     ],
                 ),
                 compexpr=qlast.DetachedExpr(
@@ -1117,6 +1110,7 @@ def prepare_rewrite_anchors(
     schema = ctx.env.schema
 
     # init set for __subject__
+    # TODO: Do we really need a separate path id for __subject__?
     subject_name = sn.QualName("__derived__", "__subject__")
     subject_path_id = irast.PathId.from_type(
         schema, stype, typename=subject_name, namespace=ctx.path_id_namespace
@@ -1148,7 +1142,7 @@ def prepare_rewrite_anchors(
                 name=pn.name,
                 val=setgen.ensure_set(
                     irast.BooleanConstant(
-                        value="true" if pn in specified_ptrs else "false",
+                        value=str(pn in specified_ptrs),
                         typeref=bool_path.target,
                     ),
                     ctx=ctx
--- a/edb/edgeql/compiler/viewgen.py
+++ b/edb/edgeql/compiler/viewgen.py
@@ -890,11 +890,11 @@ def _compile_rewrites(
             path_id = irast.PathId.from_pointer(schema, element.ptrcls)
 
             # construct a new set with correct path_id
-            ptr_set = irast.Set(
+            ptr_set = setgen.new_set_from_set(
+                target,
                 path_id=path_id,
-                typeref=target.typeref,
-                expr=target.expr,
-                rptr=None
+                rptr=None,
+                ctx=ctx,
             )
 
             # construct a new set with correct path_id
@@ -1032,18 +1032,28 @@ def _compile_rewrites_for_stype(
         rewrite_expr = rewrite.get_expr(ctx.env.schema)
         assert rewrite_expr
 
-        with ctx.new() as scopectx:
+        with ctx.newscope(fenced=True) as scopectx:
             # prepare context
             scopectx.partial_path_prefix = subject_set
-            scopectx.anchors["__specified__"] = specified_set
-            scopectx.anchors["__subject__"] = subject_set
+            nanchors = {}
+            nanchors["__specified__"] = specified_set
+            nanchors["__subject__"] = subject_set
             if old_set:
-                scopectx.anchors["__old__"] = old_set
+                nanchors["__old__"] = old_set
 
-            scopectx.env.path_scope.attach_path(
-                subject_set.path_id, context=None
-            )
+            for key, anchor in nanchors.items():
+                scopectx.path_scope.attach_path(
+                    anchor.path_id, context=None,
+                    optional=(anchor is subject_set),
+                )
+                scopectx.iterator_path_ids |= {anchor.path_id}
+                scopectx.anchors[key] = anchor
+
+            # XXX: I am pretty sure this must be wrong, but we get
+            # a failure without due to volatility issues in
+            # test_edgeql_rewrites_16
             scopectx.env.singletons.append(subject_set.path_id)
+
             ctx.path_scope.factoring_allowlist.add(subject_set.path_id)
 
             # prepare expression
@@ -1150,6 +1160,7 @@ def prepare_rewrite_anchors(
         old_set = setgen.new_set(
             stype=stype, path_id=old_path_id, ctx=ctx
         )
+        old_set.expr = irast.TriggerAnchor(typeref=old_set.typeref)
     else:
         old_set = None
 
--- a/edb/edgeql/compiler/viewgen.py
+++ b/edb/edgeql/compiler/viewgen.py
@@ -1073,7 +1073,7 @@ def _compile_rewrites_for_stype(
 
             # compile as normal shape element
             pointer, ptr_set = _normalize_view_ptr_expr(
-                ir_set,
+                subject_set,
                 shape_ql_desc,
                 rewrite_view,
                 path_id=path_id,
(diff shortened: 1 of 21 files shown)


edb/edgeql/compiler/viewgen.py
```diff
@@ -446,6 +473,133 @@ def gen_pointers_from_defaults(
...
+    bool_path = irast.PathId.from_type(
+        ctx.env.schema,
+        bool_type,
+        typename=sn.QualName(module="std", name="bool"),
+    )
+
+    # init set that will be used for __specified__
+    specified_path = irast.PathId.from_type(
+        ctx.env.schema,
+        stype,  # TODO: what's the type of this? named tuple?
```
aljazerzen, line 523: Also this.

edb/edgeql/compiler/viewgen.py
```diff
@@ -446,6 +473,133 @@ def gen_pointers_from_defaults(
...
+        path_id=subject_path,
+        typeref=subject_path.target,
+        shape=tuple(
+            [
+                (ptr_set, qlast.ShapeOp.ASSIGN)
+                for ptr_set in pointers.values()
+                if ptr_set
+            ]
+        ),
+    )
```
aljazerzen, line 509: @msullivan Can you take a look at this?

Is it ok to use stype as the type? I have a problem with `__subject__.title` compiling to a read from the table instead of the constant value that is provided in `subject_set.shape`.

edb/edgeql/compiler/viewgen.py
```diff
@@ -446,6 +473,133 @@ def gen_pointers_from_defaults(
...
+        #     continue
+
+        with ctx.new() as scopectx:
+            scopectx.anchors["__specified__"] = specified_set
+            scopectx.anchors["__subject__"] = subject_set
+            scopectx.partial_path_prefix = subject_set
+
+            # scopectx.env.path_scope.attach_path(subject_set.path_id, context=None)
+            scopectx.env.singletons.append(subject_set.path_id)
+            # scopectx.iterator_path_ids |= {subject_set.path_id}
```
aljazerzen, line 568: My rudimentary way of setting a singleton during compilation :D

edb/edgeql/compiler/viewgen.py
```diff
@@ -243,25 +241,54 @@ def _process_view(
...
+            ctx,
+        )
+        for pointer, ptr_set in defaults:
+            pointers[pointer] = ptr_set
+
+    # rewrites
+    rewrite_kind = (
+        qltypes.RewriteKind.Insert
+        if exprtype.is_insert()
+        else qltypes.RewriteKind.Insert
```
msullivan, line 269: Probably this is supposed to be `Update`?

edb/edgeql/compiler/viewgen.py
msullivan: Yeah, you'll need to generate a fresh tuple type. There's a `create` method on the `Tuple` type in schema

edb/edgeql/compiler/viewgen.py
```diff
@@ -446,6 +473,133 @@ def gen_pointers_from_defaults(
...
+    path_id = ir_set.path_id
+    scls_pointers = stype.get_pointers(ctx.env.schema)
+
+    # init set that will be used for __subject__
+    subject_path = irast.PathId.from_type(
+        ctx.env.schema,
+        stype,
+        typename=sn.QualName(module="__derived__", name="__subject__"),
+    )
+    subject_set = irast.Set(
```
msullivan, line 499: Don't create the `Set` explicitly, go through `new_set`

edb/edgeql/compiler/viewgen.py
msullivan: Yes, I think using `stype` is probably right. You'll need to do some plumbing on the pgsql side to get `__subject__` referring to the right stuff

edb/edgeql/compiler/viewgen.py
msullivan: Oh, actually, since you're constructing the whole tuple here, try `setgen.new_tuple_set`

edb/edgeql/compiler/viewgen.py
msullivan: And `ensure_set`/`expression_set` for when you have an Expr you want to turn into a `Set`. It will generate new pathids for it, too, if you don't give one

edb/edgeql/compiler/viewgen.py
msullivan: Hm... the `attach_path` approach should be the right one

edb/edgeql/compiler/viewgen.py
msullivan: Hm... There's a question of implementation strategies, though:
The thing I had in mind was that you'd add some additional IR fields to represent the rewrites.
I haven't thought about what would take to make it work if you are trying to do it by rewriting the shape.

I think that you can probably make the pure shape approach work for `INSERT` but I don't know if it can be made to work for `UPDATE`, where subtypes may have rewrites that don't appear in the type being `UPDATE`d

msullivan:
Alright I think this might be done!

edb/pgsql/compiler/ARCHITECTURE.md
```diff
@@ -25,6 +25,31 @@ materializing too much data, sets are compiled in two steps:
    - when an actual table is encountered, a plain ColRef to it's columns is
      returned.
 
+## Overlays
+
+Postgres as a limitation where effects of any DML are not visible in the same
```
elprans, line 30:
```suggestion
Postgres has a limitation where effects of any DML are not visible in the same
```

edb/pgsql/compiler/dml.py
```diff
@@ -749,9 +707,75 @@ def dynamic_get_path(
...
+    # The first serious bit of trickiness: if there are rewrites, the link
+    # table updates need to be done *before* we compute the rewrites, since
+    # the rewrites might refer to them.
+    #
+    # However, we can't unconditionally do it like this, because we
+    # want to be able to use ON CONFLICT to implement UNLESS CONFLICT
+    # ON when possible, and in that case the link table operations
+    # need to be done after the *actual insert*, because it is the actual
+    # insert that filters out conflicting rows. (This also means that we
+    # can't ues ON CONFLICT if there are rewrites.)
```
elprans, line 730:
```suggestion
    # can't use ON CONFLICT if there are rewrites.)
```

tests/test_edgeql_rewrites.py
```diff
@@ -0,0 +1,808 @@
...
+#
+
+
+import os.path
+
+import edgedb
+
+from edb.testbase import server as tb
+
+# from edb.tools import test
```
elprans, line 26:
```suggestion
```

edb/pgsql/compiler/dml.py
```diff
@@ -749,9 +707,75 @@ def dynamic_get_path(
     ctx.toplevel_stmt.append_cte(contents_cte)
     contents_rvar = relctx.rvar_for_rel(contents_cte, ctx=ctx)
 
+    rewrites = ir_stmt.rewrites and ir_stmt.rewrites.by_type.get(typeref)
+
+    pol_expr = ir_stmt.write_policies.get(typeref.id)
+    pol_ctx = None
+    if pol_expr or rewrites:
+        with ctx.new() as pol_ctx:
```
elprans, line 715: What's going on here?

elprans (approved):
Do we have tests on how rewrite alters affect baked expressions, e.g. function bodies?

msullivan:
> Do we have tests on how rewrite alters affect baked expressions, e.g. function bodies?

Can't put DML in function bodies

edb/pgsql/compiler/dml.py
```diff
@@ -841,6 +847,207 @@ def dynamic_get_path(
...
+    iterator_id: Optional[pgast.BaseExpr],
+    force_optional: bool,
+    ctx: context.CompilerContextLevel,
+) -> None:
+
+    with ctx.newscope() as insvalctx:
+        # This method is only called if the upper cardinality of
+        # the expression is one, so we check for AT_MOST_ONE
+        # to determine nullability.
+        # FIXME: Doing force_optional for all rewrites is kind of a bummer.
```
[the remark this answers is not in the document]
msullivan: The issue here is that some "required" pointers might be empty in the originally specified DML and we don't properly understand that yet. We'll want to fix it

edb/pgsql/compiler/dml.py
msullivan: I'll add a comment; I experimented with adding a function but that seems more confusing

edb/pgsql/compiler/dml.py
```diff
@@ -1206,13 +1283,22 @@ def insert_needs_conflict_cte(
...
         (type_id, True) in ctx.env.type_rewrites
         or (type_id, False) in ctx.env.type_rewrites
     ):
         return True
 
+    # We can't use ON CONFLICT if there are rewrites on the type
+    # because rewrites might reference multi pointers, which means
+    # we need to execute link operations before the final INSERT.
+    if ir_stmt.rewrites and ir_stmt.rewrites.by_type:
+        return True
```
aljazerzen, line 1300: What happens if there are ON CONFLICT and rewrites? Shouldn't the result of `merge_overlays_globally` be checked and an error thrown when True?

edb/edgeql/compiler/viewgen.py
```diff
@@ -1080,9 +1110,10 @@ def prepare_rewrite_anchors(
     schema = ctx.env.schema
 
     # init set for __subject__
+    # TODO: Do we really need a separate path id for __subject__?
```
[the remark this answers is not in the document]
msullivan: Right, but that `Foo` will be in a detached namespace so it will have a path_id like `ns~2@Foo`

edb/edgeql/compiler/viewgen.py
```diff
@@ -997,50 +970,104 @@ def _compile_rewrites_for_stype(
...
+                )
+            )
+            subject_set = setgen.class_set(
+                rewrite_view, path_id=subject_set.path_id, ctx=ctx)
+            if old_set:
+                old_set = setgen.class_set(
+                    rewrite_view, path_id=old_set.path_id, ctx=ctx)
+
+            ir_set = setgen.class_set(
+                rewrite_view, path_id=ir_set.path_id, ctx=ctx)
```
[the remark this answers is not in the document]
msullivan: Probably I should have put a comment here actually:
This gives the input sets the type of the relevant child, not of the parent it might be inheriting from.
Also `_normalize_view_ptr_expr` really wants its input to have a view type (that is, a subtype of the real object type), and not a material object type, so actually we make a new derived type

edb/pgsql/compiler/dml.py
msullivan: If we can't use SQL's `ON CONFLICT`, then we do conflict detection manually, and that's fine.

... unless the UNLESS CONFLICT is actually on a rewritten pointer, in which case I'm pretty sure we'll do the wrong thing, because the conflict checks on INSERT are compiled in a way that duplicates the original expressions in the set. (UNLESS CONFLICT doesn't work on volatile properties for this reason. I think we have the infrastructure that would be required to mostly fix this, though.)

For now we should probably disallow using UNLESS CONFLICT on a rewritten property though

PR #5014, opened 2023-02-10
Fix SQL introspection for Postgres 13

aljazerzen:


msullivan (approved):
Do we lose much by dropping these? Would it be worth doing something conditional?

fantix:
Per meeting discussion - mostly quoting @elprans: we shall advertise the EdgeDB server as a compatible Postgres cluster at **a particular version** through the binary protocol (and version-related SQL functions, etc.), regardless of the actual version of the backend Postgres in use. To unbreak CI and future releases, let's start from [Postgres 13](https://github.com/edgedb/edgedb/issues/5089) and merge this PR.

PR #5025, opened 2023-02-13
Document version pinning

### LINKED ISSUE #4929: Document version pinning behavior for `edgedb instance create` et al
Currently
```
edgedb instance create --version 2.6 demo2something
```
will install the latest version of 2.x while in order to install 2.6 you need to put an `=` in front, like so:
```
edgedb instance create --version =2.6 demo26
```

There is similar behavior in pyproject.toml.

We should document this.

(I also still think that on the command line, it should install the actual specified version, but I won't insist on relitigating that)

raddevon:
Closes #4929

### DIFF
--- a/docs/cli/edgedb_instance/edgedb_instance_create.rst
+++ b/docs/cli/edgedb_instance/edgedb_instance_create.rst
@@ -55,3 +55,23 @@ Options
     Specifies the version of the EdgeDB server to be used to run the
     new instance. To list the currently available options use
     :ref:`ref_cli_edgedb_server_list_versions`.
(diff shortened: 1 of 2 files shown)


elprans (approved):

1st1:
Should we also document `edgedb.toml` and how pinning is specified there?

raddevon:
> Should we also document `edgedb.toml` and how pinning is specified there?

@1st1 I almost included it here, but I didn't want to delay this behind a decision about where to put it. I'll merge this and work on that as a separate PR.

PR #5027, opened 2023-02-13
Write 2.10 changelog

msullivan:


docs/changelog/2_x.rst
```diff
@@ -876,6 +876,76 @@ Other changes
...
+* Fix array arguments in HTTP interface
+  (:eql:gh:`#4956`)
+
+* Support multi properties in ``UNLESS CONFLICT ON``
+  (:eql:gh:`#4955`)
+
+* Fix polymorphic type tests on result of update
+  (:eql:gh:`#4954`)
+
+* Upgrade our python linters
```
elprans, line 912: I would drop this and the `setuptools` note, as those are only relevant to EdgeDB devs.

msullivan: Oh yeah, oops

elprans (approved):

PR #5028, opened 2023-02-13
Fix non-transactional errors in Postgres 14.7

fantix:
Cherry-pick #5026 without bumping Postgres to 14.7

### DIFF
--- a/edb/server/pgcon/pgcon.pyx
+++ b/edb/server/pgcon/pgcon.pyx
@@ -935,11 +935,13 @@ cdef class PGConnection:
 
         if state is not None:
             self._build_apply_state_req(state, out)
-            if query.tx_id:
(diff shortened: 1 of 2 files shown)


elprans:
Why not just bump 2.x to 14.7?

fantix:
I'm ok to bump - it wasn't up to date with the master (14.4 comparing to 14.6) so I thought it was kept back for a reason.

msullivan:
Are we OK bumping the actual release of 2.x to 14.7?
(I'd prefer to keep the dev version synced with the release version, though it doesn't really matter)

elprans:
> Are we OK bumping the actual release of 2.x to 14.7?

Yes. I'll do that.  No reason not to bump to the latest bugfix release.

msullivan:
Roger, sounds good

PR #5031, opened 2023-02-14
Fix set single on required properties

### LINKED ISSUE #5029: `SET SINGLE USING` on a `required multi property` spuriously fails
```
edgedb> create type X { create required multi property foo -> str };
OK: CREATE TYPE
edgedb> insert X { foo := "test" };
[{"id": "e2b1ea82-ac12-11ed-8ac2-43e409d70c9b"}]
```

Trying to `set single` the property fails:
```
edgedb> alter type X alter property foo set single using (assert_single(.foo));
Error: MissingRequiredError: missing value for required property 'foo' of object type 'default::X'
```

I also tried with a constant in the using, still fails:
```
edgedb> alter type X alter property foo set single using ('x');
Error: MissingRequiredError: missing value for required property 'foo' of object type 'default::X'
```

Fails on 2.9 and on master. (I had thought maybe it was a regression introduced by #4873, but doesn't look like it)

aljazerzen:
Closes #5029

### DIFF
--- a/edb/pgsql/delta.py
+++ b/edb/pgsql/delta.py
@@ -19,6 +19,7 @@
 
 from __future__ import annotations
 from typing import *
+from copy import copy
 
 import collections.abc
 import itertools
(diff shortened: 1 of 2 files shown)


msullivan (approved):
Looks good!

PR #5032, opened 2023-02-14
Fix bug that dropping non-existing db leaves with unaccessible state

fantix:
Refs #4422, when dropping a database that doesn't exist, the database name is remembered as unaccessible by mistake, making the database unaccessible when actually created with that name. Fixed in a way that the existance of the database to be dropped (or used as a template) is checked before actually proceeding on.

Found in edgedb/edgedb-examples#105


elprans (commented):
I'm not sure this is the best possible fix.  Rather, the server should call `_allow_database_connections(dbname)` _after_ `DROP DATABASE` was completed, (regardless of possible errors).  Please also audit if there are other users of `_block_new_connections` that aren't re-allowing connections after exiting the critical section.

elprans (approved):

PR #5033, opened 2023-02-14
docs: Fix factual error about `EDGEDB_DSN` in client connection guide

elprans:
`EDGEDB_DSN` cannot be an instance name, `EDGEDB_INSTANCE` should be
used instead.

### DIFF
--- a/docs/clients/connection.rst
+++ b/docs/clients/connection.rst
@@ -3,7 +3,7 @@
 Connection
 ----------
 
-There are a couple ways to provide connection information to a client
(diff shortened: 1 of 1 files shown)


raddevon:
This looks good, @elprans. One quick suggestion while we're in here:
> There are a couple ways to provide connection information to a client library.

Could we change this first line to say "There are a few..." or "There are several..."?

raddevon (approved):
Approved, pending one suggestion for the intro sentence.

zackelan (approved):

Issue #4180, opened 2022-08-01
Add property removal example to docs


tehChromic:
Please add a simple example of a property remove:

`update Person filter .id = <uuid>'abc' set { middle_name-=.middle_name }`

it's not intuitive how to do this simple operation!

msullivan:
Not a bad idea.

I think a more idiomatic version (and what we should put in the docs) would be:
`set { middle_name := {} }`

PR #5039, opened 2023-02-15
Add example of unsetting a property to docs

fmoor:
fixes https://github.com/edgedb/edgedb/issues/4180

### DIFF
--- a/docs/datamodel/properties.rst
+++ b/docs/datamodel/properties.rst
@@ -66,6 +66,14 @@ particular order. If order is important, use an :ref:`array
 more involved discussion, see :ref:`EdgeQL > Sets
 <ref_eql_set_array_conversion>`.
 
+**Unsetting properties**
+
+To explicitly unset a property that is not required set it to an empty set.
+
+.. code-block:: edegql
+
+   update Person filter .id = <uuid>&id set { middle_name := {} };
+
 Default values
 --------------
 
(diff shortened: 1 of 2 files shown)


docs/datamodel/properties.rst
```diff
@@ -66,6 +66,14 @@ particular order. If order is important, use an :ref:`array
 more involved discussion, see :ref:`EdgeQL > Sets
 <ref_eql_set_array_conversion>`.
 
+**Unsetting properties**
+
+To explicitly unset a property that is not required set it to an empty set.
+
+.. code-block:: edegql
```
raddevon:
```suggestion
.. code-block:: edgeql
```

raddevon (commented):
Content looks good. I wonder about the organization though. Right now, it's under "Property cardinality" as a sub-section in the page on properties in schema. (Here's what it looks like rendered, btw)
image

Thinking about when the user will try to use this, I think it might be better as a new sub-section under ["Syntax" on the update documentation page](https://www.edgedb.com/docs/edgeql/update#syntax). Would that location that make sense?

fmoor:
> it might be better as a new sub-section under ["Syntax" on the update documentation page](https://www.edgedb.com/docs/edgeql/update#syntax)

Moved :+1:

raddevon (approved):

PR #5040, opened 2023-02-15
Add an `assert` function to the standard library

msullivan:
Using error raising functions in postgres (and thus in edgeql) is
somewhat fraught, since the semantics of when errors are raised and
when they may be "optimized out" is somewhat unclear.

We should document some recommended usage patterns:

This is probably the most important usage pattern, and works without
complication.
Just put it at the top level or in a `FOR` loop:
```
for obj in __new__ union (
  assert(
    not exists (obj.friends intersect obj.enemies),
    message := "invalid frenemies",
  )
)
```

Put the assertion in an `ORDER BY` clause:
```
select File { name, size }
filter .name in array_unpack(<array<str>>$names)
order by assert(.size <= 64*1024, message := "file too big")
```

Putting the `assert` call in an `order by` clause ensures it is only
evaluated on objects that passed the `filter`.
(In practice, putting the `assert` as part of the `filter` will
probably work, if it depends on the object itself, but `order by`
should be less fragile.)

Here things get kind of ugly, there are a couple options.

Put the assertion in a `FOR` iterator:
```
for _ in assert(not DatabaseStatus.closed, message := "db closed") union (
  select User { name, email } filter .name = 'Yury'
)
```
This should always work if the data is non-empty, but in some cases
might not raise the assertion if the object is empty.

Return a free object, and put a `FILTER` or `ORDER BY` on the free
object:
```
select { user := (select User { name, email } filter .name = 'Yury') }
order by assert(not DatabaseStatus.closed, message := "db closed");
```
This works but means the data is wrapped in a free object.

Make DML depend on the value:
```
with cond := assert(not DatabaseStatus.closed, message := "db closed"),
     _  := (for _ in (select 0 filter not cond) union (insert Dummy)),
select User { name, email } filter .name = 'Yury'
```
This is solid, but kind of atrocious. It will be less nasty maybe once
we have DML-supporting `IF`.


edb/lib/std/20-genericfuncs.edgeql
```diff
@@ -73,6 +73,27 @@ std::assert_distinct(
...
+) -> bool
+{
+    CREATE ANNOTATION std::description :=
+        "Assert that a boolean value is true.";
+    SET volatility := 'Stable';
+    USING SQL $$
+    SELECT (
+        edgedb.raise_on_null(
+            nullif("input", false),
+            'assert_failure',
```
elprans, line 91: Hm.  Not sure it's totally safe to use this.  Postgres may throw this as well (though currently only seems to be produced by `assert` in pl/pgSQL.  To be completely safe, I would use the same technique we use for other assertions: use a constraint violation error and specify `std::assert` as the constraint (`raise` has a named arg for the constraint name).

edb/pgsql/compiler/dml.py
```diff
@@ -560,23 +560,14 @@ def process_insert_body(
...
         typeref = typeref.material_type
 
-    type_val = pgast.TypeCast(
-        arg=pgast.StringConstant(val=str(typeref.id)),
-        type_name=pgast.TypeName(name=('uuid',)),
-    )
-    values.append(
-        pgast.ResTarget(name='__type__', val=type_val)
-    )
-
```
aljazerzen, line 579: This is not needed anymore because we don't store __type__ in the backing table anymore, right?

But what if a statement after `Insert` needs __type__? Something like:

```
select (Insert Movie { ... })[is Content]
```

aljazerzen (approved):
This feels like it should be two separate commits (one for assert and one for __type__ stuff).

edb/pgsql/compiler/dml.py
msullivan: Oh yeah that was just some git screwup. All of that is already merged. That's right, for why we aren't doing it.
It'll still get injected into the RETURNING clause of the insert when needed.

(This comes up a lot actually, when injecting type names for the CLI)

PR #5052, opened 2023-02-17
Do not store __type__ in the backing database

### LINKED ISSUE #4637: Consider not actually storing `__type__` in database
`__type__` is fully determined by what table a row comes from. We could populate it in the inheritance views instead of actually storing it in the database.

The space saving could add up and if there is a significant downside, I haven't though of it yet.
(I guess the downside is that it makes `__type__` a hacky special case instead of just a link, though there are already special cased things about it)

msullivan:
Every edgedb object has a `__type__` link, but its value is determined
uniquely by the type, and so we can save a uuid column per object in
the database by not actually storing it.

Inject the `__type__` values explicitly into the inheritance views we
generate, and inject the values in the compiler when we operate
directly on tables.

Fixes #4637.


edb/server/compiler/compiler.py
```diff
@@ -930,6 +930,14 @@ def describe_database_restore(
             and dump_server_ver.stage is not verutils.VersionStage.DEV
         )
 
+        dump_with_dunder_type = (
+            (
+                dump_server_ver is None
+                or dump_server_ver < (3, 0),
+            )
+            and dump_server_ver.stage is not verutils.VersionStage.DEV
```
elprans, line 938: I think we need to actually start recording the catalog version in dumps to avoid this imprecise dance.

edb/server/protocol/binary_v0.pyx
```diff
@@ -338,11 +338,14 @@ cdef class EdgeConnectionBackwardsCompatible(EdgeConnection):
 
             msg_buf = WriteBuffer.new_message(b'@')
 
-            msg_buf.write_int16(3)  # number of headers
+            msg_buf.write_int16(4)  # number of headers
             msg_buf.write_int16(DUMP_HEADER_BLOCK_TYPE)
             msg_buf.write_len_prefixed_bytes(DUMP_HEADER_BLOCK_TYPE_INFO)
             msg_buf.write_int16(DUMP_HEADER_SERVER_VER)
             msg_buf.write_len_prefixed_utf8(str(buildmeta.get_version()))
+            msg_buf.write_int16(DUMP_HEADER_SERVER_CATALOG_VERSION)
```
elprans, line 346: Need to also document the new header.

elprans (approved):

PR #5058, opened 2023-02-17
Fix parser build step

msullivan:


fantix (approved):

elprans (commented):
Should this be done in setup.py instead?

msullivan:
No, the setup doesn't actually write there, we just copy the output there after

PR #5072, opened 2023-02-21
Increase various test timeouts to hopefully reduce flakes

msullivan:
Our CI runners are super slow


elprans (approved):

fantix (approved):

msullivan:
When it failed I was really hoping that it had flaked, but it was just a lint :P

PR #5077, opened 2023-02-22
A bit of viewgen refactor

elprans:
In preparation for adding support for splats, refactor some of the shape
processing bits:

1. Use named tuples for shape element descriptors
2. Factor out early processing of `ShapeElement` and pass around an
   enhanced version of it.

### DIFF
--- a/edb/edgeql/compiler/viewgen.py
+++ b/edb/edgeql/compiler/viewgen.py
@@ -62,22 +62,34 @@ if TYPE_CHECKING:
 
 
 class ShapeElementDesc(NamedTuple):
+    """Annotated QL shape element for processing convenience"""
+
+    #: Shape element AST
     ql: qlast.ShapeElement
+    #: Canonical Path AST for the shape element
     path_ql: qlast.Path
+    #: The underlying pointer AST
     ptr_ql: qlast.Ptr
+    #: The name of the pointer
     ptr_name: str
+    #: Pointer source object
     source: s_sources.Source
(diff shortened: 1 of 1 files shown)


aljazerzen (approved):
This is going to have a few conflicts with mt rewrites, but it will also solve a few of my problems.

edb/edgeql/compiler/viewgen.py
```diff
@@ -57,12 +58,30 @@
...
 
 if TYPE_CHECKING:
-    from edb.schema import properties as s_props
     from edb.schema import sources as s_sources
 
-    ShapePtr = Tuple[
-        irast.Set, s_pointers.Pointer, qlast.ShapeOp, Optional[irast.Set]
-    ]
+
+class ShapeElementDesc(NamedTuple):
```
msullivan, line 64: Add some quick doc comments?

edb/edgeql/compiler/viewgen.py
```diff
@@ -341,6 +380,94 @@ def _process_view(
     return view_scls, ir_set
 
 
+def _shape_el_ql_to_shape_el_desc(
```
msullivan, line 383: Document what processing happens as part of this?

msullivan (approved):

msullivan:
This looks great; added a couple requests for a bit more documentation, since refactoring is a great time to add more docs, since it is all fresh in mind

PR #5080, opened 2023-02-22
Implement support for splats (RFC 1023)

elprans:


edb/edgeql/parser/grammar/expressions.py
```diff
@@ -1485,6 +1633,12 @@ def reduce_PathNodeName(self, *kids):
     def reduce_DUNDERTYPE(self, *kids):
         self.val = qlast.ObjectRef(name=kids[0].val)
 
+    # def reduce_STAR(self, *kids):
+    #     self.val = qlast.Splat(depth=1)
+
+    # def reduce_DOUBLESTAR(self, *kids):
+    #     self.val = qlast.Splat(depth=2)
+
```
aljazerzen, line 1641: You've probably forgot about this.

edb/edgeql/parser/grammar/expressions.py
```diff
@@ -511,6 +511,9 @@ def reduce_PathStepName_OptTypeIntersection(self, *kids):
 
         self.val = qlast.Path(steps=steps)
 
+    def reduce_Splat(self, *kids):
+        self.val = kids[0].val
+
```
aljazerzen, line 516: Comment in the class above needs to be updated.

aljazerzen (approved):

edb/edgeql/parser/grammar/expressions.py
```diff
@@ -1782,6 +1933,26 @@ def reduce_TypeExpr_AMPER_TypeExpr(self, *kids):
...
+    def reduce_SimpleTypeName(self, *kids):
+        self.val = kids[0].val
+
+    def reduce_NontrivialTypeExpr(self, *kids):
+        self.val = kids[0].val
+
+
+# The same as above, except complex forms must be parenthesized
+class ParenTypeExpr(Nonterm):
+    # def reduce_SimpleTypeName(self, *kids):
```
msullivan, line 1949: Should this be uncommented or should it be deleted?

edb/edgeql/compiler/viewgen.py
```diff
@@ -244,6 +250,100 @@ def _process_view(
...
 
+    explicit_ptr_names = {
+        desc.ptr_name for desc in shape_desc if not desc.is_linkprop
+    }
+
+    explicit_lprop_names = {
+        desc.ptr_name for desc in shape_desc if desc.is_linkprop
+    }
+
+    # Now look for any splats and expand them.
```
msullivan, line 261: Do you think it would be cleaner to break some of this off into a function?
(Maybe not)

edb/edgeql/compiler/viewgen.py
```diff
@@ -244,6 +250,100 @@ def _process_view(
...
+        if not isinstance(shape_el.expr.steps[0], qlast.Splat):
+            continue
+
+        if exprtype is not s_types.ExprType.Select:
+            raise errors.QueryError(
+                "unexpected splat operator in non-SELECT shape",
+                context=shape_el.expr.context,
+            )
+
+        if ctx.env.options.func_params is not None:
```
msullivan, line 272: We might want to make sure to disable it in other schema stuff, like constraints and indexes? (Though I think trying to do that will fail for other reasons, so)

msullivan (approved):
This is great

edb/edgeql/compiler/viewgen.py
elprans: We don't allow anything complex in those contexts already.

PR #5084, opened 2023-02-22
Merge delta_plan_input debug flag into delta_plan

msullivan:
I never want them separately and I never remember what the difference is.

### DIFF
--- a/edb/server/bootstrap.py
+++ b/edb/server/bootstrap.py
@@ -728,7 +728,7 @@ async def _make_stdlib(
         delta_command = s_ddl.delta_from_ddl(
             ddl_cmd, modaliases={}, schema=schema, stdmode=True)
 
-        if debug.flags.delta_plan_input:
+        if debug.flags.delta_plan:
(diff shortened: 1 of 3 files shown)


edb/server/bootstrap.py
```diff
@@ -728,7 +728,7 @@ async def _make_stdlib(
         delta_command = s_ddl.delta_from_ddl(
             ddl_cmd, modaliases={}, schema=schema, stdmode=True)
 
-        if debug.flags.delta_plan_input:
+        if debug.flags.delta_plan:
```
elprans, line 731: This should just be dropped, because `_process_delta` already dumps things, so you'll just be dumping the same thing twice.

edb/server/bootstrap.py
```diff
@@ -849,7 +849,7 @@ async def _amend_stdlib(
...
```
elprans, line 852: Ditto

elprans (approved):

PR #5088, opened 2023-02-23
Implement compilation of triggers

### LINKED ISSUE #4936: Implement triggers RFC

msullivan:
Closes #4936.

The interesting part of the compilation runs at the end of IR->SQL and 
injects the triggers into CTEs. The overlay system is (somewhat hackily)
used to drive what objects are triggered on and (slightly less hackily) to
provide a view of the new database state.

I have one known bug at the moment, which is approximately that we apply
access policies *too much* in one case. I have a plan for fixing it that
doesn't change that much of the big picture, and I wanted to get something
up for review.

### DIFF
--- a/edb/edgeql/compiler/stmtctx.py
+++ b/edb/edgeql/compiler/stmtctx.py
@@ -141,8 +141,7 @@ def fini_expression(
         ir = setgen.scoped_set(ir, ctx=ctx)
 
     # Compile any triggers that were triggered by the query
-    ir_triggers = triggers.compile_triggers(
-        ast_visitor.find_children(ir, irast.MutatingStmt), ctx=ctx)
+    ir_triggers = triggers.compile_triggers(ctx.env.dml_stmts, ctx=ctx)
 
     # Collect all of the expressions stored in various side sets
     # that can make it into the output, so that we can make sure
--- a/edb/edgeql/compiler/stmtctx.py
+++ b/edb/edgeql/compiler/stmtctx.py
@@ -56,6 +56,7 @@ from . import pathctx
 from . import setgen
 from . import viewgen
 from . import schemactx
+from . import triggers
 from . import tuple_args
 from . import typegen
 
@@ -139,6 +140,10 @@ def fini_expression(
     ):
         ir = setgen.scoped_set(ir, ctx=ctx)
 
+    # Compile any triggers that were triggered by the query
+    ir_triggers = triggers.compile_triggers(
+        ast_visitor.find_children(ir, irast.MutatingStmt), ctx=ctx)
+
     # Collect all of the expressions stored in various side sets
     # that can make it into the output, so that we can make sure
     # to catch them all in our fixups and analyses.
@@ -153,6 +158,7 @@ def fini_expression(
         p.sub_params.decoder_ir for p in ctx.env.query_parameters.values()
         if p.sub_params and p.sub_params.decoder_ir
     ]
+    extra_exprs += [trigger.expr for trigger in ir_triggers]
 
     all_exprs = [ir] + extra_exprs
 
@@ -268,6 +274,7 @@ def fini_expression(
             if isinstance(s, irast.Set)},
         dml_exprs=ctx.env.dml_exprs,
         singletons=ctx.env.singletons,
+        triggers=ir_triggers,
     )
     return result
 
--- a/edb/edgeql/compiler/triggers.py
+++ b/edb/edgeql/compiler/triggers.py
@@ -1,0 +1,155 @@
+#
+# This source file is part of the EdgeDB open source project.
+#
+# Copyright 2008-present MagicStack Inc. and the EdgeDB authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+"""EdgeQL trigger compilation."""
+
+
+from __future__ import annotations
+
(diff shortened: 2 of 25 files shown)


edb/schema/inheriting.py
```diff
@@ -98,10 +98,15 @@ def inherit_fields(
         inherited_fields_update = {}
         deferred_complex_ops = []
 
-        for field_name in field_names:
+        # Iterate over mcls.get_schema_fields() instead of field_names for
```
elprans, line 101: Why is this important?

edb/edgeql/compiler/stmtctx.py
```diff
@@ -139,6 +140,10 @@ def fini_expression(
     ):
         ir = setgen.scoped_set(ir, ctx=ctx)
 
+    # Compile any triggers that were triggered by the query
+    ir_triggers = triggers.compile_triggers(
+        ast_visitor.find_children(ir, irast.MutatingStmt), ctx=ctx)
```
elprans, line 145: Do we have any idea how slow those traversals are?  Any reason not to collect pending triggers in context as you make DML IR?

edb/pgsql/ast.py
```diff
@@ -1025,6 +1025,8 @@ class IteratorCTE(ImmutableBase):
     cte: CommonTableExpr
     parent: typing.Optional[IteratorCTE]
 
+    other_paths: tuple[tuple[irast.PathId, str], ...] = ()
```
elprans, line 1028: Add a comment?

edb/schema/inheriting.py
msullivan: Compilation of the trigger expression depends on the values of `kind` and `scope`, and so we need to arrange to inherit those before we try to inherit `expr`

elprans (approved):
:+1:

edb/edgeql/compiler/stmtctx.py
msullivan: I haven't measured it. I did a traversal because I think it gets hard to keep track of too much state, and we already have a `dml_stmts` that tracked something slightly different (deletes weren't included) and I didn't want to be tracking dml in two ways. I think the actually right answer though is to just track deletes in that path

edb/edgeql/compiler/triggers.py
```diff
@@ -0,0 +1,150 @@
...
+    schema = ctx.env.schema
+
+    trigger_map: dict[
+        s_triggers.Trigger,
+        set[tuple[s_objtypes.ObjectType, irast.MutatingStmt]],
+    ] = {}
+    for stmt in dml_stmts:
+        kind = TRIGGER_KINDS[type(stmt)]
+
+        stype = schemactx.concretify(
```
aljazerzen, line 124: What does this do? I cannot gather from the function documentation. What are "views" in this context?

edb/edgeql/compiler/triggers.py
```diff
@@ -0,0 +1,150 @@
...
+        # Process all the types, starting with the base type
+        for subtype in sorted(stypes, key=lambda t: t != stype):
+            for trigger in subtype.get_relevant_triggers(kind, schema):
+                mro = (trigger, *trigger.get_ancestors(schema).objects(schema))
+                base = mro[-1]
+                tmap = trigger_map.setdefault(base, set())
+                # N.B: If the *base type* of the DML appears, that suffices,
+                # because it covers everything, and we don't need to duplicate.
+                if (stype, stmt) not in tmap:
+                    tmap.add((subtype, stmt))
```
aljazerzen, line 143: I think this could be "If *any of the previously handled triggers of ancestors* appears, ...

This would have effect when you have a trigger on `Child1` in hierarchy `Updated > Child1 > Child2`. 
You would first add the trigger on Child1, but when processing Child2, the trigger would be added again, even though it's effects have already been covered.

(I may not be understanding this whole chunk correctly)

aljazerzen (approved):
Looks good!

I still have to re-read the pg compiler side with a fresh set of brain, but I can do that even after the merge.

edb/edgeql/compiler/triggers.py
```diff
@@ -0,0 +1,150 @@
...
+        # descendant types have triggers.
+        if isinstance(stmt, irast.InsertStmt):
+            stypes = {stype}
+        else:
+            stypes = schemactx.get_all_concrete(stype, ctx=ctx)
+
+        # Process all the types, starting with the base type
+        for subtype in sorted(stypes, key=lambda t: t != stype):
+            for trigger in subtype.get_relevant_triggers(kind, schema):
+                mro = (trigger, *trigger.get_ancestors(schema).objects(schema))
```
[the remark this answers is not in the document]
msullivan: Yeah, it does, though we don't actually care here.

Though also: https://github.com/edgedb/edgedb/issues/5118 :cry:

edb/edgeql/compiler/triggers.py
msullivan: Yeah, it is a specific interaction with how the pgsql/dml.py side is implemented. I'm elaborating a bit more in the comment.

edb/edgeql/compiler/triggers.py
```diff
@@ -0,0 +1,150 @@
...
+    *,
+    ctx: context.ContextLevel,
+) -> irast.Trigger:
+    schema = ctx.env.schema
+
+    scope = trigger.get_scope(schema)
+    kinds = set(trigger.get_kinds(schema))
+    source = trigger.get_subject(schema)
+
+    with ctx.detached() as _, ctx.newscope(fenced=True) as sctx:
```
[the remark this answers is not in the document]
msullivan: DetachedExpr will trigger a new `detached` context when processing. I think the main difference is I'm just doing it directly

edb/edgeql/compiler/triggers.py
msullivan: "view" is what "shape" used to be called, and which it is still sometimes called internally. But more generally, all of the transient subtypes we generate during compilation.

`concretify` finds non-view concrete base types, but it does it recursively through unions and intersections.

Arguably this should be the normal thing? Idk.

PR #5090, opened 2023-02-23
Implement `fts::match` function.

vpetrovykh:
The function that takes weights into account and can produce ranking and match highlighs.


msullivan (approved):
Also bump the catalog version please

vpetrovykh:
Catalog version updated.

PR #5092, opened 2023-02-24
[DRAFT] An initial take on an fts::test that hits indexes

msullivan:
Implement an inlined version of fts::test that succefully hits
indexes. (I tested using imdbench extended with some fts::textsearch
indexes.)

Per some discussions with @vpetrovykh, I introduced an fts::make_doc
function that is a variadic optional function that produces a string
that can be reasonable passed to an fts::textsearch index and that
works well will fts::test. (fts::test does the same operations.)

make_doc essentially does `array_to_string("docs", ' ')`, but inlined.

The most unpleasant bit about all of this, though, in handling the
language argument to postgres's `to_tsvector`, and how it interacts
with our constant extraction preprocessing. Because we extract string
literals during preprocessing, the `language` argument to `fts::test`
and similar will typically be a parameter.

The language argument is of type `regconfig` (which is an int id);
there is a cast from `text` to `regconfig`, but if the argument to
language is a text query parameter cast to `regconfig`, the index will
not match and the query will be slow.

The index *will* match if the argument is of type `regconfig` (which
is to say, an int32), but that doesn't help us any.

What *does* work is if it the argument is a string literal that is
cast directly to `regconfig`.

The only way I could come up with of accomplishing this is that the
normalization pass no longer extracts string literals that appear
directly after the token sequence `language :=`.
If anybody has a better idea I'd love to know it.

### DIFF
--- a/edb/edgeql-rust/src/normalize.rs
+++ b/edb/edgeql-rust/src/normalize.rs
@@ -136,8 +136,10 @@ pub fn normalize<'x>(text: &'x str)
         }
     };
     let mut last_was_set = false;
+    let mut language_state = 0;
     for (idx, tok) in tokens.iter().enumerate() {
         let mut is_set = false;
+        let mut new_language_state = 0;
         match tok.kind {
             Kind::IntConst
             // Don't replace `.12` because this is a tuple access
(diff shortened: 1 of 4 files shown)


elprans:
> If anybody has a better idea I'd love to know it.

Perhaps you can do this at the IR level?  I.e. specially handle the `language` argument and put a hint in the argument cast for the real desired type (i.e. `regconfig`).

msullivan:
I'm closing this because it is bad, I have a second try coming soon

PR #5095, opened 2023-02-24
Prettyprint : instead of -> in links and properties

### LINKED ISSUE #5049: Pretty-print SDL and DDL using colons
We introduced new syntax for pointer declarations: `link foo: Bar` is allowed in addition to `link foo -> Bar`.

We should pretty-print to the new syntax. The main visible effect of this will be in generated migration scripts.

Most of the work here will be test suite stuff.

Follow up to #4942

efaust:
Change from -> to : when prettyprinting links and properties, and fix up the resulting havoc in the test suite.

Fixes #5049


msullivan (commented):
This PR is solid, but I'm not approving it now because we are all going to argue about syntax on slack for a while.

msullivan (approved):

PR #5101, opened 2023-02-27
Document deprecation policy

raddevon:
I'm not sure about the placement of this. It's the best I could come up with. Open to suggestions.

I don't think it fits cleanly under any of our existing sections. I didn't want to add a new section for it. I reasoned that it might be relevant to people who are looking at changelogs, so that's where I have it currently.

### DIFF
--- a/docs/changelog/deprecation.rst
+++ b/docs/changelog/deprecation.rst
@@ -1,0 +1,10 @@
+.. _ref_changelog_deprecation:
+
+==================
+Deprecation Policy
+==================
+
+* We continue to support one previous version of EdgeDB with critical bug fixes
+  only.
+* Client bindings support the current and previous versions.
+* CLI supports all versions from version 1.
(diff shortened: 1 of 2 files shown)


docs/changelog/deprecation.rst
```diff
@@ -0,0 +1,10 @@
+.. _ref_changelog_deprecation:
+
+==================
+Deprecation Policy
+==================
+
+* We continue to support one previous version of EdgeDB with critical bug fixes
+  only.
```
1st1, line 8: Drop "only"

docs/changelog/deprecation.rst
```diff
@@ -0,0 +1,10 @@
...
+* Client bindings support the current and previous versions.
```
1st1, line 9:
```suggestion
* Client bindings will support the current and the previous version.
```

1st1 (approved):

PR #5102, opened 2023-02-27
fix: SQL introspection on Postgres 13

aljazerzen:
Ref https://github.com/edgedb/edgedb/pull/4991


fantix (approved):
Tested and passed! Thanks!

PR #5103, opened 2023-02-27
Refactor viewgen to prepare for rewrites

aljazerzen:
Introduces a new dataclass `ShapeContext`, which contains common
params of functions in viewgen.

I needed this, because I with rewrites I was adding ~5 functions 
with ~10 params each, most of which were just passed on to
`_normalize_view_ptr_expr`.

### DIFF
--- a/edb/edgeql/compiler/viewgen.py
+++ b/edb/edgeql/compiler/viewgen.py
@@ -23,12 +23,12 @@
 from __future__ import annotations
 
 import collections
+import dataclasses
 import functools
 from typing import *
 
 from edb import errors
 from edb.common import ast
-from edb.common import context as pctx
 from edb.common import topological
 from edb.common.typeutils import not_none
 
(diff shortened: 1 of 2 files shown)


edb/edgeql/compiler/viewgen.py
```diff
@@ -98,6 +98,21 @@ class ShapePtr(NamedTuple):
     target_set: Optional[irast.Set]
 
 
+@dataclasses.dataclass(kw_only=True, frozen=True)
+class ShapeContext:
+    # a helper object for passing shape compile parameters
+
+    ctx: context.ContextLevel
```
msullivan, line 105: Hm... Not 100% sold on including ctx, but I dont' feel strongly

msullivan (approved):

edb/edgeql/compiler/viewgen.py
elprans: Yeah, I think I'd prefer it to remain an explicit argument.

PR #5108, opened 2023-02-28
Support casting between scalars with a common concrete base

msullivan:
Seems like this was not supported because of an oversight, and then I
unthinkingly added a test that asserted the behavior after I permuted
some related code in #3662.

### DIFF
--- a/edb/edgeql/compiler/casts.py
+++ b/edb/edgeql/compiler/casts.py
@@ -36,6 +36,7 @@ from edb.schema import constraints as s_constr
 from edb.schema import functions as s_func
 from edb.schema import indexes as s_indexes
 from edb.schema import name as sn
+from edb.schema import scalars as s_scalars
 from edb.schema import types as s_types
 from edb.schema import utils as s_utils
 from edb.schema import name as s_name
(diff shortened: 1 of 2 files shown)


edb/edgeql/compiler/casts.py
```diff
@@ -236,6 +241,20 @@ def compile_cast(
...
+        orig_stype: s_types.Type,
+        new_stype: s_types.Type, *,
+        ctx: context.ContextLevel) -> bool:
+    return bool(
+        isinstance(orig_stype, s_scalars.ScalarType)
+        and isinstance(new_stype, s_scalars.ScalarType)
+        and (nearest := s_utils.get_class_nearest_common_ancestors(
+            ctx.env.schema, [new_stype, orig_stype]))
+        and len(nearest) == 1
+        and not nearest[0].get_abstract(ctx.env.schema)
```
elprans, line 254: This wouldn't work if the nearest common ancestor is an abstract subtype of a non-abstract parent, e.g

```
create abstract scalar type foo extending str;
create scalar type bar1 extending foo;
create scalar type bar2 extending foo;
```

So:

```suggestion
        and nearest[0].maybe_get_topmost_concrete_base(ctx.env.schema) is not None
```

msullivan: Good catch

elprans: Actually, I think the whole check should just compare the results of `maybe_get_topmost_concrete_base`, because

```
create abstract scalar type foo extending str;
create abstract scalar type bar extending str;
create scalar type bar1 extending foo, bar;
create scalar type bar2 extending foo, bar;
```

msullivan: Oh, yeah, indeed; I had misremembered and thought scalars could only have one base, but it's actually they can have one *concrete* ancestor

msullivan: And this is probably faster too

elprans (approved):

PR #5110, opened 2023-02-28
Disallow variadic arguments with optional types in user code.

### LINKED ISSUE #4982: User-defined functions with variadic+optional arguments have problems
Variadic functions have all their variadic arguments packed into an array. If one of those arguments is `NULL`, the array will have a `NULL` in it, which is not legit in edgeql.

```
edgedb> create function foo(variadic doc: optional int64) -> array<int64> using (assert_exists(doc));
OK: CREATE FUNCTION
edgedb> select foo(1, <int64>{}, 3);
[[1, null, 3]]
edgedb> \set output-format default
edgedb> select foo(1, <int64>{}, 3);
edgedb error: ProtocolEncodingError: ProtocolEncodingError: unexpected end of frame
Reconnecting...
```

I'm not sure that there is any sensible way to handle this in edgeql as it exists now, so we should probably just ban this.

efaust:
optionally typed variadic arguments are fraught, because we cannot put `null`
into arrays. Rather than try to find a roundabout solution, just ban optionally
typed variadic parameters in user code.

Fixes #4982.

### DIFF
--- a/edb/schema/functions.py
+++ b/edb/schema/functions.py
@@ -1879,6 +1879,20 @@ class CreateFunction(CreateCallableObject[Function], FunctionCommand):
                         f'{p_type.get_displayname(schema)}',
                         context=self.source_context)
 
+        # Make sure variadic parameters do not contain optional types in
+        # user-defined functions
+        if language == qlast.Language.EdgeQL:
+            if variadic := params.find_variadic(schema):
+                typemod = variadic.get_typemod(schema)
+                if typemod is ft.TypeModifier.OptionalType:
+                    raise errors.InvalidFunctionDefinitionError(
+                        f'cannot create the `{signature}` function: '
+                        f'variadic argument '
+                        f'`{variadic.get_displayname(schema)}` '
+                        f'illegally declared with optional type in '
+                        f'user-defined function',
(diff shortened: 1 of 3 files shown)


edb/schema/functions.py
```diff
@@ -1879,6 +1879,19 @@ def _create_begin(
...
+        # user-defined functions
+        if language == qlast.Language.EdgeQL:
+            if variadic := params.find_variadic(schema):
+                typemod = variadic.get_typemod(schema)
+                if typemod is ft.TypeModifier.OptionalType:
+                    raise errors.InvalidFunctionDefinitionError(
+                        f'cannot create the `{signature}` function: '
+                        f'variadic argument '
+                        f'`{variadic.get_displayname(schema)}` '
+                        f'illegally declared with optional type in user code',
```
msullivan, line 1892: I'd drop the `in user code` part; users will only be defining edgeql user functions

msullivan (approved):

edb/schema/functions.py
efaust: Discussed. Changed to "in user-defined function" to match many other errors in the file.

PR #5111, opened 2023-03-01
[DRAFT] An fts::test that hits indexes

msullivan:
Implement an inlined version of fts::test that succefully hits
indexes. (I tested using imdbench extended with some fts::textsearch
indexes, and added some tests to the explain tests)

Per some discussions with @vpetrovykh, I introduced an fts::make_doc
function that is a variadic optional function that produces a string
that can be reasonable passed to an fts::textsearch index and that
works well will fts::test. (fts::test does the same operations.)

make_doc essentially does `array_to_string("docs", ' ')`, but inlined.

The most fiddly bit about all of this, though, in handling the
language argument to postgres's `to_tsvector`, and how it interacts
with our constant extraction preprocessing. Because we extract string
literals during preprocessing, the `language` argument to `fts::test`
and similar will typically be a parameter.

The language argument is of type `regconfig` (which is an int id);
there is a cast from `text` to `regconfig`, but if the argument to
language is a text query parameter cast to `regconfig`, the index will
not match and the query will be slow.

The index *will* match if the argument is of type `regconfig`, and so
we try to make sure this happens. We introduce a new scalar type,
`fts::language`, that is compiled to `regconfig`.

The binary serialization for `regconfig` is as an int32, so we
introduce machinery to use the text serialization for regconfig params
instead. (Thanks to @elprans for figuring this trick out.)

We then automatically "promote" `str` params that are casted (possibly
implicitly) to `fts::language` to be `fts::language` params, which lets
using a constant literal param to `language` work right.

The bulk of the code here ends up being code to handle introducing our
first type that we want to use text serialization for in the binary
protocol.


edb/edgeql/compiler/stmtctx.py
```diff
@@ -272,6 +277,54 @@ def fini_expression(
...
+    """Rewrite parameters that are cast to fts::language
+
+    FTS is very fiddly about its language arguments; it has an
+    internal type for them (regconfig), and it will not match indexes
+    if you produce it by casting from text, though casting directly
+    from a string literal works. This causes complications for us,
+    because our parameter extraction runs before anything.
+
+    We solve the problem by changing the type of any parameter that is
+    only used once and is directly cast to fts::language (likely
```
elprans, line 292: Why the "only used once" restriction?  I think we can safely do this as long as every use is an `<fts::language><str>` cast.

edb/ir/typeutils.py
```diff
@@ -120,6 +120,16 @@ def is_bytes(typeref: irast.TypeRef) -> bool:
     return typeref.real_base_type.id == s_obj.get_known_type_id('std::bytes')
 
 
+def needs_text_serialization(typeref: irast.TypeRef) -> bool:
+    return (
+        typeref.real_base_type.id == s_obj.get_known_type_id('fts::language'))
+
+
+def needs_semitext_serialization(typeref: irast.TypeRef) -> bool:
```
elprans, line 128: Why not fold these into a single function?

edb/edgeql/compiler/stmtctx.py
msullivan: I did "only used once" because it was simpler to implement and because I was mostly thinking about the constant-extraction case, where each extracted constant is only used once, but I agree that we could do your thing and that it would be better in cases with an explicit <str> param.

edb/ir/typeutils.py
msullivan: Hm, we could. I think of them as being conceptually distinct, but all the existing cases where we use `needs_text_serialization` could be `needs_semitext_serialization` since the cases can be distinguished by context.

msullivan:
Going to close this because we went a different direction with FTS.

It's possible that in the future we will end up with a type that needs text serialization, in which case we can recover the whole `needs_text_serialization` machinery from this PR.

Ideally we won't have to, though.

PR #5115, opened 2023-03-01
pg-ext: pretend to be Postgres 13

### LINKED ISSUE #5089: Advertise EdgeDB server as Postgres 13 compatible through SQL
The current SQL support (through Postgres binary protocol) in EdgeDB is actually only compatible with Postgres 13, regardless of the actual version of the backend Postgres in use. We should also advertise `13` instead of `14` or higher through the binary protocol.

fantix:
Patched `server_version` and `server_version_num` to 13.9

* Also fixed conn params `session_authorization` and `application_name`
* SQL `version()` now shows EdgeDB version info

Fixes #5089


edb/server/compiler/compiler.py
```diff
@@ -526,17 +526,28 @@ def parse_search_path(search_path_str: str) -> list[str]:
...
+        # frontend-only settings and their mutability
+        frontend_only = {
+            'search_path': True,
+            'server_version': False,
+            'server_version_num': False,
+        }
         stmts = pg_parser.parse(query_str)
         sql_units = []
         for stmt in stmts:
+            fe_only = stmt.name in frontend_only
```
aljazerzen, line 538: Did you mean to write `fe_only = frontend_only[stmt.name]`?

fantix: Ah no, that's a bad naming - let me update, thanks!

aljazerzen (approved):

PR #5116, opened 2023-03-01
CI: Add support to wait until all is ready to publish

### LINKED ISSUE #5043: Have release builds sync up publishing
Currently every target publishes independently, so we can have half of the targets published while we are dealing with build errors and whatnot on other targets.

I think probably we should wait to publish until everything is done?
This could either be done by actually delaying the publish jobs or else by publishing to a staging area and then having a final job that flips the staging area live

fantix:
Fixes #5043 

[sample run](https://github.com/edgedb/edgedb/actions/runs/4306353850)

### DIFF
--- a/.github/workflows.src/release.tpl.yml
+++ b/.github/workflows.src/release.tpl.yml
@@ -6,4 +6,4 @@ on:
     inputs: {}
 
 jobs:
-  <<- workflow(targets, subdist="") ->>
+  <<- workflow(targets, subdist="", publish_all=True) ->>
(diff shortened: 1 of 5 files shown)


.github/workflows.src/release.tpl.yml
```diff
@@ -6,4 +6,4 @@ on:
     inputs: {}
 
 jobs:
-  <<- workflow(targets, subdist="") ->>
+  <<- workflow(targets, subdist="", publish_all=True) ->>
```
elprans, line 9: The testing channel should get this too.

elprans (approved):

PR #5117, opened 2023-03-01
CI: Fix tests on different versions of Postgres

### LINKED ISSUE #4081: Fix CI on different Postgres

fantix:
Now the CI workflow `test-pg-versions` runs on Postgres:
 * 13
 * 14
 * 14 single-role
 * 14 single-db (failure allowed)
 * 15 (failure allowed, but we should fix)

Fixes #4081

[sample run](https://github.com/edgedb/edgedb/actions/runs/4304779276)

Also **dropped pre-1.0 dumps** and fixed a flakiness in single-db test.

- [ ] Document in 3.0 changelog that upgrading pre-1.0 instances would need 2 steps of dump-restore through 2.x.


Syzuna:
Postgres 15 fails because they removed datlastsysoid end of January.
I had to make the painful experience of EdgeDB failing to work on it after that.
I know it is my fault for even trying 15 but it worked at first and then in a point release they removed datlastsysoid and broke it all

elprans (approved):

fantix:
@Syzuna thanks for trying anyways! Postgres 15 is now tracked in #5231

PR #5124, opened 2023-03-03
Store the rel_overlays in immutable data structures

msullivan:
In conflict clauses, and triggers, and policies, and probably soon in
rewrites, we set up temporary modifications to the overlay state.
Overlays are currently stored as dicts of dicts of lists, and so doing
anything short of a full clone terrifies me, but a full clone of a 3
level nested data structure is also definitely a smell.

Switch to using immutable data structures for it. Since *usually*
though we want to make mutable updates to it that affect other
contexts, we store pointers to it in a new RelOverlays object.

See the comments in the PR for more info. I also wrote a big new
comment about overlays in general, which we'll want to include even if
I don't merge this PR. I think I'm like 80% on merging this, but want
to try fixing a triggers bug on top of it first.

### DIFF
--- a/edb/pgsql/compiler/context.py
+++ b/edb/pgsql/compiler/context.py
@@ -71,6 +71,13 @@ class OutputFormat(enum.Enum):
 NO_STMT = pgast.SelectStmt()
 
 
+OverlayEntry = tuple[
+    str,
+    Union[pgast.BaseRelation, pgast.CommonTableExpr],
+    'irast.PathId',
+]
+
+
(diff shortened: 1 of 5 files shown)


elprans (approved):
Nice!

aljazerzen (approved):
Nice documentation write up.

I added a little docs on that into ARCHITECTURE.md, which should probably be moved here.

https://github.com/edgedb/edgedb/pull/5006/commits/86292691da6f30f14a4394bddf6fe9d9dd117963

Issue #4938, opened 2023-01-26
Add some way to run SQL administration commands like ANALYZE and REINDEX


msullivan:
The main complications here are annoying syntactic things:

* Maybe we want to put them all in some heading, like `ADMINISTRATE`, instead of adding a lot of new bare commands
* But there is nothing like `ADMINISTRATE` that would make a good reserved keyword
* `ANALYZE` and `REINDEX` are already reserved
* If we don't allow `WITH` bindings before the commands, we can use a non-reserved keyword

So I guess my proposal is to add `ADMINISTRATE ANALYZE` and `ADMINISTRATE REINDEX` as commands, and to disallow the use of `WITH` before them.

We can even make `ANALYZE` and `REINDEX` non-reserved.

(My main direct use case here is that I want to run `ANALYZE` after we create the database in the explain test case)

vpetrovykh:
I think that `ADMINISTRATE` is not a very clear choice of keyword to be used as an umbrella term. I'm not sure how `ANALYZE` and `REINDEX` fit under this umbrella, while `ROLE` commands do not. It may be that we're trying to distinguish commands that are used to configure how EdgeDB works vs commands that are involved in maintenance. Perhaps a `SYSTEM` umbrella might be a better indicator that the commands there are more about system maintenance than configuring day-to-day operation. Generally we can expect any future maintenance command to be a verb and follow `SYSTEM <VERB>` pattern.

vpetrovykh:
`ANALYZE` command in Postgres has several levels of granularity: everything, table, and column. In EdgeDB we would like to have similar granularity, but since we don't have tables and columns, we would specify it in terms of Objects and individual pointers.

The most general usage would still be just `ANALYZE` without any parameters. It would collect statistics on all the database contents.

The next level of granularity would be `ANALYZE Foo` where `Foo` denotes some object type. This is roughly equivalent to `ANALYZE <table>` in Postgres, however there are some notable differences. Since out objects may include things like multi links, the data about the object may be spread across several tables (one for the core properties and such, plus one for each of the link tables). 

The final level of granularity is at the link or property level. The basic format would be `ANALYZE Foo.bar`. Link properties would always be included in the analysis as long as their parent link is included and would never need to be specified explicitly. It's not clear there's much benefit to extending this granularity to allow specifying just the link property as the target of the command, although syntactically, it's also possible to provide this option.

[...]

PR #5126, opened 2023-03-03
Add `ANALYZE` to Edgedb commands.

vpetrovykh:
Add a plain `ANALYZE` command without options or arguments.

Issue #4938

### DIFF
--- a/edb/edgeql/parser/__init__.py
+++ b/edb/edgeql/parser/__init__.py
@@ -58,14 +58,45 @@ def parse_fragment(
     return res
 
 
-def parse(
+def parse_single(
+    source: Union[qltokenizer.Source, str],
+    filename: Optional[str]=None,
+) -> qlast.Expr:
+    if isinstance(source, str):
+        source = qltokenizer.Source.from_string(source)
+    parser = qlparser.EdgeQLSingleParser()
+    res = parser.parse(source, filename=filename)
+    assert isinstance(res, qlast.Expr)
+    return res
+
+
+def parse_query(
+    source: Union[qltokenizer.Source, str],
+    module_aliases: Optional[Mapping[Optional[str], str]] = None,
+) -> qlast.Query:
+    """Parse some EdgeQL potentially adding some module aliases.
+
+    This will parse EdgeQL queries and expressions. If the source is an
+    expression, the result will be wrapped into a SelectQuery.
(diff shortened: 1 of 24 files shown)


vpetrovykh:
I suspect that there might be more `dbstate` stuff that I should account for in my `MaintenanceQuery`.

edb/server/compiler/compiler.py
```diff
@@ -1796,6 +1814,11 @@ def _compile_dispatch_ql(
             caps |= enums.Capability.MODIFICATIONS
         return (query, caps)
 
+    elif isinstance(ql, qlast.AnalyzeStmt):
+        query = _compile_ql_analyze(ctx, ql, script_info=script_info)
+        caps = enums.Capability(0)
```
msullivan, line 1819: We'll probably want some capabilities required here if we ever take this out of dev mode.
Not sure if it matters now, though.

msullivan (approved):
Looks good!

Could add a non dev mode test (in userddl or some such) for the error message, though it's pretty marginal

tests/test_edgeql_explain.py
```diff
@@ -55,60 +53,11 @@ class TestEdgeQLExplain(tb.QueryTestCase):
             update User set {
               todo += (select .owned_issues filter <int64>.number % 3 = 0)
             };
+
+            analyze;
         '''
     ]
 
-    @classmethod
```
msullivan, line 61: Very happy to have all this gone!

edb/edgeql/parser/__init__.py
```diff
@@ -58,14 +58,37 @@ def parse_fragment(
...
+    parser = qlparser.EdgeQLSingleParser()
+    res = parser.parse(source, filename=filename)
+    assert isinstance(res, qlast.Expr)
+    return res
+
+
 def parse(
     source: Union[qltokenizer.Source, str],
     module_aliases: Optional[Mapping[Optional[str], str]] = None,
+    as_expr: bool = True,
```
elprans, line 76: I'd prefer if we had a separate `parse_expr` function instead.

elprans: And maybe then rename `parse` to `parse_command`

msullivan (approved):
Great

PR #5129, opened 2023-03-06
Optional --binary-endpoint-security under insecure_dev_mode

fantix:
`--security=insecure_dev_mode` modifies the default
`--http-endpoint-security` to `optional`, we should probably do the same to `--binary-endpoint-security` to be consistent, so that `insecure_dev_mode` is the only option to set in order to allow cleartext under any protocols, while we still preserve the options to accept cleartext in a particular protocol (binary or http).


elprans:
Is this needed by something?

fantix:
> Is this needed by something?

Not really, it's just users found binary cleartext not working under `insecure_dev_mode` as its name suggests.

Issue #4952, opened 2023-01-30
Ban `inserts` and `updates` in `filter` and `order by` clauses of `selects`


zchenb:
We should not allow queries like the following:

```
with X := Person select (X {name}, 1) filter {count ((insert Person {name := "t7"})) = 1} ;
```

```
select Foo
filter (insert Bar{val := 1}).val = 1 and false
```

PR #5130, opened 2023-03-06
Trying to enforce no DML constraint in select Where and Orderby Clauses

zchenb:
Fixes https://github.com/edgedb/edgedb/issues/4952

### DIFF
--- a/edb/edgeql/compiler/context.py
+++ b/edb/edgeql/compiler/context.py
@@ -538,7 +538,7 @@ class ContextLevel(compiler.ContextLevel):
     """A ordered set of currently compiling computeds"""
 
     disallow_dml: bool
-    """Whether we are currently in a place where no dml is allowed, 
+    """Whether we are currently in a place where no dml is allowed,
         e.g. the where and order clauses of a select"""
 
     def __init__(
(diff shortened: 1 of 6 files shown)


edb/edgeql/compiler/stmt.py
```diff
@@ -133,10 +133,14 @@ def compile_SelectQuery(
             forward_rptr=forward_rptr,
             ctx=sctx)
 
-        stmt.where = clauses.compile_where_clause(expr.where, ctx=sctx)
+        with sctx.new(context.ContextSwitchMode.NEW) as sctx_no_dml:
+            sctx_no_dml.disallow_dml = True
```
msullivan, line 137: I'd do this in `clauses` itself, for two reasons:
 * A new context is already getting made there, so it'll be a bit faster
 * But more importantly, we want this to be true for where clauses on updates and deletes also

msullivan: And I think limit/offset should get checked too

edb/edgeql/compiler/context.py
```diff
@@ -537,6 +537,10 @@ class ContextLevel(compiler.ContextLevel):
     active_computeds: ordered.OrderedSet[s_pointers.Pointer]
     """A ordered set of currently compiling computeds"""
 
+    disallow_dml: bool
```
msullivan, line 540: Maybe make this an `Optional[str]` so we can say "UPDATE may not be done in a FILTER clause" or whatever

msullivan (commented):
Good start!

edb/edgeql/compiler/stmt.py
zchenb: I think that DMLs in limits and offsets are fine as they only get executed once.

msullivan (approved):
Great!

PR #5133, opened 2023-03-06
Make sys:: and cfg:: fit propertly into the object model

### LINKED ISSUE #4491: Objects in `cfg` and `sys` are in a somewhat busted state
Most of types in `cfg` and `sys` are represented in the database as views that live in the `edgedbss` schema, and which are created in `metaschema.py`.

However the knowledge that they should live in `edgedbss` is somewhat inconsistent.
Duplicates of each of those types exist in `edgedbstd`, and those are used by `BaseObject`s views.
Some of the abstract types in `cfg`/`sys` have views created explicitly in `metaschema`, but some don't.

`cfg` and `sys` special cased in relctx to point at `edgedbss`.

This should all get rationalized, so that one version of each of these tables exists and is routed to in one place (pgsql/common.py, ideally).

What is the reason why these views are in `edgedbss` instead of just `edgedbstd`? Is this important?

Here are some of the bugs that result from this state:
```
edgedb> select cfg::ConfigObject;
edgedb error: InternalServerError: relation "edgedbss.734bd415-4670-11ed-88a5-6d9ebab1a5f7_t" does not exist
```

```
edgedb> select count(BaseObject[is cfg::AbstractConfig]);
[0]
edgedb> select count(cfg::AbstractConfig);
[3]
```

msullivan:
Objects in `sys` and `cfg` are implemented using custom views.  These
views live in the `edgedbss` module and do not get correctly
incorporated into inheritence views for BaseObject (and even for other
abstract objects in the cfg hierarchy).

Move all of those views into `edgedbstd` with the names they ought to
have. Since we run all the schema DDL before running metaschema
scripts, we need to create dummy views in our DDL, which we do by
doing a `select *` from a dummy table that has the DDL run on it.
Then, when running metaschema, we replace those views with our real
ones.

Doing this means we need to take care to preserve the order of
columns, which is annoying but not hard.

Fixes #4491.

Currently objects in sys:: and cfg::, which


msullivan:
Alright, I think this is solid now.

aljazerzen (approved):
Is there a reason these views were in edgedbss in the first place? Just the order of columns?

msullivan:
Elvis said "for historical reasons". I think a big part probably was that this was kind of a pain to get right and the bug it fixed wasn't important

edb/schema/objtypes.py
```diff
@@ -125,8 +125,8 @@ def is_free_object_type(self, schema: s_schema.Schema) -> bool:
 
     def is_fake_object_type(self, schema: s_schema.Schema) -> bool:
         return (
-            self.get_name(schema).module == 'cfg'
-            or self.is_free_object_type(schema)
+            # self.get_name(schema).module == 'cfg' or
```
elprans, line 128: Is this intended?

msullivan: Well, it should have been deleted, not commented out, but yeah

PR #5135, opened 2023-03-06
Add remove data directory command

### LINKED ISSUE #5075: Add `rm-data-dir` command to `edb`
The option should remove the default data directory in the case of version conflict.

zchenb:
Fixes #5075


msullivan:
I think the idea was to add it to the `edb` dev helper commands, not to the real server. `edb/tools/edb.py` would be where to look

zchenb:
Do you mean that I should add it as a command, so I could run `edb rm-data-dir`?

zchenb:
It should be a command. Confirmed in Slack.

zchenb:
Fixed. Please take another look!

1st1 (approved):

PR #5138, opened 2023-03-07
Update PG version

raddevon:


fmoor (approved):
Thanks @raddevon!

msullivan:
In the commit message you merge, please make it clear that that PR is updating the suggested PG version in the DO docs, and not the version of PG that we use internally :P

raddevon:
Ah yes. Will do. Sorry about that, @msullivan.

PR #5139, opened 2023-03-07
Fix DigitalOcean support: allow its custom error in bootstrap

### LINKED ISSUE #5134: EdgeDB doesn't work with managed postgres on Digital Ocean
- EdgeDB Version: 2.12

Steps to Reproduce:

1. Follow the [digital ocean deployment guide using a managed postgres](https://www.edgedb.com/docs/guides/deployment/digitalocean#deploy-with-managed-postgresql) It will fail after setting the backend dsn and restarting edgedb.

fantix:
Fixes #5134 

[DO test](https://github.com/edgedb/edgedb/actions/runs/4357306831)

### DIFF
--- a/edb/server/pgcluster.py
+++ b/edb/server/pgcluster.py
@@ -968,7 +968,9 @@ async def get_remote_pg_cluster(
         except pgcon.BackendPrivilegeError:
             can_make_superusers = False
         except pgcon.BackendError as e:
-            if e.code_is(errors.ERROR_INTERNAL_ERROR):
+            if e.code_is(
+                errors.ERROR_INTERNAL_ERROR
+            ) and "not in permitted superuser list" in str(e):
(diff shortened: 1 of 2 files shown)


edb/server/pgcluster.py
```diff
@@ -966,6 +967,13 @@ async def _detect_capabilities(
             )
         except pgcon.BackendPrivilegeError:
             can_make_superusers = False
+        except pgcon.BackendError as e:
+            if e.code_is(errors.ERROR_INTERNAL_ERROR):
+                # DigitalOcean raises a custom error:
+                # XX000: Role ... not in permitted superuser list
```
elprans, line 973: I think we should also match the error message, since this is very specific behavior and we probably don't want to mask other errors like this.

elprans (approved):

PR #5142, opened 2023-03-08
Fix the leading dot notation in inserts

### LINKED ISSUE #4958: Leading dot notation should not appear in `insert` shapes
The following query should not be compiled and executed successfully:

```
insert Person {name := "<name>", friends := .friends};
```

zchenb:
Fixes #4958


msullivan:
The problem is that while we wanted to disallow these leading dots in the expression itself, we do now allow them in *defaults*

zchenb:
The current semantics seems to be that, all properties have defaults being the empty set unless otherwise specified. Inserts are really inserting the object after applying the insert shape to the default object. That is, `insert A {S}` is really inserting `default(A) {S}`.

zchenb:
This PR is ready for merge!

edb/edgeql/compiler/viewgen.py
```diff
@@ -1268,7 +1269,10 @@ def _compile_qlexpr(
         source_set = setgen.fixup_computable_source_set(
             ir_source, ctx=shape_expr_ctx
         )
-        shape_expr_ctx.partial_path_prefix = source_set
+
+        # always set parital path prefix unless otherwise requested
+        if do_not_set_partial_prefix is False:
```
msullivan, line 1274: `not do_not_set_partial_prefix`

msullivan (approved):
Looks good! Thanks.

PR #5144, opened 2023-03-08
Refactor a bit of SQL introspection metaschema

aljazerzen:
Somehow bootstrap was now failing on master. I suspect #5115.

When I applied this old commit, things started working again. Go figure.


msullivan:
Was it failing consistently? What was the failure?
#5133 is also a solid culprit, if it was a "column does not exist" error. I had nondeterministic errors like that in #5133 but I though I had nailed them all

msullivan (approved):
Changes seem fine, though

aljazerzen:
It turns out it was a little bug in the refactor :D

The old commits was not needed.

PR #5145, opened 2023-03-08
pg-ext: Fix attribute error and simple query sync count

aljazerzen:
* Fix attribute error on SQL statements without a name (caused by #5115)
* Fix bug in simple query sync count (caused by #4957)

### DIFF
--- a/edb/server/compiler/compiler.py
+++ b/edb/server/compiler/compiler.py
@@ -536,7 +536,17 @@ class Compiler:
         sql_units = []
         for stmt in stmts:
             # GOTCHA: a setting is frontend-only regardless of its mutability
-            fe_only = stmt.name in fe_settings_mutable
+            if isinstance(
+                stmt,
+                (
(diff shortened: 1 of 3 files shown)


edb/server/compiler/compiler.py
```diff
@@ -536,7 +536,17 @@ def parse_search_path(search_path_str: str) -> list[str]:
...
-            fe_only = stmt.name in fe_settings_mutable
+            if isinstance(
+                stmt,
+                (
+                    pgast.VariableSetStmt,
+                    pgast.VariableResetStmt,
+                    pgast.VariableShowStmt,
+                ),
+            ):
+                fe_only = stmt.name in fe_settings_mutable
```
fantix, line 547: Ah the `stmt` may not have a `name` 🤦‍♂️ thanks for the fix! Tho perhaps we can push this variable into the `if` branches as it's only needed there

fantix (approved):

PR #5153, opened 2023-03-09
Document nested modules

raddevon:


docs/reference/sdl/modules.rst
```diff
@@ -100,3 +114,11 @@ examples, but it declares module ``my_module`` implicitly:
     }
 
     type my_module::User extending my_module::Named;
+
+
+.. versionadded:: 3.0
+
+   A module block can be nested inside another module block to create a nested
+   module. Fully-qualified names of entities inside the nested block reference
```
msullivan, line 122: I'm having trouble following this sentence

msullivan (approved):

docs/reference/sdl/modules.rst
raddevon: Yeah, it's a bad sentence. 😅 I've rewritten it so that it hopefully makes more sense.

PR #5154, opened 2023-03-09
Document json_object_pack

raddevon:


vpetrovykh (approved):
It's probably worth mentioning what happens to duplicate keys, i.e. the case when more than one tuple has the same string as the first element:
```
db> select json_object_pack({("hello", <json>"world"), ("hello", <json>true)});
{Json("{\"hello\": true}")}
```
It's pretty typical to use the last value in such a case, but I think this is worth stating explicitly.

PR #5155, opened 2023-03-09
Document set intersect, except operators

raddevon:


docs/edgeql/sets.rst
```diff
@@ -248,6 +248,43 @@ Use the :eql:op:`union` operator to merge two sets.
...
+
+Use the :eql:op:`intersect` operator to find common members between two sets.
+
+.. code-block:: edgeql-repl
+
+    db> select {1, 2, 3, 4, 5} intersect {3, 4, 5, 6, 7};
+    {3, 5, 4}
+    db> select {'a', 'b', 'c', 'd', 'e'} intersect {'c', 'd', 'e', 'f', 'g'};
+    {'e', 'd', 'c'}
+
```
vpetrovykh, line 264: It's important to explain what happens in the case of multiple copies of the same elements:
```
db> select {0, 1, 1, 1, 2, 3, 3} intersect {1, 3, 3, 3, 3, 3};
{1, 3, 3}
```
Basically, because the intersection contains only common elements, it means that the minimum number of copies will be present in the output. There's probably a nicer way of phrasing this, though.

docs/edgeql/sets.rst
```diff
@@ -248,6 +248,43 @@ Use the :eql:op:`union` operator to merge two sets.
...
+Use the :eql:op:`except` operator to find members in the first set that do not
+appear in the second set.
+
+.. code-block:: edgeql-repl
+
+    db> select {1, 2, 3, 4, 5} except {3, 4, 5, 6, 7};
+    {1, 2}
+    db> select {'a', 'b', 'c', 'd', 'e'} except {'c', 'd', 'e', 'f', 'g'};
+    {'b', 'a'}
+
```
vpetrovykh, line 284: Same as for the intersection, it's important to show what happens with multiple copies:
```
db> select {0, 1, 1, 1, 2, 3, 3} except {1, 3, 3, 3, 3, 3};
{0, 1, 1, 2}
```
The number of copies that will be excluded depends on how many copies are present in the second set.

docs/edgeql/sets.rst
```diff
@@ -248,6 +248,43 @@ Use the :eql:op:`union` operator to merge two sets.
...
+    {3, 5, 4}
+    db> select {'a', 'b', 'c', 'd', 'e'} intersect {'c', 'd', 'e', 'f', 'g'};
+    {'e', 'd', 'c'}
+
+.. note::
+
+    The ordering of the returned set may not match that of the operands.
+
+
+Finding unique members
```
vpetrovykh, line 270: Surely, it's not about finding unique members as much as removing certain subset from the original set.

raddevon:
@vpetrovykh See if these changes get the job done. I'm not married to any of it, but I have at least covered what the operators due with duplicated members.

docs/edgeql/sets.rst
```diff
@@ -248,6 +248,75 @@ Use the :eql:op:`union` operator to merge two sets.
...
+    {1, 3, 3}
+
+In this example, ``1`` appears three times in the first set but only once in
+the second, so it appears only once in the result. ``3`` appears twice in the
+first set and five times in the second. Both ``3`` appearances in the first set
+are overlapped by ``3`` appearances in the second, so they both end up in the
+resulting set.
+
+
+.. note::
```
msullivan, line 281: My inclination is to drop these notes, since there's no real reason to think that ordering /would/ be preserved, and I worry that noting this might make people assume it is preserved in other cases we don't mention

msullivan (approved):
This looks good to me

docs/edgeql/sets.rst
raddevon: Makes sense. I removed those.

PR #5156, opened 2023-03-09
Document 3.0 unless conflict on multi property

raddevon:


docs/edgeql/insert.rst
```diff
@@ -255,6 +255,11 @@ conflicts and provide a fallback expression.
   apply additional filters on ``Movie``; in the context of the ``else`` clause,
   ``Movie`` is bound to the conflicting object.
 
+.. note::
+
+    Although 2.x does *not* support ``unless conflict`` on :ref:`multi
```
zackelan, line 260: I think this might be more readable if inverted - "using `unless conflict` on multi properties is only supported in 3.0 and later"

zackelan (approved):

msullivan (approved):

msullivan:
Oh, actually, we fixed it in 2.10

raddevon:
Thanks, Sully! I've updated this to reflect it works as of 2.10 instead of as of 3.0.

PR #5157, opened 2023-03-09
Document triggers

raddevon:


raddevon:
One issue with this: we have versioned content in the introspection guide although we had said we won't be versioning guides at all. I believe the solution is going to be to move the introspection guide somewhere else, but I'm collecting some feedback before I make that move. I'll probably need to integrate whatever change we land on into this PR since it's the catalyst for the change.

Any feedback is welcome. My tentatively plan is to move the introspection guide into the "EdgeQL" section of the docs. This will keep a clear separation between the versioned and static documentation.

haikyuu:
Very clear. Spot on 👍

raddevon:
Moved introspection documentation to "Schema" after discussing with @1st1. I have a PR in on the web site repo to redirect the old introspection URLs to the new: https://github.com/edgedb/edgedb.com/pull/581

msullivan (commented):
What is there is pretty good!

I think we also need to touch on `for all` triggers.

raddevon:
> I think we also need to touch on `for all` triggers.

Yes, you're right. I've added an example to cover that.

Cleaned up some other minor things too.

- Moved the note about changing the object that triggered closer to the top of the page
- Changed the `User` type to `Person`
- Spaced out some of the REPL output so it's easier to read on the page, even though it doesn't get output that way by the REPL. (I think readability on the page is more important than 100% accuracy in reproducing the REPL output here.)

raddevon:
Ready for you to take another look when you have a chance, @msullivan. I think this will be a quick one for you.

msullivan (approved):
This looks good!

I think there are two more things that I would note, which aren't in the RFC because they got designed in concert with seeing what was reasonably implementable:
 * If a triggers goes and accesses objects *not* through the `__old__` and `__new__` objects, it will see the state of the database from *after* the query, not before it (So, seeing any changes already made, but not seeing changes that might have been made by another trigger that fired "earlier")
 * Triggers *can* cause other triggers to fire. The way it works is that all the triggers activated by the query itself fire in one "stage", then all the triggers activated by those triggers fire in a second stage, and so on. We reject triggers that lead to a cycle as well as queries that would cause the same trigger to need to fire in multiple "stages"

raddevon:
Thank you, @msullivan!

> * If a triggers goes and accesses objects _not_ through the `__old__` and `__new__` objects, it will see the state of the database from _after_ the query, not before it (So, seeing any changes already made, but not seeing changes that might have been made by another trigger that fired "earlier")

Does this make `__new__.<prop_name>` and `.<prop-name>` equivalent unless the path prefix has changed?

> We reject triggers that lead to a cycle as well as queries that would cause the same trigger to need to fire in multiple "stages"

Is there a case when cycling is not the same as causing the same trigger to fire in multiple stages?

raddevon:
> So, seeing any changes already made, but not seeing changes that might have been made by another trigger that fired "earlier"

Maybe by pointing this out, you mean that `__new__` will show the changes made by prior triggers where not using `__new__` will not?

msullivan:
> Thank you, @msullivan!
> 
> > * If a triggers goes and accesses objects _not_ through the `__old__` and `__new__` objects, it will see the state of the database from _after_ the query, not before it (So, seeing any changes already made, but not seeing changes that might have been made by another trigger that fired "earlier")
> 
> Does this make `__new__.<prop_name>` and `.<prop-name>` equivalent unless the path prefix has changed? Or is `__new__` different in some way? I'd like to contrast the two (those being accessing the object via `__new__` or without it) or let users know they are the same, whatever the case may be.
> 
There is no partial path prefix in triggers.
What I mean is that if you go directly select from some object by name, it sees the state of that object from *after* the query

> > We reject triggers that lead to a cycle as well as queries that would cause the same trigger to need to fire in multiple "stages"
> 
> Is there a case when cycling is not the same as causing the same trigger to fire in multiple stages?

Yeah.
Imagine that there is a trigger on Foo that inserts a Bar, and a trigger on Bar that inserts a Baz, and the main query inserts Foo and Bar.
Then the first stage of triggers will insert Bar and Baz, and the second stage of triggers would need to also insert Baz, but we reject that

msullivan:
> > So, seeing any changes already made, but not seeing changes that might have been made by another trigger that fired "earlier"
> 
> Maybe by pointing this out, you mean that `__new__` will show the changes made by prior triggers where not using `__new__` will not?

No, neither will show it

raddevon:
Would you mind to take a quick look at my new additions here, @msullivan, and make sure they are accurate before I merge? https://github.com/edgedb/edgedb/pull/5157/commits/53e95c4e91aec3425513c6f01a6c4dfdb5341990 and https://github.com/edgedb/edgedb/pull/5157/commits/399a7a79e7cd0bfc3855ff9faaa84aa33e925e6b

docs/reference/sdl/triggers.rst
```diff
@@ -0,0 +1,88 @@
...
+
+:eql:synopsis:`<name>`
+    The name of the trigger.
+
+:eql:synopsis:`insert | update | delete [, ...]`
+    The query type (or types) to trigger on. Separate multiple values with
+    commas to invoke the same trigger for multiple types of queries.
+
+:eql:synopsis:`each`
+    The expression will be evaluted once per modified object. ``__new__`` and
```
zackelan, line 66:
```suggestion
    The expression will be evaluated once per modified object. ``__new__`` and
```

docs/datamodel/triggers.rst
```diff
@@ -0,0 +1,257 @@
...
+
+.. _ref_datamodel_triggers:
+
+========
+Triggers
+========
+
+Triggers allow you to define an expression to be executed whenever a given
+query type is run on an object type. The original query will *trigger* your
+pre-defined expression to run along with the original query. These can be
```
zackelan, line 11: maybe mention here that the triggers are executed in the same transaction, if any, as the triggering query? (obvious to us but worth making clear to users)

raddevon: Good suggestion!

PR #5167, opened 2023-03-13
For scopetree experimentation first three commits

zchenb:


msullivan (approved):

msullivan:
Actually, wait, I wrote all this code so I should review it; @aljazerzen could you take a look?
It's adding some experimental features to `FOR`

edb/edgeql/compiler/inference/multiplicity.py
```diff
@@ -570,7 +570,8 @@ def _infer_for_multiplicity(
     scope_tree: irast.ScopeTreeNode,
     ctx: inf_ctx.InfCtx,
 ) -> inf_ctx.MultiplicityInfo:
-    itset = ir.iterator_stmt
+    # XXX: This is wrong.
```
elprans, line 573: Why, and does this need fixing?

aljazerzen: This just looks at the first iterator and disregards the rest. We'd have to implement multiplicity inference for Cartesian product:

- infer multiplicity for each of the product elements,
- if any elements is EMPTY, multiplicity of the product is also EMPTY,
- if any elements is DUPLICATE, multiplicity of the product is also DUPLICATE,
- otherwise it's UNIQUE.

edb/edgeql/compiler/normalization.py
```diff
@@ -400,6 +401,17 @@ def normalize_ForQuery(
...
+    for field in ('orderby',):
+        value = getattr(node, field, None)
+        _normalize_recursively(
+            node,
+            field,
+            value,
+            schema=schema,
+            modaliases=modaliases,
+            localnames=localnames,
+        )
```
aljazerzen, line 413: What's this? An unrelated fix?

edb/edgeql/compiler/stmt.py
```diff
@@ -216,6 +217,37 @@ def compile_ForQuery(
...
+            raise errors.UnsupportedFeatureError(
+                "Multiple 'FOR' iterators is an internal testing feature",
+                context=qlstmt.context,
+            )
+
+        iterator_stmts = []
+        for binding in qlstmt.iterator_bindings:
+            # XXX: can one iterator reference an earlier one?
+            iterator_stmts.append(
+                _compile_for_binding(qlstmt, binding, ctx=sctx))
```
aljazerzen, line 243: This would be equivalent to `LATERAL` in SQL.

I'd say no, let's keep it simple for now.

edb/edgeql/parser/grammar/expressions.py
```diff
@@ -150,14 +150,35 @@ class GroupingElementList(
     pass
 
 
+class OptionalOptional(Nonterm):
```
aljazerzen, line 153: Hehe, what a confusing congruence.

aljazerzen (approved):
So, `optional` basically says "if this set is empty, for statement should emit a single element: an empty set". The default here would be to emit no elements.

```
edgedb> for a in <Movie>{} union Book { id, x:= a };
{}

edgedb> for optional a in <Movie>{} union Book { id, x:= a };
{default::Book {id: 61837c5b-d46d-11ed-8fd0-27f6f33e0ce5, x: {}}, default::Book {id: 61837c5c-d46d-11ed-8fd0-6b1c67f09aa2, x: {}}}
```

Makes sense. But it's not intuitive (at least not to me).

PR #5171, opened 2023-03-14
Fix dump/restore of migrations with annotations on them

### LINKED ISSUE #5151: Annotations on migrations can't survive dump
- EdgeDB Version: EdgeDB 3.0-dev.7305+local
- EdgeDB CLI Version: 3.0.0-dev

Steps to Reproduce:

1. Run `edgedb migrate --dev-mode` with any schema change
2. Run `edgedb dump ./filename.dump`
```
edgedb error: EdgeQLSyntaxError: error receiving dump header: Unexpected 'generated_by' (on line 1, column 21)
  Server traceback:
        File "/work/edb/common/parsing.py", line 419, in parse
          self.parser.token(token)
        File "/usr/local/lib/python3.10/dist-packages/parsing/lrparser.py", line 57, in token
          self._act(token, tokenSpec)  # type: ignore
        File "/usr/local/lib/python3.10/dist-packages/parsing/lrparser.py", line 81, in _act
          raise UnexpectedToken("Unexpected token: %r" % sym)
      parsing.errors.UnexpectedToken: Unexpected token: <Token IDENT "generated_by">
      
      The above exception was the direct cause of the following exception:
      
        File "/work/edb/server/compiler_pool/worker_proc.py", line 55, in worker
          res = meth(*args)
        File "/work/edb/server/compiler/compiler.py", line 816, in describe_database_dump
          schema_ddl = s_ddl.ddl_text_from_schema(
        File "/work/edb/schema/ddl.py", line 867, in ddl_text_from_schema
          return ddl_text_from_delta(None, schema, diff)
        File "/work/edb/schema/ddl.py", line 785, in ddl_text_from_delta
          return text_from_delta(schema_a, schema_b, delta, sdlmode=False)
        File "/work/edb/schema/ddl.py", line 755, in text_from_delta
          stmts = statements_from_delta(
        File "/work/edb/schema/ddl.py", line 682, in statements_from_delta
          stmts = ddlast_from_delta(
        File "/work/edb/schema/ddl.py", line 663, in ddlast_from_delta
          ql_ast = command.get_ast(schema, context)
        File "/work/edb/schema/delta.py", line 2675, in get_ast
          return self._get_ast(schema, context, parent_node=parent_node)
        File "/work/edb/schema/migrations.py", line 210, in _get_ast
          node = super()._get_ast(schema, context, parent_node=parent_node)
        File "/work/edb/schema/delta.py", line 2993, in _get_ast
          node = super()._get_ast(schema, context, parent_node=parent_node)
        File "/work/edb/schema/delta.py", line 2229, in _get_ast
          self._apply_fields_ast(schema, context, op)
        File "/work/edb/schema/delta.py", line 2264, in _apply_fields_ast
          self._apply_field_ast(schema, context, node, fop)
        File "/work/edb/schema/migrations.py", line 227, in _apply_field_ast
          qlparser.parse_block(op.new_value),
        File "/work/edb/edgeql/parser/__init__.py", line 80, in parse_block
          return parser.parse(source)
        File "/work/edb/common/parsing.py", line 432, in parse
          raise self.get_exception(
      edb.errors.EdgeQLSyntaxError: Unexpected 'generated_by'
```

Migration looks like this:
```
edgedb> select schema::Migration {script} FILTER .name = 'm1d7wj3tklth5kvaxwvnqwvh57ir7b7xni5kkyvjfenlkyo5me2ugq';
{
  schema::Migration {
    script: '                SET generated_by := schema::MigrationGeneratedBy.DevMode;
                ALTER TYPE vSvnXEUmQu::atyyiNKARe {
    CREATE REQUIRED PROPERTY newone: std::duration;
};',
  },
}
```

msullivan:
When we parse the script back, parse it as a migration body, not a
generic block of commands.

Fixes #5151.

Related to this, we drop the DDLStatement enum value for generated_by.
This is because there isn't a good way currently to preserve it across 
dumps. We could add
`SET generated_by := schema::MigrationGeneratedBy.DevMode` to the stored
script, but that cure seems worse than the disease.

There is another potential approach (to both parts of this) that I
might actually prefer, but which has some other downsides: strip `SET
...` commands from the stored text of migration (and only allow them
at the start of the migration). Then we don't need to do anything
special on the parsing side and can easily carry through whatever
state we want.
The downside is that it's a clear compatability break... but only if
there are any users who are setting `message` (which isn't documented)
in a migration.
(If there are, and if we care about them, then dump/restore is broken
for them right now and we'll want to cherry-pick this PR to 2.x to let
them dump).

### DIFF
--- a/edb/edgeql/codegen.py
+++ b/edb/edgeql/codegen.py
@@ -1139,8 +1139,9 @@ class EdgeQLSourceGenerator(codegen.SourceGenerator):
             self.write(self.indent_text(node.body.text))
             self._block_ws(-1)
             self.write('}')
-        elif node.body.commands:
-            self._ddl_visit_body(node.body.commands)
+        elif node.commands or node.body.commands:
+            commands = [*node.commands, *node.body.commands]
+            self._ddl_visit_body(commands)
 
     def visit_StartMigration(self, node: qlast.StartMigration) -> None:
         if isinstance(node.target, qlast.CommittedSchema):
--- a/edb/edgeql/parser/__init__.py
+++ b/edb/edgeql/parser/__init__.py
@@ -80,6 +80,21 @@ def parse_block(source: Union[qltokenizer.Source, str]) -> List[qlast.Base]:
     return parser.parse(source)
 
 
+def parse_migration_body_block(
+    source: str,
+) -> tuple[qlast.NestedQLBlock, list[qlast.SetField]]:
+    # For parser-internal technical reasons, we don't have a
+    # production that means "just the *inside* of a migration block
+    # (without braces)", so we just hack around this by adding braces.
+    # This is only really workable because we only use this in a place
+    # where the source contexts don't matter anyway.
+    source = '{' + source + '}'
+
+    tsource = qltokenizer.Source.from_string(source)
+    parser = qlparser.EdgeQLMigrationBodyParser()
+    return parser.parse(tsource)
+
+
 def parse_sdl(expr: str):
     parser = qlparser.EdgeSDLParser()
     return parser.parse(expr)
--- a/edb/edgeql/parser/grammar/migration_body.py
+++ b/edb/edgeql/parser/grammar/migration_body.py
@@ -1,0 +1,33 @@
+#
+# This source file is part of the EdgeDB open source project.
+#
+# Copyright 2008-present MagicStack Inc. and the EdgeDB authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+from __future__ import annotations
+
+from .expressions import Nonterm
+from .precedence import *  # NOQA
+from .tokens import *  # NOQA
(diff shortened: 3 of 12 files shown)


tailhook:
> Related to this, we drop the DDLStatement enum value for generated_by.

Sorry, I'm not sure what that means. Do you mean that `generated_by` is a string now? It looks like tests still use enum.

Or do you mean that `generated_by` value is just dropped in dump/restore cycle? That would be weird. We can make it a string if that makes it simpler to implement.

> We could add `SET generated_by := schema::MigrationGeneratedBy.DevMode` to the stored script,

It is in the script, currently:
```
  schema::Migration {
    id: 4c6a2843-c255-11ed-a35c-2193c57f7a29,
    name: 'm1egbemkuo4mkf3tnvmo4blvnwnpyhuhc574pqkud5yqohk446joeq',
    internal: false,
    builtin: false,
    computed_fields: [],
    script: 'SET generated_by := schema::MigrationGeneratedBy.DevMode;
CREATE TYPE default::TestType;',
    message: {},
    generated_by: DevMode,
  },
```
Does your change removes that, or am I misunderstanding something?

> only allow them at the start of the migration

Fine by me.

> The downside is that it's a clear compatability break... but only if there are any users who are setting `message` 

Personally I don't think anybody uses messages so far. We don't even display them anywhere (which probably should be improved eventually) and it's undocumented. But if it's just a matter of cherry-picking this PR, we should probably do it.

msullivan:
Sorry, I forgot to post an update on the PR: Per discussions offline with @elprans, I undid the whole `generated_by` thing (just straight reverted it) and then started sticking `SET generated_by := schema::MigrationGeneratedBy.DevMode;`.

I meant "add `SET generated_by := schema::MigrationGeneratedBy.DDLStatement`" to the stored script, which I have now done

tailhook (approved):

elprans (approved):

PR #5177, opened 2023-03-14
Rollback the SQL of EXPLAIN ANALYZE

### LINKED ISSUE #5146: EXPLAIN should rollback any side effects
In simpler words `EXPLAIN ANALYZE INSERT ... ` should not commit anything to the DB.

fantix:
This is only applied on EdgeQL EXPLAIN ANALYZE: an explicit transaction will wrap the analyzing SQL and it's rolled back immediately afterwards; a savepoint is used instead if an outer transaction exists.

Fixes #5146

### DIFF
--- a/tests/test_edgeql_explain.py
+++ b/tests/test_edgeql_explain.py
@@ -488,3 +488,39 @@ class TestEdgeQLExplain(tb.QueryTestCase):
         }
 
         self.assert_plan(res['Plan'], shape)
+
+    async def test_edgeql_explain_insert_01(self):
+        res = await self.explain('''
+            insert User { name := 'Fantix' }
+        ''', analyze=True)
+        self.assert_plan(res['Plan'], {
+            'Node Type': 'Nested Loop',
+        })
+        self.assertFalse(await self.con.query('''
+            select User { id, name } filter .name = 'Fantix'
+        '''))
+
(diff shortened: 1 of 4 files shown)


tests/test_edgeql_explain.py
```diff
@@ -488,3 +488,39 @@ async def test_edgeql_explain_type_intersect_01(self):
         }
 
         self.assert_plan(res['Plan'], shape)
+
+    async def test_edgeql_explain_insert_01(self):
```
msullivan, line 492: Tests run in transactions by default, so this test is still testing the "already in a transaction" flow. 

To test not in a transaction, the test suite needs to be marked `TRANSACTION_ISOLATION = False`. It's probably not worth creating a new test suite for this, though (since it requires a new database to be created, which is slowish), so maybe just wedge it into TestServerProto or something

msullivan (approved):
Great, thank you!

Just one thing about the test to change

tests/test_edgeql_explain.py
fantix: Ahhh right - let me create a ad-hoc connection that is not affected by the transaction isolation

msullivan: Oh even better

msullivan:
Great!

PR #5180, opened 2023-03-14
Infer cardinality of `required multi` pointers as AT_LEAST_ONE

msullivan:
Currently we only actually compute the cardinality when accessing
single pointers, and explicitly choose MANY for multi pointers.

I'm not totally sure why it was like this. I think maybe the skeleton
of this logic predated `required multi`? I don't know.

This bug was originally filed last year and was spuriously attributed
to a query builder bug and moved to edgedb/edgedb-js#333, and a query
builder workaround (insert assert_exists calls) was inserted. It was
identified as being a compiler bug, though---but the issue was never
moved back to the edgedb repo and then the workaround got merged
and the issue was closed.

### DIFF
--- a/edb/edgeql/compiler/inference/cardinality.py
+++ b/edb/edgeql/compiler/inference/cardinality.py
@@ -640,10 +640,7 @@ def _infer_set_inner(
             else:
                 rptrref_card = rptrref.dir_cardinality(rptr.direction)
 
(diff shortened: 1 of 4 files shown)


elprans (approved):
Yeah, this seems like an original bug in `acecddf238`.

Issue #5168, opened 2023-03-13
Potential 3.x migration/ddl perf regression?


msullivan:
I think that I've observed a fairly noticeable perf regression in DDL on master (an extra couple seconds on trivial operations).
I /think/ the time might be spent running reflection queries, but I haven't investigated yet.

msullivan:
The regression was introduced by #5133. It slows down startup time when running my "fasttests" from about 1 minute to about 2.5 minutes, which is quite bad.
The issue is that actually properly merging the `cfg`/`sys` object views into `std::BaseObject` makes explicit id checks much slower (since it needs to go through a bunch of `cfg`/`sys` glue, I think). Since DDL schema update queries do explicit id inserts, this slows down DDL a ton.

In particular, it looks like `select {cfg::ConfigObject, sys::SystemObject}.id` takes about 60ms on my machine, while `select Named.id` on the scope database (which will return all the real objects) takes less than 1ms. 
(I didn't get reasonable results testing with `select Object.id` because there are apparently some test mode only `cfg` objects inheriting directly from `Object` which makes that take about 30ms)

Of course, this is a genuine bug fix: previously objects could be inserted that collided with `cfg`/`sys` objects, though the likelihood of this causing an *actual* problem was low. If we had to pick either the bug or this slowdown, we'd probably take the bug.
Ideally there is some real solution.

In #4412, we briefly started skipping explicit ID checks, but reverted it the next day in #4419 because I realized it could allow collisions with collection types that have deterministic IDs. We fixed that but we actually just allow setting `id` explicitly in DDL. We could get rid of that except for in dumps (and do some other check in dumps). But that would still be regressing explicit id insertion performance substantially for user code.

Hopefully there is an (simple) solution along the lines of making a fast way to query for these ids, but I haven't thought hard about it yet.

msullivan:
I think there might also have been a introspection *read* regression? I'm not sure yet.

msullivan:
Elvis suggested taking advantage of UUID structure: we could use a particular unusual uuid version/variety for all our internal types, and then disallow user specified ids of that sort.

I think the best choice would be "variety 0", which I think is technically what we already use for our builtins.

The problem though is that this is a backward compatibility break, and I think it would be pretty hard to recover from if you had explicitly created objects with variety 0 uuids or whatever. Maybe we need to have a way to disable the check (with unsafe consequences)?

msullivan:
[...]

PR #5181, opened 2023-03-14
Temporarily(?) exclude sys/cfg views from std::BaseObject

msullivan:
Moving the sys/cfg views into std::BaseObject's inhview (in #5133)
caused catastrophic performance regressions on DDL because of DDL's
reliance on explicit id specification. See #5168.

There is plenty of discussion to be had about what we should do about
this (#5168), but first I want to merge this immediately because the
regression makes our development cycle pretty miserable.

We also move some test config objects under std::BaseObject from
std::Object. (Though of course, they won't show up there now.)

This obviously has to xfail the test that we added in #5133.  Add a
new test that tests some of the other cases that #5133 fixed that we
still support. (Making sure that the abstract parents in sys and cfg
work instead of producing ISEs when used.)

### DIFF
--- a/edb/lib/_testmode.edgeql
+++ b/edb/lib/_testmode.edgeql
@@ -20,14 +20,14 @@
 # These definitions are picked up if the EdgeDB instance is bootstrapped
 # with --testmode.
 
-CREATE TYPE cfg::TestSessionConfig {
+CREATE TYPE cfg::TestSessionConfig EXTENDING std::BaseObject {
     CREATE REQUIRED PROPERTY name -> std::str {
         CREATE CONSTRAINT std::exclusive;
     }
 };
 
 
-CREATE ABSTRACT TYPE cfg::Base {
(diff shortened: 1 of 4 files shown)


edb/lib/_testmode.edgeql
```diff
@@ -20,14 +20,14 @@
 # These definitions are picked up if the EdgeDB instance is bootstrapped
 # with --testmode.
 
-CREATE TYPE cfg::TestSessionConfig {
+CREATE TYPE cfg::TestSessionConfig EXTENDING std::BaseObject {
```
elprans, line 23: maybe

```suggestion
CREATE TYPE cfg::TestSessionConfig EXTENDING cfg::ConfigObject {
```

?

elprans (approved):

edb/lib/_testmode.edgeql
msullivan: Yeah, good call

PR #5184, opened 2023-03-15
Normalize all fields in explain to snake_case

tailhook:


edb/server/compiler/explain.py
```diff
@@ -48,6 +48,7 @@
     rf'(\.?"?({uuid_core})"?)',
     re.I
 )
+case_re = re.compile(r'(?<!^)(?<!\s|-)[\s-]*(?=[A-Z])')
```
msullivan, line 51: Could you add a comment explaining this regex, since I don't understand it at a glance

tailhook: Added some reasoning. If anyone wants to piece-by-piece description, AI does that well :)

PR #5188, opened 2023-03-15
Make uuidgen properly set versions in uuid4/uuid5

msullivan:


elprans (approved):

msullivan:
I think this will change shape ids, which might invalidate some caches that
client lib codegens use?
That seems fine for a major version though

PR #5189, opened 2023-03-15
Forbid `on target delete deferred restrict` on required links.

### LINKED ISSUE #4984: Disallow `deferred restrict` on required links
If `on target delete deferred restrict` is specified on a required link and the target is deleted, the link will be empty in subsequent queries in the same transaction, which violates our model.

I can think of two solutions:
1. Disallow this combination
2. Track what types may have had objects deleted by earlier queries in the transaction and insert `assert_exists` around accesses to them, like we do when something may be filtered by an access policy.

I think option 1 seems clearly the right thing here

efaust:
We want to forbid `on target delete deferred restrict` for required links: it
allows intermediate queries to observe the NULL, which breaks the schema's
invariants.

Actually doing that is trivial: just check and throw.

HOWEVER, we were using it a bunch of places in the standard library.

Clean up a bunch of them with the following steps:

- Defer if_unused subcommands for object deletions, to try to ensure that the
  generated schema types are deleted after the containing objects.
- Change @target deletetion handling, by moving it directly to DeletePointer,
  cleaning up some cruft between DeleteLink and DeleteProperty in the process.

STILL TODO:
 - Aliases are a mess. Just punt for aliases and globals, instead marking the
   links optional for now.

Fixes #4984.

### DIFF
--- a/edb/lib/schema.edgeql
+++ b/edb/lib/schema.edgeql
@@ -130,9 +130,7 @@ CREATE ABSTRACT TYPE schema::CollectionType EXTENDING schema::PrimitiveType;
 
 
 CREATE TYPE schema::Array EXTENDING schema::CollectionType {
-    CREATE REQUIRED LINK element_type -> schema::Type {
-        ON TARGET DELETE DEFERRED RESTRICT;
-    };
+    CREATE REQUIRED LINK element_type -> schema::Type;
     CREATE PROPERTY dimensions -> array<std::int16>;
 };
 
@@ -141,9 +139,7 @@ CREATE TYPE schema::ArrayExprAlias EXTENDING schema::Array;
 
 
 CREATE TYPE schema::TupleElement EXTENDING std::BaseObject {
-    CREATE REQUIRED LINK type -> schema::Type {
-        ON TARGET DELETE DEFERRED RESTRICT;
-    };
+    CREATE REQUIRED LINK type -> schema::Type;
     CREATE PROPERTY name -> std::str;
 };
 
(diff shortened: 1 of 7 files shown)


edb/schema/links.py
```diff
@@ -288,6 +288,17 @@ def validate_object(
                 context=srcctx,
             )
 
+        if (
+            self.get_attribute_value('required') and
+            self.get_attribute_value('on_target_delete') ==
+                qltypes.LinkTargetDeleteAction.DeferredRestrict
+        ):
+            raise errors.InvalidLinkTargetError(
+                'required link illegally defined `on target delete '
```
msullivan, line 297: How about "required links may not use ..."

edb/schema/links.py
```diff
@@ -288,6 +288,17 @@ def validate_object(
                 context=srcctx,
             )
 
+        if (
+            self.get_attribute_value('required') and
```
msullivan, line 292: Check the values on `scls` with `get_required`/etc. That will ensure it works in all cases, regardless of what things are specified in the actual delta

edb/lib/schema.edgeql
```diff
@@ -322,8 +312,8 @@ ALTER TYPE schema::Source {
 CREATE TYPE schema::Alias EXTENDING schema::AnnotationSubject
 {
     CREATE REQUIRED PROPERTY expr -> std::str;
-    CREATE REQUIRED LINK type -> schema::Type {
-        ON TARGET DELETE DEFERRED RESTRICT;
+    CREATE OPTIONAL LINK type -> schema::Type {
```
msullivan, line 315: Add comments explaining that this is a hack and it is not *really* optional

msullivan (approved):
Looks good!

Left a few suggestions; go ahead and merge once you've addressed them

PR #5197, opened 2023-03-16
Add support for new JWT layout

elprans:
Add support for JWT version "1", which adds support for database and
instance scoping and shortens claim names.


edb/server/cluster.py
```diff
@@ -418,6 +421,38 @@ async def _new_pg_cluster(self) -> pgcluster.Cluster:
...
+        jws_key_file = self._get_jws_key_path()
+        try:
+            with open(jws_key_file, 'rb') as kf:
+                jws_key = jwk.JWK.from_pem(kf.read())
+        except Exception as e:
+            raise ClusterError(f"cannot load JWS key: {e}") from e
+
+        if (
+            not jws_key.has_public
+            or jws_key['kty'] not in {"RSA", "EC"}
```
tailhook, line 442: Just curious, why we have two kinds of keys supported?

tailhook (approved):
Overall, how we determine that token belongs to the right organization/account? Do every organization have their own key pair?

edb/server/protocol/binary.pyx
```diff
@@ -449,26 +452,68 @@ cdef class EdgeConnection(frontend.FrontendConnection):
...
+            namespace = "edgedb.server"
+            if not claims.get(f"{namespace}.any_role"):
+                token_roles = claims.get(f"{namespace}.roles")
+                if not isinstance(token_roles, list):
+                    raise errors.AuthenticationError(
+                        f'authentication failed: malformed claims section in'
+                        f' JWT: expected a list in "{namespace}.roles"'
+                    )
+            else:
+                token_roles = None
```
fantix, line 483: Redundant

```suggestion
```

edb/server/protocol/binary.pyx
```diff
@@ -449,26 +452,68 @@ cdef class EdgeConnection(frontend.FrontendConnection):
...
+
+        if (
+            token_instances is not None
+            and self.server.get_instance_name() not in token_instances
```
fantix, line 487: Similar (future) question as @tailhook 's, regardless of this PR tho: instance name is currently set without the org slug in the cloud for now, this allows token survive transferring instance among orgs, but not renaming the instance within the org. I noticed that instance name in the token can be a list (to support multiple instances of course) - this happens to allow creating new tokens to also support old servers lingering before the renaming is complete, but not old token on servers with new instance names after renaming. Q: do we want the tokens to be still valid after moving/renaming instances?

fantix (approved):
Except for `aud` in HTTP SCRAM, I'm good with this.

pyproject.toml
```diff
@@ -4,7 +4,7 @@ description = "EdgeDB Server"
 requires-python = '>=3.10.0'
 dynamic = ["entry-points", "version"]
 dependencies = [
-    'edgedb==1.1.0',
+    'edgedb~=1.3.0',
```
fantix, line 7: There is another `edgedb` under `build-system.requires`.

edb/server/protocol/binary.pyx
tailhook: > do we want the tokens to be still valid after moving/renaming instances?

I think in most cases instance names and tokens will either be stored together, or instance name would be harder to update than token (i.e. because instance name is in static environment, but token is in secrets). So I don't think we should optimize for that use case.

fantix: Right, in that case maybe we should just add the full instance name into the token like `org_slug/inst_name` instead of just `inst_name` 🤔 (but this is a cloud thing, out of this PR)

edb/server/protocol/binary.pyx
```diff
@@ -449,26 +452,68 @@ cdef class EdgeConnection(frontend.FrontendConnection):
...
-                    f' expected mapping in "role_names"'
-                )
+        self._check_jwt_authz(claims, token_version, user)
+
+    def _check_jwt_authz(self, claims, token_version, user):
+        token_instances = None
+        token_roles = None
+        token_databases = None
+
+        if token_version == 1:
```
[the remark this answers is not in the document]
elprans: There is no need for long backwards compat here as there are no cloud tokens in the wild yet, and server-generated ones are generated by the same server.

edb/server/protocol/binary.pyx
elprans: Yes, we'll use the full name.

edb/server/cluster.py
elprans: Historical reasons, but we mostly settled on EC now, so I guess we can remove the RSA option.

elprans:
> Overall, how we determine that token belongs to the right organization/account?

This will be added in a follow-up PR.  These changes are mostly not Cloud-specific.

edb/common/secretkey.py
```diff
@@ -0,0 +1,60 @@
...
+    skey: jwk.JWK,
+    *,
+    instances: Optional[list[str] | AbstractSet[str]] = None,
+    roles: Optional[list[str] | AbstractSet[str]] = None,
+    databases: Optional[list[str] | AbstractSet[str]] = None,
+) -> str:
+    claims = {
+        "iat": int(datetime.now(timezone.utc).timestamp()),
+        "iss": "edgedb-server",
+        "aud": "edgedb tests",
```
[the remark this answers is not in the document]
elprans: Ah, I don't think we really need it at all

Neustradamus:
Good job!

Linked to:

PR #5198, opened 2023-03-16
Minor EXPLAIN improvements

msullivan:
* Always compile the inner EXPLAIN query in binary mode
 * Expose key configuration settings and what globals are used in EXPLAIN


elprans (approved):

edb/server/compiler/explain.py
```diff
@@ -416,8 +422,15 @@ def analyze_explain_output(
 
     collapse_plan(plan)
 
+    config_vals = {
+        k: v for k, v in config_vals.items() if k not in OMITTED_CONFIG_VALS
```
1st1, line 426: Maybe having a list of config names we want to include is a better strategy?

msullivan: I figured that a new option we will probably be more likely to want to include it than to not include it, and that we are likely to forget either way.

Issue #5118, opened 2023-03-01
Our ancestor resolution order isn't *quite* C3


msullivan:
I think that our ancestor ordering is intended to be the C3 linearization ordering (https://en.wikipedia.org/wiki/C3_linearization), which is what Python uses, but we don't include the list of base classes in the list of orders to merge.

I believe this results in us not always having the second of the C properties, "preservation of local precedence order" (which makes our algorithm one of several possible "C2" algorithms :P).

I think, though I may be wrong, that the cases in which we *don't* preserve local precedence order are all cases that would be rejected by the C3 algorithm.

I'd like to fix this, but it might not be worth the compatibility break. No idea how common this will be in user code. Probably not that common.

```
with X := schema::InheritingObject
select X {
  name,
  bases: {name, @index} order by @index,
  ancestors: {name, @index} filter X.ancestors in X.bases order by @index,
  mismatch := (array_agg((select .bases order by @index).name) != array_agg((select .ancestors {name} filter X.ancestors in X.bases order by @index).name)),
}
filter .mismatch;
```
will find all the object types in a schema that don't obey the ordering.

There are a few in the schema, and most seem to be caused by the reflection machinery adding in base classes that weren't written in the schema and not being careful about where in the list they get added.

elprans:
Ideally we should document and fix this.  The release notes can have a section on porting and we can go as far as introducing a pre-flight check for CLI-driven upgrades using the introspection query above.

msullivan:
Well, this'll be a good chance to put our new tools for rewriting and squashing migrations into practice, then :)

msullivan:
(Also that query takes almost 20s, which is way more than makes sense to me)

PR #5204, opened 2023-03-17
Untangle the inheritance hierachies in the standard library

msullivan:
Currently there are a bunch of places where we have inheritance orders
that are not valid C3 orders, because the order bases are specified is
not consistent with the inheritance hierarchies of those bases.

For several, this is just because things are broken in the schema. For
many, it is because the code in reflection that computes how to update
the bases was incorrectly figuring out how to combine the python
orders and the edgeql orders. Typically I fixed those by making
the python mirror the schema better.

Work towards fixing #5118.

### DIFF
--- a/edb/lib/schema.edgeql
+++ b/edb/lib/schema.edgeql
@@ -277,7 +277,6 @@ CREATE ABSTRACT TYPE schema::Source EXTENDING schema::Object {
 CREATE ABSTRACT TYPE schema::Pointer
     EXTENDING
         schema::ConsistencySubject,
-        schema::InheritingObject,
         schema::AnnotationSubject
 {
     CREATE PROPERTY cardinality -> schema::Cardinality;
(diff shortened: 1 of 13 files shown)


edb/lib/schema.edgeql
```diff
@@ -275,7 +276,8 @@ CREATE ABSTRACT TYPE schema::Source EXTENDING schema::Object {
 
 CREATE ABSTRACT TYPE schema::Pointer
     EXTENDING
-        schema::InheritingObject, schema::ConsistencySubject,
+        schema::ConsistencySubject,
+        schema::InheritingObject,
```
elprans, line 280: You made `ConsistencySubject` an `InheritingObject`, so perhaps drop the entry here?

edb/lib/schema.edgeql
```diff
@@ -323,8 +325,10 @@ CREATE TYPE schema::Alias EXTENDING schema::AnnotationSubject
 
 CREATE TYPE schema::ScalarType
     EXTENDING
-        schema::InheritingObject, schema::ConsistencySubject,
-        schema::AnnotationSubject, schema::PrimitiveType
+        schema::PrimitiveType,
+        schema::ConsistencySubject,
+        schema::InheritingObject,
```
elprans, line 330: Ditto

```suggestion
```

elprans (approved):

PR #5210, opened 2023-03-19
Adding `--use-experimental-interpreter` to `edb test`, which will divert the query engine to the experimental interpreter

zchenb:


zchenb:
This does not merge into master but my experimental interpreter branch. The only relevant changes are to `server.py` and `runner.py`. My experimental interpreter will eventually be merged into master, it is nevertheless good to review those portions that change the main codebase.

PR #5215, opened 2023-03-21
Make internal trigger names non-obfuscated

msullivan:
Currently they are directly produced by get_backend_name by passing in
all the extra stuff as an aspect, which trips us over the length
limit, so the name gets hashed. That length limit doesn't actually
apply to triggers and functions so just produce the name directly.

This will produce a marginal increase in my quality of life when
trying to debug things related to our triggers.


elprans (approved):
LGTM, but perhaps a more thorough fix would be to add an argument to `get_backend_name` to disable obfuscation.  It'd definitely be a more annoying changeset, so I'm not insisting.

PR #5217, opened 2023-03-21
Add new SDL syntax

raddevon:
This is a big PR. The good news is that most of it consists of mirroring the existing SDL examples in the new syntax and marking the pre-3.0 syntax examples with `:version-lt: 3.0`.

Other stuff that's included:
- making sure all the code blocks I touched are indented by 4 spaces
- unwrapping "link" as inline code since the keyword is no longer required in SDL in most cases (in datamodel/links.rst for example)
- adding descriptions of the new syntax in prose where that made sense
- adding new 3.0 SDL synopses for links and properties
- fixes for very minor unrelated issues I found
  - `#New` label switched to a `versionadded` directive in `datamodel/constraints.rst` "Partial constraints" heading
  - various other spacing fixes

Stuff that *isn't* included:
- New SDL examples for the introspection guide
  - The introspection guide is being moved as part of #5157. I'm afraid of what the merge might be like if I also make changes to it in this PR, so I will circle back after that PR is merged to add new SDL examples there.
- Updated SDL in other guides
  - Guides will not be versioned, so we will circle back to the few SDL samples in them and update them to the new syntax once 3.0 launches.

### DIFF
--- a/docs/datamodel/access_policies.rst
+++ b/docs/datamodel/access_policies.rst
@@ -13,16 +13,27 @@ This is known as *object-level security*.
 Let's start with a simple schema.
 
 .. code-block:: sdl
+    :version-lt: 3.0
 
-  type User {
-    required property email -> str { constraint exclusive; };
-  }
+    type User {
+      required property email -> str { constraint exclusive; };
+    }
 
-  type BlogPost {
-    required property title -> str;
-    required link author -> User;
-  }
+    type BlogPost {
+      required property title -> str;
+      required link author -> User;
+    }
 
+.. code-block:: sdl
+
+    type User {
+      required email: str { constraint exclusive; };
+    }
+
+    type BlogPost {
+      required title: str;
+      required author: User;
+    }
 
 When no access policies are defined, object-level security is not activated.
 Any properly authenticated client can select or modify any object in the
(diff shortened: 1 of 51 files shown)


Dhghomon:
Not intimately familiar with the new SDL and doc formatting but I did go over every change and besides the one comment I wrote last week I didn't see anything that looked out of the ordinary!

msullivan (approved):
This seems reasonable to me, though obviously all the duplication is pretty tragic.
(Is having only one master copy of the documentation actually a good idea??)

Just to confirm, only one copy of the blobs will get displayed on a page

raddevon:
> This seems reasonable to me, though obviously all the duplication is pretty tragic.

Yeah, we went back and forth on this. @jaclarke started building something that would automatically update SDL syntax when the docs version selector was changed, but I believe that got complicated. We decided to do it the simple way in response to that, which caused the duplication but also gives us more control over what we want to display.

This had the positive side-effect of giving us a way to conditionally display different synopses across versions too which we would have needed anyway for cases like the addition of `errmessage` for access policies.

> Is having only one master copy of the documentation actually a good idea??

I'm not sure what you mean here, @msullivan.

> Just to confirm, only one copy of the blobs will get displayed on a page

Yes, that's right. The docs will only display one code-block per grouping based on the user's version selection.

PR #5220, opened 2023-03-21
Doc v3 access policy errors

raddevon:
Also cleaned up a couple of warnings that access policies are a 2.0 feature by replacing those with the `versionadded` directive.


aljazerzen (approved):
LGTM

PR #5225, opened 2023-03-21
Allow specifying EXTENDING as a pointer subcommand.

### LINKED ISSUE #5203: Implement extending as a pointer subcommand
We've agreed that we want to support syntax like

```
type A;
abstract link Bar;
type B {
    link Foo -> A {
        extending Bar;
    };
```

We should do that.

efaust:
When we changed from -> to : for pointer type spacing, we were left with a bunch of ugly cases.

`CREATE LINK Foo EXTENDING default::Bar, default::Baz: default::Target;`

Instead, allow specifying the EXTENDING information as a subcommand:

```
CREATE LINK Foo: default:Target {
    EXTENDING default::Bar, default::Baz;
};
```

Also fixup SDL, Abstract links, and a bunch of tests.

Switch pretty printing to always print in this mode.

Fixes #5203.


edb/edgeql/codegen.py
```diff
@@ -837,6 +837,24 @@ def _ddl_visit_bases(self, node: qlast.BasesMixin) -> None:
...
+        qlast.CreateLink,
+        qlast.CreateProperty
+    )
+
+    def _ddl_add_bases_command(
+        self,
+        node: PointerNode,
+    ) -> PointerNode:
+        if node.bases:
+            return node.replace(commands=(  # type: ignore
```
msullivan, line 852: OK I think the actual fix for this is to update the definition of `replace` in `edb/common/ast/base.py` to have a generic type annotation on `self` and the return type.

edb/edgeql/codegen.py
```diff
@@ -837,6 +837,24 @@ def _ddl_visit_bases(self, node: qlast.BasesMixin) -> None:
...
             self.visit_list(node.bases, newlines=False)
 
+    PointerNode = TypeVar(
+        'PointerNode',
+        qlast.CreateConcretePointer,
+        qlast.CreateLink,
+        qlast.CreateProperty
+    )
+
+    def _ddl_add_bases_command(
```
msullivan, line 847: add_pointer_bases maybe?

edb/edgeql/parser/grammar/sdl.py
```diff
@@ -943,17 +943,30 @@ def reduce_CreateOverloadedQualifiedPointer(self, *kids):
 #
 # Properties
 #
+sdl_commands_block(
```
msullivan, line 946: I think we also need to handle UnknownPointer

tests/test_schema_syntax.py
```diff
@@ -1346,22 +1419,70 @@ def test_eschema_syntax_link_12(self):
...
+                extending intersect;
                 property except: str;
             };
         };
         """
 
+    def test_eschema_syntax_link_15(self):
+        """
+        module test {
+            abstract link foo {
```
msullivan, line 1444: Hm, I think we could have skipped doing this for abstract links themselves and only done it for concrete ones. I don't have a super strong feeling though.

tests/test_schema_syntax.py
```diff
@@ -1346,22 +1419,70 @@ def test_eschema_syntax_link_12(self):
...
+            };
+        };
+
+        """
+
+    def test_eschema_syntax_link_18(self):
+        """
+        module test {
+            type foo {
+                link bar: test::foo {
```
msullivan, line 1478: Also add a test for the case where `link ` is omitted. (I think it will need to be implemented also)

PR #5226, opened 2023-03-21
Add support for JWT allowlist and revocation list

elprans:
The new `--jwt-sub-allowlist-file` server argument allows providing a
list of subjects whose keys are authorized, and the new
`--jwt-revocation-list-file` server argument allows providing a list of
key ids that are explicitly revoked.


fantix (approved):
LGTM!

PR #5227, opened 2023-03-21
Support link properties on computed backlinks

### LINKED ISSUE #4825: Support link properties on computed backlinks
In our `cards` test database, 
```
select Card { owners: { @count } };
```
ought to work

(Currently we can access properties on backlinks when they are spelled out explicitly, but not via computeds)

msullivan:
Replicating the machinery that non-schema computed backlinks use
(faking inheritance to get the link properties inherited, then undoing
it) turned out to require a huge amount of pain to transplant to a
schema context, so we dropped that whole system and instead check for
link properties on computed backlinks dynamically at compile time.

Fixes #4825.

### DIFF
--- a/edb/edgeql/compiler/schemactx.py
+++ b/edb/edgeql/compiler/schemactx.py
@@ -325,9 +325,15 @@ def derive_ptr(
     if ptr.get_name(ctx.env.schema) == derived_name:
         qualifiers = qualifiers + (ctx.aliases.get('d'),)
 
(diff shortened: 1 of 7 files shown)


msullivan:
It might not be obvious from first glance but this PR I think does substantially reduce jank (or at worst is jank-neutral). Faking inheritance to populate the backlink with the right link properties really was unjustifiable

elprans (approved):
Makes sense

PR #5228, opened 2023-03-22
Make schema changes needed for rewrites

msullivan:
Pulled out of the pending rewrites work (#5006) to reduce schema
change related development grief.

### DIFF
--- a/edb/schema/pointers.py
+++ b/edb/schema/pointers.py
@@ -46,6 +46,7 @@ from . import inheriting
 from . import name as sn
 from . import objects as so
 from . import referencing
+from . import rewrites as s_rewrites
 from . import schema as s_schema
 from . import types as s_types
 from . import utils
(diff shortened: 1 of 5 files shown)


edb/schema/pointers.py
```diff
@@ -534,6 +535,22 @@ class Pointer(referencing.NamedReferencedInheritingObject,
...
     )
 
+    rewrites_refs = so.RefDict(
+        attr="rewrites",
+        requires_explicit_overloaded=True,
+        backref_attr="subject",
+        ref_cls=s_rewrites.Rewrite,
+    )
+
+    rewrites = so.SchemaField(
```
elprans, line 545: Add to `schema.edgeql` or not yet?

elprans (approved):

edb/schema/pointers.py
msullivan: Wasn't done in the PR yet; isn't needed yet

PR #5237, opened 2023-03-23
Document `assert`

raddevon:


raddevon:
I'm not super confident in this one. I think the baseline is covered, but there's probably more to do here. Might lean on your feedback to point me in the right direction, @vpetrovykh.

msullivan:
Damn it, I just noticed that the commit/PR message in the implementation PR https://github.com/edgedb/edgedb/pull/5040 is very confusing. This is because I wrote it all as a git commit message but I put github markdown section headers in it. Except those are git commit message comments and they all got removed.

I just updated that PR to have the section headers I meant to. Could you take a look at that and probably write something about the triggers case, at least?

Thanks, and sorry!

raddevon:
Of course, I can take another pass! No worries, @msullivan.

raddevon:
Your updated commit message was really helpful, @msullivan! I reworked this just a bit based on your changes. I added a trigger example, and I tried to clarify the usefulness around using `assert` based on my revised understanding after reading your message.

I nearly included something from your third heading, but it seemed like it came with several caveats. My thinking is that I might leave that example out until we have `if` in DML. Sounds like that will make it much more straightforward. I figure we can direct people to your commit message in the meantime. If demand for that usage is high, I'll circle back and add something to the docs about it sooner.

Let me know if this doesn't work, and I can take another pass.

raddevon:
Once triggers is merged, I may also circle back and add the assert example (or a similar one) to that documentation as well.

docs/stdlib/bool.rst
```diff
@@ -220,3 +222,114 @@ any(a) or any(b)``.
...
+
+    .. code-block:: sdl
+        
+          type Person {
+            required name: str;
+            multi friends: User;
+            multi enemies: User;
+
+            trigger prohibit_frenemies after insert, update for each do (
+              for obj in __new__ union (
```
msullivan, line 267: I'm not totally sure why I did it with a for loop in the commit-message example, but I think it would be better to skip the for loop and just use `__new__` directly in the trigger body.
(Putting a for loop in would be necessary if you are doing `for all`, but I don't think there's a good reason to do it that way.)

msullivan (approved):

PR #5250, opened 2023-03-24
Added system config statement_cache_size

### LINKED ISSUE #5233: Add a knob to control the size of Postgres prepared statement cache
Needed to help balance performance with memory usage of the Postgres backend.  I think this should be an instance-level configuration variable.

fantix:
Used to control the size of Postgres prepared statement cache.

Fixes #5233

### DIFF
--- a/edb/edgeql/compiler/config.py
+++ b/edb/edgeql/compiler/config.py
@@ -52,6 +52,7 @@ from . import setgen
 from . import compile_ast_to_ir
 from . import options
 
+
 class SettingInfo(NamedTuple):
     param_name: str
     param_type: s_types.Type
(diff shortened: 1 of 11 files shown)


elprans (commented):
LGTM, but I'd like a better name, e.g `pg_prepared_statement_cache_size`

fantix:
Do we want to use "backend" instead of "pg"? But I also understand this config name is getting really long 🚄🚋🚋🚋🚋🚋🚋🚋

elprans:
> Do we want to use "backend" instead of "pg"? 

In this case this is specifically about the Postgres implementation, so I'd use `pg_`.  Likewise, other exposed Postgres settings should ideally be prefixed, but let's leave that to when we'll properly implement config vars (a-la global vars), which will bring natural namespacing via modules.

PR #5252, opened 2023-03-27
Add context (code location) to pointers in IR

tailhook:
This fixes issue with explain query like this:
```
SELECT { title, avg_rating }
		^^^^^^^^^^ --- computed property now having context
```
(works on schema from [imdbench](https://github.com/edgedb/imdbench/blob/c1f9a4/dbschema/default.esdl))


tailhook:
Looks like it does something weird. Have to take a closer look.

Issue #5234, opened 2023-03-22
Add a way to specify instance-level config parameters as command-line arguments / environment variables


elprans:
Related: #5234

PR #5266, opened 2023-03-29
Specify instance-level config parameters as CLI arguments / env vars

fantix:
Fixes https://github.com/edgedb/edgedb/issues/5234

Examples:

```
edgedb-server --cfg-session-idle-timeout 2m18s --cfg-no-apply-access-policies
```
```
EDGEDB_SERVER_CFG_APPLY_ACCESS_POLICIES=false edgedb-server
```

### [Add new config layers through _edgecon_state](https://github.com/edgedb/edgedb/pull/5266/commits/c67f2dd45fc2db87df932859a221d1cdd5d5b753) 

The new "environment variable" and "command line" layers of config are
set when starting the EdgeDB server through set_init_con_script_data(),
and keeps static for the lifetime of the server process.

Because one Postgres may have multiple EdgeDB servers, the temporary
table _edgecon_state is reused to bridge the per-instance config set in
environment variables or command-line arguments into the backend.

Also fixed so that INIT_CON_SCRIPT is only built when apply_init_script.

### [Take and use cfg args from command-line and env](https://github.com/edgedb/edgedb/pull/5266/commits/48fdc3eb102831f672f093d4e4606bc64a7fabc9) 

The extra command-line arguments are parsed after pg cluster initialization
when we have the config spec. If any of these config mirrors a backend
setting, the pg cluster will be restarted to take that setting as a
command-line argument, in order to match the priority of the origin.


edb/pgsql/metaschema.py
```diff
@@ -3107,11 +3107,13 @@ def __init__(self) -> None:
...
                 'postgres environment variable',
                 'postgres configuration file',
+                'environment variable',
+                'command line',
                 'postgres command line',
                 'postgres global',
+                'postgres client',
                 'system override',
                 'database',
-                'postgres client',
```
fantix, line 3114: I think `postgres client` was after `system override` and `database` because we don't want users to override our server-specific postgres config set at connection time like `client_encoding` and `timezone`. But with this PR on particularly remote Postgres, we use the same `postgres client` config for those `backend_setting` set in command-line/env, they are meant to be override by `system override` or `database`. Because we didn't expose `client_encoding` or `timezone` as EdgeDB config, I hope it is fine to change the priority of `postgres client`.

edb/server/main.py
```diff
@@ -687,10 +700,41 @@ def server_main(**kwargs):
             asyncio.run(run_server(server_args))
 
 
+class CfgArgsGroup(click.Group):
+    # The default click.Group cannot simply take remaining arguments because
+    # it won't pass the subcommand check; on the other hand, an argument with
+    # `nargs=-1` would cripple subcommands. So we patch click.Group here to
+    # support both, with a drawback that extra arguments must all come together
+    # in the end.
```
elprans, line 708: What subcommands are you referring to here?

edb/server/main.py
```diff
@@ -719,5 +763,114 @@ def main_dev():
...
+        })
+        if setting.backend_setting:
+            if isinstance(value, statypes.ScalarType):
+                value = value.to_backend_str()
+            backend_settings[setting.backend_setting] = str(value)
+
+    env_value: Any
+    setting: config.Setting
+    for env_name, env_value in os.environ.items():
+        cfg_name = env_name.removeprefix("EDGEDB_SERVER_CFG_")
```
elprans, line 810: There's an issue of namespacing the config vars.  If we only allow builtin configuration values to be passed this way, then it's probably fine, but if we generalize it to namespaced configuration variables, then a different scheme is needed, e.g `EDGEDB_SERVER_CONFIG_<module>_<name>`.  The downside is that for builtin vars this will result in `EDGEDB_SERVER_CONFIG_CFG_FOO`, which is a bit awkward looking.

fantix: Conclusion from the meeting:

1. For env vars, we prefix whatever qualified config names with `EDGEDB_SERVER_CONFIG_`, for example, `EDGEDB_SERVER_CONFIG_cfg::Config.session_idle_timeout=10m`
2. We'll do the same for command-line arguments, like `--config-cfg::Config.session_idle_timeout 10m`
3. To avoid duplicate config names caused by case sensitivity, we may add a unique constraint on the upper/lower-cased config names.
4. To avoid invalid config names caused by spaces or equal signs, we could add another constraint to either forbid those characters, or allow ASCII only.

edb/server/main.py
```diff
@@ -810,12 +781,22 @@ def add_config(name, value, type_):
...
+    def iter_environ():
+        translate_env = {
+            "EDGEDB_SERVER_BIND_ADDRESS": "listen_addresses",
+            "EDGEDB_SERVER_PORT": "listen_port"
+        }
+        for name, value in os.environ.items():
+            if cfg := translate_env.get(name):
+                yield name, value, cfg
+            else:
+                cfg = name.removeprefix("EDGEDB_SERVER_CONFIG_cfg::Config.")
```
elprans, line 793: We are planning to deprecate `Config` objects and replace with top-level variables instead, so the variables should be `EDGEB_SERVER_CONFIG_cfg::apply_access_policies`.  IOW, remove `Config.`

edb/buildmeta.py
```diff
@@ -44,7 +44,7 @@
 
 
 # Increment this whenever the database layout or stdlib changes.
-EDGEDB_CATALOG_VERSION = 2023_03_30_00_01
+EDGEDB_CATALOG_VERSION = 2023_05_03_00_01
```
elprans, line 47: Why a catver bump?

elprans, line 47: This needs another bump, as I just merged another PR with the same catver.

elprans (approved):

Issue #5186, opened 2023-03-15
Finish up last parts of triggers


msullivan:
- [x] Make triggers chain and implement a recursion check
- [ ] There is an XXX in the implementation regarding overlays for the DML done inside the trigger. Investigate and fix if needed

PR #5267, opened 2023-03-30
Make triggers chain but disallow recursion

msullivan:
The semantics here is that triggers fire off in "phases"; phase 1 is
all triggers that were triggered by the real query, phase 2 has
triggers that were triggered by phase 1, etc.

Currently we disallow firing the same trigger in multiple phases.

Progress on #5186.

### DIFF
--- a/edb/edgeql/compiler/triggers.py
+++ b/edb/edgeql/compiler/triggers.py
@@ -150,7 +150,7 @@ def compile_triggers_phase(
                     ctx.env.schema, defining_trigger_on):
                 name = str(defining_trigger_on.get_name(ctx.env.schema))
                 raise errors.SchemaDefinitionError(
-                    f"trigger on {name} would need to fire recursively"
+                    f"trigger on {name} is recursive"
                 )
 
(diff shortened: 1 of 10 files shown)


edb/edgeql/compiler/triggers.py
```diff
@@ -134,10 +146,17 @@ def compile_triggers(
 
         # Process all the types, starting with the base type
         for subtype in sorted(stypes, key=lambda t: t != stype):
+            if defining_trigger_on and subtype.issubclass(
+                    ctx.env.schema, defining_trigger_on):
+                name = str(defining_trigger_on.get_name(ctx.env.schema))
+                raise errors.SchemaDefinitionError(
+                    f"trigger on {name} would need to fire recursively"
```
elprans, line 153:
```suggestion
                raise errors.UnsupportedFeatureError(
                    f"trigger {name} is potentially recursive"
```

msullivan: I think "on" is a better preposition for this relationship than "in", but I took the other half of this suggestion.

elprans (approved):

PR #5268, opened 2023-03-30
Allow globals to be used in defaults

### LINKED ISSUE #5223: Consistently allow putting `global` in defaults
Currently we technically allow globals inside pointer defaults, but in practice it will frequently fail if the migration thinks there might be data in the table already. This is because we try to populate all the objects with the default using the `set required using ...` machinery, which disallows globals.

We should be able to fix this by treating the globals as always being unset, in this context.

msullivan:
Currently it is supported only on newly created objects and not when
migrating. Compile default conversion expressions in a mode where
defaults are always treated as empty.

Fixes #5223.

### DIFF
--- a/edb/edgeql/compiler/expr.py
+++ b/edb/edgeql/compiler/expr.py
@@ -479,14 +479,23 @@ def compile_GlobalExpr(
         qry = qlast.SelectQuery(result=qlast.Path(steps=[obj_ref]))
         return dispatch.compile(qry, ctx=ctx)
 
+    default = glob.get_default(ctx.env.schema)
+
+    # If we are compiling with globals suppressed but still allowed, always
+    # treat it as being empty.
+    if ctx.env.options.make_globals_empty:
+        if default:
+            return dispatch.compile(default.qlast, ctx=ctx)
+        else:
+            return setgen.new_empty_set(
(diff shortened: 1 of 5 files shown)


edb/edgeql/compiler/expr.py
```diff
@@ -479,15 +479,24 @@ def compile_GlobalExpr(
         qry = qlast.SelectQuery(result=qlast.Path(steps=[obj_ref]))
         return dispatch.compile(qry, ctx=ctx)
 
+    default = glob.get_default(ctx.env.schema)
+
+    # If we are compiling with globals suppressed but still allowed, always
+    # treat it as being empty.
```
elprans, line 485: I'm not sure about this.  Semantically, migrations are just EdgeQL scripts, so technically I should be allowed to `SET GLOBAL` in them.

elprans: Fair enough, but can you please add a test then in case we somehow allow them to be parsed but forget to actually support things properly.

elprans (approved):

edb/edgeql/compiler/expr.py
msullivan: Sorry, merged without reading, will do a follow up

Issue #5245, opened 2023-03-24
Find a better separator for the list of Access Policy violations


avedetvedea:
A comma doesn't look like it was a well thought out decision. There is a good chance that a comma will be placed in the `errmessage`s of Access Policies, so things will look messy. A semicolon would be better, both in terms of meaning and is not as omnipotent as a comma, but better options may be found

msullivan:
I agree that a semicolon is better and we should switch to it.
Not sure if there is anything preferable to that, though. We want it to read naturally, so don't want to do anything obscure.

PR #5269, opened 2023-03-30
Change the separate in access policy error messages from comma to semicolon

msullivan:
Suggested in #5245, and I think I agree.


msullivan:
But also we apparently have no tests for this behavior; maybe @aljazerzen should write a couple when he gets back

elprans (approved):

PR #5274, opened 2023-03-30
Rename EXPLAIN to ANALYZE, rework syntax

### LINKED ISSUE #5212: Get rid of `ANALYZE` in `EXPLAIN`
Since queries running postres's `EXPLAIN ANALYZE` are reverted already #5177 (#5146). 
 should do `ANALYZE` by default. So the syntax becomes:

1. `EXPLAIN ANALYZE` becomes `EXPLAIN`
2. `EXPLAIN` becomes `EXPLAIN { execute := false }` (or whatever syntax we choose for options)

The syntax for (2) is probably optional for 3.0.

msullivan:
As discussed on offline, arguments will be specified using named tuple
syntax, like
```
analyze (execute := false, buffers := true)
```

Executing the query becomes the default, and can be disabled with
`execute := false`. This is by far the best name for that param that
we've discussed, though using it requires abadoning it as a future
reserved keyword. I feel OK about that.

This turns `ANALYZE` back into a reserved keyword; it was future
reserved in 2.x so this isn't a break.

Fixes #5212.


edb/edgeql/parser/grammar/statements.py
```diff
@@ -274,12 +274,19 @@ def reduce_empty(self, *kids):
         self.val = False
 
 
-class ExplainStmt(Nonterm):
+class AnalyzeStmt(Nonterm):
 
-    def reduce_EXPLAIN_OptAnalyze_ExprStmt(self, *kids):
+    def reduce_ANALYZE_NamedTuple_ExprStmt(self, *kids):
```
elprans, line 279: Do we want to introduce the explicit `ANALYZE (opts) QUERY` form in 3.0?

elprans (approved):

edb/edgeql/parser/grammar/statements.py
msullivan: I don't see why we wouldn't? We've got options that would be useful, so might as well

msullivan: I was confused about the question. My actual answer is "we could, if you want"

PR #5276, opened 2023-03-30
Require inheritance order to be consistent with the specified base order

### LINKED ISSUE #5118: Our ancestor resolution order isn't *quite* C3

msullivan:
It was always, we believe, the intention to implement the C3 ordering
(like Python), and so it was just a bug that this was not enforced in the
past.

Unfortunately this breaks backwards compatability in a pretty obvious way.

This PR does not do include any backward compatability mechanisms, and 
instead will rely on users fixing the issue first and using an 
in-development migration squashing tool (edgedb/edgedb-cli#976), the same
as any other bug fix that impacts compatability.

This one might have a bigger impact than most, though, so maybe we do need
to do something? Or we could just release a beta/RC with it and see if it
seems to cause a lot of grief, and then decide?

Fixes #5118.

### DIFF
--- a/edb/schema/objects.py
+++ b/edb/schema/objects.py
@@ -3326,7 +3326,7 @@ def _merge_lineage(
                 break
         else:
             raise errors.SchemaError(
-                f"Could not find consistent ancestor order for {subject_name}"
+                f"could not find consistent ancestor order for {subject_name}"
             )
 
         result.append(candidate)
@@ -3346,6 +3346,7 @@ def _compute_lineage(
 
     for base in bases:
         lineage.append(_compute_lineage(schema, base, subject_name))
+    lineage.append(list(bases))
 
     return _merge_lineage(lineage, subject_name)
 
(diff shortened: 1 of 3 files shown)


elprans (approved):
We should draft a good "Porting" section about this (and, possibly, other breaking fixes in 3.0).

PR #5283, opened 2023-03-31
SQL pg_dump

vpetrovykh:
Things that still need to be exposed for the SQL dump:
- [x] - Base built-in scalar types
- [x] - Built-in ranges
- [x] - User-defined scalar types
- [x] - Tuples

We generally need to make sure that whatever derived type is used in the schema it is reflected for the `pg_dump`. This means that we need to make sure that arbitrary nested `array<tuple<array<tuple<...>>>` types get reflected in addition to the base types.

### DIFF
--- a/edb/pgsql/metaschema.py
+++ b/edb/pgsql/metaschema.py
@@ -4362,6 +4362,53 @@ class FTSNormalizeDocFunction(dbops.Function):
         )
 
 
+class FormatTypeFunction(dbops.Function):
+    """Used instead of pg_catalog.format_type in pg_dump."""
+
+    text = r'''
+    SELECT
+        CASE WHEN t.typcategory = 'A'
+        THEN (
+            SELECT
+                quote_ident(nspname) || '.' ||
+                quote_ident(el.typname) || tm.mod || '[]'
+            FROM edgedbsql.pg_namespace
+            WHERE oid = el.typnamespace
+        )
+        ELSE (
+            SELECT
+                quote_ident(nspname) || '.' ||
+                quote_ident(t.typname) || tm.mod
+            FROM edgedbsql.pg_namespace
+            WHERE oid = t.typnamespace
+        )
+        END
+    FROM
+        (
+            SELECT
+                CASE WHEN typemod >= 0
+                THEN '(' || typemod::text || ')'
+                ELSE ''
+                END AS mod
+        ) as tm,
+        edgedbsql.pg_type t
+    LEFT JOIN edgedbsql.pg_type el ON t.typelem = el.oid
+    WHERE t.oid = typeoid
+    '''
+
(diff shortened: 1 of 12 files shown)


edb/pgsql/metaschema.py
```diff
@@ -5419,7 +5559,7 @@ def _generate_sql_information_schema() -> List[dbops.Command]:
             tableoid, xmin, cmin, xmax, cmax, ctid
         FROM pg_namespace
         WHERE nspname IN ('pg_catalog', 'pg_toast', 'information_schema',
-                          'edgedb')
+                          'edgedb', 'edgedbstd')
```
aljazerzen, line 5562: Why do we need to expose "edgedb" and "edgedbstd" namespaces? Shouldn't everything that was in them get exposed somewhere else?

edb/pgsql/patches.py
```diff
@@ -44,4 +44,5 @@ def get_version_key(num_patches: int):
 
 
 PATCHES: list[tuple[str, str]] = [
+    ('edgeql+schema', ''),
```
msullivan, line 47: ditch this

edb/pgsql/parser/ast_builder.py
```diff
@@ -955,8 +955,14 @@ def _build_a_expr(n: Node, c: Context) -> pgast.BaseExpr:
 
 def _build_func_call(n: Node, c: Context) -> pgast.FuncCall:
     n = _unwrap(n, "FuncCall")
+
+    name: Tuple[str, ...] = tuple(_list(n, c, "funcname", _build_str))
+    # Substitute our custom function in place of "format_type".
+    if name == ('pg_catalog', 'format_type'):
+        name = ('edgedb', '_format_type')
```
aljazerzen, line 962: I'm still not sure why this needs to be in the parser. We do exactly the same stuff for `pg_relation_filenode` function in `resolver/static.py`.

And also, if you'd have `SELECT format_type(x)'` that should also resolve to `pg_catalog.format_type`, because pg_catalog is implicitly always the first name in search_path. (we are already handling this in static.py)

aljazerzen (approved):
This is colossal amount of work, great job.

edb/pgsql/parser/ast_builder.py
vpetrovykh: Now that the rest of it works and is reasonably stable with lots of tests, let me see if I can move it to the resolver where it should logically belong.

edb/pgsql/metaschema.py
vpetrovykh: `edgedbstd` and `edgedb` have a bunch of types that are slightly different from PG built-in types. Since they are various EdgeDB built-in types, they get used in schemas all the time and they need their definitions.

PR #5292, opened 2023-04-03
Document link property support on backlinks

raddevon:
Added this in a guide even thought they are not versioned because that's the only place I find we talk about how to specify a link property. Even though the note talks about 3.0, it's really more for 2.x users anyway since this is less about what 3.0 _can_ do and more about what 2.x _can't_, so I think this is still appropriate for a 2.x guide.

### DIFF
--- a/docs/guides/cheatsheet/link_properties.rst
+++ b/docs/guides/cheatsheet/link_properties.rst
@@ -158,6 +158,53 @@ Querying
     },
   }
 
+.. note::
+
+    Specifying link properties of a computed backlink in your shape is
+    supported as of EdgeDB 3.0.
+
+    If you have this schema:
+
+    .. code-block:: sdl
+
+        type Person {
+          required name: str;
+          multi follows: Person {
+            followed: datetime {
+              default := datetime_of_statement();
+            };
+          };
+          multi link followers := .<follows[is Person];
+        }
+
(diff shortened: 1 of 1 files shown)


msullivan (changes requested):
Actually, wait, looking at the context, I don't think this helps any.
I think we need to actually add a section explaining what we mean by it and probably giving an example

raddevon:
I've added two examples here, @msullivan: one to show how it works in 3.0 and one to show a workaround for pre-3.0.

msullivan (approved):
Great!

docs/guides/cheatsheet/link_properties.rst
```diff
@@ -158,6 +158,53 @@ Querying
...
+    Specifying link properties of a computed backlink in your shape is
+    supported as of EdgeDB 3.0.
+
+    If you have this schema:
+
+    .. code-block:: sdl
+
+        type Person {
+          required name: str;
+          multi follows: Person {
```
haikyuu, line 172: is link optional when using `multi link`? or is that a typo?

raddevon: `link` is optional in the new SDL syntax shipping with 3.0, except in the case of computed links.

PR #5295, opened 2023-04-03
Fix errmessage interpolation

### LINKED ISSUE #5257: ISE with constraint `errmessage`
- EdgeDB Version: 3.0-dev.7343+4d0b071
- EdgeDB CLI Version: 3.0.0-dev.941+730fc35
- OS Version: ubuntu 22.04

Steps to Reproduce:
```bash
$ edgedb migration create && edgedb migration apply
Created dbschema/migrations/00001.edgeql, id: m1zpd7uqb35mnfakmdohsn4qwc3on427s42ptwfaa6lfznh4fgxs7q
Applied m1zpd7uqb35mnfakmdohsn4qwc3on427s42ptwfaa6lfznh4fgxs7q (00001.edgeql)

$ edgedb
EdgeDB 3.0-dev.7343+4d0b071 (repl 3.0.0-dev.941+730fc35)
edgedb> insert User { email := "" };
edgedb error: InternalServerError: unhandled error while calling interpret_backend_error(); run with EDGEDB_DEBUG_SERVER to debug.
  Server traceback:
      RuntimeError: unhandled error while calling interpret_backend_error(); run with EDGEDB_DEBUG_SERVER to debug.

```

Schema:
```esdl
module default {
  type User {
      required property email -> str {
          constraint min_len_value(4) {
              errmessage := '{"fields": ["email"], "message": "Emails must be longer than 4 characters."}';
          };
      };
  }
}
```

aljazerzen:
Closes #5257

Problem was in errmessage containing json where a JSON object in
curly braces was recognized as an interpolation in Python's `.format`.

Because `errmessage` was not documented to have the whole functionality
that `.format` has, I just replaced `.format` with a regex search and replace.

This will ignore any interpolations that have params (e.g. `{var:02f}`).
It will also not throw InternalServerError when you use interpolations of variables
that were are not provided. This was the reason for this error.

### DIFF
--- a/docs/reference/sdl/constraints.rst
+++ b/docs/reference/sdl/constraints.rst
@@ -139,6 +139,10 @@ The valid SDL sub-declarations are listed below:
       scalar type, property or link on which the constraint is
       defined.
 
+    If the content of curly braces does not match any variables,
+    the curly braces are emitted as-is. They can also be escaped by 
+    using double curly braces.
+
 :sdl:synopsis:`<annotation-declarations>`
     Set constraint :ref:`annotation <ref_eql_sdl_annotations>`
     to a given *value*.
--- a/edb/schema/constraints.py
+++ b/edb/schema/constraints.py
@@ -1571,18 +1571,27 @@ class RebaseConstraint(
 
 def interpolate_errmessage(message: str, args: Dict[str, str]) -> str:
     """
-    Converts "hello {world}! {nope}" and {"world":"Alice", "hell":"Eve"}
-    into "hello Alice! {nope}".
+    Converts message template "hello {world}! {nope}{{world}}" and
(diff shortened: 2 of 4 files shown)


msullivan (approved):
This seems like a plausible approach.

I think the other option would be to make it so that `{` needs to be escaped and to check errmessage for correctness in the schema? (And again do our own interpolation instead of using format?)

Maybe we want to allow escaping `{` (as `{{` and `}}`) anyway?

aljazerzen:
I don't was an error message template to be able to fail, so I don't want to forbid unescaped bad interpolations.

But I did add escaping for double curly braces, so now it is possible to produce `{__subject__}` by using `{{__subject__}}` template.

PR #5300, opened 2023-04-04
Refactor away a few unused args

aljazerzen:
- Refactor away unused function arguments
- Rename get_track_x into get_x_and_track
- Remove env arg of pgsql.pathctx.put_path_var
- Remove more env args from pathctx
- Remove env arg of pgsql.pathctx.put_path_rvar
- Remove env arg of pgsql.pathctx.get_path_rvar
- Remove env args from pgsql.pathctx
- darker

I suggest reviewing commits separately. This PR can also be split to make possible future reverting easier.

### DIFF
--- a/edb/edgeql/compiler/casts.py
+++ b/edb/edgeql/compiler/casts.py
@@ -97,14 +97,14 @@ def compile_cast(
             f'`...[IS {new_stype.get_displayname(ctx.env.schema)}]` instead',
             context=srcctx)
 
-    uuid_t = ctx.env.get_track_schema_type(sn.QualName('std', 'uuid'))
+    uuid_t = ctx.env.get_schema_type_and_track(sn.QualName('std', 'uuid'))
     if (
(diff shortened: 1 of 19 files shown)


edb/edgeql/compiler/casts.py
```diff
@@ -97,15 +97,14 @@ def compile_cast(
             f'`...[IS {new_stype.get_displayname(ctx.env.schema)}]` instead',
             context=srcctx)
 
-    uuid_t = ctx.env.get_schema_type_and_track(sn.QualName('std', 'uuid'))
```
msullivan: I don't like the `"` quote thing at all; could you redo the darker step with that turned off? (IIRC that is like one of two config flags that black has.)

msullivan (approved):
Looks good. I think it would probably be worth doing 'rebase and merge' for this, so make sure the commits all have plausible messages?

PR #5305, opened 2023-04-05
pg-ext: fix version() to include PostgreSQL

fantix:
Also fixed alias name of statically-evaluated FuncCall

### DIFF
--- a/edb/pgsql/resolver/static.py
+++ b/edb/pgsql/resolver/static.py
@@ -19,11 +19,13 @@
 """Static evaluation for SQL."""
 
 import functools
+import platform
 from typing import *
 
 from edb import errors
 
 from edb.pgsql import ast as pgast
+from edb.server import defines
 
 from . import context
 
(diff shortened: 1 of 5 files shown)


edb/server/main.py
```diff
@@ -439,7 +439,24 @@ async def run_server(
...
+        source += f" on {nice_date}"
+    if ver_meta["scm_revision"]:
+        source += f" from revision {ver_meta['scm_revision']}"
+        if source_date := ver_meta["source_date"]:
+            nice_date = source_date.strftime("%Y-%m-%dT%H:%MZ")
+            source += f" ({nice_date})"
+    if source:
+        extras.append(f", built{source}")
+    if ver_meta["target"]:
+        extras.append(f"for {ver_meta['target']}")
```
elprans, line 456: What's the point of open-coding `get_version_line()` here?

fantix: This is a partial revert of https://github.com/edgedb/edgedb/pull/5115, it was open-coding before #5115; but we can still keep the `get_version_line()` here.

elprans: I would keep it

edb/pgsql/resolver/static.py
```diff
@@ -113,8 +115,21 @@ def eval_FuncCall(
 
     if fn_name == 'version':
         from edb import buildmeta
+
+        args = ["EdgeDB", buildmeta.get_version_string()]
```
elprans, line 119: Why not keep using `get_version_line` here also?

aljazerzen (approved):
Fivetran is satisfied, thanks.

PR #5310, opened 2023-04-06
Forbid DML in non-scalar function args

### LINKED ISSUE #5172: Disallow passing insert/update objects to user-defined functions
(I thought there was an issue for this but I couldn't find it)

If a function is implemented as a SQL function, it won't be able to use overlays to see newly created data. This can cause a bunch of problems, including breaking cardinality.

I think that we don't have any stdlib functions that take object types that are implemented as functions, so this probably just affects user defined functions.

This will affect some access policy uses also

aljazerzen:
Closes #5172

### DIFF
--- a/edb/edgeql/compiler/func.py
+++ b/edb/edgeql/compiler/func.py
@@ -214,7 +214,10 @@ def compile_FunctionCall(
         ctx=ctx,
     )
 
-    if func_name.module != 'std':
+    # Forbid DML in non-scalar function args
+    if func.get_nativecode(env.schema):
+        # We are sure that there is no such functions implemented with SQL
+
         for arg in final_args:
             if arg.expr.typeref.is_scalar:
                 continue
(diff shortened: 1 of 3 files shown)


tests/test_edgeql_calls.py
```diff
@@ -1673,6 +1673,39 @@ async def test_edgeql_calls_obj_04(self):
...
+            await self.con.execute(
+                r"SELECT boo((UPDATE Ghost SET { name := 'Tom' }))",
+            )
+
+        async with self.assertRaisesRegexTx(
+            edgedb.UnsupportedFeatureError,
+            r'newly created or updated objects cannot be passed to functions',
+        ):
+            await self.con.execute(
+                r"SELECT boo((INSERT Ghost { name := 'Jack' }));",
```
msullivan, line 1706: Could you add a test that it works when the insert is bound to a variable?

edb/edgeql/compiler/func.py
```diff
@@ -214,6 +214,18 @@ def compile_FunctionCall(
         ctx=ctx,
     )
 
+    if func_name.module != 'std':
```
msullivan, line 217: Let's check whether the function has a definition in edgeql instead; (`nativecode`)

msullivan (approved):

PR #5313, opened 2023-04-06
Fix adding certain computed links to a type with an alias

msullivan:
Pointers that are a direct alias for another pointer were being marked
as `is_derived`, which is mostly supposed to mean things that are
*automatically* derived in a view or some such.

When a type with such a computed pointer had an alias created of it,
`_elide_derived_ancestors` in the edgeql compiler was marking the base
of the pointer in the view subclass as being the aliased pointer in
the base class, which is inconsistent with what happened when an ALTER
was emitted.

Downside of this fix is that it cherry-picking it is dodgy, since it
changes how things are stored after a simple CREATE.


msullivan:
Closing because I'm going to do it pretty differently

PR #5317, opened 2023-04-10
SQL pg_catalog views for pg_dump

aljazerzen:
- Prune SQL pg_catalog metaschema for pg_dump
- Fix some pg_dump issues.
- SQL pg_catalog.pg_index

### DIFF
--- a/edb/pgsql/metaschema.py
+++ b/edb/pgsql/metaschema.py
@@ -5418,7 +5418,8 @@ def _generate_sql_information_schema() -> List[dbops.Command]:
         SELECT oid, nspname, nspowner, nspacl,
             tableoid, xmin, cmin, xmax, cmax, ctid
         FROM pg_namespace
-        WHERE nspname IN ('pg_catalog', 'pg_toast', 'information_schema')
+        WHERE nspname IN ('pg_catalog', 'pg_toast', 'information_schema',
+                          'edgedb')
         UNION ALL
         SELECT edgedbsql.uuid_to_oid(t.module_id), t.schema_name, 10, NULL,
             NULL, NULL, NULL, NULL, NULL, NULL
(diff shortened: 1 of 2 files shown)


edb/pgsql/metaschema.py
```diff
@@ -5588,16 +5611,18 @@ def _generate_sql_information_schema() -> List[dbops.Command]:
...
-            name=("edgedbsql", "pg_range"),
-            query="""
-        SELECT pr.*, pr.tableoid, pr.xmin, pr.cmin, pr.xmax, pr.cmax, pr.ctid
-        FROM pg_range pr
-        JOIN pg_type pt ON pt.oid = pr.rngtypid
-        JOIN pg_namespace pn ON pt.typnamespace = pn.oid
-        WHERE nspname IN ('pg_catalog', 'pg_toast', 'information_schema')
-        """,
-        ),
+        # # We want to expose our built-in ranges in addition to PG native
```
msullivan, line 5614: Why is this commented out?

edb/pgsql/metaschema.py
```diff
@@ -5621,6 +5646,165 @@ def _generate_sql_information_schema() -> List[dbops.Command]:
...
+            s.stanumbers1 AS histogram_bounds,
+            s.stanumbers1[1] AS correlation,
+            NULL::real[] AS most_common_elems,
+            s.stanumbers1 AS most_common_elem_freqs,
+            s.stanumbers1 AS elem_count_histogram
+        FROM pg_statistic s
+        JOIN pg_class c ON c.oid = s.starelid
+        JOIN pg_attribute a ON c.oid = a.attrelid and a.attnum = s.staattnum
+        LEFT JOIN pg_namespace n ON n.oid = c.relnamespace
+        WHERE FALSE
```
msullivan, line 5670: Why do all these joins and then filter it all out?

msullivan (approved):
I think that a lot of these views probably could use some more comments explaining them (including existing views, not just the new ones). That's not a requirement for right now, though, but it is genuinely something we should try to clean up in the near future

PR #5319, opened 2023-04-10
SQL pg_catalog function wrappers and static eval

aljazerzen:


edb/pgsql/metaschema.py
```diff
@@ -5952,7 +5960,9 @@ def construct_pg_view(table_name: str, columns: List[str]) -> dbops.View:
             ),
             returns=('bool',),
             text="""
-                SELECT has_schema_privilege(schema_oid, privilege)
+                SELECT COALESCE(
```
msullivan, line 5963: Do you need both this and the above UNION trick?

edb/pgsql/resolver/static.py
```diff
@@ -162,16 +234,90 @@ def eval_FuncCall(
...
+
+    namespace, rel_name = name
+
+    # A bit hacky to parse SQL here, but I don't want to construct pgast
+    [stmt] = pgparser.parse(
+        f'''
+        SELECT pc.oid
+        FROM edgedbsql.pg_class pc
+        JOIN edgedbsql.pg_namespace pn ON pn.oid = pc.relnamespace
+        WHERE '{namespace}' = pn.nspname AND pc.relname = '{rel_name}'
```
msullivan, line 314: Use `common.quote_literal` instead of quoting stuff yourself, to avoid exposure to injection vulnerabilities.

msullivan (approved):

edb/pgsql/metaschema.py
aljazerzen: These are separate functions because they have different signatures.

I wanted to avoid the UNION trick, but couldn't figure out how.

aljazerzen: Oh, I can just use COALESCE whose first arg is a sublink.

Note: SQL's behavior when converting a result of a relational query to a scalar that can be coalesced and returned from a function is not intuitive.

aljazerzen: Also, I'll be squashing this fix into a commit, because I intend to "rebase and merge"

PR #5323, opened 2023-04-11
SQL COPY parsing, codegen, and resolving

aljazerzen:


tests/test_sql_query.py
```diff
@@ -859,3 +859,8 @@ async def test_sql_transaction_02(self):
         await self.scon.execute("ROLLBACK")
         v2 = await self.scon.fetchval("SHOW transaction_isolation")
         self.assertNotEqual(v1, v2)
+
+    async def test_sql_query_copy_01(self):
+        await self.scon.fetch(
+            '''COPY "Movie" TO STDOUT (FORMAT CSV, DELIMITER '\t')'''
+        )
```
aljazerzen, line 866: I've tried capturing and comparing the result, but I'm not sure how asyncpg and "STDOUT" works here. I'd need an expert's opinion here @fantix .

PR #5329, opened 2023-04-11
Fix adding certain computed links to a type with an alias

msullivan:
When a type with such a computed pointer that was an alias of another
pointer had an alias created of it, `_elide_derived_ancestors` in the
edgeql compiler was marking the base of the pointer in the view
subclass as being the aliased pointer in the base class, which is
inconsistent with what happened when an ALTER was emitted.

Fix this by having `_elide_derived_ancestors` be more careful about
what it skips over.

Downside of this fix is that it cherry-picking it is dodgy, since it
changes how things are stored after a simple CREATE.

### DIFF
--- a/edb/edgeql/compiler/stmtctx.py
+++ b/edb/edgeql/compiler/stmtctx.py
@@ -507,6 +507,7 @@ def _get_nearest_non_source_derived_parent(
     obj: s_obj.DerivableInheritingObjectT,
     ctx: context.ContextLevel
 ) -> s_obj.DerivableInheritingObjectT:
+    """Find the nearest ancestor of obj whose "root source" is not derived"""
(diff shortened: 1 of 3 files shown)


edb/edgeql/compiler/stmtctx.py
```diff
@@ -503,6 +503,20 @@ def _fixup_schema_view(
                 )
 
 
+def _get_nearest_non_source_derived_parent(
+    obj: s_obj.DerivableInheritingObjectT,
+    ctx: context.ContextLevel
+) -> s_obj.DerivableInheritingObjectT:
```
elprans, line 509: Add a docstring?

elprans (approved):

PR #5332, opened 2023-04-12
Update graphiql to forked version with globals support

jaclarke:
Updates the `/graphql/explore` endpoint to serve a customised version of the GraphiQL UI built from this fork: https://github.com/edgedb/graphiql with support for globals.

The bundled js/css files from this custom build are committed to this repo, so as not to add further complexity to the edgedb build (and they're unlikely to be updated frequently anyway). To re-build these files run the `edb/graphql/explore/build.sh` script.

image


msullivan:
This seems reasonable, but I just realized that if we want, we could just support passing globals in the `variables` argument. Globals all have `::` in their name and parameters can't, so it would work just fine

AndreasPB:
Would it be possible to add the explorer plugin? It makes working with GraphiQL a fantastic experience

https://github.com/graphql/graphiql/blob/HEAD/packages/graphiql-plugin-explorer

Demo

https://github.com/edgedb/edgedb/assets/43907402/df6b1aee-0ade-49dc-b3d6-2f6c34691c46

msullivan:
In 4.0, we allow passing in globals via variables using `__globals__`.
Does that obsolete this?

PR #5337, opened 2023-04-12
Implement a schema repair mechanism

### LINKED ISSUE #5321: Cardinality bug fixes can leave migration state wedged
This is a general class of problems we face, along with a specific way that a 2.13 bugfix triggers it. This is minimized from an actual user issue.

Create and migrate to this schema in 2.12:
```
module default {
    type Foo {
        required multi property spam -> str;
        property eggs := (select .spam);
    }
};
```

Then, upgrade to 2.13, and run `edgedb migration create` again (with no changes):
```
$ edgedb migration create
edgedb error: EdgeDB could not resolve migration with the provided answers. Please retry with different answers.
```

If we run with `EDGEDB_DEBUG_DELTA_PLAN=1`, we get some insight into the situation:
```
================================================================================
DESCRIBE CURRENT MIGRATION AS JSON mismatch
================================================================================
<DeltaRoot canonical=True at 0x7f2793600670> (
    <AlterObjectType canonical=True, classname=default::Foo, annotations={'confidence': 1.0}, orig_cmd_type=<class 'edb.schema.objtypes.AlterObjectType'> at 0x7f2793601690> (
        <AlterProperty canonical=True, classname=default::__|eggs@default|Foo, annotations={'confidence': 1.0}, aux_object_data={'from_alias': None} at 0x7f2791ca4160> (
            <AlterPropertyLowerCardinality classname=default::__|eggs@default|Foo at 0x7f2791ca5540>(required=False | True # computed)
        )
    )
)
```

We can see that `Foo.eggs` in the database has `required=False` but that we want it to have `required=True`, and we fail to create a migration.

----

The issue here is that in 2.13, we fixed a cardinality inference bug: https://github.com/edgedb/edgedb/pull/5180
and as a result, `eggs` infers a different cardinality in 2.13 than in 2.12. This creates a mismatch between the actual live schema and the schema that we create in 2.13, and so we aren't able to do any migrations.

Making a trivial change to `eggs` lets us recover from this state, since that will cause it to get re-inferred directly.

Of course, there is no good way for an actual user to discover that really simple workaround, so this is kind of a mess.

Dump/restore also would work, I think
----

I'm not sure what the best approach to this category of issues is. I'm not sure that we can reasonably rule out these kinds of changes.

I think we might need some sort of repair tool. Maybe a command that does something along the lines of computing what the schema *should* be (from `describe` output, maybe), comparing that to the actual schema, and just making the fixups.

msullivan:
As discussed in #5321, certain changes and bugfixes can result in the
schema state being inconsistent, with inferred attributes that do not
match what would be created if the schema was created freshly.

Implement a repair system that works by serializing the current schema
to DDL, generating a fresh schema from that, then comparing the
current schema to this new target and applying the delta.
For safety, we refuse if the change is not data-safe.

This mechanism is exposed as a new "patch kind" for the patch mechanism,
and as a test-mode ADMINISTER function.

I tested this manually using #5180 and #5329. When we add 'repair'
patches to stable branches we should endeavor to make sure that it
actually gets exercised by the old versions CI job.
Also tested with an edgeql+schema patch made after it.

Fixes #5321.

### DIFF
--- a/edb/server/compiler/ddl.py
+++ b/edb/server/compiler/ddl.py
@@ -42,6 +42,7 @@ from edb.schema import migrations as s_migrations
 from edb.schema import objects as s_obj
 from edb.schema import schema as s_schema
 from edb.schema import utils as s_utils
+from edb.schema import version as s_ver
 
 from edb.pgsql import common as pg_common
 from edb.pgsql import delta as pg_delta
@@ -288,12 +289,12 @@ def compile_and_apply_ddl_stmt(
     )
 
 
-def _new_delta_context(ctx: compiler.CompileContext):
+def _new_delta_context(ctx: compiler.CompileContext, args: Any=None):
     return s_delta.CommandContext(
         backend_runtime_params=ctx.compiler_state.backend_runtime_params,
         stdmode=ctx.bootstrap_mode,
         internal_schema_mode=ctx.internal_schema_mode,
-        **_get_delta_context_args(ctx),
+        **(_get_delta_context_args(ctx) if args is None else args),
     )
 
 
@@ -695,7 +696,7 @@ def _describe_current_migration(
         if proposed_desc is None:
             diff = s_ddl.delta_schemas(schema, mstate.target_schema)
             complete = not bool(diff.get_subcommands())
-            if debug.flags.delta_plan:
+            if debug.flags.delta_plan and not complete:
                 debug.header('DESCRIBE CURRENT MIGRATION AS JSON mismatch')
                 debug.dump(diff)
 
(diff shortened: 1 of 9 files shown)


edb/server/compiler/ddl.py
```diff
@@ -1160,3 +1161,99 @@ def _reset_schema(
...
+    if not mismatch:
+        return None
+
+    if debug.flags.delta_plan:
+        debug.header('Repair Delta')
+        debug.dump(delta)
+
+    if not delta.is_data_safe():
+        raise AssertionError(
+            'Repair script for version upgrade is not data safe'
```
elprans, line 1211: Would this be reraised as an ISE?

msullivan:
Yeah. We could do something else, but we do want to be generating a "file a
bug" message

elprans:
> Yeah. We could do something else, but we do want to be generating a "file a bug" message

That's exactly what we want to do, just making sure we are actually turning vanilla exceptions into ISEs in the bootstrap path.

msullivan:
Oh sure enough, we are not actually

aljazerzen (approved):
Looks good to me

msullivan:
Hmmmmm. There's a danger with this scheme, though, which is: what to do if trying to reload the schema fails for a legitimate reason?

After reviewing the 2.0 changelog, I'm less worried about this than I used to be. The only PR I spotted in a quick pass with a risk of this was https://github.com/edgedb/edgedb/pull/4616, which is pretty marginal. I know the 1.x series had some worse ones though

It might be fine to just make the log message in that case say that you might need to fix-and-squash?

PR #5342, opened 2023-04-13
Document `edgedb watch` and new dev workflow

raddevon:
Keeping this as a draft PR until 3.0 launches since it is in non-versioned documentation, but the content is ready for review.


docs/cli/edgedb_watch.rst
```diff
@@ -0,0 +1,23 @@
+.. _ref_cli_edgedb_watch:
+
+
+============
+edgedb watch
```
tailhook, line 5: I think this command can have a "version added" marker. Or don't we have one?

docs/intro/migrations.rst
```diff
@@ -21,69 +39,77 @@ your codebase.
...
+      required title: str;
+      required author: User;
     }
 
   + type Comment {
-  +   required property content -> str;
+  +   required content: str;
   + }
 
+When you save your changes, the updates will be immediately applied to your
```
tailhook, line 91: Is it okay to say "immediately" when in reality it's 3-5-10 seconds?

docs/intro/migrations.rst
```diff
@@ -21,69 +39,77 @@ your codebase.
...
+    If your schema cannot be applied, the ``watch`` command will generate an
+    error. If you're using one of our client bindings as you update your schema
+    with ``watch``, you will see the error there the next time you execute a
+    query using that client binding.
+
+    If things aren't working the way you expect after making a schema change,
+    take a look at the ``watch`` console to find out why.
+
+Once you have the schema the way you want it, and you're ready to lock it in
+and commit it to version control, stop your ``watch`` command with Ctrl-c. It's
```
tailhook, line 105: You don't have to stop ``watch`` to run ``migration create`` or commit. Did you see some failures if it doesn't?

docs/intro/migrations.rst
```diff
@@ -105,87 +131,129 @@ advanced options.
...
+      default::Post {id: a4a0a40c-d9f5-11ed-8912-1397f7af9fdf},
+      default::Post {id: cc051bea-d9f5-11ed-a26d-2b64b6b273a4}
+    }
+
+Now, if we save the schema again, ``edgedb watch`` will be able to apply it. If
+we have data in here we don't want to lose though, that's not a good option. In
+that case, we might drop back to creating and applying the migration outside of
+``edgedb watch``.
+
+To start, end your ``watch`` with Ctrl-c. Then run ``edgedb migration create``.
```
tailhook, line 213: Same here.

tailhook (approved):

docs/intro/migrations.rst
Dhghomon: The etymologist in me likes immediately since at its core it means "without mediation" / "nothing in between" but I guess that meaning is pretty rare now. Maybe a move to active tense and change to something like "immediately move to apply" could be good:

When you save your changes, ```watch``` will immediately move to apply the changes to your database as long as it is still running.

docs/cli/edgedb_watch.rst
raddevon: We have one, but this is in part of the documentation that is not versioned. (Technically, it s versioned in that I can drop `versionadded` anywhere, but we've split out the docs sections that are versioned from those that aren't in the UI.)

image

I believe the thinking in not including the CLI in the overall doc versioning is that it has it's own versions that might not map cleanly to EdgeDB versions, so having that content change with the dropdown wouldn't necessarily work. Is that right, @1st1?

Do we need to rethink that or separately version the CLI section of the docs in some way? Maybe we keep using `versionadded` here in CLI but handle it differently? Maybe we just display the version number from `versionadded` but don't actually show or hide content based on the dropdown setting? That way, the version numbers we use in the CLI docs could correlate not to EdgeDB server versions but to the CLI's versions.

docs/intro/migrations.rst
raddevon: I like the suggestion, @Dhghomon. What about "When you save your changes, the CLI will immediately begin applying the changes to your..."?

docs/intro/migrations.rst
raddevon: No, this was a bad assumption on my part.

docs/intro/migrations.rst
Dhghomon: That looks good too, yeah. The only change I might suggest then is to mention the CLI once a bit earlier so there's no mistaking that `watch` is happening through the CLI. Perhaps here:

This long-running task will monitor your schema files

change to

This long-running task uses the CLI to monitor your schema files

or

This long-running task gets the CLI to monitor your schema files

Or something along those lines.

raddevon:
Feedback has been addressed with the exception of `versionadded`. I'll keep this PR as draft either until 3.0 is released or until we make a decision re: versioning for the CLI docs that would make merging earlier practical.

docs/cli/edgedb_watch.rst
tailhook: Yes, I think we can just display version number here but not show/hide. This specific feature depends on EdgeDB 3.0, but we can also have features that don't. And we can introduce features in minor versions of the CLI tools too. For example `migration create --squash` is introduced in CLI 3.0, but by design also works in 2.x (although, I'm not sure which one of 2.x). Even if we port it to 2.x branch the minor version of CLI tool is not in sync with minor version of the server.

tailhook (approved):

PR #5345, opened 2023-04-13
Forbid ranges of user-defined scalars

aljazerzen:
This is the offending schema:

```
module default {
    type User {
        property age -> range<Age>;
    };

    scalar type Age extending int16 {
        constraint min_value(13);
        constraint max_value(99);
    };
} 
```

@msullivan, am I correct to assume that we should not allow this, with error:

```
unsupported range subtype: default::Age
```

In this specific case, we could allow it, by using a range of int64.

### DIFF
--- a/edb/pgsql/types.py
+++ b/edb/pgsql/types.py
@@ -193,9 +193,8 @@ def pg_type_range(tp: Tuple[str, ...]) -> Tuple[str, ...]:
 
 
 def pg_type_from_object(
-        schema: s_schema.Schema,
-        obj: s_obj.Object,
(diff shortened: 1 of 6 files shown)


msullivan:
Yeah, disallowing it for now seems reasonable if supporting it would be a pain.
@vpetrovykh do you remember much about how hard it would be to support this?

msullivan (commented):
Hm, would it be possible to do this check earlier?
I worry that this wouldn't fire when creating a migration, just when applying it.
(We have a lot of stuff like that but we'd like to avoid it)

msullivan (approved):

PR #5350, opened 2023-04-13
Turn `_describe_type` into `singledispatch`

elprans:
It is currently a simple `if isinstance` ladder, so this is another
trivial Shift+Tab.


msullivan (approved):
Oops I thought I got this one already

Issue #5322, opened 2023-04-11
Schema comparison is way too slow for largeish schemas


msullivan:
I think the heart of this problem (which is not the only migration perf problem) is that schema comparisons are quadratic in the number of objects in each top-level object class.

We need to do something to dramatically reduce how many comparisons we do.

There is a mechanism for computing an `unchanged` set based on computing `hash_criteria` for objects, but it doesn't seem to work because the criteria include things that will always be different between old and new schemas (like references to the target type). It might be possible to get that actually working, though.

Another possibility is having a trivial name-based rule, like: if an object with the same name exists in both the old and the new schema, don't compare it to anything else!

PR #5351, opened 2023-04-13
Optimize delta_objects by taking into account name information

msullivan:
The idea is simply that if an object named `Foo` exists in both the
old and new schemas, compare those `Foo`s against each other and
nothing else, saving the quadratic cross-product based comparison
for objects that don't appear on both sides.

On the pretty modestly sized `cards` database, this reduces the cost
of the initial `describe current migration as json` for a trivial
property addition from about 3.5s to 0.9s. Dropping the inoperative
`hash_criteria` optimization shaves another 0.05s.

Progress on #5322.

### DIFF
--- a/edb/schema/delta.py
+++ b/edb/schema/delta.py
@@ -50,8 +50,8 @@ from . import utils
 
 
 def delta_objects(
-    old: Iterable[so.Object_T],
-    new: Iterable[so.Object_T],
+    old_in: Iterable[so.Object_T],
+    new_in: Iterable[so.Object_T],
     sclass: Type[so.Object_T],
     *,
     parent_confidence: Optional[float] = None,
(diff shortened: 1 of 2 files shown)


edb/schema/delta.py
```diff
@@ -62,35 +62,43 @@ def delta_objects(
...
-    )
-
-    oldnames = {o.get_name(old_schema) for o in old}
-    newnames = {o.get_name(new_schema) for o in new}
-    common_names = oldnames & newnames
-
-    pairs = sorted(
-        itertools.product(new, old),
-        key=lambda pair: pair[0].get_name(new_schema) not in common_names,
+    # TODO: Previously, we attempted to do an optimization based on
```
elprans, line 65: Let's make an issue to fix it and bring it back.

elprans (approved):

PR #5352, opened 2023-04-13
Improve the docs for `array_unpack`

msullivan:
array_unpack's docs correctly call out that the order of the returned
set is not guaranteed (though in practice is probably is for non
objects). Note in the docs that wrapping the call in enumerate will
preserve the indexes.


Dhghomon:
Looks good. One other teensy change I've been meaning to make to this page is near the bottom:

Return an array -> Returns an array (add an s)

raddevon (approved):

PR #5353, opened 2023-04-14
Small grammar fixes and wording suggestions

Dhghomon:
Just a small PR with some grammar fixes and wording suggestions I've been jotting down bit by bit over the past few weeks.


docs/edgeql/paths.rst
```diff
@@ -37,7 +37,7 @@ friend of another User*.
 
   select User.friends;
 
-Paths can traverse arbitrarily many links.
+Paths can arbitrarily traverse multiple links.
```
raddevon, line 40: I agree this needs a rewrite, but I think this one may change the intended meaning slightly. What about this?

```suggestion
Paths can traverse to an arbitrary depth in a series of nested links.
```

Judging based on the example, I think this gets closer to what we're trying to show.

docs/intro/schema.rst
```diff
@@ -226,7 +226,8 @@ understand backlink syntax is to split it into two parts:
 
 ``[is Movie]``
   This is a *type filter* that filters out all objects that aren't ``Movie``
-  objects.
+  objects. A backlink still works without this filter, but could contain any 
+  other number of objects besides `` Movie``s.
```
raddevon, line 230:
```suggestion
  other number of objects besides `` Movie`` objects.
```

Sphinx doesn't seem to cooperate when you have inline code butted up against another character. I usually just write around it like this.

raddevon (changes requested):
Two minor suggestions. The rest looks good!

docs/intro/schema.rst
Dhghomon: Ah! Good to know.

raddevon (approved):

PR #5355, opened 2023-04-14
Fix insert default cycle

### LINKED ISSUE #5348: Recursive defaults with DML will cause ISE: recursion depth limit exceeded
EdgeDB 3.0-dev.7456+local (repl 3.0.0-dev)

Create migration on the following schema
```
type Test3 {
    d1 : str ;
    d7 : Test3 {
        default :=  (insert Test3 {
            d1 := "d1"
        })
    }
}
```

aljazerzen:
Closes #5348

### DIFF
--- a/tests/test_edgeql_ddl.py
+++ b/tests/test_edgeql_ddl.py
@@ -1952,6 +1952,19 @@ class TestEdgeQLDDL(tb.DDLTestCase):
                 };
             """)
 
+    async def test_edgeql_ddl_default_12(self):
+        with self.assertRaisesRegex(
+            edgedb.QueryError,
+            'is part of a default rule cycle'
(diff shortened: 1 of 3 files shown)


tests/test_edgeql_ddl.py
```diff
@@ -1952,6 +1952,19 @@ async def test_edgeql_ddl_default_11(self):
                 };
             """)
 
+    async def test_edgeql_ddl_default_12(self):
+        with self.assertRaisesRegex(
+            edgedb.QueryError,
+            'is part of a default rule cycle'
```
msullivan, line 1958: drop "rule", I guess?

msullivan (approved):

PR #5359, opened 2023-04-14
More cleanups in sertypes

elprans:
Use helpers to encode strings and descriptor references, add comments,
and clean things up a bit.


msullivan (approved):
This code is starting to seem downright reasonable

PR #5378, opened 2023-04-24
Fix minor TypeScript syntax error in docs

Arrow7000:
Which presumably was copied from the EdgeQL code sample


scotttrinh (approved):
Made one minor change, thanks so much for catching this!

PR #5379, opened 2023-04-25
Document SQL support

raddevon:
I didn't make any massive changes here to what @aljazerzen gave me to work with, but please look over it carefully and make sure I didn't change the meaning of anything.

> Make sure to require SSL encryption when using SQL support

Could we very briefly explain why this is particularly important in this context?

Last note: I linked to all the tested SQL tools. Could you click through on them just to confirm I found the correct links for all of them?


docs/reference/sql_support.rst
```diff
@@ -0,0 +1,130 @@
...
+- `pg_dump <https://www.postgresql.org/docs/13/app-pgdump.html>`_
+- `Metabase <https://www.metabase.com/>`_
+- `Cluvio <https://www.cluvio.com/>`_
+- `Tableau <https://www.tableau.com/>`_
+- `DataGrip <https://www.jetbrains.com/datagrip/>`_
+- `Airbyte <https://airbyte.com/>`_
+- `Fivetran <https://www.fivetran.com/>`_
+- `Hevo <https://hevodata.com/>`_
+- `Stitch <https://www.stitchdata.com/>`_
+- `dbt <https://www.getdbt.com/>`_ [#]_
```
aljazerzen, line 122: Looks good.

docs/reference/sql_support.rst
```diff
@@ -0,0 +1,130 @@
...
+
+
+.. [#] dbt models are built and stored in the database as either tables or
+   views. Because the EdgeDB SQL connector does not allow writing or even
+   creating schemas, view, or tables, any attempt to materialize dbt models
+   will result in errors. If you want to build the models, we suggest first
+   transferring your data to a true Postgres instance via pg_dump or Airbyte.
+   Tests and previews can still be run directy against the EdgeDB instance.
```
aljazerzen, line 130: Airbyte, Fivetran, Hevo and Stitch should also have this comment:


> At the moment, EdgeDB does not support "Log replication" (i.e. using [Postgres replication mechanism](https://www.postgresql.org/docs/current/runtime-config-replication.html)). Supported replication methods include [XMIN Replication](https://www.postgresql.org/docs/15/ddl-system-columns.html), incremental updates using "a user-defined monotonically increasing id" and full table updates.


The methods described don't have official names - I'm just using the terminology that the tools have in their interfaces.

aljazerzen (approved):

raddevon:
> To be honest, I don't know much about why this is so. I've just noticed that it does not work if I don't tick that box when entering database details. 🤷 @fantix ?

Oh, I see. I thought you were saying this is something that needs to be enabled on the EdgeDB side, but this is actually on the tool side. Let me rework it to note that, and we may not really need an explanation unless @fantix has something that would make sense to add here.

raddevon:
Added some more explicit instructions for connecting, including implementing @tailhook's suggestion of recommending the user create a role for the SQL client.

@aljazerzen Would you take one more pass on this with the connection instructions and make sure everything looks good before I merge?

docs/reference/sql_support.rst
```diff
@@ -0,0 +1,195 @@
...
+
+    SELECT id, title FROM common."Content";
+
+To omit objects of extended types, use ``ONLY``. This query will return
+``common::Content`` objects but not ``Movie`` objects.
+
+.. code-block:: sql
+
+    SELECT id, title FROM ONLY common."Content";
+
```
aljazerzen, line 133: Nice explanation, very readable.

docs/reference/sql_support.rst
```diff
@@ -0,0 +1,195 @@
...
+===========
+
+
+Connecting
+==========
+
+EdgeDB supports running SQL queries via the Postgres protocol. Any
+Postgres-compatible client can connect to your EdgeDB database by using the
+same port that is used for the EdgeDB protocol and the same database name,
+username, and password you already use for your database.
```
aljazerzen, line 16: We should also highlight that we only support read-only queries. No DML and DDL.

raddevon: Done. All I did here was chage "SQL queries" to "read-only SQL queries." We describe the implications of this later so I don't think detail here is necessary, but I do think it's a good idea to call it out up-front.

PR #5380, opened 2023-04-27
Fix SQL introspection on custom Postgres versions

aljazerzen:


edb/server/compiler/compiler.py
```diff
@@ -666,6 +666,7 @@ def parse_search_path(search_path_str: str) -> list[str]:
                 )
                 resolved = pg_resolver.resolve(stmt, schema, options)
                 source = pg_codegen.generate_source(resolved)
+                print(source)
```
msullivan, line 669: stray

msullivan (approved):
Looks good!

PR #5387, opened 2023-04-28
Clarification that backlinks are multi by default

Dhghomon:
Noticed today that backlinks are multi by default. This seems like the best page to inform the user of this.

### DIFF
--- a/docs/datamodel/links.rst
+++ b/docs/datamodel/links.rst
@@ -48,6 +48,31 @@ link.
       multi friends: Person;
     }
 
+On the other hand, backlinks work in reverse to find objects that link to the
+object, and thus assume  ``multi`` as a default. Use the ``single`` keyword 
+to declare a "to-one" backlink.
+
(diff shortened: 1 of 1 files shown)


docs/datamodel/links.rst
```diff
@@ -48,6 +48,31 @@ link.
...
+    :version-lt: 3.0
+
+    type Author {
+      link posts := .<authors[is Article];
+    }
+
+    type CompanyEmployee {
+      single link company := .<employees[is Company];
+    }
+
```
raddevon, line 65: You could probably just delete this since these are identical.
```suggestion
```

raddevon (approved):
Great find!

docs/datamodel/links.rst
Dhghomon: Ah, no need for the separate 3.0 version if they are the same? Okay, sounds good.

PR #5399, opened 2023-05-01
Refactor EdgeQL AST to remove multiple inheritance

aljazerzen:
- inline LimitOffsetMixin
- inline Clause abstract class
- remove qlast.astmatch
- inline ReturningMixin
- inline FilterMixin
- inline OrderByMixin
- inline SubjectMixin
- inject SelectClauseMixin
- inline BaseAlias
- nest SessionSetAliasDecl
- ast refactor
- Refactor to make code darker

Converts inheritance tree of node types in our AST from this:

ast-fdp-initial

into this:

ast-fdp-refactored

(excluding DDL/SDL node types and node types extending only `Base` class)


edb/edgeql/compiler/viewgen.py
```diff
@@ -1497,10 +1497,12 @@ def _normalize_view_ptr_expr(
...
             qlexpr = astutils.ensure_ql_select(qlexpr)
 
-        if ((ctx.expr_exposed or ctx.stmt is ctx.toplevel_stmt)
-                and ctx.implicit_limit
-                and isinstance(qlexpr, qlast.OffsetLimitMixin)
-                and not qlexpr.limit):
+        if (
+            (ctx.expr_exposed or ctx.stmt is ctx.toplevel_stmt)
+            and ctx.implicit_limit
+            and isinstance(qlexpr, (qlast.PipelinedQuery, qlast.ShapeElement))
```
msullivan, line 1503: It looks like, starting with Python 3.10, it would actually work to keep saying `qlast.OffsetLimitMixin`! Maybe mypy doesn't believe that yet, though?

(Whether it actually works probably depends on the type. Maybe I need to make a mypy patch?)

edb/edgeql/ast.py
```diff
@@ -1578,3 +1575,15 @@ def has_ddl_subcommand(
     cmdtype: typing.Type[DDLOperation],
 ) -> bool:
     return bool(get_ddl_subcommand(ddlcmd, cmdtype))
+
+
+ReturningMixin = SelectQuery | ForQuery | InternalGroupQuery
```
msullivan, line 1580: I'd rename these, since they aren't mixins anymore.

msullivan (approved):
I think I mostly agree this is worth merging even though the original rationale isn't important.

A bit (but only a bit) more duplication in exchange for a simpler object model.

aljazerzen:
ast-fdp-refactored2

<details>
<summary>Code to generate the graph</summary>

```python

# Generates an inheritance graph of Python classes
# (using regex to extract it from .py file).
# 
# Usage:
# $ python gen_graph.py | fdp -T svg -o ast-fdp.svg
#
# Requirements:
# - graphviz

import re

with open('./edgedb/edb/edgeql/ast.py', 'r') as ast_file:
    ast_text = ast_file.read()

graph = {}

for match in re.finditer(
    r'class ([a-zA-Z_]+)\(((\n?\s*([a-zA-Z._]+),?)*)', ast_text
):
    deps = [
        dep.strip() for dep in match.group(2).split(',') if dep.strip() != ''
    ]
    graph[match.group(1)] = deps


hide_class = {
    'AlterObject',
    'CreateObject',
    'DropObject',
    'BasesMixin',
    'Base',
}

print('digraph G {')
for name, deps in graph.items():

    if name in hide_class:
        continue

    name = name.replace('.', '_')

    # if len(deps) < 2:
    #     continue

    for dep in deps:

        dep = dep.replace('.', '_')

        if dep in hide_class:
            continue

        print(f'  {name} -> {dep};')

print('}')

```

</details>

PR #5402, opened 2023-05-01
docs: Phase out legacy support

fantix:
* 3.0 drops support for pre-1.0 dump formats
* 3.0 deprecates support for pre-1.0 binary protocols


docs/changelog/3_x.rst
```diff
@@ -91,6 +91,12 @@ commands to use our testing channel and the beta release.
 
 This process will involve some downtime, specifically during steps 2 and 3.
 
+**Pre-1.0 Instances**
+
+If you're still running pre-1.0 EdgeDB instances like 1.0-beta3, and would like
+to upgrade to 3.0, you would need a dump-restore towards version 2.x first,
+followed by another dump-restore to 3.0.
```
raddevon, line 98:
```suggestion
If you're still running pre-1.0 EdgeDB instances (e.g., 1.0-beta3) and want
to upgrade to 3.0, you would need a dump/restore to version 2.x first,
followed by another dump/restore to 3.0.
```

Also, would it be more accurate to say you need to first **upgrade** to 2.x with a dump/restore and then **upgrade** to 3.0 with a dump/restore? I'm afraid leaving out the word "upgrade" might imply you can just dump/restore without an upgrade... unless that is your intent here.

docs/changelog/3_x.rst
```diff
@@ -515,3 +521,20 @@ Bug fixes
...
 
 * Resolver GROUP BY alised column (:eql:gh:`#4946`)
+
+
+Deprecations
+------------
+
+The support of version pre-1.0 binary protocol is deprecated in EdgeDB 3.0, and
+will be completely dropped in EdgeDB 4.0. If you're still using a deprecated
+version of the binary protocol, or any client libraries that *only* support the
```
raddevon, line 531:
```suggestion
version of the binary protocol or any client libraries that *only* support the
```

raddevon (approved):
A couple of minor suggestions. Looks great otherwise!

docs/changelog/3_x.rst
fantix: Right, it should be "upgrade" ❤️

PR #5404, opened 2023-05-02
Parser with chumksy

aljazerzen:
Edit: this PR is for the parser only. AST is in #5449

----

Original message:

The strategy is this:
- declare plain Rust structs and enums, using composition for doing sum types,
  ```rust
  # ast.rs
  struct PipelinedQuery {
    offset: Option<Expr>,
    limit: Option<Expr>,

    # this enum contains variants for inheritance children
    kind: PipelinedQueryKind,
  }

  enum PipelinedQueryKind {
    SelectQuery(SelectQuery),
    DeleteQuery(DeleteQuery),
  }

  struct SelectQuery {
     ...
  }
  ```
- parse with chumsky,
- convert the AST into `#[pyclass]`-annotated structs that implement `IntoPyObject` trait (see inheritance.rs). This step will be generated using derive macros (probably).
- generate `pyo3_example.pyi` file that contains correct type annotations (which are not present in `#[pyclass]`-annotated structs),
- generate the `#[pymodule]` function, to include the classes in the module, so we can do things like `isinstance(my_node, pyo3_example.PipelinedQuery)`,
- use the converted AST as regular Python classes in the compiler. If I understand correctly, CPython 3.11 optimizations should kick-in and we get PreeettyFastCompiler(tm).

I'll merge the new crate into edgeql-parser, as discussed. It's a standalone crate now because I needed to get familiar with PyO3 and maturin.


edb/compose-inherit/src/lib.rs
```diff
@@ -0,0 +1,39 @@
+mod py_children;
+mod py_inherit;
+
+use proc_macro::TokenStream;
+use quote::format_ident;
+use syn::parse_macro_input;
+
+/// Takes a struct declaration and emits:
+/// - struct declaration (basically) unchanged,
```
tailhook, line 9: what means "basically unchanged"? If it's unchanged, you don't need macro attribute, derive macro is okay (even emitting extra structures is fine)

edb/edgeql-chumsky/src/lib.rs
```diff
@@ -0,0 +1,53 @@
...
+    m.add_class::<crate::ast::QueryPyAttrs>()?;
+    m.add_class::<crate::ast::PipelinedQueryPyAttrs>()?;
+    m.add_class::<crate::ast::SelectQueryPyAttrs>()?;
+    Ok(())
+}
+
+#[pyfunction]
+fn parse(text: String) -> PyResult<PyObject> {
+    let mut token_stream = TokenStream::new(&text);
+    let mut tokens = Vec::new();
```
tailhook, line 28: Eventually have to get rid of `TokenStream` from `combine`. Initially I wanted to use `combine` from parser instead of chumsky. But that was long time ago. Probably `TokenStream` should emit something more chumsky-compatible and not buffer all the tokens into a vector.

tailhook (commented):
Few nits so far. I have to admit that proc macros are hard to review so I didn't do a deep review of that.

edb/compose-inherit/src/lib.rs
aljazerzen: It removes `#[pychildren]` attribute.

aljazerzen: Oh, I could just define the `#[pychildren]` macro as a no-op. And then it could really be unchanged. And then this could be a derive macro - if I didn't need the attributes.

aljazerzen:
Superseded by #5693

PR #5413, opened 2023-05-03
Parameterize schema for pg extension (uuid-ossp)

fantix:
This supports the latest Heroku where extensions must be created under the schema "heroku_ext" instead of our preferred "edgedbext".

### DIFF
--- a/edb/pgsql/compiler/config.py
+++ b/edb/pgsql/compiler/config.py
@@ -412,10 +412,7 @@ def _rewrite_config_insert(
 
     overwrite_query = pgast.SelectStmt()
     id_expr = pgast.FuncCall(
-        name=(
-            ctx.env.backend_runtime_params.instance_params.ext_schema,
-            'uuid_generate_v1mc',
-        ),
+        name=('edgedb', 'uuid_generate_v1mc'),
         args=[],
     )
     pathctx.put_path_identity_var(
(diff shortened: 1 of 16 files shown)


edb/pgsql/compiler/config.py
```diff
@@ -412,7 +412,10 @@ def _rewrite_config_insert(
 
     overwrite_query = pgast.SelectStmt()
     id_expr = pgast.FuncCall(
-        name=('edgedbext', 'uuid_generate_v1mc',),
+        name=(
+            ctx.env.backend_runtime_params.instance_params.ext_schema,
```
elprans, line 416: No need to use `instance_params` here, just use the wrapper in `edgedb` here.

fantix:
To get a Heroku Postgres at version 14:

```
heroku addons:create -a app_name heroku-postgresql:basic --version 14
```

edb/pgsql/metaschema.py
```diff
@@ -4409,9 +4409,54 @@ def __init__(self) -> None:
         )
 
 
+class UuidGenerateV1mcFunction(dbops.Function):
```
msullivan, line 4412: Is there any chance these extra wrappers will slow things down? Probably they can get inlined?
(Is preventing that why they are marked parallel_safe?)

msullivan (approved):
Great

edb/pgsql/metaschema.py
elprans: They should get inlined.  See criteria for inlining [here](https://github.com/postgres/postgres/blob/919c486a275bcdd83d2add77da87b6edbd91418c/src/backend/optimizer/util/clauses.c#L4401)

>  (Is preventing that why they are marked parallel_safe?)

No, the underlying function is marked as parallel-safe, so the wrapper should also.

PR #5417, opened 2023-05-04
Fix tuples in global variables and do error checks on arrays

### LINKED ISSUE #5356: ISE when providing tuple global
- EdgeDB Version: 3.0-dev.7404+local

Providing a tuple global value throws the error `InternalServerError: binary data has type 0 (-) instead of expected 25 (text) in record column 1`. (This error happens in both js and python clients, so I think it's a server bug instead of a problem with the client's encoding of globals).

Steps to Reproduce:

```js
const {createClient} = require("edgedb");

const client = createClient("_localdev");

(async () => {

  console.log(
    await client
      .withGlobals({"default::tupleGlobal": ["test", 132]})
      .query(`select global tupleGlobal`)
  );
  // InternalServerError: binary data has type 0 (-) instead of expected 25 (text) in record column 1

})();
```

```python
import edgedb

client = edgedb.create_client('_localdev')

print(
    client.with_globals({'default::tupleGlobal': ('test', 123)}).query('select global tupleGlobal')
)
# edgedb.errors.InternalServerError: binary data has type 0 (-) instead of expected 25 (text) in record column 1

client.close()
```

Schema:

```esdl
global default::tupleGlobal -> tuple<std::str, std::int64>;
```

msullivan:
We need to recode tuples to inject OIDs, like we do with arrays and
the like. Somewhat annoyingly, this ends up as a totally separate code
path from the existing re-encoding of tuples for parameters.

This is because for globals, which are schema declared, we have
postgres composite types, and it would be wasteful and convoluted
to split the tuples up instead of just using them.

And then, once we are injecting the OIDs, we can do missing error
checking that could allow invalid data in. (This was the main thing
that motivated fixing this now instead of deffering.)

Fixes #5356.

### DIFF
--- a/edb/server/compiler/sertypes.py
+++ b/edb/server/compiler/sertypes.py
@@ -1094,13 +1094,14 @@ def _make_global_rep(typ: s_types.Type, ctx: Context) -> object:
     if isinstance(typ, s_types.Tuple):
         subtyps = typ.get_subtypes(ctx.schema)
         return (
-            1,
+            int(enums.TypeTag.TUPLE),
             tuple(subtyp.id for subtyp in subtyps),
             tuple(_make_global_rep(subtyp, ctx) for subtyp in subtyps),
         )
     elif isinstance(typ, s_types.Array):
         subtyp = typ.get_element_type(ctx.schema)
-        return (2, subtyp.id, _make_global_rep(subtyp, ctx))
+        return (
+            int(enums.TypeTag.ARRAY), subtyp.id, _make_global_rep(subtyp, ctx))
     else:
         return None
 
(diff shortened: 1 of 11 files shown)


edb/server/compiler/sertypes.py
```diff
@@ -1090,6 +1090,21 @@ def _parse_range_descriptor(
     return RangeDesc(tid=tid, inner=subtype)
 
 
+def _make_global_rep(typ: s_types.Type, ctx: Context) -> object:
+    if isinstance(typ, s_types.Tuple):
+        subtyps = typ.get_subtypes(ctx.schema)
+        return (
+            1,
```
1st1, line 1097: Can you add an enum?

1st1 (approved):
Damn. LGTM

PR #5420, opened 2023-05-04
Add more 3.0 dump tests and fix reflected name of triggers

msullivan:
The *first* instance of triggers in the schema got it right, and then
d4a5a1676e5c927c5df51110b582c72e8734169f factored out duplicated name
handling but forgot to use the factored out macinery in triggers!


aljazerzen (approved):
Introspection tests. Nice.

PR #5422, opened 2023-05-04
Increase the line length limit from 79 to 80

msullivan:
This is very marginal, and because it so marginal and because 80 is a
nice round number, I expect that nobody will spend any of their
"one wild and precious life" arguing about it.


aljazerzen (approved):
Great quality of life improvement.

PR #5424, opened 2023-05-04
Try to make the NOCREATEDB test suite pass

msullivan:
Refresh the connection to avoid a state mismatch after restoring from dump.

Fix the explain tests to not depend on a bunch of cost numbers that can't be relied on
(though that just came up on heroku, not in non-heroku NOCREATEDB).

Skip the SQL pg_dump tests.

### DIFF
--- a/.github/workflows/tests-managed-pg.yml
+++ b/.github/workflows/tests-managed-pg.yml
@@ -326,7 +326,7 @@ jobs:
           submodules: false
 
       - name: Setup Terraform
-        uses: hashicorp/setup-terraform@v1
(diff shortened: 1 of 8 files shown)


.github/workflows/tests-managed-pg.yml
fantix, line 1: This also has a tpl file

msullivan: oh these changes were just picked up by make; guess a previous change never got regenereated

elprans (approved):

PR #5425, opened 2023-05-04
Support disabling dynamic configuration of dangerous config

### LINKED ISSUE #5235: Add a command-line option to disable dynamic configuration (via `configure`) of instance-level config values
Related: #5234, #5233

fantix:
Fixes #5235 

- [x] Manual test
- [x] Add test code
- [ ] Test cloud dump/restore

### DIFF
--- a/edb/lib/cfg.edgeql
+++ b/edb/lib/cfg.edgeql
@@ -34,11 +34,6 @@ CREATE ABSTRACT INHERITABLE ANNOTATION cfg::requires_restart;
 # in the binary protocol state.
 CREATE ABSTRACT INHERITABLE ANNOTATION cfg::system;
 
-# Changing dangerous config values may cause the database work very
-# differently. You may lose connectivity, consume lots of hardware resources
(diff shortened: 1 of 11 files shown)


edb/lib/cfg.edgeql
```diff
@@ -20,10 +20,25 @@
...
+
+# System config means that config value can only be modified using
+# CONFIGURE INSTANCE command. System config is therefore *not* included
+# in the binary protocol state.
 CREATE ABSTRACT INHERITABLE ANNOTATION cfg::system;
+
+# Changing dangerous config values may cause the database work very
+# differently. You may lose connectivity, consume lots of hardware resources
+# or end up with not enough resources to run normally, if misconfigured.
+CREATE ABSTRACT INHERITABLE ANNOTATION cfg::dangerous;
```
elprans, line 40: Is `cfg:::dangerous` == `cfg::system` and if so, should we just use the latter?

msullivan (approved):
Looks good!

PR #5435, opened 2023-05-05
backport: Add support for JWT allowlist and revocation list (#5226)

elprans:
The new `--jwt-sub-allowlist-file` server argument allows providing a
list of subjects whose keys are authorized, and the new
`--jwt-revocation-list-file` server argument allows providing a list of
key ids that are explicitly revoked.

### DIFF
--- a/edb/server/args.py
+++ b/edb/server/args.py
@@ -200,6 +200,8 @@ class ServerConfig(NamedTuple):
     jws_key_file: pathlib.Path
     jwe_key_file: pathlib.Path
     jose_key_mode: JOSEKeyMode
+    jwt_sub_allowlist_file: Optional[pathlib.Path]
+    jwt_revocation_list_file: Optional[pathlib.Path]
 
     default_auth_method: ServerAuthMethods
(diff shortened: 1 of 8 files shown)


edb/server/args.py
```diff
@@ -815,6 +817,26 @@ def resolve_envvar_value(self, ctx: click.Context):
              '--runstate-dir.\n\nThe default is "require_file" when the '
              '--security option is set to "strict", and "generate" when the '
              '--security option is set to "insecure_dev_mode"'),
+    click.option(
+        '--jwt-sub-allowlist-file',
```
1st1, line 821: nit:

```suggestion
        '--jwt-sub-allow-list-file',
```

to look nice along `--jwt-revocation-list-file`

1st1 (approved):

edb/server/args.py
elprans: Allowlist is a single word (instead of whitelist).  Also, this is a backport from master.

fantix (approved):

PR #5442, opened 2023-05-08
Enable GraphQL support for type unions.

### LINKED ISSUE #3851: GraphQL doesn't work on schemas that have union links
I tried to fire up the graphql client on our `issues` database, and it fails with

```
    /home/sully/src/e/edgedb/edb/server/compiler_pool/worker.py, line 244, in compile_graphql
        > return graphql.compile_graphql(
    /home/sully/src/e/edgedb/edb/graphql/compiler.py, line 85, in compile_graphql
        > op = graphql.translate_ast(
    /home/sully/src/e/edgedb/edb/graphql/translator.py, line 1807, in translate_ast
        > graphql.validate(gqlcore.graphql_schema, document_ast),
    /home/sully/src/e/edgedb-dev/lib/python3.10/site-packages/graphql/validation/validate.py, line 44, in validate
        > assert_valid_schema(schema)
    /home/sully/src/e/edgedb-dev/lib/python3.10/site-packages/graphql/type/validate.py, line 91, in assert_valid_schema
        > raise TypeError("\n\n".join(error.message for error in errors))

TypeError: Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but '(default:File | default:Publication | default:URL)' does not.

Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but 'Order(default:File | default:Publication | default:URL)' does not.

Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but 'Filter(default:File | default:Publication | default:URL)' does not.

Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but 'NestedFilter(default:File | default:Publication | default:URL)' does not.

Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but 'NestedInsert(default:File | default:Publication | default:URL)' does not.

Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but 'NestedUpdate(default:File | default:Publication | default:URL)' does not.

Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but '(default:File | default:Publication | default:URL)' does not.
```

It looks like no graphql queries against the database can work, even ones not involving those fields 

Looks like we'll have to do some sort of name mangling?

vpetrovykh:
In EdgeDB a type union includes all the common fields of the individual member types. This is different from a GraphQL union type which has no fields at all. Therefore it's more accurate to use GraphQL interfaces to reflect type unions from EdgeDB.

Fixes #3851

### DIFF
--- a/edb/graphql/types.py
+++ b/edb/graphql/types.py
@@ -341,6 +341,7 @@ class GQLCoreSchema:
         } - HIDDEN_MODULES))
 
         self._gql_interfaces = {}
+        self._gql_uniontypes: Set[s_name.QualName] = set()
         self._gql_objtypes_from_alias = {}
         self._gql_objtypes = {}
         self._gql_inobjtypes = {}
@@ -396,16 +397,26 @@ class GQLCoreSchema:
 
     def get_gql_name(self, name: s_name.QualName) -> str:
         module, shortname = name.module, name.name
+
+        # Adjust the shortname.
+        if shortname.startswith('__'):
+            # Use '_edb' prefix to mark derived and otherwise
+            # internal types. We opt out of '__edb' because we
+            # still rely on the first occurrence of '__' in
+            # GraphQL names to separate the module from the rest
+            # of the name in some code.
+            shortname = f'_edb{shortname}'
+        elif shortname.startswith('('):
+            # Looks like a union type, so we'll need to process individual
+            # parts of the name.
+            names = []
+            for part in shortname[1:-1].split(' | '):
+                names.append(
+                    self.get_gql_name(s_name.QualName(*part.split(':', 1))))
+            shortname = '_OR_'.join(names)
+
         if module in {'default', 'std'}:
-            if shortname.startswith('__'):
-                # Use '_edb' prefix to mark derived and otherwise
-                # internal types. We opt out of '__edb' because we
-                # still rely on the first occurrence of '__' in
-                # GraphQL names to separate the module from the rest
-                # of the name in some code.
-                return '_edb' + shortname
-            else:
-                return shortname
+            return shortname
         else:
             assert module != '', f'get_gl_name {name=}'
             return f'{module}__{shortname}'
@@ -608,6 +619,9 @@ class GQLCoreSchema:
                 # (e.g. nested aliased type), which should not be
                 # exposed as a top-level query option.
                 if name in TOP_LEVEL_TYPES or gqliface.name.startswith('_edb'):
+                    continue
+                # Check that the underlying type is not a union type.
+                if name in self._gql_uniontypes:
                     continue
                 fields[gqliface.name] = GraphQLField(
                     GraphQLList(GraphQLNonNull(gqliface)),
@@ -1232,6 +1246,7 @@ class GQLCoreSchema:
     def _define_types(self) -> None:
         interface_types = []
         obj_types = []
+        from_union = {}
 
         self.define_enums()
         self.define_generic_filter_types()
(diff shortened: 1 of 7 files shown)


edb/graphql/types.py
```diff
@@ -396,16 +397,26 @@ def graphql_schema(self) -> GraphQLSchema:
...
+
+        # Adjust the shortname.
+        if shortname.startswith('__'):
+            # Use '_edb' prefix to mark derived and otherwise
+            # internal types. We opt out of '__edb' because we
+            # still rely on the first occurrence of '__' in
+            # GraphQL names to separate the module from the rest
+            # of the name in some code.
+            shortname = f'_edb{shortname}'
+        elif shortname.startswith('('):
```
msullivan, line 409: This is obviously kind of hacky. If it would be easy to actually be looking at the type instead of just the name, that would be an improvement, but if not then hacky isn't the end of the world.

edb/graphql/types.py
```diff
@@ -1278,6 +1293,29 @@ def _type_resolver(
                     description=self._get_description(t),
                 )
 
+            if t.is_union_type(self.edb_schema):
+                # NOTE: EdgeDB union types and GraphQL union types are
+                # different in some important ways. In EdgerDB a union object
```
msullivan, line 1298: s/EdgerDB/EdgeDB/

tests/test_http_graphql_schema.py
```diff
@@ -2495,6 +2505,404 @@ def test_graphql_schema_type_19(self):
             }
         })
 
+    def test_graphql_schema_type_20(self):
+        # make sure the union type got reflected
+        self.assert_graphql_query_result(r"""
+        query {
+            __type(name: "Profile_or_Setting") {
```
msullivan, line 2512: Do we really want to support this? Supporting it as a link target is important but supporting it directly isn't really.
Or does graphql basically require we support it if we support it as a link target?

edb/graphql/types.py
```diff
@@ -396,16 +397,26 @@ def graphql_schema(self) -> GraphQLSchema:
...
+            # of the name in some code.
+            shortname = f'_edb{shortname}'
+        elif shortname.startswith('('):
+            # Looks like a union type, so we'll need to process individual
+            # parts of the name.
+            names = []
+            for part in shortname[1:-1].split(' | '):
+                names.append(
+                    self.get_gql_name(s_name.QualName(*part.split(':', 1))))
+            shortname = '_or_'.join(names)
```
msullivan, line 416: Should we mangle this more somehow? This seems like it could plausibly collide

tests/test_http_graphql_schema.py
vpetrovykh: This is the type that is returned by the link.

edb/graphql/types.py
vpetrovykh: Realistically, any readable GraphQL reflected name can collide, so we're already relying on unlikelihood of collisions. Eventually, in the long term, we hope to have a way to remap entire schemas (possibly via the mechanism we intend primarily to be used for forms). Generally, once remapping is possible we don't need to care as much about collisions.

tests/test_http_graphql_schema.py
msullivan: Right, but does that mean we need to actually expose it as a standalone thing that you can select directly?

vpetrovykh: This test checks the available types, which should exist. The `INTERFACE` is basically an abstract return type.

There's another test a bit later `test_graphql_reflection_03` that actually checks that the corresponding queries or mutations don't get implemented, which is how we know it's not available for direct querying.

msullivan (approved):
Let's make it `_OR_` instead of `_or_` and call it a day.

Looks good!

PR #5448, opened 2023-05-09
Fix a potential typo in docs

KyleXie:
* the `filter .ilike ...` expression seems should be `filter .title ilike ...`


elprans (approved):
Thanks!

PR #5449, opened 2023-05-09
Generate EdgeQL rust AST and IntoPython trait

aljazerzen:


edb/edgeql-parser/src/ast.rs
```diff
@@ -0,0 +1,2633 @@
+// DO NOT EDIT. This file was generated with:
```
tailhook, line 1: Is it temporary, while we have not merged that to master. Or do you expect python to be source of truth for the long time?

tailhook (commented):
LGTM

edb/edgeql-parser/src/ast.rs
aljazerzen: That's here to stay as long as we have the eql compiler written in Python. That's because we want to have Python definitions of the AST so mypy works correctly and because we want to be able to construct AST nodes from Python too.

We could generate Python definitions from Rust ones, but from my experiments, it's easier to do it this way around.

edb/tools/gen_rust_ast.py
```diff
@@ -0,0 +1,274 @@
...
+            '''
+        )
+    )
+
+    # discover all nodes
+    for name, typ in qlast.__dict__.items():
+        if not isinstance(typ, type) or not hasattr(typ, '_direct_fields'):
+            continue
+
+        if hasattr(typ, '__rust_ignore__') and typ.__rust_ignore__:
```
msullivan, line 62: I'd probably just use `getattr`?

edb/tools/gen_rust_ast.py
```diff
@@ -0,0 +1,274 @@
...
+            typ = f'Box<{typ}>'
+
+        f_name = quote_rust_ident(f.name)
+        field_names.add(f_name)
+
+        fields += f'    pub {f_name}: {typ},\n'
+
+    if len(cls.children) > 0:
+
+        for i in range(0, 10):
```
msullivan, line 116: using `range(0, 10)` is hacky; try `itertools.count` if you don't want to write a while loop

edb/tools/gen_rust_ast.py
```diff
@@ -0,0 +1,274 @@
...
+        typ, set(f.name for f in typ._fields.values() if not f.hidden)
+    )
+    return ancestor.__name__
+
+
+def find_covering_ancestor(typ: typing.Type, fields: typing.Set[str]):
+    # In Rust, a type will not inherit fields from parent types.
+    # This means that we need to omit some ancestor of this type, which
+    # would include all fields of the type.
+    # We loose a bit of type checking strictness here.
```
msullivan, line 266: lose

edb/tools/gen_rust_ast.py
```diff
@@ -0,0 +1,274 @@
...
+    for parent in typ.__mro__:
+        if not hasattr(parent, '_direct_fields'):
+            continue
+
+        fields = fields.difference((f.name for f in parent._direct_fields))
+        if len(fields) == 0:
+            return parent
+    raise AssertionError()
```
msullivan, line 274: give it a message?

edb/tools/gen_rust_ast.py
```diff
@@ -0,0 +1,274 @@
...
+        )
+        return union_name
+
+    if typing_inspect.is_generic_type(typ) and hasattr(typ, '_name'):
+
+        if typ._name in ('List', 'Sequence'):
+            return f'Vec<{params[0]}>'
+
+        if typ._name == 'Dict':
+            return f'HashMap<{params[0]}, {params[1]}>'
```
msullivan, line 226: We need to make sure to use an order preserving map. Probably https://docs.rs/indexmap/latest/indexmap/.

msullivan (approved):
This seems basically reasonable, though I don't have any experience with rust macros.

How tested is it?

aljazerzen:
> This seems basically reasonable, though I don't have any experience with rust macros.
> 
> How tested is it?

If you mean "how tested are Rust macros", the answer is very.

If you mean "how tested is our derive macro + ast generation", the answer is not at all. But the parser syntax tests should test all nodes and conversions to Python.

edb/tools/gen_rust_ast.py
aljazerzen: Let's stick with [`std::collections::BTreeMap`](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html) for now. It has a bit worse performance, but it's in `std` so we avoid a dependency.

It's easy to switch anyway.

tailhook: BTreeMap is not order-preserving. Its sorted.

(switch might not be easy, as there are different traits on the key of BTreeMap vs Hash/Indexmap, although they are probably all strings in AST)

aljazerzen: Oh, that easy to mix up :D It's indexmap then.

PR #5454, opened 2023-05-09
Handle error of too many constraint args

### LINKED ISSUE #5326: InternalServerError on migration with new constraint
I figured this constraint probably wouldn't work but wanted to try it anyway. It gives me an ISE.

- EdgeDB Version: 2.13+ad0eb0f
- EdgeDB CLI Version: 3.0.0-dev.946+93cec91
- OS Version: macOS 13.2.1

Steps to Reproduce:

1. Create a project with `edgedb project init`
2. Add the problematic schema to `dbschema/default.esdl`
3. Attempt to create migration with `edgedb migration create`

Schema:

```
module default {
  type Cause {
    required multi link charities -> Charity {
      property weight -> float64;
      constraint expression (sum(.charities@weight) <= 10);
    }
  }
  type Charity {
  }
}
```

aljazerzen:
Closes #5326

We do already check that "constraint on object types require ON clause", but this will guide users in all cases.

### DIFF
--- a/edb/edgeql/utils.py
+++ b/edb/edgeql/utils.py
@@ -23,6 +23,7 @@ import copy
 import itertools
 from typing import *
 
+from edb import errors
 from edb.common import ast
 from edb.schema import schema as s_schema
 from edb.schema import functions as s_func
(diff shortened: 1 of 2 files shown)


edb/edgeql/utils.py
```diff
@@ -80,10 +81,21 @@ def index_parameters(
...
+    params = parameters.objects(schema)
+
+    if not variadic and len(ql_args) > len(params):
+        # In error message we discount the implicit __subject__ param.
+        raise errors.SchemaDefinitionError(
+            f'Expected {len(params) - 1} arguments, but found '
+            f'{len(ql_args) - 1}',
+            context=ql_args[-1].context,
+            details='Did you mean to use ON (...) for specifying the subject?',
+        )
```
msullivan, line 93: Consider moving this to the call-site somehow?

msullivan (approved):
Looks good

PR #5455, opened 2023-05-09
Update documentation to reflect 2.0 has released

raddevon:
I went in with the goal of updating the language suggesting 2.0 is a future release. Ended up doing a quick tune-up while I was in here too.

### DIFF
--- a/docs/clients/graphql/cheatsheet.rst
+++ b/docs/clients/graphql/cheatsheet.rst
@@ -5,13 +5,13 @@ Cheatsheet
 
 .. note::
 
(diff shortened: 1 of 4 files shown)


raddevon:
I ended up removing the section about "Backendless mode" since the tools we have now still don't allow for a truly backendless mode in the way a user might expect. You can run with a fairly minimal backend, but you still need authentication and some way to get user IDs into your globals.

See my recent discussion with a user for more context: https://github.com/edgedb/edgedb/discussions/5468

elprans (approved):

PR #5462, opened 2023-05-10
Translate generated pgsql error locations to input text

efaust:
When a user writes some error-ful query into psql, we are currently returning
errors in terms of the generated query. This is unsightly and hard to follow:

```
edgedb=# select 1 + 'goo';
ERROR:  invalid input syntax for type integer: "goo"
LINE 2:     (1 + 'goo')
                 ^
QUERY:  SELECT/*<pg.SelectStmt at 0x1213e37c0>*/
    (1 + 'goo')
```

Instead, return error locations against the input query. This yields:

```
edgedb=# select 1 + 'goo';
ERROR:  invalid input syntax for type integer: "goo"
LINE 1: select 1 + 'goo';
                   ^
QUERY:  select 1 + 'goo';
```

Much better.

### DIFF
--- a/edb/common/ast/base.py
+++ b/edb/common/ast/base.py
@@ -243,7 +243,10 @@ class AST:
     def __copy__(self):
         copied = self.__class__()
         for field, value in iter_fields(self, include_meta=False):
-            object.__setattr__(copied, field, value)
+            try:
+                object.__setattr__(copied, field, value)
+            except AttributeError:
+                continue
         return copied
 
     def __deepcopy__(self, memo):
--- a/edb/common/ast/base.py
+++ b/edb/common/ast/base.py
@@ -246,6 +246,7 @@ class AST:
             try:
                 object.__setattr__(copied, field, value)
             except AttributeError:
+                # don't mind not setting getter_only attrs.
                 continue
         return copied
 
--- a/edb/common/ast/codegen.py
+++ b/edb/common/ast/codegen.py
@@ -60,7 +60,7 @@ class SourceGenerator(NodeVisitor):
         if indent:
             self.new_lines = 1
             self.char_indentation += 1
-        res = super().visit(node)
+        res = self.visit(node)
         if indent:
             self.char_indentation -= 1
         if nest:
--- a/edb/pgsql/ast.py
+++ b/edb/pgsql/ast.py
@@ -669,8 +669,9 @@ class Expr(ImmutableBaseExpr):
 class BaseConstant(ImmutableBaseExpr):
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
-        if not isinstance(self, NullConstant) and self.val is None:
-            raise ValueError('cannot create a pgast.Constant without a value')
+        '''if not isinstance(self, NullConstant) and self.val is None:
+            raise ValueError('cannot create a pgast.Constant without a
+            value')'''
 
 
 class StringConstant(BaseConstant):
--- a/edb/pgsql/ast.py
+++ b/edb/pgsql/ast.py
@@ -667,11 +667,7 @@ class Expr(ImmutableBaseExpr):
 
 
 class BaseConstant(ImmutableBaseExpr):
-    def __init__(self, **kwargs):
-        super().__init__(**kwargs)
-        '''if not isinstance(self, NullConstant) and self.val is None:
-            raise ValueError('cannot create a pgast.Constant without a
-            value')'''
+    pass
 
 
 class StringConstant(BaseConstant):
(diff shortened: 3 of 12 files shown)


edb/common/ast/base.py
```diff
@@ -243,7 +243,10 @@ def __init__(self, **kwargs):
     def __copy__(self):
         copied = self.__class__()
         for field, value in iter_fields(self, include_meta=False):
-            object.__setattr__(copied, field, value)
+            try:
```
efaust, line 246: Some fields are readonly, and we hit this in the impl of `.replace()`

Pretty scary/ugly to skip here.

edb/pgsql/ast.py
```diff
@@ -667,10 +667,7 @@ class Expr(ImmutableBaseExpr):
 
 
 class BaseConstant(ImmutableBaseExpr):
-    def __init__(self, **kwargs):
-        super().__init__(**kwargs)
-        if not isinstance(self, NullConstant) and self.val is None:
-            raise ValueError('cannot create a pgast.Constant without a value')
+    pass
```
efaust, line 670: Again, `.replace()` calls __init__(), and then copies the fields across, so we lose our pretty safety check.

edb/pgsql/codegen.py
```diff
@@ -117,6 +145,57 @@ def ctes_to_source(cls, ctes: List[pgast.CommonTableExpr]) -> str:
...
+    ) -> None:
+        start = len(self.result)
+        super().write(*x, delimiter=delimiter)
+        for new in range(start, len(self.result)):
+            self.write_index += len(self.result[new])
+
+    def visit(self, node):  # type: ignore
+        if self.with_translation_data:
+            translation_data = SQLSourceGeneratorTranslationData(
+                source_start=node.context.start if node.context else 0,
```
efaust, line 186: This needs a comment. Only the outermost statement lacks a context.

edb/common/ast/codegen.py
```diff
@@ -60,7 +60,7 @@ def visit_indented(
         if indent:
             self.new_lines = 1
             self.char_indentation += 1
-        res = super().visit(node)
+        res = self.visit(node)
```
efaust, line 63: _please_ someone check this. This looks right to me, but it also scares the bejeezus out of me.

edb/common/ast/base.py
msullivan: Add a comment

edb/server/compiler/compiler.py
```diff
@@ -668,8 +668,14 @@ def parse_search_path(search_path_str: str) -> list[str]:
                     **args
                 )
                 resolved = pg_resolver.resolve(stmt, schema, options)
-                source = pg_codegen.generate_source(resolved)
-                unit = dbstate.SQLQueryUnit(query=source)
+                source, tl_data = \
```
msullivan, line 671: Use parens instead of a backslash. I hate backslashes

edb/common/ast/codegen.py
msullivan: This seems like it should be right, to me.

But also some tests are failing with syntax errors so maybe we're wrong

msullivan (commented):
Looks pretty good!

Needs tests, and it seems to have broken some existing tests legitimately

edb/server/pgcon/pgcon.pyx
```diff
@@ -1667,13 +1667,21 @@ cdef class PGConnection:
                             field_type = self.buffer.read_byte()
                             if field_type == b'P':  # Position
                                 msg_buf.write_byte(b'q')  # Internal query
-                                msg_buf.write_bytestring(query)
+                                msg_buf.write_bytestring(unit.input_query.encode('utf8'))
```
fantix, line 1670: After translation, we don't need to include the original query with "internal query" (`q`) any more - we can simply return the translated data with the "position" (`P`), instead of the "internal position" (`p`). I believe psql already has the SQL and it could produce a proper error output.

fantix (approved):
Other than the issue above, I think this is good!

msullivan (approved):
Looks good to me! I think fantix is right about position v. internal position, so follow up on that. (Which might need a test change.)

efaust:
Updated for extended queries. Would like @fantix to take another look to make sure I'm not crazy

edb/server/pgcon/pgcon.pyx
```diff
@@ -2062,12 +2086,17 @@ cdef class PGConnection:
...
                 if field_type == b'P':  # Position
-                    # Internal query
-                    msg_buf.write_byte(b'q')
-                    # Compiled SQL
-                    msg_buf.write_bytestring(action.args[0])
-                    # Internal position
-                    msg_buf.write_byte(b'p')
+                    qu = (action.query_unit.translation_data
+                          if isinstance(action.query_unit,
+                                        compiler.dbstate.SQLQueryUnit)
```
fantix, line 2091: `action.query_unit` has type `SQLQueryUnit | None`, so I think just checking `is None` is sufficient here.

fantix (approved):
Looks good to me!

PR #5464, opened 2023-05-10
Document rounding for float to int casts

raddevon:
Addresses a question posed by @msullivan on Slack a while back (while I was focused on 3.0 docs).

### DIFF
--- a/docs/reference/edgeql/casts.rst
+++ b/docs/reference/edgeql/casts.rst
@@ -148,4 +148,8 @@ Casting Table
 
 - ``<>`` - can be cast explicitly
 - ``:=`` - assignment cast is supported
-- "impl" - implicit cast is supported
+- ``impl`` - implicit cast is supported
+- ``*``- When casting a float type to an integer type, the fractional value
+  naturally cannot be preserved after the cast. When executing this cast, we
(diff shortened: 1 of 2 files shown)


docs/reference/edgeql/casts.rst
```diff
@@ -148,4 +148,8 @@ Casting Table
 
 - ``<>`` - can be cast explicitly
 - ``:=`` - assignment cast is supported
-- "impl" - implicit cast is supported
+- ``impl`` - implicit cast is supported
+- ``*``- When casting a float type to an integer type, the fractional value
+  naturally cannot be preserved after the cast. When executing this cast, we
+  round to the nearest integer, rounding up on a tie (i.e., values of .5 and
```
msullivan, line 154: It looks like it actually does "round to even" on a tie:
```
scope> select <int64>{0.5, 1.5, 2.5, 3.5, 4.5, 5.5, 6.5};
[0, 2, 2, 4, 4, 6, 6]
```

msullivan (approved):
one correction, but good

docs/reference/edgeql/casts.rst
raddevon: Thanks for catching that, @msullivan!

PR #5465, opened 2023-05-10
Use setup-python cache and bump add-and-commit

scotttrinh:
Due to the soon-to-be [deprecated GitHub Actions commands](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/), bump a few dependencies that were using the `save-state` and `set-output` commands.

- https://github.com/syphar/restore-virtualenv is no longer maintained and has a note that "Most of its functionality was added to the standard setup-python action" with a link to the `setup-python` docs for caching. This change follows the patterns outlined there.
- https://github.com/endbug/add-and-commit fixed this in v9

I'm not super familiar with `pip`, but it seemed straightforward enough of a translation.


.github/workflows/test-pool.yml
```diff
@@ -33,22 +33,20 @@ jobs:
...
     - name: Generate requirements.txt
       run: |
         echo 'uvloop==0.15.2' > requirements.txt
         mkdir -p pool-simulation/reports
 
-    - name: Handle virtualenv
-      uses: syphar/restore-virtualenv@v1.1
-      id: venv-cache
+    - name: Set up Python
+      id: py310
```
elprans, line 42:
```suggestion
      id: python
```

so that we don't forget/have to change this every time we bump the Python version

.github/workflows/test-pool.yml
```diff
@@ -60,7 +58,7 @@ jobs:
       run: |
         python tests/test_server_pool.py
 
-    - uses: EndBug/add-and-commit@v7.0.0
+    - uses: EndBug/add-and-commit@v9.1.3
```
elprans, line 61: Let's use this opportunity to use the commit hash per actions best practices (and indicate the tag in a comment)

msullivan (commented):
Also, to make this all more fun:
Most of the files in `workflows` are generated files, created using jinga templating from files in `workflows.src`.
So `tests.yml` is generated, and you want to edit the template files in `workflows.src/` instead and then run `make`.

(We should probably also make it so that *all* of the files are generated, and that they include a header to that effect!)

PR #5472, opened 2023-05-11
Make coalesce infer a union type

### LINKED ISSUE #5407: Trying to assign result of two coalesced select queries to a union type link throws an InvalidLinkTargetError
- EdgeDB Version: 2.13+c171f92
- EdgeDB CLI Version: 2.3.1+ef99779
- OS Version: Windows 10 Home build 19045.2846

I'm trying to update a union type link based on whichever of the two types contains the record I'm looking for

Schema:

```
# Vehicles
abstract type Vehicle {
    required property model -> str { constraint exclusive; }
}
type Motorcycle extending Vehicle {
    required property geared -> bool;
}
type Car extending Vehicle {
    required property numOfDoors -> int16;
}
type Bicycle extending Vehicle {
    required property frameMaterial -> str;
}

# Drivers/riders
abstract type Driver {
    required link vehicle -> Vehicle;
}
type CarDriver extending Driver {
    overloaded required link vehicle -> Car;
}
type MotorcycleRider extending Driver {
    overloaded required link vehicle -> Motorcycle;
}
type Cyclist extending Driver {
    overloaded required link vehicle -> Bicycle;

    # Don't want to use the common abstract type Vehicle because this should not include other Bicycles
    link hobbyVehicle -> Car | Motorcycle; 
}
```

Query:
```
with
  hobbyVehicleModel := <str>$hobbyVehicleModel
update Cyclist
filter .vehicle.id = <uuid>$lorryId
set {
  hobbyVehicle := (
    # Try to look in both types to find the model that matches 🔍
    (select Motorcycle filter .model = hobbyVehicleModel) ?? (select Car filter .model = hobbyVehicleModel)
    # Also tried: `(select Vehicle filter .model = vehicleModel and (Vehicle is Car or Vehicle is Motorcycle))`
  )
}
```
The above update statement fails with error
```
InvalidLinkTargetError: invalid target for link 'hobbyVehicle' of object type 'default::Cyclist': 'default::Vehicle' (expecting 'default::Car | default::Motorcycle')
```
Even though as far as I can tell this should work, since the `select` query should return a record of type `Car | Motorcycle`, so why does it think I'm trying to assign the link a value of type `Vehicle`? 
But either way, how would I get this working without needing to run a separate update statement to try to find each of the allowed subtypes separately?

msullivan:
UNION and IF already do, coalesce is a good candidate to as well.
Started a type inference test suite to test it easily.

Fixes #5407.

### DIFF
--- a/edb/edgeql/compiler/func.py
+++ b/edb/edgeql/compiler/func.py
@@ -551,13 +551,15 @@ def compile_operator(
         ctx=ctx,
     )
 
-    if str_oper_name in {'std::UNION', 'std::IF'} and rtype.is_object_type():
+    if str_oper_name in {
+        'std::UNION', 'std::IF', 'std::??'
+    } and rtype.is_object_type():
         # Special case for the UNION and IF operators, instead of common
         # parent type, we return a union type.
-        if str_oper_name == 'std::UNION':
-            larg, rarg = (a.expr for a in final_args)
-        else:
+        if str_oper_name == 'std::IF':
             larg, _, rarg = (a.expr for a in final_args)
+        else:
+            larg, rarg = (a.expr for a in final_args)
 
         left_type = setgen.get_set_type(larg, ctx=ctx)
         right_type = setgen.get_set_type(rarg, ctx=ctx)
--- a/edb/edgeql/compiler/func.py
+++ b/edb/edgeql/compiler/func.py
@@ -554,7 +554,7 @@ def compile_operator(
     if str_oper_name in {
         'std::UNION', 'std::IF', 'std::??'
     } and rtype.is_object_type():
-        # Special case for the UNION and IF operators, instead of common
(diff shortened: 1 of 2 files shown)


edb/edgeql/compiler/func.py
```diff
@@ -551,13 +551,15 @@ def compile_operator(
         ctx=ctx,
     )
 
-    if str_oper_name in {'std::UNION', 'std::IF'} and rtype.is_object_type():
+    if str_oper_name in {
+        'std::UNION', 'std::IF', 'std::??'
+    } and rtype.is_object_type():
         # Special case for the UNION and IF operators, instead of common
         # parent type, we return a union type.
```
aljazerzen, line 558: Outdated comment

aljazerzen (approved):

Arrow7000:
Thanks so much for doing this work! Do you know when this fix will be released?

Issue #5478, opened 2023-05-12
pgvector support


msullivan:
We want to be able to support storing vectors from https://github.com/pgvector/pgvector

This probably means
- [ ] Make our extension system powerful enough to specify it as an optional extension package
- [ ] Write an extension package for pgvector
- [ ] Figure out how to package it for when running edgedb-managed postgres

msullivan:
I went and hacked up an initial take on enough of an extension packaging system to probably build this.

I also did a little bit of investigating into pgvector, so here are my very basic notes:
 * The pgvector build system uses `pg_config` to find what to build against and where to install, so I set `PATH=$HOME/src/e/edgedb/build/postgres/install/bin:$PATH` and then just ran `make && make install` and it installed into my edgedb dev build of postgres
 * Probably for dev builds we will just want to add pgvector as a git submodule and then build it as described above. It's under the PostgreSQL license so we can bundle it without any trouble
 * The examples in their README all give a vector length in the types (`vector(3)`, etc`), but just using `vector` without a length works just fine and so we should do that instead
 * By default, `to_json` just takes the text representation of a value and makes it a json string. In my extension system PR, I also made sure to make it so that giving a custom JSON cast works properly and that json output mode works with it, so I think we should consider having the JSON representation be a json array of numbers

elprans:
> pgvector as a git submodule

No need.  Just clone directly in `setup.py`.

PR #5479, opened 2023-05-12
Initial take on expanded extension package system

msullivan:
This should support enough machinery to allow defining opt-in extension
packages that wrap postgres types that come from extensions.

The major changes are:
* Extension packages have an optional list of postgres extensions to
  create
* Extension packages get an optional module name for things to go
* Actually run the script associated with a package when creating
  an extension
* Support specifying the name of an underlying SQL type for a scalar
  in DDL instead of needing to hardcode it in the compiler.
  (We could probably move all the hardcoding into the schema now,
  but I haven't yet.)
* Delete everything in the relevant module when deleting an extension

There are still some open questions:
* Currently I create a postgres schema with the name of the
  extension. This won't work on heroku, but do we care?
* How will we deal with exposing binary encodings to clients?
* Do we need to expose a mechanism for running some raw SQL,
  or at least for defining helper functions?

I've tested this by writing a very minimal extension package that wraps the
"ltree" extension that ships with postgres.

Work towards #5478.

### DIFF
--- a/edb/buildmeta.py
+++ b/edb/buildmeta.py
@@ -44,7 +44,7 @@ from edb.common import verutils
 
 
 # Increment this whenever the database layout or stdlib changes.
-EDGEDB_CATALOG_VERSION = 2023_05_04_00_02
+EDGEDB_CATALOG_VERSION = 2023_05_23_00_00
 EDGEDB_MAJOR_VERSION = 4
 
 
--- a/edb/edgeql/parser/__init__.py
+++ b/edb/edgeql/parser/__init__.py
@@ -127,6 +127,21 @@ def parse_migration_body_block(
     return parser.parse(tsource)
 
 
+def parse_extension_package_body_block(
+    source: str,
+) -> tuple[qlast.NestedQLBlock, list[qlast.SetField]]:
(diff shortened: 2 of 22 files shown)


elprans:
Another open question: how do we specify extension migrations?  Postgres ships a bunch of SQL files like `ext-1.0.0-1.0.1.sql`.  This also makes me think that entries in `sql_extensions` should contain version constraints.

msullivan:
`sql_extensions` can be treated as a list of names+versions using constraints, so you'll eventually be able to write 'pgvector~=2.0' as the dependency.
For how to resolve these constraints when executing `CREATE EXTENSION`, there are at least two options:
1. We can introspect `pg_available_extension_versions` in advance (at startup, probably) and make the decision at compilation time
2. We could generate pl/pgsql that introspects `pg_available_extension_versions` at runtime and generates the appropriate CREATE EXTENSION

We can do `CREATE EXTENSION PACKAGE MIGRATION` to register extension package migration scripts. Extensions would come bundled with scripts to create them.

I think none of this has any compatibility dangers.

elprans (approved):

PR #5487, opened 2023-05-15
Document Cloud CLI

raddevon:


raddevon:
Does `edgedb project` gain any Cloud-related superpowers I need to capture here? I included the ability to create a new Cloud instance in `edgedb instance create`, but I feel like I remember talk about something `project` can do here too.

fantix:
Yes, `edgedb project` can use the Cloud-format instance names just like regular instance names, and Cloud instances will be created/used/destroyed correspondingly.

raddevon:
Thanks, @fantix! Got that incorporated as well. I think this is ready for review when you have a chance. Please let me know if there's anything I've missed.

docs/cli/edgedb_instance/edgedb_instance_create.rst
```diff
@@ -21,6 +21,20 @@ instance and creating a corresponding credentials file in
...
+------------
+
+.. TODO: Cloud release
+.. Update this after Cloud has released
+
+Users with access to the EdgeDB Cloud beta may use this command to create a
+Cloud instance after logging in using :ref:`ref_cli_edgedb_cloud_login`.
+
+To create a Cloud instance, your instance name should be in the format
+``<github-org-name>/<instance-name>``. Cloud instance names may contain
```
fantix, line 34: Refs "org-name": I don't think we already supported GitHub organizations - at this moment it is just personal GitHub accounts. But we will support organizations in the future I believe, so perhaps we could adopt something more general here?

raddevon: Hmm. OK. I was thinking about labeling it instead `github-identified`, but then I'd feel the need to circle back and tell users we only support the username for now but will also support org name in the future. That will need to be changed once we support org name, so I think I might as well just call it `github-username` for now. Once we roll out support for organizations, I will circle back and change it to `github-identifier` and explain that can be either username or org name.

raddevon:
Got these updated. Your comments spilled over into the CLI repo since some of the descriptions were pulled directly out of the CLI help. I've updated them there as well in https://github.com/edgedb/edgedb-cli/pull/1038

I'll go ahead and merge this once CI is complete since I've addressed those unless you see something else in the meantime.

PR #5488, opened 2023-05-16
Draft a 3.0 RC1 release log

msullivan:
@raddevon, this has all the content but I'm not sure if you want it
to go somewhere else or be organized differently.

I separated fixes to 3.0 features from fixes to more general things,
on the theory that once 3.0 is actually out, the main page shouldn't
contain the former, but I'm not sure.

### DIFF
--- a/docs/changelog/3_x.rst
+++ b/docs/changelog/3_x.rst
@@ -848,3 +848,6 @@ Other changes and fixes
 
 * Work around postgres server crashes on Digital Ocean during edgedb setup
   (:eql:gh:`#5505`)
+
+* Always set cardinality of derived __tname__ and __tid__ pointers
+  (:eql:gh:`#5508`)
--- a/docs/changelog/3_x.rst
+++ b/docs/changelog/3_x.rst
@@ -780,7 +780,7 @@ Changes to new 3.0 features
 * Do nicer translation of relation names in ``ANALYZE``
   (:eql:gh:`#5467`)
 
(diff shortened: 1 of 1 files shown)


raddevon (approved):
This looks wonderful, @msullivan. I think this is exactly how we should organize the changes. I agree that we don't need fixes and changes inside pre-release 3.0 to be reflected in the public changelog for the final release.

Before we merge, I think we should also update installation instructions. I'm guessing `testing` is still the correct channel name (Is that correct?), but the version and package names in the installation instructions probably need updating. Here are the instances of those:

- https://github.com/edgedb/edgedb/pull/5488/files#diff-d0f8ef12b0710c19f70319fc83ae2550365a0626afcf5b923400895b781dbed6R27
- https://github.com/edgedb/edgedb/pull/5488/files#diff-d0f8ef12b0710c19f70319fc83ae2550365a0626afcf5b923400895b781dbed6R115
- https://github.com/edgedb/edgedb/pull/5488/files#diff-d0f8ef12b0710c19f70319fc83ae2550365a0626afcf5b923400895b781dbed6R121
- https://github.com/edgedb/edgedb/pull/5488/files#diff-d0f8ef12b0710c19f70319fc83ae2550365a0626afcf5b923400895b781dbed6R126

Not trying to put this on you. I'm happy to update them, but I want to make sure I'm adding the correct version numbers. If you're not sure, could you mention whoever might know, and I'll circle back once those answers are in.

PR #5495, opened 2023-05-17
Add extra path + shape explanation

Dhghomon:
Hi @raddevon,

I've been thinking about how to more easily explain backlinks and coincidentally someone on the Discord mentioned that the syntax looks a bit abnormal. I think the syntax is best explained showing an example of backlinks without an [is TypeName] and that they do work, but followed by an attempt to impose a shape that fails because it's a BaseObject. That leads into the type intersection operator which allows there to be a guaranteed shape, by which time the syntax is no longer so foreign.

I put a draft together that explains this and also adds a few inserts to make the output interesting for anyone following along. (Or maybe add the insert output into the post? Though doing that could get a bit long)

Any thoughts?

### DIFF
--- a/docs/edgeql/paths.rst
+++ b/docs/edgeql/paths.rst
@@ -24,6 +24,11 @@ Consider the following schema:
       required link author -> User;
     }
 
+    type Comment {
+      required property text -> str;
+      required link author -> User;
+    }
+
 .. code-block:: sdl
 
     type User {
@@ -35,6 +40,56 @@ Consider the following schema:
       required title: str;
       required author: User;
     }
+
+    type Comment {
+      required property text: str;
+      required link author: User;
+    }
+
+A few simple inserts will allow some experimentation with paths.
+
+Start with a first user:
+
+.. code-block:: edgeql
+  db> insert User {
+  email := "user1@email.com",
+  };
+
+Along comes another user who adds the first user as a friend:
+
+.. code-block:: edgeql
+
+  db> insert User {
+  email := "user2@email.com",
+  friends := (select detached User filter .email = "user1@email.com")
+  };
+
+The first user reciprocates, adding the new user as a friend:
+
+.. code-block:: edgeql
+
+  db> update User filter .email = "user1@email.com" 
+    set { 
+      friends += (select detached User filter .email = "user2@email.com")
+  };
+
+The second user writes a blog post about how nice EdgeDB is:
+
+.. code-block:: edgeql
+
+  db> insert BlogPost {
+  title := "EdgeDB is awesome",
+  author := assert_single((select User filter .email = "user2@email.com"))
+  };
+
+And the first user follows it up with a comment below the post:
+
+.. code-block:: edgeql
+
+  db> insert Comment {
+    text := "Nice post, user2!",
+    author := assert_single((select User filter .email = "user1@email.com"))
+  };
 
 The simplest path is simply ``User``. This is a :ref:`set reference
 <ref_eql_set_references>` that refers to all ``User`` objects in the database.
@@ -51,10 +106,13 @@ friend of another User*.
   select User.friends;
 
 Paths can traverse to an arbitrary depth in a series of nested links.
+The ``select`` below ends up showing the author of the BlogPost
+(The friend of the friend of the BlogPost, which is the author)
 
 .. code-block:: edgeql
 
-  select BlogPost.author.friends.friends;
+  select BlogPost.author; # The author
+  select BlogPost.author.friends.friends; # The author again
 
 Paths can terminate with a property reference.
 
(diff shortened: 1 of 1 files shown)


raddevon:
Working on reviewing this now, @Dhghomon, and I'm having trouble getting it to build. Does it build for you?

Dhghomon:
> Working on reviewing this now, @Dhghomon, and I'm having trouble getting it to build. Does it build for you?

@raddevon 

This is the main error I noticed when I made the PR https://github.com/edgedb/edgedb/actions/runs/4998969112/jobs/8954921525?pr=5495#step:17:705 but thought I'd first see what you thought about the code samples and their formatting before getting to the tests. What do you see on your end?

raddevon:
That's the first one, but when I resolve it, I get to another one. Haven't gone past that. Figured I'd check in with you first. Generally, on the code block formatting, I think we first need to make sure we always have an empty line between the `code-block` directive and the first line of code. I'd like to keep with our convention of 4 spaces of indentation for the actual code inside the code block. (Some existing blocks use only 2 now, but we're trying to standardize around 4 going forward.)

A lot of these seem to be REPL input/output, so the code blocks should be `edgeql-repl` instead of vanilla `edgeql`. Last, subsequent lines of input in our REPL blocks lead with `... ` so that the input is aligned from the first line to subsequent lines, and so that it's easy to distinguish input from output. Here's an example:

```
.. code-block:: edgeql-repl

    db> with data := <json>$data
    ... insert Movie {
    ...   title := <str>data['title'],
    ...   release_year := <int64>data['release_year'],
    ... };
    Parameter <json>$data: {"title": "The Marvels", "release_year": 2023}
    {default::Movie {id: 8d286cfe-3c0a-11ec-aa68-3f3076ebd97f}}
```

Those are the technical bits. Once it's building, I want to take a look at the content so I can evaluate it in context.

Thank you for putting this together!

Dhghomon:
> That's the first one, but when I resolve it, I get to another one. Haven't gone past that. Figured I'd check in with you first. Generally, on the code block formatting, I think we first need to make sure we always have an empty line between the `code-block` directive and the first line of code. I'd like to keep with our convention of 4 spaces of indentation for the actual code inside the code block. (Some existing blocks use only 2 now, but we're trying to standardize around 4 going forward.)
> 
> A lot of these seem to be REPL input/output, so the code blocks should be `edgeql-repl` instead of vanilla `edgeql`. Last, subsequent lines of input in our REPL blocks lead with `... ` so that the input is aligned from the first line to subsequent lines, and so that it's easy to distinguish input from output. Here's an example:
> 
> ```
> .. code-block:: edgeql-repl
> 
>     db> with data := <json>$data
>     ... insert Movie {
>     ...   title := <str>data['title'],
>     ...   release_year := <int64>data['release_year'],
>     ... };
>     Parameter <json>$data: {"title": "The Marvels", "release_year": 2023}
>     {default::Movie {id: 8d286cfe-3c0a-11ec-aa68-3f3076ebd97f}}
> ```
> 
> Those are the technical bits. Once it's building, I want to take a look at the content so I can evaluate it in context.
> 
> Thank you for putting this together!

Got all the tests to pass! 🎉

docs/edgeql/paths.rst
```diff
@@ -78,58 +137,75 @@ Starting from each user, the path below traverses all *incoming* links labeled
...
+
+.. code-block:: edgeql
+
+    select User.<author { text };
 
 As written, EdgeDB infers the *type* of this expression to be
-:eql:type:`BaseObject`, not ``BlogPost``. Why? Because in theory, there may be
-several links named ``author`` that point to ``User``.
+:eql:type:`BaseObject`. Why? Because in theory, there may be
+several links named ``author`` that point to ``User``. And there
```
raddevon, line 151: "...several links **from different object types** that point to ``User``, all named ``author``..."

to emphasize the reason not all links would have the `text` property

raddevon (changes requested):
This is *really* good stuff! I suggested some minor changes.

I'd also like to eventually include some natural language readings of backlinks. Doesn't have to happen in this PR, but if you want to try, feel free. I find these help people to understand backlink syntax a little better. Here's what I mean.

> The backlink `User.<author` can be read as "objects that link to the user via a link named `author`."

> The backlink `User.<author[is BlogPost]` can be read as "`BlogPost` objects that link to the user via a link named `author`."

Doesn't have to be those verbatim. There may be a natural way to put them in here, or maybe there's just a sub-section "Understanding backlink syntax" or "Reading backlink syntax" or something like that. Does that seem helpful to you, @Dhghomon?

Dhghomon:
> This is _really_ good stuff! I suggested some minor changes.
> 
> I'd also like to eventually include some natural language readings of backlinks. Doesn't have to happen in this PR, but if you want to try, feel free. I find these help people to understand backlink syntax a little better. Here's what I mean.
> 
> > The backlink `User.<author` can be read as "objects that link to the user via a link named `author`."
> 
> > The backlink `User.<author[is BlogPost]` can be read as "`BlogPost` objects that link to the user via a link named `author`."
> 
> Doesn't have to be those verbatim. There may be a natural way to put them in here, or maybe there's just a sub-section "Understanding backlink syntax" or "Reading backlink syntax" or something like that. Does that seem helpful to you, @Dhghomon?

A had a go at the human readable part and I think it can just be added as comments in the code in addition to one more section that breaks it down bit by bit. (That final part I really like, would have been great when I first saw backlinks) I went with ```edgeql formatting for that one to show the spacing, not sure if there is a better annotation though looks good on my preview.

https://github.com/edgedb/edgedb/commit/02c618d35cfeb2dacf5f8edba45aff8e2b3f1094

docs/edgeql/paths.rst
```diff
@@ -36,33 +41,87 @@ Consider the following schema:
...
 
 Paths can traverse to an arbitrary depth in a series of nested links.
+The ``select`` below ends up showing the author of the BlogPost
+(The friend of the friend of the BlogPost, which is the author)
 
 .. code-block:: edgeql
 
-  select BlogPost.author.friends.friends;
+    select BlogPost.author; # The author
+    select BlogPost.author.friends.friends; # The author again
```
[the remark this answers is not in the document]
Dhghomon: I like the wording you suggested here because it describes accurately what it does but is also enough of a tongue twister that the reader will know that this a demonstration and certainly not a recommended practice or anything.

https://github.com/edgedb/edgedb/commit/7f10efeb9fab8e6047073e2756c7ff296f58cb4c

docs/edgeql/paths.rst
Dhghomon: Okay, took your wording in mostly verbatim: https://github.com/edgedb/edgedb/commit/8512b52ec6882d7fa67d73e99d0afd63e27f3ee6

docs/edgeql/paths.rst
```diff
@@ -36,33 +41,88 @@ Consider the following schema:
...
 
 .. code-block:: edgeql
 
-  select User.friends;
+    select User.friends;
 
 Paths can traverse to an arbitrary depth in a series of nested links.
+The ``select`` below ends up showing the author of the BlogPost
+(The friends of the friends of the author of the BlogPost, 
+which in this case is just the author)
```
raddevon, line 112: Reworked slightly so it's clear which query we're talking about.
```suggestion
Both ``select`` queries below ends up showing the author of the ``BlogPost``.
The second query returns the friends of the friends of the author of the
 ``BlogPost``, which in this case is just the author.
```

docs/edgeql/paths.rst
```diff
@@ -78,58 +138,96 @@ Starting from each user, the path below traverses all *incoming* links labeled
...
+
+    select User.<author { text };
 
 As written, EdgeDB infers the *type* of this expression to be
-:eql:type:`BaseObject`, not ``BlogPost``. Why? Because in theory, there may be
-several links named ``author`` that point to ``User``.
+:eql:type:`BaseObject`. Why? Because in theory, there may be
+several links named ``author`` from different object types
+that point to ``User``. And there is no guarantee that each 
+of these links will have a property called ``text``.
```
raddevon, line 154:
```suggestion
of these types will have a property called ``text``.
```

docs/edgeql/paths.rst
```diff
@@ -78,58 +138,96 @@ Starting from each user, the path below traverses all *incoming* links labeled
...
+    # User.<        Objects that link to the user
+    # author        via a link named author
+
+    # User.<        Objects that link to the user
+    # author        via a link named author
+    # [is BlogPost] that are BlogPosts
+
+    # User.<        Objects that link to the user
+    # author        via a link named author
+    # [is Comment]  that are Comments
```
raddevon, line 188: I was concerned about the edgeql code block because that gives the user a "copy" button. I don't want to imply here that there is something a user could or should copy and paste somewhere. Almost just swapped it to a bare code block which doesn't get the copy button. Then I realized we're really just making a table here. Even though we could lay it out with a code block, it becomes less accessible than if we just make a proper table. Does that approach make sense?

```suggestion
================================ ===================================
Syntax                           Meaning
================================ ===================================
``User.<``                       Objects that link to the user
``author``                       via a link named author
================================ ===================================

================================ ===================================
Syntax                           Meaning
================================ ===================================
``User.<``                       Objects that link to the user
``author``                       via a link named author
``[is BlogPost]``                that are ``BlogPost`` objects
================================ ===================================

================================ ===================================
Syntax                           Meaning
================================ ===================================
``User.<``                       Objects that link to the user
``author``                       via a link named author
``[is Comment]``                 that are ``Comment`` objects
================================ ===================================
```

raddevon (approved):
This looks good! I made a couple of minor suggestions and one bigger one: suggesting we swap the code block with the plain-language explanations of backlink syntax for a series of tables with the same content. (I haven't looked at these rendered yet, so if you accept the suggestions, you may want to pull them and make sure they render properly before merging.)

Thank you for this. It's a big improvement!

docs/edgeql/paths.rst
Dhghomon: Yes! Looks good that way.

raddevon:
🤦 
Sorry about that, @Dhghomon. Thanks for patching it up. Great work on this!

PR #5505, opened 2023-05-18
Work around mysterious Digital Ocean crashes

msullivan:
When bootstrapping against managed Digital Ocean databases, the
postgres occasionally has a server process crash. This only seems to
happen if we keep our main connection open while initializing the
template DB (!).

So don't do that, I guess.

With this change I've managed 100+ bootstraps in a row without a
failure, so I am pretty confident this is a fix, even though I don't
have a satisfactory explanation of it.

### DIFF
--- a/edb/server/bootstrap.py
+++ b/edb/server/bootstrap.py
@@ -20,6 +20,7 @@
 from __future__ import annotations
 from typing import *
 
+import dataclasses
 import enum
 import json
 import logging
@@ -28,6 +29,7 @@ import pathlib
 import pickle
 import re
 import struct
+import sys
 import textwrap
 
 from edb import buildmeta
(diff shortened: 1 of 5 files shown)


edb/server/bootstrap.py
```diff
@@ -1828,27 +1843,36 @@ async def _bootstrap(ctx: BootstrapContext) -> None:
...
     else:
         new_template_db_id = uuidgen.uuid1mc()
 
-    try:
-        if backend_params.has_create_database:
-            tpl_ctx = ctx._replace(conn=conn)
-            conn.add_log_listener(_pg_log_listener)
-        else:
-            tpl_ctx = ctx
+    if backend_params.has_create_database:
```
fantix, line 1862: (trivial) this `if ... else ...` arms can be merged into the previous `if ... else ...` with identical condition now

fantix (approved):
Wow you found a solution! Looks good to me :+1:

msullivan (commented):
Yeah, this is better

PR #5510, opened 2023-05-18
Document `delegated` constraints in Schema,Ref>DDL

raddevon:


docs/reference/ddl/constraints.rst
```diff
@@ -370,10 +370,16 @@ Parameters
...
+    not be enforced on the type it's declared on, and the enforcement will be
+    delegated to the subtypes of this type. This is particularly useful for
+    :eql:constraint:`exclusive` constraints in abstract types. This is only
+    valid for *concrete constraints*.
 
 :eql:synopsis:`set not delegated`
-    Makes the constraint non-delegated.
+    If set, the constraint is defined as *not delegated*, which means that it
+    will be enforced globally across the type it's declared on and any
+    extending types type.
```
Dhghomon, line 382: Typo here? Extending types type -> Extending types.

docs/datamodel/constraints.rst
```diff
@@ -318,6 +318,97 @@ using arbitrary EdgeQL expressions. The example below uses the built-in
     );
   }
 
+
+Constraints and object inheritence
```
Dhghomon, line 322: Maybe Constraints and type inheritance? I see that term more and object inheritance feels more like a programming language as opposed to query language thing.

docs/datamodel/constraints.rst
```diff
@@ -318,6 +318,97 @@ using arbitrary EdgeQL expressions. The example below uses the built-in
...
+----------------------------------
+
+If you define a constraint on a parent type, the constraint will apply to all
```
Dhghomon, line 325: I keep reading apply to all as "apply to each one" which feels like delegated. (Like how "I will give an apple to all the children" doesn't feel like they are sharing a single apple) Maybe "the constraint will apply irrespective of the types that extend it" or "A constraint on a parent type is evaluated once globally regardless of how many types extend it"?

Dhghomon:
Looks good, just three small comments on wording!

Dhghomon (approved):
Okay, looks good now!

elprans (approved):

PR #5512, opened 2023-05-19
Adjust error messages for GraphQL.

### LINKED ISSUE #3816: Constraint violation error messages should be exposed in GraphQL
Currently, the GraphQL extension does not give very friendly errors when you try to do something that would violate a constraint.  For example, given this minimal schema:

```
using extension graphql;
module default {
	type ConstraintDemo {;
		property tenorless -> int16 {
			constraint max_value(10);
		}
	}
}
```

Trying to insert a ConstraintDemo with a tenorless of 11 via the CLI gives a nice error message:

```
edgedb> insert ConstraintDemo {tenorless := 11};
edgedb error: ConstraintViolationError: Maximum allowed value for tenorless is 10.
  Detail: Maximum allowed value for value of property 'tenorless' of object type 'default::ConstraintDemo' is 10.
```

But trying to do so through the GraphQL API does not:
```graphql
mutation {
  insert_ConstraintDemo(data: {tenorless: 11}) {
    id
  }
}
# returns
{
  "errors": [
    {
      "message": "QueryError: new row for relation \"81c174b3-ca67-11ec-90a9-7114d4b7def4\" violates check constraint \"a56dc4cd-ca67-11ec-a985-2df63feca51f;schemaconstr\""
    }
  ]
}
```

vpetrovykh:
In GraphQL mode the error messages should refer to various types by their exposed names (as opposed to SQL names or EdgeDB names) wherever possible. Custom scalars are an exception to this rule since they get exposed as their underlying base types instead.

Fixes #3816

### DIFF
--- a/edb/graphql/extension.pyx
+++ b/edb/graphql/extension.pyx
@@ -41,6 +41,9 @@ from edb.server import compiler
 from edb.server import defines as edbdef
 from edb.server.pgcon import errors as pgerrors
 from edb.server.protocol import execute
+from edb.server.compiler import errormech
+
+from edb.schema import schema as s_schema
 
 from edb.common import debug
 from edb.common import markup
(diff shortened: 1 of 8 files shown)


msullivan (approved):
Good DX improvement on our graphql side!

PR #5515, opened 2023-05-19
Fix create database conflicts with name in schema

### LINKED ISSUE #4566: Can't create a database that shares a name with a standard library function
```
edgedb> create database range;
edgedb error: SchemaError: function 'std::range(lower: OPTIONAL std::anypoint={}, upper: OPTIONAL std::anypoint={}, NAMED ONLY inc_lower: std::bool=true, NAMED ONLY inc_upper: std::bool=false, NAMED ONLY empty: std::bool=false)' already exists
```

aljazerzen:
Closes #4566

### DIFF
--- a/tests/test_database.py
+++ b/tests/test_database.py
@@ -83,6 +83,13 @@ class TestDatabase(tb.ConnectedTestCase):
         # create database name that conflicts with names in schema
         await self.con.execute('CREATE DATABASE range;')
 
+        conn = await self.connect(database='range')
+
+        res = await conn.query('select range(5, 10)')
+        self.assertEqual(res, [edgedb.Range(5, 10)])
+
+        await conn.aclose()
+
         await tb.drop_db(self.con, 'range')
 
     async def test_database_drop_01(self):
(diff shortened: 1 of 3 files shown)


aljazerzen:
Why does "CreateDatabase" even go through DDL? Does it needs to compute a delta? To be applied to the global schema?

It seems to me that all `DatabaseCommand`s should be simple operations - at least delta-wise.

msullivan:
> Why does "CreateDatabase" even go through DDL? Does it needs to compute a delta? To be applied to the global schema?

I believe that's right, yeah.

tests/test_database.py
```diff
@@ -76,6 +76,15 @@ async def test_database_create_03(self):
         finally:
             await tb.drop_db(self.con, 'databasename')
 
+    async def test_database_create_04(self):
+        if not self.has_create_database:
+            self.skipTest("create database is not supported by the backend")
+
+        # create database name that conflicts with names in schema
+        await self.con.execute('CREATE DATABASE range;')
```
msullivan, line 84: Could we do some testing that having this name collision doesn't lead to any trouble?

aljazerzen: Like what? Write to the new database?

msullivan: No, like using the objects that it collided with

PR #5525, opened 2023-05-22
Fix error messages for invalid parameters

efaust:


efaust:
Looking for insight from @msullivan on the approach.

edb/edgeql/compiler/astutils.py
```diff
@@ -150,16 +151,28 @@ def _visit_with_stmt(self, n: qlast.Statement) -> None:
...
         if isinstance(n.expr, qlast.Parameter):
             self.params.append((n, self.modaliases))
-        self.generic_visit(n)
+        else:
+            self.generic_visit(n)
+
+    def visit_Parameter(self, n: qlast.Parameter) -> None:
+        self.loose_params.append(n)
+
+    def visit_CreateFunction(self, n: qlast.CreateFunction) -> None:
```
msullivan, line 160: This is to prevent picking stuff up inside function bodies?

efaust: Yes. The two checks now in `throw_on_loose_param` happen in a SECOND pass. When we get a command, we'll compile the outer shell (which was looking into contraint and function bodies in this analysis pass) and then compile the contraint exprs and function bodies as part of execution. This skips them on the outer passes, and just waits until we execute to throw.

msullivan (approved):
Looks good!

PR #5526, opened 2023-05-22
Add more detailed Cloud usage instructions

raddevon:
Is this good for user-facing instructions for Cloud? I also suspect this (or something very similar) needs to be somewhere else in our documentation to make it more visible. Do you think it makes sense to have a new top-level section for Cloud? I think, with this addition, we probably have everything in the regular docs that we need (aside from maybe something about the Cloud UI), but it might be nice to surface everything that's Cloud-related somewhere else in addition to inside the CLI docs.

### DIFF
--- a/docs/cli/edgedb_cloud/index.rst
+++ b/docs/cli/edgedb_cloud/index.rst
@@ -70,6 +70,10 @@ these commands:
       $ edgedb project init \
         --server-instance <github-username>/<instance-name>
 
+  Alternatively, you can run ``edgedb project init`` *without* the
+  ``--server-instance`` option and enter an instance name in the
+  ``<github-username>/<instance-name>`` format when prompted interactively.
+
 
 3. Configure your application
 -----------------------------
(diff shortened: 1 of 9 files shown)


docs/cli/edgedb_cloud/index.rst
```diff
@@ -28,15 +47,36 @@ log in to EdgeDB Cloud.
...
-    * - :ref:`ref_cli_edgedb_cloud_logout`
-      - Forget the stored access token
-    * - :ref:`ref_cli_edgedb_cloud_secretkey`
-      - Manage your secret keys
+  .. code-block:: bash
+
+      edgedb instance create <github-username>/<instance-name>
+
+* :ref:`ref_cli_edgedb_project_init` with the ``--server-instance`` option. Set
+  the server instance name to ``<github-username>/<instance-name>``.
```
elprans, line 66: Or, you can specify `<github-username>/<instance-name>` when `project init` prompts you interactively.

elprans (approved):

PR #5531, opened 2023-05-22
Make staeval more resilient to ops not being supported on a type

msullivan:
staeval supports comparison operators (in order to check constraints
on config values in certain cases) and it supports durations (also for
config reasons). But our duration objects don't support comparison,
so doing a comparison of duration literals crashes in staeval.

This only really happens in the http/notebook protocols, because in
the binary protocol constant extraction breaks any hope of statically
evaluating things.

### DIFF
--- a/edb/ir/staeval.py
+++ b/edb/ir/staeval.py
@@ -188,7 +188,13 @@ def evaluate_OperatorCall(
 
         args.append(arg_val)
 
-    value = eval_func(*args)
+    try:
+        value = eval_func(*args)
+    except TypeError:
+        raise UnsupportedExpressionError(
+            f'operator {opcall.func_shortname} not actually supported on '
+            f'input types',
+            context=opcall.context)
     qlconst: qlast.BaseConstant
     if isinstance(value, str):
         qlconst = qlast.StringConstant.from_python(value)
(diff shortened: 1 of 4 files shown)


edb/ir/staeval.py
```diff
@@ -176,7 +176,13 @@ def evaluate_OperatorCall(
 
         args.append(arg_val)
 
-    value = eval_func(*args)
+    try:
+        value = eval_func(*args)
+    except TypeError:
+        raise UnsupportedExpressionError(
```
elprans, line 182: I think the rule should be that `statypes` *must* implement ordering (all types in EdgeQL are orderable), and then this should be an ISE.

msullivan: We can definitely support ordering for everything, but I don't think it is going to necessarily be the case that every operator that we support for some type in staeval will necessarily be supported for all of them.

elprans: I've been reluctant to add more operators to staeval given the subtle differences in results between Python and Postgres, especially when it comes to math.  

`TypeError` is not very specific so the approach in this patch risks masking implementation bugs unrelated to actual support or non-support.

msullivan: (I did just implement support though)

elprans (approved):

PR #5533, opened 2023-05-23
Type unions should not appear when computing the source of a back link

### LINKED ISSUE #5500: ISE: Relation edgedbpub.<uuid> does not exist during introspection
On latest master branch, executing 

```
select schema::InheritingObject {
  children := .<bases[IS schema::Type],
  descendants := .<ancestors[IS schema::Type]
};
```

will trigger the above ISE.

This is a separate issue from #5358.

zchenb:
Fix #5500

### DIFF
--- a/edb/schema/objtypes.py
+++ b/edb/schema/objtypes.py
@@ -192,6 +192,7 @@ class ObjectType(
             if (
                 lnk.get_shortname(schema).name == name
                 and not lnk.get_source_type(schema).is_view(schema)
+                and not lnk.get_source_type(schema).is_union_type(schema)
                 # Only grab the "base" pointers
                 and all(
                     b.generic(schema)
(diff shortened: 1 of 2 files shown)


edb/schema/objtypes.py
```diff
@@ -192,6 +192,7 @@ def getrptrs(
             if (
                 lnk.get_shortname(schema).name == name
                 and not lnk.get_source_type(schema).is_view(schema)
+                and not lnk.get_source_type(schema).is_union_type(schema)
```
msullivan, line 195: I think this needs to be duplicated in the condition in the loop over ancestors below, also.

Which, actually: we should get rid of the code duplication and just have one loop over `(lnk,) + ancestors`

msullivan (approved):
This looks good, modulo my one comment.

"Ban union in links" is a pretty inaccurate commit message, though. Try "Fix backlinks in the presence of union pointers" or something along those lines?

zchenb:
@msullivan  Could you please take a another look at this PR? The changes are in place.

msullivan (approved):
Yeah looks good

PR #5537, opened 2023-05-23
Do not allow link property on properties

### LINKED ISSUE #5327: Link property on properties are parsed incorrectly
Version: EdgeDB 3.0-dev.7454+local (repl 3.0.0-dev)

This schema is not processed correctly.
```
type Test1 {
    title : str {
        sub_title : str
    }
}
```

The query result is 
```
edgedb> describe type Test1 as sdl;
{
  'type default::Test1 {
    property sub_title: std::str;
    property title: std::str;
};',
}
```

zchenb:
Fix #5327


msullivan (approved):
Good catch

PR #5540, opened 2023-05-24
'Fix' ISE for TypeExprs in function returns

efaust:
We are already happy to create a type shell from a TypeOp. Everything already
works...except that functions were not applying their qualfied namespace to any generated types. Fixing that makes it go.


msullivan (approved):
Looks good!

PR #5542, opened 2023-05-24
Reduce JWT validation clock skew leeway to 5 seconds

elprans:
The default 60 seconds is unreasonably long and also makes testing of
key expiration more awkward than necessary.


fantix (approved):

msullivan:
Do we still want this?

PR #5543, opened 2023-05-24
Add read-only mode

fantix:
- [x] Add test
- [x] Check other writable entrypoints

### DIFF
--- a/edb/server/protocol/binary_v0.pyx
+++ b/edb/server/protocol/binary_v0.pyx
@@ -1034,6 +1034,9 @@ cdef class EdgeConnectionBackwardsCompatible(EdgeConnection):
 
             compiled = self._last_anon_compiled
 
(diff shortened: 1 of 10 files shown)


edb/server/protocol/binary_v0.pyx
```diff
@@ -1031,6 +1034,9 @@ cdef class EdgeConnectionBackwardsCompatible(EdgeConnection):
 
             compiled = self._last_anon_compiled
 
+        if self.server.is_readonly():
+            allow_capabilities &= ~enums.Capability.WRITE
```
elprans, line 1038: We should pass a `detail` text to `make_error` below saying "the instance is in read-only mode" so users have a better explanation why the query is being rejected.

elprans (approved):

PR #5544, opened 2023-05-24
docs: Improve `to_str(datetime)` docs a bit

elprans:
* Replace `to_char` references with `to_str`
* Recommend use of double quotes for literal text in format strings
* Quote `of` in examples as "of" is a valid format specifier in
  PostgreSQL 15

### DIFF
--- a/docs/stdlib/string.rst
+++ b/docs/stdlib/string.rst
@@ -664,7 +664,7 @@ Strings
     .. code-block:: edgeql-repl
 
         db> select to_str(<datetime>'2018-05-07 15:01:22.306916-05',
-        ...               'FMDDth of FMMonth, YYYY');
+        ...               'FMDDth "of" FMMonth, YYYY');
         {'7th of May, 2018'}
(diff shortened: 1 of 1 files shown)


docs/stdlib/string.rst
```diff
@@ -664,7 +664,7 @@ Strings
     .. code-block:: edgeql-repl
 
         db> select to_str(<datetime>'2018-05-07 15:01:22.306916-05',
-        ...               'FMDDth of FMMonth, YYYY');
+        ...               'FMDDth "of" FMMonth, YYYY');
         {'7th of May, 2018'}
         db> select to_str(<cal::local_date>'2018-05-07', 'CCth "century"');
         {'21st century'}
```
raddevon, line 670: Could we add a shorter version of the literal text note here where it's being demonstrated? Something like this maybe?
```suggestion
        {'21st century'}

    .. note::
    
        If you want to use literal text in your format string, it's best to
        enclose it in double quotes as shown above with ``of`` and
        ``century``.
```

raddevon (approved):
One suggestion, but this looks great!

PR #5545, opened 2023-05-24
Clean up handling of unions in link targets

msullivan:
There is no need for any special handling, we just need to
make sure to pass down the module name.


efaust (approved):
Muuuuch simpler. This even makes sense to me, now.

msullivan:
> Muuuuch simpler. This even makes sense to me, now.

Yeah, me too :P

PR #5560, opened 2023-05-26
Disallow partial paths in the defaults of link properties

### LINKED ISSUE #5200: Error during compiled migration: QueryError: could not resolve partial path
Version: newest master branch

Steps to reproduce:
1. Replace your project's schema with the following:
2. run `edb cli migration create`, this step should succeeded. If this step fails, then this bug has been fixed.
3. run `edb cli migrate`, this should now give an error that is hard to understand. 

I believe this has to do with how link properties interact with defaults.

Schema:
```
module default {
type Person {
    required name : str {
        constraint exclusive;
    }

    multi l3 : Person {
        l3note : str {
            default := .name
        }
    }

}
}
```

zchenb:
Fix #5200


zchenb:
With #5142, we should probably provide a hint to why the partial paths cannot be resolved. This error message is obscure. Creating an issue #5561.

edb/schema/pointers.py
```diff
@@ -1539,12 +1543,30 @@ def compile_expr_field(
             else:
                 in_ddl_context_name = None
 
+            # If we are in a link property's default field
+            # do not set path prefix anchor, because link properties
+            # cannot have defaults that reference the object being inserted
+            should_set_path_prefix_anchor = True
+            if field.name == 'default':
+                from .objtypes import ObjectTypeCommandContext
```
msullivan, line 1551: This is a pretty gnarly way to check if it is a link property.
I would try something like
```
parent_ctx = self.get_referrer_context_or_die(context)
source = parent_ctx.op.get_object(schema, context)
```
and then check if source is a `Pointer`. If it is, this is a link property.
(There are similar checks in `_compile_expr`)

msullivan (approved):
These seems like a solid fix, though I posted some quibbles about the exact approach

zchenb:
I have made the updates. Could you please take another look @msullivan?

edb/schema/pointers.py
```diff
@@ -1539,12 +1543,26 @@ def compile_expr_field(
...
+            # do not set path prefix anchor, because link properties
+            # cannot have defaults that reference the object being inserted
+            should_set_path_prefix_anchor = True
+            if field.name == 'default':
+                # We are checking if the parent context is a pointer
+                # (i.e. a link or a property).
+                # If so, do not set the path prefix anchor.
+                parent_ctx = self.get_referrer_context_or_die(context)
+                source = parent_ctx.op.get_object(schema, context)
+                if (isinstance(source, Pointer)):
```
msullivan, line 1556: Drop the enclosing parens

msullivan (approved):
Looks good.

Hope the merge conflicts aren't bad...

PR #5562, opened 2023-05-26
Improve the error reporting for missing pointers in schema inserts

### LINKED ISSUE #5549: ISE when validation within a rewrite fails
- EdgeDB Version: Current master
- EdgeDB CLI Version: dev
- OS Version: Linux

Steps to Reproduce:

1. Just migrate to this schema

Schema:

```
module default {

    
    type User {
        
        required password: str;
        
        multi link projects := .<owner[is Project];

        link default_project: Project {
            rewrite insert using (
                insert Project {
                    
                    owner := __subject__,
                }
            )
        };
    }

    type Project {
        required name: str;
        required owner: User;

        constraint exclusive on ( (.name, .owner) );
    }
}
```

msullivan:
There were two problems:
1. We didn't report an error context. Surprisingly much plumbing
   needed to get it there.
2. We reported a MissingRequiredError, which is a sub-error of
   ExecutionError. The CLI didn't like this, and reported a whole
   stack trace, presumbably because it felt that we weren't really
   *executing* anything yet, which is a fair point.
   Report a QueryError when we hit this case from SDL/DDL instead.

There are two other plausible options for dealing with point 2:
 * Always report a QueryError. I didn't do this because
   MissingRequiredError /is/ more informative and a lot of tests
   use it, and in non-migration cases it can also basically come
   from execution.
 * Keep reporting MissingRequiredError and fix the CLI instead.
   I didn't do this because we /shouldn't/ be generating an
   ExecutionError here, and I think this is the only case where we
   might.

Fixes #5549.


zchenb:
I have pushed #5564 that should fix the failing CI.

msullivan:
> I have pushed #5564 that should fix the failing CI.

Thanks!

aljazerzen (approved):
I agree that ExecutionError does not make sense here. DDL/SDL is not executing a query, just defining a query to be executed, so this could even be `SchemaDefinitionError`.

msullivan:
> I agree that ExecutionError does not make sense here. DDL/SDL is not executing a query, just defining a query to be executed, so this could even be `SchemaDefinitionError`.

Good call

PR #5563, opened 2023-05-26
Add disallow partial path hint

zchenb:


edb/schema/pointers.py
```diff
@@ -1561,6 +1563,23 @@ def compile_expr_field(
                 if isinstance(source, Pointer):
                     should_set_path_prefix_anchor = False
 
+            # If we are in a link property's default field
```
msullivan, line 1566: I think that this is spurious now?

edb/testbase/server.py
```diff
@@ -231,11 +232,18 @@ def annotate(self, **kwargs):
...
 
     @contextlib.contextmanager
-    def assertRaisesRegex(self, exception, regex, msg=None, **kwargs):
+    def assertRaisesRegex(self, exception, regex, msg=None, hint=None,
+                          **kwargs):
         with super().assertRaisesRegex(exception, regex, msg=msg):
             try:
                 yield
             except BaseException as e:
+                if hint is not None:
```
msullivan, line 241: We shouldn't need to add this; it's already supported below in a generic way

msullivan (commented):
Seems good, though a few changes are needed.
I don't think it fully resolves the issue, though; we probably want to come up with a hint to display in more common cases too

edb/schema/pointers.py
zchenb: Yes

PR #5564, opened 2023-05-26
Fix schema introspection with computed backlinks test case

zchenb:
Apparently schema changes have broke this test case. The idea of the test case is not to inspect the actual schema, but to make sure backlinks wouldn't cause SQL errors.


zchenb:
The test case was originally introduced in #5533 as a regression test case.

msullivan (approved):

PR #5565, opened 2023-05-27
Fix several issues related to schema loading in declarative

### LINKED ISSUE #4683: Migrations fail when abstract properties are referenced in globals
- EdgeDB Version: 2.6
- OS Version: macOS 13.0

Steps to Reproduce:

1. Create the schema below
2. Run `edgedb migration create`

Expected behavior: Migration should run successfully.
Actual behavior: This produces the error `error: object type 'default::Role' has no link or property 'slug'`

Schema:

```
module default {
  global current_user_id -> uuid;
  global current_user := (
    select User filter .id = global current_user_id
  );
  global current_user_role := (
    (global current_user).role.slug
  );

  abstract type WithSlug {
    property slug -> str
  }

  type Role extending WithSlug {
    property description -> str
  }

  type User {
    required link role -> Role;
  }
}
```

### LINKED ISSUE #5163: assert_exists makes the compiler claim "object type ... has no link or property ..."
- EdgeDB Version: "2.12+815c3e2"
- EdgeDB CLI Version: EdgeDB CLI 2.3.1+5d93f42
- OS Version: MacOS 13.0

Steps to Reproduce:
`edgedb migration create`

```
error: object type 'default::DataEntry' has no link or property 'branch'
  ┌─ dbschema/common.esdl:7:45
  │
7 │         link branch := .datapoint.data_entry.branch;
  │                                             ^^^^^^^ error

```

Schema:

```
module default {
    type Branch{
    }
    type CO2DataPoint{
        required link datapoint -> DataPoint;
        link branch := .datapoint.data_entry.branch; # THIS LINE
    }
    type DataPoint{
        required link data_entry := assert_exists(.<data_points[is DataEntry]); # THIS GIVES ERROR BUT ONLY IF THE LINE ABOVE IS UNCOMMENTED
        # link data_entry := .<data_points[is DataEntry]; # THIS WORKS
    }

    type DataEntry{
        required link branch -> Branch;
        multi link data_points -> DataPoint;
    }
}
```

msullivan:
The main change is a new schema where if we can't figure out the type
of an expression in a path, create weak dependencies to all pointers
with the same name.

This doesn't always work but in practice it will do very well.

I have some ideas for how we could incorporate correct typechecking
into schema loading, but in the mean time this probably solves 90% of
the problem at least.

In this PR there is also a commit that adds a system for categorizing
objects so that things without expressions will go first, and things
with expressions will be prioritized based on whether they can be
referred to.
I wound up reverting it because it isn't needed to fix the issues that
weak pointers fix, and other than those, I think it mostly would work
around bugs that could be fixed in straightforward ways.

Fixes #4683. Fixes #5163.

### DIFF
--- a/edb/edgeql/tracer.py
+++ b/edb/edgeql/tracer.py
@@ -319,16 +319,21 @@ def trace_refs(
     path_prefix: Optional[sn.QualName] = None,
     module: str,
     objects: Dict[sn.QualName, Optional[ObjectLike]],
+    pointers: Mapping[sn.UnqualName, Set[sn.QualName]],
     params: Mapping[str, sn.QualName],
     local_modules: AbstractSet[str]
-) -> FrozenSet[sn.QualName]:
+) -> Tuple[FrozenSet[sn.QualName], FrozenSet[sn.QualName]]:
 
-    """Return a list of schema item names used in an expression."""
+    """Return a list of schema item names used in an expression.
+
+    First set is strong deps, second is weak.
+    """
 
     ctx = TracerContext(
         schema=schema,
         module=module,
         objects=objects,
+        pointers=pointers,
         anchors=anchors or {},
         path_prefix=path_prefix,
         modaliases={},
@@ -337,7 +342,7 @@ def trace_refs(
         local_modules=local_modules,
     )
     trace(qltree, ctx=ctx)
-    return frozenset(ctx.refs)
+    return frozenset(ctx.refs), frozenset(ctx.weak_refs)
 
 
 def resolve_name(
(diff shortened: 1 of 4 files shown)


aljazerzen (commented):
I've a few conceptual questions, but let's talk about that on the call today.

edb/edgeql/tracer.py
```diff
@@ -461,6 +469,8 @@ def _fork_context(ctx: TracerContext) -> TracerContext:
         schema=ctx.schema,
         module=ctx.module,
         objects=dict(ctx.objects),
+        # XXX?
```
elprans, line 472: Can you elaborate the XXX or remove it?

elprans (approved):

PR #5576, opened 2023-05-31
Add youtube directive

jaclarke:
Adds a `edb:youtube-embed` directive which takes the video id as an argument, and optionally any content will be rendered as a caption, eg: 

```rst
.. edb:youtube-embed:: OZ_UURzDkow

    Watch our video tour of this example project to get a preview of what you'll be
    building in this guide
```
image

~~Also the `react-element`, `section-intro-page` and `gh` directives/role were part of the `eql` domain, but aren't really related to edgeql documentation, so they have been moved to a new more general `edb` domain along with the `youtube-embed` directive.~~


raddevon (approved):
This is great!

1st1:
I think we can just keep `eql` as is -- it's a bad naming for sure, but having both `eql` and `edb` can also be confusing.

PR #5586, opened 2023-06-01
Fix nightly build by removing typing_inspect dependency

msullivan:
Nightlies are failing because:
 * gen_rust_ast.py uses typing_inspect, which is a test dependency
 * The nightly tests run `edb.tool`, which imports all the tools
 * Test dependencies aren't included in the built artifacts
   (tests that use test dependencies are all written to be skipped
   if the dep is missing)

There are a bunch of ways to fix this but the one that I apparently
chose was to make gen_rust_ast instead use the custom simplified
version of typing_inspect that we have for some reason. (Maybe just to
avoid the dependency.) This required fixing it to understand unions
created with `|`.

### DIFF
--- a/edb/common/_typing_inspect.py
+++ b/edb/common/_typing_inspect.py
@@ -29,7 +29,7 @@ relies on that only works on Python 3.9+.
 from __future__ import annotations
 
 import collections
-from types import GenericAlias  # type: ignore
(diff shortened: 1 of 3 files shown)


elprans (approved):

aljazerzen (commented):
We have _typing_inspect.py! I should have taken a look around before adding a dependency.

PR #5594, opened 2023-06-05
Fix `pg_namespace` emulation

elprans:
Current emulation of `pg_namespace` is somewhat broken because it
returns `NULL` for `tableoid` for synthesized schemes, which throws off
catalog map logic in `pg_dump` of PostgreSQL 15.  Emit OID of
`pg_namespace` correctly instead.  While at it, set `xmin`, `xmax`,
`cmin`, `cmax` to non-`NULL` values as well.  `ctid` is harder to
emulate correctly so leave it as `NULL`.


aljazerzen (approved):
I was foolish enough to think that no software would use these.

pg_dump really is the ultimate test suite for Postgres compatibility.

elprans:
Hmm, I think this also needs to be backported as a metaschema patch.

PR #5596, opened 2023-06-05
Fix "upgrades" spelling error

BD103:
:)


raddevon (approved):
Thanks for catching this!

PR #5598, opened 2023-06-05
Refactor pg types

aljazerzen:
Add type annotations to `edb/pgsql/types.py`

Remove `_PointerStorageInfo` class that only has classmethods and pretends to be another class, in favor of `PointerStorageInfo`, which I converted from extending a NamedTuple into a dataclass.

### DIFF
--- a/edb/pgsql/types.py
+++ b/edb/pgsql/types.py
@@ -539,9 +539,6 @@ def _get_ptrref_storage_info(
         table_type = 'ObjectType'
         col_name = ptrref.shortname.name
 
-    elif isinstance(ptrref, irast.TypeIntersectionPointerRef):
-        raise NotImplementedError()
-
     elif ptrref.source_ptr is not None:
         # link property
         assert isinstance(ptrref, irast.PointerRef)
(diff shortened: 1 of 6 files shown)


edb/pgsql/types.py
```diff
@@ -440,18 +399,82 @@ def get_pointer_storage_info(
...
+
+@dataclasses.dataclass(kw_only=True, repr=False, eq=False, slots=True)
+class PointerStorageInfo:
 
     table_name: Optional[Tuple[str, str]]
     table_type: str
     column_name: str
     column_type: Tuple[str, str]
 
+    def __repr__(self) -> str:
```
msullivan, line 465: didn't like the default dataclass repr?

msullivan (approved):
Great cleanups!

edb/pgsql/types.py
aljazerzen: I've copied this from `_PointerStorageInfo`, thinking it's a part of some clever framework for `dump()`. Apparently not.

PR #5603, opened 2023-06-06
Support a primitive form of parameterized types

msullivan:
Types declared in extensions can declare a type scheme that allows
user-defined scalars that extend them to provide the argument value.
Eventually we may want to generalize this to work in arbitrary
situations, but we aren't there yet.

I may have built more generic of infrastructure than necessary: types
declare an array describing the parameters they take and a "type
scheme" that says how to format the modified type names.

But maybe that's just silly and it should just take a number of
parameters, which will need to be integers, and it will format them in
the obvious way?

### DIFF
--- a/edb/schema/scalars.py
+++ b/edb/schema/scalars.py
@@ -27,6 +27,8 @@ from edb.common import checked
 from edb.edgeql import ast as qlast
 from edb.edgeql import qltypes
 
+from edb.common.typeutils import downcast
+
 from . import abc as s_abc
 from . import annos as s_anno
 from . import casts as s_casts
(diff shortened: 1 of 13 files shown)


edb/schema/scalars.py
```diff
@@ -60,7 +62,29 @@ class ScalarType(
...
+    # If present, describes what the sql_type of children scalars
+    # should be, such as 'varchar({__arg_0__})'.
+    sql_type_scheme = so.SchemaField(
+        str, default=None, inheritable=False, compcoef=0.9)
+
+    # Param definitions for a parameterized type scheme.
+    # Array values notionally say what the type should be,
+    # but currently we only support integers.
+    params = so.SchemaField(
+        checked.FrozenCheckedList[str], default=None,
```
elprans, line 77: Why a `str` though?  Shouldn't this be an `ObjectList[Type]`?

msullivan: Two reasons:
1. Making it `str` was very easy, because all the code to support that exists already. Making it a list of `Type` would require doing some work. We could definitely do it.
2. I wasn't sure if in the future, the params *will* always be edgedb types. They might be "options" or some such?

I don't have a good sense of what all we might be putting there in the future. Looking at postgres, it *might* just end up always being integers, in which case we should just have a *count* and no param types at all.

I can make it be a list of `Type`, though, if you think that's best

msullivan: I think I am going to just make it so that we take a count of parameters.
It's not exposed to introspection, and doesn't need to be backwards compatible, so we can totally rework it in the future, so I'm just going to go with an internal API that is clean and obvious for its current uses.

elprans: Ok if it's internal

elprans (approved):

PR #5604, opened 2023-06-06
Inheritance clarification

Dhghomon:
At the moment almost every example of `extending` in our documentation happens on abstract types, giving the impression that only abstract types can be extended. This PR adds an explanation and example of a concrete type being extended to show that concrete types can be extended in exactly the same way.

Also includes a fix on a type `Addressable` elsewhere in the documentation that clearly was not meant to be a concrete type.


docs/datamodel/inheritance.rst
```diff
@@ -47,6 +47,42 @@ Object types can *extend* other object types. The extending type (AKA the
       breed: str;
     }
 
+Both abstract and concrete object types can be extended. Whether to make a
+type abstract or concrete is a fairly simple decision: concrete types can be
+inserted on their own, while abstract types cannot. In the schema below the
```
raddevon, line 52: This is subtle and probably not super important but…

You lead in by saying the choice "is a fairly simple decision," but then what follows is a description of the *difference* in the two, not the description of the decision. I would tweak this in one of two ways.

> Both abstract and concrete object types can be extended. The difference in an abstract type versus a concrete on is fairly simple: concrete types can be inserted on their own, while abstract types cannot. In the schema below…

leaving them to figure out how to make the decision, _or_ outline the decision itself

> Both abstract and concrete object types can be extended. Whether to make a type abstract or concrete is a fairly simple decision: if you need to be able to insert objects of the type, make it a concrete type. If objects of the type should never be inserted and it exists only to be extended, make it an abstract one. In the schema below…

raddevon (approved):
Suggested one tweak in a comment, but this looks great otherwise!

docs/datamodel/inheritance.rst
Dhghomon: Nice suggestions. I choose the latter!

PR #5606, opened 2023-06-06
Fix during migration defaults are not compiled as detached

### LINKED ISSUE #5547: ISE when inserting defaults and refering to parent object type
Migration to this schema fails:

```
module default {
    type User {
        link default_project: Project {
            default := (insert Project {
                owner := User
            })
        };
    }

    type Project {
        required owner: User;
    }
}
```

- EdgeDB Version: 3.0-rc1
- EdgeDB CLI Version: EdgeDB CLI 3.1.0+69bd230
- OS Version: Linux Ubuntu something

aljazerzen:
Closes #5547

### DIFF
--- a/edb/schema/pointers.py
+++ b/edb/schema/pointers.py
@@ -1439,6 +1439,7 @@ class PointerCommandOrFragment(
         no_query_rewrites: bool = False,
         make_globals_empty: bool = False,
         source_context: Optional[parsing.ParserContext] = None,
+        detached: bool = False,
     ) -> s_expr.CompiledExpression:
         singletons: List[Union[s_types.Type, Pointer]] = []
 
@@ -1499,7 +1500,9 @@ class PointerCommandOrFragment(
                 in_ddl_context_name=in_ddl_context_name,
             )
 
-            compiled = expr.compiled(schema=schema, options=options)
+            compiled = expr.compiled(
+                schema=schema, options=options, detached=detached
+            )
 
             if singleton_result_expected and compiled.cardinality.is_multi():
                 if expr_description is None:
@@ -1536,8 +1539,10 @@ class PointerCommandOrFragment(
                 parent_vname = source.get_verbosename(schema)
                 ptr_name = self.get_verbosename(parent=parent_vname)
                 in_ddl_context_name = f'computed {ptr_name}'
+                detached = False
             else:
                 in_ddl_context_name = None
+                detached = True
 
             return self._compile_expr(
                 schema,
@@ -1545,6 +1550,7 @@ class PointerCommandOrFragment(
                 value,
                 in_ddl_context_name=in_ddl_context_name,
                 track_schema_ref_exprs=track_schema_ref_exprs,
+                detached=detached,
             )
         else:
             return super().compile_expr_field(
@@ -1727,7 +1733,7 @@ class PointerCommand(
 
             if not default_expr.irast:
                 default_expr = self._compile_expr(
-                    schema, context, default_expr
+                    schema, context, default_expr, detached=True,
                 )
                 assert default_expr.irast
 
@@ -1838,7 +1844,7 @@ class PointerCommand(
             'std::uuid_generate_v4',
         )
 
-        if (
+        while (
             isinstance(expr, irast.Set)
             and expr.expr
             and irutils.is_trivial_select(expr.expr)
--- a/tests/test_edgeql_ddl.py
+++ b/tests/test_edgeql_ddl.py
@@ -608,7 +608,7 @@ class TestEdgeQLDDL(tb.DDLTestCase):
             CREATE TYPE TestSelfLink3 {
                 CREATE PROPERTY foo3 -> std::str;
                 CREATE PROPERTY bar3 -> std::str {
-                    SET default := TestSelfLink3.foo3;
+                    SET default := .foo3;
                 };
             };
         """)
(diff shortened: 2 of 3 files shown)


tests/test_edgeql_ddl.py
```diff
@@ -608,7 +608,7 @@ async def test_edgeql_ddl_16(self):
             CREATE TYPE TestSelfLink3 {
                 CREATE PROPERTY foo3 -> std::str;
                 CREATE PROPERTY bar3 -> std::str {
-                    SET default := TestSelfLink3.foo3;
+                    SET default := .foo3;
                 };
             };
```
aljazerzen, line 613: The fix broke this test.

Turns out if you tried to do an INSERT after this test, it would fail. So it was wrong before.

edb/schema/pointers.py
```diff
@@ -1838,7 +1844,7 @@ def _check_id_default(
             'std::uuid_generate_v4',
         )
 
-        if (
+        while (
             isinstance(expr, irast.Set)
             and expr.expr
             and irutils.is_trivial_select(expr.expr)
```
aljazerzen, line 1850: Compiling as detached produced double nested `irast.Set`

msullivan (commented):
Could you double check how other things that are detached handle this (policies, rewrites), since I don't remember, and we might want to unify the mechanisms?

We also need to have a more general discussion of what the plan is for schema things being detached, since the backstory here is kind of twisted and we're in a pretty mixed state right now

aljazerzen:
I've done an analysis of supported features and this is the resulting matrix:


|                  | .property                      | User.property | dunder                 
|------------------|--------------------------------|---------------|------------------------
|defaults          | yes (correlated)               | detached      | no                     
|rewrites          | yes (correlated, __subject__)  | detached      | __subject__, __new__   
|triggers for each | no                             | correlated    | __old__, __new__       
|triggers for all  | no                             | detached      | __old__, __new__       
|policies          | yes (correlated, __subject__)  | detached      | __subject__            

All in all, pretty consistent. These are the things we could improve:
- add `__subject__` for defaults,
- make `User.property` in `triggers for each` detached,
- allow `.property` in triggers.


PS: these are all tested for properties only, no links or link properties.


<details>
<summary>Schema used</summary>

```
module default {

    type User1 {

        required property nick: str;
        
        required property name: str {

            default := (select .nick);

            rewrite insert using (select .nick);

        };

        # trigger name_insert after insert for each do (
        #     assert_single((select .nick))
        # );

        # trigger name_insert_all after insert for all do (
        #     assert_single((select .nick))
        # );

        access policy hide_unnamed
            allow select
            using (select .nick != "");
        access policy write
            allow insert, update, delete;
    }

    type User2 {

        required property nick: str;

        required property name: str {

            default := (select User2.nick);

            # error: missing value for required property 'name'
            # -> non-correlated
            # rewrite insert using (select User2.nick);

        };

        trigger name_insert after insert for each do (
            assert_single((select User2.nick))
        );

        trigger name_insert_all after insert for all do (
            assert_single((select User2.nick))
        );

        # Error: possibly empty set -> non-correlated
        # access policy hide_unnamed
        #     allow select
        #     using (User2.nick != "");
    }

    type User3 {

        required property nick: str;

        required property name: str {

            # error: KeyError __subject__
            # default := (select __subject__.nick);

            rewrite insert using (select __subject__.nick);

        };

        trigger name_insert after insert for each do (
            assert_single((select __new__.nick))
        );
        trigger name_insert_all after insert for all do (
            assert_single((select __new__.nick))
        );


        access policy hide_unnamed
            allow select
            using (__subject__.nick != "");
        access policy write
            allow insert, update, delete;
    }
    
}
```

</details>

msullivan:
That's a great table! Could you finish it up with indexes, constraints, and computeds?

msullivan:
I think that `User.property` in `FOR EACH` is properly detached
When I try
```
    type User2 {
        required property nick: str;
        trigger name_insert after insert for each do (
            { single x := User2.nick }
        );
    }
```
I get
```
possibly more than one element returned by an expression for a computed property 'x' declared as 'single'
```

Using your version with a runtime `assert_single` I also got an error. Maybe you only tested that one when there wasn't actually other data?

aljazerzen:
I've filled in the table for the remaining constructs.

The `User.property` in "trigger for each" is indeed detached, my mistake.

aljazerzen:
Should I open a new issue about this?

elprans:
@aljazerzen, computeds have the `__source__` dunder, referring to the source object of the computed pointer.

elprans:
(defaults too)

aljazerzen:
@elprans I'm getting these two errors:

```
type User3 {

    required property nick: str;

    required property name: str {

        # Error: __subject__ cannot be used in this expression
        default := (select __subject__.nick);
    };


    # Error: __subject__ cannot be used in this expression
    single property upper_name := str_upper(__subject__.name);
}
```
... it this only in SDL maybe? Or am I doing something wrong?

aljazerzen:
The default is still giving me errors, but only when I insert, which must be a bug:

```
edgedb> insert User3 { nick := "helllllo" };
error: InvalidReferenceError: __source__ cannot be used in this expression
  ┌─ <query>:1:1
  │
1 │ insert User3 { nick := "helllllo" };
  │ ^^^^^^^^^^ error
```

msullivan (approved):

PR #5608, opened 2023-06-06
Add math::sqrt to stdlib

efaust:
Add math::sqrt to stdlib. Along the way, clean up an ISE for errors in l{n,g,og}
which were defined nearby.


msullivan (approved):
Looks good.

Go and bump `EDGEDB_CATALOG_VERSION` in `edb/buildmeta.py` before you land.

PR #5611, opened 2023-06-07
Do not corrupt cardinality inference results when doing group analysis

msullivan:
Group aggregate use analysis makes calls to card inference... which can
side effect the IR and the schema! Add a flag to disable this.

The other option I considered was making a deep copy of the IR before
doing this analysis, which doesn't seem wonderful.


elprans (approved):
A bit messy, but deepcopies are certainly worse

PR #5612, opened 2023-06-07
Fix explicit references on indexes

msullivan:


elprans (commented):
Didn't we decide that set refs should be detached in schema fields?

edb/schema/indexes.py
```diff
@@ -244,16 +244,10 @@ def get_root(
         self,
         schema: s_schema.Schema,
     ) -> Index:
-        if not self.get_abstract(schema):
-            name = sn.shortname_from_fullname(self.get_name(schema))
-            index = schema.get(name, type=Index)
```
aljazerzen, line 249: Is this a fix for an unrelated bug? Abstract indexes not finding root correctly?

aljazerzen (approved):

PR #5613, opened 2023-06-07
Support doing patch tests against beta/rc branches

msullivan:
Tested it on 3.x and it works


elprans (approved):

msullivan:
Test failures are because I screwed up the branch naming; `patch-test-*` fires the patch tests, but I didn't have a version number in the branch

PR #5622, opened 2023-06-08
Backport pgvector to 3.x

msullivan:
This uses a squashed version of the latest `pgvector` branch that's been pushed and the pending https://github.com/edgedb/edgedb/pull/5621.

It seems to work, but I haven't really put it through its paces too heavily yet.

### DIFF
--- a/edb/lib/ext/vector.edgeql
+++ b/edb/lib/ext/vector.edgeql
@@ -1,0 +1,153 @@
+#
+# This source file is part of the EdgeDB open source project.
+#
+# Copyright 2023-present MagicStack Inc. and the EdgeDB authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+create extension package vector version '1.0' {
+    set ext_module := "vector";
+    set sql_extensions := ["vector"];
+
+    create module vector;
+
+    create scalar type vector::vector extending std::anyscalar {
+        set id := <uuid>"9565dd88-04f5-11ee-a691-0b6ebe179825";
+        set sql_type := "vector";
+        set sql_type_scheme := "vector({__arg_0__})";
+        set num_params := 1;
+    };
+
+    create cast from vector::vector to std::str {
+        set volatility := 'Immutable';
+        using sql cast;
+    };
+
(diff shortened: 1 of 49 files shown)


edb/lib/ext/vector.edgeql
```diff
@@ -0,0 +1,153 @@
...
+    create cast from vector::vector to std::json {
+        set volatility := 'Immutable';
+        using sql 'SELECT val::text::jsonb';
+    };
+
+    create cast from std::json to vector::vector {
+        set volatility := 'Immutable';
+        using sql $$
+        SELECT (
+            CASE WHEN nullif(val, 'null'::jsonb) IS NULL THEN NULL
```
1st1, line 52: This NULL seems suspicious.

edb/lib/ext/vector.edgeql
```diff
@@ -0,0 +1,153 @@
...
+        )::float[]::vector
+        $$;
+    };
+
+    create function vector::euclidean_distance(
+        a: vector::vector,
+        b: vector::vector,
+    ) -> std::float64 {
+        set volatility := 'Immutable';
+        # Needed to pick up the indexes when used in ORDER BY.
```
1st1, line 66: Not sure how exactly this is coded, @msullivan, but is there any chance for name collision with `a` / `b`? (this might be a totally wrongly formulated question)

msullivan: No, it's a function

edb/lib/ext/vector.edgeql
msullivan: This looks like it is directly lifted from the code for `json` -> `array<json>` casts.
When casting from json, `null` produces `{}` rather than an error, and we need to handle that explicitly.

edb/lib/ext/vector.edgeql
```diff
@@ -0,0 +1,153 @@
...
+        set prefer_subquery_args := true;
+        using sql 'SELECT a <=> b';
+    };
+
+    create function vector::len(a: vector::vector) -> std::int64 {
```
1st1, line 91: Let's kill this one, @msullivan @vpetrovykh, for the same reason we'll be killing concatenation functions. We'll add `std::len()` in 4.0 or later.

edb/lib/ext/vector.edgeql
```diff
@@ -0,0 +1,153 @@
...
+
+    create cast from vector::vector to std::str {
+        set volatility := 'Immutable';
+        using sql cast;
+    };
+
+    create cast from std::str to vector::vector {
+        set volatility := 'Immutable';
+        using sql cast;
+    };
```
1st1, line 41: Drop these casts please. We don't allow casting arrays to strings (or vice versa) in EdgeDB.

Adding a cast to/from `array<float32>` would be great though.

edb/lib/ext/vector.edgeql
msullivan: But I did end up redoing this cast

msullivan:
I'm closing this and will reopen it, for annoying technical reasons

PR #5623, opened 2023-06-08
Refactor tokenizer

aljazerzen:
Pushes token validation earlier into the pipeline. Removes dependency on `combine`. Unifies a few structs (`CowToken`, `SpannedToken`, `Token`).

This is all in preparation of removing roundtrip to Python for parsing.

This does change public API of `edgeql-parser` crate so CLI will also have to be updated. I have a commit lined up.


edb/edgeql-parser/src/validation.rs
```diff
@@ -0,0 +1,207 @@
...
+            peeked
+        } else {
+            self.inner.next()
+        }
+    }
+
+    /// Mimics behavior of [std::iter::Peekable]. We could use that, but it
+    /// hides access to underlying iterator.
+    fn peek(&mut self) -> &Option<Result<Token<'a>, Error>> {
+        if self.peeked.is_none() {
```
msullivan, line 65: Could this be cleaner with a `match` instead of `is_none`+`unwrap`?
Or does that run into some obscure borrow checker disaster?

msullivan:
What all changes about the public API?

msullivan (commented):
This seems basically plausible to me, but I'd like Paul's feedback too

edb/edgeql-parser/src/validation.rs
aljazerzen: You could do probably do something like:
```rust
match &self.peeked {
   Some(x) => x,
   None => {
       self.peeked = Some(self.inner.next());
       self.peeked.as_ref().unwrap()
   }
}
```
... but that doesn't seem cleaner to me.

aljazerzen:
> What all changes about the public API?

- `TokenStream` was renamed to `Tokenizer`,
- `Tokenizer` has a function `fn validated_values(self) -> Validator`,
- `Validator` does the work that was previously done in `convert_token(...) -> PyObject`,
- both of them now produce `Token`, which is a cleaned up version of `SpannedToken`,
- `Tokenizer` does not implement traits from `combine` crate, but has two new functions that allow their implementation elsewhere,

In short a few fiddly changes that are needed for either Rust parser or just simplification reasons.

msullivan (approved):

PR #5624, opened 2023-06-08
Fix casts to scalars

### LINKED ISSUE #5616: casts from json to scalar subtypes don't check constraints
```
edgedb> create scalar type x extending str { create constraint expression on (__subject__ not like 'a%') };
OK: CREATE SCALAR TYPE
edgedb> select <x>to_json('"a"');
["a"]
edgedb> select <x><str>to_json('"a"');
edgedb error: ConstraintViolationError: invalid x
  Detail: invalid scalar type 'default::x'
```

` select <x>to_json('"a"')` ought to raise an error also.

(I don't think this is specific to json, of course.)
I think the issue is that we are only ever casting to `text`. We never emit a cast to the real target type.

aljazerzen:
Closes #5616

We've had a codepath to fix this bug for enum already, but now I extended it to anything that:
- is a scalar,
- it not the top-most concrete base of itself,
- is not an object (does not have a `ir_set.path_id`).

### DIFF
--- a/edb/edgeql/compiler/casts.py
+++ b/edb/edgeql/compiler/casts.py
@@ -154,7 +154,10 @@ def compile_cast(
     elif orig_stype.issubclass(ctx.env.schema, json_t):
         if (
             new_stype.is_enum(ctx.env.schema)
-            or not _is_topmost_concrete_scalar(new_stype, ctx)
+            or (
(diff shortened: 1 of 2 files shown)


msullivan (approved):
This looks great! Thanks

PR #5627, opened 2023-06-09
Add a pgvector extension package

msullivan:
Contents go into ext::pgvector.

Some compiler changes were made to get various array <-> vector casts
to work.

The test suite includes tests that the indexes are used.

### DIFF
--- a/edb/lib/ext/pgvector.edgeql
+++ b/edb/lib/ext/pgvector.edgeql
@@ -1,0 +1,165 @@
+#
+# This source file is part of the EdgeDB open source project.
+#
+# Copyright 2023-present MagicStack Inc. and the EdgeDB authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+create extension package pgvector version '0.4.0' {
+    set ext_module := "ext::pgvector";
+    set sql_extensions := ["vector >=0.4.0,<0.5.0"];
+
+    create module ext::pgvector;
+
+    create scalar type ext::pgvector::vector extending std::anyscalar {
+        set id := <uuid>"9565dd88-04f5-11ee-a691-0b6ebe179825";
+        set sql_type := "vector";
+        set sql_type_scheme := "vector({__arg_0__})";
+        set num_params := 1;
+    };
+
+    create cast from ext::pgvector::vector to std::json {
+        set volatility := 'Immutable';
+        using sql 'SELECT val::text::jsonb';
+    };
+
+    create cast from std::json to ext::pgvector::vector {
+        set volatility := 'Immutable';
+        using sql $$
+        SELECT (
+            nullif(val, 'null'::jsonb)::text::vector
+        )
+        $$;
+    };
+
+    # All casts from numerical arrays should allow assignment casts.
+    create cast from array<std::float32> to ext::pgvector::vector {
+        set volatility := 'Immutable';
+        using sql cast;
+        allow assignment;
+    };
+
+    create cast from array<std::float64> to ext::pgvector::vector {
+        set volatility := 'Immutable';
+        using sql cast;
+        allow assignment;
+    };
+
+    create cast from array<std::int16> to ext::pgvector::vector {
+        set volatility := 'Immutable';
+        using sql $$
+        SELECT val::float4[]::vector
+        $$;
+        allow assignment;
+    };
+
+    create cast from array<std::int32> to ext::pgvector::vector {
+        set volatility := 'Immutable';
+        using sql cast;
+        allow assignment;
+    };
+
+    create cast from array<std::int64> to ext::pgvector::vector {
+        set volatility := 'Immutable';
+        using sql $$
+        SELECT val::float4[]::vector
+        $$;
+        allow assignment;
+    };
+
+    create cast from ext::pgvector::vector to array<std::float32> {
+        set volatility := 'Immutable';
+        using sql cast;
+    };
+
+    create function ext::pgvector::euclidean_distance(
+        a: ext::pgvector::vector,
+        b: ext::pgvector::vector,
+    ) -> std::float64 {
+        set volatility := 'Immutable';
+        # Needed to pick up the indexes when used in ORDER BY.
+        set prefer_subquery_args := true;
+        using sql 'SELECT a <-> b';
+    };
+
+    create function ext::pgvector::neg_inner_product(
+        a: ext::pgvector::vector,
+        b: ext::pgvector::vector,
+    ) -> std::float64 {
+        set volatility := 'Immutable';
+        # Needed to pick up the indexes when used in ORDER BY.
+        set prefer_subquery_args := true;
+        using sql 'SELECT (a <#> b)';
+    };
+
+    create function ext::pgvector::cosine_distance(
+        a: ext::pgvector::vector,
+        b: ext::pgvector::vector,
+    ) -> std::float64 {
+        set volatility := 'Immutable';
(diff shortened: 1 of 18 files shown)


edb/lib/ext/pgvector.edgeql
```diff
@@ -0,0 +1,189 @@
...
+
+    create cast from ext::pgvector::vector to std::str {
+        set volatility := 'Immutable';
+        using sql cast;
+    };
+
+    create cast from std::str to ext::pgvector::vector {
+        set volatility := 'Immutable';
+        using sql cast;
+    };
```
1st1, line 41: My stance is that we should drop this. We don't have `<str>` casts for every format of EdgeDB, we use `to_str()` and `to_datatype()` functions for it. The reason for that is configurability of casting + making sure that we only accept data input compatible with EdgeDB's notion of what's acceptable for the given type.

Given that we have this: `select <pgvector::vector>[1.1, 2.1]` I don't think we need to have string casts at least in 3.0.

Moreover, we don't even know how stable the string casting is in pgvector itself, see https://github.com/pgvector/pgvector/issues/147 as an example.

cc @elvis @msullivan @vpetrovykh

1st1 (changes requested):

edb/lib/ext/pgvector.edgeql
elprans: Yes, we don't support `str -> array` casts, and neither should we do here.

1st1:
Can we make `ext::` optional in our name resolution algo? Basically the idea was to only require it if there's shadowing of user namespace of the extension.

cc @elprans

elprans:
> Can we make `ext::` optional in our name resolution algo?

I don't think it's a good idea.  This basically makes things equivalent to `std`, then what's the point of `ext`?

1st1:
> I don't think it's a good idea. This basically makes things equivalent to std, then what's the point of ext?

I have an answer but the fact that there's *some* discussion here means that we don't want to hastily implement it anyways. Let's have the discussion some other time.

@msullivan let's just drop the string casts and this is good to go, IMO, unless @elprans wants to do a deeper review.

edb/lib/ext/pgvector.edgeql
```diff
@@ -0,0 +1,189 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+create extension package pgvector version '1.0' {
```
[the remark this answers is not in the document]
msullivan: I just pushed extensions constraints and a version pin

edb/lib/ext/pgvector.edgeql
msullivan: My inclination is to keep the casts, but I don't feel strongly.

elprans: We don't have a cast from `str` to `array`, so just from that standpoint I would hold off on adding it for `pgvector`.

vpetrovykh: I'm fine with the idea that similar to `array` we skip the direct `str` casts in favour of `to_vector` and `to_str`.

vpetrovykh: At the end of the day the more "natural" cast from a literal is still from an array of numbers of some sort. Which we have. And we also have a cast to `array<float32>`.

edb/lib/ext/pgvector.edgeql
```diff
@@ -0,0 +1,189 @@
...
+    };
+
+    create function ext::pgvector::inner_product(
+        a: ext::pgvector::vector,
+        b: ext::pgvector::vector,
+    ) -> std::float64 {
+        set volatility := 'Immutable';
+        # Needed to pick up the indexes when used in ORDER BY.
+        set prefer_subquery_args := true;
+        using sql 'SELECT -(a <#> b)';
```
elprans, line 128: Hmm, so from pgvector README:

> Note: `<#>` returns the negative inner product since Postgres only supports ASC order index scans on operators

Why are we reverting this?  Have you checked that `ORDER BY inner_product(a, b) DESC` still uses the index?

edb/lib/ext/pgvector.edgeql
msullivan: Alright, I removed the `str` casts. Not adding conversion functions for now because `to_str` wouldn't be able to go into `std`

edb/lib/ext/pgvector.edgeql
msullivan: No, I was sloppy and only checked euclidean distance.

I generalized the test and sure enough it didn't work, so I turned inner_product into neg_inner_product

edb/schema/extensions.py
```diff
@@ -337,6 +337,20 @@ class DeleteExtension(
 
     astnode = qlast.DropExtension
 
+    def apply(
+        self,
+        schema: s_schema.Schema,
+        context: sd.CommandContext,
+    ) -> s_schema.Schema:
+        testmode = context.testmode
+        context.testmode = True
```
elprans, line 346: Hm. Why?

tests/schemas/dump_v4_default.esdl
```diff
@@ -0,0 +1,25 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+scalar type v3 extending ext::pgvector::vector<3>;
```
1st1, line 20: Shouldn't we require you to have `using extension pgvector;` like with GraphQL? I think we should.

tests/test_edgeql_vector.py
```diff
@@ -0,0 +1,875 @@
...
+            '''
+                select <array<float32>>
+                    <ext::pgvector::vector>array_agg(
+                        Raw.p_deepf64 order by Raw.p_deepf64
+                    );
+            ''',
+            [res],
+        )
+
+    @test.xfail('''
```
elprans, line 349: Wasn't this fixed by #5629

tests/schemas/dump_v4_default.esdl
msullivan: We do; it's just configured in the test because test schemas elide the `module default` and so you aren't at the toplevel

edb/schema/extensions.py
msullivan: Because create/deleting extensions does operations that aren't allowed in normal DDL. The create side already had this, but there was something on the delete side that ended up needing it too

tests/test_edgeql_vector.py
msullivan: Yeah, I just haven't rebased this yet

msullivan: done

tests/schemas/dump_v4_default.esdl
1st1: Got it, thanks!

msullivan:
I just force pushed a new history, which I will rebase to land

elprans (approved):

edb/lib/ext/pgvector.edgeql
```diff
@@ -0,0 +1,179 @@
...
+    };
+
+    create cast from array<std::bigint> to ext::pgvector::vector {
+        set volatility := 'Immutable';
+        using sql $$
+        SELECT val::numeric[]::vector
+        $$;
+        allow assignment;
+    };
+
```
1st1, line 95: I don't think we need decimal or bigint/decimal casts -- we don't need them and it's questionable if they're useful at all, as there will be major loss of precision here

1st1 (approved):
Approving, but I think we should drop decimal/bigint casts.

edb/lib/ext/pgvector.edgeql
elprans: We allow `float <> decimal/bigint` casts, though.

elprans: E.g

```
select <array<float32>>[1n, 2n, 3n];
```

works, and I don't see a fundamental difference between `array<float32>` and `pgvector::vector` in terms of casting.

PR #5508, opened 2023-05-18
Fix cardinality of derived __tname__ and __tid__ pointers

aljazerzen:
Closes #4719

This might not be the best place to set the cardinality, but it works. Also not sure why this matters only for `group` and not for plain `select`s.

PR #5634, opened 2023-06-09
Fix the `__tname__` cardinality fix

aljazerzen:
A followup for #5508, which is a hack that sets the cardinality of pointers during eql compilation.

This fix was an exploration project and it involved debugging:
- how new stypes are derived and when,
- how shape pointers are created and their cardinality requirements saved into the context,
- how cardinality is inferred (and validated) after the main eql compilation,
- how paths are extended and how does `Pointer.defined_here` work.

Fun times, many breakpoints, a bit better code.

### DIFF
--- a/edb/edgeql/compiler/viewgen.py
+++ b/edb/edgeql/compiler/viewgen.py
@@ -2102,9 +2102,6 @@ def _inline_type_computable(
                 ctx=scopectx
             )
 
-        ctx.env.schema = ptr.set_field_value(
-            ctx.env.schema, 'cardinality', qltypes.SchemaCardinality.One)
-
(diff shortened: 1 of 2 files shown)


edb/edgeql/compiler/viewgen.py
```diff
@@ -2104,19 +2102,15 @@ def _inline_type_computable(
                 ctx=scopectx
             )
 
-        ctx.env.schema = ptr.set_field_value(
-            ctx.env.schema, 'cardinality', qltypes.SchemaCardinality.One)
+    ctx.env.schema = ptr.set_field_value(
```
msullivan, line 2105: Add a comment

msullivan (approved):
Good detective work!

PR #5643, opened 2023-06-12
Emphasize that changes are not just cosmetic

Dhghomon:
Just a tiny PR for a minor quibble: I think the future-proofing reasons for the SDL changes in 3.0 could be emphasized a bit more and moved up before the note so that all of its justifications are made clear up front.


msullivan:
Closing because we didn't end up merging this and 3.x was released a while ago

PR #5649, opened 2023-06-13
Fix a bug where indexes on a computable property that directly references another property would trigger a `property not found` exception

### LINKED ISSUE #5548: Indexes on Computed Properties cause Invalid Reference Error
EdgeDB 4.0-dev.7528+local (repl 3.1.0-dev)

Reproduction:
1. run `edb cli migration create` on the following schema:

```
module default {
type Person {

    property name2 := .name;

    index on (.name2);
}

}
```

The error is:
```
edgedb error: InvalidReferenceError: object type 'default::Person' has no link or property 'name2' (on line 1, column 1)
  Hint: did you mean 'name'?
<stacktraces>
        File "edgedb/edb/edgeql/compiler/setgen.py", line 843, in resolve_ptr_with_intersections
      edb.errors.InvalidReferenceError: object type 'default::Person' has no link or property 'name2'
```

zchenb:
Fix #5548


zchenb:
No longer needed as the original issue is no longer present

PR #5650, opened 2023-06-13
Minor refactors on `schema.py` to avoid assignment

zchenb:
I find this piece of code confusing and not easy to debug. Changing to avoid reassignment.


msullivan (approved):

msullivan:
Change the commit message, though.
It's not a fix, but a refactor

PR #5654, opened 2023-06-14
Refactor Rust libraries

aljazerzen:
Follow up for #5645 and  https://github.com/edgedb/edgedb-cli/pull/1080.

- removes "python" from default features of edgeql-parser,
- renames edgeql-rust to edgeql-parser-python,
- moves edgeql-parser-* libraries under edgeql-parser dir,


tailhook (approved):
Looks goon. Although, I'm Not sure I understand why everything is nested under `edgeql-parser` now.

PR #5657, opened 2023-06-14
Expose maintenance_work_mem via config

msullivan:


elprans (approved):

msullivan:
I confirmed that this can be used to get a big index creation to go through

PR #5659, opened 2023-06-15
Add an ADMINISTER command to do reindex

msullivan:
I had originally called it `reindex_object`, but remembered that
Elvis wanted `NOUN_VERB` names for these functions, so I made it
`object_reindex`.

Verified that it does work to get pgvector indexes that were
created too early working.

### DIFF
--- a/edb/server/compiler/compiler.py
+++ b/edb/server/compiler/compiler.py
@@ -1440,7 +1440,7 @@ def _compile_ql_reindex(
 
     if len(ql.expr.args) != 1 or ql.expr.kwargs:
         raise errors.QueryError(
-            'object_reindex() takes exactly one position argument',
+            'reindex() takes exactly one position argument',
             context=ql.expr.context,
         )
 
@@ -1453,7 +1453,7 @@ def _compile_ql_reindex(
             pass
         case _:
             raise errors.QueryError(
-                'argument to object_reindex() must be an object type',
+                'argument to reindex() must be an object type',
                 context=arg.context,
             )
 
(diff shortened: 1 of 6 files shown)


edb/server/compiler/compiler.py
```diff
@@ -1449,8 +1506,10 @@ def _compile_ql_administer(
             )
 
         return dbstate.MaintenanceQuery(sql=(b'ANALYZE',))
-    elif ql.expr.func == 'repair_schema':
+    elif ql.expr.func == 'schema_repair':
         return ddl.administer_repair_schema(ctx, ql)
+    elif ql.expr.func == 'object_reindex':
```
elprans, line 1511: Why `object_reindex` and not just `reindex`?  Is this somehow specific to object relations or should also apply to link relations (e.g. indexes on link properties).

msullivan:
I didn't call it "reindex", because "reindex" is reserved. Obviously that
makes no sense. I'll change it.

1st1:
Yeah, let's call it "reindex", other than that lgtm!

elprans (commented):
What if I want to reindex a specific index?

msullivan:
> What if I want to reindex a specific index?

That's not supported currently.

I'm actually not sure if there is any particularly way to specify a specific index while using function call syntax. For DDL ALTERs/DROPs we recapitulate the definition, basically, and that wouldn't be syntactically valid here. If we were doing custom syntax, it would be easy, of course.

One option would be to support specifying pointers, and then reindex all indexes that reference that pointer.

1st1 (approved):

elprans (approved):

elprans (approved):

PR #5664, opened 2023-06-16
Ensure changing scalar base types throws appropriately

efaust:
Rework scalar error messages. There were several wrinkles here, including enum
scalars being treated differently when generating the alter delta and when
running commands from scratch. Try to sanify the checks a little so it always
works as expected, and spruce up some error messages along the way.

### DIFF
--- a/edb/schema/scalars.py
+++ b/edb/schema/scalars.py
@@ -321,6 +321,12 @@ class ScalarType(
                 ),
             )
             alter.add(rebase)
+
+        # Changing enum_values is the respoinsiblity of the rebase command.
+        # Either it's in the one we synthesized above, or, the rebase is doomed
+        # to throw. When we run the ddl directly, the ALTER will not have a
+        # enum_values set, so discard here for symmetry.
+        alter.discard_attribute('enum_values')
         return alter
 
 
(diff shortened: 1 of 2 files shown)


edb/schema/scalars.py
```diff
@@ -677,55 +677,48 @@ def apply(
...
                     raise errors.SchemaError(
-                        f'cannot add another enum as supertype, '
-                        f'use EXTENDING without position qualification')
-
-                all_bases.extend(bases)
-
-            if len(all_bases) > 1:
-                raise errors.SchemaError(
-                    f'cannot set more than one enum as supertype')
+                        f'enums may not have multiple supertypes')
```
msullivan, line 691: While you're here, could you either add a name or a source context here?

edb/schema/scalars.py
```diff
@@ -677,55 +677,48 @@ def apply(
         self.scls = scls
         assert isinstance(scls, ScalarType)
 
-        cur_labels = scls.get_enum_values(schema)
-
-        if cur_labels:
-
+        if self.scls.is_concrete_enum(schema):
```
msullivan, line 680: Could we add a comment about the cursed nature of `enum_values` along this code path? This nicely avoids the problem, and is definitely cleaner, but it's still a potential landmind in the future

msullivan (approved):
Great clean up. This was a tangle.

edb/schema/scalars.py
```diff
@@ -309,8 +309,9 @@ def as_alter_delta(
 
         # If this is an enum and enum_values changed, we need to
         # generate a rebase.
+        old_enum_values = self.get_enum_values(self_schema)
         enum_values = alter.get_local_attribute_value('enum_values')
-        if enum_values is not None:
+        if old_enum_values and enum_values:
```
msullivan, line 314: Actually, what if we removed the `enum_values` attribute set from `alter` in this case?
That might eliminate the cursedness by fully pushing things down to the rebase, even in the delta case.

PR #5666, opened 2023-06-16
Fix constraints on custom abstract scalars

### LINKED ISSUE #4653: ISE when creating a scalar inheriting from an abstract scalar and a non-abstract scalar
```
abstract scalar type non_empty_str extending str {
    constraint min_len_value(1);
}

scalar type name_str extending str {
    constraint max_len_value(40);
}

scalar type required_name_str extending non_empty_str, name_str;
```

RESULT:

```
➜  edgedb git:(dev) ✗ edgedb -I <project_name> migrate
edgedb error: InternalServerError: type "edgedbpub.615f49c8-6117-11ed-b3bf-5d058c5f93bb_domain" does not exist
  Server traceback:
      edb.errors.InternalServerError: type "edgedbpub.615f49c8-6117-11ed-b3bf-5d058c5f93bb_domain" does not exist
```

aljazerzen:
Closes #4653


edb/pgsql/delta.py
```diff
@@ -2035,18 +2037,20 @@ def constraint_is_effective(cls, schema, constraint):
...
+        match subject:
+            case s_pointers.Pointer():
+                if subject.generic(schema):
+                    return True
+                else:
+                    return has_table(subject.get_source(schema), schema)
+            case s_objtypes.ObjectType():
+                return has_table(subject, schema)
+            case s_scalars.ScalarType():
+                return not subject.get_abstract(schema)
```
aljazerzen, line 2052: This is the only impactful change.

edb/pgsql/delta.py
```diff
@@ -2155,10 +2170,16 @@ def apply(
             and not context.is_creating(subject)
         ):
             op = self.enforce_constraint(
-                constraint, schema, context, self.source_context)
+                constraint, schema, self.source_context
+            )
+
+            # XXX: PostCommand or maybe even pg_ops have incorrect type
+            # annotations, but I cannot figure them out.
+            op2: sd.Command = op  # type: ignore
```
aljazerzen, line 2178: @msullivan

What is the difference between:
- `schema.delta.Command` and
- `pgsql.dbops.Command`?

`sd.Command` gets produced when we diff two schemas? And is then compiled to `pgsql.dbops.Command`?

----

`pgsql.delta.MetaCommand` has a field `pgops`, which collects all operations to be executed.

It should contain `pgsql.dbops.Command`, right? And not `sd.Command`?

aljazerzen: If I was correct, this should work:
```diff python
 class MetaCommand(sd.Command, metaclass=CommandMeta):
+    pgops: ordered.OrderedSet[dbops.Command]

```

edb/pgsql/delta.py
```diff
@@ -2019,7 +2019,9 @@ class DeleteAnnotationValue(
 
 class ConstraintCommand(MetaCommand):
     @classmethod
-    def constraint_is_effective(cls, schema, constraint):
+    def constraint_is_effective(
+        cls, schema: s_schema.Schema, constraint: s_constr.Constraint
+    ):
```
msullivan, line 2024: `-> bool`

edb/pgsql/delta.py
msullivan: A `dbops.Command` represents some concrete postgres DDL command. A `delta.Command` represents a change between edgedb schemas and/or an edgedb DDL command.

I think that `pgops` should probably be just `dbops.Command`s, but I'm not 100% sure what the situation is

msullivan (approved):
Really good cleanups here.
Could you split it into the minimal fix and the cleanups, though, so we can cherry-pick the fix to 3.x?
(Unless you think this whole thing can be reasonably backported without much risk)

edb/pgsql/delta.py
aljazerzen: Cool, i'll continue this in #5672

PR #5668, opened 2023-06-16
Make Object.property in indexes detached

aljazerzen:
A followup for https://github.com/edgedb/edgedb/pull/5606#issuecomment-1584376671

This is techincally a breaking change, since any index definitions
that are using `Object.property` will now throw errors. I do think it is
quite minor and will probably not affect anyone, as this pattern is not documented
and treated as detached in all other SDL expressions apart from computeds.

### DIFF
--- a/edb/schema/indexes.py
+++ b/edb/schema/indexes.py
@@ -589,12 +589,13 @@ class IndexCommand(
                     singletons=frozenset([subject]),
                     apply_query_rewrites=False,
                     track_schema_ref_exprs=track_schema_ref_exprs,
+                    detached=True,
                 ),
             )
 
             # Check that the inferred cardinality is no more than 1
             if expr.irast.cardinality.is_multi():
-                raise errors.ResultCardinalityMismatchError(
+                raise errors.SchemaDefinitionError(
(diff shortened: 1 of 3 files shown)


edb/schema/indexes.py
```diff
@@ -589,12 +589,13 @@ def compile_expr_field(
...
                     apply_query_rewrites=False,
                     track_schema_ref_exprs=track_schema_ref_exprs,
+                    detached=True,
                 ),
             )
 
             # Check that the inferred cardinality is no more than 1
             if expr.irast.cardinality.is_multi():
-                raise errors.ResultCardinalityMismatchError(
+                raise errors.SchemaDefinitionError(
```
aljazerzen, line 598: I've changed this because when running `edgedb migration create`, the CLI would print a stacktrace instead of a nice error. Does it have any other effect?

msullivan: No, I don't think so

msullivan (approved):
Looks good.

The only question is whether it should go into 3.0, I think.
(I'm leaning no, at this point?)

PR #5669, opened 2023-06-16
Partial paths in triggers refer to __new__

aljazerzen:
Follow up for https://github.com/edgedb/edgedb/pull/5606#issuecomment-1584376671.

Triggers that are not Delete, can now contain partial paths
that refer to `__new__`. Not a breaking change.


msullivan:
If I recall correctly, it was fairly deliberate to not have partials paths, to avoid potential confusion between `__old__` and __new__`, though I'm open to an argument against that.
(It *would* probably be more convenient sometimes)

aljazerzen:
Only now I've realized that triggers are only constructs that don't support neither of `__source__` or `__subject__`. This is a valid justification for triggers not having partial paths.

PR #5672, opened 2023-06-16
A bit of pg delta refactor

aljazerzen:
Split off #5666


edb/edgeql/compiler/options.py
```diff
@@ -141,7 +141,7 @@ class CompilerOptions(GlobalCompilerOptions):
     #: Type references that should be remaped to another type.  This
     #: is for dealing with remapping explicit type names in schema
     #: expressions to their subtypes when necessary.
-    type_remaps: Dict[s_obj.Object, s_types.Type] = dc_field(
+    type_remaps: Dict[s_obj.Object, s_obj.Object] = dc_field(
```
msullivan, line 144: What non types go in here?

msullivan (approved):

edb/edgeql/compiler/options.py
aljazerzen: It can also be `s_pointers.Pointer` when compiling constraints.

PR #5677, opened 2023-06-21
mention that links are distinct

ssendev:
I was surprised to learn that `many-to-many` relationships aren't quite as many as i believed.

I found this discussion https://github.com/edgedb/edgedb/discussions/4923 which mentions it but it would have been nice to have known it sooner.

I guess ideally there would be a `multi link not distinct` even if that means upserts/on-conflict do not work.


raddevon (approved):
I made a few tweaks here. Wanted to mention an alternative solution that doesn't require an intermediary object type.

Really appreciate this contribution. Great stuff! Apologies for the delay getting it merged.

PR #5688, opened 2023-06-23
Add Java to Get Started > Client Libraries docs

raddevon:


quinchs (changes requested):

docs/intro/clients.rst
```diff
@@ -94,6 +95,21 @@ Configure the environment as needed for your preferred language.
...
+        <groupId>com.edgedb</groupId>
+        <artifactId>driver</artifactId>
+    </dependency>
+
+  .. code-tab::
+    :caption: Gradle (Java)
+
+    // build.gradle
+    implementation 'com.edgedb:driver'
+
```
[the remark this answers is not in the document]
raddevon: @quinchs OK, cool. This should do it. Let me know if you see any other ways to improve.

docs/intro/clients.rst
```diff
@@ -94,6 +95,16 @@ Configure the environment as needed for your preferred language.
...
 
+  .. code-tab:: bash
+    :caption: Maven (Java)
+
+    $ touch Main.java
+
+  .. code-tab:: bash
+    :caption: Gradle (Java)
+
+    $ touch Main.java
```
quinchs, line 106:
```suggestion
    :caption: Java

    $ touch Main.java
```

quinchs (approved):

docs/intro/clients.rst
raddevon: The one nice thing about having both is that, if the tab labels match, the docs will automatically switch all of the tabbed sections at once. So, when I select the "Maven (Java)" tab in this section, it will automatically select that tab in the following tabs as well. I understand it doesn't make much sense otherwise, but do you think that's a worthwhile reason to keep both or would it just lead to confusion?

PR #5698, opened 2023-06-26
Pin Postgres version in `BackendRuntimeParams` passed to compiler

elprans:
Force Postgres version in `BackendRuntimeParams` to be the
minimal supported, because the compiler does not rely on
the version, and not pinning it would make the remote compiler
pool refuse connections from clients that have differing versions
of Postgres backing them.

### DIFF
--- a/edb/server/server.py
+++ b/edb/server/server.py
@@ -41,6 +41,7 @@ import uuid
 import immutables
 from jwcrypto import jwk
 
+from edb import buildmeta
 from edb import errors
 
 from edb.common import devmode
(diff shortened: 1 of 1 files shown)


edb/server/server.py
```diff
@@ -552,6 +553,19 @@ def _idle_gc_collector(self):
...
 
     async def _create_compiler_pool(self):
+        # Force Postgres version in BackendRuntimeParams to be the
+        # minimal supported, because the compiler does not rely on
+        # the version, and not pinning it would make the remote compiler
+        # pool refuse connections from clients that have differing versions
+        # of Postgres backing them.
+        runtime_params = self.get_backend_runtime_params()
+        min_ver = '.'.join(str(v) for v in defines.MIN_POSTGRES_VERSION)
+        runtime_params = runtime_params._replace(
```
nsidnev, line 563: Shouldn't we put this after into the `args` dict below?

elprans: :facepalm:

fantix (approved):

nsidnev (approved):

PR #5701, opened 2023-06-27
Do not lose constraint errmessage when a constraint is ALTERed

msullivan:
The worst part here was that this could occur when the constraint
was altered *implicitly* by _propagate_if_expr_refs because a
pointer that it used was modified.

### DIFF
--- a/edb/schema/constraints.py
+++ b/edb/schema/constraints.py
@@ -766,13 +766,6 @@ class ConstraintCommand(
                 attrs['subjectexpr'] = base_subjectexpr
                 inherited['subjectexpr'] = True
 
-        errmessage = attrs.get('errmessage')
-        if not errmessage:
-            errmessage = constr_base.get_errmessage(schema)
-            inherited['errmessage'] = True
-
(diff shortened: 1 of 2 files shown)


edb/schema/constraints.py
```diff
@@ -766,13 +766,6 @@ def _populate_concrete_constraint_attrs(
                 attrs['subjectexpr'] = base_subjectexpr
                 inherited['subjectexpr'] = True
 
-        errmessage = attrs.get('errmessage')
-        if not errmessage:
-            errmessage = constr_base.get_errmessage(schema)
-            inherited['errmessage'] = True
-
-        attrs['errmessage'] = errmessage
-
```
elprans, line 775: I'm a bit puzzled how `errmessage` gets inherited from abstract constraints now

msullivan: Through the normal generic inheritance mechanisms, right?

As far as I could tell, errmessage doesn't need any special handling

elprans (approved):

PR #5705, opened 2023-06-27
Remove Heroku deployment guide warning

fmoor:


fmoor:
I ran through the deployment guide and it still works. I did have to bump the dyno memory limit, but that was the only real change needed.

elprans (approved):

PR #5706, opened 2023-06-27
Avoid very slow error formatting in non-error cases

msullivan:
When resolve_ptr fails, it does a very slow error enrichment step that
computes levenshtein distances with a ton of different schema
objects. This is fine for actual error cases, but resolve_ptr gets
called in some call sites that swallow the error.

Skip doing the formatting in those cases. This gave a 30% speedup on
some queries.

I did this by checking the `track_ref` variable, since the only cases
where it makes sense to not track references is when we aren't serious
about the lookup, and because adding another param seemed
clunky. Maybe we should just add another, though.

### DIFF
--- a/edb/edgeql/compiler/viewgen.py
+++ b/edb/edgeql/compiler/viewgen.py
@@ -1502,10 +1502,19 @@ def _normalize_view_ptr_expr(
             )
 
             try:
+                is_linkprop_mutation = (
+                    is_linkprop
+                    and s_ctx.view_rptr is not None
+                    and s_ctx.view_rptr.exprtype.is_mutation()
+                )
+
                 ptrcls = setgen.resolve_ptr(
                     ptrsource,
                     ptrname,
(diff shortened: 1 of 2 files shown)


edb/edgeql/compiler/viewgen.py
```diff
@@ -1502,10 +1502,19 @@ def _normalize_view_ptr_expr(
...
+                    and s_ctx.view_rptr.exprtype.is_mutation()
+                )
+
                 ptrcls = setgen.resolve_ptr(
                     ptrsource,
                     ptrname,
-                    track_ref=False,
+                    track_ref=(
+                        False if not is_linkprop_mutation
+                        else shape_el_desc.ptr_ql
```
efaust, line 1516: This seems like it now tracks strictly more. I take it we were actually maybe using the errors sometimes?

efaust (approved):
OK, I buy this.

PR #5708, opened 2023-06-28
Move Typescript-related comment into own block

Dhghomon:
Quick doc fix for this page:

https://www.edgedb.com/docs/datamodel/globals

At the moment it contains a line that applies only to Typescript that shows up no matter which language sample is selected:

```The .withGlobals/.with_globals method returns a new Client instance that stores the provided globals and sends them along with all future queries.```

Easiest fix is probably just to turn it to a doc comment inside the Typescript code sample.

### DIFF
--- a/docs/datamodel/globals.rst
+++ b/docs/datamodel/globals.rst
@@ -44,7 +44,9 @@ which client library you're using.
 
     import createClient from 'edgedb';
 
-    const baseClient = createClient()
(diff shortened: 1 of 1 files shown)


scotttrinh (approved):
Looks good! We already have some pretty explicit documentation of this behavior in the TypeScript docs, so this feels like enough imo.

PR #5719, opened 2023-06-29
Remove a hack only necessary on our stable branches

msullivan:
Remove a hack from #5715 that made the fix work on databases
created prior to the fix.


elprans:
Seems like it's still needed?

msullivan:
Yeah, ugh. The hack was not intended to ever be necessary for operating on a freshly created database, but I think I added the second test *after* I added the hack, so I didn't catch that in the initial PR

PR #5733, opened 2023-07-03
Generalize Consul watching facility

elprans:
In preparation for adding more watches, factor out and generalize the KV
watching bits.

[Test run](https://github.com/edgedb/edgedb/actions/runs/5448300366).


edb/server/ha/base.py
```diff
@@ -16,33 +16,33 @@
...
 
-import urllib.parse
+from __future__ import annotations
 from typing import *
 
+import urllib.parse
 
-class ClusterProtocol:
-    def on_switch_over(self):
-        pass
```
fantix, line 25: Not an issue but this abstract method is still typing the following usage:

https://github.com/edgedb/edgedb/blob/82284a22a7e6877a595ed194e65a48c561e9f8cd/edb/server/ha/adaptive.py#L111

fantix (approved):
LGTM other than one issue with the repeated key prefix, pls see below.

edb/server/ha/base.py
elprans: Hm. I admin I'm a bit confused that codepath works and how to test it.  Mind taking a look?

fantix: The `test_ha_adaptive()` covered this codepath I believe, because I can see `adaptive: HA failover detected` in the logs running this test with `debug=True`.

https://github.com/edgedb/edgedb/blob/82284a22a7e6877a595ed194e65a48c561e9f8cd/edb/server/ha/adaptive.py#L109-L111

PR #5741, opened 2023-07-06
a array -> an array,etc::pgvector -> ext::pgvector

raddevon:


docs/stdlib/pgvector.rst
```diff
@@ -48,8 +48,8 @@ vector and specifying the desired length in angle brackets:
         embedding: v3
     }
 
-To populate your data, you can cast a array of any of the numeric types into
-``etc::pgvector::vector`` or simply assign that array directly:
+To populate your data, you can cast an array of any of the numeric types into
```
DijanaPavlovic, line 51: I think that we can also remove line 61 and only leave:
You can also cast the vectors into an
``array<float32>>``:

PR #5760, opened 2023-07-10
Update baremetal password instructions

### LINKED ISSUE #5644: Doc: CLI error when following the bare metal installation guide: `The --admin option requires --unix-path or local instance name`
When following the current version of the [bare metal installation guide](https://www.edgedb.com/docs/guides/deployment/bare_metal) (or [source here](https://github.com/edgedb/edgedb/blob/master/docs/guides/deployment/bare_metal.rst)), the admin is unable to set a password with this command:
```
$ edgedb --port 5656 --tls-security insecure --admin query \
   "ALTER ROLE edgedb SET password := 'test'"
```

The output is:
```
edgedb error: The --admin option requires --unix-path or local instance name
```

- EdgeDB Version: ? (could not execute the given statement due to the missing password when following the guide)
- EdgeDB CLI Version: `EdgeDB CLI 3.1.2+82e02cf`
- OS Version: `Debian GNU/Linux 11 (bullseye) x86_64, Kernel 5.10.0-23-amd64`

Steps to Reproduce:

(Basically follow the bare metal guide as linked above.)

1. Use an empty Debian VM/system without any `edgedb`-related packages installed.
2. Install the `edgedb` packages.
3. Configure and start the service.
4. Try to set a password for the `edgedb` role.

How should one set a "global" administrator password using the CLI?

raddevon:
Bare metal password instructions need to change since `--admin` without `--unix-path` no longer works.

See https://edgedb.slack.com/archives/CNECHA9EW/p1685527561807589 and https://github.com/edgedb/edgedb/issues/5644 for context.

Closes #5644

### DIFF
--- a/docs/guides/deployment/bare_metal.rst
+++ b/docs/guides/deployment/bare_metal.rst
@@ -109,13 +109,26 @@ Save the file and exit, then restart the service.
 
 Set a password
 ==============
-There is no default password. Set a password by connecting from localhost.
+There is no default password. To set one, you will first need to get the Unix
+socket directory. You can find this by looking at your system.d unit file.
+
+.. code-block:: bash
+
+    $ sudo systemctl cat edgedb-server-3
+
+Look for a line starting with ``ExecStart=``. Get the value being passed for
+``--runstate-dir``. If that value contains ``%t``, replace that with ``/run``.
+So, ``%t/edgedb`` would become ``/run/edgedb``. Use that value in the next
+series of commands as shown.
+
+Set a password by connecting from localhost.
 
 .. code-block:: bash
 
    $ echo -n "> " && read -s PASSWORD
-   $ sudo edgedb --port 5656 --tls-security insecure --admin query \
-      "ALTER ROLE edgedb SET password := '$PASSWORD'"
+   $ sudo edgedb --port 5656 --tls-security insecure --admin \
+      --unix-path <your-runstate-dir-value> \
+      query "ALTER ROLE edgedb SET password := '$PASSWORD'"
 
 The server listens on localhost by default. Changing this looks like this.
 


fantix:
We could make the server "--emit-server-status" into a JSON file and read `socket_dir` from that file.

raddevon:
@fantix Does this mean the CLI would automatically get this info from the server and parse it, eliminating any manual work by the user, or would this just be a different process for the user?

In either case, sounds like there would be some work to do as far as implementation. If that's true, should this PR still go through as a stop-gap measure?

docs/guides/deployment/bare_metal.rst
```diff
@@ -109,13 +109,26 @@ Save the file and exit, then restart the service.
...
+Set a password by connecting from localhost.
 
 .. code-block:: bash
 
    $ echo -n "> " && read -s PASSWORD
-   $ sudo edgedb --port 5656 --tls-security insecure --admin query \
-      "ALTER ROLE edgedb SET password := '$PASSWORD'"
+   $ sudo edgedb --port 5656 --tls-security insecure --admin \
+      --unix-path <your-runstate-dir-value> \
+      query "ALTER ROLE edgedb SET password := '$PASSWORD'"
```
fantix, line 131: Ah it's too complicated to use `--emit-server-status`, we can just extract the value of `--runstate-dir` from the service definition:

```suggestion
   $ RUNSTATE_DIR=$(systemctl show edgedb-server-3 -P ExecStart | \
      grep -o -m 1 -- "--runstate-dir=[^ ]\+" | \
      awk -F "=" '{print $2}')
   $ sudo edgedb --port 5656 --tls-security insecure --admin \
      --unix-path $RUNSTATE_DIR \
      query "ALTER ROLE edgedb SET password := '$PASSWORD'"
```

docs/guides/deployment/bare_metal.rst
```diff
@@ -109,13 +109,29 @@ Save the file and exit, then restart the service.
...
+socket directory. You can find this by looking at your system.d unit file.
+
+.. code-block:: bash
+
+    $ sudo systemctl cat edgedb-server-3
+
+Look for a line starting with ``ExecStart=``. Get the value being passed for
+``--runstate-dir``. If that value contains ``%t``, replace that with ``/run``.
+So, ``%t/edgedb`` would become ``/run/edgedb``. Use that value in the next
+series of commands as shown.
```
fantix, line 122: Ah I mean the `systemctl show ... | grep ... | awk ...` command does exactly what you described here, so that the user could just copy-paste and execute. If you feel explaining the command is reasonable, maybe we can move this explainer *after* the command into an optional block?

fantix (approved):

docs/guides/deployment/bare_metal.rst
raddevon: Ah, right. I overlooked taking this out. Thanks for catching it!

PR #5769, opened 2023-07-12
Update insert docs to supply required release_year

scotttrinh:
A user helpfully pointed out that these insert statements were missing the required `release_year` value. The other solution was to make it optional, but I think we use this same schema in other places in the docs, so it seemed best to actually add it.


raddevon (approved):
👏

PR #5788, opened 2023-07-14
Multirange

vpetrovykh:
This is a draft PR because there are some issues left to work out:
1) There's a problem with `contains(multirange<cal::local_date>, cal::local_date)` not being "found" even though the generic function works for other multiranges.
2) Assorted regressions introduced (presumably mostly via cast changes, but possibly some other issues) in other existing tests.

The `multirange` functionality itself is there, though, with a bunch of new functions, etc. So it can be reviewed and commented on (and we can tweak the function names).


edb/edgeql/compiler/polyres.py
```diff
@@ -156,7 +158,7 @@ def find_callable(
             matched = [call]
 
     if len(matched) <= 1:
-        # Unabiguios resolution
+        # Unambiguios resolution
```
elprans, line 161:
```suggestion
        # Unambiguous resolution
```

edb/edgeql/compiler/polyres.py
```diff
@@ -238,7 +240,15 @@ def _get_cast_distance(
                 resolved_poly_base_type = resolved
 
             if resolved_poly_base_type == resolved:
-                return s_types.MAX_TYPE_DISTANCE if is_abstract else 0
+                if is_abstract:
+                    return s_types.MAX_TYPE_DISTANCE
+                elif arg_type.is_range() and param_type.is_multirange():
+                    # Ranges are implicitly cast into multiranges of the same
```
msullivan, line 246: Are we confident that's the behavior we want?

msullivan:
Take a look at https://github.com/edgedb/edgedb/blob/master/dev-notes/newtype-checklist.md and make sure you hit everything

edb/pgsql/compiler/relgen.py
```diff
@@ -2851,6 +2851,37 @@ def process_set_as_std_range(
...
+    #     ranges: array<range<anypoint>>,
+    #   )
+    #
+    #     into
+    #
+    #   <pg_range_type>(variadic ranges)
+    expr = ir_set.expr
+    assert isinstance(expr, irast.FunctionCall)
+
+    # N.B: kwargs go first and are sorted by name
```
msullivan, line 2872: There are no kwargs here?

edb/schema/operators.py
```diff
@@ -265,7 +266,9 @@ def _create_begin(
                         oper_all_tuples
                         and ptype.is_tuple(schema)
                     )
-                    oper_all_ranges = oper_all_ranges and ptype.is_range()
+                    oper_all_ranges = oper_all_ranges and (
```
msullivan, line 269: I'm not really sure I understand what all of this is trying to accomplish?

edb/schema/types.py
```diff
@@ -1359,7 +1362,13 @@ def _to_nonpolymorphic(
         schema: s_schema.Schema,
         concrete_type: Type,
     ) -> typing.Tuple[s_schema.Schema, Array]:
-        return Array.from_subtypes(schema, (concrete_type,))
+        st = self.get_subtypes(schema=schema)[0]
```
msullivan, line 1365: It sketches me out to have specific range handling here instead of a more generalized system for resolving nested polymorphic things.
Add a TODO comment about fixing it, at least?

edb/schema/types.py
```diff
@@ -2478,6 +2513,353 @@ def get_underlying_schema_class(cls) -> typing.Type[Collection]:
...
+    CollectionExprAlias,
+    Multirange,
+    qlkind=qltypes.SchemaObjectClass.ALIAS,
+):
+    # N.B: Don't add any SchemaFields to this class, they won't be
+    # reflected properly (since this inherits from the concrete Multirange).
+
+    @classmethod
+    def get_underlying_schema_class(cls) -> typing.Type[Collection]:
+        return Mulrirange
```
msullivan, line 2860: spelling

msullivan:
There is so much possibly needlessly duplicated code between each sort of collection type. It's not worth trying to fix that in this PR, but I wonder if we should do it at some point. The big question is whether we'll need more collection types, of course.

edb/pgsql/metaschema.py
```diff
@@ -387,6 +387,30 @@ def __init__(self) -> None:
         )
 
 
+class MultirangeToJsonFunction(dbops.Function):
+    """Convert anymultirange to a jsonb object."""
+    text = r'''
```
msullivan, line 392: This does the wrong thing when the input is `NULL`... and so does the existing `range_to_jsonb` function.

We have a test that I wrote that tries to test casts from NULL for every cast we have... but it doesn't handle the polymorphic casts, so it missed this.

I'll go fix that

edb/edgeql/compiler/polyres.py
vpetrovykh: Fixed

edb/edgeql/compiler/polyres.py
vpetrovykh: Yes, because it makes it so much easier to work with multiranges when you don't need to explicitly upcast ranges. The operations like `+`, `-`, `contains`, `overlaps`, etc. all can work with a mixture of ranges and multiranges.

The relationship between `range` and `multirange` is more like that of `int32` and `float64` than numbers and arrays. Every `int32` _can_ be represented as `float64`, but `float64` has more things it can represent, such as the result of division of 2 `int32` numbers, which _only sometimes_ is another `int32`. Similarly every `range` has a corresponding exact `multirange` to represent it, but you can do more elaborate operations on multiranges that aren't merely storing or removing "elements" like for arrays. Consider this:
```
edgedb> select range(1, 10) - range(2, 3);
edgedb error: InvalidValueError: result of range difference would not be contiguous
edgedb> select multirange([range(1,10)]) - range(2,3);
[
  [
    {"lower": 1, "upper": 2, "inc_lower": true, "inc_upper": false},
    {"lower": 3, "upper": 10, "inc_lower": true, "inc_upper": false}
  ]
]
```

edb/pgsql/compiler/relgen.py
vpetrovykh: Removed the copy-pasted comment

edb/schema/types.py
vpetrovykh: Added a TODO

vpetrovykh:
> There is so much possibly needlessly duplicated code between each sort of collection type. It's not worth trying to fix that in this PR, but I wonder if we should do it at some point. The big question is whether we'll need more collection types, of course.

I agree that there's a lot of overlap and duplication between range and multirange code. It might simplify future maintenance to refactor this, but it's not that clear that there's much benefit.

tests/test_edgeql_casts.py
```diff
@@ -2911,9 +2911,13 @@ async def test_edgeql_casts_json_set_02(self):
...
         # For *every* cast, try casting a value we know
         # will be represented as NULL.
+        #
+        # FIXME: eventually multirange casts should be included.
         casts = await self.con.query('''
             select schema::Cast { from_type: {name}, to_type: {name} }
-            filter not .from_type IS schema::ObjectType
+            filter
+                not .from_type is schema::ObjectType
+                and .to_type.name not like 'multirange%'
```
msullivan, line 2920: Don't opt multiranges out of the test, just use `json_only` for them when calling assert_query_result

edb/pgsql/metaschema.py
vpetrovykh: I've updated the SQL to include a `NULL` check just like for ranges.

msullivan (approved):
Looks good!

Please fix `test_edgeql_casts_all_null` to use `json_only` instead of skipping multirange casts, though.

tests/test_edgeql_casts.py
vpetrovykh: Oh, yeah, wrong fix!

msullivan (approved):

PR #5790, opened 2023-07-14
Doc cloud updates

raddevon:


docs/guides/cloud.rst
```diff
@@ -16,6 +16,76 @@ EdgeDB Cloud is the easiest way to host your EdgeDB instance. We offer two ways
...
+
+    * can contain only Latin alpha-numeric characters or ``-``
+    * cannot start with a dash (``-``) or contain double dashes (``--``)
+    * maximum instance name length is 61 characters minus the length of your
+      GitHub username (i.e., length of GitHub username + length of instance
+      name must be fewer than 62 characters)
+
+
+
+3. Configure your application
```
scotttrinh, line 77: I wonder if we can/should just move this into a section below the CLI and Cloud interfaces sections, like:

```
Deploying your application
====================

```

WDYT?

raddevon: Yeah, I like that. You also pushed me to re-evaluate more broadly. A lot of what I brought over was already in what we had. I've pulled out what wasn't there from the old guide and better integrated it which allowed me to delete the whole "Usage" section altogether, alongside moving the deployment info. Does this look better?

scotttrinh (approved):

PR #5813, opened 2023-07-19
Some test setup/teardown cleanups

msullivan:
* Rename SETUP/TEARDOWN_COMMANDS to PER_TEST_SETUP/TEARDOWN.
  They are run per-*test*, not per-database, which is the distinction
  from SETUP and TEARDOWN, but that is not at all clear.
* Get rid of TestExpressionsWithoutConstantFolding (incidentally
  the only user of SETUP_COMMANDS`) and merge it into TestExpressions.
  No reason to spin up a database for those pretty trivial tests.
* Drop empty SETUP/TEARDOWN scripts from TestExpressions.

### DIFF
--- a/edb/testbase/server.py
+++ b/edb/testbase/server.py
@@ -626,8 +626,9 @@ class ClusterTestCase(BaseHTTPTestCase):
     # library (e.g. declaring casts).
     INTERNAL_TESTMODE = True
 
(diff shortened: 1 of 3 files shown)


scotttrinh (approved):
🤘

PR #5816, opened 2023-07-20
Add basic routing and provider redirect for GitHub

scotttrinh:
Very far from finished, but this seems like a good point to stop and reflect on how to refactor and organize what I have here before plowing forward with the rest of the GitHub-specific flow and adding additional providers.

Some questions I have specifically: 

1. Seems like we might want the code to be a bit more modular, but it also seems like we tend to like larger files that encapsulate everything rather than splitting things into smaller modules.
2. I've mostly been googling/llm'ing around to try to find out how to do things the pythonic way, but would appreciate any direction on patterns that are common in our particular style. Things like dictionary access patterns, typing thoroughness, function declaration ordering and naming conventions, etc.
3. Looking for feedback on the config system here and how it might fit with the "newer" config system we have in mind. Specifically, I'd love to have optional values and constraints like `min_length` working. 

---

This PR adds a not-yet-working route at:

- `/db/{db_name}/ext/auth/authorize` that takes a `provider` query parameter and redirects to the provider's authorization endpoint with the required configuration.


edb/testbase/server.py
```diff
@@ -382,7 +382,7 @@ def http_con_read_response(
     ) -> tuple[bytes, dict[str, str], int]:
         resp = http_con.getresponse()
         resp_body = resp.read()
-        resp_headers = {k.lower(): v.lower() for k, v in resp.getheaders()}
+        resp_headers = {k.lower(): v for k, v in resp.getheaders()}
```
scotttrinh, line 385: I think this will break some existing tests, but I'll fix those in a separate PR based on this to keep the noise down.

pyproject.toml
```diff
@@ -17,6 +17,9 @@ dependencies = [
     'jwcrypto~=1.3.1',
     'psutil~=5.8',
     'setproctitle~=1.2',
+
+    'httpx~=0.24.1',
+    'respx~=0.20.1',
```
scotttrinh, line 22: `respx` unused at the moment, but figured I'd leave this for now. Fine to kill it if the noise is distracting.

edb/testbase/server.py
1st1: Yeah, it's a good idea to do that as a totally separate PR

pyproject.toml
1st1: Kill it please. Or we **will** forget and we try to keep the number of dependencies (test or runtime) to the absolute minimum.

edb/server/protocol/execute.pyi
```diff
@@ -0,0 +1,38 @@
...
+async def parse_execute_json(
+    db: dbview.Database,
+    query: str,
+    *,
+    variables: Mapping[str, Any] = immutables.Map(),
+    globals_: Optional[Mapping[str, Any]] = None,
+    output_format: compiler.OutputFormat = compiler.OutputFormat.JSON,
+    query_cache_enabled: Optional[bool] = None,
+) -> bytes:
+    ...
```
1st1, line 38: Please install editorconfig extension in your IDE/editor. It will then auto-configure to always insert the trailing newline to files.

edb/server/protocol/ext/auth.py
```diff
@@ -0,0 +1,282 @@
...
+                response=response,
+                redirect_uri=redirect_uri,
+                iss=request.host.decode(),
+                provider_name=provider_name,
+                db_config=db.db_config,
+            )
+        elif args[0] == "callback":
+            await handle_auth_callback(request, response)
+        else:
+            raise errors.BackendError
```
1st1, line 115: Better to always write error messages, even in WIPs :)

edb/server/protocol/ext/auth.py
```diff
@@ -0,0 +1,282 @@
...
+
+async def handle_request(request, response, db, args):
+    try:
+        # Set up routing to the appropriate handler
+        if args[0] == "authorize":
+            provider_name: str | None = _get_search_param(
+                request.url.query.decode("ascii"), "provider"
+            )
+            if provider_name is None:
+                raise errors.InternalServerError("No provider specified")
```
1st1, line 101: ISE are reserved only for unhandled server errors, they must never leak to an API or it's a bug.

You can find the full list of errors here: https://github.com/edgedb/edgedb/blob/master/edb/api/errors.txt

I'm not actually sure what error types we should use for auth. Maybe stick with BackendError for now...

edb/server/protocol/ext/auth.py
scotttrinh: Yeah, I think I should just make all of these `InternalServerError`s, but I'd like to discuss the error strategy here more generally, too.

edb/server/protocol/ext/auth.py
```diff
@@ -0,0 +1,282 @@
...
+            )
+        elif args[0] == "callback":
+            await handle_auth_callback(request, response)
+        else:
+            raise errors.BackendError
+    except Exception as ex:
+        if debug.flags.server:
+            markup.dump(ex)
+
+        # XXX Fix this when LSP "location" objects are implemented
```
1st1, line 120: Might take a while 💨

scotttrinh: Yeah, this is just copy pasta from somewhere else

edb/server/protocol/ext/auth.py
```diff
@@ -0,0 +1,282 @@
...
+        ex_type = errors.InternalServerError
+
+        _response_error(
+            response, http.HTTPStatus.INTERNAL_SERVER_ERROR, str(ex), ex_type
+        )
+
+
+def _response_error(response, status, message, ex_type):
```
1st1, line 128: FWIW I suggest using keyword-only arguments here for clarity, e.g.

```suggestion
def _response_error(*, response, status, message, ex_type):
```

and then:

```python
        _response_error(
            response=response, 
            status=http.HTTPStatus.INTERNAL_SERVER_ERROR, 
            message=str(ex),
            ex_type=ex_type
        )
```

Also better to have verbs in method name; this method should be called `fail_with_error` or something.

edb/server/protocol/ext/auth.py
```diff
@@ -0,0 +1,282 @@
...
+
+    response.body = json.dumps({"error": err_dct}).encode()
+    response.status = status
+    response.close_connection = True
+
+
+# OAuth 2 Provider for GitHub
+class GitHubProvider(BaseProvider):
+    def __init__(self, client_id, client_secret):
+        super().__init__("github", client_id, client_secret)
```
1st1, line 143:
```suggestion
    def __init__(self, *args, **kwargs):
        super().__init__("github", *args, **kwargs)
```

edb/server/protocol/ext/auth.py
scotttrinh: Ok, I'll switch these all to `BackendError` pending more details about how errors ought to work for extensions like this.

edb/server/protocol/ext/auth.py
```diff
@@ -0,0 +1,282 @@
...
+        # Check state value
+        # TODO: Look up state value from FlowState object
+        # flow_state = await db.get_flow_state(state, "github")
+        # if flow_state is None:
+        #    raise errors.UnauthorizedError("invalid state value")
+        data = {
+            'grant_type': 'authorization_code',
+            'code': code,
+            'state': state,
+            "redirect_uri": "http://localhost:8080/auth/callback",  # TODO: Get correct db-namespaced URL
```
1st1, line 167: I think you can pass the `server` object from the `protocol.pyx` to get host/port. And yeah, let's DB-namespace the URL.

edb/server/protocol/ext/auth.py
```diff
@@ -0,0 +1,282 @@
...
+    if provider_class is None:
+        raise errors.InternalServerError(f"unknown provider: {name}")
+    return provider_class(client_id, client_secret)
+
+
+def _get_auth_signing_key(db_config: SettingsMap):
+    auth_signing_key = _maybe_get_config(db_config, "xxx_auth_signing_key")
+    if auth_signing_key is None:
+        raise errors.InternalServerError(
+            "No auth signing key configured: Please set `cfg::Config.xxx_auth_signing_key`"
```
1st1, line 201: nit: Our code style dictates max 79 characters per line
```suggestion
            "No auth signing key configured: "
            "Please set `cfg::Config.xxx_auth_signing_key`"
```

edb/server/protocol/ext/auth.py
```diff
@@ -0,0 +1,282 @@
...
+    return urllib.parse.parse_qs(query).get(key, [None])[0]
+
+
+def _get_extension_path(url: httptools.parser.url_parser.URL) -> str:
+    path_parts = url.path.decode().split('/')
+
+    path_seg_parts = ["ext", "auth"]
+
+    try:
+        index = path_parts.index(path_seg_parts[0])
```
1st1, line 269: this stuff seems fragile. Is there a better simpler way to build the URL?

edb/server/protocol/ext/auth.py
```diff
@@ -0,0 +1,282 @@
...
+        if path_parts[index : index + len(path_seg_parts)] == path_seg_parts:
+            index += len(path_seg_parts) - 1
+    except ValueError:
+        print(f"ext/auth not found in path")
```
1st1, line 273: debug

1st1 (commented):
Lots of good stuff, @scotttrinh, awesome work. I've left a few nit comments.

edb/server/protocol/ext/auth.py
scotttrinh: This is copy pasta from the other protocol handlers, but I can definitely model this here.

edb/server/protocol/ext/auth.py
scotttrinh: Yeah, I'm curious about this: there is nothing I could come up with that would know enough about the routing to do this consistently. I could get to this a little easier from the `protocol.pyx` code since we're parsing the request there and have more structured details about the route at that point. wdyt?

edb/server/protocol/ext/auth.py
scotttrinh: Yeah, I ended up building this dynamically from the request URL in that `_get_extension_path` function, but as you pointed out later, it's pretty fragile. I'll move the extension's base route building to the router logic itself, it seems like that is really the right place that owns the route up to this point.

edb/server/protocol/ext/auth.py
scotttrinh: Can you suggest good reading material on this? I _think_ I understand how this works by unpacking the `args` list and the `kwargs` dict, but just to fill in the holes in my understanding.

1st1: I'd glance though the docs -- https://docs.python.org/3/tutorial/controlflow.html#more-on-defining-functions -- I just tried to google for a good blog post on this and they're all garbage. Python argument packing is pretty powerful (more so than in most other languages) so I suggest reading the docs fully just so that you know what's possible.

edb/server/protocol/ext/auth.py
scotttrinh: ⚡ [`6d508ea` (#5816)](https://github.com/edgedb/edgedb/pull/5816/commits/6d508eab75c7e433d5ad3b7137910a0e5ab63df7) @1st1 

How does that look?

edb/testbase/server.py
scotttrinh: https://github.com/edgedb/edgedb/pull/5824

edb/server/protocol/auth_ext/github.py
```diff
@@ -0,0 +1,50 @@
+import httpx
```
1st1, line 1: Please add copyright notice comments to all py files

edb/testbase/server.py
1st1: I don't see this change in #5824.

1st1 (approved):
Good stuff, the code looks great and the layout is 👌

edb/server/protocol/protocol.pyx
```diff
@@ -541,6 +544,17 @@ cdef class HttpProtocol:
...
                         request, response, db, args, self.server
                     )
+                elif extname == 'auth':
+                    netloc = (
+                        f"{request.url.host.decode()}:{request.url.port}"
+                            if request.url.port
+                            else request.url.host.decode()
+                    )
+                    extension_base_path = f"{request.url.schema.decode()}://" \
+                                          f"{netloc}/db/{dbname}/ext/auth"
```
fantix, line 554: I think there is also the `X-Forwarded-*` [headers family](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For) to consider here.

edb/server/protocol/auth_ext/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+from . import oauth
+
+
+async def handle_request(
+    request, response, db, extension_base_path: str, args: list[str]
+):
+    try:
+        # Set up routing to the appropriate handler
+        if args[0] == "authorize":
+            provider_name: str | None = _get_search_param(
```
fantix, line 36: (not something to fix) repeating the type here feels a bit redundant to me as `_get_search_param()` already defined a returning type 🤔

edb/server/protocol/auth_ext/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+        "code": ex_type.get_code(),
+    }
+
+    response.body = json.dumps({"error": err_dct}).encode()
+    response.status = status
+    response.close_connection = True
+
+
+def _get_search_param(query: str, key: str) -> str | None:
+    params = urllib.parse.parse_qs(query).get(key)
```
fantix, line 85: Would be nice to return `400 Bad Request` rather than `500 Internal Server Error` if `parse_qs()` fails, or the `url.query.decode('ascii')` above.

edb/server/protocol/auth_ext/__init__.py
fantix, line 1: Would be nice if the supported URLs are listed in the PR description :heart:

edb/server/protocol/protocol.pyx
```diff
@@ -523,6 +524,8 @@ cdef class HttpProtocol:
                 # Check if this is a request to a registered extension
                 if extname == 'edgeql':
                     extname = 'edgeql_http'
+                if extname == 'ext':
+                    extname = path_parts[3]
```
fantix, line 528: A comment is appreciated here as the logic can be confusing - we want to support both the old style API `/db/{dbname}/{ext_name}` and the new style API `/db/{dbname}/ext/{ext_name}` here I think?

Also the later `args = path_parts[3:]` is inconsistent, suggested:

```suggestion
                if extname == 'ext':
                    extname = path_parts[3]
                    args = path_parts[4:]
                else:
                    args = path_parts[3:]
```

So that you don't need to use `args[1:]`.

edb/server/protocol/auth_ext/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+async def handle_request(
+    request, response, db, extension_base_path: str, args: list[str]
+):
+    try:
+        # Set up routing to the appropriate handler
+        if args[0] == "authorize":
+            provider_name: str | None = _get_search_param(
+                request.url.query.decode("ascii"), "provider"
```
fantix, line 37: Question: why we chose API like `/db/{dbname}/ext/auth/authorize?provider=github` instead of `/db/{dbname}/ext/auth/github/authorize`? The same for the callback URL. The former is assuming we will always have `/authorize` and `/callback` for any providers.

edb/server/protocol/auth_ext/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+
+import http
+import json
+import urllib.parse
+
+from edb import errors
+from edb.common import debug
+from edb.common import markup
+
+from . import oauth
```
fantix, line 27: I would encapsulate this "OAuth" implementation detail into the provider, like having a `OAuthProvider` class handling the request. But at the same time, we don't need to optimize too early I guess, as far as the API is good.

edb/server/protocol/auth_ext/oauth.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+@overload
+def _maybe_get_config(db_config: SettingsMap, key: str) -> str | None:
+    ...
+
+
+def _maybe_get_config(
+    db_config: SettingsMap, key: str, expected_type: Type[object] = str
+) -> object:
```
fantix, line 127: Now that you defined a `base.py`, I think this method will not be a part of the `oauth.py` implementation - other providers would use config too.

edb/server/protocol/auth_ext/oauth.py
```diff
@@ -0,0 +1,155 @@
...
+from typing import Type, TypeVar, overload
+from jwcrypto import jwt, jwk
+
+from edb import errors
+from edb.server.config import ops
+
+from . import github
+
+
+# TODO get this from edb.server.config.ops directly?
```
fantix, line 34: nah :stuck_out_tongue:

```suggestion
```

edb/server/protocol/auth_ext/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+                raise errors.BackendError(
+                    "No provider specified in URL search parameters."
+                )
+
+            redirect_uri = f"{extension_base_path}/callback"
+
+            return oauth.redirect_to_auth_provider(
+                response=response,
+                redirect_uri=redirect_uri,
+                iss=request.host.decode(),
```
fantix, line 49: As this might be related to security, I don't think it's the best idea to blindly use the `Host` header as token issuer. I don't know but we probably want to have a separate optional config for it, and falls back to `Host/X-Forwarded-*`.

fantix (approved):
I'm okay with merging this first. Other than some nitpicking and a few refactor preferences, I only have a question (or 2) about the API (URL design and issuer config, pls see below).

edb/server/protocol/auth_ext/__init__.py
scotttrinh: Yeah, good call, I personally prefer to lean into inference when I can. I think I had this in place before I factored out the `_get_search_param` function. Will fix!

edb/server/protocol/auth_ext/__init__.py
scotttrinh: Is the suggestion to pass the bytes in here, and then raise a `BackendError` if any part of this fails, but still return `None` if the lookup fails?

edb/server/protocol/auth_ext/__init__.py
scotttrinh: Adding the provider to the search params feels lighter-weight and the callback itself doesn't need to refer to the provider since we don't want to get the provider name from the URL but rather from the `provider` claim within the state JWT. Doing it this way means that we avoid the case where we accidentally copy-pasta-fail the callback URL structure from GitHub to GitLab and don't notice that the URL is out of sync with the state variable. In that case we have to decide whether that's a fatal error, or just an accident we don't really care about.

edb/server/protocol/auth_ext/__init__.py
fantix: It's not quite a `BackendError` because it's actually the user's fault. I'd say we either need a new exception type (just for rendering a different response), or handle the particular exception types from `parse_qs()` (`ValueError`) and `decode()` (`UnicodeDecodeError`, which is also a `ValueError`) differently (if they are not expected from other code that should trigger internal server errors).

(sry, by "return" I meant overall HTTP response instead of returning from just this method)

edb/server/protocol/auth_ext/__init__.py
scotttrinh: Ah, so intead of `oauth.redirect_to_auth_provider(...)`, it'd be like:

```py
return provider.authorize(
  response=response,
  redirect_uri=redirect_uri,
  iss=iss,
  db_config=db_config
)
```

—?

That seems reasonable. It's been a while since I've worked a more OO-style so if that makes more sense or flows better from that perspective, I'll definitely incorporate that feedback. Maybe I can do that refactor as a separate PR.

edb/server/protocol/auth_ext/oauth.py
scotttrinh: Move this into `base.py`? Seems reasonable. I was thinking this would just hold the base class for just the provider, but if it's common practice to treat a module like this as a kind of utility module that holds some base functions and classes for the rest of the `auth_ext` module, that works for me!

edb/server/protocol/auth_ext/__init__.py
scotttrinh: Oh actually, this should be the server's netloc (e.g. `https://localhost:8080`). We should pass this into the handler I think, or maybe derive it from the `base_extension_path`.

edb/server/protocol/auth_ext/__init__.py
scotttrinh: Yeah, at the moment we're just throwing `BackendError` for any obvious issues, but I would love to have some error types that map to actual HTTP status codes that make sense, like `400` but I didn't see any existing examples of how we do this kind of error -> status matching. I'll look into that!

edb/server/protocol/auth_ext/__init__.py
scotttrinh: Well... It's not really "supported" yet since this endpoint will redirect into a flow that we can't currently finish. Should we still add a note in the description?

edb/server/protocol/auth_ext/__init__.py
scotttrinh: Also, I'm glad you brought this up because I need to remember to check the `iss` in the callback state check, too. 🙏

edb/server/protocol/protocol.pyx
scotttrinh: Would that be [`X-Forwarded-Host`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host) in this case?

edb/server/protocol/auth_ext/__init__.py
fantix: I see. I don't like long URLs either.

But I'm still a bit concerned about the API tbh. What if a future provider has a different `/callback` args format that we can't simply reuse the same handler method? If we will have `/callback/provider8`, I'd rather make it `/provider8/callback` now and tolerate the length of the URL. 🤔

Also I'm not sure why we have to sign the `provider` string into the JWT. If the attacker manipulated the provider, the final access token exchange will fail anyways, won't it? Also the provider is something configured inside the system, the attacker won't be able to redirect to a non-existing one.

Anyways, I think it's okay now that we only have OAuth 2.0 and stay with shorter `/authorize` and `/callback`, and let the future providers worry about their own URLs.

edb/server/protocol/auth_ext/__init__.py
fantix: Yeah, and we can probably save the effort e.g. passing `db_config` here and there by initializing the provider with necessary config loaded already. Separate PR is great, not something must to have too.

edb/server/protocol/protocol.pyx
fantix: Yup, as well as [`X-Forwarded-Proto`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto) for the original protocol.

Sorry about the previous misleading link.

edb/server/protocol/auth_ext/__init__.py
fantix: sry nvm 😅 was just a random rant trying to understand from the code what the actual URL we are using here.

edb/server/protocol/auth_ext/oauth.py
fantix: It's just a personal preference. It's just natural for me to think about the inheritance of `base` <- `oauth` <- `github`, where `base` has all the config stuff, `oauth` has the OAuth request handlers and `github` has only GitHub-specific stuff.

 It also makes total sense in Python to have flat function modules.

edb/server/protocol/auth_ext/__init__.py
scotttrinh: > What if a future provider has a different /callback args format that we can't simply reuse the same handler method?

We specify the callback URL in the `redirect_uri`, so we control this. And each provider has some slightly different (but very similar) set of query parameters that they will add to this URL or in some cases that data will be provided as form data in a POST request. Each `Provider` will know where to look for these values. But as far as these two URLs are concerned we control them entirely: client libraries will know to call the `/authorize?provider` URL in order to start the flow, and the server here specifies the callback URL.

> Also I'm not sure why we have to sign the provider string into the JWT. If the attacker manipulated the provider, the final access token exchange will fail anyways, won't it?

Indeed! But why have to trust the URL when the server itself can authoritatively make the claim that "this state string is intended to be used by this certain provider" and our callback can trust it due to the signature.

> Also the provider is something configured inside the system, the attacker won't be able to redirect to a non-existing one.

Sure this is less of a security issue (as far as I can tell) and more of a correctness issue: if we hardcode the provider into the callback URL itself, we're open to bugs where we have typos or copy-paste mistakes. If we get the provider from the state string, we at least know the value provided there is the same one that we built during the authorize endpoint.

I could see maybe:
- `/db/{db_name}/ext/auth/{provider_name}/authorize`
- `/db/{db_name}/ext/auth/callback`

But that feels surprising to me.

edb/server/protocol/protocol.pyx
scotttrinh: So, I'm thinking something like `def _get_netloc(request) -> str:` that does a bunch of lookups with fallbacks to the `request.url`, is that roughly the suggestion here?

edb/server/protocol/auth_ext/oauth.py
fantix: Technically there is no barrier.

`SettingsMap` is defined under `if TYPE_CHECKING` in `edb/server/config/ops.py` and not exposed in `edb/server/config/__init__.py`, so I assume it's not intended for being imported but rather a local shortcut. Strictly speaking it should be `Mapping[str, SettingValue]` for explicitness I think, so copying the shortcut complies with this explicitness.

edb/server/protocol/auth_ext/__init__.py
fantix: OK cool cool!

> And each provider has some slightly different (but very similar) set of query parameters

This is true in the scope of OAuth 2.0 / OpenID Connect. I'm okay with assuming by default `/ext/auth/authorize` and `/ext/auth/callback` are OAuth 2.0 client URLs.

> Each `Provider` will know where to look for these values.

We'll need "these values" first to know which `Provider` to use in case of `/callback` ;)

---

Oh actually that reminds me of another thing - will EdgeDB auth ext itself become a provider? I have impression we mentioned the Implicit Flow or PKCE stuff for user application login in the meetings, but I'm not sure if we decided on anything yet. I'm just thinking if there could be any URL/prefix conflict. But it shouldn't be a big problem.

edb/server/protocol/protocol.pyx
fantix: Yeah, that'll do! 🙏

edb/server/protocol/auth_ext/__init__.py
scotttrinh: > We'll need "these values" first to know which Provider to use in case of /callback ;)

Not as it's currently written: we get the `Provider` from the JWT claims in the `state` query parameter (or form data if it's a POST), and then evaluate the rest of the request with the `Provider`. True that it's specific to OAuth 2.0 but I think the current scope (no pun intended) is to get a single provider working (GitHub in this case) and add more as we go. If we need to support OAuth 1.0a for some reason (doesn't seem like there are any providers we should spend time on that are 1.0a only) I think it should be handled by a separate set of routes?

Maybe there are other auth schemes I haven't really explored much here (SAML maybe?) that makes this less compatible?

> Oh actually that reminds me of another thing - will EdgeDB auth ext itself become a provider? I have impression we mentioned the Implicit Flow or PKCE stuff for user application login in the meetings, but I'm not sure if we decided on anything yet. I'm just thinking if there could be any URL/prefix conflict. But it shouldn't be a big problem.

Yeah, I absolutely want to support username+password as a kind of self-hosted auth server that maybe skips the OAuth flow entirely but writes a similar `Profile` object to the database. I haven't really set about tackling this, but I think it should be the second provider we try to configure. FWIW, I don't think we need to consider this "provider" as a full-on OIDC style identity provider for any external services, so I suspect the actual implementation will just be `authorize?provider=email` (or maybe `password`) and then have the server just short-circuit the whole back-and-forth with itself and provide a signed OIDC identity on sucessful login. Sign up will probably be the big lift there, but I haven't thought that through yet.

---

Luckily, we have time to go in a few different directions to see if we end up with conflicts and I feel pretty free to change everything here at-will until we ship.

edb/server/protocol/protocol.pyx
scotttrinh: Alright, I went down a little rabbit hole about how to parse this and the related standardized `Forwarded` header, and I'm not sure we can just blanket trust these headers. It seems like servers tend to make trusting the proxy a configurable thing with whitelist IP ranges, and often need some tweaking.

Ultimately, if what we're trying to support here is custom domains or reverse proxies, I wonder if we should make this configurable? In that case, instead of trying to parse the headers ourselves, we have the user tell us what the callback_uri base ought to be and if that configuration value is unset (once we support that...) we fall back to the `request.url`. Thoughts?

scotttrinh: After some discussion, we decided that we'll need to address multiple cases where the `request.url` is not canonical, and each case will likely be addressed slightly differently. Will address as part of ongoing work on getting this ready for production.

edb/server/protocol/auth_ext/oauth.py
scotttrinh: Going to save this as a TODO for when refactoring of the OAuth and Provider classes.

edb/server/protocol/auth_ext/__init__.py
scotttrinh: Turns out, the issuer is really the full `base_extension_path` since a single server might be issuing tokens for multiple databases and we don't want to consider them all a single issuer.

⚡ [`55cceb8` (#5816)](https://github.com/edgedb/edgedb/pull/5816/commits/55cceb8b712498e5c4c439e843b3142afef266e9)

How does that look?

fantix: ahh good point, makes sense to me!

PR #5817, opened 2023-07-20
Document edgedb.toml

raddevon:


docs/cli/edgedb_toml.rst
```diff
@@ -0,0 +1,44 @@
...
+- ``server-version``- The server version of the EdgeDB project.
+
+  .. note::
+
+      The version specification is assumed to be **a minimum version**, but the
+      CLI will *not* upgrade to subsequent major versions. This means if the
+      version specified is ``3.1`` and versions 3.2 and 3.3 are available, 3.3
+      will be installed, even if version 4.0 is also available.
+
+      To specify an exact version, prepend with ``=`` like this: ``=3.1``
```
scotttrinh, line 25: I think we support the full range of Rust cargo-flavored semver specifiers? Do we want to point that out and link to some reference material like https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html?

docs/cli/edgedb_toml.rst
```diff
@@ -0,0 +1,44 @@
+.. _ref_cli_edgedb_toml:
+
+===========
+edgedb.toml
+===========
+
+The ``edgedb.toml`` file is a key component of EdgeDB's project system. If this
```
scotttrinh, line 7: Do you think it's helpful to link out to any TOML format documentation like https://toml.io/en/v1.0.0 ?

docs/cli/edgedb_toml.rst
```diff
@@ -0,0 +1,44 @@
...
+file is present in a directory, it signals to the CLI that the directory is an
```
fantix, line 8: > it signals to the CLI

also to any client bindings ;)

scotttrinh: Oh, good point: I wonder if this should live outside of the CLI docs? Or at least also live somewhere about the general project structure of an EdgeDB project?

raddevon: I almost put it in the reference section instead, but I thought only the CLI cared about it. If client bindings also use it, I think it makes sense to put it there instead. Thanks, @fantix!

elprans (approved):

docs/cli/edgedb_toml.rst
raddevon: I found a note with a link to TOML docs in the projects reference, so I just copied that one over. I did like your target better, so I changed it from targeting the TOML quickstart to what you suggested.

PR #5818, opened 2023-07-20
Update syntax test.

vpetrovykh:
Clearly our syntax error detection and recovery has changed the interpretation of `count(foo 1)` as probably missing a comma and not missing another parenthesis.

In practice (in REPL) it seems that `foo 1` is interpreted as a call and results in different error if function `foo` is not found.

### DIFF
--- a/edb/edgeql-parser/src/parser.rs
+++ b/edb/edgeql-parser/src/parser.rs
@@ -511,8 +511,17 @@ impl Spec {
         let actions = v
             .actions
             .into_iter()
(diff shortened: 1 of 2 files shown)


vpetrovykh:
Hmmm... this might be non-deterministic error as it's different from my local tests.

aljazerzen:
Ah, I've had this problem before and I though I solved it. I can go over parser code again and try to find the source of this.

vpetrovykh:
No longer reproduces locally for me either.

PR #5824, opened 2023-07-21
Fix tests and address linting/typing issues

scotttrinh:
I didn't want to add a bunch of noise to the base PR, so I moved this effeort into a separate PR. Should be pretty uncontroversial, but would love at least a quick check from someone.

### DIFF
--- a/edb/testbase/http.py
+++ b/edb/testbase/http.py
@@ -37,7 +37,6 @@ bag = assert_data_shape.bag
 
 
 class BaseHttpExtensionTest(server.QueryTestCase):
-
     @classmethod
     def get_extension_name(cls):
         raise NotImplementedError
@@ -69,8 +68,8 @@ class BaseHttpExtensionTest(server.QueryTestCase):
         )
         super().tearDownClass()
 
-class ExtAuthTestCase(BaseHttpExtensionTest):
 
+class ExtAuthTestCase(BaseHttpExtensionTest):
     @classmethod
     def get_extension_name(cls):
         return 'auth'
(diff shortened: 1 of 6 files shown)


tests/test_http_auth.py
```diff
@@ -38,7 +38,7 @@ def _scram_auth(self, user, password):
             _, headers, status = self.http_con_request(con, {}, path="token")
             self.assertEqual(status, 401)
             self.assertEqual(
-                headers, headers | {"www-authenticate": "scram-sha-256"}
+                headers, headers | {"www-authenticate": "SCRAM-SHA-256"}
```
scotttrinh, line 41: Here is the one place where we were depending on the lowercasing of the header values in the testbase class.

1st1:
This should probably be against the `master` branch. And you can later rebase `auth_ext` on top of the updated `master`

edb/testbase/http.py
```diff
@@ -69,6 +69,7 @@ def tearDownClass(cls):
         )
         super().tearDownClass()
 
+
```
1st1, line 72: This one isn't necessary?

scotttrinh: It was required by the linter:

```
==================================================================================================
FAIL: test_cqa_flake8 (test_sourcecode.TestCodeQuality)
--------------------------------------------------------------------------------------------------
Py_HashSecret: 000000000000000000000000000000000000000000000000
random.seed(): 39c18bca6ae5c9a5
--------------------------------------------------------------------------------------------------
  File "/Users/scotttrinh/edgedb/edgedb/tests/test_sourcecode.py", line 77, in test_cqa_flake8
    raise AssertionError(
AssertionError: flake8 validation failed:
./testbase/http.py:72:1: E302 expected 2 blank lines, found 1
```

scotttrinh:
> This should probably be against the master branch. And you can later rebase auth_ext on top of the updated master

There is only one change that actually exists on `master` which is the SCRAM change and it requires the update to the testbase, which I can definitely move.

PR #5830, opened 2023-07-24
[1/4] Extract Tenant from Server

fantix:


edb/server/protocol/binary.pyx
```diff
@@ -151,13 +151,14 @@ cdef class EdgeConnection(frontend.FrontendConnection):
     def __init__(
         self,
         server,
+        tenant,
```
elprans, line 154: Why pass both `server` and `tenant` and not just `tenant` instead?

elprans (approved):

edb/server/protocol/binary.pyx
fantix: Mostly because the Postgres protocol may not have a tenant already (TLS handshake is inside the Postgres protocol).

PR #5831, opened 2023-07-24
Add support for query params + JSON output to notebook protocol

jaclarke:


msullivan (commented):
Sure, looks reasonable.

Sorry I didn't notice this when I got back from vacation. What's the use case for this?

msullivan (approved):

PR #5834, opened 2023-07-24
Do some fairly low-quality typing improvements to server.py

msullivan:
Make it check with check-untyped-defs, and annotate some high-risk
code paths. This isn't that much of an improvement, but will help with
a pending refactor. (Though probably not help so much that it saved
time on net.)

### DIFF
--- a/edb/server/server.py
+++ b/edb/server/server.py
@@ -1,3 +1,5 @@
+# mypy: check-untyped-defs
+
 #
 # This source file is part of the EdgeDB open source project.
 #
@@ -77,6 +79,8 @@ if TYPE_CHECKING:
     import asyncio.base_events
     import pathlib
 
+    from . import bootstrap
+
 
 ADMIN_PLACEHOLDER = "<edgedb:admin>"
 logger = logging.getLogger('edb.server')
@@ -95,7 +99,7 @@ class StartupError(Exception):
 
 class Server(ha_base.ClusterProtocol):
 
-    _sys_pgcon: Optional[pgcon.PGConnection]
+    __sys_pgcon: pgcon.PGConnection | Any
 
     _roles: Mapping[str, RoleDescriptor]
     _instance_data: Mapping[str, str]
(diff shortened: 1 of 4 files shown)


edb/server/server.py
```diff
@@ -934,18 +944,20 @@ async def _introspect_dbs(self):
                 # information about them.
                 g.create_task(self._early_introspect_db(dbname))
 
-    async def get_patch_count(self, conn):
+    async def get_patch_count(self, conn: pgcon.PGConnection) -> int:
         """Get the number of applied patches."""
-        num_patches = await conn.sql_fetch_val(
+        num_patches_b = await conn.sql_fetch_val(
```
scotttrinh, line 949: I assume this change is to avoid reassigning `num_patches` with a different type, first `str` and then `int`? If so is there an advantage to the intermediate `num_patches` variable rather than returning the expression?

edb/server/server.py
```diff
@@ -934,18 +944,20 @@ async def _introspect_dbs(self):
...
             ''',
         )
-        num_patches = json.loads(num_patches) if num_patches else 0
+        num_patches = json.loads(num_patches_b) if num_patches_b else 0
         return num_patches
 
-    async def _get_patch_log(self, conn, idx):
+    async def _get_patch_log(
+        self, conn: pgcon.PGConnection, idx: int
+    ) -> Optional[bootstrap.PatchEntry]:
```
scotttrinh, line 960: We prefer `Optional[T]` over `T | None`, is that right?

scotttrinh (approved):
Awesome, thanks so much for doing this! This is helping me get a sense of how we like to hint types, which I will bring over to my work on the Auth extension. I have a few questions just to help me understand our style.

edb/server/server.py
msullivan: *I* prefer `Optional[T]`, at least, and most of the code base predates being able to write `T | None` anyway

edb/server/server.py
msullivan: Yeah, that's right.

I suspect that the reason it is assigned to an intermediate variable instead of returning it directly is because at some point while I was developing it, something was going wrong and I wanted it split up so I could stick a print or a breakpoint in between, and then I never merged them.

scotttrinh: Awesome, yeah I have a similar habit especially when developing. Do we prefer this to enable easy debugging or is there no official preference for assignment before returning?

msullivan: No official preference.

I think that constructing stuff so that it is easy to debug on the *macro* scale, but on micro stuff like this, it's easy enough to make the changes on the fly, and code gets read more often than it gets instrumented. I think it's pretty marginal either way. I'm not sure if I really have a personal preference on which version ought to get committed.

PR #5837, opened 2023-07-25
Fix some breakages in the patch system

msullivan:


aljazerzen (approved):
To be honest, I didn't figure out what does this do.

edb/server/bootstrap.py
```diff
@@ -836,6 +836,8 @@ def prepare_patch(
             configspec=config.spec_to_json(config_spec).encode('utf-8'),
         ))
 
+        support_view_commands.generate(subblock)
```
msullivan, line 839: During a refactor I accidentally deleted this line that actually recreates the views

PR #5845, opened 2023-07-26
EdgeDB + SQL equivalent clarification

Dhghomon:
Just a small clarification that the EdgeDB equivalents to primary key and foreign key are the id property and links. The current "these fields are automatically indexed" wording also gives the impression that primary and foreign key are EdgeDB terms.

Context: Doing some writing on indexing and have never used SQL before so wasn't sure what a primary and foreign key was and had to do a bit of searching. There are probably others out there that aren't immediately familiar with SQL.

### DIFF
--- a/docs/datamodel/indexes.rst
+++ b/docs/datamodel/indexes.rst
@@ -184,11 +184,11 @@ Indexes can be augmented with annotations.
   **Foreign and primary keys**
 
   In SQL databases, indexes are commonly used to index *primary keys* and
-  *foreign keys*. The EdgeDB equivalents of these are the ``id`` property
-  and links; these are automatically indexed and there is no need to manually
-  declare them. Moreover, any property with an :eql:constraint:`exclusive`
-  constraint is also automatically indexed.
-
+  *foreign keys*. EdgeDB's analog to SQL's primary key is the ``id`` field
+  that gets automatically created for each object, while a link in EdgeDB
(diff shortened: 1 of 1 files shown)


docs/datamodel/indexes.rst
```diff
@@ -184,9 +184,10 @@ Indexes can be augmented with annotations.
   **Foreign and primary keys**
 
   In SQL databases, indexes are commonly used to index *primary keys* and
-  *foreign keys*. In EdgeDB, these fields are automatically indexed; there's no
-  need to manually declare them. Moreover, any property with an
-  :eql:constraint:`exclusive` constraint is also automatically indexed.
+  *foreign keys*. The EdgeDB equivalents of these are the ``id`` property
+  and links; these are automatically indexed and there is no need to manually
```
raddevon, line 188: Maybe add "respectively" here just to make it extra clear that primiary key -> EdgeDB `id` and foreign key -> EdgeDB link.

```rst
…are the ``id`` property and links respectively.
```

or even just spell it all out.

```rst
EdgeDB's analog to SQL's primary key is the `id` field that gets automatically
created for each object, and the analog to SQL's foreign key is the link. These
are automatically indexed…
```

raddevon (approved):
One minor thing. Otherwise, looks good!

docs/datamodel/indexes.rst
Dhghomon: That spelled-out wording looks nice, have changed to that.

PR #5847, opened 2023-07-26
Doc clarify deletion policy if orphan

raddevon:
This is an attempt to clear up confusion around `if orphan` in deletion policies, prompted by a question on Discord: https://discord.com/channels/841451783728529451/849377751370432573/1133652031635214366

The note was intended to explain that `if orphan` on a link does not refer to an object being globally orphaned but only with respect to the link containing the deletion policy. In trying to explain this, the note ignored the fact that deletion might also be restricted by `on target` deletion policies of other links linking to the same object, including the default target deletion policy of `restrict`.

This change caveats the explanation to include the complicating factor of `on target` deletion policies.

### DIFF
--- a/docs/datamodel/links.rst
+++ b/docs/datamodel/links.rst
@@ -49,8 +49,8 @@ link.
     }
 
 On the other hand, backlinks work in reverse to find objects that link to the
-object, and thus assume  ``multi`` as a default. Use the ``single`` keyword 
-to declare a "to-one" backlink.
+object, and thus assume ``multi`` as a default. Use the ``single`` keyword to
+declare a "to-one" backlink.
 
 .. code-block:: sdl
 
--- a/docs/datamodel/links.rst
+++ b/docs/datamodel/links.rst
@@ -553,23 +553,32 @@ deletion policy.
 
 .. note::
 
+    The ``if orphan`` qualifier does not apply globally across all links in the
+    database or across any other links even if they're from the same type.
     Deletion policies using ``if orphan`` will result in the target being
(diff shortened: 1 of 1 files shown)


scotttrinh (approved):
Seems like a good update for clarity! In addition, I think we need to add some visualizations to truly make this understandable when reading it casually.

Dhghomon:
> In addition, I think we need to add some visualizations to truly make this understandable when reading it casually.

Yeah, I was thinking the same thing when reading through this and then noticed your comment. Maybe it could be done by adding the mentioned Channel type with a diff adding `on target delete allow` to the first schema using `on source delete delete target`. Something like this perhaps:

```type Channel {
  title: str;
  multi messages: Message {
   # Without "allow", any Channel objects linking to
   # Messages will prevent MessageThread's deletion
   # policy from automatically deleting Message objects
   + on target delete allow;
  }
}```

And then re-emphasize in point 2 that `restrict` is the default:

- its deletion is restricted by another link's on target delete policy (which defaults to `restrict` unless specified otherwise)

raddevon:
I like the feedback, @scotttrinh @Dhghomon. Do these changes help, or do they just make it worse? 😅

I showed the case that would allow for the deletion first since the point I'm trying to make here is that `if orphan` is not global, but then I circled back to the default example omitting the target policy on the other linking type to show the other case that would also prevent `Message` deletion.

Dhghomon:
Yeah, even just thinking about a deletion policy itself regardless of language is a mental exercise in itself so this is about as good as it gets I think (until an image is added later on). Nice work!

PR #5854, opened 2023-07-26
Update index.rst

avedetvedea:
both camel-case is wrong, because it should have been camelCase, and I think even that's wrong, because the convention actually is PascalCase


msullivan:
I agree that writing "camel-case" is wrong but "PascalCase" feels too obscure to me

PR #5855, opened 2023-07-26
Refactor OAuth into the base provider class

scotttrinh:
Took a swing at incorporating the feedback from here:

https://github.com/edgedb/edgedb/pull/5816#discussion_r1273944870
> I would encapsulate this "OAuth" implementation detail into the provider, like having a `OAuthProvider` class handling the request.

Wanted to get some feedback as to whether I understood the suggestion. I'll make some notes and questions inline.


edb/server/protocol/auth_ext/base.py
```diff
@@ -16,18 +16,129 @@
...
+    ):
+        state_token = self._make_signed_token(iss=issuer, provider=self.name)
+
+        auth_url = self._get_code_url(
+            state=state_token.serialize(), redirect_uri=redirect_uri
+        )
+
+        response.status = http.HTTPStatus.FOUND
+        response.custom_headers["Location"] = auth_url
+        response.close_connection = True
```
scotttrinh, line 57: This _seems_ generic enough that I moved it into the base class, but I'd be curious to know if it doesn't seem right. I think this roughly matches the separation we had with the separate OAuth module where the `Provider` knows how to build the auth-provider-specific redirect URL, but the OAuth logic is generalized.

edb/server/protocol/auth_ext/github.py
```diff
@@ -17,29 +17,28 @@
...
+        super().__init__("github", *args[1:], **kwargs)
 
-    def get_code_url(
-        self, state: str, redirect_uri: str, scope: str = "read:user"
-    ) -> str:
+    def _get_code_url(self, state: str, redirect_uri: str) -> str:
         params = {
             "client_id": self.client_id,
-            "scope": scope,
+            "scope": "read:user",
```
scotttrinh, line 33: I don't think the caller should ever need to customize this scope behavior: it feels firmly like something the Provider should know: how to get to the user info.

edb/server/protocol/auth_ext/providers.py
```diff
@@ -0,0 +1,39 @@
...
+    "github": github.GitHubProvider,
+}
+
+
+def make(db_config: SettingsMap, name: str):
+    provider_class = _providers.get(name)
+    if provider_class is None:
+        raise errors.BackendError(f"Unknown provider: {name}")
+
+    return provider_class(name, db_config)
```
scotttrinh, line 39: I wasn't sure where to put this logic that instantiates classes based on a map from strings to provider classes. It's lightweight enough to live in `__init__.py`, but it felt weird to import the `github` module in `__init__.py` for some reason 🤔 Any suggestions?

edb/server/protocol/auth_ext/base.py
```diff
@@ -16,18 +16,129 @@
...
 
-    def get_code_url(self, state: str, redirect_uri: str, scope: str) -> str:
+    async def complete(self, request, response):
```
scotttrinh, line 59: Open to suggestions on naming here: This will be what the `/callback` route calls and it will do the state check, code exchange, get the user info with the token, save a profile to the database, and redirect back to the application.

We can hash out the name in a further PR but thought it was worth getting some input on it now.

edb/server/protocol/auth_ext/providers.py
fantix: > it felt weird to import the `github` module in `__init__.py` for some reason

I don't know that, but it's a common practice to do such import if you need to expose symbols from submodules onto package level.

I would import `github` within the if/case branches lazily to reduce memory usage.

scotttrinh: Yeah, I actually ended up using a `match` here so I can do it lazily. Does it still make sense to have a kind of separate module that is used as a dynamic constructor of another class based on some runtime information like this?

scotttrinh:
Closing in favor of keeping the Provider separate from the Authorization Flow mechanism (OAuth in this case).

PR #5856, opened 2023-07-26
Refactor loading of parsers

elprans:
We have recently switched to a new parser in Rust.  This parser does not
need the heavy `parsing` Spec instances to parse things as it translates
them into its own internal representation.  Hence, keeping the `Spec`
instances around is pure memory waste.  Additionally, the current parser
loading abstractions are unnecessarily complex and provide for
customization that we don't actually use.

This rectifies both issues by simplifying the whole thing to
`parse(grammar_module: types.ModuleType, source: str)`.  The Rust parser
still loads and caches its internal spec and there is still `preload()`,
so there should be no behavioral changes.

Finally, reduce the list of parsers loaded by the I/O process to just
`fragment`, which is used for the processing of config spec.  There are
no other uses.

This brings the resident set of the I/O process down from 307MiB to
160MiB, a nearly 50% reduction!  Compiler processes get a more modest
reduction from 251MiB to 196MiB (22%), however this figure slightly
favors the old implementation as it currently forgets to preload the
`migration_body` and `extension_package_body` grammars.


edb/edgeql/parser/parser.py
```diff
@@ -1,193 +0,0 @@
...
-class EdgeQLParserSpec(parsing.ParserSpec):
-    def get_parser(self):
-        return EdgeQLParser(self)
-
-
-class EdgeQLSingleSpec(EdgeQLParserSpec):
-    def get_parser_spec_module(self):
-        from .grammar import single
-
-        return single
```
aljazerzen, line 40: I missed an unnecessary class, that I could have removed in favor of a single module reference!

aljazerzen (approved):
This is great! It took me about to understand how spec are loaded and cached in the rust parser, but I makes sense.

Nice to see a commit that reduces memory footprint by 50% **and** has removes 100 lines of code :D

aljazerzen:
Do you think that we could take this further and generate parser spec before the server / compiler is even started?

One option is from `build.rs` script and include the spec in the Rust binary directly.

Another is an edb tool that you'd call every time the grammar is changed and it would call the parser that would pull in the grammar and store is as a blob in the repo. That would then have to be committed and distributed along the parser binary.

elprans:
> One option is from `build.rs` script and include the spec in the Rust binary directly.

Sure.  Though I'd keep the hot-reload feature in devmode for the sake of grammar iteration.

PR #5857, opened 2023-07-27
Pull more initial state out of bootstrap compiler state

elprans:
Continuing on Sully's work in #5836, avoid loading schemas and various
other internal paraphernalia twice, which saves ~20MiB of memory in the
I/O process.


msullivan (approved):
Oh, unfortunately I think this means the 20 MiB was a regression in https://github.com/edgedb/edgedb/pull/5836, so this isn't a memory improvement relative to 3.x.
(Though it will be a server startup time improvement)

fantix (approved):

PR #5858, opened 2023-07-27
+ Recent changes relevant to indexes

Dhghomon:
I confirmed yesterday that indexing has a beneficial effect on grouping as well, which was added to EdgeDB recently(-ish) but this hasn't been reflected in the documentation yet. This is the only page I could find that needed this change but there could be others out there.

The changes are:

- Note that indexing speeds up queries for grouping in addition to the existing filtering and order by,
- Remove the "by a database" part as probably not all databases in general use grouping (plus no need to refer to about databases in general on this page IMO)
- Adds a note to give the analyze keyword a try which I notice shows the difference quite clearly.


docs/datamodel/indexes.rst
```diff
@@ -4,17 +4,17 @@
...
+type declarations and reference a particular property; this will speed up
+any query that references that property in a ``filter``, ``order by``, or
+``group`` clause.
 
 .. note::
 
   While improving query performance, indexes also increase disk and memory
   usage and slow down insertions and updates. Creating too many indexes may be
-  detrimental; only index properties you often filter or order by.
+  detrimental; only index properties you often filter, ordero or group by.
```
scotttrinh, line 17:
```suggestion
  detrimental; only index properties you often filter, order, or group by.
```

scotttrinh (commented):
Didn't mean to jump in here, but I was interested in this topic when you brought it up. I'll leave my review as just a comment since @raddevon is the real reviewer here, but it looks great to me.

docs/datamodel/indexes.rst
```diff
@@ -37,9 +37,13 @@ notation shorthand <ref_dot_notation>`: ``.name``.
       index on (.name);
     }
 
-By indexing on ``User.name``, queries that filter by the ``name`` property will
-be faster, as the database can look up a name in the index instead of scanning
-through all Users sequentially.
+By indexing on ``User.name``, queries that filter, order, or group by the
```
scotttrinh, line 40: Oh, I just noticed that we say "order" here, but "sort" in the other two places. I kinda like "order" since that's the related keyword, even though "sort" is probably the better verb. At any rate, just being consistent is all I really wanted to call out here, so whichever direction makes the most sense to y'all is good with me.

docs/datamodel/indexes.rst
```diff
@@ -37,9 +37,13 @@ notation shorthand <ref_dot_notation>`: ``.name``.
       index on (.name);
     }
 
-By indexing on ``User.name``, queries that filter by the ``name`` property will
-be faster, as the database can look up a name in the index instead of scanning
-through all Users sequentially.
+By indexing on ``User.name``, queries that filter, order, or group by the 
+``name`` property will be faster, as the database can look up a name in 
+the index instead of scanning through all Users sequentially.
```
raddevon, line 42:
```suggestion
the index instead of scanning through all ``User`` objects sequentially.
```

docs/datamodel/indexes.rst
```diff
@@ -4,17 +4,17 @@
 Indexes
 =======
 
-An index is a data structure used internally by a database to speed up
-filtering and sorting operations. Most commonly, indexes are declared within
-object type declarations and reference a particular property; this will speed
-up any query that references that property in a ``filter`` or ``order by``
-clause.
+An index is a data structure used internally to speed up filtering, sorting,
```
raddevon, line 7:
```suggestion
An index is a data structure used internally to speed up filtering, ordering,
```

docs/datamodel/indexes.rst
```diff
@@ -73,8 +77,8 @@ Index on multiple properties
 ----------------------------
 
 A *composite index* is an index that references multiple properties. This will
-speed up queries that filter or sort on *both properties*. In EdgeDB, this is
-accomplished by indexing on a ``tuple`` of properties.
+speed up queries that filter, sort, or group on *both properties*. In EdgeDB,
```
raddevon, line 80:
```suggestion
speed up queries that filter, order, or group on *both properties*. In EdgeDB,
```

raddevon (approved):
Looks good! I agree with Scott's comment on sort vs. order, and I went ahead and added suggestions for those changes, along with one other minor thing.

Thanks, @Dhghomon!

Issue #5697, opened 2023-06-26
Databases dumped via SQL protocol fail to restore on Postgres 15


elprans:
See test failures in #5590

PR #5860, opened 2023-07-27
fix: SQL name resolving

aljazerzen:
A result of tedious pg_dump debugging.
Tests on Postgres 15 were failing becuase they are using pg_dump 15,
which was producing the following statements that were
not produced by pg_dump 14:

```sql
CREATE TYPE edgedb."__SchemaAccessPolicy" (
    INTERNALLENGTH = variable,
    INPUT = array_in,
    OUTPUT = array_out,
    RECEIVE = array_recv,
    SEND = array_send,
    ANALYZE = array_typanalyze,
    ELEMENT = edgedb."_SchemaAccessPolicy",
    CATEGORY = 'A',
    ALIGNMENT = double,
    STORAGE = extended
);
```

Applying these dumps failed because type
edgedb."_SchemaAccessPolicy" did not exist.

After digging in pg_dump source code a bunch, trying to figure out
why these statements were showing up, I found a query similar to this:

```sql
select
  typname,
  (
    SELECT typarray FROM pg_type te WHERE oid = pg_type.typelem
  ) = oid AS isarray
from pg_type;
```

After inspecting the results, I found that our resolver does not match
`pg_type.typelem` to the outer `pg_type` (as it should), but to the inner
`pg_type`, even though the inner `pg_type` has an alias which should make
the orignal name inaccessible.

So that's the bug. With an easy fix.
Turns out that when I implemented the resolver I did not know this detail,
becuase we even had a test for it.

### DIFF
--- a/edb/pgsql/resolver/expr.py
+++ b/edb/pgsql/resolver/expr.py
@@ -201,7 +201,8 @@ def _lookup_in_table(
 def _lookup_table(tab_name: str, ctx: Context) -> context.Table:
     matched_tables: List[context.Table] = []
     for t in ctx.scope.tables:
-        if t.name == tab_name or t.alias == tab_name:
+        t_name = t.alias or t.name
+        if t_name == tab_name:
             matched_tables.append(t)
 
     if not matched_tables:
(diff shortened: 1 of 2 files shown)


aljazerzen:
Ref #5697

elprans (approved):

msullivan:
Nice detective work here

PR #5864, opened 2023-07-27
Change how globals are passed in GraphQL queries.

### LINKED ISSUE #5863: Change how globals are passed in GraphQL
Currently global variables are passed as their own special `globals` field. This is a non-standard field and it gets in the way of existing frameworks and libraries.

An alternative to that would be to use the `variables` field that's part of the GraphQL standard and add `__globals__` to that.

vpetrovykh:
Instead of using a separate `globals` field, use `variables` to add a `__globals__` object containing the global variables.

Fixes #5863

### DIFF
--- a/edb/graphql/extension.pyx
+++ b/edb/graphql/extension.pyx
@@ -94,6 +94,7 @@ async def handle_request(
     operation_name = None
     variables = None
     globals = None
+    deprecated_globals = None
     query = None
 
     try:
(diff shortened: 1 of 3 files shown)


msullivan (approved):

vpetrovykh:
I've added tests explicitly trying the scenario where new and deprecated globals don't match, but one of them is an empty dict to make sure that it's still an error.

msullivan:
Let's merge this for 4.x

PR #5869, opened 2023-07-28
[3/4] Multi-tenant Compiler

fantix:
This PR reuses the remote compiler worker code, but intentionally keeps the remote compiler server code untouched for future drop.


edb/server/compiler_pool/pool.py
```diff
@@ -1177,11 +1203,316 @@ async def _compute_compile_preargs(self, *args):
...
+            "compile_graphql",
+            "compile_sql",
+        }:
+            idx = args.index(METHOD_NAME_PLACEHOLDER)
+            args = (
+                *args[:idx],
+                method_name,
+                *args[idx + 1:],
+            )
+            method_name = "call_for_client"
```
fantix, line 1307: This can be refactored into `_compute_compile_preargs()` I think

elprans (approved):

PR #5873, opened 2023-07-28
Refactor Auth extension router structure

scotttrinh:
Took some parts of the proposed refactors and came up with the following responsibilities, and stuck with a flatter class hierarchy. The shape is kinda similar, but it pulls in some object-orientation in places that were more strictly functional before.

## Classes

- `http.Router`: Responsible for interacting with the HTTP data itself. Responsible for things like query params, catching domain errors and mapping them to appropriate HTTP responses, returning the response, and any other HTTP-specific thing we come across.
- `oauth.Client`: Responsible for knowing everything that is strictly OAuth2 based: authorization code URL building, and eventually code exchange and querying for user info. It also knows how to map from a `str` to a `Provider`. (TODO: use IDs instead of names somehow)
- `BaseProvider`: abstract class that defines the sorts of things that `Provider`s need to know how to do to fulfill the flow. Will specify things like where the authorization URL is vs the API endpoint for getting user info, etc. This might get specialized into `OAuth2Provider`, but there is such a thin layer for now between the `oauth.Client` and the final concrete `Provider`.
- `GitHubProvider`: The actual specification of how to OAuth2 Authorization Code Flow for GitHub.

## Notes

The `oauth.Client` and `BaseProvider` might move closer together or instead become more specialized, I'm curious to see how that works out.


edb/server/protocol/auth_ext/http.py
```diff
@@ -0,0 +1,169 @@
...
+from edb.common import markup
+
+from . import errors
+from . import util
+
+from edb.server.protocol.protocol import HttpRequest, HttpResponse
+
+
+class Router:
+    def __init__(self, db: Any, base_path: str):
```
scotttrinh, line 40: For some reason, I was getting runtime errors when I was trying to use the imported `Database` from `dbview`...

```
        > class Client:
    /Users/scotttrinh/edgedb/edgedb/edb/server/protocol/auth_ext/oauth.py, line 29, in Client
        > def __init__(self, db: Database, provider: str):

NameError: name 'Database' is not defined
```

scotttrinh: I had a `dbview.pyi` file initially that I was defining the `Database` class with in detail since the Cython classes are not picked up by my tooling (seems like this is a "by design" thing based on reading a bit), which I think was throwing this off. It works at runtime if I remove the `pyi` typing stub and just import it directly, but then the type is basically `Unknown` which isn't very helpful.

Any advice? I'm not going to get stuck on this, but I'm hoping there is a way to get useful typing here.

PR #5876, opened 2023-07-31
Analyze cli update

Dhghomon:
This PR has two small changes:

1) A note that analyze is available in the REPL and UI as well. This is the documentation for the CLI but it's also the first search result via Google as well as pretty much only search result when searching through the website so IMO it bears a certain responsibility that goes beyond just documenting the CLI. A quick link to this video https://www.youtube.com/watch?v=WoHJu0nq5z0 inside the note might not hurt either.
2) Updates the input to what is actually produced now.


docs/cli/edgedb_analyze.rst
```diff
@@ -9,23 +9,31 @@ edgedb analyze
 
     This CLI feature is compatible with EdgeDB server 3.0 and above.
 
+.. note::
+
+    This feature is also available inside the EdgeDB REPL and UI in which
```
scotttrinh, line 14: Yes, and!

I think we should also maybe add a section somewhere on https://www.edgedb.com/docs/cli/edgedb to talk about this feature and update this to be a link to that documentation.

Dhghomon:
Oh yeah, might as well put that in the same PR. I'll add that part tomorrow
then.

2023년 7월 31일 (월) 오후 9:35, Scott Trinh ***@***.***>님이 작성:

docs/cli/edgedb_analyze.rst
Dhghomon: I see the page does mention the command but makes it look like it's backslash only so changed that. Also added the video plus another while I was at it, and a quick something else I noticed where a classic Windows command wasn't specified.

scotttrinh: What do you think about making the text "EdgeDB REPL" a link to the appropriate place in the REPL docs?

raddevon: I would like to have documentation of the UI, but so far, the most we have is the documentation of the `edgedb ui` command which doesn't really tell you anything about the UI except how to invoke it.

We have also made a distinction between the CLI REPL and the UI REPL, and we might want to keep with that here. `analyze` works across the CLI REPL, the UI REPL, and the UI's query editor.

docs/cli/edgedb_analyze.rst
```diff
@@ -9,23 +9,31 @@ edgedb analyze
...
+    prefacing it with ``edgedb`` is not required.
```
raddevon, line 15: Here's a suggestion that I believe will incorporate both my notes and @scotttrinh's notes from the previous comment thread.
```suggestion
    Performance analysis is also available in our :ref:`CLI REPL
    <ref_cli_edgedb>` and the UI's REPL and query builder (both accessible by
    running :ref:`ref_cli_edgedb_ui` to invoke your instance's UI). Use it by
    prepending your query with ``analyze``.
```

raddevon (approved):
This is good stuff, @Dhghomon. I left a couple of suggestions (open for further discussion if you don't like my solutions). Otherwise, this looks ready to go!

docs/cli/edgedb.rst
```diff
@@ -20,7 +20,12 @@ print all the responses into the standard output:
 
     cat myscript.edgeql | edgedb [<connection-option>...]
 
+This command also works on PowerShell in Windows, while the classic
+Command Prompt uses a different keyword:
```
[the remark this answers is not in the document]
Dhghomon: Yes, I think you're right!

PR #5880, opened 2023-08-01
Add no-op GitHub OAuth callback handler

scotttrinh:
This adds the necessary requests to complete the authorization code flow for GitHub as well as request user data and emails for the authenticated user. It purposefully does not yet do anything with that data (going to do that as a separate PR).

### DIFF
--- a/tests/test_http_ext_auth.py
+++ b/tests/test_http_ext_auth.py
@@ -58,6 +58,7 @@ class MockAuthProvider:
             tuple[str, str, str],
             tuple[dict[str, Any] | list[dict[str, Any]], int],
         ] = {}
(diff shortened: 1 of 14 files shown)


tests/test_http_ext_auth.py
```diff
@@ -17,16 +17,124 @@
...
+                int(handler.headers['Content-Length'])
+            ).decode()
+            if 'Content-Length' in handler.headers
+            else None,
+        }
+        self.requests[key].append(request_details)
+
+        if key not in self.routes:
+            handler.send_error(404)
+            return
```
scotttrinh, line 101: I'm thinking maybe we default to raising an Exception here?

PR #5885, opened 2023-08-02
Add tuple alias example

Dhghomon:
The documentation on aliases currently has examples for scalars, objects, and expressions, but none on tuples. Being able to make a tuple might seem to be a bit of a no-brainer but IMO it's good to explicitly mention and give an example as an alias that stands sort of between scalars and objects: has multpile values and works like an object but doesn't need an insert so it's a pointer to a set of values instead of objects in a database.

### DIFF
--- a/docs/datamodel/aliases.rst
+++ b/docs/datamodel/aliases.rst
@@ -62,7 +62,7 @@ properties or links.
 In effect, this creates a *virtual subtype* of the base type, which can be
 referenced in queries just like any other type.
 
-**Query alias**
+**Expression alias**
 
 Aliases can correspond to an arbitrary EdgeQL expression, including entire
 queries.
@@ -70,8 +70,10 @@ queries.
 .. code-block:: sdl
     :version-lt: 3.0
 
+    # Expression producing a tuple alias
     alias Color := ("Purple", 128, 0, 128);
 
+    # Expression producing a named tuple alias
     alias GameInfo := (
       name := "Li Europan Lingues",
       country := "Iceland",
@@ -87,6 +89,7 @@ queries.
       required property is_published -> bool;
     }
 
+    # Alias produced from a full query
     alias PublishedPosts := (
       select BlogPost
       filter .is_published = true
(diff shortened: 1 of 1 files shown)


docs/datamodel/aliases.rst
```diff
@@ -33,6 +33,27 @@ of all User objects*. After declaring the alias below, you can use ``User`` and
 
   alias UserAlias := User;
 
+**Tuple alias**
+
+A tuple alias has features in common with both scalar and object type aliases.
+Like object type aliases, a tuple alias can hold multiple values that can be
+accessed using the ``.`` dot operator. But like scalar aliases, a tuple alias
+is a standalone expression that does not involve object types.
```
raddevon, line 41: I'm trying to understand whether it's accurate to say that a tuple alias "does not involve object types." My hypothesis was that a tuple could _contain_ an object type and as a result maybe we need to use different phrasing here.

In researching this, I uncovered some InternalServerErrors, so I'm not sure what the intended behavior actually is. 😅  See #5887.

Dhghomon: Okay, as discussed looks like this would serve best as a single example inside the existing section on expression aliases. This works just as well since the desired outcome is just that the reader come away from the page knowing that aliases can be more than simple scalar values on one end and expressions on schema-based types on the other.

raddevon: The example feels disconnected from the text that precedes it. What if we replaced the "heading" and the sentence with this?
> **Expression alias**
> Aliases can correspond to arbitrary EdgeQL expressions, for example tuples or entire queries.

Then, within the SDL block, we could add comments that would label each of those.

```
# Tuple alias
…

# Named tuple alias
…

# Query alias
…
```

Dhghomon: Yeah. Rereading it as it currently stands it starts out by introducing the section as "Query alias" but then says that these aliases "can correspond...to entire queries" which reads a bit oddly since it's already a given by the title. So that looks good I think.

Except I think I'll change the "Tuple alias" and such comments to "Expression producing a tuple alias" which is longer but no risk of giving the impression that we're introducing a strict set of possible aliases and that others aren't possible. Want to keep the sky's the limit feel.

Dhghomon: Although...reading this again the whole page starts by describing aliases as expressions. So maybe instead of Expression alias the title should be a bit different. Like "Other arbitrary alias examples".

raddevon (approved):

PR #5886, opened 2023-08-02
Add schema for ext::auth module

scotttrinh:
The basic structure of the schema is:

- An `Identity` type that represents some authorization server (`iss`) and some identity that exists on that AS (`sub`).

We pared back the modeling here to the barest essentials to allow updating based on usage, so for now, consumers will need to do their own profile-building (gathering any PII they might be interested in).

### DIFF
--- a/edb/lib/ext/auth.edgeql
+++ b/edb/lib/ext/auth.edgeql
@@ -24,10 +24,6 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
 
     create abstract type ext::auth::User;
 
-    create type ext::auth::Identity {
-        create required link user: ext::auth::User;
-    };
-
     create type ext::auth::Provider {
         create required property name: std::str;
         create required property url: std::str {
@@ -35,19 +31,9 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
         };
     };
 
-    create type ext::auth::Email {
-        create required property address: std::str;
-        create required property verified: std::bool;
-        create required property primary: std::bool;
-
-        create required link identity: ext::auth::Identity;
-
-        create constraint exclusive on ((.identity, .primary));
-    };
-
-    create type ext::auth::Claims {
+    create type ext::auth::Identity {
         create required link issuer: ext::auth::Provider;
-        create required link subject: ext::auth::Identity;
+        create required link subject: ext::auth::User;
 
         # Standard claims
         create required property sub: std::str;
@@ -65,6 +51,16 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
         create property zoneinfo: std::str;
         create property locale: std::str;
         create property updated_at: std::datetime;
+    };
+
+    create type ext::auth::Email {
+        create required property address: std::str;
+        create required property verified: std::bool;
+        create required property primary: std::bool;
+
+        create required link identity: ext::auth::Identity;
+
+        create constraint exclusive on ((.identity, .primary));
     };
 
     create type ext::auth::Session {
--- a/edb/lib/ext/auth.edgeql
+++ b/edb/lib/ext/auth.edgeql
@@ -22,8 +22,6 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
 
     create module ext::auth;
 
-    create abstract type ext::auth::User;
-
     create type ext::auth::Provider {
         create required property name: std::str;
         create required property url: std::str {
@@ -33,9 +31,8 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
 
     create type ext::auth::Identity {
         create required link issuer: ext::auth::Provider;
-        create required link subject: ext::auth::User;
 
-        # Standard claims
+        # Standard OpenID claims
         create required property sub: std::str;
         create property name: std::str;
         create property given_name: std::str;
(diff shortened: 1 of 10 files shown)


edb/lib/ext/auth.edgeql
```diff
@@ -17,4 +17,49 @@
...
+        create property nickname: std::str;
+        create property preferred_username: std::str;
+        create property profile: std::str;
+        create property picture: std::str;
+        create property website: std::str;
+        create property gender: std::str;
+        create property birthdate: std::str;
+        create property zoneinfo: std::str;
+        create property locale: std::str;
+        create property updated_at: std::datetime;
```
fantix, line 43: I usually keep only `sub` here (indexed with provider? unique?) and make the rest computed properties of a single JSON "claims" property, so that it's easier to update as a whole and allow storing custom claims (also semantically allows non-OpenID user profiles)

fantix: Yes to the constraint :v:

> We would store a `property claims: std::json`, right? And then use `json_get` to map from the json claim blob to the OpenID-like property name, right?

Yes!

> How would we deal with the fact that each provider might use different names for something like `picture` (`avatar_url` for GitHub, for example)?

Theoretically, we can choose either preprocessing or postprocessing - your current code would preprocess the received claims and store normalized properties, right? We can probably do the same with an additional `normalized_claims` property. (Yes I tend to keep the original claims we received, even just for debugging purposes)

Or, `picture` can be a smart computed property that also considers the `provider` value, like `picture := claims.avatar_url if .provider = 'github' else claims.picture` ... but I'm not so sure if we want to go there, because every time we add a provider, we may have to change the schema.

scotttrinh: > Theoretically, we can choose either preprocessing or postprocessing - your current code would preprocess the received claims and store normalized properties, right?

Yeah, I felt like that has the best DX for the consumer: we just document these well-known OpenID-like properties and the application developer doesn't have to think about the provider shape (and if it changes in the future, _we_ have the burden of updating our mappings).

> We can probably do the same with an additional normalized_claims property.

Why store this normalized as JSON rather than properties?

> (Yes I tend to keep the original claims we received, even just for debugging purposes)

Cool.

One point of clarification here: most of these providers are _not_ OpenID providers, so the data might come from disparate endpoints from the resource server. In GitHub's case, the `/user` and `/user/emails` endpoints map pretty cleanly to our own types, but other providers might require multiple requests for different important identity data. Given this, do you think it makes sense to maybe make the "original claims" JSON property hold some kind of map from URI -> JSON response?

```json 
{
  "https://api.github.com/user": {
    "id": "abc123",
    "login": "octocat"
  },
  "https://api.github.com/user/emails": [
    {
      "email": "octocat@example.com",
      "verified": true,
      "primary": true
    }
  ]
}
```

fantix: > Why store this normalized data as JSON rather than properties?

Good point! If we keep the original `claims` already, we don't need another cache layer before the individual properties.

> Given this, do you think it makes sense to maybe make the "original claims" JSON property hold some kind of map from URI -> JSON response?

Oh wow, now it's getting more complicated than I expected. That makes sense, however: If we are feeling relaxed with modifying the schema when adding providers, I'm thinking maybe we can have `GithubIdentity extending Identity` with extra properties for original responses (for easier and consistent user API access), together with computed properties like `picture := .user_response.avatar_url`. (Or rather, `Identity` holds the preprocessed normalized properties as the "cache")

scotttrinh: Yeah, my original idea was to have different types for each provider, but in our initial review of my proposed schema we (I think @elprans maybe suggested it?) came to the conclusion that there was more value in providing the application developer with a single consistent interface and not much value (at least for the application developer) to having a bunch of subtypes that captures the original shape of the data, since it might change over time.

I still see value in capturing raw data for debugging purposes though, so if we can come to some reasonable design there that doesn't add a lot of noise to the server code here, I'm all for it!

edb/lib/ext/auth.edgeql
```diff
@@ -17,4 +17,24 @@
...
 
 
-CREATE EXTENSION PACKAGE auth VERSION '1.0';
+CREATE EXTENSION PACKAGE auth VERSION '1.0' {
+    set ext_module := "ext::auth";
+
+    create module ext::auth;
+
+    create type ext::auth::Identity {
+        create required property iss: std::str;
```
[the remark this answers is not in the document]
fantix: Yeah, sure - even if it's just that we happened to borrow the concept from OpenID, `iss` sounds like a general-enough term.

Oh, does this mean, `iss` would be a URL as in OIDC, instead of a code name/identifier?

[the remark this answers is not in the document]
scotttrinh: > Oh, does this mean, iss would be a URL as in OIDC, instead of a code name/identifier?

Oh, great question. I wasn't really intending for it to be a URL, but that kinda makes sense especially if we implement some kind of generic provider. Any drawbacks? I wonder what the issuer would be in the case of GitHub: `https://github.com` or `https://api.github.com`?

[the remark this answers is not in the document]
fantix: The only drawback I can think of is a possible mismatch/confusion between our `iss` and the `iss` returned from the remote provider - perhaps we will add some special protocol (like `github+https://private.github.deploy.org`) when the `iss` returned from the provider is not quite enough for selecting a provider backend on our side. Then the user must be aware of such differences and not use the value in a wrong way.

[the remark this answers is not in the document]
scotttrinh: Alright, that seems fine if/when we get to a problem like that in the future. So for now, I'll add another property to the `BaseProvider` class for the provider's issuer URL 👍

fantix (approved):

1st1 (approved):

PR #5891, opened 2023-08-04
Initial HTTP server for auth extension

scotttrinh:
Adds two OAuth2 authorization code flow endpoints to the EdgeDB HTTP server:

- `/db/{dbname}/ext/auth/authorize?provider=github`: Configures the appropriate payload for the authorization code redirect to the authorization service to allow the user to authenticate, in this case with GitHub
- `/db/{dbname}/ext/auth/callback`: Receives the `code` and `state` on successful authentication by the end user

Plenty more to do here, but this much is ready to land into the trunk while we carry on development. Most of this has been thoroughly reviewed already, but definitely looking for feedback on anything that might block inclusion into `master`

## Notes

- Bumps us to Python 3.11 (due to use of `asyncio.TaskGroup`) and to the latest mypy
- Uses soon-to-be-replaced configuration values (to be replaced by a new extension configuration system)
- Only considers the extremely happy path for now


.github/workflows.src/tests.inc.yml
```diff
@@ -12,7 +12,7 @@
     - name: Set up Python
       uses: actions/setup-python@v4
       with:
-        python-version: '3.10.12'
```
fantix, line 15: @elprans this probably means we should issue future (nightly) releases in 3.11 too, right?

elprans: We already do. 3.x as well

elprans: https://github.com/edgedb/edgedb-pkg/blob/4cfef6561b97bf54de134d3b9e81c70bdc7197ba/edgedbpkg/edgedb/__init__.py#L64-L73

edb/server/protocol/auth_ext/base.py
```diff
@@ -0,0 +1,48 @@
...
+
+    async def fetch_user_info(self, token: str) -> data.UserInfo:
+        raise NotImplementedError
+
+    async def fetch_emails(self, token: str) -> list[data.Email]:
+        raise NotImplementedError
+
+    def _maybe_isoformat_to_timestamp(self, value: str | None) -> int | None:
+        return (
+            int(datetime.fromisoformat(value).timestamp()) if value else None
```
1st1, line 47: Are you rounding the timestamp intentionally here, btw?

edb/server/protocol/auth_ext/oauth.py
```diff
@@ -0,0 +1,99 @@
...
+        )
+
+    async def handle_callback(self, code: str) -> None:
+        token = await self.provider.exchange_code(code)
+
+        async with asyncio.TaskGroup() as g:
+            user_info_t = g.create_task(self.provider.fetch_user_info(token))
+            emails_t = g.create_task(self.provider.fetch_emails(token))
+        user_info = await user_info_t
+        emails = await emails_t
```
1st1, line 83: Actually, a faster way:

```suggestion
        user_info = user_info_t.get_result()
        emails = emails_t.get_result()
```

1st1 (approved):
Epic

edb/server/protocol/auth_ext/base.py
scotttrinh: Actually, there isn't a strong reason to prefer `int` here since the underlying spec just says "JSON number", but I kind of interpreted the spec to be "number of seconds" -> "integer of seconds".

See: https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims

edb/server/protocol/auth_ext/oauth.py
scotttrinh: @1st1 Do you mean [`result()`](https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.result) here? Couldn't find anything called `get_result()`

scotttrinh: @1st1 ⚡ [`501fab1` (#5891)](https://github.com/edgedb/edgedb/pull/5891/commits/501fab1bf82aee9931537cac818b8811fb8fe45b)

edb/server/protocol/auth_ext/base.py
scotttrinh: @1st1 ⚡ [`d6fb2fc` (#5891)](https://github.com/edgedb/edgedb/pull/5891/commits/d6fb2fcdaa0a7e11223a11606319af27272f7e57)

Issue #5895, opened 2023-08-04
 edb.server: Received signal: 15. Shutting down


petersamokhin:
hello there! still trying to make edgedb work on a remote postgres backend. while #5894 is not yet merged, I tried to test the setup locally to figure out the minimal requirements required, and I receive a weird error:

```
INFO 47 2023-08-04T18:53:37.035 edb.server: Connection established to backend database: edgedb
INFO 47 2023-08-04T18:53:37.902 edb.server: Received signal: 15.
INFO 47 2023-08-04T18:53:37.903 edb.server: Shutting down.
ERROR: Could not complete instance bootstrap in 30000000 seconds. If you have slow hardware, consider increasing the timeout via the EDGEDB_DOCKER_BOOTSTRAP_TIMEOUT_SEC variable.
```

I changed the timeout from 300 to 30000000 seconds to demonstrate the wrong message. 
The actual setup takes just a few seconds, so the timeout was not reached.

however, the main issue here is that the server fails with exit code 1, and shows some "signal: 15" in logs

<details>
  <summary>SQL setup</summary>
  
```sql
CREATE DATABASE edgedb;
CREATE USER edgedb WITH ENCRYPTED PASSWORD 'edgedb';
ALTER DATABASE edgedb OWNER TO edgedb;
```
  
</details>

<details>
  <summary>Full log (67 lines)</summary>
  
```
2023-08-04T18:53:18.903472095Z Bootstrapping EdgeDB instance on remote Postgres cluster...
2023-08-04T18:53:20.121165721Z INFO 47 2023-08-04T18:53:20.120 edb.server: starting EdgeDB server 3.2+740bb0d, built on 2023-07-21T19:08Z from revision b7dc55395 (2023-07-21T01:09Z) for aarch64-unknown-linux-gnu
INFO 47 2023-08-04T18:53:20.121 edb.server: instance name: '_unknown'
INFO 47 2023-08-04T18:53:21.372 edb.server: Detected 97 backend connections available.
INFO 47 2023-08-04T18:53:26.834 edb.server: Creating the necessary PostgreSQL extensions...
INFO 47 2023-08-04T18:53:26.860 edb.server: Initializing the standard library...
WARNING 47 2023-08-04T18:53:29.248 edb.server: skipping "pg_toast_1262" --- only superuser can analyze it
WARNING 47 2023-08-04T18:53:29.248 edb.server: skipping "pg_toast_1262_index" --- only superuser can analyze it
[...]

Issue #5592, opened 2023-06-04
underlying database names


petersamokhin:
hello there! tl;dr is there a way to control the underlying database names?
currently we see:
```
 V2f147ded60___edgedbtpl__
 V2f147ded60___edgedbsys__
 V2f147ded60_edgedb
```

This is not possible to use with the managed DB instances where the databases are pre-created.
However, we could pre-create multiple databases once and provide the EdgeDB instance with their names.
Is that supported? If not, can it be supported in the future?

elprans:
EdgeDB supports a "single database" mode, whereby all metadata is stored in a single database and the `CREATE DATABASE` statement is disabled, so you only get the `edgedb` database.  This mode is normally auto-detected on startup.  Are you seeing a different behavior?

petersamokhin:
@elprans could you please document this behaviour somewhere? 
How is it determined exactly? e.g. by trying to create a database and falling back to the single database named `edgedb`?
Is it possible to control the default/fallback db name?

elprans:
> by trying to create a database and falling back to the single database

Yes

>  Is it possible to control the default/fallback db name?

In single db mode EdgeDB uses the database provided in `--backend-dsn` (`EDGEDB_SERVER_BACKEND_DSN`), e.g. if you give it `postgres://host:port/mydb` it'll use `mydb`.

petersamokhin:
@elprans  Thank you for clarification! The last question: if we have the postgres backups, that is used as the backend for the edgedb instance, do we still need to backup (dump) edgedb? Does edgedb store anything anywhere else apart from the postgres?

petersamokhin:
Perfect! Thanks a lot.

petersamokhin:
hey @elprans, just realised that the last time I got it wrong.

> and the CREATE DATABASE statement is disabled, so you only get the edgedb database. 

I thought that there will be only a single underlying postgres database, but >1 databases in edgedb.
i.e. it is not really clear that each edgedb database requires a separate&dedicated postgres database.

could you please confirm?

elprans:
> each edgedb database requires a separate&dedicated postgres database.

Yes.  If you don't have a permission to `CREATE DATABASE` in Postgres, `CREATE DATABASE` in EdgeDB will likewise not be available.

elprans:
To clarify, the underlying database name in single-database mode would be whatever you specified in `--backend-dsn`

PR #5900, opened 2023-08-05
Fix healthchecks on single-database instances

elprans:
Instances running on backends without `CREATE_DATABASE` capability will
only have the `edgedb` database in them, while the healthcheck endpoints
attempt to connect to the system database which is only present in
multi-database mode.

Fixes: #5895


edb/server/protocol/system_api.py
```diff
@@ -89,8 +89,13 @@ def _response_ok(response, message):
 
 
 async def _ping(server):
+    if server.get_backend_runtime_params().has_create_database:
+        dbname = edbdef.EDGEDB_SYSTEM_DB
+    else:
+        dbname = edbdef.EDGEDB_SUPERUSER_DB
```
petersamokhin, line 95: Will it be adjusted to the database name provided in BACKEND_DSN? i.e. if the (one and only) available database is named differently (not `edgedb`), will it still work? 

https://github.com/edgedb/edgedb/issues/5592#issuecomment-1577786582
> In single db mode EdgeDB uses the database provided in --backend-dsn (EDGEDB_SERVER_BACKEND_DSN), e.g. if you give it postgres://host:port/mydb it'll use mydb.

elprans: Yes. In single-database mode `edgedb` gets translated to whatever was specified in the backend DSN.

fantix (approved):

fantix:
[test ran](https://github.com/edgedb/edgedb/actions/runs/5776933393)

PR #5909, opened 2023-08-10
Make state serializers in the compiler

fantix:
This is to release the I/O process from long-running blocking computation.

- [x] Make serializer in the compiler
- [x] Use serializer properly in the dbview

### DIFF
--- a/edb/server/compiler/compiler.py
+++ b/edb/server/compiler/compiler.py
@@ -384,6 +384,12 @@ class CompilerState:
     local_intro_query: Optional[str]
     global_intro_query: Optional[str]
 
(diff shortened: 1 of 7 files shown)


edb/server/compiler/compiler.py
```diff
@@ -2351,6 +2366,13 @@ def _try_compile(
         rv.in_type_args = in_type_args
         rv.in_type_data = in_type_data
 
+    if final_user_schema:
```
elprans, line 2369:
```suggestion
    if final_user_schema is not None:
```

elprans (approved):

PR #5921, opened 2023-08-17
Don't unpickle schema in I/O process

fantix:
- [x] Add pickled user schema
- [x] Add pickled global schema
- [x] Use pickled user/global schema in compiler client
- [x] Move inteprete_backend_error back to the compiler
- [x] Move parsing of JSON schema to the compiler
- [x] Allow compiler initialization with JSON schema
- [x] Edit introspect_user/global_schema
- [x] Deal with state serializer
- [x] Drop unpickled schema in dbview

### DIFF
--- a/edb/graphql/extension.pyx
+++ b/edb/graphql/extension.pyx
@@ -186,14 +186,12 @@ async def handle_request(
 
             # only use the backend if schema is required
             if static_exc is errormech.SchemaRequired:
-                ex = errormech.interpret_backend_error(
-                    s_schema.ChainedSchema(
-                        tenant.server.get_std_schema(),
-                        db.user_schema,
-                        tenant.get_global_schema(),
-                    ),
+                compiler_pool = tenant.server.get_compiler_pool()
+                ex = await compiler_pool.interpret_backend_error(
+                    db.user_schema_pickled,
+                    tenant.get_global_schema_pickled(),
                     ex.fields,
-                    from_graphql=True,
+                    True,
                 )
             else:
                 ex = static_exc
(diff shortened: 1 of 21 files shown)


edb/graphql/extension.pyx
```diff
@@ -186,14 +186,12 @@ async def handle_request(
...
             if static_exc is errormech.SchemaRequired:
-                ex = errormech.interpret_backend_error(
-                    s_schema.ChainedSchema(
-                        tenant.server.get_std_schema(),
-                        db.user_schema,
-                        tenant.get_global_schema(),
-                    ),
+                compiler_pool = tenant.server.get_compiler_pool()
+                ex = await compiler_pool.interpret_backend_error(
+                    db.user_schema_pickled,
```
elprans, line 191: It's either a "pickled user schema" or "user schema pickle", I prefer the latter.

```suggestion
                    db.user_schema_pickle,
```

edb/server/compiler/compiler.py
```diff
@@ -387,6 +387,8 @@ class CompilerState:
 
     @functools.cached_property
     def state_serializer_factory(self) -> sertypes.StateSerializerFactory:
+        # TODO: This factory will probably need to become per-db once
```
elprans, line 390: @msullivan, need your input here.  Config spec can already be different w.r.t enabled extensions, or am I confusing things here?

msullivan: We don't support doing session configs with extension configs yet, so I think it is fine

msullivan: Hopefully I will implement that in the next month but I'll try to fix it then

elprans (approved):

PR #5931, opened 2023-08-21
Avoid using global_schema for roles in I/O process

fantix:
* Introspect roles directly without global schema.
* Include `roles` in QueryUnit when global_schema is updated.

This is a partial revert of [this commit](https://github.com/edgedb/edgedb/pull/2183/commits/7c40e2fc306a945dd88dfa1437669ab85cb9efb9#diff-397281e6358a1aadc225ecbf91ee76794703c819aea1faf7913df30561ff0911L336-L347).

### DIFF
--- a/edb/server/compiler/compiler.py
+++ b/edb/server/compiler/compiler.py
@@ -60,6 +60,7 @@ from edb.schema import objects as s_obj
 from edb.schema import objtypes as s_objtypes
 from edb.schema import pointers as s_pointers
 from edb.schema import reflection as s_refl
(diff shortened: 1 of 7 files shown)


fantix:
Ah, the `roles_intro_query` could also be one of the `sys_queries` I guess?

elprans:
> Ah, the `roles_intro_query` could also be one of the `sys_queries` I guess?

Yes.  In fact, we already have it there.

elprans (approved):

PR #5932, opened 2023-08-21
Add a library for producing nicely indented things with f-strings

msullivan:
I hacked this up a few months ago and finally got fed up enough trying to
read the mangled config views that we generate to try actually using it.

We could use this to generate nicer code in a bunch of places in
pgsql/delta too, where we create triggers and run queries.

I like this, but we can skip it if people don't like it.

### DIFF
--- a/edb/common/xdedent.py
+++ b/edb/common/xdedent.py
@@ -1,0 +1,138 @@
+#
+# This source file is part of the EdgeDB open source project.
+#
+# Copyright 2011-present MagicStack Inc. and the EdgeDB authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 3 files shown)


elprans (approved):
LGTM, but I'd be explicit that we reserve this for places where we generate-once-and-save because of the extra formatting cost.

tests/common/test_xdedent.py
```diff
@@ -0,0 +1,124 @@
...
+
+        q = '''
+        (SELECT
+            ARRAY[(q0.val)->>'name'] AS key
+        ) AS k0
+        '''
+        sources = [q, q]
+        fromlist = ',\n'.join(f'LATERAL {X(s)}' for s in sources)
+        res = xdedent.xdedent(fromlist)
+        self.assertEqual(res, EXPECTED)
```
aljazerzen, line 124: These tests are not that easy to read :D

edb/common/xdedent.py
```diff
@@ -0,0 +1,138 @@
...
+the nesting structure and interprets it.
+Obviously, as with all schemes for
+in-band signalling, all hell can break loose if the signals appear in
+the input data unescaped.
+
+Our signal sequences contain a null byte and both kinds of quote
+character, so you should be fine as long as any untrusted data
+either:
+ * Has no null bytes
+ * Has at least one kind of quote character escaped in it somehow
```
aljazerzen, line 49:
```suggestion
Our signal sequences contain a null byte and both kinds of quote
character, so you should be fine unless the untrusted data:
 * has null bytes and
 * does not have any kind of quote character escaped in it somehow
```

aljazerzen (approved):

PR #5934, opened 2023-08-21
Move DESCRIBE CONFIG into the compiler

msullivan:
Currently it gets precompiled at bootstrap time and cached in
the functions _describe_system_config_as_ddl and
_describe_database_config_as_ddl

Once we have extension/user-creatable configs, having a single
pre-compiled config description function won't work.

Here, I've fixed it by moving it out of the function and recompiling
it in the compiler at query execution time. This is obviously slower
but I don't think this is on critical paths?

Another option would be to regenerate _describe_database_config_as_ddl
whenever a config object is created/deleted/altered.

I think that would work, and would have better performance
characteristics than this, but I'm not totally comfortable with having
standard library functions dynamically getting recompiled based on the
contents of the user schema. Maybe I should just get over that, though?

### DIFF
--- a/edb/edgeql/compiler/config_desc.py
+++ b/edb/edgeql/compiler/config_desc.py
@@ -128,13 +128,15 @@ def _describe_config(
         " ?? false"
     )
     query = (
-        f"FOR conf IN {{cfg::get_config_json(sources := [{ql(source)}])}} "
+        "assert_exists(assert_single(("
+        + f"FOR conf IN {{cfg::get_config_json(sources := [{ql(source)}])}} "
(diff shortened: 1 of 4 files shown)


elprans:
How much slower is this compared to the current implementation?

msullivan:
Current version: 250ms uncached, 215ms cached
New version: 770ms uncached, 20ms cached

Looks like the actual query performance is quite a bit better with the non-function version!
(At the cost of our compiler, of course)

elprans:
I wonder why that is

elprans (approved):

PR #5937, opened 2023-08-22
Fix computed dunder `__subject__` -> `__source__`

raddevon:
`__subject__` is not available in computeds


Dhghomon (approved):
Change to __source__ for now since that's the reality, and then work on supporting __subject__ later on, right? Looks good to me.

PR #5941, opened 2023-08-22
Upsert `Identity` object on successful auth

scotttrinh:
Create or update an `Identity` object when authentication succeeds.

### DIFF
--- a/edb/server/protocol/auth_ext/oauth.py
+++ b/edb/server/protocol/auth_ext/oauth.py
@@ -20,6 +20,7 @@
 import urllib.parse
 
 import httpx
+from edb.server.protocol import execute
 
 
 from typing import Any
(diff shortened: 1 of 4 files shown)


edb/server/protocol/auth_ext/oauth.py
```diff
@@ -81,7 +82,43 @@ async def handle_callback(self, code: str) -> None:
...
+
+        await execute.parse_execute_json(
+            db=self.db,
+            query="""\
+with
+  iss := <str>$issuer_url,
+  sub := <str>$provider_id,
+  email := <optional str>$email,
+
+  # update an existing identity, if it exists
```
fantix, line 95: why not use [our upsert](https://www.edgedb.com/docs/edgeql/insert#upserts) 😏

fantix (approved):

edb/server/protocol/auth_ext/oauth.py
scotttrinh: Oh true, I originally had a much more complicated structure that had to look-up existing emails and find identities, etc. Lemme see if it can just be simplified back down to our `unless conflict` call.

scotttrinh: @fantix ⚡ [`c10bc1e` (#5941)](https://github.com/edgedb/edgedb/pull/5941/commits/c10bc1e05ca8a6e857f7ba7479f03d30c3b55125)

Thanks for noticing this!

Issue #5945, opened 2023-08-23
Upload a new plpgsql function


AdrienPensart:
Hello!

suppose I want to extend edgedb without having to include a new function in its std library, how can it be done?
custom edgeql function do not support plpgsql yet, I don't even know if it is in the roadmap.

For the moment, I need this : 

bytes_to_human => pg_pretty_size

I tested it and implement it (see linked PR).

Awesome project again guys!

AdrienPensart:
I finally implemented it this way : 

`
function bytes_to_human(size: int64, k: int64 = 1000, decimals: int64 = 2, units: array<str> = [' B', ' KB', ' MB', ' GB', ' TB', ' PB', ' EB', ' ZB', ' YB']) -> str {                                                                                         using (                                                                                                                                                                                                                                                         select '0' ++ units[0] if size = 0                                                                                                                                                                                                                          else (                                                                                                                                                                                                                                                          with i := <int64>math::floor(math::ln(math::abs(size)) / math::ln(k))                                                                                                                                                                                       select to_str(std::round(<decimal>(size / k ^ i), decimals)) ++ units[i]                                                                                                                                                                                )                                                                                                                                                                                                                                                       );                                                                                                                                                                                                                                                          annotation title := "Convert a byte size to human readable string";                                                                                                                                                                                     };
`

PR #5946, opened 2023-08-23
add std::bytes_to_human which maps to pg_size_pretty

AdrienPensart:
Linked to issue #5945


elprans (commented):
I think it would make more sense for this to perhaps be a `to_str(size: cfg::memory, format='pretty')` since this is explicitly about formatting memory units.

PR #5951, opened 2023-08-24
Add chatbot guide

raddevon:


docs/guides/tutorials/chatgpt_bot.rst
```diff
@@ -0,0 +1,2285 @@
...
+With that out of the way, let's walk through how the pieces fit together.
+
+
+Implementation overview
+-----------------------
+
+The general implementation has two steps (which we'll also follow in the
+guide):
+
+1. create embeddings for each section using `OpenAI's embeddings API
```
diksipav, line 77: I think that we should not exclude the first part which is getting the docs files into the right format and splitting that into sections. This is crucial part of the whole thing. 

Also I'm not sure when you say the general implementation has 2 steps, do you think on creating and storing embeddings. In general, also the next 3 steps are part of the implementation.

raddevon: > I think that we should not exclude the first part which is getting the docs files into the right format and splitting that into sections. This is crucial part of the whole thing.

What do you mean by "getting the docs files into the right format?" Can you show me where in the example app this happens and where we split the sections? I think I have missed those parts. My understanding was we didn't need them for this app, so that's why I removed them here.

> Also I'm not sure when you say the general implementation has 2 steps, do you think on creating and storing embeddings. In general, also the next 3 steps are part of the implementation.

Let me work on rewording this. You're right that it isn't strictly accurate the way I've written it.

docs/guides/tutorials/chatgpt_bot.rst
```diff
@@ -0,0 +1,2285 @@
...
+    into multiple sections because of the LLM's token limit. LLMs divide text
+    into tokens. For English text, 1 token is approximately 4 characters or
+    0.75 words. LLMs have limits on the number of tokens they can receive and
+    send back.
+
+    One approach to mitigate this is to parse your documentation files and
+    create new sections every time you encounter a header. If you use this
+    approach, consider section lengths when writing your documentation. Use
+    your token limit to set a hard limit on documentation section length to
+    make sure it's accessible to your chatbot. If you find a section is too
```
diksipav, line 391: `Use your token limit to set a hard limit on documentation section length to make sure it’s accessible to your chatbot. ` I'm not sure about this sentence.

Maybe we can move this whole section to the end or just rephrase/remove the sentence from above: 
`One approach to mitigate this is to parse your documentation files and create new sections every time you encounter a header. If you use this approach, consider section lengths when writing your documentation. Use your token limit to set a hard limit on documentation section length to make sure it’s accessible to your chatbot. If you find a section is too long, consider ways you might break it up with additional headings. This will probably make it easier to read for your users too!`

docs/guides/tutorials/chatgpt_bot.rst
raddevon: Pushed a change to address your second concern.

docs/guides/tutorials/chatgpt_bot.rst
raddevon: Yeah, I don't think it adds much. I'll remove it.

docs/guides/tutorials/chatgpt_bot.rst
```diff
@@ -0,0 +1,2305 @@
...
+project, this will make it easier to import them.
+
+.. lint-off
+
+.. code-block:: json-diff
+    :caption: tsconfig.json
+
+      {
+        "compilerOptions": {
+          "target": "es5",
```
diksipav, line 154: I don't know if you mentioned this later (couldn't find it with ctrl+F), we need es6 here so maybe we can show it now since we show the whole tsconfig file anyway.

docs/guides/tutorials/chatgpt_bot.rst
```diff
@@ -0,0 +1,2305 @@
...
+.. code-block:: typescript
+
+    import OpenAI from "openai";
+
+    export function initOpenAIClient() {
+      if (!process.env.OPENAI_API_KEY)
+        throw new Error("Missing environment variable OPENAI_API_KEY");
+
+      return new OpenAI({
+        apiKey: process.env.OPENAI_API_KEY!,
```
diksipav, line 331: we don't need exclamation mark at the end, I forgot to remove it (its from previous versions), I updated in the example repo too (have to push it)

docs/guides/tutorials/chatgpt_bot.rst
raddevon: I missed this. Thank you! What's the reason for it? I will provide a brief explanation too.

docs/guides/tutorials/chatgpt_bot.rst
```diff
@@ -0,0 +1,2305 @@
...
+.. code-block:: -diff
+
+      EDGEDB_DSN=<your-dsn>
+      EDGEDB_CLIENT_TLS_SECURITY="insecure"
+    + OPENAI_API_KEY="<your-openai-api-key>"
+
+Instead of ``<your-openai-api-key>``, paste in the key you just created.
+
+While we're here, let's get that key ready to be used. We will be making calls
+to the OpenAI API. We'll create a ``utils`` module and export a function from
```
diksipav, line 318: Maybe we can write something like:
`We’ll create a `utils` module and define in it a `initOpenAIClient` function that initialises an OpenAI client. We will reuse this function in every file we need a new client instance.`

raddevon: Right. Because we're not actually reusing a single client. Makes sense.

docs/guides/tutorials/chatgpt_bot.rst
```diff
@@ -0,0 +1,2305 @@
...
+
+Create a ``.env.local`` file in the root of your new Next.js project.
+
+.. code-block:: bash
+
+    $ touch .env.local
+
+We're going to add a couple of variables to that file to configure EdgeDB.
+We'll need to run a command on our new instance to get the value for one of
+those. Since the runtime we'll be using in our Next.js project does not have
```
diksipav, line 262: I think we can explain this a bit better. @jaclarke helped me to come with this:

`We're going to add a couple of variables to that file to configure the EdgeDB client. We'll need to run a command on our new instance to get the value for one of those. Since we'll be using the [Edge Runtime](https://nextjs.org/docs/app/api-reference/edge) in our Next.js project, the edgedb-js client won't be able to access the Node.js filesystem API's it usually uses to automatically find your instance, so we need to provide the DSN for the instance instead. To get that, run this command:`

docs/guides/tutorials/chatgpt_bot.rst
```diff
@@ -0,0 +1,2305 @@
...
+        required content: str;
+        required tokens: int16;
+        required embedding: OpenAIEmbedding;
+
+        index ext::pgvector::ivfflat_cosine(lists := 3)
+          on (.embedding);
+      }
+    …
+
+The ``Section`` contains properties to store the content, a count of tokens,
```
diksipav, line 499: do we need here `to store`?

docs/guides/tutorials/chatgpt_bot.rst
```diff
@@ -0,0 +1,2305 @@
...
+and the embedding, which is of the custom scalar type we created in the
+previous step.
+
+We've also added an index inside the ``Section`` type to speed up queries. In
```
diksipav, line 503: I think we should remove index completely, our guide has 3 files :D 
But we should mention that it should be used when there are lots of files, I don't really know when the results start being visible? @jaclarke 
@vpetrovykh is on vacation, maybe @elprans can help? Do you know when it makes sense to add index, after 100 files, 1000, 10000?

elprans: There's nothing wrong with adding an index on a small table. I would leave this as-is.

docs/guides/tutorials/chatgpt_bot.rst
```diff
@@ -0,0 +1,2305 @@
...
+    only documentation sections which have changed.
+
+    You can achieve this by saving content checksums and section paths as part
+    of your ``Section`` objects. The next time you run generation, compare the
+    section's current checksum with the one you stored in the database, finding
+    it by its path. You don't need to generate embeddings and update the
+    database for a given section unless the two checksums are different
+    indicating something has changed.
+
+    If you decide to go this route, add properties your ``Section`` as shown
```
diksipav, line 562: we miss here "to your Section" I guess

In general users can save path or title, there should be a property that is unique, I used paths to allow having the same file titles under different folders... not sure is it easy to explain that it doesn't have to be `only path`

docs/guides/tutorials/chatgpt_bot.rst
```diff
@@ -0,0 +1,2305 @@
...
+    import e from "dbschema/edgeql-js";
+    import { initOpenAIClient } from "./utils";
+
+    dotenv.config({ path: ".env.local" });
+
+    const openai = initOpenAIClient();
+
+    interface Section {
+      id?: string;
+      path: string;
```
diksipav, line 652: we don't have path anymore : )

docs/guides/tutorials/chatgpt_bot.rst
raddevon: Good catch!

I went ahead and generalized this advice too to let readers know they can use any unique identifier.

docs/guides/tutorials/chatgpt_bot.rst
```diff
@@ -0,0 +1,2316 @@
...
+        });
+      });
+
+      await query.run(client, { sections });
+      console.log("Embedding generation complete");
+    }
+    // …
+
+
+Putting it all together
```
diksipav, line 1013: Do we want this repetition'? Users can go and copy from the example project
I am also fine with leaving this here, just the guide is very long.

docs/guides/tutorials/chatgpt_bot.rst
diksipav: The edgedb-js needs es6 because it uses some data structures that are available from es6 onward. I know about Set() but maybe there is sth more.

docs/guides/tutorials/chatgpt_bot.rst
```diff
@@ -0,0 +1,2316 @@
...
+As we write our handler, one important consideration is that answers can be
+quite long. We could wait on the server side to get the whole answer from
+OpenAI and then send it to the client, but that would feel slow to the user.
+OpenAI supports streaming, so instead we can send the answer to the client in
+chunks, as they arrive to the server. With this approach, the user doesn't have
+to wait for the entire response before they start getting feedback and our API
+seems faster.
+
+In order to stream responses, we will use the browser's `server-sent events
+(SSE) API`_. Server-sent events enables a client to receive automatic updates
```
diksipav, line 1202: I found this same sentence on Wikipedia: "Server-sent events enables a cli..."

But I think it's better to say that server maintains data transmissions instead initiates: 

`Server-sent events enables a client to receive automatic updates from a server via an HTTP connection, and describes how server maintains data transmissions to client once an initial client connection has been established. The client sends a request and with that request initiates a connection with our server. The server then sends data back to the client in chunks until all of the data is sent, at which point it closes the connection.`

docs/guides/tutorials/chatgpt_bot.rst
```diff
@@ -0,0 +1,2316 @@
...
+``route.ts`` file inside the route's folder.
+
+Let's create a new folder for the answer generation route inside ``app/api``.
+
+.. code-block:: bash
+
+    $ mkdir app/api && cd app/api
+    $ mkdir generate-answer && touch generate-answer/route.ts
+
+As the final setup step, we will install the ``common-tags`` NPM package which
```
diksipav, line 1233: We are not really generating HTML, codeBlock and oneLineTrim are more used to properly indent/unindent code:  

We also need to install the ``common-tags`` NPM package (and its corresponding
types package) which gives us some useful template tags that  we will use later
when we create the prompt from user's question and related sections.

.. code-block:: bash

    $ npm install common-tags @types/common-tags

diksipav: I updated in the example repo to use stripIndents instead of codeBlock.

docs/guides/tutorials/chatgpt_bot.rst
```diff
@@ -0,0 +1,2316 @@
...
+
+.. note::
+
+    Like before, you may want to read along for understanding and copy/paste
+    the completed route at the end of this section.
+
+.. code-block:: typescript
+    :caption: app/api/generate-answer/route.ts
+
+    import { codeBlock, oneLineTrim } from "common-tags";
```
diksipav, line 1270:
```suggestion
    import { codeBlock, oneLineTrim } from "common-tags";
```

docs/guides/tutorials/chatgpt_bot.rst
raddevon: OK. I'll do a little work on that one and slot it in. Thank you!

docs/guides/tutorials/chatgpt_bot.rst
raddevon: That's also from `common-tags`? I can't see the change on the examples repo yet.

docs/guides/tutorials/chatgpt_bot.rst
```diff
@@ -0,0 +1,2318 @@
...
+    function createFullPrompt(query: string, context: string) {
+        const systemMessage = `
+            As an enthusiastic EdgeDB expert keen to assist,
+            respond to questions referencing the given EdgeDB
+            sections.
+
+            If unable to help based on documentation, respond
+            with: "Sorry, I don't know how to help with that."`;
+
+        return codeBlock`
```
diksipav, line 1600: return codeBlock`
```suggestion
        return stripIndents`
```

docs/guides/tutorials/chatgpt_bot.rst
```diff
@@ -153,8 +153,7 @@ will make it easier to import them.
 
       {
         "compilerOptions": {
-    -     "target": "es5",
-    +     "target": "es6",
+          "target": "es6",
```
raddevon, line 156: We're showing the changes the user should make, so this needs to be a diff. Please revert.

docs/guides/tutorials/chatgpt_bot.rst
```diff
@@ -0,0 +1,2318 @@
...
+* ``target``: Embedding array to compare against to find related sections. In
+  this case, these will be the questions's embeddings we just generated.
+* ``matchThreshold``: Similarity threshold. Only matches with a similarity
+  score below this threshold will be returned. This will be a number between
+  ``0.0`` and ``1.0``. Values closer to ``0.0`` mean the documentation sections
+  must be very similar to the question while values closer to ``1.0`` allow for
+  more variance.
+* ``matchCount``: Maximum number of sections to return
+* ``minContentLength``: Minimum number of characters the sections should have
+  in order to be considered
```
diksipav, line 1498: Is there a way to make list items more consistent in the terms of punctuation we use? If we list them, I think it's fine to separate them with `,`, if we have longer explanations and sentences to finish them with `.`.

Probably it's also fine to mix things, not many people pay attention to this.

docs/guides/tutorials/chatgpt_bot.rst
```diff
@@ -0,0 +1,2318 @@
...
+
+This function takes the embeddings of the question (the ``embedding``
+parameter) and returns the related documentation sections.
+
+We start by running the query and passing in some values for the parameters:
+
+- the question embeddings that were passed to the function
+- a ``matchThreshold`` value of ``0.3``. You can tinker with this if you don't
+  like the results.
+- a ``matchCount``. We've chosen ``8`` here which represents the most sections
```
diksipav, line 1559: For this whole list I have the same suggestion as above. : )

docs/guides/tutorials/chatgpt_bot.rst
raddevon: I'm OK making sure every list item ends with a period in a case where any of them do for a given list. I have a consistent rule I apply for this, but that doesn't always result in a consistent aesthetic within a given list. Maybe the better rule is, "if any items in a list end in a period, they all do." Would that work better? If an entire list is contains sentence fragments, it doesn't feel right to me to end those with a period.

If that does work better, this feels like something that should go in an issue that we can revisit later. I don't think it's critical to getting this published.

docs/guides/tutorials/chatgpt_bot.rst
diksipav: We should never do that? I believe that explanations on Wikipedia can be correct and the same as on the other places on more established websites like MDN. 
Not sure where I get this from when I firstly wrote it (now when I searched for better wording I found it there), but I never went through the guide the 2nd time so now I realised it can be better explained.

diksipav: Oh, we have to give credit to the organisations that came up with the text if we copy/paste it. Tnx

raddevon: Yeah, it's OK as long as we give attribution.

diksipav (approved):

PR #5957, opened 2023-08-25
Avoid duplicating outputs for value/serialized for tuples

msullivan:
Tuple queries with extra levels of nesting like
`select (select (1, 'foo'))` currently duplicates some of the query outputs:
```
    "expr-3~2"."0_value~1" AS "0_value~2",
    "expr-3~2"."1_value~1" AS "1_value~2",
    "expr-3~2"."0_value~1" AS "0_serialized~1",
    "expr-3~2"."1_value~1" AS "1_serialized~1"
```

This seemed weird, because we do have machinery for reusing existing
outputs in order to avoid exactly this sort of duplication
(`find_path_output`).

I originally thought that the bug was in `find_path_output`, which worked by
searching through `path_namespace` for an expression that matched and
whose `(path_id, aspect)` key was already present in `path_outputs`.
I thought that the problem was that if the `path_id` got remapped by the
`view_path_id_map`, the existing output would be under a different
path_id, and so we wouldn't find it.

But after thinking about it more, I realized that this didn't really
make sense: we remap path_ids at the start of get_path_output, so
there shouldn't be a mismatch between `path_namespace` and
`path_outputs`: both should use the inner remapped name.

Then, looking at backtraces from where this divergence showed up, I
realized that we were actually going out of our way to make this
happen: we were *reversing* the map on the tuple element path ids and
recursively calling `_get_path_output`, which then called
`get_path_var` which reapplied the map, causing the divergence.

Fortunately I was able to just rip all that code out. I suspect that
it was made unnecessary by #2274 or #3034.

### DIFF
--- a/edb/pgsql/compiler/pathctx.py
+++ b/edb/pgsql/compiler/pathctx.py
@@ -135,13 +135,6 @@ def get_path_var(
         rel = rel.query
 
     if flavor == 'normal':
-        # Check if we already have a var, before remapping the path_id.
-        # This is useful for serialized aspect disambiguation in tuples,
-        # since process_set_as_tuple() records serialized vars with
-        # original path_id.
-        if (path_id, aspect) in rel.path_namespace:
-            return rel.path_namespace[path_id, aspect]
-
         if rel.view_path_id_map:
             path_id = map_path_id(path_id, rel.view_path_id_map)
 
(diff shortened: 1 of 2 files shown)


aljazerzen (approved):
Nice, removed code and improved SQL.

Issue #4281, opened 2022-08-20
Add `to_bytes` converter function.


vpetrovykh:
Currently `bytes` are hard to work with because they can't be cast into any other type (except `json`), nor can other types be cast into `bytes`. So this type becomes a kind of opaque blob that is basically good for storing some raw data, but not for much else.

We can introduce converter functions that take "encoding" as the second argument and then we could find more uses for the `bytes` type.

We can have the `str`/`bytes` converters:
```
function to_bytes(val: str, encoding: str) -> bytes
function to_str(val: bytes, encoding: str) -> str
```
We can also have `int16`/`int32`/`int64`/`bytes` converters:
```
function to_bytes(val: int16, little_endian: bool = true) -> bytes
function to_bytes(val: int32, little_endian: bool = true) -> bytes
function to_bytes(val: int64, little_endian: bool = true) -> bytes
function to_bytes(val: bytes, little_endian: bool = true) -> int16
function to_bytes(val: bytes, little_endian: bool = true) -> int32
function to_bytes(val: bytes, little_endian: bool = true) -> int64
```

tailhook:
I'm not sure why `little_endian = true` is the default. I would say that network byte order should be default. As it's more common in storage and data transmission.

I would also prefer enum instead of boolean flag, but that is probably too slow.

Also what are the downsides of using `utf-8` encoding by default?

1st1:
> I would say that network byte order should be default.

+1

1st1:
> Also what are the downsides of using utf-8 encoding by default?

None, I also think this is what we should do.

elprans:
int -> bytes converters have dubious value IMO.  I added the `to_str` / `to_bytes` in #5960.

PR #5960, opened 2023-08-25
Add `to_str(bytes)` and `to_bytes(str)` functions

elprans:
These can be used to convert bytes to strings and vice versa with
the specified encoding.

Fixes: #4281


docs/stdlib/bytes.rst
```diff
@@ -124,6 +127,24 @@ Bytes
 
 ---------
 
+.. eql:function:: std::to_bytes(s: str, encoding: optional str="utf-8") \
```
1st1, line 130: Should 'encoding' be an enum?

msullivan (approved):

docs/stdlib/bytes.rst
msullivan: That sounds like it would be annoying for us with little gain

1st1: We use enum for alphabet in the parallel PR for base64. Enum enables strict typing and removes ambiguity about which case the string literal should be in.

tests/test_edgeql_functions.py
```diff
@@ -3098,6 +3098,29 @@ async def test_edgeql_functions_to_str_07(self):
                 await self.con.query(
                     r'''SELECT to_str(<cal::local_time>'15:01:22', '');''',)
 
+    async def test_edgeql_functions_string_bytes_conversion(self):
+        encodings = ["utf-8", "koi8-u", "ISO_8859_5"]
+        string = "Паляниця"
+
+        for encoding in encodings:
```
1st1, line 3105: Would be great to see an "expect raises" test to make sure we don't leak SQL's NULLs as errors.

1st1: (And that the errors we can propagate are actually mapped to something but an ISE)

docs/stdlib/bytes.rst
elprans: I agree with Sully that this would be annoying to maintain as an enum.  The list is quite large, could be Postgres-version dependent, and is actually [dynamic](https://www.postgresql.org/docs/current/sql-createconversion.html).

> removes ambiguity about which case the string literal should be in

Its case-insensitive (should probably document this fact).  Also, common variations of encoding names are accepted, e.g 'utf-8' and 'utf8'.

1st1: I don't like this. Exposing a standard library function that's tied to the underlying postgres version, poorly documented because of that, can fail at runtime when you deploy to prod, etc. I'd much prefer exposing a narrow set of encodings that we can support across a wide list of versions guaranteed. I'm -1 on adding this in its current form.

tests/test_edgeql_functions.py
elprans: Done.

docs/stdlib/bytes.rst
elprans: Are you suggesting we only support utf-8?  Because I'm not sure which criteria to apply when considering which encodings are "blessed" then.

1st1: The ones supported by all versions of Postgres we support? I assume it's unlikely for Postgres to drop an encoding they already support in some future version?

1st1: We can gradually add more encodings in major versions as we shift the window of supported postgreses. But it would be weird if some encoding works in prod and doesn't in local, or vice versa. We have to somehow control the list of encodings we support and document.

elprans: I decided to just omit the whole encoding thing for now.

PR #5961, opened 2023-08-25
Add anyobject pseudo type

aljazerzen:
This is needed for fts::search. It mimics implementation of `anytype`.

Personally I don't think this is good language design, though. We are basically implementing generics with constraints and now adding new pseudo types for different constraints we need.

With this PR, we have two such constraints: anytuple and anyobject (anytype is just plain generic type argument with no constraints).

Proper generics would look like:

```
create function fts::search<T: schema::ObjectType>(
     object: T,
     ...
) -> Tuple<
    object: T,
    ...
>
```


edb/edgeql/ast.py
```diff
@@ -184,6 +184,10 @@ class AnyTuple(PseudoObjectRef):
     pass
 
 
+class AnyObject(PseudoObjectRef):
```
msullivan, line 187: Maybe we should merge `anytype/anyobject/anytuple` into one node for these things?

msullivan (approved):
I agree that this design isn't wonderful.

I think one of the reasons why polymorphic functions aren't user-definable is because our polymorphism system is so half-assed and we'd like to do better eventually

aljazerzen:
> polymorphic functions aren't user-definable

Oh, that is good news - it means we can change it without breaking user code, right? No one would depend on the reflection of our functions?

edb/edgeql/ast.py
aljazerzen: I've merged the AST nodes and this cleans the code up a bit already.

Something similar could probably be also done on `s_types.Type` methods (`is_any`, `is_anyobject`, `is_anytuple` & friends).

1st1:
> No one would depend on the reflection of our functions?

Query builders (ours own TypeScript one) do. But that should be fine.

> Proper generics would look like:

Yeah, we know. The problem is that postgres polymorphism is very limited and we're basically building around its limitations. We might be able to do better eventually, deprecating all of the `any*` pseudotypes, but that's far into the future.

aljazerzen:
> The problem is that postgres polymorphism is very limited

Ah, I see. Our functions mostly compile to Postgres functions, so we are inheriting those limitations.

PR #5962, opened 2023-08-25
wip Move provider-specific config to config object

scotttrinh:
Instead of having `{provider_name}_id` and `{provider_name}_secret`, make a proper `ClientConfig` type. I originally thought about it being exclusive on the URL, but then remembered @fantix's use case of having multiple "apps" using the same auth system to have separate GitHub apps, for instance. That means the client will need to know the _id_ of this object in order to properly select the right ID and secret. I'm not 100% sure this is going to work, but thought it would be useful to discuss over a concrete PR.


edb/server/protocol/auth_ext/http.py
```diff
@@ -57,14 +57,11 @@ async def handle_request(
         try:
             match args:
                 case ("authorize",):
-                    # TODO: this is ambiguous whether it's a name or ID which
-                    # is useful now, but we'll need to pivot to ID sooner than
-                    # later and then rename all of this to provider_id
-                    provider = _get_search_param(
+                    provider_id = _get_search_param(
                         request.url.query.decode("ascii"), "provider"
```
scotttrinh, line 61: Should I also rename this query param to `provider_id`?

edb/server/protocol/auth_ext/http.py
```diff
@@ -78,11 +75,11 @@ async def handle_request(
                     query = request.url.query.decode("ascii")
                     state = _get_search_param(query, "state")
                     code = _get_search_param(query, "code")
-                    provider = self._get_from_claims(state, "provider")
+                    provider_id = self._get_from_claims(state, "provider")
```
scotttrinh, line 78: ditto about maybe renaming?

edb/server/protocol/auth_ext/util.py
```diff
@@ -76,3 +82,15 @@ def get_config(
             description="Missing configuration value",
         )
     return value
+
+
+def get_config_unchecked(
```
scotttrinh, line 87: I didn't end up using this, but seems like a nice function to have anyway, so I left it. 🤷

tests/test_http_ext_auth.py
```diff
@@ -152,10 +152,27 @@ def __exit__(self, *exc):
...
-        ("ext::auth::AuthConfig::github_client_id", f"<str>'{uuid.uuid4()}'"),
+    EXTENSION_SETUP = [
+        f"""
+        CONFIGURE CURRENT DATABASE SET
+        ext::auth::AuthConfig::auth_signing_key := <str>'{'a' * 32}';
+        """,
+        (
+            """
+            CONFIGURE CURRENT DATABASE
+            INSERT ext::auth::ClientConfig {
```
elprans, line 163: No need to split the docstring like that, curly braces can be quoted with another curly brace: `{{` and `}}`

PR #5963, opened 2023-08-25
Implement base64 encoding and decoding functions

elprans:
This adds the new `std::encoding::base64_encode` and
`std::encoding::base64_decode` functions to encode to and decode from
base64-encoded strings.  There is support for RFC 4648 §4 standard
alphabet as well as RFC 4648 §5 URL- and filename-safe alphabet.
Padding requirement can be controlled by the `padding` named only
argument.

### DIFF
--- a/tests/test_edgeql_functions.py
+++ b/tests/test_edgeql_functions.py
@@ -17,9 +17,11 @@
 #
 
 
+import base64
 import decimal
 import json
 import os.path
+import random
 
 import edgedb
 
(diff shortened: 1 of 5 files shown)


1st1:
I like "enc" as the name of the new namespace better. Would work better with existing abbreviations like "cal"

tests/test_edgeql_functions.py
```diff
@@ -7476,3 +7478,77 @@ async def test_edgeql_functions_multirange_unpack_01(self):
                 ],
                 json_only=True,
             )
+
+    async def test_edgeql_functions_encoding_base64_fuzz(self):
```
1st1, line 7482: Please add a test to see how gracefully it fails if you try to decode a non-base64 garbage. I'm curious if stray NULLs can appear or unexpected errors.

1st1: Also test passing an empty string.

elprans: Empty strings are already tested: `random.randrange(0, 1000)`

elprans: Done.

elprans:
> I like "enc" as the name of the new namespace better. Would work better with existing abbreviations like "cal"

Changed.

1st1 (approved):

PR #5967, opened 2023-08-28
Restore source context printing in server parse errors

elprans:
This was lost in the parser rewrite.


aljazerzen (approved):
Oh, right. I forgot about this.

PR #5970, opened 2023-08-28
Postpone evaluation of function parameter defaults

elprans:
Currently, the default expression is compiled too early and might fail
if it depends on schema bits that have not been defined yet.

### DIFF
--- a/edb/schema/functions.py
+++ b/edb/schema/functions.py
@@ -195,15 +195,8 @@ class ParameterDesc(ParameterLike):
 
         paramd = None
         if astnode.default is not None:
-            defexpr = s_expr.Expression.from_ast(
+            paramd = s_expr.Expression.from_ast(
                 astnode.default, schema, modaliases, as_fragment=True)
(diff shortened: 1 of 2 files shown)


edb/schema/functions.py
```diff
@@ -516,6 +509,28 @@ def canonicalize_attributes(
...
+        track_schema_ref_exprs: bool=False,
+    ) -> s_expr.CompiledExpression:
+        if field.name == 'default':
+            return value.compiled(
+                schema=schema,
+                as_fragment=True,
+                options=qlcompiler.CompilerOptions(
+                    schema_object_context=self.get_schema_metaclass(),
+                    apply_query_rewrites=not context.stdmode,
+                    track_schema_ref_exprs=track_schema_ref_exprs,
```
aljazerzen, line 527: I'm not sure what all of these options do, but was a drop of modalises intentional?

aljazerzen (approved):

PR #5973, opened 2023-08-29
Add missing `edgedb database` commands to CLI docs

raddevon:
This started as a PR to add the `database wipe` command, but `database create` was also missing. Discovered and fixed some other issues while I was in there:

- Updated CLI connection options to include missing options and to reflect current defaults
- The commands lists were a bit haphazardly sorted, so I've sorted them all alphabetically
- Found some issues in the `migration upgrade-check` command, including an intro that was too long and a command that was mistyped
- Some other minor copy touch-ups


docs/cli/edgedb_connopts.rst
```diff
@@ -44,18 +44,27 @@ Connection flags
...
     if not set, to ``5656``.
 
+:cli:synopsis:`--unix-path /path/to/socket`
+    Specifies a path to a Unix socket for an EdgeDB connection. If the path is
+    a directory, the actual path will be computed using the ``port`` and
+    ``admin`` parameters.
+
+:cli:synopsis:`--admin`
+    Connect to a password-less Unix socket (specified by the ``unix_path`` with
+    superuser privileges by default.
```
scotttrinh, line 57: Missing an ending paren? Or maybe no parens?

```suggestion
    Connect to a password-less Unix socket. By default, specified by the
    ``unix_path`` with superuser privileges.
```

docs/cli/edgedb_database/edgedb_database_wipe.rst
```diff
@@ -0,0 +1,32 @@
...
+====================
+
+Destroy the contents of a :ref:`database <ref_datamodel_databases>`
+
+.. cli:synopsis::
+
+    edgedb database wipe [<options>]
+
+
+Description
```
scotttrinh, line 15: Do you think it's worth noting that [it's equivalent to `reset schema to initial`](https://github.com/edgedb/edgedb-cli/blob/7292db83cf64b9eeb7ef10a83ef5262141fc466c/src/commands/database.rs#L64)?

docs/cli/edgedb_database/edgedb_database_wipe.rst
```diff
@@ -0,0 +1,32 @@
...
+===========
+
+The database deleted will be one of these values: the value passed for the
+``--database``/``-d`` option, the value of ``EDGEDB_DATABASE``, or ``edgedb``.
+The contents of the database will be destroyed but the database itself will be
+preserved.
```
scotttrinh, line 21: The content is removed and the schema itself is also reset to the initial state before migrations. Seems somewhat like deleting and recreating a database with the same name, although I'm not 100% sure what other side effects there are.

docs/cli/edgedb_project/index.rst
```diff
@@ -15,18 +15,18 @@ an :ref:`ref_reference_edgedb_toml` file in the project directory.
...
     :class: funcoptable
 
     * - :ref:`ref_cli_edgedb_project_init`
       - Initialize a new or existing project
-    * - :ref:`ref_cli_edgedb_project_unlink`
-      - Clean-up the project configuration
     * - :ref:`ref_cli_edgedb_project_info`
       - Get various metadata about the project
+    * - :ref:`ref_cli_edgedb_project_unlink`
+      - Clean-up the project configuration
```
scotttrinh, line 30: I know this just moved but I find "clean-up" a confusing description of what `unlink` does here. Not a blocker, but even something redundant like "Unlinks project from currently linked database" seems better than this. WDYT?

docs/cli/edgedb_connopts.rst
raddevon: Good catch! I think I like the parenthetical here so I went ahead and closed it.

docs/cli/edgedb_database/edgedb_database_wipe.rst
raddevon: Added that the schema is also reset.

docs/cli/edgedb_project/index.rst
raddevon: No reason I can't update it while I'm in here! I took inspiration from the one-liner on that command's page since a lot of these are at least inspired by those brief descriptions anyway. Gives it good symmetry.

scotttrinh (approved):

PR #5974, opened 2023-08-29
Add migrations recovery documentation

raddevon:
This is the PR that started me down the rabbit hole that got me to #5973.

It will become obsolete if we add a CLI command for this as suggested in https://github.com/edgedb/edgedb-cli/issues/270, but I thought it would be good to document the current process since it comes up from time to time.

### DIFF
--- a/docs/guides/migrations/recovering.rst
+++ b/docs/guides/migrations/recovering.rst
@@ -50,7 +50,8 @@ or if this is the first migration:
       <script>
     };
 
-Replace ``<name>`` and ``<script>`` with the values from your ``Migration`` query results.
+Replace ``<name>``, ``<script>``, and ``<parent.name>`` with the values from
+your ``Migration`` query results.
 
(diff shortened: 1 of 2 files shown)


docs/guides/migrations/recovering.rst
```diff
@@ -0,0 +1,59 @@
...
+
+.. code-block:: edgeql
+
+    CREATE MIGRATION <name>
+        ONTO initial
+    {
+      <script>
+    };
+
+Just slot in the values from your ``Migration`` query results.
```
scotttrinh, line 53: nit: No "just", and maybe be more explicit about what to replace here?

```suggestion
Replace ``<name>`` and ``<script>`` with the values from your ``Migration`` query results.
```

raddevon: Good suggestion. Fixed the line length and also added `<parent.name>`.

PR #5977, opened 2023-08-29
Add JWT parsing functions to ext::auth

elprans:
With this, setting the `ext::auth::client_token` global will yield an
identity via the `ext::auth::ClientTokenIdentity` global if the token is
valid.

### DIFF
--- a/edb/lib/ext/auth.edgeql
+++ b/edb/lib/ext/auth.edgeql
@@ -93,4 +93,119 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
             set default := <std::duration>'336 hours';
         };
     };
+
+    create scalar type ext::auth::JWTAlgo extending enum<RS256, HS256>;
+
+    create function ext::auth::_jwt_check_signature(
+        jwt: tuple<header: std::str, payload: std::str, signature: std::str>,
+        key: std::str,
+        algo: ext::auth::JWTAlgo = ext::auth::JWTAlgo.RS256,
+    ) -> std::json
+    {
+        set volatility := 'Stable';
+        using (
+            with
+                module ext::auth,
+                msg := jwt.header ++ "." ++ jwt.payload,
+                hash := (
+                    "sha256" if algo = JWTAlgo.RS256 or algo = JWTAlgo.HS256
+                    else <str>std::assert(
+                        false, message := "unsupported JWT algo")
+                ),
+            select
+                std::to_json(
+                    std::to_str(
+                        std::enc::base64_decode(
+                            jwt.payload,
+                            padding := false,
+                            alphabet := std::enc::Base64Alphabet.urlsafe,
+                        ),
+                    ),
+                )
+            order by
+                assert(
+                    std::enc::base64_encode(
+                        ext::pgcrypto::hmac(msg, key, hash),
+                        padding := false,
+                        alphabet := std::enc::Base64Alphabet.urlsafe,
+                    ) = jwt.signature,
+                    message := "JWT signature mismatch",
+                )
+        );
+    };
+
+    create function ext::auth::_jwt_parse(
+        token: std::str,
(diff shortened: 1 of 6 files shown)


edb/lib/ext/auth.edgeql
```diff
@@ -48,4 +48,114 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
...
+                        false, message := "unsupported JWT algo")
+                ),
+            select
+                std::to_json(
+                    std::to_str(
+                        std::enc::base64_decode(
+                            jwt.payload,
+                            padding := false,
+                            alphabet := std::enc::Base64Alphabet.urlsafe,
+                        ),
```
scotttrinh, line 77: This is lovely ❤️‍🔥

edb/lib/ext/auth.edgeql
```diff
@@ -48,4 +48,114 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
...
+                )
+            order by
+                assert(
+                    std::enc::base64_encode(
+                        ext::pgcrypto::hmac(msg, key, hash),
+                        padding := false,
+                        alphabet := std::enc::Base64Alphabet.urlsafe,
+                    ) = jwt.signature,
+                    message := "JWT signature mismatch",
+                )
```
scotttrinh, line 88: Out of curiosity, why are the asserts in the `order by` clause?

scotttrinh (approved):
This is really great! Reads well and easy to understand. 🏆

scotttrinh:
> With this, setting the ext::auth::client_token global will yield an identity via the ext::auth::ClientTokenIdentity global if the token is valid.

What happens when there is no `client_token` set is it just an empty set?

edb/lib/ext/auth.edgeql
raddevon: It makes sure the assertion only happens on objects that pass your filter.

edb/lib/ext/auth.edgeql
```diff
@@ -48,4 +48,114 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
...
+                    t := ext::auth::_jwt_check_signature(
+                        ext::auth::_jwt_parse(token),
+                        key,
+                        algo,
+                    ),
+                },
+                validity_range := std::range(
+                    std::to_datetime(<float64>json_get(jwt.t, "nbf")),
+                    std::to_datetime(<float64>json_get(jwt.t, "exp")),
+                ),
```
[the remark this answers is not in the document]
elprans: That's fine, it'll be an unbounded range in that case.  If neither is set, then it's an infinite range which will pass the check.

elprans:
> What happens when there is no `client_token` set is it just an empty set?

Yes

elprans:
The test failure will go away once this is rebased on master.

edb/lib/ext/auth.edgeql
elprans: Semantically it makes more sense to put assertions in `filter`, but that affects result cardinality inference, so it's in `order by`.  Assertions need to be placed in a way that makes sure they actually get executed (just putting them in a `with` and not using the alias will not work).

tests/test_http_ext_auth.py
```diff
@@ -171,6 +171,29 @@ class TestHttpExtAuth(tb.ExtAuthTestCase):
...
 
+    @classmethod
+    def get_setup_script(cls):
+        res = super().get_setup_script()
+
+        import os.path
+
+        # HACK: As a debugging cycle hack, when RELOAD is true, we reload the
+        # extension package from the file, so we can test without a bootstrap.
+        RELOAD = True
```
msullivan, line 182: Mark it False

msullivan:
Are we ready to merge this?

PR #5981, opened 2023-08-30
Explain link properties cannot be links

raddevon:
This question comes up frequently

### DIFF
--- a/docs/datamodel/links.rst
+++ b/docs/datamodel/links.rst
@@ -395,6 +395,37 @@ the *nature/strength* of the relationship.
       }
     }
 
+.. note::
+
+    The divide between "link" and "property" is important when it comes to
+    understanding what link properties can do. They are link **properties**,
+    not link **links**. This means link properties can contain only primitive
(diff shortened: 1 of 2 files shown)


scotttrinh (commented):
Changes look good, but it looks like we need some way to indicate that the SDL is invalid in this example.

raddevon:
Thanks, @scotttrinh. I'm checking to see if we already have something for this. If not, I might need to change this up a bit. I think I can just make it a code block without syntax highlighting to get around it if need be.

raddevon:
We don't currently have a way to do this. Talked to @jaclarke who suspected I could just use `sdl-invalid` now to avoid the block being parsed and we could circle back later to make sure those blocks would get highlighting, but it seems even `sdl-invalid` still gets parsed. I've used bare code blocks instead and filed issues to add this feature:

- https://github.com/edgedb/edgedb/issues/5982
- https://github.com/edgedb/edgedb.com/issues/689

scotttrinh (approved):

Issue #5922, opened 2023-08-17
`test_edgeql_explain_ranges_contains_03` is flaky


fantix:
Occasionally we'll meet:

```
FAIL: test_edgeql_explain_ranges_contains_03 (test_edgeql_explain.TestEdgeQLExplain.test_edgeql_explain_ranges_contains_03)
--------------------------------------------------------------------------------
Py_HashSecret: 54e98b67da9764fde81d641a2d39dda55956a050a37ee029
random.seed(): f3cbefd753e1987b
--------------------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/runner/work/edgedb/edgedb/edb/testbase/server.py", line 154, in wrapper
    self.loop.run_until_complete(
  File "/opt/hostedtoolcache/Python/3.11.3/x64/lib/python3.11/asyncio/base_events.py", line [653](https://github.com/edgedb/edgedb/actions/runs/5887339910/job/15966638814#step:17:654), in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/home/runner/work/edgedb/edgedb/tests/test_edgeql_explain.py", line 1394, in test_edgeql_explain_ranges_contains_03
    self.assert_plan(
  File "/home/runner/work/edgedb/edgedb/tests/test_edgeql_explain.py", line 47, in assert_plan
    assert_data_shape.assert_data_shape(
  File "/home/runner/work/edgedb/edgedb/edb/common/assert_data_shape.py", line 369, in assert_data_shape
    return _assert_generic_shape((), data, shape)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/edgedb/edgedb/edb/common/assert_data_shape.py", line 329, in _assert_generic_shape
    return _assert_dict_shape(path, data, shape)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/edgedb/edgedb/edb/common/assert_data_shape.py", line 109, in _assert_dict_shape
    _assert_generic_shape(path + (f'["{sk}"]',), data[sk], sv)
  File "/home/runner/work/edgedb/edgedb/edb/common/assert_data_shape.py", line 325, in _assert_generic_shape
    return _assert_list_shape(path, data, shape)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #5991, opened 2023-08-30
Adjust explain tests to be less flaky.

vpetrovykh:
There are tests that are intended to trigger indexes. It matters that the specific indexes get used, but not necessarily the details of how they get used, so whether the plan uses IndexScan or BitmapHeapScan is secondary. Thus these tests should be flexible to allow either one of those outcomes as valid.

Fixes: #5922


msullivan (approved):
Great! More correct and simpler.

PR #5995, opened 2023-08-31
Add failure notification

raddevon:
I don't love this, but I can't find a way to say "needs all." It looks like you just have to list each job individually. The only alternative I'm aware of is to generate the file. Not sure it's worth that, although we will need to babysit this if we add new jobs.

Uses this action: https://github.com/Gamesight/slack-workflow-status


raddevon:
Fantix alerted me that this file _is_ actually generated. I'll take another crack at it so maybe we don't need to hard-code all these job names.

PR #5998, opened 2023-08-31
Document libpq 14 required to connect

raddevon:


fantix (commented):
Ah, SNI is only for the Cloud - it should be fine for non-Cloud instances.

fantix (approved):

PR #5999, opened 2023-08-31
Handle auth errors in callback route

scotttrinh:
Adds some handling for when auth fails and returns an `error` and (optional) `error_description`. In this case, we still redirect you back to the app, but we pass through the `error` and `error_description`.

### DIFF
--- a/tests/test_http_ext_auth.py
+++ b/tests/test_http_ext_auth.py
@@ -403,9 +403,7 @@ class TestHttpExtAuth(tb.ExtAuthTestCase):
                 """
             )
 
-            expires_at = now + datetime.timedelta(
-                minutes=5
-            )
+            expires_at = now + datetime.timedelta(minutes=5)
             state_claims = {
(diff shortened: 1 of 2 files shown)


tests/test_http_ext_auth.py
```diff
@@ -187,30 +187,55 @@ def http_con_send_request(self, *args, headers=None, **kwargs):
...
+            """,
+            provider_name,
+        )
+
+    async def get_auth_config_value(self, key: str):
+        return await self.con.query_single(
+            f"""
+            SELECT assert_single(
+                cfg::Config.extensions[is ext::auth::AuthConfig]
+                    .{key}
```
scotttrinh, line 206: Is there a way to parameterize this in the query rather than interpolating the value into the query string itself?

elprans: No as that would make the result type undecidable.  All queries must type check at compile time.

PR #6000, opened 2023-08-31
Fix collisions between function parameters named id and the property

### LINKED ISSUE #5935: Parameter name conflicting with property name
- EdgeDB Version: 3.2+75dd258
- EdgeDB CLI Version: EdgeDB CLI 3.4.0+97cad0e
- OS Version: Mac OS 13.0.1 (22A400)

Steps to Reproduce:

Schema:
```module default {
    type User{ 
    }
    function getUser(id:uuid) -> set of User {
        using(
            select User filter .id = id
        )
    };
    function getUserStr(id:str) -> set of User {
        using(
            select User filter .id = <uuid>id
        )
    };
    function getUserById(idx:uuid) -> set of User {
        using(
            select User filter .id = idx
        )
    };
}```
edgedb> select getUser("a90edcd6-40af-11ee-8e22-4f46532b9eab");
{
  default::User {id: a90edcd6-40af-11ee-8e22-4f46532b9eab},
  default::User {id: a974e86e-40af-11ee-8e22-bffa776ab289},
  default::User {id: a9b8b292-40af-11ee-8e22-33a518703dff},
  default::User {id: a9faa440-40af-11ee-8e22-2b4b9170c10b},
}
Only getUserById works ok. getUserStr("") returns all users.

Note the difference is the parameter name.

### PR DESCRIPTION
**Pull request by msullivan**
We name function params their actual names, which can collide with the
`id` column in tables and produce bad results. It could maybe collide
with source and target in link tables too. It doesn't collide with
everything, because most columns get uuid names.

Qualify references to params with the function name.

Fixes #5935.

### DIFF
--- a/edb/pgsql/compiler/__init__.py
+++ b/edb/pgsql/compiler/__init__.py
@@ -57,7 +57,7 @@ def compile_ir_to_sql_tree(
     ignore_shapes: bool = False,
     explicit_top_cast: Optional[irast.TypeRef] = None,
     singleton_mode: bool = False,
-    use_named_params: Optional[tuple[str, ...]] = None,
+    named_param_prefix: Optional[tuple[str, ...]] = None,
     expected_cardinality_one: bool = False,
     expand_inhviews: bool = False,
     external_rvars: Optional[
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: edb/pgsql/compiler/__init__.py [hunk 1]
```diff
@@ -57,7 +57,7 @@ def compile_ir_to_sql_tree(
     ignore_shapes: bool = False,
     explicit_top_cast: Optional[irast.TypeRef] = None,
     singleton_mode: bool = False,
-    use_named_params: bool = False,
+    use_named_params: Optional[tuple[str, ...]] = None,
```
elprans, line 60: Should we rename the arg to "named_arg_prefix" or something?

elprans (approved):

Issue #5569, opened 2023-05-29
Weird staeval infinite recursion


lithdew:
- EdgeDB Version: 3.0-rc.1+23c3462 (also tested on nightly 4.0-dev.7515+6f86d18)
- EdgeDB CLI Version: 3.1.0-dev.998+d16f478
- OS Version: macOS 13.3.1 (22E261)

Steps to Reproduce:

This query used to work on certain versions of nightly (specifically 4.0-dev.7512+bca7e08). Testing on 3.0-rc.1 and the latest nightly however causes a maximum recursive depth exceeded in comparison.

```
with
  cursor := <optional tuple<
    createdAt: datetime,
    id: uuid,
  >>$cursor,
  `limit` := <optional int64>$limit ?? 10,
  notifications := (
    select Notification
    filter all({
      exists .account,
      .account.id = global currentAccountId,
      any({
        .createdAt < cursor.createdAt,
        .createdAt = cursor.createdAt and .id > cursor.id,
      }) if exists cursor else true,
    })
    order by .createdAt desc
    then .id asc
    limit `limit`
  ),
  newStakeNotifications := (
    select NewStakeNotification {
      stakerId,
      staker: {
        avatar: { storageId, etagId, type, url := generate_etag_url(.storageId, .etagId, true) },
        cover: { storageId, etagId, type, url := generate_etag_url(.storageId, .etagId, true) },
        *,
      },
      amount,
    }
  ),
  marketplaceListingFilledNotifications := (
    select MarketplaceListingFilledNotification {
      lister: {
        avatar: { storageId, etagId, type, url := generate_etag_url(.storageId, .etagId, true) },
        cover: { storageId, etagId, type, url := generate_etag_url(.storageId, .etagId, true) },
        *,
      },
      buyer: {
        avatar: { storageId, etagId, type, url := generate_etag_url(.storageId, .etagId, true) },
        cover: { storageId, etagId, type, url := generate_etag_url(.storageId, .etagId, true) },
        *,
      },
      amount,
      primarySale,
    }
  )
select notifications {
  *,
  data := assert_exists((
    select
    (
      <json>(select newStakeNotifications filter .id = notifications.id) union
      <json>(select marketplaceListingFilledNotifications filter .id = notifications.id)
    )
    limit 1
  ))
};
```

Stack trace:

```
edgedb error: InternalServerError: maximum recursion depth exceeded in comparison
[...]

PR #6009, opened 2023-09-01
Get rid of constant folding in the compiler

msullivan:
It activates approximately never, since it can't fold very many things
and we typically extract constants before running queries.

Mostly it just is some extra code that takes time to run but doesn't
do much except occasionally cause weird bugs like #5569.

### DIFF
--- a/edb/edgeql/compiler/expr.py
+++ b/edb/edgeql/compiler/expr.py
@@ -27,7 +27,6 @@ from typing import *
 from edb import errors
 
 from edb.common import context as ctx_utils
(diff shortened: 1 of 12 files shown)


msullivan:
A bunch of the multiplicity tests depended on us collapsing set literals down to ConstantSets, and I didn't feel like breaking them, so I added a new path to build ConstantSet literals, and extended it to work with parameters, so now it is kind of useful again.

We should maybe drop the multiplicity inference on constant set literals, though, actually.

aljazerzen (approved):

PR #6013, opened 2023-09-01
Add Google, Azure/MS, Apple OIDC providers

scotttrinh:
Support Google as an OAuth provider using their [OpenID Connect authorization flow.](https://developers.google.com/identity/openid-connect/openid-connect)

A few things of note:

- Abstracted the common OpenID Connect functionality into it's own base provider.
- Added a caching transport (https://github.com/obendidi/httpx-cache) to avoid refetching the very cacheable contents of the discovery endpoint and JWKS endpoint.
- Updated the `get_code` function to be async even though in the normal OAuth flow, it doesn't need it. Didn't feel worth caring about the Coroutine overhead to try to do something more specialized like an Async Sentinel or something.

### DIFF
--- a/edb/server/protocol/auth_ext/base.py
+++ b/edb/server/protocol/auth_ext/base.py
@@ -89,7 +89,6 @@ class OpenIDProvider(BaseProvider):
                 token_endpoint.path,
                 json=data,
             )
-            from edb.common import markup
             token = resp.json()["id_token"]
 
             return token
--- a/edb/server/protocol/auth_ext/base.py
+++ b/edb/server/protocol/auth_ext/base.py
@@ -20,6 +20,7 @@ import uuid
 import urllib.parse
 import json
 
+from typing import Union
 from jwcrypto import jwt, jwk
 from datetime import datetime
 
@@ -45,10 +46,12 @@ class BaseProvider:
     async def get_code_url(self, state: str, redirect_uri: str) -> str:
         raise NotImplementedError
 
-    async def exchange_code(self, code: str) -> str:
+    async def exchange_code(self, code: str) -> data.OAuthAccessTokenResponse:
         raise NotImplementedError
 
-    async def fetch_user_info(self, token: str) -> data.UserInfo:
+    async def fetch_user_info(
+        self, token_response: data.OAuthAccessTokenResponse
+    ) -> data.UserInfo:
         raise NotImplementedError
 
     def _maybe_isoformat_to_timestamp(self, value: str | None) -> float | None:
@@ -72,14 +75,10 @@ class OpenIDProvider(BaseProvider):
         encoded = urllib.parse.urlencode(params)
         return f"{oidc_config.authorization_endpoint}?{encoded}"
 
-    async def exchange_code(self, code: str) -> str:
+    async def exchange_code(
+        self, code: str
+    ) -> data.OpenIDConnectAccessTokenResponse:
         oidc_config = await self._get_oidc_config()
-        data = {
-            "grant_type": "authorization_code",
-            "code": code,
-            "client_id": self.client_id,
-            "client_secret": self.client_secret,
-        }
 
         token_endpoint = urllib.parse.urlparse(oidc_config.token_endpoint)
         async with self.http_factory(
@@ -87,13 +86,29 @@ class OpenIDProvider(BaseProvider):
         ) as client:
             resp = await client.post(
                 token_endpoint.path,
-                json=data,
+                json={
+                    "grant_type": "authorization_code",
+                    "code": code,
+                    "client_id": self.client_id,
+                    "client_secret": self.client_secret,
+                },
             )
-            token = resp.json()["id_token"]
+            json = resp.json()
 
-            return token
+            return data.OpenIDConnectAccessTokenResponse(**json)
 
-    async def fetch_user_info(self, token: str) -> data.UserInfo:
+    async def fetch_user_info(
+        self, token_response: data.OAuthAccessTokenResponse
+    ) -> data.UserInfo:
+        if not isinstance(
+            token_response, data.OpenIDConnectAccessTokenResponse
+        ):
+            raise TypeError(
+                "token_response must be of type "
+                "OpenIDConnectAccessTokenResponse"
+            )
+        id_token = token_response.id_token
+
         # Retrieve JWK Set
         oidc_config = await self._get_oidc_config()
         jwks_uri = urllib.parse.urlparse(oidc_config.jwks_uri)
@@ -104,7 +119,7 @@ class OpenIDProvider(BaseProvider):
 
         # Load the token as a JWT object and verify it directly
         jwk_set = jwk.JWKSet.from_json(r.text)
-        id_token_verified = jwt.JWT(key=jwk_set, jwt=token)
+        id_token_verified = jwt.JWT(key=jwk_set, jwt=id_token)
         payload = json.loads(id_token_verified.claims)
         if payload["iss"] != self.issuer_url:
             raise errors.InvalidData("Invalid value for iss in id_token")
--- a/edb/server/protocol/auth_ext/base.py
+++ b/edb/server/protocol/auth_ext/base.py
@@ -37,7 +37,7 @@ class BaseProvider:
         self.client_secret = client_secret
         self.http_factory = http_factory
 
-    def get_code_url(self, state: str, redirect_uri: str) -> str:
+    async def get_code_url(self, state: str, redirect_uri: str) -> str:
         raise NotImplementedError
 
     async def exchange_code(self, code: str) -> str:
--- a/edb/server/protocol/auth_ext/base.py
+++ b/edb/server/protocol/auth_ext/base.py
@@ -116,4 +116,7 @@ class OpenIDProvider(BaseProvider):
         )
 
     async def _get_oidc_config(self):
-        raise NotImplementedError
+        client = self.http_factory(base_url=self.issuer_url)
+        response = await client.get('/.well-known/openid-configuration')
+        config = response.json()
+        return data.OpenIDConfig(**config)
--- a/edb/server/protocol/auth_ext/base.py
+++ b/edb/server/protocol/auth_ext/base.py
@@ -20,7 +20,6 @@ import uuid
 import urllib.parse
 import json
 
-from typing import Union
 from jwcrypto import jwt, jwk
 from datetime import datetime
 
--- a/edb/server/protocol/auth_ext/base.py
+++ b/edb/server/protocol/auth_ext/base.py
@@ -17,7 +17,7 @@
 #
 
 import uuid
-import urllib
+import urllib.parse
 import json
 
 from jwcrypto import jwt, jwk
@@ -82,11 +82,14 @@ class OpenIDProvider(BaseProvider):
         }
 
         token_endpoint = urllib.parse.urlparse(oidc_config.token_endpoint)
-        async with self.http_factory(base_url=token_endpoint.netloc) as client:
+        async with self.http_factory(
+            base_url=f"{token_endpoint.scheme}://{token_endpoint.netloc}"
+        ) as client:
             resp = await client.post(
                 token_endpoint.path,
                 json=data,
             )
+            from edb.common import markup
             token = resp.json()["id_token"]
 
             return token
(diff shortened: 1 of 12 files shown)


edb/server/protocol/auth_ext/base.py
```diff
@@ -48,3 +53,73 @@ async def fetch_user_info(self, token: str) -> data.UserInfo:
...
+
+        token_endpoint = urllib.parse.urlparse(oidc_config.token_endpoint)
+        async with self.http_factory(
+            base_url=f"{token_endpoint.scheme}://{token_endpoint.netloc}"
+        ) as client:
+            resp = await client.post(
+                token_endpoint.path,
+                json=data,
+            )
+            token = resp.json()["id_token"]
```
scotttrinh, line 92: I'm a little meh about this, but basically:

OpenID connect returns an `id_token` which contains the data we care about for the "user info" already. So instead of returning the authorization token and making a request to the user info endpoint, I am just returning the `id_token` and extracting the user info from the claims.

I considered refactoring this to return a tuple of both tokens, does that seem cleaner? I feel like we might want that once we want to start returning the authorization token to the user, but I didn't want to prematurely complicate the code if there was another/better way to approach this.

edb/server/protocol/auth_ext/data.py
```diff
@@ -80,3 +80,69 @@ def __repr__(self) -> str:
             f"iss={self.iss!r} "
             f"email={self.email!r})"
         )
+
+
+@dataclasses.dataclass
+class OpenIDConfig:
```
scotttrinh, line 86: Is this dataclass too fat? We only care about a few of these properties, maybe I should just list those and ignore the others?

edb/server/protocol/auth_ext/base.py
scotttrinh: Alright, I decided that I can't live with this how it is, so refactored it a bunch in [`d995fef` (#6013)](https://github.com/edgedb/edgedb/pull/6013/commits/d995fef386e7e36027f67a7fce899e882b8ec714)

edb/server/protocol/auth_ext/base.py
```diff
@@ -37,14 +43,98 @@ def __init__(
...
+    async def fetch_user_info(
+        self, token_response: data.OAuthAccessTokenResponse
+    ) -> data.UserInfo:
+        if not isinstance(
+            token_response, data.OpenIDConnectAccessTokenResponse
+        ):
+            raise TypeError(
+                "token_response must be of type "
+                "OpenIDConnectAccessTokenResponse"
+            )
```
scotttrinh, line 109: I battled with ChatGPT for a while trying to figure out a way to restrict this at compile time such that callers would need to pass `OpenIDConnectAccessTokenResponse` here (which in practice we do since we just do `exchange_code -> fetch_user_info`), but this is all I could come up with. Open to suggestions on a better way to do this!

edb/server/protocol/auth_ext/data.py
```diff
@@ -80,3 +80,126 @@ def __repr__(self) -> str:
...
+        return self.access_token
+
+    def __repr__(self) -> str:
+        return (
+            f"{self.__class__.__name__}("
+            f"access_token={self.access_token!r}, "
+            f"token_type={self.token_type!r}, "
+            f"expires_in={self.expires_in!r}, "
+            f"refresh_token={self.refresh_token!r})"
+        )
```
scotttrinh, line 177: Any best practices about `__str__` and `__repr__` for things that contain potentially sensitive data like this?

edb/server/protocol/auth_ext/base.py
```diff
@@ -37,14 +43,98 @@ def __init__(
...
+        jwks_uri = urllib.parse.urlparse(oidc_config.jwks_uri)
+        async with self.http_factory(
+            base_url=f"{jwks_uri.scheme}://{jwks_uri.netloc}"
+        ) as client:
+            r = await client.get(jwks_uri.path)
+
+        # Load the token as a JWT object and verify it directly
+        jwk_set = jwk.JWKSet.from_json(r.text)
+        id_token_verified = jwt.JWT(key=jwk_set, jwt=id_token)
+        payload = json.loads(id_token_verified.claims)
```
fantix, line 123: We should wrap the possible parsing errors here instead of propagating and triggering ISEs. The same for the following key references.

edb/server/protocol/auth_ext/data.py
```diff
@@ -80,3 +80,126 @@ def __repr__(self) -> str:
...
+    claims_parameter_supported: Optional[bool]
+    request_parameter_supported: Optional[bool]
+    request_uri_parameter_supported: Optional[bool]
+    require_request_uri_registration: Optional[bool]
+    op_policy_uri: Optional[str]
+    op_tos_uri: Optional[str]
+
+    def __init__(self, **kwargs):
+        for field in dataclasses.fields(self):
+            setattr(self, field.name, kwargs.get(field.name))
```
fantix, line 132: We could probably store the unused remaining values from `kwargs` into a single dict field for debugging purposes.

edb/server/protocol/auth_ext/data.py
fantix: Yeah, we can always add them back when needed, right?

edb/server/protocol/auth_ext/base.py
```diff
@@ -37,14 +43,98 @@ def __init__(
...
     def _maybe_isoformat_to_timestamp(self, value: str | None) -> float | None:
         return datetime.fromisoformat(value).timestamp() if value else None
+
+
+class OpenIDProvider(BaseProvider):
+    def __init__(self, name: str, issuer_url: str, *args, **kwargs):
+        super().__init__(name, issuer_url, *args, **kwargs)
+
+    async def get_code_url(self, state: str, redirect_uri: str) -> str:
+        oidc_config = await self._get_oidc_config()
```
fantix, line 66: My OIDC knowledge is a bit rusty here - are we supposed to cache the OIDC config for some time?

edb/server/protocol/auth_ext/data.py
```diff
@@ -80,3 +80,126 @@ def __repr__(self) -> str:
...
+
+    def __str__(self) -> str:
+        return self.issuer
```
fantix, line 135: What is this for?

edb/server/protocol/auth_ext/data.py
```diff
@@ -80,3 +80,126 @@ def __repr__(self) -> str:
...
+
+    def __repr__(self) -> str:
```
fantix, line 137: Isn't `@dataclass` providing something similar here?

edb/server/protocol/auth_ext/data.py
fantix: Let's mask the secrets here if `EDGEDB_DEBUG_EXT_AUTH` (`edb.common.debug.flags.ext_auth`) is set - so that we can safely log the response object when needed.

edb/server/protocol/auth_ext/base.py
fantix: Ah, nvm, now I see you're using `httpx_cache` below.

edb/server/protocol/auth_ext/oauth.py
```diff
@@ -34,7 +35,8 @@ def __init__(
         if edgedb_test_url:
             self.edgedb_orig_base_url = urllib.parse.quote(base_url, safe='')
             base_url = edgedb_test_url
-        super().__init__(*args, base_url=base_url, **kwargs)
+        cache = httpx_cache.AsyncCacheControlTransport()
```
fantix, line 38: If I understand your code correctly, before we share client instances among requests, this cache is only available within the handling of one user request, right? Like we may call get_oidc_config() multiple times but make an actual HTTP call only once.

This is fine now, but when we want to cache the OIDC config for a longer time (probably by sharing the same client instance), I'm not sure if `httpx_cache` is sufficient - HTTP cache is rather complicated, according to my likely outdated knowledge. For example, I don't see the use of "ETag" or "HEAD request" in `httpx_cache`.

scotttrinh: > If I understand your code correctly, before we share client instances among requests, this cache is only available within the handling of one user request, right? Like we may call get_oidc_config() multiple times but make an actual HTTP call only once.

I honestly don't know, I'm glad you asked! I was hoping the cache was shared by every client that uses the `http_factory`, but if that doesn't seem to be the case here, we're not getting the benefit I was hoping for with the cache. The `http_factory` related stuff was introduced by @1st1 to help with testing external HTTP requests, so maybe Yury can shed some light here?

> I'm not sure if httpx_cache is sufficient - HTTP cache is rather complicated, according to my likely outdated knowledge. For example, I don't see the use of "ETag" or "HEAD request" in httpx_cache.

Yeah, I was also surprised that `httpx_cache` didn't seem to leverage `etag`, but it _does_ respect the cache-control headers and correctly handle caching based on HTTP verb and status code which is better than a more naive LRU cache. Worst case scenario, it downloads duplicate data more than it might if there was a working etag system, so doesn't seem like a hard blocker.

edb/server/protocol/auth_ext/data.py
scotttrinh: It's likely to be unique amongst instances of this same class, yeah? And if two objects _do_ have the same `issuer`, they likely represent the same resource. Seemed like the correct value to stand in for the `__str__` value.

edb/server/protocol/auth_ext/data.py
scotttrinh: Ahh, true. And they even have some facilities to exclude certain fields. Will fix!

edb/server/protocol/auth_ext/data.py
scotttrinh: I wonder if it's even worth having a nice `__repr__` for these classes since the only non-sensitive data is pretty useless for disambiguation (`token_type`, `expires_in`). WDYT?

edb/server/protocol/auth_ext/base.py
scotttrinh: > If we are sure/confident of the given type of response, I'd say using assert is also cool here as it is semantically for debugging (asserts are not run under python -O, though I don't think we are using that option).

The thing I was hoping to find was some way for the type system to catch if we accidentally broke the type of `exchange_code` in the OpenID Connect providers (e.g. by accidentally using `data.OAuthAccesTokenResponse(**json)`). Or if we did some other refactor that accidentally broke our contract here.

It's fine if the answer is: there is nothing the type system can do here to prevent the issue, so all solutions will be some runtime check. If so, I'll just leave what I have.

edb/server/protocol/auth_ext/data.py
fantix: True! So we can just use dataclass's repr for ext::auth debugging then. Makes sense!

scotttrinh: I'm suggesting turning off the dataclass `__repr__` via `@dataclasses.dataclass(repr=False)`. That wouldn't give us access to a nicer debug output though, even during debugging, right?

edb/server/protocol/auth_ext/base.py
fantix: I think one try/except is enough. It'll be all the fault of the provider, right? For debugging, we still have the traceback.

scotttrinh: > It'll be all the fault of the provider, right?

Yeah, I believe so. If the first function fails, it's because their `jwks_uri` isn't return a valid `JWKSet`, if the `JWT` instantiation fails it'll be due to a malformed `id_token`, and if the JSON parsing fails, it'll be due to a malformed JWT claim, which _might_ throw in the function before it since it will attempt to check expiration, etc. so it probably does some JSON parsing of it's own?

fantix: Yeah, `jwt._check_provided_claims()` decoded the claim string and abandoned the parsed value.

scotttrinh: @fantix ⚡ [`00ec539` (#6013)](https://github.com/edgedb/edgedb/pull/6013/commits/00ec5391fa86b68c669e76d49ee7a95abb849f1a)

Does this seem about right?

edb/server/protocol/auth_ext/data.py
scotttrinh: @fantix ⚡ [`d6da52d` (#6013)](https://github.com/edgedb/edgedb/pull/6013/commits/d6da52d1206ad2a8b655b0a90267754cf1e35625)

edb/server/protocol/auth_ext/data.py
scotttrinh: @fantix 👀 [`d4f6cde` (#6013)](https://github.com/edgedb/edgedb/pull/6013/commits/d4f6cde44e3db95bd2744f8e29aad258f67de953)

this is what I was thinking

edb/server/protocol/auth_ext/base.py
fantix: Awesome!

edb/server/protocol/auth_ext/data.py
fantix: Cool! It's actually okay to just `self._extra_fields = kwargs` but this is trivial, pls ignore it.

scotttrinh: Oh, no I think this is more obvious what it's doing, will fix!

PR #6014, opened 2023-09-01
Add ext::auth schema and handle auth

scotttrinh:
- Adds a bare-bones schema to the `ext::auth` extension module
- Creates `Identity` objects on successful auth
- Creates a session JWT on successful auth
- Redirect, passing any `error` or `error_description` from the authorization server to the redirect URL if auth fails

### DIFF
--- a/edb/lib/ext/auth.edgeql
+++ b/edb/lib/ext/auth.edgeql
@@ -35,6 +35,7 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
             set readonly := true;
             create annotation std::description :=
                 "ID of the auth provider";
+            create constraint exclusive;
         };
 
         create required property provider_name: std::str {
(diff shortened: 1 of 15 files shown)


edb/lib/ext/auth.edgeql
```diff
@@ -72,5 +72,11 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
                 least 32 characters long.";
             set default := "00000000000000000000000000000000";
         };
+
+        create property token_time_to_live -> std::duration {
+            create annotation std::description :=
+                'The time after which an auth token expires.';
+            set default := <std::duration>'336 hours';
```
elprans, line 79: Why was this specific default chosen?  Also, there doesn't seem to be a way to have an unexpiring token (other than setting the ttl to some obscenely large value).

scotttrinh: No reason, I just thought "2 weeks, sure!" and went with it.

> Also, there doesn't seem to be a way to have an unexpiring token (other than setting the ttl to some obscenely large value).

That's true! Open to suggestion here: is there a way to make this optional, but still have a default if you don't set it?

elprans (approved):

PR #6016, opened 2023-09-01
Fix object->json casts inside functions

### LINKED ISSUE #5938: InternalServerError: cannot cast type uuid to jsonb
EdgeDB Version: 3.2+75dd258
EdgeDB CLI Version: EdgeDB CLI 3.4.0+97cad0e
OS Version: Mac OS 13.0.1 (22A400)

Steps to Reproduce:

1. edgedb migration create succeeds
2. edgedb migrate fails with "edb.errors.InternalServerError: cannot cast type uuid to jsonb"

```
type User { 
    email:str;
}
function getUser(idx:str) -> set of json {
    using (
        select <json> (
            select User {
                email
            } filter .id = <uuid>idx
        )
    )
};
```

msullivan:
Functions are compiled with object shapes suppressed, so the generated
code was invalid.

Whenever we are switching the output_format for shape serialization
purposes, we should unsuppress object shapes. This also fixes an issue
with materialization.

Fixes #5938.

### DIFF
--- a/edb/pgsql/compiler/context.py
+++ b/edb/pgsql/compiler/context.py
@@ -536,8 +536,11 @@ def output_format(
     output_format: OutputFormat,
 ) -> Generator[None, None, None]:
     original_output_format = ctx.env.output_format
+    original_ignore_object_shapes = ctx.env.ignore_object_shapes
     ctx.env.output_format = output_format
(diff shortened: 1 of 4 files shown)


elprans (approved):
This is a nice cleanup also

PR #6028, opened 2023-09-06
Add Azure OIDC compatible OAuth provider

scotttrinh:
Add Azure/Microsoft "common" tenant OpenID Connect compatible provider.

`common` means:

> Users with both a personal Microsoft account and a work or school account from Azure AD can sign in to the application.

[Source](https://learn.microsoft.com/en-us/azure/active-directory/develop/v2-protocols-oidc#find-your-apps-openid-configuration-document-uri)

### DIFF
--- a/tests/test_http_ext_auth.py
+++ b/tests/test_http_ext_auth.py
@@ -80,6 +80,69 @@ GOOGLE_DISCOVERY_DOCUMENT = {
     "code_challenge_methods_supported": ["plain", "S256"],
 }
 
+AZURE_DISCOVERY_DOCUMENT = {
+    "token_endpoint": (
+        "https://login.microsoftonline.com/common/oauth2/v2.0/token"
+    ),
+    "token_endpoint_auth_methods_supported": [
+        "client_secret_post",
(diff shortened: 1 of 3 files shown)


tests/test_http_ext_auth.py
```diff
@@ -80,6 +80,69 @@
...
+    "subject_types_supported": ["pairwise"],
+    "id_token_signing_alg_values_supported": ["RS256"],
+    "response_types_supported": [
+        "code",
+        "id_token",
+        "code id_token",
+        "id_token token",
+    ],
+    "scopes_supported": ["openid", "profile", "email", "offline_access"],
+    "issuer": "https://login.microsoftonline.com/{tenantid}/v2.0",
```
scotttrinh, line 103: Notice the weird template string here. This is a copy-paste from the _actual_ JSON response here, so _hopefully_ we never need to use this `issuer` property and just just safely ignore it. I feel like they should be hydrating this themselves since it was requested from a specific tenant (`common`) and the rest of the URLs present here are hydrated already.

fantix (approved):

PR #6030, opened 2023-09-06
Allow optional signing key

scotttrinh:
The signing key needs to be optional to allow the extension to be enabled and the secret to be set via some UI.

While I was in here, I wanted to also update the token TTL to treat `0` as non-expiring.

### DIFF
--- a/edb/lib/ext/auth.edgeql
+++ b/edb/lib/ext/auth.edgeql
@@ -70,7 +70,6 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
             create annotation std::description :=
                 "The signing key used for auth extension. Must be at \
                 least 32 characters long.";
-            set default := "00000000000000000000000000000000";
         };
 
         create property token_time_to_live -> std::duration {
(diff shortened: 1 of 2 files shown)


edb/server/protocol/auth_ext/http.py
```diff
@@ -185,11 +185,12 @@ def _make_session_token(self, identity_id: str) -> str:
...
         expires_at = datetime.datetime.utcnow() + expires_in
 
-        claims = {
+        claims: dict[str, Any] = {
             "iss": self.base_path,
             "sub": identity_id,
-            "exp": expires_at.astimezone().timestamp(),
         }
+        if expires_in.total_seconds() != 0:
+            claims["exp"] = expires_at.astimezone().timestamp()
```
scotttrinh, line 193: I'm not sure what the easiest way to add a test case for this is since we have the extension setup in the test class itself and I don't want to lose testing the non-zero case. Any suggestions?

edb/lib/ext/auth.edgeql
```diff
@@ -70,12 +70,12 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
             create annotation std::description :=
                 "The signing key used for auth extension. Must be at \
                 least 32 characters long.";
-            set default := "00000000000000000000000000000000";
```
1st1, line 73: You never checked for this default in Python code?

scotttrinh: Nope! I was treating that as valid hoping that optional config values would make it moot at some point.

elprans (approved):

PR #6032, opened 2023-09-07
Add `PasswordCredential` type for local IdP

scotttrinh:
For cases when the server itself is the Identity Provider, we will store things like password hash, reset password token, magic link tokens, etc.

Opening up for early feedback on the model here: I am imagining creating both an `Identity` and `LocalIdentity` when you sign up.

### DIFF
--- a/edb/lib/ext/auth.edgeql
+++ b/edb/lib/ext/auth.edgeql
@@ -1,7 +1,7 @@
 #
 # This source file is part of the EdgeDB open source project.
 #
-# Copyright 2018-present MagicStack Inc. and the EdgeDB authors.
+# Copyright 2023-present MagicStack Inc. and the EdgeDB authors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -28,6 +28,11 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
         create property email: std::str;
 
         create constraint exclusive on ((.iss, .sub))
+    };
+
+    create type ext::auth::LocalIdentity {
+        create property password_hash: std::str;
+        create required link identity: ext::auth::Identity;
     };
 
     create type ext::auth::ClientConfig extending cfg::ConfigObject {
--- a/edb/lib/ext/auth.edgeql
+++ b/edb/lib/ext/auth.edgeql
@@ -30,9 +30,11 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
         create constraint exclusive on ((.iss, .sub))
     };
 
-    create type ext::auth::LocalIdentity {
-        create property password_hash: std::str;
-        create required link identity: ext::auth::Identity;
+    create type ext::auth::PasswordCredential {
+        create required property password_hash: std::str;
+        create required link identity: ext::auth::Identity {
+            create constraint exclusive;
+        };
     };
 
     create type ext::auth::ClientConfig extending cfg::ConfigObject {
--- a/edb/lib/ext/auth.edgeql
+++ b/edb/lib/ext/auth.edgeql
@@ -30,7 +30,7 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
         create constraint exclusive on ((.iss, .sub))
     };
 
-    create type ext::auth::LocalIdentity {
+    create type ext::auth::LocalIdentity extending ext::auth::Identity {
         create required property handle: std::str {
             create constraint exclusive;
         };
--- a/edb/lib/ext/auth.edgeql
+++ b/edb/lib/ext/auth.edgeql
@@ -30,9 +30,15 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
         create constraint exclusive on ((.iss, .sub))
     };
 
+    create type ext::auth::LocalIdentity {
+        create required property handle: std::str {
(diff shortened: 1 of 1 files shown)


1st1 (approved):

edb/lib/ext/auth.edgeql
```diff
@@ -30,6 +30,11 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
         create constraint exclusive on ((.iss, .sub))
     };
 
+    create type ext::auth::LocalIdentity {
+        create property password_hash: std::str;
+        create required link identity: ext::auth::Identity;
```
fantix, line 35: Do we want to make this link unique 1:1 instead of N:1? During authentication, I think we will first select the identity (corresponding to the user choosing to log in with a password), then check if the password matches. It would be weird if the schema allows multiple passwords per identity.

scotttrinh: Yeah, that makes sense, nice catch. `create constraint exclusive` right?

scotttrinh:
Based on the IRL discussion:

Instead of considering this a different kind of "identity", we want to support multiple kinds of credentials. Password being one, but other credentials like WebAuthn, magic links, recovery codes, etc. will be modeled as different types.

edb/lib/ext/auth.edgeql
```diff
@@ -30,6 +30,13 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
         create constraint exclusive on ((.iss, .sub))
     };
 
+    create type ext::auth::PasswordCredential {
```
elprans, line 33: Do we want to introduce an abstract parent to make it easier to get all credentials for an identity?

scotttrinh: Sure, good call! Maybe it also makes sense to add a computed backlink to that abstract credential type to make it even easier? Or does it make more sense to have separate backlinks to each different type of credential?

How about naming here? `LocalCredential`? `AbstractCredential`? `Credential`?

fantix: > Maybe it also makes sense to add a computed backlink in Identity to that abstract credential type to make it even easier?

This makes me think again about the link direction stuff. What will the query look like - in what case will the user have a credential object first, and then find its identity following the current link (not the proposed backlink)?

scotttrinh: The user will have their email and password, and submit that via basic auth, and the server will see you're trying to make a password authenticated login, will look up the `Identity` based on the provided `email`, and then test the password hashes. So, yeah, I think they'll have the `Identity` first.

elprans: `AbstractCredential` or `Credential` both work, go for whichever you like more.

fantix: >  I think they'll have the Identity first.

In that case, I think a single optional link from Identity to Credential is more natural?

Or rather, with the exclusive constraint, it's actually equivalent in SQL performance, users just need to write the backlink filter?

fantix (approved):

edb/lib/ext/auth.edgeql
scotttrinh: > Or rather, with the exclusive constraint, it's actually equivalent in SQL performance, users just need to write the backlink filter?

Hopefully there won't be any need for a "user" (e.g. application developer) to actually query the `*Credential` types: they'll just call the login endpoint with the credentials and the server will do the rest of the work. So the direction of the link here is something we mostly have to deal with.

Oh, also, the same `Identity` might have multiple local `Credential`s given that we'll likely implement forgot password, magic link, recovery code, MFA, etc. as separate `Credential`s for the same `Identity`.

edb/lib/ext/auth.edgeql
```diff
@@ -30,6 +30,19 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
         create constraint exclusive on ((.iss, .sub))
     };
 
+    create type ext::auth::LocalIdentity {
+        create required property handle: std::str {
```
scotttrinh, line 34: How do we feel about the term `handle` here? This property is what will be provided as the lookup for a set of credentials. So for password flow, it'll be an email or username. For something like WebAuthn, it'll be a [credential ID](https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredential/id).

Here are some alternatives:

1. login
2. key
3. identifier
4. alias
5. ??

elprans: IMO, `handle` is most neutral and unambiguous.

edb/lib/ext/auth.edgeql
```diff
@@ -30,6 +30,19 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
         create constraint exclusive on ((.iss, .sub))
     };
 
+    create type ext::auth::LocalIdentity {
```
elprans, line 33: Not extending `Identity` on purpose?

scotttrinh: Oops! Nope, an oversight.

Also, how do I provide a constant override for the `iss` property from `Identity`? Maybe it should be readonly as well?

scotttrinh: I couldn't figure out how to get the default of "local" or a constant "local", so we can just follow up with that on a separate PR.

PR #6034, opened 2023-09-08
point package server to new infra in builds

RobertBlumen:


.github/workflows.src/build.targets.yml
```diff
@@ -1,7 +1,7 @@
 publications:
     - name: prod
       suffix: ""
-      server: sftp://uploader@upload-packages.edgedb.com:2222/
+      server: sftp://uploader@package-upload.edgedb.com:22/
```
elprans, line 4:
```suggestion
      server: sftp://uploader@package-upload.edgedb.net:22/
```

elprans (approved):

Issue #5985, opened 2023-08-30
GraphQL errors when querying/mutating models in a non-default module


tomnz:
I am experiencing multiple issues while attempting to use the GraphQL query/mutation endpoints whenever a model is part of a non-`default` module. See inline below for specific issues and how to trigger them.

- EdgeDB Version: **3.2+740bb0d**
- EdgeDB CLI Version: **EdgeDB CLI 3.4.0+15c5e18**
- OS Version: **Docker image @ :latest**

## Schema

(Untested, but this _should_ reproduce the issues):

```
using extension graphql;
module default {
  type Parent {
    required single link data: Data {
      constraint exclusive;
      default := (
        insert Data {}
      }
    }

    multi link children := .<parent[is child::Child];
  }
  # Included to demonstrate that errors are not related to nested fields
  type Data {
    link parent := .<data[is Parent];
    value: str;
  }
}
module child {
  type Child {
    required single link parent: default::Parent;
    data: str;
  }
}
```

### A) Invalid `__typename` for module-scoped model in query

```
query {
  Parent(first: 1) {
    __typename
    id
    data {
      __typename
      id
      value
    }
    children {
      __typename
      id
      data
    }
  }
}
```

The response may look something like:

```
Parent: [
  {
    __typename: "Parent_Type",
    id: "...",
    data: {
      __typename: "Data_Type",  # This is correct - so the issue is not nested fields
      value: "..."
    },
    children: [
      {
        __typename: "\u0001__\u0002_Type",  # This is invalid!
        id: "...",
        data: "..."
      }
    ]
  }
]
```

The `__typename` field for `child::Child` does not resolve correctly, and will return `\u0001__\u0002_Type` for any type inside a non-`default` module. I would expect it to instead return `child__Child_Type`.

This causes issues for regular GraphQL clients, as they rely on the typename to figure out the concrete type returned for the respective interfaces that the query defines as the return type (e.g. `Parent` -> `Parent_Type`).

Note that the `__typename` works as expected for models in `default`.

### B) Unsupported inline fragment spread for module-scoped model in query

```
query {
  child__Child {
    ... on child__Child {
      id
      data
    }
  }
}
```

[...]

PR #6035, opened 2023-09-08
GQL module fixes

vpetrovykh:
Fixes to the several issues reported in #5985

### DIFF
--- a/edb/graphql/types.py
+++ b/edb/graphql/types.py
@@ -1897,7 +1897,7 @@ class GQLBaseType(metaclass=GQLTypeMeta):
                 eql = parse_fragment(f'{self.gql_typename!r}')
             else:
                 # Construct the GraphQL type name from the actual type name.
-                eql = parse_fragment(f'''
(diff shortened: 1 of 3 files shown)


edb/graphql/types.py
```diff
@@ -1504,11 +1504,20 @@ def get(self, name: str, *, dummy: bool = False) -> GQLBaseType:
...
+                    module: Union[s_name.Name, str]
+
                     if '::' in tname:
                         edb_base = self.edb_schema.get(
                             tname,
                             type=s_types.Type,
                         )
+                    elif '__' in tname:
+                        # Looks like it's coming from a specific module
+                        module, tname = tname.split('__', maxsplit=1)
```
msullivan, line 1516: Will this work right for nested modules?

msullivan (approved):
Seems reasonable but double check the nested module situation

edb/graphql/types.py
vpetrovykh: Fixed that to be consistent with nested modules.

PR #6036, opened 2023-09-08
Fix the undefined order of columns in `COPY`.

### LINKED ISSUE #5716: SQL mode COPY includes internal columns, produces columns in nondiscoverable order
```
scope=# copy "User" to stdout;
27995	898	1149	0	1149	(0,1)	Alice	f5ab708e-1534-11ee-8439-6fca9e7c1d0c	ecb718b1-1534-11ee-b27d-ab578c0a5e0c	f5ab7086-1534-11ee-8439-973fa5acc1b4
...
```
I don't think we should include `ctid` and friends. They don't get included when running `copy` on normal postgres.

Additionally, the order that the columns are reported out does not correspond to the `ordinal_position`s reported when querying `information_schema`/`pg_catalog`

It also includes `__type__`, while `information_schema` omits it. (It does so deliberately, but I think it should probably start including it?).

I think that I am indifferent to whether the column order produced by `COPY` is changed to match `ordinal_position` or vice-versa.

(Assigning to Victor because I know he's looking at another SQL thing right now, but you should feel free to kick it over to @aljazerzen if you think that makes more sense)

vpetrovykh:
Match the order of columns in the `COPY` command and introspection_schema.

Fixes #5716

### DIFF
--- a/edb/pgsql/resolver/command.py
+++ b/edb/pgsql/resolver/command.py
@@ -42,19 +42,26 @@ def resolve_CopyStmt(stmt: pgast.CopyStmt, *, ctx: Context) -> pgast.CopyStmt:
         # cannot be copied as a table, but needs to be wrapped into a
         # `SELECT * FROM view`.
         relation, table = dispatch.resolve_relation(stmt.relation, ctx=ctx)
+        col_map: Dict[str, str] = {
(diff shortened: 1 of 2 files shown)


msullivan (approved):
Looks good!

Issue #4941, opened 2023-01-26
Implement initial take on full text search: RFC 1015


msullivan:
https://github.com/edgedb/rfcs/blob/master/text/1015-full-text-search.rst

In progress in #4889

PR #6038, opened 2023-09-08
Full text search on objects

aljazerzen:
Major revamp of full text search API.

Ref #4941

This PR is ready for review, but given that it is very close to full functionality, I will add a few commits in the future. PR can also be split into smaller ones, if anyone finds self-encompassing chunk of changes.

TODO for this PR:

- [x] Basics
- [x] Inheritance - basic
- [x] fts::search
- [x] anytype
- [x] anyobject
- [x] Opaque language object and a function to construct it
- [x] Inheritance - redefining in subtypes
- [x] Error for multiple inheritance
- [x] Error for multiple FTS indexes
- [x] replace GENERATED STORED with TRIGGER
- [x] GENERATED VIRTUAL - does it hit the index?
  -> PostgreSQL currently implements only stored generated columns.
- [x] add CONSTRAINT name so we can drop the constraint and re-add it,
- [x] Multiple properties pg
- [x] Multiple properties zombo
- [x] Language per index
- [x] Language per property
- [x] Language per object
  -> does it work on zombo?
  -> No, it does not. You can specify analyzers via zdb.define_field_mapping,
  -> but you must pass literal JSON that cannot depend on row contents.
- [x] fts::search on objects without FTS indexes should return no results
- [x] ~fts::search should reject anything else than plain Object references~
- [x] Test that `select fts::search(Recipe, 'spiced').object` does not compute rank at all
- [x] test null fields
- [x] deltafts: update the fts document when the trigger is created
- [x] replace fts::with_language to fts::with_options
- [x] make language default to english
- [x] make fts::Language use ISO language codes instead of names
- [x] weights

For future PRs:

- fts::search_with_highlights() -> array<str>
- figure out what to do with separator (dynamically generated?)
- Boosting


edb/lib/fts.edgeql
```diff
@@ -19,128 +19,145 @@
...
+        Norwegian,
+        Portuguese,
+        Romanian,
+        Russian,
+        Serbian,
+        Spanish,
+        Swedish,
+        Tamil,
+        Turkish,
+        Yiddis,
```
msullivan, line 89: Did the `h` get dropped here?

edb/pgsql/compiler/relgen.py
```diff
@@ -3859,3 +3889,219 @@ def process_encoded_param(
...
+def create_subrel_for_expr(
+    expr: pgast.BaseExpr,
+    *,
+    alias: str = 'expr',
+    an_id: irast.PathId,
+    ctx: context.CompilerContextLevel
+) -> irast.PathId:
+    """
+    Creates a dummy relation that contains the given expression.
+    That expr should not use any references to other relvars, since it will be
```
msullivan, line 4079: I think that where this is used, the `expr` sometimes might have references to other rvars, if any nontrivial computation is done to produce the language or the query?

I don't think that should be a big problem, though? If we specify `lateral=True` when we create the rvar, it should be able to refer to the earlier things.

tests/test_edgeql_fts.py
```diff
@@ -19,24 +19,37 @@
 
 import os.path
 
+import edgedb
+
 from edb.testbase import server as tb
 
 
-class TestEdgeQLFTS(tb.QueryTestCase):
+class TestEdgeQLFTSQuery(tb.QueryTestCase):
```
msullivan, line 27: Add a test where the query is produced by some nontrivial computation. Concatenating some strings together might be good enough.

(I *think* it will fail, for reasons I discussed in another comment, but should be easy to fix)

msullivan (commented):
I've reviewed everything but `deltafts.py` so far.

Left a bunch of comments, but nothing major. It's looking great!

edb/schema/sources.py
```diff
@@ -154,6 +154,34 @@ def add_pointer(
...
+        """
+        Returns a list of columns that are present in the backing table of
+        this source, apart from the columns for pointers.
+        """
+        res = []
+        from edb.common import debug
+        if not debug.flags.zombodb:
+            fts_textsearch = sn.QualName('fts', 'textsearch')
+            has_fts_index = False
+            for index in self.get_indexes(schema).objects(schema):
```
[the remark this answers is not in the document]
msullivan: (If you prefer the current approach, probably still add a `break` to the loop)

msullivan:
>     * [ ]  Test that `select fts::search(Recipe, 'spiced').object` does not compute rank at all
> 
It shouldn't be too hard to do that with a [test_edgeql_sql_codegen.py](https://github.com/edgedb/edgedb/blob/master/tests/test_edgeql_sql_codegen.py) test

edb/pgsql/deltafts.py
```diff
@@ -0,0 +1,471 @@
...
+                relation=pgast.Relation(name='NEW'),
+            )
+        },
+    )
+    exprs = astutils.maybe_unpack_row(sql_res.ast)
+
+    ops = dbops.CommandGroup()
+    from edb.common import debug
+
+    if debug.flags.zombodb:
```
msullivan, line 146: I'd probably split zombo into another function unless it depends on a ludicrous number of shared variables?

edb/pgsql/deltafts.py
```diff
@@ -0,0 +1,471 @@
...
+            _raise_unsupported_language_error(unsupported)
+
+        text_sql = codegen.generate_source(expr.text)
+        language_sql = codegen.generate_source(expr.language)
+
+        document_exprs.append(
+            f'''
+            setweight(
+                to_tsvector(
+                    ({language_sql}::text)::pg_catalog.regconfig,
```
msullivan, line 376: Extra parens around `{language_sql}` maybe?

edb/pgsql/deltafts.py
```diff
@@ -0,0 +1,471 @@
...
+    raise errors.UnsupportedFeatureError(msg)
+
+
+# --- pg fts ---
+
+
+def _pg_create_trigger(
+    table_name: Tuple[str, str],
+    exprs: Sequence[pgast.BaseExpr],
+) -> dbops.Command:
```
msullivan, line 356: This needs to also update `__fts_document__` for all the existing rows, right?

msullivan: (Add a test for that case, too)

msullivan (commented):
PG side looks solid too!

edb/lib/fts.edgeql
```diff
@@ -19,128 +19,145 @@
...
 
-CREATE ABSTRACT INDEX fts::textsearch(named only language: std::str) {
+CREATE ABSTRACT INDEX fts::textsearch {
     CREATE ANNOTATION std::description :=
         "Full-text search index based on the Postgres's GIN index.";
-    SET code := 'gin (to_tsvector(__kw_language__, __col__))';
+    SET code := 'gin (__col__)';
 };
 
 CREATE SCALAR TYPE fts::searchable_str {
```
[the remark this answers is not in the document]
aljazerzen: I'll move this into `deltafts.py`

edb/pgsql/compiler/relgen.py
```diff
@@ -3859,3 +3889,219 @@ def process_encoded_param(
...
+    # create a dummy path id for a dummy object
+    el_name = sn.QualName('__dummy__', alias)
+    query_ptrref = irast.SpecialPointerRef(
+        name=el_name,
+        shortname=el_name,
+        out_source=an_id.target,
+        out_target=an_id.target,
+        out_cardinality=qltypes.Cardinality.AT_MOST_ONE,
+    )
+    expr_id = an_id.extend(ptrref=query_ptrref)
```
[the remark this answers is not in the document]
aljazerzen: Oh yeah, we talked about this, I forgot.

edb/pgsql/compiler/relgen.py
```diff
@@ -3859,3 +3889,219 @@ def process_encoded_param(
...
+                    name=('zdb', 'score'),
+                    args=[ctid]
+                )
+                where_clause = pgast.Expr(
+                    lexpr=ctid,
+                    name='==>',
+                    rexpr=query_pg,
+                )
+            else:
+                el_name = sn.QualName('__object__', '__fts_document__')
```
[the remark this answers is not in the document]
aljazerzen: There is a comment on `SpecialPointerRef`

edb/pgsql/deltafts.py
```diff
@@ -0,0 +1,471 @@
...
+        return _delete_fts_document(index, drop_index, schema, context)
+    else:
+        # effective index remains: don't drop the fts document
+
+        effective_subject = effective.get_subject(schema)
+        is_eff_on_direct_parent = effective_subject in subject.get_bases(
+            schema
+        ).objects(schema)
+
+        if is_eff_on_direct_parent:
```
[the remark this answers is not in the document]
aljazerzen:
```
type GrandParent {
   index fts::textsearch;
}
type Parent {
   index fts::textsearch;
}
type Child {
}

alter GrandParent {
   drop index fts::textsearch;
}
```

This will also drop the index that is inherited into the Child, but it will not change Child's effective index.

edb/pgsql/deltafts.py
aljazerzen: Oh, yea it does. But not `_pg_create_trigger`, `_refresh_fts_document` should.

aljazerzen: Wait, no you are right - it has to be updated when creating the column AND when refreshing.

edb/pgsql/compiler/relgen.py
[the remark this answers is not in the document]
msullivan: Yeah, but I'd still recap it here, since there's just one call site

edb/pgsql/compiler/relgen.py
aljazerzen: When expr gets complex, this subrel gets joined in with `CROSS LATERAL JOIN`. So this is not a problem, apparently.

I'll also remove the comment saying it should not work.

tests/test_edgeql_fts.py
aljazerzen: Added the test. It does not fail.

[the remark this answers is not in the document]
aljazerzen: I've added a test that has `select X limit 10` and a test that has `select X filter ...`. Both work.

TODO: make unions work. Right now, they fail with message "fts::search required fts::index on type ...".

tests/schemas/fts_setup_wip.edgeql
```diff
@@ -0,0 +1,522 @@
+start migration to {
```
elprans, line 1: Is this meant to be committed? If so, change the name from `_wip`.

aljazerzen: @vpetrovykh ?

aljazerzen:
A bit of documentation on analyzers:

FTS analyzers can be selected using an 'analyzer identifer':
- `fts::search` takes a str which is the identifier,
- `fts::with_options` takes any enum, which is cast to string and then used as the identifer,

ISO languages need the `ISO_` prefix in the identifier, because we want to allow using any custom idenifier. For example, when you want to use elastic's `cjk`, you don't want that to be interpreted as `iso_cjk`, which would be Chokwe and not ChineseJapaneseKorean.

msullivan:
Bump the catalog version

aljazerzen:
@msullivan I have trouble with making `fts::with_options` VARIADIC.

1. What's the return type? It cannot be a single `fts::Document`, since it might contain multiple fields. It should be a tuple of `fts::Document`, but I don't think we can express that.
2. Making first arg `text` VARIADIC means that we need to make following args NAMED ONLY (`language` and `weight_category`).

Since this goes on top of the problems we discussed around nesting `with_options` calls and required implicit casting / with_options overloading, I'll postpone this and get around to it after the parser stuff.

aljazerzen:
A bit of documentation of how languages work:

- `fts::with_options` takes an enum, whose value is cast to a string. That string is a *language code*.
- `fts::search` takes a string, which is treated directly as a *language code*

Mapping from *language codes* to backend text analyzers is described with this function:

```python
def determine_analyzer(language_code: str) -> Analyzer:
  # xxx_ codes are passed to backend directly
  if language_code.startswith('xxx_'):
    analyzer = language_code[4:]
    return analyzer

  # lookup in ISO 639-3
  if iso_lang := iso_langs_by_code.get(language_code):

    # on Postgres, convert `eng` to `'english'::regconfig`
    return get_backend_analyzer(iso_lang)

  # lookup in EdgeDB language codes
  if edgedb_lang := edgedb_langs_by_code.get(language_code):

    # on Elastic, convert `edb_chineese_japanese_korean` to `cjk`
    return get_backend_analyzer(edgedb_lang)

  raise get_backend_analyzer('eng')
```

At the moment, *EdgeDB language codes* do not contain any entries, but are meant to include:
```
edb_Brazilian
edb_ChineseJapaneseKorean
edb_Indian
```

----

This mapping is based on following requirements:

- languages that have an ISO 639-3 code should use that code (across all supported backends)
- other languages should use some naming defined in EdgeDB, so they work across all supported backends,
- it should be possible to pass custom values to backends, so we support access to, for example, a custom Elastic analyzer for a dialect of Slovenian,


There is still the question of what to do when a backend does not support a language. Currently this will result in an error. It would be convenient to default to some other language, as that would allow local development deployments using Postgres FTS and production deployment using languages not supported by Postgres FTS.

aljazerzen:
For reference:

ISO 639-3 codes were pulled from https://iso639-3.sil.org/code_tables/download_tables

A snippet for matching languages by their names:

```python
import csv

by_name = {}
with open('./iso-639-3_20230123.tab') as csvfile:
    for row in csv.reader(csvfile, delimiter='\t'):
        code = row[0]
        name = row[6]
        name = name.split(' (')[0]

        by_name[name] = code

languages = '''Arabic,English,German'''

for lang_name in languages.split(','):
    lang_code = by_name.get(lang_name, ' ? ')
    print(lang_code, lang_name.lower())
```

msullivan (approved):
Lets make `with_options` take NAMED ONLY arguments, and then this is good to merge!

PR #6045, opened 2023-09-11
Fix unclosed socket warning from http_ext_auth tests

msullivan:
Apparently while `HTTPServer.shutdown` knocks the serving thread out
of `serve_forever`, it does not actually close the server socket.
Explicitly call `server_close` to do that.


scotttrinh (approved):
🆒

PR #6048, opened 2023-09-12
Use an abstract Credential type

scotttrinh:
Currently fails due to some issue with dropping the extension.


scotttrinh:
Superseded by #6229

PR #6050, opened 2023-09-12
Describe what indexes do

raddevon:


docs/datamodel/indexes.rst
```diff
@@ -37,8 +49,8 @@ notation shorthand <ref_dot_notation>`: ``.name``.
       index on (.name);
     }
 
-By indexing on ``User.name``, queries that filter, order, or group by the 
-``name`` property will be faster, as the database can look up a name in 
+By indexing on ``User.name``, queries that filter, order, or group by the
+``name`` property will be faster, as the database can look up a name in
```
scotttrinh, line 53: Not sure where the best place to put this note is, but it's important to know that the query planner gets to choose whether to use the index or not. It can often be faster to skip the index until the table grows to a certain size, so it's best to profile with realistic datasets to see if an index will be helpful. My personal approach to indexes with Postgres has been to create them based on how I know I'll query the data, and even if they aren't being hit at first, in time, they ought to be.

raddevon: See what you think about these changes, @scotttrinh. I had a note early in the page about the query planner, and I've added to that note an example of when it might ignore an index. I've also tried to make sure that we never guarantee that an index will be used. I added another note where we recommend trying an index to see the performance benefit to tell the user they may note see the index being used in a small database.

PR #6055, opened 2023-09-13
Note object-level security is analogous to RLS

raddevon:
Should make it easier to find answers if users try searching for (or asking the chatbot about) "row-level security."

### DIFF
--- a/docs/datamodel/access_policies.rst
+++ b/docs/datamodel/access_policies.rst
@@ -8,7 +8,8 @@ Access Policies
 
 Object types can contain security policies that restrict the set of objects
 that can be selected, inserted, updated, or deleted by a particular query.
-This is known as *object-level security*.
+This is known as *object-level security* and it is similar in function to SQL's
+row-level security.
 
 Let's start with a simple schema without any access policies.
 


docs/datamodel/access_policies.rst
```diff
@@ -8,7 +8,8 @@ Access Policies
 
 Object types can contain security policies that restrict the set of objects
 that can be selected, inserted, updated, or deleted by a particular query.
-This is known as *object-level security*.
+This is EdgeDB's analog to SQL's row-level security, known as *object-level
+security*.
```
scotttrinh, line 12: Not a blocker but it feels a little strange to see a mention of SQL here. Do we have precedent for comparing one of our features to a SQL feature like this?

Maybe a weaker correspondence would feel better to me, like:

"This is known as *object-level security* and it is similar in function to SQL's row-level security."

Something like that? Not sure if that's any better, honestly, and this is a pretty minor quibble.

scotttrinh (approved):

docs/datamodel/access_policies.rst
Dhghomon: I've mostly only seen writing about joins but not much else such as rows, primary key and foreign key (I think that's what they're called) and so on. I'm probably one of the few people that started with EdgeDB jargon before learning some of the SQL equivalents. Could be interesting to later see a small sheet of equivalent / near-equivalent terms.

raddevon: We don't really have a precedent for doing this, but I think we should start, especially since most people using EdgeQL will probably have some familiarity with SQL concepts and are more likely to search for the names they know. (I'm considering asking for help coming up with a list of these analogs and mentioning them throughout the docs.) 

I do prefer the way you've worded it here though. I'll change it up. Thank you!

Issue #6054, opened 2023-09-13
Make ids of std schema objects stable across minor versions


msullivan:
Currently, all the ids are generated randomly at bootstrap time. We do the bootstrap as part of our release artifact build, which means that the ids of std objects in a database are determined by which version of edgedb the database was created with, not the version that is currently running.

That is, if a database is created with 3.2 and then upgraded to 3.3, it will have 3.2 ids, not 3.3 ids.

Making ids stable would have some benefits:
1. You could run a shared compiler pool that doesn't need to worry about stdschema mismatch
2. We might not need to load stdschema/reflschema from the database *at all*, we can just read them out from our cache. We could *almost* not even store the schemas in the database at all, except that we need to old schema state to apply patches!

I think this should actually be pretty easy to do; we can use uuid5s based on name+type when operating in stdmode.
We'll also want some CI tests that check:
1. Schema generation is deterministic
2. All ids from a previous minor version still exist and have the same name

PR #6058, opened 2023-09-13
Make ids of schema objects stable

msullivan:
Instead of generating them randomly, use a uuid5 generated using the
name/type of the object.

Does the implementation work of #6054, but I don't want to close that
(and we shouldn't depend on this) until we have a testing setup.

I tested this using:
```
import pickle
from edb.common import devmode
devmode.enable_dev_mode()

p1 = pickle.loads(open('backend-stdlib.pickle', mode='rb').read()[20:])
p2 = pickle.loads(open('build/cache/backend-stdlib.pickle', mode='rb').read()[20:])

mismatched = [k for k in p1.reflschema._id_to_data if p1.reflschema._id_to_data.get(k) != p2.reflschema._id_to_data.get(k)]

print("reflschema:", p1.reflschema.__dict__ == p2.reflschema.__dict__)
print("stdschema:", p1.stdschema.__dict__ == p2.stdschema.__dict__)
```

### DIFF
--- a/edb/schema/objects.py
+++ b/edb/schema/objects.py
@@ -1084,14 +1084,6 @@ class Object(s_abc.Object, ObjectContainer, metaclass=ObjectMeta):
         return hash(self.id)
 
     @classmethod
-    def generate_id(
-        cls,
-        schema: s_schema.Schema,
-        data: Dict[str, Any],
-    ) -> uuid.UUID:
-        return uuidgen.uuid1mc()
-
-    @classmethod
     def _prepare_id(
         cls,
         schema: s_schema.Schema,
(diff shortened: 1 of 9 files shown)


edb/schema/objects.py
```diff
@@ -1101,7 +1104,15 @@ def _prepare_id(
...
-            return cls.generate_id(schema, data)
+            if stdmode:
+                # When compiling the standard library, we generate
+                # stable ids based on the internal name and the type's
+                # name. This keeps std schemas compatible across
+                # minor versions at least.
+                return uuidgen.uuid5(
+                    TYPE_ID_NAMESPACE, f'{name}-{cls.__name__}')
+            else:
+                return cls.generate_id(schema, data)
```
elprans, line 1115: We don't seem to actually overload `generate_id` anywhere anymore, do you want to get rid of it while you're here?

elprans: Or, if you want to keep it for some reason, I'd push the `stdmode` thing into it.

msullivan: Yeah we'll just drop it

elprans (approved):
Needs a catver bump

PR #6061, opened 2023-09-13
Move the patch testing scripts into standalone files

msullivan:
This makes it less of a pain to test them locally


msullivan:
(I've already pushed this to 3.x)

elprans (approved):

PR #6064, opened 2023-09-13
Make patches tests able to catch busted migrations

msullivan:
Upgrading from rc.1/rc.2 to later versions leaves user schemas in a bad state, since
the `prefer_subquery_args` field is `None` on old versions and `False` on new ones.
 
To make sure we won't miss issues like this again, fix the patches tests to
create an instance of the database for the `scope` tests (probably the most complex
of the test schemas). If the schema state is corrupted, then the (now no-op) migration
when running the tests will fail.

Also fix the object-level issue by adding a repair pass that only runs if the database 
was created with rc1 or rc2. We'll maybe want to add a mechanism to properly detect
that, but since we don't have one, we have to be a bit hacky.
We could also skip the auto-repair and just document that affected users should run
`administer schema_repair()`.

### DIFF
--- a/.github/scripts/patches/create-databases.py
+++ b/.github/scripts/patches/create-databases.py
@@ -18,6 +18,23 @@ try:
     for name in ['json', 'functions', 'expressions', 'casts', 'policies', 'vector', 'scope']:
         db.execute(f'create database {name};')
 
+    # For the scope database, let's actually migration to it.  This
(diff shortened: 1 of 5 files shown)


elprans (approved):
Yeah, we want to store the original server version at bootstrap in instancedata or something like that.

PR #6067, opened 2023-09-14
Make explain tests use deterministic data.

vpetrovykh:
Deterministic dataset should ensure that the explain index tests are not flaky because they should have the same statistics to base the index optimization on.


msullivan (approved):
Let's hope it works!

PR #6072, opened 2023-09-15
Add test for wrong cardinality on polymorphic prop

scotttrinh:
Reported as a bug in the queries generator here: https://github.com/edgedb/edgedb/issues/6254


msullivan:
I'm gong to close this, but it will be included in an upcoming fix. I transfered the issue to https://github.com/edgedb/edgedb/issues/6254

PR #6073, opened 2023-09-15
Allow computed pointers on types to omit link/property kind specification

msullivan:
This requires dropping the field setting syntax from object types, but
that's mostly OK because object types don't have any fields that can
be set from SDL.
This introduces some inconsistency but it is probably acceptable.

If we want to add fields to object types in the future, probably
we'll just want to allow writing `SET foo := ...`.
Technically, id can be set on object types, but it only works
now in DDL, so there is no breakage.

The implementation is a pile of hacks, where we run creation using
only the generic Pointer code, determine whether the target is an
object, then reinterpret the original ast node with that information
and run the resulting delta. Oh well.

### DIFF
--- a/edb/edgeql/ast.py
+++ b/edb/edgeql/ast.py
@@ -1049,6 +1049,10 @@ class CreateConcreteUnknownPointer(CreateConcretePointer):
     pass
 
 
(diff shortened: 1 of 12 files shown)


elprans (approved):

msullivan:
@raddevon We'll need doc updates for this

PR #6074, opened 2023-09-15
Support functional-style if/then/else syntax

msullivan:
In addition to our current "Python-style"
`then_expr if cond_expr else else_expr`, add support for
Haskell/SML/OCaml-style
`if cond_expr then then_expr else else_expr`.

The Python-style syntax did a great job fitting a conditional
expression into Python under a lot of tough syntactic constraints, but
non-Python people tend to hate it and it will look pretty bad once we
have conditional DML.

There aren't any grammar conflicts, so this is fully backward
compatible, and both syntaxes can coexist peacefully, though we should
change all of our documentation to prefer the new syntax.

### DIFF
--- a/edb/edgeql/ast.py
+++ b/edb/edgeql/ast.py
@@ -385,6 +385,8 @@ class IfElse(Expr):
     condition: Expr
     if_expr: Expr
     else_expr: Expr
+    # Just affects pretty-printing
+    python_style: bool = False
 
 
 class TupleElement(Base):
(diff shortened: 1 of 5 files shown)


elprans (approved):

1st1:
Please don't merge it before the next compiler call.

edb/edgeql/ast.py
```diff
@@ -385,6 +385,8 @@ class IfElse(Expr):
     condition: Expr
     if_expr: Expr
     else_expr: Expr
+    # Just affects pretty-printing
+    python_style: bool = False
```
aljazerzen, line 389: We probably don't need this as we can always codegen the new syntax - semantically they are equivalent and we can pick a preferred opinionated syntax.

aljazerzen (approved):

1st1 (approved):

msullivan:
@raddevon We'll need documentation updates for this

PR #6090, opened 2023-09-18
Adjust error message for some INSERT cases.

vpetrovykh:
Inserting an expression can indicate that the intent was to perform a conditional insert. Check the expression and provide more specific error message.

### DIFF
--- a/edb/edgeql/parser/grammar/expressions.py
+++ b/edb/edgeql/parser/grammar/expressions.py
@@ -302,9 +302,29 @@ class SimpleInsert(Nonterm):
                 len(subj_path.steps) == 1 and  \
                 isinstance(subj_path.steps[0], qlast.ObjectRef):
             objtype = subj_path.steps[0]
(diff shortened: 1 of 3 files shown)


aljazerzen (approved):

edb/edgeql/parser/grammar/expressions.py
```diff
@@ -293,9 +293,31 @@ def reduce_Insert(self, *kids):
...
+            raise errors.EdgeQLSyntaxError(
+                f"INSERT statements cannot be used inside conditional "
+                f"expressions",
+                hint=(
+                    f"To resolve this try to factor out the mutation "
+                    f"expression into the top-level WITH block. "
+                    f"Alternatively, use a FOR loop to iterate over "
+                    f"some data that's constructed using conditional "
+                    f"expressions and then supplied to the INSERT."
+                ),
```
msullivan, line 310: Is the idea that once we support conditional inserts, we'll do something different here?

msullivan (approved):

PR #6104, opened 2023-09-19
Handle local auth errors when redirecting

scotttrinh:
When implementing an HTML form login flow, we need to be able to redirect back to the form with errors rather than redirecting to the "success" handler to avoid application developers needing to setup this redirect themselves.

### DIFF
--- a/edb/server/protocol/auth_ext/http.py
+++ b/edb/server/protocol/auth_ext/http.py
@@ -143,34 +143,53 @@ class Router:
                     local_client = local.Client(
                         db=self.db, provider_id=register_provider_id
                     )
-                    identity = await local_client.register(data)
-
-                    session_token = self._make_session_token(identity.id)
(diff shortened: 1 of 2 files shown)


edb/server/protocol/auth_ext/http.py
```diff
@@ -143,34 +143,53 @@ async def handle_request(
...
+                            )
+                            redirect_url = (
+                                f"{data['redirect_to']}?{redirect_params}"
+                            )
+                            response.custom_headers["Location"] = redirect_url
+                        else:
+                            response.status = http.HTTPStatus.CREATED
+                            response.custom_headers[
+                                "content-type"
+                            ] = "application/json"
```
jaclarke, line 169: I think `content_type` can be set directly on the `response`.

jaclarke (approved):

PR #6131, opened 2023-09-20
Add 3.0 dumps

msullivan:


elprans (approved):

edb/testbase/server.py
```diff
@@ -1471,7 +1471,7 @@ class StablePGDumpTestCase(BaseQueryTestCase):
 
     BASE_TEST_CLASS = True
     ISOLATED_METHODS = False
-    STABLE_DUMP = True
+    STABLE_DUMP = False
```
elprans, line 1474: ?

msullivan: `STABLE_DUMP = True` is used exclusively for `gen-test-dumps` to figure out which tests to create dumps for, and we don't actually need or want dumps for the pg dump tests. I've updated it to just drop the line, though, instead of setting it to false explicitly.

msullivan: (This doesn't really need to happen in *this* PR but it seemed like a fine time to do it)

PR #6132, opened 2023-09-21
Fix fts::index DDL

aljazerzen:
Bug discovered by @vpetrovykh


edb/pgsql/delta.py
```diff
@@ -3739,19 +3739,39 @@ def apply(
             # the indexes are dropped automatically in this case.
```
msullivan, line 3739: Huh, I'm surprised we have this optimization for indexes. We mostly don't for anything else. (We delete columns one at a time before deleting the table. Maybe we should do better though)

msullivan (approved):

PR #6137, opened 2023-09-21
Doc param defaults

raddevon:
I'm not in love with this, but we need to get something documented here. Open to suggestions, particularly for this sentence which I'm not crazy about:

> Since parameters can only be scalars, arrays of scalars, and, as of EdgeDB 3.0, tuples of scalars, only parts of the query that would be one of those types can be parameterized.

Maybe there's a better way to express this. Maybe there's a name either for all the parts of the query that can be parameterized or that can't, but I'm just not aware of it (or it's not in my head right now).

Debated whether to put in the gnarly example for "paramaterizing" `order by`, but I think we should because it's something people want to do. Maybe for some people, this will work well enough.

### DIFF
--- a/docs/edgeql/parameters.rst
+++ b/docs/edgeql/parameters.rst
@@ -150,6 +150,22 @@ When using a client library, pass the idiomatic null pointer for your language:
     select <required str>$name;
 
 
+Default parameter values
+------------------------
+
+When using optional parameters, you may want to provide a default value to use
+in case the parameter is not passed.
+
+.. code-block:: edgeql-repl
+
+  db> select 'Hello ' ++ <optional str>$name ?? 'there';
+  Parameter <str>$name (Ctrl+D for empty set `{}`): EdgeDB
+  {'Hello EdgeDB'}
+  db> select 'Hello ' ++ <optional str>$name ?? 'there';
+  Parameter <str>$name (Ctrl+D for empty set `{}`):
+  {'Hello there'}
+
+
 What can be parameterized?
 --------------------------
 
--- a/docs/edgeql/parameters.rst
+++ b/docs/edgeql/parameters.rst
@@ -96,10 +96,8 @@ Parameter types and JSON
 
 Prior to EdgeDB 3.0, parameters can be only :ref:`scalars
 <ref_datamodel_scalar_types>` or arrays of scalars. In EdgeDB 3.0, parameters
-can also be tuples. This may seem limiting at first, but in actuality this
-doesn't impose any practical limitation on what can be parameterized. To pass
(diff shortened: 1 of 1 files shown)


scotttrinh (approved):
I actually think this is clear enough, especially with the examples given in the "This excluses parts of the query" part.

docs/edgeql/parameters.rst
```diff
@@ -150,11 +148,54 @@ When using a client library, pass the idiomatic null pointer for your language:
     select <required str>$name;
 
 
+Default parameter values
+------------------------
+
+When using optional parameters, you may want to provide a default value to use
```
Dhghomon, line 154: Maybe add a link to here at this point? https://www.edgedb.com/docs/edgeql/sets#coalescing Seeing a sudden ?? can be jarring at time for those new to the operator.

(But not the word 'coalescing' in this page as having a new keyword in the middle of everything can be confusing)

raddevon: I like the idea of linking. I'm not sure how to do it without saying "coalescing" though. Do you have a suggestion? Does it help if I give a brief explanation inline as well as using the term and linking it?

Dhghomon: Maybe just a link at the word default? Personally my mind just sees the word coalescing and turns it to the word default. I guess coalescing is originally a SQL term so makes sense to name it that way though I originally seized on a similar Rust pattern when learning ?? that includes the word default:

https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap_or_default

(Though it's not an exact equivalent, actually more similar to a method called or https://doc.rust-lang.org/std/option/enum.Option.html#method.or) https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=10e2f8fe3c3311b2177992b45d2931e3

raddevon: I'm afraid it might be weird for the word "default" to link to coalescing documentation. I feel like coalesce is a tool that can be used to create a default value, but the two are not synonymous. I'm afraid it could be especially confusing for a low-vision user using a screen reader if they just had it read the links and didn't get the rest of the context.

I do like the idea of a link though. Let me see what I can do with it…

raddevon: CleanShot 2023-09-22 at 08 19 29@2x
I think this is a decent compromise. I refer to is as the `??` operator but also call it by its name in the parenthetical. I think it's important that we do call it by name so that users will know how to talk about it if they need to ask a question or share how they accomplished something with another user. This may be the only context they see it in, so I think it's actually a good idea to use the name here as well.

It was definitely deficient without this link though. I gave no context, so I'm glad you called it out!

PR #5670, opened 2023-06-16
Query params in config commands

aljazerzen:
Closes #4934

PR #6140, opened 2023-09-21
3.4

msullivan:


msullivan:
https://github.com/edgedb/edgedb/pull/5670 was considered feature work, and came in right before the 3.0 release, and so didn't make the cut for the release. I forgot that feature was new, and used it in a test I backported, and it turned out that the existing 3.x behavior was an ISE. I started going to fix the ISE with an error message, then decided that made way less sense than just backporting it.

elprans:
@msullivan, #5670 doesn't seem to have a test for `configure` as you requested there.  Can we add one now?

msullivan:
> @msullivan, #5670 doesn't seem to have a test for `configure` as you requested there. Can we add one now?

It looks like one got added in #5683

aljazerzen (approved):

PR #6160, opened 2023-09-25
Add password reset flow

jaclarke:


edb/server/protocol/auth_ext/http.py
```diff
@@ -305,28 +432,51 @@ async def handle_request(
...
-                            "ext::auth::AuthConfig::providers",
-                            frozenset
-                        )
-                        password_providers = [
-                            p for p in providers
-                            if (util.get_config_typename(p) ==
-                                'ext::auth::PasswordClientConfig')
-                        ]
-                        assert(len(password_providers) == 1)
+                        password_provider = self._get_password_provider()
```
scotttrinh, line 435: Can we return a `404` for these routes if the `password_provider` is `None` instead of throwing an error?

edb/server/protocol/auth_ext/local.py
```diff
@@ -171,3 +186,100 @@ async def authenticate(self, db: Any, input: dict[str, Any]):
...
+        password = input["password"]
+
+        local_identity = await self.validate_reset_secret(
+            db, identity_id, secret)
+
+        if local_identity is None:
+            raise errors.InvalidData("Invalid 'reset_token'")
+
+        # TODO: check if race between validating secret and updating password
+        #       is a problem
```
scotttrinh, line 269: Yeah, it would be nice to take a row-lock on the `PasswordCredential` when validating. We don't have a mechanism for that in EdgeQL yet though, right?

edb/server/protocol/auth_ext/ui.py
```diff
@@ -16,37 +16,41 @@
...
 
 from .util import get_config_typename
 
 
 oauth_provider_names = {
-  'github': 'Github',
-  'google': 'Google',
-  'apple': 'Apple',
-  'azure': 'Azure'
+    'github': 'Github',
```
scotttrinh, line 27: tiny nit:

```suggestion
    'github': 'GitHub',
```

tests/test_http_ext_auth.py
```diff
@@ -175,6 +176,10 @@
 }
 
 
+def utcnow():
+    return datetime.datetime.now(datetime.timezone.utc)
```
scotttrinh, line 180: ❤️

tests/test_http_ext_auth.py
```diff
@@ -1952,3 +1957,384 @@ async def test_http_auth_ext_local_password_authenticate_01(self):
...
+                reset_url.startswith(form_data['reset_url'] + '?reset_token=')
+            )
+
+            claims = await self.extract_jwt_claims(
+                reset_url.split('=', maxsplit=1)[1]
+            )
+            self.assertEqual(claims.get("sub"), str(identity[0].id))
+            self.assertEqual(claims.get("iss"), str(self.http_addr))
+            now = utcnow()
+            tomorrow = now + datetime.timedelta(minutes=10)
```
scotttrinh, line 2032: Nit:

```suggestion
            tenMinutesLater = now + datetime.timedelta(minutes=10)
```

—?

scotttrinh (changes requested):
Looks great! A few minor things. Only the `404` request is somewhat blocking, but open to feedback on that.

scotttrinh (approved):
No blockers, just a few points worth discussing.

edb/server/protocol/auth_ext/http.py
```diff
@@ -337,6 +317,350 @@ async def handle_request(
...
+                                query, 'error'
+                            ),
+                            email=_maybe_get_search_param(query, 'email'),
+                            app_name=ui_config.app_name,
+                            logo_url=ui_config.logo_url,
+                            dark_logo_url=ui_config.dark_logo_url,
+                            brand_color=ui_config.brand_color,
+                        )
+
+                case ('signup',):
```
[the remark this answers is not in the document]
jaclarke: Yeah, that's probably a good idea. I did have that thought when I ending up naming a page the same as an api endpoint, and meant to come back to it later, then forgot.

PR #6163, opened 2023-09-26
Improve FTS weights

aljazerzen:
Limits what's accepted into `weight_category`.

Ideally, `with_options(weight_category)` would be an `anyenum`,
similar to language, so people can use their own enum,
with their own naming.
It would ignore the names of the enum variants and instead only
use their ordering which would have to match the ordering of
the weight values in the `fts::search`.

Right now, we cannot have it be an `anyenum`,
because the `anyenum` type would get conflated with `language: anyenum`.

Naive option would be to stick with `str` for now, but the future change
to enums would be backward in-compatible.

This PR adds `fts::Weight` enum that must be used in
`with_options(weight_category: fts::WeightCategory)`.
In the future, we can implement proper generics and change it
to `weight_category: anyenum`.

The category now also defaults to 'A',
instead of inheriting Postgres's behavior of defaulting to 'D'.

---

This PR also adjust `fts:search`'s `weights` parameter':
- it now defaults to `[1.0, 0.5, 0.25, 0.125]` (as agreed in the RFC),
- if less than four values are provided, it pads the array with zeros,
- if more than four values are provided, it ignores them,
- the four values of the array correspond to categories A, B, C and D, respectively.
  (instead of inheriting Postgres's behavior of D, C, B and A)

### DIFF
--- a/edb/edgeql/compiler/func.py
+++ b/edb/edgeql/compiler/func.py
@@ -1003,7 +1003,7 @@ def compile_fts_with_options(
     weight_expr = call.args[1].expr
     if not irutils.is_const(weight_expr):
         raise errors.InvalidValueError(
-            f"fts::search weight_category must be a literal",
+            f"fts::search weight_category must be a constant",
             context=weight_expr.context,
         )
     weight_const = irutils.as_const(weight_expr)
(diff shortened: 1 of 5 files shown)


vpetrovykh (approved):
Using a built-in enum for now is OK. Making it more generic in the future (once we have a good way of distinguishing one `anyenum` from another) would be backwards compatible.

PR #6165, opened 2023-09-26
Fix remote compiler unpickling None bug

fantix:
This bug was introduced in [#5921#diff](https://github.com/edgedb/edgedb/pull/5921/commits/4bde3adc1075ce4acee66f53655f9f8c41bac662#diff-cb023165d37a6f71ee27d63c154a2ad31c6c0509668ec482846db124ab1b5bec)

Before that change, we used to pickle and unpickle `None`. But now we no longer pickle/unpickle the user schema in remote compiler server, so the `None` literal is passed to the worker.

This also fixed 2 multi-tenant server issues when 0 tenant configured.

### DIFF
--- a/edb/server/protocol/system_api.py
+++ b/edb/server/protocol/system_api.py
@@ -49,7 +49,7 @@ async def handle_request(
         if path_parts == ['status', 'ready'] and request.method == b'GET':
             if tenant is None:
                 # TODO(fantix): test the compiler pool
-                _response_ok(response, "OK")
+                _response_ok(response, b"OK")
             else:
                 await tenant.create_task(
                     handle_readiness_query(request, response, tenant),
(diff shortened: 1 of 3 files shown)


edb/server/protocol/system_api.py
```diff
@@ -49,7 +49,7 @@ async def handle_request(
         if path_parts == ['status', 'ready'] and request.method == b'GET':
             if tenant is None:
                 # TODO(fantix): test the compiler pool
```
elprans, line 51: Still TODO?

elprans (approved):

edb/server/protocol/system_api.py
fantix: Yeah, I haven't got to that part. I'd also like to reuse the syscon for pings if possible. I'll start a new PR for that.

Issue #6086, opened 2023-09-18
Document how to get back type in polymorphic queries


raddevon:
Document here: http://www.edgedb.com/docs/edgeql/select#polymorphic-queries

`select MyAbstractType {__type__: {name}};` or if you need it at the top level, you can do this: `select MyAbstractType {type := .__type__.name};`

PR #6171, opened 2023-09-27
Select page cleanup

Dhghomon:
I started work on https://github.com/edgedb/edgedb/issues/6086 but noticed that the query output is a bit hard for the reader to predict: sometimes the output assumes certain heroes, movies and villains and other times another. But by keeping the same schema and adding a few inserts the resulting input is now predictable for the reader and can be reproduced and experimented with locally as well.

### DIFF
--- a/docs/edgeql/select.rst
+++ b/docs/edgeql/select.rst
@@ -49,7 +49,7 @@ Selecting objects
 -----------------
 
 However most queries are selecting *objects* that live in the database. For
-demonstration purposes, the queries below assume the following schema.
+demonstration purposes, the queries below assume the following schema:
 
 .. code-block:: sdl
     :version-lt: 3.0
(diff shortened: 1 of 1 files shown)


raddevon:
We should update the results on 174-178 to reflect the new contents of the database by making the number of results and the IDs of those results correct.

raddevon (approved):
This is a great change, @Dhghomon! I'm approving so you'll be able to merge after some changes, but I do think there are a couple of things that could make this even more consistent. I called out one of them in my previous comment, but I think there may be a few places where IDs appear that are not consistent with your new data. I know it's a small thing, but, if we're going to be consistent, may as well apply it to the object IDs too.

I make a pass through this entire page and make sure that every instance of an ID reflects what's in your new data. Aside from the one I called out, I noticed a couple of instances of old IDs in the filtering section. There may be others I missed.

PR #6175, opened 2023-09-27
Combine all EdgeQL grammars into a single one

aljazerzen:
As discussed, this PR combines all grammars into a single one, introducing new special token kinds that are injected before all other tokens.

I've also removed bits of code that were used to handle multiple grammars, as I don't think they will be used again (this can easily be reverted).

Comparisons of grammar sizes:

```
Length of encoded JSON string:
old "block" grammar: 7707556
new "start" grammar: 7727965

➜  edgedb git:(parser-combine) ll -S edb/edgeql/parser/grammar/*.pickle
-rw-r--r-- 1 aljaz users 4,6M sep 27 11:38 edb/edgeql/parser/grammar/block.pickle
-rw-r--r-- 1 aljaz users 4,6M sep 27 12:25 edb/edgeql/parser/grammar/start.pickle
-rw-r--r-- 1 aljaz users 4,6M sep 26 09:23 edb/edgeql/parser/grammar/single.pickle
-rw-r--r-- 1 aljaz users 3,8M sep 27 10:37 edb/edgeql/parser/grammar/sdldocument.pickle
-rw-r--r-- 1 aljaz users 2,8M sep 27 10:37 edb/edgeql/parser/grammar/extension_package_body.pickle
-rw-r--r-- 1 aljaz users 2,8M sep 27 10:37 edb/edgeql/parser/grammar/migration_body.pickle
-rw-r--r-- 1 aljaz users 1,6M sep 27 11:38 edb/edgeql/parser/grammar/fragment.pickle
```


msullivan (approved):
Do we want to have `fragment` available as a separate grammar, since IIRC the server process only uses that one?

Not sure if it is necessary

aljazerzen:
> Do we want to have fragment available as a separate grammar, since IIRC the server process only uses that one?

I cannot find where that is accessed from.- if it is true, we'd be increasing memory usage of the server from 1.2MB to 4.2MB. For our cloud that would not be that problematic, as we have many compiler processes anyway and this would be net decrease in memory. But for smaller deployments, this is not ideal.  

Good thing that I put the "remove code dealing with multiple grammars" into separate commits.

elprans:
> since IIRC the server process only uses that one?

It also needs `block`.

msullivan:
> > since IIRC the server process only uses that one?
> 
> It also needs `block`.

Ah, then there's no point

Issue #4437, opened 2022-09-27
Consider allowing DML in conditionals


msullivan:
I think we could do it without too much trouble.
The key insight here is that we actually already *do* support conditional DML: DML can appear inside of `FOR` statements, and so the number of times a DML statement is executed (including possibly zero) depends on the value of the for iterator.

Then,
```
IF_BRANCH if COND else ELSE_BRANCH
```
is (modulo the fact that it adds a fence around COND and potentially increases the inferred cardinality)
equivalent to
```
for b in COND union (
  {
    (for _ in (select () filter b) union (IF_BRANCH)),
    (for _ in (select () filter not b) union (ELSE_BRANCH)),
  }
)
```

elprans:
Also, `COND` must not be correlated with anything, as is the general rule for DML dependencies.

msullivan:
And if we *don't* do this, we should document this sort of pattern somewhere

avedetvedea:
@msullivan I have used your for-loop workaround that you provided in another thread to do some pretty complicated things and never hit a bug. The only problem is I have crazy for loops in my queries when I want to do conditional-DMLs. Looking forward to having this feature which is so coherent with the EQL's do-all-in-a-single-query ethos

avedetvedea:
Could this ```if else``` improvement help with things like returning different shaped data (read empty vs loaded array) depending on a condition, i.e., returning different 'loaded' data based on that condition? https://discord.com/channels/841451783728529451/1101544217597190164 

I have been working around that problem with the same for loop hack with success

AnsonHwang86:
Waiting for this crazy pattern.

PR #6177, opened 2023-09-27
Document for conditional DML workaround

### LINKED ISSUE #6091: Document `for` conditional workaround

raddevon:
This becomes obsolete with https://github.com/edgedb/edgedb/issues/4437, but I'm not sure how long that implementation will take and whether or not it will release before 4.0. In the meantime, this is a workaround that gets passed around a ton on Discord, so we may as well have it documented.

Closes #6091


msullivan (approved):
Though we also might want to mention that you can often use UNLESS CONFLICT too.

PR #6180, opened 2023-09-27
Skip notification job if event is PR

raddevon:


elprans:
While here, can you also pin the action to a [specific commit SHA](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions) please?

fantix (approved):

PR #6189, opened 2023-09-28
Implement `blocked` readiness state

elprans:
The difference from the `offline` state is that the server still accepts
connections, but immediately closes them with a protocol-appropriate
error.


fantix (approved):
👍

PR #6195, opened 2023-09-29
Add a PKCE flow to end of OAuth

scotttrinh:
This is the most secure way to pass an `auth_token` back to a server given that the Cookie we create is actually between the browser _and the EdgeDB server_ not the application's server.

Here's a quick overview of how it works:

1. The application server generates a random string (called the "verifier") and then hashes the verifier and sends this value (called the "challenge") at the beginning of the OAuth flow with the EdgeDB server. Importantly **the application server stores the verifier in an `HttpOnlly` cookie with the browser**. This allows the application to extract the verifier later.
2. The EdgeDB server stores this "challenge" in the database and adds it as a value in the `state` JWT that is passed back and forth through the typical OAuth flow.
3. At the end of the OAuth flow with the third-party Identity Provider, the EdgeDB server returns the ID to the PKCE object as a "code" and passes this back through the browser redirect to the application server.
4. The browser redirects to the application server where it then extracts the verifier from the cookie it set in step 1, and sends the `code` and `verifier` to the EdgeDB server (at `/token`).
5. The EdgeDB server looks up the `PKCE` object via the `code` (which is the `PKCE.id`), and then does the same sha256 hash on the verifier. If that hashed value matches the `PKCE.challenge` we know it's a good request, and we return the auth token and identity ID back to the application server. This completes the PKCE flow.

(additionally, we remove the used `PKCE` object to avoid replay attacks, and expire these `PKCE` objects after 10 minutes)

---

I didn't take a lot of time on naming and modularizing the various parts, so I would ❤️ LOVE ❤️  to get some feedback on naming, code structure, etc.

### DIFF
--- a/edb/lib/ext/auth.edgeql
+++ b/edb/lib/ext/auth.edgeql
@@ -65,6 +65,11 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
         create required property password_hash: std::str;
     };
 
+    create type ext::auth::PKCE extending ext::auth::Auditable {
+        create required property challenge: std::str;
+        create link identity: ext::auth::Identity;
+    };
+
     create type ext::auth::ClientConfig extending cfg::ConfigObject {
         create required property provider_id: std::str {
             set readonly := true;
--- a/edb/lib/ext/auth.edgeql
+++ b/edb/lib/ext/auth.edgeql
@@ -66,7 +66,9 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
     };
 
     create type ext::auth::PKCE extending ext::auth::Auditable {
-        create required property challenge: std::str;
+        create required property challenge: std::str {
+            create constraint exclusive;
+        };
         create link identity: ext::auth::Identity;
     };
 
--- a/edb/lib/ext/auth.edgeql
+++ b/edb/lib/ext/auth.edgeql
@@ -65,7 +65,7 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
         create required property password_hash: std::str;
     };
 
-    create type ext::auth::PKCE extending ext::auth::Auditable {
+    create type ext::auth::PKCEChallenge extending ext::auth::Auditable {
         create required property challenge: std::str {
             create constraint exclusive;
         };
--- a/edb/server/protocol/auth_ext/pkce.py
+++ b/edb/server/protocol/auth_ext/pkce.py
@@ -69,7 +69,9 @@ class PKCE:
               id,
               challenge,
               identity_id := .identity.id
-            } filter .id = <uuid>$id
+            }
+            filter .id = <uuid>$id
+            and (datetime_current() - .created_at) < <duration>'10 minutes';
             """,
             variables={"id": id},
         )
--- a/edb/server/protocol/auth_ext/pkce.py
+++ b/edb/server/protocol/auth_ext/pkce.py
@@ -18,77 +18,85 @@
 
 
 import json
+import dataclasses
 
 from edb.server.protocol import execute
-from . import data
 
 
-class PKCE:
-    def __init__(self, db):
-        self.db = db
+@dataclasses.dataclass(repr=False)
+class PKCEChallenge:
+    """
+    Object that represents the ext::auth::PKCEChallenge type
+    """
 
-    async def create(self, challenge: str):
-        await execute.parse_execute_json(
-            self.db,
-            """
-            insert ext::auth::PKCE {
-              challenge := <str>$challenge,
-            }
-            """,
-            variables={
-                "challenge": challenge,
-            },
-        )
+    id: str
+    challenge: str
+    identity_id: str | None
 
-    async def link_identity_challenge(
-        self, identity_id: str, challenge: str
-    ) -> str:
-        r = await execute.parse_execute_json(
-            self.db,
-            """
-            update ext::auth::PKCE
-            filter .challenge = <str>$challenge
-            set { identity := <ext::auth::Identity><uuid>$identity_id }
-            """,
-            variables={
-                "challenge": challenge,
-                "identity_id": identity_id,
-            },
-        )
 
-        result_json = json.loads(r.decode())
-        assert len(result_json) == 1
+async def create(db, challenge: str):
+    await execute.parse_execute_json(
+        db,
+        """
+        insert ext::auth::PKCEChallenge {
+            challenge := <str>$challenge,
+        }
+        """,
+        variables={
+            "challenge": challenge,
+        },
+    )
 
-        return result_json[0]["id"]
 
-    async def get_by_id(self, id: str) -> data.PKCE:
-        r = await execute.parse_execute_json(
-            self.db,
-            """
-            select ext::auth::PKCE {
-              id,
-              challenge,
-              identity_id := .identity.id
-            }
-            filter .id = <uuid>$id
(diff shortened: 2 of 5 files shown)


1st1:
> The application server generates a random string (called the "verifier") and then hashes the verifier and sends this value (called the "challenge")

Should we include the name of the hash function into the challenge string?

1st1:
> The application server generates a random string (called the "verifier")

Should we require minimum length for the verifier? (can be enforced at the last step)

edb/lib/ext/auth.edgeql
```diff
@@ -65,6 +65,13 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
         create required property password_hash: std::str;
     };
 
+    create type ext::auth::PKCE extending ext::auth::Auditable {
+        create required property challenge: std::str {
```
1st1, line 69: We will never remove these objects?

edb/server/protocol/auth_ext/pkce.py
```diff
@@ -0,0 +1,94 @@
...
+
+from edb.server.protocol import execute
+from . import data
+
+
+class PKCE:
+    def __init__(self, db):
+        self.db = db
+
+    async def create(self, challenge: str):
```
1st1, line 30: IMO all of these methods should be top level functions in the `pkce.py` module. There's no need for the `PKCE` class as it holds no state; functions can just accept `db` as the first argument.

edb/lib/ext/auth.edgeql
```diff
@@ -65,6 +65,13 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
         create required property password_hash: std::str;
     };
 
+    create type ext::auth::PKCE extending ext::auth::Auditable {
```
1st1, line 68: Perhaps we should rename

```suggestion
    create type ext::auth::PKCEChallenge extending ext::auth::Auditable {
```

edb/server/protocol/auth_ext/pkce.py
1st1: And then I'd move the PKCE dataclass definition to this file; it would make much more sense.

1st1:
> The application server generates a random string (called the "verifier") and then hashes the verifier and sends this value (called the "challenge") at the beginning of the OAuth flow with the EdgeDB server. Importantly the application server stores the verifier in an HttpOnlly cookie with the browser. This allows the application to extract the verifier later.

So here's how I understand it, correct me if I'm wrong: there will be a few ways developers can take advantage of EdgeDB Auth:

1. Implement auth client-side in their one-page JS apps. They render their own login forms and use the future version `egdedb-js@auth` package to invoke APIs like:

   * `client.signIn({user, password})` -- send a request to the EdgeDB API endpoint, no need for PKCE here?
   * `client.signInWithOAuth({...})` -- the page would redirect to the relevant OAuth provider; the PKCE verifier/challenge would be generated by the client and persisted in `localStorage` until redirected back, at which point it can issue a request to an EdgeDB API endpoint to get the JWT. All done completely automatically for the developer.
   
   In this scenario there's no need in any server-side code for the user to have. Everything would be handled by edgedb-js client side + some of our server-side API endpoints.

2. Use our pre-built UI. For this the developer would have to implement one page/URL, say `app.com/login` that would generate the verifier, set it as an HTTPOnly cookie, and then redirect the browser to our UI. Once the login is successful, we redirect the logged in user to the pre-configured `app.com` location, where the server code will always be checking for the HTTPOnly cookie presence. If it's there it would be asking EdgeDB to issue a token.

3. Not use our pre-built UI, but have an app with a sizable backend, where it makes sense to call, say, `edgedb-js@auth` APIs from the server side to get the auth JWT. Good for server-rendered apps etc.

Am I missing or not getting something?

edb/lib/ext/auth.edgeql
scotttrinh: We delete these at the end of the flow once you've completed.

There is an additional issue with "abandoned" PKCE flows that would pile up without GC. I had a couple of ideas here:

1. Maybe we already have some kind of cron-like/job-queue that we can schedule a cleanup with?
2. We could remove all outdated challenges at the beginning of each new challenge.

edb/server/protocol/auth_ext/pkce.py
scotttrinh: Yeah, this would be my natural style in TypeScript, so makes sense!

edb/lib/ext/auth.edgeql
scotttrinh: Oh, I like that name!

scotttrinh:
> Implement auth client-side in their one-page JS apps. They render their own login forms and use the future version egdedb-js@auth package to invoke APIs like:
> - client.signIn({user, password}) -- send a request to the EdgeDB API endpoint, no need for PKCE here?
> - client.signInWithOAuth({...}) -- the page would redirect to the relevant OAuth provider; the PKCE verifier/challenge would be generated by the client and persisted in localStorage until redirected back, at which point it can issue a request to an EdgeDB API endpoint to get the JWT. All done completely automatically for the developer.

Yeah, PKCE is absolutely something we can use for client-side only, but the missing piece here is that if we allow "unauthenticated" users to make any kind of query to the database, why even have authentication? 😅 At the end of this authentication flow, we've created an `ext::auth::Identity`, but the application-level `User` type has to be made by the application. If we encourage people to do this from a SPA, I think we need a better backend-less story around RBAC with database roles to lock down what people can do based on their database user.

But that's a bit orthogonal to this: yeah PKCE can absolutely be implemented client-side only (and mobile app only).

> If the above is right, then specifically for use cases (1) and (3) we should also allow specifying the OAuth "scope" and obtaining the auth token to fetch further info.

Yep, you got all of that right! I agree that we should allow passing the IdP's auth (and refresh) tokens back to the application for all use cases, including when using the built-in UI. The payload of the PKCE request is secure, so we can pass anything back. Now, backend-less apps have to consider the security implications of not having a "secure" place to store sensitive information, but that's already a going to be a problem for them (until we figure out our own backend-less solution and then we can be that secure location).

edb/lib/ext/auth.edgeql
1st1: We can remove all outdated ones on server startup for now.

1st1:
@scotttrinh couple of comments from the above review that I still want to get your opinion on:

* Should we include the name of the hash function into the challenge string?
* Should we require minimum length for the verifier? (can be enforced at the last step)

scotttrinh:
@1st1 

> Should we include the name of the hash function into the challenge string?

We could also include it as a separate value to avoid needing any finicky parsing on the other end.

> Should we require minimum length for the verifier? (can be enforced at the last step)

The last step feels kind of late: the user will have already consented to the auth at the IdP, and then they'll be confronted with a runtime error that they cannot fix. Seems fine to leave it unchecked, but I'm also fine with rejecting a weak verifier. Hopefully this will be "solved" with introducing a library of helpers and documenting things well.

1st1 (approved):

1st1:
> We could also include it as a separate value to avoid needing any finicky parsing on the other end.

As discussed on the call we don't have to do it now in 4.0. The main use case is potential lack of sha256 on the client, but we can write one in vanilla JS or whatnot if we have to. In any case, we can add a second parameter later.

> Should we require minimum length for the verifier? (can be enforced at the last step)

Let's do this before merging this, aside from that it's cool.

Issue #5954, opened 2023-08-24
Finish up extension-defined configs


msullivan:
https://github.com/edgedb/edgedb/pull/5949 did the bulk of the work for extension-defined configs and nested database configs, but there are some loose ends:
- [ ] Support extension-defined SESSION configs
- [ ] Support redacting config values?
- [ ] Support a second layer of nesting in extension objects
- [ ] See if introspection performance can be improved (cache the get config function in `_edgecon_state`?)
- [ ] Add syntax for "CREATE CONFIG TYPE" to distinguish the types further?
- [ ] Documentation updates?

PR #6198, opened 2023-09-29
Cache config JSON to speed up extension config access

msullivan:
Every config object and multi link view makes a call to
_read_sys_config, which adds up pretty badly when accessing extension
configs.

Add a temporary _config_cache table that is used to cache results, and
modify _read_sys_config_full to be able to compute the output for all
three relevant config levels at once.

`select cfg::Config.extensions[is ext::_conf::Config].config_name`
goes from taking about 50ms on average to taking ~23ms on a cache
miss and ~6.5ms on a cache hit.
(Before I removed instance level extension config objects in #6193,
it took ~300ms.)

Probably the last part of #5954 that will get done before 4.x.

### DIFF
--- a/edb/buildmeta.py
+++ b/edb/buildmeta.py
@@ -44,7 +44,7 @@ from edb.common import verutils
 
 
 # Increment this whenever the database layout or stdlib changes.
-EDGEDB_CATALOG_VERSION = 2023_10_02_00_02
+EDGEDB_CATALOG_VERSION = 2023_10_02_00_03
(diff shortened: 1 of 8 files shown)


elprans (approved):

msullivan:
I had planned to specifically write a test that verified that the `__dbver__` cache entry was working properly (that would mimic how I tested it manually), but when I disable the `__dbver__` cache entry, a ton of tests fail, and I am now convinced it doesn't need a new test to exercise it

PR #6199, opened 2023-09-29
Fix parser for old versions of Rust

aljazerzen:
Edit: I've replaced usage of `std::sync::OnceLock` with `once_cell::OnceCell`, since `std::sync::OnceLock` is unstable on whatever version we are using in `macos-x86_64` build.

I've also updated our declared MRSV to 1.65.0, since that's what our code supports. It is gated by `let...else` syntax:

```
│ 502 | /     let CSTNode::Terminal(terminal) = value else {                                           
│ 503 | |         return;                                                                              
│ 504 | |     };  
```

I don't think we should be concerned with raising MSRV, since these packages are probably compiled by us only. But maybe we should change our tests to build with MSRV instead of latest or stable, so bumps to the MSRV don't go unnoticed as they did here.

### DIFF
--- a/edb/edgeql-parser/Cargo.toml
+++ b/edb/edgeql-parser/Cargo.toml
@@ -3,7 +3,7 @@ name = "edgeql-parser"
 version = "0.1.0"
 license = "MIT/Apache-2.0"
 authors = ["MagicStack Inc. <hello@magic.io>"]
-rust-version = "1.59"
+rust-version = "1.65.0"
 edition = "2021"
 
 [dependencies]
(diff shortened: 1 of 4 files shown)


scotttrinh (approved):
Changes look good! Might be worth noting (maybe in the PR description) why we bumped the rust version here.

msullivan:
Yeah, we should pin a rust version in CI.

I think the dev environment setup doc also might reference a rust version

aljazerzen:
It does reference it: 

> Rust compiler and Cargo 1.65 or later;

PR #6201, opened 2023-09-29
Poll a bit more aggressively when waiting for postgres to start up

msullivan:
Shaves like 0.8s off my server startup time, which started bothering
me again since I fixed `--use-db-cache` in #6194...


fantix (approved):
Yeah that makes sense.

aljazerzen (approved):

PR #6209, opened 2023-10-02
Add a '_protected' flag to pointers

msullivan:
This is useful for config purposes.

Turns out that we *didn't* do a safety check for `__type__` on insert,
but would generate bogus SQL and ISE, so this fixes that.

Clean up some error messages in similar id related code, also.

I think we could just call it 'protected', though.

### DIFF
--- a/edb/buildmeta.py
+++ b/edb/buildmeta.py
@@ -44,7 +44,7 @@ from edb.common import verutils
 
 
 # Increment this whenever the database layout or stdlib changes.
-EDGEDB_CATALOG_VERSION = 2023_09_28_00_01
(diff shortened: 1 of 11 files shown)


elprans (approved):
I'm OK with calling it `protected`.

1st1:
I don't see this being gated for use in stdlib mode only. IIRC we decided to gate it.

msullivan:
> I don't see this being gated for use in stdlib mode only. IIRC we decided to gate it.

Schema fields are automatically gated for stdlib mode only unless we opt in with `allow_ddl_set=True`

1st1:
> Schema fields are automatically gated for stdlib mode only unless we opt in with allow_ddl_set=True

Got it, cool

PR #6210, opened 2023-10-02
Update OIDC provider based on Azure flow testing

scotttrinh:
A few changes based on testing the OAuth flow for Azure:

1. Azure requires that requests are form encoded, but hilariously respond with JSON even if you ask for a form encoded response. 🤡 
2. The issuer for the Azure JWT will be the specific tenant that the user belongs to rather than the `common` tenant you requested with. Seems _fine_ to drop the requirement for matching issuer since there is no reasonable way to generically support arbitrary checks.

For point 2, maybe we can make this configurable in the sub-class for the `AzureProvider`? I'm open to suggestions here!

### DIFF
--- a/edb/server/protocol/auth_ext/base.py
+++ b/edb/server/protocol/auth_ext/base.py
@@ -127,7 +127,10 @@ class OpenIDProvider(BaseProvider):
             if content_type.startswith(str(ContentType.JSON)):
                 response_body = resp.json()
             else:
-                response_body = urllib.parse.parse_qs(resp.text)
+                response_body = {
+                    k: v[0] if len(v) == 1 else v
+                    for k, v in urllib.parse.parse_qs(resp.text).items()
+                }
 
             return data.OpenIDConnectAccessTokenResponse(**response_body)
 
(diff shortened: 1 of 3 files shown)


edb/server/protocol/auth_ext/base.py
```diff
@@ -86,23 +100,36 @@ async def exchange_code(
...
+                "code": code,
+                "client_id": self.client_id,
+                "client_secret": self.client_secret,
+                "redirect_uri": redirect_uri,
+            }
+            if self.content_type == ContentType.JSON:
+                resp = await client.post(
+                    token_endpoint.path,
+                    json=request_body,
+                    headers={"Accept": "application/json"},
```
fantix, line 114: This can be a common headers dict shared by the 2 if/else branches I think?

```suggestion
                    headers={"Accept": self.content_type.value},
```

edb/server/protocol/auth_ext/base.py
```diff
@@ -133,10 +160,10 @@ async def fetch_user_info(
             raise errors.MisconfiguredProvider(
                 "Failed to parse ID token with provider keyset"
             ) from e
-        if payload.get("iss") != self.issuer_url:
-            raise errors.InvalidData("Invalid value for iss in id_token")
```
fantix, line 137: Just to be sure, this deletion was intentional, right?

fantix (approved):

1st1 (approved):

PR #6211, opened 2023-10-02
Add email-sending support as SMTP client

fantix:
- [x] SMTP server config
- [x] Email sending API
- [x] Manual testing
- [x] Add unit tests
- [x] Integrate with auth for sending password reset email


edb/server/ext/smtp.py
```diff
@@ -0,0 +1,174 @@
...
+import asyncio
+import email
+
+import aiosmtplib
+
+from edb.common import retryloop
+from edb.ir import statypes
+
+from . import util
+from ... import errors
```
1st1, line 30: Please use absolute imports for when we're importing from parents and grandparents

edb/lib/ext/smtp.edgeql
```diff
@@ -0,0 +1,49 @@
...
+        create required property security -> ext::smtp::ConnectionSecurity {
+            set default := ext::smtp::ConnectionSecurity.PreferSTARTTLS;
+        };
+        create required property validate_certs -> std::bool {
+            set default := true;
+        };
+        create required property concurrency -> std::int32 {
+            create constraint std::min_value(1);
+            set default := 2;
+        };
```
1st1, line 41: I saw how you use it in code, but I'm not sure I fully understand it. How does configuring `concurrency` affects things? Can we hide this config option and auto-configure it on the server somehow?

fantix: This limits the number of maximum concurrent SMTP connections - for example, setting this value to `1` will serialize all SMTP requests. So this is basically a customer feature to be able to play nice with their SMTP providers (ideally, there should also be a rate limit).

fantix: The caller will block until the email is sent.

edb/server/protocol/auth_ext/ui.py
```diff
@@ -409,3 +411,43 @@ def _render_button(text: str):
...
+    )
+    alternative.attach(plain_text_msg)
+    html_msg = mime_text.MIMEText(
+        f"""
+<!DOCTYPE html>
+<html>
+  <body>
+    <a href="{reset_url}">Reset password</a>
+  </body>
+</html>
```
fantix, line 446: We need proper title / plain text / HTML email content here.

tests/test_http_ext_auth.py
```diff
@@ -315,6 +317,21 @@ class TestHttpExtAuth(tb.ExtAuthTestCase):
...
+        ext::auth::SMTPConfig::timeout_per_attempt := <duration>'1 seconds';
+
+        CONFIGURE CURRENT DATABASE SET
+        ext::auth::SMTPConfig::timeout_per_email := <duration>'1 seconds';
+
+        CONFIGURE CURRENT DATABASE SET
+        ext::auth::SMTPConfig::validate_certs := true;
+
+        CONFIGURE CURRENT DATABASE SET
+        ext::auth::SMTPConfig::security := ext::auth::SMTPSecurity.TLS;
```
fantix, line 333: I was expecting the default values to kick in without setting these manually, but it turned out not the case.

elprans: Hm, this is odd. @msullivan, any idea why?

edb/server/protocol/auth_ext/smtp.py
```diff
@@ -0,0 +1,152 @@
...
+    password = util.maybe_get_config(
+        db.db_config,  # type: ignore
+        "ext::auth::SMTPConfig::password",
+    )
+    timeout_per_attempt = util.get_config(
+        db.db_config,  # type: ignore
+        "ext::auth::SMTPConfig::timeout_per_attempt",
+        expected_type=statypes.Duration,
+    )
+    req_timeout = timeout_per_attempt.to_microseconds() / 1_000_000.0
```
scotttrinh, line 77: `req_timeout` is measured in millionths of a microsecond?

scotttrinh (commented):
Looks really good so far! One clarifying question.

edb/server/protocol/auth_ext/smtp.py
fantix: Yes I believe so: https://www.postgresql.org/docs/current/datatype-datetime.html#DATATYPE-INTERVAL-INPUT

fantix: Ah, should use `*` - good catch!

scotttrinh (approved):

edb/lib/ext/auth.edgeql
```diff
@@ -144,6 +144,59 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
                 indicates that the token should never expire.";
             set default := <std::duration>'336 hours';
         };
+
+    };
+
+    create scalar type ext::auth::SMTPSecurity extending enum<PlainText, TLS, STARTTLS, STARTTLSOrPlainText>;
```
1st1, line 150: Should it be `StartTLS`?

fantix: Yeah I'm seeing both in different places. [RFC 3207](https://www.ietf.org/rfc/rfc3207.txt) uses "STARTTLS".

edb/server/protocol/auth_ext/smtp.py
```diff
@@ -0,0 +1,152 @@
...
+        email.message.Message,
+        str,
+        bytes,
+    ],
+    sender: Optional[str] = None,
+    recipients: Optional[Union[str, Sequence[str]]] = None,
+    test_mode: bool = False,
+) -> None:
+    global _semaphore
+    if _semaphore is None:
```
1st1, line 50: So this is a global semaphore for all tenants?

fantix: Yes that's correct.

tests/test_http_ext_auth.py
fantix: This is probably because we didn't use the `spec.Setting.default` value in `auth/utils.py`, I'll push a fix.

PR #6212, opened 2023-10-02
Fix exclusive constraints in configuration objects

msullivan:
Right now we use the full *tuple* of exclusive constraints as the
exclusive key. It needs to be each exclusive pointer individually.

### DIFF
--- a/edb/buildmeta.py
+++ b/edb/buildmeta.py
@@ -44,7 +44,7 @@ from edb.common import verutils
 
 
 # Increment this whenever the database layout or stdlib changes.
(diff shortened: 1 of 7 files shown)


msullivan:
My one concern here is with `cfg::Auth`: https://github.com/edgedb/edgedb/blob/97c818511bd9efe2100000224278306764a25295/edb/lib/cfg.edgeql#L66-L79

I don't actually really understand our auth system: did we *intend* for this to be an exclusive constraint on the tuple (as was implemented before), or separate exclusive constraints on the properties?

elprans (approved):

PR #6214, opened 2023-10-03
+'please specify another' for nonexistent database

Dhghomon:
A bit of Python code needs to be changed to add the final 'please specify another one' output mentioned here so gave it a go: https://github.com/edgedb/edgedb-cli/issues/1125

I've changed test_database_drop_01 to match as well but the workflow tests are still showing the old output. Revert the change to test_database_drop_01 or do something else? @msullivan

### DIFF
--- a/edb/server/dbview/dbview.pyx
+++ b/edb/server/dbview/dbview.pyx
@@ -1185,7 +1185,7 @@ cdef class DatabaseIndex:
             return self._dbs[dbname]
         except KeyError:
             raise errors.UnknownDatabaseError(
-                f'database {dbname!r} does not exist')
+                f'database {dbname!r} does not exist; please specify another')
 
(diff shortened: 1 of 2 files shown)


msullivan:
The issue with the test here is that the error for not being able to drop a database doesn't seem to come from that check of ours in dbview, but is instead triggered by postgres failing to drop the database when we send it a `DROP DATABASE` command. Probably that gets turned into an edgedb error by `edb/server/compiler/errormech.py`. I don't think it would be an improvement to add "please specify another" there.

I don't really think that it's an improvement to the connect path either to have this extra bit of error message, though. It seems reasonable as a message from the CLI or an interactive tool but seems off for the server to say it

Dhghomon:
Sounds good, I'll look for a good place to have the CLI display the hint.

PR #6215, opened 2023-10-03
Support some simple constraints on config object properties

msullivan:
We have to be able to evaluate them at compile time, which is a bit of a
downside, but probably doesn't matter in practice.

### DIFF
--- a/edb/edgeql/compiler/config.py
+++ b/edb/edgeql/compiler/config.py
@@ -34,6 +34,7 @@ from edb.ir import staeval as ireval
 from edb.ir import statypes as statypes
 from edb.ir import typeutils as irtyputils
 
+from edb.schema import constraints as s_constr
 from edb.schema import globals as s_globals
 from edb.schema import links as s_links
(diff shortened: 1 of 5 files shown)


edb/edgeql/compiler/config.py
```diff
@@ -291,7 +304,41 @@ def _validate_global_op(
...
+        with ctx.detached() as sctx:
+            sctx.partial_path_prefix = expr
+            sctx.anchors = ctx.anchors.copy()
+            sctx.anchors[qlast.Subject().name] = expr
+
+            final_expr = constraint.get_finalexpr(ctx.env.schema)
+            assert final_expr is not None and final_expr.qlast is not None
+            ir = dispatch.compile(final_expr.qlast, ctx=sctx)
+
+        result = ireval.evaluate(ir, schema=ctx.env.schema)
```
elprans, line 332: Ideally we need to check if the constraint you added can be staevaled at the point of config type definition.

elprans (approved):

PR #6216, opened 2023-10-03
Constraint documentation fixes

Dhghomon:
Some fixes and rewriting for the page on constraints and related table:

* min_len_value: only applies to str so change extending type
* Type names: change to PascalCase (UpperCamelCase) which seems to be the convention and stands out better
* Section on automatic indexing for `exclusive` states that links get indexed with it, but indexes are already indexed by default. However, index properties do get indexed with `exclusive` so simply added the word property
* Boolean expression: may read as "expression taking a boolean" so changed to expression returning one
* I assume it's correct that *any* possible expression works for `expression on` as long as it adheres to the limitations specified directly after, so reworded as that.
* Table: change strings to str, rewrite a bit, group str-only functions together on the bottom for readability


docs/stdlib/constraints.rst
```diff
@@ -238,9 +237,9 @@ Constraints
     .. note::
 
         This constraint also has an additional effect of creating an
```
raddevon, line 239: "also" + "additional" seem redundant
```suggestion
        This constraint has an additional effect of creating an
```

docs/stdlib/constraints.rst
```diff
@@ -224,9 +224,8 @@ Constraints
             constraint exclusive on ( (.x, .y) );
         }
 
-    In principle, many possible expressions can appear in the ``on
-    (<expr>)`` clause of the ``exclusive`` constraint with a few
-    caveats:
+    Any possible expression can appear in the ``on (<expr>)`` clause of
+    the ``exclusive`` constraint as long as it adheres to the following:
 
     * The expression can only contain references to the immediate
```
raddevon, line 230: While we're in here, it would be really cool if we could show examples of these. For each item, I would love to see one example that works and one that doesn't. If that's too much, maybe we just explain these better. Areas of confusion for me:

- What does it mean that the expression can contain a reference to "links of the type?" I thought constraints could not dig into links. (I think it may mean you can constrain on the link itself but not on properties of the linked object.)
- What is a "long path?"
- How do I know which functions are `Immutable`?

docs/stdlib/constraints.rst
```diff
@@ -238,9 +237,9 @@ Constraints
     .. note::
 
         This constraint also has an additional effect of creating an
-        implicit :ref:`index <ref_datamodel_indexes>` on the link or
-        property. This means that in the above example there's no need to
-        add explicit indexes for the ``name`` property.
+        implicit :ref:`index <ref_datamodel_indexes>` on a link or
+        link property. This means that in the ``User`` example there's no need
```
raddevon, line 241: I think this was correct as originally written. Not sure if it's possible to constrain on link properties to be honest with you. If it is and if this also holds true for them, I believe that would be in *addition* to properties, rather than *instead* of them.
```suggestion
        property. This means that in the ``User`` example there's no need
```

Maybe @vpetrovykh can confirm?

raddevon (changes requested):
This looks pretty good. There's one change that @vpetrovykh might need to look at. One other minor change.

docs/stdlib/constraints.rst
raddevon: Actually, this probably isn't in scope for this change. I'll file a new issue for it.

raddevon: https://github.com/edgedb/edgedb/issues/6217

docs/stdlib/constraints.rst
Dhghomon: Ah, I just noticed that one is a typo on my part: should say "on a property or link property", not "link or link property".

raddevon: Are you sure that's accurate? This should be about properties and link properties rather than properties and links?

elprans: A more correct statement would be "an implicit index is created using the expression specified in the `exclusive` constraint", so if we are constraining on a bare property or a link property, that would have an effect of creating an index on it, so you don't have to do this:

```
type Foo {
   name: str { constraint exclusive };
   index on (.name)
}
```

All links are already indexed, so this point is moot for those.

Dhghomon: Yeah, maybe could use that example as a show-not-tell with inline comments:

```
type Foo {
   name: str { constraint exclusive }; # Makes implicit index
   index on (.name)                          # So you don't need to do this
}
```

Dhghomon: > Are you sure that's accurate? This should be about properties and link properties rather than properties and links?

Reading this again I think I'll just delete the word link and leave `property` since a link property is also a property and don't want to distract the reader from the subject which is just a quick note on the `exclusive` constraint.

Also I think maybe this will work better: also un-notify the note section and move it up to directly under `type User`. That lets the following section breathe on its own and finally end on the part that you brought up the new issue for yesterday.

https://github.com/edgedb/edgedb/commit/7c017bf37f684cff121124f3cbd6b545447cbaeb

Dhghomon: Actually...this could work. Move the note above the example and add a commented out line to explicitly show what you don't need to do.

https://github.com/edgedb/edgedb/commit/65763f74aad1b499971a1815e57f360bf7c92c98

raddevon (approved):
Looks great! 👏

PR #6219, opened 2023-10-03
Return OAuth provider tokens

scotttrinh:
This saves the provider token(s) that we used during the OAuth flow into the PKCE flow so that when you complete the OAuth flow, you'll receive the provider token (and refresh token) in the response body.

Notes:
- We are storing the identity provider's token, which we should do carefully and as temporarily as possible. We're going to work on a periodic cleanup job separately that will GC any expired `PKCEChallenge` objects.
- The provider may or may not return a refresh token.
- We don't currently pass any additional scopes to the provider, but that's a pretty common use-case. We'll support that in a separate PR.

### DIFF
--- a/edb/server/protocol/auth_ext/data.py
+++ b/edb/server/protocol/auth_ext/data.py
@@ -133,7 +133,4 @@ class OpenIDConnectAccessTokenResponse(OAuthAccessTokenResponse):
     id_token: str
 
     def __init__(self, **kwargs):
-        for field in dataclasses.fields(self):
-            if field.name in kwargs:
-                setattr(self, field.name, kwargs.pop(field.name))
-        self._extra_fields = kwargs
+        super().__init__(**kwargs)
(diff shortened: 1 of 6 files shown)


edb/server/protocol/auth_ext/data.py
```diff
@@ -112,12 +112,14 @@ class OAuthAccessTokenResponse:
...
     token_type: str
     expires_in: int
-    refresh_token: str
+    refresh_token: str | None
 
     def __init__(self, **kwargs):
         for field in dataclasses.fields(self):
             if field.name in kwargs:
                 setattr(self, field.name, kwargs.pop(field.name))
+            else:
```
1st1, line 121: FWIW I don't think you're supposed to overload the `__init__` method on dataclasses

1st1: Never mind, it seems to be supported by the API

> init: If true (the default), a [__init__()](https://docs.python.org/3/reference/datamodel.html#object.__init__) method will be generated.

> If the class already defines [__init__()](https://docs.python.org/3/reference/datamodel.html#object.__init__), this parameter is ignored.

1st1 (approved):
Let's add support for `additional_scopes`

Issue #6222, opened 2023-10-03
SELECT can incorrectly return data from an `UPDATE` overlay when ensure_source_rvar is used


msullivan:
This is a placeholder issue. Once I land a pending PR, I am going to close #3030 in favor of this issue, which I will update. Filing it now to reserve a bug number to put into a test xfail.

PR #6223, opened 2023-10-03
Be able to depend on multiple dml sources when producing overlays

### LINKED ISSUE #3030: Trouble with DML/overlays and ensure_source_rvar
In #3023 I added an xfailed insert test. A simplified version of it is
```
WITH                                                                      
     F := (INSERT Subordinate {name := "!"}),                             
     Z := enumerate((F,)),                                                
SELECT Z.1.0; 
```
and it returns an empty set. The issue turns out to be:
 1. no source rvar for Z.1.0 gets advertised by enumerate, so the compiler generates a new_primitive_rvar and joins it in
 2. we don't detect the insert as a dml_source of `Z.1.0`, and so the overlay isn't applied. This is because `get_nearest_dml` just looks down the main path.

I think this particular case might be easy to fix by making sure that enumerate can preserve source rvars for its elements, but there are plenty of operations that won't be able to preserve source rvars, like:
```
WITH                                                                      
     F := (INSERT Subordinate {name := "!"}),                             
SELECT [F][0];
```

Part 2 could be addressed by taking a wider view of finding dml sources; perhaps by grabbing any dml source in the ir, at all. But this made me realize that there is a more general problem here.
Consider
```
WITH                                                                      
     A := (SELECT UpdateTest FILTER .name = 'update-test2'),              
     B := (UPDATE A SET { comment := "foo" })                             
SELECT ((SELECT {A, B} {name, comment}));
```
We will output
```
[
  {"comment": "second", "name": "update-test2"},
  {"comment": "foo", "name": "update-test2"}
]
```
One object is from the select, and reflects the old data, and one from the update, and reflects the new.

If we try something where we lose track of the source rvars, though:
```
WITH                                                                      
     A := (SELECT UpdateTest FILTER .name = 'update-test2'),              
     B := (UPDATE UpdateTest FILTER .name = 'update-test2' SET { comment := "foo" })                                                               
SELECT assert_exists((SELECT {A, B} {name, comment}));
```
... then we trip an assertion. But if we remove that assertion, then we get
```
[
  {"comment": "second", "name": "update-test2"},
  {"comment": "second", "name": "update-test2"}
]
```
with both objects reflecting the old data, since we don't use the overlay.
But if we *did* arrange to use the overlay, then both objects would reflect the new data, which is also wrong!

Getting this right in cases like this seems pretty tricky. Possible approaches:
1. Some sort of dynamic provenance tracking
2. Reusing materialization to track any data that might be output

I'm not sure if option 2 scales to dealing with links that are also changed, and the like, though.

This issue seems very tricky to get right.

### LINKED ISSUE #6057: tuples + with + insert can lose objects
```
with noobs := {                                                           
  ((insert Person { name := "foo" }), "bar"),                             
  ((insert Person { name := "spam" }), "eggs"),                           
},                                                                        
select noobs; 
```
gives back
```
{}
```

Probably some sort of overlay thing. Ugh!

The original issue is some variant of:
```
with noobs := {
  ((insert Person { name := "foo" }), "bar"),
  ((insert Person { name := "spam" }), "eggs"),
},
insert Movie {
  title := "test",
  actors := distinct (noobs.0),
};
```
And then
```
movies> select Movie { title, actors: {name, @role} };
[{"title": "test", "actors": []}]
```

msullivan:
Queries like:
```
with noobs := {
  ((insert Person { name := "foo" }), "bar"),
  ((insert Person { name := "spam" }), "eggs"),
},
select noobs;
```
would return an empty set, since (for annoying reasons that could
possibly be avoided in this simple case but not in the general case),
we need to do an `ensure_source_rvar` on `noobs`, but
`get_nearest_dml_stmt` is highly limited and can only find one DML
statement. This means we can't produce the proper overlay that
includes both Persons.

Upgrade `get_nearest_dml_stmt` to be able to collect all of the
relevant DML sources.

Fixes #6057. That bug was just a special case of #3030; this fix thus
represents a combination of forward progress on #3030 as well as some
lateral movement: as discussed in that issue, some queries are now
*differently* wrong (see test_edgeql_update_union_overlay_02, which
previously would have returned the old value twice instead of the
new value twice). I am going to close #3030 in favor of an
updated #6222 once this is merged.

### DIFF
--- a/edb/common/typeutils.py
+++ b/edb/common/typeutils.py
@@ -61,3 +61,11 @@ def is_container(obj):
 
 def is_container_type(type_):
     return isinstance(type_, type) and _is_container_type(type_)
+
+
+def dedup(old: Collection[T]) -> List[T]:
+    new: List[T] = []
+    for x in old:
+        if x not in new:
+            new.append(x)
+    return new
--- a/edb/common/typeutils.py
+++ b/edb/common/typeutils.py
@@ -61,11 +61,3 @@ def is_container(obj):
 
 def is_container_type(type_):
     return isinstance(type_, type) and _is_container_type(type_)
-
-
-def dedup(old: Collection[T]) -> List[T]:
-    new: List[T] = []
-    for x in old:
-        if x not in new:
-            new.append(x)
-    return new
--- a/edb/ir/typeutils.py
+++ b/edb/ir/typeutils.py
@@ -1042,6 +1042,20 @@ def replace_pathid_prefix(
         dir = part.rptr_dir()
         assert dir
 
+        if (
+            isinstance(ptrref, irast.TupleIndirectionPointerRef)
+            and result.target.collection == 'tuple'
+        ):
+            # For tuple indirections, we want to update the target
+            # type when we get mapped to a subtype.
+            idx = get_tuple_element_index(ptrref)
+            target = result.target
+            if target.id != target.subtypes[idx].id:
+                ptrref = ptrref.replace(
+                    out_source=target,
+                    out_target=target.subtypes[idx],
+                )
+
         if ptrref.source_ptr:
             result = result.ptr_path()
         result = result.extend(
--- a/edb/ir/utils.py
+++ b/edb/ir/utils.py
@@ -29,7 +29,7 @@ import json
 from edb import errors
 
 from edb.common import ast
-from edb.common.typeutils import dedup
+from edb.common import ordered
 
 from edb.edgeql import qltypes as ft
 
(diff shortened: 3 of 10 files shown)


edb/common/typeutils.py
```diff
@@ -61,3 +61,11 @@ def is_container(obj):
 
 def is_container_type(type_):
     return isinstance(type_, type) and _is_container_type(type_)
+
+
+def dedup(old: Collection[T]) -> List[T]:
+    new: List[T] = []
+    for x in old:
```
elprans, line 68: Since this looks like a generic helper, there will be temptations to use it for larger collections, so maybe use an `OrderedSet` here?  Or just use it directly at call sites?

elprans (approved):

edb/common/typeutils.py
msullivan: For some reason I had in my head that the `dedup` implementation I put in here used a `set` to track whether things had been seen while building up an ordered result, though obviously it did not.  I'll just use an OrderedSet.

Issue #6118, opened 2023-09-20
Document how to select an item via a known backlink


raddevon:
https://stackoverflow.com/questions/74501898/how-to-update-a-nested-object-in-edgedb/74869015#74869015

PR #6225, opened 2023-10-04
Add known backlink select example

Dhghomon:
Adds an example of using a backlink to select an item such as during an update (see the Stackoverflow question inside https://github.com/edgedb/edgedb/issues/6118 )

This one is a bit tricky as selecting via a backlink is probably not the first thing that comes to a user's mind, and we also don't want this example to give the impression that you should prefer using backlinks to filter as opposed to any other method that might come to mind - which is why there is a sentence mentioning that there are other ways to filter. It should just be a neat example of something you *can* do.


raddevon:
I feel like this is more likely to be tripped over if we move it to the `select` documentation (`docs/edgeql/select.rst`). It's tangentially related to paths, but I think it's more directly related to selecting. This could be a new section under filtering alongside "Filtering by ID" and "Nested filters" — maybe "Filtering by known backlink" or something similar. Content is good though.

I do wonder if we should add a note that in practice, you would probably want to achieve this `owned` property as a computed property of the instrument that checks to see if a backlink exists (I think this would work.), just so readers don't try to replicate this example in their projects. Or maybe there's another example that wouldn't have a tidier alternative you could use in practice… 🤔

docs/edgeql/paths.rst
```diff
@@ -241,6 +241,61 @@ that the type name (in this case ``User``) doesn't need to be specified.
...
+.. code-block:: edgeql-repl
+
+    db> insert Instrument { 
+    ...   description := "Tenor Stratocaster ukulele", 
+    ...   owned := false 
+    ... };
+    {default::Instrument {id: 0d3b3bc8-6013-11ee-8bab-6b7c6f4da55a}}
+
+Then a ``Musician`` with a link to the instrument. (Note the ability to cast
+from a ``uuid`` to an object type, which was added in EdgeDB 3.0!)
```
raddevon, line 275: Love that you called back here. I would cross-link to the docs on that as well. (This might require adding a label if that section doesn't already have one.)

raddevon (changes requested):

Dhghomon:
> I feel like this is more likely to be tripped over if we move it to the `select` documentation (`docs/edgeql/select.rst`). It's tangentially related to paths, but I think it's more directly related to selecting. This could be a new section under filtering alongside "Filtering by ID" and "Nested filters" — maybe "Filtering by known backlink" or something similar. Content is good though.
> 
> I do wonder if we should add a note that in practice, you would probably want to achieve this `owned` property as a computed property of the instrument that checks to see if a backlink exists (I think this would work.), just so readers don't try to replicate this example in their projects. Or maybe there's another example that wouldn't have a tidier alternative you could use in practice… 🤔
> 
> **Update:** You know, this change could be as simple as adding some other arbitrary property that you change on the instrument besides `owned`. Something like `color`. "I saw Bert perform recently and noticed his ukelele is , so I'm going to go back and update the object with that information."

One tough part about moving to select.rst is that that page has its own schema with the superheroes. Select is also a bit over three times the size of the page on paths. I'll move it over somewhere inside select though if you're feeling adamant about it.

As for the example, I've put it together to follow this answer on Stackoverflow as closely as possible:

update Bar
  filter .<bar.id = <uuid>'df492862-80aa-11ed-832e-e72d12452736'
  set { baz := false };

And I guess in that case too a backlink in the schema would be recommended.

Your color suggestion works I think (and have changed the example to match), since it's just a property you could quickly tack on and doesn't feel too 'linky'.

raddevon:
I think `select` much more closely matches what a user is trying to do here, and alongside the other "Filtering" sub-sections seems like the more natural home. In my head, this information is _about_ filtering and merely _uses_ a path rather than being _about_ a path and merely _using_ filtering. I'm open to arguments going in the other direction though. We can poll in Slack if you want to see what the others think.

I don't think balancing the length of the two pages better is the right consideration here for deciding where to place it. It seems natural to me that the `select` page is much longer than path since it's such a fundamental thing people are doing with their database. Paths are fundamental too, but you could get a long way with EdgeDB without reading that page by intuiting what's happening. (You're certainly better off reading it, but it doesn't feel like required reading to the same degree as the `select` documentation.)

raddevon:
Meant to add we could adapt the query to the schema on the `select` page by selecting a `Movie` based on one of its characters.

Dhghomon:
> Meant to add we could adapt the query to the schema on the `select` page by selecting a `Movie` based on one of its characters.

Yeah, I was going to say that I have an idea for a nice query that should work for the select page schema. I'll give it a shot and see how it feels.

Dhghomon:
> Meant to add we could adapt the query to the schema on the `select` page by selecting a `Movie` based on one of its characters.

Okay, second attempt is now up using the schema inside the select page. Made sure to end it with a note that you don't need to use backlinks for such an operation, just another tool in one's belt that can make a query pretty concise.

docs/edgeql/select.rst
```diff
@@ -843,6 +843,48 @@ shapes just like a non-computed link.
   } filter .name = "Iron Man";
 
 
+Using backlinks to find items
+-----------------------------
```
raddevon, line 847:
```suggestion
Filtering on a known backlink
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

- Change the heading to more closely mirror our other "filtering" sections
- Change the heading level so that this falls under "Backlinks" instead of adjacent to it

docs/edgeql/select.rst
```diff
@@ -843,6 +843,48 @@ shapes just like a non-computed link.
...
+
+Another handy use for backlinks is using them to filter and find items
+such as during an update. This can work as a nice shortcut when you have
+the id of one object that links to a second object without a link back
+to the first.
+
+Spider-Man's villains always have a grudging respect for him, and their names
+can be changed to reflect that if we know the id of a movie that they starred
+in. (Note the ability to :ref:`cast <ref_eql_casts>` from a ``uuid`` to an
```
raddevon, line 856: I think it would be more useful here to link to the specific part about casting from a UUID since that's the functionality you're calling out. You might have to add a label to be able to link to it directly. (I don't think it has one yet.) If you do, I would also link the whole phrase: "cast from a uuid" (Note: you won't be able to keep `uuid` as inline code if you do this because RST doesn't want any inline styles on links.)

docs/edgeql/select.rst
```diff
@@ -843,6 +843,48 @@ shapes just like a non-computed link.
   } filter .name = "Iron Man";
 
 
+Using backlinks to find items
+-----------------------------
+
+Another handy use for backlinks is using them to filter and find items
+such as during an update. This can work as a nice shortcut when you have
```
raddevon, line 850: Since we're in the `select` docs, I think our example needs to be a `select` rather than an `update`. We have already talked about computed fields at this point on the page, so your example can use one and end up being practically the same. Just `select` instead of `update` and, since you're not actually updating anything, add a shape with a computed that gives your new name to Spiderman nemeses. Something like this:

```
select Villain filter .<characters =
  <Movie><uuid>'6c60c28a-5c03-11ee-99ff-dfa425012a05' {
    name := .name ++ ', who got to see Spider-Man!'
};
```
Then you can just show the output of this and skip the `select`.

Your without backlink example could look like this:
```
with movie := 
  <Movie><uuid>'6c60c28a-5c03-11ee-99ff-dfa425012a05',
select movie.characters[is Villain] {
  name := .name ++ ', who got to see Spider-Man!'
};
```

Not tested, but I think it should work. At any rate, I think you get what I'm going for. Just adapting what you have to a `select` to better fit the context.

docs/edgeql/select.rst
```diff
@@ -843,6 +843,48 @@ shapes just like a non-computed link.
...
+        'Green Goblin, who got to see Spider-Man!',
+        'Doc Ock, who got to see Spider-Man!',
+      }
+
+In other words, "update every ``Villain`` object that the ``Movie`` object
+of this id links to via a link called ``characters``".
+
+A backlink is naturally not required, however. The same update without
+traversing a backlink would look like this:
+
```
raddevon, line 881:
```suggestion

.. code-block:: edgeql-repl

```

Before you push your final changes, run the site dev server to make sure they build locally. Let me know if you're having trouble building it generally. We can try to figure out the issue and make sure it builds for you.

raddevon (changes requested):
I like your example here. I think we need to adapt it to a `select` since that's the page we're on, but by adding a shape with a computed, I think we can preserve it pretty well.

docs/edgeql/select.rst
```diff
@@ -843,6 +843,46 @@ shapes just like a non-computed link.
   } filter .name = "Iron Man";
 
 
+Filtering on a known backlink
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Another handy use for backlinks is using them to filter and find items
+when doing a ```select`` (or an ``update`` or other operation, of course).
+This can work as a nice shortcut when you have the id of one object that
```
raddevon, line 851: I would either style this as "ID" or make it clear you are talking about the property name by making "id" inline code: "`id`"

Whichever you prefer, propagate it to the other instances.

raddevon (changes requested):
Content is 🤌. Just some very minor suggestions.

I think I still feel like this should go under "Filtering" instead of "Backlinks" though since "filter" will be the verb the user wants to do. Same reason we probably wouldn't want to take our "Filtering by ID" sub-section and move it to wherever we talk about IDs.

Dhghomon:
> Content is 🤌. Just some very minor suggestions.
> 
> I think I still feel like this should go under "Filtering" instead of "Backlinks" though since "filter" will be the verb the user wants to do. Same reason we probably wouldn't want to take our "Filtering by ID" sub-section and move it to wherever we talk about IDs.

Tough to say, because it would be nice to have everything filtering-related inside the filter section but conceptualizing and working with backlinks tends to take the most mental effort and would feel odd to make a quick mention of them in the filtering section before finally waiting four more sections to get around to just what this backlink actually was. Especially since it's the backlink section on this page that leads to the larger reference inside paths.rst, and a user wondering just what a backlink is would have to first make their way down the page to find out. To sum up:

Order of operations for a reader with this PR as is:

- Read about filtering
- Read about backlinks, see quick note about how backlinks work for filtering too, neat. Immediate link to more backlink info

Order of operations for reader if moved up to filtering section:

- Read about filtering
- See backlink example, may or may not be clear what a backlink is
- Hopefully continues reading down to the backlink section where more info and link to paths.rst is

Does that change your mind at all? If not then I'll move it up to filtering.

docs/edgeql/select.rst
Dhghomon: Yeah. Ordinarily I'd say backtick but it's such a small word that capitals minus backtick is probably better for visibility so will go with that.

Dhghomon:
> > Order of operations for a reader with this PR as is:
> > 
> > * Read about filtering
> > * Read about backlinks, see quick note about how backlinks work for filtering too, neat. Immediate link to more backlink info
> > 
> > Order of operations for reader if moved up to filtering section:
> > 
> > * Read about filtering
> > * See backlink example, may or may not be clear what a backlink is
> > * Hopefully continues reading down to the backlink section where more info and link to paths.rst is
> > 
> > Does that change your mind at all? If not then I'll move it up to filtering.
> 
> I feel like this assumes a user is _reading_ the docs versus _referencing_ them, which I think will be the exception not the rule. (This could be my bias because I never "read" docs from front to back. I may read a tutorial of a "getting started" guide, but I reference almost everything else.) If they're referencing this, organizing around the verbs become more important than a logical progression. I still vote for organizing this with other "filtering" sections, but I'm open to other arguments against.

I think I'm going to have to agree with this now. I'm a top-to-bottom reader myself 90% of the time but this select page is so weighty that I found my eyes drifting to the slider on the right where the category selector is. So up it goes to the filtering section..

PR #6227, opened 2023-10-04
+Accessing types in polymorphic queries

Dhghomon:
Getting to https://github.com/edgedb/edgedb/issues/6086 now that the select page cleanup is merged. Two points stand out here that I felt were important:

* The best way to demonstrate the utility of this is by reminding that the `Default::SomeType` output you see is just a convenience in the REPL but not a parameter you can access, and output in <json> is the best way to show this up front
* A bit of explanation that `__type__` isn't just magic, and that the splat operator is a good way to get a look into the internals. I remember way back in the day just sort of rote remembering `__type__` without knowing exactly what was inside and feels better now with the splat operator able to show you everything.


docs/edgeql/select.rst
```diff
@@ -1026,6 +1026,70 @@ exclusively fetch the ``Movie.characters`` of type ``Hero``.
...
+is simply a convenience of the REPL and not a property that can be accessed.
+This is particularly noticeable if you cast an object to ``json``, making it
+impossible to determine the type if the query is polymorphic:
+
+.. code-block:: edgeql-repl
+
+    db> select Person limit 1;
+    {default::Villain {id: 6c22bdf0-5c03-11ee-99ff-dfaea4d947ce}}
+    db> select <json>Person limit 1;
+    {Json("{\"id\": \"6c22bdf0-5c03-11ee-99ff-dfaea4d947ce\"}")}
```
raddevon, line 1042: The way GitHub is diffing this makes it look a lot more different than it actually is. 😅 I'll explain what I was trying to accomplish here after the suggestion:
```suggestion
While the type of an object is displayed alongside the results of polymorphic
queries run in the REPL, this is simply a convenience of the REPL and not a
property that can be accessed. This is particularly noticeable if you cast an
object to ``json``, making it impossible to determine the type if the query is
polymorphic. First, the result of a query as the REPL presents it with type
annotations displayed:

.. code-block:: edgeql-repl

    db> select Person limit 1;
    {default::Villain {id: 6c22bdf0-5c03-11ee-99ff-dfaea4d947ce}}

Note the type ``default::Villain``, which is displayed for the user's
convenience but is not actually part of the data returned. This is the same
query with the result cast as ``json`` to show only the data returned:

.. code-block:: edgeql-repl

    db> select <json>Person limit 1;
    {Json("{\"id\": \"6c22bdf0-5c03-11ee-99ff-dfaea4d947ce\"}")}

.. note::

    We will continue to cast subesequent examples in this section to ``json``,
    not because this is required for any of the functionality being
    demonstrated, but to remove the convenience type annotations provided by
    the REPL and make it easier to see what data is actually being returned by
    the query.
```

- Clarify that the type annotation only happens in the REPL
- Split the examples with and without `json` cast to drive home the difference we're trying to call attention to
- Add a note that explains the continued use of `json` cast to any readers who might not understand and might wrongly assume you must cast to `json` to get type information back on a query.

docs/edgeql/select.rst
```diff
@@ -1026,6 +1026,70 @@ exclusively fetch the ``Movie.characters`` of type ``Hero``.
...
+    ...    name 
+    ...    } 
+    ...  } limit 1;
+    {Json("{\"__type__\": {\"name\": \"default::Villain\"}}")}
+
+This information can be pulled into the top level by assigning a name to
+the ``name`` property inside ``__type__``:
+
+    db> select <json>Person { n := .__type__.name } limit 1;
+    {Json("{\"n\": \"default::Villain\"}")}
```
raddevon, line 1060:
```suggestion
.. code-block:: edgeql-repl

    db> select <json>Person { type := .__type__.name } limit 1;
    {Json("{\"type\": \"default::Villain\"}")}
```

- Add `code-block` directive so it builds
- Change `n` to `type`. `type` feels more descriptive and since the `Person` type has a `name` property, `n` could be conflated to have something to do with that rather than the name of the type.

docs/edgeql/select.rst
```diff
@@ -1026,6 +1026,70 @@ exclusively fetch the ``Movie.characters`` of type ``Hero``.
...
+
+There is nothing magical about ``__type__``; it is a simple link to an object
+called ``ObjectType`` that leads to all of the possible information to know
+about any type in the database. The splat operator can be used to see this
+object's makeup, while the double splat operator produces too much input to
+show on this page. Playing around with the splat and double splat operator
+inside ``__type__`` is a quick way to get some insight into the internals
+of EdgeDB.
```
raddevon, line 1068: Shifted this slightly to center more around this object's `__type__` link rather than about the `ObjectType` type since we're starting this description from `__type__`. Information is basically the same but the perspective has been shifted.
```suggestion
There is nothing magical about ``__type__``; it is a simple link to an object
of the type ``ObjectType`` which contains all of the possible information to
know about the type of the current object. The splat operator can be used to
see this object's makeup, while the double splat operator produces too much
output to show on this page. Playing around with the splat and double splat
operator inside ``__type__`` is a quick way to get some insight into the
internals of EdgeDB.
```

raddevon (changes requested):
The guts of this are great. I have a few suggestions that I think can clarify some parts of it.

docs/edgeql/select.rst
```diff
@@ -1026,6 +1026,70 @@ exclusively fetch the ``Movie.characters`` of type ``Hero``.
...
+
+.. code-block:: edgeql-repl
+
+    db> select Person limit 1;
+    {default::Villain {id: 6c22bdf0-5c03-11ee-99ff-dfaea4d947ce}}
+    db> select <json>Person limit 1;
+    {Json("{\"id\": \"6c22bdf0-5c03-11ee-99ff-dfaea4d947ce\"}")}
+
+The type of an object is found inside ``__type__`` which is a link that
+carries various information about a type, including its ``name``.
```
raddevon, line 1045:
```suggestion
carries various information about the object's type, including its ``name``.
```
I always fear there could be confusion between data type and object type, so I like to be explicit about which we're talking about.

Dhghomon:
Nice suggestions! I applied them all.

raddevon (approved):

PR #6228, opened 2023-10-04
Show how to determine the outcome of upsert

### LINKED ISSUE #5975: Document how to determine the result of an upsert (`insert` or `update`)
```
select (
  insert Person {name := 'Test', phone := '123456789'}
  unless conflict on (.name) else (
    update Person set {phone := '123456789'}
  )
) {is_new := (select not exists (select Person filter .name = 'Test'))};
```

`name` is exclusive here. Your query isn't really an upsert since there's no `update`, but you could use the same technique there too, I believe. This leverages the fact that, since everything here happens in a single transaction, the `Person` will not be found with a `select` if they are inserted with this query and checks for that with the computed `is_new`.

raddevon:
Closes #5975

### DIFF
--- a/docs/edgeql/insert.rst
+++ b/docs/edgeql/insert.rst
@@ -321,6 +321,43 @@ to the ``update`` statement in the ``else`` clause. This updates the
 To learn to use upserts by trying them yourself, see `our interactive upserts
 tutorial </tutorial/data-mutations/upsert>`_.
 
+.. note::
+
+    It can be useful to know the outcome of an upsert. Here's an example
+    showing how you can return that:
+
+    .. code-block:: edgeql-repl
+
+      db> with
+      ...   title := "Eternals",
+      ...   release_year := 2021
+      ... select (
+      ...   insert Movie {
+      ...     title := title,
+      ...     release_year := release_year
+      ...   }
+      ...   unless conflict on .title
+      ...   else (
+      ...     update Movie set { release_year := release_year }
+      ...   )
+      ... ) {
+      ...   is_new := (
+      ...     select not exists (
+      ...       select Movie filter .title = title
+      ...     )
+      ...   )
+      ... };
+      {default::Movie {is_new: true}}
+
+    This technique exploits the fact that a ``select`` will not return an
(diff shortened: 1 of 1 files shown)


docs/edgeql/insert.rst
```diff
@@ -321,6 +321,43 @@ to the ``update`` statement in the ``else`` clause. This updates the
...
+      ...     update Movie set { release_year := release_year }
+      ...   )
+      ... ) {
+      ...   is_new := (
+      ...     select not exists (
+      ...       select Movie filter .title = title
+      ...     )
+      ...   )
+      ... };
+      {default::Movie {is_new: true}}
```
scotttrinh, line 350: Not a blocker: In practice, I feel like people will want to select the movie and also the "new/existing" status at the same time in some kind of `{ is_new, data }` "envelope". Is that easy to show here?

```
      db> with
      ...   title := "Eternals",
      ...   release_year := 2021,
      ...   M := (
      ...     insert Movie {
      ...       title := title,
      ...       release_year := release_year
      ...     }
      ...     unless conflict on .title
      ...     else (
      ...       update Movie set { release_year := release_year }
      ...     )
      ...   )
      ... select {
      ...   is_new := (
      ...     select not exists (
      ...       select Movie filter .title = title
      ...     )
      ...   ),
      ...   data := select M { * },
      ... };
      {default::Movie {is_new: true}}
```

Or something like that? Not even sure if that works as written 😅

scotttrinh (approved):

docs/edgeql/insert.rst
raddevon: Yeah, I think you're probably right. Let me see what I can come up with…

raddevon: Added some more info and another example. I thought about just replacing this example with one returning more data, but I decided to keep it since it more closely mirrors the plain upsert example which returns just the ID since it has no shape. This makes it a bit longer, but I think it's fine. I do agree that this is probably the next question people will have given this example. Might as well get ahead of it.

scotttrinh (approved):
👏

PR #6229, opened 2023-10-04
Auth providers schema update

jaclarke:


edb/lib/ext/auth.edgeql
```diff
@@ -105,10 +98,73 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
...
+
+        create required property display_name: std::str {
+            set readonly := true;
+            set protected := true;
+        };
+
+        create property additional_scope: std::str;
+    };
+
+    # create type ext::auth::OpenIdProviderConfig
```
1st1, line 110: We're not shipping this? cc @scotttrinh

scotttrinh: This is arbitrarily defined OIDC providers, not the three we have built-in (apple, google, azure).

edb/server/protocol/auth_ext/oauth.py
```diff
@@ -28,49 +28,46 @@
...
         )
 
-        (provider_name, client_id, client_secret) = self._get_provider_config(
-            provider_id
-        )
-
         match provider_name:
-            case "github":
+            case "builtin::oauth_github":
                 from . import github
```
1st1, line 46: Do we need conditional imports? If it works with moving them to the top of the file please just do that.

edb/server/protocol/auth_ext/ui.py
```diff
@@ -170,7 +166,7 @@ def render_signup_page(
 
       {_render_error_message(error_message)}
 
-      <input type="hidden" name="provider" value="{provider_id}" />
+      <input type="hidden" name="provider" value="{provider_name}" />
```
1st1, line 169: Change the below email input from `<input type="text">` to `<input type="email">` (ditto for sign in if not there already)

edb/lib/ext/auth.edgeql
1st1: James, let's just kill these lines then. I don't like shipping commented out snippets in code.

edb/server/protocol/auth_ext/oauth.py
```diff
@@ -97,7 +94,7 @@ async def _handle_identity(self, user_info: data.UserInfo) -> data.Identity:
             query="""\
 with
   iss := <str>$issuer_url,
-  sub := <str>$provider_id,
+  sub := <str>$provider_name,
```
scotttrinh, line 97: This should be `provider_id` still actually: it's the ID of the user in the provider's system not our "provider ID". Maybe just change it to `subject`?

scotttrinh (changes requested):

edb/server/protocol/auth_ext/oauth.py
jaclarke: I thought we dropped `provider_id`?

scotttrinh: no, this is literally just a poorly-named variable for the `sub`/`subject` in this query. "subject's ID at the provider's service" = "provider_id". I think `subject` is the correct name now.

jaclarke: Oh right, yeah it's just a variable name. Looks like it caught up in a find/replace. 🤦‍♂️

scotttrinh: yep! it was poorly named before anyway.

scotttrinh (approved):

PR #6233, opened 2023-10-04
Run auth::PKCEChallenge GC periodically

fantix:
- [x] Manual test

### DIFF
--- a/edb/server/protocol/auth_ext/pkce.py
+++ b/edb/server/protocol/auth_ext/pkce.py
@@ -145,19 +145,26 @@ async def delete(db, id: str) -> None:
 
 
 async def _gc(tenant: edbtenant.Tenant):
-    async with taskgroup.TaskGroup() as g:
-        for db in tenant.iter_dbs():
-            if "auth" in db.extensions:
-                g.create_task(
-                    execute.parse_execute_json(
-                        db,
-                        """
-                        delete ext::auth::PKCEChallenge filter
-                            (datetime_of_statement() - .created_at) >
-                            <duration>'10 minutes'
-                        """,
-                    ),
-                )
+    try:
+        async with taskgroup.TaskGroup() as g:
+            for db in tenant.iter_dbs():
+                if "auth" in db.extensions:
+                    g.create_task(
+                        execute.parse_execute_json(
+                            db,
+                            """
+                            delete ext::auth::PKCEChallenge filter
+                                (datetime_of_statement() - .created_at) >
+                                <duration>'10 minutes'
+                            """,
+                        ),
+                    )
+    except Exception as ex:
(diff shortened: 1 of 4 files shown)


edb/server/protocol/auth_ext/pkce.py
```diff
@@ -130,3 +142,42 @@ async def delete(db, id: str) -> None:
...
+                if tenant.accept_new_tasks
+            ]
+            if tasks:
+                done, _ = await asyncio.wait(tasks)
+                for task in done:
+                    ex = task.exception()
+                    if ex is not None:
+                        logger.debug(
+                            "GC of ext::auth::PKCEChallenge failed",
+                            exc_info=ex,
```
1st1, line 178: Probably worth logging which tenant it crashed for?

edb/server/protocol/auth_ext/pkce.py
```diff
@@ -130,3 +142,42 @@ async def delete(db, id: str) -> None:
...
+async def gc(server: edbserver.BaseServer):
+    while True:
+        try:
+            tasks = [
+                tenant.create_task(_gc(tenant), interruptable=False)
+                for tenant in server.iter_tenants()
+                if tenant.accept_new_tasks
+            ]
+            if tasks:
+                done, _ = await asyncio.wait(tasks)
```
1st1, line 172: Is there a potential of race here, when we schedule to do some work for a tenant and they are suspended before the work has started? Warning: I don't have a good grasp on how we implemented multi-tenancy.

fantix: `accept_new_tasks` should guard us against such races, and the tenant will wait for our task to finish if started by `tenant.create_task(..., interruptable=False)`

1st1: sounds good

edb/server/protocol/auth_ext/pkce.py
```diff
@@ -130,3 +142,42 @@ async def delete(db, id: str) -> None:
...
+        async with taskgroup.TaskGroup() as g:
+            for db in tenant.iter_dbs():
+                if "auth" in db.extensions:
+                    g.create_task(
+                        execute.parse_execute_json(
+                            db,
+                            """
+                            delete ext::auth::PKCEChallenge filter
+                                (datetime_of_statement() - .created_at) >
+                                <duration>'10 minutes'
```
scotttrinh, line 158: Not a blocker: I wonder if we should make this duration value some kind of shared const between this and the `get_by_id` function?

scotttrinh (approved):
Cool! Code looks good to me as far as I can read it.

scotttrinh (approved):
Love it!

PR #6234, opened 2023-10-05
Request any configured additional OAuth scope

scotttrinh:
Passes along any configured scopes to the provider in the initial "authorize" request. When we return the token(s) from the Identity Provider, they will be scoped with our default scope (needed to get an email address and subject ID) and any other scope requested here.

### DIFF
--- a/edb/server/protocol/auth_ext/oauth.py
+++ b/edb/server/protocol/auth_ext/oauth.py
@@ -19,7 +19,7 @@
 
 import json
 
-from typing import Any
+from typing import Any, Type
 from edb.server.protocol import execute
 
 from . import github, google, azure, apple
@@ -47,19 +47,22 @@ class Client:
             "additional_scope": additional_scope,
         }
 
+        provider_class: Type[base.BaseProvider]
         match provider_name:
             case "builtin::oauth_github":
-                provider_class = github.GitHubProvider  # type: ignore
+                provider_class = github.GitHubProvider
             case "builtin::oauth_google":
-                provider_class = google.GoogleProvider  # type: ignore
+                provider_class = google.GoogleProvider
             case "builtin::oauth_azure":
-                provider_class = azure.AzureProvider  # type: ignore
+                provider_class = azure.AzureProvider
             case "builtin::oauth_apple":
-                provider_class = apple.AppleProvider  # type: ignore
+                provider_class = apple.AppleProvider
             case _:
                 raise errors.InvalidData(f"Invalid provider: {provider_name}")
 
-        self.provider = provider_class(*provider_args, **provider_kwargs)
+        self.provider = provider_class(
+            *provider_args, **provider_kwargs  # type: ignore
+        )
 
     async def get_authorize_url(self, state: str, redirect_uri: str) -> str:
         return await self.provider.get_code_url(
--- a/edb/server/protocol/auth_ext/oauth.py
+++ b/edb/server/protocol/auth_ext/oauth.py
@@ -38,13 +38,11 @@ class Client:
             *args, edgedb_test_url=base_url, **kwargs
         )
 
-        (client_id, secret, additional_scope) = self._get_provider_config(
-            provider_name
-        )
-        provider_args = (client_id, secret)
+        provider_config = self._get_provider_config(provider_name)
+        provider_args = (provider_config.client_id, provider_config.secret)
         provider_kwargs = {
             "http_factory": http_factory,
-            "additional_scope": additional_scope,
+            "additional_scope": provider_config.additional_scope,
         }
 
         provider_class: Type[base.BaseProvider]
@@ -111,15 +109,15 @@ select (insert ext::auth::Identity {
 
         return data.Identity(**result_json[0])
 
-    def _get_provider_config(
-        self, provider_name: str
-    ) -> tuple[str, str, str | None]:
+    def _get_provider_config(self, provider_name: str):
         provider_client_config = util.get_config(
             self.db, "ext::auth::AuthConfig::providers", frozenset
         )
         for cfg in provider_client_config:
             if cfg.name == provider_name:
-                return (cfg.client_id, cfg.secret, cfg.additional_scope)
+                return data.ProviderConfig(
+                    cfg.client_id, cfg.secret, cfg.additional_scope
+                )
 
         raise errors.MissingConfiguration(
             provider_name, "Provider is not configured"
--- a/edb/server/protocol/auth_ext/oauth.py
+++ b/edb/server/protocol/auth_ext/oauth.py
@@ -38,10 +38,13 @@ class Client:
             *args, edgedb_test_url=base_url, **kwargs
         )
 
-        provider_args = self._get_provider_config(provider_name)
+        (client_id, secret, additional_scope) = self._get_provider_config(
+            provider_name
+        )
+        provider_args = (client_id, secret)
         provider_kwargs = {
             "http_factory": http_factory,
-            "additional_scope": self._get_additional_scope(provider_name),
+            "additional_scope": additional_scope,
         }
 
         match provider_name:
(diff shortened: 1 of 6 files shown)


edb/server/protocol/auth_ext/oauth.py
```diff
@@ -30,44 +30,39 @@ class Client:
...
+            "additional_scope": self._get_additional_scope(provider_name),
+        }
+
         match provider_name:
             case "builtin::oauth_github":
-                self.provider = github.GitHubProvider(
-                    *self._get_provider_config(provider_name),
-                    http_factory=http_factory,
-                )
+                provider_class = github.GitHubProvider  # type: ignore
```
scotttrinh, line 49: I couldn't get `mypy` to shut-up about the type of `provider_class`, even if I annotated the variable like `provider_class: Type[base.BaseProvider]`, so I just silenced it.

This refactor seemed worth it since it feels error prone to manually have every branch of this match have to instantiate the provider with the exact same arguments. Especially an "optional" argument like this, it would be easy to forget to add it in one branch.

tests/test_http_ext_auth.py
```diff
@@ -1229,7 +1215,9 @@ async def test_http_auth_ext_azure_authorize_01(self):
             self.assertEqual(url.scheme, "https")
             self.assertEqual(url.hostname, "login.microsoftonline.com")
             self.assertEqual(url.path, "/common/oauth2/v2.0/authorize")
-            self.assertEqual(qs.get("scope"), ["openid profile email"])
+            self.assertEqual(
+                qs.get("scope"), ["openid profile email offline_access"]
```
scotttrinh, line 1219: Note this scope has the additional `offline_access` scope we configured in the setup for the Azure provider.

tests/test_http_ext_auth.py
```diff
@@ -350,29 +350,29 @@ class TestHttpExtAuth(tb.ExtAuthTestCase):
...
+            client_id := '{uuid.uuid4()}',
         }};
 
         CONFIGURE CURRENT DATABASE
         INSERT ext::auth::AzureOAuthProvider {{
-            secret := <str>'{AZURE_SECRET}',
-            client_id := <str>'{uuid.uuid4()}'
+            secret := '{AZURE_SECRET}',
+            client_id := '{uuid.uuid4()}',
+            additional_scope := 'offline_access',
```
jaclarke, line 367: Should we name this `additional_scopes`? (I assume we allow multiple scopes?)

scotttrinh: I went back and forth on this: is the plural of scope really scopes? Or is it a single scope that is now bigger (increasing the scope of something). The parameter in OAuth2 is called `scope`, so this _does_ feel kinda right.

And, indeed, this is a space-separated list of scope strings.

edb/server/protocol/auth_ext/oauth.py
1st1: That's OK, but what was mypy saying?

scotttrinh:
```
edb/server/protocol/auth_ext/oauth.py:51:34: error: Incompatible types in assignment (expression has type "Type[GoogleProvider]", variable has type "Type[GitHubProvider]")  [assignment]
edb/server/protocol/auth_ext/oauth.py:53:34: error: Incompatible types in assignment (expression has type "Type[AzureProvider]", variable has type "Type[GitHubProvider]")  [assignment]
edb/server/protocol/auth_ext/oauth.py:55:34: error: Incompatible types in assignment (expression has type "Type[AppleProvider]", variable has type "Type[GitHubProvider]")  [assignment]
```

edb/server/protocol/auth_ext/oauth.py
```diff
@@ -30,44 +30,39 @@ class Client:
...
         http_factory = lambda *args, **kwargs: http_client.HttpClient(
             *args, edgedb_test_url=base_url, **kwargs
         )
 
+        provider_args = self._get_provider_config(provider_name)
+        provider_kwargs = {
+            "http_factory": http_factory,
+            "additional_scope": self._get_additional_scope(provider_name),
+        }
+
```
1st1, line 46: Try declaring `provider_class: Type[base.BaseProvider]` here, before `match`

scotttrinh: Yeah, [I did that](https://github.com/edgedb/edgedb/pull/6234#discussion_r1347648942) and it still gave me the exact same error.

edb/server/protocol/auth_ext/oauth.py
```diff
@@ -30,44 +30,39 @@ class Client:
...
         self.db = db
 
         http_factory = lambda *args, **kwargs: http_client.HttpClient(
             *args, edgedb_test_url=base_url, **kwargs
         )
 
+        provider_args = self._get_provider_config(provider_name)
+        provider_kwargs = {
+            "http_factory": http_factory,
+            "additional_scope": self._get_additional_scope(provider_name),
```
1st1, line 44: Can we fold `_get_additional_scope` into `_get_provider_config`? It's not the fastest APIs I think

scotttrinh: It's a little tricky because we're returning a tuple and `additional_scope` is a keyword-only argument, but I think I can still do that. Will fix.

tests/test_http_ext_auth.py
1st1: I think it should be "scopes" -- as in the purpose of this is to extend the information we request and potentially add more information "bits" -- so using a plural "scopes" here seems like a better option to me.

edb/server/protocol/auth_ext/oauth.py
1st1: Weird. @msullivan can you teach us to use mypy properly here? I'm mostly curious at this point, it's not urgent

1st1: (we can merge as is obviously)

edb/server/protocol/auth_ext/oauth.py
scotttrinh: @1st1 ⚡ [`948aaa4` (#6234)](https://github.com/edgedb/edgedb/pull/6234/commits/948aaa4b9fab456b642c5481dc1b8b6707389b65)

How does this look?

edb/server/protocol/auth_ext/oauth.py
elprans: Not exactly same:

```
edb/server/protocol/auth_ext/oauth.py:61:58: error: Argument 2 to "BaseProvider" has incompatible type "**Dict[str, object]"; expected "str"  [arg-type]
edb/server/protocol/auth_ext/oauth.py:61:58: error: Argument 2 to "BaseProvider" has incompatible type "**Dict[str, object]"; expected "Optional[str]"  [arg-type]
edb/server/protocol/auth_ext/oauth.py:61:58: error: Argument 2 to "BaseProvider" has incompatible type "**Dict[str, object]"; expected "Callable[..., HttpClient]"  [arg-type]
```

You can silence this one though for now as you're doing dynamic argument passing and properly typing that is only possible in 3.12

edb/server/protocol/auth_ext/oauth.py
```diff
@@ -110,13 +108,15 @@ async def _handle_identity(self, user_info: data.UserInfo) -> data.Identity:
...
+    def _get_provider_config(
+        self, provider_name: str
+    ) -> tuple[str, str, str | None]:
         provider_client_config = util.get_config(
             self.db, "ext::auth::AuthConfig::providers", frozenset
         )
         for cfg in provider_client_config:
             if cfg.name == provider_name:
-                return (cfg.client_id, cfg.secret)
+                return (cfg.client_id, cfg.secret, cfg.additional_scope)
```
1st1, line 119: Alternatively you can just return `cfg` here.

scotttrinh: Oh, that's a good idea. I think it lacks some typing that made this more awkward than hiding it behind this function, I'll take a look.

edb/server/protocol/auth_ext/oauth.py
1st1: Better! You can also just return `cfg` from that method, that would probably be more future proof, but up to you. This PR is good to go as far as I'm concerned

edb/server/protocol/auth_ext/oauth.py
scotttrinh: Hmm... when I tried it locally it gave me literally the same error, maybe I was in a weird state with unsaved files or something. Fixed in [`8483e46` (#6234)](https://github.com/edgedb/edgedb/pull/6234/commits/8483e46fadc409a525d05c68ffb13e25a5768d6e)

edb/server/protocol/auth_ext/oauth.py
1st1: You can declare a subtype of `typing.NamedTuple` and cast `cfg` into that.

edb/server/protocol/auth_ext/oauth.py
scotttrinh: That _does_ seem better. Should I make a dataclass for `ProviderConfig` to get proper typing when accessing the properties of the returned config object?

PR #6235, opened 2023-10-05
Make ids for extension objects stable

msullivan:


1st1 (approved):
That was quick! Any way we can add a test for this, if you have such tests for the std schema?

PR #6236, opened 2023-10-05
Document missing environment variables

### LINKED ISSUE #6113: Add `EDGEDB_DOCKER_APPLY_MIGRATIONS` and `EDGEDB_SERVER_HTTP_ENDPOINT_SECURITY` to the list of environment variables
Detail on `EDGEDB_DOCKER_APPLY_MIGRATIONS`: https://www.edgedb.com/docs/guides/deployment/docker#schema-migrations

Add here: https://www.edgedb.com/docs/reference/environment#environment-variables

raddevon:
In addition to adding missing env vars, I alphabetized all the variables and did some very light copy editing.

I came out of this with a few questions:
- Does `EDGEDB_SERVER_GENERATE_SELF_SIGNED_CERT` need to be documented since it is also set as a side-effect of setting `EDGEDB_SERVER_TLS_CERT_MODE` to `generate_self_signed`? Is there a use case for setting it alone?
- Should `EDGEDB_SERVER_UID` be documented? If so, what is it?
- Should `EDGEDB_SERVER_ALLOW_INSECURE_HTTP_CLIENTS` be documented? How does it interact with `EDGEDB_SERVER_HTTP_ENDPOINT_SECURITY` and what's the distinction between the two?
- What is the difference between `EDGEDB_SERVER_SKIP_MIGRATIONS` and `EDGEDB_DOCKER_APPLY_MIGRATIONS`? How do they interact?

Closes #6113

### DIFF
--- a/docs/reference/environment.rst
+++ b/docs/reference/environment.rst
@@ -26,6 +26,13 @@ variants.
 Supported variables
 -------------------
 
+EDGEDB_DOCKER_ABORT_CODE
+........................
+
+If the process fails, the arguments are logged to stderr and the script is
+terminated with this exit code. Default is ``1``.
+
+
 EDGEDB_DOCKER_APPLY_MIGRATIONS
 ..............................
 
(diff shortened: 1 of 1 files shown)


fmoor:
> Does EDGEDB_SERVER_GENERATE_SELF_SIGNED_CERT need to be documented since it is also set as a side-effect of setting EDGEDB_SERVER_TLS_CERT_MODE to generate_self_signed? Is there a use case for setting it alone?

I believe this is deprecated https://github.com/edgedb/edgedb/blob/d0929e95416f6c19e4dd6f690d0b1293a42e80f1/edb/server/args.py#L775

> Should EDGEDB_SERVER_UID be documented? If so, what is it?

This is the user id that should run the `edgedb-server` binary in the docker container. I'm not sure if it should be documented though. @elprans?

> Should EDGEDB_SERVER_ALLOW_INSECURE_HTTP_CLIENTS be documented? How does it interact with EDGEDB_SERVER_HTTP_ENDPOINT_SECURITY and what's the distinction between the two?

This one is also deprecated. https://github.com/edgedb/edgedb/blob/d0929e95416f6c19e4dd6f690d0b1293a42e80f1/edb/server/args.py#L1077

> What is the difference between EDGEDB_SERVER_SKIP_MIGRATIONS and EDGEDB_DOCKER_APPLY_MIGRATIONS? How do they interact?

`EDGEDB_SERVER_SKIP_MIGRATIONS` is deprecated. `EDGEDB_DOCKER_APPLY_MIGRATIONS` should be used instead. https://github.com/edgedb/edgedb-docker/blob/8f42796be51b739ab4438a1826a20c1ae796909f/docker-entrypoint-funcs.sh#L477

docs/reference/environment.rst
```diff
@@ -26,6 +26,93 @@ variants.
 Supported variables
 -------------------
 
+EDGEDB_DOCKER_ABORT_CODE
```
fmoor, line 29: `edgedb-server` doesn't know anything about the `EDGEDB_DOCKER_*` variables. These are only used by the docker image. At least some of these are already documented here https://github.com/edgedb/edgedb/blob/master/docs/guides/deployment/docker.rst Maybe it makes sense to keep them separate?

fmoor: Actually there are others that the server doesn't know about too like `EDGEDB_SERVER_BOOTSTRAP_SCRIPT_FILE` that don't start with `EDGEDB_DOCKER_*`.

raddevon: > edgedb-server doesn't know anything about the EDGEDB_DOCKER_* variables. These are only used by the docker image. Maybe it makes sense to keep them separate?

Would you keep them on this page but divide them under separate headings?

> Actually there are others that the server doesn't know about too like EDGEDB_SERVER_BOOTSTRAP_SCRIPT_FILE that don't start with EDGEDB_DOCKER_*.

Is this a Docker env var that just didn't get the prefix, or is it something different?

raddevon:
@fmoor Should deprecated variables stay in the docs but with a warning, or is it better to delete them altogether?

docs/reference/environment.rst
fmoor: > Would you keep them on this page but divide them under separate headings?

The docker environment variables seem to cohere more with the docker deployment instructions rather than with the server environment variables to me. Readers only care about them when deploying with docker. But this may be subjective :shrug:.

> Is this a Docker env var that just didn't get the prefix, or is it something different?

Yea. I _think_ this is an OG docker var that was created before we had this naming convention.

fmoor:
> is it better to delete them altogether?

This is probably not a good option because people reading existing infra code will try to look them up and find nothing. On the other hand, I don't know how long we want to keep deprecated things around. I think databases tend to keep things for a long time :thinking:

docs/reference/environment.rst
raddevon: This change was originally inspired by [a Discord thread](https://discord.com/channels/841451783728529451/1058029984486273108) in which a user was expecting to find Docker env vars in this file as well, so it seems like people are looking for them here. It seems reasonable to me, and I think we can do it without any harm as long as we separate them and make clear that those Docker variables are invisible to the server. I'll give it a shot and see what I can come up with…

fmoor: Sure, sounds good. Maybe the environment variable section in the docker deploy guide should just be a link to this page then?

raddevon: > Actually there are others that the server doesn't know about too…

Is there an easy way I could discover which variables the server is aware of and which it isn't? I'll try to sort them into sections.

fmoor: I grep the server repo and the edgedb-docker repos.

raddevon:
I think I have this in shape, @fmoor, if you get a chance to take another look. I think I found all the variables that work only in Docker and got everything sorted into those two categories. Added the missing deprecated variables too, although I don't know how to determine which version they were deprecated in. That might be nice to add if you know or know how I can find out, but I don't think it's necessarily a showstopper given we're going from not having these variables documented at all to having them, albeit without the version they were deprecated in.

fmoor (approved):

fmoor:
> I don't know how to determine which version they were deprecated in.

Only thing I can think of is check out each release tag and have a look at each deprecated variable.

raddevon:
> Only thing I can think of is check out each release tag and have a look at each deprecated variable.

I couldn't find any tags on the edgedb-docker repo or anything else that appeared to indicate which commits mapped to which version, so I'll go ahead and merge without versions for now. We can circle back later if there's another way. I did find two other deprecated env vars I had missed, so it was worthwhile to revisit before the merge!

Thanks for all your help, @fmoor! 🙏

PR #6239, opened 2023-10-06
Pass PKCE `challenge` through to authorize redirect

scotttrinh:
We require that `authorize` calls contain a `challenge` string to use PKCE to protect the auth token from being exposed in untrusted contexts. This passes the provided `challenge` string from the application through.

We would like to allow this behavior to be configurable, where developers could opt into the implicit flow instead if they are ok with that security trade-off.

### DIFF
--- a/edb/server/protocol/auth_ext/http.py
+++ b/edb/server/protocol/auth_ext/http.py
@@ -82,7 +82,7 @@ class Router:
                     oauth_client = oauth.Client(
                         db=self.db,
                         provider_name=provider_name,
-                        base_url=test_url,
+                        base_url=test_url
                     )
                     await pkce.create(self.db, challenge)
                     authorize_url = await oauth_client.get_authorize_url(
@@ -304,7 +304,7 @@ class Router:
                             redirect_params = urllib.parse.urlencode(
                                 {
                                     "error": str(ex),
-                                    "email": data.get('email', ''),
+                                    "email": data.get('email', '')
                                 }
                             )
                             redirect_url = (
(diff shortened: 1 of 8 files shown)


edb/server/protocol/auth_ext/http.py
```diff
@@ -589,13 +596,16 @@ async def handle_request(
                                 query, 'error'
                             ),
                             email=_maybe_get_search_param(query, 'email'),
+                            challenge=_get_search_param(
+                                query, 'challenge'
+                            ),
```
fantix, line 601: This breaks the "Sign in" link on the "Sign up" page.

scotttrinh: ah, yeah, I can pass it to sign in as well so it can pass them on. I originally did that assuming we had the social buttons there too. Probably want to add them?

jaclarke: Maybe we could put it in a session cookie? So if that 'challenge' query param exists, we use it and put it in a session cookie, if not then we check the cookie. Then we can also handle the less secure flow, by checking if neither exist, then we create a new verifier and put that in a session cookie, and use that to derive the challenge when we need it.

scotttrinh: That's a great idea @jaclarke 🙌

scotttrinh: @jaclarke @fantix ⚡ [`6e84957` (#6239)](https://github.com/edgedb/edgedb/pull/6239/commits/6e84957532ac2be1a796ac004deae708477abd8b)

Gotta love writing cookie parsers from scratch 😅

fantix (approved):
I'll double check the cookies part once I have my keyboard, but this looks good overall!

fantix:
Done - but I think there is a merge conflict against #6252

1st1 (approved):

PR #6240, opened 2023-10-06
Improvements to string function page

Dhghomon:
Noticed today that this page could use some care.

Notable changes:

* Change to third person singular tense for the function intros ("Test if a string..." -> "Tests if a string...")
* Some extra examples, especially in re_replace - the `flags` parameter was new to me and inclarity on how to use it was what led me to notice this page
* General rewriting. One notable change is that I've removed all the instances of italics for *string*. Here the word *string* is italicized to indicate that it's a parameter name, but it feels awkward when reading as gives the impression of a type name when the type is actually *str*, plus the context I think is clear enough that any reader will know that string means the input string.
* The deprecated functions str_ltrim and str_rtrim are still referenced here so changed them. Don't see them anywhere else in the documentation now except to mention that they are deprecated.

The page looks good to me now but feel free to hop in and propose any other changes while the PR is open.

### DIFF
--- a/docs/stdlib/string.rst
+++ b/docs/stdlib/string.rst
@@ -36,13 +36,13 @@ Strings
       - :eql:func-desc:`to_str`
 
     * - :eql:func:`len`
-      - Return string's length.
+      - Returns a string's length.
 
     * - :eql:func:`contains`
-      - Test if a string contains a substring.
+      - Tests if a string contains a substring.
 
     * - :eql:func:`find`
-      - Find index of a substring.
+      - Finds the index of a substring.
 
     * - :eql:func:`str_lower`
       - :eql:func-desc:`str_lower`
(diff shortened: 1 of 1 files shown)


scotttrinh (approved):
Changes look good to me! Is this tense consistent across the other function pages as well?

Dhghomon:
> Changes look good to me! Is this tense consistent across the other function pages as well?

Beyond this string page I don't think it's 100% yet, but I have noticed that pretty much all and especially the newer pages do use this tense which is fortunate. One example from the page on sequences:

`Increments the given sequence to its next value and returns that value.`

A full extra PR just changing tense might be a bit overkill considering all the other things to do but we could keep an eye out them while working on other tasks.

raddevon:
@Dhghomon Do you want to add something to the contributing guide about tense, so that we have a guideline going forward for what tense we should be using? Then, we can just slowly bring everything in line with that as we continue working on the docs.

PR #6241, opened 2023-10-06
Update docs upsert example

jaclarke:
@raddevon I came across a need to check result of an upsert the other day, and this is what I came up with. Then I checked if we had this documented, and it seems we do, but to me the existing example seems more complicated with the need to do a filter again on the same props as in the insert conflict clause. What do you think?

### DIFF
--- a/docs/edgeql/insert.rst
+++ b/docs/edgeql/insert.rst
@@ -365,7 +365,7 @@ tutorial </tutorial/data-mutations/upsert>`_.
 
       db> with
       ...   title := "Eternals",
-      ...   release_year := 2021
+      ...   release_year := 2021,
       ...   movie := (
       ...     insert Movie {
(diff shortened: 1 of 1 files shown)


raddevon (approved):
Beautiful! This example didn't exist until this week 😅, so we'll get your change merged and hardly anyone will have seen the more complicated one. Definitely nicer to avoid having to select on the potentially conflicting property.

docs/edgeql/insert.rst
```diff
@@ -369,25 +366,22 @@ tutorial </tutorial/data-mutations/upsert>`_.
       db> with
       ...   title := "Eternals",
       ...   release_year := 2021
```
jaclarke, line 368:
```suggestion
      ...   release_year := 2021,
```

PR #6243, opened 2023-10-06
Cache compiled system query globally

fantix:
This PR adds a new switch `cached_globally` in `parse_execute_json()`. When set, the compiled query is cached globally among all tenants and databases. Thus, it's the caller's duty to make sure only the shared schema (std schema, or the extension schemas in the user schema) is used in the query.

- [x] Add test for multiple user schemas

### DIFF
--- a/edb/server/compiler/compiler.py
+++ b/edb/server/compiler/compiler.py
@@ -921,6 +921,45 @@ class Compiler:
 
         return compile(ctx=ctx, source=source), ctx.state
 
+    def compile_system(
+        self,
+        source: edgeql.Source,
+        output_format: enums.OutputFormat,
+        expect_one: bool,
+        implicit_limit: int,
+        inline_typeids: bool,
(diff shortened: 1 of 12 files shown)


edb/server/compiler/compiler.py
```diff
@@ -921,6 +921,45 @@ def compile_in_tx(
...
+        expect_one: bool,
+        implicit_limit: int,
+        inline_typeids: bool,
+        inline_typenames: bool,
+        protocol_version: defines.ProtocolVersion,
+        inline_objectids: bool = True,
+        json_parameters: bool = False,
+    ):
+        state = dbstate.CompilerConnectionState(
+            user_schema=s_schema.EMPTY_SCHEMA,
```
fantix, line 937: I think this dodged the `ext::auth::` schema 🥀

msullivan (approved):

PR #6244, opened 2023-10-06
Elevate link prop insert, update, query docs to "Links" page

### LINKED ISSUE #6124: Show how to update link properties on "Links" doc page
We show this already in the "Using link properties" cheatsheet, but it's pretty fundamental so could probably also be in the ["Links" doc page section on link properties](https://www.edgedb.com/docs/datamodel/links#link-properties).

Also on [the JS "Update" documentation](https://www.edgedb.com/docs/clients/js/update#updating-links) and maybe in other client libraries' documentation as well.

raddevon:
The information we have on link properties on the "Links" page doesn't currently give you enough to start using them. It has always seemed weird to me that information that is fundamental to using the feature is hidden away in a cheatsheet. I feel like a cheatsheet should be a place to catch interesting use cases but not for the foundational info the user needs to use the feature. Elevating that info as a result.

Closes #6124

### DIFF
--- a/docs/datamodel/links.rst
+++ b/docs/datamodel/links.rst
@@ -586,13 +586,13 @@ a shape on the link.
 
 .. code-block:: edgeql-repl
 
-    edgedb> select Person {
-    .......   name,
-    .......   family_members: {
-    .......     name,
-    .......     @relationship
-    .......   }
-    ....... };
+    db> select Person {
+    ...   name,
+    ...   family_members: {
+    ...     name,
+    ...     @relationship
+    ...   }
+    ... };
     {
       default::Person {name: 'Alice', family_members: {}},
       default::Person {
--- a/docs/datamodel/links.rst
+++ b/docs/datamodel/links.rst
@@ -538,7 +538,7 @@ property allows us to modify the link property of the existing link.
         # 🚫
         insert Movie {
           title := 'The Incredible Hulk',
-          actors := {(
+          characters := {(
               select Person {
                 @character_name := 'The Hulk'
               } filter .name = 'Mark Ruffalo'
@@ -560,12 +560,12 @@ property allows us to modify the link property of the existing link.
         # ✅
         insert Movie {
           title := 'The Incredible Hulk',
-          actors := assert_distinct((
-            with characters := {
+          characters := assert_distinct((
+            with actors := {
               ('The Hulk', 'Mark Ruffalo'),
               ('Abomination', 'Tim Roth')
             },
-            for character in characters union (
+            for actor in actors union (
               select Person {
                 @character_name := character.0
               } filter .name = character.1
--- a/docs/datamodel/links.rst
+++ b/docs/datamodel/links.rst
@@ -497,7 +497,7 @@ linked object being added. Be sure to prepend the link property's name with
         }
         filter .name = "Alice"
       )
-    }
+    };
 
 The shape could alternatively be included on an insert if the object being
 linked (the ``Person`` named "Alice" in this example) is being inserted as part
@@ -522,7 +522,7 @@ already-established link between the two:
         }
         filter .name = "Alice"
       )
-    }
+    };
 
 Using ``select .family_members`` here with the shape including the link
 property allows us to modify the existing link's link property.
--- a/docs/datamodel/links.rst
+++ b/docs/datamodel/links.rst
@@ -509,7 +509,7 @@ already-established link between the two:
 
 .. code-block:: edgeql
 
-    update Person {
+    update Person set {
       name := "Bob",
       family_members := (
         select .family_members {
--- a/docs/datamodel/links.rst
+++ b/docs/datamodel/links.rst
@@ -453,6 +453,10 @@ the *nature/strength* of the relationship.
           }
         }
 
+.. note::
+
+    Link properties cannot be made required. They are always optional.
+
 Above, we model a family tree with a single ``Person`` type. The ``Person.
 family_members`` link is a many-to-many relation; each ``family_members`` link
 can contain a string ``relationship`` describing the relationship of the two
@@ -603,7 +607,9 @@ a shape on the link.
     In the query results above, Alice appears to have no family members even
     though we know that, if she is Bob's step-sister, he must be her
     step-brother. We would need to update Alice manually before this is
-    reflected in the database.
+    reflected in the database. Since link properties cannot be required, not
+    setting one is always allowed and results in the value being the empty set
+    (``{}``).
 
 .. note::
 
--- a/docs/datamodel/links.rst
+++ b/docs/datamodel/links.rst
@@ -485,15 +485,15 @@ linked object being added. Be sure to prepend the link property's name with
 
 .. code-block:: edgeql
 
-  insert Person {
-    name := "Bob",
-    family_members := (
-      select detached Person {
-        @relationship := "sister"
-      }
-      filter .name = "Alice"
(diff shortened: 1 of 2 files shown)


docs/datamodel/links.rst
```diff
@@ -475,11 +475,134 @@ in these cases due to better ergonomics of selecting, updating, and even
...
+      }
+    },
+  }
+
+.. note::
+
+    In the query results above, Alice appears to have no family members even
+    though we know that, if she is Bob's step-sister, he must be her
+    step-brother. We would need to update Alice manually before this is
+    reflected in the database.
```
raddevon, line 602: Question for whoever reviews: Is this necessary? When I was reading this, I noticed that the `family_members` link on `Alice` was empty, and for just a second, it hit me as weird until my brain turned the corner to, "of course it's blank because the database doesn't know anything about reciprocal familial relationships." Thought it would be good to just briefly explain, but maybe it's silly. It feels a little out of place, but also it hits weird for a split second when you see it without a note.

docs/datamodel/links.rst
```diff
@@ -475,11 +475,134 @@ in these cases due to better ergonomics of selecting, updating, and even
...
+              select Person {
+                @character_name := character.0
+              } filter .name = character.1
+            )
+          ))
+        };
+
+    Note that we are also required to wrap the ``actors`` query with
+    :eql:func:`assert_distinct` here to assure the compiler that the result set
+    is distinct.
```
raddevon, line 569: This is duplicated from the cheatsheet, but I think it makes sense because this is such a common mistake. Open to other opinions though. Maybe it should come out of the cheatsheet and live only here instead?

docs/datamodel/links.rst
```diff
@@ -475,11 +475,140 @@ in these cases due to better ergonomics of selecting, updating, and even
...
+
+Updating a link's link property is similar to adding a new one except you can
+select the link on the object instead of selecting from the object type since
+the link has already been established. Here, we've discovered that Alice is
+actually Bob's *step*-sister, so we want to change the link property on the
+already-established link between the two:
+
+.. code-block:: edgeql
+
+    update Person {
```
jaclarke, line 512:
```suggestion
    update Person set {
```

raddevon: 🤦‍♂️

docs/datamodel/links.rst
Dhghomon: I'd say keep it in both locations like you've done, as this is both a bit of a headache to remember as well as an unfortunate compiler message, which means brute force memorization for the time being.

I see that there's a Movie type already on this page though with `characters` instead of `actors` so could change actors to characters to match it.

docs/datamodel/links.rst
```diff
@@ -475,11 +475,141 @@ in these cases due to better ergonomics of selecting, updating, and even
...
+
+    insert Person {
+      name := "Bob",
+      family_members := (
+        select detached Person {
+          @relationship := "sister"
+        }
+        filter .name = "Alice"
+      )
+    }
```
Dhghomon, line 496: Add a semicolon here

docs/datamodel/links.rst
```diff
@@ -475,11 +475,141 @@ in these cases due to better ergonomics of selecting, updating, and even
...
+
```
Dhghomon, line 497: I noticed when playing around with this that link properties cannot be made required in a schema, I wonder where would be a good place to add that note. I was thinking maybe a sample query output here showing the relationship between Bob and Alice, then point out to the reader that the relationship shows up as {} if not indicated as it is not required...and indeed link properties cannot be made required so this schema will not work:

```
type Person {
  name: str;
  multi family_members: Person {
    required relationship: str; # Won't work
  }
}
```

But maybe some other part of the page or via a different introduction could work. Or even just skip all that and add a quick "link properties cannot be made required" here:

"They cannot contain links to other objects, and cannot be `required`."

docs/datamodel/links.rst
```diff
@@ -475,11 +475,141 @@ in these cases due to better ergonomics of selecting, updating, and even
...
+The shape could alternatively be included on an insert if the object being
+linked (the ``Person`` named "Alice" in this example) is being inserted as part
+of the query. If the outer person ("Bob" in the example) already exists and
+only the links need to be added, this can be done in an ``update`` query
+instead of an ``insert`` as shown in the example above.
+
+Updating a link's link property is similar to adding a new one except you can
```
Dhghomon, line 504: This is good to point out and I think could be talked up a bit more to ensure that the reader notices that it's a bit convenient, something like this perhaps:

"Updating a link's property  is similar to adding a new one except that now you don't need to select from the object type: you can select the link on the object itself because the link has already been established."

docs/datamodel/links.rst
```diff
@@ -475,11 +475,141 @@ in these cases due to better ergonomics of selecting, updating, and even
...
+      family_members := (
+        select .family_members {
+          @relationship := "step-sister"
+        }
+        filter .name = "Alice"
+      )
+    }
+
+Using ``select .family_members`` here with the shape including the link
+property allows us to modify the existing link's link property.
```
Dhghomon, line 524: Maybe "the link property of the existing link" or just "the property of the existing link"? I guess by definition a property of an existing link is a link property..

docs/datamodel/links.rst
```diff
@@ -475,11 +475,141 @@ in these cases due to better ergonomics of selecting, updating, and even
...
+Querying link properties
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+To query a link property, add the link property's name prepended with ``@`` to
+a shape on the link.
+
+.. code-block:: edgeql-repl
+
+    edgedb> select Person {
+    .......   name,
```
Dhghomon, line 585: I usually see db> instead of edgedb> and three dots instead of seven for edgeql-repl blocks, could change to that

docs/datamodel/links.rst
Dhghomon: I think it's good to err on the side of extra info as you've done here.

(And actually this could be a good place to mention that link properties cannot be required if you want to add it to this PR)

Dhghomon (approved):
Looks good! I've approved and added a few possible changes/additions if you feel like making them.

docs/datamodel/links.rst
raddevon: I doubled up here to try to disambiguate between a link property and a property on the linked object. I guess I probably went overboard. 😅

docs/datamodel/links.rst
raddevon: I ended up mentioning this as an additional note at the top under the note about "link links" vs. link properties. (I feel like this is kind of a place for general caveats, so it made sense to include both here. We've just told you what they are, but here's what you should know before you get any deeper.) I also reiterated it at the very end in the note about the familial relationship only being reflected in the direction you explicitly added the link to reinforce.

docs/datamodel/links.rst
raddevon: Didn't note this here, but I did add it in two other places as noted in previous comment.

docs/datamodel/links.rst
raddevon: Went with your first suggestion. It's unambiguous without also being ridiculous like my initial attempt. 😆

docs/datamodel/links.rst
raddevon: Good catch!

PR #6247, opened 2023-10-06
Allow configuration of hosted auth UI's OAuth flow

scotttrinh:
We do not yet look at this configuration value, and we should not expose it in the configuration UI until we do, but this will get the value in place in the schema.

### DIFF
--- a/edb/lib/ext/auth.edgeql
+++ b/edb/lib/ext/auth.edgeql
@@ -179,6 +179,8 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
         };
     };
 
+    create scalar type ext::auth::FlowType extending std::enum<PKCE, Implicit>;
+
     create type ext::auth::UIConfig extending cfg::ConfigObject {
         create required property redirect_to: std::str {
             create annotation std::description :=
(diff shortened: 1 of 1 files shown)


1st1 (approved):
lgtm but please ask james/elvis to also give it a look

edb/lib/ext/auth.edgeql
```diff
@@ -167,8 +167,16 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
...
     };
 
+    create scalar type ext::auth::FlowType extending std::enum<PKCE, Implicit>;
+
     create type ext::auth::UIConfig extending cfg::ConfigObject {
         create required property redirect_to: std::str;
+        create required property oauth_flow_type: ext::auth::FlowType {
+            create annotation std::description :=
+                "Which flow does the OAuth flow use when returning an auth \
+                token from an Identity Provider.";
```
jaclarke, line 177: Using the word flow twice sounds awkward, maybe something this?
```suggestion
                "The OAuth flow used when returning an auth token from an \
                identity provider.";
```

jaclarke: Actually do we also need to do the PKCE flow for email/password logins? Don't we still return the token as a query param in that flow?

scotttrinh: > Actually do we also need to do the PKCE flow for email/password logins? Don't we still return the token as a query param in that flow?

Oh, good question, I haven't tested the email+password flow with the hosted UI, but I think you're right. The email+password flow doesn't actually have the PKCE flow implemented in it. Shouldn't be to hard to conditionally add PKCE awareness to the email+password flow! I'll update the name and description to reflect that.

scotttrinh: @jaclarke ⚡ [`790fb3a` (#6247)](https://github.com/edgedb/edgedb/pull/6247/commits/790fb3a531f6f46f02e1e035f7ce7d7057bbe0f9)

For simplicity and clarity, I think I'll make the PKCE flow required for email+password, too, and then when we make this configurable, we can make it optional here again.

PR #6252, opened 2023-10-06
Add PKCE to email+password flow

scotttrinh:
In order to support the hosted UI, we also need the PKCE flow for email+password. This is just as well since it's slightly more secure anyway. That means that developers will need to implement PKCE even if they aren't using an OAuth provider.


fantix:
We should also update the hosted UI to pass on the challenge, right?

scotttrinh:
Oh good call @fantix , I can do that in this PR.

scotttrinh:
@fantix Actually I'll need to do that in #6239 since that's where I get the challenge from the query params. I'll rebase that PR onto this branch and pass the challenge there.

1st1 (approved):

PR #6258, opened 2023-10-07
Fix a protocol bug involving CONFIGURE INSTANCE

msullivan:
CONFIGURE INSTANCE didn't sync the state to a connection but we still
indicated that it did on the pgcon. This could lead to state mismatch
issues.

Also fix another potential vector for such things that I don't think
is actually currently possible (op that changes state but runs no
sql).


fantix (approved):
👍

PR #6269, opened 2023-10-11
Add OAuth providers to sign up hosted auth

scotttrinh:
Add the OAuth provider buttons to the "Sign Up" hosted auth page.

Previously this page was disabled unless you had the password provider configured, but now we conditionally show you all of the providers, including the password provider.

### DIFF
--- a/edb/server/protocol/auth_ext/http.py
+++ b/edb/server/protocol/auth_ext/http.py
@@ -626,19 +626,21 @@ class Router:
 
                 case ('ui', 'signup'):
                     ui_config = self._get_ui_config()
-                    password_provider = (
-                        self._get_password_provider()
-                        if ui_config is not None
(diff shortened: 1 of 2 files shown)


fantix (approved):
Works like a charm!

Q: Is it by design that there is no links for sign-in on the sign-up page, if email-password provider is not configured?

image

scotttrinh:
@fantix 

> Q: Is it by design that there is no links for sign-in on the sign-up page, if email-password provider is not configured?

Good question. There isn't actually any difference between "sign in" and "sign up" for OAuth, so I suppose in that case we could render a completely different UI that just says "Continue with GitHub" maybe? Doesn't seem like a blocker for now, but I can see it being somewhat confusing if you only have OAuth configured.

PR #6271, opened 2023-10-11
Reuse the same PKCE object when reusing challenge

scotttrinh:
The most legitimate use case for this is in the hosted UI, you might go back and forth with a few providers before completing with one. In each case, it'll reuse the same `challenge` string since that is set up only once when doing the initial request to the hosted auth UI.

This doesn't seem to have any ill-effects from a security standpoint since the challenge itself is meant to be intercepted and in cleartext. And in order to use it, you need the verifier, which you will get at the end, but the `PKCEChallenge` wiill have already been deleted by then, so it will not expose the `Identity` that was made with it.


fantix (approved):
Manually verified :+1:

PR #6272, opened 2023-10-11
Avoid using hashed postgres names for triggers

msullivan:
Triggers don't have the 64 character name limit, so don't worry about
it there.

I think this is the only place left in the *schema* where we hash
names to make them short enough, though we sometimes do it in the
queries we generate too. (Though that should be fixable as well.)

### DIFF
--- a/edb/buildmeta.py
+++ b/edb/buildmeta.py
@@ -44,7 +44,7 @@ from edb.common import verutils
 
 
 # Increment this whenever the database layout or stdlib changes.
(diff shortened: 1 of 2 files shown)


msullivan:
Oh, oops. I had rebased onto https://github.com/edgedb/edgedb/pull/6270 to avoid needing to bootstrap, and then I forgot to undo the rebase before force pushing the fixed thing, so this already got merged as part of https://github.com/edgedb/edgedb/pull/6270.

@elprans can you still take a look, though, and then I'll close it

PR #6284, opened 2023-10-12
Doc empty set operations

### LINKED ISSUE #6115: Better surface info about how booleans work
In particular, people are caught off guard that `or` breaks with an empty set. We need to be sure that, pretty much everywhere we talk about filtering, we also note this and that users can use `?=` for values that might be empty.

Can link to [Working with booleans — Cheatsheets | EdgeDB Docs](https://www.edgedb.com/docs/guides/cheatsheet/boolean)

raddevon:
The goal here was to make some sort of note about this in any place the user might like to trip over it. Let me know if there are any good candidates I've missed. I stopped short of noting it on every page that discusses filtering — particularly ``update``, and ``delete`` — because we don't really discuss filtering at all on those pages, the assumption being, I think, that you learn about filtering from the ``select`` documentation. I think this is OK, but I could be convinced otherwise.

Even though the greater than/less than operators are also subject to this, I have decided not to add notes for them because I can't really think of a practical scenario where a user will trip over this with those particular operators. That could be more of a failure of my own imagination than a reflection of reality, so I'm open to dissenting opinions.

Closes #6115

### DIFF
--- a/docs/edgeql/select.rst
+++ b/docs/edgeql/select.rst
@@ -102,19 +102,19 @@ And the following inserts:
 
 .. code-block:: edgeql-repl
 
-  db> insert Hero { 
+  db> insert Hero {
   ...   name := "Spider-Man",
-  ...   secret_identity := "Peter Parker" 
+  ...   secret_identity := "Peter Parker"
   ... };
   {default::Hero {id: 6be1c9c6...}}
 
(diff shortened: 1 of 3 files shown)


scotttrinh (approved):
This all seems clear to me! 🏆

PR #6285, opened 2023-10-12
Fix errors.IdleTransactionTimeoutError error code

msullivan:
This is a regression in 4.0, probably introduced by #5921 moving
interpret_error out of write_error. Some code paths, purely involving
protocol things, can't call interpret_error before calling
write_error, so make write_error at least do static error
interpretation.

Refactor ISE handling while we're at it.


fantix (approved):
:+1:

PR #6286, opened 2023-10-12
Add big scary warning about custom IDs

raddevon:


docs/stdlib/cfg.rst
```diff
@@ -116,15 +116,26 @@ Client connections
...
-     application might not expect.
+      1. An unprivileged user can discover ids that already exist in the
+         database by trying to insert new values and noting when there is a
+         constraint violation on ``.id`` even if the user doesn't have access
+         to the relevant table.
+
+      2. It allows re-using object ids for a different object type, which the
+         application might not expect.
+
+      Additionally, enabling can have serious performance implications as, on
```
msullivan, line 131: I would say "every object type must be checked for collisions" instead. It's kind of a fine distinction, but "scanned" implies worse than it actually is, because my reading of it is that it would actually need to check every object, which it shouldn't: it will do a separate index lookup for each type.

msullivan (approved):

docs/stdlib/cfg.rst
```diff
@@ -116,15 +116,26 @@ Client connections
...
+      an ``insert``, every object will have to be scanned.
+
+      As a result, we don't recommend enabling this. If you need to preserve
```
Dhghomon, line 134: If you want an even scarier message Rust's most famous unsafe function (called `transmute`) has some good wording:

```
The compiler will generate code assuming that you, the programmer, ensure that there will never be undefined behavior. It is therefore your responsibility to guarantee that every value passed to transmute is valid at both types Src and Dst. Failing to uphold this condition may lead to unexpected and unstable compilation results. This makes transmute incredibly unsafe. transmute should be the absolute last resort.
```

"Is is therefore your responsibility" + "failing to uphold this condition may lead to" + "should be the absolute last resort" are pretty nice ways to get the point across that please don't use this.

raddevon: Oh, yikes! Yeah, that's scary alright. We'll keep this in our back pocket, and if the current message doesn't have the desired effect, we'll roll out something more strongly implicating the user in their custom ID related self-destruction. 😅

Issue #5798, opened 2023-07-16
An `unless conflict else` inside a trigger fails with "object type or alias does not exist" error


TheHyphen:
- EdgeDB Version: 3.1+a33a828
- EdgeDB CLI Version: 3.4.0+4d95a2c
- OS Version: MacOS 12.3

Steps to Reproduce:

An `on conflict` with an `else` clause in a trigger makes the `edgedb migrate --dev-mode` fail with the error
> object type or alias 'Stock' does not exist

Schema:
```
module default {
    type Product {
        required name: str;
    }

    type Stock {
        required product: Product {
            constraint exclusive;
        };
        required quantity: int32;
    }

    type IncomingStock {
        required product: Product;
        required quantity: int32;

        trigger StockCreate after insert for each do (
            (
                insert Stock {
                    product := __new__.product,
                    quantity := __new__.quantity
                }
                unless conflict on .product
                else (
                    update Stock set { quantity := .quantity + __new__.quantity }
                )
            )
        );
    }
}
```

Full error message:
```
❯ edgedb migrate --dev-mode
error: object type or alias 'Stock' does not exist
   ┌─ /Users/personal/projects/edgedb/dbschema/default.esdl:17:54
   │
17 │           trigger StockCreate after insert for each do (
   │ ╭──────────────────────────────────────────────────────^
18 │ │             insert Stock {
19 │ │                 product := __new__.product,
20 │ │                 quantity := __new__.quantity
   · │
25 │ │             )
26 │ │         );
   │ ╰─────────^ error

edgedb error: cannot proceed until .esdl files are fixed
```

This error does not occur without the `else` clause in the schema.

mkhlsvn:
This annoying bug still actual not only for 3.4+301ba34 but also for 5.0-dev.7931+b66b7b5

```
error: object type or alias 'DayClosureEvent' does not exist
     ┌─ C:\Users\...\dbschema\default-v1.33.1.esdl:1621:13
     │
1621 │ ╭             FOR d IN range_unpack(range(func::getLocalDate(new.date), func::getCurrentDate())) UNION (
1622 │ │                 INSERT LoanAccruals {date:= d, loan:= <Loan>__new__.id,
1623 │ │                     # error: object type or alias 'DayClosureEvent' does not exist
1624 │ │                     # https://github.com/edgedb/edgedb/issues/5798
[...]

PR #6297, opened 2023-10-13
Fix use of UNLESS CONFLICT in triggers

### LINKED ISSUE #6798: Debuggability for tx serialization errors
Our schema/queries seem to result in a vastly disproportionate number of transaction serialization errors. I understand there are optimizations we can make and in extreme cases file bugs with you to make compiler improvements.

Because our first and foremost priority is the queries actually working in production, so we have an extremely high maximum retries of 10 using the default TS client backoff function. This intentionally obscures many would-be errors. Eg, if something is retried even 5 times, there is already probably something wrong.

We only ever see errors in the cases where it fails 10 times. Even then there is no guarantee that the particular query that failed is the one to optimize and not something that was competing in the background. It is also currently not practical to correlate running queries with the output of the backoff function.

I propose that edgedb support logging the content of queries that are retried, or perhaps retried more than a given number of times. This would let us get ahead of tx serialization problems.

msullivan:
The bug was that normalization didn't process unless_conflict, so we
weren't getting module names in the DDL.

Fixes #6798.

### DIFF
--- a/edb/edgeql/compiler/normalization.py
+++ b/edb/edgeql/compiler/normalization.py
@@ -278,7 +278,7 @@ def normalize_InsertQuery(
         localnames=localnames,
     )
 
-    for field in ('shape',):
+    for field in ('shape', 'unless_conflict',):
         value = getattr(node, field, None)
         _normalize_recursively(
             node,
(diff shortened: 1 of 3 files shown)


aljazerzen (approved):

aljazerzen (approved):

msullivan:
Rather, fixes #5798.

PR #6302, opened 2023-10-16
Build and serialize parser spec during build process

aljazerzen:
Before this PR, parser grammar spec was being retrieved from our Python code where it was read from .pickle files (or rebuild in case of development environment).

This meant that:
- we had to unpickle, serialize to JSON, pass to Rust edgeql-parser, deserialize,
- Python did not free the memory used for spec or its JSON-serialized-form,
- we were basically using 3x as much memory as required.

This PR builds the spec from setup.py and then serializes it into `edb/grammar.bc`. During runtime, this file is read, deserialized and caches. This process can also be triggered by calling `cache_spec()`, so the first call to `parse()` will not need to touch the disk.

Memory usage of the server process dedicated to the parser went from 86.9MB to 17.7MB.
I did not benchmark the performance as it would only optimize first run, but removing serialization of >4MB of JSON have probably helped.

For serialization format I've chosen [BitCode](https://github.com/SoftbearStudios/bitcode), because it has very good serialized size and decent deserialization performance, based on https://github.com/djkoloski/rust_serialization_benchmark. I've made sure to pick something that goes trough serde, which means that this format is easily changeable.


edb/edgeql/parser/__init__.py
```diff
@@ -229,24 +228,15 @@ def _cst_to_ast(
...
-        spec = parsing.load_parser_spec(grammar, allow_rebuild=False)
-    except parsing.ParserSpecIncompatibleError as e:
-        if allow_rebuild:
-            spec = parsing.load_parser_spec(grammar, allow_rebuild=True)
-        else:
-            raise errors.InternalServerError(e.args[0]) from None
-
-    rust_parser.cache_spec(spec)
+def preload() -> None:
+    rust_parser.preload_spec()
```
elprans, line 242: Can we keep the hot reload ability for devmode?  It's a lot better if you only have to reload the server as opposed to running `make rust` every time you've changed the parser.

aljazerzen: It's just `make parsers`, rebuilding Rust is not required.

The problem is that because we don't even init the `parsing.Spec` during runtime anymore, we cannot compare it to find out if it is incompatible.

So we can either rebuild the parsers at every run, or never.

I could port the code from parsing that detects the changes in the spec to the new parser, but it'll take a bit more effort.

aljazerzen:
I'm a bit scared of picking this into 4.x branch - I added changes to the build process, that I'm not confident in. It will probably be fine, but if we really don't want breakage, this could wait for 4.1

elprans:
> I'm a bit scared of picking this into 4.x branch - I added changes to the build process, that I'm not confident in. It will probably be fine, but if we really don't want breakage, this could wait for 4.1

That's what nightlies are for. To relieve the release manager anxiety.

edb/edgeql-parser/edgeql-parser-python/src/parser.rs
```diff
@@ -105,46 +105,66 @@ fn downcast_tokens<'a>(
...
-    let grammar_name = "edb.edgeql.parser.grammar.start";
-    let grammar_mod = py.import(grammar_name)?;
-    let py_spec = load_parser_spec.call(py, (grammar_mod,), None)?;
+    std::fs::write(dst, spec_serialized).ok().unwrap();
+    Ok(PyNone)
+}
 
-    let x = load_spec(py, &py_spec)?;
+fn load_spec() -> parser::Spec {
+    let bytes = std::fs::read("./edb/grammar.bc").ok().unwrap();
```
elprans, line 141: It's a bad idea to be relying on the current working directory to load the spec.  Pass the path as an argument to `parse()` and `get_spec()` instead (or pass the module object and build the path relative to `module.__file__`).

edb/edgeql/parser/__init__.py
elprans: `make parsers` isn't a thing currently, are you going to add it?

edb/edgeql-parser/edgeql-parser-python/src/lib.rs
```diff
@@ -52,12 +52,17 @@ py_module_initializer!(
         m.add(
             py,
             "parse",
-            py_fn!(py, parse(parser_name: &PyString, data: PyObject)),
+            py_fn!(py, parse(spec_filepath: &PyString, parser_name: &PyString, data: PyObject)),
```
msullivan, line 55: It's a little sketchy to me that parse takes a path but will use the cached version even if it's from a different path. Maybe `parse` should take no path and always use a global cache, or maybe the cache should be keyed by the path?

edb/edgeql-parser/edgeql-parser-python/src/parser.rs
```diff
@@ -105,46 +110,69 @@ fn downcast_tokens<'a>(
...
+    Ok(PyNone)
+}
 
-    let x = load_spec(py, &py_spec)?;
+fn load_spec(spec_filepath: &str) -> parser::Spec {
+    let bytes = std::fs::read(spec_filepath).ok().unwrap();
 
-    PARSER_SPECS.set(x).ok();
-    Ok(PARSER_SPECS.get().unwrap())
+    bitcode::deserialize::<parser::SpecSerializable>(&bytes)
```
msullivan, line 151: How big are the savings from storing the serialized spec as bitcode instead of just saving the json?
Mostly just curious.

edb/edgeql-parser/edgeql-parser-python/src/parser.rs
```diff
@@ -105,46 +110,69 @@ fn downcast_tokens<'a>(
...
+        .unwrap()
+        .into()
 }
 
-fn load_spec(py: Python, py_spec: &PyObject) -> PyResult<(parser::Spec, PyObject)> {
+fn spec_from_python(py: Python, py_spec: &PyObject) -> PyResult<parser::SpecSerializable> {
```
msullivan, line 156: Maybe have the python side do the serialization to json?

msullivan (approved):
Looks good.

Some nits, but they can be ignored if you think I'm wrong

edb/edgeql-parser/edgeql-parser-python/src/parser.rs
aljazerzen: bitcode is 47% smaller:
```
[edb/edgeql-parser/edgeql-parser-python/src/parser.rs:160] spec_json.len() = 9473372
[edb/edgeql-parser/edgeql-parser-python/src/parser.rs:144] spec_serialized.len() = 4470568
```

edb/edgeql/parser/grammar/__init__.py
```diff
@@ -6,6 +6,13 @@
 ##
 
 from __future__ import annotations
-
+import pathlib
 
 from . import start as start  # noqa
+
+
+def get_spec_filepath():
```
elprans, line 14:
```suggestion
def get_spec_filepath() -> str:
```

PR #6308, opened 2023-10-18
Add an FTS section to the Standard Library.

vpetrovykh:


docs/stdlib/abstract.rst
```diff
@@ -25,6 +25,18 @@ Abstract types are used to describe polymorphic functions, otherwise known as
 ----------
 
 
+.. eql:type:: std::anyobject
+
+    :index: any anytype object
+
+    An abstract base object type.
```
elprans, line 32: This is wrong.  `anyobject` is not an abstract base, that's what `BaseObject` is.  It is instead a generic type variable like `anytype` but bound to objects.

msullivan: (I think we should maybe make `anyscalar` *not* be a generic base, also, and only be a type variable.)

Also, it's just `anyobject`, not `std::anyobject`

vpetrovykh: I'm not sure how exactly should these `any` pseudo-types interact with hierarchies.

Currently, they are kind of inconsistent:
```
edgedb> select Foo is anytype;
{false}
edgedb> select Foo is anyobject;
{false}
edgedb> select 2 is anytype;
{true}
edgedb> select 2 is anyscalar;
{true}
edgedb> select [2] is (array<anyscalar>);
{true}
edgedb> select [2] is (array<anytype>);
{true}
edgedb> select (1,2) is anytuple;
{false}
edgedb> select (1,2) is (tuple<anyscalar, anyscalar>);
{true}
edgedb> select (1,2) is (tuple<anytype, anytype>);
{true}
```

I think, it makes sense for `any` pseudotypes to be valid in `is` queries, for example. For scalars we have a whole bunch of `any` abstract types that establish the scalar hierarchies. I'm not sure why `anyscalar` is really more special than `anyobject`, I think they ought to express a similar level of abstraction and it makes sense for both of them to be part of the type system. The `any` types have a special feature that they "preserve" the actual argument types when used in function signatures. This feature is more obvious for `anyobject`, `anytype` and `anytuple` as they preserve internal fields, but for scalars it only comes up when chaining functions and the return type must match the parameter type of the next function.

msullivan: I don't think any of those queries should be allowed, honestly

msullivan: But all of this is neither here nor there: `anyobject` should match the copy for `anytype` and `anytuples`

aljazerzen: I agree, those queries are not the intended used of `anytype` and `anyobject` and should not be allowed.

Correct me if I'm wrong, `anytype` and `anyobject` are used for a basic implementation of generic functions. In Java syntax, using `anytype` in a function declaration, would be similar to:

```java
<T extends Type> T function_name(T arg1) {
    ...
}
```

And anyobject would be:
```java
<O extends ObjectType> O function_name(O arg1) {
    ...
}
```

If they are indeed just basic generic arguments, referring to them outside of the context of a function does not make sense.

(I'm assuming `anyscalar` or `anytuple` work similarly)

---

Looking at reflection, the type hierarchy is:

```
schema::Type                <- anytype
  schema::PrimitiveType
    schema::ScalarType      <- anyscalar
    schema::CollectionType
      schema::Array
      schema::Tuple         <- anytuple
      schema::Range
      schema::MultiRange
  schema::ObjectType        <- anyobject
```

docs/stdlib/fts.rst
```diff
@@ -0,0 +1,575 @@
...
+to full-text search. This is an unusual index as it actually affects the
+results of :eql:func:`fts::search` function. This is unlike most indexes which
+only affect the performance and not the actual results. Another special
+feature of ``fts::index`` is that at most one such index can be declared per
+type. If a type inherits this index from a parent and also declares its own,
+only the new index applies and fully overrides the ``fts::index`` inherited
+from the parent type. This means that when dealing with a hierarchy of
+full-text-searchable types, each type can customize what gets searched as
+needed.
+
```
aljazerzen, line 61: Should we also mention what happens when two parents both define their own fts::index?

The error message is clear enough, so it might not be needed.

https://github.com/edgedb/edgedb/blob/master/edb/schema/indexes.py#L1222-L1230

docs/stdlib/fts.rst
```diff
@@ -0,0 +1,575 @@
...
+    * ``gle``
+    * ``ita``
+    * ``nor``
+    * ``por``
+    * ``ron``
+    * ``rus``
+    * ``spa``
+    * ``swe``
+    * ``tur``
+
```
aljazerzen, line 333: Maybe link to the list of language codes: https://iso639-3.sil.org/code_tables/639/data?field_iso639_cd_st_mmbrshp_639_1_tid=94671

docs/stdlib/fts.rst
```diff
@@ -0,0 +1,575 @@
...
+      - Opaque document type
+
+    * - :eql:func:`fts::search`
+      - :eql:func-desc:`fts::search`
+
+    * - :eql:func:`fts::with_options`
+      - :eql:func-desc:`fts::with_options`
+
+When considering FTS functionality our goal was to come up with an interface
+that could support different backend FTS providers. To achieve that we've
```
aljazerzen, line 40: We should mention somewhere that at the moment, EdgeDB supports only postgres FTS backend

aljazerzen (approved):

PR #6310, opened 2023-10-18
Followup for parser grammar serialization

aljazerzen:
I forgot to push a few last refactoring commits in #6302, so here they come.

### DIFF
--- a/edb/edgeql-parser/edgeql-parser-python/src/parser.rs
+++ b/edb/edgeql-parser/edgeql-parser-python/src/parser.rs
@@ -1,9 +1,9 @@
-use once_cell::sync::OnceCell;
-
+use cpython::exc::AssertionError;
 use cpython::{
     ObjectProtocol, PyClone, PyInt, PyList, PyNone, PyObject, PyResult, PyString, PyTuple, Python,
     PythonObject, PythonObjectWithCheckedDowncast, ToPyObject,
 };
+use once_cell::sync::OnceCell;
 
 use edgeql_parser::parser;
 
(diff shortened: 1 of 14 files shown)


edb/edgeql-parser/edgeql-parser-python/src/parser.rs
```diff
@@ -110,60 +105,49 @@ fn downcast_tokens<'a>(
...
-
-    let res = spec_to_json.call(py, (py_spec,), None)?;
-
-    let spec_json = PyString::downcast_from(py, res)?;
-    let spec_json = spec_json.to_string(py).unwrap();
-
-    Ok(serde_json::from_str::<parser::SpecSerializable>(&spec_json).unwrap())
-}
-
 fn load_productions(py: Python<'_>, spec: &parser::Spec) -> PyResult<PyObject> {
```
msullivan, line 152: This isn't a blocker, but maybe we could make `preload_spec` return the list of production/classname tuples, and call `load_spec_productions` from python when we call `preload_spec` and cache that side of it in python?

Then there wouldn't be any more tight coupling between the python side and the rust side.

msullivan (approved):

edb/edgeql-parser/edgeql-parser-python/src/parser.rs
aljazerzen: That a bit of work for marginal benefit - rust side is already quite decoupled, since only `edgeql-parser-python` crate depends on Python.

PR #6311, opened 2023-10-18
Pin the CLI

msullivan:
Pending https://github.com/edgedb/edgedb-cli/issues/1152


elprans:
I posted fixes for https://github.com/edgedb/edgedb-cli/issues/1152, so this is probably unnecessary.

PR #6315, opened 2023-10-19
Clean up some volatility_ref related code

msullivan:
Now that #6313 made newrel always clear volatility_ref, we can get rid
of a bunch of code that was doing that explicitly.


elprans (approved):

msullivan:
I marked it for backport to 4.x because I think it's low risk and it might save us from a few merge conflicts in the future, but feel free to decide not to backport if you don't think it should go in, @aljazerzen

PR #6317, opened 2023-10-19
Store std schema and reflection schema in the same pickle

elprans:
The reflection schema is a thin modification of the std schema so them
sharing the same serialization container basically halves the memory
consumption of cached schema (saving about 20MiB).

### DIFF
--- a/edb/buildmeta.py
+++ b/edb/buildmeta.py
@@ -44,7 +44,7 @@ from edb.common import verutils
 
 
 # Increment this whenever the database layout or stdlib changes.
-EDGEDB_CATALOG_VERSION = 2023_10_11_00_00
+EDGEDB_CATALOG_VERSION = 2023_10_19_00_00
 EDGEDB_MAJOR_VERSION = 5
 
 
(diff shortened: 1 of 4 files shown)


msullivan (approved):
Great!

Do you know if the data path that gets stdschema/reflschema to the compiler pool preserves this?

Do we want to backport this to 4.x? I think the approach would be to put the combined schema in a new field in the database, and fall back to loading from the old ones if it isn't present.

elprans:
> Do you know if the data path that gets stdschema/reflschema to the compiler pool preserves this?

It does, RPC args are pickled wholesale.

> Do we want to backport this to 4.x? I think the approach would be to put the combined schema in a new field in the database, and fall back to loading from the old ones if it isn't present.

Yeah, sure!  Perhaps I should rename the field here then.

msullivan:
Should I do the backport, or do you want to?

msullivan:
As an aside: this preserves sharing between the keys and values, but doesn't preserve the HAMTs internal sharing of nodes. Probably that's much smaller, though. I don't know if there's an easy way to manage that just using pickle's mechanisms, since the internal nodes aren't pyobjects

elprans:
> As an aside: this preserves sharing between the keys and values, but doesn't preserve the HAMTs internal sharing of nodes. Probably that's much smaller, though. I don't know if there's an easy way to manage that just using pickle's mechanisms, since the internal nodes aren't pyobjects

If memray is to be believed that overhead is minimal.

PR #6324, opened 2023-10-20
Add ext::auth email verification

scotttrinh:
Send a verification email when registering for a new `LocalIdentity` when using the `EmailPasswordFactor`. If you have `require_verification := true` (which is the default) this will also block logging in until you verify.

One interesting side-effect here is that we delay creating the PKCE flow until you verify so that the PKCE token can still maintain a short validity window. What's not obvious about this change is that it means it's a bit more likely that the user ends up accidentally clobbering their PKCE flow session if they haven't been careful to reuse the same verifier/challenge combination. You can always initiate an log in flow, and it'll just work since there is no elapsed time between starting your PKCE flow and completing it.


msullivan:
This might not be worth the trouble at the moment, but I'd recommend opting your code into stricter mypy checks in `pyproject.toml`

edb/server/protocol/auth_ext/email.py
```diff
@@ -0,0 +1,107 @@
...
+        db,
+        msg,
+        sender=from_addr,
+        recipients=to_addr,
+        test_mode=test_mode,
+    )
+    await _protected_send(coro, tenant)
+
+
+async def _protected_send(
```
msullivan, line 99: I don't 100% understand what's going on here. The idea is to execute `coro` but take a random amount of time before returning?
(What's the timing attack we're worried about here?)

The reason I came here to comment was to say that the create_task usage is dangerous because it doesn't save a reference, but then I noticed it was `tenant.create_task`, so clearly fine.

scotttrinh: I took this code, pretty much verbatim from: https://github.com/edgedb/edgedb/pull/6211/files#diff-ceb3a40c7d3d1fc8a40b20a415f88b76616edc12d2d37ddd46c99b5e94cff9a2R375-R395

If there is a better way to abstract this, I'm definitely open for suggestion, but I'm not sure what the original concern was w.r.t. timing attacks other than making the request take slightly longer to make it a bit more expensive to make the request.

@fantix any thoughts about @msullivan 's question here?

scotttrinh:
@msullivan 

> This might not be worth the trouble at the moment, but I'd recommend opting your code into stricter mypy checks in pyproject.toml

I actually just turned them off in my editor. At the moment, due to how the body of `handle_request` is just a big `match` statement, the LSP/typechecking was just completely broken (mostly complaining about every variable being unused). I plan on doing a big refactor to split out the routing and handling, but even before we got to this point, the amount of `Unknown` happening due to the cython <-> python translation made almost the entire file show as errors which was obscuring actual errors/typos.

I'm definitely 100% pro-types, and I know this draft PR is currently failing CI—I'm still working on manually testing the flow. Any advice or help getting better types for things like `request`, `db`, config stuff, etc. is absolutely welcome!

edb/server/protocol/auth_ext/email.py
fantix: I think Yury's idea was to mask the cases where e.g. the email address doesn't exist, or it fails fast to send the email, so that the attackers cannot guess what's happening on the server side from how long it runs, thus protecting the email addresses that do exist from spam or phishing (like, a potential attacker could run a test with a public email list against the server without this sleep, get a list of registered email addresses, and phish saying something as if it's actually from this server). But after reviewing the code, I think the sleep should be done earlier, and `NoIdentityFound` must not be raised in such APIs.

scotttrinh: @fantix thanks for the reminder on not leaking whether an email is registered!

Looks like we potentially raising `NoIdentityFound` in two places:
1. `resend-verification-email`: if we couldn't find the `email`.
2. `send_reset_email`: if we couldn't find the `email`.
3. `verify`: if we couldn't find the `email` referenced in the verification token.

I think we can and should trust the `email` within the token, so it still seems appropriate to send a useful message here. It _should_ be extremely unlikely to get into this state normally, but if you delete the `LocalIdentity` and `EmailPasswordFactor` for a user who has not verified their email, they would hit this case.

---

What do you think of the `send_fake_email` approach [here](https://github.com/edgedb/edgedb/pull/6324/commits/5e40b54fb771bfba4c5022b5d45bf201c2aeb173) rather than moving the sleep?

fantix: Yeah, that looks good to me 👍

fantix (approved):
Tested and worked! 🚀

PR #6325, opened 2023-10-20
Update bundled Postgres to 16

aljazerzen:
Closes #6308

There is a few breakages with explain plans. @msullivan can you immediately spot what's going on before I dive too deep?

### DIFF
--- a/.github/workflows.src/tests-pg-versions.tpl.yml
+++ b/.github/workflows.src/tests-pg-versions.tpl.yml
@@ -35,6 +35,9 @@ jobs:
          - 'CREATEDB NOCREATEROLE'
         multi-tenant-mode: [ '' ]
         include:
+          - postgres-version: 14
+            single-mode: ''
+            multi-tenant-mode: ''
           - postgres-version: 15
             single-mode: ''
             multi-tenant-mode: ''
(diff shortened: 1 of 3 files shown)


msullivan:
Looking quickly at `test_edgeql_explain_ranges_strictly_below_02`, it seems that one of the nodes changed from
```
{'important': True, 'title': 'scan_direction', 'type': 'text', 'value': 'NoMovement'}
```
to
```
{'title': 'scan_direction', 'value': 'Forward', 'type': 'text', 'important': True}
```
All of the rest seem to be similar.
Seems unlikely we care about the detail here. We can just replace `NoMovement` with str?

.github/workflows.src/tests-pg-versions.tpl.yml
```diff
@@ -28,20 +28,20 @@ jobs:
...
-        postgres-version: [ 14 ]
+        postgres-version: [ 16 ]
         single-mode:
          - ''
          - 'NOCREATEDB NOCREATEROLE'
          - 'CREATEDB NOCREATEROLE'
         multi-tenant-mode: [ '' ]
         include:
-          - postgres-version: 14
+          - postgres-version: 15
```
msullivan, line 38: Let's keep 14 in the mix also

msullivan (approved):

PR #6328, opened 2023-10-20
Add a special param token for normalization

aljazerzen:
When working on the parser, I had a test case:
```
select count(foo 1);
```
... which should report a missing comma.
But when issuing that to the server, I got an error saying `InvalidReferenceError: function 'default::foo' does not exist`. That's because we do query "normalization", where the literals are extracted and replaced with params:

```
select count(foo (<int>$1);
```

... which is syntactically valid EdgeQL

... which is not good.

So this PR creates a new special token without textual representation that's produced during normalization, so literals don't expand to something that messes up parsing.

### DIFF
--- a/edb/edgeql/parser/grammar/tokens.py
+++ b/edb/edgeql/parser/grammar/tokens.py
@@ -164,8 +164,13 @@ class T_AT(Token, lextoken='@'):
     pass
 
 
-class T_ARGUMENT(Token):
+class T_PARAMETER(Token):
     pass
+
+
+class T_PARAMETERANDTYPE(Token):
+    pass
+    "A special token produced by normalization"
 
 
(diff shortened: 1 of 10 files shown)


edb/edgeql/parser/grammar/tokens.py
```diff
@@ -164,10 +164,15 @@ class T_AT(Token, lextoken='@'):
     pass
 
 
-class T_ARGUMENT(Token):
+class T_PARAMETER(Token):
     pass
 
 
+class T_PARAMETERANDTYPE(Token):
+    pass
```
msullivan, line 172: Either drop `pass` or move it after docstring

msullivan (approved):

msullivan:
Probably not worth trying to backport this to 3.x, since so much changed about the parser?

PR #6330, opened 2023-10-20
Fix GA dryrun workflow 

aljazerzen:


msullivan (approved):

elprans (changes requested):
This file is generated, you want to fix the template and run `make` in `.github`.

aljazerzen:
I've refactored whitespace in the generated yaml, see separate commits.

elprans (approved):

PR #5751, opened 2023-07-07
Update index.rst

hitrust:
Fix typo

PR #5750, opened 2023-07-07
Update connection.rst

hitrust:
EDGEDB_USER, EDGEDB_DATABASE and EDGEDB_PASSWORD are deprecated using EDGEDB_SERVER_USER, EDGEDB_SERVER_DATABASE and EDGEDB_SERVER_PASSWORD instead respectively.

PR #6331, opened 2023-10-20
Fix some docs typos

msullivan:
This is a rollup of of #5750, #5751, #5752, and #5756.

### DIFF
--- a/docs/reference/connection.rst
+++ b/docs/reference/connection.rst
@@ -248,11 +248,11 @@ instance-level configuration object.
 
   * - **Environment variable**
     - **CLI flag**
-  * - ``EDGEDB_DATABASE``
+  * - ``EDGEDB_SERVER_DATABASE``
     - ``--database/-d <name>``
-  * - ``EDGEDB_USER``
+  * - ``EDGEDB_SERVER_USER``
     - ``--user/-u <user>``
-  * - ``EDGEDB_PASSWORD``
+  * - ``EDGEDB_SERVER_PASSWORD``
     - ``--password <pass>``
   * - ``EDGEDB_TLS_CA_FILE``
     - ``--tls-ca-file <path>``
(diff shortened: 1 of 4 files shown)


docs/reference/connection.rst
```diff
@@ -248,11 +248,11 @@ instance-level configuration object.
...
   * - **Environment variable**
     - **CLI flag**
-  * - ``EDGEDB_DATABASE``
+  * - ``EDGEDB_SERVER_DATABASE``
     - ``--database/-d <name>``
-  * - ``EDGEDB_USER``
+  * - ``EDGEDB_SERVER_USER``
     - ``--user/-u <user>``
-  * - ``EDGEDB_PASSWORD``
+  * - ``EDGEDB_SERVER_PASSWORD``
```
raddevon, line 255: I believe these are actually different variables. I think the originals were correct in this context. The `SERVER` variables specify the default database, user, and password that will be created when the server is brought up, but in the context of client connection, the user is specifying what they want to connect to, which doesn't use the `SERVER` variables.

Hey, @fantix, can you check my work on that? Does it sound right to you, or are these changes correct?

aljazerzen: I can confirm that this is the case, @raddevon.

These changes are wrong.

aljazerzen (changes requested):
The env vars commit should be reverted.

raddevon:
@hitrust If you can revisit this and revert the environment variable changes, I will get the other changes merged. Thank you for finding these issues and taking the time to write a fix!

hitrust:
@raddevon @aljazerzen 

The reason why I make these changes is that when I use EDGEDB_USER, EDGEDB_DATABASE and EDGEDB_PASSWORD in my docker-compose.yml file, there is a warning record in edgedb container logs. the record says EDGEDB_USER, EDGEDB_DATABASE and EDGEDB_PASSWORD are deprecated.

However when using EDGEDB_SERVER_USER, EDGEDB_SERVER_DATABASE and EDGEDB_SERVER_PASSWORD instead respectively, there is no warning record in logs.

It's been a while. maybe things changed.

raddevon:
Yes, Docker is looking for the server variables, so it doesn't want the client connection variables. The documentation changed here though is about the client connection variables, which were correct. The Docker/server variables are documented in two places:

- https://www.edgedb.com/docs/reference/environment#server-variables
- https://www.edgedb.com/docs/guides/deployment/docker#configuration

hitrust:
@raddevon 

Thanks for your explanation. 

It seems that I confused the client connection parameters​ and docker environment  variables (server variables).

Dhghomon:
> @raddevon
> 
> Thanks for your explanation.
> 
> It seems that I confused the client connection parameters​ and docker environment variables (server variables).
> 
> Sorry for my fault.

No problem! That part has been reverted so should be fine now.

Dhghomon (approved):

PR #6333, opened 2023-10-20
docs: Fix the docs in several places

elliotwaite:
Here are some suggestions I thought of while reading over the docs.

I'll add some inline comments to the changed files to add some extra info to some of the changes.

Also, while reading over the docs, there were a couple of places where I wasn't sure what part of the syntax was for. For example, in the [Inserting and updating link properties](https://www.edgedb.com/docs/datamodel/links#inserting-and-updating-link-properties) code example, the `detached` keyword was used for the first time and I wasn't sure what it did. Perhaps a short description could be added in the surrounding text with a link to where the reader could find a longer description. Also, in the [What can be parameterized?](https://www.edgedb.com/docs/edgeql/parameters#what-can-be-parameterized) code​ example, the `then` keyword is used to specify multiple `order by` expressions. I hadn't seen the `then` keyword used in that context before so at first I wasn't sure what it did. So maybe a short explanation could be added to the surrounding text that just said that the `then` keyword is used to specify multiple `order by` expressions. I was less confident about writing the text for these changes, so I didn't include them in this PR, but I thought I'd just mention them here in case someone else with better writing ability wanted to add them.

There were a couple of other things I noticed while reading over the docs that I wanted to ask about. These are things that seemed a bit odd to me, but they also seemed like there was probably a good reason behind them:

If the `union` keyword is required for `for` loops:
```
for number in {0, 1, 2, 3}
union (
  select { number, number + 0.5 }
);
```
...why not place it on the same line:
```
for number in {0, 1, 2, 3} union (
  select { number, number + 0.5 }
);
```

Also, why does the [System Functions](https://www.edgedb.com/docs/stdlib/sys) page list the `sys::get_transaction_isolation()` function in the lower part of the page but not in the top section that lists all the other functions?

P.S. Thank you for building EdgeDB! I've been enjoying learning about it and am excited to start using it.

### DIFF
--- a/docs/datamodel/constraints.rst
+++ b/docs/datamodel/constraints.rst
@@ -260,7 +260,7 @@ Constraints on links
 --------------------
 
 When defining a constraint on a link, ``__subject__`` refers to the *link
-itself*. This is commonly used add constraints to :ref:`link properties
+itself*. This is commonly used to add constraints to :ref:`link properties
 <ref_datamodel_link_properties>`.
 
 .. code-block:: sdl
@@ -281,7 +281,7 @@ itself*. This is commonly used add constraints to :ref:`link properties
     type User {
       name: str;
       multi friends: User {
-        single strength: float64;
+        strength: float64;
         constraint expression on (
           __subject__@strength >= 0
         );
--- a/docs/datamodel/inheritance.rst
+++ b/docs/datamodel/inheritance.rst
@@ -238,7 +238,7 @@ Use ``abstract`` to declare reusable, user-defined constraint types.
     abstract constraint in_range(min: anyreal, max: anyreal) {
       errmessage :=
         'Value must be in range [{min}, {max}].';
-      using (max > __subject__ and __subject__ >= min);
+      using (min <= __subject__ and __subject__ < max);
     }
 
     type Player {
@@ -252,7 +252,7 @@ Use ``abstract`` to declare reusable, user-defined constraint types.
     abstract constraint in_range(min: anyreal, max: anyreal) {
       errmessage :=
         'Value must be in range [{min}, {max}].';
-      using (max > __subject__ and __subject__ >= min);
+      using (min <= __subject__ and __subject__ < max);
     }
 
     type Player {
--- a/docs/datamodel/introspection/objects.rst
+++ b/docs/datamodel/introspection/objects.rst
@@ -47,6 +47,7 @@ Introspection of the ``schema::ObjectType``:
 Consider the following schema:
 
 .. code-block:: sdl
+    :version-lt: 3.0
 
     abstract type Addressable {
         property address -> str;
@@ -57,6 +58,22 @@ Consider the following schema:
         required property name -> str;
 
         multi link friends -> User;
+
+        # define an index for User based on name
+        index on (.name);
+    }
+
+.. code-block:: sdl
+
+    abstract type Addressable {
+        address: str;
+    }
+
+    type User extending Addressable {
+        # define some properties and a link
+        required name: str;
+
+        multi friends: User;
 
         # define an index for User based on name
         index on (.name);
--- a/docs/datamodel/links.rst
+++ b/docs/datamodel/links.rst
@@ -67,8 +67,9 @@ Required links
 
 All links are either ``optional`` or ``required``; the default is ``optional``.
 Use the ``required`` keyword to declare a required link. A required link must
-point to *at least one* target instance. In this scenario, every ``Person``
-must have a ``best_friend``:
+point to *at least one* target instance, and if the cardinality of the required
+link is ``single``, it must point to *exactly one* target instance. In this
+scenario, every ``Person`` must have *exactly one* ``best_friend``:
 
 .. code-block:: sdl
     :version-lt: 3.0
@@ -143,7 +144,7 @@ instances can link to the same target(s).
 
 In the ``GroupChat`` example, the ``GroupChat.members`` link is now
 ``exclusive``. Two ``GroupChat`` objects cannot link to the same ``Person``;
-put differently, no ``Person`` can be a ``member`` of multiple ``GroupChat``.
+put differently, no ``Person`` can be a ``member`` of multiple ``GroupChat``s.
 
 .. _ref_guide_modeling_relations:
 
@@ -340,10 +341,10 @@ A user can like multiple movies. And in the absence of an ``exclusive``
 constraint, each movie can be liked by multiple users. Thus this is a
 *many-to-many* relationship.
 
-Filtering, limiting, and ordering links
+Filtering, ordering, and limiting links
 ---------------------------------------
 
-The clauses ``order by``, ``filter`` and ``limit`` can be used on links
+The clauses ``filter``, ``order by`` and ``limit`` can be used on links
 as well.
 
 If no properties of a link are selected, you can put the relevant clauses
@@ -472,8 +473,8 @@ properties of the relationships and properties of the target objects. On the
 other hand, for one-to-one, one-to-many, and many-to-one relationships there's
 an exact correspondence between the link and one of the objects being linked.
 In these situations any property of the relationship can be equally expressed
-as the property of the target object (for one-to-many and one-to-one cases) or
-as the property of the source object (for many-to-one and one-to-one cases).
+as the property of the source object (for one-to-many and one-to-one cases) or
+as the property of the target object (for many-to-one and one-to-one cases).
 It is generally advisable to use object properties instead of link properties
 in these cases due to better ergonomics of selecting, updating, and even
 casting into :eql:type:`json` when keeping all data in the same place rather
--- a/docs/datamodel/links.rst
+++ b/docs/datamodel/links.rst
@@ -144,7 +144,8 @@ instances can link to the same target(s).
 
 In the ``GroupChat`` example, the ``GroupChat.members`` link is now
 ``exclusive``. Two ``GroupChat`` objects cannot link to the same ``Person``;
-put differently, no ``Person`` can be a ``member`` of multiple ``GroupChat``s.
+put differently, no ``Person`` can be a ``member`` of multiple ``GroupChat``
+objects.
 
 .. _ref_guide_modeling_relations:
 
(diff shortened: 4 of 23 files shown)


docs/datamodel/links.rst
```diff
@@ -143,7 +144,7 @@ instances can link to the same target(s).
 
 In the ``GroupChat`` example, the ``GroupChat.members`` link is now
 ``exclusive``. Two ``GroupChat`` objects cannot link to the same ``Person``;
-put differently, no ``Person`` can be a ``member`` of multiple ``GroupChat``.
+put differently, no ``Person`` can be a ``member`` of multiple ``GroupChat``s.
```
elliotwaite, line 147: I wasn’t sure the best way to make `GroupChat` plural. Could also be changed to "`GroupChat` objects."

docs/datamodel/links.rst
```diff
@@ -340,10 +341,10 @@ A user can like multiple movies. And in the absence of an ``exclusive``
 constraint, each movie can be liked by multiple users. Thus this is a
 *many-to-many* relationship.
 
-Filtering, limiting, and ordering links
+Filtering, ordering, and limiting links
```
elliotwaite, line 344: I thought it might be better to list "filtering, ordering, and limiting" in a consistent order that reflects the order in which they are used in the select syntax.

docs/datamodel/links.rst
```diff
@@ -472,8 +473,8 @@ properties of the relationships and properties of the target objects. On the
 other hand, for one-to-one, one-to-many, and many-to-one relationships there's
 an exact correspondence between the link and one of the objects being linked.
 In these situations any property of the relationship can be equally expressed
-as the property of the target object (for one-to-many and one-to-one cases) or
-as the property of the source object (for many-to-one and one-to-one cases).
+as the property of the source object (for one-to-many and one-to-one cases) or
+as the property of the target object (for many-to-one and one-to-one cases).
```
elliotwaite, line 477: I though these should be switched, unless I am misunderstanding something.

docs/datamodel/objects.rst
```diff
@@ -8,9 +8,11 @@ Object Types
...
 <ref_datamodel_props>` and :ref:`links <ref_datamodel_links>`.
 
-Properties are used to attach primitive data to an object type. They are
-declared with the ``property`` keyword. For the full documentation on
-properties, see :ref:`Properties <ref_datamodel_props>`.
+Properties are used to attach primitive data to an object type. Prior to
+EdgeDB 3.0, they must be declared with the ``property`` keyword, but the
+keyword is not required in 3.0+ unless the property is a :ref:`computed
+property <ref_datamodel_computed>`. For the full documentation on properties,
+see :ref:`Properties <ref_datamodel_props>`.
```
elliotwaite, line 15: I thought it might be good to remind the reader why the `property` and `link` keywords are used here, since when I first read this section, I thought they were typos left over from the pre-3.0 syntax. However, let me know if you have any suggestions for how this could be worded better, or if you don't think this extra info is needed.

docs/datamodel/constraints.rst
```diff
@@ -281,7 +281,7 @@ itself*. This is commonly used add constraints to :ref:`link properties
     type User {
       name: str;
       multi friends: User {
-        single strength: float64;
+        strength: float64;
```
elliotwaite, line 284: It didn’t seem like `single` was needed here. But please correct me if I am wrong about this.

docs/datamodel/inheritance.rst
```diff
@@ -238,7 +238,7 @@ Use ``abstract`` to declare reusable, user-defined constraint types.
     abstract constraint in_range(min: anyreal, max: anyreal) {
       errmessage :=
         'Value must be in range [{min}, {max}].';
-      using (max > __subject__ and __subject__ >= min);
+      using (min <= __subject__ and __subject__ < max);
```
elliotwaite, line 241: This is subjective, but I thought this ordering was easier to read.

docs/stdlib/set.rst
```diff
@@ -407,8 +407,8 @@ Sets
 
         select User.<owner[is Issue];
 
-        # With the use of type intersection it's possible to refer to
-        # specific property of Issue now:
+        # With the use of type intersection it's possible to refer
+        # to a specific property of Issue now:
```
elliotwaite, line 411: I added the missing “a” but also changed the word wrap location so that the comment would be fully visible without needing to horizontally scroll the code example when viewing the page in a medium-width browser.

docs/stdlib/type.rst
```diff
@@ -231,10 +231,11 @@ Types
         with module example
         select Text {
             name :=
-                Text[is Issue].name IF Text is Issue ELSE
+                Text[is Issue].name if Text is Issue else
                 <str>{},
                 # the cast to str is necessary here, because
-                # the type of the computed expression must be defined
+                # the type of the computed expression must be
+                # defined
```
elliotwaite, line 238: I changed the word wrap location here so that the comment would be fully visible without needing to horizontally scroll the code example when viewing the page in a medium-width browser.

docs/stdlib/string.rst
```diff
@@ -259,9 +259,9 @@ Strings
 
     Case-sensitive simple string matching.
 
-    Returns ``true`` if the *value* ``V`` matches the *pattern* ``P``
-    and ``false`` otherwise.  The operator ``not like`` is
-    the negation of ``like``.
+    Returns ``true`` if the *value* (the ``str`` on the left) matches the
+    *pattern* (the ``str`` on the right) and ``false`` otherwise. The operator
+    ``not like`` is the negation of ``like``.
```
elliotwaite, line 264: I wasn’t sure what the `V` and `P` were referring to, so I thought something like this might be more clear. But let me know if you think something else would be better, or if you prefer how it originally was.

docs/edgeql/delete.rst
```diff
@@ -13,7 +13,7 @@ The ``delete`` command is used to delete objects from the database.
 Clauses
 -------
 
-Deletion statements support ``filter``, ``order by``, and ``offset/limit``
+Deletion statements support ``filter``, ``order by``, ``offset``, and ``limit``
```
elliotwaite, line 16: When I first saw this as `offset/limit`, I thought that it implied that those clauses couldn’t be used separately. But since they can, I thought it would be better to list `offset` and `limit` as separate items.

Dhghomon:
Thanks for the PR! I just noticed this and will give it a look over.

docs/datamodel/constraints.rst
Dhghomon: Agreed, don't see any need for `single` here.

docs/datamodel/inheritance.rst
Dhghomon: True!

docs/datamodel/introspection/objects.rst
```diff
@@ -62,6 +63,22 @@ Consider the following schema:
...
+    type User extending Addressable {
+        # define some properties and a link
+        required name: str;
+
+        multi friends: User;
+
+        # define an index for User based on name
+        index on (.name);
+    }
+
```
Dhghomon, line 81: Good catch. I see the same old syntax in introspection/indexes.rst as well.

docs/datamodel/links.rst
Dhghomon: I think I remember @raddevon mentioning that s for plural right after double backticks often upsets the formatter. So ``GroupChat`` objects is probably the safer option.

docs/datamodel/links.rst
Dhghomon: I think so...@raddevon can you confirm? Feels like maybe a chart could be useful here as I feel a bit dizzy reading that paragraph.

docs/datamodel/objects.rst
Dhghomon: Looks good and the extra reminder makes sense. I'd make a small change to this because as is it feels like "must" should be in the past tense:

"In versions prior to EdgeDB 3.0 they must be declared with the..."

docs/edgeql/delete.rst
Dhghomon: Good point.

docs/stdlib/string.rst
Dhghomon: Looks good to me.

Dhghomon:
Looks good to me with just a tiny suggestion or two, I'll just see what @raddevon thinks about the link properties part and maybe @scotttrinh will be interested in giving the PR a look too. They are involved with something fairly large at the moment so they might be out of action for a day or two. Thanks again!

docs/datamodel/links.rst
elliotwaite: Sounds good. I've updated it to "`GroupChat` objects".

docs/datamodel/objects.rst
elliotwaite: I agree, that would read better. I've updated it to use your suggestion.

docs/datamodel/links.rst
raddevon: This change looks correct to me!

Dhghomon (approved):
Looks good! Devon confirmed the bit about the link properties so nothing holding this one back.

docs/edgeql/path_resolution.rst
```diff
@@ -185,5 +185,6 @@ outer query. This is because clauses like :ref:`filter
 <ref_eql_select_filter>` and :ref:`order by <ref_eql_select_order>` need to
 be applied to each value in the result.
 
-The :ref:`limit <ref_eql_select_pagination>` clause is not nested in the scope
-because it needs to be applied globally to the entire result set of your query.
+The :ref:`offset <ref_eql_select_pagination>` and
+:ref:`limit <ref_eql_select_pagination>` clauses are not nested in the scope
+because they need to be applied globally to the entire result set of your query.
```
Dhghomon, line 190: Looks like this is the only line holding up a merge (longer than 79 characters).

elliotwaite: Ah, missed this one, my mistake. I just pushed an update to fix this.

PR #6335, opened 2023-10-21
Experiment with UNION-less FOR

msullivan:
Elvis suggested we try dropping the UNION from FOR. When I tried
dropping it and requiring just a ParenExpr, I was sucked into a rabbit
hole of apparent bugs in our parser that I haven't followed up on yet.

But it *does* work with braces... Thoughts?

### DIFF
--- a/edb/edgeql-parser/src/parser/custom_errors.rs
+++ b/edb/edgeql-parser/src/parser/custom_errors.rs
@@ -219,7 +219,20 @@ impl<'s> Parser<'s> {
                 && self.compare_stack(
                     &[
                         Cond::keyword("for"),
-                        Cond::Production("OptionalOptional"),
+                        Cond::Production("Identifier"),
+                        Cond::keyword("in"),
+                    ],
+                    i,
+                    ctx,
+                )
+            {
+                return Some((i + 3, ParserRule::ForIterator));
+            }
+            if !found_union
+                && self.compare_stack(
+                    &[
+                        Cond::keyword("for"),
+                        Cond::keyword("optional"),
                         Cond::Production("Identifier"),
                         Cond::keyword("in"),
                     ],
--- a/edb/edgeql/parser/grammar/expressions.py
+++ b/edb/edgeql/parser/grammar/expressions.py
@@ -159,23 +159,45 @@ class GroupingElementList(
     pass
 
 
-class OptionalOptional(Nonterm):
-    def reduce_OPTIONAL(self, *kids):
-        self.val = True
-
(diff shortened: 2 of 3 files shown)


aljazerzen:
For in EdgeQL works a bit different than for in say Python, and I feel like UNION is a nice hint at that.

Also, I don't like having a lot of very similar syntax doing the same thing - we can stick with UNION and own it.

All that said, the new syntax is ok, if you don't know the current one, because then you try to scan the code for the UNION to determine when the FOR body starts.

Overall, I'm -0.5 for this change on the scale from -1 to 1.

haikyuu:
As an end user, I agree with @aljazerzen 

Maybe even the `for` itself is confusing!
```
edgedb> FOR x in {1, 2, 3} UNION { x, 2 };
{1, 2, 2, 2, 3, 2}
```

`For` hints to imperative programming ... So maybe that's why it seems a bit off.

For example, this is the equivalent in programming languages:
```ruby
[1, 2, 3].flatMap((a)=>[a, 2]);
# or a more readable
[1, 2, 3].flatMap do(a) [a, 2]
```

So The `for` conveys iteration but not the `flatMap`... It's conveyed by the `union` operator.
```
for x in {1, 2, 3} union { x, 2 };
```

Maybe what's off is the way it's interpreted by the brain while reading: We expect `for x in set, do something` whereas `union` is a noun.

So I think either we change `for` to something else, or `union`. And since `union` has its place in edgedb, I'm in favor of changing the `for`.
```
flatmap x in {1, 2, 3} into {x, 2}

```
`flatmap x in set into set` is a concept in standard libraries for all major programming languages. So it is definitely more readable than `for x in set union set`.

The downside is that - **like `for`** - it doesn't have an equivalent in set theory. But it's a functional programming concept, which makes it fit nicely in Edgedb.

msullivan:
Oh, @elprans's original suggestion of using a ParenExpr doesn't work because `for x in foo(expr)` is ambiguous between whether it is a function call or a UNION-less for. This was complicated by a parser generator bug that failed to report the conflicts and instead generated a broken parser.

msullivan:
The flattening is a consequence of edgedb's underlying "set theory", where we don't have sets of sets, but collapse them all.
I think flatmap might be too obscure of a name for such a core primitive

1st1:
I'm also -1 on changing the syntax. It works and I saw few people complaining about it (unlike about our previous SDL syntax, the previous IF expression, quirks with computeds etc.) IMO we should focus on other important things.

PR #6343, opened 2023-10-24
Add styled container to verification pages

scotttrinh:
Verification pages lacked a container element since they are not forms. Moves the container element styles to a `container` class, and then updates our existing `form` styling and the new non-form "pages" to have this class.

---

Ran Black on the UI module on accident, but decided to just roll with it here as a separate commit.

### DIFF
--- a/edb/server/protocol/auth_ext/_static/styles.css
+++ b/edb/server/protocol/auth_ext/_static/styles.css
@@ -34,7 +34,7 @@ body:after {
   max-height: 100px;
 }
 
-form {
+.container {
   grid-row: 2;
   background: #fff;
   padding: 24px;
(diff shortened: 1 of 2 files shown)


edb/server/protocol/auth_ext/_static/styles.css
```diff
@@ -232,25 +232,26 @@ a.field-note:hover {
 @media (prefers-color-scheme: dark) {
   body {
     background: #191C1F;
+    color: #DEE2E6;
```
scotttrinh, line 235: This is a change due to the default color of the UI text being the light-themed color still. Lemme know if there is a more appropriate body text color.

jaclarke (approved):

PR #6344, opened 2023-10-24
Refactor match/case into dispatcher

scotttrinh:
Our `handle_request` method had organically grown from a few endpoints to many, and ended up becoming too complex for type checking and IDE integration 😬  It also required being really careful about the scope of variables defined in each `case` and made it pretty hard to follow.

This change makes `handle_request` just a dispatch method that calls other methods on the class to actually handle the logic of the endpoint.

Now that type checking is back, it found a few cases of calls to potentially unbound variables, and a few other type issues, so I think those are all resolved now.

### DIFF
--- a/edb/server/protocol/auth_ext/config.py
+++ b/edb/server/protocol/auth_ext/config.py
@@ -21,14 +21,13 @@ from typing import Optional
 
 
 class UIConfig:
-    def __init__(
-        self,
-        app_name: Optional[str] = None,
-        logo_url: Optional[str] = None,
-        dark_logo_url: Optional[str] = None,
-        brand_color: Optional[str] = None,
-    ) -> None:
-        self.app_name = app_name
-        self.logo_url = logo_url
-        self.dark_logo_url = dark_logo_url
-        self.brand_color = brand_color
+    redirect_to: str
+    redirect_to_on_signup: Optional[str]
+    app_name: Optional[str]
+    logo_url: Optional[str]
+    dark_logo_url: Optional[str]
+    brand_color: Optional[str]
+
+
+class ProviderConfig:
+    name: str
--- a/edb/server/protocol/auth_ext/http.py
+++ b/edb/server/protocol/auth_ext/http.py
@@ -372,6 +372,7 @@ class Router:
             db=self.db, provider_name=register_provider_name
         )
         require_verification = local_client.provider.config.require_verification
+        pkce_code: Optional[str] = None
 
         try:
             identity = await local_client.register(data)
@@ -416,6 +417,8 @@ class Router:
                         {"verification_email_sent_at": (now_iso8601)}
                     ).encode()
                 else:
+                    if pkce_code is None:
+                        raise errors.PKCECreationFailed
                     response.body = json.dumps({"code": pkce_code}).encode()
         except Exception as ex:
             redirect_on_failure = data.get(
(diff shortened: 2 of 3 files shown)


edb/server/protocol/auth_ext/http.py
```diff
@@ -70,995 +72,49 @@ async def handle_request(
...
-                    query_params["code"] = [pkce_code]
-                    new_query = urllib.parse.urlencode(query_params, doseq=True)
-                    new_url = parsed_url._replace(query=new_query).geturl()
-
-                    session_token = self._make_session_token(identity.id)
-                    response.status = http.HTTPStatus.FOUND
-                    response.custom_headers["Location"] = new_url
-                    _set_cookie(response, "edgedb-session", session_token)
-
+                    return await self.handle_callback(*handler_args)
```
scotttrinh, line 82: I also considered a dictionary of tuples to handlers, but this seemed a _bit_ more flexible just in case we needed to add any logic in the dispatch in the future. I'm very open to feedback here!

edb/server/protocol/auth_ext/config.py
```diff
@@ -21,14 +21,13 @@
...
+    redirect_to: str
+    redirect_to_on_signup: Optional[str]
+    app_name: Optional[str]
+    logo_url: Optional[str]
+    dark_logo_url: Optional[str]
+    brand_color: Optional[str]
+
+
+class ProviderConfig:
+    name: str
```
scotttrinh, line 33: I'm not sure I _quite_ understand the difference here, but it seemed weird to have an initializer on this class when we're really just using it so we can cast it as an interface/type.

edb/server/protocol/auth_ext/http.py
fantix: As we're here, perhaps `edb.common.value_dispatch` would be nice too?

scotttrinh: Hmm.. Looking at the implementation and I'm not sure that meets the usage here: we're dispatch based on (for now exact) matches of a list of strings rather than the value of an argument. Maybe I'm misunderstanding how it'd be used?

edb/server/protocol/auth_ext/config.py
fantix: :+1: Right, we don't need the initializer here I believe - other than the ability to instantiate this type, I don't think the initializer brings anything else differently compared to the new way you used here for typing purposes.

edb/server/protocol/auth_ext/http.py
fantix: Not tested, but I'm thinking about something like this:

```python
    async def handle_request(
        self, request: Any, response: Any, args: list[str]
    ):
        ...
        try:
            return await self.handle(args, request, response)
        except errors.NotFound as ex:
            ...

    @value_dispatch.value_dispatch
    async def handle(self, args, request: Any, response: Any):
        ...

    @handle.register(("authorize",))
    async def handle_authorize(self, _args, request: Any, response: Any):
        ...
```

fantix: You're right, `value_dispatch` doesn't support pattern bindings yet, but it shouldn't be hard to add a `pattern_dispatch` or something if needed. Before that, we could mix using `match` in a way like:

```python
    @value_dispatch.value_dispatch
    async def handle(self, args, request: Any, response: Any):
        match args:
            case('ui', '_static', filename):
                ...

            case _:
                raise errors.NotFound("Unknown auth endpoint")

    @handle.register(("authorize",))
    async def handle_authorize(self, _args, request: Any, response: Any):
        ...
```

scotttrinh: Oh that's so clever! I love it. Will do.

scotttrinh: @fantix The one thing that isn't yet obvious to me is what the `handle` function looks like so that it has the right fallback order. I'll play around with it and if I can't figure it out can I merge what we have here and revisit later?

fantix (approved):

edb/server/protocol/auth_ext/http.py
scotttrinh: Alright, looks like it's a little bit of a larger refactor since this would require moving from a class to flat functions with a context struct, so I've spun that work off into a separate branch so this can land without being blocked.

PR #6346, opened 2023-10-24
Fix coalesce with DML in non-dev builds

msullivan:
Additionally, stop conditioning these testing features on whether it
is dev build but instead based on whether the testmode flag is
set. Moreover, get rid of the code path for testing whether an
instance is a dev instance, since anything conditioned on that *can't*
be tested by our test suite as it currently stands.


elprans (approved):

msullivan:
We should make `INTERNAL_TESTMODE = False` the default, though, but I didn't want to deal with that today

PR #6352, opened 2023-10-25
Support auth for HTTP endpoints

### LINKED ISSUE #6345: Implement auth for GraphQL and HTTP
A user wants to expose the SQL protocol outside their VPC, but they can't because this would also expose GQL and HTTP protocols without access control. Implement RBAC on these protocols so that it is safe to expose outside the VPC.

msullivan:
Currently, all extension endpoints except for 'auth' will go through
authentication, since it seems like the safe default.

Fixes #6345.

This is a backwards incompatible change! Unless "Trust" is configured
as a method for HTTP transports (which it will be in dev instances),
HTTP endpoints will stop working when not authenticated to.

Authentication is done by adding a header of the form:
```
Authorization: bearer <JWT>
```

We'll need to update the docs. For cloud instances, you just use the
JWT token.

I'm not sure how painful it will be to adjust other hosted prod
deployments of edgedb to be able to use JWT tokens, though?
Thoughts?

If it seems like it might be nontrivial, we can add a new transport
type so that the simple HTTP endpoints and tunnelled edgedb+http can
be separated, which would make it easy to opt-in to exactly the old
behavior.

EDIT: I added a separate transport type for SIMPLE_HTTP (feel free to give me a better name)
and support HTTP Basic auth for password authentication, which should make life easier for everyone.


msullivan:
(I have a patch to add a new transport type to split the edgedb+http from the other HTTP endpoints ready, if we want to do that)

elprans (commented):
We should be able to support normal password auth with basic HTTP auth as long as the connection is secure (i.e TLS).

aljazerzen:
I don't think I can review this, but my safe guess is that it blocks the rc1.

msullivan:
I just pushed basic password auth over simple http endpoints. Might be worth reviewing now, though there is a still pending commit to add tests. (I've tested it manually, though)

edb/server/args.py
```diff
@@ -150,6 +151,7 @@ def __str__(self):
     ServerConnTransport.TCP: ServerAuthMethod.Scram,
     ServerConnTransport.TCP_PG: ServerAuthMethod.Scram,
     ServerConnTransport.HTTP: ServerAuthMethod.JWT,
+    ServerConnTransport.SIMPLE_HTTP: ServerAuthMethod.Scram,
```
fantix, line 154: This is a bit misleading, as we're not sending any "Salted Challenges" as in SCRAM. Should we have a new `ServerAuthMethod.Password` enum?

edb/server/protocol/protocol.pyx
```diff
@@ -706,6 +715,52 @@ cdef class HttpProtocol:
...
+    async def _check_http_auth(
+        self,
+        HttpRequest request,
+        HttpResponse response,
+        str dbname,
+    ):
+        dbindex: dbview.DatabaseIndex = self.tenant._dbindex
+
+        try:
+            user = 'edgedb'  # I guess?
```
fantix, line 727: I think this should be the `"sub"` in the JWT token or the username in the `Basic` authorization value, or `cfg::Auth.user` won't work if it's set to non-`edgedb` values. But that's probably an acceptable unimplemented feature at this moment, because this endpoint wasn't protected at all.

fantix (approved):
Otherwise, this looks good to me!

edb/server/args.py
msullivan: Yeah, let me test if the config patches that would be necessary to backport it to 4.x work.
(Since it's actually an object in the config system)

edb/server/protocol/protocol.pyx
msullivan: `sub` isn't always present in the JWTs we generate, though. Ones for local instances have no `sub` but do have `"edgedb.server.any_role"` set.

Our cloud tokens do have a `sub` set, and it seems to be a random alphanumeric string; does it actually correspond to a role in edgedb?

edb/server/args.py
msullivan: Yeah I think this should work

msullivan: ... maybe. I'm getting some really weird errors testing the new auth extension after in an place upgrade... We'll see.

msullivan: Ok got it

edb/server/protocol/protocol.pyx
elprans: `sub` in the cloud token is the id of the cloud user, no relation to role, as those are specified in token scope instead.  We have the `X-EdgeDB-User` header to communicate the desired role instead.  Default to `edgedb` if it isn't present I guess.

msullivan:
Alright, I added proper user support, so I think this is basically there.

I might also tweak it to respond `401 Unauthorized` and send a `[WWW-Authenticate](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate)` header if no authorization header is sent. urllib at least required a very particular incantation to get it to send a password if you don't do that.

(Of course, just injecting the basic auth header manually is actually way easier than navigating that...)

fantix (approved):

PR #6353, opened 2023-10-25
Tweak error recovery costs

aljazerzen:


aljazerzen:
I promise I'll make up a better strategy to improve error messages compared to what I'm doing now (which is just adding a new "tweak error recovery" commit every time we find a new confusing keyword injection).

But for 4.x, this is what it is.

msullivan (approved):

PR #6361, opened 2023-10-26
[4.x] Support auth for HTTP endpoints (#6352)

### LINKED ISSUE #6345: Implement auth for GraphQL and HTTP

msullivan:
4.x backport notes: This required a patch, and to help test that I added more tests to the patch tests.

This makes all extension endpoints except for 'auth' will go through authentication, since it seems like the safe default.

We add a new SIMPLE_HTTP transport method for HTTP endpoints. (Unfortunately, edgedb+http already uses HTTP, even though HTTP would be a better name for *this* one.)

We also add a new Password auth method that only works for SIMPLE_HTTP, and which is the default method for SIMPLE_HTTP. JWT authentication is also supported, but SCRAM is not.

This is a backwards incompatible change! Unless "Trust" is configured as a method for HTTP transports (which it will be in dev instances), HTTP endpoints will stop working when not authenticated to.

Fixes #6345.

Password authentication uses HTTP basic auth.

JWT authentication is done by adding a header of the form:
```
Authorization: bearer <JWT>
```

The username for JWT and Trust auth can be specified with the X-EdgeDB-User header.

### DIFF
--- a/.github/workflows.src/tests-patches.tpl.yml
+++ b/.github/workflows.src/tests-patches.tpl.yml
@@ -67,7 +67,7 @@ jobs:
         # has timeouts.
         edb server --bootstrap-only --data-dir test-dir
         # Should we run *all* the tests?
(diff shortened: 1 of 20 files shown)


jnak:
Sweet. Will this be included in EdgeDB 4.0 ?

msullivan:
Yes

PR #6364, opened 2023-10-26
Do not log exception on invalid data

scotttrinh:
The response should include enough information to be useful for debugging, no need to also log to stdout.


msullivan (approved):

msullivan:
In some places in the server, we check if a debug flag is set and dump exceptions then. If you think that would help your debugging in the future, go ahead and do thta

scotttrinh:
@msullivan 

meh, we're already providing the message in the response, so this was strictly duplicate output IMO. good to know though!

PR #6367, opened 2023-10-26
Don't cache UI assets in memory when the Admin UI is disabled

elprans:
Saves ~10MiB or about 10% of all Python allocations in the baseline.


fantix (approved):

msullivan:
Do many realistic workflows not have the admin UI enabled?

What would be best is to not cache in memory at all, except that file reads aren't async and sendfile isn't implemented in uvloop

elprans:
> Do many realistic workflows not have the admin UI enabled?

Cloud doesn't, because we run a dedicated UI for all instances.

PR #6369, opened 2023-10-26
Make ext::auth::ClientTokenIdentity single

msullivan:
We don't have a great story currently for how to fix this kind of
extension issue, so for now let's resort to the tried-and-true method
and fixing it in the compiler :(.


1st1:
Should we also fix the auth schema for the future major releases?

msullivan:
> Should we also fix the auth schema for the future major releases?

Yeah that'll be a follow up

elprans (approved):

PR #6377, opened 2023-10-27
Include auth_ext static assets in the package build

msullivan:
I verified that the assets are in the sdist.

Also add a test that fetches one of the assets. I'll kick off a
release dry-run to verify that it works.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -4,3 +4,4 @@ include LICENSE README.md logo.svg
 
 recursive-include edb/edgeql-parser *
 recursive-include edb/edgeql-parser/edgeql-parser-python *
+recursive-include edb/server/protocol/auth_ext/_static *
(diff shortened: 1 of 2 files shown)


scotttrinh (approved):
Cool! Test makes sense to me, but does that really test that it is packaged?

elprans:
> Cool! Test makes sense to me, but does that really test that it is packaged?

Tests run on production packages, so yes.

msullivan:
> > Cool! Test makes sense to me, but does that really test that it is packaged?
> 
> Tests run on production packages, so yes.

To clarify, release/nightly/dry-run build tests run on production packages, and I kicked off a dry-run to test this, which looks good (all the failed tests are parser related and will be fixed by https://github.com/edgedb/edgedb/pull/6375): https://github.com/edgedb/edgedb/actions/runs/6672556465/job/18137460047

Running tests locally (and I think in CI) use the dev build.

scotttrinh:
Ahh, that was the piece I was missing to tie together the dry run and this test 🙌

PR #6383, opened 2023-10-28
Provide an error message for failed PKCE verification

scotttrinh:
Provide a helpful message when challenge and verifier do not match.


msullivan (approved):
Rebasing should fix the test failure

PR #6385, opened 2023-10-29
Allow `if .. then .. else` at the top level

elprans:
Since `if` is similar to `for` in its control-like nature and we allow
the latter, allow writing `if` at the top level without an extra
`select`, which is especially useful with the recent ability to do
conditional DML.

    edgedb> if user_id
    ....... then (select User filter .id = user_id)
    ....... else (insert User)


msullivan (approved):

1st1:
Typo in the commit message: conditional DML

1st1:
Can you add one functional test?

PR #6393, opened 2023-10-30
Fix `ext::auth::ClientTokenIdentity` cardinality

scotttrinh:
Since `conf.key` is a set, we end up with a multi cardinality


msullivan (approved):
This looks good.

Also revert the compiler changes in https://github.com/edgedb/edgedb/pull/6369 as part of this

msullivan (approved):

aljazerzen:
Just to confirm @msullivan: this would be hard to patch into 4.1?

PR #6398, opened 2023-10-31
Create an allow-list of redirect URLs

scotttrinh:
In various flows, we use application-provided URLs to redirect the user back into the application. These URLs should be allow-listed and checked before redirecting the user to mitigate against spoofing.

### DIFF
--- a/edb/lib/ext/auth.edgeql
+++ b/edb/lib/ext/auth.edgeql
@@ -252,7 +252,7 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
             create annotation std::description :=
                 "When redirecting the user in various flows, the URL will be \
                 checked against this list to ensure they are going \
-                to a trusted domain controlled by thee application. URLs are \
+                to a trusted domain controlled by the application. URLs are \
                 matched based on checking if the candidate redirect URL is \
                 a match or a subdirectory of any of these allowed URLs";
         };
--- a/edb/lib/ext/auth.edgeql
+++ b/edb/lib/ext/auth.edgeql
@@ -247,6 +247,15 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
                 indicates that the token should never expire.";
             set default := <std::duration>'336 hours';
         };
+
+        create multi property allowed_redirect_urls -> std::str {
+            create annotation std::description :=
+                "When redirecting the user in various flows, the URL will be \
+                checked against this list to ensure they are going \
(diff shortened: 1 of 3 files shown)


edb/lib/ext/auth.edgeql
```diff
@@ -247,6 +247,15 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
                 indicates that the token should never expire.";
             set default := <std::duration>'336 hours';
         };
+
+        create multi property allowed_redirect_urls -> std::str {
+            create annotation std::description :=
+                "When redirecting the user in various flows, the URL will be \
+                checked against this list to ensure they are going \
+                to a trusted domain controlled by thee application. URLs are \
```
msullivan, line 255: s/thee/the

msullivan:
The test failure is yet another casualty of the just-in-time implementation of extension module config. It's fixed in https://github.com/edgedb/edgedb/pull/6400

edb/server/protocol/auth_ext/http.py
```diff
@@ -1317,6 +1390,37 @@ async def _try_verify_email(
...
+    def _is_url_allowed(self, url: str) -> bool:
+        allowed_urls = util.get_config(
+            self.db,
+            "ext::auth::AuthConfig::allowed_redirect_urls",
+            frozenset,
+        )
+        allowed_urls = cast(Set[str], allowed_urls)
+        parsed_url = urllib.parse.urlparse(url)
+
+        for allowed_url in allowed_urls:
```
elprans, line 1402: Why not just do string prefix comparison as `any(url.startswith(allowed) for allowed in allowed_urls))`?  Also, should this be done case-insensitively?

scotttrinh: I was trying to match GitHub's matching algorithm which does this domain-minus-subdomain matching.

scotttrinh: @elprans I ended up just going with a `for` loop with an early `return` since that seems _slightly_ better than iterating the whole list? Thoughts?

msullivan: The `any`+generator fragment that Elvis suggested will short circuit also, since `any` will return as soon as it sees a true element, and so the rest of the generator won't execute.
Though probably your version is still faster because it doesn't need to deal with a generator.

msullivan (approved):

PR #6412, opened 2023-11-01
Add auth guide

raddevon:


haikyuu:
Trying to wrap my head about auth. I think this one requires a video example ... Maybe using express and edgedb (nothing fancy)

MiroslavPetrik:
Perhaps the [schema extensions](https://www.edgedb.com/docs/datamodel/extensions) page needs an update as well?

scotttrinh:
@haikyuu 

We plan on publishing libraries for popular web frameworks like Express, so rather than show the guide using a framework, we plan on showing it as a standard Node HTTP server, so it's easy to translate into other languages and contexts that we won't be publishing libraries for any time soon.

Stay tuned!

docs/guides/auth/built_in_ui.rst
```diff
@@ -0,0 +1,241 @@
...
+Configuring
+===========
+
+We will demonstrate the various steps below by building a NodeJS HTTP server in
+a single file that we will use to simulate a typical web application. **Note**:
+We are in the process of publishing helper libraries that you can use with
+popular languages and web frameworks. The details below show the inner workings
+of how data is exchanged with the Auth extension from a web app using HTTP. You
+can use this as a guide to integrate with your application written in any
+language that can send and receive HTTP requests.
```
raddevon, line 38: Once the first library launches, maybe we can add another admonition (`warning` type) that says something about users of the respective framework (Next.js, I assume) using that instead of trying to integrate from scratch.

```suggestion
a single file that we will use to simulate a typical web application.

.. note::

    We are in the process of publishing helper libraries that you can use with
    popular languages and web frameworks. The details below show the inner
    workings of how data is exchanged with the Auth extension from a web app
    using HTTP. You can use this as a guide to integrate with your application
    written in any language that can send and receive HTTP requests.
```

scotttrinh: > users of the respective framework (Next.js, I assume) using that instead of trying to integrate from scratch.

Yeah, and I think we should have the docs for each of those frameworks in the docs here so we can link directly from that note.

scotttrinh (approved):
Great! Assuming CI works, we are good to go here.

docs/guides/auth/index.rst
```diff
@@ -0,0 +1,308 @@
...
+
+
+token_time_to_live
+------------------
+
+This value controls the expiration time on the authentication token’s
+JSON Web Token. This is effectively the “session” time.
+
+
+allowed_redirect_urls
```
MiroslavPetrik, line 67: @raddevon 

Since this parameter is not updatable via the `edgedb ui` (yet i suppose), I tried to update it manually and had hard time doing so.

Finally was able to do it via the configure command:

```
configure current database set ext::auth::AuthConfig::allowed_redirect_urls := {"http://localhost:4000/", "http://localhost:3000/"}
```

Maybe it would be good to add code snippets on how to set each of the property via REPL? 

First I was trying to update the 'table' via regular query, with no luck
```
 update  ext::auth::AuthConfig set { allowed_redirect_urls := "http://localhost:3000/" };
edgedb error: InternalServerError: cannot delete from view "eb39dffd-d950-53ad-9974-c14c22ebd918"
  Server traceback:
      edb.errors.InternalServerError: cannot delete from view "eb39dffd-d950-53ad-9974-c14c22ebd918"
```

So I imagine multiple users could struggle with this as well.

PR #6421, opened 2023-11-02
Fix DML with constraints on abstract types

### LINKED ISSUE #6419: Queries fail on abstract types
- EdgeDB Version: 4.0+d568f08
- EdgeDB CLI Version: EdgeDB CLI 4.1.0-dev
- OS Version: x86_64 GNU/Linux 6.5.9-arch2-1

Even simple query fails: 
```
update ChatBase
filter (.id = <uuid>$chat_id) and (<uuid>$my_id in .members.id)
set { 
  messages += (
    insert Message { 
      from := (select User filter .id = <uuid>$my_id), 
      text := <str>$text 
    }
  ) 
}
```
with error:
```
InternalServerError: <ir.TypeRef 'default::ChatBase' at 0x7fd7d64c4dd0> 
```
And it works if I repalce ChatBase with Dialog

Schema:
messaging.esdl
```edgeql
module default {
   abstract type ChatBase {
      multi members: User {
         on target delete allow;
      };
      multi messages: Message {
         on target delete allow;
         on source delete delete target;
         constraint exclusive;
      }
   }
   type Dialog extending ChatBase {
      overloaded required multi members: User;
      trigger require_two_members after insert, update for each do (
         assert(
            count(__new__.members) <= 2,
            message := "Dialog can have only two members, now are: " ++ to_str(count(__new__.members))
         )
      );
      trigger prohibit_same_dialogs after insert for each do (
         assert_single(
            (select Dialog filter (count(__new__.members intersect .members) = 2)),
            message := "Can't create more than one dialog with same members"
         )
      );
   }
   type Monolog extending ChatBase {
      required single link owner: User {
         constraint exclusive;
      };
   }
   type GroupChat extending ChatBase {
      pictureId: str;
   }
   scalar type MessageText extending str {
      constraint min_len_value(1);
      constraint max_len_value(1000);
   }
   type Message {
      from: User {
      };
      required single link chat := assert_exists(.<messages[is ChatBase]);
      required text: MessageText;
      required createdAt: datetime {
         readonly := true;
         default := datetime_of_statement();
      }
   }
   type DraftMessage {
      required from: User {
         on source delete delete target;
      }; 
      chat: ChatBase;
      recipient: User;
      constraint expression on (
         (exists .chat and not exists .recipient) 
         or 
         (exists .recipient and not exists .chat)
      );
      required text: MessageText;
   }
}
```
default.esdl:
```edgeql
module default {
   scalar type Activity extending enum<Investor, Entrepreneur, Freelancer>;
   function is_unique_array(arr: array<Activity>) -> bool
      using (count(distinct (array_unpack(arr))) = count(array_unpack(arr)));
   type User {
      required phone: str {
         constraint exclusive;
         constraint expression on (len(__subject__) = 12);
      };
      name: str;
      nickname: str {
         constraint exclusive;
      };
      pictureId: str;
      required activities: array<Activity> {
         constraint expression on (is_unique_array(__subject__));
         default := <array<Activity>>[];
      };
      required interests: array<str> {
         default := <array<str>>[];
      };
      statusMessage: str;
      property canBeVisible := exists(.name) and exists(.nickname);
      link chats := .<members[is ChatBase];
   }
   type OtpPhoneRequest {
      required phone: str;
      required otp: int32;
      required sentAt: datetime;
      confirmedAt: datetime;
   }
}
```

aljazerzen:
Some constraints are inforced by adding a CTE that does the check.
When an a constraint was defined on an abstract type, we were trying
and failing to generate this check for the abstract type as well.

This PR removes this check for all abstract object types.

Closes #6419

### DIFF
--- a/edb/edgeql/compiler/conflicts.py
+++ b/edb/edgeql/compiler/conflicts.py
@@ -800,6 +800,11 @@ def compile_inheritance_conflict_checks(
 
     conflicters = []
     for subject_stype, anc_type, ir in modified_ancestors:
+
+        # don't enforce any constraints for abstract object type
+        if subject_stype.get_abstract(schema=ctx.env.schema):
+            continue
+
         conflicters.extend(
             _compile_inheritance_conflict_selects(
                 stmt, ir, anc_type, subject_stype, ctx=ctx
--- a/edb/pgsql/compiler/dml.py
+++ b/edb/pgsql/compiler/dml.py
@@ -431,6 +431,12 @@ def fini_dml_stmt(
             else:
                 stop_ref = base_typeref
 
+            # When the base type is abstract, there will be no CTE for it,
+            # so the overlays of children types have to apply to the whole
+            # ancestry tree.
+            if base_typeref.is_abstract:
+                stop_ref = None
+
             # The overlay for update is in two parts:
             # First, filter out objects that have been updated, then union them
             # back in. (If we just did union, we'd see the old values also.)
--- a/tests/test_constraints.py
+++ b/tests/test_constraints.py
@@ -1774,7 +1774,7 @@ class TestConstraintsDDL(tb.DDLTestCase):
         ):
             await self.con.execute("insert X { y := -1 }")
 
-    async def test_constraints_abstract_object(self):
+    async def test_constraints_abstract_object_01(self):
         await self.con.execute(
             """
                 create abstract type ChatBase {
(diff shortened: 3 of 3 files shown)


edb/edgeql/compiler/conflicts.py
```diff
@@ -800,6 +800,11 @@ def compile_inheritance_conflict_checks(
 
     conflicters = []
     for subject_stype, anc_type, ir in modified_ancestors:
+
+        # don't enforce any constraints for abstract object type
```
msullivan, line 804: I think this is wrong. I believe that skipping this check means that we won't generate a constraint check against an abstract common ancestor, so we might miss enforcing a constraint.

I suspected that was the case, so I tried a variant of your test case that made the property single, and it does incorrectly succeed.

I'm not sure why the multi version works.

msullivan (commented):
I also checked, and this was a regression introduced by #6135.

We could just revert that, if we can't find a better fix, though.

I suspect the right fix here is on the pgsql side

edb/edgeql/compiler/conflicts.py
msullivan: I pushed the test case I wrote

msullivan: I was wrong that this was wrong. I had mixed up which type was which, and thought that it was being skipped when the source of the constraint was abstract. This code should be right.

But it was good that I was wrong about this and decided to poke at it, because the test is of course still failing!

msullivan:
I tracked down the source of the failure I mentioned. One key thing was that it *did* work when it was a multi prop. I looked at the code generated for multi props, and the obvious difference was that the `else~` CTEs for multi props were applying any overlay, while the single ones weren't. This is bad, because the data we are checking for conflicts is the newly added data; we need to use an overlay!

Then I did a quick search to see where `process_update_conflicts` was called from, and the relevant buggy code turned out to be immediately above it: https://github.com/edgedb/edgedb/blob/master/edb/pgsql/compiler/dml.py#L424-L446

When we add an overlay to our map of overlays, we register it as being an overlay for all of its ancestor types, too. (Since it adds things to or modifies those ancestor types).

When we do updates, though, we skip adding overlays for all the descendants to the overlay map for the base type of the update; this is because updates generate a `union_cte` that unions all of the different subtype updates together, and so we use that as the overlay for the update base type, instead of having a separate overlay for each subtype.

But because https://github.com/edgedb/edgedb/pull/6135 skipped generating an update for the abstract base, it isn't in `dml_ctes`, so that `union_cte` nevers get added as an overlay!

Does that make sense?

aljazerzen:
I does somewhat make sense. It's a lot and I can try to explain it back to you on Tuesday, as a practice :D

So:
- We don't have to check the constraints for abstract types - even for catching exclusive constraints between different subtypes. My fix is correct.
- My fix and your investigation uncovered similar bug for single properties, which I'll fix now.

msullivan (approved):

PR #6443, opened 2023-11-08
docs: Update the Get Started docs to 4.1

elliotwaite:
With the release of v4, I thought it might be a good idea to update the Get Started docs to match.


elliotwaite:
Also, other parts of the Quickstart could be updated to correspond to the new UI, but doing so would also require updating the screenshots to match the new instructions. Some places that could be updated:
* "Then click `Open REPL`" could be changed to "Then click `Open Editor`".
* "Click `COPY AS JSON`" could be changed to "Click `COPY`".

Dhghomon (approved):
Looks good to me, thanks for another helpful PR!

PR #6446, opened 2023-11-08
Introspection for Property.secret

### LINKED ISSUE #6444: Expose if a property is marked as secret in introspection
Some properties are marked as secret in the auth ext schema, and cause an error to be thrown if you try to query them. It would be useful if whether a property is secret was exposed in schema introspection, so that the UI could omit these properties when querying data for the data view. (See https://github.com/edgedb/edgedb-ui/issues/282)

aljazerzen:
Closes #6444

### DIFF
--- a/tests/test_edgeql_introspection.py
+++ b/tests/test_edgeql_introspection.py
@@ -1590,3 +1590,18 @@ class TestIntrospection(tb.QueryTestCase):
             } LIMIT 0''',
             []
         )
+
+    async def test_edgeql_introspection_secret(self):
+        await self.con.execute(
+            '''
+            CREATE EXTENSION pgcrypto;
+            CREATE EXTENSION auth;
+            '''
+        )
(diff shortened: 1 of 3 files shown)


tests/test_edgeql_introspection.py
```diff
@@ -1590,3 +1590,18 @@ async def test_edgeql_schema_inheritance_computed_backlinks_01(self):
...
         )
+
+    async def test_edgeql_introspection_secret(self):
+        await self.con.execute(
+            '''
+            CREATE EXTENSION pgcrypto;
+            CREATE EXTENSION auth;
+            '''
+        )
+        await self.assert_query_result(
```
msullivan, line 1601: Could you move this test case into `test_http_ext_auth.py` or somewhere else `auth` is already created? Creating it is pretty slow.

msullivan (approved):

PR #6447, opened 2023-11-08
Add more metrics

fantix:
- [x] Current number of databases.
- [x] Time a client connection is open.
- [x] Number of compiled/cached GraphQL queries.
- [x] Number of SQL queries since instance startup.
- [x] Number of SQL compilations since instance startup.
- [x] Time it takes to compile a query or script.
- [x] Queries per connection (EdgeQL / SQL only).
- [x] Number of transaction serialization errors.
- [x] Number of network errors.
- [x] Number of Auth API calls.
- [x] Number of Auth UI form rendered.
- [x] Number of Auth successful logins.
- [x] Number of OAuth providers configured.
- [x] Average size of queries.

### DIFF
--- a/edb/server/metrics.py
+++ b/edb/server/metrics.py
@@ -157,6 +157,12 @@ transaction_serialization_errors = registry.new_labeled_counter(
     labels=('tenant',)
 )
 
+connection_errors = registry.new_labeled_counter(
+    'connection_errors_total',
+    'Number of network connection errors.',
+    labels=('tenant',)
+)
+
 ha_events_total = registry.new_labeled_counter(
     "ha_events_total",
     "Number of each high-availability watch event.",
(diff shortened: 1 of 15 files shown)


msullivan:
Should we consider putting some of this stuff behind authentication?

fantix:
> Should we consider putting some of this stuff behind authentication?

Good point! Or perhaps we can serve the /metrics endpoint only on specified host address (like localhost) or from certain client IP

edb/server/metrics.py
```diff
@@ -101,14 +114,88 @@
...
+auth_ui_renders = registry.new_labeled_counter(
+    "auth_ui_renders_total",
+    "Number of UI pages rendered by the Auth extension.",
+    labels=("tenant",),
+)
+
+auth_providers = registry.new_labeled_gauge(
+    'auth_providers',
+    'Number of Auth providers configured.',
+    labels=('tenant', 'database'),
```
fantix, line 194: Shall we add `database` to many other metrics?

msullivan (approved):

PR #6460, opened 2023-11-13
Add mTLS support

fantix:
`/metrics` and `/server/*` would require a valid client TLS certificate if `--tls-client-ca-file` is set. This behavior can be customized by setting the default authentication methods for `http_metrics` or `http_health` under `--default-auth-method`, for example:

```
edgedb-server --default-auth-method http_metrics:mTLS,http_health:trust --tls-client-ca-file ...
```

- [x] Add --tls-client-ca-file
- [x] Protect /metrics
- [x] Add mTLS tests, ~fix metrics tests~
- [x] Protect /server/*
- [x] Add test for mTLS authentication over binary protocol


edb/lib/cfg.edgeql
```diff
@@ -67,6 +67,14 @@ CREATE TYPE cfg::Password EXTENDING cfg::AuthMethod {
         SET default := { cfg::ConnectionTransport.SIMPLE_HTTP };
     };
 };
+CREATE TYPE cfg::MutualTLS EXTENDING cfg::AuthMethod {
```
elprans, line 70:
```suggestion
CREATE TYPE cfg::mTLS EXTENDING cfg::AuthMethod {
```

edb/server/args.py
```diff
@@ -135,6 +136,7 @@ class ServerConnTransport(enum.StrEnum):
     TCP = "TCP"
     TCP_PG = "TCP_PG"
     SIMPLE_HTTP = "SIMPLE_HTTP"
+    METRICS = "METRICS"
```
elprans, line 139:
```suggestion
    HTTP_METRICS = "HTTP_METRICS"
```

edb/server/protocol/auth_helpers.pyx
```diff
@@ -235,3 +232,31 @@ cdef auth_basic(tenant, username: str, password: str):
...
+
+cdef auth_mtls(transport):
+    sslobj = transport.get_extra_info('ssl_object')
+    if sslobj is None:
+        raise errors.AuthenticationError(
+            "mTLS authentication is not supported over plaintext transport")
+    cert_data = sslobj.getpeercert()
+    if not cert_data:  # None or empty dict
+        raise errors.AuthenticationError(
+            "valid client certificate required")
```
msullivan, line 245: Could you add a comment to the effect that peer certs are always verified against the configured client_ca_file, and so all we need to do is check that one exists?

elprans (approved):

msullivan (approved):

PR #6469, opened 2023-11-14
Fixes to auth redirect urls

jaclarke:
Fixes multiple issues with auth redirects:
- Fix `redirect_to_on_signup` not being used/being used when it shouldn't as the redirect url in the builtin ui
- Fix broken joining of params to redirect urls when the url already contains query params
- Add `provider` name to the params of redirects from the oauth and email/password register flows
- Add verify_url to verification_token claims for use in resend verification flow + check `verify_url` is an allowed url

### DIFF
--- a/edb/server/protocol/auth_ext/http.py
+++ b/edb/server/protocol/auth_ext/http.py
@@ -803,6 +803,7 @@ class Router:
                 base_path=self.base_path,
                 providers=providers,
                 redirect_to=ui_config.redirect_to,
+                redirect_to_on_signup=ui_config.redirect_to_on_signup,
                 error_message=_maybe_get_search_param(query, 'error'),
                 email=_maybe_get_search_param(query, 'email'),
                 challenge=maybe_challenge,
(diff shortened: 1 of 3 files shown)


edb/server/protocol/auth_ext/http.py
```diff
@@ -1492,3 +1492,13 @@ def _check_keyset(candidate: dict[str, Any], keyset: set[str]):
         raise errors.InvalidData(
             "Missing required fields: " ", ".join(missing_fields)
         )
+
+
+def _join_url_params(url: str, params: dict[str, str]):
```
scotttrinh, line 1497: ❤️

scotttrinh (approved):

edb/server/protocol/auth_ext/http.py
```diff
@@ -1240,7 +1244,7 @@ def _get_data_from_reset_token(self, token: str) -> Tuple[str, str, str]:
 
     def _get_data_from_verification_token(
         self, token: str
-    ) -> Tuple[str, float, Optional[str], Optional[str]]:
+    ) -> Tuple[str, float, str, Optional[str], Optional[str]]:
```
scotttrinh, line 1247: This is borderline ready to be a dataclass, I think, but this is still fine for now. 😆

scotttrinh (approved):

edb/server/protocol/auth_ext/http.py
jaclarke: As far as I can tell, tuples can contain up to 2^63 elements, so we should be able to pack in a few more values yet. 😄

PR #6473, opened 2023-11-14
Add note relating `migrate --dev-mode` to `watch`

raddevon:
This question came up in a recent issue (https://github.com/edgedb/edgedb-cli/issues/1170). This note should make it easier for users to discover in the future.


scotttrinh (approved):

Dhghomon:
Looks good, one other tiny change we could make while we're at it: I notice that on the current edgedb watch page the command is so simple that it almost looks like a typo or something (just a sudden `edgedb watch` all alone on its own line). Maybe add a sentence before there: "Starting edgedb watch couldn't be simpler:" or "To start the process, type:" or something similar.

PR #6475, opened 2023-11-15
Fix std::range in singleton mode

aljazerzen:


edb/pgsql/compiler/expr.py
```diff
@@ -647,9 +647,15 @@ def compile_FunctionCall(
         expr: irast.FunctionCall, *,
         ctx: context.CompilerContextLevel) -> pgast.BaseExpr:
 
-    if sfunc := relgen._SIMPLE_SPECIAL_FUNCTIONS.get(str(expr.func_shortname)):
+    fname = str(expr.func_shortname)
+    if sfunc := relgen._SIMPLE_SPECIAL_FUNCTIONS.get(fname):
         return sfunc(expr, ctx=ctx)
 
+    if expr.func_sql_expr:
+        raise errors.InternalServerError(
```
msullivan, line 655: Let's make it unimplemented feature error?

msullivan (approved):

PR #6495, opened 2023-11-17
Remove statement that tuple > JSON not reversible

### LINKED ISSUE #5888: Docs claim that we can't cast JSON to tuple
Our documentation says that JSON cannot be cast to a tuple.

> Unlike other casts to JSON, tuple casts to JSON are not reversible (i.e., it is not possible to cast a JSON value directly into a [tuple](https://www.edgedb.com/docs/stdlib/tuple#type::std::tuple)).

 This holds true in the CLI REPL, but casting _does_ appear to work when querying over HTTP.

- EdgeDB Version: 3.2+44862ed
- EdgeDB CLI Version: 3.4.0+4d95a2c
- OS Version: macOS 13.4.1

Steps to Reproduce:

1. Create new project
2. Migrate schema below
3. Send this query over http:
```bash
curl --data '{"query":"with data := <tuple<a:str,b:str>><json>$data, select data;", "variables":{"data":{"a":"eee","b":"smstr"}}}' --header 'Content-Type: application/json' http://localhost:<port>/db/edgedb/edgeql
```

The output of this query is `{"data":[{"a" : "eee", "b" : "smstr"}]}` which indicates the JSON to tuple cast has succeeded.

Schema:
```
using extension edgeql_http;
module default {
}
```

raddevon:
Closes #5888

### DIFF
--- a/docs/stdlib/json.rst
+++ b/docs/stdlib/json.rst
@@ -128,9 +128,7 @@ conditions for casting:
 
 A named :eql:type:`tuple` is converted into a JSON object when cast as a
 :eql:type:`json` while a standard :eql:type:`tuple` is converted into a
-JSON array. Unlike other casts to JSON, tuple casts to JSON are *not*
-reversible (i.e., it is not possible to cast a JSON value directly into a
-:eql:type:`tuple`).
+JSON array.
 
 ----------
 


msullivan (approved):

raddevon:
I believe you're right, @scotttrinh. Not any arbitrary JSON can be cast to tuple, but a JSON value that was cast from tuple can always be cast back.

PR #6496, opened 2023-11-17
Add connection documentation for EdgeDB Cloud

raddevon:
I tried to hit every spot we talk about connections where we didn't address EdgeDB Cloud. Please point out any I might have missed.

Not sure if we need to say anything further about CLI network usage. What I've said here is pretty minimal.

### DIFF
--- a/docs/cli/edgedb_connopts.rst
+++ b/docs/cli/edgedb_connopts.rst
@@ -23,10 +23,13 @@ Connection flags
 
 :cli:synopsis:`-I <name>, --instance=<name>`
     Specifies the named instance to connect to. The actual connection
-    parameters are stored in ``<edgedb_config_dir>/credentials`` and are
-    usually created by :ref:`ref_cli_edgedb_instance_create` or similar
-    commands. Run ``edgedb info`` to see the location of
-    ``<edgedb_config_dir>`` on your machine.
+    parameters for local and self-hosted instances are stored in
+    ``<edgedb_config_dir>/credentials`` and are usually created by
+    :ref:`ref_cli_edgedb_instance_create` or similar commands. Run ``edgedb
+    info`` to see the location of ``<edgedb_config_dir>`` on your machine.
+
+    EdgeDB Cloud instance names are in the format
+    ``<github-username>/<instance-name>``.
 
     This option overrides host and port.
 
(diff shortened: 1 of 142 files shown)


docs/cli/edgedb_connopts.rst
```diff
@@ -23,10 +23,13 @@ Connection flags
...
-    usually created by :ref:`ref_cli_edgedb_instance_create` or similar
-    commands. Run ``edgedb info`` to see the location of
-    ``<edgedb_config_dir>`` on your machine.
+    parameters for local and self-hosted instances are stored in
+    ``<edgedb_config_dir>/credentials`` and are usually created by
+    :ref:`ref_cli_edgedb_instance_create` or similar commands. Run ``edgedb
+    info`` to see the location of ``<edgedb_config_dir>`` on your machine.
+
+    EdgeDB Cloud instance names are in the format
+    ``<github-username>/<instance-name>``.
```
elprans, line 32: Although we are tying orgs to Github currently, it will not necessarily be the case forever, so it's worth not to tie ourselves in the docs.  Also, it's org-name, not `username`.

```suggestion
    ``<org-name>/<instance-name>``.
```

raddevon: Thanks, @elprans! I'll update this globally.

raddevon: Got this changed globally to reflect that the org name is the first segment of the Cloud instance name.

docs/intro/clients.rst
```diff
@@ -388,12 +391,74 @@ will automatically connect to the project-linked instance—no need for
...
+instance.
+
+If you have not yet created a secret key, you can do so in the EdgeDB Cloud UI
+or by running :ref:`ref_cli_edgedb_cloud_secretkey_create` via the CLI.
+
+.. lint-off
+
+.. note::
+
+    You may also set other environment variables like ``EDGEDB_DSN`` for EdgeDB
```
elprans, line 415: Using the underlying instance DSN directly is not really recommended practice.  I wouldn't highlight this unless there's a specific reason to.

raddevon: I believe the reason I did it was that we had a user trying to do this when I was putting together the PR, and they couldn't find how. I don't remember what the circumstances were though exactly. I'll go ahead and remove this note.

elprans (approved):

PR #6499, opened 2023-11-20
More info, tips etc. to Cloud front page

Dhghomon:


docs/guides/cloud.rst
```diff
@@ -8,7 +8,7 @@ Cloud
 
 EdgeDB Cloud is the easiest way to host your EdgeDB instance. We offer two ways
 to interact with EdgeDB Cloud: via our CLI or through a graphical web
-interface.
+interface nearly identical to the :ref:`ref_cli_edgedb_ui`.
```
raddevon, line 11: It's a little pedantic, but "the EdgeDB UI" or something like "the interface summoned by the edgedb ui command" so it doesn't look like we're playing fast and loose with our branding. Should be capitalized unless we're referring to the command, in other words.
```suggestion
interface nearly identical to the :ref:`EdgeDB UI <ref_cli_edgedb_ui>`.
```

docs/guides/cloud.rst
```diff
@@ -49,25 +49,25 @@ create a local project linked to your instance.
     * can contain only Latin alpha-numeric characters or ``-``
     * cannot start with a dash (``-``) or contain double dashes (``--``)
     * maximum instance name length is 61 characters minus the length of your
-      GitHub username (i.e., length of GitHub username + length of instance
+      Org name (i.e., length of Org name + length of instance
```
raddevon, line 52: Rebase this on my recent branch ([doc-cloud-connection](https://github.com/edgedb/edgedb/tree/doc-cloud-connection)), and you'll get my fixes on this, which are slightly different from yours.

docs/guides/cloud.rst
```diff
@@ -77,32 +77,59 @@ If you'd prefer, you can also manage your account via `the EdgeDB Cloud
 web-based GUI <https://cloud.edgedb.com/>`_.
 
 The first time you access the web UI, you will be prompted to log in. Once you
-log in with your GitHub account, you'll see your instance list.
+log in with your account, you'll be on the Instances tab of the front page
+which shows your instance list. The other two tabs allow you to manage your
+Org Settings, and Billing.
```
raddevon, line 82: Quoting in the context where we are referring to the name of the tab and lower-casing in the instances where we aren't.
```suggestion
log in with your account, you'll be on the "Instances" tab of the front page
which shows your instance list. The other two tabs allow you to manage your
organization settings and billing.
```

docs/guides/cloud.rst
```diff
@@ -77,32 +77,59 @@ If you'd prefer, you can also manage your account via `the EdgeDB Cloud
...
+
+Instances
+---------
 
 If this is your first time accessing EdgeDB Cloud, this list will be empty. To
-create an instance, click "Create new instance." This will pop up a modal
+create an instance, click "Create new instance". This will pop up a modal
```
raddevon, line 88: Period inside the quotation is the prevailing grammatical convention in the US. Since I don't think it will create any confusion here, I'd vote for sticking to it (even though it honestly really bugs me).
```suggestion
create an instance, click "Create new instance." This will pop up a modal
```

docs/guides/cloud.rst
```diff
@@ -77,32 +77,59 @@ If you'd prefer, you can also manage your account via `the EdgeDB Cloud
...
 allowing you to name your instance and specify the version of EdgeDB and the
 region for the instance.
 
 Once the instance has been created, you'll see the instance dashboard which
-allows you to monitor your instance, navigate to the database management page,
-and create secret keys.
+allows you to monitor your instance, navigate to the management page for its
+database(s), and create secret keys.
```
raddevon, line 94: I think this looks nicer and is still OK even if the instance has only one database.
```suggestion
databases, and create secret keys.
```

docs/guides/cloud.rst
```diff
@@ -77,32 +77,59 @@ If you'd prefer, you can also manage your account via `the EdgeDB Cloud
...
+
+.. lint-on
+
+Billing
+-------
+
+On this page you can use `Link <https://link.com/>`_,
+`Stripe's <https://stripe.com/>`_ fast-checkout solution, to manage your
+billing for EdgeDB Cloud via `a large number of card brands 
+<https://stripe.com/docs/payments/cards/supported-card-brands>`_.
```
raddevon, line 132: Link is not the only option, right? I actually don't see it at all on my Billing tab. I like the idea of mentioning it, but I want to make sure we don't create a different confusion by talking only about Link when there are other card management options or when not all users will see it.

Maybe @jaclarke can shed some light on which options will appear here and under which circumstances…

docs/guides/cloud.rst
```diff
@@ -125,8 +152,11 @@ better!
...
-  interruptions.
+* If you're experiencing a service interruption, check `our status page
+  <https://www.edgedbstatus.com/>`_ for information on what may be
+  causing it.
 * Report any bugs you find by `submitting a support ticket
-  <https://edgedb.com/p/cloud-support>`_.
+  <https://edgedb.com/p/cloud-support>`_. Note: when using EdgeDB Cloud
+  through the CLI, setting the ``RUST_LOG`` environment variable to ``info``,
+  ``debug``, or ``trace`` may also provide further info on what is happening
+  behind the scenes.
```
raddevon, line 162: > may also provide further info on what is happening behind the scenes.

I wonder if this implies we want the user to fix it themselves (when I'm guessing that won't often be possible). Maybe we change it to something like "…may provide additional debugging information which will be useful to include with your ticket."

If the user sees something in the logs they can fix, they'll probably still do it, but this makes it clear we're not trying to put the responsibility onto them.

raddevon (changes requested):
Good stuff! I suggested a few tweaks.

docs/guides/cloud.rst
Dhghomon: Ew. But okay, we'll follow that convention's lead I guess.

docs/guides/cloud.rst
Dhghomon: Looks like the layout gave me the impression that users needed to 1. Sign up for Link 2. Enter payment information, when the opposite is true: enter payment information as usual, and optionally sign up for Link for quicker payment next time. So best here is probably just to clarify that it's optional (just pushed a commit to change to that)

docs/guides/cloud.rst
Dhghomon: Sounds good, went with your suggestion.

docs/guides/cloud.rst
```diff
@@ -77,32 +77,59 @@ If you'd prefer, you can also manage your account via `the EdgeDB Cloud
...
+
+.. lint-on
+
+Billing
+-------
+
+On this page you can manage your account type, payment methods, and set your
+email to send billing info. Optionally, you can also save your payment info
+using `Link <https://link.com/>`_, `Stripe's <https://stripe.com/>`_
+fast-checkout solution.
```
raddevon, line 132:
```suggestion
email for receiving billing info. Optionally, you can also save your payment
info using `Link <https://link.com/>`_, `Stripe's <https://stripe.com/>`_
fast-checkout solution.
```

docs/guides/cloud.rst
```diff
@@ -77,32 +77,59 @@ If you'd prefer, you can also manage your account via `the EdgeDB Cloud
...
+
+.. _org setings:
+  https://docs.github.com/en/organizations/managing-oauth-access-to-your-organizations-data/approving-oauth-apps-for-your-organization
+
+.. lint-on
+
+Billing
+-------
+
+On this page you can manage your account type, payment methods, and set your
```
raddevon, line 129: Distinguishing between the two things we're saying you can manage — account type and payment methods — and the other thing you can do — set your email.
```suggestion
On this page you can manage your account type and payment methods, and set your
```

raddevon (approved):
Looks great! Two very minor suggestions, and I think we're ready to merge it.

Dhghomon:
> Went ahead and applied those suggestions so I can merge. I'll do that once the build completes.

Thanks!

PR #6500, opened 2023-11-20
Use an in-memory cache for auth extension HTTP requests

scotttrinh:
Our original httpx caching library defaulted to using an in-memory cache. However the replacement library does not offer an in-memory cache so we are implementing our own here.


fantix (approved):
I'll try to upstream this so that we don't need to maintain it by ourselves.

PR #6501, opened 2023-11-20
Add a guide for migrating data from Postgres to EdgeDB.

vpetrovykh:


docs/guides/datamigrations/postgres.rst
```diff
@@ -0,0 +1,574 @@
...
+    Referenced by:
+        TABLE "users" CONSTRAINT "users_status_id_fkey" FOREIGN KEY (status_id) REFERENCES statuses(id)
+
+.. lint-on
+
+The ``badges`` table actually uses the ``name`` of the badge as a primary key
+as opposed to a separate ``id``. So in order to reflect that properly in
+EdgeDB we would have to add :eql:constraint:`exclusive` constraint to this
+property. Meanwhile ``not null`` makes the property ``required``, so we end up
+with a type like this:
```
raddevon, line 83:
```suggestion
The ``badges`` table uses the ``name`` of the badge as a primary key as opposed
to a separate ``id``. In order to reflect that properly in EdgeDB, we would
have to add an :eql:constraint:`exclusive` constraint to this property.
Meanwhile ``not null`` makes the property ``required``, leaving us with a type
like this:
```

docs/guides/datamigrations/postgres.rst
```diff
@@ -0,0 +1,574 @@
...
+        }
+        required description: str;
+    }
+
+The ``statuses`` table actually has a dedicated ``id`` column in addition to
+``title``. However, ``id`` in EdgeDB is a :eql:type:`uuid`, whereas in our
+original dataset it is an ``integer``. Luckily, the ``title`` is ``UNIQUE``
+and we might be able to use that to correctly map the data during our
+migration and leave the old ``id`` entirely behind (assuming that no code
+actually depends on it). This leaves us with the following type:
```
raddevon, line 99:
```suggestion
The ``statuses`` table has a dedicated ``id`` column in addition to ``title``.
However, the automatic ``id`` in EdgeDB is a :eql:type:`uuid`, whereas in our
original dataset it is an ``integer``. Luckily, the ``title`` in the source
data is ``UNIQUE``, and we might be able to use that to correctly map the data
during our migration, leaving the old ``id`` entirely behind (assuming that no
code depends on it). This leaves us with the following type:
```

docs/guides/datamigrations/postgres.rst
```diff
@@ -0,0 +1,574 @@
...
+.. lint-on
+
+The ``users`` table also has an ``id`` column, which is not a
+:eql:type:`uuid`. Instead of omitting this information we'll record it as
+``app_id`` to facilitate the transition. Potentially we still want to
+eventually drop it in favor of the built-in ``id`` from EdgeDB, but we need it
+for now. The column ``client_settings`` would become a :eql:type:`json`
+property. The columns ``badge_name`` and ``status_id`` reference ``badges``
+and ``statuses``, respectively and will become *links* instead of
+*properties*:
```
raddevon, line 146:
```suggestion
The ``users`` table, like ``statuses``, has an ``id`` column, which is not a
:eql:type:`uuid`. Instead of omitting the ``id`` data, we'll record it as
``app_id`` in EdgeDB to facilitate the transition. We may still want to
eventually drop it in favor of the built-in ``id`` from EdgeDB, but we need it
for now. The column ``client_settings`` would become a :eql:type:`json`
property. The columns ``badge_name`` and ``status_id`` reference ``badges`` and
``statuses`` respectively and will become *links* in EdgeDB instead of
*properties*, even though a property would more closely mirror how they are
stored in Postgres:
```

raddevon (changes requested):
Awesome guide, @vpetrovykh! I left several suggestions, most of them minor. One more idea here: the first time we mention an EdgeDB concept, we should link to that concept's documentation page for easy cross-referencing.

docs/guides/datamigrations/postgres.rst
```diff
@@ -0,0 +1,581 @@
...
+            constraint exclusive;
+        }
+        required description: str;
+    }
+
+The ``statuses`` table has a dedicated ``id`` column in addition to ``title``.
+However, the automatic ``id`` in EdgeDB is a :eql:type:`uuid`, whereas in our
+original dataset it is an ``integer``. Luckily, the ``title`` in the source
+data is ``UNIQUE``, and we might be able to use that to correctly map the data
+during our migration, leaving the old ``id`` entirely behind (assuming that no
```
Dhghomon, line 98: I'm curious about this part vs. the next types - is there a reason for not mapping the old id into a property like app_id (as is done later) or is it just serving as a first example of assuming that you have the id problem solved through some other means?

If the latter then I might rephrase a bit to something like "Let's assume with ``statuses`` here that we already have the old ids correctly mapped during the data, allowing us to leave the old ``id`` entirely behind..." or however you would like to put it.

docs/guides/datamigrations/postgres.rst
```diff
@@ -0,0 +1,581 @@
...
+<filename>`` command. We can then read the files and use a simple loop to
+import the data into EdgeDB:
+
+.. note::
+
+    When Postgres dumps the JSON data as text files there will be a known
+    gotcha causing all the backslashes used to escape characters inside JSON
+    string values to be doubled. This is because the *text* format causes
+    backslashes themselves to be escaped. This needs to be accounted for when
+    reading the resulting files.
```
Dhghomon, line 403: Here I would make clear that this is Python code as I could see people misreading this as EdgeQL or some other language. Maybe add this:

`Accomplishing this through the Python client library would look like the following:`

docs/guides/datamigrations/postgres.rst
```diff
@@ -0,0 +1,574 @@
...
+also mimics how the bookmarks might be created in the app going forward.
+
+After all the import scripts we end up with data that looks something like
+this:
+
+.. lint-off
+
+.. code-block:: edgeql-repl
+
+    social:edgedb> select User {
```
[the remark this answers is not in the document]
Dhghomon: Yeah, good question. Off the top of my head maybe change to this convention depending on if it's local or cloud? Would show how the instance:database prompt looks now without being overly long.

`local:db>` 

`cloud:db>`

raddevon (approved):

PR #6507, opened 2023-11-22
Fix SQL introspection of __fts_document__

### LINKED ISSUE #6486: Cannot pg_dump a database with an fts index
```
pg_dump: dumping contents of table "public.ChatMessage"
pg_dump: error: query failed: FATAL:  '__fts_document__'
pg_dump: detail: Query was: COPY public."ChatMessage" (last_updated_at, created_at, id, user_id, body_chunks, replied_to_id, audio, files, gifs, images, local_id, local_sent_at, open_graphs, videos, room_id, text_body_from_chunks, __fts_document__) TO stdout;
```

Bug is probably the fact that we are exposing `__fts_document__` somewhere in introspection, even though it is not a real pointer.

aljazerzen:
Closes #6486


msullivan (approved):
We'll need to do a patch in the backport. Doing an empty `edgeql+schema` patch should do it, I *think*, but there might be a lighter-weight way.

PR #6511, opened 2023-11-24
Improve error message for illegal casts and parameters.

vpetrovykh:
You cannot have a cast or a parameter of a generic "any" type.


msullivan (approved):
Sorry I thought I approved this

PR #6519, opened 2023-11-27
Expose __type__ via SQL adapter

### LINKED ISSUE #6208: Expose `__type__` column in the SQL interface
When we are querying the inhviews, we can just resolve it to `__type__`. When not accessing subtypes, we'll need to inject the uuid directly.

aljazerzen:
Closes #6208

Exposes __type__ as an UUID columns of every table, including when using `FROM ONLY` (and we are compiling SQL to load data from tables that actually *do not* have the `__type__` column).

If using SQL adapter only, values of `__type__` are useless, as there is not way to query which type is associated with this id, as we don't expose `schema::ObjectType` (or anything from `schema` module) over SQL.

TODO:
- [x] add introspection

### DIFF
--- a/edb/pgsql/resolver/relation.py
+++ b/edb/pgsql/resolver/relation.py
@@ -48,7 +48,6 @@ Context = context.ResolverContextLevel
 def resolve_SelectStmt(
     stmt: pgast.SelectStmt, *, ctx: Context
 ) -> Tuple[pgast.SelectStmt, context.Table]:
-
     # VALUES
     if stmt.values:
         values = dispatch.resolve_list(stmt.values, ctx=ctx)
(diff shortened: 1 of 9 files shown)


aljazerzen:
@msullivan this is now ready

msullivan:
> If using SQL adapter only, values of `__type__` are useless, as there is not way to query which type is associated with this id, as we don't expose `schema::ObjectType` (or anything from `schema` module) over SQL.
> 
Oh, lol. Maybe we should not have bothered with this.

edb/pgsql/resolver/relation.py
```diff
@@ -385,16 +390,26 @@ def _construct_column(p: s_pointers.Pointer, ctx: Context) -> context.Column:
...
 
-    if isinstance(p, s_links.Link):
-        col.name = p.get_shortname(ctx.schema).name + '_id'
-
-        _, dbname = pgcommon.get_backend_name(ctx.schema, p, catenate=False)
-        col.reference_as = dbname
+    elif isinstance(p, s_links.Link):
+        if short_name.name == '__type__':
+            col.name = '__type__'
+            col.reference_as = '__type__'
```
msullivan, line 399: Ugh, should we call it `__type__id`? That seems awful though

msullivan (approved):

edb/pgsql/resolver/relation.py
aljazerzen: It does look awful. It's a special-case column anyway, so we can change it into `the_id_of_the_type_of_this_object`, but I do think we can just leave it as `__type__`.

PR #6521, opened 2023-11-27
Fix non-rewritten tuple literals

aljazerzen:


msullivan (approved):
Yeah, I think probably we could refine the type hierachy to avoid this problem, yeah, by having an expression type that doesn't include TupleVarBase (I think?), and requiring that in some places?

PR #6522, opened 2023-11-27
Retry download-artifacts when it fails

### LINKED ISSUE #6429: Make CI retry failed "download shared artifacts"
This flakes a decent amount. Maybe it is possible to arrange to have it retried?
(Maybe that requires making a custom replacement action though?)

msullivan:
For now, I've only made the change when doing "Download shared
artifacts", which is the place where we have observed flakiness.

Fixes #6429.

### DIFF
--- a/.github/workflows.src/tests.inc.yml
+++ b/.github/workflows.src/tests.inc.yml
@@ -315,10 +315,12 @@
     # Restore the artifacts and environment variables
 
     - name: Download shared artifacts
-      uses: actions/download-artifact@v3
+      uses: Wandalen/wretry.action@1.3.0
       with:
-        name: shared-artifacts
-        path: .tmp
+        action: actions/download-artifact@v3
+        with:
+          name: shared-artifacts
+          path: .tmp
 
     - name: Set environment variables
       run: |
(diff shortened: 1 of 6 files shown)


.github/workflows.src/tests.inc.yml
```diff
@@ -315,10 +315,12 @@
     # Restore the artifacts and environment variables
 
     - name: Download shared artifacts
-      uses: actions/download-artifact@v3
+      uses: Wandalen/wretry.action@1.3.0
```
fantix, line 318: I think it's `v1.3.0` - but let's use hash with a tag comment like:
```suggestion
      uses: Wandalen/wretry.action@a163f62ae554a8f3cbe27b23db15b60c0ae2e93c  # v1.3.0
```

fantix (approved):

.github/workflows.src/tests.inc.yml
msullivan: Why do you prefer the hash over the tag?

msullivan: Oh, because the tag can be changed and force pushed?

fantix: Yeah I think so? Given the action providers vary a lot, using hashes would be more stable

PR #6526, opened 2023-11-28
Fix coalesced DML in FOR loops over objects

### LINKED ISSUE #6464: `for a in b union (a ?? (insert X))` doesn't work
- EdgeDB Version: https://packages.edgedb.com/archive/x86_64-unknown-linux-gnu/edgedb-server-4.1%2B2194ddc.tar.zst
- EdgeDB CLI Version: 4.1.0-dev
- OS Version: x86_64 GNU/Linux 6.6.1-arch1-1

```edgeql
with
   my_num := <int64>$my_num,
   chats := (
      select ChatBase { members: { * }, messages: { * } } 
      filter my_num in .members.num
      order by assert_single(( select .messages order by .created_at limit 1 ))
   ),
   memberships := distinct (
      for chat in chats union (
         (select ChatMembership filter .chat = chat and .user.num = my_num) ?? (insert ChatMembership { chat := chat, user := (select User filter .num = my_num) })
      )
   ),
select { 
   memberships := memberships { * }
}
```
When executing query, `memberships` is always empty and new entries are not 
Replacing `memberships := ...` with:
```
memberships := distinct (
  for chat in chats union (
     insert ChatMembership {
        chat := chat,
        user := (select User filter .num = my_num)
     } unless conflict on ((.user, .chat))
     else (select ChatMembership)
  )
),
```
make it work as intended.

Sample schema:
```
scalar type UsersNumericSequence extending sequence;
type User {
   required num: UsersNumericSequence { readonly := true; constraint exclusive; };
}
scalar type ChatsNumericSequence extending sequence;
abstract type ChatBase {
   required num: ChatsNumericSequence { readonly := true; constraint exclusive; };
   multi members: User {
      on target delete allow;
   };
   multi messages: Message {
      constraint exclusive;
      on target delete allow;
      on source delete delete target;
   }
   index on (.num);
}
type ChatMembership {
   required user: User {
   };
   required chat: ChatBase {
   };
   required last_read_message_num: int64 { default := -1; };
   constraint exclusive on ((.chat, .user));
}
```

msullivan:
More generally, the issue here is that is that because of
pull_path_namespace, the outer iterator was being routed to an
optional wrapper that has the iterator in its non-NULL case, but which
would produce NULL in the NULL branch. We already have code to prevent
this sort of thing by putting visible paths in the path_id_mask in
update_scope (see #5575), but it was a little too limited.

I do still think that we probably should be doing this masking
unconditionally, but it causes lots of trouble with materialization
cases. I should follow up on that.

I also need to spend some time deciding whether the entire
pull_path_namespace system actually makes any sense at all
or whether it is just a hack we should hope to eliminate.

Fixes #6464

### DIFF
--- a/edb/pgsql/compiler/relctx.py
+++ b/edb/pgsql/compiler/relctx.py
@@ -818,7 +818,7 @@ def update_scope(
     # with some DML linkprop cases (probably easy to fix) and a number
     # of materialization cases (possibly hard to fix), so I'm going
     # with a more conservative approach.
-    if scope_tree.optional:
+    if scope_tree.is_optional(ir_set.path_id):
         # Since compilation is done, anything visible to us *will* be
         # up on the spine. Anything tucked away under a node must have
         # been pulled up.
--- a/tests/test_edgeql_for.py
+++ b/tests/test_edgeql_for.py
@@ -1107,6 +1107,25 @@ class TestEdgeQLFor(tb.QueryTestCase):
             [{"key": "Earth"}, {"key": "Water"}]
         )
 
+    async def test_edgeql_for_tuple_optional_01(self):
+        await self.assert_query_result(
+            r'''
+                for user in User union (
+                  ((select (1,) filter false) ?? (2,)).0
+                );
+            ''',
+            [2, 2, 2, 2],
+        )
+
+        await self.assert_query_result(
+            r'''
+                for user in User union (
+                  ((select (1,) filter user.name = 'Alice') ?? (2,)).0
+                );
+            ''',
+            tb.bag([1, 2, 2, 2]),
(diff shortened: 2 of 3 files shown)


msullivan:
This can be backported to 3.x, though the changelog message would need to change and some of the tests would need dropped. `test_edgeql_for_tuple_optional_01` exists in 3.x, but the other manifestations are in new features

aljazerzen:
> I also need to spend some time deciding whether the entire
pull_path_namespace system actually makes any sense at all
or whether it is just a hack we should hope to eliminate.

We should talk about this.

aljazerzen (approved):

PR #6529, opened 2023-11-29
Simplify ?? implementation when RHS is AT_MOST_ONE

msullivan:
A lot of the code is dedicated to filtering out NULL results after the
COALESCE is done, but that is avoidable: if we make sure that
`nullable` is properly set on the expression, then this will be
handled automatically at higher levels of the compilation pipeline.

Once that simplification is made, the only difference from the ONE
case is that the RHS needs to be placed in a sub-SELECT, and we can
use set_as_subquery to do that. This lets us merge the two cases,
with the only difference being in how the RHS is compiled.

### DIFF
--- a/edb/pgsql/ast.py
+++ b/edb/pgsql/ast.py
@@ -105,8 +105,7 @@ class BaseExpr(Base):
         for v in kwargs.values():
             if typeutils.is_container(v):
                 items = typing.cast(typing.Iterable, v)
-                # FIXME: Is this right???
-                nullable = all(getattr(vv, 'nullable', False) for vv in items)
+                nullable = any(getattr(vv, 'nullable', False) for vv in items)
 
             elif getattr(v, 'nullable', None):
                 nullable = True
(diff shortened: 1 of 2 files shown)


edb/pgsql/ast.py
```diff
@@ -105,7 +105,8 @@ def _infer_nullability(self, kwargs: typing.Dict[str, object]) -> bool:
         for v in kwargs.values():
             if typeutils.is_container(v):
                 items = typing.cast(typing.Iterable, v)
-                nullable = any(getattr(vv, 'nullable', False) for vv in items)
+                # FIXME: Is this right???
+                nullable = all(getattr(vv, 'nullable', False) for vv in items)
```
aljazerzen, line 109: This is not true.

Counter example:
```
POWER(4, NULL)
```

`4` is not nullable, so `all(False, True)` will be `False`, but `POWER(4, NULL)` is nullable.

This should be a proper AST pass, not this kwargs hackey.

aljazerzen (approved):
Looks good. The nullable inference can be fixed in a separate PR, I might take a stab at that tomorrow.

edb/pgsql/ast.py
aljazerzen: `any` in most cases and `all` in the case of `COALESCE`

PR #6536, opened 2023-11-30
Fix test_connection_connect_timeout for macOS

fantix:
The TCP listen backlog on macOS is not for pending clients before accept() but rather concurrent TCP handshakes. Looks like not doing `listen()` blocks the client connect() properly on macOS.


elprans (approved):

aljazerzen:
@fantix, I'll also backport this

PR #6553, opened 2023-12-01
Integer/UUID to bytes conversion

mritzerfeld:
This pull requests extends the available to_bytes() function to support the conversion of integers and UUIDs into binary representation. Reverse conversion for integers is available through extended function to_int16(), to_int32() and to_int64().

Additionally type casts for such data are introduced.


tests/test_edgeql_functions.py
```diff
@@ -3116,6 +3116,57 @@ async def test_edgeql_functions_string_bytes_conversion_error(self):
                 ''',
             )
 
+    async def test_edgeql_functions_int16_bytes_conversion(self):
```
msullivan, line 3119: Could you add some more direct tests?

edb/lib/std/25-numoperators.edgeql
```diff
@@ -2680,3 +2680,47 @@ CREATE CAST FROM std::decimal TO std::str {
     SET volatility := 'Immutable';
     USING SQL CAST;
 };
+
+
+## Bytes casts.
```
msullivan, line 2685: I think my inclination is to take the functions but omit adding the casts.

Do you have an argument in favor of doing both?

msullivan (commented):
This looks good, and I think we should be able to merge something along these lines.

I have a question and a request for tests.

I am happy with the function names but I would not be surprised if some of my colleagues wish to argue about them.

edb/lib/std/25-numoperators.edgeql
mritzerfeld: My only argument for adding the casts would be consistency. So if, aside from `json`, conversion is solely done via functions it should be the same here.

docs/stdlib/bytes.rst
```diff
@@ -136,19 +148,48 @@ Bytes
...
+    .. code-block:: edgeql-repl
+
+        db> select to_bytes(<int16>31);
+        {b'\x00\x1f'}
+        db> select to_bytes(<int32>31);
+        {b'\x00\x00\x00\x1f'}
+        db> select to_bytes(123456789123456789);
+        {b'\x01\xb6\x9bK\xac\xd0_\x15'}
+
+    The UUID values are converted to the underlying 16-byte string:
```
elprans, line 182:
```suggestion
    The UUID values are converted to the underlying string of 16 bytes:
```

(otherwise one might think we are sending the string representation).

docs/stdlib/numbers.rst
```diff
@@ -814,41 +814,104 @@ from :eql:type:`str` and :eql:type:`json`.
...
+    The bytes conversion function expects big-endian representation. Only the
+    lower 2 bytes will be used to determine the value, the rest will be
+    ignored. This means that a 64-bit integer can be passed as long as the
+    actual value is contained in the lower 2 bytes.
+
+    .. code-block:: edgeql-repl
+
+        db> select to_int16(b'\x00\x07');
+        {7}
+        db> select to_int16(to_bytes(<int64>123));
```
elprans, line 843: Hm, I'm not sure that's the behavior we want.  I would expect an error, not silent truncation.

msullivan (approved):

1st1:
Endianness should be specified  explicitly (we shouldn't just assume BE). IMO we must fix this before this is shipped. @vpetrovykh @msullivan

PR #6560, opened 2023-12-04
Populate __fts_document__ after restoring a dump

### LINKED ISSUE #6550: FTS indexes are broken after a restore
`__fts_document__` doesn't get dumped, and we disable triggers when restoring, so `__fts_document__` never gets populated. We should probably run SQL that populates it after the restore

aljazerzen:
Closes #6550

There is a few things I'm not sure about:

- I'm returning `repopulate_units` in `RestoreDescriptor` as plain str.
  Should they be wrapped into `SqlUnit`?
  We have something similar for executing units that create the schema,
  then we populate the tables with COPY,
  but we didn't have anything that would execute custom SQL afterwards.

- I'm calling code for fts indexes directly from the server/compiler.
  This means that I'm I'm constructing an ad-hoc edgeql.compiler.CompilerOptions
  for indexes, and calling stuff that is very related to FTS only.
  Should I create an abstraction where schema objects can define a
  "post-restore repopulation" actions?

- Is it ok do add to an existing dump test?
  Will that break restore from v1 tests?
  Should I add a new dump test for this?

### DIFF
--- a/edb/pgsql/dbops/base.py
+++ b/edb/pgsql/dbops/base.py
@@ -421,7 +421,7 @@ class Query(Command):
         else:
             return self.text
 
-    def code(self, block: PLBlock) -> str:
+    def code(self, block: Any) -> str:
         return self.text
 
     def __repr__(self):
--- a/edb/pgsql/delta.py
+++ b/edb/pgsql/delta.py
@@ -3547,6 +3547,25 @@ def get_index_compile_options(
     )
 
 
+def get_reindex_command(
+    obj: s_objtypes.ObjectType,
+    schema: s_schema.Schema,
+    modalises: Mapping[Optional[str], str],
+) -> Optional[str]:
+    "Generate SQL statement that repopulates the index after a restore."
+    "Currently this only applies to FTS indexes."
+
+    (fts_index, _) = s_indexes.get_effective_fts_index(obj, schema)
+    if fts_index:
+        options = get_index_compile_options(
+            fts_index, schema, modalises, None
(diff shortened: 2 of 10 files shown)


msullivan:
You should add the test to `test_dump_v4.py`, which is the dump test for new 4.0 features. `dump0x` are all pre-1.0 and 1.0 dump tests (with meaningless numbers), after which we started using `test_dump_vN.py`.

Each dump test is run several times:
 * Once on just the current dev version where the database is created, populated, dumped, and restored
 * Once for each released major version created after the dump test, where we just restore a dump that is checked in. (Or more precisely, once for each version created after the test where we ran `edb gen-test-dumps` to create the test dumps.) 

You'll still get a test failure when the the existing 4.0 dump is tried (since the FTS fields aren't there), so what I'd suggest for now is to add a hack to skip those parts of the test in that case, like this: https://github.com/edgedb/edgedb/blob/master/tests/test_dump_v3.py#L48-L56

Then maybe we'll also run `gen-test-dumps` for 4.3, so we have an actual dump where this comes up? Or maybe I'll go do a hacky thing where I cherry-pick the changes to the dump test setup onto `v4.0` and then recreate the 4.0 dumps like that. (I think the existing 4.0 dumps were created similarly, with more stuff in the dumps than was actually present in the real 4.0. Oh, except that I'm not sure exactly what that was, which complicates thing...)

aljazerzen:
This comment should be pinned to a edgedb compiler knowledge database :D

PR #6562, opened 2023-12-04
Fix reloading readiness state and JWT *_list files in multi-tenant mode

fantix:
* SIGHUP will now read `readiness-state-file`, `jwt-sub-allowlist-file` and `jwt-revocation-list-file` (if set) of all tenants
* ~Set `EDGEDB_SERVER_DISABLE_MONITOR_FS=1` to disable all `monitor_fs()`~
* Added option `--reload-config-files` / env var `EDGEDB_SERVER_RELOAD_CONFIG_FILES` to configure when to reload:
  * `default`: reload on both SIGHUP and fsevent
  * `never`: disable the reload function
  * `signal`: only reload on SIGHUP
  * `fsevent`: watch the files for changes and reload when it happens

- [x] Manual test
- [ ] Under `default` / `fsevent`, also watch the JWT-related files, as well as the multitenant-config file
- [ ] Add test

Please review each commit.

### DIFF
--- a/edb/server/tenant.py
+++ b/edb/server/tenant.py
@@ -1134,6 +1134,18 @@ class Tenant(ha_base.ClusterProtocol):
 
         self._accepting_connections = self.is_online()
 
+    def reload(self):
+        # In multi-tenant mode, the file paths for the following states may be
+        # unset in a reload, while it's impossible in a regular server.
+        # Therefore, we are clearing the states here first, rather than doing
+        # so in reload_readiness_state() or load_jwcrypto().
+        self._readiness = srvargs.ReadinessState.Default
+        self._jwt_sub_allowlist = None
+        self._jwt_revocation_list = None
+
+        self.reload_readiness_state()
+        self.load_jwcrypto()
+
     async def on_before_drop_db(
         self, dbname: str, current_dbname: str
     ) -> None:
--- a/edb/server/tenant.py
+++ b/edb/server/tenant.py
@@ -78,6 +78,7 @@ class Tenant(ha_base.ClusterProtocol):
     _task_group: taskgroup.TaskGroup | None
     _tasks: Set[asyncio.Task]
     _accept_new_tasks: bool
+    _file_watch_finalizers: list[Callable[[], None]]
 
     __sys_pgcon: pgcon.PGConnection | None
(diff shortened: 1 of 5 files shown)


edb/server/tenant.py
```diff
@@ -358,13 +372,15 @@ async def init(self) -> None:
...
 
-            def reload_state_file(_file_modified, _event):
-                self.reload_readiness_state()
-
             self.reload_readiness_state()
-            self._server.monitor_fs(
-                self._readiness_state_file, reload_state_file
-            )
+
+            if not os.getenv("EDGEDB_SERVER_DISABLE_MONITOR_FS"):
```
elprans, line 377: This should be something like `EDGEDB_SERVER_CONFIG_FILE_RELOAD` and be an enum (`none`, `signal`, `fsevent`).

fantix: Yeah I heard you said that in our meeting too - TL; DR: I'm just concerned this may be a breaking change and decided to add a less intrusive "internal" env var.

Currently, we have the following types of "config files":

1. TLS certificate and key (server)
2. JWS key (server)
3. Readiness state (tenant)
4. JWT sub allowlist and revocation list (tenant)

And before this PR, the reloading condition is like:

| Config file  | Single-tenant | Multi-tenant |
|--------------|--------------------|-------------------|
| 1. TLS       | signal, fsevent    | signal, fsevent   |
| 2. JWS key   | signal             | signal            |
| 3. Readiness | signal, fsevent    | fsevent           |
| 4. JWT lists | signal             | none              |

(This PR adds the missing `signal` to 3 and 4 under multi-tenant mode and allow stripping `fsevent`.)

My concern is, if `--config-file-reload` is a single global enum (defaults to `signal`), we'll change the default dual reload of 1. TLS and 3. Readiness for regular instances. It's probably not a problem as we didn't advertise the `fsevent` reload as a stable feature, but it's been out for a while. So if you think that's okay, I can add this enum as a proper option/env var and add `fsevent` to 2. JWS key and 4. JWT lists.

elprans: The default value for the env var should be `"default"` and it should mimic the pre-PR behavior.

elprans (approved):

Issue #4223, opened 2022-08-06
Add documentation to explain benefits of single links (with backlink) vs multi link 


vonkoff:
**Think this was a great write up by elprans that should be simplified and included somewhere in the EdgeDB documentation.** 

Comment
--------------

In general, `single` links are more efficient than `multi` links, because a `multi` link involves an intermediate table, which needs to be queried when reading data and updated whenever a new post is created.

Here's a rough guide on when to use `multi` and when to use a backlink, possibly with a computed link on the other end for convenience.

Use a `multi` link if:

1. The relationship is relatively stable, i.e. you don't update it frequently, for example a list of associated postal addresses in a user profile.
2. The number of elements in the link is usually fairly small.

If either of the above isn't true, use a `single` link from the other end and couple it with a computed based on a backlink, like so:

```
type Post {
  required link author -> User;
}

type User {
  link posts := (.<author[IS Post])
}
```

_Originally posted by @elprans in https://github.com/edgedb/edgedb/discussions/3105#discussioncomment-1546892_

PR #6563, opened 2023-12-05
Add advice on multi link on linking object vs. backlink on linked-to object

Dhghomon:
Resolves https://github.com/edgedb/edgedb/issues/4223 where there is some good advice on choosing between the two. Looks like this page has just the section to stick it in.

### DIFF
--- a/docs/datamodel/links.rst
+++ b/docs/datamodel/links.rst
@@ -278,9 +278,10 @@ table <https://en.wikipedia.org/wiki/Associative_entity>`_, whereas a
 .. note::
 
   Choosing a link direction can be tricky when modeling these kinds of
-  relationships. Should you model the relationship as one-to-many using a ``multi`` link, or as
-  many-to-one using a ``single`` link with a backlink to traverse in the
-  other direction? A general rule of thumb in this case is as follows.
+  relationships. Should you model the relationship as one-to-many using a
+  ``multi`` link, or as   many-to-one using a ``single`` link with a
+  backlink to traverse in the other direction? A general rule of thumb
+  in this case is as follows.
 
   Use a ``multi`` link if:
 
--- a/docs/datamodel/links.rst
+++ b/docs/datamodel/links.rst
@@ -287,8 +287,8 @@ Use a ``multi`` link if:
   For example, a list of postal addresses in a user profile.
 - The number of elements in the link tends to be small.
 
-Otherwise, prefer a single link from one object coupled with a computed backlink
-on the other:
+Otherwise, prefer a single link from one object coupled with a computed
+backlink on the other:
 
 .. code-block:: sdl
 
--- a/docs/datamodel/links.rst
+++ b/docs/datamodel/links.rst
@@ -277,6 +277,29 @@ table <https://en.wikipedia.org/wiki/Associative_entity>`_, whereas a
 As a result, single links are marginally more efficient. Generally ``single``
 links are recommended when modeling 1:N relations.
 
+Another decision often faced by EdgeDB users is whether to use a ``multi``
+link on the linking object or a backlink on the other (the linked object).
+A general rule of thumb in this case is as follows.
+
+Use a ``multi`` link if:
+
+- The relationship is relatively stable and thus not updated very frequently.
+  For example, a list of postal addresses in a user profile.
+- The number of elements in the link tends to be small.
+
+Otherwise, prefer a single link from one object coupled with a computed backlink
+on the other:
+
+.. code-block:: sdl
+
+    type Post {
+      required author: User;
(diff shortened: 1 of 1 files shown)


docs/datamodel/links.rst
```diff
@@ -277,6 +277,29 @@ table <https://en.wikipedia.org/wiki/Associative_entity>`_, whereas a
...
+A general rule of thumb in this case is as follows.
+
+Use a ``multi`` link if:
+
+- The relationship is relatively stable and thus not updated very frequently.
+  For example, a list of postal addresses in a user profile.
+- The number of elements in the link tends to be small.
+
+Otherwise, prefer a single link from one object coupled with a computed
+backlink on the other:
```
scotttrinh, line 291: I think we can fold the prior sentence from this change into this same paragraph. Otherwise it's weird that we say "Generally ``single`` links are recommended when modeling 1:N relations." But then go on to caveat it further.

What about:

```suggestion
Otherwise, prefer a single link from one object coupled with a computed
backlink on the other. This is marginally more efficient and generally
recommended when modeling 1:N relations:
```

And remove that sentence from lines 277:278?

Dhghomon: Ah yes, I was thinking about that a bit and looks like a good solution to the small duplication.

Dhghomon: Okay, deleted the other sentence now. Looks nice

docs/datamodel/links.rst
```diff
@@ -274,8 +274,30 @@ same "shirt owner" relationship is represented with a ``multi`` link.
...
+
+.. code-block:: sdl
+
+    type Post {
+      required author: User;
+    }
+
+    type User {
+      link posts := (.<author[is Post])
+    }
```
raddevon, line 300: I might do this whole block as a `.. note::`, just to call out that it isn't really part of the "one-to-many" section. It's related to both that and the "many-to-one" section since it's advising between the two.

Also, maybe we should characterize this as the "direction" of the link. I think that's a bit clearer than the way you've characterized it here. Spitballing a rewrite of the first paragraph:
> Choosing a link direction can be tricky when dealing with these kinds of links. Should you model the link as a one-to-many using `multi`, or should you model a many-to-one (single) link and use backlinks to traverse in the other direction? A general rule of thumb in this case is as follows.

Does this approach make sense?

Oh, one more:
> Otherwise, prefer a single link from one object **type** coupled with a computed backlink on the other.

raddevon (changes requested):
Very good change! I made a couple of suggestions to hopefully clarify some things. To summarize:
- Put this in a `.. note::`
- Characterize as the "direction" of the link
- Since we're talking about modeling, talk about a link from an object _type_ rather than from an object. (Becoming less convinced of the importance of this one as I try to describe it, so feel free to push back if you think it's wrong or insignificant. 😅)

docs/datamodel/links.rst
```diff
@@ -274,8 +274,34 @@ same "shirt owner" relationship is represented with a ``multi`` link.
...
 ``single`` link is stored as a column in the object type where it is declared.
-As a result, single links are marginally more efficient. Generally ``single``
-links are recommended when modeling 1:N relations.
+
+.. note::
+
+  Choosing a link direction can be tricky when dealing with these kinds of
+  links. Should you model the link as a one-to-many using ``multi``, or should
+  you model a many-to-one (single) link and use backlinks to traverse in the
+  other direction? A general rule of thumb in this case is as follows.
```
scotttrinh, line 283: A few small suggestions for clarity. Mostly trying to tease out how _relationships_ are modeled _using_ links. In other words, you don't "model links" you "model relationships (one-to-many, or many-to-one) with links (multi or single)".

```suggestion
  Choosing a link direction can be tricky when modeling with these kinds of
  relationships. Should you model the relationship as one-to-many using a ``multi`` link, or as
  many-to-one using a ``single`` link with a backlink to traverse in the
  other direction? A general rule of thumb in this case is as follows.
```

—Note that I didn't reformat the line lengths to highlight the changes here, so this suggestion isn't something that can be applied directly.

scotttrinh: @Dhghomon did you see this suggestion? any thoughts?

Dhghomon: Weird, was this hidden? I'll commit it now.

scotttrinh: @Dhghomon Not sure! But, as I noted, the formatting is not quite right, I just edited it inline to highlight the suggested changes, so you'll need to format the line lengths (looking forward to not caring about this soon!)

scotttrinh (approved):

PR #6565, opened 2023-12-05
Fix changing a pointer to be computed when there is a subtype

### LINKED ISSUE #6542: `cannot drop columns from view` on migrating link->computed with a subtype
- EdgeDB Version: 4.2+833dd8d
- EdgeDB CLI Version: 4.0.2+500be79
- OS Version: macos 13

Steps to Reproduce:

1. Perform the following migrations
```
CREATE MIGRATION m1dtiamxhlhyqgliwi763otwsfwyxyky6zepk7le4kyy4adqzmlmna
    ONTO initial
{
  CREATE TYPE default::Random;
  CREATE TYPE default::Parent {
      CREATE REQUIRED SINGLE LINK r: default::Random;
  };
  CREATE TYPE default::Child EXTENDING default::Parent;
};
```

```
CREATE MIGRATION m1fscwyfelzqkqtzx6tcbvuwggj44btex53now4in45jtjga7oovaq
    ONTO m1dtiamxhlhyqgliwi763otwsfwyxyky6zepk7le4kyy4adqzmlmna
{
  ALTER TYPE default::Parent {
      ALTER LINK r {
          USING (std::assert_exists((SELECT
              default::Random
          LIMIT
              1
          )));
      };
  };
};
```

```
edgedb error: InternalServerError: cannot drop columns from view
  Server traceback:
      edb.errors.InternalServerError: cannot drop columns from view
```

Schema:
```
module default {

    type Random {}

    type Parent {
        # required single link r: Random;
        link r:=(select Random);
    }

    type Child extending Parent {
    }

}
```

msullivan:
This was a regression introduced by #6542, which moved the deletion of
the pointer to happen *after* processing subcommands. It turns out
that that was correct, but in order for it to work right around
inheritance, we need to make sure that the propagation of alters into
child types occurs as "caused" actions, so that they happen after all
the processing for the parent objects themselves. (Including the
actual deletion of a pointer when it is changed to computed.)

Also, make sure that constraints get propagated when this change
happens.  That change then required a knock-on change to expression
propagation to avoid messing up error messages in that case.

Fixes #6542

### DIFF
--- a/edb/schema/delta.py
+++ b/edb/schema/delta.py
@@ -2123,7 +2123,9 @@ class ObjectCommand(Command, Generic[so.Object_T]):
                 delta_drop, cmd_drop, _ = ref.init_delta_branch(
                     schema, context, cmdtype=AlterObject)
                 delta_create, cmd_create, ctx_stack = ref.init_delta_branch(
-                    schema, context, cmdtype=AlterObject)
+                    schema, context, cmdtype=AlterObject,
+                    possible_parent=self,  # type: ignore
+                )
 
                 cmd_drop.from_expr_propagation = True
                 cmd_create.from_expr_propagation = True
--- a/edb/schema/inheriting.py
+++ b/edb/schema/inheriting.py
@@ -955,7 +955,7 @@ class AlterInheritingObjectOrFragment(
                 schema = d_alter_cmd.inherit_fields(
                     schema, context, d_bases, fields=props)
 
-            self.add(d_root_cmd)
+            self.add_caused(d_root_cmd)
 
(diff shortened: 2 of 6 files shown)


elprans:
> This was a regression introduced by #6542

You meant some other PR here, not the reported issue?

msullivan:
Sorry, introduced by #5411

aljazerzen:
What are "caused" actions in delta commands?

msullivan:
> What are "caused" actions in delta commands?

They run *after* the command is finished

aljazerzen (approved):

PR #6566, opened 2023-12-05
Fix migrating between aliased and non-aliased computeds

msullivan:
This version is designed to be cherry-pickable to 4.x, so it doesn't
change how any of our schemas are represented, it just tries to
actually fully make the transition when we switch.

I'm going to do a follow-up where I rip all of this out and the cost
of schema changes.

### DIFF
--- a/edb/schema/pointers.py
+++ b/edb/schema/pointers.py
@@ -1191,11 +1191,36 @@ class PointerCommandOrFragment(
             self.set_attribute_value(
                 'is_derived', True
             )
+            if isinstance(self, inheriting.AlterInheritingObject):
+                self._propagate_is_derived(schema, context, True)
 
             if context.declarative:
                 self.set_attribute_value(
                     'declared_overloaded', True
                 )
+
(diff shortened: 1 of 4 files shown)


edb/schema/pointers.py
```diff
@@ -1191,12 +1191,37 @@ def canonicalize_attributes(
...
 
+        # If we *were* an aliased computed but are changing to not be
+        # one, we need to update our base and is_derived to not be.
+        elif (
+            isinstance(self, inheriting.AlterInheritingObject)
+            and self.has_attribute_value('expr')
+            and (
+                (cur_base := self.scls.get_bases(schema).objects(schema)[0])
+                and (subj := cur_base.get_subject(schema))
+                and subj == self.scls.get_subject(schema)
```
aljazerzen, line 1210: This checks if my subject is also the subject of my first base.

So it checks that I'm haven't been rebased to a different subject?

Why is this here?

msullivan:
@elprans Could you take a look at this?

Issue #6389, opened 2023-10-30
Document Access Policy Annotations


CarsonF:
The docs are misleading right now about where annotations can be declared with access policies in the schema.
It took me awhile to figure out they are supposed to be inside the `{}` & a sibling of the `errmessage` set clause.
There were no examples of this, so I was just using the synopsis block.
```diff
 access policy name
   [ when (condition) ]
   { allow | deny } action [, action ... ]
   [ using (expr) ]
-  [ annotation-declarations ]
   [ "{"
      [ errmessage := value ; ]
+     [ annotation-declarations ]
     "}" ] ;
```
```diff
 [ with with-item [, ...] ]
 { create | alter } type TypeName "{"
   [ ... ]
   create access policy name
     [ when (condition) ; ]
     { allow | deny } action [, action ... ; ]
     [ using (expr) ; ]
-     [ create annotation annotation-name := value ; ]
     [ "{"
        [ set errmessage := value ; ]
+       [ create annotation annotation-name := value ; ]
       "}" ; ]
 "}"
```

Dhghomon:
Good catch! I'll put a PR together fixing this along with adding an annotation to the example on the access policy page.

PR #6573, opened 2023-12-06
Doc: access policy fix and expansion

Dhghomon:
Fixes https://github.com/edgedb/edgedb/issues/6389 where a user noticed that the syntax declaration for access policies makes it look like annotations should be outside the braces when they should be inside, and that we don't have any access policy examples that show where annotations belong either so it took some experimentation to figure it out.

In our current example there is a bit of new learning to do so instead of declaring a new annotation at the same time, it uses the built-in `errmessage` to show where to add such notes and then mentions in the breakdown below that other annotations can go in this space as well. Plus adds an example where you might see this error: where an app is configured wrongly such that a user is allowed to *try* to write a post but is prevented by the access policy.


raddevon (changes requested):
Great additions, @Dhghomon! I have some minor suggestions and one concern about an access policy result that I don't follow.

docs/datamodel/access_policies.rst
```diff
@@ -253,7 +261,36 @@ we'll see the post we just created.
...
+.. code-block:: edgeql-repl
+
+  db> set global current_user := <uuid>'d1c64b84-8e3c-11ee-86f0-d7ddecf3e9bd';
+  OK: SET GLOBAL
+  db> insert BlogPost {
+  ...    title := "My post",
+  ...    author := (select User filter .id = global current_user)
+  ...  };
+  edgedb error: AccessPolicyError: access policy violation on insert of
+  default::BlogPost (User must have same ID as blog author)
```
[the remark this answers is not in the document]
Dhghomon: Yeah, not a good example as we discussed last week. A better idea struck me which requires two globals but still simple enough and feels much more like a real world example.

docs/datamodel/access_policies.rst
```diff
@@ -205,28 +247,41 @@ Let's add a policy to our sample schema.
...
-one else will be able to edit, delete, or view this post.
+Let's break down the access policy syntax piece-by-piece. These policies grant
+full read-write access (``all``) to the ``author`` of each ``BlogPost``, if
+the author is in a country that allows full access to the service. Otherwise,
+the same author will be restricted to either read-only access or no access at
+all, depending on the country.
 
 .. note::
 
   We're using the *coalescing equality* operator ``?=`` which returns
```
raddevon, line 268: Not your change, I know, but while we're here… What do you think?
```suggestion
  We're using the *coalescing equality* operator ``?=`` because it returns
```
This makes it sound less like a side effect and more like the intended effect (which I assume it was).

docs/datamodel/access_policies.rst
```diff
@@ -205,28 +247,41 @@ Let's add a policy to our sample schema.
...
   ``false`` even if one of its arguments is an empty set.
 
-- ``access policy``: The keyword used to declare a policy inside an object
+- ``access policy`` The keyword used to declare a policy inside an object
```
raddevon, line 271: I'd vote for restoring the colon here. Maybe it was unintentional?
```suggestion
- ``access policy``: The keyword used to declare a policy inside an object
```

docs/datamodel/access_policies.rst
```diff
@@ -205,28 +247,41 @@ Let's add a policy to our sample schema.
...
   type.
-- ``author_has_full_access``: The name of this policy; could be any string.
+- ``author_has_full_access`` and ``author_has_read_access``: The name of these
```
raddevon, line 273:
```suggestion
- ``author_has_full_access`` and ``author_has_read_access``: The names of these
```

docs/datamodel/access_policies.rst
```diff
@@ -205,28 +247,41 @@ Let's add a policy to our sample schema.
...
+  policies; could be any string.
 - ``allow``: The kind of policy; could be ``allow`` or ``deny``
 - ``all``: The set of operations being allowed/denied; a comma-separated list
   of the following: ``all``, ``select``, ``insert``, ``delete``, ``update``,
   ``update read``, ``update write``.
```
raddevon, line 278: Making sure it's clear the user doesn't have to list all of these.
```suggestion
  of any number of the following: ``all``, ``select``, ``insert``, ``delete``,
  ``update``, ``update read``, and ``update write``.
```

docs/datamodel/access_policies.rst
```diff
@@ -205,28 +247,41 @@ Let's add a policy to our sample schema.
 
     +     access policy author_has_full_access
     +       allow all
-    +       using (global current_user ?= .author.id);
+    +       using (global current_user    ?= .author.id
+    +         and  global current_country ?= Country.Full) {
+    +        errmessage := "User not in country with full access";
```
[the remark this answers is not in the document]
raddevon: If you do change this, make sure you update the REPL output too.

raddevon (changes requested):
Looks good. A few minor changes, and this is ready to go!

docs/datamodel/access_policies.rst
```diff
@@ -37,29 +37,48 @@ Let's start with a simple schema without any access policies.
...
+    access policy! See the subsection on resolution order below for details.
 
 Defining a global
 ^^^^^^^^^^^^^^^^^
 
-To start, we'll add a *global variable* to our schema. We'll use this global
-to represent the identity of the user executing the query.
+Global variables are the a convenient way to provide the context needed to
+determine what sort of access should be allowed for a given object, as they
+can be set and reset as needed.
```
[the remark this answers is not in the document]
Dhghomon: I think so!

docs/datamodel/access_policies.rst
Dhghomon: Yes, I think I killed that comma unintentionally -_-

Dhghomon:
> Looks good. A few minor changes, and this is ready to go!

Okay, I think I got them all!

Dhghomon:
@raddevon Forgot all about this PR! In the meantime a small merge conflict popped up but has now been resolved.

Dhghomon:
Okay, just added https://github.com/edgedb/edgedb/pull/6573/commits/f005df87bfee937cb942a9fe106177d8cdc6b821 on which operations return empty sets vs. which will show errors. Adding this is pretty nice as it allows a more thorough explanation on `update read` vs. `update write` in the section in which they are introduced.

raddevon (approved):

PR #6579, opened 2023-12-07
Some (mostly micro) optimizations worth about ~52% on an update test

msullivan:


elprans:
Is it still 38% after the rewrite_anchors revert?

elprans (approved):

msullivan:
> Is it still 38% after the rewrite_anchors revert?

It's a few percent worse, but I think I've fixed it and have restored it

msullivan:
In particular, this is on a benchmark that is testing an update of a large type in a large schema. Our test suite does not get a particular dramatic speedup, for example

edb/schema/schema.py
```diff
@@ -1331,6 +1331,10 @@ def _get_by_id(
             # Avoid the overhead of cast(Object_T) below
             return obj  # type: ignore
 
+    # Important micro-optimization
+    if not TYPE_CHECKING:
+        get_by_id = _get_by_id
```
aljazerzen, line 1340: This seems very hacky. Would overloaded functions perform any better here?

edb/schema/objects.py
```diff
@@ -2688,11 +2703,15 @@ def get(
         name: Key_T,
         default: Optional[Object_T | NoDefaultT] = NoDefault,
     ) -> Optional[Object_T]:
-        items = dict(self.items(schema))
+        # TODO: Should we store an actual dict?
```
aljazerzen, line 2730: Somewhere, we are probably iterating over names and calling `x.get(name)`, so this would be quadratic.

If I understand correctly, this `ObjectIndexBase` is used for example, to define pointers of an object or types in a module. Number of keys can be >50, right?

So I'd say yes, this should be dict.

aljazerzen (approved):
This is great!

Some of them are quite obvious, but some require arcane levels of Python and edgedb knowledge :D

edb/schema/schema.py
msullivan: What's going on here is that `get_by_id` is defined overloaded in the parent, to handle all the type stuff, but then it calls `_get_by_id` which is defined in the child types. The point was to avoid duplicating the big list of overloads for all three types. But it turns out that `get_by_id` was hot enough that eliminating the extra call was noticiable.

PR #6583, opened 2023-12-08
fts restore

aljazerzen:
- Populate __fts_document__ after restoring a dump
- move the dump test to dumpv4
- checkpoint
- fix
- fix
- make aliased_views contain sets, not types


msullivan:
I think I like this but I think you didn't update any of the types, which is the first thing I wanted to look at

tests/test_edgeql_expressions.py
```diff
@@ -3058,7 +3058,7 @@ async def test_edgeql_expr_introspect_bad_01(self):
     async def test_edgeql_expr_introspect_bad_02(self):
         with self.assertRaisesRegex(
                 edgedb.QueryError,
-                r'cannot introspect transient type variant'):
+                r"type 'A' does not exist"):
```
aljazerzen, line 3061: I think this an improvement to the error message, because a "transient type variant" is not something that we document anywhere.

Saying that "type 'A' does not exist" hints that "A :=" does not declare a type, but just a value binding. So you have to do `INTROSPECT TYPEOF A` instead.

msullivan (approved):

PR #6589, opened 2023-12-11
Persistent cache

fantix:
Closes #6349 

- [x] ~Add LRUMapping.gc()~ we can use `stmt_cache.pyx`
- [x] Pass over the user schema version in DB
- [x] Replace `dbver` with the non-local schema version
- [x] Create cache registry table
- [x] Rewrite QueryUnits to use SQL functions
- [x] Implement persistent cache
- [x] Implement cache eviction
- [ ] Adopt #6349
- [ ] Add tests


edb/server/dbview/dbview.pyx
```diff
@@ -535,7 +535,13 @@ cdef class DatabaseConnectionView:
             raise errors.InternalServerError(
                 'no need to serialize state while in transaction')
 
-        ver = self._db.schema_version
+        ver = hash(
```
1st1, line 538: Will `ver` ever end up in Postgres or in another Python process? `hash()` as you know is randomized

fantix: Negative. In Postgres, it's in the state temporary table and never used.

fantix:
Close in favor of #6881 and many others.

PR #6590, opened 2023-12-11
Fix some bugs involving union and coalescing of optional values

### LINKED ISSUE #6438: Coalesce DML express return more than one Element
- EdgeDB Version: v4 on Cloud

1.defind schema
```
module default {
    type Foo {
	title: str;
        bar : Bar;
    }

    type Bar {
	title: str; 
    }

}
```
2.insert Foo
```
insert Foo {title:="Foo1"}
Here is the return value.
[
  {
    "id": "64682096-7d0a-11ee-8549-df022b184faa"
  }
]
```
3. runing following edgedql
```
with data := <json>$data, 
upserts := (
  for item in json_array_unpack(data["bar"]) union (
    (update Bar filter .id = <uuid>json_get(item, "barId") set { title:=<str>json_get(item,"barTitle") })??
    (insert Bar { title:=<str>json_get(item,"barTitle") })
  )), 
update Foo filter .id = <uuid>data["fooId"] set {bar:=assert_single(upserts)}

Here is variable

{"fooId":"64682096-7d0a-11ee-8549-df022b184faa","bar":[{"barId":"1735ff7b-0a7c-40fa-9ac3-e5f1219ca9b7","barTitle":"bar1"}]}
** Pls note that the Bar of id "1735ff7b-0a7c-40fa-9ac3-e5f1219ca9b7" is not yet save in database yet, it was randomly provided by the customer
```

4. by runing above, it will return a error
```
CardinalityViolationError: assert_single violation: more than one element returned by an expression
```

let's simply edgedb to show more exactly by removing array.

modify edgedb to 
```
with data := <json>$data, 
upserts := (
 
    (update Bar filter .id = <uuid>json_get(data, "bar", "barId") set { title:=<str>json_get(data,"bar","barTitle") })??
    (insert Bar { title:=<str>json_get(data, "bar","barTitle") })
), 
update Foo filter .id = <uuid>data["fooId"] set {bar:=assert_single(upserts)}

here is data

{"fooId":"64682096-7d0a-11ee-8549-df022b184faa","bar":{"barId":"1735ff7b-0a7c-40fa-9ac3-e5f1219ca9b7","barTitle":"bar2"}}
```

Here is error
```
InternalServerError: more than one row returned by a subquery used as an expression
```

While I run 
```
with data := <json>$data, 
upserts := (
 
    (update Bar filter .id = <uuid>json_get(data, "bar", "barId") set { title:=<str>json_get(data,"bar","barTitle") })??
    (insert Bar { title:=<str>json_get(data, "bar","barTitle") })
), 
select upserts

here is data

{"fooId":"64682096-7d0a-11ee-8549-df022b184faa","bar":{"barId":"1735ff7b-0a7c-40fa-9ac3-e5f1219ca9b7","barTitle":"bar2"}}
```
It did return only one element.
image

If I change id of Bar to one of id has existed in database, it will run correctly without this bug.
image

msullivan:
There are a number of places that improperly handle a set that has
multiple NULL values in it:
 * INSERT on single values
 * UPDATE on single values
 * assert_single

The assert_single case is a long standing bug that can be triggered
without much trouble. The INSERT/UPDATE bugs are new, and stem from
the new ?? on DML, since I think it that was the first thing in the
compiler backend where extra NULLS might be produced for something
with single cardinality.

Fix this by adding null checks at the consumer sites. One of the
consumer sites is set_as_subquery, and I modified update to always use
that.
I also added a helper function to add the null checks, and I'll be
going back and changing lots of other code to use the helper.

We could also flip this around a bit and try saying that we shouldn't
ever produce extra NULLs in a single set, and try to stamp it out on
the producer side, though that seems a little less consistent to me?

Fixes #6438.

### DIFF
--- a/edb/edgeql/compiler/expr.py
+++ b/edb/edgeql/compiler/expr.py
@@ -393,7 +393,8 @@ def _compile_dml_coalesce(
         res = dispatch.compile(full, ctx=subctx)
         # Indicate that the original ?? code should determine the
         # cardinality/multiplicity.
-        res.card_inference_override = ir
+        assert isinstance(res.expr, irast.SelectStmt)
+        res.expr.card_inference_override = ir
 
         return res
 
@@ -479,7 +480,8 @@ def _compile_dml_ifelse(
         res = dispatch.compile(full, ctx=subctx)
         # Indicate that the original IF/ELSE code should determine the
         # cardinality/multiplicity.
-        res.card_inference_override = ir
+        assert isinstance(res.expr, irast.SelectStmt)
+        res.expr.card_inference_override = ir
 
         return res
 
--- a/edb/edgeql/compiler/inference/cardinality.py
+++ b/edb/edgeql/compiler/inference/cardinality.py
@@ -519,12 +519,6 @@ def _infer_set(
             ir, is_mutation=is_mutation,
             scope_tree=scope_tree, ctx=ctx)
 
-        # But actually! Check if it is overridden
-        if ir.card_inference_override:
-            result = _infer_set_inner(
-                ir.card_inference_override, is_mutation=is_mutation,
-                scope_tree=scope_tree, ctx=ctx)
-
         # We need to cache the main result before doing the shape,
         # since sometimes the shape will refer to the enclosing set.
         ctx.inferred_cardinality[ir] = result
@@ -1272,6 +1266,11 @@ def __infer_select_stmt(
 
     if ir.iterator_stmt:
         stmt_card = cartesian_cardinality((stmt_card, iter_card))
+
+    # But actually! Check if it is overridden
+    if ir.card_inference_override:
+        stmt_card = infer_cardinality(
+            ir.card_inference_override, scope_tree=scope_tree, ctx=ctx)
 
     return stmt_card
 
--- a/edb/edgeql/compiler/inference/multiplicity.py
+++ b/edb/edgeql/compiler/inference/multiplicity.py
@@ -227,11 +227,6 @@ def _infer_set(
     result = _infer_set_inner(
         ir, is_mutation=is_mutation, scope_tree=scope_tree, ctx=ctx
     )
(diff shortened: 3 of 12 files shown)


elprans:
> We could also flip this around a bit and try saying that we shouldn't
> ever produce extra NULLs in a single set, and try to stamp it out on
> the producer side, though that seems a little less consistent to me?

Why? I have the opposite feeling that stamping out all callers could be a more error-prone whack-a-mole?

msullivan:
> > We could also flip this around a bit and try saying that we shouldn't
> > ever produce extra NULLs in a single set, and try to stamp it out on
> > the producer side, though that seems a little less consistent to me?
> 
> Why? I have the opposite feeling that stamping out all callers could be a more error-prone whack-a-mole?

There is sort of a general question of "when is having NULLs output by a producer permitted" that needs to be answered.

The answer can't be "never", because at the very least we have our optional wrappers, but more generally we want some operations to always produce NULL and for us to be able to elide the optional wrapper in that case. (Object property references, parameters, json casts are current operations where this happens).

1. In this PR, I took the approach of "whenever the set could be empty", since it seemed an easy and consistent rule. But you're right that it might be whack-a-mole.
2. Another possibility is "also whenever the set could be empty, but single sets must actually output at most one row whether it is NULL or not". I think I had it in my head that this is a less consistent rule than option 1, but honestly it probably makes more sense. I think this was obeyed by everything until the conditional DML work added constructs that were implemented with unions but might have their cardinality overridden to be single.
3. Or "no extra NULLs ever, even for multi sets". This is plausible, but would require inserting NULL checks in multi operations in cases where we could *not* want to insert NULL checks for single operations: We don't want a NULL check for `<str><json>$0` but this rule would require a NULL check in `<str>{<json>$0, <json>$1}`

Alright I have convinced myself that option 2 is the best.
I'll revise the PR.

aljazerzen (approved):

PR #6593, opened 2023-12-12
Test option to re-run negative cases from last test result

aljazerzen:
Based on #6592 

This PR:
- adds `--log-result` and `--no-log-result` flags to `edb test` that save the test results into a JSON file in `build/test-results/` (default: true)
- moves code for rendering results from `runner.py` into `results.py`,
- adds `--include-unsuccessful-result` flag that finds the last result and includes all tests with negative outcome (default: false)


edb/tools/test/__init__.py
```diff
@@ -89,20 +90,45 @@
                    '(e.g --cov edb.common --cov edb.server)')
 @click.option('--running-times-log', 'running_times_log_file',
               type=click.File('a+'), metavar='FILEPATH',
-              help='Maintain a running time log file at FILEPATH')
+              help='maintain a running time log file at FILEPATH')
+@click.option('--log-result/--no-log-result', is_flag=True,
+              help='write the test result to a log file',
+              default=True)
+@click.option('--include-unsuccessful-result', is_flag=True,
```
msullivan, line 97: Just `--include-unsuccesful`, I suppose

edb/tools/test/results.py
```diff
@@ -0,0 +1,298 @@
...
+        return []
+    return [
+        case.name
+        for case in result.failures
+        + result.errors
+        + result.unexpected_successes
+    ]
+
+
+def _dataclass_from_dict(klass: typing.Type, d: typing.Dict):
```
msullivan, line 285: `cls` is more traditional in python.

edb/tools/test/results.py
```diff
@@ -0,0 +1,298 @@
...
+    # JSON -> dataclasses is surprisingly hard.
+    # Maybe we could use a library here.
+
+    # HACK: hardcoded List[TestCase]
```
msullivan, line 289: You should be able to use our ersatz `typing_inspect` to extract the argument from the list type with `get_args`

edb/tools/test/results.py
```diff
@@ -0,0 +1,298 @@
...
+    if isinstance(d, list):
+        return [_dataclass_from_dict(TestCase, e) for e in d]
+
+    try:
```
msullivan, line 293: Might be better to check with `dataclasses.is_dataclass` instead of catching an exception

edb/tools/test/results.py
```diff
@@ -0,0 +1,298 @@
...
+        field_types = {f.name: f.type for f in dataclasses.fields(klass)}
+    except BaseException:
```
msullivan, line 295: You almost never want to handle BaseException. Importantly, it doesn't include `KeyboardInterrupt`.

msullivan (approved):
Basically all of my comments are about the json deserialization.

This will be great.

PR #6594, opened 2023-12-12
Document `@source`/`@target`

raddevon:


elprans (approved):

msullivan:
Could we have a @source example too? I suspect it will be a more common
pattern. Maybe something like ***@***.***, @order) for something where the
link is ordered

raddevon:
> Could we have a @source example too? I suspect it will be a more common pattern. Maybe something like ***@***.***, @order) for something where the link is ordered

I think this request got mangled by either an email client or GitHub, but I believe the intent is to demo constraint on `((@source, @order))`. I tried to do that, and it doesn't seem to be working for me or else I've missed something. I'm modeling a Simon-esque memory game with this:

```
module default {
  type Player {
    required name: str;
    multi picks: Color {
      order: int16;
      constraint exclusive on ((@source, @order));
    }
  }
  type Color {
    required name: str;
  }
}
```

When I try to migrate this, I get
```
error: cannot reference multiple links or properties in a constraint where at least one link or property is MULTI
  ┌─ /Users/raddevon/Documents/projects/edgedb/demos/multi-link-order/dbschema/default.esdl:6:7
  │
6 │       constraint exclusive on ((@source, @order));
  │       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ error

```

Am I missing something here, @msullivan?

msullivan:
Oh, I don't think this will work in 4.x until 4.3 goes out. It should work on nightly though?
(Unless we haven't had nightlies go out in a week?)

PR #6595, opened 2023-12-12
Document comparing objects by casting to `json`

raddevon:


raddevon:
@aljazerzen Since it seemed like this was useful in a very narrow use case, I've documented it only where users are likely to run into that use case instead of trying to put it in something broader like casting documentation, JSON documentation, or comparison operator documentation. Let me know how you feel about that.

aljazerzen (approved):
This is great, thanks!

PR #6596, opened 2023-12-12
Make trigger `when` docs visible only for 4.0

### LINKED ISSUE #6416: Document 4.0 trigger changes
Looks like `when` was added to the syntax on the [DDL](https://www.edgedb.com/docs/reference/ddl/triggers#triggers)/[SDL](https://www.edgedb.com/docs/reference/sdl/triggers#ref-eql-sdl-triggers-syntax) reference pages, but there are a few problems:

- `when` is visible in the syntax even when I switch back to v3 in the docs. If this had been backported to v3, the addition of `when` wouldn't be in the 4.0 changelog, so I think it's incorrect to show it in the syntax when switching the version selector to v3.
- The function of `when` is not described in the "Description" on [the SDL reference page](https://www.edgedb.com/docs/reference/sdl/triggers#ref-eql-sdl-triggers-syntax).
- Might be good to add a `when` to the example on these pages (although only for v4)
- We probably need mention and maybe an example of `when` on [the Schema > Triggers page](https://www.edgedb.com/docs/datamodel/triggers#triggers) (behind a `versionadded:: 4.0` directive).

raddevon:
`when` was added to the synopses, but it was visible regardless of the version selected. This hides `when` for versions lower than 4.0. Also adds an example of `when` usage.

https://github.com/edgedb/edgedb/pull/6595 will add a `when` usage example to the `datamodel/triggers` documentation page, so I have omitted that here.

Closes #6416

### DIFF
--- a/docs/reference/sdl/triggers.rst
+++ b/docs/reference/sdl/triggers.rst
@@ -28,6 +28,27 @@ Declare a trigger that inserts a ``Log`` object for each new ``User`` object:
       );
     }
 
+..versionadded:: 4.0
+
+Declare a trigger that inserts a ``Log`` object conditionally when an update
+query makes a change to a ``User`` object:
+
+.. code-block:: sdl
+
+    type User {
+      required name: str;
+
+      trigger log_update after update for each
+      when (<json>__old__ {**} != <json>__new__ {**})
+      do (
+        insert Log {
+          action := 'update',
+          target_name := __new__.name,
+          change := __old__.name ++ '->' ++ __new__.name
+        }
+      );
+    }
+
 .. _ref_eql_sdl_triggers_syntax:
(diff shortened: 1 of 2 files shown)


docs/reference/sdl/triggers.rst
```diff
@@ -28,6 +28,27 @@ Declare a trigger that inserts a ``Log`` object for each new ``User`` object:
       );
     }
 
+..versionadded:: 4.0
```
scotttrinh, line 31:
```suggestion
.. versionadded:: 4.0
```

scotttrinh (approved):
One little typo, but looks good after fixing that!

aljazerzen (approved):

PR #6601, opened 2023-12-13
Avoid doing a join when injecting type ids

### LINKED ISSUE #6599: Don't join against __type__ when injecting type ids
See https://github.com/edgedb/edgedb/commit/006308ffe2006b38c08163e1db373abda77f0ac0, which broke it

msullivan:
We have an optimization to allow skipping joins when all we want is
the id of a link, but it is suppressed if there is an access policy on
the target. Unsuppress it when the target is schema::ObjectType, since
there shouldn't be any user-visible links to an internal ObjectType
anyway, and it lets us avoid a join when injecting typeids.

Even better, this will in practice let us do type name injection
without a join also, since clients get a type mapping in the new
protocol.

Fixes #6599.

### DIFF
--- a/edb/pgsql/compiler/pathctx.py
+++ b/edb/pgsql/compiler/pathctx.py
@@ -1087,6 +1087,19 @@ def has_type_rewrite(
     )
 
 
+def link_needs_type_rewrite(
+        typeref: irast.TypeRef, *, env: context.Environment) -> bool:
+    return (
+        has_type_rewrite(typeref, env=env)
+        # Typically we need to apply rewrites when looking at a link
+        # target that has a policy on it, but we suppress this for
+        # schema::ObjectType. None of the hidden objects should be
+        # user visible anyway, and this allows us to do type id
+        # injection without a join.
+        and str(typeref.real_material_type.name_hint) != 'schema::ObjectType'
+    )
+
+
 def find_path_output(
     rel: pgast.BaseRelation, ref: pgast.BaseExpr
 ) -> Optional[pgast.OutputVar]:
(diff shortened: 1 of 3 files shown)


elprans (approved):

edb/pgsql/compiler/pathctx.py
```diff
@@ -1087,6 +1087,19 @@ def has_type_rewrite(
...
+def link_needs_type_rewrite(
+        typeref: irast.TypeRef, *, env: context.Environment) -> bool:
+    return (
+        has_type_rewrite(typeref, env=env)
+        # Typically we need to apply rewrites when looking at a link
+        # target that has a policy on it, but we suppress this for
+        # schema::ObjectType. None of the hidden objects should be
+        # user visible anyway, and this allows us to do type id
+        # injection without a join.
+        and str(typeref.real_material_type.name_hint) != 'schema::ObjectType'
```
aljazerzen, line 1099: Comparing `name_hint`? Is lookup into the schema that slow?

aljazerzen (approved):
If I understand correctly, this is also a perf improvement, that should be noticeable when injecting type names (i.e. the CLI default)

edb/pgsql/compiler/pathctx.py
msullivan: We don't have the schema in the ir->pg compiler

msullivan:
> If I understand correctly, this is also a perf improvement, that should be noticeable when injecting type names (i.e. the CLI default)

Not directly, since injecting type names still requires joining, but then we'll be able to update the CLI to inject type ids instead of type names and do the type name lookup iteself

edb/pgsql/compiler/pathctx.py
aljazerzen: Of course, yes

PR #6603, opened 2023-12-13
Docs: Add offending code block to error message in docs tests

jaclarke:
Sometimes the reported approximate line number of broken code blocks in the docs tests are not particularly accurate, so also output the code block itself in the error message.


raddevon (approved):
🎉

PR #6604, opened 2023-12-13
4.3 cherry picks

fantix:
* #6531 
* #6562 
* #6588

### DIFF
--- a/edb/server/args.py
+++ b/edb/server/args.py
@@ -128,6 +128,29 @@ class ServerConnTransport(enum.StrEnum):
     SIMPLE_HTTP = "SIMPLE_HTTP"
 
 
+class ReloadTrigger(enum.StrEnum):
+    """
+    Configure what triggers the reload of the following config files:
+    1. TLS certificate and key (server config)
+    2. JWS key (server config)
+    3. Multi-tenant config file (server config)
+    4. Readiness state (server or tenant config)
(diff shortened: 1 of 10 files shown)


aljazerzen:
Hmm, I'll move these commits to the releases branch one by one, so they don't get squashed and we don't get a merge commit.

fantix:
I can do "rebase and merge" instead of "squash and merge" on this PR, which does the same thing I think

aljazerzen:
I think "rebase and merge" will still produce a merge commit:

```
A --------------------- Merge
 \__ CP1 __ CP2 __ CP3 __/
```

Since releases/4.x cannot be forcepushed to, I'd prefer not to rick non-linear history.

msullivan:
> I think "rebase and merge" will still produce a merge commit:
> 
> ```
> A --------------------- Merge
>  \__ CP1 __ CP2 __ CP3 __/
> ```
> 
> Since releases/4.x cannot be forcepushed to, I'd prefer not to rick non-linear history.

Rebase and merge doesn't produce a merge commit, it does a rebase. "Create a merge commit" will do a merge.

... I've pretty sure I've force pushed to release branches, but we should at least try not to

aljazerzen:
Oh, ok I see that I'm mistaken.

But yeah, I've force-pushed to release too, but I think Elvis disabled that because I was too liberal with it :dagger:

PR #6609, opened 2023-12-14
Fix DML-containing FOR when the iterator is an object set with duplicates

### LINKED ISSUE #6608: DML-containing FOR misbehaves on iterator sets with duplicates
See:
```
_localdev:insert> insert Subordinate { name := "foo" };
[{"id": "c850a212-9a16-11ee-b85c-e35d56999f20"}]
_localdev:insert> select count(Subordinate);
[1]
_localdev:insert> for x in {Subordinate, Subordinate} union (
.................   (x { name }, (insert Note { name := '', subject := x }))
................. );
[
  [{"name": "foo"}, {"id": "cb95a9e0-9a16-11ee-b85c-275254b2af44"}],
  [{"name": "foo"}, {"id": "cb95a9e1-9a16-11ee-b85c-8ba642b6289b"}],
  [{"name": "foo"}, {"id": "cb95a9e0-9a16-11ee-b85c-275254b2af44"}],
  [{"name": "foo"}, {"id": "cb95a9e1-9a16-11ee-b85c-8ba642b6289b"}],
  [{"name": "foo"}, {"id": "cb95a9e0-9a16-11ee-b85c-275254b2af44"}],
  [{"name": "foo"}, {"id": "cb95a9e1-9a16-11ee-b85c-8ba642b6289b"}],
  [{"name": "foo"}, {"id": "cb95a9e0-9a16-11ee-b85c-275254b2af44"}],
  [{"name": "foo"}, {"id": "cb95a9e1-9a16-11ee-b85c-8ba642b6289b"}]
]
_localdev:insert> for x in {Subordinate, Subordinate} union (
.................   (x, (insert Note { name := '', subject := x }))
................. );
```

(And actually, even without the tuple it will produce extra output rows)

There's actually a nice comment explaining what we do to prevent this bug: https://github.com/edgedb/edgedb/blob/f2e721e8782c4f71b9cfe6b93175e03e1869d6d6/edb/pgsql/compiler/clauses.py#L207-L210
Unfortunately the comment isn't true anymore and if we did actually previously handle this case, we never wrote a test for it :(

The tricky thing here is that even if we do force the iterator to have a transient identity value, there are places that expect object identity to be the actual object id

msullivan:
DML-containing FOR loops get compiled into CTEs, and after the DML is
performed, the CTEs need to get joined back with the DML CTEs in order
to produce the output. Currently we do these joins on the 'identity'
of the iterator set; for non-object sets, we generate a transient
identity with uuid_generate_v4, and for object sets we ues the actual
object id.

That works ok if all the objects in the iterator set are unique, but
if they aren't we produce a lot of duplicate output rows.

To solve this, we need to use transient ids for object types too, but
overriding object identity with something transient causes problems:
if we need to actually join the object against some *actual* table, we
are in trouble, since we've overriden the identity.

Solve this by introducing a new 'iterator' aspect and distinguishing
between whether we want to join on an iterator aspect or a real identity
when making path bonds.

Fixes #6608

### DIFF
--- a/edb/pgsql/ast.py
+++ b/edb/pgsql/ast.py
@@ -148,7 +148,7 @@ class EdgeQLPathInfo(Base):
 
     # Ignore the below fields in AST visitor/transformer.
     __ast_meta__ = {
-        'path_id', 'path_scope', 'path_outputs', 'is_distinct',
+        'path_id', 'path_bonds', 'path_outputs', 'is_distinct',
         'path_id_mask', 'path_namespace',
         'packed_path_outputs', 'packed_path_namespace',
     }
@@ -160,7 +160,7 @@ class EdgeQLPathInfo(Base):
     is_distinct: bool = True
 
     # A subset of paths necessary to perform joining.
-    path_scope: typing.Set[irast.PathId] = ast.field(factory=set)
+    path_bonds: typing.Set[tuple[irast.PathId, bool]] = ast.field(factory=set)
 
     # Map of res target names corresponding to paths.
     path_outputs: typing.Dict[
@@ -880,7 +880,7 @@ class RangeSubselect(PathRangeVar):
     subquery: Query
 
     @property
-    def query(self):
+    def query(self) -> Query:
         return self.subquery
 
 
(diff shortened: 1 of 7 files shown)


elprans (approved):

msullivan:
If this is tough to backport we can skip it, but otherwise we should

PR #6613, opened 2023-12-15
Compile single globals into materialized CTEs

msullivan:
Hopefully this will result in a lot less recomputation in certain
cases.


elprans (approved):

msullivan:
I think we do want to eventually want to backport this but maybe it's worth seeing if any of the interested users want to try it out with the nightly first?

msullivan:
I've reverted this on 4.x because it's caused two regressions:
* https://github.com/edgedb/edgedb/issues/7056
* https://discord.com/channels/841451783728529451/1217949944800088105

4.7 will go out Monday probably and will have this reverted.

I'll revert it on 5.x/main if I can't figure out fixes today.

PR #6616, opened 2023-12-17
Docs: Fix up some documentation typos and outdated syntax

johnpyp:
I'm new to EdgeDB so I could've gotten some syntax wrong, but:

- (access policies) Added a `version-lt` for the old 3.0 example and a new 4.0 one.
- (access policies) Using the selected global, it now compares `current_user` to `.author` rather than `.author.id`, as it should be `User` to `User` comparisons.
- Add some trailing semicolons to `default` statements
- Use `is` rather than `Is` during link assertions

- Replaced `link` with `multi` in a backlink declaration... I don't know if this is correct though, or if it should be omitted altogether. It isn't exactly clear to me.

### DIFF
--- a/docs/datamodel/access_policies.rst
+++ b/docs/datamodel/access_policies.rst
@@ -404,7 +404,7 @@ making the current user able to see their own ``User`` record.
     This change is being made to simplify reasoning about access policies and
     to allow certain patterns to be express efficiently. Since those who have
     access to modifying the schema can remove unwanted access policies, no
-    additional security is provided by applying access policies to each 
+    additional security is provided by applying access policies to each
     other's expressions.
 
     It is possible (and recommended) to enable this :ref:`future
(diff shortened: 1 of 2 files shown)


raddevon:
This is an excellent PR. Thank you for contributing it!

I made a few minor changes. I'll explain them since they might be helpful to understand if you'd like to contribute more in the future:

* You were right to update the syntax for the access policies custom error message example, but there's no need for the example in older syntax since this feature was introduced in 3.0 alongside the newer syntax. I removed the `:version-lt: 3.0` code block and kept your revised code block in the new syntax.
* Dropping the `link` keyword is a good idea, but it only works in 4.0 (whereas _having_ `link` still works in 4.0). In order to drop it, we need a new syntax block for versions prior to 4.0 where it is still present. I added that. (That said, I believe the example should have shown `multi link` originally, so I kept `multi` in the new pre-4.0 block. Nice catch there.)
* This change didn't make it into the final PR after my update but, FYI, if you have a diff code block with a `:version-lt:`, the indentation level of the diff symbols (`+`/`-`) must match the indentation level of the `:version-lt:`.

Thanks again for contributing this! I'll get it merged once the tests complete. We'd love to see more from you in the future if you're so inclined, and we hope you're enjoying EdgeDB!

raddevon (approved):

PR #6619, opened 2023-12-18
Remove handling for 'releases/*' in get_version_from_scm

msullivan:
This has been breaking all of our tests for the `releases/4.x` branch
basically for the entire 4.x cycle.

When that branch was introduced in 8481c088748862922bb360a9711ad5e5617aba25,
it looks like we were creating a 'releases/X.Y' branch
*for each minor version*. It looks like we then stopped
creating 'releases/X.Y' branches some time during the 2.x cycle.

Then for 4.x we decided to switch from using `stable/4.x` to using
`release/4.x`, which I think morphed into `releases/4.x` in some
attempt for consistency with all those old branches, which is now
triggering that version handling code which is resulting in version
numbers like '4.3-dev.8052', which the CLI rejects in certain paths
because it doesn't think -dev should be on a 4.3 version.

Delete that 'releases/*' handling, which doesn't really match our
current workflows.

I *also* think that we should delete all the old `releases/` branches
(they all have equivalent tags) and make the branch `release/4.x`
instead of `releases/4.x`.

### DIFF
--- a/edb/buildmeta.py
+++ b/edb/buildmeta.py
@@ -427,15 +427,6 @@ def get_version_from_scm(root: pathlib.Path) -> str:
     )
 
     proc = subprocess.run(
-        ['git', 'rev-parse', '--abbrev-ref', 'HEAD'],
-        stdout=subprocess.PIPE,
-        universal_newlines=True,
-        check=True,
-        cwd=root,
-    )
-    branch = proc.stdout.strip()
-
-    proc = subprocess.run(
         ['git', 'tag', '--list', 'v*'],
         stdout=subprocess.PIPE,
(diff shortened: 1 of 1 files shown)


aljazerzen:
I don't know what failures you are talking about - there are some that are show up when CLI interacts with the server that was built on `releases/` but does not have a tag. This is a tad annoying.

msullivan:
> I don't know what failures you are talking about - there are some that are show up when CLI interacts with the server that was built on `releases/` but does not have a tag. This is a tad annoying.

Yes, those failures.

aljazerzen (approved):

PR #6621, opened 2023-12-19
Fix test reporting when arbitrary errors are injected

aljazerzen:


msullivan (approved):

msullivan:
(I went ahead and merged this because I had some tests failing locally that were running into this problem)

PR #6633, opened 2023-12-21
Fix very slow global introspection query when there are lots of databases

msullivan:
The introspection query was fetching annotations, which can never be
populated on databases and which was exhibiting some pretty
pathological quadratic time introspection queries. With 500 dbs,
it took me about 30s.

Also add a new knob to the patch system so that we can apply this in a
patch without it needing to do anything expensive on each database.


elprans (approved):

msullivan:
```
    ('edgeql+schema+globalonly', ''),
```
is the patch we want, though if there is already an `edgeql+schema` patch in the same version then it isn't needed.

PR #6634, opened 2023-12-21
Add a dedicated sql-introspection patch type

msullivan:
This lets us refresh sql introspection views without needing to do an
edgeql+schema patch.

It up applying it to 100 dbs from ~90s to ~65s, which isn't as much as
I hoped but is still nicer.


elprans (approved):

msullivan:
@aljazerzen consider using this

PR #6639, opened 2023-12-22
Stop having sys::Database be an AnnotationSubject

msullivan:
Follow-up on #6633. That can be cherry-picked, this is for 5.0.

This is technically a backwards compatability break but a super
trivial one.


aljazerzen (approved):

elprans (changes requested):
I think we need to sit on this for a bit and see if we can improve annotation introspection on databases instead.  User/system-defined annotations on databases could be a useful feature, especially in context of some planned features.

PR #6643, opened 2023-12-22
Use correct signature for in-memory cache method

scotttrinh:
The `store` method on Hishel's `AsyncBaseStorage` is `async`, see: https://github.com/karpetrosyan/hishel/blob/3b3d5287869dc190118fc91c97a4e79a5b586ae2/hishel/_async/_storages.py#L44

### DIFF
--- a/edb/server/protocol/auth_ext/http_client.py
+++ b/edb/server/protocol/auth_ext/http_client.py
@@ -53,7 +53,7 @@ class InMemoryStorage(hishel._async._storages.AsyncBaseStorage):
         super().__init__()
         self._storage = lru.LRUMapping(maxsize=maxsize)
 
-    def store(self, key: str, response, request, metadata):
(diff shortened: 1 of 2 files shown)


aljazerzen (commented):
Can we write a test for this bug?

I know this part is mocked in tests, so we don't have an easy way to do it, but I do want to have a way of verifying this before releasing.

scotttrinh:
@aljazerzen 

I'm not entirely sure I know how we'd test it, since, yeah outgoing HTTP requests from the server are being mocked in the tests. We've tested it manually and it seems to resolve the issue, and it's strictly more correct, but I'm open to suggestion on a strategy here.

@1st1 any ideas?

scotttrinh:
Turns out we just were not hitting the cache! Just added a `cache-control` header and was able to reproduce the issue and confirm that the change fixes the problem.

PR #6644, opened 2023-12-22
Use newly released hishel in-memory cache

scotttrinh:
An alternative to #6643 , not sure which approach we would prefer to take, honestly.

Hishel released a new in-memory cache last week in [`0.0.20`](https://github.com/karpetrosyan/hishel/releases/tag/0.0.20) which negates our need to specify our own.


karpetrosyan (approved):

scotttrinh:
As discussed, we fixed out own storage in 4.x, but going into 5.x we'll switch to Hishel's built-in memory caching.

PR #6653, opened 2024-01-02
Fix slow global introspection of roles also

msullivan:
This is just a simple extension to #6633.


elprans (approved):

msullivan:
It looks like this is insufficient for roles to make it fast. Roles also have `bases`, `ancestors`, and `member_of` multi links and all of them are slow too.

I'll need to think a bit more.

msullivan:
I'm going to drop this approach

PR #6656, opened 2024-01-03
Add schema migration guide, reorder and rename in places

Dhghomon:
I've just finished rst-ifying and touching the schema migration guide that I showed as a pdf on our Discord before Christmas. In addition to the guide, I've added some opinionated changes and renaming to the /guides section of the docs and am also looking for opinions on whether this guide should be a single page or multiple files.

I'm leaning towards a single page, simply because (despite the length) it's a single subject and a menu pops up on the right anyway to differentiate between sections.

Taking a look at what else we have inside /guides, I've noticed the following:

- Deployment and Importing existing data show where multiple files make sense, as someone looking to deploy to AWS will have no interest in Azure, and vice versa. Similarily, inside *Important existing data* there is a section on Postgres and the structure is set up to add more later.
- Same for tutorials (which could work better as "EdgeDB with..." as discussed recently as the word tutorials makes you think of a sit-down course whereas these are more like tutorialized blueprints)
- Migration Patterns: this one is the odd man out as it's currently a collection of tips that could potentially be a single page, and feels like the predecessor to this guide. I would like to incorporate them as is into this new guide as a single page, but am curious what others think.

## Single page vs. multiple pages

Single page can look a bit intimidating, but otherwise IMO is much more convenient. It's a single subject, and making changes down the road would be much easier if left as one. (Syntax changes, new migration features, etc.) And splitting into multiple pages (about 10 or more) would necessitate coming up with file names that might not explain the content very well. e.g. names.rst is actually titled "Making a property required" and reqlink.rst is "Adding a required link". The file names are a good effort to match the titles for the most part but I think having difficulty in choosing a file name is a sign that maybe there doesn't need to be more than one.

For comparison, the file names inside /deployment are clear as day: aws_aurora_ecs.rst, docker.rst, heroku.rst, and so on.

Note: the website built with this will show both the single "Schema migration guide" and current "Migration Guide" so if there is agreement that single page is okay then Migration Guide will disappear.

The PR changes the order to the following:

Overview, Cloud, Cheatsheets, Deployment, Moving to EdgeDB, Using EdgeDB with, Auth, Schema migrations guide, Contributing

Renames: "Important existing data" to "Moving to EdgeDB", "Tutorials" to "Using EdgeDB with". I think Cheatsheets should be way up near the top.

image

Hopefully the line length limit gets killed soon! If not then I'll manually do it when we've agreed on everything else.


raddevon:
Just got finished going through this. It's really good! I'll have some edits to suggest, but here are some of my thoughts on the organizational changes (which I generally like a lot):
- I have a slight preference for "Switching to EdgeDB" vs. "Moving to EdgeDB." "Switching" feels closer to what the user is doing vs. "moving" to me. Although I can see an argument that "switching" might imply broader content than what that guide entails.
- "Migration **Guide**" feels redundant as a title since it's already under "Guides." I would favor changing it to just "Migrations"
- "Schema migrations guide" also feels a bit redundant and doesn't really help me understand what that piece of content is. I  notice that "Overview" is just a list of the same links that are already in the sidebar. I don't think that adds much value. I would favor keeping the content that is currently in "Schema migrations guide" but with the title "Overview" (or perhaps "Basics"). I think we can just ditch the page of links that is currently "Overview" altogether.
- We currently have both Guides > Schema migrations guide and Guides > Migrations Guide > Schema migrations guide pointing to the same piece of content. I like your idea of pulling the existing recipes into a subsection alongside your guide, but I would favor not also having the link directly under "Guides."
- Would it make sense to combine all the others as headings under a single page called "Recipes?" Then, combined with my other suggestions here, that would simplify the structure to this:
  - Guides
    - Migrations
      - Overview (or Basics)
      - Recipes

I'll start working my way through, suggesting other minor edits. Great work on this! Excited to publish it!

Dhghomon:
> Just got finished going through this. It's really good! I'll have some edits to suggest, but here are some of my thoughts on the organizational changes (which I generally like a lot):
> 
> * I have a slight preference for "Switching to EdgeDB" vs. "Moving to EdgeDB." "Switching" feels closer to what the user is doing vs. "moving" to me. Although I can see an argument that "switching" might imply broader content than what that guide entails.
> * "Migration **Guide**" feels redundant as a title since it's already under "Guides." I would favor changing it to just "Migrations"
> * "Schema migrations guide" also feels a bit redundant and doesn't really help me understand what that piece of content is. I  notice that "Overview" is just a list of the same links that are already in the sidebar. I don't think that adds much value. I would favor keeping the content that is currently in "Schema migrations guide" but with the title "Overview" (or perhaps "Basics"). I think we can just ditch the page of links that is currently "Overview" altogether.
> * We currently have both Guides > Schema migrations guide and Guides > Migrations Guide > Schema migrations guide pointing to the same piece of content. I like your idea of pulling the existing recipes into a subsection alongside your guide, but I would favor not also having the link directly under "Guides."
> * Would it make sense to combine all the others as headings under a single page called "Recipes?" Then, combined with my other suggestions here, that would simplify the structure to this:
>   
>   * Guides
>     
>     * Migrations
>       
>       * Overview (or Basics)
>       * Recipes
> 
> I'll start working my way through, suggesting other minor edits. Great work on this! Excited to publish it!

I'd still keep the "schema" part of the "schema migrations guide" (otherwise it could read as migrating from Postgres, etc. when this guide is completely internal to EdgeDB) but taking out the "guide" part does make sense.

Overview: It looks funny at the moment because an Overview page gets automatically generated for any section longer than a single page (see all the other multi-section parts of the Guide) and that's the part you mention about pointing to the same spot. That was to get your input before doing away with the multi-fileness of the current section - and that would get rid of the Overview part as well and replace the whole thing with a single guide.

So the structure would be even a bit simpler, just a page called Schema Migrations underneath Guides.

Dhghomon:
> Just got finished going through this. It's really good! I'll have some edits to suggest, but here are some of my thoughts on the organizational changes (which I generally like a lot):
> 
> * I have a slight preference for "Switching to EdgeDB" vs. "Moving to EdgeDB." "Switching" feels closer to what the user is doing vs. "moving" to me. Although I can see an argument that "switching" might imply broader content than what that guide entails.
> * "Migration **Guide**" feels redundant as a title since it's already under "Guides." I would favor changing it to just "Migrations"
> * "Schema migrations guide" also feels a bit redundant and doesn't really help me understand what that piece of content is. I  notice that "Overview" is just a list of the same links that are already in the sidebar. I don't think that adds much value. I would favor keeping the content that is currently in "Schema migrations guide" but with the title "Overview" (or perhaps "Basics"). I think we can just ditch the page of links that is currently "Overview" altogether.
> * We currently have both Guides > Schema migrations guide and Guides > Migrations Guide > Schema migrations guide pointing to the same piece of content. I like your idea of pulling the existing recipes into a subsection alongside your guide, but I would favor not also having the link directly under "Guides."
> * Would it make sense to combine all the others as headings under a single page called "Recipes?" Then, combined with my other suggestions here, that would simplify the structure to this:
>   
>   * Guides
>     
>     * Migrations
>       
>       * Overview (or Basics)
>       * Recipes
> 
> I'll start working my way through, suggesting other minor edits. Great work on this! Excited to publish it!

Okay, playing around with this a bit more. I was going to try to squish the tips into the guide but I think they work best unchanged, just put into a single file of their own. That lets the guide stand on its own, while the tips file is a bit more freeform and as a single file makes it easy for later PRs to add tips if we decide to add them. There is a bit of duplication (e.g. making a property required) but I don't think that's a problem especially since a second retelling of the same concept helps it to gel better.

Also might reorder the tips since some reference each other, in the meantime have just put them into this single file to give an idea of the final layout. Looks cleaner, I think!

image

Dhghomon:
Update: I've been putting the UI docs together as well and since they assume a reordering following this PR I think I'll add them here too.

image

raddevon (changes requested):
Still working through this, but I figured I'd go ahead and post what I have.

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+                RENAME TO nam;
+            };
+        };
+    But confidence is 0.67, below minimum threshold of 0.99999
+    Error executing command: EdgeDB is unable to make a decision.
+
+    Please run in interactive mode to confirm changes, or use
+    `--allow-unsafe`
+
+As the output suggests, you can add ``allow-unsafe`` to a non-interactive
```
raddevon, line 339:
```suggestion
As the output suggests, you can add ``--allow-unsafe`` to a non-interactive 
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+        };
+    But confidence is 0.67, below minimum threshold of 0.99999
+    Error executing command: EdgeDB is unable to make a decision.
+
+    Please run in interactive mode to confirm changes, or use
+    `--allow-unsafe`
+
+As the output suggests, you can add ``allow-unsafe`` to a non-interactive 
+migration if you truly want to push the suggestions through regardless 
+of the migration tool's confidence. But it's more likely in this case
```
raddevon, line 341:
```suggestion
of the migration tool's confidence, but it's more likely in this case 
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+      ALTER TYPE default::User {
+        CREATE PROPERTY nam: default::Name;
+        DROP PROPERTY name;
+      };
+    };
+
+.. note::
+
+    See the section below on migration hashes if you are curious about 
+    how migrations are named)
```
raddevon, line 376:
```suggestion
    how migrations are named.
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+
+So the ``User`` type was altered via creating a new property, and
```
raddevon, line 378:
```suggestion
This migration will alter the ``User`` type by creating a new property and 
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+    };
+
+.. note::
+
+    See the section below on migration hashes if you are curious about 
+    how migrations are named)
+
+So the ``User`` type was altered via creating a new property, and 
+dropping the old one. If that is what we wanted, then we can now type 
+``edgedb migrate`` to complete the migration.
```
raddevon, line 380:
```suggestion
``\migrate`` in the REPL or ``edgedb migrate`` at the command line to complete the migration.
```
Leading with the REPL command since this is what you used in the previous step.

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+
+Questions from the CLI
+======================
+
+So far we've only learned how to say yes or no to the CLI's questions
```
raddevon, line 385:
```suggestion
So far we've only learned how to say "yes" or "no" to the CLI's questions 
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+So far we've only learned how to say yes or no to the CLI's questions 
+when we migrate a schema, but quite a few other options are presented 
+when the CLI asks us a question:
+
+.. code-block::
+
+    did you create object type 'default::PlayerCharacter'? [y,n,l,c,b,s,q,?]
+    > y
+
+The choices ``y`` and ``n`` are obviously yes and no, and you can
```
raddevon, line 394:
```suggestion
The choices ``y`` and ``n`` are obviously "yes" and "no," and you can 
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+The choices ``y`` and ``n`` are obviously yes and no, and you can 
+probably guess that ``?`` will pull up a help menu, but the others 
+aren't so clear. Let's go over every option to make sure we understand 
+them.
+
+``y`` (or ``yes``)
+------------------
+
+This will accept the proposed change and move on to the next step. 
+If it's the last proposed change, the migration will now be complete.
```
raddevon, line 403:
```suggestion
If it's the last proposed change, the migration will now be created.
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+``n`` (or ``no``)
+-----------------
+
+This will reject the proposed change. At this point, the migration 
+tool will try to suggest a different change if it can, but it won't 
+always be able to do so.
+ 
+We can see this behavior with the same tiny schema change we made 
+above where we changed a property name from ``name`` to ``nam``. In 
+the output below, we see the following:
```
raddevon, line 414:
```suggestion
the output of that ``migration create``, we see the following:
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+
+- The CLI first asks us if we renamed the property, to which we say no.
+- It then tries to confirm that we have altered the ``User`` type. 
+  We say no again. 
+- The CLI then guesses that maybe we are dropping and creating the 
+  whole ``User`` type instead. This time, we say yes.
```
raddevon, line 420:
```suggestion
- The CLI first asks us if we renamed the property, to which we say "no".
- It then tries to confirm that we have altered the ``User`` type. 
  We say "no" again. 
- The CLI then guesses that maybe we are dropping and creating the 
  whole ``User`` type instead. This time, we say "yes."
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+- It then asks us to confirm that we are creating a ``User`` type, 
+  since we have decided to drop the existing one.
+
+But if we say no again to this question, the CLI will throw its hands
```
raddevon, line 424:
```suggestion
If we say "no" again to the final question, the CLI will throw its hands 
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+  whole ``User`` type instead. This time, we say yes.
+- It then asks us to confirm that we are creating a ``User`` type, 
+  since we have decided to drop the existing one.
+
+But if we say no again to this question, the CLI will throw its hands 
+up and tell us that it doesn't know what we are trying to do because 
+there is no way left for it to migrate to the schema that we have 
+told it to move to.
+
+Here is the output:
```
raddevon, line 429:
```suggestion
Here is what that would look like:
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+    the provided answers. Please retry with different answers.
+
+``l`` (or ``list``)
+-------------------
+
+This is used to see (list) the actual DDL statements that are being proposed.
+When asked the question ``did you alter object type 'default::User'?``
+in the example above, we might be wondering exactly what changes will 
+be made here. How exactly does the database intend to alter the ``User`` 
+type if we say yes? Simply clicking ``l`` will show it:
```
raddevon, line 453:
```suggestion
type if we say "yes?" Simply pressing ``l`` will show it:
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+
+.. code-block::
+
+    The following DDL statements will be applied:
+      ALTER TYPE default::User {
+          CREATE PROPERTY nam: std::str;
+          DROP PROPERTY name;
+      };
+
+This shows us clear as day that saying ``yes`` will result in creating
```
raddevon, line 463:
```suggestion
This shows us clear as day that saying "yes" will result in creating 
```

It looks like in most instances up to now, you're not styling "yes" as inline code, which I think is correct since the actual response is simply `y`. I've made this match those other instances but also surrounded it with quotes to reflect this is our reply like I have with the others.

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+    The following DDL statements will be applied:
+      ALTER TYPE default::User {
+          CREATE PROPERTY nam: std::str;
+          DROP PROPERTY name;
+      };
+
+This shows us clear as day that saying ``yes`` will result in creating 
+a new property called ``nam`` and dropping the existing ``name`` property.
+
+So when doubts dwell, click the letter l!
```
raddevon, line 466:
```suggestion
So when doubts dwell, press the letter "l!"
```
I dig the mnemonic!

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+
+The following two keys will stop the migration, but in different ways:
+
+``s`` (or ``stop``)
+-------------------
+
+This is also known as a 'split'. Pressing ``s`` will complete the 
+migration at the current point. Any statements that you have applied 
+will be applied, but the schema will not yet match the schema in your 
+``.esdl`` file(s). But you can easily start another migration to complete
```
raddevon, line 490:
```suggestion
``.esdl`` file(s). You can easily start another migration to complete 
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+    {
+      ALTER TYPE default::User {
+        CREATE PROPERTY nam: default::Name;
+        DROP PROPERTY name;
+      };
+    };
+
+.. note::
+
+    See the section below on migration hashes if you are curious about
```
raddevon, line 375: We should add a label to the section we're referencing and link the phrase "migration hashes" to it.

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+So how do you create a migration with only data? To do this, just 
+add ``--allow-empty`` to the command:
+
+.. code-block:: bash
+
+    $ edgedb migration create --allow-empty
+    Created myproject/dbschema/migrations/00002.edgeql,
+    id: m1xseswmheqzxutr55cu66ko4oracannpddujg7gkna2zsjpqm2g3a
+
+You will now see an empty migration in which you can enter some queries.
```
raddevon, line 525:
```suggestion
You will now see an empty migration in ``dbschema/migrations`` in which you can enter some queries. 
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+the remaining changes once you have applied the migration that was 
+just created. This effectively splits the migration into two or more 
+files.
+
+``q`` (or ``quit``)
+-------------------
+
+Pressing ``q`` will simply quit without saving any of your progress.
+
+Migration hashes and data migrations
```
raddevon, line 500:
```suggestion
Data migrations and migration hashes
```
Changing the order to reflect the order they're discussed.

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+
+    CREATE MIGRATION m1xseswmheqzxutr55cu66ko4oracannpddujg7gkna2zsjpqm2g3a
+        ONTO m1n5lfw7n74626cverbjwdhcafnhmbezjhwec2rbt46gh3ztoo7mqa
+    {
+        insert User { name := 'User 1'};
+        insert User { name := 'User 2'};
+        delete User filter .name = 'User 2';
+    };
+
+But when you type ``edgedb migrate``, the CLI will then complain that
```
raddevon, line 549:
```suggestion
The problem is, if you save that migration and run ``edgedb migrate``, the CLI will complain that 
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+    {
+        insert User { name := 'User 1'};
+        insert User { name := 'User 2'};
+        delete User filter .name = 'User 2';
+    };
+
+But when you type ``edgedb migrate``, the CLI will then complain that 
+the migration hash doesn't match what it is supposed to be. However, 
+it helpfully provides the reason: "Migration names are computed from 
+the hash of the migration contents".
```
raddevon, line 552:
```suggestion
the hash of the migration contents."
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+
+Fortunately, it also tells you exactly what the hash (the migration 
+name) will need to be and you can simply change it to that.
```
raddevon, line 555:
```suggestion
name) will need to be, and you can simply change it to that.
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+
+
+    Migration names are computed from the hash of the migration contents.
+
+    To proceed you must fix the statement to read as:
+    CREATE MIGRATION m13g7j2tqu23yaffv6wkn2adp6hayp76su2qtg2lutdh3mmj5xyk6q 
+    ONTO ...
+    Alternatively, revert the changes to the file.
+
+You can also create a migration that combines schema changes and data.
```
raddevon, line 576:
```suggestion
Aside from exclusive data migrations, you can also create a migration that combines schema changes *and* data. 
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+
+    Migration names are computed from the hash of the migration contents.
+
+    To proceed you must fix the statement to read as:
+    CREATE MIGRATION m13g7j2tqu23yaffv6wkn2adp6hayp76su2qtg2lutdh3mmj5xyk6q 
+    ONTO ...
+    Alternatively, revert the changes to the file.
+
+You can also create a migration that combines schema changes and data. 
+This is even easier, since it doesn't even require appending ``allow-empty``
```
raddevon, line 577:
```suggestion
This is even easier, since it doesn't even require appending ``--allow-empty`` 
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+You can also create a migration that combines schema changes and data. 
+This is even easier, since it doesn't even require appending ``allow-empty`` 
+to the command. Just do the following:
+
+- Change your schema,
+- Type ``edgedb migration create`` and respond to the CLI's questions,
+- Add your queries to the file (best done on the bottom after the 
+  DDL statements have changed the schema),
+- Type ``edgedb migrate`` and change the migration name to the suggested name,
+- Type ``edgedb migrate`` again.
```
raddevon, line 585:
```suggestion
1. Change your schema
2. Type ``edgedb migration create`` and respond to the CLI's questions
3. Add your queries to the file (best done on the bottom after the 
   DDL statements have changed the schema)
4. Type ``edgedb migrate`` and change the migration name to the suggested name
5. Type ``edgedb migrate`` again
```
Changing to an ordered list since order is important here. Also ditching the commas since a series is implied by the list.

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+
+The `EdgeDB tutorial <tutorial_>`_ is a good example of a database 
+set up with both a schema migration and a data migration. Setting 
+up a database with schema changes in one file and default data in 
+a second file is a nice way to separate the two operations and maintain 
+high readability at the same time. These two files can be `seen here 
+<tutorial_files_>`_.
```
raddevon, line 592:
```suggestion
up a database with `schema changes in one file and default data in 
a second file <tutorial_files_>`_ is a nice way to separate the two operations and maintain 
high readability at the same time.
```
Trying to make the link label more reflective of what we're linking to for accessibility.

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+
+Squashing migrations
+====================
+
+Users often end up making frequent changes to their schema because
```
raddevon, line 597:
```suggestion
Users often end up making many changes to their schema because 
```
My thinking here is that it's the quantity of changes rather than the frequency that creates the need for squashing.

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+a second file is a nice way to separate the two operations and maintain 
+high readability at the same time. These two files can be `seen here 
+<tutorial_files_>`_.
+
+Squashing migrations
+====================
+
+Users often end up making frequent changes to their schema because 
+of how effortless it is to do. (And in the next section we will learn 
+about EdgeDB Watch, which is ever more effortless!) This leads to
```
raddevon, line 599:
```suggestion
about EdgeDB Watch, which is even more effortless!) This leads to 
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+
+Squashing migrations
+====================
+
+Users often end up making frequent changes to their schema because 
+of how effortless it is to do. (And in the next section we will learn 
+about EdgeDB Watch, which is ever more effortless!) This leads to 
+an interesting side effect: lots of ``.edgeql`` files, many of which 
+represent trials and approaches that don't end up making it to the 
+final form for your schema.
```
raddevon, line 602:
```suggestion
final schema.
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+    (00002.edgeql)
+    Applied m1ixxlsdgrlinfijnrbmxdicmpfav33snidudqi7fu4yfhg4nngoza 
+    (00003.edgeql)
+    Applied m1tsi4amrdbcfjypu72duyckrlvvyb46r3wybd7qnbmem4rjvnbcla 
+    (00004.edgeql)
+    ...and so on...
+    Project initialized.
+
+To squash your migrations, just add ``--squash`` after ``edgedb migration 
+create``. Running this command will first display some helpful info
```
raddevon, line 627:
```suggestion
To squash your migrations, just run ``edgedb migration create`` with the
 ``--squash`` option. Running this command will first display some helpful info 
```
"just add…" implies the user would have already been running `edgedb migration create`, but in this instance, the user would only run it if their intention was to squash since there are no outstanding schema changes to create migrations from. Reformulated to reflect this.

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+    2. Ensure that other users of the database have the revision
+    above or can create database from scratch.
+        To check a specific instance, run:
+        edgedb -I <name> migration log --from-db --newest-first --limit 1
+    1. Merge version control branches that contain schema changes
+    if possible.
+
+    Proceed? [y/n]
+
+Then just press ``y`` to squash all of your existing migrations into
```
raddevon, line 648:
```suggestion
Press ``y`` to squash all of your existing migrations into 
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+
+Then just press ``y`` to squash all of your existing migrations into 
+a single file.
+
+Fixups during a squash
+----------------------
+
+If your schema doesn't match the schema in the database, EdgeDB will 
+prompt you to create a *fixup* file, which can be useful (as the CLI 
+says) to "automate upgrading other instances to a squashed revision".
```
raddevon, line 656:
```suggestion
prompt you to create a *fixup* file, which can be useful to, as the CLI 
says, "automate upgrading other instances to a squashed revision". 
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+a single file.
+
+Fixups during a squash
+----------------------
+
+If your schema doesn't match the schema in the database, EdgeDB will 
+prompt you to create a *fixup* file, which can be useful (as the CLI 
+says) to "automate upgrading other instances to a squashed revision". 
+You'll see fixups inside a folder called ``/fixups``. Their file names 
+are extremely long, but are simply two migration hashes joined together
```
raddevon, line 658:
```suggestion
are extremely long because they are simply two migration hashes joined together 
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+
+Fixups during a squash
+----------------------
+
+If your schema doesn't match the schema in the database, EdgeDB will 
+prompt you to create a *fixup* file, which can be useful (as the CLI 
+says) to "automate upgrading other instances to a squashed revision". 
+You'll see fixups inside a folder called ``/fixups``. Their file names 
+are extremely long, but are simply two migration hashes joined together 
+by a dash. So a fixup that begins with
```
raddevon, line 659:
```suggestion
by a dash. This means a fixup that begins with
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+    m1v3vqmwif4ml3ucbzi555mjgm4myxs2husqemopo2sz2m7otr22ka 
+    ONTO m16awk2tzhtbupjrzoc4fikgw5okxpfnaazupb6rxudxwin2qfgy5q
+
+will have a file name a full 116 characters in length.
+
+The CLI output when using squash along with a fixup is pretty informative 
+on its own, so let's just walk through the output as you'll see it 
+in practice.
+
+First we'll begin with this schema:
```
raddevon, line 673:
```suggestion
in practice. First we'll begin with this schema:
```

raddevon: Also, we should specify how we want the reader to follow along here. Should they start a new project or migrate the project we created earlier to this new schema?

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+
+.. code-block:: sdl
+
+  type User {
+    name: str;
+  }
+
+Then remove ``name: str;`` from the ``User`` type, migrate, put it 
+back again, and migrate. You'll now have multiple files in your
```
raddevon, line 682: Maybe we can suggest "removing" it by commenting it out since this could also be a useful technique to learn.

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+``/migrations`` folder and will now be able to squash them.
+
+Then change to this schema without migrating it:
```
raddevon, line 685:
```suggestion
Then change to this schema **without migrating it**:
```
Just want to make sure this part gets noticed.

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+
+.. code-block:: sdl
+
+  type User {
+    name: str;
+    nickname: str;
+  }
+
+Upon using ``edgedb migration create --squash``, the output is first
```
raddevon, line 694:
```suggestion
Now run ``edgedb migration create --squash``. The output is first 
```
Want to make sure it's clear what the reader should do.

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+    to automate upgrading other instances to a squashed revision.
+    This starts the usual migration creation process.
+
+    Feel free to skip this step if you don't have
+    other instances to migrate
+
+    Create a fixup file? [y/n]
+    > y
+
+You will then see the the same questions that otherwise show up in
```
raddevon, line 733:
```suggestion
You will then see the the same questions that otherwise show up in 
```
```suggestion
You will then see the the same questions that would otherwise show up in 
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+Then just press ``y`` to squash all of your existing migrations into 
+a single file.
+
+Fixups during a squash
+----------------------
+
+If your schema doesn't match the schema in the database, EdgeDB will 
+prompt you to create a *fixup* file, which can be useful (as the CLI 
+says) to "automate upgrading other instances to a squashed revision". 
+You'll see fixups inside a folder called ``/fixups``. Their file names
```
raddevon, line 657:
```suggestion
You'll see fixups inside ``/dbschema/fixups``. Their file names 
```
to avoid implying fixups will appear at `fixups` from the project root

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+        git add dbschema
+
+    The normal migration process will update your migration history:
+        edgedb migrate
+
+    $ edgedb migrate
+    
+    Applied m1v3vqmwif4ml3ucbzi555mjgm4myxs2husqemopo2sz2m7otr22ka 
+    (m16awk2tzhtbupjrzoc4fikgw5okxpfnaazupb6rxudxwin2qfgy5q-
+    m1oih6aevfcftysukvofwuth2bsuj5aahkdnpabscry7p7ljkgbxma.edgeql)
```
raddevon, line 759: I would split this code block with a quick sentence to show this is where you entered a new command.
```suggestion

We'll take its suggestion to apply the migration:

.. code-block:: bash

    $ edgedb migrate
    
    Applied m1v3vqmwif4ml3ucbzi555mjgm4myxs2husqemopo2sz2m7otr22ka 
    (m16awk2tzhtbupjrzoc4fikgw5okxpfnaazupb6rxudxwin2qfgy5q-
    m1oih6aevfcftysukvofwuth2bsuj5aahkdnpabscry7p7ljkgbxma.edgeql)
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+
+
+.. note::
+
+    Squashing is still limited to schema changes, so queries inside
```
raddevon, line 764:
```suggestion
    Squashing is limited to schema changes, so queries inside 
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+    (m16awk2tzhtbupjrzoc4fikgw5okxpfnaazupb6rxudxwin2qfgy5q-
+    m1oih6aevfcftysukvofwuth2bsuj5aahkdnpabscry7p7ljkgbxma.edgeql)
+
+
+.. note::
+
+    Squashing is still limited to schema changes, so queries inside 
+    data migrations will be discarded during a squash.
+
+EdgeDB Watch
```
raddevon, line 767: I'm not sure how I feel about sorta "branding" the `edgedb watch` command by title casing it. I think I lean toward just referring to it as the command since it only ever exists in that context. If there were also an interface to it via the UI, I could see this making sense, but since it's exclusively that command, I'd lean toward just referring to it as a command.

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+============
+
+Another option when quickly iterating over schema changes is ``edgedb watch``.
+This will create a long-running process that keeps track of every time you
+save as ``.esdl`` file inside your ``/migrations`` folder, letting you know
```
raddevon, line 772:
```suggestion
save an ``.esdl`` file inside your ``/migrations`` folder, letting you know
```

In addition to this suggestion, we may want to consider a broader rewrite.

> …letting you know if your changes have successfully compiled or not.

This almost seems like a side effect. Yes, it lets you know whether your changes successfully applied, but that's because it needed to let you know the outcome of applying them. Watching your schema files and _applying_ the changes seems like it's actually the important thing the command does. As a consequence, if it _can't_ apply them, it will let you know.

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+    error: type 'default::i32' does not exist
+    ┌─ myproject/dbschema/default.esdl:5:25
+    │
+    5 │         wrong_property: i32;
+    │                         ^^^ error
+
+    Schema migration error:
+    cannot proceed until .esdl files are fixed
+
+Once you correct the ``i32`` property to ``int32``, EdgeDB Watch will
```
raddevon, line 841:
```suggestion
Once you correct the ``i32`` type to ``int32``, EdgeDB Watch will 
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+    ┌─ myproject/dbschema/default.esdl:5:25
+    │
+    5 │         wrong_property: i32;
+    │                         ^^^ error
+
+    Schema migration error:
+    cannot proceed until .esdl files are fixed
+
+Once you correct the ``i32`` property to ``int32``, EdgeDB Watch will 
+let you know that things are okay now, and will become quiet again:
```
raddevon, line 842:
```suggestion
let you know that things are okay now and will resume quietly watching your schema and applying migrations:
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+
+.. code-block::
+
+    Resolved. Schema is up to date now.
+
+EdgeDB Watch is best run in a separate tab on your command line so
```
raddevon, line 848:
```suggestion
EdgeDB Watch is best run in a separate instance of your command line so 
```
to account for the fact not all terminal emulators/CLI interfaces support tabs

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+Once you correct the ``i32`` property to ``int32``, EdgeDB Watch will 
+let you know that things are okay now, and will become quiet again:
+
+.. code-block::
+
+    Resolved. Schema is up to date now.
+
+EdgeDB Watch is best run in a separate tab on your command line so 
+that you can take care of other tasks—including officially migrating 
+when you are satisfied with your current schema—without having to
```
raddevon, line 850:
```suggestion
that you can take care of other tasks — including officially migrating 
when you are satisfied with your current schema — without having to 
```
This isn't more correct, but this is the way I've been trying to present em dashes so I'll replicate it here to be consistent.

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+try the following query after you have made some changes:
+
+.. code-block::
+
+    group schema::Migration {
+        name,
+        script
+    } by .generated_by;
+
+Some migrations will contain nothing in their ``generated_by`` property,
```
raddevon, line 863: Add a brief sentence after the query telling what it does. Maybe something like this: "This will return a list of applied migrations."

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+    group schema::Migration {
+        name,
+        script
+    } by .generated_by;
+
+Some migrations will contain nothing in their ``generated_by`` property, 
+while those generated by EdgeDB Watch will have a 
+``MigrationGeneratedBy.DevMode``. 
+
+(The final option for ``generated_by`` is
```
raddevon, line 867:
```suggestion
(The final option aside from `DevMode` and the empty set for ``generated_by`` is
```
assuming these are, in fact, the other two 😅

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+``MigrationGeneratedBy.DDLStatement``, which will show up if you directly
+change your schema by using DDL - which is generally not recommended)
```
raddevon, line 869:
```suggestion
change your schema by using DDL, which is generally not recommended)
```

raddevon, line 869:
```suggestion
.. note::

    The final option (aside from ``DevMode`` and the empty set) for ``generated_by`` is
    ``MigrationGeneratedBy.DDLStatement``, which will show up if you directly
    change your schema by using DDL, which is generally not recommended.
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+
+Once you are satisfied with your changes while running EdgeDB Watch, 
+just create the migration with ``edgedb migration create`` and then 
+apply them with one small change: ``edgedb migrate --dev-mode`` to
```
raddevon, line 873:
```suggestion
apply them with one small tweak to the ``migrate`` command: ``edgedb migrate --dev-mode`` to 
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+change your schema by using DDL - which is generally not recommended)
+
+Once you are satisfied with your changes while running EdgeDB Watch, 
+just create the migration with ``edgedb migration create`` and then 
+apply them with one small change: ``edgedb migrate --dev-mode`` to 
+let the CLI know to apply the migrations made during dev mode that 
+were made by EdgeDB Watch.
+
+So you really want to use DDL
+=============================
```
raddevon, line 878:
```suggestion
So, you really want to use DDL?
===============================
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+
+You might have a good reason to use a direct DDL statement or two 
+to change your schema. How do you make that happen? EdgeDB disables 
+the usage of DDL by default, so this attempt to use DDL will not work:
```
raddevon, line 882: DDL is only disabled after you run your first migration, right? If so, we should note that.

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+    1 │ create type MyType;
+    │ ^^^^^^^^^^^^^^^^^^ Use the migration commands instead.
+    │
+    = The `allow_bare_ddl` configuration variable is set to
+    'NeverAllow'.  The `edgedb migrate` command normally sets
+    this to avoid accidental schema changes outside of the 
+    migration flow.
+
+This configuration can be overridden by the following command which 
+changes the enum ``allow_bare_ddl``from the default ``NeverAllow``
```
raddevon, line 900:
```suggestion
changes the enum ``allow_bare_ddl`` from the default ``NeverAllow`` 
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+
+.. code-block:: edgeql-repl
+
+    db> configure current database set allow_bare_ddl := 'AlwaysAllow';
+
+Note that the command is ``configure current database`` and not ``configure 
+instance``, as ``allow_bare_ddl`` is evaluated on the database level.
+
+That wasn't so bad, so why did the CLI tell us to try to "avoid accidental 
+schema changes outside of the migration flow"? Why is DDL disabled
```
raddevon, line 911:
```suggestion
schema changes outside of the migration flow?" Why is DDL disabled 
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+.. code-block:: edgeql-repl
+
+    db> configure current database set allow_bare_ddl := 'AlwaysAllow';
+
+Note that the command is ``configure current database`` and not ``configure 
+instance``, as ``allow_bare_ddl`` is evaluated on the database level.
+
+That wasn't so bad, so why did the CLI tell us to try to "avoid accidental 
+schema changes outside of the migration flow"? Why is DDL disabled 
+by default in the first place?
```
raddevon, line 912:
```suggestion
after running a migration in the first place?
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+
+So you really wanted to use DDL but now regret it
+=================================================
```
raddevon, line 915:
```suggestion
So, you really wanted to use DDL but now regret it?
===================================================
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+        computed_fields: [],
+        script: 'SET generated_by := 
+            (schema::MigrationGeneratedBy.DDLStatement);
+    CREATE TYPE SomeType;',
+        message: {},
+        generated_by: DDLStatement,
+    },
+    }
+
+Fortunately, the fix is not too hard: we can use a command called
```
raddevon, line 999:
```suggestion
Fortunately, the fix is not too hard: we can use the command
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+manually changed via DDL. If in doubt, use ``describe schema as sdl`` 
+to compare or use ``edgedb migration create`` and check the output. 
+If the CLI is asking you if you want to drop a type, that means that 
+you forgot to add it to the schema inside your ``.esdl`` file(s).
+
+
+Multiple migrations to keep data 
+================================
+
+Let's say that we have a simple ``User`` type with a ``friends`` link
```
raddevon, line 1019: We should either tell readers _not_ to follow along or give them instructions how they should. Do they migrate what they have, or start a new project?

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+    CREATE TYPE SomeType;',
+        message: {},
+        generated_by: DDLStatement,
+    },
+    }
+
+Fortunately, the fix is not too hard: we can use a command called 
+``edgedb migration extract``. This command will retrieve the migration(s) 
+created using DDL and assign each of them a proper file name and hash 
+inside the ``/migrations`` folder, effectively giving them a proper
```
raddevon, line 1002:
```suggestion
inside the ``/dbschema/migrations`` folder, effectively giving them a proper 
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+
+.. code-block::
+
+    CREATE MIGRATION m1xseswmheqzxutr55cu66ko4oracannpddujg7gkna2zsjpqm2g3a
+        ONTO m1n5lfw7n74626cverbjwdhcafnhmbezjhwec2rbt46gh3ztoo7mqa
+    {
+    };
+
+You can now add your queries inside the braces. Assuming a schema 
+with a simple ``User`` type, we could then add a bunch of queries
```
raddevon, line 536: Rather than showing this way to add the data migrations and manually update the hash, should we demonstrate [`edgedb migration edit`](https://www.edgedb.com/docs/cli/edgedb_migration/edgedb_migration_edit) which will open the migration in your editor and automatically update the hash for you? Then we could circle back and say that, if you had edited the migration directly instead, you would have gotten this error and could manually update the hash.

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+      type User {
+        name: str;
+        multi friends: User;
+        friend_names: array<str>;
+      }
+    }
+
+The CLI will simply ask us if we created a property called ``friend_names``. 
+We haven't applied the migration yet, so we might as well put the 
+data inside the same migration. A simple update will do the job!
```
raddevon, line 1086: We don't say how the reader should do this, and I think it would be better if we did. We should probably recommend `edgedb migration edit` for this to make sure the hash gets updated.

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+from our schema and do a final migration.
+
+Migration internals
+===================
+
+We've now reached the most optional part of the migrations tutorial, 
+but an interesting one for those curious about what goes on behind 
+the scenes during a migration.
+
+Migrations in EdgeDB before the advent of the EdgeDB Project flow
```
raddevon, line 1139: Similar feedback here to earlier uses of "EdgeDB Watch" as if it were a term we wanted to establish as a brand. There exists a concept of a lower-case 'p' "project" in EdgeDB, and there is a command `edgedb project`. Not sure about "EdgeDB Project" though.
```suggestion
Migrations in EdgeDB before the advent of the ``edgedb project`` flow 
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+
+Migration internals
+===================
+
+We've now reached the most optional part of the migrations tutorial, 
+but an interesting one for those curious about what goes on behind 
+the scenes during a migration.
+
+Migrations in EdgeDB before the advent of the EdgeDB Project flow 
+were still automated, but required more manual work if you didn't
```
raddevon, line 1140:
```suggestion
were still automated but required more manual work if you didn't 
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+but an interesting one for those curious about what goes on behind 
+the scenes during a migration.
+
+Migrations in EdgeDB before the advent of the EdgeDB Project flow 
+were still automated, but required more manual work if you didn't 
+want to accept all of the suggestions provided by the server. This 
+process is in fact still used to migrate even today; the CLI just 
+facilitates it by making it easy to respond to the generated suggestions.
+
+They took place `inside a transaction <transaction_>`_ handled by
```
raddevon, line 1145:
```suggestion
Pre-project migrations took place `inside a transaction <transaction_>`_ handled by 
```
I think you've said enough here between the pronoun and its antecedent to justify naming it again.

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+
+This starts the migration, after which the quickest process was to 
+type ``populate migration`` to accept the statements suggested by 
+the server, and then ``commit migration`` to finish the process.
+
+Now, there is another option besides simply typing ``populate migration`` 
+that allows you to look at and handle the suggestions every step of 
+the way (in the same way the CLI does today), and this is what we 
+are going to have some fun with. You can see `the original RFC <rfc_>`_ 
+for this if you are curious.
```
raddevon, line 1160:
```suggestion
are going to have some fun with. You can see `the original migrations RFC <rfc_>`_ 
if you are curious.
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+The server is telling us with ``"complete": false`` that this suggestion 
+is not the final step in the migration, that it is 67% confident that 
+its suggestion is correct, and that we should probably type the following 
+statement:
+
+.. code-block::
+
+    ALTER TYPE default::User { ALTER PROPERTY name { RENAME TO nam; };};
+
+Don't forget to remove the ``\n`` from inside the original suggestion
```
raddevon, line 1248:
```suggestion
Don't forget to remove the newlines (``\n``) from inside the original suggestion 
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+for this if you are curious.
+
+It is *very* finicky compared to the CLI, resulting in a failed transaction 
+if any step along the way is different from the expected behavior, 
+but is an entertaining challenge to try to get right if you want to 
+truly understand how migrations work in EdgeDB.
+
+This process requires looking at the server's proposed solutions every 
+step of the way, and these steps are best seen in JSON format. Let's 
+first set the CLI to make the format nicely readable with this command:
```
raddevon, line 1169: We should tell readers to start a new project before we get to this point if they want to follow along. (I assume that's the intention here anyway is for them to have a fresh project.)

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+
+.. code-block:: edgeql-repl
+
+    db> \set output format json-pretty
```
raddevon, line 1173:
```suggestion
    db> \set output-format json-pretty
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+    ...     }
+    ...   }
+    ... };
+
+You should get the output ``OK: START MIGRATION``, followed by a prompt 
+that ends with ``[tx]`` to show that we are inside of a transaction. 
+Anything we do here will have no effect on the current registered 
+schema until we finally commit the migration.
+
+So now what do we do? We could simply type ``populate migration``
```
raddevon, line 1214: I think we've skipped a step here. The `describe migration` output suggests we have migrated to the new property name `nam`, but I don't think we've actually performed that migration yet. We need to first complete the initial migration and then start a new one with the new property name before we can get to this point.

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+Don't forget to remove the ``\n`` from inside the original suggestion 
+- the transaction will fail if you don't take them out. If the migration 
+fails at any step you will see ``[tx]`` change to ``[tx:failed]`` 
+and you will have to type ``abort migration`` to leave the transaction 
+and begin the migration again.
+
+Technically, at this point you are permitted to write any DDL statement 
+you like and the migration tool will adapt its suggestions to reach 
+the desired schema. But this is more than likely to generate an error 
+when you try to commit the migration, and is bad practice in any case.
```
raddevon, line 1257:
```suggestion
the desired schema. Doing so though is bad practice and is more than likely to generate an error 
when you try to commit the migration. 
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+- the transaction will fail if you don't take them out. If the migration 
+fails at any step you will see ``[tx]`` change to ``[tx:failed]`` 
+and you will have to type ``abort migration`` to leave the transaction 
+and begin the migration again.
+
+Technically, at this point you are permitted to write any DDL statement 
+you like and the migration tool will adapt its suggestions to reach 
+the desired schema. But this is more than likely to generate an error 
+when you try to commit the migration, and is bad practice in any case. 
+(But give it a try if you're curious)
```
raddevon, line 1258:
```suggestion
(Even so, give it a try if you're curious.)
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+
+So let's dutifully type the suggested statement above, and then use
```
raddevon, line 1260:
```suggestion
Let's dutifully type the suggested statement above, and then use 
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+the desired schema. But this is more than likely to generate an error 
+when you try to commit the migration, and is bad practice in any case. 
+(But give it a try if you're curious)
+
+So let's dutifully type the suggested statement above, and then use 
+``describe current migration as json`` again to see what the current 
+status of the migration is. This time we see two major differences: 
+"complete" is now ``true``, meaning that we are at the end of the 
+proposed migration, and "proposed" does not contain anything. We can 
+also see our confirmed statement(s) inside "confirmed" at the bottom.
```
raddevon, line 1265:
```suggestion
also see our confirmed statement inside "confirmed" at the bottom.
```
since we're just talking about the case at hand here and not the generic case.

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+
+    {
+    "parent": "m1fgpuxbvd74m6pb72rdikakjv3fv7cftrez7r56qjgonboimp5zoa",
+    "complete": true,
+    "proposed": null,
+    "confirmed": ["ALTER TYPE default::User {\n ALTER PROPERTY name 
+    {\n RENAME TO nam;\n };\n};"]
+    }
+
+With this done, you can type ``commit migration`` and the migration
```
raddevon, line 1277:
```suggestion
With this done, you can commit the migration and the migration 
```

Lean toward referring to the action rather than the command here, both since you show the command below and since the command would actually require a semicolon which is a bit awkward to place inline in prose.

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+
+With this done, you can type ``commit migration`` and the migration 
+will be complete.
+
+.. code-block:: edgeql-repl
+
+    db[tx]> commit migration;
+    OK: COMMIT MIGRATION
+
+And because this migration was created using direct DDL statements,
```
raddevon, line 1285:
```suggestion
Since this migration was created using direct DDL statements, 
```

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+
+.. code-block:: edgeql-repl
+
+    db[tx]> commit migration;
+    OK: COMMIT MIGRATION
+
+And because this migration was created using direct DDL statements, 
+you will need to use ``edgedb migration extract`` to extract the latest
+migration and give it a proper ``.edgeql`` file in the same way we 
+did above in the "So you really wanted to use DDL but now regret it"
```
raddevon, line 1288:
```suggestion
did above in the "So you really wanted to use DDL but now regret it?"
```

raddevon:
Just completed my review of the guide. Two additional notes:

- I didn't pay attention to line lengths in my suggestions, so those may need to be adjusted if and when the suggestions are accepted.
- I feel like it could use something to wrap it up — maybe just a "Further reading" section with a few relevant links to things like CLI commands and maybe the DDL documentation?

This was really great. I learned a ton reading it and had fun doing it! 👏 

I'll take a look at the other changes in this PR too, but I suspect those will be much lighter if I have anything at all. This is the bulk of my review, I'm sure.

docs/guides/migrations/guide.rst
Dhghomon: Ha! I couldn't resist throwing it in.

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+
+.. code-block::
+
+  ALTER TYPE default::User {
+      DROP PROPERTY name;
+  };
+
+.. note::
+
+    Migrations (and other commands) take even less time if you are
```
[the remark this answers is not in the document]
Dhghomon: Ah, now that I think about it the difference is pretty negligible...if you're not using Windows. On Windows (outside of WSL) the difference is pretty huge (a few seconds vs. instantaneous) so I almost exclusively do migrations while connected. So I'll incorporate the suggestion but change the last sentence a bit.

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+===========================
+
+To see what a schema migration file looks like, type ``edgedb migration 
+create`` and then hit ``y`` to say yes to the question you see below:
+
+
+.. code-block::
+
+    db> did you create object type 'default::User'? [y,n,l,c,b,s,q,?]
+    > y
```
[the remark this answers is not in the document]
Dhghomon: True. Added an explanation: https://github.com/edgedb/edgedb/pull/6656/commits/c1abca92aaf128f2e9ab27881d4e173f6c6d8d6e

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+from this migration script alone is that there is a ``User`` type 
+inside a module called ``default`` that *doesn't* have a property 
+called ``name`` anymore.
+
+.. code-block::
+
+  ALTER TYPE default::User {
+      DROP PROPERTY name;
+  };
+
```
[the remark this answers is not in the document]
Dhghomon: https://github.com/edgedb/edgedb/pull/6656/commits/79862688c02a51ea5b6d6a8189c865b1664ef1a3

docs/guides/migrations/guide.rst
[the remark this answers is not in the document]
Dhghomon: Rewrote to this: https://github.com/edgedb/edgedb/pull/6656/commits/b5fd29cb734238debf9c98fc5ebea6abfb5ff39b

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+finally the ``User`` type which is now able to have a property of 
+type ``Name``.
+
+.. code-block::
+
+    create module default if not exists;
+    create scalar type default::Name extending std::str;
+    create type default::User {
+        create property name: default::Name;
+    };
```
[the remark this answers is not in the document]
Dhghomon: https://github.com/edgedb/edgedb/pull/6656/commits/9d4a07637204e79acc300dac3771cdcb409e14f7

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+
+.. code-block:: sdl
+
+    module default {
+      type SomeType;
+    }
+
+Creating a migration with ``edgedb migration create`` will result 
+in two questions, one to confirm that we wanted to drop the ``name`` 
+property, and another to drop the ``Name`` type.
```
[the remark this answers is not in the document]
Dhghomon: Good catch, definitely should be User. https://github.com/edgedb/edgedb/pull/6656/commits/b973f9a0f697493dd14d6f10263e4088739723a5

docs/guides/migrations/guide.rst
```diff
@@ -0,0 +1,1298 @@
...
+A non-interactive migration will fail if we make changes to our schema 
+that are ambiguous. Let's see if we can make a non-interactive migration 
+fail by doing just that. Delete the most recent ``.edgeql`` migration 
+file again, and change the schema to this:
+
+.. code-block:: sdl
+
+    module default {
+      type User {
+        nam: Name;
```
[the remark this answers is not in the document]
Dhghomon: I originally considered that but it's just changing a name by a letter and (unless there is a way to show a diff of a single letter) that would take a bit longer to parse, I think. In the meantime I've opted for a quick "Can you spot the difference?" before the code block to get the reader's attention. https://github.com/edgedb/edgedb/pull/6656/commits/0b2663eca415758c93e23bc78b7903ce7976b8b7

docs/guides/migrations/guide.rst
Dhghomon: Good idea. Interestingly I recently added migration edit to Windows but didn't know that it also updated the hash.

https://github.com/edgedb/edgedb/commit/e7ff59dbf5b5eed95cb4600c158b0bb8eac35a4e

I've opted for a different order though as I want to walk the reader through the process first before showing the more convenient way to do it. Plus, even though `edgedb migration edit` is easier to *use*, it's still an extra item to remember when getting used to migrations in the first place whereas everyone knows how to edit a file, click save and follow the instructions in the output. So I'd still call it the more advanced concept and put it second.

docs/guides/migrations/guide.rst
Dhghomon: On the subject in general: I've seen EdgeDB Projects in a few areas but agreed that lowercasing is better.

Now with EdgeDB Watch there's a bit of a parsing annoyance for the reader as it's a verb, and also a persistent process that you can refer to.

* You will see EdgeDB watch quickly display
* If you are curious what is happening as EdgeDB watch does its thing, 
* EdgeDB watch will let you know
* EdgeDB watch is best run in a separate instance
* Then EdgeDB watch will suddenly pipe up
* those generated by EdgeDB watch will have a ``MigrationGeneratedBy.DevMode``
* Once you are satisfied with your changes while running EdgeDB watch
* that were made by EdgeDB watch.

A lot of them look for an instant like someone has mistyped "EdgeDB watches" or started typing watch and then opted for another verb.

So now I think I'm leaning towards keeping the capitalization, just because of how easy it is to refer to it and because otherwise you have to rewrite as either the command or "the process created by edgedb watch" (long) or "the process" (vague...what process was that again?).

docs/guides/migrations/guide.rst
Dhghomon: Interesting! Did not know that. https://github.com/edgedb/edgedb/pull/6656/commits/f16ce487aae31069feda0574ec7da1af8aa117fa

docs/guides/migrations/guide.rst
Dhghomon: https://github.com/edgedb/edgedb/pull/6656/commits/7c4494f93ec40c69bda1d94f6ec3eb072e1293a2

(A lowercased Edgedb watch slipped in there, reverted until we decide what to do with it)

docs/guides/migrations/guide.rst
raddevon: I think linking "the section below" is not as nice for accessibility. One mode of traversing the page with a screen reader is to read only the text of each link and allow the user to click on any of the links. "The section below" would not indicate to a screen reader user what they are navigating to. Relative position of the section on the page isn't as meaningful as what the section is about would be, IMO.

docs/guides/migrations/guide.rst
raddevon: I think I would prefer styling as inline code. It would help with the parsing issue but without having to "brand" the feature.

> You will see `edgedb watch` quickly display

> If you are curious what is happening as `edgedb watch` does its thing…

etc.

It might not work at the beginning of headings or inside links (due to RST quirks), but those instances are rare.
>

docs/guides/migrations/guide.rst
Dhghomon: Good idea! https://github.com/edgedb/edgedb/pull/6656/commits/59c9dd8bd7c8fa902f56a60c1239e8a10acad8dc

docs/guides/migrations/guide.rst
Dhghomon: Seems to work. There are only two capitalized instances left, one in the section title which is fine, and the CLI output which we are just faithfully passing on here. (Maybe small PR later on to change the output for the CLI but in the meantime that's what it shows)

https://github.com/edgedb/edgedb/pull/6656/commits/b50232905c2893ff990d23adc7b21e6ddfbdbaeb

docs/guides/migrations/guide.rst
Dhghomon: I don't want to give the impression that using the old method to migrate a schema requires a complete refresh of anything (the lesson to take home is that this is simply what happens unseen to the user every time the CLI does a migration for you), but you've reminded me that the reader won't know how to destroy and restart an instance when necessary. So added that here: https://github.com/edgedb/edgedb/pull/6656/commits/986d069ef6db99e83f49cee35e1c90f1ac6234b2

docs/guides/migrations/guide.rst
Dhghomon: True that it should be more inviting: https://github.com/edgedb/edgedb/pull/6656/commits/f4cea9899aa0da77125e002e770d8a9f3bd15f25

docs/guides/migrations/guide.rst
Dhghomon: https://github.com/edgedb/edgedb/pull/6656/commits/f4cea9899aa0da77125e002e770d8a9f3bd15f25

docs/guides/migrations/guide.rst
Dhghomon: Ah, typo there. I've rephrased a bit while fixing it: https://github.com/edgedb/edgedb/pull/6656/commits/e6b19a539d1d3d1377d7203203d502ad39bd1300

docs/guides/migrations/guide.rst
Dhghomon: True! Pretty long link but got it within the line length limit: https://github.com/edgedb/edgedb/commit/ca7203dadef82941680f11cc887c25c00d997cb3

raddevon (approved):
I pushed a few more changes to titles that we had discussed way back at the beginning. I think this is ready to go!

Dhghomon:
At long last! Nice to see it merged. 🎉

PR #6660, opened 2024-01-03
Note that if..else now supports DML

raddevon:


docs/edgeql/for.rst
```diff
@@ -74,6 +74,10 @@ A similar approach can be used for bulk updates.
 Conditional DML
 ---------------
 
+.. versionadded:: 4.0
+
+    DML is now supported in :eql:op:`if..else` as of EdgeDB 4.0.
+
 DML (i.e., :ref:`insert <ref_eql_insert>`, :ref:`update <ref_eql_update>`,
 :ref:`delete <ref_eql_delete>`) is not supported in :eql:op:`if..else`. If you
```
scotttrinh, line 82: This reads kinda of contradictory still. We're saying it is now supported in >= 4.0 but then saying it is not supported at all. I know the logic works out fine still, but maybe we should mention that it wasn't supported prior to 4.0 to make it super clear that this whole section can be ignored if you are using >= 4.0?

As an aside: I wonder if there is a section of the docs that we should update to show a demonstration of conditional DML as well?

raddevon: We could add an example to the `if..else` docs where we mention this is now possible. There are probably some other places it could go too. We could mention it in `unless conflict` since that is itself a way to do conditional DML.

For this PR though, I'll focus on getting this one change shored up, and we can look at other opportunities to show conditional DML as a separate PR.

raddevon: I think this is clearer. Thanks for the suggestion, @scotttrinh!

scotttrinh (approved):

PR #6666, opened 2024-01-05
Fix some alter+rename combos on pointers in POPULATE MIGRATION

### LINKED ISSUE #6664: Migration to remove property is failing when properties have constraints
I have created a database with below migration applied.

```
start migration to { 
    module default { 
        Type Branch{ 
          property branchName: std::str { 
            constraint min_len_value(0);
            constraint max_len_value(255);
          };
          property branchCode: std::int64;
          property branchURL: std::str { 
            constraint max_len_value(500);
            constraint regexp(r"^(https?|ftp|file)://[-a-zA-Z0-9+&@#/%?=~|!:,.;][-a-zA-Z0-9+&@#/%=~_|]");
            constraint min_len_value(5);
          };
        };
    };
};
populate migration ;
commit migration ;
```

Then I wanted to remove "branchURL" and add some new properties. I applied below migration.

```
start migration to { 
    module default { 
        Type Branch{ 
          property branchName: std::str { 
            constraint min_len_value(0);
            constraint max_len_value(255);
          };
          property branchCode: std::int64;
          property phoneNumber: std::str { 
            constraint min_len_value(5);
            constraint max_len_value(50);
            constraint regexp(r"^(+\d{1,2}\s)?(?\d{3})?[\s.-]\d{3}[\s.-]\d{4}$");
          };
          property email: std::str { 
            constraint min_len_value(5);
            constraint max_len_value(50);
            constraint regexp(r"^[a-zA-Z0-9.!#$%&'+=?^_`{|}~-]+@[a-zA-Z0-9-]+(?:.[a-zA-Z0-9-]+)*$");
          };
        };
    };
};
populate migration ;
commit migration ;
```

This migration gives the error :  `edgedb.InvalidReferenceError: property 'branchURL' does not exist`

If I remove the constraints, the migration works.

Migrating from
```
start migration to { 
    module default { 
        Type Branch{ 
          property branchName: std::str;
          property branchCode: std::int64;
          property branchURL: std::str;
        };
    };
};
populate migration ;
commit migration ;
```

to

```
start migration to { 
    module default { 
        Type Branch{ 
          property branchName: std::str;
          property branchCode: std::int64;
          property phoneNumber: std::str;
          property email: std::str;
        };
    };
};
populate migration ;
commit migration ;
```

msullivan:
Doing multiple operations on a pointer and renaming it at the same
time in a migration created with POPULATE MIGRATION sometimes fails,
because we weren't properly applying renames when generating the AST.

This *didn't* happen when using the normal migration flow, because the
renames get processed incrementally there.

I think I introduced this as a regression in #4007.

I also fixed an issue where in a CREATE MIGRATION script, you could
refer to the old name of something in a later command. The reason I
dealt with that here is because it was preventing schema tests from
finding the bug here, since in the schema test the migration script
was only checked by applying it with CREATE MIGRATION.
(In the actual server, POPULATE MIGRATION explicitly checks
the commands validity in a way not afected by this, which is
where these failures were happening.)

Because of those explicit validity checks, I'm confident that there
aren't auto-generated migrations out there that rely on this bug.

Fixes #6664.

### DIFF
--- a/tests/test_edgeql_ddl.py
+++ b/tests/test_edgeql_ddl.py
@@ -13085,6 +13085,25 @@ CREATE MIGRATION m14i24uhm6przo3bpl2lqndphuomfrtq3qdjaqdg6fza7h6m7tlbra
         #     ['test'],
         # )
 
+    async def test_edgeql_ddl_create_migration_05(self):
+        await self.con.execute('''
+            create type X { create property x -> str; };
+        ''')
+        async with self.assertRaisesRegexTx(
+                edgedb.InvalidReferenceError,
+                "property 'x' does not"):
+            await self.con.execute('''
+                CREATE MIGRATION
+                {
+                    alter type default::X {
+                        alter property x rename to y;
+                    };
+                    alter type default::X {
+                        alter property x create constraint exclusive;
+                    };
+                };
+            ''')
+
     async def test_edgeql_ddl_naked_backlink_in_computable(self):
         await self.con.execute('''
             CREATE TYPE User {
--- a/edb/schema/delta.py
+++ b/edb/schema/delta.py
@@ -2316,8 +2316,15 @@ class ObjectCommand(Command, Generic[so.Object_T]):
         astnode = self._get_ast_node(schema, context)
 
         if astnode.get_field('name'):
-            name = sn.shortname_from_fullname(self.classname)
-            name = context.early_renames.get(name, name)
+            # We need to be able to catch both renames of the object
+            # itself, which might have a long name (for pointers, for
+            # example) as well as an object being referenced by
+            # shortname, if this is (for example) a concrete
+            # constraint and the abstract constraint was renamed.
+            name = context.early_renames.get(self.classname, self.classname)
+            name = sn.shortname_from_fullname(name)
+            if self.classname not in context.early_renames:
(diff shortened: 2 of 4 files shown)


elprans (approved):

tests/test_edgeql_ddl.py
```diff
@@ -13085,6 +13085,25 @@ async def test_edgeql_ddl_create_migration_04(self):
...
+                CREATE MIGRATION
+                {
+                    alter type default::X {
+                        alter property x rename to y;
+                    };
+                    alter type default::X {
+                        alter property x create constraint exclusive;
+                    };
+                };
+            ''')
```
aljazerzen, line 13105: Comments mention a concrete constraint referenced by a short name, based on a renamed abstract constraint.

Can you add a test for that?

aljazerzen (approved):

tests/test_edgeql_ddl.py
msullivan: That's tested by `test_schema_migrations_equivalence_annotation_08`, which was added in https://github.com/edgedb/edgedb/pull/4007, which introduced this bug as a regression. My first take at fixing it broke that one, which is how I realized it was a regression

PR #6667, opened 2024-01-05
Add force_database_error documentation

Dhghomon:
Writing documentation for the UI today and noticed that force_database_error takes a str instead of a bool, and that the strs it accepts can be pretty verbose. Pretty nice feature to have compared to a simple bool so added two examples of valid input to get users started.


raddevon (approved):
Nice catch! Would it also make sense to provide a link to a list of errors? Maybe https://github.com/edgedb/edgedb/blob/master/edb/errors/__init__.py

Dhghomon:
> Nice catch! Would it also make sense to provide a link to a list of errors? Maybe https://github.com/edgedb/edgedb/blob/master/edb/errors/__init__.py

Yeah! Good call, I linked to the page containing them. (There's one in .rst format too)

PR #6674, opened 2024-01-05
Reduce the size of data for analyze tests.

### LINKED ISSUE #6316: Try using `enable_seqscan=off` to guide index analyze tests
We have a bunch of tests that use ANALYZE to make sure that we can hit the index in various cases.

Unfortunately creating tables big enough that we feel confident postgres will use the indexes is slow and it *still* flakes sometimes (https://github.com/edgedb/edgedb/actions/runs/6567901995/job/17841774091?pr=6314)

I think that disabling seqscans in some of the tests where we are trying to verify index use might make them less flaky and let us make the tests faster.

We could expose `enable_seqscan` as a config var (testmode only, probably), but I think a simpler approach would be to have the analyze test suite create a function that disables them, that works like `pgvector::set_probes` does:
```
create function _disable_seqscan() -> std::int64 {
    using sql $$
        select 1 from (
            select set_config('enable_seqscan', 'off'::text, true)
        ) as dummy;
    $$;
};
```
(That way we can easily cherry-pick the test changes without needing to patch in a config)

vpetrovykh:
Instead of generating a large dataset to force usage of indexes, disable sequential scan and use a smaller data set.

Fixes #6316

### DIFF
--- a/tests/test_edgeql_explain.py
+++ b/tests/test_edgeql_explain.py
@@ -29,7 +29,7 @@ from edb.server.compiler.explain import pg_tree
 
 
 class TestEdgeQLExplain(tb.QueryTestCase):
-    '''Tests for EXPLAIN.
+    '''Tests for ANALYZE.
 
     This is a good way of testing explain functionality, but also this can be
     used to test indexes.
@@ -54,8 +54,12 @@ class TestEdgeQLExplain(tb.QueryTestCase):
             data, shape, fail=self.fail, message=message)
 
     async def explain(self, query, *, execute=True, con=None):
+        con = (con or self.con)
+        # Disable sequential scan so that we hit the index even on small
+        # datasets.
+        await con.query_single('select _set_seqscan("off")')
         no_ex = '(execute := False) ' if not execute else ''
-        return json.loads(await (con or self.con).query_single(
+        return json.loads(await con.query_single(
             f'analyze {no_ex}{query}'
         ))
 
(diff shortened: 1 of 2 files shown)


tests/test_edgeql_explain.py
```diff
@@ -54,8 +54,12 @@ def assert_plan(self, data, shape, message=None):
             data, shape, fail=self.fail, message=message)
 
     async def explain(self, query, *, execute=True, con=None):
+        con = (con or self.con)
+        # Disable sequential scan so that we hit the index even on small
+        # datasets.
+        await con.query_single('select _set_seqscan("off")')
         no_ex = '(execute := False) ' if not execute else ''
-        return json.loads(await (con or self.con).query_single(
+        return json.loads(await con.query_single(
```
msullivan, line 62: Should we make it configurable whether we turn of seqscan? Maybe some of tests could leave it on which would allow us to reduce test churn. Maybe not really, though.

msullivan (approved):
Great! Maybe too much test churn to cherry-pick it cleanly, though we will see.

How quick is startup now?

vpetrovykh:
On my machine the startup for just the analyze tests went from ~55 seconds to ~25 seconds. For comparison bootstrapping the "edgeql_select" tests takes about 23 seconds. So I think this cut down the bootstrap time to about as low as it gets.

tests/test_edgeql_explain.py
vpetrovykh: For analyze tests leaving it on doesn't really help us, I think. If in the future we run into analyze tests that specifically need seqscan on then we can make it configurable.

vpetrovykh: To be clear regarding the test churn, part of this is also reducing the number of tests objects created dramatically. So the results of the original tests would likely not align even if you turned the seqscan on for them.

PR #6706, opened 2024-01-18
Fix issue with schema changes when trigger depends on a rewrite

msullivan:
Dependencies on rewrites, policies, and triggers don't get explicitly
tracked as references, but the things they depend on transitively
through those do.

The dummy expression for rewrites is not the correct type, so if a
trigger and a rewrite are being reprocessed in _propagate_if_expr refs
at the same time, and the trigger uses the rewrite, and is processed
first, then it will error.
Fix this by having the rewrite dummy expression have the right type.

I've experimented with having rewrites and triggers explicitly tracked
as dependencies, but it complicates some code and I don't think it
actually fixes any real problems.

### DIFF
--- a/edb/schema/rewrites.py
+++ b/edb/schema/rewrites.py
@@ -36,6 +36,9 @@ from . import referencing
 from . import schema as s_schema
 from . import types as s_types
 
+if TYPE_CHECKING:
+    from . import pointers as s_pointers
+
 
 class Rewrite(
     referencing.NamedReferencedInheritingObject,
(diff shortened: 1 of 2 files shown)


edb/schema/rewrites.py
```diff
@@ -223,7 +233,9 @@ def get_dummy_expr_field_value(
         value: Any,
     ) -> Optional[s_expr.Expression]:
         if field.name == 'expr':
-            return s_expr.Expression(text='false')
+            pt = self.scls.get_ptr_target(schema)
+            text = f'<{pt.get_displayname(schema)}>{{}}'
```
aljazerzen, line 237: So this is just a cast of an empty set to the correct type?

aljazerzen (approved):

edb/schema/rewrites.py
msullivan: Yeah

PR #6714, opened 2024-01-19
DML in cast_expr

### LINKED ISSUE #5050: Support DML in cast_expr
If I recall correctly, we put off supporting DML in cast_expr because cast_expr is executed per *target*, not per *source*, and `(source, target)` on a multi link may not be unique, which made it difficult to join things back up if they were split into multiple CTEs.

I think we can solve this problem by using the row's `ctid` as the join key.

If that doesn't work for some reason, then I think we should be able to support it for links (but not properties) by using `(source, target)`. (Supporting it for properties is pretty marginal from a feature perspective, but there is a big win to not needing separate code paths.)

Follow up on #3592.

aljazerzen:
Closes #5050

### DIFF
--- a/edb/pgsql/compiler/relctx.py
+++ b/edb/pgsql/compiler/relctx.py
@@ -767,18 +767,20 @@ def create_iterator_identity_for_path(
     ctx: context.CompilerContextLevel,
     apply_volatility: bool=True,
 ) -> None:
-
-    id_expr = pgast.FuncCall(
-        name=('edgedb', 'uuid_generate_v4'),
-        args=[],
-    )
-
     if isinstance(stmt, pgast.SelectStmt):
         path_id = pathctx.map_path_id(path_id, stmt.view_path_id_map)
         if apply_volatility:
             apply_volatility_ref(stmt, ctx=ctx)
 
-    pathctx.put_path_var(stmt, path_id, id_expr, force=True, aspect='iterator')
+    already_exists = isinstance(stmt, pgast.Query) and (
+        maybe_get_path_rvar(stmt, path_id, aspect='iterator', ctx=ctx) != None
+    )
+    if not already_exists:
+        id_expr = pgast.FuncCall(
+            name=('edgedb', 'uuid_generate_v4'),
+            args=[],
+        )
+        pathctx.put_path_var(stmt, path_id, id_expr, aspect='iterator')
     pathctx.put_path_bond(stmt, path_id, iterator=True)
 
 
--- a/edb/pgsql/delta.py
+++ b/edb/pgsql/delta.py
@@ -4899,13 +4899,6 @@ class PointerMetaCommand(
                 orig_schema, pointer, env=None
             )
 
-        # refs = irutils.get_longest_paths(ir.expr)
-        # ref_tables = schemamech.get_ref_storage_info(ir.schema, refs)
-        # local_table_only = all(
-        #     t == old_ptr_stor_info.table_name
-        #     for t in ref_tables
-        # )
-
         ptr_path_id = tgt_path_id.ptr_path()
         src_path_id = ptr_path_id.src_path()
         assert src_path_id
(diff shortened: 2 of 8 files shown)


edb/pgsql/compiler/relctx.py
```diff
@@ -749,9 +749,13 @@ def create_iterator_identity_for_path(
         if apply_volatility:
             apply_volatility_ref(stmt, ctx=ctx)
 
-    pathctx.put_path_var(stmt, path_id, id_expr, force=True, aspect='iterator')
     pathctx.put_path_bond(stmt, path_id, iterator=True)
 
+    if (path_id, 'iterator') in stmt.path_rvar_map:
+        return
+
+    pathctx.put_path_var(stmt, path_id, id_expr, aspect='iterator')
```
aljazerzen, line 757: @msullivan I'm not sure if this is the correct way to do it.

I had a problem where DML within iteration was producing this:
```SQL
    WITH
    source_rel AS
        ((
            SELECT *, ctid FROM "the_link_table"
            -- this ctid should be used as the iterator
        )),
    "iter~1" AS
        ((
            FROM source_rel AS "source_rel~1"
            CROSS JOIN LATERAL (
                SELECT/*<pg.SelectStmt at 0x7f82c0915f10>*/
                    "source_rel~1"."hell" AS "hell_value~1",
                    edgedb.uuid_generate_v4() AS "hell_iterator~1"
                    -- this uuid_generate call should not be here.
                    -- ctid should be used instead.
                    -- I inject its path and rel vars so it would be picked up
                    -- but the code above overriddes it.
                    -- So I've disabled that.
                WHERE
                    ("source_rel~1"."hell" IS NOT NULL)
            ) AS "Hello~3"
            SELECT/*<pg.SelectStmt at 0x7f82c0914110>*/
                "Hello~3"."hell_iterator~1" AS "hell_iterator~2"
        )),
    "ins_contents~1" AS
        (...)
    ...
```

edb/pgsql/delta.py
```diff
@@ -4850,59 +4864,79 @@ def _compile_conversion_expr(
...
-                external_rvars[src_path_id, 'source'] = src_rvar
+                src_rel.path_outputs[(tgt_path_id, 'iterator')] = \
+                    pgast.ColumnRef(name=('id',))
+
+            external_rels[ptr_path_id] = \
+                (src_rel, ('source', 'value'))
+            external_rels[tgt_path_id] = \
+                (src_rel, ('identity', 'source', 'value', 'iterator'))
+
+        #     if ptr_table:
```
aljazerzen, line 4905: So I have the following DDL:
```
alter type Base {
    alter link foo {
        alter property bar set type int32;
    };
};
```

I set `external_rels` for `Base`, `Base::foo` and `Base::foo@`.
But on lookup for set with path_id `(default::Base).>foo[IS default::Child]@bar[IS std::int64]`, I'm not finding it in `external_rels`.

This should happen here: https://github.com/edgedb/edgedb/blob/7a145a9596fd21c4d5a330839459947e0582e79a/edb/pgsql/compiler/relgen.py#L426-L428

edb/pgsql/compiler/relctx.py
msullivan: This seems plausible, though I'll ponder it a bit.
(I would restructure it so that the put_path_var is in the conditional instead of having an early return.)

edb/pgsql/delta.py
msullivan: Huh, you mean `external_rels` is getting cleared somewhere?

edb/pgsql/delta.py
```diff
@@ -4850,59 +4865,60 @@ def _compile_conversion_expr(
...
-                external_rvars[ptr_path_id, 'source'] = rvar
-                external_rvars[ptr_path_id, 'value'] = rvar
-                external_rvars[src_path_id, 'identity'] = rvar
-                external_rvars[tgt_path_id, 'identity'] = rvar
-                if local_table_only and not is_lprop:
-                    external_rvars[src_path_id, 'source'] = rvar
-                    external_rvars[src_path_id, 'value'] = rvar
-                elif is_lprop:
-                    external_rvars[tgt_path_id, 'value'] = rvar
+                src_rel.path_outputs[(ptr_path_id, 'identity')] = \
```
msullivan, line 4891: Hm, I'm a little nervous about making ctid an `identity` aspect; what needs that?

aljazerzen:
I give up. I might come back when I understand how our backend compiler works better.

PR #6717, opened 2024-01-19
Fix some postgres submodule check issues

msullivan:


.github/workflows/pull-request-meta.yml
```diff
@@ -19,12 +19,13 @@ jobs:
...
 
     - name: Verify that postgres/ was not changed unintentionally
+      env:
+        PR_TITLE: ${{ github.event.pull_request.title }}
       shell: bash
       run: |
         required_prefix="Update bundled PostgreSQL"
 
-        title="${{ github.event.pull_request.title }}"
-        if [[ $title == $required_prefix* ]]; then
```
aljazerzen, line 27: Ohhh, this was not quoted so it was being interpreted as commands. Strange that it didn't happen before with my PR titles.

I shouldn't have done this in bash.

aljazerzen (approved):

.github/workflows/pull-request-meta.yml
msullivan: The quoting around $title was my first suspicion as well, but that wasn't really it. It not being quoted means it will be expanded as *multiple* arguments to `[[`, which might be fine, though I quoted it to be safe.

The big issue was that backquotes are evaluated inside of double quotes, so when I had a PR title with backquotes in it, it got executed.
Fix was to pass it in as an env var.

PR #6720, opened 2024-01-22
Add Slack OpenID Connect provider

scotttrinh:
Apologies for the little bit of formatting noise in the test file. I decided it was worth just leaving it to avoid diffs in the future.


edb/lib/ext/auth.edgeql
```diff
@@ -145,6 +145,17 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
...
     };
 
+    create type ext::auth::SlackOAuthProvider
+        extending ext::auth::OAuthProviderConfig {
+        alter property name {
+            set default := 'builtin::oauth_slack';
+        };
+
+        alter property display_name {
+            set default := 'slack';
```
jaclarke, line 155:
```suggestion
            set default := 'Slack';
```

jaclarke (approved):

PR #6721, opened 2024-01-23
Branches, and fast branch creation

msullivan:
The idea is to use pg_dump. This actually requires invoking pg_dump three times, and making some tweaks to the SQL before executing it, in order to get all the pieces in. This is so that we can apply it on top of the template.

I also tried just doing a full dump/restore onto an empty database, which was much simpler to implement but was a lot slower.


edb/edgeql/ast.py
```diff
@@ -872,6 +872,7 @@ class DatabaseCommand(ExternalObjectCommand):
 class CreateDatabase(CreateObject, DatabaseCommand):
 
     template: typing.Optional[ObjectRef] = None
+    mode: str
```
elprans, line 875: Make it an enum?

edb/server/bootstrap.py
```diff
@@ -1035,6 +1036,109 @@ def prepare_patch(
...
+    data_arg = ['--schema=edgedbpub'] if mode == 'data' else []
+    dump_args = [
+        '--data-only',
+        '--schema=edgedbstd',
+        *data_arg,
+        '--disable-triggers',
+        # We need to use --inserts so that we can use --on-conflict-do-nothing
+        '--inserts',
+        '--rows-per-insert=100',
+        '--on-conflict-do-nothing',
```
elprans, line 1128: Can you add an explanation why this is necessary please?

edb/server/protocol/execute.pyx
```diff
@@ -197,10 +204,10 @@ async def execute(
     finally:
         if query_unit.drop_db:
             tenant.allow_database_connections(query_unit.drop_db)
-        if query_unit.create_db_template:
-            tenant.allow_database_connections(
-                query_unit.create_db_template,
-            )
+        # if query_unit.create_db_template:
```
elprans, line 207: What's up here?

edb/server/tenant.py
```diff
@@ -783,7 +783,7 @@ async def _acquire_intro_pgcon(
         try:
             conn = await self.acquire_pgcon(dbname)
         except pgcon_errors.BackendError as e:
-            if e.code_is(pgcon_errors.ERROR_INVALID_CATALOG_NAME):
+            if e.code_is(pgcon_errors.ERROR_INVALID_CATALOG_NAME) or True:
```
elprans, line 786: Dont wanna keep this

edb/server/tenant.py
```diff
@@ -1175,13 +1175,43 @@ async def on_before_drop_db(
...
+                    self._cluster,
+                    self._server._refl_schema,
+                    con,
+                    real_src_dbname,
+                    real_tgt_dbname,
+                    mode,
+                    self._server._sys_queries['backend_id_fixup'],
+                )
+        except Exception:
+            # Is this the right place (no)??
```
elprans, line 1207: Probably better to handle a level above

edb/server/tenant.py
msullivan: Oh, yeah. That was there because in the initial take, we created a totally empty database, and if the copy failed and then the remove didn't happen, we'd end up with a database that couldn't start up, so you couldn't even clean it up from edgedb. Now, though, since we still use the template, that's not a problem.

msullivan:
Alright I think I've mostly get this in shape. If there is some straightforward better way to do the config dumping, though, that could be good.

msullivan:
I came up with two new approaches for the config copying, and there are patches for both.

elprans (approved):

PR #6735, opened 2024-01-25
nix code block -> bare code block

raddevon:
We seemingly don't support nix syntax highlighting. It breaks the build.


scotttrinh:
Nix syntax highlighting _should_ work (added in Pygments 2.0 https://pygments.org/docs/lexers/#pygments.lexers.nix.NixLexer) but maybe there is some syntax issue in that nix codeblock?

aljazerzen (approved):

PR #6736, opened 2024-01-25
Docs: Add `nix` language code

raddevon:


raddevon:
Putting this on hold for a bit. Even though this allows us to use `code-block:: nix`, it does not actually result in syntax highlighting in the rendered output.

PR #6737, opened 2024-01-25
Bump test parallelization in CI to 4

elprans:
Github now runs 4-core VMs by default.


fantix (approved):

elprans:
Somehow the run here is slower than with `-j2`.

PR #6751, opened 2024-01-30
Use a consistent interface for `ext::auth` errors

scotttrinh:
When we return HTTP responses for unsuccessful requests, we now return a simplified and consistent interface that includes the name of the exception being caught and serialized to JSON.

In our auth helper libraries, we can use this to instantiate equivalent errors to allow consumers to more easily catch classes of errors.

### DIFF
--- a/edb/server/protocol/auth_ext/http.py
+++ b/edb/server/protocol/auth_ext/http.py
@@ -130,60 +130,55 @@ class Router:
                 case _:
                     raise errors.NotFound("Unknown auth endpoint")
 
+        # User-facing errors
         except errors.NotFound as ex:
             _fail_with_error(
                 response=response,
                 status=http.HTTPStatus.NOT_FOUND,
-                message=str(ex),
-                ex_type=edb_errors.ProtocolError,
+                ex=ex,
             )
 
         except errors.InvalidData as ex:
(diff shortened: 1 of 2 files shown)


edb/server/protocol/auth_ext/http.py
```diff
@@ -192,8 +187,7 @@ async def handle_request(
             _fail_with_error(
                 response=response,
                 status=http.HTTPStatus.INTERNAL_SERVER_ERROR,
-                message=str(ex),
-                ex_type=edb_errors.InternalServerError,
+                ex=ex,
```
jaclarke, line 190: Do we need to wrap this catch-all error in an `InternalServerError`, so it gets serialised with that type name?

scotttrinh: I'm not sure, to be honest. I guess that would be consistent though, so if that makes sense to you, I can do that!

scotttrinh: @jaclarke ⚡ [`892b5ba` (#6751)](https://github.com/edgedb/edgedb/pull/6751/commits/892b5ba759f83f79bc974b52c2be64967a45091c)

jaclarke (approved):

PR #6752, opened 2024-01-30
Bump proc-macro2 to fix build under recent nightly Rust

elprans:


elprans:
> We build in recent nightly Rust?

I do locally.

PR #6753, opened 2024-01-30
Don't unconditionally populate 'iterator' aspect in include_rvar

msullivan:
That causes us to have lots of incorrect 'iterator' aspects in rvar
maps.  Instead, just properly populate the aspects explicitly at the
one relevant call site.

This might help with #6714.

### DIFF
--- a/edb/pgsql/compiler/dml.py
+++ b/edb/pgsql/compiler/dml.py
@@ -382,7 +382,9 @@ def merge_iterator(
         put_iterator_bond(iterator, select)
         relctx.include_rvar(
             select, iterator_rvar,
-            aspects=('source', 'value', iterator.aspect),
+            aspects=('value', iterator.aspect) + (
+                ('source',) if iterator.path_id.is_objtype_path() else ()
+            ),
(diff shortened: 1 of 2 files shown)


edb/pgsql/compiler/dml.py
```diff
@@ -382,6 +382,7 @@ def merge_iterator(
         put_iterator_bond(iterator, select)
         relctx.include_rvar(
             select, iterator_rvar,
+            aspects=('source', 'value', iterator.aspect),
```
aljazerzen, line 385: This was changed from `'value', ('source' if some condition)`, to `'value', 'source', 'iterator/identity`.

Are you sure that this `source` condition is not needed anymore?

aljazerzen (approved):

edb/pgsql/compiler/dml.py
msullivan: I'm pretty confident it isn't a bug right now, but actually it is kind of sloppy in the same way that that iterator sloppiness led to your trouble, so...

PR #6754, opened 2024-01-31
Move app details configuration into `AuthConfig`

scotttrinh:
We need these details in more places (email templates, WebAuthn, etc) so copy the existing `UIConfig` properties into the base `AuthConfig` object.

For now, we have the old properties still there. We are going to discuss if (and how) to migrate any existing user provided values from the old config into the new config as part of the upgrade separately. Going to leave this as a draft for now while we discuss.

(n.b. some Black formatting churn, apologies)


msullivan (approved):
This seems solid. I need to look into making the migration smooth

PR #6756, opened 2024-01-31
Formatting of stack tracebacks in tests

aljazerzen:
Make tracebacks in error reporting look like this:
```
  /home/aljaz/EdgeDB/edgedb/tests/test_sourcecode.py:111, in test_cqa_mypy
    raise AssertionError(
```
instead of this:
```
  File "/home/aljaz/EdgeDB/edgedb/tests/test_sourcecode.py", line 111, in test_cqa_mypy
    raise AssertionError(
```

### DIFF
--- a/edb/common/traceback.py
+++ b/edb/common/traceback.py
@@ -1,0 +1,192 @@
+#
+# This source file is part of the EdgeDB open source project.
+#
+# Copyright 2009-present MagicStack Inc. and the EdgeDB authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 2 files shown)


edb/common/traceback.py
```diff
@@ -0,0 +1,192 @@
...
+# limitations under the License.
+#
+# mypy: disable-error-code="attr-defined"
+
+"""
+Provides stack trace formatting that prints `{filename}:{line}`, instead of
+`"{filename}", line {line}`.
+
+Stolen from Python's traceback module.
+"""
```
msullivan, line 25: Since this code is taken from CPython, we need a relevant cpython copyright notice.

https://github.com/edgedb/edgedb/blob/master/edb/tools/profiling/tracing_singledispatch.py#L4-L5 is what we do in some existing code adapted from CPython.

msullivan (approved):
Seems reasonable, and is probably the most principled way to do this.

Should we consider just running a regex over the output instead, though?

PR #6760, opened 2024-02-01
Supress idle tx timeout during compilation

### LINKED ISSUE #6758: We should suspend the pg transaction timeout while waiting for the compiler
If there is a long running compile job (probably computing a migration?), we can spuriously trigger an idle-in-transaction timeout, since even though the transaction is not idle from edgedb's perspective, it *is* from postgres's.

msullivan:
I tested this by adding a sleep when the variable `slowcompile` is used. 
Fixes #6758.

### DIFF
--- a/edb/server/protocol/binary.pyx
+++ b/edb/server/protocol/binary.pyx
@@ -517,7 +517,7 @@ cdef class EdgeConnection(frontend.FrontendConnection):
         async with self.with_pgcon() as conn:
             await conn.sql_execute(f'''
                 select pg_catalog.set_config(
-                    'idle_in_transaction_session_timeout', {timeout}, false)
+                    'idle_in_transaction_session_timeout', {timeout}, true)
             '''.encode('utf-8'))
 
     async def _parse(
--- a/edb/server/protocol/binary.pyx
+++ b/edb/server/protocol/binary.pyx
@@ -503,7 +503,7 @@ cdef class EdgeConnection(frontend.FrontendConnection):
         async with self.with_pgcon() as conn:
             await conn.sql_execute(b'''
                 select pg_catalog.set_config(
-                    'idle_in_transaction_session_timeout', '0', false)
+                    'idle_in_transaction_session_timeout', '0', true)
             ''')
 
     async def _restore_tx_timeout(self, dbview.DatabaseConnectionView dbv):
--- a/edb/server/protocol/binary.pyx
+++ b/edb/server/protocol/binary.pyx
@@ -552,7 +552,7 @@ cdef class EdgeConnection(frontend.FrontendConnection):
 
         # If we have to do a compile within a transaction, spin up a
         # heartbeat task to keep the SQL connection from going idle.
-        if dbv.in_tx() and query_unit_group is None:
+        if dbv.in_tx() and not dbv.in_tx_error() and query_unit_group is None:
             event = asyncio.Event()
             heartbeat_task = self.tenant.create_task(
                 self._heartbeat(event), interruptable=False)
--- a/edb/server/protocol/binary.pyx
+++ b/edb/server/protocol/binary.pyx
@@ -499,27 +499,26 @@ cdef class EdgeConnection(frontend.FrontendConnection):
         else:
             return edgeql.NormalizedSource.from_string(text)
 
-    async def _heartbeat(self, stop_event):
-        """A heartbeat function that keeps the connection from going idle.
+    async def _suppress_tx_timeout(self):
+        async with self.with_pgcon() as conn:
+            await conn.sql_execute(b'''
+                select pg_catalog.set_config(
+                    'idle_in_transaction_session_timeout', '0', false)
+            ''')
 
-        We run this while we are running the compiler from within a
-        transaction, to keep the postgres transaction from going idle
-        during long-running compiles.
-
-        The heartbeat duration is tuned for the default 10s timeout.
-        """
-        while True:
-            conn = await self.get_pgcon()
-            try:
-                await conn.sql_execute(b'SELECT 1')
-            finally:
-                self.maybe_release_pgcon(conn)
-
-            try:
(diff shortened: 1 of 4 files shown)


msullivan:
As an aside, I wasted a bunch of time thinking about whether it would be more elegant to use `cancel` to stop the heartbeat task instead of signalling a separate event. In theory, yes, but in practice it is very hairy, because unless I missed something, there is no simple way to *defer* a cancellation until after some block of code is complete. There is `asyncio.shield`, but it will leave the inner task running asynchronously.
So I wound up writing this code:

```
async def deferred_shield(arg):
    '''Wait for a future, deferring cancellation until it is complete.

    If you do
        await deferred_shield(something())

    it is approximately equivalent to
        await something()

    except that if the coroutine containing it is cancelled,
    something() is protected from cancellation, and *additionally*
    CancelledError is not raised in the caller until something()
    completes.

    This can be useful if something() contains something that
    shouldn't be interrupted but also can't be safely left running
    asynchronously.
    '''
    task = asyncio.create_task(arg)

    ex = None
    while not task.done():
        try:
            await asyncio.shield(task)
        except asyncio.CancelledError as cex:
            ex = cex
        except Exception:
            if ex:  # Or would we rather raise the actual exception?
                raise ex
            raise

    if ex:
        raise ex
    return task.result()
```

which is cute, and might be useful for someone someday, but I decided to just use an event.

edb/server/protocol/binary.pyx
```diff
@@ -499,6 +499,28 @@ cdef class EdgeConnection(frontend.FrontendConnection):
         else:
             return edgeql.NormalizedSource.from_string(text)
 
+    async def _heartbeat(self, stop_event):
+        """A heartbeat function that keeps the connection from going idle.
```
elprans, line 503: Probably a better fix would be to just set `idle_session_timeout` to `0` while the compiler is running?  Possibly delayed (to, say, 1/2 of the timeout in the future) so that fast compiles don't trigger it.  Postgres allows transaction-local configuration so this should be safe even if restoration of the original timeout fails somehow.

msullivan: The problem with the possibly delayed approach is that if a bunch of the timeout has already elapsed when the query is sent, then we can still timeout while compiling. That's why my version does the first select right away, instead of delayed.

Part of my rationale for going with this version is that then, the common case is one extra query and not two.

We can do it by clearing idle_session_timeout, though

1st1: > Part of my rationale for going with this version is that then, the common case is one extra query and not two.

It's only affecting queries in transaction blocks so even 2 extra quick queries might be OK.

msullivan:
Alright we'll do it the other way

edb/server/protocol/binary.pyx
```diff
@@ -499,6 +499,27 @@ cdef class EdgeConnection(frontend.FrontendConnection):
         else:
             return edgeql.NormalizedSource.from_string(text)
 
+    async def _suppress_tx_timeout(self):
+        async with self.with_pgcon() as conn:
+            await conn.sql_execute(b'''
+                select pg_catalog.set_config(
+                    'idle_in_transaction_session_timeout', '0', false)
```
elprans, line 506: You probably want the opposite?

```suggestion
                    'idle_in_transaction_session_timeout', '0', true)
```

edb/server/protocol/binary.pyx
```diff
@@ -526,7 +547,21 @@ cdef class EdgeConnection(frontend.FrontendConnection):
...
-        return await dbv.parse(query_req)
+        query_unit_group = dbv.lookup_compiled_query(query_req)
+
+        # If we have to do a compile within a transaction, suppress
+        # the idle_in_transaction_session_timeout.
+        suppress_timeout = (
+            dbv.in_tx() and not dbv.in_tx_error() and query_unit_group is None
+        )
+        if suppress_timeout:
+            await self._suppress_tx_timeout()
```
elprans, line 558: What do we think about doing a `loop.call_at` half the current timeout here instead?  Too fiddly?

msullivan: I think we shouldn't do that because:
> The problem with the possibly delayed approach is that if a bunch of the timeout has already elapsed when the query is sent, then we can still timeout while compiling.

edb/server/protocol/binary.pyx
msullivan: Sure. I don't think it should matter in practice, since there shouldn't be any way for this set_config to ever get committed

msullivan:
```
#!/usr/bin/env python3

import time

import edgedb

def run():
    client = edgedb.create_client(
        port=5656, database='edgedb', tls_security='insecure'
    )

    loop = 0
    for tx in client.transaction():
        with tx:
            tx.query('select 1')
            time.sleep(8)
            loop += 1
            print('LOOP!')

            tx.query('''
                with slowcompile := 10, select slowcompile
            ''')

    assert loop == 1

run()
```
was how I tested this, by the way

edb/server/protocol/binary.pyx
```diff
@@ -499,6 +499,27 @@ cdef class EdgeConnection(frontend.FrontendConnection):
...
+            'session_idle_transaction_timeout',
+        )
+        timeout = (
+            'NULL' if not old_timeout
+            else repr(old_timeout.value.to_backend_str())
+        )
+        async with self.with_pgcon() as conn:
+            await conn.sql_execute(f'''
+                select pg_catalog.set_config(
+                    'idle_in_transaction_session_timeout', {timeout}, false)
```
elprans, line 520: Should probably also be `true` here for symmetry.

msullivan: Hm, though that setting is *not* transaction local. It should persist when the transaction is committed. It looks like that does still work if it is true (because if it was configured earlier in the transaction, that config wasn't local), but that seems more subtle

elprans (approved):

PR #6762, opened 2024-02-01
Branch

vpetrovykh:
Add SDL comparisons to branch tests.

Fix an SDL tracing bug that caused extending scalar types to sometimes fail.

### DIFF
--- a/edb/testbase/server.py
+++ b/edb/testbase/server.py
@@ -1542,12 +1542,14 @@ class StableDumpTestCase(QueryTestCase, CLITestCaseMixin):
 
         orig_branch = self.get_database_name()
         new_branch = f'new_{orig_branch}'
+        # record the original schema
+        orig_schema = await self.con.query_single('describe schema as sdl')
 
         # connect to a default branch so we can create a new branch
-        branch_type = 'DATA' if include_data else 'SCHEMA'
+        branch_type = 'data' if include_data else 'schema'
         await self.con.execute(
-            f'CREATE {branch_type} BRANCH {new_branch} '
-            f'FROM {orig_branch}'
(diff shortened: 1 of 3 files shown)


edb/testbase/server.py
```diff
@@ -1559,6 +1561,22 @@ async def check_branching(self, include_data=False, *, check_method):
...
+            # original schema should have the "complete" status right away.
+            await self.con.execute(f'start migration to {{ {orig_schema} }}')
+            mig_status = json.loads(
+                await self.con.query_single_json(
+                    'describe current migration as json'
+                )
+            )
+            self.assertTrue(mig_status.get('complete'))
+            # cancel the migration
+            await self.con.execute('rollback')
```
msullivan, line 1578: Do `abort migration` instead of `rollback`. `rollback` will rollback an enclosing connection too; that doesn't really matter here, but it's cleaner

edb/testbase/server.py
```diff
@@ -1559,6 +1561,22 @@ async def check_branching(self, include_data=False, *, check_method):
...
+            # compare the new branch schema to the original. We expect there
+            # to be no difference and therefore a new migration to the
+            # original schema should have the "complete" status right away.
+            await self.con.execute(f'start migration to {{ {orig_schema} }}')
+            mig_status = json.loads(
+                await self.con.query_single_json(
+                    'describe current migration as json'
+                )
+            )
+            self.assertTrue(mig_status.get('complete'))
```
msullivan, line 1576: Good strategy

msullivan (approved):
Looks great

edb/testbase/server.py
vpetrovykh: Fixed

Issue #6767, opened 2024-02-01
Coalesce `??` doesn't evaluate link's existence


CarsonF:
- EdgeDB Version: 4.5+641a8f3
- EdgeDB CLI Version: 4.0.2+500be79

Steps to Reproduce:

1. Schema
```esdl
module default {
  type User {
    primary: Content;
    secondary: Content;
  };
  type Content {
    name: str
  };
}
```
2. Some data
```edgeql
insert User {
  primary := (insert Content)
};
insert User {
  secondary := (insert Content)
};
```
3. Bug
```edgeql
select User {
  coalesce := (.primary ?? .secondary),
  conditional := (if exists .primary then .primary else .secondary)
}
```
The `coalesce` property returns empty if there's no `primary`, ignoring `secondary`.
The `conditional` correctly picks `primary` or `secondary`.

![Screenshot 2024-02-01 at 9 08 45 AM](https://github.com/edgedb/edgedb/assets/932566/fa7ec3e3-20f6-41cd-8e4d-700b9842a915)

msullivan:
The history here is weird: it looks like this was a 3.0 regression, but also it seems to be fixed on master...

msullivan:
Oh and to complicate it even more, it seems to be working on my local builds of 4.5, even though I was able to reproduce it on a 4.5 release build.

msullivan:
OK, actually, that analysis was basically all wrong: the bug shows up only when there is an implicit limit set for how many rows get returned.
I have that disabled typically, which was what was throwing me off.
(`\set limit 0`)

msullivan:
Oh, no, it *is* still fixed on master

msullivan:
It looks like it was fixed by https://github.com/edgedb/edgedb/pull/6747, which is already slated for backporting. I need to do some thinking about whether that makes sense to me that it is fixed.

msullivan:
Alright, I am convinced it makes sense. The extra scope introduced by the implicit limit (or by adding some explicit extra wrapping) was triggering the issue fixed in #6747. I'm going to close this as fixed, and we'll cherry-pick the test I'm adding in https://github.com/edgedb/edgedb/pull/6771 along with the fix.

CarsonF:
So fix will be released in 4.6?

PR #6771, opened 2024-02-01
Add a test for #6767

msullivan:
It passes on master already, though fails on 3.x and 4.x.


msullivan:
It looks like the issue here was fixed by https://github.com/edgedb/edgedb/pull/6747

elprans (approved):

PR #6772, opened 2024-02-02
Fix some major issues with the error context printer

msullivan:
There were two major problems:
 * The line immediately before the error was always skipped in the context
 * The long line truncating, which would try to show the contents around
   the error column, would try to do that when truncating context lines too,
   which made no sense and meant that truncated context lines often
   contained newlines.

Additionally, things were complicated by doing a search for each line
instead of doing one search for the error line and indexing from it.

I decided to just get rid of all the truncation logic, since it would mostly
just serve to impede debugging, though it would be possible to do
something more correct than what we do know. (Though I'd bump the
limit up from 120, which is very low.)

Hopefully now this will actually be useful in debugging bootstrap SQL
failures.

### DIFF
--- a/edb/common/context.py
+++ b/edb/common/context.py
@@ -90,10 +90,6 @@ class ParserContext(markup.MarkupExceptionContext):
     def as_markup(cls, self, *, ctx):
         me = markup.elements
 
-        body = []
-
-        lines = []
-        line_numbers = []
         start = self.start_point
         # TODO: do more with end?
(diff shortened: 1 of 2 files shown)


vpetrovykh:
LGTM. It did actually show the error in the bootstrap SQL clearly, so I guess it works for the intended purpose :)

vpetrovykh (approved):

PR #6775, opened 2024-02-02
Partial rework of how lprop scope tree visibility works

msullivan:
The core motivating thing here is that currently basically any use of
linkprops breaks the `[is Foo]` optimization where we only select from
`Foo`.
So if you have something like:
```
select schema::Constraint {
    ancestors[is schema::Constraint]: { @index },
};
```
it will select all `InheritingObject`s.

This is because `@index` is attached to the scopetree in a *fully*
flattened way, as
```
  Constraint.ancestors[is Constraint]@index
  Constraint.ancestors[is Constraint]
  Constraint.ancestors
  Constraint
```
because referring to a link property requires the source *of the link*
to be visible (to prevent the link from deduplicating), but this is
implemented by forcing the every path element to be visible. This means
that in the backend compiler, `Constraint.ancestors` is visible when we
compile the link property, which suppresses the optimization because
it looks like `Constraint.ancestors` may be getting used somewhere
*without* immediately type intersecting it.

Instead we try to do
```
  Constraint.ancestors[is Constraint]@index
    |-Constraint.ancestors[is Constraint]
      |- Constraint.ancestors
  Constraint
```

I think overall this is a code cleanup (_maybe_fixup_lprop can be
deleted), though I suspect there might still be a cleaner way to
structure the attach_path loop.

Some parts of relgen needed some fixing up in order to still be able
to find linkprop sources now that less is flattened.

### DIFF
--- a/edb/ir/scopetree.py
+++ b/edb/ir/scopetree.py
@@ -32,6 +32,7 @@ from edb import errors
 from edb.common import context as pctx
 from edb.common import term
 from . import pathid
+from . import ast as irast
 
 
 class FenceInfo(NamedTuple):
@@ -438,8 +439,10 @@ class ScopeTreeNode:
             #       |-Foo.bar
             #   Foo
             #
-            # I forget why this is necessary for tuples, but it is permissable
-            # because their fields are always singletons.
+            # This is permissable because their fields are always singletons.
+            # FIXME: I think that it should not be *necessary* for tuples,
+            # but test_edgeql_volatility_select_tuples_* fail if it is changed,
+            # I think for incidental reasons.
             #
             # For link properties, this is necessary because referring
             # to a link property at the end of a path suppresses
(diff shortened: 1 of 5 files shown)


edb/ir/scopetree.py
```diff
@@ -439,24 +432,49 @@ def attach_path(
...
+            #   Foo
             #
-            # We could include other cases with invariant cardinality here,
-            # like type intersection, but we want to preserve the prefix
-            # visibility information for the sake of possible optimizations.
-            if (
-                not (is_lprop or prefix.is_linkprop_path())
-                and not prefix.is_tuple_indirection_path()
-            ):
+            # I forget why this is necessary for tuples, but it is permissable
```
elprans, line 441: For tuples because we want 

```edgeql
select (Foo,) filter Foo.name = 'bar';
```

to work as expected (tuples are not fencing)

msullivan: That would work anyway?

msullivan: Or could easily be made to work by adjusting `_node_paths_are_not_links`. (The `name` case works anyway, but cases with links currently would break.)

Once that is fixed, two materialization tests fail without it; I'm confident that it is not necessary in principle, but I'm not sure if it is worth tracking down at the moment

msullivan:
Could I get a review on this?

aljazerzen (approved):

PR #6780, opened 2024-02-05
Support closing all connections to a database on database drop

### LINKED ISSUE #6776: Provide a way to force close all connections to a database
This would be useful for certain anticipated branch-related dev workflows.

It doesn't need to be instant, necessarily. Maybe it would be enough to close all the client connections and then wait for everything to clean itself up?

fantix:
`drop database` will now interrupt and close idle frontend connections before the actual backend database is dropped. Note, idle-in-transaction frontend connections will also be closed.

@msullivan you can configure `close_frontend_conns=True` conditionally in `execute.pyx`.

Fixes #6776 

- [x] Add a test


fantix:
Hmm, while adding a test, I realized this was not enough. `on_before_drop_db()` will abort if any frontend connections are found, regardless of being idle or not. So we can't even reach the backend `drop database`. I'll add an early kick here.

tests/test_database.py
```diff
@@ -238,6 +238,32 @@ async def test_branch_drop_recreate(self):
...
+                'SELECT sys::get_current_database();')
+            self.assertEqual(dbname, ['test_db_disconnect'])
+
+            # Drop branch while the frontend connection is active
+            await self.con.execute(
+                'DROP BRANCH RESET CONNECTIONS test_db_disconnect'
+            )
+
+            # The frontend connection should be closed by the server now
+            self.assertTrue(conn.is_closed())
```
msullivan, line 259: Are you confident that this will be reflected in the client state already?

fantix: Yeah, the test passed at [b851e5c](https://github.com/edgedb/edgedb/actions/runs/7791023176/job/21246262435#step:18:222) - could be flaky but I haven't met

edb/edgeql/parser/grammar/ddl.py
```diff
@@ -851,7 +851,12 @@ def reduce_create_data_branch(self, *kids):
 #
 class DropBranchStmt(Nonterm):
     def reduce_DROP_BRANCH_DatabaseName(self, *kids):
-        self.val = qlast.DropDatabase(name=kids[2].val)
+        _, _, name = kids
+        self.val = qlast.DropDatabase(name=name.val)
+
+    def reduce_DROP_BRANCH_RESET_CONNECTIONS_DatabaseName(self, *kids):
```
elprans, line 857: Not sure I like the formulation. It doesn't read very well and would scale poorly if we add more options. I actually like PostgreSQL's `drop database foo [with option, ...]` somewhat more. I also like `force` as a shorter spelling (a-la Postgres also)

msullivan: Sure

elprans (approved):

PR #6782, opened 2024-02-05
Make DROP DATABASE properly wait for connections to drop

### LINKED ISSUE #4567: DROP DATABASE can fail even after last connection is closed

msullivan:
DROP DATABASE attempts to wait in a retry loop for all incoming
connections to be dropped, but it doesn't quite work, and so we have
hacked around it with another retry loop in the test suite.

Turns out the problem was that we were looking for the *edgedb*
database name in pg_stat_activity, not the *postgres* database name.

Fixes #4567.


elprans (approved):

fantix (approved):

msullivan:
It turns out I was right the first time when I thought that the connection pool was restablishing connections to databases we were trying to disconnect from.

On top of that, though, when trying to tear down `TestServerProtoConcurrentDDL`, sometimes there is still an edgedb connection open. I haven't tracked down how that is possible exactly:
Reading the code, `aclose` sends a termination message and then waits for the server to close the connection, and the server removes the dbview before closing the connection...

msullivan:
Aha! The problem was that some of the client queries were being *cancelled*, which meant they weren't getting a clean shutdown.

Prevent the cancellation by using the `deferred_shield` function that I wrote last week while down an apparently unproductive rabbit hole: https://github.com/edgedb/edgedb/pull/6760#issuecomment-1920352252

edb/server/connpool/pool.py
```diff
@@ -226,11 +228,7 @@ def try_steal(
...
 
-    async def acquire(self) -> C:
-        # There can be a race between a waiter scheduled for to wake up
-        # and a connection being stolen (due to quota being enforced,
-        # for example).  In which case the waiter might get finally
-        # woken up with an empty queue -- hence we use a `while` loop here.
+    async def try_acquire(self) -> typing.Optional[C]:
         self.conn_waiters_num += 1
         try:
             attempts = 0
```
fantix, line 234: `attempts` should go with the while loop into the new outer `acquire()`, and pass in as a parameter.

fantix (approved):

msullivan:
> Prevent the cancellation by using the `deferred_shield` function that I wrote last week while down an apparently unproductive rabbit hole: [#6760 (comment)](https://github.com/edgedb/edgedb/pull/6760#issuecomment-1920352252)

Another option would be to rewrite it to use `gather` instead of TaskGroup, I guess?

msullivan:
I've observed some weird failures locally that I'm trying to track down, which is why I'm not merging yet

msullivan:
Ok, fantix has managed to reproduce the hangs I saw on master (though I haven't yet), so we can go ahead and merge this.
On fantix's (16 core) machine, `edb test -v -j32 -k test_server_adjacent -k test_server_proto` hangs reliably

PR #6783, opened 2024-02-06
Rust axum guide

Dhghomon:


MrFoxPro:
Would like to see some good practices for organaizing structs / code / files when there are a bunch of queries that shares similiar `shapes` and how to organize all structs which describes return types (derived from Queryable).

Dhghomon:
> Would like to see some good practices for organaizing structs / code / files when there are a bunch of queries that shares similiar `shapes` and how to organize all structs which describes return types (derived from Queryable).

Something really comprehensive might turn this from a quick guide into something else but have added two tips off the top of my head that fit with the Open-Meteo output:

https://github.com/edgedb/edgedb/pull/6783#issuecomment-1930044896

MrFoxPro:
Let's say I want to return Json directly without using serde. How can I do it?
How to convert `edgedb_protocol::model::Json` to `axum::Json`?

MrFoxPro:
```rust
impl IntoResponse for edb_model::Json {
   fn into_response(self) -> axum::response::Response {
      (
         [(
            axum::http::header::CONTENT_TYPE,
            "application/json".parse::<axum::http::HeaderValue>().unwrap(),
         )],
         self.to_string(),
      )
   }
}
```

only traits defined in the current crate can be implemented for types defined outside of the crate

PR #6784, opened 2024-02-06
Remove `http_inject_cors` debug flag and replace with `cors_allow_origins` configuration option

### LINKED ISSUE #6655: Allow configuring CORS `Access-Control-Allow-Origin`
Currently if the environment variable `EDGEDB_DEBUG_HTTP_INJECT_CORS` is set to `1`, we will respond with `*`.

We should make this configurable, and ideally (though optionally) have UI for it as well, and then get rid of the environment variable ideally.
Not sure if there are other related fields that should be configurable also.

See https://github.com/edgedb/edgedb/issues/5755, https://github.com/edgedb/edgedb/issues/6569

jaclarke:
Fixes #6655

### DIFF
--- a/edb/server/protocol/protocol.pyx
+++ b/edb/server/protocol/protocol.pyx
@@ -219,6 +219,8 @@ cdef class HttpProtocol:
             self.current_request.content_type = value
         elif name == b'host':
             self.current_request.host = value
+        elif name == b'origin':
+            self.current_request.origin = value
         elif name == b'accept':
             if self.current_request.accept:
                 self.current_request.accept += b',' + value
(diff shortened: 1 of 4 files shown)


scotttrinh (approved):
I spent some time thinking about if there could be a more clear name than `_handle_cors` for this kind of vaguely middleware-like function, but didn't come up with anything better.

Code looks great, and the tests made it easy to follow the change here, thanks for being thorough there!

edb/server/protocol/protocol.pyx
```diff
@@ -724,6 +722,59 @@ cdef class HttpProtocol:
...
+            return False
+
+        origin = request.origin.decode() if request.origin else None
+        origin_allowed = origin is not None and (
+            origin in allowed_origins or '*' in allowed_origins)
+
+        if origin_allowed:
+            response.custom_headers['Access-Control-Allow-Origin'] = origin
+            if expose_headers is not None:
+                response.custom_headers['Access-Control-Expose-Headers'] = \
```
msullivan, line 759: I would suggesting using an open paren to continue the line instead of a backslash

msullivan:
Good to merge?

1st1 (approved):

msullivan:
@raddevon we'll need some documentation support

PR #6787, opened 2024-02-06
Stop writing extension version in the dump.

vpetrovykh:
We currently lack the ability of installing arbitrary extension versions, so when dumping or restoring a database we ignore the version number and only care whether the extension by that name exists or not.

### DIFF
--- a/edb/schema/ddl.py
+++ b/edb/schema/ddl.py
@@ -725,11 +725,13 @@ def ddlast_from_delta(
     sdlmode: bool = False,
     testmode: bool = False,
     descriptive_mode: bool = False,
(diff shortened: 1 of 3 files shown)


msullivan (commented):
I think it might be preferable to handle this in the delta -> AST conversion process. So in `get_ast` for `CreateExtension`

msullivan (approved):

msullivan:
Do you think it is worth updating one of the dump tests to check that it doesn't appear in the dump file? Maybe not.

Did you double check it worked manually?

vpetrovykh:
I checked manually that it works and I don't think it's worth making a separate test for that.

PR #6791, opened 2024-02-07
Add WebAuthn registration flow

scotttrinh:
- Spec: https://w3c.github.io/webauthn
- WebAuthn Primers
  - https://webauthn.guide/#registration
  - https://webauthn.me/introduction

WebAuthn Registration Sequence Diagram

For us, the EdgeDB server is acting as the "Relying Party". The Relying Party is the entity that saves a given Public Key and links it to a local concept of "user" or "identity". The Relying Party has two separate pieces: a stateful server that is responsible for registering and verifying credentials and associating them with their "user"/"identity" concept; and a script that runs in the browser (the "user agent"). We currently facilitate the browser script by running it in our built-in UI, and we add API endpoints to configure the script and verify and save the linked credentials. We will add the equivalent scripting to our `auth-core` (and supporting library code) libraries.

### DIFF
--- a/edb/lib/ext/auth.edgeql
+++ b/edb/lib/ext/auth.edgeql
@@ -55,15 +55,42 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
     };
 
     create type ext::auth::EmailFactor extending ext::auth::Factor {
-        create required property email: str {
-            create delegated constraint exclusive;
-        };
+        create required property email: str;
         create property verified_at: std::datetime;
     };
 
     create type ext::auth::EmailPasswordFactor
         extending ext::auth::EmailFactor {
+        alter property email {
+            create constraint exclusive;
+        };
         create required property password_hash: std::str;
+    };
+
+    create type ext::auth::WebAuthnFactor extending ext::auth::EmailFactor {
+        create required property user_handle: std::bytes {
+            create constraint exclusive;
+        };
+        create required property credential_id: std::bytes {
+            create constraint exclusive;
+        };
+        create required property public_key: std::bytes {
+            create constraint exclusive;
+        };
+
+        create constraint exclusive on ((.email, .credential_id));
+        create constraint exclusive on ((.email, .user_handle));
+    };
+
+    create type ext::auth::WebAuthnRegistrationChallenge
+        extending ext::auth::Auditable {
+        create required property challenge: std::bytes {
+            create constraint exclusive;
+        };
+        create required property email: std::str;
+        create required property user_handle: std::bytes;
+
+        create constraint exclusive on ((.user_handle, .email, .challenge));
     };
 
     create type ext::auth::PKCEChallenge extending ext::auth::Auditable {
@@ -194,6 +221,25 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
         alter property name {
             set default := 'builtin::local_emailpassword';
             set protected := true;
+        };
+
+        create required property require_verification: std::bool {
+            set default := true;
+        };
+    };
+
+    create type ext::auth::WebAuthnProviderConfig
+        extending ext::auth::ProviderConfig {
+        alter property name {
+            set default := 'builtin::local_webauthn';
+            set protected := true;
+        };
+
+        create required property relying_party_origin: std::str {
+            create annotation std::description :=
+                "The full origin of the sign-in page including protocol and \
+                port of the application. If using the built-in UI, this \
+                should be the origin of the EdgeDB server.";
         };
 
         create required property require_verification: std::bool {
--- a/edb/lib/ext/auth.edgeql
+++ b/edb/lib/ext/auth.edgeql
@@ -79,7 +79,6 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
         };
 
         create constraint exclusive on ((.email, .credential_id));
-        create constraint exclusive on ((.email, .user_handle));
     };
 
     create type ext::auth::WebAuthnRegistrationChallenge
--- a/edb/server/protocol/auth_ext/_static/webauthn-register.js
+++ b/edb/server/protocol/auth_ext/_static/webauthn-register.js
@@ -34,7 +34,7 @@ document.addEventListener("DOMContentLoaded", () => {
       });
 
       const redirectUrl = new URL(redirectTo);
-      redirectUrl.searchParams.append("isSignup", "true");
+      redirectUrl.searchParams.append("isSignUp", "true");
       if (maybeCode !== null) {
         redirectUrl.searchParams.append("code", maybeCode);
       }
--- a/edb/server/protocol/auth_ext/_static/webauthn-register.js
+++ b/edb/server/protocol/auth_ext/_static/webauthn-register.js
@@ -1,0 +1,188 @@
+import { decodeBase64Url, encodeBase64Url } from "./utils.js";
+
+document.addEventListener("DOMContentLoaded", () => {
+  const registerForm = document.getElementById("email-factor");
+
+  if (registerForm === null) {
+    return;
+  }
+
+  registerForm.addEventListener("submit", async (event) => {
+    if (event.submitter?.id !== "webauthn-signup") {
+      return;
+    }
+    event.preventDefault();
+
+    const formData = new FormData(/** @type {HTMLFormElement} */ registerForm);
+    const email = formData.get("email");
+    const provider = "builtin::local_webauthn";
+    const challenge = formData.get("challenge");
+    const redirectOnFailure = formData.get("redirect_on_failure");
+    const redirectTo = formData.get("redirect_to");
+    const verifyUrl = formData.get("verify_url");
+
+    if (redirectTo === null) {
+      throw new Error("Missing redirect_to parameter");
+    }
+
+    try {
+      const maybeCode = await register({
+        email,
+        provider,
+        challenge,
+        verifyUrl,
+      });
+
+      const redirectUrl = new URL(redirectTo);
+      redirectUrl.searchParams.append("isSignup", "true");
+      if (maybeCode !== null) {
+        redirectUrl.searchParams.append("code", maybeCode);
+      }
+
+      window.location.href = redirectUrl.href;
+    } catch (error) {
+      console.error("Failed to register WebAuthn credentials:", error);
+      const url = new URL(redirectOnFailure ?? redirectTo);
+      url.searchParams.append("error", error.message);
+      window.location.href = url.href;
+    }
+  });
+});
+
+const WEBAUTHN_OPTIONS_URL = new URL(
+  "../webauthn/register/options",
+  window.location
+);
+const WEBAUTHN_REGISTER_URL = new URL("../webauthn/register", window.location);
+
+/**
+ * Register a new WebAuthn credential for the given email address
+ * @param {Object} props - The properties for registration
+ * @param {string} props.email - Email address to register
+ * @param {string} props.provider - WebAuthn provider
+ * @param {string} props.challenge - PKCE challenge
+ * @param {string} props.verifyUrl - URL to verify email after registration
+ * @returns {Promise<string | null>} - The PKCE code or null if the application
+ *   requires email verification
+ */
+export async function register({ email, provider, challenge, verifyUrl }) {
+  // Check if WebAuthn is supported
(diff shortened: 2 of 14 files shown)


edb/server/protocol/auth_ext/webauthn.py
```diff
@@ -0,0 +1,232 @@
...
+        challenge_dict = result_json[0]
+
+        return WebAuthnRegistrationChallenge(
+            id=challenge_dict["id"],
+            challenge=base64.b64decode(challenge_dict["challenge"]),
+            user_handle=base64.b64decode(challenge_dict["user_handle"]),
+            email=challenge_dict["email"],
+        )
+
+    async def _delete_registration_challenges(
```
jaclarke, line 215: Should we also do cleanup of old challenges, like we do for pkce?

edb/server/protocol/auth_ext/webauthn.py
```diff
@@ -0,0 +1,232 @@
...
+            """
+with
+    challenge := <str>$challenge,
+    user_handle := <str>$user_handle,
+    email := <str>$email,
+insert ext::auth::WebAuthnRegistrationChallenge {
+    challenge := enc::base64_decode(challenge),
+    user_handle := enc::base64_decode(user_handle),
+    email := email,
+}""",
```
jaclarke, line 106: I'm fairly sure that with the `execute_json` api you can just pass base64 encoded strings for `bytes` params, and edgedb handles the decoding without you needing to use `enc::base64_decode`, as `execute_json` is the backend for the edgeql over http endpoint. But maybe your way is more explicit?

```suggestion
with
    challenge := <bytes>$challenge,
    user_handle := <bytes>$user_handle,
    email := <str>$email,
insert ext::auth::WebAuthnRegistrationChallenge {
    challenge := challenge,
    user_handle := user_handle,
    email := email,
}""",
```

edb/server/protocol/auth_ext/webauthn.py
scotttrinh: Since we delete all challenges after a successful registration which means the only "old challenges" we would have are ones where someone began the WebAuthn process and then decided not to use WebAuthn. I guess that's a valid case for cleaning up old challenges but that's far less likely to happen than abandoned/timed-out PKCE flows.

I can look at adding that as a separate PR maybe?

edb/server/protocol/auth_ext/webauthn.py
scotttrinh: Currently it does not:

<details>
<summary>Server gives me this error:</summary>

```
1. TypeError: Object of type bytes is not JSON serializable

------------------------------ Traceback -------------------------------

    /Users/scotttrinh/edgedb/edgedb/edb/server/protocol/auth_ext/http.py, line 113, in handle_request
        > return await self.handle_webauthn_options(*handler_args)
    /Users/scotttrinh/edgedb/edgedb/edb/server/protocol/auth_ext/http.py, line 835, in handle_webauthn_options
        > await webauthn.create_registration_challenge(
    /Users/scotttrinh/edgedb/edgedb/edb/server/protocol/auth_ext/webauthn.py, line 43, in create_registration_challenge
        > await execute.parse_execute_json(
    edb/server/protocol/execute.pyx, line 523, in parse_execute_json
        > return await execute_json(
    edb/server/protocol/execute.pyx, line 561, in execute_json
        > bind_args = _encode_args(args)
    edb/server/protocol/execute.pyx, line 622, in edb.server.protocol.execute._encode_args
        > jval = _encode_json_value(arg)
    edb/server/protocol/execute.pyx, line 606, in edb.server.protocol.execute._encode_json_value
        > jarg = json.dumps(val, cls=DecimalEncoder)
    /nix/store/263fr5l24468xxbxma197gkafc6v0qxz-python3-3.11.4/lib/python3.11/json/__init__.py, line 238, in dumps
        > **kw).encode(obj)
    edb/server/protocol/execute.pyx, line 602, in edb.server.protocol.execute.DecimalEncoder.encode
        > return super().encode(obj)
    /nix/store/263fr5l24468xxbxma197gkafc6v0qxz-python3-3.11.4/lib/python3.11/json/encoder.py, line 200, in encode
        > chunks = self.iterencode(o, _one_shot=True)
    /nix/store/263fr5l24468xxbxma197gkafc6v0qxz-python3-3.11.4/lib/python3.11/json/encoder.py, line 258, in iterencode
        > return _iterencode(o, 0)
    /nix/store/263fr5l24468xxbxma197gkafc6v0qxz-python3-3.11.4/lib/python3.11/json/encoder.py, line 180, in default
        > raise TypeError(f'Object of type {o.__class__.__name__} '
```

</details>

But @1st1 has suggested moving this serialize/deserialize dance into the execute function directly to avoid needing to do it manually. I'm still hoping there is a way we can avoid needing to serialize it at all and use the binary protocol directly, but maybe I'm thinking about this in the wrong way? Do we always have to serialize to JSON when communicating with the server from the server itself?

scotttrinh: Oh, actually I think I misunderstood. You're not saying that I need to drop _encoding_, just that passing b64 strings will automatically be decoded on the server. Lemme try that.

edb/server/protocol/auth_ext/_static/webauthn-register.js
```diff
@@ -0,0 +1,188 @@
...
+ * @param {string} props.provider - WebAuthn provider
+ * @param {string} props.challenge - PKCE challenge
+ * @param {string} props.verifyUrl - URL to verify email after registration
+ * @returns {Promise<string | null>} - The PKCE code or null if the application
+ *   requires email verification
+ */
+export async function register({ email, provider, challenge, verifyUrl }) {
+  // Check if WebAuthn is supported
+  if (!window.PublicKeyCredential) {
+    console.error("WebAuthn is not supported in this browser.");
```
jaclarke, line 71: Maybe we could do this check on page load, before the user tries to register? But we can probably leave that for a later PR when we've decided on the ui design.

scotttrinh: Yeah, I agree that we should hide any WebAuthn affordances if we detect that you cannot use WebAuthn. I suppose in the case where an application has decided to do WebAuthn only, we need to show some kind of helpful empty state as well.

edb/server/protocol/auth_ext/_static/webauthn-register.js
```diff
@@ -0,0 +1,188 @@
...
+    try {
+      const maybeCode = await register({
+        email,
+        provider,
+        challenge,
+        verifyUrl,
+      });
+
+      const redirectUrl = new URL(redirectTo);
+      redirectUrl.searchParams.append("isSignup", "true");
```
jaclarke, line 37:
```suggestion
      redirectUrl.searchParams.append("isSignUp", "true");
```

edb/server/protocol/auth_ext/webauthn.py
scotttrinh: Yep, looks like that works as you've suggested here, thanks! Still would love the avoid the overhead here, but seems strictly better than what I'm doing here!

jaclarke: > But @1st1 has suggested moving this serialize/deserialize dance into the execute function directly to avoid needing to do it manually.

I think that could be done here, in the same way I added support for decimals: https://github.com/edgedb/edgedb/pull/5848/files#diff-32be6e82c8d465d3f9979e75008993c3ad08d677a2456d59c668df53eab631d1

scotttrinh: Maybe we still want to avoid having to explicitly wrap `b64_encode` to the arguments though.

edb/lib/ext/auth.edgeql
```diff
@@ -55,17 +55,44 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
...
     create type ext::auth::EmailPasswordFactor
         extending ext::auth::EmailFactor {
+        alter property email {
+            create constraint exclusive;
+        };
         create required property password_hash: std::str;
     };
 
+    create type ext::auth::WebAuthnFactor extending ext::auth::EmailFactor {
+        create required property user_handle: std::bytes {
```
jaclarke, line 71: I don't think this is an issue, just something that confused me initially: I was wondering why there was an email and a user handle, until I realised webauthn uses user handle to mean user id.

jaclarke: But maybe we should call this `user_id`? To me 'handle' sounds like a user provided name, instead of a generated id that we create for them.

scotttrinh: Yeah, I decided to use the WebAuthn name for this so that it wasn't expected to actually match an "id" in our own system: it's a randomly generated byte sequence. It _should_ be unique for each `LocalIdentity` such that if you use multiple WebAuthn authenticators with the same email (this is allowed and encouraged!) you have the same `user_handle`.

Some additional reading:
  - https://w3c.github.io/webauthn/#sctn-user-handle-privacy
  - https://w3c.github.io/webauthn/#user-handle

I'm open to renaming this, but wanted to at least give some thoughts on why I chose this name.

edb/server/protocol/auth_ext/webauthn.py
jaclarke: That's what I mean. In that custom JSON encoder, we do b64_encode on any bytes values, so when we call `parse_execute_json`, the params in the query are typed with `<bytes>`, and we can pass bytes in the variables.

scotttrinh: @jaclarke ⚡ https://github.com/edgedb/edgedb/pull/6791/files/d30b5e6e4cb4e0e29e47b74cd3331e08236334f3..7ad4733f547d553d6bb57fa8113410b792098209

edb/lib/ext/auth.edgeql
jaclarke: > such that if you use multiple WebAuthn authenticators with the same email (this is allowed and encouraged!) you have the same user_handle

Don't these constraints prevent that though? https://github.com/edgedb/edgedb/pull/6791/files#diff-77c4b6c26c6397edcdda77f3d29f3f96a657838c3fbbba98312e61c790c0affaR81-R82
I assumed that was the reason for having two constraints instead of `create constraint exclusive on ((.email, .credential_id, .user_handle));`

scotttrinh: Ah, you're right: we shouldn't drop that second constraint. We want to avoid having more than one object with the same email and credential ID, but an email can have multiple user handles and multiple credential IDs. Will fix.

edb/server/protocol/execute.pyx
```diff
@@ -611,6 +611,8 @@ class DecimalEncoder(json.JSONEncoder):
                 ) + '}'
         if isinstance(obj, list):
             return '[' + ', '.join(map(self.encode, obj)) + ']'
+        if isinstance(obj, bytes):
+            return base64.b64encode(obj).decode()
```
jaclarke: I think the string needs to be wrapped in `"` so it's valid json?

scotttrinh: Yep, just recursed on `self.encode` again, but maybe it's better to just wrap it in quotes? wdyt?

jaclarke (approved):

edb/lib/ext/auth.edgeql
elprans: Is `WebAuthn` actually an email factor?  Is email always guaranteed to be provided?

scotttrinh: > Is WebAuthn actually an email factor? Is email always guaranteed to be provided?

It requires some kind of identifier, which doesn't strictly have to be an email address, but we kind of married ourselves to email as our "handle" right? It would be weird if webauthn allowed you to use usernames but you couldn't use usernames and passwords.

Or maybe you're asking if the authenticator gives us an email? It does not: the user provides that through the form.

scotttrinh: Strictly speaking you could have no identifier at all, but that's not typical UX.

edb/server/protocol/execute.pyx
jaclarke: I guess wrapping in quotes is slightly more efficient, since we already know it's a string type with no characters that need escaping, but it's probably only a marginal difference.

PR #6792, opened 2024-02-07
Add WebAuthn authentication flow

scotttrinh:
- Spec: https://w3c.github.io/webauthn
- WebAuthn Primers
  - https://webauthn.guide/#authentication
  - https://webauthn.me/introduction

Builds on #6791 

Authentication is almost the same as registration:
1. Fetch the "options" from the server which include a `challenge` and `allowCredentials` that identifies the user based on the provided `email` + `user_handle`
2. The authenticator will sign with the stored private key and generate an `assertion` which includes the signed challenge, and other information about the authenticator.
3. The server verifies the signature and some other metadata, and if it passes considers the session authenticated.

Like the registration flow, it uses PKCE to do the final exchange of the actual token.

### DIFF
--- a/edb/lib/ext/auth.edgeql
+++ b/edb/lib/ext/auth.edgeql
@@ -92,6 +92,16 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
         create constraint exclusive on ((.user_handle, .email, .challenge));
     };
 
+    create type ext::auth::WebAuthnAuthenticationChallenge
+        extending ext::auth::Auditable {
+        create required property challenge: std::bytes {
+            create constraint exclusive;
+        };
+        create required link factor: ext::auth::WebAuthnFactor {
+            create constraint exclusive;
+        };
+    };
+
     create type ext::auth::PKCEChallenge extending ext::auth::Auditable {
         create required property challenge: std::str {
             create constraint exclusive;
--- a/edb/lib/ext/auth.edgeql
+++ b/edb/lib/ext/auth.edgeql
@@ -76,6 +76,16 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
             create constraint exclusive;
         };
 
+        create trigger email_shares_user_handle after insert for each do (
+            std::assert(
+                (__new__.user_handle = (
+                    select detached ext::auth::WebAuthnFactor
+                    filter .email = __new__.email
+                    and not .id = __new__.id
+                ).user_handle) ?? true,
+                message := "user_handle must be the same for a given email"
+            )
+        );
         create constraint exclusive on ((.email, .credential_id));
     };
 
--- a/edb/lib/ext/auth.edgeql
+++ b/edb/lib/ext/auth.edgeql
@@ -68,9 +68,7 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
     };
 
     create type ext::auth::WebAuthnFactor extending ext::auth::EmailFactor {
-        create required property user_handle: std::bytes {
-            create constraint exclusive;
-        };
+        create required property user_handle: std::bytes;
         create required property credential_id: std::bytes {
             create constraint exclusive;
         };
--- a/edb/lib/ext/auth.edgeql
+++ b/edb/lib/ext/auth.edgeql
@@ -78,11 +78,11 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
 
         create trigger email_shares_user_handle after insert for each do (
             std::assert(
-                (__new__.user_handle = (
+                __new__.user_handle = (
                     select detached ext::auth::WebAuthnFactor
                     filter .email = __new__.email
                     and not .id = __new__.id
-                ).user_handle) ?? true,
+                ).user_handle,
                 message := "user_handle must be the same for a given email"
             )
         );
--- a/edb/server/protocol/auth_ext/_static/webauthn-authenticate.js
+++ b/edb/server/protocol/auth_ext/_static/webauthn-authenticate.js
@@ -1,19 +1,26 @@
-import { decodeBase64Url, encodeBase64Url } from "./utils.js";
+import {
+  decodeBase64Url,
+  encodeBase64Url,
+  parseResponseAsJSON,
+} from "./utils.js";
 
 document.addEventListener("DOMContentLoaded", () => {
-  const registerForm = document.getElementById("email-factor");
+  /** @type {HTMLFormElement | null} */
+  const authenticateForm = document.getElementById("email-factor");
 
-  if (registerForm === null) {
+  if (authenticateForm === null) {
     return;
   }
 
-  registerForm.addEventListener("submit", async (event) => {
+  authenticateForm.addEventListener("submit", async (event) => {
     if (event.submitter?.id !== "webauthn-signin") {
       return;
     }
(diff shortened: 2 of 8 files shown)


edb/server/protocol/auth_ext/webauthn.py
```diff
@@ -230,3 +234,207 @@ async def _delete_registration_challenges(
...
+}
+filter .email = <str>$email;""",
+            variables={
+                "email": email,
+            },
+            cached_globally=True,
+        )
+        result_json = json.loads(result.decode())
+        user_handles: set[str] = {x["user_handle"] for x in result_json}
+        assert len(user_handles) == 1
```
jaclarke, line 260: Why do we assert that there's only a single user_handle associated with a given email? I thought we wanted to allow you to use the same email with multiple authenticators: https://github.com/edgedb/edgedb/pull/6791#discussion_r1483149382
Unless they should all have the same user_handle (which I guess was the reason for dropping the exclusive constraint on `(.email, .user_handle)`?), in which case in `create_registration_options_for_email` we should check if there's already a `WebAuthnFactor` for that email and re-use the `user_handle`. I don't know if there's some security issue with being able to retrieve an email's user_handle like that though?

edb/server/protocol/auth_ext/_static/webauthn-authenticate.js
```diff
@@ -0,0 +1,188 @@
...
+    body: JSON.stringify({
+      email: props.email,
+      assertion: encodedAssertion,
+      provider: props.provider,
+      challenge: props.challenge,
+    }),
+  });
+
+  if (!authenticateResponse.ok) {
+    console.error(
```
jaclarke, line 174: I don't think this handles the case where email verification is required, so we'll end up returning a misleading error that webauthn auth failed.

edb/server/protocol/auth_ext/webauthn.py
scotttrinh: Yeah, we're _supposed_ to use the same user_handle if the user is "the same user" to allow multiple WebAuthn factors for a single user. You're right that I'm missing that check from the `create_registration_options_for_email` but I need to think about the UX here a bit more.

The ostensible reason for allowing multiple factors is to allow what are basically back-up factors (or multi-factors) so you can seamlessly move between passkeys, hardware authenticators, etc. but in our "registration" flow it's not super obvious that you might have multiple identities with the same email which point to a single `User` in the downstream application because we kind of leave that up to the application developer to do.

But, yeah I think at the very least we should reuse the same user handle if you provide an existing email, will fix.

edb/server/protocol/auth_ext/_static/webauthn-authenticate.js
scotttrinh: Good point. Now that we have stable type names in our error json, I can detect this case and at least throw a more helpful error. We still navigate you to the "error" state of the sign in page in the email+password case, so I think that throwing the correct error should do the right thing.

edb/server/protocol/auth_ext/webauthn.py
scotttrinh: @jaclarke ⚡ [`ebda3e8` (#6792)](https://github.com/edgedb/edgedb/pull/6792/commits/ebda3e809602670e77b623b401cd6951aff914bf)

edb/server/protocol/auth_ext/_static/webauthn-authenticate.js
scotttrinh: @jaclarke ⚡ [`64eaa1e` (#6792)](https://github.com/edgedb/edgedb/pull/6792/commits/64eaa1ec349562a66d7e38a65f988bbc3ffd3075)

edb/server/protocol/auth_ext/webauthn.py
jaclarke: From testing it seems there's another issue with this assert. If the email isn't registered, then there's no user-handles,the assert fails and we return an ise. I'm not sure what the correct error should be here? I know we're trying not to leak the existence of emails, but I don't think we can get around that as we can only return a challenge if we have the stored credentials for an email.

scotttrinh: Yeah, good catch: assert is the wrong tool here, we need to raise a specific exception for proper error handling.

As far as data leakage is concerned, I think we talked about `register` being fine to leak whether an email is signed up or not, but `authenticate` would always pretend it succeeds. From that perspective, I think we can return a sensible error here.

elprans (approved):

edb/server/protocol/auth_ext/webauthn.py
scotttrinh: @jaclarke ⚡ https://github.com/edgedb/edgedb/pull/6792/files/40d636291fca7bb5db36ebd142e7e58fa2214bdd..1431063a019694d9028966b1ae1cbce2d8c6e93b

Turns out I needed a bit more work to ensure the invariant that multiple factors with the same `email` have the same `user_handle`. Added a test for the new trigger.

edb/lib/ext/auth.edgeql
```diff
@@ -76,6 +76,16 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
             create constraint exclusive;
         };
 
+        create trigger email_shares_user_handle after insert for each do (
+            std::assert(
+                (__new__.user_handle = (
+                    select detached ext::auth::WebAuthnFactor
+                    filter .email = __new__.email
+                    and not .id = __new__.id
+                ).user_handle) ?? true,
```
jaclarke, line 85: Minor thing, but I think `?? true` probably isn't needed. If the result of the select is an empty set, the assertion just won't run anyway.

edb/server/protocol/auth_ext/webauthn.py
```diff
@@ -284,8 +284,17 @@ async def create_authentication_options_for_email(
...
+            raise errors.WebAuthnAuthenticationFailed(
+                "No WebAuthn credentials found for the email."
+            )
+
         user_handles: set[str] = {x["user_handle"] for x in result_json}
-        assert len(user_handles) == 1
+        if len(user_handles) > 1:
+            raise errors.WebAuthnAuthenticationFailed(
+                "Multiple user handles found for the email."
+            )
```
jaclarke, line 296: Could still be an assert. It's really more of an ISE if there's multiple user handles, since the db should enforce that.

jaclarke (approved):

edb/lib/ext/auth.edgeql
scotttrinh: TIL about the empty set behavior of `assert`. I'll add an example to the documentation for `assert`.

jaclarke: I think it just follows the normal set behaviour, where if the function arg isn't `set of` (like in `count`, for example) then it's expanded out. So `assert({true, false})` is executed as `{assert(true), assert(false)}`, and similarly `assert({})` becomes `{}`.

scotttrinh: That makes sense. The documentation hints at this to since it says:

> If the input bool is false, assert raises a QueryAssertionError. _Otherwise, this function returns true._ (emphasis mine)

I think it'd just be helpful to show some examples of calling this function with a sets to show that. I'll follow up.

PR #6796, opened 2024-02-07
Make constraint error details contain useful information for developers

### LINKED ISSUE #6768: `message` and `details` fields are the same when you specify an `errmessage` for a constraint.
- EdgeDB Version: 4.4+aa576e1
- EdgeDB CLI Version:  4.0.2+500be79
- OS Version: macOS ventura

Steps to Reproduce:

1. For one of the properties in your schema add a `errmessage`, to customize the default errors emiitted for constraint validation.
2. Try to insert a record using the edgedb repl that fails this constraint
3. It returns the same error for `message` as well as `detail`

Screenshot 2024-02-01 at 17 34 03

Schema:
```
  type User extending Model {
    first_name: str;
    last_name: str;
    required email: str {
      constraint exclusive {
        errmessage := 'non-unique email'
      };
      constraint regexp(r'^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$') {
        errmessage := 'invalid email, try again later'
      };
    };
    is_email_verified: bool {
      default := false;
    };
  }
```

Originally asked here:
https://discord.com/channels/841451783728529451/849374705210490900/1202341124673634304

aljazerzen:
Closes #6768

Changes error details of constraint violations to:

```
f'violated {constraint_description} on {subject_description}'
```
... where `x_description` is the `x.get_verbosename()`

### DIFF
--- a/edb/schema/constraints.py
+++ b/edb/schema/constraints.py
@@ -198,19 +198,6 @@ class Constraint(
         # if there no origin, return self
         return [self] if not origins else list(origins)
 
-    def get_ancestor_from_std(
-        self, schema: s_schema.Schema
-    ) -> Optional[Constraint]:
-        name: sn.QualName = self.get_name(schema)
-        if name.module == 'std':
-            return self
-
-        for base in self.get_bases(schema).objects(schema):
-            std_ancestor = base.get_ancestor_from_std(schema)
-            if std_ancestor:
-                return std_ancestor
-        return None
-
     def is_independent(self, schema: s_schema.Schema) -> bool:
         return (
             not self.descendants(schema)
--- a/edb/schema/constraints.py
+++ b/edb/schema/constraints.py
@@ -218,21 +218,6 @@ class Constraint(
             self.get_field_value(schema, 'subject'),
         )
 
-    def format_error(
-        self,
-        schema: s_schema.Schema,
-    ) -> str:
-        subject = self.get_subject(schema)
-        titleattr = subject.get_annotation(schema, sn.QualName('std', 'title'))
-
-        if not titleattr:
-            subjname = subject.get_shortname(schema)
-            subjtitle = subjname.name
-        else:
-            subjtitle = titleattr
-
-        return self.format_error_message(schema, subjtitle)
-
     def format_error_message(
         self,
         schema: s_schema.Schema,
--- a/edb/schema/constraints.py
+++ b/edb/schema/constraints.py
@@ -261,6 +261,7 @@ class Constraint(
         subject_name: str,
     ) -> str:
         text = self.get_errmessage(schema)
+        assert text
         args = self.get_args(schema)
         if args:
             args_ql: List[qlast.Base] = [
--- a/edb/schema/constraints.py
+++ b/edb/schema/constraints.py
@@ -218,6 +218,21 @@ class Constraint(
             self.get_field_value(schema, 'subject'),
         )
 
+    def format_error(
+        self,
+        schema: s_schema.Schema,
+    ) -> str:
+        subject = self.get_subject(schema)
+        titleattr = subject.get_annotation(schema, sn.QualName('std', 'title'))
+
+        if not titleattr:
+            subjname = subject.get_shortname(schema)
+            subjtitle = subjname.name
+        else:
+            subjtitle = titleattr
+
+        return self.format_error_message(schema, subjtitle)
+
     def format_error_message(
         self,
         schema: s_schema.Schema,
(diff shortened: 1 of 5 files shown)


edb/schema/constraints.py
```diff
@@ -176,18 +176,27 @@ def get_name_impacting_ancestors(
...
+            # if this base applies
+            is_concrete = not base.is_non_concrete(schema)
+            if is_concrete and not base.get_delegated(schema):
+
+                # add all of its origins
+                for x in base.get_constraint_origins(schema):
+                    origins.add(x)
+
+        # if there no origin, return self
+        return [self] if not origins else list(origins)
```
aljazerzen, line 199: This is a refactor, me trying to understand if I can use this code for my case

edb/server/compiler/errormech.py
```diff
@@ -620,10 +622,17 @@ def _interpret_constraint_errors(
...
             constraint = obj_ptr.get_exclusive_constraints(schema)[0]
 
-        msg = constraint.format_error(schema)
+        # msg is for the "end user" that should not mention pointers and object
+        # type it is also affected by setting `errmessage` in user schema.
+        msg = constraint.format_error_message(schema)
+
+        # details is for the "developer" that must explain what's going on
+        # under the hood. It should be picked up from the errmessage on the
+        # first ancestor constraint that's in std module.
```
aljazerzen, line 631: This was my initial approach for the details

tests/test_constraints.py
```diff
@@ -1148,14 +1147,26 @@ async def test_constraints_ddl_04(self):
...
+                await self.con.execute(
+                    """
+                    INSERT ConstraintOnTest3 {
+                        foo := 'Test'
+                    };
+                    """
+                )
+            except edgedb.ConstraintViolationError as e:
+                # edgedb-python does not expose a nicer access to details field
+                details = e._attrs[2].decode("utf-8")
```
aljazerzen, line 1162: Am I doing something wrong here? Or does edgedb-python really don't expose error details field?

aljazerzen: Turns out that edgedb-python really does not expose details.

edb/server/compiler/errormech.py
```diff
@@ -599,6 +599,8 @@ def _interpret_constraint_errors(
     # no need for else clause since it would have been handled by
     # the static version
 
+    print(error_type)
```
elprans, line 602:
```suggestion
```

edb/schema/constraints.py
msullivan: Probably not, right?

edb/schema/constraints.py
```diff
@@ -176,18 +176,27 @@ def get_name_impacting_ancestors(
...
     def get_constraint_origins(
-            self, schema: s_schema.Schema) -> List[Constraint]:
-        origins: List[Constraint] = []
+        self, schema: s_schema.Schema
+    ) -> List[Constraint]:
+        """
+        Find the transitive closure of all constraints that apply.
+        Return parents of these constraints.
+        """
+        # I'm not sure that this explanation is clear (or even entirely correct)
```
msullivan, line 185: I don't think this explanation is quite right. I think it would be worth having clear documentation here, because there is a subtle issue at play.

The purpose of this function is to find which types an exclusive constraint originate from. If we have `Baz <: Bar <: Foo`, and Foo declares some exclusive constraint, then when we do an insert into `Baz` we need to check all `Foo`s to make sure there is no conflict. (Which we do using triggers, and `get_constraint_origin` helps drive the generation of those.)

The subtleties here are:
 1. Because of multiple inheritance, we might have an exclusive constraint that originates from multiple unrelated ancestors
 2. Delegated exclusive constraints are declared on a parent, but their "origins" are their children. (So that `abstract type Named` or whatever can declare an exclusive name, but the names will be exclusive *within* each of the types that subtypes it.)

msullivan (approved):

edb/schema/constraints.py
aljazerzen: No, it does something else (but that was not obvious to me before the refactor)

tests/test_constraints.py
aljazerzen: I'll wait for a release of edgedb-python and clean this up before merging.

edb/schema/constraints.py
aljazerzen: I'll add these docs in a separate PR

msullivan:
Let's merge it without waiting for edgedb-python, so we can backport it without needing to bump edgedb-python on those branches

PR #6804, opened 2024-02-09
Update `pgvector` extension.

vpetrovykh:
Add getter and setter for `ef_search` used by HNSW.


msullivan:
Ugh, we can merge this if we need to but I should try to make session config of extension configs actually work

msullivan:
Session and backend extension configs work now, so let's do it that way instead

PR #6806, opened 2024-02-09
Fix empty array literals in SQL adapter

msullivan:
Apparently the `elements` field gets omitted from pg_query's output
when there aren't elements.

### DIFF
--- a/edb/pgsql/parser/ast_builder.py
+++ b/edb/pgsql/parser/ast_builder.py
@@ -77,7 +77,7 @@ def _list(
     builder: Builder,
     mapper: Callable[[T], U] = _ident,
 ) -> List[U]:
-    return [mapper(builder(n, ctx)) for n in node[name]]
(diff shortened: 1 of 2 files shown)


aljazerzen (approved):
I hate non-consistent shit like this.

An uncompromising system would have zero-cost empty array, so there would be no benefit of omitting it. But we use JSON, so there is a cost and libpg_query deems it appropriate to omit the key, so we need a special case. A small initial compromise wastes hours down the line.

msullivan:
Having the zero-cost empty array requires a more heavily schema-full system, though, which has pluses and minuses. Though I think that libpg_query *does* support protobufs...

I suspect the core of it is that when the array is empty, postgres represents it as NULL instead of allocating a zero length array...

aljazerzen:
We should have used the protobuf interface instead of JSON. But regarding NULL for empty array: [this is the solution](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.new)

PR #6807, opened 2024-02-09
IDEA: Introduce a WITH based syntax for FOR-style iteration

msullivan:
I dislike how `for` looks and feels so much "heavier" than `with`, due
to the `union` keyword and the need to parenthesize the body if it is
a query, and especially if you need to have *nested* `for`s.

So my idea is to introduce `<-` binding, which behaves like a `for`
binding.

So then you can do something like
```
with name <- {'Sully', 'Zhibo', 'Aljaž'}
insert Person { name := name }
```
I modified some tests you can look at to compare.

Is this a bad idea, and I'm just sleep-deprived?
Does anybody but me like this?
(Does anybody have any other ideas about lighter-weight `for`?)

### DIFF
--- a/edb/edgeql/ast.py
+++ b/edb/edgeql/ast.py
@@ -155,6 +155,7 @@ class SortExpr(Base):
 class AliasedExpr(Base):
     alias: str
     expr: Expr
+    monadic: bool = False
 
 
 class ModuleAliasDecl(Base):
--- a/edb/edgeql/codegen.py
+++ b/edb/edgeql/codegen.py
@@ -195,7 +195,10 @@ class EdgeQLSourceGenerator(codegen.SourceGenerator):
 
     def visit_AliasedExpr(self, node: qlast.AliasedExpr) -> None:
         self.write(ident_to_str(node.alias))
-        self.write(' := ')
+        if node.monadic:
+            self.write(' <- ')
+        else:
+            self.write(' := ')
         self._block_ws(1)
 
         self.visit(node.expr)
--- a/edb/edgeql/compiler/stmt.py
+++ b/edb/edgeql/compiler/stmt.py
@@ -24,6 +24,7 @@ from __future__ import annotations
 from typing import *
 
 from collections import defaultdict
+import functools
 import textwrap
 
 from edb import errors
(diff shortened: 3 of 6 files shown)


msullivan:
How about something like
```
for name <- {'Sully', 'Zhibo', 'Aljaž'}
insert Person { name := name }
```

Do people like that at all? Still light weight, no grammar conflicts, no overloading of `with`

Dhghomon:
> I dislike how `for` looks and feels so much "heavier" than `with`, due to the `union` keyword and the need to parenthesize the body if it is a query, and especially if you need to have _nested_ `for`s.
> 
> So my idea is to introduce `<-` binding, which behaves like a `for` binding.
> 
> So then you can do something like
> 
> ```
> with name <- {'Sully', 'Zhibo', 'Aljaž'}
> insert Person { name := name }
> ```
> 
> I modified some tests you can look at to compare.
> 
> Is this a bad idea, and I'm just sleep-deprived? Does anybody but me like this? (Does anybody have any other ideas about lighter-weight `for`?)

On the documentation/promoting EdgeDB side too explaining why `union` is needed is always a bit unfortunate. Takes you from the usual "Check out this new feature that just works" to "Okay, sit down for a sec and hear some justification for why this syntax is the way it is..." Finding some way to have `for` feel like `with` would definitely be a plus.

corbinday:
How would this work with existing syntax for scenarios where `with` is used with `for`? I'm thinking along the lines of where there are variables that use both `<-` and `:=` in the with block.

I like the idea that `<-` binding would imply the `for` behavior when doing an operation. 

I'm imagining something like this:
```edgeql
with
  races <- array_unpack((
    with module schema
    select ScalarType.enum_values 
    filter ScalarType.name = 'default::Race'
  )),
  classes <- array_unpack((
    with module schema
    select ScalarType.enum_values 
    filter ScalarType.name = 'default::Class'
  )),
  created := datetime_of_statement() 
insert NPC {
  race := races,
  class := classes
  created := created
}
```
where the `<-` binding would imply that the insert will happen for every combination in the Cartesian product of `race` and `class`. However, `created` would not factor into the Cartesian product. It would be equivalent to the existing syntax:
```edgeql
with
  races := array_unpack((
    with module schema
    select ScalarType.enum_values 
    filter ScalarType.name = 'default::Race'
  )),
  classes := array_unpack((
    with module schema
    select ScalarType.enum_values
    filter ScalarType.name = 'default::Class'
  )),
  date_created := datetime_of_statement()
for r in races
union (
  for c in classes
  union (
    insert NPC {
      race := races,
      class := classes
      created := created
    }
  )
)
```

Perhaps alternatively, the `<-` could be used inside the query statement rather than the `with` block because really it's the stuff after the `with` block that is behaving differently. It's already common to have sets prepared in the `with` block using `:=`. 
```edgeql
with
  races := array_unpack((
    with module schema
    select ScalarType.enum_values 
    filter ScalarType.name = 'default::Race'
  )),
  classes := array_unpack((
    with module schema
    select ScalarType.enum_values 
    filter ScalarType.name = 'default::Class'
  )),
  created := datetime_of_statement() 
insert NPC {
  race <- races,
  class <- classes
  created := created
}
```

Also, maybe something like `:*` would make more sense  semantically than `<-` since `*` is commonly associated with taking the product. 
```edgeql
with
  races := array_unpack((
    with module schema
    select ScalarType.enum_values 
    filter ScalarType.name = 'default::Race'
  )),
  classes := array_unpack((
    with module schema
    select ScalarType.enum_values 
    filter ScalarType.name = 'default::Class'
  )),
  created := datetime_of_statement() 
insert NPC {
  race :* races,
  class :* classes
  created := created
}
```

elprans:
#6810 has a better solution to the `UNION(..)` verbosity problem in cases of when the `for` body is a statement.

PR #6810, opened 2024-02-10
Allow omitting `UNION` in `FOR` if the body is a statement

elprans:
Lots (probably most) uses of `FOR` in EdgeQL involve a statement in the
loop body, so the current formulation requiring a `UNION` keyword AND a
set of parentheses around the query are seen as more unwieldy than
necessary.  Fortunately, since statements always start with a reserved
keyword, the grammar allows us to simplify the syntactic form to:

    FOR var IN <atomic-expr>
    <stmt-core>

For example:

    FOR name IN {"Alice", "Bob"}
    INSERT User { name := name }

### DIFF
--- a/edb/edgeql/parser/grammar/expressions.py
+++ b/edb/edgeql/parser/grammar/expressions.py
@@ -170,20 +170,23 @@ class OptionalOptional(Nonterm):
 class SimpleFor(Nonterm):
     def reduce_ForIn(self, *kids):
         r"%reduce FOR OptionalOptional Identifier IN AtomicExpr UNION Expr"
+        _, optional, iterator_alias, _, iterator, _, body = kids
         self.val = qlast.ForQuery(
-            optional=kids[1].val,
-            iterator_alias=kids[2].val,
-            iterator=kids[4].val,
-            result=kids[6].val,
+            optional=optional.val,
+            iterator_alias=iterator_alias.val,
+            iterator=iterator.val,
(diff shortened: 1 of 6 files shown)


edb/edgeql/parser/grammar/expressions.py
```diff
@@ -168,16 +168,24 @@ def reduce_empty(self, *kids):
...
+        r"%reduce FOR OptionalOptional Identifier IN AtomicExpr UNION Expr"
         self.val = qlast.ForQuery(
             optional=kids[1].val,
             iterator_alias=kids[2].val,
             iterator=kids[4].val,
             result=kids[6].val,
         )
 
+    def reduce_ForInStmt(self, *kids):
+        r"%reduce FOR OptionalOptional Identifier IN AtomicExpr ExprStmtCore"
```
aljazerzen, line 181:
```suggestion
        r"%reduce FOR OptionalOptional Identifier IN AtomicExpr ExprStmt"
```

aljazerzen (approved):

msullivan (approved):

vpetrovykh (approved):
Ideally our codegen would not inject `UNION` and be able to use the shorter syntax if the result is an `ast.Query`. We can fix it later, but this would affect our normalized expressions in the schema so probably better to fix it now.

vpetrovykh (approved):
LGTM

PR #6816, opened 2024-02-12
Migrate rust-python crates from cpython to pyo3

aljazerzen:
- refactor: migrate graphql-rewrite crate from cpython to pyo3
- refactor: migrate edgeql-parser crate from cpython-rust to pyo3

I've disabled the macros that generate AST because that is not used currently and has to be changed up a lot before it becomes useful.


elprans (approved):
Tested prod builds (including on 3.12), works great!

PR #6819, opened 2024-02-12
Fix query cache dbver issue with concurrent DDL

fantix:
During a query compilation, if the schema is changed, the server was caching the compiled query under a wrong dbver, causing future queries to fail. See the test for issue reproduction.

@msullivan, I think you mentioned adding a similar compiler sleep hook for testing, but I can't find it. Please let me know if there's a better way than the sleep sentinel I added here in the query string.

### DIFF
--- a/edb/server/compiler/compiler.py
+++ b/edb/server/compiler/compiler.py
@@ -2337,6 +2337,9 @@ def _try_compile(
     source: edgeql.Source,
 ) -> dbstate.QueryUnitGroup:
     if _get_config_val(ctx, '__internal_testmode'):
+        # This is a bad but simple way to emulate a slow compilation for tests.
+        # Ideally, we should have a testmode function that is hooked to sleep
+        # as `simple_special_case`, or wait for a notification from the test.
         sentinel = "# EDGEDB_TEST_COMPILER_SLEEP = "
         text = source.text()
         if text.startswith(sentinel):
--- a/edb/server/compiler/compiler.py
+++ b/edb/server/compiler/compiler.py
@@ -26,6 +26,7 @@ import json
 import hashlib
 import pickle
 import textwrap
+import time
 import uuid
 
 import immutables
@@ -2335,6 +2336,11 @@ def _try_compile(
     ctx: CompileContext,
     source: edgeql.Source,
 ) -> dbstate.QueryUnitGroup:
+    if _get_config_val(ctx, '__internal_testmode'):
+        sentinel = "# EDGEDB_TEST_COMPILER_SLEEP = "
+        text = source.text()
+        if text.startswith(sentinel):
+            time.sleep(float(text[len(sentinel):text.index("\n")]))
 
     default_cardinality = enums.Cardinality.NO_RESULT
     statements = edgeql.parse_block(source)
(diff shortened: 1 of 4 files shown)


msullivan:
> @msullivan, I think you mentioned adding a similar compiler sleep hook for testing, but I can't find it. Please let me know if there's a better way than the sleep sentinel I added here in the query string.

I don't know if there is really any *good* way to do such things. (I did a different bad way, and then didn't merge it.)

msullivan:
Maybe a better (I guess?) way would be to add some testmode function, and then have a compiler function hook using `simple_special_case` that sleeps. Downside there though is that it is tough to pass in the sleep amount, since constants will be extracted. (You could have it wait for the the *length* of an array, or something silly like that)

edb/server/compiler/compiler.py
```diff
@@ -2335,6 +2336,11 @@ def _try_compile(
     ctx: CompileContext,
     source: edgeql.Source,
 ) -> dbstate.QueryUnitGroup:
+    if _get_config_val(ctx, '__internal_testmode'):
```
msullivan, line 2339: Add a comment I suppose

edb/server/dbview/dbview.pyx
```diff
@@ -230,15 +230,17 @@ cdef class Database:
...
-    cdef _cache_compiled_query(self, key, compiled: dbstate.QueryUnitGroup):
+    cdef _cache_compiled_query(
+        self, key, compiled: dbstate.QueryUnitGroup, int dbver
+    ):
         assert compiled.cacheable
 
-        existing, dbver = self._eql_to_compiled.get(key, DICTDEFAULT)
-        if existing is not None and dbver == self.dbver:
+        existing, existing_dbver = self._eql_to_compiled.get(key, DICTDEFAULT)
+        if existing is not None and existing_dbver == self.dbver:
```
msullivan, line 239: is it right to be checking against `self.dbver` but populating it with `dbver`?

msullivan: (yes, I think, right?)

tests/test_server_proto.py
```diff
@@ -2827,6 +2828,63 @@ async def test_server_proto_query_cache_invalidate_09(self):
...
+                }};
+            ''')
+            if time.monotonic() - start > sleep:
+                self.skipTest("The host is too slow for this test.")
+                # If this happens too much, consider increasing the sleep time
+
+            with self.assertRaisesRegex(
+                edgedb.errors.InternalServerError,
+                "column .* does not exist",
+            ):
```
msullivan, line 2877: This seems wrong as intended behavior. At least in this case, where the schema is changed while we were compiling, it should be easy to notice that and retry.

It does seem like it might be hard to prevent in the general case, though, without locking things in ways we don't want to? Probably worth thinking about.

msullivan (approved):
This makes sense to me.

It's probably worth thinking more about the race conditions against DDL, though, that can cause transient ISEs like this. That's longstanding, though, I think, so probably doesn't need to be fixed right now

edb/server/dbview/dbview.pyx
fantix: Right, `dbver` is the only guaranteed version that the current query unit was compiled upon. I'll add a comment.

tests/test_server_proto.py
fantix: Created an issue for a future fix: https://github.com/edgedb/edgedb/issues/6820

PR #6821, opened 2024-02-12
Use `asyncio.TaskGroup` instead of the original in-tree implementation

elprans:
The `edb.common.TaskGroup` stuff has been upstreamed to Python since
3.11, and since that's the minimum required version, switch over to
using the implementation from the standard library.

Also, drop `MultiError` as we now have the `ExceptionGroup`


1st1 (approved):
Obviously this is a good idea. Thanks for doing the tedious work of cleaning this up.

fantix (approved):

PR #6823, opened 2024-02-12
Run CI tests with Python 3.12

elprans:


fantix (approved):

tests/common/test_signalctl.py
```diff
@@ -313,7 +313,7 @@ async def _task():
                 except asyncio.CancelledError:
                     # In case the task cancellation is hanging, ..
                     fut.set_result(None)
-                    await self.wait_for_parent(2)
+                    await self.loop.create_future()
```
msullivan, line 316: Could we put a comment about why we are intentionally waiting forever?

elprans: @fantix I pulled out your fix to #6835

PR #6826, opened 2024-02-13
Add Magic Links passwordless provider

scotttrinh:
Adds another passwordless authentication option that sends an expiring token to an email address. This is similar in spirit to a reset password flow. This is commonly called "Magic Link", so that's what I've adopted here at least as an internal feature name. For users, we plan on calling it "Email a sign in link".

### DIFF
--- a/edb/server/protocol/auth_ext/http.py
+++ b/edb/server/protocol/auth_ext/http.py
@@ -51,6 +51,7 @@ from . import (
     config,
     email as auth_emails,
     webauthn,
+    magic_link,
 )
 
 
@@ -106,6 +107,14 @@ class Router:
                     return await self.handle_send_reset_email(*handler_args)
                 case ('reset-password',):
                     return await self.handle_reset_password(*handler_args)
+                case ('magic-link', 'register'):
+                    return await self.handle_magic_link_register(*handler_args)
+                case ('magic-link', 'email'):
+                    return await self.handle_magic_link_email(*handler_args)
+                case ('magic-link', 'authenticate'):
+                    return await self.handle_magic_link_authenticate(
+                        *handler_args
+                    )
 
                 # WebAuthn routes
                 case ('webauthn', 'register'):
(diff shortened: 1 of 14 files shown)


TreeOfLearning:
We use passwordless authentication in our product and will likely switch to this once it is integrated as edgedb drives our db system. Very exciting. 

The term "magic link" is pretty ubiquitous and well understood. Any reason you're adopting different user-facing phrasing? I think it actually makes it less clear - a "sign in" link sounds like a link where you _can_ sign in, not a link that _does the signing in_ for you. In my opinion, at least :)

scotttrinh:
> The term "magic link" is pretty ubiquitous and well understood. Any reason you're adopting different user-facing phrasing? I think it actually makes it less clear - a "sign in" link sounds like a link where you _can_ sign in, not a link that _does the signing in_ for you. In my opinion, at least :) 

Thanks so much for the feedback here, that's actually really helpful.

For our built-in UI, we have to treat it like a product we have a strong opinion on. We had some internal discussion about this and landed on a more general-public phrasing here. Since this is something you can make your own UI for, we expect that people will adjust the wording to fit their audience.

Nothing here is set in stone and we are even open to making tweaks like this in point releases if we find that most developers simply prefer another wording, so duly noted!

edb/server/protocol/auth_ext/http.py
```diff
@@ -818,12 +829,178 @@ async def handle_reset_password(self, request: Any, response: Any):
...
+            test_mode=self.test_mode,
+            issuer=self.base_path,
+        )
+        query = urllib.parse.parse_qs(
+            request.url.query.decode("ascii") if request.url.query else ""
+        )
+        token = _get_search_param(query, "token")
+
+        (identity_id, challenge, callback_url) = (
+            self._get_data_from_magic_link_token(token)
```
jaclarke, line 989: We should probably catch all the errors in this endpoint (like the expired token error), and always redirect, since this is a link that the user follows from an email and not an api call, so we don't want to return json.

scotttrinh: oh, that's tricky since we don't know where to redirect errors to. I suppose we'll need to pass that through as part of the URL search params. Or maybe the redirect error should be configured in the provider config? 🤔 But we'll want to use the built-in UI _if_ you're coming from the built-in UI, so maybe that means search params are the way to go there.

Furthermore, if the redirect or token is missing from the search params, I guess we'll need to render an HTML error page?

scotttrinh: Yeah, I guess we do all of this in the reset password handler, so I'll just follow that same pattern here.

jaclarke (approved):

PR #6832, opened 2024-02-13
Fix f-string style flakes in auth UI

elprans:
The upcoming upgrade to Python 3.12 (with PEP 701) and corresponding
`pycodestyle` bump, there are a bunch of (mostly just annoying) flakes
in multiline f-strings.  The `auth_ext/ui.py` is by far the biggest
source, so I'm splitting formatting fixes out into a standalone commit
for ease of review.

### DIFF
--- a/edb/server/protocol/auth_ext/ui.py
+++ b/edb/server/protocol/auth_ext/ui.py
@@ -36,6 +36,35 @@ known_oauth_provider_names = [
 ]
 
 
+def _render_oauth_buttons(
+    oauth_providers: list,
+    oauth_params: dict[str, str],
+    oauth_label: str,
+) -> str:
+    if not oauth_providers:
+        return ""
+
+    oauth_buttons = [
+        {
+            "authparams": urllib.parse.urlencode(
+                {'provider': p.name, **oauth_params}
+            ),
+            "image": (
+                f'<img src="_static/icon_{p.name[15:]}.svg"'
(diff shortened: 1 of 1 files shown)


edb/server/protocol/auth_ext/ui.py
```diff
@@ -36,6 +36,35 @@
...
+        return ""
+
+    oauth_buttons = [
+        {
+            "authparams": urllib.parse.urlencode(
+                {'provider': p.name, **oauth_params}
+            ),
+            "image": (
+                f'<img src="_static/icon_{p.name[15:]}.svg"'
+                f'alt="{p.display_name} Icon" />'
```
scotttrinh, line 54: We need a space here, not sure which makes more sense:

```suggestion
                f'<img src="_static/icon_{p.name[15:]}.svg" '
                f'alt="{p.display_name} Icon" />'
```

or

```suggestion
                f'<img src="_static/icon_{p.name[15:]}.svg"'
                f' alt="{p.display_name} Icon" />'
```

scotttrinh (changes requested):
Only saw one small thing.

edb/server/protocol/auth_ext/ui.py
elprans: Fixed.

scotttrinh (approved):

PR #6835, opened 2024-02-13
Fix `test_signalctl_wait_for_09`

elprans:
Under Python 3.12+ `stdin.readline()` is returning `EOF` after the process
is terminated; let's replace it with a infinite future.


fantix (approved):
Thanks!

Issue #6732, opened 2024-01-25
Allow setting app name and other details even without using the built-in auth UI


scotttrinh:
We currently only allow setting the app name and details like logos and colors when using the built-in UI. However, we can and should use these details in other places like in verification emails.

scotttrinh:
Done in #6754

PR #6844, opened 2024-02-14
Fall back to `ui_config` for getting app details

scotttrinh:
We plan on keeping the `ui_config` around for now, so allow falling back.

See also #6732


msullivan (approved):
This looks reasonable to me. (Though also it fails tests, so doesn't seem to quite be there somehow)

Issue #6705, opened 2024-01-17
Query cache is keyed on all session config, not just the ones that affect compilation


msullivan:
This is probably kind of marginal, since there aren't *that* many non-compilation options frequently used as session configs (session_idle_transaction_timeout mostly), but they are all used in the key, which might mean more recompiles than needed.

(I realized this because I forgot to put `affects_compilation` on the debugging configs I added recently and they still worked)

PR #6847, opened 2024-02-14
Cache by schema version

fantix:
Drop integer `dbver` and replace it with the UUID `schema_version`, so that we can persist the query cache under consistent keys in the next steps.

`dbver` was bumped on config changes but `schema_version` is only for user schema. We handle this difference by adding such config into the compile cache key / PG state key.

`get_session_config()` for cache key [should be optimized](https://github.com/edgedb/edgedb/issues/6705) to honor `affects_compilation`, this is fixed in a future persistent cache PR.

### DIFF
--- a/edb/server/dbview/dbview.pyx
+++ b/edb/server/dbview/dbview.pyx
@@ -245,15 +245,17 @@ cdef class Database:
             # We already have a cached query for a more recent DB version.
             return
 
+        # Store the matching schema version, see also the comments at origin
         self._eql_to_compiled[key] = compiled, schema_version
 
-    def cache_compiled_sql(self, key, compiled: list[str]):
+    def cache_compiled_sql(self, key, compiled: list[str], schema_version):
         existing, ver = self._sql_to_compiled.get(key, DICTDEFAULT)
         if existing is not None and ver == self.schema_version:
             # We already have a cached query for a more recent DB version.
             return
 
-        self._sql_to_compiled[key] = compiled, self.schema_version
+        # Store the matching schema version, see also the comments at origin
+        self._sql_to_compiled[key] = compiled, schema_version
 
     def lookup_compiled_sql(self, key):
         rv, cached_ver = self._sql_to_compiled.get(key, DICTDEFAULT)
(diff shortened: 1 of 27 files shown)


edb/server/dbview/dbview.pyx
```diff
@@ -542,9 +537,15 @@ cdef class DatabaseConnectionView:
...
                 'no need to serialize state while in transaction')
 
-        dbver = self._db.dbver
+        ver = hash(
+            (
+                self._db.schema_version,
+                self.get_database_config(),
+                self.get_system_config(),
+            )
+        )
```
msullivan, line 546: I don't think that the python `hash` function is reliable enough to use as a cache key

msullivan: (I think I like the third option best?)

msullivan (approved):
This looks good, modulo the concerns I have about the changes to `_config_cache` invalidation

edb/server/dbview/dbview.pyx
```diff
@@ -557,12 +558,12 @@ cdef class DatabaseConnectionView:
                 state.append({"name": sval.name, "value": jval, "type": kind})
 
         # Include the database version in the state so that we are forced
-        # to clear the config cache on dbver changes.
+        # to clear the config cache on changes.
         state.append(
-            {"name": '__dbver__', "value": dbver, "type": 'C'})
+            {"name": '__ver__', "value": ver, "type": 'C'})
```
elprans, line 563: Where are we actually reading this?

msullivan: https://github.com/edgedb/edgedb/blob/669ba34661ad04a605b12f745fa4b92122557570/edb/server/protocol/execute.pyx#L88-L91 and https://github.com/edgedb/edgedb/blob/669ba34661ad04a605b12f745fa4b92122557570/edb/server/protocol/execute.pyx#L253-L256 are the relevant places.

It doesn't get used directly, but changes in it are used to force a state sync

edb/server/dbview/dbview.pyx
fantix: In that case, I think it's easier if we keep the `dbver`, but only use schema_version for compile cache.

msullivan (approved):

PR #6848, opened 2024-02-14
Cleanup of pgvector.

vpetrovykh:
Drop undocumented `set_probes` and `_get_probes` in favor of actual configarion parameters.

We now have `probes` and `ef_search` config parameters that effectively expose the corresponding parameters from the postgres extension.

Add tests for the exposed config parameters.

### DIFF
--- a/edb/lib/_testmode.edgeql
+++ b/edb/lib/_testmode.edgeql
@@ -359,6 +359,15 @@ std::_datetime_range_buckets(
     $$;
 };
 
+
+CREATE FUNCTION
+std::_current_settings(sqlname: str) -> OPTIONAL std::int64 {
+    USING SQL $$
+      SELECT nullif(current_setting(sqlname, true), '')::int8
+    $$;
+};
+
+
 CREATE MODULE std::_test;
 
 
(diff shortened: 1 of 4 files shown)


edb/lib/_testmode.edgeql
```diff
@@ -359,6 +359,15 @@ std::_datetime_range_buckets(
     $$;
 };
 
+
+CREATE FUNCTION
+std::_current_settings(sqlname: str) -> OPTIONAL std::int64 {
+    USING SQL $$
+      SELECT nullif(current_setting(sqlname, true), '')::int8
+    $$;
+};
```
msullivan, line 368: Let's make it return `str` and drop the `nullif` bit. If we're going to make it generic, we should

edb/lib/ext/pgvector.edgeql
```diff
@@ -122,20 +147,6 @@ create extension package pgvector version '0.5.0' {
         set force_return_cast := true;
     };
 
-    create function ext::pgvector::set_probes(num: std::int64) -> std::int64 {
-	using sql $$
-            select num from (
-	        select set_config('ivfflat.probes', num::text, true)
-            ) as dummy;
-	$$;
-    };
```
msullivan, line 131: We should keep `set_probes` but mark it as deprecated.

msullivan (approved):

PR #6851, opened 2024-02-14
Switch to using Github's runners for macOS builds

elprans:
Those are now available, and though a tad slower than our current hosted
runners, allow us to avoid having to maintain hosted macOS runners.


msullivan (approved):
I am very happy about this

PR #6852, opened 2024-02-14
Properly report errors involving newly created types

### LINKED ISSUE #6845: `MissingRequiredError` no longer thrown. Now it's just an ISE
I noticed recently that I no longer get the `MissingRequiredError` like mentioned [here](https://www.edgedb.com/docs/guides/migrations/tips#adding-a-required-link) when trying to apply a migration that doesn't properly set a newly required property. Instead, the migration fails with the following error:

```
edgedb error: InternalServerError: RuntimeError: unhandled error while calling interpret_backend_error(); run with EDGEDB_DEBUG_SERVER to debug.
  Server traceback:
      edb.errors.InternalServerError: RuntimeError: unhandled error while calling interpret_backend_error(); run with EDGEDB_DEBUG_SERVER to debug.
```

I tested this out with edgedb versions `4.5`, `4.4`, `4.3`, `4.0`, and `3.5`, so I think it's an issue introduced by the recent CLI `4.1` release.

- EdgeDB Version:
- EdgeDB CLI Version:
- OS Version:

Steps to Reproduce:

1. Init a project with the below schema.
2. Seed some data without lastName's.
```edgeql
for name in { 'hello', 'world'} union ( insert User { name := name } );
```
3. Uncomment the required lastName property.
4. Attempt migration.

Schema:

```edgeql
module default {
    type User {
        name: str;
        # required lastName: str;
    }
}
```

msullivan:
Currently, if we run a DDL command that errors, and we are not in a
transaction, we will ISE instead of interpreting the error correctly.
This is because the updated user schema doesn't get stored anywhere,
because we weren't in a transaction. When inside a transaction, it
would be stored in the dbview.

I originally fixed this by using implicit transactions to keep track
of the new user schema, but this required keeping implicit
transactions open on failures, and in opening up an implicit
transaction even in single statement execution of DDL, and that all
seemed questionable.

I wound up fixing it by stashing the schema in the exception object
and extracting it in interpret_error.

Fixes #6845.

### DIFF
--- a/edb/server/protocol/execute.pyx
+++ b/edb/server/protocol/execute.pyx
@@ -182,6 +182,11 @@ async def execute(
                 await dbv.apply_config_ops(be_conn, config_ops)
 
     except Exception as ex:
+        # If we made schema changes, include the new schema in the
+        # exception so that it can be used when interpreting.
+        if query_unit.user_schema:
+            ex._user_schema = query_unit.user_schema
+
         dbv.on_error()
 
         if query_unit.tx_commit and not be_conn.in_tx() and dbv.in_tx():
@@ -358,6 +363,10 @@ async def execute_script(
 
     except Exception as e:
         dbv.on_error()
+
+        # Include the new schema in the exception so that it can be
+        # used when interpreting.
+        e._user_schema = dbv.get_user_schema_pickle()
 
         if not in_tx and dbv.in_tx():
             # Abort the implicit transaction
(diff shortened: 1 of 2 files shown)


edb/server/protocol/execute.pyx
```diff
@@ -182,6 +182,11 @@ async def execute(
                 await dbv.apply_config_ops(be_conn, config_ops)
 
     except Exception as ex:
+        # If we made schema changes, include the new schema in the
+        # exception so that it can be used when interpreting.
+        if query_unit.user_schema:
+            ex._user_schema = query_unit.user_schema
```
elprans, line 188: Given that we only need this for `pgerror.BackendError`, should we check for that before mutating it?

elprans: I feel kind-of uneasy about attempting to `setattr` something on a random exception object.

msullivan: Yeah, that's probably a good idea

elprans (approved):

PR #6856, opened 2024-02-15
Fix cors tests not resetting instance config

jaclarke:
Should fix the flaky tests disabled here: https://github.com/edgedb/edgedb/pull/6850

### DIFF
--- a/tests/test_http_auth.py
+++ b/tests/test_http_auth.py
@@ -28,7 +28,6 @@ from edb import protocol
 from edb.server import defines as edbdef
 from edb.testbase import server as tb_server
 from edb.testbase import http as tb_http
-from edb.tools import test
 
 
 class BaseTestHttpAuth(
@@ -129,6 +128,8 @@ class BaseTestHttpAuth(
 
 
 class TestHttpAuth(BaseTestHttpAuth):
+    PARALLELISM_GRANULARITY = 'system'
+
     def test_http_auth_scram_valid(self):
         args = self.get_connect_args()
         (token, headers, status, sid, expected_server_sig) = self._scram_auth(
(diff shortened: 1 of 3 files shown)


tests/test_http_auth.py
```diff
@@ -203,64 +204,70 @@ def test_http_auth_scram_invalid_password(self):
...
+                    'WWW-Authenticate',
+                    headers['access-control-expose-headers']
+                )
+                self.assertIn(
+                    'Authentication-Info',
+                    headers['access-control-expose-headers']
+                )
+        finally:
+            await self.con.execute(
+                'configure instance reset cors_allow_origins')
```
scotttrinh, line 269: Does it make more sense to (also?) `reset` before the initial assertion to move the default state closer to the assertion to make it less likely there is a race condition? I guess the only way to get a totally isolated test run would be to somehow spin up another instance, so maybe it's not worth going too overboard.

tests/test_http_auth.py
```diff
@@ -203,64 +204,70 @@ def test_http_auth_scram_invalid_password(self):
...
                 'Authentication-Info',
-                headers['access-control-expose-headers']
+                headers['Access-Control-Expose-Headers']
             )
 
+            with self.http_con() as con:
+                _, headers, status = self.http_con_request(
+                    con, {}, path="token",
+                    headers={'Origin': 'https://example.edgedb.com'}
+                )
```
scotttrinh, line 255: Since we're in here, is there a reason we use `urllib.request` in the first case, but `http_con` here?

scotttrinh (approved):
No blockers, and this does seem right to me. 👍

tests/test_http_auth.py
jaclarke: I don't know how the config system works enough to really know what race conditions are possible, but if the tests can't overlap because of the parallelism option, then I'm not sure I understand what the difference is between the config reset being at the start of the same test, versus at the end of an immediately preceding test? If a race can happen aren't those cases pretty much the same?

scotttrinh: > but if the tests can't overlap because of the parallelism option

Yeah you're right that this is honestly moot. Definitely not a blocker!

I think the thing that stuck out to me was my normal thought process when looking at a test case, which is:

1. Set up the state we're testing.
2. Make assertions.
3. (Optional) Reset/teardown

So in the case of the first block of assertions, I'd expect:

1. Setup: ensure default value for `cors_allow_origins`.
2. Assert: no `Access-Control-Allow-Origin` header.

msullivan: Reset/teardown is unfortunately not optional for this kind of test case, which is modifying global state on a shared edgedb server. I think the current structure is right.

msullivan (approved):
This is fine to merge now, since I don't want this blocking anything.

But as I think about it more, I think we should probably make this use `tb.start_edgedb_server` so that any testing that uses `instance` configs is being done against an isolated server.

tests/test_http_auth.py
scotttrinh: > Reset/teardown is unfortunately not optional for this kind of test case

Yeah, I just meant optional in general the general case: if your "setup" just involves local state, there is nothing to teardown.

> which is modifying global state on a shared edgedb server. I think the current structure is right.

My question was, should we _also_ reset the state at the beginning to ensure that the state is as-expected at the site of the expectation/assertion?

Fine if not, and really just a suggestion based on reading through the code and noticing that we're asserting on an implicit assumption about global state: that the instance config for `allowed_cors_origins` is in the initial state.

msullivan: The reason I think it is fine is because I think it's totally reasonable to test what the default state is, here

scotttrinh: Yeah, but as we found out, who knows what other tests might've changed the default state since this is shared state, right? Not worth too much bikeshedding, but I'm just thinking about applying this principle more generally: setup your expectation explicitly, then assert. If reading either of these tests in isolation, I find it helpful to see: "Ahh, we're testing what the behavior is when the `allowed_cors_origins` config value is in it's default state." And also reseting the value again at the end of the test to be a good citizen.

We've certainly fixed this case with the blunt tool of serializing them, and I think we should just merge this, just more of a general discussion at this point.

msullivan:
I went and switched to using `start_edgedb_server` for the auth test. For the http test, that was going to be a little more annoying, and there doesn't seem to be any major benefit to testing the instance configuration stuff twice. (And start_edgedb_server is pretty slow, too)

PR #6857, opened 2024-02-15
Fix in_tx_dbver

fantix:
Looks like we had in_tx_dbver since [2021](https://github.com/edgedb/edgedb/pull/2183/files#r1491189100) but never really used it. The added test demonstrates an issue where we cached a query under the wrong dbver in a transaction, and caused ISE in later calls.


elprans (approved):

aljazerzen:
@fantix, I'm not able to backport this.

In the test, the last query fails with missing column too. See back-ported commit 7dc2b35a9.

I've tried first backporting #6819, but it's test is failing too.

Is this fixable, or do we just don't backport it?

PR #6863, opened 2024-02-16
Make CREATE EXTENSION ... VERSION more flexible

msullivan:
It will now select something >= the specified version while still
matching the major version.

In the future we can start adding qualifiers to the string to be more
or less specific.

Doing something like this is desired because existing migrations have
pgvector 0.4.0 written in them, which we'd prefer to not have exist in
edgedb 5.0.

### DIFF
--- a/edb/schema/extensions.py
+++ b/edb/schema/extensions.py
@@ -196,7 +196,10 @@ def get_package(
     ]
     if version is not None:
         filters.append(
-            lambda schema, pkg: pkg.get_version(schema) == version,
+            lambda schema, pkg: (
+                pkg.get_version(schema) >= version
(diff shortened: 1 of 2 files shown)


vpetrovykh:
Matching the major version is a bit arbitrary, but I kinda understand why semantically this makes sense as we generally expect minor versions to be compatible, whereas major ones might break that.

I guess that by the time major version rolls out for `pgvector` we should have a plan for how that affects old migrations that start off at 0.4.0.

elprans (approved):

vpetrovykh (approved):

scotttrinh (approved):

PR #6865, opened 2024-02-16
Output all ExceptionGroup sub-exceptions on failure

msullivan:
Our custom MultiError would include all of the sub-exceptions as part of
its message, so even though they wouldn't get the markup/renderer 
treatment, they would always be output. The switch to using the standard
library TaskGroup and ExceptionGroup broke this, so failures in a TaskGroup
would be opaque.

Fix this by teaching the serializer about ExceptionGroup. I wanted to get
things indented nicely, so I added a SubNode field that indents its body
without doing anything else, and then I realized that if I wanted headers
to look good we would need to shorten them when the line has been indented.

Doing that required fixes a TODO in the terminal renderer that dates back
to 2012, from 84d9febd1b!

### DIFF
--- a/edb/common/markup/elements/doc.py
+++ b/edb/common/markup/elements/doc.py
@@ -40,6 +40,10 @@ class Section(DocMarkup):
     collapsed = Field(bool, coerce=True, default=False)
 
 
+class SubNode(DocMarkup):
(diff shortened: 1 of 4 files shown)


1st1 (approved):
oh dear...

PR #6868, opened 2024-02-16
Initial schema sketch for email templates

scotttrinh:


edb/lib/ext/auth.edgeql
```diff
@@ -299,6 +299,33 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
...
+        };
+        create required property html: std::str {
+            create annotation std::description :=
+                "Body of the email.";
+        };
+        create required property plain_text: std::str {
+            create annotation std::description :=
+                "Body of the email in plain text.";
+        };
+    };
```
scotttrinh, line 315: @elprans I wanted to check and see if this is what you had in mind based on our discussion earlier today:

```suggestion
    create type ext::auth::EmailTemplate extending cfg::ConfigObject {
        create required property purpose: std::str {
            create constraint exclusive;
        };
        create required property subject: std::str {
            create annotation std::description :=
                "Subject of the email.";
        };
        create required property html: std::str {
            create annotation std::description :=
                "Body of the email.";
        };
        create required property plain_text: std::str {
            create annotation std::description :=
                "Body of the email in plain text.";
        };
    };
```

And then adding `multi email_templates: ext::auth::EmailTemplate;` to the existing `AuthConfig` object?

elprans: One advantage of the original link approach is that we can populate some default and make them `required` for robustness?

scotttrinh: Yeah, I thought about `default` here, but I was wondering if a huge blob of HTML is really something we want to have in the HTML. Also might be nice to have the default HTML be something that we "upgrade" over time without having to require a schema change.

PR #6872, opened 2024-02-16
Add annotation on internal config that affects compilation

fantix:
We will honor `affects_compilation` for including (session) config values in the compile request, so `__internal_testmode` should be included by then.


elprans (approved):
Don't forget to bump catver

PR #6873, opened 2024-02-16
Magic link fixes

jaclarke:


edb/server/protocol/auth_ext/local.py
```diff
@@ -102,15 +102,20 @@ async def get_verified_by_identity_id(self, identity_id: str) -> str | None:
...
         r = await execute.parse_execute_json(
             self.db,
-            """
+            f"""
 with
     email := <str>$email,
     identity := (
         select ext::auth::LocalIdentity
-        filter .<identity[is ext::auth::EmailFactor].email = email
+        filter .<identity[is ext::auth::{factor_type}].email = email
```
scotttrinh, line 118: I was a little confused by this at first, but now I get it: when there are multiple identities with the same email, we want this specific identity type.

I wonder if we should just move this method to the subclasses since we pretty much _always_ want "this specific factor type". 🤔

PR #6875, opened 2024-02-16
Remove use of `syphar/restore-virtualenv`

elprans:
Use regular `setup-python`'s caching capabilities in tandem with `pip-compile`.


fantix (approved):
Looks good and installation from cached package tarballs is nice too

PR #6880, opened 2024-02-16
Add ADMINISTER command to vacuum.

### LINKED ISSUE #6663: Expose `VACUUM [FULL]` as an `administer` function
`VACUUM` is useful for when autovacuum is not enough, and `VACUUM FULL` can be used to reclaim disk space after a wide-ranging update on a relation.

vpetrovykh:
Calling `administer vacuum(SomeType, OtherType.ptr, full := true)` will vacuum the table for `SomeType` and all its descendants, as well as the link table associated with the `ptr` (assuming that `ptr` is multi) and all its descendants. The named only argument `full` being set to `true` indicates the `FULL` option.

All of these arguments can be omitted. Not specifying any types or pointers to vacuum will result in vacuuming everything that is accessible to the user.

Co-authored-by: Michael J. Sullivan <sully@msully.net>

Fixes #6663

### DIFF
--- a/edb/server/compiler/ddl.py
+++ b/edb/server/compiler/ddl.py
@@ -1404,3 +1404,153 @@ def administer_reindex(
     return dbstate.MaintenanceQuery(
         sql=tuple(q.encode('utf-8') for q in commands)
     )
+
+
+def administer_vacuum(
+    ctx: compiler.CompileContext,
+    ql: qlast.AdministerStmt,
+) -> dbstate.BaseQuery:
+    from edb.ir import ast as irast
+    from edb.ir import typeutils as irtypeutils
+    from edb.schema import objtypes as s_objtypes
+    from edb.schema import pointers as s_pointers
+
+    # check that the kwargs are valid
+    kwargs: Dict[str, str] = {}
+    for name, val in ql.expr.kwargs.items():
+        if name != 'full':
+            raise errors.QueryError(
+                f'unrecognized keyword argument {name!r} for vacuum()',
+                context=val.context,
+            )
+        elif not isinstance(val, qlast.BooleanConstant):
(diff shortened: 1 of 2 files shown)


edb/server/compiler/ddl.py
```diff
@@ -1404,3 +1404,132 @@ def administer_reindex(
...
+            ):
+                ptr = False
+            case qlast.Path(
+                steps=[qlast.ObjectRef(), qlast.Ptr()],
+                partial=False,
+            ):
+                ptr = True
+            case _:
+                raise errors.QueryError(
+                    f'argument to vacuum() must be an object type',
```
msullivan, line 1453: or link or property reference?

edb/server/compiler/ddl.py
```diff
@@ -1404,3 +1404,132 @@ def administer_reindex(
...
+    for arg, (rptr, obj) in zip(ql.expr.args, args):
+        if not rptr:
+            # On a type, we just vacuum the type and its descendants
+            tables.update({obj} | {
+                desc for desc in obj.descendants(schema)
+                if desc.is_material_object_type(schema)
+            })
+        else:
+            # On a pointer, we must go over the pointer and its descendants
+            # and convert them to table and colum format.
```
msullivan, line 1503: I think this comment is not right

edb/server/compiler/ddl.py
```diff
@@ -1404,3 +1404,132 @@ def administer_reindex(
...
+            if not isinstance(rptr.ptrref, irast.PointerRef):
```
msullivan, line 1504: What happens if we try it on a single property? I'm worried it will ISE

msullivan (approved):

msullivan (approved):

PR #6881, opened 2024-02-16
Implement persistent query cache

fantix:


fantix:
@msullivan or @elprans, I think I can use some help here with the `record` type again. 🙏 

At this moment, there's the `record` error at the bottom running on this PR:

```
_localdev:edgedb> create type User { create property name: str; };
OK: CREATE TYPE

_localdev:edgedb> insert User { name := 'Alice' };
{default::User {id: 42603752-cd2f-11ee-912d-5f636ada9c1e}}

_localdev:edgedb> select User;
{default::User {id: 42603752-cd2f-11ee-912d-5f636ada9c1e}}

_localdev:edgedb> delete (select User);
edgedb error: InvalidValueError: record type has not been registered
```

And Postgres logs:

```
ERROR 387503 2024-02-16T19:55:50.172 postgres: record type has not been registered
INFO 387503 2024-02-16T19:55:50.172 postgres: SELECT/*<pg.SelectStmt at 0x7948ba81bb90>*/
INFO 387503 2024-02-16T19:55:50.172 postgres: 	        edgedb.__qh_9019920f4e0ae1852e7f1e9b26938d6e()
```

@elprans previously suggested running this SQL like this:

```
SELECT * FROM edgedb.__qh_9019920f4e0ae1852e7f1e9b26938d6e() AS q(result record)
```

That works fine for the `delete`, but it fails with the `select` query. I have a patch to generate this SQL:

```patch
diff --git a/edb/server/compiler/compiler.py b/edb/server/compiler/compiler.py
index 826a1dbf6..b34a59cd7 100644
--- a/edb/server/compiler/compiler.py
+++ b/edb/server/compiler/compiler.py
@@ -1816,6 +1816,7 @@ def _compile_ql_query(
             returns='',
             text=sql_text,
         )
+        returns_record = False

         match ctx.output_format:
             case enums.OutputFormat.NONE:
@@ -1825,6 +1826,7 @@ def _compile_ql_query(
                 func.set_returning = ir.cardinality.is_multi()
                 if ir.stype.is_object_type():
                     func.returns = 'record',
+                    returns_record = True
                 else:
                     func.returns = pg_types.pg_type_from_ir_typeref(
                         ir.expr.typeref)
@@ -1861,9 +1863,16 @@ def _compile_ql_query(
                 )
             ],
         )
-        sql_text = pg_codegen.generate_source(
-            pgast.SelectStmt(target_list=[pgast.ResTarget(val=func_call)])
-        )
+        if returns_record:
+            # libpg_query doesn't parse `AS q(result record)`, and I didn't find a proper AST
+            func_call_sql = pg_codegen.generate_source(func_call)
+            sql_text = (
+                f"SELECT * FROM {func_call_sql} AS q(result record)"
+            )
+        else:
+            sql_text = pg_codegen.generate_source(
+                pgast.SelectStmt(target_list=[pgast.ResTarget(val=func_call)])
+            )
         cache_sql = (
             cf.to_string().encode(defines.EDGEDB_ENCODING),
             df.to_string().encode(defines.EDGEDB_ENCODING),
```

I think `returns_record` needs to be more accurate to only address the Postgres issue (the `delete` case). I looked into our IR, but it seems the returning shapes for either `delete` or `select` are identical. So my question is, can we predict the Postgres issue with any information from our IR?

msullivan:
That's strange, it seems like the same type is being returned in both cases.

msullivan:
I do not understand at all what the difference between the `select` and `delete` cases are once your patch is applied.
Both `RETURNS SETOF record`, both return something that looks like `ROW("m~2"."expr~1_value~1") AS "expr~1_serialized~1"`, where the value is a `uuid`.
But the `select` one errors on ` AS q(result record)`, though it succeeds on ` AS q(result uuid)`.
The `delete` one works on ` AS q(result record)`, but errors on ` AS q(result uuid)`.

I am really failing to see any meaningful difference here, though.
@elprans does this make any sense to you?

msullivan:
It feels to me like there is a postgres bug around here.

fantix:
> It feels to me like there is a postgres bug around here.

Yeah, that's also what Elvis said.

If it's out of our hands, I think I'll have to find a different way to define/call the functions.

msullivan:
Should we just cache the text and not bother with functions?

msullivan:
Another option might be to actually generate composite types for everything returned by queries. That seems kind of miserable, though.

elprans:
Which query is failing exactly?  I'm not getting any errors on either `delete` or `select` (after applying ths workaround patch).

elprans:
Hm, though the `insert` query is failing with `function return row and query-specified return row do not match`, which seems fixable.

fantix:
Oh maybe you need data in the table, empty table works fine with both the select and insert query

elprans:
I pushed a formulation that seems to work.  Also, the AST you were looking for is `coldeflist` on `FuncCall`

msullivan:
Let's make sure to document this

msullivan:
It seems like we don't populate the cache in transactions, which means that the test suite is not going to be exercising the cache very much.

fantix:
> It seems like we don't populate the cache in transactions, which means that the test suite is not going to be exercising the cache very much.

Yeah, that seemed to be the case for some time already, I do have a TODO in the pre-RFC to add the in-transaction cache back.

edb/pgsql/compiler/context.py
```diff
@@ -304,6 +304,10 @@ class CompilerContextLevel(compiler.ContextLevel):
     #: needed by DML.
     shapes_needed_by_dml: Set[irast.Set]
 
+    #: Aggregated output of detached SQL parameters (position integer) and
+    #: their SQL types (qualified names tuple)
+    detached_params: Dict[int, Tuple[str, ...]]
+
```
msullivan, line 310: What is a detached parameter?

fantix: We normally compile in a way that the types of input parameters are placed inline, like

```sql
... where users.name = $3::text
```

but in functions, we'd have to define the type of parameters in the function signature, like:

```sql
create function find_user(name: text)
...
begin
    ... where users.name = $name
...
```

`detached_params` is just my way to find those types of parameters.

edb/server/protocol/execute.pyx
```diff
@@ -255,8 +344,14 @@ async def execute_script(
         orig_state = state = dbv.serialize_state()
 
     data = None
+    recompile_requests = None
 
     try:
+        if debug.flags.func_cache:
+            if any(query_unit.user_schema for query_unit in unit_group):
+                # TODO(fantix): do this in the same transaction
+                recompile_requests = await dbv.clear_cache_keys(conn)
```
msullivan, line 353: We should still try to do recompiles with the text cache, right? We could drive it straight from the in-memory table, I think

msullivan:
Some other things we might want:
- [ ] A way to disable recompilation
- [ ] A way for the user to forcibly clear the cache. Probably an `administer` command.

Thing two is mostly something that *I* need, but arguably these two things can be the same, and the way to "disable recompilation" is just to clear the cache first.

edb/common/debug.py
```diff
@@ -176,6 +176,8 @@ class flags(metaclass=FlagsMeta):
 
     zombodb = Flag(doc="Enabled zombodb and disables postgres FTS")
 
+    func_cache = Flag(doc="Use stored functions for persistent cache")
```
msullivan, line 179: Maybe document that it is experimental and doesn't work yet fully

edb/server/compiler/compiler.py
```diff
@@ -1822,17 +1836,133 @@ def _compile_ql_query(
...
+        script_info is None
+    )
+
     sql_res = pg_compiler.compile_ir_to_sql_tree(
         ir,
         expected_cardinality_one=ctx.expected_cardinality_one,
         output_format=_convert_format(ctx.output_format),
         backend_runtime_params=ctx.backend_runtime_params,
         expand_inhviews=options.expand_inhviews,
+        detach_params=bool(use_persistent_cache and debug.flags.func_cache),
```
msullivan, line 1854: Shouldn't need the `bool`, right?

edb/server/compiler/compiler.py
```diff
@@ -1822,17 +1836,133 @@ def _compile_ql_query(
...
     result_cardinality = enums.cardinality_from_ir_value(ir.cardinality)
 
+    # This low-hanging-fruit is temporary; persistent cache should cover all
+    # cacheable cases properly in future changes.
+    use_persistent_cache = (
+        ctx.persistent_cache and
+        cacheable and
+        not ctx.bootstrap_mode and
+        script_info is None
+    )
```
msullivan, line 1846: I don't think this is worth your time to change everywhere now, and we aren't fully consistent, but the style in the compiler tends towards putting the `and` at the start of lines

edb/server/compiler/compiler.py
```diff
@@ -1822,17 +1836,133 @@ def _compile_ql_query(
...
+            )
+        else:
+            sql_ast = pgast.SelectStmt(
+                target_list=[pgast.ResTarget(val=func_call)],
+            )
+
+        cache_sql = (
+            cf.to_string().encode(defines.EDGEDB_ENCODING),
+            df.to_string().encode(defines.EDGEDB_ENCODING),
+        )
```
msullivan, line 1963: Put all this into a helper function

edb/server/compiler/compiler.py
```diff
@@ -1822,17 +1836,133 @@ def _compile_ql_query(
...
         output_format=_convert_format(ctx.output_format),
         backend_runtime_params=ctx.backend_runtime_params,
         expand_inhviews=options.expand_inhviews,
+        detach_params=bool(use_persistent_cache and debug.flags.func_cache),
     )
 
-    sql_text = pg_codegen.generate_source(sql_res.ast)
+    sql_ast = sql_res.ast
+
+    pg_debug.dump_ast_and_query(sql_ast, ir)
```
elprans, line 1859: Remove this dump

edb/server/dbview/dbview.pyx
```diff
@@ -887,6 +920,79 @@ cdef class DatabaseConnectionView:
...
+            ).deserialize(row[0], "<unknown>")
+            rv.append(query_req)
+            self._db._eql_to_compiled.pop(query_req, None)
+        return rv
+
+    async def recompile_all(
+        self, conn, requests: typing.Iterable[rpc.CompilationRequest]
+    ):
+        compiler_pool = self.server.get_compiler_pool()
+        concurrency = max(1, compiler_pool.get_size_hint() - 1)
```
elprans, line 938: This needs to be at most half of the compiler pool to avoid blocking live queries while cache refresh is being done.

edb/server/defines.py
```diff
@@ -53,6 +53,7 @@
 BACKEND_COMPILER_TEMPLATE_PROC_RESTART_INTERVAL = 1
 
 _MAX_QUERIES_CACHE = 1000
+_MAX_QUERIES_CACHE_DB = 10_000
```
elprans, line 56: This probably should stay at 1000 now that we keep SQL text in memory.

edb/server/dbview/dbview.pyx
```diff
@@ -887,6 +920,79 @@ cdef class DatabaseConnectionView:
...
+                    )
+
+        async with asyncio.TaskGroup() as g:
+            for _ in range(concurrency):
+                g.create_task(recompile_request())
+            g.create_task(persist_cache_task())
+            for data in requests:
+                await i.put(data)
+            for _ in range(concurrency):
+                await i.put(None)
```
msullivan, line 994: The code would be cleaner if we spun off one task per request and used a semaphore to limit concurrency, right?

I guess that would use more memory, though?

msullivan (approved):

edb/server/compiler/compiler.py
fantix: Right if we can let mypy think `Flag` is a `bool`:

```
edb/server/compiler/compiler.py:1854:23: error: Argument "detach_params" to "compile_ir_to_sql_tree" has incompatible type "Flag | bool"; expected "bool"  [arg-type]
```

edb/server/dbview/dbview.pyx
```diff
@@ -191,6 +194,14 @@ cdef class Database:
 
         # Store the matching schema version, see also the comments at origin
         self._eql_to_compiled[key] = compiled, schema_version
+        keys = []
+        while self._eql_to_compiled.needs_cleanup():
+            keys.append(self._eql_to_compiled.cleanup_one().get_cache_key())
+        if keys:
+            self.tenant.create_task(
+                self.tenant.evict_query_cache(self.name, keys),
+                interruptable=False,
```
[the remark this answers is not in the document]
fantix: Ah, script cache! Now fixed and I'll add a test to cover this branch.

edb/server/dbview/dbview.pyx
fantix: Not too much more memory, I suppose; all the requests are already in memory.

msullivan: Yeah, just I think a task is bigger than a queue entry

edb/server/compiler/compiler.py
msullivan: Please put this dump back; it's conditional in `pg_debug`

msullivan: Actually I just did it.
(I was doing some testing of something unrelated on top of this branch, to avoid needing to rebuild, and ran into this being gone)

PR #6883, opened 2024-02-17
Report topological cycle errors in migrations as real errors

msullivan:


elprans (approved):

aljazerzen:
Do we actually have a case where this error is needed? The few times I've seen this error it turned out to be a bug in ref collection.

PR #6887, opened 2024-02-19
Cache `libpg_query` builds in CI

elprans:
This was left as a TODO, but never done.  Shaves ~30s from the `build`
job.


fantix (approved):
👍

PR #6893, opened 2024-02-20
Docs: Document HTTP and GQL usage for EdgeDB Cloud users

raddevon:
This was inspired by [a recent Discord thread](https://discord.com/channels/841451783728529451/1208369085893845002) in which a user asked how to do this on EdgeDB Cloud.

While I was in here, I discovered that some of our HTTP documentation has been hidden because of the placement of the TOC. (Honestly, not sure why the placement matters, but it seems to.)


1st1 (commented):
We also need to add a page to our Cloud guide.

scotttrinh (approved):

PR #6900, opened 2024-02-21
Replace flake8 with ruff

aljazerzen:
Replaces flake8 with ruff. Also fixes a few lint errors that show up because of that.

Total time on my machine:
- flake 1.415 sec
- ruff 0.05 sec

The important question: what have I done with the `from typing import *` custom extension that we had? I've disabled the rule. I plan to change all `from typing import *` into `from typing import List, Dict` before merging this PR.


aljazerzen:
Re `import *`: A pretty good way of doing this would be search and replace with all names in `typing`:
```
from typing import Any, ClassVar, Generic, Optional, Tuple, Type, TypeVar, Union, AbstractSet, Iterable, Iterator, MutableMapping, MutableSequence, MutableSet, Sequence, Dict, List, Set, cast, overload
```
... and then running `ruff check --fix .`, which would remove unused imports.

aljazerzen:
https://github.com/edgedb/edgedb/pull/6874 has a few problems with the new f-string syntax and flake8, upgrading to ruff fixes that.

msullivan:
Should we investigate whether ruff would take a patch to support `from typing import *` nicely?

msullivan (approved):
I guess. I really do hate the typing stuff though.

pyproject.toml
```diff
@@ -213,3 +210,19 @@ ignore_errors = true
...
+    "B008",
+    "B904",
+    "B905",
+    "B019",
+    "E731",
+    "E741",
+]
+line-length = 80
+indent-width = 4
+target-version = "py310"
```
elprans, line 227: No need for this, because `ruff` looks at `requires-python` that we already have above

elprans (approved):

PR #6901, opened 2024-02-21
Add endianness to int converter functions.

vpetrovykh:
Add `std::Endian` enum with two values `big` and `little`.

Update the bytes <-> int converter functions with a required `Endian` parameter.

Update the docs to reflect these changes.

### DIFF
--- a/edb/lib/std/70-converters.edgeql
+++ b/edb/lib/std/70-converters.edgeql
@@ -270,38 +270,77 @@ std::to_bytes(s: std::str) -> std::bytes {
 };
 
 
+CREATE SCALAR TYPE
+std::Endian EXTENDING enum<Little, Big>;
+
+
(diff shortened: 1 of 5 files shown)


edb/lib/std/70-converters.edgeql
```diff
@@ -270,38 +270,77 @@ std::to_bytes(s: std::str) -> std::bytes {
 };
 
 
+CREATE SCALAR TYPE
+std::Endian EXTENDING enum<little, big>;
```
msullivan, line 274: Should it be `Little` and `Big`?

We *almost* always capitalize enum values. (But not, it seems, *always* always.)

msullivan (approved):

1st1 (approved):

PR #6903, opened 2024-02-21
Add Next.js App Router guide

beerose:


docs/guides/tutorials/nextjs_pages_router.rst
```diff
@@ -1,10 +1,10 @@
-.. _ref_guide_nextjs:
+.. _ref_guide_nextjs_pages_router:
```
raddevon, line 1: I also like to update the label names, but it does require us to change any existing references to the old label. In this case, that would require a companion PR [in the web site repo](https://github.com/edgedb/edgedb.com/blob/master/content/blog/2023-08-04-keep-it-ryl-in-your-documentation/post.rst?plain=1#L94).

docs/guides/tutorials/nextjs_app_router.rst
```diff
@@ -0,0 +1,627 @@
...
+app with Next.js's ``create-next-app`` tool.
+
+.. code-block:: bash
+
+  $ npx create-next-app@latest
+
+You'll be prompted to provide a name (we'll use ``nextjs-blog``) for your 
+app and choose project options. For this tutorial, we'll go with the 
+recommended settings including TypeScript, App Router, and 
+**opt-ing out** of the ``src/`` directory.
```
raddevon, line 21: > …we'll go with the recommended settings…

Since we don't know for certain all these options will continue to be the recommended settings in perpetuity, I think we should delineate each of the settings we want the user to select. We've picked out a few of them here and called them out, but I think we're leaving the rest for the user to choose the "recommended setting" which could change in the future.

Here are the prompts I answered with the ones we haven't called out in bold:

> ✔ Would you like to use TypeScript? … No / Yes
> ✔ **Would you like to use ESLint? … No / Yes**
> ✔ **Would you like to use Tailwind CSS? … No / Yes**
> ✔ Would you like to use `src/` directory? … No / Yes
> ✔ Would you like to use App Router? (recommended) … No / Yes
> ✔ **Would you like to customize the default import alias (@/*)? … No / Yes**

docs/guides/tutorials/nextjs_app_router.rst
```diff
@@ -0,0 +1,627 @@
...
+dependencies. Once it's done, you can navigate to the app's directory and 
+start the development server.
+
+.. code-block:: bash
+
+  $ cd nextjs-blog
+  $ yarn dev # or pnpm dev or npm run dev or bun dev
+
+Open `localhost:3000 <http://localhost:3000>`_ to see the default Next.js
+homepage. At this point the app's file structure looks like this:
```
raddevon, line 33: If you already have something running on 3000, the dev server will listen on a different port. Does it matter, or should we assume that, if you're running something on 3000, you'll probably be able to correct for that?

Maybe we just say something like this to avoid the issue:
> When the dev server starts, it will log out a local URL. Visit that URL to see the default Next.js homepage.

docs/guides/tutorials/nextjs_app_router.rst
```diff
@@ -0,0 +1,627 @@
...
+
+Updating the homepage
+---------------------
+
+Let's start by implementing a simple homepage for our blog application using
+static data. Replace the contents of ``app/page.tsx`` with the following.
+
+.. code-block:: tsx
+
+  // app/page.tsx
```
raddevon, line 70:
```suggestion
.. code-block:: tsx
  :caption: app/page.tsx

```

docs/guides/tutorials/nextjs_app_router.rst
```diff
@@ -0,0 +1,627 @@
...
+
+.. code-block:: bash
+
+  $ edgedb project init
+  No `edgedb.toml` found in `~/nextjs-blog` or above
+  Do you want to initialize a new project? [Y/n]
+  > Y
+  Specify the name of EdgeDB instance to use with this project [default:
+  nextjs-blog]:
+  > nextjs-blog
```
raddevon, line 152: Was this the default for you? In my experience, the CLI always replaces hyphens with underscores when suggesting an instance name. On my system, it recommended `nextjs_blog` instead of `nextjs-blog`.

It seems to work if I change it to a hyphen, so I'm not really sure why the CLI does that replacement. Might still be good to reflect the real-world output in what we show here though.

docs/guides/tutorials/nextjs_app_router.rst
```diff
@@ -0,0 +1,627 @@
...
+  Checking EdgeDB versions...
+  Specify the version of EdgeDB to use with this project [default: 2.x]:
+  >
+  ┌─────────────────────┬──────────────────────────────────────────────┐
+  │ Project directory   │ ~/nextjs-blog                                │
+  │ Project config      │ ~/nextjs-blog/edgedb.toml                    │
+  │ Schema dir (empty)  │ ~/nextjs-blog/dbschema                       │
+  │ Installation method │ portable package                             │
+  │ Start configuration │ manual                                       │
+  │ Version             │ 2.x                                          │
```
raddevon, line 162: Maybe show the current version (`4.5+641a8f3`) or `x.x` to make it feel more evergreen?
```suggestion
  │ Version             │ x.x                                          │
```

raddevon: Oh, same feedback up on line 154

docs/guides/tutorials/nextjs_app_router.rst
```diff
@@ -0,0 +1,627 @@
...
+  │ Instance name       │ nextjs-blog                                  │
+  └─────────────────────┴──────────────────────────────────────────────┘
+  Initializing EdgeDB instance...
+  Applying migrations...
+  Everything is up to date. Revision initial.
+  Project initialized.
+
+This process has spun up an EdgeDB instance called ``nextjs-blog`` and
+"linked" it with your current directory. As long as you're inside that
```
raddevon, line 171: Since the CLI has a concept of "linking" which basically sets a handle you can use to run CLI commands against a remote instance, maybe it would be better to use a different word in this context. I'd recommend "associated." I think we can drop the quotes here too.
```suggestion
associated it with your current directory. As long as you're inside that
```

docs/guides/tutorials/nextjs_app_router.rst
```diff
@@ -0,0 +1,627 @@
...
+Next.js app.
+To do that, we'll need the ``edgedb`` client library. Let's install that from
+NPM:
+
+.. code-block:: bash
+
+  $ npm install edgedb 
+  # or yarn add edgedb or pnpm add edgedb or bun add edgedb
+
+Then go to the ``app/page.tsx`` file to replace the static data with a
```
raddevon, line 282:
```suggestion
Then go to the ``app/page.tsx`` file to replace the static data with
```

docs/guides/tutorials/nextjs_app_router.rst
```diff
@@ -0,0 +1,627 @@
...
+builder. This provides a schema-aware client API that makes writing strongly
+typed EdgeQL queries easy and painless. The result type of our queries will be
+automatically inferred, so we won't need to manually type something like
+``type Post = { id: string; ... }``.
+
+First, install the generator to your project.
+
+.. code-block:: bash
+
+  $ yarn add --dev @edgedb/generate
```
raddevon, line 360: We should probably stick with one package manager/script runner as the recommended choice. We start by recommending `yarn dev` to start the dev server before switching to `npm install edgedb` and then back to `yarn add` here. I would lean toward `npm` as the default recommendation since users who meet the Node prerequisite will already have that installed as well, but I'm fine with `yarn` if there's a compelling case for it in the context of this tutorial.

docs/guides/tutorials/nextjs_app_router.rst
```diff
@@ -0,0 +1,627 @@
...
+                </Link>
+              </li>
+            ))}
+          </ul>
+        </div>
+      )
+    }
+
+Instead of writing our query as a plain string, we're now using the query
+builder to declare our query in a code-first way. As you can see we import the
```
raddevon, line 447:
```suggestion
builder to declare our query in a code-first way. As you can see, we import the
```

docs/guides/tutorials/nextjs_app_router.rst
```diff
@@ -0,0 +1,627 @@
...
+query builder as a single default import ``e`` from the ``dbschema/edgeql-js``
+directory.
+
+Now, when we update our ``selectPosts`` query, the type of our dynamically
+loaded ``posts`` variable will update automatically—no need to keep
```
raddevon, line 452:
```suggestion
loaded ``posts`` variable will update automatically — no need to keep
```

docs/guides/tutorials/nextjs_app_router.rst
```diff
@@ -0,0 +1,627 @@
...
+``BlogPost`` type by its ``id``. We're also using the ``uuid`` function 
+from the query builder to convert the ``id`` parameter to a UUID.
+
+Now, click on one of the blog post links on the homepage. This should bring
+you to ``/post/<uuid>``.
+
+Deploying to Vercel
+-------------------
+
+**#1 Deploy EdgeDB**
```
raddevon, line 518: I know this was just copied over directly from the other guide, but should we take this opportunity to set up EdgeDB Cloud as the default way to deploy EdgeDB? Then, we can just link to the deployment guides as an alternative for people who want to self-host.

docs/guides/tutorials/nextjs_app_router.rst
```diff
@@ -0,0 +1,627 @@
...
+We've also created a dynamic route that fetches a single post from the 
+database and displays it on a separate page.
+
+The next step is to add a ``/newpost`` page with a form for writing new blog
+posts and saving them into EdgeDB. That's left as an exercise for the reader.
+
+To see the final code for this tutorial, refer to 
+`github.com/edgedb/edgedb-examples/tree/main/nextjs-blog
+<https://github.com/edgedb/edgedb-examples/tree/main/
+nextjs-blog-app-router>`_.
```
raddevon, line 627:
```suggestion
`github.com/edgedb/edgedb-examples/tree/main/nextjs-blog-app-router
<https://github.com/edgedb/edgedb-examples/tree/main/
nextjs-blog-app-router>`_.
```

raddevon (changes requested):
This looks great! I added some comments and suggestions.

raddevon:
Oh, forgot to mention, you might need to update submodules and push. That's one of the reasons the build is failing. The other is a line longer than 79 characters, I think (a check which will be going away soon 🎉 ).

docs/guides/tutorials/nextjs_pages_router.rst
beerose: https://github.com/edgedb/edgedb.com/pull/797

beerose: This was the only reference I found. Let me know it I missed something

docs/guides/tutorials/nextjs_app_router.rst
beerose: I'll add a code snippet, highlighting the options we're picking for this particular tutorial.

docs/guides/tutorials/nextjs_app_router.rst
beerose: Makes sense.

docs/guides/tutorials/nextjs_app_router.rst
beerose: Ohhh, nice!

docs/guides/tutorials/nextjs_app_router.rst
beerose: I thought about it for a moment, but then I wasn't sure since we don't have a free tier yet. We can mention it, but since it's only a tutorial, EdgeDB could be more costly compared to other providers (+ some have free tiers).

docs/guides/tutorials/nextjs_app_router.rst
```diff
@@ -0,0 +1,627 @@
...
+            </li>
+          ))}
+        </ul>
+      </div>
+    )
+  }
+
+
+After saving, Next.js should hot-reload, and you should see the updated
+homepage in your browser.
```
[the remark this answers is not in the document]
beerose: Hmm, actually, it is required as it's using RSC. My bad here.

docs/guides/tutorials/nextjs_app_router.rst
beerose: Hmm, it actually only matters for the install commands. I used a different style here as the command was longer than the others.

raddevon: Maybe if they have both Yarn and npm, it doesn't matter which they use to install a dependency, but it _will_ matter if they *don't* have Yarn installed. Should we assume they do? If so, should we maybe list pre-requisites at the top of the guide, including Yarn among them? Should we take them through the steps to install it or maybe link them to the Yarn installation instructions?

I suggest we avoid all those questions by showing everything with npm by default which we _know_ they have if they have Node.

beerose: > I suggest we avoid all those questions by showing everything with npm by default which we know they have if they have Node.

I think this is a good fallback solution. Will go with this one.

docs/guides/tutorials/nextjs_app_router.rst
beerose: I've added the EdgeDB Cloud deployment. There's a bit of repetition, but I figured it might be easier to follow this way (there are "with EdgeDB Cloud" and "with other providers" sections.

docs/guides/tutorials/nextjs_app_router.rst
```diff
@@ -528,9 +528,86 @@ you to ``/post/<uuid>``.
...
+
+  $ edgedb project unlink
+
+Then run the following command to link the project to the EdgeDB Cloud:
+
+.. code-block:: bash
+
+  $ edgedb project init --server-instance "<org>/<instance-name>"
+
+The migrations and schema will be automatically uploaded to the cloud.
```
raddevon: I'm a little unsure about this as the recommended workflow. I would think the user might want to be able to continue to work locally, so I would probably recommend migrating via `edgedb migrate -I <org>/<instance-name>` and keeping the project link to the local instance.

Maybe we should get some input from @elprans and @1st1 on this to see what workflow we should recommend here…

docs/guides/tutorials/nextjs_app_router.rst
```diff
@@ -541,6 +541,16 @@ and create a new instance. Create and make note of a secret key for your EdgeDB
...
 
+.. note::
+
+  If you want to restore your data from a local instance to the cloud, you 
+  can use the ``edgedb dump`` and ``edgedb restore`` commands.
+
+.. code-block:: bash
+
+  $ edgedb dump <your-dump.dump>
+  $ edgedb restore -I <org>/<instance-name> <your-dump.dump>
```
raddevon, line 553: For whatever we end up recommending below this, we may want to mention here that migrating is no longer necessary if you choose to dump/restore.

docs/guides/tutorials/nextjs_app_router.rst
beerose: Ahh, I see. I thought that this is what you meant by "having to start from scratch".

raddevon (approved):

PR #6910, opened 2024-02-22
Fix CI by trusting pgvector extension

fantix:
Fixes the [failing CI](https://github.com/edgedb/edgedb/actions/runs/7998861394/job/21845763716#step:23:7047)

### DIFF
--- a/tests/test_edgeql_data_migration.py
+++ b/tests/test_edgeql_data_migration.py
@@ -12098,8 +12098,6 @@ class TestEdgeQLDataMigrationNonisolated(EdgeQLDataMigrationTestCase):
     async def test_edgeql_migration_extension_01(self):
         # Test migrations getting from an array in integers to vector and then
         # revert to an array of floats.
-        if not self.is_superuser:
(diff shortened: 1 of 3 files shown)


tests/test_edgeql_data_migration.py
```diff
@@ -3344,6 +3344,9 @@ async def test_edgeql_migration_reject_prop_05(self):
         await self.fast_forward_describe_migration()
 
     async def test_edgeql_migration_vector_change_01(self):
+        if not self.is_superuser:
```
elprans, line 3347: Superuser is not required to create extensions, so we need to adjust our `NOCREATEROLE` mode to include that permission bit instead.  Otherwise we'll be skipping this test on hosted Postgres tests and I'd like to avoid that.

fantix: Fixed with [sample run](https://github.com/edgedb/edgedb/actions/runs/8008439238)

elprans (approved):

PR #6911, opened 2024-02-22
Refactor alias and global delta operations

aljazerzen:
Followup for #6866

Findings:
- linearize_delta will:
  - silently drop all declarations that have a namespace (`default::__FooAlias__x@w1`)
  - break when delta contains an un-setting alter property i.e. (`alter_alias.set_attribute_value('type', None)`),

I have a hard time differentiating between the initial and the canonicalized version of the delta tree. Could we completely separate these two passes? So we have one full pass of the tree for canonicalization and then a separate apply pass.


msullivan (approved):
This is great! I have a couple nits that I am going to try to just fix myself for timezone/release reasons, so I'm holding off on merging for a bit

PR #6912, opened 2024-02-22
Add styled default email templates

scotttrinh:
Used [MJML](https://mjml.io/) to produce some reasonable email templates based on some designs provided by Roman. I considered basically translating MJML into "components" but there are some tricky ways they wrap certain kinds of elements. Probably still doable, but I ran out of patience.

Here is the general shape (showing the magic link email)

image

### DIFF
--- a/edb/server/protocol/auth_ext/ui/components.py
+++ b/edb/server/protocol/auth_ext/ui/components.py
@@ -169,7 +169,7 @@ def _oauth_button(
 
 
 def button(
-    text: str,
+    text: Optional[str],
     *,
     id: Optional[str] = None,
     secondary: Optional[bool] = False,
@@ -189,7 +189,7 @@ def button(
 
     return f'''
       <button {attrs}>
-        {f'<span>{text}</span>' if text is not None else ''}
+        {f'<span>{text}</span>' if text else ''}
         <svg
           xmlns="http://www.w3.org/2000/svg"
           width="24"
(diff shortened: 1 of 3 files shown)


1st1 (approved):
Looks good conceptually, but I wasn't looking at the code closely.

edb/server/protocol/auth_ext/ui/components.py
```diff
@@ -336,3 +336,269 @@ def success_message(message: str) -> str:
...
+                                      border: none;
+                                      display: block;
+                                      outline: none;
+                                      text-decoration: none;
+                                      height: 150px;
+                                      width: 100%;
+                                      font-size: 13px;
+                                    "
+                                    width="150"
+                                  />
```
scotttrinh, line 430: I went with 150x150 here for lack of a better idea. Another option is to have the width set to 150 and have the height be `auto` which would preserve the aspect ratio. We don't give much guidance to the developer about aspect ratio and size, and I think we should.

emrosenf:
Would you consider exposing an option for the user to retrieve the link and send their own transactional e-mails?

scotttrinh:
@emrosenf 
> Would you consider exposing an option for the user to retrieve the link and send their own transactional e-mails so that they can be styled by the user?

We have a near term plan to allow setting the email templates (keep an eye on #6868) that would allow totally custom emails via your configured SMTP server, but we're likely going to still maintain the flows within the auth server/extension rather than exposing an API that you could piece something together with to send your own emails.

In some senses, all of these flows are something that you could replace with your own bespoke flows since all of the data lives in your database. If you store the signing key securely on your server you could issue your own auth tokens, so you could implement your own password recovery and email verification flows directly. A bit of a heavy lift, but totally doable since the server code we're including here isn't privileged in any way that your own code isn't.

Issue #6794, opened 2024-02-07
Conditional DML in `if...else` has various bugs and `Internal Server Errors`


corbinday:
As I've been trying to write some conditional DML, my approaches have failed for one reason or another:
- `Internal Server Error: cannot reference correlated set [ YOUR SET ] here`
- `Internal Server Error: maximum recursion depth exceeded while calling a Python object`
- nothing is inserted or updated

Here's a simplified schema and some query examples to showcase.

### `Internal Server Error: cannot reference correlated set [ YOUR SET ] here`
```edgeql
with
  user := (select User limit 1),
  question := (select B_Question filter .color = Color.Green limit 1),
  color := question.color,
  previousReply := (select Reply filter .user = user and .question = question)
select (
  if question is B_Question and color = Color.Red and exists previousReply then (
    update previousReply[is H_Reply] set { value := 'value'}
  ) else if question is B_Question and color = Color.Red and not exists previousReply then (
    insert H_Reply {
      user := user,
      question := question,
      value := 'value'
    }
  ) else if question is B_Question and color = Color.Green and exists previousReply then (
    update previousReply[is W_Reply] set {
      value := 'new value'
    }
  ) else (
    insert W_Reply {
      user := user,
      question := question,
      value := 'value'
    }
  )
)
```

### `Internal Server Error: maximum recursion depth exceeded while calling a Python object`
I'm guessing that hitting this error is inevitable at some level of `n` `if...else`'s due to the underlying implementation that makes this work with postgres...
```edgeql
with
  user := (select User limit 1),
  question := (select B_Question filter .color = Color.Green limit 1),
  color := question.color,
  previousReply := (select Reply filter .user = user and .question = question),
select (
  if false then (
    (update previousReply[is H_Reply] set { value := 'value 1'})
  ) else if false then (
    (update previousReply[is H_Reply] set { value := 'value 2'})
  ) else if false then (
    (update previousReply[is H_Reply] set { value := 'value 3'})
  ) else if false then (
    (update previousReply[is H_Reply] set { value := 'value 4'})
  ) else if false then (
    (update previousReply[is H_Reply] set { value := 'value 5'})
  ) else if false then (
    (update previousReply[is H_Reply] set { value := 'value 6'})
  ) else if false then (
    (update previousReply[is H_Reply] set { value := 'value 7'})
  ) else if false then (
    (update previousReply[is H_Reply] set { value := 'value 8'})
  ) else if false then (
[...]

PR #6917, opened 2024-02-23
Fix DML coalesce inside of IF/ELSE

msullivan:
The core thing here is that when we move a compiled IR expression into
a FOR loop, we need to adjust the scope tree so that it is inside the
loop's scope.

Also fix DML appearing directly as a FOR iterator, which I ran into
while testing a minimized bug (that turned out to not be the full
issue). I was nervous my change there might mess up the GROUP
optimizations, so I added another codegen test.

Fixes main part of #6794.

### DIFF
--- a/edb/edgeql/compiler/expr.py
+++ b/edb/edgeql/compiler/expr.py
@@ -312,6 +312,36 @@ def compile_Array(
     return setgen.new_array_set(elements, ctx=ctx, srcctx=expr.context)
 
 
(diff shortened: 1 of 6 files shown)


aljazerzen:
Lets talk about this on the call tomorrow.

msullivan:
@aljazerzen Yes

@elprans Could I get a review now? This is a pretty bad bug

elprans (approved):

PR #6918, opened 2024-02-23
Add `bytes` option to `array_join`.

vpetrovykh:
The `array_join` function can now also be used to join bytes into a single byte-string.


msullivan (approved):
Double check your spelling in the commit message, but otherwise looks great

PR #6923, opened 2024-02-26
Update edgedb-python dep

aljazerzen:


elprans (approved):

pyproject.toml
```diff
@@ -4,7 +4,7 @@ description = "EdgeDB Server"
 requires-python = '>=3.11.0'
 dynamic = ["version"]
 dependencies = [
-    'edgedb==1.6.0',
+    'edgedb==1.9.0',
```
elprans, line 7: There's a similar requirement in `build-system.requires`, you need to bump that also.

PR #6924, opened 2024-02-26
Add Config.enable_recompilation

fantix:
It's turned off by default in `ClusterTestCase`s

### DIFF
--- a/edb/lib/cfg.edgeql
+++ b/edb/lib/cfg.edgeql
@@ -178,7 +178,7 @@ ALTER TYPE cfg::AbstractConfig {
             Access-Control-Allow-Origin HTTP header';
     };
 
-    CREATE PROPERTY enable_recompilation -> std::bool {
+    CREATE PROPERTY auto_rebuild_query_cache -> std::bool {
         SET default := true;
         CREATE ANNOTATION std::description :=
             'Recompile all cached queries on DDL if enabled.';
--- a/edb/lib/cfg.edgeql
+++ b/edb/lib/cfg.edgeql
@@ -178,6 +178,13 @@ ALTER TYPE cfg::AbstractConfig {
             Access-Control-Allow-Origin HTTP header';
     };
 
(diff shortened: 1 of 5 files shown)


edb/lib/cfg.edgeql
```diff
@@ -178,6 +178,13 @@ ALTER TYPE cfg::AbstractConfig {
             Access-Control-Allow-Origin HTTP header';
     };
 
+    CREATE PROPERTY enable_recompilation -> std::bool {
+        SET default := true;
+        CREATE ANNOTATION cfg::system := 'true';
+        CREATE ANNOTATION std::description :=
+            'Recompile all cached queries on DDL if enabled.';
+    };
```
msullivan, line 186: This shouldn't need to be system, right? It could be compiled at any level and should be fine, and then we could disable/enable on specific test suites more easily

edb/lib/cfg.edgeql
```diff
@@ -178,6 +178,12 @@ ALTER TYPE cfg::AbstractConfig {
             Access-Control-Allow-Origin HTTP header';
     };
 
+    CREATE PROPERTY enable_recompilation -> std::bool {
```
elprans, line 181: The name seems too unspecific.  I'd use something like `auto_rebuild_query_cache`

edb/server/protocol/execute.pyx
```diff
@@ -515,19 +521,27 @@ async def execute_script(
...
-                    ]
+        if (
+            not in_tx
+            and dbv.server.config_lookup(
+                "enable_recompilation",
+                dbv.get_session_config(),
+                dbv.get_database_config(),
+                dbv.get_system_config(),
+            )
+            and any(query_unit.user_schema for query_unit in unit_group)
```
msullivan, line 532: Let's check whether there is DDL before we do the config lookup, as a microopt

msullivan (approved):

PR #6926, opened 2024-02-27
Don't leak objects out of access policies when used in a computed global

msullivan:
The caching for our computed globals failed to account for whether
access policies are in effect, which could allow leaking of objects
that should not have been visible. Fix this by augmenting the cache
key.

A downside of this fix is that sometimes now we will generate two
identical computations of a computed global. This is probably fine for
now, since this is an important bug to fix, and 2 is still a lot
better than the N that it was without the cache.

### DIFF
--- a/edb/edgeql/compiler/context.py
+++ b/edb/edgeql/compiler/context.py
@@ -151,7 +151,10 @@ class Environment:
     path_scope: irast.ScopeTreeNode
     """Overrall expression path scope tree."""
 
-    schema_view_cache: Dict[s_types.Type, tuple[s_types.Type, irast.Set]]
+    schema_view_cache: Dict[
+        tuple[s_types.Type, object],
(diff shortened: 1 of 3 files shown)


edb/edgeql/compiler/context.py
```diff
@@ -151,7 +151,10 @@ class Environment:
     path_scope: irast.ScopeTreeNode
     """Overrall expression path scope tree."""
 
-    schema_view_cache: Dict[s_types.Type, tuple[s_types.Type, irast.Set]]
+    schema_view_cache: Dict[
+        tuple[s_types.Type, object],
```
elprans, line 155: Why `object` and not `Any`?

elprans (approved):

msullivan:
Because it is intentionally opaque

PR #6928, opened 2024-02-27
Updated versions of python and rust in contribution guide.

dnwpark:
The versions of python and rust required for the contribution guide are out of date and should be updated.

The following guides are not updated but they also explicitly reference python 3.10. They will likely need to be updated in the future:

- `docs/guides/tutorials/graphql_apis_with_strawberry.rst`
- `docs/guides/tutorials/rest_apis_with_fastapi.rst`
- `docs/guides/tutorials/rest_apis_with_flask.rst`

### DIFF
--- a/docs/guides/contributing/code.rst
+++ b/docs/guides/contributing/code.rst
@@ -26,9 +26,9 @@ Linux or macOS.  Windows is not currently supported.
 
 * GNU make version 3.80 or newer;
 * C compiler (GCC or clang);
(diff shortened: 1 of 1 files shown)


msullivan (approved):
Having tutorials reference 3.10 is fine, from the perspective of compatability. It's still pretty new, I'm sure the tutorials still work. edgedb-python supports 3.8 and later

PR #6931, opened 2024-02-27
Change internal group keyword INTO to IN.

dnwpark:
`INTO` is not a reserved keyword and causes issues with trying to modify how the previous `BY` clause is defined. Since this is an internal statement, changing to a reserved keyword should be ok.


elprans (commented):
> Changed internal group

Prefer imperative mood in commit messages, e.g "Change internal group"

msullivan (approved):

PR #6935, opened 2024-02-28
Support concurrent testing of emails

scotttrinh:
Reusing the same temporary file for all emails only works for serialized (or at least non-concurrent) execution contexts. If you ran the test suite in parallel, you'd get collisions and test failures.

This change makes each run of each test write to a deterministic but randomized location based on a hash of the sender and recipients, which is known to each test case, but is randomly generated (UUIDv4) for each test run.


fantix (approved):
Thanks!

PR #6936, opened 2024-02-28
Document wrapping `insert` with `select`

raddevon:
I thought we already had this documented — and I believe it *is* in the book — but it's a pretty fundamental concept and should also be documented here.

### DIFF
--- a/docs/edgeql/insert.rst
+++ b/docs/edgeql/insert.rst
@@ -117,6 +117,25 @@ You can only ``insert`` instances of concrete (non-abstract) object types.
   ... };
   error: QueryError: cannot insert into abstract object type 'default::Person'
 
+By default, ``insert`` returns only the inserted object's ``id`` as seen in the
+examples above. If you want to get additional data back, you may wrap your
+``insert`` with a ``select`` and apply a shape specifying any properties and
+links you want returned:
+
+.. code-block:: edgeql-repl
+
+  db> select (insert Hero {
+  ...   name := "Spider-Man"
+  ...   # secret_identity is omitted
+  ... }) {id, name};
+  {
+    default::Hero {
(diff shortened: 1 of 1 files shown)


scotttrinh (approved):
We added some details about this to the edgedb-js docs (see https://github.com/edgedb/edgedb-js/pull/838). To that end, makes sense to add this to the `update` docs as well, yeah?

Maybe we also need to update `select` in a way that makes this pattern intuitive?

docs/edgeql/insert.rst
```diff
@@ -117,6 +117,25 @@ You can only ``insert`` instances of concrete (non-abstract) object types.
...
   error: QueryError: cannot insert into abstract object type 'default::Person'
 
+By default, ``insert`` returns only the inserted object's ``id`` as seen in the
+examples above. If you want to get additional data back, you may wrap your
+``insert`` with a ``select`` and apply a shape specifying any properties and
+links you want returned:
+
+.. code-block:: edgeql-repl
+
+  db> select (insert Hero {
```
elprans, line 127: Let's also offer an example of a `with` style, which I personally prefer as a better-looking formulation:

```
with
   NewHero := (insert)
select
   NewHero {
      id,
      name,
   }
```

raddevon:
@scotttrinh Good idea. Maybe we just cover `delete` too while we're in there and complete the set…

raddevon:
@scotttrinh Added this to the `update` docs as well. It looks a little different because that page is surprisingly different in structure from the `select` one, but I tried to match the structure of the `update` page while keeping the new information in basically the same shape it has on the `select` page. (Turned out, `delete` already mentions this.)

Would you take a peek at the new stuff and make sure it looks reasonable before I merge?

scotttrinh (approved):
New changes look good!

PR #6937, opened 2024-02-29
Add 5.0 dumps.

vpetrovykh:


msullivan:
@vpetrovykh could you revisit this and generate 5.2 dumps with the current tests?

PR #6938, opened 2024-02-29
Start refactoring irast.Set

msullivan:
Here is a series of refactors for irast.Set, with the general eventual
goal of making irast.Set *just* a metadata holder, and not also
responsible for a bunch of other operations directly.

Probably review the commits individually.

An eventual goal is to also eliminate the `rptr` field by making
`Pointer` an `Expr` (and putting computed expressions inside `Pointer`
in the cases where currently both `expr` and `rptr` are set), and I am
making good progress on that, but that one is a bigger project.

### DIFF
--- a/edb/edgeql/compiler/inference/cardinality.py
+++ b/edb/edgeql/compiler/inference/cardinality.py
@@ -332,6 +332,19 @@ def __infer_type_introspection(
     return ONE
 
 
+@_infer_cardinality.register
+def __infer_type_root(
+    ir: irast.TypeRoot,
+    *,
+    scope_tree: irast.ScopeTreeNode,
+    ctx: inference_context.InfCtx,
+) -> qltypes.Cardinality:
+    if typeutils.is_exactly_free_object(ir.typeref):
+        return ONE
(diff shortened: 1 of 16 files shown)


msullivan:
Looks like there is still some DDL-related stuff to fix

aljazerzen (approved):

edb/edgeql/compiler/inference/cardinality.py
```diff
@@ -332,6 +332,19 @@ def __infer_type_introspection(
     return ONE
 
 
+@_infer_cardinality.register
+def __infer_type_root(
+    ir: irast.TypeRoot,
```
elprans, line 337: I know we have `root_rvar`, but do we think there is a better name than `root`?

msullivan: I'm not sure. I picked it by analogy with `root_rvar` but also don't love it.

msullivan:
I'm going to merge this, but if anybody has a better name for TypeRoot, let's change it

PR #6943, opened 2024-02-29
Remove Nonterm metaclasses

dnwpark:
Removed the following `Nonterm` related metaclasses:

- `NontermMeta`
- `ListNontermMeta`
- `KeywordMeta`

Replaced functionality with `__init_subclass__`.

### DIFF
--- a/edb/common/parsing.py
+++ b/edb/common/parsing.py
@@ -18,7 +18,8 @@
 
 
 from __future__ import annotations
-from typing import Any, Callable, Tuple, Type, Dict, List, Set, cast
+from typing import Any, Callable, Tuple, Type, Dict, List, Set, \
+                   Optional, cast
 
 import json
 import logging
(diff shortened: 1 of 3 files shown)


edb/common/parsing.py
```diff
@@ -18,7 +18,8 @@
 
 
 from __future__ import annotations
-from typing import Any, Callable, Tuple, Type, Dict, List, Set, cast
+from typing import Any, Callable, Tuple, Type, Dict, List, Set, \
+                   Optional, cast
```
msullivan, line 22: We typically prefer doing continuations with parens, when it is possible, so something like
```
from typing import (
    Any, Callable, Tuple, Type, Dict, List, Set,
    Optional, cast
)
```
would be preferred

edb/common/parsing.py
```diff
@@ -109,17 +111,19 @@ def decorator(func: Any):
...
-            return result
+class Nonterm(parsing.Nonterm):
 
-        if not result.__doc__:
-            result.__doc__ = '%nonterm'
+    def __init_subclass__(cls, is_internal=False, **kwargs):
+        super().__init_subclass__(**kwargs)
 
-        for name, attr in result.__dict__.items():
+        if is_internal:
```
msullivan, line 120: Let's add a docstring explaining that we read method names and use them to generate `%reduce` docstrings.

(Might as well clean things up a bit while you are here)

edb/common/parsing.py
```diff
@@ -142,49 +146,47 @@ def __new__(mcls, name, bases, dct):
...
-            elif isinstance(element, NontermMeta):
-                element = element.__name__
-
-            tokens.append(element)
-
-            if separator:
-                prod = lambda self, lst, sep, el: self._reduce_list(lst, el)
+            if separator_name:
+                tail_prod = \
+                    lambda self, lst, sep, el: self._reduce_list(lst, el)
```
msullivan, line 162: Use parens for the continuation

edb/common/parsing.py
```diff
@@ -142,49 +146,47 @@ def __new__(mcls, name, bases, dct):
...
-    def __init__(cls, name, bases, dct, *, element, separator=None):
-        super().__init__(name, bases, dct)
-
-
-class ListNonterm(Nonterm, metaclass=ListNontermMeta, element=None):
+                tail_prod = \
+                    lambda self, lst, el: self._reduce_list(lst, el)
+                tail_prod_name = 'reduce_{}_{}'.format(
+                    cls.__name__, element_name)
+                setattr(cls, tail_prod_name, tail_prod)
```
msullivan, line 171: This is shared between both branches, so maybe do it once after the branches

msullivan (approved):
Looks good! A few nits.

Very happy to be burning down these metaclasses.

edb/common/parsing.py
```diff
@@ -142,49 +146,47 @@ def __new__(mcls, name, bases, dct):
...
-
-class ListNontermMeta(NontermMeta):
-    def __new__(mcls, name, bases, dct, *, element, separator=None):
-        if name != 'ListNonterm':
-            if issubclass(separator, Token):
-                separator = separator._token
-            elif isinstance(separator, NontermMeta):
-                separator = separator.__name__
+class ListNonterm(Nonterm, is_internal=True):
+    def __init_subclass__(cls, element, separator=None, is_internal=False,
```
msullivan, line 153: I would keep the `*, ` before `element`, which requires the arguments after it to be keyword arguments.

(That said, it doesn't really matter here, since there's not any way to pass non-keyword arguments to class creation)

PR #6945, opened 2024-02-29
Handle WebAuthn challenge having multiple factors

scotttrinh:
We made some big changes late in the game for allowing multiple WebAuthn factors for a given (email, user_handle) pair, but didn't update this in the authentication challenge flow.

Since we cannot remove the old link, this adds an additional link and ignores the old one.

### DIFF
--- a/edb/server/protocol/auth_ext/webauthn.py
+++ b/edb/server/protocol/auth_ext/webauthn.py
@@ -390,7 +390,7 @@ select ext::auth::WebAuthnAuthenticationChallenge {
     created_at,
     modified_at,
     challenge,
-    factors: {
+    factors := (distinct (.factors union .factor)) {
         id,
         created_at,
         modified_at,
--- a/edb/server/protocol/auth_ext/webauthn.py
+++ b/edb/server/protocol/auth_ext/webauthn.py
@@ -390,7 +390,7 @@ select ext::auth::WebAuthnAuthenticationChallenge {
     created_at,
     modified_at,
     challenge,
-    factors := (distinct (.factors union .factor)) {
+    factors: {
         id,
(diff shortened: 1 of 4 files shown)


msullivan:
I would kind of prefer if we *also* looked at the old link, but I guess it's not totally critical for something like this

jaclarke (approved):

scotttrinh:
@msullivan 

Oh yeah, I remember we talked about that. Lemme see if it's easy to distinct union them together in the query so that everything else just keeps working.

scotttrinh:
@msullivan ⚡ [`2af19e9` (#6945)](https://github.com/edgedb/edgedb/pull/6945/commits/2af19e9ec68c8e01ada3f9fbe760f533c02e5cf2)

Let me know if that makes sense!

scotttrinh:
@msullivan ⚡ [`8fe06a2` (#6945)](https://github.com/edgedb/edgedb/pull/6945/commits/8fe06a2d61660f25e7ea7c6881666dfcf54d7a54)

As discussed, just removing the property.

edb/server/protocol/auth_ext/webauthn.py
```diff
@@ -315,18 +315,19 @@ async def create_authentication_options_for_email(
...
             select ext::auth::WebAuthnFactor
             filter .user_handle = user_handle
             and .email = email
-        )))
+        ))
     )
 insert ext::auth::WebAuthnAuthenticationChallenge {
     challenge := challenge,
-    factor := factor,
+    factor := assert_exists(assert_single((select factors limit 1))),
```
jaclarke, line 327:
```suggestion
```

PR #6946, opened 2024-02-29
Fix ISE when creating an alias with a name that already exists

aljazerzen:


aljazerzen:
@msullivan, I found something strange: migrating from an actual type to an alias:

```
# from test test_schema_migrations_equivalence_23
start migration to { module default {
            type Child {
                property foo -> str;
            }

            type Base {
                link bar -> Child;
            }
}}; populate migration; commit migration;
start migration to { module default {
            type Child {
                property foo -> str;
            }

            # exchange a type for an alias
            alias Base := (
                SELECT Child {
                    # bar is the same as the root object
                    bar := Child
                }
            );
}}; populate migration; commit migration;
```

The old alias implementation did not drop the type at all, it just found it already existing, so it didn't generate a new create type. This is the generated DDL:

```
    alter type default::Base extending default::Child LAST;
    create alias default::Base := (
        select
            default::Child {
                bar := default::Child
            }
    );
```

So what's happening is that we are just creating an alias "over" an existing type.

This feels strange:
- does this mean that also the backing pg table is not dropped?
- my new check that prevents creating an alias if that name already exists in the schema now throws an error
- and it is not wrong, a type with this name already exists.

Do we really need to support this?

aljazerzen:
I've tried disabling this and forcing a `DROP TYPE` + `CREATE ALIAS`. But this won't work if there are other types that depend on this type (which is exactly what test `test_schema_migrations_equivalence_23` is checking).

aljazerzen:
Ok, actually, this not that of a problem - conversion from an actual type to an alias does make sense (if underlying pg tables are really handled correctly).

msullivan:
The plan is to drop alias overwriting in this PR, right?

tests/test_schema.py
```diff
@@ -5404,14 +5418,45 @@ def test_schema_migrations_equivalence_60(self):
                 required property name -> str;
                 index pg::spgist on (.name);
             };
-        """, r"""
+        """,
+                r"""
```
msullivan, line 5422: Indent should line up.

ruff is not checking any whitespace errors, and it is a big problem.

msullivan (approved):

tests/test_schema.py
aljazerzen: And `r"""` is indented 'correctly', which is one indentation more than `[`, which is one indentation more than `self._assert_migration_equivalence`.

But yea, looks weird.

PR #6950, opened 2024-02-29
Some test suite perf tweaks

msullivan:
Because of the new implementation of branches, creating databases from
a template is serialized now, which slows down test suites that rely
on creating database clones. Fortunately most of them don't really
need it, so we can simplify matters by avoiding branches for test
suites with empty databases.

Improve the rendering of tests a bit too by including a count of the
currently running tests and rerendering when a test *starts*, instead
of just when one finishes.

### DIFF
--- a/edb/testbase/server.py
+++ b/edb/testbase/server.py
@@ -1092,7 +1092,8 @@ class DatabaseTestCase(ConnectedTestCase):
             # The retry here allows the test to survive a concurrent testing
             # EdgeDB server (e.g. async with tb.start_edgedb_server()) whose
             # introspection holds a lock on the base_db here
-            create_command = f'CREATE BRANCH {qlquote.quote_ident(dbname)}'
+            create_command = (
+                f'CREATE SCHEMA BRANCH {qlquote.quote_ident(dbname)}')
             if cls.get_setup_script():
                 create_command += f' FROM {qlquote.quote_ident(base_db_name)}'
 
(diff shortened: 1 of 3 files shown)


edb/testbase/server.py
```diff
@@ -1092,19 +1092,16 @@ async def setup_and_connect(cls):
...
-                ):
-                    async with tr:
-                        await admin_conn.execute(
-                            f'''
-                                CREATE DATABASE {qlquote.quote_ident(dbname)}
-                                FROM {qlquote.quote_ident(base_db_name)}
-                            ''',
-                        )
-            await create_db()
+            create_command = f'CREATE DATABASE {qlquote.quote_ident(dbname)}'
```
elprans, line 1095:
```suggestion
            create_command = f'CREATE BRANCH {qlquote.quote_ident(dbname)}'
```

msullivan: I went to try to fix this in the web UI because I was in the middle of something else locally and I wound up in a full vscode and terminal environment in my browser. Fun.

edb/testbase/server.py
```diff
@@ -1838,7 +1841,7 @@ def get_test_cases_setup(
         except unittest.SkipTest:
             continue
 
-        if not setup_script:
+        if setup_script is None:
```
elprans, line 1844: Giving `''` special treatment gives me heebie-jeebies.  Also is this really ever `None` currently? Maybe that's what we should be returning instead?

msullivan: Yeah, sure

msullivan: I'll drop the special handling of `''` but its best to keep returning that instead of `None`, since some subclasses want to concatenate stuff

elprans (approved):

PR #6954, opened 2024-03-01
Move extension tests to their own file

msullivan:
They are slow and large and I usually don't actually care to run them
when running ddl tests.


elprans (approved):

aljazerzen (approved):

msullivan:
I observed flakes here earlier, but they went away. I'm going to merge this now, then send up a clean PR to address that flake and another, since it involves some infrastructure

PR #6955, opened 2024-03-01
Fix CI test failure by cleaning up in test_edgeql_ddl_reindex

msullivan:
I'm not totally sure what happened to cause test_edgeql_ddl_reindex
to get sorted earlier than test_edgeql_ddl_extensions_01, though...


msullivan:
Oh, the `get_setup_script` change was causing `get_cases_by_shard` to lose track of the order, which I also fixed

elprans (approved):

PR #6956, opened 2024-03-01
Alter alias even from ref propagation

aljazerzen:
- uncomment the hack
- make commands produced by affected refs execute as causes instead of innards
- fix for reflection writer when we create a type and drop it in the same migration


aljazerzen:
The remaining error is this:
```
# from test_schema_migrations_computed_optionality_01
# schema 1:
            abstract type Removable {
                optional single property removed := false;
            };

            type Topic extending Removable { };
            type Definition { };

            alias VisibleTopic := (
                SELECT Topic {
                    defs := (SELECT Definition),
                }
                FILTER NOT .removed
            );

# schema 2:
            abstract type Removable {
                property removed := false;
            };

            type Topic extending Removable { };
            type Definition { };

            alias VisibleTopic := (
                SELECT Topic {
                    defs := (SELECT Definition),
                }
                FILTER NOT .removed
            );
```

We have an "alter property removed", which propagates via two paths:
- via Topic, because extends `Removable`, which propagates into `VisibleTopic`, which drops alias types and recreates,
- via reference to property `removed` in the `VisibleTopic`, which triggers alter alias again.

This fails because the property `VisibleTopic.removed` is found to be descendant of `Removable.removed` and is scheduled to be updated here:
https://github.com/edgedb/edgedb/blob/main/edb/schema/inheriting.py#L958
... but because one `alter alias` runs in the first iteration, the property does not exists anymore. Well it does exist, but with a different id.

edb/schema/inheriting.py
```diff
@@ -955,7 +958,14 @@ def _propagate_field_alter(
...
         props: Tuple[str, ...],
     ) -> None:
-        for descendant in scls.ordered_descendants(schema):
+        descendant_names = [
+            d.get_name(schema) for d in scls.ordered_descendants(schema)
+        ]
+
+        for descendant_name in descendant_names:
+            descendant = schema.get(descendant_name, type=so.InheritingObject)
+            assert descendant, '.inherit_fields caused a drop of a descendant?'
```
msullivan, line 967: `get` will raise an error by itself, right?
If you want it deferred to this assert, pass a `default`

edb/schema/expraliases.py
```diff
@@ -426,7 +421,6 @@ def _alter_begin(
 
         return super()._alter_begin(schema, context)
 
-
```
msullivan, line 429: We need to enable ruff's prefix whitespace linting!

msullivan (approved):

PR #6959, opened 2024-03-01
Split recompile and writeback in transaction

fantix:
This extracted the actual query cache recompilation to the time we compile the DDL itself, assuming the compiled DDL will soon execute with the recompiled cache updated in the same transaction.

We currently skip the recompilation for DDLs in a transaction and let the future COMMIT do it. The recompilation may unnecessarily block the DDL transaction (locking tables) for a long time because it's done before the COMMIT after DDLs. This shall be changed and improved after we add back the per-dbview cache for queries `_in_tx_with_ddl`, in a way that we recompile before applying the DDL in the database (locking tables) and hope there is no more DDLs in the same transaction.

### DIFF
--- a/edb/server/dbview/dbview.pyx
+++ b/edb/server/dbview/dbview.pyx
@@ -22,6 +22,7 @@ from typing import (
 
 import asyncio
 import base64
+import copy
 import json
 import os.path
 import pickle
@@ -85,12 +86,14 @@ cdef class CompiledQuery:
         extra_counts=(),
         extra_blobs=(),
         request=None,
+        recompiled_cache=None,
     ):
         self.query_unit_group = query_unit_group
         self.first_extra = first_extra
         self.extra_counts = extra_counts
         self.extra_blobs = extra_blobs
         self.request = request
+        self.recompiled_cache = recompiled_cache
 
 
 cdef class Database:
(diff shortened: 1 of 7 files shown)


edb/server/dbview/dbview.pyx
```diff
@@ -1124,6 +1080,42 @@ cdef class DatabaseConnectionView:
                 if not lock._waiters:
                     del lock_table[query_req]
 
+        recompiled_cache = None
```
msullivan, line 1083: Should this logic go in a separate function? (Fine if you don't think so)

msullivan (approved):

edb/server/protocol/execute.pyx
```diff
@@ -358,28 +344,16 @@ async def execute(
...
-            and server.config_lookup(
-                "auto_rebuild_query_cache",
-                dbv.get_session_config(),
-                dbv.get_database_config(),
-                dbv.get_system_config(),
-            )
+            not debug.flags.disable_persistent_cache
+            and compiled.recompiled_cache
+            or compiled.request
+            and query_unit.cache_sql
```
msullivan, line 354: Add extra parens to separate the `and` and `or` clauses. I can never remember their precedence so I can't understand it really

msullivan: (OK I'm pretty confident that `and` binds tighter because `and` is like multiplication and is sometimes in logic written as juxtaposition, but I still think we should avoid it.)

edb/server/dbview/dbview.pyx
fantix: Yeah, I was thinking about having all this decided in `server/compiler/compiler.py` or deeper where we can have per-statement "dbview" of session_config, database_config and all (it's like what `dbview.pyx` does in `start()`/`on_success()`, but it'll be too late for recompilation), so that we can handle scripts like:

```
configure session set auto_rebuild_query_cache := false;
create type Foo;
configure session set auto_rebuild_query_cache := true;
```

or use the correct config to recompile after scripts like:

```
configure current database set apply_access_policies := true;
create Type Foo;
```

So basically this logic is temporarily here and I'm expecting it to be replaced by a simple `if query_unit_group.should_recompile` or something like that in the near future.

edb/server/protocol/execute.pyx
```diff
@@ -231,7 +201,23 @@ async def execute(
 
             if query_unit.sql:
                 if query_unit.user_schema:
-                    ddl_ret = await be_conn.run_ddl(query_unit, state)
+                    if (
+                        not debug.flags.disable_persistent_cache
+                        and compiled.recompiled_cache
+                    ):
```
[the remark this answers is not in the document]
fantix: Added a comment. I'd defer the refactor after we have proper cache for scripts too, they might share the same code in a different way. `execute.pyx` does deserve a proper refactor:

https://github.com/edgedb/edgedb/blob/5812b60dbc6dafc2f93d7727ee2eda80661046d1/edb/server/protocol/execute.pyx#L49

edb/server/protocol/execute.pyx
fantix: Extracted 2 variables and added more comments :v:

PR #6960, opened 2024-03-01
Implement _run_and_rollback_retrying

msullivan:
In general we should be using retrying transactions. Our isolated
tests do retries at the outside, but non-isolated ones don't
necessarily.

Implement _run_and_rollback_retrying and use it some places that
TransactionSerializationError flakes have been seen (in #6954 and
in nightly CI).

Honestly I don't really think that these places ought to be having
TransactionSerializationErrors, but also it really is pretty required
to do retries when using SERIALIZABLE, so we should start moving
towards being more consistent about it.

### DIFF
--- a/edb/testbase/connection.py
+++ b/edb/testbase/connection.py
@@ -222,7 +222,7 @@ class Iteration(BaseTransaction, abstract.AsyncIOExecutor):
         self._options = retry._options.transaction_options
         self.__retry = retry
         self.__iteration = iteration
-        self.__started = False
+        self._started = False
 
     async def __aenter__(self):
         if self._managed:
@@ -233,7 +233,7 @@ class Iteration(BaseTransaction, abstract.AsyncIOExecutor):
 
     async def __aexit__(self, extype, ex, tb):
         self._managed = False
-        if not self.__started:
+        if not self._started:
             return False
 
         try:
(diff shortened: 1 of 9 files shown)


edb/testbase/connection.py
```diff
@@ -283,22 +283,29 @@ async def _ensure_transaction(self):
...
 
+class RawIteration(RawTransaction, Iteration):
+    async def _ensure_transaction(self):
+        self._started = True
+        await super()._ensure_transaction()
+
+
 class Retry:
-    def __init__(self, connection):
+    def __init__(self, connection, raw=False):
```
msullivan, line 302: Or should we make all our `Iteration`s be `RawTransaction`s?

msullivan: I ended up taking a different approach that allows Iteration to always contain nested transactions, but not to be in one

edb/testbase/connection.py
```diff
@@ -293,7 +291,7 @@ async def _ensure_transaction(self):
 
 
 class Retry:
-    def __init__(self, connection):
+    def __init__(self, connection, raw=False):
```
fantix, line 294:
```suggestion
    def __init__(self, connection):
```

Unused now

fantix (approved):
Yeah I think this is good.

PR #6963, opened 2024-03-01
Allow trailing separator for ListNonterm

dnwpark:


edb/common/parsing.py
```diff
@@ -147,36 +148,31 @@ def __init_subclass__(cls, *, is_internal=False, **kwargs):
...
-                tail_prod = (
-                    lambda self, lst, el: self._reduce_list(lst, el)
-                )
-                tail_prod_name = 'reduce_{}_{}'.format(
-                    cls.__name__, element_name)
-            setattr(cls, tail_prod_name, tail_prod)
-
-            setattr(cls, 'reduce_' + element_name,
-                lambda self, el: self._reduce_el(el))
+class ListHelper(Nonterm, is_internal=True):
```
msullivan, line 151: Does it help any to have this in a separate class? These could probably be staticmethods on `ListNonterm` itself, I think

msullivan (approved):
I like this!

It might be worth trying in a follow-up to add an `allow_empty` flag too, though we might run into LR(1) trouble again...

dnwpark:
I'll give a try for the `allow_empty` in a following PR.

PR #6967, opened 2024-03-01
Change grouping element to use trailing commas.

dnwpark:
Change `GroupingAtomList` and `GroupingElementList` to use trailing comma.


msullivan (commented):
Looks good, but please add a test to `test_edgeql_syntax.py`

msullivan (approved):

PR #6975, opened 2024-03-04
Explicitly pass WebAuthn credential properties

scotttrinh:
Not sure if this was working before due to some idiosyncratic behavior of Bitwarden's passkey implementation, but when I tried this with the built-in Chrome passkey functionality, I was getting errors about missing properties.

### DIFF
--- a/edb/server/protocol/auth_ext/_static/webauthn-authenticate.js
+++ b/edb/server/protocol/auth_ext/_static/webauthn-authenticate.js
@@ -150,10 +150,12 @@ async function authenticateAssertion(props) {
   // Assertion includes raw bytes, so need to be encoded as base64url
   // for transmission
   const encodedAssertion = {
-    ...props.assertion,
+    type: props.assertion.type,
+    id: props.assertion.id,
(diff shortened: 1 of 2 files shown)


edb/server/protocol/auth_ext/_static/webauthn-authenticate.js
```diff
@@ -150,10 +150,12 @@ async function authenticateAssertion(props) {
   // Assertion includes raw bytes, so need to be encoded as base64url
   // for transmission
   const encodedAssertion = {
-    ...props.assertion,
+    type: props.assertion.type,
+    id: props.assertion.id,
+    authenticatorAttachments: props.assertion.authenticatorAttachment,
```
jaclarke, line 155: Should this be `authenticatorAttachment` (no 's')?

scotttrinh: Good eye. It's optional so I guess it worked in my testing just accidentally 🙈 . Getting pretty close to creating a mock authenticator so we can actually test these end-to-end...

jaclarke (approved):

PR #6979, opened 2024-03-04
Allow `/branch/{dbname}` in extension routes

scotttrinh:
Allows `/branch` as the base route for our extensions while keeping support for `/db`. We do not currently plan on deprecating the `/db` route, but will be promoting `/branch` in our documentation.

I'll update the documentation as a separate step since we will probably want to align that with release, right?

### DIFF
--- a/edb/server/protocol/protocol.pyx
+++ b/edb/server/protocol/protocol.pyx
@@ -510,7 +510,7 @@ cdef class HttpProtocol:
                     b'The server is closing.',
                 )
 
-        if route == 'db':
+        if route == 'db' or route == 'branch':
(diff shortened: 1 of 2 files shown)


jaclarke:
I think this also needs changing: https://github.com/edgedb/edgedb/pull/6979/files#diff-b4dd339c8b00dd7f89d940c1b692e3bd860325daf67edf2a830d80cfda10e5e1R499

scotttrinh:
@jaclarke 
> I think this also needs changing:

Nice catch. I'll see if I can add a test as well.

msullivan (approved):

PR #6983, opened 2024-03-04
Add app deployment guides for Fly.io, Render, and Railway

beerose:


docs/guides/cloud/deploy/fly.rst
```diff
@@ -0,0 +1,34 @@
...
+    +      EDGEDB_SECRET_KEY="$(cat /run/secrets/EDGEDB_SECRET_KEY)" \
+    +      pnpm run build
+
+5. Run ``flyctl deploy`` to deploy your app to Fly.io. You need to pass the 
+   secrets to the deployment command:
+
+   .. code-block:: bash
+
+    $ flyctl deploy --secrets EDGEDB_INSTANCE="<EDGEDB_INSTANCE>" \
+        --secrets EDGEDB_SECRET_KEY="<EDGEDB_SECRET_KEY>"
```
beerose, line 34:
```suggestion
    $ flyctl deploy --build-secret EDGEDB_INSTANCE="<EDGEDB_INSTANCE>" \
        --build-secret EDGEDB_SECRET_KEY="<EDGEDB_SECRET_KEY>"
```

docs/guides/cloud/deploy/fly.rst
```diff
@@ -0,0 +1,34 @@
...
+    +      EDGEDB_SECRET_KEY="$(cat /run/secrets/EDGEDB_SECRET_KEY)" \
+    +      pnpm run build
+
+5. Run ``flyctl deploy`` to deploy your app to Fly.io. You need to pass the 
+   secrets to the deployment command:
+
+   .. code-block:: bash
+
+    $ flyctl deploy --build-secret EDGEDB_INSTANCE="<EDGEDB_INSTANCE>" \
+        --build-secret EDGEDB_SECRET_KEY="<EDGEDB_SECRET_KEY>"
```
raddevon, line 34: The build secrets didn't work for me. It may be something particular to my app I'm deploying, but I don't think it is. I believe for any of our clients, the env vars will need to be available at runtime. This required me to set them with `flyctl secrets set EDGEDB_INSTANCE=val EDGEDB_SECRET_KEY=val` instead of by using `--build-secret` on the `deploy` command.

Even if this isn't a universal requirement, we should probably mention both and describe the use case for each. I'm not sure of a case where `--build-secret` will work, but I could be wrong on that.

docs/guides/cloud/deploy/netlify.rst
```diff
@@ -1,8 +1,8 @@
 .. _ref_guide_cloud_deploy_netlify:
 
-========================
-Deploy an app on Netlify
-========================
+=======
+Netlify
```
raddevon, line 4: Saw your message on this earlier. Yeah, good call on this change. 👍

docs/guides/cloud/deploy/render.rst
```diff
@@ -0,0 +1,27 @@
...
+Render
+======
+
+:edb-alt-title: Deploying applications built on EdgeDB Cloud to Render
+
+1. Push project to GitHub or some other Git remote repository
+2. Create and make note of a secret key for your EdgeDB Cloud instance
+3. From Render's dashboard, click "New > Web Service"
+4. Import your project's repository
+5. In your new project's page, click "Environment" to add two variables:
```
raddevon, line 13: These can be set directly from the import without having to go through the project page. That's probably the most straightforward way for the purposes of the guide.
CleanShot 2024-03-06 at 07 45 27@2x

raddevon (changes requested):
Looking good! I had some issues with the fly.io guide, but the others all worked. Suggested a small optimization on the Render guide.

raddevon (approved):

PR #6990, opened 2024-03-05
Schema version in query cache key

fantix:
This PR adds schema_version into the query cache key, so that the cache table can be append-only to avoid concurrent update errors.

### DIFF
--- a/edb/server/dbview/dbview.pyx
+++ b/edb/server/dbview/dbview.pyx
@@ -1040,46 +1040,58 @@ cdef class DatabaseConnectionView:
         self,
         query_req: rpc.CompilationRequest,
         cached_globally=False,
-        use_metrics=True,
-        # allow passing in the query_unit_group, in case the call site needs
-        # to make decisions based on whether the cache is hit
-        query_unit_group=None,
+        bint use_metrics=True,
         uint64_t allow_capabilities = <uint64_t>compiler.Capability.ALL,
     ) -> CompiledQuery:
-        source = query_req.source
-        if cached_globally:
-            # WARNING: only set cached_globally to True when the query is
-            # strictly referring to only shared stable objects in user schema
-            # or anything from std schema, for example:
-            #     YES:  select ext::auth::UIConfig { ... }
-            #     NO:   select default::User { ... }
-            query_unit_group = (
-                self.server.system_compile_cache.get(query_req)
-                if self._query_cache_enabled
-                else None
-            )
-        elif query_unit_group is None:
-            query_unit_group = self.lookup_compiled_query(query_req)
+        query_unit_group = None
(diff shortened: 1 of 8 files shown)


edb/server/dbview/dbview.pyx
```diff
@@ -182,22 +182,18 @@ cdef class Database:
...
-        # Store the matching schema version, see also the comments at origin
-        self._eql_to_compiled[key] = compiled, schema_version
+        self._eql_to_compiled[key] = compiled
         keys = []
         while self._eql_to_compiled.needs_cleanup():
-            query_req, (unit_group, _) = self._eql_to_compiled.cleanup_one()
+            query_req, unit_group = self._eql_to_compiled.cleanup_one()
             if len(unit_group) == 1:
                 keys.append(query_req.get_cache_key())
         if keys:
```
msullivan, line 199: TODO reminder: I think we need to defer these evictions until after the transaction completes.

The question of how to handle cache entries created during transactions is a tricky one in general, and will be more so once we switch back to functions, since it won't be visible to other tasks until it commits.
We might want to track explicitly whether it has been persisted succesfully yet?

edb/server/dbview/dbview.pyx
```diff
@@ -966,9 +951,17 @@ cdef class DatabaseConnectionView:
...
                     self._db._eql_to_compiled.pop(query_req, None)
                 else:
-                    await compiled_queue.put(
-                        (query_req, result[0], schema_version)
-                    )
+                    # schema_version, database_config and system_config are not
+                    # serialized but they affect the cache key. We only update
+                    # these values *after* the compilation because 1) they are
+                    # not needed for compilation, and 2) we can evict the in-
+                    # memory cache by the same key when recompilation fails.
```
msullivan, line 958: `database_config` and `system_config` can affect the compilation, though. Am I missing something?

edb/server/dbview/dbview.pyx
fantix: Yes, I'm thinking about something like what we do in dbviews - once applied successfully in DB, we replay the stuff in memory (`dbview` for in-tx cache and `database` for shared ones)

msullivan (approved):

edb/server/dbview/dbview.pyx
fantix: user_schema, database_config and system_config are not serialized in the CompilationRequest because they are sent to the compiler process separately, while they are all affecting the cache key because they affect the compilation. I think I shouldn't say "they are not needed for compilation", but rather like "the compiler takes their values separately"

fantix: I tweaked the comments, and fixed a potential race issue on db_config and sys_config

PR #6995, opened 2024-03-05
Add tool to convert grammar to ebnf

dnwpark:


edb/edgeql/parser/grammar/convert.py
```diff
@@ -0,0 +1,418 @@
...
+import parsing
+import pathlib
+from typing import Dict, List, Sequence, Tuple, Callable
+
+
+class EBNF_Item:
+    pass
+
+
+class EBNF_Literal(EBNF_Item):
```
msullivan, line 29: Python style is no underscores in class names. I'd probably just drop the entire `EBNF` from the name

edb/edgeql/parser/grammar/convert.py
```diff
@@ -0,0 +1,418 @@
...
+            )
+
+    # repeatedly simplify until no simplifications are made
+    while True:
+        count = len(ebnf_productions)
+        ebnf_productions = simplify_productions(ebnf_productions)
+        if count == len(ebnf_productions):
+            break
+
+    # output
```
msullivan, line 414: I think the best place for this code to live would be as a tool in `edb/tools`. There are a bunch of examples there, and then we'd run it like `edb generate_ebnf` or some such.

edb/edgeql/parser/grammar/convert.py
```diff
@@ -0,0 +1,418 @@
...
+
+def ebnf_single_or_choice(items: Sequence[EBNF_Item]) -> EBNF_Item:
+    if len(items) == 1:
+        return items[0]
+    else:
+        return EBNF_Choice(items)
+
+
+def expand_iso_ebnf(item: EBNF_Item) -> str:
+    if isinstance(item, EBNF_Literal):
```
msullivan, line 85: This (and a bunch of other places) would be a perfect time to try out python's new pattern matching system (`match`)

edb/edgeql/parser/grammar/convert.py
```diff
@@ -0,0 +1,418 @@
...
+#
+
+
+import parsing
+import pathlib
+from typing import Dict, List, Sequence, Tuple, Callable
+
+
+class EBNF_Item:
+    pass
```
msullivan, line 26: I'd recommend using either `dataclass` or else our own `ast` mechanism. (Probably `dataclass`).

You get to skip writing `__init__`, the attributes and their types are spelled out more cleanly, and you automatically get string formatting.

edb/edgeql/parser/grammar/convert.py
```diff
@@ -0,0 +1,418 @@
...
+                for inner_item in item.inner
+            ) + ')'
+        )
+    else:
+        raise NotImplementedError
+
+
+def to_iso_ebnf(productions: List[EBNF_Production]) -> List[str]:
+    return [
+        production.name + ' = ' + expand_iso_ebnf(production.item) + ';'
```
msullivan, line 117: Consider using f-strings instead of doing string concatenation.

msullivan: (Though use your judgement about whether it's actually an improvement

edb/edgeql/parser/grammar/convert.py
```diff
@@ -0,0 +1,418 @@
...
+                ])), True
+
+        return item, False
+
+    # apply replacements until no changes are made
+
+    def replace_repeatedly(
+                item: EBNF_Item,
+                funcs: List[Callable[[EBNF_Item], Tuple[EBNF_Item, bool]]]
+            ) -> EBNF_Item:
```
msullivan, line 303: We aren't totally consistent about these things, but typically edgedb style for this kind of definition is to have the closing paren lined up with the `def` and the body indented one step. (This is what black does, among others)

edb/edgeql/parser/grammar/convert.py
```diff
@@ -0,0 +1,418 @@
...
+        while changed:
+            changed = False
+            for func in funcs:
+                item, curr_changed = func(item)
+                changed = changed or curr_changed
+        return item
+
+    def replace_inlines(item: EBNF_Item) -> EBNF_Item:
+        return replace_repeatedly(
+                item, [
```
msullivan, line 315: Only indent one step

edb/edgeql/parser/grammar/convert.py
dnwpark: I thought this bit was a bit too short to use formatting.

edb/edgeql/parser/grammar/convert.py
dnwpark: Moved to separate file. (For now, will move things to tools later)

edb/tools/railroad_diagram.py
```diff
@@ -0,0 +1,519 @@
...
+    class Literal(Item):
+        token: str
+
+    @dataclass(eq=False, match_args=False)
+    class Reference(Item):
+        name: str
+
+    @dataclass(eq=False, match_args=False)
+    class Single(Item):
+        inner: 'ebnf.Item'
```
msullivan, line 69: Use `from __future__ import annotations`, and then drop the quotes.

edb/tools/railroad_diagram.py
```diff
@@ -0,0 +1,519 @@
...
+        count = len(ebnf_productions)
+        ebnf_productions = simplify_productions(ebnf_productions)
+        if count == len(ebnf_productions):
+            break
+
+    # output
+    with open(path / 'grammar.iso.ebnf', 'w') as file:
+        file.write('\n'.join(to_iso_ebnf(ebnf_productions)))
+    with open(path / 'grammar.w3c.ebnf', 'w') as file:
+        file.write('\n'.join(to_w3c_ebnf(ebnf_productions)))
```
msullivan, line 519: I'd split the output out from the function doing the work

msullivan (approved):

PR #6996, opened 2024-03-06
Support new default branch behavior

msullivan:
* Default branch is now 'main'
 * --default-branch allows specifying a different default
 * If 'branch' is sent with ClientHandshake, it takes priority.
   If it is `__default__`, then the default branch is used
 * The 'database' ClientHandshake param is used otherwise.
   If it is `edgedb` and `edgedb` does not exist, then
   the default branch is used.
 * Similar behavior for HTTP routes.


edb/server/defines.py
```diff
@@ -29,7 +29,8 @@
 EDGEDB_SUPERGROUP = 'edgedb_supergroup'
 EDGEDB_SUPERUSER = s_def.EDGEDB_SUPERUSER
 EDGEDB_TEMPLATE_DB = s_def.EDGEDB_TEMPLATE_DB
-EDGEDB_SUPERUSER_DB = 'edgedb'
+EDGEDB_OLD_DEFAULT_DB = 'edgedb'
+EDGEDB_SUPERUSER_DB = 'main'
```
fantix, line 33: `EDGEDB_SUPERUSER_DB` is also used in 2 functions in `metaschema.py`, should we patch up those 2 functions?

edb/server/bootstrap.py
```diff
@@ -2484,13 +2491,13 @@ async def _bootstrap(ctx: BootstrapContext) -> edbcompiler.CompilerState:
...
-            edbdef.EDGEDB_SUPERUSER_DB,
-            edbdef.EDGEDB_SUPERUSER,
+            default_branch,
+            args.default_database_user or edbdef.EDGEDB_SUPERUSER,
         )
     else:
         await _set_edgedb_database_metadata(
             ctx,
-            edbdef.EDGEDB_SUPERUSER_DB,
+            default_branch,
```
fantix, line 2500: We should also use `default_branch` in `system_api.py` if not `has_create_database`?

> ```
> async def _ping(tenant):
>     if tenant.get_backend_runtime_params().has_create_database:
>         dbname = edbdef.EDGEDB_SYSTEM_DB
>     else:
>         dbname = edbdef.EDGEDB_SUPERUSER_DB  # FIXME
> ```

fantix (approved):

edb/server/defines.py
msullivan: I'm not sure if those functions are even right, really?

msullivan: Hmmm... and I bet that patches don't work in single-db mode

fantix: Probably true, we don't have a patch test CI on single-db mode. Though the regular single-db CI has been green.

fantix:
Looks like now I can access the same default database with both `edgedb` and `main`:

```
$ edgedb -H localhost --database main query 'select schema::ObjectType { name }'
...
{"name": "default::A"}

$ edgedb -H localhost --database edgedb query 'select schema::ObjectType { name }'
...
{"name": "default::A"}
```

But this doesn't apply on the emulated Postgres endpoint:

```
$ psql -h localhost -U edgedb -p 5656 main -c 'select * from "A"'
 id | __type__
----+----------
(0 rows)

$ psql -h localhost -U edgedb -p 5656 edgedb -c 'select * from "A"'
psql: error: connection to server at "localhost" (::1), port 5656 failed: ERROR:  database 'edgedb' does not exist
```

PR #6999, opened 2024-03-06
Fix computed single scalar globals

### LINKED ISSUE #6988: ISE: cannot create root rvar for non-object path
- EdgeDB Version: 5.0-beta.1+15775e6
- EdgeDB CLI Version: 4.1.0+127aeb9
- OS Version: Win 11

Steps to Reproduce:

1. Try to make a migration `edgedb migrate` with attached initial migration: 
[00001_edgeql.txt](https://github.com/edgedb/edgedb/files/14491317/00001_edgeql.txt)

msullivan:
They were being cached, but we were rejecting attempts to use the
cache because of an overly zealous assert. Make the assert more
permissive.

Fixes #6988.

### DIFF
--- a/edb/pgsql/compiler/relctx.py
+++ b/edb/pgsql/compiler/relctx.py
@@ -501,9 +501,6 @@ def new_primitive_rvar(
     lateral: bool,
     ctx: context.CompilerContextLevel,
 ) -> pgast.PathRangeVar:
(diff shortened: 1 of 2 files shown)


elprans (approved):

aljazerzen:
To backport this, I had to pull-in a few snippets from #6613

PR #7002, opened 2024-03-07
[5.x] Allow grouping to have trailing comma.

dnwpark:


msullivan (approved):

msullivan:
Actually we can backport this to other versions super easily now too

msullivan:
Oh, though #6931 is a prereq

PR #7003, opened 2024-03-07
Remove all the unused nonterms and tokens from the parser

msullivan:


aljazerzen (approved):

dnwpark (approved):
LGTM. Making parser no longer produces warnings for unused productions.

PR #7005, opened 2024-03-07
Enable more ruff lints

aljazerzen:
Not a lot, but now it should find trailing whitespace and a few other things.
See: https://docs.astral.sh/ruff/rules/#warning-w

There is a lot more lints in "preview" and they can be ran with:
```
ruff check --preview .
```
This PR also fixes most of them and disables other checks.


msullivan (approved):
We can probably just turn on `--preview`, I'm not too worried

msullivan:
(Not actually backported)

PR #7009, opened 2024-03-07
Fix a ResourceWarning caused by SimpleAdaptivePool

msullivan:
This was a huge pain to track down and fix and probably not worth
spending my morning on it. I ended up editing my installed
`subprocess.py` to see what the command line was for the process that
was leaking.

I then spent a long time thinking there might be a uvloop bug, because
I confused myself and thought we were using uvloop in the test suite
(after I added a debug print that failed under uvloop---but the
failure occured when running the real server).

The confusing part to me was that I felt that a subprocess protocol
shouldn't be getting notified that it has disconnected unless the
child process has actually exited, but I had missed that the protocol
we use to communicate with the workers is a normal socket protocol
which *obviously* closes when the socket does.

Anyway the fix is to arrange to wait for the children to exit, which
like everything involving protocols is annoying because the callback
isn't async.

### DIFF
--- a/edb/server/compiler_pool/pool.py
+++ b/edb/server/compiler_pool/pool.py
@@ -885,6 +885,7 @@ class SimpleAdaptivePool(BaseLocalPool):
         self._expected_num_workers = 0
         self._scale_down_handle = None
         self._max_num_workers = pool_size
+        self._cleanups = {}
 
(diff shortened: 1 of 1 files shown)


fantix (approved):
LGTM!

PR #7015, opened 2024-03-08
Handle in-transaction query cache properly

fantix:
- [x] Disable in-tx cache while persistent_cache is on ([dry run](https://github.com/edgedb/edgedb/actions/runs/8205312820))
- [ ] Properly fix in-tx cache by tracking in dbview and writeback after transaction commits

### DIFF
--- a/edb/server/dbview/dbview.pyx
+++ b/edb/server/dbview/dbview.pyx
@@ -200,7 +200,7 @@ cdef class Database:
             query_req, unit_group = self._eql_to_compiled.cleanup_one()
             if len(unit_group) == 1:
                 keys.append(query_req.get_cache_key())
-        if keys and debug.flags.persistent_cache:
+        if keys and not debug.flags.disable_persistent_cache:
             self.tenant.create_task(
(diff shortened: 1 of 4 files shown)


edb/server/dbview/dbview.pyx
```diff
@@ -711,8 +711,14 @@ cdef class DatabaseConnectionView:
...
 
-        if not self._in_tx_with_ddl:
+        if not self._in_tx:
+            self._db._cache_compiled_query(key, query_unit_group)
+        elif not self._in_tx_with_ddl and not debug.flags.persistent_cache:
             self._db._cache_compiled_query(key, query_unit_group)
+        # Query compiled in transaction is not yet cached when persistent_cache
+        # is turned on, because eviction of such query would fail if the tx
+        # ended up rolling back, or simply committed after the eviction.
+        # TODO(fantix): handle in-tx cache properly
```
msullivan: Wait, this means it won't even go into the in-memory cache?
I think that is probably too harsh.

fantix:
Close in favor of #7025

PR #7018, opened 2024-03-08
Document drop branch force

raddevon:
Added this to the Reference > Administration > Database page since I think it needs to be there. Its canonical home should be on a new Reference > Administration > Branch page. As part of a future PR, I will comb through the docs to update "database" to "branch" wherever appropriate, and I will also add the aforementioned page with a note on the Administration > Database page that "databases" were renamed "branches" in v5.

### DIFF
--- a/docs/reference/admin/databases.rst
+++ b/docs/reference/admin/databases.rst
@@ -50,6 +50,16 @@ Remove a database.
 
     drop database <name> ;
 
+.. versionadded:: 5.0
+
+    You may now drop a branch, even if it has active connections by adding
+    ``force``:
+
+    .. eql:synopsis::
+
+        drop branch <name> [force] ;
+
+
 Description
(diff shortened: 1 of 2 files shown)


docs/reference/admin/databases.rst
```diff
@@ -50,6 +50,16 @@ Remove a database.
 
     drop database <name> ;
 
+.. versionadded:: 5.0
+
+    You may now drop a branch, even if it has active connections by adding
+    ``force``:
```
fantix, line 56: Do we need some kind of a warning here to say, this is not usually recommended because it may interrupt ongoing transactions, use with extra caution? @msullivan

fantix (approved):

raddevon:
I think this one is no longer relevant after the branches documentation I've merged.

PR #7021, opened 2024-03-08
Doc v5 pgvector

raddevon:


vpetrovykh (approved):
Looks good.

It's probably worth adding that the config commands that are valid for this extension are `administer config session` and `administer config current branch`. So you can set the config in the current branch and have it persist that way.

raddevon:
@vpetrovykh Would you take another look and see if I've understood you correctly?

PR #7022, opened 2024-03-08
Add "branch" to database does not exist error message

msullivan:


edb/server/dbview/dbview.pyx
```diff
@@ -1303,7 +1303,7 @@ cdef class DatabaseIndex:
             return self._dbs[dbname]
         except KeyError:
             raise errors.UnknownDatabaseError(
-                f'database {dbname!r} does not exist')
+                f'database branch {dbname!r} does not exist')
```
scotttrinh, line 1306: Keeping `database` here seems debatable, but I don't have a strong argument in favor of either direction, so lgtm!

scotttrinh (approved):

edb/server/dbview/dbview.pyx
msullivan: We've done similar in a lot of places in the server

PR #7025, opened 2024-03-08
 Do persistent cache writebacks asynchronously

msullivan:
Whenever a entry is added to the in memory cache, enqueue it in a
queue that a per-db worker task consumes. The worker task manages all
cache inserts and evictions, which should fully eliminate the
serialization error dangers. This also lets us persistently cache
queries that were first compiled from within a rolled back
transaction.

I also re-enabled the query-cache-changes notifications, with some
very minor rate limiting. We'll see how it goes.

Thoughts?

FUTURE NOTE: The situation for function caching once we want to
reintroduce that will be a little more complicated. We'll need to put
both the SQL text and the function call code in the in-memory cache,
and use the text until the function is visible everywhere (at which
point we can drop the text.) We'll also need to do some thinking about
how to test it properly, because the downside of the approach is that
in the typically path, the first execution of a query can't use the
function.  We may need some sort of testing path to allow us to
exercise the functions easily in the test code.

### DIFF
--- a/edb/common/asyncutil.py
+++ b/edb/common/asyncutil.py
@@ -98,8 +98,11 @@ async def debounce(
 
     while True:
         try:
-            async with asyncio.timeout_at(target_time):
+            if target_time is None:
                 v = await input()
+            else:
+                async with asyncio.timeout_at(target_time):
+                    v = await input()
         except TimeoutError:
             t = loop.time()
         else:
--- a/edb/common/asyncutil.py
+++ b/edb/common/asyncutil.py
@@ -18,9 +18,10 @@
 
 
 from __future__ import annotations
-from typing import TypeVar, Awaitable
+from typing import Callable, TypeVar, Awaitable
 
 import asyncio
+import time
 
 
 _T = TypeVar('_T')
@@ -62,3 +63,77 @@ async def deferred_shield(arg: Awaitable[_T]) -> _T:
     if ex:
         raise ex
     return task.result()
+
+
+async def debounce(
+    input: Callable[[], Awaitable[_T]],
+    output: Callable[[list[_T]], Awaitable[None]],
+    *,
+    max_wait: float,
+    delay_amt: float,
+    max_batch_size: int,
+) -> None:
+    '''Debounce and batch async events.
+
+    Loops forever unless an operation fails, so should probably be run
+    from a task.
+
+    The basic algorithm is that if an event comes in less than
+    `delay_amt` since the previous one, then instead of sending it
+    immediately, we wait an additional `delay_amt` from then. If we are
+    already waiting, any message also extends the wait, up to
+    `max_wait`.
+
+    Also, cap the maximum batch size to `max_batch_size`.
+    '''
+    # I think the algorithm reads more clearly with the params
+    # capitalized as constants, though we don't want them like that in
+    # the argument list, so reassign them.
+    MAX_WAIT, DELAY_AMT, MAX_BATCH_SIZE = max_wait, delay_amt, max_batch_size
+
+    batch = []
+    last_signal = 0.0
+    target_time = None
+
+    while True:
+        target_delta = target_time and target_time - time.monotonic()
+        try:
+            v = await asyncio.wait_for(
+                input(),
+                timeout=target_delta,
+            )
+        except TimeoutError:
+            t = time.monotonic()
+        else:
+            batch.append(v)
+
+            t = time.monotonic()
+
+            # If we aren't current waiting, and we got a
+            # notification recently, arrange to wait some before
+            # sending it.
+            if (
+                target_time is None
+                and t - last_signal < DELAY_AMT
+            ):
+                target_time = t + DELAY_AMT
+            # If we were already waiting, wait a little longer, though
+            # not longer than MAX_WAIT.
+            elif (
+                target_time is not None
+                and target_time < last_signal + MAX_WAIT
+            ):
+                target_time = max(t + DELAY_AMT, target_time)
+
+        # Skip sending the event if we need to wait longer.
+        if (
+            target_time is not None
+            and t < target_time
+            and len(batch) < MAX_BATCH_SIZE
+        ):
+            continue
+
+        await output(batch)
+        batch.clear()
+        last_signal = t
+        target_time = None
--- a/edb/common/asyncutil.py
+++ b/edb/common/asyncutil.py
@@ -91,23 +91,22 @@ async def debounce(
     # the argument list, so reassign them.
     MAX_WAIT, DELAY_AMT, MAX_BATCH_SIZE = max_wait, delay_amt, max_batch_size
 
+    loop = asyncio.get_running_loop()
+
     batch = []
     last_signal = 0.0
     target_time = None
 
     while True:
-        target_delta = target_time and target_time - time.monotonic()
         try:
-            v = await asyncio.wait_for(
-                input(),
-                timeout=target_delta,
-            )
+            async with asyncio.timeout_at(target_time):
+                v = await input()
         except TimeoutError:
-            t = time.monotonic()
+            t = loop.time()
         else:
             batch.append(v)
 
-            t = time.monotonic()
+            t = loop.time()
 
             # If we aren't current waiting, and we got a
             # notification recently, arrange to wait some before
@@ -134,6 +133,6 @@ async def debounce(
             continue
 
         await output(batch)
-        batch.clear()
+        batch = []
         last_signal = t
         target_time = None
--- a/edb/common/asyncutil.py
+++ b/edb/common/asyncutil.py
@@ -21,7 +21,6 @@ from __future__ import annotations
 from typing import Callable, TypeVar, Awaitable
 
 import asyncio
-import time
 
 
 _T = TypeVar('_T')
@@ -94,7 +93,7 @@ async def debounce(
     loop = asyncio.get_running_loop()
 
     batch = []
-    last_signal = 0.0
+    last_signal = -MAX_WAIT
     target_time = None
 
     while True:
@@ -120,9 +119,11 @@ async def debounce(
             # not longer than MAX_WAIT.
             elif (
                 target_time is not None
-                and target_time < last_signal + MAX_WAIT
             ):
-                target_time = max(t + DELAY_AMT, target_time)
+                target_time = min(
+                    max(t + DELAY_AMT, target_time),
+                    last_signal + MAX_WAIT,
+                )
 
         # Skip sending the event if we need to wait longer.
         if (
--- a/edb/server/dbview/dbview.pyx
+++ b/edb/server/dbview/dbview.pyx
@@ -220,7 +220,7 @@ cdef class Database:
             query_req, unit_group = self._eql_to_compiled.cleanup_one()
             if len(unit_group) == 1:
                 keys.append(query_req.get_cache_key())
-        if keys and debug.flags.persistent_cache:
+        if keys and not debug.flags.disable_persistent_cache:
             self.tenant.create_task(
                 self.tenant.evict_query_cache(self.name, keys),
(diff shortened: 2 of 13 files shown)


edb/server/dbview/dbview.pyx
```diff
@@ -148,6 +160,73 @@ cdef class Database:
...
+            while self._eql_to_compiled.needs_cleanup():
+                query_req, unit_group = self._eql_to_compiled.cleanup_one()
+                if len(unit_group) == 1 and unit_group.cache_state == 1:
+                    keys.append(query_req.get_cache_key())
+                unit_group.cache_state = CacheState.Evicted
+            if keys:
+                await self.tenant.evict_query_cache(self.name, keys)
+
+            # Now, populate the cache
+            query_req, units = await self._cache_queue.get()
```
1st1, line 194: Do you want to drain the queue right here to group DB writes into one query?

edb/server/dbview/dbview.pyx
```diff
@@ -148,6 +160,73 @@ cdef class Database:
...
+
+    async def cache_worker(self):
+        while True:
+            try:
+                await self._cache_worker()
+            except Exception as ex:
+                debug.dump(ex)
+                metrics.background_errors.inc(
+                    1.0, self.tenant._instance_name, "cache_worker"
+                )
```
1st1, line 176: I'm usually very nervous about `while True` that doesn't wait before the next attempt. I'd add `await sleep(0.1)` to the `except` clause. In case of some errors you might want to give the system time to recover, and not saturate it with BS work until it's done.

edb/server/dbview/dbview.pyx
```diff
@@ -148,6 +160,73 @@ cdef class Database:
...
+
+                units.cache_state = CacheState.Present
+
+                added_since_signal += 1
+
+            # Only signal query-cache-changes if we are quiescent or
+            # we have a lot that have updated since the last send.
+            # TODO: Is this good enough? Should we be smarter and try
+            # to rate-limit more and/or include a detailed payload.
+            await asyncio.sleep(0)
```
1st1, line 216: What's this `sleep(0)` is doing here? I don't see how the comment above explains it. And if it's not clear, expect somebody to remove it one day.

1st1 (commented):
Good work, Sully. Added some comments, lmk if you need help implementing the debounce thingy.

edb/server/dbview/dbview.pyx
1st1: Basically I'd do an `await queue.get()` followed by a series of `queue.get_nowait()` until they complain

edb/server/dbview/dbview.pyx
msullivan: The idea is to give anything in-flight a chance to enqueue things without actually waiting any. It did lead to an increase in how often this case got hit.

edb/server/dbview/dbview.pyx
```diff
@@ -148,6 +160,73 @@ cdef class Database:
...
+            if (
+                added_since_signal
+                and (
+                    self._cache_queue.empty()
+                    or added_since_signal > 100
```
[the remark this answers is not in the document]
msullivan: Yeah I planned to do something along those lines, but I wanted to see if the aarch64 trouble recurred after these changes without it, but the aarch64 builds are too broken because of spot instance trouble to know yet really

edb/server/dbview/dbview.pyx
msullivan: Sure. @fantix's existing infrastructure makes it super trivial to do

edb/server/dbview/dbview.pyx
```diff
@@ -174,6 +174,9 @@ cdef class Database:
                 metrics.background_errors.inc(
```
1st1, line 174: ^^ Also we should probably stop using `debug.dump` directly in code and do proper `logging.error` or whatnot. If it's not printing when debugging that should be debugged

edb/server/dbview/dbview.pyx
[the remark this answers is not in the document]
msullivan: Alright I implemented something. Let me know if it basically matches what you were thinking

edb/server/dbview/dbview.pyx
```diff
@@ -221,30 +221,82 @@ cdef class Database:
...
+            # any message also extends the wait, up to MAX_WAIT.
+            # Also, cap the batch size to MAX_KEYS, since
+            target_delta = target_time and target_time - time.time()
+            try:
+                key = await asyncio.wait_for(
+                    self._cache_notify_queue.get(),
+                    timeout=target_delta,
+                )
+            except TimeoutError:
+                t = time.time()
```
1st1, line 262: Please use `time.monotonic()` for things like this, `time.time()` can jump forward/backward pretty arbitrarily

edb/server/dbview/dbview.pyx
```diff
@@ -191,26 +194,35 @@ cdef class Database:
...
 
             # Now, populate the cache
-            query_req, units = await self._cache_queue.get()
-            # TODO: Should we do any sort of error handling here?
-            if (
-                len(units) == 1
+            # Empty the queue, for batching reasons.
+            ops = [await self._cache_queue.get()]
+            while not self._cache_queue.empty():
+                ops.append(self._cache_queue.get_nowait())
```
1st1, line 200: I'd add a comment that this is safe for asyncio queues (unlike thread queues, where the empty check /get_nowait combo is not guaranteed to succeed due to races)

1st1: (I froze here for a few moments thinking, mistakingly, that your code is incorrect)

edb/server/dbview/dbview.pyx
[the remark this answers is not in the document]
1st1: I read the code a few times and I think it's good. I want @fantix to thoroughly read it too.

1st1 (approved):
I think this is in a good state now. `time.time()` must be swapped with `time.monotonic()` -- that's my last major comment. @fantix please review this thoroughly.

edb/server/dbview/dbview.pyx
msullivan: Oh yeah; I felt naughty while doing this, but should have added a comment

edb/server/dbview/dbview.pyx
```diff
@@ -148,6 +163,141 @@ cdef class Database:
     def tenant(self):
         return self._index._tenant
 
+    def stop(self):
+        if self._cache_worker_task:
+            self._cache_worker_task.cancel()
+            self._cache_worker_task = None
```
fantix, line 169: Also cancel the `_cache_notify_task`?

edb/server/dbview/dbview.pyx
```diff
@@ -148,6 +163,141 @@ cdef class Database:
...
+            try:
+                await self._cache_notifier()
+            except Exception as ex:
+                debug.dump(ex)
+                metrics.background_errors.inc(
+                    1.0, self.tenant._instance_name, "cache_notifier"
+                )
+                # Give things time to recover, since the likely
+                # failure mode here is a failover or some such.
+                await asyncio.sleep(0.1)
```
fantix, line 234: Perhaps we can have just one copy of this method, passing the worker function in

edb/server/dbview/dbview.pyx
```diff
@@ -148,6 +163,141 @@ cdef class Database:
...
+            ops = [
+                (query_req, units) for query_req, units in ops
+                if len(units) == 1
+                and units[0].cache_sql
+                and units.cache_state == CacheState.Pending
+            ]
+            if not ops:
+                continue
+
+            # TODO: Should we do any sort of error handling here?
```
fantix, line 211: I guess this can wait until func cache is turned on.

edb/server/dbview/dbview.pyx
```diff
@@ -148,6 +163,141 @@ cdef class Database:
...
+        target_time = None
+
+        while True:
+            # "Debounce" and batch query-cache-change notifications.
+            # The basic algorithm is that if a notification comes in
+            # less than DELAY_AMT since we triggered the previous one,
+            # then instead of sending it immediately, we wait an
+            # additional DELAY_AMT from then. If we are already waiting,
+            # any message also extends the wait, up to MAX_WAIT.
+            # Also, cap the batch size to MAX_KEYS, since
```
fantix, line 254:
```suggestion
            # Also, cap the batch size to MAX_KEYS, since we include the keys in the
            # Postgres event which has a size limit.
```

edb/server/dbview/dbview.pyx
```diff
@@ -148,6 +163,141 @@ cdef class Database:
...
+            target_delta = target_time and target_time - time.time()
+            try:
+                key = await asyncio.wait_for(
+                    self._cache_notify_queue.get(),
+                    timeout=target_delta,
+                )
+            except TimeoutError:
+                t = time.time()
+            else:
+                pending_keys.append(str(key))
```
fantix, line 264: I would also drain the queue up to `MAX_KEYS` pending keys here, because `asyncio.wait_for()` jumps at least 1 iteration in the event loop if timeout is not None, and we may end up with 100 tight asyncio event loops with our `while True` loop, which sounds like an expensive CPU moment.

fantix (changes requested):
I think the caching part is good, some nits below; but the pgcon state tracking is broken (I'm surprised no test caught it, I'll add a test or fix one).

edb/server/dbview/dbview.pyx
msullivan: I may just try changing it to use `timeout_at` instead of `wait_for`, which looks like it won't create a task and so won't yield back to the event loop in the case of `get` succeeding immediately.

One reason I'm reluctant to switch to explicitly draining the queue is because the latest version has refactored it to just have an `input` function instead of operating on a queue (though I could change it back to having an input queue).

edb/server/protocol/execute.pyx
```diff
@@ -349,22 +309,11 @@ async def execute(
...
-                # In 3 cases the state is changed:
-                #   1. The non-tx query changed the state
-                #   2. The state is synced with dbview (orig_state is None)
-                #   3. We came out from a transaction (orig_state is None)
-                # Excluding two special case when the state is NOT changed:
-                #   1. An orphan ROLLBACK command without a paring start tx
-                #   2. There was no SQL, so the state can't have been synced.
-                be_conn.last_state = state
+        if query_unit.tx_rollback:
+            be_conn.last_state = None
```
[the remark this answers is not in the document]
msullivan: Good catch.

The problem was that `run_ddl` didn't sync state in all cases.

msullivan:
I pushed some changes but they aren't visible yet, and there appears to be a github incident: https://www.githubstatus.com/incidents/q3gl44chgxwv

fantix (approved):

edb/common/asyncutil.py
```diff
@@ -62,3 +62,78 @@ async def deferred_shield(arg: Awaitable[_T]) -> _T:
...
+    loop = asyncio.get_running_loop()
+
+    batch = []
+    last_signal = -MAX_WAIT
+    target_time = None
+
+    while True:
+        try:
+            async with asyncio.timeout_at(target_time):
+                v = await input()
```
1st1, line 102: Hm, so you will be waiting with timeout even for the first `input()` return, but that's not what we want to do. We don't want to keep looping through timeouts until `input()` returns for the first time.

We want to start debouncing only when it returns. The first await on it shouldn't be with a timeout at all.

1st1: Oh sigh, this is why I don't like None for cases like this. Please add an `if`.

```
if target_time is not None:
    async with ...
else:
    await input()
```

1st1 (approved):
One nit comment which I do want to be fixed, otherwise all good.

PR #7030, opened 2024-03-11
Document release process of edgedb-server

aljazerzen:


dev-notes/release-process.md
```diff
@@ -0,0 +1,161 @@
...
+practice is to list all commits since the last release:
+
+```
+git show releases/4.x # to see the last commit that has been cherry-picked
+git log master # find the hash of that commit on master
+git log hash_of_that_commit..master > ../to-backport.txt
+```
+
+Now, one can go trough the list and see if the commits are worth back-porting.
+A few pointers:
```
msullivan, line 38: Generally I'd say don't backport features also, unless it's high priority in some way

dev-notes/release-process.md
```diff
@@ -0,0 +1,161 @@
...
+  Disregard this rule early on after the fork of the release branch, since
+  porting refactors will decrease chances merge conflicts of other commits later
+  on.
+- Don't backport "build" commits (updating of build deps, refactoring of the
+  release pipeline), since that might trigger problems in the release process.
+- If a PR changes any of the schema objects (i.e. adding a field to
+  `s_types.Type`) or metaschema (i.e. changing a pg function
+  `edgedb.range_to_jsonb`), a "patch" needs to be added into `pgsql/patches.py`.
+  This is needed, because minor releases don't require a "dump and restore", so
+  we must apply these changes to existing user databases.
```
msullivan, line 52: Also, adding or modifying any standard library function. (Which is *hopefully* more of a common case than adding a field to a schema object)

msullivan: Should mention that whenever a patch is added, we should make sure that the "Tests of patching old EdgeDB Versions" works: https://github.com/edgedb/edgedb/actions/workflows/tests-patches.yml

PR #7034, opened 2024-03-12
Fallback to PKCE RFC parameter names

### LINKED ISSUE #7026: Aliasing the PKCE Parameter "challenge" for RFC Alignment
Currently the PKCE flow for the Auth Module expects the following parameter, `challenge`.

The [PKCE RFC's](https://datatracker.ietf.org/doc/html/rfc7636#section-6.1) Parameter for this is `code_challenge`.

Given the amount of code that likely targets `challenge`, I'd assume it would be best to let both `challenge` and `code_challenge` work for a period of time, with `challenge` taking precedence for the foreseeable future.

This would allow for easier integrations with Spring Security for example, as it currently supports EdgeDB Auth until you hit login.

Documentation on if EdgeDB supports the `code_challenge_method` would be good as well.

scotttrinh:
I played it a little fast-and-loose when implementing our PKCE implementation since we are not an OAuth authorization flow and therefore are not building an _actual_ PKCE flow, just something inspired by it. However, there are some existing tools that can take advantage of the code exchange if we align with the same parameter names outlined in [RFC 7636](https://datatracker.ietf.org/doc/html/rfc7636#section-6.1).

Added a note in the docs for anyone who might be familiar with how RFC 7636 defines PKCE to highlight the differences.

Closes #7026 

---

Special note here about omitting support for `code_challenge_method`: 

>   If the client is capable of using "S256", it MUST use "S256", as
>   "S256" is Mandatory To Implement (MTI) on the server.  Clients are
>   permitted to use "plain" only if they cannot support "S256" for some
>   technical reason and know via out-of-band configuration that the
>   server supports "plain".

[RFC 7636 Section 4.2](https://datatracker.ietf.org/doc/html/rfc7636#section-4.2)

Since our PKCE flow is _always_ a server-to-server flow, we _require_ that clients provide a SHA256 hashed challenge, and do not accept `plain` as a `code_challenge_method` since it negates the added security of PKCE. The RFC specifies that if the `code_challenge_method` is omitted that the server treat the challenge as `plain`, but we do not support `plain`. I considered conforming here but always requiring implementors to provide the constant `S256` value, but that seemed low value and an unnecessary hurdle. In the future we could chose to require it and return a failure status of 400 if you do not provide it if we want to add additional compatibility with RFC 7636.

### DIFF
--- a/edb/server/protocol/auth_ext/http.py
+++ b/edb/server/protocol/auth_ext/http.py
@@ -388,8 +388,8 @@ class Router:
             request.url.query.decode("ascii") if request.url.query else ""
         )
         code = _get_search_param(query, "code")
-        verifier = _get_search_param(query, "verifier") or _get_search_param(
-            query, "code_verifier"
+        verifier = _get_search_param(
+            query, "verifier", fallback_keys=["code_verifier"]
         )
 
         verifier_size = len(verifier)
--- a/edb/server/protocol/auth_ext/http.py
+++ b/edb/server/protocol/auth_ext/http.py
@@ -253,7 +253,9 @@ class Router:
             raise errors.InvalidData(
                 "Redirect URL does not match any allowed URLs.",
             )
-        challenge = _get_search_param(query, "challenge")
+        challenge = _get_search_param(
+            query, "challenge", fallback_keys=["code_challenge"]
+        )
         oauth_client = oauth.Client(
             db=self.db, provider_name=provider_name, base_url=self.test_url
         )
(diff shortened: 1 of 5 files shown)


edb/server/protocol/auth_ext/http.py
```diff
@@ -386,7 +388,9 @@ async def handle_token(self, request: Any, response: Any):
             request.url.query.decode("ascii") if request.url.query else ""
         )
         code = _get_search_param(query, "code")
-        verifier = _get_search_param(query, "verifier")
+        verifier = _get_search_param(query, "verifier") or _get_search_param(
+            query, "code_verifier"
+        )
```
fantix, line 393: Use `fallback_keys` here?

fantix (approved):

edb/server/protocol/auth_ext/http.py
scotttrinh: oh, thank you. I did a search for challenge, but forgot about verifier. Lemme double check and add a test.

PR #7040, opened 2024-03-13
Disable persistent cache on arm64 linux

msullivan:


edb/common/debug.py
```diff
@@ -176,7 +178,15 @@ class flags(metaclass=FlagsMeta):
 
     zombodb = Flag(doc="Enabled zombodb and disables postgres FTS")
 
-    disable_persistent_cache = Flag(doc="Don't use persistent cache")
+    disable_persistent_cache = Flag(
```
fantix, line 181: Do we want to again flip this to `persistent_cache` now that we allow `default=True` and overriding to `False` to save writing the `DISABLE_` prefix in env var?

fantix (approved):

edb/common/debug.py
msullivan: We could, though similar things are mostly `DISABLE` currently

msullivan: Yeah, I think I'll stick with DISABLE for nwo

fantix: Oh true, yeah DISABLE then.

Issue #6605, opened 2023-12-13
Confusing `missing a type cast before the parameter` on query syntax order


jackfischer:
- EdgeDB Version: 4.2
- EdgeDB CLI Version: 4.0.2
- OS Version:

Steps to Reproduce:

Run the following incorrect query where the filter is before the shape. Obviously the issue is not a missing type cast, though because one is involved the error message is an especially confusing red herring. I fell for this as a reasonably experienced edgeql-er.

```
credalai_app_9964:edgedb> select User filter .id = <uuid>$userId { id };
error: QueryError: missing a type cast before the parameter
  ┌─ <query>:1:32
  │
1 │ select User filter .id = <uuid>$userId { id };
  │                                ^^^^^^^ error
```

Schema:
```
type User {}
```

PR #7044, opened 2024-03-14
Improve error when applying a shape to a parameter

dnwpark:
When passing the query from #6605, the error has been changed to:

```
select User filter .id = <uuid>$userId { id };
error: QueryError: cannot apply a shape to the parameter
  ┌─ <query>:1:32
  │
1 │ select User filter .id = <uuid>$userId { id };
  │                                ^^^^^^^^^^^^ Consider adding parentheses around the parameter and type cast
```

When parentheses are added:
```
error: QueryError: shapes cannot be applied to scalar type 'std::uuid'
  ┌─ <query>:1:27
  │
1 │ select User filter .id = (<uuid>$userId) { id };
  │                           ^^^^^^^^^^^^^^^^^^^ error
```

### DIFF
--- a/edb/edgeql/compiler/astutils.py
+++ b/edb/edgeql/compiler/astutils.py
@@ -21,6 +21,7 @@
 
 
 from __future__ import annotations
+from dataclasses import dataclass
 from typing import Optional, Tuple, Dict, List
 
 from edb.common import ast
@@ -121,6 +122,12 @@ def extend_path(expr: qlast.Expr, field: str) -> qlast.Path:
         return qlast.Path(steps=[expr, step])
 
 
+@dataclass
+class Params:
+    params: List[
+        Tuple[qlast.TypeCast, Dict[Optional[str], str]]] = []
+    loose_params: List[qlast.Parameter] = []
+
 class FindParams(ast.NodeVisitor):
     """Visitor to find all the parameters.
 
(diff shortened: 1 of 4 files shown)


edb/edgeql/compiler/astutils.py
```diff
@@ -121,16 +122,22 @@ def extend_path(expr: qlast.Expr, field: str) -> qlast.Path:
         return qlast.Path(steps=[expr, step])
 
 
+@dataclass
+class Params:
+    cast_params: List[
+        Tuple[qlast.TypeCast, Dict[Optional[str], str]]
+    ] = field(default_factory=list)
+    shaped_params: List[qlast.Parameter] = field(default_factory=list)
```
msullivan, line 130: Let's store the Shape here so we can report the source span over all of `$userId { id }`

edb/edgeql/compiler/stmtctx.py
```diff
@@ -818,6 +818,15 @@ def check_params(params: Dict[str, irast.Param]) -> None:
                 f'{"s" if len(missing_args) > 1 else ""}')
 
 
+def throw_on_shaped_param(
+    param: qlast.Parameter,
+    ctx: context.ContextLevel
+) -> None:
+    raise errors.QueryError(
+        f'a shape cannot be applied to the parameter',
+        context=param.context)
```
msullivan, line 827: Lets add a `hint=` explaining that probably parentheses should be added around the cast+param

msullivan (approved):
Needs a test but looks pretty good

msullivan (approved):

PR #7046, opened 2024-03-14
Fix parser at unrecoverable errors

aljazerzen:
In some cases, our parser will not be able to recover from an error.
What we want in this case is to skip all tokens until EOF.
This was working correctly, but:

1. It could be skiping tokens for up to 10 different recovery paths.
   (i.e. each token would processed 10 times).
   This maked parsing of such cases ~10x slower than it could be.
   We do have a mechanism that prevents injecting new tokens after
   error recovery cost gets too large, so I've made it so that when that
   happens, we discard all but the best recovery path.

2. Each time a parser would skip a token, it would first check all
   possible tokens to inject. It would then figure out it is not
   allowed to inject token in this case, so loop was done completely
   unnecessarily.

The pathological case from BeatGig now parses in ~400ms on my machine.

Another thing that we should probably look into, is not reporting
15000 errors of unexpected tokens, but a single error saying\
"cannot parse from here on".

### DIFF
--- a/edb/edgeql-parser/src/parser.rs
+++ b/edb/edgeql-parser/src/parser.rs
@@ -55,6 +55,7 @@ pub fn parse<'a>(input: &'a [Terminal], ctx: &'a Context) -> (Option<&'a CSTNode
     let mut new_parsers = Vec::with_capacity(parsers.len() + 5);
 
     for token in input {
+        // println!("token {:?}", token);
         while let Some(mut parser) = parsers.pop() {
             let res = parser.act(ctx, token);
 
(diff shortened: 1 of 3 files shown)


edb/edgeql-parser/src/parser.rs
```diff
@@ -55,6 +55,7 @@ pub fn parse<'a>(input: &'a [Terminal], ctx: &'a Context) -> (Option<&'a CSTNode
     let mut new_parsers = Vec::with_capacity(parsers.len() + 5);
 
     for token in input {
+        // println!("token {:?}", token);
```
msullivan, line 58: Do you want to keep the debug code?

msullivan (approved):

edb/edgeql-parser/src/parser.rs
aljazerzen: Yes, ive used it a few times now.

Maybe i should have put it behind a feature flag instead of a comment.

PR #7047, opened 2024-03-14
Optimize pickled in-tx CompilerConnectionState

fantix:
This is an optimization for transactions without DDL, especially effective on branches with heavy schemas. Until a DDL command is issued, the user schema is now excluded from the pickled in-transaction state that is calculated and transmitted after the compilation of `start transaction` and `commit`.

Thanks to @themajashurka for capturing this, and providing repro schema+script.

- [x] Fix and [test remote compiler](https://github.com/edgedb/edgedb/actions/runs/8300568883/job/22718902256)
- [x] Fix and test [multi-tenant mode](https://github.com/edgedb/edgedb/actions/runs/8300568883/job/22718902471)

### DIFF
--- a/edb/server/compiler/dbstate.py
+++ b/edb/server/compiler/dbstate.py
@@ -676,7 +676,7 @@ class TransactionState(NamedTuple):
 
     id: int
     name: Optional[str]
-    user_schema: s_schema.FlatSchema
+    local_user_schema: s_schema.FlatSchema | None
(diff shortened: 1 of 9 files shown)


msullivan (approved):
Looks good, though this is all getting pretty hairy.

msullivan:
(Let's not backport to 4.x)

PR #7050, opened 2024-03-15
Restore libpg_query cache

fantix:
Refs #6887, #7042


elprans (approved):

msullivan (changes requested):
I don't think this is right; I think that the extension key needs to be dependent on libpg_query. See https://github.com/edgedb/edgedb/pull/7053

And, we need to revert libpg_query, since the breakages are genuine and we need to update edgedb for the new version to use it

fantix:
Close in favor of #7053

PR #7052, opened 2024-03-15
Revert "Bump libpg_query (#7042)"

msullivan:
This reverts commit 2db80d9be974963a931613f2b2b00a10e196b300.

The update causes a lot of breakages of SQL related tests.  I think
that the caching in CI is faulty and failed to catch it, and our local
build system won't automatically rebuild it.


msullivan:
We'll need to do some real work to update it, unfortunately.
If we want, we can temporarily use a fork of our own where we backport just the glibc fix.

msullivan:
We'll need to do some real work to update it, unfortunately.
If we want, we can temporarily use a fork of our own where we backport just the glibc fix.

aljazerzen (approved):

PR #7053, opened 2024-03-15
Fix submodule CI cache stamps

msullivan:
Since we don't sync submodules in our checkout for build unless we
have to, we need to pass `--cached` or else the commit we get might be
whatever is lying around.

Also, add the libpg_query hash to the ext cache hash, since it is part
of that.


msullivan:
This problem is why the libpg_query upgrade didn't break things in CI.

... of course, I think this also means that this PR will fail CI until https://github.com/edgedb/edgedb/pull/7052 is merged

PR #7058, opened 2024-03-15
Docs: Add UI fixes to changelog

raddevon:
I highlighted several and then linked to a GitHub list of commits since February 25th for the rest. I tried linking each commit in parens in the last bullet item, but it was a big mess so I think this might be a better approach.

If I called out any that probably most people won't care about, or if I didn't call out a big one, happy to revise.

### DIFF
--- a/docs/changelog/4_x.rst
+++ b/docs/changelog/4_x.rst
@@ -709,6 +709,9 @@ Bug fixes
 
 * Several UI fixes
 
+  * Update behaviour of datatable editor to save input by default on closing
+    (:eql:gh:`edgedb/edgedb-ui/#325`)
+
   * Store instance version with schema data, so schema is refreshed when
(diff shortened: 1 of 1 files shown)


docs/changelog/4_x.rst
```diff
@@ -704,5 +704,30 @@ Bug fixes
...
+  * Always order ``id`` column first in dataview
+    (:eql:gh:`edgedb/edgedb-ui/9a7c352e`)
+
+  * Improve rendering of long/multiline strings in dataview changes preview
+    (:eql:gh:`edgedb/edgedb-ui/13511ebd`)
+
+  * Fix link props in visual query builder
+    (:eql:gh:`edgedb/edgedb-ui/42492465`)
+
+  * Update behaviour of datatable editor to save input by default on closing
```
jaclarke, line 730: Maybe this one should be first, as it's a fairly important behaviour change.

jaclarke (approved):

PR #7059, opened 2024-03-15
Make escaping strings more consistent.

dnwpark:


tests/test_edgeql_data_migration.py
```diff
@@ -140,7 +141,7 @@ async def fast_forward_describe_migration(
                         interpolations[var_name] = var_value
 
                 for stmt in mig['proposed']['statements']:
-                    curddl = stmt['text']
```
msullivan, line 143: We don't want to have to an add `unescape_string` here. This is testing exposed behavior; if we update this, we'd need to update our CLI's migration flows, and we'd break any user using this.

edb/edgeql/quote.py
```diff
@@ -35,17 +35,26 @@
...
+        # https://www.edgedb.com/docs/reference/edgeql/lexical#strings
+        result = s
 
-    return ''.join((r if i % 2 else r.replace(r"'", r"\'"))
-                   for i, r in enumerate(split))
+        # escape backslash first
+        result = result.replace('\\', '\\\\')
 
+        result = result.replace('\'', '\\\'')
+        result = result.replace('\"', '\\\"')
```
msullivan, line 49: I'd prefer not to escape double quotes

edb/edgeql/quote.py
```diff
@@ -35,17 +35,26 @@
...
+        result = result.replace('\b', '\\b')
+        result = result.replace('\f', '\\f')
+        result = result.replace('\n', '\\n')
+        result = result.replace('\r', '\\r')
+        result = result.replace('\t', '\\t')
```
msullivan, line 54: Should we make this table driven?
(Or maybe just not bother)

msullivan (approved):

msullivan (approved):
Failure's just in one of your new tests

PR #7062, opened 2024-03-15
Fix issues with cached global shapes and global cardinality inference

### LINKED ISSUE #7056: Initial migration fails after upgrading to v4.6
- EdgeDB Version: 4.6+0632dff
- EdgeDB CLI Version: 4.1.0+03ae624
- OS Version: Linux Mint 21.3

After I reset my dev database instance, and upgrade to v4.6, I get a cryptic error that suggests a bug was introduced in 4.6 (everything works fine in 4.5).

Steps to Reproduce:

1. Run `edgedb project init` with `server-version="4.6'` in `edgedb.toml`
2. When applying initial migrations, the following error is thrown:
```
edgedb error: InternalServerError: relation "edgedbpub.19b9270f-e2cf-11ee-86bc-774ffaf8bef2_t" does not exist
  Server traceback:
      edb.errors.InternalServerError: relation "edgedbpub.19b9270f-e2cf-11ee-86bc-774ffaf8bef2_t" does not exist
```

Schema:

The schema definition is too large to be provided in the issue body, but I can upload it somewhere if that helps in any way.

### LINKED ISSUE #7061: Single globals no longer correctly infer optional return type
Regression in 4.6 and 5.x caused by https://github.com/edgedb/edgedb/pull/6613

```
_localdev:main> create global y := (select Object filter false limit 1);
OK: CREATE GLOBAL
_localdev:main> select { required y := global y };
[{"y": null}]
```

Originally via https://discord.com/channels/841451783728529451/1217949944800088105

msullivan:
Having an actual shape on a cached global caused confusion with the
convoluted logic used to determine whether something was a cached
global in range_for_material_objtype.
Instead, track it directly in TypeRoot.

Also, properly compile the type_rewrite for cached globals in a
detached scope so that the cardinality inference for it doesn't
collide with the enclosing scope.

Fixes #7056. Fixes #7061.

### DIFF
--- a/edb/ir/ast.py
+++ b/edb/ir/ast.py
@@ -511,6 +511,9 @@ class TypeRoot(Expr):
     # This will be replicated in the enclosing set.
     typeref: TypeRef
 
+    # Whether this is a reference to a cached global.
+    is_cached_global: bool = False
+
     # Whether to force this to not select subtypes
     skip_subtypes: bool = False
 
--- a/edb/ir/ast.py
+++ b/edb/ir/ast.py
@@ -511,7 +511,8 @@ class TypeRoot(Expr):
     # This will be replicated in the enclosing set.
     typeref: TypeRef
 
-    # Whether this is a reference to a cached global.
+    # Whether this is a reference to a global that is cached in a
+    # materialized CTE in the query.
     is_cached_global: bool = False
 
     # Whether to force this to not select subtypes
--- a/edb/pgsql/compiler/relctx.py
+++ b/edb/pgsql/compiler/relctx.py
@@ -505,7 +505,8 @@ def new_primitive_rvar(
         skip_subtypes = ir_set.expr.skip_subtypes
         is_global = ir_set.expr.is_cached_global
     else:
-        skip_subtypes = is_global = False
+        skip_subtypes = False
+        is_global = False
 
     typeref = ir_set.typeref
     dml_source = irutils.get_dml_sources(ir_set)
--- a/edb/pgsql/compiler/relctx.py
+++ b/edb/pgsql/compiler/relctx.py
@@ -503,15 +503,19 @@ def new_primitive_rvar(
 ) -> pgast.PathRangeVar:
     if isinstance(ir_set.expr, irast.TypeRoot):
         skip_subtypes = ir_set.expr.skip_subtypes
+        is_global = ir_set.expr.is_cached_global
     else:
-        skip_subtypes = False
+        skip_subtypes = is_global = False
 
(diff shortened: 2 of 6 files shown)


edb/pgsql/compiler/relctx.py
```diff
@@ -503,15 +503,19 @@ def new_primitive_rvar(
 ) -> pgast.PathRangeVar:
     if isinstance(ir_set.expr, irast.TypeRoot):
         skip_subtypes = ir_set.expr.skip_subtypes
+        is_global = ir_set.expr.is_cached_global
     else:
-        skip_subtypes = False
+        skip_subtypes = is_global = False
```
aljazerzen, line 508:
```suggestion
        skip_subtypes = False
        is_global = False
```

edb/pgsql/compiler/relctx.py
```diff
@@ -1413,20 +1417,15 @@ def range_for_material_objtype(
...
 
     env = ctx.env
 
-    # If this is a view type, but it still appears in rewrites, that is
-    # because it is a global that we are caching.
-    if (
-        typeref.material_type is not None
-        and (typeref.id, include_descendants) not in ctx.env.type_rewrites
-    ):
-        typeref = typeref.material_type
```
aljazerzen, line 1428: This doesn't work because the shape on the global causes a new derived type, which will not appear in `type_rewrites`, right?

edb/ir/ast.py
```diff
@@ -511,6 +511,9 @@ class TypeRoot(Expr):
     # This will be replicated in the enclosing set.
     typeref: TypeRef
 
+    # Whether this is a reference to a cached global.
+    is_cached_global: bool = False
```
aljazerzen, line 515: Cached global?

Do we actually cache the value, or is this just your way of saying "a global with an assigned expression whose value we compute at runtime"?

aljazerzen (approved):

edb/pgsql/compiler/relctx.py
msullivan: It was sillier than that, I think. When there was a shape, the derived type appeared *inside* the `type_rewrite` also, and so it got picked up as being a global (since it *was* in type_rewrites), but it didn't get compiled that way because we were already compiling it.
Instead of trying to tease out that case, I made the mechanism more explicit.

edb/ir/ast.py
aljazerzen: Oh, cached in the query, right.

I though it was cached in edgedb connection or even edgedb database.

PR #7064, opened 2024-03-15
Simplify compilation of TypeIntrospection

msullivan:
The currently implementation is needlessly verbose, since it provides
the source aspect for schema::Type. We can avoid all that work and
just directly output the id value from a `compile` function, and we'll
automatically join it against a source aspect if needed.

### DIFF
--- a/edb/pgsql/compiler/expr.py
+++ b/edb/pgsql/compiler/expr.py
@@ -350,14 +350,6 @@ def _inline_array_slicing(
     )
 
 
-@dispatch.compile.register(irast.TypeIntrospection)
-def compile_TypeIntrospection(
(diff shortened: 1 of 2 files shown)


aljazerzen:
> The currently implementation is needlessly verbose, since it provides
the source aspect for schema::Type

You mean it provides the source aspect for `X::__type__`?

msullivan:
> > The currently implementation is needlessly verbose, since it provides
> > the source aspect for schema::Type
> 
> You mean it provides the source aspect for `X::__type__`?

Well, a source aspect for whatever the introspect expression path id is. But the table it selects from is `schema::Type`

PR #7071, opened 2024-03-18
Refactor naming of parser source span

### LINKED ISSUE #6968: Rename ParserContext to Span and make sure it is exported from all the `ast.py` modules
`Span` is a nicer name for source spans and there is no reason we should need to import another module to get the type name.

We should then start renaming the attributes that hold them to `span`  from `context`/`srcctx`

aljazerzen:
Closes #6968

As a references, this used to be named:
- `parser_context`,
- `source_context`,
- `context`,
- `ParsingContext`,
- `pctx`,
- `span`.

This PR does have an enormous diff, it cannot be backported to old branches, but it probably should be backported to 5.x, so new fixed don't cause too much conflicts.

### DIFF
--- a/edb/common/context.py
+++ b/edb/common/context.py
@@ -46,7 +46,7 @@ from edb.common import typeutils
 NEW_LINE = re.compile(br'\r\n?|\n')
 
 
-class ParserContext(markup.MarkupExceptionContext):
+class Span(markup.MarkupExceptionContext):
     title = 'Source Context'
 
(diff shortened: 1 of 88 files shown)


aljazerzen:
It looks like I've missed a bunch of tests (since I only ran "#fasttests" locally). And now I want to first make the thing that collects all of the failed tests into the "collect" step of GHA workflow.

msullivan (approved):
Very happy about this

PR #7072, opened 2024-03-18
Restore cheatsheets to side navigation

raddevon:
Discovered this problem this morning. Position of the TOC in the file now determines what gets rendered in the navigation sidebar, which resulted in the individual cheatsheet links not being rendered there. This was further complicated by the fact that the booleans cheatsheet link at Cheatsheets > Overview was set to the incorrect target, making that cheatsheet difficult to access except via search.

### DIFF
--- a/docs/guides/cheatsheet/index.rst
+++ b/docs/guides/cheatsheet/index.rst
@@ -45,8 +45,8 @@ EdgeQL
 Schema
 ======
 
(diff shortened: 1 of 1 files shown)


scotttrinh (approved):
I don't know anything about the docs implementation here, so I'd treat my approval as a bit of a rubber stamp just saying I read through the code and it seems reasonable, but can't speak to the underlying issue of the `toctree` placement.

PR #7073, opened 2024-03-18
Fix update rewrites on types that are children of updated type

### LINKED ISSUE #6801: `MissingRequiredError` in mutation rewrite when updating base type
- EdgeDB Version: 4.5+28216a1
- EdgeDB CLI Version: 4.0.2+be1da73
- OS Version: GNU/Linux 6.7.3-arch1-1

```esdl
abstract type A {
   abstract_value: str;
}

type AA extending A {
   value: str;

   required value_changed_at: datetime {
      default := to_datetime(0);
      rewrite update using (
         with 
            j := <json>__specified__,
         select(
            if <bool>json_get(j, "value")
            then datetime_current()
            else __subject__.value_changed_at
         )
      );
   };
}
```

```edgeql
insert AA { value := "aa" };
update A set { abstract_value := "f" }

// edgedb error: MissingRequiredError: missing value for required property 'value_changed_at' of object type 'default::AA'

### LINKED ISSUE #7048: Rewrites are wrongfully applied to parent types
Duplicate of https://github.com/edgedb/edgedb/issues/6801

- EdgeDB Version: 5.0-prerelease

Steps to Reproduce:

```
> insert Document { actual_type := 'Document' };
OK

> insert Account { actual_type := 'Account' };
OK

> update Document filter .actual_type = 'Document' set { ingestionStatus := 'Pending' };
OK

> update Account filter .actual_type = 'Account' set { ingestionStatus := 'Pending' };
edgedb error: InvalidReferenceError: object type 'default::Account' has no link or property 'ingestionStatus'

> update std::Object set { };
edgedb error: InvalidReferenceError: ingestionStatus is not a member of tuple<id: std::bool, __type__: std::bool>
```

Schema:

```
module default {
    abstract type Entity {
        actual_type -> str;
    }

    type Document extending Entity {
        required property ingestionStatus -> str {
            default := 'Ingested';
        }

        REQUIRED PROPERTY statusUpdatedAt: std::datetime {
            default := (std::datetime_of_statement());
            REWRITE
                UPDATE 
                USING ((std::datetime_of_statement() IF __specified__.ingestionStatus ELSE __old__.statusUpdatedAt));
        };
 

    }

    type Account extending Entity;
}
```

### PR DESCRIPTION
**Pull request by aljazerzen**
Closes #7048
Closes #6801

When type X is updated, but has child Y with an update rewrite, we were using type X as the subject where we should have been using Y.

Review commit-by-commit as I had to do a little refactor first.

### DIFF
--- a/edb/edgeql/compiler/viewgen.py
+++ b/edb/edgeql/compiler/viewgen.py
@@ -907,9 +907,6 @@ class RewriteContext:
     base_type: s_objtypes.ObjectType
     shape_type: s_objtypes.ObjectType
 
-    # Cache for computing `prepare_rewrite_anchors`, since that is expensive
-    anchor_cache: Dict[s_objtypes.ObjectType, RewriteAnchors]
-
 
 def _compile_rewrites(
     specified_ptrs: Set[sn.UnqualName],
@@ -926,14 +923,12 @@ def _compile_rewrites(
         kind=kind,
         base_type=stype,
         shape_type=view_scls,
-        anchor_cache=dict(),
     )
 
     # Computing anchors isn't cheap, so we want to only do it once,
     # and only do it when it is necessary.
     anchors: Dict[s_objtypes.ObjectType, RewriteAnchors] = {}
     def get_anchors(stype: s_objtypes.ObjectType) -> RewriteAnchors:
-        nonlocal anchors
         if stype not in anchors:
             anchors[stype] = prepare_rewrite_anchors(stype, r_ctx, s_ctx, ctx)
         return anchors[stype]
--- a/edb/edgeql/compiler/viewgen.py
+++ b/edb/edgeql/compiler/viewgen.py
@@ -929,14 +929,14 @@ def _compile_rewrites(
         anchor_cache=dict(),
     )
 
-    anchors: RewriteAnchors | None = None
-
     # Computing anchors isn't cheap, so we want to only do it once,
     # and only do it when it is necessary.
+    anchors: Dict[s_objtypes.ObjectType, RewriteAnchors] = {}
     def get_anchors(stype: s_objtypes.ObjectType) -> RewriteAnchors:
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: edb/edgeql/compiler/viewgen.py [hunk 1]
```diff
@@ -996,12 +996,13 @@ def get_anchors(stype: s_objtypes.ObjectType) -> RewriteAnchors:
 
             by_type[ty][pn] = (ptr_set, ptrref.real_material_ptr)
 
-    if not anchors:
+    anc = next(iter(anchors.values()), None)
```
msullivan, line 999: Is the first one always the one we want?

edb/edgeql/compiler/viewgen.py
```diff
@@ -899,6 +899,18 @@ def _raise_on_missing(
...
+@dataclasses.dataclass(kw_only=True, repr=False, eq=False)
+class RewriteContext:
+    specified_ptrs: Set[sn.UnqualName]
+    kind: qltypes.RewriteKind
+
+    base_type: s_objtypes.ObjectType
+    shape_type: s_objtypes.ObjectType
+
+    # Cache for computing `prepare_rewrite_anchors`, since that is expensive
+    anchor_cache: Dict[s_objtypes.ObjectType, RewriteAnchors]
```
msullivan, line 911: This seems unused?

msullivan (approved):

edb/edgeql/compiler/viewgen.py
aljazerzen: We want any of them, I think it doesn't matter

edb/edgeql/compiler/viewgen.py
aljazerzen: Oh, yeah, I wanted to used this instead of a local anchors variable, but forgot about it.

It is not really any better, so your commit is fine.

PR #7082, opened 2024-03-19
Separate triggers by kind in compilation.

### LINKED ISSUE #5794: [Triggers] False positive recursive trigger
- EdgeDB Version: "3.1+a33a828"
- EdgeDB CLI Version: `EdgeDB CLI 3.4.0+4d95a2c`
- OS Version: MacOS

Steps to Reproduce:

1. Migrate
2. see the error

The trigger should happen on update only, and only inserts an object. This wouldn't trigger another update.
```
error: trigger on default::Theme is recursive
```

Schema:
```
abstract type WithRevisions {}
	type Revision {
		required object: WithRevisions;
	}
type Theme extending WithRevisions {
	        required name: str;
		trigger revision_create after update for each do (
			insert Revision {
				object := (
					insert Theme {
						name := __old__.name
					}
				)
			}
		);
	}
```

dnwpark:
Fix #5794

When checking for recursive triggers, the compilation now checks `insert`, `update`, and `delete` separately.

Given the schema in #5794, running the following:
```
insert Theme {name := "hello"};
update Theme filter Theme.name = "hello" set {name := "world
"};
select Theme {*}
```

Produces:
```
{
  default::Theme {id: c893e790-e645-11ee-a4e6-a74c0e50141c, name: 'world'},
  default::Theme {id: f4f70b0a-e645-11ee-a4e6-c360cb91f3c6, name: 'hello'},
}
```

As another example, this schema is now valid but becomes invalid if `update` is changed to `insert`
```  type A {
    required name: str;
    trigger create_A after update for each do (
      insert B {
        name := __new__.name
      }
    );
  }

  type B {
    required name: str;
    trigger create_B after insert for each do (
      insert A {
        name := __new__.name
      }
    );
  }
```

### DIFF
--- a/edb/edgeql/compiler/triggers.py
+++ b/edb/edgeql/compiler/triggers.py
@@ -138,6 +138,7 @@ def compile_trigger(
 def compile_triggers_phase(
     dml_stmts: Collection[irast.MutatingStmt],
     defining_trigger_on: Optional[s_types.Type],
+    defining_trigger_kinds: Optional[Collection[qltypes.TriggerKind]],
     *,
     ctx: context.ContextLevel,
 ) -> tuple[irast.Trigger, ...]:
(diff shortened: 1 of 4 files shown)


edb/schema/triggers.py
```diff
@@ -204,27 +204,36 @@ def compile_expr_field(
...
+            anchors: dict[
+                str, pathid.PathId | s_types.Type | qlast.Expr
+            ] = {k: v for k, v in singleton_anchors.items()}
             anchors['__trigger_type__'] = source
+            anchors['__trigger_kinds__'] = qlast.Array(
+                elements=[
+                    qlast.StringConstant.from_python(kind)
+                    for kind in kinds
+                ]
+            )
```
msullivan, line 236: I think we probably should add a new field to Options rather than smuggle it in through anchors

(Though smuggling it in via an array literal *is* clever)

edb/edgeql/compiler/triggers.py
```diff
@@ -165,11 +167,13 @@ def compile_triggers_phase(
 
         # Process all the types, starting with the base type
         for subtype in sorted(stypes, key=lambda t: t != stype):
-            if defining_trigger_on and subtype.issubclass(
-                    ctx.env.schema, defining_trigger_on):
+            if (defining_trigger_on and defining_trigger_kinds and
```
msullivan, line 170: Typically we prefer putting the `and` on the continuation line

PR #7083, opened 2024-03-19
Skip `test_server_auth_mtls` on MacOS.

vpetrovykh:
PostgreSQL's request for a shared memory segment exceeded your kernel's SHMALL parameter causing an error in out builds.

### DIFF
--- a/tests/test_server_auth.py
+++ b/tests/test_server_auth.py
@@ -19,6 +19,7 @@
 import asyncio
 import os
 import pathlib
+import platform
 import signal
 import ssl
 import tempfile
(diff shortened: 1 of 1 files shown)


msullivan (approved):

tests/test_server_auth.py
```diff
@@ -581,6 +581,11 @@ async def test_server_auth_in_transaction(self):
                 DROP ROLE foo;
             ''')
 
+    @unittest.skipIf(
+        platform.system() == "Darwin" and platform.machine() == 'x86_64',
+        "Postgres is not getting getting enough shared memory on macOS "
```
fantix, line 586: > on macOS

Probably on the `macos-14` GitHub runner? This test passed on my local x86 macOS 14.

fantix (approved):

tests/test_server_auth.py
vpetrovykh: I've updated the message.

PR #7090, opened 2024-03-21
Merge Pointer into expr field

msullivan:
Parts of this migration are pretty half-assed, and use compatibility properties `rptr` and `old_expr`. Aljaž and I will burn those down in the next couple weeks.


msullivan:
@aljazerzen and I decided to merge ASAP and then do the cleanups

aljazerzen (approved):

PR #7094, opened 2024-03-21
Doc v5 branches guide

raddevon:
I won't guarantee this is everything in the "Guides" section that needs to change. (It likely isn't.) I do think this handles the critical stuff though and does most of what we need to surface the feature development + branching workflow.

I'll start next on combing the docs for mentions of database and reworking as necessary. That should allow me to pick up any stragglers I've missed here.


docs/cli/edgedb_branch/edgedb_branch_rebase.rst
```diff
@@ -26,6 +26,9 @@ Description
 Creates a new branch that is based on the target branch, but also contains any new migrations on the
```
diksipav, line 26: maybe we should provide here a more info about how rebasing works, sth like: Creates a new branch that is based on the target branch, and also contains new migrations from the current branch. Drops the current branch and rename new branch to the current branch name.  (this is not the best but I needed a bit more info when working with rebasing)

vpetrovykh:
I think that the `note` for commands like `edgedb branch merge` or `edgedb branch rebase` should not just say that you should be using the `database` command suite in earlier versions because there are no equivalent commands.

Creating, dropping, and renaming are all fine because they have equivalents in the `database` suite (even if creating options are different), but there more complex `branch` commands are very much new and only exist as branch commands.

You can compare the command suites in the CLI to make sure that the older versions have equivalent things.

vpetrovykh:
Also I think you use both `git` and VCS in your descriptions. In any particular section you probably don't want to mix the terminology. If the language is generic enough to use VCS instead of `git`, go for it. If the examples rely on `git`, that's also fine, but perhaps then just stick to that in the description maybe acknowledging that other VCS should work in some equivalent manner.

raddevon:
@vpetrovykh I circled back on the commands that do not have database command analogs to add that in those notes, and while I was there, I figured it made sense to expose the alternatives where there are alternatives that make sense (like `rename`). While I was doing that, I decided it would make sense to just make each of the command notes point to the particular alternative when there is a clean analog (`create`, `drop`, `wipe`). This ended up making some notes very long, so I moved them after the initial command descriptions, except for the note on the `edgedb branch` overview page which I left at the top since it is short and since that page is just a table of links.

> Also I think you use both git and VCS in your descriptions. In any particular section you probably don't want to mix the terminology. If the language is generic enough to use VCS instead of git, go for it. If the examples rely on git, that's also fine, but perhaps then just stick to that in the description maybe acknowledging that other VCS should work in some equivalent manner.

I only found instances of this in the "Getting started" section, so point out any others if I've missed them. I circled back to `intro/branches.rst`, and I think it's important to talk about it generically as a way to bridge the gap between EdgeDB schema development and application development with a VCS rather than to make it appear as though it's somehow married to Git. At the same time, the examples need to be concrete to get readers from 0 to 1 as quickly as we can.

I can see your concern about the transition between those two being jarring though. I don't think consistency is necessarily the answer though because we risk either implying that the feature is married to Git or leaving the users with abstract and difficult-to-follow examples. Instead, I've added a note near the beginning to tell users that the workflow is compatible with any VCS that offers branches and rebasing but that we will show examples in Git to make them concrete. I've left the rest of the guide as it was. Does that approach make sense, and does it adequately address the concern?

vpetrovykh (approved):
Looks good now :)

PR #7106, opened 2024-03-22
Fix fts::search volatility

aljazerzen:


aljazerzen:
Backporting this requires a "schema" patch.

PR #7111, opened 2024-03-25
Refactor irast.Set.rptr into Set.expr  (part 1 / n)

aljazerzen:
Followup for #7090

- indexes.py
- multiplicity.py
- drop unsed param
- cardinality 1
- cardinality 2 & 3
- viewgen 1
- viewgen 2
- refactor qlcompiler  stmt.py
- ql compiler stmt.py

### DIFF
--- a/edb/edgeql/compiler/conflicts.py
+++ b/edb/edgeql/compiler/conflicts.py
@@ -494,20 +494,20 @@ def compile_insert_unless_conflict_on(
 
     # We accept a property, link, or a list of them in the form of a
     # tuple.
-    if cspec_res.rptr is None and isinstance(cspec_res.expr, irast.Tuple):
+    if isinstance(cspec_res.expr, irast.Tuple):
         cspec_args = [elem.val for elem in cspec_res.expr.elements]
     else:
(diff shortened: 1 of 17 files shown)


edb/edgeql/compiler/conflicts.py
```diff
@@ -593,8 +593,8 @@ def compile_insert_unless_conflict_on(
 
 def _has_explicit_id_write(stmt: irast.MutatingStmt) -> bool:
     for elem, _ in stmt.subject.shape:
-        assert elem.rptr is not None
-        if elem.rptr.ptrref.shortname.name == 'id':
+        assert isinstance(elem.expr, irast.Pointer)
```
msullivan, line 596: This is an aside, and not a change request for this PR, but we should be able to make `shape` contain `SetE[Pointer]`

msullivan (approved):
Great work on this.
Very happy about this refactor.

edb/edgeql/compiler/conflicts.py
msullivan: Oh wait, that's already done and we can drop this assert

PR #7114, opened 2024-03-25
Refactor Set.rptr into Set.expr (2 / 2)

msullivan:
Get rid of `rptr` and `old_expr`. (`old_expr` becomes
`irutils.sub_expr` and gets used in some call sites)

PR #7116, opened 2024-03-26
Make expr non-Optional in irast.Set

msullivan:
The bulk of the work to make this happen was done in #6938
(introducing TypeRoot) and #7090 (moving Pointer into expr).

The final pieces are making EmptySet an Expr instead of a Set subclass
and getting rid of the cases where we "clear" expressions out of a
Set.

The biggest benefit here is that the meaning of a Set is much clearer
now, but there are some nice direct benefits too.

This lets have fully table dispatched set compilation in several places
now.
`_get_set_rvar`, which was 155 lines long in 2.x, is now finally completely
gone!
`_compile_set_in_singleton_mode` is also gone, replaced with simply 
compiling the set's expression.

Typechecker error messages involving SetE are also much nicer now.

### DIFF
--- a/edb/ir/ast.py
+++ b/edb/ir/ast.py
@@ -792,7 +792,7 @@ class ConstExpr(Expr):
     typeref: TypeRef
 
 
-class EmptySet(Set, ConstExpr):
+class EmptySetExpr(ConstExpr):
     pass
 
 
--- a/edb/ir/ast.py
+++ b/edb/ir/ast.py
@@ -792,7 +792,7 @@ class ConstExpr(Expr):
     typeref: TypeRef
 
 
-class EmptySetExpr(ConstExpr):
+class EmptySet(ConstExpr):
     pass
 
 
--- a/edb/ir/ast.py
+++ b/edb/ir/ast.py
@@ -530,14 +530,28 @@ class TypeRoot(Expr):
     skip_subtypes: bool = False
 
 
-T_co = typing.TypeVar('T_co', covariant=True, bound=typing.Optional[Expr])
+class RefExpr(Expr):
+    '''Different expressions sorts that refer to some kind of binding.'''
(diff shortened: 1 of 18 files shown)


elprans (approved):

msullivan:
Depends on https://github.com/edgedb/edgedb/pull/7114

edb/ir/ast.py
```diff
@@ -530,14 +530,28 @@ class TypeRoot(Expr):
...
+    __abstract_node__ = True
+    typeref: TypeRef
+
+
+class MaterializedExpr(RefExpr):
+    pass
+
+
+class VisibleBindingExpr(RefExpr):
+    pass
```
aljazerzen, line 544: So before we had just `Set.expr = None` and now we have separate subclasses of Expr to be more explicit?

edb/pgsql/compiler/relgen.py
```diff
@@ -1493,6 +1493,7 @@ def process_set_as_subquery(
 
 
 register_get_rvar(irast.Stmt)(process_set_as_subquery)
+register_get_rvar(irast.MaterializedExpr)(process_set_as_subquery)
```
aljazerzen, line 1496: Why cannot we declare these two as decorators?

aljazerzen (approved):

edb/ir/ast.py
msullivan: Yeah

edb/pgsql/compiler/relgen.py
msullivan: "Because mypy doesn't support higher-kinded polymorphism"

It would narrow the type of the function, and this function gets called directly, and because of not having the higher-kinded polymorphism we can't give the decorator the type we'd need to have it work the way we want and also enforce the correct type for the function.

PR #7117, opened 2024-03-26
Don't pretty print SQL text by default

msullivan:
Only do it during the debug endpoints. I think this changed by
accident at some point.  This shouldn't matter *too* much, but will
reduce the size of the SQL sent to postgres.


elprans (approved):

aljazerzen (commented):
This one is on me (probably).

PR #7118, opened 2024-03-26
Add a debug flag to inject the edgeql as a comment in the SQL

msullivan:
This should make it easier to extract information from postgres logs
in certain situations.

Right now it is a debug flag named EDGEDB_DEBUG_PRESERVE_QUERIES.
Other options would be to do it unconditionally or to have a config
value.
Taking suggestions on names.

It injects the pre-constant extraction source.

### DIFF
--- a/edb/common/debug.py
+++ b/edb/common/debug.py
@@ -144,6 +144,9 @@ class flags(metaclass=FlagsMeta):
     server_clobber_pg_conns = Flag(
         doc="Discard Postgres connections when releasing them to the pool.")
 
+    preserve_queries = Flag(
+        doc="Include the EdgeQL query text in the SQL sent to postgres.")
+
     print_locals = Flag(
         doc="Include values of local variables in tracebacks.")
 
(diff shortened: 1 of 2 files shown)


fantix (approved):
Looks good and I remember I wanted this some time ago too

edb/server/compiler/compiler.py
```diff
@@ -1880,6 +1881,9 @@ def _compile_ql_query(
         return dbstate.NullQuery()
 
     sql_text = pg_codegen.generate_source(sql_ast)
+    # If requested, embed the EdgeQL text in the SQL.
+    if debug.flags.preserve_queries and source:
+        sql_text += '\n-- ' + repr(source.text())
```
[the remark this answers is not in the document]
elprans: Oh, hm, OK.

edb/common/debug.py
```diff
@@ -144,6 +144,9 @@ class flags(metaclass=FlagsMeta):
     server_clobber_pg_conns = Flag(
         doc="Discard Postgres connections when releasing them to the pool.")
 
+    preserve_queries = Flag(
```
elprans, line 147:
```suggestion
    edgeql_text_in_sql = Flag(
```

msullivan: Great

PR #7121, opened 2024-03-26
Refactor func defs

aljazerzen:
`s/def /def  /g`

`darker -r HEAD..:WORKTREE: -S .`


edb/common/ast/transformer.py
```diff
@@ -39,7 +39,7 @@ class NodeTransformer(visitor.NodeVisitor):
 
        class RewriteName(NodeTransformer):
 
-           def visit_Name(self, node):
```
msullivan, line 42: lol.

but also we should delete this file

edb/common/value_dispatch.py
```diff
@@ -52,23 +51,23 @@ def value_dispatch(func: Callable[..., _T]) -> _ValueDispatchCallable[_T]:
...
           return f"I don't want a {fruit}..."
 
       @eat.register('apple')
-      def _eat_apple(fruit):
+      def  _eat_apple(fruit):
           return "I love apples!"
 
       @eat.register('eggplant')
       @eat.register('squash')
-      def _eat_what(fruit):
```
msullivan, line 64: Fix docs

msullivan (approved):
I like probably 80% of the changes made to function definition blocks and probably less than 50% of the changes that got made to code that happened to be at the tops of functions.

Typically the `def` changes I don't like are ones where it was already one-line-per-argument but black collapsed it because there wasn't a trailing comma and there was space to put it on one line. There's no probably mechanical way to detect the ones where I wish we had put the trailing comma in instead, though, so we can just live with it for now.

If there's an easy way to convince darker/black to really only look at the `def` blocks, and not reformat the first three lines of every function, then it would be nice to do that, otherwise whatever

aljazerzen:
I've reverted all of the non-func-def code changes and also a few changes to func defs that we minimal and pedantic.

---

With this PR I abandon my quest for using formatters on this codebase, since their benefits really kick-in only after the whole codebase is being formatted on every commit. The chances of adopting such workflow are minimal, so there is no point making effort toward this on every commit. Fiddling with PRs like one, I'm mostly just wasting time on a Sisyphus's task.

I do still believe that we are missing a lot of ease-of-use since we do all formatting by-hand, while having less-readable commit diffs than what we'd have, if we would be removing personal coding styles out of the commited code.

PR #7122, opened 2024-03-26
Add cfg::Config._query_cache_mode

### LINKED ISSUE #7102: Add a temporary config to switch query cache mode
Before the end goal of proper persistent query cache is achieved with Postgres functions, we'd like to have a config to control the cache mode the server is running in, that is `Config._query_cache_mode`, which is an enum of:

* `IN_MEMORY` - query cache lives in memory and resets on server restart, no persistent cache is used
* `REG_INLINE` - query cache (the actual SQL) is placed inline in the cache registration table
* `PG_FUNC` - the actual SQL is encapsulated in Postgres functions and the cache registration table contains function calls
* `DEFAULT` - the distribution-default mode, currently `REG_INLINE` but it will be `PG_FUNC` in a future release

This will be a system config and server restart is needed for changes to be effective. If the effective config value changes on start, the cache registration table will be cleared.

This config is not permanent and will be dropped once the persistent cache system is stable.

fantix:
Query cache mode is a system config. Changes are only effective after server restart. If changed, the persistent query cache will be reset on server start.

Fixes #7102

### DIFF
--- a/edb/buildmeta.py
+++ b/edb/buildmeta.py
@@ -55,7 +55,7 @@ from edb.common import verutils
 
 
 # Increment this whenever the database layout or stdlib changes.
-EDGEDB_CATALOG_VERSION = 2024_03_06_00_00
+EDGEDB_CATALOG_VERSION = 2024_03_26_00_00
 EDGEDB_MAJOR_VERSION = 6
 
 
--- a/edb/common/debug.py
+++ b/edb/common/debug.py
@@ -35,7 +35,6 @@ from __future__ import annotations
 
 import builtins
 import contextlib
-import platform
 import os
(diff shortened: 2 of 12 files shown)


msullivan:
Does it have to be a system config? That will mean that it can't be configured in multi-tenant mode, doesn't it?

fantix:
We can use env vars (#5266) to set "static" system config in multi-tenant mode.

msullivan:
> We can use env vars (#5266) to set "static" system config in multi-tenant mode.

My point is that we want *users* to be able to configure it, right?

fantix:
I see, I'll start a new PR.

PR #7123, opened 2024-03-27
Add error when enum length exceeds 63

### LINKED ISSUE #6186: InternalServerError on long enum labels
- EdgeDB Version: '3.1+1f030a7'
- EdgeDB CLI Version: EdgeDB CLI 3.5.0+907ff37
- OS Version: Ubuntu 20.04.5 LTS (Focal Fossa)

Steps to Reproduce:

1. Declare a scalar type extending enum which contains a value that is 65 characters long
2. Create migration
3. Apply migration
```
 `edgedb error: InternalServerError: invalid enum label`

### PR DESCRIPTION
**Pull request by dnwpark**
Works for both SDL and DDL.

Note: it is possible to change this max length in a custom postgres build.

Close #6186

### DIFF
--- a/edb/pgsql/common.py
+++ b/edb/pgsql/common.py
@@ -47,6 +47,12 @@ from edb.pgsql import ast as pgast
 from . import keywords as pg_keywords
 
 
+# This is a postgres limitation.
+# Note that this can be overridden in custom builds.
+# https://www.postgresql.org/docs/current/datatype-enum.html
+max_enum_label_length = 63
+
+
 def quote_e_literal(string: str) -> str:
     def escape_sq(s):
         split = re.split(r"(\n|\\\\|\\')", s)
--- a/edb/pgsql/common.py
+++ b/edb/pgsql/common.py
@@ -50,7 +50,7 @@ from . import keywords as pg_keywords
 # This is a postgres limitation.
 # Note that this can be overridden in custom builds.
 # https://www.postgresql.org/docs/current/datatype-enum.html
-max_enum_label_length = 63
+MAX_ENUM_LABEL_LENGTH = 63
 
 
 def quote_e_literal(string: str) -> str:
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by scotttrinh** [comment]
(Apologies for the drive-by review) Might make sense to add a note about this restriction in the docs, too. Maybe somewhere around: https://github.com/edgedb/edgedb/blob/ac9264a2381592927bdbbb05cce7dbdb6309e276/docs/stdlib/enum.rst?plain=1#L44-L52

**Review by aljazerzen — APPROVED** [review]

File: edb/schema/utils.py [hunk 1]
```diff
@@ -208,26 +208,49 @@ def ast_to_type_shell(
...
                     )
-                elements.append(est.maintype.name)
-            return s_scalars.AnonymousEnumTypeShell(  # type: ignore
-                elements=elements
-            )
+                elements.append(subtype_type_name.maintype.name)
+                element_spans.append(subtype_type_name.maintype.span)
+
+        for element, element_span in zip(elements, element_spans):
+            if len(element) > 63:
```
msullivan, line 242: Let's put this as a constant somewhere instead of embedding it here.

dnwpark: @msullivan is there a good place for these kinds of constants atm?

msullivan: Not really. Maybe `pgsql.common`, I guess.

edb/pgsql/common.py
```diff
@@ -47,6 +47,12 @@
 from . import keywords as pg_keywords
 
 
+# This is a postgres limitation.
+# Note that this can be overridden in custom builds.
+# https://www.postgresql.org/docs/current/datatype-enum.html
+max_enum_label_length = 63
```
msullivan, line 53: Python constants are traditional `ALL_CAPS`

msullivan (approved):

PR #7126, opened 2024-03-27
Proposal: eliminate `id` on `FreeObject`

msullivan:
We do this by making FreeObject not extend BaseObject.

FreeObject ids are currently generated; this is not particularly
useful and requires some machinations to make work correctly.

Having FreeObject not have ids makes them more sensible, I think,
especially once we are going to support taking them as input.

This breaks compatibility but probably not in a way many people depend
on.

It might require some fixes to bindings. I fixed the python bindings a
while ago in prep for
this. (https://github.com/edgedb/edgedb-python/pull/464)

### DIFF
--- a/edb/pgsql/compiler/shapecomp.py
+++ b/edb/pgsql/compiler/shapecomp.py
@@ -114,7 +114,7 @@ def compile_shape(
             assert isinstance(tuple_el, pgast.TupleElement)
             elements.append(tuple_el)
 
-        # OH HO HO
+        # If there wasn't an id (because its a FreeObject), add a fake one.
         if ctx.materializing and not has_id:
             elements.append(pgast.TupleElement(
                 path_id=ir_set.path_id,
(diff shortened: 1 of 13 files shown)


edb/pgsql/compiler/shapecomp.py
```diff
@@ -112,4 +114,11 @@ def compile_shape(
             assert isinstance(tuple_el, pgast.TupleElement)
             elements.append(tuple_el)
 
+        # OH HO HO
```
aljazerzen, line 117: Oh ho ho?

aljazerzen (approved):

elprans:
So this has the effect of removing the "forced materialization" feature from free objects, which would affect performance of queries where we use this hack.  We should introduce an alternative (`materialized(<expr>`?) before this can land.

msullivan:
> So this has the effect of removing the "forced materialization" feature from free objects, which would affect performance of queries where we use this hack. We should introduce an alternative (`materialized(<expr>`?) before this can land.

Oh, this doesn't actually remove the forced materialization yet. We still do that for now.

edb/pgsql/compiler/shapecomp.py
msullivan: ... added a real comment

PR #7128, opened 2024-03-27
Check singleton expressions in constraints and indexes

### LINKED ISSUE #4833: invalid constraint expression reported on apply instead of create
- EdgeDB Version: 2.8+afca054
- EdgeDB CLI Version: 2.3.0-dev.895+cba06af
- OS Version: ubuntu 22.04

The following schema doesn't generate errors on migration create, but it produces `edgedb error: UnsupportedFeatureError: set returning operator <QualName std::DISTINCT> is not supported in simple expressions` on migration apply. It seems like this could be detected at create time instead.

```
module default {
  type MyThing {
    property traits -> array<str> {
      constraint expression on (array_agg(distinct array_unpack(__subject__)) = __subject__)
    }
  };
}
```

dnwpark:
Since constraint and index expressions are singleton, some functions and operations are not allowed. However, the ql to ir compilation step is not aware of singleton mode and cannot check for these errors. As a result, some errors in the schema could not be checked during the `migration create` step.

By compiling the relevant expressions to sql, any potential downstream errors can be caught in advance.

Additionally the error message was updated:
- to refer to "singleton" expressions" instead of "simpleton expressions"
- to drop "QualName"

close #4833


edb/schema/constraints.py
```diff
@@ -582,6 +582,10 @@ def compile_expr_field(
                     ),
                 )
 
+                utils.try_compile_irast_to_sql_tree(compiled, self.span)
```
msullivan, line 585: Ad a comment about why

msullivan (approved):

PR #7135, opened 2024-03-28
Add bit_count function.

dnwpark:
Add `std::bit_count` for `int16`, `int32`, `int64`, and `bytes`.

### DIFF
--- a/docs/stdlib/numbers.rst
+++ b/docs/stdlib/numbers.rst
@@ -775,17 +775,21 @@ Definitions
 .. eql:function:: std::bit_count(val: int16) -> int64
                   std::bit_count(val: int32) -> int64
                   std::bit_count(val: int64) -> int64
+                  std::bit_count(bytes: bytes) -> int64
 
-    Return the number of bits set in the binary representation of the absolute
-    value of the integer. This is also known as the population count.
+    Return the number of bits set in the :eql:type:`bytes` value.
+    
+    This is also known as the population count.
 
     .. code-block:: edgeql-repl
 
         db> select bit_count(255);
         {8}
+        db> select bit_count(b'\xff\xff');
+        {16}
 
 
-----------
+------------
 
 
 .. eql:function:: std::to_bigint(s: str, fmt: optional str={}) -> bigint
--- a/docs/stdlib/numbers.rst
+++ b/docs/stdlib/numbers.rst
@@ -103,6 +103,9 @@ Bitwise functions
   * - :eql:func:`bit_rshift`
     - :eql:func-desc:`bit_rshift`
 
+  * - :eql:func:`bit_count`
+    - :eql:func-desc:`bit_count`
+
 String parsing
 --------------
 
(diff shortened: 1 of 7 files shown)


docs/stdlib/numbers.rst
```diff
@@ -769,6 +772,21 @@ Definitions
 ------------
 
 
+.. eql:function:: std::bit_count(val: int16) -> int64
+                  std::bit_count(val: int32) -> int64
+                  std::bit_count(val: int64) -> int64
+
+    Count the number of set bits in an integer.
```
elprans, line 779:
```suggestion
    Return the number of bits set in the binary representation of the absolute value of the integer. This is also known as the population count.
```

edb/lib/std/26-bitwisefuncs.edgeql
```diff
@@ -333,3 +333,36 @@ std::bit_lshift(val: std::int64, n: std::int64) -> std::int64
     )
     $$;
 };
+
+CREATE FUNCTION
+std::bit_count(val: std::int16) -> std::int64
+{
+    CREATE ANNOTATION std::description :=
+        'Count the number of set bits in a 32-bit integer.';
```
Gobot1234, line 341:
```suggestion
        'Count the number of set bits in a 16-bit integer.';
```

edb/lib/std/26-bitwisefuncs.edgeql
```diff
@@ -333,3 +333,36 @@ std::bit_lshift(val: std::int64, n: std::int64) -> std::int64
...
+    USING SQL $$
+    SELECT bit_count(val::bit(32))
+    $$;
+};
+
+CREATE FUNCTION
+std::bit_count(val: std::int64) -> std::int64
+{
+    CREATE ANNOTATION std::description :=
+        'Count the number of set bits in a 32-bit integer.';
```
Gobot1234, line 363:
```suggestion
        'Count the number of set bits in a 64-bit integer.';
```

elprans (commented):
Should also add a `bit_count(val: std::bytes)` variant.

msullivan:
> Should also add a `bit_count(val: std::bytes)` variant.

That requires a little actual work, but should be doable.

It's *almost* doable easily enough in EdgeQL using these bit_counts, but not quite, because I don't think we have an easy way to turn single bytes into integers and back.

We have int16, 32, and 64 converters, but nothing for single bytes. (Probably because we don't have a type that exactly matches.)

I think we should also add converters between length-1 bytes values and int64 that transforms a single byte

msullivan:
bit_count on bytes could be implemented in edgeql with
```
sum((bit_count(ord(b[range_unpack(range(0, len(b)))
```
where `ord` turns a length-1 `bytes` into an integer

I implemented it with
```
create function ord(b: bytes) -> int64 using (
  <int64>to_int16(b ++ b'\x00', Endian.Little)
);
```
which is obviously wildly silly

elprans:
Postgres has a bit_count(bytea) function.

msullivan:
> Postgres has a bit_count(bytea) function.

Then we should probably use that instead of doing any of the things that I said.

(Though I think we do still want `ord`/`chr` for bytes)

elprans:
> (Though I think we do still want `ord`/`chr` for bytes)

Postgres has those also (`ord` is called `ascii` which I think is more appropriate)

msullivan:
> > (Though I think we do still want `ord`/`chr` for bytes)
> 
> Postgres has those also (`ord` is called `ascii` which I think is more appropriate)

Not for `bytes`. (Honestly `ord` and `chr` aren't really appropriate for `bytes` either though.)

msullivan (approved):
Looks good. Need to fix something with docs test.

One other thing that is needed is, whenever the standard library is changed, to update `EDGEDB_CATALOG_VERSION` in `edb/buildmeta.py`

PR #7144, opened 2024-04-02
Add link properties to GroupingIdent

### LINKED ISSUE #7013: Unable to group by link property
- EdgeDB Version: 5.0-beta.1+15775e6

Schema:
```
multi user_reactions: User {
   emoji: str;
}
```

```
with grouped := (group .user_reactions by @emoji)
select grouped
```
`Unexpected '@'`

dnwpark:
Allow link properties to be included in the `by` clause of grouping statements.

Given a schema:
```
  type User {
    required name: str;
  }

  type Post {
    multi user_reactions: User {
      emoji: str;
    };
  }
```

It allows the following query:
```
select Post {
    blah := (
        with grouped := (
            group .user_reactions
            by @emoji
        )
        select grouped {**}
    )
};
```

close #7013

### DIFF
--- a/edb/edgeql/desugar_group.py
+++ b/edb/edgeql/desugar_group.py
@@ -95,7 +95,7 @@ def desugar_group(
     aliases: AliasGenerator,
 ) -> qlast.InternalGroupQuery:
     assert not isinstance(node, qlast.InternalGroupQuery)
-    alias_map: Dict[str, Tuple[str, qlast.Expr]] = {}
+    by_alias_map: Dict[str, Tuple[str, qlast.Path]] = {}
 
     def rewrite_atom(el: qlast.GroupingAtom) -> qlast.GroupingAtom:
         if isinstance(el, qlast.ObjectRef):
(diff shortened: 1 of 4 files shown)


aljazerzen (approved):

edb/edgeql/desugar_group.py
```diff
@@ -147,7 +147,17 @@ def rewrite(el: qlast.GroupingElement) -> qlast.GroupingElement:
...
+        collect_grouping_atoms(by)
+    )
+    actual_keys: ordered.OrderedSet[str] = ordered.OrderedSet()
+    for key, key_span in keys:
+        if key in actual_keys:
+            raise errors.QueryError(
+                f"BY clause cannot refer to two properties with the same name",
+                span=key_span,
+            )
+        actual_keys.add(key)
```
msullivan, line 160: I think this is actually somewhat overzealous, now that I think about it.
```
_localdev:scope> group Card { name, cost, element } by {.cost, (.cost, .element)};
edgedb error: QueryError: BY clause cannot refer to two properties with the same name
_localdev:scope> 
```
This ought to work, certainly. (Also, the span isn't getting reported.)

We *could* try to do something more careful, and consider whether something is being used in a compound key, but I don't think it's worth it. I now think it's fine to allow stuff like `{.foo, .foo}` as long as we disallow `{.foo, @foo}`.

(The former case is a little ambiguous, but we *do* at least report the correct grouping and the correct key value, even if we can't distinguish the two identical versions. The latter will always omit key information.)

msullivan:
Also, the final commit message ought to have something like "Support link properties in GROUP BY clauses" -- I tend to prefer commit message headers that describe the outcome, not the goal, because it makes life easier for the release manager preparing the changelog.

msullivan (approved):

PR #7149, opened 2024-04-02
Add error when renaming module

dnwpark:
When running the following ddl:
```
  ALTER MODULE ModuleTest RENAME TO ModuleTest2;
```

The following error is raised:
```
error: SchemaError: Renaming modules is not supported
  │
4 │   ALTER MODULE ModuleTest RENAME TO ModuleTest2;
  │                           ^^^^^^^^^^^^^^^^^^^^^ error

```


edb/schema/delta.py
```diff
@@ -3530,10 +3530,17 @@ def _cmd_from_ast(
         astnode: qlast.DDLOperation,
         context: CommandContext,
     ) -> RenameObject[so.Object_T]:
+        from . import modules
+
         parent_ctx = context.current()
         parent_op = parent_ctx.op
         assert isinstance(parent_op, ObjectCommand)
         parent_class = parent_op.get_schema_metaclass()
+        if parent_class == modules.Module:
```
msullivan, line 3539: I think a cleaner approach would be to create a `RenameModule` implementation in `modules.py` and have it override `apply` to raise the error.

dnwpark: Right now, renaming in general seems to be done via `AlterX` with a `Rename` subcommand. I think adding a `RenameModule` would be a significant deviation from the current implementation.

msullivan: It's always a `RenameFoo` subcommand. We have one for everything that supports renaming, though frequently it has no class body.

msullivan:
Put the issue being fixed in the PR text/commit message; it makes it easier to keep track of things, and automatically will close the issue when the fix is merged

msullivan (approved):
There's a mypy/lint error and what looks like a wierd gh actions flake

PR #7151, opened 2024-04-03
Add note about `\expand` command in CLI REPL

scotttrinh:
Took the text here from the help text in the CLI, but maybe we want to expand (🥁 ) on this description a bit?

### DIFF
--- a/docs/cli/edgedb.rst
+++ b/docs/cli/edgedb.rst
@@ -194,6 +194,13 @@ many of the commands:
   Run a query performance analysis on the given query. Most conveniently used
   without a backslash by just adding ``analyze`` before any query.
 
+:cli:synopsis:`\\expand`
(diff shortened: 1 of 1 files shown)


raddevon:
> …maybe we want to expand (🥁 ) on this description a bit?

I believe I suggested this once upon a time, and I was told that we may move things around here so we don't want to be explicit about what is in here versus the standard output. I don't love that approach. Maybe things have changed since then?

raddevon (approved):

scotttrinh:
> I believe I suggested this once upon a time, and I was told that we may move things around here so we don't want to be explicit about what is in here versus the standard output. I don't love that approach. Maybe things have changed since then?

Cool. I'll merge this as-is for now and we can revisit with the team.

PR #7152, opened 2024-04-03
Add branches documentation throughout

raddevon:
This should fill in most of the remaining gaps left by previous branches documentation PRs. This does has some gaps of its own where documentation is blocked by ongoing development, and those have been marked here with `TODO` comments.

I also did some other tune-ups to bring the docs up to speed for v5 and fixed minor issues I found as I was auditing everything. It's a big PR, but I don't think there's another way that makes much sense for an effort like this.

Based on the branch that is the subject of this PR: https://github.com/edgedb/edgedb/pull/7094


raddevon:
This has a soft dependency on https://github.com/edgedb/edgedb.com/pull/856. The build shouldn't fail without that being merged, but that PR moves our cheatsheets to the "Database" section of the docs to communicate they are versioned.

PR #7154, opened 2024-04-03
Add hint when calling a type or casting to function.

### LINKED ISSUE #1601: Better suggestions when a function doesn't exist but a type does
```
ERROR: InvalidReferenceError: function 'cal::local_date' does not exist
```

could suggest something in the vein of "Did you mean:".

This is similar but not identical to #915.

dnwpark:
Given `select <to_str>(1);`, the following error is produced:
```
error: InvalidReferenceError: type 'to_str' does not exist
  ┌─ <query>:1:9
  │
1 │ select <to_str>(1);
  │         ^^^^^^ did you mean to call 'to_str'?
```

Given `select str(1);`, the following error is produced:
```
error: InvalidReferenceError: function 'default::str' does not exist
  ┌─ <query>:1:8
  │
1 │ select str(1);
  │        ^^^^^^ did you mean to cast to 'str'?
```

close #1601

### DIFF
--- a/edb/edgeql/compiler/expr.py
+++ b/edb/edgeql/compiler/expr.py
@@ -35,6 +35,7 @@ from edb.ir import utils as irutils
 
 from edb.schema import abc as s_abc
 from edb.schema import constraints as s_constr
+from edb.schema import functions as s_func
 from edb.schema import globals as s_globals
 from edb.schema import indexes as s_indexes
 from edb.schema import name as sn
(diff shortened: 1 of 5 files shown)


edb/edgeql/compiler/expr.py
```diff
@@ -646,7 +647,34 @@ def compile_GlobalExpr(
...
+            and isinstance(expr.type.maintype, qlast.ObjectRef)
+        ):
+            type_name: str = expr.type.maintype.name
+            type_module: Optional[str] = expr.type.maintype.module
+
+            name: sn.Name
+            if type_module is not None:
+                name = sn.QualName(type_module, type_name)
+            else:
+                name = sn.UnqualName(type_name)
```
msullivan, line 665: There's a function in schema.utils (I think?) for converting from an ObjectRef to a Name

edb/edgeql/compiler/func.py
```diff
@@ -95,7 +96,23 @@ def compile_FunctionCall(
...
+    except errors.InvalidReferenceError as e:
+        s_utils.enrich_schema_lookup_error(
+            e,
+            funcname,
+            modaliases=ctx.modaliases,
+            schema=env.schema,
+            suggestion_limit=1,
+            span=expr.span,
+            hint_text='did you mean to cast to'
+        )
```
msullivan, line 113: Probably want to pass an item_type

msullivan (approved):

PR #7157, opened 2024-04-04
Refactor QL AST

aljazerzen:
This PR refactors QL AST to make it more "flat": such that there each node is either inheriting from an abstract node, or is abstract itself.

This is fully accomplished from the "expression" part of the AST. The "DDL" and "SDL" part is still left to do.

- Refactor qlast.SpecialAnchor to also include subject and source
- Refactor out SetConstructorOp
- Refactor our qlast.BaseRealConstant
- Refactor out PipelinedQuery
- Mark TypeExpr abstract
- gen-ast-inheritance

### DIFF
--- a/edb/edgeql/ast.py
+++ b/edb/edgeql/ast.py
@@ -174,7 +174,7 @@ class ObjectRef(BaseObjectRef):
 
 
 class PseudoObjectRef(BaseObjectRef):
-    ''' anytype, anytuple or anyobject '''
+    '''anytype, anytuple or anyobject'''
     name: str
 
 
--- a/edb/edgeql/ast.py
+++ b/edb/edgeql/ast.py
@@ -224,10 +224,6 @@ class BinOp(Expr):
     rebalanced: bool = False
 
 
-class SetConstructorOp(BinOp):
-    op: str = 'UNION'
-
-
 class WindowSpec(Base):
     orderby: typing.List[SortExpr]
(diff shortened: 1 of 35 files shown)


edb/edgeql/ast.py
```diff
@@ -173,11 +173,15 @@ class ObjectRef(BaseObjectRef):
 
 
 class PseudoObjectRef(BaseObjectRef):
-    # anytype, anytuple or anyobject
+    ''' anytype, anytuple or anyobject '''
```
msullivan, line 176: no enclosing spaces

edb/edgeql/compiler/astutils.py
```diff
@@ -93,7 +93,9 @@ def is_ql_path(qlexpr: qlast.Expr) -> bool:
 
     start = qlexpr.steps[0]
 
-    return isinstance(start, (qlast.Source, qlast.ObjectRef, qlast.Ptr))
+    return isinstance(start, (qlast.ObjectRef, qlast.Ptr)) or (
+        isinstance(start, qlast.Anchor) and start.name == '__source__'
```
msullivan, line 97: Hm... Should other anchors be true here?

aljazerzen: Yes

aljazerzen: although this function is never used

PR #7158, opened 2024-04-04
Add cfg::Config._query_cache_mode

### LINKED ISSUE #7102: Add a temporary config to switch query cache mode

fantix:
Now users may control the cache mode per statement or database. Changing the effective mode on a database will clear its cache registration table.

Fixes #7102, replaces #7122

### DIFF
--- a/edb/lib/cfg.edgeql
+++ b/edb/lib/cfg.edgeql
@@ -194,7 +194,7 @@ ALTER TYPE cfg::AbstractConfig {
             'Recompile all cached queries on DDL if enabled.';
     };
 
-    CREATE PROPERTY _query_cache_mode -> cfg::QueryCacheMode {
+    CREATE PROPERTY query_cache_mode -> cfg::QueryCacheMode {
         SET default := cfg::QueryCacheMode.Default;
         CREATE ANNOTATION cfg::affects_compilation := 'true';
         CREATE ANNOTATION std::description :=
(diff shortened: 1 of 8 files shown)


edb/lib/cfg.edgeql
```diff
@@ -192,6 +194,13 @@ ALTER TYPE cfg::AbstractConfig {
             'Recompile all cached queries on DDL if enabled.';
     };
 
+    CREATE PROPERTY _query_cache_mode -> cfg::QueryCacheMode {
```
msullivan, line 197: Are we sure we want the leading `_`?

msullivan (approved):
Looks good!

edb/lib/cfg.edgeql
fantix: I guess so, like we'll probably drop this once cache is stable? Also I think it needs more work to be a public config, e.g. it's not yet clearing cache if changed on system level, or if we really want to allow all modes of cache exist at once by allowing per-statement cache policy (it works implicitly, just confusing).

msullivan: Hm, ok.
I sort of expect we'll keep it, though?

fantix: I'll drop the leading underscore - we just don't have to advertise it now.

PR #7159, opened 2024-04-04
Refactor irast call

dnwpark:
This PR does two things:
- Moves the parameter typemods into CallArg
- Changes the arguments of irast.Call into a dictionary, which stores named parameters by name.
    - Currently, arguments are stored as a list, with the named parameters coming first, sorted alphabetically.


dnwpark:
Not entirely sure why mypy fails...

edb/ir/ast.py
```diff
@@ -952,11 +953,7 @@ class Call(ImmutableExpr):
     # Bound arguments.
     # Named arguments will come first, sorted by name,
     # followed by positional arguments, in order of declaration.
```
aljazerzen, line 955: This comment is not true anymore, right?

Named arguments don't come before positional ones.

edb/common/ast/match.py
```diff
@@ -1,206 +0,0 @@
...
-from __future__ import annotations
-
-import collections.abc
-import types
-
-from edb.common import adapter
-
-
-class MatchASTMeta(adapter.Adapter):
-    pass
```
aljazerzen, line 31: @msullivan your dept toward our remove-meta-classes fund has been erased.

aljazerzen (approved):
Looks good.

edb/common/ast/match.py
dnwpark: I removed a bunch of metaclasses from the parser a while back so I think he's in the surplus now and should re-add some more for balance.

edb/ir/ast.py
dnwpark: Good catch. Did another sweep of the code to check if i missed any other comments.

PR #7164, opened 2024-04-08
Fix missing span when overloading fields with mismatch type

XeniaLu:
An overloaded field may raise either `SchemaDefinitionError` or `SchemaError` when the cardinality or target type mismatches its base. In case of type mismatch, the `SchemaError` is not patched with the corresponding span with error, leading to an error without source indication, like:

```
$ edgedb migration create
edgedb error: SchemaError: cannot redefine property 'v' of object type 'default::B' as scalar type 'std::bool'
  Detail: property 'v' is defined as scalar type 'std::str' in parent object type 'default::A'
```

For schema:

```esdl
type A {
  required v: str;
}

type B extending A {
  overloaded v: bool;  # added this line
}
```

With this PR:

```
error: cannot redefine property 'v' of object type 'default::B' as scalar type 'std::bool'
  ┌─ dbschema/default.esdl:6:17
  │
8 │   overloaded v: bool;
  │                 ^^^^ error
  │
  = property 'v' is defined as scalar type 'std::str' in parent object type 'default::A'

```

The added test covers this fix with the `position` assertion.

### DIFF
--- a/edb/schema/inheriting.py
+++ b/edb/schema/inheriting.py
@@ -141,7 +141,7 @@ class InheritingObjectCommand(sd.ObjectCommand[so.InheritingObjectT]):
                     ignore_local=ignore_local_field,
                     schema=schema,
                 )
-            except errors.SchemaDefinitionError as e:
+            except (errors.SchemaDefinitionError, errors.SchemaError) as e:
                 if (span := self.get_attribute_span(field_name)):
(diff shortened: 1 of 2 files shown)


msullivan (approved):

msullivan:
Thanks!

PR #7166, opened 2024-04-08
Refactor to add type annotations for s_expr.Expression

aljazerzen:
In cases where we use field getter on schema objects, we have a custom mypy extension that figures out the type of the field. This does not work with Pylance unfortunately, so I went around and added type annotations.

```diff
- expr = pointer.get_expr(schema)
+ expr: Optional[s_expr.Expression] = pointer.get_expr(schema)
```


msullivan (approved):
I guess?

I'm not against adding any of the specific annotations, but I don't think we want to do this universally

aljazerzen:
I've used LSP "rename property" action, which didn't find all usages of the property.
I know that mypy works and I do have it in vscode, but there is no "go to definition", "rename property" or "find usages".

PR #7169, opened 2024-04-08
Document the CLI's `migration extract` command

### LINKED ISSUE #7147: Document `edgedb migration extract`
From the CLI help:
> Extract migration history from the database and write it to <schema-dir>/migrations. Useful when a direct DDL command has been used to change the schema and now `edgedb migrate` will not comply because the database migration history is ahead of the migration history inside <schema-dir>/migrations

raddevon:
Resolves #7147


scotttrinh (approved):
🎉

PR #7172, opened 2024-04-09
Add support for type expressions in intersections

dnwpark:
Allow '|' and '&' in intersections.

Examples using the advanced types database:

```
{
_localdev:advancedtypes> select Ba[is Bb | Bc];
  default::CBaBc {id: 90e5466b-f69e-11ee-9882-db9098d38acf},
  default::CBaBc {id: 90e5466c-f69e-11ee-9882-9b06405472f4},
  default::CBaBbBc {id: 9234cc50-f69e-11ee-9882-7bef563d5961},
  default::CBaBbBc {id: 9234cc51-f69e-11ee-9882-c7a7a4594329},
  default::CBaBb {id: 90e54669-f69e-11ee-9882-0bcc2d82fdca},
  default::CBaBb {id: 90e5466a-f69e-11ee-9882-db6f00d601e9},
  default::CBaBbBc {id: 9234cc50-f69e-11ee-9882-7bef563d5961},
  default::CBaBbBc {id: 9234cc51-f69e-11ee-9882-c7a7a4594329},
}
```

```
_localdev:advancedtypes> select Ba[is Bb & Bc];
{
  default::CBaBbBc {id: 9234cc50-f69e-11ee-9882-7bef563d5961},
  default::CBaBbBc {id: 9234cc51-f69e-11ee-9882-c7a7a4594329},
}
```

```
_localdev:advancedtypes> select Ba[is CBa | Bb & Bc];
{
  default::CBaBbBc {id: 9234cc50-f69e-11ee-9882-7bef563d5961},
  default::CBaBbBc {id: 9234cc51-f69e-11ee-9882-c7a7a4594329},
  default::CBa {id: 900a300a-f69e-11ee-9882-5bf50e8b2919},
  default::CBa {id: 900a300b-f69e-11ee-9882-57c0233146e7},
}

```
```
_localdev:advancedtypes> select Ba[is (CBa | Bb) & Bc];
{
  default::CBaBbBc {id: 9234cc50-f69e-11ee-9882-7bef563d5961},
  default::CBaBbBc {id: 9234cc51-f69e-11ee-9882-c7a7a4594329},
}
```

### DIFF
--- a/edb/edgeql/compiler/schemactx.py
+++ b/edb/edgeql/compiler/schemactx.py
@@ -536,10 +536,6 @@ def apply_intersection(
 ) -> TypeIntersectionResult:
     """Compute an intersection of two types: *left* and *right*.
 
-    In theory, this should handle all combinations of unions and intersections
-    recursively, but currently this handles only the common case of
-    intersecting a regular type or a union type with a regular type.
-
     Returns:
         A :class:`~TypeIntersectionResult` named tuple containing the
         result intersection type, whether the type system considers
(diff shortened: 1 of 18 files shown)


edb/edgeql/compiler/schemactx.py
```diff
@@ -30,6 +30,8 @@
     Union,
     Iterable,
     Sequence,
+    List,
```
elprans, line 33: We are on Python 3.11+ so prefer using builtin types instead, e.g `list[foo]` instead of `List[foo]`

edb/edgeql/compiler/schemactx.py
```diff
@@ -552,40 +548,34 @@ def apply_intersection(
...
+            stype=right,
+            is_empty=False,
+            is_subtype=True,
+        )
+
+    if (
+        left.get_is_opaque_union(ctx.env.schema)
+        and (left_union := left.get_union_of(ctx.env.schema))
+    ):
+        left = get_union_type(left_union.objects(ctx.env.schema), ctx=ctx)
```
msullivan, line 566: The opaque union stuff really is dodgy in general. Does this represent a change of behavior, or are we doing basically what we were before for opaque unions?

msullivan: In particular, I think that intersecting with an opaque union really *ought* to just produce the intersector type---but the reason we don't want that is for implementation purposes, IIRC?

msullivan (approved):
This looks good, pending some follow-ups on opaque unions.

Let's make sure we understand exactly what the behavior is there, what we want, and have it documented where we are applying the intersections.

Great work on this!
Classic onboarding rabbit hole.
I'm happy I never did it during my first months :)

edb/schema/objects.py
```diff
@@ -3147,6 +3148,7 @@ def _issubclass(
         self,
         schema: s_schema.Schema,
         parent: SubclassableObject,
+        **kwargs: Any,
```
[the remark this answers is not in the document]
dnwpark: Seems unpacking opaque unions and views as thoroughly as I initially attempted is not necessary. I've removed the kwargs entirely.

edb/edgeql/compiler/schemactx.py
dnwpark: Always producing intersections here causes issues when compiling the standard library because it tries to create an intersection `TypeShell`.

PR #7173, opened 2024-04-09
Add rudimentary support for inline function calls

elprans:
If the IR produces a `FunctionCall` with non-empty `body` attribute, we
use that instead of compiling an SQL function call.  It is currently the
responsibility of the IR producer to make sure that any arguments are
wired properly via anchors.

### DIFF
--- a/edb/ir/ast.py
+++ b/edb/ir/ast.py
@@ -935,6 +935,8 @@ class Call(ImmutableExpr):
     """Operator or a function call."""
     __abstract_node__ = True
 
+    __ast_mutable_fields__ = frozenset(('extras', 'body'))
+
     # Bound callable has polymorphic parameters and
     # a polymorphic return type.
     func_polymorphic: bool
@@ -980,6 +982,12 @@ class Call(ImmutableExpr):
     # to this function as a subquery-as-an-expression.
     # See comment in schema/functions.py for more discussion.
     prefer_subquery_args: bool = False
+
+    # Any extra information useful for compilation of special-case callables.
+    extras: typing.Optional[dict[str, typing.Any]] = None
+
+    # Inline body of the callable.
+    body: typing.Optional[Set] = None
 
 
 class FunctionCall(Call):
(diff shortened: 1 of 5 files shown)


edb/ir/ast.py
```diff
@@ -981,6 +983,12 @@ class Call(ImmutableExpr):
     # See comment in schema/functions.py for more discussion.
     prefer_subquery_args: bool = False
 
+    # Any extra information useful for compilation of special-case callables.
+    extras: typing.Optional[dict[str, typing.Any]] = None
+
+    # Inline body of the callable.
+    body: typing.Optional[Set] = None
```
aljazerzen, line 990: So, this expression replaces the function call in backend compilation, which is effectively "inlining function in the frontend compilation".

Why cannot we inline this fully, so there `qlast.FuncCall` is compiled to the body expression, so just `irast.Expr`? Without a wrapping `irast.Call`?

aljazerzen: From the call: because we still need to compile the arguments.

edb/pgsql/compiler/expr.py
```diff
@@ -433,6 +433,10 @@ def compile_operator(
     lexpr = rexpr = None
     result: Optional[pgast.BaseExpr] = None
 
+    if expr.body is not None:
+        raise NotImplementedError(
+            "support for generic inline operator calls is not implemented")
```
aljazerzen, line 438: Maybe move body from `irast.Call` to `irast.FunctionCall`?

aljazerzen (approved):

edb/pgsql/compiler/expr.py
elprans: OK

dnwpark: I feel like inline functions ought to be a separate class entirely.

elprans: OK, I think my over-ambitious naming of the PR got people spooked. This isn't really anywhere close to actual inline functions.  Instead, think of this as support for special functions that only need to be "special" in IR, but don't require special SQL handling.

We'll revisit this when we start adding _actual_ inline functions.

PR #7176, opened 2024-04-09
Add a server helper returning a parsed type descriptor from a query string

elprans:
The new `execute.describe(query) -> TypeDesc` helper is useful for
situations when the server (or its extension) needs to reason about what
a given query returns.  This is essentially a nicely packaged
`DatabaseConnectionView.parse`.


fantix (approved):
This one looks good, though I think we forgot to `remove_dbview()` in `parse_execute_json()` in the first place (unless I misread some of our code 🤔 ).

elprans:
> This one looks good, though I think we forgot to `remove_dbview()`

We were actually missing it in a few other places too.

PR #7177, opened 2024-04-09
Provide infrastructure for long-running extension tasks

elprans:
Some extensions might benefit from the ability to run an async task
whenever the extension is present in a branch.  Add a
`start_stop_extensions()` hook which is called whenever the set of
enabled extensions in a branch changes.

Additionally, make it possible to create named tasks associated with
the server tenant.

### DIFF
--- a/edb/server/tenant.py
+++ b/edb/server/tenant.py
@@ -139,6 +139,7 @@ class Tenant(ha_base.ClusterProtocol):
 
         self._task_group = None
         self._tasks = set()
+        self._named_tasks: dict[str, asyncio.Task] = dict()
         self._accept_new_tasks = False
         self._file_watch_finalizers = []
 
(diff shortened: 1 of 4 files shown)


edb/server/tenant.py
```diff
@@ -445,19 +453,27 @@ def create_task(
...
+                rv = self.__loop.create_task(coro, name=name)
             else:
-                rv = self._task_group.create_task(coro)
+                rv = self._task_group.create_task(coro, name=name)
 
             # Keep a strong reference of the created Task
-            self._tasks.add(rv)
-            rv.add_done_callback(self._tasks.discard)
+            if name is not None:
+                self._named_tasks[name] = rv
```
fantix, line 462: If we're sure about the uniqueness of task names, I'd suggest:

```suggestion
                assert name not in self._named_tasks
                self._named_tasks[name] = rv
```

to expose potential surprises in tests.

fantix (approved):
LGTM!

PR #7185, opened 2024-04-11
Fix toy_eval_model.py so we can run it directly

msullivan:
It got broken by #6900

### DIFF
--- a/edb/tools/toy_eval_model.py
+++ b/edb/tools/toy_eval_model.py
@@ -49,6 +49,16 @@ against the real DB.
 
 from __future__ import annotations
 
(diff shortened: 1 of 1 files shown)


aljazerzen (approved):
Ohh, that here so you can do `python ./toy_eval_model.py` and have it "reset" to the root of the repo?

msullivan:
No, it's so I can run it as `python edb/tools/toy_eval_model.py` and have it work.

Probably I should delete this and make it a proper `edb` tool run as `edb toy-model` or some such

msullivan:
... or realized I could run it like `python -m edb.tools.toy_eval_model`

PR #7186, opened 2024-04-12
schema type annotations

aljazerzen:
- generate type definitions for schema classes
- definitions


aljazerzen:
The verdict here is:
- we want this, but working better. This would best be done with these .pyi files being actual
  .py files in `edb/schema/generated/` and the base classes adding the generated type-annotation classes as a mixin.

  ```python
  # edb/schema/pointers.py

  class Pointer(..., PointerMixin):
      source = so.SchemaField(
        so.InheritingObject,
        default=None, compcoef=None,
        inheritable=False
      )

  # edb/schema/generated/pointers.py
  class PointerMixin:
      def get_source(schema: Schema) -> so.InheritingObject: ...
  ```

- alternatively, we could stop creating these functions at run-time and just generate them in the mixin classes:

  ```python
  # edb/schema/pointers.py

  class Pointer(..., PointerMixin):
      source = so.SchemaField(
        so.InheritingObject,
        default=None, compcoef=None,
        inheritable=False
      )

  # edb/schema/generated/pointers.py
  class PointerMixin:
      def get_source(self, schema: Schema) -> so.InheritingObject:
          self.get_field_value('source', schema)
  ```

aljazerzen:
@msullivan, can you have a look into why mypy is failing? A new realization is that that if I disable our plugin, the internal error disappears. I've also figured that if I disable `StructTransformer`, it disappears:

https://github.com/edgedb/edgedb/blob/schema-type-annotations/edb/tools/mypy/plugin.py#L73-L82

aljazerzen:
I've tested out performance impact of this change. Currently, there are a few commits on this PR, starting with the most basic setup and then optimizing from there:


| git ref | compile_edgeql_script_webapp_get_movie | compared to master |
|---------|----------------------------------------|--------------------|
| master      	| 74.8 ms += 2.1 ms | non-significant change |
| schema-mixins        	| 93.4 ms += 3.4 ms | 24.87% SLOWER |
| getter-specialization	| 84.3 ms += 2.9 ms | 12.66% SLOWER |
| inlined-getters	| 79.7 ms += 3.0 ms | 6.52% SLOWER |
| static-field	| 75.3 ms += 1.6 ms | non-significant change |


The `static-field` approach does generate a lot of code, which could probably be cleaned-up a little without a major impact on performance.

PR #7187, opened 2024-04-12
Accept session changes in transactions

### LINKED ISSUE #7138: Session cannot change within a Transaction (at least AP enforcement)
- EdgeDB Version: 4.7+45e4acc
- EdgeDB CLI Version: 4.1
- OS Version: mac 14.4

Minimal reproduction repo:
https://github.com/CarsonF/edgedb-bug-ap-config-ignored

Essentially:
1. Start transaction with non default session, with APs enabled
2. Run query inside transaction with a different session, with APs disabled
   Query is executed with APs enabled instead

<details>
<summary>Original Post</summary>

This is very hard to reproduce. I've spent hours narrowing it down.

Query
```edgeql
WITH __param__email := <std::str>$email
SELECT (
  SELECT std::assert_single(
    (
      WITH __scope_0_AuthIdentity := DETACHED Auth::Identity
      SELECT __scope_0_AuthIdentity { id }
      FILTER (__scope_0_AuthIdentity.user.email = __param__email)
    )
  ).passwordHash
)
```
```ts
const client = createClient().withConfig({
  apply_access_policies: false,
});

client.query("...") // fine
client.querySingle("...") // error
```
```
CardinalityViolationError: required link 'user' of object type 'Auth::Identity' is hidden by access policy
```

It appears that nothing about this query can change, and still reproduce the error.
I've tried:
- inlining email param
- removing `{ id }` shape from Identity
- not aliasing a detached Identity

All work fine. And querying with a `Many` cardinality works, but `AtMostOne` does not.

I'm starting to think that maybe the AP checks are baked into the query which is then cached, and then that cached query is used even when called with APs disabled.

</details>

fantix:
This is a leftover from the protocol 1.0 upgrade when we moved the session state to the protocol level in an HTTP-cookie style. Specifically, any custom session state from the client was not synchronized into the compiler if the command was issued in a transaction. This wasn't an issue because we don't have the `tx.with_config()` API in client bindings.

This PR fixes so that session states (config and module aliases) are also synchronized in the compiler state even in transactions.

For 5.0 users, please run the following commands to clear the query cache (assuming this config `query_cache_mode` was never touched):

```
configure current database set query_cache_mode := <cfg::QueryCacheMode>'InMemory';
configure current database reset query_cache_mode;
```

- [x] Add a test

Fixes #7138

### DIFF
--- a/edb/server/compiler/compiler.py
+++ b/edb/server/compiler/compiler.py
@@ -956,6 +956,18 @@ class Compiler:
         )
         request.deserialize(serialized_request, original_query)
 
+        # Apply session differences if any
+        if (
+            request.modaliases is not None
+            and state.current_tx().get_modaliases() != request.modaliases
+        ):
+            state.current_tx().update_modaliases(request.modaliases)
+        if (
+            (session_config := request.session_config) is not None
+            and state.current_tx().get_session_config() != session_config
+        ):
+            state.current_tx().update_session_config(session_config)
+
         units, cstate = self.compile_in_tx(
             state,
             txid,
--- a/tests/test_protocol.py
+++ b/tests/test_protocol.py
@@ -683,7 +683,10 @@ class TestProtocol(ProtocolTestCase):
         await self.con.connect()
 
         # Fixture
-        await self._execute('CREATE TYPE StateChangeInTx')
+        await self._execute('CREATE MODULE TestStateChangeInTx')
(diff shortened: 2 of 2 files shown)


scotttrinh (approved):
Seems reasonable! Do we need to add a test for the `modaliases`, too?

msullivan (approved):

CarsonF:
Thank you @fantix! 🎉

PR #7188, opened 2024-04-12
Document `cors_allow_origins` configuration

scotttrinh:
Configuration added in #6784

### DIFF
--- a/docs/reference/configuration.rst
+++ b/docs/reference/configuration.rst
@@ -51,6 +51,10 @@ Connection settings
 :eql:synopsis:`listen_port -> int16`
   The TCP port the server listens on; defaults to ``5656``.
 
+:eql:synopsis:`cors_allow_origins -> multi str`
+  Origins that will be calling the server that need Cross-Origin Resource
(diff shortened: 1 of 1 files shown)


docs/reference/configuration.rst
```diff
@@ -51,6 +51,10 @@ Connection settings
 :eql:synopsis:`listen_port -> int16`
   The TCP port the server listens on; defaults to ``5656``.
 
+:eql:synopsis:`cors_allow_origins -> multi str`
+  Origins that will be calling the server that need Cross-Origin Resource
+  Sharing (CORS) support. Can use ``*`` to allow any origin.
+
```
jaclarke, line 57: Maybe we should be more specific and mention that this configures which `Origin`s on a request the server will do preflight responses, and set the `Access-Control-Allow-Origin` header on responses for?

scotttrinh: @jaclarke ⚡  [`f514a10` (#7188)](https://github.com/edgedb/edgedb/pull/7188/commits/f514a10198030b05c743ff126ce106127dfd1e28)

wdyt?

jaclarke (approved):

PR #7189, opened 2024-04-12
Generalize test._xfail a bit more

msullivan:
For some future work I'm doing I want to be able to set up an xfail
decorator that allows either an error or a failure.

### DIFF
--- a/edb/tools/test/__init__.py
+++ b/edb/tools/test/__init__.py
@@ -37,6 +37,7 @@ from edb.testbase.server import get_test_cases
 from edb.tools.edb import edbcommands
 
 from .decorators import not_implemented
+from .decorators import _xfail
 from .decorators import xfail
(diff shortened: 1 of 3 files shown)


aljazerzen (approved):
Ok, so "fail" means that there was an assertion error, "error" means that server/compiler threw an error.

This PR allows xfail to have `allow_errors`, so you can say "either assertion error or an edgedb error".

I guess, useful.

Overall, I think we shouldn't be differentiating between fail and error on the test-suite level. It should be only success or fail and edgedb error assertions should be specified in the test itself.

msullivan:
The reason we have the distinction in general is because a miscompilation is worse than an ISE, and we had a couple instances where tests that were xfailed because they ISEd morphed into a miscompilation after some change, but that never got noticed.

PR #7194, opened 2024-04-12
[5.x] Allow either b1/b2 catver or b3 catver

msullivan:
EDGEDB_CATALOG_VERSION got spuriously bumped in b890bd79, which prevents
b3 from reading b1/b2 databases.

Make sure the next version can upgrade from either.

### DIFF
--- a/edb/buildmeta.py
+++ b/edb/buildmeta.py
@@ -55,7 +55,15 @@ from edb.common import verutils
 
 
 # Increment this whenever the database layout or stdlib changes.
+#
+# N.B: DO NOT INCREMENT THIS WHEN BACKPORTING CHANGES TO A RELEASE BRANCH.
+# The merge conflict there is a nice reminder that you probably need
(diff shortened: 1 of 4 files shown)


elprans (approved):

edb/server/server.py
```diff
@@ -1273,9 +1273,18 @@ async def _maybe_apply_patches(
                     patch_info = await bootstrap.gather_patch_info(
                         num, kind, patch, conn
                     )
+
+                    # Reload the compiler state from this database in
+                    # particular, so we can compiler from exactly the
+                    # right state. (Since self._std_schema and the like might
+                    # be further advanced.)
```
aljazerzen, line 1280:
```suggestion
                    # particular, so we can compile from exactly the
                    # right state. (Since self._std_schema and the like might
                    # be further advanced.)
```

aljazerzen (approved):
If I understand correctly, this will make it so 5.x databases that have been created from beta1 or beta2 will have a bit different std schema?

PR #7200, opened 2024-04-15
pg_ext: don't yield NoData in SimpleQuery

### LINKED ISSUE #6682: Incompatibility with lib/pq Golang PostgreSQL driver for read-only SQL
We utilize the [read-only SQL](https://www.edgedb.com/docs/reference/sql_support) feature of EdgeDB to export data to various third-party tools. We were having difficulty hooking up one of these tools - and after working with that provider, we were able to narrow down the issue to some unsupported functionality on the EdgeDB PostgreSQL side. Reporting it here so that it can be fixed, as it may be affecting others.

- The provider uses [lib/pq](https://github.com/lib/pq/) (a Golang library) to connect to Postgres databases.
- A common pattern with that library is to use `db.Ping()` ([code here](https://github.com/lib/pq/blob/3d613208bca2e74f2a20e04126ed30bcb5c4cc27/conn_go18.go#L90-L100)) to test the new database connection before continuing.
- Under the hood, this appears to make some sort of "no-op" query to PostgreSQL, although the [code](https://github.com/lib/pq/blob/master/conn.go#L701) is hard for me to grok what is going on.
- This no-op query made by `db.Ping()` is what ultimately raises an error and prevents the code from continuing. The error reportedly seen by the provider was: `pq: unknown response for simple query: 'n'`

Reproducing should be as simple as connecting to the read-only SQL endpoint of an EdgeDB instance with lib/pq and calling the `db.Ping()` function.

The provider was able to work around the issue by switching to use `sqlx.Open` - and avoiding the use of the `Ping()` function.

- EdgeDB Version: 4.2+539e5c8
- EdgeDB CLI Version: 4.0.2+6d63edc
- OS Version: Debian Bullseye (Docker)

fantix:
Fixes #6682

### DIFF
--- a/edb/server/pgcon/pgcon.pyx
+++ b/edb/server/pgcon/pgcon.pyx
@@ -1660,6 +1660,7 @@ cdef class PGConnection:
         cdef:
             WriteBuffer buf, msg_buf
             PGMessage action
+            bint be_parse
 
         buf = WriteBuffer.new()
         state = None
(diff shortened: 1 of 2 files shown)


msullivan (approved):

aljazerzen (commented):
Looks good to me.

PR #7202, opened 2024-04-15
[5.x] TEST: Investigate some problems with cache after upgrade

msullivan:


msullivan:
Closing in favor of https://github.com/edgedb/edgedb/pull/7203 which I will backport normally

PR #7203, opened 2024-04-15
Fix deserialization of persistent cache entries after upgrade

msullivan:
Deserialize config in rpc using current proto version always.
We always encode that way, but would decode using proto v1 if the
query was submitted with it.

And populate the query cache with the CLI in the patches test so that
this gets caught there.


msullivan:
Forward version of https://github.com/edgedb/edgedb/pull/7202

elprans (approved):

fantix (approved):

PR #7220, opened 2024-04-18
[5.x] Fix several bugs in the AI extension

msullivan:
I'm making these changes on 5.x first and will forward port to
master. This is because I think that the patch mechanism is a critical
component.

 * Fix trivial index overriding by tweaking ddl_identity.
 * Make UPDATE triggers not always ISE
 * Filter out rows with NULL embeddings in fts::search
 * Update the triggers when overriding an index
 * Fix nontrivial index overriding.

The last one is the most impactful thing here.

The issue was that indexes in children were expected to have the same
name as in the parent, but did not. The "effective object index" in a
child, if the child defined the index itself, won't actually be a
child of the parent index, which was how the name was found.  It would
not be hard to do this computation right (and I had that part
working), but this approach has a major downside: *adding* an index to
a parent can require the column name to be changed in the child.
Additionally, it can't work if there are multiple independent base types
with ai indexes.

Another approach would be to always use the indexes *own* id as the
name.  I started along this path too, and had it much of the way. The
main downside (which I hadn't started on) is that handling the EXPLAIN
mode where inhviews are not used would be a pain.  This is *almost*
compatible with the existing database layouts (since nontrivial
inheritance doesn't work anyway) but differs when there is trivial
inheritance of indexes (where the child has no index specified).
So we would need to apply a patch to rename columns and whatnot.

Instead I went with a simplifying approach of making everything
use the same name. It's generalized in a way were it would be
very easy to use the model_name in the name instead, though.

This did also require some very custom patching code to rename all of
the generated columns, refresh the triggers, and refresh some views.
Refreshing the triggers was also necessary on is own, since they never
worked.

### DIFF
--- a/edb/pgsql/delta_ext_ai.py
+++ b/edb/pgsql/delta_ext_ai.py
@@ -61,9 +61,6 @@
 #                            -- embedding-text-3- models).
 #  )
 #
-# TODO: We could probably drop target_attr since the names should be
-# predictable now.
-#
 # The above view is a UNION of SELECTs over object relations, where each
 # UNION element is roughly this:
 #
@@ -796,7 +793,7 @@ def _get_index_root_id(
     schema: s_schema.Schema,
     index: s_indexes.Index,
 ) -> str:
-    return s_indexes.get_fts_index_id(schema, index)
+    return s_indexes.get_ai_index_id(schema, index)
 
 
 def patch_fix_ai_indexes(
--- a/edb/pgsql/delta_ext_ai.py
+++ b/edb/pgsql/delta_ext_ai.py
@@ -768,4 +768,4 @@ def _get_index_root_id(
     schema: s_schema.Schema,
     index: s_indexes.Index,
 ) -> str:
-    return index.get_topmost_concrete_base(schema).id.hex
+    return s_indexes.get_fts_index_id(schema, index)
--- a/edb/pgsql/delta_ext_ai.py
+++ b/edb/pgsql/delta_ext_ai.py
@@ -135,7 +135,7 @@ def create_ext_ai_index(
     # inherited from the parent, we don't need to create the index, but just
     # update the populating expressions.
     if has_overridden:
-        return _refresh_ai_embeddings_source_view(
+        return _refresh_ai_embeddings(
             index,
             options,
             schema,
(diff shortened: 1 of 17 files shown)


edb/schema/indexes.py
```diff
@@ -1694,6 +1702,14 @@ def is_fts_index(
     return index.issubclass(schema, fts_index)
 
 
+def get_fts_index_id(
```
elprans, line 1705: Why is this named `fts_index`?

edb/schema/indexes.py
```diff
@@ -1694,6 +1702,14 @@ def is_fts_index(
...
+    schema: s_schema.Schema,
+    index: Index,
+) -> str:
+    # TODO: Use the model name?
+    return f'base'
```
elprans, line 1710: Hmm. This is OK for 5.x since we prohibit more than one index, but ideally the name should be derived from index identity (perhaps just use DDL identity?).  Model name is not enough, because we may want to allow multiple indexes with different index parameters using the same model to generate embeddings.  Semantic search is very fiddly with respect to the kind of index and its parameters (there are 6 permutations of main params + free standing iffvlat/hnsw params).

edb/pgsql/delta_ext_ai.py
```diff
@@ -506,6 +532,7 @@ def _pg_create_trigger(
         text=f"""
         BEGIN
             NEW."__ext_ai_{idx_id}_embedding__" := NULL;
+            RETURN NEW;
```
elprans, line 535: :man_facepalming: this was lost somehow in the torrent of amends

edb/pgsql/delta_ext_ai.py
```diff
@@ -61,6 +61,9 @@
 #                            -- embedding-text-3- models).
 #  )
 #
+# TODO: We could probably drop target_attr since the names should be
```
elprans, line 64: Probably not, see review comment on `get_fts_index_id`.  We want to minimize the computation done in the indexer since CPU time there is a scarce resource.

edb/schema/indexes.py
elprans: To clarify: index (and the embedding col) selection would happen based on `ai::search` arguments.

elprans (approved):
Solid fixes, thanks!

PR #7224, opened 2024-04-18
Allow multiple authentication methods per transport in `--default-auth-method`

elprans:
Currently, `--default-auth-method` only allows one default method per transport.  This isn't very flexible and so allow multiple authentication methods to be tried in sequence (according to the
specified order in `--default-auth-method`).

Technically, this infrastructure also allows trying multiple configured `cfg::Auth` methods in order of `.priority`, but that is a bigger (and breaking) change so I left it out.

### DIFF
--- a/edb/server/args.py
+++ b/edb/server/args.py
@@ -543,7 +543,7 @@ def _validate_default_auth_method(
                     f"invalid connection transport: {transport_name}, "
                     f"supported values are: {', '.join(transport_names)})"
                 )
-
+            transport_methods = []
             for method_name in method_names.split('/'):
                 method = names.get(method_name)
                 if not method:
@@ -551,10 +551,8 @@ def _validate_default_auth_method(
                         f"invalid authentication method: {method_name}, "
                         f"supported values are: {', '.join(names)})"
                     )
-                try:
-                    methods[transport].append(method)
-                except KeyError:
-                    methods[transport] = [method]
+                transport_methods.append(method)
+            methods[transport] = transport_methods
 
     return ServerAuthMethods(methods)
 
--- a/edb/server/args.py
+++ b/edb/server/args.py
@@ -167,29 +167,31 @@ class ServerAuthMethods:
 
     def __init__(
         self,
-        methods: Mapping[ServerConnTransport, ServerAuthMethod],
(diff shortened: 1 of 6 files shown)


edb/server/args.py
```diff
@@ -1222,24 +1230,33 @@ def parse_args(**kwargs: Any):
...
+                methods[pos] = method
             else:
-                method = DEFAULT_AUTH_METHODS.get(transport)
-            methods[transport] = method
+                methods = (
+                    methods[:pos]
+                    + DEFAULT_AUTH_METHODS.get(transport)
+                    + methods[pos + 1:]
+                )
+            transport_methods[transport] = [method]
```
msullivan, line 1240: Should it not be `= methods`?

edb/server/protocol/protocol.pyx
```diff
@@ -865,43 +865,65 @@ cdef class HttpProtocol:
...
+
+            for authmethod in authmethods:
+                authmethod_name = authmethod._tspec.name.split('::')[1]
+                try:
+                    # If the auth method and the provided auth information
+                    # match, try to resolve the authentication.
+                    if authmethod_name == 'JWT' and scheme == 'bearer':
+                        if not self.is_tls:
+                            raise errors.AuthenticationError(
+                                'JWT HTTP auth must use HTTPS')
```
msullivan, line 882: We talked about dropping this also, right?
Do you want to do that separately?

msullivan (approved):

msullivan:
This is technically a potentially breaking change, so we want to make sure to give it some prominence in our release notes.
(In practice I doubt there will be much trouble. I don't think a lot of people are setting up Trust methods with low priorities that keep them from being active.)

edb/server/protocol/protocol.pyx
elprans: Separately

edb/server/args.py
elprans: Yeah, that logic was wrong entirely, I pushed a fix.

elprans:
> This is technically a potentially breaking change, so we want to make sure to give it some prominence in our release notes.

I realized that too and walked the `priority` thing back, we're still selecting only one method if configured explicitly in `cfg::Auth`.

msullivan (approved):

PR #7226, opened 2024-04-18
Stop the AI extension task when a branch is dropped

msullivan:


edb/server/dbview/dbview.pyx
```diff
@@ -173,6 +173,7 @@ cdef class Database:
         if self._cache_notify_task:
             self._cache_notify_task.cancel()
             self._cache_notify_task = None
+        ai_ext.stop_extension(self.tenant, self.name)
```
elprans, line 176:
```suggestion
        self.extensions = set()
        self.start_stop_extensions()
```

elprans (approved):

PR #7228, opened 2024-04-19
Remove type simplification from schema.

### LINKED ISSUE #7221: Property or link with union type of type and a derived type produces error
Given the schema
```
  type A;
  type A_ extending A;

  type Bad {
    property foo -> A | A_;
  };
```

The following error is produced:
```
SchemaError: object type 'default::A' already exists
```

Given the schema
```
  type A;
  type A_ extending A;
  type B;

  type Bad {
    property foo -> A | A_ | B;
  };
```

The following error is produced:
```
error: type 'default::(default:A | default:A_ | default:B)' does not exist
   │
22 │     property foo -> A | A_ | B;
   │                     ^^^^^^^^^^ error
```

dnwpark:
close #7221

Make type simplification for schema utils `ensure_union_type` and `ensure_intersection_type` explicit. Remove the type simplification when generating types from a user schema.

When creating types from a schema, the implicit simplification can cause an error when the name of the simplified type differs from the non-simplified type. This can occur if the type of a link is the union of a type and a base type (eg. `type A; type A_ extending A;` `link foo -> A | A_`).

### DIFF
--- a/edb/schema/objtypes.py
+++ b/edb/schema/objtypes.py
@@ -640,7 +640,7 @@ class AlterObjectType(
                 # them in the schema to produce the correct as_alter_delta.
                 nschema = _delete_to_delist(delete, schema)
 
-                nschema, nunion = utils.get_union_type(
+                nschema, nunion, _ = utils.ensure_union_type(
                     nschema,
                     types=union.get_union_of(schema).objects(schema),
                     opaque=union.get_is_opaque_union(schema),
--- a/edb/schema/objtypes.py
+++ b/edb/schema/objtypes.py
@@ -641,7 +641,7 @@ class AlterObjectType(
                 nschema = _delete_to_delist(delete, schema)
 
                 types = utils.simplify_union_types(
-                    schema, union.get_union_of(schema).objects(schema), False
+                    schema, union.get_union_of(schema).objects(schema)
                 )
                 nschema, nunion, _ = utils.ensure_union_type(
                     nschema,
--- a/edb/schema/objtypes.py
+++ b/edb/schema/objtypes.py
@@ -640,9 +640,12 @@ class AlterObjectType(
                 # them in the schema to produce the correct as_alter_delta.
                 nschema = _delete_to_delist(delete, schema)
 
+                types = utils.simplify_union_types(
+                    schema, union.get_union_of(schema).objects(schema), False
+                )
                 nschema, nunion, _ = utils.ensure_union_type(
                     nschema,
-                    types=union.get_union_of(schema).objects(schema),
+                    types=types,
                     opaque=union.get_is_opaque_union(schema),
                     module=union.get_name(schema).module,
                 )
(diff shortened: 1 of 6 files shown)


edb/schema/objtypes.py
```diff
@@ -640,9 +640,12 @@ def _alter_finalize(
                 # them in the schema to produce the correct as_alter_delta.
                 nschema = _delete_to_delist(delete, schema)
 
-                nschema, nunion = utils.get_union_type(
+                types = utils.simplify_union_types(
+                    schema, union.get_union_of(schema).objects(schema)
+                )
```
msullivan, line 645: Do we want this simplification? It seems like probably not?

Also the surrounding code seems bizarre and makes no sense to me, and so of course I wrote it.

edb/schema/pointers.py
```diff
@@ -3139,8 +3139,9 @@ def get_or_create_union_pointer(
 
     far_endpoints = [p.get_far_endpoint(schema, direction)
                      for p in components]
-    targets: List[s_types.Type] = [p for p in far_endpoints
+    targets: Sequence[s_types.Type] = [p for p in far_endpoints
```
msullivan, line 3142: Fix the indentation here. (flake8 would have caught this)

(Probably by dropping the iterator body to the next line.)

edb/schema/utils.py
```diff
@@ -1109,18 +1074,65 @@ def ensure_union_type(
...
+        return list(components)
 
-    schema, union, _ = ensure_union_type(
-        schema, types, opaque=opaque, module=module)
 
-    return schema, union
+def simplify_union_types_preserve_derived(
+    schema: s_schema.Schema,
+    types: Sequence[s_types.Type],
+) -> Sequence[s_types.Type]:
```
msullivan, line 1104: I probably would have kept the `preserve_derived` flag and made this one function, but I also don't object to having it be separated.

Could you add docstring comments to both?

msullivan (approved):
Looks good!

tests/test_schema.py
```diff
@@ -4029,6 +4029,18 @@ def test_schema_get_migration_union_ptrs_01(self):
...
+        schema = r'''
+        type A;
+        type A_ extending A;
+        type B;
+        type C {
+            link foo -> A | A_ | B;
+        };
+        '''
+
+        self._assert_migration_consistency(schema)
```
msullivan, line 4042: Could you reframe this as being one of the `self._assert_migration_equivalence(` tests, and testing that it works to *remove* a link with a type like this?

edb/schema/utils.py
dnwpark: I couldn't figure out a good way to do it, but some day it would be nice to have the preserve derived variant removed entirely.

msullivan: The code that it was lifted from was conditional though, right?

msullivan (approved):

edb/schema/utils.py
dnwpark: Yeah, but I think ultimately it may be best to figure out how to move the special logic all the way out to `func.compile_operator`, which is the only place which uses `preserve_derived`

PR #7237, opened 2024-04-19
Skip rebuilding query string if normalization didn't extract variables

fantix:
Hope I get @elprans's idea correctly

### DIFF
--- a/edb/edgeql-parser/edgeql-parser-python/src/normalize.rs
+++ b/edb/edgeql-parser/edgeql-parser-python/src/normalize.rs
@@ -187,10 +187,12 @@ pub fn normalize(text: &str) -> Result<Entry, Error> {
     }
 
     all_variables.push(variables);
-    let processed_source = if !all_variables.is_empty() {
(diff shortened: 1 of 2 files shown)


fantix:
@1st1 This one too - though profiling and benchmarking didn't reflect significant improvement

edb/edgeql-parser/edgeql-parser-python/src/normalize.rs
```diff
@@ -188,7 +187,11 @@ pub fn normalize(text: &str) -> Result<Entry, Error> {
     }
 
     all_variables.push(variables);
-    let processed_source = serialize_tokens(&rewritten_tokens[..]);
+    let processed_source = if !all_variables.is_empty() {
```
elprans, line 190: This is wrong, as `all_variables` is a `Vec` of `Vec` and so will not be empty.  Use the `counter <= var_idx` check below instead.

fantix: Fixed

elprans (approved):

PR #7239, opened 2024-04-19
Fix quotes in an f-string.

vpetrovykh:
This seems like an odd thing to slip through, is Python > 3.11 required for that syntax to work?

### DIFF
--- a/docs/guides/contributing/code.rst
+++ b/docs/guides/contributing/code.rst
@@ -62,8 +62,8 @@ be built with the following ``shell.nix`` file.
        venvDir = "./venv";
 
        buildInputs = with pkgs; [
-           python311Packages.python
-           python311Packages.venvShellHook
+           python312Packages.python
+           python312Packages.venvShellHook
            rustup
            autoconf
            automake
--- a/edb/server/args.py
+++ b/edb/server/args.py
@@ -179,7 +179,7 @@ class ServerAuthMethods:
 
     def __str__(self):
         return ','.join(
-            f'{t.lower()}:{'/'.join(m.lower() for m in mm)}'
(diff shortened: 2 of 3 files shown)


elprans:
EdgeDB requires Python 3.11, where this syntax it [totally OK](https://peps.python.org/pep-0701/)

vpetrovykh:
Wow, that's weird because I have python 3.11.9 in the dev environment and this caused a syntax error:
```
  File "/home/victor/.virtualenvs/edgedb/bin/edb", line 5, in <module>
    from edb.tools.edb import edbcommands
  File "/home/victor/dev/magicstack/edgedb/edb/tools/edb.py", line 34, in <module>
    from edb.server import args as srv_args
  File "/home/victor/dev/magicstack/edgedb/edb/server/args.py", line 182
    f'{t.lower()}:{'/'.join(m.lower() for m in mm)}'
                    ^
SyntaxError: f-string: expecting '}'
```
Actually, the link you posted says: `Python-Version: 3.12`
So maybe you just have 3.12 already installed?

elprans:
Ah, right, sorry 3.12 of course (we should probably just bump the requirement)

elprans (approved):

elprans:
FWIW, tests and releases use 3.12

vpetrovykh:
Yeah, maybe instead of this fix we should bump python requirement to 3.12, because this thing with f-strings will likely keep happening more in the codebase.

aljazerzen:
I've pushed a commit that bumps the python version requirement. Maybe this should go into a separate PR.

msullivan:
Well, if we bump to 3.12 we don't need to fix the quotes.

Could we just bump to 3.12, with an appropriate message?

PR #7250, opened 2024-04-23
Allow except in link constraints.

dnwpark:
Enables schemas like this:
```
    create type Foo;
    create type ExceptTest {
        create link l -> Foo {
            create property flag -> bool;
            create constraint always_fail except (@flag);
        };
    };
```


msullivan (approved):
Looks good!

PR #7251, opened 2024-04-23
Force return cast on range get upper and lower functions.

### LINKED ISSUE #6786: InternalServerError: operator is not unique: date `-` date_duration_t (from range)
- EdgeDB Version: 4.5

Steps to Reproduce:

```edgeql
with
  period := range(
    <cal::local_date>"2024-01-01",
    <cal::local_date>"2025-01-05",
    inc_upper := true
  ),
  upper := range_get_upper(period)

select upper - <cal::date_duration>"1 day"
```
```
InternalServerError: operator is not unique: date `-` date_duration_t
```

dnwpark:
Add `force_return_cast` to `range_get_upper` and `range_get_lower` functions.

These function use the built in `upper` and `lower` postgres functions. Because `range<cal::local_date>` uses the builtin `daterange`, the result has the type `date`. Without a forced cast, this causes issues when calling further functions/operators with the result.

close #6786

### DIFF
--- a/edb/buildmeta.py
+++ b/edb/buildmeta.py
@@ -55,7 +55,7 @@ from edb.common import verutils
 
 
 # Increment this whenever the database layout or stdlib changes.
-EDGEDB_CATALOG_VERSION = 2024_04_18_00_01
+EDGEDB_CATALOG_VERSION = 2024_04_25_00_00
 EDGEDB_MAJOR_VERSION = 6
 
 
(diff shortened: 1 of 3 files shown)


dnwpark:
Much cleaner than explicitly implementing the `cal` variants.

msullivan (approved):

msullivan:
Please update EDGEDB_CATALOG_VERSION also

PR #7258, opened 2024-04-25
Backport force return cast on range get upper and lower functions.

dnwpark:
From PR: #7251
Issue: #6786


msullivan:
When you squash and merge this, make sure to preserve the original commit message.

dnwpark:
Reverted for not getting tests working

PR #7265, opened 2024-04-26
Refactor to make rewrites use object's id for id of __subject__ 

aljazerzen:
Before we were creating a separate path id with name `__derived__::__subject__`.

### DIFF
--- a/edb/pgsql/compiler/dml.py
+++ b/edb/pgsql/compiler/dml.py
@@ -1831,13 +1831,12 @@ def process_update_rewrites(
     list[tuple[pgast.ResTarget, irast.PathId]],
 ]:
     # assert ir_stmt.rewrites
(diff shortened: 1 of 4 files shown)


msullivan (approved):
Oh great! I had hoped this would be possible

edb/pgsql/compiler/dml.py
```diff
@@ -907,32 +905,21 @@ def process_insert_rewrites(
 ) -> tuple[pgast.CommonTableExpr, pgast.PathRangeVar]:
     typeref = ir_stmt.subject.typeref.real_material_type
 
-    object_path_id = ir_stmt.subject.path_id
-    subject_path_id = (
-        ir_stmt.rewrites.subject_path_id
```
msullivan, line 912: Can we get rid of `subject_path_id` from the `rewrites` object too?

PR #7267, opened 2024-04-26
Switch EDGEDB_DEBUG_EDGEQL_TEXT_IN_SQL to encode string as json

msullivan:
This will make it marginally easier to consume from non-python clients
than the current approach of calling repr.

Though maybe it's not worth doing if we're going to want to consume it
from versions both before and after this.

### DIFF
--- a/edb/server/compiler/compiler.py
+++ b/edb/server/compiler/compiler.py
@@ -1917,7 +1917,8 @@ def _compile_ql_query(
     sql_text = pg_codegen.generate_source(sql_ast)
     # If requested, embed the EdgeQL text in the SQL.
     if debug.flags.edgeql_text_in_sql and source:
-        sql_text += '\n-- ' + repr(source.text())
+        sql_debug_obj = dict(edgeql=source.text())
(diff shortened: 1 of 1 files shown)


edb/server/compiler/compiler.py
```diff
@@ -1917,7 +1917,7 @@ def _compile_ql_query(
     sql_text = pg_codegen.generate_source(sql_ast)
     # If requested, embed the EdgeQL text in the SQL.
     if debug.flags.edgeql_text_in_sql and source:
-        sql_text += '\n-- ' + repr(source.text())
+        sql_text += '\n-- ' + json.dumps(source.text())
```
elprans, line 1920: Can we also prepend instead of appending it?  Postgres truncates large queries in its diagnostic views, so this will almost always be truncated away.

msullivan:
One other format question: maybe it would be worth sticking it in an object like `{"source": <source>}`, which would give us a place to embed future metadata too?

elprans:
> One other format question: maybe it would be worth sticking it in an object like `{"edgeql": <source>}`, which would give us a place to embed future metadata too?

Sure, why not

elprans (approved):

PR #7278, opened 2024-04-29
Incompatible property type message.

### LINKED ISSUE #3568: Better error message when schema contains union of types with incompatible properties
- EdgeDB Version: 2.0-dev.6375+75976cc

For the following schema:
```
type X {
  property a -> str;
}
type Y {
  property a -> int64;
}
type Z {
  link xy -> X | Y;
}
```
`edgedb migration create` returns the error:
```
edgedb error: SchemaError: cannot create a union of std::int64, std::str
```
This error message could be improved to explain the source of the error, for example something like: `cannot create link 'xy' on type 'Z' because target is a union of types 'X | Y' which have property 'a' with incompatible types 'std::int64' and 'std::str'`

dnwpark:
Given a schema:
```
type X {
  property a -> str;
}
type Y {
  property a -> int64;
}
type Z {
  link xy -> X | Y;
}
```

Produces the error:
```
error: cannot create union (default::X | default::Y) with property 'a' using incompatible types std::int64, std::str
10 │   link xy -> X | Y;
   │              ^^^^^ error
```

Similar errors are produced for:
- `select X union Y`
- `select Object is X | Y`

close #3568

### DIFF
--- a/edb/edgeql/compiler/func.py
+++ b/edb/edgeql/compiler/func.py
@@ -624,7 +624,11 @@ def compile_operator(
         left_type = setgen.get_set_type(larg, ctx=ctx)
         right_type = setgen.get_set_type(rarg, ctx=ctx)
         rtype = schemactx.get_union_type(
-            [left_type, right_type], preserve_derived=True, ctx=ctx)
+            [left_type, right_type],
+            preserve_derived=True,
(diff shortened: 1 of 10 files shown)


msullivan (approved):
Great improvement!

PR #7280, opened 2024-04-30
Add remaining config file fs watch for multi-tenants

fantix:
A known issue is that, if the watched file is moved/renamed, uvloop `_monitor_fs()` will still report the changes of the NEW path, while the reloading tasks are all using the OLD path. This effectively cripples the reload feature on fs change. This is particularly found when any config file is edited using vim. I'll create a new PR to fix it to survive renames (that also includes watching the containing folder so that a new file at the OLD path can be picked up, but that may need some discussion).

Updating the config file content works fine.

### DIFF
--- a/edb/server/server.py
+++ b/edb/server/server.py
@@ -936,6 +936,10 @@ class BaseServer:
         if client_ca_file is not None:
             self.monitor_fs(client_ca_file, reload_tls)
 
+    def start_watching_files(self):
+        # TODO(fantix): include the monitor_fs() lines above
+        pass
+
     def load_jwcrypto(self, jws_key_file: pathlib.Path) -> None:
         try:
             self._jws_key = secretkey.load_secret_key(jws_key_file)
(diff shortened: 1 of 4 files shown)


edb/server/server.py
```diff
@@ -936,6 +936,10 @@ def reload_tls(_file_modified, _event, retry=0):
         if client_ca_file is not None:
             self.monitor_fs(client_ca_file, reload_tls)
 
+    def start_watching_files(self):
+        # TODO(fantix): include the monitor_fs() lines above
```
elprans, line 940: What is this TODO and why isn't it done in this PR?

fantix: Ah that was for the watches in `init_tls()` above - it has some complications and it was too late last night. I'll fix it today with either this PR or a new one.

elprans:
> A known issue is that, if the watched file is moved/renamed, uvloop `_monitor_fs()` will still report the changes of the NEW path, while the reloading tasks are all using the OLD path. This effectively cripples the reload feature on fs change.

By the way, you shouldn't ignore the callback's `event` parameter and only do the reload on `UV_CHANGE`.  If you receive a `UV_RENAME` instead, you should stop watching that handle.

For renames, watching the containing directory is the correct fix I think.

elprans (approved):

PR #7285, opened 2024-04-30
Create key derivation function

scotttrinh:
Use a separate derived key for signing each different kind of JWTs other than the auth_token JWT. This avoids accidentally being able to use these other (short-lived) tokens as auth tokens directly.

### DIFF
--- a/edb/server/protocol/auth_ext/http.py
+++ b/edb/server/protocol/auth_ext/http.py
@@ -1694,8 +1694,8 @@ class Router:
         ) = None,
         expires_in: datetime.timedelta | None = None,
     ) -> str:
-        initial_key_material = self._get_auth_signing_key()
-        signing_key = util.derive_key(initial_key_material, derive_for_info)
+        input_key_material = self._get_auth_signing_key()
+        signing_key = util.derive_key(input_key_material, derive_for_info)
         expires_in = (
             datetime.timedelta(minutes=10) if expires_in is None else expires_in
         )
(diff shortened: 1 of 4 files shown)


edb/server/protocol/auth_ext/http.py
```diff
@@ -1685,12 +1688,14 @@ def _make_secret_token(
...
         secret: str,
+        derive_for_info: str,
         additional_claims: (
             dict[str, str | int | float | bool | None] | None
         ) = None,
         expires_in: datetime.timedelta | None = None,
     ) -> str:
-        signing_key = self._get_auth_signing_key()
+        input_key_material = self._get_auth_signing_key()
+        signing_key = util.derive_key(input_key_material, derive_for_info)
```
scotttrinh, line 1698: We talked about putting the derive function on a separate thread but that change is a little non-trivial since most of the call sites like this one are synchronous. Not a huge deal if we still want to do that since the call sites of _these_ functions are usually async, so it would just be one layer of refactoring.

I'll provide some measurements done locally to decide if it's worth doing so now.

scotttrinh: So, locally, it takes betwen 0.2ms and 0.05ms to run the derive_key function.

elprans: This seems fast enough, so let's roll with sync generation for now and revisit if this becomes a bottleneck.

elprans (approved):

fantix (approved):
LGTM!

PR #7286, opened 2024-04-30
server: fix monitor_fs() to track target recreation

fantix:
Refs #7280, this PR wraps the callbacks given to `monitor_fs()` so as to follow target file recreation.

### DIFF
--- a/edb/server/server.py
+++ b/edb/server/server.py
@@ -38,6 +38,7 @@ import itertools
 import json
 import logging
 import os
+import pathlib
 import pickle
 import socket
 import ssl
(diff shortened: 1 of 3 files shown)


edb/server/server.py
```diff
@@ -371,13 +371,46 @@ def cancel_pgext_connection(self, pid, secret):
...
+                else:
+                    finalizer()
+                    handle = new_handle
+                    self._file_watch_handles.append(handle)
+
+        def callback(_file_modified, event):
+            nonlocal handle
+            if event == 2:  # CHANGE
+                cb()
+            else:  # RENAME, RENAME_CHANGE
```
elprans, line 399: Test for these events explicitly. Otherwise you might react wrongly to a non-rename event.

elprans (approved):

PR #7290, opened 2024-05-01
Fix handling of enums in arrays and multi properties.

### LINKED ISSUE #3990: Property of array<ENUM> breaks the graphql extension
- EdgeDB Version: 1.4
- OS Version: tested on many

Steps to Reproduce:

1. See schema below and add the graphql extension line and then go to the gql endpoint.

Schema:

```
scalar type Tag extending enum<Rock, HipHop>;

type Artist {
   property name -> str;
   property tags -> array<Tag>;
}
```

When you go to the graphql endpoint you get the error `TypeError: InsertArtist fields cannot be resolved.` When you comment out the property `tags` it works again. If you do `multi property tags -> Tag` it works. So there is something with the `array<Tag>` that is breaking the gql extension.

vpetrovykh:
Declaring an enum array property or a multi property should not result in an error in GraphQL. Both of these are valid types and can be inserted and updated.

Fixes #3990

### DIFF
--- a/edb/graphql/translator.py
+++ b/edb/graphql/translator.py
@@ -1683,12 +1683,14 @@ class GraphQLTranslator:
             optional = False
 
         if self.is_list_type(vartype):
-            if target.is_array:
-                raise errors.QueryError(err_msg)
-            elif target.is_multirange:
+            if target.is_multirange:
                 subtype = target.edb_base.get_subtypes(target.edb_schema)[0]
                 st_name = subtype.get_name(target.edb_schema)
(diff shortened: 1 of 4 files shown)


fmoor:
@vpetrovykh, maybe we can create a new group for people that are qualified to review `edgedb/edgedb` PRs instead of requesting review from `edgedb/engineering`? This is noise for a lot of people.

msullivan (approved):
Make sure to mention that it is a graphql fix in the final commit message

PR #7291, opened 2024-05-01
Clean up some word usage in release-process docs

msullivan:
Say "labelling" instead of "tagging"


RobertBlumen (approved):
I learned that labelled and labeled are both correct.

dev-notes/release-process.md
RobertBlumen, line 1: fix " This can be done is setup.py with variable""

msullivan:
huh I though I merged this already

PR #7293, opened 2024-05-02
Backport 5.x

vpetrovykh:


msullivan (approved):
Let's drop #7228, which I want to think about a little bit more to convince myself it can't break anything if backported. I think that means dropping #7278 also, for now.
(Though we can have @dnwpark do a new backport of it later)

Issue #2327, opened 2021-03-05
errors casting json values don't provide location


fmoor:
It would be much more helpful if errors gave a path to the error location in the json data. Ideally this information would be both obvious to humans reading the error message and available programmatically.

```python
>>> con.query("SELECT <tuple<a: int64, b: str>><json>$0", '{"a": 1}')
InvalidValueError: invalid null value in cast

>>> con.query("SELECT <tuple<a: int64, b: array<str>>><json>$0", '{"a": 1, "b": [2]}')
InvalidValueError: expected json string or null; got json number
```

- EdgeDB Version: package:1.0b2.dev5500+g2f67186c8-2021030200~focal.nightly
- OS Version: Linux 5.8.0-44-generic #50~20.04.1-Ubuntu SMP Wed Feb 10 21:07:30 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

msullivan:
This should be possible but it will touch a wide range of things.

We'll want to report paths in the json something like:
* `."a"`
* ".b"[0]`
for the examples above.

Our big composite casts from json work by breaking down the operation into a bunch of smaller casts, so we'll need to arrange to pass down extra information for error messages into those casts, and create a way for that to be passed into the cast functions.

It's possible that actually including the indexes in array casts will make it too expensive, though, in which case we'd need to be a bit less ambitious.

PR #7300, opened 2024-05-03
Add context to errors when casting to collections.

dnwpark:
If there is an incompatible element when casting to a collection, the `to` and `from` type are added as well as the name of the element.

For example, given the query `SELECT <tuple<a: array<tuple<b: int64>>>>(c:=[(d:=[1])]);`, the following error is produced:
```
error: QueryError: while casting 'tuple<c: array<tuple<d: array<std::int64>>>>' to 'tuple<a: array<tuple<b: std::int64>>>', at tuple element 'a', in array elements, at tuple element 'b', cannot cast 'array<std::int64>' to 'std::int64'
  ┌─ <query>:1:8
  │
1 │ SELECT <tuple<a: array<tuple<b: int64>>>>(c:=[(d:=[1])]);
  │        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ error
```

### DIFF
--- a/edb/edgeql/compiler/casts.py
+++ b/edb/edgeql/compiler/casts.py
@@ -158,14 +158,14 @@ def compile_cast(
         )
     ):
         cast_element = ('array', None)
-        if ctx.collection_cast_path is not None:
-            ctx.collection_cast_path.append(cast_element)
+        if ctx.collection_cast_info is not None:
+            ctx.collection_cast_info.path_elements.append(cast_element)
 
         result = _cast_array_literal(
             ir_set, orig_stype, new_stype, span=span, ctx=ctx)
 
-        if ctx.collection_cast_path is not None:
-            ctx.collection_cast_path.pop()
+        if ctx.collection_cast_info is not None:
+            ctx.collection_cast_info.path_elements.pop()
 
         return result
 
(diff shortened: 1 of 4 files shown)


dnwpark:
This should be helpful in resolving #2327

msullivan (commented):
I like the new error messages, but I don't really like catching the error at so many call sites and gradually extending it while *also* tracking the full path in the context.

How bad would it be to include the full message closer to the root of where the exception was raised, and ideally only modify it once?

edb/edgeql/compiler/context.py
```diff
@@ -569,6 +569,28 @@ class ContextLevel(compiler.ContextLevel):
...
+    In the event of a cast error, the list is preserved at the outermost cast.
+    """
+
+    collection_cast_from_type: Optional[s_types.Type]
+    """For generating errors messages when casting to collections.
+        Only ever set in outermost cast."""
+
+    collection_cast_to_type: Optional[s_types.Type]
+    """For generating errors messages when casting to collections.
+        Only ever set in outermost cast."""
```
msullivan, line 592: Maybe this would be nicer if we moved it all to one `CollectionCastInfo` object?

edb/edgeql/compiler/casts.py
```diff
@@ -157,9 +157,18 @@ def compile_cast(
             or new_stype.issubclass(ctx.env.schema, json_t)
         )
     ):
-        return _cast_array_literal(
+        cast_element = ('array', None)
+        if ctx.collection_cast_path is not None:
+            ctx.collection_cast_path.append(cast_element)
```
msullivan, line 162: So the idea is that if the compile fails, then the pops don't happen and the message gets collected at the end?

That's a little subtle, though might still be fine.

Are there too many base cases for this error to inject the message closer to the initial sites where it is raised?

msullivan: (OK yeah, and that pattern is properly documented too)

msullivan (approved):
I think this is workable, but let's at least think about whether the errors can reasonably be pushed to the leaves instead of relying on the mutable list trick here

edb/edgeql/compiler/casts.py
dnwpark: There "should" always be a base TypeCast operation which handles all this.

PR #7302, opened 2024-05-03
Add some more notes about release communication

msullivan:
Some things that I think will improve our lines of communication and
help avoid at least some of the recent snafus


vpetrovykh (approved):
I think making these steps more explicit is a very good idea.

PR #7312, opened 2024-05-07
Improve json cast error messages

### LINKED ISSUE #2327: errors casting json values don't provide location

dnwpark:
Provide context to error messages when casting to from json to collections.

Given `SELECT <tuple<a: int64>>to_json('{"b": 1}');` the following error is produced:
```
InvalidValueError: while casting 'std::json' to 'tuple<a: std::int64>', at tuple element 'a', missing value in JSON object
```

Given `SELECT <tuple<a: int64>>to_json('{"a": "b"}');` the following error is produced:
```
InvalidValueError: while casting 'std::json' to 'tuple<a: std::int64>', at tuple element 'a', expected JSON number or null; got JSON string
```

### DIFF
--- a/edb/edgeql/compiler/casts.py
+++ b/edb/edgeql/compiler/casts.py
@@ -690,19 +690,26 @@ def _cast_json_to_tuple(
         # json_get calls and see if that is faster.
         elements = []
         for new_el_name, new_st in new_stype.iter_subtypes(ctx.env.schema):
+            cast_element = ('tuple', new_el_name)
+            if subctx.collection_cast_info is not None:
+                subctx.collection_cast_info.path_elements.append(cast_element)
+
+            json_get_kwargs: dict[str, qlast.Expr] = {}
+            if error_message_context := cast_message_context(subctx):
+                detail = qlast.Constant.string(
+                    '{"error_message_context": "' + error_message_context + '"}'
+                )
+                json_get_kwargs['detail'] = detail
             val_e = qlast.FunctionCall(
-                func=('__std__', 'json_get'),
+                func=('__std__', '__json_get_not_null'),
                 args=[
                     source_path,
                     qlast.Constant.string(new_el_name),
                 ],
+                kwargs=json_get_kwargs
             )
 
             val = dispatch.compile(val_e, ctx=subctx)
-
-            cast_element = ('tuple', new_el_name)
-            if subctx.collection_cast_info is not None:
-                subctx.collection_cast_info.path_elements.append(cast_element)
 
             val = compile_cast(
                 val, new_st,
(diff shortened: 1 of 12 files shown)


dnwpark:
Looks bigger than it is :)

edb/edgeql/compiler/casts.py
```diff
@@ -646,48 +648,70 @@ def _cast_json_to_tuple(
...
+        allow_null = cardinality_mod != qlast.CardinalityModifier.Required
+
+        # Only json arrays or objects can be cast to tuple.
+        # If not in the top level cast, raise an exception here
+        json_object_args: list[qlast.Expr] = [
+            source_path,
+            qlast.Constant.boolean(allow_null),
+        ]
+        if error_message_context := cast_message_context(subctx):
+            detail = qlast.Constant.string(
```
msullivan, line 666: I think it would be cleaner to build the dict and `json.dumps` it

edb/lib/std/30-jsonfuncs.edgeql
```diff
@@ -85,6 +85,31 @@ std::json_get(
     $$;
 };
 
+
+CREATE FUNCTION
```
msullivan, line 89: We want this because it produces a better error message than leaving it to the cast, or what?

edb/server/compiler/errormech.py
```diff
@@ -453,9 +453,15 @@ def _static_interpret_invalid_param_value(
     err_details: ErrorDetails,
     from_graphql: bool = False,
 ):
+    error_message_context = ''
+    if err_details.detail_json:
+        error_message_context = (
+            err_details.detail_json.get('error_message_context', '')
+        )
+
```
msullivan, line 461: The big question for me with this PR is:
Is it better to have the context prepended to the actual message, or would it be better to put it in the `details` field?

My inclination had been `details`, but this looks pretty nice.

msullivan (approved):
This looks great, and will be a nice improvement.

Remember to bump `EDGEDB_CATALOG_VERSION`

edb/lib/std/30-jsonfuncs.edgeql
dnwpark: This function lets us produce different error messages for when a tuple is missing an element and when the element is there but explicitly set to null.

edb/server/compiler/errormech.py
dnwpark: I guess either works, but it seemed to me that details were often more about providing information from beyond the immediate context. (eg. stuff about the inheritance of the types involved in an error)

PR #7314, opened 2024-05-07
Semi-rationalize get_dummy_expr_field_value implementations

msullivan:
If there is an actual type involved that might matter,
get_dummy_expr_field_value ought to produce a value of that type.

Write one helper function for this in `edb.schema.types` and call it
everywhere that cares.

This fixes at least one bug involving globals in a somewhat twisted
interaction with policies (though I am still doing some more thinking
about those cases, so I may fix it in an additional way in a
follow-up).

### DIFF
--- a/edb/schema/types.py
+++ b/edb/schema/types.py
@@ -2910,6 +2910,16 @@ def ensure_schema_type_expr_type(
     return cmd
 
 
+def type_dummy_expr(
+    typ: Type,
+    schema: s_schema.Schema,
+) -> Optional[s_expr.Expression]:
+    if isinstance(typ, so.DerivableInheritingObject):
+        typ = typ.get_nearest_non_derived_parent(schema)
+    text = f'assert_exists(<{typ.get_displayname(schema)}>{{}})'
+    return s_expr.Expression(text=text)
+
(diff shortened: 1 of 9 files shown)


elprans (approved):

edb/schema/types.py
```diff
@@ -2910,6 +2910,16 @@ def ensure_schema_type_expr_type(
...
 
 
+def type_dummy_expr(
+    typ: Type,
+    schema: s_schema.Schema,
+) -> Optional[s_expr.Expression]:
+    if isinstance(typ, so.DerivableInheritingObject):
+        typ = typ.get_nearest_non_derived_parent(schema)
+    text = f'assert_exists(<{typ.get_displayname(schema)}>{{}})'
+    return s_expr.Expression(text=text)
```
aljazerzen, line 2920: Woah, we can construct schema objects directly using the constructor? Didn't know that.

aljazerzen (approved):

edb/schema/types.py
msullivan: `Expression` isn't really a schema object, I think

PR #7317, opened 2024-05-08
Docs: Flesh out AI docs and move to top level nav

raddevon:
This depends on https://github.com/edgedb/edgedb.com/pull/897 to get "AI" into the top-level navigation. Otherwise, it will not be linked from anywhere.


docs/ai/reference.rst
```diff
@@ -0,0 +1,357 @@
...
+
+RAG
+---
+
+``https://<edgedb-host>:<port>/branch/<branch-name>/ai/rag``
+
+Responds with text generated by the specified text generation model.
+
+
+Request body
```
scotttrinh, line 292: We should note that the body needs to be JSON.

docs/ai/reference.rst
```diff
@@ -0,0 +1,357 @@
...
+  * ``globals`` (object, optional): A dictionary of globals for use in the
+    context query.
+
+  * ``max_object_count`` (int, optional): Maximum number of objects to return;
+    default is 5.
+
+* ``stream`` (boolean, optional): Specifies whether the response should be
+  streamed. Defaults to false.
+
+* ``prompt`` (object, optional): Settings that define a prompt, overriding the
```
scotttrinh, line 318: The pseudo type language here is a little bit imprecise, but maybe that's fine.

In reality this is not an object with a bunch of optional fields but rather a union of different types. Like in TypeScript this would be:

```ts
type Prompt =
  | { name: string }
  | { id: string }
  | [{ role: string; content: string }];
```

There are more precise formats like OpenAPI/Swagger or JSON Schema which we could leverage here to be perfectly precise if we don't want to adopt a specific language's type system like TypeScript.

wdyt?

docs/ai/reference.rst
```diff
@@ -0,0 +1,357 @@
...
+  * ``custom`` (array of objects, optional): Custom prompt messages, each
+    containing a ``role`` and ``content``.
+
+
+Response
+^^^^^^^^
+
+A JSON object or a stream of events, depending on the request body's ``stream``
+parameter, containing the AI-generated data or error messages if the request
+failed.
```
scotttrinh, line 334: Yeah, we probably want more precision in the return type here as well. What is the shape of the JSON data that is returned? What does the stream data look like? Maybe some reference to [server-sent events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events)  might be useful?

docs/ai/python.rst
```diff
@@ -0,0 +1,414 @@
...
+
+   :param client:
+       An EdgeDB client instance.
+
+   :param kwargs:
+       Keyword arguments that are passed to the ``AIOptions`` data class to
+       configure AI-specific options. These options are:
+
+       * ``model``: The name of the model to be used. (required)
+       * ``prompt``: An optional prompt to guide the model's behavior. (default: None)
```
scotttrinh, line 121: I _think_ setting this to `None` explicitly or via default means you end up using the system prompt? This description is strictly true, but probably misleading as to the end result.

docs/ai/reference.rst
raddevon: I took a look at these formats, and it seems they are intended to generate documentation, not to be user-facing documentation themselves. I guess we could try to implement this, but we would probably need to add something to the documentation build system to generate docs from one of the formats.

I do take your point here though. I missed the point of the properties of `prompt`, so I've tried to clarify that you should pass only one of the properties under prompt and which to use depending on what you want to achieve. See how you feel about the outcome.

docs/ai/reference.rst
raddevon: I provided examples of both successful and error responses from each of the endpoints, and I better explained the SSE response and linked to that MDN article as well.

docs/ai/python.rst
raddevon: I added an explanation here of what the result of `None` actually is.

docs/ai/reference.rst
scotttrinh: Yeah, I guess what we could do is not use OpenAPI or any of those tools, but see how they structure the types in the documentation that they generate. I agree that we don't have to worry too much about giving an actual machine-readable format like one of those mentioned tools, and we certainly shouldn't complicate our build process.

Maybe we could even take inspiration a bit from something like the [Stripe API docs](https://docs.stripe.com/api/balance/balance_object)? Ultimately I think we just need to add a bit more detail/precision here so that someone knows _exactly_ what they can send and _exactly_ what to expect as a response.

docs/ai/reference.rst
```diff
@@ -0,0 +1,520 @@
...
+<https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events>`__
+(SSE) to stream responses. Here is a detailed breakdown of the typical
+sequence and structure of events in a streaming response:
+
+* **HTTP Status**: 200 OK
+* **Content-Type**: text/event-stream
+* **Cache-Control**: no-cache
+
+The stream consists of a sequence of five events, each encapsulating part of
+the response in a structured format:
```
scotttrinh, line 378: I wonder if we should show the raw text of what the SSE looks like in real life rather than this slightly restructured text? Or do you think that linking to the MDN article is enough?

The things that are not terribly obvious here are:
1. What does an SSE event actually look like:

   ```
   event: message_start
   data: {"type":"message_start","message":{"id":"<message_id>","role":"assistant","model":"<model_name>"}}
   
   ```
2. You might receive the text one character at a time, or one whole event at a time, or several events at once, or the whole stream at once, so you'll need to do some work to buffer the messages and restructure them yourself. Maybe this part is a bit too much for these reference-level documentation since this really due to how chunked responses work over HTTP.

raddevon: I think it wouldn't be bad to show an example response. I'll add it.

docs/ai/reference.rst
raddevon: Makes sense. I'll turn this over a bit and see what I can come up with.

Thank you!

raddevon: I've added some request examples, which I think could be helpful and some other minor details, like a note about authentication.

I guess the thing that's missing in terms of precision is the types of the response object properties? Is that what you're referring to when you say that it's missing detail and precision, or is there something else I'm overlooking? The request objects already give the types and descriptions of each property, so I'm not sure what else we can expose there. Aside from documentation like the Stripe ones you linked being prettier, I can't tell they're giving much more detail there. They definitely are on the responses, although I wonder how necessary it is with the examples. I think the request object details are pretty necessary since an example wouldn't capture every possible property on the object.

scotttrinh: In this specific case I mean that the definition for prompt is wrong in a specific kind of way. Given the current defintion you it seems like you could supply this object:

```json
{
  "prompt": {
    "name": "foo",
    "id": 123,
    "custom": [{ "role": null, "content": 42.38 }],
  }
}
```

But really this is a disjoint union like the TypeScript type I referenced earlier. So, what I'm looking for is something that makes this union-type a little more explicit. Something like "one of", or "this or that" rather than marking each property as optional. Does that make more sense? I think how Stripe would do it is to define different types of `Prompt`s and say this property is one of those types.

scotttrinh: Oh, the plot thickens a bit here after looking at the code. The type is actually more like:

```ts
interface CustomPrompt {
  role: "User" | "System";
  content: string;
}

type Prompt = 
  ({ name: string } | { id: string })
  & { custom: CustomPrompt[] };
```

Since you can provide a `custom` key along with the name or id (which is exclusive and will throw an error if you set both) and those messages will be appended to the prompt that you specified either by name or id.

raddevon: Ah, OK. I missed that you could append to an existing prompt. Did you see the prose I added to try to explain though? Does that resolve the other concern?

```rst
  You may specify an existing prompt by its ``name`` or ``id``, or you may
  define a custom prompt inline by sending an array of objects, each containing
  a ``role`` and ``content``. Choose to specify any one of these properties, or
  omit the ``prompt`` property to use the default prompt:

  * ``name`` (string, optional): The ``name`` of an existing custom prompt to
    use.

  * ``id`` (string, optional): The ``id`` of an existing custom prompt to use.

  * ``custom`` (array of objects, optional): Custom prompt messages, each
    containing a ``role`` and ``content``.
```

If I update that to address the ability to append to an existing prompt, does this solve the problem?

scotttrinh: Maybe it's not worth belaboring this, but I think my main concern is that if you squint at the definition, it's not clear that `name` and `id` are exclusive just based on looking at the structure. I think something that makes it clear structurally that `{ name: string }` and `{ id: string }` are exclusive rather than them being both optional siblings in the object's definition. Not 100% sure how to solve that (or if it's worth solving given the explanatory prose) but that's my main concern in this specific case.

But the general thing here is: how do we structure these type definitions in a friendly cross-language way that is precise enough to be non-ambiguous, but not too tied to a specific type annotation style. That's the broader question, and something we can iterate on if we're feeling stuck here.

raddevon: OK, I see. So, it's more about the implication of having the three presented as they are in the list. Yeah, that makes sense.

raddevon:
Resolved by #7169

docs/ai/javascript.rst
```diff
@@ -0,0 +1,167 @@
...
+:edb-alt-title: EdgeDB AI's JavaScript package
+
+``@edgedb/ai`` offers a convenient wrapper around ``ext::ai``. Install it with
+npm or via your package manager of choice:
+
+.. code-block:: bash
+
+    $ npm install @edgedb/ai # or
+    $ yarn add @edgedb/ai # or
+    $ pnpm add @edgedb/ai
```
beerose, line 16: Should we start including `bun` too?

docs/ai/javascript.rst
```diff
@@ -0,0 +1,167 @@
...
+EdgeDB client and any options for the AI provider (like the text generation
+model):
+
+.. code-block:: typescript
+
+    const client = createClient();
+
+    const gpt4Ai = createAI(client, {
+      model: "gpt-4-turbo-preview",
+    });
```
beerose, line 40: Should we link to somewhere where we explain what models user can use?

docs/ai/javascript.rst
```diff
@@ -0,0 +1,167 @@
...
+
+Call your AI client's ``queryRag`` method, passing in a text query.
+
+.. code-block:: typescript
+
+    console.time("gpt-4 Time");
+    console.log(
+      await astronomyAi.queryRag("What color is the sky on Mars?")
+    );
+    console.timeEnd("gpt-4 Time");
```
beerose, line 62: Is leaving `console.time` here relevant?

raddevon: I don't think so. I'll clean that up.

docs/ai/javascript.rst
raddevon: I'm glad you caught this one because I didn't have these listed in the reference at all. (They were listed only on the overview page.)

I listed them, linked to them from here, and took all of the model lists out of the overview page and instead linked to the lists in the reference.

scotttrinh (approved):

docs/ai/python.rst
```diff
@@ -0,0 +1,416 @@
...
+   An asynchronous class for creating EdgeDB AI clients.
+
+   This class provides methods to send queries and receive responses using both
+   blocking and streaming communication modes asynchronously.
+
+   :ivar client:
+       An instance of ``httpx.AsyncClient`` used for making HTTP requests
+       asynchronously.
+
+.. py:method:: query_rag(message, context=None) -> str
```
fantix, line 271: This is a `py:coroutinemethod::` I believe

fantix (commented):
LGTM

PR #7325, opened 2024-05-09
Allow code blocks in docs to contain errors if the lang ends with `-invalid`

jaclarke:


raddevon (approved):
Tests pass, even with an invalid block. Looks good!

PR #7327, opened 2024-05-09
Update libpg to 16.

dnwpark:
Update libpg to a fork of 16-latest which has a fix to how json integers are output.

Changes to the ast builder handle how newer versions of libpg describe constants.

### DIFF
--- a/edb/pgsql/parser/ast_builder.py
+++ b/edb/pgsql/parser/ast_builder.py
@@ -136,9 +136,7 @@ def _build_any(node: Node, _: Context) -> Any:
 
 
 def _build_str(node: Node, _: Context) -> str:
-    node = _unwrap(node, "String")
-    node = _unwrap(node, "str")
-    node = _unwrap(node, "sval")
+    node = _unwrap_string(node)
     return str(node)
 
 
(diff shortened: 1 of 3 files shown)


edb/pgsql/parser/ast_builder.py
```diff
@@ -156,6 +155,42 @@ def _unwrap(node: Node, name: str) -> Node:
...
+    n = _unwrap(n, 'fval')
+    n = _unwrap(n, 'fval')
+    return n
+
+
+def _unwrap_string(n: Node) -> Node:
+    n = _unwrap(n, 'String')
+    n = _unwrap(n, 'str')
+    n = _unwrap(n, 'sval')
+    n = _unwrap(n, 'sval')
```
aljazerzen, line 190: Double `sval`? What is libpg_query doing?

aljazerzen (approved):

edb/pgsql/parser/ast_builder.py
dnwpark: I tested a few different versions of libpg and the behaviour changed slightly between each one. I kept the double unwrap just in case they decide to shift the behaviour back.

PR #7337, opened 2024-05-10
Automatically do github releases in release jobs

msullivan:
I'm not sure what the best way to test/debug this is. Should we
release a 6.0a1 in order to validate it, or should we just see how it
does on 5.x point releases and go from there?

### DIFF
--- a/.github/workflows.src/build.inc.yml
+++ b/.github/workflows.src/build.inc.yml
@@ -364,6 +364,27 @@
 <%- endfor %>
 <%- endif %>
 
+<% if subdist == "" or subdist == "testing" %>
+  publish-github-release:
+    needs: collect
+    runs-on: ubuntu-latest
+
+    steps:
+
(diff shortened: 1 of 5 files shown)


.github/workflows.src/build.inc.yml
```diff
@@ -364,6 +364,27 @@
...
+
+    - name: Publish Github Release
+      uses: elprans/gh-action-create-release@master
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        tag_name: ${{ github.ref }}
+        release_name: ${{ github.ref }}
+        target: ${{ github.event.pull_request.base.ref }}
+        body: EdgeDB v${{ steps.relver.outputs.version }}
```
elprans, line 382: Should probably include a list of major features and a link to the changelog. Not sure how to automate the feature list, but perhaps just edit manually for now.

Issue #7264, opened 2024-04-26
Migration creation allows invalid aggregate functions in constraints


jollygreenlaser:
- EdgeDB Version: 4.7+45e4acc
- EdgeDB CLI Version: EdgeDB CLI 5.0.0+c9dfeec
- OS Version: Ubuntu 22, OSX 14.4.1

```
type User {
   name: str;
}

type HasTwoOrFewerUsers {
   users: User;
   constraint expression on (count(.users) <= 2);
}
```

This will work to create a migration, but then fail to apply:

```
InternalServerError: aggregate functions are not allowed in check constraints
```

If `users` is updated to be correctly `multi` then it properly fails at migration creation time:

```
error: cannot use aggregate functions or operators in a non-aggregating constraint
```

Seems this should always be caught at creation time.

msullivan:
I thought this was going to have been fixed by https://github.com/edgedb/edgedb/pull/7128/, but it looks like we don't catch this at all, and the error comes from postgres.

The funny bit is that if `users` is actually a multi link, *then* we correctly reject it at migration create time!

dnwpark:
@msullivan Thoughts on adding whether the set op is permitted in singleton to the schema?

PR #7343, opened 2024-05-13
Catch aggregate calls in constraints and indexes in ql compiler

dnwpark:
related #7264

- Catches aggregate calls in constraints and indexes earlier in the compilation process
    - Raises errors in migration create which were previously only caught on migrate
- Fixes abstract constraints not raising errors when using aggregate functions
- Adds more information in errors about aggregate functions

Given the schema:
```
  type User {
     name: str;
  }

  type HasTwoOrFewerUsers {
     users: User;
     constraint expression on (count(.users) <= 2);
  }
```

The following error is produced:
```
error: cannot use aggregate functions or operators in a non-aggregating constraint
   │
25 │      constraint expression on (count(.users) <= 2);
   │                                ^^^^^^^^^^^^^ error
```

### DIFF
--- a/edb/ir/utils.py
+++ b/edb/ir/utils.py
@@ -48,6 +48,7 @@ from edb.common import ast
 from edb.common import ordered
 
 from edb.edgeql import qltypes as ft
+from edb.schema import name as sn
 
 from . import ast as irast
 from . import typeutils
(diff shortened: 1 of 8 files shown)


edb/ir/utils.py
```diff
@@ -489,13 +490,47 @@ def find_potentially_visible(
...
-        return any(
-            arg.param_typemod == ft.TypeModifier.SetOfType
-            for arg in n.args.values()
+        return (
+            (has_multi_param or not is_singleton_set_of_call(n))
+            and (has_set_of_param(n) or returns_set_of(n))
         )
-    return bool(ast.find_children(ir, irast.Call, flt, terminate_early=True))
+    calls = ast.find_children(ir, irast.Call, flt, terminate_early=True)
+    return next(iter(calls or []), None)
```
msullivan, line 533: Does the `bool` thing not actually work?

msullivan: Oh, you're trying to get the actual value, OK

edb/pgsql/compiler/expr.py
```diff
@@ -433,10 +433,16 @@ def compile_OperatorCall(
                 dispatch.compile(r_expr, ctx=ctx),
             ],
         )
-    elif expr.typemod is ql_ft.TypeModifier.SetOfType:
+    elif irutils.is_singleton_set_of_call(expr):
+        ...
```
msullivan, line 437: `pass` is traditional for an empty branch

edb/ir/utils.py
```diff
@@ -498,6 +499,49 @@ def flt(n: irast.Call) -> bool:
...
+    # Some set functions and operators are allowed in singleton mode
+    # as long as their inputs are singletons
+
+    return call.func_shortname in {
+        sn.QualName('std', 'IN'),
+        sn.QualName('std', 'NOT IN'),
+        sn.QualName('std', 'EXISTS'),
+        sn.QualName('std', '??'),
+        sn.QualName('std', 'IF'),
+    }
```
msullivan, line 505: We *could* make this a new flag on operators that we configure in the standard library creation code. No sure if it is worth it, though?

tests/test_edgeql_ddl.py
```diff
@@ -13217,16 +13217,28 @@ async def test_edgeql_ddl_index_07(self):
 
     async def test_edgeql_ddl_index_08(self):
         async with self.assertRaisesRegexTx(
-                edgedb.UnsupportedFeatureError,
-                'set returning operator std::DISTINCT is not supported '
-                'in singleton expressions'):
+                edgedb.SchemaDefinitionError,
+                "cannot use aggregate operator 'std::DISTINCT' "
```
msullivan, line 13221: we should say "SET OF operator", not "aggregate operator"

edb/schema/constraints.py
```diff
@@ -931,12 +931,20 @@ def _populate_concrete_constraint_attrs(
...
+                final_subjectexpr.irast,
+                has_any_multi,
+            ):
+                label = (
+                    'function'
+                    if isinstance(set_of_op, irast.FunctionCall) else
+                    'operator'
+                )
+                op_name = str(set_of_op.func_shortname)
+                raise errors.UnsupportedFeatureError(
```
msullivan, line 944: We should say SET OF instead of aggregate.

Also, I know I wrote the original error message, but I have no idea what "non-aggregating constraint" means, so we should stop saying it. :P

msullivan (approved):

PR #7345, opened 2024-05-14
Docs: Add notes about Cloud authentication in scripts

raddevon:


docs/guides/cloud/cli.rst
```diff
@@ -7,9 +7,21 @@ CLI
...
-:ref:`ref_cli_edgedb_cloud_login`. Once your login is successful, you will be
-able to create an instance using either :ref:`ref_cli_edgedb_instance_create`
-or :ref:`ref_cli_edgedb_project_init`, depending on whether you also want to
+:ref:`ref_cli_edgedb_cloud_login`.
+
+.. note::
+
+    This is the way you'll log in interactively on your development machine,
+    but when interacting with EdgeDB Cloud via a script or in CI, you'll
+    instead set the ``EDGEDB_SECRET_KEY`` environment variable to your secret
```
scotttrinh, line 16: Probably want to mention setting the `EDGEDB_INSTANCE` environment variable here, too?

raddevon: I guess I could. The reason I didn't to start is that I'm explaining that you shouldn't use `edgedb cloud login`, so I wanted to explain the alternative to that. `edgedb cloud login` doesn't do anything to define an instance, so you'll still need to do that with any other command as you would have even if you had used `edgedb cloud login`. Do you think it's relevant here?

And if so, maybe it just needs to go in each of these notes?

scotttrinh (approved):
Looks good, as discussed! 🚢

PR #7284, opened 2024-04-30
Make linkprops (including @source/@target) work in conflict selects

msullivan:
Exclusive constraints on the linkprops themselves still don't work,
though it should be fixable.

Fixes #7263

PR #7349, opened 2024-05-14
Fix an UNLESS CONFLICT on links perf regression

msullivan:
The regression was introduced in #7284, which made
`utils.subject_substitute` substitute `__subject__` in for the
implicit path prefix. That fixed an issue in some cases, but
introduced an extra level of Set/SelectStmt indirection in
others.

That level of indirection could cause us to emit redundant semijoins
in some cases, which caused a major performance regression on UNLESS
CONFLICT on object level exclusive constraints that referred to a link
using a partial path prefix. (Referring to the link using
`__subject__.` was already slow!)

Fix that by making `subject_substitute` smarter.

The redundant semi-join issue can come up in other cases too, like:
```
select (User, (User,).0.deck);
```

### DIFF
--- a/edb/edgeql/utils.py
+++ b/edb/edgeql/utils.py
@@ -163,7 +163,7 @@ def find_subject_ptrs(ast: qlast.Base) -> Set[str]:
     return ptrs
 
 
-def is_anchor(expr: qlast.PathElement, name: str):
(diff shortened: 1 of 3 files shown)


dnwpark (approved):
Looks good.

Feels bad that avatars are first come first serve.

elprans (approved):

PR #7361, opened 2024-05-17
Docs: Add Auth integration to the fastapi guide

beerose:


docs/guides/tutorials/rest_apis_with_fastapi.rst
```diff
@@ -1167,16 +1170,403 @@ payload:
...
+    # Value should be:
+    # {protocol}://${host}:${port}/branch/${branch}/ext/auth/
+    EDGEDB_AUTH_BASE_URL = os.getenv('EDGEDB_AUTH_BASE_URL')
+
+    @router.post("/auth/signup")
+    async def handle_signup(request: Request):
+        body = await request.json()
+        email = body.get("email")
+        password = body.get("password")
+        name = body.get("name")
```
scotttrinh, line 1351: I don't think we can do anything with `name` here, or maybe this is supposed to be `provider`?

docs/guides/tutorials/rest_apis_with_fastapi.rst
```diff
@@ -1167,16 +1170,403 @@ payload:
...
+        if not email or not password or not name:
+            raise HTTPException(status_code=400, detail="Missing email, password, or name.")
+
+        verifier, challenge = generate_pkce()
+        register_url = f"{EDGEDB_AUTH_BASE_URL}/register"
+        response = httpx.post(register_url, json={
+            "challenge": challenge,
+            "email": email,
+            "password": password,
+            "provider": provider,
```
scotttrinh, line 1362: Where does `provider` come from here? I think we can hardcode this as `builtin::local_emailpassword` since this endpoint just deals with email+password sign up.

docs/guides/tutorials/rest_apis_with_fastapi.rst
```diff
@@ -1167,16 +1170,403 @@ payload:
...
+        email = body.get("email")
+        password = body.get("password")
+        name = body.get("name")
+
+        if not email or not password or not name:
+            raise HTTPException(status_code=400, detail="Missing email, password, or name.")
+
+        verifier, challenge = generate_pkce()
+        register_url = f"{EDGEDB_AUTH_BASE_URL}/register"
+        response = httpx.post(register_url, json={
```
scotttrinh, line 1358: Maybe this should be `register_response` to avoid conflicts with the `response` that is returned from this handler later in this function?

docs/guides/tutorials/rest_apis_with_fastapi.rst
```diff
@@ -1167,16 +1170,403 @@ payload:
...
+            "challenge": challenge,
+            "email": email,
+            "password": password,
+            "provider": provider,
+            "verify_url": "http://localhost:8000/auth/verify",
+        })
+
+        if response.status_code != 200 and response.status_code != 201:
```
scotttrinh, line 1366: We need to look at the result from the http request and see if it contains a `code`. If so, we need to do a code exchange, like with sign-in. I totally left this part out of the Auth Guide on accident, so we should update both 🙈

docs/guides/tutorials/rest_apis_with_fastapi.rst
```diff
@@ -1167,16 +1170,403 @@ payload:
...
+            return JSONResponse(status_code=400, content={"message": "Registration failed"})
+
+        response = JSONResponse(content={"message": "User registered"})
+        response.set_cookie(key="edgedb-pkce-verifier", value=verifier, httponly=True, secure=True, samesite='strict')
+        return response
+
+The sign-up endpoint sends a POST request to the EdgeDB Auth server to register
+a new user. It also generates a PKCE verifier and sets it as an HttpOnly cookie
+in the response.
```
scotttrinh, line 1375: It doesn't need to set the verifier cookie in this endpoint because the whole PKCE session happens within this one endpoint, but we _do_ set the auth token cookie here.

docs/guides/tutorials/rest_apis_with_fastapi.rst
```diff
@@ -1167,16 +1170,403 @@ payload:
...
+
+First, let's update the EdgeDB schema to include a new field in the User type
+to store the EdgeDB Auth identity and a new ``current_user`` type.
+
+.. code-block:: sdl-diff
+    :caption: dbschema/default.esdl
+    
+    + global current_user := assert_single(
+    +     ((
+    +         select User { id, name }
```
scotttrinh, line 1445: Minor nit: Remove this `{ id, name }` shape from the select: the shape is used when selecting from the global, not in the global definition itself.

docs/guides/tutorials/rest_apis_with_fastapi.rst
```diff
@@ -1167,16 +1170,403 @@ payload:
...
+
+      select (
+    +     with identity := (select ext::auth::Identity filter .id = <uuid>$identity_id),
+          insert User {
+            name := <str>$name,
+    +       identity := identity
+          }) {
+          name,
+          created_at,
+      };
```
scotttrinh, line 1478: Maybe prefer the UUID->Object cast style here?

```suggestion
      select (
          insert User {
            name := <str>$name,
    +       identity := <ext::auth::Identity><uuid>$identity_id,
          }) {
          name,
          created_at,
      };
```

docs/guides/tutorials/rest_apis_with_fastapi.rst
```diff
@@ -1167,16 +1170,403 @@ payload:
...
+      
+          if response.status_code != 200 and response.status_code != 201:
+              return JSONResponse(status_code=400, content={"message": "Registration failed"})
+          
+    +     code = response.json().get("code")
+    +     token_url = f"{EDGEDB_AUTH_BASE_URL}/token"
+    +     token_response = httpx.get(token_url, params={"code": code, "verifier": verifier})
+    + 
+    +     if token_response.status_code != 200:
+    +         return JSONResponse(status_code=400, content={"message": "Token exchange failed"})
```
scotttrinh, line 1526: Oh, I see we're doing the code exchange here! We should move it up to the previous endpoint definition and add the token cookie rather than the verifier cookie.

docs/guides/tutorials/rest_apis_with_fastapi.rst
```diff
@@ -1167,16 +1170,403 @@ payload:
...
+    +     try:
+    +         created_user = await create_user_qry.create_user(client, name=name, identity_id=identity_id)
+    +     except edgedb.errors.ConstraintViolationError:
+    +         raise HTTPException(
+    +             status_code=400,
+    +             detail={"error": f"User with email '{email}' already exists."},
+    +         )
+              
+          response = JSONResponse(content={"message": "User registered"})
+          response.set_cookie(key="edgedb-pkce-verifier", value=verifier, httponly=True, secure=True, samesite='strict')
```
scotttrinh, line 1538: This should be the auth token cookie instead of the verifier cookie.

docs/guides/tutorials/rest_apis_with_fastapi.rst
scotttrinh: Ahh, I see we use this later on when creating the user. Might make more sense to remove it here and add it as part of the later diff?

scotttrinh (approved):
Looks great!

PR #7374, opened 2024-05-21
Raise error when computed global is set or reset.

### LINKED ISSUE #7368: Don't allow setting a global if it was declared with a computable expression
Schema:

```
module default {
    global test := 'hello';
}
```

```
> set global test := 'world'; <--- this should error out, but it does not
OK

> select global test;
{'hello'}    <--- because this is confusing
```

dnwpark:
After `set global foo := false;`, attempting to run `set global lol := false;` produces the following error:
```
error: ConfigurationError: global 'foo' is computed from an expression and cannot be modified
  ┌─ <query>:1:12
  │
1 │ set global foo := false;
  │            ^^^ error
```

close #7368

### DIFF
--- a/edb/edgeql/compiler/config.py
+++ b/edb/edgeql/compiler/config.py
@@ -292,10 +292,19 @@ def _validate_config_object(
 def _validate_global_op(
     expr: qlast.ConfigOp, *, ctx: context.ContextLevel
 ) -> SettingInfo:
+    glob_name = s_utils.ast_ref_to_name(expr.name)
     glob = ctx.env.get_schema_object_and_track(
-        s_utils.ast_ref_to_name(expr.name), expr.name,
+        glob_name, expr.name,
         modaliases=ctx.modaliases, type=s_globals.Global)
     assert isinstance(glob, s_globals.Global)
+
+    if isinstance(expr, (qlast.ConfigSet, qlast.ConfigReset)):
+        if glob.get_expr(ctx.env.schema):
+            raise errors.ConfigurationError(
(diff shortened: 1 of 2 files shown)


edb/edgeql/compiler/config.py
```diff
@@ -292,11 +292,20 @@ def _validate_config_object(
...
     glob = ctx.env.get_schema_object_and_track(
-        s_utils.ast_ref_to_name(expr.name), expr.name,
+        glob_name, expr.name,
         modaliases=ctx.modaliases, type=s_globals.Global)
     assert isinstance(glob, s_globals.Global)
 
+    if isinstance(expr, (qlast.ConfigSet, qlast.ConfigReset)):
+        if glob.get_expr(ctx.env.schema):
+            raise errors.ConfigurationError(
+                f"global '{glob_name}' is computed from an expression and "
```
msullivan, line 304: I'd probably just say "computed global 'whatever' may not be modified" or some such

msullivan (approved):
This seems good for `set global`. We want to also check what happens when the client specifies the values via the protocol

dnwpark:
@msullivan I saw the issue mentioned that. With these changes, it seems that setting the global silently deletes it from the state.

msullivan:
> @msullivan I saw the issue mentioned that. With these changes, it seems that setting the global silently deletes it from the state.

I'm not sure what you mean by that

msullivan (approved):

dnwpark:
Also added a test to the python client here: https://github.com/edgedb/edgedb-python/pull/494

PR #7375, opened 2024-05-22
Fix make sure collection types land in env.created_schema_types

aljazerzen:


aljazerzen:
@msullivan The problem with the `invalid syntax for type uuid: "None"` was that we were trying to re-create collection types. I'm not really sure which field was passing `None` for uuid. Even weirder, I got this message: `modification of computed property 'backend_id' is prohibited`, which is even weirder because that originates from the frontend compiler, which is probably being invoked from reflection.

Regardless, it is all fixed now. Should be safe to backport without any patches.

tests/test_edgeql_ddl.py
```diff
@@ -14503,7 +14503,7 @@ async def test_edgeql_ddl_collection_cleanup_05(self):
 
         self.assertEqual(
             await self.con.query_single(count_query),
-            orig_count + 1,
+            orig_count + 2,  # one for tuple<str, str>, one for TupleExprAlias
```
msullivan, line 14506: This change is actually a backporting problem, since it changes the schema for existing working cases:
```
module default {
    scalar type a extending str;
    scalar type b extending str;
    alias Bar := (<a>"", <b>"");
}
```
If you migrate to this schema on master, then apply this PR, and try to create migration again, it will fail with no proposed DDL but a schema mismatch, which is a pretty bad state.

Fortunately, `administer schema_repair();` can fix this one.

So our options are:
 * Find a way to avoid changing the behavior in cases that currently work
 * Always run a schema repair as part of the 5.4 patching
 * Figure this one will hit few enough users that we can just document `administer schema_repair();` in the changelog if they hit the problem?

Do you have thoughts on how possible the first option is?

aljazerzen: Option 3 is off the charts, as this will happen for anyone who has an alias or a computed global that contains a collection type. My guess would be that 25% of all edgedb databases qualify for this.

---

Old behavior:
- a single type is generated,
- TupleExprAlias with name `default::Bar`

New behavior:
- two types are generated,
- TupleExprAlias with name `default::Bar`,
- Tuple with name `tuple<default|a, default|b>` (this tuple is the material type of the TupleExprAlias)

I could make mechanism where we don't explicitly create any type that is material types of some other type.

msullivan (approved):
All right. Let's land this and do a repair patch.
Make sure to bump the catalog version.

tests/test_edgeql_ddl.py
aljazerzen: On the call we decided to get with option 2, because it is more correct and less likely to cause problems than option 1.

msullivan:
Oh, I had meant to say, but forgot: make the commit message be something that explains what the bug fixed is, not how it was fixed. This makes it easier to produce a useful changelog message.

msullivan:
> @msullivan The problem with the `invalid syntax for type uuid: "None"` was that we were trying to re-create collection types. I'm not really sure which field was passing `None` for uuid. 

It was `id`

> Even weirder, I got this message: `modification of computed property 'backend_id' is prohibited`, which is even weirder because that originates from the frontend compiler, which is probably being invoked from reflection.

Yeah, reflection invokes the frontend compiler, but that failure is pretty weird. `backend_id` is *not* computed, but I think I remember some bug where we produce a message like that incorrectly in some broken update case...

PR #7376, opened 2024-05-22
Docs: Add note about config timeout values

raddevon:
I could have explained that values under 501 microseconds will work in the REPL, but I figured better to keep it simple and just tell users not to use values under 1ms.

### DIFF
--- a/docs/reference/configuration.rst
+++ b/docs/reference/configuration.rst
@@ -124,14 +124,22 @@ Client connections
   while in a transaction. Defaults to 10 seconds; set to ``<duration>'0'`` to
   disable the mechanism.
 
+  .. note::
+
+      For ``session_idle_transaction_timeout`` and ``query_execution_timeout``,
+      values under 1ms are rounded down to zero, which will disable the timeout.
+      In order to set a timeout, please set a duration of 1ms or greater.
+
+      ``session_idle_timeout`` can take values below 1ms.
+
 :eql:synopsis:`query_execution_timeout -> std::duration`
   How long an individual query can run before being aborted. A value of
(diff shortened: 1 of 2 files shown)


docs/reference/configuration.rst
```diff
@@ -127,3 +127,11 @@ Client connections
...
   How long an individual query can run before being aborted. A value of
   ``<duration>'0'`` disables the mechanism; it is disabled by default.
+
+.. note::
+
+    For ``session_idle_transaction_timeout`` and ``query_execution_timeout``,
+    values under 1ms are rounded down to zero, which will disable the timeout.
+    In order to set a timeout, please set a duration of 1ms or greater.
+
+    ``session_idle_timeout`` can take values below 1ms.
```
fmoor, line 137: I could see missing this note if I was only looking for `session_idle_transaction_timeout` documentation. Its repetitive but maybe this should just be part of the documentation and duplicated for each config value?

docs/stdlib/cfg.rst
```diff
@@ -243,6 +243,14 @@ Client connections
...
   The timeout isn't enabled by default.
 
+.. note::
+
+    For ``session_idle_transaction_timeout`` and ``query_execution_timeout``,
+    values under 1ms are rounded down to zero, which will disable the timeout.
+    In order to set a timeout, please set a duration of 1ms or greater.
+
+    ``session_idle_timeout`` can take values below 1ms.
+
```
fmoor, line 253: same here

fmoor (approved):

docs/reference/configuration.rst
raddevon: Yeah, I like that idea.

PR #7377, opened 2024-05-22
Patch for 5.4

vpetrovykh:
Looks like I managed to adjust the commits so that the tests pass locally.


vpetrovykh:
#7278 and #7310 are the two commits that were causing errors, so they are the ones worth double-checking in terms of me manually resolving the conflicts there.

msullivan (approved):

PR #7385, opened 2024-05-23
Fix single-tenant metrics not to filter by tenant

fantix:
This is a bug introduced in #7024 that `/metrics` returns different results using HTTP or HTTPS protocol under single-tenant mode. This PR fixes so that `/metrics` never filter by tenant under single-tenant mode.

This PR also fixes the test to `fetch_metrics()` with TLS certificate and adds a test to make sure `edgedb_server_compiler_processes_current` - as an example - is not filtered out.

### DIFF
--- a/edb/common/prometheus.py
+++ b/edb/common/prometheus.py
@@ -409,7 +409,6 @@ class BaseLabeledCounter(BaseMetric):
             fmt_label = ','.join(
                 f'{label}="{_format_label_val(label_val)}"'
                 for label, label_val in zip(self._labels, labels)
-                if label not in label_filters
             )
             buffer.append(
                 f'{self._name}{self._suffix}{{{fmt_label}}} {float(value)}'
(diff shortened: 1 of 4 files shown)


msullivan (approved):

msullivan:
Though there is also the question of: should we be removing the tenant label when filtering by tenant at all?

zackelan:
> should we be removing the tenant label when filtering by tenant at all?

I would prefer not to - if querying metrics for tenant X does not include the tenant label, then at the point where we scrape the metrics, we would have to go out of our way to add the tenant label back in, so that in our metrics storage system per-tenant metrics are queryable. since the server already knows the per-tenant metrics, there's very little point in stripping the label out only to immediately add it back.

zackelan (approved):

fantix:
Sounds good, I'll add the label back.

PR #7386, opened 2024-05-23
server: expose /server/branches system API

fantix:
This endpoint lists all branches in the current tenant, or such of all tenants if queried by IP of a multi-tenant server.

This endpoint takes additional options for extra branch information, like `/server/branches/schema-version` will include loaded schema version UUID for each branch.

This endpoint must be protected by a non-trust authentication method, ~which is only mTLS for now~. For example:

```
edgedb-server ... --default-auth-method http_health:mtls/trust --tls-client-ca-file tests/certs/client_ca.cert.pem
```

This allows unauthorized access to `/server/status/alive` and `/server/status/ready`, but mTLS is required to access `/server/branches/...`. For example:

```
$ curl -k  https://127.0.0.1:5656/server/branches/schema-version | jq
{
  "error": {
    "message": "this endpoint requires authentication",
    "type": "AuthenticationError",
    "code": 117506048
  }
}
```

(multi-tenant)

```
$ curl -k --cert tests/certs/client.cert.pem --key tests/certs/client.key.pem https://127.0.0.1:5656/server/branches/schema-version  | jq
{
  "local/tt2": [
    {
      "name": "main",
      "schema_version": "32e7161c-17ac-11ef-8dd8-75fd3d451c70"
    }
  ],
  "localtest": [
    {
      "name": "main",
      "schema_version": null
    }
  ]
}
```

(selected tenant)

```
$ curl -k --cert tests/certs/client.cert.pem --key tests/certs/client.key.pem https://tt2.local:5656/server/branches/schema-version  | jq
[
  {
    "name": "main",
    "schema_version": "32e7161c-17ac-11ef-8dd8-75fd3d451c70"
  }
]
```

UPDATE from #7622: we use `__edgedbsys__` as the "database" to check the authorization of the JWT token. This change also applies to `/server/*` URLs.


jaclarke:
Can it also work with jwt auth? I don't it's possible to do mTLS from the browser.

fantix:
Close in favor of #7654

PR #7388, opened 2024-05-23
[5.x] Fix make sure collection types land in env.created_schema_types

msullivan:


vpetrovykh (approved):
Doesn't look suspicious and passes the tests.

PR #7397, opened 2024-05-24
Add array index number to output when casting from json to array.

dnwpark:
Given the query `SELECT <array<int64>>to_json('[1, 2, "asdf"]')`, produces the error:
```
InvalidValueError: while casting 'std::json' to 'array<std::int64>', in array elements, at index 2, expected JSON number or null; got JSON string
```

### DIFF
--- a/edb/edgeql/compiler/casts.py
+++ b/edb/edgeql/compiler/casts.py
@@ -453,7 +453,7 @@ def _cast_to_ir(
         sql_function=cast.get_from_function(ctx.env.schema),
         sql_cast=cast.get_from_cast(ctx.env.schema),
         sql_expr=bool(cast.get_code(ctx.env.schema)),
-        error_message_context=cast_message_context(ctx),
+        error_message_context=get_error_message_context_ir(ctx),
     )
 
     return setgen.ensure_set(cast_ir, ctx=ctx)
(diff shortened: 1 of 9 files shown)


dnwpark:
Adding `error_message_context` in `fini_expression` is super dubious imo

msullivan:
> Adding `error_message_context` in `fini_expression` is super dubious imo

Why is that necessary?

dnwpark:
In `_compile_call_args` there is a check to ensure that all args have a known multiplicity. The multiplicity is added in `fini_expression`. I suppose the other alternative is to add an override to `infer_multiplicity` on `TypeCast`.

msullivan:
Yeah, adding it to multiplicity and cardinality checks is the correct thing

dnwpark:
Benchmarking against three queries:

**Simple null** - 10000 times
```
client.query(
    "SELECT <array<int64>>to_json(<str>$input)",
    input='[null]'
)
```
- `master`: `8.927s`
- With array indexes: `13.453s`

**Long array** - 1000 times
```
client.query(
    "SELECT <array<int64>>to_json(<str>$input)",
    input='['+','.join(str(i) for i in range(1000))+',null'+']'
)
```
- `master`: `4.647s`
- With array indexes: `10.134s`

**Long and deep array** - 1000 times
```
client.query(
    "SELECT <array<tuple<array<tuple<array<int64>>>>>>to_json(<str>$input)",
    input='['+','.join('[[[[1,2,3],[4,5,6],[7,8,9]]]]' for i in range(1000))+',[[[[1,2,null]]]]'+']'
)
```
- `master`: `9.406s`
- With array indexes: `90.841s`

dnwpark:
@msullivan non-trivial cases seem convincingly much slower.

msullivan:
Great work experimenting with this, though!

msullivan:
Also, we support `json` arguments directly too

PR #7404, opened 2024-05-28
Add is_singleton_set_of to schema.

### LINKED ISSUE #7264: Migration creation allows invalid aggregate functions in constraints

dnwpark:
SET OF functions which are allowed in singleton expressions were previously listed by name within the code. 

close #7264

### DIFF
--- a/edb/buildmeta.py
+++ b/edb/buildmeta.py
@@ -60,7 +60,7 @@ from edb.common import verutils
 # The merge conflict there is a nice reminder that you probably need
 # to write a patch in edb/pgsql/patches.py, and then you should preserve
 # the old value.
-EDGEDB_CATALOG_VERSION = 2024_06_06_00_00
+EDGEDB_CATALOG_VERSION = 2024_06_14_00_00
 EDGEDB_MAJOR_VERSION = 6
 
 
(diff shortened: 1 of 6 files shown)


msullivan:
This is what I was thinking of, yeah. Though I forgot that we'd need to plumb it into `irast.Call`.
The question now is whether this is actually better.

dnwpark:
@msullivan I suppose this is technically cleaner, since it keeps the "configuration" of a function separate from how it's used.

In terms of the test errors i mentioned, I've been needing to re-run all tests several times to get the runs to finish consistently. There wasn't any particular error, just that one or more of the test jobs got cancelled.

msullivan (approved):

PR #7405, opened 2024-05-28
Fix dumps of types that have both AI and FTS indexes

msullivan:
When adding "fake" columns to the schema for the dump descriptor
for the shadow columns for AI and FTS indexes, we used the wrong
version of the immutable schema object, so only added the last
such column to the schema.

### DIFF
--- a/edb/server/compiler/compiler.py
+++ b/edb/server/compiler/compiler.py
@@ -3149,7 +3149,7 @@ def _describe_object(
         # and include them in the dump.
         nschema = schema
         for (name, col, _type) in source.get_addon_columns(schema):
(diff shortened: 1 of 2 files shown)


elprans (approved):

edb/server/compiler/compiler.py
```diff
@@ -3149,7 +3149,7 @@ def _describe_object(
         # and include them in the dump.
         nschema = schema
         for (name, col, _type) in source.get_addon_columns(schema):
-            nschema, fake_ptr = _add_fake_property(source, name, schema)
+            nschema, fake_ptr = _add_fake_property(source, name, nschema)
```
aljazerzen, line 3152: Oh, god, what an easy mistake to make.

PR #7407, opened 2024-05-29
Rename interpreter tests

aljazerzen:
I often run a single test by copy pasting its name and it is annoying that two other tests also run because of that.


zchenb:
Interpreter tests are already prefixed by `test_edgeql_select_interpreter_`, may I ask why two other tests are run?

aljazerzen:
Oh, I didn't even notice - you are right.

I guess it was just `test_edgeql_with_rebind_01` that was not prefixed correctly

PR #7422, opened 2024-06-03
Fix PgFunc compiled query cache

fantix:
This PR fixes the `PgFunc` query cache mode and sets it as the default mode.

The `QueryUnit` under `PgFunc` was directly a function call, but we deferred the creation of the function into an asynchronous task, so the cache was not immediately available; it was failing because of this.

This PR first modified the `QueryUnit` to include the compiled SQL that is immediately available ("inline" SQL) and the function-call SQL that replaces the former immediately after we know the function is created. This resolves the error above.

However, there are 4 more issues with the functions in transactions.

1. The function creation was queued and executed as soon as the compilation was completed, even when it was in a transaction with DDL applied. The function creation may fail because it may refer to the new types that only exist in the transaction (is this true? update: yes for custom scalars in function signatures). ~The solution here is to store the query cache within the transaction first and push it up to the database-global cache registration for persistence once the transaction is committed.~ Queries in transactions after DDL are not cached now.

2. A function cache created after a transaction starts can be picked up by the transaction. In this case, we may not be able to call the function due to transaction isolation. ~To address this issue, this PR creates a snapshot of the cache registry when a transaction starts and only uses the cache that was available at the start of the transaction.~ This PR tracks the active transactions and the relative sequence of when each cache function is actually created in DB. The function cache is only activated after all transactions that can't see the function are closed. After activation, the "inline" SQL is dropped to save memory usage. UPDATE: Before activation, transactions started after the function creation could also use the function cache (last CRF).

3. As Sully mentioned below, scalar types are sometimes referred in cache function signatures, like `select <array<MyStr>>$0`. This is fixed by evicting all depending query cache when running DDL to drop such scalar types.

4. We're generating SQL functions that `returning setof record[]` which is an error in Postgres. This is reproducible with EdgeQL queries like `select [(1,),(2,)]`. Such errors are currently ignored (printing a warning) and the corresponding query cache falls back to `InMemory`.

### DIFF
--- a/edb/pgsql/delta.py
+++ b/edb/pgsql/delta.py
@@ -2879,14 +2879,74 @@ class DeleteScalarType(ScalarTypeMetaCommand,
     def delete_scalar(
         cls, scalar: s_scalars.ScalarType, orig_schema: s_schema.Schema
     ) -> dbops.Command:
+        ops = dbops.CommandGroup()
         old_domain_name = common.get_backend_name(
             orig_schema, scalar, catenate=False)
 
+        drop_func_cache_sql = textwrap.dedent(f'''
+            DO $$
+            DECLARE
+                query_cache_row RECORD;
+            BEGIN
+                FOR query_cache_row IN
+                    WITH
+                    types AS (
+                        SELECT
+                            pt.oid AS oid
+                        FROM
+                            pg_type pt
+                            JOIN pg_namespace pn
+                                ON pt.typnamespace = pn.oid
+                        WHERE
+                            pn.nspname = '{old_domain_name[0]}'
+                            AND (
+                                pt.typname = '{old_domain_name[1]}'
+                                OR pt.typname = '_{old_domain_name[1]}'
+                            )
+                    ),
+                    class AS (
+                        SELECT
+                            pc.oid AS oid
+                        FROM
+                            pg_class pc
+                            JOIN pg_namespace pn
+                                ON pc.relnamespace = pn.oid
+                        WHERE
+                            pn.nspname = 'pg_catalog'
+                            AND pc.relname = 'pg_type'
+                    ),
+                    funcs AS (
+                        SELECT
+                            substring(p.proname FROM 6)::uuid AS key
+                        FROM
+                            pg_proc p
+                            JOIN pg_depend d
+                                ON d.objid = p.oid
+                            JOIN types t
+                                ON d.refobjid = t.oid
+                            JOIN class c
+                                ON d.refclassid = c.oid
+                        WHERE
+                            p.proname LIKE '__qh_%'
+                    )
+                    SELECT
+                        c.evict
+                    FROM
+                        funcs f
+                        LEFT JOIN edgedb._query_cache c
+                            ON f.key = c.key
+                    WHERE
+                        c.evict IS NOT NULL
+                LOOP
+                    EXECUTE query_cache_row.evict;
+                END LOOP;
+            END $$;
+        ''')
+        ops.add_command(dbops.Query(drop_func_cache_sql))
+
         cond: dbops.Condition
         if scalar.is_concrete_enum(orig_schema):
-            ops = dbops.CommandGroup()
-            old_enum_name = common.get_backend_name(
-                orig_schema, scalar, catenate=False)
+            old_enum_name = old_domain_name
             cond = dbops.EnumExists(old_enum_name)
 
             cast_func_name = common.get_backend_name(
@@ -2911,10 +2971,13 @@ class DeleteScalarType(ScalarTypeMetaCommand,
 
             enum = dbops.DropEnum(name=old_enum_name, conditions=[cond])
             ops.add_command(enum)
-            return ops
+
         else:
             cond = dbops.DomainExists(old_domain_name)
-            return dbops.DropDomain(name=old_domain_name, conditions=[cond])
+            domain = dbops.DropDomain(name=old_domain_name, conditions=[cond])
+            ops.add_command(domain)
+
+        return ops
 
     def apply(
         self,
--- a/edb/pgsql/delta.py
+++ b/edb/pgsql/delta.py
@@ -2883,12 +2883,16 @@ class DeleteScalarType(ScalarTypeMetaCommand,
         old_domain_name = common.get_backend_name(
             orig_schema, scalar, catenate=False)
 
+        # The custom scalar types are sometimes included in the function
+        # signatures of query cache functions under QueryCacheMode.PgFunc.
+        # We need to find such functions through pg_depend and evict the cache
+        # before dropping the custom scalar type.
         drop_func_cache_sql = textwrap.dedent(f'''
             DO $$
             DECLARE
-                query_cache_row RECORD;
+                qc RECORD;
             BEGIN
-                FOR query_cache_row IN
+                FOR qc IN
                     WITH
                     types AS (
                         SELECT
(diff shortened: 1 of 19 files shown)


edb/server/compiler/dbstate.py
```diff
@@ -220,6 +224,16 @@ class QueryUnit:
...
 
+    # cache_func_call stores the calling SQL of the cache function
+    # if QueryCacheMode is set to PgFunc.
+    cache_func_call: Optional[Tuple[bytes, bytes]] = None  # (sql, hash)
+
+    # cache_inline is an in-memory temporary reservation of the inlined
+    # query cache when cache_func_call is applied in self.sql and sql_hash,
+    # used by transactions that don't yet have the visibility of the newly
+    # persisted query cache outside the transaction.
+    cache_inline: Optional[Tuple[bytes, bytes]] = None  # (sql, hash)
```
msullivan, line 235: Do we ever purge the `cache_inline` data?
One of the goals for the function cache is to avoid the memory usage of caching everything in memory, right?
(Though we might not want to purge it *immediately* after creating the function, because other transactions might want to grab it?)

msullivan:
What are the still known issues?
- [ ] Deleting scalar types that are used in a cached function
- [ ] Queries with 1000 (or something like that?) parameters

edb/server/dbview/dbview.pyx
```diff
@@ -837,15 +843,45 @@ cdef class DatabaseConnectionView:
...
+                not self._tx_error and
+                self._query_cache_enabled and
+                not self._in_tx_with_ddl and
+                key not in self._in_tx_eql_to_compiled
+            ):
+                self._in_tx_eql_to_compiled[key] = query_unit_group
+                while self._in_tx_eql_to_compiled.needs_cleanup():
+                    self._in_tx_eql_to_compiled.cleanup_one()
+        else:
+            self._db._cache_compiled_query(key, query_unit_group)
```
msullivan, line 857: Is this the behavior we want?

For transactions without any DDL, is there anything stopping us from calling `_cache_compiled_query` immediately?

msullivan:
One other thing we need to figure out: how do we properly test this? 
Most of our test suite runs queries inside transactions, which means the functions won't actually get called.
Worse, I think with this current version, since the transactions don't get committed, most of the functions generated as part of the test suite won't even get *created*.

(Though on the flip side, maybe that's good for performance...)

edb/server/compiler/dbstate.py
fantix: > Do we ever purge the `cache_inline` data?

No, not yet

> ... because other transactions might want to grab it?

Yes, that's my understanding.

I think I can add at least 2 changes to purge this data:

1. Purge in `hydrate_cache()`, this is usually done at server start;
2. Add a new cache registry to store the "inline" cache that rolls forward by some internally-tracked incremental-only transaction ID.

edb/server/dbview/dbview.pyx
fantix: Now that I'm reviewing this, perhaps we can still cache the entries compiled in transactions (without DDL for now). Let me try that with a new way to deal with the inline cache.

edb/server/compiler/dbstate.py
msullivan: We could also prune after some amount of real wall-clock time, which would probably be fine. (Though would sometimes result in having a cache miss in a tx when the inline version is missing)

edb/pgsql/delta.py
```diff
@@ -2909,6 +2909,64 @@ def delete_scalar(
...
+                        WITH
+                        types AS (
+                            SELECT
+                                pt.oid AS oid
+                            FROM
+                                pg_type pt
+                                JOIN pg_namespace pn
+                                    ON pt.typnamespace = pn.oid
+                            WHERE
+                                pn.nspname = '{old_enum_name[0]}'
```
msullivan, line 2927: Use the `ql` function to quote the literals

edb/pgsql/delta.py
```diff
@@ -2909,6 +2909,64 @@ def delete_scalar(
...
+                        SELECT
+                            c.evict
+                        FROM
+                            funcs f
+                            LEFT JOIN edgedb._query_cache c
+                                ON f.key = c.key
+                        WHERE
+                            c.evict IS NOT NULL
+                    LOOP
+                        EXECUTE query_cache_row.evict;
```
msullivan, line 2964: I really like that we can do this by executing the evict function

edb/pgsql/delta.py
```diff
@@ -2879,14 +2879,71 @@ class DeleteScalarType(ScalarTypeMetaCommand,
     def delete_scalar(
         cls, scalar: s_scalars.ScalarType, orig_schema: s_schema.Schema
     ) -> dbops.Command:
+        ops = dbops.CommandGroup()
         old_domain_name = common.get_backend_name(
             orig_schema, scalar, catenate=False)
 
+        drop_func_cache_sql = textwrap.dedent(f'''
```
msullivan, line 2886: Add a comment about what we are up to here, too

edb/pgsql/delta.py
fantix: Oh yes I should just use the `edgedb."_evict_query_cache"()` function!

edb/server/dbview/dbview.pyx
```diff
@@ -218,18 +233,81 @@ cdef class Database:
...
+        for query_req, units in self._func_cache_gt_tx_seq.items():
+            if query_req is None:
+                # Hit a wrapping marker, grab the real oldest active tx ID
+                seq = next(iter(self._active_tx_list.keys()))
+                drops.append(None)
+            elif units.tx_seq_id < seq:
+                units[0].maybe_use_func_cache()
+                drops.append(query_req)
+            else:
+                break
```
msullivan, line 307: Instead of waiting for every finish transaction to drop before we start using the function, could we track the transaction id in the cache and start using it in later transactions even while old ones are still pending?

I think it probably doesn't matter *that* much for real-world use cases, but it will play a lot nicer with an idea for testing I want to do.

The testing idea is that we can have a test mode that runs each query test twice, with the idea that it will hopefully hit the cache on the second loop, and I think all the extra transactions are likely to cause some trouble.

msullivan: Actually, maybe that test idea won't work that well anyway, since it often won't have been written in time for the second loop anyway...

Maybe the strategy should be to run a full test suite, then run the full suite a second time now that the cache is populated.

fantix: > Instead of waiting for every finish transaction to drop before we start using the function, could we track the transaction id in the cache and start using it in later transactions even while old ones are still pending?

Yeah, I thought about this. Without changing the QueryUnit data structure, I think I can also patch pgcon.pyx to take an additional param to use func call.

> Maybe the strategy should be to run a full test suite, then run the full suite a second time now that the cache is populated.

This sounds more reasonable to me.

msullivan (commented):
I just thought of one last problem: when we `hydrate_cache` because of a signal from another frontend, we will run into the same sorts of transaction visibility issues. Part of the solution is probably to track the txid we hydrated at, but we'll need to deal with not having SQL available, I think?

fantix:
> when we `hydrate_cache` because of a signal from another frontend, we will run into the same sorts of transaction visibility issues. Part of the solution is probably to track the txid we hydrated at,

If I understand it correctly, this is already taken care of by [not using the function](https://github.com/edgedb/edgedb/pull/7422/files#diff-1503bde887dee692dada5bf2b3a8e924233fabc6df2ca972baffe86c2581cf3cR436-R441) immediately after we received the signal (and followed by the same activation on transaction ends). Receiving the signal indicates the function is already created in another frontend and ready to use.

>  but we'll need to deal with not having SQL available, I think?

SQL is always available in the database, so the hydration will load both SQL and the function call. We strip out the SQL whenever the function cache is activated.

msullivan (approved):

msullivan:
You're right, totally missed that being handled!

msullivan:
I'm dropping `to-backport-5.x` because there are follow-up bugs that need to be fixed before we can backport.
(Maybe we should change the default back on master to avoid breaking nightlies?)

fantix:
The bug is fixed in #7564 which is marked with `to-backport-5.x`, so adding the same tag on its dependency too.

PR #7427, opened 2024-06-04
Add array modifying functions

### LINKED ISSUE #5737: Add "array_set" function to mutate array elements by index
Currently the way to mutate an array element by index would be something like this

`
with id = <uuid>$id
update Object filter .id = id 
set {
  some_array := some_array[:5] ++ [3] ++ some_array[6:]
}
`
it would be nice to have a "array_set" function to mutate array elements by index.

dnwpark:
Added `array_set`, `array_insert`, and `array_remove`.

The query `SELECT array_set([1, 2, 3, 4], 2, 9);` returns `[1, 2, 9, 4]`.

The query `SELECT array_insert([1, 2, 3, 4], 2, 9);` returns `[1, 2, 9, 3, 4]`.

The query `SELECT array_remove([1, 2, 3, 4], 2);` returns `[1, 2, 4]`.

close #5737


dnwpark:
There are some subtleties, such as `array_set` not throwing out of bounds errors, to mirror `array_get`. Also, negative indexes in general.

tests/test_edgeql_functions.py
```diff
@@ -996,6 +996,309 @@ async def test_edgeql_functions_array_get_08(self):
...
+        # Size 0 array
+        await self.assert_query_result(
+            r'''SELECT array_set(<array<int64>>[], 0, 9);''',
+            [[]],
+        )
+
+        await self.assert_query_result(
+            r'''SELECT array_set(<array<int64>>[], 1, 9);''',
+            [[]],
+        )
```
vpetrovykh, line 1082: _Assigning_ a value outside of bounds of an array does not reasonably produce an unchanged array as we don't support gaps in indexes. So getting an error is reasonable here (instead of silently ignoring an undefined operation).
I also assume that for `array_set` the _length_ of the array is invariant, so we cannot just add an element at index 0 of an `[]`.

Whatever bounds-checking policy we have for `array_set` should apply to `array_insert`. In fact, the only difference is that `array_set([0, 1, 2], 3, 9)` is illegal (as there is no element there to set), while `array_insert([0, 1, 2], 3, 9) == [0, 1, 2, 9]` since it is possible to insert an element after the last array element and still end up with a contiguous array. Other than that corner case, both `array_set` and `array_insert` are logically trying to create a non-contiguous array when they are putting elements out of bounds. As far as I can see, `array_insert` behaves reasonably w.r.t. acceptable and unacceptable index values/bound checks.

tests/test_edgeql_functions.py
```diff
@@ -996,6 +996,309 @@ async def test_edgeql_functions_array_get_08(self):
...
+
+    async def test_edgeql_functions_array_remove_02(self):
+        with self.assertRaisesRegex(
+            edgedb.InvalidValueError,
+            'array index 4 is out of bounds'
+        ):
+            await self.con.query(
+                r'''SELECT array_remove([1, 2, 3, 4], 4);''',
+            )
```
vpetrovykh, line 1264: The precedent of `array_get` is that you never get an out-of-bounds error, but instead you get `{}` (which is otherwise impossible) or whatever the default value that you specified.

This logic _should_ work for `array_remove` as well. Removing a non-existing element from an array is a NOP and the original array is unchanged. Instead I see "array index <blah> is out of bounds".

vpetrovykh (commented):
Overall comment from me is that I think the out-of-bounds checks need to be different.

This is trying to balance usability with correctness and error handling. If you wish to out-of-bounds errors, you can use `[]` instead of `array_get`. On the other hand an element outside of the array doesn't exist, so "getting" or "removing" them doesn't do anything, while trying to "set" them is potentially incorrect and makes no sense in treating as "nothing changed".

So `array_remove` should never produce an "index out of bounds" error, while both `array_set` and `array_insert` should check the index bounds and produce an error when the resulting array would have to have a "gap" in elements.

dnwpark:
@vpetrovykh  

I think you are correct that `array_set` should throw on out of bounds. I was thinking it should imitate `array_get` as their names are similar, but as a modifying function, it should be treated differently.

Your point about the length of the arrays being an invariant is an interesting way of looking at it.

I don't think it follows then that `array_remove` should not throw though. If `array_set` always keeps the length of the array fixed, and `array_insert` always increases the length by 1, then I think `array_remove` should always decrease it by 1.

I'm not sure why `array_get` forms a precedent here.

vpetrovykh:
`array_remove` is an interesting case.
First of all there might be two different variants:
1) remove an element at the specified _index_
2) remove all elements matching the specified _value_ (this is what `array_remove` does in Postgres)

Perhaps we want _both_ of the variants? `array_remove_index` and `array_remove_value`?

So let's consider a few things about arrays and removing stuff from them. We'll start with option 2), actually, as I think it's easier to conceptualize. I have an array of integers. Should it always be legal to remove the value `3` from it? Sure, why not? If the array has some `3`s in it, remove them all, if not, leave the array as is.

For the case 1) it's a little weirder, but the "algorithm" is that when I tell you to remove an element at a certain index `n` what I mean is kinda "omit this one spot in the array and splice the rest together". In other words, this: `val[:n] ++ val[n+1:]`. If you try this in EdgeDB right now, you can set `n` to be any index you like without a problem.

vpetrovykh:
When we try to generalize this logic to _adding_ or _setting_ values at a certain index, we run into trouble because the "reasonable" interpretation of "insert value 99 at index 9 of [1, 2]" is something like `[1, 2,,,,,,,,99]` which is not a type of array we support. It's that latter fact that makes it illegal.

We can decide that `array_set` does not actually care about the length of the array being invariant. Then the different case from `array_insert` will not be how the element right _past_ the end of the array is handled, they both will happily append values at index `len(arr)`. The difference will be how the element _before_ index 0 is handled.

`array_insert` preserves all existing elements and shifts them over to the right (and then updates the indexes). So `array_insert([1, 2], -3, 99)` should result in `[99, ,1, 2]` which is illegal.
`array_set` however _replaces_ the value at the specified index (and then updates the indexes). So `array_set([1, 2], -3, 99)` should result in `[99,1, 2]` which is fine.
Personally, I think it's more intuitive if `array_set` implies that the length of the array is not affected, so I propose that as the invariant.

dnwpark:
@vpetrovykh I've updated `array_set` to raise an out of bounds exception.

Given that the slicing works well to remove values, perhaps we should hold off on adding a remove function.

In an earlier chat I did briefly mention how c# uses `Remove` and `RemoveAt` for your proposed `array_remove_index` and `array_remove_value`. There is also a postgres function `array_positions` which returns all indexes of a value and so we may want `find_indexes` and `array_remove_indexes` as well.

Since remove leads to a much bigger design space, maybe we just stick to `array_set` and `array_insert` for this first bit.

vpetrovykh:
Let's do `array_set` and `array_insert` for this first iteration and revisit "remove" and "find" functionality.

dnwpark:
@vpetrovykh Agreed. I've reduced this PR to just `array_set` and `array_insert`.

dnwpark:
@vpetrovykh Did some benchmarking on whether to "combine" branches in the implementation.

Running each 10k times:

- set_empty: `SELECT array_set(<array<int64>>[], 0, 1)`
- set_out_of_bounds: `SELECT array_set(<array<int64>>[1,2,3,4,5,6,7,8,9], 25, -1)`
- set_front: `SELECT array_set(<array<int64>>[1,2,3,4,5,6,7,8,9], 0, -1)`
- set_middle: `SELECT array_set(<array<int64>>[1,2,3,4,5,6,7,8,9], 1, -1)`
- set_back: `SELECT array_set(<array<int64>>[1,2,3,4,5,6,7,8,9], 8, -1)`
- insert_empty: `SELECT array_insert(<array<int64>>[], 0, 1)`
- insert_empty_out_of_bounds: `SELECT array_insert(<array<int64>>[], 1, 1)`
- insert_out_of_bounds: `SELECT array_insert(<array<int64>>[1,2,3,4,5,6,7,8,9], 25, -1)`
- insert_front: `SELECT array_insert(<array<int64>>[1,2,3,4,5,6,7,8,9], 0, -1)`
- insert_middle: `SELECT array_insert(<array<int64>>[1,2,3,4,5,6,7,8,9], 1, -1)`
- insert_back: `SELECT array_insert(<array<int64>>[1,2,3,4,5,6,7,8,9], 9, -1)`

Results:

|Operation|Raises|Existing|Refactored|
|-|-|-|-|
|`array_set` empty|Yes|`7.479`|`7.310`|
|`array_set` out of bounds|Yes|`9.838`|`7.121`|
|`array_set` front|No|`5.387`|`7.522`|
|`array_set` middle|No|`5.713`|`7.608`|
|`array_set` back|No|`5.686`|`6.061`|
|`array_insert` empty|No|`5.951`|`7.025`|
|`array_insert` empty out of bounds|Yes|`6.853`|`7.145`|
|`array_insert` out of bounds|Yes|`8.010`|`7.390`|
|`array_insert` front|No|`5.971`|`5.793`|
|`array_insert` middle|No|`6.015`|`5.675`|
|`array_insert` back|No|`5.881`|`5.433`|

I don't think there's a clear benefit.

vpetrovykh (approved):
Looks good

PR #7435, opened 2024-06-06
Interpreter: Add indexes to links

zchenb:


msullivan (approved):

msullivan:
We also want them on multi links

Issue #6697, opened 2024-01-12
In-place major version upgrades


msullivan:
Currently, while we allow in-place upgrades between minor versions, upgrading between major versions requires a dump and restore. This is slow, and requires downtime. 

Currently we use the patch system to make catalog changes during minor version upgrades, but using the patch system for major versions is probably unworkable:
1. The number of patches needed would be much higher, which is problematic because they are a pain to develop and maintain. Currently we only need patches when a backported bugfix requires a catalog change. Now it would be needed for all *features* on the main branch.
2. Worse, there is no good way with that approach to populate new fields in the reflection schema.

The plan then, is this:
 * `edgedbstd` gets a version suffix, so `edgedbstd_vX`
 * `edgedbstd` still exists, and will contain trampoline functions and views that point to the equivalent functions/tables in the namespaced schema
 * When compiling normal queries, the compiler can use the versioned std namespace directly, but when compiling user-defined functions, constraints, etc, it will use trampoline schema `edgedbstd`.
 * When we do an upgrade, we will populate the new `edgedbstd_vY` schema, and then update everything in `edgedbstd` to point to it. (This probably means we can't ever remove fields from user visible tables, but that's probably OK.)
 * The reflection schema also lives in the new `edgedbstd_vX` database. Before the upgrade, we will `DESCRIBE SCHEMA` the old database, along with all the schema object ids, (like dump does), and during the upgrade we will run that DDL script in order to populate the new reflection schema. (But we will *not* compile it to postgres, since the postgres tables already exist and are supposed to be compatible)

Eventually we might need to *actually* change layout, in which case some major versions might not be upgradable. But also there's a good chance we could do it gradually, with some tables or functions or whatnot using the new layout and allowing gradual backfill migration.

Open questions (partial list):
 * This may require us to change our policy on postgres version support and new features. For example, we dropped support for postgres 13 in edgedb 4.0 because it didn't support multiranges. In-place upgrades increase the benefit of maintaining support for old postgres versions for longer. Maybe this would have made it worthwhile to disable multirange support on postgres 13, or to require it live in an extension.
 * Maybe we could query pg_catalog to make sure that the database looks like we expect it to still?
 * How do we handle extensions?

msullivan:
I've focused mostly on the compiler component, but there is also a lot of room for cleverness to make this truly in place. With appropriate setup, almost all of this can happen while the previous version is still running, and even serving write queries. (As long as no migrations are being done)

[...]

PR #7439, opened 2024-06-07
Start namespacing and trampolining the standard library

msullivan:
The idea here is that all of the functions, tables, and views in
edgedb/edgedbstd/edgedbsql should be moved into namespaced libraries
of the form `edgedb_VER`, where VER will be substituted with some
version identifier.

Then, for anything that might be referenced by a function, constraint,
etc, we will create a *trampoline* in the un-suffixed namespace.
When doing (eventually) in-place version upgrades, we will create the
new namespace and then update the trampolines to point to it.

So far, functions and views are mostly namespaced. Standard library
schema object tables aren't yet.

### DIFF
--- a/edb/pgsql/trampoline.py
+++ b/edb/pgsql/trampoline.py
@@ -17,7 +17,21 @@
 #
 
 
-"""Support for namespacing and trampolining the standard library."""
+"""Support for namespacing and trampolining the standard library.
+
+The idea here is that all of the functions, tables, and views in
+edgedb/edgedbstd/edgedbsql should be moved into namespaced libraries
+of the form `edgedb_VER`, where VER will be substituted with some
+version identifier.
+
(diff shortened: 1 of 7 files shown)


edb/pgsql/trampoline.py
```diff
@@ -0,0 +1,134 @@
...
+
+q = common.qname
+qi = common.quote_ident
+
+
+def versioned_schema(s: str, version: Optional[int]=None) -> str:
+    if version is None:
+        # ... get_version_dict() is cached
+        version = buildmeta.get_version_dict()['major']
+    return f'{s}_v{version}'
```
elprans, line 59: Since we're not consistently quoting schema names in SQL, I'd add a comment here that the return value MUST be lowercase.

elprans (approved):

msullivan:
Progress towards #6697.

PR #7440, opened 2024-06-07
ai: Properly forward non-successful responses from non-streaming chat

elprans:


msullivan (approved):

msullivan:
Should we merge this?

PR #7443, opened 2024-06-10
Expose EdgeQL computed pointers over SQL adapter

### LINKED ISSUE #6692: Computeds in SQL adapter
Continuing discussion from #6017

aljazerzen:
Closes #6692

The funny thing is that this was mostly easy, the hard part was introspection. I had to:
- pay off a bit of tech dept (we weren't matching all rows for `id` in `information_schema.columns`/`pg_attribute` with correct row for the pointer in `_SchemaPointer`).
- change the direction of joining (before we started with all rows in the `information_schema.columns`/`pg_attributes` and tried to join in schema pointers. Now we do it the other way around. This means that "hidden" columns (`__fts_document__`) don't need to be filtered out, which is nice.

### DIFF
--- a/edb/pgsql/resolver/expr.py
+++ b/edb/pgsql/resolver/expr.py
@@ -1,4 +1,5 @@
 #
+#
 # This source file is part of the EdgeDB open source project.
 #
 # Copyright 2008-present MagicStack Inc. and the EdgeDB authors.
@@ -26,6 +27,14 @@ from edb import errors
 
 from edb.pgsql import ast as pgast
 from edb.pgsql import trampoline
+from edb.pgsql import compiler as pgcompiler
+
+from edb.schema import types as s_types
+
+from edb.ir import ast as irast
+
+
+from edb.edgeql import compiler as qlcompiler
 
 from . import dispatch
 from . import context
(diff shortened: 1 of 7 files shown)


aljazerzen:
What is left now, it to figure out how to invent values for `pg_attribute` for computed columns. Right now, I've added a bunch of default values, which seems to cause trouble.

aljazerzen:
This PR contains a few commits that I've separated into different PRs, so I'll rebase it when they are merged.

edb/pgsql/resolver/expr.py
```diff
@@ -117,6 +126,40 @@ def resolve_column_kind(
...
             # special case: __type__ static value
             return _uuid_const(val)
+        case context.ColumnComputable(pointer=pointer):
+
+            expr = pointer.get_expr(ctx.schema)
+            assert expr
+
+            source = pointer.get_source(ctx.schema)
+            assert isinstance(source, s_types.Type)
+            # assert isinstance(source, (s_types.Type, s_pointers.Pointer))
```
msullivan, line 136: Stray

msullivan (approved):

edb/pgsql/resolver/expr.py
aljazerzen: This was left here because I was preparing to support link properties on computeds.

Can you confirm that we don't support them in EdgeQL?

PR #7445, opened 2024-06-10
Fix pg_dump of an empty schema

### LINKED ISSUE #7444: Bad SQL adapter introspection of an empty database
- EdgeDB Version: 6.0-dev

1. Init a new, empty database.
2. `pg_dump "sslmode=require host=localhost port=5656 user=edgedb" > dump.sql`

```
pg_dump: error: schema with OID 163001966 does not exist
```

This happens because pg_class contains the made-up pg_namespace OIDs, which are needed for pg_class row for "global types", such as `tuple<std::int64, std::int64>`.

SQL adapter exposes tables for object types (and links) in a separate schemas, but those schemas don't exist in the backing PostgreSQL. So we must generate OIDs for such namespaces.

Now, when database is empty, virtual_tables is empty (as there is no virtual tables), but we still expose all tuple types in pg_class. These rows are the ones that contain the invalid oid-s.

aljazerzen:
Closes #7444

@msullivan should i add a test for pg_dump? I'm not sure how expensive this is.

### DIFF
--- a/edb/pgsql/metaschema.py
+++ b/edb/pgsql/metaschema.py
@@ -5914,6 +5914,7 @@ def _generate_sql_information_schema() -> List[dbops.Command]:
         trampoline.VersionedView(
             name=("edgedbsql", "pg_namespace"),
             query="""
+        -- system schemas
         SELECT
             oid,
             nspname,
@@ -5929,6 +5930,8 @@ def _generate_sql_information_schema() -> List[dbops.Command]:
         WHERE nspname IN ('pg_catalog', 'pg_toast', 'information_schema',
                           'edgedb', 'edgedbstd')
         UNION ALL
+
+        -- virtual schemas
         SELECT
             edgedbsql_VER.uuid_to_oid(t.module_id)  AS oid,
             t.schema_name                       AS nspname,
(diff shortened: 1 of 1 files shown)


edb/pgsql/metaschema.py
```diff
@@ -5949,8 +5952,16 @@ def _generate_sql_information_schema() -> List[dbops.Command]:
             '0'::cid                            AS cmax,
             NULL                                AS ctid
         FROM (
-            SELECT DISTINCT schema_name, module_id
+            SELECT schema_name, module_id
```
msullivan, line 5955: Can this wind up with something duplicated?

aljazerzen: No, I use `UNION` just below (and UNION removed all duplicates, I've tested that)

aljazerzen:
> @msullivan should i add a test for pg_dump? I'm not sure how expensive this is.

edb/pgsql/metaschema.py
msullivan: :+1:

msullivan (approved):
I think our pg_dump tests *are* pretty slow. Make a judgement call based on how likely you think it is that somebody might break this in the future

PR #7446, opened 2024-06-10
Fix SQL adapter COPY command

aljazerzen:
Fixes a few bugs in COPY:

- copying link tables would fail because of missing id column,
- colnames were not handled correctly,

### DIFF
--- a/edb/pgsql/resolver/command.py
+++ b/edb/pgsql/resolver/command.py
@@ -19,12 +19,13 @@
 """SQL resolver that compiles public SQL to internal SQL which is executable
 in our internal Postgres instance."""
 
-from typing import Optional, Dict, List
+from typing import Optional, Dict
 
 from edb.pgsql import ast as pgast
 
 from . import dispatch
 from . import context
+from . import expr as pg_res_expr
 
 Context = context.ResolverContextLevel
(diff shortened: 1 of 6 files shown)


edb/pgsql/resolver/command.py
```diff
@@ -19,75 +19,58 @@
...
+                col.name: col for col in table.columns
+            }
+            selected_columns = (col_map[name] for name in stmt.colnames)
+        else:
+            selected_columns = (c for c in table.columns if not c.hidden)
+
+        # The relation being copied is potentially a view and views cannot be
+        # copied if referenced by name, so we just always wrap it into a SELECT.
 
-        if relation.schemaname == 'edgedbpub':
```
msullivan, line 53: Special case wasn't needed?

msullivan (approved):

edb/pgsql/resolver/command.py
aljazerzen: It doesn't hurt to always wrap the relation into a SELECT. This if assumes that we only expose VIEWs as TABLEs in `edgedbpub` schema, which is probably true, but also something can and will change with the trampolined schemas.

PR #7459, opened 2024-06-13
Create a sub rvar for inherited ptrref rvars.

dnwpark:
Following on #7454, add a sub rvar for the group of queries generated when expanding a pointer reference's inheritance.

In order to ensure correct behaviour when narrowing the type of a link union, the `schema_obj_id` is persisted when creating a range from a single query.

### DIFF
--- a/edb/pgsql/compiler/relctx.py
+++ b/edb/pgsql/compiler/relctx.py
@@ -1838,15 +1838,7 @@ def range_from_queryset(
         # Just one class table, so return it directly
         from_rvar = set_ops[0][1].from_clause[0]
         assert isinstance(from_rvar, pgast.PathRangeVar)
-        from_rvar = from_rvar.replace(
-            typeref=typeref,
-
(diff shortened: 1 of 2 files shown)


edb/pgsql/compiler/relctx.py
```diff
@@ -1838,7 +1838,15 @@ def range_from_queryset(
         # Just one class table, so return it directly
         from_rvar = set_ops[0][1].from_clause[0]
         assert isinstance(from_rvar, pgast.PathRangeVar)
-        from_rvar = from_rvar.replace(typeref=typeref)
+        from_rvar = from_rvar.replace(
```
msullivan, line 1841: Let's see if we can fix this in `ast.replace` instead

msullivan: Good job tracking this down though! It is weird and subtle

msullivan (approved):

PR #7460, opened 2024-06-13
Version a lot more, trampoline a bit less

msullivan:
Don't trampoline schema alias views, but do still version them.

Version schema stdlib types.

Teach common.get_backend_name and friends about versioning.
Importantly, and unlike previously, the *default* is now to version
the names.

Currently this results in us basically *always* using the
versioned names and almost never using the trampolines.

Next up is making whether to use versioned names fully configurable
throughout the compiler and configuring it properly when calling into
it.

Progress towards #6697.


elprans (approved):

msullivan:
Oh, notably, we *don't* trampoline the tables yet (with views), which we will need to

aljazerzen (approved):

PR #7461, opened 2024-06-14
Copy meta fields when copying AST nodes.

dnwpark:
Meta fields ought to be copied when copying ast nodes as well.

Additionally, re-add `schema_object_id` to the meta fields for pg ast.


msullivan (approved):

msullivan:
Oh good, I was a little afraid something would break

PR #7466, opened 2024-06-17
Propagate spans in parsed SQL

aljazerzen:
libpg_query emits a `"location"` fields that we already use and convert to our spans, but not on all nodes.

This PR make it so we now propagates spans from children into parents. It uses and improves existing `span.SpanPropagator`.


aljazerzen:
Oh, to be able to update spans, I've removed the `pgast.ImmutableBase`, so nodes can be updated. I can revert that if there is a principled reason for it.

PR #5364, opened 2023-04-16
fix edgedb-docker:#72 tenant-id isn't supported by edgedb-docker

chickeaterbanana:
for better configurability and support in docker, it's needed to have a environment variable for the tenant id. This helps to implement the tenant id support in docker.

see also: https://github.com/edgedb/edgedb-docker/issues/72

PR #7470, opened 2024-06-19
server: add missing args env vars

fantix:
This is somehow missing and mismatching the documentation.

Closes #5364

### DIFF
--- a/docs/reference/environment.rst
+++ b/docs/reference/environment.rst
@@ -274,6 +274,13 @@ and ``*_ENV`` variants are also supported.
 .. _URI format:
    https://www.postgresql.org/docs/13/libpq-connect.html#id-1.7.3.8.3.6
 
+EDGEDB_SERVER_MAX_BACKEND_CONNECTIONS
+.....................................
+
+The maximum NUM of connections this EdgeDB instance could make to the backend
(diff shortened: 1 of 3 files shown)


jackfischer:
Possible to add `EDGEDB_DEBUG_EDGEQL_TEXT_IN_SQL`?

fantix:
> Possible to add `EDGEDB_DEBUG_EDGEQL_TEXT_IN_SQL`?

Oh all the `EDGEDB_DEBUG_*` [debug switches](https://github.com/edgedb/edgedb/blob/e1f5c1a6bc94677fad43fa3a8657b7b9fa86c57d/edb/common/debug.py#L264) are always supported; they don't need an extra option in the server command parameter list.

jackfischer:
I'm actually getting `/usr/lib/x86_64-linux-gnu/edgedb-server-4/lib/python3.11/site-packages/edb/common/debug.py:259: UserWarning: Unknown debug flag: 'EDGEDB_DEBUG_EDGEQL_TEXT_IN_SQL'` but this may be because we are on 4.7

fantix:
> this may be because we are on 4.7

Right, this option is only added since [5.0](https://github.com/edgedb/edgedb/blob/v5.0/edb/common/debug.py#L146)

msullivan (approved):

PR #7473, opened 2024-06-19
Add sql_setup_script and sql_teardown_script fields to extension packages

msullivan:
This lets us create any pure SQL things we need. Teardown obviously
needs to clean up after setup.
Specified as string literals, so using $$ strings is recommended.


vpetrovykh (approved):
LGTM

PR #7475, opened 2024-06-20
Postgis

vpetrovykh:
This is a WIP PR.

Mostly at this point I expect to go over the exposed functions, test that they _run_ (don't care for functionality tests mostly, as that is presumably covered by postgis itself). There are probably still a few internal functions that we don't care to expose left in there.

I haven't written schema and index tests yet, but I have created properties and used them in `gist`, `spgist`, and `brin` indexes to see that it can be created.

The extension code has hand-written parts in the beginning, type definitions and casts. After that the content is generated by a script. The script is in the root directory `get_postgis_func.py`. It prints out the extracted operators and functions. I tend to redirect its output into a file (`python get_postgis_func.py > gen.edgeql`) where I either examine the output for specific issues I'm fixing or copy-paste it into the extension `edgeql` file.

I'll be updating this branch as I write more tests and refine the automatic extractor script.


vpetrovykh:
In the latest iteration I cleaned up some more functions (fixing signatures mostly). This uncovered an issue with `array<geometry>>` (at least it seems to affect functions that have `array<geometry>>` in the signature and goes away when I commented then out). Those functions cause an error when attempting to initialize the extension: `edb.errors.InvalidValueError: cannot determine OID of EdgeDB type 'b855d9a4-eb15-5850-9a93-2fd6e35197ae'`. I haven't looked into the details of it, yet.

There are more tests now, they cover about 1/4 or so of functions directly and probably validate more since the basic principle of how function signatures were handled is largely the same for most functions. As the issue with arrays showed it's still important to run every function to make sure that no error creeped in when converting signatures.

vpetrovykh:
This branch has the extension working well enough to run the test suite, so it should be OK to use this in order to polish how we add/package the extension.

I was referring to https://postgis.net/docs/manual-3.4/postgis_installation.html when I was making sure that my Ubuntu had all the libraries necessary to run the setup. I ended up installing GEOS, proj, protobuf, docbook-xls, xsltproc, but obviously this will vary from system to system. My takeaway is that the postgis build guide helped in this process.

get_postgis_func.py
```diff
@@ -0,0 +1,548 @@
+import json
```
1st1, line 1: This should be an `edb` tool, so move it into `edb/tools` and make it a proper command

edb/pgsql/delta.py
```diff
@@ -7557,28 +7557,34 @@ def _create_extension(self, ext_spec: str) -> None:
             lclauses.append(f'v.split {op} {pver}')
         cond = ' and '.join(lclauses) if lclauses else 'true'
 
+        ver_regexp = r'^\d+(\.\d+)+$'
```
msullivan, line 7560: Why do we need this version regexp?

vpetrovykh: Because of this:
```
V2f206a4ba0_main=# select name, version from pg_available_extension_versions where name = 'postgis';
  name   |  version   
---------+------------
 postgis | 3.4.2
 postgis | ANY
 postgis | unpackaged
(3 rows)
```
The version is not necessarily numeric and that causes issues when trying to match the version number by splitting it and converting to ints.

edb/ir/typeutils.py
```diff
@@ -196,6 +196,28 @@ def is_persistent_tuple(typeref: irast.TypeRef) -> bool:
...
+) -> Optional[str]:
+    # FIXME: instead of hardcode we need to extract this from the
+    # schema/extension
+    if typeref.real_base_type.id in {
+        s_obj.get_known_type_id('ext::postgis::box2d'),
+        s_obj.get_known_type_id('ext::postgis::box3d'),
+    }:
+        return 'geometry'
+
+    return None
```
msullivan, line 210: We should add a `custom_serialization` field to `ScalarType`, I guess.

(We should skip using known ids if we are temporarily hacking around it. Can we use `name_hint`?)

edb/schema/indexes.py
```diff
@@ -91,14 +91,39 @@ def is_index_valid_for_type(
...
+                or
+                expr_type.issubclass(
+                    schema,
+                    (
+                        schema.get('ext::postgis::geometry',
+                                   type=s_scalars.ScalarType,
+                                   default=None),
+                        schema.get('ext::postgis::geography',
+                                   type=s_scalars.ScalarType,
+                                   default=None),
```
msullivan, line 107: Maybe schema types should be able to declare a list of indexes that are known to support them?

edb/schema/objects.py
```diff
@@ -3032,11 +3032,14 @@ def _issubclass(
     def issubclass(
         self,
         schema: s_schema.Schema,
-        parent: Union[SubclassableObject, Tuple[SubclassableObject, ...]],
+        parent: Union[Optional[SubclassableObject],
+                      Tuple[Optional[SubclassableObject], ...]],
```
msullivan, line 3036: How necessary is this? It seems hacky.

msullivan:
I think there are three PRs to extract from this and land separately:
* Custom serialization (with the hardcoding of names, I guess)
* Index support
* The various bugfixes to extensions

We can land those with some hardcoded bits to get everything working, but then should follow up with
 * Configure the custom serialization with a field on ScalarType
 * Configure the indexes somehow; probably with a field on ScalarType that lists valid index types

PR #7478, opened 2024-06-20
Rust connection pool implementation

mmastrac:
Note: this is not a Python integration -- that part of the code was pulled out of this PR and will happen in https://github.com/edgedb/edgedb/pull/7544

This PR migrates the connection pooling algorithm from Python to Rust. It attempts to maintain as much of the algorithm as possible, though there will be some differences as a result of how the code is structured.

We maintain the same terminology of "blocks" from the existing implementation and the spirit of the overall algorithm is the same as before. There are some unavoidable differences due to code structure and language changes, however the QoS results show the similar behaviour in all cases.

The code tries to maintain separation between the following components to make it easier to reason about:

 1. The low-level blocks/block, connections, and metrics code. This code
    creates, destroys and transfers connections without understanding of
    policies, quotas or any sort of algorithm. We ensure that the blocks and
    metrics are reliable, and use this as a building block for our pool.
 2. The algorithm. This performs planning operations for acquisition, release
    and rebalancing of the pool. The algorithm does not perform operations, but
    rather informs that caller what it should do.
 3. The pool itself. This drives the blocks and the connector interface, and
    polls the algorithm to plan next steps during acquisition, release and
    during the timer-based planning callback.
 4. The Python integration code. This is behind an optional feature, and exposes
    PyO3-based interface that allows a connection factory to be implemented in
    Python.

QoS:

```
2024-07-15T19:05:01.550982Z  INFO conn_pool::pool::tests: QoS[test_connpool_1] = [94.19]
2024-07-15T19:05:01.550987Z  INFO conn_pool::pool::tests: QoS[test_connpool_10] = [100.00]
2024-07-15T19:05:01.550992Z  INFO conn_pool::pool::tests: QoS[test_connpool_2] = [78.73]
2024-07-15T19:05:01.550996Z  INFO conn_pool::pool::tests: QoS[test_connpool_3] = [99.67]
2024-07-15T19:05:01.551000Z  INFO conn_pool::pool::tests: QoS[test_connpool_4] = [91.84]
2024-07-15T19:05:01.551004Z  INFO conn_pool::pool::tests: QoS[test_connpool_5] = [96.70]
2024-07-15T19:05:01.551009Z  INFO conn_pool::pool::tests: QoS[test_connpool_6] = [100.00]
2024-07-15T19:05:01.551013Z  INFO conn_pool::pool::tests: QoS[test_connpool_7] = [99.53]
2024-07-15T19:05:01.551017Z  INFO conn_pool::pool::tests: QoS[test_connpool_8] = [100.00]
2024-07-15T19:05:01.551022Z  INFO conn_pool::pool::tests: QoS[test_connpool_9] = [84.71]
2024-07-15T19:05:01.551029Z  INFO conn_pool::pool::tests: QoS = [94.54] (rms=91.07)
```

The following items are left for a future, integration TODO:

  - [ ] Error states for connections
  - [ ] Python/Rust integration
  - [ ] Generate compatible JSON for previous reports
  - [ ] Hook up old connection tests
  - [ ] iterate_connections API
  - [ ] _build_snapshot (metrics)
  - [ ] Garbage collection may be too aggressive
  - [ ] Try LRU/MRU for connection acquisition in blocks

### DIFF
--- a/edb/server/conn_pool/src/algo.rs
+++ b/edb/server/conn_pool/src/algo.rs
@@ -1,8 +1,5 @@
 use scopeguard::defer;
-use std::{
-    cell::{Cell, RefCell},
-    num::NonZeroUsize,
-};
+use std::cell::{Cell, RefCell};
 use tracing::trace;
 
 use crate::{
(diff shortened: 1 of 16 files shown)


edb/server/conn_pool/src/algo.rs
```diff
@@ -0,0 +1,121 @@
...
+
+    }
+}
+
+#[cfg(test)]
+mod tests {
+
+    #[test]
+    fn test_one_block() {}
+}
```
1st1, line 137: I'm not opposed to having unit tests right in the source files (it's idiomatic for Rust, right?) but let's also have big functional tests as they were, written in Python (more declarative and accessible for the rest of the team)

edb/server/conn_pool/src/python.rs
```diff
@@ -0,0 +1,371 @@
...
+        &mut self,
+        py: Python,
+        conn: PyObject,
+    ) -> Option<PoolHandle<PythonConnectionFactory>> {
+        let Some(py_dict) = &mut self.py_dict else {
+            return None;
+        };
+        let py_dict = py_dict.as_ref(py);
+        let item = py_dict.get_item(conn.clone_ref(py)).ok()??;
+        _ = py_dict.del_item(conn);
```
1st1, line 65: Python dicts have also pop_item() / pop() methods, which could be useful here

mmastrac: Ah, yeah that could be useful. I may need to PR some bindings for that in PyO3, or extract and save the dict.pop method. I'll check the CPython binding API to see if it is possible to call.

1st1: It's a minor nit comment though, ultimately it doesn't matter.

mmastrac:
Review TODO:

 - [ ] Confirm that all-time Closed connections are correct
 - [ ] Make sure that queuing is fair (don't skip the line)

edb/server/conn_pool/src/metrics.rs
```diff
@@ -0,0 +1,373 @@
...
+    cumulative: u64,
+    ptr: u8,
+    /// If we've never rolled over, we cannot divide the entire array by `SIZE`
+    /// and have to use `ptr` instead.
+    rollover: bool,
+}
+
+impl<const SIZE: usize> Default for RollingAverageU32<SIZE> {
+    fn default() -> Self {
+        assert!(SIZE <= u8::MAX as _);
```
aljazerzen, line 36: Why not use  `<const SIZE: u8>`?

aljazerzen (approved):
I've skimmed trough this, it looks good at the glance, but I didn't invest the time to understand the algorithm itself.

edb/server/conn_pool/src/metrics.rs
mmastrac: Good question. This doesn't compile when using a `u8` [1] but there might be a workaround [2] I can trying today:

[1] https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=5cac05a391babb48572e722be85588f2

[2] https://github.com/rust-lang/rust/issues/104949

aljazerzen: Ah ok, doesn't matter then.

fantix (approved):
This is a great improvement in general!

It's worth noting that the Rust implementation differs from the Python pool in a way that the Rust pool tries to use a consistent overfullness-based (pluggable!) algorithm to rebalance/redistribute connections among blocks that are overwhelming the pool, instead of the per-case manually optimized algorithm in the Python pool (rebalance for mode C and round-robin for mode D). This is working great with config knobs tweakable towards QoS results directly, making the code more straightforward too.

edb/server/conn_pool/src/conn.rs
```diff
@@ -0,0 +1,348 @@
...
+                ConnInner::Disconnecting(Instant::now(), f)
+            }
+            _ => unreachable!(),
+        });
+    }
+
+    pub fn transfer(&self, connector: &C, from: &MetricsAccum, to: &MetricsAccum, db: &str) {
+        self.untrack(from);
+        self.transition(|inner| match inner {
+            ConnInner::Idle(_t, conn, ..) | ConnInner::Active(_t, conn, ..) => {
```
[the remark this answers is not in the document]
mmastrac: `transfer` is used for two different cases -- basically this handles the case of "stealing" a connection on release (when it would still be marked as active) vs "stealing" an idle connection.

edb/server/conn_pool/src/block.rs
```diff
@@ -0,0 +1,921 @@
...
+        }
+    }
+
+    #[inline]
+    pub fn metrics(&self) -> Rc<MetricsAccum> {
+        self.state.metrics.clone()
+    }
+
+    fn try_acquire_used(&self) -> Option<Conn<C>> {
+        for conn in &*self.conns.borrow() {
```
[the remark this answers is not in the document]
mmastrac: Good point. I think I'll experiment with MRU in a future patch as it's probably not a terribly difficult thing to do. Adding to the TODO list.

mmastrac:
This PR migrates the connection pooling algorithm from Python to Rust. It attempts to maintain as much of the algorithm as possible, though there will be some differences as a result of how the code is structured.

We maintain the same terminology of "blocks" from the existing implementation and the spirit of the overall algorithm is the same as before. There are some unavoidable differences due to code structure and language changes, however the QoS results show the similar behaviour in all cases.

PR #7480, opened 2024-06-21
Workaround for developing with GCC 14

fantix:
Refs: https://github.com/cython/cython/issues/5240

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,14 +12,15 @@ build-reqs:
 
 cython: build-reqs
 	find edb -name '*.pyx' | xargs touch
-	BUILD_EXT_MODE=py-only python setup.py build_ext --inplace
+	# See also: https://github.com/cython/cython/issues/5240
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -12,14 +12,15 @@ build-reqs:
 
 cython: build-reqs
 	find edb -name '*.pyx' | xargs touch
-	BUILD_EXT_MODE=py-only python setup.py build_ext --inplace
+	# See also: https://github.com/cython/cython/issues/5240
+	CFLAGS="${CFLAGS} -Wno-error=incompatible-pointer-types" BUILD_EXT_MODE=py-only python setup.py build_ext --inplace
```
elprans, line 16: Add these to `setup.py` `CFLAGS` instead.

fantix: Ah makes sense!

elprans (approved):

PR #7488, opened 2024-06-26
Allow subdomains in redirects

scotttrinh:
This was the original design intention, but didn't actually test or implement allowing subdomains.

See https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/authorizing-oauth-apps#redirect-urls for an explanation of the algorithm used here, which is: "If provided, the redirect URL's host (excluding sub-domains) and port must exactly match the callback URL. The redirect URL's path must reference a subdirectory of the callback URL."


fantix (approved):
LGTM!

PR #7489, opened 2024-06-26
Enable type checking in edb.server.protocol.

dnwpark:
Most changes are adding type annotations or casting return values.

However, I did make a few changes to make some things clearer:
- Moved/renamed `auth_ext.data.Config` to `auth_ext.config.OAuthProviderConfig`
    - Changed to inherit `auth_ext.config.ProviderConfig`
    - Changed to `dataclass`
- Added `ai_ext.ProviderConfig` and `ai_ext.ApiStyle`
- Added `auth.scram.Session`

### DIFF
--- a/edb/server/protocol/auth_ext/http.py
+++ b/edb/server/protocol/auth_ext/http.py
@@ -17,6 +17,8 @@
 #
 
 
+from __future__ import annotations
+
 import datetime
 import http
 import http.cookies
@@ -29,7 +31,7 @@ import os
 import mimetypes
 import uuid
 
-from typing import Any, Optional, Tuple, FrozenSet, cast
+from typing import Any, Optional, Tuple, FrozenSet, cast, TYPE_CHECKING
 
 import aiosmtplib
 from jwcrypto import jwk, jwt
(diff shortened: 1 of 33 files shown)


msullivan (approved):

edb/server/protocol/auth_ext/http.py
```diff
@@ -83,72 +97,67 @@ async def handle_request(
...
-                    return await self.handle_ui_verify(*handler_args)
+                    await self.handle_ui_verify(request, response)
                 case ("ui", "resend-verification"):
-                    return await self.handle_ui_resend_verification(
-                        *handler_args
-                    )
+                    await self.handle_ui_resend_verification(request, response)
                 case ("ui", "magic-link-sent"):
-                    return await self.handle_ui_magic_link_sent(*handler_args)
+                    return await self.handle_ui_magic_link_sent(
```
scotttrinh, line 158: I'm not sure why we dropped the `return` in the other cases, but should we do it here too?

scotttrinh (approved):
Thanks so much for doing this! I tried to go down this path when building the auth extension, but ran into a bunch of hurdles, so I'm glad you got it working. 🙏 

No blockers, just one question about a `return await`

edb/server/protocol/auth_ext/http.py
dnwpark: Good catch!

PR #7455, opened 2024-06-12
Use CTE instead of inheritance views when generating ranges for types

dnwpark:
If a type or pointer has children and the compiler is not using EXPLAIN, use a CTE instead of the current inheritance views.

Also remove the `expand_inhview` debug flag.

Issue #7410, opened 2024-05-29
Remove inheritance views


msullivan:
Currently we generate an "inheritance view" for every object type and multi pointer in the schema, named the same as the object but suffixed with "_t" and selecting from the object and all of its descendants.

There are bugs (https://github.com/edgedb/edgedb/issues/6304) and performance issues (https://github.com/edgedb/edgedb/issues/5127, #5844) that result from this, in addition to making DDL basically unreadable and requiring a lot of logic in `pgsql/delta.py` to maintain.

We should remove them and instead generate non-materialized CTEs with the relevant inheritance queries when needed. (We can even omit those from the debug spew if we want.) We already have most of the required machinery in the pgsql compiler to help with ANALYZE (see `EDGEDB_DEBUG_EDGEQL_EXPAND_INHVIEWS`), but will need to update it to put it in a CTE typically.

Places that need to be updated:
 - [x] The edgeql compiler backend
 - [x] The SQL protocol resolver
 - [x] `_generate_schema_alias_view` - this can maybe use the same infrastructure as for SQL protocol
 - [x] Constraint trigger functions need to have their targets updated
 - [x] Endpoint actions

msullivan:
Oh, annoyingly, we'll also need to recompile functions when descendants from types they select are added or removed

PR #7494, opened 2024-06-26
Use inheritance CTEs instead of views in SQL resolver

aljazerzen:
- extract code for creating VIEWS from delta.py into inheritance.py and adapt it to construct pgast instead of strings,
- use that asts as inheritance CTEs in resolver,
- cleanup,

Prerequisite for #7455 

Ref #7410


msullivan (approved):
Fantastic, thanks!

PR #7495, opened 2024-06-26
Fix tenant shutdown in multi-tenant mode

fantix:
1. Request frontend connections to stop
2. Add a timeout to interrupt if they don't stop in time

### DIFF
--- a/edb/server/defines.py
+++ b/edb/server/defines.py
@@ -89,3 +89,7 @@ _TLS_CERT_RELOAD_EXP_INTERVAL = 0.1
 
 PGEXT_POSTGRES_VERSION = "13.9"
 PGEXT_POSTGRES_VERSION_NUM = 130009
+
+# The time in seconds the EdgeDB server will wait for a tenant to be gracefully
+# shutdown when removed from a multi-tenant host.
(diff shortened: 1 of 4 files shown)


nsidnev (approved):

elprans:
How does this interact with the `offline` readiness state?

fantix:
> How does this interact with the `offline` readiness state?

It's not currently reporting `offline` during the removal of a tenant, expecting the controlling plane to set `offline` in advance. However, I think it makes sense to report `offline` regardless of the value in the `readiness-state-file`, as soon as a tenant is removed from the `multitenant-config-file`.

elprans:
What I meant is that we set tenant state to `offline` already and that should effectively terminate all connections (possibly after a timeout).  This change is still valid, but toggling the readiness state should have a similar effect.

fantix:
> What I meant is that we set tenant state to `offline` already and that should effectively terminate all connections (possibly after a timeout). This change is still valid, but toggling the readiness state should have a similar effect.

Sounds reasonable! Will create a new PR for that - it'll involve some complexity of shutdown task management (which should be cancelled in case the state is switched back to online before timeout).

PR #7497, opened 2024-06-26
Add EDGEDB_TEST_REPEATS env var for testing function cache

msullivan:
When EDGEDB_TEST_REPEATS is set, the test suite duplicates
transactional query tests, and the re-run version will typically use
the function cache.

There are some actual failures when I ran this that need to be
investigated.


elprans (approved):

msullivan:
Changes look good. I can't mark "approved" because github thinks of me as the author, but let's merge

PR #7504, opened 2024-06-27
Use PathAspect enum wherever possible.

dnwpark:
Also moved the enum from `edb/pgsql/compiler/pathctx.py` to a new file `edb/pgsql/compiler/enums.py`.

### DIFF
--- a/edb/pgsql/ast.py
+++ b/edb/pgsql/ast.py
@@ -29,6 +29,9 @@ from edb.common import typeutils
 from edb.edgeql import ast as qlast
 from edb.ir import ast as irast
 
+if typing.TYPE_CHECKING:
+    from .compiler.enums import PathAspect
+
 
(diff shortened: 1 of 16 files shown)


aljazerzen (commented):
This is definitely makes some snippets span more lines now, but I think it is worth the increased readability.

I do have another nit: we usually use module-qualified identifiers (`irast.PathId` instead of just `PathId`).
Do you think that we should be importing like this:
```python
from edb.pgsql.compiler import enums as pgce

def x(aspect: pgce.PathAspect): ...
```

Arguably, pgce is not a great acronym for "pg compiler enums". But we could then shorten `PathAspect` to just `Aspect`, becuase pg compiler has only this aspect (the other one is more related to delta.py).

aljazerzen (approved):

PR #7505, opened 2024-06-27
Fix nested config objects that add new exclusive props

msullivan:
We use the exclusive properties to generate the id of an object, but
this causes trouble is a subtype adds new ones, since we won't take
that into account when generating the ids in the link.
Account for that properly by only using the target type's props.

No test, but an incoming PR from Scott will exercise this.


elprans (approved):

scotttrinh (approved):
Tested by rebasing my branch on this, and works as expected! Thanks for the super quick fix!

PR #7507, opened 2024-06-28
Docs fix on object types

diksipav:


docs/datamodel/objects.rst
```diff
@@ -11,8 +11,8 @@ analogous to SQL *tables* or ORM *models*, and consist of :ref:`properties
 Properties are used to attach primitive data to an object type. In versions
 prior to EdgeDB 3.0 they must be declared with the ``property`` keyword, but
 the keyword is not required in 3.0+ unless the property is a :ref:`computed
-property <ref_datamodel_computed>`. For the full documentation on properties,
-see :ref:`Properties <ref_datamodel_props>`.
+property <ref_datamodel_computed>` (from 4+ is not required for computeds neither).
+For the full documentation on properties, see :ref:`Properties <ref_datamodel_props>`.
```
scotttrinh, line 15: Actually, we should drop all of the language around the ``property`` keyword. We can show computed properties in the code blocks if we want to be thorough, but having all three historic styles being called out here is totally unnecessary.

scotttrinh (changes requested):
Thanks for jumping in here! I think we should remove all of the language around the keywords for both properties and links.

docs/datamodel/objects.rst
diksipav: Hm ok, I am fine with deleting these altogether. Will do it now

diksipav: I wasn't sure did u mean to write:
```
Properties are used to attach primitive data to an object type. In versions
prior to EdgeDB 3.0 they must be declared with the ``property`` keyword, but
the keyword is not required in 3.0+. For the full documentation on properties...
```

or to completely remove text about property:
```
Properties are used to attach primitive data to an object type. 
For the full documentation on properties...
```
I understood we want the 2nd.

scotttrinh (approved):
Yep, this is perfect!

PR #7508, opened 2024-06-28
Fix migration creation when adding a new base class with certain constraints

msullivan:
If constraints that have an inherited subjectexpr have a new base type
added to the type they are on, we were losing 'subjectexpr' as an
inherited field, because a code path that was intended to only be able
to make fields inherited was sometimes clearing them.

### DIFF
--- a/edb/schema/inheriting.py
+++ b/edb/schema/inheriting.py
@@ -1228,13 +1228,14 @@ class RebaseInheritingObject(
         fields.update(self.EXTRA_INHERITED_FIELDS)
 
         for field in fields:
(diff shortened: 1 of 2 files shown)


edb/schema/inheriting.py
```diff
@@ -1229,11 +1229,12 @@ def compute_inherited_fields(
 
         for field in fields:
             if field not in result:
-                result[field] = (
+                if (
```
elprans, line 1232: collapse the `if`?

elprans (approved):

PR #7509, opened 2024-06-29
Fix interaction of implicit limit with explicit OFFSET

### LINKED ISSUE #3627: implicit limit breaks `offset`
The implicit limit gets injected in *before* the offset

msullivan:
Currently, we will inject a LIMIT at any exposed SELECT, which results
in us applying implicit LIMIT before OFFSET, which is wrong.

Drop the injection inside of SelectStmt compilation and inject the
LIMIT at the top-level instead directly.

This also fixes an issue that always annoyed me where limits would be
inserted multiple redundant times.

Fixes #3627.

### DIFF
--- a/edb/edgeql/compiler/viewgen.py
+++ b/edb/edgeql/compiler/viewgen.py
@@ -1534,7 +1534,6 @@ def _normalize_view_ptr_expr(
 
             if (
                 (ctx.expr_exposed or ctx.stmt is ctx.toplevel_stmt)
-                and not qlexpr.limit
                 and ctx.implicit_limit
                 and not base_is_singleton
             ):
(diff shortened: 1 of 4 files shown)


edb/edgeql/compiler/viewgen.py
```diff
@@ -1534,7 +1534,6 @@ def _normalize_view_ptr_expr(
 
             if (
                 (ctx.expr_exposed or ctx.stmt is ctx.toplevel_stmt)
-                and not qlexpr.limit
```
elprans, line 1537: Why is the check for an explicit `limit` on a shape element dropped?

msullivan: Because giving an explicit limit shouldn't override the explicit limit, the minimum should apply.

In most cases, that is what happened with the existing code, since we would typically inject the implicit limit a second time in the inner SelectStmt, but that is no longer the case, so we need to inject here.

msullivan: In theory, one might argue that an explicit LIMIT should fully override an implicit limit, but I don't think that really holds up: there is no way for the CLI/UI to know about the explicit LIMIT, so it will truncate based on the implicit limit anyway.

elprans (approved):

PR #7510, opened 2024-07-01
Support arbitrary OpenID Connect providers

scotttrinh:
Adds support for arbitrary well-behaved OpenID Connect providers that support discovery. Does not include a PKCE session, which some providers might require. It uses a direct code exchange with a shared client secret. Tested using Google OpenID Connect.

### DIFF
--- a/edb/server/protocol/auth_ext/base.py
+++ b/edb/server/protocol/auth_ext/base.py
@@ -75,12 +75,10 @@ class OpenIDProvider(BaseProvider):
         self,
         name: str,
         issuer_url: str,
-        *args: Any,
-        content_type: ContentType = ContentType.JSON,
-        **kwargs: Any,
+        *args,
+        **kwargs,
     ):
         super().__init__(name, issuer_url, *args, **kwargs)
-        self.content_type = content_type
 
     async def get_code_url(
         self, state: str, redirect_uri: str, additional_scope: str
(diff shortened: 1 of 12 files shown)


edb/server/protocol/auth_ext/base.py
```diff
@@ -70,17 +70,15 @@ class ContentType(enum.StrEnum):
...
 
 
-class OpenIDProvider(BaseProvider):
+class OpenIDConnectProvider(BaseProvider):
     def __init__(
         self,
         name: str,
         issuer_url: str,
         *args: Any,
-        content_type: ContentType = ContentType.JSON,
```
scotttrinh, line 79: The OpenID Connect spec specifies that requests are done via form encoded POST, so I removed this extraneous configuration from the provider since they all support form encoded bodies.

edb/server/protocol/auth_ext/http_client.py
```diff
@@ -50,7 +50,7 @@ async def post(  # type: ignore[override]
         **kwargs: Any,
     ) -> httpx.Response:
         if self.edgedb_orig_base_url:
-            path = f'{self.edgedb_orig_base_url}/{path}'
+            path = f'{self.edgedb_orig_base_url}{path}'
```
scotttrinh, line 53: Our HTTP test URLs had an extra `/` in them due to how we were parsing these URLs. Not really a bug, but when I went to write new tests, I realized that these URLs were wrong, so just fixed this while I was in here. Resulting in some additional churn here.

edb/server/protocol/auth_ext/ui/__init__.py
```diff
@@ -68,7 +66,8 @@ def render_signin_page(
       <input id="email" name="email" type="email" value="{email or ''}" />
     """
 
-    password_input = f"""
+    password_input = (
+        f"""
```
scotttrinh, line 70: Some black formatting churn, apologies. Thought it was worth commit now.

edb/server/protocol/auth_ext/oauth.py
```diff
@@ -39,26 +39,36 @@ def __init__(
         )
 
         provider_config = self._get_provider_config(provider_name)
-        provider_args = (provider_config.client_id, provider_config.secret)
+        provider_args: tuple[str, str] | tuple[str, str, str, str] = (
+            provider_config.client_id,
+            provider_config.secret,
+        )
```
fantix, line 45: maybe we can just inline the following `self.provider = github.GitHubProvider(*provider_args, ...)` in each case branch? I feel like it'll be easier that way

fantix (approved):
LGTM!

edb/server/protocol/auth_ext/oauth.py
scotttrinh: yeah, I think you're right that it's both more straightforward, and probably easier to understand. Will fix.

jaclarke (approved):

PR #7512, opened 2024-07-02
Revert "Skip rebuilding query string if normalization didn't extract variables (#7237)"

msullivan:
This reverts commit 6aef9f57908cb2c0627fa29a187318dcc431f701.

Reverting this for now because it breaks the persistent cache for
queries that don't have any constants extracted. This is because the
serialized source uses a PackedEntry, which always works by
serializing tokens, and so it won't match the non-normalized queries.

### DIFF
--- a/edb/edgeql-parser/edgeql-parser-python/src/normalize.rs
+++ b/edb/edgeql-parser/edgeql-parser-python/src/normalize.rs
@@ -188,6 +188,14 @@ pub fn normalize(text: &str) -> Result<Entry, Error> {
     }
 
     all_variables.push(variables);
+    // N.B: We always serialize the tokens to produce
+    // processed_source, even when no changes have been made. This is
+    // because when Source gets serialized, it always uses a
+    // PackedEntry, which will result in it being normalized *there*,
+    // and so if we don't do it *here*, then we won't be able to hit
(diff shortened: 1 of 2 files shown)


edb/edgeql-parser/edgeql-parser-python/src/normalize.rs
```diff
@@ -159,9 +159,10 @@ pub fn normalize(text: &str) -> Result<Entry, Error> {
                 matches!(kw, "configure"|"create"|"alter"|"drop"|"start"|"analyze")
                 || (last_was_set && kw == "global")
             ) => {
+                let processed_source = serialize_tokens(&tokens);
```
elprans, line 162: Please add a comment here why this is, indeed, needed.

msullivan: *This* one probably isn't really, and I could drop it if you want. (This case is when we are bailing because it is DDL or some such, and can't hit a cache anyway)

I can add a comment to the other case

msullivan: Added

elprans (approved):

PR #7515, opened 2024-07-02
Unbreak cache recompilation after a restart

msullivan:
In #7099, I made it so we only recompile when the schema_version of a 
request matches the current schema_version. Unfortunately, we don't 
serialize the schema_version out, so it was always None.

Bump the CompilationRequest serialization version number and include 
schema_version.


fantix (approved):
LGTM 👍

PR #7524, opened 2024-07-02
Create a task for each DB to ensure we generate the correct load

mmastrac:
The previous QoS tests were "slipping" and generating an insufficient amount of load when the overhead of the pool was too high. This modifies the query generation to create one task per DB and sleep an appropriate amount of time. The tests were unable to hit the query/sec numbers requested and this was causing the QoS numbers to be lower than they should have been. The QoS output for the various tests change to reflect this, but the overall behaviour of the pool is identical.

Note that this had the side-effect of increasing the total number of queries run in most of the tests, which lowering most of the `ConnectionOverhead` scores. The `ConnectionOverhead` scoring now uses "# of disconnections / # of queries" which is a more stable number and can be more easily compared across tests.

Additional fixes:

 - The `LatencyRatio` ratio output was incorrect -- it printed `divisor` twice
 - One test only had a total weight of 0.9, fixed

### DIFF
--- a/tests/test_server_pool.py
+++ b/tests/test_server_pool.py
@@ -241,7 +241,7 @@ class LatencyRatio(PercentileBasedScoreMethod):
         ratio = dividend_percentile / divisor_percentile
         score = self._calculate(ratio)
         sim.record_scoring(
-            f'{self.percentile} ratio {self.divisor}/{self.divisor}',
+            f'{self.percentile} ratio {self.divisor}/{self.dividend}',
             ratio, score, self.weight
         )
         return score * self.weight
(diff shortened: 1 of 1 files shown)


tests/test_server_pool.py
```diff
@@ -240,7 +234,7 @@ def calculate(self, sim: Simulation) -> float:
         ratio = dividend_percentile / divisor_percentile
         score = self._calculate(ratio)
         sim.record_scoring(
-            f'{self.percentile} ratio {self.divisor}/{self.divisor}',
+            f'{self.percentile} ratio {self.divisor}/{self.dividend}',
```
mmastrac, line 237:
```suggestion
            f'{self.percentile} ratio {self.dividend}/{self.divisor}',
```

fantix (approved):
This is awesome, thank you!

PR #7527, opened 2024-07-03
ci: add Ubuntu 24.04 build

fantix:
[Sample build](https://github.com/edgedb/edgedb/actions/runs/9765378684/job/27017938915)


elprans (approved):
We have the same checking script in edgedb-cli, please open a PR to fix that one up as well.

aljazerzen:
There were build problems with libpg_query. Backporting update of libpg_query was non-trivial. So I won't backport this one.

PR #7538, opened 2024-07-09
Move edgedb's fixed internal types to a edgedbt schema

msullivan:
Our internal domains are specifically not versioned, so let's start
making that clear by putting them in an unversioned schema.


aljazerzen (approved):
I'll always read this schema name as "edge dbt" :)

msullivan:
I could name it something else, but as usually I am struggling with a good name

PR #7541, opened 2024-07-10
Basic shape of ext::auth::Event

scotttrinh:
The intention here is that an event will be triggered at certain points in the lifecycle of the various authentication ceremonies and situations. For instance, when you sign up with email+password, you would get `IdentityEventType.Created` and `EmailFactorEventType.Created` events.

I've also included a `event_log_retention_period` configuration value since otherwise the event log might grow unbounded. We have an existing garbage collector so we can hook into that workflow to remove older events based on the `timestamp`. We might also want to have some kind of max_size setting in the future.

### DIFF
--- a/edb/lib/ext/auth.edgeql
+++ b/edb/lib/ext/auth.edgeql
@@ -135,6 +135,95 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
         };
     };
 
+    create abstract type ext::auth::Event extending std::BaseObject {
+        create required property timestamp: std::datetime {
+            set default := std::datetime_current();
+            set readonly := true;
+        };
+    };
+
+    create abstract type ext::auth::IdentityEvent extending ext::auth::Event {
+        create required link identity: ext::auth::Identity {
+            on target delete delete source;
+        };
+    };
+
+    create type ext::auth::IdentityCreated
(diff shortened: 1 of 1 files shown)


edb/lib/ext/auth.edgeql
```diff
@@ -135,6 +135,44 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
...
+        Authenticated,
+    >;
+
+    create abstract type ext::auth::IdentityEvent extending ext::auth::Event {
+        create required property event_type: ext::auth::IdentityEventType;
+
+        create required link identity: ext::auth::Identity {
+            on target delete delete source;
+        };
+    };
```
1st1, line 157: How about making `identity` link optional and moving it into `ext::auth::Event`? IMO for logging kind of types it's more convenient to "compress" all relevant links/properties into just one type.

scotttrinh: There is also the `token` property for some of the events (needed for sending an email with a reset or verification token, for instance). We could also drop that entirely and make it _just_ for the email/webhook and not store it in the event, but I kind of wanted the event to be the source of truth here. But, that was also before the net module existed which has a record of everything anyway.

scotttrinh:
Closed in favor of #7796

PR #7545, opened 2024-07-10
Fix nightly build breakage due to metapkg weirdness

msullivan:
For reasons I haven't been able to remember, metapkg/edgedb-pkg invokes setup.py using an ancient version of Python. (Before eventually invoking it with the correct version.) #7536 started importing tomllib at the top level, but this breaks. Fix it by just importing it at the use site.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,6 @@ import platform
 import shutil
 import subprocess
 import textwrap
-import tomllib
 
 import setuptools
 from setuptools import extension as setuptools_extension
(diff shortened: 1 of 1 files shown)


mmastrac (approved):
Is it possible to smoketest metapkg in the CI builds?

mmastrac:
Is it possible to smoketest metapkg in the CI builds?

mmastrac (approved):
LGTM

msullivan:
> Is it possible to smoketest metapkg in the CI builds?

metapkg builds are much slower than than our main CI runs, so I'm not sure if there's much we can do worthwhile

PR #7556, opened 2024-07-14
feat: add trigonometric functions from postgres

MrFoxPro:
https://www.postgresql.org/docs/current/functions-math.html#FUNCTIONS-MATH-TRIG-TABLE
Took this: https://github.com/edgedb/edgedb/pull/5608 as example for current PR.

Need assistance:
1. Are there special error codes? For reverse functions there is domain restriction -1 <= x <= 1, idk if postgres returns special error for this. https://www.postgresql.org/docs/current/errcodes-appendix.html

2. Need tests. I can't do that because I afraid building and bootstrapping edgedb will take for me too much time, so I wouldn't able to run tests anyway.


dnwpark:
Hi @MrFoxPro, thanks for this useful contribution!

I think the only relevant error will be `22003` which is `NUMERIC_VALUE_OUT_OF_RANGE` which occurs for out of range inputs to acos and asin.

Not immediately relevant, but worth noting that since `tand(90)` returns infinity, if that gets add in the future, some special handling may be necessary.

In terms of tests, I would go ahead and add tests if you can. The CI runs relatively quickly so you can push and let it check things. The query tests check floats with a precision of 15 decimal places so make sure there are enough digits in the assertions!

Lastly, the `EDGEDB_CATALOG_VERSION` in `buildmeta.py` should be updated again.

Please let me know if you have any more questions. If you would like a bit more direct help, I could also make a pr for your pr I think?

dnwpark:
I've added some tests and merged it in #8071

PR #7558, opened 2024-07-15
Consume ReadyForQuery message after an error over SQL

aljazerzen:


msullivan:
This seems basically reasonable to me, but I'd like @fantix to take a look, since I don't understand the state machine fully.

In particular, I don't understand the intricacies of `send_sync_on_error`

edb/server/pgcon/pgcon.pyx
```diff
@@ -2123,8 +2123,8 @@ cdef class PGConnection:
                             self.debug_print("sent backend message: 'Z'")
                         self.write_sync(be_buf)
                         self.write(be_buf)
-                    else:
-                        break
+
+                    # don't break, because we want to also consume ReadyForQuery
```
fantix, line 2127:
```suggestion
                    # don't break, because we want to also exhaust the
                    # remaining buffer with "ignore_till_sync" and consume
                    # ReadyForQuery if there is one
```

fantix (approved):
👍

fantix:
Oh wait I just saw Sully saying `send_sync_on_error`

edb/server/pgcon/pgcon.pyx
fantix: I'm sorry. Now, I think it was right to break here - because it only breaks the inner loop that handles pg responses for the outer loop's current action, and an ErrorResponse does mean the action is resolved.

Then, the outer loop will continue with `ignore_till_sync=True` and skip all the actions until SYNC, which is guaranteed to be there in `pg_ext.simple_query()` (which is the only case when `send_sync_on_error=True`). The ReadyForQuery will then be processed and forwarded to the frontend correctly.

We had an issue because `send_sync_on_error=True` sent an extra SYNC, and we didn't have a corresponding action to consume it.

The first returning value of `_parse_sql_extended_query()` simply means that the frontend should remain in its own `ignore_till_sync` mode. For simple query, this returned value is simply ignored because simple query should always clear `ignore_till_sync` state returning a ReadyForQuery (equivalent to finalizing with SYNC).

fantix (changes requested):

aljazerzen:
Great, thanks fantix.

PR #7562, opened 2024-07-16
Remove inheritance views from schema alias views

dnwpark:
- Used `inheritance.get_inheritance_view` to generate an inheritance select statement.
- Added `sys::Role::member_of` schema alias view to support `DescribeRolesAsDDLFunction`

### DIFF
--- a/edb/pgsql/inheritance.py
+++ b/edb/pgsql/inheritance.py
@@ -22,7 +22,7 @@ def get_inheritance_view(
     ] = frozenset(),
     exclude_ptrs: AbstractSet[s_pointers.Pointer] = frozenset(),
 ) -> pgast.SelectStmt:
-    ptrs: Dict[sn.UnqualName, Tuple[str, Tuple[str, ...]]] = {}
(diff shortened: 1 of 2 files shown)


msullivan (approved):
Looks good!

edb/pgsql/inheritance.py
```diff
@@ -44,22 +44,26 @@ def get_inheritance_view(
...
             ):
                 ptrs[ptrname] = (
-                    ptr_stor_info.column_name,
+                    [ptr_stor_info.column_name],
                     ptr_stor_info.column_type,
                 )
 
+                shortname = ptr.get_shortname(schema).name
+                if shortname != ptr_stor_info.column_name:
+                    ptrs[ptrname][0].append(common.quote_ident(shortname))
```
aljazerzen, line 53: Am I right in that this makes it such that each pointer is represented by two columns in the view?

One named after ptr's uuid and one using ptr shortname?
Well, except things like `id`, `source` and `target` which have only shortnames and no uuids.

We should add a comment explaining this. Or make the ptrs dict contain better types (datastructs with proper names).

PR #7564, opened 2024-07-16
func cache: fix dropping extension with scalar type

fantix:
Scalar types in extensions may define `sql_type` thus not using `common.get_backend_name()`. Instead, here I think we should use `types.pg_type_from_scalar()` to cover more cases.

This is a follow up fix for https://github.com/edgedb/edgedb/pull/7422/commits/d1b2c4d22ee1cd38fe0bb81bc708e6910e33a6b4, captured in [edgedb-js CI](https://github.com/edgedb/edgedb-js/actions/runs/9959418228/job/27516218887?pr=1054#step:17:20). This should've been caught by https://github.com/edgedb/edgedb/pull/7497, I'll double check why it didn't, and make sure either it does, or write a new test in this PR.

Manually tested for enum tuples as well as the pgvector `vector` type.

### DIFF
--- a/edb/pgsql/delta.py
+++ b/edb/pgsql/delta.py
@@ -2832,21 +2832,24 @@ class DeleteScalarType(ScalarTypeMetaCommand,
         cls, scalar: s_scalars.ScalarType, orig_schema: s_schema.Schema
     ) -> dbops.Command:
         ops = dbops.CommandGroup()
-        old_domain_name = common.get_backend_name(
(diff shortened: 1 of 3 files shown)


msullivan (approved):

fantix:
We do need a separate test here. The `EXTENSIONS` in `TestEdgeQLVector` is only created once and reused in the `zREPEAT` test cases, while `test_edgeql_extensions` didn't really address the case we're fixing here.

PR #7583, opened 2024-07-19
Rust connpool: provide connections in LRU/MRU

mmastrac:
Connections were previously provided in a non-deterministic way. This PR changes the connpool to provide connections in MRU for re-use, and LRU for closing idle connections.


fantix (approved):
I like the abstraction, allowing us to tweak the implementation separately if needed in the future. 👍

PR #7584, opened 2024-07-19
Rust connpool: Implement an explicit reconnection phase

mmastrac:
We want to track reconnections for QoS and algorithm purposes -- we want to account for the time spend disconnecting and reconnecting when calculating the expense of switching connections.


fantix (approved):
Thanks @mmastrac for the explanation in PM! So basically:

> * Each connect and each reconnect has a matching disconnect event
> * The calculataion for how long it takes to reconnect is just the greatest of
>   1. the actual average reconnection time (which is a disconnect + connect) and
>   2. the average times to connect and disconnect.
>
>   This just ensures we can get that number even when our stats are sparse. Eventually, `sum(avg(connect) + avg(disconnect)` should roughly equal `avg(reconnect)` once there's enough data.

PR #7585, opened 2024-07-19
Rust connpool: Error handling

mmastrac:
This implements error handling for the Rust connection pool. 

For disconnection errors: all disconnection errors are routed through the task.

For (re-)connection errors: a failed connection holds a spot in a block until someone queues for it. That acquisition request then gets an error rather than a connection. There is no retry logic at this moment, but the intent would be that it would live in the pool itself.

Shutdown is reworked to explicitly close idle and failed connections, as failed connections would not GC otherwise. In a future patch where we improve the GC for idle connections, it might make sense to GC those errors as well.

### DIFF
--- a/edb/server/conn_pool/src/algo.rs
+++ b/edb/server/conn_pool/src/algo.rs
@@ -148,7 +148,7 @@ constants! {
     /// The weight we apply to waiting connections.
     const DEMAND_WEIGHT_WAITING: usize = 3;
     /// The weight we apply to active connections.
-    const DEMAND_WEIGHT_ACTIVE: usize = 277;
+    const DEMAND_WEIGHT_ACTIVE: usize = 4;
     /// The minimum non-zero demand. This makes the demand calculations less noisy
     /// when we are competing at lower levels of demand, allowing for more
     /// reproducable results.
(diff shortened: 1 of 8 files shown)


edb/server/conn_pool/src/algo.rs
```diff
@@ -288,7 +295,8 @@ pub trait PoolAlgorithmDataBlock: PoolAlgorithmDataMetrics {
     /// and there are waiting elements; otherwise, returns `None`.
     fn hunger_score(&self, will_release: bool) -> Option<isize> {
         let waiting = self.count(MetricVariant::Waiting);
-        let connecting = self.count(MetricVariant::Connecting);
+        let connecting =
+            self.count(MetricVariant::Connecting) + self.count(MetricVariant::Reconnecting);
```
mmastrac, line 299: This was missed in the previous reconnecting PR -- including here. Required rebalancing of constants.

fantix (approved):
👍

edb/server/conn_pool/src/block.rs
```diff
@@ -183,22 +199,21 @@ impl<C: Connector, D: Default> Block<C, D> {
     fn create_if_needed(
         self: Rc<Self>,
         connector: &C,
-    ) -> impl Future<Output = ConnResult<ConnHandle<C>>> {
-        if let Some(conn) = self.conns.try_acquire_idle_mru(&self.state.metrics) {
-            return Either::Left(ready(Ok(self.conn(conn))));
+    ) -> impl Future<Output = ConnResult<ConnHandle<C>, C::Error>> {
+        use futures::future::Either;
+        use std::future::ready;
```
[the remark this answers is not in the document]
mmastrac: Oh good catch. These were not imported at the top level when I first added them. Will remove.

mmastrac, line 204:
```suggestion
```

PR #7587, opened 2024-07-22
Fix an issue with certain array types.

vpetrovykh:
When arrays of objects or some scalars appear in the schema it causes errors as these array types are not properly resolved.

Adjust the helper that resolves the backend type based on the EdgeDB type id.

### DIFF
--- a/edb/pgsql/metaschema.py
+++ b/edb/pgsql/metaschema.py
@@ -4221,6 +4221,155 @@ class GetPgTypeForEdgeDBTypeFunction(trampoline.VersionedFunction):
         )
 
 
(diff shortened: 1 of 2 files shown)


vpetrovykh:
The helper `get_pg_type_for_edgedb_type` has trouble handling array types derived from some of the scalars like `range` or new scalars defined in extensions (like `geometry` from postgis).

At the time when `get_pg_type_for_edgedb_type` is defined the `_SchemaType` view is not yet defined, so I ended up updating the `get_pg_type_for_edgedb_type` definition after the schema alias views are defined.
Potentially we can bootstrap the `_SchemaType` view earlier instead, but for me redefining the helper function required a smaller change.

aljazerzen (approved):

PR #7593, opened 2024-07-24
Recompile constraint triggers

dnwpark:
- In `delta`, create a mechanism to schedule updates to triggers after all other changes to the schema (`schedule_constraint_trigger_update`)
    - Ensure that any `create_ops` or `alter_ops`, also schedule updating triggers.
    - If a constraint becomes a relative or stops being the relative of another, also update its triggers.
    - Update mechanism similar to existing post inhview updates.
- Implement the trigger updates using new `schemamech.SchemaTableConstraint.update_trigger_ops` and `deltadbops.AlterTableUpdateConstraintTrigger`
- In `dbops`, remove all create trigger and trigger function ops from create and alter ops.
    - Add `if exists` to any drop trigger and trigger function ops since a trigger may be dropped multiple times by a single command.
- Add tests to check that constraints work after changes to schema.
- Fixed a bug where when changing the bases of an abstract link (either removing or adding), it's triggers and those of its relatives (new and old) are not correctly updated.

Related #7410

### DIFF
--- a/edb/pgsql/delta.py
+++ b/edb/pgsql/delta.py
@@ -214,6 +214,17 @@ class MetaCommand(sd.Command, metaclass=CommandMeta):
     def _get_tenant_id(self, context: sd.CommandContext) -> str:
         return self._get_instance_params(context).tenant_id
 
+    def _get_topmost_command_op(
+        self,
+        context: sd.CommandContext,
+        ctxcls: Type[sd.CommandContextToken[sd.Command]],
+    ) -> CompositeMetaCommand:
+        ctx = context.get_topmost_ancestor(ctxcls)
+        if ctx is None:
+            raise AssertionError(f"there is no {ctxcls} in context stack")
+        assert isinstance(ctx.op, CompositeMetaCommand)
+        return ctx.op
+
     def schedule_inhview_update(
         self,
         schema: s_schema.Schema,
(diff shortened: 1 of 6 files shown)


edb/pgsql/delta.py
```diff
@@ -2019,52 +2044,157 @@ def constraint_is_effective(
...
+    def schedule_relatives_constraint_trigger_update(
+        self,
+        constraint: s_constr.Constraint,
+        orig_schema: s_schema.Schema,
+        curr_schema: s_schema.Schema,
+        context: sd.CommandContext,
     ):
-        base_schema = orig_schema if is_delete else schema
+        # Find all origins whose relationship with the constraint has changed.
+        orig_origins: dict[uuid.UUID, s_constr.Constraint] = {}
```
msullivan, line 2055: My inclination would be to use the `Object` as the key instead of the uuid. It doesn't really matter, but feels cleaner to me. (Behavior is the same.)

edb/pgsql/delta.py
```diff
@@ -2251,17 +2358,19 @@ def apply(self, schema, context):
...
-                     if nschema.has_object(constraint.id)
-                     else None),
+                    (
+                        lambda nschema, ncontext:
+                            self.enforce_constraint(
+                                constraint, nschema, self.span
+                            )
+                            if nschema.has_object(constraint.id)
+                            else None
+                    ),
```
msullivan, line 2368: I think that we might want to schedule these `enforce_constraint`s to happen at the end, at the same time we create the triggers.

Since migrations can contain DML queries, we want to avoid the situation where a conflicting object is inserted *before* we create the trigger. I think that doing the enforce check when we create the trigger might solve that problem?

msullivan: Oh, actually nevermind about that, since the trigger updates get run at the same time as post_inhview updates.

msullivan (approved):
Great work!

edb/pgsql/delta.py
dnwpark: Ah also, this was me reformatting that code since during one of my attempts I ended up doing some pretty unreasonable things with the post inhview commands.

PR #7598, opened 2024-07-25
CI: fix multiple flakiness

fantix:
Please see each commit

### DIFF
--- a/.github/workflows.src/build.targets.yml
+++ b/.github/workflows.src/build.targets.yml
@@ -11,14 +11,12 @@ targets:
           platform_version: buster
           family: debian
           runs_on: [package-builder, self-hosted, linux, x64]
-          docker_arch: linux/amd64
         - name: debian-buster-aarch64
           arch: aarch64
           platform: debian
           platform_version: buster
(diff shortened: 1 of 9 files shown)


.github/workflows.src/build.targets.yml
```diff
@@ -11,14 +11,12 @@ targets:
...
           family: debian
           runs_on: [package-builder, self-hosted, linux, x64]
-          docker_arch: linux/amd64
         - name: debian-buster-aarch64
           arch: aarch64
           platform: debian
           platform_version: buster
           family: debian
           runs_on: [package-builder, self-hosted, linux, arm64]
-          docker_arch: linux/arm64
```
elprans, line 21: Why?

fantix: https://github.com/edgedb/edgedb-docker/pull/91 changed to use `bookworm` for Docker base, and updating `docker_arch` to `debian-bookworm-*` ensures the correct CI job dependency and version referring.

elprans (approved):

PR #7605, opened 2024-07-26
Rust PostgreSQL connection client

mmastrac:
Implements a basic Rust postgres connection library and re-uses its internal state machine to replace the guts of `pgcon`.

User-facing changes:
 - Bugfix: EdgeDB will choose the smaller of server `max_connections` or the maximum number of connections defined for the user's current `role` to prevent hitting connection limits.
 - Change: `sslmode`: SSL is always requested unless sslmode=disable. If the server returns `S` to indicate SSL support and then fails to accept the user credentials, we do not retry without SSL.
 - Feature: Plain-text password authentication in backends is now supported, though discouraged.

Changes:

 - New `PGRawConn` that behaves as a transport and fully connects to postgres before returning the transport to the protocol. This transport can then be used transparently and with zero overhead as it only uses Rust code for connection setup.
 - New `source_description` parameter wherever we use connections directly or indirectly to assist with tracking down what particular part of the system requested a connection. This is not currently propagated into the parameters we send to the server, but we could do this in the future.
 - Major improvements to DSN parsing, as well as moving all DSN code to Rust. There is now a thin shim in `pgconnparams.py`, but it delegates all logic to the Rust side of things.
 - `class CreateParamsKwargs(TypedDict)` is used wherever we hold a bag of connection parameters to assist with typing.
 - `pgaddr` has been removed from the EdgeDB server settings as it was not sufficient to connect to the backend. It has been replaced with a fully-rendered DSN. `edgedb-rust` will need to be updated to support this.

## Testing strategies

DSN parsing: we have a full battery of DSN parsing tests that are compared against libpq. The tests that were previously implemented for pgconnparams have been migrated into this rust codebase. The tests used internally for postgres to test DSN parsing were also extracted and added to the library. On top of this, a number of "hard" examples were added to exercise the code.

Auth code: auth code is tested against a "capture" postgres that we inittb, boot up on a socket and test. Tests for each auth method (scram, md5, password) are tested with and without SSL.

## DSN notes

The DSN parsing code behaves effectively the same as the previous version of the code with a few small differences:

 - Backslash escaping/unescaping in pgpass and the connection string now fully match the behaviour of libpq
 - A few additional warnings were added in cases where an explicit passfile was specified but cannot be read (we still don't warn for implicit passfiles).

## Examples:

Parse a DSN:
```
cargo run --example dsn 'postgres://username@/database_name?host=path-to-socket'
```

Query a running server:
```
cargo run --example connect -- -U foo -P secret -u <<direct path to unix socket .s.PGSQL.5432>>
```


tests/test_backend_connect.py
```diff
@@ -1026,14 +1026,14 @@ def test_connect_pgpass_regular(self):
             # test escaping
             self.run_testcase({
```
mmastrac, line 1027: This testcase was actually incorrect. Backslashes and colons in a postgres:// user/db are treated as literals.

```
psql 'postgres://test\@fgh/test:db?host=/home/matt/.local/share/edgedb/_localdev'
psql: error: connection to server on socket "/home/matt/.local/share/edgedb/_localdev/.s.PGSQL.5432" failed: FATAL:  no pg_hba.conf entry for host "[local]", user "test\", database "test:db", no encryption
```

tests/test_dump_basic.py
```diff
@@ -87,17 +87,16 @@ async def test_dump_fuzz_01(self):
         dbname = self.get_database_name()
         restored_dbname = f'{dbname}_restored'
 
-        with tempfile.TemporaryDirectory() as f:
```
mmastrac, line 90: This test was failing and it appears that it was timing-related. I reordered the create/dump steps to make it more reliable.

edb/server/pgcon/rust_transport.py
```diff
@@ -0,0 +1,475 @@
...
+            pg_state,
+            host if protocol != "unix" else None,
+            ready_future
+        )
+        state.update = update
+        state.drive_initial()
+
+        await ready_future
+        upgraded_transport = update.raw_transport
+        conn = PGRawConn(
```
mmastrac, line 458: @fantix I wonder if it would make sense to just return a raw transport here rather than having `PGRawConn`. The only question I have is how we might be able to expose the DB state (connection params, etc).

edb/server/pgcon/rust_transport.py
```diff
@@ -0,0 +1,475 @@
...
+                self.state.drive_message(memoryview(data))
+            except Exception as e:
+                if error := self.pg_state.server_error:
+                    self.pg_state.server_error = None
+                    self.ready_future.set_exception(pgerror.BackendConnectionError(fields=dict(error)))
+                else:
+                    self.ready_future.set_exception(ConnectionError(e))
+            if self.state.is_ready():
+                self._ready = True
+                self.ready_future.set_result(True)
```
fantix, line 148: Setting result/exception multiple times on the same `Future` object is an error - maybe it's guaranteed on the Rust side, but this still looks concerning on the Python side.

edb/server/pgcon/rust_transport.py
```diff
@@ -0,0 +1,475 @@
...
+
+    def connection_lost(self, exc):
+        if self._ready:
+            self.protocol.connection_lost(exc)
+        else:
+            if not self.ready_future.done:
```
fantix, line 154: `done` is a method - it'll be always true if not called:

```suggestion
            if not self.ready_future.done():
```

edb/server/pgcon/rust_transport.py
```diff
@@ -0,0 +1,475 @@
...
+                if exc:
+                    self.ready_future.set_exception(exc)
+                else:
+                    self.ready_future.set_exception(RuntimeError(
+                        "Connection unexpectedly lost"
+                    ))
```
fantix, line 160: Is this the branch we fall into when shutting down the server while trying to establish a pgcon? It would be kinda expected in that case, right?

edb/server/pgcon/rust_transport.py
```diff
@@ -0,0 +1,475 @@
...
+        ssl.minimum_version = _parse_tls_version(
+            ssl_config['ssl_min_protocol_version'])
+    # OpenSSL 1.1.1 keylog file
+    if hasattr(ssl, 'keylog_filename'):
+        if ssl_config['keylog_filename']:
+            ssl.keylog_filename = ssl_config['keylog_filename']
+    return ssl
+
+
+class PGConnectionProtocol(asyncio.Protocol):
```
fantix, line 113: Let's compare the performance with `asyncio.BufferedProtocol`, which advocates for zero-copy.

edb/server/pgcon/rust_transport.py
```diff
@@ -0,0 +1,475 @@
...
+            pg_state: PGState,
+        ):
+        self.state = state
+        self.pg_state = pg_state
+        self.protocol = protocol
+        self.ready_future = ready_future
+        self._writing_paused = False
+        self._ready = False
+
+    def data_received(self, data: bytes):
```
fantix, line 134: There's also the `eof_received()`; though we didn't quite use it much previously, we may do so later, so let's forward the callback here too so that we don't have to think about it again.

edb/server/pgcon/rust_transport.py
```diff
@@ -0,0 +1,475 @@
...
+            source_description,
+            ConnectionParams._create(state.config),
+            upgraded_transport,
+            pg_state,
+            (host, port)
+        )
+        upgraded_transport.set_protocol(user_protocol)
+
+        # Notify the user protocol of successful connection
+        user_protocol.connection_made(conn)
```
fantix, line 468: How do we make sure that `connection_made()` is always called before any `data_received()` calls? Given the concerns of `ready_future` being set multiple times, I'm thinking if it's possible to combine this logic into the `PGConnectionProtocol` in e.g. the `connection_made()` callback or something, and lift some - if not all - usages of `ready_future`.

edb/server/pgcon/rust_transport.py
```diff
@@ -0,0 +1,475 @@
...
+    ready_future: asyncio.Future = asyncio.Future()
+    pg_state = PGState(
+        parameters={},
+        cancellation_key=None,
+        auth=None,
+        server_error=None,
+        ssl=False
+    )
+
+    user_protocol = protocol_factory()
```
fantix, line 431: This may not be a problem now, but I'd defer the user protocol instantiation to when it's really needed (that also means multiple instantiations during retry), so that we don't get to the wrong case when the same user protocol instance gets multiple `connection_made()` calls in the future by mistake.

fantix (commented):
This is epic! Submitting some comments for the Python code first.

edb/server/pgcon/rust_transport.py
fantix: I think in `PGConnection.connection_made()`, we will have the `PGRawConn` and we can just setup `PGConnection` properties with values from `PGRawConn` there.

edb/server/pgrust/src/connection/mod.rs
```diff
@@ -0,0 +1,202 @@
...
+/// The resolved target of a connection attempt.
+pub enum ResolvedTarget {
+    SocketAddr(std::net::SocketAddr),
+    #[cfg(unix)]
+    UnixSocketAddr(std::os::unix::net::SocketAddr),
+}
+
+impl ResolvedTarget {
+    /// Resolves the target addresses for a given host.
+    pub fn to_addrs_sync(host: dsn::Host) -> Result<Vec<ResolvedTarget>, std::io::Error> {
```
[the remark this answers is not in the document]
mmastrac: Ack, good catch. I will need to move this off-thread. The plan was originally to make use of the tokio blocking pool, but I removed tokio from this PR.

Will address this.

edb/server/pgcon/rust_transport.py
mmastrac: I think it might be best to keep it as asyncio.Protocol. This code will go away at some point in the future and I don't believe we were using it before.

edb/server/pgcon/rust_transport.py
mmastrac: As this code shouldn't be long-lived I think we can skip this. I hope to remove as much of that calling Python code as possible.

tests/test_backend_connect.py
```diff
@@ -1234,21 +674,17 @@ def check():
             await self.con.restore(None, b'', {})
 
     async def test_connection_ssl_to_no_ssl_server(self):
-        ssl_context = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
-        ssl_context.load_verify_locations(SSL_CA_CERT_FILE)
```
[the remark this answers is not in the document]
mmastrac: No need -- the SSL startup message will result in a failure before SSL is used, so this would be a no-op.

edb/server/pgcon/connect.py
```diff
@@ -0,0 +1,213 @@
...
+            connection,
+            lambda: PGConnection(dbname=connection.database),
+            source_description=source_description
+        )
+    except pgerror.BackendConnectionError:
+        # If we don't mandate SSL, try again with no SSL
+        if (connection.sslmode == SSLMode.allow or
+            connection.sslmode == SSLMode.prefer):
+            connection = connection.clone()
+            connection.update(sslmode=SSLMode.disable)
```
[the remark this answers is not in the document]
mmastrac: Per conversation on slack, this behaviour will be documented in changelog.

edb/server/pgcon/connect.py
```diff
@@ -0,0 +1,213 @@
...
+        connection = ConnectionParams(dsn=dsn_or_connection)
+    else:
+        connection = dsn_or_connection
+    try:
+        pgrawcon, pgcon = await create_postgres_connection(
+            connection,
+            lambda: PGConnection(dbname=connection.database),
+            source_description=source_description
+        )
+    except pgerror.BackendConnectionError:
```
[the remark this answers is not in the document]
mmastrac: Per conversation on slack, retry is being removed.

edb/server/pgcon/rust_transport.py
```diff
@@ -0,0 +1,475 @@
...
+        self._pg_state.server_error = error
+
+
+async def _create_connection_to(
+        protocol_factory: Callable[[], asyncio.Protocol],
+        protocol: str,
+        host: str | bytes,
+        hostname: str,
+        port: int
+    ) -> Tuple[str, str, int, asyncio.Transport]:
```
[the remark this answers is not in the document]
mmastrac: Reformatted whole file with black :)

edb/server/pgcon/rust_transport.py
```diff
@@ -0,0 +1,475 @@
...
+    def auth(self, auth: int) -> None:
+        self._pg_state.auth = Authentication(auth)
+
+    def server_error(self, error: list[tuple[str, str]]) -> None:
+        self._pg_state.server_error = error
+
+
+async def _create_connection_to(
+        protocol_factory: Callable[[], asyncio.Protocol],
+        protocol: str,
```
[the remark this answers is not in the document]
mmastrac: Good catch, fixed

edb/server/pgcon/rust_transport.py
```diff
@@ -0,0 +1,475 @@
...
+from typing import Optional, List, Tuple, Protocol, Callable, Dict, Any, TypeVar
+from enum import Enum, auto
+from edb.server._pg_rust import PyConnectionState
+from dataclasses import dataclass
+from edb.server.pgconnparams import (
+    ConnectionParams,
+    SSLMode,
+    get_pg_home_directory
+)
+from . import errors as pgerror
```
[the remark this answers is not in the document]
mmastrac: Partially applied this one by hand, thanks!

edb/server/pgcon/connect.py
```diff
@@ -0,0 +1,213 @@
...
+import logging
+import textwrap
+
+from edb.server.pgconnparams import SSLMode
+from edb.pgsql.common import quote_ident as pg_qi
+from edb.pgsql.common import quote_literal as pg_ql
+from edb.pgsql import params as pg_params
+from . import errors as pgerror
+from .rust_transport import ConnectionParams, create_postgres_connection
+from .pgcon import PGConnection
```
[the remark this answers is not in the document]
mmastrac: Applied by hand, thanks

edb/server/pgcon/__init__.pyi
```diff
@@ -21,11 +21,12 @@ from __future__ import annotations
 from typing import (
     Any,
     Callable,
-    Dict,
     Optional,
 )
 
 from edb.pgsql import params as pg_params
+from edb.server import pgconnparams
+from .rust_transport import ConnectionParams
```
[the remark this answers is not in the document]
mmastrac: I reworked these files a bit in the latest changes -- originally we had the top-level pyi and that was causing issues with the type checker. I'm now just providing pgcon.pyi which fixes those problems.

edb/server/pgcon/__init__.pyi
```diff
@@ -45,9 +46,10 @@ class BackendCatalogNameError(BackendError):
     ...
 
 async def connect(
```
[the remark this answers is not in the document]
mmastrac: Reworked this file

edb/server/pgrust/src/connection/mod.rs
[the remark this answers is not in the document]
mmastrac: To fix this, I used an executor to resolve the candidates, and in the property we drop the GIL so that we can effectively run it in parallel. The uvloop resolution works in a similar way and this should be sufficiently performant (it'll only be called for connection setup) until I move more of this code to rust.

edb/server/pgcon/rust_transport.py
```diff
@@ -0,0 +1,475 @@
...
+        return (protocol, hostname, port, t)
+
+
+async def _create_connection(
+        protocol_factory: Callable[[], asyncio.Protocol],
+        connect_timeout: Optional[int],
+        host_candidates: List[Tuple[str, str | bytes, str, int]]
+    ) -> Tuple[str, str, int, asyncio.Transport]:
+    e = None
+    for protocol, host, hostname, port in host_candidates:
```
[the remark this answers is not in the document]
mmastrac: That's pretty much it -- `localhost` might resolve to `127.0.0.1` and `::1`, but the server might only be listening on IPv4 _or_ IPv6. We need to try all provided addresses in sequence to find the first that connects.

edb/server/pgcon/rust_transport.py
mmastrac: With the latest changes, we only create this at the last moment

edb/server/pgcon/rust_transport.py
mmastrac: Everything now happens in PGConnectionProtocol which should hopefully make this more obviously correct.

edb/server/pgcon/rust_transport.py
mmastrac: All the cases where we might set the exception now test not done() first

edb/server/pgcon/rust_transport.py
mmastrac: I changed this to raise BackendConnectionError (without any fields set)

edb/server/pgcon/rust_transport.py
```diff
@@ -0,0 +1,475 @@
...
+        async with asyncio.timeout(connect_timeout if connect_timeout else 60):
+            try:
+                return await _create_connection_to(
+                    protocol_factory,
+                    protocol,
+                    host,
+                    hostname,
+                    port
+                )
+            except asyncio.CancelledError as ex:
```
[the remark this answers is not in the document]
mmastrac: Rewrote this to catch asyncio.TimeoutError

edb/server/pgcon/rust_transport.py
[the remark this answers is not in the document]
fantix: I pushed a fix, making the generated SQL slightly better formatted, too.

edb/server/pgcon/rust_transport.py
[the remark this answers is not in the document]
fantix: Ohh sorry I mean why rewrite what's done in `loop.create_connection()` and have Rust doing the resolution separately?

edb/server/pgrust/src/connection/mod.rs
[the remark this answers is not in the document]
fantix: > an executor

Yup, this would effectively flip the `PY_ALLOW_THREADS` macros from no-op into actual overheads. But

> until I move more of this code to rust.

I assume the Python thread pool will go away after this, so I think it's fine that we have the overhead for a while.

fantix (approved):
Epic! LGTM

(There're a few `println!("Read:");` in the future connection Rust code; a kind reminder to replace them in the next PR)

Issue #7294, opened 2024-05-02
Support writes in the SQL adapter


msullivan:
The trickiness here is in supporting all of the various special edgeqlisms that affect writes:
 * Rewrites
 * Triggers
 * Access policies? (We'd probably want that to work on reads also -- which is doable)
 * Cross-table exclusive constraints
 * Built-in foreign keys
 * `required multi` pointers

My theory for how to implement this, which is not fully formed, is to turn the DML operations in the SQL queries into not-DML CTEs, and then basically use the edgeql compiler to compile edgeql DML that is rigged up to take those CTEs as input. (Using a generalization of external rvars, probably).
That should then let all our hooks operate.

We'd also probably need to compile in assertions that link uuids actually map to a real object.

There are still a bunch of questions, especially around link tables, but I think this *should* be workable.

msullivan:
One thing we need to figure out an answer for is how to specify global variables to the SQL protocol. The *simplest* approach would be to expose it as a config value where you specify JSON encoded globals, but we might want something smarter.

aljazerzen:
My strategy is to first implement INSERT and then DELETE and UPDATE with reusing code blocks / functions from the INSERT implementation.

Todo for INSERT:
- [ ] multiple values (properly set for iterator),
- [ ] links (`INSERT "Movie" (director_id) VALUES ("61e477c9-5ef7-419f-804c-86f5f1a2d78e"::uuid)`),
- [ ] make sure that triggers run at the end of the top-level-query, not at the end of current DML query,
- [ ] DML in CTEs (push generated CTEs one level up so they are always top-level, and reject DML in non-top-level CTEs)
- [ ] reuse alias generator between compiler invocations to prevent name conflicts,
- [ ] RETURNING values,

aljazerzen:
Modified rows in query result:

When INSERT is the top-level statement in the query, Postgres response will contain a [CommandComplete (B)](https://www.postgresql.org/docs/current/protocol-message-formats.html#PROTOCOL-MESSAGE-FORMATS-COMMANDCOMPLETE) with tag `INSERT oid rows`, where `oid` is always 0 and `rows` is number of inserted rows.

Because we compile INSERT into SELECT with many CTEs, INSERT is not top-level statement in the query, which means that the postgres response will not contain the appropriate tag.

Also, if the INSERT did not specify RETURNING columns, there should be no result. Again, because we compile to SELECT with CTEs, that will produce a result, with no columns.

[...]

PR #7608, opened 2024-07-30
INSERT into link tables over SQL adapter

aljazerzen:
Ref #7294

### DIFF
--- a/edb/pgsql/resolver/command.py
+++ b/edb/pgsql/resolver/command.py
@@ -41,6 +41,7 @@ from edb.schema import pointers as s_pointers
 from edb.schema import links as s_links
 from edb.schema import properties as s_properties
 from edb.schema import name as sn
+from edb.schema import utils as s_utils
 
 
 from . import dispatch
@@ -198,6 +199,12 @@ def _preprocess_insert_stmt(
     Takes SQL INSERT query and produces an equivalent EdgeQL insert query
     and a bunch of metadata needed to extract associated CTEs from result of the
     EdgeQL compiler.
+
+    In this context:
+    - subject is the pointer being updated,
+    - source is the source of the subject (when subject is a link / property),
+    - value is the relation that provides new value to be inserted,
+    - ptr-s are (usually) pointers on the subject.
     """
 
     # determine the subject object we are inserting into
@@ -220,13 +227,9 @@ def _preprocess_insert_stmt(
         res = _preprocess_insert_object_stmt(
             stmt, sub, sub_table, expected_columns, ctx=ctx
         )
-    elif isinstance(sub, s_links.Link):
-        res = _preprocess_insert_link_stmt(
+    elif isinstance(sub, (s_links.Link, s_properties.Property)):
+        res = _preprocess_insert_pointer_stmt(
             stmt, sub, sub_table, expected_columns, ctx=ctx
-        )
-    elif isinstance(sub, s_properties.Property):
-        res = _preprocess_insert_property_stmt(
-            stmt, sub, expected_columns, ctx=ctx
         )
     else:
         raise NotImplementedError()
@@ -253,6 +256,9 @@ def _preprocess_insert_object_stmt(
     *,
     ctx: Context,
 ) -> PreprocessedDML:
+    """
+    Translates a 'SQL INSERT into an object type table' to an EdgeQL insert.
+    """
 
     # handle DEFAULT and prepare the value relation
     value_relation, expected_columns = _preprocess_insert_value(
@@ -323,10 +329,7 @@ def _preprocess_insert_object_stmt(
     sub_name = sub.get_name(ctx.schema)
     assert isinstance(sub_name, sn.QualName)
     ql_stmt: qlast.Expr = qlast.InsertQuery(
-        subject=qlast.ObjectRef(
-            name=sub_name.name,
-            module=sub_name.module,
-        ),
+        subject=s_utils.name_to_ast_ref(sub_name),
         shape=insert_shape,
     )
     if not is_value_single:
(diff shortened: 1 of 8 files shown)


edb/pgsql/resolver/command.py
```diff
@@ -339,6 +369,409 @@ def _preprocess_insert_stmt(
...
+            output_var = pgast.ColumnRef(name=(ptr_name,), nullable=True)
+            value_rel.path_outputs[(ptr_id, pgce.PathAspect.VALUE)] = output_var
+            value_rel.path_outputs[(ptr_id, pgce.PathAspect.IDENTITY)] = (
+                output_var
+            )
+        else:
+            # link pointer
+            ptr_name = expected_col.name
+            ptr = link.maybe_get_ptr(ctx.schema, sn.UnqualName(ptr_name))
+            assert ptr
```
msullivan, line 500: We want a real error message here, I think? (Unless it is done elsewhere?)

edb/pgsql/resolver/command.py
```diff
@@ -339,6 +369,409 @@ def _preprocess_insert_stmt(
...
+        else qlast.ObjectRef(name=iterator_name)
+    )
+
+    # a phantom relation that is supposed to hold the inserted value
+    # (in the resolver, this will be replaced by the real value relation)
+    value_cte_name = ctx.alias_generator.get('ins_value')
+    value_rel = pgast.Relation(name=value_cte_name)
+    value_columns: List[Tuple[str, bool]] = []
+    for expected_col in expected_columns:
+        if expected_col.name == 'source':
```
msullivan, line 475: Can these cases be unified more? I think there might be a cleanup possible, but no problem if not

edb/pgsql/resolver/command.py
```diff
@@ -339,6 +369,409 @@ def _preprocess_insert_stmt(
...
+    )
+
+
+def _preprocess_insert_property_stmt(
+    stmt: pgast.InsertStmt,
+    prop: s_properties.Property,
+    expected_columns: List[context.Column],
+    *,
+    ctx: Context,
+) -> PreprocessedDML:
```
msullivan, line 645: I think it should be possible to unify the property and link cases and have it simplify things more than it complicates them?
(Maybe I'm wrong, but probably worth trying for a bit)

edb/pgsql/resolver/command.py
```diff
@@ -339,6 +369,409 @@ def _preprocess_insert_stmt(
...
+    )
+    if not is_value_single:
+        # value relation might contain multiple rows
+        # to express this in EdgeQL, we must wrap `insert` into a `for` query
+        ql_stmt = qlast.ForQuery(
+            iterator=qlast.Path(steps=[qlast.IRAnchor(name=value_name)]),
+            iterator_alias=iterator_name,
+            result=ql_stmt,
+        )
+    # ql_stmt = qlast.Path(steps=[ql_stmt, qlast.Ptr(name=sub_name.name)])
```
msullivan, line 591: Since we are generating an UPDATE, can we do it without the iterator?
It might still be easiest to do it with the iterator, though.

edb/pgsql/resolver/command.py
```diff
@@ -339,6 +369,409 @@ def _preprocess_insert_stmt(
...
+    # Here we only decide on the name of that iterator column, the actual column
+    # is generated later, when resolving the DML stmt.
+    value_iterator = ctx.alias_generator.get('iter')
+    output_var = pgast.ColumnRef(name=(value_iterator,))
+    value_rel.path_outputs[(source_id, pgce.PathAspect.ITERATOR)] = output_var
+    value_rel.path_outputs[(value_id, pgce.PathAspect.ITERATOR)] = output_var
+
+    # construct the EdgeQL DML AST
+    sub_name = link.get_name(ctx.schema)
+    assert isinstance(sub_name, sn.QualName)
```
msullivan, line 519: These asserts shouldn't be needed, I think? `Link` is a `QualifiedObject`, so its name is always a QualName.
(If it is more about convincing `pylance` than `mypy`, maybe put an annotation on the variable instead of adding an assert?)

edb/pgsql/resolver/command.py
```diff
@@ -339,6 +369,409 @@ def _preprocess_insert_stmt(
...
+    assert isinstance(sub_source_name, sn.QualName)
+    sub_target_name = sub_target.get_name(ctx.schema)
+    assert isinstance(sub_target_name, sn.QualName)
+
+    sub_name = link.get_shortname(ctx.schema)
+
+    ql_sub_source_ref = qlast.ObjectRef(
+        name=sub_source_name.name,
+        module=sub_source_name.module,
+    )
```
msullivan, line 530: Maybe use `edb.schema.utils.name_to_ast_ref`

msullivan (approved):

edb/pgsql/resolver/command.py
aljazerzen: No, if this fails, it's a bug.

The ptr_name is coming from a `context.Column`, which means that this column does exist. If this fails, it means that for a given SQL column, we don't know how to map it to an object property.

edb/pgsql/resolver/command.py
aljazerzen: They *could* possibly be more unified, but I don't think they should.

There are many small differences. I've started with one large function with many if stmts to choose between the three cases, but it got way too messy, so I decided to have some simple bits of code duplicated for the sake of readability.

edb/pgsql/resolver/command.py
aljazerzen: I've managed to combine link and multi-property functions.

edb/pgsql/resolver/command.py
aljazerzen: I don't think so.

The value relation might contains multiple different `source` values, which cannot be translated to a non-iterated update.

edb/pgsql/resolver/command.py
aljazerzen: You were right.

I think I've manually went and looked at the definition of `name = so.Field(...)` on `so.Object` and derived the assert from there.

PR #7612, opened 2024-07-31
DELETE object table over SQL adapter

aljazerzen:
Support for DELETE statement over SQL adapter.

TODO tracked here: https://github.com/edgedb/edgedb/issues/7294#issuecomment-2260120714

### DIFF
--- a/edb/pgsql/parser/ast_builder.py
+++ b/edb/pgsql/parser/ast_builder.py
@@ -293,8 +293,9 @@ def _build_select_stmt(n: Node, c: Context) -> pgast.SelectStmt:
 def _build_insert_stmt(n: Node, c: Context) -> pgast.InsertStmt:
     return pgast.InsertStmt(
         relation=_maybe(n, c, "relation", _build_rel_range_var),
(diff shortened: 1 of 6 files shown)


edb/pgsql/parser/ast_builder.py
```diff
@@ -369,27 +377,37 @@ def _build_variable_set_stmt(n: Node, c: Context) -> pgast.Statement:
...
-            scope=pgast.OptionsScope.TRANSACTION
-            if n["name"] in tx_only_vars or "is_local" in n and n["is_local"]
-            else pgast.OptionsScope.SESSION,
+            scope=(
+                pgast.OptionsScope.TRANSACTION
+                if n["name"] in tx_only_vars
+                or "is_local" in n
+                and n["is_local"]
+                else pgast.OptionsScope.SESSION
+            ),
```
msullivan, line 397: While here, could you add some parens? I don't like relying on the `or`/`and` precedence rule.
(I *think* and binds tighter (because it is more like multiplication?))

msullivan (approved):

PR #7626, opened 2024-08-07
Refactor static func eval over SQL adaptor

aljazerzen:
Move a few local constants out of functions, mainly to improve performance.


msullivan (approved):
Add a commit message

PR #7627, opened 2024-08-07
UPDATE of object tables over SQL adapter

aljazerzen:
Ref #7294

### DIFF
--- a/edb/pgsql/resolver/command.py
+++ b/edb/pgsql/resolver/command.py
@@ -1213,20 +1213,11 @@ def _preprocess_update_object_stmt(
     )
     stmt.ctes = []
 
-    # prepare anchors for inserted value columns
-    value_name = ctx.alias_generator.get('upd_val')
-    value_id = irast.PathId.from_type(
-        ctx.schema,
-        sub,
-        typename=sn.QualName('__derived__', value_name),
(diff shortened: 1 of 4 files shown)


edb/pgsql/resolver/command.py
```diff
@@ -1099,6 +1099,245 @@ def _preprocess_delete_pointer_stmt(
...
+
+    # prepare anchors for inserted value columns
+    value_name = ctx.alias_generator.get('ins_val')
+    iterator_name = ctx.alias_generator.get('ins_iter')
+    value_id = irast.PathId.from_type(
+        ctx.schema,
+        sub,
+        typename=sn.QualName('__derived__', value_name),
+        env=None,
+    )
```
msullivan, line 1235: Some weird code duplication here

msullivan (approved):
... diff really does not do a good job helping understand the refactor here

aljazerzen:
No, it does not.

The cleanup was in the second part (`dispatch.resolve`) of resolving. Instead of having `_resolve_InsertStmt`, `_resolve_DeleteStmt` and `_resolve_UpdateStmt`, I was able to generalize down to just `_resolve_DMLQuery`.

Maybe this should be two separate PRs.

PR #7628, opened 2024-08-07
Test infrastructure improvements

mmastrac:
Various test infrastructure improvements extracted from PR #7544:

 - The test server now properly redirects output to the exceptions, even if `EDGEDB_DEBUG_SERVER` is set
 - Adds a `test.async_timeout` decorator to simplify test timeouts

### DIFF
--- a/edb/testbase/server.py
+++ b/edb/testbase/server.py
@@ -2440,7 +2440,7 @@ class _EdgeDBServer:
         self.proc: asyncio.Process = await asyncio.create_subprocess_exec(
             *cmd,
             env=env,
-            stdout=subprocess.PIPE if not self.debug else None,
+            stdout=subprocess.PIPE,
             stderr=subprocess.STDOUT,
             pass_fds=(status_w.fileno(),),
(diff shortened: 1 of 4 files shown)


edb/testbase/server.py
```diff
@@ -2451,6 +2451,32 @@ async def __aenter__(self):
...
+
+        output = b''
+
+        async def read_stdout():
+            nonlocal output
+            # Tee the log temporarily to a tempfile that exists as long as the
+            # test is running. This helps debug hanging tests.
+            with tempfile.NamedTemporaryFile(mode='w+t',
+                                             prefix='edgedb-test-log-',
+                                             delete=False) as temp_file:
```
msullivan, line 2463: Why `delete=False` and a manual `remove`?

msullivan (approved):

edb/testbase/server.py
mmastrac: Ack, that was an oversight, will fix before landing.

PR #7632, opened 2024-08-08
DELETE/UPDATE ONLY over SQL adapter

aljazerzen:
Ref #7294

### DIFF
--- a/edb/pgsql/resolver/command.py
+++ b/edb/pgsql/resolver/command.py
@@ -868,13 +868,22 @@ def _uncompile_delete_object_stmt(
 
     # construct the EdgeQL DML AST
     sub_name = sub.get_name(ctx.schema)
+    where = qlast.BinOp(
+        left=qlast.Path(partial=True, steps=[qlast.Ptr(name='id')]),
+        op='IN',
+        right=qlast.Path(steps=[value_ql, qlast.Ptr(name='id')]),
+    )
+    if not stmt.relation.include_inherited:
+        # DELETE ONLY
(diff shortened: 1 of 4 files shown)


aljazerzen (commented):
```sql
DELETE ONLY Foo USING Bar WHERE Foo.owner = Bar.id

-- should be compiled to:
SELECT id FROM ONLY Foo, Bar WHERE Foo.owner = Bar.id

-- so include_inherited=False
```

edb/pgsql/resolver/command.py
```diff
@@ -1345,6 +1363,31 @@ def is_default(e: pgast.BaseExpr) -> bool:
...
+        right=qlast.TypeCast(
+            expr=qlast.TypeCast(
+                expr=qlast.Constant.string(str(sub.id)),
+                type=qlast.TypeName(maintype=qlast.ObjectRef(name='uuid'))
+            ),
+            type=qlast.TypeName(
+                maintype=qlast.ObjectRef(module='schema', name='Type')
+            )
+        )
+    )
```
msullivan, line 1388: We should do either:
- `.__type__ = (introspect [type name])`
- `.__type__.id = <uuid>'[type id]'`

Both should skip having to do an actual lookup.

msullivan (approved):

edb/pgsql/resolver/command.py
aljazerzen: Yes, this would be much better option.

I've removed this whole chunk anyway, since I can use ONLY on value relation and filter out child types there.

PR #7636, opened 2024-08-08
Enforce rustfmt and clippy on all Rust code

mmastrac:
This starts enforcing rustfmt and clippy on all Rust code, and adds a test to `tests/test_sourcecode.py` to test this at build time.

~The `graphql-rewrite` codebase is removed as part of this work -- we can recover it from git history if we revisit it in the future.~


elprans:
> The `graphql-rewrite` codebase is removed as part of this work -- we can recover it from git history if we revisit it in the future.

Any specific reason for that.  It isn't dead code, we use it [here](https://github.com/edgedb/edgedb/blob/be333a4e7d91c082bbe83a6e1a6caa490652893e/edb/graphql/extension.pyx#L287).

mmastrac:
> > The `graphql-rewrite` codebase is removed as part of this work -- we can recover it from git history if we revisit it in the future.
> 
> Any specific reason for that? It isn't dead code, we use it [here](https://github.com/edgedb/edgedb/blob/be333a4e7d91c082bbe83a6e1a6caa490652893e/edb/graphql/extension.pyx#L287).

Ah, I missed that it was used. I can restore that code.

elprans (approved):

tests/test_sourcecode.py
```diff
@@ -107,3 +107,54 @@ def test_cqa_mypy(self):
                     output += '\n\n' + ex.stderr.decode()
                 raise AssertionError(
                     f'mypy validation failed:\n{output}') from None
+
+    def test_clippy(self):
```
fantix, line 111: These 2 tests are failing on nightlies (where we run tests with released binary):

https://github.com/edgedb/edgedb/actions/runs/10481913983/job/29032928532

You can test this by pushing code into a branch in the main repo and manually triggering the `Dry Run` CI.

PR #7637, opened 2024-08-12
Replace yanked `types-pkg-resources` dependency with `types-setuptools`

jaclarke:
It seems the `types-pkg-resources` package no longer exists: https://pypi.org/project/types-pkg-resources/, so replace it with the suggested `types-setuptools` package. The replacement seems to work without any errors, but I'm not sure exactly what this package does in edgedb, so there may be some issues I don't know about.


mmastrac (approved):
LGTM - this fix also works for me and it appears to have prior art in other projects https://github.com/sqlfluff/sqlfluff/pull/6039.

fantix (approved):
Yeah, this should be fine. It's under the `test` group, so it won't be installed in production. Also, the passing test is also a thumbs-up.

PR #7639, opened 2024-08-12
Allow for GC interval in Rust connection pool

mmastrac:
Instead of aggressively closing all connections when idle, we now perform regular GC sweeps of the pool and scan for any idle connection older than the GC age.


fantix (approved):
Perfect!

PR #7641, opened 2024-08-13
Remove module from globals' internal alias names.

### LINKED ISSUE #7516: When an object global contains a shape, the introspected `name` contains extraneous `__` before the module name
- EdgeDB Version: 5.4

Schema:

```esdl
module default {
  global global_object := (select O { name } limit 1);

  type O {
    name: str;
  }
}
```

Query:

```edgeql
with module schema
select Type {
  id,
  name,
} filter .name ilike "%global_object";
```

The resulting name is: `__default::global_object` instead of `default::global_object`.

dnwpark:
Close #7516

Cause of the issue:
- When creating a `global` ie. `CreateGlobal.apply`:
    - During `CreateAliasLike._create_begin`
    - `alias_name` is set to `GlobalCommand._get_alias_name`
        - This mangles `classname`, which adds some `'@'`s in the process
- When compiling the alias expression in `compile_alias_expr`:
    - A persistent type is generated in `_process_view`
    - The `generated_name` is prefixed by `__`
- When updating the schema, in `write_meta_create_object`:
    - During `_build_object_mutation_shape`, when processing `n == 'name'`
    - `target_value` is set to `Object.get_displayname_static(v)`
        - Calls `QualifiedObject.get_shortname_static`
        - Calls `schema.name.shortname_from_fullname`
        - This unmangles the name after splitting by `'@'`, but preserves the module

This PR removes the module from the alias name produced for the global. This allows the unmangled form to not prefix the module with `'__'`.

### DIFF
--- a/edb/schema/expraliases.py
+++ b/edb/schema/expraliases.py
@@ -174,8 +174,17 @@ class AliasLikeCommand(
         return obj
 
     @classmethod
-    def _mangle_name(cls, type_name: sn.QualName) -> sn.QualName:
-        base_name = type_name
+    def _mangle_name(
+        cls,
+        type_name: sn.QualName,
+        *,
+        include_module_in_name: bool,
+    ) -> sn.QualName:
+        base_name = (
+            type_name
+            if include_module_in_name else
+            type_name.get_local_name()
+        )
         quals = (cls.get_schema_metaclass().get_schema_class_displayname(),)
         pnn = sn.get_specialized_name(base_name, str(type_name), *quals)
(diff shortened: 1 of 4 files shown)


aljazerzen (commented):
Great break down of what went wrong!

Can you add the following test:

```
create module my_mod;
create global my_mod::my_global -> str;

select schema::Type { name }
filter .name ilike "%my_global%";
```

dnwpark:
@aljazerzen Good call!

aljazerzen (approved):

PR #7653, opened 2024-08-16
Rename sys::Database to sys::Branch

fantix:
Rename `sys::Database` to `sys::Branch`. Also, make the old name an alias of the new name.

~This PR contains the minimal changes to make it work. We still leave behind lots of "database" terms in the code.~


fantix:
Seems like the `CREATE ALIAS` command failed:

```
CREATE ALIAS sys::Database := sys::Branch;
```

```
RuntimeError: non-abstract object type 'sys::Database' has non-abstract ancestors
```

elprans:
> Seems like the `CREATE ALIAS` command failed:
> 
> ```
> CREATE ALIAS sys::Database := sys::Branch;
> ```
> 
> ```
> RuntimeError: non-abstract object type 'sys::Database' has non-abstract ancestors
> ```

This is not because of `ALIAS`, this is because you also need to rename `class Database` to `class Branch`.

edb/schema/database.py
```diff
@@ -36,22 +36,22 @@
...
 ):
     pass
 
 
-class DatabaseCommandContext(sd.ObjectCommandContext[Database]):
+class DatabaseCommandContext(sd.ObjectCommandContext[Branch]):
     pass
 
 
 class DatabaseCommand(
```
elprans, line 53: These should be renamed also.

msullivan (approved):
This is *technically* not backward compatible, as now you can't do thinks like `[is sys::Database]`, but oh well.

(We should maybe support that also)

fantix:
@elprans wdyt?

fantix:
I'm merging this first; we can change it later again.

PR #7659, opened 2024-08-19
Rust connection string parser

mmastrac:
Extracted from #7605

This implements a connection string parser in Rust and replaces the older Python DSN parser.

The DSN parsing code behaves effectively the same as the previous version of the code with a few small differences:

 -  Backslash escaping/unescaping in pgpass and the connection string now fully match the behaviour of libpq
 -  A few additional warnings were added in cases where an explicit passfile was specified but cannot be read (we still don't warn for implicit passfiles).

What's left in the Python DSN parsing code is just SSL setup, which will eventually be migrated to Rust as well.


tests/test_backend_connect.py
```diff
@@ -1026,14 +1026,14 @@ def test_connect_pgpass_regular(self):
             # test escaping
             self.run_testcase({
                 'dsn': 'postgres://{}@fgh/{}?passfile={}'.format(
-                    R'test\\', R'test\:db', passfile.name
+                    'test\\', 'test:db', passfile.name
```
mmastrac, line 1029: This testcase was actually incorrect. Backslashes and colons in a postgres:// user/db are treated literally as seen below:

```
psql 'postgres://test\@fgh/test\:db?host=/home/matt/.local/share/edgedb/_localdev'
psql: error: connection to server on socket "/home/matt/.local/share/edgedb/_localdev/.s.PGSQL.5432" failed: FATAL:  no pg_hba.conf entry for host "[local]", user "test\", database "test\:db", no encryption
```

edb/server/pgrust/Cargo.toml
```diff
@@ -0,0 +1,67 @@
...
+edition = "2021"
+
+[lint]
+workspace = true
+
+[features]
+python_extension = ["pyo3/extension-module", "pyo3/serde"]
+optimizer = []
+
+[dependencies]
```
mmastrac, line 15: There are more dependencies here than necessary but we'll land the rest when #7605 lands

edb/server/pgrust/src/conn_string.rs
```diff
@@ -0,0 +1,978 @@
...
+                passfile
+            }
+            Password::Specified(_) => return Ok(None),
+            Password::Passfile(passfile) => {
+                let passfile = passfile.clone();
+                if matches!(passfile.try_exists(), Ok(false)) {
+                    *self = Password::Unspecified;
+                    return Ok(Some(PasswordWarning::NotExists(passfile)));
+                }
+                if passfile.exists() && !passfile.is_file() {
```
fantix, line 123: `exists()` duplicates with the `try_exists()` above?

fantix (approved):
Overall this is great! Some nitpickings

edb/server/pgrust/src/conn_string.rs
mmastrac: It's not quite the same -- `try_exists` will return `Ok(...)` if and only if the file can be guaranteed to exist or not exist, while `exists()` returns `false` if the file definitely doesn't exist OR there was an error. This is somewhat awkward code that just preserves the existing error functionality so the tests pass with the correct failures.

edb/server/pgconnparams.py
```diff
@@ -66,195 +62,23 @@ class ConnectionParameters:
...
         # home directory, whereas Postgres stores its config in
         # %AppData% on Windows.
         buf = ctypes.create_unicode_buffer(ctypes.wintypes.MAX_PATH)
         r = ctypes.windll.shell32.SHGetFolderPathW(  # type: ignore
             0, CSIDL_APPDATA, 0, 0, buf)
         if r:
-            return None
+            # Fall back to the home directory
+            warnings.warn("Could not resolve %AppData%", stacklevel=2)
+            return pathlib.Path.home()
```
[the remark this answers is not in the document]
mmastrac: Good point. I switched this to raise an exception -- it is an extremely unlikely code path, and only on Windows in the case where the OS has spectacularly failed.

edb/server/pgrust/src/conn_string.rs
```diff
@@ -0,0 +1,978 @@
...
+                Some(user.into())
+            }
+        }
+    };
+    let mut password: Option<Cow<str>> = url.password().map(|p| p.into()).map(maybe_decode);
+    let mut database: Option<Cow<str>> = match url.path() {
+        "" | "/" => None,
+        path => Some(path.trim_start_matches('/').into()),
+    }
+    .map(maybe_decode);
```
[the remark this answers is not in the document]
mmastrac: In this case it's a style thing where I prefer the terminal ends of multiple statements to use the same structure. Will skip the combinator work to land this sooner. :)

edb/server/pgconnparams.py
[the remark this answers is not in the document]
mmastrac: Due to a git failure, this last commit failed to push. I'll return to this code in the future as it will likely become pure Rust and we can revisit them.

PR #7660, opened 2024-08-19
Fix for Rust toolchain issues

mmastrac:


.github/workflows.src/tests.inc.yml
```diff
@@ -9,6 +9,9 @@
         fetch-depth: 50
         submodules: true
 
+    - name: Install Rust toolchain
+      uses: dsherret/rust-toolchain-file@v1
+
```
fantix, line 14: I think it'll also work if we move this to `tests.tpl.yml` at line 87:

```diff
diff --git a/.github/workflows.src/tests.tpl.yml b/.github/workflows.src/tests.tpl.yml
index 032de7c44..940f20e86 100644
--- a/.github/workflows.src/tests.tpl.yml
+++ b/.github/workflows.src/tests.tpl.yml
@@ -84,6 +84,9 @@ jobs:
     steps:
     <<- restore_cache() >>

+    - name: Install Rust toolchain
+      uses: dsherret/rust-toolchain-file@v1
+
     # Run the test

     - name: Test
```

fantix (approved):

.github/workflows.src/tests.inc.yml
mmastrac: Running this one now...

PR #7664, opened 2024-08-21
Add trpc guide

beerose:


docs/guides/tutorials/trpc.rst
```diff
@@ -0,0 +1,468 @@
...
+============================
+
+This guide explains how to integrate **EdgeDB** with **tRPC** for a modern, 
+type-safe API. We'll cover setting up database interactions, API routing, 
+and implementing authentication, all while ensuring type safety across the 
+client and server.
+
+You can reference the following repositories for more context:
+
+- `TurboRepo Template <https://github.com/scotttrinh/turbo-edb-app/blob/main-template/>`_: A monorepo template using **TurboRepo**, **EdgeDB**, and **tRPC**.
```
scotttrinh, line 13: Maybe we should move or rebuild this into our official examples repo?

docs/guides/tutorials/trpc.rst
```diff
@@ -0,0 +1,468 @@
...
+This will create an EdgeDB project and set up a schema to start with.
+
+Define the EdgeDB Schema
+------------------------
+
+The previous command generated a schema file in the `dbschema` directory.
+
+Here’s an example schema that defines a `User` model:
+
+.. code-block:: sdl
```
scotttrinh, line 40: Maybe add `:caption:` blocks to these files to make it a little more concrete what the structure is?

```suggestion
.. code-block:: sdl
   :caption: dbschema/default.esdl
```

docs/guides/tutorials/trpc.rst
```diff
@@ -0,0 +1,468 @@
...
+   $ # or npm install @trpc/react-query
+   $ # or bun add @trpc/react-query
+
+Define the tRPC Router
+-----------------------
+
+Here’s how to define a simple tRPC query that interacts with **EdgeDB**:
+
+.. code-block:: typescript
+   :caption: server/routers/_app.ts
```
scotttrinh, line 122: Ahh, yeah, exactly.

docs/guides/tutorials/trpc.rst
```diff
@@ -0,0 +1,468 @@
...
+   import { AppRouter } from './routers/_app';
+
+   export const api = createTRPCReact<AppRouter>();
+
+Client-Side Usage in Next.js
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can then use **tRPC** hooks to query the API from the client:
+
+.. code-block:: typescript
```
scotttrinh, line 187: Is this a client component?

docs/guides/tutorials/trpc.rst
```diff
@@ -0,0 +1,468 @@
...
+
+   The **tRPC** context provides the session and the EdgeDB client to the tRPC 
+   procedures:
+
+   .. code-block:: typescript
+
+      import { initTRPC } from '@trpc/server';
+      import { headers } from "next/headers";
+      import { auth } from "./path-to-auth";
+      import { edgedbClient } from "./path-to-edgedb";
```
scotttrinh, line 304: These are coming from the same module, I think, so if we make the module location and name explicit above, we can make this more explicit.

docs/guides/tutorials/trpc.rst
```diff
@@ -0,0 +1,468 @@
...
+
+      export const appRouter = t.router({
+        getUserData: t.procedure.query(async ({ ctx }) => {
+          if (!ctx.session) {
+            throw new Error("Not authenticated");
+          }
+          // Fetch data from EdgeDB using the client's session
+          const userData = await ctx.edgedbClient.query(`
+            SELECT User { name, email } WHERE .id = <uuid>$id
+          `, { id: ctx.session.user.id });
```
scotttrinh, line 349:
```suggestion
          if (!(await ctx.session.isSignedIn())) {
            throw new Error("Not authenticated");
          }
          // Fetch data from EdgeDB using the authenticated client
          const userData = await ctx.session.client.query(`
            select User { name, email }
          `);
```

This assumes you've set up some access policies to disallow selecting users other than yourself. Otherwise:

```suggestion
          if (!(await ctx.session.isSignedIn())) {
            throw new Error("Not authenticated");
          }
          // Fetch data from EdgeDB using the authenticated client
          const userData = await ctx.session.client.query(`
            select User { name, email } filter global ext::auth::ClientTokenIdentity in .identities
          `);
```

(or for single identity<->user models `global ext::auth::ClientTokenIdentity = .identity`)

docs/guides/tutorials/trpc.rst
```diff
@@ -0,0 +1,468 @@
...
+
+By integrating **EdgeDB Auth** into the tRPC context, you ensure that 
+authenticated sessions are securely passed to API procedures, enabling 
+user authentication and protecting routes.
+
+You can also reference these projects for further examples:
+
+- `TurboRepo Template <https://github.com/scotttrinh/turbo-edb-app/blob/main-template/>`_
+- `LookFeel Project <https://github.com/LewTrn/lookfeel>`_ - A real-world 
+  example using **EdgeDB** and **tRPC**.
```
scotttrinh, line 468: Do we want to restate these links here, or just keep them in the beginning or ending?

docs/guides/tutorials/trpc.rst
beerose: Or maybe change it to an official template so that people can click "use it as a template"?

scotttrinh: Yeah, maybe both? Can a subdirectory be a template? Probably not. I certainly don't mind linking to a personal repo of mine, but it feels like it's worth trying to make it a bit more official if we can. I guess since we already have the next+edb-auth template, there is already precedent for having a template repo in the edgedb org. I'll see about transferring it as-is and making it a template.

docs/guides/tutorials/trpc.rst
beerose: It could be both

docs/guides/tutorials/trpc.rst
beerose: I guess both options make sense: in the beginning for people who'd like to skip to the code, and at the end for a summary.

docs/guides/tutorials/trpc.rst
```diff
@@ -0,0 +1,475 @@
...
+----------
+
+By integrating **EdgeDB Auth** into the tRPC context, you ensure that 
+authenticated sessions are securely passed to API procedures, enabling 
+user authentication and protecting routes.
+
+You can also reference these projects for further examples:
+
+- `create-t3-turbo-edgedb <https://github.com/edgedb/create-t3-turbo-edgedb>`_ - 
+  A monorepo template using TurboRepo, EdgeDB, and tRPC.
```
beerose, line 473:
```suggestion
  A monorepo template using Turborepo, EdgeDB, and tRPC.
```

scotttrinh (approved):

PR #7672, opened 2024-08-22
Validation of object ids of prior DML over SQL

aljazerzen:
Ref #7294

## Problem

Such SQL queries:
```SQL
WITH inserted as (INSERT INTO "User" (name) VALUES ('aljaz') RETURNING id)
INSERT INTO "Document" (owner_id) SELECT id FROM inserted
```
... and uncompiled into this EdgeQL:
```
with
  dml_0 := (
    for x in `ins_val~0` union
      insert User { name := x.name }
  ),
  dml_1 := (
    for x in `ins_val~1` union
      insert Document { owner := (select User filter .id = x.owner.id) }
  )
select { dml_0, dml_1 }
```
... where `ins_val~0` and `ins_val~1` are `IRAnchor`s which are provided by the SQL query.

The problem is that with dml_1, `select User` will not contain the object inserted in dml_0.
It would error out, saying that `User` with this id does not exists.

This is a known limitation of EdgeQL (and has its reasons). But it cannot be justified in the original SQL, because there are no SELECT from base tables.

## Solution

This PR replaces `(select User filter .id = x.owner.id)` with `select {User, dml_0} filter .id = x.owner.id`. To be precise, it is replaced with:

```
if exists x.owner
  select {User, dml_0}
  filter .id = x.owner.id
  limit 1
else
  {}
```

---

Currently this is done for links in object insert/updates only.

TODO:
- [x] fix single row optimization,
- [x] inserting link's source,
- [x] inserting link's target,
- [x] ~deleting object's id~ [not needed as this cannot happen because of Postgres rules around DML](https://github.com/edgedb/edgedb/pull/7672#issuecomment-2411320341),
- [x] ~deleting link's source~ ditto,
- [x] ~updating object's id~ ditto,

---

This PR also contains additional minor changes:
- test result sorting,
- cleanup to remove `detached` in DML (it is apparently not needed anymore),
- fix for scope tree's find_visible_ex (that give me problems in prior attempts at this),

### DIFF
--- a/edb/pgsql/resolver/command.py
+++ b/edb/pgsql/resolver/command.py
@@ -474,11 +474,7 @@ def _construct_assign_element_for_ptr(
         target = ptr.get_target(ctx.schema)
         assert isinstance(target, s_objtypes.ObjectType)
 
-        # FIXME: We should change this to use the non-old variant, and
-        # then eliminate old. There's not enough test coverage of
-        # single links for me to want to make the change yet,
-        # though. -sully
-        ptr_ql = _construct_cast_from_uuid_to_obj_type_old(
+        ptr_ql = _construct_cast_from_uuid_to_obj_type(
             ptr_ql, target, stype_refs, optional=True, ctx=ctx
         )
 
@@ -489,14 +485,13 @@ def _construct_assign_element_for_ptr(
     )
 
 
-def _construct_cast_from_uuid_to_obj_type_old(
+def _construct_cast_from_uuid_to_obj_type(
     ptr_ql: qlast.Path,
     object: s_objtypes.ObjectType,
     stype_refs: Dict[uuid.UUID, List[qlast.Set]],
     *,
     optional: bool,
     ctx: Context,
-    no_id: bool = False,
 ) -> qlast.Expr:
     # Constructs AST that converts a UUID provided by ptr_ql to an object type.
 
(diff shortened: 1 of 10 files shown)


edb/pgsql/resolver/command.py
```diff
@@ -696,13 +797,7 @@ def _uncompile_insert_pointer_stmt(
 def _has_at_most_one_row(query: pgast.Query | None) -> bool:
     if not query:
         return True
-    return isinstance(query, pgast.SelectStmt) and (
-        (query.values and len(query.values) <= 1)
-        or (
-            isinstance(query.limit_count, pgast.NumericConstant)
-            and query.limit_count.val == '1'
-        )
-    )
```
aljazerzen, line 705: This change should be reverted: it is here because otherwise the pgcompiler will produce other bugs.

aljazerzen:
This PR is in a good shape and could be merged now, but it does disable an optimization that otherwise produces errors.

aljazerzen:
In the PR description I mention that this cast is currently implemented for "links in object inserts/updated". With this I mean following pattern:

```sql
WITH inserted as (INSERT ... RETURNING id)
INSERT/UPDATE ... VALUES/SET link_id = inserted.id
```

But we also cast ids to object types in:

- deleting object's id,

  ```sql
  DELETE FROM "User" WHERE id = '7a3a9c1a-146d-47af-9544-98ace90e6576'
  ```

- deleting link's source,

  ```sql
  DELETE FROM "User.shared_with" WHERE source = '7a3a9c1a-146d-47af-9544-98ace90e6576'
  ```

- updating object's id,

  ```sql
  UPDATE "User" SET ... WHERE id = '7a3a9c1a-146d-47af-9544-98ace90e6576'
  ```

- ~updating link's source~ (we don't support UPDATEs in link tables),

Initially, I believed that I'd have to implement intelligent casts to object types for each of these cases, but it turns out that this is not necessary. That's because when the UUID belongs to a newly inserted object, PostgreSQL's rules around DML visibility prevent the newly inserted rows from affecting the following DML.

For example:

```sql
WITH inserted AS (INSERT INTO "User" DEFAULT VALUES RETURNING id)
DELETE FROM "User" WHERE id = (SELECT id FROM inserted)
```
... should *not* DELETE any rows and should result in one new rows in `User`, because DELETE does not yet "see" the new row.

Similar applies for "deleting link's source" and "updating object's id".

msullivan (approved):

Issue #7621, opened 2024-08-06
Render and save SDL after successful migrations


1st1:
For the UI (and possibly the CLI) it would be nice to have a migrations browser that shows diffs as textual diffs of SDL, not just DDL deltas. The latter is quite low level and requires time for the reader to parse and understand.

I think the correct way to approach this would be to render SDL with the same machinery `DESCRIBE SCHEMA`. We do need to make sure the ordering of types/pointers/indexes/constraints/annotations/etc in the output is ordered, so that text diffing is possible.

The SDL should be saved in `schema::Migration` objects.

Separately, we should consider an "upgrade path", i.e. some process that would allow users of prior versions to populate the SDL migration state to their DB.

PR #7673, opened 2024-08-22
Add SDL to migrations.

dnwpark:
Add a config property `store_migration_sdl` with two options `cfg::StoreMigrationSDL::NeverStore` and `cfg::StoreMigrationSDL::AlwaysStore`.

When this is set to `cfg::StoreMigrationSDL::AlwaysStore`, during migrations or migration rewrites, the stored `schema::Migration` objects will have their resulting SDL saved.

related #7621

### DIFF
--- a/edb/server/compiler/ddl.py
+++ b/edb/server/compiler/ddl.py
@@ -165,6 +165,9 @@ def compile_and_apply_ddl_stmt(
             is_transactional=True,
         )
 
+    if isinstance(stmt, qlast.CreateMigration):
+        stmt.target_sdl = s_ddl.sdl_text_from_schema(new_schema)
+
     # If we are in a migration rewrite, we also don't actually
     # apply the DDL, just record it. (The DDL also needs to be a
     # CreateMigration.)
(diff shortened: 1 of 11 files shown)


dnwpark:
@msullivan In the end, figuring exactly how the migration was created is a bit too complex, so I've just left it as a always/never setting.

edb/server/compiler/ddl.py
```diff
@@ -165,6 +165,13 @@ def compile_and_apply_ddl_stmt(
             is_transactional=True,
         )
 
+    store_migration_sdl = compiler._get_config_val(ctx, 'store_migration_sdl')
+    if (
+        isinstance(stmt, qlast.CreateMigration)
+        and store_migration_sdl == 'AlwaysStore'
+    ):
+        stmt.target_sdl = s_ddl.sdl_text_from_schema(new_schema)
```
msullivan, line 173: Do we need to do this? Wouldn't the logic in `apply` take care of it?

msullivan (approved):

PR #7674, opened 2024-08-22
CI: fix Rust toolchain on macOS

fantix:
Thanks @mmastrac for the fix. Looks like the toolchain.toml checkout should go before the edgedb-pkg checkout, or the latter will get overwritten.

[Sample run](https://github.com/edgedb/edgedb/actions/runs/10510224013)


mmastrac (approved):
LGTM!

PR #7676, opened 2024-08-22
Add `net` and `net::http` modules

scotttrinh:
The `net` module contains types and functions that will send network requests. This change includes a no-op placeholder for the `net::http::schedule_request` function.

### DIFF
--- a/edb/lib/net.edgeql
+++ b/edb/lib/net.edgeql
@@ -45,7 +45,7 @@ CREATE SCALAR TYPE net::http::Method EXTENDING std::enum<
 CREATE TYPE net::http::ScheduledRequest extending std::BaseObject {
     CREATE REQUIRED PROPERTY state: net::RequestState;
     CREATE REQUIRED PROPERTY created_at: std::datetime;
-    CREATE REQUIRED PROPERTY failure: tuple<kind: net::RequestFailureKind, message: str>;
+    CREATE PROPERTY failure: tuple<kind: net::RequestFailureKind, message: str>;
 
     CREATE REQUIRED PROPERTY url: str;
     CREATE REQUIRED PROPERTY method: net::http::Method;
--- a/edb/lib/net.edgeql
+++ b/edb/lib/net.edgeql
@@ -17,3 +17,63 @@
 #
 
 CREATE MODULE net;
+
+CREATE SCALAR TYPE net::RequestState EXTENDING std::enum<
+    Pending,
+    InProgress,
+    Completed,
+    Failed
+>;
+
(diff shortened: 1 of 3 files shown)


edb/lib/net.edgeql
```diff
@@ -0,0 +1,79 @@
...
+    `DELETE`,
+    HEAD,
+    OPTIONS,
+    PATCH
+>;
+
+CREATE TYPE net::http::ScheduledRequest extending std::BaseObject {
+    CREATE REQUIRED PROPERTY state: net::RequestState;
+    CREATE REQUIRED PROPERTY created_at: std::datetime;
+    CREATE REQUIRED PROPERTY failure: tuple<kind: net::RequestFailureKind, message: str>;
```
fantix, line 48: `failure` should be optional?

edb/lib/net.edgeql
```diff
@@ -0,0 +1,79 @@
...
+CREATE SCALAR TYPE net::RequestState EXTENDING std::enum<
+    Pending,
+    InProgress,
+    Completed,
+    Failed
+>;
+
+CREATE SCALAR TYPE net::RequestFailureKind EXTENDING std::enum<
+    NetworkError,
+    Timeout
```
fantix, line 30: My concern here is that I - as a user - don't know if I should re-schedule my request upon these errors, like `NetworkError` could be `cannot resolve hostname` (e.g. should backoff for some time and retry) or `connection reset receiving response` (e.g. should check remote first to avoid double transaction or something), `Timeout` may also happen before or after sending a request.

Regex-matching the `message` is suboptimal - e.g. `LANG=zh_CN.UTF-8` may localize the message in some cases of propagated low-level errors. I know we decided to let user decide on retrying, but I currently don't have a better suggestion to expose more info about the error (so that the user could actually decide) while keeping the API simple. So I guess it's fine for now.

fantix (approved):

edb/lib/net.edgeql
scotttrinh: Good catch, thanks.

edb/lib/net.edgeql
scotttrinh: Oh, we should absolutely have a lot more enums here, but I'm not even sure what all of the different error possibilities are yet. Mostly I imagine these are the rough groups of kinds of errors we could get from the underlying library: either they've timed out (which we would control I think?) or the request library (httpx in python or reqwest in rust) would throw an exception and we'd bubble that back up. Makes sense to match the granularity of the underlying library to make it easiest for the consumer.

PR #7682, opened 2024-08-23
Stop using catalog version for the default tenant id

msullivan:
I don't think it ever is really helpful, except maybe in some dev
situations where we don't actually use it.

Anybody relying on tenant ids almost certainly needs to be specifying
them explicitly to get anything to work.

Just use `'E`' as the tenant id. This will give us nice predictible
database names and eases inplace upgrades.

### DIFF
--- a/edb/buildmeta.py
+++ b/edb/buildmeta.py
@@ -60,7 +60,7 @@ from edb.common import verutils
 # The merge conflict there is a nice reminder that you probably need
 # to write a patch in edb/pgsql/patches.py, and then you should preserve
 # the old value.
-EDGEDB_CATALOG_VERSION = 2024_08_23_00_00
+EDGEDB_CATALOG_VERSION = 2024_08_23_00_01
 EDGEDB_MAJOR_VERSION = 6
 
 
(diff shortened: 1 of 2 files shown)


elprans:
> I don't think it ever is really helpful, except maybe in some dev
> situations where we don't actually use it.

Wouldn't this essentially render `--ignore-other-tenants` broken?

msullivan:
> > I don't think it ever is really helpful, except maybe in some dev
> > situations where we don't actually use it.
> 
> Wouldn't this essentially render `--ignore-other-tenants` broken?

It breaks the use case of running different versions of `edgedb` against the same backend without specifying tenant names. That's a pretty marginal use case, and if anybody cares about it they can explicitly specify a tenant name, right?

msullivan:
Plus, the flag is mostly undocumented

elprans:
> Plus, the flag is mostly undocumented

Fair enough, but please do remember to document this in the upgrade considerations of the 6.0 changelog.

elprans (approved):

PR #7683, opened 2024-08-23
Run gen-meta-grammars

scotttrinh:
Update pygments meta with latest changes


scotttrinh:
Is there anything else that needs to be done with whatever uses this?

msullivan (approved):

PR #7686, opened 2024-08-26
Fix TagUnpackRow encoding over SQL adapter

aljazerzen:
A sister PR to #7684

When there is DML without RETURNING clause, we inject a COUNT(*) to the
top-level compiled query. We then intercept the single row produced and
convert it tag of to `CommandComplete` message.

This PR makes sure that `Bind` of such queries requests the result to be
in binary encoding and not the encoding that that client might have
requested. What they have requested does not matter, since we are not
returning any results.

This does also simplify our code for parsing the data row into the tag.

### DIFF
--- a/edb/server/pgcon/pgcon.pyx
+++ b/edb/server/pgcon/pgcon.pyx
@@ -1741,7 +1741,38 @@ cdef class PGConnection:
                             action.frontend_only = True
                             prepared.add(be_stmt_name)
 
-                if not action.is_frontend_only():
+                if action.is_frontend_only():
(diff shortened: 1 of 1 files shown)


aljazerzen:
@msullivan Fantix was able to help here: the result encoding can be specified in [`Bind` command](https://www.postgresql.org/docs/current/protocol-message-formats.html#PROTOCOL-MESSAGE-FORMATS-BIND).

fantix (approved):

msullivan (approved):

PR #7689, opened 2024-08-28
Initial implementation of in-place upgrades

msullivan:
This builds on all of the work done to put the standard library into versioned namespaces.

The components currently are:
 * A new `ADMINISTER prepare_upgrade()` command that returns the metadata about the schema necessary to perform the upgrade. Currently a user script runs this on every database and aggregates it, and we pass that to the server.
 * An `--inplace-upgrade <json file>` flag is added to the server, which tells it to do an inplace upgrade using the `prepare_upgrade` information in the file. It does this by running the a partial standard library tpldbdump on each database, and then rebuilding the schema tables for each database.
 * Trampoline swapping, which requires some subtlety: we can't drop or reorder columns on views, so I wrote some pl/pgsql that regenerates a view preserving the order of previously existing fields and inserting NULL for removed ones.
 * Testing: I've added an initial testing script that runs `inittestdb`, collects the `prepare_upgrade` results, applies a patch that adds and removes some stuff to the std schema and bumps the version number, upgrades the database in place, and then runs the full test suite.

Major progress on #6697.

Remaining TODOs (can largely be in follow ups):
- [ ] Extension upgrade support
- [ ] Parallelize upgrade (by using compiler pool)?
- [ ] Support upgrade while old version is running
- [ ] Better fault tolerance
- [ ] Extension triggers and user functions: regenerate them if a user type points to schema object that has grown a new subtype


misc/inplace-testing/make-and-prep.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash -ex
```
elprans, line 1: Any reason why `misc/` and not `tests/` or `edb/testbase/`?

elprans (approved):

misc/inplace-testing/make-and-prep.sh
msullivan: Because it isn't part of the real test suite, I guess. I can move it though.

PR #7694, opened 2024-08-28
Disable a broken assert in the rust parser

msullivan:
@aljazerzen, could you follow up on this?


mmastrac (approved):
LGTM.

aljazerzen:
I've noticed that a few weeks back but coult not figure out what was going wrong.

Its probably an artifact of some weird combination of recovery steps.

The worst part is that the panic produced is converted to an exception object that cannot be picked, so test suite has additional problems with that.

PR #7695, opened 2024-08-28
Add sorting for constraints and indexes when generating code.

dnwpark:
Sorts constraints and indexes by the text of their expressions.

For example, given a schema:
```
type Foo {
    constraint exclusive on ('C');
    constraint exclusive on ('A');
    constraint exclusive on ('D);
    constraint exclusive on ('B');
};
```

The query `describe schema as sdl` will list the constraints sorted alphabetically:
```
module default {
    type Foo {
        constraint std::exclusive on ('A');
        constraint std::exclusive on ('B');
        constraint std::exclusive on ('C');
        constraint std::exclusive on ('D');
    };
};
```

related #7621

### DIFF
--- a/edb/edgeql/codegen.py
+++ b/edb/edgeql/codegen.py
@@ -933,14 +933,43 @@ class EdgeQLSourceGenerator(codegen.SourceGenerator):
                         self.new_lines = 1
                     self.visit_list(list(items), terminator=';')
             elif self.descmode or self.sdlmode:
-                sort_key = lambda c: (
-                    (c.name.itemclass or '', c.name.name)
-                    if isinstance(c, qlast.ObjectDDL)
-                    else ('', c.name if isinstance(c, qlast.SetField) else '')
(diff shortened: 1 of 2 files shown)


msullivan (approved):

edb/edgeql/codegen.py
```diff
@@ -933,14 +933,43 @@ def _ddl_visit_body(
...
                     self.visit_list(list(items), terminator=';')
             elif self.descmode or self.sdlmode:
-                sort_key = lambda c: (
-                    (c.name.itemclass or '', c.name.name)
-                    if isinstance(c, qlast.ObjectDDL)
-                    else ('', c.name if isinstance(c, qlast.SetField) else '')
+                sort_desc_or_sdl: Callable[
+                    [qlast.Base],
+                    tuple[str, ...]
+                ] = lambda c: (
```
msullivan, line 939: Add a comment here about what we're doing and why

PR #7697, opened 2024-08-29
Load `net` module and fix issues

scotttrinh:
In #7676, we were not actually loading the `net.edgeql` module. Once we began loading it, we discovered many issues with syntax and DDL ordering which are fixed here.

### DIFF
--- a/edb/lib/net.edgeql
+++ b/edb/lib/net.edgeql
@@ -42,36 +42,39 @@ CREATE SCALAR TYPE net::http::Method EXTENDING std::enum<
     PATCH
 >;
 
+CREATE TYPE net::http::Response EXTENDING std::BaseObject {
+    CREATE REQUIRED PROPERTY created_at: std::datetime;
+    CREATE PROPERTY status: std::int16;
+    CREATE PROPERTY headers: std::array<std::tuple<name: std::str, value: std::str>>;
+    CREATE PROPERTY body: std::bytes;
+};
+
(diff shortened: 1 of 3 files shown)


fantix (approved):
aha, right

aljazerzen:
@scotttrinh I've added a fix, it was an easy one. The two new modules & their types were showing up in SQL introspection (`SELECT * FROM information_schema.tables`), so `pg_dump` tried to create those tables and schemas, but failed for a reason unknown to me.

scotttrinh:
@aljazerzen 

Don't we want these tables to be present in a dump?

aljazerzen:
No, we don't. Accessing EdgeDB via SQL is not meant to expose "internal" or EdgeDB-specific mechanisms. That's because when using `pg_dump` (not `edgedb dump`), the goal is to export data out of EgdeDB into a plain postgres instance. In such cases, user probably does not want to have tables `"net::http"."ScheduledRequest"` and `"cfg"."Auth"` created.

aljazerzen (approved):

scotttrinh:
@aljazerzen 

> That's because when using pg_dump (not edgedb dump), the goal is to export data out of EgdeDB into a plain postgres instance.

That's helpful context, thanks! I think once we have write SQL, we might want to support writing to this table and reading from it from Postgres, but maybe that's a separate concern than dumping. Thanks for the explanation!

PR #7698, opened 2024-08-30
Fix few typos & add SNI in connection docs

diksipav:
I started updating connection docs with SNI, but realized it lacks info on some other connection options too.

I posted questions in Slack.


scotttrinh (approved):
I think it's fine to land this without the other documentation, we can add that in a separate PR once we get all of the answers there.

1st1 (approved):

PR #7705, opened 2024-09-03
CI: don't use hidden dir for shared artifacts

fantix:
[actions/upload-artifact@v4.4.0](https://github.com/actions/upload-artifact/releases/tag/v4.4.0) dropped support for hidden paths (e.g., with a leading dot `.results`), which breaks our CI when using such paths - fix so that we don't.

Also, pinned such action versions so that breakage won't happen accidentally.


msullivan (approved):
Put an explanation in the commit message?

PR #7710, opened 2024-09-04
Fix SET within transactions over SQL adapter

aljazerzen:
I've found the following behavior:

```
edgedb=# begin;
         show search_path;
         set search_path = 'my_schema';
         show search_path;
         commit;
         show search_path;
 search_path
-------------
 public

 search_path
-------------
 public         -- this is wrong: it should be my_schema

 search_path
-------------
 my_schema
```

When using SET (or SET SESSION) within a transaction, this SET is applied only after the transaction is committed. It should be applied immediately.

I've debugged it down to this:
- in edb/server/compiler/compiler.py, we process SQL commands in `compile_sql(...)`,
- we have a special branch for SET command, which applies [correct `set_vars` to `SQLQueryUnit`](https://github.com/edgedb/edgedb/blob/d85ced24431cda0d08f3d612b31b37b8b2a7c21b/edb/server/compiler/compiler.py#L663-L664),
- at the end of command processing, we correctly apply the [unit to the `SQLTransactionState`](https://github.com/edgedb/edgedb/blob/d85ced24431cda0d08f3d612b31b37b8b2a7c21b/edb/server/compiler/compiler.py#L808),
- the call to the compiler then finishes and discards `SQLTransactionState`,
- subsequent calls to the compiler receive the `SQLTransactionState` from before the transaction, not the one that preceding units have been applied to.

TLDR; within transactions, the modified `SQLTransactionState` is not being passed to the compiler correctly.


aljazerzen:
(there are many tests failing because we change the search_path variable and don't reset it back)

At first, look only at `test_sql_query_set_01`.

fantix:
> the call to the compiler then finishes and discards SQLTransactionState,

This was actually kinda intentional - the frontend-only "state" is replayed in the I/O server by applying each `SQLQueryUnit` in the dbview, and passed unidirectionally to the compiler process. (the `fe_settings` on `SQLQueryUnit` is only used to check if a recompilation is needed)

edb/server/protocol/pg_ext.pyx
```diff
@@ -70,27 +70,34 @@ def managed_error():
...
         self._in_tx_explicit = False
         self._in_tx_implicit = False
+
+        # Kepp track of backend settings so that we can sync to use different
+        # backend connections (pgcon) within the same frontend connection,
+        # see serialize_state() below and its usages in pgcon.pyx.
+        self._settings = DEFAULT_SETTINGS
         self._in_tx_settings = None
+
+        # Frontend-only settings are defined by the high-level compiler, and
```
aljazerzen, line 82: High-level compiler meaning the `edb/compiler/compiler.py`, `compile_sql` function?

fantix (approved):
I think this is the correct fix; but please also double check if that all makes sense.

aljazerzen (commented):
I don't really understand the changes here, but if it makes all of the tests pass, including the new ones, it must be correct.

edb/server/protocol/pg_ext.pyx
fantix: ya the one you just moved to `edb/server/compiler/sql.py` ;)

PR #7711, opened 2024-09-04
Fix iterators in DML over SQL adapter

aljazerzen:
Test case:

```sql
INSERT INTO "Document" (title) VALUES ('Report'), ('Briefing')
```

Here, the input relation is injected with an iterator (as it is needed for DML), but is is also registered as identity & value.
This causes problems in specific test case, where the iterator will be joined with base
table to compare ids - this produces no results so no objects are inserted.

If I remove the identity/value, the ql compiler will complain because it is
first doing a plain `compile_Set` on the iterator.


aljazerzen:
Fixed by #7672

Issue #7692, opened 2024-08-28
Proper function DML support


msullivan:
In order to support the Networking Module RFC (https://github.com/edgedb/rfcs/blob/master/text/1026-net-module.rst), I think we need to support functions that contain DML.

Currently there is a flag that allows it, but it doesn't work very well. The DML is actually contained inside the postgres function, and it doesn't work very well. In particular, it can be potentially optimized away and also can't properly return the created object.

I think the correct approach here is to support them by inlining functions that contain DML.

I'm filing this now as a tracking issue but there are some serious semantic questions to answer.

msullivan:
The semantic question:

Imagine a function like:
```
function make_user(name: str) -> User using (insert User { name := name })
```

The behavior of `select make_user('Bob')` is clear, but how about `select make_user(<str>{})` or `select make_user({'foo', 'bar'})`.

I think we really do *not* want to have the number of inserts done be implicitly determined by the cardinality of the argument (we firmly disallow that sort of thing with DML currently).

I think the alternative is rejecting those queries with a cardinality error, which requires making DML containing functions have a very different argument matching behavior than regular ones.

Sigh. I think the second choice is the right one but it means more design to be done.

elprans:
> I think we really do not want to have the number of inserts done be implicitly determined by the cardinality of the argument (we firmly disallow that sort of thing with DML currently).

How is this different from DML in a `for` over a set with variable cardinality?

msullivan:
Because the FOR makes it explicit.

msullivan:
We should write some docs but this is done, right?

dnwpark:
@msullivan Yeah, I would say so. Just tidying some stuff up with the volatility before starting on the docs.

PR #7713, opened 2024-09-04
Implement function inlining.

dnwpark:
Allows for the ir ast trees to be directly embedded into the tree of a calling expression.

This is accomplished by:
- Retrieving the function body and recompiling ir in the ir compiler
    - Use a new context detached from the inlining context
    - Added a function `compile_function_inline` to do only the necessary steps, excluding things materializing sets.
- Substituting any `irast.Parameter` nodes from the function with a new `irast.InlinedParameterExpr`
    - This is necessary because `irast.Parameter` is used for runtime parameters when compiling queries.
- Compiling the arguments in a subrelation in the pg compiler
    - Include the rvar and path value in the appropriate relations.

related #7692

Note: The volatility check for inlined function bodies was removed. This is because functions which use globals are `immutable`, since they expect a globals json to be passed in. This is not true of their inlined bodies, which become `stable`.

### DIFF
--- a/edb/edgeql/compiler/__init__.py
+++ b/edb/edgeql/compiler/__init__.py
@@ -164,7 +164,6 @@ if TYPE_CHECKING:
     from edb.ir import ast as irast
     from edb.ir import staeval as ireval
 
-    from . import context
     from . import dispatch as dispatch_mod
     from . import inference as inference_mod
     from . import normalization as norm_mod
@@ -274,10 +273,7 @@ def compile_ast_to_ir(
         debug.header('EdgeQL AST')
         debug.dump(tree, schema=schema)
 
-    ctx = stmtctx_mod.init_context(
-        schema=schema,
-        options=options,
-    )
+    ctx = stmtctx_mod.init_context(schema=schema, options=options)
 
     if isinstance(tree, qlast.Expr) and ctx.implicit_limit:
         tree = qlast.SelectQuery(result=tree, implicit=True)
@@ -321,7 +317,6 @@ def compile_ast_fragment_to_ir(
     schema: s_schema.Schema,
     *,
     options: Optional[CompilerOptions] = None,
-    inlining_context: Optional[context.ContextLevel] = None,
 ) -> irast.Statement:
     """Compile given EdgeQL AST fragment into EdgeDB IR.
 
@@ -347,9 +342,7 @@ def compile_ast_fragment_to_ir(
     if options is None:
         options = CompilerOptions()
 
-    ctx = stmtctx_mod.init_context(
-        schema=schema, options=options, inlining_context=inlining_context
-    )
+    ctx = stmtctx_mod.init_context(schema=schema, options=options)
     ir_set = dispatch_mod.compile(tree, ctx=ctx)
 
     result_type = ctx.env.set_types[ir_set]
--- a/edb/edgeql/compiler/__init__.py
+++ b/edb/edgeql/compiler/__init__.py
@@ -326,6 +326,7 @@ def compile_ast_fragment_to_ir(
     schema: s_schema.Schema,
     *,
     options: Optional[CompilerOptions] = None,
+    inlining_context: Optional[context.ContextLevel] = None,
 ) -> irast.Statement:
     """Compile given EdgeQL AST fragment into EdgeDB IR.
 
@@ -351,7 +352,9 @@ def compile_ast_fragment_to_ir(
     if options is None:
         options = CompilerOptions()
 
-    ctx = stmtctx_mod.init_context(schema=schema, options=options)
+    ctx = stmtctx_mod.init_context(
+        schema=schema, options=options, inlining_context=inlining_context
+    )
     ir_set = dispatch_mod.compile(tree, ctx=ctx)
 
     result_type = ctx.env.set_types[ir_set]
--- a/edb/edgeql/compiler/__init__.py
+++ b/edb/edgeql/compiler/__init__.py
@@ -202,7 +202,6 @@ def compile_ast_to_ir(
     *,
     script_info: Optional[irast.ScriptInfo] = None,
     options: Optional[CompilerOptions] = None,
-    inlining_context: Optional[context.ContextLevel] = None,
 ) -> irast.Statement:
     pass
 
@@ -214,7 +213,6 @@ def compile_ast_to_ir(
     *,
     script_info: Optional[irast.ScriptInfo] = None,
     options: Optional[CompilerOptions] = None,
-    inlining_context: Optional[context.ContextLevel] = None,
 ) -> irast.ConfigCommand:
     pass
 
@@ -226,7 +224,6 @@ def compile_ast_to_ir(
     *,
     script_info: Optional[irast.ScriptInfo] = None,
     options: Optional[CompilerOptions] = None,
-    inlining_context: Optional[context.ContextLevel] = None,
 ) -> irast.Statement | irast.ConfigCommand:
     pass
 
@@ -238,7 +235,6 @@ def compile_ast_to_ir(
     *,
     script_info: Optional[irast.ScriptInfo] = None,
     options: Optional[CompilerOptions] = None,
-    inlining_context: Optional[context.ContextLevel] = None,
 ) -> irast.Statement | irast.ConfigCommand:
     """Compile given EdgeQL AST into EdgeDB IR.
 
@@ -281,7 +277,6 @@ def compile_ast_to_ir(
     ctx = stmtctx_mod.init_context(
         schema=schema,
         options=options,
-        inlining_context=inlining_context,
     )
 
     if isinstance(tree, qlast.Expr) and ctx.implicit_limit:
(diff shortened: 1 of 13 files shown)


msullivan:
I think it would be useful to try a 2-argument function as part of your early tests

edb/edgeql/compiler/stmtctx.py
```diff
@@ -88,9 +89,33 @@ def init_context(
...
+        aliases=(
+            inlining_context.aliases
+            if inlining_context is not None else
+            None
+        ),
+        scope_id_ctr=(
+            inlining_context.scope_id_ctr
+            if inlining_context is not None else
+            None
+        ),
```
msullivan, line 117: We should use the existing context, with appropriate new scopes applied to it.
Doing it in a completely separate context is not going to work at all in anything but the most simple situations.

msullivan: And especially the same environment. (Or at least, an almost identical environment, since we might need to change options or something, though that is a little fraught too.)

edb/ir/ast.py
```diff
@@ -886,6 +886,20 @@ def is_global(self) -> bool:
...
+    # Parameter is used for both function parameters and runtime parameters.
+    # If a function is inlined in compile_FunctionCall, this is used to
+    # specify the function's parameters.
+
+    # During function compilation, arguments are converted from names to either
+    # int for positional argument or str for named argument.
+    arg_key: int | str
+
+    required: bool
+    is_global: bool
```
msullivan, line 900: The more typical way to do something like this in our compiler would probably be to substitute in `irast.Set` objects that have the same `path_id` as as the arguments but have an `irast.RefExpr` as their `expr`, to indicate that they are bound somewhere.

They should then get picked up by their path id when doing the backend compilation, without needing to add an extra map.

edb/edgeql/compiler/context.py
```diff
@@ -285,6 +285,14 @@ def __init__(
...
         alias_result_view_name: Optional[s_name.QualName] = None,
         options: Optional[GlobalCompilerOptions] = None,
+        set_types: Optional[Dict[irast.Set, s_types.Type]] = None,
+        scope_tree_nodes: Optional[
+            MutableMapping[int, irast.ScopeTreeNode]
+        ] = None,
+        type_rewrites: Optional[dict[
+            Tuple[s_types.Type, bool],
+            irast.Set | None | Literal[True]
+        ]] = None,
```
msullivan, line 295: This is missing a lot of fields that need to be preserved in Environment for correct operation in some case or another, especially once DML is supported.
Definitely needed:
- materialized_sets
- dml_exprs
- dml_stmts
- dml_rewrites
- schema_refs

I think possibly needed, or needed if things are done a little differently:
- pointer_derivation_map (for card inference?)
- pointer_specified_info (for card inference?)
- query_globals

Plus, I think having all of the caches shared will help get more cache hits, even if it isn't *necessary*.
And I think many of the other things wouldn't hurt to have be shared.

So I think instead of constructing a fully new env, we should probably make a straight copy of the env and just update the few fields we care about changing.

It might /just/ be `options` we need to change, though? In which case it might actually just be best to update the `options` field and then restore it after.

edb/schema/functions.py
```diff
@@ -2440,3 +2443,103 @@ def compile_function(
...
+
+    # Infer cardinality, multiplicity, and volatility
+    inf_ctx = inference.make_ctx(env=ctx.env)
+    inference.infer_cardinality(
+        ir_set, scope_tree=ctx.path_scope, ctx=inf_ctx
+    )
+    inference.infer_multiplicity(
+        ir_set, scope_tree=ctx.path_scope, ctx=inf_ctx
+    )
+    inference.infer_volatility(ir_set, env=ctx.env)
```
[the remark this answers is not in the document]
msullivan: Oh, I think the reason why it doesn't work without inferring them here is because not enough of ctx/env are being shared

edb/schema/functions.py
```diff
@@ -2440,3 +2443,103 @@ def compile_function(
...
+        options=get_compiler_options(
+            schema,
+            context,
+            func_name=func_name,
+            params=params,
+            track_schema_ref_exprs=track_schema_ref_exprs,
+            inlining_context=inlining_context,
+        ),
+        inlining_context=inlining_context,
+    )
```
msullivan, line 2481: Do we really need to go through `init_context` here? I think we probably should be trying to share the context mostly. (I think it probably would work if we create a `detached()` context?)

msullivan (commented):
Overall this looks pretty good.

I think we need to share more of env and ctx when compiling the function body, though.

I think there are some things that are being done mostly to workaround not sharing enough state, and a bunch more will come up when doing DML or when writing other sorts of tests.

edb/edgeql/compiler/func.py
```diff
@@ -356,14 +460,54 @@ def compile_FunctionCall(
...
+                    irast.InlinedParameterExpr(
+                        typeref=arg.expr.typeref,
+                        required=node.expr.required,
+                        is_global=node.expr.is_global,
+                    ),
+                    path_id=arg.expr.path_id,
+                    ctx=self.ctx,
+                )
+            else:
+                return arg
```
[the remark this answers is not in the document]
dnwpark: Default values get added to `inlined_args` as a `irast.Set` instead of `CallArg`. In that case, the default value is inlined directly. This should work because defaults are always constants.

edb/edgeql/compiler/func.py
```diff
@@ -356,14 +460,54 @@ def compile_FunctionCall(
...
                 ctx=ctx,
             )
 
     ir_set = setgen.ensure_set(res, typehint=rtype, path_id=path_id, ctx=ctx)
     return stmt.maybe_add_view(ir_set, ctx=ctx)
 
 
+class ArgumentInliner(ast.NodeTransformer):
+
+    mapped_args: dict[irast.PathId, irast.PathId]
```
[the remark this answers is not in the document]
dnwpark: I decided to not go down that path since the backend compiler does stuff with args too. Probably possible though.

edb/schema/functions.py
dnwpark: I tried to see if I could get a detached context working but I think the inlined context needs to have more than what is available in a detached context.

edb/edgeql/compiler/func.py
```diff
@@ -356,14 +460,54 @@ def compile_FunctionCall(
         for arg in res.args.values():
             pathctx.register_set_in_scope(
                 arg.expr,
-                optional=arg.param_typemod == ft.TypeModifier.OptionalType,
+                optional=(
+                    arg.param_typemod == ft.TypeModifier.OptionalType
+                    or arg.is_default
```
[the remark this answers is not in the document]
dnwpark: It was a consequence of how `has_inlined_defaults` interacted with the rest of the inlining. The simplest solution seems to be to just disable inlining defaults when inlining the function body.

edb/pgsql/compiler/relgen.py
```diff
@@ -1013,6 +1023,18 @@ def process_set_as_path_type_intersection(
...
+        ) is None:
+            assert isinstance(source_rvar.query, pgast.Query)
+            # When inlining functions, the prefix path will be the argument
+            # path id, while ir_source.path_id will be the "correct" path id.
+            # Remap the argument path id here to ensure the path bond is
+            # able to find the source rvar when joining.
+            prefix_path_id = pathctx.map_path_id(
+                prefix_path_id, source_rvar.query.view_path_id_map
+            )
+
```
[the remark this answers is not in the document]
dnwpark: The most straightforward solution here was to just update the paths of all pointers in the edgeql->ir side instead of trying to rely on the path mapping.

tests/test_edgeql_functions.py
```diff
@@ -8308,3 +8308,2149 @@ async def test_edgeql_functions_complex_types_04(self):
...
+                create required property a -> int64;
+            };
+            insert Bar{a := 1};
+            insert Bar{a := 2};
+            insert Bar{a := 3};
+            create function foo(x: Bar) -> int64 {
+                set is_inlined := true;
+                using (x.a);
+            };
+        ''')
```
msullivan, line 9554: Could we do a test like
```
            create function foo() -> int64 {
                set is_inlined := true;
                using (count(Bar));
            };
```
and then query
```
select (Bar, foo())
```
and make sure that we get `3` back as the result from `foo()`.
I want to make sure that paths in the inlined functions don't end up getting captured by factoring with the enclosing query.

edb/edgeql/compiler/__init__.py
```diff
@@ -201,6 +202,7 @@ def compile_ast_to_ir(
     *,
     script_info: Optional[irast.ScriptInfo] = None,
     options: Optional[CompilerOptions] = None,
+    inlining_context: Optional[context.ContextLevel] = None,
```
msullivan, line 205: You don't use these entrypoints anymore, right? Can we drop the `inlining_context` from them?

msullivan (commented):
This looks pretty good, though I'm worried about the path factoring issue I asked for a test for, which we'll need to talk about

msullivan (approved):

msullivan:
Great work!

PR #7734, opened 2024-09-10
Add smoke tests for `net` and `net::http` types

scotttrinh:
These tests uncovered the `_is_forbidden_stdlib_type_for_mod` bit, and I wasn't 100% sure if this was the right fix here, but tests pass.


fantix (approved):
I think it's good

PR #7735, opened 2024-09-10
Store module name instead of dynamic look-up

scotttrinh:
The value should be stable across all of these invocations


fantix (approved):
Yeah, that's correct.

PR #7736, opened 2024-09-10
Move `net` into `std` namespace

scotttrinh:


tests/test_edgeql_net_schema.py
```diff
@@ -66,7 +66,7 @@ async def test_edgeql_net_http_request_response_01(self):
         await self.assert_query_result(
             '''
             with
-                nh as module net::http,
+                nh as module std::net::http,
```
scotttrinh, line 69: I did have to fully qualify this.

fantix (approved):
I saw the discussion :+1:

tests/test_edgeql_net_schema.py
1st1: Yeah, we should indeed fix WITH to accept unqualified names.

Issue #7733, opened 2024-09-09
Extension packaging


msullivan:
Key design takeaways:
- CLI will download and install packages into portable builds, *including* installing postgres extensions into the postgres tree
- Distro package extensions will need to be packaged as distro packages?
   - Or in a pinch installed by the CLI running as root. (Probably more acceptable in docker than in general.)
- Docker?
   - Have an environment variable that triggers package installation using CLI?
   - Have users trigger it with `docker exec`.
   - Bottom line: nothing too special to help docker

----

The key rejected ideas are:
 * Have the server be in charge of installing into the postgres directory: it's a bit dodgy for the program to be modifying its own installation like that--often it won't be writable anyway (and postgres's extension config path isn't configurable, so we can't easily point it elsewhere)
 * Have the server be able to *download* packages: all of the above problems, plus production instances often won't have network access anyway. (The plus side of this is that it might make installing into docker much easier)

msullivan:
Remaining question: should we do "fat binaries" with compiled postgres extensions for all platforms, or build separate packages for each target?

msullivan:
Oh, ugh -- we are *also* going to need to be able to package all of the library dependencies for our portable builds. I don't think that affects the package loading too much though, just the build step.

PR #7738, opened 2024-09-10
Support installing packaged extensions

msullivan:
An packaged extension is a .zip file with a MANIFEST.toml file.
Installing an extension consists of:
 1. Dropping the .zip file into the extensions/ search path of edgedb
 2. If present, installing the postgres extension files into
    the packaged postgres directory.

The installation is currently done by an `install-ext.py` python
script, which should get turned into part of the CLI.

On server startup, the server checks the search path for any .zip
files, checks the MANIFEST for the names and versions, and if the
package does not exist, runs the script to install it.

Work on https://github.com/edgedb/edgedb/issues/7733.

The major rationale of having a separate MANIFEST with name/version is
so that we don't need to parse the whole extension package in order to
determine the name and version of extension, but maybe that isn't
necessary. We could just parse to do it, or we could try getting those
from the filename?


msullivan:
Another question: postgres extensions put files in both `lib` and `share`. Would it be worth invoking `pg_config` to get the `LIBDIR` and `SHAREDIR` separately? Currently we assume they are siblings.

elprans:
> Another question: postgres extensions put files in both `lib` and `share`. Would it be worth invoking `pg_config` to get the `LIBDIR` and `SHAREDIR` separately? Currently we assume they are siblings.

Invoking separately is probably best.

msullivan:
Maybe we should fully drop the MANIFEST, get the name and version from the file name, have the postgres extension always live in `pg/{lib,share}`, and run the `*.edgeql` files in alphabetical order?

msullivan:
On a suggestion from @aljazerzen that I really liked, I turned `install-ext.py` from an "external script that we will rewrite in rust" into an `edgedb-load-ext` command that the CLI will invoke. This is nice because it reduces the amount of logic overall and narrows the interface between CLI and edgedb.

edb/load_ext/main.py
```diff
@@ -0,0 +1,156 @@
...
+    'include': 'includedir',
+}
+
+
+def install_pg_extension(
+    pkg: pathlib.Path,
+    pg_config: dict[str, str],
+) -> None:
+    with zipfile.ZipFile(pkg) as z:
+        with z.open('MANIFEST.toml') as m:
```
1st1, line 53: I'd consider renaming to `GEL_EXTENSION.toml`. cc @elprans

edb/tools/config.py
```diff
@@ -0,0 +1,48 @@
...
+from __future__ import annotations
+
+
+import click
+
+from edb import buildmeta
+from edb.tools.edb import edbcommands
+
+
+@edbcommands.command("config")
```
1st1, line 29: can you add a help line for this command describing what it does

1st1 (commented):
Nice work!

edb/load_ext/main.py
elprans: I don't see a need to be this specific, it's a generated file in a tarball.

msullivan: Currently it's a hand-written file, though it could also be generated by the build system

elprans (approved):

PR #7740, opened 2024-09-11
Refactor how click arguments to the server are processed

msullivan:
It makes things a little cleaner, I think.  I pulled this out of
another pending change (#7738) for cleanliness and also because I
think I might change that script back to using argparse, to make it
easier to extract if someone wants to.


fantix (approved):
Cool!

PR #7742, opened 2024-09-12
Send `std::net:http` requests

scotttrinh:
### TODO

- [X] Get batches of pending `std::net::http::ScheduledRequest` objects
- [X] Create httpx request task
- [x] Run this work queue on server
- [x] Add end-to-end tests with test HTTP server

### DIFF
--- a/edb/server/net_worker.py
+++ b/edb/server/net_worker.py
@@ -126,18 +126,23 @@ class ScheduledRequest:
     method: str
     url: str
     body: typing.Optional[bytes]
-    headers: typing.Optional[list[tuple[str, str]]]
+    headers: typing.Optional[list[dict]]
 
 
 async def handle_request(
     client: httpx.AsyncClient, db: dbview.Database, request: ScheduledRequest
 ) -> None:
     try:
+        headers = (
+            [(header["name"], header["value"]) for header in request.headers]
+            if request.headers
+            else None
+        )
         response = await client.request(
             request.method,
             request.url,
             content=request.body,
-            headers=request.headers,
+            headers=headers,
         )
         request_state = 'Completed'
         failure = None
--- a/edb/server/net_worker.py
+++ b/edb/server/net_worker.py
@@ -1,0 +1,243 @@
+#
+# This source file is part of the EdgeDB open source project.
+#
+# Copyright 2024-present MagicStack Inc. and the EdgeDB authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from __future__ import annotations
+from dataclasses import dataclass
+from enum import Enum
+
+import json
+import typing
+
+import asyncio
+import logging
+
+import httpx
+
+from edb.ir import statypes
+from edb.server.protocol import execute
+from . import dbview
+
+if typing.TYPE_CHECKING:
+    from edb.server import server as edbserver
+    from edb.server import tenant as edbtenant
+
+
+logger = logging.getLogger("edb.server")
+POLLING_INTERVAL = statypes.Duration(microseconds=10 * 1_000_000)  # 10 seconds
+
+
+async def _http(tenant: edbtenant.Tenant) -> None:
+    try:
+        async with asyncio.TaskGroup() as g:
+            for db in tenant.iter_dbs():
+                json_bytes = await execute.parse_execute_json(
+                    db,
+                    """
+                    with
+                        PENDING_REQUESTS := (
+                            select std::net::http::ScheduledRequest
+                            filter .status = std::net::RequestState.Pending
+                        ),
+                        UPDATED := (
+                            update PENDING_REQUESTS
+                            set {
+                                status := std::net::RequestState.InProgress,
+                            }
+                        ),
+                    select UPDATED {
+                        id,
+                        method,
+                        url,
+                        body,
+                        headers,
+                    }
+                    """,
+                    cached_globally=True,
+                )
+                pending_requests: list[dict] = json.loads(json_bytes)
+                for pending_request in pending_requests:
+                    request = ScheduledRequest.from_db_record(pending_request)
+                    g.create_task(handle_request(db, request))
+    except Exception as ex:
+        logger.debug(
+            "HTTP send failed (instance: %s)",
+            tenant.get_instance_name(),
+            exc_info=ex,
+        )
+
+
+async def http(server: edbserver.BaseServer) -> None:
+    while True:
+        try:
+            tasks = [
+                tenant.create_task(_http(tenant), interruptable=False)
+                for tenant in server.iter_tenants()
+                if tenant.accept_new_tasks
+            ]
+            if tasks:
+                await asyncio.wait(tasks)
+        except Exception as ex:
+            logger.debug("HTTP worker failed", exc_info=ex)
+        finally:
+            await asyncio.sleep(
+                POLLING_INTERVAL.to_microseconds() / 1_000_000.0
+            )
+
+
+class HTTPMethod(str, Enum):
+    GET = 'GET'
+    POST = 'POST'
+    PUT = 'PUT'
+    DELETE = 'DELETE'
+    HEAD = 'HEAD'
+    OPTIONS = 'OPTIONS'
+    PATCH = 'PATCH'
+
+
+@dataclass
+class ScheduledRequest:
+    id: str
+    method: HTTPMethod
+    url: str
+    body: typing.Optional[bytes]
+    headers: typing.Optional[dict[str, str]]
+
+    @staticmethod
+    def from_db_record(record: dict) -> ScheduledRequest:
+        id = record['id']
+        if not isinstance(id, str):
+            type_name = type(id).__name__
+            raise TypeError(f"Expected 'id' to be str, got {type_name}")
+
+        method = record['method']
+        if not isinstance(method, str):
+            type_name = type(method).__name__
+            raise TypeError(
+                f"Expected 'method' to be str, got {type_name}"
+            )
+        else:
+            try:
+                method = HTTPMethod(method)
+            except ValueError:
+                raise ValueError(f"Invalid HTTP method: {method}")
+
+        url = record['url']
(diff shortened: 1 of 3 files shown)


edb/server/net_worker.py
```diff
@@ -0,0 +1,243 @@
...
+            id=id,
+            method=method,
+            url=url,
+            body=body,
+            headers=headers,
+        )
+
+
+async def http_send(request: ScheduledRequest) -> httpx.Response:
+    async with httpx.AsyncClient() as client:
```
fantix, line 183: Maybe we can cache an `httpx.AsyncClient` per tenant to pool connections, but that also leads to racing cookie issues among concurrent requests. So yeah, this is probably fine.

edb/server/net_worker.py
```diff
@@ -0,0 +1,243 @@
...
+                        body,
+                        headers,
+                    }
+                    """,
+                    cached_globally=True,
+                )
+                pending_requests: list[dict] = json.loads(json_bytes)
+                for pending_request in pending_requests:
+                    request = ScheduledRequest.from_db_record(pending_request)
+                    g.create_task(handle_request(db, request))
```
fantix, line 75: We should limit the maximum number of concurrent requests here, with probably a config like `net::Config.max_concurrent_net_requests` or something like that (Cloud users shouldn't be able to modify this). Implementation-wise, we can use either an `asyncio.Semaphore` or `asyncio.Queue`, both have some usage in our codebase I think.

edb/server/net_worker.py
```diff
@@ -0,0 +1,243 @@
...
+                if tenant.accept_new_tasks
+            ]
+            if tasks:
+                await asyncio.wait(tasks)
+        except Exception as ex:
+            logger.debug("HTTP worker failed", exc_info=ex)
+        finally:
+            await asyncio.sleep(
+                POLLING_INTERVAL.to_microseconds() / 1_000_000.0
+            )
```
fantix, line 99: Are we planning to interrupt this sleep with notifications later?

fantix (commented):
Overall this looks good!

edb/server/net_worker.py
scotttrinh: "notifications" meaning a Postgres `notify` event? I think so! I'm not sure exactly how to set that up, but the end goal is definitely push rather than polling (maybe with a longer poll still happening in the case of missed notifications).

edb/server/net_worker.py
scotttrinh: I'll start out with a hard-coded max in this initial implementation.

1st1: +1 to what Fantix said.

edb/server/net_worker.py
1st1: Is the idea of caching AsyncClient to minimize number of Python object invocations? Is it expensive to instantiate? Or would it reuse some underlying resources?

edb/server/net_worker.py
```diff
@@ -0,0 +1,243 @@
...
+
+
+async def handle_request(
+    db: dbview.Database, request: ScheduledRequest
+) -> None:
+    try:
+        response = await http_send(request)
+        request_state = 'Completed'
+        response_status = response.status_code
+        response_body = response.content
```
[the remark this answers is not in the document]
scotttrinh: `read` isn't awaitable? It seems like it returns `bytes` here? Maybe the change is:

```suggestion
        response_body = response.read()
```

The documentation doesn't seem to give a really clear picture on how to do what we're trying to do here which is save the response bytes without any extra decoding. Although... Maybe we _want_ it decoded (to account for compression) and the re-encode it so that we're saving something that can be easily to deal with on the consumer's end?

edb/server/net_worker.py
```diff
@@ -0,0 +1,243 @@
...
+    PATCH = 'PATCH'
+
+
+@dataclass
+class ScheduledRequest:
+    id: str
+    method: HTTPMethod
+    url: str
+    body: typing.Optional[bytes]
+    headers: typing.Optional[dict[str, str]]
```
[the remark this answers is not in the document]
scotttrinh: Who knew? 😅  The docs are pretty sparse on details here: https://www.python-httpx.org/api/#headers but the types are helpful: https://github.com/encode/httpx/blob/0aa20e449e742f81fbc518ce793878015226652d/httpx/_types.py#L57-L63

edb/server/net_worker.py
fantix: > reuse some underlying resources

Yes, it reuses the TCP/HTTP connections with `keep-alive`, avoiding some connection establishment time (which we don't care much, because it is supposed to be async and slow). In case of HTTP/2, it could multiplex concurrent requests over the same TCP connection, allowing higher concurrency with minimal resource.

The cost to create an `AsyncClient` is not that much, Python objects are also cleared away after `__aexit__()`. However, the cookies store is shared among all requests on the same client, this may or may not be the expected behavior. For example, `AsyncClient` would follow redirections; if the redirected request depends on the cookie from the initial response, 2 concurrent requests with both redirections on the same domain may step on each other's feet.

In the end of the day, it's still simpler to just use a separate client per request. But I guess in the future we can make it an option/config for the user to decide if needed.

edb/server/net_worker.py
scotttrinh: I'm not 100% sure I understand how tenants and instances interact, but would the semaphore be global to this module, or would it want to be scoped within the `_http` function such that it was per-tenant?

edb/server/net_worker.py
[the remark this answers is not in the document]
fantix: Oh sorry it should be `await response.aread()`.

> without any extra decoding

I suppose you meant the textual codecs (UTF-8, ISO-8859-1, etc.) instead of the protocol codes (gzip, deflate, etc.), right? There is the `response.aiter_text()` but we'll then need to handle the decoding errors - coming back to the previous question of storing partial responses. I'd prefer we just keep the `bytes` (only decoding the gzip/deflate/... stuff).

Or maybe we can try to decode it into `str` and - like you said - store encoded bytes with confidence (like with an additional optional field on the `Response` indicating the confirmed text encoding, or even always re-encoded as UTF-8), or just store the raw bytes if we failed to decode.

edb/server/net_worker.py
scotttrinh: It's handy that the client can handle cookies, but I don't think any user (ourselves included) should depend on something stateful like cookies when using this module. Maybe there is a way to disable cookies altogether?

edb/server/net_worker.py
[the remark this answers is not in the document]
scotttrinh: > I suppose you meant the textual codecs (UTF-8, ISO-8859-1, etc.) instead of the protocol codes (gzip, deflate, etc.), right?

Ahh, I mostly meant compression, not text encodings, but it looks like `response.content` already handles decompression:

https://www.python-httpx.org/quickstart/#binary-response-content

> Any gzip and deflate HTTP response encodings will automatically be decoded for you. If brotlipy is installed, then the brotli response encoding will be supported. If zstandard is installed, then zstd response encodings will also be supported.

> Oh sorry it should be await response.aread().

👍 

> Or maybe we can try to decode it into str and - like you said - store encoded bytes with confidence (like with an additional optional field on the Response indicating the confirmed text encoding, or even always re-encoded as UTF-8), or just store the raw bytes if we failed to decode.

The main problem I was worried about was having consumers have to read a bunch of headers to know how to decompress the bytes into the encoded text, but since that's not really the issue here, I'm personally fine with just storing the raw bytes, and having the consumer decode themselves in what is most likely utf-8. Although, maybe that's relatively hostile since we imagine that these responses will typically only be read at troubleshooting time, and typically via the UI which would be a subpar experience.

On the other hand, I _really_ don't want to make this part of the module super complicated and the simplest thing to actually store is the raw bytes. @1st1 any thoughts here?

edb/server/net_worker.py
scotttrinh: Oh, one other nice thing we'd get out of sharing a client is that we can set the max connections in the client itself to avoid managing the concurrency directly ourselves.

edb/server/net_worker.py
fantix: I think the per-tenant `_http()`-scoped Semaphore is better, as it mitigates the noisy neighbor issue on multi-tenant hosts (one tenant could still issue requests if another tenant is overloading its Semaphore). It does make it slightly harder to estimate system resources allocation when configuring multi-tenant hosts, like you'll need to multiply the per-tenant max connections by the number of tenants, but that's all.

This has a precondition that `_http()` tasks don't overlap for a single tenant (like it is now). Maybe we'll change how this is done with probably a Queue in the future when we add the notification stuff, but I guess it's okay for now.

edb/server/net_worker.py
fantix: > Maybe there is a way to disable cookies altogether?

https://github.com/encode/httpx/issues/2992 looks like we'll need to hack our own way in.

> we can set the max connections in the client itself

Yep, there's also that. And it'll be the hardware max connections in case of HTTP/2 (like the number of FDs I believe), which is something we really want I guess (instead of logical number of concurrent requests).

scotttrinh: I think it makes sense to have a null cookie jar, share a single AsyncClient per-tenant, and set the max connections on the client. I'll throw that together and we can iterate from there if we want something more robust.

scotttrinh: @fantix ⚡ [`b7f7731` (#7742)](https://github.com/edgedb/edgedb/pull/7742/commits/b7f773114d32e6e97ae936dc832c5e6c29c4bd4e)

edb/server/net_worker.py
```diff
@@ -0,0 +1,195 @@
...
+    DELETE = 'DELETE'
+    HEAD = 'HEAD'
+    OPTIONS = 'OPTIONS'
+    PATCH = 'PATCH'
+
+
+@dataclasses.dataclass
+class ScheduledRequest:
+    id: str
+    method: HTTPMethod
```
fantix, line 134: Maybe adding a converter in `__post_init__()` to set `self.method = HTTPMethod(self.method)`?

(also I think we can use https://docs.python.org/3/library/enum.html#enum.StrEnum now 🤔 )

fantix (approved):
Awesome!

edb/server/net_worker.py
scotttrinh: Do you think there is much reason to prefer an enum here over a raw str since we're not validating any of the rest of the data?

fantix: Ah not really - since you defined the enum 🤷‍♂️ not a big deal I suppose

scotttrinh: Lemme just get rid of the enum, I mostly meant that for validation and didn't remember to 🪓 it.

fantix (approved):
Good to go! :+1:

Issue #7737, opened 2024-09-10
Move standard libraries into std module


dnwpark:
For example, `math::abs` should become `std::math::abs`.

Additionally, expand unqualified names in `with module` statements (eg. `with module math`)

dnwpark:
Leaving `cfg`, `schema`, and `sys` as top level modules.

PR #7743, opened 2024-09-12
Move standard libraries into std module.

dnwpark:
Moved `cal`, `fts`, `math`, and `pg` to `std::cal`, `std::fts`, `std::math`, and `std::pg` respectively.

related #7737


1st1:
I think we also need to fix WITH as part of this PR

aljazerzen (approved):

PR #7748, opened 2024-09-13
Globals over SQL adapter

aljazerzen:
Before this PR, SQL params were not changed at all during compilation so the IO process was able to pass the external arguments (from user) directly as internal argument to postgres.

This PR:
- teaches resolver to care about params, which will combine params from actual params in SQL and params needed from injecting globals into DML and computeds,
- adds code for remapping external arguments to internal arguments, in Bind message and also ParameterDescription message,
- pulls the value of global params from connection settings of format `global my_module::my_global`,
- encode global values into their binary representation when using SET (currently, we only support strings)


edb/pgsql/resolver/command.py
```diff
@@ -1434,6 +1437,23 @@ def _compile_uncompiled_dml(
             output_format=pgcompiler.OutputFormat.NATIVE_INTERNAL,
             alias_generator=ctx.alias_generator,
         )
+
+        # merge the params need for DML with params for the rest of the query
+        param_remapping: Dict[int, int] = {}
+        for arg_name, arg in sql_result.argmap.items():
+            internal_index = len(ctx.query_params) + 1
```
msullivan, line 1444: Should we reuse an index if the same global is used more than once?

aljazerzen:
A bit of documentation for this:

Schema

```
module default {
  global current_user_id: uuid;
}
module my_mod {
  global my_glob: tuple<str, int16>;
}
```

SQL "set global":

```sql
SET "global default::current_user_id" TO '05ecd797-0a29-4ade-bf69-326d0aa3e1b4';

SET "global my_mod::current_user_id" TO '05ecd797-0a29-4ade-bf69-326d0aa3e1b4';
```

SQL variable commands

```sql
SHOW search_path;

SHOW "global default::current_user_id";
> '05ecd797-0a29-4ade-bf69-326d0aa3e1b4'

RESET "global default::current_user_id";
```

Works with transactions

```sql
BEGIN;

SET LOCAL "global default::current_user_id" TO '05ecd797-0a29-4ade-bf69-326d0aa3e1b4';

SHOW "global default::current_user_id";
> '05ecd797-0a29-4ade-bf69-326d0aa3e1b4'

COMMIT;

SHOW "global default::current_user_id";
> NULL
```

```sql
BEGIN;

SET "global default::current_user_id" TO '05ecd797-0a29-4ade-bf69-326d0aa3e1b4';

SHOW "global default::current_user_id";
> '05ecd797-0a29-4ade-bf69-326d0aa3e1b4'

ROLLBACK;

SHOW "global default::current_user_id";
> NULL
```

dev-notes/SQL_DML.drawio.png
msullivan, line 1: Is there a source for this we can check in also?

aljazerzen: It's embedded in the .png (some comment field), so it can be imported into app.diagrams.net application.

edb/server/protocol/pg_ext.pyx
```diff
@@ -950,16 +953,26 @@ cdef class PgConnection(frontend.FrontendConnection):
                 )
             parse_unit = parse_action.query_unit
             actions.append(parse_action)
+
+            # remap argumnets, which will inject globals
+            bind_data = bytes(
+                remap_arguments(
+                    b"\x00\x01\x00\x01\x00\x00\x00\x00",
```
msullivan, line 960: Could you add a comment explaining this bytestring?

edb/server/protocol/pg_ext.pyx
```diff
@@ -1543,6 +1566,100 @@ cdef class PgConnection(frontend.FrontendConnection):
         return qu
 
 
+cdef WriteBuffer remap_arguments(
+    data: bytes,
+    params: list[dbstate.SQLParam] | None,
+    fe_settings: dbstate.SQLSettings
+):    
+    # The "external" parameters (that are visible to the user)
```
msullivan, line 1574: Probably it is worth declaring some of the variables with c types, so the code uses unboxed ints

msullivan (approved):

aljazerzen:
More docs:

Due to SQL SET syntactical limitations, setting values of globals is supported only for the following types:
- str,
- uuid (using hex-encoded strings),
- int64, int32, int16,
- float32, float64,
- bool (using integers 0 and 1).

PR #7753, opened 2024-09-16
Search std for module name when using `with module`.

dnwpark:
related #7737

### DIFF
--- a/tests/test_edgeql_expressions.py
+++ b/tests/test_edgeql_expressions.py
@@ -9905,3 +9905,196 @@ aa \
             await self.con.execute(f"""
                 select <cal::to_local_date>1;
             """)
+
+    async def test_edgeql_expr_with_module_01(self):
+        await self.con.execute(f"""
+            create module dummy;
+        """)
+
+        valid_queries = [
+            'SELECT <int64>{} = 1',
+            'SELECT <std::int64>{} = 1',
+            'WITH MODULE dummy SELECT <int64>{} = 1',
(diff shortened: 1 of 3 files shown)


msullivan (commented):
There is unfortunately *another* location where names get resolved that probably also needs to be updated: https://github.com/edgedb/edgedb/blob/6ba3152590fa55d9b779f790bf0ee68065080767/edb/edgeql/tracer.py#L371-L419

That operates when loading schemas, so you'll want to try write `test_schema.py` tests that load schemas that have this sort of WITH module nonsense inside a computed or an alias, I guess.

tests/test_edgeql_expressions.py
```diff
@@ -9905,3 +9905,200 @@ async def test_edgeql_cast_to_function_01(self):
...
+                edgedb.errors.InvalidReferenceError,
+                "Foo' does not exist",
+            ):
+                await self.con.execute(query)
+
+    async def test_edgeql_expr_with_module_05(self):
+        await self.con.execute(f"""
+            create module dummy;
+            create module std::test;
+            create function std::test::Foo(x: int64) -> int64 using(x);
```
msullivan, line 10035: There is already a `std::_test` module and a `std::_test::abs` function; we can use those

dnwpark:
@msullivan Not entirely sure this is what you had in mind, but I've added schema and describe tests.

dnwpark:
@msullivan the tracer code already did what we wanted, it was just the schema that needed updating.

msullivan (approved):

PR #7756, opened 2024-09-16
Add support for postgis types in indexes.

vpetrovykh:
Postgis adds `geometry` and `geography` to `gist`, `spgist`, and `brin` indexes.

### DIFF
--- a/edb/schema/indexes.py
+++ b/edb/schema/indexes.py
@@ -73,6 +73,14 @@ def is_index_valid_for_type(
 ) -> bool:
     # HACK: currently this helper just hardcodes the permitted index & type
     # combinations, but this should be inferred based on index definition.
+
+    # TODO: remove the explicit hardcode of types compatible with indexes and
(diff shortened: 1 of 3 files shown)


edb/schema/indexes.py
```diff
@@ -91,14 +91,39 @@ def is_index_valid_for_type(
...
-            return expr_type.is_range() or expr_type.is_multirange()
+            return (
+                expr_type.is_range()
+                or
+                expr_type.is_multirange()
+                or
+                expr_type.issubclass(
+                    schema,
+                    (
+                        schema.get('ext::postgis::geometry',
```
msullivan, line 102: Could you add a `TODO` comment here reminding us to remove these *and* reminding us to remove the `Optional` hack in `issubclass` when we do?

msullivan (approved):

edb/schema/indexes.py
vpetrovykh: Added

PR #7757, opened 2024-09-16
Add custom serialization for some PostGIS types.

vpetrovykh:
Scalars `box2d` and `box3d` need custom serialization for a binary protocol because they don't have one provided by PostGIS. However, they can be serialized as `geometry` instead because there are casts between both of these types and geometry that preserve all relevant information.

### DIFF
--- a/edb/ir/typeutils.py
+++ b/edb/ir/typeutils.py
@@ -196,6 +196,28 @@ def is_persistent_tuple(typeref: irast.TypeRef) -> bool:
         return False
 
 
+def get_custom_serialization(
+    typeref: irast.TypeRef,
+) -> Optional[str]:
+    # FIXME: instead of hardcode we need to extract this from the
+    # schema/extension
+    if str(typeref.real_base_type.name_hint) in {
+        'ext::postgis::box2d',
+        'ext::postgis::box3d',
+    }:
+        return 'geometry'
+
+    return None
+
+
+def needs_custom_serialization(typeref: irast.TypeRef) -> bool:
+    # True if any component needs custom serialization
+    return contains_predicate(
+        typeref,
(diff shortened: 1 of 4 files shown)


edb/ir/typeutils.py
```diff
@@ -196,6 +196,28 @@ def is_persistent_tuple(typeref: irast.TypeRef) -> bool:
         return False
 
 
+def get_custom_serialization(
+    typeref: irast.TypeRef,
+) -> Optional[str]:
+    # FIXME: instead of hardcode we need to extract this from the
+    # schema/extension
+    if typeref.real_base_type.id in {
+        s_obj.get_known_type_id('ext::postgis::box2d'),
```
msullivan, line 205: Let's use `name_hint` instead of adding known type ids for this.

edb/schema/_types.py
```diff
@@ -66,4 +66,8 @@
         UUID('00000000-0000-0000-0000-000000000112'),
     sn.name_from_string('cfg::memory'):
         UUID('00000000-0000-0000-0000-000000000130'),
+    sn.name_from_string('ext::postgis::box2d'):
+        UUID('7fae5536-6311-4f60-8eb9-096a5d972f48'),
+    sn.name_from_string('ext::postgis::box3d'):
+        UUID('c1a50ff8-fded-48b0-85c2-4905a8481433'),
```
msullivan, line 72: Drop this

msullivan (approved):

edb/ir/typeutils.py
vpetrovykh: Replaced with `name_hint`

edb/schema/_types.py
vpetrovykh: Removed

PR #7759, opened 2024-09-17
Allow complex types as function params.

dnwpark:
Allow functions such as `create function foo(x: A | B) -> int64 using (x.n);`


msullivan (approved):
Nice catch

msullivan:
Oh but it still doesn't work if it's in the schema!

PR #7763, opened 2024-09-17
Simplify path factoring in the scope tree

msullivan:
There isn't really any good reason to have the "factor with a visible
path" and "factor with a not visible path" code paths be different,
except that there were some bugs in the latter.

Fix and merge them.


elprans (approved):
Is there a bug this is fixing or a new test that exposes the fixed bug?

msullivan:
I think there is probably a bug that this fixes, but it'll be pretty obscure

PR #7764, opened 2024-09-18
Workaround Cython bug under Clang

fantix:
`cimport`-ing an `cdef inline` function is compiled into a CYTHON_INLINE function pointer by Cython (both 0.x and 3.x), which isn't allowed in Clang.

Refs #7747, fixes nightlies on macOS


msullivan (approved):
Thanks for figuring this out!

PR #7767, opened 2024-09-18
Get std::net::http max connections from config

scotttrinh:


edb/server/net_worker.py
```diff
@@ -55,10 +52,14 @@ def set_cookie(self, _):
 
 
 async def _http(tenant: edbtenant.Tenant) -> None:
+    net_http_max_connections = tenant._server.config_lookup(
+        'net_http_max_connections', tenant.get_sys_config()
+    )
```
scotttrinh, line 57: Is this the correct way to get per-tenant config values?

fantix: Yes I think so - I'd also add:

```
        CREATE ANNOTATION cfg::system := 'true';
```

in `cfg.edgeql`, because database-level or session-level config is never effective in this impl.

Btw, usually we shall also apply new config values in `on_remote_system_config_change()`, however in our case it's currently reloaded every 10 seconds, so we're good for now.

fantix (approved):

docs/reference/configuration.rst
fantix, line 1: Also bumping the `EDGEDB_CATALOG_VERSION`?

edb/server/net_worker.py
scotttrinh: > Btw, usually we shall also apply new config values in on_remote_system_config_change(), however in our case it's currently reloaded every 10 seconds, so we're good for now.

Do we have an example of that? I couldn't find anything just grepping around. Seems worth adding now so we don't accidentally regress this if we change the current polling situation (seems likely we'll change this soon)

fantix: It'll be just like calling `reinit_idle_gc_collector()` in `on_remote_system_config_change()`, but we can't really do this now because the `AsyncClient` instance is not in this context. This is also related to - I was thinking - how we handle future notifications, like it'll take up 1 connection per branch (not feasible if the user has many many branches), and forward the notification to a pool of workers through a queue or something, and this `on_remote_system_config_change()` would update the number of workers in the pool. This is a bit more complex than I expected, I'll also need to think more about it; or maybe you have some other ideas already ❤️

PR #7772, opened 2024-09-19
Disable net_worker for conflicting tests

fantix:
This adds a hidden option/envvar `--net-worker-mode` that is turned off for tests that have conflict with the current net_worker.

Also takes back a retry around DROP DATABASE for tests as a workaround.

[Sample run](https://github.com/edgedb/edgedb/actions/runs/10946518415)

### DIFF
--- a/edb/server/args.py
+++ b/edb/server/args.py
@@ -164,6 +164,11 @@ class ReloadTrigger(enum.StrEnum):
     """Watch the files for changes and reload when it happens."""
 
 
+class NetWorkerMode(enum.StrEnum):
+    Default = "default"
+    Disabled = "disabled"
+
+
 class ServerAuthMethods:
 
     def __init__(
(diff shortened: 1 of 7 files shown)


msullivan (approved):
This seems solid and correct but let's discuss out-of-band whether we think it is worth merging (given that we are landing https://github.com/edgedb/edgedb/pull/7773)

fantix:
Flipped as discussed in chat; also cherry-picked a retry.

msullivan (approved):
Do we need to do the DROP DATABASE retry in tests? Getting rid of that has surfaced a lot of bugs, and this should fix all the ones we know about, right?

(Also, the *internal* retry loop inside DROP DATABASE is already 30s and we reemit the signals inside it now, so I don't know if this loop would actually help unless it was much longer.)

fantix:
The internal 30s is only effective using `WITH (FORCE)` - should we just add it to the `drop_db()` function in `testbase/server.py`?

PR #7773, opened 2024-09-19
Don't run the net_worker when branch is not connectable

fantix:
This allows `drop branch`, `create branch ... from ...` to `ensure_database_not_connected()` correctly, avoiding `failed to ..., database being accessed by others`.


msullivan (approved):

msullivan:
We should make sure to do this for the ai workers too (doesn't need to be this PR)

PR #7777, opened 2024-09-19
Inline functions with dml.

### LINKED ISSUE #7692: Proper function DML support

dnwpark:
- Compile arguments to an iterator CTE when an inlined function contains dml
- In the edgeql -> ir compiler, detect dml in function calls by checking functions in the schema

close #7692

### DIFF
--- a/edb/edgeql/compiler/astutils.py
+++ b/edb/edgeql/compiler/astutils.py
@@ -22,12 +22,21 @@
 
 from __future__ import annotations
 from dataclasses import dataclass, field
-from typing import Optional, Tuple, Dict, List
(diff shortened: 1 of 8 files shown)


msullivan (approved):
This looks great.

PR #7778, opened 2024-09-19
Reduce disk used by inplace upgrade testing

msullivan:
Don't save tarballs of everything by default.
We were running out of disk in CI runs. This might help.


fantix (approved):
LGTM

PR #7779, opened 2024-09-19
Add custom serialization for scalars.

vpetrovykh:
Some scalar types need to be cast into some other SQL type during serialization. This can be defined as `custom_sql_serialization` on the scalar types in schema.


msullivan (approved):

msullivan:
Oh, darn, we forgot to bump the catalog version

PR #7782, opened 2024-09-19
tests: replace retry-dropdb with force-dropdb

fantix:
This is a follow-up of #7772 that uses `DROP DATABASE ... FORCE` to enter the inner 30s loop.


edb/server/dbview/dbview.pxd
```diff
@@ -124,6 +124,7 @@ cdef class DatabaseConnectionView:
         Database _db
         bint _query_cache_enabled
         object _protocol_version
+        public bint is_internal  # internal dbview doesn't have a frontend connection
```
msullivan, line 127: Maybe `is_transient`? Since IIRC it can also apply to a query from the HTTP protocol, which is not really internal, but *also* isn't long lived.

msullivan (approved):
One concern here is that a database might now get `unregister_db`d while it still has `_views`. Do we think that's fine?

fantix:
> One concern here is that a database might now get `unregister_db`d while it still has `_views`. Do we think that's fine?

Yeah, it's fine. `unregister_db()` happens in `on_after_drop_db()`, while in `on_before_drop_db()` we call `_pg_ensure_database_not_connected()` and wait for all actual pgcons to disappear. I added a piece of comments to remind the `is_transient` users to keep such pgcons transient too.

PR #7787, opened 2024-09-21
Generalize matching types to indexes.

vpetrovykh:
Make it possible to declare in the schema which types are valid for which indexes.

`create index match for <type_expr> using <index_name>;`

This enables defining new types of indexes in extensions and declare how they relate to each other without needing to update the server code.

### DIFF
--- a/edb/schema/extensions.py
+++ b/edb/schema/extensions.py
@@ -36,6 +36,7 @@ from edb.common import checked
 from . import annos as s_anno
 from . import casts as s_casts
 from . import delta as sd
+from . import indexes as s_indexes
 from . import name as sn
 from . import objects as so
 from . import schema as s_schema
@@ -522,6 +523,18 @@ class DeleteExtension(
                 )
                 casts_cleanup.append(drop)
 
+        # Similarly, index matches are kept in __ext_index_matches__. We can
+        # remove them first since nothing else depends on them.
+        for im in schema.get_objects(
+            included_modules=(sn.UnqualName('__ext_index_matches__'),),
+            type=s_indexes.IndexMatch,
+        ):
+            if (
+                _name_in_mod(im.get_valid_type(schema).get_name(schema))
+                or _name_in_mod(im.get_index(schema).get_name(schema))
+            ):
+                self.add(im.init_delta_command(schema, sd.DeleteObject))
+
         def filt(schema: s_schema.Schema, obj: so.Object) -> bool:
             return not _name_in_mod(obj.get_name(schema)) or obj == self.scls
 
@@ -548,7 +561,7 @@ class DeleteExtension(
             # scalar types. Cleaning casts earlier may cause issues with
             # functions that use casts in their signatures as part of the
             # default expression.
-            if isinstance(subast, qlast.DropScalarType) and casts_cleanup:
+            if casts_cleanup and isinstance(subast, qlast.DropScalarType):
                 self.update(casts_cleanup)
                 casts_cleanup.clear()
 
--- a/edb/schema/indexes.py
+++ b/edb/schema/indexes.py
@@ -53,12 +53,12 @@ from . import objects as so
 from . import referencing
 from . import scalars as s_scalars
 from . import types as s_types
+from . import schema as s_schema
 from . import utils
 
 
 if TYPE_CHECKING:
     from . import objtypes as s_objtypes
-    from . import schema as s_schema
 
 
 # The name used for default concrete indexes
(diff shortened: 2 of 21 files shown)


msullivan:
types -> types in the title line

edb/schema/extensions.py
```diff
@@ -548,9 +565,20 @@ def filt(schema: s_schema.Schema, obj: so.Object) -> bool:
             # scalar types. Cleaning casts earlier may cause issues with
             # functions that use casts in their signatures as part of the
             # default expression.
-            if isinstance(subast, qlast.DropScalarType) and casts_cleanup:
-                self.update(casts_cleanup)
-                casts_cleanup.clear()
+            #
+            # For the same reasons we also want to clean up index matches
+            # before we're cleaning indexes or scalar types.
```
msullivan, line 570: Can we just always drop the index matches first? We don't need to have them since they can't be used in functions like casts

edb/schema/indexes.py
```diff
@@ -74,131 +74,21 @@ def is_index_valid_for_type(
...
-                schema,
-                schema.get('std::str', type=s_scalars.ScalarType),
-            )
-        case (
-            'ext::ai::index'
-        ):
-            return expr_type.issubclass(
-                schema,
-                schema.get('std::str', type=s_scalars.ScalarType),
-            )
```
msullivan, line 201: Hell yeah.

edb/schema/indexes.py
```diff
@@ -74,131 +74,21 @@ def is_index_valid_for_type(
     # HACK: currently this helper just hardcodes the permitted index & type
     # combinations, but this should be inferred based on index definition.
 
-    # TODO: remove the explicit hardcode of types compatible with indexes and
-    # make them configurable in the schema instead. This is especially
-    # important for extension types.
-    #
+    # TODO:
```
msullivan, line 77: Can you clean up issubclass Optional bit now?

msullivan (approved):
This is fantastic, thanks

elprans (changes requested):
This is a nice cleanup, but I'd like us to think about this a bit to make sure we're choosing the right abstraction here.  Postgres matches indexes to types via [operator classes and families](https://www.postgresql.org/docs/current/indexes-opclass.html), and that system is more flexible as it allows indexing the same type with the same index but differently (e.g using different collation rules for comparison).  So I wonder if we should replicate that system in some way instead.

For example, here's how pgvector index matches are defined via operator classes:

```
CREATE OPERATOR CLASS vector_ops
	DEFAULT FOR TYPE vector USING btree AS
	OPERATOR 1 < ,
	OPERATOR 2 <= ,
	OPERATOR 3 = ,
	OPERATOR 4 >= ,
	OPERATOR 5 > ,
	FUNCTION 1 vector_cmp(vector, vector);

CREATE OPERATOR CLASS vector_l2_ops
	DEFAULT FOR TYPE vector USING ivfflat AS
	OPERATOR 1 <-> (vector, vector) FOR ORDER BY float_ops,
	FUNCTION 1 vector_l2_squared_distance(vector, vector),
	FUNCTION 3 l2_distance(vector, vector);

CREATE OPERATOR CLASS vector_ip_ops
	FOR TYPE vector USING ivfflat AS
	OPERATOR 1 <#> (vector, vector) FOR ORDER BY float_ops,
	FUNCTION 1 vector_negative_inner_product(vector, vector),
	FUNCTION 3 vector_spherical_distance(vector, vector),
	FUNCTION 4 vector_norm(vector);

CREATE OPERATOR CLASS vector_cosine_ops
	FOR TYPE vector USING ivfflat AS
	OPERATOR 1 <=> (vector, vector) FOR ORDER BY float_ops,
	FUNCTION 1 vector_negative_inner_product(vector, vector),
	FUNCTION 2 vector_norm(vector),
	FUNCTION 3 vector_spherical_distance(vector, vector),
	FUNCTION 4 vector_norm(vector);

CREATE OPERATOR CLASS vector_l2_ops
	FOR TYPE vector USING hnsw AS
	OPERATOR 1 <-> (vector, vector) FOR ORDER BY float_ops,
	FUNCTION 1 vector_l2_squared_distance(vector, vector);

CREATE OPERATOR CLASS vector_ip_ops
	FOR TYPE vector USING hnsw AS
	OPERATOR 1 <#> (vector, vector) FOR ORDER BY float_ops,
	FUNCTION 1 vector_negative_inner_product(vector, vector);

CREATE OPERATOR CLASS vector_cosine_ops
	FOR TYPE vector USING hnsw AS
	OPERATOR 1 <=> (vector, vector) FOR ORDER BY float_ops,
	FUNCTION 1 vector_negative_inner_product(vector, vector),
	FUNCTION 2 vector_norm(vector);

CREATE OPERATOR CLASS vector_l1_ops
	FOR TYPE vector USING hnsw AS
	OPERATOR 1 <+> (vector, vector) FOR ORDER BY float_ops,
	FUNCTION 1 l1_distance(vector, vector);
```

Note: different indexes require different operators (aka index access method strategies).

edb/schema/indexes.py
vpetrovykh: Cleaned up

edb/schema/extensions.py
vpetrovykh: You're right, I've moved the index matches to be dropped first.

msullivan:
I pushed a tweak that made `schema::IndexMatch` not exposed to users at all, so that this would be basically a fully internal API. I had to tweak the reflection infra a bit to make it work.

msullivan:
I think we are prepared to merge this, right? After our discussion yesterday?

elprans (approved):

Issue #6493, opened 2023-11-16
Persistent compile cache


fantix:
## Objective

* Compiled EdgeQL cache (SQLs) should survive server restarts, so that restarts won't cause potentially large drops in quality of service (like a slow query) when rebuilding the cache.
* (secondary) Reduce server memory footprint and backend network load.

## Proposal

Dynamically create individual SQL functions for each compiled EdgeQL (named/indexed by query / argument types hash and schema version) in order to replace the current in-memory LRUMapping. The EdgeDB server should keep an up-to-date list of cached queries (names of functions) to avoid extra round-trip to the backend checking the existence of the cache of a certain query, so that the server could either just execute that function (cache hit), or compile + create function +  execute function (cache miss) with minimal I/O to the backend, when handling an EdgeQL query request.

## Action items

- [ ] Test how well Postgres handles concurrent changes to the functions table under load
- [ ] Design for multi-backend-tenancy (multiple EdgeDB servers on one backend) - basically a notification for other tenants to reload what's cached in the backend
- [ ] Design the cleanup of the "cache" (functions table) when hitting `maxsize`
- [x] Make sure prepared statements work fine
- [ ] Benchmark to see how the design performs
- [ ] ~Use `CALL` Postgres protocol message to call the cached funcs~

msullivan:
We have merged a working persistent cache, but need some additional work to get functions working.

msullivan:
Current issues:

- [ ] Test failures when using https://github.com/edgedb/edgedb/pull/7497
- [ ] Increased log spew because of `DROP FUNCTION IF EXISTS`
- [ ] I once observed "function does not exist" failures trying to call the cache function. It seemed nondeterministic. Need to try to reproduce.

PR #7792, opened 2024-09-24
Reduce the postgres INFO spew when using function cache

msullivan:
Whenever postgres skips a DROP because IF EXISTS fails, it emits an
INFO. Since this happens every time a query cache misses, I found this
annoyingly noisy.

Make those operate by using an explicit condition check instead. Doing
this required removing a check that converted such checks into IF
EXISTS as well as fixing some problems with our checks.

This is one of two checkboxes left on #6493, I think, the other being
a flake I dimly remembered, so we can probably close that?

### DIFF
--- a/edb/pgsql/dbops/functions.py
+++ b/edb/pgsql/dbops/functions.py
@@ -23,7 +23,6 @@ import textwrap
 from typing import Optional, Tuple, Sequence, List, cast
 
 from ..common import qname as qn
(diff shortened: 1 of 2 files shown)


elprans (approved):
:+1:

fantix (approved):
> the other being a flake I dimly remembered, so we can probably close that?

Right, I was also paying attention to that and didn't quite see it.

PR #7796, opened 2024-09-24
Add ext::auth::WebhookConfig

scotttrinh:
Allows users to configure webhooks for certain auth extension events.

Possibly an alternative to #7541 . Perhaps we don't need an actual event type that these webhooks should use, but just have the concept of these events be the content of the bodies of the requests we make.


1st1 (commented):
Looks good!

elprans (approved):

PR #7805, opened 2024-09-26
Support accessing link properties through FOR bindings

msullivan:
So now you can write something like
```
SELECT User {
    cards := (
        (FOR d IN .deck[is SpecialCard] SELECT (d.name, d@count))
    ),

}
```
and have it work.

See [RFC 1027 - Simplifying path resolution](https://github.com/edgedb/rfcs/blob/master/text/1027-no-factoring.rst)
for some of the motivation here - this removes the necessity of using
path factoring in one of the main places it was hard to avoid.


aljazerzen (approved):

1st1 (commented):
Wow, this is nice.

dnwpark (approved):

Issue #7766, opened 2024-09-18
Remove path factoring


msullivan:
See [RFC 1027 - Simplifying path resolution](https://github.com/edgedb/rfcs/blob/master/text/1027-no-factoring.rst)

msullivan:
Reopening this, because we still must remove it.

PR #7809, opened 2024-09-26
Initial implementation of "RFC 1027 - Simplfying path resolution"

msullivan:
See [RFC 1027 - Simplifying path resolution](https://github.com/edgedb/rfcs/blob/master/text/1027-no-factoring.rst) and  #7766.

There are still some improvements I would like to make to the `warn_old_scoping` mode, and we should update and triage more of the test suite to use `simple_scoping` and/or `warn_old_scoping`.
    
Note that while a lot of tests rely on path factoring, I expect this is very unrepresentative of actual EdgeQL in the wild. There are a lot of tests that use path factoring because path factoring needed a lot of tests to make sure it was right.

### DIFF
--- a/edb/lib/cfg.edgeql
+++ b/edb/lib/cfg.edgeql
@@ -192,6 +192,19 @@ ALTER TYPE cfg::AbstractConfig {
             'Whether inserts are allowed to set the \'id\' property.';
     };
 
+    CREATE PROPERTY simple_scoping -> std::bool {
+        CREATE ANNOTATION cfg::affects_compilation := 'true';
+        CREATE ANNOTATION std::description :=
+            'Whether to use the new simple scoping behavior \
+            (disable path factoring)';
+    };
+
+    CREATE PROPERTY warn_old_scoping -> std::bool {
+        CREATE ANNOTATION cfg::affects_compilation := 'true';
+        CREATE ANNOTATION std::description :=
+            'Whether to warn when depending on old scoping behavior.';
+    };
+
     CREATE MULTI PROPERTY cors_allow_origins -> std::str {
         CREATE ANNOTATION std::description :=
             'List of origins that can be returned in the \
--- a/tests/test_edgeql_insert.py
+++ b/tests/test_edgeql_insert.py
@@ -17,6 +17,7 @@
 #
 
 
+import contextlib
 import os.path
 import uuid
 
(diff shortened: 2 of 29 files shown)


elprans (approved):
LGTM overall, though there is a bunch of XXX's which should probably be addressed before this is merged.

tests/test_edgeql_insert.py
```diff
@@ -6473,6 +6487,7 @@ async def test_edgeql_insert_conditional_01(self):
             variables=([True, False],)
         )
 
+        # XXX: THIS IS FAILING BUT THAT IS BULLSHIT
```
aljazerzen, line 6490: Why would this be failing?

Which prefix is being factored out?

edb/lib/cfg.edgeql
```diff
@@ -192,6 +192,19 @@ ALTER TYPE cfg::AbstractConfig {
...
+        CREATE ANNOTATION std::description :=
+            'Whether to use the new simple scoping behavior \
+            (disable path factoring)';
+    };
+
+    CREATE PROPERTY warn_old_scoping -> std::bool {
+        CREATE ANNOTATION cfg::affects_compilation := 'true';
+        CREATE ANNOTATION std::description :=
+            'Whether to warn when depending on old scoping behavior.';
+    };
```
aljazerzen, line 206: What's the default value here? False? Should we have this be True?

aljazerzen (approved):
This was a big one, great work.

edb/lib/cfg.edgeql
msullivan: The default value is not having a value, and falling back to the presence or absence of the future.
(This is kind of tricky; there is a table in the RFC)

msullivan:
> Honestly, I was expecting more path factoring being used in the tests. If I understand correctly, of all insert tests, only three used it?

Yeah. I'll want to check more test suites, but I expect factoring to be heavily used in `select`, `scope`, `coalesce`, and `volatility`, but not all that much elsewhere?
I could be wrong, or forgetting some suites.

tests/test_edgeql_insert.py
msullivan: I think it's something hinky with how IF is compiled for DML.

Hm.

msullivan: Ah, but more important, it is *not* failing. I fixed it (because it seemed critically bad), and never removed the comment.

PR #7810, opened 2024-09-27
Precompile SELECT set_config to SET

aljazerzen:


aljazerzen:
The conclusion here is that:
- we need to get pg_dump working now, so let's just add another workaround instead of this PR,
- I will come back to this later, hopefully after pgcon.pxd will be rewritten in Rust.

PR #7811, opened 2024-09-27
Add debug flag to disable net worker

aljazerzen:
Adds `EDGEDB_DEBUG_DISABLE_NET_WORKER` flag


scotttrinh:
I think this already exists, no?

https://github.com/edgedb/edgedb/pull/7772

aljazerzen:
Ah yes it is, with env var EDGEDB_SERVER_NET_WORKER_MODE!

It was bothering me when I had compiler debug output enabled, spewing 100+ lines of AST dumps every few seconds.

PR #7812, opened 2024-09-27
Update the pgvector extension.

vpetrovykh:
Update the underlying pgvector version to 0.7.4.
Add `sparsevec` and `halfvec` types and update the valid indexes. Add one more distance operator (L1 distance).

### DIFF
--- a/edb/lib/ext/pgvector.edgeql
+++ b/edb/lib/ext/pgvector.edgeql
@@ -17,9 +17,135 @@
 #
 
 
-create extension package pgvector version '0.5.0' {
+create extension package pgvector version '0.7.4' {
     set ext_module := "ext::pgvector";
-    set sql_extensions := ["vector >=0.5.0,<0.7.0"];
+    set sql_extensions := ["vector >=0.7.4,<0.8.0"];
+
+    set sql_setup_script := $script$
+        -- Rename the vector_norm to be consistent with l2_norm
+        ALTER FUNCTION edgedb.vector_norm(edgedb.vector) RENAME TO l2_norm;
+
+        -- Add some helpers
+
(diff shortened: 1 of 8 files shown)


vpetrovykh:
After working on the bindings I am thinking that maybe `sparsevec` should have a custom `json` cast that mimics the Python dict rather than just wraps the string representation.

vpetrovykh:
Annoyingly, the 1-based indexing is leaking into the `str` (and `json`) representation for `sparsevec`. This, in turn, means that `subvector` (not included here because I'd like to finalize the question around indexes) is also using 1-based indexes.

I'm wondering if we have any patching options rather than slow re-parsing of the strings in SQL to adjust indexes.

vpetrovykh:
The binary representation (both for encoding and decoding) as well as _internal_ representation all use 0-based indexes. These get adjusted when handling SQL strings. Since we're re-packaging this extension is it possible for us to patch the functions involved in handling the strings at all (like adding our own functions we can use in casts maybe)?

edb/lib/ext/pgvector.edgeql
```diff
@@ -17,9 +17,135 @@
...
+
+        -- about 10 times slower than a cast
+        CREATE FUNCTION text_to_sparsevec(val text) RETURNS sparsevec AS $$
+        DECLARE
+            mid text[];
+            kv text[];
+            i int8;
+            res text := '{';
+        BEGIN
+            IF val ~ '^\s*{\s*(\d+\s*:.+?,\s*)*\d+\s*:.+}\s*/\s*\d+\s*$'
```
msullivan, line 60: I'm going to take your word on this regex.

msullivan (approved):

PR #7813, opened 2024-09-27
Send configured Webhooks

scotttrinh:
In the auth `http.py` module, send webhooks via the `std::net` module for every configured url for a given event.

### DIFF
--- a/edb/server/net_worker.py
+++ b/edb/server/net_worker.py
@@ -25,6 +25,7 @@ import typing
 import asyncio
 import logging
 import httpx
+import base64
 
 from edb.ir import statypes
 from edb.server import defines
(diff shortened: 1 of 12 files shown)


edb/server/net_worker.py
```diff
@@ -134,6 +135,10 @@ class ScheduledRequest:
     body: typing.Optional[bytes]
     headers: typing.Optional[list[dict]]
 
+    def __post_init__(self):
+        if self.body is not None:
+            self.body = base64.b64decode(self.body).decode('utf-8').encode()
```
fantix, line 140: Ahhh, this is unfortunate but okay for now - we should figure out a better way to run queries internally other than `parse_execute_json()` I suppose

fantix (commented):
This looks promising!

Issue #7725, opened 2024-09-06
Query performance observability


msullivan:
Mostly a placeholder issue for now.
We want to provide visibility into query performance

I think the overall plan here is to use `pg_stat_statements` (https://www.postgresql.org/docs/current/pgstatstatements.html) to collect information.

I think the core questions are:

1. How do we connect our EdgeQL queries with information from pg_stat_statments?

   We forked pg_stat_statements to record custom information like EdgeQL query string, compilation settings, etc.

2. How do we expose the information to our users?

   `sys::QueryStats` is added as a view of the `pg_stat_statements` backend view.

- [x] Add edb_stat_statements (#7869)
- [x] Fix CI build (edgedb/edgedb-pkg#124)
- [x] Add config to turn off stats (#8033)
- [x] [Add `tag` to identify the source of queries](https://github.com/edgedb/edgedb/pull/7869#issuecomment-2471526873) (#8036)
- [x] Set query tag in CLI (edgedb/edgedb-cli#1408)
- [x] Set query tag in edgedb-python (edgedb/edgedb-python#552)
- [ ] Set query tag in edgedb-js (edgedb/edgedb-js#1133)

fantix:
Closing as fixed, please see each linked PR for details.

PR #7814, opened 2024-09-27
Add EdgeDB query performance statistics

fantix:
Refs #7725, this is ~a quick'n'dirty first attempt~ the first implementation with a standalone extension that manipulates the PG query source text (and optionally the queryID to map to the cache key). The upper side is that the extension is kinda clean without patching PG codebase (easier to port), ~the down side is that we lose the raw SQL stats data for all~.

- [x] Fix tests
- [x] Deal with SQL user permission issue
- [x] Fix missing hooks
- [x] Add EdgeQL view
- [x] Add backend capabilities
- [x] Handle multitenancy correctly
- [x] Play nice with persistent cache
- [x] Complete the reset function parameters
- [x] Count SQL queries in
- [x] Add columns that reflect compilation settings
- [x] Track EdgeQL scripts correctly
- [ ] Fix CI builds
- [ ] Add test

Sample results:

```
_localdev:main> select sys::QueryStats { * } filter .query not like '%QueryStats%' and .query not like '%sys::%' limit 2;
{
  sys::QueryStats {
    computed_fields: [],
    builtin: false,
    internal: false,
    name: '4ed18524-e5ee-6288-a0b0-9a171bcedd98',
    id: 4ed18524-e5ee-6288-a0b0-9a171bcedd98,
    query: 'SELECT 89',
    query_id: 5679466998613828232,
    query_type: SQL,
    compilation_config: Json("{\"allow_user_specified_id\": [\"false\"]}"),
    protocol_version: (major := 3, minor := 0),
    default_namespace: 'public',
    namespace_aliases: {},
    migration_name: 'initial',
    output_format: {},
    expect_one: {},
    implicit_limit: {},
    inline_typeids: {},
    inline_typenames: {},
    inline_objectids: {},
    plans: 0,
    total_plan_time: <duration>'0:00:00',
    min_plan_time: <duration>'0:00:00',
    max_plan_time: <duration>'0:00:00',
    mean_plan_time: <duration>'0:00:00',
    stddev_plan_time: <duration>'0:00:00',
    calls: 1,
    total_exec_time: <duration>'0:00:00.000001',
    min_exec_time: <duration>'0:00:00.000001',
    max_exec_time: <duration>'0:00:00.000001',
    mean_exec_time: <duration>'0:00:00.000001',
    stddev_exec_time: <duration>'0:00:00',
    rows: 1,
  },
  sys::QueryStats {
    computed_fields: [],
    builtin: false,
    internal: false,
    name: 'bcecb044-dd0b-af72-d014-345032436dfc',
    id: bcecb044-dd0b-af72-d014-345032436dfc,
    query: 'select
    (BB, (<__std__::int64>$0 + <__std__::int64>$1))',
    query_id: -4833294490262589582,
    query_type: EdgeQL,
    compilation_config: Json("{\"allow_bare_ddl\": \"AlwaysAllow\", \"query_cache_mode\": \"Default\", \"__internal_testmode\": false, \"store_migration_sdl\": \"NeverStore\", \"force_database_error\": \"false\", \"apply_access_policies\": true, \"allow_dml_in_functions\": false, \"allow_user_specified_id\": false, \"__internal_query_reflschema\": false, \"__internal_no_apply_query_rewrites\": false}"),
    protocol_version: (major := 2, minor := 0),
    default_namespace: 'bb',
    namespace_aliases: {},
    migration_name: 'm1ndk7eqxayvb5wdfkuv7kslyx2wnhr3e3afcmakqtgo4hjcxwspka',
    output_format: BINARY,
    expect_one: false,
    implicit_limit: 0,
    inline_typeids: false,
    inline_typenames: false,
    inline_objectids: true,
    plans: 6,
    total_plan_time: <duration>'0:00:00.001099',
    min_plan_time: <duration>'0:00:00.000086',
    max_plan_time: <duration>'0:00:00.000227',
    mean_plan_time: <duration>'0:00:00.000183',
    stddev_plan_time: <duration>'0:00:00.000048',
    calls: 10,
    total_exec_time: <duration>'0:00:00.000118',
    min_exec_time: <duration>'0:00:00.000007',
    max_exec_time: <duration>'0:00:00.000019',
    mean_exec_time: <duration>'0:00:00.000012',
    stddev_exec_time: <duration>'0:00:00.000004',
    rows: 0,
  },
}
```


elprans:
>  the down side is that we lose the raw SQL stats data for all.

But we don't have to, do we?  If we don't recognize a query as being a compiler product we can avoid modifying its state.

edb_stat_statements/edb_stat_statements.c
```diff
@@ -0,0 +1,219 @@
...
+    queryDesc->sourceText = orig_sourceText;
+}
+
+/*
+ * Extract EdgeDB statement info and overwrite source text and queryId
+ */
+static uint64
+edbss_extract_stmt_info(char **query_str) {
+    if (strncmp(*query_str, EDB_STMT_MAGIC_PREFIX, strlen(EDB_STMT_MAGIC_PREFIX)) == 0) {
+        EdbStmtInfoSemState state;
```
1st1, line 133: What version of C will this be compiled with? Because you could just do this in C99:

```suggestion
        EdbStmtInfoSemState state = {0};
```

edb/server/compiler/compiler.py
```diff
@@ -1557,6 +1559,72 @@ def _compile_ql_query(
...
+            # compilation_config
+            'cc': config.to_json_obj(
+                spec,
+                {
+                    **current_tx.get_system_config(),
+                    **current_tx.get_database_config(),
+                    **current_tx.get_session_config(),
+                },
+                setting_filter=lambda v: v.name in spec
+                                         and spec[v.name].affects_compilation,
```
msullivan, line 1589: Can we make sure nothing `secret` goes in also?

edb/server/compiler/compiler.py
```diff
@@ -1557,6 +1559,72 @@ def _compile_ql_query(
...
+    #      extension, used to replace the compiled SQL with the frontend query
+    #      in this JSON, as well as the `queryId` (unique index of statistics).
+    #   2. query_info: the inner comment that is stored in pg_stat_statements,
+    #      used as a supplement storage of the additional query information,
+    #      like query_type, output_format, etc.
+    sql_info: Dict[str, Any] = {}
+    if (
+        not ctx.bootstrap_mode
+        and ctx.backend_runtime_params.has_stat_statements
+    ):
```
msullivan, line 1574: Do we want whether we collect this to also depend on a config flag? This seems like a lot of extra stuff to store unconditionally.
(Curious about @elprans's opinion on this)

edb_stat_statements/edb_stat_statements.c
fantix: Yeah, mostly assuming C99 in #7869 now

edb/server/compiler/compiler.py
fantix: I think the only compilation setting that may include free text is `force_database_error`, which shouldn't have secrets. I can also add a double-safe to filter out extension configs in #7869.

edb/server/compiler/compiler.py
fantix: Answered in the meeting: we do need a config to turn the tracking on/off.

Follow-up question: do we want to make this config `affects_compilation` (thus controlling the compiled SQL with or without the magic JSON-comment prefix, but impacting query cache), or a runtime session config that the extension can read and behave (but we end up with prefixed SQLs all the time)?

PR #7816, opened 2024-10-01
Raise error when passing multi cardinality args to modifying functions.

dnwpark:
Related #7692

### DIFF
--- a/edb/edgeql/compiler/inference/cardinality.py
+++ b/edb/edgeql/compiler/inference/cardinality.py
@@ -187,6 +187,7 @@ def _union_cardinality(
 
 
 VOLATILE = qltypes.Volatility.Volatile
+MODIFYING = qltypes.Volatility.Modifying
 
 
 def _check_op_volatility(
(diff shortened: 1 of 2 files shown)


msullivan (commented):
I think that we also need to reject potentially empty arguments for parameters that are not OPTIONAL.

(I'm pretty sure that will break some of the tests written for https://github.com/edgedb/edgedb/pull/7777/, but oh well)

edb/edgeql/compiler/inference/cardinality.py
```diff
@@ -775,6 +776,19 @@ def __infer_func_call(
...
+            raise errors.QueryError(
+                'possibly more than one element passed into modifying function',
+                span=ir.span,
+            )
+
+        if any(card.can_be_zero() for card in cards):
+            raise errors.QueryError(
+                'possibly an empty set passed into modifying function',
+                span=ir.span,
+            )
```
msullivan, line 790: Should work if the argument is OPTIONAL

dnwpark: @msullivan Ah sorry, forgot to push the change.

msullivan (approved):

PR #7820, opened 2024-10-02
Update bundled PostgreSQL to 17.0

aljazerzen:


tests/test_edgeql_explain.py
```diff
@@ -235,25 +235,25 @@ async def test_edgeql_explain_with_bound_01(self):
...
+                                #         "index of object type 'default::User'"
+                                #         " on (__subject__.name)",
+                                #     "type": "index",
+                                #     "important": True,
+                                # },
+                                # {
+                                #     "title": "index_cond",
+                                #     "type": "expr",
+                                #     "important": False,
+                                # },
```
msullivan, line 256: We need to track this down more before merging, right?

aljazerzen: Yes

msullivan:
We should verify it still works on 16, also

aljazerzen:
TLDR; with upgrade to pg 17, some queries will stop hitting the index, unless we set operator class of the default index to `text_pattern_ops` (for all things that are `str`, probably).

---

So, the explain errors are here because a query that used to be IndexScan on pg16 gets a SeqScan on pg17. I don't think it is a PostgreSQL bug.

```
_localdev:main> create type User {
    create required property name: str;
    create index on (__subject__.name);
};
OK: CREATE TYPE
_localdev:main> with a := (for i in ({1, 2}, {1, 2}, {1, 2}, {1, 2}, {1, 2}, {1, 2}, {1, 2}, {1, 2}, {1, 2}, {1, 2}, {1, 2}, {1,
 2}, {1, 2}, {1, 2}, {1, 2}, {1, 2}, {1, 2}, {1, 2}, {1, 2}, {1, 2}) union (insert User { name := 'hello' })) select 1;
{1}
_localdev:main> select _set_config("enable_seqscan", "off");
{'off'}
_localdev:main> administer reindex (User);
OK: ADMINISTER
_localdev:main> analyze select User filter .name like 'E%';
─────────────────── Query ───────────────────
analyze select ➊  User filter .name like 'E%';

──────────── Coarse-grained Query Plan ────────────
       │ Time     Cost Loops Rows Width │ Relations
➊ root │ 20.0 13140.43   1.0  0.0    32 │ User
_localdev:main> \expand
────────────────── Fine-grained Query Plan ──────────────────
 │ Time     Cost Loops Rows Width │ Plan Info
 │ 20.0 13140.43   1.0  0.0    32 │ ➊ Gather
 │ 44.9 12140.33   3.0  0.0    32 │ SeqScan relation_name=User
```

This test case is resulting in:
- SeqScan on pg 17,
- IndexScan on pg 16,
- IndexScan if I explicitly set `text_pattern_ops` as the operator class of the index,

Current code will not specify the operator class when creating the index, which results in `text_ops` being used, both on pg 16 and pg 17 (I've verified).


<details>
<summary>
SQL for getting operator class of an index
</summary>

```sql
select pi.* from pg_index pi join pg_class pc on pc.oid = pi.indrelid where relname = 'hello';

select * from pg_opclass where oid = 3126;
```
</details>

Documentation on [operator classes](https://www.postgresql.org/docs/current/indexes-opclass.html) does imply that for pattern matching expressions, `XXX_pattern_ops` should be used instead of the default operator class.

So my conclusion is that PostgreSQL used to hit the index because it was somehow able to use `text_ops` for the LIKE operator. This behavior changes with PostgreSQL 17. We now have to decide if we want to:
- always use `text_pattern_ops` for `str` indexes, or
- be fine with SeqScan in this case.

I think that EdgeDB never guaranteed IndexScan with pattern matching expressions, so a change of behavior to SeqScan is not a breaking change.

aljazerzen:
I've decided that this change in index behavior is minor enough that I can merge this PR and fix it later.

I need to investigate further, because I'm now again suspecting that this was an unintentional change in PostgreSQL.

msullivan (approved):

PR #7825, opened 2024-10-03
Cleanup old EdgeDBMessage

aljazerzen:


aljazerzen:
Apparently log messages *are* a part of the protocol.

https://docs.edgedb.com/database/reference/protocol/messages#errorresponse

PR #7830, opened 2024-10-03
Server-side SCRAM should work with stored keys, not hashed keys

mmastrac:
In preparation for landing a new Rust front-end, implement the server-side SCRAM correctly by using stored/server keys like Postgres does (ie: `SCRAM-SHA-256$<iterations>:<salt>$<stored_key>:<server_key>`).

This code is not currently used.

### DIFF
--- a/edb/server/pgrust/src/auth/scram.rs
+++ b/edb/server/pgrust/src/auth/scram.rs
@@ -157,7 +157,6 @@ impl ServerTransaction {
                     &stored_key,
                 );
 
-                eprintln!("{calculated_stored_key:?} {stored_key:?}");
                 if calculated_stored_key.as_slice() != stored_key {
                     return Err(SCRAMError::ProtocolError);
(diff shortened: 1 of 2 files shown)


edb/server/pgrust/src/auth/scram.rs
```diff
@@ -137,23 +139,31 @@ impl ServerTransaction {
...
+                    &server_key,
+                    &stored_key,
                 );
-                let mut proof = vec![];
-                BASE64_STANDARD
-                    .decode_vec(message.proof.as_bytes(), &mut proof)
-                    .map_err(|_| SCRAMError::ProtocolError)?;
-                if proof != client_proof {
+
+                eprintln!("{calculated_stored_key:?} {stored_key:?}");
```
mmastrac, line 160:
```suggestion
```

fantix (approved):

PR #7831, opened 2024-10-04
Add `std::net::http::schedule_request` function

scotttrinh:


edb/lib/net.edgeql
```diff
@@ -71,12 +71,21 @@ ALTER TYPE std::net::http::Response {
 CREATE FUNCTION
 std::net::http::schedule_request(
     url: str,
-    named only body: optional std::bytes,
-    named only method: std::net::http::Method = std::net::http::Method.`GET`,
-    named only headers: optional std::array<std::tuple<name: std::str, value: std::str>>,
-) -> std::int16
+    named only body: optional std::bytes = <std::bytes>{},
+    named only method: optional std::net::http::Method = std::net::http::Method.`GET`,
```
fantix, line 75: I don't quite understand how an `optional method` worked inserting into a `required property method` 🤔

fantix (approved):
Looks good!

edb/lib/net.edgeql
scotttrinh: I assume having a default here means the inferred cardinality for this argument is `one`?

fantix:
Does `optional` allow passing in an empty set explicitly?


Scott Trinh ***@***.***>于2024年10月4日 周五下午1:58写道：

1st1 (commented):
Awesome.

How about we make it more awesome and you right some basic docs too? Could be very basic at this stage, but I think we should make it a rule from now on (I'll discuss on our Monday call)

1st1:
> Yeah, I need to add reference docs for the net module, so I'll do that, but I'll do it in a separate PR if that's alright.

Sure

edb/lib/net.edgeql
1st1: Is GET a keyword? I assume since it's a default people won't have to mess with it often

scotttrinh: Yeah, `GET` and `DELETE` are keywords and need to be escaped.

scotttrinh: We can add other functions for the common HTTP methods, if we want! Might make sense to use that to enforce, for instance, GET and DELETE not having `body` is practical usage, for example.

scotttrinh:
cc @1st1 
Super quick wip draft for the basic docs: https://github.com/edgedb/edgedb/pull/7833

Not ready, but wanted to link here for posterity.

edb/lib/net.edgeql
fantix: If I set `method := <std::net::http::Method>{}`, I get:

```
edgedb error: MissingRequiredError: missing value for required property 'method' of object type 'std::net::http::ScheduledRequest'
```

I think this is a compiler issue that allowed AT_MOST_ONE passed to ONE in functions 🤔

scotttrinh: Interesting! What is the type of a missing argument if not empty set 😅 . I didn't realize we had an undefined-like type.

I needed to make these are `optional` in order to not pass them, or maybe I messed up the function declaration somewhere? I'll play around with it to see if I can understand it better.

scotttrinh: OK, I think I figured it out: optional arguments need a default `{}` otherwise you get a "function does not exist" error. When I got that error before, I thought it was because any argument that has a default needed `optional` in order to be found if you don't pass that argument. I think I'm still a little surprised that `= {}` is the way you make an argument optional and `optional` is the way you affect the cardinality of the argument type itself, but now that I know it, I think I understand it a little better.

fantix (approved):

PR #7832, opened 2024-10-04
Install extensions unpacked into a directory

msullivan:


edb/load_ext/main.py
```diff
@@ -104,6 +104,32 @@ def get_pg_config(pg_config_path: pathlib.Path) -> dict[str, str]:
...
 
 
+def install_edgedb_extension(
+    pkg: pathlib.Path,
+    ext_dir: pathlib.Path,
+) -> None:
+    target = ext_dir / pkg.stem
+    print("Installing", target)
+
+    shutil.rmtree(target, ignore_errors=True)
```
elprans, line 114: I'm a bit uneasy about nuking first.  Let's install into a tempdir and then do a rename followed by rmtree.

msullivan: No `rmtree` needed then, right?

elprans: You still want to remove the old thing no?

```
unpack -> tmpdir
mv ext ext~
mv tmpdir ext
rm -rf ext~
```

edb/load_ext/main.py
```diff
@@ -123,12 +125,16 @@ def install_edgedb_extension(
...
+            ttarget_file = ttarget / f
 
             with z.open(f) as src:
-                with open(target_file, "wb") as dst:
+                with open(ttarget_file, "wb") as dst:
                     print("Installing", target_file)
                     shutil.copyfileobj(src, dst)
 
+        os.makedirs(ext_dir, exist_ok=True)
+        shutil.move(ttarget, ext_dir)
```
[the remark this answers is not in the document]
msullivan: I assumed that failing in that case was *why* you didn't like rmtree. Otherwise I don't really understand why we are making this change?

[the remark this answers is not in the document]
elprans: Failing if already exists is OK as long as we raise a slightly better error message (e.g. "extension foo is already installed").  The initial `rmtree` concern was that we were removing things permanently _before_ installing new stuff rather than _after_ once we've been satisfied that the extension was installed correctly.

elprans (approved):

msullivan:
I'll put the better error message in a follow-up, since I have some pending changes

PR #7834, opened 2024-10-04
Disable sql query cache when the environment variable is set also

msullivan:


edb/server/protocol/pg_ext.pyx
```diff
@@ -1493,6 +1493,10 @@ cdef class PgConnection(frontend.FrontendConnection):
             self.debug_print("Compile", query_str)
         fe_settings = dbv.current_fe_settings()
         key = (hashlib.sha1(query_str.encode("utf-8")).digest(), fe_settings)
+
+        # XXX: Slow?
```
elprans, line 1497: slow?

msullivan: The equivalent code paths for the edgeql protocol cache it in a cython member variable, and I felt bad about not doing that here.

I went and did it.

fantix (approved):

PR #7836, opened 2024-10-07
Search std for module name when using with x as module.

dnwpark:
- Allows queries such as `with http as net::http select http::Response`. The aliased module `net::http` will resolve to `std::net::http`.
- Apply module aliases to function names during normalization, even if no function was found.
    - This fixes an issue where given the following modules:
        - `module A { function foo() -> int64 using (1) }`
        - `module B {}`
        - `module C { alias query := (with A as B select A::foo()}`
    - `C::query` would normalize to `select A::foo()` with the expectation that `A::Foo` does not exist.

related #7737


msullivan (approved):
Great work on this. Tests look very exhaustive.

PR #7838, opened 2024-10-08
Server handshake implementation in Rust

mmastrac:
Implements the server-side of the Postgres handshake in Rust. This isn't currently hooked up, but we test it against the client-side handshake for correctness.

As client-side data is possibly malformed or malicious, we add a new validation layer in the protocol which automatically bails if any messages cannot be decoded as expected.  There's a new fuzz tester that attempts to exercise each message type with various mutations: deletions, randomization, etc.

We also add a new `PgError` for strongly-typed Postgres error codes.

SSL implementation on the server isn't tested currently. This will come in a future revision.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -342,6 +342,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "constant_time_eq"
+version = "0.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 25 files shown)


fantix (approved):
I skimmed through and this looks pretty well-done. We're reusing the same protocol/message framework for EdgeDB binary protocol too, right?

mmastrac:
> I skimmed through and this looks pretty well-done. We're reusing the same protocol/message framework for EdgeDB binary protocol too, right?

We will be, in a future PR. :)

Issue #7822, opened 2024-10-02
Warning system


msullivan:
We've decided that we need this for [RFC 1027 - Simplifying path resolution](https://github.com/edgedb/rfcs/blob/master/text/1027-no-factoring.rst).

Initial plan is that it will be encoded in JSON as a header for CommandDataDescription. Users of the clients should be able to provide a callback that is invoked on the list of warnings; the default will be to log them.

TODO:

- [x] Server implementation: #7823
- [x] Rust binding and CLI update
- [x] Python binding update: https://github.com/edgedb/edgedb-python/pull/523
- [x] JS binding update
- [x] Go binding update
- [ ] Expose warning count metrics?

msullivan:
Thinking about the protocol a bit more, it's clearly unfortunate that we are encoding the warnings fully as JSON, since it means clients need two different decoders for errors.

Headers are defined as being JSON, though (and *definitely* need to be utf-8 strings). Maybe the best approach (though it is kind of ugly) would be to have the JSON be a list of base64-encoded exceptions in the normal wire format?

1st1:
Imo let's stop where we are. JSON is ok. We should work on a proper system
(using our type descriptors) for 7.x.

jollygreenlaser:
In the future, will it be possible for us to add our own warnings to the system? This could be a very powerful debug/diagnostic tool as an outright feature of edgedb.

Can we get a warning flag for a `=` that hits a `{}` and otherwise silently fails, aka should've been `?=`?

aljazerzen:
Apparently, `start`, `end`, `line` and `col` are all strings that contain the integers. Do we want to change that?

msullivan:
> Apparently, `start`, `end`, `line` and `col` are all strings that contain the integers. Do we want to change that?

Hm... The plus side for doing so is that it would be easy and would make more sense.
The minus side is that attributes in the binary encoding are all strings, I think, so the current thing is more consistent between encodings.

I think probably we should change it?

msullivan:
> In the future, will it be possible for us to add our own warnings to the system? This could be a very powerful debug/diagnostic tool as an outright feature of edgedb.
> 

Possibly. Right now, though, all the warnings are *static*, produced by the compiler. Being able to support raising a warning dynamically (which is what a user-defined error would need) would have some runtime overhead. (Probably it would require inserting the warning into a temporary table, and then querying the temporary table after the query.)

[...]

PR #7842, opened 2024-10-09
Encode offset positions as ints in json error encoding

msullivan:
This is a tweak to the warning protocol. See #7822.
Some of the pending bindings changes may need to be updated.
edgedb-python shouldn't need to be. (Cast to `int` becomes a no-op.)

### DIFF
--- a/edb/errors/base.py
+++ b/edb/errors/base.py
@@ -128,7 +128,10 @@ class EdgeDBError(Exception, metaclass=EdgeDBErrorMeta):
         }
         for name, field in _JSON_FIELDS.items():
             if field in self._attrs:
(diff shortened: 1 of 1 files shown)


msullivan:
I'm not 100% sure I stand by this change though?
Looking at the edgedb-python again, it expects to store all its attributes as strings (or bytes!), since that is what we get on the binary protocol...

fmoor (approved):

aljazerzen:
Usually I'm for consistency across data structures and would agree that we should have offsets as strings everywhere. But having strings contain decimal-encoded integers is a large red flag for me and I think it outweighs the consistency perspective.

In the end, this is all more or less subjective and does not have large impact.

Issue #7330, opened 2024-05-10
Database configure not always properly applying immediately on current connection


msullivan:
`test_server_proto_configure_error` is flaking occasionally: https://github.com/edgedb/edgedb/actions/runs/9007197736/job/24746488573

I would think it could *also* be cache related, but I think I saw the reverse happen too: the error having *not* been cleared during teardown.

I'm probably going to put a flake workaround

msullivan:
Here's one where it isn't getting cleared right: https://github.com/edgedb/edgedb/actions/runs/9025625652/job/24801671577?pr=7327

msullivan:
I think it is a race with the scheduled reintrospection

PR #7844, opened 2024-10-09
Try to fix "danger" flakes

msullivan:
We've had test flakes where `test_server_config.py` tests other than
`test_server_proto_configure_error` were failing with "danger" errors
set by `test_server_proto_configure_error`.

`test_server_proto_configure_error` does wait to try to make sure that
the reset of `force_database_error` has made it to other connections,
by checking that other connections can run a query. The problem is
that it might be that the reason it can run a query is because the
*error* config hasn't propagated to it yet.

Fix this by making sure the final error we configure makes it to the
other connection.

### DIFF
--- a/tests/test_server_config.py
+++ b/tests/test_server_config.py
@@ -1353,15 +1353,26 @@ class TestServerConfig(tb.QueryTestCase):
 
             err = {
                 'type': 'SchemaError',
-                'message': 'danger',
+                'message': 'danger 1',
                 'context': {'start': 42},
             }
+
+            # N.B: When CONFIGURE CURRENT DATABASE is run, the new
+            # setting becomes live immediately on the current
+            # connection, but does not become live on other
+            # connections until after a scheduled background
+            # reintrospection is done.
+            #
+            # There is a bug related to this (#7330) in which an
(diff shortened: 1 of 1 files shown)


tests/test_server_config.py
```diff
@@ -1353,15 +1353,26 @@ async def test_server_proto_configure_error(self):
...
             }
+
+            # N.B: When CONFIGURE CURRENT DATABASE is run, the new
+            # setting becomes live immediately on the current
+            # connection, but does not become live on other
+            # connections until after a scheduled background
+            # reintrospection is done.
+            #
+            # There is a bug related to this (#7330) in which an
+            # *older* scheduled reintrospection can temporarily
```
fantix, line 1367: Ugh, why introspection lock (#7485) didn't fix #7330 again? Cache-related issues?

fantix (approved):

tests/test_server_config.py
msullivan: The introspection lock makes sure the introspections don't get reordered, but doesn't prevent an older introspection from overwriting a newer "session level" application of a database config.

PR #7845, opened 2024-10-10
Fix inplace upgrade tests after pg_addr changes

msullivan:
PR #7605 broke the inplace-upgrade tests, since it removed host and
port from the `pg_addr` field on the `server-info` endpoint and turned
the `server_settings` field into the toplevel.

Fix it by exposing the full DSN to server-info as well.

(An earlier iteration made the inplace upgrade tests use the testbase
python connection to read pgdsn from settings.)

### DIFF
--- a/edb/server/tenant.py
+++ b/edb/server/tenant.py
@@ -1797,6 +1797,12 @@ class Tenant(ha_base.ClusterProtocol):
         self.create_task(task(), interruptable=True)
 
     def get_debug_info(self) -> dict[str, Any]:
+        pgaddr = self.get_pgaddr()
+        pgaddr.clear_server_settings()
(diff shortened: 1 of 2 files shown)


msullivan:
The inplace upgrade tests run scheduled daily, since they are pretty slow. See https://github.com/edgedb/edgedb/actions/runs/11247508356 for a failure

mmastrac (approved):
Ahh JQ fun. LGTM.

fantix (approved):

msullivan:
> Ahh JQ fun. LGTM.

Yeah, busting out `jq` is always when I start thinking "... maybe I should have written this in Python"

PR #7851, opened 2024-10-12
Fix inplace upgrade rollbacks

msullivan:
A new parameter got added to PGConnectionProxy, and it swalloed one
call site's dbname argument.  Fix that, and make the params
keyword-only.


mmastrac (approved):
LGTM

PR #7860, opened 2024-10-15
Fix COPY that uses globals over SQL adapter

aljazerzen:


msullivan:
Should COPY just not respect policies?

msullivan:
Sigh, idk.

jackfischer:
My user feedback is that policies don't matter over the SQL adapter. SQL adapter is really for BI and ETL. Policies in gel-protocol SQL are desirable of course.

aljazerzen:
@jackfischer That's very close to how I think about it. But still, COPY should eventually support globals/permissions.

And sql-over-gel-proto cannot support COPY, becuase gel-proto does not have messages for copy output.

This PR now just improves error messages (I've repurposed an old PR that only added a test).

PR #7865, opened 2024-10-15
Make CONFIGURE CURRENT DATABASE block until configuration is actually set

### LINKED ISSUE #7330: Database configure not always properly applying immediately on current connection

msullivan:
... on the local instance, at least. Multi-frontend will still be racy.

Fixes #7330.

### DIFF
--- a/edb/server/tenant.py
+++ b/edb/server/tenant.py
@@ -1154,6 +1154,7 @@ class Tenant(ha_base.ClusterProtocol):
         cache_mode = config.QueryCacheMode.effective(
             db.lookup_config('query_cache_mode')
         )
+        # TODO: don't always hydrate on reintrospection
         if query_cache and cache_mode is not config.QueryCacheMode.InMemory:
             db.hydrate_cache(query_cache)
         return old_cache_mode is not cache_mode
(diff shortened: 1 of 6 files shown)


msullivan:
I think the background introspections were the cause of the flakes disabled by https://github.com/edgedb/edgedb/pull/7859.
I think they caused the schema to go backwards temporarily.

edb/server/tenant.py
```diff
@@ -1656,29 +1657,18 @@ async def task():
...
-                            b'SELECT edgedb._clear_query_cache()')
-                        self._dbindex.get_db(dbname).clear_query_cache()
-            except Exception:
-                metrics.background_errors.inc(
-                    1.0, self._instance_name, "on_local_database_config_change"
-                )
-                raise
-
-        self.create_task(task(), interruptable=True)
+        # TODO: Can we just do config?
```
fantix, line 1663: I think we do need the user schema for extension config specs.

edb/server/tenant.py
```diff
@@ -1656,29 +1657,18 @@ async def task():
...
+        if await self.introspect_db(dbname):
```
fantix, line 1664: self-note: `introspect_db()` has a lock inside, so concurrent database config changes from different frontends will be properly (?) serialized.

fantix (approved):
Yeah I think this is good, maybe @1st1 wanna have a look too?

fantix (approved):

PR #7866, opened 2024-10-15
Move new std::net backend to Rust

mmastrac:
Migrate std::net to a new Rust library that provides access to `reqwest` to Python code.

Enhances the std::net tests to fully test the output of the network request, as well as adding an error-case test.

### DIFF
--- a/edb/server/http/src/python.rs
+++ b/edb/server/http/src/python.rs
@@ -1,0 +1,400 @@
+use futures::future::poll_fn;
+use pyo3::{exceptions::PyException, prelude::*, types::PyByteArray};
+use reqwest::Method;
+use scopeguard::defer;
+use serde_pickle::SerOptions;
+use std::{
+    cell::RefCell, collections::HashMap, os::fd::IntoRawFd, pin::Pin, rc::Rc, sync::Mutex, thread,
+};
+use tokio::{
+    io::AsyncWrite,
+    sync::{AcquireError, Semaphore, SemaphorePermit},
+    task::LocalSet,
+};
+use tracing::{error, info, trace};
+
+pyo3::create_exception!(_http, InternalError, PyException);
+
+type PythonConnId = u64;
+
+#[derive(Debug)]
+enum RustToPythonMessage {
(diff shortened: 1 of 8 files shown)


edb/server/http/src/python.rs
```diff
@@ -0,0 +1,434 @@
...
+        let mut counts = self.counts.lock().unwrap();
+        let old_capacity = counts.capacity;
+        counts.capacity = new_limit;
+
+        if new_limit > old_capacity {
+            let added = new_limit - old_capacity;
+            self.semaphore.add_permits(added);
+        } else if new_limit < old_capacity {
+            let removed = old_capacity - new_limit;
+            self.semaphore.forget_permits(removed);
```
msullivan, line 188: Add a comment that this might forget too few permits, but we make up for it in release?

msullivan: Oh, I think there might actually be a race here.
Imagine that `capacity` is 10 and `active` is 6, so there are 4 permits remaining.
We call `update_limit(5)`, and try to forget `5` permits, but only `4` actually get forgotten.
Now capacity is 5, active is 6, and there are 0 permits remaining.
If `update_limit(10)` gets called now, we'll call `add_permits(5)` and update `capacity` to 10.
Then, when the 6 active threads release, they will all return their permits (since they are no longer over the limit), and we will wind up with 11 permits in the semaphore.

scotttrinh (approved):
Additions to the tests look great, thanks for that! Code looks good to me. 🏆

edb/server/http/src/python.rs
mmastrac: Ack. I think you are correct. Let me add a test for this case.

msullivan:
Finding concurrency bugs is genuinely one of my greatest joys.

edb/server/http/src/python.rs
mmastrac: Added a bunch more validation, and some extensive tests around acquire/release that run only in the `debug_assertions` cfg.

PR #7869, opened 2024-10-16
Add query performance statistics

fantix:
This is the 2nd take after #7814, forking the builtin pg_stat_statement extension from the upstream master branch. It's different in a way that we can extract JSON query info only once per query across parse/plan/execute runs (unless reset cleared the hashtable row), and some custom stats columns are directly stored as a hashtable column.

Please review each commit separately.

Refs #7725

- [x] Add compiler config columns
- [ ] Add config to turn off stats


edb/server/compiler/sql.py
```diff
@@ -248,8 +257,26 @@ def compile_sql(
...
-        unit.stmt_name = compute_stmt_name(unit.query, tx_state).encode("utf-8")
+            track_stats = True
+
+        stmt_hash = hash_stmt_name(unit.query, tx_state)
+        unit.stmt_name = compute_stmt_name(stmt_hash).encode("utf-8")
+
+        if track_stats and backend_runtime_params.has_stat_statements:
+            cache_key = uuidgen.from_bytes(stmt_hash.digest()[:16])
+            sql_debug_obj = {
+                'query': orig_text,
```
fantix, line 268: Normalize?

edb_stat_statements/edb_stat_statements.c
```diff
@@ -468,6 +525,133 @@ _PG_init(void)
...
+		JsonLexContext *lex = makeJsonLexContextCstringLen(
+#if PG_VERSION_NUM >= 170000
+				NULL,
+				info_str,
+#else
+				(char *) info_str,  // not actually mutating
+#endif
+				info_len,
+				PG_UTF8,
+				true);
```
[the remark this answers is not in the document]
msullivan: Oh, and if we can't rely on that I think we also would need to be worrying about the strings we get passed in the callbacks, which are also `palloc`d?

msullivan (approved):
OK this all looks good except I realize I might not understand the memory management model in play here.

edb_stat_statements/edb_stat_statements.c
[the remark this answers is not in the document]
fantix: Yeah, I believe it's still safer to manage memory manually; now it's fixed.

edb/lib/sys.edgeql
```diff
@@ -86,6 +90,43 @@ ALTER TYPE sys::Role {
...
+
+    CREATE PROPERTY calls -> std::int64;
+    CREATE PROPERTY total_exec_time -> std::duration;
+    CREATE PROPERTY min_exec_time -> std::duration;
+    CREATE PROPERTY max_exec_time -> std::duration;
+    CREATE PROPERTY mean_exec_time -> std::duration;
+    CREATE PROPERTY stddev_exec_time -> std::duration;
+
+    CREATE PROPERTY rows -> std::int64;
+    CREATE PROPERTY stats_since -> std::datetime;
```
1st1, line 114: @elprans Elvis, can you add `description` annos to every field? Just one sentence

fantix: Added in [this crf commit](https://github.com/edgedb/edgedb/pull/7869/commits/eb453cf0772b70cfd894086a8096117e7e305209)

1st1:
As discussed on the call, we need to make a few adjustments.

* As @elprans mentioned, we want to exclude migration hash from the query id, so that schema migrations don't "reset" the stats. Most of schema migrations are relatively minor for real world application, affecting only a subset of queries anyway.

* We want to add query "tagging". The motivation for that is to make it easier for users to browse through the query log and identify which queries are coming from EdgeDB extensions like Auth, or from EdgeDB UI (those can be quite slow which could be confusing), or from various parts of the user application.

The plan for tagging is as follows:

- [ ] Add `tag: string` property to `sys::QueryStats`.
- [ ] We should use "headers" in our protocol to apply tags on query by query basis. Server should pass the "tag" to the compiler, which should embed it into a SQL query, which the extension can later parse out.
- [ ] Our client libraries will likely get `withTag()` method, so users will be able to run `await client.withTag('myapp').query(...)`

Tags should be documented as arbitrary strings. We can reserve `"edgedb/"` "namespace" for EdgeDB-related things, e.g. `"edgedb/ui"`, `"edgedb/auth"`, etc.

I think we should put a max-length constraint on this field and limit tags to 200-or-something characters.

cc @elprans @msullivan and obviously @fantix

msullivan:
I think Elvis wanted to inject the tag on the server side, to avoid compiler round trips.

edb_stat_statements/edb_stat_statements.c
```diff
@@ -608,14 +603,31 @@ edbss_extract_stmt_info(const char* query_str, int query_len) {
...
+#else
+					(char *) info_str,  // not actually mutating
+#endif
+					info_len,
+					PG_UTF8,
+					true);
+			JsonParseErrorType parse_rv = pg_parse_json(lex, &sem);
+			freeJsonLexContext(lex);
+
+			if (parse_rv == JSON_SUCCESS)
```
msullivan, line 629: Does the state get mutated even if there is an error? I guess that's probably fine?

What happens to `info_len` on an error? Is it still updated with however much got consumed?

edb_stat_statements/edb_stat_statements.c
```diff
@@ -608,14 +603,31 @@ edbss_extract_stmt_info(const char* query_str, int query_len) {
...
+				if ((state.found & EDB_STMT_INFO_PARSE_REQUIRED) == EDB_STMT_INFO_PARSE_REQUIRED)
+					return info->query_id != UINT64CONST(0) ? info : NULL;
+
+			info_str += info_len + 1;
```
msullivan, line 633: The +1 makes me nervous about the case where there isn't a newline at the end of an info line?

msullivan: I'm not sure if the cases where there are untrusted entries in the query log is that likely, but this is C so I want to be extra careful about our boundary cases.

edb_stat_statements/edb_stat_statements.c
```diff
@@ -699,22 +696,12 @@ edbss_json_scalar(void *semstate, char *token, JsonTokenType tokenType) {
...
-			}
-			goto fail;
-
-		case EDB_STMT_INFO_PARSE_CACHE_KEY:
+		case EDB_STMT_INFO_PARSE_ID:
 			if (tokenType == JSON_TOKEN_STRING) {
-				Datum cache_key = DirectFunctionCall1(uuid_in, CStringGetDatum(token));
-				state->info->cache_key = DatumGetUUIDP(cache_key);
+				Datum id_datum = DirectFunctionCall1(uuid_in, CStringGetDatum(token));
+				pg_uuid_t *id_ptr = DatumGetUUIDP(id_datum);
```
msullivan, line 712: Can these fail?

msullivan (commented):
This looks basically good but I want to make sure I understand the behavior in some of the edge cases where the input might be malformed.

In particular, what happens if uuid_in fails because the `id` is malformed?

(To some extent, this is just me not knowing postgres internals well. It might be that the answer is simple? Does failure do a longjmp or something and just abort everything?)

edb_stat_statements/edb_stat_statements.c
fantix: > Does the state get mutated even if there is an error? I guess that's probably fine?

Yes, it'll keep the parsed values and continue to the next line, which I think is fine.

> What happens to `info_len` on an error? Is it still updated with however much got consumed?

On error, `info_len` will be updated to skip the whole failing line and restart on the next line.

edb_stat_statements/edb_stat_statements.c
fantix: Right, good question. The `edbss_extract_info_line()` function will tag the `\n` or the end of the `query_str`, so this `+1` will either skip the `\n` properly, or go beyond the end of the `query_str` and cause the next call to `edbss_extract_info_line()` to return `NULL`.

edb_stat_statements/edb_stat_statements.c
fantix: Yes. The outer-most `PG_CATCH` in the current session will - like you said - do a longjmp and recover by sending an error to the peer. Such error will be propagated to the client as an `edb.errors.InvalidValueError`:

```
ERROR 116169 - 2024-11-20T11:22:27.377 postgres: invalid input syntax for type uuid: "b2f8e457-a4f8-ab73-1979-afb333f9c"
INFO 116169 - 2024-11-20T11:22:27.377 postgres: -- {"query": "select\n    (<__std__::int64>$0 + <__std__::int64>$1)", "type": 1, "extras": "{\"cc\": {\"__internal_no_apply_query_rewrites\": false, \"__internal_query_reflschema\": false, \"__internal_testmode\": false, \"allow_bare_ddl\": \"AlwaysAllow\", \"allow_dml_in_functions\": false, \"allow_user_specified_id\": false, \"apply_access_policies\": true, \"force_database_error\": \"false\", \"query_cache_mode\": \"Default\", \"simple_scoping\": null, \"store_migration_sdl\": \"NeverStore\", \"warn_old_scoping\": null}, \"pv\": [3, 0], \"of\": \"BINARY\", \"e1\": false, \"il\": 101, \"ii\": false, \"in\": true, \"io\": false, \"dn\": \"default\"}", "id": "b2f8e457-a4f8-ab73-1979-afb333f9c"}
INFO 116169 - 2024-11-20T11:22:27.377 postgres: 	SELECT edgedb_v6_2f20a50ab0.__qh_bd20a1eba9bb696335db87182e5b207f(($1)::int8, ($2)::int8)
---------------------------------------------------------------------- Exception occurred: invalid input syntax for type std::uuid: "b2f8e457-a4f8-ab73-1979-afb333f9c" ----------------------------------------------------------------------

1. edb.errors.InvalidValueError: invalid input syntax for type std::uuid: "b2f8e457-a4f8-ab73-1979-afb333f9c"

------------------------------------------------------------------------------------------------------------------ Details -------------------------------------------------------------------------------------------------------------------


edb.errors.InvalidValueError: invalid input syntax for type std::uuid: "b2f8e457-a4f8-ab73-1979-afb333f9c"
ERROR 116111 _localdev 2024-11-20T11:22:27.377 asyncio: an error in edgedb protocol
protocol: <edb.server.protocol.binary.EdgeConnection object at 0x73c6cfabf370>
transport: <uvloop.loop._SSLProtocolTransport object at 0x73c6d48086c0>
```

edb_stat_statements/edb_stat_statements.c
msullivan: Because len is negative, at that point?

msullivan (approved):

msullivan:
If you merge this now, please update https://github.com/edgedb/edgedb/issues/7725 with the remaining pending tasks so we can track them

edb_stat_statements/edb_stat_statements.c
fantix: Yes exactly

PR #7875, opened 2024-10-16
Test cached binaries and use /usr/lib contents as cache key

mmastrac:


.github/workflows.src/tests.inc.yml
```diff
@@ -432,12 +432,7 @@
 
         # Validate Rust extensions
         for ext in ./build/rust_extensions/edb/*.so; do
-          ldd "$ext" || exit 1
-        done
-
```
msullivan: Why?

msullivan (approved):

.github/workflows.src/tests.inc.yml
mmastrac: I thought I'd add a test for all the stuff we're restoring from cache -- it seems unlikely that the extensions will break but I suppose it's possible if there's some weirdness going on with openssl libraries moving around, and it would be pretty annoying to track down if it did.

I can take that out.

msullivan: Oh I had just noticed the removal, not the addition, in the "[Skip validation of cython extensions](https://github.com/edgedb/edgedb/pull/7875/commits/505779b1674b86fe96e809a083cc003c21be43a9)" commit

PR #7883, opened 2024-10-17
Migrate OAuth HTTP code to new worker

mmastrac:
This migrates the auth extensions from httpx to the new Rust HTTP handler. We add a new, shared HttpClient to the tenant which anyone can make use of.

### DIFF
--- a/edb/server/http.py
+++ b/edb/server/http.py
@@ -1,0 +1,191 @@
+#
+# This source file is part of the EdgeDB open source project.
+#
+# Copyright 2024-present MagicStack Inc. and the EdgeDB authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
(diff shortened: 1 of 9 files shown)


edb/server/http.py
```diff
@@ -0,0 +1,191 @@
...
+        if args:
+            data = args[0]
+            if isinstance(data, Mapping):
+                for key, value in data.items():
+                    self[key] = value
+            else:
+                for key, value in data:
+                    self[key] = value
+        for key, value in kwargs.items():
+            self[key] = value
```
msullivan, line 171: Is it really worth doing this?

tests/test_http_ext_auth.py
```diff
@@ -1417,8 +1417,8 @@ async def test_http_auth_ext_azure_authorize_01(self):
 
             discovery_request = (
                 "GET",
-                "https://login.microsoftonline.com/common/v2.0",
-                ".well-known/openid-configuration",
+                "https://login.microsoftonline.com",
+                "common/v2.0/.well-known/openid-configuration",
```
msullivan, line 1421: Why do these need to change?

msullivan (approved):

tests/test_http_ext_auth.py
mmastrac: This was relying on some very obscure parsing and encoding for base URLs. In the end I just cleaned up the URL handling so we didn't have to special case the base URL part of things.

edb/server/http.py
mmastrac: The getter is definitely necessary to make sure the older code works, but the setter is more for ensuring we don't hit any landmines with code expecting headers to work as it does in httpx. I can take another pass at this once I finish up the AI code and figure out exactly what level of compat is necessary.

PR #7886, opened 2024-10-18
fix sql introspection

aljazerzen:
- **minor refactor**
- **fix a flake in sql tests**
- **postgres-version-aware sql introspection**


msullivan (approved):
Looks good!

msullivan (changes requested):
Actually we have some more discussion needed before we go with this route

msullivan (approved):
Oh actually as you pointed on on Slack, this is fine, since we make these views after the tpldbdump

PR #7887, opened 2024-10-18
Use tpldbdump cache only with latest PostgreSQL version

aljazerzen:
pg_dump scripts cannot be (reliably) restored into databases of older PostgreSQL versions.

Because of that, we have failures when using edgedb with non-latest postgres.

This PR does not use tpldbdump cache only when backend version is latest (17).

### DIFF
--- a/edb/server/bootstrap.py
+++ b/edb/server/bootstrap.py
@@ -1556,12 +1556,17 @@ async def _init_stdlib(
         src_hash=src_hash,
         cache_dir=cache_dir,
     )
-    tpldbdump_package = read_data_cache(
-        tpldbdump_cache,
-        pickled=True,
-        src_hash=src_hash,
-        cache_dir=cache_dir,
-    )
+
+    backend_params = cluster.get_runtime_params()
+
+    tpldbdump_package = None
+    if backend_params.instance_params.version.major == 17:
+        tpldbdump_package = read_data_cache(
+            tpldbdump_cache,
+            pickled=True,
+            src_hash=src_hash,
+            cache_dir=cache_dir,
+        )
     if args.inplace_upgrade_prepare:
         tpldbdump = None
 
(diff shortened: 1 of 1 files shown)


edb/server/bootstrap.py
```diff
@@ -1556,12 +1556,17 @@ async def _init_stdlib(
...
         cache_dir=cache_dir,
     )
-    tpldbdump_package = read_data_cache(
-        tpldbdump_cache,
-        pickled=True,
-        src_hash=src_hash,
-        cache_dir=cache_dir,
-    )
+
+    backend_params = cluster.get_runtime_params()
```
aljazerzen, line 1560: Is it ok to execute this piece of code earlier than it was before?

edb/server/bootstrap.py
```diff
@@ -1556,12 +1556,17 @@ async def _init_stdlib(
...
+
+    tpldbdump_package = None
+    if backend_params.instance_params.version.major == 17:
```
aljazerzen, line 1563: Do we already keep a set of supported PostgreSQL versions somewhere?

This number should definetely be a constant somewhere, but i don't know where.

msullivan: I don't think we do, but it should probably be in `defines.py` if anywhere?

edb/server/bootstrap.py
msullivan: Hm..... I think it might actually need to be after `populate_misc_instance_data` to work right?
Let me track it down

msullivan: OK, nevermind, I think it's fine

msullivan: I was thinking of the `edgedb.get_backend_capabilities()` *postgres* function

msullivan (approved):

msullivan (changes requested):
Elvis thinks the perf hit here for old versions is probably too bad to do; I'm investigating whether we can get away with using old tpldbdumps still

msullivan:
OK, yeah, I think we don't want to do this and it was a bad idea for me to suggest it. The things that have the most version dependency currently aren't part of the dump anyway.
We should be able to keep hacking it into working by doing some surgery on the tpldbdump script.

PR #7889, opened 2024-10-18
Fix RETURNING clause rel var name resolution over SQL adapter

### LINKED ISSUE #7890: RETURNING clause in SQL queries doesn't resolve table names
- EdgeDB Version: EdgeDB 6.0-dev.8902+local

Steps to Reproduce:

See #7889 for tests reproducing the bug.

Basically referring to the tables from the `INSERT`, `UPDATE`, or `DELETE` query fails in the `RETURNING` clause.
```
edgedb=# insert into "User" (name) values('Dana') returning id;
                  id                  
--------------------------------------
 29ce6dd6-8d7c-11ef-a351-5781971bf2b4
(1 row)

INSERT 0 1
```
vs
```
edgedb=# insert into "User" (name) values('Dana') returning "User".id;
ERROR:  cannot find table `User`
LINE 1: ...nsert into "User" (name) values('Dana') returning "User".id;
                                                             ^
```

vpetrovykh:
Closes #7890

RETURNING clauses don't resolve table names.

### DIFF
--- a/tests/test_sql_dml.py
+++ b/tests/test_sql_dml.py
@@ -967,6 +967,23 @@ class TestSQLDataModificationLanguage(tb.SQLQueryTestCase):
         )
         self.assertEqual(res, [[doc_id, user_id]])
 
+    @test.xfail('''
+        RETURNING clause doesn't support table names
+
+        ERROR:  cannot find table `Document`
+    ''')
+    async def test_sql_dml_insert_44(self):
+        # Test that RETURNING supports "Table".col format
+        res = await self.squery_values(
+            '''
+            INSERT INTO "Document" (title) VALUES ('Test returning')
+            RETURNING "Document".id
+            '''
+        )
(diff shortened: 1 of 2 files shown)


tests/test_sql_dml.py
```diff
@@ -967,6 +967,23 @@ async def test_sql_dml_insert_43(self):
         )
         self.assertEqual(res, [[doc_id, user_id]])
 
+    @test.xfail('''
```
msullivan, line 970: Needs to be `xerror` I think

msullivan (approved):

msullivan:
Please file an issue to track these xfailed tests you are adding, also. (You can just reference the tracking issue in the PR commits and that will build a good enough list)

aljazerzen:
@vpetrovykh, sorry for taking over this PR, it was an easy fix.

PR #7892, opened 2024-10-18
Fix older postgres versions

msullivan:
This still *leaves out* fields that were removed in postgres 17,
though.

They seem pretty marginal, but it's worth discussing what we really
want to do there. I think this approach is good enough to ship an
alpha with, though.


msullivan:
My theory is that we can land this now and then discuss next week whether we want to land https://github.com/edgedb/edgedb/pull/7886 or whether we should try to only expose 13.9 stuff or whether we should land it and then also try to start claiming the real underlying versions?

aljazerzen (approved):

aljazerzen:
https://github.com/edgedb/edgedb/actions/workflows/tests-pg-versions.yml?query=branch%3Apg-17-wtf

aljazerzen:
This is the last remaining error, but I'm not sure it is related to this PR:

```
RuntimeError: exception during initialization of 'dumpv5' test DB: UnsupportedBackendFeatureError(could not find extension satisfying vector >=0.7.4,<0.8.0: only found versions 0.6.2)
```

PR #7896, opened 2024-10-21
Increase timeout in inplace upgrade test

msullivan:


msullivan:
The test failures are flakes that seem unrelated to inplace upgrades

PR #7897, opened 2024-10-21
Initialize `CompilationRequest` in one step

elprans:
The current two-step initialization of `CompilationRequest` seems
unnecessary and error-prone.  Have a normal constructor and make
`deserialize` a classmethod instead.


fantix (approved):
LGTM

PR #7899, opened 2024-10-22
Fix error when dropping non overloaded function.

dnwpark:
Related #7692


aljazerzen (approved):

aljazerzen:
The test failure is a flake.

msullivan:
A weird flake, too

PR #7904, opened 2024-10-22
ext: fix issues in net gc task

fantix:
* don't run on dropping databases
* don't call asyncio.wait([])
* fix wrong logging without decoding JSON

Refs #7888


scotttrinh (approved):
🙏

edb/server/net_worker.py
```diff
@@ -382,7 +387,8 @@ async def gc(server: edbserver.BaseServer) -> None:
             if tenant.accept_new_tasks
         ]
         try:
-            await asyncio.wait(tasks)
+            if tasks:
+                await asyncio.wait(tasks)
```
elprans, line 391: Why aren't we using a `TaskGroup` here?

edb/server/net_worker.py
```diff
@@ -363,6 +364,10 @@ async def _gc(tenant: edbtenant.Tenant, expires_in: statypes.Duration) -> None:
             for db in tenant.iter_dbs():
                 if db.name == defines.EDGEDB_SYSTEM_DB:
                     continue
+                if not tenant.is_database_connectable(db.name):
+                    # Don't run the net_worker if the database is not
+                    # connectable, e.g. being dropped
+                    continue
```
msullivan, line 370: Should we be checking in `_delete_requests` also?

(More generally, in the medium term I think we need to figure out design patterns/interfaces for worker tasks that get these things right by default)

edb/server/net_worker.py
```diff
@@ -351,6 +351,7 @@ def _warn(e):
                 variables={"expires_in": expires_in.to_backend_str()},
                 cached_globally=True,
             )
+            result: list[dict] = json.loads(result_json)
             if len(result) > 0:
                 logger.info(f"Deleted requests: {result!r}")
```
elprans, line 356: I don't think it's a good idea to log all requests here probably. The list could be considerably large.  Perhaps just log the count?

edb/server/net_worker.py
fantix: Yeah, I guess we can; this was a carried-over from [an old commit](https://github.com/edgedb/edgedb/pull/6233/commits/47c3c09973adf08c20c7917b7a836ad26ddaa387), let me try again so that we know for sure.

msullivan:
... looks like this doesn't fix that DROP flake though

edb/server/net_worker.py
fantix: Ah it's just unnecessary, because the Task we're in here is safe without cancellation - the server doesn't need to wait for `gc()` itself to clean up anything, while all child Tasks are already managed by the `TaskGroup` in each Tenant (by `tenant.create_task()`).

We would need an extra coroutine wrapper to use `TaskGroup` here:

```python
async def _await_task(task: asyncio.Task):
    return await task


async def gc(server: edbserver.BaseServer) -> None:
    while True:
        try:
            async with asyncio.TaskGroup() as g:
                for tenant in server.iter_tenants():
                    if tenant.accept_new_tasks:
                        task = tenant.create_task(
                            _gc(tenant, NET_HTTP_REQUEST_TTL),
                            interruptable=False,
                        )
                        g.create_task(_await_task(task))
...
```

edb/server/net_worker.py
fantix: Fixed for now - let's see if it fixes the test

edb/server/net_worker.py
fantix: Changed to log the count

edb/server/net_worker.py
fantix: Nope, still failing.

PR #7907, opened 2024-10-22
Try to fix some of the DROP DATABASE flakes

msullivan:
When in dev/test mode, dump out all the info about the other
connections.  Hopefully this will let us figure out where they are
coming from.

See also #7904


fantix (approved):

msullivan:
There are still other server test race conditions I'm able to reproduce locally now, though, that I'm still tracking

PR #7914, opened 2024-10-23
Rename net_http_max_connections to http_max_connections

msullivan:
We're going to use it to cap other outgoing http connections also.


mmastrac (approved):
LGTM

PR #7917, opened 2024-10-23
Some more changes to reduce adjacent server flakes

msullivan:
1. Make worker threads (like net) not attempt new connections
   once the server has started shutting down.
2. Don't respond to messages that prompt updating database
   metadata for database's we don't have introspected anyway.
3. Mark a lot of worker tasks as being interruptable. When they
   weren't interruptable, server exit had to wait for them to
   exit, which they didn't always do. This caused connections
   to get held open.

*Most* of the locally observable flakes I've been seeing were fixed by
these changes, though I am still getting
test_server_adjacent_database_propagation failures where it fails to
connect to the newly create database. Still working on that.

### DIFF
--- a/edb/server/dbview/dbview.pyi
+++ b/edb/server/dbview/dbview.pyi
@@ -118,6 +118,9 @@ class Database:
     def lookup_config(self, name: str) -> Any:
         ...
 
+    def is_introspected(self) -> bool:
+        ...
+
 class DatabaseConnectionView:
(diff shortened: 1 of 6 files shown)


msullivan:
I have never been happier in my life to have had a lint error. When I saw one test suite failed I assumed it was the flake again.

elprans (approved):

fantix (approved):

PR #7920, opened 2024-10-24
Avoid computing globals json if not needed for inlined function.

dnwpark:
related #7692

### DIFF
--- a/edb/edgeql/compiler/func.py
+++ b/edb/edgeql/compiler/func.py
@@ -542,6 +542,28 @@ class ArgumentInliner(ast.NodeTransformer):
 
         return cast(irast.Base, self.generic_visit(node))
 
+    # Don't transform pointer refs.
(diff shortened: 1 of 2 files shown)


msullivan:
This is only sort of what I was looking for. I think that even if the inlined function uses global, we should use them directly instead of building and using a globals json.
This will probably require changing the global codepath to understand that when it is being inlined, it shouldn't use the "in a function" path?

dnwpark:
@msullivan Inlined functions already will use whatever global is available: either the globals param or from the json in non-inlined function bodies.

The function `get_compiler_options` copies `options.func_params` from an inlining context. In general, the check for whether an expr is within a function body or not is to check `func_params is None`. As a result, inlined globals will do whatever they would have done if the function body was directly substituted.

I've removed the unneeded extra global args compilation.

msullivan (approved):

PR #7921, opened 2024-10-24
Implement --uninstall and --list-packages commands for load_ext

msullivan:
Note, though, that this only uninstalls the extension from the "server
installation".

The extension package may have already been loaded into the
sys::ExtensionPackage tables of instances, though, and this is not
able to remove it from there.

Removing a package that is actually in use will probably cause
trouble.

### DIFF
--- a/edb/load_ext/main.py
+++ b/edb/load_ext/main.py
@@ -257,15 +257,53 @@ def load_ext_uninstall(
         shutil.rmtree(target_dir)
 
 
+def load_ext_list_packages() -> None:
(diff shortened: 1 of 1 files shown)


elprans:
> Removing a package that is actually in use will probably cause trouble.

You will not be allowed to, normal schema dangling ref checks apply.  A possible future improvement might be to add an option to issue a `DROP EXTENSION PACKAGE` as well.

mmastrac (approved):

PR #7923, opened 2024-10-25
Factoring of columns in USING clause over SQL adapter

aljazerzen:
This PR makes the last `id` identifier in this query:
```
SELECT
FROM x
JOIN y USING (id)
WHERE id = 4  -- this id is not ambiguous 
```
... not ambiguous, but instead equivalent to `x.id`.

### DIFF
--- a/edb/pgsql/resolver/context.py
+++ b/edb/pgsql/resolver/context.py
@@ -73,7 +73,7 @@ class Scope:
     ctes: List[CTE] = field(default_factory=lambda: [])
 
     # Pairs of columns of the same name that have been compared in a USING
-    # clause. This makes unqualified references to their name them un-ambagious.
+    # clause. This makes unqualified references to their name them un-ambiguous.
     # The fourth tuple element is the join type.
     factored_columns: List[Tuple[str, Table, Table, str]] = field(
         default_factory=lambda: []
--- a/edb/pgsql/resolver/context.py
+++ b/edb/pgsql/resolver/context.py
@@ -72,6 +72,13 @@ class Scope:
     # Common Table Expressions
     ctes: List[CTE] = field(default_factory=lambda: [])
 
+    # Pairs of columns of the same name that have been compared in a USING
+    # clause. This makes unqualified references to their name them un-ambagious,
+    # it resolves to the column in the left (first) table.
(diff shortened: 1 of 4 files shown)


edb/pgsql/resolver/context.py
```diff
@@ -72,6 +72,13 @@ class Scope:
     # Common Table Expressions
     ctes: List[CTE] = field(default_factory=lambda: [])
 
+    # Pairs of columns of the same name that have been compared in a USING
+    # clause. This makes unqualified references to their name them un-ambagious,
+    # it resolves to the column in the left (first) table.
```
msullivan, line 77: Does it do the left table for RIGHT JOIN and FULL OUTER JOIN as well?

elprans (approved):

edb/pgsql/resolver/context.py
msullivan: No, it looks like for RIGHT JOIN it uses the right key and for FULL OUTER JOIN it uses whichever of them is non-NULL.
So I guess we need to be sensitive to what the join type is, and take the right for RIGHT JOIN and take `coalesce(left, right)` for FULL OUTER JOIN?
(Or maybe just decide we don't care much and always do the coalesce??)

elprans: coalescing might interfere with index use, so need to be cognizant of that

edb/pgsql/resolver/context.py
```diff
@@ -72,6 +72,13 @@ class Scope:
     # Common Table Expressions
     ctes: List[CTE] = field(default_factory=lambda: [])
 
+    # Pairs of columns of the same name that have been compared in a USING
+    # clause. This makes unqualified references to their name them un-ambagious.
```
msullivan, line 76: s/ambagious/ambiguous/

edb/pgsql/resolver/expr.py
```diff
@@ -268,6 +270,44 @@ def _lookup_column(
             (t, c) for t, c in matched_columns if t.precedence == max_precedence
         ]
 
+    # when ambiguous references have been used in USING clause,
+    # pick the column from the left table
```
msullivan, line 274: Comment out of date?

PR #7926, opened 2024-10-25
Support DROP EXTENSION PACKAGE of user-installed extensions

msullivan:
The main lift here is in distinguishing them from builtin extensions.


edb/server/tenant.py
```diff
@@ -520,10 +520,22 @@ async def _load_extension_package(
             compiler_state=compiler.state,
             global_schema=global_schema,
             user_schema=s_schema.FlatSchema(),
-            bootstrap_mode=True,
             internal_schema_mode=True,
         )
 
+        # Extension installation only works if stdmode or testmode is
```
elprans, line 526: Should we maybe remove this constraint?

elprans (approved):

PR #7927, opened 2024-10-25
Full http in Rust support, remove httpx

mmastrac:
Makes use of HttpClient throughout the entire server process:

 - Auth
 - AI extension (incl. SSE)
 - std::net::http

Notes:
 - `httpx` is no longer a dep (kept as a dev-dependency)
 - Uses an `EdgeDB ${version}` User-Agent for all HTTP requests
 - All HTTP requests are subject to `http_max_connections` config option. This option is refreshed from the std::net worker.
 - SSE supports backpressure, and queue is fixed at 100 messages for now
 - Streaming and non-streaming requests use separate APIs to simplify the implementation
 - One thread per tenant is used if HTTP is required (we boot this thread only when needed). Eventually we'll move to a one-runtime-per-tenant and this thread will disappear.


edb/server/http/src/python.rs
```diff
@@ -284,34 +397,108 @@ async fn run_and_block(capacity: usize, rpc_pipe: RpcPipe) {
...
-            use PythonToRustMessage::*;
-            match rpc {
-                UpdateLimit(limit) => {
-                    permit_manager.update_limit(limit);
+        let (id, backpressure) = match rpc {
+            PythonToRustMessage::Request(id, ..) | PythonToRustMessage::RequestSse(id, ..) => {
+                (Some(id), Some(Semaphore::new(SSE_QUEUE_SIZE).into()))
+            }
+            _ => (None, None),
+        };
```
msullivan, line 405: Add a comment on what we're doing here exactly?

msullivan (approved):

msullivan:
I decided to go get rid of it as a test dependency: https://github.com/edgedb/edgedb/pull/7950

PR #7931, opened 2024-10-28
Start renaming server components

msullivan:
- Update a number of messages we print
- Change some docstrings that were showing up in my search of string
  constants but aren't user facing
- Update user-facing environment variables to also support versions
  with the new name. If both are specified, emit a warning and prefer
  the new one.
- Create new entry point executables for our commands. (Do we need to
  do this a second time in edgedb-pkg?)

My philosophy here is that "edgedb" will remain the "internal
codename" of the product probably forever. Anything that is just
developer facing will retain its current name.


aljazerzen (approved):
Don't merge this before we cut the 6.x branch.

msullivan:
I think the plan is that this should be in 6.x

1st1:
> I think the plan is that this should be in 6.x

Correct

MartinCura:
👀 I'm dying of curiosity of why the rename

PR #7935, opened 2024-10-29
Export metrics about extension use

msullivan:
Export a per-extension metric indicating which extensions are in use.
Really it reports how many branches each extension is in use by.

### DIFF
--- a/edb/server/dbview/dbview.pyx
+++ b/edb/server/dbview/dbview.pyx
@@ -307,15 +307,11 @@ cdef class Database:
         tname = self.tenant.get_instance_name()
         for ext in self.extensions:
             if ext not in extensions:
-                metrics.extension_used.dec(
-                    1, self.tenant.get_instance_name(), ext
-                )
+                metrics.extension_used.dec(1, tname, ext)
 
         for ext in extensions:
             if ext not in self.extensions:
-                metrics.extension_used.inc(
-                    1, self.tenant.get_instance_name(), ext
-                )
+                metrics.extension_used.inc(1, tname, ext)
 
         self.extensions = extensions
 
--- a/edb/server/dbview/dbview.pyx
+++ b/edb/server/dbview/dbview.pyx
@@ -159,7 +159,8 @@ cdef class Database:
             self.user_config_spec = config.FlatSpec(*ext_config_settings)
         self.reflection_cache = reflection_cache
         self.backend_ids = backend_ids
-        self.extensions = extensions
+        self.extensions = set()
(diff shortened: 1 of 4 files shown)


edb/server/metrics.py
```diff
@@ -194,6 +194,12 @@
     labels=('tenant', 'branch'),
 )
 
+extension_used = registry.new_labeled_gauge(
+    'extension_used',
+    'How many branches an extension is used by.',
```
zackelan, line 199: we should follow https://prometheus.io/docs/practices/naming/#metric-names whenever possible - something like `extension_used_branch_count`  means you can tell what 4 vs 5 means without having to refer back to the help text (which appears in the `/metrics` output, but isn't always propagated from there to a database like Prometheus, a visualizer like Grafana, etc)

edb/server/dbview/dbview.pyx
```diff
@@ -301,6 +302,23 @@ cdef class Database:
             max_batch_size=100,
         )
 
+    cdef _set_extensions(self, extensions):
+        # Update metrics about extension use
+        tname = self.tenant.get_instance_name()
```
scotttrinh, line 307: Did you mean to use `tname` below instead of calling `get_instance_name` in the two loops?

scotttrinh (commented):
General question: does this mean this number goes up every time the server restarts? Assuming the server contains 5 branches that all use `ext::ai`, for instance: Does restarting the server 3 times mean the count for `ext::ai` is `15`?

edb/server/dbview/dbview.pyx
msullivan: Indeed. Thanks for the catch

edb/server/metrics.py
msullivan: Sure. How about `extension_used_branch_count_current`, since our prometheus implementation in python seems to think we shouldn't end a metric with `_count`.

msullivan:
> General question: does this mean this number goes up every time the server restarts? Assuming the server contains 5 branches that all use `ext::ai`, for instance: Does restarting the server 3 times mean the count for `ext::ai` is `15`?

No, it is a "gauge", so it measures current levels

dnwpark (approved):
Maybe worth adding a test to check that multiple branches works as expected, otherwise looks fine.

msullivan:
> Maybe worth adding a test to check that multiple branches works as expected, otherwise looks fine.

I've got a follow-up PR coming that adds some more metrics; I'll put a multiple branch test in that one, to save myself some rebasing and rerunning of CI

Issue #7922, opened 2024-10-25
Improvements of SQL introspection


aljazerzen:
Currently, introspection over SQL adapter is a bit lacking. In some cases, we leak internal database structure that should not be visible to end users (i.e. schema `edgedbpub`) and in some cases we don't expose features because they don't exist in the internal database (foreign keys).

This issue is a tracking issue of all such problems:

- [ ] defaults - some tools look at pg_attribute to see if a column is AUTO generated or not. We don't use PostgreSQL defaults, but inject the defaults in our compilation process. We should set column `pg_attribute.atthasdef` for those columns and populate table `pg_attrdef`,
- [ ] primary keys - I don't know where, but `id` is not being picked up (by Django) as the primary key. 
- [ ] foreign keys - ORMs depend on detecting foreign keys to translate them into links,
- [ ] format_type - is implemented incorrectly and should be redone fully,
- [ ] pg_tables and pg_views are not being populated, https://github.com/edgedb/edgedb/issues/6403
- [ ] `FROM x JOIN y ON x.id = y.id WHERE id = 4` is **not** ambiguous (not really an introspection issue)

aljazerzen:
We also need to decide on which schemas to expose. Currently, this is what is seen in an empty database:
```
pg_toast                -- PostgreSQL
pg_catalog              -- PostgreSQL
information_schema      -- PostgreSQL

edgedb_v6_2f209700f0    -- for module schema::
edgedb                  -- for module schema::, trampolined

edgedbstd_v6_2f209700f0 -- for module std::
edgedbstd               -- for module std::, trampolined

edgedbt 				-- for datetime types in std::

public                  -- for module default
```

I think we should:
- hide all trampolined schemas,
- rename `edgedb` into `schema`,
- rename `edgedbstd` into `std`,
- merge `edgedbt` into `std`.

PR #7937, opened 2024-10-29
Emulate introspection of primary keys over SQL adapter

aljazerzen:
Ref #7922


msullivan (approved):
Add a comment explaining the helper view technique being used and why?

PR #7940, opened 2024-10-29
Export metric about schema feature use

msullivan:
Currently report whether the following are used:
 - policies
 - triggers
 - rewrites
 - globals
 - computed globals
 - aliases
 - functions
 - computed pointers
 - FTS

### DIFF
--- a/edb/server/compiler/ddl.py
+++ b/edb/server/compiler/ddl.py
@@ -37,6 +37,9 @@ from edb.edgeql import compiler as qlcompiler
 from edb.edgeql import qltypes
 from edb.edgeql import quote as qlquote
 
+
+from edb.schema import annos as s_annos
+from edb.schema import constraints as s_constraints
 from edb.schema import database as s_db
 from edb.schema import ddl as s_ddl
 from edb.schema import delta as s_delta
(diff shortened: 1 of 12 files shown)


edb/server/metrics.py
```diff
@@ -200,6 +200,12 @@
     labels=('tenant', 'extension'),
 )
 
+feature_used = registry.new_labeled_gauge(
+    'feature_used_branch_count_current',
+    'How many branches a schema feature is used by.',
+    labels=('tenant', 'feature'),
+)
```
1st1, line 207: @zackelan please confirm we can plot this later

edb/server/compiler/ddl.py
```diff
@@ -1169,6 +1184,53 @@ def _reset_schema(
     )
 
 
+_FEATURE_NAMES: dict[type[s_obj.Object], str] = {
+    s_policies.AccessPolicy: 'policy',
+    s_triggers.Trigger: 'trigger',
+    s_rewrites.Rewrite: 'rewrite',
+    s_globals.Global: 'global',
+    s_expraliases.Alias: 'alias',
+    s_func.Function: 'function',
```
1st1, line 1193: Not necessarily for this PR, but maybe we can add:

* link properties
* more than one base type on types
* annotations
* indexes (i'd like to see how often people define them!)
* constraints (ditto)
* computed properties (ditto)
* computed links
* multi properties
* enums

1st1 (approved):
Amazing stuff, I can't wait until we start seeing these stats.

edb/server/metrics.py
zackelan: yes, this fits well into the Prometheus data model, so shouldn't pose any problems with aggregating it after deployment

edb/server/compiler/ddl.py
msullivan: @zackelan I can't tell if I'm missing something here: is there are major difference between what you are suggesting we do and what I'm doing here?

msullivan: @1st1 I added all those (a third or so were already there)

zackelan: nothing in particular - just a +1 to Yury's suggestion of adding metrics like Oprah giving away cars

PR #7942, opened 2024-10-30
Migrate `ext::auth::SMTPConfig` -> `cfg::EmailProvider`

scotttrinh:
This will provide multiple choices for email providers allowing switching
between providers. Moves the SMTP config into the main config namespace to allow for instance-wide configuration.


msullivan:
Yeah, this is what I was thinking of basically.

edb/server/smtp.py
```diff
@@ -0,0 +1,225 @@
...
+        *,
+        test_mode: bool = False,
+    ) -> None:
+        global _semaphore
+        if _semaphore is None:
+            _semaphore = asyncio.BoundedSemaphore(
+                int(
+                    os.environ.get(
+                        "EDGEDB_SERVER_AUTH_SMTP_CONCURRENCY",
+                        os.environ.get("EDGEDB_SERVER_SMTP_CONCURRENCY", 5),
```
scotttrinh, line 122: wdyt?

msullivan: Can this not go through the config system somehow?

msullivan: I guess "probably, but also the existing behavior doesn't"?

scotttrinh: Well, now is a great time to add it to the config!

edb/lib/cfg.edgeql
```diff
@@ -102,6 +102,71 @@ CREATE TYPE cfg::Auth EXTENDING cfg::ConfigObject {
...
+CREATE ABSTRACT TYPE cfg::EmailProviderConfig EXTENDING cfg::ConfigObject {
+    CREATE REQUIRED PROPERTY name -> std::str {
+        CREATE CONSTRAINT std::exclusive;
+        CREATE ANNOTATION std::description :=
+            "The name of the email provider.";
+    };
+};
+
+CREATE TYPE cfg::SMTPProviderConfig EXTENDING cfg::EmailProviderConfig {
+    CREATE PROPERTY sender -> std::str {
```
scotttrinh, line 121: We should make `sender` required, I'm not sure why we didn't have that before.

edb/lib/cfg.edgeql
```diff
@@ -102,6 +102,71 @@ CREATE TYPE cfg::Auth EXTENDING cfg::ConfigObject {
...
+    CREATE REQUIRED PROPERTY timeout_per_email -> std::duration {
+        SET default := <std::duration>'60 seconds';
+        CREATE ANNOTATION std::description :=
+            "Maximum time to send an email, including retry attempts.";
+    };
+    CREATE REQUIRED PROPERTY timeout_per_attempt -> std::duration {
+        SET default := <std::duration>'15 seconds';
+        CREATE ANNOTATION std::description :=
+            "Maximum time for each SMTP request.";
+    };
```
scotttrinh, line 167: Max concurrency config here?

scotttrinh (commented):
Notes to self after a quick self-review.

edb/server/server.py
```diff
@@ -1282,6 +1283,20 @@ def get_default_tenant(self) -> edbtenant.Tenant:
     def iter_tenants(self) -> Iterator[edbtenant.Tenant]:
         yield self._tenant
 
+    def _load_sidechannel_configs(self) -> None:
+        # TODO(fantix): Do something like this for multitenant
```
fantix, line 1287: ✔️

fantix (approved):
LGTM!

edb/lib/cfg.edgeql
fantix: Yeah, I think so. Though this would become a per-provider config, instead of a server/tenant-wide config. It should be fine for now, but in the long run we may want to have both: per-provider config makes sure of not overwhelming the provider, while the server/tenant-wide config makes sure e.g. the local number of FDs is not exceeded.

msullivan:
Anything blocking this?

PR #7944, opened 2024-10-30
Add logs to auth extension

scotttrinh:
To help with debugging, troubleshooting, and observability.

The general logging strategy is:
- Add info message for when the auth extension is attempting to handle an HTTP request
- Add logging to the main `except` block that catches uncaught exceptions so that you get server logs when a request fails
- Add logging to every authentication endpoint for both success (`info` messages) and failures (`debug` and/or `error` messages)
- Logs include information like identity IDs, and other useful identifying information (`email`, `email_factor_id`, `credential_id`, etc).

---

I ended up doing a little bit of unnecessary refactoring to some of the endpoints that made some cases a bit easier to log errors from, but also aligned endpoints written earlier in the extensions life with more recent innovations. Apologies for the noise.

### DIFF
--- a/edb/server/protocol/auth_ext/http.py
+++ b/edb/server/protocol/auth_ext/http.py
@@ -30,6 +30,7 @@ import hashlib
 import os
 import mimetypes
 import uuid
+import dataclasses
 
 from typing import (
     Any,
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -23,7 +23,7 @@ dependencies = [
     'hishel==0.0.24',
     'webauthn~=2.0.0',
     'argon2-cffi~=23.1.0',
-    'aiosmtplib~=2.0',
+    'aiosmtplib~=3.0',
```
scotttrinh, line 26: I wanted types for the `send` function, so I bumped this.

edb/server/protocol/auth_ext/http.py
```diff
@@ -1219,14 +1301,15 @@ async def handle_magic_link_email(
             if redirect_on_failure is None:
                 raise ex
             else:
-                if not self._is_url_allowed(redirect_on_failure):
-                    raise errors.InvalidData(
-                        "Redirect URL does not match any allowed URLs.",
-                    )
```
jaclarke, line 1225: Is this intentionally removed?

scotttrinh: oof, good catch. I'll do another audit to make sure there aren't more unintentional omissions like this.

scotttrinh: Well... That escalated quickly: #7966

jaclarke (approved):

PR #7945, opened 2024-10-30
Enable DML in user-defined functions.

dnwpark:
Allows users to create dml functions.


1st1:
I'm not sure we should remove it. Maybe just make it no-op?

msullivan:
> I'm not sure we should remove it. Maybe just make it no-op?

Yeah, removing it would break dump compat. (See also: the test failures)

msullivan:
I would title this commit "Enable DML support in user-defined functions" or some such, since it is now available for users, right?

Issue #6403, opened 2023-10-31
Fill-out pg_catalog introspection SQL adapter


msullivan:
`pg_class` has an `attnum` field that we should probably treat like `ordinal_position`

There are also `pg_table` and `pg_view` views that maybe we need to intercept?

aljazerzen:
You probably mean "`pg_attribute` has an `attnum` column". I've fixed that in #6507, because it was tripping pg_dump when `attnum` were not sequential, which was happening when `__fts_document__` got hidden.

`pg_tables` and `pg_views` remains to be done.

Can I assume we can hide all views outside pg_catalog and information_schema?

For `pg_tables` I intend to pull data from our `edgedbsql.virtual_tables`.

PR #7946, opened 2024-10-30
Emulate foreign keys for single links in introspection over SQL adapter

aljazerzen:
Ref https://github.com/edgedb/edgedb/issues/6403

Foreign & primary keys on link tables are left to do, in a separate PR.

### DIFF
--- a/edb/pgsql/metaschema.py
+++ b/edb/pgsql/metaschema.py
@@ -6364,7 +6364,10 @@ def _generate_sql_information_schema(
             pi.indisclustered,
             pi.indisvalid,
             pi.indcheckxmin,
-            FALSE AS indisready, -- prevent pg_dump from recreating indexes
+            CASE
+                WHEN COALESCE(is_id.t, FALSE) THEN TRUE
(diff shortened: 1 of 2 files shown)


edb/pgsql/metaschema.py
```diff
@@ -6378,10 +6364,16 @@ def _generate_sql_information_schema(
             pi.indisclustered,
             pi.indisvalid,
             pi.indcheckxmin,
-            pi.indisready,
+            CASE
+                WHEN COALESCE(is_id.t, FALSE) THEN TRUE
+                ELSE FALSE -- override so pg_dump won't try to recreate them
+            END AS indisready,
```
msullivan, line 6370: This is equivalent to `COALESCE(is_id.t, FALSE)`, right? The `CASE` seems unnecessary.

msullivan (approved):

edb/pgsql/metaschema.py
aljazerzen: The CASE is detecting whether this is a primary key and has this form for a few different columns in the view.

PR #7950, opened 2024-10-31
Drop use of httpx as a test dependency

msullivan:
We only use it in one test and are getting rid of it as a real
dependency (see #7927), so let's just drop it.


scotttrinh (approved):
🆒

mmastrac (commented):
LGTM -- just needs the pyproject.toml changes

PR #7951, opened 2024-10-31
Fix load-ext when .zip filename doesn't match internal folder name

msullivan:
There was a mismatch that was causing trouble.


mmastrac (approved):
Nice find, LGTM

PR #7958, opened 2024-11-02
Add a debug flag to validate schema reflection

### LINKED ISSUE #5169: Better testing of schema reflection
Our mechanism for reading/rewriting the schema to the database is surprisingly under-tested. We only do the load at database startup time (otherwise we just use the in memory representation), and so if I understand correctly we almost never exercise the "reader" (load schema from database) path. I think we only really use it in dump tests.

We should add a test mode that reloads the schema after every DDL command and compares it to the in-memory version.
If it turns out to be fast enough, we can just always turn it on. If (more likely) it is dog slow, we can run it in a CI job once a day or something.

msullivan:
When the flag is set, after every DDL command we reintrospect the
database and compare that to the in-meory schema. I've recently landed
a few PRs fixing issues exposed by this. (IIRC, I also fixed a few
earlier this year when I first wrote this PR.) Overall the system is
pretty solid.

Run it in CI once a day, since I think it is too slow to run all the
time.

Fixes #5169.


.github/Makefile
```diff
@@ -13,7 +13,8 @@ all: workflows/nightly.yml \
 	workflows/tests-ha.yml \
 	workflows/tests-pg-versions.yml \
 	workflows/tests-patches.yml \
-	workflows/tests-inplace.yml
+	workflows/tests-inplace.yml \
+	workflows/tests-reflection.yml \
```
dnwpark, line 17: minor but probably don't want to end with `\`

dnwpark (commented):
Looks good as far as I can tell, but I'll wait for someone else to look at it for a review

elprans (approved):

PR #7963, opened 2024-11-04
Update ai ext

diksipav:


edb/server/protocol/ai_ext.py
```diff
@@ -354,11 +355,13 @@ async def _ext_ai_fetch_active_models(
...
-                model[1].decode("utf-8"),
-                model[2].decode("utf-8"),
-            ))
+            result.append(
+                (
+                    int.from_bytes(model[0], byteorder="big", signed=True),
+                    model[1].decode("utf-8"),
+                    model[2].decode("utf-8"),
+                )
+            )
```
1st1, line 364: @diksipav I'm not sure what happened here -- a code formatter tool? but please revert all changes that are pure syntax/whitespace changes.

edb/server/protocol/ai_ext.py
```diff
@@ -1205,53 +1251,133 @@ async def _start_openai_like_chat(
...
+                            + b'data: {"type": "content_block_stop",'
+                            + b'"index": ' + str(tool_index).encode() + b'}\n\n'
+                        )
+                        protocol.write_raw(event)
                     event = (
                         b'event: message_stop\n'
                         + b'data: {"type": "message_stop"}\n\n'
                     )
                     protocol.write_raw(event)
-                    continue
```
diksipav, line 1213: continue here was working before we switched to using HttpClient, now it seems like stream doesn't finish

edb/server/protocol/ai_ext.py
diksipav: Hey, I fixed formatting. @elprans, @dnwpark can you pls take a look at this PR, I'll be on vacation next week. But would like to move this forward, and be able to add new things in new PRs.

edb/server/protocol/ai_ext.py
diksipav: The PR is ready for review. @elvis I hope u can understand my question below.

But I just want to clarify how is the ext meant to be used, especially the `prompt.custom` part. I think I've already asked this. And I just wanna be sure when I start writing docs.

When using provider user can provide `messages` (chat history), but user can also provide `prompt.custom` messages.  Inside the provider I merge the too and put everything inside `prompt.custom`.  

```ts
  const textModel = edgedbRag.languageModel("gpt-4-turbo-preview");
  
  const result = await streamText({
    model: textModel.withSettings({
      context: { query: "Book" },
      prompt: { name: "builtin::rag-default" },
    }),
    messages: requestData.messages,
  });
  ```
  
This is the prompt interface:
```ts
export type Prompt =
  | { name: string; custom?: EdgeDBRagMessage }
  | { id: string; custom?: EdgeDBRagMessage[] }
  | { custom: EdgeDBRagMessage[] };
  ```.
  
In the ext if user doesn't provide prompt at all by default the "builtin::rag-default" will be used, but if user provides just prompt.custom without name or id, then prompt.custom will be used. 

So now when user provides messages but not prompt.custom, inside the provider I still set the default builtin prompt to be used because user didn't deliberately said: hey use this prompt.custom only. I do this because I want user to be able to do just:

  const result = await streamText({
    model: textModel.withSettings({
      context: { query: "Book" },
    }),
    messages: requestData.messages,
  });

instead to have to also provide the prompt name/id like in the example above.

Comment (elprans) [reply edb/server/protocol/ai_ext.py:1213]: Seems reasonable to me.

File: edb/server/protocol/ai_ext.py [hunk 3]
```diff
@@ -1125,15 +1126,27 @@ async def _start_chat(
     model_name: str,
     messages: list[dict[str, Any]],
     stream: bool,
+    temperature: Optional[float],
```
elprans, line 1129: Let's make these keyword arguments.

edb/server/protocol/ai_ext.py
```diff
@@ -1181,18 +1194,51 @@ async def _start_openai_like_chat(
...
```
elprans, line 1197: Make these kwargs

edb/server/protocol/ai_ext.py
```diff
@@ -1181,18 +1194,51 @@ async def _start_openai_like_chat(
...
+    logit_bias: Optional[dict[int, int]],
+    logprobs: Optional[bool],
+    user: Optional[str],
+    tools: Optional[list[dict[str, Any]]]
 ) -> None:
+    isOpenAI = "openai" in str(getattr(client, "base_url", ""))
+    params = {
+        "model": model_name,
+        "messages": messages,
+        **({"temperature": temperature} if temperature is not None else {}),
```
elprans, line 1211: This would probably be easier to read if you just used a plain `if`:

```python
params = {
    "model": model_name,
    "messages": messages,
}
if temperature is not None:
    params["temperature"] = temperature
...
```

edb/server/protocol/ai_ext.py
```diff
@@ -1205,53 +1251,133 @@ async def _start_openai_like_chat(
...
+                            + b'"index": ' + str(tool_index).encode() + b'}\n\n'
+                        )
+                        protocol.write_raw(event)
                     event = (
                         b'event: message_stop\n'
                         + b'data: {"type": "message_stop"}\n\n'
                     )
                     protocol.write_raw(event)
-                    continue
+                    break
```
elprans, line 1267: Hmm, @mmastrac any idea why this could be the case?  FWIW breaking explicitly on the client side is probably OK, but this might indicate a bug in the new http client code.

edb/server/protocol/ai_ext.py
```diff
@@ -1591,18 +1909,19 @@ async def _handle_rag_request(
             if custom_prompt:
                 if not isinstance(custom_prompt, list):
                     raise TypeError(
-                        "prompt.custom must be a list of {role, content} "
+                        "prompt.custom must be a list"
```
elprans, line 1912:
```suggestion
                        "prompt.custom must be a list of "
```

edb/server/protocol/ai_ext.py
diksipav: hey @elprans can u pls tell me do u mean all arg should be positional or only everything after http_client for example:
  ```
  protocol: protocol.HttpProtocol,
    request: protocol.HttpRequest,
    response: protocol.HttpResponse,
    provider: ProviderConfig,
    http_client: http.HttpClient,
    *,
    model_name: str,
    messages: list[dict[str, Any]],
    ....
```

File: edb/server/protocol/ai_ext.py (hunk 7 above)
Comment (diksipav) [reply edb/server/protocol/ai_ext.py:1912]: hm these are possible values, not sure how to add all this into an err message
Screenshot 2024-11-12 at 16 34 05

File: edb/server/protocol/ai_ext.py (hunk 3 above)
Comment (elprans) [reply edb/server/protocol/ai_ext.py:1129]: I would make everything keyword

File: edb/server/protocol/ai_ext.py (hunk 6 above)
Comment (mmastrac) [reply edb/server/protocol/ai_ext.py:1267]: Fixed this in https://github.com/edgedb/edgedb/pull/7984

File: edb/server/protocol/ai_ext.py [hunk 8]
```diff
@@ -1591,19 +1949,24 @@ async def _handle_rag_request(
             if custom_prompt:
                 if not isinstance(custom_prompt, list):
                     raise TypeError(
-                        "prompt.custom must be a list of {role, content} "
-                        "objects"
+                        "prompt.custom must be a list, where each element is \
+                            one of the following types: \
+                        - EdgeDBSystemMessage: { role, content } \
```
elprans, line 1954: Where does `EdgeDBSystemMessage` and others come from?  Seems a bit confusing since those aren't actual keys in input?

elprans: Yeah, but this is server-side.  If one interacts with the API via another SDK the errors would not make sense.

diksipav: I get it, I believe the error is better now. It should be understandable enough for everyone.

edb/server/protocol/ai_ext.py
```diff
@@ -1205,53 +1289,133 @@ async def _start_openai_like_chat(
...
+                                        },
+                                    }).encode("utf-8")
+                                event = (
+                                    b'event: content_block_delta\n'
+                                    + b'data:' + event_data + b'\n\n'
+                                )
+                                protocol.write_raw(event)
                     elif finish_reason := data.get("finish_reason"):
+                        index = tool_index \
+                        if finish_reason == "tool_calls" else 0
```
dnwpark, line 1405: Generally avoid `\` to continue a line.

Here, i would use
```py
index = (
    tool_index if finish_reason == "tool_calls" else 0
)
```

edb/server/protocol/ai_ext.py
```diff
@@ -1401,44 +1691,85 @@ async def _start_anthropic_chat(
...
                         b'event: content_block_start\n'
-                        + b'data: ' + sse.data.encode("utf-8") + b'\n\n'
+                        + b'data: ' + json.dumps(sse_data).encode("utf-8")
+                        + b'\n\n'
                     )
+                    # we don't send content_block_stop event when text
+                    # chunk ends, should be okay since we don't consume
+                    # this event on the client side
+                    data = sse.json()
+                    if data.get("content_block") and \
```
dnwpark, line 1704: Similar to above regarding `\`.

edb/server/protocol/ai_ext.py
```diff
@@ -1447,10 +1778,37 @@ async def _start_anthropic_chat(
...
         response.status = http.HTTPStatus.OK
         response.content_type = b'application/json'
-        response_text = result.json()["content"][0]["text"]
-        response.body = json.dumps({
-            "response": response_text,
-        }).encode("utf-8")
+
+        result_data = result.json()
+        tool_calls = [item for item in result_data["content"]
+                      if item.get("type") == "tool_use"]
```
dnwpark, line 1784: Put list contents in new indent:
```py
tool_calls = [
    item for item in result_data["content"]
    if item.get("type") == "tool_use"
]
```

edb/server/protocol/ai_ext.py
```diff
@@ -1591,19 +1949,27 @@ async def _handle_rag_request(
...
                     )
                 for entry in custom_prompt:
                     if (
                         not isinstance(entry, dict)
                         or not entry.get("role")
-                        or not entry.get("content")
-                        or len(entry) > 2
+                        # content can be empty string too
+                        or "content" not in entry
+                        or len(entry) > 3
```
dnwpark, line 1969: why is it ok to have `len(entry) == 3`?

edb/server/protocol/ai_ext.py
```diff
@@ -1174,25 +1214,69 @@ async def aconnect_sse(
...
+        params["top_p"] = top_p
+    if tools is not None:
+        params["tools"] = tools
+    if isOpenAI and logit_bias is not None:
+        params["logit_bias"] = logit_bias
+    if isOpenAI and logprobs is not None:
+        params["logprobs"] = logprobs
+    if isOpenAI and user is not None:
+        params["user"] = user
+    if not isOpenAI and safe_prompt is not None:
```
dnwpark, line 1252: Also since the only two possibilities currently are OpenAI and Mistral, maybe we should use have `isMistral`. Ok with not doing this though.

dnwpark (commented):
Looks good, I've mostly commented on some style things.

edb/server/protocol/ai_ext.py
diksipav: thanks a lot!

edb/server/protocol/ai_ext.py
diksipav: hm I will add it, but it is questionable will we use the same code in the future if we add support for the new AI provider, mistral and openai have quite similar APIs so it was possible to use one flow for both, anthropic has its own separate handler func in the extension

edb/server/protocol/ai_ext.py
```diff
@@ -1174,25 +1214,69 @@ async def aconnect_sse(
...
+    top_p: Optional[float],
+    max_tokens: Optional[int],
+    seed: Optional[int],
+    safe_prompt: Optional[bool],
+    logit_bias: Optional[dict[int, int]],
+    logprobs: Optional[bool],
+    user: Optional[str],
+    tools: Optional[list[dict[str, Any]]],
 ) -> None:
+    isOpenAI = "openai" in str(getattr(client, "base_url", ""))
```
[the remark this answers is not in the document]
diksipav: I updated the type err here too, to:
  ```
raise TypeError(
                            """prompt.custom must be a list, where each element is
                                one of the following types:\n
                            - system message: { role: 'system', content: str }\n
                            - user message: { role: 'user', content:
                                [{ type: 'text', text: str }] }\n
                            - assistant message: { role: 'assistant', content: str,
                                optional tool_calls: [{id: str, type: 'function',
                                function: { name: str, arguments: str }}] }\n
                            - tool message:
                                { role: 'tool', content: str, tool_call_id: str }"""
                        )
```
the thing is that the message can take few shapes, role and content should always be there but there can be `tool_call_id` or `tool_calls` too depending on the message role. Maybe I should add all these somehow to the if part?

File: edb/server/protocol/ai_ext.py (hunk 12 above)
Comment (diksipav) [reply edb/server/protocol/ai_ext.py:1969]: one of the following types:\n
                            - system message: { role: 'system', content: str }\n
                            - user message: { role: 'user', content:
                                [{ type: 'text', text: str }] }\n
                            - assistant message: { role: 'assistant', content: str,
                                optional tool_calls: [{id: str, type: 'function',
                                function: { name: str, arguments: str }}] }\n
                            - tool message:
                                { role: 'tool', content: str, tool_call_id: str }"""
                        )
```
the thing is that the message can take few shapes, role and content should always be there but there can be `tool_call_id` or `tool_calls` too depending on the message role. Maybe I should add all these somehow to the if part?

edb/server/protocol/ai_ext.py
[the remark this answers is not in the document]
diksipav: hm maybe but I dont get then why we have all this, it will be too much to provide both the provider and client, but imo I'd use one func here for `_start_openai_chat` and `_start_openai_like_chat` 
Screenshot 2024-11-18 at 20 50 16

edb/server/protocol/ai_ext.py
dnwpark: I think that might make the intent clearer.

Perhaps worth just checking all the data:
```py
if not isinstance(entry, dict) or not entry.get('role'):
    raise Something

entry_role = entry.get('role')
if entry_role == 'system':
    if set(entry.keys()) != {'role', content'}:
        raise Something
elif entry_role == 'user':
# ...
```

edb/server/protocol/ai_ext.py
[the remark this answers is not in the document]
dnwpark: Or perhaps we can pass `provider.name` into `_start_open_ai_like_chat`. I just think it's weird to search through the client url.

dnwpark (approved):
Comments all resolved, looks good!

elprans (approved):

PR #7964, opened 2024-11-04
Add constraints to auth app config

scotttrinh:
Since these values end up in user-facing templates, ensure that they are limited to sane lengths and are also well sanitized.

### DIFF
--- a/edb/server/protocol/auth_ext/util.py
+++ b/edb/server/protocol/auth_ext/util.py
@@ -106,7 +106,7 @@ def get_config_typename(config_value: edb_config.SettingValue) -> str:
 def escape_and_truncate(input_str: str | None, max_len: int) -> str | None:
     if input_str is None:
         return None
-    return html.escape(input_str)[:max_len]
+    return html.escape(input_str[:max_len])
 
 
 def get_app_details_config(db: Any) -> config.AppDetailsConfig:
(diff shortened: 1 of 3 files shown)


edb/server/protocol/auth_ext/util.py
```diff
@@ -101,28 +103,38 @@ def get_config_typename(config_value: edb_config.SettingValue) -> str:
     return config_value._tspec.name  # type: ignore
 
 
+def escape_and_truncate(input_str: str | None, max_len: int) -> str | None:
+    if input_str is None:
+        return None
+    return html.escape(input_str)[:max_len]
```
1st1, line 109: escape should happen on the trimmed string, not after -- can risk having broken escape codes

1st1 (commented):
Can you add tests by any chance? Can be a unit test

edb/server/protocol/auth_ext/util.py
```diff
@@ -101,28 +103,38 @@ def get_config_typename(config_value: edb_config.SettingValue) -> str:
     return config_value._tspec.name  # type: ignore
 
 
+def escape_and_truncate(input_str: str | None, max_len: int) -> str | None:
+    if input_str is None:
+        return None
+    return html.escape(input_str[:max_len])
```
1st1, line 109:
```suggestion
    trunc = (
        f"{input_str[:max_len]}..." 
        if len(input_str) > max_len else
        input_str
    )
    return html.escape(trunc)
```

scotttrinh:
@1st1 ⚡ 

Added tests for both the UI and the WebAuthn relying party info (we weren't truncating and escaping there), and added the suggested change showing ellipsis.

1st1 (approved):

PR #7966, opened 2024-11-05
More thoroughly protect redirects

scotttrinh:
Ensure that URLs are checked against the allow list before sending a
redirect. Make a separate type for checked URLs, and centralize the
checking of URLs and constructing of redirects.

### DIFF
--- a/edb/server/protocol/auth_ext/http.py
+++ b/edb/server/protocol/auth_ext/http.py
@@ -30,6 +30,7 @@ import hashlib
 import os
 import mimetypes
 import uuid
+import dataclasses
 
 from typing import (
     Any,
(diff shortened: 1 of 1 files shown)


edb/server/protocol/auth_ext/http.py
```diff
@@ -2534,15 +2540,6 @@ def _set_cookie(
     response.custom_headers["Set-Cookie"] = val.OutputString()
 
 
-def _with_appended_qs(url: str, query: dict[str, list[str]]) -> str:
```
scotttrinh, line 2537: We had duplicate ways of adding to a URL's query string, so went with the more recent `util`-based function and removed this one. The only difference is that this one handles arrays of values, but we were not actually using that functionality.

edb/server/protocol/auth_ext/http.py
```diff
@@ -2445,6 +2421,36 @@ def _is_url_allowed(self, url: str) -> bool:
...
+    ) -> Optional[AllowedUrl]:
+        return self._make_allowed_url(url) if url else None
+
+
+@dataclasses.dataclass
+class AllowedUrl:
+    url: str
+
+    def map(self, f: Callable[[str], str]) -> "AllowedUrl":
+        return AllowedUrl(f(self.url))
```
scotttrinh, line 2452: 😬 uwu i'm just a wittle functional programmer, please let me keep this uwu 🥺

jaclarke (approved):

PR #7968, opened 2024-11-05
Upgrade cython version

msullivan:


pyproject.toml
```diff
@@ -96,14 +96,14 @@ language-server = ['pygls~=1.3.1']
 
 [build-system]
 requires = [
-    "Cython (>=0.29.32, <0.30.0)",
+    "Cython(>=3.0.11,<3.1.0)",
```
elprans, line 99: Let's change the upper bound to 4.0.0.

msullivan: Why? I'm nervous about random breakage, since we've hit it in the past.

elprans (approved):

pyproject.toml
elprans: Mostly concerned with too narrow of a constraint leading to unresolveable build dependency graph given that metapkg doesn't isolate build deps per package.

Issue #2809, opened 2021-08-12
WITH bound volatiles don't work right in DML


msullivan:
```
WITH name := <str>random(),
INSERT Person { name := name, tag := name };
```
will have different `name` and `tag`.
Somewhat related to #1699.

That particular failure can just be attributed to not compiling materialized sets on DML, but slight generalizations become trickier:
```
WITH name := <str>random(),
SELECT (INSERT Person { name := name, tag := name })
```
or worse
```
WITH name := <str>random(),                                               
INSERT Person {                                                           
    name := name,                                                         
    note := (INSERT Note { name := name, note := name }),                 
};
```

The main issue is that it needs to be in a CTE

PR #7969, opened 2024-11-06
Allow volatile WITH in dml statements.

dnwpark:
Given:
```
type Foo { property a -> float64; };
```

Allows the query:
```
with x := random() insert Foo { a := x };
```

Related #2809
Related #7692

### DIFF
--- a/edb/pgsql/compiler/clauses.py
+++ b/edb/pgsql/compiler/clauses.py
@@ -34,6 +34,7 @@ from edb.pgsql import types as pg_types
 from . import astutils
 from . import context
 from . import dispatch
+from . import dml
 from . import enums as pgce
 from . import output
 from . import pathctx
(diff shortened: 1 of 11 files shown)


edb/pgsql/compiler/clauses.py
```diff
@@ -273,6 +283,70 @@ def compile_dml_bindings(
             with ctx.substmt() as bctx:
                 dispatch.compile(binding, ctx=bctx)
 
+        # If something we are WITH binding is volatile and the stmt contains
+        # dml, we similarly want to compile it *now*. Unlike with DML, we need
+        # to manually construct the CTE here.
```
msullivan, line 288: Maybe put this all in a different function

tests/test_edgeql_insert.py
```diff
@@ -5822,23 +5822,34 @@ async def test_edgeql_insert_cardinality_assertion(self):
...
-                SELECT (INSERT Person { name := name, tag := name });
-            ''')
+        await self.assert_query_result(
+            r'''
+                SELECT all(Person.name = Person.tag)
+            ''',
+            [True]
+        )
+
+    @tb.needs_factoring_weakly
```
msullivan, line 5837: I'd like to see some more tests, since this is a fiddly area.

Definitely one with a WITH binding inside of a FOR loop.
Maybe one with it inside a nested INSERT also.

msullivan (approved):

PR #7973, opened 2024-11-06
Bump uvloop, edgedb, and asyncpg deps

msullivan:
So that we are using the same Cython version everywhere.


elprans (approved):
Bump asyncpg also

Issue #7972, opened 2024-11-06
Fix auth email button background color


scotttrinh:
Somehow we're getting the string `"None"` for color if you do not set one rather than falling back to the default color, which causes auth emails to set the button background color to `"None"`.

PR #7974, opened 2024-11-06
Fix email button background

scotttrinh:
When passing `None` explicitly, which we are always doing, the fallback color was not being set. It was also the wrong format considering that we store them without the leading `#`.

---

I tried for an unreasonable amount of time to add tests for this, but the configuration was not stable when running the full test suite, so I just reverted back to no additional tests.

### DIFF
--- a/edb/server/protocol/auth_ext/ui/__init__.py
+++ b/edb/server/protocol/auth_ext/ui/__init__.py
@@ -704,9 +704,9 @@ def render_password_reset_email(
     app_name: Optional[str] = None,
     logo_url: Optional[str] = None,
     dark_logo_url: Optional[str] = None,
-    brand_color: Optional[str] = "007bff",
+    brand_color: Optional[str] = render.DEFAULT_BRAND_COLOR,
 ) -> email.message.EmailMessage:
-    brand_color = brand_color if brand_color is not None else "007bff"
+    brand_color = brand_color or render.DEFAULT_BRAND_COLOR
     msg = email.message.EmailMessage()
     msg["To"] = to_addr
     msg["Subject"] = "Reset password"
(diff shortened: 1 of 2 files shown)


scotttrinh:
Closes #7972

edb/server/protocol/auth_ext/ui/__init__.py
```diff
@@ -707,8 +707,9 @@ def render_password_reset_email(
     app_name: Optional[str] = None,
     logo_url: Optional[str] = None,
     dark_logo_url: Optional[str] = None,
-    brand_color: Optional[str] = "#007bff",
+    brand_color: Optional[str] = "007bff",
 ) -> multipart.MIMEMultipart:
+    brand_color = brand_color if brand_color is not None else "007bff"
```
jaclarke, line 712: Nitpick: I guess the brand color should be const somewhere so we don't have to repeat it.

jaclarke (approved):

edb/server/protocol/auth_ext/ui/__init__.py
scotttrinh: oh, good call, I can do that now.

edb/server/protocol/auth_ext/ui/components.py
```diff
@@ -37,14 +37,17 @@
 ]
 
 
+DEFAULT_BRAND_COLOR = "1f8aed"
```
scotttrinh, line 40: Oddly, the UI uses this as the default color, but emails used `007bff`, or at least _tried_ to use that color. I thought since the fallback never worked, it would be better to standardize on the one color we've actually successfully used 😅

PR #7985, opened 2024-11-12
Remove multiplicity from `QueryUnit.sql`

elprans:
Currently, `QueryUnit.sql` is a tuple representing, possibly, multiple
SQL statements corresponding to the original EdgeQL statement.  This
introduces significant complexity to consumers of `QueryUnit`, which are
mostly unprepared to handle more than one SQL statement anyway.

Originally the SQL tuple was used to represent multiple EdgeQL statements,
a task which is now handled by the `QueryUnitGroup` stuff.  Another use
case are the non-transactional commands (`CREATE BRANCH` and friends).
Finally, since this facility was available, more uses of it were added
without actually _needing_ be executed as multiple SQL statements
with no other recourse: those are mostly maintenance commands and the
DDL type_id readback.

I think the above uses are no longer a sufficient reason to keep the
tuple complexity and so I'm ripping it out here, making `QueryUnit.sql`
a `bytes` property with the invariant of _always_ containing exactly one
SQL statement and thus not needing any special handling.  The users are
fixed up as follows:

- Non-transactional branch units encode the extra SQL needed to
  represent them in the new `QueryUnit.db_op_trailer` property which is
  a tuple of SQL.  Branch commands have special handling for them
  already, so this is not a nuisance.

- The newly-added type id mappings produced by DDL commands are now
  communicated via the new "indirect return" mechanism, whereby a DDL
  PLBlock can communicate a return value via a specially-formatted
  `NoticeResponse` message.

- All other unwitting users of multi-statement `QueryUnit` are converted
  to use a single statement instead (primarily by converting them to use a
  `DO` block).

### DIFF
--- a/edb/pgsql/dbops/base.py
+++ b/edb/pgsql/dbops/base.py
@@ -504,6 +504,35 @@ class Query(Command):
         return f'<Query {self.text!r}>'
 
 
+class PLQuery(Command):
+    def __init__(
+        self,
+        text: str,
+        *,
+        type: Optional[str | Tuple[str, str]] = None,
+        trampoline_fixup: bool = True,
+    ) -> None:
+        from ..import trampoline
+
+        super().__init__()
+        if trampoline_fixup:
+            text = trampoline.fixup_query(text)
+        self.text = text
+        self.type = type
+
+    def to_sql_expr(self) -> str:
+        if self.type:
+            return f'({self.text})::{qn(*self.type)}'
+        else:
+            return self.text
+
+    def code_with_block(self, block: PLBlock) -> str:
+        return self.text
+
+    def __repr__(self) -> str:
+        return f'<PLQuery {self.text!r}>'
+
+
 class DefaultMeta(type):
     def __bool__(cls):
         return False
--- a/edb/pgsql/dbops/base.py
+++ b/edb/pgsql/dbops/base.py
@@ -504,33 +504,8 @@ class Query(Command):
         return f'<Query {self.text!r}>'
 
 
-class PLQuery(Command):
-    def __init__(
-        self,
-        text: str,
-        *,
-        type: Optional[str | Tuple[str, str]] = None,
-        trampoline_fixup: bool = True,
-    ) -> None:
-        from ..import trampoline
-
-        super().__init__()
-        if trampoline_fixup:
-            text = trampoline.fixup_query(text)
-        self.text = text
-        self.type = type
-
-    def to_sql_expr(self) -> str:
-        if self.type:
-            return f'({self.text})::{qn(*self.type)}'
-        else:
-            return self.text
-
-    def code_with_block(self, block: PLBlock) -> str:
-        return self.text
-
-    def __repr__(self) -> str:
-        return f'<PLQuery {self.text!r}>'
+class PLQuery(Query):
+    pass
 
 
 class DefaultMeta(type):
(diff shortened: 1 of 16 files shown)


edb/server/compiler/compiler.py
```diff
@@ -2715,6 +2485,261 @@ def _try_compile(
...
+            if comp.cached_reflection is not None:
+                unit.cached_reflection = \
+                    pickle.dumps(comp.cached_reflection, -1)
+        unit.ddl_stmt_id = comp.ddl_stmt_id
+
+        if comp.modaliases is not None:
+            unit.modaliases = comp.modaliases
+
+        if comp.tx_action == dbstate.TxAction.START:
+            # units[0:0] = _make_tx_units(ctx, qlast.StartTransaction())
```
msullivan, line 2667: ?

edb/server/compiler/ddl.py
```diff
@@ -207,38 +220,42 @@ def compile_and_apply_ddl_stmt(
...
             ddl_stmt_id = str(uuidgen.uuid1mc())
             new_type_ids = [
                 f'{pg_common.quote_literal(tid)}::uuid' for tid in new_types
             ]
-            sql = sql + (
-                trampoline.fixup_query(textwrap.dedent(
-                    f'''\
-                SELECT
-                    json_build_object(
+            new_types_sql = textwrap.dedent(f"""\
```
msullivan, line 254: Add a comment about what we're doing here.

edb/server/compiler/dbstate.py
```diff
@@ -748,17 +737,16 @@ class ProposedMigrationStep(NamedTuple):
...
-            'prompt': self.prompt,
-            'prompt_id': self.prompt_id,
-            'data_safe': self.data_safe,
-            'required_user_input': list(self.required_user_input),
+            "statements": [{"text": stmt} for stmt in self.statements],
+            "confidence": self.confidence,
+            "prompt": self.prompt,
+            "prompt_id": self.prompt_id,
+            "data_safe": self.data_safe,
+            "required_user_input": list(self.required_user_input),
```
msullivan, line 745: It's probably not worth undoing but all these automated changes make this diff pointlessly noisy

msullivan (approved):
It's a little unclear to me that this is actually simpler, since pulling results out via NOTICE feels wildly janky. I guess it won't feel as janky the next time we have to do it, for better or worse.

edb/pgsql/dbops/base.py
```diff
@@ -504,6 +504,35 @@ def __repr__(self) -> str:
         return f'<Query {self.text!r}>'
 
 
+class PLQuery(Command):
```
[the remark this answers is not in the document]
elprans: It used to be more different, I think it's probably worth keeping it around for type-as-documentation purposes to highlight that you can't just run `PLQuery` as you can `Query` and must add it into a `PLBlock` instead.

edb/server/compiler/dbstate.py
1st1: You clearly have something enabled, plenty of changes here just change `'` to `"`

1st1 (approved):

elprans:
> It's a little unclear to me that this is actually simpler, since pulling results out via NOTICE feels wildly janky

Yeah, it does seem like it at first, but the mechanism is reliable.  There are more cleanups to be realized from the single-statement invariant of `QueryUnit` that aren't here.

edb/server/compiler/ddl.py
elprans: Added, I also added an explicit `edgedb.indirect_return()` helper to make things even more clear.

edb/server/compiler/dbstate.py
elprans: Must have pressed a "format file" shortcut or something. It's just this file.

PR #7987, opened 2024-11-12
Fix clippy complaints

elprans:
These are firing on my machine

    error: this manual char comparison can be written more succinctly
       --> edb/edgeql-parser/src/position.rs:127:48
        |
    127 |             if let Some(loff) = prefix_s.rfind(|c| c == '\r' || c == '\n') {
        |                                                ^^^^^^^^^^^^^^^^^^^^^^^^^^ help: consider using an array of `char`: `['\r', '\n']`
        |
        = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#manual_pattern_char_comparison

Not sure why these aren't picked up in CI, but the fixes seem to make
sense.


aljazerzen (approved):
It's not showing in CI because you have a more recent version of clippy on your machine.

Issue #7983, opened 2024-11-11
ISE when enumerating array aggregates


dnwpark:
Given a type:
```
type Person;
```

The query `select enumerate(array_agg((insert Person)));` produces the error:
```
edgedb error: InternalServerError: aggregate functions are not allowed in functions in FROM
```

Additionally, the query `with y := array_agg((insert Person)), x := enumerate(y), select x;` produces the error:
```
edgedb error: InternalServerError: a column definition list is required for functions returning "record"
```

dnwpark:
The second error is related to #4090 and is related to how arrays are unpacked.

In standard pgsql, given:
```pgsql
CREATE TABLE PERSON (
  id int
);

INSERT INTO PERSON(id) VALUES (1);
INSERT INTO PERSON(id) VALUES (2);
INSERT INTO PERSON(id) VALUES (3);
```

The query generates something like:
```pgsql
SELECT
  *
FROM (
  SELECT
    array_agg("Person~1"."ps~1") as "ps~2"
  FROM (
    SELECT
      ROW(id) as "ps~1"
    FROM
      Person
  ) as "Person~1"
) as "Person~2"
CROSS JOIN LATERAL (
  SELECT
    *
  FROM LATERAL ROWS FROM (
    unnest("Person~2"."ps~2") -- ERROR HERE
  ) WITH ORDINALITY as "f"("id")
) as "F"
;
```

Where the error line should be something like
```pgsql
    unnest("Person~2"."ps~2") as ("id" int)
````

The `array_agg` and `unnest` are generated during `eta_expand_ordered`

PR #7988, opened 2024-11-12
Fix ISE when enumerating a call to an aggregate function.

dnwpark:
Aggregate functions are not allowed in `ROWS FROM (...)`.

Related #7983

### DIFF
--- a/edb/pgsql/compiler/relgen.py
+++ b/edb/pgsql/compiler/relgen.py
@@ -2828,7 +2828,7 @@ def process_set_as_enumerate(
             for _, f_arg in arg_subj.args.items()
         )
     ):
-        # Enumeration of a non-aggregate SET-returning function
+        # Enumeration of a non-aggregate function
         rvars = process_set_as_func_enumerate(ir_set, ctx=ctx)
     else:
         rvars = process_set_as_simple_enumerate(ir_set, ctx=ctx)
--- a/edb/pgsql/compiler/relgen.py
+++ b/edb/pgsql/compiler/relgen.py
@@ -2823,9 +2823,12 @@ def process_set_as_enumerate(
                 or arg_expr.limit
                 or arg_expr.offset
             )
+        ) and not any(
+            f_arg.param_typemod == qltypes.TypeModifier.SetOfType
+            for _, f_arg in arg_subj.args.items()
         )
     ):
-        # Enumeration of a SET-returning function
+        # Enumeration of a non-aggregate SET-returning function
         rvars = process_set_as_func_enumerate(ir_set, ctx=ctx)
     else:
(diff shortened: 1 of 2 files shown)


msullivan (commented):
I'm a little confused; none of those aggregate functions are SET returning?

dnwpark:
@msullivan the error is generated by postgres for putting aggregate functions in `ROWS FROM`, where postgres defines aggregate functions as taking a set of inputs.

We do have a few functions which take and return sets, such as `assert_exists`, but they special case functions and don't end up using `process_set_as_func_enumerate`.

dnwpark:
The SQL generated when doing `select enumerate(count((select Person)));`:

```pgsql
FROM LATERAL (
    FROM LATERAL (
        FROM LATERAL (
            FROM LATERAL (
                FROM LATERAL (
                    FROM edgedbpub."default::Person" AS "Person~4"
                    SELECT
                        "Person~4".id AS "Person_value~1"
                ) AS "Person~5"
                SELECT
                    "Person~5"."Person_value~1" AS "Person_value~2"
            ) AS "Person~6",
            LATERAL ROWS FROM (count("Person~6"."Person_value~2")) WITH ORDINALITY  AS "f~1"("v~1", _i)
            SELECT
                ("f~1"._i - 1) AS "0_value~1",
                "f~1"."v~1" AS "1_value~1"
        ) AS "q~1"
        SELECT
            "q~1"."0_value~1" AS "0_value~2",
            "q~1"."1_value~1" AS "1_value~2"
    ) AS "expr-11~2"
    SELECT
        "expr-11~2"."0_value~2" AS "0_value~3",
        "expr-11~2"."1_value~2" AS "1_value~3"
) AS "expr-12~2"
SELECT
    (("expr-12~2"."0_value~3", "expr-12~2"."1_value~3"))::edgedbpub."tuple<std|int64, std|int64>_t" AS "tuple_serialized~1"
LIMIT (
    SELECT
        (101)::int8 AS "expr~13_value~1"
)
```

edb/pgsql/compiler/relgen.py
```diff
@@ -2823,9 +2823,12 @@ def process_set_as_enumerate(
                 or arg_expr.limit
                 or arg_expr.offset
             )
+        ) and not any(
+            f_arg.param_typemod == qltypes.TypeModifier.SetOfType
+            for _, f_arg in arg_subj.args.items()
         )
     ):
-        # Enumeration of a SET-returning function
+        # Enumeration of a non-aggregate SET-returning function
```
msullivan, line 2831: OK, I think the problem is that this comment is wrong. SET-returning has nothing to do with it.

msullivan (approved):

PR #7991, opened 2024-11-13
Initial support for SQL over native protocol

elprans:
The binary protocol version is bumped to 3.0 to accommodate the addition
of the new `input_language` field to the `Parse` and `Execute` messages.
The field can be set to `b'E'` for EdgeQL or to `b'S'` for SQL.  The
result of a query sent as SQL is always encoded as a free object.  In
other regards the SQL execution and caching pipeline is identical to
EdgeQL.

Caveats and limitations:

Not all query features are supported when in SQL mode, specifically
cardinality assertions and output format are always `MANY` and `BINARY`
correspondingly.

DDL and explicit prepared statements are unsupported.

Multi-statement queries are not supported yet.


edb/protocol/messages.py
```diff
@@ -488,6 +488,12 @@ def dump(self) -> bytes:
 ###############################################################################
 
 
+class InputLanguage(enum.Enum):
+
+    EDGEQL = 0x45
+    SQL = 0x53
```
1st1, line 494:
```suggestion
    EDGEQL = 0x45  # 'E'
    SQL = 0x53  # 'S'
```

edb/pgsql/parser/parser.pyx
```diff
@@ -49,3 +83,261 @@ def pg_parse(query) -> str:
...
+        normalized_query = result.normalized_query.decode('utf8')
+        consts = []
+        for i in range(result.clocations_count):
+            loc = result.clocations[i]
+            if loc.length != -1:
+                if loc.param_id < 0:
+                    param_id = (
+                        abs(loc.param_id)
+                        + result.highest_extern_param_id
+                    )
```
1st1, line 137: Can you add a comment explaining what's going on here?

edb/server/compiler/compiler.py
```diff
@@ -2343,6 +2364,129 @@ def compile(
...
+            intro_sql = None
+        if isinstance(sql_unit.command_complete_tag, dbstate.TagPlain):
+            status = sql_unit.command_complete_tag.tag
+        elif isinstance(
+            sql_unit.command_complete_tag,
+            (dbstate.TagCountMessages, dbstate.TagUnpackRow),
+        ):
+            status = sql_unit.command_complete_tag.prefix.encode("utf-8")
+        elif sql_unit.command_complete_tag is None:
+            status = b"SELECT"  # XXX
```
1st1, line 2431: Flagging your XXX

1st1 (approved):
Epic stuff. I've left a couple nit comments, but really looks good to go.

edb/server/protocol/args_ser.pyx
```diff
@@ -155,11 +152,36 @@ cdef WriteBuffer recode_bind_args(
...
-    num_args += _count_globals(qug)
+    num_globals = _count_globals(qug)
+    num_args += num_globals
 
     if live:
+        if not compiled.extra_formatted_as_text:
+            # all parameter values are in binary
+            out_buf.write_int32(0x00010001)
+        elif not recv_args and not num_globals:
+            # all parameter values are in text (i.e extracted constants)
```
msullivan, line 163: Extracted *SQL* constants

msullivan (approved):
Looks good, but absolutely needs tests.

1st1:
I'm working on adding `querySQL()` to the JS driver, encountered this weird thing while writing tests:

```
  ● querySQL

    EdgeQLSyntaxError: Unexpected '::'
       |
     1 | select 1 + $1::int
       |              ^^
```

Full test case (just to show that my implementation does send SQL and works without query arguments):

```js
  test("querySQL", async () => {
    let client = getClient();

    try {
      let res = await client.querySQL("select 1");
      expect(JSON.stringify(res)).toEqual("[{\"col~1\":1}]");

      res = await client.querySQL("select 1 AS foo, 2 AS bar");
      expect(JSON.stringify(res)).toEqual("[{\"foo\":1,\"bar\":2}]");

      res = await client.querySQL("select 1 + $1::int", [41]);
      expect(JSON.stringify(res)).toEqual("[{\"col~1\":1}]");
    } finally {
      await client.close();
    }
  });
```

elprans:
> Two problems here, first -- "EdgeQLSyntaxError"

Errors are translated, we have no separate tree for pure SQL errors

> second -- the error is weird, might be something related to constant extraction?

Looks like probably something on your end, because it works for me with Python:

```python
import edgedb

client = edgedb.create_client("_localdev")
print(client.query_sql("select 1 + $1::int", 10))
```

```shell
$ python 1.py
[Object{col~1 := 11}]
```

edb/server/compiler/rpc.pyx
```diff
@@ -339,19 +399,18 @@ cdef _deserialize_comp_req_v0(
...
 
-    req.serialized_cache = data
-    req.cache_key = cache_key
-
     return req
 
 
-cdef _serialize_comp_req_v0(req: CompilationRequest):
+cdef _serialize_comp_req(req: CompilationRequest):
     # Please see _deserialize_v0 for the format doc
```
fantix, line 410:
```suggestion
    # Please see _deserialize_comp_req_v1() for the format doc
```

edb/server/compiler/rpc.pyx
```diff
@@ -210,17 +239,44 @@ cdef class CompilationRequest:
...
+        ReadBuffer buf = ReadBuffer.new_message_parser(data)
+        CompilationRequest req
+
+    if data[0] == 1:
+        req = _deserialize_comp_req_v1(
+            buf, query_text, compilation_config_serializer)
+    else:
+        raise errors.UnsupportedProtocolVersionError(
+            f"unsupported compile cache: version {data[0]}"
         )
```
fantix, line 270: I think this would fail in `_introspect_db()` -> `hydrate_cache()` after instance upgrade with v0 cache. Would it be possible to still load v0 cache entries and upgrade to v1 cache entries?

fantix (approved):

edb/server/compiler/rpc.pyx
elprans: This only applies to in-place upgrades and you can only upgrade in-place from a v1 cache instance no?

fantix: ok I see

edb/pgsql/resolver/__init__.py
```diff
@@ -56,6 +62,10 @@ def resolve(
 
     if debug.flags.sql_input:
         debug.header('SQL Input')
+
+        import edb.common.debug
+        edb.common.debug.dump(query)
```
aljazerzen, line 67:
```suggestion
        debug.dump(query)
```

aljazerzen (approved):

PR #7992, opened 2024-11-14
edb test: In verbose mode, print message when test starts

msullivan:


elprans:
There's a more elaborate implementation in #7991

PR #7995, opened 2024-11-15
Improve unsupported error message over SQL adapter

aljazerzen:


elprans (approved):
Better than it was, but I'd still prefer if we eventually made a proper dispatch over statement AST like we do for status tags for EdgeQL statements.

PR #7996, opened 2024-11-15
Prevent duplicate column names over SQL adapter

aljazerzen:
SQL over edgedb protocol requires unique column names in query results.

In PostgreSQL, duplicate column names are possible in 3 cases:

- explicit column names: `SELECT 1 as a, 2 as a;`
  This PR adds an error in this case (but only when SQL is being resolved for EdgeDB protocol)

- duplicate names are inferred: `SELECT a.id, b.id FROM a, b;`
  I've already removed this behavior and disabled column name inference if it would result in duplicate names. This was needed for DML commands, so results of such queries can be used in later SQL.

- duplicate names result from wildcard includes:

  ```
  WITH
    x(a) AS (VALUES (1)),
    y(a) AS (VALUES (2))
  SELECT * FROM x, y
  ```
  This PR makes it so inferred names in this case are `a` and `ya`.


tests/test_sql_query.py
```diff
@@ -871,6 +871,53 @@ async def test_sql_query_44(self):
...
+              x(a) AS (VALUES (1)),
+              y(a) AS (VALUES (2)),
+              z(a) AS (VALUES (3))
+            SELECT * FROM x, y JOIN z u on TRUE
+            '''
+        )
+
+        # `a` would be duplicated,
+        # so second and third instance are prefixed with rel var name
+        self.assert_shape(res, 1, ['a', 'ya', 'ua'])
```
aljazerzen, line 895: Do we want `xa`, `ya` and `ua` instead?

(it was a bit easier to do it this way)

tests/test_sql_query.py
```diff
@@ -871,6 +871,53 @@ async def test_sql_query_44(self):
...
+            '''
+            WITH
+              x(a) AS (VALUES (1)),
+              y(a) AS (VALUES (2), (3))
+            SELECT * FROM x, y, y
+            '''
+        )
+
+        # duplicate rel var names can yield duplicate column names
+        self.assert_shape(res, 4, ['a', 'ya', 'ya'])
```
aljazerzen, line 919: We probably don't want this behavior, but I'm not sure what we'd want instead.

Error out? This might break some existing SQL tools / user queries.

Emit `y1a`, `y2a` instead?

edb/server/compiler/sql.py
```diff
@@ -314,6 +314,7 @@ def resolve_query(
         search_path=search_path,
         allow_user_specified_id=allow_user_specified_id,
         apply_access_policies=apply_access_policies,
+        for_edgedb_protocol=True,
```
aljazerzen, line 317: @elprans Do pass this information to the compiler somewhere already?

tests/test_sql_query.py
elprans: I think we want an underscore, so `y_a`, `u_a` etc.

tests/test_sql_query.py
elprans: It's unlikely that tooling relies on colnames in those cases as opposed to col indexes, so emit fully-disambiguated names always.

aljazerzen: If we error out, that might break tools that were looking into results based on index.

aljazerzen:
My question have been resolved.

This can now be merged, but it will cause merge conflicts with #7991. I can go and fix those.

Also, all tests added in this PR should also be added to be executed over edgedb-protocol in #7991. I've written comments about what's the expected behavior.

elprans (approved):

edb/pgsql/resolver/expr.py
```diff
@@ -75,7 +75,20 @@ def infer_alias(res_target: pgast.ResTarget) -> Optional[str]:
...
+    ctx: Context,
+) -> Tuple[Sequence[pgast.ResTarget], Sequence[context.Column]]:
+    targets, columns = _resolve_ResTarget(
+        res_target, existing_names=existing_names, ctx=ctx
+    )
+
+    return (targets, columns)
+
+
+def _resolve_ResTarget(
```
msullivan, line 88: Why this extra level of indirection?

edb/pgsql/resolver/expr.py
```diff
@@ -87,11 +100,38 @@ def resolve_ResTarget(
...
         columns = []
         for table, column in col_res:
-            columns.append(column)
+            val = resolve_column_kind(table, column.kind, ctx=ctx)
+
+            # make sure name is not duplicated
+            # this behavior is technically different then Postgres, but EdgeDB
+            # protocol does not support duplicate names. And we doubt that
+            # anyone is depending on original behavior.
+            nam: str = column.name
```
msullivan, line 109: `nam` for `name` is a more aggressive abbreviation than I would normally do :P

msullivan (approved):

edb/pgsql/resolver/expr.py
aljazerzen: That's because variable `name` is already used down the function and python doesn't know about "block scopes". It does have the same type (str), but mypy was complaining about variable being re-defined.

edb/pgsql/resolver/expr.py
aljazerzen: Initially i thought I was gonna do validation after the call of the inner function, but I've later realized that validation depends on data I discard during resolution, so I had to do validation during resolution.

So this is a leftover.

PR #8002, opened 2024-11-18
Rework the async communication with Rust

mmastrac:
Avoid using the `asyncio.StreamReader` and collect the functionality of the pipe in a new `RustAsyncChannel`.

In addition, apply a number of correctness fixes as suggested by @1st1


msullivan:
Hm, looks like it didn't fix the hang

msullivan (approved):

PR #8008, opened 2024-11-19
Bump pyo3 and tokio

mmastrac:


aljazerzen (approved):
Looks good.

I'm having a deja vu of updating to a recent version of pyo3. I guess that was to 0.22.

PR #8011, opened 2024-11-20
Initial take at new documentation for path factoring

msullivan:
Please give feedback.

### DIFF
--- a/docs/edgeql/path_resolution.rst
+++ b/docs/edgeql/path_resolution.rst
@@ -1,8 +1,183 @@
 .. _ref_eql_path_resolution:
 
-===============
-Path resolution
-===============
+============
+Path scoping
+============
+
+Beginning with EdgeDB 6.0, we are phasing out our historical (and
+somewhat notorious)
+:ref:`"path scoping" algorithm <ref_eql_old_path_resolution>`
+in favor of a much simpler algorithm that nevertheless behaves
+identically on *most* idiomatic EdgeQL queries.
+
+EdgeDB 6.0 will contain features to support migration to and testing
+of the new semantics.  We expect the migration to be relatively
+painless for most users.
+
+Discussion of rationale for this change is available in
+`the RFC <rfc_>`_.
+
+
+New path scoping
+----------------
+
+When applying a shape to a path (or to a path that has shapes applied
+to it already), the path will be be bound inside computed
+pointers in that shape:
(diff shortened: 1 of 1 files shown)


docs/edgeql/path_resolution.rst
```diff
@@ -1,8 +1,183 @@
...
+However, when a path is used multiple times in "sibling" contexts,
+a cross-product will be computed:
+
+.. code-block:: edgeql-repl
+
+    db> select User.first_name ++ ' ' ++ User.last_name;
+    {'Peter Parker', 'Peter Stark', 'Tony Parker', 'Tony Stark'}
+
+
+If the old behavior is desired, you can typically rewrite the query
```
dnwpark, line 58: Maybe phrase this as, `If the "sibling" contexts should be bound together, you can typically ..... This was the old behaviour.`

docs/edgeql/path_resolution.rst
```diff
@@ -1,8 +1,183 @@
...
+
+Discussion of rationale for this change is available in
+`the RFC <rfc_>`_.
+
+
+New path scoping
+----------------
+
+When applying a shape to a path (or to a path that has shapes applied
+to it already), the path will be be bound inside computed
```
dnwpark, line 25: Maybe needs a more precise definition of `bound`.

aljazerzen (approved):
This reads well for me, but I'm not confident that average edgedb user will understand the change in behavior. That said, they don't need to - the upgrade plan is the only really important for them.

Maybe add a TLDR at the top?

docs/edgeql/path_resolution.rst
```diff
@@ -1,8 +1,183 @@
...
+.. code-block:: edgeql-repl
+
+    db> select User {
+    ...   name := User.first_name ++ ' ' ++ User.last_name
+    ... }
+    {User {name: 'Peter Parker'}, User {name: 'Tony Stark'}}
+
+
+When doing ``SELECT``, ``UPDATE``, or ``DELETE``, if the subject is a
+path with zero or more shapes applied to it, the path will be
```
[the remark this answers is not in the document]
msullivan: Yeah, zero or more!

Maybe instead I should say "is a path (or a path with shapes applied to it)"?
Or "is a path, optionally with shapes applied to it,"?

PR #8012, opened 2024-11-20
Fix ISE when taking intersection of types with pointers of the same name

### LINKED ISSUE #5846: Using type intersection with computed fails with `InternalServerError: can't lookup computable ptrref`
- EdgeDB Version: 3.2+bb17a05
- EdgeDB CLI Version: 3.5.0-dev.1025+21b3681
- OS Version: Windows 11 (WSL)

Steps to Reproduce:

1. Run the follwing query:
```
select Dog {
  uppername := [is Cat].name_upper
}
```
2. Server returns: `InternalServerError: can't lookup computable ptrref`

Same problem when using a computed link, although it fails with a different error: `InternalServerError: relation "edgedbpub.e6ce820e-2bb9-11ee-90ee-29650b039dc4_t" does not exist`

Schema:
```
module default {
    abstract type Pet {
        name: str;
        property name_upper := std::str_upper(.name);
    }

    type Cat extending Pet {
    }

    type Dog extending Pet {
    }
}
```

dnwpark:
Such intersections produce a proper error if:
- The pointer is computed but do not come from a common base type
- Do not have the same cardinality

close #5846


msullivan (approved):
Looks good!
Great testing work

PR #8014, opened 2024-11-21
Fix introspection over SQL adapter

aljazerzen:


fantix (approved):

aljazerzen:
I've ran "Test PostgreSQL versions" action on this PR and it is not reporting "column pi.indnullsnotdistinct does not exist" anymore.

https://github.com/edgedb/edgedb/actions/runs/11955740478

It does report errors about vector extension versions, which have to be fixed separately.

PR #8022, opened 2024-11-22
Allow OAuth callback to be specified directly

scotttrinh:
Currently, we expect that the OAuth Identity Provider should always redirect back to the auth extension's server endpoint, so we build this URL ourselves. However, there might be times when users want to control the OAuth flow themselves, so the callback should redirect to some URL that they've specified.

One confusing thing here is that sometimes the URL provided at `redirect_to` might be refered to as the "callback" URL, but that URL specifies where the auth extension server should call the application at the end of the flow.

### DIFF
--- a/edb/server/protocol/auth_ext/http.py
+++ b/edb/server/protocol/auth_ext/http.py
@@ -135,27 +135,27 @@ class Router:
 
         try:
             match args:
-                # API routes
+                # PKCE token exchange route
+                case ("token",):
+                    await self.handle_token(request, response)
+
+                # OAuth routes
                 case ("authorize",):
                     await self.handle_authorize(request, response)
                 case ("callback",):
                     await self.handle_callback(request, response)
-                case ("token",):
-                    await self.handle_token(request, response)
+
+                # Email/password routes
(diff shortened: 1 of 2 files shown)


fantix:
Oh I thought such `redirect_url` is configured inside the server (like cfg::Provider.redirect_url or something?). Would a URL parameter allow malicious 3rd parties to kidnap customers somehow?

scotttrinh:
> Oh I thought such redirect_url is configured inside the server (like cfg::Provider.redirect_url or something?).

No, we don't configure any URLs in the config since multiple apps could use the same auth extension (like web and mobile). The built-in UI is the only exception here since it doesn't have a way of configuring multiple options (unless we allowed configuring them via the built-in UIs URL 🤔 😆 )

> Would a URL parameter allow malicious 3rd parties to kidnap customers somehow?

Luckily, we added the allowed URLs thing, so it would only be possible to redirect to one of those URLs. FWIW, we allow specifying other URLs (`redirect_to`, `verify_url`, etc).

fantix (approved):
I see, all good :+1:

PR #8025, opened 2024-11-22
Refactor QL AST

aljazerzen:
A few minor changes to the AST, which should help with our work on grammar.

I suggest review commit-by-commit.

- **refactor: _Optional**
- **mark a few nodes as abstract**
- **refactor: make DDLCommand not extend Command**
- **refactor: Alias**
- **refactor: remove schema_class**
- **remove __rust_ignore__**

### DIFF
--- a/edb/edgeql/ast.py
+++ b/edb/edgeql/ast.py
@@ -154,12 +154,16 @@ class SortExpr(Base):
     nones_order: typing.Optional[NonesOrder] = None
 
 
-class AliasedExpr(Base):
+class Alias(Base):
+    __abstract_node__ = True
+
+
+class AliasedExpr(Alias):
     alias: str
     expr: Expr
 
 
-class ModuleAliasDecl(Base):
+class ModuleAliasDecl(Alias):
     module: str
     alias: typing.Optional[str]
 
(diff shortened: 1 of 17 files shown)


edb/edgeql/ast.py
```diff
@@ -524,7 +520,10 @@ class SelectQuery(Query):
 
 
 class GroupingIdentList(Base):
-    elements: typing.Tuple[GroupingAtom, ...]
+    elements: typing.Tuple[
+        typing.Union[ObjectRef, Path, GroupingIdentList],
```
msullivan, line 523: Why expand this out? That seems worse

edb/edgeql/ast.py
```diff
@@ -154,12 +154,16 @@ class SortExpr(Base):
     nones_order: typing.Optional[NonesOrder] = None
 
 
-class AliasedExpr(Base):
+class Alias(Base):
```
msullivan, line 156: Do we want a more descriptive name for `Alias`?

msullivan (approved):

edb/edgeql/ast.py
aljazerzen: What about variable? With binding binds multiple "variables"?

It has to be something that encapsulated both "a variable that is an expression" (`AliasedExpr`) and "module scope" (`ModuleAliasDecl`).

aljazerzen: We can change this later.

PR #8032, opened 2024-11-22
Rust project refactoring

mmastrac:
This moves all PyO3 modules into a single binary, reducing the overall memory footprint of Rust modules, allowing multiple modules to share .text sections.

Modules are exported via a single `.so` living in `edb/server/_rust_native`. 

As the module loads, it places the submodules in `sys.modules` directly, allowing consumers to directly import from `edb.server._rust_native._MODULE`.

## Logging

As part of this work, the custom per-project logging has been replaced with one single logger integration. This logger avoids asyncio and creates a blocking i/o thread using a unix dgram socket to send messages internally. 

Logging may happen around Python shutdown which PyO3 doesn't handle gracefully (unhappily failing to acquire the GIL in some cases), so a blocking thread avoids any complexity around that, as well as any async I/O shutdown issues.

Log messages that are not enabled (ie: more fine-grained than `info!` in Rust) are filtered out in the Rust log subscriber, and are effectively free.


aljazerzen (approved):
Looks good. I've not read the logger impl fully, but I concur on the approach you've outlined in the comment.

Should this refactoring also include the parser? It is used by the compiler process, so it would not benefit from merging binaries.

mmastrac:
> Looks good. I've not read the logger impl fully, but I concur on the approach you've outlined in the comment.
> 
> Should this refactoring also include the parser? It is used by the compiler process, so it would not benefit from merging binaries.

I think we can move the parser, but it doesn't make sense to include it in the mega-binary I believe at this time. It _might_ be useful to share pages between the compiler and server but I don't think it's a pressing need right now.

PR #8037, opened 2024-11-25
Replace headers with annotations in Parse/Execute

fantix:
Take annotations (`dict[str, str]`) instead of headers (`dict[int, bytes]`) in Parse/Execute messages.

We forgot this in protocol 1.0 but had been ignoring headers, so it was fine until we're trying to make use of annotations now. This change is a part of the binary protocol version 3.0 (EdgeDB 6.0a1 or nightlies before this PR will not work properly with annotations).

### DIFF
--- a/docs/reference/protocol/messages.rst
+++ b/docs/reference/protocol/messages.rst
@@ -114,7 +114,7 @@ Format:
 See the :ref:`list of error codes <ref_protocol_error_codes>` for all possible
 error codes.
 
-Known headers:
+Known attributes:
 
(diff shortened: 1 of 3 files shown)


elprans (commented):
There are more messages with Annotations in them, please check the protocol docs.

elprans (approved):

msullivan (approved):

fantix:
I'm adding the `to-backport-5.x` tag so that we can render the Protocol 2.0 docs correctly reflecting this fix; it's marginal though because annotations were not in use until Protocol 3.0.

PR #8038, opened 2024-11-26
Implement EdgeDB server protocol in pgrust (currently unused)

mmastrac:
Partial work towards landing the Rust frontend for EdgeDB:

 - Lands the prototype EdgeDB protocol + server implementation
 - Common server auth state machines are extracted to a new `ServerAuth` in `gel_auth::handshake`
 - Splits the auth library into "gel_auth" which we can use to replace some Cython code in a later patch
 - Cleans up unused deps for pgrust (and a few others scattered around)

This work was extracted from a branch implementing a Rust frontend (not included here)


rust/pgrust/src/protocol/edgedb.rs
fantix, line 1: * It's a bit weird that we have the EdgeDB binary protocol under `pgrust` project.
* We need to find a way to support both protocol v2.0 and v3.0 at the same time, at least for a while when people are connecting with old CLI/clients.

mmastrac: Agreed, I will likely need to separate this in the future. For now, they share the codegen from the Postgres protocol and this will also need to be extracted.

This protocol is also 2.0 for now, but I'll need to ensure both are implemented.

PR #8039, opened 2024-11-26
Changes Dump to use flags instead of headers in Protocol 3.0

fantix:
Refs #6106

The test is failing because the CLI doesn't support it, added in edgedb/edgedb-rust#365


msullivan (approved):
Sure, but why do we need a new dump message?

PR #8041, opened 2024-11-26
Support macos with new rust structure

scotttrinh:
See: https://pyo3.rs/v0.14.4/building_and_distribution.html#macos


elprans:
The docs say that `setuptools-rust` is supposed to do this automatically. I wonder why that doesn't work.

scotttrinh:
Yeah it certainly seems to be doing the right thing here: https://github.com/PyO3/setuptools-rust/blob/1e7c7d3b62a03fe60b54f3f25e385c2bbb8e2de6/setuptools_rust/build.py#L204

elprans:
The problem is that the Rust restructure turned extension into _submodules_, so those flags never reach those.  We can probably get away with setting the linker flags in global `RUSTFLAGS`, but I'm not 100% sure of the side-effects of that.  @mmastrac, we need your input here.

scotttrinh:
Closed in favor of #8042

PR #8046, opened 2024-11-27
Bump edgedb-python to 3.0.0b4

msullivan:
I incorrectly thought that edgedb/edgedb-python#552 was going to
require matching changes to testbase/connection.py, so I did an
edgedb-python release so we could easily upgrade.

It *didn't* require changes, and this was all a waste of time.
Oh well.


fantix (approved):
Though, I can maybe write annotation tests on top of this.

PR #8047, opened 2024-11-28
Add a test for renaming property without dropping index

### LINKED ISSUE #8016: Support auto renaming indexes
When a property is renamed the migration system should, ideally, offer to rename the index, not prompt to drop/recreate it.

```
did you drop index '__::idx' of object type 'default::XXX'? [y,n,l,c,b,s,q,?]
> n
edgedb error: SchemaError: index of object type 'default::XXX' already exists
```

msullivan:
This does work now, though you need to say no to the index drop.  It
didn't work in 4.x, I think, but I'm not sure exactly what fixed it.

Closes #8016

### DIFF
--- a/tests/test_edgeql_data_migration.py
+++ b/tests/test_edgeql_data_migration.py
@@ -1646,6 +1646,37 @@ class TestEdgeQLDataMigration(EdgeQLDataMigrationTestCase):
             }
         })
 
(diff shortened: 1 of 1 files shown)


elprans:
Any way we can quash the index drop prompt?  The diff is probably mis-generating the coefficient.

aljazerzen (approved):
Nice.

PR #8055, opened 2024-12-02
Enhanced query protocol in pgrust (to replace pgcon)

mmastrac:
Support proper pipelining in the Client and the required message parsing to do so. This is not sufficient to replace `pgcon.py*` yet, but a big step towards that.

The `--example connect` now supports an `--extended` flag allowing us to send extended messages instead of `Query`.

This supports:

 - Non-extended `Query` messages
 - Extended query messages in a `Sync`-delimited pipeline (including mixing `Query` messages with those)
 - `COPY OUT` style statements in both `Query` and `Execute` (not `COPY IN` at this time, however)

Unsupported:

 - `COPY IN`
 - Async notifications and `ParameterStatus` messages are handled as `warn!` and not exposed yet

### DIFF
--- a/rust/pgrust/src/connection/flow.rs
+++ b/rust/pgrust/src/connection/flow.rs
@@ -1,0 +1,886 @@
+//! Postgres flow notes:
+//!
+//! https://www.postgresql.org/docs/current/protocol-flow.html#PROTOCOL-FLOW-PIPELINING
+//!
+//! Extended query messages Parse, Bind, Describe, Execute, Close put the server
+//! into a "skip-til-sync" mode when erroring. All messages other than Terminate (including
+//! those not part of the extended query protocol) are skipped until an explicit Sync message is received.
+//!
+//! Sync closes _implicit_ but not _explicit_ transactions.
+//!
+//! Both Query and Execute may return COPY responses rather than rows. In the case of Query,
+//! RowDescription + DataRow is replaced by CopyOutResponse + CopyData + CopyDone. In the case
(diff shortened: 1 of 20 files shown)


rust/pgrust/src/connection/queue.rs
```diff
@@ -0,0 +1,166 @@
+use std::future::Future;
+use std::ops::DerefMut;
+use std::pin::Pin;
+use std::task::{Context, Poll};
+
+/// A queue of futures that can be polled in order.
+///
+/// Only only future will be active at a time. If no futures are active, the
```
fantix, line 8:
```suggestion
/// Only one future will be active at a time. If no futures are active, the
```

rust/pgrust/src/connection/flow.rs
```diff
@@ -0,0 +1,1234 @@
...
+    fn is_sync(&self) -> bool {
+        false
+    }
+}
+
+pub(crate) struct SyncMessageHandler;
+
+impl MessageHandler for SyncMessageHandler {
+    fn handle(&mut self, message: Message) -> MessageResult {
+        if ReadyForQuery::try_new(&message).is_some() {
```
fantix, line 312: Take out the `ReadyForQuery.status` and store in the connection? Or even return as a result of `pipeline_sync()`

fantix (approved):

rust/pgrust/src/connection/flow.rs
mmastrac: Totally agree -- I will have to add this to a future revision of this API, but it absolutely will need to be present.

PR #8066, opened 2024-12-03
Tweak compilation cache for SQL queries

aljazerzen:
- we had `QueryUnit.cachable` set to False for sql-over-edgedb-protocol,
- we were using the hash of the whole `NormalizedSource` as cache key, which includes extracted constants, which defeats (most) of the purpose of caching

Commits:
- **fix NormalizedSource.variables**
- **compute cache_key from pg_query_fingerprint**
- **Revert "compute cache_key from pg_query_fingerprint"**
- **pgsql Source.cache_key**
- **more tests**

### DIFF
--- a/edb/pgsql/parser/parser.pyx
+++ b/edb/pgsql/parser/parser.pyx
@@ -34,7 +34,7 @@ from edb.server.pgproto.pgproto cimport (
     WriteBuffer,
 )
 
-from libc.stdint cimport int8_t, uint8_t, int32_t
+from libc.stdint cimport int8_t, uint8_t, int32_t, uint64_t
 
 
 cdef extern from "pg_query.h":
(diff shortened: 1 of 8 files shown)


msullivan:
Any idea what's going on with the test failure?

aljazerzen:
We were rejecting multi-statement queries in execute, after trying to fetch type descriptors from PostgreSQL.

edb/pgsql/parser/parser.pyx
```diff
@@ -145,6 +145,7 @@ def pg_normalize(query: str) -> NormalizedQuery:
             raise error
 
         normalized_query = result.normalized_query.decode('utf8')
+        print("normalized_query", normalized_query)
```
msullivan, line 148: Drop this

edb/testbase/server.py
```diff
@@ -401,6 +402,19 @@ def __getstate__(self):
...
+        self, measure_fn: typing.Callable[[], int | float],
+        expected_change: int | float
+    ):
+        before = measure_fn()
+        try:
+            yield
+        finally:
+            after = measure_fn()
+            change = after - before
+            self.assertEqual(expected_change, change)
```
msullivan, line 416: I like this

msullivan (approved):

PR #8068, opened 2024-12-03
ext::ai: Allow specifying underlying vector dimensions when creating an index

elprans:


msullivan (approved):

diksipav:
Wow, thank you!

edb/schema/indexes.py
```diff
@@ -1479,30 +1479,63 @@ def _copy_ext_ai_model_annotations(
...
+                    f"{model_dimensions} dimensions, "
+                    f"got {specified_dimensions}",
+                    span=self.span,
+                )
+            elif (
+                specified_dimensions != model_dimensions
+                and not supports_shortening
+            ):
+                raise errors.SchemaDefinitionError(
+                    f"{model_stype_vn} returns embeddings with over "
```
diksipav, line 1531: should here just be written: `returns embeddings with ... ` instead of `with over....`?

PR #8082, opened 2024-12-08
Optimize compilation in transactions

1st1:
Do not re-pickle state (which includes schemas and other heavy objects) on every query in transactions unless that state changes.


msullivan (approved):
I'm nervous about things being added without `get_state_key` being properly updated. Could you add some comments in places where it would be hard to miss that?

fantix (approved):

PR #8083, opened 2024-12-09
Expand SQL docs

aljazerzen:


docs/reference/sql_adapter.rst
```diff
@@ -300,17 +300,20 @@ construct is mapped to PostgreSQL schema:
...
+          SET "global default::username" TO 'Tom'``;
+          SHOW "global default::username";
+
+    - Access policies are applied to object type tables when setting
+      ``apply_access_policies_pg`` is set to ``true``.
+
+    - Mutation rewrites and triggers are applied to all DML commands.
 
 
 DML commands
```
aljazerzen, line 319: @scotttrinh How can I mark these three whole sections as "added in EdgeDB 6.0"?

scotttrinh: Like [this](https://github.com/edgedb/edgedb/blob/0e8d80100e3d7c104747f336a8b19991d6a607c3/docs/reference/edgeql/casts.rst?plain=1#L37)

```rst
.. versionadded:: 6.0
```

scotttrinh: Oh, you mean the whole sections, 🤔

scotttrinh: I think you'll just need to mark each section separately? @jaclarke is there any better way to do this?

aljazerzen: But wouldn't that leave the section header in the old versions?

aljazerzen: I guess I could move the sections into separate pages.

scotttrinh: > But wouldn't that leave the section header in the old versions?

I don't think so? I think it shows as a half-hidden field that you can expand so you can see the new feature even from older versions which might make you want to upgrade. Worth experimenting and seeing how it renders.

aljazerzen: How do I test that? I see that I can `make docs` to get the output HTML in `build/html`. But how to I set the `versionadded` variable?

scotttrinh: Ahh, you'll need to get the docs infra setup locally via the https://github.com/edgedb/edgedb.com repo, which I think is missing the setup instructions about linking to your local repos 🤔

docs/reference/sql_adapter.rst
```diff
@@ -300,54 +300,167 @@ construct is mapped to PostgreSQL schema:
...
+
+    In addition, there are the following EdgeDB-specific settings:
+
+    - settings prefixed with ``"global "`` set the values of globals.
+      Because SQL syntax allows only string, integer and float constants in
+      ``SET`` command, globals of other types such as ``datetime`` cannot be set
+      this way. Special handling is in place to enable setting UUID globals via
+      hex-encoded strings. To set globals of other types via SQL, it is
+      recommended to change the global to use one of the simple types instead,
+      and use appropriate casts where the global is used.
```
scotttrinh, line 374: Can we maybe provide an example here? Does SQL syntax not support boolean? Maybe we can show an example of a bool global?

docs/reference/sql_adapter.rst
```diff
@@ -300,54 +300,167 @@ construct is mapped to PostgreSQL schema:
...
+=============
+
+The adapter emulates introspection schemas of PostgreSQL: ``information_schema``
+and ``pg_catalog``.
+
+Both schemas are not perfectly emulated, since they are quite large and
+complicated stores of information, that also changed between versions of
+PostgreSQL.
+
+Because of that so some tools might show objects that are not queryable or might
```
scotttrinh, line 400:
```suggestion
Because of that some tools might show objects that are not queryable or might
```

docs/reference/sql_adapter.rst
```diff
@@ -300,54 +300,167 @@ construct is mapped to PostgreSQL schema:
...
+report problems when introspecting. In such cases, please report the problem on
+GitHub so be can track the incompatibility down.
```
scotttrinh, line 402:
```suggestion
GitHub so we can track the incompatibility down.
```

docs/reference/sql_adapter.rst
```diff
@@ -300,54 +300,167 @@ construct is mapped to PostgreSQL schema:
...
+
+    - it cannot be used on tables that represent object types with access
+      properties or links of such objects,
+    - it cannot be used on tables that represent object types that have child
+      types extending them.
+
+Query cache
+===========
+
+An SQL query is issued to EdgeDB, it is compiled an internal SQL query, which is
```
scotttrinh, line 430:
```suggestion
An SQL query is issued to EdgeDB, it is compiled to an internal SQL query, which is
```

docs/reference/sql_adapter.rst
```diff
@@ -300,54 +300,167 @@ construct is mapped to PostgreSQL schema:
...
+then issued to the backing PostgreSQL instance. The compiler query is then
+cached, so each following issue of the same query will not perform any
+compilation, but just pass trough to PostgreSQL.
```
scotttrinh, line 433:
```suggestion
compilation, but just pass through to PostgreSQL.
```

or maybe

```suggestion
compilation, but just pass through the compiled query.
```

docs/reference/sql_adapter.rst
aljazerzen: Yea, good idea.

aljazerzen: > Does SQL syntax not support boolean?

No. Strangely it does not. Even though there are a ton of settings that expect boolean values. But they just look for `'true'` or `'yes'`. Bizarre.

aljazerzen: Well, syntactically it cannot be a boolean. But for postgres's settings, they have a bunch of string settings that they allow: https://www.postgresql.org/docs/current/config-setting.html

I'll implement these for globals too.

aljazerzen: Ref #8131

scotttrinh: > No. Strangely it does not. Even though there are a ton of settings that expect boolean values. But they just look for 'true' or 'yes'. Bizarre.

Oh my god, I hate it. 🙈 

Thanks for the update!

aljazerzen:
Ping?

scotttrinh (approved):
This was pending and I forgot to submit! Sorry for the hold up! Just a minor typo.

beerose (approved):

PR #8085, opened 2024-12-09
Update supported llm models

diksipav:


edb/lib/ext/ai.edgeql
```diff
@@ -316,57 +316,63 @@ CREATE EXTENSION PACKAGE ai VERSION '1.0' {
...
-            ext::ai::embedding_model_max_output_dimensions := "1024";
+            ext::ai::model_provider := "builtin::openai";
+        alter annotation
+            ext::ai::text_gen_model_context_window := "128000";
     };
 
-    create abstract type ext::ai::MistralSmallModel
+    # Mistral models
+    create abstract type ext::ai::MistralLargeModel
         extending ext::ai::TextGenerationModel
```
aljazerzen, line 343: These are breaking changes. What's our policy on breaking `ext::ai`?

diksipav: oh what is a breaking change? I didn' t realize, I just wanted to update models

aljazerzen: For future reference: we very strongly strongly prefer not to add breaking changes into ext modules. Because this makes database dumps un-restorable on newer versions of EdgeDB.

diksipav: I believe I fixed this : ) Thank u a lot!

elprans (approved):

msullivan (changes requested):
The diff is too noisy to easily verify that nothing is deleted; I'm going to try to clean it up

msullivan (approved):

msullivan:
I reordered things so that it is clear in the diff that none of the old models from 5.x broke.

Note that some names are still changed, which is not necessarily great, and might break dumps from earlier 6.x dev versions. I'm not too worried about this in the context of the AI extension on alphas, though.

PR #8088, opened 2024-12-09
Update bundled PostgreSQL to 17.2

elprans:
This also includes the `ALTER TABLE` rewrite optimization for domains
with constraints, see edgedb/postgres#3


msullivan (approved):

aljazerzen:
Do I understand correctly that we will be shipping a modified version of postgres, but edgedb would still be compatible with the official builds of postgres?

elprans:
Yes.

PR #8090, opened 2024-12-10
Send dimensions & user args to embedding models

diksipav:
If provided, send dimensions and user in embedding requests.

### DIFF
--- a/edb/server/protocol/ai_ext.py
+++ b/edb/server/protocol/ai_ext.py
@@ -759,6 +759,7 @@ async def _generate_embeddings_params(
                         inputs=inputs,
                         token_count=batch_token_count,
                         shortening=shortening,
+                        user=None,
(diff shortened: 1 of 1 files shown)


edb/server/protocol/ai_ext.py
```diff
@@ -2225,6 +2232,9 @@ async def _handle_embeddings_request(
             raise TypeError(
                 'missing or empty required "model" value in request')
 
+        shortening = body.get("dimensions")
```
diksipav, line 2235: I'd like above on line 2225 to use ` inputs = body.get("inputs")` and to expect `inputs` from the clients, but this is a breaking change so probably we wanna leave it as `input`.

elprans: Sure, accept both, but throw an error when both are specified.

PR #8091, opened 2024-12-10
Add docd for AI provider & update AI docs

diksipav:


deepbuzin:
@diksipav is there any way to find out a full list of things that the provider package ships with, other than browsing the source code?

PR #8059, opened 2024-12-02
[WIP] Support TOML config file

fantix:
```toml
[cfg]
session_idle_timeout = "PT15S"
shared_buffers = "128MiB"
_pg_prepared_statement_cache_size = 10

[[cfg.email_providers]]
_tname = "cfg::SMTPProviderConfig"
name = "gmail"
sender = "alice@example.com"
port = 587
security = "STARTTLS"

[[cfg.email_providers]]
_tname = "cfg::SMTPProviderConfig"
name = "outlook"
sender = "bob@example.com"
timeout_per_email = "PT1M48S"
```

The single-tenant server takes a new `--config-file` argument that points to a TOML config file like above. At the same time, a multi-tenant server accepts a new key `"config-file"` in its per-tenant config JSON. Each tenant could define its own TOML config file.

The TOML config file itself is reloadable (defined by `--reload-config-files`), and the `config-file` path in a multitenant config JSON file is also reloadable. The reloaded values do not affect running queries or active transactions; they are only effective with new queries, transactions, or backend connections.

The config values loaded from the TOML config file are placed in a new config tier, "configuration file," which overrides "environment variable" (and lower tiers) and is overridden by (not-yet-implemented) "command line" (and higher tiers). The root section in the TOML must be `cfg` at this moment.

Implementation-wise, the TOML values are JSON-ified and stored in the `_edgecon_state` temp table, so that different servers on the same backend Postgres may define different values in their TOML config files while not affecting each other (multi-tenancy for the backend). On reload, all alive pgcons shall run a SQL to re-insert reloaded TOML values as soon as they become idle; new pgcons automatically get to insert the reloaded TOML values.

PR #8098, opened 2024-12-10
Add structured config compilation

fantix:
Structured config values are Python objects from environment variables or (future) TOML config files. The compilation re- assembles the objects into `ConfigOp` ASTs and uses the static evaluation mechanism to generate verified config values.

This supersedes #8059 as a less-hacky way to validate config values.

As a side-effect, such config sources now support specifying EdgeQL values directly in a double-braces, for example:

    env GEL_SERVER_CONFIG_cfg::session_idle_timeout="{{<duration>'8 minutes 32 seconds'}}"

Or even simpler:

    env GEL_SERVER_CONFIG_cfg::session_idle_timeout='8 minutes 32 seconds'

(because the typecast is included by default)

This allows setting nested config objects with INSERT statements too:

    env GEL_SERVER_CONFIG_cfg::email_providers="{{ { (insert cfg::SMTPProviderConfig { name := 'gmail'}) } }}"

Sample:

```python
ops = compile_structured_config(
    {
        "cfg::Config": {
            "singleprop": "value",
            "memprop": 512,
            "durprop": "16s",
            "enumprop": "One",
            "multiprop": ["v1", "v2", "v3"],
            "listen_port": 5,
            "email_providers": [
                {"_tname": "cfg::SMTPProviderConfig", "name": "gmail"},
                {"_tname": "cfg::SMTPProviderConfig", "name": "outlook"},
            ],
            "sysobj": [
                {
                    "name": "1",
                    "obj": {
                        "_tname": "cfg::Subclass1",
                        "name": "aa",
                        "sub1": "aa",
                    },
                },
                {
                    "name": "2",
                    "_tname": "cfg::TestInstanceConfigStatTypes",
                    "memprop": 128,
                },
            ],
        },
    },
    spec=spec,
    schema=stdlib.stdschema,
    source="config file",
)
print(ops)
```

```
{
 'cfg::Config': immutables.Map({
  'durprop': SettingValue(name='durprop', value=<statypes.Duration 'PT16S'>, source='config file', scope=<ConfigScope.INSTANCE: 'INSTANCE'>, secret=False),
  'email_providers': SettingValue(name='email_providers', value=frozenset({cfg::SMTPProviderConfig(timeout_per_email=<statypes.Duration 'PT1M'>, security='STARTTLSOrPlainText', sender=None, name='outlook', port=None, timeout_per_attempt=<statypes.Duration 'PT15S'>, validate_certs=True, password=None, host=None, username=None), cfg::SMTPProviderConfig(timeout_per_email=<statypes.Duration 'PT1M'>, security='STARTTLSOrPlainText', sender=None, name='gmail', port=None, timeout_per_attempt=<statypes.Duration 'PT15S'>, validate_certs=True, password=None, host=None, username=None)}), source='config file', scope=<ConfigScope.INSTANCE: 'INSTANCE'>, secret=False),
  'enumprop': SettingValue(name='enumprop', value='One', source='config file', scope=<ConfigScope.INSTANCE: 'INSTANCE'>, secret=False),
  'listen_port': SettingValue(name='listen_port', value=5, source='config file', scope=<ConfigScope.INSTANCE: 'INSTANCE'>, secret=False),
  'memprop': SettingValue(name='memprop', value=<statypes.ConfigMemory '512B'>, source='config file', scope=<ConfigScope.INSTANCE: 'INSTANCE'>, secret=False),
  'multiprop': SettingValue(name='multiprop', value=frozenset({'v3'}), source='config file', scope=<ConfigScope.INSTANCE: 'INSTANCE'>, secret=False),
  'singleprop': SettingValue(name='singleprop', value='value', source='config file', scope=<ConfigScope.INSTANCE: 'INSTANCE'>, secret=False),
  'sysobj': SettingValue(name='sysobj', value=frozenset({cfg::TestInstanceConfigStatTypes(obj=None, name='2', durprop=None, memprop=<statypes.ConfigMemory '128B'>), cfg::TestInstanceConfig(obj=cfg::Subclass1(sub1='aa', name='aa'), name='1')}), source='config file', scope=<ConfigScope.INSTANCE: 'INSTANCE'>, secret=False)
 })
}
```

### DIFF
--- a/edb/ir/staeval.py
+++ b/edb/ir/staeval.py
@@ -512,7 +512,8 @@ def python_cast_str(sval: str, pytype: type) -> Any:
             return False
         else:
             raise errors.InvalidValueError(
-                f"invalid input syntax for type bool: {sval!r}"
+                f"invalid input syntax for type bool: {sval!r}",
+                hint="bool value can only be one of: true, false"
             )
     else:
         return pytype(sval)
--- a/edb/ir/staeval.py
+++ b/edb/ir/staeval.py
@@ -34,6 +34,7 @@ from typing import (
 
 import decimal
 import functools
+import uuid
 
 import immutables
 
@@ -43,6 +44,7 @@ from edb import errors
 from edb.common import typeutils
 from edb.common import parsing
 from edb.common import uuidgen
+from edb.common import value_dispatch
 from edb.edgeql import ast as qlast
 from edb.edgeql import compiler as qlcompiler
 from edb.edgeql import qltypes
@@ -52,6 +54,7 @@ from edb.ir import typeutils as irtyputils
 from edb.ir import statypes as statypes
 from edb.ir import utils as irutils
 
+from edb.schema import name as sn
 from edb.schema import objects as s_obj
 from edb.schema import objtypes as s_objtypes
 from edb.schema import types as s_types
@@ -100,6 +103,40 @@ def evaluate_SelectStmt(
             'expression is not constant', span=ir_stmt.span)
 
 
+@evaluate.register(irast.InsertStmt)
+def evaluate_InsertStmt(
+    ir: irast.InsertStmt, schema: s_schema.Schema
+) -> EvaluationResult:
+    # XXX: raise for unsupported InsertStmt?
+
+    return irast.Tuple(
+        named=True,
+        typeref=ir.subject.typeref,
+        elements=[
+            irast.TupleElement(
+                name=ptr_set.expr.ptrref.shortname.name,
+                val=irast.Set(
+                    expr=evaluate(ptr_set.expr.expr, schema),
+                    typeref=ptr_set.typeref,
+                    path_id=ptr_set.path_id,
+                ),
+            )
+            for ptr_set, _ in ir.subject.shape
+            if ptr_set.expr.ptrref.shortname.name != "id"
+            and ptr_set.expr.expr is not None
+        ],
+    )
+
+
+@evaluate.register(irast.TypeIntrospection)
+def evaluate_TypeIntrospection(
+    ir: irast.TypeIntrospection, schema: s_schema.Schema
+) -> EvaluationResult:
+    return irast.StaticIntrospection(
+        named=True, ir=ir, schema=schema, elements=[], typeref=ir.typeref
+    )
+
+
 @evaluate.register(irast.TypeCast)
 def evaluate_TypeCast(
     ir_cast: irast.TypeCast, schema: s_schema.Schema
@@ -108,7 +145,7 @@ def evaluate_TypeCast(
     schema, from_type = irtyputils.ir_typeref_to_type(
         schema, ir_cast.from_type)
     schema, to_type = irtyputils.ir_typeref_to_type(
-        schema, ir_cast.from_type)
+        schema, ir_cast.to_type)
 
     if (
         not isinstance(from_type, s_scalars.ScalarType)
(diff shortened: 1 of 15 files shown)


elprans:
> Technically, this allows setting nested config objects with INSERT statements too.

Can you give an example of how that might look?

fantix:
This works:

    env GEL_SERVER_CONFIG_cfg::email_providers="{{ { (insert cfg::SMTPProviderConfig { name := 'gmail'}) } }}" edb server

tests/test_server_compiler.py
```diff
@@ -77,6 +84,142 @@ def test_server_compiler_compile_edgeql_script(self):
...
+        result = self.compiler.compile_structured_config(
+            {"cfg::Config": values}, source=source
+        )
+        rv = dict(result["cfg::Config"])
+        for name, setting in rv.items():
+            self.assertEqual(setting.name, name)
+            self.assertEqual(setting.scope, config.ConfigScope.INSTANCE)
+            self.assertEqual(setting.source, source)
+        self.assertDictEqual({k: v.value for k, v in rv.items()}, expected)
+        return rv
```
msullivan, line 103: Could we get an end-to-end test, also?

One thing I am concerned about is how INSERTed objects specified via structured config will interact with objects inserted manually via `CONFIGURE INSTANCE INSERT ...`.

The *best* option is for the sets of objects to get merged. I spent about a day trying to make that work, and came to the conclusion that it would probably take me closer to a week, and would be pretty hairy. This is why I introduced `EDGEDB_MAGIC_SMTP_CONFIG`, so we wouldn't need to handle it.

If manually inserted objects just fully override, that is probably acceptable behavior (though it wouldn't work for our SMTP config case, so we would still need EDGEDB_MAGIC_SMTP_CONFIG), but I think it will take a little care to make that work properly also. (If you INSERT an object and then RESET it away, probably the structured config one should come back? But I think as implemented it won't quite work.)

Anyway, I don't think this really *blocks* us, but I think we might need to disable object config for now, and stick with EDGEDB_MAGIC_SMTP_CONFIG for the case we need. (We should probably use the structured parser for EDGEDB_MAGIC_SMTP_CONFIG though!)
I think getting objects to work well enough to cover that case for us will be a decent amount of extra work, and not a priority yet.

msullivan (commented):
Looks good broadly, but I have some concerns we need to figure out about the config object situtation

tests/test_server_compiler.py
fantix: Right, I also left 2 similar questions in [the previous PR](https://github.com/edgedb/edgedb/pull/8059#issuecomment-2515142722)

edb/server/compiler/compiler.py
```diff
@@ -1290,6 +1291,26 @@ def validate_schema_equivalence(
...
         )
 
+    def compile_structured_config(
+        self,
+        objects: Mapping[str, config_compiler.ConfigObject],
+        source: str | None = None,
+    ) -> dict[str, immutables.Map[str, config.SettingValue]]:
+        # XXX: only config in the stdlib is supported currently, so the only
+        # key allowed in objects is "cfg::Config". API for future compatibility
+        if list(objects) != ["cfg::Config"]:
```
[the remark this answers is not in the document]
fantix: Yes? Though, it's a bit weird for the user to write `[cfg::AbstractConfig]` in the TOML config file.

Fwiw using `cfg::Config` here is following [this line](https://github.com/edgedb/edgedb/commit/1ccc83f7339d1c34d8b85c3cac7b18c8550f50d4#diff-5dadc787222c2b6f04979c96c03df710b04a9390596044aaa9919edd80674682R96) which was added before we even had `cfg::AbstractConfig` and never got updated.

edb/ir/staeval.py
```diff
@@ -100,6 +103,40 @@ def evaluate_SelectStmt(
...
 
+@evaluate.register(irast.InsertStmt)
+def evaluate_InsertStmt(
+    ir: irast.InsertStmt, schema: s_schema.Schema
+) -> EvaluationResult:
+    # XXX: raise for unsupported InsertStmt?
+
+    return irast.Tuple(
+        named=True,
+        typeref=ir.subject.typeref,
```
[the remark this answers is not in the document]
fantix: Fixed!

edb/server/compiler/compiler.py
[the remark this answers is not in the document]
msullivan: Sure, ok

tests/test_server_compiler.py
fantix: > We should probably use the structured parser for EDGEDB_MAGIC_SMTP_CONFIG

Fixed!

edb/server/main.py
```diff
@@ -244,18 +245,27 @@ async def _run_server(
...
-            compiler_state=compiler_state,
+            compiler_state=compiler.state,
             tenant=tenant,
             use_monitor_fs=args.reload_config_files in [
                 srvargs.ReloadTrigger.Default,
                 srvargs.ReloadTrigger.FileSystemEvent,
             ],
             net_worker_mode=args.net_worker_mode,
         )
+        magic_smtp = os.getenv('EDGEDB_MAGIC_SMTP_CONFIG')
```
msullivan, line 256: Did you do a manual test of this, since apparently I half-assed it and didn't write one

msullivan (approved):

edb/server/main.py
```diff
@@ -244,18 +245,27 @@ async def _run_server(
...
+        if magic_smtp:
+            magic_smtp = json.loads(magic_smtp)
+            if isinstance(magic_smtp, dict):
+                # for backward compatibility
+                magic_smtp = [magic_smtp]
```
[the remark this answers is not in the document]
fantix: True!

edb/server/main.py
fantix: Yes, manually tested ✅

PR #8100, opened 2024-12-10
Fix cache collision in normalized SQL queries

### LINKED ISSUE #8099: SQL-over-native protocol constant extraction + cache gets confused
```
    async def test_sql_native_query_21(self):
        await self.assert_sql_query_result(
            """
                SELECT 1
            """,
            [{'col~1': 1}],
        )

        await self.assert_sql_query_result(
            """
                SELECT 'test'
            """,
            [{'col~1': 'test'}],
        )
```
leads to
```
AssertionError: data shape differs: 1952805748 != 'test' PATH: [0]["col~1"]
shape: [{'col~1': 'test'}]
data: [{'col~1': 1952805748}]
```

(via @aljazerzen )

msullivan:
Currently the cache key is just based on the normalized text, which
means that queries that differ only in the types of extracted
constants will collide.

Include the type oids in the key.

Fixes #8099.

### DIFF
--- a/edb/pgsql/parser/parser.pyx
+++ b/edb/pgsql/parser/parser.pyx
@@ -236,11 +236,14 @@ cdef class Source:
     def original_text(self) -> str:
         return self._text
 
+    def _compute_cache_key(self) -> bytes:
+        h = hashlib.blake2b(self._tag().to_bytes())
+        h.update(bytes(self.text(), 'UTF-8'))
+        return h.digest()
+
     def cache_key(self) -> bytes:
         if not self._cache_key:
-            h = hashlib.blake2b(self._tag().to_bytes())
-            h.update(bytes(self.text(), 'UTF-8'))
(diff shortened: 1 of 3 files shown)


elprans (approved):

fantix (approved):

edb/pgsql/parser/parser.pyx
```diff
@@ -346,6 +349,13 @@ cdef class NormalizedSource(Source):
 
         return oids
 
+    def _compute_cache_key(self) -> bytes:
+        h = hashlib.blake2b(super()._compute_cache_key())
+        # Include the types of the extracted constants
+        for oid in self.extra_type_oids():
+            h.update(oid.to_bytes(length=2))
```
aljazerzen, line 356: oid is int32, this should be `to_bytes(length=4, signed=True)`

PR #8103, opened 2024-12-11
Fix cache key of source

aljazerzen:
Followup for #8100

### DIFF
--- a/edb/pgsql/parser/parser.pyx
+++ b/edb/pgsql/parser/parser.pyx
@@ -236,11 +236,6 @@ cdef class Source:
     def original_text(self) -> str:
         return self._text
 
-    def _compute_cache_key(self) -> bytes:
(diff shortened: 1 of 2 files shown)


edb/pgsql/parser/parser.pyx
```diff
@@ -349,13 +356,6 @@ cdef class NormalizedSource(Source):
 
         return oids
 
-    def _compute_cache_key(self) -> bytes:
```
msullivan, line 352: Drop it in the parent too

msullivan (approved):
Sure, but I don't really understand what was wrong?

msullivan:
> Sure, but I don't really understand what was wrong?

Oh, the length/signedness. Sure. I made it 2 bytes because all of the ints were coming from `PgLiteralTypeOID` above, and are all small and positive.

PR #8104, opened 2024-12-11
Disable normalization for SQL over EdgeDB proto

aljazerzen:
Currently, there are queries that work without normalization
but fail with normalization, so I'll disable it for now.


msullivan:
This is some really dumb github nonsense. It automatically closed this issue because of a mention in #8122, and now I can't reopen it because I force-pushed an update to this PR.

PR #8106, opened 2024-12-11
Update docs to cover modifying functions.

dnwpark:


docs/datamodel/functions.rst
```diff
@@ -32,6 +32,67 @@ This function accepts a :eql:type:`str` as an argument and produces a
...
+      }
+    );
+
+.. code-block:: edgeql-repl
+
+    db> select add_user('Jan') {name, joined_at};
+    {default::User {name: 'Jan', joined_at: <datetime>'2024-12-11T11:49:47Z'}}
+
+Unlike other functions, the arguments of modifying functions **must** have a
+:ref:`cardinality <_ref_reference_cardinality>` of ``One``.
```
scotttrinh, line 63: Practical question: if some property that you're trying to insert or update is `multi`, does that mean you'll have to pass it as an array and unpack it? Maybe we should show this as an example?

docs/datamodel/computeds.rst
```diff
@@ -53,9 +53,9 @@ field is referenced in a query.
 
 .. warning::
 
-  Volatile functions are not allowed in computed properties defined in schema.
-  This means that, for example, your schema-defined computed property cannot
-  call :eql:func:`datetime_current`, but it *can* call
+  Volatile and modifying functions are not allowed in computed properties
+  defined in schema. This means that, for example, your schema-defined computed
+  property cannot call :eql:func:`datetime_current`, but it *can* call
```
beerose, line 58: Maybe we can add a link to the corresponding docs for modyfing functions so that it's clear what we refer to?

docs/datamodel/functions.rst
```diff
@@ -32,6 +32,67 @@ This function accepts a :eql:type:`str` as an argument and produces a
...
+
+    db> select add_user({'Feb','Mar'});
+    edgedb error: QueryError: possibly more than one element passed into
+    modifying function
+    db> select add_user(<str>{});
+    edgedb error: QueryError: possibly an empty set passed as non-optional
+    argument into modifying function
+
+Optional arguments can still accept empty sets. For example, if ``add_user``
+was defined as
```
beerose, line 75:
```suggestion
was defined as:
```

docs/datamodel/functions.rst
```diff
@@ -32,6 +32,67 @@ This function accepts a :eql:type:`str` as an argument and produces a
...
+
+  function add_user(name: str, joined_at: optional datetime) -> User
+    using (
+      insert User {
+        name := name,
+        joined_at := joined_at ?? std::datetime_current(),
+      }
+    );
+
+then the following queries are valid:
```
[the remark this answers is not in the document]
dnwpark: Not entirely sure that's customary, since it's technically the continuation of the previous sentence.

docs/datamodel/computeds.rst
dnwpark: Linked the volatility reference page and changed it to refer to expressions in general.

docs/datamodel/functions.rst
dnwpark: I've added a more general section on passing in sets as arguments as well.

beerose (approved):

PR #8109, opened 2024-12-12
Implement a "SQL Row" type descriptor

1st1:
Couple of notes:

* A dedicated descriptor is needed to reduce complexity in client libraries, as the "record" type has a different API from the "free object" type. Adding multiple cases to the codecs pipeline to track what language was used -- EdgeQL or SQL -- is more error prone and time consuming than adding a dedicated descriptor.

* This commit uses a hack to call the compiler process to build the actual descriptor. The IO<->compiler RPC code is incredibly complex and brittle, so adding two new remote methods would require copy/pasting (as well as careful testing) of a lot of code that manages compiler state and related optimizations.

  It's easier to leave all of that code alone and multiplex the new method into the existing `compile()` and `compile_in_tx()`.

  An alternative would be quite an invasive refactoring of the entire compiler_pool package, something that I'd like to avoid, especially that close to the release.


edb/server/compiler/rpc.pyx
```diff
@@ -17,9 +17,12 @@
 #
 
 from typing import (
+    List,
```
elprans, line 20: Just use plain `tuple` and `list`

edb/server/compiler/compiler.py
```diff
@@ -707,6 +722,19 @@ def compile_in_tx(
     ) -> Tuple[
         dbstate.QueryUnitGroup, Optional[dbstate.CompilerConnectionState]
     ]:
+        if request.input_language is enums.InputLanguage.SQL_PARAMS:
+            tx = state.current_tx()
+            return (
```
elprans, line 727: This doesn't type check? Can we return a proper QueryUnitGroup instead?

1st1: I fixed all type checking issues.

1st1:
This PR depends on https://github.com/edgedb/edgedb-python/pull/560 and then a new version of edgedb-python

elprans (approved):

PR #8110, opened 2024-12-12
Fix bogus transaction error in SQL-over-native protocol

msullivan:


edb/server/protocol/binary.pyx
```diff
@@ -554,7 +554,11 @@ cdef class EdgeConnection(frontend.FrontendConnection):
                     )
             finally:
                 if suppress_timeout:
-                    await self._restore_tx_timeout(dbv)
+                    # Some failure modes abort the transaction
+                    try:
+                        await self._restore_tx_timeout(dbv)
+                    except pgerror.BackendError:
+                        pass
```
1st1, line 561: How about we only suppress it when `query_req.input_language is LANG_SQL`. Also please amend the comment with an explanation, that backend might get into a broken transaction state if `dbv.parse` sends a bogus SQL query to get its in/out types.

1st1 (approved):

PR #8116, opened 2024-12-13
Implement position mapping for SQL over native proto

### LINKED ISSUE #8077: Fix error spans for SQL queries with extracted constants
```
_localdev:scope[sql]> select name, 100000, asdf from "User";
error: QueryError: cannot find column `asdf`
  ┌─ <query>:1:18
  │
1 │ select name, 100000, asdf from "User";
  │                  ^ error

```

msullivan:
This is a little fiddly, since we need to account for query
normalization. We do this by traversing the pre and post normalization
ASTs in parallel, building up a map.  In my testing, postgres only
returned error positions for errors that could be detected while
describing the query, so to simplify the implementation we only remap
errors that happen there.

Also do a better job translating the errors:
 * Don't ISE when it doesn't map to an edgedb error
 * Preserve hint and details when we can

Fixes #8077.

### DIFF
--- a/edb/pgsql/codegen.py
+++ b/edb/pgsql/codegen.py
@@ -21,6 +21,7 @@ from __future__ import annotations
 
 from typing import Any, Optional, Sequence, List
 
+import abc
 import collections
 import dataclasses
 
@@ -113,21 +114,18 @@ def generate_ctes_source(
 
 
 class TranslationData:
+    @abc.abstractmethod
+    def translate(self, pos: int) -> int:
+        ...
+
+
+@dataclasses.dataclass(kw_only=True)
+class BaseTranslationData(TranslationData):
     source_start: int
     output_start: int
-    output_end: int
-    children: List[TranslationData]
-
-    def __init__(
-        self,
-        *,
-        source_start: int,
-        output_start: int,
-    ):
-        self.source_start = source_start
-        self.output_start = output_start
-        self.output_end = -1
-        self.children = []
+    output_end: int | None = None
+    children: List[BaseTranslationData] = (
+        dataclasses.field(default_factory=list))
 
     def translate(self, pos: int) -> int:
         bu = None
@@ -135,9 +133,20 @@ class TranslationData:
             if u.output_start >= pos:
                 break
             bu = u
-        if bu and bu.output_end > pos:
+        if bu and (bu.output_end is None or bu.output_end > pos):
             return bu.translate(pos)
         return self.source_start
+
+
+@dataclasses.dataclass
+class ChainedTranslationData(TranslationData):
+    parts: List[TranslationData] = (
+        dataclasses.field(default_factory=list))
+
+    def translate(self, pos: int) -> int:
+        for part in self.parts:
+            pos = part.translate(pos)
+        return pos
 
 
 @dataclasses.dataclass(frozen=True)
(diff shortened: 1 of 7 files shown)


edb/pgsql/codegen.py
```diff
@@ -113,33 +114,41 @@ def generate_ctes_source(
...
 
     def translate(self, pos: int) -> int:
         bu = None
         for u in self.children:
             if u.output_start >= pos:
                 break
             bu = u
-        if bu and bu.output_end > pos:
+        if bu and (bu.output_end is None or bu.output_end > pos):
             return bu.translate(pos)
```
aljazerzen, line 137: This could be a binary search. Might not be worth it.

edb/pgsql/codegen.py
```diff
@@ -113,33 +114,41 @@ def generate_ctes_source(
 
 
 class TranslationData:
```
aljazerzen, line 116: Can we rename this to something less generic? "Translation data" could be anything that compiler produces. I suggest "SourceMap".

edb/server/compiler/sql.py
```diff
@@ -109,9 +111,60 @@ def _try(q: str) -> List[dbstate.SQLQueryUnit]:
...
+        # HACK: I don't know why, but this - 1 helps a lot.
+        output_start=out.span.start - 1 if out.span else 0,
+    )
+    if type(src) != type(out):
+        return tdata
+    children = tdata.children
+    for (k1, v1), (k2, v2) in zip(ast.iter_fields(src), ast.iter_fields(out)):
+        assert k1 == k2
+        if k1 != k2:
+            continue
```
aljazerzen, line 141: This is probably not needed if you have assert above.

edb/server/compiler/sql.py
```diff
@@ -109,9 +111,60 @@ def _try(q: str) -> List[dbstate.SQLQueryUnit]:
...
+    This is used to translate SQL spans reported on a normalized query
+    to ones that make sense on the pre-normalization version.
+
+    Note that we only use this map for errors reported during the
+    "parse" phase, so we don't need to worry about it being reused
+    with different constants.
+    """
+    tdata = pg_codegen.BaseTranslationData(
+        source_start=src.span.start if src.span else 0,
+        # HACK: I don't know why, but this - 1 helps a lot.
```
aljazerzen, line 132: Postgres's "locations" are 1-based. Our spans are 0-based. Maybe this is why.

edb/server/compiler/sql.py
```diff
@@ -109,9 +111,60 @@ def _try(q: str) -> List[dbstate.SQLQueryUnit]:
...
+        output_start=out.span.start - 1 if out.span else 0,
+    )
+    if type(src) != type(out):
+        return tdata
```
aljazerzen, line 136: Maybe also check that one of the types is a `ParamRef`? Just for the case when in the future we do more aggressive normalization.

aljazerzen (approved):
Looks good.

elprans (approved):

edb/pgsql/codegen.py
msullivan: I'll add a TODO

edb/pgsql/codegen.py
msullivan: I'll do a separate clean PR for that rename

edb/server/compiler/sql.py
msullivan: I think we'll still probably want to map the spans then, though?

PR #8121, opened 2024-12-13
Add TOML config file support

### LINKED ISSUE #1325: Allow reading configuration from a local file
Need an option to read server configuration from a specified JSON file in the local filesystem.

### LINKED ISSUE #7990: Per-tenant configuration in multitenant
I think there are two components here:

- [ ] Per tenant "environmental" configuration, where we can do the equivalent of passing config in environment variables. I think we decided that a toml file was the plan, but I'm not sure we agreed on any specifics.
- [ ] `CONFIGURE INSTANCE`, which I think is currently disabled in multi-tenant

fantix:
Sample TOML config file:

```toml
["cfg::Config"]
session_idle_timeout = "1 minute"
shared_buffers = "128MiB"
_pg_prepared_statement_cache_size = 10
listen_port = 5858

[[magic_smtp_config]]
_tname = "cfg::SMTPProviderConfig"
name = "gmail"
sender = "alice@example.com"
port = 587
security = "STARTTLS"

[[magic_smtp_config]]
_tname = "cfg::SMTPProviderConfig"
name = "outlook"
sender = "bob@example.com"
timeout_per_email = "PT1M48S"
```

The single-tenant server takes a new --config-file argument that points to a TOML config file like above. At the same time, a multi-tenant server accepts a new key "config-file" in its per-tenant config JSON. Each tenant could define its own TOML config file.

The TOML config file itself is reloadable (defined by --reload-config-files), and the config-file path in a multitenant config JSON file is also reloadable. The reloaded values do not affect running queries or active transactions; they are only effective with new queries, transactions, or backend connections.

The config values loaded from the TOML config file are placed in a new config tier, "configuration file," which overrides "environment variable" (and lower tiers) and is overridden by (not-yet-implemented) "command line" (and higher tiers). The root section in the TOML must be cfg at this moment.

Implementation-wise, the TOML values are JSON-ified and stored in the _edgecon_state temp table, so that different servers on the same backend Postgres may define different values in their TOML config files while not affecting each other (multi-tenancy for the backend). On reload, all alive pgcons shall run a SQL to re-insert reloaded TOML values as soon as they become idle; new pgcons automatically get to insert the reloaded TOML values.

- [x] Deal with backend_setting after reload
- [x] Add tests

Replaces #8059, Fixes #1325, Fixes #7990


msullivan:
Add a backport label?

edb/server/tenant.py
```diff
@@ -965,11 +1025,26 @@ async def acquire_pgcon(self, dbname: str) -> pgcon.PGConnection:
...
         for _ in range(self._pg_pool.max_capacity):
             conn = await self._pg_pool.acquire(dbname)
-            if conn.is_healthy():
-                return conn
+            if not conn.is_healthy():
+                logger.warning("acquired an unhealthy pgcon; discard now")
+            elif conn.last_init_con_data is not self._init_con_data:
+                if self._update_init_con_sql:
+                    try:
+                        await conn.sql_execute(self._update_init_con_sql)
```
elprans, line 1033: Is there any way to avoid doing this every time?  Ideally we should track whether a connection has the updated state.  Maybe a dict with version counter or something like that?

fantix: `last_init_con_data` would be for "track whether a connection has the updated state"? The `is` reference check should be as fast as a version counter.

elprans: I'm a bit confused about the interaction between `_update_init_con_sql` and that check then. Can there be a case when the state is not matching but `_update_init_con_sql` is `None`?

fantix: Theoretically no - whenever we update the state, we set a new update SQL. Let me change the code a bit to ensure that it is always done "atomically" with a single function.

In the worst case when it's really happening, it prints a warning, discard that pgcon and acquire a new one, and the new one may just `pg_connect()` with a proper state.

fantix:
I've spent some time investigating the overall `backend_setting` with *static config* (=config set in env var, config files, etc.). I think we have a few undecided issues and questions:

1. Config with `backend_setting` is not 100% working in static config.

Config with `backend_setting` is only effective when the "backend setting" value is set in the backend Postgres. When it comes to static config, the current server implemented by injecting command-line arguments when starting the backend Postgres like `postgres ... -c statement_timeout=30s`. This is fine with local PG clusters, but not remote ones with `--backend-dsn` - it's simply not effective.

2. EdgeQL `CONFIGURE SYSTEM` stops working on backend config if set in static config.

Background: `postgresql.auto.conf` is the single-source-of-truth for EdgeDB dynamic system config with `backend_setting` (if the PG cluster has filesystem access). It's set by SQL `ALTER SYSTEM SET ` followed by `pg_reload_conf()`, and read from `pg_file_settings`. At the same time, we keep a copy of its value in the `system override` store (usually the "metadata" JSON on `__edgedbsys__` database description).

Because Postgres command-line arguments have higher precedence than "configuration file", the value set by the `ALTER SYSTEM` SQL was never effective, even though it displays to be up to date. For example:

```
$ env GEL_SERVER_CONFIG_cfg::query_execution_timeout='1s'  edb server &
$ edb cli
> select cfg::Config.query_execution_timeout;
{<duration>'0:00:01'}
> select sys::_sleep(1.5);
gel error: QueryTimeoutError: canceling statement due to statement timeout
```

Now increase the timeout. It looks like a success but it's actually not:

```
> configure system set query_execution_timeout := <duration>'2s';
OK: CONFIGURE INSTANCE
> select cfg::Config.query_execution_timeout;
{<duration>'0:00:02'}
> select sys::_sleep(1.5);
gel error: QueryTimeoutError: canceling statement due to statement timeout
```

For the mismatching display issue, it can be fixed by correcting the config source order like below, but I'm not sure if we want to do it:

```diff
diff --git a/edb/pgsql/metaschema.py b/edb/pgsql/metaschema.py
index 035b13e71..a2748a2ac 100644
--- a/edb/pgsql/metaschema.py
+++ b/edb/pgsql/metaschema.py
@@ -3119,10 +3119,10 @@ class SysConfigSourceType(dbops.Enum):
                 'environment variable',
                 'configuration file',
                 'command line',
+                'system override',
                 'postgres command line',
                 'postgres global',
                 'postgres client',
-                'system override',
                 'database',
                 'postgres override',
                 'postgres interactive',
```

For issue 1 and 2, another direction is to stop using command-line arguments as the implementation of static config. As non of the lower-level PG config sources are available to us, we'll have to share "postgres configuration file" with dynamic system config. This means, we'll need to:

* Sync `postgresql.auto.conf` with static config + dynamic config from `__edgedbsys__` metadata, because static config may be different from it was in the last run.
* EdgeQL `CONFIGURE SYSTEM RESET` may issue `ALTER SYSTEM SET` instead of just `RESET`, because logically dynamic config is a different layer on top of static config, and users can only reset dynamic config.
* On EdgeDB config file reload, we'll also need a sync just like at server start.

3. Do we actually want backend settings being configurable through static config?

One backend Postgres may have more than one EdgeDB server, each of them may have different static config. This was fine with the current server, when backend config are never effective if `--backend-dsn` is used. But if we fix that issue, the tenants will race to set their version of the static config to the same backend Postgres. This may lead to some confusion for multi-tenant-on-single-PG users especially after TOML config file and reloading support is added.

This may be potentially relevant to EdgeDB server command-line option `--disable-dynamic-system-config`, which prevents `CONFIGURE SYSTEM` commands on configs with `cfg::system = true`. However, this option does not stop dynamic user changes to 100% configs with `cfg::backend_setting`, because the 2 types of configs are not 100% overlapping. So this option is not an answer to question 3.

My answer would be "no", because we have to be able to statically configure those system configs which also have `cfg::backend_setting`, or there'll be no way to set them at all.

4. How about static config precedence over dynamic config?

This is to say, if one config is set in either environment variable or TOML config file, you can no longer run `CONFIGURE SYSTEM` on it. Only the server admin could modify the TOML file and reload the server to make hot changes. This almost deprecates `--disable-dynamic-system-config` and feels cleaner to implement.

An alternative design is just like `postgresql.auto.conf`, when `CONFIGURE SYSTEM` would actually store the values in a disk file. My gut feeling tells me this was something in early EdgeDB and long abandoned.

msullivan:
I think we can probably safely disallow static config of configs with `backend_setting`. 

What is the current situation for how static config interacts with `CONFIGURE INSTANCE` for *non* `backend_setting` settings? Does it work right? (It's not the end of the world if it doesn't, really)

fantix:
> What is the current situation for how static config interacts with CONFIGURE INSTANCE for non backend_setting settings? Does it work right? 

Yeah, those work just fine.

edb/pgsql/metaschema.py
```diff
@@ -3580,6 +3581,23 @@ class SysConfigFullFunction(trampoline.VersionedFunction):
...
+                    WHEN s.type = 'E' THEN 'environment variable'
+                    ELSE 'configuration file'  -- 'F', matches WHERE below
+                END) AS source,
+                config_spec.backend_setting IS NOT NULL AS is_backend
+            FROM
+                _edgecon_state s
+                INNER JOIN config_spec ON (config_spec.name = s.name)
+            WHERE
+                s.type = 'A' OR s.type = 'E' OR s.type = 'F'
+        ),
```
msullivan, line 3599: Bump CATALOG_VERSION

edb/pgsql/metaschema.py
```diff
@@ -3117,6 +3117,7 @@ def __init__(self) -> None:
                 'postgres environment variable',
                 'postgres configuration file',
                 'environment variable',
+                'configuration file',
```
msullivan, line 3120: Would it be basically workable to just pretend it is an `environment variable`?

Patching this type to add the new field is kind of hairy, since you can't reorder enum values in postgres.

edb/server/tenant.py
```diff
@@ -243,21 +254,35 @@ def set_reloadable_files(
...
             compiler = self._server.get_compiler_pool()
         result = compiler.compile_structured_config(
-            {"cfg::Config": {"email_providers": value}}, source="magic",
-            allow_nested=True,
+            {"cfg::Config": {"email_providers": value}},
+            "magic",  # source
+            True,  # allow_nested
         )
+        if asyncio.iscoroutine(result):
+            result = await result
```
msullivan, line 285: This smells; can we at least document it?

edb/server/tenant.py
```diff
@@ -1563,6 +1640,92 @@ def set_readiness_state(self, state: srvargs.ReadinessState, reason: str):
...
+
+        # Update sys pgcon and reload system config
+        async with self.use_sys_pgcon() as syscon:
+            if syscon.last_init_con_data is not self._init_con_data:
+                await syscon.sql_execute(
+                    pgcon.RESET_STATIC_CFG_SCRIPT + (self._init_con_sql or b'')
+                )
+                syscon.last_init_con_data = self._init_con_data
+            sys_config = await self._load_sys_config(syscon=syscon)
+            # GOTCHA: don't notify the change of sysconfig because it's local
```
msullivan, line 1726: Hm?

msullivan (approved):

msullivan:
Great work!

edb/server/tenant.py
fantix: Fixed to leverage typing

edb/server/tenant.py
fantix: Ah, don't notify other tenants on the same backend; comment updated.

edb/common/asyncutil.py
```diff
@@ -140,3 +151,186 @@ async def debounce(
...
+            d[self._name] = rv
+            return rv
+
+        # getattr on an object with __slots__
+        else:
+            if rv := getattr(instance, self._name, None):
+                return rv
+            rv = ExclusiveTask(self._method.__get__(instance, owner))
+            setattr(instance, self._name, rv)
+            return rv
```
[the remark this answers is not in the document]
fantix: In case of classes without `__slots__`, the `self._name` is usually just the same; using getattr/setattr will cause a infinite loop back to this `__get__()`

edb/pgsql/metaschema.py
fantix: I see, let me try!

msullivan: No, because now we have in-place upgrades, unfortunately, so we would want to just always do it this way, unfortunately

fantix: Roger; I'll test and commit this diff.

PR #8122, opened 2024-12-13
Fix issue in SQL-over-native where normalization is required

msullivan:
We index into extra_counts unconditionally, which is wrong.
This will fix #8104.

I'm splitting it out because if we merge #8104, we will want to revert
it eventually, but we want this fix to survive.


elprans (approved):

elprans (commented):
Technically the cardinality of `extra_counts` should be equal to the number of statements in input.  The fix is not incorrect per se, but I wonder if something else is amiss somewhere.

msullivan:
It is always empty for `Source`; in edgeql as well.

PR #8132, opened 2024-12-17
Fix RECURSIVE over SQL adapter

aljazerzen:


msullivan (approved):

msullivan:
How does using it multiple times work?

aljazerzen:
In SQL, recursive keyword does not apply to a specific CTE, but it applies to the whole query. So any/all CTEs can be recursive when the keyword is present.

Location of the flag in our AST is the questionable part here. We should probably has a separate ast node `With { ctes: List[CommonTableExpression], recursive: bool }`.

aljazerzen:
(... but my code does not take that into account properly)

msullivan:
Remember to label for backport. too

PR #8136, opened 2024-12-17
Update .gitignore to include edb/cli/gel

msullivan:
Follow-up for #8130


fantix (approved):

msullivan:
Is there a single thing I'm going to do today that somebody else won't have already done without me noticing?

aljazerzen:
I've sneaked it into one of my commits - sorry.

PR #8140, opened 2024-12-17
Fix psql "\d table" by making casting to regclass more resilient

### LINKED ISSUE #8138: "forbidden function" error when running `\d TableName` through SQL adapter
EdgeDB Version: `6.0-beta.1+966017c`

repro steps:

(these are identical to #8137 except using `\d TableName` instead of `\d`)

- create a 6.0b1 instance in the cloud (haven't tested to see if this repros outside the cloud, but don't see any reason it shouldn't)

- connect using the EdgeDB CLI and create a very simple schema (in my case, just `create type Ballast { create required property blob -> str; };`)

- set a password for the SQL connection (in the cloud UI, this is "how to connect?" -> SQL client -> change password)

- connect using `psql` and run `\d "Ballast"`

results:

```
> time psql -d "postgresql://admin:hunter2@free-60b1-us-west-2--zackelan.c-36.i.aws-test.edgedb.cloud:5656/main?sslmode=require" -c '\d "Ballast"'
ERROR:  forbidden function
LINE 8: WHERE p.puballtables AND pg_catalog.pg_relation_is_publishab...
                                 ^

________________________________________________________
Executed in   44.95 secs      fish           external
   usr time    6.85 millis    0.00 micros    6.85 millis
   sys time    6.77 millis  912.00 micros    5.86 millis
```

expected results:

it shouldn't fail, and ideally shouldn't take 45 seconds.

msullivan:
Apparently '12345'::regclass needs to be supported as interpreting the
number.

Fixes #8138.

### DIFF
--- a/edb/pgsql/resolver/static.py
+++ b/edb/pgsql/resolver/static.py
@@ -448,14 +448,17 @@ def cast_to_regclass(param: pgast.BaseExpr, ctx: Context) -> pgast.BaseExpr:
     """
 
     expr = eval(param, ctx=ctx)
+    res: pgast.BaseExpr
     if isinstance(expr, pgast.NullConstant):
-        return pgast.NullConstant()
-    if isinstance(expr, pgast.StringConstant):
-        return to_regclass(expr.val, ctx=ctx)
+        res = pgast.NullConstant()
+    elif isinstance(expr, pgast.StringConstant) and expr.val.isnumeric():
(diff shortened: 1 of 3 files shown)


aljazerzen:
> Apparently '12345'::regclass needs to be supported as interpreting the
number.

SQL is not a sane language.

aljazerzen (approved):

PR #8145, opened 2024-12-19
Change some pg_catalog views to be MATERIALIZED

### LINKED ISSUE #8137: slow performance when running psql `\d` through SQL adapter
EdgeDB Version: `6.0-beta.1+966017c`

repro steps:

- create a 6.0b1 instance in the cloud (haven't tested to see if this repros outside the cloud, but don't see any reason it shouldn't)

- connect using the EdgeDB CLI and create a very simple schema (in my case, just `create type Ballast { create required property blob -> str; };`)

- set a password for the SQL connection (in the cloud UI, this is "how to connect?" -> SQL client -> change password)

- connect using `psql` and run `\d`

results:

it consistently takes ~17 seconds to run `\d`:

```
> time psql -d "postgresql://admin:hunter2@free-60b1-us-west-2--zackelan.c-36.i.aws-test.edgedb.cloud:5656/main?sslmode=require" -c '\d'
          List of relations
 Schema |  Name   | Type  |  Owner
--------+---------+-------+----------
 public | Ballast | table | postgres
(1 row)


________________________________________________________
Executed in   17.64 secs      fish           external
   usr time   88.93 millis   81.39 millis    7.54 millis
   sys time   29.80 millis   25.49 millis    4.32 millis

> time psql -d "postgresql://admin:hunter2@free-60b1-us-west-2--zackelan.c-36.i.aws-test.edgedb.cloud:5656/main?sslmode=require" -c '\d'
          List of relations
 Schema |  Name   | Type  |  Owner
--------+---------+-------+----------
 public | Ballast | table | postgres
(1 row)


________________________________________________________
Executed in   18.00 secs      fish           external
   usr time    9.29 millis  735.00 micros    8.56 millis
   sys time    5.27 millis  143.00 micros    5.12 millis

> time psql -d "postgresql://admin:hunter2@free-60b1-us-west-2--zackelan.c-36.i.aws-test.edgedb.cloud:5656/main?sslmode=require" -c '\d'
          List of relations
 Schema |  Name   | Type  |  Owner
--------+---------+-------+----------
 public | Ballast | table | postgres
(1 row)


________________________________________________________
Executed in   17.60 secs      fish           external
   usr time    9.45 millis   84.00 micros    9.37 millis
   sys time    4.96 millis  799.00 micros    4.16 millis
```

expected results:

it should not take 17 seconds to list a single table

msullivan:
This *hugely* speeds up most pg_catalog SQL introspection operations, like
`\d` and pg_dump.

Unfortunately it slows down DDL operations, because we need to refresh the
views. On my machine, it looks like it typically adds about 150ms to
applying a migration.

This is probably fine for our users, but is pretty bad for us because our
test suite is pretty heavy on DDL; this slows it down on my machine from
15.5m to 19.5m, which might be too much to merge.
Thoughts?

Fixes #8137.

### DIFF
--- a/edb/buildmeta.py
+++ b/edb/buildmeta.py
@@ -60,7 +60,7 @@ from edb.common import verutils
 # The merge conflict there is a nice reminder that you probably need
 # to write a patch in edb/pgsql/patches.py, and then you should preserve
 # the old value.
-EDGEDB_CATALOG_VERSION = 2024_12_13_99_00
+EDGEDB_CATALOG_VERSION = 2024_12_17_00_00
 EDGEDB_MAJOR_VERSION = 7
 
 
--- a/edb/pgsql/dbops/views.py
+++ b/edb/pgsql/dbops/views.py
@@ -29,13 +29,14 @@ from . import ddl
 
 
 class View(base.DBObject):
-    def __init__(self, name, query):
+    def __init__(self, name, query, materialized=False):
         super().__init__()
         self.name = name
         self.query = query
+        self.materialized = materialized
 
     def get_type(self) -> str:
-        return "VIEW"
+        return "VIEW" if not self.materialized else "MATERIALIZED VIEW"
 
     def get_id(self):
         return qn(*self.name)
@@ -59,7 +60,7 @@ class CreateView(ddl.SchemaObjectOperation):
         query = textwrap.indent(textwrap.dedent(self.view.query), '    ')
         return (
             f'CREATE {"OR REPLACE" if self.or_replace else ""}'
-            f' VIEW {qn(*self.view.name)} AS\n{query}'
+            f' {self.view.get_type()} {qn(*self.view.name)} AS\n{query}'
         )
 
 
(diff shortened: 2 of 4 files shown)


1st1:
> This is probably fine for our users, but is pretty bad for us because our
> test suite is pretty heavy on DDL; this slows it down on my machine from
> 15.5m to 19.5m, which might be too much to merge.

I think we have to suck it up and still do it. We can potentially disable materialization when under "dev mode" (but have a separate test to ensure that it works in prod)

msullivan:
I kind of wish we had made SQL protocol support an opt-in extension like graphql and edgeql_http. That would let us make this change with no-cost for non-SQL users. We *could* still do it; breaking changes that require adjusting configuration are not too bad, as these things go (though I suspect @1st1 would hate that idea).

Another hacky option is a testmode config flag that disables refreshing the views for test suites that don't ever do SQL.

Another idea is to run the refresh asynchronously. Depending on how much we care about consistency and correctness, this can be various degrees of difficult. If "eventual consistency" is good enough, then it is pretty easy.

elprans:
I think we can merge this as-is and add optimizations later.  Without this ORM-driven introspection is unbearably slow.

aljazerzen (approved):

PR #8146, opened 2024-12-19
Actually validate reflection in the reflection validation action

msullivan:
...

It's also failing for seemingly no reason (startup timeout), so I
might debug that here too.


msullivan:
test_sql_sys_query_stats flake again

Issue #8147, opened 2024-12-19
SQL: can't set enum config with values that should be quoted


fantix:
- EdgeDB Version: 7.0-dev.9101+local
- EdgeDB CLI Version: -
- OS Version: -

Steps to Reproduce:

1.
   ```
   main=# set IntervalStyle TO 'ISO_8601';
   SET
   ```

2. ```
   main=# show IntervalStyle;
   ERROR:  invalid value for parameter "intervalstyle": ""ISO_8601""
   提示:  Available values: postgres, postgres_verbose, sql_standard, iso_8601.
   ```

This was found when I needed to

```
set "edb_stat_statements.track" TO 'All';
```

where `all` is a reserved keyword as in `needs_quoting()`.

PR #8148, opened 2024-12-19
test: try to fix QueryStats test to be more reliable

fantix:
This PR changes the stats test so that schema changes are not needed. The counter is added to escape all kinds of Postgres cache.

#8147 is needed for SQL config.


msullivan:
Label for backport?

msullivan (approved):

PR #8151, opened 2024-12-19
pg-ext: fix initial client_encoding

fantix:
PR #7747 changed SQL settings to always be a tuple, and missed this one.

The added tests don't fail without this fix, because PostgreSQL is actually happy with `'\"U\", \"T\", \"F\", \"-\", \"8\"'` as UTF-8; The assertion in `setting_to_sql()` will fail without the fix.


msullivan (approved):

msullivan:
Backport?

PR #8152, opened 2024-12-19
doc: add docs for sys::QueryStats

fantix:
This is mostly translated by the free version of GitHub Copilot from the source EdgeQL. I double-checked and added a few lines, but I haven't run a build to verify the doc syntax etc.

### DIFF
--- a/docs/stdlib/sys.rst
+++ b/docs/stdlib/sys.rst
@@ -5,7 +5,7 @@
 System
 ======
 
-:edb-alt-title: System Functions
+:edb-alt-title: System Functions and Types
 
 
 .. list-table::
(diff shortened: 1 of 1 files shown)


beerose (approved):

docs/stdlib/sys.rst
themajashurka, line 1: Do the `plan` related things refer to Postgres backend exclusively? Will QueryStats collect information from edgeDB server in the future?

fantix: > Do the `plan` related things refer to Postgres backend exclusively?

Currently yes, all mutating values in this view are collected in the backend (postgres).

> Will QueryStats collect information about the edgeDB-server perspective in the future?

Yes, I think so. The infra is ready for EdgeDB server to inject extra per-query info like EdgeQL compilation time, I/O time, etc.

themajashurka: Very nice, thanks! I look forward to that!

PR #8164, opened 2025-01-02
test: add lock around _test_sys_query_stats() to prevent interference

fantix:
Trying to fix: https://github.com/edgedb/edgedb/actions/runs/12589046828


msullivan (approved):

msullivan:
Are the locks system wide, or per database? Per-database won't help us here, right?

fantix:
Yeah, it's per-database. I think I'll just skip the reset part of the test.

msullivan:
> Yeah, it's per-database. I think I'll just skip the reset part of the test.

We could also test reset against a `start_edgedb_server` standalone db

PR #8170, opened 2025-01-03
Send `identity_id` on require_verification

scotttrinh:
During the sign up flow, we want to allow applications to still create and associate an Identity, since further flows like the verification flow might happen as a sign in, such as when the user verifies their email from a different device than they signed up with. That changes the semantics a bit from "we hide the identity from the application until the email is verified" to "we do not give the application an auth token until the email is verified" which seems more honest.

Backport from #8129

### DIFF
--- a/edb/server/protocol/auth_ext/http.py
+++ b/edb/server/protocol/auth_ext/http.py
@@ -446,19 +446,32 @@ class Router:
 
         email_password_client = email_password.Client(db=self.db)
         require_verification = email_password_client.config.require_verification
+        if not require_verification and maybe_challenge is None:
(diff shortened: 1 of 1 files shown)


fantix (approved):
looks like the CI failed though
 * test_http_auth_ext_local_password_register_json_02
 * I'm not sure about the test_patches failuers

scotttrinh:
Ah, thanks for the heads up, I should've run those tests locally! Looks like the patch failures are just the same test failure. Will fix.

scotttrinh:
Welp, it was a real bug! Was trying to set the `response.body` to `str` instead of `bytes`.

PR #8193, opened 2025-01-09
[DEBUG] Debug HTTP failures

msullivan:
This is an experiment that I don't expect to work, aimed at fixing
flakes like https://github.com/edgedb/edgedb/actions/runs/12662375681/job/35288956004
where something is hanging in the rust/python async HTTP library.

The (unlikely seeming to me) hypothesis here is that the pipe
write+read is not establishing a "synchronizes-with" relationship
and so is missing the message to the queue.

I don't really think this is possible, though, and even if it were I
kind of expect that it would still probably work.

### DIFF
--- a/.github/workflows.src/build.targets.yml
+++ b/.github/workflows.src/build.targets.yml
@@ -5,153 +5,64 @@ publications:
 
 targets:
     linux:
-        - name: debian-buster-x86_64
-          arch: x86_64
-          platform: debian
-          platform_version: buster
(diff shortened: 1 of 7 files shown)


msullivan:
I think I have basically convinced myself that at least in the obvious case, this ordering thing can't actually matter.

For one, it shouldn't *matter* if the channel `send` "happens-before" the `recv`; the `recv` should block then and pick it up later

PR #8201, opened 2025-01-10
Use rustls for http fetches

mmastrac:
@msullivan and I tracked down an issue where HTTP tests were causing crashes on aarch64 runners. The culprit turned out to be `openssl-probe` calling `setenv()` in a multithreaded environment.

Switch http to rustls + native certs. We still have the probe dep in Cargo.lock, but as long as we don't call the functions that set envvars, we should be fine.

### DIFF
--- a/rust/http/Cargo.toml
+++ b/rust/http/Cargo.toml
@@ -20,7 +20,9 @@ tracing.workspace = true
 scopeguard = "1"
 eventsource-stream = "0.2.3"
 
-# We want to use rustls to avoid setenv issues OpenSSL and the system certs.
+# We want to use rustls to avoid setenv issues OpenSSL and the system certs. As long
+# as we don't call `openssl_probe::*init*env*()` functions, we should be fine.
+#
 # We add these features:
 # - http2: to use HTTP/2
 # - charset: to support charset encoding
--- a/rust/http/Cargo.toml
+++ b/rust/http/Cargo.toml
@@ -17,9 +17,17 @@ tokio.workspace = true
 pyo3_util.workspace = true
 tracing.workspace = true
 
-reqwest = { version = "0.12", features = ["gzip", "deflate", "stream"] }
 scopeguard = "1"
 eventsource-stream = "0.2.3"
+
+# We want to use rustls to avoid setenv issues OpenSSL and the system certs.
+# We add these features:
+# - http2: to use HTTP/2
+# - charset: to support charset encoding
+# - gzip/deflate/brotli: to support compression
+# - stream: to support streaming responses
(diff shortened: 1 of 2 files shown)


msullivan:
Tag for backport?

rust/http/Cargo.toml
```diff
@@ -17,10 +17,20 @@ tokio.workspace = true
 pyo3_util.workspace = true
 tracing.workspace = true
 
-reqwest = { version = "0.12", features = ["gzip", "deflate", "stream"] }
 scopeguard = "1"
 eventsource-stream = "0.2.3"
 
+# We want to use rustls to avoid setenv issues OpenSSL and the system certs. As long
```
msullivan, line 23: Some sort of preposition seems needed after "setenv issues"

msullivan (approved):
What causes it to still be a dependency?

rust/http/Cargo.toml
mmastrac: Updated the docs and linked them here.

mmastrac:
> What causes it to still be a dependency?

It's also used to probe the location of the system certs on Linux which I believe is just really a fancy way to use this list of possible locations:

https://github.com/alexcrichton/openssl-probe/blob/master/src/lib.rs#L24

In theory we could (and probably should) replace this with our own cert probing functions, but I think we should consider that design a bit more carefully.

rust/http/Cargo.toml
msullivan: No, I mean it should be "setenv issues with OpenSSL" or something like that

aljazerzen (approved):

mmastrac:
> I think the unsafe env function can still be used, because rustls-native-certs [calls openssl_probe::probe()](https://github.com/rustls/rustls-native-certs/blob/main/src/unix.rs) and that [calls set_env](https://github.com/alexcrichton/openssl-probe/blob/master/src/lib.rs#L90-L103).

I think we are only calling the read function here `env::var_os(name)` which should be safe but I may have misread the code -- please correct me if I did. 

It might be better for us to use the system roots if setenv isn't being called to more easily allow end-users to update them.

rust/http/Cargo.toml
mmastrac: Yeah -- updated those docs with a preposition and added a link back to this PR :)

```
# We want to use rustls to avoid setenv issues w/ OpenSSL and the system certs. As long
# as we don't call `openssl_probe::*init*env*()` functions (functions that call setenv
# in a thread-unsafe way), we should be fine.
```

aljazerzen:
Oh yes, my mistake - we are now only invoking `env::var_os` and not `env::set_var`.

The test failure now is just ruff, nothing else.

PR #8206, opened 2025-01-13
Implict limit for SQL over native

aljazerzen:


msullivan:
Oh damn, I merged the wrong thing first

PR #8209, opened 2025-01-13
Detect crashed test runners and report and exit

msullivan:
Probably a better job could be done than this, but this should get us
a clean failure instead of a long hang.

### DIFF
--- a/edb/tools/test/mproc_fixes.py
+++ b/edb/tools/test/mproc_fixes.py
@@ -124,5 +124,3 @@ def patch_multiprocessing(debug: bool):
     _orig_pool_join_exited_workers = (
         multiprocessing.pool.Pool._join_exited_workers)
     multiprocessing.pool.Pool._join_exited_workers = join_exited_workers
-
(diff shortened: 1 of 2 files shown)


mmastrac (approved):

edb/tools/test/mproc_fixes.py
```diff
@@ -111,3 +120,9 @@ def patch_multiprocessing(debug: bool):
     # Allow workers some time to shut down gracefully.
     _orig_pool_worker_handler = multiprocessing.pool.Pool._handle_workers
     multiprocessing.pool.Pool._handle_workers = multiprocessing_worker_handler
+
+    _orig_pool_join_exited_workers = (
+        multiprocessing.pool.Pool._join_exited_workers)
+    multiprocessing.pool.Pool._join_exited_workers = join_exited_workers
+
+    multiprocessing.pool.Pool._crashed_workers = None
```
fantix, line 128: Unused?

fantix (approved):

PR #8210, opened 2025-01-13
Implement \(expr)-style string interpolation

### LINKED ISSUE #272: Consider implementing '\(expr)' string interpolation like in Swift

msullivan:
`\(` is currently rejected inside string literals, so there is no
compatability issue.

Expressions in the `\( )` blocks get automatically cast to str.

The implement is based on the lexer producing `STRINTERPSTART`,
`STRINTERPCONT`, and `STRINTERPEND` tokens that are parsed by the
parser.  The lexer needs to be a little stateful, unfortunately, to
track what kind of opening quote opened the string.

Fixes #272.

### DIFF
--- a/edb/edgeql/ast.py
+++ b/edb/edgeql/ast.py
@@ -251,7 +251,7 @@ class FunctionCall(Expr):
     window: typing.Optional[WindowSpec] = None
 
 
-class StrInterpFragment(Expr):
+class StrInterpFragment(Base):
     expr: Expr
     suffix: str
 
--- a/edb/edgeql/ast.py
+++ b/edb/edgeql/ast.py
@@ -251,6 +251,16 @@ class FunctionCall(Expr):
     window: typing.Optional[WindowSpec] = None
 
 
+class StrInterpFragment(Expr):
+    expr: Expr
+    suffix: str
+
+
+class StrInterp(Expr):
+    prefix: str
+    interpolations: list[StrInterpFragment]
+
+
 class BaseConstant(Expr):
     """Constant (a literal value)."""
     __abstract_node__ = True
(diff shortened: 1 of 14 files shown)


1st1:
The idea was to allow `\( ...)` syntax, leaving the closing paren unescaped (https://www.hackingwithswift.com/read/0/5/string-interpolation, Swift language style)

1st1 (changes requested):
This is awesome, but should be `)`, not `\)`, see my other comment.

msullivan:
Oh, lol

1st1 (approved):

msullivan:
Oh, there is a bug in the paren handling

edb/edgeql/ast.py
```diff
@@ -251,6 +251,16 @@ class FunctionCall(Expr):
     window: typing.Optional[WindowSpec] = None
 
 
+class StrInterpFragment(Expr):
```
aljazerzen, line 254:
```suggestion
class StrInterpFragment(Base):
```

aljazerzen (approved):
I'd though this would be harder to parser. But apparently not. I've tested a few deeply nested string with different quote styles and it works.

Nice.

Gobot1234:
Can this be backported for 6.0?

msullivan:
> I'd though this would be harder to parser. But apparently not. I've tested a few deeply nested string with different quote styles and it works.
> 
> Nice.

Could you add some of the things you tried as tests?

msullivan:
> Can this be backported for 6.0?

Yeah, sure. It makes using the new net request stuff easier

PR #8213, opened 2025-01-13
Implict limit for SQL over native (#8206)

msullivan:


msullivan:
Redo of https://github.com/edgedb/edgedb/pull/8206 after I screwed it up

PR #8216, opened 2025-01-14
test: bootstrap without testmode for multitenant test

fantix:
The compiler in multitenant server always use the stdlib cache pickle regardless of the actual stdlib in each tenant's backend. So we must bootstrap the backends without testmode schemas, because the stdlib cache pickle does not always have testmode schemas, e.g. in release CIs.

[Sample nightly run](https://github.com/edgedb/edgedb/actions/runs/12772132339)

### DIFF
--- a/edb/server/args.py
+++ b/edb/server/args.py
@@ -1552,6 +1552,8 @@ def parse_args(**kwargs: Any):
         if kwargs['compiler_pool_mode'] is not CompilerPoolMode.MultiTenant:
             abort("must use --compiler-pool-mode=fixed_multi_tenant "
                   "in multi-tenant mode")
+        if kwargs['testmode']:
+            abort("cannot use --testmode in multi-tenant mode")
 
     bootstrap_script_text: Optional[str]
     if kwargs['bootstrap_command_file']:
--- a/edb/testbase/server.py
+++ b/edb/testbase/server.py
@@ -2394,6 +2394,7 @@ class _EdgeDBServer:
         extra_args: Optional[List[str]] = None,
         net_worker_mode: Optional[str] = None,
         password: Optional[str] = None,
+        testmode: bool = True,
     ) -> None:
         self.bind_addrs = bind_addrs
         self.auto_shutdown_after = auto_shutdown_after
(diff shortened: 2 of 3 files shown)


msullivan:
I'm not sure this is the right approach; multitenant should work with `--testmode`, I think.

If that's substantially hard to make happen, then at least we should clearly fail with a real error message.

fantix:
Right. I'll disable `--testmode` for multitenant server in this PR to unblock the release, while supporting testmode would need to deliver with an extra testmode stdlib (patch?) pickle or something, I'll draft a new PR and start discussion there.

msullivan:
> Right. I'll disable `--testmode` for multitenant server in this PR to unblock the release, while supporting testmode would need to deliver with an extra testmode stdlib (patch?) pickle or something, I'll draft a new PR and start discussion there.

The issue is that we don't want to read from the DB to find it in multitenant mode?

edb/server/args.py
```diff
@@ -1552,6 +1552,8 @@ def parse_args(**kwargs: Any):
         if kwargs['compiler_pool_mode'] is not CompilerPoolMode.MultiTenant:
             abort("must use --compiler-pool-mode=fixed_multi_tenant "
                   "in multi-tenant mode")
+        if kwargs['testmode']:
+            abort("cannot use --testmode in multi-tenant mode")
```
msullivan, line 1557: One thing that is annoying here is that `--testmode` is only passed at bootstrap time, but multitenant won't work on any database that was created with `--testmode`.

Sigh.

msullivan (approved):
Alright, let's get the release out and then figure out if and how to support this

edb/server/args.py
fantix: Yeah, we can potentially load the stdschema from instdata for each tenant and verify its compatibility (by version?) with the stdlib loaded in multitenant server/compiler, and raise an appropriate error when they don't match. This might be useful regardless of how we deal with the testmode, though it's not a blocker as the main use case of multitenant server is the Cloud ourselves - the `--multitenant-config-file` option is still hidden from the general public.

fantix:
Some jobs in the sample nightly CI failed due to crashed workers, I suppose it's irrelevant to the testmode changes here.

msullivan:
It was an error that this was labelled as backported and not to-backport, right?

PR #8217, opened 2025-01-14
Monitor open FDs

mmastrac:
Add an FD monitor for servers. This is potentially something we can do more efficiently in Rust at a later date, but a Python implementation will allow for an easier backport to 5.6.

We use a new shortcut available in modern Linux kernels to avoid counting directory entries (os.stat's size == # of open FDs).

Threads are used to simplify the implementation.

### DIFF
--- a/edb/server/metrics.py
+++ b/edb/server/metrics.py
@@ -16,6 +16,8 @@
 # limitations under the License.
 #
 
+import os
+import sys
 
 from edb.common import prometheus as prom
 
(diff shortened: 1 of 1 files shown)


edb/server/metrics.py
```diff
@@ -262,3 +264,77 @@
...
+def monitor_open_fds_linux():
+    import time
+    while True:
+        max_open_fds.set(os.sysconf('SC_OPEN_MAX'))
+        # To get the current number of open files, stat /proc/self/fd/
+        # and get the size. If zero, count the number of entries in the
+        # directory.
+        #
+        # This is supported in modern Linux kernels.
+        # https://github.com/torvalds/linux/commit/f1f1f2569901ec5b9d425f2e91c09a0e320768f3
```
elprans, line 293: Oh nice!

PR #8218, opened 2025-01-14
Add `edb::env-switcher` and `edb::split-section` sphinx directives

jaclarke:
- `edb::env-switcher`:
This directive is used to mark a group of pages as a single 'logical' page in the docs navigation, within which the user can switch between the content of each rst page using the environment switcher ui. To use create an empty `index.rst` page with this directive at the start, and a toctree containing the nested pages:
```rst
.. edb:env-switcher::

=============
Some tutorial
=============

.. toctree::
    :maxdepth: 3
    :hidden:

    any
    typescript
    python
```
The name of each nested page must be defined in the env names mapping in the docs site repo.

- `edb::split-section`:
Any page containing at least one of these directives will be rendered with the split layout design. The last child element in the directive must be a `code-block` or `tabs` directive, which will be rendered on the right side of the page, and the rest of the content in the directive will be on the left, aligned with the code block. Any other page content not in `split-section` directives will also be rendered on the left.
- `edb::split-point`:
Overrides the default behaviour above where the last code block is rendered on the right, and instead explicitly marks the point in the `split-section` after which the content is rendered on the right side of the page. Has an optional arg which can be either `above` or `below`, to pick whether the content before or after the `split-point` is rendered on the right side of the page (on mobile the order is the same as in the rst)

### DIFF
--- a/edb/tools/docs/__init__.py
+++ b/edb/tools/docs/__init__.py
@@ -30,6 +30,7 @@ from . import eql
 from . import js
 from . import sdl
 from . import graphql
+from . import go
 from . import shared
 
 
@@ -119,6 +120,7 @@ def setup(app):
     js.setup_domain(app)
     sdl.setup_domain(app)
     graphql.setup_domain(app)
+    go.setup_domain(app)
 
     app.add_directive('versionadded', VersionAdded, True)
     app.add_directive('versionchanged', VersionChanged, True)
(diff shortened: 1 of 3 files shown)


scotttrinh (approved):
I was expecting a bit more code to review here 😅

jaclarke:
> I was expecting a bit more code to review here 😅

Yeah, most of the code is on the docs site side. 😄 I added you as reviewer just to check that how the directives work/are used makes sense. I've also flipped the `split-section` directive around, so the code block is at the end now, which I think matches the intended reading order.

PR #8220, opened 2025-01-14
Try to fix test_server_ops_multi_tenant in release builds

msullivan:
One of the tests needs to access the /server-info endpoint in order to
test magic_smtp_config. But since testmode doesn't work with
multitenant (see #8216), this does not work when being run outside
devmode.

Skip that part of the test when not in devmode, then.

I don't find this particularly satisfying.


fantix (approved):
Roger; I'll just add testmode support

PR #8223, opened 2025-01-15
Add a full test for notice handling on handshake

mmastrac:
This improves the captive postgres utility to make it more flexible and ensures that we can test a postgres server configured with `client_min_messages=debug5`.


aljazerzen (approved):
Looks good

PR #8229, opened 2025-01-16
multitenant: patch stdlib with testmode schema if requested

fantix:
- [x] patch stdlib starting multitenant server with --testmode
- [x] fix tests
- [ ] reject tenant backends with incompatible std schema
- [x] [Build CI sample run](https://github.com/edgedb/edgedb/actions/runs/12817630831)
- [x] [Standalone backend CI sample run](https://github.com/edgedb/edgedb/actions/runs/12821283901)

### DIFF
--- a/edb/server/args.py
+++ b/edb/server/args.py
@@ -1552,8 +1552,6 @@ def parse_args(**kwargs: Any):
         if kwargs['compiler_pool_mode'] is not CompilerPoolMode.MultiTenant:
             abort("must use --compiler-pool-mode=fixed_multi_tenant "
                   "in multi-tenant mode")
-        if kwargs['testmode']:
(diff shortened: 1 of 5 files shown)


msullivan (approved):
This is a big improvement.

There's still a somewhat weird mismatch with non-multitenant mode: there, `--testmode` only needs to be specified when *bootstrapping* the instance, whereas in multi-tenant it will be needed every time

It would be nice to avoid this but it is obscure and won't hurt anyone but us, testing things, but we should at least leave a comment about it

fantix:
I'd like to still check the schema compatibility when adding new tenants; though it's a bit more complex than I expected, I'll merge this PR with added comments for now, and start a new PR for catalog version / std schema check.

PR #8230, opened 2025-01-16
Bump Rust toolchain to 1.81

fantix:
Refs: https://github.com/edgedb/edgedb-cli/pull/1439/commits/70f4491397a65eae14272c612d6e92085f80e8da

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.80.1"
+channel = "1.81"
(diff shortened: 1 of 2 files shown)


mmastrac (approved):
LGTM -- maybe need a cargo fmt

fantix:
> LGTM -- maybe need a cargo fmt

That failed cargo test seems to be a real difference in Rust 1.81 where `crate` became `$crate` - your custom macro skipped whitespace, but the error message didn't so it was hard to find the real difference.

PR #8233, opened 2025-01-17
Add an `override-cors-allowed-origins` option

jaclarke:
For cloud instances we currently default to setting the `cors_allow_origins` config to `*` to let the cloud ui make queries to the instance. But the user can change that config, which breaks the UI, and also for security it's probably better to leave the default `cors_allow_origins` config empty.

Since we want to allow the cloud UI to always be able to query the instance, regardless of how the user configures `cors_allow_origins`, this server option will allow us to set the cloud ui as an always allowed cors origin. Supporting multiple origins and wildcards is to handle the gel rename (allow both our new and old cloud ui urls to work) and for staging ui builds.

### DIFF
--- a/edb/server/args.py
+++ b/edb/server/args.py
@@ -292,6 +292,8 @@ class ServerConfig(NamedTuple):
 
     admin_ui: bool
 
+    override_cors_allowed_origins: Optional[str]
+
 
 class PathPath(click.Path):
     name = 'path'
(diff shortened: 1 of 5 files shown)


scotttrinh:
I wonder if we can/should use the same mechanism we use for the magic SMTP provider?

edb/server/args.py
```diff
@@ -1081,6 +1083,16 @@ def resolve_envvar_value(self, ctx: click.Context):
         ),
         default='default',
         help='Enable admin UI.'),
+    click.option(
+        '--override-cors-allowed-origins',
+        envvar="GEL_SERVER_OVERRIDE_CORS_ALLOWED_ORIGINS",
```
elprans, line 1088:
```suggestion
        envvar="GEL_SERVER_CORS_ALWAYS_ALLOWED_ORIGINS",
```

elprans (approved):

PR #8237, opened 2025-01-21
Allow dropping isolation level to REPEATABLE READ in a READ ONLY tx

elprans:
This re-enabled REPEATABLE READ isolation in explicit transaction
blocks, but only in those that are READ ONLY.  Anomalies observed in
read statements are normal non-serializability anomalies, so with an
explicit opt-in it is a reasonable tradeoff between lessening the
transaction serialization load and consistency.

### DIFF
--- a/edb/edgeql-parser/src/keywords.rs
+++ b/edb/edgeql-parser/src/keywords.rs
@@ -79,6 +79,7 @@ pub const UNRESERVED_KEYWORDS: phf::Set<&str> = phf_set!(
     "reject",
     "release",
     "rename",
+    "repeatable",
     "required",
     "reset",
     "restrict",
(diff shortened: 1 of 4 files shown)


1st1 (approved):

msullivan (approved):
We'll need to support this in bindings too, right?

elprans:
> We'll need to support this in bindings too, right?

Yes

themajashurka:
What about implicit transactions, like standalone queries? Shall I wrap my existing queries in tx blocks to leverage a lower isolation level?

msullivan:
> What about implicit transactions, like standalone queries? Shall I wrap my existing queries in tx blocks to leverage a lower isolation level?

I think we're going to be doing something shortly to support that better

CodesInChaos:
Would it make sense to name it SNAPSHOT instead of REPEATABLE READ, since postgres actually offers snapshot semantics?

> The Repeatable Read isolation level is implemented using a technique known in academic database literature and in some other database products as Snapshot Isolation.

PR #8248, opened 2025-01-23
Delete old .s.EDGEDB.admin.XXX sockets

msullivan:
This will make sure we can create the fallback symlink
even if an old socket wasn't deleted, which should fix
edgedb/edgedb-cli#1418.

(Though maybe we want a CLI-side fix *also*, so 6.0b2 works better?)


fantix (approved):
(backport tag)

PR #8249, opened 2025-01-24
Retry test suite queries after TransactionConflictError

msullivan:
Only outside of transactions. This should reduce flakes
on global DDL conflicts, among other things.

### DIFF
--- a/edb/testbase/connection.py
+++ b/edb/testbase/connection.py
@@ -144,6 +144,11 @@ class BaseTransaction(abc.ABC):
     async def _commit(self) -> None:
         query = self._make_commit_query()
         try:
+            # Use _fetchall to ensure there is no retry performed.
+            # The protocol level apparently thinks the transaction is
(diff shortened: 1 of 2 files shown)


edb/testbase/connection.py
```diff
@@ -144,7 +144,7 @@ async def commit(self) -> None:
     async def _commit(self) -> None:
         query = self._make_commit_query()
         try:
-            await self._connection.execute(query)
```
aljazerzen, line 147: Why this change?

aljazerzen (approved):

edb/testbase/connection.py
msullivan: Because once the COMMIT fails, the protoocl driver thinks it isn't in a tx anymore and wants to retry it. This prevents that.

There might be a cleaner way but I didn't see it right away.

I'll add a comment.

PR #8251, opened 2025-01-24
[sql] Support inserting into column with _id suffix

msullivan:
Some special handling needed to not assume it is a link.


elprans (approved):

aljazerzen (approved):
Great, you've paid off my dept :D

PR #8254, opened 2025-01-24
[sql] Fix (SELECT * FROM ...) in subqueries

msullivan:
Currently it will select all columns in *all* tables in scope. We only
want to select from the things in the FROM clause.


elprans (approved):

aljazerzen (commented):
Looks good.

PR #8261, opened 2025-01-26
Avoid memory leak by moving `lru_cache` to `__init__`

### LINKED ISSUE #5377: Make sure we don't use lru_cache on methods
I think we probably do, and I think I disabled the lint for it.

It leaks memory by saving references to `self`

DavidRomanovizc:
Related to Issue #5377
Closes #5377

### What was wrong?
The codebase currently use the `lru_cache` decorator in the classes over methods. The ruff linter reported a warning, indicating potential memory leaks in class methods, because the `lru_cache` decorator retain instance references, impeding garbage collection and potentially leading to memory-related issues.

### How was it fixed?
I stopped using decorators directly over methods and instead applied the decorator functions during class instance initialization.

### DIFF
--- a/edb/common/enum.py
+++ b/edb/common/enum.py
@@ -32,7 +32,7 @@ class StrEnum(str, enum.Enum):
 @functools.total_ordering
 class OrderedEnumMixin():
 
-    def __init__(self) -> None:
+    def __init__(self, *args, **kwargs) -> None:
         self._index_of = functools.lru_cache(maxsize=None)(self.__index_of)
 
     @classmethod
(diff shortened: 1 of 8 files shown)


msullivan (approved):
Thanks! I think this should do the right thing.

msullivan (changes requested):
Hm, maybe not *quite* the right thing, since it doesn't seem to work

msullivan:
Out of curiosity, what prompted you to tackle this?

DavidRomanovizc:
> Out of curiosity, what prompted you to tackle this?

I tackled a similar issue in another repository, so I have some idea how it can be fixed

msullivan:
Oh, this has another problem, which is that I think the cache will get pickled now.

I think I am going to try another approach inspired by this.

msullivan:
I'm going to close this, because I think I'm going to solve this with https://github.com/edgedb/edgedb/pull/8267.

Thanks, though! This lead me directly to that PR.

PR #8269, opened 2025-01-29
Quickstart tutorial with Next.js

scotttrinh:


docs/intro/quickstart/index.rst
```diff
@@ -0,0 +1,795 @@
...
+            "id": "123e4567-e89b-12d3-a456-426614174004",
+            "front": "What is the highest freshwater lake on Earth?",
+            "back": "Lake Titicaca"
+          }
+        ]
+      }
+
+.. edb:split-section::
+
+  As time goes on, and our planet changes, perhaps we'll want to update one of the cards with the latest in geographical knowledge. Let's add a route to update a card by its ID.
```
scotttrinh, line 395: I think making an "add a card to the deck" route is more interesting than updating an existing card.

docs/intro/quickstart/index.rst
```diff
@@ -0,0 +1,795 @@
...
+      update Card
+      filter .id = cardId
+      set {
+        front := front,
+        back := back,
+      };
+
+.. edb:split-section::
+
+  Now that the queries are in separate files, we can generate the functions for each query. This will create a file next to the EdgeQL file with a fully type-safe function based on introspecting the query with the server.
```
scotttrinh, line 724: This sentence needs some work.

docs/intro/quickstart/index.rst
```diff
@@ -0,0 +1,795 @@
...
+              id,
+              name,
+              description,
+              cards := (select .<deck[is Card] {
+                id,
+                front,
+                back,
+              } order by .order),
+            } filter .id = deckId
+          `,
```
beerose, line 339: Don't we want to show the query builder in the quick start? It's the recommended way in TypeScript projects, right?

scotttrinh: > Don't we want to show the query builder in the quick start?

We can, and I'm open to introducing it in the same way we are introducing the `queries` generator by rewriting an EdgeQL query into the query builder, but I have a few issues with jumping into the query builder right away:

1. It's a lot of syntax for a non-trivial query like this. For example, this one would be:

```ts
e.params({ deckId: e.uuid }, (params) =>
  e.select(e.Deck, (deck) => ({
    filter: e.op(deck.id, "=", params.deckId),
    id: true,
    name: true,
    description: true,
    cards: e.select(deck["<deck[is Card]"], (card) => ({
      id: true,
      front: true,
      back: true,
      order_by: card.order,
    }),
  }))
);
```

That's not much worse than the EdgeQL, but you have to be able to follow the EdgeQL concepts encoded in the query builder, and also intuit what all of these bits of query builder syntax are at the same time. Compared to EdgeQL + `client.query*` it just seems a lot less heavy to me for an introduction. Especially since we're not really explaining much here, just showing it to you.

> It's the recommended way in TypeScript projects, right?

2. For static queries, I personally prefer the `queries` generator. It's less typechecking overhead and I think EdgeQL is a really great query language so I like leaning into it as much as possible.

---

Having said that, maybe we can show off a dynamic query which is where the query builder really shines. And after you've gotten accustom to the EdgeQL you've read so far, maybe it'll be an easier leap to see the query builder since it tries to resemble EdgeQL?

docs/intro/quickstart/setup.rst
```diff
@@ -0,0 +1,69 @@
...
+      $ npm run test
+
+
+.. edb:split-section::
+
+  Let's quickly take a poke around the empty database with our CLI REPL.
+
+  .. code-block:: sh
+
+      $ npx gel
```
1st1, line 27: Wait, we haven't run `npx gel project init`, so `npx gel` will just fail as it won't be able to connect to anything

docs/intro/quickstart/setup.rst
```diff
@@ -0,0 +1,69 @@
...
+  Fun! We'll create a proper data model for our application in the next step, but for now, let's take a look around the project we've just created. Most of the project files will be familiar to you if you've worked with Next.js before. So let's focus on the new files that integrate Gel.
+
+  - ``gel.toml``: This is the configuration file for the Gel project instance.
+  - ``dbschema/``: This directory contains the schema for the database, and later supporting files like migrations, and generated code.
+  - ``dbschema/default.gel``: This is the default schema file that we'll use to define our data model. It is empty for now, but we'll add our data model to this file in the next step.
+  - ``lib/gel.ts``: This file contains the Gel client, which we'll use to interact with the database.
+
+  .. code-block:: sh
+
+    $ tree
```
1st1, line 68: No output here so it looks confusing

CleanShot 2025-02-07 at 15 49 07@2x

1st1: Also I don't have the `tree` command installed, so I wouldn't assume everybody has it.

I'd just copy/paste the output of tree here, not asking the user to run anything.

CleanShot 2025-02-07 at 15 51 41@2x

docs/intro/quickstart/modeling.rst
```diff
@@ -0,0 +1,51 @@
...
+    +     required front: str;
+    +     required back: str;
+
+    +     required deck: Deck;
+    +   }
+      };
+
+.. edb:split-section::
+
+  Now that we've written our first version of our data model's schema, we will create a migration to apply this schema to the database. When making changes to our schema, the CLI migration tool will ask some questions to ensure that the changes we are making are what we expect. Since we had an empty schema before, the CLI will skip asking any questions and simply create the migration file.
```
1st1, line 33: > Now that we've written our first version of our data model's schema

This is a perfect opportunity to explain what a migration even means.

Basically:

```suggestion

  Congratulations! We have the first version of our data model's schema *stored in a file on disk*. Now we need to signal the database to actually create types for `Deck` and `Card` so you can store data and query data from them.
  
  To make Gel do that, we need to do two quick steps:
  
  * Create a migration: a file with a list of low-level instructions. 
  
    Note: when you are changing existing schema, the CLI migration tool might ask questions to ensure that it understands your changes exactly. 
  
  * Apply the migration: basically, tell Gel "I want you to use these instructions and get my types ready for action.
  
   Since we had an empty schema before, the CLI will skip asking any questions and simply create the migration file.
```

docs/intro/quickstart/modeling.rst
```diff
@@ -0,0 +1,51 @@
...
+      Created ./dbschema/migrations/00001-m125ajr.edgeql, id: m125ajrbqp7ov36s7aniefxc376ofxdlketzspy4yddd3hrh4lxmla
+      $ npx gel migrate
+      Applying m125ajrbqp7ov36s7aniefxc376ofxdlketzspy4yddd3hrh4lxmla (00001-m125ajr.edgeql)
+      ... parsed
+      ... applied
+
+
+.. edb:split-section::
+
+  Let's take a look at the schema we've generated in our built-in database UI. We can use this tool to visualize our data model and see the object types and links we've defined.
```
1st1, line 47: We need like a gallery widget here. First screenshot will be dashboard with circled button for "Schema Viewer", second will be the viewer

```suggestion
  Let's take a look at the schema we've generated in our built-in database UI. We can use it to visualize our data model and see the object types and links we've defined.
```

docs/intro/quickstart/working.rst
```diff
@@ -0,0 +1,412 @@
+.. _ref_quickstart_working:
+
+=====================
+Working with our data
+=====================
+
+.. edb:split-section::
+
+  With TypeScript, there are three ways to run a query: use a string EdgeQL query, use our ``queries`` generator to turn a string of EdgeQL into a TypeScript function, or use our query builder API to build queries dynamically in a type-safe manner. In the next example, we'll show you each of these methods, but for the rest of the tutorial we'll use the query builder API.
```
1st1, line 9:
```suggestion
  With TypeScript, there are three ways to run a query: use a string EdgeQL query, use our ``queries`` generator to turn a string of EdgeQL into a TypeScript function, or use our query builder API to build queries dynamically in a type-safe manner. In the next example, we'll show you each of these methods, but for the rest of this section of the tutorial we'll use the query builder API.
```

?

docs/intro/quickstart/working.rst
```diff
@@ -0,0 +1,412 @@
...
+      :caption: Queries generator
+
+      $ npx @gel/generate queries
+
+.. edb:split-section::
+
+  Now that we have a schema defined, let's create a simple page with a button that allows users to import a deck of cards from a JSON file. We'll use Next.js server actions to handle the file upload and insert the data into our database. The JSON file will contain the deck name, optional description, and an array of cards with front and back text.
+
+  .. note::
+      If you are seeing TypeScript or ESLint errors, you may need to restart the TypeScript language server, or the ESLint server. Sometimes when adding new files, the language server or ESLint will not pick up the new files until you restart the server. This will be true for the rest of the tutorial, but the majority of development is not creating new files, so after this initial onboarding pain, you'll find that editor tooling works well. This is not a Gel-specific issue, but rather a general issue with starting a new project.
```
1st1, line 28: > restart the TypeScript language server, or the ESLint server.

(for VSCode do Shift+Cmd/Ctrl+P "reload....")

docs/intro/quickstart/working.rst
```diff
@@ -0,0 +1,412 @@
...
+      $ npx @gel/generate edgeql-js
+
+    .. code-tab:: sh
+      :caption: Queries generator
+
+      $ npx @gel/generate queries
+
+.. edb:split-section::
+
+  Now that we have a schema defined, let's create a simple page with a button that allows users to import a deck of cards from a JSON file. We'll use Next.js server actions to handle the file upload and insert the data into our database. The JSON file will contain the deck name, optional description, and an array of cards with front and back text.
```
1st1, line 25: image

We have to explain the actual querying. Or what's the point of this getting started tutorial? Next section is access control, but at this point the use has no clue about the querying unless they click through the files and just stare at the code.

We cannot rely on this being "self-explanatory".

1st1: I propose a different flow. The problem here is that we can't seem to find a way to escape NextJS complexity of several files, UI and its styling, etc. But we have to or we're not educating people on **how** to actually use Gel.

I think that the next section after "Modeling our data" should be:

* "Connecting to the database", followed by 
* "Your first queries", followed by 
* "Using Gel from NextJS" (the current "Working with our data").

👉 The "Connecting to the database" section should tell the reader to create an empty file, import gel, create a client, and run a simple query:

```
import createClient from 'gel';

client = createClient()

console.log(await client.query(
  'select 'hello from Gel'
))
```

```
$ node test.ts
hello from Gel
```

👉 The "Your first queries" section should keep editing in that file, and introduce a very simple select query via the QB API, then do the JSON insert stuff. Slowly introducing the user to our QB.

👉 Then "Using Gel from NextJS" section can actually be more brief and focus on explaining **where to plug** `createClient` and **where to use it from**, followed by a few snippets (pretty much what you have in the current "Working with our data​" but slightly adjusted.)

docs/intro/quickstart/workflow.rst
```diff
@@ -0,0 +1,111 @@
+.. _ref_quickstart_workflow:
+
+===============================
+A Smoother Development Workflow
```
1st1, line 4: I think titles should be more descriptive that will result in better "searchability" (and SEO, ultimately), e.g. in this case: "Improving workflow with CLI hooks"

docs/intro/quickstart/workflow.rst
```diff
@@ -0,0 +1,111 @@
...
+
+  .. code-block:: sdl-diff
+    :caption: dbschema/default.gel
+
+      module default {
+        type Deck {
+          required name: str;
+          description: str;
+    +
+    +     cards := (select .<deck[is Card] order by .order);
```
1st1, line 37: I think we should set max line length to 50 or less. Or just prefer to spend more space vertically than horizontally. People on mobile will say thank you. But even on desktop:


CleanShot 2025-02-07 at 17 15 16@2x

1st1: @jaclarke we need to make horizontal scrollbars always visible on macos for code snippets

docs/intro/quickstart/access.rst
```diff
@@ -0,0 +1,423 @@
...
+    +
+    +   type User {
+    +     required name: str;
+    +   }
+    +
+    +   type AccessToken {
+    +     required user: User;
+    +   }
+    +
+        type Deck {
```
1st1, line 34: We have to split this long example into smaller blocks -- one talking about adding just `single optional global access_token` line, adding a sentence on what it does, etc.

1st1:
@scotttrinh I've made a pass through 60% of the tutorial and I have a feeling that the rest isn't ready for a review. Let's touch base on Monday. Good start though, I love the new format.

docs/intro/quickstart/setup.rst
scotttrinh: Oh right, this was originally starting with the `npm create @gel` flow which _does_ the `project init` flow for you. I'll add it and add some explanation.

docs/intro/quickstart/setup.rst
scotttrinh: Sorry, some of this content is "TODO", this is a reminder for me to add the output of such a command.

docs/intro/quickstart/working.rst
scotttrinh: I use the query builder for the entire tutorial, there are no examples of using the other two methods. Should there be?

docs/intro/quickstart/working.rst
scotttrinh: > But we have to or we're not educating people on how to actually use Gel.

Yeah, I am taking a pretty opinionated stance on what "quick start" means, and coming down on the side of "show, don't tell", and hope that leaving the user with a working application written in the typical style will result in a good jumping off point of editing existing, well-written code, rather than trying to learn the language over the course of the tutorial. I'm not convinced this is correct form, but it allows us to cover a lot of ground without asking that much from the reader. Give them a sense of what it looks and feels like to use Gel, instead of it being more "lesson" like.

---

> I propose a different flow. The problem here is that we can't seem to find a way to escape NextJS complexity of several files, UI and its styling, etc.

I like your suggestion here, and I think it's probably what a traditional "getting started" guide would do. Here's another approach that I was just thinking of this morning that also addresses the "there's too much Next.js in this tutorial" feeling (cc @beerose ):

Have _all_ of the Next.js stuff already written and using mock data, so we don't focus much on the application logic and it can look very nice. And most of the tutorial you're updating `actions.ts` and `queries.ts` to show converting the static data into queries, and wiring up the mutations. That focuses the tutorial on "using Gel at the data layer of your Next.js application" rather than it being a "make an app with Next.js and Gel". That would give us more time and space for going a little deeper into the queries and mutations itself.

So what that might look like is:
-  setup: a quick tour of the application itself, then a tour of the Gel-specific files like we have currently. We can look at the `@/lib/gel.ts` file and see that the `createClient()` call will connect to the database. We can even demo your script, but I know there were strong feelings about _not_ starting out with strings of EdgeQL and going right to the query builder, so I'm not sure if that approach fits with the "baby's first trivial query in a script". I'm fine with it though, and personally was proposing that we start with strings -> queries code gen -> query builder. 😅 
- modeling: look at the mock data, and write the schema. We can talk about the nested nature of the data with decks containing cards. End this section with generating the query builder, and maybe replacing our `select "Hello from Gel"` query with a query builder version?
- working: Replace the mock data with queries and mutations. No Next.js component work, just updating the `actions.ts` and `queries.ts` and looking at the prebuilt UI.
- the rest: keep it pretty much the same, but no UI/Next.js specific stuff except for adding the access control, etc.

wdyt?

docs/intro/quickstart/workflow.rst
scotttrinh: Related: I also think we should use more of the horizontal space of the page for the code snippets. Screenshots and code feel cramped due to the space we leave for the in-page TOC.

docs/intro/quickstart/setup.rst
1st1: let's try `ls -R` which should be OK for a small dir tree.

1st1: and everybody has it

docs/intro/quickstart/index.rst
```diff
@@ -0,0 +1,59 @@
...
+
+Don't worry if you're new to Gel - we'll have you up and running with a Next.js starter project and a local Gel database in just about 5 minutes. From there, we'll guide you through building the complete application in roughly 60-90 minutes.
+
+Our Flashcards app will be a modern web application with the following features:
+
+* Create, edit and delete flashcard decks
+* Add and remove cards from decks
+* Display cards with front/back text content
+* Simple UI with Next.js and Tailwind CSS
+* Clean, type-safe data modeling using Gel's schema system
```
beerose, line 30: How about changing this to "By the end of this tutorial, you will:
- Learn how to create, update and delete Flashcards with Gel 
- Learn how to select data from Gel and display in the UI
- Create a simple UI with Next.js and Tailwind CSS
- Explore type-safe data modeling using Gel's schema system"
? So that it focuses more on learning Gel while doing something else rather than on the app.

docs/intro/quickstart/dynamic.rst
```diff
@@ -0,0 +1,136 @@
...
+===============
+Dynamic Queries
+===============
+
+When updating data, we often want to modify only specific fields while leaving others unchanged. For example, we might want to update just the front text of a flashcard or only the description of a deck. There are two main approaches to handle these partial updates:
+
+1. Write a single complex query that conditionally handles optional parameters
+2. Build the query dynamically in the application code based on which fields need updating
+
+The second approach using dynamic queries tends to be more performant and maintainable. EdgeDB's TypeScript query builder excels at this use case. It allows you to construct queries dynamically while maintaining full type safety. Let's see how to implement this pattern.
```
beerose, line 12:
```suggestion
The second approach using dynamic queries tends to be more performant and maintainable. Gel's TypeScript query builder excels at this use case. It allows you to construct queries dynamically while maintaining full type safety. Let's see how to implement this pattern.
```

beerose (commented):
A note on "we" vs "you": I noticed the quickstart switches between the two, which is generally not a _wrong_ practice, but I was thinking we should stick to one. 

"We" is great for creating personal connections between the user and the company, but it's sometimes unclear what "we" refers to - is it user+author, the company, or the tool? I saw another argument somewhere against "we" that "you're not sitting with the author". 

For example, if there's "now, we'll create a function", it can be unclear whether "we will" means _there's a Gel tool that does this_ or it's metaphorically "we". In most cases "we" don't do these things - user "you" does. Or "we will create a migration to apply this schema" can make total sense in the context but if someones loses the context, they can think "gel (we) will create a migration to apply this schema".

I would use "we" only for introductions or summaries - "we'll walk you through...", "we covered", "we have a template app for you to clone" - i.e. in places where it make sense to refer to us (authors/company). And for all instructions we can go with "you" - "make sure you have ...",  "(you) clone our repo", for example:

`We'll create a proper data model for our application in the next step` 
can be 
`We will show you how to create a proper data model for your application in the next step`
or  
`In the next step, you will create a proper data model for your application`.

Switching to "you" for instructions takes away some of this connection vibe, but can make it more _straight to the point_. FWIW, "let's make ..." is fine as is. 

Also, for titles, we can skip it altogether, and say "Modeling **the** data", "Setting up **the** environment", etc.

docs/intro/quickstart/index.rst
scotttrinh: Yeah, I like that! It's a strict improvement no matter what we decide about the rest of the content. FWIW, the third bullet ("Create a simpleUI with Next.js and Tailwind CSS") might be something we decide to drop by making the starting point a fully functional app with static, mock data, and we spend the whole tutorial working through replacing this static data layer with a Gel-powered one. I'd love to here your thoughts on that over on [the main thread](https://github.com/edgedb/edgedb/pull/8269#discussion_r1947387337) ☝️

docs/intro/quickstart/setup.rst
beerose: I think showing the file structured (with nesting level being 2?) would be good. People will have an idea what to expect, but also this would fill some space (there's much more text on the left).

docs/intro/quickstart/setup.rst
```diff
@@ -0,0 +1,69 @@
+.. _ref_quickstart_setup:
+
+===========================
+Setting up your environment
```
beerose, line 4: We should also include `createClient` here as we're already jumping to using Gel in Next.js and this file seems to be a good opportunity to explain the Gel-related files + how to get a client.

(I think it was already mentioned somewhere, but I couldn't find it, sorry for repeating)

docs/intro/quickstart/modeling.rst
```diff
@@ -0,0 +1,51 @@
+.. _ref_quickstart_modeling:
```
beerose, line 1: What if we add a bit more here? Since it's relatively short, maybe we can also explain links here?

docs/intro/quickstart/modeling.rst
```diff
@@ -0,0 +1,51 @@
...
+    :caption: dbschema/default.gel
+
+      module default {
+    +   type Deck {
+    +     required name: str;
+    +     description: str;
+    +   };
+
+    +   type Card {
+    +     required order: int64;
```
beerose, line 23: I wonder if we should explain `order` a bit more - can be me, but I only got it after reading the next chapters 😅

scotttrinh: Yeah, good call, we can say that link sets are unordered, and since we want the cards to have a fixed order, we add this property so we can order them when we query for the deck and its cards.

docs/intro/quickstart/access.rst
```diff
@@ -0,0 +1,423 @@
+.. _ref_quickstart_access:
```
beerose, line 1: How about making this the last section? I'm thinking so as this one gets a bit more difficult and there's a lot going one. Maybe we can cover everything else before access policies, so that we don't scare people away

docs/intro/quickstart/modeling.rst
scotttrinh: Yeah, I think in the mock-data -> Gel data version of this modeling section, we can look at the structure of the mock data and talk more about how to map this structure onto the Gel schema language's data modeling concepts, like links, cardinality, sets vs arrays, etc.

docs/intro/quickstart/access.rst
scotttrinh: Yeah, I'm fine with that!

docs/intro/quickstart/access.rst
```diff
@@ -0,0 +1,423 @@
...
+    Generation complete! 🤘
+
+.. edb:split-section::
+
+  Let's create a page for creating a new user and getting an access token. Let's start by creating the query to create a new user which will return the ``AccessToken.id`` which we will use as the access token itself. We will save this access token in a cookie so that we can authenticate requests in other server actions and route handlers.
+
+  .. tabs::
+
+    .. code-tab:: typescript
+      :caption: app/signup/actions.ts
```
beerose, line 113: It looks like the indentation here is too big (comparing to other code blocks):

CleanShot 2025-02-11 at 15 51 29@2x

docs/intro/quickstart/access.rst
beerose: Another thought - maybe this can be a part of another auth-focused guide?

docs/intro/quickstart/access.rst
scotttrinh: Yeah, I think _all_ of my non-diff code blocks are wrong. I wrongly assumed I could make the diff and non-diff code blocks have the same indentation and that the directive would normalize them, but it looks like I cannot. That makes it _very_ annoying to start with a code block and then switch to the same code in a diff 😅 . Will fix once it's all finished.

docs/intro/quickstart/index.rst
```diff
@@ -0,0 +1,59 @@
...
+  :maxdepth: 1
+  :hidden:
+
+  setup
+  modeling
+  working
+  workflow
+  access
+  inheritance
+  dynamic
```
beerose, line 17: We could also have "Summary" section with stuff like: 
- summary of what we build
- "what's next" with links to deployment, auth, ai, etc.
- if we're deploying the quickstart app, we can paste a demo link there
- links to gel examples and templates

docs/intro/quickstart/access.rst
scotttrinh: As discussed, we're going to remove the access control/policies stuff from the initial quickstart and just save most of that for the auth extension tutorial that will build on this.

docs/intro/quickstart/dynamic.rst
scotttrinh: Removed the separate "dynamic queries" section. This content mostly moved into the "Working with the data" section.

docs/intro/quickstart/setup.rst
```diff
@@ -0,0 +1,84 @@
...
+  - ``dbschema/``: This directory contains the schema for the database, and later supporting files like migrations, and generated code.
+  - ``dbschema/default.gel``: The default schema file that you'll use to define your data model. It is empty for now, but you'll add your data model to this file in the next step.
+  - ``lib/gel.ts``: A utility module that exports the Gel client, which you'll use to interact with the database.
+
+  .. code-block:: sh
+
+    $ tree
+
+
+.. edb:split-section::
```
beerose, line 71: Thinking if this is the best place - on one hand yes, because it's part of the setup, and on the other user didn't see the query builder yet. What do you think about having it in "Connecting to the database" section?

docs/intro/quickstart/connecting.rst
```diff
@@ -0,0 +1,94 @@
...
+  .. code-block:: sh
+
+    $ npx tsx test.ts
+    [ 'Hello from Gel!' ]
+    ┌─────────┬────────────────────────────────────────┬────────────┐
+    │ (index) │ id                                     │ name       │
+    ├─────────┼────────────────────────────────────────┼────────────┤
+    │ 0       │ 'f4cd3e6c-ea75-11ef-83ec-037350ea8a6e' │ 'I am one' │
+    │ 1       │ 'f4cf27ae-ea75-11ef-83ec-3f7b2fceab24' │ 'I am two' │
+    └─────────┴────────────────────────────────────────┴────────────┘
```
beerose, line 94: Should we also show where and how that client setup is done in the cloned flashcards app? As the next section (working with the data) is using it - maybe would be good to have it here so it's not surprising where the client is coming from.

docs/intro/quickstart/setup.rst
scotttrinh: Yeah, I originally had this closer to the point where you first start using the query builder, but now we've paired down the actual migrations to just two: the initial one, and the one at the very end. That means that setting up the hook only pays off for the very last migration.

So, if I moved it _after_ the modeling section, I was having to tell the reader to set up the hook, but then since they've already migrated, you'll have to run this command manually anyway. That wasn't a big deal when were where doing a lot more migrations because you'd see it paying off as you went along. Another option is to move it to just before the inheritance migration. It feels a bit more out of place there, to be honest.

Oh! I just had an idea while writing all of this out: We could have this hook already set up in the template, and we can talk about the hook in the section where we talk about what a migration is. wdyt?

docs/intro/quickstart/connecting.rst
scotttrinh: Do you mean show the `/lib/gel.ts` file here? It's pretty barebones now:

```ts
import { createClient } from "gel";

export const client = createClient();
```

Maybe I can have a small note in the first query about importing the client from the `@/lib/gel` module. Just say "In the template we've exported an initialized `Client` instance from the module `/lib/gel.ts`, so import that client run the queries below." or something? Is that worth pointing out? Seems kind of trivial and if the reader is curious they have the source code already locally. I'm fine to add this though if it feels important or helpful.

docs/intro/quickstart/setup.rst
beerose: > Oh! I just had an idea while writing all of this out: We could have this hook already set up in the template, and we can talk about the hook in the section where we talk about what a migration is. wdyt?

This is great! When talking about project structure (where you call `tree`) we can mention that we'll talk about hooks later.

docs/intro/quickstart/connecting.rst
beerose: OR maybe in the "Connecting to the database", you can mention "in our app, the client setup is in ..., go take a look".

docs/intro/quickstart/setup.rst
scotttrinh: awesome, will fix, thanks for the nudge.

docs/intro/quickstart/setup.rst
scotttrinh: I'm not sure why I didn't think of this, but I just decided to show the highlighted files on the right side. Makes sense to me!

docs/intro/quickstart/setup.rst
```diff
@@ -0,0 +1,89 @@
+.. _ref_quickstart_setup:
+
+===========================
+Setting up your environment
+===========================
+
+.. edb:split-section::
+
+  Use git to clone the Next.js starter template into a new directory called ``flashcards``. This will create a fully configured Next.js project and a local |Gel| instance with an empty schema. You will see the database instance being installed and the project being initialized. You are now ready to start building the application.
```
diksipav, line 9: link to github

docs/intro/quickstart/setup.rst
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+.. edb:split-section::
+
+  Fun! You will create a proper data model for the application in the next step, but for now, take a look around the project you've just created. Most of the project files will be familiar if you've worked with Next.js before. Here are the new files that integrate |Gel|:
+
+  - ``gel.toml``: The configuration file for the Gel project instance. Notice that we have a ``hooks.migration.apply.after`` hook that will run ``npx @gel/generate edgeql-js`` after migrations are applied. This will generate the query builder code that you'll use to interact with the database. More details on that to come!
+  - ``dbschema/``: This directory contains the schema for the database, and later supporting files like migrations, and generated code.
+  - ``dbschema/default.gel``: The default schema file that you'll use to define your data model. It is empty for now, but you'll add your data model to this file in the next step.
+  - ``lib/gel.ts``: A utility module that exports the Gel client, which you'll use to interact with the database.
```
diksipav, line 64: maybe just say `files` instead of `new files` (all files are new), I am biased to think that new files are generated with `gel project init`, but here actually nothing new is created because it already exist in the cloned folder, ofc I know that new users don't have this bias... I just find it weird to say `new files`, maybe is just me.

docs/intro/quickstart/modeling.rst
```diff
@@ -0,0 +1,101 @@
+.. _ref_quickstart_modeling:
+
+=================
+Modeling the data
+=================
+
+.. edb:split-section::
+
+  The flashcards application has a simple data model, but it's interesting enough to get a taste of many of the features of the |Gel| schema language. You have a ``Card`` type that describes a single flashcard, which for now contains two required string properties: ``front`` and ``back``. Each ``Card`` belongs to a ``Deck``, and there is an explicit ordering to the cards in a given deck.
```
diksipav, line 9: `of many features` instead of `of many of the features`, or we can say sth along the lines: The flashcards application has a straightforward data model, yet it's engaging enough to showcase many features of the Gel schema language.

docs/intro/quickstart/modeling.rst
```diff
@@ -0,0 +1,101 @@
...
+
+  Looking at the mock data, you can see this structure in the JSON.
+
+  .. code-block:: typescript
+
+    interface Card {
```
diksipav, line 15: in the tutorial app we define `RawJSONCard` and `RawJSONDeck` in both `lib/models` and in the `app/model`, is that needed?

docs/intro/quickstart/modeling.rst
```diff
@@ -0,0 +1,101 @@
...
+
+=================
+Modeling the data
+=================
+
+.. edb:split-section::
+
+  The flashcards application has a simple data model, but it's interesting enough to get a taste of many of the features of the |Gel| schema language. You have a ``Card`` type that describes a single flashcard, which for now contains two required string properties: ``front`` and ``back``. Each ``Card`` belongs to a ``Deck``, and there is an explicit ordering to the cards in a given deck.
+
+  Looking at the mock data, you can see this structure in the JSON.
```
diksipav, line 11: maybe to mention here that we mean the file `deck-edgeql.json`?

docs/intro/quickstart/modeling.rst
```diff
@@ -0,0 +1,101 @@
...
+    +   };
+      };
+
+.. edb:split-section::
+
+  Congratulations! This first version of the data model's schema is *stored in a file on disk*. Now you need to signal the database to actually create types for ``Deck`` and ``Card`` in the database.
+
+  To make |Gel| do that, you need to do two quick steps:
+
+  1. **Create a migration**: a file with a list of low-level instructions.
```
diksipav, line 57: I'd explain this a bit more clearly: a file containing a set of low level instructions that define how the database schema should change. It records any additions, modifications, or deletions to your schema in a way that the database can understand.

and something along the lines:
**Apply the migration**: This executes the migration file, instructing Gel to implement the recorded changes in the database. Essentially, this step updates the database structure to match your defined schema, ensuring that the Deck and Card types are created and ready for use.

docs/intro/quickstart/connecting.rst
```diff
@@ -0,0 +1,99 @@
...
+Connecting to the database
+==========================
+
+.. edb:split-section::
+
+  Before diving into the application, let's take a quick look at how to connect to the database from your code. We will intialize a client and use it to make a simple, static query to the database, and log the result to the console.
+
+  .. note::
+
+    Notice that the ``createClient`` function isn't being passed any connection details. How does |Gel| know how to connect to the database you set up earlier? When we ran ``npx gel project init`` earlier, the CLI created credentials for the local database and stored them in a well-known location. When you initialize your client with ``createClient()``, |Gel| will check the places it knows about for connection details.
```
diksipav, line 13: maybe: the CLI created credentials for the local database and stored them in a well-known location `on disk`. When you initialize your client with createClient(), `Gel will check these known locations on your machine `for connection details."

docs/intro/quickstart/connecting.rst
```diff
@@ -0,0 +1,99 @@
...
+
+    With |Gel|, you do not need to come up with your own scheme for how to build the correct database connection credentials and worry about leaking them into your code. You simply use |Gel| "projects" for local development, and set the appropriate environment variables when you're ready to deploy, and the client knows what to do!
```
diksipav, line 15: maybe we include link to the docs with environment variables or connections docs in general

docs/intro/quickstart/modeling.rst
beerose, line 9: It looks like there's no `order` in the TypeScript version

docs/intro/quickstart/connecting.rst
```diff
@@ -0,0 +1,99 @@
...
+
+    $ npx tsx test.ts
+    [ 'Hello from Gel!' ]
+
+.. edb:split-section::
+
+
+  With TypeScript, there are three ways to run a query: use a string EdgeQL query, use the ``queries`` generator to turn a string of EdgeQL into a TypeScript function, or use the query builder API to build queries dynamically in a type-safe manner. In this tutorial, you will use the TypeScript query builder API.
+
+  This query builder must be generated any time the schema changes, so a hook has been added to the ``gel.toml`` file to generate the query builder any time the schema is updated. Moving beyond this simple query, use the query builder API to insert a few ``Deck`` objects into the database, and then select them back.
```
diksipav, line 49: file to `re`generate the query

docs/intro/quickstart/working.rst
```diff
@@ -0,0 +1,385 @@
...
+    -           back: card.back,
+    -           order: card.order,
+    -         })
+    -         .run(tx);
+    -
+    -       cardIds.push(createdCard.id);
+    -     }
+    -
+    -     await e
+    -       .params({ cardIds: e.array(e.uuid) }, (params) =>
```
diksipav, line 174: we didn't have params in the previous step, we still used cardIdsLiteral

docs/intro/quickstart/setup.rst
scotttrinh: Yeah, I meant "new to you" as in: "you already recognize most of the files since they're Next.js app router files, here are the ones you do not recognize". I'll reword!

docs/intro/quickstart/modeling.rst
scotttrinh: Yeah, we don't need both, PRs welcome, but that's more of an implementation detail than something related to the quickstart text itself. We don't actually mention those model modules directly.

docs/intro/quickstart/modeling.rst
scotttrinh: `decks.json` is where the data lives once you import the `deck-edgeql.json`, so I'd like to keep that rather than make the actions module more complicated by needing to check if it exists, etc.

> maybe to mention here that we mean the file deck-edgeql.json?

Yeah, that's a good call. Maybe show a bit of the data itself.

docs/intro/quickstart/connecting.rst
scotttrinh: Yeah, good idea. Once we have those written, I'll try to circle back and add links.

docs/intro/quickstart/modeling.rst
scotttrinh: Right: it's just the order of the array itself. That's why I didn't mention an `order` property, just said "there _is_ an explicit ordering". I'll maybe make this more clear that I mean arrays are ordered and we want to keep this order in the database.

docs/intro/quickstart/working.rst
```diff
@@ -0,0 +1,385 @@
...
+    +       e.select(e.Deck, (d) => ({
+    +         filter_single: e.op(d.id, "=", params.deckId),
+    +       }))
+    +     );
+    +
+    +     const order = e.cast(e.int64, e.max(deck.cards.order));
+    +     return e.insert(e.Card, {
+    +       front: params.front,
+    +       back: params.back,
+    +       deck: e.cast(e.Deck, params.deckId),
```
diksipav, line 278: is this supposed to work? deck is not part of the Card type

docs/intro/quickstart/working.rst
```diff
@@ -0,0 +1,385 @@
...
+      }
+
+.. edb:split-section::
+
+  Next, you will update the Server Actions for each ``Deck`` object: ``updateDeck``, ``addCard``, and ``deleteCard``. Start with ``updateDeck``, which is the most complex because it is dynamic. You can set either the ``title`` or ``description`` fields in an update. Use the dynamic nature of the query builder to generate separate queries based on which fields are present in the form data.
+
+  This may seem a bit intimidating at first, but the key to making this query dynamic is the ``nameSet`` and ``descriptionSet`` variables. These variables conditionally add the ``name`` or ``description`` fields to the ``set`` parameter of the ``update`` call.
+
+  .. code-block:: typescript-diff
+    :caption: app/(authenticated)/deck/[id]/actions.ts
```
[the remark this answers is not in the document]
scotttrinh: ahh, this is a vestige from when we also had access control in the tutorial. Will fix.

docs/intro/quickstart/working.rst
scotttrinh: yeah, Aleksandra also caught that. We used to have a slightly different data model, but I need to update this one.

PR #8270, opened 2025-01-29
Remove aws-lc-* packages

mmastrac:
These were causing issues on certain Ubuntu runners. As a bonus, it also removes a number of build support crates.


msullivan (approved):
Thank you!

PR #8275, opened 2025-01-29
config: Add support for remapping Postgres configs into Gel enums

elprans:
We have a long-standing rule that configuration parameters should avoid
using a boolean type and use enums instead (though we've not generally
been super diligent about this always).  Postgres, on the other hand,
has lots of boolean settings.  To solve this, teach the config framework
how to remap backend config values onto arbitrary frontend enums.  In
general the below is all that is required:

    class EnabledDisabledEnum(enum.StrEnum):
        Enabled = "Enabled"
        Disabled = "Disabled"

    class EnabledDisabledType(
        EnumScalarType[EnabledDisabledEnum],
        edgeql_type="cfg::TestEnabledDisabledEnum",
    ):
        @classmethod
        def get_translation_map(cls) -> Mapping[EnabledDisabledEnum, str]:
            return {
                EnabledDisabledEnum.Enabled: "true",
                EnabledDisabledEnum.Disabled: "false",
            }

### DIFF
--- a/edb/ir/statypes.py
+++ b/edb/ir/statypes.py
@@ -580,6 +580,9 @@ class EnumScalarType(
     """Configuration value represented by a custom string enum type that
     supports arbitrary value mapping to backend (Postgres) configuration
     values, e.g mapping "Enabled"/"Disabled" enum to a bool value, etc.
+
+    We use SingleParametricType to obtain runtime access to the Generic
+    type arg to avoid having to copy-paste the constructors.
     """
 
     _val: E
--- a/edb/ir/statypes.py
+++ b/edb/ir/statypes.py
@@ -18,17 +18,34 @@
 
 
 from __future__ import annotations
-from typing import Any, Optional
+from typing import (
+    Any,
+    Callable,
+    ClassVar,
+    Generic,
+    Mapping,
+    Optional,
+    Self,
+    TypeVar,
+)
 
 import dataclasses
+import datetime
+import decimal
+import enum
 import functools
 import re
 import struct
-import datetime
+import uuid
 
 import immutables
 
 from edb import errors
+from edb.common import parametric
+from edb.common import uuidgen
+
+from edb.schema import name as s_name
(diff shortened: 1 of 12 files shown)


edb/pgsql/metaschema.py
```diff
@@ -3444,6 +3444,53 @@ def __init__(self) -> None:
         )
 
 
+class PostgresJsonConfigValueToGelConfigValueFunction(
+    trampoline.VersionedFunction,
+):
+    """Convert a Postgres config value to Gel config value.
```
msullivan, line 3450: I think my inclination is to keep calling it "edgedb" as a "codename" in internal code like this, but I don't feel strongly.

Just makes it a little easier to be consistent in function names and stuff, though obviously it is kind of fucked no matter what.

msullivan (approved):
An epic yak shave.

Probably will be a bit of a pain in my ass to backport this but I think I have it coming after sending you down this rabbit hole.

edb/ir/statypes.py
```diff
@@ -503,8 +544,192 @@ def __repr__(self) -> str:
...
+        cls,
+        typeargs: Optional[tuple[type, ...] | type],
+        val: E,
+    ) -> Self:
+        if typeargs is None or cls.is_anon_parametrized():
+            obj = cls(val)
+        else:
+            obj = cls[typeargs](val)  # type: ignore[index]
+
+        return obj
```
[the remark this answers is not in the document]
elprans: `Parametric` basically gives us runtime access to the generic type arg and lets us avoid copy-pasting the constructors.  Given a bunch of enums are added #8276 (and possibly more settings later) I figured it's worth complicating the base type to make implementations essentially glorified mappings.

edb/pgsql/metaschema.py
elprans: We also have the `backend` (postgres) and `frontend` (edgedb-server) used around (which I think I prefer)

PR #8277, opened 2025-01-30
[5.x] TEST: Remove insertion of extra repairs in patch queue

msullivan:
I'm developing this against 5.x, so we can just test it against the
actual patches.
We won't actually use it for 5.x, though.


msullivan:
Alright, full success.
Not going to merge it, though.

Issue #5805, opened 2023-07-17
Tests for the patch system *itself*


msullivan:
We have the "Tests of patching old EdgeDB Versions" CI job that does a pretty good job testing our concrete uses of the patching system, but we don't have any general tests of the patching system itself.

This means that when I break some part of the patch system, the breakage will just lurk until some patch actually uses it. For an example, see https://github.com/edgedb/edgedb/pull/5804

The setup should be something like:

1. Bootstrap a database with the current checkout
2. Inject a variety of patches into `pgsql/patches.py`
3. Run tests

PR #8278, opened 2025-01-30
[6.x] Remove insertion of extra repairs in patch queue

msullivan:
Currently, if there is *any* repair in a patch queue, we insert a
repair after *every* patch that modifies the std schema.

This was in combination with deferring all repairs until the last one
in the queue, which allowed us to avoid needing to load the current
standard schema in order to do a repair. But we implemented support
for doing that as part of patching user extensions, so merge all of
the schema repair code into that.

I tested this on 5.x's substantial collection of complex patches (see https://github.com/edgedb/edgedb/pull/8277), and
then forward ported to master, and then backported this to 6.x.

### DIFF
--- a/edb/pgsql/patches.py
+++ b/edb/pgsql/patches.py
@@ -43,24 +43,6 @@ def get_version_key(num_patches: int):
         return f'_v{num_major}'
 
 
-def _setup_patches(patches: list[tuple[str, str]]) -> list[tuple[str, str]]:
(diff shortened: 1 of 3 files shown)


msullivan:
Forward porting had enough conflicts that I'm doing a synthetic test for 6.x.

It would be really nice to have tests for the patching system: https://github.com/edgedb/edgedb/issues/5805

PR #8281, opened 2025-01-30
update nextjs guides

beerose:
Since we will have a new quickstart tutorial for building apps with Next.js, I merged our two existing Next.js tutorials into a short guide showing the setup and basics of using EdgeDB with Next.


docs/guides/tutorials/nextjs.rst
```diff
@@ -0,0 +1,270 @@
...
+    );
+  }
+
+With API Routes
+===============
+
+Create an API route and fetch data from the client side:
+
+.. code-block:: tsx
+  :caption: pages/api/posts.ts
```
scotttrinh, line 171: We should update this to be an app-router Route Handler, right?

docs/guides/tutorials/nextjs.rst
```diff
@@ -0,0 +1,270 @@
...
+          </article>
+        ))}
+      </div>
+    );
+  }
+
+Deployment
+-----------
+
+First, add a ``prebuild`` script to your ``package.json`` to generate the query builder during deployment:
```
scotttrinh, line 223: We should move this to `gel.toml` hooks

docs/guides/tutorials/nextjs.rst
beerose: I was thinking to have two options there - with app router and with pages router. I'll make it clear in the title.

scotttrinh: Do we want to support the pages router still? There is nothing particularly special about this code, and if we show an app router Route Handler, it should be pretty easy for someone two write the equivalent pages router API route, I would think.

docs/guides/tutorials/nextjs.rst
```diff
@@ -0,0 +1,270 @@
...
+.. code-block:: bash
+
+  $ npx @edgedb/generate edgeql-js
+
+The command introspects your schema and generates a query builder in the ``dbschema/edgeql-js`` directory.
+
+Using EdgeDB in Next.js
+-----------------------
+
+EdgeDB with React Server Components (App Router)
```
scotttrinh, line 133: Can we show both server components and server actions?

docs/guides/tutorials/nextjs.rst
beerose: But don't we need this, too? So that t works with a cloud provider when building the app. As I understand, the hooks will work after applying the migration etc, so if I deploy to e.g. Vercel and connect to the Cloud with env variables, it won't run the generator.

scotttrinh: yeah, good point, we should do both. We probably need some additional direction to get it to even _work_ during deployment since it needs access to a live database. I think it works fine in the Vercel integration which I'm pretty sure sets the `EDGEDB_INSTANCE` and `EDGEDB_SECRET_KEY` environment variables correctly.

Maybe this section should be merged into the sections below?

PR #8292, opened 2025-02-04
Mark the assert_* functions as being Immutable

msullivan:
This seems basically semantically correct to me.

The motivation is that we use `assert_exists(<typname>{})` as a dummy
expression when making schema changes that affect schema expressions,
and currently it is possible to get errors due to that expression
being `Stable`.

### DIFF
--- a/edb/buildmeta.py
+++ b/edb/buildmeta.py
@@ -60,7 +60,7 @@ from edb.common import verutils
 # The merge conflict there is a nice reminder that you probably need
 # to write a patch in edb/pgsql/patches.py, and then you should preserve
 # the old value.
-EDGEDB_CATALOG_VERSION = 2024_01_28_00_00
+EDGEDB_CATALOG_VERSION = 2024_02_03_00_00
 EDGEDB_MAJOR_VERSION = 7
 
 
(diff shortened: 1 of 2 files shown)


aljazerzen:
Well, they are immutable. Any reason why they were marked as "stable" before?

dnwpark (approved):
Looking at the git history, these functions were initially created as `Stable` but no explicit reason was given. Probably as a placeholder value until a "correct" one could be properly thought out?

As an aside, `std::assert` probably ought to be `Immutable` as well.

elprans:
The `Stable` comes from an unfortunate fact that Postgres uses volatility for optimization, i.e. it may evaluate `immutable` expressions _early_ and regardless of the control flow, so e.g something like `if (foo) then ... else assert()` may trigger the assert even if `foo` is `true`.  That said, this is only applicable to SQL functions and we inline the assertions, so this _should_ be safe.

msullivan:
> Looking at the git history, these functions were initially created as `Stable` but no explicit reason was given. Probably as a placeholder value until a "correct" one could be properly thought out?
> 
> As an aside, `std::assert` probably ought to be `Immutable` as well.

Sigh, but marking it Immutable in the obvious way fails, because that one *isn't* implemented by inlining.

PR #8312, opened 2025-02-06
Fix default_transaction_isolation on system/database level

fantix:
`client` GucSource (backend connection parameter) has higher priority than `database` or `configuration file` (system). Now that `default_transaction_isolation` is exposed, it has to be removed from the backend connection parameters. Instead, it now becomes a permanent system config that always has a value.

This breaks remote backends without config file access unless they are preconfigured with `default_transaction_isolation = serializable` system-wide.

Refs #8276 

- [ ] Add test

### DIFF
--- a/edb/edgeql/compiler/config.py
+++ b/edb/edgeql/compiler/config.py
@@ -146,7 +146,24 @@ def compile_ConfigReset(
             span=expr.span,
         )
 
-    elif isinstance(info.param_type, s_objtypes.ObjectType):
+    if (
+        expr.scope == qltypes.ConfigScope.INSTANCE and
+        info.backend_setting == 'default_transaction_isolation'
+    ):
+        # Special case for default_transaction_isolation:
+        # Gel default is `serializable`, while PG default is `read committed`.
+        default = info.ptr.get_default(ctx.env.schema)
+        if default is not None:
+            return compile_ConfigSet(
+                qlast.ConfigSet(
+                    name=expr.name,
+                    scope=expr.scope,
+                    expr=default.parse(),
+                ),
+                ctx=ctx,
+            )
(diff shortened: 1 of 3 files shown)


edb/edgeql/compiler/config.py
```diff
@@ -146,7 +146,24 @@ def compile_ConfigReset(
             span=expr.span,
         )
 
-    elif isinstance(info.param_type, s_objtypes.ObjectType):
+    if (
+        expr.scope == qltypes.ConfigScope.INSTANCE and
+        info.backend_setting == 'default_transaction_isolation'
+    ):
+        # Special case for default_transaction_isolation:
+        # Gel default is `serializable`, while PG default is `read committed`.
```
msullivan, line 154: Call this out as a `HACK` in the comments.
Some day we may want this to be a general mechanism?
(Not today.)

msullivan (approved):

msullivan:
Actually wait, I don't fully understand; when does `default_transaction_isolation` get set, in the normal flow?

fantix:
> Actually wait, I don't fully understand; when does `default_transaction_isolation` get set, in the normal flow?

During bootstrap, backend settings that have defaults are applied to the PG config file?

edb/edgeql/compiler/config.py
fantix: Yeah, now that I'm thinking more about it, maybe we should *always* reset the system backend settings to their Gel defaults explicitly, as we cannot rely on the backend defaults, like users could put anything in `postgresql.conf` and run Gel with `--backend-dsn`.

msullivan:
Wait --- cloud doesn't have configfile access, right?

PR #8315, opened 2025-02-06
[sql] Fix non-normalized queries over native protocol

msullivan:
Queries where we had to fall back to the non-normalized version were
broken under the native protocol, because we were still feeding the
arguments to postgres.

Also, make `pg_get_serial_sequence` force non-normalization, since it
ignores its arguments. It still fails if actual params are used,
though.


elprans (approved):

aljazerzen (commented):
Thanks

PR #8318, opened 2025-02-07
Handle session state sync properly with tx-behavior settings

fantix:
See the comments added in `pgcon.pyx` for details

- [x] Add tests

### DIFF
--- a/tests/test_server_proto.py
+++ b/tests/test_server_proto.py
@@ -2069,6 +2069,46 @@ class TestServerProto(tb.QueryTestCase):
 
         self.assertEqual(await self.con.query_single('SELECT 42'), 42)
 
+    async def assert_tx_isolation_and_default(
+        self, expected: str, *, default : str | None = None, conn=None
+    ):
+        if conn is None:
+            conn = self.con
+        if default is None:
+            default = expected
+        self.assertEqual(
+            await conn.query_single('''
+                select (
+                    <str>sys::get_transaction_isolation(),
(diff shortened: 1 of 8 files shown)


fantix:
This makes the state sync more complex, making me wonder if we should maintain such complications. An easy simplification is to always commit (add a SYNC message) after syncing the state; in that way, we can drop many special cases (`if` branches, state trackings) from this PR as well as previous code.

msullivan (approved):
I don't think I'd object to syncing unconditionally either.

tests/test_server_proto.py
```diff
@@ -2367,6 +2341,31 @@ async def test_server_proto_tx_31(self):
...
+            await con2.query('''
+                CONFIGURE SESSION
+                    SET default_transaction_isolation := 'RepeatableRead';
+            ''')
+            await self.assert_tx_isolation_and_default(
+                'RepeatableRead', conn=con2
+            )
+
+            # Try a few times back and forth - this should be enough to hit
+            # the same backend connection.
```
msullivan, line 2357: Of course, that being true indicates that maybe we should do a little more work to try to select connections with matching state more consistently?

(Not now, though. And unfortunately doing that will make these bugs much subtler, too.)

fantix: Yeah, the connection pool could also be smarter to open new backend connections on state mismatch when the load is low. (cc @mmastrac)

PR #8320, opened 2025-02-08
Use sys_pgcon for long-term advisory locks

fantix:
* Fixes the issue of advisory locks getting leaked back to the connection pool.
* Fixes the `test_ext_ai_indexing_*` failures in CI.

Refs [#7183](https://github.com/edgedb/edgedb/pull/7183#discussion_r1947900380), [sample run](https://github.com/edgedb/edgedb/actions/runs/13219234251)

### DIFF
--- a/edb/server/protocol/ai_ext.py
+++ b/edb/server/protocol/ai_ext.py
@@ -363,39 +363,61 @@ async def _ext_ai_fetch_active_models(
     return result
 
 
-_EXT_AI_ADVISORY_LOCK_LOCK = b"3987734540"
-_EXT_AI_ADVISORY_LOCK = b"3987734541"
+# The _ext_ai_lock() is a long-term lock held in the system pgcon. It is used
+# in the index builder job above guarding multiple alternating database pgcons
+# and outgoing HTTP requests (free up pgcons while waiting for a response from
+# external services), so that different Gel tenants on the same backend
+# run this job exclusively.
+#
+# The following implementation is also safe to be used by multiple tasks within
+# the same tenant (though at the time of writing, there is only one such task
+# per tenant). To achieve this, we added an extra query on pg_locks to check if
+# it's already held by another task, because advisory locks allow reentrancy
+# from the same session (the same sys_pgcon). And to avoid racing conditions,
+# we use another advisory lock over the 2 queries of check-and-lock in the
+# local session. This also means, one must use _ext_ai_lock() instead of an
+# individual lock of the 2 locks here to avoid misuse.
+#
+# If you are editing the magic numbers here: make sure it fits in a Postgres
+# Oid type (uint32), or you'll need to change the `classid` query below.
+_EXT_AI_ADVISORY_LOCK = b"3987734540"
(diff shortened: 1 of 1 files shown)


edb/server/protocol/ai_ext.py
```diff
@@ -363,22 +363,51 @@ async def _ext_ai_fetch_active_models(
...
 
 
 async def _ext_ai_lock(
+    tenant: srv_tenant.Tenant,
     pgconn: pgcon.PGConnection,
 ) -> bool:
-    b = await pgconn.sql_fetch_val(
-        b"SELECT pg_try_advisory_lock(" + _EXT_AI_ADVISORY_LOCK + b")")
-    return b == b'\x01'
+    async def _try_xact_lock(lock: bytes) -> bool:
```
mmastrac, line 374: Could we skip the second lock and just check pg_locks for reentrancy checks?

edb/server/protocol/ai_ext.py
```diff
@@ -363,22 +363,51 @@ async def _ext_ai_fetch_active_models(
...
+        # Acquire the outer lock to operate on the real lock
+        if await _try_xact_lock(_EXT_AI_ADVISORY_LOCK_LOCK):
+            # If success, acquire the real lock in the current session
+            # temporarily to make sure it's not locked by system session
+            sp = b"ext_ai_lock_sp"
+            await pgconn.sql_execute(b"SAVEPOINT " + sp)
+            if await _try_xact_lock(_EXT_AI_ADVISORY_LOCK):
+                # If success, move the lock to the system session
+                await pgconn.sql_execute(b"ROLLBACK TO SAVEPOINT " + sp)
+                async with tenant.use_sys_pgcon() as syscon:
```
msullivan, line 391: What is the try-lock for?
And couldn't it get taken by another connection after the ROLLBACK but before the lock?
Is that a problem?

edb/server/protocol/ai_ext.py
fantix: I think yes!

edb/server/protocol/ai_ext.py
fantix: > What is the try-lock for?

It checks if the lock is being held by another worker (through sys_pgcon). We cannot acquire this lock directly in sys_pgcon because, say if another worker is already holding this lock through sys_pgcon, re-acquiring the lock will not block (or fail with the try_* method) from the same sys_pgcon session.

> And couldn't it get taken by another connection after the ROLLBACK but before the lock?

That's guaranteed not to happen by the outter/first lock, unless other code misused the same advisory lock ID.

msullivan (approved):
Alright, I understand.
Probably could use some more comments, because the protocol here is subtle.
(Or are you changing the protocol per @mmastrac's comments?)

edb/server/protocol/ai_ext.py
mmastrac: I don't think this is necessary for the PR to land -- we could always address in a follow-up.

PR #8325, opened 2025-02-10
Try doing retries in test_ext_ai_indexing_02

msullivan:
The more general fix is that assert_query_result ought to do retries
when not in a transaction, of course, but let's try to get this
release out.

This is a forward-port from something already on the 6.x branch.

### DIFF
--- a/tests/test_ext_ai.py
+++ b/tests/test_ext_ai.py
@@ -20,6 +20,8 @@ import json
 import pathlib
 import unittest
 
+import edgedb
+
 from edb.server.protocol import ai_ext
 from edb.testbase import http as tb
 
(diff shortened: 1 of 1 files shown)


tests/test_ext_ai.py
```diff
@@ -255,20 +257,32 @@ async def test_ext_ai_indexing_02(self):
...
-            )
+            async for tr in self.try_until_succeeds(
+                ignore=(
+                    edgedb.TransactionConflictError,
+                    edgedb.TransactionSerializationError,
+                ),
+                timeout=30.0,
+            ):
+                async with tr:
+                    try:
```
scotttrinh, line 268: Do we need this extra `try`/`except` wrapping?

msullivan: oh, lol

scotttrinh: Definitely makes it easier to find this particular error.

scotttrinh (approved):

PR #8327, opened 2025-02-10
Drop tx isolation of interal queries in ext::ai to RepeatableRead

elprans:
Avoid unnecessary isolation load.  We already do this in other
internal loops.


msullivan (approved):

msullivan:
Tag for backport?

dnwpark (approved):

PR #8332, opened 2025-02-11
Implement caching for JWKSet

mmastrac:
In the previous HTTP rewrite, we accidentally removed caching for JWK keys and OIDC configurations. This restores it using an in-memory cache.

This adds a cache parameter to the Python-side `get(...)`.

The cache is enabled only for OIDC and JWK discovery requests -- the assumption is that we'd prefer fresh user information when fetching during login over speed.


scotttrinh (approved):
🙏

PR #8340, opened 2025-02-14
Improve parser recovery for CREATE BRANCH

aljazerzen:


tests/test_edgeql_syntax.py
```diff
@@ -4168,6 +4168,14 @@ def test_edgeql_syntax_ddl_branch_06(self):
         CREATE DATA BRANCH foo FROM bar;
         """
 
+    @tb.must_fail(errors.EdgeQLSyntaxError,
+                  "Missing keyword 'EMPTY'",
```
1st1, line 4172: I think we need to improve this a bit further :/

We have a bunch of different variants here: https://docs.edgedb.com/database/reference/admin/branches#create-schema-branch

So it's "empty" or "schema" or "data", and ideally we need a human-friendly message just explaining in one sentence that there are a few variants and the user has to choose one.

aljazerzen: This is hard to do - it'd have to change how parser works and that would break a lot of current messages that we already have.

It is possible, it shouldn't make the parser any slower on the happy path, but it might make it slower in error recovery.

The main problem is that I don't have the time to work on this now.

msullivan: Can the `custom_errors` mechanism not be used to do this?

aljazerzen: It can be yes.

I wanted to solve this in general, so we could have corrected `CREATE DEFAULT` into `Missing one of "PROPERTY", "LINK" or "INDEX"`.

msullivan (approved):
Tag for backport?

PR #8342, opened 2025-02-15
Report the number of times a feature is used as a metric

msullivan:
Currently we only report how many branches it is used in total.

Also report how many branches have their schemas introspected
currently.  This is helpful here because we only get feature count
info from introspected branches.

### DIFF
--- a/edb/server/metrics.py
+++ b/edb/server/metrics.py
@@ -224,7 +224,7 @@ feature_used_branches = registry.new_labeled_gauge(
 )
 feature_used = registry.new_labeled_gauge(
     'feature_used_num_count_current',
-    'How many branches a schema feature is used by.',
+    'How many times a schema feature is used.',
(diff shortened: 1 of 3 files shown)


elprans (approved):

edb/server/metrics.py
```diff
@@ -211,11 +217,16 @@
...
 
-feature_used = registry.new_labeled_gauge(
+feature_used_branches = registry.new_labeled_gauge(
     'feature_used_branch_count_current',
     'How many branches a schema feature is used by.',
     labels=('tenant', 'feature'),
 )
+feature_used = registry.new_labeled_gauge(
+    'feature_used_num_count_current',
+    'How many branches a schema feature is used by.',
```
zackelan, line 227: can you update this description so it's not the same as the other one?

zackelan (approved):

Issue #7411, opened 2024-05-29
More language server feature work


msullivan:
We have an LSP "language server now", but should support more features.

This is a placeholder issue. @aljazerzen could you add a checklist?

aljazerzen:
We also need to look into how much work would be support Windows. Maybe via WSL, same as our portable edgedb-server is started.

aljazerzen:
Ref https://github.com/geldata/gel/pull/8509 and https://github.com/geldata/gel/pull/8513

TextDocumentDefinition TODO:
- [ ] cleanup the code on #8513 and merge it,
- [ ] go to pointers instead of the target type of the pointer,
- [ ] running in schema files (for computeds, index expression, rewrites, triggers and so on),
- [ ] go to WITH bindings,
- [ ] go to `std` definitions,

PR #8349, opened 2025-02-19
Schema diagnostics in language server

aljazerzen:
Ref #7411

- adds `edb ls` sub command, which can be invoked from the vscode extension directly, which makes development much easier,
- treats both `.esdl` and `.gel` files as schema files,
- keeps track of modified `dbschema/` files (even those that have not been saved to fs yet),
- reports compilation errors from schema (we have previously done that for .edgeql files only)

### DIFF
--- a/edb/language_server/__init__.py
+++ b/edb/language_server/__init__.py
@@ -24,7 +24,11 @@ T = TypeVar('T', covariant=True)
 E = TypeVar('E', covariant=True)
 
 
-@dataclasses.dataclass(kw_only=True, slots=True)
+@dataclasses.dataclass(kw_only=True, slots=True, frozen=True)
 class Result(Generic[T, E]):
     ok: Optional[T] = None
     err: Optional[E] = None
+
+
+def is_schema_file(path: str) -> bool:
+    return path.endswith(('.esdl', '.gel'))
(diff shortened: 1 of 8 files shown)


edb/language_server/__init__.py
```diff
@@ -15,3 +15,16 @@
...
+
+import dataclasses
+from typing import Generic, Optional, TypeVar
+
+
+T = TypeVar('T', covariant=True)
+E = TypeVar('E', covariant=True)
+
+
+@dataclasses.dataclass(kw_only=True, slots=True)
```
msullivan, line 27: Make it `frozen` also, or the `covariant` is unsound.

edb/language_server/main.py
```diff
@@ -84,29 +84,46 @@ def completions(params: lsp_types.CompletionParams):
...
-        ls.publish_diagnostics(document.uri, diagnostics, document.version)
-        return
-    assert ql_ast.ok
+    diagnostic_set: ls_server.DiagnosticsSet
 
     try:
-        if isinstance(ql_ast.ok, list):
-            diagnostics = ls_server.compile(ls, ql_ast.ok)
-            ls.publish_diagnostics(document.uri, diagnostics, document.version)
+        if doc_uri.endswith('.esdl') or doc_uri.endswith('.gel'):
```
msullivan, line 95: You could also do `doc_uri.endswith(('.esdl', '.gel'))`

PR #8354, opened 2025-02-20
Make \(name)-style string interpolation work for errmessage

msullivan:
This is (somewhat unfortunately) implemented by translating it to
the old {name} style used previously for errmessage.

It would be somewhat nicer to store the string using \(name)
style, but doing that loses the ability to prevent interpolation
by doing \\(name). (Since the lexed version is stored in that
case, we wouldn't be able to distinguish between the cases.)

### DIFF
--- a/edb/schema/constraints.py
+++ b/edb/schema/constraints.py
@@ -175,7 +175,12 @@ class Constraint(
     )
 
     errmessage = so.SchemaField(
-        str, default=None, compcoef=0.971, allow_ddl_set=True)
+        str,
+        default=None,
+        compcoef=0.971,
+        allow_ddl_set=True,
+        allow_interpolation=True,
+    )
 
     is_aggregate = so.SchemaField(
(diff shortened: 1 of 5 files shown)


msullivan:
Should get some docs too, I guess

aljazerzen (approved):
Hmm, interesting. So this implies that `errmessage` is a normal expression, that is evaluated in a context that contains `__subject__` and constraint params?

One might assume that you could do this:
```
errmessage := std::to_lower(__subject__.__type__.name);
```

msullivan:
> Hmm, interesting. So this implies that `errmessage` is a normal expression, that is evaluated in a context that contains `__subject__` and constraint params?
> 
> One might assume that you could do this:
> 
> ```
> errmessage := std::to_lower(__subject__.__type__.name);
> ```

I think what we are going for is "we always allow the same interpolation syntax", and hopefully people don't read too much more into it.

PR #8383, opened 2025-02-25
ORM docs added to guides.

vpetrovykh:
"Using EdgeDB with ..." section now has 4 entries corresponding to SQLAlchemy, SQLModel, Django, and Prisma.


vpetrovykh:
OK, we're trying to consolidate the generator scripts, so I'm closing this until we land the new version.

PR #8384, opened 2025-02-25
Increase recursion limit in compiler worker processes

msullivan:
I've been threatening to do this for a while and it is finally time, I
think. Compiling the (complex!) query for `DESCRIBE INSTANCE CONFIG AS
DDL` as part of `dump --all` has been failing.


msullivan:
some github tooling screwup

PR #8387, opened 2025-02-25
Update the `gel watch` docs.

vpetrovykh:
Add `[hooks]` and `[[watch]]` documentation.
Update the docs for the `gel watch` command as well.


docs/reference/cli/gel_watch.rst
```diff
@@ -5,25 +5,51 @@
...
 
-	gel watch
+    Any output that the triggered scripts produce will be shown in the
+    :gelcmd:`watch` console. This includes any error messages. So if you're
+    not seeing a change you've expected, check on the watch process to make
+    sure there aren't any unexpected errors in the triggered scripts.
 
-.. note::
+To learn about our recommended development migration workflow using
+:gelcmd:`watch`, read our :ref:`intro to migrations <ref_intro_migrations>`.
```
scotttrinh, line 28: I think we want to move this section to the `--migrate` section?

docs/reference/reference/gel_toml.rst
```diff
@@ -56,9 +141,22 @@ Example
     [project]
     schema-dir = "db/schema"
 
+    [hooks]
+    project.init.after="setup_dsn.sh"
+    branch.wipe.after=""
+    branch.switch.after="setup_dsn.sh"
+    schema.update.after="gel-orm sqlalchemy --mod compat --out compat"
```
scotttrinh, line 148: tiny nit: can we format these lines the same as the other lines:

```suggestion
    project.init.after = "setup_dsn.sh"
    branch.switch.after = "setup_dsn.sh"
    schema.update.after = "gel-orm sqlalchemy --mod compat --out compat"
```
Also, was there a reason we showed an empty hook?

vpetrovykh:
There is a discrepancy in the migration docs:

The "Migrations" section appearing under "Learn" does not mention the need to run `gel migrate --dev-mode` after creating a migration in `watch` mode.
However, the Migration GUIDE in the `Gel watch` section actually suggests that there are 2 steps to create and apply migrations in this mode: `gel migration create` + `gel migrate --dev-mode`.

I'm pretty sure that the current version of `gel watch --migrate` doesn't require the `gel migrate --dev-mode` since that's what it already does. But I wanted to double check whether I'm missing something before removing it.

scotttrinh (approved):

PR #8391, opened 2025-02-25
Re-structure docs to match structure on docs site

jaclarke:
There's no actual content changes in this PR, it's just renames and git is getting confused by edgeql -> reference/edgeql and reference/edgeql -> reference/reference/edgeql.

Related: https://github.com/geldata/edgedb.com/pull/970


scotttrinh (approved):
i-aint-reading-that

PR #8392, opened 2025-02-25
Mention that we dropped ``id`` from free objects in the changelog

msullivan:


docs/resources/changelog/6_x.rst
```diff
@@ -305,6 +305,11 @@ Additional changes
 EdgeQL
 ------
 
+* Free objects no longer have an ``id`` property. This is a breaking
+  change, though we expect the real-world impact to be minor. If you
+  need a dynamically generated new ID for free objects, it can be
+  added manually.
```
scotttrinh, line 311: Would it be too much to show an example here? Maybe

```suggestion
  added manually, using one of our UUID generation functions.
  
  .. code-block:: edgeql
    select {
        id := uuid_generate_v1mc(),
        property := 'Some string',
    }
```

scotttrinh (approved):
No blockers.

PR #8402, opened 2025-02-26
Fix migration failure if a DML function refers to required property.

dnwpark:
Splits altering DML functions into a paired DeleteFunction and CreateFunction. This fixes issues when adding/removing `required`, renaming, or changing the type of a pointer in the same migration as the function change.

For example, when migrating from:
```
  type Foo { required a: int64; }
  function insert_foo() -> Foo using ( insert Foo { a := 1 } );
```

to:
```
  type Foo {}
  function insert_foo() -> Foo using ( insert Foo {} );
```


dnwpark:
tests to be added.

PR #8404, opened 2025-02-26
Update the gel-go docs

fmoor:
gel-go introduces breaking changes from the old edgedb-go module. Mostly it is package organization that has changed. This change updates the docs to reflect the new layout.


scotttrinh (approved):
LGTM!

PR #8405, opened 2025-02-26
Update scope tests to use simple_scoping

msullivan:
Some tests I updated (either by modifying the query or the results),
some I deleted.

### DIFF
--- a/tests/test_edgeql_scope.py
+++ b/tests/test_edgeql_scope.py
@@ -29,6 +29,8 @@ from edb.tools import test
 class TestEdgeQLScope(tb.QueryTestCase):
     '''This tests the scoping rules.'''
 
+    NO_FACTOR = True
+
     SCHEMA = os.path.join(os.path.dirname(__file__), 'schemas',
                           'cards.esdl')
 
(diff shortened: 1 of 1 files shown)


tests/test_edgeql_scope.py
```diff
@@ -1097,16 +210,28 @@ async def test_edgeql_scope_binding_02a(self):
                 name := {'Alice', 'Bob'},
                 L := (name, (
                     SELECT User
-                    FILTER .name = name
```
aljazerzen, line 1100: Ok, so `.name` and `name` shouldn't have been factored even before this change.

Why did this work before? Had we allowed FILTER condition to be a set and applied `any` implicitly?

aljazerzen (approved):

tests/test_edgeql_scope.py
msullivan: `name` was factored with that use of `name` in the tuple above

But also we *do* allow `FILTER` conditions to be a set and apply `any` implicitly, though in 6.0 we now emit a warning; I don't think is happening here, though, since it name is getting factored

aljazerzen: Oh, I missed the `(name, (`, right.

PR #8407, opened 2025-02-26
Use correct docker repository when checking tags

elprans:


fantix (approved):

msullivan:
Does this need to be backported?

Is it ready to merge?

elprans:
> Does this need to be backported?

To 6.x, yes.

PR #8411, opened 2025-02-27
Fix Magic Link guide

scotttrinh:
- Add missing `isSignUp` query parameter to the callback URL
- Correctly get the `identity_id` from the token exchange response
- Pass the required `callback_url` to the send magic link request

### DIFF
--- a/docs/reference/auth/magic_link.rst
+++ b/docs/reference/auth/magic_link.rst
@@ -176,6 +176,7 @@ Sign up
        }
 
        const registerUrl = new URL("magic-link/register", GEL_AUTH_BASE_URL);
+       const callbackUrl = new URL("auth/magic-link/callback", "http://localhost:${SERVER_PORT}");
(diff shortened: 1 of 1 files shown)


gracecodes144:
Oh damn, I missed this, it went by so fast! I also found the `provider` argument to the various endpoints a little confusing and lacking in documentation. Sorry to reset the clock 🙄

scotttrinh:
@gracecodes144 

>  I also found the provider argument to the various endpoints a little confusing and lacking in documentation.

Yeah, they're really an implementation detail and I don't think there is much to say about them other than "please provide this value". We have plans for writing proper documentation for the HTTP API that the auth extension exposes which might say a bit more there, but they're really just constants that help with some HTTP routing on our end. They're a bit more prevalent and useful in the OAuth flows where we need to match which OAuth provider's logic we need to match, but all of the auth flows have a "provider" type.

PR #8418, opened 2025-02-28
Add workaround to bare metal deployment guide

fmoor:


docs/resources/guides/deployment/bare_metal.rst
```diff
@@ -67,6 +67,18 @@ Install the Gel package.
 Enable a systemd unit
 =====================
 
+We released gel-server-6 versions 6.0 and 6.1 without renaming the user and
+group that runs the server. This will be fixed in version 6.2. For the versions
+with the wrong user you will have to create the group, user, and data directory
+before enabling the systemd unit.
```
beerose, line 73: How about:

```suggestion
Versions 6.0 and 6.1 of ``gel-server-6`` retain the previous user and group configurations for server execution. This configuration will be updated in version 6.2.

For systems running versions 6.0-6.1, please ensure you create the required group, user, and data directory before enabling the ``systemd`` unit. This additional setup step will be eliminated in version 6.2.
```

beerose (approved):

PR #8425, opened 2025-03-03
Fix CentOS/RHEL deployment guide

fmoor:
see also https://github.com/geldata/gel-pkg/issues/145

### DIFF
--- a/docs/resources/guides/deployment/bare_metal.rst
+++ b/docs/resources/guides/deployment/bare_metal.rst
@@ -61,6 +61,13 @@ Install the Gel package.
 
    $ sudo yum install gel-6
 
+Disable SELinux.
+
+.. code-block:: bash
+
+   $ sed -i 's/SELINUX=enforcing/SELINUX=disabled/' /etc/selinux/config
+   $ reboot
+
 
 .. _ref_guide_deployment_bare_metal_enable_unit:
 
@@ -86,7 +93,7 @@ default. You can start the server by enabling the unit.
    $ sudo systemctl enable --now gel-server-6
 
 This will start the server on port 5656, and the data directory will be
-``/var/lib/gel/1/data``.
+``/var/lib/gel/6/data``.
 
 .. warning::
 


docs/resources/guides/deployment/bare_metal.rst
```diff
@@ -86,7 +93,7 @@ default. You can start the server by enabling the unit.
    $ sudo systemctl enable --now gel-server-6
 
 This will start the server on port 5656, and the data directory will be
-``/var/lib/gel/1/data``.
+``/var/lib/gel/6/data``.
```
fmoor, line 96: For versions 6.0 and 6.1 this is a lie. The directory is `/var/lib/edgedb/6/data` for those versions, but I'm not sure it is worth adding an explanation here. The directory will be correct starting in version 6.2.

docs/resources/guides/deployment/bare_metal.rst
```diff
@@ -61,6 +61,13 @@ Install the Gel package.
 
    $ sudo yum install gel-6
 
+Disable SELinux.
+
+.. code-block:: bash
+
+   $ sed -i 's/SELINUX=enforcing/SELINUX=disabled/' /etc/selinux/config
+   $ reboot
+
```
fmoor, line 70: This will no longer be necessary when https://github.com/geldata/gel-pkg/issues/145 is fixed. If this change is accepted I'll open an issue reminding us to remove this blurb once the gel-pkg issue is fixed.

docs/resources/guides/deployment/bare_metal.rst
scotttrinh: Seems fine to pretend we always intended this and the old behavior was an unintentional defect.

scotttrinh (approved):
I can't speak to the technical accuracy here, but I don't see any blockers from team-devrel 🙏

docs/resources/guides/deployment/bare_metal.rst
fmoor: follow up issue https://github.com/geldata/gel/issues/8440

docs/resources/guides/deployment/bare_metal.rst
fmoor: follow up issue https://github.com/geldata/gel/issues/8440

PR #8428, opened 2025-03-04
Do not duplicate kwarg in WebAuthnFactor creation

scotttrinh:
You cannot provide the same argument twice in a dataclass instantiation like I was doing attempting to overide the existing argument with a new one, but instead we pop the value from the `identity` property off when instantiating the `LocalIdentity` class instance which removes it from the source dictionary.

### DIFF
--- a/edb/server/protocol/auth_ext/webauthn.py
+++ b/edb/server/protocol/auth_ext/webauthn.py
@@ -221,7 +221,7 @@ select factor { ** };""",
         assert len(result_json) == 1
 
         factor_dict = result_json[0]
-        local_identity = data.LocalIdentity(**factor_dict["identity"])
+        local_identity = data.LocalIdentity(**factor_dict.pop("identity"))
         return data.WebAuthnFactor(**factor_dict, identity=local_identity)
 
     async def _get_registration_challenge(


msullivan (approved):

edb/server/protocol/auth_ext/webauthn.py
```diff
@@ -215,13 +218,14 @@ async def register(
...
             else:
+                logger.error(f"WebAuthn registration failed: {exc}")
                 raise exc
 
         result_json = json.loads(result.decode())
         assert len(result_json) == 1
 
         factor_dict = result_json[0]
-        local_identity = data.LocalIdentity(**factor_dict["identity"])
+        local_identity = data.LocalIdentity(**factor_dict.pop("identity"))
```
scotttrinh, line 228: This is the actual fix

edb/server/protocol/auth_ext/http.py
```diff
@@ -1469,11 +1482,18 @@ async def handle_webauthn_register(
...
+            email_factor = await webauthn_client.register(
+                credentials=credentials,
+                email=email,
+                user_handle=user_handle,
+            )
+        except Exception as e:
+            logger.error(f"Failed to register WebAuthn: {e}")
+            raise errors.WebAuthnRegistrationFailed(
+                "Failed to register WebAuthn"
+            ) from e
```
fantix, line 1495: It's a bit weird to me to catch all exceptions (including unexpected errors) *just* for a certain function call in one of the `handle_*()` callbacks: that looks like this particular call to `register()` must not fail arbitrarily, while the rest of the code in this handler can fail. I'd wrap all the `handle_webauthn_*()` functions with a `try...except Exception` (probably as a method decorator), and wrap the arbitrary exception into a `WebAuthnAuthenticationFailed` error. We can also `logger.error(..., exc_info=True)` in the outer handler of `WebAuthnAuthenticationFailed` only once - the traceback will show more details for this unexpected error.

fantix (approved):

edb/server/protocol/auth_ext/http.py
scotttrinh: > It's a bit weird to me to catch all exceptions (including unexpected errors) just for a certain function call in one of the handle_*() callbacks: that looks like this particular call to register() must not fail arbitrarily, while the rest of the code in this handler can fail.

Yeah, that's a valid point, but this is somewhat on purpose: I'm specifically saying that "failing to successfully execute `webauthn_client.register` should raise a `errors.WebAuthnRegistrationFailed` exception, but the other errors (expected and unexpected) ought to bubble as necessary. There are in fact other parts of this handler (like the missing `user_handle` exception on line 1480) that already have exception handling, and I don't want to map them all into the same `WebAuthnRegistrationFailed` exception since it wasn't a failure to register the WebAuthn credentials.

> We can also logger.error(..., exc_info=True) in the outer handler of WebAuthnAuthenticationFailed only once - the traceback will show more details for this unexpected error.

Oh, that's an interesting idea! I wonder if _all_ of these "expected" errors (these _are_ actually expected errors rather than some kind of fatal defect) ought to give some helpful log output to the user. Will fix. How do you feel about `logger.error` vs. `logger.debug` in this case?

scotttrinh: Oh actually, I was wrong, `_fail_with_error` absolutely _does_ already log the error out, I just assumed/remembered  wrong. I'll remove the extra log output here.

edb/server/protocol/auth_ext/http.py
```diff
@@ -2287,7 +2315,7 @@ def _fail_with_error(
         "type": str(ex.__class__.__name__),
     }
 
-    logger.error(f"Failed to handle HTTP request: {err_dct!r}")
+    logger.error(f"Failed to handle HTTP request: {err_dct!r}", exc_info=True)
```
fantix, line 2318: Would this be too noisy? As `_fail_with_error()` is used by many handlers, maybe it's better if we:

```
def _fail_with_error(..., exc_info: bool):
    ....
    logger.error(...., exc_info=exc_info)
```

and call `_fail_with_error(exc_info=True)` only in branches that - we know - are handling unexpected errors, like `except Exception` or arguably `except WebAuthnRegistrationFailed`

scotttrinh: Fair question. One of the goals of 6.x for the auth extension was to make it easier to troubleshoot and debug while developing, so I'd personally rather show more and more helpful errors in more places than continue to be conservative here, but maybe this is too large a change for too surgical of a bug fix to warrant making big moves here. I'm fine with your suggestion of reducing the scope of this change and re-addressing the question of logging in it's own context.

scotttrinh: @fantix ⚡ [`5e02c24` (#8428)](https://github.com/geldata/gel/pull/8428/commits/5e02c2453010fa1cb16e55e8cfd37158dcc5bc2f)

PR #8432, opened 2025-03-04
Default `cors_always_allowed_origins` to localhost for `edb server`

jaclarke:
To make local ui development easier, when running the `edb server` command set the `cors_always_allowed_origins` setting to `http://localhost:*` by default, so I don't have to remember to either set it via the command line arg/env var, or set the `cors_allow_origins` config.


msullivan (approved):
Sure, why not

PR #8455, opened 2025-03-07
Upgrading guide

scotttrinh:


docs/resources/upgrading.rst
```diff
@@ -0,0 +1,173 @@
...
+
+   .. code-block:: bash
+
+       $ gel restore --all my_database.dump/ --dsn <new dsn>
+
+   Once the restore is complete, update your application to connect to the new instance.
+
+   This process will involve some downtime, specifically during steps 2 and 3.
+
+GitHub Action
```
scotttrinh, line 163: I think we're still waiting on https://github.com/geldata/setup-gel/pull/39 before we can actually suggest this.

docs/resources/upgrading.rst
```diff
@@ -0,0 +1,173 @@
+.. _ref_upgrading:
+
+=================
+Upgrading from v5
```
scotttrinh, line 4: Maybe there is a better title for this? Maybe "Upgrading from EdgeDB"?

docs/resources/index.rst
```diff
@@ -6,6 +6,7 @@ Resources
     :maxdepth: 3
     :hidden:
 
+    upgrading
```
scotttrinh, line 9: Couldn't think of a better place for this, but Resources is maybe a bit too hidden?

1st1 (approved):

PR #8473, opened 2025-03-14
Test: add test for enum params in func cache

fantix:
Add test for #8471


tests/test_server_ops.py
```diff
@@ -938,7 +943,7 @@ def measure_sql_compilations(
                 try:
                     # It should hit the cache no problem.
                     with self.assertChange(measure_compilations(sd), 0):
-                        await con.query(qry)
+                        await con.query(qry, 'Two')
```
fantix, line 946: This is verified to fail without the fix in #8471

msullivan (approved):

PR #8474, opened 2025-03-16
Fix DSN env var usage docs

janwirth:
Bug / Misdocumentation
```
❯ GEL_DSN=gel://hostname.com:1234?password_env=MY_PASSWORD
                                                            
❯ echo GEL_DSN
GEL_DSN
                                                            
❯ echo $GEL_DSN
gel://hostname.com:1234?password_env=MY_PASSWORD
```

### DIFF
--- a/docs/reference/reference/dsn.rst
+++ b/docs/reference/reference/dsn.rst
@@ -101,5 +101,5 @@ containing the value (``?host_file=./hostname.txt``).
   .. code-block::
 
     MY_PASSWORD=p@$$w0rd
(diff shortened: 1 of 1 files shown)


mmastrac:
The existing documentation is actually correct: when we parse a DSN, `_env` query parameters instruct the connection string parser to pull that particular value from the environment variable of that name.

No `$` is required as this happens entirely within our connection setup code for each driver.

janwirth:
Ah you are right, my bad.

mmastrac:
No worries, thanks for the report. If we can clarify that documentation, please feel free to suggest changes. It can certainly use more explanation. 😄

PR #8477, opened 2025-03-17
multitenant: improve system API responsiveness

### LINKED ISSUE #8355: make `/server/status/` healthcheck endpoints non-blocking with respect to the compiler pool
current behavior:

calling the the `/server/status/alive` and/or `/server/status/ready` healthcheck endpoints makes an RPC to the compiler pool, and if the system is overly busy, that RPC may wait in a queue before it's serviced, potentially causing the healthcheck to time out

desired behavior:

make the check non-blocking so that it won't time out when the system is under heavy load

sketch of a design:

- keep a timestamp of the most recent successful RPC call made to the compiler (`time.perf_counter_ns` is fine for this, because we don't care about calendar time, just elapsed time)

- in the healthcheck handler, if that timestamp is less than N seconds old (1 <= N <= 60, I'd suggest 10), then treat that as "compiler pool is reachable". if it's more than N seconds old, then follow the current logic of making an RPC to check.

very-nice-to-have:

backportable to 5.x to allow patching existing Cloud instances without requiring an upgrade to 6.x

fantix:
The status/{ready,alive} endpoints now checks the compiler pool status from the server-side first without blocking, if requested from the multitenant host. This reduces the possible-long response time of system APIs when the compilation pool is fully busy.

Also allows HTTP access on multitenat host system API.

Fixes #8355

### DIFF
--- a/edb/server/compiler_pool/pool.py
+++ b/edb/server/compiler_pool/pool.py
@@ -185,6 +185,7 @@ class AbstractPool:
         self._refl_schema = kwargs["refl_schema"]
         self._schema_class_layout = kwargs["schema_class_layout"]
         self._dbindex = kwargs.get("dbindex")
+        self._last_active_time = 0
 
     def _get_init_args(self):
         assert self._dbindex is not None
@@ -618,6 +619,16 @@ class AbstractPool:
     def get_size_hint(self) -> int:
         raise NotImplementedError
 
+    def _maybe_update_last_active_time(self):
+        if sys.exc_info()[0] is None:
+            self._last_active_time = time.monotonic()
+
+    async def health_check(self) -> bool:
+        if time.monotonic() - self._last_active_time > 10:
+            await self.make_compilation_config_serializer()
+            self._maybe_update_last_active_time()
+        return True
+
 
 class BaseLocalPool(
     AbstractPool, amsg.ServerProtocol, asyncio.SubprocessProtocol
@@ -810,12 +821,22 @@ class BaseLocalPool(
         # Skip disconnected workers
         if worker.get_pid() in self._workers:
             self._workers_queue.release(worker, put_in_front=put_in_front)
+        self._maybe_update_last_active_time()
 
     def get_debug_info(self):
         return dict(
             worker_pids=list(self._workers.keys()),
             template_pid=self.get_template_pid(),
         )
+
+    async def health_check(self) -> bool:
+        if not (
+            self._running
+            and self._ready_evt.is_set()
+            and len(self._workers) > 0
+        ):
+            return False
+        return await super().health_check()
 
 
 @srvargs.CompilerPoolMode.Fixed.assign_implementation
@@ -1173,6 +1194,7 @@ class RemotePool(AbstractPool):
         if self._sync_lock.locked():
             self._sync_lock.release()
         self._semaphore.release()
+        self._maybe_update_last_active_time()
 
     async def compile_in_tx(
         self,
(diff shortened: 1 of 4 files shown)


edb/server/compiler_pool/pool.py
```diff
@@ -618,6 +619,16 @@ def get_debug_info(self):
     def get_size_hint(self) -> int:
         raise NotImplementedError
 
+    def _maybe_update_last_active_time(self):
+        if sys.exc_info()[0] is None:
+            self._last_active_time = time.monotonic()
+
+    async def health_check(self) -> bool:
+        if time.monotonic() - self._last_active_time > 10:
+            await self.make_compilation_config_serializer()
```
zackelan, line 628: possibly in a followup PR, doesn't need to be this one:

can we enforce a timeout on this (configurable through an env var would be nice, but hardcoded to 10 seconds or something would be fine too) so that if we time out this connection, we return a healthcheck failure message instead of letting upstream (Consul or whatever else is doing the healthchecks) hit its own request timeout

edb/server/protocol/protocol.pyx
```diff
@@ -749,7 +749,7 @@ cdef class HttpProtocol:
                 path_parts[1:],
                 self.server,
                 self.tenant,
-                self.is_tenant_host,
+                is_tenant_host=not self.is_tls or self.is_tenant_host,
```
zackelan, line 752: I don't understand what this change is / why we're making it?

fantix: > Also allows HTTP access on multitenat host system API.

This is mostly for the test who accesses the tenant-host system APIs (`/status/ready` and `/status/alive`) through cleartext HTTP, but also opens up access from any unencrypted clients. Without this change, such clients will get `No such tenant configured` from `/status/ready` or `/status/alive`.

zackelan: can we make this only work for the test then? allowing anyone to access metrics from any tenant without authentication is a minor security hole, but there's no reason to add it unnecessarily

edb/server/compiler_pool/pool.py
```diff
@@ -618,6 +625,24 @@ def get_debug_info(self):
...
+        if elapsed > HEALTH_CHECK_MIN_INTERVAL:
+            task = self._loop.create_task(
+                self.make_compilation_config_serializer()
+            )
+            task.add_done_callback(
+                lambda _: self._maybe_update_last_active_time()
+            )
+            await asyncio.wait([task], timeout=HEALTH_CHECK_TIMEOUT)
+            if not task.done():
+                return False
```
zackelan, line 643: [Unlike wait_for(), wait() does not cancel the futures when a timeout occurs.](https://docs.python.org/3/library/asyncio-task.html#waiting-primitives)

so we should cancel them on timeout to avoid leaking that memory and clogging up the event loop

elprans, line 643:
```suggestion
            try:
                with asyncio.timeout(HEALTH_CHECK_TIMEOUT):
                    await self.make_compilation_config_serializer()
            except TimeoutError:
                # log timeout / increment metric
                return False
            except Exception:
                return False
            else:
                self._maybe_update_last_active_time()
                return True
```

elprans, line 643:
```suggestion
            try:
                with asyncio.timeout(HEALTH_CHECK_TIMEOUT):
                    await self.make_compilation_config_serializer()
            except TimeoutError:
                # log timeout / increment metric
                return False
            except Exception:
                return False
            else:
                self._maybe_update_last_active_time()
```

fantix: It was intentional (to avoid hairy asyncio cancellation, should've added a gotcha), but okay, will fix

> to avoid leaking that memory and clogging up the event loop

to avoid adding more burden to the already overwhelmed compiler processes. FYI we don't yet have a way to cancel an RPC that is already sent to a compiler process, the cancellation here is only meaningful while the RPC is in the queue, which is ~90% of the case I think.

PR #8485, opened 2025-03-18
Fixed computed ptr with group by incorrectly inferring multiplicity.

### LINKED ISSUE #8481: Regression: Using `for` with `group` returning a free object is raising a QueryError
- EdgeDB Version: 6.3

Noticeable from the built-in UI when trying to use the Data Explorer on an object type with links. We have a query that shows the count of the linked objects grouped by their type. This query used to work in 6.2, but no longer in 6.3:

```
QueryError: possibly not a distinct set returned by an expression for a computed link '__count_questions_3'
    |
 10 | __count_questions_3 := (for g in (
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 11 |               group .questions
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 12 |               using typename := .__type__.name
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 13 |               by typename
    | ^^^^^^^^^^^^^^^^^^^^^^^^^
 14 |             ) union {
    | ^^^^^^^^^^^^^^^^^^^^^
 15 |               typename := g.key.typename,
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 16 |               count := <std::float64>count(g.elements)
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 17 |             })
    | ^^^^^^^^^^^^^
Hint: You can use assert_distinct() around the expression to turn this into a runtime assertion, or the DISTINCT operator to silently discard duplicate elements.
```

dnwpark:
Fixes query errors caused by using group by in a computed pointer.

For example, given the schema:
```
module default {
    type Foo { n: int64 };
    type Bar { multi foo: Foo };
}
```

The query `SELECT Bar { g := (group .foo by .n) };` would produce the error:
```
QueryError: possibly not a distinct set returned by an expression for a computed link 'g'
  ┌─ <query>:1:14
  │
1 │ SELECT Bar { g := (group .foo by .n) };
  │              ^^^^^^^^^^^^^^^^^^^^^^ You can use assert_distinct() around the expression to turn this into a runtime assertion, or the DISTINCT operator to silently discard duplicate elements.
```

close #8481

### DIFF
--- a/edb/edgeql/compiler/inference/multiplicity.py
+++ b/edb/edgeql/compiler/inference/multiplicity.py
@@ -815,9 +815,11 @@ def __infer_group_stmt(
         if set:
             infer_multiplicity(set, scope_tree=scope_tree, ctx=ctx)
 
-    # N.B: Since the type is usually a free object (except in some
-    # internal tests), this will typically get turned UNIQUE by the
-    # enclosing Set.
+    # N.B: The type is usually a free object (except in some
+    # internal tests), which are always unique
+    if irtyputils.is_free_object(ir.typeref):
+        return UNIQUE
+
     return DUPLICATE
 
 
--- a/tests/test_edgeql_group.py
+++ b/tests/test_edgeql_group.py
@@ -3853,6 +3853,120 @@ class TestEdgeQLGroup(tb.QueryTestCase):
             ],
         )
 
+    async def test_edgeql_group_link_property_02(self):
+        await self.assert_query_result(
+            r'''
+            WITH MODULE cards
+            select User { cards := (group .deck { name } by .element) };
+            ''',
+            tb.bag([
+                {
+                    'cards': tb.bag([
+                        {
+                            'key': {'element': 'Water'},
+                            'grouping': ['element'],
+                            'elements': tb.bag([
+                                {'name': 'Bog monster'},
+                                {'name': 'Giant turtle'}
+                            ])
+                        },
+                        {
+                            'key': {'element': 'Fire'},
+                            'grouping': ['element'],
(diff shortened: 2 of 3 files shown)


aljazerzen (approved):

edb/edgeql/compiler/inference/multiplicity.py
```diff
@@ -815,9 +815,11 @@ def __infer_group_stmt(
         if set:
             infer_multiplicity(set, scope_tree=scope_tree, ctx=ctx)
 
-    # N.B: Since the type is usually a free object (except in some
-    # internal tests), this will typically get turned UNIQUE by the
-    # enclosing Set.
+    # N.B: The type is usually a free object (except in some
+    # internal tests), which are always unique
+    if irtyputils.is_free_object(ir.typeref):
```
scotttrinh, line 820: I don't think I understand enough about what side effects the above `infer_multiplicity` calls have, but is it possible to return early with this call?

scotttrinh (approved):
Assuming manual test of the UI has shown that this change is enough to fix the linked bug issue, lgtm!

edb/edgeql/compiler/inference/multiplicity.py
dnwpark: It recurses through the entire tree to cache the multiplicity for different sub-branches later on, so unfortunately no.

PR #8487, opened 2025-03-19
Add ui tests

jaclarke:
Runs the ui tests alongside the normal test suite to hopefully catch possible future regressions that break the ui.

### DIFF
--- a/.github/workflows.src/tests.tpl.yml
+++ b/.github/workflows.src/tests.tpl.yml
@@ -135,42 +135,42 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      <<- restore_cache() >>
+    <<- restore_cache() >>
 
-      # Setup UI test dependencies
+    # Setup UI test dependencies
 
-      - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


scotttrinh:
I think you want to make this change in https://github.com/geldata/gel/blob/master/.github/workflows.src/tests.tpl.yml and then run the proper make command to generate the the workflow yaml.

.github/workflows.src/tests.tpl.yml
```diff
@@ -130,8 +130,50 @@ jobs:
...
+        path: build/gel-ui
+        fetch-depth: 2
+
+    - name: Set up Node
+      uses: actions/setup-node@v3
+      with:
+        node-version: lts/*
+
+    - name: Set up chrome
+      uses: browser-actions/setup-chrome@v1
```
elprans, line 154: Please use commit hashes instead of tags on *all* third-party actions (i.e everything that is not in `actions/` or `geldata/`).

msullivan (approved):
Sure, let's try it

PR #8489, opened 2025-03-19
Pin CLI version

fantix:
geldata/gel-cli#1552 requires a proper fix in the server repo, so pin the CLI to the version before that to unbreak master for now.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,7 @@ import setuptools_rust
 
 EDGEDBCLI_REPO = 'https://github.com/edgedb/edgedb-cli'
 # This can be a branch, tag, or commit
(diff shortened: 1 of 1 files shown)


elprans (approved):
FWIW, I think we should just get rid of the CLI build in server altogether.  `gel-python` installs a `gel` entrypoint and downloads a pre-built binary, that's ought to be enough.

fantix:
> FWIW, I think we should just get rid of the CLI build in server altogether. `gel-python` installs a `gel` entrypoint and downloads a pre-built binary, that's ought to be enough.

Sounds good, I'll create a PR.

PR #8491, opened 2025-03-19
CI: don't build CLI in CI

fantix:


elprans (commented):
You should also get rid of the entrypoints in `pyproject.toml` and update `edb/cli` to just run `gel` from `PATH`.

elprans (approved):

PR #8501, opened 2025-03-24
Fix SQL introspection on old PostgreSQL versions

aljazerzen:
Fixes backward compatiblity issues with postgres <17 >=14 introduced in #8488


fantix (approved):
Thanks!

PR #8503, opened 2025-03-24
Add docs redirects

jaclarke:
See https://github.com/geldata/edgedb.com/pull/998

@scotttrinh Feel free to overwrite this with the actual redirects.

### DIFF
--- a/docs/redirects.js
+++ b/docs/redirects.js
@@ -1,9 +1,28 @@
 // See https://nextjs.org/docs/app/api-reference/config/next-config-js/redirects
 module.exports = [
   {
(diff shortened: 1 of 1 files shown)


scotttrinh:
Amazing. I'll copy what we have in dotcom already into a separate PR on this one, and we can get it merged before switching the dotcom one to reading from this one. 🙏

scotttrinh (approved):

scotttrinh:
There is no harm in landing this well before the dotcom side is merged, so let's merge it now. Worst case scenario, we'll have to update both places if we need an emergency redirect.

PR #8507, opened 2025-03-25
Prevent server auto shutdown during building AI index

fantix:
The keepalive token is only added when there is actually AI indexing work to do, in order to avoid delaying the auto shutdown indefinitely. The token is only removed when the "AI lock" is released.


dnwpark (approved):
Looks good!

PR #8509, opened 2025-03-26
Gel-ls: read gel.toml and schema-dir

aljazerzen:


edb/language_server/server.py
```diff
@@ -66,6 +76,36 @@ class DiagnosticsSet:
...
+        self,
+        doc: pygls.workspace.TextDocument,
+        diagnostics: Iterable[lsp_types.Diagnostic],
+    ):
+        if doc not in self.by_doc:
+            self.by_doc[doc] = []
+        self.by_doc[doc].extend(diagnostics)
+
+
+def _new_diagnostic_at_the_top_at_the_top(message: str) -> lsp_types.Diagnostic:
```
scotttrinh, line 98: Does this intentionally stutter? `at_the_top_at_the_top`?

edb/language_server/project.py
```diff
@@ -0,0 +1,72 @@
+from typing import Optional, cast, Any, Tuple
```
elprans, line 1: `typing.Tuple` is deprecated, use `tuple`

edb/language_server/server.py
aljazerzen: Lol, no it does not - it's a new python language server, which is much worse than pylance.

PR #8513, opened 2025-03-27
Gel-ls: go to definition for .edgeql

aljazerzen:
Language server side of "get location of definition of symbol at given source location".
The things left are:
- tweak how spans are inferred in IR: in `select Hello`, we will infer the span of `Hello` to pointer `Hello::id`, which is then what gets picked. And id should not have any span, since it was not in the source code at all.
- actually save `sourcectx` in schema objects. Most objects I inspected do not have it, I don't know why. I use `s_ddl.apply_sdl` with nodes that do have spans, but out I get schema which does not have sourcectx.

### DIFF
--- a/edb/common/span.py
+++ b/edb/common/span.py
@@ -302,6 +302,9 @@ class SpanValidator(ast.NodeVisitor):
         super().generic_visit(node)
 
 
+# Finds the node in AST by position within the source.
+# It returns the first node whose span contains the target offset in a
+# post-order traversal of the AST.
 def find_by_source_position(
     node: ast.AST, target_offset: int
 ) -> ast.AST | None:
--- a/edb/common/span.py
+++ b/edb/common/span.py
@@ -33,7 +33,7 @@ contexts through the AST structure.
 
 from __future__ import annotations
 
-from typing import Iterable, Optional
+from typing import Iterable, Any, Optional
 import re
 import bisect
 
@@ -300,3 +300,36 @@ class SpanValidator(ast.NodeVisitor):
         if getattr(node, 'span', None) is None:
             raise RuntimeError('node {} has no span'.format(node))
         super().generic_visit(node)
+
+
+def find_by_source_position(
+    node: ast.AST, target_offset: int
+) -> ast.AST | None:
+    finder = SpanFinder(target_offset)
+    finder.visit(node)
+    return finder.found_node
+
+
+class SpanFinder(ast.NodeVisitor):
+    target_offset: int
+    found_node: Any | None
+
+    def __init__(self, target_offset: int):
+        super().__init__()
+        self.target_offset = target_offset
+        self.found_node = None
+
+    def generic_visit(self, node, *, combine_results=None) -> Any:
+        has_span = False
+        if node_span := getattr(node, 'span', None):
+            has_span = True
+            if not span_contains(node_span, self.target_offset):
+                return
+
+        super().generic_visit(node)
+        if self.found_node is None and has_span:
+            self.found_node = node
+
+
+def span_contains(span: Span, target_offset: int) -> bool:
+    return span.start <= target_offset and target_offset <= span.end
--- a/edb/language_server/main.py
+++ b/edb/language_server/main.py
@@ -177,22 +177,10 @@ def document_definition(
 
             if isinstance(ql_ast, list):
 
-                # <DEBUG>
-                from edb.common import markup
-                with open('ql_stmt.txt', 'w') as file:
-                    markup.dump(ql_ast, file=file)
-                # </DEBUG>
-
                 _, ir_stmts = ls_server.compile(ls, document, ql_ast)
 
                 ir_stmt: irast.Statement
                 for ir_stmt in ir_stmts:
-
-                    # <DEBUG>
-                    from edb.common import markup
-                    with open('ir_stmt.txt', 'w') as file:
-                        markup.dump(ir_stmt, file=file)
-                    # </DEBUG>
 
                     node = edb_span.find_by_source_position(ir_stmt, position)
                     if not node:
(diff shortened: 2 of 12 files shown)


aljazerzen:
This PR now contains many commits, half of which will be converted into separate PRs. Rebasing of this PR incoming.

---

Apparently, we did have `Object.sourcectx`, but it was marked as "ephemeral" and was never written to.

I've changed that and made it so that when schema.CreateObjectCommand is being applied, it takes own `span` and writes it to attributes of the created object. Problems:
- this is done for all objects, which might mean significant increase of size of the schema in bytes, since span also includes original source code in `buffer` field,
- if an object is `Alter`-ed, the span will be incorrect. But this does not matter, since gel-ls will always create the schema from SQL directly and there will be no `Alter`s,

---

When an EdgeQL query is compiled to IR, we emitting confusing spans. I've done following changes:

- in `select Player { name }`, the resulting IR was changed:
  ```diff
    Set:
  -   span: 7..13
  +   span: 0..20
      expr: Path:
        span: 7..13
      elements: [ ... ]
  ```

- in `select Player { name }`, I've removed the span of injected `id` pointer
  ```diff
    Set:
      span: 0..20
      elements: [
         Ptr:
           name: id
  -        span: 0..20
  +        span: None
         ...
      ]
  ```

- in `select Player { name }`, I've changed the span of `name` from the span of the whole shape to just the span of `name` pointer
  ```diff
    Set:
      span: 0..20
      elements: [
         Ptr:
           name: id
         Ptr:
           name: name
  -        span: 0..20
  +        span: 16..20
         ...
      ]
  ```

aljazerzen:
The result is this:

vscode-gel-ls-definition

aljazerzen:
Now also supporting:
- "goto pointers" and
- "goto bindings".

vscode-ls-definition-2

edb/common/span.py
```diff
@@ -300,3 +300,36 @@ def generic_visit(self, node):
         if getattr(node, 'span', None) is None:
             raise RuntimeError('node {} has no span'.format(node))
         super().generic_visit(node)
+
+
+def find_by_source_position(
```
msullivan, line 305: Add a comment somewhere around here that explains what we are looking for and sketches how. (The first node to contain the span in a post-order traversal of the AST, right? So, leaves preferred over their parents)

edb/language_server/main.py
```diff
@@ -126,22 +134,140 @@ def document_updated(ls: ls_server.GelLanguageServer, doc_uri: str):
...
+                return None
+            ql_ast = ql_ast_res.ok
+
+            if isinstance(ql_ast, list):
+
+                # <DEBUG>
+                from edb.common import markup
+                with open('ql_stmt.txt', 'w') as file:
+                    markup.dump(ql_ast, file=file)
+                # </DEBUG>
```
msullivan, line 184: We can't ship this without being behind a flag

edb/language_server/main.py
```diff
@@ -126,22 +134,140 @@ def document_updated(ls: ls_server.GelLanguageServer, doc_uri: str):
...
+                _, ir_stmts = ls_server.compile(ls, document, ql_ast)
+
+                ir_stmt: irast.Statement
+                for ir_stmt in ir_stmts:
+
+                    # <DEBUG>
+                    from edb.common import markup
+                    with open('ir_stmt.txt', 'w') as file:
+                        markup.dump(ir_stmt, file=file)
+                    # </DEBUG>
```
msullivan, line 195: More stray debug

edb/language_server/main.py
```diff
@@ -126,22 +134,140 @@ def document_updated(ls: ls_server.GelLanguageServer, doc_uri: str):
...
+                    assert isinstance(node, irast.Base), node
+
+                    ls.show_message_log(f'node: {str(node)}')
+                    ls.show_message_log(f'span: {str(node.span)}')
+
+                    if not isinstance(node, irast.Set):
+                        return None
+                    assert ir_stmt.schema
+
+                    # if this is a ptr, find it in the schema
```
msullivan, line 209: My inclination would be to check if `expr` is a `Pointer`, and use the `ptrref` there instead. Not sure that it really matters, though.

edb/language_server/main.py
```diff
@@ -126,22 +134,140 @@ def document_updated(ls: ls_server.GelLanguageServer, doc_uri: str):
...
+                        ls.show_message_log(f'rptr: {ptr}')
+                        if isinstance(ptr, irast.PointerRef):
+                            target = ir_stmt.schema.get_by_id(ptr.id)
+
+                    # fallback to getting the target type
+                    if not target:
+                        target = ir_stmt.schema.get_by_id(
+                            node.path_id.target.id
+                        )
+                        assert target
```
msullivan, line 221: Does checking if the `expr` is `TypeRoot` and using that work? Maybe that is too narrow?
Always using the path_id's target seems a little too broad, because it means more complex expressions could point to the definition, even when that might not make as much sense?

(Writing `'foo'` would want to look up the definition of `std::str`, using this implementation. Not that that would work anyway.)

edb/language_server/utils.py
```diff
@@ -0,0 +1,19 @@
+from lsprotocol import types as lsp_types
+
+from edb.edgeql import tokenizer
+
+
+def convert_span(source: str, span: tuple[int, int | None]) -> lsp_types.Range:
```
msullivan, line 6: Brief comment

msullivan (approved):
This is great

edb/schema/objects.py
```diff
@@ -1025,7 +1025,9 @@ class Object(s_abc.Object, ObjectContainer, metaclass=ObjectMeta):
         inheritable=False,
     )
 
-    # Span of source text that contained definition of this object
+    # Span of source text that contained definition of this object.
+    # This field is ephemeral, which means it not serilaized and saved
```
msullivan, line 1029: serialized

Issue #6763, opened 2024-02-01
RuntimeError: Could not determine home directory


spacemanspiff2007:
Steps to Reproduce:

1. Spin up Docker image of Postgres DB
2. Spin up Docker image of EdgeDB and set ``EDGEDB_SERVER_BACKEND_DSN``, ``EDGEDB_SERVER_SECURITY: insecure_dev_mode`` and run as Non-Root

Observe Crash

````
    /usr/lib/x86_64-linux-gnu/edgedb-server-4/lib/python3.11/site-packages/edb/server/main.py, line 545, in run_server
        > cluster, args = await _get_remote_pgcluster(args, tenant_id)
    /usr/lib/x86_64-linux-gnu/edgedb-server-4/lib/python3.11/site-packages/edb/server/main.py, line 361, in _get_remote_pgcluster
        > cluster = await pgcluster.get_remote_pg_cluster(
    /usr/lib/x86_64-linux-gnu/edgedb-server-4/lib/python3.11/site-packages/edb/server/pgcluster.py, line 907, in get_remote_pg_cluster
        > addrs, params = pgconnparams.parse_dsn(dsn)
    /usr/lib/x86_64-linux-gnu/edgedb-server-4/lib/python3.11/site-packages/edb/server/pgconnparams.py, line 572, in parse_dsn
        > sslkey = _dot_postgresql_path('postgresql.key')
    /usr/lib/x86_64-linux-gnu/edgedb-server-4/lib/python3.11/site-packages/edb/server/pgconnparams.py, line 271, in _dot_postgresql_path
        > return str((pathlib.Path.home() / '.postgresql' / filename).resolve())
    /usr/lib/x86_64-linux-gnu/edgedb-server-4/lib/python3.11/pathlib.py, line 914, in home
        > return cls("~").expanduser()
    /usr/lib/x86_64-linux-gnu/edgedb-server-4/lib/python3.11/pathlib.py, line 1385, in expanduser
        > raise RuntimeError("Could not determine home directory.")

RuntimeError: Could not determine home directory.
ERROR: Could not complete instance bootstrap
````

This can be fixed by adding ``?sslmode=disable`` to the DSN.

spacemanspiff2007:
Issue still there with the new gel server, however the workaround does not work anymore 😢

mmastrac:
Are you able to set the HOME environment variable before running the server?

spacemanspiff2007:
[...]

PR #8514, opened 2025-03-28
Update pgconnparams.py

spacemanspiff2007:
#6763

### DIFF
--- a/edb/server/pgconnparams.py
+++ b/edb/server/pgconnparams.py
@@ -53,7 +53,7 @@ else:
         except RuntimeError:
             # In some container environments the home dir is not
             # available so we return a path to a nonexistent file
-            return pathlib.Path('/.postgresql')
+            return pathlib.Path('')
 
 
 class SSLMode(enum.IntEnum):
--- a/edb/server/pgconnparams.py
+++ b/edb/server/pgconnparams.py
@@ -51,8 +51,8 @@ else:
         try:
             return pathlib.Path.home() / '.postgresql'
         except RuntimeError:
-            # In some container environments the home dir is available 
-            # so we return a path to a nonexistent file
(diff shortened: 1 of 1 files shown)


edb/server/pgconnparams.py
```diff
@@ -48,7 +48,12 @@ def get_pg_home_directory() -> pathlib.Path:
 else:
 
     def get_pg_home_directory() -> pathlib.Path:
-        return pathlib.Path.home() / '.postgresql'
+        try:
+            return pathlib.Path.home() / '.postgresql'
+        except RuntimeError:
+            # In some container environments the home dir is not
+            # available so we return a path to a nonexistent file
+            return pathlib.Path('/.postgresql')
```
fantix, line 56: This (maybe load keys from `/.postgresql` if it unexpectedly exists) is different than what libpq does upstream (don't load anything if homedir isn't available):

https://github.com/postgres/postgres/blob/2247281c470502c799fc2153a3583c025e894a9c/src/interfaces/libpq/fe-secure-openssl.c#L1149-L1159

Ideally, we should make `home_dir` an optional value here:

https://github.com/geldata/gel/blob/5c9955b4e6197ca677ebaa4d1b3867fad3b1d288/rust/pgrust/src/python/mod.rs#L280

spacemanspiff2007: I changed the code to return an empty path, however I don't know any rust so I would need help to change that part ...

spacemanspiff2007:
@fantix Do you expect me to do rust changes or could you ping someone who could help with that?

fantix:
I'll update your PR

fantix:
> Please sign our Contributor License Agreement.

@spacemanspiff2007

PR #8518, opened 2025-03-28
Refactor python typing with builtins (UP006)

aljazerzen:
So `typing.List` is replaced by just `list`.

I would like to thank ruff for contributing to this PR.


elprans (commented):
Can we also add `UP007`?

elprans (commented):
Oh and `UP046` and `UP047`

aljazerzen:
I'll split this PR into a few separate ones.

1. #8523 
2. this one
3. one for [UP007](https://docs.astral.sh/ruff/rules/non-pep604-annotation-union/)
4. one for [UP046](https://docs.astral.sh/ruff/rules/non-pep695-generic-class/) [UP047](https://docs.astral.sh/ruff/rules/non-pep695-generic-function/)

PR #8523, opened 2025-03-29
Update ruff

aljazerzen:


edb/pgsql/compiler/relgen.py
```diff
@@ -71,26 +72,19 @@
 from . import relctx
 
 
+@dataclasses.dataclass(repr=False, eq=False)
```
msullivan, line 75: This is a good change but why was ruff complaining?

msullivan (approved):

edb/pgsql/compiler/relgen.py
aljazerzen: There is a rule for detecting stuff that could be a dataclass: https://docs.astral.sh/ruff/rules/class-as-data-structure/

PR #8525, opened 2025-03-30
Refactor typing.TypeVar into PEP695 syntax (UP046 & UP047)

aljazerzen:
Converts this:

```py
from typing import TypeVar

T = TypeVar("T")


class GenericClass(Generic[T]):
    var: T
```

... to this:

```py
class GenericClass[T]:
    var: T
```

And similarly for functions.

https://docs.astral.sh/ruff/rules/non-pep695-generic-class/

Since this is quite large, I don't plan on backporting it.


edb/pgsql/parser/ast_builder.py
```diff
@@ -44,7 +44,6 @@ class Context:
 # Node = bool | str | int | float | List[Any] | dict[str, Any]
 Node = Any
 T = TypeVar("T")
-U = TypeVar("U")
 Builder = Callable[[Node, Context], T]
```
dnwpark: Maybe this should be removed too. Will the `T` here be the same as the `T` bound in `_enum[T]` for example?

edb/schema/pointers.py
```diff
@@ -19,12 +19,13 @@
...
-    Optional,
-    TypeVar,
+    cast,
     Iterable,
+    Optional,
+    Self,
     Sequence,
-    cast,
     TYPE_CHECKING,
+    TypeVar,
```
dnwpark: Is `TypeVar` still used here? Probably can remove `Pointer_T`.

edb/schema/delta.py
```diff
@@ -367,24 +368,21 @@ def sort_by_cross_refs_key(
...
-
 class CommandMeta(
     adapter.Adapter,
     struct.MixedStructMeta,
 ):
 
     _astnode_map: dict[type[qlast.DDLOperation], type[Command]] = {}
 
-    def __new__(
+    def __new__[CommandMeta_T: CommandMeta](
```
dnwpark, line 385: This can't be `Self`?

edb/schema/objects.py
```diff
@@ -108,7 +109,6 @@ class DefaultConstructorT(enum.Enum):
 
 
 T = TypeVar("T")
```
dnwpark: Can be removed?

dnwpark (approved):

edb/schema/delta.py
aljazerzen: I don't think so, but I'm not sure I can explain why

PR #8531, opened 2025-04-01
Tweak how we set IR source spans

aljazerzen:
Extracted from #8513.

This is needed for language server, which looks at compiled IR to determine what's the type under the cursor.

For that to work, we need search IR for a given cursor position. For that to work, IR nodes should have the span that contains all spans of its children. This was not the case, so I've made the following changes:

- in `select Player { name }`, the resulting IR was changed:
  ```diff
    Set:
  -   span: 7..13
  +   span: 0..20
      expr: Path:
        span: 7..13
      elements: [ ... ]
  ```

- in `select Player { name }`, I've removed the span of injected `id` pointer
  ```diff
    Set:
      span: 0..20
      elements: [
         Ptr:
           name: id
  -        span: 0..20
  +        span: None
         ...
      ]
  ```

- in `select Player { name }`, I've changed the span of `name` from the span of the whole shape to just the span of `name` pointer
  ```diff
    Set:
      span: 0..20
      elements: [
         Ptr:
           name: id
         Ptr:
           name: name
  -        span: 0..20
  +        span: 16..20
         ...
      ]
  ```

---

On paper, these changes make sense, but our explain code will probably disagree.

### DIFF
--- a/edb/edgeql/compiler/stmt.py
+++ b/edb/edgeql/compiler/stmt.py
@@ -33,6 +33,7 @@ import textwrap
 
 from edb import errors
 from edb.common import ast
+from edb.common import span as edb_span
 from edb.common.typeutils import not_none
 
 from edb.ir import ast as irast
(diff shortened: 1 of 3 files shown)


msullivan (approved):
Seems plausible

PR #8532, opened 2025-04-01
Improve ast span markup

aljazerzen:
This PR makes it so markup of ql and ir ASTs contains a simple `filename:start..end`.

Markup of span is a fully rendered source code, with the span characters underlined and arrows and such. We didn't emit that in AST markup, because it is too large.

### DIFF
--- a/edb/edgeql/ast.py
+++ b/edb/edgeql/ast.py
@@ -25,7 +25,7 @@ from __future__ import annotations
 import typing
 
 from edb.common import enum as s_enum
-from edb.common import ast, span, markup
+from edb.common import ast, span
 
 from . import qltypes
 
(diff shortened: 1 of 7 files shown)


edb/edgeql/ast.py
```diff
@@ -99,6 +99,19 @@ def dump_edgeql(self) -> None:
...
+
+    if not base.span:
+        return node
+
+    node.add_child(label='span', node=markup.serialize(str(base.span), ctx=ctx))
+    child = node.children.pop()
+    node.children.insert(1, child)
+    return node
+
+
```
msullivan, line 114: Always including the spans for the AST seems super noisy

msullivan (approved):

PR #8534, opened 2025-04-01
Remove typing.Optional.

dnwpark:
Fixes a lot of spam when running mypy locally.


msullivan (changes requested):
I think I actually prefer `Optional` to `| None`, but I'm not implacably opposed.

Some work does need to be done to convince me, though

PR #8536, opened 2025-04-01
Clarify link readonly and deletion policies

scotttrinh:
Clarify how readonly and deletion policies interact with links, as this was not explicitly detailed before. Specifying `on target delete allow` will change the set of links to no longer include a link to the deleted object.


docs/reference/datamodel/links.rst
```diff
@@ -855,9 +855,7 @@ The valid SDL sub-declarations are listed below:
     The expression must be :ref:`Stable <ref_reference_volatility>`.
 
 :eql:synopsis:`readonly := {true | false}`
-    If ``true``, the link is considered *read-only*.  Modifications
-    of this link are prohibited once an object is created.  All of the
-    derived links **must** preserve the original *read-only* value.
+    If ``true``, the link is considered *read-only*.  Modifications of this link using ``update`` are prohibited once an object is created. All of the derived links **must** preserve the original *read-only* value. Changes to this link **can** occur if a link is deleted and the appropriate deletion policy allows it.
```
1st1, line 858: I'd hyperlink "derived" to an example/section in the docs where we show link inheritance / overload

1st1 (approved):

docs/reference/datamodel/links.rst
aljazerzen, line 858:
```suggestion
    If ``true``, the link is considered *read-only*.  Modifications of this link using ``update`` are prohibited once an object is created. All of the derived links **must** preserve the original *read-only* value. Changes to this link **might** still occur if the target object is deleted and the deletion policy allows it.
```

aljazerzen (approved):

docs/reference/datamodel/links.rst
scotttrinh: @aljazerzen I agree with everything except the change from "can" to "might still", since if you _have_ deleted a target object with the appropriate deletion policy, it _will_ update the link set. wdyt?

scotttrinh: I guess the language we're using here isn't completely clear: maybe "will" is better than "can" or "might still"?

1st1: Why do we have to use vague "can" / "might" instead of a concrete "will"?

aljazerzen: I wanted to say that 'even on "required", it is not guaranteed that the value will not change'. So the value "might still change".

But "will" also conveys the same thing.

PR #8538, opened 2025-04-01
Cleanup Span name and filename

aljazerzen:
Our `Span` contained following fields: name, filename, document, buffer.

- `document` was never used and never set, so I removed it,
- `name` and `filename` were both used as "the name of the buffer that contains this span", so I consolidated them both to `filename`,
- `buffer` names sense, but I added type annotation that caused mypy to complain about it somewhere else,
- for ad-hoc queries, we were setting filename to `<string>` or `<empty>`, which seemed hacky. Especially, because we were later checking if filename is `<string>` so we would not pass that to clients. So I've removed that and now it is `None` not things that do not come from files.

### DIFF
--- a/edb/common/span.py
+++ b/edb/common/span.py
@@ -33,7 +33,7 @@ contexts through the AST structure.
 
 from __future__ import annotations
 
-from typing import Iterable
+from typing import Iterable, Optional
 import re
 import bisect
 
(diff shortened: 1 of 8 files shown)


msullivan (approved):
Looks good

PR #8540, opened 2025-04-01
Send a LogMessage to the client about query cache recompilation

msullivan:
The CLI could learn to do something useful with it.

### DIFF
--- a/edb/api/errors.txt
+++ b/edb/api/errors.txt
@@ -166,6 +166,8 @@
 
 0x_F0_01_00_00   WarningMessage
 
+0x_F0_02_00_00   MigrationStatusMessage
+
 
 #### Suggested errors for Gel clients
 
--- a/edb/server/dbview/dbview.pyx
+++ b/edb/server/dbview/dbview.pyx
@@ -1215,7 +1215,12 @@ cdef class DatabaseConnectionView:
             self.get_system_config(),
         )
 
-    async def recompile_cached_queries(self, user_schema, schema_version):
+    async def recompile_cached_queries(
+        self,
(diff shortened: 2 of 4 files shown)


edb/api/errors.txt
```diff
@@ -166,6 +166,8 @@
 
 0x_F0_01_00_00   WarningMessage
 
+0x_F0_02_00_00   MigrationStatusMessage
```
elprans, line 169: Let's add a generic `StatusMessage` and make this a subtype of that.

elprans (approved):

edb/server/dbview/dbview.pyx
```diff
@@ -1215,7 +1215,12 @@ cdef class DatabaseConnectionView:
             self.get_system_config(),
         )
 
-    async def recompile_cached_queries(self, user_schema, schema_version):
+    async def recompile_cached_queries(
+        self,
+        user_schema,
+        schema_version,
+        send_log_message: object | None = None,
```
aljazerzen, line 1222: Ideally, this type annotation should say that it expects a function. Unless `.pyx` does not support that.

edb/server/dbview/dbview.pyx
```diff
@@ -1313,6 +1326,7 @@ cdef class DatabaseConnectionView:
         allow_capabilities: uint64_t = <uint64_t>compiler.Capability.ALL,
         pgcon: pgcon.PGConnection | None = None,
         tag: str | None = None,
+        send_log_message: object | None = None,
```
aljazerzen, line 1329: ditto

PR #8542, opened 2025-04-01
Complete SQL ast parsing for ON CONFLICT

aljazerzen:
Fill out the mapping from dicts created by libpg_query to our pg AST
for ON CONFLICT nodes.

They are quite complicated, especially the "constraint target", because
to specify an index, you can list the index columns, with sorting modifiers.

I need this for implementing ON CONFLICT in SQL adapter, but mostly just so
I can throw "Unsupported" exceptions :D.

Most of the changes are from:
- change of type of index_elems from `list[Expr]` to `list[IndexElem]`,
  which then requires a fix at all locations  where we construct the node,
- ~accidental black run on the ast_builder.py~.

### DIFF
--- a/edb/pgsql/codegen.py
+++ b/edb/pgsql/codegen.py
@@ -471,17 +471,23 @@ class SQLSourceGenerator(codegen.SourceGenerator):
             self.new_lines = 1
             self.write('ON CONFLICT')
 
-            if node.on_conflict.infer:
-                self.visit(node.on_conflict.infer)
+            if node.on_conflict.target:
+                self.visit(node.on_conflict.target)
 
             self.write(' DO ')
-            self.write(node.on_conflict.action.upper())
+            self.write(node.on_conflict.action[3:])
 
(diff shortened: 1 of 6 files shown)


edb/pgsql/codegen.py
```diff
@@ -471,17 +471,23 @@ def visit_InsertStmt(self, node: pgast.InsertStmt) -> None:
...
             self.write('ON CONFLICT')
 
-            if node.on_conflict.infer:
-                self.visit(node.on_conflict.infer)
+            if node.on_conflict.target:
+                self.visit(node.on_conflict.target)
 
             self.write(' DO ')
-            self.write(node.on_conflict.action.upper())
+            self.write(node.on_conflict.action[3:])
```
msullivan, line 478: Hm...

msullivan: s/3/len('DO_')/ ?

msullivan (approved):

PR #8543, opened 2025-04-01
Update bundled PostgreSQL to version 17.4

elprans:
(also includes cherry-picked fix for macOS 15.4)


aljazerzen (approved):

msullivan:
Are we going to need to backport?

elprans:
> Are we going to need to backport?

Sure.  Postgres we ship in prod builds is configured in gel-pkg, so the backport is mostly for consistency and to be able to build 6.x in dev mode.

PR #8511, opened 2025-03-26
pg_ext: fix `to_regclass()`, `::regclass` and `pg_settings`

fantix:
Fixes #8508
Refs #8488

- [x] Implement frontend-aware `pg_settings`
- [x] Implement `split_identifier_string()`
- [x] Implement `current_schemas()` function in the backend
- [x] Fix SQL implementation of `to_regclass()`
- [x] Fix regclass typecasting
- [x] Drop static evaluation of replaced functions
- [x] Add test if not yet

PR #8547, opened 2025-04-02
pg_ext: apply sql_state immediately after each SET

fantix:
During adding tests for #8511, I noticed that #8511 only synced the sql_state into the temp table `_edgecon_state` once at fresh pgcon acquisition, but SET commands - especially within transactions - are not changing the `_edgecon_state` table immediately. This defeats the attempt to have frontend-only SQL state in the backend database. More over, `SET LOCAL` commands was completely untouched in #8511, those wasn't working correctly either.

This PR injects PGMessages before executing the frontend-only SET commands to actually modify the `_edgecon_state` table, so that the SQL state in the backend is always up to date. For transaction-level session settings a.k.a. SET LOCAL, a deferred constraint trigger is added to automatically "reset" (remove) the local session settings at the end of each transaction.

This PR also fixes the `pg_settings` view to properly reflect frontend-only settings and their sources, as well as the emulated default values (stored in instdata).

This PR adds tests for #8511.

At last, as a verification of all the changes here, this PR also replaced the old `get_var` frontend implementation of the `SHOW` command with proper SQL backend implementation. That also fixed the bad description of `SHOW ALL` statement, as well as the `pg_show_all_settings()` function.

This merges back to #8511


edb/server/bootstrap.py
```diff
@@ -2258,6 +2259,18 @@ async def _populate_misc_instance_data(
...
+    await _store_static_json_cache(
+        ctx,
+        'sql_default_fe_settings',
+        json.dumps(
+            [
+                {"name": key, "value": pg_common.setting_to_sql(key, val)}
+                for key, val in dbstate.DEFAULT_SQL_FE_SETTINGS.items()
+            ]
+        )
+    )
```
fantix, line 2272: Do we need to do anything for inplace upgrade?

edb/server/compiler/sql.py
```diff
@@ -293,13 +293,95 @@ def _compile_sql(
                 from edb.pgsql import resolver as pg_resolver
                 pg_resolver.dispatch._raise_unsupported(stmt)
 
-            unit.get_var = stmt.name
-            unit.frontend_only = (
-                stmt.name in FE_SETTINGS_MUTABLE
-                or stmt.name.startswith('global ')
+            pg_settings_for_show = pgast.RelRangeVar(
```
aljazerzen, line 296: Can this all go in a sub function? It is getting pretty large.

aljazerzen (approved):
Looks good

edb/server/bootstrap.py
aljazerzen: @msullivan

edb/server/compiler/sql.py
fantix: Yeah, that sounds like a good idea, let me fix it

edb/pgsql/common.py
```diff
@@ -609,3 +610,180 @@ def get_sql_value_function_op(op: pgast.SQLValueFunctionOP) -> str:
...
+def setting_to_sql(name, setting):
+    is_enum = name.lower() in ENUM_SETTINGS
+
+    assert typeutils.is_container(setting)
+    return ', '.join(setting_val_to_sql(v, is_enum) for v in setting)
+
+
+def setting_val_to_sql(val: str | int | float, is_enum: bool):
+    if isinstance(val, str):
+        if is_enum:
```
elprans, line 780: I'm not sure why this is needed? Quoted enum config vars seem to work fine?

fantix: @aljazerzen

aljazerzen: https://github.com/geldata/gel/pull/7745

aljazerzen: I *think* the original problem was that I was trying to set a setting that is expected to be an int, but our machinery was converting it to string.

So I refactored the machinery so that python always passes around `string | int | float` and then correctly serializes to sql.

elprans: Well, the `to_sql` is a misnomer here then, because those values don't actually end up in SQL and are exposed at protocol level.

edb/server/bootstrap.py
msullivan: Oh I missed this question; I'll follow up.

msullivan: Nothing needs to be done for inplace upgrade. instdata is versioned and so gets all the new stuff without intervention.
(Though maybe something should have been done for 6.x??)

PR #8548, opened 2025-04-02
Create an auth webhooks reference section

scotttrinh:
I went to start adding more detail to the webhook section and realized that there is a lot to say about webhooks, so I made a separate section, and linked to it.

### DIFF
--- a/docs/reference/auth/webhooks.rst
+++ b/docs/reference/auth/webhooks.rst
@@ -1,0 +1,256 @@
+.. _ref_auth_webhooks:
+
+========
+Webhooks
+========
+
+The auth extension supports sending webhooks for a variety of auth events. You can use these webhooks to, for instance, send a fully customized email for email verification, or password reset instead of our built-in email verification and password reset emails. You could also use them to trigger analytics events, start an email drip campaign, create an audit log, or trigger other side effects in your application.
+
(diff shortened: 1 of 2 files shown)


docs/reference/auth/webhooks.rst
```diff
@@ -0,0 +1,256 @@
...
+          ext::auth::WebhookEvent.EmailVerificationRequested,
+          ext::auth::WebhookEvent.PasswordResetRequested,
+        },
+        # Optional, only needed if you want to verify the webhook request
+        signing_secret_key := '1234567890',
+      };
+
+.. note::
+
+  URLs must be unique across all webhooks configured for each branch. If you want to send multiple events to the same URL, you can do so by adding multiple ``ext::auth::WebhookEvent`` values to the ``events`` set.
```
beerose, line 35: Maybe we can add "as demonstrated in the example above"?

beerose (approved):

docs/reference/auth/webhooks.rst
scotttrinh: Oh good call. I think maybe I'll just move this whole note into the paragraph that precedes the example so it's clear what to look out for when reading the example.

scotttrinh: @beerose ⚡ [`f0134cf` (#8548)](https://github.com/geldata/gel/pull/8548/commits/f0134cf26e0c171285437f48573894cc419a3e7c)

PR #8551, opened 2025-04-03
Update AI HTTP reference docs

scotttrinh:
1. Embeddings request body: only document the `inputs` parameter
2. RAG request body: new optional `tools` parameter
3. RAG JSON response: completely new data layout
4. RAG streaming response: new event types

### DIFF
--- a/docs/reference/ai/http.rst
+++ b/docs/reference/ai/http.rst
@@ -29,13 +29,11 @@ Request headers
 Request body
 ------------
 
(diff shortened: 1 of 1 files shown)


1st1 (approved):

anbuzin (approved):
I'm going to have to build something using the HTTP API to actually find out if it is complete enough. But just from reading it seems great 👍

PR #8553, opened 2025-04-03
Fix AI ext openai streaming

jaclarke:
Should fix issue reported on discord: https://discord.com/channels/841451783728529451/1356441709856292914/1356625059585986722

When translating the openai response (https://platform.openai.com/docs/api-reference/chat-streaming/streaming) to the anthropic style streaming response (https://docs.anthropic.com/en/api/messages-streaming#event-types), we're not sending a `content_block_start` event before the `content_block_delta` events for the text response as required. Also the default prompt doesn't contain a user message, so the user's query never actually gets sent to the llm, just the system message with the context.


jaclarke:
Question for @diksipav or @anbuzin:
https://github.com/geldata/gel/blob/c001579b09b470779c53ed3ebc8c23df52aed645/edb/server/protocol/ai_ext.py#L1430
Do these tool call indexes start at 0 or 1? If it's 0, then I think this can conflict with the text content block which we also give index 0 right? I'm not sure how to test this to check though.

anbuzin:
> Do these tool call indexes start at 0 or 1?

I'll investigate and get back to you as soon as I can

msullivan:
Bump the `EDGEDB_CATALOG_VERSION`, since the library is changed.

Do you want to backport this?

anbuzin:
@jaclarke they start from 0: [link](https://platform.openai.com/docs/guides/function-calling#streaming)

jaclarke:
> @jaclarke they start from 0: [link](https://platform.openai.com/docs/guides/function-calling#streaming). Could you point me somewhere close to where the conflict is happening, please? I'm not familiar with this code just yet.

Thanks. I think this should fix it? https://github.com/geldata/gel/pull/8553/commits/cf73148dbda26e6fa18908b5d59bde42ff8b1a88
Where I think the conflict can happen, is here: https://github.com/geldata/gel/pull/8553/files#diff-3b86041f767ff6a0e871eda744b21d72d1a0d60f173b80c7c7f127de04fc76acR1426 we now send a `content_block_start` event (and subsequent `content_block_delta`s) for the text part of the response with index 0, and then also send `content_block_start` events for the tool calls using the index from openai: https://github.com/geldata/gel/pull/8553/files#diff-3b86041f767ff6a0e871eda744b21d72d1a0d60f173b80c7c7f127de04fc76acL1430

jaclarke:
> Do you want to backport this?

I guess we should as it currently seems to be completely broken.

msullivan:
> > Do you want to backport this?
> 
> I guess we should as it currently seems to be completely broken.

Roger; label with to-backport-6.x

anbuzin (approved):

diksipav:
@jaclarke let me know if you still more help, I can take a better look into all this.

Also, I added [example projects](https://github.com/geldata/gel-examples/tree/add-nextjs-ai-example/nextjs-edgedb-ai/app) a while ago for both our AI plain usage and the provider (but it never got merged). AI examples provide tests for streaming, non-streaming, completion and chat for all 3 LLMs. May be of use to you and users. They were all working perfectly when I wrote them.

jaclarke:
Thanks @diksipav. I've tested these changes with that ai example project and everything seems to work including tools calls (after updating the packages to gel), so it looks like it's ok.

PR #8555, opened 2025-04-03
Add docs for the JS `withWarningHandler` method

scotttrinh:


docs/reference/using/js/client.rst
```diff
@@ -743,6 +743,33 @@ Client Reference
               // ...
             });
 
+    .. js:method:: withWarningHandler(handler: (warnings: errors.GelError[]) => void): Client
+
+        Returns a clone of the ``Client`` instance with the specified warning handler. Some queries may generate warnings while still returning a result. The ``handler`` function will be called with an array of ``GelError`` objects whenever the client encounters warnings during query execution.
```
1st1, line 748: > Returns a clone

Do we have a dedicated section where we explain our "cloning"? IMO every time we mention cloning it should be linked to that page.

Also I'm not sure I like "cloning" 100% -- "deriving" might be a better word (but also it's a more complex word...) Thoughts?

scotttrinh: I went with the existing language from the rest of these methods, but it would be helpful to explain this more thoroughly and use consistent language, agreed.

jaclarke (approved):
Not related to the docs, but I wonder if maybe the warning handler should also be passed the query as well? Currently I think the only way to correlate the warnings with the query that produced them is to set a new warnings handler for each query.

scotttrinh:
> should also be passed the query as well? 

Oh, that's a good call. Somewhat related here: the current warning does not include enough detail to even make the warning actionable, and once we get that `query` and `attrs` stuff, that'll be even more helpful than providing the callback with the full query text. To be clear, I think we should do both.

docs/reference/using/js/client.rst
scotttrinh: I'm going to keep what we have here and follow up if/when we decide to clarify (or change!) this behavior elsewhere in the docs.

PR #8558, opened 2025-04-04
Expand LS testing framework and cases

aljazerzen:
Language server test previously included only a basic test
that started the server and compared output to
expected string of data.

Now, we can startup a LspRunner, give it files in the workspace,
send it init messages that bring up the server, send requests
and recieve responses.

Tests for:
- basic init,
- syntax err,
- name err,
- get definition.

### DIFF
--- a/edb/errors/base.py
+++ b/edb/errors/base.py
@@ -28,9 +28,7 @@ import contextlib
 
 
 __all__ = (
-    "EdgeDBError",
-    "EdgeDBMessage",
-    "ensure_span",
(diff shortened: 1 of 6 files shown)


edb/errors/base.py
```diff
@@ -42,7 +44,7 @@ def __new__(mcls, name, bases, dct):
         assert name not in mcls._name_map
         mcls._name_map[name] = cls
 
-        code = dct.get('_code')
+        code = dct.get("_code")
```
msullivan, line 47: Boo

msullivan (approved):
lg

Issue #8259, opened 2025-01-25
Support ON CONFLICT in SQL inserts


msullivan:
Hopefully we can compile it into edgeql UNLESS CONFLICT.

It might take some thought which I haven't done yet.

---

TODO:

- [x] parsing full ON CONFLICT syntax,
- [x] specifying constraints with column names,
- [ ] ~specifying constraints with their name~,
- [x] ON CONFLICT DO NOTHING
- [x] ON CONFLICT DO UPDATE
- [x] references to `excluded` table,
- [x] references to the subject table,
- [x] ON CONFLICT DO UPDATE WHERE,
- [x] CommandComplete tag & RETURNING,
- [x] more tests (multiple column update)

msullivan:
We discussed this some today, and the trickiest piece we identified is handling ON CONFLICT UPDATE.

The basic idea is to turn it into something like `INSERT Foo {CTE ANCHORS 1} UNLESS CONFLICT ON (...) ELSE (UPDATE Foo SET {CTE ANCHORS 2})`

The tricky bit is that we will need to stick the SQL CTE for the update (`CTE ANCHORS 2`) in the *middle* of the SQL generated by compiling the `INSERT`, and it will need to select from the CTE that contains all of the conflicting objects. This is the `else_select_cte`: https://github.com/geldata/gel/blob/4267072b03988c8fcf52ba3b3a8ee85279d33504/edb/pgsql/compiler/dml.py#L1460-L1487

We might need to stick some extra metadata into the `else_select_cte` so that we can pick it out during our postprocessing.

PR #8560, opened 2025-04-04
ON CONFLICT basic impl over SQL adapter

aljazerzen:
First stab at #8259
I recommend review commit by commit, since there is
a few refactoring commits appended.

Adds support for INSERT ON CONFLICT DO ...
Currently, there is quite a few limitations:

- Constraint can only be specified by column names
  (ON CONSTRAINT is not supported).
- These column names have to be plain names without any
  ordering or null ordering specified.
- DO UPDATE does not provide any relation in its scope.
  That is: columns can only be updated to constants.
  This restriction includes WHERE clause of UPDATE.
- CommandComplete tag (`INSERT 0 2`) might sometimes
  be wrong. Also, RETURNING might sometimes be wrong.

### DIFF
--- a/edb/pgsql/resolver/command.py
+++ b/edb/pgsql/resolver/command.py
@@ -435,11 +435,11 @@ def _uncompile_insert_object_stmt(
 
     if conflict and conflict.update_name is not None:
         assert conflict.update_id
-        assert conflict.update_rel
+        assert conflict.update_input_placeholder
         ql_singletons.add(conflict.update_id)
         ql_anchors[conflict.update_name] = conflict.update_id
         external_rels[conflict.update_id] = (
-            conflict.update_rel, (pgce.PathAspect.SOURCE,),
+            conflict.update_input_placeholder, (pgce.PathAspect.SOURCE,),
         )
 
     return UncompiledDML(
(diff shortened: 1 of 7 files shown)


edb/pgsql/resolver/command.py
```diff
@@ -452,6 +469,192 @@ def _uncompile_insert_object_stmt(
...
+    update_rel: Optional[pgast.Relation] = None
+
+
+def _uncompile_on_conflict(
+    stmt: pgast.InsertStmt,
+    sub: s_objtypes.ObjectType,
+    sub_table: context.Table,
+    ctx: Context,
+    stype_refs: dict[uuid.UUID, list[qlast.Set]]
+) -> Optional[UncompileOnConflict]:
```
msullivan, line 488: Add a big comment describing what the edgeql we are generating is and what we do with it?

edb/pgsql/resolver/command.py
```diff
@@ -1770,6 +1994,42 @@ def _compile_uncompiled_dml(
     return result, ctes
 
 
+def _collect_stmt_ctes(
+    ctes: list[pgast.CommonTableExpr],
+    ir_stmt: irast.MutatingStmt
+) -> list[pgast.CommonTableExpr]:
+    # We compile all SQL DML queries in a single EdgeQL query.
+    # Result is an enormous SQL query with many CTEs.
+    # This function looks trough these CTEs and matches them to the original
```
msullivan, line 2003: s/trough/through/

msullivan (approved):
Great work

Issue #8458, opened 2025-03-10
Support doing DML in READ REPEATABLE transactions


msullivan:
The big thing that is preventing us from doing writes from READ REPEATABLE transactions is the spectre of #2875, where conflicting data could be inserted into related types in separate transactions, because both transactions are operating on a snapshot taken before either transaction started.

Some possible approaches:

1. READ REPEATABLE DML safety inference. Allow doing the DML if we know it doesn't require looking at other types for exclusive conflict checks. Pros: Easy to do, no cost. Cons: Doesn't solve whole problem.
2. Shadow tables? Have extra side tables for the parent types containing just the fields used in exclusive constraints. When inserting into an object, you would also insert into the shadow tables for it and all its ancestors. Pros: This would let us enforce the constraints just using UNIQUE constraints with no messing about. Cons: likely to be expensive and a decent amount of work.
3. Try to use pg_try_advisory_xact_lock with a data-derived key to explicitly detect when there is a conflict. We could potentially raise either a ConstraintViolation or a SerializationError; we'd need to think about it. Pros: Implementation should be lightweight, doesn't require a data model change. Cons: Deriving the key will be hacky, potentially slow. Just feels dodgy overall.

I think something like 3 is probably the right approach here.

I was hoping that it would work to put the lock in a trigger, but it doesn't, so we get to extend the `conflicts` machinery some more.

elprans:
Should probably be a combination of 1 and 3.

msullivan:
Update: 3 fully does not work, and I'm embarrassed that I thought it could.

If connection A starts a transaction, then connection B inserts an object, B will take and drop the lock before A ever takes it. Then A, inserting a conflicting object, will take the lock, but succeed.

msullivan:
I'm going to close this for now, since we have it working in many cases

PR #8561, opened 2025-04-04
Allow DML in REPEATABLE READ transactions but perform safety checks

msullivan:
- In conflicts.py, implement analysis for what inserts and updates
   affect a cross-table constraint
 - Plumb that information out into the QueryUnit and to the server
 - Report the conflicts as part of CommandDataDescription in an
   annotation called `unsafe_isolation_dangers`, so clients can
   potentially make proactive decisions.
 - Raise an error when attempting to execute something with isolation
   dangers in REPEATABLE READ mode.

There are still a few follow-ups before I can merge, but this is
reviewable now:
 - [x] Write tests
 - [x] Follow-up on some possible bugs I found in the conflict mechanism
       that might have wider scope once this is implemented.
 - [x] Improve some documentation

Advances #8458.

### DIFF
--- a/edb/api/errors.txt
+++ b/edb/api/errors.txt
@@ -44,6 +44,7 @@
 0x_03_04_00_00   CapabilityError
 0x_03_04_01_00   UnsupportedCapabilityError
 0x_03_04_02_00   DisabledCapabilityError
+0x_03_04_03_00   UnsafeIsolationLevelError
 
 
 ####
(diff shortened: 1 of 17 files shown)


elprans:
> * Report the conflicts as part of CommandDataDescription in an annotation called `unsafe_isolation_dangers`

Should we also raise it as a `Capability` bit to make client impl simpler?

msullivan:
> > * Report the conflicts as part of CommandDataDescription in an annotation called `unsafe_isolation_dangers`
> 
> Should we also raise it as a `Capability` bit to make client impl simpler?

Would it help much?

I'm also nervous about capability updates, because I think clients differ based on whether they default to enabling things they don't know about or disabling them. (Sigh; and I ran into that years ago and never followed up---sorry).
The last time I dealt with this I screwed it up a lot and rolled the new capability bit back fully (#4688)

aljazerzen (approved):

elprans (approved):

PR #8562, opened 2025-04-04
sql dml output

aljazerzen:
- **ON CONFLICT basic impl over SQL adapter**
- **comments, minor refactoring**
- **refactor of fini_dml_stmt**
- **fix exception handling on pg proto**
- **Fix how we pick output CTE of DML over SQL adapter**


msullivan (approved):
This looks great; could you put a bit of the info in the (great!) commit message into comments?

PR #8571, opened 2025-04-07
Pin calver to 2025.03.31

aljazerzen:
calver 2025.04.01 is broken (pyproject license)
calver 2025.04.02 requires a new version of setuptools, which are hard to upgrade. We are currently pinned on <70.2. I've tried updating that: https://github.com/geldata/gel-pkg/pull/152

Ref: https://github.com/geldata/gel-pkg/issues/153


mmastrac (approved):

aljazerzen:
https://github.com/geldata/gel/actions/runs/14320396745

PR #8573, opened 2025-04-08
Fix MacOS build by installing sentencepiece

aljazerzen:
It started failing recently with the following
error message:

```
Package sentencepiece was not found in the pkg-config search path.
Perhaps you should add the directory containing `sentencepiece.pc'
to the PKG_CONFIG_PATH environment variable
Package 'sentencepiece' not found
CMake Error at CMakeLists.txt:15 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.

-- Configuring incomplete, errors occurred!
./build_bundled.sh: line 22: nproc: command not found
Error: could not find CMAKE_PROJECT_NAME in Cache
```

If missing sentencepiece is the problem, installing
it with brew should fix the build.

### DIFF
--- a/.github/workflows.src/build.inc.yml
+++ b/.github/workflows.src/build.inc.yml
@@ -149,7 +149,7 @@
     - name: Install dependencies
       if: << 'false' if tgt.runs_on and 'self-hosted' in tgt.runs_on else 'true' >>
       run: |
-        env HOMEBREW_NO_AUTO_UPDATE=1 brew install libmagic
+        env HOMEBREW_NO_AUTO_UPDATE=1 brew install libmagic sentencepiece
 
     - name: Build
       env:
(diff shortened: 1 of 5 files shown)


msullivan (approved):

msullivan:
Let's try it but this does actually seem kind of dodgy

aljazerzen:
Latest run:

https://github.com/geldata/gel/actions/runs/14337625917/job/40188717028

... which is now failing with the same error about calver as the other builds, which I count as issue fixed.

PR #8577, opened 2025-04-09
Add reference docs for `extension` command group

scotttrinh:
Also remove vestigial `-E` option in extension guide.

### DIFF
--- a/docs/reference/using/cli/gel_extension/list-available.rst
+++ b/docs/reference/using/cli/gel_extension/list-available.rst
@@ -1,0 +1,27 @@
+.. _ref_cli_gel_extension_list_available:
+
+============================
+gel extension list-available
+============================
+
+List available extensions.
+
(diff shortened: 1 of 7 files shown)


docs/reference/using/cli/gel_extension/list-available.rst
```diff
@@ -0,0 +1,27 @@
...
+
+============================
+gel extension list-available
+============================
+
+List available extensions.
+
+.. cli:synopsis::
+
+  gel extension list-available [<options>]
```
beerose, line 11: Unrelated and nitpick, but is there a plan for a shortcut for this? `list-available` is long to type 😅

scotttrinh: Eh, hopefully this something you need to do often enough for it to be annoying 😅 Maybe one day we can have an alias feature like git.

PR #8579, opened 2025-04-09
Ensure server conversions don't cache normalized values.

dnwpark:


msullivan (commented):
I'm missing something about what is happening here --- 456 ought to be getting extracted anyway and passed as a parameter, and we *should* be caching the query, right?

dnwpark:
@msullivan 456 is put into `Source.extra_blobs` by the tokenizer, but we get no information about the contents of that blob.

In order to convert 456, we read the value in `compiler._compile_ql_query` (just a few lines above the change). Then we store that value in the ServerParamConversion. Since this is part of the QueryUnitGroup, it can get cached, but we don't want that to happen.

Ideally, we would have more information about the extra blobs during `execute._convert_parameters` but I think that would require some changes to the tokenizer.

msullivan (approved):
OK, I think we can land this *for now* since it is a bug fix.

But my inclination is that we ought to fix it and properly cache these queries before we release it in a major version?

msullivan:
Tag for backport still though

PR #8582, opened 2025-04-09
Cleanup a hack for DML on inherited links over SQL adapter

aljazerzen:
For the following query:

```sql
WITH d AS (
  INSERT INTO "Document" (title) VALUES ('Report') RETURNING id
)
INSERT INTO "Document.shared_with" (source, target, can_edit)
SELECT d.id, $1, TRUE FROM d
```
... we generate an EdgeQL:
```
select {
  dml_0 :=insert Document,
  dml_1 :=update Document { shared_with += ...}
}
```

But to validate that ids inserted into `shared_with` exists in `Document`,
we have to join with `Document` prior to insertion. That is problematic,
because we cannot just use the concrete table, since it does not yet
contain the inserted `Document`.

Until now, we used a hack that just didn't join with the concrete table.

This PR makes it so we join with concrete table AND the overlay that contains
objects created during current query.

And removes the hack.

---

I've found a problem with this hack when I was trying to add a followup test 
case for #8562. The test case is not working correctly yet.

### DIFF
--- a/edb/pgsql/compiler/dml.py
+++ b/edb/pgsql/compiler/dml.py
@@ -321,10 +321,6 @@ def gen_dml_cte(
     # need them.
     ctx.path_scope.maps.clear()
 
-    skip_rel = (
-        isinstance(ir_stmt, irast.UpdateStmt) and ir_stmt.sql_mode_link_only
-    )
-
     if range_rvar is not None:
         relctx.pull_path_namespace(
             target=dml_stmt, source=range_rvar, ctx=ctx)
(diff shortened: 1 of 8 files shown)


edb/pgsql/compiler/dml.py
```diff
@@ -245,40 +245,68 @@ def gen_dml_cte(
...
+
+        subject_rel_overlayed = relctx.range_for_typeref(
+            typeref,
+            subject_path_id,
+            for_mutation=False,
+            include_descendants=False,
+            dml_source=[
+                k for k in ctx.dml_stmts.keys()
+                if isinstance(k, irast.MutatingLikeStmt)
+            ],
```
msullivan, line 283: Do we want to only do this in SQL mode, though?

We should also document why we do this.

msullivan (approved):

msullivan:
to backport?

edb/pgsql/compiler/dml.py
aljazerzen: We could yeah, but I don't have a nice variable close by that would indicate if this is "SQL mode" or not.

PR #8590, opened 2025-04-10
Fix memory leak in pgrust-pgcon

fantix:
PGConnectionProtocol - PyConnectionState - PyConnectionStateUpdate form a cross-Rust-Python circular reference loop that cannot be detected by Python GC, leading to memory leak when creating new backend connections.

Fixed by manually breaking the loop once we're done with the PGConnectionProtocol.

The test is verified to fail without the patch.


msullivan (approved):
Great work finding this!

mmastrac (approved):
Nice find

PR #8591, opened 2025-04-10
Fix _clear_fe_local_sql_settings breaking inplace upgrades

msullivan:
_clear_fe_local_sql_settings was not versioned, which broke IPUs. It
wasn't versioned because it could not be trampolined. But we don't
need to trampoline it, so arranged to version but not trampoline it.


msullivan:
The IPU tests still failed, due to still needing #8593. Going to merge this now.

PR #8595, opened 2025-04-11
ON CONFLICT subject rvar over SQL adapter

aljazerzen:
This PR makes subject rvar accessible from ON CONFLICT UPDATE and WHERE
clauses.

This allows a cleanup of hack from #8587.

Specifically, this PR:
- moves "conflict update" CTE to be *after* "else" CTE (where rows with
  conflicts are computed),
- uses "else" CTE as the source of values for subject's columns,
- enforces correlation between "excluded" and "subject" rvars by
  joining on the iterator that is generated in insert value relation.
  We cannot use real ids for this because they do not exist (yet) for
  these inserted objects.
- removes the hacky pgcompiler join of "conflict update" relation and
  update "range" and replaces it by EdgeQL comparison of ids of the
  subject and update iterator.

To make all of these things work, the *resolving* part of ON CONFLICT
needs more data about the subject, dml value and pg Environment, so it
can use pg compiler's `get_path_output` machinery. This is not the ideal
apprach, because it is supposed to be isolated within pg compiler.

This PR contains a few snippets of low code quality (because they break
boundaries between abstractions and then have to depend on internal
details of the implementation). An example of this is matching the CTE
by name. Fortuantely, it is isolated in SQL DML resolver.


aljazerzen:
Ref #8259

edb/pgsql/resolver/command.py
```diff
@@ -2315,17 +2396,82 @@ def _resolve_conflict_update_rel(
...
 
-    cu_table.alias = ctx.alias_generator.get('rel')
+        # We start with a plain value_id, but because of for loops, rvar_map
+        # will contain path_ids polluted with namespaces. So instead of a plain
+        # one, we find a path_id from rvar_map that matches the plain one in all
+        # but the namespace.
+        value_id = next(
+            p for p, _ in else_cte.query.path_rvar_map.keys()
+            if p.replace_namespace(set()) == compiled_dml.value_id
+        )
```
msullivan, line 2420: Another option would be to use the IR to find it?

edb/pgsql/resolver/command.py
```diff
@@ -2261,6 +2294,7 @@ def _resolve_dml_value_rel(
 
 def _resolve_conflict_update_rel(
     compiled_dml: context.CompiledDML,
+    subject_name: tuple[str, str | None],
```
msullivan, line 2297: Add a docstring sketching out what the relevant CTEs and joins are?

msullivan (approved):
Great work on this!

edb/pgsql/resolver/command.py
aljazerzen: Yes, that's possible, but:
- we'd have to attach the IR stmt from the middle step (in `_compile_uncompiled_dml`) to `CompiledDML` so it can be used later on during resolving,
- value_id does not appear at a convenient location in IR. It is used only when referencing the value relation to get values for the update shape
image

Alternatively, we could just look at namespaces on the subject's path_id and apply those namespaces to the plain value_id.

Issue #8200, opened 2025-01-10
Implement ai::search(... "text")


anbuzin:
We discussed Python AI API updates ([PR](https://github.com/edgedb/rfcs/pull/102)) and concluded that having an `ai::search(... "text")` function in the database is the cleanest way to fix peculiarities of the current interface.

msullivan:
The difficulty to resolve here is that we need to turn the text into an embedding, and that means an external API call that we don't want to do from inside a postgres query.

We will be able to support it only for string literals (handled by extraction) and query parameters, which is ugly but explainable.

What we need to do is extract the argument, send it off to the API call, and replace it with 

Elvis had some ideas about making this type-driven, but I didn't love that. I think my proposal is to allow extension functions to mark arguments as requiring server-based rewriting, along with what the replacement type will be. Then the compiler will disallow uses that are not query parameters, and inject a new parameter for the rewritten version.
This information will get described in the QueryUnit somehow, which will tell the server what to do.

anbuzin:
One more suggestion: in addition to this, we could have a `ext::ai::to_vector("text")` function that would be used like this:

```edgeql
with v := ext::ai::to_vector("some text")
select ext::ai::search(Type, v);
```

@elprans thinks is would require a generic vector type to be implemented first, so maybe just keep this one in mind.

PR #8598, opened 2025-04-11
Enable queries with server param conversions to be cached.

dnwpark:
Related #8200

The server param conversions previously explicitly stored any normalized values alongside other conversion information. This meant that they could not be cached.

This PR replaces the explicitly stored value with information about where in the extra blobs a conversion needs to look for its input data.

The tokenizer has been updated to provide a dictionary mapping normalized parameter names to its blob index and index within the blob.

For scripts, the server now produces a dictionary of `converted_arg`s, one for each query unit ahead of time, instead of leaving it to `recode_bind_args_for_script` to assemble the appropriate args.

### DIFF
--- a/edb/edgeql-parser/edgeql-parser-python/src/pynormalize.rs
+++ b/edb/edgeql-parser/edgeql-parser-python/src/pynormalize.rs
@@ -227,7 +227,7 @@ pub fn serialize_extra(variables: &[Variable]) -> Result<(Bytes, Vec<usize>), St
                 .map_err(|_| "element isn't too long".to_owned())?
                 .to_be_bytes(),
         );
-        offsets.push(pos + 4);
+        offsets.push(pos);
     }
     offsets.push(buf.len());
     Ok((buf.freeze(), offsets))
--- a/edb/edgeql-parser/edgeql-parser-python/src/pynormalize.rs
+++ b/edb/edgeql-parser/edgeql-parser-python/src/pynormalize.rs
@@ -8,7 +8,7 @@ use gel_protocol::codec;
 use gel_protocol::model::{BigInt, Decimal};
 use pyo3::exceptions::{PyAssertionError, PyValueError};
 use pyo3::prelude::*;
-use pyo3::types::{PyBytes, PyDict, PyFloat, PyInt, PyList, PyString};
+use pyo3::types::{PyBytes, PyDict, PyFloat, PyInt, PyList, PyString, PyTuple};
 
 use crate::errors::SyntaxError;
 use crate::normalize::{normalize as _normalize, Error, PackedEntry, Variable};
@@ -42,6 +42,9 @@ pub struct Entry {
     #[pyo3(get)]
     extra_blobs: PyObject,
 
+    #[pyo3(get)]
+    extra_offsets: PyObject,
+
     extra_named: bool,
 
     #[pyo3(get)]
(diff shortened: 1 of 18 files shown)


edb/edgeql-parser/edgeql-parser-python/src/pynormalize.rs
```diff
@@ -98,12 +106,80 @@ impl Entry {
...
+}
+
+pub fn serialize_extra(variables: &[Variable]) -> Result<(Bytes, Vec<usize>), String> {
     use gel_protocol::codec::Codec;
     use gel_protocol::value::Value as P;
 
     let mut buf = BytesMut::new();
+    let mut offsets = Vec::new();
     buf.reserve(4 * variables.len());
+    offsets.reserve(1 + variables.len());
```
mmastrac, line 182: These can be replaced with new_with_capacity

edb/edgeql-parser/edgeql-parser-python/src/pynormalize.rs
```diff
@@ -98,12 +106,80 @@ impl Entry {
...
+        };
+        if self.blob_index >= self.entry_pack.variables.len() {
+            return None;
+        }
+        if self.var_index >= self.entry_pack.variables[self.blob_index].len() {
+            return None;
+        }
+
+        // Get result
+        let blob_vars = &self.entry_pack.variables[self.blob_index];
```
mmastrac, line 148: Nothing wrong with this code, but the blob vars part of this could be replaced with an if let ... using `.get(...)` rather than indexing which will help make it more "obviously correct".

edb/edgeql-parser/edgeql-parser-python/src/pynormalize.rs
```diff
@@ -98,12 +106,80 @@ impl Entry {
...
+
+impl<'a> Iterator for VariableNameIter<'a> {
+    type Item = (String, usize, usize, &'a Value);
+
+    fn next(&mut self) -> Option<Self::Item> {
+        // Check termination
+        let first = match self.first_extra {
+            Some(first) => first,
+            None => return None,
+        };
```
mmastrac, line 139: More concisely written as `let first = self.first_extra?;`, I believe. This will use the Rust try operation to early exit. I believe clippy might recommend this to you.

mmastrac:
Rust bits LGTM

edb/server/protocol/args_ser.pyx
```diff
@@ -685,59 +685,94 @@ cdef list[ParamConversion] get_param_conversions(
...
-    param_names = set(
+    # from bind args
+    bind_args_names: set[str] = set(
         param_conversion.param_name
         for param_conversion in server_param_conversions
     )
 
     # Next extract bytes from the bind_args
-    param_datas: dict[str, bytes] = {}
+    bind_arg_datas: dict[str, bytes] = {}
```
msullivan, line 695: Could this whole setup be made simpler and involve a little less data being threaded through if we handled the extracted arguments similarly to the real ones, and parsed the extra blob in `get_param_conversions`?

Not sure if it is worth doing extra work to do that with everything working already, though

msullivan (approved):

edb/server/protocol/execute.pyx
```diff
@@ -451,15 +452,98 @@ async def execute(
...
+
+            elif extra_blob_offset_indexes := (
+                conversion.get_extra_blob_offset_indexes()
+            ):
+                blob_index, offset_index = extra_blob_offset_indexes
+                # skip the length part of the data
+                offset_start = compiled.extra_offsets[blob_index][offset_index] + 4
+                offset_end = compiled.extra_offsets[blob_index][offset_index+1]
+                data = compiled.extra_blobs[blob_index][offset_start:offset_end]
+                # Don't cache conversion, since it may differ per unit
```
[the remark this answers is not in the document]
dnwpark: Reworked the extra blobs indexing so that these offsets are not needed, and the cache acts the same for all cases.

edb/server/protocol/args_ser.pyx
dnwpark: The data needs to be threaded through regardless, but I've reworked things to be hopefully clearer.

msullivan: Why?

msullivan: I think I must be missing something important

PR #8603, opened 2025-04-14
Batch embeddings requests in queries with multiple ai text search calls.

### LINKED ISSUE #8200: Implement ai::search(... "text")

dnwpark:
Close #8200 

- Refactor the ai extension code to simplify text truncation and batching requests
- Add `ai_ext.generate_embeddings_for_texts` which gets embeddings for an arbitrary number of inputs which may have different indexes, models, providers, etc.
- Batch the param conversions for ai text search

### DIFF
--- a/edb/server/protocol/ai_ext.py
+++ b/edb/server/protocol/ai_ext.py
@@ -970,9 +970,10 @@ def _batch_embeddings_inputs(
     ]
 
     # Get indexes of inputs, sorted from shortest to longest by token count
+    # Use the text itself as a tie breaker for consistency
     unbatched_input_indexes = list(range(len(inputs)))
     unbatched_input_indexes.sort(
-        key=lambda index: input_token_counts[index],
+        key=lambda index: (input_token_counts[index], inputs[index]),
         reverse=False,
     )
 
--- a/edb/server/protocol/ai_ext.py
+++ b/edb/server/protocol/ai_ext.py
@@ -129,12 +129,14 @@ class Tokenizer(abc.ABC):
         """Decode tokens into text."""
         raise NotImplementedError
 
-    def shorten_to_token_length(self, text: str, token_length: int) -> str:
+    def shorten_to_token_length(
+        self, text: str, token_length: int
+    ) -> tuple[str, int]:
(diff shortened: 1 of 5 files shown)


dnwpark:
Note: most of the changes are new tests, I also renamed and moved `test_ext_ai_indexing_06` to `test_ext_ai_text_search_01`

edb/server/protocol/ai_ext.py
```diff
@@ -2592,12 +2658,210 @@ async def generate_embeddings_for_text(
...
+    for group_key, batched_texts_and_indexes in (
+        group_batch_texts_and_indexes.items()
+    ):
+        model_name, shortening = group_key
+        provider = model_providers[model_name]
+
+        provider_config = provider_configs[provider]
+
+        for batched_texts, batched_input_indexes in batched_texts_and_indexes:
+            embeddings_result = await _generate_embeddings(
```
elprans, line 2834: Can we do this concurrently (i.e in a taskgroup) or will this be tricky with rate-limits?

dnwpark: In theory dealing with the rate limits is not hard, but there are some things to consider.

- The embeddings worker is caching rate/token limits in `provider_schedulers`, which is local to `_ext_ai_index_builder_controller_loop`.
  - It may be possible to share limits (at least total limits), with `generate_embeddings_for_texts`, but it might get fiddly.
  - If we have no rate limit yet (eg in the case of a new index), should we share this limit back?
- The rate limits are shared between every application that uses the same API key, including the embeddings worker itself.
  - The worker uses a backoff mechanism to retry requests that fail in this case.
  - Should this function do this too and return the query result after a longer delay?

I feel like for now, the implementation is simple enough that we can re-visit if people really do start using multiple calls to `ext::ai::search` with very large blocks of text within a search. Since the function calls are restricted to only searching for text constants or query parameters, I feel this is unlikely.

elprans (approved):

PR #8605, opened 2025-04-15
Bump pg version that we expose over SQL adapter

aljazerzen:
I've bumped just the `server_version` and `server_version_num`, all
behavior remains the same (including pg_catalog, which mostly follows
the pg_catalog of the backend postgres).


aljazerzen:
Oh, of course pg_dump changes behavior when we present a different version. I'll have to fixup the pg_catalog afterall.

Issue #4309, opened 2022-08-31
Support nested arrays


msullivan:
postgres does not directly support nested arrays (`array<array<str>>`) and so we don't either. However, it does work when wrapped in tuples (`array<tuple<array<str>>>`).

We probably should be able to do this transformation automatically.

Sikarii:
I asked Colin earlier, but is there a limitation as to why tuples inside arrays have weird looking type names when introspected?
Interesting as in the colons seemingly get replaced by `|` when nested.

`select schema::Array { name }`
```
{
  schema::Array {name: 'array<anytype>'},
  schema::Array {name: 'array<std::str>'},
  schema::Array {name: 'array<std::json>'},
  schema::Array {name: 'array<std::int16>'},
  schema::Array {name: 'array<cal::local_datetime>'},
  schema::Array {name: 'array<cal::local_date>'},
  schema::Array {name: 'array<cal::local_time>'},
  schema::Array {name: 'array<cal::relative_duration>'},
  schema::Array {name: 'array<cal::date_duration>'},
  schema::Array {name: 'array<std::uuid>'},
  schema::Array {name: 'array<tuple<text:std|str, refs:array<std||uuid>>>'},
  schema::Array {name: 'array<tuple<std|int16, std|int32>>'},
  schema::Array {name: 'array<tuple<std|str, std|int16, std|int32, std|int64, array<tuple<std||||int16, std||||int32>>>>'},
}
```

PR #8611, opened 2025-04-16
Support non-computed nested array globals.

dnwpark:
related #4309

Allow globals such as:
```
global foo -> array<array<int64>>;
```

Which can then be set by:
```
set global foo := [[1, 2, 3], [4, 5, 6]];
```

### DIFF
--- a/edb/pgsql/compiler/astutils.py
+++ b/edb/pgsql/compiler/astutils.py
@@ -139,84 +139,113 @@ def array_get_inner_array(
     an element is accessed, it needs to be unwrapped.
 
     Essentially, this function takes tuple<array<...>> and returns array<...>
-
-    Postgres does not support arbitrarily accessing fields out of unnamed
-    composites and so we need to do an extra unnest(array[]) to be able to
-    specify the name and type our resulting array.
-
-    For example, the query: `select [[1]][0];` will produce the following SQL:
-
-    SELECT
-            "expr-6~2"."array_value~4" AS "array_serialized~1"
-        FROM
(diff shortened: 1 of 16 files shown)


edb/pgsql/compiler/astutils.py
```diff
@@ -139,84 +139,113 @@ def array_get_inner_array(
...
+                                        ARRAY[ROW("expr-2~2"."array_value~1")]
+                                        AS "array_value~2"
+                                    FROM
+                                        LATERAL
+                                        (SELECT
+                                                ARRAY[($1)::int8]
+                                                AS "array_value~1"
+                                        ) AS "expr-2~2"
+                                ) AS "expr-3~2"
+                        ) AS "expr-5~2"
```
msullivan, line 212: I know this is from a previous PR, but trim the comment example code down. It includes some extra irrelevant stuff: an extra wrapper to produce serialized, the toplevel NULL filter, an inject LIMIT

msullivan:
I found one problem while poking around with testing:
```
_localdev:main> 
select schema::Array { name } filter .name = 'array<array<std|int64>>';
{}
_localdev:main> 
create global compositeAA -> array<array<int64>>;
OK: CREATE GLOBAL
_localdev:main> 
drop global compositeAA;
OK: DROP GLOBAL
_localdev:main> 
select schema::Array { name } filter .name = 'array<array<std|int64>>';
{Object {name: 'array<array<std|int64>>'}}
```

Not all of the types get cleaned up on delete.

PR #8612, opened 2025-04-16
Remove extra blob arg indexes from server param conversions.

dnwpark:
The server param conversions passed in a `extra_blob_arg_indexes` which described how to extract the parameter's conversion data from the extra blobs. This is redundant if the bind_args_index is extended to index into the extra blobs.

Related #8200


msullivan (approved):
Looking great

msullivan:
Tag for backport?

PR #8614, opened 2025-04-17
Improve syntactial keyword suggestion for gel-ls

aljazerzen:
Add syntactical suggestions to gel-ls.

We did already have a basic impl of this which relied on looking at
errors produced by a parsing of the whole file.

This PR replaces that with:
- tokenizing of the whole file,
- slicing tokens before the cursor,
- parsing without error recovery,
- inspection of the parser state and extraction of possible next tokens.

This extraction looks at available actions in current state.
An important detail is that not all of these actions are valid.
They might trigger a chain of reductions that ends in a state that
does not accept the suggested token.

This approach works great for DDL and SDL which are heavily based on keywords.
So `START` completes with `MIGRATION` and `TRANSACTION`.
See added tests for more examples.

It does not work well for completing "identifier-like" elements.
That's because of our different keyword categories.
I've made it so that when `Ident` is in suggestions, we don't suggest
unreserved keywords.
This removes all moot suggestions for `create type _`.
But it also removes suggestion `property` from within `type X { ... }`.
Also, it does get rid of reserved keywords from locations where
reserved keywords can be identifiers i.e. in `create type default::_`.

Maybe we should simulate more of action application to see if suggested
token results in any kind of ident production.
And reject only those suggestions.

Idea: we could add "suggestion priority" to grammar rules, which would
put them to the top of the suggestion list.


edb/edgeql-parser/src/parser/mod.rs
```diff
@@ -206,6 +206,54 @@ pub fn parse<'a>(input: &'a [Terminal], ctx: &'a Context) -> (Option<CSTNode<'a>
     (node, errors)
 }
 
+pub fn suggest_next_keyword<'a>(input: &'a [Terminal], ctx: &'a Context) -> Vec<Keyword> {
```
msullivan, line 209: Add some comments: what the goal is, what the algorithm is

edb/edgeql-parser/src/parser/mod.rs
```diff
@@ -206,6 +206,54 @@ pub fn parse<'a>(input: &'a [Terminal], ctx: &'a Context) -> (Option<CSTNode<'a>
...
+        .filter_map(|kind| {
+            if let Kind::Keyword(keyword) = kind {
+                Some(*keyword)
+            } else {
+                None
+            }
+        })
+        // never suggest dunder or bools, they should be suggested semantically
+        .filter(|k| !k.is_dunder() && !k.is_bool())
+        // if next token cab be ident, hide all unreserved keywords
```
msullivan, line 250: s/cab/can/

edb/edgeql-parser/src/parser/mod.rs
```diff
@@ -394,6 +442,43 @@ impl<'s> Parser<'s> {
...
+                    let mut to_pop = reduce.cnt;
+                    let cancel_out = usize::min(pushed, to_pop);
+                    to_pop -= cancel_out;
+                    pushed -= cancel_out;
+                    for _ in 0..to_pop {
+                        node = node.parent.as_ref().unwrap();
+                    }
+
+                    // get state of current stack top
+                    let stack_state = if reduce.cnt == 0 { state } else { node.state };
```
msullivan, line 472: Can't we always take `node.state`?

edb/edgeql-parser/src/parser/mod.rs
```diff
@@ -394,6 +442,43 @@ impl<'s> Parser<'s> {
...
+
+        loop {
+            // find next action
+            let action = ctx.spec.actions[state].get(token)?;
+
+            match action {
+                Action::Shift(next) => {
+                    return Some(*next);
+                }
+                Action::Reduce(reduce) => {
```
msullivan, line 461: We simulate Reduce in case the return is to somewhere invalid?

msullivan (approved):

edb/edgeql-parser/src/parser/mod.rs
aljazerzen: No, we must simulate Reduce in case `goto` (of the reduce) takes us into a state that does not accept current token.

This is the line that will fail out if that happens: https://github.com/geldata/gel/pull/8614/files/6a2ba1437ee39283b1354ff8db0a701fb0e6a8af#diff-8edb2474c2efee3785b6c9aa287628641a7e1879b259eaaced1989d76a4db5b4R455
```
let action = ctx.spec.actions[state].get(token)?;
```

edb/edgeql-parser/src/parser/mod.rs
aljazerzen: No, because when we should have pushed onto the stack, we really didn't because I try to be conservative with allocations.

Instead, I save that into `pushed` count. We never really need the top stack node, because we pop nodes down from the stack at the beginning of the reduce. Unless if we don't because `reduce.cnt` is zero. In that case we have to look for the state the top node would have been, if we had pushed, and that is just state.

Complicated, but we don't need to know how this works because we have the proper `act` impl for that.

aljazerzen: I've added comments and named these "nodes that we should have added" as "ghost nodes"

PR #8620, opened 2025-04-22
Tweak parser recovery

### LINKED ISSUE #8609: Bad error message
- EdgeDB Version: 7.0-dev.9502+local
- EdgeDB CLI Version: 7.2.0+c5b8c5e

Steps to Reproduce:

```
_localdev:main> std::assert_single((select 1));
error: EdgeQLSyntaxError: Missing keyword 'IF'
  ┌─ <query>:1:1
  │
1 │ std::assert_single((select 1));
  │ ^ error
```

It feels like `select` would be a better suggestion for the missing keyword.

aljazerzen:
Closes #8609

This PR increases cost of `if` and `group` keywords to prevent them from
being suggested, since they are usually not what is missing.

Also, I've adjusted how parsers are recovered and added a special path
for parsers with custom errors: when we detect a custom error, we will
not try to fix any other error and just skip tokens until EOI. This will
have the effect of never producing an AST if there is a custom file, but
reporting just the custom error and no other errors.

It does add a guarantee that the custom error will never be discarded in
lieu of other recovery paths.

### DIFF
--- a/edb/edgeql-parser/src/parser/mod.rs
+++ b/edb/edgeql-parser/src/parser/mod.rs
@@ -71,6 +71,7 @@ pub fn parse<'a>(input: &'a [Terminal], ctx: &'a Context) -> (Option<CSTNode<'a>
 
     for token in input {
         // println!("token {:?}", token);
+
         while let Some(mut parser) = parsers.pop() {
             let res = parser.act(ctx, token);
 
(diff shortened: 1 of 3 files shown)


edb/edgeql-parser/src/parser/mod.rs
```diff
@@ -154,36 +159,43 @@ pub fn parse<'a>(input: &'a [Terminal], ctx: &'a Context) -> (Option<CSTNode<'a>
...
+        //     print!(
+        //         "p{index} {:06} {:5}:",
+        //         parser.error_cost, parser.can_recover
+        //     );
+        //     for e in &parser.errors {
+        //         print!(" {}", e.message);
+        //     }
+        //     println!("");
+        // }
+        // println!("");
```
msullivan, line 198: drop?

msullivan (approved):
Will only using the custom error interfere with some of our other plans for the parser?

aljazerzen:
It was already interfering with the parser recovery we already use to produce error messages.

In many cases, we were able to find other recovery paths that fully recovered the parser, which then discarded the custom recovery rule.

PR #8623, opened 2025-04-22
Fix proto hang running script after setting default_transaction_*

msullivan:
If `state_reset_needs_commit` is set in `send_query_unit_group`, then
it will ask for a sync, even if needs_commit_state was not set. But
`wait_for_state_resp` didn't know about this, so it did not wait for
the sync, which then arrived later, during `wait_for_command`, which
choked on it during `fallthrough`.

Then, during the cleanup for `parse_execute_script_context`, the
processing hung forever looking for a sync that already arrived.

So, we:
 * Make sure wait_for_state_resp uses the same wait condition
 * Process a sync in fallthrough before raising the exception,
   so that a sync appearing at the wrong time due to a state machine
   bug doesn't get caught later waiting for the sync.

There is a test that triggers this in a pending gel-python patch I'm
working on.

I wasn't able to produce a test that worked in this test suite,
because our testbase connection's `with_options` is broken for reasons
that are still mysterious to me. The state *is* getting threaded in,
so I'm guessing that something is wrong with the codec and it isn't
getting encoded right.

### DIFF
--- a/edb/server/pgcon/pgcon.pyx
+++ b/edb/server/pgcon/pgcon.pyx
@@ -571,6 +571,7 @@ cdef class PGConnection:
 
         if state is not None and start == 0:
             self._build_apply_state_req(state, out)
+            # N.B: Condition here needs to match that in wait_for_state_resp
             if needs_commit_state or self.state_reset_needs_commit:
                 self.write_sync(out)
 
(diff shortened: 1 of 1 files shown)


fantix (approved):
Ahhh, good catch!

PR #8626, opened 2025-04-23
Refactor language server code into more modules

aljazerzen:


msullivan (approved):
Add a commit message describingi the new organization. And probably a comment too somewhere?

aljazerzen:
The reason on why organized it the way I did is "because this split felt the most balanced".
I also wanted to move code out of the two generic files "main.py" and "server.py", since they don't add much to code discoverability.

There are going to be "declaration.py" and "completion.py" for the two requests we support. Maybe I should also make a "diagnostics.py" for the `compile` function.

PR #8632, opened 2025-04-24
Semantic completion for LSP

aljazerzen:
Currently limited to .edgeql files and suggestions of object types and
local with bindings.

How it works:
- we start with source (e.g. `select _;`, where `_` is cursor position),
- we parse the file with recovery which hopefully produces an AST with
  bogus node at the cursor location (usually `()` is injected, so we get
  `SelectQuery(result=Tuple(elements=[]))`,
- we replace the node under the cursor with `qlast.Cursor`
  (we get `SelectQuery(result=Cursor)`),
- in qlcompiler, we raise `IdentCompletionException`, which halts
  compilation and contains the list of completion items.

We need the compilation context for its state: scopes, with bindings,
partial_path_prefix, schema are all in there. This way, we will be able
to complete names of computed pointers that have been defined in the
current query.


aljazerzen:
https://github.com/user-attachments/assets/25356276-6577-44dc-bab5-097e21b6efc6

edb/edgeql-parser/src/parser/mod.rs
```diff
@@ -224,7 +225,7 @@ pub fn parse<'a>(input: &'a [Terminal], ctx: &'a Context) -> (Option<CSTNode<'a>
 /// An important detail is that not all of these actions are valid.
 /// They might trigger a chain of reductions that ends in a state that
 /// does not accept the suggested token.
-pub fn suggest_next_keyword<'a>(input: &'a [Terminal], ctx: &'a Context) -> Vec<Keyword> {
+pub fn suggest_next_keyword<'a>(input: &'a [Terminal], ctx: &'a Context) -> (Vec<Keyword>, bool) {
```
msullivan, line 228: document the bool?

msullivan (changes requested):
There is too unrelated much code reformatting in `expr.py` here to review that file for actual changes. Please split it into another PR if you feel strongly about merging it.

edb/edgeql/compiler/expr.py
```diff
@@ -1189,3 +1218,39 @@ def try_constant_set(expr: irast.Base) -> Optional[irast.ConstantSet]:
...
+    """An exception that is raised to halt the compilation and return a list of
+    suggested idents to be used at the location of qlast.Cursor node.
+    """
+
+    def __init__(self, suggestions: list[str]):
+        self.suggestions = suggestions
+
+
+@dispatch.compile.register(qlast.Cursor)
+def compile_Cursor(
```
msullivan, line 1233: Some other things worth considering support for: module names, object names inside a module once a module is specified, functions

msullivan (commented):
Otherwise looks great!

msullivan (approved):
Thanks!

PR #8634, opened 2025-04-25
Fix ISE when with binding in dml function contains a set function/operator

dnwpark:
Set funcs/ops such as ?? compile their arguments in a new scope.

If such funcs/ops are used in a volatile binding in a dml function, a non-volatile bindings which is compiled from the volatile binding will not have direct access to the args iterator rvar.

Instead, they need access via the args' scope stmt.


msullivan (approved):
Good catch.
(I might try to take another look tomorrow to see if there is something that can be done to simplify things here, but feel free to merge)

msullivan:
Label for backport?

PR #8635, opened 2025-04-25
Add error when using DML in a read only transaction.

### LINKED ISSUE #8454: Read only transaction error message is wrong
- EdgeDB Version: 6.2 & 7.0-dev.9399
- OS Version: ubuntu 22.04.1

The read only transaction error says `SELECT`, but it seems it should say `INSERT`.

Steps to Reproduce:
```go
opts := gelcfg.NewTxOptions().WithReadOnly(true)
configured := client.WithTxOptions(opts)

err := configured.Tx(ctx, func(ctx context.Context, tx geltypes.Tx) error {
	return tx.Execute(ctx, "INSERT User")
})
fmt.Println(err)
// Output:
// gel.TransactionError: cannot execute SELECT in a read-only transaction
```

Schema:
```
START MIGRATION TO {
	module default {
		global used_everywhere: int64;
		type User {}
	}
};
POPULATE MIGRATION;
COMMIT MIGRATION;
```

dnwpark:
close #8454

### DIFF
--- a/edb/server/compiler/errormech.py
+++ b/edb/server/compiler/errormech.py
@@ -87,10 +87,13 @@ branch_errors = {
     pgerrors.ERROR_DUPLICATE_DATABASE: errors.DuplicateDatabaseDefinitionError,
 }
 
-directly_mappable = {
+directly_mappable: dict[str, type | tuple[type, str]] = {
     pgerrors.ERROR_DIVISION_BY_ZERO: errors.DivisionByZeroError,
     pgerrors.ERROR_INTERVAL_FIELD_OVERFLOW: errors.NumericOutOfRangeError,
-    pgerrors.ERROR_READ_ONLY_SQL_TRANSACTION: errors.TransactionError,
+    pgerrors.ERROR_READ_ONLY_SQL_TRANSACTION: (
+        errors.TransactionError,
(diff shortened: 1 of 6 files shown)


msullivan (changes requested):
We shouldn't do this in the compiler, at all. Whether the transaction is readonly is not part of the cache key for the query cache (and shouldn't be), so it won't work

We should do this check in the server. QueryUnit has a `capabilities` field that has a bit set in it if the query contains DML.

msullivan:
> @msullivan I've moved the checking to be entirely server side.

That is still all on the compiler side (thought it is in `server/compiler`, which is the server-compiler interface layer). But it is still *after* the query cache.

This change will need to be to cython; probably to `execute` and `execute_script` in `execute.pyx`.

A test that is needed here is to execute the query not in a read-only transaction, so that it is cached, and then execute it again in an RO transaction and make sure we get the right error message.

msullivan:
Oh, actually, probably a better idea for how to do it: we could intercept postgres's error message and replace it with a better one, in `errormech.py`.

dnwpark:
@msullivan one more go at this

msullivan (approved):

PR #8638, opened 2025-04-25
Start invalidating evicted cache entries on other frontends

### LINKED ISSUE #8636: Function cache invalidation issue with multiple servers running
We seem to not handle eviction from the function cache properly in a multi-server setup.

Run two servers:
```
edb server --port 5656
edb server --backend-dsn $(curl -s http://localhost:5656/server-info | jq -r .pg_addr.dsn) --port 5657
```

Use this test script:
```
import gel
import random
import time

# just for debug spew, where to draw the line when guessing if the
# query is cached.
# If we make the query more complicated, then it gets really easy to
# distinguish, but a small query is super fast.
CACHE_CUTOFF = 0.004

def main():
    db = gel.create_client()
    msg = None
    while True:
        i = random.randint(0, 2000)
        name = f'x{i:04}'
        t0 = time.time()
        msg = 'OK  '
        try:
            db.query('select {%s := "lol"}' % name)
        except gel.InternalServerError:
            msg = 'FAIL'
        t1 = time.time()
        print(name, msg, "CACHED" if t1-t0 < CACHE_CUTOFF else "uncached")


if __name__ == '__main__':
    main()

```

Then run the script twice, once against each database:
```
GEL_PORT=5656 GEL_CLIENT_TLS_SECURITY=insecure python3 test.py
```
```
GEL_PORT=5657 GEL_CLIENT_TLS_SECURITY=insecure python3 test.py
```

Failures are of the form:
```
1. edb.errors.InternalServerError: function edgedb_v7_2f262f1200.__qh_c7185e59e9593b38371f9a8e8e1d57b8(text) does not exist
```
which come from trying to call a cached function that doesn't exist.

The script repeatedly runs a randomly chosen query out of 2000 possible very simple queries.
The cache size is currently 1000, so this will fill the cache and then have a hit rate of 50%.

msullivan:
We just weren't ever doing that. There was a commented out
signal_sysevent, but that forced a full reread, which was too
expensive to work.

That substantially *reduces* the frequency of errors
(by about two orders of magnitude in the test in #8636),
but there is still a race condition where the other frontends
may not have processed the eviction in time. So we also need
to handle the errors and cause retries, as well as evicting
the entry.

Fixes #8636.

### DIFF
--- a/edb/server/compiler/rpc.pyx
+++ b/edb/server/compiler/rpc.pyx
@@ -314,9 +314,10 @@ cdef class CompilationRequestIdHandle:
         return hash(self.cache_key)
 
     def __eq__(self, rhs) -> bool:
-        if isinstance(rhs, CompilationRequestIdHandle):
+        ty = type(rhs)
+        if ty is CompilationRequestIdHandle:
             return self.cache_key == rhs.cache_key
-        elif isinstance(rhs, CompilationRequest):
+        elif ty is CompilationRequest:
             return self.cache_key == rhs.get_cache_key()
         else:
             return NotImplemented
--- a/edb/server/compiler/rpc.pyx
+++ b/edb/server/compiler/rpc.pyx
@@ -277,7 +277,14 @@ cdef class CompilationRequest:
     def __hash__(self):
         return hash(self.get_cache_key())
 
-    def __eq__(self, other: CompilationRequest) -> bool:
+    def __eq__(self, rhs) -> bool:
+        cdef:
+            CompilationRequest other
+
+        if not isinstance(rhs, CompilationRequest):
+            return NotImplemented
+        other = rhs
+
         return (
             self.source.cache_key() == other.source.cache_key() and
             self.protocol_version == other.protocol_version and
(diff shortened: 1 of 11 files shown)


msullivan:
And in addition to retrying, I think we need to invalidate the cache entry on the failure

msullivan:
I just also added debouncing and batching of the invalidations. That has some downsides, though, and makes the test script in https://github.com/geldata/gel/issues/8636 perform worse. This is because it (by design!) delays invalidation messages, and so entries are more often not invalidated in time, which means we have to do more retries.

Thoughts on this? I think that this probably won't be that much of a problem for realistic workloads: the query is being evicted because it hasn't been used in a long time, so the odds of it showing up in the next 1s are pretty low. (Not so low that we can afford to be *incorrect* in that case, I feel, but probably low enough that a retry is OK?)

edb/server/compiler/rpc.pyx
```diff
@@ -294,6 +301,27 @@ cdef class CompilationRequest:
...
+        self.cache_key = cache_key
+
+    def __hash__(self):
+        return hash(self.cache_key)
+
+    def __eq__(self, rhs) -> bool:
+        if isinstance(rhs, CompilationRequestIdHandle):
+            return self.cache_key == rhs.cache_key
+        elif isinstance(rhs, CompilationRequest):
+            return self.cache_key == rhs.get_cache_key()
```
fantix, line 320: trivial: might be faster?

```suggestion
        ty = type(rhs)
        if ty is CompilationRequestIdHandle:
            return self.cache_key == rhs.cache_key
        elif ty is CompilationRequest:
            return self.cache_key == rhs.get_cache_key()
```

edb/server/tenant.py
```diff
@@ -2100,17 +2095,27 @@ async def evict_query_cache(
...
-        keys: Optional[list[str]],
+        to_add: Optional[list[str]],
+        to_invalidate: Optional[list[str]],
     ) -> None:
         if not self.is_db_ready(dbname):
             return
 
+        if to_invalidate:
+            if db := self.maybe_get_db(dbname=dbname):
+                # print("invalidating", to_invalidate)
```
fantix, line 2106:
```suggestion
```

fantix (approved):
Awesome!

fantix:
> the query is being evicted because it hasn't been used in a long time, so the odds of it showing up in the next 1s are pretty low

Agree :+1:

Issue #8516, opened 2025-03-28
Implement support for Ollama and HuggingFace TGI


anbuzin:
Those two API's ([Ollama](https://github.com/ollama/ollama/blob/main/docs/api.md), [HF](https://huggingface.co/docs/text-generation-inference/en/index)) are not compatible with OpenAI or Anthropic. Technically you _could_ use them via a proxy like LiteLLM, but there's no reason not to support them directly.

dnwpark:
Ollama support done for now.
HF support probably needs to come after #8515

PR #8646, opened 2025-04-30
Add Ollama support.

dnwpark:
related #8516

Adds support for Ollama indexes, rags queries, and embeddings queries.

Supports the following models:
- `llama3.2`
- `llama3.3`
- `nomic-embed-text`
- `bge-m3`

Note: Adding more flexible models will require some re-thinking of how we deal with models in the schema.


elprans (approved):
This is great!

PR #8650, opened 2025-04-30
Try to optimize away a per-object check in access policies

msullivan:
We want postgres to not be able to decide that a type rewrite CTE is
empty, because that causes it to do optimizations that break our use
of `assert_exists` for required links.

In order to prevent postgres from determining that the CTE for a type
rewrite was empty in some cases (when it has something like `global
foo` in the filter and it is false), we were injecting an extra
`.id ?= <uuid>{}` condition to each access policy filter.
This is always false but postgres doesn't know that, so it defeats
that optimization.

It also defeats the use of indexes, which is bad. Add an analysis of
the access policy query to determine whether it is safe to omit that
check (by making sure that the object itself is always referenced).

This is somewhat dodgy, though.

### DIFF
--- a/edb/edgeql/compiler/policies.py
+++ b/edb/edgeql/compiler/policies.py
@@ -228,12 +228,30 @@ def get_rewrite_filter(
         )
         filter_expr = astutils.extend_binop(filter_expr, deny_expr)
 
+    # We compile the expression again so we can do an IR based
+    # analysis on it below.
+    with ctx.new() as dctx:
+        # HACK: to prevent filter_ir from being warning fenced
+        dctx.allow_factoring()
+        dctx.expr_exposed = context.Exposure.UNEXPOSED
+        filter_ir = dispatch.compile(filter_expr, ctx=dctx)
+    filter_expr = ctx.create_anchor(filter_ir)
+
     # This is a bad hack, but add an always false condition that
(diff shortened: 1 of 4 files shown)


edb/edgeql/compiler/policies.py
```diff
@@ -248,6 +266,100 @@ def get_rewrite_filter(
...
+    We track the invertedness and invert the AND and OR behavior when
+    underneath a NOT, kind of for fun.
+    """
+    if isinstance(ir, irast.Set):
+        if ir is ref:
+            return True
+        ir = ir.expr
+
+    match ir:
+        case irast.SelectStmt(result=result):
```
elprans, line 299: Is it guaranteed that we would ever only see `select` in this context and not anything else?

msullivan: DML isn't allowed in access policies.

elprans: I'd probably still add an assertion

elprans (approved):

edb/edgeql/compiler/policies.py
msullivan: We return false for the fall-through case, though, so I feel fine about it

PR #8667, opened 2025-05-05
Add proper remote compiler protocol versioning

fantix:
This is currently just a placeholder, but allows future implementation
to hook into old server/client with non-hacky compatibility support.

This reverts the compat hacks added in #8621


msullivan (approved):

msullivan:
Do we backport this? Yes, right?

msullivan:
> > Do we backport this? Yes, right?
> 
> That question caught me for a second. Yes, I suppose we should backport for the next 6.x release, but not the 6.x or 5.x Cloud patches

Right.

PR #8672, opened 2025-05-06
Fix schema type inconsistencies with collection aliases

dnwpark:
Related #8691

Aliases and computed globals will create any new types in the schema and marked them as `from_alias`. They do not store the existing types that they use in any existing fields.

This produces the following inconsistencies.

1. `from_alias` types are added to the `new_types` of a delta, and so the server dbview would not update its backend ids

For example:
```
create global foo := [(f := 1)];
create global bar -> array<tuple<f: int64>>;
set global bar := [(f := 1)];
select global bar;
```

The first global creates the types `tuple<f: int64>` and `array<tuple<f: int64>>`. But since these are marked as `from_alias`, they are not added to `new_types` in `TypeCommand._create_begin`.

Then setting then selecting `bar`, the backend oid is unavailable when attempting to recode globals, resulting in an ISE.

2. Types are not always properly tidied up after a non-computed global has its type altered, or is deleted.

For example, after:
```
create global foo -> tuple<hello: int64>;
drop global foo;
```

The query `select schema::Type {*} filter contains(.name, 'hello');` will return a leftover type.

3. Types can be deleted while still being "used" by an alias expression.

For example, after:
```
create type Bar { create property x -> tuple<hello: std::int64>; };
create global foo := (hello := 1);
drop type Bar;
```

The query `select schema::Type {*} filter contains(.name, 'hello');` will return nothing.

These issues were solved by:
1. No longer setting `from_alias` for new collection types from alias expressions (other than the alias expr itself)
2. Ensuring that unused types are deleted after altering or deleting non-computed globals
3. Storing types from an alias expression that already exist in a new `existing_types` field to count them as "used"

### DIFF
--- a/edb/schema/expraliases.py
+++ b/edb/schema/expraliases.py
@@ -64,10 +64,6 @@ class Alias(
         so.ObjectSet[s_types.Type],
         default=so.DEFAULT_CONSTRUCTOR,
     )
-    existing_types = so.SchemaField(
-        so.ObjectSet[s_types.Type],
-        default=so.DEFAULT_CONSTRUCTOR,
-    )
 
 
 class AliasCommandContext(
@@ -222,7 +218,6 @@ class AliasLikeCommand(
         s_types.TypeShell[s_types.Type],
         s_expr.Expression,
         set[so.ObjectShell[s_types.Type]],
-        set[so.ObjectShell[s_types.Type]],
     ]:
         pschema = schema
 
@@ -247,7 +242,7 @@ class AliasLikeCommand(
 
         is_global = (self.get_schema_metaclass().
                      get_schema_class_displayname() == 'global')
-        cmd, type_shell, created_types, existing_types = _create_alias_types(
+        cmd, type_shell, created_types = _create_alias_types(
             expr=expr,
             classname=classname,
             schema=schema,
@@ -257,7 +252,7 @@ class AliasLikeCommand(
         if drop_old_types_cmd:
             cmd.prepend(drop_old_types_cmd)
 
-        return cmd, type_shell, expr, created_types, existing_types
+        return cmd, type_shell, expr, created_types
 
 
 class AliasCommand(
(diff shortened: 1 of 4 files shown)


dnwpark:
Some of the new tests are quite slow. Could use some advice on making them faster?

msullivan:
Probably we should not backport this, btw

edb/schema/expraliases.py
```diff
@@ -64,6 +64,10 @@ class Alias(
         so.ObjectSet[s_types.Type],
         default=so.DEFAULT_CONSTRUCTOR,
     )
+    existing_types = so.SchemaField(
+        so.ObjectSet[s_types.Type],
+        default=so.DEFAULT_CONSTRUCTOR,
+    )
```
msullivan, line 70: Hmmmm.... This seems dodgy, since it makes our schema representation order-dependent in a way it shouldn't be, I think.

msullivan:
Tagging copilot as an experiment

tests/test_edgeql_expr_aliases.py
```diff
@@ -1304,6 +1304,437 @@ async def test_edgeql_aliases_schema_types_03(self):
             []
         )
 
+    async def _check_ddl_type_changes(self, ddls_and_expected_types):
```
msullivan, line 1307: Can you test:
 * Create non-alias tuple reference
 * Create alias tuple reference
 * Drop non-alias tuple reference
 * Drop alias tuple reference
 * Type should not exist

I *think* this case is not handled, but I could be wrong?

edb/schema/expraliases.py
dnwpark: I've removed this field and now just store any referenced collection types in `created_types`. This *should* just work.

dnwpark:
@msullivan this should be ready for another look

edb/schema/globals.py
```diff
@@ -619,3 +623,17 @@ class DeleteGlobal(
...
+        self,
+        schema: s_schema.Schema,
+        context: sd.CommandContext,
+    ) -> s_schema.Schema:
+        schema = super()._delete_begin(schema, context)
+        scls = self.scls
+        if not self._is_computable(scls, schema):
+            target = scls.get_target(schema)
+            if op := target.as_type_delete_if_unused(schema):
+                self.add_caused(op)
```
msullivan, line 637: Why did we need to add this? Were we not deleting target types before?

msullivan (approved):

edb/schema/globals.py
dnwpark: Yeah, there were a few places where things *should* be deleted but seemed to not be.

PR #8680, opened 2025-05-08
Add check for redundancy in Index validation.

dnwpark:
Indexes are automatically generated for links and exclusive constraints.


dnwpark:
Note: This will only check CreateIndex. We can also check AlterIndex, constraints and links as well. Is this worth doing?

msullivan:
> Note: This will only check CreateIndex. We can also check AlterIndex, constraints and links as well. Is this worth doing?

The thing we really want to check on is not running `CREATE INDEX` itself, but doing a *migration* to something with redundant indexes. That code path will probably hit CreateIndex though, so maybe this covers it.

We want to test it though

PR #8682, opened 2025-05-08
In LSP, compile on save instead of change

scotttrinh:
The `DID_CHANGE` callback is called for every edit to the document, so it would try to re-parse on every keystroke. We plan on debouncing this as the full fix for the DX issue, but for now, we can switch to re-parsing on save.

### DIFF
--- a/edb/language_server/main.py
+++ b/edb/language_server/main.py
@@ -75,7 +75,7 @@ def init(options_json: str | None) -> ls_server.GelLanguageServer:
     def text_document_did_open(params: lsp_types.DidOpenTextDocumentParams):
         ls_server.document_updated(ls, params.text_document.uri)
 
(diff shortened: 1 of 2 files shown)


aljazerzen (approved):
Looks good. We probably would want to parse the file on every DID_CHANGE, but compile it just on DID_SAVE.

I'll try it out and see how it feels.

In any case, we can merge this and add onto it later.

aljazerzen:
Yeah, syntax errors should how even before saving. I think our parser is fast enough for that.

https://github.com/user-attachments/assets/3e793c92-0866-46fa-9eed-91dc208c2145

scotttrinh:
> Yeah, syntax errors should how even before saving. I think our parser is fast enough for that.

Can we try this on a big, realistic schema? I feel like even syntax/parser errors should be debounced to just be more resource-friendly in general, but then once we do that, we might as well also compile?

But, right, we can address this later more holistically.

PR #8685, opened 2025-05-08
Fix having unnecessary constraint triggers after restore

### LINKED ISSUE #8683: After restoring from a dump, unnecessary triggers are enabled
We create triggers unconditionally for all exclusive constraints, but disable them whenever there aren't any children or parents.

But during restore, we disable all triggers and then reenable them, so they get turned back on.

This slows down inserts

msullivan:
Don't create constraint triggers that we then always disable.
The triggers get enabled while doing a restore, which has real perf
impacts.

In our current implementation, it is totally safe to just stop creating
them. Since #7593, we now no longer ever enable the disabled triggers,
because we always need to recompile them to include new descendants or some
such.

Fixes #8683.

I also added an `ADMINISTER remove_pointless_triggers()` command that
will remove the unneeded triggers from an existing database.

We could also skip that and do it unconditionally as part of a patch,
though.

### DIFF
--- a/edb/buildmeta.py
+++ b/edb/buildmeta.py
@@ -57,7 +57,7 @@ from edb.common import verutils
 # The merge conflict there is a nice reminder that you probably need
 # to write a patch in edb/pgsql/patches.py, and then you should preserve
 # the old value.
-EDGEDB_CATALOG_VERSION = 2025_04_30_00_00
+EDGEDB_CATALOG_VERSION = 2025_05_09_00_00
 EDGEDB_MAJOR_VERSION = 7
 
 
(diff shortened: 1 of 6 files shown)


elprans:
> We could also skip that and do it unconditionally as part of a patch, though.

I think I would prefer that.  `ADMINISTER foo` is technically a valid public API and we ideally shouldn't be populating it with random fixlets.  Or maybe we can fold this into `repair_schema`?

msullivan:
> > We could also skip that and do it unconditionally as part of a patch, though.
> 
> I think I would prefer that. `ADMINISTER foo` is technically a valid public API and we ideally shouldn't be populating it with random fixlets. Or maybe we can fold this into `repair_schema`?

Alright, I've prefixed the ADMINISTER command with an underscore and restricted it to testmode, which lets us test that it doesn't break constraints when run and makes the code available for use in backport and IPU.

The other question is whether it is worth having a new dedicated code path for `fixup`; we could also just call `update_trigger_ops()`, which would recreate all of the necessary triggers instead of just deleting the unnecessary ones.

PR #8688, opened 2025-05-09
Gel-ls: go to definition for schema files

aljazerzen:
Based on #7704
Ref #7411

The implementation is a bit lacking, because it does not use the
existing code we have for name resolution, but instead just finds
ObjectRef under the cursor and looks up that name in the schema.

It does also look for enclosing `module` declarations, so it can resolve
unqualified names, but that's it. I hope our actual name resolution is
not any more complicated than this. We don't have a way to specify mod
aliases, right? We do have some custom handling for `std::`, but we
don't support definitions in `std` (yet) anyway.

I did make sure that definition lookup does *not* recompile schema,
since that is slow on large schema. It even caches parsed SDL, so
pressing `ctrl` and hovering over many identifiers does not overwhelm
the server.

To make this work, I had to improve how we set spans on AST, so we never
propagate spans from parents into children, but instead use spans
produced by the parser.


aljazerzen:
https://github.com/user-attachments/assets/c1e7bbe5-8325-45bb-a043-512acb0f2ca0

edb/language_server/definition.py
```diff
@@ -114,18 +114,83 @@ def _get_definition_in_ql(
...
+    # find the span in ql ast
+    node_path = edb_span.find_by_source_position(ls.state.schema_sdl, position)
+    if not node_path:
+        return None
+
+    # ls.show_message_log(f"found node: {dump_to_str(node_path[0])}")
+
+    # only resolve ObjectRefs
+    if not isinstance(node_path[0], qlast.ObjectRef):
+        return None
```
msullivan, line 152: Make sure pointers are on a TODO list somewhere

msullivan (approved):

edb/language_server/definition.py
msullivan: I guess more broadly I mean, "understand expressions nested in the schema" better?

aljazerzen: I don't think we have any place in schema where we would allow pointer references that is not a `s_expr.Expression`.

And "understand expressions nested in the schema" is on the list, yes

PR #8689, opened 2025-05-09
Gel-ls: parse on change, but compile only on save

aljazerzen:
Follow up for #8682

That PR made is so we don't parse or compile on every document change,
but only when document is saved. This was needed because large schemas
resulted in a lot of lag of diagnostics.

This PR does parsing on every change but recompiles only on saves.
This gives syntactical diagnostics during typing. This is feasible
because they are cheap performance wise. Compilation of EdgeQL or schema
is done only on save.

But still, there is something bad going on with schema compilation.
Name errors are quick to show (<1 sec), but successful compilations are
slow. We should investigate.

### DIFF
--- a/edb/language_server/completion.py
+++ b/edb/language_server/completion.py
@@ -73,7 +73,7 @@ def _get_completion_in_ql(
     # compile the stmt that now contains the qlast.Cursor,
     # which should halt compilation, when it gets to the cursor
     try:
-        diagnostics, _ir_stmts = ls_server.compile(ls, document, [ql_stmt])
+        diagnostics, _ir_stmts = ls_server.compile_ql(ls, document, [ql_stmt])
     except qlcompiler.expr.IdentCompletionException as e:
         return [
             lsp_types.CompletionItem(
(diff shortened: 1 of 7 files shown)


aljazerzen:
Uploading vscode-gel-ls-parse-on-change.mp4…

scotttrinh:
My _only_ concern here is that I think it might be surprising to folks that there are some errors that will only show up after save. I haven't seen another LSP integration that has this specific behavior, typically I think most LSPs will either run constantly (I imagine they're debouncing here?) or they only run on save.

I agree that in the spirit of show more help sooner, it'd be nice. Maybe we use different levels for parser errors vs compilation errors?

msullivan (approved):
LG from a technical perspective.
We will want to think about what Scott is saying though

aljazerzen:
1) this is exactly how rust-analyzer runs. When I first started using it, I've noticed that this is happening, but I didn't (and still don't) find it confusing. Although my experience here is less applicable, since I probably understand difference between syntactic and semantic errors better than the average dev.

2) with this PR, if you have semantic errors and continue editing and you introduce more syntactic errors, *both* new syntactic errors and semantic errors from last compilation are shown.

So my opinion is that this is better than no syntactic errors on DID_CHANGE and also that this is better than debounced compilation on DID_CHANGE.

But Yury did mention another thing: vscode has option to save the file on every keystroke. This might still be flooding the server, so we'd have to do debouncing anyway.

PR #8692, opened 2025-05-09
Implement sys::approximate_count()

### LINKED ISSUE #4164: Add std::approximate_count()

msullivan:
It takes a schema::ObjectType, so it is called like
`sys::approximate_count(introspect Card)`.
This is because the implementation works by querying the `reltuples`
field of `pg_class`, and it really only makes sense for getting a
full table. Having it take `anytype` and special casing would
create weird discontinuities where the simple case worked like this
and complex cases have a correctness or performance cliff.

Maybe we want to name it something different?

By default it includes all descendant types too; the ignore_subtypes
parameter changes that.

Fixes #4164.

### DIFF
--- a/edb/lib/sys.edgeql
+++ b/edb/lib/sys.edgeql
@@ -405,3 +405,15 @@ sys::__pg_or(a: OPTIONAL std::bool, b: OPTIONAL std::bool) -> std::bool
         SELECT a OR b;
     $$;
 };
+
+
+CREATE FUNCTION
+sys::approximate_count(
+    type: schema::ObjectType,
+    NAMED ONLY ignore_subtypes: std::bool=false,
+) -> int64
+{
+    SET volatility := 'Stable';
+    USING SQL FUNCTION 'edgedb.approximate_count';
+    set impl_is_strict := false;
+};
--- a/docs/reference/stdlib/sys.rst
+++ b/docs/reference/stdlib/sys.rst
@@ -11,6 +11,9 @@ System
 .. list-table::
     :class: funcoptable
 
+    * - :eql:func:`sys::approximate_count`
+      - :eql:func-desc:`sys::approximate_count`
+
     * - :eql:func:`sys::get_version`
       - :eql:func-desc:`sys::get_version`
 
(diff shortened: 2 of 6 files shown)


1st1:
Can you add some basic docs? Also, this won't also make with access policies, right? Given that we (our users) want to potentially to use access policies to restrict what LLMs can see and whatnot, maybe exposing this function is a bad idea.  Also..we won't be using it for our UI probably (access policies need to be taken into account for rendering pagination).

Anyway, I'm not against merging this (I also was pinging you guys to add it from time to time), just brain-dumping my current thought here.

edb/lib/sys.edgeql
```diff
@@ -405,3 +405,15 @@ sys::__pg_or(a: OPTIONAL std::bool, b: OPTIONAL std::bool) -> std::bool
         SELECT a OR b;
     $$;
 };
+
+
+CREATE FUNCTION
+sys::approximate_count(
```
elprans, line 411: Arguably this could be in `std::` instead?

msullivan:
> Given that we (our users) want to potentially to use access policies to restrict what LLMs can see and whatnot, maybe exposing this function is a bad idea (as it could "leak" some knowledge). 

Yeah it will go on the list of functions that probably should be restricted by RBAC.

> Arguably this could be in `std::` instead?

The above being part of why I like it in `sys` over `std`.
Thought I don't feel that strongly.

aljazerzen (approved):
> reltuples float4
> Number of live rows in the table. This is only an estimate used by the planner. It is updated by [VACUUM](https://www.postgresql.org/docs/current/sql-vacuum.html), [ANALYZE](https://www.postgresql.org/docs/current/sql-analyze.html), and a few DDL commands such as [CREATE INDEX](https://www.postgresql.org/docs/current/sql-createindex.html). If the table has never yet been vacuumed or analyzed, reltuples contains -1 indicating that the row count is unknown.

We probably do have auto VACUUM enabled, right?

msullivan:
Oh sigh, I forgot to bump the catalog

Issue #8691, opened 2025-05-09
Schema type inconsistencies after create/alter/drop aliases and computed globals.


dnwpark:
1. `from_alias` types are added to the `new_types` of a delta, and so the server dbview would not update its backend ids

For example:
```
create global foo := [(f := 1)];
create global bar -> array<tuple<f: int64>>;
set global bar := [(f := 1)];
select global bar;
```

The first global creates the types `tuple<f: int64>` and `array<tuple<f: int64>>`. But since these are marked as `from_alias`, they are not added to `new_types` in `TypeCommand._create_begin`.

Then setting then selecting `bar`, the backend oid is unavailable when attempting to recode globals, resulting in an ISE.

2. Types are not always properly tidied up after a non-computed global has its type altered, or is deleted.

For example, after:
```
create global foo -> tuple<hello: int64>;
drop global foo;
```

The query `select schema::Type {*} filter contains(.name, 'hello');` will return a leftover type.

3. Types can be deleted while still being "used" by an alias expression.

For example, after:
```
create type Bar { create property x -> tuple<hello: std::int64>; };
create global foo := (hello := 1);
drop type Bar;
```

The query `select schema::Type {*} filter contains(.name, 'hello');` will return nothing.

PR #8693, opened 2025-05-09
Add tests to check schema type consistency after ddl with aliases and globals.

dnwpark:
Related #8691


tests/test_edgeql_expr_aliases.py
```diff
@@ -1304,6 +1304,446 @@ async def test_edgeql_aliases_schema_types_03(self):
             []
         )
 
+    async def _check_ddl_type_changes(self, ddls_and_expected_types):
```
msullivan, line 1307: Let's put these tests in test_edgeql_ddl, or somewhere else.

We need to move away from adding lots of DDL-doing tests to QueryTestCase test suites, since it prevents them from parallelizing.

tests/test_edgeql_expr_aliases.py
```diff
@@ -1304,6 +1304,446 @@ async def test_edgeql_aliases_schema_types_03(self):
...
+                f'''
+                    select schema::Type {{
+                        name,
+                        from_alias,
+                        type_name := (.__type__.name),
+                    }}
+                    filter .id not in {{{','.join(
+                        f'<uuid>"{str(existing_type)}"'
+                        for existing_type in existing_types
+                    )}}}
```
msullivan, line 1326: Can you do this by passing in a parameter instead? It doesn't really matter but is cleaner

msullivan (approved):

PR #8694, opened 2025-05-09
Make the patch system support making tweaks to user databases

msullivan:
Currently there is the edgeql+user_ext patch mode, which makes some
change to a user database when a specific extension is installed.

Generalize it to also support an edgeql+user mode, which doesn't have
an extension.

The main goal here is to be hook-up fixup functions in there, like for https://github.com/geldata/gel/pull/8685

### DIFF
--- a/edb/server/bootstrap.py
+++ b/edb/server/bootstrap.py
@@ -908,22 +908,23 @@ def prepare_patch(
 
         metadata_user_schema = reflschema
 
-    elif kind.startswith('edgeql+user_ext'):
+    elif kind.startswith('edgeql+user'):
         assert '+schema' not in kind
 
         # There isn't anything to do on the system database for
-        # userext updates.
+        # user updates.
         if user_schema is None:
             return (update,), (), dict(is_user_update=True)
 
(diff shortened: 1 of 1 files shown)


elprans (approved):

aljazerzen:
> The main goal here is to be hook-up fixup functions in there, like for

... like for what?

msullivan:
> > The main goal here is to be hook-up fixup functions in there, like for
> 
> ... like for what?

Sigh, had a PR reference at the start of a line that became a comment.
Like for https://github.com/geldata/gel/pull/8685

aljazerzen (approved):

edb/server/bootstrap.py
```diff
@@ -908,22 +908,23 @@ def prepare_patch(
 
         metadata_user_schema = reflschema
 
-    elif kind.startswith('edgeql+user_ext'):
+    elif kind.startswith('edgeql+user'):
         assert '+schema' not in kind
```
aljazerzen, line 912: Maybe limit this a bit, this currently also supports `edgeql+user_etx`, which we probably want to catch before releasing.

PR #8690, opened 2025-05-09
Remove unnecessary drop and create constraints.

dnwpark:
close #8670

Issue #8670, opened 2025-05-05
We drop and recreate the constraint for id whenever a table is rebased


msullivan:
If we run
```
create type X0;
create type X1;
```
and then
```
alter type X1 extending X0;
```

we will do:
```
    ALTER TABLE edgedbpub."35dd32e1-2a0b-11f0-a969-edfb37ea298d" DROP CONSTRAINT "35de7483-2a0b-11f0-adff-35b4e065e437;schemaconstr";
    
    ALTER TABLE edgedbpub."35dd32e1-2a0b-11f0-a969-edfb37ea298d" ADD CONSTRAINT "35de7483-2a0b-11f0-adff-35b4e065e437;schemaconstr" UNIQUE (id);
```
which is pointless and probably expensive.

msullivan:
I expect that we will also have the same problem with `name` in this example:


```
create abstract type N {
    create property name -> str { create constraint exclusive; };
};
create type X0 extending N;
create type X1 extending N;
```
and then
```
alter type X1 extending X0;
```

It would be best to fix this also, but it isn't as important.

PR #8700, opened 2025-05-12
Fix typo in constraint comparison and adding more comments.

dnwpark:
Fix to #8690
Related #8670


aljazerzen (approved):

msullivan:
Honestly this still doesn't make sense to me; are we really confident that this case never requries recreating the triggers?

PR #8703, opened 2025-05-13
Gel-ls: show hints produced in all stages in diagnostics

aljazerzen:
We had code that was showing hints from some parts of the compiler.
This PR adds tests for that and fixes error conversion into diagnostics
for CST to AST mapper.


scotttrinh (approved):
I don't have any technical feedback here, but looking at the tests, this is great!

msullivan (approved):

PR #8706, opened 2025-05-13
Pin to click 8.1

msullivan:
Something in 8.2.0 changed handling of case sensitivity.


msullivan:
Looks like there is already an issue filed on click: https://github.com/pallets/click/issues/2911

There wasn't anything in the changelog announcing a breaking change about this.

fantix (approved):

PR #8713, opened 2025-05-15
Disable DML on instered objects that needed for SQL

aljazerzen:
Followup for #8582

For DML, I needed updates of objects that were inserted in the same
query. To be exact, I needed only *updates of their links*, which
translate to update of the link table only.

But my fix also affected EdgeQL DML, so this PR disables that.

We could keep this feature, but it would only work when updates really
are on links and multi properties only. Which would be a weird design
choice, so I would prefer not to have this feature in the first place.

### DIFF
--- a/edb/pgsql/compiler/__init__.py
+++ b/edb/pgsql/compiler/__init__.py
@@ -88,6 +88,7 @@ def compile_ir_to_sql_tree(
     versioned_stdlib: bool = True,
     # HACK?
     versioned_singleton: bool = False,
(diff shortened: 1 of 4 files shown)


msullivan (approved):
Great!
Could we make the test assertRaises instead of xfail also?

aljazerzen:
We cannot, because we don't actually detect "update of inserted objects", but just don't update it.

So it fails with "data shape differs" after selecting the supposedly updated element.

PR #8723, opened 2025-05-20
Add json implementation for to_bytes function.

dnwpark:
The query `select to_bytes(to_json("[1]"))` produces `{b'[1]'}`.


tests/test_edgeql_functions.py
```diff
@@ -3631,6 +3631,24 @@ async def test_edgeql_functions_string_bytes_conversion_error(self):
                 ''',
             )
 
+    async def test_edgeql_functions_json_bytes_conversion(self):
+        string = json.dumps({"a": [1, 2, 3], "b": "foo"})
+
+        await self.assert_query_result(
```
scotttrinh, line 3637: Love a good roundtrip test case 👏 . Might be worth adding some weird stuff like utf-8 characters maybe?

scotttrinh, line 3637: Love a good roundtrip test case 👏 . Might be worth adding some weird stuff like utf-8 characters maybe?

scotttrinh (approved):
Consider this a non-technical API-only review, but looks great to me!

tests/test_edgeql_functions.py
```diff
@@ -3631,6 +3631,31 @@ async def test_edgeql_functions_string_bytes_conversion_error(self):
                 ''',
             )
 
+    async def test_edgeql_functions_json_bytes_conversion(self):
+        strings = [
+            json.dumps({"a": [1, 2, 3], "b": "foo"}),
+            json.dumps(
+                {"數": [1, 2, 3], "言": "你好世界！"},
+                ensure_ascii=False
```
msullivan, line 3639: What happens if this is not set?

msullivan (approved):

PR #8724, opened 2025-05-21
Add http schedule_request with json body.

### LINKED ISSUE #8697: Simplify scheduling HTTP JSON request
Based on a discussion regarding an upcoming blog post and common needs when using `net::http`, there is a clear use case for simpler handling of JSON data when making HTTP requests.

We should consider adding:

 1. A built-in function, e.g., `to_bytes(<json>)`, to easily convert JSON values to bytes. This would be useful for preparing JSON bodies for HTTP requests or other byte-oriented operations.
 2. An overload or sugar function for net::http::request (e.g., `net::http::schedule_json_request`) that simplifies sending JSON bodies. This function could automatically handle setting the `Content-Type: application/json` header and converting the provided JSON value to bytes using the new to_bytes function internally.

This would reduce boilerplate for users making JSON-based API calls using the `net::http` module.

dnwpark:
close #8697

### DIFF
--- a/edb/lib/net.edgeql
+++ b/edb/lib/net.edgeql
@@ -99,3 +99,36 @@ std::net::http::schedule_request(
         }
     ));
 };
+
+CREATE FUNCTION
+std::net::http::schedule_request(
+    url: str,
+    named only body: std::json,
+    named only headers: optional std::array<
+        std::tuple<
+            name: std::str,
+            value: std::str
+        >
+    > = {},
+    named only method: std::net::http::Method = std::net::http::Method.`GET`,
+) -> std::net::http::ScheduledRequest
+{
+    SET is_inlined := true;
+    USING ((
+        INSERT std::net::http::ScheduledRequest {
+            url := url,
+            method := method,
+            headers := headers ++ [
+                (
+                    name := "Content-Type",
+                    value := "application/json",
+                ),
+            ],
+            body := to_bytes(body),
+
+            created_at := std::datetime_of_statement(),
+            updated_at := std::datetime_of_statement(),
+            state := std::net::RequestState.Pending,
+        }
+    ));
+};
--- a/edb/lib/net.edgeql
+++ b/edb/lib/net.edgeql
@@ -117,7 +117,7 @@ std::net::http::schedule_request(
     USING ((
         WITH 
             has_content_type := any(
-                std::array_unpack(headers).0 = 'Content-Type'
+                std::str_lower(std::array_unpack(headers).0) = 'content-type'
             ),
             actual_headers := (
                 IF has_content_type
--- a/edb/lib/net.edgeql
+++ b/edb/lib/net.edgeql
@@ -115,20 +115,22 @@ std::net::http::schedule_request(
 {
     SET is_inlined := true;
     USING ((
-        INSERT std::net::http::ScheduledRequest {
-            url := url,
-            method := method,
-            headers := headers ++ [
-                (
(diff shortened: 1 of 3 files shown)


dnwpark:
This appends `(name := "Content-Type", value := "application/json")` to the headers. What happens if there is a duplicate?

scotttrinh:
> What happens if there is a duplicate?

I think that's determined by the underlying HTTP layer, so maybe @mmastrac can shed some light here? IIRC, typically duplicate headers are handled by taking the last one in most cases. I wonder if that means we should prepend this header to allow people to overwrite the Content-type with a more specific JSON MIME type, which is somewhat common.

dnwpark:
Did some digging on this.

According to the http spec:
- [sec 5.2](https://www.rfc-editor.org/rfc/rfc9110.html#name-field-lines-and-combined-fi): multiple fields are concatenated with a `,` delimiter
- [sec 8.3](https://www.rfc-editor.org/rfc/rfc9110.html#field.content-type): `content-type` is a singleton field. some apps will try to take the last value, while others will produce an error

Relying on taking the last value is not a good solution imo. We could either try to flag for duplicates either in `schedule_request` or possibly in `net_worker.handle_request`.

dnwpark:
@scotttrinh

I think having an explicit additional parameter is going to be kinda clunky. I've changed it so that it only adds the header if it's not already present.

edb/lib/net.edgeql
```diff
@@ -99,3 +99,38 @@ std::net::http::schedule_request(
...
+        >
+    > = {},
+    named only method: std::net::http::Method = std::net::http::Method.`GET`,
+) -> std::net::http::ScheduledRequest
+{
+    SET is_inlined := true;
+    USING ((
+        WITH 
+            has_content_type := any(
+                std::array_unpack(headers).0 = 'Content-Type'
```
scotttrinh: I think we need to compare this in a case insensitive way.

dnwpark: @scotttrinh Fixed, take another look when you have time!

edb/lib/net.edgeql
```diff
@@ -99,3 +99,38 @@ std::net::http::schedule_request(
...
+    > = {},
+    named only method: std::net::http::Method = std::net::http::Method.`GET`,
+) -> std::net::http::ScheduledRequest
+{
+    SET is_inlined := true;
+    USING ((
+        WITH 
+            has_content_type := any(
+                std::str_lower(std::array_unpack(headers).0) = 'content-type'
+            ),
```
scotttrinh, line 121: OK, apologies for the golfing, I just have a feeling like we're rarely going to have a chance to revisit this, so pays to put in a little effort now. 😅 

I wonder if this is better performance wise:

```suggestion
            has_content_type := (
                "content-type" in std::str_lower(std::array_unpack(headers).0)
            ),
```

@msullivan or @elprans have any insight here?

Not a blocker!

scotttrinh (approved):

edb/lib/net.edgeql
dnwpark: In postgres, `any =` is equivalent to `in`, so hopefully this ends up roughly similar here.

scotttrinh: Yeah, I get that they're logically equivalent, I just wasn't sure if there was a performance reason to prefer `in` as I've found that `any` and `all` functions have been sources of performance issues in my own usage when combining multiple conditions for a filter. But definitely not a blocker!

PR #8726, opened 2025-05-22
Update the docs for backup/restore CLI.

vpetrovykh:


docs/reference/using/cli/gel_instance/gel_instance_backup.rst
```diff
@@ -0,0 +1,65 @@
+.. _ref_cli_gel_instance_backup:
+
+
+===================
+gel instance backup
+===================
+
+Backup a |Gel| instance to disk.
```
scotttrinh, line 8: As far as I know, these commands only work for local project managed (and Docker compose managed) and Gel Cloud instances, right? We should probably say something about that either here or in the description.

docs/reference/using/cli/gel_instance/gel_instance_backup.rst
```diff
@@ -0,0 +1,65 @@
...
+.. cli:synopsis::
+
+    gel instance backup [<options>]
+
+
+Description
+===========
+
+The :gelcmd:`instance backup` command creates a backup of the specified
+insatnce. This backup can be restored using the
```
scotttrinh, line 19:
```suggestion
instance. This backup can be restored using the
```

docs/reference/using/cli/gel_instance/gel_instance_backup.rst
```diff
@@ -0,0 +1,65 @@
...
+:ref:`ref_cli_gel_instance_restore` command.
+
+The tool will automatically determine whether to perform a full or an
+incremental backup. The goal is to balance the backup size with the restore
+speed. So the rule of thumb is that an incremental backup will be created if
```
scotttrinh, line 24: Is it "restore speed" we're optimizing or "backup speed"?

docs/reference/using/cli/gel_instance/gel_instance_restore.rst
```diff
@@ -0,0 +1,66 @@
...
+
+:cli:synopsis:` --backup-id <backup_id>`
+    The backup ID as indicated by the
+    :ref:`ref_cli_gel_instance_listbackups`.
+
+:cli:synopsis:`--latest`
+    Restore the latest backup available for the instance.
+
+:cli:synopsis:`--source-instance <source>`
+    Name of source instance to restore the backup from.
```
scotttrinh, line 38: This only works for Cloud, right?

scotttrinh (commented):
Thanks for taking this on, it was on my to do list for this week, and keeps getting pushed. 🙏 

Reading through the docs made me wonder how easy it would be to add a `deletebackup` command, or have instructions on where to find them so users can delete them themselves.

docs/reference/using/cli/gel_instance/gel_instance_backup.rst
mmastrac: These work for locally managed (incl WSL) and cloud instances. Docker and remote aren't supported.

docs/reference/using/cli/gel_instance/gel_instance_backup.rst
mmastrac: I think we should say backup speed. We prefer to take snapshots as quickly as we can.

docs/reference/using/cli/gel_instance/gel_instance_restore.rst
mmastrac: Currently yes

Issue #8515, opened 2025-03-28
Implement model selection via a URI string for chat models


anbuzin:
Currently if you would like to use a chat model, you need to either pick one that's explicitly specified in the `ext::ai` or change your schema to use a custom model.

Proposal: let the user specify a model with a URI, e.g. `provider:model_name`.

PR #8729, opened 2025-05-22
Add ability to specify provider:model when using rags api.

dnwpark:
Related #8515

When using the rags api, you can now do `gel.ai.create_rag_client(client, model="openai:gpt-4.1")` as long as openai config is provided.

Note: This is not supported for embeddings models/queries, and is unlikely to be in the near future.

### DIFF
--- a/edb/server/protocol/ai_ext.py
+++ b/edb/server/protocol/ai_ext.py
@@ -2563,7 +2563,7 @@ async def _handle_rag_request(
         if ctx_globals is not None and not isinstance(ctx_globals, dict):
             raise TypeError('"globals" must be a JSON object')
 
-        model = body.get('model')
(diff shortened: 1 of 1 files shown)


dnwpark:
@anbuzin Is this kind of what you had in mind?

anbuzin:
@dnwpark yes, this seems exactly like what I had in mind! This unties us from having to keep a whitelist of models in the schema, correct?

anbuzin:
A couple questions:

1. How would I define a custom provider? Is there a way to enable "<custom_provider_name>:<custom_model>" somehow?
2. What kind of error gets thrown if I ask for a non-existent model?

anbuzin (approved):

PR #8736, opened 2025-05-23
Add permissions to schema.

dnwpark:
related #8177

Allows creation of custom permissions within the schema:
```
Permission foo;
```

Permissions can be used in general expressions by using the `global` keyword:
```
select global foo;
```

Currently, all such expressions evaluate to `false`.


edb/schema/schema.py
```diff
@@ -1413,7 +1420,11 @@ def getter(schema: FlatSchema, name: sn.Name) -> Optional[so.Object]:
         if obj is not so.NoDefault:
             # We do our own type check, instead of using get_by_id's, so
             # we can produce a user-facing error message.
-            if obj and type is not None and not isinstance(obj, type):
+            if (
+                obj
+                and type is not None
+                and (not skip_type_check and not isinstance(obj, type))
+            ):
```
msullivan, line 1427: We should not do this. It makes the type annotations wrong for the overloads.

It would be OK if we treated it as not being found if the type is wrong and `skip_type_check` is set, but I think that's probably not worth the hassle, and you should get rid of `skip_type_check` and handle it at the call site.

You could either pass `Object` as the type and do all the checks there, or you could catch the exception.
Probably the first is cleaner.

edb/edgeql/compiler/expr.py
```diff
@@ -549,10 +552,50 @@ def compile_UnaryOp(
...
+    )
+
+    # Check None
+    if expr_obj is None:
+        s_schema.Schema.raise_bad_reference(
+            expr_schema_name,
+            module_aliases=ctx.modaliases,
+            span=expr.span,
+            type=s_globals.Global,
+        )
```
msullivan, line 573: Document why you are doing this instead of letting `get` handle it. (I think to get `global` mentioned in the error`?

edb/edgeql/compiler/expr.py
```diff
@@ -549,10 +552,50 @@ def compile_UnaryOp(
...
+
+    # Check for non-global
+    if not isinstance(expr_obj, s_globals.Global):
+        s_schema.Schema.raise_wrong_type(
+            expr_schema_name,
+            expr_obj.__class__,
+            s_globals.Global,
+            span=expr.span,
+        )
+        raise  # for type inference
```
msullivan, line 595: I'd do a `raise AssertionError('something')`. This bare raise in a place it isn't allowed squicks me out.

msullivan: maybe it's a mypy bug

dnwpark: Yeah it's a mypy thing. I'll add a proper assertion though.

PR #8738, opened 2025-05-24
Add prompt param to discord's `get_code_url`

asiebens:
Default behaviour of Discord reprompts user for authorization of scopes every time (even if nothing changed when user initially authorized the requested scopes).

This change will set the `prompt` param to `none` in the request and will have the following effect:

- When the user is first redirected to discord, they get a prompt to authorize the scopes requested by the application
- When the user is redirected to discord and no scopes have changed, no prompt is shown to the user
- When the user is redirected to discord and the scopes have changes since the user authorized the application, the prompt is shown again to re-authorize the request


scotttrinh:
Implemented in #9111 

Thanks for the contribution!

PR #8747, opened 2025-05-29
Support creating indexes concurrently

### LINKED ISSUE #8175: Concurrent index creation
Postgres supports `CREATE INDEX CONCURRENTLY`, and it would be good if we can do something similar.

The two big hurdles are:
 - How should it interact with our migration system
 - `CREATE INDEX CONCURRENTLY` does not work in a transaction, and we always do our DDL transactionally now

msullivan:
The implementation in this PR works well and is solid, but there is plenty of room for us to argue about whether the interface is good enough.

The core idea, though, is to have an index field called `create_concurrently` that indicates that an index should be created concurrently, and then there will be a separate CLI command used to actually trigger and wait for the concurrent creation.

The crux of the complication here is that CREATE INDEX CONCURRENTLY cannot be run from inside a transaction or a pl/pgsql script, and so it can't be transactionally packaged with anything else, like updating our metadata.

The approach, then, is to have three separate components:
 * The actual DDL, generated by `migration create`, which will put the index
   into our schema but not actually create any indexes.
   The `create_concurrently` flag is exposed, as well as a new `active` flag
   that indicates whether the backing index actually exists.
 * `ADMINISTER concurrent_index_create(<uuid>)`, which takes the `id` of
    an index and runs `CREATE INDEX CONCURRENTLY IF NOT EXISTS` for it,
    and then sets `active := true` on the schema object (and populates postgres
    schema comments). 
    It does *not* do these things transactionally, though, so we could crash or
    shutdown after creating the index but before updating the schema. That's fine,
    though: once the command is run again, it will skip the CREATE INDEX then update
    metadata.

Fixes #8175

------

The big questions here are:
 * Is it acceptable to have create_concurrently as a schema flag?
   I chose that because users have a very schema-driven workflow,
   and the alternative seemed to be to force them to edit the migrations
   to add it.
 * Is outsourcing the state machine processing for actually creating
   the indexes to the CLI invoking ADMINISTER commands reasonable?
   It was quite straightforward to build.

I didn't try to make use of any of the `deferred` mechanism, because
that seems like it solving a different issue.

### DIFF
--- a/edb/server/compiler/ddl.py
+++ b/edb/server/compiler/ddl.py
@@ -1853,14 +1853,23 @@ def _get_index(
     # This is janky, and we shouldn't do it.
     arg = ql.expr.args[0]
     match arg:
-        case qlast.Constant(
-            kind=qlast.ConstantKind.STRING,
-            value=id_string,
+        case qlast.TypeCast(
+            type=qlast.TypeName(
+                maintype=qlast.ObjectRef(
+                    name='uuid',
+                    module='std' | None,
+                ),
+                subtypes=None,
+            ),
+            expr=qlast.Constant(
+                kind=qlast.ConstantKind.STRING,
+                value=id_string,
+            )
         ):
             pass
         case _:
             raise errors.QueryError(
-                f'argument to {ql.expr.func}() must be a string literal',
+                f'argument to {ql.expr.func}() must be a uuid literal',
                 span=arg.span,
             )
 
--- a/edb/server/compiler/ddl.py
+++ b/edb/server/compiler/ddl.py
@@ -1960,7 +1960,6 @@ def administer_concurrent_index_update(
         statements = subblock.commands[0].get_statements()
         _, comments = statements
 
-
         # Update the schema::Index
         eql = f'''
             update schema::Index filter .id = <uuid>"{index.id}"
--- a/edb/server/compiler/ddl.py
+++ b/edb/server/compiler/ddl.py
@@ -67,7 +67,6 @@ from edb.pgsql import delta as pg_delta
 from edb.pgsql import dbops as pg_dbops
 
 from . import dbstate
-from . import enums
 from . import compiler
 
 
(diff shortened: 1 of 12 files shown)


msullivan:
@elprans alright, after some thought, I think I want to go ahead with this approach. I'd like to merge this PR so I can have @aljazerzen more easily start on the CLI side.

(I do think I might try to get rid of `ADMINISTER concurrent_index_update()` in a follow-up, though. It really shouldn't be necessary.)

aljazerzen:
> Is it acceptable to have create_concurrently as a schema flag?

Another way of doing it would be a CLI flag on `migration apply`, which would set a protocol flag, saying "don't activate any indexes created in this migration in the main transaction".

1st1:
Haven't we decided to call it `build_concurrently`?

edb/server/compiler/ddl.py
```diff
@@ -1839,6 +1840,108 @@ def administer_prepare_upgrade(
...
+        case qlast.Constant(
+            kind=qlast.ConstantKind.STRING,
+            value=id_string,
+        ):
+            pass
+        case _:
+            raise errors.QueryError(
+                f'argument to {ql.expr.func}() must be a string literal',
+                span=arg.span,
+            )
```
aljazerzen: String literal? How hard would it be to accept a `<uuid>$0`?

Does not really matter

aljazerzen (commented):
The user-facing messages should make a distinction between index "creation" and index "activation".

Because we already have `CREATE INDEX` as a command, it might be confusing if CLI prints "Creating index concurrently ...". So instead I'll print "Activating index ...".

Is "activating" a good word?

edb/server/compiler/ddl.py
msullivan: I can make it support `<uuid>"string literal"` but supporting `$0` is kind of hard, since arguments never go to the compiler.

msullivan:
I think we should say "building" rather than activating, if we make the field `build_concurrently`. Then I will also change the function name to `concurrent_index_create`

edb/server/compiler/ddl.py
msullivan: I've made it support and require `<uuid>"..."` instead. Still kind of a bummer but less of a hack

PR #8754, opened 2025-05-31
Do better multiplicity inference in nested FOR expressions

### LINKED ISSUE #8748: Incorrect multiplicity inference of `for` expressions
The following

```edgeql
FOR movie IN default::Movie SELECT movie {
  title,
  actors := (FOR person IN movie.actors SELECT person),
};
```

currently produces this:

```
error: QueryError: possibly not a distinct set returned by an expression for a computed link 'actors'
  ┌─ <query>:3:3
  │
3 │   actors := (FOR person IN movie.actors SELECT person),
  │   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ You can use assert_distinct() around the expression to turn this into a runtime assertion, or the DISTINCT operator to silently discard duplicate elements.
```

msullivan:
Currently we disable our `disjoint_union` tricks inside nested FOR
expressions (because we need to make sure expressions like
`for x in {1,2} union (for y in {3, 4} union y)`) are DUPLICATE.

This is unnecessarily strict, though, especially because if there is a
shape in between the two fors, whether the inner FOR loop is UNIQUE
can matter, like in:
```
FOR movie IN Movie SELECT movie {
  title,
  actors := (FOR person IN movie.actors SELECT person),
};
```

This is easy to fix by considering the inner iterator but making sure
disjoint_union doesn't leak.

Fixes #8748.

### DIFF
--- a/edb/edgeql/compiler/inference/multiplicity.py
+++ b/edb/edgeql/compiler/inference/multiplicity.py
@@ -628,8 +628,7 @@ def _infer_for_multiplicity(
     itmult = infer_multiplicity(itset, scope_tree=scope_tree, ctx=ctx)
 
     if itmult != DUPLICATE:
-        new_iter = itset.path_id if not ctx.distinct_iterator else None
-        ctx = ctx._replace(distinct_iterator=new_iter)
+        ctx = ctx._replace(distinct_iterator=itset.path_id)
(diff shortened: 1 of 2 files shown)


elprans (approved):
We should probably backpatch this

PR #8755, opened 2025-06-02
Add note about `init` and basic alias doc

scotttrinh:
~~`init` isn't yet documented or used consistently throughout the docs, so prefer the current command instead of the upcoming style.~~

Discussed internally and this is the change we decided was the best first step.

### DIFF
--- a/docs/intro/localdev.rst
+++ b/docs/intro/localdev.rst
@@ -16,7 +16,7 @@ It's easy to get started with a local Gel instance. Navigate to the root of your
 
 .. code-block:: bash
 
-  $ gel init
+  $ gel project init
 
 This command takes care of several things for you:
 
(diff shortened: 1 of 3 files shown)


1st1 (approved):

docs/intro/localdev.rst
```diff
@@ -16,7 +16,7 @@ It's easy to get started with a local Gel instance. Navigate to the root of your
 
 .. code-block:: bash
 
-  $ gel init
+  $ gel project init
```
1st1: Can we do this?

```suggestion
  $ gel init
```

This command (an alias to `edb:cmd:'gel project init`) takes care...

^^ might be even better, because you'll have a link to the actual command docs.

PR #8756, opened 2025-06-02
Add all_permissions property to roles.

dnwpark:
related https://github.com/geldata/gel/issues/8177

Adds an `all_permissions` computed properties to roles, which provides access to all permissions, directly from the role, as well as inherited from its memberships.

This is implemented by adding a new `_all_role_memberships` internal function which recursively collects a role's `member_of` property.

### DIFF
--- a/edb/buildmeta.py
+++ b/edb/buildmeta.py
@@ -57,7 +57,7 @@ from edb.common import verutils
 # The merge conflict there is a nice reminder that you probably need
 # to write a patch in edb/pgsql/patches.py, and then you should preserve
 # the old value.
-EDGEDB_CATALOG_VERSION = 2025_05_20_00_00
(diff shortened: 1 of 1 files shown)


edb/pgsql/metaschema.py
```diff
@@ -3156,6 +3156,82 @@ def __init__(self, schema: s_schema.Schema) -> None:
...
+        permissions_ptr = role_obj.getptr(
+            schema, s_name.UnqualName('permissions'), type=s_props.Property
+        )
+        permissions = _schema_alias_view_name(schema, permissions_ptr)
+        permissions = (
+            common.maybe_versioned_schema(permissions[0]), permissions[1]
+        )
+
+        text = f"""
+            WITH RECURSIVE memberships (id, member_of)
```
msullivan, line 3193: wooooooo

PR #8757, opened 2025-06-02
Implement `splat_strategy` qualifier for pointers

### LINKED ISSUE #8601: Add `lazy` schema field
1. `lazy := true` on a computed property or property excludes it from being fetched on `select {*}`
2. When set on a link / computed link it excludes it from `select {**}`
3. It has to be introspectable like other fields

msullivan:
Basically implement Yury's proposal.

I haven't added a future yet, though, so this changes splat behavior in a backward incompatible way.
We can add a future if we want.

Fixes #8601.

### DIFF
--- a/edb/edgeql/compiler/viewgen.py
+++ b/edb/edgeql/compiler/viewgen.py
@@ -995,7 +995,7 @@ def _raise_on_missing(
             if pn.name in next(iter(rewrites.by_type.values())):
                 continue
 
-        if ptrcls.is_property(ctx.env.schema):
+        if ptrcls.is_property():
             # If the target is a sequence, there's no need
             # for an explicit value.
             ptrcls_target = ptrcls.get_target(ctx.env.schema)
@@ -1809,7 +1809,7 @@ def _normalize_view_ptr_expr(
                 ]
 
                 ercls: type[errors.EdgeDBError]
-                if ptrcls.is_property(ctx.env.schema):
+                if ptrcls.is_property():
                     ercls = errors.InvalidPropertyTargetError
                 else:
                     ercls = errors.InvalidLinkTargetError
--- a/edb/edgeql/compiler/viewgen.py
+++ b/edb/edgeql/compiler/viewgen.py
@@ -47,6 +47,7 @@ from edb.ir import typeutils
 from edb.ir import utils as irutils
 import edb.ir.typeutils as irtypeutils
 
+from edb.schema import futures as s_futures
 from edb.schema import links as s_links
 from edb.schema import name as sn
 from edb.schema import objtypes as s_objtypes
@@ -778,7 +779,13 @@ def _expand_splat(
             splat_strat == qltypes.SplatStrategy.Explicit
             or (
                 splat_strat == qltypes.SplatStrategy.Default
-                and ptr.get_linkful(ctx.env.schema)
+                and (
+                    ptr.get_linkful(ctx.env.schema)
+                    if s_futures.future_enabled(
+                        ctx.env.schema, 'no_linkful_computed_splats'
+                    )
+                    else isinstance(ptr, s_links.Link)
+                )
             )
         ):
             continue
--- a/edb/edgeql/compiler/viewgen.py
+++ b/edb/edgeql/compiler/viewgen.py
@@ -775,8 +775,15 @@ def _expand_splat(
             continue
         if ptr.get_secret(ctx.env.schema):
             continue
-        # if ptr.get_lazy(ctx.env.schema):
-        #     continue
+        splat_strat = ptr.get_splat_strategy(ctx.env.schema)
+        if (
+            splat_strat == qltypes.SplatStrategy.Explicit
+            or (
+                splat_strat == qltypes.SplatStrategy.Default
+                and ptr.get_linkful(ctx.env.schema)
+            )
+        ):
+            continue
         sname = ptr.get_shortname(ctx.env.schema)
         # Skip any dunder properties; these are injected properties like
         # __tid__ and __tname__, and we want to manage injecting them
@@ -824,8 +831,9 @@ def _expand_splat(
                 continue
             if ptr.get_secret(ctx.env.schema):
                 continue
-            # if ptr.get_lazy(ctx.env.schema):
-            #     continue
+            splat_strat = ptr.get_splat_strategy(ctx.env.schema)
+            if splat_strat == qltypes.SplatStrategy.Explicit:
+                continue
             pn = ptr.get_shortname(ctx.env.schema)
             if (
                 (pn.name.startswith('__') and pn.name.endswith('__'))
--- a/edb/edgeql/compiler/viewgen.py
+++ b/edb/edgeql/compiler/viewgen.py
@@ -773,8 +773,10 @@ def _expand_splat(
     for ptr in pointers.objects(ctx.env.schema):
         if not isinstance(ptr, s_props.Property):
             continue
-        if ptr.get_secret(ctx.env.schema) or ptr.get_lazy(ctx.env.schema):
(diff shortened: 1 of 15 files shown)


edb/lib/schema.edgeql
```diff
@@ -298,6 +298,7 @@ CREATE ABSTRACT TYPE schema::Pointer
     CREATE PROPERTY default -> std::str;
     CREATE PROPERTY expr -> std::str;
     CREATE PROPERTY secret -> std::bool;
+    CREATE PROPERTY lazy -> std::bool;
```
elprans: Can this be an enum so that we can leave the door open for marking things as eager by default?

msullivan: We can make it an enum, but I don't totally understand this suggested use case? What are the values you propose and what values might we add later?

1st1:
I think we've made one mistake with the splats design and are about to make one with the ORM. 

### The problem

Let's me unwrap this:

1. We have, broadly, two categories of computeds -- properties (ones that return scalars) and links (return objects). Single star `*` splats include all properties and all computeds that return scalars, double splats `**` include links and computeds returning links.

3. We've made a decision for Python ORM to basically do `*` splats in expressions like `User.seelct()`. So the default fetching API would *always* generate queries to fetch all properties and all scalar computeds.

4. This PR is needed for (2) to let users make certain computeds opt-out from `*` splats, and consequently, the ORM eager fetching API.

### The mistake

Again, we have, broadly, two categories of computeds -- properties and links. But in reality it's way more nuanced than this.  Some users have crazy computeds that are "properties" - return a scalar, but are computed using *aggregate functions and complex queries in them*.  Add to the mix access policies on those complex computeds and you have a recipe for a disaster with eager computed loading.

### Path forward

I propose we:

* **Refine our classification:** re-classify computeds into two categories: 
 
  * those that involve only properties of the host type, 
  * those that do link traversal, even if done in an aggregate function and the return type is still a scalar.
  * Maybe *linkful computeds* and *linkless computeds" are good names? Idk you tell me.

* **Fix splats in Gel 7+:** philosophically, single star splats were always envisioned as convenience sugar, a loose equivalent to SQL's `SELECT *`, and were all about "give me properties but don't traverse links".
 
  * It would be really nice if we could find a way to stop including linkful computeds in `*`. Yes, it's backwards incompatible, but it's a mistake on our end. 
  * Maybe it should be a **schema-level "future"** that enables the new `*` behavior on all linkful computeds for the instance. This way new users will use the new behavior and old users... well, they can switch to "the new way" carefully if they wish so. **If not too complicated I'd do this for Gel 7.**

* **Don't repeat the mistake in the Python ORM** -- new APIs shouldn't load linkful computeds by default, they **must** be an opt-in either at the code (or at the schema level, more on that later):

  * But we don't have enough introspection power (as @elprans told me; I assume we'd need an EdgeQL parser/compiler in the Python ORM to deal with this) to understand if a computed is linkful or linkless, so...
  * ...we should not include *any* computeds in Python's `*` API equivalent.
  * When we fix computeds in Gel 7+, we can start eager-loading linkless computeds, *this should be a backwards compatible change* typing-wise.

* **Fix this PR:** seems like what we need is the opposite of "lazy". 
  
  * **Side note** full transparency, I didn't like any flavor of "lazy" when I first heard of this idea and still don't fully like it. Opting out linkful computeds one by one, keeping that a good practice, dealing with schema migrations, etc. all make this mechanism complicated and I suspect not widely used. 
  
    Users will keep hitting performance problems and our solution would be  -- take this pipette and meticulously apply this thing over all linkful computeds in your schema. But then you start doing that when you already hit the problem, meaning that after marking a thing lazy you'd have to grep/fix all your code. Ugly.
  * But if we **don't** load linkful computeds by default in the Python ORM (and in the future in `select Type{*}`) then we still want a mechanism to include some of them by default.
  * Why? Schema evolution. Computeds can be clutch when a property / link should be dropped, but to make existing code work you introduce a computed to fill in. Those computeds can be, in theory, complex, so what was a property might become a linkful computed. We'd still want that computed to be included in `*` splat, so...
  * ...we still need this PR, but we have to flip its logic.

-------

Easy fix to this PR is to rename "lazy" to "eager". But just like with our decision to not use bools for env variables and config values, we should do better here with an enum. I'll seed the bike shedding with `LoadStrategy`, with following members:

* `default` -- let Gel determine the outcome
* `explicit` (old `lazy`) -- don't include this in splats, the user has to explicitly name the field
* `implicit` (= `eager`) -- include this computed in splats implicitly because it's really really needed there.

-------

Thoughts? cc @vpetrovykh @elprans @msullivan

CarsonF:
Since this is being brought up, I'd like to offer my two cents as a user.
Yury mentions property vs link, but really wanting to avoid path traversal.

It is common in my app to eagerly fetch links, but only their ID, so that they can be lazily fetched later.
**Neither single nor double splats solve this use case.**
It seems like just the ID of a (single) link pointer, is just a foreign key and does not need joins / traversal.
It would be really helpful if there were a way to include those links in some splat form.
Even if via an `implicit/eager` annotation.
```edgeql
type User;
type Member {
  required user: User;
}
select Member {*} #=> [{ user: { id: "..." } }]
```

edb/schema/pointers.py
```diff
@@ -187,6 +187,27 @@ def merge_readonly(
     return current
 
 
+# def merge_lazy(
```
elprans: Drop this?

edb/schema/pointers.py
```diff
@@ -459,6 +480,18 @@ class Pointer(referencing.NamedReferencedInheritingObject,
...
+    splat_strategy = so.SchemaField(
+        qltypes.SplatStrategy,
+        allow_ddl_set=True,
+        describe_visibility=(
+            so.DescribeVisibilityPolicy.SHOW_IF_EXPLICIT_OR_DERIVED_NOT_DEFAULT
+        ),
+        coerce=True,
+        default=qltypes.SplatStrategy.Default,
+        compcoef=0.909,
+        # merge_fn=merge_splat,
```
elprans: Stray comment

tests/test_edgeql_select.py
```diff
@@ -1993,6 +1993,20 @@ async def test_edgeql_select_splat_06(self):
...
+    async def test_edgeql_select_splat_07(self):
+        res = json.loads(await self.con.query_json(
+            r'''
+            select Issue {
+                **,
+            }
+            filter .number = "1"
+            '''
+        ))
+        # Make sure the lazy properties aren't included
```
elprans:
```suggestion
        # Make sure the explicit properties aren't included
```

tests/test_edgeql_select.py
```diff
@@ -2005,6 +2005,24 @@ async def test_edgeql_select_splat_07(self):
         # Make sure the explicit properties aren't included
         self.assertNotIn('tags', res[0])
         self.assertNotIn('related_to', res[0])
+        # num_watchers should be included, without the future!
+        self.assertIn('num_watchers', res[0])
+
+        await self.con.execute('''
+            create future no_linkful_computed_splats
```
1st1, line 2012: Now need to teach `gel init` to do it... Can you file an issue?

msullivan:
Darn, I realized as I merged that I didn't update the comment - I *did* add a future

edb/edgeql/compiler/viewgen.py
```diff
@@ -820,6 +834,11 @@ def _expand_splat(
         for ptr in pointers.objects(ctx.env.schema):
             if not isinstance(ptr, s_links.Link):
                 continue
+            if ptr.get_secret(ctx.env.schema):
+                continue
```
aljazerzen, line 838: Ooo, sneaky "fix" and an undocumented change

PR #8760, opened 2025-06-03
Look up role permissions when executing queries.

dnwpark:
Enables permissions in queries and access policies.

### DIFF
--- a/edb/edgeql/compiler/expr.py
+++ b/edb/edgeql/compiler/expr.py
@@ -639,10 +639,9 @@ def compile_GlobalExpr(
 
     # If we are compiling with globals suppressed but still allowed, always
     # treat it as being empty.
-    if (
-        isinstance(glob, s_globals.Global)
-        and ctx.env.options.make_globals_empty
-    ):
+    if ctx.env.options.make_globals_empty:
+        if isinstance(glob, s_permissions.Permission):
+            default_ql = qlast.Constant.boolean(False)
         if default_ql:
             return dispatch.compile(default_ql, ctx=ctx)
         else:
@@ -664,10 +663,7 @@ def compile_GlobalExpr(
         and not ctx.env.options.json_parameters
     ):
         param_set, present_set = setgen.get_global_param_sets(
-            glob,
-            ctx=ctx,
-            # Permissions act as implicit globals
-            is_implicit_global=isinstance(glob, s_permissions.Permission)
+            glob, ctx=ctx,
         )
     else:
         param_set, present_set = setgen.get_func_global_param_sets(
--- a/edb/edgeql/compiler/expr.py
+++ b/edb/edgeql/compiler/expr.py
@@ -578,15 +578,7 @@ def compile_GlobalExpr(
 
     # Check for permission
     if isinstance(expr_obj, s_permissions.Permission):
-        std_type = sn.QualName('std', 'bool')
-        ct = typegen.type_to_typeref(
-            ctx.env.get_schema_type_and_track(std_type),
-            env=ctx.env,
-        )
-        ir_expr = irast.BooleanConstant(
-            value="false", typeref=ct, span=expr.span
-        )
-        return setgen.ensure_set(ir_expr, ctx=ctx)
+        return setgen.get_permission_param_set(expr_obj, ctx=ctx)
 
     # Check for non-global
     if not isinstance(expr_obj, s_globals.Global):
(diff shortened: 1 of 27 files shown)


edb/edgeql/compiler/expr.py
```diff
@@ -578,15 +578,7 @@ def compile_GlobalExpr(
...
-        std_type = sn.QualName('std', 'bool')
-        ct = typegen.type_to_typeref(
-            ctx.env.get_schema_type_and_track(std_type),
-            env=ctx.env,
-        )
-        ir_expr = irast.BooleanConstant(
-            value="false", typeref=ct, span=expr.span
-        )
-        return setgen.ensure_set(ir_expr, ctx=ctx)
+        return setgen.get_permission_param_set(expr_obj, ctx=ctx)
```
msullivan: We probably need to support the same "reading it out of json" stuff we do for globals, right?

msullivan: Would it be better to treat them *more* like globals, and treat them identically in the compiler (with a flag in `irast.Global` maybe) before handling them differently in the server.

edb/server/compiler/compiler.py
```diff
@@ -3682,15 +3691,41 @@ def _extract_extensions(
...
+        role_memberships[role] = list(
+            role.get_bases(global_schema).objects(global_schema)
+        )
+        role_permissions[role] = list(sorted(
+            role.get_permissions(global_schema) or ()
+        ))
+
+    for role in schema_roles:
+        role_name = str(role.get_name(global_schema))
+        extracted_roles[role_name]['all_permissions'] = list(set(
```
msullivan: Make it a tuple so it's immutable

edb/server/compiler/compiler.py
```diff
@@ -1898,11 +1899,17 @@ def _compile_ql_query(
...
     if ir.globals:
         globals = [
             (str(glob.global_name), glob.has_present_arg)
             for glob in ir.globals
         ]
+    if isinstance(ir, irast.Statement) and ir.permissions:
+        permissions = [
+            str(permission.global_name)
+            for permission in ir.permissions
+        ]
```
msullivan: I think maybe we should keep the permissions treated as globals until this stage, and split them here?

msullivan (commented):
This is looking pretty good.

We also need to inject permissions in the json execute paths

edb/server/compiler/compiler.py
```diff
@@ -3682,15 +3693,41 @@ def _extract_extensions(
...
         )
-    return immutables.Map(roles)
+
+    # To populate all_permissions, combine the permissions of each role
+    # and its ancestors.
+    role_memberships: MutableMapping[s_role.Role, list[s_role.Role]] = {}
+    role_permissions: MutableMapping[s_role.Role, Sequence[str]] = {}
+    for role in schema_roles:
+        role_memberships[role] = list(
+            role.get_bases(global_schema).objects(global_schema)
```
msullivan, line 3723: Is this just direct parents, though?

tests/test_edgeql_ddl.py
```diff
@@ -8917,7 +8917,8 @@ async def test_edgeql_ddl_permissions_01(self):
         """)
         await self.assert_query_result(
             'select global foo;',
-            [False],
+            # Tests run as superuser
```
msullivan, line 8920: We need dump tests too

tests/test_server_permissions.py
```diff
@@ -0,0 +1,307 @@
...
+import json
+
+import edgedb
+
+from edb.testbase import server as tb
+
+
+class TestServerPermissions(tb.ConnectedTestCase):
+
+    PARALLELISM_GRANULARITY = 'system'
```
msullivan, line 28: I wonder if we can figure out something better

msullivan (approved):

msullivan:
There are some mandatory follow-ups here but I'm going to merge it now to get the catalog changes merged

edb/server/compiler/compiler.py
msullivan: Yeah, looks like there is a bug where we only get permissions from direct parents

msullivan: Lol but it was a one line fix so I really should have just included it in this PR

msullivan:
Work on #8177

PR #8764, opened 2025-06-04
Reduce the indentation churn in #8721

msullivan:


msullivan:
We could merge this if we want but don't need to

PR #8771, opened 2025-06-04
Add sys::data_modification to grant non-superusers the ability to run DML.

dnwpark:


msullivan (commented):
I think the logic is right but I'd tweak the flow a little: instead of directly manipulating the protocol's `allow_capabilities` bitmask based on logic, we should compute a mask based just on the role, and then and the masks together.

Then, ideally, we would arrange to not recompute the mask each time.

edb/lib/sys.edgeql
```diff
@@ -443,3 +443,8 @@ sys::approximate_count(
     USING SQL FUNCTION 'edgedb.approximate_count';
     set impl_is_strict := false;
 };
+
+
+CREATE PERMISSION sys::data_modification;
+CREATE PERMISSION sys::ddl;
+CREATE PERMISSION sys::persistent_config;
```
1st1: Can we move these into a sub-namespace? `sys::perm::*`?

msullivan: Yeah why not

msullivan: https://github.com/geldata/rfcs/pull/115

PR #8772, opened 2025-06-05
Fix mangled type names leaking into introspectable names

### LINKED ISSUE #7878: The introspect name of type is INCONSISTENT
image

v4.5 on Cloud

Schema:
```type ShipmentItem  {
		printCounterTuples: array<tuple<labelName: str, printedQty:int16, note: str, qtyPerPrint:int16>>;
}```

as show in picture, the target.name should be 

```array<tuple<labelName:std::str, printedQty:std::int16, note:std::str, qtyPerPrint:std::int16>>"``` instead of ```array<tuple<labelName:std|str, printedQty:std|int16, note:std|str, qtyPerPrint:std|int16>>"```

### PR DESCRIPTION
**Pull request by msullivan**
For nested tuple/array types and for some computed global types, we
were putting mangled names as the display name.

Fix this by "recursively" unmangled ::s in type names, and by only
using the shortname for types.

Fixes #7878

### DIFF
--- a/edb/schema/name.py
+++ b/edb/schema/name.py
@@ -233,6 +233,7 @@ def shortname_from_fullname(fullname: Name) -> Name:
 
 unmangle_re_1 = re.compile(r'\|+')
 
+
 def recursively_unmangle_shortname(name: str) -> str:
     # Any number of pipes becomes a single ::.
     return unmangle_re_1.sub('::', name)

### CODE REVIEW

File: tests/test_edgeql_ddl.py [hunk 1]
```diff
@@ -7928,17 +7928,17 @@ async def test_edgeql_ddl_global_type_changes_06(self):
                 'create global foo := [(f := [1])]',
                 [
                     {
-                        'name': 'array<tuple<f:array<std||int64>>>',
+                        'name': 'array<tuple<f:array<std::int64>>>',
```
elprans, line 7931: It'd be nice if we inserted a space after the `:` in named tuples, but not critical.

msullivan:
Planning to do that in a follow up PR

PR #8777, opened 2025-06-06
Test running inplace upgrades from instances created with 6.x

### LINKED ISSUE #8776: Run inplace upgrade tests against 6.0
Currently we have inplace upgrade tests that test some synthetic upgrades from master, but we need to test from 6.0 to master.

msullivan:
It uses the dump tests, since those should (should!) not change
between versions.

We'll test against 6.0 and 6.latest.
Fixes #8776.

### DIFF
--- a/.github/scripts/patches/compute-ipu-versions.py
+++ b/.github/scripts/patches/compute-ipu-versions.py
@@ -1,0 +1,62 @@
+# Compute prior minor versions to test upgrading from
+
+
+import json
+import os
+import pathlib
+import re
+import sys
+from urllib import request
+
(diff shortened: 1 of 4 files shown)


.github/scripts/patches/compute-ipu-versions.py
```diff
@@ -0,0 +1,58 @@
+# Compute prior minor versions to test upgrading from
+
+import edb.buildmeta
+
+import json
+import os
+import re
+from urllib import request
+
+base = 'https://packages.edgedb.com'
```
elprans:
```suggestion
base = 'https://packages.geldata.com'
```

elprans (approved):

PR #8800, opened 2025-06-13
Fix rewrites making UPDATE infer as ONE

### LINKED ISSUE #8791: Mutation rewrites changes cardinality on update to single
- Gel Version: 6.7+130d82b
- Gel CLI Version: Gel CLI 7.2.0+c5b8c5e
- OS Version: MacOs Sonoma 14.5

Steps to Reproduce:

1. Create the schema
  ```
  type A {
    updatedAt: datetime;
  }

  type B {
    updatedAt: datetime {
      rewrite update using (
        datetime_of_statement()
      )
    }
  }
  ```
2. Create a static query file `test.edgeql` with `update B set {}`
3. Generate the query file: `bunx @gel/generate queries --target ts`
4. The inferred return is now a singular object and not an array:
  ```
export type TestReturns = {
  "id": string;
};

  ```

This doesn't impact runtime queries but anything relying on the inferred cardinality should get problems.

msullivan:
This was caused by some code doing something nonsensical, with a
comment above it that I wrote saying that it must be wrong but made a
particular test pass. That test still passes if I remove it, so all
right.

Fixes #8791.

### DIFF
--- a/edb/edgeql/compiler/viewgen.py
+++ b/edb/edgeql/compiler/viewgen.py
@@ -1238,11 +1238,6 @@ def _compile_rewrites_for_stype(
                 scopectx.iterator_path_ids |= {anchor.path_id}
                 scopectx.anchors[key] = anchor
 
-            # XXX: I am pretty sure this must be wrong, but we get
-            # a failure without due to volatility issues in
-            # test_edgeql_rewrites_16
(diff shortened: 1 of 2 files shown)


msullivan:
> Oh, maybe the reason it passes now is because path factoring is off...

Nope, works fine with path factoring on.

elprans (approved):

PR #8803, opened 2025-06-16
Add std::lang::py annotations

msullivan:
See https://github.com/geldata/gel-python/issues/627#issuecomment-2977922610

### DIFF
--- a/edb/schema/schema.py
+++ b/edb/schema/schema.py
@@ -83,7 +83,10 @@ STD_MODULES = (
     sn.UnqualName('std::_test'),
     sn.UnqualName('std::fts'),
     sn.UnqualName('std::lang'),
+    sn.UnqualName('std::lang::go'),
+    sn.UnqualName('std::lang::js'),
     sn.UnqualName('std::lang::py'),
+    sn.UnqualName('std::lang::rs'),
     EXT_MODULE,
     sn.UnqualName('std::enc'),
 )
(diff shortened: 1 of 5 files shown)


edb/schema/schema.py
```diff
@@ -82,6 +82,8 @@
     sn.UnqualName('std::pg'),
     sn.UnqualName('std::_test'),
     sn.UnqualName('std::fts'),
+    sn.UnqualName('std::lang'),
+    sn.UnqualName('std::lang::py'),
```
1st1, line 88: Can you add `::js` / `::go` / `::rust` too?

But now I'm wondering... for Python we'll have just a fully qualified type name and then we can auto-import. For Go/JS/Rust the situation might be different...

@elprans @msullivan thoughts?

msullivan: Do we know what we want to put in js/go/rust?

1st1: Probably same thing - for JS it can be a zod schema, for rust - a serde struct to unpack JSON. But... now I'm thinking that our annotations are text - so we don't need to define the exact format. So let's just add the annotations and we'll find the format for them in the near (hopefully) future. `</train-of-thoughts>`

msullivan: I think `rs` instead of `rust` if we are abbreviating python and javascript

1st1: yeah sure

1st1 (approved):

PR #8822, opened 2025-06-23
rbac: Make a branch_config permission but require superuser for system

msullivan:
See geldata/rfcs#119

### DIFF
--- a/tests/test_server_permissions.py
+++ b/tests/test_server_permissions.py
@@ -799,8 +799,7 @@ class TestServerPermissions(tb.EdgeQLTestCase, server_tb.CLITestCaseMixin):
             ''')
 
             await conn.execute('''
-                CONFIGURE CURRENT BRANCH SET apply_access_policies
-                    := false
+                CONFIGURE CURRENT BRANCH RESET apply_access_policies
             ''')
 
         finally:


edb/server/compiler/enums.py
```diff
@@ -42,6 +42,8 @@ class Capability(enum.IntFlag):
     DDL               = 1 << 3    # noqa
     PERSISTENT_CONFIG = 1 << 4    # noqa
 
+    BRANCH_CONFIG     = 1 << 59   # noqa
+    INSTANCE_CONFIG   = 1 << 60   # noqa
```
dnwpark, line 46: `INSTANCE_CONFIG` doesn't seem like it's used anywhere?

tests/test_server_permissions.py
```diff
@@ -762,6 +762,53 @@ async def test_server_permissions_session_config_02(self):
                 DROP ROLE foo;
             ''')
 
+    async def test_server_permissions_persistent_config_01(self):
```
dnwpark, line 765: maybe should be named `test_server_permissions_branch_config_01`

dnwpark (approved):

tests/test_server_permissions.py
msullivan: it tests that instance config isn't allowed too

edb/server/compiler/enums.py
msullivan: https://github.com/geldata/gel/pull/8822/files#diff-d2fc37bc724ef58ac44d52da73c773483ba90dd9818f3acede2664abd700e850R2618-R2621

PR #8828, opened 2025-06-24
Add sys::perm::sql_session_config.

dnwpark:
Adds `sys::perm::sql_session_config` which guards sql `SET [LOCAL | SESSION]` and `set_config()`.

related #8177

### DIFF
--- a/edb/server/protocol/pg_ext.pyx
+++ b/edb/server/protocol/pg_ext.pyx
@@ -30,6 +30,7 @@ import sys
 import time
 import uuid
 from collections import deque
+from typing import Sequence
 
 cimport cython
 import immutables
@@ -1000,6 +1001,9 @@ cdef class PgConnection(frontend.FrontendConnection):
             source = pg_parser.Source.from_string(query_str)
             query_units = await self.compile(source, dbv)
 
+        for qu in query_units:
+            self.check_capabilities(qu)
+
         already_in_implicit_tx = dbv._in_tx_implicit
         metrics.sql_queries.inc(
             len(query_units), self.tenant.get_instance_name()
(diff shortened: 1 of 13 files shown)


edb/server/protocol/pg_ext.pyx
```diff
@@ -1329,6 +1333,33 @@ cdef class PgConnection(frontend.FrontendConnection):
...
+                    if perm not in permissions:
+                        missing = sorted(
+                            set(query_unit.required_permissions)
+                            - set(permissions)
+                        )
+                        plural = 's' if len(missing) > 1 else ''
+                        raise errors.DisabledCapabilityError(
+                            f'role {self.username} does not have required '
+                            f'permission{plural}: {", ".join(missing)}'
+                        )
```
msullivan: Another option would be to add a SQL_SESSION_CONFIG to the `Capability` enum.
I don't know if I have a particularly strong opinion

tests/test_server_permissions.py
```diff
@@ -980,3 +980,141 @@ async def test_server_permissions_dump_01(self):
...
+
+            with self.assertRaisesRegex(
+                asyncpg.exceptions.InternalServerError,
+                # should be edgedb.DisabledCapabilityError?
+                'role foo does not have required permission: '
+                'sys::perm::sql_session_config'
+            ):
+                await conn.execute("""
+                    SET SESSION transaction_isolation TO 'serializable'
+                """)
```
msullivan, line 1023: I think we also need to disable it for `query_sql` calls over the gel protocol

msullivan (approved):

tests/test_server_permissions.py
dnwpark: Added a test to check for this.

edb/server/protocol/pg_ext.pyx
dnwpark: Had a think about this, and I changed it to use capabilities.

Maybe we use `required_permissions` for things that come out of the schema, such as for functions.

msullivan: Oh, functions are something we need to check too, if we support calling edgeql functions from SQL, which I can't remember if we do

msullivan (approved):

PR #8837, opened 2025-06-27
Implement permissions in access policies over SQL adapter

dnwpark:
Given the schema:
```
PERMISSION CanSelect;
TYPE Widget {
    PROPERTY n -> int64;
    ACCESS POLICY can_select ALLOW SELECT USING (global CanSelect);
};
ROLE Foo {
    SET permissions := default::CanSelect;
};
ROLE Bar;
```

The SQL query `SELECT "n" FROM "Widget"` will only return values for `admin` and `Foo`.

related #8177

### DIFF
--- a/edb/pgsql/resolver/command.py
+++ b/edb/pgsql/resolver/command.py
@@ -2738,7 +2738,9 @@ def merge_params(
             )
             ctx.query_params.append(
                 dbstate.SQLParamGlobal(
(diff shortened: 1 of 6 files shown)


aljazerzen:
(I'm using a convention to append "over SQL adapter" to PRs that deal with sql resolver or pgproto stuff)

PR #8839, opened 2025-06-27
Fix ISE when computed global is introspected using GraphQL.

### LINKED ISSUE #8814: In GraphQL extension wrongly handled computed global
- Gel Version: 7.0-dev.9624+d3c72d1 / 6.8+7ffd5c3
- Gel CLI Version: 7.7.0-dev.1402+460093f
- OS Version: Windows 11

Steps to Reproduce:

1. create a simple schema:
   ```
   using extension graphql;
   
   module default {
       global current_user_id: uuid;
       global current_user := (select User filter .id = global current_user_id);
   
       type User {
           name: str;
       }
   }
   ```
2. run the graphql introspection query
3. the query result with error something like this:
   ```
   InternalServerError: TypeError: Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but 'Filtercurrent_user@default|current_user@global' does not.
   
   Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but 'current_user@default|current_user@global' does not.
   
   Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but 'current_user@default|current_user@global' does not.
   
   Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but 'Ordercurrent_user@default|current_user@global' does not.
   
   Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but 'delete_current_user@default|current_user@global' does not.
   ```

dnwpark:
close #8814

Given the schema:
```
using extension graphql;

module default {
    global current_user_id: uuid;
    global current_user := (select User filter .id = global current_user_id);
    type User;
}
```

The graphql query `query { __schema { types { name } } }` would produce an ISE.

### DIFF
--- a/edb/graphql/types.py
+++ b/edb/graphql/types.py
@@ -405,6 +405,15 @@ class GQLCoreSchema:
     def graphql_schema(self) -> GraphQLSchema:
         return self._gql_schema
 
+    def _get_type_gql_name(self, type: s_types.Type) -> str:
+        if type.get_from_global(self.edb_schema):
+            # The names of global types are mangled so use shortname instead
+            typename = type.get_shortname(self.edb_schema)
+        else:
+            typename = type.get_name(self.edb_schema)
+        assert isinstance(typename, s_name.QualName)
+        return self.get_gql_name(typename)
+
     @classmethod
     def get_gql_name(cls, name: s_name.QualName) -> str:
         module, shortname = name.module, name.name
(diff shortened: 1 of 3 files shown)


edb/graphql/types.py
```diff
@@ -405,6 +405,15 @@ def edgedb_schema(self) -> s_schema.Schema:
     def graphql_schema(self) -> GraphQLSchema:
         return self._gql_schema
 
+    def _get_type_gql_name(self, type: s_types.Type) -> str:
+        if type.get_from_global(self.edb_schema):
+            # The names of global types are mangled so use shortname instead
+            typename = type.get_shortname(self.edb_schema)
```
msullivan, line 411: We should maybe fix global type names, but this is a good first thing.

PR #8846, opened 2025-06-30
testing

dnwpark:
seeing what happens when we add permissions to things in the schema


edb/schema/extensions.py
```diff
@@ -869,8 +869,14 @@ def _delete_begin(
 
         def _name_in_mod(name: sn.Name) -> bool:
             return (
-                (isinstance(name, sn.QualName) and name.module == module)
-                or name == module_name
+                (
+                    isinstance(name, sn.QualName)
+                    and name.module.startswith(module_name.name)
+                )
```
aljazerzen: I assume that the intention is to match module `foo` with name `foo::bar::Hello`.

But this will also match `foo_fighters::Singer`

dnwpark: whoops! good catch!

edb/lib/sys.edgeql
```diff
@@ -452,3 +452,15 @@ CREATE PERMISSION sys::perm::data_modification;
 CREATE PERMISSION sys::perm::ddl;
 CREATE PERMISSION sys::perm::branch_config;
 CREATE PERMISSION sys::perm::sql_session_config;
+
+ALTER FUNCTION std::sequence_reset(
```
msullivan, line 460: Put a comment about why this is defined late?

msullivan (approved):

msullivan:
The only thing I kind of wonder now is whether it would be better to error when trying to do reads, instead of allowing them and filtering with the policy.

Merge this though, and we can argue about that later.

(We don't currently have that good of a mechanism for doing it, but we could add one; and even now we could hack it up by calling a function in the policy that has the rule, though that is kind of awful)

dnwpark:
@msullivan It might be possible to add something to types similar to how functions have `required_permissions`

PR #8848, opened 2025-07-01
Fix a bad assert

aljazerzen:
I was asserting a original source text was found,
but that is the case only when a span has been
found too.

Introduced in #7071


msullivan:
Is there a case that this breaks? Should it be backported?

msullivan (approved):

PR #8849, opened 2025-07-01
Access policies for link tables over SQL adapter

aljazerzen:


edb/pgsql/resolver/relation.py
```diff
@@ -624,3 +636,70 @@ def _compile_read_of_obj_table(
...
     )
     return pgast.Relation(name=cte_name)
+
+
+def _compile_read_of_link_table(
+    obj: s_links.Link | s_properties.Property,
+    include_inherited: bool,
+    table: context.Table,
+    ctx: Context,
+) -> pgast.BaseRelation:
```
msullivan, line 646: Add some comments making clear that this is for the policy case, and what is being done?

edb/pgsql/resolver/relation.py
```diff
@@ -624,3 +636,70 @@ def _compile_read_of_obj_table(
         )
     )
     return pgast.Relation(name=cte_name)
+
+
+def _compile_read_of_link_table(
+    obj: s_links.Link | s_properties.Property,
+    include_inherited: bool,
```
msullivan, line 643: `include_inherited` is unused in the function, which I think is a problem?
I think we need to respect that also.

(It looks like that is an existing bug: links with access policies on them don't have the policies apply, but *do* lose their subtypes?)

aljazerzen: I've added handling for inherited link table (and tests)

msullivan (approved):
Looks good!

PR #8855, opened 2025-07-03
Try making schema::Object filtered by a permission

msullivan:
Required some fixes in other places

But I don't think we should land this.

### DIFF
--- a/edb/server/protocol/execute.pyx
+++ b/edb/server/protocol/execute.pyx
@@ -912,7 +912,13 @@ async def execute_system_config(
     await conn.sql_fetch(b'select 1', state=state)
 
     if query_unit.sql:
-        data = await conn.sql_fetch_col(query_unit.sql)
+        data = await conn.sql_fetch_col(
+            query_unit.sql,
+            args=(
+                tuple(b'\x01' for _ in query_unit.permissions)
+                if query_unit.permissions else ()
+            ),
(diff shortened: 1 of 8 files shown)


edb/server/protocol/execute.pyx
```diff
@@ -912,7 +912,13 @@ async def execute_system_config(
     await conn.sql_fetch(b'select 1', state=state)
 
     if query_unit.sql:
-        data = await conn.sql_fetch_col(query_unit.sql)
+        data = await conn.sql_fetch_col(
+            query_unit.sql,
+            args=(
+                tuple(b'\x01' for _ in query_unit.permissions)
```
dnwpark, line 918: This is probably only being checked by superuser, but probably this should still check `dbv.get_permissions`.

edb/server/tenant.py
```diff
@@ -457,6 +457,10 @@ def _refresh_role_capabilities(self) -> None:
                     capability |= compiler_enums.Capability.BRANCH_CONFIG
                 if 'sys::perm::sql_session_config' in available_permissions:
                     capability |= compiler_enums.Capability.SQL_SESSION_CONFIG
+                if 'sys::perm::analyze' in available_permissions:
```
dnwpark, line 460: Analyze hasn't been created?

PR #8858, opened 2025-07-04
Refactor a few grammar rules

aljazerzen:
This PR contains a few changes to the grammar, none of the should have any real impact.

----

The original goal of this PR:

My initial goal was to simplify grammar a lot, by merging all possible statements into one giant non-term. That statement would then be used in the top-level block, in `CRATE MODULE`, in `CREATE TYPE` and basically all commands that use a sub-block. That would significantly reduce the size of the grammar, the serialized spec (which is currently 7.5MiB) and would also reduce the memory usage of the parser.

The obvious downside would be that following statements would be syntactically correct:

```
RENAME TO x;
ALTER BRANCH blah {
     CREATE TYPE User;
     COMMIT;
     CREATE MIGRATION X {
         DROP EXTENDING;
     };
};
```

Because of this, we would need small compiler stage after parsing (or directly in qlcompiler) that would reject all statements that appear in wrong contexts.

---

This plan fell into water (if this idiom does not exist in english, I mean fell apart), because we have statements that have overlapping syntax. For example, syntax for `EXTENDING` depends on the context, for example:

```
CREATE TYPE A EXTENDING B;
------------------------- this is all CreateTypeStmt

ALTER TYPE A { EXTENDING B };
------------   -----------
        |               \__ this is AlterSimpleExtending
         \__ AlterTypeStmt

# So the first stmt could also be parsed as:
CREATE TYPE A EXTENDING B;
------------- -----------
        |               \__ this is AlterSimpleExtending
         \__ CreateTypeStmt
... which causes a parser conflict.
```

There are more examples of this, and my conclusion is that this is unfixable. The language we constructed is inherently large and cannot be deconstructed into a few simple components such as Expression, Statement and StatementBlock, but really does need a lot of different kinds of expressions, statements and blocks.

### DIFF
--- a/edb/edgeql/parser/grammar/ddl.py
+++ b/edb/edgeql/parser/grammar/ddl.py
@@ -68,27 +68,35 @@ class DDLStmt(Nonterm):
         pass
 
     @parsing.inline(0)
-    def reduce_AlterRoleExtending(self, *_):
-        pass
-
-    @parsing.inline(0)
     def reduce_ExtensionPackageStmt(self, *_):
-        pass
-
-    @parsing.inline(0)
-    def reduce_MigrationStmt(self, *_):
         pass
 
     @parsing.inline(0)
     def reduce_OptWithDDLStmt(self, *_):
         pass
 
+    @parsing.inline(0)
+    def reduce_MigrationStmt(self, *_):
+        pass
+
+
+class DDLWithBlock(Nonterm):
+    @parsing.inline(0)
+    def reduce_WithBlock(self, *_):
+        pass
+
 
 class OptWithDDLStmt(Nonterm):
-    def reduce_WithBlock_InnerDDLStmt(self, *kids):
+    def reduce_DDLWithBlock_WithDDLStmt(self, *kids):
         self.val = kids[1].val
         self.val.aliases = kids[0].val.aliases
 
+    @parsing.inline(0)
+    def reduce_WithDDLStmt(self, *_):
+        pass
+
+
+class WithDDLStmt(Nonterm):
     @parsing.inline(0)
     def reduce_InnerDDLStmt(self, *_):
         pass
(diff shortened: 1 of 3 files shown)


aljazerzen:
One way to get around this is to remove `EXTENDING` from `CreateTypeStmt`, which make sure that the example is unambiguously parsed as `CreateTypeStmt(commands=[AlterSimpleExtending])`. But this does not work, because `CreateTypeStmt` has non-terminals *following* extending part:

```
CREATE TYPE A EXTENDING B IF NOT EXISTS;
------------- ----------- ------------- 
           |           |             \_ this must be part of CreateTypeStmt
           |            \_ AlterSimpleExtending
            \_ CreateTypeStmt

If we wanted allowed that, we would also have to allow this:
CREATE TYPE A { EXTENDING B; } IF NOT EXISTS;
-------------   ------------   ------------- 
           |           |             \_ this must be part of CreateTypeStmt
           |            \_ AlterSimpleExtending
            \_ CreateTypeStmt
```

aljazerzen:
Anyway, I've attempted this in 1c3fcf03dacc3f617309af7ddb662b0dbd4d8023, but it is reverted so the small changes can be merged.

I'm disappointed and content on never trying to simplify our grammar again.

edb/tools/parser_demo.py
```diff
@@ -107,6 +107,13 @@ def main():
...
     '''
-    CREATE MIGRATION { ;;; CREATE TYPE Foo ;;; CREATE TYPE Bar ;;; };
+    CREATE MIGRATION
+    {
+        set global foo := "test";
+        alter type Foo {
+            create required property name -> str {
+            set default := (global foo);
+        }
+    }; }
```
dnwpark, line 117: Extra brace here?

aljazerzen: That's the test case.

PR #8860, opened 2025-07-07
Implement URI lookup for AI embedding models.

dnwpark:
WIP


edb/schema/indexes.py
```diff
@@ -356,6 +361,12 @@ class Index(
         allow_ddl_set=True,
     )
 
+    # If this is a generated AI index, track the created type
+    generated_ai_model_type = so.SchemaField(
```
elprans, line 367: Is this just to keep the reference?  I wonder if you can formulate the generated ref injection via `find_extra_refs` callback in `compile_expr_field` instead?

dnwpark: Ah, I am not actually compiling anything into an expression, so I'm not sure that would work.

First, I manually create a `qlast.CreateObjectType` for the new model type:
https://github.com/geldata/gel/blob/5f8e00b879bec7f7b2e11781798e3a6df0185802/edb/schema/indexes.py#L1272

I compile it:
https://github.com/geldata/gel/blob/5f8e00b879bec7f7b2e11781798e3a6df0185802/edb/schema/indexes.py#L1335-L1339

Then, return to `IndexCommand._handle_ai_index_op` and  make it a prerequisite of the `CreateIndex`/`AlterIndex` command:
https://github.com/geldata/gel/blob/5f8e00b879bec7f7b2e11781798e3a6df0185802/edb/schema/indexes.py#L1025

However, to get the typeshell of the type produced by the command, I compile the command a second time and actually apply it to the schema:
https://github.com/geldata/gel/blob/5f8e00b879bec7f7b2e11781798e3a6df0185802/edb/schema/indexes.py#L1341-L1355

edb/schema/types.py
```diff
@@ -101,6 +101,13 @@ class Type(
         ExprType,
         default=None, compcoef=0.909)
 
+    # Whether this type was generated from elsewhere in the schema.
+    is_schema_generated = so.SchemaField(
+        bool,
+        default=False,
+        compcoef=0.0,
+    )
```
aljazerzen, line 109: We have similar fields `from_alias` and `from_global`. Maybe they can be reused?

aljazerzen (commented):
I understand about half of this, but also don't really understand what it does - even after reading the tests. From what I gather, it allows using AI models from the JSON file in the user schema?

I can spend more time here, but I'd need an explanation.

edb/schema/types.py
dnwpark: `from_alias` has a lot of associated logic with links and views. I thought it best to avoid adding more meanings to it. @msullivan will hopefully have more insight

dnwpark:
@anbuzin What sort of issues were you having with ollama? Right now, not all models are in the reference json yet so that might be it.

anbuzin:
@dnwpark It was more like, you have to insert a config with an empty string for a secret, and I couldn't get the RAG to work with `llama3.2`, but I haven't had a chance to figure out exactly why

dnwpark:
@anbuzin Ah yeah, you need to config the provider no matter what. In theory, you could override `api_url` to have ollama to run on a server.

tests/test_ext_ai.py
```diff
@@ -1422,3 +1425,737 @@ def test_batch_embeddings_inputs_01(self):
                 ([4, 1], 6),
             ],
         )
+
+
+class TestExtAIDDL(tb_server.DDLTestCase):
```
msullivan, line 1432: We need tests where we add new entries to the manifest, after having done the initial creation.

We should probably only check the file once per server startup (maybe? or in the eventual case maybe once per 24 hours of something).

Since this version doesn't support the http loading anyway, I think the testing approach should be to support passing the file name via an environment variable. Then we can create a temp test dir, start up the server in that directory with start_edgedb_server, do some work, stop it, populate a new updated file, start the server pointing at that instead, and then do the rest of the testing.

Does that make sense?

edb/schema/ddl.py
```diff
@@ -669,6 +673,7 @@ def delta_from_ddl(
         Mapping[tuple[sn.Name, Optional[str]], uuid.UUID]
     ]=None,
     compat_ver: Optional[verutils.Version] = None,
+    reference_paths: Optional[Mapping[str, pathlib.Path]] = None
```
msullivan, line 676: I think we should pass this in as a mapping to the *data*. Or if we want to generalize things further, to some abstract interface type for retrieving data?

msullivan (changes requested):
I've concluded that I don't like this approach. I think that this will be cleaner if we take a "JSON-forward" approach, where instead of driving the extension worker based on querying the introspection schema, we drive it based on an in-memory datastructure driven by the JSON. It should be pretty easy to produce a data structure for the db-configured stuff with an introspection query; probably we could directly produce the JSON directly using `json_object_pack`, if we wanted...

One thing that we will need to do either in that new approach or if we wanted to continue this approach, is to cache the last downloaded version of the file. Otherwise, if the fetch fails on server startup or whatever, there will be a degradation of service. In the JSON-forward approach it would be worse, but even in this schema-based approach it would prevent running `migration create`.

msullivan:
Hm... let me dig into something a bit more. I had forgotten that we actually copy the annotations onto the indexes themselves... but I can't remember why?

msullivan:
Here's a query that will generate the exact same JSON I think, from introspecting:
```
WITH ai_objs := (
  SELECT schema::ObjectType FILTER 'ext::ai::EmbeddingModel' IN .ancestors.name
)
SELECT json_object_pack(
  FOR obj IN ai_objs
  FOR j IN json_object_pack(
    FOR a IN obj.annotations
    SELECT (
        re_replace('^embedding_model_', '',
          re_replace('^ext::ai::', '', a.name),
        ),
        if a.name like 'ext::ai::model_%'
        then <json>a@value else to_json(a@value)
    )
    FILTER a.name LIKE 'ext::ai::%'
  )
  SELECT (<str>j['model_name'], j)
);
```

Though maybe you'd be better off just doing
```
SELECT schema::ObjectType {
    annotations: {name, @value},
}
FILTER 'ext::ai::EmbeddingModel' IN .ancestors.name;
```
and postprocessing in python :)

Issue #8786, opened 2025-06-10
Function with union type ISE's in SDL


msullivan:
```
type Foo {
  required name: str;
}
type Bar {
  name: str;
}

function name(x: Foo | Bar) -> optional str using (x.name);
```

It works in DDL (https://github.com/geldata/gel/pull/7759).

```
    /home/sully/src/e/edgedb/edb/server/compiler_pool/worker.py, line 169, in compile
        > units, cstate = COMPILER.compile_serialized_request(
    /home/sully/src/e/edgedb/edb/server/compiler/compiler.py, line 607, in compile_serialized_request
        > return self.compile(
    /home/sully/src/e/edgedb/edb/server/compiler/compiler.py, line 686, in compile
        > unit_group = compile(ctx=ctx, source=request.source)
    /home/sully/src/e/edgedb/edb/server/compiler/compiler.py, line 2656, in compile
        > return _try_compile(ctx=ctx, source=source)
    /home/sully/src/e/edgedb/edb/server/compiler/compiler.py, line 2830, in _try_compile
        > return _try_compile_ast(statements=statements, source=source, ctx=ctx)
    /home/sully/src/e/edgedb/edb/server/compiler/compiler.py, line 2882, in _try_compile_ast
        > comp, capabilities = _compile_dispatch_ql(
    /home/sully/src/e/edgedb/edb/server/compiler/compiler.py, line 2561, in _compile_dispatch_ql
        > query = ddl.compile_dispatch_ql_migration(
    /home/sully/src/e/edgedb/edb/server/compiler/ddl.py, line 450, in compile_dispatch_ql_migration
        > return _start_migration(ctx, ql, in_script)
    /home/sully/src/e/edgedb/edb/server/compiler/ddl.py, line 530, in _start_migration
        > target_schema, warnings = s_ddl.apply_sdl(
    /home/sully/src/e/edgedb/edb/schema/ddl.py, line 574, in apply_sdl
        > ddl_stmts = s_decl.sdl_to_ddl(target_schema, documents)
    /home/sully/src/e/edgedb/edb/edgeql/declarative.py, line 448, in sdl_to_ddl
        > trace_dependencies(decl_ast, ctx=tracectx)
    /usr/local/lib/python3.12/functools.py, line 909, in wrapper
        > return dispatch(args[0].__class__)(*args, **kw)
    /home/sully/src/e/edgedb/edb/edgeql/declarative.py, line 1094, in trace_Function
[...]

PR #8863, opened 2025-07-10
Fix params with union types producing ISEs 

dnwpark:
Related #8786

Previously, this schema would produce an error:
```
type Foo {
  required name: str;
}
type Bar {
  name: str;
}

function name(x: Foo | Bar) -> optional str using (x.name);
```


dnwpark:
A followup PR will address function params with type intersections.

aljazerzen (approved):

PR #8865, opened 2025-07-15
Add permissions to sys types and functions.

dnwpark:
related #8177

Create the following access policies and function requirements.

In `sys`
- `TYPE sys::Role`
  - `ACCESS POLICY ap_read ALLOW select USING (global sys::perm::superuser);`
- `FUNCTION sys::_describe_roles_as_ddl`
  - ` required_permissions := sys::perm::ddl`
- `FUNCTION sys::_get_all_role_memberships`
  - ` required_permissions := sys::perm::ddl`
- `TYPE sys::QueryStats`
  - `ACCESS POLICY ap_read ALLOW select USING (global sys::perm::superuser);`
- `FUNCTION sys::reset_query_stats`
  - ` required_permissions := sys::perm::ddl`

In `schema`
- `Type schema::Permission`
  - `ACCESS POLICY ap_read ALLOW select USING (global sys::perm::superuser);`
- `Type schema::Migration`
  - `ACCESS POLICY ap_read ALLOW select USING (global sys::perm::ddl);`

### DIFF
--- a/edb/buildmeta.py
+++ b/edb/buildmeta.py
@@ -57,7 +57,7 @@ from edb.common import verutils
 # The merge conflict there is a nice reminder that you probably need
 # to write a patch in edb/pgsql/patches.py, and then you should preserve
 # the old value.
-EDGEDB_CATALOG_VERSION = 2025_07_23_00_00
+EDGEDB_CATALOG_VERSION = 2025_07_24_00_00
 EDGEDB_MAJOR_VERSION = 7
 
 
--- a/edb/lib/sys.edgeql
+++ b/edb/lib/sys.edgeql
@@ -465,10 +465,21 @@ sys::approximate_count(
     set required_permissions := { sys::perm::superuser };
 };
 
-# Add permissions to std.
+# Add permissions to schema and std.
 
-# The std module is populated before sys permissions so we need to
+# These modules are populated before sys permissions so we need to
 # add these restrictions here.
+
+ALTER TYPE schema::Permission {
+    CREATE ACCESS POLICY ap_read allow select using (
+        global sys::perm::superuser
+    );
+};
+ALTER TYPE schema::Migration {
+    CREATE ACCESS POLICY ap_read allow select using (
+        global sys::perm::ddl
+    );
+};
 
 ALTER FUNCTION std::sequence_reset(
     seq: schema::ScalarType,
--- a/edb/lib/sys.edgeql
+++ b/edb/lib/sys.edgeql
@@ -20,6 +20,14 @@
 CREATE MODULE sys;
 
 
+CREATE MODULE sys::perm;
+CREATE PERMISSION sys::perm::superuser;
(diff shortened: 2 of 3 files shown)


edb/ir/typeutils.py
```diff
@@ -87,6 +87,16 @@ def is_cfg_view(
     )
 
 
+def is_excluded_cfg_view(
+    child: s_obj.Object,
+    *,
+    ancestor: s_obj.Object,
+    schema: s_schema.Schema,
+) -> bool:
+    """Used to exclude sys/cfg tables from non sys/cfg views for performance."""
```
msullivan: Include a link to the issue and the comment that we don't really like this anyway

edb/edgeql/compiler/policies.py
```diff
@@ -469,16 +490,27 @@ def try_type_rewrite(
...
+            # We need to explicitly exclude cfg views to prevent them from
+            # from showing up in type rewrites. Normally this happens in
+            # inheritance.get_inheritance_view.
+            children = frozenset(
+                child
+                for child in children
+                if not irtyputils.is_excluded_cfg_view(
+                    child, ancestor=stype, schema=schema
+                )
+            )
```
msullivan: Can we do this unconditionally? I would think so; am I missing something?

edb/edgeql/compiler/policies.py
```diff
@@ -412,16 +420,29 @@ def try_type_rewrite(
...
+            expand_children = True
+
+        # One of the children contains a cfg view.
+        if any(
+            irtyputils.is_excluded_cfg_view(
+                desc, ancestor=stype, schema=schema
+            )
+            for desc in all_child_descs
+        ):
+            expand_children = True
```
msullivan: Do we need to force the expansion here? If it falls back to relying on the `inheritance` expansion, that's fine?

msullivan (approved):

edb/edgeql/compiler/policies.py
dnwpark: We only need to do this if we detected a cfg view earlier, so we can skip this in most cases I think.

edb/edgeql/compiler/policies.py
```diff
@@ -109,6 +111,9 @@ def has_own_policies(
     return any(
         has_own_policies(stype=child, skip_from=stype, ctx=ctx)
         for child in stype.children(schema)
+        if irtyputils.is_excluded_cfg_view(
+            child, ancestor=stype, schema=schema
+        )
```
msullivan: This seems clearly wrong to me. (Like, it seems that there should be a `not` there!)

It makes me think that either this whole check is not needed or the tests are missing something important...

msullivan: If you have `A extending B`, `B extending C`, and a policy on C, and then `select A`, this change will make the `C` policy not apply.

Please fix this here and add a test for that case to test_edgeql_policies.py

PR #8867, opened 2025-07-15
Separate Parameter into QueryParameter and FunctionParameter.

dnwpark:
WIP


edb/edgeql/compiler/setgen.py
```diff
@@ -2162,9 +2167,14 @@ def get_func_global_json_arg(*, ctx: context.ContextLevel) -> irast.Set:
...
 
+    is_func_param = ctx.env.options.func_params is not None
+    parameter_type = (
+        irast.FunctionParameter if is_func_param else irast.QueryParameter
+    )
+
     # If this is because we have json params, not because we're in a
     # function, we need to register it.
-    if ctx.env.options.json_parameters:
+    if bool(ctx.env.options.json_parameters):
```
msullivan: Why? Calling `bool` in an `if` is always unnecessary (and rarely if ever makes the code clearer)

edb/edgeql/compiler/setgen.py
```diff
@@ -2125,8 +2125,13 @@ def get_global_param_sets(
         None
     )
 
+    is_func_param = ctx.env.options.func_params is not None
+    parameter_type = (
+        irast.FunctionParameter if is_func_param else irast.QueryParameter
+    )
```
msullivan: I think this is all unnecessary, and this code path is only ever called with `func_params is None`, so I think we should just assert that and not bother with all the casing?

msullivan: (Unless I'm just wrong here?)

edb/edgeql/tracer.py
```diff
@@ -647,13 +647,25 @@ def trace_Constant(node: qlast.BaseConstant, *, ctx: TracerContext) -> None:
...
 
 
+@trace.register
+def trace_FunctionParameter(
+    node: qlast.FunctionParameter, *, ctx: TracerContext
+) -> None:
+    raise errors.SchemaError(
+        'function parameters are not allowed in schemas',
+        span=node.span,
+    )
```
msullivan: This is a bad message -- "function parameters" clearly *are* allowed in schemas. But the `FunctionParameter` ast node isn't, because it gets created on-the-fly and substituted for a path during compilation. This should be an AssertionError I think.

(And maybe we should add a comment to the definition of qlast.FunctionParameter explaining where it gets created, since it's not really in the parsed version!)

edb/edgeql/compiler/setgen.py
dnwpark: I thought it might possible when called via `compile_FunctionCall`, but it turns out to not be the case.

PR #8869, opened 2025-07-16
Defer loading cookies until needed in auth ext + return better error message

jaclarke:
If there are invalid cookies in the request to any of the http endpoints, the server returns a not very helpful `HttpParserCallbackError: User callback error` error message.

As only the auth ext endpoints actually read the cookies, defer parsing the cookie header until the request is being handled by the auth ext, and return a better error message in the case of a `CookieError`.

### DIFF
--- a/edb/server/protocol/auth_ext/http.py
+++ b/edb/server/protocol/auth_ext/http.py
@@ -138,6 +138,14 @@ class Router:
         )
 
         try:
+            try:
+                request.load_cookies()
+            except http.cookies.CookieError as ex:
+                raise errors.InvalidData(
+                    f'invalid cookies header: {ex}, '
+                    f'try clearing cookies and trying again'
+                )
+
             match args:
                 # PKCE token exchange route
                 case ("token",):
--- a/edb/server/protocol/protocol.pxd
+++ b/edb/server/protocol/protocol.pxd
@@ -35,6 +35,7 @@ cdef class HttpRequest:
         public bytes authorization
         public object params
         public object forwarded
+        public bytes cookie_header
         public object cookies
 
 
--- a/edb/server/protocol/protocol.pyi
+++ b/edb/server/protocol/protocol.pyi
@@ -40,6 +40,9 @@ class HttpRequest:
     forwarded: dict[bytes, bytes]
     cookies: http.cookies.SimpleCookie
 
+    def load_cookies(self) -> None:
+        ...
+
 class HttpResponse:
     status: http.HTTPStatus
     close_connection: bool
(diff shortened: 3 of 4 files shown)


edb/server/protocol/auth_ext/http.py
```diff
@@ -138,6 +138,14 @@ async def handle_request(
         )
 
         try:
+            try:
+                request.load_cookies()
+            except http.cookies.CookieError as ex:
+                raise errors.InvalidData(
+                    f'invalid cookies header: {ex}, '
+                    f'try clearing cookies and trying again'
+                )
```
scotttrinh: Seems better than the status quo, for sure! I wonder if we can do one better and actually ignore and clear out bad cookies based on the error? Conceptually, I always feel like cookies are things owned by the server, not by the client. If course, it's possible that the cookies are just completely mangled to the point where it's impossible to even know which key needs to be cleared.

scotttrinh (approved):
I don't think we need to address my question right this second, since it'll be hard to even test for this until we see what the exception actually says to know what code path we're hitting. So :shipit:

edb/server/protocol/auth_ext/http.py
1st1: I think our error pages should be a bit richer / nicely designed, but also have a JS snippet that can nuke cookies and any other local state (this error page could just do that, right?)

jaclarke: > I always feel like cookies are things owned by the server

Normally, yes. But for local instances we're using localhost, and the users app is likely also using localhost, so we effectively share cookies with their app. It feels like we're just going to cause more weird, hard to debug, problems if we start destroying cookies when the users app has a bug that causes it to set bad cookies.

jaclarke: Maybe the best approach is to have some custom cookie parsing that just ignores invalid cookies (we can log a warning), as we only need to read the known good cookies we set anyway.

jaclarke: Ok, I've updated it to ignore invalid cookies: https://github.com/geldata/gel/pull/8869/commits/2f060fd8e93d7ce9cf689c05d2411493063d1b06

scotttrinh: The downside of ignoring invalid cookies is that you don't know it's messed up, right? I guess another approach is to lazily load cookies at the point when we need them and to do it selectively rather than populating a whole cookie jar, just ask for the morsels we need? Because then if something is messed up, there is clearly a bug. This isn't a blocker, though: this is still superior to the status quo in every way.

msullivan:
Should we merge this?

mmastrac:
Can we send back `Clear-Site-Data: "cache", "cookies"` when this fails?

Issue #8831, opened 2025-06-25
Expose graphql and potentially ai::rag in the protocol


1st1:
Let's add a new language field(s) for GraphQL queries (return JSON) and potentially for ai::rag (streaming of our data messages with JSON?)

GraphQL use case is more important, as we want to simplify repl and webui workflows. Currently setting values for globals is complicated, because no GrpahQL client supports our non-standard stuff. @msullivan points out that GraphQL can work over the protocol even if the extension isn't enabled, as extension, mostly, is more about enabling the HTTP endpoint than it is about anything else.

The rag functionality can be a cool REPL gimmick but also will allow us shipping `gel` client libs with fewer dependencies.

msullivan:
>  because no GrpahQL client supports our non-standard stuff

We fixed this ages ago. Globals can be passed in as variables

msullivan:
Other than that, I can think of three other options:
1. Add a graphql endpoint that *doesn't* work using the normal Parse/Execute flow, but via a totally new kind of message, but which would still run in the same transaction and use session globals, etc. This would probably not be too hard to implement, but would require more work on client authors, who will need to actually do new protocol stuff. I think this is super ugly and would rather just do it right, I think.
2. Add a new protocol message for doing requests against HTTP endpoints over an existing and authenticated binary connection. This would let the CLI and UI do graphql and RAG requests without establishing a new connection. This is more general than option 1, but wouldn't natively include session globals.
3. Don't do anything; the CLI and UI can just do HTTP requests, what's the big deal?

PR #8878, opened 2025-07-23
Support graphql over the binary protocol

msullivan:
This is implemented by adding another language enum field, and then
using the same flow as edgeql and SQL.

We'll only expose the JSON output mode in clients for now, because the
binary one doesn't work right for queries that do things like
introspection or @include.

Handling things like introspection and @include is the only reason
that this PR isn't trivial: graphql introspection queries are
implemented by having the compiler call out to the graphql-core
implementation to execute the query, and then including the json
response in the output. Making this work when the query might depend
on a variable requires tracking "key variables" and including them in
the cache keys, and recompiling when they don't match.

We already did this in the graphql HTTP endpoint but now need to
support it in the main flow.

Two main caveats to follow-up on:
 * This first take doesn't support doing the query normalization to
   extract constants. I'll do that in a follow-up.
 * We don't support queries that have default values for parameters.
   This is because the gel bindings don't have a way to *not pass*
   a parameter. (They can pass None, but that's not the same thing!)
   Not sure the best thing to do here.

Mostly finishes #8831.
(I'll close that when we land this and open follow-ups.)

### DIFF
--- a/edb/graphql/translator.py
+++ b/edb/graphql/translator.py
@@ -2017,8 +2017,6 @@ def translate_ast(
     if variables is None:
         variables = {}
 
-    # variables = {}
-
     validation_errors = convert_errors(
         graphql.validate(gqlcore.graphql_schema, document_ast),
         substitutions=substitutions)
--- a/edb/graphql/translator.py
+++ b/edb/graphql/translator.py
@@ -98,6 +98,8 @@ class GraphQLTranslatorContext:
         query,
         document_ast,
         operation_name,
+        native_input,
+        parse_only_mode,
     ):
         self.variables = variables
         self.fragments = {}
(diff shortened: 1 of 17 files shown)


edb/graphql/translator.py
```diff
@@ -1930,11 +2005,20 @@ def translate_ast(
...
+    # If no variables have been provided, and we are in native
+    # protocol mode, that means we need to be tolerant of critical
+    # variables not having values. We'll report out which ones
+    # existed, and then recompile with the variables present.
+    parse_only_mode = native_input and variables is None
+
     if variables is None:
         variables = {}
 
+    # variables = {}
```
elprans: stray

edb/graphql/translator.py
```diff
@@ -170,10 +174,23 @@ def __init__(self, values):
         self.update(values)
         self.touched = set()
 
+    def __bool__(self):
+        # HACK! And kind of wrong!
```
elprans, line 178: 😱

PR #8881, opened 2025-07-24
Fix policy object duplication in degenerate multiple inheritance cases

msullivan:
If we have `B extending A` and `C extending B, A`, the logic in
policies.py for determining if `A` has overlapping children fails to
realize that it clearly does, and so generates broken code that
selects from `C` twice.

Should fix one of the test failures in #8865.


dnwpark (approved):
This makes sense

PR #8886, opened 2025-07-25
Do constant extraction on graphql over the native protocol

### LINKED ISSUE #8879: Normalize graphql queries over binary protocol
Follow up to #8831

msullivan:
Fixes #8879

### DIFF
--- a/edb/server/compiler/rpc.pyi
+++ b/edb/server/compiler/rpc.pyi
@@ -25,6 +25,9 @@ from edb import edgeql
 from edb.server import defines, config
 from edb.server.compiler import sertypes, enums
 
+from edb import graphql
+from edb.pgsql import parser as pgparser
+
 class SQLParamsSource:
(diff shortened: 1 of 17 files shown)


edb/server/compiler/rpc.pyi
```diff
@@ -42,7 +45,10 @@ class SQLParamsSource:
         ...
 
 class CompilationRequest:
-    source: edgeql.Source
+    # FIXME: I didn't want to track down the errors when I put the right
+    # annotation in.
+    # source: edgeql.Source | graphql.Source | pgparser.Source
+    source: typing.Any
```
aljazerzen: I've fixed this in a commit I've just pushed. I've also missed `SQLParamsSource`

aljazerzen (approved):
Looks good.

edb/server/compiler/rpc.pyi
msullivan: Thanks

PR #8887, opened 2025-07-26
Make default value of apply_access_policies_pg to True for non-superusers.

dnwpark:
related #8177


msullivan (changes requested):
Yeah, let's not do this superuser based stuff.

Let's just flip the flag and warn users about it.

(I had though I might reuse some of the infrastructure here for a PR I'm working on that is related, but I'm actually doing it a totally different way)

PR #8889, opened 2025-07-28
Add `global sys::current_role`

msullivan:
Work on #8177. The thing that convinced me to actually do this is that
SQL already supports this, but it is done in the compiler and
interacts incorrectly with the query caches. Fix that by making
SQL use the new global.


edb/edgeql/compiler/config.py
```diff
@@ -298,6 +299,14 @@ def _validate_global_op(
         modaliases=ctx.modaliases, type=s_globals.Global)
     assert isinstance(glob, s_globals.Global)
 
+    fullname = glob.get_name(ctx.env.schema)
+    # wait is this wrong???
```
elprans: Is it?

msullivan: No, I don't think so. (If it was wrong in the way I had worried about, tests would have failed.)

Issue #8876, opened 2025-07-22
Daily first redirect to Google auth screen fails


anieuwland:
- Gel Version: v6.7
- Gel CLI Version: v7.6 but issue is older
- OS Version: Linux, but using Gel Cloud

## Issue

When I attempt to log in user a Google account, this fails on my first attempt. I am using the built-in auth ui and have a valid Google oauth2 client configured. When I choose Google in the auth ui, the first time I do so in a day or so, I am forwarded to the Gel page and get the message `{"error": {"message": "Expecting value: line 1 column 1 (char 0)", "type": "InternalServerError"}}`. I am not forwarded to Google.

If I subsequently refresh (ctrl + r) or go back and choose google again, Gel is successful in whatever it tries to do and forwards me to Google for actually logging in.

## Steps to Reproduce

0. I have not visited cokeryen.nl today or recently and go to cokeryen.nl.
1. I choose 'Inloggen' (= Log in) and am forwarded to Gel's built-in auth UI. I am now on `https://cokeryen--anieuwland.c-21.i.aws.edgedb.cloud/db/prd/ext/auth/ui/signin?challenge=NLiw8VYSpPDe-sLwzGdvcIpuKIyuwJmcGdmk8005NB8`
2. I choose Continue with Google and immediately receive `{"error": {"message": "Expecting value: line 1 column 1 (char 0)", "type": "InternalServerError"}}`.
3. Pressing ctrl+r, I am correctly forwarded.

* Use Gel Cloud
* Use built-in auth provider Google

In step 2, the exact request done that results in the 500 error is:

```bash
[...]

PR #8891, opened 2025-07-29
Fix bad HTTP cache response when cache expires and upstream is not modified

mmastrac:
This fixes the issue in #8876 where the Google auth page would fail once per day. There was a race condition between cache expiry and re-validation: if the LRU cache holding the request body expired after the request was made but before the response was returned, the empty body of the "not modified" response was returned.

### DIFF
--- a/rust/gel-http/src/cache.rs
+++ b/rust/gel-http/src/cache.rs
@@ -1,3 +1,4 @@
+use bytes::Bytes;
 use http::{HeaderMap, Method, Request, Response, Uri};
 use http_cache_semantics::{AfterResponse, BeforeRequest, CacheOptions, CachePolicy};
 use lru::LruCache;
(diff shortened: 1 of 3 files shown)


rust/gel-http/src/cache.rs
```diff
@@ -7,14 +8,34 @@ use std::{
...
-    Response(http::Response<Vec<u8>>),
+    /// The cache is stale or missing, so we need to make a request. Note that CachedItem
+    /// may contain a stale cache item and must be returned in `after_request`.
+    Request(http::Request<Vec<u8>>, Cacheditem),
+    /// The cache is fresh, so we can return the response
+    Response(http::Response<Bytes>),
+}
+
+#[derive(Debug)]
+pub struct Cacheditem {
```
scotttrinh:
```suggestion
pub struct CachedItem {
```

For consistency?

rust/gel-http/src/cache.rs
```diff
@@ -65,22 +94,30 @@ impl Cache {
             },
             cache: Arc::new(Mutex::new(CacheItems::new(
                 NonZero::new(100).unwrap(),
-                1024 * 1024,
+                10 * 1024 * 1024,
```
scotttrinh, line 97: Just checking: this increases the cache max size to 10MB right?

scotttrinh (approved):
No blockers! Thanks for the extensive test suites here.

rust/gel-http/src/cache.rs
mmastrac: Good eye, I missed this!

rust/gel-http/src/cache.rs
mmastrac: Correct. TBH I don't understand _why_ that cache request was falling out of the LRU (I wonder if there was a second race?), but I think it's reasonable to hold up to 10MB of cached JWK keys _just in case_ that was somehow related.

PR #8905, opened 2025-07-31
One-time code implementation for `ext::auth`

scotttrinh:
This PR implements One-Time Code (OTC) based authentication as an alternative to traditional link-based verification, based on the design outlined in https://github.com/geldata/rfcs/pull/117. This change introduces a `verification_method` configuration option on authentication providers, allowing developers to choose between `Link` and `Code` verification methods for actions like email verification and passwordless sign-in.

### Things to Look For:

*   **Conditional Branching**: Handlers for authentication flows are structured to conditionally branch based on the `verification_method` config value. This allows the same endpoint to initiate either a link-based flow or an OTC flow depending on the provider's configuration.
*   **Rate Limiting**: Rate limiting is implemented by logging each verification attempt. When a code is submitted, a single atomic query checks the code's validity and the number of recent attempts to prevent brute-force attacks. OTCs also have a time-to-live (TTL) and are deleted after successful use or eventual cleanup.
*   **`OneTimeCode` Type**: A new `ext::auth::OneTimeCode` object is introduced to store the hashed code, its expiration, and other metadata.
*   **UI and Email Templates**: New UI components and email templates have been added to support the code input flow.

---

I'm currently doing a bunch of manual testing, but this is a good checkpoint to get feedback on the general shape of the solution. If the general approach is working, we'll just merge this and merge fixes on top of this if manual testing finds any bugs.


edb/server/protocol/auth_ext/otc.py
```diff
@@ -0,0 +1,238 @@
...
+
+from edb.server.protocol import execute
+
+from . import errors
+
+logger = logging.getLogger(__name__)
+
+
+def generate_code() -> str:
+    return f"{secrets.randbelow(1000000):06d}"
```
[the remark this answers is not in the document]
scotttrinh: > could potentially generate codes with leading zeros that might be displayed differently

I'm not sure what this means: we're already using an f-string here to make sure that it's formatted into a string which would preserve leading zeros, right?

```pycon
>>> import secrets
>>> f"{secrets.randbelow(100):02d}"
'05'
```

Someone chime in here if the suggestion with `zfill` is better than what is here.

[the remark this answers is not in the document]
msullivan: No, it's not

edb/lib/ext/auth.edgeql
```diff
@@ -139,6 +139,46 @@ CREATE EXTENSION PACKAGE auth VERSION '1.0' {
         };
     };
 
+    create type ext::auth::OneTimeCode extending ext::auth::Auditable {
```
msullivan, line 142: Adding as a note, though this probably doesn't need action as part of *this* PR:
To support in-place upgrades, we are going to also need a patch for upgrading to the old version to this one.
The contents of the patch will be the same either way, but we have two basic options:
1. Bump the version number to 1.1 or 2.0, and use the new `extension package migration` system to specify the patch. This will require some user action to do the upgrade.
2. Build the patch in and apply it automatically on in-place upgrades. We already do need to do this to apply the AI patches that got added in the 6.x cycle. (See #8900, which I think I almost have working)

I think I would in theory like to always move towards versioned extensions and stop applying extension patches automatically on upgrades, but we are already still doing the magic upgrades in 7.0 so maybe whatever

scotttrinh: > Bump the version number to 1.1 or 2.0, and use the new extension package migration system to specify the patch. This will require some user action to do the upgrade.

I think I would prefer this, but I assume users _cannot_ stay on the old version, right? What does the user action look like?

edb/server/protocol/auth_ext/http.py
```diff
@@ -753,66 +755,152 @@ async def handle_verify(
...
-        try:
-            email_factor = await self._try_verify_email(
-                provider=data["provider"],
-                identity_id=token.subject,
-            )
-            await self._maybe_send_webhook(
-                webhook.EmailVerified(
-                    event_id=str(uuid.uuid4()),
-                    timestamp=datetime.datetime.now(datetime.timezone.utc),
+        if verification_token:
```
msullivan, line 766: The commit message claims that this should be branching on `verification_method`?

In any case it is worth having a comment here about the logic, and maybe to have the subparts in separate functions?

edb/lib/ext/auth.edgeql
msullivan: They can keep using the old version for a bit, though unless we ship the old version with the new release, it will only work on instances that have been upgraded in place.

If they don't specify the version number, it might be as simple as running `migration create` on the new version and applying it. If they do specify a version number, they'll need to bump it I think.

I'll experiment slightly with this.

edb/server/protocol/auth_ext/http.py
```diff
@@ -1039,32 +1156,118 @@ async def handle_reset_password(
...
 
+        email_password_client = email_password.Client(db=self.db)
+
         try:
-            token = jwt.ResetToken.verify(
-                reset_token,
-                self.signing_key,
-            )
+            if reset_token:
+                # Token-based password reset (existing flow)
```
msullivan: These "existing flow" and "new flow" comments are actually quite useful for me here but maybe not so much as long-term comments.

Might be worth splitting into functions here also?

edb/server/protocol/auth_ext/http.py
```diff
@@ -1373,49 +1637,147 @@ async def handle_magic_link_authenticate(
...
+                    failure_reason = "unknown"
+                    error_msg = str(ex)
+                    if "Maximum verification attempts exceeded" in error_msg:
+                        failure_reason = "rate_limited"
+                    elif "Invalid code" in error_msg:
+                        failure_reason = "invalid_code"
+                    elif "Code has expired" in error_msg:
+                        failure_reason = "expired"
+                    elif "Verification failed" in error_msg:
+                        failure_reason = "verification_failed"
```
msullivan: This code is duplicated a bunch

edb/server/protocol/auth_ext/otc.py
```diff
@@ -0,0 +1,238 @@
...
+    ),
+
+    # Find the OTC
+    otc := (
+        select ext::auth::OneTimeCode
+        filter .factor = FACTOR and .code_hash = code_hash
+        limit 1
+    ),
+
+    is_rate_limited := failed_attempts >= 5,
```
msullivan: Make `5` either a query parameter or with bind it separately as a "constant"?

edb/server/protocol/auth_ext/otc.py
```diff
@@ -0,0 +1,238 @@
...
+    is_code_found := (exists otc) ?? false,
```
msullivan: `?? false` not needed

edb/server/protocol/auth_ext/otc.py
```diff
@@ -0,0 +1,238 @@
...
+        if (exists FACTOR)
+        then (
+            insert ext::auth::AuthenticationAttempt {
+                factor := FACTOR,
+                attempt_type :=
+                    ext::auth::AuthenticationAttemptType.OneTimeCode,
+                successful := is_code_valid,
+            }
+        )
+        else (<ext::auth::AuthenticationAttempt>{})
```
msullivan: Did it require the cast? I at least tried to make `else` blocks not require it on empty sets...

edb/server/protocol/auth_ext/otc.py
```diff
@@ -0,0 +1,238 @@
...
+
+    Returns:
+        The OneTimeCode ID if verification succeeds
+
+    Raises:
+        NoIdentityFound: If code is invalid, expired, or max attempts exceeded
+    """
+    code_hash = hash_code(code)
+
+    # Single composed query that handles the entire verification workflow
```
msullivan: I love edgeql sometimes

msullivan (approved):
(I didn't review the UI stuff at all)

edb/server/protocol/auth_ext/http.py
scotttrinh: Yeah, that's good point. We are branching on the config for whether or not we send a token or a code, but the endpoint doesn't care: it'll work for either method. The upshot of that is that if you change between the two methods, your old codes or links will still work since this endpoint can handle either. That's not an intentional decision, but it seems ... fine?

Splitting up the logic seems rational, and this entire http module is due for a pretty thorough refactoring, but right now, each route corresponds to a single route handler function. Ideally, in a refactoring pass, I'd like the http module to only handle HTTP stuff (input parsing, response marshalling), and to have _all_ of the logic and side effects to happen in other modules/classes with less orchestration here. Once that's accomplished, I hope there is less of a need to split this up.

edb/lib/ext/auth.edgeql
scotttrinh: Seems like the old version (without the new `verification_method` property) would just flat out run into runtime errors, no?

edb/server/protocol/auth_ext/otc.py
scotttrinh: Yeah, I actually want to make this configurable at some point in the future, but for now, I'll bind it as a constant, good idea.

edb/server/protocol/auth_ext/otc.py
scotttrinh: Dude, seriously. I was shocked it actually worked as expected, and had to have the tests actually prove it to me.

edb/server/protocol/auth_ext/otc.py
scotttrinh: Ya know, I didn't actually check, I've just been in the habit of writing casts for empty sets because sometimes I need them, and I don't have a good intuition for when I can omit them. I'll give it a shot!

edb/lib/ext/auth.edgeql
msullivan: I think we'd need to make it handle it the case where it is missing, and have the right default?

scotttrinh: > I think we'd need to make it handle it the case where it is missing, and have the right default?

It feels weird for the server to know about the schema changes and try to support multiple versions, but I guess this is new territory for us and we'll need to decide what we want to do. The places where we write the `.verification_method` to the provider config are centralized enough that we can do the fallback for this case, but all of the machinery around rate limiting authentication attempts just doesn't exist without the updated schema. I'm not using it yet for rate limiting things _except_ Code validations, but ideally we rate limit every kind of authentication attempt (wrong password, etc.) and the server would need to know to skip it if it's on `1.0` or something I suppose?

Maybe what we really want here is a way to have completely different `auth_ext` modules load based on the version number and keep around the old code when we upgrade? I guess that smells an awful lot like publishing packages...

msullivan: I think for this one we can look at whether verification_method is set, or something?

For more general things, we'll have to check the versions.

PR #8907, opened 2025-07-31
Get rid of mistral_common dependency

msullivan:
mistral_common has a lot of dependencies, soem of which are causing us
problems. Mistral support is kind of marginal, so we'll drop the
dependency and fall back to doing the same dumb truncation we do for
ollama.


dnwpark (approved):
:')

elprans:
FWIW, I got things building without kicking `mistral_common` out.

PR #8913, opened 2025-08-04
Fix OTC flow for built-in auth UI

scotttrinh:
### Testing status

Each flow with both verification required and not, and link vs code.

- [X] Magic Link Sign Up
- [X] Magic Link Sign In
- [X] Email+password email verification
- [x] Email+password forgot password
- [x] Email+password resend verification
- [x] WebAuthn email verification
- [x] WebAuthn resent verification

### Changes

Externally, this fixes some issues with not having enough context passed to the built-in UI. Internally, we're now using only the config to decide if the built-in UI shows the code input or not, which was suggested a while back, but now it's required due to testing and finding that we already use `code` in query parameter to pass the PKCE code in the email+password registration.

### DIFF
--- a/edb/server/protocol/auth_ext/http.py
+++ b/edb/server/protocol/auth_ext/http.py
@@ -1596,6 +1596,10 @@ class Router:
                 response.content_type = b"application/json"
                 response.body = json.dumps(return_data).encode()
         except Exception as ex:
+            request_accepts_json: bool = request.accept == b"application/json"
+            if request_accepts_json:
+                raise ex
+
             redirect_on_failure = data.get(
                 "redirect_on_failure", data.get("redirect_to")
             )


edb/server/protocol/auth_ext/http.py
```diff
@@ -1551,9 +1580,9 @@ async def handle_magic_link_email(
...
 
-            return_data = {
-                "email_sent": email,
-            }
+                    return_data = {
+                        "email_sent": email,
+                    }
 
             if allowed_redirect_to:
                 return self._do_redirect(
```
fantix, line 1588: At line 1598:

```python
        except Exception as ex:
            redirect_on_failure = data.get(
                "redirect_on_failure", data.get("redirect_to")
            )
            if redirect_on_failure is None:
                raise ex
            else:
```

Because `redirect_on_failure` is always set, the error response will always be a redirect response, which is different than the success path when `redirect_to` is not set. Maybe here `magic-link/email` should borrow the same `Accept: application/json` logic as in `magic-link/register` above:

```python
            if request_accepts_json:
                raise ex
```

scotttrinh: I think in this specific flow (which is the link that you get when signing in or up with a Magic Link), it's always going to be someone following a link from an email, so returning JSON can never make sense since the browser won't want a JSON error. Or maybe I'm misunderstanding your suggestion?

scotttrinh: Oh, I've misunderstood the structure here, this is not the handler for link's URL, this is the handler for sending the link! I agree we should do what you've suggested here. Will fix!

scotttrinh: @fantix ⚡ [`fb8fff3` (#8913)](https://github.com/geldata/gel/pull/8913/commits/fb8fff3ce3ff96ac1813eac628c0335343f073cf)

fantix (approved):

PR #8918, opened 2025-08-05
Add apply_access_policies_pg_default flag to Role

msullivan:
This allows an individual Role to override the per-branch
configuration of apply_access_policies_pg. The point of this is so
that we can make apply_access_policies_pg the *default*, but still
make it possible to create a role where the policies aren't applied.

This is mainly for the purpose of integration with external tools that
don't provide some sort of way to inject arbitrary config commands to
the connection estabalishment.

### DIFF
--- a/edb/buildmeta.py
+++ b/edb/buildmeta.py
@@ -57,7 +57,7 @@ from edb.common import verutils
 # The merge conflict there is a nice reminder that you probably need
 # to write a patch in edb/pgsql/patches.py, and then you should preserve
 # the old value.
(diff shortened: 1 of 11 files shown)


aljazerzen (approved):

edb/server/compiler/dbstate.py
```diff
@@ -747,6 +747,7 @@ class ParsedDatabase:
         "server_version_num": cast(
             SQLSetting, (defines.PGEXT_POSTGRES_VERSION_NUM,)
         ),
+        # "apply_access_policies_pg": ("true",),
```
dnwpark: ?

PR #8926, opened 2025-08-06
Add `administer rebuild_sql_introspection` command

aljazerzen:
Gel server supports updates of the backend postgresql version, without
any modification to contents of the database. That is, gel-server can
be shutdown, PostgreSQL database upgraded and then gel-server can
connect to the upgraded database without any additional steps.

Except SQL introspection views need to be rebuilt. That's because they
depend on the PostgreSQL version.

So this PR introduces `administer rebuild_sql_introspection()` command,
that does just that. It is idempotent (it can be run multiple times)
and should be run any time the version of the backend databases changes.

### DIFF
--- a/edb/server/compiler/ddl.py
+++ b/edb/server/compiler/ddl.py
@@ -1431,17 +1431,15 @@ def administer_repair_schema(
     )
 
 
-def administer_rebuild_sql_introspection(
+def administer_fixup_backend_upgrade(
     ctx: compiler.CompileContext,
     ql: qlast.AdministerStmt,
 ) -> dbstate.BaseQuery:
     if ql.expr.args or ql.expr.kwargs:
         raise errors.QueryError(
-            'rebuild_sql_introspection() does not take arguments',
+            'fixup_backend_upgrade() does not take arguments',
             span=ql.expr.span,
         )
-
-    current_tx = ctx.state.current_tx()
 
     from edb.pgsql import metaschema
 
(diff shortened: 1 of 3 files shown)


aljazerzen:
I didn't add any tests, because testing this is hard:
- we need to bootstrap an instance on an old PostgreSQL,
- we do `pg_dumpall` from PostgreSQL,
- we spin-up a new PostgreSQL, with a newer version, and apply the dump,
- we point gel-server at the new PostgreSQL,
- we try to do a pg_dump over gel-server SQL adapter, which should thoroughly test the SQL introspection.

I've run this on my machine, but I won't add tests for it.

edb/server/compiler/ddl.py
```diff
@@ -1431,6 +1431,41 @@ def administer_repair_schema(
...
+    cmd_group = pg_dbops.CommandGroup()
+    cmd_group.add_commands(cmds)
+    cmd_group.generate(block)
+
+    assert block.is_transactional()
+
+    return dbstate.DDLQuery(
+        sql=block.to_string().encode('utf-8'),
+        user_schema=current_tx.get_user_schema_if_updated(),
+        global_schema=current_tx.get_global_schema_if_updated(),
```
msullivan: Drop these?

msullivan (approved):
Should we name it something more general, like `fixup_postgres_backend_update()`?
(I don't love that but it's ok)

aljazerzen:
> Should we name it something more general, like fixup_postgres_backend_update()?

I'm using `fixup_backend_upgrade`

PR #8960, opened 2025-08-19
sql: Fix COLLATE clauses that require quotation

msullivan:
We were failing to quote the name of the collation.

### DIFF
--- a/edb/pgsql/codegen.py
+++ b/edb/pgsql/codegen.py
@@ -1019,7 +1019,7 @@ class SQLSourceGenerator(codegen.SourceGenerator):
 
     def visit_CollateClause(self, node: pgast.CollateClause) -> None:
         self.visit(node.arg)
-        self.write(f' COLLATE {node.collname}')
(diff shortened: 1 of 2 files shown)


edb/pgsql/parser/ast_builder.py
```diff
@@ -685,7 +685,7 @@ def _build_param_ref(n: Node, c: Context) -> pgast.ParamRef:
 def _build_collate_clause(n: Node, c: Context) -> pgast.CollateClause:
     return pgast.CollateClause(
         arg=_build_base_expr(n['arg'], c),
-        collname='.'.join(_list(n, c, 'collname', _build_str)),
+        collname=_list(n, c, 'collname', _build_str),
```
msullivan: Should we make it a tuple, like we do for most other names?

aljazerzen: Sure. But what's really the difference?

msullivan: Just mutability

PR #8961, opened 2025-08-19
Fix handling of SQL syntax error over gel proto

aljazerzen:
When string normalization was disabled, handling crashed.


msullivan (approved):
Is there some way to handle this at the source of the issue?

aljazerzen:
Yes, that would be much better, thanks for spotting.

msullivan (approved):

Issue #8974, opened 2025-08-21
Aliases/globals have `expr` duplicated in a `Type`, and that sometimes get recompiled


msullivan:
In particular, it seems that *both* can get triggered by `self._propagate_if_expr_refs`, which causes some weird behavior.

PR #8975, opened 2025-08-21
Work around for bug #8974 when updating extension globals using secrets

msullivan:
This is needed to backport #8953 and #8964, but I don't feel great
about it.

### DIFF
--- a/edb/schema/types.py
+++ b/edb/schema/types.py
@@ -3082,10 +3082,14 @@ class TypeCommand[TypeT: Type](sd.ObjectCommand[TypeT]):
         value: s_expr.Expression,
         track_schema_ref_exprs: bool=False,
     ) -> s_expr.CompiledExpression:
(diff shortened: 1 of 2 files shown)


dnwpark (approved):
:smiling_face_with_tear:

edb/schema/types.py
```diff
@@ -3082,10 +3082,14 @@ def compile_expr_field(
         value: s_expr.Expression,
         track_schema_ref_exprs: bool=False,
     ) -> s_expr.CompiledExpression:
+        # XXX: This seems like pointless duplication of work from
+        # globals/aliases... why is expr even here?
+        # (... because we export it in the introspection schema)
```
aljazerzen, line 3087: Oh. We cannot remove it unless we want to break introspection. Oh no

msullivan: We can maybe make it a computed?

PR #8978, opened 2025-08-22
Add a fake test to enable populating db with ORM tests from gel-python

msullivan:
You can do something like:
edb inittestdb -k test_orm --update --tests-dir misc/fake-tests/

### DIFF
--- a/misc/fake-tests/test_orm.py
+++ b/misc/fake-tests/test_orm.py
@@ -1,0 +1,126 @@
+#
+# This source file is part of the EdgeDB open source project.
+#
+# Copyright 2012-present MagicStack Inc. and the EdgeDB authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


misc/fake-tests/test_orm.py
```diff
@@ -0,0 +1,44 @@
...
+'''
+
+import os.path
+import pathlib
+
+from edb.testbase import server as tb
+
+
+TEST_DIR = (
+    pathlib.Path(__file__).parent.parent.parent.parent
```
dnwpark, line 32: This seems kinda suspicious.

Also, after [this PR](https://github.com/geldata/gel-python/pull/864), the gel-python tests support more than just the one schema.

msullivan: It sounds to me like you are volunteering to produce an inittestdb for gel-python? :P

dnwpark: Yeah sure, why not xd

msullivan: I think that Elvis is trying to make it actually use `edb test`

PR #8982, opened 2025-08-25
Add math::exp and math::e

Gobot1234:


docs/reference/stdlib/math_funcops_table.rst
```diff
@@ -12,6 +12,9 @@
     * - :eql:func:`math::floor`
       - :eql:func-desc:`math::floor`
 
+    * - :eql:func:`math::exp`
+      - :eql:func-desc:`math::exp`
+
```
scotttrinh, line 17: You'll also need to add the descriptions to https://github.com/Gobot1234/edgedb/blob/64144e0dabb33ee9234712747cbc7000b7950a41/docs/reference/stdlib/math.rst this is just a kind of table of contents that gets included.

tests/test_edgeql_functions.py
```diff
@@ -5872,15 +5878,40 @@ async def test_edgeql_functions_math_exp_03(self):
             abs_tol=1e-5,
         )
 
-        with self.assertRaises(edgedb.NumericOutOfRangeError):
+        async with self.assertRaisesRegexTx(
+            edgedb.NumericOutOfRangeError, 'value out of range: overflow'
+        ):
             await self.con.query(r'''SELECT math::exp(1000);''')
```
Gobot1234, line 5884: any idea why this wasn't throwing the error properly before this? was it because of the change in assert_data_shape?

aljazerzen: It was throwing an error and `with` correctly caught that error. The problem was that transaction for this test case was aborted so subsequent interactions with the database failed.

It's a thing that is very specific to how our test suite runs.

Gobot1234: How do the tests after this run then for future reference?

aljazerzen: `assertRaisesRegexTx` restarts the transaction: https://github.com/edgedb/edgedb/blob/master/edb/testbase/server.py#L915-L926

aljazerzen: Pretty gnarly, a tiny bit too magic for my taste :D

Gobot1234: Ah right that makes sense. I also don't like that lol

msullivan: It doesn't "restart" the transaction, exactly, but it runs the body in a "nested" transaction that it rolls back.

(We don't really support nested transactions, though. The test suite's client simulates them with `SAVEPOINT`s; this is a test-suite special, though--the main gel-python client doesn't support it)

PR #8987, opened 2025-08-26
Fix some new simple_scoping + inline functions regressions

msullivan:
Another bug introduced just now in #8985 and caught by my in-progress
PR to rip out old scoping.


edb/edgeql/compiler/stmtctx.py
```diff
@@ -653,6 +653,11 @@ def _collapse_factoring_protected(
         if not (parent := node.parent):
             continue
 
+        # If the underlying thing has already been factored fully, we
+        # skip it, because it might be no-factor fenced?
```
dnwpark, line 657: hmm....

PR #8988, opened 2025-08-27
Remove path factoring

msullivan:
This makes simple_scoping the only supported behavior.

This rips out some of the machinery for it, but most of it is still
there for now, because the scope tree is still pretty fundamental and
a bunch of internal codegen stuff actually still relies on enable path
factoring for edgeql AST that it generates.

First step is ripping that out.


dnwpark (approved):
:tada:

PR #8991, opened 2025-08-28
Enable FOR OPTIONAL

msullivan:
I think it's likely to be in decent shape, actually.
A big part of the original disabling was that the first
implementation didn't work on object types, but that
has been fixed.


aljazerzen:
I'll merge and backport it.

PR #9000, opened 2025-09-05
Loosen Magic Code register/email data requirements

scotttrinh:
Code-based flows do _not_ require PKCE sessions. Also fixed a bunch of issue we had with unbound variables being referenced in the catch branch of the exceptions handling code.


fantix (approved):
LGTM!

PR #9002, opened 2025-09-08
Docs for 7.x features

aljazerzen:
- **docs: permissions**
- **docs: SQL COPY**
- **docs: Role.branches**
- **docs: Index.create_concurrently**

### DIFF
--- a/docs/reference/datamodel/permissions.rst
+++ b/docs/reference/datamodel/permissions.rst
@@ -187,17 +187,20 @@ Public readonly database
 ------------------------
 
 Gel server can be exposed to public internet, with clients connecting directy
-from browsers. In such scenarios, it is recommended to create a separate role
-that will be used by the JavaScript client (e.g. ``webapp``) and give it
-a restricted set of permissions.
+from browsers. Let's assume that only want to grant read access to the public
+browser client.
 
-For example, if the webapp only needs read access and no data modification,
-it doesn't need any permissions set. This way, it will not be able to issue
-``DROP TYPE`` or ``DELETE`` commands, but will be able to read all data in the
-database.
+In such scenarios, it is recommended to create a separate role
+that will be used by the JavaScript client (e.g. ``webapp``) and not grant it
+any permissions.
 
-If we want to limit that access further, for example read access to type
-``Secrets``, we can use this schema:
+This way, it will not be able to issue ``DROP TYPE`` or ``DELETE`` commands,
+but will be able to read all data in the database. More importantly, it will
+not be able to configure ``apply_access_policies`` to ``false`` to bypass
+our restrictions.
+
+If we want to limit that access further, for example limit read access to type
+``Secrets``, we can use such schema:
 
 .. code-block:: sdl
 
@@ -210,9 +213,10 @@ If we want to limit that access further, for example read access to type
     };
 
 
-Now, ``webapp`` will not possess permission ``server_access``, which is
-required to read or modify ``Secret``. To be able to use the ``Secrets``, we
-have use super user role, or some other role with ``server_access`` permission:
+Because ``webapp`` role will not possess permission ``server_access`` it will
+not be able to read (or modify) ``Secret``. For other, trusted clients, which
+should be able to access ``Secrets``, we have use *superuser* role, or some
+other role with ``server_access`` permission:
 
 .. code-block:: edgeql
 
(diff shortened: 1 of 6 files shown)


aljazerzen:
Depends on: https://github.com/geldata/edgedb.com/pull/1032

docs/reference/datamodel/permissions.rst
```diff
@@ -0,0 +1,184 @@
...
+:ref:`global variables<ref_datamodel_globals>` of the same name:
+
+.. code-block:: edgeql
+
+    select global sys::perm::data_modification;
+
+Roles that are qualified as *superuser* are implicitly granted all permissions.
+
+Some commands are currently limited to *superuser* roles only. These include
+commands to manage branches, extensions and roles.
```
dnwpark: We probably need a more complete list of these somewhere.

docs/reference/datamodel/permissions.rst
```diff
@@ -0,0 +1,184 @@
...
+
+    alter role warehouse {
+      set permissions := {default::data_export};
+    };
+
+.. note::
+
+    Role's permissions are applied to all branches in the instance.
+    If branches are using different schemas, make sure that names of permissions
+    do not overlap.
```
scotttrinh: We probably want to surface that roles are cross-branch a bit closer to the top to help people understand how/where roles are in the hierarchy of instances and branches. This note can stay as-is, it'll just be a reminder and caution.

docs/reference/datamodel/permissions.rst
```diff
@@ -0,0 +1,184 @@
...
+:eql:synopsis:`ext::auth::perm::auth_read`
+
+:eql:synopsis:`ext::auth::perm::auth_write`
+
+:eql:synopsis:`ext::auth::perm::auth_read_user`
+
+
+Permissions for :ref:`ai <ref_ai_extai_reference>` extension:
+
+:eql:synopsis:`ext::ai::perm::provider_call`
```
dnwpark: I added some descriptions for the ai permissions in `extai.rst`. Some link to/from here and there maybe?

docs/reference/datamodel/permissions.rst
```diff
@@ -0,0 +1,184 @@
...
+            allow select
+            using (global data_export);
+
+        access policy anyone_can_insert
+            allow insert;
+    }
+
+In this example, we have type ``AuditLog`` into which all roles are allowed to
+insert new log entries. But reading is allowed only to roles that posses
+``data_export`` permission (or are qualified as a *superuser*).
```
scotttrinh, line 172: We should add other common examples and recipes. Here are some off-the-cuff ideas:

- limiting developer or BI access to the built-in UI
- client-side data access (serverless, js-client)
- CI role (single branch?)

dnwpark (approved):

docs/reference/datamodel/permissions.rst
aljazerzen: Good ideas. Let's do them is a separate PR.

aljazerzen: btw, do you have a todo list for docs?

scotttrinh: No real todo list, we either do them right away or they just don't get done until someone asks about something and we have the time right in that moment to pull it off. I'll see if we can add some more structure there to keep us from losing this.

aljazerzen: @scotttrinh I've added more examples. I don't think this covers all you've suggests, but to be honest, I cannot pin down what's missing.

scotttrinh: Yeah, this is roughly fine. I think the web app example needs some thought about how it interacts with `ext::auth` to provide more granular access control, but this is already useful.

aljazerzen: That's almost orthogonal - permissions are for controlling access of gel *clients*, while `ext::auth` is for controlling access of *users*.

These is some correlation in that both use access policies to exert that control and that access policies can be disabled if client has `configure_apply_access_policies`.

But yes, these space for this to be expanded.

scotttrinh: For posterity more than argument, I think the use case I have in mind is a completely backendless application where each user is given a role (along with an `ext::auth::Identity`) and you are making direct queries to the database from your client-side JavaScript (or Swift/Kotlin).

aljazerzen: Ohh, right - I'm having a hard time getting out of "there must be a trusted backend somewhere" mindset. It is actually possible to have just the client frontend app.

yea, we have to add an example for this.

PR #9012, opened 2025-09-09
Return json from magic code authenticate endpoint if no callback_url is given

jaclarke:
Match the behaviour of the email+password authenticate endpoint, and return a json response if the `callback_url` is not passed to the magic code authenticate endpoint.


edb/server/protocol/auth_ext/http.py
```diff
@@ -1753,10 +1759,17 @@ async def handle_magic_link_authenticate(
                     datetime.datetime.now(datetime.timezone.utc),
                 )
 
-                return self._try_redirect(
-                    response,
-                    util.join_url_params(callback_url, {"code": auth_code}),
-                )
+                response_dict = {"code": auth_code}
+
+                if maybe_callback_url:
```
scotttrinh, line 1772: Mostly a note for the future: I think the better approach would be to check the accept header and vary the required request body parameters based on whether the request accepts JSON or not. But what we have here matches our current defensive strategy and I wouldn't want to change this without making it a wider-reaching change across the whole auth server.

PR #9016, opened 2025-09-11
Minimize FlatSchema interface

aljazerzen:
Goal of this PR is twofold:
- improve performance of `s_schema.Schema` data structure,
- minimize its interface,

The idea is to move as much behavior out of the data structure itself into helper functions around it. For example:
- `def get()` performs name resolution (which includes application of module aliases, implicit prefixing with `std` and a few more things). If this would be replaced with a simpler:
  - `def get_by_qualified_name(sn.Name)` and
  - `get_by_shortname(str, SchemaClass)` and
  - `get_by_globalname(str, SchemaClass)`
  Now most calls of `get` could be replaced with more specific, faster function. Additionally, name resolution could be decoupled in a separate `def lookup` function.
- A few functions are too specific (`get_functions`) and could be generalized.
- Functions should refrain from applying default values, they could return `None` instead.
- Functions should refrain from throwing user-facing exceptions, this could be handled by helper functions instead.

When I say "helper functions", these can be standalone functions that take schema as an arg, but they could also be methods on the Schema abstract class. The important part is that they are not implemented for `ChainedSchema` and `FlatSchema` separately.

I suspect that the PR will include a lot of small incremental refactors, slowly porting one method after another.

- **Remove disallow_module**
- **Revert "Remove disallow_module"**
- **Remove support for __current__**

### DIFF
--- a/edb/edgeql/compiler/expr.py
+++ b/edb/edgeql/compiler/expr.py
@@ -913,27 +913,27 @@ def _infer_type_introspection(
     span: Optional[parsing.Span]=None,
 ) -> s_types.Type:
     if irtyputils.is_scalar(typeref):
-        return env.schema.fetch(
+        return env.schema.get_by_name(
             'schema::ScalarType', type=s_objtypes.ObjectType
         )
     elif irtyputils.is_object(typeref):
-        return env.schema.fetch(
+        return env.schema.get_by_name(
             'schema::ObjectType', type=s_objtypes.ObjectType
         )
     elif irtyputils.is_array(typeref):
-        return env.schema.fetch(
+        return env.schema.get_by_name(
             'schema::Array', type=s_objtypes.ObjectType
         )
     elif irtyputils.is_tuple(typeref):
-        return env.schema.fetch(
+        return env.schema.get_by_name(
             'schema::Tuple', type=s_objtypes.ObjectType
         )
     elif irtyputils.is_range(typeref):
-        return env.schema.fetch(
+        return env.schema.get_by_name(
             'schema::Range', type=s_objtypes.ObjectType
         )
     elif irtyputils.is_multirange(typeref):
-        return env.schema.fetch(
+        return env.schema.get_by_name(
             'schema::MultiRange', type=s_objtypes.ObjectType
         )
     else:
@@ -951,7 +951,7 @@ def compile_Introspect(
         typeref = typeref.material_type
     if typeref.is_opaque_union:
         typeref = typegen.type_to_typeref(
-            ctx.env.schema.fetch(
+            ctx.env.schema.get_by_name(
                 'std::BaseObject', type=s_objtypes.ObjectType
             ),
             env=ctx.env,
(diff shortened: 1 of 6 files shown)


edb/edgeql/compiler/expr.py
```diff
@@ -985,10 +983,14 @@ def _infer_index_type(
     node_type = setgen.get_expr_type(expr, ctx=ctx)
     index_type = setgen.get_set_type(index, ctx=ctx)
 
-    str_t = env.schema.get('std::str', type=s_scalars.ScalarType)
-    bytes_t = env.schema.get('std::bytes', type=s_scalars.ScalarType)
-    int_t = env.schema.get('std::int64', type=s_scalars.ScalarType)
-    json_t = env.schema.get('std::json', type=s_scalars.ScalarType)
+    str_t = env.schema.get_by_name(sn.QualName('std', 'str'))
+    assert isinstance(str_t, s_scalars.ScalarType)
```
msullivan: I don't think this change is an improvement -- it's more verbose and produces worse messages on failure. (Though that won't happen in these cases, I suppose).

Though it looks like `get` is preserved, so this isn't necessary anyway?

aljazerzen: Yes, this is net-negative change, I'll revert it.

I've changed this in preparation to renaming `.get()` to `.lookup()`. The idea is that:
- `.get_by_*()` methods are fast, low-level and simple: they do not have `default` param, `str` to `sn.Name` parsing, and do not raise exceptions,
- `.lookup()` method would be the full name resolution, which would be moved out into the ql compiler,

I guess we need middle ground, for cases like this one, where we need convenience of `str` to `sn.Name` parsing, but don't do full name resolution.

edb/schema/schema.py
```diff
@@ -117,7 +116,57 @@
...
+    Objects can be retrieved by:
+    - id,
+    - name (fully qualified),
+    - shortname (only for function and operator class),
+    - references.
+
+    A bit of terminology:
+    - `get_*` methods are for low-level data retrieval,
+    - `fetch_*` methods are wrappers for `get_` with nicer interface (name
+       parsing, default values, raising exceptions),
```
msullivan: I would maybe reverse `get` and `fetch`, since `get` is so heavily used now?

aljazerzen: I've tried to do that but:
- then we'd have to rename `get_obj_data_raw`, `get_casts` and `get_referrers` and similar less used methods,
- we'd still have to rename `get` to `lookup` (which is the majority of work)
- `get` really feels like the "lowest level operation", because if feels like it is just a getter. `fetch` feels like doing some work during the retrieval.

We could reverse `lookup` and `get`, but this feels very wrong.

Current situation is not good, because when anyone needs to get something from schema, the first thing they try is `get`, which is the most expensive of them all. They should be trying `get` and then realizing that `fetch` is the more convenient.

aljazerzen: I wish we could just use rust-analyzer and "refactor > rename" all the methods.

msullivan (approved):

PR #9049, opened 2025-09-25
Fix some issues with overloading AI indexes

### LINKED ISSUE #9048: AI index names can get out of sync between pg and gel, confusing ANALYZE
The names are assigned based on the first `ai` index to be placed on a type. If another overriding index is declared on a subtype, the pg index is not renamed. If the first index is then deleted, the pg index name won't match anything in the database, and ANALYZE won't be able to resolve it to a name.

This is what's happening in https://github.com/geldata/gel/pull/9046

It can be triggered via:
```
    async def test_ext_ai_indexing_05f(self):
            await self.con.execute('''
                alter type Astronomy
                drop index
                  ext::ai::index(embedding_model := 'text-embedding-test')
                on (.content);
            ''')

            qv = [1.0, -2.0, 3.0, -4.0, 5.0, -6.0, 7.0, -8.0, 9.0, -10.0]

            await self._assert_index_use(
                f'''
                with vector := <array<float32>>$0
                select ext::ai::search(OnExpression, vector) limit 5;
                ''',
                qv,
            )
```

msullivan:
When the "effective" index changes, we need to rename the pg index, so
ANALYZE won't get confused.

We also need to do a `refresh` when the effective index changes due to
a delete index, like we do for creates, since that will invalidate the
vectors.

Fix this by adding the new refresh call and making refresh rename the
index.

In PR #2474, I yanked out all the rename dbops, and remarked:
> If something changes in the future and we need to generate a pgsql
> rename, we can pull the old dbops code out of git.

Fixes #9048.

### DIFF
--- a/edb/pgsql/dbops/ddl.py
+++ b/edb/pgsql/dbops/ddl.py
@@ -366,6 +366,23 @@ class CreateObject(SchemaObjectOperation):
             block.add_command(mdata.code_with_block(block))
 
 
+class RenameObject(SchemaObjectOperation):
+    def __init__(self, object, *, new_name, **kwargs):
+        super().__init__(name=object.name, **kwargs)
+        self.object = object
+        self.altered_object = object.copy()
+        self.altered_object.rename(new_name)
+        self.new_name = new_name
+
+    def generate_extra(self, block: base.PLBlock) -> None:
+        super().generate_extra(block)
+        # if self.object.metadata:
+        #     breakpoint()
+        #     mdata = UpdateMetadata(
+        #         self.altered_object, self.altered_object.metadata)
+        #     block.add_command(mdata.code_with_block(block))
+
+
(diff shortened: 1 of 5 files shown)


edb/pgsql/dbops/ddl.py
```diff
@@ -366,6 +366,23 @@ def generate_extra(self, block: base.PLBlock) -> None:
...
+    def __init__(self, object, *, new_name, **kwargs):
+        super().__init__(name=object.name, **kwargs)
+        self.object = object
+        self.altered_object = object.copy()
+        self.altered_object.rename(new_name)
+        self.new_name = new_name
+
+    def generate_extra(self, block: base.PLBlock) -> None:
+        super().generate_extra(block)
+        # if self.object.metadata:
```
elprans, line 381: Stray

elprans (approved):

PR #9057, opened 2025-09-30
6.11 changelog

jackfischer:
I just asked Claude code to explore the git history to make the changelog and it wrote this.


aljazerzen:
This looks quite good. I've removed a few points, because they are non-user facing, and reworded the others.

aljazerzen:
We do have script for this that takes a git range and produces the changelog, but the main work I need to put in is translate git-commit-messages into user-facing terms.

PR #9064, opened 2025-10-03
Fix order of declaration issue when declaring computed with `using`

### LINKED ISSUE #9030: Order of declaration bug when having a global and an extended form computed
Seems like using the _computed_ extended form in a type after a global that references the type triggers a weird bug where it doesn't find the type's properties.

I've reduced it as much as i could. Schema:

```edgeql
module default {

  # Global referencing same type declared before
  global some_user := ((
    select User limit 1
  ));

  type User {
    required foo: bool;

    # Bug
    required computed_bug: bool {
      using (
        .foo
      # ^^^^ error: object type 'default::User' has no link or property 'foo'
      );
    }

    # These work
    required computed_working_1 := .foo;  # Non-extended form
    required computed_working_2 { # No explicit type
      using (
        .foo
      );
    }
  }

  # Moving global to the end (or to another .gel module) also fixes it
  # global some_user := ((
  #   select User limit 1
  # ));

}
```

Couldn't find a similar issue but maybe didn't search right.

- Gel Version: 6.10+b9cb801 (also happens in 6.9)
- Gel CLI Version: 7.9.0+070b371
- OS Version: Fedora release 42 (Adams)

Steps to Reproduce:

1. Use provided schema
2. See LSP complain or try to create a migration with `gel migration create`

msullivan:
We need to look for an `expr` SetField, and track that as a
dependency.

Fixes #9030.

### DIFF
--- a/edb/edgeql/declarative.py
+++ b/edb/edgeql/declarative.py
@@ -673,7 +673,7 @@ def _trace_item_layout(
             target_expr: Optional[qlast.Expr]
             if isinstance(decl.target, qlast.TypeExpr):
                 target = _resolve_type_expr(decl.target, ctx=ctx)
-                target_expr = None
+                target_expr = _get_expr_field(decl)
             else:
                 target = None
                 target_expr = decl.target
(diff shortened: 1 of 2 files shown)


aljazerzen (approved):

aljazerzen:
Backport?

PR #9065, opened 2025-10-03
Docs for PreferRepeatableRead

aljazerzen:
Since this is *virtual*, client side option, it has to be explained for
each client separately.


msullivan (approved):
Maybe add a note to `serialzable` about potentially perf issues?

PR #9068, opened 2025-10-03
Warning on order by .id without empty last

jackfischer:
Motivation: This has been a very common problem for us and today I realized it's also the culprit in https://github.com/geldata/gel/issues/8035 . I assume there is a bigbrain reason the `empty last` can't be assumed and silently injected and that this is hard to fix to not require `empty last` at all.

Obviously I did not write this, this is written by you-know-who (claude)

> [IR-first detection] in _is_sort_on_id()
> - Check direct pointer via ir_sortexpr.expr if it is an irast.Pointer.
> - Fallback to ir_sortexpr.path_id.rptr().
> 
> [Minimal syntactic fallback]
> - If IR checks don’t match, conservatively check if the qlast path’s terminal step is literally id.
> - Keeps logic robust without spelunking deeply into AST/IR structures.

Considered moving the code to a better place and probably splitting it up

- Create is_set_on_id(ir_set: irast.Set) -> bool in pathctx.py
- Small qlast helper to edb/edgeql/astutils.py

but not doing anything like that yet unless we want to

### DIFF
--- a/edb/edgeql/compiler/clauses.py
+++ b/edb/edgeql/compiler/clauses.py
@@ -141,14 +141,14 @@ def compile_orderby_clause(
                     ctx=exprctx)
                 if len(matched) != 1:
                     sort_type_name = schemactx.get_material_type(
-                        sort_type, ctx=ctx
-                    ).get_displayname(env.schema)
+                        sort_type, ctx=ctx).get_displayname(env.schema)
                     if len(matched) == 0:
                         raise errors.QueryError(
                             f'type {sort_type_name!r} cannot be used in '
                             f'ORDER BY clause because ordering is not '
                             f'defined for it',
                             span=sortexpr.span)
+
                     elif len(matched) > 1:
                         raise errors.QueryError(
                             f'type {sort_type_name!r} cannot be used in '
--- a/edb/edgeql/compiler/clauses.py
+++ b/edb/edgeql/compiler/clauses.py
@@ -42,9 +42,11 @@ from edb.ir import typeutils as irtyputils
 def _is_sort_on_id(ir_sortexpr: irast.Set, sortexpr: qlast.SortExpr) -> bool:
     """Return True if the sort expression is ordering by the builtin .id.
 
-    Uses IR-level detection only:
+    Primarily uses IR-level detection:
     - direct pointer in ir_sortexpr.expr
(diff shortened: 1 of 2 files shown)


edb/edgeql/compiler/clauses.py
```diff
@@ -36,6 +36,44 @@
...
+
+    # Syntactic fallback: if the original qlast path ends with `.id`.
+    steps = getattr(getattr(sortexpr, 'path', None), 'steps', None)
+    if steps:
+        last = steps[-1]
+        name = getattr(last, 'ptr', None) or getattr(last, 'name', None)
+        if hasattr(name, 'name'):
+            name = name.name
+        if isinstance(name, str) and name == 'id':
+            return True
```
msullivan, line 74: ... this is wildly atrocious but probably works.

msullivan (changes requested):
(I'm marking this as Request changes, but we'll take it over from here)

Oh, argh, this is kind of nasty.

The heart of the issue is that we default to `empty first` while postgres defaults to `NULLS LAST`. I'm not sure why that's the decision we made.

For *indexes*, we specify `NULLS FIRST` when we create them, but when declaring `UNIQUE` constraints, we can't.

The right thing to do would probably be to use `CREATE UNIQUE INDEX` to implement exclusive constraints, since that would let us specify `NULLS FIRST`.
The complication here is that we now support in-place upgrades, and this would require rebuilding indexes. We'll need to think about it.
I think we *probably* could make it happen?

I think that the best approach for now would be to detect `ORDER BY` on all `required` properties that have `exclusive` constraints and inject `NULLS LAST`.

edb/edgeql/compiler/clauses.py
jackfischer: It first came up with

```python
rptr = getattr(ir_sortexpr, 'expr', None)
if (
    isinstance(rptr, irast.Pointer)
    and irtyputils.is_id_ptrref(rptr.ptrref)
    and sortexpr.nones_order is None
):
```

Alas that was too good to be true.

msullivan: That was certainly *closer* than what it ended up coming up with.
I also wonder why it is so excited to use `getattr`

PR #9073, opened 2025-10-06
[7.x] Fix in-place upgrade repairs

msullivan:
The repair needs to be done against the *new* schema for stuff like a
new field in the AI extension to get picked up.

(I desperately hate the annotation copying thing---it has caused so
much grief.)

This probably needs to be forward-ported.

### DIFF
--- a/edb/graphql/compiler.py
+++ b/edb/graphql/compiler.py
@@ -29,9 +29,9 @@ from graphql.language import lexer as gql_lexer
 
 
 def _get_gqlcore(
-    std_schema: s_schema.FlatSchema,
-    user_schema: s_schema.FlatSchema,
-    global_schema: s_schema.FlatSchema,
+    std_schema: s_schema.Schema,
+    user_schema: s_schema.Schema,
+    global_schema: s_schema.Schema,
 ) -> graphql.GQLCoreSchema:
     return graphql.GQLCoreSchema(
         s_schema.ChainedSchema(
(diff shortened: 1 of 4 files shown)


edb/graphql/compiler.py
```diff
@@ -29,9 +29,9 @@
 
 
 def _get_gqlcore(
-    std_schema: s_schema.FlatSchema,
-    user_schema: s_schema.FlatSchema,
-    global_schema: s_schema.FlatSchema,
+    std_schema: s_schema.Schema,
+    user_schema: s_schema.Schema,
+    global_schema: s_schema.Schema,
```
aljazerzen, line 34: I've also made this change, for completely unrelated reasons, in #9016

edb/schema/schema.py
```diff
@@ -2068,15 +2068,21 @@ def _get_global(
...
-            return self._global_schema.get_global(  # type: ignore
+        obj: Optional[so.Object_T]
+        if (
+            issubclass(objtype, so.GlobalObject)
+            and self._global_schema is not EMPTY_SCHEMA
+        ):
+            obj = self._global_schema.get_global(  # type: ignore
+                objtype, name, default=None)
+            if obj is not None:
+                return obj
```
aljazerzen, line 2079: Why is this necessary? Do we have `so.GlobalObject`s that are in `_top_schema` or `_base_schema`?

This is how I've done in https://github.com/geldata/gel/pull/9016 

https://github.com/geldata/gel/blob/a18ca332d1b657285f2f5be2f3800b666544bc9e/edb/schema/schema.py#L1864-L1875

msullivan: Not before this PR, but in this PR I set up a ChainedSchema that has two layered global schemas

PR #9074, opened 2025-10-07
Fix creating branch when a schema has nested arrays

### LINKED ISSUE #8910: Can't create branch from `cards` database
```
_localdev:scope> create data branch scope2 from scope;
gel error: InvalidValueError: cannot determine Postgres OID of Gel schema::Array('96527047-6f04-11f0-a18e-3b68071c21e2') with element type 96526cfc-6f04-11f0-975c-35b65f94f3a9
_localdev:scope> create schema branch scope2 from scope;
gel error: InvalidValueError: cannot determine Postgres OID of Gel schema::Array('96527047-6f04-11f0-a18e-3b68071c21e2') with element type 96526cfc-6f04-11f0-975c-35b65f94f3a9
```

msullivan:
We can't compute backend_ids for nested arrays, but we also don't need
to, since they don't actually get passed in as arguments.

Fixes #8910.

### DIFF
--- a/edb/server/bootstrap.py
+++ b/edb/server/bootstrap.py
@@ -2182,6 +2182,7 @@ def compile_sys_queries(
         FILTER
             NOT (.abstract ?? False)
             AND NOT (.transient ?? False)
+            AND NOT .element_type IS schema::Array
         SET {
             backend_id := sys::_get_pg_type_for_edgedb_type(
                 .id,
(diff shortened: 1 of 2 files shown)


aljazerzen (approved):

aljazerzen:
Do we need a patch for this?

msullivan:
Yeah, this will need a patch to get it to work right. In particular, we need to regenerate `sysqueries`.
Doing an empty `edgeql+schema` patch should do it, I think.
That's a little heavyweight, but won't require any new features in the patching system.

PR #9077, opened 2025-10-07
Fix in-place upgrade repairs

msullivan:
The repair needs to be done against the *new* schema for stuff like a
new field in the AI extension to get picked up.

(I desperately hate the annotation copying thing---it has caused so
much grief.)

This is the forward-ported version of #9073.

### DIFF
--- a/.github/scripts/patches/compute-ipu-versions.py
+++ b/.github/scripts/patches/compute-ipu-versions.py
@@ -42,8 +42,12 @@ for obj in data['packages'] + data_testing['packages']:
             base + obj['installrefs'][0]['ref'],
         ))
 
+prerelease_versions.sort(key=lambda x: x[0])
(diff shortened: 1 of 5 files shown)


msullivan:
I think what I *want* to do is declare that after 7.0-final, we are done with extension patching via the patch system, and we will version extensions via the actual extension versioning system!

Then, we will not allow IPUs to 8.x from 7.x prereleases, only from final versions.

Which hopefully will mean we won't need to understand 7.x patches further on!
Though maybe we'll want to check the patch log to see if we want to reject it, since we don't have a record of the actual latest version.

aljazerzen:
Yeah, the additional amount of work we have to do for patches of extensions is too large.

But extension versioning will be a pain in the ass too.

PR #9084, opened 2025-10-08
Rewrite Schema in Rust

aljazerzen:
This PR implements the Schema data structure in Rust.

It:
- implements the data structure in [gel-schema crate in gel-rust repo](https://github.com/geldata/gel-rust/compare/master...schema),
- adds a PyO3 wrapper crate that provides a `Schema` class,
- adds a Python class `RustSchema` that proxies method calls to the PyO3 class,
- replaces usages of `FlatSchema` with `RustSchema`.

Challenges:
- the `edb.schema.Schema` abstract class has a pretty large interface. In #9016, I've tried to push as many abstract methods from schema implementations up into Schema itself.
- data that we put in the schema can be of many different types (Python primitives, uuid, None, ObjectSet, ObjectList, Expressions, Version, ...), each of which needs a rust-native representation,
- all data we put in the schema is stored there in so-called "reduced representation" (a tuple of data that can be pickled). This means that the PyO3 class must produce precisely this repr, so it can be `schema_restore`-d later by getter methods.
- conversion between "Python reduced repr" and "Rust repr" is currently slow. That's because:
  - I'm importing classes just to check if a value is an instance of it,
  - some data needs to be copied (strings, uuids, lists, ...),

I've managed to get it so far that it compiles std lib, bootstraps and works on all queries that I've tried. Let's see the test suite.

Plan:
- [x] Cache imports using `static PyOnceLock`,
- [x] Reducing values is no longer needed, because serialization is now done with serde and bincode. This means that we should remove the "reduced repr" and make PyO3 class consume and produce the "normal" Python repr of values. I suspect that this will provide a huge speed-up.
- [ ] `ObjectList`, `ObjectSet`, `ObjectDict` and `ObjectIndex` are currently copied on each access. This means that if we want to, for example, lookup a Pointer of an ObjectType, we do `get_pointers()`, which copies all pointers from schema into a new `ObjectIndex` instance, just to pick a single `Uuid` out of it. This is so wrong that it feels unethical and immoral. To improve this, I want to store these values in schema in an `Rc`, so that when we retrieve a field value, we just clone the `Rc` and not the value itself. This means that eacn of these object containers would need its PyO3 wrapper class. A bit of work, but huge potential speed-up.

Current benchmark: sometimes this is 3x slower than master

time

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1174,9 +1174,9 @@ dependencies = [
 [[package]]
 name = "gel-schema"
 version = "0.1.0"
-source = "git+https://github.com/geldata/gel-rust.git?rev=e162dbbf22d05b785aaa31e03cbbe78296cdbe1e#e162dbbf22d05b785aaa31e03cbbe78296cdbe1e"
+source = "git+https://github.com/geldata/gel-rust.git?rev=05032c45444b8786cc455cfd45a9797283123021#05032c45444b8786cc455cfd45a9797283123021"
 dependencies = [
- "im-rc",
+ "im",
  "indexmap",
  "regex",
  "serde",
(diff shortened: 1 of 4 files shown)


aljazerzen:
One concerning data point is still the `compile_migration_01` benchmark, I have to investigate that.

PR #9085, opened 2025-10-09
Docs for AI agent

vpetrovykh:
Describing the gel-python generated QB API for the AI agents. The format is markdown with minimal explanation focusing on examples.

### DIFF
--- a/docs7/pydantic.md
+++ b/docs7/pydantic.md
@@ -1,0 +1,74 @@
+# Schema
+
+The schema used in examples. There are only two types representing characters and items for a simple game.
+```
+module default {
+    type Character {
+        required first_name: str {
+            constraint exclusive;
+        }
+
+        link equipped: Item;
+        multi link inventory: Item;
+    }
+
+    type Item {
+        required description: str {
+            default := 'Junk';
+        }
+
+        rightful_owner: Character {
+            constraint exclusive;
+        }
+    }
+}
+```
+
+# Reflected Pydantic Models
+
+The command `gel generate py/models` creates models that can be used by the Gel query builder. These models are also Pydantic models and so they have access to various Pydantic features like `model_dump()` or `model_validate()`, etc.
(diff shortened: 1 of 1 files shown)


docs7/qb.md
```diff
@@ -0,0 +1,359 @@
...
+icecream = default.Item(description='Chocolate Ice Cream')
+elsa = default.Character(
+    first_name='Elsa',
+    inventory=[
+        icecream,
+    ],
+    equipped=icecream,
+)
+icecream.rightful_owner = elsa
+db.sync(elsa)
```
dnwpark, line 90: not sure how AI will understand, but perhaps create also another item with elsa as the `rightful_owner` to show that the same logic holds when two different objects link the same target

docs7/qb.md
```diff
@@ -0,0 +1,359 @@
...
+```python
+junk = default.Item()
+assert junk.description != 'Junk'
+db.sync(junk)
+assert junk.description == 'Junk'
+```
+
+## Bulk write
+
+It is better to use `.save()` For writing data to Gel in bulk. This method avoids re-fetching overhead and will attempt to batch similar objects together.
```
dnwpark, line 119: Sync uses the same batching logic, so it should have the same batching behaviour.

docs7/qb.md
```diff
@@ -0,0 +1,359 @@
...
+db.save(alice)
+print(alice.id)  # the id is populated by save
+```
+
+Use `.sync()` method when you need to not only save the model changes, but also to fetch the updated state from the database. This method will populate the new `id` values as well as any server-side defaults in case of creating objects.
+
+In our case this can be seen with the Item example:
+```python
+junk = default.Item()
+assert junk.description != 'Junk'
```
dnwpark, line 112: No attribute error here?

elprans (approved):

PR #9088, opened 2025-10-14
Stop enforcing serializable isolation writes touching `id`

jackfischer:


edb/edgeql/compiler/conflicts.py
```diff
@@ -938,6 +938,12 @@ def check_for_isolation_conflicts(
                     root_subject_obj = root_subject
 
                 if isinstance(root_subject_obj, s_objtypes.ObjectType):
+                    # Skip abstract types like BaseObject or Object, as they
+                    # don't have real postgres tables and their constraints
+                    # don't require cross-table trigger validation.
+                    if root_subject_obj.get_abstract(schema):
+                        continue
```
msullivan, line 945: Also this I think might fix the issue you have had, but at the expense of introducing bugs. Abstract types do still need cross-table validation (between their children)

PR #9093, opened 2025-10-16
Update assert_single description for clarity

spacemanspiff2007:


docs/reference/stdlib/set.rst
```diff
@@ -544,7 +544,7 @@ Sets
 
     .. index:: cardinality, singleton
 
-    Checks that the input set contains no more than one element.
+    Checks that the input set is empty or contains no more than one element.
```
scotttrinh: Maybe even more clear:


```suggestion
    Checks that the input set has exactly zero or one elements.
```

wdyt?

scotttrinh (approved):
Seems fine, thanks for the suggestion!

PR #9096, opened 2025-10-17
Fix rvar lookup when applying type intersection to aliases and iterators

### LINKED ISSUE #9092: ISE when applying shape to for statement containing type intersection
Given a schema:
```edgeql
create type Obj_A {
    create property a: int64;
};
create type Obj_B {
    create property b: int64;
};
create type Obj_AB extending Obj_A, Obj_B {
    create property ab: int64;
};
```

The following queries:
```edgeql
FOR obj_a IN (default::Obj_A) UNION (obj_a [is default::Obj_B] {a});
SELECT (FOR obj_a IN (default::Obj_A) UNION (obj_a [is default::Obj_B])) {a};
```

produce an ISE:
```
edb.errors.InternalServerError: no rvar in intersection matches path id (__derived__::obj_a@w~1).>indirection[IS __derived__::(__derived__:obj_a@w~1 & default:Obj_B)].>a[IS std::int64]
```

dnwpark:
close #9092

Given a schema:
```edgeql
create type Obj_A {
    create property a: int64;
};
create type Obj_B {
    create property b: int64;
};
create type Obj_AB extending Obj_A, Obj_B {
    create property ab: int64;
};
```

Fixes the following queries:
```edgeql
FOR x IN Obj_A UNION (x [is Obj_B] {a});
SELECT (FOR x IN Obj_A UNION (x [is Obj_B])) {a};
WITH x IN Obj_A SELECT x [is Obj_A] {a};
```

### DIFF
--- a/edb/pgsql/compiler/pathctx.py
+++ b/edb/pgsql/compiler/pathctx.py
@@ -470,11 +470,8 @@ def _find_rvar_in_intersection_by_typeref(
 
     assert component_rvars
 
-    pid_rptr = path_id.rptr()
-    if pid_rptr is not None:
-        if pid_rptr.material_ptr is not None:
-            pid_rptr = pid_rptr.material_ptr
-        tref = pid_rptr.out_source
+    if src_path := path_id.src_path():
+        tref = src_path.target
     else:
         tref = path_id.target
 


msullivan (approved):
The new way seems more like what it should be, but I am a bit curious why it was that way before. (And *should* the things match??)
Not sure it's worth the digging though

PR #9097, opened 2025-10-18
Implement sum for durations

### LINKED ISSUE #7403: Implementing `sum(SET OF duration)`
I'd really like to be able to call sum on durations directly.

Currently you have to lossily call totalseconds and then sum the number and then convert back to a duration which isn't ideal.

Happy to try implementing this if I can get a pointer where the implementation at runtime for sum is. I'd guess a new overload needs adding and something writing in https://github.com/edgedb/edgedb/blob/37317eaef98c89d57392ebc2d966c8a7f7036d5d/edb/tools/experimental_interpreter/basis/std_funcs.py#L94.

Gobot1234:
closes #7403

### DIFF
--- a/docs/reference/stdlib/set.rst
+++ b/docs/reference/stdlib/set.rst
@@ -644,7 +644,7 @@ Sets
 
     .. index:: aggregate
 
-    Returns the sum of the set.
+    Return the arithmetic sum of values in a set.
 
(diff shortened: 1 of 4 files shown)


aljazerzen:
I've reworded the description, otherwise this looks good. Thanks!

aljazerzen (approved):

PR #9099, opened 2025-10-20
Reference docs for QB.

vpetrovykh:
I've updated the reference for `Client` as well as added some refernce docs for the reflected models/query-builder (in codegen section).

### DIFF
--- a/docs/reference/using/python/api/codegen.rst
+++ b/docs/reference/using/python/api/codegen.rst
@@ -149,6 +149,81 @@ You can use the standard |Gel| client created by either :py:func:`~gel.create_cl
 
     There are also some common methods and fields available to all reflected models.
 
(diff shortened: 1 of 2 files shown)


docs/reference/using/python/api/codegen.rst
```diff
@@ -89,33 +89,484 @@ Models
...
+            ppl_elast = client.query(
+                default.Person.order_by(email=("asc", "empty last"))
+            )
+
+        It is also possible to pass valid query builder expressions as *exprs* to specify the order, if some non-trivial sub-query is required:
+
+        .. code-block:: python
+
+            # Order people by name length, then by name
+            res = client.query(
```
elprans, line 457: Need an example of specifying direction with expressions.

PR #9101, opened 2025-10-21
Add a `.?>link` operator to dereference link while suppressing "hidden by policy"

### LINKED ISSUE #8522: Add an intrinsic to mask the "required link foo is hidden by access policy" errors
We currently wrap references to required links with `assert_exists` if the target is covered by an access policy.  This is done to avoid breaking the cardinality invariant in case the access policy masks all target objects.  However, in certain contexts, such as autogenerated queries (of which the UI is a heavy user) this necessitates potentially expensive workarounds (such as avoiding the link reference by computing the referenced object via a backlink combined with an explicit join).

We need an explicit escape hatch to remove the need for the workarounds.  I suggest something like `try()` or `optional()`,  e.g `select Foo { req_made_optional := optional(.required_link_possibly_masked_by_ap) }`

msullivan:
Users (and us) occasionally have problems dealing with "required link
hidden by access policy" errors. They can be worked around often by
using backlinks and an explicit join, but that that is kind of
horrible and seems to sometimes be slow.

Issue #8522 proposes adding an intrinsic function to mask the error,
but I really hate that. I don't think that a simple function call
ought to be impacting the semantics of the expression contained
within, and I think it's really unclear what its behavior would be in
more complex cases.

Instead, add an operator, `Obj.?>link` that does an *optional*
dereference of `link`.  The result is always optional, and it
suppresses any "hidden by policy" error.

I'm happy to argue about syntax if anyone has something they like
more. I also considered `.?` but that was worse.

Closes #8522, but not in the way proposed there.

### DIFF
--- a/edb/edgeql-parser/src/tokenizer.rs
+++ b/edb/edgeql-parser/src/tokenizer.rs
@@ -78,7 +78,7 @@ pub enum Kind {
     Coalesce,         // ??
     Namespace,        // ::
     BackwardLink,     // .<
-    OptionalLink,     // .?
+    OptionalLink,     // .?>
     FloorDiv,         // //
     Concat,           // ++
     GreaterEq,        // >=


edb/edgeql-parser/src/tokenizer.rs
```diff
@@ -78,6 +78,7 @@ pub enum Kind {
     Coalesce,         // ??
     Namespace,        // ::
     BackwardLink,     // .<
+    OptionalLink,     // .?
```
aljazerzen:
```suggestion
    OptionalLink,     // .?>
```

aljazerzen (approved):

msullivan:
> What about: `Object.link?` meaning that the `?` makes the previous thing optional? Also, how do we plan on actually using this in EdgeQL, assuming there is some declared link on the type, will we support redefining it as optional using this syntax?
> 
No, I don't think we can allow that.

PR #9104, opened 2025-10-23
Configurable query cache size

### LINKED ISSUE #9095: Query cache sizes should be configurable
Right now they are hardcoded to be 1000

aljazerzen:
Closes #9095

TODO:
- [x] ~do we want to have a separate option for gel native proto cache and SQL proto?~ (no)
- [x] ~tests~ (I don't know how to test this. Maybe set it to 1, restart the server, execute query A, execute query B, validate that executing query A triggers re-compilation)
- [ ] docs

### DIFF
--- a/docs/cloud/http_gql.rst
+++ b/docs/cloud/http_gql.rst
@@ -78,7 +78,7 @@ HTTP
 ----
 
 - :ref:`Overview <ref_edgeql_http>`
-- :ref:`ref_edgeqlql_protocol`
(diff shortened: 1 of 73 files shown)


msullivan (approved):

msullivan (commented):
Joint cache is probably fine.

I'd test manually and not worry too much about automated testing I guess?

I am curious what happens if the cache size is decreased below what is persisted though.

PR #9105, opened 2025-10-23
Add note about `overloaded` to inheritance docs

scotttrinh:
This concept only applies to inheritance, so it's a bit strange that it didn't get mentioned here, but only in the properties documentation.

Question: is there anything else to note about `overloaded` that isn't already covered in the docs? I was wondering if we have any constraints on how you can overload like not being able to change the type, or no `required` -> `optional` or something like that?


Gobot1234:
Only thing I can think of that's maybe worth mentioning is that it doesn't work on computeds

PR #9107, opened 2025-10-25
VoyageAI improvements

fzowl:
Adding VoyageAI voyage-3.5 model
Adding VoyageAI voyage-context-3 model

### DIFF
--- a/edb/lib/ext/ai.edgeql
+++ b/edb/lib/ext/ai.edgeql
@@ -596,7 +596,7 @@ CREATE EXTENSION PACKAGE ai VERSION '1.0' {
         alter annotation
             ext::ai::embedding_model_max_batch_tokens := "32000";
         alter annotation
-            ext::ai::embedding_model_max_output_dimensions := "2048";
+            ext::ai::embedding_model_max_output_dimensions := "1024";
         alter annotation
             ext::ai::embedding_model_supports_shortening := "true";
     };
@@ -613,7 +613,7 @@ CREATE EXTENSION PACKAGE ai VERSION '1.0' {
         alter annotation
             ext::ai::embedding_model_max_batch_tokens := "32000";
         alter annotation
-            ext::ai::embedding_model_max_output_dimensions := "2048";
+            ext::ai::embedding_model_max_output_dimensions := "1024";
         alter annotation
             ext::ai::embedding_model_supports_shortening := "true";
     };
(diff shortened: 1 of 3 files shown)


edb/lib/ext/ai.edgeql
```diff
@@ -562,6 +583,167 @@ CREATE EXTENSION PACKAGE ai VERSION '1.0' {
...
+            ext::ai::embedding_model_max_batch_tokens := "32000";
+        alter annotation
+            ext::ai::embedding_model_max_output_dimensions := "2048";
+        alter annotation
+            ext::ai::embedding_model_supports_shortening := "true";
+    };
+
+    create abstract type ext::ai::Voyage3EmbedModel
+        extending ext::ai::EmbeddingModel
+    {
```
dnwpark, line 712: Add something like
```edgeql
        create annotation std::deprecated :=
        "This model is noted as a legacy model in the VoyageAI docs.";
```

edb/lib/ext/ai.edgeql
```diff
@@ -562,6 +583,167 @@ CREATE EXTENSION PACKAGE ai VERSION '1.0' {
...
+            ext::ai::embedding_model_max_input_tokens := "32000";
+        alter annotation
+            ext::ai::embedding_model_max_batch_tokens := "32000";
+        alter annotation
+            ext::ai::embedding_model_max_output_dimensions := "1024";
+    };
+
+    create abstract type ext::ai::Voyage3LiteEmbedModel
+        extending ext::ai::EmbeddingModel
+    {
```
dnwpark, line 729: Another old model

edb/lib/ext/ai.edgeql
```diff
@@ -562,6 +583,167 @@ CREATE EXTENSION PACKAGE ai VERSION '1.0' {
...
+        alter annotation
+            ext::ai::model_name := "voyage-context-3";
+        alter annotation
+            ext::ai::model_provider := "builtin::voyageai";
+        alter annotation
+            ext::ai::embedding_model_max_input_tokens := "32000";
+        alter annotation
+            ext::ai::embedding_model_max_batch_tokens := "320000";
+        alter annotation
+            ext::ai::embedding_model_max_output_dimensions := "1024";
```
dnwpark: Should be 2048?

dnwpark (commented):
Thanks for the contribution! Here are some notes from my testing too.

The max vector size is currently 2000, this causes the indexing to get truncated.

Also it seems that the default embedding length for a lot of models is 1024. Some kind of parameter seems to be missing from the request, and this causes an error when searching text. It's probably fine to just use 1024 for max size.

fzowl:
@dnwpark Sorry, i missed your review and comments. Can you please check the changes now?

PR #9111, opened 2025-10-30
Add Discord OAuth `prompt` parameter with default

scotttrinh:
A reimagining of #8738 where we pass through a `prompt` string in Discord's OAuth flow or default it to `"consent"`. More reading available here https://discord.com/developers/docs/topics/oauth2#authorization-code-grant-authorization-url-example

### DIFF
--- a/edb/server/protocol/auth_ext/oauth.py
+++ b/edb/server/protocol/auth_ext/oauth.py
@@ -87,9 +87,9 @@ class Client:
                     *provider_args,
                     **provider_kwargs,
                 )
-            case (prov_name, issuer_url) if isinstance(issuer_url, str):
+            case (provider_name, str(issuer_url)):
                 self.provider = base.OpenIDConnectProvider(
-                    prov_name,
+                    provider_name,
                     issuer_url,
                     *provider_args,
                     **provider_kwargs,


edb/server/protocol/auth_ext/oauth.py
```diff
@@ -84,9 +87,9 @@ def __init__(
                     *provider_args,
                     **provider_kwargs,
                 )
-            case (provider_name, str(issuer_url)):
+            case (prov_name, issuer_url) if isinstance(issuer_url, str):
```
msullivan: Why was this change made? The old version is nicer --- was there some typechecking problem?

msullivan (approved):
Sure, seems reasonable.
If there are any backport problems we might skip it still though.

edb/server/protocol/auth_ext/oauth.py
scotttrinh: Yeah, I meant to call this out: I got a type check error saying you cannot use `str` for match because it doesn't not have `__match_args__`.  Is that wrong? I can separate this out and try to write up a test case for this if that's better.

scotttrinh: ya know what, I have tests for this already, I'm going to revert this and wait for an actual reason to make this change.

PR #9123, opened 2025-11-25
add a note on how to reset default Cloud password

nsidnev:


docs/reference/running/deployment/note_cloud_reset_password.rst
```diff
@@ -0,0 +1,7 @@
+.. note:: Gel Cloud: Reset the default password for the admin role
+
+    If you want to dump an existing Gel Cloud instance, you will need to change the automatically generated password for the standard admin role - ``edgedb`` or ``admin``. To do this, execute the following query in the instance:
```
fmoor: Does the password need to be changed to the one used in the dump?

nsidnev: Yes, the password should be the same as in a new non-cloud instance so that the tooling can continue to work properly during the dump/restore. Do you have a suggestion on how to phrase this more clearly?

fmoor: Maybe something like this?

> If you want to restore a dump the target instance must have the same admin password as the dump. To change the password on the target instance do this, ...

fmoor: I guess it should also call out that the admin user must be the same as the dump too.

docs/reference/running/deployment/note_cloud_reset_password.rst
```diff
@@ -0,0 +1,9 @@
+.. note:: Gel Cloud: Reset the default password for the admin role
+
+    If you want to dump an existing Gel Cloud instance and restore it to a new self-managed instance, you need to change the automatically generated password for the default admin role - ``edgedb`` or ``admin``.
+    The administrator role name and its password used in the dump/restore process must be the same in both the Cloud and self-managed instances for the Gel tooling to continue functioning properly.
```
fmoor: This is a bit nit picky. Seems like the documentation should be applicable more broadly than just the cloud.

```suggestion
    The administrator role name and its password used in the dump/restore process must be the same in both the instance dumped from and the instance restored to for the Gel tooling to continue functioning properly.
```

fmoor (approved):

